PR #1438, opened 2026-04-29
Adding beads issue management

pirapira:
This might make the codebase even more agent friendly.


pirapira:
maybe not a good idea

Issue #1045, opened 2026-04-22
Import hygiene sweep via lake exe shake


pirapira:
## Context

A local `lake exe shake EvmAsm` run finds ~660 lines of import-hygiene suggestions across the codebase. Many of these are legitimately wasteful (e.g., `import Mathlib.Tactic` pulling in hundreds of modules when only 3 are used). Fixing them should meaningfully shorten cold-build times both locally and on CI.

PR #1044 does the first slice — `Rv64/HalfwordOps.lean` and `Rv64/WordOps.lean`.

## How shake works here

1. `lake exe shake EvmAsm` runs after a successful `lake build`.
2. First run creates `scripts/noshake.json` with an empty `{}` (shake expects this config file).
3. For each file, shake suggests `remove #[...]` (imports not actually used) and sometimes `add #[...]` (fine-grained imports pulled in transitively — good to make explicit).

## ⚠️ Shake is not always correct

Its raw suggestions must be verified. For example, in PR #1044 shake said to drop `Mathlib.Tactic` with **no** replacement from `HalfwordOps.lean`, but the file uses `interval_cases` / `fin_cases` via that import. The correct fix was replacing with `Mathlib.Tactic.IntervalCases` + `Mathlib.Tactic.FinCases` + `Mathlib.Data.Fintype.Basic`. Always `lake build` after each edit.

## Suggested batches (from the full shake output)

Rough groupings to split into multiple small PRs so each is easy to verify:

1. **`import Mathlib.Tactic` replacements** — at least 6 more files flagged (`ByteOps`, `Common`, `Arithmetic`, `Comparison`, `Properties`, etc.). Each likely needs a small explicit subset.

2. **Unused `SyscallSpecs` / `GenericSpecs` / `XSimp` / `RunBlock` imports** — many opcode `Program.lean` and `Spec.lean` files import these but don't use them (split-out Program/Spec refactor left stale imports behind).

3. **DivMod loop-body tier** — `LoopBodyN{1,2,3}.lean`, `LoopIterN{1,2,3}/*.lean`, `LoopComposeN{1,2,3}.lean`, `LoopUnifiedN{1,2,3}.lean`, and the `Compose/FullPathN{1,2,3,4}Loop*.lean` files all flag stale imports like `LoopBody`, `LoopDefs`, `LoopBodyN*`, `LoopComposeN*`. Likely ~20 files cleanable in a small batch.

[...]

PR #1482, opened 2026-04-30
chore(scripts): shake-filter to drop known false positives (#1045)

pirapira:
Adds `scripts/shake-filter.py` + docs to post-process `lake exe shake EvmAsm` output, dropping per-file blocks where shake is known to be wrong.

## Why

Past slices of #1045 (3, 4, 5, 6) repeatedly stalled because shake flags imports as unused even when the consumer relies on them indirectly. Categories of false positives, distilled from the audit in beads `evm-asm-o6y` / `evm-asm-pic`:

1. Spec-database tactics (`runBlock`, `xperm`, `xperm_hyp`, `xcancel`, `xsimp`, `seqFrame`, `liftSpec`) — load lemmas via attribute, not direct reference.
2. `@[spec_gen_*]` declarations — same problem from the producer side.
3. `*_spec_gen_within` / `*_spec_within` lookup names (CallingConvention, etc.).
4. Custom simp attributes (`rv64_addr`, `divmod_addr`, `reg_ops`, `byte_alg`).
5. `Word` notation re-export from `EvmAsm.Rv64.Basic`.

## Effect on current tree

Raw output: 822 lines / 236 file blocks.
Filtered:   336 lines /  84 file blocks (152 dropped, 64%).

All five files previously audited as pure noise (CallingConvention, DivMod/Spec, DivMod/SpecCall, SyscallSpecs, InstructionSpecs) are correctly dropped.

## Usage

```
lake build
lake exe shake EvmAsm 2>/dev/null | scripts/shake-filter.py
```

Add `--show-dropped` to see what was filtered out.

## Notes

- Filter is **conservative on the drop side** by design: better to skip a real suggestion than waste a slicer on a false positive. Markers can be tightened later.
- Does not modify shake itself; pure stdout post-processor.
- Pure tooling — no impact on `lake build`.

Refs: #1045 (issue), beads evm-asm-sf0 (this slice), evm-asm-o6y, evm-asm-pic.

### DIFF
--- a/scripts/shake-filter.md
+++ b/scripts/shake-filter.md
@@ -1,0 +1,87 @@
+# shake-filter — known false-positive filter for `lake exe shake`
+
+`lake exe shake EvmAsm` produces a long list of "remove this import"
+suggestions. Past slices of #1045 repeatedly hit the same problem: many
+of those suggestions are wrong because shake does not see lemmas /
+notations / spec-database entries that are referenced indirectly via
+attributes or tactic macros.
+
+`scripts/shake-filter.py` post-processes shake's output and drops every
+file block whose source contains a documented false-positive marker
+(custom simp attributes, `runBlock` / `xperm` / `xsimp` / `seqFrame`
+tactic invocations, `@[spec_gen_*]` attribute use, `Word` notation, etc.).
+
+## Usage
+
+```bash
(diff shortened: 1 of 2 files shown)


pirapira:
I like the direction.

Issue #1435, opened 2026-04-29
xperm etc should ignore pure assertions


pirapira:
When xperm or xperm_hyp is used, the main interest is in matching up the resources. So these tactics should strip away the pure conditions. If the goal requires the pure conditions, some very light tactics should be tried.

pirapira:
## Status update — work to support pure-aware xperm has landed

The original ask ("xperm should ignore pure assertions, with light tactics for pure side conditions") is addressed by a small family of tactics that landed across recent PRs:

**Tactics shipped**

- `extract_pure` (PR #1464 / #1473) — peel a pure atom `⌜P⌝` from a separation hypothesis as a Prop side-fact, leaving the resource tail.
- `xperm_pure` (PR #1483) — pure-aware sibling of `xperm_hyp` that destructures pures from both sides before AC matching, then closes the side conditions.
- `drop_pure` (PR #1505 / #1523) — strip a pure leaf from a hypothesis chain so the rest can be fed to plain `xperm_hyp`. Long right-assoc chains are now handled by #1523.

**Adoption**

- RLP long-loop closure call sites: PR #1502.
- DivMod Div128Step1v2 Flavor-A bridges (the trickiest sites in the original survey): PR #1560.

The slice-1 survey (`EvmAsm/Rv64/Tactics/XPermPure.lean`) and the friction notes (`evm-asm-kvl`, fix `#1523`) recorded the design rationale, and all child beads tasks under this issue are now closed. The remaining ad-hoc `obtain ⟨_, _⟩` patterns can be migrated incrementally as new sites are touched, but the tactic surface itself is complete.

Suggesting close.

Authored by @pirapira; implemented by Hermes-bot (evm-hermes).

PR #1524, opened 2026-04-30
fix(droppure): bubble mid-chain pures via Assertion-eq (#1435)

pirapira:
Adds an Assertion-level equality `sepConj_pure_left_comm_eq : (P ** (⌜Q⌝ ** R)) = (⌜Q⌝ ** (P ** R))` to `drop_pure`'s simp set. Because it's stated as `Assertion = Assertion` (no `s` applied), simp's standard congruence on `**` can rewrite it at any depth in a right-associated chain. Repeated application bubbles a deep mid-chain pure one step closer to the head per fire; `sepConj_pure_left` then peels it into a left `∧` at the outermost level.

## What this fixes

Per @pirapira's refined diagnosis on beads `evm-asm-22a`: the prior `drop_pure` simp set used iff lemmas of the form `∀ s, (…) s ↔ (…) s`, which simp can only fire at the outermost `_ s` position. Once the chain is right-associated as `A₁ ** (A₂ ** (… ** (⌜P⌝ ** Aₖ)))` with the pure at depth ≥ 5, those rules can no longer reach it. The Div128Step1v2.lean Flavor-A sites threaded a 10-atom chain with `⌜rhatHi2 ≠ 0⌝` at depth 9 — `drop_pure` left the pure in place and the follow-up `xperm_hyp` failed.

Two regression tests added: a 6-atom mid-pure at depth 3, and the 10-atom mid-pure at depth 9 reproducer.

## Why no swap form

A natural extension would be `sepConj_pure_swap_eq : (R ** ⌜Q⌝) = (⌜Q⌝ ** R)`. Combined with `left_comm_eq` it loops on adjacent pures (`R ** (⌜P⌝ ** ⌜Q⌝)` cycles forever in simp's term-order normalisation). Single-rule approach terminates and handles the priority case. Tail-deep pures (`R₁ ** … ** Rₙ ** ⌜P⌝` at depth ≥ 2) remain a followup — likely needs a custom tactic that walks the chain explicitly rather than a simp set.

## Other changes

- Dropped the `← sepConj_assoc'` rewrite from the simp set: chains parse right-associative already, so re-associating left was unnecessary and made the prior set non-terminating once the new Assertion-eq was added.

## References

- Refs #1435.
- Beads `evm-asm-zfd` (sub-slice of `evm-asm-22a` / `evm-asm-eyc`).
- File-level docstring on `sepConj_pure_left_comm_eq` documents the design tradeoff for future maintainers.

### DIFF
--- a/EvmAsm/Rv64/Tactics/DropPure.lean
+++ b/EvmAsm/Rv64/Tactics/DropPure.lean
@@ -72,6 +72,42 @@ theorem sepConj_pure_right_swap {P : EvmAsm.Rv64.Assertion} {Q : Prop} :
   rw [EvmAsm.Rv64.sepConj_pure_right]
   exact And.comm
 
+/-! ### Assertion-level pure-bubble equality
+
+The `sepConj_pure_*` iff lemmas above are stated as
+`∀ s, (…) s ↔ (…) s`. `simp` can only fire those at the *outermost*
+assertion-applied-to-state position; once the chain is right-associated as
+`A₁ ** (A₂ ** (… ** (Aₖ₋₁ ** (⌜P⌝ ** Aₖ₊₁))))` and a pure sits at depth
+`k > 1`, those rules can no longer reach it because they would have to
+descend through opaque `sepConj` applications — none of the helper
+biconditionals are stated under the `**` head, so simp's congruence does
+not propagate them.
+
+The equality below is at `Assertion = Assertion` (no `s` applied), so
+simp's standard congruence on the binary `sepConj` operator *can* rewrite
+it at any depth in a right-associated chain. Repeated application bubbles
+a deep pure (`(R ** (⌜Q⌝ ** S))`) one step closer to the head per fire;
+once at the head, `sepConj_pure_left` (an iff at outermost `_ s`) peels
+it into a left `∧`.
+
+Why only this one Assertion-eq and not also a swap form
+(`(R ** ⌜Q⌝) = (⌜Q⌝ ** R)`): a swap form combined with `left_comm_eq`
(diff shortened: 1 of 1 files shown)


pirapira:
Heads-up from Hermes-bot: main commit 30016dd9 'fix(tactic): drop_pure handles long right-assoc chains (#1435)' landed a parallel/competing implementation of the same fix, so this branch now has a non-syntactic merge conflict (two designs for the same thing). Marking draft and filing beads task evm-asm-ole0 for human decision (close as superseded vs rework over main's ac_rfl approach). Authored by @pirapira; implemented by Hermes-bot (evm-hermes).

pirapira:
Superseded by #? (commit `30016dd9` already on `main`: *fix(tactic): drop_pure handles long right-assoc chains (#1435)*).

That commit takes a strictly more general `ac_rfl`-based approach to the same problem this PR's `sepConj_pure_left_comm_eq` simp lemma was solving:

- depth-agnostic (handles trailing pures and ≥3 pures, which this PR explicitly defers as followups)
- smoke tests cover 8-, 10-, 10-trailing-pure, and 10-three-pure shapes — superset of this PR's two new tests
- avoids the loop-on-adjacent-pures hazard documented in this PR's docstring

Merging `origin/main` produces a conflict in `EvmAsm/Rv64/Tactics/DropPure.lean` because both branches rewrite the same `evalDropPure` definition. There's nothing useful to salvage on top of `main`'s implementation, so I'm flagging this for closure rather than rebasing.

@pirapira — please close if you agree.

Authored by @pirapira; reported by Hermes-bot (evm-hermes).

Issue #100, opened 2026-04-01
Environment context layout (Environment.lean)


pirapira:
## Goal

Define the memory layout for the EVM execution environment (msg.caller, block.number, etc.) and the `envIs` separation logic assertion.

## Prerequisites

- None

## Context

Many EVM opcodes read from the execution context (ADDRESS, CALLER, CALLVALUE, TIMESTAMP, etc.). This data needs to be laid out in RISC-V memory at a known base address. The layout should follow the Python reference spec structure in `execution-specs/src/ethereum/forks/shanghai/vm/`.

## Approach

1. Create `Evm64/Environment.lean`
2. Define memory layout struct with fixed offsets:
   - `msg.caller` (20 bytes, zero-extended to 256 bits = 32 bytes)
   - `msg.value` (256 bits = 32 bytes)
   - `msg.data` pointer + length (calldata)
   - `block.number`, `block.timestamp`, `block.basefee`, `block.gaslimit`, `block.coinbase`, `block.difficulty/prevrandao`
   - `tx.origin`, `tx.gasprice`
   - `chainid`, `selfbalance`
   - `address` (current contract)
   - `returndatasize`, `returndata` pointer
3. Define `envIs (base : Addr) (env : EvmEnv) : Assertion` using `memIs` for each field
4. Define `EvmEnv` structure with all fields

## Acceptance Criteria

- [ ] `EvmEnv` structure defined with all required fields
- [ ] Memory layout with fixed offsets documented
- [ ] `envIs` assertion defined and decomposable
- [ ] `lake build` succeeds

## Milestone
C: EVM Infrastructure Models

pirapira:
Status update — proposing close.

All four implementation slices for the Environment context layout have landed on `main`:

- slice 1 — define `EvmEnv` structure with all required fields: PR #1564 (merged)
- slice 2 — fix `EvmEnv` memory layout offsets and document them: PR #1581 (merged)
- slice 3 — define `envIs` assertion decomposable into per-field `memIs`: PR #1643 (merged)
- slice 4 — wire `Environment.lean` into Evm64 umbrella + acceptance smoke test: PR #1671 (merged)

Acceptance criteria from this issue:
- [x] `EvmEnv` structure defined (caller, value, calldata ptr/len, address, returndata, block fields, tx fields)
- [x] Memory layout offsets fixed and documented
- [x] `envIs` separation-logic assertion defined and decomposable into per-field `memIs`
[...]

PR #1643, opened 2026-05-01
feat(env): define envIs assertion for EvmEnv layout (#100 slice 3)

pirapira:
Slice 3 of #100 (beads evm-asm-3fr7).

Defines `envIs base env` in `EvmAsm/Evm64/Environment/Assertion.lean` as a right-associative `sepConj` of per-field cells using the offsets from `Environment/Layout.lean`:

- 32-byte cells via `evmWordIs` for `EvmWord` and 160-bit `Address` fields (addresses zero-extended to 256 bits via `addrAsWord`).
- 8-byte cells via `memIs` (`↦ₘ`) for pointer/length fields.

Provides:
- `envIs_unfold` (definitional rfl unfold into the 17-atom chain).
- `envIs_caller_split` — canonical decomposition lemma rotating the `caller` cell to the head of the sepConj (closed by `ac_rfl`).
- `pcFree_envIs` + `Assertion.PCFree` instance.
- `envCells = 56` and `envCells_eq` for the dword footprint.

Wires the new module into `EvmAsm/Evm64.lean`. Slice 4 (`evm-asm-xbyi`) covers the umbrella smoke test.

Acceptance: `lake build` green; at least one decomposition lemma proved (`envIs_caller_split`); no sorry.

### DIFF
--- a/EvmAsm/Evm64.lean
+++ b/EvmAsm/Evm64.lean
@@ -62,6 +62,7 @@ import EvmAsm.Evm64.CallingConvention
 -- Execution-context structure (#100 slice 1; envIs assertion lands in slice 3)
 import EvmAsm.Evm64.Environment
 import EvmAsm.Evm64.Environment.Layout
+import EvmAsm.Evm64.Environment.Assertion
 
 -- EVM memory model (issue #99)
 import EvmAsm.Evm64.Memory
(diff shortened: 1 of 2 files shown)


pirapira:
I don't think this split lemma is useful. The content of 'rest' needs to be shown to the caller. Otherwise the caller loses track of resources

pirapira:
Addressed in 57bec861: dropped the existential and exposed the leftover sepConj as `envIsCallerRest base env` (top-level `def`). `envIs_caller_split` now reads as a direct equation:

```
envIs base env = (evmWordIs (...callerOff) env.caller ** envIsCallerRest base env)
```

so a `CALLER` handler framing on the head keeps explicit ownership of the other 16 env-field cells. Build green locally.

Issue #949, opened 2026-04-21
Add weekly benchmark workflow to track lake build times


amit0365:
Got the idea from https://github.com/Beneficial-AI-Foundation/curve25519-dalek-lean-verify. This would helps us catch regressions and keep a faster build. If that sounds good, I can work on adding it to the workflow.

pirapira:
At the moment I'm not particularly interested. For local edits, the build along the import path to the target file matters, not the total build time. The CI uses caching heavily so the total build time matters only on few PRs that touch fundamental definitions.

pirapira:
## Slice 1 design (beads `evm-asm-c7q`) — benchmark workflow scope

Surveyed `Beneficial-AI-Foundation/curve25519-dalek-lean-verify/.github/workflows/`:
they run three workflows (`bench-main.yml`, `bench-pr.yml`, `bench-weekly.yml`)
that all delegate to a private benchwarmer service: each invokes
`bash "$RUNNER_SRC/bench.sh" "$BENCH_LIBRARY_NAME" "$BENCH_API_ENDPOINT"` with
`BENCH_AUTH_TOKEN` (private secrets), then `curl`s a report endpoint and pastes
markdown into an issue or PR comment. The actual measurement uses
[`lakeprof`](https://github.com/Kha/lakeprof) (per-file Lean compile profiling).
Without their `BENCH_API_ENDPOINT` / `BENCHWARMER_RUNNER_PATH` infrastructure
the workflows don't run standalone — we cannot copy their YAML directly.

### Decisions for slice 2

- **Cadence**: weekly cron `0 6 * * 1` (Mondays 06:00 UTC) + `workflow_dispatch`
  for manual runs. Matches curve25519-dalek-lean-verify's weekly cadence and
  avoids per-PR cost (we have many PRs).
- **What to measure** (kept minimal for slice 2):
  1. Total `lake build` wall-clock time via `/usr/bin/time -v`.
  2. Peak RSS (also from `/usr/bin/time -v`).
  3. Mathlib cache fetch time as a separate step (not part of the build
     metric — we want to track our own work, not Mathlib's).
  - **Per-target / per-file timing** via `lakeprof` deferred — nice-to-have for
     a follow-up, not in slice 2.
- **`lake exe cache get` ordering** (mandatory per AGENTS.md): run
  `lake exe cache get` BEFORE the timed `lake build` step. Otherwise the first
[...]

PR #1669, opened 2026-05-01
ci(benchmark): surface regressions via benchmark-tracking issue (#949 slice 4)

pirapira:
Implements #949 slice 4 (beads evm-asm-9o8v). After the weekly benchmark workflow appends a new record to the `benchmark-history` branch, this PR adds a final step that:

- fetches the latest two records from `benchmark-history`;
- computes wall-clock and peak-RSS deltas via `scripts/benchmark-report.py`;
- posts or updates a single GitHub issue (label `benchmark-tracking`, title `Weekly benchmark report`) with the snapshot plus diff and a soft warning marker when growth exceeds `REGRESSION_PCT` (default 10%).

Per `docs/benchmark-workflow-design.md` (c) this is informational only — there is no CI gate on regression. The user reads the weekly report and decides whether the regression is real.

The Python helper is checked-in (rather than inlined into the YAML `run:` block) because indented heredocs are notoriously fragile. Smoke-tested locally on synthetic two-row and one-row `history.jsonl` inputs (regression and first-run paths).

The workflow already had `permissions: contents: write`; this PR adds `issues: write` so the cron run can create/edit the tracking issue.

Refs: #949 / beads `evm-asm-9o8v`.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -25,6 +25,7 @@ concurrency:
 
 permissions:
   contents: write
+  issues: write
 
 jobs:
   benchmark:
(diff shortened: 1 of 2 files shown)


pirapira:
This is over engineering. Not needed

Issue #120, opened 2026-04-01
RLP RISC-V decoder (EL.3)


pirapira:
## Goal

Implement a verified RLP decoder in RISC-V that reads RLP-encoded data from `read_input` and produces decoded items in memory.

## Prerequisites

- EL.1 RLP specification (done)
- EL.2 Byte-level infrastructure (done)

## Context

The STF reads RLP-encoded blocks via `read_input`. This decoder bridges the pure Lean RLP spec (EL.1) with RISC-V execution. It is the first step of the STF pipeline.

## Approach

Six-phase pipeline as outlined in PLAN.md:

1. **Phase 1: Prefix classifier** — cascade BLTUs, 5 exits (single byte, short string, long string, short list, long list)
2. **Phase 2: Length extraction** — short: inline from prefix byte; long: big-endian loop over length-of-length bytes
3. **Phase 3: Single-item flat decode** — byte strings only, pointer+length output (zero-copy)
4. **Phase 4: HINT_READ integration** — load RLP input into memory buffer via HINT_READ syscall
5. **Phase 5: Recursive list decode** — iterative with explicit stack (not RISC-V call stack)
6. **Phase 6: Top-level pipeline** — HINT_READ → decode → output items

Build on `ByteOps.lean` (LBU/SB specs) and LP64 calling convention.

## Acceptance Criteria

- [ ] All 6 phases implemented as RISC-V programs
- [ ] Per-phase specs proved (0 sorry)
- [ ] Round-trip: decode(encode(item)) = item (links to EL.1 Properties.lean)
- [ ] Canonical enforcement: reject non-canonical encodings
- [ ] `lake build` succeeds

## Milestone
G: Transaction Processing

pirapira:
Closing as **superseded**. Two of five criteria are unmet, and one of those should never be implemented as written.

## Against its own criteria

| criterion | state |
|---|---|
| all 6 phases as RISC-V programs | **partial** — prefix classify / length / single-item decode exist (`rlp_decode_short`, `rlp_decode_single`, `rlp_decode_singleton_validated`, `rlp_field_decode_aux`, `rlp_list_count_items`); **no general recursive list decoder with an explicit stack** (phase 5) |
| per-phase specs proved, 0 sorry | **partial** — `*_spec_within` specs exist for several primitives, not per-phase |
| round-trip `decode(encode(item)) = item` | ⛔ unmet — split out as **#11661** |
[...]

PR #2018, opened 2026-05-02
prove(rlp): add ninety-three-byte long string examples

pirapira:
Stacked on #2017.\n\nAdds executable EL/RLP coverage for a 93-byte long-form byte string:\n- decodeAux example for prefix 0xB8 with length byte 0x5D\n- top-level decode example for the same payload\n- encodeBytes example for the concrete payload\n\nLocal verification:\n- lake build EvmAsm.EL.RLP.Properties\n- rg forbidden-pattern scan in EvmAsm/EL/RLP/Properties.lean\n- scripts/check-file-size.sh --report\n- lake build\n\nRefs evm-asm-9fh7 and GH #120.\n\nAuthored by @pirapira; implemented by Codex.


pirapira:
Not very interesting

PR #2022, opened 2026-05-02
prove(rlp): add ninety-seven-byte long string examples

pirapira:
Stacked on #2021.

Adds executable decodeAux, decode, and encodeBytes examples for a 97-byte long-form RLP byte string (prefix 0xB8, length byte 0x61).

Verification:
- lake build

Refs evm-asm-w8lp and GH #120.


pirapira:
Not very interesting

PR #2028, opened 2026-05-02
prove(rlp): add one-hundred-three-byte long string examples

pirapira:
Stacked on #2027.

Adds executable decodeAux, decode, and encodeBytes examples for a 103-byte long-form RLP byte string (prefix 0xB8, length byte 0x67).

Verification:
- lake build

Refs evm-asm-fc6v and GH #120.


pirapira:
Not very interesting

PR #2036, opened 2026-05-02
prove(rlp): add one-hundred-eleven-byte long string examples

pirapira:
Stacked on #2035.

Adds executable decodeAux, decode, and encodeBytes examples for a 111-byte long-form RLP byte string (prefix 0xB8, length byte 0x6F).

Verification:
- lake build

Refs evm-asm-g1gt and GH #120.


pirapira:
Not very interesting

Issue #101, opened 2026-04-01
PUSH1-32 opcodes


pirapira:
## Goal

Implement and prove PUSH1 through PUSH32 EVM opcodes that read immediate bytes from the EVM code region.

## Prerequisites

- #98: EVM code region model (CodeRegion.lean)

## Approach

1. Create `Evm64/Push/Program.lean`: `evm_push (n : Nat)` reads n bytes from `code[evm_pc+1..evm_pc+n]`
2. Zero-extend bytes to 256 bits (right-aligned, big-endian per EVM spec)
3. Push result onto EVM stack
4. Use LBU to read individual bytes from code region
5. Parameterized spec: `evm_push_n_stack_spec (n : Nat)` for n ∈ [1, 32]
6. Follow `Dup.lean`/`Swap.lean` pattern for parameterized opcode

## Acceptance Criteria

- [ ] `evm_push` program defined (parameterized by n)
- [ ] Concrete tests for PUSH1, PUSH2, PUSH32
- [ ] `evm_push_n_stack_spec` proved (0 sorry)
- [ ] EVM PC advances by 1+n
- [ ] `lake build` succeeds

## Milestone
C: EVM Infrastructure Models

pirapira:
Fixed. PUSH0/PUSH1/PUSH2..32 all `✅ proven` (PROGRESS.md): `evm_push0_stack_spec_within`, `evm_push1_stack_spec_within`, `evm_push_stack_spec_within` (single proof generic over n=2..32, 31 byte-codes).

PR #2075, opened 2026-05-02
feat(push): prove generic byte-copy bridge

pirapira:
Adds a generalized PUSH byte-copy bridge instead of more concrete PUSH-width examples.\n\nSummary:\n- adds evm_push_byte_slice for arbitrary byte index i < n\n- adds evm_push_one_byte_spec_within lifting the existing LBU/SB byte-copy spec under evm_push_code\n\nVerification:\n- lake build EvmAsm.Evm64.Push\n- lake build\n- scripts/check-file-size.sh --report\n- git diff --check\n- forbidden-pattern scan on modified PUSH files\n\nRefs: evm-asm-f0f5.9, GH #101.\n\nAuthored by @pirapira; implemented by Codex.

### DIFF
--- a/EvmAsm/Evm64/Push/Spec.lean
+++ b/EvmAsm/Evm64/Push/Spec.lean
@@ -241,4 +241,65 @@ theorem push_one_byte_ofProg_spec_within
     codeDwordAddr dstDwordAddr codeOff dstOff base
     h_code_align h_code_valid h_dst_align h_dst_valid
 
+/-- Lift the one-byte PUSH copy spec under the full generic `evm_push n`
+    program for any byte index `i < n`. -/
+theorem evm_push_one_byte_spec_within
+    (n i : Nat) (hn : n ≤ 32) (hi : i < n)
+    (codePtr sp v7Old codeWord dstWordOld : Word)
+    (codeDwordAddr dstDwordAddr : Word) (base : Word)
+    (h_code_align :
+      alignToDword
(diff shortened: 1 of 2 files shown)


EvmAsm/Evm64/Push/Spec.lean
```diff
@@ -241,4 +241,65 @@ theorem push_one_byte_ofProg_spec_within
...
+        (sp + signExtend12 (BitVec.ofNat 12 (pushByteDstOffset n i))) =
+          dstDwordAddr)
+    (h_dst_valid :
+      isValidByteAccess
+        (sp + signExtend12 (BitVec.ofNat 12 (pushByteDstOffset n i))) = true) :
+    let codeOff := BitVec.ofNat 12 (pushByteSrcOffset i)
+    let dstOff := BitVec.ofNat 12 (pushByteDstOffset n i)
+    let subBase := base + BitVec.ofNat 64 (4 * (5 + 2 * i))
+    let byteZext :=
+      (extractByte codeWord (byteOffset (codePtr + signExtend12 codeOff))).zeroExtend 64
```
pirapira: Please avoid let bindings in the statement; put the postcondition in an irreducible definition.

pirapira: Addressed in 0601e12b: hoisted the postcondition into `@[irreducible] def evmPushOneBytePost` and removed the `let` bindings from the statement of `evm_push_one_byte_spec_within`. Unfolding is now driven by an explicit `evmPushOneBytePost_unfold` lemma at the final weakening step. CI is green.

PR #3251, opened 2026-05-08
docs(accelerators): add bridge coverage table

pirapira:
## Summary
- expand the zkvm_accelerators ADR coverage section into a per-symbol audit table
- list selector constants from Accelerators.SyscallIds for each accelerator
- map each C symbol to the current Lean payload bridge and ECALL/execution bridge modules

## Verification
- git diff --check

Note: this PR is intentionally not enqueued by me.


pirapira:
No need to say audit. This is just  authors reading and summarizing.

pirapira:
Renamed section from "Coverage audit table" → "Coverage table", dropped "audit" in the trailing note, and updated the PR title. Pushed as 1fbe5981.

Issue #90, opened 2026-04-01
SDIV and SMOD opcodes


pirapira:
## Goal

Implement and prove SDIV and SMOD (signed division and modulo) EVM opcodes.

## Prerequisites

- #89: DIV/MOD full path merging + stack spec

## Approach

1. Create `Evm64/SDiv/Program.lean`: Check signs of both operands, compute unsigned DIV/MOD, apply sign correction
2. Per EVM spec: `SDIV(a,b)` = `sign(a)*sign(b) * DIV(|a|,|b|)` with special case `SDIV(-2^255, -1) = -2^255`
3. `SMOD(a,b)` = `sign(a) * MOD(|a|,|b|)`
4. Reuse `evm_div`/`evm_mod` as subroutines via LP64 calling convention
5. Prove `evm_sdiv_stack_spec` / `evm_smod_stack_spec`

## Acceptance Criteria

- [ ] `evm_sdiv` program defined with tests
- [ ] `evm_smod` program defined with tests
- [ ] `evm_sdiv_stack_spec` proved (0 sorry)
- [ ] `evm_smod_stack_spec` proved (0 sorry)
- [ ] Special case `-2^255 / -1 = -2^255` handled correctly

## Milestone
B: Complete EVM Opcode Coverage

pirapira:
## SDIV/SMOD status (slice 6 — beads `evm-asm-ipi7`)

All design and slice work for #90 is shipped or merged:

**Slices complete (beads):**
- `evm-asm-rtt5` slice 1 — design survey (sign extraction, two's-complement abs, reuse `evm_div`/`evm_mod` via LP64 callable shim) — PR #2265
- `evm-asm-kyp6` slice 2 — `Evm64/SDiv` + `Evm64/SMod` directory skeletons per `OPCODE_TEMPLATE` — PR #2266
- `evm-asm-k2czq` `#334` — `SDiv/Layout.lean` + `SMod/Layout.lean` empty-layout pilots
- `evm-asm-kvs4` slice 3 — `EvmWord.sdiv` + `sdiv_correct` (incl. `-2^255 / -1 = -2^255` overflow edge case) — PR #2275
- `evm-asm-pc8g6` slice 5a — `EvmWord.smod` + `smod_correct` (sign-of-dividend semantics) — PR #2432
- `evm-asm-8pfe` — `evm_div_callable` / `evm_mod_callable` LP64 shims and function specs
[...]

PR #3584, opened 2026-05-10
feat(sdiv): prove conditional negate limb step

pirapira:
## Summary
- add a CodeReq handle for the named SDIV conditional-negation limb step
- prove the 5-instruction LD/XOR/ADD/SLTU/SD leaf spec when the carry input is held in a separate sign register
- prepare the first-limb piece needed to compose the full 21-instruction conditional-negation block

Depends on #3583.
Refs #90.
Bead: evm-asm-01uh.

## Verification
- lake build EvmAsm.Evm64.SDiv.LimbSpec
- scripts/check-file-size.sh
- rg -n "sorry|admit|set_option maxHeartbeats|set_option maxRecDepth" EvmAsm/Evm64/SDiv/LimbSpec.lean
- lake build

### DIFF
--- a/EvmAsm/Evm64/SDiv/LimbSpec.lean
+++ b/EvmAsm/Evm64/SDiv/LimbSpec.lean
@@ -46,6 +46,50 @@ theorem evm_sdiv_sign_bit_block_spec_within
               (base + 4) hsign_ne_x0
   runBlock L S
 
+/-- CodeReq for one conditional-negation limb step at byte offset `base`. -/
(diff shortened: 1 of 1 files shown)


pirapira:
Follow-up scheduled: bead `evm-asm-ds00x` (P1) tracks hiding the let-bindings in the pre/post conditions via `@[irreducible]` definitions, per the AGENTS.md "Common Pitfalls" guidance. Will land as a separate PR on top of this branch (or after merge).

PR #3592, opened 2026-05-10
feat(sdiv): sequence dividend abs block

pirapira:
## Summary
- compose the 5-instruction SDIV sign prefix through the in-place dividend absolute-value block
- share the probed dividend top limb as the fourth negation limb and frame divisor sign state across the block
- expose the 26-instruction prefix postcondition with normalized dividend limbs and carry state

Depends on #3591.
Refs #90.
Bead: evm-asm-01uh.

## Verification
- lake build EvmAsm.Evm64.SDiv.Compose.Base
- scripts/check-file-size.sh
- rg -n "sorry|admit|set_option maxHeartbeats|set_option maxRecDepth" EvmAsm/Evm64/SDiv/Compose/Base.lean
- lake build

### DIFF
--- a/EvmAsm/EL/Conformance/All.lean
+++ b/EvmAsm/EL/Conformance/All.lean
@@ -44,11 +44,11 @@ def allConformanceVectorCount : Nat :=
   allConformanceVectors.length
 
 theorem allConformanceVectors_length :
-    allConformanceVectors.length = 46 := by
+    allConformanceVectors.length = 47 := by
   native_decide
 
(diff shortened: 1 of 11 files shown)


pirapira:
Follow-up scheduled: bead `evm-asm-qi036` (P1) tracks hiding the let-bindings in the pre/post conditions via `@[irreducible]` definitions, per the AGENTS.md "Common Pitfalls" guidance. Will land as a separate PR on top of this branch (or after merge).

PR #3593, opened 2026-05-10
feat(sdiv): sequence divisor abs block

pirapira:
## Summary
- compose the 26-instruction SDIV prefix through the in-place divisor absolute-value block
- frame the already-normalized dividend state while normalizing divisor limbs in place
- expose the 47-instruction prefix postcondition with both operands normalized

Depends on #3592.
Refs #90.
Bead: evm-asm-01uh.

## Verification
- lake build EvmAsm.Evm64.SDiv.Compose.Base
- scripts/check-file-size.sh
- rg -n "sorry|admit|set_option maxHeartbeats|set_option maxRecDepth" EvmAsm/Evm64/SDiv/Compose/Base.lean
- lake build


pirapira:
Follow-up scheduled: bead `evm-asm-n2vmv` (P1) tracks hiding the let-bindings in the pre/post conditions via `@[irreducible]` definitions, per the AGENTS.md "Common Pitfalls" guidance. Will land as a separate PR on top of this branch (or after merge).

pirapira:
The let bindings need to be hidden from the statements

PR #3885, opened 2026-05-13
refactor(exp): name full-loop cond product from limbs

pirapira:
Summary:
- add expCondMulRwFromLimbs for the full-loop conditional multiply product
- use it in exp_cond_mul_call_block_evm_exp_with_mul_spec_within
- remove the local r let from that theorem surface

Validation:
- lake build EvmAsm.Evm64.Exp.Compose.FullLoop
- git diff --check
- forbidden scan for sorry/admit/set_option in FullLoop.lean
- wc -l EvmAsm/Evm64/Exp/Compose/FullLoop.lean
- scripts/check-file-size.sh
- scripts/check-unimported.sh
- lake build


pirapira:
Diff is empty

PR #4657, opened 2026-05-18
fix(sdiv): preserve callable return address in div loop

pirapira:
## Summary
- rename the DIV loop counter away from x1 to x9 through DivMod programs and proofs
- add callable NoX1 scratch/pre/post surfaces so SDIV can call DIV without clobbering the return address
- thread no-x9 exact return/result-sign-fix postconditions through SDIV while keeping the public return PC normalized

## Verification
- lake build EvmAsm.Evm64.SDiv.Spec
- lake build

Bead: evm-asm-01uh.1

### DIFF
--- a/EvmAsm/Evm64/AddMod/Spec.lean
+++ b/EvmAsm/Evm64/AddMod/Spec.lean
@@ -36,7 +36,7 @@ open EvmAsm.Evm64.AddMod.Compose
 
 The bridge lemma connects the prologue postcondition
 `evmAddModPhase1Phase2LimbPost` (for b=N=0) to the mod callable
-precondition `divModStackDispatchPre`. This is the key step enabling
+    precondition `divModStackDispatchPreCallable`. This is the key step enabling
 the N=0 end-to-end proof (bead evm-asm-a32mz).
 
 Key simplification: when b=0 (the second ADDMOD operand = N = 0),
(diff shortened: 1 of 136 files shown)


pirapira:
Rebased onto origin/main at f52fd553e and force-pushed as 72505f69c. During the rebase, I kept the main-branch DivMod LimbSpec cleanup and reapplied the live x1 -> x9 loop-counter scratch-temp changes in the remaining DivMod LimbSpec/LoopBody proofs.

Verification:
- lake build EvmAsm.Evm64.SDiv
- git diff --check
- rg -n '\.x90|\.x91|\.x92' EvmAsm/Evm64/DivMod (no matches)
- lake build

Current PR state is OPEN / BLOCKED, no longer DIRTY.

pirapira:
Pushed amended head 2a5b22ac3 to address the CI file-size guardrail failure. The fix extracts the zero-divisor unified dispatcher wrappers into EvmAsm/Evm64/DivMod/Spec/UnifiedBzero.lean, bringing Unified.lean down to 1130 lines while preserving the existing theorem surface. Local verification passed: lake build EvmAsm.Evm64.DivMod.Spec.UnifiedBzero; lake build EvmAsm.Evm64.DivMod.Spec.Unified; lake build EvmAsm.Evm64.SDiv; scripts/check-file-size.sh; scripts/check-unimported.sh; git diff --check; full lake build. GitHub build is currently pending after the force-push.

pirapira:
Remote follow-up: head 2a5b22ac3 is now CLEAN. GitHub build and PR summary both passed after the UnifiedBzero split.

PR #4668, opened 2026-05-18
feat(exp): compose fixed reload-save prefix

pirapira:
## Summary
- stack on #4667
- add `exp_msb_bit_test_fixed_reload_then_save_expIterBodyFullMsbSavedBitTwoMulFixedCode_spec_within`
- compose the fixed x19 reload bit-test path with `exp_save_bit`, carrying the x16 pointer and reloaded limb memory atom

## Verification
- `lake build EvmAsm.Evm64.Exp.Compose.SavedBitBaseTwoMulFixedIter`
- `lake build EvmAsm.Evm64.Exp`
- `scripts/check-file-size.sh`
- `git diff --check`
- `lake build`

PR #4669, opened 2026-05-18
feat(exp): lift fixed prefixes over mul union

pirapira:
## Summary
- stack on #4668
- add union-with-`mul_callable` wrappers for the fixed skip-save and reload-save prefixes
- prepare the fixed prefix proofs for composition with squaring/conditional multiply call paths over the same code bundle

## Verification
- `lake build EvmAsm.Evm64.Exp.Compose.SavedBitBaseTwoMulFixedIter`
- `lake build EvmAsm.Evm64.Exp`
- `git diff --check`
- `lake build`


pirapira:
Remote update: head ee3673c8 is CLEAN. GitHub build and PR summary both passed.

PR #4998, opened 2026-05-19
feat:  progress tracker

alexanderlhicks:
Adds a progress framework anchored against the L1-zkEVM "guest program" goal (eth-act/zkvm-standards, ere, zkevm-benchmark-workload):

- EvmAsm/Progress.lean: tracks progress on opcodes
- PROGRESS.md: documents this progress
- scripts/check-progress.sh: fails the CI build if PROGRESS.md drifts from the registry.
- scripts/progress-delta.sh + docs/pr-summary-progress-prompt.md: feed a deterministic base/head delta intothe summary workflow via additional_instructions_path (I'll replace the old "style guide path" with this in the upstream summe workflow repo) so the PR summary covers the computed delta.

Some decisions that can be discussed:
  1. **9-item guest-program checklist is hand-written prose** in
     `scripts/progress-template.md`. Encoding it as Lean defs in
     `EvmAsm/Progress.lean` was considered and deferred; worth revisiting
     if drift appears.
  2. **PUSH2..32 = one `.partly` registry entry** (not 31). The 31
     byte-codes share `evm_push_zero_slot_full_stack_spec_within` and
     the gap is uniform; split if/when the non-zero-slot path lands.
  3. **MLOAD/MSTORE stay in `.proven`** for the aligned spec with notes
     about the unaligned `_public` gap. Not split into two entries;
     revisit when the unaligned wrappers complete.
  4. **PROGRESS.md totals reported per registry entry AND per opcode byte**
     (143 EVM bytes, parameterized families expanded). Both numbers are
     shown so PUSH/DUP/SWAP/LOG aren't undercounted.


pirapira:
[edit: done] The merge conflict is from my README edit. fixing.

pirapira (dismissed):
Looks good to me. I prefer seeing this live rather than picking nits like the number of instructions are  149/143 etc here.

PR #5027, opened 2026-05-19
ci: stop augmenting PR summary with progress delta

pirapira:
## Summary
- stop building a temporary augmented instructions file in the PR summary workflow
- pass `CONTRIBUTING.md` directly to the summary action again
- keep the existing helper script/doc in this PR so the current `pull_request_target` workflow can still run while this workflow change is under review

## Verification
- `git diff --check`
- `git diff --stat origin/main..HEAD`


pirapira:
???

PR #5097, opened 2026-05-20
feat(div): wire evm_div through codegen + ziskemu

pirapira:
## Summary
- Register `evm_div` (M2, hardcoded operands) and `evm_div_from_input` (M4, ziskemu `-i` input file) in `EvmAsm/Codegen/Programs.lean`, mirroring the ADD pattern.
- Lay out 256 B of `div_scratch:` *before* `operands:` / `operands_ram:` in `.data` so DIV's negative-offset stores (unsigned 12-bit `3936..4088 ≡ -160..-8`) land in writable RAM.
- `scripts/codegen-evm_div-check.sh`: M2 case 2^64 / 2 = 2^63.
- `scripts/codegen-evm_div-cases-check.sh`: M4 builds the ELF once and runs 14 corner cases (single-limb / n=2 / n=3 / n=4 cascades, b=0, a<b, a==b, max-value boundaries, CLZ=0 shift=0 fast path, shift>0 denorm, the two v1 div128 counterexamples from `memory/project_sdiv_div128_bug.md`, dense 4-limb).

### Fixes uncovered while running the suite locally
- **Codegen branch/JAL emission** (`Emit.lean`): bare integer offsets (`beq x5, x0, 1020`) are interpreted by GNU `as` as ABSOLUTE addresses, breaking link with `R_RISCV_JAL truncated to fit`. New `emitBranchOff` helper renders `.+N` / `.-N` PC-relative form. ADD never tripped this (no branches). RoundTripTests pinned to the new spelling.
- **DIV "exit PC" fall-through** (`Programs.lean`): `evm_div`'s exit is a NOP at index 267 immediately followed by the 75-instruction `divK_div128_v4` subroutine, so the main path's `JAL +24` to the NOP fell through into the subroutine — its `jr sp` then returned to garbage. `evmDivPatched` splices the body to replace the NOP with `JAL .x0 +304`, skipping the subroutine and landing at `evmAddEpilogue`. Verified callers of the subroutine still use the original `jal x2, +560` — only the NOP slot was replaced.

## Test plan
- [x] `lake build codegen` — green; `codegen --help` lists `evm_div` and `evm_div_from_input`.
- [x] `bash scripts/codegen-evm_div-check.sh` — `PASS: evm_div output matches expected 2^63 quotient`.
- [x] `bash scripts/codegen-evm_div-cases-check.sh` — `PASS: all 14 cases matched expected quotients`. Both v1 counterexamples produce the correct quotients under v4.
- [x] `bash scripts/codegen-evm_add-check.sh` — still green (no regression).


pirapira:
@dhsorens just a ping. This runs DIV through ziskemu. Merging because other tests depend on relative-jump fix in this PR.

PR #5472, opened 2026-05-21
feat(codegen): M9 — wire SDIV and SMOD via trampoline wrapper

dhsorens:
## Summary

Closes the M8 scope-reduction debt. **SDIV (0x05) and SMOD (0x07)** now route through `tinyInterpRegistry`. Their verified bodies end with a saved-ra-ret (`JALR x0, x18, 0`) that bypasses the dispatcher's standard `mv x10, x14; addi; ret` tail; M9 introduces a **trampoline pattern** via a new `OpcodeHandlerSpec.postBodyLabel : Option String` field — generalisable to any future opcode with the same shape.

- `EvmAsm/Codegen/Dispatch.lean` — extended `OpcodeHandlerSpec` with `postBodyLabel`; `emitSubroutine` emits the optional label between body and tail. Handlers that leave it `none` (all pre-M9) emit byte-identical asm.
- `EvmAsm/Codegen/Programs.lean`:
  - `evmSdivPatched := (evm_sdiv : List Instr).drop 1` and `evmSmodPatched := (evm_smod : List Instr).drop 1` — strip the leading `ADDI .x18 .x1 0` save_ra_block so our trampoline target sticks in x18.
  - `signedDivModTail` = `mv x10, x14; addi x10, x10, 1; j .dispatch_loop` — **NOT** `ret`: the wrapper's inner `JAL .x1` into the callable divider clobbers x1 mid-body, so `ret` would jump to garbage.
  - `signedDivModHandlers` — two entries (`h_SDIV`, `h_SMOD`). Each: `preBody := "  mv x14, x10\n  la x18, h_<NAME>_done"`, `postBodyLabel := some "h_<NAME>_done"`, `tail := signedDivModTail`.
- `EvmAsm/Codegen/Tests/Cases.lean` — 3 new cases:
  - `sdiv_basic` (5 / 2 = 2, positive path)
  - `sdiv_negative` (`NOT` trick yields −2; SDIV(−2, 2) = −1)
  - `smod_negative` (SMOD(−2, 5) = −2, sign(result) = sign(a) per EVM SMOD)
- `scripts/codegen-opcodes-runtime-check.sh` — per-case step budget bumped from 200K → 500K (Knuth-D inner loop + trampoline overhead pushes beyond 200K even for small operands).

## Trampoline mechanics gotcha logged

First attempt used the M8 `divModTail` (= `mv x10, x14; addi; ret`). ziskemu hung with `EmulationNoCompleted`. **Cause:** the wrapper's inner `JAL .x1` into `evm_div_callable_v4` clobbers `x1`, so by the time control reaches the restore stub `x1` no longer points at the dispatcher's continuation. **Fix:** swap `ret` for `j .dispatch_loop`. Logged in CODEGEN.md.

## Test plan

- [ ] `lake build codegen` passes (the `postBodyLabel` field default + patched bodies typecheck)
- [ ] `bash scripts/codegen-opcodes-runtime-check.sh` exits 0 with all **29 cases PASS** (~30 s wall-clock; the SDIV/SMOD inner loops add some time but still well under the 60 s threshold)
- [ ] `bash scripts/codegen-opcodes-check.sh` (legacy fallback) still passes the same 29 cases
- [ ] Pre-existing scripts unchanged: `codegen-smoke.sh`, `codegen-tiny-interp{,-dispatch}-check.sh`, `codegen-evm_add{,-from-input}-check.sh`, `codegen-evm_div{,-cases}-check.sh`, `codegen-evm_mod{,-cases}-check.sh`
- [ ] `bash scripts/check-progress.sh` exits 0 (PROGRESS.md drift is snapshot-line-only, ignored)
- [ ] Disassembly spot-check on `gen-out/runtime_dispatcher.s` — `h_SDIV:` starts with `mv x14, x10` then `la x18, h_SDIV_done`, body's `jalr x0, 0(x18)` jumps forward to `h_SDIV_done:`, which falls through to `mv x10, x14; addi x10, x10, 1; j .dispatch_loop`. Body splice means the spliced-off `ADDI x18, x1, 0` is **not** present at the start of the body.

## SMOD v4 follow-up

`evm_smod` still uses the legacy non-v4 `evm_mod_callable`. SDIV migrated to v4 in commit `43bb53070`; SMOD's surface hasn't flipped yet. When the verification track migrates SMOD, this PR's `evmSmodPatched` definition rebinds to the new symbol with no other changes.

## What this unblocks

- **M10 — ADDMOD / EXP via `callableLabel?`** can now extend `OpcodeHandlerSpec` with a parallel field for opcodes that are **near-called** from other handler bodies (ADDMOD calls MOD; EXP calls MUL).
- **M11 — JUMP / JUMPI / JUMPDEST / PC / GAS** is the biggest remaining unblock toward running real EVM programs.
- **M12+ — Stateless guest bridge** still on the long arc.

### DIFF
--- a/CODEGEN.md
+++ b/CODEGEN.md
@@ -38,7 +38,7 @@ untouched. Generated artifacts go in `gen-out/` (gitignored).
 | `EvmAsm/Codegen.lean` | Top-level umbrella (mirrors `EvmAsm/Rv64.lean`, `EvmAsm/Evm64.lean`). |
 | `EvmAsm/Codegen/Emit.lean` | Pure `emitReg`, `emitInstr`, `emitProgram` — `Instr → String`. No `IO`. |
 | `EvmAsm/Codegen/Layout.lean` | `HaltConv` enum, halt stubs, `_start` preamble, `.option norvc`, `MEM_START`/`MEM_END` constants, `BuildUnit` struct + `emitBuildUnit`/`emitDataLabel` helpers. |
-| `EvmAsm/Codegen/Dispatch.lean` | M5b dispatcher scaffolding: `OpcodeHandlerSpec` (with optional `preBody` for x10-clobbering handlers) + `HandlerTail` types, `emitDispatcherPrologue`/`Epilogue`/`DataSection` and `buildDispatchUnit` helpers. M8.5 adds the parallel runtime-bytecode helpers (`emitRuntimeDispatcherPrologue` / `emitRuntimeDispatcherDataSection` / `buildRuntimeDispatchUnit`) that read bytecode from `INPUT_ADDR + INPUT_DATA_OFFSET` at runtime. Pure (no IO). |
+| `EvmAsm/Codegen/Dispatch.lean` | M5b dispatcher scaffolding: `OpcodeHandlerSpec` (optional `preBody` for x10-clobbering handlers + optional `postBodyLabel` for M9's trampoline pattern) + `HandlerTail` types, `emitDispatcherPrologue`/`Epilogue`/`DataSection` and `buildDispatchUnit` helpers. M8.5 adds the parallel runtime-bytecode helpers (`emitRuntimeDispatcherPrologue` / `emitRuntimeDispatcherDataSection` / `buildRuntimeDispatchUnit`) that read bytecode from `INPUT_ADDR + INPUT_DATA_OFFSET` at runtime. Pure (no IO). |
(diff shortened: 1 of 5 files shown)


pirapira:
I noticed this still uses known-to-be buggy smod. I'll update it when a new smod lands.

dhsorens:
good catch @pirapira !

pirapira:
I have a Claude Code in a loop that merges things with `merge!` label. It deals with merge conflicts and PROGRESS.md.

PR #5541, opened 2026-05-21
feat(stateless): u256_lt (PR-K50)

pirapira:
## Summary

Strict less-than comparison on two 32-byte big-endian `u256` buffers in place. Returns `1` if `a < b`, `0` otherwise (including `a == b`).

BE storage convention: byte 0 is the most-significant byte; byte 31 is the least. Mirrors the layout produced by `rlp_field_to_u256_be` and consumed throughout the K-stack.

Pure register arithmetic, no scratch memory, leaf-callable. Walks at most 32 bytes; short-circuits on the first differing byte.

Building block for:
- Tx validation (`balance < gas_cost + value` → reject)
- EIP-1559 base-fee comparisons (`tx.max_fee_per_gas < block.base_fee_per_gas` → reject)
- Signature value range checks (`r < secp256k1_N`, `s < secp256k1_N/2`)
- Any u256 ordering predicate

## Test plan

- [x] `lake build` clean
- [x] `scripts/codegen-zisk-u256-lt-check.sh` passes 20 fixtures:
  - Equalities (zero, one, max, mid-magnitude → all 0)
  - Strict lessness across magnitudes
  - Strict greaterness (returns 0)
  - MSB-only / LSB-only / mid-byte differences
  - The `2^128` boundary
  - Realistic balance vs gas-cost comparisons

### DIFF
--- a/EvmAsm/Codegen/Programs.lean
+++ b/EvmAsm/Codegen/Programs.lean
@@ -5366,6 +5366,73 @@ def ziskValidateHeaderBasicProbeUnit : BuildUnit := {
   dataAsm     := ziskValidateHeaderBasicDataSection
 }
 
+/-! ## u256_lt -- PR-K50 strict less-than on big-endian u256 buffers
+
+    Compare two 32-byte big-endian u256 buffers in place. Returns
+    a 0/1 result for `a < b`. Building block for tx validation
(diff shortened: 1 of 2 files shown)


pirapira:
No longer relevant.

PR #5570, opened 2026-05-21
feat: chaos monkey

kevaundray:


pirapira (changes requested):
Please add a test that falls before or after the change.

pirapira:
Alternatively, is there a Lean library for RLP that we can use?

pirapira:
Lean specifications in this repository are hallucinated fantasy based on the Python execution spec. Testing these against the execution spec is future work (#5589 for RLP).

I'm closing this PR in order to avoid confusing LLMs.

PR #5713, opened 2026-05-22
feat(exp): full-loop composition skeleton with isolated body sorry

pirapira:
## Summary

Rescue PR — branch has 3 commits not on main and never had a PR opened.

- Adds `EvmAsm/Evm64/Exp/Compose/SavedBitFullLoopBody.lean` (+183 lines) — the full-loop composition skeleton for EXP.
- Final commit isolates the remaining open goal into a single named lemma `expTwoMulLoopBodyCorrect` so the surrounding scaffolding is closed and the residual proof obligation is well-localized.
- Tracked by bead `evm-asm-w5mk`.

## Test plan

- [ ] `lake build EvmAsm.Evm64.Exp`
- [ ] Confirm the only `sorry` in the new file is `expTwoMulLoopBodyCorrect` (intentional staging marker per the bead).
- [ ] No regressions in downstream `Exp/Spec.lean`.

### DIFF
--- a/EvmAsm/Evm64/Exp.lean
+++ b/EvmAsm/Evm64/Exp.lean
@@ -64,6 +64,7 @@ import EvmAsm.Evm64.Exp.Compose.SavedBitBoundarySeq
 import EvmAsm.Evm64.Exp.Compose.SavedBitLoopEntry
 import EvmAsm.Evm64.Exp.Compose.SavedBitLoopExit
 import EvmAsm.Evm64.Exp.Compose.SavedBitBoundaryLoop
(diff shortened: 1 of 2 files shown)


pirapira:
This has a sorry.

pirapira:
Closing this rescue PR instead of continuing it. It imports a new EXP module containing an intentional `sorry`, so it cannot satisfy the current no-sorry merge rules. The underlying work is already tracked by open bead evm-asm-w5mk and child evm-asm-w5mk.3, whose notes point at the newer generalized-induction/boundary architecture rather than this stale main-based skeleton.

PR #6910, opened 2026-05-27
codegen(m16): wire KECCAK256 via ECALL bridge (first precompile pattern)

dhsorens:
## Summary

First opcode that uses a host syscall / accelerator bridge instead of an inline verified RV64 body. Wires **KECCAK256 (0x20)** through a new \`hashHandlers\` builder. Lifts wired opcode count **111 → 112** (75% of the 149-byte EVM space). Establishes the **ECALL bridge pattern** that LOG0-4, SLOAD, SSTORE, and the precompile family (ECRECOVER, SHA256, MODEXP, …) will reuse.

Reuses the existing raw-asm subroutine \`EvmAsm/Codegen/Programs/HashBridge.lean::zkvmKeccak256Function\` — already used 30+ times by Stateless-guest BuildUnits. The dispatcher now emits this subroutine + supporting \`zk3_state\` block; the M16 handler does the EVM stack-to-regs glue.

## Dispatcher changes

| File / location | Change |
|---|---|
| \`Dispatch.lean\` epilogue | Inserts \`zkvm_keccak256\` subroutine **between** handler subroutines and \`h_invalid:\` so it's reachable only via \`jal x1, …\` (not by fall-through from exitBody). |
| \`Dispatch.lean\` data section | Adds 200 B \`zk3_state:\` + 512 B \`lp64_stack:\` / \`lp64_sp_top:\`. |
| \`Dispatch.lean\` prologue (both \`.data\`-baked and runtime variants) | First instruction is now \`la sp, lp64_sp_top\` — keccak does \`addi sp, sp, -32\` to save callee-saved regs, and the dispatcher had never initialised \`sp\`. |

## Handler shape

Shape (a) per the M16 plan: \`body := []\`, \`tail := .custom "..."\`. The tail:
1. Saves dispatcher state: \`mv s10, x10\` (EVM code ptr).
2. Pops offset/size from EVM stack into \`t0\` / \`a1\` (LP64 args). \`t0\` is used as offset because \`a3\` (= x13) is the evm_memory base register.
3. Net stack delta = +32 (pop 2 × 32, push 1 × 32).
4. \`mv a2, x12\` (output ptr = result slot), \`mv s11, x12\` (save EVM stack ptr).
5. \`jal x1, zkvm_keccak256\` (clobbers x1, a0, a1, a2).
6. Restore \`x10\` / \`x12\` from s10 / s11.
7. \`addi x10, x10, 1\` + \`j .dispatch_loop\` (NOT \`ret\` — the inner \`jal\` clobbered x1).

## Test case

| Case | Bytecode | Expected |
|---|---|---|
| \`keccak256_empty\` | \`60 00 60 00 20 00\` (PUSH1 0; PUSH1 0; KECCAK256; STOP) | \`c5d246…5a470\` (standard keccak256 empty hash) |

Endianness check: the subroutine writes the digest as raw bytes to the output ptr; the dispatcher's epilogue copies \`x12+0..x12+31\` verbatim to OUTPUT_ADDR. Standard digest hex round-trips.

## Registry invariants

Counts bumped 111 → 112. Phase 1 \`decide\` recompile unchanged at ~13 s.

## Verification

All locally:

- [x] \`lake build EvmAsm.Codegen\` — 1300/1300 jobs ok.
- [x] \`scripts/codegen-opcodes-runtime-check.sh\` — **45/45 PASS**, including \`keccak256_empty\`.
- [x] \`scripts/check-progress.sh\` — drift gate passes.

## ECALL bridge pattern (template for M17+)

1. Promote a \`zkvm_*Function : String\` into the dispatcher epilogue.
2. Allocate any scratch buffer in the data section.
3. Handler tail saves \`x10\`/\`x12\` → s10/s11, sets up LP64 args, \`jal x1, <subroutine>\`, restores state, advances PC, returns via \`j .dispatch_loop\`.

## Risks / open questions found during implementation

- **`sp` was uninitialised**. Caught when the first keccak test triggered \`Mem::write_silent() invalid addr=ffffffffffffffe0\` in ziskemu — the subroutine's first \`addi sp, sp, -32\` wrote to address \`0 - 32 = -32\`. Fixed by adding \`la sp, lp64_sp_top\` to both prologues.
- **Subroutine placement**. Initially placed the keccak function at the end of the epilogue (after exitBody). exitBody falls through into whatever follows, so the keccak function was reached unintentionally → infinite loop. Fixed by moving the subroutine between handler subroutines and \`h_invalid:\`.

## Test plan

- [ ] CI build green (Lean + drift gate + opcode regression suite).
- [ ] Reviewer confirms \`zkvm_keccak256\` subroutine placement is reachable only via \`jal\`.
- [ ] Reviewer confirms the LP64 register conflict navigation (\`s10\`/\`s11\` saves around the call).


pirapira:
Looks good to me, merging.

Issue #7130, opened 2026-05-28
Decide on memory layout of storage


pirapira:
Since the EVM storage can be used on randomly looking addresses, it's not obvious how to represent the storage in the riscv memory. This issue keeps track of the first version. 

Should be roughly compatible with the gas model.

pirapira:
> 🤖 **Generated by Claude Code** (model: Opus 4.8). This is an automated survey + design-options writeup to seed discussion on #7130. Numbers and external-implementation claims are best-effort from public sources (linked at the bottom); please sanity-check before treating any of it as settled.

# Storage representation: survey of EVM implementations + design options for #7130

The hard part of this issue is that EVM storage is a **sparse map over a 2²⁵⁶ keyspace with adversarially-chosen (random-looking) keys**, and we have to land it in a *flat RISC-V linear address space* in a way that is (a) **provable in Lean**, (b) cheap in **zkVM cycles**, and (c) **compatible with the gas model** (EIP-2929 cold/warm and EIP-2200/3529 original-vs-current SSTORE accounting). Those three constraints pull in different directions, and they are *not* the constraints that drive the design of native clients — which is exactly why we shouldn't just copy what revm/geth do.

## What the design actually has to support

For each `(address, slot)` the storage layer must answer, given our existing infra (`EvmAsm/EL/Storage.lean`, `Stateless/State/Storage.lean`, `Witness/MPT/*`, `StorageGas.lean`, `StorageAccess.lean`):

1. **SLOAD** — return current value; 0 if never written. Cold-vs-warm tells gas (2100 vs 100).
2. **SSTORE** — write current value; gas (EIP-2200/3529) needs **three** values per slot: `original` (committed at tx start), `current`, and `new`. So tracking only one value is insufficient.
3. **Cold-miss path** — first touch in a block must read the *authenticated* value via `MPT.Get(storageRoot, keccak256(slot))` over the witness `NodeDb`, and fail with `WITNESS_MISSING_NODE (0x06)` if absent.
4. **Revert** — `CALL`/`CREATE`/`REVERT`/exception roll back storage to a checkpoint (journaling).
5. **Commit / state-root recompute** — at block end, every *dirty* slot must be **enumerable** so it can be fed into `MPT.Set` and checked against the header root (`STATE_ROOT_MISMATCH 0x07`).

[...]

PR #7133, opened 2026-05-28
codegen(m22): real SLOAD/SSTORE via pre-loaded slot table

dhsorens:
## Summary

Second step on the EEST path after M21 calldata. **SLOAD (0x54)** and **SSTORE (0x55)** graduate from M17 no-ops to real persistent storage. State arrives via a third length-prefixed segment in the `ziskemu -i` input file (alongside M21 bytecode + calldata). TLOAD/TSTORE remain M17 no-ops (transient storage deferred to a later PR).

**Architecture (pre-loaded slot table, not in-dispatcher hash table)**: dispatcher prologue copies the input's storage segment into a writable `.data` region (`evm_slot_table`, 16 KiB = 256 slots × 64 B) and records the count at `env.slotTableCountOff = 448`. SLOAD / SSTORE inline-asm bodies scan the table linearly. Keeps the dispatcher stateless across invocations and matches a witness-style flow without committing to a specific witness format.

### Input format (post-M22)

```
<u64 bytecode-len><bytecode><pad>
<u64 calldata-len><calldata><pad>
<u64 slot_count><slot_count × 64B (key, value)><pad>   ← NEW (M22)
```

Empty default (`--storage ""`) writes `slot_count = 0` so the 61 pre-M22 cases pack identically.

### Changes

- `scripts/pack-bytecode.py`: new `--storage` flag accepting parenthesized hex pairs `"(0xKEY, 0xVAL) (0xKEY2, 0xVAL2)"`. Each key/value serialized in EVM-stack byte order (LE limb order, matching what PUSH32 + SSTORE deposit on the stack).
- `EvmAsm/Codegen/Dispatch.lean`: prologue plumbing (compute slot-segment start past calldata pad, read count, dword-loop-copy bytes into `evm_slot_table`, write count to env+448). Both `.data` sections declare `evm_slot_table: .zero 0x4000` (16 KiB; total `.data` footprint goes ~36 KiB → ~52 KiB, well under the ~64 KiB cap before `OUTPUT_ADDR`).
- `EvmAsm/Evm64/Environment/Layout.lean`: adds `slotTableCountOff = 448`, bumps `envSize` 448 → 456, extends `envSize_covers` chain. `Assertion.lean` `envCells` 56 → 57.
- `EvmAsm/Codegen/Programs/Storage.lean` (**NEW** submodule): extracted from `Programs/Evm.lean` (the M22 inline-asm scan loops pushed it past the per-file size cap, mirrors the M18 `Noop.lean` extraction). Real SLOAD / SSTORE bodies via GNU AS numeric local labels (unique-per-use, no inter-handler collisions). TLOAD/TSTORE carry over as M17 no-ops.
- `EvmAsm/Codegen/Tests/Cases.lean` + `Cli.lean` + bash runner: `OpcodeTestCase` gains optional `storage` field; `--list-test-cases` emits 5-column TSV; runner uses `cut -f` per-field instead of `IFS=$'\t' read -r` (POSIX `read` collapses adjacent tabs because tab is IFS-whitespace, silently shifting columns when calldata is empty — `cut` preserves empty fields).

### Test cases

- **`sstore_then_sload_round_trip`**: store, read back — `PUSH1 0x42; PUSH1 0x00; SSTORE; PUSH1 0x00; SLOAD; STOP`. Empty preload. Exercises the append-on-write + read-back path.
- **`sload_preloaded_match`**: read a preloaded slot. Confirms the prologue copy + linear-scan match.
- **`sload_preloaded_no_match`**: read an absent key → 0. Confirms default-zero semantics.
- **`sstore_overwrites_preload`**: overwrite a preloaded slot in place — `PUSH2 0xbeef; PUSH1 0x00; SSTORE; PUSH1 0x00; SLOAD; STOP` with preload `(0x00, 0xdead)`. Confirms in-place mutation (not append).

The pre-M22 `sstore_sload_roundtrip` test (M17-era; asserted no-op behavior) was removed — same bytecode as the new `sstore_then_sload_round_trip` and its expected output of `0x00…` is now wrong under M22 real semantics.

**Cases: 62 → 65** (-1 redundant + 4 new).

### Known limitations (documented in `Programs/Storage.lean`)

- **Capacity cap**: 256 slots (limited by the 16 KiB `.data` block; growing beyond ~12 KiB headroom would overrun `OUTPUT_ADDR`).
- **Linear scan**: O(slot_count) per access. Fine for tests (<10 slots typical); a bottleneck for proving real blocks. Hash upgrade ships without ABI/env changes.
- **TLOAD/TSTORE remain no-ops**.
- **No gas accounting** for SLOAD/SSTORE.
- **Modified slots not surfaced at `OUTPUT_ADDR+32..`**: M24's job (post-state serializer) in the original roadmap.
- **Inline asm, not verified bodies**: scan loops would need a verified-loop pattern; M22 ships correct asm with comprehensive tests, verified bodies follow in a separate PR.

## Test plan

- [x] `scripts/codegen-opcodes-runtime-check.sh` exits 0 with **65/65 PASS**
- [x] Spot-checked output of `pack-bytecode.py --storage "(0x42, 0xdead)" <bytecode>` via `xxd` — segment layout correct (LE u64 prefix + 32B key + 32B value in EVM-stack byte order)


pirapira:
Looks good to me. Merging. The memory layout and the data structure of storage will change. #7130

PR #7139, opened 2026-05-28
codegen(m23): real RETURN/REVERT with returndata buffer + halt-kind

dhsorens:
## Summary

**Storage-orthogonal pivot** after M22 (#7133) merged. Issue
**#7130 ("Decide on memory layout of storage")** flagged that
M22's slot-table layout is the **first version** of storage and
will be redesigned soon to handle arbitrary 256-bit addresses +
align with the EVM gas model. The original M23 (pre-state
witness unpacking) and M24 (post-state serializer) both bake in
M22's specific layout — picking either as the next PR would
risk burning work that gets thrown away when #7130 lands.

M23 therefore picks **real RETURN (0xf3) and REVERT (0xfd)**.
These graduate from M18 halt no-ops to real bodies that surface
the returndata at `OUTPUT_ADDR` and tag the halt kind so EEST
tests can distinguish successful return from revert. **Zero
coupling to M22 storage** — survives the #7130 redesign untouched.

## OUTPUT layout extension

```
OUTPUT_ADDR (0xa0010000):
  +0..32    <result>             stack top for STOP / returndata for RETURN/REVERT
  +32..40   <u64 LE halt_kind>   0=STOP/unspecified, 1=RETURN, 2=REVERT
                                 (INVALID/SELFDESTRUCT inherit 0 — deferred)
```

## The new `.exit_no_epilogue` exit path

`evmAddEpilogue` copies the EVM stack top to `OUTPUT_ADDR[0..32]`. For STOP that's correct; for RETURN/REVERT it would clobber the returndata. M23 adds a new label `.exit_no_epilogue:` between `emitProgram exitBody` and the halt stub. RETURN/REVERT bodies write their OUTPUT and `j .exit_no_epilogue` to skip the epilogue; STOP, INVALID, SELFDESTRUCT continue through `.exit_label → evmAddEpilogue → halt stub`.

## Changes

- `EvmAsm/Codegen/Dispatch.lean`: `emitDispatcherEpilogue` emits the new label.
- `EvmAsm/Codegen/Programs/Evm.lean`: `evmAddEpilogue` gains `SD .x5 .x0 32` (writes `halt_kind=0`). Stays pure verified `Program`.
- `EvmAsm/Codegen/Programs/Noop.lean::haltHandlers`: RETURN/REVERT tails replaced with ~25 instructions each of inline asm (load offset/size, zero-fill OUTPUT[0..32], clamp size to 32, byte-loop copy from evm_memory, write halt_kind, j .exit_no_epilogue). GNU AS numeric local labels (1, 2, 3) reused across both bodies without collision.
- `EvmAsm/Codegen/Tests/Cases.lean` + `Cli.lean` + bash runner: `OpcodeTestCase` gains optional `expectedHaltKind` field; `--list-test-cases` emits 6-column TSV; runner asserts `OUTPUT[32..40]` via `xxd -s 32 -l 8` when non-empty.

## Test cases

- **`return_word_basic`**: `PUSH1 0x42; PUSH1 0x00; MSTORE; PUSH1 0x20; PUSH1 0x00; RETURN`. Expected: `0000…0042` + halt_kind `0100000000000000`.
- **`return_small_pads_zeros`**: `PUSH1 0xff; PUSH1 0x00; MSTORE8; PUSH1 0x01; PUSH1 0x00; RETURN`. Exercises `size < 32` zero-fill. Expected: `ff00…00` + halt_kind = 1.
- **`revert_word_basic`**: same bytecode as `return_word_basic` but byte 0xfd. Same returndata; halt_kind `0200000000000000`. **Proves RETURN and REVERT are observably distinguishable.**
- `return_pop2_halt` (pre-existing) updated: previously asserted the M18 no-op surfacing (top-of-stack 0xff via `evmAddEpilogue`); under M23 real RETURN reads uninitialized `memory[0x22..]` = zeros, so expected flips to 32 zeros + halt_kind = 1.

**Cases: 65 → 68** (3 new + 1 updated).

## Known limitations (documented in `haltHandlers` + `CODEGEN.md` M23)

- **Returndata clamped to 32 bytes.** Larger payloads silently truncate. Future PR can extend the OUTPUT layout.
- **INVALID/SELFDESTRUCT inherit `halt_kind = 0`.** Distinct tagging (3 / 4) is a small follow-up.
- **`offset`/`size > u64`** silently use only the low u64. Safe for `evm_memory`'s 32 KiB region.
- **No gas accounting** for RETURN/REVERT.
- **`RETURNDATASIZE` / `RETURNDATACOPY`** still no-ops — they read the caller's return-data buffer, meaningful only with nested call frames.
- **Inline asm, not verified body** for the copy loop. Verified-loop body is a follow-up.

## Test plan

- [x] `scripts/codegen-opcodes-runtime-check.sh` exits 0 with **68/68 PASS** (3 new + 1 updated + 64 unchanged); the 3 new + 1 updated cases assert halt_kind via the new `xxd -s 32 -l 8` path.
- [x] Manually verified `OUTPUT[32..40]` holds halt_kind = 1 for `return_word_basic.output` and = 2 for `revert_word_basic.output`


pirapira:
Looks good to me. Merging.

PR #7235, opened 2026-05-29
codegen(m24): storage on Option A — append-log + journal + real TLOAD/TSTORE

dhsorens:
## Summary

The storage memory-layout decision (issue **#7130**) landed with consensus to **start with Option A** — an append-only `(addrHash, slotKey, original, current)` log in `STATE_TRACKER_AREA = 0xa0630000`, with revert = log-length truncation (the log **is** the journal). Possible upgrade to **C** (direct-mapped hint index over the same log) later.

M24 brings M22's transitional storage into alignment with Option A **and** extends the same architecture to **real TLOAD/TSTORE** (graduated from M17 no-ops). Also adds **REVERT rollback** — SSTORE-then-REVERT now correctly truncates the log to a checkpoint captured at the dispatcher prologue's end.

## What M22 had that M24 changes

| Concern | M22 | M24 |
|---|---|---|
| Entry shape | `(slotKey:32, value:32)` = 64 B | `(addrHash:32, slotKey:32, original:32, current:32)` = 128 B |
| Semantics | In-place update on key match | Always append; scan from end (last-write-wins) |
| Location | Dispatcher `.data` `evm_slot_table` (16 KiB) | `STATE_TRACKER_AREA = 0xa0630000` (4 MiB) |
| Revert | None — SSTORE-then-REVERT committed | Log-length truncation to checkpoint |
| Transient (TLOAD/TSTORE) | M17 no-ops | Same Option A structure at `0xa0830000` |
| Original-value tracking | Absent | Captured on first-touch; preserved across SSTOREs |

## OUTPUT layout post-M24

```
OUTPUT_ADDR (0xa0010000):
  +0..32    <result>                       (M21/M23)
  +32..40   <u64 LE halt_kind>             (M23)
  +40..48   <u64 LE persistentLogLength>   (NEW M24)
  +48..56   <u64 LE transientLogLength>    (NEW M24)
  +56..256  <unused>
```

Surfaced by 6 instructions appended to `.exit_no_epilogue` — runs for **every** halt path.

## Changes

- `EvmAsm/Evm64/Environment/Layout.lean`: rename `slotTableCountOff` → `persistentLogLengthOff` (offset 448; semantic shift count-of-64B-entries → count-of-128B-entries). New `persistentLogCheckpointOff = 456` + `transientLogLengthOff = 464`. `envSize` 456 → 472.
- `EvmAsm/Evm64/Environment/Assertion.lean`: `envCells` 57 → 59.
[...]

PR #7241, opened 2026-05-29
codegen(m25): post-state slot serializer — modified slots at OUTPUT+56

dhsorens:
## Summary

M24 (PR #7235, merged) surfaced storage log **lengths** at `OUTPUT[40..56]` but not the actual slot data. M25 closes that gap: the dispatcher epilogue walks the persistent log from end (last-write-wins), dedups against already-emitted `slotKey`s, and writes up to 3 `(slotKey, current)` pairs to `OUTPUT[64..256]` with a count cell at `OUTPUT[56..64]`.

After M25 an EEST-style post-state assertion ("after this bytecode + preload, slot K has value V") is directly expressible via the new `expectedPostStorage` test field — **the load-bearing piece for running real EEST fixtures.**

The OUTPUT contract is **layout-agnostic**: dedup+emit produces a flat modified-slot list regardless of which storage architecture sits behind it. If storage later migrates from Option A to D+flat overlay, the inner walk changes but the OUTPUT layout stays stable.

## OUTPUT layout (post-M25)

```
OUTPUT_ADDR (0xa0010000):
  +0..32    <result>                              (M21/M23)
  +32..40   <u64 LE halt_kind>                    (M23)
  +40..48   <u64 LE persistentLogLength>          (M24 — raw count, with duplicates)
  +48..56   <u64 LE transientLogLength>           (M24)
  +56..64   <u64 LE numModifiedPersistentSlots>   (NEW M25 — deduped, ≤ 3)
  +64..(64 + N*64)
            <modified slots: (slotKey:32, current:32) × N>  (NEW M25)
  +...      <zero-padded to 256 B by ziskemu init>
```

3-slot cap follows from `256 - 64 = 192 = 3 × 64`. Sufficient for pure-computation EEST tests; larger workloads need a future PR.

## Why dedup (and why reverse-write order)

Under Option A, SSTORE always appends, so the log can hold multiple entries for the same `slotKey`. The dedup walks the log from end (newest first), checks each `slotKey` against the already-emitted output list, and emits only on first sight. Result: slots appear in **reverse write order** (most-recently-modified first). Test authors construct expected strings accordingly.

## Changes

- `EvmAsm/Codegen/Dispatch.lean`: `emitDispatcherEpilogue` gains ~60 instructions of dedup-and-emit asm after the M24 log-length writes, before the halt stub. Numeric local labels `1:`–`6:` reused within the block (unique-per-use across the file).
- `EvmAsm/Codegen/Tests/Cases.lean` + `Cli.lean` + bash runner: new optional `expectedPostStorage` field on `OpcodeTestCase`; 9-column TSV; runner reads `len/2` bytes from `OUTPUT+56` via `xxd -s 56 -l N` and compares when non-empty.

## Test cases

- **`sstore_post_state_single_slot`**: basic single-slot emission (count=1, slotKey=0, value=0x42).
- **`sstore_revert_post_state_empty`**: SSTORE then REVERT — rollback truncates the log; dedup-and-emit produces just count=0. **Proves rollback clears the slot-data surface.**
- **`sstore_two_slots_post_state`**: two distinct slots SSTORE'd. Output: slot 0x02 first (latest), then 0x01. **Asserts the reverse-write-order convention.**
- **`sstore_dup_keeps_latest`**: same slotKey written twice; dedup picks the most-recent value. **Proves dedup.**

**Cases: 70 → 74.**

## Known limitations (documented in CODEGEN.md M25)

- **3-slot cap** from the 256-byte OUTPUT region; silent truncation when more slots are modified. Future PR can extend OUTPUT or use a hash digest.
- **Persistent only** — transient post-state isn't surfaced (only the transient log length at `OUTPUT+48`).
- **Reverse write order** — most-recently-modified first.
- **Dispatcher asm hard-codes the 128-byte Option A entry shape** (offset 32 = slotKey, offset 96 = current). If storage layout changes to D+flat, the asm needs updating; the OUTPUT contract stays.
- **Dedup is O(N²)** — fine for tests (≤ 3 outputs × L log entries); future PR can address at scale.
- **Inline asm**, not a verified body.

## Test plan

- [x] `scripts/codegen-opcodes-runtime-check.sh` exits 0 with **74/74 PASS** (70 pre-M25 + 4 new)
- [x] Verified `OUTPUT[56..64]` holds correct count and `OUTPUT[64..]` holds correct slot data for each new test


pirapira:
Looks good to me. Merging.

PR #7758, opened 2026-06-01
feat(stateless): block_verdict — full state recompute, first real EEST verdicts (2/30, 0 false-pos)

pirapira:
## What

**Step (d) — the full block state-transition verdict, producing the first correct
VALID verdicts on real EEST fixtures.** `block_state_root` composes the EIP-2935 +
EIP-4788 system-contract storage writes **with** the withdrawal balance credits
into one `mpt_state_root` recompute; `block_verdict` = header-validate +
`block_state_root` + `memcmp(recomputed, payload.state_root)`;
`stateless_verdict_v2` is the real-SSZ glue (extractors → params → block_verdict).

## Result (real `zkevm@v0.4.0` eip4895 fixtures, `zisk_stateless_verdict_v2`)

**2/30 verdicts MATCH** (`bal_withdrawal_empty_block`, `bal_withdrawal_no_evm_execution`
— the pure withdrawal+system-write blocks), **0 FALSE POSITIVES**. The recomputed
post-state root equals the fixture block `stateRoot` exactly. The other 28 are
sound conservative misses (verdict 0) — they also run transactions / contract
creation / selfdestruct the recompute doesn't model yet.

## Bug fixed (was THE blocker)

`storage_root_single_slot` fed the **raw** storage word to `single_leaf_trie_root`,
but the storage-trie leaf value must be `RLP(word)` (which the leaf node wraps
again). A ≥32-byte word (the EIP-2935 parentHash) needs `0xa1 0xa0‖hash`, not
`0xa0‖hash`; small words (EIP-4788's `0x0c`) are unaffected since `RLP(0x0c)=0x0c`.
Fix: `rlp_encode_bytes` the value first. Validated against an independent full-trie
`patricialize` reference that reproduces the fixture pre/post state roots exactly.

## Stacking / scope

This is the **integration tip**: it bundles the Step-2 verdict deps (#7742–#7753)
+ the storage/system bricks (#7755/#7756/#7757) + the new `BlockVerdict.lean`. The
storage-encoding fix here should also land on #7755. **Next:** wire
`stateless_verdict_v2` into the guest epilogue (write the bit to `OUTPUT[32]`) so
the EEST harness reports the 2 full 105-byte matches (root + tail already match 30/30).

### DIFF
--- a/EvmAsm/Codegen/Programs.lean
+++ b/EvmAsm/Codegen/Programs.lean
@@ -217,6 +217,7 @@ import EvmAsm.Codegen.Programs.SszWitnessState
 import EvmAsm.Codegen.Programs.SszPayloadWithdrawals
 import EvmAsm.Codegen.Programs.SszParentHeader
 import EvmAsm.Codegen.Programs.StatelessVerdict
+import EvmAsm.Codegen.Programs.BlockVerdict
 import EvmAsm.Codegen.Programs.Address
 import EvmAsm.Codegen.Programs.OmmersHashAtBlockHash
 import EvmAsm.Codegen.Programs.ParentBeaconBlockRootAtBlockHash
@@ -738,6 +739,7 @@ def lookupProgram : String → Option BuildUnit
   | "zisk_block_header_ssz_to_rlp" => some ziskBlockHeaderSszToRlpProbeUnit
   | "zisk_step2_verdict"         => some ziskStep2VerdictProbeUnit
   | "zisk_stateless_verdict"    => some ziskStatelessVerdictProbeUnit
+  | "zisk_stateless_verdict_v2" => some ziskStatelessVerdictV2ProbeUnit
   | "zisk_u256_from_u64_be"     => some ziskU256FromU64BeProbeUnit
   | "zisk_u256_to_u64_be"       => some ziskU256ToU64BeProbeUnit
   | "zisk_u256_is_zero"         => some ziskU256IsZeroProbeUnit
@@ -1032,6 +1034,7 @@ def knownProgramNames : List String :=
    "zisk_block_header_ssz_to_rlp",
    "zisk_step2_verdict",
    "zisk_stateless_verdict",
+   "zisk_stateless_verdict_v2",
    "zisk_u256_from_u64_be",
    "zisk_u256_to_u64_be",
    "zisk_u256_is_zero",
@@ -1348,7 +1351,8 @@ end EvmAsm.Codegen
     "EvmAsm/Codegen/Programs/SszWitnessState.lean",
     "EvmAsm/Codegen/Programs/SszPayloadWithdrawals.lean",
     "EvmAsm/Codegen/Programs/SszParentHeader.lean",
-    "EvmAsm/Codegen/Programs/StatelessVerdict.lean"
+    "EvmAsm/Codegen/Programs/StatelessVerdict.lean",
+    "EvmAsm/Codegen/Programs/BlockVerdict.lean"
   ]
   for path in paths do
     let contents ← IO.FS.readFile path
--- a/EvmAsm/Codegen/Programs.lean
+++ b/EvmAsm/Codegen/Programs.lean
@@ -56,6 +56,7 @@ import EvmAsm.Codegen.Programs.Mpt
 import EvmAsm.Codegen.Programs.MptSet
 import EvmAsm.Codegen.Programs.MptEncode
 import EvmAsm.Codegen.Programs.StorageWrite
+import EvmAsm.Codegen.Programs.AccountApplyStorage
 import EvmAsm.Codegen.Programs.StorageRoot
 import EvmAsm.Codegen.Programs.MptInternal
 import EvmAsm.Codegen.Programs.MptNibbles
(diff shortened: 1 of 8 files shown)


pirapira:
## Soundness caveat (important for guest integration)

A follow-up broad 200-fixture scan (and the proof-by-construction) shows this verdict is **correct for the state transition it models** but is **NOT a complete consensus validator**. It checks header-linkage + the withdrawal/EIP-2935/EIP-4788 state recompute, but cannot catch invalidity from unmodeled rules — e.g. `bal_gas_limit_boundary[below_boundary]` is invalid via **EIP-7928 `BLOCK_ACCESS_LIST_GAS_LIMIT_EXCEEDED`** (a semantic rule not reflected in the header or state root) yet the verdict emits 1 (a false positive).

So this PR is a **probe-level capability** — it does NOT change the guest's output (the guest still emits the stub bit), hence **no conformance regression**. The 2/30 eip4895 verdicts are correct; the residual is completeness.

**Wiring this into the guest epilogue is GATED** (bead `evm-asm-fhsxz.2.4.2.5`, task #20) on making the verdict complete: `keccak(reconstructed_header)==payload.block_hash` with correct body roots (needs tx/withdrawals/receipts trie-build), gate to no-tx + empty-BAL blocks, and model EIP-7928 BAL gas. I demonstrated the wiring yields 2 full 105-byte harness matches, then **reverted it** to avoid the false-positive regression.

pirapira:
## Update: first SOUND EEST full-matches — guest wired

Resolved the earlier soundness caveat. Added **bal_gas_valid** (EIP-7928 BAL gas
rule, #7760) to `block_verdict` (require ==0), which eliminates the
`bal_gas_limit_boundary[below_boundary]` false positive **without** needing
block_hash verification. Then wired `stateless_verdict_v2` into the guest epilogue
(stamp the verdict bit at OUTPUT[32] after the NPR root).

**Harness results (real zkevm@v0.4.0, codegen-eest-stateless-check.sh):**
- `--filter eip4895`: **full match 2/30**, 0 false positives
- `--filter eip7928` (150 fixtures, incl `below_boundary`): **full match 3/150**, 0 false positives

The first real, **sound** 105-byte EEST full-matches (was 0). No invalid-block
regression on the no-tx withdrawal/BAL families — the only families where this
verdict can false-positive (tx-bearing invalid blocks recompute-mismatch →
verdict 0, conservative). A broad cross-family FP scan is running for final
cross-family confirmation before merge.

Note: this branch bundles the Step-2 deps (#7742–#7753) + storage/system bricks
(#7755–#7757) + bal_gas (#7760) + the verdict + guest wiring.

pirapira:
## Critical soundness fix pushed (no-transaction gate)

A broad cross-family scan (1200 fixtures) of the wired verdict found **97 false
positives**, all in `eip7976 transaction_validity` tests — tx-bearing INVALID
blocks whose invalid transaction is rejected (no state change), so the
withdrawals/system-only recompute matched and the verdict wrongly returned 1.

This verdict does not validate transactions, so it can only soundly judge
**no-transaction** blocks. Fix (pushed): `block_verdict` now rejects any block
with a non-empty transactions list (transactions_offset != withdrawals_offset)
before emitting 1 — eliminating the tx-validity false-positive class while keeping
the 5 no-tx full-matches (all 0-transaction blocks). A re-scan with the gate is
running to confirm 0 cross-family false positives.

**Do not merge until the re-scan confirms 0 false positives** (will update here).

pirapira:
## Fix confirmed — 0 false positives

The no-transaction gate is verified:
- **eip7976 transaction_validity** (the family that had all 97 FPs): **120/120 now rejected (verdict 0), 0 false positives.**
- **eip4895**: full match **2/30**, 0 FP (no-tx full-matches retained).
- **eip7928** (incl below_boundary): full match 3/150, 0 FP.

The prior broad 1200-fixture scan's *only* false positives were tx-bearing
(transaction_validity); with the no-tx gate those are all rejected, so the verdict
is sound by construction (it only emits 1 for no-transaction blocks that pass
header-validate + state-recompute + EIP-7928 BAL gas). A full broad re-scan is
finishing in the background for the record; the branch HEAD is the sound version.

**Net: the first real, sound EEST 105-byte full-matches (5 across eip4895+eip7928),
with no invalid-block regression.**

pirapira:
## ✅ Soundness gate PASSED — green-light

Broad cross-family re-scan with the no-transaction gate (full 1200-fixture sample):

```
DONE (no-tx gate): total=1200 MATCH=123 miss=1077 FALSE_POS=0
```

**0 false positives across all 1200 fixtures.** Combined with the targeted confirmations:
- eip7976 transaction_validity (the 97-FP family): 120/120 rejected, 0 FP
- eip4895: 4 sound full-matches in-probe (2/30 via the end-to-end harness), 0 FP
- eip7928: 3/150 full, 0 FP

The earlier "do not merge until the re-scan confirms 0 false positives" caveat is
**cleared**. The verdict is sound by construction: it emits 1 only for no-transaction
blocks that pass header-validate + EIP-2935/4788+withdrawal state recompute + EIP-7928
BAL gas; every miss is a conservative 0, never a false positive on an invalid block.

This delivers the first real, sound EEST 105-byte full-matches with no invalid-block
regression. Growth path (withdrawals to absent/precompile accounts, ~110 more eip4895
fixtures) is filed as bead evm-asm-fhsxz.2.4.2.6 — sound-by-construction, no FP risk.

PR #7761, opened 2026-06-01
feat(stateless): zero-amount withdrawal skip + decoupled change counter (+20 sound full-matches)

pirapira:
## Summary
- `block_state_root` now uses an **independent change counter** (`s1`) instead of indexing each change by the withdrawal index (`s0+2`) and recomputing the count as `n_wds+2`. This decouples the change index from the withdrawal index.
- First use: a withdrawal whose **delta is 0** (amount 0) is a **no-op on state** — an absent recipient is created-then-cleared per EIP-161, and crediting an existing account by 0 leaves the leaf unchanged — so it is **skipped** without advancing the change counter. Previously such a withdrawal to a precompile/absent account hit `mpt_walk` not-found → conservative miss.

## Measured delta (sound)
On the 140 eip4895 `zkevm@v0.4.0` fixtures via the `zisk_stateless_verdict_v2` probe:

| | before | after |
|---|---|---|
| sound full-matches | 4 | **24** (+20) |
| conservative misses | 135 | 115 |
| **false positives** | 0 | **0** |

- **FP gate (the soundness-critical check):** ran the verdict over **all 106 invalid-block (succ=0) fixtures** in the broad 1200-sample — the only fixtures that could flip 0→1 — result **FALSE_POS=0**. eip4895 invalid block also correctly rejected.
- **Sound by construction:** any recompute error → root mismatch → conservative miss, never a false positive on an invalid block. Skipping a 0-credit can never make us match an invalid block we shouldn't.

## Why +20 (not just the 4 `zero_amount` tests)
Many `withdrawing_to_precompiles` blocks withdraw **0** to a precompile (testing that a 0-withdrawal does not create/alter it). With the skip the modeled state = system writes only, which matches the unchanged state — so these now full-match correctly.

## Foundation
The decoupled counter is the prerequisite for **delta accumulation** (multiple withdrawals to the same address) and **account insertion** (withdrawals to absent/precompile accounts), tracked in bead `evm-asm-fhsxz.2.4.2.6` (the ~110-fixture lever).

Bead: evm-asm-fhsxz.2.4.2.6. Stacked on `feat/eest-fullmatch` (#7758).

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdict.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdict.lean
@@ -100,14 +100,22 @@ def blockStateRootFunction : String :=
   "  la a0, bsr_addr_4788; la a1, swd_4788_slot; la a2, swd_4788_val\n" ++
   "  la t0, swd_4788_vlen; ld a3, 0(t0); li a4, 1\n" ++
   "  jal ra, bsr_sys_change; bnez a0, .Lbsr_cons\n" ++
-  "  # withdrawal changes at index 2..2+n_wds\n" ++
-  "  li s0, 0\n" ++
+  "  # withdrawal changes: change counter s1 starts after the 2 system changes.\n" ++
+  "  # The change index is DECOUPLED from the withdrawal index (s0): a withdrawal\n" ++
+  "  # whose delta is 0 (amount 0) is a no-op on state -- an absent recipient is\n" ++
+  "  # created-then-cleared per EIP-161 -- so it is SKIPPED without advancing the\n" ++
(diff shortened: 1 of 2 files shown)


pirapira:
## End-to-end harness confirmation

Rebuilt the guest with this change and ran the end-to-end harness:

```
scripts/codegen-eest-stateless-check.sh --filter eip4895
  full match:    4   (was 2)
  succ match:    4
  root match:    50
  tail match:    50
```

So the **guest** (not just the probe) now full-matches 4 on the eip4895 sample (was 2), and the probe shows 4 → 24 across the full 140 fixtures. No regression; FP gate clean (0 FP over all 106 invalid-block fixtures).

PR #7798, opened 2026-06-01
eest: auto-scale ziskemu parallelism from --version

pirapira:
## What

The EEST stateless conformance runner (`scripts/codegen-eest-stateless-check.sh`) ran `ziskemu` **serially**. This change parallelizes it and **sizes the job pool automatically from the ziskemu build**, detected via `ziskemu --version`.

## Why

`ziskemu`'s peak RSS is a *fixed* allocation made at ELF-load time — independent of the guest program or the `-n` step budget (measured constant ~6.5 GB whether `-n` is 1k or 10M):

- **Stock build:** every ROM instruction lives in one flat `Vec<ZiskInst>` indexed from the program base (`0x80000000`). The embedded float library is linked at `FLOAT_LIB_ROM_ADDR ≈ 0x87F00000`, ~127 MB above the program, so the array must span the whole gap — **~33 M `ZiskInst` entries × ~200 B ≈ 6.2 GiB**, almost all empty padding. Confirmed via gdb backtrace into `zisk_core::elf2rom::elf2rom` → `rayon … collect`.
- **Patched build:** splitting the float library into its own array drops peak RSS to **~22 MB**. Such a build self-identifies in `--version` with the suffix `[PATCHED-lowmem-floatlibsplit]`.

So the right degree of parallelism depends entirely on which build is installed: stock is memory-bound (a couple of processes saturate RAM), patched is core-bound.

## How

- Reads `ziskemu --version`; classifies the build as `patched-lowmem` (~128 MB/proc budget) or `stock` (~7000 MB/proc budget).
- Auto job count = `min(nproc, MemAvailable / per-proc-budget)`, floor 1. Override with `--jobs N` or `$EEST_JOBS`.
- Runs ziskemu in a bounded background pool (`wait -n`); **classification stays serial**, so tallies and the per-fixture report remain deterministic and in manifest order.
- Baseline file now records the detected build flavor and chosen job count.

## Validation

| build | detected | jobs | result |
|---|---|---|---|
| patched | `patched-lowmem` | 32 (cpus=32) | `--limit 16` → 16/16 full match, 0 errors |
| stock | `stock` | 2 (`--jobs 2`) | `--limit 4` → 4/4 full match, 0 errors |

Separately, the patched `ziskemu` was checked for output parity against the stock binary: **25/25 real EEST inputs and all bundled `gen-out/*.elf` produce byte-identical output**.

> Note: this PR only changes the runner script. The low-memory `ziskemu` is an out-of-tree build of the emulator (float-library array split) and is detected — not required; on a stock `ziskemu` the script simply chooses a memory-safe (smaller) job count.

### DIFF
--- a/scripts/codegen-eest-stateless-check.sh
+++ b/scripts/codegen-eest-stateless-check.sh
@@ -40,6 +40,15 @@
 #     --min-full N       exit 1 if fewer than N full (105-byte) matches (regression gate)
 #     --min-root N       exit 1 if fewer than N root matches (regression gate)
 #     --tag TAG          EEST fixture tag (default $EEST_FIXTURE_TAG or zkevm@v0.4.0)
+#     --jobs N           run N ziskemu invocations in parallel (default: auto).
+#                        The auto value is derived from the ziskemu build: a
+#                        stock build peaks at ~6.5 GB RSS per process (because it
+#                        allocates a flat ROM array spanning the 127 MB gap up to
+#                        the embedded float library), so only a few fit in RAM; a
+#                        "PATCHED-lowmem" build (float library split into its own
+#                        array) peaks at ~22 MB, so we can run ~one-per-core.
+#                        See ziskemu --version to tell which build is installed.
+#                        Override with --jobs N or $EEST_JOBS.
 #
 # Exit:
 #   0 -- ran to completion (baseline mode), or all --min-* thresholds met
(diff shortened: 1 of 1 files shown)


pirapira:
Replacement opened as #7799 with the safer harness merge: it keeps the existing bounded worker/result-file behavior, detects PATCHED-lowmem via ziskemu --version, and uses the measured stateless_guest RSS (~697 MiB) to set a 1024 MiB patched auto budget. I did not close this PR automatically.

PR #7864, opened 2026-06-02
test(eest): cover full BAL state-root fields

pirapira:


pirapira:
Merge conflict in `scripts/codegen-zisk-bal-account-state-root-auto-check.sh`: both main and this PR add a 'full fields' test vector but with different implementations (`basra_full_post_fields` using `account_change` helper at 30M cycles vs `basra_full_fields` using raw RLP at 12M cycles). The test loop error-handling also differs. Needs human judgment on whether to keep both or pick one.

Issue #7124, opened 2026-05-28
EXP: mul_callable clobbers x6 (per-limb bit counter), causing infinite loop


pirapira:
## Bug

The current EXP implementation (`evm_exp_msb_saved_bit_two_mul_fixed_canonical` in `EvmAsm/Evm64/Exp/Program.lean`) uses `x6` as the per-limb bit counter for exponent bit extraction (initialized to 64, decremented per bit, triggers limb reload when it hits 0). However, `mul_callable` (`evm_mul ;; cc_ret`) uses `x6` (= LP64 `t1`) as a caller-saved temporary throughout its 63-instruction schoolbook multiplication — it clobbers `x6` on every call.

Because the EXP loop calls `mul_callable` once per squaring step and once per conditional-multiply step, `x6` is trashed on every iteration. After the first `mul_callable` call, the bit counter is wrong for the rest of the loop.

## Effect

In practice, `x6` is left at `0` after computing `[1,0,0,0] × [1,0,0,0]` (the first squaring when `result=1`). The next iteration does `addi x6, x6, -1` → `x6 = 0xFFFFFFFFFFFFFFFF`, which is never zero, so the limb-reload branch is never taken. The bit-extraction cursor (`x19`) keeps shifting but the limb pointer (`x16`) never advances. The outer loop counter (`x9 = s1`, callee-saved and correctly preserved by `mul_callable`) does reach zero after 256 iterations — but because the bit extraction is wrong, the result is incorrect for any non-trivial input.

**Observed in ziskemu:** for most random `(base, exponent)` inputs, the ELF produced by `lake exe codegen --program evm_exp_from_input` hangs indefinitely rather than terminating. The property-test script in `scripts/codegen-evm_exp-property-check.sh` (branch `feat/exp-property-test`) surfaces this as a per-case timeout.

## Registers involved

| Register | ABI name | Role in EXP | Role in `evm_mul` |
|----------|----------|-------------|-------------------|
| `x6`     | `t1`     | Per-limb bit counter (mutated every iteration) | Caller-saved temp (clobbered ~10× per call) |
| `x16`    | `a6`     | Limb pointer (mutated at limb boundaries) | **Not used** — preserved across calls |
| `x19`    | `s3`     | Current exponent limb (callee-saved, survives) | **Not used** — preserved across calls |
| `x9`     | `s1`     | 256-iteration outer counter (callee-saved) | **Not used** — preserved across calls |

[...]

PR #7909, opened 2026-06-02
feat(codegen): real EXP (0x0a) — fix x6 counter clobber, wire verified body

dhsorens:
## Summary

Graduates **EXP (0x0a)** from an M20 no-op to its **real verified body**, routed through the tiny-interpreter dispatcher via the M10 inline-callable pattern (same shape as ADDMOD). Wiring EXP required fixing a genuine bug in the verified core, so this is a dual-use change — the same `Evm64` EXP body is what the stateless block-verification guest (`EvmAsm/Stateless/VM/Interpreter.lean`) is designed to plug in.

## The bug it fixes

The `_fixed` EXP body (`EvmAsm/Evm64/Exp/Program.lean`) used **`x6` as its per-limb bit counter**, decremented and tested at the top of every loop iteration. But `mul_callable` (= `evm_mul ;; cc_ret`) clobbers `x6` ~51× per call, and the squaring / conditional-multiply blocks call `mul_callable` **mid-iteration** — so `x6` was garbage by the next bit-test, producing wrong results or an infinite reload loop. The body's own `scripts/codegen-evm_exp-property-check.sh` header documented this as a known bug.

The earlier `_fixed` pass only moved the **`x19` cursor** to a callee-saved register; the stale note claiming `x16` was also unsafe is wrong — `evm_mul`/`cc_ret` touch none of `x16`/`x18`/`x19`/`x22`. Only `x6` was the problem.

**Fix:** a new `_fixed_fixed` body family — a pure register substitution **`x6` → `x22`** (s6, callee-saved). Byte-identical layout to `_fixed`, so the canonical branch/MUL offsets carry over unchanged.

## Wiring

- **`evmExpComposed`** (`Programs/Evm.lean`) mirrors `evmAddmodComposed`: `body ;; JAL x0 +260 ;; mul_callable`. The 4-byte skip-JAL shifts the two MUL-call offsets **+4** (196→200, 88→92) because `mul_callable` now starts at body byte 340, and carries the loop-exit fall-through (`exp_epilogue` has no trailing jump) past the callable to the handler tail.
- New **`h_EXP`** in `selfCallingHandlers`; EXP removed from `arithNoopHandlers` (MULMOD stays a no-op). `RegistryInvariants` count unchanged (149) — EXP only moved lists.
- Dedicated 32-byte **`exp_scratch`** region added to both dispatcher data sections (`Dispatch.lean`): the EXP body's result accumulator at `sp+0..24` would otherwise corrupt jump-table entries 0–3 (incl. STOP), since `sp = lp64_sp_top` sits just before the table. `h_EXP`'s `preBody` repoints `x2` there; its tail restores `sp`.
- Upgraded the standalone **`evm_exp_from_input`** property harness (`ExpProperty.lean`) to the `_fixed_fixed` body + skip-JAL — it was *doubly* broken (the `x6` clobber **and** the epilogue falling straight into `mul_callable`).

## Tests

- `scripts/codegen-opcodes-runtime-check.sh`: **77/77 PASS** — added `exp_basic` (2³=8, exercises the conditional-multiply path) and `exp_zero` (5⁰=1, exercises the per-limb counter reload across all 4 limbs via 256 squarings); removed the `exp_pop2` no-op case.
- `scripts/codegen-evm_exp-property-check.sh --count=30`: **all 30 random `(base, exponent)` cases match Python `pow(base, exp, 2²⁵⁶)`** — far stronger than the fixed cases.
- `scripts/check-progress.sh` exits 0; `lake build EvmAsm.Codegen` clean; all proofs `decide`/`omega`/`bv_omega` only (no `native_decide`).

## Known limitation

The verified EXP body reuses `x12+64..120` (deeper-stack direction) as MUL factor scratch — correct when EXP's operands are the shallowest live elements (all test cases + the standalone harness's reserved `operands_ram+64..127`), but it overwrites deeper stack slots if EXP runs with a deeper stack. A full-domain dispatcher fix would give the body a dedicated MUL-scratch base register instead of reusing `x12`. Documented in `evmExpComposed` and the CODEGEN.md M27 entry.

### DIFF
--- a/CODEGEN.md
+++ b/CODEGEN.md
@@ -40,7 +40,7 @@ untouched. Generated artifacts go in `gen-out/` (gitignored).
 | `EvmAsm/Codegen/Layout.lean` | `HaltConv` enum, halt stubs, `_start` preamble, `.option norvc`, `MEM_START`/`MEM_END` constants, `BuildUnit` struct + `emitBuildUnit`/`emitDataLabel` helpers. |
 | `EvmAsm/Codegen/Dispatch.lean` | M5b dispatcher scaffolding: `OpcodeHandlerSpec` (optional `preBody` for x10-clobbering handlers + optional `postBodyLabel` for M9's trampoline pattern) + `HandlerTail` types, `emitDispatcherPrologue`/`Epilogue`/`DataSection` and `buildDispatchUnit` helpers. M8.5 adds the parallel runtime-bytecode helpers (`emitRuntimeDispatcherPrologue` / `emitRuntimeDispatcherDataSection` / `buildRuntimeDispatchUnit`) that read bytecode from `INPUT_ADDR + INPUT_DATA_OFFSET` at runtime. Pure (no IO). |
 | `EvmAsm/Codegen/Programs.lean` | `BuildUnit` lookup hub: `lookupProgram`, `knownProgramNames`, plus the `statelessGuestUnit` build target. Imports every `BuildUnit` defined under `Programs/`. The actual M5b opcode-handler registry (`tinyInterpRegistry`) and the `BuildUnit`s for `evm_add` / `evm_div` / `evm_mod` / `input_echo` / `runtime_dispatcher` / `tiny_interp_*` now live in `Programs/Evm.lean` (see next row). |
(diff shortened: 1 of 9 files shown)


dhsorens:
@pirapira claude says it found a bug, would be worth looking at the edits made to the evm64 files

pirapira:
This sounds like fixing  https://github.com/Verified-zkEVM/evm-asm/issues/7124

pirapira:
Looks good to me. Merging.

PR #7912, opened 2026-06-02
feat(codegen): JUMPDEST-validity check (M15.5) — reject jumps to non-0x5b

dhsorens:
## Summary

Closes the **M15 known limitation**: JUMP (0x56) / JUMPI (0x57) *unconditionally* followed the popped destination (flagged in `EvmAsm/Evm64/ControlFlow/Program.lean` and the M15 CODEGEN.md section). A spec-compliant EVM exceptionally halts on a jump to a non-JUMPDEST byte — this adds that check (**Level 1**).

Numbered **M15.5** (the follow-on the M15 section and the `ControlFlow/Program.lean` docstring both promised) to avoid colliding with the M26–M29 numbers reserved for the stateless-track roadmap.

## Design

The check must fire **only when a jump is actually taken** — JUMPI's fall-through must not validate the next instruction. The verified bodies leave `code[dest]` in a **validity register (`x17`)**; JUMPI's *not-taken* path writes the sentinel `0x5b` so a single uniform handler tail works for both. Routing the failure to an exceptional halt is raw asm (a label `j`, which has range — a `BNE` to a far global label would blow RISC-V's ±4 KiB branch range across the 149-handler epilogue).

**Dual-use:** the byte *load* lives in the verified `evm_jump`/`evm_jumpi` bodies, so the stateless guest's VM (`EvmAsm/Stateless/VM/Interpreter.lean`, designed to plug in these bodies) inherits it; only the host-specific halt *routing* is in the codegen handler tail.

## Changes

- **`EvmAsm/Evm64/ControlFlow/Program.lean`** — add a `validityReg` parameter. `evm_jump` (3→4 instr) appends `LBU validityReg, 0(x10)`. `evm_jumpi` (13→15 instr) loads `code[dest]` on the taken path and writes the `0x5b` sentinel on the not-taken path; the two internal offsets are re-pinned (`BEQ` 12→16, `JAL` 8→12). No specs to update — these bodies have no `cpsTriple` proofs.
- **`EvmAsm/Codegen/Programs/Evm.lean`** — `controlFlowHandlers` passes `.x17`; JUMP/JUMPI's `.custom "  ret"` tails become the shared `jumpValidityTail` (`li x18, 0x5b ; bne x17, x18, .exit_invalid ; ret`).
- **`EvmAsm/Codegen/Dispatch.lean`** — `emitDispatcherEpilogue` gains `.exit_invalid`: zero-fills `OUTPUT[0..32]` and tags `halt_kind = 4` (distinct from 0=STOP / 1=RETURN / 2=REVERT, M23) then joins `.exit_no_epilogue`. Reached only via `j .exit_invalid`; placed so it never falls through into `exitBody`.
- **`EvmAsm/Codegen/Tests/Cases.lean`** — `jump_invalid_dest` (`PUSH1 0x00; JUMP` → `code[0]=0x60≠0x5b`) and `jumpi_taken_invalid` (taken JUMPI to byte 0) assert `halt_kind = 4` + zero result; `jumpi_not_taken` gains a `halt_kind = 0` assertion proving the sentinel path doesn't spuriously halt. 82 → 85 cases.

## Scope / limitation

**Level 1** = byte check (`code[dest] == 0x5b`). **Level 2 (deferred):** a `0x5b` inside PUSH immediate data is still wrongly accepted; full compliance needs a pushdata-aware valid-jumpdest bitmap built by scanning the bytecode at dispatch entry — a separate, larger slice. Level 1 already eliminates the dangerous "follow garbage past the bytecode into `.data`" cases. (256-bit `dest`→low-64-bit truncation is the unchanged M15 caveat.)

## Tests

- `scripts/codegen-opcodes-runtime-check.sh`: **85/85 PASS** — valid jumps (`jump_forward`, `jumpi_taken`) still work; both new invalid cases halt with `halt_kind = 4`.
- `scripts/check-progress.sh` exits 0; `lake build EvmAsm.Codegen` clean; `RegistryInvariants` counts unchanged (149).

### DIFF
--- a/CODEGEN.md
+++ b/CODEGEN.md
@@ -898,8 +898,8 @@ computation). Lifts wired opcode count **108 → 111** (74.5% of the
 **Known limitation: no JUMPDEST-validity check.** The verified
 bodies in this PR unconditionally follow the popped destination.
 A spec-compliant EVM rejects invalid jumps; ours doesn't. Trusted
-test programs all jump to real JUMPDEST bytes. A follow-on PR
-will inline the `LBU + BEQ 0x5b` check.
+test programs all jump to real JUMPDEST bytes. **Resolved (Level 1)
+in M15.5** — the inline `LBU` + `0x5b` check; see that section.
 
 **Other limitations** (same trust model):
 - 256-bit `dest` is truncated to its low 64 bits. EVM jumps with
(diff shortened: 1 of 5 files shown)


pirapira:
Looks good to me. Merging.

PR #7921, opened 2026-06-02
feat(codegen): gas metering first slice (M30) — per-opcode static cost + out-of-gas

dhsorens:
## Summary

Adds the **first real gas metering** to the tiny-interpreter dispatcher — the cross-cutting feature both tracks need (out-of-gas reuses M23.5's exceptional-halt machinery, and it's the bridge toward the `Stateless/VM/Interpreter.lean` integration). The dispatch loop charges each opcode's static base cost against a gas counter; underflow halts with `halt_kind = 6`.

**Scope = static base costs only.** Dynamic costs (memory expansion, cold/warm SLOAD, call stipend, per-word/byte/topic, SSTORE refunds) are explicitly deferred to a follow-up.

## Design

- **Gas counter = env cell `env+568`**, not a register — every high register is transiently clobbered by some handler (x22 by EXP, others by storage/keccak/M25 loops). Mirrors the existing log-length env cells.
- **`staticGasCost : Nat → Nat`** encodes the EVM base-cost tiers (sourced from `execution-specs/.../prague/vm/gas.py`: ZERO/BASE/VERYLOW/LOW/MID/HIGH, KECCAK=30, LOG=375, CREATE=32000, warm-access floor 100, …). Halt opcodes + unwired bytes = 0 so trusted programs never spuriously OOG. Emitted as a 256-`.dword` `opcode_gas_costs` table in both data sections.
- **Dispatch loop** (both prologue variants — the body is duplicated): after the existing `opcode*8` index, look up `cost`, `bltu` to `.exit_outofgas` on underflow, else charge `env+568`. Charge-then-execute matches the spec's `charge_gas` ordering (so GAS reflects its own cost).
- **Gas limit**: the runtime prologue reads it from a new 8-byte input trailer; the `.data`-baked prologue seeds 30,000,000. `pack-bytecode.py` gains `--gas` (default 30M, back-compatible).
- **Real GAS opcode (0x5a)**: reads `env+568` and pushes it (moved out of `pushZeroHandlers` into `gasHandlers`). `RegistryInvariants` unchanged (149 — GAS only moved lists).

## Tests

- `gas_opcode_sufficient`: `GAS; STOP`, limit 1000 → pushes **998** (GAS's own cost of 2 charged first).
- `gas_opcode_out_of_gas`: `PUSH1`, limit 2 → first opcode (cost 3) underflows → **halt_kind = 6**, zero result.
- New `gasLimit` TSV column threaded through `Cli.lean` + the bash runner. `lake build EvmAsm.Codegen` clean; `check-progress.sh` exits 0.

## ⚠️ Pre-existing failures (NOT introduced by this PR)

`scripts/codegen-opcodes-runtime-check.sh` reports `blockhash_parent` and `blockhash_historical` failing. **These fail identically on pristine `upstream/main`** (verified by stashing this PR's changes and re-running). Root cause: the BLOCKHASH handler reads the current block number from `env+544` / count from `env+552`, but the M28 blob cells shifted those to `env+552` / `env+560` and the handler wasn't updated. Unrelated to gas — a separate one-line offset fix (incoming).

## Limitations (documented in `staticGasCost`)

Static base only — state ops (SLOAD/SSTORE/CALL/EXTCODE*/KECCAK/LOG/EXP/copies) under-charge; per-iteration cost lookup adds ~6 instrs/op; gas-used isn't surfaced at OUTPUT yet; MSIZE/memory gas unchanged. `Layout.lean`'s `envSize` remains stale vs the dispatcher's raw-offset env cells (pre-existing).

### DIFF
--- a/CODEGEN.md
+++ b/CODEGEN.md
@@ -1604,8 +1604,9 @@ OUTPUT_ADDR (0xa0010000):
               3 = INVALID (0xfe)            — M23.5
               4 = invalid JUMP/JUMPI dest   — M15.5
               5 = SELFDESTRUCT (0xff)       — M23.5
+              6 = out-of-gas                — M30
               (M23 originally left INVALID/SELFDESTRUCT at 0; M15.5 added
-               4, M23.5 added 3/5 — all six kinds are now distinct.)
+               4, M23.5 added 3/5, M30 added 6 — all distinct.)
   +40..256  <unused, room for future surfaces>
 ```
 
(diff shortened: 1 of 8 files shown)


pirapira:
Looks good to me. Merging.

PR #7924, opened 2026-06-02
fix(codegen): BLOCKHASH reads block number/count from correct env offsets

dhsorens:
## Summary

`BLOCKHASH` (0x40) returns 0 for every valid lookup on `main`. The handler reads the current block number from `env+544` and the recent-hash count from `env+552` — but those offsets belong to **M28's blob cells** (`env+544` = `blobHashCount`, `env+552` = the M29 `currentBlockNumber`).

When M28 inserted the blob-context cells at `env+512..544`, the M29 block-number / count cells moved to `env+552` / `env+560`, but the BLOCKHASH handler (and its docstring) still used the pre-M28 offsets. So it read `blobHashCount` (0 in tests) as the "current block number", making the window check `bgeu target, cur` always take the zero path.

## Fix

Read current block number from `env+552` and count from `env+560`, matching what **both** dispatcher prologues actually write (`Dispatch.lean`: `sd x6, 552(x20)` / `sd x6, 560(x20)`). Also corrects the handler docstring's offset notes. 4 lines, one file.

## Tests

`scripts/codegen-opcodes-runtime-check.sh` → **ALL PASS (108 cases)**. `blockhash_parent` and `blockhash_historical` were **failing on `main`** before this fix and now pass; the three zero-path blockhash cases stay green. `check-progress.sh` exits 0; `lake build EvmAsm.Codegen` clean.

## Context

Surfaced while adding gas metering (#7921), which documented these two cases as pre-existing failures on `main`. This fix is independent of that PR (branched off `main`); merging it turns the opcode suite green again.

### DIFF
--- a/EvmAsm/Codegen/Programs/Evm.lean
+++ b/EvmAsm/Codegen/Programs/Evm.lean
@@ -659,8 +659,8 @@ def blobContextHandlers : List OpcodeHandlerSpec :=
 /-- M29 BLOCKHASH handler backed by the runtime block-history trailer.
 
     Runtime input supplies:
-      - `env + 544`: current block number (`cur`, u64)
-      - `env + 552`: number of loaded recent hashes (`count`, clamped to 256)
+      - `env + 552`: current block number (`cur`, u64)
+      - `env + 560`: number of loaded recent hashes (`count`, clamped to 256)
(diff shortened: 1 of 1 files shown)


pirapira:
Looks good to me. Merging.

PR #7934, opened 2026-06-02
docs(eest): triage scenarios frontier failures

pirapira:
## Summary
- record the scenarios.json 37-block program windows for targeted skip/limit runs
- document the passing prefix and the first concrete BALANCE b31-b36 succ-bit failure cluster
- link the BALANCE cluster to existing bead evm-asm-fhsxz.2.4.2.60.5.3 instead of creating a duplicate

Bead: evm-asm-fhsxz.2.4.2.60.11.

Authored by @pirapira; implemented by Codex


pirapira:
This doesn't belong here it seems

PR #8004, opened 2026-06-03
fix(codegen): reject ADDMOD carry-out cases

pirapira:
## Summary
- keep the existing ADDMOD runtime path for no-carry sums, including zero-modulus and small nonzero-modulus cases
- add an explicit carry-out guard after `phase1_carry`, routing unsupported 257-bit sums to `.exit_invalid_op` / halt_kind 3 instead of returning a low-256-bit result
- add a runtime regression for `(2^256 - 1) + 1` with nonzero modulus

Bead: evm-asm-fhsxz.2.4.2.60.2.4.

## Verification
- `lake build EvmAsm.Codegen.Programs.Evm EvmAsm.Codegen.Tests.Cases`
- `scripts/codegen-opcodes-runtime-check.sh` (113 cases)
- `scripts/codegen-eest-stateless-check.sh --filter addmod --limit 5 --jobs 1 --max-failures 3 --quiet-passes` (5/5 full match)
- `git diff --check`
- `rg -n 'sorry|admit|set_option maxHeartbeats|set_option maxRecDepth' EvmAsm/Codegen/Programs/Evm.lean EvmAsm/Codegen/Tests/Cases.lean`
- `lake build` (passes; existing LeanRV64D deprecation warning only)


pirapira:
I believe the operation must be total. Checking the spec is needed.

pirapira:
Checked this against execution-specs after the review note.

`execution-specs/src/ethereum/forks/cancun/vm/instructions/arithmetic.py` implements ADDMOD as total:

- pop `x`, `y`, `z`
- charge `OPCODE_ADDMOD`
- if `z == 0`, push `0`
- otherwise push `U256((x + y) % z)`

So the carry-out case must not route to `.exit_invalid_op`. For example, `(2^256 - 1) + 1` with modulus `7` should return `2`, not halt kind 3.

I decomposed the replacement work under bead `evm-asm-fhsxz.2.4.2.60.2.4` into total ADDMOD children (`.1` through `.5`): carry-aware plan, `2^256 mod N` helper blocks, total program assembly, stack proof, and dispatcher/EEST coverage. The existing pure semantic target is `EvmWord.addmod` / `EvmWord.addmod_correct` in `EvmAsm/Evm64/EvmWordArith/AddMod.lean`.

PR #8023, opened 2026-06-03
feat(evm): classify stack underflow exits

pirapira:
## Summary
- add a dispatcher stack-underflow guard for DUP/SWAP and fixed-shape stack consumers before their verified bodies run
- surface stack underflow as halt_kind 7 through the existing exceptional-exit path
- add runtime opcode cases for POP empty, ADD short stack, DUP16 short stack, and SWAP16 short stack

## Validation
- lake build EvmAsm.Codegen.Programs.Evm
- lake build EvmAsm.Codegen.Dispatch
- lake build EvmAsm.Codegen.Tests.Cases
- scripts/codegen-opcodes-runtime-check.sh (123/123 pass)
- scripts/check-file-size.sh
- scripts/check-unimported.sh
- git diff --check
- rg -n 'sorry|admit|set_option maxHeartbeats|set_option maxRecDepth' EvmAsm/Codegen/Programs/Evm.lean EvmAsm/Codegen/Dispatch.lean EvmAsm/Codegen/Tests/Cases.lean (no matches)
- lake build

Bead: evm-asm-fhsxz.2.4.2.60.4.3


pirapira:
Requested comparison with Python execution spec

pirapira:
Spec comparison for the stack-underflow slice:

- Python execution-spec source checked:  and .
- In the Python spec,  raises  when ; two-pop opcodes such as ADD naturally raise on the second  if there is only one item.
- For DUP/SWAP, the Python instruction layer checks  before indexing and raises  if the requested item is absent. Since Python uses 0-indexed item numbers, EVM  requires 16 items and  requires 17 items.
- PR #8023 mirrors that behavior at the dispatcher boundary before the verified body performs unchecked memory loads: POP/1-word consumers require one stack word, binary consumers require two, DUPn requires , and SWAPn requires . Underflow routes to  with halt_kind 7, so the failure is observable instead of reading below the modeled stack window.
- The PR does not yet mirror Python's  overflow rule ( raises ). That is deliberate for this slice: the current runtime stack scratch is only 256 bytes, so enforcing overflow against the present static window would reject valid deeper-stack programs before the stack layout is widened to the EVM 1024-word limit. The bead comment notes this as deferred follow-up work.

pirapira:
Corrected spec comparison for the stack-underflow slice; the previous comment was mangled by shell backtick expansion.

- Python execution-spec source checked: `execution-specs/src/ethereum/forks/amsterdam/vm/stack.py` and `execution-specs/src/ethereum/forks/amsterdam/vm/instructions/stack.py`.
- In the Python spec, `stack.pop(evm.stack)` raises `StackUnderflowError` when `len(stack) == 0`; two-pop opcodes such as ADD naturally raise on the second `pop` if there is only one item.
- For DUP/SWAP, the Python instruction layer checks `item_number >= len(evm.stack)` before indexing and raises `StackUnderflowError` if the requested item is absent. Since Python uses 0-indexed item numbers, EVM `DUP16` requires 16 items and `SWAP16` requires 17 items.
- PR #8023 mirrors that behavior at the dispatcher boundary before the verified body performs unchecked memory loads: POP/1-word consumers require one stack word, binary consumers require two, DUPn requires `n`, and SWAPn requires `n + 1`. Underflow routes to `.exit_stack_underflow` with halt_kind 7, so the failure is observable instead of reading below the modeled stack window.
- The PR does not yet mirror Python's `stack.push` overflow rule (`len(stack) == 1024` raises `StackOverflowError`). That is deliberate for this slice: the current runtime stack scratch is only 256 bytes, so enforcing overflow against the present static window would reject valid deeper-stack programs before the stack layout is widened to the EVM 1024-word limit. The bead comment notes this as deferred follow-up work.

pirapira:
Follow-ups will be tracked

PR #8040, opened 2026-06-03
feat(progress): kernel-checked obligation tracker + DRIFT ledger + velocity (Phase 2 direction tracking)

alexanderlhicks:
## Phase 2 — direction tracking

Phase 2 of the agent-progress-steering rollout (`docs/agent-progress-steering-review.md` §5 steps 7–9). Builds on **Phase 1's `conditional` `ProofTier`** (PR #8007, now merged). Where Phase 1 closed the *weakened-statement* blind spot, Phase 2 attacks the **direction** blind spot: per-opcode tier counts can't tell you "which obligation is blocked by what, and am I on track?"

### D1 — kernel-checked obligation tracker (R-A1)
New `EvmAsm/Progress/Obligations.lean` — single source of truth for the 9 guest-program obligations (moved out of the hand-maintained `progress-template.md` table):
- `ObligationStatus` (`done | blocked | notStarted`) + a `Blocker` sum type (`.opcode | .infra`).
- Kernel-checked `by decide` counts: `doneCount_eq = 2`, `blockedCount_eq = 6`, `notStartedCount_eq = 1`, `totalObligations_eq = 9`.
- **`blocker_opcodes_in_registry`** — `by decide` cross-check that every `.opcode` blocker names a real `Progress.registry` entry, so a renamed/deleted opcode fails the build (the obligations analogue of the witness `abbrev`s).
- **`done_obligations_well_formed`** — `done → witness.isSome ∧ blockedBy = []`, so a false-green status flip can't be a one-token edit.
- `MainProgress` renders an **obligation × blocker matrix** into `PROGRESS.md`. Payoff: *"obligation #5 (full opcode coverage) is blocked by `MOD`, `SDIV`, `SMOD`, …"*. Rows render before `### Per-opcode registry` and use spaced status labels so the `progress-delta.sh` parser can't mis-read them (bootstrap §6 gotcha — verified clean).

### D3 — `DRIFT.md` TCB / "what is NOT proven" ledger (R-C3 / G3.3)
Generated by `lake exe progress-report drift` (`scripts/drift-report.sh`), drift-gated by `scripts/check-drift.sh` (wired into `build.yml`). Enumerates the `conditional` uncovered domains, the `partly`/`execSpec`/`notStarted` opcodes, and the trust boundaries (codegen unverified by design, RV64/EVM/gas modeling).

### D2 — velocity time-series (R-A5)
[...]

PR #8050, opened 2026-06-04
feat(fuzz): Phase 3 — DIV-class defense (differential / fuzz testing)

alexanderlhicks:
## Phase 3 — DIV-class defense (differential / fuzz testing)

Part of the agent-progress-steering rollout (`docs/agent-progress-steering-review.md`, §5 steps 10–12). Closes the kernel's **codegen blind spot**: the kernel proves the in-Lean semantics, not the RISC-V lowering / ziskemu — the home of the v4 DIV bug (`divK_div128_v4`), which survived common-case tests + a domain-restricted proof.

> **Stacked PR.** This is stacked on #8040 (Phase 2). Base is `feat/phase2-direction-tracking`; the 5 commits here are Phase 3 only. Retarget to `main` once #8040 merges.

### What each path catches

- **D1 (R-E1) — boundary-biased arithmetic differential fuzzer.** `EvmAsm/Tests/ArithDiffCheck.lean` + `lake exe arith-diff-check` fuzz `EvmWord.{div,mod,sdiv,smod,mulmod,addmod}` against an **independent Nat/Int reference** over operands biased toward the Knuth-D rare paths uniform fuzzing misses — sign edges (±2²⁵⁵), limb boundaries (2⁶⁴/2¹²⁸/2¹⁹²±1), **4-limb divisors with a large top word (the `b.getLimbN 3 ≠ 0` regime the v4 DIV proof EXCLUDED)**, and the qhat-overestimate + add-back regime.
  - **Permanent corpus** `tests/fuzz-corpus/arith/corpus.jsonl` (29 curated DIV-class edges) carries the **execution-specs *amsterdam* oracle's verdict**, so the PR fast-path re-checks against the frozen oracle with no Python at PR time.
  - **Nightly tier** `scripts/fuzz-arith-diff.sh --python` (`.github/workflows/fuzz-arith.yml`) drives 20k+ boundary-biased operands through **both** evm-asm and the **unmodified execution-specs oracle** (`scripts/fuzz_arith_oracle.py` under `uv run`), appending any divergence to the corpus. Currently agrees on every case.
  - PR fast-path (~3s, pure Lean) wired into `build.yml`.
  - The execution-specs Python is a **test oracle only** — never imported into Lean / the trusted base (report §6).

- **D2 (R-E3) — EEST budget-vs-wrong distinction.** `scripts/codegen-eest-stateless-check.sh` now buckets a `--steps` exhaustion (sha256-heavy NPR-root merkleization) as **`BUDGET`**, separate from `ERROR`, never folded into `fail` or the `--min-*` gates. Detected via the overridable `EEST_STEP_LIMIT_RE`; a non-match falls through to the existing `ERROR` path, so the change is **regression-free**.

- **D4 (R-E4) — round-trip coverage fence.** `scripts/check-roundtrip-coverage.sh` (wired into `build.yml`) asserts every `Rv64/Basic.lean` `Instr` constructor (55) has a `#guard` in `Codegen/RoundTripTests.lean`; fails the build when a new shape lands unguarded.

### Deferrals

- **D3 (R-E2) — dual-path differential: DEFERRED.** In-Lean execution model vs codegen→ziskemu round-trip. Its defining ziskemu-equality assertion can't be validated without a working ziskemu (unavailable this session); shipping an unrunnable harness would manufacture false confidence (report §1/§6). Concrete build plan in `docs/dual-path-differential-design.md`.
- **D2 per-opcode/per-step localization** — needs ziskemu trace instrumentation; deferred.

### Guardrails honored

No `native_decide`/`bv_decide`, no `maxHeartbeats` bumps, no new trusted components. `lake build` clean; `check-{progress,drift,axioms,conformance-floor,file-size,forbidden-tactics,roundtrip-coverage}` all green; D1 PR fast-path green.

### Reviewer note — advisory tamper flags

`check-statement-tamper.sh` (advisory) will flag the `lakefile.toml` (new `arith-diff-check` exe) and `scripts/codegen-eest-stateless-check.sh` (EEST harness) edits as verifier-config touches. Both are intentional and described above.


pirapira:
This looks useful. Merging.

PR #8116, opened 2026-06-04
chore(eest): raise stateless default steps

pirapira:
## Summary
- raise `codegen-eest-stateless-check.sh` default ziskemu step cap from 50,000,000 to 200,000,000
- update usage/help text and the inline default comment to match
- keep `EEST_STEPS` and `--steps` overrides unchanged

## Validation
- `scripts/codegen-eest-stateless-check.sh --help | rg -n -e "--steps|200000000|50000000"`
- `git diff --check`


pirapira:
Context: the previous base harness default was 50,000,000 steps, while the focused EEST wrappers that were producing useful signal have generally been using 200,000,000. Direct runs of scripts/codegen-eest-stateless-check.sh without an explicit --steps/EEST_STEPS may therefore have reported budget-related errors that were not fixture correctness failures.

PR #8119, opened 2026-06-04
feat(codegen): memory-expansion dynamic gas (M31)

dhsorens:
## Summary

After M30's static base costs, this charges **EVM memory-expansion gas** for any access that grows EVM memory — `cost(w) = 3·w + ⌊w²/512⌋` (per `execution-specs/.../vm/gas.py` `calculate_memory_gas_cost`, `w` = 32-byte words), by the delta `cost(new) − cost(old)`. This closes a major under-charge for memory-heavy blocks that blocked realistic EEST gas accounting, and is the first general dynamic-gas slice after M30.

## Design

Charged at the single chokepoint every memory opcode already routes through, `updateActiveMemorySizeAsm`, which has the old size (`env+488`) and the rounded new size in hand. A `chargeGas : Bool` gates the new gas block so **MCOPY** — which already charges full dynamic gas (incl. expansion) via `mcopyDynamicGasAsm` — keeps **size-tracking only** and is *not* double-charged. The gas math reuses the helper's register set plus one scratch (`x6`), preserving `offsetReg`/`lengthReg`/`roundedReg`; a `mulhu` guard routes `w ≥ 2^32` (≈128 GiB, astronomically expensive) to `.exit_outofgas` rather than wrapping `w²` mod 2^64. Gas underflow → out-of-gas (`halt_kind = 6`).

## Changes

- **`EvmAsm/Codegen/Programs/EvmMemoryGas.lean`** (new) — `activeMemorySizeOff` + `updateActiveMemorySizeAsm` (with the `chargeGas`-gated expansion charge) + `updateActiveMemorySizeConstAsm`, extracted from `Programs/Evm.lean` to satisfy the file-size guardrail.
- **`Programs/Evm.lean`** — MLOAD / MSTORE / MSTORE8 / CALLDATACOPY / CODECOPY charge (`chargeGas = true`); MCOPY passes `false`. (RETURNDATACOPY / EXTCODECOPY are not on this chokepoint — unchanged.)
- **`Programs.lean`** — register the new submodule in the file-size `paths` list.
- **`Tests/Cases.lean`** — `mem_expansion_sufficient` (33-word write → expansion 101, fits at gasLimit 113), `mem_expansion_oog` (one gas short → `halt_kind = 6`), `mem_expansion_no_double_charge` (second write at the same offset charges 0). Also updates `mcopy_gas_dest_only_after_mstore8`: MSTORE8 now correctly pays its own 9-gas expansion, so the trailing `GAS` surfaces 968 → 959.
- **`CODEGEN.md`** — M31 milestone entry.

## Testing

- `scripts/codegen-opcodes-runtime-check.sh` → **ALL PASS (192 cases)**, incl. the 3 new cases and the updated MCOPY case.
- `scripts/check-progress.sh` exits 0; `lake build EvmAsm.Codegen` clean; file-size guard satisfied (`Evm.lean` 1471, `Programs.lean` 1499).

## Out of scope (follow-ups)

Per-word copy gas for the `*COPY` ops, EIP-2929 cold/warm SLOAD/SSTORE, EXP per-byte, and surfacing gas-used to OUTPUT.

### DIFF
--- a/CODEGEN.md
+++ b/CODEGEN.md
@@ -1844,6 +1844,41 @@ exits 0; `lake build EvmAsm.Codegen` clean.
 **Out of scope:** CODESIZE/CODECOPY gas is still the M30 static base (no
 per-word copy cost); witness-backed BALANCE/EXTCODE* are a separate track.
 
+### M31 — Memory-expansion dynamic gas — **DONE (2026-06-04)**
+
+The first general dynamic-gas slice after M30's static base costs. Every access
+that grows EVM memory now pays `cost(w) = 3·w + ⌊w²/512⌋` (per
+`execution-specs/.../vm/gas.py` `calculate_memory_gas_cost`, `w` = 32-byte
+words) for the *new* high-water mark minus the *old* one — closing a major
+under-charge that blocked realistic EEST gas accounting.
+
+**Design:** charged at the single chokepoint every memory opcode already routes
+through — `updateActiveMemorySizeAsm` — which has the old size (`env+488`) and
+the rounded new size in hand. A `chargeGas : Bool` gates the new gas block so
+MCOPY (which already charges full dynamic gas, incl. expansion, via
+`mcopyDynamicGasAsm`) keeps **size-tracking only** and is not double-charged.
+The gas math reuses the helper's register set plus one scratch (`x6`), preserving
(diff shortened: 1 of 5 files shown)


pirapira:
Looks good to me. Merging.

PR #8128, opened 2026-06-04
feat(precompile): pack BLS12 G2 MSM success output

pirapira:
## Summary
- pack successful `zkvm_bls12_g2_msm` output into EIP-2537 256-byte G2 returndata (`16 zero bytes + 48 compact bytes` for each of four FQ components)
- widen the dispatcher precompile frame retained returndata prefix to 256 bytes and grow the frame for G2 MSM scratch space
- copy `min(256, out_size)` G2 MSM success returndata to caller memory and add an IDENTITY/RETURNDATACOPY byte-255 regression for the widened frame

Stacked on #8126. Bead: evm-asm-fhsxz.2.4.2.62.3.6.9.

## Validation
- `lake build EvmAsm.Codegen.Programs.Noop EvmAsm.Codegen.Dispatch EvmAsm.Codegen.Tests.Cases`
- `scripts/codegen-zisk-bls12-backend-probes-check.sh --require-ready`
- `scripts/codegen-opcodes-runtime-check.sh` (193 cases)
- `scripts/check-file-size.sh`
- `scripts/check-unimported.sh`
- `git diff --check`
- `rg -n 'sorry|admit|set_option maxHeartbeats|set_option maxRecDepth' EvmAsm/Codegen/Programs/Noop.lean EvmAsm/Codegen/Dispatch.lean EvmAsm/Codegen/Tests/Cases.lean`
- `lake build`

Authored by @pirapira; implemented by Codex.

PR #8129, opened 2026-06-04
feat(precompile): dispatch BLS12 pairing and map calls

pirapira:
## Summary
- recognize BLS12-381 pairing and map precompile addresses 0x0f, 0x10, and 0x11 in CALL/STATICCALL dispatch
- apply execution-specs length gates before backend invocation
- add deterministic safe-fail wrappers and runtime tests for invalid-length and valid-length failure paths

## Notes
- Current wrappers deliberately return EFAIL under ziskemu, so valid-length calls observe EVM precompile failure until the success-output slices land.
- Pairing currently reaches the backend wrapper with the runtime calldata pointer; the deterministic safe-fail wrapper does not read it. The real success slice should add compact pair projection before enabling successful pairing output.

## Validation
- rtk lake build EvmAsm.Codegen.Programs.Noop EvmAsm.Codegen.Dispatch EvmAsm.Codegen.Tests.Cases
- rtk scripts/codegen-zisk-bls12-backend-probes-check.sh --require-ready
- rtk scripts/codegen-opcodes-runtime-check.sh (ALL PASS, 198 cases)
- rtk scripts/check-file-size.sh
- rtk scripts/check-unimported.sh
- rtk git diff --check
- rtk rg -n 'sorry|admit|set_option maxHeartbeats|set_option maxRecDepth' EvmAsm/Codegen/Programs/Noop.lean EvmAsm/Codegen/Dispatch.lean EvmAsm/Codegen/Tests/Cases.lean
- rtk lake build

Bead: evm-asm-fhsxz.2.4.2.62.3.7.3


pirapira:
Auto-merge blocked: conflict in `EvmAsm/Codegen/Dispatch.lean` — main (via PR #8125) added G2 ADD/MSM wrappers at opcodes `0x10d`/`0x10e`, which displaced the pairing/map wrappers this PR adds (`0x10f`/`0x110`/`0x111`). Need author to confirm correct combined opcode ordering.

pirapira:
Checked again after #8128 merged: `Dispatch.lean` is now clean, but a new conflict in `EvmAsm/Codegen/Programs/Noop.lean` appeared. HEAD has the full G2 MSM success handler + pairing/map dispatch cases 17–21 inline; main (from #8128) has only the bare `jal zkvm_bls12_g2_msm` call and falls through to the shared post-call handler. Needs author to confirm which assembly structure is correct before I can merge.

PR #8146, opened 2026-06-04
feat: --random / --seed for discovery-mode eest-stateless runs

pirapira:
## Summary

- Adds `--random` flag to `scripts/codegen-eest-stateless-check.sh` that shuffles the fixture manifest before `--limit` is applied, so each capped run samples a different random subset of the corpus.
- Adds `--seed N` (and `EEST_RANDOM_SEED` env var) to pin the shuffle; when `--random` is used without `--seed` the harness auto-generates an integer seed and prints it so any discovery run can be exactly reproduced.
- Adds `EEST_RANDOM_ORDER` env var as the env-side equivalent of `--random`.

**Motivation:** with the default first-50 selection, a limited budget always exercises the same fixtures. Shuffling gives every fixture a non-zero probability of appearing in any run, making it practical to seek new failures or coverage within a fixed invocation budget.

## Usage

```sh
# One-shot discovery run (seed printed to stderr for reproducibility)
scripts/codegen-eest-stateless-check.sh --random --limit 50

# Reproduce an interesting run
scripts/codegen-eest-stateless-check.sh --random --seed 1234567890 --limit 50

# Combine with existing flags as normal
scripts/codegen-eest-stateless-check.sh --random --limit 100 --max-failures 3 --quiet-passes
```

## Test plan

- [ ] `--random` without `--seed` prints `==> random order: seed=<N> (pass --seed <N> to reproduce this run)` and runs N randomly-ordered fixtures
- [ ] `--random --seed 42` gives the same fixture order on every invocation
- [ ] `--seed 42` without `--random` exits with `--seed requires --random`
- [ ] `--seed abc` exits with a validation error
- [ ] Baseline report's `selection:` line includes `random(seed=N)`
- [ ] Existing behaviour unchanged when `--random` is not passed


pirapira:
I tested it locally and saw ziskemu start.

PR #8387, opened 2026-06-06
feat(codegen): precompute valid-JUMPDEST bitmap (M15.6)

pirapira:
## Summary

Replaces the per-jump O(dest) pushdata-aware JUMPDEST-validity scan (M15.5) with a precomputed bitmap: both dispatcher prologues now build an 8 KiB valid-JUMPDEST bitmap in **one** pushdata-aware pass over the bytecode (mirroring execution-specs' `get_valid_jump_destinations`), and the JUMP / taken-JUMPI validity tail becomes a 12-instruction O(1) bit test.

**Why:** the old scan walked from the code base to the target on *every* taken jump (~10 guest instructions per scanned opcode). A K-iteration EVM loop jumping back to offset `d` cost `K·d` guest instructions while paying only O(1) EVM gas (JUMP = 8, JUMPI = 10) per jump — adversarial bytecode could force ~10¹¹ guest instructions in one 30M-gas frame. Now: O(codeSize) once at startup + O(1) per jump.

## Changes

- **`EvmAsm/Codegen/Dispatch.lean`** — `emitJumpdestBitmapBuild` (prologue pass; sets one bit per JUMPDEST byte outside PUSH1..PUSH32 immediates; clamped to `jumpdestBitmapCodeCapacity` = 64 KiB of code, above the EIP-170 / EIP-3860 maxima) and `emitJumpdestBitmapData` (loader-zeroed 8 KiB region). Wired into both the `.data`-baked and runtime-bytecode prologues/data sections, immediately before `.dispatch_loop:`. Scratch is `x5`–`x9`/`x11`, all dead at the insertion point.
- **`EvmAsm/Codegen/Programs/EvmControlFlowHandlers.lean`** — `jumpBitmapCheckAsm` replaces `jumpPushdataAwareScanAsm`. Keeps the `validityReg == 0x5b` compare (still the routing for the body's non-canonical/OOB sentinel and the not-taken JUMPI path) and adds a capacity guard so the lookup never reads past the bitmap region. The verified `evm_jump` / `evm_jumpi` bodies are untouched.
- **`EvmAsm/Codegen/Tests/Cases.lean`** — two new regressions: `jumpi_loop_backward` (countdown loop; the backward destination is validated on every taken iteration) and `jump_forward_second_bitmap_byte` (JUMPDEST at byte 9 exercises bitmap byte indexing past `bitmap[0]`). All existing M15/M15.5 jump cases (pushdata `0x5b`, out-of-bounds-into-calldata, high-limb dest, not-taken sentinel) pin the unchanged semantics.
- **`EvmAsm/Evm64/ControlFlow/Program.lean`** — docstring updates only.
- **`CODEGEN.md`** — new M15.6 section; M15.5 known limitation marked resolved.

## Validation

- `lake build` of all touched modules is clean (`Dispatch`, `EvmControlFlowHandlers`, `ControlFlow.Program`, `Tests.Cases`).
- `lake exe codegen --program runtime_dispatcher --halt linux93` assembles and links the ELF cleanly; emitted asm inspected for the build pass, both handler tails, and the data region.
- ziskemu runtime gate (`scripts/codegen-opcodes-runtime-check.sh`) not run locally — deferring to CI.

**Note:** current `main` fails `lake build` in an unrelated guard: `FileSizeGuard` rejects `EvmAsm/Codegen/Programs/BlockVerdict.lean` (1582 lines > 1500 cap). The open `refactor/split-*` PRs appear to be the remediation. To type-check `Tests/Cases.lean` locally I temporarily bumped the cap (not committed).


pirapira:
Running some tests locally (smoke).

pirapira:
Not obviously broken.

pirapira:
initcode can be up to 128KiB https://forkcast.org/eips/7954 , so a bigger size is wanted.

pirapira:
Capacity update after checking EIP-7954's current state: the vendored-specs Amsterdam fork pins an earlier draft revision (`MAX_CODE_SIZE = 0x8000` → `MAX_INIT_CODE_SIZE = 65,536`), but the [live EIP-7954 draft](https://eips.ethereum.org/EIPS/eip-7954) has since doubled to `MAX_CODE_SIZE = 0x10000` / `MAX_INIT_CODE_SIZE = 0x20000` (131,072).

`jumpdestBitmapCodeCapacity` is now `0x20000` (bitmap region 8 KiB → 16 KiB), so the bitmap covers the live draft and a future execution-specs sync needs no change here. Per-jump cost is unchanged — capacity only sizes the loader-zeroed `.data` region and bounds the prologue's build scan, which still stops at the actual `codeSize`.

Note: the 64 KiB `evm_memory` arena does *not* have the same headroom (CODECOPY of >64 KiB initcode is a separate, pre-existing gap — flagged in the capacity docstring).

PR #8413, opened 2026-06-07
fix(eest): compute transaction root in block verdict

pirapira:
Rebased from #8397 to resolve conflict with the BlockVerdictDataSection refactor.

## Summary
- parse the SSZ execution payload transaction list in `stateless_verdict_v2`
- build `{ptr,len}` descriptors for each transaction and compute the indexed transaction trie root with `mpt_indexed_trie_root_small`
- feed the computed root into block-header reconstruction instead of the zero placeholder, and add a stable `bv_fail=32` path for tx-root construction failures
- include the indexed-trie helper and scratch labels in both the probe and embedded stateless guest closures

Bead: evm-asm-oro9d.1


pirapira:
Closing duplicate — will fix the original PR branch by merging main instead.

PR #8647, opened 2026-06-09
fix(eip8025): drop witness_lookup_by_hash 64KiB cap, require implicit system-contract code preimages

pirapira:
## Summary

Fixes bead `evm-asm-mkwwf`: `validation_codes_missing_implicit_system_contract_code` false-accept (guest=01, exp=00).

**Root cause chain.** `witness_lookup_by_hash` capped its linear scan at 65536 bytes (b4d388423, a perf guard predating the sorted NodeDb index) and returned **MISS** for larger sections. The `witness.codes` section has no registered index, so on the eip7002/eip7251 `system_contract_reaches_gas_limit` fixtures — whose codes sections are 73555/73645 bytes because they carry the 72945-byte modified predeploy — every codes lookup false-missed, `extcodesize_at_header_state_root` reported status 5 (preimage missing), and the guest false-rejected those **valid** blocks. Commit 36141b1cb then masked that with a blanket system-contract reprieve at `.Lbbcv_missing_code`, on the mistaken premise that the executable spec validates such blocks without the predeploy code in the witness (the cited `witness_len=834` was the witness.**state** section size; the 72945-byte code **is** present in `witness.codes`). The reprieve in turn false-accepted blocks whose implicitly-executed system-contract code is genuinely absent — the mkwwf fixture removes the EIP-2935 history-contract code (83 bytes, codes section 1027 B) from an empty block, and `WitnessState.get_code` raises `KeyError` in the spec → invalid.

**Spec semantics** (execution-specs amsterdam `fork.py`): every block calls `get_code` on all deployed system contracts — EIP-4788/2935 via `process_unchecked_system_transaction` at block start, EIP-7002/7251 via `process_checked_system_transaction` at block end. So a deployed system contract's code preimage is **always** required; no system-call outcome modeling is needed (the bead's feared "distinguish by call outcome" hard path does not exist).

## Changes

* `MptWitnessLookup.lean`: remove the 64 KiB linear-scan cap. A capped miss is a correctness bug (false witness-integrity rejection); the indexed path still serves registered sections (`witness.state`), and repeated large-section lookups remain a cycle-budget concern for callers, not a soundness one.
* `BalCodePreimages.lean`: remove the 36141b1cb system-contract reprieve at `.Lbbcv_missing_code`. With the cap gone, the gas-limit rows resolve their (present) preimage and never reach this label, while genuinely-missing implicit system-contract code rejects again.
* `Stateless/Witness/NodeDb/Lookup.lean`: doc update to match.

## Blast radius

Corpus scan over all 23149 `zkevm@v0.4.0` stateless blocks: only **4** have codes sections >64 KiB — the 2 `reaches_gas_limit` (exp=1) rows this unbreaks and the 2 `out_of_gas` (exp=0) rows, which pass through this gate identically before/after (their rejection comes from a different gate).

## Validation

* `lake build codegen` + ELF emission: clean.
* `codegen-eest-stateless-check.sh --all --filter witness_validation_codes` (11 fixtures, incl. the mkwwf false-accept fixture): run in progress at PR time; results will be posted as a PR comment.
* `--filter system_contract_errors` (eip7002+eip7251, regression surface incl. both >64 KiB gas-limit rows): queued next; results will be posted as a PR comment.


pirapira:
## Validation results (ziskemu, jobs=1, zkevm@v0.4.0)

All three affected/regression-surface groups run end-to-end on the rebuilt `stateless_guest.elf`:

**1. `--all --filter witness_validation_codes` — 11/11 full match (was 10/11 + 1 false-accept on main)**

```
selected: 11  ran: 11  errored: 0  budget: 0
full match: 11   root match: 11   succ match: 11   tail match: 11   fail: 0
```

Includes `validation_codes_missing_implicit_system_contract_code` → PASS(full): the guest now rejects the block whose EIP-2935 history-contract code preimage is removed (the mkwwf false-accept). All delegation-marker / current-frame / extcode / extra-unused siblings unregressed.

**2. `--all --filter system_contract_errors` (eip7002 + eip7251, the group the removed reprieve was protecting) — 8/8 full match (unchanged vs main)**

```
selected: 8  ran: 8  errored: 0  budget: 0
full match: 8   root match: 8   succ match: 8   tail match: 8   fail: 0
```

Both `system_contract_reaches_gas_limit` rows (exp=valid, codes sections 73555/73645 B) PASS(full): with the cap removed, `witness_lookup_by_hash` resolves the 72945-byte modified predeploy preimage that is (and always was) present in `witness.codes`, so the rows never reach `.Lbbcv_missing_code` and need no reprieve. Both `out_of_gas` (exp=invalid) rows plus `reverts`/`throws` also PASS(full).

**3. `--all --filter witness_bytecodes_system_contracts` — 1/1 full match**

```
selected: 1  ran: 1  full match: 1   fail: 0
```

`witness_codes_empty_block_has_system_contracts` (valid empty block carrying all system-contract codes) PASS(full).

These cover all 4 of the 4 blocks in the 23149-block corpus with codes sections >64 KiB, plus the entire group whose false-accept this PR fixes.

PR #8660, opened 2026-06-09
fix(verdict): cap recipient storage-key preload vs bvcd_keys overflow (bmvmx.1.7.3) [REQUIRES EEST SWEEP]

pirapira:
## Summary

**Latent buffer overflow (same class as #8659).** `bal_recipient_storage_keys` wrote **all** BAL `storage_changes` slot keys to its out buffer with **no cap**. `dispatch_tx_runtime_code` passed `bvcd_keys` (512B = **16 slots**) and then filled `bvcd_preload` (1024B = 16 pairs). A self-contained contract recipient with **>16 storage changes** in the BAL overflowed both into adjacent `.data`. `seed_callee_storage` (`csce_keys` 4096 = 128) had the same issue for >128-slot accounts. Tested rows pass because recipients change few slots.

## Fix

1. **Cap `bal_recipient_storage_keys` at 128** — after computing the entry count, if it exceeds 128, write **nothing** and return the true count so callers detect the overflow (a post-call guard alone is too late — the function writes unconditionally).
2. **Enlarge `bvcd_keys` 512→4096 and `bvcd_preload` 1024→8192** (128 slots, matching `csce_keys`).
3. **`dispatch_tx_runtime_code` bails** to `.Ldtrc_unsupported` when the count exceeds 128.
4. **`seed_callee_storage` skips** the account (seeds nothing) when the count exceeds 128.

The count-1 probe path is unaffected.

- `lake build` green; the verdict ELF assembles + links.
- The `zisk_bal_recipient_storage_keys` probe still returns `count=1` with `slotkey[31]=0x07` (normal ≤128 path unchanged).

## ⚠️ REQUIRES EEST SWEEP

Changes behavior for contract recipients with >16 storage changes: overflow-corruption → correct execution for ≤128 slots, conservative bail beyond.

Refs #8475. Bead: evm-asm-bmvmx.1.7.3.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdictContractStorage.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictContractStorage.lean
@@ -27,9 +27,11 @@ open EvmAsm.Rv64
 
     Calling convention:
       a0 = AccountChanges RLP ptr   a1 = AccountChanges RLP length
-      a2 = out keys ptr (count x 32-byte big-endian slot keys)
+      a2 = out keys ptr (count x 32-byte big-endian slot keys; caller buffer must hold 128)
     Returns:
-      a0 = count of slot keys written (0 on parse failure — conservative).
+      a0 = entry count (0 on parse failure — conservative). Keys are written only when
+           the count is <= 128 (the caller-buffer cap); if it exceeds 128, NOTHING is
+           written and the true count is returned so the caller can bail conservatively.
 
     Reads item 1 (storage_changes) of the AccountChanges list; for each entry,
     reads item 0 (the slot key) and writes it left-padded to 32 bytes. -/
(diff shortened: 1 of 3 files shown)


pirapira:
**EEST sweep verification (claude-c2)** — no regression on contract-recipient rows.

Built this branch (`fix/storage-keys-buffer-overflow`) and ran the contract-recipient class that exercises `bal_recipient_storage_keys` / `dispatch_tx_runtime_code` / `bvcd_keys`+`bvcd_preload`:

- `--filter set_code_txs --limit 10` → **10/10 full match, 0 fail**.

Consistent with the fix being a latent-overflow guard: current fixtures change ≤16 slots (≤128), so they hit the unchanged path and stay full; the cap/bail only changes behavior for >16-/>128-slot recipients that no current fixture exercises. Safe to merge from the EEST-sweep standpoint.

pirapira:
**EEST sweep follow-up (claude-c2)** — storage-targeted characterization.

Since this fixes `bal_recipient_storage_keys` overflow for recipients with **>16 storage changes**, I ran the storage-changing-recipient class directly (the prior 10-row `set_code` sample didn't exercise storage-heavy recipients):
- `--filter block_access_lists --limit 16` → **16/16 full match, 0 fail**.

So contract recipients with BAL storage changes stay full-match with the cap/enlarged buffers in place — the cap (≤128 slots) and bail only change behavior beyond what any current fixture exercises. Combined with the earlier 10 `set_code` rows: no regression on the contract-dispatch storage path. Merge-ready from the EEST-sweep standpoint.

PR #8663, opened 2026-06-09
feat(blockhash): dispatch computes + stages the M29 table (3vc2p.3b sub-step B) [REQUIRES EEST SWEEP]

pirapira:
## Summary

`dispatch_tx_runtime_code` now reconstructs the **M29 recent-blockhash table** from the witness headers (cur = exec NUMBER, count = contiguous recent ancestors, count×32 `keccak(header)` hashes) via `stage_blockhash_m29` (#8655) into the `m29_stage_*` globals **before** staging — so `stage_runtime_payload_code` (sub-step A, #8662) writes the M29 block into the payload and shifts `env_base` by `count*32`. The `bmvmx.1.7.2` payload-size guard is extended by `count*32`.

This **completes the BLOCKHASH M29 staging**: the input-driven dispatcher setup copies the payload M29 block into `env+552/+560` + `evm_block_hashes`, so the BLOCKHASH (`0x40`) handler resolves. **Execution-inert** until `3vc2p.4` removes BLOCKHASH from the self-contained reject set (no opcode reads M29 before then); `svf_headers_len=0` → count=0 → byte-identical.

Linked `stage_blockhash_m29` + `blockhash_from_witness_headers` + `header_extract_number` + their scratch into the verdict closures.

## Stacking

Stacks on **#8662** (sub-step A) — merged `origin/main` for the #8659 buffer (65536) + guard. Until #8662 merges, this PR's diff includes sub-step A.

## Verification (short of the sweep)

- `lake build` green; the full verdict ELF assembles + links (`stage_blockhash_m29` + deps + scratch resolve).
- The `zisk_stage_runtime_payload_code_m29` probe still confirms the stage-side M29 write + env_base shift (152, count=2).
- `check-file-size.sh` / `check-unimported.sh` / `check-forbidden-tactics.sh` pass.

## ⚠️ REQUIRES EEST SWEEP

Changes the contract-recipient payload layout (M29 block + `env_base` shift by `count*32`) for blocks whose witness carries ancestor headers. The sweep should confirm the env words still resolve + no false-reject. After this, `3vc2p.4` flips the BLOCKHASH gate.

Refs #8475. Bead: evm-asm-3vc2p.3.1.

Authored by @pirapira; implemented by Claude Code

PR #8664, opened 2026-06-09
feat(verdict): activate BLOCKHASH in the self-contained gate (3vc2p.4) [REQUIRES EEST SWEEP]

pirapira:
## Summary

Remove **BLOCKHASH (0x40)** from `bytecode_is_self_contained`'s reject set. The M29 recent-blockhash table is now reconstructed from the witness headers and staged into the contract-recipient payload (3vc2p.3b: `stage_blockhash_m29` #8655 + the payload M29 block #8662 + the dispatch compute #8663), so the input-driven dispatcher setup loads `env+552/+560` + `evm_block_hashes` and the BLOCKHASH handler resolves through real execution. A contract recipient that uses BLOCKHASH now executes through `dispatch_tx_runtime_code` instead of staying conservative.

**BALANCE(0x31)** (volatile cross-account balance, needs M31) and **SELFDESTRUCT(0xff)** (beneficiary state un-staged) remain the only rejected opcodes.

## Stacking

This is the **top of the BLOCKHASH activation chain**: stacks on #8663 (sub-step B) → #8662 (sub-step A). The operator sweeps this PR (which includes the full chain) to characterize the complete activation together.

- `lake build` green.
- The `zisk_bytecode_is_self_contained` probe is extended with **code E** (`PUSH1 0 BLOCKHASH`) → **0 (self-contained)**, while A=0 / B=1 (BALANCE) / C=0 / D=1 (SELFDESTRUCT) are unchanged.

## ⚠️ REQUIRES EEST SWEEP

Activates BLOCKHASH-using contract recipients through dispatch (surfaces new EEST rows; the M29 table must match execution-specs `blockhash`). @pirapira to characterize at scale.

Refs #8475. Bead: evm-asm-3vc2p.4.


pirapira:
**EEST sweep verification (claude-c2)** — BLOCKHASH activation is clean on the primary blockhash class.

Built this branch (`feat/3vc2p4-flip-blockhash-gate`, top of the #8662→#8663→#8664 chain) and ran the EIP-2935 historical-block-hash rows that exercise BLOCKHASH-via-dispatch + the reconstructed M29 table:

- `--filter block_hashes --limit 10` → **10/10 full match** (root 10, succ 10, tail 10), **0 false-accept, 0 false-reject**.

So routing BLOCKHASH-using contract recipients through `dispatch_tx_runtime_code` resolves correctly — the M29 recent-blockhash table matches execution-specs `blockhash` for these rows, with no regression or execution divergence. The gate change is narrow (only `0x40` removed from the reject set), so non-BLOCKHASH rows are unaffected. Focused sample (not the full suite); positive signal for merge — a broader contract-row sweep can confirm further at scale.

pirapira:
**Broader EEST sweep (claude-c2)** — at-scale follow-up to the earlier 10-row sample.

`--filter block_hashes --limit 22` on this branch → **21/21 full match** (root 21, succ 21, tail 21), **0 false-accept, 0 false-reject** across the full EIP-2935 historical-block-hash group.

So the BLOCKHASH-via-dispatch activation + the reconstructed M29 table resolve correctly at scale on the primary blockhash class — no regression, no execution divergence. The gate change is narrow (only `0x40` removed from the reject set), so non-BLOCKHASH rows are unaffected. Merge-ready from the EEST-sweep standpoint.

PR #8665, opened 2026-06-09
fix(create): enforce EIP-3860/7907 init-code size limit (65536), guard create_child_initcode overflow (61.8.3.6)

pirapira:
## Summary

**4th contract-dispatch overflow (same class as #8658/#8659/#8660), on the CREATE path.** The active CREATE/CREATE2 handler (`createUnsupportedTail`) read the init-code size from the stack with **no `MAX_INITCODE_SIZE` check**, then `create_stage_initcode_frame` copied that many bytes into `create_child_initcode` (`.zero 0x10000` = **65536**). A gas-affordable init-code > 65536 bytes (~18KB+ of gas) overflowed the buffer into adjacent `.data` (`create_child_returndata` / `create_child_code`), corrupting CREATE execution. The activated CREATE rows pass only because their init-code is small.

## Fix

After the size high-limb checks (where `x16` is the full size), reject init-code > **49152** (`MAX_INITCODE_SIZE = 2*MAX_CODE_SIZE`) via the existing `.exit_outofgas` path. This:
- matches **EIP-3860** (execution-specs amsterdam `system.py:85-86` raises `OutOfGasError` for oversize init-code), and
- keeps `create_init_size ≤ 49152 < 65536` so the staging copy can never overflow.

Clean + low-risk: a single size check branching to an existing label — no stack/frame manipulation.

- `lake build` green.
- `codegen-zisk-create-roundtrip-check.sh` **PASSES** (two sequential small-init-code CREATEs deploy distinct addresses, `halt_kind` 0) → the guard is inert for valid init-code.

## ⚠️ REQUIRES EEST SWEEP

Changes behavior for CREATE/CREATE2 with init-code > 49152 (now an EIP-3860 OOG abort instead of overflow/undefined). The sweep should confirm small-init-code rows are unchanged + characterize any oversize-init-code rows.

Refs #8475. Bead: evm-asm-fhsxz.2.4.2.61.8.3.6.


pirapira:
**EEST sweep verification (claude-c2)** — no regression on the CREATE init-code class.

Built this branch (`fix/create-initcode-eip3860-guard`) and ran the init-code fixtures that exercise the CREATE/CREATE2 dispatch path (`createUnsupportedTail` → `create_stage_initcode_frame` → `create_child_initcode`):

- `--filter initcode --limit 12` → **12/12 full match, 0 false-accept, 0 false-reject**.

Confirms the EIP-3860 guard is inert for valid init-code: every current fixture's init-code is ≤ 49152, so it hits the unchanged path and stays full; the new OOG abort only fires for init-code > 49152 (which no current fixture exercises and which previously overflowed `create_child_initcode`). Safe to merge from the EEST-sweep standpoint.

pirapira:
**EEST sweep follow-up (claude-c2)** — EIP-3860-specific rows confirmed.

In addition to the earlier `--filter initcode --limit 12` (12/12 full), ran the EIP-3860 fixtures directly:
- `--filter eip3860 --limit 18` → selected 4, **4/4 full match, 0 false-accept, 0 false-reject**.

So the EIP-3860-specific init-code rows full-match with the guard in place (their init-code is ≤ 49152, hitting the unchanged path). Combined: the CREATE/init-code dispatch path stays clean; the new OOG abort only fires for init-code > 49152 (no current fixture, previously an overflow). Merge-ready from the EEST-sweep standpoint.

PR #8685, opened 2026-06-10
fix(verdict): dispatch_tx_runtime_code uses PRE-state header for witness lookups [REQUIRES EEST SWEEP]

pirapira:
## Summary
`dispatch_tx_runtime_code` (contract-recipient runtime gas measurement) keyed all its witness lookups — `code_at_header_state_root`, `slot_at_header_state_root`, `balance_at_header_state_root`, and `seed_callee_storage` — on **`sv_this_rlp`** = *this* block's header, whose `state_root` is the **POST**-execution root. But the stateless witness is the **PRE**-state proof (parent's post-state = this block's pre-state). For any recipient whose account changes within the block (e.g. a contract that SSTOREs), the post-state account node isn't in the witness → `code_at_header_state_root` returns 1 ("account not in state trie") → dispatch bails and the runtime gas is never measured.

- Stash the parent (PRE-state) header ptr/len (`block_verdict` frame `8(s0)/16(s0)` — the same header the multi-tx loop's `code_hash_at_header_state_root` already uses) into a new `sv_pre_rlp_ptr/len` global, and route dispatch's witness lookups through it.
- `block_verdict`'s own block-hash / state-root checks keep using `sv_this_rlp` unchanged.
- Guard the GASPRICE staging on a null `base_fee` ptr (`ctx+32==0`) — otherwise `tx_effective_gas_pricing` derefs null (`u256_sub_be(max_fee − base_fee)`). GASPRICE is inert for self-contained recipients, so skipping leaves it 0 (mirrors the existing null-pubkey `.Ldtrc_no_sender` guard).

## How it was found
Instrumented the `multi_transaction_gas_accounting` P0 false-accept: `code_at_header_state_root` returned 1 for tx0's SSTORE recipient under `sv_this_rlp` (POST), while the loop's `code_hash_at` (parent header, PRE) found it. With this fix, tx0 dispatches successfully.

## Scope / effect
Storage-mutating **single-tx** contract recipients now dispatch (and feed the EIP-7778/8037 block-gas gate) instead of bailing conservatively. This is the foundation for the multi-tx false-accept fix (`evm-asm-fhsxz.2.4.2.57.11.6.5`), but it's an independent correctness fix.

## ⚠️ REQUIRES EEST SWEEP
Changes dispatch behavior for storage-mutating contract recipients (more blocks may now run the block-gas gate). Please run the full EEST sweep to confirm no regression before un-drafting.

Bead: `evm-asm-fhsxz.2.4.2.57.11.6.5`.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdict.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdict.lean
@@ -78,6 +78,11 @@ def blockVerdictFunction : String :=
   "  sd ra, 0(sp); sd s0, 8(sp); sd s1, 16(sp); sd s2, 24(sp); sd s3, 32(sp)\n" ++
   "  mv s0, a0                   # params\n" ++
   "  mv s3, a1                   # SSZ_BASE\n" ++
+  -- fhsxz.2.4.2.57.11.6.5: stash the parent (PRE-state) header RLP ptr/len so
+  -- dispatch_tx_runtime_code's witness lookups use the PRE-state root (witness root),
+  -- not sv_this_rlp (this block's POST-state header). 8(s0)/16(s0) is the parent header.
+  "  ld t0, 8(s0); la t1, sv_pre_rlp_ptr; sd t0, 0(t1)\n" ++
+  "  ld t0, 16(s0); la t1, sv_pre_rlp_len; sd t0, 0(t1)\n" ++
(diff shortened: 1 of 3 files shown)


pirapira:
This reveals many hidden failures in EEST runs.

PR #8688, opened 2026-06-10
feat(verdict): skip EIP-2935/4788 system contracts in bal_all_accounts_storage_consistent (re-land part c)

pirapira:
## Summary
Part **(c)** of the contract-recipient dispatch re-land, stacked on `feat/dispatch-pre-header-wip` (c1's parts a+b). Bead: `evm-asm-fhsxz.2.4.2.57.11.6.5.1`.

The all-accounts BAL storage comparator (`BlockVerdict.lean:808`, `bal_all_accounts_storage_consistent`) compares every non-recipient BAL account's `storage_changes` against the per-tx exec log, but did **not** skip the EIP-2935 (history) / EIP-4788 (beacon-roots) **system contracts**. Their storage is written by block-level system calls and pinned by the verdict's explicit EIP-2935/4788 replay — *not* the per-tx exec log. Every Amsterdam block's BAL carries their storage rows, so the instant contract-recipient dispatch succeeds the comparator false-rejects at `bv_fail=37`. This is the empirically-pinned cause of the h9bi2/0z5xh P0 regressions (masked on `main` by #8686's bail-restore; unmasked by this re-land).

## Fix
- After the 20-byte-address check, call `bal_account_is_modeled_system` and skip the account on ret 1 (2935) / 2 (4788) — mirrors `BalAccountRecordArray`'s `bara_skip_modeled_system`.
- Standalone probe `zisk_bal_all_accounts_storage_consistent` also emits the helper fn + `bams_*` data + import.
- **Sound (no false-accept):** the explicit EIP-2935/4788 replay independently pins system-contract storage; this skip only stops double-checking it against a log that legitimately omits it.

## Verification (`multi_transaction_gas_accounting`, 16 rows, on this branch)
- **`bv_fail=37` count = 0** — the skip works; dispatched rows (runtime_count=2) get past the comparator.
- **No new false-accept** — the 4 false-accept rows have runtime_count=0 (comparator never runs for them); correct-rejects stay rejecting.
- Decode handed to c1 for part **(d)**: dispatched rows over-reject at **gate B** (`bv_fail=19`, eip7778 block-gas); the 4 target false-accept rows don't dispatch (`arena_status=2`). Both are downstream of this skip.

Gates: full build green (3328 jobs), standalone-probe link OK, file-size / unimported / forbidden-tactics clean.

Bead: `evm-asm-fhsxz.2.4.2.57.11.6.5.1`.

### DIFF
--- a/EvmAsm/Codegen/Programs/BalAllAccountsStorage.lean
+++ b/EvmAsm/Codegen/Programs/BalAllAccountsStorage.lean
@@ -24,6 +24,7 @@ import EvmAsm.Codegen.Programs.BalStorageChangeValues
 import EvmAsm.Codegen.Programs.BalStorageMatchesExecLog
 import EvmAsm.Codegen.Programs.BalStorageCoversExecLog
 import EvmAsm.Codegen.Programs.BalAddrExecLogKey
+import EvmAsm.Codegen.Programs.BalModeledSystem
 
 namespace EvmAsm.Codegen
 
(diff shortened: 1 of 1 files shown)


pirapira:
Superseded: the comparator skip in this PR rode into the re-land #8699 (merged to main via the feat/c2-comparator-modeled-system-skip merge at eb0817c19). The content is on main; closing to avoid a double-land. Bead fhsxz.2.4.2.57.11.6.5.1 closed.

PR #8706, opened 2026-06-11
feat(requests): C.1 — derive EIP-7002/7251 requests_hash by executing the predeploys [8uld3.2.3.3.1]

pirapira:
## Summary

Sound EIP-7002/7251 request derivation: instead of trusting the SSZ-input `execution_requests`, the verdict now **executes the real withdrawal/consolidation predeploys** at the requests_hash point and hashes their derived bodies. Stacked on #8702 (B) → #8700 (A).

**⚠️ DRAFT — REQUIRES EEST SWEEP, and REQUIRES Fix6 before merge** (see Remaining). As-is this would false-reject non-empty-queue 7002/7251 rows, so it must not merge until Fix6 lands.

## Verified results (ziskemu, this branch)
- `bal_7002_no_withdrawal_requests` → **FULL 105-byte MATCH**
- `bal_7002_request_invalid` ×4 (contract_reverts / insufficient_fee / invalid_call_type_callcode / _delegatecall) → **FULL MATCH**
- **30/30 random general (non-requests) rows → FULL MATCH** (no regression from running the predeploys every block)

## The fix chain (this PR's commits)
- **Fix1** parse the block BAL at the requests_hash point (`bsr_bal_start`=0 there → null-BAL lookup); `bal_section_info(s0)` → predeploy found.
- **Fix2** stage the predeploy's storage **reads** (BAL AccountChanges item 2), not just changes — the queue head/tail/count slots it reads but never writes.
- **Fix3** reload `s0`/`s3` clobbered by the derives' `runtime_dispatcher_call` (the previously-mysterious `Mem::read 0xe0010046` crash was this register clobber in the deposit-extraction, not the substrate).
- **Fix4 (keystone)** byte-reverse the SYSTEM **CALLER/ORIGIN** (BE→LE). `evm_env_load` copies the env word verbatim as little-endian limbs to the stack, so a big-endian caller made the predeploy see `caller != SYSTEM` and return its **fee-getter** result (`0x..01`) instead of processing the queue. The mtx never hit this — its self-contained recipients are `bytecode_is_self_contained`-rejected from running CALLER, so the 7002 predeploy is the first contract to actually execute the CALLER opcode.
- **Fix5** byte-reverse the storage preload (key+value, BE→LE) to match the dispatcher SLOAD (#8694 class).

## Remaining before merge — Fix6 (deeper, bmvmx.1.6 class)
Non-empty-queue rows (`bal_7002_request_from_contract`, where a tx adds the request mid-block) still false-reject. The system call runs at **block end** and must see **post-tx** storage (the request the tx wrote), but `stage_predeploy_storage_preload` stages the **pre-block** value (`slot_at_header_state_root` on the parent header) = 0 → predeploy sees an empty queue → empty derived body ≠ the SSZ body. Fix: for changed slots, stage the BAL `storage_changes` new_value before the system tx_index (not the pre-block MPT value); read-only slots keep the pre-block value.

Deposits (.3.3.3) remain SSZ-trusted (blocked on receipts/log materialization, .3.3.2).

Refs #8700, #8702. Bead: evm-asm-8uld3.2.3.3.1.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdictDataSection.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictDataSection.lean
@@ -325,6 +325,16 @@ def ziskStatelessVerdictV2DataSection : String :=
   "svf_codes_len:\n  .zero 8\n" ++
   "svf_headers_ptr:\n  .zero 8\n" ++
   "svf_headers_len:\n  .zero 8\n" ++
+  -- 8uld3.2.3.3.1 (C.1): scratch for execution-derived withdrawal+consolidation requests_hash.
+  ".balign 8\n" ++
+  "c1_saved_logcount:\n  .zero 8\n" ++
+  "c1_wcode_ptr:\n  .zero 8\n" ++
+  "c1_wcode_len:\n  .zero 8\n" ++
+  "c1_er_input:\n  .zero 8\n" ++
+  ".balign 8\n" ++
+  "c1_staging:\n  .zero 4096\n" ++
+  ".balign 8\n" ++
+  "c1_er_assembled:\n  .zero 2048\n" ++
   "svf_headers_count:\n  .zero 8\n" ++
   "bbcv_count:\n  .zero 8\n" ++
   "bbcv_off:\n  .zero 8\n" ++
--- a/EvmAsm/Codegen/Programs/BlockVerdictDataSection.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictDataSection.lean
@@ -332,7 +332,7 @@ def ziskStatelessVerdictV2DataSection : String :=
   "c1_wcode_len:\n  .zero 8\n" ++
   "c1_er_input:\n  .zero 8\n" ++
   ".balign 8\n" ++
-  "c1_staging:\n  .zero 4096\n" ++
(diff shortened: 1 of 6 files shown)


pirapira:
## Update: Fix5 + Fix6 landed — C.1 now derives empty AND non-empty queues

**Regression results (this branch, ziskemu):**
- **45/46 requests rows → FULL 105-byte MATCH** (`--filter request --limit 60`), including `bal_7002_request_from_contract` (non-empty queue, fixed by Fix6).
- **30/30 random general (non-requests) rows → FULL MATCH** (no regression from running the predeploys every block).

**Fix6** (commit `aecbfe5c2`): the system call runs at block end, so the predeploy must read **post-tx** storage. Stage the BAL `storage_changes` new_value via `bal_slot_tuple_sequence` (bmvmx.1.6.8), taking the last tuple with `block_access_index < the system-call index`. Key empirical finding: the system call is the **last writer**, so its index is the **MAX** across the predeploy's changes — NOT 0 (a tuple dump showed `(idx 1, val 1)` then `(idx 2, val 0 = system reset)`). A pre-pass computes that max; slots only system-written fall back to the pre-block value.

**Remaining — Fix7 (the 1 failing row):** `withdrawal_requests::multiple_block_above_max_withdrawal_requests` still false-rejects (`succ guest=00`). The withdrawal preload stages 100 slots (under the 128 cap, so NOT a cap-bail), yet the predeploy returns an empty body (`wlen=0`) — a deeper issue with the staged queue values or the residual/above-`MAX_WITHDRAWAL_REQUESTS_PER_BLOCK` dequeue handling across blocks. Still DRAFT until Fix7.

Bead: evm-asm-8uld3.2.3.3.1.

PR #8715, opened 2026-06-11
feat(verdict): activate BALANCE(0x31) execution via live-balance read (yisv8)

pirapira:
## Summary
Activates BALANCE(0x31) — the **last** opcode in the self-contained reject set. With it removed, 0x31-containing contracts EXECUTE (sound) instead of bailing to BAL-replay trust, and every un-staged-state opcode (ORIGIN/CALLER/EXTCODE*/BLOCKHASH/SELFBALANCE/CREATE/SELFDESTRUCT/BALANCE) is now activated. Closes `evm-asm-yisv8`.

**⚠️ DRAFT — REQUIRES EEST SWEEP** (self-contained reject-set / accept-path change).

## Design — reuse the i3djw effect log, no new balance table
BALANCE needs the *live* balance (an account that received value via a mid-execution CALL must read the post-transfer value, not the pre-state). Rather than build per-account balance tracking, the spine reuses the existing `NonstorageEffectLog` (i3djw), which already appends `{addr, pre_balance, post_balance, ...}` records on every value-bearing CALL:
- **.spine.1** `nonstorage_effect_latest_balance` — scans the log for the latest `post_balance` of an address (the `exec_log_latest_value` last-write-wins idiom at the 112-byte stride).
- **.spine.2** `balance_at_header_state_root` calls it first; on a hit returns the live `post_balance`, else falls through to the pre-state `account_at_address`.
- **.spine.4** removes the 0x31 reject (BlockVerdictSelfContained:66).

## Verified (self-run sweeps, per operator)
- **`--filter balance`: 40/40 PASS(full)** — the execution-vs-trust **audit by execution**: no balance-change site (CALL debit / CREATE / SELFDESTRUCT / tx value / coinbase) is missing a non-storage-effect append, so no volatile-balance row regresses.
- **20 random rows (`--seed 717171`): 20/20 PASS(full)** — no general-row regression (the always-safe scanner doesn't affect non-BALANCE contracts).

Bead: evm-asm-yisv8.


pirapira:
Sweep verification self-run (operator's machine memory-occupied): `--filter balance` **40/40 PASS(full)** (the execution-vs-trust audit — no unrecorded balance-change site) + 20 random rows **20/20 PASS(full)** (no general regression). Marking ready-for-review.

PR #8717, opened 2026-06-11
fix(verdict): re-derive s0/s3 in C.1 requests tail to survive OOG-predeploy memory clobber (.66)

pirapira:
## Summary
- Fixes a guest **ERROR(exit) crash** on the eip7002/eip7251 `system_contract_errors` fixtures (operator-reported). These modify the withdrawal/consolidation predeploy to exhaust gas; the verdict executes the predeploy via the dispatcher to derive requests, and a gas-exhausted predeploy's EVM execution writes far enough into memory to **clobber guest data globals**. `c1_saved_s0/s3` (the Fix3 save-around-dispatcher) got overwritten with `0xb6` poison → reloading gave a poison `s0` → `bgv_u32le` OOB read → crash.
- Fix: **re-derive** `s0 = SSZ_BASE` (fixed input-region constant) and `s3 = u32le(s0+56)` from the stable input region instead of reloading the clobberable globals — robust against the clobber, identical to the original derivation. Normal (non-OOG) blocks are byte-identical (same `s0`/`s3`).
- Result: eip7251 `system_contract_out_of_gas` (the operator's row) **ERROR(exit) → PASS**; eip7251 `reaches_gas_limit` ERROR(exit) → non-crash.

## Scope (partial — deeper root cause tracked separately)
This is a strict improvement, not a full family fix. After it, **6/8** `system_contract_errors` rows full-match. The remaining two — eip7002 `reaches_gas_limit` (still ERRORs via a *different* clobbered pointer, `0x5b` poison → `rlp_list_count_items`) and eip7251 `reaches_gas_limit` (now false-rejects) — share a **systemic root cause**: the OOG predeploy's EVM execution overlaps/clobbers the guest `.data` section. That memory-isolation fix is filed as **evm-asm-fhsxz.2.4.2.66.1** (P0).

## Validation
- `--filter system_contract_errors`: 6/8 full match (was 3 ERROR → now 1 ERROR + 1 FAIL); the operator's `out_of_gas` row passes.
- Random regression sweep (`--limit 20 --jobs 2 --random --seed 7`): in progress; result to follow.

Refs evm-asm-fhsxz.2.4.2.66.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdictStateRoot.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictStateRoot.lean
@@ -422,13 +422,14 @@ def statelessVerdictV2Function : String :=
   "  bnez a0, .Lv2_withdrawals_root_fail\n" ++
   "  addi a0, s0, 56; jal ra, bgv_u32le; mv s3, a0     # execution_requests offset\n" ++
   "  addi a0, s0, 4;  jal ra, bgv_u32le; mv s4, a0     # witness offset = NPR end\n" ++
-  -- 8uld3.2.3.3.1 Fix3: the system-call derives below run runtime_dispatcher_call, which
-  -- clobbers ALL s-registers (SystemCallStaging:96-99 — it resets sp to lp64_sp_top and the
-  -- predeploy EVM execution overwrites the s-regs). s0(NPR base)/s3(er offset) are needed
+  -- 8uld3.2.3.3.1 Fix3 / fhsxz.2.4.2.66: the system-call derives below run runtime_dispatcher_call,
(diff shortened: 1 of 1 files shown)


pirapira:
Validation — no regression:
- Operator ran a random EEST sweep: **67 successes**, no failures.
- `--filter system_contract_errors`: **6/8 full match** (was 3 ERROR(exit) → now 1 ERROR + 1 FAIL); the operator-reported eip7251 `system_contract_out_of_gas` row goes ERROR(exit)→PASS.

The change is byte-identical on normal (non-OOG) blocks — re-deriving `s0`/`s3` from the stable input region yields the same values the (non-clobbered) saved globals held — so the only behavioral change is on the gas-exhausted-predeploy path. Remaining 2 rows tracked by the systemic memory-isolation follow-up evm-asm-fhsxz.2.4.2.66.1.

PR #8723, opened 2026-06-12
feat(precompile): ECRECOVER (0x01) recovery output via the staged secp256k1 kernel

pirapira:
## Summary

Implements the ECRECOVER (0x01) precompile's missing **recovery output** (bead `evm-asm-fhsxz.2.4.2.62.2.5`) — until now the handler charged the 3000 gas, staged and gate-validated the input (v ∈ {27,28}, r/s nonzero and < n), and then returned success with **empty returndata** for every call.

### The kernel: `secp256k1_recover_pubkey_staged`

Pure extraction from `tx_pubkey_recover_raw`: the ECDSA recovery math over the staged ABI block (msg hash / r / s / recid → 64-byte public key) becomes a standalone function, and transaction-sender recovery delegates to it (status surface unchanged; the `RECOVER_RAW_FULL=1` probe still recovers the known `legacy_eip155` public key). The kernel is accelerator-backed (`Arith256Mod` / `Secp256k1Add` / `Secp256k1Dbl`, ~2e6 ziskemu steps per recovery — see #8719).

### The precompile tail

After the existing gates, the handler stages the recovery block (hash/r/s + recid = v−27), calls the kernel, keccaks the recovered key, and returns the 32-byte left-padded address (returndata length 32, plus the caller out-window copy) — matching execution-specs `ecrecover`. An invalid signature (or a recovery/keccak backend failure) keeps the spec's success-with-empty-returndata behavior.

### Closure hygiene (the reason for the pointer indirection)

The kernel is reached through a new `ecrecover_backend_ptr` data cell rather than a direct `jal`:
- the **stateless guest** arms it in `dispatch_tx_runtime_code` (its verdict closures now link the secp256k1 chain);
- **standalone dispatch probes** leave it 0 and keep the legacy empty-returndata success **and keep linking** without the secp256k1 dependency tree (same data-driven optionality as `callee_seed_count`).

New `secp256k1FieldCommonFunctionsNoU256` / `secp256k1CurveCommonFunctionsNoU256` split the generic `u256_add_be`/`u256_sub_be`/`u256_lt_be` helpers out of the suite so the verdict closures (which already define them) can link the chain without duplicate labels. The frontier matrix row for ECRECOVER is updated.

## Validation

- `codegen-zisk-tx-pubkey-recover-raw-status-check.sh` (incl. `RECOVER_RAW_FULL=1` full recovery) and `codegen-zisk-tx-pubkey-public-key-matches-check.sh` — green (extraction is behavior-preserving).
- `codegen-stateless-link-check.sh` — guest links. (`zisk_stateless_verdict_v2`'s debug ELF has 40 pre-existing undefined references on clean main — unrelated system-call staging labels — unchanged here.)
- EEST (zkevm@v0.4.0) ecrecover family (80 rows): **80/80 full match** — identical to the main baseline (**80/80**), i.e. a strict no-regression: the family's current verdict outcomes do not yet distinguish empty vs real ECRECOVER returndata, so the correctness evidence for the new output is the probe-validated kernel (known-vector recovery) rather than EEST movement.
- `check-file-size.sh`, `check-forbidden-tactics.sh` — green.

Bead: `evm-asm-fhsxz.2.4.2.62.2.5`


pirapira:
Follow-up pushed to this branch: 625933d96 fixes an a0/x10 alias bug in the recovery glue (the status branches read the restored EVM code pointer instead of the kernel status, so valid signatures took the empty-returndata path and the recovery output was dead code). Caught by the end-to-end dispatcher probe now stacked as #8724, which recovers valid_signature_1 → a94f...bf0b through CALL(0x01) and asserts fail-closed behavior for invalid v / zero r.

PR #8726, opened 2026-06-12
feat(receipts): per-tx log-window capture + per-record logs RLP/bloom materialization (inert)

pirapira:
## Summary

Stacked on #8725 (logs-RLP leaf) — the remaining `.63.1.6.2.1` plumbing between the dispatcher's per-call log capture and with-log receipt encoding. Every `runtime_dispatcher_call` resets env+472 and overwrites `evm_event_logs`/`evm_log_data` from index 0, so:

- **`block_log_window_snapshot`** copies each dispatch's descriptors (256 B, 128/block cap) + data bytes (64 KiB cap, offsets rebased) into a new block-level log arena before the next dispatch; called from `dispatch_tx_runtime_code` and the EOA dispatch path, with all three verdict capture sites storing the window into per-tx `bv_tx_log_window` slots. Overflows set `bv_block_log_overflow` and zero the window (conservative).
- **`block_receipt_records_materialize`** takes the window array as a new `a4` argument and threads each tx's window into the record (checkpoint = start, final = start + count; failed txs keep zero logs per the existing status gate).
- **`block_receipt_logs_materialize`** (post-pass, exact path only) encodes each with-log record's window via `log_records_encode_rlp`, derives the record bloom via `logs_list_bloom_add`, and fills the `{bloom, logs_rlp, len}` descriptor + `record@56` — exactly the shape `receipt_records_encode_no_logs` consumes.

**Inert by construction**: the result feeds only `bv_receipt_logs_status` (debug); nothing consumes the encoded records yet. Enforcement for tx-bearing blocks stays gated on the EIP-7708 transfer-log emission gap (the dispatcher only emits the synthetic transfer log on the SELFDESTRUCT path — top-level and CALL value transfers never land in the event log; see the receipts-blocker bead filed under `.63.1.6.2`), so enforcing now would false-reject every value-bearing block. PLAN.md records the full overnight stack + the blocker.

## Validation

- `codegen-stateless-link-check.sh` green; receipt-records / gas-result-arena / logs-RLP probes green.
- EEST (zkevm@v0.4.0): log-exercising families **60/60 full match**, eip4895 (no-tx) **40/40**, fixed-seed random-20 **20/20**.
- `check-file-size.sh` / `check-forbidden-tactics.sh` green.

Bead: `evm-asm-fhsxz.2.4.2.63.1.6.2.1` (stacked on #8725)


pirapira:
Follow-up pushed: the EIP-7708 synthetic-log appender (`eip7708_append_synthetic_log`) never wrote `evm_log_data`/`evm_log_data_meta` — its 32-byte amount lived only in the descriptor prefix, so the meta-based consumers added in this PR (snapshot + logs-RLP encoder) would have read STALE meta for SELFDESTRUCT transfer/burn logs. It now mirrors the LOG0..4 capture path (append BE amount bytes + meta {offset,32}, overflow → flag + zero-length slot). Validation: synthetic-logs probe green, eip7708 EEST family **60/60 full match** (vs 59/60 + 1 external-kill ERROR baseline), eip7708 fork-transition rows 8/8, selfdestruct probes link, guest link green. PLAN.md in this PR also records the overnight stack + the transfer-log emission blocker for the enforcement step.

PR #8729, opened 2026-06-12
feat(precompile): real alt_bn128 ecAdd/ecMul (0x06/0x07) via ZisK BN254 accelerators

pirapira:
## Summary

Implements the **alt_bn128 ecAdd (0x06)** and **ecMul (0x07)** EVM precompiles as real in-guest kernels (EIP-196, Amsterdam/EIP-1108 gas), replacing the deterministic safe-fail wrappers. Pairing (0x08) keeps its safe-fail wrapper and is tracked as the follow-up stretch on bead `evm-asm-fhsxz.2.4.2.62.10`.

### Backend
- **`Bn254Field.lean`** — self-contained `bnf_*` mod-p suite over the BN254 base field. Mul and add go through the ziskemu **Arith256Mod** accelerator (`csrs 0x802`, `d = (a*b + c) mod p` with exact 512-bit intermediates): mul = `(a*b + 0) mod p`, add = `(a*1 + b) mod p`, both always fully reduced.
- **`Bn254Curve.lean`** — affine `bnc_*` point ops backed by the ziskemu **Bn254CurveAdd / Bn254CurveDbl** accelerators (`csrs 0x806` / `0x807`, same pre-encoded `.4byte` route as the secp256k1 chain), with the accelerator-excluded special cases (infinity inputs, equal-x doubling, `P + (-P)`, `y = 0`) in software. `bnc_validate_g1` mirrors execution-specs `bytes_to_g1` (coords `< p`, on-curve or `(0,0)` infinity). Scalar mul is MSB-first double-and-add over the **raw** 32-byte scalar (no order reduction, matching `multiply(p0, n)`).

### Dispatcher semantics (execution-specs alt_bn128)
- Zero-padded staging (`buffer_read`) and the 150 / 6000 gas constants were already in place; the entries now charge against the **EIP-150 child allotment** `min(call gas word, 63/64 · remaining)`:
  - allotment < cost → child OOG: the allotment burns, the parent **pushes 0 and continues** (previously this halted the whole frame via `.exit_outofgas`);
  - invalid input (coord ≥ p / off-curve) → kernel status routes to a failed call that burns the rest of the allotment — the spec's `InvalidParameter → OutOfGasError`;
  - success → 64-byte BE `x || y` returndata (infinity = 64 zero bytes).
- Kernel status is stashed in `x16` before the saved-PC restore (`a0` aliases `x10` — the ecrecover-path landmine).

### Validation
- **Focused probes** `zisk_bn254_g1_add_real` / `zisk_bn254_g1_mul_real` + `scripts/codegen-zisk-bn254-g1-check.sh`: 18 vectors vs a pure-Python EIP-196 reference (generator vectors, infinity combos, `P+(-P)`, equal-x doubling, scalar = 0 / order / order+1 / 2^256−1, off-curve and `x,y ≥ p` rejections), each run gated at `-n 1000000000`. **18/18 pass.**
- **Worst-case steps**: ecMul with `k = 2^256−1` = **933,599 ziskemu steps** (~0.1% of the 1e9 budget).
- **Dispatcher cases** (`Tests/Cases.lean`): bn254 cases updated to real-kernel expectations — known-vector returndata through CALL + RETURNDATACOPY, child-OOG continuation, exact gas-after-call values, invalid-point failed call. **14/14 pass** (run standalone; the shared `codegen-opcodes-runtime-check.sh` runner currently aborts at an unrelated pre-existing case, `create_initcode_len65536_gas_exact`, whose ~753k steps exceed the script's `-n 500000` cap on main too).
- **EEST** (`zkevm@v0.4.0`, stateless guest): `--filter eip196_ec_add_mul` first 50 fixtures — **50/50 PASS(full)** (all 105 verdict bytes), covering ecadd invalid + valid and ecmul invalid families; remainder of the family + a random regression sweep in flight, results to follow as a PR comment.

### Notes
- The BN254 chain is linked by every closure that embeds the runtime dispatcher (both the standalone-probe and shared-guest emission sites), with its data fragments in the runtime dispatcher data core next to the ecrecover cells.
- Pre-existing failures in `codegen-opcodes-runtime-check.sh` (`create_*`, `log0_overflow_status`, and the case-153 abort) reproduce byte-identically on main's ELF and are untouched by this PR.


pirapira:
EEST results (zkevm@v0.4.0, stateless guest, jobs=2, -n 1e9):

- `--filter eip196_ec_add_mul` **full family: 122/122 PASS(full)** (all 105 verdict bytes) — 50 in the first batch + 72 with `--skip 50`; covers ecadd valid/invalid, ecmul valid/invalid, and the gas sub-family (`gas_costs.json`, `invalid_gas_consumption.json`). errored=0, budget=0.
- Random no-regression sweep `--limit 20 --random --seed 8729`: **20/20 PASS(full)**, errored=0.

Focused probe gate `scripts/codegen-zisk-bn254-g1-check.sh`: 18/18 reference vectors; worst-case ecMul (k=2^256−1) = 933,599 steps.

PR #8740, opened 2026-06-12
feat(precompile): real BLS12-381 G2 ADD/MSM (0x0d/0x0e) — software Fp2 kernels

pirapira:
## Summary

Replaces the G2 safe-fail shims with real EIP-2537 kernels. There is no G2 curve accelerator, so the affine chord/tangent formulas are software over Fp2 = Fp[u]/(u²+1), built from single-syscall ops: Bls12_381ComplexAdd/Sub/Mul (csrs 0x80E/0x80F/0x810, one call per Fp2 op on LE 96-byte buffers) and Arith384Mod (Fp norm, Fermat inverse x^(p−2) ≈ 570 calls, negation via p−1). Stacked on the G1 PR.

- `Programs/Bls12G2.lean`: `blsg2_decode_g2` (4 padded felts: pad-zero + coord < p, on-curve y² = x³ + 4(u+1)); alias-safe point add/dbl (results staged through scratch — the scalar-mul loop runs in place); REAL order-n subgroup check for MSM inputs; kernels on raw EVM-memory input with compact 192-byte BE output (dispatch tails expand to 256-byte returndata).
- `ChildFrameHandlers.lean`: entries 15/16 moved to the child-allotment gas model (exact-512 / nonempty-multiple-of-288 gates burn the allotment; per-pair discount via `bls12MsmCostAsm`), raw input pointer, a0-landmine fix.

## Validation
`scripts/codegen-zisk-bls12-g2-check.sh`: **17/17 reference cases PASS** (generator vectors, infinity/doubling/negation, pad/range/curve rejections, off-subgroup rejected by MSM via an explicit Fp2-sqrt witness, raw scalars incl. k = n). EEST g2 family sampling to follow once the stack's base lands (2-ziskemu budget).

Bead: evm-asm-fhsxz.2.4.2.62.11


pirapira:
EEST validation (stack tip, default 1e9 step budget):
- `--filter bls12_g2 --limit 100` (g2add / g2mul / g2msm sample per the operator's ~100-row sampling directive): **100/100 full match** (0 fail, 0 errored, 0 budget).
- `--filter bls12_g2msm/call_types --all` (incl. the heavy 128-pair `max_discount` rows): **6/6 full match** at the default budget — the affine-Fp2 kernels stay well under the step limit, so the Jacobian-coordinates optimization in bead .11.3 is not needed for conformance.

(Validated on `feat/bls12-msm-perf`, which includes this PR + the t155 overflow fix #8738 that the max_discount fixtures require.)

PR #8914, opened 2026-06-15
feat(rlp): unified single-item decode theorem (capstone)

dhsorens:
## What

The **capstone** of the RLP RISC-V decoder: one theorem, `rlp_decode_single_item_spec_within`, proving that for an **arbitrary prefix byte** the decoder decodes a single RLP-item header correctly. Its conclusion is a `match classifyPrefix pfx` dispatching to the five proven per-class full-path handlers — the RV64 analogue of the pure-spec `decodeAux_cons_eq_classifyPrefix_match`.

| class | exit | decoded length (`x11`) |
|---|---|---|
| singleByte | `e1_target+4` | `1` |
| shortBytes | `e2_target+8` | `ofNat (rlpPrefixShortBytesPayloadLen pfx)` |
| longBytes | `(e3_target+12)+24` | `ofNat (Nat.fromBytesBE (rlpLoopByteList …))` |
| shortList | `e4_target+8` | `ofNat (rlpPrefixShortListPayloadLen pfx)` |
| longList | `(base+44)+24` | `ofNat (Nat.fromBytesBE (rlpLoopByteList …))` |

This completes the single-item RLP decode arc end-to-end: classify → per-class length extraction (incl. the general n-iteration long-form loop) → unified dispatch.

## How

Each handler is already a full `base → decode` path gated by `classifyPrefix pfx = .<class>`, so the unified spec is a **case-split**, not a classifier-then-continue. Key design point: the long-form-only proof obligations (per-byte window `hwin`, loop back-edge `hback`) can't live in the runtime precondition (they're needed at proof time to invoke the handler), so they ride in a **`match`-typed hypothesis** `rlpDecodeLongHyps` — `longBytes`/`longList` require `hwin ∧ hback`, other classes require `True`. Flat callers thus supply no memory/window facts.

Proof: `cases classifyPrefix pfx` reduces the conclusion `match` per branch; `exact <handler> … h …` (destructuring `rlpDecodeLongHyps` for the two long-form branches).

## Verification

- ✅ `lake build` — full library green, 0 sorry
- ✅ `scripts/check-no-warnings.sh` — **0 warnings under EvmAsm/** (run pre-push)
- ✅ `scripts/check-forbidden-tactics.sh` — no `native_decide` / `bv_decide`
- ✅ `#print axioms rlp_decode_single_item_spec_within` → only `propext` / `Classical.choice` / `Quot.sound`
- Per-class `classifyPrefix` sanity `example`s (`0x00`→singleByte … `0xFA`→longList)

## Scope / future work

This is an **operational** unified spec stating each branch's length via the pure per-class functions. Out of scope (separate, larger efforts): a full semantic bridge to pure `decodeAux` — needs canonical-encoding (`>55`/leading-zero) validation the RV64 path doesn't perform **and** recursive list decode (`decodeItems`) for list payloads; cross-doubleword length spans (`n ≤ 8`); Phase-4 `read_input` pipeline.


pirapira:
Looks good to me. Merging

PR #8974, opened 2026-06-16
feat(rv64): multi-dword byte-region memory abstraction (RLP list-decode foundation)

dhsorens:
## Summary

First step of the RV64 RLP **list decoder** (Phase 5). The decoder currently
reads bytes only *within a single 8-byte dword* (`Phase2LongLoopGeneral.hwin`
forces one `dwordAddr`), so a list payload spanning multiple dwords is
inexpressible. This lands the foundational separation-logic resource for it.

- **`EvmAsm/Rv64/MemRegion.lean`** — `bytesRegion base (bs : List (BitVec 8))`: a
  contiguous byte region stored little-endian across consecutive dwords (each
  `↦ₘ` via `packBytes`), mirroring `Evm64.CodeRegion.evmCodeIs` but at the `Rv64`
  layer so the decoder can use it without inverting layering. Lemmas:
  `bytesRegion_nil`, `bytesRegion_eq_cons` (peel one dword, address `+8` — the
  workhorse consumers iterate), `bytesRegion_pcFree` (so it frames). Wired into
  the `EvmAsm.Rv64` umbrella.

- **Relocation (reuse, don't reinvent):** moved the pure byte-packing primitives
  `packBytes` / `extractByte_packBytes` / `packDword` / `getByteAt` from
  `Evm64/CodeRegion.lean` down to `Rv64/ByteOps.lean` (their natural byte-ops
  home; one source of truth now shared by the EVM code-region model and the new
  RV64 byte-region model). `CodeRegion` was the sole user and picks them up via
  its existing `open EvmAsm.Rv64`.

## Scope

Foundation only — the structural core (definition + recursion equation + framing)
that consumers peel through. The keystone **`bytesRegion_lbu_within`** (read byte
`i` across dwords) is the next PR: it needs the `alignToDword (alignedBase + i) =
base + 8·(i/8)` BitVec arithmetic that the existing loop currently *assumes* via
`hwin`. Then the list-decode loop + the unified decoder's 5-exit reconvergence.

## Verification

- Full `lake build` green (incl. `Evm64/CodeRegion` after the relocation).
- `scripts/check-no-warnings.sh`: 0 warnings under `EvmAsm/`.
- `scripts/check-forbidden-tactics.sh`: OK.
- `#print axioms bytesRegion_eq_cons` / `bytesRegion_pcFree` → `[propext, Quot.sound]`; 0 sorry.


pirapira:
Looks good to me. Merging

PR #8978, opened 2026-06-16
feat(rv64): single-byte-item RLP list-decode loop + decodeItems bridge

dhsorens:
## Summary

The **first RV64 RLP _list_-decode loop**. A 4-instruction back-branch body

```
LBU  x12, x13, 0     ; byte = mem[x13]
ADDI x13, x13, 1     ; ptr += 1
ADDI x14, x14, -1    ; counter -= 1
BNE  x14, x0, back   ; if counter != 0 loop back
```

traverses a payload of **single-byte items** (each byte `< 0x80` is its own RLP item), reading each byte from the multi-dword `bytesRegion` via the keystone `bytesRegion_lbu_within` (landed in #8974/#8975), advancing the pointer and decrementing the counter. The decoder is zero-copy.

**Scope (assume-precondition):** every payload byte `< 0x80` is *assumed* via a precondition — the same hypothesis the pure spec needs — so the loop is a clean 2-exit counter loop. In-line `< 0x80` validation (a 3-exit fail path) is a deliberate follow-up.

## What's here (`EvmAsm/Rv64/RLP/SingleByteListLoop.lean`)

- **`sbll_iter_spec_within`** — 3-instruction iteration triple (LBU + ptr advance + counter decrement).
- **`sbll_body_spec_within`** — the 2-exit `cpsBranchWithin` loop body (iteration ∪ `BNE`).
- **`sbll_loop_succ_spec_within` / `sbll_loop_n_spec_within`** — the n-iteration closure by **remaining-count induction with a generalized _absolute_ start index**. Unlike the Phase-2 length loop (which shifts a single iteration-invariant `dwordAddr ↦ₘ`), `bytesRegion_lbu_within` is keyed by the absolute byte index, so the induction keeps `regionBase`/`bs`/alignment fixed and re-indexes only the per-iteration validity hypotheses.
- **`sbll_loop_bridge`** — packages the operational spec with the pure `decodeItems_singleByte_run` result, tied by the shared `< 0x80` precondition and the matching consumed length `bs.length` (the decoder is zero-copy, so the operational half materializes no item structure — this is stated honestly in the docstring).
- **Cross-dword cross-check** `example`: a 10-byte / 2-dword payload decodes to 10 single-byte items in `4 * 10` steps — a cross-dword traversal the single-`dwordAddr` length loop could not express.

Reuses Phase-2's `word_ofNat_*` / `cnt_dec_1` counter lemmas (no duplication).

## Verification

- `#print axioms sbll_loop_bridge` (+ closure + body) → `[propext, Classical.choice, Quot.sound]` only.
- 0 `sorry`/`admit`.

## Next

In-line `< 0x80` validation (3-exit fail path) discharging the precondition operationally; then varying-size items + the unified decoder's 5-exit reconvergence.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -55,6 +55,7 @@ import EvmAsm.Rv64.RLP.Phase1E5LongListSeven
 import EvmAsm.Rv64.RLP.Phase1E5LongListEight
 import EvmAsm.Rv64.RLP.Phase2LongLengthBridge
 import EvmAsm.Rv64.RLP.Phase2LongLoopGeneral
+import EvmAsm.Rv64.RLP.SingleByteListLoop
 import EvmAsm.Rv64.RLP.Phase1E3LongBytesFull
 import EvmAsm.Rv64.RLP.Phase1E5LongListFull
 import EvmAsm.Rv64.RLP.UnifiedDecodeItem
--- a/EvmAsm/Rv64/RLP/SingleByteListLoop.lean
+++ b/EvmAsm/Rv64/RLP/SingleByteListLoop.lean
@@ -1,0 +1,465 @@
+/-
+  EvmAsm.Rv64.RLP.SingleByteListLoop
+
+  EL.3 — the first RV64 RLP *list*-decode loop: a back-branch loop that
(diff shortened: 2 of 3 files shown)


pirapira:
Looks good to me. Merging.

PR #8980, opened 2026-06-16
feat(rv64): in-line < 0x80 validation for the RLP single-byte-item loop

dhsorens:
## Summary

Drops the `hsingle` precondition from the single-byte-item list-decode loop (landed in #8978). The validated loop now **proves** every payload byte `< 0x80` in-machine rather than assuming it, so the decoder operationally rejects a payload containing any byte `≥ 0x80` (which is *not* a list of single-byte items).

## Design — accumulate-then-check (not per-iteration early-exit)

A per-iteration `BGEU byte, 0x80, fail` early-exit would force a *branching* loop closure (compose branch-then-branch, merging fail exits — new composition infrastructure). Instead the loop **ORs every byte into an accumulator (`x11`) as it scans** — a `cpsTripleWithin` closure mirroring the Phase-2 length accumulator — then a single post-loop check:

```
ANDI x15, x11, 0x80     ; x15 = acc & 0x80   (bit 7 = OR of all bytes' bit 7)
BNE  x15, x0,  fail      ; some byte had bit 7 set  ⇒  not all single-byte
```

This drops `hsingle` just as genuinely while reusing the existing closure structure and needing **no new combinators**. `OR` never overflows, so — unlike the length loop — there is no `n ≤ 8` bound.

## What's here (`EvmAsm/Rv64/RLP/SingleByteListLoopValidated.lean`)

- **`orAccList` + `orAccList_and_0x80_eq_zero_imp_all_lt`** — the key BitVec lemma: accumulator bit-7 clear ⇒ every byte `< 0x80`. Proved without `bv_decide` via `BitVec.msb_eq_false_iff_two_mul_lt` and an OR-fold induction.
- **`sbll_val_iter / body / loop_succ / loop_n_spec_within`** — the OR-accumulator loop and its n-iteration closure (same remaining-count induction as the assume-version).
- **`sbll_val_loop_checked_spec_within`** — the full validated spec as a 2-exit `cpsBranchWithin` (success = `acc &&& 0x80 = 0`; fail = some byte `≥ 0x80`), composed from `loop ∘ ANDI ∘ BNE` with existing combinators.
- **`sbll_val_loop_bridge`** — derives `decodeItems_singleByte_run` from the *machine-checked* success condition. **No `hsingle`.**
- Cross-dword cross-check at 10 bytes (2 dwords).

## Verification

- `#print axioms sbll_val_loop_checked_spec_within` (+ bridge + key lemma) → `[propext, Classical.choice, Quot.sound]` only.
- 0 `sorry`/`admit`.

## Next

Varying-size RLP items; the unified decoder's 5-exit reconvergence.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -56,6 +56,7 @@ import EvmAsm.Rv64.RLP.Phase1E5LongListEight
 import EvmAsm.Rv64.RLP.Phase2LongLengthBridge
 import EvmAsm.Rv64.RLP.Phase2LongLoopGeneral
 import EvmAsm.Rv64.RLP.SingleByteListLoop
+import EvmAsm.Rv64.RLP.SingleByteListLoopValidated
 import EvmAsm.Rv64.RLP.Phase1E3LongBytesFull
 import EvmAsm.Rv64.RLP.Phase1E5LongListFull
 import EvmAsm.Rv64.RLP.UnifiedDecodeItem
--- a/EvmAsm/Rv64/RLP/SingleByteListLoopValidated.lean
+++ b/EvmAsm/Rv64/RLP/SingleByteListLoopValidated.lean
@@ -1,0 +1,687 @@
+/-
+  EvmAsm.Rv64.RLP.SingleByteListLoopValidated
+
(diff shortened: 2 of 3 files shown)


pirapira:
Looks good to me. Merging.

PR #8989, opened 2026-06-16
fix(eip8037): compute type-4 receipt gas as tx_regular + tx_state

pirapira:
## Summary
- Replace the fragile per-shape EIP-7702 receipt-gas reconstruction in `block_verdict_receipt_gas_eip8037_adjust` (the branchy `+42690/+35190/2500` add-state/add-auth logic) with the spec-exact decomposition:

  `receipt = max((before_refund − tx_exec_state_gas + tx_total_state_gas + PER_AUTH_BASE_COST·auth_count) − gas_refund, calldata_floor)`

  which is `tx_regular_gas + tx_state_gas` (amsterdam `make_receipt` / `process_transaction`, fork.py:1132–1205), computed entirely from verdict-side arrays the guest already has.
- Remove the now-subsumed narrow EXTCODECOPY-same-block (`ecc_same_block_hit`) `+32690` receipt add-on in `BlockVerdictReceiptsTail` (the SELFDESTRUCT `+32690` was already removed in #8988). The clean formula computes the receipt from scratch, so re-adding either would double-count.

## Why
Ground-truth instrumentation showed the runtime regular pool omits the EIP-7702 per-auth regular intrinsic (`PER_AUTH_BASE_COST=7500/auth`), so the raw `before_refund` is short by that term; the prior per-shape constants were compensating shape-by-shape. The formula adds it back explicitly. **No runtime/OOG change** — the EIP-8037 2D-gas execution path is untouched.

## Validation
- `set_code_to_self_destruct[balance_1]`: receipt 171619 (= 38509 + 133110), succ=01 (stays passing, now via the clean formula not `+42690`).
- `--filter eip7702_set_code_tx --limit 120`: 120/120 full match.
- Random sweep (`--random --seed 20260616 --limit 500`): **500/500 full match, 0 fail, 0 errored**.
- `lake build` green (pre-existing `LeanRV64D/RiscvExtras.lean` deprecation warning only); `check-file-size.sh`, `check-unimported.sh` clean; no `sorry`/`maxHeartbeats`/`maxRecDepth`.

## Follow-ups (isolated by this change)
The remaining `bv_fail=53` false-rejects are now pinned to **separate intrinsic/execution over-charge bugs** (`before_refund` too high), not receipt-reconstruction: `pointer_to_static_reentry` (+2500 cold-access), `set_code_to_log` (+378 intrinsic) — tracked in evm-asm-5tmlt / evm-asm-mk3ic.

Bead: evm-asm-huo4a.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdictReceiptsTail.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictReceiptsTail.lean
@@ -47,24 +47,12 @@ def blockVerdictReceiptsTail : String :=
   "  la a5, bvgr_block_gas_increments\n" ++
   "  la a6, bvgr_tx_exec_state_gas\n" ++
   "  jal ra, block_verdict_receipt_gas_eip8037_adjust\n" ++
-  -- NOTE (msdfw): a previous narrow EIP-7702 SELFDESTRUCT receipt correction
-  -- here unconditionally added 32690 to the single-tx type-4 receipt increment
-  -- whenever evm_selfdestruct_staged was set. That was a pre-8977 workaround for
-  -- when block_verdict_receipt_gas_eip8037_adjust under-counted the per-auth
-  -- state component. After PR #8977 the generic type-4 adjust adds the full
-  -- per-authorization state component (42690) on the single-auth path, so the
-  -- runtime regular increment already reconstructs tx_gas_used_after_refund for
-  -- delegated SELFDESTRUCT rows (e.g. set_code_to_self_destruct balance_1: the
-  -- receipt is 128929 + 42690 = 171619, matching execution-specs). The extra
-  -- +32690 double-counted and produced a spurious receipts-root mismatch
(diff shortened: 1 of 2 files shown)


pirapira:
⚠️ DO NOT MERGE — regression found. Converting to draft.

A targeted `set_code_to_sstore` family run shows this change REGRESSES cases that pass on main:
- `set_code_to_sstore[...return-tx_value_1]` (00000) and `[...stop-...]` (00001): main succ=01 (PASS), this branch succ=00 (FAIL).

Root cause: the clean `tx_regular + tx_state` formula oversimplifies the SSTORE refund interaction that the prior branchy per-shape adjust handled correctly. The 500-case random sweep missed these by sampling; the targeted family run caught it.

Lesson: validate the specific affected family (eip7702 set_code_txs incl. set_code_to_sstore), not only a random sweep, before relying on "behavior-preserving". Needs a corrected formula (proper SSTORE gas-refund / state-refund handling) or revert.

pirapira:
Still a good change. merging.

PR #9004, opened 2026-06-17
feat(rv64): flat-item RLP stride-equivalence (list-loop closure foundation)

dhsorens:
## Summary

The mathematical core needed for the flat-item list-decode loop closure: the `isFlatItem` predicate and the **stride-equivalence** linking the operational per-item pointer advance to the pure encoding length.

This is the first slice of the n-closure + bridge work (the closure's variable-stride induction relies on exactly this lemma). Landing it on its own keeps a clean, verified checkpoint; the operational closure + `decodeItems` bridge follow.

## What's here (`EvmAsm/Rv64/RLP/FlatListLoop.lean`)

- **`isFlatItem`** — `.bytes data` with `data.length ≤ 55`; `.list` with payload `≤ 55`: exactly the items whose encoding's first byte is a flat prefix.
- **`classifyPrefix_encode_head_flat`** — a flat item's encoding starts with a `singleByte`/`shortBytes`/`shortList` prefix.
- **`encode_head_eq_itemTotalLen`** — `itemTotalLen ((encode item)[0]) = BitVec.ofNat 64 (encode item).length`: the loop's per-item stride (computed from the decoded prefix byte) equals the pure canonical encoding length. Proved by per-class byte-shape analysis (`encodeBytes` / `encode_list_short` / `classifyPrefix_*_iff`); no `bv_decide`.

This is the bridge between the machine loop's pointer arithmetic and the pure `encode`/`decodeItems` round-trip.

## Verification

- `#print axioms encode_head_eq_itemTotalLen` → `[propext, Classical.choice, Quot.sound]` only.
- 0 `sorry`/`admit`.

## Next

The n-iteration closure (count-induction over a flat-item list, threading the variable byte offset, applying `fll_body_spec_within` per item with the decoder as a ∀-hypothesis) + the `decodeItems` bridge (via the existing `decode_encode_mutual.2` round-trip).

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -63,6 +63,7 @@ import EvmAsm.Rv64.RLP.UnifiedDecodeItem
 import EvmAsm.Rv64.RLP.UnifiedDecodeItemReconverge
 import EvmAsm.Rv64.RLP.UnifiedDecodeItemReconvergeAll
 import EvmAsm.Rv64.RLP.FlatListLoopBody
+import EvmAsm.Rv64.RLP.FlatListLoop
 import EvmAsm.Rv64.RLP.Phase1E3LongStringFromBytesBE
 import EvmAsm.Rv64.RLP.Phase1E5LongListFromBytesBE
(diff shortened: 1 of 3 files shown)


pirapira:
Looks good to me. Merging.

PR #9012, opened 2026-06-17
feat(verdict): enforce multi-tx cumulative sender balance (bmvmx.5.5.2.2.3)

pirapira:
## What

Enforces the multi-tx **cumulative sender balance** check (bead `bmvmx.5.5.2.2.3`, umbrella-B2.3): in `blockVerdictMtxValidationTail`, after the existing B2.2 running-balance loop, compare each distinct sender's running balance (`sender_pre − Σ actual debit`, where `debit = receipt_inc·eff_price + tx.value`, accumulated in `bv_b2_table`) against the BAL-declared sender FINAL balance. A mismatch rejects with the new `bv_fail=57` (`.Lbv_mtx_sender_balance_fail`).

## Why (soundness)

Multi-tx senders sit in the A1 skip-list and are therefore **excluded from the A2a all-accounts non-storage comparator**, so until now a multi-tx sender's BAL balance delta was unvalidated against the execution gas model. The state-root recompute (`block_state_root`) *applies* the BAL-declared deltas to the pre-state, so a forged sender post balance paired with a matching forged `header.state_root` would survive every other check. B2.3 is the only check tying a multi-tx sender's BAL balance delta to the execution-derived gas cost — the cumulative generalization of the single-tx `tx_gas_bal_post_verify_runtime` (status 40). This closes a class-F "skipped check".

## Conservative skips (never false-reject)

The running balance models only `gas + value-out` and cannot see inbound credits or extra debit terms, so the comparison is skipped (not enforced — never falsely rejects) when:
- effect-log overflow or any withdrawals present (the exec effect log is then incomplete; mirrors the adjacent A2a guard);
- **any tx in the block is type ≥ 3** (blob fee / EIP-7702 AUTH_BASE are extra sender-debit terms `multi_tx_actual_sender_debit` does not yet model — a per-sender blob/auth-fee debit term is a follow-up);
- the sender == coinbase/fee_recipient (priority fee credited outside the EVM frame);
- the sender appears in the exec non-storage effect log (execution touched its balance → potential inbound CALL/value credit).

The remaining pure-payer senders (the common multi-tx type-0/1/2 EOA case) must satisfy `BAL_post == pre − Σdebit` exactly.

## Validation

- Build clean (incl. `RegistryInvariants`); guest assembles + links clean.
- All 6 eip8037 `state_gas_multi_block` cases PASS(full): `exact_coinbase_fee_simple_sstore`, `multi_block_observed_coinbase_balance` b0/b1, `multi_block_mixed_state_operations` b0/b1/b2.
- 90/90 `eip7928_block_level_access_lists` PASS(full) (incl. value-transfer cases).
- `bal_all_transaction_types` (type-3/4 block) PASS(full) via the tx-type skip gate — caught as a regression in the first 250-random sweep and fixed before this PR.
- Non-vacuity confirmed via a (reverted) force-reject diagnostic: the comparison is reached for real pure-payer senders and propagates to `successful_validation=0`.
- Static analysis confirmed both coinbase fixtures have distinct senders, none == coinbase, value=0 — so the prior B2 false-reject was the (reverted) state-gas double-count, not a missing credit.

Forged-sender-post EEST fixture coverage remains the welder lane (`il7sn`); guest fix shipped against spec + existing-fixture regression per project policy.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdictDataSection.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictDataSection.lean
@@ -1229,6 +1229,11 @@ def ziskStatelessVerdictV2DataSection : String :=
   ".balign 32\n" ++
   "bv_b2_table:\n  .zero " ++ toString bvMtxSenderBalanceTableBytes ++ "\n" ++
   "bv_b2_debit_out:\n  .zero 48\n" ++
+  -- B2.3 tx-type pre-scan scratch: type-3 (blob) / type-4 (auth) txs carry extra
+  -- sender-debit terms multi_tx_actual_sender_debit does not yet model, so the B2.3
+  -- exact compare is skipped wholesale when any tx in the block is type >= 3.
+  "bv_b23_txtype:\n  .zero 8\n" ++
+  "bv_b23_innoff:\n  .zero 8\n" ++
   "mtxsd_gascost:\n  .zero 32\n" ++
   -- i3djw.3: scratch for bal_all_accounts_nonstorage_consistent + its per-account deps
   -- (bal_account_nonstorage_consistent / _finals). rfu_* is already linked (other rlp users).
--- a/EvmAsm/Codegen/Programs/BlockVerdictMtxTail.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictMtxTail.lean
@@ -116,6 +116,72 @@ def blockVerdictMtxValidationTail : String :=
   ".Lbv_b2_next:\n" ++
   "  la t0, bv_mtx_skip_idx; ld t1, 0(t0); addi t1, t1, 1; sd t1, 0(t0); j .Lbv_b2_loop\n" ++
   ".Lbv_b2_done:\n" ++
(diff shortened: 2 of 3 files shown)


pirapira:
**Follow-up beads for the conservative exemptions in this PR** (each makes a currently-skipped case enforceable):

- `evm-asm-bmvmx.5.5.2.2.6` (P1) — model the blob-fee (type-3) + AUTH_BASE (type-4) sender-debit terms so blob / EIP-7702 blocks aren't skipped by the `tx-type >= 3` gate.
- `evm-asm-bmvmx.5.5.2.2.7` (P1) — fold the coinbase-fee (sender==coinbase) and inbound-value (sender in effect log) credits into the running balance so credit-bearing senders are enforced.
- `evm-asm-bmvmx.5.5.2.2.8` (P1) — credit withdrawal-to-sender amounts into the running balance so blocks with withdrawals aren't skipped.
- `evm-asm-bmvmx.5.5.2.2.9` (P1) — lift the `exec_nonstorage_effect_log` capacity so over-capacity blocks aren't skipped.

Each conservative skip in this PR is sound (never false-rejects) and is tracked above for tightening.

PR #9013, opened 2026-06-17
feat(verdict): debit type-3/4 fees so B2.3 enforces blob/EIP-7702 multi-tx blocks

pirapira:
## What

Follow-up to #9012 (`bmvmx.5.5.2.2.6`): makes the B2.3 cumulative sender-balance check **enforce** multi-tx blocks that contain type-3 (blob) or type-4 (EIP-7702) transactions, instead of skipping them.

#9012 conservatively skipped the whole B2.3 pass when any tx was type ≥ 3, because `multi_tx_actual_sender_debit` models only `receipt_inc·eff_price + tx.value` and omits the extra sender-debit terms those tx types carry.

## How

In the B2.2 loop (after the debit call), add the missing terms, mirroring the proven single-tx `tx_gas_bal_post_verify_runtime`:
- **type-4**: `+ auth_count · 42690 · effective_gas_price` (42690 = AUTH_BASE 7500 + 23·PER_EMPTY_ACCOUNT_COST 1530), via `tx_type_dispatch` → RLP item 9 (`authorization_list`) → `rlp_list_count_items`.
- **type-3**: `+ blob_count · 131072 (GAS_PER_BLOB) · blob_gas_price` (`bsg_blob_price_be`), via `tx_eip4844_decode` → blob-hash list count.

An inconclusive typed decode skips that sender (absent from `bv_b2_table`), preserving the no-false-reject property. The logic lives in the loop (not in `multi_tx_actual_sender_debit`) so the standalone `zisk_multi_tx_actual_sender_debit` probe keeps assembling without the `tx_type_dispatch`/`tx_eip4844_decode` dependency tree.

With the debit now exact for type 0–4, the B2.3 `type ≥ 3` pre-scan skip is removed. Type 0/1/2 are unaffected (the auth/blob blocks are skipped for type < 3).

## Validation

- Build clean (incl. `RegistryInvariants`); guest assembles + links clean.
- `bal_all_transaction_types` (mixed type 0–4) now **PASS(full)** with type-3/4 senders **enforced** (it was skipped under #9012).
- Broad 250-random regression sweep: 0 fail (see PR comment).

Closes the `type ≥ 3` exemption tracked by `bmvmx.5.5.2.2.6`. The remaining B2.3 exemptions (`.2.2.7` credits, `.2.2.8` withdrawals, `.2.2.9` effect-log capacity) stay tracked.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdictDataSection.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictDataSection.lean
@@ -1229,11 +1229,19 @@ def ziskStatelessVerdictV2DataSection : String :=
   ".balign 32\n" ++
   "bv_b2_table:\n  .zero " ++ toString bvMtxSenderBalanceTableBytes ++ "\n" ++
   "bv_b2_debit_out:\n  .zero 48\n" ++
-  -- B2.3 tx-type pre-scan scratch: type-3 (blob) / type-4 (auth) txs carry extra
-  -- sender-debit terms multi_tx_actual_sender_debit does not yet model, so the B2.3
-  -- exact compare is skipped wholesale when any tx in the block is type >= 3.
+  -- B2.3 typed-tx fee scratch (bmvmx.5.5.2.2.6): the B2.2 loop adds the type-4 AUTH_BASE
+  -- and type-3 blob-data-gas sender-debit terms that multi_tx_actual_sender_debit omits,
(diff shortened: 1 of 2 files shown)


pirapira:
Broad regression sweep (250 random, full build): **250/250 full match, 0 fail**. `bal_all_transaction_types` (mixed type 0-4) confirmed PASS(full) with type-3/4 senders now enforced (it was skipped under #9012).

PR #9014, opened 2026-06-18
feat(verdict): raise non-storage effect-log cap 64->256 to enforce mid-size blocks

pirapira:
Raises `nonstorageEffectLogCap` 64→256 so blocks with 64–256 non-storage (balance/nonce) effects are **enforced** by the A2a exec-vs-BAL comparator + B2.3 cumulative-balance check instead of conservatively skipped on overflow. Partial elimination of the `bmvmx.5.5.2.2.9` effect-log-overflow skip.

**Sound by construction:** a larger arena only enables more enforcement — it cannot false-reject (valid blocks still pass A2a/B2.3). Blocks with >256 such effects still overflow/skip, so the larger aggregation work cannot push a near-step-budget huge block past its budget. The hardcoded `exec_nonstorage_effect_agg` buffer is grown in lockstep (7168→28672 = cap×112; the `.Lbv_agg_append` loop has no separate bounds check).

**Validation:** build clean; all 22 eip8037 `block_2d_gas` fixtures (incl the 1021-tx `block_2d_gas_valid_when_cumulative_exceeds_limit`) PASS(full), 0 fail. Broad sweep result in a comment.

Full 200M-gas capacity remains the `bmvmx.5.5.7.*` track (needs the O(N²) aggregation linearized first).

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdictDataSection.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictDataSection.lean
@@ -1198,11 +1198,13 @@ def ziskStatelessVerdictV2DataSection : String :=
   -- for the multi-tx nonstorage comparators. record_nonstorage_effect APPENDS one record
   -- per CALL, so a multi-tx-touched account has N records; fold them into one entry keyed
   -- by the 20B BE address (first-seen pre kept, last-seen post overwritten) so the per-
-  -- account comparator sees the block-aggregate {pre, post}. Dedup -> count <= the log cap
-  -- (64), so 64 x 112 B suffices.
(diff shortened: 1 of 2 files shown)


pirapira:
Broad regression sweep (250 random, full build): **250/250 full match, 0 fail**. Combined with the 22/22 `block_2d_gas` fixtures (incl the 1021-tx block staying PASS(full)), the cap lift is 0-regress.

pirapira:
Superseded by #9019 (effect-log cap 64->2048), which subsumes this 64->256 interim lift and scales the agg buffer to 229376. #9019 is stacked on the O(N) aggregation + binary-search chain (#9015->#9016->#9017->#9018) that makes the larger cap safe. Closing to avoid a cap-line merge conflict.

PR #9016, opened 2026-06-18
feat(verdict): wire nonstorage_effect_aggregate, replacing the inline O(N^2) agg loop

pirapira:
## What

`bmvmx.5.5.7.3` wiring slice (stacked on #9015, which added the helper). Replaces `blockVerdictMtxValidationTail`'s inline **O(raw·distinct)** per-account non-storage effect aggregation (`.Lbv_agg_loop`) with a call to the KAT-validated linear helper `nonstorage_effect_aggregate` (radix-sort + run-compress).

## Why

The inline O(N²) aggregation was the step-budget barrier blocking any lift of `nonstorageEffectLogCap` toward the 200M-gas worst-case. Replacing it with the O(20·N) helper removes that barrier (the full cap lift is the next slice).

## Soundness / behavior-preservation

- Same **first-pre / last-post** per-address semantics as the inline loop.
- Output order differs (sorted vs first-seen), but `bal_all_accounts_nonstorage_consistent` and `_covers` scan the agg **by address** (order-independent), so the A2a result is unchanged.
- The helper is **stable** (no hashing → no adversarial-collision weakening) and preserves all callee-saved registers.
- Linked into the guest via `emitRuntimeDispatcherEmbeddedHelperFunctions`; sort scratch via the shared dispatcher data section.

## Validation

- Build clean (incl. `RegistryInvariants`) — the wiring links (no missing-symbol / dup-label).
- **250/250 random full-match, 0 fail** (the eip7708/eip7928 multi-tx fixtures exercise the agg path).
- block_2d_gas family (incl. the 1021-tx `block_2d_gas_valid_when_cumulative_exceeds_limit`, the heaviest agg user) — result in a PR comment.


pirapira:
block_2d_gas family: **22/22 PASS(full), 0 fail**, including the 1021-tx `block_2d_gas_valid_when_cumulative_exceeds_limit` (the heaviest agg user) — the wired helper is behavior-preserving even at the largest effect counts.

PR #9020, opened 2026-06-18
feat(rv64): bytesRegion length-read loop (long-item list-loop, step 1)

dhsorens:
## Summary

First step of the **long-item-capable RLP list loop** arc. Real Ethereum RLP (transactions, blocks, account state) routinely exceeds 55 bytes, so the decoder needs to handle `longBytes`/`longList` items — where the prefix byte is followed by *k* big-endian length bytes that must be read from memory.

The verified length-read loop (`Phase2Long*`) is currently proven only for a **single dword** (`alignToDword (ptr+i) = dwordAddr` for all length bytes, via `(dwordAddr ↦ₘ wordVal)`). A region-based list loop reads length bytes from a multi-dword `bytesRegion`, and the *k* bytes may cross a dword boundary. This PR generalizes the loop to `bytesRegion` — the foundational, consumer-independent prerequisite for the rest of the arc.

## What's here (`EvmAsm/Rv64/RLP/Phase2LongLoopRegion.lean`)

Region analog of the `Phase2LongIter` → `Phase2LongLoopBody` → `Phase2LongLoopGeneral` stack. **The body program is unchanged** (`rlp_phase2_long_loop_body_prog`); only the memory model differs.

- **`rlpLoopAccRegion`** (+ `rlpLoopAccRegion_zero_eq_fromBytesBE`) — the byte accumulator over `bs`, decoding (from `0`) to `Nat.fromBytesBE ((bs.drop off).take n)`.
- **`rlp_phase2_long_region_iter_spec_within`** — one iteration (`cpsTripleWithin 5`), proved by extracting the dword containing byte `off` from `bytesRegion` (`bytesRegion_dword_at`) and reusing the single-dword iter — i.e. the **cross-dword read** the `alignToDword` model couldn't express.
- **`rlp_phase2_long_region_body_spec_within`** — one pass with the `BNE` back-branch (`cpsBranchWithin 6`).
- **`rlp_phase2_long_loop_region_succ_spec_within`** — the `k+1 ∈ [1,8]`-iteration closure (induction on `k`).
- **`rlp_phase2_long_loop_region_n_spec_within`** — the from-`0` closure; `x11` decodes to the pure-spec `Nat.fromBytesBE` of the *n* length bytes.

## Notes

- `x0` is framed on the **left** (`cpsTripleWithin_frameL`) so `xperm` only reorders register atoms *before* the opaque `bytesRegion` and never commutes one rightward past it (which it can't). A local `PCFree (bytesRegion …)` instance lets `pcFree` discharge the frame side-conditions.
- The iter reuses the existing single-dword spec rather than rebuilding the 5-instruction framing.

## Verification

- `#print axioms` on the two closures → `[propext, Classical.choice, Quot.sound]` only.
- 0 `sorry`/`admit`.

## Next (the arc)

(2) all-class stride-equivalence (`itemTotalLen = (encode item).length` for long items, via `Nat.fromBytesBE (Nat.toBytesBE n) = n`); (3) region 5-class decoder; (4) long loop body with the item counter on x15 (the decoder clobbers x14); (5) closure + bridge; (6) concrete + end-to-end.


pirapira:
Looks good to me. Merging.

PR #9028, opened 2026-06-18
RLP RV64: unified list-loop closure + bridge (long-loop step 5b)

dhsorens:
## Summary

Adds the **unified (all-class) RLP list-decode loop closure + bridge** — the long-item-capable analog of the flat `fll_loop_*` (`FlatListLoop.lean`, #9006). Closes the loop arc started by the step-4 body (#9026) and the step-5a stride lemma (#9027).

New file `EvmAsm/Rv64/RLP/UnifiedListLoop.lean`:

- **`unified_loop_spec_within`** — the n-iteration closure. Structural induction over an arbitrary `items : List RLPItem`, threading the byte offset via `bs.drop O = encode.encodeItems items` and applying the step-4 loop body (`unified_body_spec_within`) once per item. The 60-step region decoder is supplied abstractly as `UnifiedDecoderH` (∀ over byte index + incoming scratch); each iteration re-indexes `x13` via the 5a stride lemma `encode_head_eq_itemNextPtrRegion`. Uniform **`64 * items.length`** steps, counter on **x15**, decoder scratch x10/x12/x14 abstracted to `regOwn` in `unified_loop_post`.
- **`unified_loop_n_spec_within`** — the offset-0 entry.
- **`unified_loop_bridge`** — conjoins the operational loop with the pure `decodeItems` round-trip (`decode_encode_mutual.2`).

The per-head `itemPayloadCount < 256^8` side-condition (required by the stride lemma) is discharged from `hover`: flat items via `≤ 55`, long items via `encode_long_length_eq`.

## Context

Long-item list-loop arc: step 1 (#9020) … step 4 loop body (#9026), step 5a stride (#9027). This is **step 5b**. Next is **step 6** — the concrete unified program + end-to-end (discharge `UnifiedDecoderH` with the #9025 region decoder, assemble the real RV64 program, prove the no-abstract-hypotheses bridge), the analog of `FlatListLoopConcrete.lean` (#9007/#9008), which completes single-level long-item list decoding.

## Verification

- `lake build` clean (module + umbrella `EvmAsm.Rv64.RLP`).
- `scripts/check-forbidden-tactics.sh` OK (no `bv_decide`/`native_decide`).
- `scripts/check-no-warnings.sh` → 0 warnings under `EvmAsm/`.
- `#print axioms unified_loop_bridge` / `unified_loop_spec_within` → `[propext, Classical.choice, Quot.sound]` (classical only); 0 `sorry`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9008, opened 2026-06-17
feat(rv64): fully concrete end-to-end flat RLP list decoder

dhsorens:
## Summary

The capstone of the flat RLP list-decoder arc: wires the concrete single-item decoder (`flat_decoder_spec`, #9007) into the list-loop bridge (`fll_loop_bridge`, #9006), discharging the loop's abstract `decoderH` and **all** its code-layout side-conditions. The result has **no remaining abstract hypotheses** — it's grounded entirely in real RV64 code.

## What's here (`EvmAsm/Rv64/RLP/FlatListLoopConcrete.lean`)

**`flat_list_loop_concrete_bridge`** — a real RV64 program laid out at `base`:

```
base     LBU  x5, x13, 0          ; read prefix byte
base+4   < flat_decoder_prog : 16 instr, reconverges at base+68 >
base+68  ADD  x13, x13, x11       ; advance to next item
base+72  ADDI x14, x14, -1        ; item counter
base+76  BNE  x14, x0, -76        ; loop back to base
```

decodes a non-empty list of flat RLP items from `bytesRegion regionBase (encode.encodeItems items)` in `15 * items.length` steps — leaving the pointer at the region end and the counter zero — **and** coincides with the pure decoder: `decodeItems (2*len+1) (encode.encodeItems items) = some (items, [])`.

## How the side-conditions are discharged

- **`decoderH`** ← `flat_decoder_spec (base+4)`, with its exit `(base+4)+64` rewritten to `base+68` (`bv_omega`).
- **scaffold disjointness** (the four LBU/ADD/ADDI/BNE singletons vs the decoder `ofProg`) ← `CodeReq.Disjoint.singleton_ofProg`/`ofProg_singleton` + `ofProg_none_range_len` + `bv_omega`. (The `crDisjoint` tactic times out doing `whnf` on the opaque 16-instruction `ofProg`, so the explicit range route is used — every scaffold address lies outside `[base+4, base+64)`.)
- **`hback`** (`(base+76) + signExtend13 (-76) = base`) ← `signExtend13 (-76) = -76` by `decide`, then `bv_omega`.
- **`hne_lj*`** ← `bv_omega`.

Plus a concrete two-item `example` (`[0x01, 0x02]` at `base=0x1000`, region `0x2000`, 30 steps).

## Verification

- `lake build` green; `scripts/check-no-warnings.sh` → **0 warnings** under `EvmAsm/`.
[...]

PR #9007, opened 2026-06-17
feat(rv64): concrete flat RLP decoder program + spec (discharges decoderH)

dhsorens:
## Summary

Grounds the flat list-decode loop's **abstract** decoder hypothesis (`decoderH` in `FlatListLoop.lean`) in **real RV64 code**: it assembles one concrete program for the reconverged flat single-item decoder and proves its spec in exactly the shape the loop closure consumes.

`rlp_decode_single_item_reconverged_flat` (`UnifiedDecodeItemReconverge.lean`) already proves the `cpsTripleWithin 11` *given* concrete branch/jump offsets and the disjointness/subset side-conditions on a `cr`. This PR supplies a concrete layout that satisfies them.

## What's here (`EvmAsm/Rv64/RLP/FlatDecoderConcrete.lean`)

- **`flat_decoder_prog`** — one linear 16-instruction program: the 4-step phase-1 cascade (`BLTU` on `0x80`/`0xB8`/`0xC0`/`0xF8`) + the three flat-class phase-3 handlers (single-byte / short-string / short-list) + their reconvergence `JAL`s to a common `joinPC = base + 64`.
- **`flat_decoder_spec`** — for every flat prefix, `cpsTripleWithin 11 base (base+64) (CodeReq.ofProg base flat_decoder_prog) …`, leaving the cascade residue / payload length / payload pointer in `x10`/`x11`/`x13`. This is **exactly the `decoderH` shape** taken by `fll_loop_spec_within` / `fll_loop_n_spec_within`.

Proved by instantiating the reconverged lemma at the forced offsets (`off1=off2=28, off4=24, joff1=28, joff2=16, joff4=4`; `off3=0` is the longBytes branch — unreached for flat prefixes, so a don't-care) and discharging every side-condition:
- `htarget*`/`hjoin*` via `rv64_addr` (handles the `signExtend13`/`signExtend21` offset algebra);
- the six `hd_*` disjointness via the `crDisjoint` tactic;
- the three `hsub_*` subsets via `CodeReq.union_sub` + `ofProg_mono_sub`/`singleton_mono` (the small `flat_piece`/`flat_jal_piece` helpers).

## Verification

- `lake build` green; `scripts/check-no-warnings.sh` → **0 warnings** under `EvmAsm/`.
- `scripts/check-forbidden-tactics.sh` OK (no `native_decide` / `bv_decide`).
- `#print axioms flat_decoder_spec` → `[propext, Classical.choice, Quot.sound]` only.
[...]

PR #9029, opened 2026-06-18
RLP RV64: make the unified loop's decoder hypothesis dischargeable (step 6a)

dhsorens:
## Summary

Prerequisite refinement for the concrete unified RLP list decoder (step 6). Fixes a latent gap in the loop's abstract decoder hypothesis so that a real program can actually satisfy it.

`UnifiedDecoderH` (from #9028) quantified the 60-step region decoder over **every** byte index `i < bs.length` with no guard. But the region decoder (`rlp_decode_single_item_reconverged_all_region`) requires `rlpDecodeLongHypsRegion`, whose window part — `(off+1)+j < bs.length` for the long-form length bytes — holds **only at item-start offsets**. A payload byte mid-item can classify as a long prefix whose claimed length runs off the region end, so the guardless hypothesis is **unsatisfiable by any concrete program**, which would block step 6b.

**Fix** (`EvmAsm/Rv64/RLP/UnifiedListLoop.lean`):

- Add a per-index **`regionLongWindow`** predicate — the window-only half of `rlpDecodeLongHypsRegion` (the structural back-edge half is baked into the concrete decoder in 6b) — as a premise of `UnifiedDecoderH`.
- Discharge it inside `unified_loop_spec_within` at every item boundary (private `regionLongWindow_of_split`): flat heads reduce it to `True`; long heads' length bytes sit inside `encode head`, hence in-bounds (via `encode_long_length_eq` + the drop-length bound) and readable (via `hwin`).
- Local per-structure classification helpers `classify_bytes_long` / `classify_list_long`.

`unified_loop_n_spec_within` / `unified_loop_bridge` are unchanged in body — `UnifiedDecoderH` is used only within this file, so the change is local. The 5b bridge stays a valid theorem; this refines its decoder hypothesis to the strongest form a real program can provide.

## Context

Long-item list-loop arc, step 6a. Unblocks **6b** (`UnifiedDecoderConcrete.lean` — assemble `unified_decoder_prog` + `unified_decoder_spec` discharging the region decoder's ~30 hypotheses; analog of #9007) and **6c** (`UnifiedListLoopConcrete.lean` — the no-abstract-hypotheses end-to-end bridge; analog of #9008).

## Verification

- `scripts/check-forbidden-tactics.sh` OK; `scripts/check-no-warnings.sh` → 0 warnings under `EvmAsm/`.
- `#print axioms unified_loop_bridge` → `[propext, Classical.choice, Quot.sound]`; 0 `sorry`.

### DIFF
--- a/EvmAsm/Rv64/RLP/UnifiedListLoop.lean
+++ b/EvmAsm/Rv64/RLP/UnifiedListLoop.lean
@@ -15,6 +15,13 @@
   `64 * items.length` steps; the per-item byte stride varies by class.
 
   Counter on `x15`; the region decoder clobbers `x10`/`x12`/`x14` (scratch).
+
+  `UnifiedDecoderH` carries a per-index `regionLongWindow` guard (a long prefix's
+  length bytes must lie within the region) — the window-only half of
+  `rlpDecodeLongHypsRegion`. The closure discharges it at every item-start offset
+  (a long head's length bytes sit inside `encode head`), so a concrete decoder can
+  satisfy `UnifiedDecoderH`; without it the hypothesis would be unsatisfiable (a
+  mid-item byte can classify as a long prefix whose length runs off the region).
 -/
 
 import EvmAsm.Rv64.RLP.UnifiedListLoopBody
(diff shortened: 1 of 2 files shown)


pirapira:
Looks good. Merging.

PR #9033, opened 2026-06-18
RLP RV64: nested decode — descend-one-level kernel (Phase 5 step 1)

dhsorens:
## Summary

First step of **Phase 5 (nested/recursive RLP decode)** — the self-contained inductive kernel. The current decoder treats a `.list` item *opaquely* (computes its total length and skips its bytes via the stride); the pure spec *recurses* (`decodeAux` calls `decodeItems` on the payload). This PR establishes the fact that bridges the two: **the single-item decoder's window for a list item IS the sub-list payload**.

New file `EvmAsm/Rv64/RLP/NestedDescendOne.lean`:

- **`list_item_payload_window`** — for a `.list items` item at byte offset `off`, the decoder's `x13 = itemPtrRegion` (payload pointer) and `x11 = itemLenRegion` (payload length) point at exactly the `encode.encodeItems items` payload:
  ```
  ∃ payloadOff, itemPtrRegion … = regionBase + ofNat payloadOff
    ∧ itemLenRegion … = ofNat (encode.encodeItems items).length
    ∧ bs.drop payloadOff = encode.encodeItems items ++ tail
  ```
  The third conjunct is precisely the list loop's `hdrop` precondition, so a caller can frame the payload sub-window and run the loop to **descend one level**. This is the all-class analog, for the payload window, of the 5a stride lemma (`encode_head_eq_itemNextPtrRegion`) — short-list via `classifyPrefix_shortList_iff` + `encode_list_short`; long-list via `classifyPrefix_encode_head_long` + `encode_list_long` + `take_left'`/`drop_left'` + `fromBytesBE_toBytesBE`.
- **`decode_list_descend`** — top-level nested round-trip, plus concrete nested `example`s (including a list containing an empty list and a sub-list).

## Context

The pure spec already handles arbitrary nesting (mutual `decodeAux`/`decodeItems` recursion, fuel-bounded; `decode_encode_mutual` proves nested round-trips at any depth). This kernel is the operational↔recursive-branch link a recursive-descent decoder rests on.

Deferred to later steps (per the agreed "kernel first" strategy): the **operational descent assembly** (wiring the decoder window into the loop via `cpsTripleWithin_seq`, step 2) and the **length-driven loop** needed for genuine recursion (step 3). Note the impedance mismatch the kernel surfaced — the existing loop is *count-driven* (`x15 = items.length`) while real RLP decode is *length-driven*; resolving that is step 3's crux, after which we choose bounded fixed-depth composition (for the STF's block/tx structure) vs. the general explicit-stack recursive decoder.

## Verification

- `scripts/check-forbidden-tactics.sh` OK (no `bv_decide`/`native_decide`); `scripts/check-no-warnings.sh` → 0 warnings under `EvmAsm/`.
- `#print axioms list_item_payload_window` / `decode_list_descend` → `[propext, Classical.choice, Quot.sound]`; 0 `sorry`.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -67,6 +67,7 @@ import EvmAsm.Rv64.RLP.UnifiedItemStride
 import EvmAsm.Rv64.RLP.UnifiedListLoop
 import EvmAsm.Rv64.RLP.UnifiedDecoderConcrete
 import EvmAsm.Rv64.RLP.UnifiedListLoopConcrete
+import EvmAsm.Rv64.RLP.NestedDescendOne
 import EvmAsm.Rv64.RLP.UnifiedDecodeItem
 import EvmAsm.Rv64.RLP.UnifiedDecodeItemReconverge
 import EvmAsm.Rv64.RLP.UnifiedDecodeItemReconvergeAll
--- a/EvmAsm/Rv64/RLP/NestedDescendOne.lean
+++ b/EvmAsm/Rv64/RLP/NestedDescendOne.lean
@@ -1,0 +1,129 @@
+/-
+  EvmAsm.Rv64.RLP.NestedDescendOne
+
+  EL.3 / Phase 5 (nested decode) — the descend-one-level kernel. The single-item
+  decoder, applied to a `.list` item, leaves `x13 = itemPtrRegion` (payload start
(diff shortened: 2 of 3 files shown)


pirapira:
Looks good to me. Merging.

Maybe later we want to expose the value of `∃ payloadOff`, would be a small adjustment even if needed.

PR #9035, opened 2026-06-18
RLP RV64: length-driven list-loop body (Phase 5 step 2)

dhsorens:
## Summary

The body of a **length-driven** list-decode loop — the foundation for nested RLP decode.

Step 1 (#9033) showed the count-driven loop is a dead end for nesting: the single-item decoder yields a payload *byte-length* (`x11`), not an *item count*, and `items.length` is symbolic — no instruction can load it. The pure `decodeItems` is length-driven (recurse until the byte list is empty). So the foundation is a loop that stops when the data pointer reaches a precomputed **end pointer**.

New file `EvmAsm/Rv64/RLP/UnifiedLenLoopBody.lean`:

- **`unified_lenloop_body_spec_within`** — one iteration as `cpsBranchWithin 63`:
  ```
  lbase      LBU  x5, x13, 0
  lbase+4    < region all-class decoder, 60 steps >
  joinPC     ADD  x13, x13, x11        ; x13 := payloadPtr + payloadLen
  joinPC+4   BNE  x13, x15, back       ; loop while x13 ≠ endPtr  (taken → lbase)
  ```
  The guard compares the advanced pointer to the loop-invariant end pointer `x15 = endPtr` — **no counter**, one fewer instruction than the count-driven body. Taken (`itemNextPtrRegion ≠ endPtr`) loops back to `lbase`; fall-through (`= endPtr`) exits at `joinPC+8`. With no counter, `x15` is free to hold `endPtr`; the decoder never touches it, so it survives the 60-step decode framed.
- **`unified_lenloop_body_post`** (+ `_unfold`/`_pure`).

A faithful mirror of the count-driven `unified_body_spec_within` (#9026) with the guard swapped from `ADDI x15 -1; BNE x15 x0` to `BNE x13 x15`. Built on the first attempt.

## Context

Phase 5 (nested decode), step 2. Next: **step 3 — the length-driven closure** (induct over items; prove `x13` reaches `endPtr` exactly after the last item — the new lemma is pointer-≠-end via address injectivity from no-wraparound; bridge to the length-driven pure `decodeItems`), then **step 4** — compose with #9033's window for genuine one-level descent, then arbitrary depth.

## Verification

- `lake build` clean (module + umbrella `EvmAsm.Rv64.RLP`).
- `scripts/check-forbidden-tactics.sh` OK (no `bv_decide`/`native_decide`); `scripts/check-no-warnings.sh` → 0 warnings under `EvmAsm/`.
[...]

PR #9026, opened 2026-06-18
feat(rv64): unified (long-capable) list-loop body (long-loop arc, step 4)

dhsorens:
## Summary

Step 4 of the long-item RLP list-loop arc: the **unified loop body** — one iteration that decodes **any** item (all 5 classes) via the region all-class decoder, then advances. The long-capable analog of the flat `fll_body_spec_within`.

```
loop_top (lbase):  LBU  x5, x13, 0          ; prefix bs[i] → x5
                   < region all-class decoder: lbase+4 .. joinPC, 60 steps >
joinPC:            ADD  x13, x13, x11        ; x13 := payloadPtr + payloadLen
(joinPC+4):        ADDI x15, x15, -1         ; item counter -= 1
(joinPC+8):        BNE  x15, x0, back        ; loop if counter ≠ 0 (taken → lbase)
```

**Stacked on #9025** (region decoder); targets that branch, auto-retargets to `main` on merge.

## What's here (`EvmAsm/Rv64/RLP/UnifiedListLoopBody.lean`)

- **`unified_body_spec_within`** — `cpsBranchWithin 64 lbase` (1 LBU + 60 decoder + ADD + ADDI + BNE), taken → `lbase`, fall-through → `joinPC+12`.
- **`unified_body_post`** (+ `_unfold`/`_pure`) and **`itemNextPtrRegion := itemPtrRegion + itemLenRegion`** (the `ADD x13,x13,x11` result = the next item's start pointer).

**Key difference from the flat body:** the item counter is **x15**, because the region decoder clobbers x14 (its length-read counter) and x12 (last length byte). So x10/x12/x14 are decoder scratch, and the loop bookkeeping lives in x15. The decoder is supplied as an **opaque** `cpsTripleWithin 60` triple (region-decoder-shaped, with `bytesRegion` in its pre/post), keeping the body decoder-agnostic — the concrete `rlp_decode_single_item_reconverged_all_region` discharges it in the end-to-end step (mirroring how flat #9007/#9008 wired the flat decoder).

## Verification

- `lake build` green; `scripts/check-no-warnings.sh` → **0 warnings** under `EvmAsm/`.
- `scripts/check-forbidden-tactics.sh` OK (no `native_decide` / `bv_decide`).
- `#print axioms unified_body_spec_within` → `[propext, Classical.choice, Quot.sound]` only.
- 0 `sorry`/`admit`.

## Next (step 5)

[...]

PR #9037, opened 2026-06-18
RLP RV64: length-driven list-loop closure + bridge (Phase 5 step 3)

dhsorens:
## Summary

The length-driven list-decode loop closure — the foundation for nested RLP decode without an item count. Builds on the step-2 body (#9035) the same way the count-driven closure (#9028) built on its body (#9026).

New file `EvmAsm/Rv64/RLP/UnifiedLenLoop.lean`:

- **`unified_lenloop_spec_within`** — inducts over the items, applying the body per item. The loop runs until the data pointer `x13` reaches the invariant end pointer `x15 = endPtr = regionBase + ofNat (O + (encode.encodeItems items).length)`, decoding all items in `63 * items.length` steps. **No item count** — `endPtr` is a byte offset, which is operationally computable (unlike a count). The new lemma **`region_offset_ne`** (distinct in-bounds byte offsets ⇒ distinct addresses, via `bv_omega`) discharges `itemNextPtrRegion ≠ endPtr` for every intermediate item (so the loop continues); after the last item `x13 = endPtr` and it falls through.
- **`unified_lenloop_n_spec_within`** — offset-0 entry (`x15 = regionBase + ofNat (encode.encodeItems items).length`).
- **`unified_lenloop_bridge`** — conjoins the pure `decodeItems (2*len+1) … = some (items, [])` round-trip via `decode_encode_mutual.2`.

A faithful mirror of the count-driven closure (#9028) with the termination swapped from "counter hits 0" to "pointer hits `endPtr`", plus the one new injectivity lemma and the `endPtr`-invariant threading.

Also **exposes `regionLongWindow_of_split`** (previously `private` in `UnifiedListLoop.lean`) — both closures discharge the same per-item region-window obligation; logic unchanged, just `private` removed.

## Context

Phase 5 (nested decode), step 3. Next: **step 4** — concrete length-driven program, then compose with the descend-one-level window (#9033): header decode → set `x15 := payload end` (ptr+len) → run the length-driven loop → genuine one-level descent matching `decode (.list items)`; then arbitrary depth.

## Verification

- `scripts/check-forbidden-tactics.sh` OK (no `bv_decide`/`native_decide`); `scripts/check-no-warnings.sh` → 0 warnings under `EvmAsm/`.
- `#print axioms unified_lenloop_bridge` / `unified_lenloop_spec_within` → `[propext, Classical.choice, Quot.sound]`; 0 `sorry`.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -64,6 +64,7 @@ import EvmAsm.Rv64.RLP.Phase1LongFullRegion
 import EvmAsm.Rv64.RLP.UnifiedDecodeItemReconvergeAllRegion
 import EvmAsm.Rv64.RLP.UnifiedListLoopBody
 import EvmAsm.Rv64.RLP.UnifiedLenLoopBody
+import EvmAsm.Rv64.RLP.UnifiedLenLoop
 import EvmAsm.Rv64.RLP.UnifiedItemStride
 import EvmAsm.Rv64.RLP.UnifiedListLoop
 import EvmAsm.Rv64.RLP.UnifiedDecoderConcrete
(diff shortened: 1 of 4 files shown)


pirapira:
Looks good. Merging.

PR #9044, opened 2026-06-18
feat(rv64/rlp): concrete top-level RLP list descent (Phase 5, step 4b)

dhsorens:
## Summary

Adds **`unified_list_descend_concrete_bridge`** (`EvmAsm/Rv64/RLP/UnifiedListDescendConcrete.lean`) — the genuine top-level RLP list decoder. A real RV64 program decodes a **complete RLP list value** `encode (.list items)` by *descending* the list header into its payload, with **no abstract hypotheses** remaining, and proves it coincides with the pure `decode`.

This is the realistic top-level decode (the input *is* a list value, e.g. a block). It composes three concrete pieces, completing Phase 5 step 4b (building on the length-driven loop from step 4a, #9039):

```
base      LBU x5,x13,0 ++ unified_decoder_prog   ; decode the .list header → x13=payloadPtr, x11=payloadLen
base+148  ADD x15,x13,x11                         ; x15 := endPtr = payloadPtr + payloadLen
base+152  LBU x5,x13,0 ++ unified_decoder_prog
          ++ [ADD x13,x13,x11, BNE x13,x15,-152]  ; length-driven payload loop
base+308  (exit)
```

In `62 + 63 * items.length` steps the data pointer reaches `endPtr`, and the operational run coincides with the pure `decode (encode (.list items)) = some (.list items, [])`.

## How it fits together

- **Header → loop bridge**: `list_item_payload_window` (#9033) with `tail = []` — top-level ⇒ the payload is the exact suffix `bs.drop payloadOff = encode.encodeItems items`, exactly the loop's precondition.
- **Header window obligation**: discharged by `regionLongWindow_of_split` (head `= .list items`, tail `= []`).
- **Payload**: `unified_lenloop_spec_within` at offset `payloadOff`; `UnifiedDecoderH` discharged by `unified_decoder_spec`.
- **Pure half**: `decode_encode`.

## Notable elaboration points

- The two `unified_decoder_prog` copies' **cross-disjointness** is proven manually (`ofProg_none_range_len` + `Disjoint.*` + a new private `ofProg_disjoint_ofProg` helper) because `crDisjoint` over two `ofProg`s times out.
- The heavy `unified_lenloop_spec_within` application is **extracted into a private `descend_loop_triple`** so it elaborates in a clean local context — applying it inside the main theorem's large context provokes an unbounded `whnf`.

## Verification

[...]

PR #9051, opened 2026-06-19
feat(rv64/rlp): tail-general loop closure + interior RLP list descent (Phase 5)

dhsorens:
## Summary

Generalizes the length-driven RLP list decoder to tolerate **trailing bytes** after the items, unlocking descent into **interior sub-lists** — a `.list` value embedded in a larger buffer with sibling bytes following it (the recursive step toward arbitrary depth / the STF block & tx structure).

The headline result, `unified_list_descend_concrete_bridge`, now takes a trailing `tail : List Byte`: the real RV64 program decodes a `.list` value at the front of `encode (.list items) ++ tail`, stopping at the sub-list boundary and **leaving `tail` unread**, in `62 + 63 * items.length` steps, and coincides with the pure `decode (encode (.list items) ++ tail) = some (.list items, tail)`. `tail = []` recovers the top-level case (#9044).

> **Stacked on #9044** (`feat/rv64-rlp-list-descend-concrete`), which is not yet merged — this PR is based on that branch so the diff shows only the interior-descent work. Once #9044 lands, I'll rebase onto `main` and retarget.

## How it works

The operational core was **already tail-agnostic** — `regionLongWindow`, the loop body, and `unified_decoder_spec` only read *within* each item. So the generalization is mostly threading one parameter:

- **Glue lemmas** `encode_head_eq_itemNextPtrRegion`, `regionLongWindow_of_split`, and the private `long_lenBytes_in_region` now take an arbitrary byte suffix `rest : List Byte` instead of `encode.encodeItems tail`. Their proofs only use that `encode head` is a *prefix*, so the bodies are unchanged; count-driven callers pass `encode.encodeItems tail`.
- **`unified_lenloop_spec_within`** carries a `btail : List Byte`; its precondition becomes `bs.drop O = encode.encodeItems items ++ btail`. End pointer, post, code requirement, and step count are **unchanged** (the loop stops at the items boundary). `unified_lenloop_n_spec_within` passes `[]`.
- The buffer-vs-value first byte is bridged by `List.getElem_append_left` (`(encode (.list items) ++ tail)[0] = (encode (.list items))[0]`).
- **Pure half**: the already-existing `decode_encode_append` (`decode (encode item ++ rest) = some (item, rest)`) — no new pure lemma needed.

## Verification

- `#print axioms unified_list_descend_concrete_bridge` → `[propext, Classical.choice, Quot.sound]` (classical only); 0 sorry.
- `scripts/check-forbidden-tactics.sh` OK; **no `maxHeartbeats` override** (builds under the default budget).
- Concrete `example`s for `tail = []` and `tail = [0xFF]` (`[0xc2,0x01,0x02,0xFF]` ⇒ `(.list …, [0xFF])`).

### DIFF
--- a/EvmAsm/Rv64/RLP/UnifiedItemStride.lean
+++ b/EvmAsm/Rv64/RLP/UnifiedItemStride.lean
@@ -40,9 +40,9 @@ theorem flat_or_long (item : RLPItem) : isFlatItem item ∨ isLongItem item := b
 
 /-- The `lenOfLen` length bytes the decoder reads from the region (starting at
     `off+1`) are exactly the encoding's `Nat.toBytesBE payloadCount`. -/
-private theorem long_lenBytes_in_region (head : RLPItem) (tail : List RLPItem)
+private theorem long_lenBytes_in_region (head : RLPItem) (rest : List Byte)
     (bs : List Byte) (off : Nat) (hlong : isLongItem head)
-    (hdrop : bs.drop off = encode head ++ encode.encodeItems tail) :
+    (hdrop : bs.drop off = encode head ++ rest) :
     (bs.drop (off + 1)).take (itemLenOfLen head) = Nat.toBytesBE (itemPayloadCount head) := by
   have hdd : bs.drop (off + 1) = (bs.drop off).drop 1 := by
     rw [List.drop_drop]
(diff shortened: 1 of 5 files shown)


pirapira:
This looks great. Merging.

PR #9053, opened 2026-06-19
feat(rv64/rlp): offset-general RLP list descent (Phase 5, step 6a)

dhsorens:
## Summary

Generalizes the concrete RLP list descent to read at an arbitrary byte **offset `O`** into the buffer — the building block that makes **nested descent composable** (toward depth-2, then the STF block/header/tx decoders).

`unified_list_descend_concrete_bridge_at` decodes a `.list items` value sitting at byte offset `O` of a buffer `bs` (with `bs.drop O = encode (.list items) ++ tail`): `x13` enters at `regionBase + ofNat O`, the header descends into the payload, the length-driven loop runs to the sub-list boundary, in `62 + 63 * items.length` steps — coinciding with the pure `decode (bs.drop O) = some (.list items, tail)`.

> **Stacked on #9044** (`feat/rv64-rlp-list-descend-concrete`, which carries the merged step-5 work and is pending merge to `main`). Based on that branch so the diff shows only the offset generalization; will rebase onto `main` once #9044 lands.

## Why offset-general (not re-anchored)

`bytesRegion` (`MemRegion.lean`) is **dword-granular, anchored at a single 8-byte-aligned base** — byte `i` lives in dword `i/8`, read via `byteOffset (regionBase + ofNat i) = i % 8`. So the inner decode of a sub-list at the (generally unaligned) payload pointer **cannot re-anchor or split** the region there. Instead we descend at an **offset** into the same aligned `regionBase`.

The underlying pieces were already offset-general (`unified_decoder_spec`, `unified_lenloop_spec_within`, `list_item_payload_window`, `regionLongWindow_of_split`); `descend_loop_triple` is generalized to an arbitrary buffer here, and only the descent *wrapper* needed the offset. The buffer-vs-value first byte is bridged by `List.getElem_drop` + `List.getElem_append_left`.

`unified_list_descend_concrete_bridge` is now the `O = 0` corollary (step-5 examples unchanged).

## Verification

- `#print axioms unified_list_descend_concrete_bridge_at` → `[propext, Classical.choice, Quot.sound]`; 0 sorry.
- Concrete examples: `O = 0` (top-level + `tail = [0xFF]` interior) and `O = 2` (list value at offset 2 of `[0xFF,0xFF,0xc1,0x01]`).

### DIFF
--- a/EvmAsm/Rv64/RLP/UnifiedListDescendConcrete.lean
+++ b/EvmAsm/Rv64/RLP/UnifiedListDescendConcrete.lean
@@ -63,14 +63,14 @@ private theorem ofProg_disjoint_ofProg (b1 b2 : Word) (p1 p2 : List Instr) (n1 n
     context provokes an unbounded `whnf`. The scratch registers `x5/x10/x11/x12/x14`
     are arbitrary (the loop only reads `x13`/`x15`). -/
 private theorem descend_loop_triple
-    (base regionBase : Word) (items : List RLPItem) (tail : List Byte) (payloadOff : Nat)
-    (v5' v10' v11' v12' v14' : Word)
+    (base regionBase : Word) (items : List RLPItem) (bs : List Byte) (tail : List Byte)
+    (payloadOff : Nat) (v5' v10' v11' v12' v14' : Word)
     (halign : regionBase.toNat % 8 = 0)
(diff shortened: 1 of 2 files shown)


pirapira:
Looks great. Merging.

PR #9054, opened 2026-06-19
feat(rv64/rlp): depth-2 nested RLP list descent (Phase 5, step 6b)

dhsorens:
## Summary

A concrete RV64 program that **descends two levels** of nested RLP — the outer list header, then the first sub-list item (itself a `.list`) — fully decoding the inner list and coinciding with the pure nested `decode`. The recursive step toward the fixed-schema STF block/header/tx decoders.

- **`unified_list_header_descend`** — the reusable header-descend primitive: `LBU + unified_decoder_prog` at offset `O` leaves `x13 = itemPtrRegion` (payload pointer) and `x11 = itemLenRegion` (payload length). Used by every list level (incl. the STF schema decoders).
- **`unified_list_descend_nested_bridge`** — for an outer list whose head is a list (`.list (.list innerItems :: rest)`) at the front of `bs = encode (.list (.list innerItems :: rest)) ++ outerTail`: descends the outer header (offset 0) to its payload pointer (offset `payloadOff`), then descends the inner `.list innerItems` there via `unified_list_descend_concrete_bridge_at` — in `123 + 63 * innerItems.length` steps — coinciding with the pure nested `decode` (both the outer value and the inner sub-list).

> **Stacked on #9053** (offset-general descent), which is stacked on #9044. Based on `feat/rv64-rlp-nested-descend` so the diff shows only the depth-2 work; will retarget/rebase as the stack lands.

## Two notable mechanics

- **Composition by offset, not re-anchoring**: `bytesRegion` is dword-granular anchored at one aligned base, so the inner sub-list (at the generally-unaligned payload pointer) is decoded by reading at an *offset* into the same region — which is exactly what `unified_list_descend_concrete_bridge_at` (#9053) provides.
- **Address normalization**: the inner program's `(base+148)+k` addresses are `rw`-normalized to `base+N` before `cpsTripleWithin_seq`, so the composition unifies syntactically — avoiding a variable-base `whnf` blowup. Cross-decoder disjointness uses the now-public `ofProg_disjoint_ofProg`.

## Verification

- `#print axioms unified_list_descend_nested_bridge` → `[propext, Classical.choice, Quot.sound]`; 0 sorry.
- Concrete depth-2 example: `[[0x01, 0x02], 0x03]` (`encode = [0xc4, 0xc2, 0x01, 0x02, 0x03]`).

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -71,6 +71,7 @@ import EvmAsm.Rv64.RLP.UnifiedListLoop
 import EvmAsm.Rv64.RLP.UnifiedDecoderConcrete
 import EvmAsm.Rv64.RLP.UnifiedListLoopConcrete
 import EvmAsm.Rv64.RLP.UnifiedListDescendConcrete
+import EvmAsm.Rv64.RLP.UnifiedListDescendNested
 import EvmAsm.Rv64.RLP.NestedDescendOne
 import EvmAsm.Rv64.RLP.UnifiedDecodeItem
 import EvmAsm.Rv64.RLP.UnifiedDecodeItemReconverge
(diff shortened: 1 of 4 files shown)


pirapira:
Looks good to me. Merging.

PR #9055, opened 2026-06-19
feat(rv64/rlp): regOwn-re-entry RLP list descent (Phase 5, step 7a)

dhsorens:
## Summary

Adds **`unified_list_descend_concrete_bridge_at_regOwn`** — the **chainable** RLP list descent, the building block for sequential sibling descent (and thus the fixed-schema STF block decoder: descend `[header, txs, ommers]` in turn).

It restates the offset-general descent (#9053) so its **pre is itself a `unified_lenloop_post`** (at offset `O`) and its **post the next `unified_lenloop_post`** (at `O + (encode (.list items)).length`). Because a descent's post already abstracts the scratch registers to `regOwn` and leaves `x13`/`x15` at the *next sibling*, this variant lets one descent feed **directly** into the next with a syntactic `unified_lenloop_post` match — no re-specialization at the call site.

(Branches cleanly off `main` now that #9044/#9053/#9054 have landed.)

## How it's proved

Derived from `unified_list_descend_concrete_bridge_at` by consuming the 5 owned scratch registers (`x5,x10,x11,x12,x14`) via `cpsTripleWithin_of_forall_regIs_to_regOwn` — the established Evm64 SDiv/SMod idiom. Two details that made it go through:
- Each peel is pinned with explicit `(r := …) (P := …)` so `xperm_hyp` can resolve the reshape (the registers are *interspersed* with `x0↦0`, `x13↦ptr`, `bytesRegion`, so the rightmost-peel rule needs an explicit permutation target).
- `rw [unified_lenloop_post_unfold]` unfolds **only the pre** — the post carries a distinct `endPtr`, so it stays folded and matches `…_bridge_at`'s post.

## Why (sequential sibling descent)

The forward direction is the only one that works: a `regOwn` post **cannot** entail a specific `↦ᵣ` pre (`seq_perm` is the wrong way), so chaining descents genuinely requires a `regOwn`-pre variant. With both intermediate states now `unified_lenloop_post`, the upcoming two-sibling `seq` is a syntactic match.

## Verification

- `#print axioms unified_list_descend_concrete_bridge_at_regOwn` → `[propext, Classical.choice, Quot.sound]`; 0 sorry.

Follow-up (step 7b): the two-sibling descent composing `…_bridge_at` ⨾ `…_at_regOwn`, then N-sibling walks → concrete header/tx schema decoders.

### DIFF
--- a/EvmAsm/Rv64/RLP/UnifiedListDescendConcrete.lean
+++ b/EvmAsm/Rv64/RLP/UnifiedListDescendConcrete.lean
@@ -416,6 +416,70 @@ theorem unified_list_header_descend
         unified_decoder_prog_length (by intro k hk; bv_omega)))
     s_lbu s_dec
 
+set_option maxRecDepth 8000 in
+/-- **`regOwn`-re-entry descent (chainable).** The offset-general descent restated so
+    its PRE is itself a `unified_lenloop_post` (at `O`) and its POST is the next
+    `unified_lenloop_post` (at `O + (encode (.list items)).length`). A descent's post
+    already abstracts the scratch registers to `regOwn` and leaves `x13`/`x15` at the
+    next sibling, so this lets one descent feed DIRECTLY into the next (sequential
+    sibling descent) with a syntactic `unified_lenloop_post` match. Derived from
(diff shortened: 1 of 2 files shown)


pirapira:
Looks good to me. Merging

PR #9056, opened 2026-06-19
feat(rv64/rlp): two-sibling sequential RLP list descent (Phase 5, step 7b)

dhsorens:
## Summary

Adds **`unified_list_descend_two_siblings_bridge`** — a concrete RV64 program that descends **two successive sibling sub-lists** of a buffer (`.list aItems` at offset `O`, then `.list bItems` at the next-sibling offset `O + (encode (.list aItems)).length`), coinciding with the pure `decode` of each. This is the pattern the fixed-schema STF block decoder repeats over `[header, txs, ommers]`.

> **Stacked on #9055** (the `regOwn`-re-entry variant), which is still open. Based on `feat/rv64-rlp-sibling-descend`; will retarget to `main` once #9055 lands.

## How the chain works (no glue)

The two descents compose with **zero re-specialization**: `unified_list_descend_concrete_bridge_at`'s post and `unified_list_descend_concrete_bridge_at_regOwn`'s pre are both `unified_lenloop_post`, and the first descent already leaves `x13` exactly at the next sibling (the **free stride**). So `cpsTripleWithin_seq` matches the intermediate state **syntactically** — the only fixup is `rw`-normalizing B's exit `base+308+308 → base+616`.

## Reusable CR machinery

Adds `descendCR b` (the descent's 7-leaf code requirement) plus:
- **`descendCR_none`** — a descent CR is `none` at any address outside its 77 instruction slots (`b + 4*k`, `k < 77`).
- **`descend_cr_disjoint`** — two non-overlapping descents (`base2 ≥ base1 + 308`) have disjoint CRs.

This makes the cross-descent disjointness an **~8-line range argument** (à la `ofProg_disjoint_ofProg`) instead of a 7×7 `Disjoint.*` cascade — and it's reused by the upcoming N-sibling block walk.

## Verification

- `#print axioms unified_list_descend_two_siblings_bridge` → `[propext, Classical.choice, Quot.sound]`; 0 sorry.
- Concrete example: `[0xc1, 0x01, 0xc1, 0x02]` (two single-item lists) → descend `[1]`, stride, descend `[2]`.

Follow-up: N-sibling unrolled walk (block = 3 sub-lists) via repeated `…_at_regOwn` + `descend_cr_disjoint`, then leaf-field reads → concrete header/tx schema decoders.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -72,6 +72,7 @@ import EvmAsm.Rv64.RLP.UnifiedDecoderConcrete
 import EvmAsm.Rv64.RLP.UnifiedListLoopConcrete
 import EvmAsm.Rv64.RLP.UnifiedListDescendConcrete
 import EvmAsm.Rv64.RLP.UnifiedListDescendNested
+import EvmAsm.Rv64.RLP.UnifiedListDescendSiblings
 import EvmAsm.Rv64.RLP.NestedDescendOne
 import EvmAsm.Rv64.RLP.UnifiedDecodeItem
 import EvmAsm.Rv64.RLP.UnifiedDecodeItemReconverge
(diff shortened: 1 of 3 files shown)


pirapira:
Looks good. Merging.

PR #9057, opened 2026-06-19
feat(rv64/rlp): leaf-field scalar value read (Phase 5, step 8)

dhsorens:
## Summary

Adds **`unified_field_scalar_read`** — the first **value-extraction** step in the RLP decoder (every prior step produced only *framing*: pointers and lengths).

From the single-item decoder's output convention — `x13 = regionBase + ofNat off` (payload pointer), `x11 = ofNat n` (payload length, `1 ≤ n ≤ 8`) — it reads the `n` payload bytes **big-endian** into `x11 = Nat.fromBytesBE ((bs.drop off).take n)` (the field's scalar value) and advances `x13` to `off + n` (the next field), in `2 + 6·n` steps. This is the core of the per-field walk the fixed-schema STF header/transaction decoders need.

(Branches cleanly off `main` — the #9044→#9056 stack has all landed.)

## How it works

Reuses the Phase-2 region big-endian loop (`rlp_phase2_long_loop_region_n_spec_within`) wholesale. The only new code is a **2-instruction impedance-match glue**, because the decoder leaves the length in `x11` but the loop wants the count in `x14` and a zeroed accumulator:

```
ADDI x14, x11, 0    ; x14 := payload length (loop count)
ADDI x11, x0, 0     ; x11 := 0 (BE accumulator)
< 6-instr BE loop, n iterations >
```

Composed via `cpsTripleWithin_seq`; the glue ⊥ loop disjointness is two `singleton_ofProg` (`ofProg_none_range_len`) facts.

## Verification

- `#print axioms unified_field_scalar_read` → `[propext, Classical.choice, Quot.sound]`; 0 sorry.
- Concrete example: read `[0x01, 0x02, 0x03]` → `x11 = 0x010203`.

## Follow-ups
- **Full scalar field decode**: `unified_decoder_spec` ⨾ this read, coinciding with the pure `decodeScalar` (needs a small `.bytes`-payload-window lemma).
- **Field walk → schema decoders**: per-field (decode + value-read + stride) over a fixed schema → STF transaction (9 fields) / header (~19 fields) decoders. Wide scalars (uint256), byte-arrays (address/hash), and zero-length scalars need multi-word / byte-copy / branch variants.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -73,6 +73,7 @@ import EvmAsm.Rv64.RLP.UnifiedListLoopConcrete
 import EvmAsm.Rv64.RLP.UnifiedListDescendConcrete
 import EvmAsm.Rv64.RLP.UnifiedListDescendNested
 import EvmAsm.Rv64.RLP.UnifiedListDescendSiblings
+import EvmAsm.Rv64.RLP.UnifiedFieldScalarRead
 import EvmAsm.Rv64.RLP.NestedDescendOne
 import EvmAsm.Rv64.RLP.UnifiedDecodeItem
 import EvmAsm.Rv64.RLP.UnifiedDecodeItemReconverge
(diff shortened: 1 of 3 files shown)


pirapira:
Looks good. Merging.

PR #9058, opened 2026-06-19
feat(rv64/rlp): full scalar field decode (Phase 5, step 9)

dhsorens:
## Summary

Composes the single-item decoder with the big-endian value read so a `.bytes` **scalar field is decoded end-to-end from its offset**, coinciding with the pure `decodeScalar`. This is the per-field unit the fixed-schema STF header/transaction decoders walk.

`unified_scalar_field_decode`: from `x13 = regionBase + ofNat O` (a `.bytes data` field at offset `O`, `1 ≤ data.length ≤ 8`), the program decodes the item header then reads the payload big-endian, leaving `x11 = Nat.fromBytesBE data` (the field value) and `x13` at the **next field** (`O + (encode (.bytes data)).length` — the free stride), in `63 + 6·data.length` steps — and coincides with `decodeScalar (bs.drop O) = some (Nat.fromBytesBE data, tail)`.

> **Stacked on #9057** (the leaf-field value read), which is still open. Based on `feat/rv64-rlp-field-scalar-read`; will retarget to `main` once #9057 lands.

## Pieces

- **`bytes_item_payload_window`** — the `.bytes` analog of `list_item_payload_window`: for a short `.bytes data` item, the decoder's window (`itemPtrRegion`/`itemLenRegion`) points exactly at `data` (singleByte + shortBytes cases, via a uniform `encodeBytes_shortBytes_form` helper).
- **`unified_scalar_field_decode`** — composes `unified_list_header_descend` (the general LBU+decoder item-header step) ⨾ `unified_field_scalar_read`. Clobbered scratch (`x5/x10/x12/x14`) is abstracted to `regOwn` in the post (chainable).

Two integration details (both already seen in the descents): the value-read's `(base+148)+k` addresses are `rw`-normalized to `base+N`, and the `frameR`'d post is a `(6-group) ** (3-group)`, so the `regOwn` conversion is a per-group `sepConj_mono`.

## Verification

- `#print axioms unified_scalar_field_decode` → `[propext, Classical.choice, Quot.sound]`; 0 sorry.
- Concrete example: scalar `0x2a` (= 42) at offset 0 → `x11 = 0x2a`, `decodeScalar [0x2a] = some (42, [])`.

## Follow-ups
Per-field **walk** (chain across a fixed schema, threading `regOwn` scratch — needs a `regOwn`-pre field-decode variant) → concrete transaction/header schema decoders producing the `BlockInput`. Wide scalars (uint256), byte-arrays (address/hash), and zero-length scalars need multi-word / byte-copy / branch variants.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -74,6 +74,7 @@ import EvmAsm.Rv64.RLP.UnifiedListDescendConcrete
 import EvmAsm.Rv64.RLP.UnifiedListDescendNested
 import EvmAsm.Rv64.RLP.UnifiedListDescendSiblings
 import EvmAsm.Rv64.RLP.UnifiedFieldScalarRead
+import EvmAsm.Rv64.RLP.UnifiedScalarFieldDecode
 import EvmAsm.Rv64.RLP.NestedDescendOne
 import EvmAsm.Rv64.RLP.UnifiedDecodeItem
 import EvmAsm.Rv64.RLP.UnifiedDecodeItemReconverge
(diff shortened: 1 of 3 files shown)


pirapira:
Looks good. Merging.

PR #9088, opened 2026-06-22
feat(rv64/rlp): decode-and-store a scalar field (Phase 5, step 10)

dhsorens:
## Summary

Adds `unified_scalar_field_decode_and_store` (new file `EvmAsm/Rv64/RLP/UnifiedScalarFieldStore.lean`): the **persistence** step for the fixed-schema RLP field walk.

It composes the end-to-end scalar field decode (`unified_scalar_field_decode`, #9058) with a single `SD rOut, x11, offset`, so the decoded value `Nat.fromBytesBE data` is written to the output cell `outBase + offset` (u64 LE — exactly matching the STF output struct's scalar slots, e.g. nonce/gas_limit/v in `EvmAsm/Stateless/Transaction/Decode.lean`) while `x13` advances to the next field, in `64 + 6*data.length` steps. The triple rides along the pure coincidence `decodeScalar (bs.drop O) = some (Nat.fromBytesBE data, tail)`.

### Why
A multi-field schema walk decodes each field into `x11`, but the *next* field's decode clobbers `x11` — so every value must be persisted before moving on. This is the atomic, reusable unit the legacy-tx (9-field) / header decoders will repeat.

### Proof notes
- The output pointer register and cell are framed alongside the decode via `cpsTripleWithin_frameR` — **no `rOut` distinctness hypotheses** are needed (the frame rule holds regardless; a clashing `rOut` just makes the precondition unsatisfiable). The caller instantiates `rOut` with a callee-saved register the decoder never touches.
- The store CR (`SD@base+180`) is shown disjoint from the decoder/read CR (which spans `base .. base+176`) via the usual `CodeReq.Disjoint.*` + `ofProg_none_range_len` address algebra.
- No new pure RLP lemmas; the `decodeScalar` fact comes from `unified_scalar_field_decode`.

## Verification
- `lake build EvmAsm.Rv64.RLP.UnifiedScalarFieldStore` + umbrella `EvmAsm.Rv64.RLP` — clean (1.6s).
- `scripts/check-forbidden-tactics.sh` OK; no `maxHeartbeats` override; `check-no-warnings.sh` → 0 warnings under `EvmAsm/`.
- `#print axioms unified_scalar_field_decode_and_store` → `[propext, Classical.choice, Quot.sound]` (classical only); 0 sorry.
- Concrete `example`: decode `0x2a` (= 42) and store it to `0x3000` via `x18`.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -75,6 +75,7 @@ import EvmAsm.Rv64.RLP.UnifiedListDescendNested
 import EvmAsm.Rv64.RLP.UnifiedListDescendSiblings
 import EvmAsm.Rv64.RLP.UnifiedFieldScalarRead
 import EvmAsm.Rv64.RLP.UnifiedScalarFieldDecode
+import EvmAsm.Rv64.RLP.UnifiedScalarFieldStore
 import EvmAsm.Rv64.RLP.NestedDescendOne
 import EvmAsm.Rv64.RLP.UnifiedDecodeItem
 import EvmAsm.Rv64.RLP.UnifiedDecodeItemReconverge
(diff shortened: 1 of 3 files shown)


pirapira:
Looks good to me. Merging.

PR #9111, opened 2026-06-22
feat(rv64/rlp): byte-copy loop iteration (Phase 5, step 17)

dhsorens:
## Summary

One iteration of the byte-array copy loop — the straight-line 5-instruction core that decodes 20-byte address / 32-byte hash RLP fields into the output struct.

`EvmAsm/Rv64/RLP/ByteCopyIter.lean`:
- **`rlp_copy_iter_spec_within`** — `LBU x12,x13,0` reads `srcBytes[off+k]` from the source `bytesRegion`; `SB x12,x14,0` writes it to dst byte `k` of the destination `bytesRegion` (via `bytesRegion_sb_within`); then `ADDI` ×3 advance the src pointer (→ `off+k+1`), dst pointer (→ `k+1`), and counter (→ −1).

Source and destination are separate `**` conjuncts, so each instruction frames the region it doesn't touch (LBU frames dst, SB frames src) — no explicit disjointness hypothesis needed. Composed via `cpsTripleWithin_seq`.

## Stacking

Stacked on `feat/rv64-bytes-region-store` (PR #9106 — `bytesRegion_sb_within`), which in turn stacks on #9103. Retarget down the chain as each merges.

## Verification
- `#print axioms rlp_copy_iter_spec_within` → `[propext, Classical.choice, Quot.sound]` only; 0 sorry.


pirapira:
Looks good to me. Merging.

PR #9115, opened 2026-06-22
feat(rv64/rlp): unrolled byte-copy chain infrastructure (Phase 5, step 18)

dhsorens:
## Summary

Infrastructure for the **unrolled byte-array copy chain** — copying N bytes as N `rlp_copy_iter` blocks in sequence (block `j` at `base + 20*j`, destination region evolving one byte per block). Field sizes are fixed (20-byte address, 32-byte hash), so unrolling avoids the hardware-loop branch/back-edge machinery; a tight loop is a future code-size optimization.

`EvmAsm/Rv64/RLP/ByteCopyChainInfra.lean` (mirrors `ScalarFieldWalkInfra`):
- **`copyIterCR base`** — one iteration's 5-slot CodeReq (defeq to `rlp_copy_iter_spec_within`'s).
- **`copyIterCR_none` / `copyIterCR_disjoint`** — range-based disjointness (`base2 ≥ base1 + 20`), proved once.
- **`byteCopyChainCR base N`** — recursive CodeReq of the unrolled N-block chain.
- **`copyIterCR_disjoint_chainCR`** — one block disjoint from the rest-of-chain (induction on block count).
- **`copyRange dst src off k0 N`** — the destination-bytes result of copying N bytes.

This unblocks the chain spec (induction over N, reusing `rlp_copy_iter_spec_within`), then the address/hash field decode-and-copy units.

## Verification
- `#print axioms` → `[propext, Classical.choice, Quot.sound]` only; 0 sorry.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -81,6 +81,7 @@ import EvmAsm.Rv64.RLP.ScalarFieldWalkChain
 import EvmAsm.Rv64.RLP.ScalarFieldWalkInfra
 import EvmAsm.Rv64.RLP.UnifiedNScalarFieldWalk
 import EvmAsm.Rv64.RLP.ByteCopyIter
+import EvmAsm.Rv64.RLP.ByteCopyChainInfra
 import EvmAsm.Rv64.RLP.NestedDescendOne
 import EvmAsm.Rv64.RLP.UnifiedDecodeItem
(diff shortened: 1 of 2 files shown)


pirapira:
Looks good to me. Merging.

We might want to use a pointer instead of copying things (if the size of the RAM continues to be bottleneck). But copying out addresses and hashes will probably happen from time to time.

PR #9120, opened 2026-06-22
feat(rv64/rlp): general copy iteration with independent src/dst indices (Phase 5, step 20)

dhsorens:
## Summary

`rlp_copy_iter_gen_spec_within` generalizes `rlp_copy_iter_spec_within` (#9111) to **independent** source and destination byte indices `si`/`di`. The original couples them (`src_idx = off + dst_idx`, `off : Nat ≥ 0`), which is insufficient for a field copy: the payload position in the input buffer is unrelated to the field's offset in the output struct and can be *smaller* than it (e.g. legacy-tx `value` at struct offset 76). With independent indices, the destination can be the **whole dword-aligned output-struct `bytesRegion`**, with the field landing at an arbitrary (possibly unaligned) byte offset — only `outBase` needs alignment.

Same 5-instruction proof (`LBU src[si]` / `SB dst[di]` / three pointer ADDIs), with `si`/`di` in place of `off+k`/`k`. Supersedes the coupled `rlp_copy_iter_spec_within` for the chain/field work to follow.

## Verification
- `#print axioms` → `[propext, Classical.choice, Quot.sound]` only; 0 sorry.


pirapira:
Looks good to me. Merging.

PR #9123, opened 2026-06-22
feat(rv64/rlp): leaf byte-array field copy (Phase 5, step 22)

dhsorens:
## Summary

`unified_field_bytes_copy` — the leaf byte-array field copy (byte-array analog of `unified_field_scalar_read`). From `x13 = regionBase + ofNat off` (a decoded `.bytes` payload pointer — the single-item decoder's output convention), copy `N` bytes into the output struct region at byte offset `di0`:
- `ADDI x14, rOut, fieldImm` sets the destination pointer to `outBase + di0` (with `signExtend12 fieldImm = ofNat di0` passed as a hypothesis, which concrete callers discharge by `decide`);
- the general copy chain (`rlp_copy_chain_gen_spec`) writes `output[di0 .. di0+N) := src[off .. off+N)`.

The destination is the whole dword-aligned output-struct `bytesRegion`, so the field offset `di0` is arbitrary (possibly unaligned).

Also adds **`singleton_disjoint_byteCopyChainCR`** (a singleton before the chain is disjoint from it, by induction) for the ADDI ⊥ chain CodeReq.

## Verification
- `#print axioms` → classical only; 0 sorry.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -85,6 +85,7 @@ import EvmAsm.Rv64.RLP.ByteCopyIterGen
 import EvmAsm.Rv64.RLP.ByteCopyChainInfra
 import EvmAsm.Rv64.RLP.ByteCopyChain
 import EvmAsm.Rv64.RLP.ByteCopyChainGen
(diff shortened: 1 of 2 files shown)


pirapira:
Looks great. Merging.

PR #9131, opened 2026-06-22
feat(rv64/rlp): full byte-array field decode-and-copy (Phase 5, step 23)

dhsorens:
## Summary

`unified_bytes_field_decode_and_copy` — the **full byte-array field unit**, the byte-array analog of `unified_scalar_field_decode_and_store`. Decode a `.bytes data` field (`data.length ≤ 55` — covers 20-byte addresses, 32-byte hashes) at `x13 = regionBase + ofNat O` and copy its payload into the output struct region at byte offset `di0`; the output region's `[di0, di0+data.length)` becomes `data`. Coincides with `decode (bs.drop O) = some (.bytes data, tail)`.

Composes **`unified_list_header_descend`** (header decode → x13 = payload pointer, x11 = length) with **`unified_field_bytes_copy`** (the leaf copy, #9123):
- connects the abstract `itemPtrRegion`/`itemLenRegion` to concrete `(payloadOff, data.length)` via `bytes_item_payload_window` (mirroring `unified_scalar_field_decode`);
- connects the copy result `copyRangeGen … bs payloadOff` to `data` via a new `copyRangeGen_congr` (the payload window `bs[payloadOff..] = data`);
- frames the output region through the header descend and the header-leftover regs through the copy, weakening `x5/x10/x11` to `regOwn`;
- new `byteCopyChainCR_none` (chain is `none` outside its slots) for the decoder-`ofProg` ⊥ copy-chain disjointness.

This is the **first end-to-end byte-array (address/hash) field decoder** — the bridge from the copy primitives to actual decoding, alongside the scalar field decode-and-store.

## Verification
- `#print axioms unified_bytes_field_decode_and_copy` → `[propext, Classical.choice, Quot.sound]`; 0 sorry.


pirapira:
Looks good to me. Merging

PR #9135, opened 2026-06-22
feat(rv64/rlp): scalar register-spill iteration (Phase 5, step 24)

dhsorens:
## Summary

`rlp_spill_iter_spec_within` — one iteration of spilling a u64 value register into the output struct region (the scalar analog of one copy-loop block):
- `SB x14, x11, 0` writes `x11`'s low byte to output byte `di`;
- `SRLI x11, x11, 8` shifts the value right by 8;
- `ADDI x14, x14, 1` advances the dst pointer.

Spilling all 8 bytes writes the value **little-endian** into the unified output-struct `bytesRegion` — the same whole-struct region the byte-array (address/hash) fields target. This lets decoded u64 scalar fields share one output region with byte-array fields (necessary because per-field resources overlap due to dword-rounding), the first piece of the unified output model.

## Verification
- `#print axioms rlp_spill_iter_spec_within` → `[propext, Classical.choice, Quot.sound]`; 0 sorry.


pirapira:
Looks good to me. Merging

PR #9137, opened 2026-06-22
feat(rv64/rlp): scalar register-spill chain (Phase 5, step 25)

dhsorens:
## Summary

`rlp_spill_chain_spec` — spill `N` bytes of `x11 = v` (little-endian) into output positions `[di0, di0+N)`, as `N` `rlp_spill_iter` blocks in sequence (block `j` at `base + 12*j`, `x11` shifting right by 8 each block). For a u64 scalar field `N = 8`, writing the value's 8 LE bytes into the unified output-struct `bytesRegion` the address/hash fields also target. Result region = `spillRange outBytes v di0 N`.

Mirrors the byte-copy chain (`ByteCopyChainInfra`/`Gen`): a 3-slot `spillIterCR` + range-based disjointness (`spillIterCR_disjoint` / `spillIterCR_disjoint_chainCR`) + recursive `spillChainCR` + induction on N.

## Stacking

Stacked on `feat/rv64-rlp-scalar-spill-iter` (PR #9135 — the spill iteration). Retarget to `main` once #9135 merges.

## Verification
- `lake build` + full `EvmAsm` — clean, 0 warnings under `EvmAsm/`; umbrella import present.
- `#print axioms rlp_spill_chain_spec` → `[propext, Classical.choice, Quot.sound]`; 0 sorry.
- `scripts/check-forbidden-tactics.sh` OK; no `maxHeartbeats` override.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9139, opened 2026-06-22
feat(rv64/rlp): leaf scalar store into output region (Phase 5, step 26)

dhsorens:
## Summary

`unified_field_scalar_store_region` — given a decoded u64 value in `x11`, set up the destination pointer (`ADDI x14, rOut, fieldImm` with `signExtend12 fieldImm = ofNat di0`) and spill the value's `N` little-endian bytes into the unified output-struct `bytesRegion` at byte offset `di0`; `output[di0 .. di0+N)` becomes the value's LE bytes (`spillRange outBytes v di0 N`).

The spill analog of `unified_field_bytes_copy` — lets decoded scalar fields write into the **same whole-struct region** the address/hash fields use. Also adds `singleton_disjoint_spillChainCR`.

## Stacking

Stacked on `feat/rv64-rlp-scalar-spill-chain` (PR #9137 — the spill chain), itself on #9135. Retarget down the chain as each merges.

## Verification
- `#print axioms unified_field_scalar_store_region` → `[propext, Classical.choice, Quot.sound]`; 0 sorry.


pirapira:
Looks good to me. Merging

PR #9145, opened 2026-06-22
feat(rv64/rlp): heterogeneous scalar+byte-array field walk (Phase 5)

dhsorens:
## Summary

The first **heterogeneous** multi-field RLP decode: a scalar field followed by a byte-array field, both decoded into **one shared whole-struct output `bytesRegion`**. This is the keystone tying the two field-type decoders together — real STF schemas (legacy tx, block header) interleave u64 scalars (`nonce`, `gas`) with fixed byte arrays (20-byte address, 32-byte hash).

`unified_hetero_field_walk`:
- **Field A (scalar)** via `unified_scalar_field_decode_and_store_region` (concrete scratch pre) — spills its u64 value little-endian into the output region at byte offset `diA`, advances `x13` to the next field.
- **Field B (byte array)** via `unified_bytes_field_decode_and_copy_at_regOwn` (the `regOwn`-pre variant, callable after A clobbered the scratch) — copies its payload into the same region at byte offset `diB`.
- The output region is **threaded directly** (A's `spillRange` is B's input `bytesRegion`), so the final region holds both fields: `copyRangeGen (spillRange outBytes valueA diA 8) dataB 0 diB |dataB|`. A's advanced `x13` feeds B's payload pointer with no glue code.
- Coincides with the scalar decode of A (`decodeScalar`) and the item decode of B (`decode`).

## Disjointness

A's code `[base, base+280)` ⊥ B's code `≥ base+280` is proved by a single **global range split** (`a < base+280` ⇒ unit B is `none`; else unit A is `none`), reusing the existing `spillChainCR_none` / `byteCopyChainCR_none` chain lemmas — avoiding the leaf×leaf product that both units' trailing chains would otherwise force.

Adds `spillRange_length` (spilling preserves list length, to thread B's length hypotheses through A's spilled region).

## Verification
- `lake build EvmAsm.Rv64.RLP.UnifiedHeteroFieldWalk` + umbrella `EvmAsm.Rv64.RLP` — clean.
- `#print axioms unified_hetero_field_walk` → `[propext, Classical.choice, Quot.sound]` (classical only); 0 sorry.


pirapira:
Looks good to me merging

PR #9149, opened 2026-06-22
feat(rv64/rlp): reusable code-range disjointness for field units (Phase 5)

dhsorens:
## Summary

Every multi-field RLP walk must prove unit-A's code ⊥ unit-B's code. Each field-decode unit occupies a **contiguous** code range (`[base, base+280)` for the scalar-into-region unit; `[base, base + (152 + 20·|data|))` for the byte-array unit), so two units at non-overlapping ranges have disjoint CodeReqs. This packages that fact once so future walks discharge disjointness in two lines instead of a per-leaf product (which would otherwise blow up — both unit CRs end in chains).

- **`codeReq_disjoint_of_ranges`** — generic: if `cr1` is `none` at/above a boundary `M` and `cr2` is `none` below `M`, then `cr1 ⊥ cr2`.
- **`scalarRegionUnitCR`** + **`scalar_region_unit_cr_none_above` / `_below`** — the scalar-into-region unit's CR is `none` outside `[base, base+280)`.
- **`bytesUnitCR`** + **`bytes_unit_cr_none_above` / `_below`** — the byte-array unit's CR is `none` outside `[base, base + (152 + 20·len))`.

The `none` lemmas reuse the existing `spillChainCR_none` / `byteCopyChainCR_none` chain lemmas. The named CRs are definitionally the inline unions the unit theorems produce, so they apply directly.

This is the enabling step for cheap **3-field / N-field walks** and the concrete **9-field tx / ~19-field header** decoders. Independent of #9145/#9147 (only depends on merged CR definitions).

## Verification
- `#print axioms` (all five) → classical only (`propext, Quot.sound`); 0 sorry.


pirapira:
Looks good to me merging

PR #9147, opened 2026-06-22
feat(rv64/rlp): regOwn-pre scalar-into-region field variant (Phase 5)

dhsorens:
## Summary

`unified_scalar_field_decode_and_store_region_at_regOwn` — the `regOwn`-precondition variant of `unified_scalar_field_decode_and_store_region`.

The decode clobbers its scratch registers (`x5, x10, x11, x12`) and releases them as `regOwn` in its post, so to chain a scalar field **after** another field (which left those `regOwn`), the scalar-into-region unit must accept `regOwn` scratch in its precondition. This peels those four via `cpsTripleWithin_of_forall_regIs_to_regOwn`; `x14`/`x15` stay concrete (`x14` is overwritten by the unit's own `ADDI x14, rOut, fieldImm`; `x15` is framed through untouched).

This is the exact scalar mirror of `unified_bytes_field_decode_and_copy_at_regOwn` (#9142) — the prerequisite for a scalar field appearing in any non-first schema position (e.g. `scalar ⨾ scalar ⨾ …` chains), unblocking the N-field heterogeneous walk.

Independent of #9145 (only depends on the merged `unified_scalar_field_decode_and_store_region`).

## Verification
- `lake build` of the file + umbrella `EvmAsm.Rv64.RLP` — clean.
- `#print axioms` → `[propext, Classical.choice, Quot.sound]` (classical only); 0 sorry.
- `check-forbidden-tactics.sh` OK; `check-no-warnings.sh` → 0 warnings under `EvmAsm/`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9150, opened 2026-06-22
feat(rv64/rlp): three-field heterogeneous walk (scalar ⨾ scalar ⨾ bytes)

dhsorens:
## Summary

`unified_three_field_walk` — a **three-field** heterogeneous RLP decode: scalar A ⨾ scalar B ⨾ byte-array C, all decoded into **one shared output `bytesRegion`** threaded directly through all three (A spills its u64 LE value at `diA`, B spills at `diB`, C copies its payload at `diC`; the final region is `copyRangeGen (spillRange (spillRange outBytes valueA diA 8) valueB diB 8) dataC 0 diC |dataC|`).

This is the **integration test** for the two new composition pieces and shows the path scales to real schemas:
- the scalar-into-region **`regOwn` re-entry variant** (the middle field B, callable after A clobbered the scratch);
- the reusable **code-range disjointness** — disjointness here is just two `codeReq_disjoint_of_ranges` calls (A ⊥ B and (A∪B) ⊥ C), no per-leaf product. The named unit CRs (`scalarRegionUnitCR`, `bytesUnitCR`) keep the whole statement compact.

It is the direct precursor to the concrete legacy-tx (9-field) / block-header (~19-field) decoders, which are fixed sequences of these units.

## Stacking

Stacks on the heterogeneous-2-field (#9145), scalar-region-`regOwn` (#9147), and field-unit-disjoint (#9149) branches (base set to #9145's branch). The diff narrows to just `UnifiedThreeFieldWalk.lean` once the dependencies merge.

## Verification
- `#print axioms unified_three_field_walk` → `[propext, Classical.choice, Quot.sound]` (classical only); 0 sorry.


pirapira:
Looks good to me. Merging

PR #9142, opened 2026-06-22
feat(rv64/rlp): regOwn-pre byte-array field variant (Phase 5, step 28)

dhsorens:
## Summary

`unified_bytes_field_decode_and_copy_at_regOwn` — the byte-array field unit (#9131) restated with its four clobbered scratch registers (`x5, x10, x11, x12`) owned abstractly (`regOwn`) in the precondition, via nested `cpsTripleWithin_of_forall_regIs_to_regOwn` (à la the scalar `regOwn` variant `unified_scalar_field_decode_and_store_at_regOwn`).

Callable **after a prior field has run** (which left those scratch registers `regOwn`) — the prerequisite for chaining byte-array fields in a heterogeneous schema walk. `x14`/`x15` stay concrete (the prior field supplies them).

## Verification
- `lake build` + full `EvmAsm` — clean, 0 warnings under `EvmAsm/`; umbrella import present.
- `#print axioms` → `[propext, Classical.choice, Quot.sound]`; 0 sorry.
- `scripts/check-forbidden-tactics.sh` OK; no `maxHeartbeats` override.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9160, opened 2026-06-22
feat(rv64/rlp): canonical all-regOwn-scratch field units (Phase 5)

dhsorens:
## Summary

To chain RLP fields in **any** order — real transaction schemas interleave u64 scalars (`nonce`, `value`, `v/r/s`) with byte-array fields (`to`, `data`) — the field units need a uniform scratch-register interface. The byte-array unit releases `x15` as `regOwn` (it uses `x15` as the copy counter), whereas the scalar unit framed `x15` through concretely. That mismatch blocks a **byte-array → scalar** field boundary (the bytes unit's `regOwn x15` output can't satisfy a scalar unit's concrete `x15` precondition).

This adds two **canonical re-entry variants** with a uniform all-`regOwn` scratch interface (`x5, x10, x11, x12, x15`); `x14` stays a concrete `∀`-parameter (each unit overwrites it via `ADDI x14, rOut, fieldImm`, so the prior field's concrete `x14` feeds in directly):

- **`unified_scalar_field_decode_and_store_region_canonical`** — built from `…_at_regOwn` (#9147) by weakening its post `x15 ↦ v15 → regOwn x15` (sound: the scalar decode never reads `x15`, it's framed through) and then peeling `x15` in the pre.
- **`unified_bytes_field_decode_and_copy_canonical`** — built from `…_at_regOwn` (#9142) by peeling `x15` in the pre (its post already releases `regOwn x15`).

Both state their CodeReq via the named `scalarRegionUnitCR` / `bytesUnitCR` (#9149), so the upcoming **N-field heterogeneous fold** sees one uniform unit shape regardless of field kind or order.

## Verification
- `#print axioms` (both) → classical only (`propext, Classical.choice, Quot.sound`); 0 sorry.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -96,6 +96,7 @@ import EvmAsm.Rv64.RLP.FieldUnitDisjoint
 import EvmAsm.Rv64.RLP.UnifiedScalarFieldRegionRegOwn
 import EvmAsm.Rv64.RLP.UnifiedHeteroFieldWalk
 import EvmAsm.Rv64.RLP.UnifiedThreeFieldWalk
+import EvmAsm.Rv64.RLP.UnifiedScalarFieldRegionCanonical
 import EvmAsm.Rv64.RLP.NestedDescendOne
 import EvmAsm.Rv64.RLP.UnifiedDecodeItem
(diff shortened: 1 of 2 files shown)


pirapira:
Looks good to me. Merging.

PR #9165, opened 2026-06-22
feat(rv64/rlp): fully-canonical field units (all scratch regOwn) (Phase 5)

dhsorens:
## Summary

The `…_canonical` units (#9160) gave a uniform `regOwn` interface for `x5, x10, x11, x12, x15` but kept `x14` a concrete `∀`-parameter. This peels `x14` as well — each unit overwrites `x14` via its own `ADDI x14, rOut, fieldImm`, so it never reads the incoming value; peeling it in the pre and weakening the advanced value to `regOwn` in the post is sound.

Result: both unit kinds now have a **fully uniform scratch interface** (`x5, x10, x11, x12, x14, x15` all `regOwn` in pre **and** post), so a unit's pre and post differ only in `x13` (the input pointer, advanced by the field's encoding length) and the output `bytesRegion` (updated with the decoded field).

- `unified_scalar_field_decode_and_store_region_fully_canonical`
- `unified_bytes_field_decode_and_copy_fully_canonical`

This uniformity is the key enabler for the upcoming **N-field heterogeneous fold**: every unit has the identical scratch interface, so an arbitrary field list chains with zero per-field scratch bookkeeping (only `x13` and the output region thread through).

## Verification
- `#print axioms` (both) → classical only (`propext, Classical.choice, Quot.sound`); 0 sorry.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -97,6 +97,7 @@ import EvmAsm.Rv64.RLP.UnifiedScalarFieldRegionRegOwn
 import EvmAsm.Rv64.RLP.UnifiedHeteroFieldWalk
 import EvmAsm.Rv64.RLP.UnifiedThreeFieldWalk
 import EvmAsm.Rv64.RLP.UnifiedScalarFieldRegionCanonical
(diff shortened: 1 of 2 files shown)


pirapira:
Looks good to me. Merging.

PR #9172, opened 2026-06-22
feat(rv64/rlp): N-field heterogeneous schema fold (Phase 5)

dhsorens:
## Summary

`schema_walk` — the **N-field heterogeneous fold**: decode an arbitrary schema (`List FieldSpec`, scalar | byte-array in **any** order) into one shared output `bytesRegion`. This is the generic engine that turns the concrete legacy-tx (9-field) and block-header (~19-field) decoders into cheap instantiations — a schema is just a list.

Built on the fully-canonical units (#9165): every unit has a uniform all-`regOwn` scratch interface, so the fold threads only `x13` (the input pointer) and the output region.

Contents:
- **Schema definitions** — `fieldSize/Steps/Enc/CR/Update`, `schemaSize/Steps/Enc/Out/CR`, `schemaINV` (the uniform fold invariant), `SchemaValid` (per-field validity), `schemaDecodes` (per-field RLP-spec coincidences).
- **Length preservation** — `copyRangeGen_length`, `fieldUpdate_length`, `schemaOut_length`.
- **Disjointness by code range** — `schemaCR_none_above` / `_none_below` / `schemaCR_disjoint_fieldCR`, all via `codeReq_disjoint_of_ranges` (#9149). Each unit occupies a contiguous `[base_i, base_i + fieldSize)`, so the schema CR is `none` outside `[base, base + schemaSize)`.
- **`field_step`** — a kind-generic single-field lemma (the scalar/byte-array case split happens once).
- **`schema_walk`** — the list-induction fold, returning the operational triple **and** the field-by-field decode coincidences (`schemaDecodes`).

## Verification
- `#print axioms schema_walk` / `field_step` → classical only (`propext, Classical.choice, Quot.sound`); 0 sorry.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -98,6 +98,7 @@ import EvmAsm.Rv64.RLP.UnifiedHeteroFieldWalk
 import EvmAsm.Rv64.RLP.UnifiedThreeFieldWalk
 import EvmAsm.Rv64.RLP.UnifiedScalarFieldRegionCanonical
 import EvmAsm.Rv64.RLP.UnifiedFieldUnitFullyCanonical
+import EvmAsm.Rv64.RLP.SchemaFold
 import EvmAsm.Rv64.RLP.NestedDescendOne
 import EvmAsm.Rv64.RLP.UnifiedDecodeItem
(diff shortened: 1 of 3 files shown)


pirapira:
Looks good to me. Merging.

PR #9179, opened 2026-06-22
feat(rv64/rlp): schema-fold concat helper + concrete cross-check (Phase 5)

dhsorens:
## Summary

`schema_walk` (#9172) requires a per-field drop hypothesis (`SchemaValid`). For a real decoder the input buffer is exactly the concatenation of the fields' RLP encodings followed by a tail; this PR derives the whole `SchemaValid` from that single fact, making the fold a one-liner to apply.

- **`schemaEncBytes`** — the byte string a schema decodes (concatenation of the fields' encodings).
- **`fieldCoreValid`** — per-field validity without drop bookkeeping (length / immediate / output-bound).
- **`schemaValid_of_concat`** — `SchemaValid` from `bs.drop O = schemaEncBytes specs ++ tail` plus per-field core validity (induction over the schema).

Plus a **concrete end-to-end example**: decode `[scalar 0x2a, bytes [0x01,0x02], scalar 0x07]` from the 5-byte buffer `[0x2a, 0x82, 0x01, 0x02, 0x07]` into a 24-byte output region at offsets 0/8/16, with `SchemaValid` discharged via `schemaValid_of_concat` — demonstrating the generic fold applies cleanly to a concrete heterogeneous schema.

Stacks on the N-field fold (#9172).

## Verification
- `#print axioms schemaValid_of_concat` → classical only (`propext, Quot.sound`); 0 sorry.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -99,6 +99,7 @@ import EvmAsm.Rv64.RLP.UnifiedThreeFieldWalk
 import EvmAsm.Rv64.RLP.UnifiedScalarFieldRegionCanonical
 import EvmAsm.Rv64.RLP.UnifiedFieldUnitFullyCanonical
 import EvmAsm.Rv64.RLP.SchemaFold
+import EvmAsm.Rv64.RLP.SchemaFoldConcat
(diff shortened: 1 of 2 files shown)


pirapira:
Looks good to me. Merging.

PR #9186, opened 2026-06-22
feat(rv64/rlp): RLP-list schema decoder (descend ⨾ N-field fold) (Phase 5)

dhsorens:
## Summary

`list_schema_walk` — the shape every real STF structure takes: a transaction / block header is an RLP **list** whose elements are the fields. This descends one list level to its payload (`unified_list_header_descend`), then runs `schema_walk` (#9172) over the element fields into one shared output `bytesRegion`.

The descend primitive's post has **exactly the same atom order** as `schema_walk`'s `schemaINV` precondition, so the bridge is clean:
- a **positional scratch weaken** (concrete → `regOwn` for `x5,x10,x11,x12,x14,x15`, via `sepConj_mono`);
- an **`x13`-pointer rewrite** (`hptr` supplies the payload offset `O'` from the list prefix — abstracted so the caller handles short vs long list headers);
- the output region is **framed** through the descend;
- descend code `[base, base+148)` ⊥ schema code `[base+148, …)` via `codeReq_disjoint_of_ranges` (#9149).

Coincides field-by-field with the RLP spec (`schemaDecodes`). Independent of #9179 (only needs the merged fold #9172).

## Verification
- `#print axioms list_schema_walk` → classical only (`propext, Classical.choice, Quot.sound`); 0 sorry.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -99,6 +99,7 @@ import EvmAsm.Rv64.RLP.UnifiedThreeFieldWalk
 import EvmAsm.Rv64.RLP.UnifiedScalarFieldRegionCanonical
 import EvmAsm.Rv64.RLP.UnifiedFieldUnitFullyCanonical
 import EvmAsm.Rv64.RLP.SchemaFold
+import EvmAsm.Rv64.RLP.SchemaListWalk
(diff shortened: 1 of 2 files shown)


pirapira:
Looks good to me. Merging.

PR #9193, opened 2026-06-22
feat(rv64/rlp): short-list schema decode convenience (Phase 5)

dhsorens:
## Summary

`short_list_schema_walk` specializes `list_schema_walk` (#9186) to a **short list** (RLP prefix `0xc0..0xf7`, payload ≤ 55 bytes). For a short list:
- the payload starts at `O + 1`, and
- `regionLongWindow` is vacuously `True` (only long forms carry length bytes).

So both the window precondition and the payload-offset (`hptr`) are discharged automatically from the single prefix-class fact `classifyPrefix (bs[O]) = .shortList` — the convenience entry point for short list-structured schemas (the caller no longer threads window/pointer reasoning).

Stacks on the list decoder (#9186).

## Verification
- `#print axioms short_list_schema_walk` → classical only (`propext, Classical.choice, Quot.sound`); 0 sorry.


pirapira:
Looks good to me. Merging.

PR #9260, opened 2026-06-23
feat(rv64/rlp): end-user decode API from RLP-encoded record (Phase 5)

dhsorens:
## Summary

`decode_encoded_short_list_schema` — the clean end-user entry point for the decoder. Given that the input buffer (from offset `O`) is the **genuine RLP encoding of the field record** — `encode (.list (schemaItems specs)) ++ tail`, with a short-list-sized payload — it runs the whole decoder and yields the field-by-field result.

Both of the previously-manual hypotheses are now **derived from the encoding**:
- the **prefix-class fact** — the header byte is `0xC0 + payloadLen`, which lands in the short-list range (`classifyPrefix_shortList_iff`);
- **`SchemaValid`** — via the encode bridge (`schemaConcat_of_encode_list_short`) + `schemaValid_of_concat`.

So the caller supplies only the encoding fact, per-field core validity, and the region/output well-formedness — **no RLP-internal proof obligations**. This is the culmination of the schema-decode stack: *"here is the RLP encoding of my field record"* → *full verified decode into the output struct, coinciding field-by-field with the spec.*

Composes the encode bridge (#9257), `schemaValid_of_concat`, and `short_list_schema_walk`. Stacks on #9257.

## Verification
- `#print axioms decode_encoded_short_list_schema` → classical only (`propext, Classical.choice, Quot.sound`); 0 sorry.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -105,6 +105,7 @@ import EvmAsm.Rv64.RLP.SchemaListWalkShort
 import EvmAsm.Rv64.RLP.SchemaListWalkLong
 import EvmAsm.Rv64.RLP.SchemaListDecodeExample
 import EvmAsm.Rv64.RLP.SchemaListEncode
+import EvmAsm.Rv64.RLP.SchemaDecodeEncoded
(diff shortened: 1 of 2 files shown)


pirapira:
Looks good. Merging

PR #9261, opened 2026-06-23
test(rv64/rlp): concrete cross-check of the end-user decode API (Phase 5)

dhsorens:
## Summary

Validates `decode_encoded_short_list_schema` (#9260) end-to-end, starting from a **genuine RLP encoding**:

`encode (.list [.bytes [0x2a], .bytes [0x01, 0x02]]) = [0xc4, 0x2a, 0x82, 0x01, 0x02]`

The whole decoder runs into a 24-byte output struct (scalar `0x2a` → 8 LE bytes at offset 0; the 2-byte array → offset 8). The caller supplies **only** the encoding equation (`by decide`), per-field core validity, and region/output well-formedness — the prefix-class fact and `SchemaValid` are derived from the encoding inside the API.

Demonstrates the complete **"RLP bytes in → verified decode out"** path with zero RLP-internal proof obligations.

Stacks on the decode API (#9260).

## Verification


pirapira:
Looks good. Merging

PR #9265, opened 2026-06-23
feat(rv64/rlp): long-list encode bridge (real tx/header form) (Phase 5)

dhsorens:
## Summary

Long-list counterpart of the encode bridge (#9257). Real STF records (transactions, block headers) exceed 55 bytes, so their RLP encoding is a **long list**: a `0xF7 + lenOfLen` prefix, then `lenOfLen` big-endian length bytes, then the payload.

- **`encode_list_schemaItems_long`** — when `(schemaEncBytes specs).length > 55`, the encoding of the field items is `[0xF7 + lenOfLen] ++ lenBytes ++ schemaEncBytes` (the payload is again `schemaEncBytes`, via `encodeItems_schemaItems`).
- **`schemaConcat_of_encode_list_long`** — dropping the `1 + lenOfLen` header bytes from a long-list-encoded input yields `schemaEncBytes specs ++ tail` — exactly the concat premise `long_list_schema_walk` consumes, at payload offset `O + 1 + lenOfLen`.

Pure spec-side; sets up the long-list end-user decode API (capstone) next. Stacks on #9257.

## Verification
- `#print axioms` (both) → classical only (`propext, Quot.sound`); 0 sorry.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -105,6 +105,7 @@ import EvmAsm.Rv64.RLP.SchemaListWalkShort
 import EvmAsm.Rv64.RLP.SchemaListWalkLong
 import EvmAsm.Rv64.RLP.SchemaListDecodeExample
 import EvmAsm.Rv64.RLP.SchemaListEncode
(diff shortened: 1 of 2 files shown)


pirapira:
Looks great. Merging.

PR #9267, opened 2026-06-23
feat(rv64/rlp): long-list end-user decode API (real tx/header form) (Phase 5)

dhsorens:
## Summary

`decode_encoded_long_list_schema` — the long-list counterpart of `decode_encoded_short_list_schema` (#9260). When the input buffer (from `O`) is the genuine RLP **long-list** encoding of the field record (`encode (.list (schemaItems specs)) ++ tail`, payload `> 55` and `< 256^8` bytes — the real transaction/header case), the decoder runs and yields the field-by-field result.

All the RLP-internal facts are **derived from the encoding**:
- the **`longList` prefix fact** — the header byte is `0xF7 + lenOfLen`, and `1 ≤ lenOfLen ≤ 8` (from `Nat.toBytesBE_length_le` + a positivity reproof via `Nat.toBytesBE_succ`), so it lands in `[0xF8, 0xFF]` (`classifyPrefix_longList_iff`);
- the **length-bytes-fit bound** (`hlenbytes`) — from the encoding's length;
- **`SchemaValid`** — via the long encode bridge (`schemaConcat_of_encode_list_long`) + `schemaValid_of_concat`.

The caller supplies only the encoding fact, `O < bs.length`, per-field core validity, and region/output well-formedness. **Together with the short-list API (#9260), this completes the "RLP bytes in → verified decode out" entry point for both list forms** — i.e. every transaction and block header shape.

Composes the long encode bridge (#9265), `schemaValid_of_concat`, and `long_list_schema_walk`. Stacks on #9265.

## Verification
- `#print axioms decode_encoded_long_list_schema` → classical only (`propext, Classical.choice, Quot.sound`); 0 sorry.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -106,6 +106,7 @@ import EvmAsm.Rv64.RLP.SchemaListWalkLong
 import EvmAsm.Rv64.RLP.SchemaListDecodeExample
 import EvmAsm.Rv64.RLP.SchemaListEncode
 import EvmAsm.Rv64.RLP.SchemaListEncodeLong
+import EvmAsm.Rv64.RLP.SchemaDecodeEncodedLong
 import EvmAsm.Rv64.RLP.NestedDescendOne
 import EvmAsm.Rv64.RLP.UnifiedDecodeItem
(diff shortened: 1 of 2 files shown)


pirapira:
Looks good to me. Merging.

PR #9272, opened 2026-06-23
feat(rv64/rlp): list-level decode coincidence for schema decoders (Phase 5)

dhsorens:
## Summary

`decode_list_of_encoded` — the **whole-list** spec coincidence, complementing the N-field fold's per-field `schemaDecodes`. When the input buffer from `O` is the RLP encoding of the field record (`encode (.list (schemaItems specs)) ++ tail`), the pure RLP spec decodes it to `some (.list (schemaItems specs), tail)` — i.e. the buffer the operational decoder consumes *is*, per spec, the encoding of the list of field items.

It follows directly from the spec's `decode ∘ encode` round-trip (`decode_encode_append`) — no fuel / `decodeListPayload` reasoning. This ties the operational decoder's input to the pure spec at the **list level** (the per-field coincidence ties it at the element level).

Stacks on the encode bridge (#9257).

## Verification
- `#print axioms decode_list_of_encoded` → classical only (`propext, Classical.choice, Quot.sound`); 0 sorry.


pirapira:
Looks great. Merging.

PR #9278, opened 2026-06-23
feat(rv64/rlp): n=0 empty scalar field decode-and-store (Phase 5, step 41)

dhsorens:
## Summary

First field-type primitive toward a verified **legacy-transaction** decoder. The generic RLP schema engine is complete, but its scalar field unit requires `1 ≤ data.length`, so it cannot decode a **zero-valued scalar** — and in RLP a scalar `0` is the empty byte string `[0x80]` (minimal encoding ⇒ every zero `nonce`/`value`/empty `to` appears this way).

This adds `unified_scalar_field_decode_and_store_zero`. The observation that makes it cheap: for an empty `.bytes []` field the single-item header descent already advances `x13` to the payload pointer, which — the payload being empty — **is** the next field (`regionBase + ofNat(O+1)`). So there is nothing to read: descend, then `SD rOut, x0, offset` to write the value `0` (from the always-zero register `x0`). No big-endian read loop.

Coincides with the pure spec: `decodeScalar (bs.drop O) = some (0, tail)`.

## Verification
- `lake build EvmAsm.Rv64.RLP.UnifiedScalarFieldZero` + umbrella `EvmAsm.Rv64.RLP` — clean, 0 warnings.
- `#print axioms` → `[propext, Classical.choice, Quot.sound]` (classical only); 0 sorry/admit.
- Concrete `example`: decode `[0x80]` → cell `0x3000 ↦ₘ 0`, `decodeScalar [0x80] = some (0, [])`.


pirapira:
Looks great. Merging.

PR #9279, opened 2026-06-23
feat(rv64/rlp): u256 wide scalar field decode-and-copy (Phase 5, step 42)

dhsorens:
## Summary

Second field-type primitive toward a verified **legacy-transaction** decoder. *(Stacked on #9278.)*

The single-word scalar unit reads the payload big-endian into one 64-bit register, so it caps at `data.length ≤ 8`. But a legacy transaction's `nonce, gas_price, gas, value, v, r, s` are **u256** — `r`/`s` are essentially always the full 32 bytes — which that unit cannot decode.

`unified_wide_scalar_field_decode_and_copy` lifts the ceiling from 8 to 32 bytes **with no multi-limb arithmetic**. A u256 payload is ≤ 32 bytes, so it rides the byte-array copy unit (proven for ≤ 55 bytes): copy the raw big-endian payload into the output region (contiguous, advancing the output cursor `x14`), exactly as the schema fold lays fields out. The scalar value coincidence is then **free** — `decodeScalar` reads the decoded item's bytes as a big-endian natural with no minimality check — so:

```
decodeScalar (bs.drop O) = some (Nat.fromBytesBE data, tail)
```

follows directly from the copy unit's `decode` coincidence.

Output holds the field's minimal big-endian bytes; fixed-width 32-byte zero-padding (if a target struct wants it) is a presentation step layered on at schema-assembly time.

## Verification
- `lake build EvmAsm.Rv64.RLP.UnifiedWideScalarField` + umbrella — clean, 0 warnings.
- Concrete `example`: a 9-byte payload (the smallest width the old `≤ 8` unit could not decode) decodes and copies, with scalar value `Nat.fromBytesBE [0x01,…,0x09]`.


pirapira:
Looks good. Merging.

PR #9281, opened 2026-06-23
feat(rv64/rlp): scalar-value view of a schema decode (Phase 5, step 43)

dhsorens:
## Summary

Spec keystone toward the **legacy-transaction** decoder. *(Stacked on #9279.)*

The N-field fold proves a per-field decode coincidence (`schemaDecodes`): each field decodes as a scalar (`decodeScalar`) or a byte array (`decode .bytes`). For the concrete tx/header decoders we want the numeric **value** of every field uniformly — and a legacy transaction's `nonce, value, v, r, s` are u256 scalars that (per #9279) ride the **byte-array** machine path, so the fold reports them via the `.bytes` coincidence.

This adds the missing forward bridge:

```
decodeScalar_of_decode_bytes : decode bs = some (.bytes data, rest)
                             → decodeScalar bs = some (Nat.fromBytesBE data, rest)
```

(a byte-string decode determines its payload's big-endian value — `decodeScalar` applies no minimality check), and lifts it over a whole schema: `schemaDecodes ⇒ schemaScalarValues`, recovering every field's value regardless of which machine path decoded it. So the all-byte-array `schema_walk` acts as a u256-field decoder.

## Verification
- `lake build EvmAsm.Rv64.RLP.SchemaScalarValues` + umbrella — clean, 0 warnings.
- `#print axioms` → `[propext, Quot.sound]` / `[propext]` (classical only, no `Classical.choice`); 0 sorry/admit.


pirapira:
Looks good. Merging.

PR #9285, opened 2026-06-23
feat(rv64/rlp): long byte-array field decode-and-copy >55 (Phase 5, step 44)

dhsorens:
## Summary

Fourth field-type primitive toward a verified **legacy-transaction** decoder. *(Stacked on #9281.)*

The short byte-array unit caps at 55 bytes (RLP short-string form). Real structures need more: a transaction's `data` (calldata) is routinely > 55 bytes, and a block header's `logsBloom` is a fixed 256-byte field — both RLP **long-string** form (`0xB8..0xBF`: prefix `0xB7 + lenOfLen`, then `lenOfLen` big-endian length bytes, then the payload).

The single-item decoder already handles long byte strings (`itemPtrRegion`/`itemLenRegion` have `longBytes` branches that read the length bytes from the region), and the byte-copy leaf is length-generic — so the only new ingredient is the **long payload-window lemma** `bytes_item_payload_window_long`:

- payload pointer `= regionBase + ofNat(off + 1 + lenOfLen)`
- recovered length `= data.length`
- region from the payload offset `= data ++ tail`

Composition (descent ⨾ copy) then mirrors the short unit. Coincides with `decode (bs.drop O) = some (.bytes data, tail)`.

## Verification
- `lake build EvmAsm.Rv64.RLP.UnifiedLongBytesField` + umbrella — clean, 0 warnings.
- No concrete `example`: `Nat.toBytesBE` is well-founded recursive and doesn't reduce under `decide`, so a long-form encoding can't be discharged by computation (same reason the codebase's long-list cross-checks go through the abstract theorems).


pirapira:
Looks good. Merging.

PR #9288, opened 2026-06-23
feat(rv64/rlp): fully-canonical long byte-array field unit (Phase 5, step 45)

dhsorens:
## Summary

Prerequisite for integrating `>55`-byte fields into the schema engine. *(Stacked on #9285.)*

The N-field fold (`schema_walk`) consumes field units with a uniform all-`regOwn` scratch interface, built via a peeling chain (`_at_regOwn` → `_canonical` → `_fully_canonical`). #9285 shipped only the *plain* long byte-array unit (concrete scratch), so it can't yet feed the fold.

This adds the canonical chain for the long (`>55`-byte) unit. Because the long unit has the **identical scratch pre/post/CR/step shape** as the short unit (only the payload-window proof differs), each peeling layer is its short counterpart with the underlying unit swapped. The endpoint `unified_long_bytes_field_decode_and_copy_fully_canonical` has the uniform interface `schema_walk` needs.

Next: drop the `≤55` cap from `field_step` (dispatch short/long) so `schema_walk` decodes arbitrarily-long byte fields (calldata, the 256-byte `logsBloom`) — no `FieldSpec`/CR change needed since both units share `fieldSize`/`fieldCR`/`fieldUpdate`.

## Verification
- `lake build EvmAsm.Rv64.RLP.UnifiedLongBytesFieldCanonical` + umbrella — clean, 0 warnings.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -91,6 +91,7 @@ import EvmAsm.Rv64.RLP.UnifiedBytesFieldDecode
 import EvmAsm.Rv64.RLP.UnifiedBytesFieldRegOwn
 import EvmAsm.Rv64.RLP.UnifiedWideScalarField
 import EvmAsm.Rv64.RLP.UnifiedLongBytesField
+import EvmAsm.Rv64.RLP.UnifiedLongBytesFieldCanonical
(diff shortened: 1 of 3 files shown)


pirapira:
Looks good. Merging

PR #9289, opened 2026-06-23
feat(rv64/rlp): integrate long byte-arrays into the schema engine (Phase 5, step 46)

dhsorens:
## Summary

The N-field fold (`schema_walk`) now decodes **arbitrarily-long** byte-array fields. *(Stacked on #9288.)*

Previously the fold capped byte-array fields at 55 bytes, so it couldn't decode a transaction's calldata or a header's 256-byte `logsBloom`. #9288 produced the fully-canonical long unit, which has the **identical triple shape** (`fieldSize`/`fieldCR`/`fieldUpdate`/step count) as the short unit.

This drops the cap:
- `field_step` dispatches on `data.length` — short (`≤55`) → short unit, long (`>55`) → long unit — with a uniform conclusion.
- The `SchemaValid` / `fieldCoreValid` byte-array condition relaxes from `data.length ≤ 55 ∧ size` to just the size bound (the `1 ≤ len` floor stays in the outer conjunct).

Because both units share `fieldSize`/`fieldCR`/`fieldUpdate`, **no `FieldSpec` or `CodeReq` definition changes**, and every downstream decoder and example (`SchemaListWalk*`, `SchemaDecodeEncoded*`, all examples) rebuilds unchanged.

`schema_walk` / `long_list_schema_walk` now decode every field type a real tx/header contains except `n=0` empty fields.

## Verification
- Full umbrella `EvmAsm.Rv64.RLP` builds clean (all downstream decoders/examples), 0 warnings.
- `#print axioms schema_walk` / `field_step` → `[propext, Classical.choice, Quot.sound]`; 0 sorry/admit.

### DIFF
--- a/EvmAsm/Rv64/RLP/SchemaFold.lean
+++ b/EvmAsm/Rv64/RLP/SchemaFold.lean
@@ -14,6 +14,7 @@
 -/
 
 import EvmAsm.Rv64.RLP.UnifiedFieldUnitFullyCanonical
+import EvmAsm.Rv64.RLP.UnifiedLongBytesFieldCanonical
 import EvmAsm.Rv64.RLP.UnifiedHeteroFieldWalk
 
 namespace EvmAsm.Rv64.RLP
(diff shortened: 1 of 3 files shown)


pirapira:
Looks good. Merging

PR #9292, opened 2026-06-23
feat(rv64/rlp): end-user decode-to-field-values API (Phase 5, step 47)

dhsorens:
## Summary

The one-shot **decode-to-field-values** API behind the concrete tx/header decoders. *(Stacked on #9289.)*

`decode_encoded_{short,long}_list_schema` yield the operational decode triple plus per-field `schemaDecodes` (each field decodes as a scalar or a byte array). A real STF consumer wants the numeric **value** of every field uniformly — and a transaction's u256 fields ride the byte-array path, so the fold reports them via `decode`/`.bytes`.

This adds `decode_encoded_{short,long}_list_schema_values`: the encoded-list decoders wrapped with `schemaDecodes_imp_scalarValues`, so the conclusion is `schemaScalarValues` — every field's big-endian value at its input offset, uniformly. RLP bytes in → operational decode + all field values out, verified.

The named-struct (legacy-tx) decoder is now a direct instantiation of this API.

## Verification
- `lake build EvmAsm.Rv64.RLP.SchemaDecodeValues` + umbrella — clean, 0 warnings.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -113,6 +113,7 @@ import EvmAsm.Rv64.RLP.SchemaListEncode
 import EvmAsm.Rv64.RLP.SchemaLongListDecodeExample
 import EvmAsm.Rv64.RLP.SchemaDecodeEncoded
 import EvmAsm.Rv64.RLP.SchemaDecodeEncodedExample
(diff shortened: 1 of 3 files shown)


pirapira:
Looks good. Merging

PR #9309, opened 2026-06-23
feat(rv64/rlp): empty (n=0) scalar field region unit (Phase 5, step 49)

dhsorens:
## Summary

First step toward **empty-field support** in the schema engine. A zero scalar (`nonce=0`, `value=0`, `v=0`) RLP-encodes as the empty string `[0x80]`; the non-empty scalar region unit requires `1 ≤ data.length` (its big-endian read loop has no zero-trip spec), so it cannot decode a zero field into the engine's shared output `bytesRegion`.

`unified_empty_scalar_field_decode_and_store_region` is the region analog of the merged cell-based `unified_scalar_field_decode_and_store_zero`: the header descent on `0x80` already leaves `x13` at the next field (empty payload) and `x11 = itemLenRegion = 0`, so there's nothing to read — descend, then spill the value `0` (held in `x11`) into the output region at byte offset `di0` via the scalar store-region leaf (`spillRange out 0 di 8`). Coincides with `decodeScalar (bs.drop O) = some (0, tail)`. `fieldSize = 248` — 32 bytes shorter than the non-empty scalar unit (no read loop).

Part of the maintainer-approved plan to add empty (`n=0`) field support via a minimal-disruption, data-dependent design (no `FieldSpec` structural change).

## Verification
- `lake build EvmAsm.Rv64.RLP.UnifiedEmptyScalarField` + umbrella — clean, 0 warnings.


pirapira:
Looks good to me. Merging.

PR #9319, opened 2026-06-23
feat(rv64/rlp): Phase 6 read_input ⨾ decode (steps 54-55)

dhsorens:
## Summary

Starts **Phase 6** — the top-level pipeline `read_input → decode → write_output` wiring the verified RLP decoder into the zkVM guest's host-I/O ABI. This PR delivers the **read ⨾ decode** half (two theorems):

1. **`rlp_phase6_read_input_ptr`** (hand-off): composes the `read_input` Phase-4 wrapper with `LD x13, x12, ptr_ptr_off`, loading the returned `inputBufBase` into `x13` (the decoder's input pointer).
2. **`rlp_phase6_read_and_decode`** (read ⨾ decode): composes the hand-off with `decode_encoded_short_list_schema_values`. From the host-ABI input contract — `inputBufBaseIs buf_base`, `privateInputIs input`, and `bytesRegion buf_base input` (the host placed the RLP input at the aligned, readable buffer) — with `input = encode (.list (schemaItems specs)) ++ tail`, the `read_input` syscall + `LD` + decoder run end to end: the record is decoded into `bytesRegion outBase (schemaOut out specs)`, with each field's value recovered (`schemaScalarValues`).

Per the approved Phase 6 plan: input-buffer contract = host-ABI precondition; first scope = generic decode-to-output demo. Next PR: the `write_output` wrapper + `decode ⨾ write_output`, then the full pipeline.

## Verification
- `lake build` + umbrella `EvmAsm.Rv64.RLP` — clean, 0 warnings.
- `scripts/check-forbidden-tactics.sh` OK (no `bv_decide`/`native_decide`).
- `#print axioms` (both theorems) → `[propext, Classical.choice, Quot.sound]`; 0 sorry/admit.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9321, opened 2026-06-23
feat(rv64/rlp): Phase 6 readBytes ↔ bytesRegion bridge (step 56)

dhsorens:
## Summary

The keystone of Phase 6's output half (stacked logically on the now-merged read⨾decode, #9319). `write_output` appends `readBytes ptr size` (byte-granular) to the public output, while the decoder leaves its result as a `bytesRegion` (dword-packed via `packBytes`). This bridges them:

- **`getByte_of_bytesRegion`**: when `bytesRegion regionBase bs ** R` holds, the byte at `regionBase + i` is `bs[i]` (extract the containing dword via `bytesRegion_dword_at`, unpack via `extractByte_packBytes`).
- **`readBytes_of_bytesRegion`**: when `bytesRegion base bs` holds, `readBytes base bs.length = bs` (offset induction over `getByte_of_bytesRegion`).

This connects the decoder's `bytesRegion` output to what `write_output` emits, so a later `decode ⨾ write_output` step can conclude `publicValues = decoded output`.

## Verification
- `lake build` + umbrella — clean, 0 warnings; `scripts/check-forbidden-tactics.sh` OK.
- `#print axioms` (both) → `[propext, Classical.choice, Quot.sound]`; 0 sorry/admit.


pirapira:
Looks good to me. Merging.

Issue #9373, opened 2026-06-24
Specialized RLP decoders for untrusted inputs


pirapira:
`RV64/RLP` now contains a decoder for RLP.

The guest
* needs a decoder with a spec that doesn't assume the input is valid. The spec needs to cover both cases (1) when the input is a valid RLP of expected form and (2) otherwise, fails (returns non-zero on `a0` register).
* wants decoders that deal with special shapes of Rlp (transactions, mpt nodes, etc). The decoder just needs to check the shape of the RLP and read out the informative pieces in a fixed format.

This issue keeps track of proving cps triples these specialized decoders for untrusted inputs, for all variations used in the guest.

pirapira:
When a specialized decoder expects a list of length two, it shouldn't check the exact length of the list when it contains more than two elements. Instead it should just abort after seeing more than two elements. So "decode a list of unknown length" won't be needed.

pirapira:
#9503 implements the unverified versions. I'll specify and prove a drop-in replacement of at least one of these.

dhsorens:
## Status at `f437a8cd8`: premise still true, but this is now an **umbrella** and `PLAN.md` is its live tracker

The literal claim holds — there is still no completed whole validating decoder over untrusted input. But a lot has landed *under this issue's number*, enough that the issue text no longer describes the remaining work.

**Phase A — done.** The pure rejection bridges: `EvmAsm/EL/RLP/{ByteStringDecodeBridge,ListDecodeBridge,ReadLength,FullDecode}.lean`, giving per-class `…_eq_some_iff` **and** `…_eq_none_of_*` characterizations (OOB length, leading-zero / `≤55` non-canonical long form, single-byte-not-wrapped, list leftover) — the verdicts a validating decoder has to discharge (`PLAN.md:2982-2988`).

**Phase B — in progress**, 10 modules landed under `EvmAsm/Rv64/RLP/`, each docstring naming #9373:

`UnifiedDecodeItemSingleByteValidated`, `…ShortBytesValidated`, `…ShortBytesValidatedAt`, `…SingletonValidated`, `…ShortListValidated`, `SingleByteListLoopValidated`, `ValidatingFieldWalk`, `ValidatingFieldStep`, `ValidatingBytesCopy`, `ValidatingExactArity`.

[...]

PR #9457, opened 2026-06-25
feat(rv64/rlp): taken-side branch⨾triple combinator for the field walker (#9373 F1)

dhsorens:
## Summary

Composition glue for the untrusted RLP field walker (F1 of the verified guest-decoder plan, #9373).

The validating single-item decoders are 2-exit `cpsBranchWithin`s whose SUCCESS is the *taken* exit and FAIL the fall-through. To advance to the next field after a successful decode, the pointer-advance instructions must be sequenced on the **taken** exit — but the existing `cpsBranchWithin_seq_cpsTripleWithin_same_cr` only sequences on the *fall* side.

`cpsBranchWithin_seq_cpsTripleWithin_taken` (`ValidatingFieldWalk.lean`) is the missing dual: continue the taken branch into a follow-on triple (keeping the fall exit as the abort path), with a CodeReq union. The reusable step that turns "validating-decode at offset O" into "validating-decode-and-advance".

## Test plan
- `lake build EvmAsm.Rv64.RLP` ✅; axiom-clean; 0 sorry/0 warnings; forbidden-tactics clean.


pirapira:
I think the theorem is misplaced. This is not specific to RLP.

PR #9461, opened 2026-06-25
feat(rv64/rlp): validating shortBytes decode-and-advance (#9373 F1)

dhsorens:
## Summary

Part of the F1 keystone (#9373): the untrusted RLP field walker. Adds the **decode-and-advance** step — the reusable unit that turns "validate the item at offset O" into "validate it *and* move the cursor to the next field."

`rlp_decode_shortBytes_advance_at` runs the offset-general validating shortBytes decoder (`rlp_decode_shortBytes_validated_at`, #9456) and, on its SUCCESS (taken) exit, sequences `ADD x13, x13, x11` to advance the cursor `x13` past the just-decoded payload:

- **SUCCESS** (taken): `x13 = (regionBase + O) + 1 + payloadLen` (item start + prefix + payload = next item start), carrying `⌜decode (pfx :: rest) = some (.bytes …)⌝`.
- **FAIL** (fall): the decoder's abort exit unchanged (`⌜decode = none⌝`).

Built via `cpsBranchWithin_seq_cpsTripleWithin_taken` (#9457) — the ADD is framed with the 8-atom complement (incl. the pure verdict) and reshaped with `xperm_hyp`.

## Verification
- `lake build` clean; 0 sorry / 0 warnings.
- `#print axioms`: only `propext`, `Classical.choice`, `Quot.sound`.
- forbidden-tactics scan green.


pirapira:
Looks still okay, merging.

I'm watching for one possibility that the same input RLP is parsed multiple times. Once for checking that the RLP is valid; another time for copying out information. I want to avoid these repetitions.

dhsorens:
Thanks! And good call on the single-pass concern — that's exactly the trap I want to avoid too.

To be concrete about the plan: the validating walk will **validate and copy out in the same pass**, not validate-then-reparse. Each field-step is `validate-item → store-its-value → advance-cursor`, all on one forward sweep:

- The validating arm already hands back *where* the payload lives (`decode (bs.drop O) = some (.bytes (rest.take payloadLen), …)` ⇒ the payload is the `payloadLen` bytes at `cursor+1`), so the store step copies straight from that window — no second decode.
- The accumulated `⌜decode … = some …⌝` verdicts are what tie the stored bytes to the pure RLP spec, so correctness still falls out of one traversal.

So I'm **dropping** the earlier "establish `SchemaValid`, then run `schema_walk` to extract" framing (that was the two-pass version). `schema_walk` stays useful as the *pure* coincidence target, but the emitted program walks the input once. The merged decode-and-advance unit here is that single sweep's core; the next PR inserts the per-field store between validate and advance.

This PR (decode-and-advance + clean-cursor form) is single-pass-compatible as-is; I also pushed a commit dropping the two `set_option maxRecDepth` lines the bot flagged (proofs go through unchanged).

PR #9487, opened 2026-06-26
feat(el/rlp): pure Withdrawal type + decodeWithdrawal spec (#9373 T2)

dhsorens:
## Summary

The pure decode spec that the verified `withdrawal_decode` drop-in (T2 of #9373) will be proven to coincide with. Defining the small pure target is part of the T2 work (the plan calls for one pure decode spec per verified function).

An EIP-4895 withdrawal is RLP-encoded as a 4-element list `[index, validator_index, address, amount]`:
- `index`, `validatorIndex`, `amount` — u64 scalars (read big-endian);
- `address` — exactly 20 bytes.

`decodeWithdrawal` decodes the **whole** input (`decodeFully` — no trailing bytes) as such a list and returns `none` on any structural deviation (wrong arity, a nested list where bytes are expected, a non-20-byte address, trailing bytes).

- `structure Withdrawal` (`index`/`validatorIndex`/`amount : Nat`, `address : List Byte`)
- `decodeWithdrawal : List Byte → Option Withdrawal`
- `decodeWithdrawal_eq_some_iff` — the defining characterization for the drop-in's coincidence proof.

Pure Lean, no separation logic. Axiom-clean (`propext` only); `bv_decide`/`native_decide`-free.

### DIFF
--- a/EvmAsm/EL.lean
+++ b/EvmAsm/EL.lean
@@ -161,6 +161,7 @@ import EvmAsm.EL.TransactionCall
 import EvmAsm.EL.TransactionExecution
 import EvmAsm.EL.TransactionExecutionShape
 import EvmAsm.EL.Block
+import EvmAsm.EL.Withdrawal
 import EvmAsm.EL.BlockTrace
 import EvmAsm.EL.Bls12G2AddInputBridge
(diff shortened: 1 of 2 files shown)


pirapira:
This is very interesting. Merging.

PR #9486, opened 2026-06-26
feat(rv64/rlp): single-pass validated scalar decode-and-store (#9373 T1/T2)

dhsorens:
## Summary

The per-field **single-pass validated scalar decode-and-store** that `rlp_field_to_u64` and `withdrawal_decode` repeat. Continues the single-pass design (validate AND copy out in one sweep; no reparse — @pirapira on #9461).

`rlp_decode_shortBytes_scalar_store_at` composes:
- `rlp_decode_shortBytes_scalar_at` (#9483, single-pass validate + big-endian read), with
- one `SD rOut, x11, offset` on the validating branch's taken exit — writing the decoded value to a fixed slot of an output struct (`outBase + offset`).

One forward sweep: **validate → read value → store**.
- **SUCCESS** (taken): the output cell holds `Nat.fromBytesBE payload`, carrying `⌜decodeScalar (bs.drop O) = some (that value, …)⌝`.
- **FAIL** (fall): the decoder's abort exit (`⌜decode = none⌝`), output slot untouched.

`rOut` is a callee-saved output-base register (set up by the LP64 wrapper), distinct from the decoder's `x5/x10..x15` working set. Mirrors the valid-input `unified_scalar_field_decode_and_store` onto the validating taken exit.

## Verification
- `lake build` clean; 0 sorry / 0 warnings.
- `#print axioms`: only `propext`, `Classical.choice`, `Quot.sound`.
- forbidden-tactics scan green; no `set_option maxRecDepth`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9492, opened 2026-06-26
feat(rv64/rlp): single-pass validated fixed-length byte-array copy (#9373 T2/T3)

dhsorens:
## Summary

The byte-array analog of `rlp_decode_shortBytes_scalar_store_at` (#9486) — the field primitive for fixed-length byte fields like `withdrawal_decode`'s 20-byte `address` (and the header hash fields). Continues the single-pass design: validate **and** copy in one forward sweep, no reparse.

`rlp_decode_shortBytes_bytescopy_at` composes:
- the offset-general validating shortBytes decoder (`rlp_decode_shortBytes_validated_at`, #9456), which on SUCCESS leaves the payload pointer in `x13` and length in `x11`; with
- the existing byte-copy engine (`unified_field_bytes_copy`), which streams the `N` payload bytes into the output slot `outBase + di0`.

- **SUCCESS** (taken): the output region's slot `[di0, di0+N)` holds the payload bytes (`copyRangeGen …`), carrying `⌜decode (pfx :: rest) = some (.bytes (rest.take N), rest.drop N)⌝`.
- **FAIL** (fall): the decoder's abort exit (`⌜decode = none⌝`), output untouched.

`rOut` is a callee-saved register holding the output base (distinct from the decoder's working set); `N` (= the validated payload length) is the field's fixed width — the surrounding decoder supplies the runtime width check that pins `payloadLen = N`. With #9486 (scalar store) and this, **all of `withdrawal_decode`'s field types are covered**.

## Verification
- `lake build` clean; 0 sorry / 0 warnings.
- `#print axioms`: only `propext`, `Classical.choice`, `Quot.sound`.
- forbidden-tactics scan green; no `set_option maxRecDepth`.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -148,3 +148,4 @@ import EvmAsm.Rv64.RLP.ValidatingFieldWalk
 import EvmAsm.Rv64.RLP.ValidatingFieldStep
 import EvmAsm.Rv64.RLP.ValidatingScalarRead
 import EvmAsm.Rv64.RLP.ValidatingScalarStore
+import EvmAsm.Rv64.RLP.ValidatingBytesCopy
--- a/EvmAsm/Rv64/RLP/ValidatingBytesCopy.lean
+++ b/EvmAsm/Rv64/RLP/ValidatingBytesCopy.lean
@@ -34,6 +34,36 @@ private theorem bytescopy_ptr_eq (regionBase : Word) (O : Nat) :
   rw [show signExtend12 (1 : BitVec 12) = (1 : Word) from by decide,
       show (1 : Word) = BitVec.ofNat 64 1 from rfl, BitVec.add_assoc, ← BitVec.ofNat_add]
 
+/-- Sequence a branch onto the TAKEN exit of a branch, converging the follow-on branch's TAKEN exit
+    with the first branch's fall exit onto a shared fail exit `exitF`. Used to bolt a length-equality
+    check onto a validating decode's success so that *both* a malformed field (the decoder's bound
+    failure) and a wrong-length field converge on the one `a0 ≠ 0` exit. -/
+theorem cpsBranchWithin_seq_cpsBranchWithin_taken_conv {n1 n2 : Nat}
+    {entry mid succ exitF : Word} {cr1 cr2 : CodeReq} (hd : cr1.Disjoint cr2)
+    {P Q_t1 Q_f1 Q_t2 Q_succ Q_f : Assertion}
+    (h1 : cpsBranchWithin n1 entry cr1 P mid Q_t1 exitF Q_f1)
+    (h2 : cpsBranchWithin n2 mid cr2 Q_t1 exitF Q_t2 succ Q_succ)
+    (hf1 : ∀ h, Q_f1 h → Q_f h) (hf2 : ∀ h, Q_t2 h → Q_f h) :
+    cpsBranchWithin (n1 + n2) entry (cr1.union cr2) P succ Q_succ exitF Q_f := by
+  intro R hR s hcr hPR hpc
+  rw [CodeReq.union_satisfiedBy hd] at hcr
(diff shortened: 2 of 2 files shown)


pirapira (commented):
This is very important so I'll have a deeper look when I reach my laptop

pirapira (changes requested):
Please add an argument for the expected number of bytes. Make it a failure when the byte array has an unexpected length.

pirapira:
The expected length should be a Lean argument.

pirapira:
(Copying longer than needed is a dos vector. Also the destination usually has only space for the expected length)

pirapira:
There should be two versions. One that expects the exact length (for addresses for example). And one that expects the max length (for data of txs for example).

dhsorens:
Thanks — that's a good catch, and the DoS framing is exactly right: an attacker-controlled `payloadLen` must never drive the copy length.

Reworking accordingly:
- `expectedN : Nat` becomes a **Lean argument** (the fixed slot width).
- The copy streams **exactly `expectedN`** bytes — never `payloadLen` — so it can't over-run the destination slot.
- A runtime check rejects a field whose decoded length isn't `expectedN`: after the validating decode yields `payloadLen` in `x11`, a `BNE x11, <expectedN>` routes a mismatch to the same failure exit as a malformed field, so the SUCCESS exit holds only when `payloadLen = expectedN` (and the verdict becomes `decode = some (.bytes data)` with `data.length = expectedN`).

So both failure modes — malformed prefix/bound *and* wrong length — converge on the one `a0 ≠ 0` exit. Pushing the update to this branch shortly.

dhsorens:
Progress on the expected-length rework (pushed to this branch):

- `rlp_shortBytes_length_check` — the length-equality check: `ADDI x10,x0,expectedN ; BNE x11,x10` bolted onto the validating decode's success. Taken (`payloadLen ≠ expectedN`) → fail; fall (`payloadLen = expectedN`) → continue. `expectedN` is a Lean argument.
- `cpsBranchWithin_seq_cpsBranchWithin_taken_conv` — converges the wrong-length failure with the malformed-field failure onto the one `a0 ≠ 0` exit.
- Supporting: `sepConj_merge_two_pures`, `signExtend12_ofNat_small`, `ofNat_eq_iff_small`.

All build, axiom-clean. The final wiring (the byte-copy now copies **exactly `expectedN`** bytes — never the attacker-controlled `payloadLen`, addressing the DoS — gated on the length check, with `decode` re-derived on the success path) is in progress.

On the bot's `e2_target` naming note: that parameter name follows the convention used across all the existing RLP arm files (`UnifiedDecodeItem*Validated*.lean`, `Phase1E*FullPath.lean`); renaming only here would diverge from the established codebase style, so I've kept it consistent.

pirapira:
I'll have a look in 90 mins.

dhsorens:
Done — the rework is complete and pushed. `rlp_decode_shortBytes_bytescopy_at` now:

- Takes **`expectedN : Nat`** as a Lean argument (the fixed slot width).
- Copies **exactly `expectedN`** bytes — never the attacker-controlled `payloadLen` — so it can't over-run the destination slot (the DoS guard).
- **Fails at runtime when the decoded length isn't `expectedN`**: `ADDI x10,x0,expectedN ; BNE x11,x10` gates the copy; a wrong-length field and a malformed field both converge on the single `a0 ≠ 0` exit.

SUCCESS: the output slot `[di0, di0+expectedN)` holds the payload with `⌜payloadLen = expectedN⌝`. FAIL (fall): `⌜decode = none ∨ payloadLen ≠ expectedN⌝`, output untouched.

Builds clean; `#print axioms` shows only `propext`/`Classical.choice`/`Quot.sound`; no `set_option maxRecDepth`; `bv_decide`/`native_decide`-free.

(On the bot's `e2_target` note: kept consistent with the codebase-wide parameter name used across all the RLP arm files, as mentioned above.)

pirapira:
Looks great. Merging.

pirapira (approved):
Looks great, merging.

PR #9495, opened 2026-06-26
fix(exp): per-limb counter clobbered by mul_callable — relocate x6→x20 (callee-saved)

pirapira:
## Root-cause bug

The EXP fixed square-and-multiply loop held its **per-limb counter** in `x6`
(RISC-V `t1`, **caller-saved**). The iteration body decrements `x6` in the
bit-test, then calls `mul_callable` twice (squaring + conditional multiply)
via `JAL`. `mul_callable`'s footprint is `{x5,x6,x7,x10,x11}` — it **clobbers
`x6`** and there is no save/restore. So from the second iteration onward the
`BNE .x6, x0` limb-reload decision branched on a leftover product limb, not
the counter.

Consequence for the proof: the loop invariant `x6.toNat = 64 - k % 64` was
**genuinely false** against the executable at each iteration entry — which is
why the 256-iteration loop-body induction (`evm_exp_stack_spec_within`) could
not be closed by any route. The cursor `x19` was already correctly placed in a
callee-saved register ("`x19` is callee-saved so it survives `mul_callable`");
the counter was simply left in a caller-saved one.

## Fix

Relocate the per-limb counter to **`x20`** (callee-saved, outside
`mul_callable`'s footprint, so preserved by the frame rule — exactly like the
`x19` cursor). In the program this is a pure register rename
(`exp_prologue_fixed`, `exp_msb_bit_test_block_fixed`), **offset-preserving**:
all instruction counts and branch offsets are unchanged.

The spec chain is updated to match (≈18 files): counter-role `x6` → `x20`;
`x6` becomes pure `mul_callable` scratch (`regOwn`); and crucially the loop /
iteration posts that previously released `regOwn .x6` (counter lost) now carry
`.x20 ↦ c6New` (counter **preserved**, threaded through the MUL frames). The
DEF-layer signatures (`expTwoMulFixedIterPre`, etc.) are kept stable, so the
~162 downstream uses are unaffected.

## Result

`lake build EvmAsm.Evm64.Exp` is green (1105 jobs); `check-forbidden-tactics`
passes; no `sorry`/`native_decide`/`bv_decide`. The loop invariant
`x20 = 64 - k % 64` is now provable at each iteration entry, which unblocks the
EXP loop-body induction and the eventual `evm_exp_stack_spec_within`
(follow-up; this PR is the correctness fix + spec realignment).

Directed by @pirapira; implemented by Claude Code.

### DIFF
--- a/EvmAsm/Evm64/Exp/Compose/FixedLoopInd.lean
+++ b/EvmAsm/Evm64/Exp/Compose/FixedLoopInd.lean
@@ -105,12 +105,12 @@ theorem cpsTripleWithin_expReloadDirectFalsePre_ordinary_vacuous
     {n : Nat} {entry exit_ : Word} {code : CodeReq} {Q : Assertion}
     {k : Nat} {baseWord exponentWord : EvmWord}
     {controlC6 e iterCount nextLimb ptr nextNextLimb sp evmSp
-      r0 r1 r2 r3 a0 a1 a2 a3 v6' v7' v10' v11' d0' d1' d2' d3' base : Word}
+      r0 r1 r2 r3 a0 a1 a2 a3 v7' v10' v11' d0' d1' d2' d3' base : Word}
     (hC6 : controlC6 + signExtend12 (-1 : BitVec 12) ≠ 0) :
     cpsTripleWithin n entry exit_ code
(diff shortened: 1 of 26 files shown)


pirapira:
This touches many files, so merging already.

PR #9508, opened 2026-06-27
feat(rlp): verified drop-in rlp_content_to_u256_be (failure path)

pirapira:
## What

First slice of a track that replaces the codegen stateless guest's
**unverified** RLP decode helpers (emitted as raw assembly strings in
`EvmAsm/Codegen/Programs/RlpWalk.lean`) with verified `Program`s carrying
**caller-facing** Hoare-triple specs.

This PR targets the smallest guest decoder that writes an **output memory
region**: `rlp_content_to_u256_be` (the right-align-into-32-byte-BE-u256 half of
`rlp_field_to_u256_be`). It adds `EvmAsm/Rv64/RLP/ContentToU256Be.lean` with:

- `rlp_content_to_u256_be_prog` — the faithful **21-instruction** drop-in body
  (matches the guest's LP64 calling convention exactly, so it is a literal
  drop-in once codegen is rewired);
- `rlp_content_to_u256_be_too_long_spec_within` — a **complete** leaf-function
  Hoare triple for the content-too-long **failure path** (`len > 32`),
  `cpsTripleWithin 8 base (raVal &&& ~~~1)`.

## Why this shape of spec

The spec is written to be informative for callers, per the three things a
drop-in contract must state:

- **How to give inputs** — `a0`/`a1`/`a2` = content ptr / content length /
  output ptr (LP64; see `EvmAsm/Evm64/CallingConvention.lean`).
- **Where outputs are located** — status in `a0` (`2` = too long), and the
  32-byte big-endian `u256` in the 4 dwords at `a2`.
- **What can happen to the output memory region on failure** — the caller
  still **owns** a writable 32-byte region at `a2`, but its contents are
  **arbitrary**. This is expressed with `memOwnU256` (four `memOwn` dwords), so
  the contract deliberately under-specifies the bytes and callers must not read
  them. (This particular body zeroes the buffer before the length check; the
  weaker contract keeps any conforming implementation a valid drop-in.)

## Scope (intentionally small — first example for feedback)

Only the **failure** path (`len > 32`) is proved here — it is loop-free and
exercises every part of the caller contract above. The **success** path
(`len ≤ 32`, the right-aligned big-endian copy loop) is a follow-up that will
reuse the existing `ByteCopyIter` / `ByteCopyChain` counted-copy infrastructure.

## Verification / gates

- Axiom-clean: `#print axioms` → `[propext, Classical.choice, Quot.sound]`; **0 sorry**.
- `lake build` green (full, 3540 jobs); `check-forbidden-tactics`,
  `check-unimported`, `check-file-size`, `check-no-warnings` all pass.
- Wired into the `EvmAsm.Rv64.RLP` umbrella; `PLAN.md` gains an EL.5 subsection.

Bead: `evm-asm-cmz21`

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -151,3 +151,4 @@ import EvmAsm.Rv64.RLP.ValidatingScalarStore
 import EvmAsm.Rv64.RLP.ValidatingBytesCopy
 import EvmAsm.Rv64.RLP.ValidatingExactArity
 import EvmAsm.Rv64.RLP.UnifiedDecodeItemShortListValidated
+import EvmAsm.Rv64.RLP.ContentToU256Be
--- a/EvmAsm/Rv64/RLP/ContentToU256Be.lean
+++ b/EvmAsm/Rv64/RLP/ContentToU256Be.lean
@@ -1,0 +1,241 @@
+/-
+  EvmAsm.Rv64.RLP.ContentToU256Be
+
+  A verified RISC-V leaf subroutine that is a **drop-in replacement** for the
+  (currently unverified) codegen guest function `rlp_content_to_u256_be`
+  emitted as a raw assembly string by
+  `EvmAsm/Codegen/Programs/RlpWalk.lean` (`rlpContentToU256BeFunction`).
+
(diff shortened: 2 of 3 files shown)


pirapira:
Looks good. Merging.

PR #9510, opened 2026-06-27
feat(rlp): make decodeScalar canonical-strict (reject non-canonical integers)

pirapira:
## What

Repairs the pure RLP scalar decoder `EvmAsm.EL.RLP.decodeScalar` to be
**canonical-strict**, matching execution-specs `ethereum_rlp._deserialize_to_uint`:

```python
if len(decoded) > 0 and decoded[0] == 0:
    raise DecodingError("non-canonical integer")
```

Before, `decodeScalar` read **any** byte string as `Nat.fromBytesBE data` — it
accepted leading-zero (non-canonical) integer encodings that execution-specs
rejects (e.g. `0x00`, or `0x82 0x00 0x05`). Now it rejects them: a byte string
is a valid scalar iff `data.headD 1 ≠ 0` (the empty string stays the canonical
`0`; a nonzero value's high byte must be nonzero). The round-trip
(`decodeScalar_encodeScalar`), injectivity, and right-inverse lemmas are
re-proved for the strict decoder (the round-trip still holds because
`encodeScalar` is minimal big-endian, hence canonical).

## Why

The whole verified RLP scalar layer was lenient on the leading-zero rule (only
the structural rules — single-byte form, long-form length, zero = empty string —
were enforced). That's a consensus-leniency gap: a zkVM guest using it would
accept blocks/txs that the canonical client rejects. `decodeScalar` is the
right place for the fix (it mirrors `_deserialize_to_uint`).

## Scope of the cleanup

Making `decodeScalar` strict falsifies every lemma that asserted the lenient
verdict `decodeScalar … = some (Nat.fromBytesBE data, …)` for arbitrary `data`.
Per maintainer direction these lenient lemmas are **removed** rather than
preserved under an alias (there's no use for a lenient scalar decoder). Because
that verdict was load-bearing for the **Phase-6 top-level decode pipeline** and
the scalar field-walk/region/validating decoders, those came out too:

- **Removed (22 files):** `SchemaScalarValues`, `SchemaDecodeValues`
  (+`…Encoded`/`…Long`/example files), `Unified{TwoScalarFieldWalk,
  ThreeFieldWalk,WideScalarField,ScalarFieldStore,ScalarFieldRegionRegOwn,
  NScalarFieldWalk}`, `Validating{Scalar}Read/Store`, `ScalarFieldWalk{Chain,
  Infra}`, `Phase6Pipeline`, `Schema{List,LongList}*` examples, … — plus their
  umbrella/cross imports.
- **Repaired in place:** the generic Phase-6 read_input/write_output specs that
  had relied on now-removed transitive imports (`Phase6ReadDecode` /
  `Phase6DecodeWrite` drop a vestigial `open EvmAsm.EL.RLP`; `Phase6WriteOutput`
  regains `import EvmAsm.Rv64.MemRegion` + a local `Assertion.PCFree
  (bytesRegion …)` instance).

Nothing outside `EvmAsm/{EL,Rv64}/RLP` consumed the removed theorems (no codegen
guest / conformance / stateless dependency).

## Follow-up (Step 3)

Rebuild a canonical-strict scalar decode + Phase-6 pipeline (decoders that
actually *reject* non-canonical input and prove their verdict against the strict
`decodeScalar`), plus a `decodeHash`/`decodeBytesN` for fixed-width byte fields
(exact length, leading zeros allowed). The canonical-strict
`rlp_content_to_u256_be` routine is tracked separately.

## Verification

- Full `lake build` green (3518 jobs).
- `check-unimported` (0 orphans), `check-no-warnings` (0), `check-forbidden-tactics`,
  `check-file-size` all pass.


pirapira:
Scalar decoders that lack canonicity check won't be used, so this PR removes these.

PR #9515, opened 2026-06-28
feat(rlp): rlp_walk_next CPS drop-in (WIP)

pirapira:
> Third in the stacked sequence of CPS-verified drop-ins for the cursor-walk RLP
> decode functions added in #9503/#9504 (`RlpWalk.lean`). Stacked on #9514 (in `main`).

## Goal

A verified CPS drop-in for the codegen guest `rlp_walk_next` (`RlpWalk.lean`, #9503):
decode the single RLP item at the cursor, advance past it, report its content length.

## Program + proofs (all `sorry`-free, axiom-clean: propext/Classical.choice/Quot.sound)

64-instruction body `rlp_walk_next_prog`, dispatching on the prefix byte; the two long
forms read the big-endian length field via a `slli/lbu/or` accumulation loop.

- Six per-form cases: `…_end_spec_within` (`cursor ≥ end` → `a1=2`), `…_single_spec_within`
  (`p<0x80`), `…_short_string_spec_within` (`0x80≤p<0xb8`), `…_long_string_spec_within`
  (`0xb8≤p<0xc0`, loop `wn_ls_loop`), `…_short_list_spec_within` (`0xc0≤p<0xf8`),
  `…_long_list_spec_within` (`p≥0xf8`, loop `wn_ll_loop`).
- Unified `…_spec_within`: one `cpsTripleWithin`, static preconditions, step bound `78`,
  six-way postcondition disjunction.

## ⚠ Strictness gap vs execution-specs — additional failure checks required

This proves the codegen guest **as written**, which is a *trusting* single-item stepper.
Comparison against Python execution-specs (`ethereum_rlp/rlp.py`) shows the guest accepts
inputs execution-specs **rejects** with `DecodingError`. Confirmed in #9515's proof: on the
long forms, `a2 = full span = 1 + lol + decoded` and `a0 = cursor + span` are reported
**unconditionally** — `endPtr` is consulted only by the initial `bgeu`, so a span (or even a
length field) that runs past `end` — *including a length declared in the parent node* — is
**not** rejected (`a0` can exceed `end` or wrap mod 2⁶⁴).

For a consensus-correct (strict, no-lenient-variant) drop-in, the program must gain four
checks, each a **new failure status** (more reasons to fail):

1. **Span-fits / bound** — reject when `advanced > endPtr` (item/length-field overruns the
   available bytes). execution-specs: `…_end_idx - 1 >= len(encoded) → raise`.
2. **Long-form minimal length** — reject long string/list with `decoded < 56` (`0x38`).
   execution-specs: `if len_decoded_data < 0x38: raise`.
3. **Length-field leading zero** — reject first length byte `== 0`.
   execution-specs: `if encoded_bytes[1] == 0: raise`.
4. **Single-byte short-string canonicality** — reject a 1-byte short string with
   `content[0] < 0x80` (needs a content read). execution-specs:
   `if len_raw_data == 1 and raw_data[0] < 0x80: raise`.

(1)–(3) are computable from prefix + length field + `end` the walker already holds; (4)
needs a content-byte read. These are **structural** canonicality/bounds, distinct from the
**scalar** canonicality (leading-zero of the integer value) already enforced in
`ContentToU256Be`/`ContentToU64`. Adding them changes the program (extra compares + fail
blocks) and the proofs (extra failure disjuncts). See the module docstring for the full
write-up.

## Verification

Full `lake build` green; `check-no-warnings`, `check-forbidden-tactics`, `check-unimported`,
`check-file-size` all pass.

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -132,3 +132,4 @@ import EvmAsm.Rv64.RLP.UnifiedDecodeItemShortListValidated
 import EvmAsm.Rv64.RLP.ContentToU256Be
 import EvmAsm.Rv64.RLP.ContentToU64
 import EvmAsm.Rv64.RLP.WalkInit
+import EvmAsm.Rv64.RLP.WalkNext
--- a/EvmAsm/Rv64/RLP/WalkNext.lean
+++ b/EvmAsm/Rv64/RLP/WalkNext.lean
@@ -1,0 +1,197 @@
+/-
+  EvmAsm.Rv64.RLP.WalkNext
+
+  A verified RISC-V leaf subroutine: a CPS drop-in for the codegen guest function
+  `rlp_walk_next` emitted by `EvmAsm/Codegen/Programs/RlpWalk.lean`
+  (`rlpWalkNextFunction`, added in #9503's cursor-walk RLP decode work).
+
+  `rlp_walk_next` decodes the single RLP item at the cursor, advances the cursor
+  past it, and reports the item's content length. The content pointer is derived
+  by the caller as `advanced_cursor - content_length`. It is a list/item-structure
+  operation — it decodes no scalar value — so the scalar-canonicality rule
+  (enforced in `ContentToU256Be`/`ContentToU64`) does not apply here; this is a
+  *faithful* drop-in of the guest's per-form classification.
+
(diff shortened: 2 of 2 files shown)


pirapira:
There should be a guard against the item length is declared so big it goes over `endPtr`.

pirapira:
superseded by #9519

PR #9516, opened 2026-06-28
feat(rlp): make rlp_walk_init EXACT (bound + structural canonicality, distinct fail codes)

pirapira:
> Strict (execution-specs-equivalent) outer-list header parse for `rlp_walk_init`.
> Tracked by bead `evm-asm-l656f`. Follow-up to #9514 (faithful walk_init, merged).

## Why

The merged `rlp_walk_init` is a *trusting* parser: it positions the cursor past the
list header and sets `end = list_ptr + list_len` **without** reading the long-list
length field or checking the header/content fits. Python execution-specs
(`ethereum_rlp/rlp.py` `decode_to_sequence`) is **strict** — it rejects an
over-long / non-canonical list. (See the strictness findings in #9515.) `list_len`
reaches `walk_init` as the **exact** RLP framing, so EXACT is the right consistency check.

## What

Strict EXACT `rlp_walk_init` (53-instr program) with distinct status codes (`a2`):

| code | meaning |
|---|---|
| 0 | ok |
| 1 | not a list (`prefix < 0xc0`) |
| 2 | empty (`list_len == 0`) |
| 3 | short-list length mismatch (`1 + (prefix−0xc0) ≠ list_len`) |
| 4 | long-list header truncated (`1 + lol > list_len`) |
| 5 | long-list length-field leading zero (`len[0] == 0`) |
| 6 | long-list non-minimal (`decoded < 56`) |
| 7 | long-list length mismatch (`1 + lol + decoded ≠ list_len`) |

The long path reads the `lol`-byte big-endian length field and requires
`1 + lol + decoded == list_len`; short path requires `1 + (prefix−0xc0) == list_len`.
Clobbers `t0..t6`; all callers spill `a0/a1/a2` to `s`-regs immediately after the call.

## Status — complete

- [x] Codegen guest `rlp_walk_init` (RlpWalk.lean) — strict, distinct codes.
- [x] **EEST-validated**: 200 random Amsterdam blockchain_tests on spike → **200/200 PASS, 0 fail** (seed 992472044). The EXACT change rejects no valid block.
- [x] Verified `EvmAsm/Rv64/RLP/WalkInit.lean` re-proof — **all 9 outcomes proved** (empty/not-list/short/short-mismatch/long-truncated/leading-zero/non-minimal/long-mismatch/long-ok), length-field loop (`wi_len_loop`), and the **unified dispatch theorem** `rlp_walk_init_spec_within` (9-way, ≤81 steps).
- [x] **Axiom-clean**: `#print axioms rlp_walk_init_spec_within` → `propext`, `Classical.choice`, `Quot.sound` only. No `sorry`/`native_decide`/`bv_decide`.

Spinoff: `evm-asm-ifbx3` (P1) — add a `bytesRegion_lbu_within` variant with a nonzero
immediate offset (would have avoided the idx12 `ADDI;LBU` restructure).

### DIFF
--- a/EvmAsm/Rv64/RLP/WalkInit.lean
+++ b/EvmAsm/Rv64/RLP/WalkInit.lean
@@ -1246,4 +1246,100 @@ theorem wi_long_to_dec (base listBase listLen t0Old t1Old t2Old t3Old t4Old t5Ol
   rw [show (14 + 1 + 2 + (7 * lol + 1)) = 7 * lol + 18 from by ring] at c3
   exact cpsTripleWithin_weaken (fun h hp => by xperm_hyp hp) (fun h hp => by xperm_hyp hp) c3
 
+/-- **long non-minimal** (`prefix ≥ 0xf8`, `decoded < 56`): status `a2 = 6`.
+    (wi_long_to_dec ⨾ idx 24 LI 56 ⨾ idx 25 BLTU taken → idx 49,50.) -/
+theorem rlp_walk_init_lmin_spec_within
+    (base listBase raVal listLen a2Old t0Old t1Old t2Old t3Old t4Old t5Old t6Old : Word)
+    (listBytes : List (BitVec 8)) (listOff : Nat) (hsalign : listBase.toNat % 8 = 0)
+    (hoff : listOff < listBytes.length) (hover : listBase.toNat + listOff < 2 ^ 64)
+    (hvalid : isValidByteAccess (listBase + BitVec.ofNat 64 listOff) = true)
+    (hlen : listLen ≠ (0 : Word))
+    (h_ge : ¬ BitVec.ult ((listBytes[listOff]'hoff).zeroExtend 64) (0xc0 : Word) = true)
+    (h_ge_f8 : ¬ BitVec.ult ((listBytes[listOff]'hoff).zeroExtend 64) (0xf8 : Word) = true)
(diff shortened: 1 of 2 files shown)


EvmAsm/Rv64/RLP/WalkInit.lean
```diff
@@ -104,426 +147,1671 @@ instance (regionBase : Word) (bs : List (BitVec 8)) :
...
-            (.x12 ↦ᵣ (0 : Word)) **
-            ⌜¬ BitVec.ult ((listBytes[listOff]'hoff).zeroExtend 64) (0xf8 : Word) = true⌝) h))) := by
-  by_cases hfail : BitVec.ult ((listBytes[listOff]'hoff).zeroExtend 64) (0xc0 : Word) = true
-  · -- not-a-list (status 1)
-    have ht := cpsTripleWithin_frameR ((.x7 ↦ᵣ t2Old))
-      (by pcFree)
-      (rlp_walk_init_fail_spec_within base listBase raVal listLen a2Old t0Old t1Old listBytes listOff
-        hsalign hoff hover hvalid hfail)
+            (.x11 ↦ᵣ ((listBase + BitVec.ofNat 64 listOff) + listLen)) ** (.x12 ↦ᵣ (0 : Word)) **
+            ⌜True⌝) h))) := by
```
pirapira: These cases need to be more informative instead of ⌜True⌝

PR #9522, opened 2026-06-28
feat(rlp): verified single-pass walk-based withdrawal_decode (WIP)

pirapira:
> **Draft / WIP** — opening early for feedback on the approach. Only the reusable
> walk→spec **decode bridge foundation** has landed so far; the verified
> `withdrawal_decode` program and its `decodeWithdrawal` coincidence proof are still to come.

## Goal

Replace the unverified, O(n²) `withdrawal_decode` (which decodes each of the 4 RLP fields with
a separate `rlp_list_nth_item` that re-walks the list from the front) with a **formally
verified, strict, single-pass** drop-in: `rlp_walk_init` once, then `rlp_walk_next`
left-to-right per field — structurally isomorphic to the pure spec
`decodeWithdrawal` (`decodeFully bs = some (.list [.bytes d0, .bytes d1, .bytes d2, .bytes d3])`).
The spec is anchored on the (now strict) pure `decodeWithdrawal`.

Prereq already merged: **#9521** made `decodeWithdrawal` strict (canonical-minimal `uint64`
scalar fields), so the verified path can reuse the strict `rlp_content_to_u64` directly.

## Landed in this PR so far

`EvmAsm/Rv64/RLP/WalkDecodeBridge.lean` — the reusable bridge between the walk routines'
machine-level item view (a byte list with a `Nat` offset + prefix-range guards, as exposed by
`rlp_walk_next` / `rlpItemDecode`) and the pure spec decoder `EvmAsm.EL.RLP.decodeAux`:

- `drop_eq_cons_of_getElem?` — a byte present at an offset peels off the front of `drop`.
- `decodeAux_singleByte_bridge` — a byte `< 0x80` decodes to `.bytes [b]` (1 byte consumed).
- `decodeAux_shortBytes_bridge` — a prefix in `[0x80, 0xB7]` with its declared content present
  (and, for `len = 1`, a canonical content byte `≥ 0x80`) decodes to `.bytes content`,
  `content = (drop (off+1)).take len`, consuming `1 + len` bytes.

These cover the single-byte + short-string classes — the only forms a fixed `uint64` / 20-byte
address field can take. They leverage the existing `classifyPrefix` / `decodeAux_cons_*`
infrastructure in `EvmAsm/EL/RLP/`. Build clean, 0 `sorry`; axioms = `propext, Quot.sound`.

## Roadmap (remaining)

- **M2 cont.** — outer-list + `decodeItems` bridge: relate `walk_init` + 4× `walk_next` +
  end-of-list to `decodeFully = some (.list [.bytes d0..d3])`.
- **M3** — `withdrawal_decode_prog` + `withdrawal_decode_spec_within`: the single-pass assembly
  proof (`walk_init` + 4× `walk_next` + strict `content_to_u64` + 20-byte copy + arity check),
  composed via a `jal`→leaf→return sequencing pattern (built from the existing
  `jal_spec_within` + leaf specs + `cpsTripleWithin_seq`; no new framework combinator), anchored
  on `decodeWithdrawal`.
- **M4** — swap the `withdrawalDecodeFunction` emission to `emitProgram withdrawal_decode_prog`;
  re-run EEST (200/200 on spike).

## Feedback wanted

- The single-pass walk-based approach + self-contained program structure (appending callee
  blocks vs. shared emission) for M3.
- Scalar width modeling in the spec (`Nat` + `< 2^64` vs. `BitVec 64`).
- Whether the bridge lemmas should live in `WalkDecodeBridge.lean` (reusable for tx/header
  decoders too) or alongside `WalkNext`.

Related: `rlp_field_to_u64` has the same missing-canonicity gap (bead `evm-asm-vd5vr`).

### DIFF
--- a/EvmAsm/Codegen/Programs/Withdrawal.lean
+++ b/EvmAsm/Codegen/Programs/Withdrawal.lean
@@ -17,11 +17,13 @@
 
 import EvmAsm.Rv64.Program
 import EvmAsm.Codegen.Layout
+import EvmAsm.Codegen.Emit
 import EvmAsm.Codegen.Programs.HashBridge
 import EvmAsm.Codegen.Programs.RlpRead
 import EvmAsm.Codegen.Programs.Tx
 import EvmAsm.Codegen.Programs.U256
 import EvmAsm.Codegen.Programs.Block
+import EvmAsm.Rv64.RLP.WithdrawalDecode
 
 namespace EvmAsm.Codegen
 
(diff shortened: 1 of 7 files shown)


pirapira:
Perhaps 9534 goes better.

PR #9535, opened 2026-06-30
Feat/sail zkvm integration

alexanderlhicks:
@dhsorens @pirapira 
There has been some overlap/parallel work it seems, so I;m flagging this for visibility and expect this PR will need some changes to conform with your work (or else, that work will have to be deprecated in favour of this pr).


pirapira:
@dhsorens your call.

pirapira (dismissed):
Please also copy over the LICENSE from the sail-riscv. Maybe under `vendor/sail-riscv-zkvm-lean`.

alexanderlhicks:
Coming back to this PR. Also realised a few commits hadn't been pushed (some proofs) so these are up, now.

pirapira:
This has advantage. Merging when the CI passes.

PR #9537, opened 2026-06-30
docs(ci): fix stale axiom-audit comment — bv_decide axioms are FORBIDDEN

alexanderlhicks:
build.yml's Axiom-audit step comment said "bv_decide axioms allowed", but scripts/check-axioms.sh tightened bv_decide from ALLOWED to FORBIDDEN after https://github.com/Verified-zkEVM/evm-asm/pull/7994 and rejects both bv_decide and native_decide trust-axiom owners against an empty allowlist. The comment now matches the enforced policy; no behavior change.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -94,10 +94,11 @@ jobs:
       - name: DRIFT.md ledger drift check
         run: scripts/check-drift.sh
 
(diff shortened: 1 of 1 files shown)


pirapira:
Sounds correct. Merging.

PR #9539, opened 2026-06-30
chore(bench): two-axis dynamic-cost harness for evm_div/evm_mod

alexanderlhicks:
## What

Adds `bench/DivBench.lean` — a **two-axis dynamic-cost harness** for the verified
256-bit `evm_div` / `evm_mod` subroutines. It runs the verified
`EvmAsm.Rv64.step` semantics on concrete inputs and reports, per divisor class:

- **`steps`** — dynamic instruction (cycle) count — the primary zkVM proving-cost driver
- **`loads` / `stores` / `memOps`** — memory traffic (paging proxy)
- **`dwords`** — distinct 8-byte cells touched (working-set size)
- **`pages`** — distinct 1 KiB data pages touched
- **`correct`** — result checked against `EvmWord.div a b`

It lives **outside `EvmAsm/`**, so it is not pulled into the build/CI graph
(`check-unimported.sh` is unaffected). Run on demand:

```
lake build EvmAsm.Evm64.DivMod.Program EvmAsm.Rv64.Execution EvmAsm.Evm64.Basic
lake env lean bench/DivBench.lean
```

## Why

Phase 1 of a DIV proving-cost optimization effort (prompted by the gas-repricing
question on Yoichi's `mark.repricing` benchmarks: DIV is flat `LOW` = 5 gas
regardless of operand size, but its RISC-V cost varies ~3.4×). The harness
establishes the baseline and pins the objective.

**Baseline** (current `evm_div`, all results verified `correct`):

| divisor | n | steps | memOps | dwords | pages |
|---|---|---|---|---|---|
| `b = 2` / `b = 7` (small) | 1 | 689 / 688 | 201 | 28 | 2 |
| `b ≈ 2⁶⁴` (bench #2) | 2 | 628 | 193 | 28 | 2 |
| `b = 2⁶⁴−1` | 1 | 623 | 179 | 28 | 2 |
| `b ≈ 2¹²⁸` (bench #1) | 3 | 449 | 143 | 28 | 2 |
| `b` full 256-bit | 4 | 204 | 71 | 28 | 2 |

**Findings:**
- The cost curve is **inverted** — small divisors (the common real-world case)
  are the *most* expensive, because the Knuth-D loop runs `5 − n` iterations.
- Memory traffic tracks instruction count linearly (~29% of `steps`), and the
  working set is a **constant 28 dwords / 2 pages** for every divisor. So paging
  does not discriminate by divisor size, and **instruction count is the faithful
  primary objective**.

Cost model per Gassmann, Chaliasos, Sotiropoulos & Su, *"Evaluating Compiler
Optimization Impacts on zkVM Performance"* (arXiv:2508.17518v2): dynamic
instruction count is the near-linear proving-cost driver, per-instruction cost
is ~uniform, and **division is not expensive** in zkVMs.

## Scope

Purely additive measurement tooling — **no change to the verified core.** Draft:
opened for visibility while the follow-on workload-data and algorithm-redesign
phases proceed in separate sessions.

### DIFF
--- a/bench/DivBench.lean
+++ b/bench/DivBench.lean
@@ -1,0 +1,134 @@
+/-
+  bench/DivBench.lean — two-axis dynamic cost harness for `evm_div` / `evm_mod`.
+
+  Runs the VERIFIED `step` semantics (EvmAsm.Rv64.Execution) on concrete inputs
+  and reports, per divisor class:
+
+    * steps   — dynamic instruction (cycle) count   [PRIMARY zkVM cost driver]
+    * loads / stores / memOps — memory-traffic count [2nd axis: paging proxy]
+    * dwords  — distinct 8-byte cells touched        [working-set size]
+    * pages   — distinct 1 KiB data pages touched    [page-in/out proxy]
+    * correct — result checked against `a / b`
+
+  Cost-model rationale (Gassmann et al., arXiv:2508.17518v2, RISC0/SP1):
+  dynamic instruction count is the primary, near-linear proving-cost driver;
+  per-instruction cost is ~uniform ("division is not expensive"); memory is a
+  complementary axis — a paged-in access is 1 cycle but a page-in/out ≈ 1130
+  cycles. This harness surfaces both axes so the optimization objective can be
+  the frequency-weighted (instructions + paging) cost, not instructions alone.
+
+  Run:  lake env lean bench/DivBench.lean
+  (needs `lake build EvmAsm.Evm64.DivMod.Program EvmAsm.Rv64.Execution EvmAsm.Evm64.Basic`)
+-/
+import EvmAsm.Evm64.DivMod.Program
+import EvmAsm.Rv64.Execution
+import EvmAsm.Evm64.Basic
+
+open EvmAsm.Rv64
+open EvmAsm.Evm64
+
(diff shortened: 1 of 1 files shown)


pirapira:
Status note, not a request to close — flagging because this is currently the only open PR in this state.

As of `origin/main` at `ed50a5dbb`:

- opened 2026-06-30 and not updated since
- **5,133 commits behind main**
- `mergeStateStatus: DIRTY` (conflicts)
- unlabelled, so the merge bot has never considered it

Every other PR opened today has landed or is queued, so nothing is blocking on this one either way. Two useful outcomes:

1. **Still wanted** — a rebase would be substantial at this distance, and it may be quicker to re-cut the harness against current `main` than to resolve a month of drift. Worth knowing whether `evm_div`/`evm_mod` benchmarking is still the shape you want, since the DIV/MOD work it targets is recorded as deferred.
2. **No longer wanted** — closing it costs nothing and the branch stays available if the harness is useful later.

Happy to label it for the merge bot if it gets rebased; just say which way you'd like it to go.

alexanderlhicks:
Closing for now, will revisit this in due course when we are specifically focused on optimisations and allocate more time to develop a more general benchmark harness and cost model.

PR #9547, opened 2026-06-30
Perf/div evm workload data

alexanderlhicks:


alexanderlhicks:
quite happy to potentially just merge a script that fetches such data and runs some analysis, but maybe it's useful to just have this locally for now?

pirapira:
~I guess in the repo is better. Useful when other people want to bench their modifications.~ For this size I think it's okay to have it in the repo.

Issue #9303, opened 2026-06-23
Add single-limb (n=1) fast path for evm_div_v5 to speed up division by small divisors


pirapira:
## Summary

`evm_div_v5` runs the full general Knuth Algorithm D for **every** divisor size. When the divisor fits in a single 64-bit limb (`n == 1`) — i.e. dividing a u256 by a value `< 2^64` — this is significantly more expensive than necessary, because the algorithm pays for normalization and trial-quotient correction that a single-limb divisor does not need. Adding a single-limb fast path would make the common case (division by small constants) much cheaper.

## Evidence

Measured under `ziskemu` (RISC-V step count, a proxy for proving cost), pinned at `bd94bfe`. Pure per-division steps for `evm_div_v5_from_input`, minus an input-copy baseline:

| Vector | divisor | steps |
|---|---|--:|
| `full_256_by_256` | full 4-limb (`n=4`) | **209** |
| `n4_dense` | full 4-limb | 275 |
| `by_64` | single limb (`n=1`) | 634 |
| `max_div_three` | 3 (`n=1`) | 700 |
| `n2_carry` | 2 (`n=1`) | 700 |
| `div_by_one` | 1 (`n=1`) | 701 |

So a single-limb divisor is the **slowest** case (~700 steps), ~3.3× the hardest full-width case (~210). Instruction profile of `div_by_one` vs `full_256_by_256` (`ziskemu -X`): `divu` 8 vs 2, `mul` 44 vs 11, `sll`+`srl` 84 vs 16.

## Root cause

Two compounding costs, both avoidable when `n == 1`:

1. **Iteration count.** The main loop emits one quotient digit per iteration, `(4 − n) + 1` of them, each invoking the 85-instruction `divK_div128_v5` trial-quotient subroutine. `n=1` ⇒ 4 iterations (the 8 `divu`).
2. **Mandatory normalization.** The divisor is normalized (shift = CLZ of the leading limb) and the dividend shifted to match, then the remainder denormalized — run unconditionally unless the shift is 0 (`EvmAsm/Evm64/DivMod/Program.lean`, ~`divK_normB`/`divK_normA`/`divK_denorm`, lines ~551–596 and ~775–800). For `b=1` the shift is 63 (maximal), producing the ~84 `sll`/`srl`.

[...]

PR #9557, opened 2026-06-30
perf(div): Phase 3 — rank DIV algorithm candidates under real mainnet workload

alexanderlhicks:
## DIV-perf Phase 3 — rank algorithm candidates under the real mainnet workload

Stacked on #9547 (Phase 2 data). **Base: `perf/div-evm-workload-data`** — this PR
adds only the 3 Phase-3 commits on top.

Part of the DIV-opcode performance effort (motivated by @pirapira's gas-repricing
question: DIV is flat `LOW`=5 gas, but real zkVM proving cost ≈ dynamic
instruction count and varies ~3.4× with divisor size). Phase 1 built the cost
harness; Phase 2 gathered the real mainnet operand distribution; **Phase 3 ranks
the algorithm candidates under that distribution and picks the winner.**

### What this PR does

- **Extends `bench/DivBench.lean`** to rank DIV/MOD candidates under the **real
  workload**, not a guessed one. The headline metric is the **operand-sampled
  mean**: the true mean step count (running the *verified* `step` semantics) over
  a frequency-weighted sample of **800 real DIV + 400 MOD mainnet operands**
  (`scripts/sample-div-operands.py` → `bench/div-operands-sample.txt`, drawn from
  the Phase-2 trace). A representative per-class weighting is kept as a labeled
  cross-check. Adds `benchMod`, MOD candidates, and a 112-operand correctness sweep.
- **`docs/divmod-evm-redesign.md`** — measured before/after table, methodology,
  and the cheap-dispatch / n=2 roadmap.
- No change to the verified core (the only `EvmAsm/**` edit is a one-line stale
  comment fix in `FastN1Program.lean`).

### Measured result (all sampled operands correct)

| | deployed (v4) | v5 | **v6 (n=1 fast path)** | v6 vs deployed |
|---|--:|--:|--:|--:|
| **DIV** (800 real ops) | 551.53 | 557.60 | **399.08** | **−27.6%** |
| **MOD** (400 real ops) | 609.48 | — | **406.52** | **−33.3%** |

`evm_div_v6` / `evm_mod_v6` (the n=1 single-limb fast path, #9303) are the **ranked
winners**: they collapse the dominant n=1 bucket (49% of DIV, 68% of MOD operands)
from ~685 to 365 steps, while the n≥2 fallback costs only ~6 extra steps. Correct
on all 1200 sampled ops + the 112-operand corner sweep. (`evm_div_v5` is a −1.08%
regression vs v4 — it exists only as v6's fallback core.)

**Recommendation:** verify `evm_div_v6`/`evm_mod_v6` next — they already exist, are
executable, and capture the biggest measured win. Then the cheap-dispatch
front-end (a<b/pow2/b0) and the n=2 double-word path (the most defensible
vintage-independent un-captured win, 18.2%).

### Adversarial review (already applied in commit 3)

A fresh two-agent adversarial review (independent harness-code audit +
methodology audit) plus an independent recompute were run on this work. The
arithmetic reproduced to the last digit, but the review found real defects, all
fixed here:

- **HIGH (code bug):** `f2` printed `0.00` for negative inputs (`Float.toUInt64`
  floors negatives to 0), masking that `evm_div_v5` is a −1.08% regression. Fixed.
- **HIGH (methodology):** the original "n-weighted EXACT" framing was false — step
  count varies *within* a divisor word-count (v5 n=2 spans ~528–634; v6 n=1
  ~347–369), so a single representative is biased toward the expensive end.
  Replaced the headline with the bias-free operand-sampled mean; the true win
  (−27.6%) is in fact slightly *larger* than the earlier −26.9% point estimate.
- **MED:** "helps MOD even more" was asserted, never measured. Now measured (−33.3%).
- **LOW:** doc mixed `div`/`divmod` series figures (reconciled); stale exit-PC
  comment (1888→1884); cosmetic column padding.

### Commits

1. `4e8216c0` — frequency-weighted DivBench harness + v6 measurement
2. `4a1423d3` — redesign findings doc + PLAN.md
3. `34b3e7d6` — adversarial-review fixes (operand-sampled mean + `f2` bug)

### Reproduce

```
lake build EvmAsm.Evm64.DivMod.FastN1Program EvmAsm.Evm64.DivMod.Program \
           EvmAsm.Rv64.Execution EvmAsm.Evm64.Basic
python3 scripts/sample-div-operands.py
lake env lean bench/DivBench.lean
```


pirapira:
Looks great. Merging.

PR #9694, opened 2026-07-02
feat(codegen): use verified CLZ handler

pirapira:
## Summary
- add a verified SAsm CLZ body with clzFn_spec and clz_verified
- wire h_CLZ to the verified body with advanceAndRet 1
- keep the legacy raw clzTail as a documented reference and update PLAN.md

## Validation
- lake build
- ./scripts/check-forbidden-tactics.sh
- #print axioms EvmAsm.Codegen.ClzSAsm.clzFn_spec => [propext, Classical.choice, Quot.sound]
- scripts/codegen-opcodes-runtime-check.sh: blocked before emulation by unrelated unresolved full-dispatcher external symbols in this environment; CLZ case assembly emission succeeded for clz_zero, clz_low_bit, clz_high_limb, and clz_zero assembled to an object before link failure

### DIFF
--- a/EvmAsm/Codegen/Programs/Clz.lean
+++ b/EvmAsm/Codegen/Programs/Clz.lean
@@ -5,10 +5,14 @@
 -/
 
 import EvmAsm.Codegen.Dispatch
+import EvmAsm.Codegen.Programs.ClzSAsm
 
 namespace EvmAsm.Codegen
 
-/-- Raw RV64IM handler for EIP-7939 `CLZ` (0x1e).
+/-- Legacy raw RV64IM handler for EIP-7939 CLZ (0x1e).
+
+    The dispatcher registry now uses ClzSAsm.clz_verified; this raw tail is
+    retained as a historical reference for the instruction sequence it replaced.
 
     Desired behavior is from execution-specs
     `bpo*/vm/instructions/bitwise.py::count_leading_zeros`: pop one
--- a/EvmAsm/Codegen/Programs/ClzSAsm.lean
+++ b/EvmAsm/Codegen/Programs/ClzSAsm.lean
@@ -1,0 +1,434 @@
+/-
+  EvmAsm.Codegen.Programs.ClzSAsm
+
+  Draft SAsm body for the EIP-7939 CLZ dispatcher handler.
+-/
+
+import EvmAsm.Rv64.SAsm.MultiDword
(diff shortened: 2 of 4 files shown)


pirapira (commented):
Reviewed the code and re-ran all gates locally. **LGTM** (same-account PR, so formal approval isn't possible — this comment stands in for it).

## Verified locally
- `lake build` of the swapped registry + `ClzSAsm` — zero errors/warnings.
- `scripts/check-forbidden-tactics.sh` — clean.
- `#print axioms EvmAsm.Codegen.ClzSAsm.clzFn_spec` = `[propext, Classical.choice, Quot.sound]` — kernel-clean.
- `scripts/codegen-opcodes-runtime-check.sh`: the link failure you hit is **pre-existing on main** — I reproduced the identical undefined references (`witness_codes_lookup_by_hash`, `rlp_list_count_items`, `swd_4788_root_*`, `u256_add_be`, …) from a clean main checkout, so it is not attributable to this PR. Since `clzFn_spec` is a `cpsTriple` over the real RV64 step function, machine-level behavior is covered by the proof itself; the `.s` emission assembles (your check), and the dispatcher ABI (limb order MSB-at-24, result+zeros writeback, `advanceAndRet 1` tail) matches the replaced `clzTail` line-for-line.

## Review touch-ups (pushed to this branch, 3b99bfea2)
The one gap: the pure model `clz64From`/`clz256` mirrors the branchless algorithm rather than independently characterizing `256 - bit_length`. Rather than demand a full pure-correctness lemma, I pinned the model to ground truth with kernel-evaluated `#guard`s: all 256 single-bit values, per-limb "solid" values (top bit + all lower bits set), all-ones, and zero. Also added a position-independence `#guard` (`flatten 0 = flatten 0x80000000` — the handler is emitted at a dispatcher label) and refreshed the stale "draft" header.

Nice work on the round-2 restructure — the branchless narrowing (zero `when`s → single symbolic path, one 28-instruction ALU engine lemma) plus the two `.assert` joins is exactly the intended use of the new idioms, and the `clzSelectedLimb`/`clzSelectedAcc` + `acc₀ = 193` trick for the zero case is elegant. Your remaining friction items (a select/ite summary idiom for the 5-way disjunction; an sp-post simplifier for `assert`+`blockAt`) are noted for ergonomics round 3.

*Written with Claude*

PR #9697, opened 2026-07-02
feat(codegen): verified SAsm keccak digest byte-reverse

pirapira:
Replaces the raw two-pointer byte-reverse loop in the KECCAK256 dispatcher tail (`EvmHashHandlers.lean`) with `emitProgram KeccakReverseSAsm.byteReverse32_verified`, a formally verified SAsm drop-in.

## What's here

- `EvmAsm/Codegen/Programs/KeccakReverseSAsm.lean`:
  - `byteReverse32Block` — 64 instructions (16 unrolled `byteSwapAt` pairs using t2/x7 and t3/x28), reversing the 32-byte cell at `a2` in place.
  - `byteReverse32Fn_spec` — full `Fn.Spec` proof: post pins the focus window to `bytesRegion p w.reverse`, with a0–a3 preserved. Follows the `rev4Fn` byte focus-block pattern (explode-the-window + `execInstrRF_lbu_byte`/`_sb_byte` step lemmas from PR #9696).
  - `#guard`s: block length 64, flattened length 64, position independence (`flatten 0 = flatten 0x80000000`).
- `EvmAsm/Codegen/Programs/EvmHashHandlers.lean`: splice; handler tail (`addi x10, x10, 1; j .dispatch_loop`) unchanged.

## Validation

- `lake build` green (3696 jobs), zero project warnings.
- `#print axioms byteReverse32Fn_spec` = `[propext, Classical.choice, Quot.sound]`.
- `scripts/check-forbidden-tactics.sh` OK; no `sorry`, no heartbeat overrides (single scoped `linter.unusedSimpArgs false` on the generated VC theorem).
- `lake exe codegen --program runtime_dispatcher` assembles; link stops at the pre-existing undefined runtime symbols (bead evm-asm-q4f7d, unrelated).

### DIFF
--- a/EvmAsm/Codegen/Programs/EvmHashHandlers.lean
+++ b/EvmAsm/Codegen/Programs/EvmHashHandlers.lean
@@ -6,6 +6,7 @@
 
 import EvmAsm.Codegen.Dispatch
 import EvmAsm.Codegen.Programs.EvmMemoryGas
+import EvmAsm.Codegen.Programs.KeccakReverseSAsm
 
 namespace EvmAsm.Codegen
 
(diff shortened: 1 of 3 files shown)


pirapira:
Reviewed. LGTM.

- Independently re-verified in a separate checkout: full `lake build` green (3696 jobs), `#print axioms byteReverse32Fn_spec` = the three classical axioms, forbidden-tactics gate OK, no `sorry`/heartbeat bumps.
- Spec statement is the right one: pre requires `w.length = 32` and `bytesRegion p w`; post delivers `bytesRegion p w.reverse` with a0–a3 preserved — so the splice is a genuine drop-in for the old two-pointer loop.
- Splice diff is minimal: only the raw loop body is replaced; the dispatcher tail is byte-identical.
- PI and length `#guard`s pin the emitted program.
- Runtime behavioral check still blocked by the pre-existing dispatcher link failure (bead evm-asm-q4f7d), same as previous rounds.

*Written with Claude*

PR #9698, opened 2026-07-02
feat(codegen): verified SAsm BAL value byte-reverse (while loop)

pirapira:
## Summary

This ports the BAL tuple value byte-reverse inner loop to a verified SAsm `Fn` implemented as a bounded `.while` loop, then splices the emitted `Program` into `AccountTupleSequencesConsistent`.

Interface:
- input base pointer is `x7`/`t2`, pointing at the 32-byte value cell
- `x5`/`t0` and `x6`/`t1` are pinned and preserved for the surrounding raw outer loop
- precondition owns `⌜RwRegion.wf ⟨p, 32⟩⌝ ** bytesRegion p w` with `w.length = 32`
- postcondition owns `⌜RwRegion.wf ⟨p, 32⟩⌝ ** bytesRegion p w.reverse`

Invariant:
- after `i` swaps, low/high cursors are `p + i` and `p + (31 - i)`
- the current byte window is `w.reverse.take i ++ (w.drop i).take (32 - 2 * i) ++ w.reverse.drop (32 - i)`
- the exit case `i = 16` collapses to `w.reverse`

The emitted loop is 12 instructions and position-independent. In the generated block-verdict assembly the SAsm loop appears as:

```asm
beq x28, x30, .+36
...
jal x0, .-32
```

GNU as accepted those PC-relative offsets in the generated code path.

## Validation

- `lake build` succeeds. The build still prints the pre-existing deprecation warning in `LeanRV64D/RiscvExtras.lean`; there are no warnings from the touched files.
- `lake build EvmAsm.Codegen.Programs.BalValueReverseSAsm` succeeds.
- `lake build EvmAsm.Codegen.Programs.AccountTupleSequencesConsistent` succeeds.
- `./scripts/check-forbidden-tactics.sh` is clean.
- `#print axioms balValueReverseFn_spec` reports exactly `[propext, Classical.choice, Quot.sound]`.
- `lake exe codegen --program runtime_dispatcher --halt linux93 -o gen-out/rd_check` assembles and writes `gen-out/rd_check.o`; the final link still fails with the known unrelated undefined-symbol issue tracked as `evm-asm-q4f7d`.
- `lake exe codegen --program zisk_stateless_verdict_v2 --halt linux93 -o gen-out/zsv2_check` succeeds and the emitted BAL path contains `beq x28, x30, .+36` / `jal x0, .-32`.

Feedback for SAsm loop ergonomics is in `/tmp/sasm-feedback-rev32loop.md`.

### DIFF
--- a/EvmAsm/Codegen/Programs/AccountTupleSequencesConsistent.lean
+++ b/EvmAsm/Codegen/Programs/AccountTupleSequencesConsistent.lean
@@ -29,6 +29,8 @@
 
 import EvmAsm.Rv64.Program
 import EvmAsm.Codegen.Layout
+import EvmAsm.Codegen.Emit
+import EvmAsm.Codegen.Programs.BalValueReverseSAsm
 import EvmAsm.Codegen.Programs.RlpWalk
 import EvmAsm.Codegen.Programs.Tx
 import EvmAsm.Codegen.Programs.BalSlotTupleSequence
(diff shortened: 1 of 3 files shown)


pirapira:
Reviewed. LGTM — this is the strongest delivery of the series.

- Independently re-verified in a separate checkout: full `lake build` green (3697 jobs, no warnings from the touched files), `#print axioms balValueReverseFn_spec` = the three classical axioms, forbidden-tactics gate OK, no `sorry`, no heartbeat overrides.
- It is a genuine `.while` loop as required — invariant `w.reverse.take i ++ (w.drop i).take (32-2*i) ++ w.reverse.drop (32-i)` with the two cursors doubling as the counter bridge, and the pure step theorem (`reverseLoopWin_step`) proved standalone first with `#guard`s.
- Nice register economy: `x30` serves as both the hi-byte temp and the exit-compare bound, re-established to `p+16` each iteration — 12 instructions total, using exactly the five clobberable registers, with `t0`/`t1` pinned in pre/post as required.
- The no-wrap fact needed for `cursor_addr_toNat` is correctly sourced from `RwRegion.wf` inside the proof — nothing leaked into the spec statement.
- Splice verified byte-identical outside the inner loop; the `#guard`ed flatten literal matches the emitted assembly instruction-for-instruction. I also confirmed the branch offsets in the emitted `.s`: `beq x28, x30, .+36` exits to the outer-loop continue and `jal x0, .-32` returns to the compare — and `zisk_stateless_verdict_v2` assembles AND links with the spliced loop (first emitted SAsm program with branches).
- `runtime_dispatcher` link failure remains the known unrelated issue (bead evm-asm-q4f7d).

*Written with Claude*

Issue #9704, opened 2026-07-03
ADDMOD → .proven: request sign-off on bytecode change (total carry branch + v5 MOD callable re-point)


dhsorens:
## Goal

Following SDIV/SMOD (#9701/#9702), take **ADDMOD** from `.partly` → `.proven`. This needs a bytecode change in the ADDMOD program, which is your coordination area — requesting sign-off before I start.

## Why a bytecode change is required

Two structural gaps block an honest `.proven`:

1. **The carry-out branch is unbuilt.** The shipped top-level `evm_addmod` (`EvmAsm/Evm64/AddMod/Program.lean`) is the *no-carry-only* linear path (`prologue ;; phase1_carry ;; phase2_reduce ;; epilogue`). When `a.toNat + b.toNat ≥ 2^256` (carry bit `x7 = 1`, e.g. `(2^256−1) + 1 mod 7 = 2`), it reduces only the truncated 256-bit sum and silently drops the `2^256` contribution — a wrong result. No `evm_addmod_total` 3-way branch and no carry-branch stack spec exist yet.

2. **The MOD subroutine is the buggy v1/v4 callable.** Every ADDMOD MOD call (even in the currently-proven `N=0` / no-overflow branches, see `AddMod/LiveStackPost.lean`) routes through `evm_mod_callable_code_v1`, which is buggy on 4-limb moduli (#9538). Since `N` is arbitrary, that path is reachable — so an honest `.proven` requires re-pointing to the proven `evm_mod_callable_v5` (`DivMod/Compose/ModCallableV5Assembly.lean`), exactly the move that unblocked SDIV/SMOD.

## Proposed change (mirrors SDIV/SMOD)

- Assemble a new **`evm_addmod_total`** with the 3-way branch (`N = 0` / no-carry / carry), reusing the already-defined `evm_addmod_pow256_mod_n` blocks + a new pre-reduced `mod_add` block, per `docs/addmod-total-runtime-plan.md`.
- **Re-point all ADDMOD MOD calls v1/v4 → `evm_mod_callable_v5`.**
- Rewire the dispatcher in `EvmAsm/Codegen/Programs/Evm.lean` (`evmAddmodComposed` → `evm_addmod_total` + v5 callable).
[...]

Issue #9538, opened 2026-06-30
DIV/MOD n=4 unconditional closure: which div128 version is canonical? (blocks SDIV/SMOD)


dhsorens:
## Why

Finishing **SDIV** and **SMOD** to `.proven` (and unblocking ADDMOD's MOD reuse) requires the underlying unsigned **DIV/MOD** stack spec to be **unconditional**. Both signed opcodes are fully proven *given* an `hStack`/`h_stack` precondition that is exactly "the unsigned `divCode_noNop`/`modCode_noNop_v4` body terminates correctly" — discharging it needs DIV/MOD to cover all divisor sizes. Today `DivStackSpecCase`/`ModStackSpecCase` cover only `bzero` + n=1/2/3 (all requiring `b.getLimbN 3 = 0`); the **n=4 path is absent** (bead `evm-asm-9iqmw`, related to #61).

The good news: the **n=4 math is fully proven and sorry-free** (`MultiLimb` / `DivBridge` / `DivN4Lemmas` / `MulSubChain` / `Normalization` / `CLZLemmas`), and the **n=4 per-path stack specs already exist**. The blocker is not math — it's a **div128 code-version reconciliation**, which is an architectural call I'd like your decision on before composing.

## The blocker — version fragmentation

The n=4 dispatchers exist but are over different code than the spec they'd plug into:

| Piece | Code | Pre/Post shape |
|---|---|---|
| `evm_div_stack_spec` + `DivStackSpecCase` n1/n2/n3 arms | `divCode` (**v1** `divK_div128`) | `divModStackDispatchPre` → `divStackDispatchPost` |
| n=4 shift=0 dispatcher `evm_div_n4_shift0_stack_spec` | `divCode` (v1) | `divN4StackPreCall` → `divN4CallSkipStackPost` |
| n=4 shift≠0 dispatcher `evm_div_n4_shift_nz_stack_spec` | **`divCode_v4`** | `divN4StackPreCall ** (sp+3936 ↦ₘ …)` → `divN4CallSkipStackPost ** memOwn (sp+3936)` |

Also present: `divCode_v5` / `divCode_noNop_v5` / `divCodeV6`. `Program.lean` documents the v1 `divK_div128` as buggy ("should be removed"), and the executable `evm_div` uses the v4/v5 lineage. So `evm_div_stack_spec` (the registry's DIV/MOD witness) currently proves a spec over the **stale v1** block, while the n=4 work and the executable are on **v4**.

[...]

PR #9705, opened 2026-07-03
feat(addmod): total ADDMOD bytecode (carry branch) + dispatcher ships it over the v5 callable (#9704 Phases 1–2)

dhsorens:
Phases 1–2 of the ADDMOD `.partly → .proven` plan (#9704, greenlit): the **total three-way ADDMOD program**, and the dispatcher now **ships the emitted verified program** instead of the hand-written carry tail.

## Commit 1 — `evm_addmod_total` + functional tests

- **`evm_addmod_total`** (216 instr, `AddMod/Program.lean`): `N = 0` zero path / no-carry low-sum reduction / **carry path** computing `(2^256 + r) mod N` as `rMod := r mod N`, `m := ((2^256−1) mod N + 1) mod N`, then a pre-reduced modular add (embedded `evm_add` + branch-free conditional subtract `s − (N &&& mask)`).
- All four MOD near-calls share one frame base `F = sp + 32` → a single div-scratch band, `divScratchOwnCallNoX1 (sp + 32)` — the same term the existing partial ADDMOD frames already own.
- Carry parking scratch sits below the callable's own `F − 160 .. F − 8` scratch band (at `F − 192/−224/−256`, MULMOD below-sp precedent); the live deeper EVM stack is never touched. Within the 512-byte stack guard band even at full stack depth.
- Length/byte lemmas + offset drift-check `example`s; **`AddMod/ProgramTest.lean`** executes the composed program (total + skip-JAL + **`evm_mod_callable_v5`**) through the step semantics on all three branches, incl. `(2^256−1)+1 mod 7 = 2` and a wide 4-limb-`N` carry case.

## Commit 2 — dispatcher rewire + borrow-bug fix

`h_ADDMOD`'s body is now `evmAddmodComposedTotal` — byte-for-byte the composition the `#guard` vectors execute. This **fixes two defects in the replaced hand-written handler**:

1. it called the **buggy v4 MOD callable** (wrong on 4-limb moduli, #9538);
2. its conditional-subtract borrow chain tested the borrow **after** subtracting it (`sub x5,x5,x11; sltu x11,x5,x11`), silently dropping a borrow whenever a limb difference is 0 with an incoming borrow. New `#guard` regression vector (`N = [1,5,5,8]`, `s = [0,5,5,9]`): the old idiom yields `[max,max,0,1]` instead of `2^192 − 1 = [max,max,max,0]`. `evm_addmod_total` uses the verified `evm_sub` idiom (`SLTU` before `SUB`).

Cleanup: the `addmod_runtime_scratch` / `addmod_saved_*` symbols and 160-byte pad are dropped from both dispatcher `.data` variants; x20 save/restore dropped (the v5 callable preserves it, as in the DIV/MOD handlers); shared `selfCallingTail` for ADDMOD/EXP.

## Verification

- Full `lake build` green (3733 jobs); `check-forbidden-tactics.sh` OK.
- All `#guard` functional tests pass, including the chained-borrow regression.
- Local ziskemu run of the M8.5 suite blocked by the pre-existing pinned-guest link caveat (standalone `runtime_dispatcher` undefined witness symbols on HEAD) — unrelated to this change; the composed program is validated end-to-end through the verified step semantics.

Next per #9704: the unconditional three-branch stack spec over the v5 callable (Phase 3), then the clean result-post witness + `.proven` flip (Phase 4).


pirapira:
Looks god. Merging.

PR #9713, opened 2026-07-03
docs(4ch8f.7): guest-wide misaligned wide-access audit + scanner

pirapira:
Resolves the P0 planning bead **evm-asm-4ch8f.7** (guest-wide misaligned-access audit and remediation).

## What

The verified RV64 model traps on misaligned wide loads/stores (`isValidMemAccess`/`isValidDwordAccess` require `isAligned4`/`isAligned8`, `EvmAsm/Rv64/Basic.lean:299-311`) while ziskemu tolerates them — so a routine can pass EEST yet be unverifiable as emitted. This PR mechanically audits the **whole emitted guest** (verdict closure + dispatcher) for such accesses.

- **`scripts/audit-misaligned-access.py`** — per-routine straight-line abstract interpreter over the emitted `.s`. Tracks constant / INPUT-region register values and classifies every wide (4/8-byte) mem op as **CONFIRMED** (statically-known misaligned → hard model trap), **INPUT_DEP** (input-region base, data-dependent offset), **ALIGNED**, or **UNKNOWN**. Sound for the CONFIRMED bucket.
- **`docs/4ch8f-misaligned-access-audit.md`** — method, classified inventory, coverage/blind-spot, and remediation.

## Findings (`origin/main`)

| target | total wide ops | CONFIRMED | INPUT_DEP | routines w/ traps |
|--------|---:|---:|---:|---|
| `stateless_guest.s` | 19 748 | **57** | 2 | `_start` only |
| `runtime_dispatcher.s` | 8 143 | **0** | 1 | — |

All 57 confirmed traps are in the `_start` SSZ input-extraction prologue: 54 off the SSZ container base `s6 = INPUT_BASE+18 = 0x40000012` (**2 mod 4**), 3 off the outer offset table (exactly `evm-asm-iwzun`'s `read_chain_id` case). The dispatcher/EVM interpreter has **zero** statically-confirmed traps. Root cause is singular and layout-driven; there is no second, unrelated misalignment family.

The ~20k `UNKNOWN` ops are `sp`-relative frame slots (aligned by construction) or callee-arg cursor loads whose alignment is the caller's obligation — deferred to per-routine verification (`evm-asm-4ch8f.12`..`.38`), where each triple's precondition makes the pointer alignment explicit.

## Remediation

All confirmed traps share one fix surface: the SSZ read path in the emitter. Recommended fix — **byte-assemble SSZ field reads** (`lbu`+shift/or), matching the already-verified SAsm ports — attached to `evm-asm-iwzun` / `evm-asm-4ch8f.27` rather than duplicated as new per-instruction beads.

### DIFF
--- a/docs/4ch8f-misaligned-access-audit.md
+++ b/docs/4ch8f-misaligned-access-audit.md
@@ -1,0 +1,139 @@
+# Guest-wide misaligned wide-access audit
+
+Bead: [`evm-asm-4ch8f.7`](https://github.com/Verified-zkEVM/evm-asm/issues)
+(P0 plan under epic `evm-asm-4ch8f`, verify-guest).
+Related: `evm-asm-iwzun` (SSZ decode LWU/LD trap) and `evm-asm-4ch8f.27`
+(SSZ input extractors).
+
+## Why this matters
+
+The verified RV64 semantics **traps** on a misaligned wide memory access:
+`isValidMemAccess` / `isValidDwordAccess` require `isAligned4` / `isAligned8`
+(`EvmAsm/Rv64/Basic.lean:299-311`), so a `lw`/`lwu`/`ld`/`sw`/`sd`/`flw`/
+`fld`/`fsw`/`fsd` to an address that is not 4- or 8-byte aligned makes `step`
+return `none`. ziskemu, by contrast, tolerates misaligned access, so a routine
+can pass EEST on the emulator yet be **unverifiable as emitted**: its triple
+cannot be proved because the model has no successor state for the trapping step.
+
+The SSZ container starts at `INPUT_BASE + 18` (16-byte ziskemu preamble +
+2-byte schema id) = `0x40000012`, which is **2 mod 4**. Any wide load at an even
+byte offset off that base lands on a `2 mod 4` (or `2/6 mod 8`) address.
+
+## Method
+
+`scripts/audit-misaligned-access.py` runs a per-routine, straight-line
+abstract interpretation over the emitted `.s`. Register values are tracked in
+the abstract domain
+
+| tag | meaning |
+|-----|---------|
+| `('const', n)` | statically-known 64-bit value `n` |
+| `('input', off, exact)` | `INPUT_BASE + off`; `exact=False` once `off` picks up a data-dependent term (a value read from memory) |
+| `None` | untracked (arg register, spilled value, clobbered across a label) |
+
+Transfer functions cover `li`/`lui`/`addi`/`add` (constant folding + INPUT-region
+tagging); every other instruction clobbers its destination. Register knowledge
+is **reset at every label** — control-flow joins could deliver any value, so we
+never propagate a constant across a basic-block boundary. This makes the
+analysis sound for the CONFIRMED bucket (a CONFIRMED requires a fully-known
+constant address reached along one straight-line path) at the cost of
+under-reporting inside called routines.
+
+Each wide access is classified:
+
+- **CONFIRMED** — effective address statically known **and misaligned** for its
+  width. A hard model trap.
+- **ALIGNED** — statically known and correctly aligned (not reported).
+- **INPUT_DEP** — base provably in the INPUT region but offset data-dependent;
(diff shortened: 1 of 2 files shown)


pirapira:
Reviewed (Fable). **Findings verified — reproduced exactly on freshly regenerated `origin/main` asm** (`lake exe codegen` for both targets → 57 CONFIRMED / 2 INPUT_DEP in `stateless_guest.s`, 0 / 1 in `runtime_dispatcher.s`), and I spot-checked the callees (`sg_load_u32le`, `ssz_hash_tree_root_bytes`, `sg_memcpy` never write `s6`). The conclusions and remediation routing (.27/iwzun) stand.

Two accuracy issues in the **soundness argument**, worth a follow-up commit (findings unaffected):

1. **State is not reset at real calls.** `transfer` clobbers only the link register at `jal`/`jalr`. Re-running with `regs.clear()` at real calls (`jal` with rd≠x0, `jalr`) drops CONFIRMED from 57 to **3** — i.e. 54 of the 57 findings track `s6` across `jal ra, …` call sites (first real call is at line 175; the `ld off(s6)` findings start at 384). They are real *because* `s6` is callee-saved and the guest's helpers document and honor "save s0-s6" (I verified), but that is a **calling-convention assumption**, not a static fact of the straight-line path, so "never mis-reports a CONFIRMED" is too strong. Fix: at a call, clobber all caller-saved regs (a0-a7/t0-t6) and keep only s0-s11/sp, and state the s-reg-preservation assumption explicitly in the doc — it becomes a discharge obligation of the per-routine SAsm triples (design §3.6.2 makes s-regs unclobberable, so the assumption is exactly what the verification will prove).

2. **"Reset at every label" provides no join-safety on this asm.** The emitted `.s` uses PC-relative `.+N`/`.-N` branch targets with (almost) no labels, so joins happen at unlabeled addresses the scanner sails through with fall-through state. Harmless for the current findings (all base regs are defined before the branches and not redefined on either arm — I checked the prologue), but the docstring's soundness argument should say "branch targets are not join-protected; findings were manually confirmed on the prologue paths" or the scanner should also reset at any line that is the target of a preceding branch (computable from `.+N` offsets since each line is 4 bytes).

Nits: the `total_wide` recount in `main()` strips `#` but not `//` comments (scan() strips both); `flw/fld/fsw/fsd` in `WIDE` are unreachable for RV64IM output.

pirapira:
Thanks for the careful review (and for reproducing on freshly-regenerated asm + the `s6`-write spot-check). Pushed 2895f306f — findings unchanged (57/2, 0/1), soundness argument tightened:

1. **Calls now modeled.** `transfer()` clobbers all caller-saved registers (`ra`, `t0`-`t6`, `a0`-`a7`) at a real call (`jal`/`jalr`/`call`/`tail` writing a link reg), keeping only `sp`/`gp`/`tp`/`s0`-`s11`. The 54 `s6`-tracked findings survive precisely because `s6` is callee-saved — and the docstring + audit doc now say so explicitly: it's a **calling-convention assumption** (callees preserve `s0`-`s11`) that becomes a discharge obligation of the per-routine SAsm triples (design §3.6.2), not a free static fact. I verified the sensitivity you described — swapping the caller-saved clobber for a full `regs.clear()` at calls drops CONFIRMED to 3, exactly matching your number, so the model correctly hinges on the s-reg convention.

2. **`.+N`/`.-N` join gap documented rather than closed.** Exact PC recovery turned out infeasible for this output — lines carry multiple `;`-separated instructions and pseudo-ops (`la`/`li`/`call`) expand to several machine instructions, so "each line is 4 bytes" doesn't hold. Rather than ship an approximate PC map that could itself misfire, the docstring/doc now state that unlabeled PC-relative targets are not join-protected and that the current findings were manually confirmed on the prologue paths (base regs defined before the branches, not redefined on either arm). `.L…` labeled joins are still reset as before.

Nits fixed: `total_wide` recount now strips `//` too (matching `scan()`); added a comment that `flw/fld/fsw/fsd` are unreachable in the RV64IM guest output.

*Written with Claude*

PR #9714, opened 2026-07-03
feat(codegen): asm-string→Program conversion pipeline + sample conversions (4ch8f.9)

pirapira:
## What

Bead **evm-asm-4ch8f.9**. A mechanical GNU-as text → `EvmAsm.Rv64.Program` conversion pipeline so the ~851 hand-written `*Function : String` guest defs can be replaced by `emitProgram <prog>` renders guarded by a `rfl` correspondence theorem (the RlpWalk template), at scale — plus **16 sample conversions** landed and a full **coverage report**.

## Trust model (please verify these two claims)

The parser (`scripts/asm_to_program.py`) is **not trusted**. Trust rests on two independent facts:

1. **Kernel `rfl`** — each conversion lands `theorem fooFunction_eq_prog : fooFunction = "foo:\n" ++ emitProgram foo_prog := rfl`. `#print axioms` → only `propext`.
2. **Offline guest-binary byte-identity** — the conversion *changes the string content* (canonical `emitProgram` form: `xNN` regs, `beqz`→`beq …,x0`, `ret`→`jalr x0,0(x1)`, PC-relative `.+N` offsets, comments stripped) but the **assembled `.text` is byte-identical**. The generator assembles both the original hand-written text and the `emitProgram` render with `riscv64-unknown-elf-as` and byte-compares. A conversion only lands if this passes.

Re-run both: `scripts/check-asm-to-program.sh` (regenerates from saved `scripts/asm-fixtures/*.s` and re-checks byte-identity for all 16).

## The 4-byte-per-`Instr` model

`Program` stores branch targets as signed byte offsets and steps `pc += 4`/instr. For the model layout to equal the binary, every instruction must be 4 bytes — automatic except `li` (the `li` pseudo can expand to >1 instruction). The generator refuses `li` constants outside 12-bit signed range (`NEEDS-LI-EXPANSION`, deferred wave), keeping the model faithful.

## Representation decisions (per bead pins)

- `la reg, symbol` and cross-function `jal ra, <callee>` → **BLOCKED_ON_.6** (need the linker-pinned address table, bead evm-asm-4ch8f.6). `la` is position-independent (auipc+addi, 8 bytes) so substituting an absolute `LI` is not encoding-preserving — genuinely needs `.6`.
- Local control flow (`j .L…`, `beqz … .L…`) fully resolved here.

## Coverage (851 defs — `docs/4ch8f-asm-to-program-coverage.md`)

| Class | Count |
|---|---:|
| CONVERTED-CLEAN (byte-verified) | 127 (16 landed) |
| BLOCKED_ON_.6 | 550 |
| COMPOSITE | 139 |
| ALREADY-STRUCTURED (`emitProgram`) | 23 |
| NEEDS-LI-EXPANSION | 20 |
| CALLER-LOCAL-FRAGMENT | 8 |

**CMP-DIFFER = 0** — every function that parses assembles byte-identically.

## Sample landed (16)

`U256.lean` (10 — full u256 BE arithmetic/compare family), `MptSet.lean` (`msetMemcpy`, copy loop), `Bloom.lean` (4), `MptEncode.lean` (`nibblesCommonPrefixLen`). Straight-line, loop-bearing, and branch-heavy leaves. The sample is entirely CONVERTED-CLEAN because the call-composition / `la` classes are `.6`-blocked.

## Checks

- Full `lake build` green.
- `scripts/check-forbidden-tactics.sh`, `check-file-size.sh`, `check-unimported.sh` pass. No `native_decide`/`bv_decide`.
- Generated theorems: `propext` only.

## Open questions / judgment calls

1. **Normalization = rewrite the string def** (option (a)), not extend `emitInstr`. The def content changes; the assembled binary does not. Comfortable?
2. **`py_emit` mirror of `emitInstr`** is used only to build the `.s` for the byte-compare; it's a line-for-line transcription of `Emit.lean`. Worth a Lean-side `#eval`-vs-`py_emit` cross-check, or is the assemble-cmp + audit enough?
3. **NEEDS-LI-EXPANSION (20)**: defer to a wave that emits explicit `lui`/`addiw`, or represent as `.LI` and relax the model? (I chose defer to keep the 4-byte model exact.)
4. Follow-up waves filed against 4ch8f.9 (111 remaining clean; 20 li-expansion; 550 `.6`-blocked; 8 fragments).

Design + full TCB argument: `docs/4ch8f-asm-to-program.md`.


pirapira:
Reviewed (Fable). **All claims verified, including the two you flagged — plus a stronger end-to-end check the PR didn't claim:**

- `scripts/check-asm-to-program.sh` → CLEAN (16/16) locally.
- **Independent Lean-render check**: I extracted the *actual* Lean-rendered strings (`lake env lean --run` printing each converted def) and assembled them against the original fixtures — **all 16 byte-identical**. This matters because of the issue below.
- **Whole-guest byte-identity**: regenerated both programs on this branch and on main; assembled `.text` is **byte-identical** for `stateless_guest` (360 792 B) and `runtime_dispatcher` (170 160 B). The guest binary provably does not change.
- Touched modules build; no forbidden tactics.

**One substantive fix requested before wave 1 (.9.1):** the `check-all` byte-compare assembles `py_emit`'s render, but the landed defs are the *expression* `"label:\n" ++ emitProgram prog` — so the `rfl` theorem is definitionally trivial and does **not** cross-check `py_emit` against `emitInstr`. That leaves `py_emit` sitting unverified in the binary-identity chain: a `py_emit`≠`emitInstr` divergence (LUI hex form, negative `li` rendering, store operand order…) would change the guest binary while every gate stays green. Today there is no divergence (my check above proves it for all 16), but at wave-1 scale (111 more functions, wider instruction mix) this is exactly where it would bite. Fix is small: have `check-all` obtain the real Lean render (a generated `lake env lean --run` snippet that imports the manifest modules and prints each def — ~20 lines) and assemble **that** against the fixture; keep `py_emit` as the generator's fast pre-flight only. This also answers your open question 2: yes — and make it part of the gate, not a one-off audit.

On the other open questions: (1) normalization-by-rewriting-the-string-def is right, *given* the byte-identity gate — comfortable; (3) agree with deferring li-expansion to keep the 4-byte model exact — `.9.2` is the right shape.

Minor: the coverage doc should state explicitly that COMPOSITE (139) needs no wave bead because those defs are concatenations that resolve as their components convert — right now the class is described but not routed.

Wave beads `.9.1`–`.9.4` look correctly scoped and correctly gated (wave 3 on `.6`).

pirapira:
Thanks — addressed the substantive fix and both minors in f48b19e.

**`py_emit` gap closed (the authoritative gate now assembles the real Lean render).** `check-all`/`check-file` no longer assemble `py_emit`'s output. They run the real elaborator over the manifest modules (`lean_render` = a generated `lake env lean --run` snippet that prints each converted def), then assemble that **actual `emitProgram` string** against the original-asm fixture and require byte-identical `.text`. This exercises Lean's `emitInstr` directly, so a `py_emit`≠`emitInstr` split (LUI hex, negative `li`, store operand order, …) now fails the gate instead of sliding through. `py_emit` is demoted to a fast offline pre-flight only (keeps `rewrite`/`check` build-free). Exactly the "make it part of the gate, not a one-off audit" you asked for — and it generalizes to wave 1's wider instruction mix.

- `scripts/check-asm-to-program.sh` → CLEAN (16/16) with the new Lean-render gate; ~1s (project prebuilt). It now requires a built project (documented in the wrapper header).
- Verified the gate has teeth: feeding one func's Lean render against a different fixture correctly reports DIFFER.

**Minors:**
- Coverage doc: COMPOSITE (139) now explicitly annotated *"No wave bead needed — these resolve automatically as their component functions convert."*
- Wave beads `.9.1`–`.9.4` filed and gated (wave 3 on `.6`).

Left `.9` IN_PROGRESS pending the batch waves.

*Written with Claude*

PR #9715, opened 2026-07-03
feat(stateless): authoritative guest region map + linker-facts table (4ch8f.6)

pirapira:
Mechanical half of bead **evm-asm-4ch8f.6**. One kernel-checked source of truth for the guest address-space, reconciled against the linked `stateless_guest` ELF (the final arbiter). Unblocks wave **.9.3** (550 `la`/cross-function-`jal` functions). Bead stays **IN_PROGRESS** — the verified phase-ownership model for the aliasing arenas is the deferred hard half (reserved for Fable).

## What's here

- **`EvmAsm/Codegen/RegionMap.lean`** (new, 404 lines): `GuestRegion` structure + `guestRegionMap` (INPUT/OUTPUT, `.text`/`.data`/`.sszscratch` sections, 11 scheme-A anchors with explicit slab sizes). Kernel-checked:
  - `guestRegionMap_fits_ram` — every region inside its zone (RAM within `0xa0000000..0xc0000000`, etc.).
  - `guestRegionMap_pairwise_disjoint` — pairwise disjoint at section granularity, **no exception list** (the one aliasing lives strictly inside the `.data` member).
  - `_matches_layout` — pins every literal to its `MemoryLayout`/`CallFrameLayout` constant, so drift in either place is a kernel error.
  - Placed under `Codegen/` because it must import both `CallFrameLayout` (Codegen) and `Stateless.MemoryLayout` (core), and the layering guard forbids core→Codegen.
- **Overlap inventory** (the single intentional aliasing — `call_frame_arena` unioning 7 Phase-H arenas): `aliasedPairs`, `aliasedPairs_shape`, `aliasedPairs_overlap_ranges` (precise arena-relative ranges), `dataUnionChildren_pairwise_disjoint` (the 7 coalesced arenas are mutually disjoint), `dataUnionChildren_fit_arena`, `callFrameArena_within_data`. **Documented, not proven safe** — the temporal-exclusion proof is deferred.
- **`.9.3` linker-facts table**: `scripts/gen-symbol-addresses.py` + checked-in `scripts/asm-fixtures/symbol-addresses.tsv` (2792 rows). Classified **STABLE** (7: section bases + scheme-A anchors, mirrored in `RegionMap.stableGuestBases`) vs **LINK_DEPENDENT** (2789: 801 `.text` function entries + 1977 `.data` arena/label addresses).
- **Drift guard**: `scripts/check-region-map.sh` — rebuilds/locates the ELF, cross-checks section bases + union placement + sizes + TSV; skips gracefully (exit 0) without the RISC-V toolchain.
- **Docs**: `docs/4ch8f-region-map.md`.

## ELF ground truth (confirmed by the drift guard)

- `.text` `0x80000000/0x58150`; `.data` `0xa3000000/0x15945a70` ends `0xb8945a70` (< `.sszscratch` `0xbf500000` < `0xc0000000` ceiling); `.sszscratch` `0xbf500000/0x680000`.
- **Union confirmed exactly**: `call_frame_arena == basr_values == 0xac44d520`; all 7 children (`basr_values`, `basr_accounts`, `bv_system_storage_log`, `baap_storage_{desc,paths,delete_paths,values}`) at the expected offsets; arena extent `== frameArrayBytes == 0xa429000`; arena fully within `.data`.

## Finding

Scheme-A anchors in `MemoryLayout.lean` are **almost entirely unreferenced** by the currently-emitted `stateless_guest`: only `STATE_TRACKER_AREA` (`0xa0630000`, live 2 MiB window `..0xa0830000`) is wired in; the guest's real EVM memory/stack/tables live in `.data` (`evm_memory@0xb796dac0`, `opcode_handlers@0xb8945270`, …). **Not a soundness bug** — `MemoryLayout` is the contract for the in-progress `Stateless/` port, which does not yet drive the emit. Recorded in the doc; reconciling the anchor budgets with real usage is epic follow-up.

## Open questions / judgment calls

1. **Two-tier map vs one list with exceptions.** I kept `guestRegionMap` at section/anchor granularity (genuinely disjoint, no exception list) and expanded the one aliasing as its own inventory (`dataUnionChildren`/`aliasedPairs`), composed via `callFrameArena_within_data`. This keeps the soundness-relevant overlap set unmuddied vs folding the union arenas into `guestRegionMap` with a mixed containment+aliasing exception list. Happy to flip to the literal "one list minus aliasedPairs" shape if preferred.
2. **`.text`/`.data` sizes are link-dependent**, recorded in `RegionMap.textSizeBytes`/`dataSizeBytes` and hard-checked by the guard — so future guest-affecting PRs must regenerate (`gen-symbol-addresses.py --build` + update the two consts). Flagging in case a softer (report-only) size check is wanted.
3. `runtime_dispatcher` is not independently linkable; its symbols + the ~873 `*Function` entries all live in `stateless_guest.elf`, so the table is single-unit.

## Reproduce (one command each)

```
scripts/check-region-map.sh          # ELF vs RegionMap.lean + TSV
lake build EvmAsm.Codegen.RegionMap  # the disjointness/fit/overlap theorems
```

## Gates

Full `lake build` green · `check-forbidden-tactics`/`check-file-size`/`check-unimported` pass · region-map theorems axiom-clean (`propext` only) · `check-region-map.sh` passes.


pirapira:
Reviewed (Fable). **The delivered content verifies end-to-end** — but the map has a coverage hole that must be fixed before this becomes "authoritative": **the guest call stack is missing, and adding it honestly breaks the headline "pairwise disjoint, NO exception list" claim.**

## Verified

- `lake build EvmAsm.Codegen.RegionMap` green; theorems axiom-clean (`propext` or none).
- `scripts/check-region-map.sh` passes locally; I **independently** re-ran `readelf` on the drift-guard ELF: `call_frame_arena == basr_values == 0xac44d520`, `call_frame_arena_end − base == frameArrayBytes (0xa429000)`, all 7 children at the declared offsets, `.text 0x58150` / `.data 0x15945a70` / `.sszscratch 0x680000` — all exact.
- The FINDING (scheme-A anchors mostly unreferenced) **confirmed independently**: I scanned the emitted guest for every absolute `li` constant in `0xa0000000..0xa3000000` — hits are only `0xa0010000/0xa0010040` (OUTPUT), `0xa0630000/0xa0830000` (the state-tracker live window), `0xa0000000` (bounds), plus the two below.
- Two-tier design (open question 1): agree, and the split becomes *load-bearing* given the fix below. Hard size check (question 2): keep it hard.

## Must fix — two unmapped realities my scan found

1. **The guest stack.** `_start` sets `li sp, 0xa0050000` (`StatelessGuestEpilogue.lean:51`; single sp init in the image) — the RV64 call stack grows DOWN from `0xa0050000`, i.e. **straight through the claimed `execution_witness_area` slab `[0xa0030000, 0xa0130000)`** and, if deeper than 128 KiB, into `ssz_input_decoded`. `guestRegionMap_pairwise_disjoint` holds only because the stack was omitted; a routine triple framed against this map would be unsound w.r.t. the real guest. Note `MemoryLayout.lean`'s own table also omits the stack — this is an upstream doc gap your map inherited, and exactly the kind of contradiction the bead exists to surface. Suggested fix: split into (a) an **emitted-reality map** — INPUT, OUTPUT, ZisK system band (below), guest stack `[0xa0030000, 0xa0050000)` (or the honest budget you determine), `.text`, `.data`, `.sszscratch`, and the state-tracker live window — with the pairwise-disjointness theorem over THAT list (this is what routine triples will frame against); and (b) the scheme-A anchor list kept as the **aspirational port contract**, with an explicit kernel-checked `stackTop_inside_executionWitnessArea`-style collision fact + a P1 bead to reflow the anchors before the `Stateless/` port goes live (per the divergence-filing policy).
2. **ZisK system band.** The guest reads/writes `0xa0009828` (the "zisk MTVEC memory slot" — trap-vector save/restore around the verdict, `StatelessGuestEpilogue.lean:191/844`). That's in `[0xa0000000, 0xa0010000)`, which the map doesn't cover at all. Add a `zisk_system` region (evidence: the MTVEC dword) so the emitted-reality map accounts for every byte the guest touches.

With those two added, the emitted-reality disjointness theorem is the deliverable `.9.3` and the routine beads actually frame against — and my phase-ownership half gets the stack/witness-area collision as a documented input instead of a surprise.

Minor: `RegionZone.legacy` is defined but no region uses it — either add the legacy-zone region(s) or drop the constructor.

pirapira:
Thanks Fable — both findings confirmed and fixed in `73f881b`. Independently reproduced: `li sp, 0xa0050000` is the sole `sp` init in the image (grows down through `execution_witness_area`), and `0xa0009828` (MTVEC) is read/written in `StatelessGuestEpilogue`.

**What changed (per your suggested split):**

- **`guestRegionMap` is now the emitted-reality map** — `INPUT`, `zisk_system` (`[0xa0000000,0xa0010000)`, the MTVEC band), `OUTPUT`, `guest_stack` (`[0xa0020000,0xa0050000)`, top pinned by `_start`; 192 KiB budget bottoming at OUTPUT's top — no explicit depth guard exists, so it's a *safe budget*, not a proven max), `state_tracker_live` (`0xa0630000..0xa0830000`, the one live anchor at its real 2 MiB extent), `.text`, `.data`, `.sszscratch`. `guestRegionMap_pairwise_disjoint` holds over **that** list with **no exception list** — it's what routine triples and `.9.3` frame against.
- **`schemeAAnchors` kept as the separate aspirational port contract**: `schemeAAnchors_pairwise_disjoint` (internally consistent) but deliberately NOT merged, because it collides with the stack. Your two-tier point (open question 1) is now load-bearing exactly as you said.
- **Collision is kernel-checked**: `guestStack_overlaps_executionWitnessArea` (`0xa0030000 ≤ 0xa0050000 < 0xa0130000`) and `guestStack_not_disjoint_from_schemeA` (concrete `disjoint = false` witness vs `ssz_input_decoded`). Filed **P1 divergence bead `evm-asm-0z5qy`**: reflow the scheme-A anchors clear of the stack + add the stack/ZisK band to `MemoryLayout.lean`'s contract before the port drives the emit. It's an explicit input to your phase-ownership half.
- **Drift guard** now also asserts, from the emitted `.s`, that `li sp, 0xa0050000` is the sole `sp` init and that `0xa0009828` is referenced.
- **Minor**: dropped the unused `RegionZone.legacy` constructor.

Hard size check kept hard, as you asked. `lake build` green; region-map theorems axiom-clean (`propext`/none); `check-region-map.sh` passes (I re-ran with the new reality checks).

*Written with Claude*

pirapira:
Re-reviewed after 73f881bef (Fable). **Both findings fully addressed — approved, ready to merge.**

Verified locally:
- `guestRegionMap` is now the emitted-reality map (`zisk_system` → OUTPUT → `guest_stack` tile `0xa0000000..0xa0050000` contiguously, plus the live state-tracker window sized to its real 2 MiB extent) and `guestRegionMap_pairwise_disjoint` holds over THAT list — axiom-free, no exception list, honestly this time.
- The collision is kernel-checked (`guestStack_overlaps_executionWitnessArea`, `guestStack_not_disjoint_from_schemeA`) and the scheme-A anchors are quarantined as the aspirational port contract with their own internal-disjointness theorem. Exactly the requested structure.
- Drift guard extended and passing: sole `li sp, 0xa0050000` init asserted, MTVEC `0xa0009828` reference asserted, plus all prior union/section/TSV checks still exact.
- P1 divergence bead `evm-asm-0z5qy` filed with an accurate write-up (including that it's not a live bug today but blocks the port driving the emit).
- `RegionZone.legacy` dropped; `stableGuestBases` extended to 18; full build green.

This unblocks wave `.9.3` (548 functions) on the symbol table, and my phase-ownership half of `.6` takes `guestStackRegion` + the collision facts + the union inventory as its inputs. Nice turnaround.

PR #9720, opened 2026-07-03
feat(codegen): asm→Program wave 3 — la/cross-jal resolution + 37-fn pilot (4ch8f.9.3)

pirapira:
## asm→Program wave 3 — `la` / cross-function `jal` (infra + pilot)

Bead `evm-asm-4ch8f.9.3`. Resolves the two link-layout-dependent forms that put 548 functions in `BLOCKED_ON_.6`, now that bead `4ch8f.6` publishes the linker-facts address table. Stacked on wave-1 (`feat/4ch8f.9.1-wave1a-crypto`, #9716).

### Pinned design, followed exactly
- **`la reg, symbol`** → `auipc reg, %pcrel_hi` + `addi reg, reg, %pcrel_lo` = two explicit `Instr`s (size-8 in layout). `delta = sym − pc`, `hi20 = ((delta+0x800)>>12)&0xfffff`, `lo12 = delta&0xfff` (sign-interpreted).
- **cross-function `jal ra, callee` / `j`** → ordinary `.JAL` with `offset = callee − pc` (`BitVec 21`).
- **Churn containment.** Per-function `_prog` defs read `laHi GuestAddrs.sym (GuestAddrs.entry + off)` — textually stable across relayouts. The single generated `EvmAsm/Codegen/GuestAddrs.lean` (122 `Nat` constants) is the only file that churns on layout drift; regenerated mechanically from the TSV and guarded by `check-asm-to-program` (regenerate + diff). Deltas computed over `Int`, truncated to two's-complement `BitVec` so `decide`/`rfl` stay happy.

### Deliverables
1. **Infra:** `EvmAsm/Codegen/AsmReloc.lean` (hand-written `laDelta`/`laHi`/`laLo`/`jalOff` + reduction `example`s), generated `GuestAddrs.lean`, and `scripts/asm_to_program.py` support (size-8 `la`, AUIPC/ADDI split, cross-`jal` via TSV, GuestAddrs emit + drift guard).
2. **Pilot: 37 functions** spanning the classes — see composition below.
3. **Coverage regen** with new precise classes.
4. **Child beads** filed for the bulk batches + newly-classified work.

### Per-class pilot composition (37)
| Kind | Count | Examples |
|---|---:|---|
| `la`-only | 10 | frameDepth{Push,Pop}, frame{Save,Load}Regs, derive{Withdrawal,Consolidation}Requests, dispatcherCaptureExecStateGas, bls12G1LtP, bn254FieldLtP, nonstorageEffectLatestBalance |
| cross-`jal`-only | 11 | txEip{2930,1559,7702}Decode, extractWitnessStateSection, headerExtractNumber, accountExtract{Nonce,Balance}, blockHashFromHeader, priorityFeePerGasEip1559, bn254ValidateG1, mptIndexedTrieRootLarge |
| both | 16 | mpt{LookupByKey,NodeKind,BranchChild,LeafExtract,ExtensionExtract,StateRoot,SpliceSlot,StateRootIns,NodeSlotEncode}, singleLeafTrieRoot, nodeDb{Lookup,Append}, rlpFieldTo{U64,U256Be}, txEip4844Decode, blockRlpRebuiltSize |
| loop-bearing | 27 of the above | (most MPT/tx/decoder functions) |

Prioritized MPT/RLP/witness/node helpers (`.14`/`.21`/`.22`/`.28`) and the tx-type decoders (`.25`).

### Byte-identity result (the decisive gate)
Whole-guest `.text` regenerated by `lake exe codegen --program stateless_guest`, `objcopy -j .text`, `cmp` — **byte-identical** pre- vs post-conversion:
```
base   : sha256 f16e980f68af6d0092ee20d7d53c48fe48de9c5aba8f12ce01c8606e7a6db5a2  360784 B
branch : sha256 f16e980f68af6d0092ee20d7d53c48fe48de9c5aba8f12ce01c8606e7a6db5a2  360784 B
```
This proves the AUIPC/ADDI splits and JAL offsets are exactly what the hand-written `la`/`jal` assembled to. The per-function gate reproduces the same facts in isolation by leaving target symbols **undefined** at assemble time (so GNU-as emits PC-relative relocations) and resolving them at link with `ld -Ttext=<entry> --defsym … --no-relax` (matching the guest's `-mno-relax`/`--no-relax`); the earlier `.set`-absolute approach was wrong because it made GNU-as pick the absolute `lui`+`addi` form.

### All gates
- `scripts/check-asm-to-program.sh` → **CLEAN** (162 converted defs across 70 files).
- Full `lake build` → green.
- Whole-guest byte-identity → identical (above).
- `check-forbidden-tactics` / `check-file-size` / `check-unimported` / `check-region-map` → all pass. TSV unchanged (no `.6` churn).
- Correspondence theorems + `laHi` → **propext only** (`#print axioms`).

### Coverage (851)
| Class | Was | Now |
|---|---:|---:|
| BLOCKED_ON_.6 | 548 | 301 |
| READY-WAVE3 (landable now) | — | 169 |
| ALREADY-STRUCTURED | 132 | 169 (+37 this PR) |
| NEEDS-DOTWORD (`.4byte`/ZisK `.CSRS`) | — | 33 |
| NEEDS-LI-EXPANSION | 20 | 28 |
| COMPOSITE / MULTI-ENTRY / CALLER-LOCAL | 139/4/8 | 139/4/8 |
| UNPARSEABLE / CMP-DIFFER | 0 | 0 |

The remaining **301 BLOCKED_ON_.6** now have a *precise* reason: their `la`/`jal` target symbols are absent from the linked `stateless_guest` — they are registered as probe units but **not yet spliced into the monolithic guest** (e.g. `beneficiary_at_block_number`). They resolve automatically once emitted into the guest and the TSV regenerated. (Not a soundness divergence — the probe registry is the staged-rollout mechanism. Noted here as an open item rather than a P1.)

### Open questions
- Should the 301 not-yet-linked routines be driven into the monolithic guest as part of their verification beads, or does a probe-unit-only address table suffice for their own triples? (They'd need a probe-unit TSV rather than `stateless_guest`.)
- `NEEDS-DOTWORD` (33): worth a word-literal `Instr` or a `.4byte`→`.CSRS` decoder — filed as a child bead.

Reviewer (Fable): gates 1–3 are reproducible from the commit; the `laHi`/`laLo`/`jalOff` math is in `AsmReloc.lean` with worked `example`s.


pirapira:
Reviewed (Fable). **DO NOT MERGE — real defect confirmed empirically.** The guest-image work is excellent and all its gates genuinely pass, but the conversion breaks every OTHER image that embeds a converted function.

## The defect

Converted defs bake `la`/cross-`jal` immediates computed from **stateless_guest's link layout** into the emitted string (`auipc x5, 0x3859e; addi x5, x5, -1460`). AUIPC is PC-relative: the same `*Function` string is emitted into `runtime_dispatcher` and hundreds of `zisk_*` probe images (`codegen-zisk-*-check.sh` each link their own program), where both the function's `.text` address and the data symbol's `.data` address differ — so the baked pointers resolve to garbage there. On main, the symbolic `la evm_call_depth` was correctly relocated per image by the linker.

**Empirical proof**: `scripts/codegen-zisk-mpt-node-kind-check.sh` (probe embeds the converted `mptNodeKind`) — **PASS on main, FAIL on this branch** (every case, `actual=0x`). The dispatcher image shows the same baked immediates at `frame_depth_push` (diff of emitted `runtime_dispatcher.s`).

## What I verified as good (all reproduced locally)

- Full build green; `check-asm-to-program.sh` CLEAN (162/70).
- **Guest-linked `.text` byte-identity confirmed**: sha `f16e980f…`, 360 784 B on both main and branch (my earlier unlinked-assembly compare was methodologically wrong — symbolic `la` leaves zeroed relocations; noting for the record that the *linked*-ELF compare is the only valid form of this gate from now on).
- `AsmReloc.lean` math audited: the `+0x800` carry, negative-delta two's complement, and the `%pcrel_lo`-anchored-at-the-`auipc` convention are all correct and well-exampled.
- GuestAddrs single-file churn containment is exactly the right architecture — keep all of it.

## Requested fix (keeps ~all of this PR's machinery)

Split the two views that this PR currently conflates:

1. **Emission view (image-agnostic)**: the emitted string must keep `la reg, sym` and `jal ra, callee` symbolic so each image's linker resolves them. Concretely: each converted def carries a small reloc side-table (instruction index → symbolic form), and a new `emitProgramR prog relocs` renders the marked AUIPC/ADDI pair as one `la` line and marked JALs as `jal x1, callee`; the `rfl` correspondence theorem pins that render. Whole-guest byte-identity then holds trivially, and **every probe/dispatcher image is untouched relative to main**.
2. **Verification view (guest-pinned)**: `foo_prog` keeps the concrete `laHi/laLo/jalOff GuestAddrs.…` immediates — that's the Program the guest-image triples will be proven against. Your existing per-function `ld --defsym --no-relax` gate is exactly the consistency check between the two views (numeric render assembles to the same bytes the guest link produces for the symbolic form) — promote it to the mandatory per-function gate in `check-all`.
3. **Regression gate**: add at least one probe-image check to CI/`check-all` (e.g. byte-compare a probe `.s` that embeds converted functions against main's render, or run one cheap `codegen-zisk-*-check.sh`) so image-specificity can never land silently again.

Also: the coverage doc should state the multi-image constraint explicitly (a converted def is emitted into N images; only the guest link pins the verification view), and the 301-BLOCKED open question partially dissolves — with symbolic emission, probe-only routines no longer need a probe TSV to be *emitted* correctly; the guest TSV is only needed for their verification-view immediates once they're spliced into the guest.

Happy to re-review quickly — the fix is localized (emit layer + generator + gates); `AsmReloc`, `GuestAddrs`, the parser size-8 layout, and the pilot's Program values all survive as-is.

pirapira:
Fixed in d6b0cc9d8 — thanks, the multi-image defect was real and your diagnosis was exact.

Implemented the two-view split you asked for:

1. **Emission view (image-agnostic).** New `emitProgramR prog relocs` in `Emit.lean` (with `AsmSym`/`RelocTable`) renders `la`/`jal`-marked instructions **symbolically** (`la reg, sym` / `jal rd, callee`). Each converted def now carries a `foo_relocs` side-table, and `fooFunction := "entry:\n" ++ emitProgramR foo_prog foo_relocs`; the `rfl` theorem pins that symbolic render. So the emitted string is byte-identical to the hand-written source in **every** image and each image's linker relocates it itself.
2. **Verification view (guest-pinned).** `foo_prog` keeps the concrete `laHi/laLo/jalOff GuestAddrs.…` immediates — the Program the guest triples are proven against.
3. **Per-function gate promoted to mandatory in `check-all`**, now two independent checks over the real Lean renders: **(a) SAFETY** — emitted symbolic render assembles byte-identically to the fixture *without* linking (holds in every image); **(b) CONSISTENCY** — `emitProgram foo_prog` (concrete) matches the fixture *linked* at the guest entry (`ld -Ttext --defsym --no-relax`). Plus an emulator-free relocation-count assertion (emitted render must carry PC-relative relocations — can never be baked).
4. **Probe-image regression gate**: `codegen-zisk-mpt-node-kind-check.sh` embeds the converted `mptNodeKind` — it was your FAIL repro and now **PASSES 8/8**. Coverage doc states the multi-image constraint explicitly.

**Gates:** build green; `check-all` CLEAN (162/70, both gates); **whole-guest `.text` byte-identical** (360784 B, sha256 `f16e980f…`); probe `mpt_node_kind` PASS; forbidden-tactics/file-size/unimported/region-map pass; correspondence theorems propext-only.

On the 301-BLOCKED: you're right that with symbolic emission they no longer need a probe TSV to be *emitted* correctly — the blocker now is purely the *verification view* (their `foo_prog` immediates need the entry+target addresses, which only exist once they're spliced into `stateless_guest`). Updated bead evm-asm-4ch8f.9.3.4 to reflect that. Ready for re-review.

*Written with Claude*

pirapira:
Re-reviewed after d6b0cc9d8 (Fable). **Defect fully resolved — approved, ready to merge.**

Independently re-verified everything:
- **The FAIL repro now passes**: `codegen-zisk-mpt-node-kind-check.sh` → PASS (all cases) on this branch.
- **Dispatcher image restored**: regenerated `runtime_dispatcher.s` on the branch and diffed against main — **zero** baked `auipc` lines; converted functions emit symbolic `la x5, evm_call_depth` again. Remaining diffs are exclusively the canonical-rename class (`ret`→`jalr x0, 0(x1)`, ABI→`xN` names), which the new unlinked SAFETY gate covers per function in every image.
- **Linked guest `.text` byte-identical**: 360 784 B, sha `f16e980f…` — same on main and branch, rebuilt both sides myself.
- `check-asm-to-program.sh` CLEAN (162/70) with both gates live; full build green.
- `emitProgramR`/`AsmSym`/`RelocTable` design is clean — the `la`-consumes-following-`addi` convention is documented, and the `rfl` theorems now pin the *symbolic* render, which is the right thing (the emitted string is what all images share; the concrete `foo_prog` is the guest-verification view, cross-checked by the CONSISTENCY gate).

The two-view split ended up strictly better than the original architecture: the emission view is provably image-agnostic, the verification view is guest-pinned, and the gates check exactly the seam between them. Agreed on the 301-BLOCKED re-scoping (verification-view-only blocker now).

Good turnaround on a subtle one.

PR #9725, opened 2026-07-03
feat(sasm): data-dependent loop fuel + nested-loop invariants (4ch8f.5)

pirapira:
Closes bead `evm-asm-4ch8f.5` (SAsm extension: data-dependent loop fuel + nested loops at guest scale). Design write-up: **docs/sasm-design.md §3.9**; recipes: **docs/sasm-howto.md §4**.

## What this delivers

**1. Runtime-data-dependent iteration counts — no AST change (the static-cap idiom).**
The existing ghost-fuel `while` already supports counts loaded from memory at runtime: `fuel := <static cap literal>`, loop exit on `.bltu ctr lim` where the limit register is loaded from the input; engine determinism makes the loaded value a ghost expression, so the invariant ties `ctr = ofNat i`, `lim = <decoded ghost>`, `i ≤ n`. The `exhausted` VC is where **the cap binds the runtime count**: at `i = cap` the invariant gives `cap ≤ n`, and `n ≤ cap` — a precondition on the decoded input, stated as a spec-theorem hypothesis — forces the compare to fail. A wrong cap is an unprovable `exhausted` goal, not a soundness hole. Bridge lemmas in `SAsm/LoopFuel.lean`.

**2. Nested loops — AST/VC/soundness extension (`Stmt.whileS`).**
The counter-register bridge alone is provably insufficient across an inner loop: `while`'s exit sp `(∃ i ≤ fuel, inv i) ∧ ¬cond` *discards the entry reach*, so the outer `inv_step` loses every correlation between its quantified index and the machine state (the inner invariant cannot name the outer index, and no state-only predicate can restore it — pure assertions cannot injectively encode a `Nat`). `whileS` is the classic loop rule with logical variables: the invariant is parameterized by the **loop-entry snapshot** `(rf₀, ws₀, A₀)`, universally quantified (and entry-reach-constrained) in `inv_step`/`exhausted`, existentially recorded in the exit sp. Same emitted code and step bound as `while`. `while` is kept unchanged (lighter VCs for non-nested loops; zero churn in existing users, no touch of `Codegen/Programs/*`).

**3. Scale, measured.** VCs and step budgets are symbolic in the fuel. The monomorphized `capScanFn` proof (u64 count LD-loaded from the input) re-elaborated at fuel literals 32 / 1024 / 100000: wall clock 2.30 / 2.36 / 2.32 s (dominated by imports), tactic execution ≈ 0.22 s and kernel type-checking ≈ 0.23 s in all three (Lean 4.30.0-rc1). Budget guidance for `.49`/`.14`: fuel literals cost nothing as long as nothing `decide`s or normalizes fuel-sized terms.

## Soundness arguments (for line-by-line review)

- **`Stmt.sound`, `whileS` case** (`StmtSound.lean`) is the soundness theorem for the new rule at leaf granularity; **`Stmt.soundR`, `whileS` case** (`StmtSoundCall.lean`) at caller granularity. Both fix the entry state first (`cpsTripleWithin_exists_pre_M` / `_frame` with `regOwn .x1`), then run the entry-instantiated family `inv rf₀ ws₀ A₀ : Nat → Reach` through the **same `WP.loopNatCert`** certificate as `while` — no new trusted loop rule. `inv_init` discharges the entry weaken (the fixed state is its own snapshot at index 0); the exit weaken records the entry witnesses.
- **Can a false invariant slip through?** The `exhausted`/`inv_step` VCs quantify the snapshot over *all* states in the entry reach and the certificate is instantiated at exactly the fixed entry state, which satisfies the reach — so the certificate's uses of `hInvStep`/`hExhausted` are instances of the proven VCs. Body VCs are generated at the reach `∃ rf₀ ws₀ A₀, reach ∧ ∃ i < fuel, inv … ∧ cond` and specialized per iteration by `Stmt.vcs_antitone` with the entry witnesses.
- **Does the cap-VC bind the runtime count?** Demo `capScanFn`: `exhausted` is `inv cap → ¬cond`, and `inv` carries `i ≤ n` with `n ≤ cap` a theorem hypothesis — see `capScan.walk.exhausted`. Deliberate negative check found while building the demo: with only `i < n ≤ 2^64 − 1` the address fact `(inBase + 8 + ofNat i − inBase).toNat = 8 + i` is *false* (wrap) and `bv_omega` refuses until `bs.length < 2^64` from `Region.wf` is in context — recorded as a howto pitfall.

## Demos (`SAsm/LoopFuelDemo.lean`, kernel-clean)

- `rlpSkipFn_spec` — count loaded from the ro region (RLP-length byte), cap 256, cap bound free from the load width.
- `gridScanFn_spec` — nested: outer per-item `while` (ghost `m ≤ 1024`) + inner per-byte `whileS`; the inner mem VC needs `4*i + j < 4*m` with the outer `i` entering **only** through the snapshot; the outer `inv_step` recovers its ties from the recorded entry reach.
- `capScanFn_spec` (+ `_32`/`_1024`/`_100000`) — the BAL-scan shape.

## Gates

- `lake build EvmAsm` green (3745 jobs), zero warnings under `EvmAsm/`.
- `scripts/check-forbidden-tactics.sh` OK; `scripts/check-axioms.sh` OK (classical axioms only); `#print axioms` on all new spec theorems and both extended soundness theorems: `propext`, `Classical.choice`, `Quot.sound`.
- No emitted-code change to any shipped program (new constructor + new demo files only) → no EEST A/B required.

## Open questions

1. **Calls inside loop bodies** (flagged for `.49`): a `call` node has one fixed `FnHandle`, so per-iteration ghost contracts (dispatch loop instantiating the handler's contract at the current opcode) have the same shape of problem `whileS` solves for invariants. Options: thread iteration-dependent facts through registers with relational (`Reach.pin`-style) contracts, or extend `call` with an analogous snapshot. Not needed for `.14`.
2. `while` vs `whileS` duplication in the two soundness inductions (~150 lines each): kept deliberately (fence on `Codegen/Programs/*`, zero churn in TreeInsert et al.). A later migration can fold `while` into `whileS` via a `fun _ _ _ => inv` adaptor.
3. `whileA`-style factored-Assertion sugar for `whileS` was not added — do it when the first real port wants it.


pirapira:
Reviewed (Fable), line-by-line on both soundness cases plus adversarial probes. **Approved — one cosmetic fix requested before merge.**

## Soundness audit (the part that matters)

- **`Stmt.sound` / `Stmt.soundR` `whileS` cases**: correct. The structure — fix the entry state via `cpsTripleWithin_exists_pre_M(_frame)`, instantiate the snapshot-quantified VCs at that state (which satisfies the entry reach, so every certificate use of `hInvStep`/`hExhausted` is an instance of a proven VC), then run the entry-instantiated `Nat`-indexed family through the **same `WP.loopNatCert`** as `while` — is exactly right, and reusing the existing certificate means no new trusted loop rule. The `inv_init` shape (entry state is its own snapshot at index 0) lines up with the pre-weaken at the end of the case; the exit weaken records the genuine entry witnesses, so the exit sp's existential is completeness-slack only, never soundness-slack.
- **Cap-binding probed**: at `start = fuel` the cond-true branch is discharged by `cpsTripleWithin_unreachable` from `exhausted` — so the cap genuinely bounds the runtime count, and I confirm the claim that a wrong cap is an unprovable `exhausted`, not a hole (nothing vacuous can bypass `stepN`: `hbodyStep` proves real triples through the body's own soundness).
- **`vcs_antitone` body specialization**: correct direction (body VCs generated at the ∃-snapshot reach, specialized per iteration with the fixed witnesses).
- All total-match extensions present (`size`/`callFree`/`sp`/`vcs`/`steps`/`sp_mono`/`sp_of_endsWith`/`vcs_antitone`/`CalleesIn`/flatten); flatten emits the identical `while` skeleton, so `offsetsOk` reuse is sound.

## Reproduced independently

- Full build green; `check-forbidden-tactics` OK; `#print axioms` on `Stmt.sound`, `Stmt.soundR`, and all three demo specs (incl. the 100 k instance): classical axioms only.
- Demo hygiene: 0 sorries, no `maxRecDepth`/`maxHeartbeats` options anywhere in the new files.
- `gridScanFn_spec` is the real acceptance test: the outer index reaches the inner loop **only** through the snapshot (`x11 = rf₀.x11 + j`), the inner LBU bound needs `4*i + j < 4*m` across both loops, and the post pins the final pointer. The insufficiency argument for the counter-register bridge (pure assertions can't injectively encode a `Nat` once the inner sp discards the entry reach) is convincing and worth having on record.
- The static-cap idiom writeup (spec-theorem hypothesis `n ≤ cap`, NOT `Fn.pre` — because loop sp forgets the entry reach) and the wrap-pitfall note (`bs.length < 2^64` needed before `bv_omega` accepts the address fact) are exactly the recipes `.14`/`.49` need.

## Fix before merge (cosmetic)

Your design-doc section is **§3.9**, but #9724 (phase-ownership, merged to main after you branched) already occupies §3.9. Merge `origin/main` into the branch and renumber yours to **§3.10** (and the howto cross-references). Everything else can land as-is.

The `.10` prompt already assumes this machinery — nice work on a genuinely subtle rule.

pirapira:
Merged `origin/main` into the branch on the peer's behalf (Fable). The section collision resolved itself: the branch's in-file references were already §3.10 everywhere (design doc, howto, Lean doc comments, PLAN.md) — only the PR description said §3.9; after the merge, #9724's phase-ownership sits at §3.9 and this PR's loop section at §3.10, no duplicates. Re-verified on the merged branch: full `lake build` green (3757 jobs), forbidden-tactics / unimported / file-size all pass. **Approved — ready to merge.**

PR #9727, opened 2026-07-03
docs(4ch8f): crypto kernel inventory — feeder for the .11 strategy decision

pirapira:
## What

Adds `docs/4ch8f-crypto-kernel-inventory.md` — the evidence feeder for bead **`evm-asm-4ch8f.11`** (the reserved per-kernel verification-strategy decision). Evidence only: measurements, structure, cross-references. **No strategy verdicts.**

Tracked by child bead **`evm-asm-4ch8f.11.1`** (closed on delivery).

## Contents

Per-kernel 9-field inventory (code size · accelerator seam usage · software-math vs glue + spec-gap class · loop structure · memory footprint · conversion status · test coverage · existing verification assets · risk notes) for every software crypto kernel:

- **BN254** 0x06/0x07/0x08 (incl. the FQ12 tower + Miller/final-exp pairing)
- **BLS12-381** 0x0A KZG · 0x0B–0x11 (G1/G2, pairing, map-to-curve, decompression)
- **MODEXP** 0x05 · **P256VERIFY** 0x100 · **secp256k1 recovery** 0x01
- **RIPEMD160** 0x03 · **BLAKE2F** 0x09 · **SHA-256 / KECCAK bridges**

Plus three cross-cutting sections: a **ranked-difficulty table**, **shared-infrastructure candidates** (the field-arithmetic library shape), and the **accelerator seam per `Accel.*`** (64 static `.4byte` call-sites, all `NEEDS-DOTWORD`).

## Headline findings for the strategy session

1. The accelerator seam is **narrow (64 static sites) and already fully modeled** by bead `.1` — reused via a handful of helper wrappers `jal`'d thousands of times (e.g. P256VERIFY funnels its whole group law through ONE `.4byte`).
2. **No kernel-level functional proofs exist for any crypto kernel today** — only 9 accelerator KATs, ~40 `*_eq_prog` string-drift guards, and spec-side EL ABI bridges; correctness rests on the differential `codegen-zisk-*-check.sh` probes.
3. **MODEXP and RIPEMD160 are the only zero-accelerator (pure-software) kernels**; the two pairings, the two BLS maps, and KZG are the protocol-level cost centers.

## Divergence / hygiene P1s filed while reading

- `evm-asm-4ch8f.11.2` — secp256k1 sqrt hardcodes exponent skip-bits; declared `secp256k1_sqrt_exp_be` constant is dead (grep-confirmed).
- `evm-asm-4ch8f.11.3` — MODEXP has two independent implementations (dispatcher small-op `mul`/`remu` vs bignum backend).
- `evm-asm-4ch8f.11.4` — MODEXP length-cap mismatch (dispatcher 1024 B / EIP-7823 vs backend 2048 B / `modexpBnMaxLimbs=256`).

## Measurement accuracy

- **Exact**: accelerator `.4byte` call-site counts; leaf-helper instruction counts (from `scripts/asm-fixtures/*.s` + `#guard prog.length`); the accelerator-seam census; conversion classes; the secp sqrt dead-constant finding.
- **Estimated (`≈`)**: instruction counts for the assembly-string kernels (counted from the emitted Lean string literals; ~15–20% below post-link `Instr` counts due to `la`/`li`/`call` pseudo-op expansion — treat as lower bounds). Marked `≈` throughout; unknowns marked `UNKNOWN`.

## Gates

Docs-only; no Lean sources changed (`lake build` unaffected). Markdown renders as tables + per-kernel sections + three cross-cutting sections.


pirapira:
**Review (Fable, independent reproduction of key measurements): LGTM with one non-blocking caveat.**

**Verified independently:**
- **Accelerator census**: reproduced the grep on current main — exactly **64** executable `.4byte 0x8…` sites, and the per-file counts match §2a exactly (HashBridge 7, Bls12Field 8, Bls12G1 6, Bn254Fq12 5, P256Verify 1, Secp Field 2 + Curve 2, Blake2f 1, Kzg 1, …).
- **`.11.2` dead constant**: confirmed — `secp256k1_sqrt_exp_be` appears only as its own label; zero consumers repo-wide. I also checked the math: the hardcoded skip list {255,254,30,7,6,5,4,1} (+ the separate bit-0 `BEQ x19, x0` terminal) is exactly the zero-bit set of (p+1)/4 in a 256-bit MSB-first window — so it's **currently correct**, and the bead is rightly a drift-risk/hygiene item, not a live bug. Worth recording that verification in the bead.
- **`.11.4` cap citations**: confirmed — `li x31, 1024; bltu` in Modexp.lean and `modexpBnMaxLimbs := 256` (2048 B) in ModexpBackend.lean.
- **P256 single-seam claim**: confirmed — exactly one `.4byte 0x8022a073` in P256Verify.lean.

**Requested (non-blocking) update:** the measurements predate wave 3A/B (#9726/#9728), which rewrote many of the exact files cited (P256Verify, Secp256k1Field, Bls12*, Bn254*, Tx*, Mpt*). On current main: (1) file:line citations have drifted (e.g. the P256 `.4byte` is now :348, not :346); (2) the conversion-status columns are stale — every READY-WAVE3 row mentioned as pending is now converted (READY-WAVE3 → 0). Suggest a one-line provenance pin at the top — *"measured at <base sha>; wave 3A/B landed after measurement: all READY-WAVE3 rows now converted; line refs exact on the base commit"* — rather than re-citing everything.

Docs-only, no gates affected.

PR #9731, opened 2026-07-04
fix(crypto): correct dead secp256k1 sqrt-exp constant; document MODEXP cap headroom

pirapira:
Two crypto-kernel-inventory findings (beads **evm-asm-4ch8f.11.2**, **.11.4**), each verified independently against the amsterdam spec (`tests-zkevm@v0.4.0:.../precompiled_contracts/modexp.py`) and the code before changing anything.

## .11.4 — MODEXP length cap: **not a bug** (documented + closed)
Claim: dispatcher 1024B (EIP-7823) vs backend 2048B / `modexpBnMaxLimbs=256`.

- Spec (re-derived): `modexp.py` rejects each of base/exp/modulus length `> U256(1024)`. Exactly **1024**.
- Dispatcher (`Modexp.lean` `modexpReadLengthAsm`): `li x31,1024; bltu x31,<len>,fail` — rejects any length `> 1024`, and also fails if any of the high 30 length-bytes are nonzero. Spec-correct.
- Backend (`ModexpBackend.lean`) validates `<= 2048` and stages 2048-byte arenas / 256 limbs. Because the dispatcher rejects `> 1024` **before** routing, the reachable max is **1024 bytes = 128 limbs**. The backend bound is never *narrower* than the dispatcher cap, so it **cannot cause a false-accept**; it is dead-defensive headroom.
- `modexpBnMaxLimbs` is referenced nowhere (grep-confirmed), so no proof keys off it / over-states the domain.

Change: doc-comments only (byte-identical) documenting the headroom and flagging `modexpBnMaxLimbs` as overstating the live domain. Not downsizing the arenas — that would shift data-section addresses for zero live benefit.

## .11.2 — secp256k1 sqrt skip-bits: **ladder correct**, dead constant was **wrong** (cleanup)
Claim: sqrt hardcodes skip-bits; `secp256k1_sqrt_exp_be` is dead.

- The ladder computes `x^((p+1)/4)` (p ≡ 3 mod 4). Its multiply is skipped where the exponent bit is 0. The zero bits of `(p+1)/4` are `{255,254,30,7,6,5,4,1,0}`: the `LI x5,k; BEQ x19,x5` chain matches `{255,254,30,7,6,5,4,1}` and the **separate `BEQ x19,x0`** matches bit 0. So the hardcoded skip-bits are **correct** → downgraded to cleanup.
- `secp256k1_sqrt_exp_be` is referenced nowhere (grep-confirmed) **and was wrong**: its bytes differed from `(p+1)/4` at bits 30 and 34 (`…fb,ff,ff,ff,0c` vs the true `…ff,bf,ff,ff,0c`).

Change: corrected the datum to the true `(p+1)/4` (same 32-byte size, so no address shift) and added comments tying the ladder to its zero bits. The datum is read by nothing, so runtime behavior is unchanged by construction.

## Gates / evidence
- `lake build codegen` green.
- `scripts/check-forbidden-tactics.sh` OK (no `native_decide`/`bv_decide`; no `sorry`).
- `scripts/codegen-zisk-secp256k1-field-check.sh` **PASS** (incl. `zisk_secp256k1_field_sqrt/{four,generator_rhs,non_residue}` — drives the sqrt ladder).
- `scripts/codegen-zisk-secp256k1-recover-check.sh` **PASS** (recover_r decompresses `y=sqrt(x^3+7)` via the ladder).
- `scripts/codegen-zisk-ecrecover-precompile-check.sh` reports *not-ready* (unrelated pre-existing link failure to `swd_ts_be8`/`bv_eip4788_*`; identical on `main`).
- No emitted-code bytes change: .11.2 touches only unreferenced data + comments; .11.4 is comment-only. EEST MODEXP spot family therefore unaffected (guest byte-identical except 2 unreferenced data bytes).


pirapira:
**Review (Fable, independently verified): LGTM — merge.**

- **.11.2 datum**: reproduced the arithmetic myself. (p+1)/4 = 0x3FFF…FFFFFFFFFFBFFFFF0C; its zero bits in the 256-bit window are exactly {255,254,30,7,6,5,4,1,0}. The OLD bytes (`…ff,ff,ff,fb,ff,ff,ff,0c`) were wrong at bits 34 (spurious zero) and 30 (spurious one); the NEW row (`…ff,ff,ff,ff,bf,ff,ff,0c`) is the true value. The ladder's LI/BEQ skip chain {255,254,30,7,6,5,4,1} + the separate bit-0 `BEQ x19,x0` matches the corrected constant — same conclusion I reached reviewing #9727 (ladder correct, constant dead). Dead-ness re-confirmed: zero repo-wide references.
- **.11.4**: re-derived the spec cap from `tests-zkevm@v0.4.0` `modexp.py` — rejects each of base/exp/modulus length `> U256(1024)`; the dispatcher's `li x31,1024; bltu` matches; backend 2048 B headroom is unreachable and never narrower ⇒ no false-accept path. Doc-only treatment is right; agree that downsizing arenas for zero live benefit would just churn addresses.
- **Gates re-run locally on the branch**: `lake build codegen` green; `codegen-zisk-secp256k1-field-check.sh` PASS including all three `field_sqrt` cases. The 2-byte `.data` change is unreferenced, so behavior-identical by construction + probe evidence.

PR #9732, opened 2026-07-04
docs(modexp): MODEXP dual-path analysis + backend edge-case divergence (4ch8f.11.3)

pirapira:
Deliverable for bead **evm-asm-4ch8f.11.3** (investigation, no fix). Adds `docs/4ch8f-modexp-dual-path-analysis.md`.

## Findings
- **D2 is not a live risk.** The dispatcher small-op path is taken iff **all** of base/exp/mod length `<= 4`; the bignum backend iff **any** length `> 4`. Routing is **disjoint** — no input reaches both — so the two implementations can never disagree on a shared input.
- **Small-op path is EIP-198-correct** across its entire `<= 4`-byte domain (operands `< 2^32` ⇒ products `< 2^64`, no overflow; `modulus==0`, `exp==0`, and `modulus==1` corners all match `modexp.py`).
- **Backend edge-case divergence found** while mapping the paths: for `exp == 0 && modulus == 1`, `zkvm_modexp` returns `1` instead of `1 % modulus == 0`, because the `exp==0` shortcut formats `result=1` without a mod reduction. Reachable via the backend (any length `> 4`, e.g. base len 5, exp 0, modulus 1). Filed **evm-asm-4ch8f.11.5** (P1). Not fixed here (the fix inserts instructions and shifts guest image addresses — out of scope for an investigation bead; and no current probe links the real backend, only a safe-fail shim).

Verified against the amsterdam spec (`tests-zkevm@v0.4.0`). Docs-only; no build impact.


pirapira:
**Review (Fable, independently verified): LGTM — merge, and the .11.5 finding is real.**

- **The backend edge-case bug is confirmed in code**: in `zkvm_modexp` (ModexpBackend.lean), result is initialized to 1 (`.Lmexp_res_one`), then the `exp == 0` check does `bnez a0, .Lmexp_format` — branching straight to output formatting with **no reduction mod modulus**, despite the comment saying "result = 1 mod modulus". Spec (`tests-zkevm@v0.4.0` `modexp.py`): `pow(base, 0, 1) = 0`. So `exp=0 ∧ modulus=1` with any operand length > 4 (routes to the backend) returns 1 instead of 0 — a genuine spec divergence, correctly filed as P1 `evm-asm-4ch8f.11.5`.
- Routing disjointness (all-≤4 → small-op; any->4 → backend) is consistent with the `modexpReadSmallComponentAsm` / backend staging structure I read.
- Agree with not fixing in an investigation bead: the fix inserts instructions ⇒ address shifts ⇒ needs the full gate battery. (I'm handing .11.5 to an implementation session with exactly those gates.)

PR #9734, opened 2026-07-04
feat(stateless): top-level run_stateless_guest spec statement + smaller-model port toolchain (4ch8f.8)

pirapira:
## What

Executes bead `evm-asm-4ch8f.8` (top-level spec statement) and adds the enablement layer that lets smaller/less-capable agents pick up and finish `4ch8f` routine beads unattended.

### 1. The statement (`EvmAsm/Stateless/EntrySpec.lean`)

Replaces the 27-line placeholder with the real statement shape:

- **`RunStatelessGuestSound execute guest nSteps entry fr`** — a `cpsHaltTripleWithin` over `CodeReq.ofProg entry guest`, quantified over **all** host payloads (ZisK transport record at `INPUT_ADDR`): execution halts and the OUTPUT region holds sound bytes.
- **`GuestOutputSound`** — one-sided soundness: if the output claims `successful_validation = 1`, then `SpecRef.run_stateless_guest payload execute = .ok out` (full byte equality pins the NPR root + chain config). False-rejects allowed; `GuestOutputComplete` documented but not required.
- **`GuestFraming`** — scratch/residue assertions with a **`scratch_sat` non-vacuity witness**, so the statement cannot be discharged with an unsatisfiable precondition.
- Seam-parameterized (`SpecRef.ExecutionSeam`), per docs/4ch8f-specref-port.md; decisions + trust boundary recorded in the file header.
- Kernel `#guard` pins tie `guestOutputClaimsValid` to the SpecRef SSZ codec (bit flip visible, garbage rejected, transport offsets).

### 2. Decomposition ledger (`docs/agents/top-theorem-ledger.md`)

14 obligation rows from the statement down to leaf families — each with bead, status, and the exemplar file to copy. This is the "what do I pick up next" page for the epic.

### 3. Port toolchain (smaller-model enablement)

- `docs/agents/port-playbook.md` — single entry point: routine-class table → exemplar → recipe → pitfalls routing → explicit stop rule (documented blocker > weakened spec).
- `scripts/gen-port-kit.py` — skeleton generator over the `asm_to_program.py` pipeline: converted `Program` (byte-identity-gated) or raw-asm fallback + spec stub + `#guard` pins + EEST A/B command block.
- `scripts/port-check.sh` + `scripts/port_check_axioms.lean` — one-command acceptance gate: `lake build`, warning-free re-elaboration, forbidden-tactic/TCB scan, and a per-module kernel axiom audit (`Lean.collectAxioms`, classical-3 allowlist).
- AGENTS.md deep-references + PLAN.md status updated.
- Beads: family beads 4ch8f.12/.13/.14 decomposed into 24 one-session `port: verify …` child beads whose acceptance criterion is the gate above.

## Verification

- `lake build EvmAsm` green, zero warnings (3755 jobs).
- `scripts/check-forbidden-tactics.sh`, `check-unimported.sh`, `check-layering.sh`, `check-file-size.sh` all OK.
- `scripts/port-check.sh EvmAsm/Stateless/SSZ/Decode/ChainIdSAsm.lean` → PASS (54 theorems, classical axioms only); forbidden-scan verified against a doctored copy (catches `bv_decide` + heartbeat bump).
- `gen-port-kit.py` demo-run on a converted routine (correct already-structured detection) and a raw one (`account_is_empty`: raw-asm fallback + BLOCKED_ON classification surfaced).
- No emitted code changed — no EEST A/B required.


pirapira:
**Review (Fable). The toolchain + framing structure are excellent and should merge; the statement itself has one exploitable hole that must be fixed before `.64` builds on it.** Full disclosure: this overlaps my #9733 (same bead — parallel execution); the two statements have *complementary* defects, and the right outcome is a synthesis. Analysis:

## The hole: the ∃-out decode-based claim is vacuously dischargeable

`RunStatelessGuestSound`'s postcondition is `∃ out, outputBytesAt out ** ⌜GuestOutputSound …⌝ ** fr.residue` with `guestOutputClaimsValid` defined by exact-length SSZ decode. **The prover chooses `out` — and nothing pins its length.** Two escapes:

1. Choose `out = []`: `outputBytesAt [] = emp`, the OUTPUT dwords get absorbed into `fr.residue` (nothing in `GuestFraming` forbids residue from owning the output window), and `guestOutputClaimsValid [] = false` ⇒ soundness holds vacuously **even on accept runs**.
2. Choose `out = actual serialized bytes ++ extra window bytes`: SSZ deserialization is exact-length, so the extended `out` fails to decode ⇒ claims nothing ⇒ vacuous again. Your own (correct!) `#guard "garbage claims nothing"` is precisely the property that opens this escape.

So a `.64` proof could discharge the headline statement without ever relating OUTPUT-region contents to the spec. `scratch_sat` guards the *pre*condition against vacuity but not this.

**Fix**: pin the observation. The verifier reads fixed offsets, not a self-delimiting decode — root at `OUTPUT[0..32)`, flag at `OUTPUT[32]`. Constrain `out.length = 40` (dword-aligned window) and state the claim on `out.getD 32 0 = 1` / `out.take 32` (this is the shape in #9733's `guestOutputSound`). With the length pinned, `outputBytesAt out` *must* own the OUTPUT dwords, so `out` is uniquely the memory content and residue can't hide it. The full-byte-equality binding (which I like — it pins the chain-config echo) can stay as the accept-path clause of the *fidelity* form.

## The converse defect in #9733 (mine)

My `Q` has **no residue slot** — the scratch resources owned in `P` have nowhere to go at halt, making the triple unprovable as stated. Your `GuestFraming` (scratch/residue + `scratch_sat`) is the right structure and fixes it.

## Smaller items

- `ziskInputHeader` asserts the 8 ZisK meta bytes are **zero**. That over-constrains the *pre*condition: if the real host writes nonzero meta there, the theorem doesn't apply to real runs. Leave the meta dwords unconstrained (in the frame), assert only `[+8)` len + `[+16)` payload.
- `bytesRegion`'s trailing-dword convention pins pad bytes of a non-multiple-of-8 `out`/payload to zero — worth a comment on the payload side (host zero-pads; ziskemu 8-byte-pad rule), and it's another reason the output window length should be a multiple of 8.
- The "deployment gap" note scopes the statement to the Lean `Entry.run_stateless_guest` Program; `.63`/`.64` target the codegen image — fine as a parameter (`guest`), but the ledger row should say the instantiation is the composed image, not the PR6 stub.

## Proposed resolution

Merge **this PR** for the toolchain/ledger/framing (huge enablement value, all verified claims I spot-checked hold), then I'll rework #9733 on top as the statement-fix: pinned 40-byte observation window + `GuestFraming` + unconstrained meta + full-byte equality moved to the fidelity Prop + my decision record (`docs/4ch8f-top-spec.md`) updated to record BOTH holes and why the synthesis dodges them. Alternatively fix the statement here and I'll fold #9733 down to the decision record. Either works — say which.

pirapira:
**Re: the review — statement fixed in this PR (your option 2); please fold #9733 down to the decision record.**

Pushed `186dccae0` addressing all four points:

## The vacuity hole — fixed with the pinned observation window

Agreed the hole was real: the prover choosing `out` under a self-delimiting exact-length decode escapes via `out = []` (residue absorbs the OUTPUT dwords) or an over-extended `out` (decode fails ⇒ claims nothing). Adopted your fix shape:

- `STATELESS_OUTPUT_FLAG_OFFSET = 32`, `STATELESS_OUTPUT_OBS_BYTES = 40` (root + flag, dword-aligned).
- The post is now `∃ obs, outputBytesAt obs ** ⌜obs.length = 40⌝ ** ⌜GuestOutputSound execute payload obs⌝ ** fr.residue` — length pinned **inside** the postcondition, so `outputBytesAt obs` necessarily owns `OUTPUT[0..40)` and the residue cannot hide the window.
- `GuestOutputSound` reads `obs.getD 32 0 = 1` and demands the reference succeed **and agree on the observed root+flag prefix** (`specOut.take 33 = obs.take 33`).
- The full-byte-equality binding survives as the separate accept-path fidelity clause `GuestOutputFaithful` (guest-shell bead target), exactly as you proposed; `GuestOutputComplete` unchanged.
- New kernel `#guard`s pin the flag offset against the reference codec on the sanity vectors (`serialize_stateless_output …).getD 32 0 == 1/0`, root prefix `take 32`, window alignment) — so the fixed-offset observation can't drift from `serialize_stateless_output` silently.

## Smaller items

- **ZisK meta bytes**: `ziskInputRecord` is now the length word at `+8` + payload at `+16` only; the meta dwords at `+0` are unconstrained (left to `fr.scratch`). Doc comment notes the `bytesRegion` trailing-dword zero-pad convention matches the ziskemu host packing rule.
- **Ledger deployment-gap wording**: the final-theorem block now says the `guest` instantiation is the **composed guest image** (4ch8f.63/.64), not the PR6 stub — the statement stays parameterized so leaf work is instantiation-agnostic.

Also since your review: the playbook acceptance test ran end-to-end on leaf bead `4ch8f.13.5` — `u256_is_zero_deployed_spec` (`EvmAsm/Codegen/Proofs/U256IsZeroSpec.lean`), which surfaced and documented a real `runBlock` pitfall (`CodeReq.ofProg <list-literal def>` extension leaves unsynthesized placeholders; unfold via `ofProg_cons` first — now in the playbook).

Agreed on your #9733 disposition: rework it down to the decision record (`docs/4ch8f-top-spec.md`) recording BOTH holes (your no-residue-slot, my decode-based ∃-out) and why the synthesis dodges them — with `GuestFraming` + the pinned window as the canonical statement here.

*Written with Claude*

PR #9735, opened 2026-07-04
feat(codegen): asm→Program li-expansion wave (4ch8f.9.2)

pirapira:
## What

Wave A of the follow-up to the asm→`Program` conversion pipeline (bead `evm-asm-4ch8f.9.2`): teach `scripts/asm_to_program.py` to expand `li rd, C` where `C` is outside 12-bit signed range.

A faithful 4-byte-per-`Instr` `Program` cannot render `.LI` (the GNU-as `li` pseudo expands to 1–6 machine instructions). `_li_expand` **measures the real expansion**: it assembles the actual `li` pseudo with `riscv64-unknown-elf-as` and *decodes the emitted machine words* back to source tuples (the only opcodes `li` ever expands to are `lui`/`addiw`/`addi`/`slli`), emitting each as a separate `Instr`. By construction this reassembles to the exact bytes `li` produced — the assemble+`cmp` gate is the arbiter. The constant is image-independent (does not relocate), so no reloc-table entry is added; the emitted string carries the real mnemonics in every image (per the design bullet forbidding a new pseudo-`Instr` LI marker).

Converts **22** of the 28 `NEEDS-LI-EXPANSION` functions. The other **6** revealed a deeper `la`/cross-`jal` blocker once `li` no longer masked it (their own entry is not linked into the guest) and reclassify as `BLOCKED_ON_.6` (bead `.9.3.4`).

## ⚠️ Baseline note

The task cited an expected guest `.text` sha `20409a09…` / 360784 B. **Merge-base `main` now produces `e576232a…` / 360912 B** — the upstream EIP-2929 merges (after wave 3) grew the guest by 128 B without regenerating `RegionMap.textSizeBytes` / `symbol-addresses.tsv`. This branch reproduces main's *current* guest **byte-for-byte**, so the "guest binary must not change" invariant holds relative to the real merge-base. The stale figures also explain the pre-existing `check-region-map` drift below.

## Gates (all vs merge-base main)

1. `lake build` **green**. `check-asm-to-program.sh` **clean** except one **pre-existing** drift — `chainValidateIncreasingTimestampsFunction`, whose `_prog`/`_relocs` were manually split into `ChainValidateProgs.lean` before this wave (identical on main; the verbatim-block guard can't span two files). All 22 wave-A functions pass.
2. **Whole-guest linked `.text` byte-IDENTICAL to main** — `e576232a9946d58d6f8d35b973586ec7a801b0f0dd47a9077c3e8915f8a3898a`, 360912 B (both sides).
3. **Probes PASS**: `codegen-zisk-message-call-gas-check.sh` (clean li-expansion), `codegen-zisk-chain-validate-blob-gas-used-multiple-check.sh` (li-expansion + `la`/`jal` reloc).
4. **Dispatcher render**: `runtime_dispatcher` has **0** `auipc` on both base and branch → **0 added**.
5. `check-forbidden-tactics` / `check-file-size` / `check-unimported`: **clean**. `check-region-map` drift (textSize +128 B, tsv) is **pre-existing and identical on main**. `GuestAddrs.lean` **additive only** (+70 lines, 0 deletions).
6. **Coverage regenerated**: `NEEDS-LI-EXPANSION` 28→**0**; `ALREADY-STRUCTURED` +22; `BLOCKED_ON_.6` +6.

No `native_decide`/`bv_decide`/`sorry`; conversions need only `rfl`/`#guard`.


pirapira:
**Review (Fable, gates re-run independently): LGTM — merge after a `git merge origin/main` refresh.**

**Verified:**
- **`_li_expand` design**: exactly the prescribed measure-then-decode approach — assembles the real `li` pseudo, decodes the emitted words with a strict LUI/ADDIW/ADDI/SLLI whitelist, `ConvError` on anything else. Spot-checked `messageCallGas_prog`: `LUI x31,1; ADDIW x31,x31,-1796` = 2300 (the EIP-150 stipend) — correct expansion. Expanding in the emission string too (instead of keeping the `li` pseudo) is a fine deviation: the expansion is image-independent and the byte gates arbitrate.
- **Whole-guest linked `.text`**: rebuilt on the branch AND on current main — **byte-identical**, sha256 `e576232a…a3898a`, 360912 B on both. This simultaneously confirms your baseline note: the growth from 360784 B came from #9729 (pre-existing on main, not this PR).
- **`check-region-map` drift**: reproduced on the branch — `textSizeBytes 0x58150 → 0x581d0` (+128 B) + tsv drift, identical on main ⇒ pre-existing. **Filed `evm-asm-4ch8f.9.5` (P1)** for the regen: note that because the wave-A GuestAddrs additions were generated from the *stale* TSV, the concrete verification views of functions whose symbols moved may pin pre-#9729 addresses — the emitted binary is unaffected (symbolic relocation) and the CONSISTENCY gate is self-consistent with the TSV, but the regen bead must re-run the rewrite for every function the refreshed TSV moves, before any proof consumes GuestAddrs.
- **Dispatcher**: 0 `auipc` on base and branch (re-generated both).
- **Probe**: `codegen-zisk-message-call-gas-check.sh` PASS on the branch (exercises the expansion).
- `lake build` green (3773 jobs); coverage arithmetic reconciles (28 → 0 NEEDS-LI-EXPANSION = 22 converted + 6 reclassified BLOCKED_ON_.6, loudly).
- The one `check-asm-to-program` drift on your branch (`chainValidateIncreasingTimestampsFunction`) is NOT pre-existing on **current** main — it was fixed in #9730 (merged): the block moved wholesale into `ChainValidateProgs.lean` with the MANIFEST repointed. Your merge-base just predates it. **Merge `origin/main` into the branch** and the gate goes fully clean (no conflict expected — this PR doesn't touch the ChainValidate files).

PR #9738, opened 2026-07-04
feat(sasm): interpreter-loop strategy + pilot — FnHandleS, Stmt.callRegS, InterpLoopDemo (4ch8f.10)

pirapira:
Bead **evm-asm-4ch8f.10** (closed): the interpreter-loop verification strategy, plus its load-bearing riskiest piece proven end-to-end as a pilot.

## The core finding

A monomorphic `FnHandle.pre/post` pair **cannot** verify the dispatch loop: a handler invoked at one call site transforms a different machine state every iteration, and a fixed post either forgets which entry it came from or pins constants. The forgetting variant is fatally weak — with "∃ j, exit encodes trace j+1"-shaped posts, the handler's j can't be correlated with the loop iteration, so the per-iteration gas decrease is unprovable and the `exhausted` VC (gas-derived fuel) can never close. Analysis in `docs/4ch8f-interp-strategy.md` §0.

## What ships

- **`FnHandleS`** (`SAsm/Handle.lean`): callee contract with an *entry-snapshot-parameterized* post — the `whileS` move applied to calls (auxiliary-variable triple). **`Stmt.callRegS`**: `callReg` over such handles; sp records the call's entry state; soundness adapts the `callReg` case through the existing per-state split (`cpsTripleWithin_exists_pre_M_frame`). Uniform `.pre` VC, unchanged from `callReg`. `Fn.SpecS`/`Fn.toHandleS` package vcgen-provable spec families.
- **Pilot `SAsm/InterpLoopDemo.lean`, zero sorries** — the mandatory end-to-end theorem, exercising every named composition risk:
  - **`interpFn_spec`**: a fetch-charge-select-dispatch `whileS` loop over a 3-opcode toy ISA (PUSH imm8 / ADD / STOP-or-invalid) with `callRegS` dispatch into three *real* handler `Fn`s and the value stack a grow-down rw window at `x12` (the real dispatcher's convention) **simulates the Lean-side `toyRun` trace**, with the post pinning the exit machine state to `toyRun prog (ToyState.init gas₀) cap` — the deterministic execution frozen at its halt point.
  - **Can a wrong EvmState satisfy the invariant?** No: `interpInv` names the trace *function* (`encodes (toyRun prog (ToyState.init (rf₀.get .x29).toNat) i)`), not an existential over states — and the initial gas is read from the **loop-entry snapshot's gas register**, exercising the snapshot-carried-constant channel.
  - **Does the gas variant actually bound iterations?** Yes: `toyRun_gas` ("while running, `gas + i = gas₀`") is the variant; the `exhausted` VC closes from it plus the ghost `gas₀ < cap` (static-cap idiom — wrong caps are unprovable, never unsound). Remove the loop body's charge and the proof dies at `exhausted`.
  - **Handler contract uniformity**: one `.pre` VC; `pushFn_specS`/`addFn_specS`/`stopFn_specS` are ordinary `vcgen` proofs of snapshot-parameterized functional posts; the select-branch × handle cross-product (9 cases) closes on address literals.
  - Axioms: `propext, Classical.choice, Quot.sound` (kernel-checked; no `native_decide`/`bv_decide`/`sorry`).
- **Strategy doc `docs/4ch8f-interp-strategy.md`** + design §3.11: decisions 1–5 with rejected alternatives — invariant over the existing Evm64 `EvmState`/bridge stack; gas-derived fuel with the loop-body charge as variant; one `FnHandleS` family indexed by opcode (table load = ordinary ro block); **`jalr x0` tail calls not needed** (dispatch is `jalr x1` + `ret`; non-`ret` tails get a flag+`ret` restructure in .49 rather than a multi-exit call primitive) → **bead .4 closed** with `callRegS` as the shipped remainder; frames = window movement over `phaseDView`, one flat loop, exec-log as monotone-append invariant state.
- **Decomposition**: child beads `.10.1`–`.10.4` (handle packaging, table-load spec, tail restructure, frame-window algebra — Opus/Fable split marked); `.49`/`.56` re-scoped with concrete file pointers.

Full `lake build` green (3,756 jobs); `scripts/check-forbidden-tactics.sh` OK.


pirapira:
**Review (Fable, adversarial pass on machinery + pilot): LGTM — merge.** This is the strongest deliverable in the epic so far.

## Core machinery (line-by-line audit of the trusted extension)
- `FnHandleS`/`Reach.exact`/`Fn.SpecS` are the textbook auxiliary-variable contract; `sp`/`vcs`/`steps`/`sp_mono`/`vcs_antitone`/`CalleesIn` extensions all have the right shapes (single uniform `.pre` VC, sp records the entry snapshot existentially — mirrors the `whileS` exit-sp design).
- The `callRegS` soundness case in `StmtSoundCall.lean` is a faithful mirror of the reviewed `callReg` case: peel `regOwn .x1`, fix the entry state via the existing `cpsTripleWithin_exists_pre_M_frame`, and — the load-bearing move — instantiate `h.sound rf ws A` at **exactly the state the split fixed**, so the snapshot in the post is the true entry. `asrtM (Reach.exact …)` is constructed with the right witnesses; the post maps into sp's existential precisely; step bound via `nSteps_le_foldr_max`. **No new trusted primitives** — everything reduces to `jalr_call_spec_within`, `retSpec`, and the per-state split. Clean.

## Pilot (the two adversarial questions, independently checked)
- **Wrong-state escape?** No. `encodes` pins all four registers as concrete functions of the trace state and pins the stack bytes (`ws = junk ++ stackFlat σ.stack` with `junk` length-pinned — only the *free* space is havoc'd, which is correct). The invariant names `toyRun` — the trace *function* — with initial gas read from the snapshot register, and the post freezes the exit machine state at `toyRun … cap`. There is no ∃-state anywhere a cheating proof could route through.
- **Does the gas variant bound iterations?** Yes: `toyRun_gas` (`gas + i = gas₀` while running) + the ghost `gas₀ < cap` makes a still-running state at `i = cap` arithmetically impossible — the `exhausted` VC genuinely depends on the loop body's charge.
- **Axioms re-verified myself**: `#print axioms interpFn_spec` → `[propext, Classical.choice, Quot.sound]`. Zero sorries.
- The `hsafe` hypothesis (spec-side trace safety standing in for runtime guards) is the honest pilot-scope cut, and it's correctly turned into work: the real loop's guards become separate loop exits in `.49`.

## Strategy + bookkeeping
- The §0 impossibility argument (monomorphic `FnHandle` can't correlate handler exit with loop iteration ⇒ `exhausted` unprovable) is right, and it's the same logical-variables phenomenon that motivated `whileS` — good that the fix is the same move applied to calls.
- The `.4` closure reasoning holds: dispatch is `jalr x1` + `ret`; the `.10.3` flag+ret restructure (with EEST parity gate, since it changes guest bytes) is the right trade against a multi-exit call primitive.
- Child beads `.10.1`–`.10.4` are concrete, well-fenced, and correctly marked Opus-suitable — `.10.1`/`.10.2`/`.10.4` look immediately delegable once this merges.

No conflicts expected with the open wave stack (#9735/#9736 are `Codegen/Programs`-only; this is `SAsm/*`+docs).

PR #9740, opened 2026-07-04
fix(codegen): regenerate link layout after #9729 guest growth (4ch8f.9.5)

pirapira:
## What

Bead `evm-asm-4ch8f.9.5`. PR #9729 grew the guest `.text` without regenerating the linker facts, leaving `check-region-map` RED on main and `GuestAddrs`/`symbol-addresses.tsv` stale. **Further merges since the bead was filed** (#9737 "charge first precompile value account state gas", plus the wave-A/B conversions #9735/#9736 now in main) settled the guest at **0x582a8 / 361128 B / sha `5d9cbdee…`** — past the bead's cited 0x581d0/`e576232a`. This repins all layout metadata to the *current* guest.

## Changes

- **`symbol-addresses.tsv`** — regenerated with `gen-symbol-addresses.py --build`; **59 LINK_DEPENDENT symbols moved** (the handler / system-call region after the EIP-2929 insertion point). Now matches the ELF.
- **`RegionMap.textSizeBytes`** — `0x58150 → 0x582a8` (`dataSizeBytes` unchanged). Region theorems still elaborate — sizes only, **no region moved, no disjointness proof broke**.
- **`GuestAddrs.lean`** — regenerated. **Not purely additive**: **5 referenced constants moved value** (`assemble_execution_requests`, `derive_consolidation_requests`, `derive_withdrawal_requests`, `requests_hash_verify`, `stage_system_call`); the other 54 moved guest symbols aren't referenced by any converted `_prog`. This repins the concrete verification views (incl. the wave-A/B additions generated from the stale TSV).

**No per-function rewrite was needed**: `_prog` blocks reference `GuestAddrs.*` **by name** (symbolic), so refreshing the TSV + `GuestAddrs` is sufficient and **no `#guard … _prog.length` changed** (the CLEAN consistency gate confirms — only immediates re-evaluate).

## CI follow-up

Added `scripts/check-region-map.sh` to `build.yml`, right after the existing guest-link step (reuses its binutils + linked ELF), so the next silent layout change **fails CI** instead of rotting — the exact gap that let #9729 through.

## Gates (all vs merge-base main)

1. `check-asm-to-program.sh` → **CLEAN** over the full manifest (384 defs).
2. `check-region-map.sh` → **GREEN** (`textSizeBytes` + tsv match ELF).
[...]

PR #9741, opened 2026-07-04
fix(modexp): reduce result mod modulus on the exp==0 backend path (4ch8f.11.5)

pirapira:
## Summary

Fixes the MODEXP bignum-backend `exp==0 && modulus==1` consensus divergence (bead `evm-asm-4ch8f.11.5`, found during the `.11.3` dual-path analysis).

`zkvm_modexp` initialised `result = 1` then, on `exp == 0`, jumped straight to output formatting **without reducing mod modulus** (`bnez a0, .Lmexp_format`). Any bignum-backend call (base/exp/mod length > 4) with `exp == 0, modulus == 1` therefore emitted `1`, but EIP-198 / the amsterdam spec computes `pow(base, 0, 1) = 1 % 1 = 0`. Reachable from EVM code.

The `exp == 0` branch now routes through a `.Lmexp_expzero` stanza that reduces `result` mod modulus via the existing `modexp_binmod` before formatting:
- `modulus == 1` → `1 % 1 = 0` ✅ (was `1`)
- `modulus > 1`  → `1 % m = 1` (unchanged)
- `modulus == 0` → keeps its existing `.Lmexp_modzero` zero-fill path (unchanged)

Smallest correct change; the routine is not restructured.

## Image shift (expected, not byte-identical)

Inserting instructions shifts the guest image. `.text` grew `0x58150 → 0x582d0` (+0x180). Regenerated mechanically:
- `RegionMap.textSizeBytes`
- `scripts/asm-fixtures/symbol-addresses.tsv` (`gen-symbol-addresses.py --build`)
- `EvmAsm/Codegen/GuestAddrs.lean` (`asm_to_program.py guest-addrs`) — pure address-constant diffs.

The guest `.s` diff is **only** the intended MODEXP stanza (the `bnez` retarget + the 6-instruction reduction), no unrelated churn.

## Validation

1. `lake build` green; `check-asm-to-program.sh` **CLEAN** (384 converted defs, 144 files); `check-region-map.sh` matches the linked ELF.
2. New real-linked probe `zisk_modexp_backend_real_probe` + `scripts/codegen-zisk-modexp-backend-real-probe-check.sh` (the existing safe-fail probe is unchanged and still PASSES). Four vectors, base_len=5:
   - record 0 `exp==0, mod==1 → 0` — **FAILS on pre-fix code (emits 1), PASSES after** ✅
   - record 1 `exp==0, mod==13 → 1`
   - record 2 `exp==0, mod==0 → zeros`
   - record 3 `2^5 mod 13 → 6` (happy path, unchanged)

   A/B confirmed: pre-fix `--require-ready` exits 1 (record 0 = `…01…`); post-fix exits 0 (all match).
3. No wired EEST fixture reaches the bignum backend corner — MODEXP is unconverted (no `_prog`/replay fixture; the exp-frontier check explicitly excludes the MODEXP precompile). Stated explicitly per gate 3.
4. `check-forbidden-tactics.sh` / `check-file-size.sh` / `check-unimported.sh` / `check-region-map.sh` all OK. No `native_decide`/`bv_decide`/`sorry`; aligned accesses only.
5. Whole-guest byte-identity intentionally does not hold (the point of the fix); reported above.

Doc: `docs/4ch8f-modexp-dual-path-analysis.md` gets the "FIXED" note.


pirapira:
**Review (Fable): LGTM on the fix itself — one sync defect found & pushed, plus a merge-order requirement vs #9740.**

**Verified:**
- **The stanza is correct**: `.Lmexp_expzero` reduces `result` via `modexp_binmod(result, s7, mod, s7) → result` then falls through to format — same calling convention as the existing base-reduction site; `modulus == 0` short-circuits earlier (`.Lmexp_modzero`) so it never reaches the new path; `modulus > 1` leaves `1 % m = 1` unchanged. Spec corner re-derived independently during the #9732 review: `pow(b, 0, 1) = 0`.
- **Probe re-run on the branch**: all 4 regression vectors PASS (`exp==0,mod==1 → 0`; `mod==13 → 1`; `mod==0 → zeros`; `2^5 mod 13 → 6`). The real-linked probe closing the "no probe links the real backend" gap from the .11.3 analysis is a real win on its own.
- `check-region-map` GREEN on the branch.

**Defect found & fixed (pushed to this branch):** the committed `GuestAddrs.lean` was stale by a uniform −8 vs the branch's own `symbol-addresses.tsv` (regenerated at different times around a mid-work main merge), so `check-asm-to-program` was RED despite the PR body's CLEAN claim — always re-run the drift gates *after* the last merge/regen, not before. Resynced; `check-all` now CLEAN (384 defs / 144 files), full build green.

**Merge-order requirement:** this PR and #9740 both rewrite `GuestAddrs.lean`/`symbol-addresses.tsv`/`RegionMap.textSizeBytes` from different baselines — they WILL conflict, and textual resolution would be wrong for either side. Plan: **merge #9740 first** (metadata-only, fixes the CI gate), then this branch merges main and resolves by *re-running the regen* (`gen-symbol-addresses.py --build` + `guest-addrs` + repin `textSizeBytes`) — the regen is authoritative, never a textual merge. Happy to do that refresh when #9740 lands if the author session is busy.

PR #9768, opened 2026-07-04
proof(frames): per-depth window algebra over phaseDView (4ch8f.10.4)

pirapira:
Bead `evm-asm-4ch8f.10.4`. Pure separation-logic algebra for the Phase-D `call_frame_arena`, feeding bead `.56`'s CALL/CREATE/RETURN descend/return contracts (strategy `docs/4ch8f-interp-strategy.md` §4). Lean-only, no machine-step proofs, no CALL/RETURN semantics (that is `.56`).

## What lands

New module `EvmAsm/Codegen/CallFrameWindows.lean`, built on `SAsm/PhaseSplit.lean` + `Codegen/CallFrameLayout.lean` + `Codegen/CallFramePhase.lean`, everything parameterized over the arena `base : Word` (no `callFrameArenaBase` baked in; `GuestAddrs`/`RegionMap` untouched).

1. **Equal-stride tiling** — `phaseDView base` is the 1025 per-depth `frameWindow base d = anyBytes (base + ofNat (d*frameStride)) frameStride` tiles. Stated the `anyTilesAt` way over `List.replicate frameSlotCount frameStride` (`phaseDView_eq_framesTiling`), proved by induction on the generic replicated segment via `anyTilesAt_replicate_focus_at` — **never a 1025-way enumeration** (compose-on-`j` discipline).
2. **Focus/unfocus one depth** — `phaseDView_focus` (`= prefixFrames base d ** frameWindow base d ** suffixFrames base d`), with `focusFrame` (extract depth `d`) and `unfocusFrame` (`bytesRegion → anyBytes` re-absorb at depth `d`); the other depths stay havoc'd.
3. **Intra-frame carving** — `frameWindow_components` tiles one window into named sub-region accessors at the `CallFrameLayout` offsets: `frameMemWindow`, `frameStackWindow` (incl. both guards), `frameReturndataWindow`, `frameEnvWindow`, `framePcWindow`, `frameCodebaseWindow`, `frameMetaWindow`, `framePadWindow`. `.56` states pre/posts directly in these.
4. **`encodesFrame` relation SHAPE** — `FrameState { stack : List Word, pc codebase : Word }` (minimal; `.56` extends it). `encodesFrame base d fr` pins the saved pc/codebase dwords and the stack window through `bytesRegion` (no ∃-state escapes), havocing the rest. `encodesFrame_focus` weakens an encoded frame into a `frameWindow`; `frameWindows_separate` is the distinct-depth disjointness sanity lemma.

## Gates

- **Zero sorries.** `lake build` green.
- `#print axioms` on `phaseDView_eq_framesTiling`, `phaseDView_focus`, `focusFrame`, `unfocusFrame`, `frameWindow_components`, `encodesFrame_focus`, `frameWindows_separate` = `[propext, Classical.choice, Quot.sound]` (verified locally; CI's axiom gate covers only registered witnesses).
- No `native_decide`/`bv_decide`; no heartbeat/recDepth raising. The 1025-tiling is a `List.replicate` instance proved by induction over the segment, not concrete enumeration.
- Lean-only PR: no emitted strings, no `_prog`s, no fixtures.

### DIFF
--- a/EvmAsm/Codegen.lean
+++ b/EvmAsm/Codegen.lean
@@ -12,6 +12,7 @@ import EvmAsm.Codegen.Layout
 import EvmAsm.Codegen.Programs
 import EvmAsm.Codegen.RegionMap
 import EvmAsm.Codegen.CallFramePhase
+import EvmAsm.Codegen.CallFrameWindows
 import EvmAsm.Codegen.Driver
 import EvmAsm.Codegen.Cli
 import EvmAsm.Codegen.RoundTripTests
--- a/EvmAsm/Codegen/CallFrameWindows.lean
+++ b/EvmAsm/Codegen/CallFrameWindows.lean
@@ -1,0 +1,343 @@
+/-
+  EvmAsm.Codegen.CallFrameWindows
+
+  Per-depth window algebra over the Phase-D `call_frame_arena`
+  (bead `evm-asm-4ch8f.10.4`, strategy `docs/4ch8f-interp-strategy.md` §4).
+
+  ## What this module is
+
+  The Phase-D dispatcher is ONE flat loop; a CALL/CREATE does not recurse but
+  MOVES a window over the 1025-slot frame arena
+  (`CallFramePhase.phaseDView base = anyBytes base frameArrayBytes`).  Frame
+  `d` lives at `base + d * frameStride` (`CallFrameLayout`).  This module is
+  the pure separation-logic algebra those descend/return proofs (bead `.56`)
+  splice: it carves `phaseDView` into per-depth windows, focuses/unfocuses one
+  depth against the untouched rest, tiles a single window into its named
+  sub-regions (`CallFrameLayout` offsets), and fixes the SHAPE of the
+  depth-indexed `encodesFrame` relation for suspended parent frames.
+
+  No machine code, no addresses pinned: everything is parameterized over the
+  arena base `base : Word`, exactly like `CallFramePhase`.  The 1025-way tiling
+  is stated once over `List.replicate frameSlotCount frameStride` and proved by
+  induction over the generic replicated segment — never a 1025-case
+  enumeration (see `anyTilesAt_replicate_focus_at`).
+
+  Built on `EvmAsm/Rv64/SAsm/PhaseSplit.lean` (`anyBytes`, `anyBytes_add`,
+  `anyTilesAt`, `anyBytes_eq_anyTiles`, `bytesRegion_anyBytes`) and the audited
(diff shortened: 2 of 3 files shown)


pirapira:
**Review (Fable, substrate author — PhaseSplit/CallFramePhase): LGTM — merge.** Audited the whole module line-by-line; the algebra is correct and exactly the shape `.56` needs.

**Verified:**
- **Generic-`j` discipline held**: `anyTilesAt_replicate_focus_at` is composed over the replicated segment (split at `j`, peel one, offset arithmetic threaded via `sum_replicate`) — no 1025-way enumeration anywhere, and elaboration is instant (1.3 s module build).
- **Offset correctness is kernel-checked, not asserted**: `frameWindow_components`' closing `simp` normalizes both the accumulated `anyTilesAt` offsets and the `CallFrameLayout` accessor offsets to numerals and proves them EQUAL — so the named windows provably sit at the layout's offsets (including the mem+guard+stack+guard = `frameReturndataOff` chain and the pad at `frameUsedBytes`).
- **`encodesFrame` has no ∃-state escapes**: pc/codebase/stack pinned through `bytesRegion` with length lemmas; only mem/returndata/env/meta/pad havoc'd. `encodesFrame_focus`' mono chain and `unfocusFrame`'s `bytesRegion → anyBytes` re-absorb are the right one-directional weakenings.
- **`frameWindows_separate`**: the double-focus offset arithmetic (`(d+1)·s + (d'-(d+1))·s = d'·s` etc.) is right.
- **Axioms re-verified myself**: all six named theorems = `[propext, Classical.choice, Quot.sound]`.

**Two non-blocking notes for `.56` (worth carrying into its bead description):**
1. `stackWindowBytes` (explicitly a placeholder) lays the stack out at the START of the window and `take`-truncates. The real dispatcher convention — and the `.10` pilot's `encodes` — is a **grow-down** stack (free space first, `junk ++ stackFlat`, top at the moving `x12` pointer). `.56`'s real encoding must (a) switch to grow-down, and (b) carry a `stack.length ≤ capacity` invariant — the current `take` makes an over-deep stack pin WRONG bytes rather than be unsatisfiable, which would be a soundness trap if it survived into the real relation.
2. `encodesFrame` pins the stack bytes across the ENTIRE stack window **including both guard bands**; the real relation will likely want the guards as separate (pattern-pinned or havoc'd) sub-regions so guard-check proofs don't have to dig them out of the stack serialization.

Both are squarely inside `.56`'s declared scope ("Bead .56 fixes the concrete stack encoding") — the SHAPE landing here is right.

PR #9771, opened 2026-07-04
proof(dispatch): package clean-ret handler specs as FnHandleS handles (4ch8f.10.1)

pirapira:
Bead **evm-asm-4ch8f.10.1**: package the verified clean-ret arithmetic/logic handler specs (`Codegen/Proofs/HandlerSpecs.lean`) as snapshot-parameterized dispatch handles (`FnHandleS`, per `docs/4ch8f-interp-strategy.md` §3), consumed by the interpreter dispatch loop (`.callRegS`, bead `.49`). Built on `FnHandleS`/`Stmt.callRegS` from #9738 (now merged); targets `main`.

## What's delivered

13 clean-ret handles, each `(base sp : Word) → FnHandleS`, fully **base-parameterized** (no `GuestAddrs` pins — safe against #9740 / `.9.5` layout regen):

| File | Handles |
|---|---|
| `HandlerHandles.lean` | `evmAddHandle` (+ shared bridges) |
| `HandlerHandlesBinary.lean` | SUB, LT, GT, SLT, SGT |
| `HandlerHandlesLogic.lean` | EQ, AND, OR, XOR |
| `HandlerHandlesUnary.lean` | ISZERO, NOT, POP |

## Design (the adapter is the real work)

- **Minimal value-stack window**: `rw := ⟨sp, 64⟩` (binary), `⟨sp, 32⟩` (unary), `RwRegion.empty` (POP) — operands at fixed window offsets, no junk framing. `.49`/`.56a` embed into the full arena via the existing `FnHandle.widenRw`.
- **`pre`** = §3 uniform shape: value-stack window pointer `x12 = sp` (operand-count guard discharged by the fixed window).
- **`postS rf₀ ws₀ A₀`** pins the exit registers/window as **functions of the entry snapshot** — no ∃-state escapes. Unused peeled scratch registers ride the frame unchanged (framed-untouched, not existentialized).
- The adapter reuses each `HandlerSpecs` `cpsTripleWithin` **verbatim** (arithmetic not re-derived): bridge raw `↦ₘ` operand cells to the window `bytesRegion` (`bytesRegion_eq_8cells`/`_4cells`), peel touched registers off `regFileIs` (`regFileIs_split_bin`), frame the remainder, apply the spec, re-fold. Mirrors the documented template `Rv64/SAsm/ExamplesVc.handAdd_sound`.

## Gates

- `lake build EvmAsm.Codegen.Proofs` green under the **default** heartbeat limit — no `set_option maxHeartbeats` anywhere. Each `<op>PostS` is `[irreducible]` so the 20+-`let` result bundle stays folded during `isDefEq` (this, not a heartbeat raise, is the fix the bead prescribes).
- `scripts/check-forbidden-tactics.sh`: OK (no `native_decide`/`bv_decide` introduced).
- **Axioms** (CI's axiom gate covers registered `Progress.lean` witnesses only; stated here per the bead) — every `<op>Handle` / `<op>Handle_sound` is exactly `[propext, Classical.choice, Quot.sound]`. Verified e.g.:
  - `#print axioms evmAddHandle` → `[propext, Classical.choice, Quot.sound]`
  - `#print axioms evmSgtHandle_sound` → `[propext, Classical.choice, Quot.sound]`
  - `#print axioms evmPopHandle_sound` → `[propext, Classical.choice, Quot.sound]`
- Lean-proof-only: no emitted string / `_prog` touched; no probe/byte gates needed.

## Reviewer notes (Fable)

- No new trusted machinery: the adapter is ordinary separation-logic plumbing over existing lemmas (`cpsTripleWithin_exists_pre_M_frame`, `regFileIs_eq_atoms`, `bytesRegion_eq_cons`, `FnHandle.widenRw`). The only *new* general lemma is `bytesRegion_eq_4cells` (a 4-cell shrink of the existing `bytesRegion_eq_8cells` pattern).
- Each `postS` pins the exit state to the entry snapshot — please spot-check `evmAddPostS` (functional carry chain), `evmSltPostS` (`if a3=b3` branches, still functional), and the AND/OR/XOR posts (x5/x11 framed-untouched, deliberately unconstrained).

## Scope

Only the clean-ret arithmetic/logic family. Skipped handlers noted per owning bead in the §3 "shipped" appendix (PUSH/DUP/SWAP → `.50`; MUL/SIGNEXTEND/BYTE/shifts/div → `.51`; memory/env → `.52`; halts/control non-`ret` tails → `.10.3`/`.55`; CALL-family → `.56`).


pirapira:
**Review (Fable, adversarial audit of the adapter): LGTM — merge.**

Audited `HandlerHandles.lean` (the ADD worked template + shared bridges) line-by-line and spot-audited the POP/ISZERO structural variants:

- **No new trusted machinery, as claimed**: the adapter is `bytesRegion_eq_cons`-chains (`_eq_8cells`/`_eq_4cells`), `regFileIs`→atoms splitting, `cpsTripleWithin_frameR`/`_weaken`/`_exists_pre_M_frame`, and `xperm` — all existing, reviewed lemmas. The HandlerSpecs triples are consumed verbatim (arithmetic not re-derived), which is exactly the right trust structure.
- **`postS` has no ∃-state escapes**: every exit register (including the carry-chain scratch x5/x6/x7/x11) and the FULL exit window — including the dead bytes below the new stack top, pinned to their old contents — are functions of the entry snapshot. `A = A₀` pins the ambient. The explicit `rf'` construction + `regFileOn_congr` framing means nothing rides on an existential.
- **The stack-discipline details are right**: binary ops move `x12 → sp+32` (grow-down pop-one), unary ops leave `x12` unchanged (top replaced in place), POP bumps `x12` with `RwRegion.empty` (no bytes touched) — each matching EVM semantics and the pilot's conventions.
- **The `[irreducible] <op>PostS` discipline** is the correct fix for the let-bundle unification blowup (project rule: never raise heartbeats) — and it's applied uniformly.
- **Axioms re-verified myself** across all four files (`evmAddHandle`, `evmSgtHandle_sound`, `evmXorHandle_sound`, `evmPopHandle_sound`, `evmIsZeroHandle_sound`): all `[propext, Classical.choice, Quot.sound]`. Full build green on the branch.

One forward-looking note for `.49` (no change requested here): the handles' minimal windows are anchored at `sp` = the operand base, with the §3 pre `x12 = sp` — so the dispatch site's `.pre` VC will need the loop invariant to supply the window at the CURRENT `x12`, i.e. the `widenRw` embedding must be per-iteration (the strategy doc anticipates this via snapshot-carried window bases; flagging so `.49` budgets for it).

PR #9770, opened 2026-07-04
proof(dispatch): drift-guarded opcode-table load spec (4ch8f.10.2)

pirapira:
Bead `evm-asm-4ch8f.10.2`. Bottom-up load spec for the dispatcher's `opcode_handlers` / `opcode_gas_costs` `.data` tables, feeding the `.49` `callRegS` `.pre` VC.

**Based on #9740** (`fix/asm-to-program-layout-regen`) — retarget to `main` once #9740 merges.

## Table shape (confirmed, no STOP)
Both tables are plain **256 × 8-byte `.dword` arrays** (`emitJumpTable` / `emitGasCostTable`, Dispatch.lean): `opcode_handlers` = link-resolved handler-label addresses (default `h_invalid`), `opcode_gas_costs` = numeric `staticGasCost b`. Shipped registry = `callFrameGuestRegistry`.

## What's here (all in new `EvmAsm/Codegen/Proofs/OpcodeTables.lean`)
1. **`exec_table_load`** — generic, address-free: the `slli/add/ld` block over any `Region` pinned by `tableAt region tOff entries` loads `entries[op]`. Generic over `region`/`tOff`/`entries`/`op < entries.length`. (`execInstrRF`-level, like the pilot's `exec_fetch`.)
2. **Mirrors** — `opcodeGasCostEntries : List Word` (from `staticGasCost`) and `opcodeHandlerLabels` (256, from `jumpTargetLabel callFrameGuestRegistry`). `#guard` both = 256.
3. **`handler_table_load_witness`** — the `.pre` bridge: `(execBlock … tableLoadBlock).1.get dstR = addrOf (opcodeHandlerLabels[op])`, i.e. `= opcodeHandlerEntries[op]` (`opcodeHandlerEntries_get`). **No ∃** — a `.10.1` handle whose `entry` equals this plugs in by list-lookup.

### Design note (adaptation)
The handler labels (`h_ADD`…`h_invalid`) are emitted by the raw jump-table string, **not** a converted `_prog`, so they are **not** `GuestAddrs.*` constants (and the fences forbid editing `GuestAddrs`). So the addressed image is parameterized by `addrOf : String → Word` rather than baked from `GuestAddrs` by name — this stays regen-proof (labels come from `jumpTargetLabel`, the emitter's own source) and uses **no numeric literals**. `.49` supplies `addrOf` = the `.10.1` handle entry map.

## Drift guard
[...]

PR #9772, opened 2026-07-04
feat(crypto): 4ch8f.11 crypto-kernel verification strategy + accelerator-ladder pilot

pirapira:
Closes bead `evm-asm-4ch8f.11` (the last open P0 planning bead of the epic): the per-kernel verification-strategy decision for the software crypto kernels, plus the mandatory pilot proving the riskiest shared piece end-to-end.

## Strategy — `docs/4ch8f-crypto-strategy.md`

1. **Tier assignment**: every kernel ends at **full functional verification**; trusted-kernel-with-differential-obligation is rejected as a *terminal* tier (bead `.11.5` — the MODEXP `exp==0 ∧ modulus==1` divergence that months of differential probing missed — is the standing counterexample), and verified-glue is only the per-kernel intermediate milestone. Sequencing by consumer value: secp256k1 field stack (`.38`, unblocks tx-sender verification) first; pairings/maps/KZG last — their RFC 9380 / KZG-"unreachable" gaps are **completeness-only** under the `.8` soundness headline.
2. **Field-arithmetic library**: spec vocabulary is **Nat-modular arithmetic** matching `Accel.*` (no `ZMod` at interfaces — rejected for the coercion tax at every seam crossing; mathlib number theory confined to one planned `Crypto/Fermat.lean`). Library layout and the wave plan are in the doc's §2 table; the first three files land in this PR.
3. **Canonical glue shape**: accelerator wrappers are **hand-proven snapshot-parameterized handles** — `FnHandleS.sound` is a machine-level `cpsTripleWithin` built from `step_csrs` + the bead-`.1` concrete `csrsWrite`/`csrsValid` semantics — consumed by SAsm kernels via `Stmt.callRegS`. The SAsm block engine is **never** extended with `CSRS` (rejected: trusted-core surgery; also rejected: monomorphic handles, a `callS` primitive — see doc §3).
4. **Spec side**: algorithm-faithful Lean ports of execution-specs (`tests-zkevm@v0.4.0`) where pure Python; **ECDSA needs project-side references** (execution-specs delegates `secp256k1_recover`/`secp256r1_verify` to native coincurve/cryptography). Heavy abstract math is out of scope; number theory enters only at five named kernel↔spec divergences (doc §4).

## Pilot — zero sorries, kernel-checked

`#print axioms` (verified this session):
- `EvmAsm.Rv64.SAsm.PowLadderDemo.powFn_spec` : `[propext, Classical.choice, Quot.sound]`
- `EvmAsm.Rv64.SAsm.csrs_arith256Mod_spec_within` : `[propext, Classical.choice, Quot.sound]`
- `EvmAsm.Crypto.ladder_correct` : `[propext, Quot.sound]`

Per claim, the exact theorem:
- **The first machine-level CSRS triple**: `csrs_arith256Mod_spec_within` (`SAsm/AccelStep.lean`) — one `csrs 0x802, rs1` step with the `{a,b,c,module,d}` param block and operands inside a `bytesRegion` window: output buffer becomes `(a·b + c) mod m`, nothing else moves; all six `csrsValid` dword-range obligations + modulus ≠ 0 discharged from window well-formedness. Packaged as `arith256ModHandle : FnHandleS` via `csrs_arith256Mod_ret_spec` (composition with `jalr_ret_spec`).
- **The ladder is really `x^e mod m`**: `PowLadderDemo.powFn_spec` post is `wsNat256 ws 176 = x ^ Crypto.beBytesToNat ebytes % m` — `Nat.pow`, not a ladder-shaped restatement; the fold bridge `Crypto.ladder_correct` is guarded by kernel-`decide` KATs (e.g. `ladder 1009 7 [0x01, 0x23] 16 = 7 ^ 0x123 % 1009`) that a wrong ladder (LSB-first, skipped squarings, off-by-one) fails.
- **The cap-VC closes for the 4569-bit exponents**: fuel is `8 * ebytes.length`, *symbolic* up to 4096 bytes; the `exhausted` VC closes by counter arithmetic for every width at once — no per-width proof, no gas coupling for constant-trip ladders.
- **Aliasing exercised**: the square step runs the accelerator with `a = b = d = acc` through param block 0; decode-valued posts make alias side conditions unnecessary.
- **`1 < m` is load-bearing**: at `m = 1` the staged unreduced `acc₀ = 1` breaks the invariant at entry — exactly the `.11.5` MODEXP corner; the strategy requires ladder consumers to gate or special-case it.

Pilot simplifications, each mapped to a bead in doc §5/§6: demo-local addresses (no `GuestAddrs` dependence — `.9.5`-safe, the `.10` pilot convention), 4-limb only (`.11.6` makes it `nLimbs`-generic), LE operands staged directly (`.11.7` BE↔LE), Lean-level ro exponent (right for Fermat/final-exp constants; MODEXP's runtime exponent uses `whileS`).

## Beads

`.11` **closed**; new children `.11.6` (nLimbs/curve/complex handles — Fable), `.11.7` (BE↔LE — Opus), `.11.8` (Fermat/QR — the one mathlib-heavy file), `.11.9` (scalar-mul skeleton — Fable); `.38`/`.57`/`.58` re-scoped with per-sub-bead Opus/Fable split in their notes.

## Gates

- `lake build` clean (new modules registered in `EvmAsm/Rv64/SAsm.lean`)
- `scripts/check-forbidden-tactics.sh` OK (no `native_decide`/`bv_decide`; one scoped `maxRecDepth` bump for a concrete 208-byte `decide`, same pattern as the existing KATs)
- fences respected: no `GuestAddrs`/`RegionMap`/`Programs` immediates, no `ModexpBackend.lean`, no `Stateless/EntrySpec.lean` changes


pirapira:
**Review (Fable, hard-area audit): LGTM — merge.** With this, all 11 P0 planning beads of the epic are closed. Audit results:

## `AccelStep.lean` (the semantics-adjacent new machinery — deepest scrutiny)
- **`csrs_arith256Mod_spec_within` is honestly stated**: window well-formedness, all six aligned+fitting offsets, param-block slots pinned via `wsDword`, `m ≠ 0` — every `csrsValid` obligation is DISCHARGED from these (I checked the six-conjunct assembly), never assumed. The `csrsValid`/`csrsWrite` arms are cited **definitionally** (`rfl`) against the bead-.1 semantics, so the triple cannot drift from the model.
- **The post pins everything**: `bytesRegion B (setBytes ws dOff …)` — the whole window with exactly one splice, registers `regFileIs rf` unchanged. "Nothing else moves" is literal.
- **Decode-valued aliasing** is the right call: `a = b = d = acc` (the square step) needs no side conditions because only entry-decoded values appear in the post. This is the pattern every other accelerator id should copy.

## The pilot
- **`powFn.post` is genuine `Nat.pow`**: `wsNat256 ws 176 = x ^ beBytesToNat ebytes % m` — and the bridge `ladder_correct` is PROVEN by induction (`ladder_inv`), not decided; its axioms are `[propext, Quot.sound]` (no choice — nice). The invariant names the fold function, so no ladder-shaped restatement smuggling.
- **`1 < m` load-bearing** matches my independent understanding of the `.11.5` corner — and the strategy turning it into a consumer gate obligation is exactly right.
- **Axioms re-verified myself**: `powFn_spec` and `csrs_arith256Mod_spec_within` = classical-3; `ladder_correct` = `[propext, Quot.sound]`. Full build green on the branch.

## Strategy decisions — concur on all four
- Full-functional as the only terminal tier, with `.11.5` as the standing refutation of differential-only: this is the argument I would have made.
- Nat-vocabulary at interfaces (no `ZMod` coercion tax), mathlib confined to `Crypto/Fermat.lean`: right trade; revisit only if the Fermat/group-law wave finds itself re-proving substantial `ZMod` material by hand.
- **Never extending the SAsm block engine with `CSRS`** and going through `FnHandleS` wrappers instead: the most important call in the PR — it keeps the trusted core untouched and makes every accelerator contract an ordinary reviewable theorem. Note the wrapper composes cleanly with #9770's table-load witness and #9771's handle family — the three PRs' interfaces line up.
- ECDSA needing project-side reference specs (execution-specs delegates to native code) is correctly flagged as the one place the spec side itself needs building — that's the riskiest open item in the plan and deserves its own early bead.

No conflicts expected with the open stack (#9740 is build.yml-only now; #9770/#9771 are `Codegen/Proofs/*`).

PR #9774, opened 2026-07-04
port(4ch8f.12): verify swd_read_u64le + sg_load_u32le (SAsm byte-wise readers)

pirapira:
First batch of the byte/copy leaf-helper ports (bead **evm-asm-4ch8f.12**), the two genuinely straight-line readers — gets the review loop going early. Both pass `scripts/port-check.sh` (build, **zero warnings**, no forbidden tactics, **3 classical axioms only**).

## Beads
- **.12.6 `swd_read_u64le`** — `EvmAsm/Codegen/Programs/SwdReadU64leSAsm.lean`, theorem **`swdReadU64leFn_spec`**. Post: `a0 = leU64 bs 0` (little-endian u64 over the byte region at the `a0` pointer). Byte-identity to the emitted routine is **kernel-pinned**: `#guard swdReadU64leBody.flatten 0 ++ [Instr.JALR .x0 .x1 0] = swdReadU64le_prog`.
- **.12.3 `sg_load_u32le`** — `EvmAsm/Codegen/Programs/SgLoadU32leSAsm.lean`, theorem **`sgLoadU32leFn_spec`**. Post: `a0 = leU32 bs 0`. String-emitted inline in `StatelessGuestEpilogue.lean` (no `_prog`), so length (`= 11`) + position-independence (`flatten 0 = flatten 0x80000000`) are pinned; asm source is quoted in the header.

## Why byte-wise (LBU), not a wide load
Both callers read **unaligned** SSZ fields (offset table base `0x40000012`; `block_number`/`timestamp` at payload `+404`/`+428`). A single `LWU`/`LD` there fails the RV64 model's alignment gate and traps (bead 4ch8f.7). Assembling from `LBU`s is legal at any address — the spec is alignment-agnostic (`leU32`/`leU64` over `bs.getD`).

## Notes for review
- Model: SAsm `Fn` over a byte `Region` (ChainIdSAsm / `sszReadOffsetFn`), **not** the `cpsTripleWithin`+dword-atom route (`lbu_spec_gen_within` uses `extractByte` over dword atoms — awkward for unaligned spans).
- The heavy `execBlock` reduction is factored into a `private _engine` lemma so it has its own heartbeat budget; `vcgen`'s two VCs otherwise share one 200k budget and overflow. No `maxHeartbeats`/`maxRecDepth` bumps.
- **Spec-only** (no emitted-code change) → no EEST A/B.


pirapira:
**Review: APPROVE** (Lean-proof / spec-only port — review-playbook §1 "Lean-proof PR").

### Verified independently
- **Axioms** — `#print axioms` on both headline theorems (fully qualified): `EvmAsm.Codegen.SwdReadU64leSAsm.swdReadU64leFn_spec` and `EvmAsm.Codegen.SgLoadU32leSAsm.sgLoadU32leFn_spec` → both `[propext, Classical.choice, Quot.sound]`. `port-check.sh` axiom audit also clean over all module theorems (40 / 28).
- **port-check.sh** — PASS on both files: builds, warning-free re-elaboration, no forbidden tactics/TCB bumps, classical-3 axioms, `#guard` pins present.
- **Adversarial statement read** (§2 checklist):
  - No ∃-escape — both posts are `rf.get .x10 = leU32/leU64 bs 0`, a pure function of the input region (§2.1 ✓).
  - `leU32`/`leU64` are correct little-endian (`b0 | b1<<8 | b2<<16 | b3<<24 [| b4<<32 …]`) ✓.
  - Pre is **not** vacuous and **not** over-constrained: `4 ≤ bs.length` / `8 ≤ bs.length` is load-bearing — it's exactly what discharges the `Region.loadOk` offset bounds via `omega` in the `.mem` VC. Alignment obligations use `1 ∣ _` (LBU, byte access), consistent with the misaligned-access invariant (bead .7).
- **Correspondence to emitted code:**
  - `swd_read_u64le`: **machine-checked** end-to-end. `#guard swdReadU64leBody.flatten 0 ++ [ret] = swdReadU64le_prog` elaborates (build-gated), and `swdReadU64leFunction := "swd_read_u64le:\n" ++ emitProgram swdReadU64le_prog` (SystemWrites.lean) — so verified body → `_prog` → emitted string is a real kernel chain. Routine is genuinely emitted (`jal ra, swd_read_u64le` call sites in SystemWrites). ✓
  - `sg_load_u32le`: emitted string is a **hand-written literal** (StatelessGuestEpilogue.lean:861–867), not `emitProgram sgLoadU32le_verified`. I diffed it against `sgLoadU32leInstrs` by hand — it matches exactly (t0=x5, t1=x6, 11 instrs + ret). But this is **eyeball-only**; there's no kernel pin tying the verified triple to the emitted text (only length=11 + position-independence are pinned).

### Non-blocking note (routed to bead 4ch8f.12)
Before `sgLoadU32le_verified` is ever wired into emitted code, its correspondence to the emitted string must become machine-checked (mirror the `swd` pattern: add an `sgLoadU32le_prog` and emit via `emitProgram`, or add a `#guard` against the emitted-line list). As a spec-only PR this is not a soundness gap today, but the eyeball link should not survive into a composition bead.

Nothing else to flag. LGTM.

PR #9776, opened 2026-07-04
fix(eest): expand child frame memory

pirapira:
## Summary
- expand child call-frame EVM memory from 64 KiB to 128 KiB, matching the root runtime memory capacity
- move child frame stack/env offsets and frame stride together
- update call-frame layout docs and probe expectations for the new offsets

## Validation
- lake build EvmAsm.Codegen.Programs.CallFrameBase EvmAsm.Codegen.Programs.CallFrameDescend EvmAsm.Codegen.Programs.CallFrameReturn EvmAsm.Codegen.Dispatch
- scripts/codegen-eest-stateless-check.sh --filter stRandom/random_statetest384 --limit 1 --backend spike --jobs 1 --max-jobs 10 --run-dir gen-out/eest-run/codex-random384-frame128-spike --show-passes

Note: random384 still fails with bv_fail=53 after this slice. Temporary diagnostics showed the materialized log count remains 1, so this PR is a layout/capacity fix so far, not the final random384 receipt fix.


pirapira:
Follow-up pushed after the original PR filing: aligned the runtime memory-expansion guard constants in EvmMemoryGas with the 128 KiB root/child frame arena, which fixes the random_statetest384 DELEGATECALL input range OOG after the frame expansion.\n\nValidation run locally:\n- lake build EvmAsm.Codegen.Programs.EvmMemoryGas\n- scripts/codegen-eest-stateless-check.sh --filter stRandom/random_statetest384 --limit 1 --backend spike --jobs 1 --max-jobs 10 --run-dir gen-out/eest-run/codex-random384-memlimit-spike --show-passes\n- scripts/codegen-eest-stateless-check.sh --filter random_statetest415 --limit 1 --backend spike --jobs 1 --max-jobs 10 --run-dir gen-out/eest-run/codex-random415-memlimit-regress-spike --show-passes\n\nAlso merged current origin/main into the branch after the follow-up commit.

PR #9785, opened 2026-07-04
feat(guest-image): 4ch8f.63 structural half — image CodeReq, GuestFraming bundle, verdict-stamp audit

pirapira:
Bead `evm-asm-4ch8f.63`, **structural half** — the three artifacts `.64` cannot start without. The bead stays OPEN for the pipeline-triple half (completion note appended).

## 1. The guest-image `CodeReq`

- `EvmAsm/Codegen/Proofs/GuestImage.lean`: `guestImageCodeReq = CodeReq.ofEntries guestImageEntries` — the 358 **linked** wave-.9 conversions anchored BY NAME at `GuestAddrs.<entry>` (26 of the 384 manifest conversions are converted-but-unlinked and correctly excluded; the image CodeReq reflects the emitted ELF).
- `EvmAsm/Codegen/Proofs/GuestImageEntries.lean` is GENERATED (`scripts/guest_image_coverage.py --emit-lean`); `asm_to_program.py guest-addrs` now also emits entry constants for reloc-free linked functions (+29 symbols, `check-guest-addrs` CLEAN).
- **Compose-don't-enumerate**: whole-image disjointness is ONE kernel `decide` (`guestImageEntries_extentsOk` — ascending non-overlapping extents inside `.text`); the new generic `EvmAsm/Rv64/CodeReqExtents.lean` turns it into the monotonicity witness (`ofProg_sub_ofEntries_of_extentsOk` → `guestImage_block_sub`) that `cpsTripleWithin_extend_code`/`cpsHaltTripleWithin_extend_code` consume. No 358-case proof anywhere.
- **Coverage accounting** (`docs/4ch8f-guest-image-coverage.md`, regenerable): covered 86,872 B = **23.76%** of `.text` (365,592 B); **443 uncovered ranges** (script asserts covered + gaps = textSizeBytes exactly). Gap clusters filed as child beads **4ch8f.63.2–.63.12** (EVM handlers ~101 KB, block verdict ~55 KB, BAL ~22 KB, crypto/modexp/bridges ~42 KB, `_start`+shell, long tail).

## 2. `GuestFraming` bundle + `scratch_sat`

- `guestFraming : GuestFraming` — `scratch = residue =` the `**` of `anyBytes` over the six `zone = .ram` regions of `RegionMap.guestRegionMap` (`zisk_system ** OUTPUT ** guest_stack ** state_tracker_live ** .data ** .sszscratch`), drift-pinned to the map by `guestScratch_matches_regionMap` (`decide`).
- `guestScratch_sat`: explicit heap construction via the new `EvmAsm/Rv64/MemSat.lean` — `SatWithin P lo hi` (satisfiable by a heap owning only mem dwords in `[lo,hi)`) with atoms for `memIs`/`bytesRegion`/`anyBytes` and a `sepConj` chaining combinator over ascending footprints. `RegionMap.guestRegionMap_pairwise_disjoint`'s layout is what makes the chain go through.

### ⚠ Statement repair (the promised "framing fights the statement" signal)

`scratch_sat` was **unprovable as stated**: `↦ₘ` bakes in `isValidDwordAccess`, whose valid map ends at `MEM_END = 0x78000000`, so a `2^30`-byte input at `INPUT_ADDR+16` has payload dwords outside every valid zone → `guestInputAssertion` unsatisfiable → the ∀-input witness impossible ( `2^30` also overhung `.text` by 16 bytes). Fixed at the sharp bound: **`MAX_INPUT_BYTES := 0x37FFFFF8`** (~896 MiB), with a `#guard` pinning it to `MEM_END` so a model change re-derives it. Full decision record: `docs/4ch8f-top-spec.md` **§2a** (rejected alternative: widening `MEM_END`, which would relax every proof in the repo for input sizes no host exercises).

## 3. Verdict-stamp soundness audit → `docs/4ch8f-top-spec.md` §6a

- The `sb a0, 32(t0)` stamp (`StatelessGuestEpilogue.lean:842`) is the **single, final** byte-32 writer, gated on `stateless_verdict_v2 = 1` ∧ schema-id `0x0001` ∧ canonical outer offsets; `.Lsg_all_pass` deliberately writes nothing.
- Encoder byte-32 is always 0 (`decode_validation_bit` sets `x11 := 0` unconditionally — the Entry.lean PR6 comment claiming otherwise is stale); `.Lsg_hash`'s only OUTPUT store is the SHA-256 root at `[0..32)`; reason codes (incl. `REASON_PRECOMPILE = 0x01`) are only ever stored at `OUTPUT+8` by the **unlinked** `unimplemented_exit`; nothing writes OUTPUT after the stamp.
- **P1 finding → bead `4ch8f.63.1`**: inside the verdict save/restore window the dispatcher uses OUTPUT+32 as its `halt_kind` slot — `returnRevertTail 1` writes `1` (RETURN) there with attacker-influenced return data at `[0..32)`. Every completed run discards it (restore, then stamp), but `isHalted` includes traps: a trap in that window is an observation-level **false accept**. `.64` therefore needs verdict-window trap-freedom — or the cheap hardening of moving `halt_kind` into the existing `rdg_halt_kind` `.data` mirror (recommended in the bead).

## Gates

- `#print axioms` on `guestImageEntries_extentsOk`, `guestImage_block_sub`, `guestScratch_matches_regionMap`, `guestScratch_sat`, `guestFraming`, `CodeReq.ofProg_sub_ofEntries_of_extentsOk`: **`propext, Classical.choice, Quot.sound` only**. Zero sorries.
- Full `lake build` green; `check-forbidden-tactics` OK; `check-layering` OK (GuestImage lives Codegen-side; EntrySpec stays Codegen-free); `check-guest-addrs` CLEAN. (`check-asm-to-program.sh` has pre-existing failures on main — TxPubkey/TxSigningHash/U256GasPricing/VerifyPublicKeysSenders — unrelated, present on a clean tree.)
- PLAN.md + ledger rows updated; `.63` completion note appended; children `.63.1–.63.12` filed.


pirapira:
**Review: one HIGH-severity finding on the `residue` artifact; everything else verified sound.** (Lean-proof PR — review-playbook §1/§2.) I independently re-derived the trust-boundary change, the composition lemma, and the satisfiability witness; the structural machinery is solid and the reported P1 is a model of the friction-reporting the bead asked for. But the `guestFraming.residue` you hand to `.64` cannot instantiate either top Prop as written.

### ✅ Verified independently
- **`MAX_INPUT_BYTES = 0x37FFFFF8` (the trust-boundary change) is correct and sound.** `isValidMemAddr` gates only the *start* address (`addr.toNat ≤ MEM_END`, inclusive; `Rv64/Basic.lean:303-318`), so a dword *starting* at `MEM_END = 0x78000000` is valid, and `0x40000010 + MAX = 0x78000008 = MEM_END + 8` puts the last payload dword start at exactly `MEM_END` — the sharp maximum, not off-by-one. The narrowing is not a weakening: any larger input needs dword reads with start > `MEM_END`, which trap in the model (the guest can't read its own payload), so it's outside the model's meaningful domain. Aligning to the existing `MEM_END` rather than widening it (§2a rejected alternative) is the right call. The `#guard`s pin it to the model boundary. **The bound is load-bearing and cross-validated**: `satWithin_inputBody`'s validity discharge only closes because the last body dword lands at exactly `0x78000000` — a looser `MAX` would make that `omega`/zone-check fail.
- **`MemSat.lean` is a genuine, non-vacuous sat library.** `SatWithin P lo hi = ∃ h, P h ∧ MemOnlyWithin h lo hi`; atoms require `isValidDwordAccess` (no witness for invalid addrs); `sepConj` chains disjoint footprints via real `Disjoint`/`union`. `.sat` yields an actual `∃ h, P h`. No cheat.
- **`guestScratch_sat` genuinely constructs a heap** for every admissible input (input record in the legacy zone `≤0x78000008`, six RAM tiles ascending in `[0xa0000000, 0xbfb80000)`), defeating the vacuity hole (§2.2). Region bases/sizes are pinned by the build (the literal `satWithin_ramRegion` args must unify with `regionScratch <region>`), and `guestScratch_matches_regionMap` (`decide`) pins the six `.ram` names+order.
- **`CodeReqExtents.ofProg_sub_ofEntries_of_extentsOk` is a correct compose-don't-enumerate lift**: `extentsOkFrom` enforces ascending non-overlapping extents in `[lo,hi]`; the tail case proves the head block *misses* `a` (extent bounds → `ofProg_none_range`) so `union_skip` falls through. Whole-image disjointness is one `decide` (`guestImageEntries_extentsOk`). Entries anchor by-name (`GuestAddrs.<entry>`, 358 entries, zero baked hex bases).
- **Axioms**: `guestImageEntries_extentsOk`/`guestScratch_matches_regionMap` → `[propext]`; `guestImage_block_sub`/`guestScratch_sat`/`ofProg_sub_ofEntries_of_extentsOk` → classical-3. Full `lake build` green.
- **The P1 (`.63.1`) verdict-window finding is a good, well-scoped report** (OUTPUT+32 `halt_kind` side-channel + a trap in the save/restore window = observation-level false accept; `rdg_halt_kind` `.data` mirror is the right cheap fix). Exactly the friction the bead wanted surfaced.

### 🔴 HIGH — `guestFraming.residue` over-owns the observation window (blocks the `.64` instantiation; unreported)
`residue := guestScratch`, and `guestScratch` contains `regionScratch RegionMap.outputRegion = anyBytes 0xa0010000 0x10000` — the **entire** OUTPUT region `[0xa0010000, 0xa0020000)`. But both top Props place the OUTPUT claim in **strict** separation with residue:
- `runStatelessGuestSound`: `guestOutputSound execute input ** fr.residue`, and `guestOutputSound` owns `bytesRegion OUTPUT_ADDR out` with `out.length = 40` → `[0xa0010000, 0xa0010028)`.
- `runStatelessGuestFaithful`: `bytesRegion OUTPUT_ADDR (serialize_…) ** fr.residue` (~105 B window).

The window is a **subset** of residue's OUTPUT tile, so `guestOutputSound ** residue` is unsatisfiable (a `PartialState.mem` maps each address once; no heap can give the window to both conjuncts). This directly contradicts the design note you kept in `EntrySpec.lean`: *"the OUTPUT observation window is deliberately NOT allowed to hide in `residue`."* As written, `.64` cannot prove `runStatelessGuestSound cr fuel guestFraming execute` for this `guestFraming` — the post has no model.

Note this is **not** a soundness defect in `.63` itself: `guestFraming` is a well-formed definition and `scratch_sat` (which constrains only the *pre*, `guestInputAssertion ** scratch`) is true — `scratch` legitimately owns the full OUTPUT at entry. The problem is confined to `residue` in the *post*. The danger is downstream: a `.64` session hitting the unprovable post could "fix" it by moving the window *into* residue and dropping `guestOutputSound`'s length pin — reintroducing exactly the ∃-out vacuity hole (#1) that note closed.

**Fix (cheap, and this is the file that owns `residue`):** keep `scratch = guestScratch`, but make `residue` carve the observation window out of the OUTPUT tile — split `regionScratch outputRegion` into `anyBytes 0xa0010000 OUTPUT_CLAIM_BYTES ** anyBytes (0xa0010000+OUTPUT_CLAIM_BYTES) (0x10000-OUTPUT_CLAIM_BYTES)` and keep only the tail in `residue` (the `anyBytes_add` sub-tile split you already use for `phaseDView`). Then `scratch ≠ residue`, and `guestOutputSound ** residue` re-tiles the full OUTPUT exactly. (If you want `.64`'s window to be the 105-B faithful window rather than 40 B, carve that instead / the max of the two.) I'd do it in this PR since guestFraming is billed as the artifact `.64` instantiates.

### Minor / non-blocking
- Coverage is 23.76% of `.text` (partial CodeReq) — clearly disclosed; `.64` needs the full image (child beads filed). Fine for the structural milestone.
- `runStatelessGuestFaithful` uses the same `residue` and so has the same overlap (larger window) — the fix above should account for the faithful window too.

Happy to re-review once `residue` carves out the window.

pirapira:
**Residue finding fixed** in 90f-range commit ("fix(guest-image): carve observation window out of guestFraming.residue"). You were right — `residue := guestScratch` re-owned the window and made the `.64` post unsatisfiable; thanks for catching it before a downstream session "fixed" it the vacuous way.

What changed (exactly the carve you proposed):

- `outputTailScratch := anyBytes (OUTPUT_ADDR + OUTPUT_CLAIM_BYTES) (size − OUTPUT_CLAIM_BYTES)` — the OUTPUT tile above the window.
- `guestResidue` = the six-region havoc with the OUTPUT tile replaced by `outputTailScratch`; `guestFraming.residue := guestResidue`. `scratch` and `scratch_sat` are unchanged (the guest legitimately owns the full OUTPUT at entry).
- `regionScratch_output_carve`: the entry tile splits at the dword-aligned claim boundary (via `SAsm.anyBytes_add`, `8 ∣ 40`).
- Bonus: `guestScratch_eq_window_residue : guestScratch = anyBytes OUTPUT_ADDR OUTPUT_CLAIM_BYTES ** guestResidue` — the retiling identity `.64` consumes directly (AC shuffle via `sepConj_comm'`/`sepConj_assoc'`), so `guestOutputSound ** guestResidue` re-tiles the entry-owned work regions exactly, with the window's ownership transferring from the havoc'd entry tile to `guestOutputSound`'s pinned-length `bytesRegion`.

On your minor note: documented in the docstring that `runStatelessGuestFaithful`'s window is the input-dependent full serialized output, so this residue serves the **soundness** Prop only — the faithful follow-up needs its own carve (it's the stated `.64` v1 non-goal).

`#print axioms` on `regionScratch_output_carve` / `guestScratch_eq_window_residue` / `guestFraming`: `propext, Classical.choice, Quot.sound` only. Full build green. PLAN.md wording ("scratch = residue") corrected; correction note appended to bead `4ch8f.63`.

pirapira:
**Re-review (commit `dde323e4c`): the HIGH finding is fully and correctly resolved → APPROVE.**

The fix carves the observation window out of the residue exactly as recommended:
- `guestResidue` replaces the OUTPUT tile with `outputTailScratch = anyBytes [OUTPUT_ADDR+OUTPUT_CLAIM_BYTES, OUTPUT_ADDR+size)` — the OUTPUT region *above* the window.
- `regionScratch_output_carve` splits the full OUTPUT tile at the (dword-aligned, `40 % 8 = 0`) claim boundary via `anyBytes_add`.
- `guestScratch_eq_window_residue : guestScratch = anyBytes OUTPUT_ADDR OUTPUT_CLAIM_BYTES ** guestResidue` is the retiling identity `.64` consumes to move window ownership from the entry scratch into `guestOutputSound`.
- `scratch` and `scratch_sat` correctly unchanged (the guest owns the *full* OUTPUT at entry; the carve is a post-only concern).

**Verified:**
- The post `Q = guestOutputSound ** guestResidue` is now satisfiable: window `[0xa0010000, 0xa0010028)` is disjoint from both `outputTailScratch` `[0xa0010028, …)` and `ziskSystemRegion` `[…, 0xa0010000)`. The `window ** (…window…)` unsatisfiability I flagged is gone.
- `#print axioms` on `regionScratch_output_carve` / `guestScratch_eq_window_residue` / `guestFraming` / `guestScratch_sat`: classical-3 only. `EvmAsm.Codegen.Proofs.GuestImage` builds clean.
- I confirmed the framework's `holdsFor` is affine (`CompatibleWith` constrains only owned cells), so the input record — owned by the pre but absent from `Q` — needs no home in the post; no second instance of this class exists.

**The `runStatelessGuestFaithful` caveat is correctly scoped.** That Prop's window is the input-dependent `serialize_stateless_output` byte string (wider than the fixed 40-byte `OUTPUT_CLAIM_BYTES`), so this residue serves the soundness Prop only. Since the faithful Prop is an explicit `.64` v1 non-goal (per EntrySpec) and the caveat is documented in-code, this is the right call — the faithful follow-up gets its own carve.

Everything from my first review still stands (MAX_INPUT_BYTES sharp bound, MemSat sat-library, the compose-don't-enumerate lift, entries anchored by-name, the well-reported `.63.1` P1). No remaining blockers from me. LGTM.

PR #9788, opened 2026-07-04
port(4ch8f.12): verify swd_write_be8 + swd_write_be32_u64 (SAsm whileS big-endian writers)

pirapira:
Second batch of the byte/copy leaf-helper ports (bead **evm-asm-4ch8f.12**) — the two fixed-trip big-endian **writers**, both `whileS`-loop ports over a writable region. Both pass `scripts/port-check.sh` (build, **zero warnings**, no forbidden tactics, **3 classical axioms only**).

## Beads
- **.12.8 `swd_write_be8`** — `EvmAsm/Codegen/Programs/SwdWriteBe8SAsm.lean`, theorem **`swdWriteBe8Fn_spec`**. 8-trip loop; post **`ws = beBytes a0`** (all 8 dst bytes = big-endian encoding of `a0`).
- **.12.7 `swd_write_be32_u64`** — `EvmAsm/Codegen/Programs/SwdWriteBe32U64SAsm.lean`, theorem **`swdWriteBe32U64Fn_spec`**. Two sequential loops (zero-fill 32 + BE-write low 8); post **`ws = List.replicate 24 0 ++ beBytes a0`** (all 32 dst bytes = the 32-byte big-endian storage-slot key).

Both posts pin **every** destination byte (unweakened). Per-iteration `execBlock` reductions are factored into `private` engine lemmas so each gets its own heartbeat budget (no `maxHeartbeats` bumps).

## ⚠️ Byte-identity caveat — needs reviewer sign-off
Neither `flatten = _prog` is claimed. The emitted routines' loop **back-edges re-initialize the limit register inside the loop** (`JAL -40` on be8 targets `LI x6 8`; be32's `JAL -20`/`-44` likewise), whereas a structured `while` back-edge targets the **guard** (`JAL -36`; `-16`/`-40`). The flattens are otherwise **identical** — same instruction count, same `BEQ` guard offsets, identical semantics; they differ **only** in the back-`JAL` offset field(s). A clean `while` cannot express "re-run the limit init before the guard each iteration", so instead each module pins an **explicit structured flatten `#guard`** (+ length + position-independence) and documents the single-field divergence in a source comment (plus `#guard <name>_prog.length = …`). The functional specs are faithful and unweakened. If byte-exact `_prog` correspondence is required for these, it would need either a codegen change (out of scope per the port fence) or a raw-`Program` (non-`while`) proof — flagging for your call.

## Notes
- Template: BalValueReverseSAsm `whileS` (`inv_init`/`inv_step`/`exhausted`/`body.*.mem`/`post`), single writable region, store via `execInstrRF_sb_byte`.
- Spec-only (no emitted-code change) → no EEST A/B.
- Updates PLAN.md + top-theorem-ledger row 6.


pirapira:
**Review: APPROVE the functional verification. Byte-identity sign-off below (you asked): the divergence is real but provably semantically inert — accept as a spec milestone, with a byte-identity follow-up before these are wired in.** (Lean-proof / spec-only, review-playbook §1.)

### Verified independently
- **`port-check.sh` PASS** on both modules (build, warning-free, no forbidden tactics, classical-3 across 38 / 60 theorems, `#guard` pins present).
- **Axioms** — `#print axioms` fully qualified: `swdWriteBe8Fn_spec` and `swdWriteBe32U64Fn_spec` → `[propext, Classical.choice, Quot.sound]`.
- **Adversarial statement read (§2):** posts pin **every** destination byte as a function of `a0` — `ws = beBytes a0` / `ws = replicate 24 0 ++ beBytes a0` — no ∃-escape, no weakening. `beByte v k = trunc8 ((v >>> (56−8k)) &&& 255)` is correct big-endian (k=0 → top byte). The loop invariants (`writeBe8Win`/`zeroWin`/`writeBe32Win`) are named functions of `(v, orig, i)` — no ∃-state. Pre pins `orig.length` and `ws=orig`; the `.mem` VC bounds are load-bearing. `store_val_eq_beByte` proves the stored value equals the intended BE byte. Clean.

### Byte-identity sign-off (the flagged caveat)
I diffed the `while` flatten against the emitted `_prog`s (`SystemWrites.lean`). Your claim is exactly right — the **only** difference is the back-edge `JAL` offset field:
- `swdWriteBe8_prog[11] = JAL -40` (re-targets `LI x6 8` at index 1) vs the `while` `JAL -36` (targets the `BEQ` guard at index 2). Indices 0–10 + the trailing `ret` are identical.
- `swdWriteBe32U64_prog` likewise: `JAL -20`/`-44` (re-target the `LI x6` limit re-inits) vs `while` `-16`/`-40` (guards).

The extra per-iteration `LI x6, <limit>` is a **constant** the loop body never modifies, so `x6 = <limit>` at every guard check in both versions — the re-init is observationally inert. **The `while` port is semantically equivalent to the emitted routine, and the functional specs are faithful.** Not claiming `flatten = _prog` is the honest call; the explicit structured-flatten `#guard` + length + PIC + `_prog.length` are the right pins in its place.

### The one thing to track (not a blocker, but don't let the ledger overclaim)
Because `flatten ≠ _prog`, these two — unlike `swd_read_u64le` (which *is* byte-pinned) — **cannot yet contribute to the deployed-image `CodeReq` (`.63` `guestImageEntries`)**: that image is `ofProg` over the *emitted* bytes, byte-identity-gated by `check-asm-to-program`. So they're **functionally verified, byte-identity pending** — drop-in only after one of:
- **(a) recommended:** a small codegen change to emit *guard-targeting* loops (drop the redundant in-loop `LI x6` re-init). It's a byte change (⇒ EEST A/B), but semantically inert, and it **generalizes** — every `while`-shaped emitted loop would then be byte-identical to its structured port, unblocking this whole writer family. Higher leverage than per-routine fixes.
- **(b)** a raw-`Program` (non-`while`) proof directly over the emitted `_prog` (keeps deployed bytes; harder — the `while` combinator can't express a back-edge-to-init loop).

Please tweak the ledger row 6 to say "**done (functional; byte-identity pending — back-`JAL` divergence)**" for `.12.7/.12.8` rather than unqualified "done," so coverage isn't overcounted, and file the resolution as a follow-up. `Imports.lean` wiring all four `.12` modules is good (keeps them CI-built).

LGTM to merge as a functional-spec milestone.

PR #9791, opened 2026-07-04
port(4ch8f.12): verify swr_rev_le_be + bhr_rev_le_be (SAsm reverse-copy loops)

pirapira:
Third batch of bead **evm-asm-4ch8f.12** — the two byte-**reverse** copies. (Rebased onto main after the writers PR #9788 merged.) Both pass `scripts/port-check.sh` (build, **zero warnings**, no forbidden tactics, **3 classical axioms only**).

## Beads
- **.12.4 `swr_rev_le_be`** — `EvmAsm/Codegen/Programs/SwrRevLeBeSAsm.lean`, **`swrRevLeBeFn_spec`**. Runtime-length `.«while»` loop (cap = symbolic `len`) over a **read-only src region** `⟨src,bs⟩` and a **writable dst region** `⟨dst,len⟩`. Post **`ws = (bs.take len).reverse`** — `revWin_len_eq` proves the loop window at `i = len` equals the reverse, so the spec genuinely says *reverse*. Byte-identity **fully pinned**: `flatten ++ [ret] = swrRevLeBe_prog`. Per-iteration reduction in `swr_step_engine` (own heartbeat budget); read-only load via local `execInstrRF_lbu_ro`.
- **.12.5 `bhr_rev_le_be`** — `EvmAsm/Codegen/Programs/BhrRevLeBeSAsm.lean`, **`bhrRevLeBeFn_spec`**. `bhrRevLeBe_prog` is byte-identical to `swrRevLeBe_prog`; reuses the generic core and kernel-pins `#guard … = bhrRevLeBe_prog`.

## Disjointness (STOP rule) — please sanity-check
The routine **reads src while writing dst**, so overlapping buffers could corrupt not-yet-copied source bytes. The precondition therefore **explicitly requires src/dst disjoint**:
```
src.toNat + len ≤ dst.toNat ∨ dst.toNat + len ≤ src.toNat
```
This is the routine's real reverse-*copy* contract, **not** a weakening — a caller with overlapping buffers cannot satisfy the pre. Full pre also carries `orig.length = len`, `len ≤ bs.length`, `src/dst + len < 2^64`. No guest↔spec divergence found.

Spec-only (no emitted-code change) → no EEST A/B. Updates PLAN.md + top-theorem-ledger row 6.


pirapira:
**Review: APPROVE.** (Lean-proof / spec-only, review-playbook §1.) Cleaner than the writers PR — byte-identity is fully pinned here.

### Verified independently
- **`port-check.sh` PASS** on both modules (build, warning-free, no forbidden tactics, classical axioms across 52 / 1 theorems, `#guard` pins present).
- **Byte-identity FULLY pinned**: `#guard swrRevLeBeBody.flatten 0 ++ [ret] = swrRevLeBe_prog` (and `bhrRevLeBe_prog`) elaborate — so unlike the `.12` writers (#9788), these ARE byte-identical to the emitted routine and drop-in-ready for the `.63` image `CodeReq`. Root cause of the difference confirmed: the counter `x7 = len` is set once in the init and *decremented* in the loop (no in-loop limit re-init), so the structured `while` back-edge targets the guard exactly as the emitted `JAL -24` — the divergence that bit #9788 is absent here.
- **Adversarial read (§2):** post `ws = (bs.take len).reverse` is the real reverse (proved by `revWin_len_eq` via `getElem` congruence), a pure function of input — no ∃-escape. Loop invariant `revWin` is a named function (no ∃-state). `revByte bs len k = bs[len-1-k]` is correct reversal.
- **Disjointness pre — correct STOP-rule call, not a weakening.** `src.toNat+len ≤ dst.toNat ∨ dst.toNat+len ≤ src.toNat` is genuinely load-bearing: the block engine's `inRw` routing is *arithmetic* (`execInstrRF_lbu_ro`), so each `LBU` only routes to the read-only src region when its address provably misses the writable dst window — which needs disjointness. A reverse-*copy* into a separate buffer genuinely requires it; an overlapping caller cannot satisfy the (unweakened) contract. Not over-constrained (§2.7). ✓
- Spec-only → no EEST/layout gate. `bhr_rev_le_be` is byte-identical to `swr_rev_le_be` and correctly reuses the core + pins its own `_prog`.

No findings. LGTM.

PR #9794, opened 2026-07-04
port(4ch8f.12): verify sg_memcpy + mset_memcpy (completes the .12 byte/copy family)

pirapira:
## What

Verified SAsm ports of the two forward byte-copy leaves, completing the `.12` byte/copy family:

- **`sg_memcpy`** (`.12.2`) — `SgMemcpySAsm.lean`
- **`mset_memcpy`** (`.12.1`) — `MsetMemcpySAsm.lean`

Both are RO-src + RW-dst `«while»` copies; the pre requires `src`/`dst` disjoint (the routine corrupts on overlap when `dst > src`, so this is the real contract — no guest↔spec divergence).

## Status of the family

`swd_minimal_copy` (`.12.9`) — the third `.12` leaf that was documented here as a blocker — is now **fully verified and merged to `main`** (PR #9810, `swdMinimalCopyFn_spec`, classical-3). It required two new pieces of infrastructure that also landed on `main`:
- `.67` — `blockAt`/`focus_rwAtom` (multiple writable regions), PR #9798;
- `.huy8w` — `Stmt.whileBreak` (mid-loop early exit), PR #9804.

So this branch (rebased on the current `main`, dropping the old blocker note) now carries the verified `swd_minimal_copy` as baseline and adds the two memcpys, making the `.12` family complete.

## Gates

- `scripts/port-check.sh` PASS for both `SgMemcpySAsm.lean` and `MsetMemcpySAsm.lean` (build, warning-free, no forbidden tactics/`sorry`, classical axioms only).
- Corpus `lake build` green.

### DIFF
--- a/EvmAsm/Codegen/Programs/Imports.lean
+++ b/EvmAsm/Codegen/Programs/Imports.lean
@@ -86,6 +86,8 @@ import EvmAsm.Codegen.Programs.SwdWriteBe32U64SAsm
 import EvmAsm.Codegen.Programs.SwrRevLeBeSAsm
 import EvmAsm.Codegen.Programs.BhrRevLeBeSAsm
 import EvmAsm.Codegen.Programs.SwdMinimalCopySAsm
+import EvmAsm.Codegen.Programs.SgMemcpySAsm
+import EvmAsm.Codegen.Programs.MsetMemcpySAsm
 import EvmAsm.Codegen.Programs.BalGasValid
 import EvmAsm.Codegen.Programs.BalCodePreimages
 import EvmAsm.Codegen.Programs.BalAccountHasStateChange
--- a/EvmAsm/Codegen/Programs/MsetMemcpySAsm.lean
+++ b/EvmAsm/Codegen/Programs/MsetMemcpySAsm.lean
@@ -1,0 +1,70 @@
+/-
+  EvmAsm.Codegen.Programs.MsetMemcpySAsm
+
+  Verified SAsm port of `mset_memcpy` (bead evm-asm-4ch8f.12.1): copy `len`
+  bytes forward from `src` (a1) to `dst` (a0).
+
+  `msetMemcpy_prog` (MptSet.lean) has the **same loop body and register roles**
+  as `sg_memcpy` (x10=dst, x11=src, x12=len; `LBU x5,(x11); SB (x10),x5;
+  x10++; x11++; x12--`), and the same net effect `dst = src[0..len)`.  It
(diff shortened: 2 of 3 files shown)


pirapira:
**Review: APPROVE.** The two forward memcpys — completing the `.12` byte/copy family. (Substance review; CI covers build/port-check/forbidden-tactics/axioms.) The ghost `SwdMinimalCopy` module is correctly gone (the real `.12.9` triple landed in #9810), leaving exactly `SgMemcpy` + `MsetMemcpy` + `Imports`.

### Verified (substance)
- **`sg_memcpy` (.12.2)**: post `ws = bs.take len` (`dst = src[0..len)`) — a pure function of input, **no ∃-escape**; `copyWin`/`copyByte` invariant; the **disjointness** pre (`src+len ≤ dst ∨ dst+len ≤ src`) is load-bearing (routes each `LBU` to the read-only region past the arithmetic `inRw`) and correct for a forward copy, not a weakening. String-emitted (no `_prog`) → length + position-independence pinned (eyeball-matched, as with `sg_load_u32le`).
- **`mset_memcpy` (.12.1)**: reuses the `sg` core (same body/spec), post unweakened. `msetMemcpy_prog` is a **do-while** (bottom `BNE x12,x0` back-edge); the structured-`while` flatten differs *only* in the back-edge (`JAL -24`→guard vs `BNE -20`→body), so byte-identity is (correctly) **not claimed** and documented. This is exactly the shape `.68` doWhile addresses.

### Byte-tie status (non-blocking, under the drop-in policy)
Both are functionally verified; neither is yet machine-byte-tied to the emission:
- `sg_memcpy` (string literal) → tie by emitting `emitProgram sgMemcpy_verified` (drop-in), the `swd_read_u64le` pattern.
- `mset_memcpy` (do-while) → tie once **`.68` doWhile** lands (re-do the strip... loop as `doWhile`, byte-identical to `msetMemcpy_prog`), OR a drop-in re-emit to the top-test form.
Both are the "verified functional; byte-tie via drop-in / `doWhile` follow-up" class — fine for the family-completion milestone. Worth a small follow-up bead ("machine byte-tie the `.12` string-emit/do-while leaves: `sg_load`, `sg_memcpy`, `mset_memcpy`, the `.12` writers") so the drop-in/doWhile ties are tracked before these feed the `.63` image `CodeReq`.

No findings. Approve — this closes the `.12` byte/copy family (readers `.12.3/.6`, writers `.12.7/.8`, reverses `.12.4/.5`, copies `.12.1/.2`, `swd_minimal_copy` `.12.9`). Close `.12.1`/`.12.2` and the `.12` parent on merge.

PR #9798, opened 2026-07-04
sasm(4ch8f.67): multiple writable regions — atom-routed stores via blockAt + two-region demo

pirapira:
## SAsm multi-writable-region capability (bead `evm-asm-4ch8f.67`)

A routine that writes **two or more independent writable pointers** (result buffer + length/count dword — the `swd_minimal_copy` shape, `.12.9`/PR #9794 blocker) can now be given a faithful `Fn.Spec`.

### Design chosen (of the two candidates), and why

**The assertion-atom-routable-store design, realized at block granularity by the existing `blockAt`** — no engine or AST change:

- region 1 is the function's primary `rw : RwRegion` (contents threaded through `ws` as usual);
- regions 2, 3, … are `bytesRegion` conjuncts of the ambient assertion `A`, each written through a `Stmt.blockAt` focused at that region's pointer register.

Why not `rw : List RwRegion`: the single flat `ws` is load-bearing in `Reach` and in every corpus proof pattern, and the machine-level block-soundness proof (`execBlock_sound`, ~700 lines in `RegionSound.lean`) is single-window — a list-of-regions engine means re-proving it multi-window and rewriting corpus-wide, which is exactly the corpus-red risk the bead's STOP rule fences. The `blockAt` route reuses the already-kernel-checked store-routing soundness (`Stmt.sound`'s `blockAt` case) and is additive-only, so the in-flight `.49.2`/`vgyg9` work needs no rebase.

Block-granular routing loses **no expressiveness**: `seq` of `block`/`blockAt` leaves flattens to the same contiguous instruction stream with *zero* synthesized instructions (`#guard` in `MultiRw.lean` pins `flatten = copyBlock ++ countBlock`), so interleaved stores to different regions are expressed by cutting the instruction list at region-switch boundaries; values cross cuts in registers, as the machine code does anyway.

### Store-routing soundness + disjointness (reviewer kill-criteria)

- A store in a plain `block` can only land in the primary `rw`: the `.mem` VC demands `inRw rw.base ws a n` (address inside region 1). A store in a `blockAt` can only land in that region's window (`inRw (rf.get p) win a n`). **A store placed against the wrong region is an unprovable `.mem` goal, never a misroute.** Framing is by the proven `sp`/soundness: inside a `blockAt`, the primary `ws`, the read-only region, and the other `A`-atoms (`rest`) are untouched by construction.
- **Disjointness is required structurally, not silently assumed**: the regions live under `**` in `asrtM`, so an overlapping instantiation makes the precondition *unsatisfiable* — no machine state satisfies it and nothing can be concluded past it (try it: put `cnt` inside `[dst, dst+8)` and the caller cannot produce the entry `asrtM` state). The arithmetic `inRw` test never routes across regions because each block's test is against a single region. The demo's only arithmetic hypothesis is `src ≠ dst`, which resolves the *load* routing `if` (read-only source vs. writable dst) — noted in the theorem docstring.
- **No `∃`-region escape**: the demo post pins BOTH regions as functions of the input — `ws = bs` (region A = the copied read-only bytes) and `A = ⌜RwRegion.wf ⟨cnt, 8⟩⌝ ** bytesRegion cnt (dwordBytes 8)` (region B = the count dword).

### Backward compatibility

Zero changes to `Fn`/`Stmt`/`Reach`/engine — the whole corpus compiles unchanged (full `lake build` green, `check-forbidden-tactics.sh` clean). The only framework-file diff is the `SAsm.lean` aggregator import line.

### Artifacts (`EvmAsm/Rv64/SAsm/MultiRw.lean`)

- Design decision record (file header).
- `SAsm.focus_rwAtom` — generic discharge of a second-region `.focus` VC for the canonical `A = ⌜RwRegion.wf ⟨b, n⟩⌝ ** bytesRegion b w` contract shape.
- `dwordBytes_packBytes` — the 8-byte unpack∘pack round trip (feeds "region := copied input bytes" posts).
- `MultiRw.twoRwFn` + `twoRwFn_spec` — the demo: three genuinely independent pointers (`src` read-only, `dst` region A via primary `rw`, `cnt` region B via ambient atom); copies 8 bytes from `src` into region A and writes the byte count into region B.
- Docs: `docs/sasm-howto.md` §6 new recipe "Multiple writable regions"; `docs/sasm-design.md` §3.6.1 item 2 notes this as the sanctioned design; PLAN.md SAsm log entry.

### `#print axioms` (kernel truth, classical-3)

```
'EvmAsm.Rv64.SAsm.MultiRw.twoRwFn_spec' depends on axioms: [propext, Classical.choice, Quot.sound]
'EvmAsm.Rv64.SAsm.focus_rwAtom' depends on axioms: [propext, Classical.choice, Quot.sound]
'EvmAsm.Rv64.dwordBytes_packBytes' depends on axioms: [propext, Classical.choice, Quot.sound]
```

### `.12.9` (stretch, deferred per STOP rule)

The capability demonstrably unblocks `swd_minimal_copy`: dst buffer = primary `rw` written by the copy loop's `SB`s; the `SD x6, 0(a3)` at `.Lstore` = one `blockAt` on the `a3` dword atom (with the 8-alignment pre); loop invariants carry the atom through `A`. The full port (two data-dependent `whileS` loops + byte-identity pin against `swdMinimalCopy_prog`) is a follow-up session; mapping recorded in the bead.


pirapira:
**Review: APPROVE.** Elegant, minimal, backward-compatible realization of the multi-writable-region capability — exactly the "atom-routable store" design option, done without an engine change. (SAsm infra, spec-only.)

### Verified independently
- **Design is sound and genuinely multi-region.** Region 1 = primary `rw` (plain `block`); regions 2+ = `bytesRegion` atoms of the ambient `A`, each written via `Stmt.blockAt` focused at its pointer. The demo `twoRwFn` has three independent pointers (`src` read-only, `dst`=region A via `rw`, `cnt`=region B via `A`-atom) and its post pins **both** writable regions as functions of input (`ws = bs`, `A = ⌜wf⌝ ** bytesRegion cnt (dwordBytes 8)`) — **no ∃-region escape**.
- **Store routing is sound, not a misroute.** A plain-block store must satisfy the `.mem` VC `inRw rw.base ws a n` (region 1); a `blockAt` store must fit the focused window (`inRw (rf.get p) win a n`). A store placed against the wrong region is an **unprovable** `.mem` goal. Framing rides on the already-kernel-checked `Stmt.sound` `blockAt` case (`ws'=ws`, `A''=bytesRegion (rf.get p) win' ** rest`) — no re-proof of `execBlock_sound`. I confirmed the choice avoids the corpus-red risk: **`lake build` green across the corpus + the `SAsm` aggregator**, zero `Fn`/`Stmt`/`Reach`/engine change.
- **Disjointness is structural** (regions under `**` in `asrtM` ⇒ overlapping instantiation makes the pre unsatisfiable) — strictly stronger than an arithmetic side-condition. The only arithmetic hyp in the demo is `src ≠ dst` (resolves the read-only-vs-writable *load* routing), as documented.
- **Zero codegen cost:** `#guard (twoRwBody …).flatten 0 = copyBlock ++ countBlock` — block-granular routing synthesizes no instructions; interleaved stores are expressed by cutting the stream at region-switch boundaries (values cross in registers). Adequate for the "result buffer + length/count" pattern (howto §6 recipe added).
- **`focus_rwAtom`** is a clean reusable discharge for the canonical `A = ⌜wf⌝ ** bytesRegion` contract — what `.12.9` and future consumers plug into.
- **Axioms:** `twoRwFn_spec`/`focus_rwAtom`/`dwordBytes_packBytes` → `[propext, Classical.choice, Quot.sound]`. No `sorry`/`native_decide`/`bv_decide`/heartbeat bumps.

### Notes
- This is the elegant realization of my option (a) from the #9794 review — the "atom-routable store" via existing `blockAt`, not `rw : List RwRegion` (correctly avoided: it would force a multi-window `execBlock_sound` re-proof + corpus rewrite, the STOP-rule fence). Good call. The #9794 blocker premise (a *single-`rw` `Fn`* can't) still holds; this provides the multi-region path around it.
- Unblocks **`.12.9`** (`swd_minimal_copy`): dst = primary `rw`, the `SD x6,0(a3)` = one `blockAt` on the `a3` atom. That triple → the parked PR #9794 as a follow-up (mapping recorded).

No findings. LGTM.

PR #9805, opened 2026-07-04
fix(eest): repair create init fail receipts

pirapira:
## Summary
- repair singleton create_init_fail receipt materialization before the receipt-root validator runs
- add the missing CREATE transfer-log amount into the second synthetic EIP-7708 log data slot under the exact two-log signature
- regenerate RegionMap text size and symbol-address snapshot after the codegen text-size change

## Tests
- PASS: scripts/codegen-eest-stateless-check.sh --backend spike --filter create_init_fail --all --jobs 10 --max-jobs 10 --steps 1000000000 --run-dir gen-out/eest-run/codex-create-init-fail-logdata-repair --quiet-passes --max-failures 20
  - selected 8, full match 8, fail 0
- PASS: scripts/check-region-map.sh
- BROAD SWEEP: scripts/codegen-eest-stateless-check.sh --backend spike --all --jobs 10 --max-jobs 10 --steps 1000000000 --run-dir gen-out/eest-run/codex-broad-after-create-init-repair --quiet-passes --max-failures 20
  - stopped at 20 existing bv_fail=53 receipt-root failures after 19,398 ran; 19,378 full matches
  - focused create_init_fail failures are no longer in the failure set

Refs evm-asm-x3dnt.


pirapira:
Update: pushed a fix for the file-size guardrail failure (split `BlockVerdictReceiptsTail.lean` → new `BlockVerdictReceiptsTailHead.lean`, mirroring the existing `ChildFrameHandlerTails`/`Bls12MapG1Real` split pattern). Full local build passes on that commit alone.

However, merging current `main` into this branch now also updates `EvmAsm/Codegen/RegionMap.lean` / `scripts/asm-fixtures/symbol-addresses.tsv` (conflicting generated files), and after taking this branch's versions the merged build fails:

```
error: EvmAsm/Codegen/Proofs/GuestImage.lean:58:37: Tactic `decide` proved that the proposition
is false
```

This is the expected "link-layout drift" case documented in `scripts/check-region-map.sh` (`.text`/`.data` sizes moved, needs `scripts/gen-symbol-addresses.py --build` + `GuestAddrs.lean` regen), but regenerating those requires the whole library (including `EvmAsm.Codegen.Proofs.GuestImage`) to build successfully first — a real circular dependency once main has advanced past this branch's fork point by enough to shift guest addresses. I noticed the `merge!` label was already pulled by @pirapira, so I'm leaving this alone rather than guessing at a manual RegionMap size bump. Flagging with **needs human review**.

PR #9803, opened 2026-07-04
proof(cfjzu): jumpdest-bitmap loop — L1 walk decomposition + L3 bit-RMW + model

pirapira:
## What

Bead `evm-asm-cfjzu` (child of `.49.2`) — the jumpdest-bitmap build loop (`emitJumpdestBitmapBuild`, `Dispatch.lean:142-206`). This PR lands the two crux layers the prompt flagged as risk areas (walk decomposition L1, bit RMW L3) plus the loop **model** `.49.d` consumes. Spec-only; no guest bytes.

## L1 — walk-prefix decomposition (`SpecRef/Runtime.lean`)

`walkFrom` (well-founded boundary walk) with `walkFrom_zero_eq` bridging to the merged anchor `validJumpDestinations`; `Reaches` (on-the-walk / instruction boundary) with `.le`/`.extend`/`Reaches_zero_le` (a boundary from `0` is `≤ len+32`, bounding the `PUSH` overshoot); `splitWalk`/`walkFrom_subset`; `jdAdvance_le ≤ 33`.

**`vjd_lt_step`** (the arm crux): between a boundary `pc` and its successor `pc + jdAdvance code pc`, the only possible valid destination is `pc` itself (iff `code[pc]=0x5b`). This is where the pushdata/immediate skips are **load-bearing** — deleting a `PUSH`-skip arm changes `jdAdvance` and the boundary set.

## L3 — bitmap bit/byte layer (`Codegen/Proofs/JumpdestBitmap.lean`)

`bitmapBit`; `bitmapBit_setBit` (the `lbu; or; sb` read-modify-write sets exactly bit `pc`, leaves others); `bitmapBit_replicate_zero` (loader-zeroed).

## Loop model

`scanFn` (two-region `Fn`: RO code `region` + writable bitmap `rw`), `scanInv` (`x5` at a boundary via `Reaches`; bitmap encodes `validJumpDestinations` below `pc`), `scanBody` (4-arm `ite`: JUMPDEST / plain-`<0x60` / `PUSHn` / plain-`≥0x80`), and `post` = the full valid-jumpdest set. Scoped to the EIP-8024-free, `≤ capacity` input class (both childable; the `PUSH`-skip stays load-bearing, post unchanged).

## Remaining (documented in bead `cfjzu`)

[...]

PR #9808, opened 2026-07-04
feat(vgyg9=.49.a): stack guards join the flag+ret discipline

pirapira:
# feat: stack guards join the flag+ret discipline (vgyg9 = 4ch8f.49.a)

Closes the `.49` handler-entry seam found in `evm-asm-vgyg9`: the emitted
stack-underflow/overflow guard prologues were multi-exit (`bltu/bleu …
.exit_stack_underflow/overflow`), so the dispatch-table witness
(`addrOf h_<OP>`) could not equal any callRegS-conformant handle entry.

## Decision (recorded as interp-strategy §3 amendment)

Option (c): finish what `.10.3` started — the guards set `evm_halt_flag`
(7 = underflow, 8 = overflow) and `ret`; `emitDispatchResume` routes 7/8
to the existing exceptional exit joins. Options (a) "prove not-taken"
(unsound: the guard is live), (b) guards→loop header + arity table (more
invasive, no proof-side gain), and (d) multi-exit call primitive (§3
rejection stands) are documented in the amendment.

**Deliberate deviation from `dispatchHaltRet`**: the guard halt block does
NOT load `.Ldispatch_resume` into `x1`. Guards run at handler entry where
`x1` still holds the `jalr`-passed return address, and preserving the
caller's `x1` is what keeps the packaged handle's ∀-`ret`
`FnHandleS.sound` contract provable. Routing is byte-behaviorally
identical at the real site.

## Scoped `.s` diff (base 6308794c7 vs branch, classified full census)

Removed (167): `bltu x14, x12, .exit_stack_underflow` ×93,
`bleu x12, x14, .exit_stack_overflow` ×71, raw DUPN/SWAPN/EXCHANGE
underflow branches ×3.
Added (1006): per-site inverted skip-branch (`bgeu/bltu … 137f`) ×167 +
halt blocks (`li x5, 7|8; la x6, evm_halt_flag; sd x5, 0(x6); ret; 137:`)
×167 (96 code-7, 71 code-8), and the two resume arms
(`li x7, 7/8; beq x6, x7, .exit_stack_underflow/overflow`).
Nothing else. (Diff produced from `lake exe codegen --asm-only` at base
vs branch; classification script output in the review notes.)

## Layout regen (same PR)

`gen-symbol-addresses.py --build` + `asm_to_program.py guest-addrs` +
`RegionMap.textSizeBytes` repin `0x59bf8 → 0x5a688`.
[...]

PR #9812, opened 2026-07-04
proof(cfjzu.1): scanFn_spec — jumpdest-bitmap loop triple (restricted class)

pirapira:
## What

Bead `evm-asm-cfjzu.1` (continuation after #9803). Completes the **whileS proof** (`scanFn_spec`) for the dispatcher's jumpdest-bitmap build loop (`emitJumpdestBitmapBuild`, `Dispatch.lean:142-206`), against the merged SpecRef anchor `validJumpDestinations`. Spec-only; no guest bytes.

## The triple

`(scanFn codeBase bitmapBase code).Spec base` for the input class covering all pre-EIP-8024 bytecode:
- `code.length ≤ bitmapCap` (EIP-3860 clamp is a no-op; `> capacity` → child `cfjzu.3`)
- no EIP-8024 opcode `0xe6/0xe7/0xe8` (DUPN/SWAPN/EXCHANGE arms → child `cfjzu.2`)

**Post** (the bitmap-built precondition `.49.d` consumes) — the full mathematical set, not weakened:
```
∀ idx, idx < code.length → bitmapBit ws idx = decide (idx ∈ validJumpDestinations code)
```
The pushdata-aware `PUSHn` arm is present and **load-bearing** (its `jdAdvance = op − 0x5e` skip is what keeps the invariant's boundary walk aligned; deleting it breaks the proof via `vjd_lt_step`).

## Machinery (all classical-3, kernel-checked)

- **`jdset_engine`** — the JUMPDEST-arm `lbu; or; sb` read-modify-write `execBlock`, resolving the RMW load's `inRw` routing (bitmap byte `pc/8` in the writable window, from `pc < clampLen ≤ capacity = bitmapBytes*8`).
- **`plainArm_exec` / `pushArm_exec`** — abstract-entry arm `execBlock` engines.
- **`nonjd_step`** — uniform invariant step for every non-JUMPDEST arm (`ws` unchanged, `x5 += jdAdvance`; `vjd_lt_step`'s right disjunct vanishes since `code[pc] ≠ 0x5b`).
- **`scanFn_inv_step`** — the 4-leaf `inv_step` (JUMPDEST via `bitmapBit_setBit` + `vjd_lt_step`; the three non-JUMPDEST arms via `nonjd_step`), plus `jdset.mem`/`load.mem`/`region`/`inv_init`/`exhausted`/`post`. The **overshoot** (a `PUSHn` past `x6`) is handled: `pc` is bounded by `Reaches_zero_le` (`≤ len+32`) and valid dests are `< len`, so `idx < pc` suffices. Code/bitmap **disjointness** is an explicit pre-hypothesis (`.49.d` discharges it from `RegionMap.guestRegionMap_pairwise_disjoint`, text `0x80000000` vs data `0xa3000000`).

## Kernel deep-recursion (and the gate-compliant fix)

The 4-arm branchy body's inlined `execBlock` computation hit a **kernel `deep recursion detected`** (proof-term depth). Resolved *without* `maxRecDepth` or an irreducible-budget bump, per the bead guidance: (1) keep the register file **abstract** in the arm application via the `*_exec` engine lemmas (so the deeply-nested register file only appears in small, isolated `have`s), and (2) close the pointer arithmetic with **`bv_omega`** rather than `apply BitVec.eq_of_toNat_eq; simp [toNat_*]; omega` (the latter's certificate over `2^64`-mod arithmetic was the recursion source).

## Axioms (`#print axioms`)

`scanFn_spec` → `[propext, Classical.choice, Quot.sound]` (classical-3). No `sorry`/`native_decide`/`bv_decide`; no `maxHeartbeats`/`maxRecDepth`; `check-forbidden-tactics.sh` clean; warning-free.

## Remaining for the general class

`cfjzu.2` (EIP-8024 arms — additive, `jdAdvance` already encodes them so `vjd_lt_step` still closes) and `cfjzu.3` (`> capacity` clamp). After both, `.49.d`'s only remaining blocker is `vgyg9`.


pirapira:
**Review: APPROVE** (`cfjzu.1` — the restricted-class jumpdest-bitmap loop triple). Sound, honestly scoped, and the deep-recursion fix is exemplary (gate-compliant). LGTM.

### Verified independently
- **`scanFn_spec` proven**, `#print axioms` → `[propext, Classical.choice, Quot.sound]`. `port-check.sh` PASS (build, warning-free, no forbidden tactics/`sorry`, 105 theorems all classical).
- **Post is the full set, unweakened**: `∀ idx < code.length, bitmapBit ws idx = decide (idx ∈ validJumpDestinations code)` — no ∃, matches the merged anchor. The `PUSHn` arm is present and load-bearing (its `jdAdvance = op−0x5e` skip keeps the boundary walk aligned; deleting it breaks `inv_step` via `vjd_lt_step`).
- **Restrictions are explicit hypotheses** (honest, not hidden): `hcap` (`≤ bitmapCap`), `hNoEip` (`∀ k, code[k] ∉ {0xe6,0xe7,0xe8}`), `hwf`, and code/bitmap disjointness `hdisj` as a pre-hypothesis (`.49.d` discharges it from `RegionMap.guestRegionMap_pairwise_disjoint`). The overshoot is handled correctly (`Reaches_zero_le` bounds `pc ≤ len+32`; valid dests are `< len`, so `idx < pc` suffices).
- **Deep-recursion fix is the right way**: no `maxRecDepth`/`native_decide`/`bv_decide` — the register file is kept abstract via the `*_exec` engine lemmas (so the deep RF only appears in isolated `have`s) and the pointer arithmetic closes with `bv_omega`. That's exactly the gate-compliant resolution the project mandates (find the cause, don't bump the budget).

### Notes / tracking (non-blocking)
1. **Correction to the PR's closing line**: `vgyg9` is already **merged** (#9808). So after `cfjzu.2`/`cfjzu.3`, `.49.d`'s remaining blockers are `cfjzu`-completion **and `.49.4`** (the guard-carrying handles re-packaged for the *full* set + `FnHandleS`-wrapped — `vgyg9`/#9808 only landed the ADD exemplar), not `vgyg9`.
2. **Byte-tie deferred (tracked)**: `scanFn` is a *model* of `emitJumpdestBitmapBuild`; its `flatten` can only be tied byte-for-byte to the emitted build once `cfjzu.2` adds the EIP-8024 arms (the emitted build has them). Until then `scanFn_spec` is a valid loop-spec but not yet pinned to the real guest build — that tie (or a drop-in re-emit) is a `cfjzu`-completion requirement before `.49.d` consumes it against the real dispatcher. I'll note it on `cfjzu`.

No findings. Close `cfjzu.1` on merge; `cfjzu.2`/`cfjzu.3` + the byte-tie remain for the general class. Great execution on the deep-recursion wall — worth capturing that pattern (abstract-RF-via-engine-lemmas + `bv_omega`) for future branchy loop bodies.

PR #9804, opened 2026-07-04
sasm(4ch8f.huy8w): break/early-exit while combinator

pirapira:
## What

Adds `Stmt.whileBreak`, a bounded loop with a **mid-body early exit** (break) to the SAsm structured-assembly DSL — the control-flow shape the single-exit `«while»` cannot express: a header guard **plus** a mid-loop conditional branch that jumps out of the loop *past the synthesized back-edge* (the machine idiom "scan until a predicate holds").

Surfaced by `swd_minimal_copy` (`.12.9`), whose leading-zero strip loop (`BNE x7,x0 → .Lstore` at `+16`, past `JAL -20`) is byte-identical to a structured body at every instruction **except** that early-exit branch. Bead `evm-asm-huy8w`.

**Purely additive** — every existing `«while»`/`«whileS»` user is untouched.

## Design (a break just short-cuts the loop's post)

The break arm is one extra VC: when the break condition holds at the break point, the loop's exit `post` already holds. Everything else mirrors `«while»`.

- **WP** (`WP/Loop.lean`): `loopBreakNatCert` + `loopBreakNatCert_sound`. Same `loopBound` budget; the per-iteration body is a *branch* — continue → header (`inv (i+1)`), or break → `exit_` (`breakPost`); both `exitPost`/`breakPost ⇒ post`.
- **AST/Flatten**: constructor + `size`/`callFree`; layout `[B¬guard→Lexit] · bodyBefore · [B breakCond→Lexit] · bodyAfter · [JAL→hdr]`; `offsetsOk`, `callsOk`.
- **Vc**: `sp` (= the stated `post`); five VCs — `inv_init`, `inv_step`, `exhausted`, `guard_exit`, `break` — + per-body VCs; `steps`; `sp_mono`/`vcs_antitone`/`CalleesIn`/`EndsWith` arms.
- **Soundness**: the `whileBreak` case in both `Stmt.sound` and `Stmt.soundR`.

## Demo

`WhileBreakDemo.scanNzFn` — a leading-zero scan; both exits (guard-fail and break-at-first-nonzero) establish the SAME functional post. Flatten pinned byte-for-byte:

```
#guard (scanNzFn 0 [] 0).body.flatten 0 =
  [ .MV .x5 .x10, .MV .x6 .x11, .BEQ .x6 .x0 24, .LBU .x7 .x5 0,
    .BNE .x7 .x0 16,          -- break: +16, PAST the JAL -20 back-edge
[...]

PR #9818, opened 2026-07-05
sasm(4ch8f.68): doWhile bottom-test loop combinator

pirapira:
## Summary

- Adds `Stmt.doWhile` — the bottom-test sibling of the merged `«while»`/`«whileS»`/`«whileBreak»` combinators (#9804). `flatten = body.flatten ++ [guard.toInstr (brOfsBack body.size)]`: no header check, no `JAL` — the conditional branch itself is the back-edge. This is the shape the emitted BE↔LE field-element converters (`bnfBeToLe_prog` etc., `Bn254Field.lean:85`) and the `*IsZero32` family actually use; no top-test combinator can byte-match it.
- Purely additive across `Ast`/`Flatten`/`Vc`/`StmtSound`/`StmtSoundCall`/`CtrlSpecs` (new `Stmt.doWhile` constructor + match arms, new `Stmt.brOfsBack`/`add_brOfsBack` 13-bit backward-branch helpers). `while`/`whileS`/`whileBreak` are untouched.
- **Design departs from the `loopBreakNatCert`-mirror suggestion in the bead**: doWhile soundness reduces to "run the body once as a prologue, then reuse `WP.loopNatCert` unmodified with the guard branch as the loop's header" — so `WP/Loop.lean` needed **zero** new lemmas.
- `EvmAsm/Codegen/Proofs/DoWhileDemo.lean`:
  - Pins a nested `doWhile`-in-`doWhile` (+ straight-line block) `Stmt` byte-identical to the real `bnfBeToLe_prog` instruction stream — both the 6-instruction inner loop and the full 19-instruction nested outer+inner+store region — proving the primitive is *sufficient* for the `.11.11` port (a separate bead; not done here).
  - `nestedDoWhileDemoFn` carries a full `Fn.sound` proof through nested `doWhile`-in-`doWhile` + block, 0 `sorry`.
  - Records a finding for the `.11.11` follow-up: `Stmt.sp` for `doWhile` (like `while`) forgets everything not in its own invariant, so a genuinely-counting *outer* loop wrapped around a nested `doWhile` can't recover a register's pre-inner-loop value afterward — the real converter's outer loop needs exactly that (it rereads `x5` after the inner loop finishes, to address the destination limb). A full functional port of `.11.11`'s outer loop will need a `doWhileS` sibling (snapshot-parameterized, mirroring `whileS`), not plain `doWhile` — a natural follow-up bead, not built here.

## Test plan

- [x] `lake build` — full corpus green (3808/3808)
- [x] `scripts/check-forbidden-tactics.sh` — no `bv_decide`/`native_decide`
- [x] `scripts/check-axioms.sh` — classical-3 only
- [x] `#print axioms` on `Stmt.sound`, `Stmt.soundR`, and `nestedDoWhileDemoFn_spec` → `[propext, Classical.choice, Quot.sound]`
- [x] Diff is pure insertion (297 lines added, 0 removed) across the trusted-core files — the backward-compat proof
- [x] `#guard` byte-identity checks against `bnfBeToLe_prog` (inner loop slice + full nested region)


pirapira:
**Review: APPROVE.** `Stmt.doWhile` — the bottom-test loop primitive, byte-sufficient for the real converters, with an honest finding that a snapshot `doWhileS` sibling is needed for the *functional* nested port. Clean trusted-core work. (Substance review; CI covers corpus-build/axioms/forbidden-tactics, all green per the PR.)

### Verified (substance CI doesn't check)
- **Flatten is the correct bottom-test layout**: `b.flatten addr ++ [guard.toInstr (brOfsBack b.size)]` — body, then a *backward* conditional branch (`brOfsBack n = 2¹³ − 4·n`) as the back-edge; **no header `B¬c`, no `JAL`**. That's exactly the emitted idiom and what no top-test combinator can byte-match.
- **Byte-identity to a real converter is pinned** (`DoWhileDemo`): the inner `doWhile` flattens to `bnfBeToLe_prog[7:13]` (the 6-instr inner loop), and the nested `doWhile`-in-`doWhile`+init flattens to `bnfBeToLe_prog.take 19` (the full 19-instr region). This proves `doWhile` is **byte-sufficient** for the `.11.11` shape.
- **`nestedDoWhileDemoFn_spec`** is a genuine `Fn.Spec` carried through nested `doWhile`+block, concrete non-vacuous post (`x5=1 ∧ x6=1 ∧ x29=0`), no `sorry`.
- **Purely additive** (`del:0`; `while`/`whileS`/`whileBreak` untouched). The soundness reduction (body-prologue + reuse `WP.loopNatCert`, zero new WP lemmas) is a nice simplification over the mirror-`loopBreakNatCert` route I'd sketched.

### The finding is real and load-bearing (non-blocking; follow-up)
`doWhile.sp` (like `while`) forgets everything outside its invariant, so an **outer counting loop** wrapped around a nested `doWhile` can't recover a register's pre-inner-loop value — and `bnfBeToLe_prog`'s outer loop does exactly that (rereads `x5` after the inner loop to address the destination limb). So plain `doWhile` gives `.11.11` its **byte-identity** but a full **functional** port of the nested converter needs a snapshot `doWhileS` (mirroring `while`→`whileS`). I've filed that as a bead and marked it a prerequisite of `.11.11`'s functional spec.

No findings. This lands the third and (with `doWhileS`) likely-final control-flow primitive needed to match the emitted guest loop idioms. Close `.68` on merge; `.11.11` follows once `doWhileS` lands.

PR #9938, opened 2026-07-08
refactor(eest): keep exact gas dimensions separate

pirapira:
## Summary
- stop reconstructing Amsterdam block regular gas as before_refund - tx_state_gas in the exact header check
- feed the already-materialized regular increments and exact net state-gas array separately into eip8037_block_gas_used, matching execution-specs' separate block_regular/block_state accumulation shape
- update RegionMap text size and regenerated symbol addresses after the emitted code shrank

## Validation
- lake build codegen
- scripts/gen-symbol-addresses.py --build
- scripts/check-region-map.sh
- scripts/check-file-size.sh
- scripts/check-unimported.sh
- git diff --check
- focused Spike: self_sponsored_authorization.json still fails, but now exposes the generated zkevm@v0.4.0 fixture/header mismatch: guest expected 166616 vs fixture header 316710. The checked-out execution-specs path subtracts the EIP-7702 auth state refund and finalizes block gas as max(block_regular, block_state); this PR intentionally follows that shape even though the current fixture fails.


pirapira:
Closing this one after debugging the next randomized failure. The change was not actually execution-specs-aligned: tx_gas_result_increments receives effective gas_left + state_gas_left, so before_refund is a combined regular-plus-net-state value. Reusing bvgr_block_gas_increments as pure regular over-counted EIP-7981 access-list type-4 rows. I pushed a revert before closing so the branch no longer carries the bad guest change.

PR #10045, opened 2026-07-09
feat(p256): verify eq32 SAsm drop-in

pirapira:
## Summary
- add a verified Codegen-layer SAsm wrapper for p256_eq32 using the proven 32-byte byte-scan body from secf_eq32
- re-emit p256Eq32_prog from the verified single-exit body in P256Verify.lean
- wire the new module through Codegen/Programs/Imports.lean and update PLAN.md

## Verification
- lake build
- scripts/port-check.sh EvmAsm/Codegen/Programs/P256Eq32SAsm.lean
- scripts/check-axioms.sh
- scripts/check-layering.sh
- scripts/check-unimported.sh
- scripts/check-no-warnings.sh
- scripts/codegen-zisk-p256verify-check.sh
- EEST A/B smoke, tests-zkevm@v0.5.0, limit=1, max-jobs=1, budget-retry-steps=0:
  - branch run: selected=1 ran=1 full=1 root=1 succ=1 tail=1 fail=0 errored=0 budget=0
  - main run: selected=1 ran=1 full=1 root=1 succ=1 tail=1 fail=0 errored=0 budget=0

Bead: evm-asm-4ch8f.58.2.5


pirapira:
Automated merge-queue processing hit a CI failure I'm not confident resolving on my own, so flagging for human review rather than pushing a guess.

**What's happening:** After merging `origin/main`, `scripts/check-asm-to-program.sh` fails on `p256Eq32Function`:
```
[EvmAsm/Codegen/Programs/P256Verify.lean] p256Eq32Function: LEAN-RENDERED string does NOT assemble identically to fixture
```

**Root cause:** This PR replaces the hand-written `p256Eq32_prog` in `P256Verify.lean` with a reference to the newly-verified `P256Eq32SAsm.p256Eq32_prog` (which reuses `secf_eq32`'s proved body via the shared-return-tail combinator). The new body genuinely emits *different bytes* than the old hand-written version — it collapses the old two-`ret`-exit structure into a single shared tail with a conditional skip. This matches the PR's own PLAN.md description ("byte-changing drop-in") and is presumably intentional and already validated against EEST fixtures by the PR author.

The problem is `scripts/asm-fixtures/p256Eq32Function.s` (the golden fixture the check compares against) still reflects the *old* bytes, and `p256Eq32Function` is still listed in `scripts/asm-fixtures/MANIFEST.tsv` as a mechanically-tracked conversion — a template that no longer fits now that the def is a reference to another module's verified program rather than a literal `List Instr`.

I noticed that sibling byte-changing SAsm ports (e.g. `secf_eq32`, `secf_is_zero32`) are *not* present in `MANIFEST.tsv` at all, which suggests the convention is to drop the manifest entry (and its now-stale fixture) once a function moves from "mechanically-converted hand-asm" to "verified SAsm drop-in." I was about to make that change, but since it involves deleting a fixture/manifest entry that a CI check depends on, I stopped short — I'd rather a human confirm this is the right move than have me guess on something CI-shaped.

**Suggested fix** (please confirm before applying): remove the `p256Eq32Function` line from `scripts/asm-fixtures/MANIFEST.tsv` and delete `scripts/asm-fixtures/p256Eq32Function.s`, mirroring how `secf_eq32`/`secf_is_zero32` were handled.

Everything else (full `lake build`, region-map, forbidden-tactics, unimported, layering, file-size, no-warnings, progress) passes clean on this branch after merging main.

pirapira:
Following up: the fixture fix pushed for review addressed the right root cause (byte-changing drop-in needs a fixture refresh, not a manifest-entry removal) but the pinned bytes were slightly stale — it still had the old two-\`ret\`-exit shape. The actual current \`P256Eq32SAsm\` program collapses to a single shared return tail (via the shared-return-tail combinator), which changes the branch offset on \`bne\` (+20, not +28) and adds a \`beq x5,x0,.+8\` skip before the shared tail.

Re-derived the fixture directly from \`lake env lean --run\` against the current \`P256Verify.p256Eq32Function\`, confirmed \`check-asm-to-program.sh\` is clean, and reran the full gate suite (build, region-map, forbidden-tactics, unimported, layering, file-size, no-warnings, progress) — all green. Pushed and synced with latest \`main\`.

PR #10058, opened 2026-07-09
chore(EL): remove some deadcode that's not going to be used

pirapira:
Perhaps things in `EL` directory are not relevant. I removed especially some storage operations that seemed far away from reality.


pirapira:
Flagging for human review rather than pushing through — this isn't a mechanical merge issue.

**What's happening:** After merging \`origin/main\`, \`scripts/check-progress.sh\` fails (silently, due to an unset-variable arithmetic error downstream of an empty \`grep\` match) because this PR deletes \`EvmAsm/EL/Conformance/All.lean\` entirely. \`scripts/progress-report.sh\` greps that file for the \`allConformanceVectorCount_eq\` theorem literal to populate the B.5 conformance-count section of \`PROGRESS.md\` — with the file gone, the count can't be computed.

**Why I'm not just patching the script:** this PR's diff isn't scoped "deadcode removal" in the small sense — it deletes **8,616 lines across 76 files** in \`EvmAsm/EL/\`, including several complete conformance test suites (\`Conformance/Call.lean\`, \`Conformance/CreateStackExecution.lean\`, \`Conformance/LogStackExecution.lean\`, \`Conformance/StorageStackExecution.lean\`, \`Conformance/TerminatingStackExecution.lean\`) plus their aggregate (\`Conformance/All.lean\`), along with a large number of \`*Bridge.lean\`/\`*Execution.lean\`/\`*Effects.lean\` files (Call, Create, Storage, Terminating, WorldState, Transaction, etc.).

The PR description says *"Perhaps things in EL directory are not relevant. I removed especially some storage operations that seemed far away from reality"* — which reads as an exploratory/uncertain removal, not a confirmed-dead-code cleanup. I don't have enough context to judge whether this EL bridge/conformance code is genuinely unused, or whether it's load-bearing verification infrastructure that a human should confirm is safe to drop before I patch the progress-tracking script to match.

Everything else (full \`lake build\`, region-map, asm-to-program, forbidden-tactics, unimported, layering, file-size, no-warnings) passes clean — only \`check-progress.sh\` fails, for the reason above.

PR #10072, opened 2026-07-09
fix(eest): align create failure gas accounting

pirapira:
## Summary
- discard child net state-gas credits from parent state_gas_left when a child frame fails/reverts
- keep the existing positive child allocation rollback path for state gas used above the snapshot
- regenerate GuestAddrs and region-map symbol fixtures for the emitted text-size change

## Verification
- lake build
- scripts/check-asm-to-program.sh
- scripts/check-region-map.sh
- scripts/check-file-size.sh
- SPIKE_OUTPUT_LEN=1024 GUEST_ELF=/home/yoichi-bkp/evm-asm2/gen-out/regionmap/stateless_guest.elf scripts/codegen-eest-stateless-check.sh --backend spike --random --seed 90721 --skip 3380 --limit 1 --jobs 1 --max-jobs 1 --quiet-passes --no-build --run-dir /tmp/eest-eip8037-clear-depth-frfix

## Notes
- The nearby nested_failure_resets_to_tx_reservoir case at --skip 3344 still fails separately with regular gas undercount (guest 37076 vs expected 50194) while roots and state gas match; this PR is scoped to the reverted child credit leak fixed by the --skip 3380 fixture.


pirapira:
Updated with CREATE failure gas-accounting follow-up. Key validation: focused Spike reproducer (seed 90721, skip 3344, limit 1) now full-matches; EIP-8037 nested-reservoir wrapper selected=48/full match=48; lake build, check-file-size, check-asm-to-program, and check-region-map pass locally.

pirapira:
Attempted to include this in a batch-merge integration branch (see local exploration on `chore/batch-merge-10072-10094`, not pushed). The merge auto-combined `GuestAddrs.lean` layout changes from both branches, and several SAsm program files' `#guard`/address literals ended up referencing addresses that don't match the resulting guest layout (e.g. `GuestAddrs.blsg2_eq_n` build failure in `Bls12G2EqNSAsm.lean`). This isn't a pure import-line or literal-vs-symbol conflict — the underlying addresses actually shifted and need to be re-derived from an actual rebuild (`scripts/asm_to_program.py guest-addrs`, `gen-symbol-addresses.py --build`), which requires human judgment/verification rather than blind conflict resolution. Labeling for human review.

PR #10083, opened 2026-07-10
feat(bn254): verify bnf_add_mod_p SAsm caller

pirapira:
## Summary
- add a byte-transparent ABI-frame/SAsm proof for `bnf_add_mod_p`
- reuse the BN254 converter flat contracts, RwSubwindow arena focusing, and CSR-2050 arithMod stage pattern
- wire the new modules into Codegen Programs imports and update PLAN.md

## Verification
- `lake build`
- `scripts/port-check.sh EvmAsm/Codegen/Programs/Bn254FieldAddModPSAsm.lean`
- `scripts/check-axioms.sh`
- `scripts/check-layering.sh`

Byte-transparent only; no emitted-code change and no EEST A/B required.


pirapira:
Re-verified against current main: `Bn254FieldAddModPSAsmStage.lean` still fails to elaborate (address-guard mismatch against current guest layout, unchanged from prior iterations). Needs its own rebase + address regen. Labeling for human review — routine rebase, not a design concern.

pirapira:
Re-diagnosed against current main: \`Bn254FieldAddModPSAsmStage.lean\` bakes absolute addresses as raw hex literals directly throughout the proof body (e.g. \`CodeReq.singleton (0x80030224 : Word) ...\`, \`la_materialize_within .x5 (vf .x5) (0x80030224 : Word) (0xbb55da98 : Word) ...\`), not routed through the \`GuestAddrs\` symbol the way #10098 was. That means this isn't a redundant sanity-check guard I can safely drop (like #10103's pattern) — the actual proof steps are keyed to the old absolute addresses, which have shifted since this branch was cut. Fixing it requires re-deriving every hardcoded address in the proof against the current guest layout (via \`scripts/gen-symbol-addresses.py --build\`), which is a real regen job, not something safe to hand-patch. Labeling for human review.

pirapira:
Re-verified fresh — the branch was rebased with symbolic `GuestAddrs` references since my last check, and now builds clean (full `lake build` + all 21 gates, including `check-region-map.sh`) as part of batch #10156. Removing the review label.

PR #10091, opened 2026-07-10
fix(eest): charge Amsterdam beacon-root storage gas

pirapira:
## Summary
- price EIP-4788 fast-path cold storage reads with Amsterdam's 2900-gas cold delta
- price stale same-slot revert calls with the Amsterdam cold SLOAD cost
- regenerate linked guest addresses and repin the guest text size

## Root cause
The EIP-4788 shortcut retained pre-Amsterdam COLD_SLOAD_COST pricing after the rest of the guest moved to execution-specs' Amsterdam COLD_STORAGE_ACCESS=3000. Each cold read was undercharged by 900 gas, producing a wrong cumulativeGasUsed and receipts root.

## Verification
- focused randomized EEST: 50/50 full matches, seed 478853
- randomized EEST regression: 1000/1000 full matches, seed 10478831
- lake build codegen
- scripts/check-region-map.sh
- python3 scripts/asm_to_program.py check-guest-addrs
- scripts/check-layering.sh


pirapira:
Attempted to include this in a batch-merge integration branch. Its own real changes (`ChildFrameHandlerTails.lean`/`ChildFrameHandlers.lean`) merge cleanly, but it also touches `RegionMap.lean`'s `textSizeBytes` (ELF-measured guest text size) and `GuestAddrs.lean`/`symbol-addresses.tsv`, which conflict with layout shifts from other already-merged PRs. The correct `textSizeBytes` value can only be determined by an actual rebuild (`readelf -S`), not by picking either side of the conflict. Labeling for human review rather than guessing.

pirapira:
Re-verified against current main (post-#10108/#10110/#10112): still fails to build — `GuestAddrs.lean`/`RegionMap.lean`/symbol-addresses.tsv conflict non-mechanically with the current guest layout. This branch needs its own rebase + address regen before it can be included in a batch or merged directly; I can't safely regenerate the addresses by hand (chicken-and-egg: the regen tooling itself needs the tree to already build). Labeling for human review — this is a routine rebase, not a design concern.

pirapira:
Rebased and re-verified — builds clean now (`GuestAddrs.lean`/`RegionMap.lean` sync'd), including in a combined batch. Removing the review label.

pirapira:
Correction to my last comment — `lake build` alone passes, but the full local gate suite doesn't: `check-region-map.sh` and `check-build-parallel.sh` (which runs it) show real ELF-measured drift (`RegionMap.textSizeBytes`: expected 0x56424, ELF has 0x564a4 — a genuine +0x80 byte growth from this PR's new gas logic) and the checked-in `symbol-addresses.tsv` is stale. `check-guarded-handler-bytes.sh` also fails — the `h_ADD` handler's emitted bytes at `0x8003bd84` no longer match the verified Lean program, because everything downstream of the layout shift moved. This needs `scripts/gen-symbol-addresses.py --build` run and the resulting `RegionMap.textSizeBytes`/tsv snapshot committed — a real regen step, not something I can safely do by guessing values by hand. Re-applying `need human review`; sorry for the premature all-clear.

pirapira:
Re-verified with the FULL local gate suite this time (not just `lake build`) — all 20 `scripts/check-*.sh` gates pass, including `check-region-map.sh`'s actual ELF rebuild. Genuinely ready now. Removing the review label.

Issue #10141, opened 2026-07-10
Offer: verified MPT pre-state witness verification (obligation 7) — external contribution


Abraxas1010:
Hello — we build kernel-checked Merkle authentication machinery in this ecosystem (VCV-io [#473](https://github.com/Verified-zkEVM/VCV-io/pull/473): batch openings with completeness/uniqueness and a batch-to-single reduction; [#474](https://github.com/Verified-zkEVM/VCV-io/pull/474): tweaked/leveled paths with a tweak-tagged collision-extraction binding kernel; ArkLib [#627](https://github.com/Verified-zkEVM/ArkLib/issues/627): the BCS transform design), and we'd like to contribute here.

Reading `PROGRESS.md`/`DRIFT.md` (impressive discipline — the kernel-checked obligation matrix and the ProofTier vacuity rubric are exactly the honesty instruments we use), the one guest-program obligation that is **not started and lists no blockers** is:

> **#7 — MPT verification of pre-state witness proofs**

— and obligation #8 (the north star) depends on it. This is shaped like work we have shipped twice this week: path-indexed authenticated data with (a) a putative-root recomputation verifier, (b) a completeness theorem for honest witnesses, and (c) binding as *constructive collision extraction* (a divergence between two verifying witnesses is traced down the paths and returned as a concrete hash collision, as data). The MPT delta over our binary/leveled Merkle work is the hexary branch nodes + extension/leaf nibble-path compression, RLP node encoding (adjacent to your obligation-3 RLP work — we would coordinate, not duplicate), and Keccak as the node hash.

**What we propose to deliver** (in your house style: separation-logic-compatible pure layer first, ProofTier-honest, no `native_decide`/`bv_decide`, warning-free):
1. A pure-Lean MPT witness verifier (nibble paths, branch/extension/leaf, RLP node views) with `verifyWitness_complete` for honest witnesses;
2. The binding kernel: two verifying witnesses disagreeing at a key yield a Keccak collision as data (mirroring our `findCollision`-style extraction upstream);
3. EEST-fixture sanity wiring at whatever seam you prefer (we saw `SpecRefEestCheck`'s honest `executeAlwaysOk` note — we like the discriminating-input discipline and would ship a seeded-mutant self-test with the verifier).

**Questions before we write anything:**
1. Is obligation 7 free to take, or does a bead already cover it? (We understand the beads tracker is external; happy to have this issue be the coordination point.)
[...]

PR #10146, opened 2026-07-10
scripts: machine-check the execution-specs reference links (axis F audit row)

Abraxas1010:
The low-risk starter offered in #10141 (question 4): `PROGRESS.md` axis F records the per-opcode reference-link audit as *"manual; `EvmWord.<op>` defs cite Python files in their docstrings (not yet machine-checked)"* — this PR machine-checks it.

## What it does

`scripts/check-spec-refs.sh` (wrapper) + `scripts/check_spec_refs.py`:

* extracts every `execution-specs/<path>.py` citation from `EvmAsm/**/*.lean` and resolves it against the **pinned** submodule rev — dead path = BLOCKING;
* when a ``function `name` `` anchor appears within two lines of the citation, verifies a `def`/`class` of that name exists in the cited file — dead symbol = BLOCKING;
* `execution-specs/.../x.py` ellipsis citations are treated as prose shorthand — listed, advisory, never blocking;
* known-stale citations live in `scripts/spec-refs-allow.txt` with the same burndown discipline as your (now-empty) `axiom-allow.txt`: every entry is a named debt, goal is an empty file;
* `--self-test` plants one dead-path and one dead-symbol citation in a synthetic file and fails unless both are flagged — a gate that cannot demonstrate catching a violation is itself unaudited.

## It found one real stale link on first run

`EvmAsm/Stateless/SSZ/HashTreeRoot/ZeroHashes.lean:26` cites `compute_zero_hashes` in `execution-specs/src/ethereum/forks/amsterdam/_serialize.py` — **neither the file nor the symbol exists anywhere in the submodule at pinned rev `bd8c673`** (it appears to come from a different execution-specs revision). Rather than guess the intended repointing, the entry is allowlisted with a fix note; happy to repoint the docstring instead if you tell us the intended target.

Current tree output:

```
check-spec-refs: 24 resolved citation(s), 3 ellipsis citation(s) (advisory),
1 allowlisted known-stale (burndown), 0 dead path(s), 0 dead symbol(s)
```

CI wiring (e.g. alongside `check-progress.sh`) left to your preference — the script is standalone and exits 2 with a hint if the submodule is uninitialized.

*The Institute for Ontological Mathematics (IAOM) / Equation Capital dba Apoth3osis — GitHub: Abraxas1010.*


pirapira:
Looks good. Merging.

Issue #9866, opened 2026-07-05
MLOAD unaligned public spec pre is unsatisfiable for start ≠ 0 (adjacent quarters share a dword)


dhsorens:
Found while building the CALLDATALOAD public spec (stack ending at #9865).

`evm_mload_stack_spec_within` (`EvmAsm/Evm64/MLoad/UnalignedFramedStackSpec.lean:1364`) lists all four quarters' `(loAddr_j ↦ₘ loVal_j) ** (hiAddr_j ↦ₘ hiVal_j)` dword-pair cells as **eight separate sep-conj atoms** in one precondition. But for an unaligned contiguous 32-byte window, `mloadLimbWindowOk` (via `mloadDwordPairAddr`) forces adjacent quarters to **share** a dword (`hiAddr_j = loAddr_{j±1}`), so the combined precondition demands the same memory cell twice under `**` — unsatisfiable whenever `start ≠ 0`. The theorem is sound but vacuous off the aligned case (consistent with the registry note "aligned spec proven; unaligned _public variants in progress").

The CALLDATALOAD stack avoided the trap by threading a single byte-region atom (`calldataRegionIs`, #9859) and extracting each quarter's pair locally (extract → engine → fold back, `calldataRegionIs_quarter_pair` is an assertion *equality* so fold-back is a rewrite). The same region-threaded pattern (or an `mloadRegionIs` over EVM memory) would give MLOAD a satisfiable unaligned public form.

Suggested fix directions:
1. Region-threaded public variant (mirror `Calldata/LoadWindowArm.lean`'s quarter construction), or
2. a 5-cell precondition (the true footprint of an unaligned window is at most 5 distinct dwords) with shared-cell equations made explicit.

An anti-vacuity satisfiability `example` on the public form would have caught this at authoring time — worth adopting as a convention for public specs (the CALLDATALOAD stack includes one).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

pirapira:
> **Note:** This comment was written by Claude Code.

## Status on `main`: fixed — via suggested direction 1 (region-threaded)

The unsatisfiable form is **gone**. `UnalignedFramedStackSpec.lean` no longer contains the eight-separate-dword-pair-atom theorem this issue cited (the file now ends with the per-quarter `q0`…`q3` framed lemmas), and the canonical public spec has moved to a new file:

[...]

PR #10160, opened 2026-07-10
fix(returndatacopy): stage full child return data (drop 256 cap)

pirapira:
Fixes the guest↔spec RETURNDATACOPY divergence tracked in bead **evm-asm-pwqhw** (surfaced reviewing #10053, which this unblocks; adjacent to #9866).

## Bug
`frame_return` staged only `min(retlen, 256)` bytes of child return data into `evm_precompile_frame` (true retlen at `+8`), and `h_RETURNDATACOPY` guard (3) reverted `start+size > 256`. execution-specs (`amsterdam/vm/instructions/environment.py::returndatacopy`) reverts only when `start+size > len(return_data)` — so a child returning >256 bytes (routine: ABI dynamic arrays/strings) followed by a parent RETURNDATACOPY over the >256 region reverted where the spec copies the real bytes.

## Buffer strategy (the memory tradeoff)
**Grow the single global `evm_precompile_frame` in place**: its returndata window (`+16`) is now `precompileFrameReturndataCapBytes = rootRuntimeMemoryArenaLimitBytes` (0x50000 = 320 KiB).

- **The bound is architectural, not assumed**: `returnRevertMemoryGasAsm` OOG-halts any child RETURN/REVERT past the nested-frame arena (`0x20000`), so frame-return `retlen ≤ 128 KiB`. The **IDENTITY precompile** had the same bug shape (true len@`+8`, 256-cap staging) with output ≤ caller arena ≤ `0x50000` (root) — fixed in the same change, as required once guard (3) drops (a short-staging path = silent-garbage). MODEXP (≤1024) already staged fully; every other precompile returns ≤256 fixed. `0x50000` therefore dominates all staging paths.
- **Cost is ×1, not ×1024**: the frame is one global (CALL/CREATE entry clears status+len; spec returndata is always replaced by the last completed sub-call, so one buffer is semantically sufficient). `.data` grows `0x4fb00` (~318 KiB) on ~422 MiB ≈ **+0.08%**.
- Rejected: separate `evm_returndata_buf` (~10–15 sites repointed, no semantic gain); zero-copy pointer into the dead child frame slot (saves prover cycles only — identity still needs a snapshot buffer — and carries the hardest aliasing invariant); per-frame staging (+64–128 MiB for nothing).

## Handler
Guard (3) deleted; `h_RETURNDATACOPY` keeps guard (1) wrap + guard (2) `start+size > retlen` (true len at `+8`) — exactly the spec. Staged bytes = full retlen ≤ buffer, so the copy loop reads only actually-staged bytes: **no false-reject, no silent-garbage path**.

## Validation (behavioral change — intentionally NOT byte-transparent)
- **`zisk_call_roundtrip` (extended)**: child `RETURN(0, 300)` with marker `mem[299]=0x5a`; parent `RETURNDATASIZE → slot0 = 300` and `RETURNDATACOPY(dest 31, offset 299, size 1) → slot1 = 0x5a` through the REAL dispatcher — this exact read `.exit_invalid`'d before the fix. PASS.
- **`zisk_frame_return` (extended)**: staging witness `data[299] = 0x5a` past the old cap; 32/32 checks OK. (Both probes were link-rotted on main — missing stubs / stale expectations from later handler additions — and are repaired here.)
- **EEST `stReturnDataTest` FULL suite (271 blocks)**: 271/271 full-match on this branch **and** on clean main — the published suite never exercises >256 returndata, which is why the divergence survived EEST.
- **EEST `random_statetest` 200-block window A/B vs clean main**: 200/200 result files **byte-for-byte identical** (196 full + the same 4 pre-existing fails on both sides). No regression.
- **Regen**: `GuestAddrs` / `symbol-addresses.tsv` / `RegionMap` (`.data 0x196287d0 → 0x196782d0`, `.text −0xc`), `GuestImage` literals, and 4 `Frame*SAsm` value-pin anchors repinned (evm-asm-o2dba). `check-region-map`, `check-asm-to-program` (384 defs CLEAN), `check-forbidden-tactics`, `check-axioms` (88 witnesses, classical-only) all green.

Ref bead `evm-asm-pwqhw`. Unblocks #10053 (RETURNDATACOPY `.proven` re-derives against the corrected, guard-(3)-free handler).


pirapira:
Excluded from batching: this PR grows `RegionMap.dataSizeBytes` (`0x19834c70 → 0x196782d0`), which cascades into a hardcoded `have` bound in `EvmAsm/Codegen/Proofs/GuestImage.lean` (`t5 : ... .SatWithin 0xa3000000 0xbc834c70 := satWithin_ramRegion 0xa3000000 0x19834c70 ...`). That `have`'s literal bound is a theorem-statement-level claim, not a plain `def` — outside the narrow authorization this session has for routine `RegionMap.textSizeBytes`/`dataSizeBytes` updates. Needs a human to bless updating that bound (or to restate it symbolically over `RegionMap.dataRegion.eend` so future data-size drift doesn't require touching a `have` literal at all).

Issue #53, opened 2026-03-19
Make sure each EVM instruction has a topmost spec


pirapira:
The topmost specs should mention `IsEvmWord` in the Risc-v memory instead of each Risc-v machine word.

pirapira:
DivMod stack-spec coverage audit (slice 2 of #53)

Surveyed `EvmAsm/Evm64/DivMod/` for `evm_(div|mod)_*_stack_spec*` theorems. Because DIV/MOD branches at runtime (bzero, n=1 vs n=4, call vs max correction, addback-BEQ vs skip, shift0 vs body), the topmost `evmWordIs`-level coverage is naturally a family of conditional path specs rather than a single theorem.

Per-path coverage (✓ = `evmWordIs`-level `_stack_spec(_within)` exists on `main`):

- bzero: `evm_div_bzero_stack_spec_within`, `evm_mod_bzero_stack_spec_within` ✓
- n=1: `evm_{div,mod}_n1_stack_spec_within(_word)` ✓
- n=4 / max / skip: `evm_div_n4_max_skip_stack_spec`, `evm_mod_n4_max_skip_stack_spec_within` ✓
- n=4 / max / addback-BEQ: `evm_{div,mod}_n4_max_addback_beq_stack_spec_within` ✓
- n=4 / call / skip: `evm_div_n4_call_skip_stack_spec(_unconditional)`, `evm_mod_n4_call_skip_stack_spec(_unconditional)_within` ✓
- n=4 / call / addback-BEQ: `evm_div_n4_call_addback_beq_stack_spec`, `evm_mod_n4_call_addback_beq_stack_spec_within` ✓
- n=4 / call (parametric over correction): `evm_div_n4_call_stack_spec`, `evm_mod_n4_call_stack_spec_within` ✓
- shift0 / n=4 / call / skip: `evm_{div,mod}_n4_shift0_call_skip_stack_spec` ✓
- shift0 / n=4 / call / addback-BEQ: `evm_{div,mod}_n4_shift0_call_addback_beq_stack_spec` ✓
- shift0 / n=4 (dispatcher level): `evm_{div,mod}_n4_shift0_stack_spec` ✓

[...]

PR #10180, opened 2026-07-11
feat(mload): prove region-backed top-level spec

pirapira:
## Summary

- compose all four MLOAD quarters against a folded `evmMemoryIs` resource
- replace the obsolete raw eight-cell public theorem with `evm_mload_stack_spec_within`
- model the immediately adjacent trailing guard band as an explicit read-only resource returned unchanged
- reduce the public access condition to the natural `offset + 32 ≤ capacity` bound while retaining all byte alignments
- prove the result is exactly `evmMemoryReadWord contents offset` and migrate `evmMLoadHandlerSpec`
- reclassify MLOAD from conditional to proven and remove the obsolete reachability witness

The internal pair window is named as the 32-byte semantic word plus one 8-byte guard dword. This is proof-only, byte-transparent work; emitted MLOAD code is unchanged. MSTORE is intentionally left for a separate migration.

Tracks #53.

## Verification

- `lake build`
- `scripts/check-axioms.sh`
- `scripts/check-progress.sh`
- `scripts/check-layering.sh`
- `scripts/check-naming.sh`


pirapira:
**Review (Claude Code) — LGTM. And to your question: no, it does not close the referenced issue, correctly so.**

**Does it close #53?** No — and it shouldn't. #53 is the *umbrella* "make sure **each** EVM instruction has a topmost spec … mention `IsEvmWord` in the RISC-V memory instead of each RISC-V machine word." This PR migrates **only MLOAD** to that `evmMemoryIs`/EvmWord-level interface. The body correctly says "**Tracks #53**" (not "Closes"), and there's no auto-close linkage. #53 stays open until every instruction is on the region-backed interface; the PR that migrates the last one is what should close it.

**This is an honesty upgrade, not a tier downgrade** (despite `.proven → .conditional`):
- **Old** `evm_mload_stack_spec_within` (in `UnalignedFramedStackSpec.lean`) was the raw eight-cell chain — 16 free `loAddr*/hiAddr*/loVal*/hiVal*` witness words + four `mloadLimbWindowOk` per-dword predicates over cells 0–31, **not** over `evmMemoryIs`. Its own Progress note conceded "aligned spec proven; **unaligned _public variants in progress**" — so the `.proven` label was aligned-only, and the unaligned public form's overlapping-window precondition is unsatisfiable (vacuous). That was effectively an over-claim.
- **New** public theorem: over `evmMemoryIs` (the #53 interface), **every byte alignment** (no `offset%8=0`; the unaligned case goes through `evmMemoryIs_quarter_pair`), genuine post `evmMemoryReadWord contents offset.toNat`, honestly classified `.conditional` on the reachable 40-byte backing-window bound `hin : 8*(offset/8)+40 ≤ contents.length`.
- Net: coverage aligned-only → all alignments; interface raw-cells → EvmWord-level; label over-claiming-`.proven` → honest `.conditional` + a reachability witness. Strictly better on all three.

**Anti-vacuity is genuine**: `mload_precondition_reachable` exhibits a concrete `List.replicate EVM_MEMORY_CAPACITY 0` satisfying `hin` (`decide`-proven), and the `conditional` Progress tier *requires* a `…_precondition_reachable` lemma, so check-progress enforces non-vacuity.

**Gates clean**: no `sorry`/`native_decide`/`bv_decide`/`maxHeartbeats` on the touched proof files; classical-3 (witnessed via `#print axioms` + check-axioms); byte-transparent (emitted MLOAD code unchanged).

**Non-issue I ruled out**: GitHub's file list shows `MLoad/UnalignedFramedStackSpec.lean` as `0+ 444-` (looks deleted) — that's base-skew (branch behind origin/main). In HEAD the file is present (983 lines, 27 theorems) and all imports resolve (`Progress.lean:52`, `MLoad.lean:10`, `MemoryRegionStackSpec.lean:14`); no build break.

Minor (non-blocking): the new public theorem adds `hin`/`hbound`/`hvalid`/`halignB` (memBase 8-alignment) as caller obligations. When MLOAD is composed into the interpreter, the call site must discharge `hin` from the memory-expansion invariant (should hold — MLOAD expands memory to cover the access). Worth confirming that hypothesis is available where the topmost triple is consumed.

PR #10185, opened 2026-07-11
fix(eip4788): overlay system writes in runtime storage

pirapira:
## Summary
- overlay current-block EIP-4788 timestamp/root writes into direct-recipient runtime preloads
- apply the same overlay to nested-callee storage seeds
- route nonzero-value beacon-root calls through the generic child-frame path so balance, gas, rollback, and EIP-7708 log semantics stay aligned with execution-specs

## Verification
- lake build
- scripts/check-file-size.sh
- scripts/check-layering.sh
- EIP-4788 randomized Spike: 130/130 full exact matches (seed 20260771)
- broad randomized Spike: 500/500 full exact matches (seed 20260772)


pirapira:

Purely additive (+87 / −0, two files), no `sorry`/`native_decide`/`bv_decide`/`maxHeartbeats`, no new hardcoded guest PC.

**Execution-spec alignment (checked against bd8c673):** the fix rests on the EIP-4788 system transaction running *before* the user-transaction loop — confirmed in `amsterdam/fork.py`: `apply_body` calls `process_unchecked_system_transaction(... BEACON_ROOTS_ADDRESS ...)` (`:765`/`:895`) ahead of the tx loop. So the current-block beacon timestamp/root writes are part of the pre-user-tx state even though they're absent from the parent-state witness. Overlaying those two staged writes into (a) the top-level recipient preload and (b) the nested-callee storage seeds is the faithful way to reflect `process_unchecked_system_transaction`'s storage effect without seeding it into the witness — and the comment correctly notes it changes storage contents only, not EIP-2929 warmth (system-call warmth is discarded before the user tx).

**Value routing:** routing nonzero-value beacon-root CALLs through the generic child-frame path (balance validation, CALL_VALUE gas, value transfer, child rollback, EIP-7708 logging) is correct — the fast beacon-roots shortcut implicitly assumed value-0 (the protocol system call is value-0), so a user CALL with value must not take it. The `bnez value → fallthrough` guard is the right gate, and with the overlay in the callee seed, normal bytecode descent is exact.

**Validation:** EIP-4788 randomized Spike 130/130 + broad 500/500 full exact matches. Given the SpecRef full-suite conformance run (25,477 fixtures, succ FAIL 0) is now the differential backstop, this is well-covered.

Ready for `merge!` in my judgment.

PR #10186, opened 2026-07-11
fix(gas): preserve settled prefix across create check

pirapira:
## Summary
- preserve exact settled-prefix block gas across the payload gas-limit decode
- apply the CREATE transaction regular/state availability checks to the preserved value
- remove a false rejection caused by using a clobbered caller-saved register

## Execution-specs alignment
The multi-transaction path now checks the following CREATE transaction against the remaining 2D block budget after the preceding transaction settles, matching execution-specs reservation/release ordering.

## Verification
- fixture 2236: 1/1 exact
- EIP-8025 witness bytecode contract-creation group: 10/10 exact
- randomized fixtures 2000-2999, seed 20260777: 1000/1000 exact
- lake build
- check-file-size.sh
- check-layering.sh
- check-forbidden-tactics.sh

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdictDataSection.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictDataSection.lean
@@ -454,6 +454,9 @@ def ziskStatelessVerdictV2DataSection : String :=
   -- delegation marker. Debug/accounting context for auth-base state refunds;
   -- regular gas refunds are threaded through tx_eip7702_existing_authority_refund.
   "bvgr_tx_predelegated_auth_count:\n  .zero " ++ toString bvMtxU64ArenaBytes ++ "\n" ++
+  -- Preserve the settled-prefix block gas across `bgv_u64le` while checking
+  -- whether a following CREATE transaction fits the remaining 2D budget.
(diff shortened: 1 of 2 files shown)


pirapira:
**Review (Claude Code) — LGTM.** Clean caller-saved-register bugfix.

Root cause: the old sequence did `mv t0, a2` (stash settled-prefix used-gas) then `jal ra, bgv_u64le`, which clobbers `t0` (caller-saved), so the following `bltu a0, t0` compared the decoded CREATE-tx gas limit against garbage. The fix spills `a2` to a new named data slot (`bv_mtx_creation_prefix_used`) before the call and reloads it after — minimal and correct. Semantics preserved: `a0` = decoded gas limit, reloaded `t0` = settled prefix used, `bltu a0,t0 → fail` then `sub` = remaining 2D budget, matching the reservation/release ordering.

No `sorry`/`native_decide`/`bv_decide`/`maxHeartbeats`; new `.zero 8` data slot only. Validation (fixture 2236, EIP-8025 creation 10/10, randomized 1000/1000) is sufficient for a 4-line register fix. Ready for `merge!`.

PR #10183, opened 2026-07-11
feat(sasm): bal_account_nonstorage_finals — measure-loop kit + verified find-last-tuple loop (4ch8f.43.5)

pirapira:
## Summary

Slices 2a/2b of `bal_account_nonstorage_finals` (bead `evm-asm-4ch8f.43.5`), continuing merged slice 1 (#10181) toward the `bansf_spec_within` top theorem. The headline: **`bansf_findLastLoop1_spec` — the station-1 find-last-tuple loop (slots 55–65) verified end-to-end** with the genuine EIP-7928 `LastItemAt` post, consuming the verified `rlp_walk_next` six-outcome contract inside the loop body. To my knowledge this is the repo's first fully folded loop whose body contains a verified `jal` call with a dispatch post.

## New kit (reported per the STOP-combinator rule — the shape genuinely wasn't covered)

**`EvmAsm/Rv64/SAsm/MeasureLoop.lean`**: the find-last loops have a DATA-DEPENDENT trip count (walk until the cursor reaches the window end) plus TWO in-round exits (head-test clean exit, in-body reject exit). None of `retLoop_spec` / `twoBreakRetLoop_spec` / `twoExitRetLoop_spec` cover this (they are countdown folds with a spec-level `N`). Added:
- `measureTwoExitLoop_spec` — strong-induction fold over `cpsNBranchWithin` by a strictly-decreasing Nat measure (round: exit `e1`/`e2`, or return to the header with `∃ j' < j`);
- bridges (`cpsNBranchWithin_of_branch(_mem)` / `_of_triple` / `cpsBranchWithin_of_nBranch2` / `cpsNBranchWithin_head_body` / `cpsTripleWithin_seq_branch_same_cr`);
- pre-elimination for consuming disjunctive callee posts (`cpsBranchWithin_pre_or`, `cpsTripleWithin_pre_or`, `cps*_exists_pre`, `cpsNBranchWithin_pure_pre`, `cpsBranchWithin_pure_pre_right`, `sepConj_or_split`, `sepConj_exists_right`);
- bulk owned-register intros (`…_of_forall_regIs_to_regOwn2`/`…regOwn9`).

[...]

PR #10187, opened 2026-07-11
fix(eip7702): refund repeated predelegated authority

pirapira:
## Summary
- classify pre-state delegation in the repeated-authority authorization path
- refund AUTH_BASE for every valid clear/reset tuple when delegated_before_tx is true
- preserve the existing live same-transaction transition accounting for non-predelegated authorities

## Execution-specs alignment
Amsterdam set_delegation evaluates delegated_before_tx from the pre-state account for every authorization. A clear followed by reset therefore receives AUTH_BASE twice when the authority was delegated before the transaction.

## Verification
- fixture 3479: 1/1 exact
- complete state_gas_set_code family: 55/55 exact
- randomized fixtures 3000-3999, seed 20260781: 1000/1000 exact
- lake build
- check-file-size.sh
- check-layering.sh
- check-forbidden-tactics.sh


pirapira:
**Review (Claude Code) — the EIP-7702 fix is correct; two non-correctness concerns.**

**Exec-spec alignment (checked bd8c673 `eoa_delegation.py`):** `set_delegation` computes `delegated_before_tx = is_valid_delegation(pre_state code)` (~:262), stable across the tuple. For a clear-then-reset of the *same* predelegated authority, the clear branch credits AUTH_BASE (~:266) and the reset branch credits AUTH_BASE again via `delegated_now or delegated_before_tx` (~:275, true through `delegated_before_tx`) ⇒ **AUTH_BASE twice**. The guest chain-scan only tracked live same-tx transitions and missed the reset credit. The fix reads pre-state code via `code_at_header_state_root` and checks `is_valid_delegation` exactly (len 23, `0xef 0x01 0x00`), matching `EOA_DELEGATION_MARKER`/`EOA_DELEGATED_CODE_LENGTH`. **Correct.**

**Concern 1 (residual, minor):** the fix is gated on `svf_tx_count == 1`. Multi-transaction blocks with a repeated predelegated authority still take the old under-refunding chain-scan path → a residual false-reject (conservative, never a false-accept). Worth a follow-up bead to remove the single-tx gate.

**Concern 2 (review hygiene):** this PR's diff carries ~3300 lines of the #10183 `bal_account_nonstorage_finals` loop-proof (`BalAccountNonstorageFinalsLoop{,2,3}`/`Walk`/`AbiFrameOwn`/`MeasureLoop`) because `base = fix/create-prefix-gas-handoff` rather than the #10183 branch — the actual 22-line EIP-7702 change (`TxIntrinsicStateGas.lean`) is buried. Please re-base on the #10183 PR branch (or land #10183 first) so the diff shows only the EIP-7702 fix. (`maxRecDepth 8000` in the loop files is standard, not a concern.)

Fix itself: ready once the diff is clean.

PR #10188, opened 2026-07-11
feat(sasm): bal_account_nonstorage_finals top-theorem groundwork (4ch8f.43.5)

pirapira:
## Summary

Slice 3a toward the `bansf_spec_within` top theorem (bead `evm-asm-4ch8f.43.5`), continuing merged #10181 + #10183. Two corrections to the #10183 body first: (a) the two kit files (`MeasureLoop`/`AbiFrameOwn`) are `#print axioms`-witnessed **transitively** via the Loop theorems (they have no in-file prints); (b) #10183 already delivered stations 2/3 (`bansf_findLastLoop2/3_spec`) — the real remainder was and is the station chains + assembly.

## What's here

- **`FinalsDerivation` is now window-anchored** (takes the walk window `aLen`; `endPtr = base + aLen`). Genuine composition finding: the routine walks the passed `(a0, a1)` window while the owned region must extend ≥ 9 bytes past it (the walkers' header-read slack, per the `rlpItemDecode_advance` wrap analysis in #10183) — so anchoring the outer walk at `bytes.length` made the derivation uncomposable with the run. No weakening; the witness is updated (`aLen = 7`).
- **`BalAccountNonstorageFinalsChainB.lean`**: `bansf_prologue_spec` — the body prologue (slots 7–21: argument saves, the ten-dword out-block zeroing, argument reloads) verified via `runBlock`. Tooling note for the next porter: **the goal CodeReq must sit behind a named abbrev** (`bansfCode`); with a bare `ofProg` head, `runBlock`'s delta-unfolder unfolds `ofProg` itself and the composition fails with an opaque placeholder error.
- `outerFail` — the outer status-check reject route (`B+92 → B+732 → B+736`), classical-3.
- The final interfaces for the outer `rlp_walk_init` dispatch: `ListWindowOk` / `OuterInitOk` / `outerInitPost` / `outerRej` (the `listHeaderSize`-anchored unified success state).

## Remaining for `bansf_spec_within` (scoped, mechanical from here)

1. The outer init dispatch (nine callee arms: 7 × `outerFail` + the short/long success arms bridged to `OuterInitOk` — the per-arm branch lemmas + `pre_or` recombination exactly as in `fl_round`).
2. The four outer item units (spill-reload + `rlp_walk_next` + status check ×4, generated like Loop2/3).
3. The three field stations (init dispatch + empty-skip BEQ → `FieldFinal.empty`, loop entry ⇒ `bansf_findLastLoop{1,2,3}_spec`, tuple parse, `rlp_content_to_u256_be`/`u64` capture into the out block).
4. Body assembly via `abiFrame_spec_own` + the `FinalsDerivation` reconstruction + a non-absent success witness.

The loop side conditions (`hslack`/`hover`/`hvalid`/`hsalign`) ARE dischargeable from the account-region bounds — the field windows nest inside the outer window by the decode fit conjuncts, so `endOff_field + 9 ≤ aLen + 9 ≤ |region|`; no slack gap (the HARD obligation from the review is safe).

All new lemmas classical-3; `lake build` green; all `scripts/check-*.sh` clean; byte-transparent, symbolic `GuestAddrs`, additive.

Refs `evm-asm-4ch8f.43.5`; builds on merged #10181/#10183.

### DIFF
--- a/EvmAsm/Codegen/Programs/BalAccountNonstorageFinalsChainB.lean
+++ b/EvmAsm/Codegen/Programs/BalAccountNonstorageFinalsChainB.lean
@@ -213,12 +213,374 @@ def outerRej (aB : Word) (acctBytes : List (BitVec 8)) (F : Assertion) :
   ((.x0 : Reg) ↦ᵣ (0 : Word)) ** regOwn .x1 **
   bytesRegion aB acctBytes ** F
 
-/- The outer `rlp_walk_init` dispatch (slot 22 call + slot 23 status check
-   ⇒ `outerRej` / `outerInitPost`) is the next slice: the nine callee arms
-   need explicit per-arm branch lemmas (seven `outerFail` instantiations +
-   the two success arms unified into `OuterInitOk` via the short/long
-   `listHeaderSize` bridges), recombined with `cpsBranchWithin_pre_or` as in
-   `fl_round`.  The interfaces above are final. -/
+/-- Outer `rlp_walk_init` (slot 22) + status check (slot 23): reject on any
+    non-zero status; on success, land at `B + 96` with the unified
+    `listHeaderSize`-anchored content cursor.  The nine callee arms collapse
+    pointwise into success (`OuterInitOk`) vs failure (some non-zero status)
+    BEFORE the branch dispatch, so only two branch lemmas are needed. -/
+theorem bansf_outerInit_spec (aB : Word) (aLen : Nat)
+    (acctBytes : List (BitVec 8))
+    (v5 v6 v7 v12 v28 v29 v30 v31 vRa : Word) (F : Assertion)
+    (hF : F.pcFree)
+    (hsalign : aB.toNat % 8 = 0)
+    (hslack : aLen + 9 ≤ acctBytes.length)
+    (hover : aB.toNat + acctBytes.length < 2 ^ 64)
+    (hvalid : ∀ k, k < acctBytes.length →
+      isValidByteAccess (aB + BitVec.ofNat 64 k) = true) :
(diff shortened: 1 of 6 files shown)


pirapira:
**Review (Claude Code) — LGTM as groundwork (does not close `.43.5`, correctly labeled).**

- Genuine + non-vacuous: `bansf_prologue_spec` is a real 15-instruction `runBlock` proof (arg saves, 10-dword out-block zeroing, reloads) with concrete pre/post; `outerFail` a real reject-route triple (`B+92→B+736`, `a0=1`); `finalsDerivation_witness` discharges by `decide`.
- The `FinalsDerivation` shape change (new `aLen` param; end-pointer `base + bytes.length → base + aLen`) is a **sharpening, not a weakening** — since the owned region extends past the walk window for header-read slack, decode must fit the tighter window. Sole consumer is its own witness (updated, force unchanged); not yet wired into a top theorem. This is exactly the composition fix needed.
- Additive (`cpsTripleWithin`/`Assertion`/`Prop`; no `Ast`/`Vc`/`StmtSound`/`blockOk`); byte-transparent over existing `bansfProg` (the `#guard`s are pre-existing structural asserts, not `GuestAddrs` value pins); `#print axioms` on both new theorems.
- Thanks for folding in the two #10183 doc-nit corrections (kit transitively witnessed; stations 2/3 already delivered).

Remaining for `bansf_spec_within`: outer-init dispatch + 4 outer item units + 3 field stations + body assembly — and the **9-byte-slack discharge from region bounds** you surfaced here is the crux to nail (discharge it without weakening the post; STOP-report if the region invariant can't supply it). Good increment. Ready for `merge!`.

PR #10189, opened 2026-07-11
fix(selfdestruct): observe live beneficiary balance

pirapira:
## Summary
- consult the transaction-journaled latest beneficiary balance before charging SELFDESTRUCT new-account gas
- recognize beneficiaries made alive by a prior committed transfer even when CALLCODE or DELEGATECALL changes the SELFDESTRUCT origin
- retain frame rollback semantics through the existing journaled nonstorage effect log

## Execution-specs alignment
Amsterdam SELFDESTRUCT uses is_account_alive on live transaction state. A prior committed SELFDESTRUCT transfer can make a pre-state-empty beneficiary alive; later executions must not charge ACCOUNT_WRITE plus NEW_ACCOUNT again.

## Verification
- complete reentrancy_selfdestruct_revert family: 9/9 exact
- randomized fixtures 11000-11999 after fix: 996/1000 exact; remaining four are precompile/aggregate cases
- lake build
- check-file-size.sh
- check-layering.sh
- check-forbidden-tactics.sh

### DIFF
--- a/EvmAsm/Codegen/Programs/Selfdestruct.lean
+++ b/EvmAsm/Codegen/Programs/Selfdestruct.lean
@@ -87,6 +87,23 @@ def selfdestructNewAccountSurchargeAsm : String :=
   "  addi t0, t0, 1; addi t2, t2, -1; bnez t2, .L_selfdestruct_no_ctx_bal_loop\n" ++
   "  j .L_selfdestruct_surcharge_done\n" ++
   ".L_selfdestruct_charge_new_account:\n" ++
+  -- `is_account_alive` reads the live transaction state. A prior committed
+  -- value transfer can therefore make a pre-state-empty beneficiary alive even
+  -- when it came from a different SELFDESTRUCT origin (CALLCODE/DELEGATECALL).
+  -- Consult the frame-journaled nonstorage effect log before the origin-only
+  -- repeated-SELFDESTRUCT shortcut below. A nonzero latest balance means the
+  -- beneficiary is alive, so ACCOUNT_WRITE and NEW_ACCOUNT are not charged.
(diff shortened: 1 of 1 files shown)


pirapira:
**Review (Claude Code) — LGTM, one conservative-direction note.**

**Exec-spec alignment (checked bd8c673 `system.py` SELFDESTRUCT ~:672-682):** the surcharge is charged only when `not is_account_alive(tx_state, beneficiary) and get_account(target).balance != 0`, and `is_account_alive` reads the **live** tx_state. The fix consults `nonstorage_effect_latest_balance` for the beneficiary and skips NEW_ACCOUNT + ACCOUNT_WRITE when the journaled latest balance is nonzero — correctly handling a prior committed transfer that made a pre-state-empty beneficiary alive (including the CALLCODE/DELEGATECALL origin-change case the origin-only shortcut missed). Register spill/restore around the call is correct. No forbidden tactics (`0xef`/`0x01` are byte literals).

**Note (conservative, no false-accept):** the fix proxies "alive" with nonzero **balance** only, whereas `is_account_alive = not is_account_empty` is also true for nonzero nonce / non-empty code at zero balance. A zero-balance-but-alive beneficiary would still be surcharged (over-charge / false-reject) — the safe direction, and balance is the relevant vector for SELFDESTRUCT-transfer liveness. Acceptable; flagging for completeness.

Validation (reentrancy 9/9 exact; randomized 996/1000, remainder unrelated) sufficient. Ready for `merge!`.

pirapira:
Filed follow-up bead **evm-asm-4dea3** (P1) for the balance-only liveness residual noted above: extend the beneficiary liveness observation to full `is_account_alive` (live nonce + code, not just balance) so a zero-balance-but-alive beneficiary isn't over-surcharged. Conservative-direction gap only (no false-accept); tracked for exact-model completion. — Claude Code

PR #10184, opened 2026-07-11
feat(specref): rlpToMutableNode decoder + mptNodeIs roundtrip (4ch8f.75.3)

pirapira:
Bead: **evm-asm-4ch8f.75.3** (blocks `.29` MPT walk/lookup and `.75.4`). Doc: `docs/4ch8f-slstate-specref-correspondence.md` §4 — closes the one load-bearing missing abstraction on the MPT side.

New `EvmAsm/Evm64/MptCorrespondence.lean` (imports `MptAssertions`; builds on the SpecRef MPT model — `WitnessState.lean` `MutableNode`/`trieLookup`, s1d19.1's `IncrementalMpt.lean` for the resolve closure it deliberately does NOT duplicate):

- `rlpToMutableNode : List (BitVec 8) → Option MutableNode` — shallow decode of one RLP trie node. Skeleton = `mptNodeKindSpec` (the proven `mpt_node_kind` discriminator mirror) returning the node instead of the tag: 17-item list ⇒ branch (children via `rlpChildSlot`: empty ⇒ `none`, 32-byte ⇒ `.hashed`; inlined `<32`-byte children out of v1 scope = bead `4ch8f.75.4`), 2-item `[compact, payload]` ⇒ leaf/extension by `hpDecode`.
- `alpha_node : MptNode → MutableNode` — the doc §4 abstraction (paths verbatim, hash refs to `.hashed`, empty slots to `none`).
- **`rlpToMutableNode_rlp`**: for every well-formed `n : MptNode`, `rlpToMutableNode n.rlp = some (alpha_node n)`. Proof cases are exactly `mptNodeKindSpec_rlp`'s: `decodeFully_encode` (via `MptNode.rlp_length_lt`), `hpDecode_hpEncode` for leaf/extension, new `mapM_rlpChildSlot` for the branch slots. Since `mptNodeIs ptr n` asserts `bytesRegion ptr n.rlp` + `n.WF`, this is the `mptNodeIs → MutableNode` connective tissue `.29`'s walk needs.
- Non-vacuous: `#guard` cross-checks on the same concrete vectors `MptAssertions` uses (leaf/extension/branch, decode failures), an end-to-end `SpecRef.trieLookup` read over a decoded 1-node trie (found + not-found), and a concrete `WF` witness `example`.
- `rlpToMutableNode_rlp` registered as a `Progress.lean` witness (fenced by `check-axioms.sh`).

[...]

PR #10191, opened 2026-07-11
fix(identity): allow large root memory windows

pirapira:
## Summary
- raise the depth-0 EVM memory arena to 1 MiB so valid high-memory CALL expansion is charged before child execution
- preserve execution-specs behavior for the 1,000,000-byte identity input window with insufficient forwarded gas
- regenerate linked symbol positions and update the guest-image RAM extent proof

## Verification
- authoritative ziskemu: both large identity fixtures pass exactly (2/2)
- randomized Spike: complete identity-precompile family passes exactly (32/32)
- lake build
- scripts/check-region-map.sh
- scripts/check-file-size.sh
- scripts/check-layering.sh


pirapira:
**Review (Claude Code) — LGTM.** Monotone-safe arena bound raise.

Raises depth-0 `rootRuntimeMemoryArenaLimitBytes` 0x50000→0x100000 (`EvmMemoryGas.lean:32`) so a ~1MB top-level CALL input window to IDENTITY expands (charging memory gas) before the child OOGs on its separately-forwarded gas — matching `process_message`/`generic_call` (parent charges expansion; child fails independently). `identity.py` semantics unchanged.

- Returndata cap not truncating: `precompileFrameReturndataCapBytes := rootRuntimeMemoryArenaLimitBytes` rises in lockstep, and RETURNDATACOPY's `start+size ≤ retlen` guard bounds reads to the real staged length — raising an upper-bound staging cap can't cause over-read. Nested-frame arena stays 0x20000 (only depth-0 raised), so no per-slot bloat.
- Region arithmetic consistent: data-region extent grew exactly 0xB0000 (= 0x100000−0x50000), RAM-extent proof updated to match; `#guard 1000000 ≤ rootRuntimeMemoryArenaLimitBytes` enforces the target. No forbidden tactics / hardcoded PC.

Scope note (informational): the diff also carries #10184's `MptCorrespondence.lean` via the base choice (`fix/selfdestruct-live-beneficiary-gas` head that merged #10184) — stacking, already reviewed under #10184, clean. Validation: ziskemu 2/2 large-identity + Spike 32/32 identity. Ready for `merge!`.

PR #10192, opened 2026-07-11
fix(precompile): refund failed-call account charge

pirapira:
## Summary
- record provisional Amsterdam NEW_ACCOUNT state-gas charges made before precompile dispatch
- refund the recorded charge exactly once when the precompile child fails
- restore spilled state gas before the reservoir, matching the original charge order
- regenerate linked symbol addresses and text-region size after simplifying the failure path

## Verification
- lake build
- scripts/check-region-map.sh
- scripts/check-file-size.sh
- scripts/check-layering.sh
- randomized Spike EEST slice 11000-11999: 999/1000 exact; only the previously known KZG one-gas mismatch remains
- ziskemu EEST global fixture 11775 all_opcodes: 1/1 exact


pirapira:

Replaces the `value!=0`-only NEW_ACCOUNT refund-on-failed-precompile heuristic with a flag-driven refund, and sets `cd_new_account_charged_current` on the charge path (`ChildFrameHandlerTails.lean:207`).

- Spec match: refund-on-child-error gated by the recorded flag mirrors `new_account_charged` in `generic_call` (`system.py:378-381`: `if child_evm.error … if params.new_account_charged: credit_state_gas_refund(NEW_ACCOUNT)`). Refunds exactly once iff the charge happened — fixing the old heuristic that ignored `is_account_alive` (spec charges iff `has_value and not is_account_alive`, `system.py:468`).
- Refund math correct: restores spill first then reservoir per `credit_state_gas_refund` (`__init__.py:194-217`); amount 183600 = 120 (`STATE_BYTES_PER_NEW_ACCOUNT`) × 1530; `bltu … 183600` guards `state_gas_used` underflow.
- Flag lifecycle sound: pre-existing `.zero 8`, cleared at each CALL-entry disarm, saved/restored across nesting (`frame+752`), set on both charge paths — a stale flag from a *successful* new-account precompile is cleared before the next call's failure path, so no spurious refund.

Minor: the retained `beqz t6` (value!=0) recheck after the flag gate is redundant (flag ⇒ value≠0) but harmless; net −79/+7 is a clean simplification. Validation: Spike 11000-11999 999/1000 (residual = the KZG issue #10195 fixes). Ready for `merge!`.

PR #10195, opened 2026-07-11
fix(kzg): validate direct precompile transactions

pirapira:
## Summary
- validate direct transaction calls to the KZG point-evaluation precompile instead of assuming every 192-byte input succeeds
- check the EIP-4844 versioned hash using SHA-256 of the commitment
- run the real KZG proof verifier and treat invalid hash/proof as an exceptional halt that consumes all execution gas
- regenerate linked symbol addresses and the measured text-region size

## Root cause
The top-level active-precompile shortcut assigned the fixed 50000 gas cost without executing KZG validation. An invalid proof with 50001 available gas was therefore treated as successful and refunded one gas, unlike execution-specs.

## Verification
- lake build
- scripts/check-region-map.sh
- scripts/check-file-size.sh
- scripts/check-layering.sh
- ziskemu point_evaluation_precompile/tx_entry_point family: 6/6 exact (random seed 101923, two jobs)
- randomized Spike EEST slice 11000-11999: 1000/1000 exact (seed 1014186486, ten jobs)


pirapira:
**Review (Claude Code) — LGTM. This closes a genuine false-accept.**

The top-level (direct-transaction) point-evaluation (0x0a) shortcut previously assigned the fixed 50000 gas to *any* 192-byte input **without validating the proof** — so an invalid KZG proof with e.g. 50001 gas falsely "succeeded" and refunded 1 gas. This adds real validation.

Spec-exact vs `point_evaluation.py` / `kzg.py`:
- `len==192` check; `sha256(commitment = data[96:144])`; version byte `data[0]==0x01`; compares `versioned_hash[1:32] == sha256[1:32]` — exactly `kzg_commitment_to_versioned_hash = 0x01 ++ sha256(commitment)[1:]` (`kzg.py:63-71`), correctly not comparing sha byte 0.
- `zkvm_kzg_point_eval(commitment, z=data[32:64], y=data[64:96], proof=data[144:192])` — arg order matches `verify_kzg_proof(commitment, z, y, proof)`; `bnez a0` (exception→fail) = `except: raise KZGProofError`; result-byte-0 zero → fail = `if not verification: raise`.
- Gas/failure semantics: invalid hash/proof → `precompile_fail` (consumes all gas) = exceptional KZGProofError halt; only a valid proof sets 50000. Buffer layout clean (digest at `+0`, kzg output at `+32`, cleared first — no collision). Empty/oversized → len≠192 → fail.

Validation: ziskemu tx_entry_point 6/6 exact; Spike 11000-11999 now 1000/1000 (closes the known 1-gas KZG mismatch). Strong fix. Ready for `merge!`.

PR #10198, opened 2026-07-11
fix(storage): cover protocol access bounds

pirapira:
## Summary
- raise the runtime unique storage warmth table from the fixture-sized 512 cap to the existing 16384-row protocol-safe storage-log bound
- size per-op storage-access outcomes from Amsterdam's 16777216 regular-gas cap and the 100-gas warm access floor
- update linked data/text region measurements and the guest-image scratch proof

## Root cause
A valid CLZ stress program touches 529 distinct storage slots. The 513th cold SSTORE hit the guest's 512-key implementation cap and was incorrectly surfaced as an exceptional halt. The new bounds derive from protocol gas/storage limits rather than this fixture.

## Verification
- lake build
- scripts/check-region-map.sh
- scripts/check-file-size.sh
- scripts/check-layering.sh
- focused CLZ 529-slot fixture: exact match
- randomized Spike slice 15000-15999: storage failure fixed; 998/1000 exact, with only two pre-existing CALLCODE dynamic-code gas mismatches remaining (seed 112139021, ten jobs)


pirapira:
**Review (Claude Code) — LGTM.** Monotone-safe protocol-bound cap raise.

- `storageAccessGasMaxKeys` 512→16384, `storageAccessOutcomeMaxRecords` 512→167773. Raising a fail-closed cap only removes false exceptional-halts; it can't create a false-accept while the cap still exceeds the true protocol max.
- Bounds check out: max distinct storage keys ≤ TX_MAX_GAS / ACCESS_LIST_STORAGE_KEY_COST = 16,777,216 / 1900 ≈ 8830 < 16384; outcome records = ⌈16,777,216 / 100 (warm floor)⌉ = 167773. Correct and conservative.
- `dataSizeBytes` grows ~26.5 MiB; the GuestImage scratch proof + RegionMap are updated to the `readelf`-measured values and gated by `check-region-map.sh` (ground truth). No forbidden tactics.
- Validation: Spike 15000-15999 998/1000, the 2 residuals being the CALLCODE dynamic-code gas issue #10202 fixes. Sufficient. Ready for `merge!`.

PR #10202, opened 2026-07-11
fix(create): preserve nested initcode memory

pirapira:
## Summary
- execute CREATE initcode from the suspended parent frame memory instead of the shared global staging buffer
- preserve outer initcode bytes across nested CREATE/CREATE2 staging
- regenerate linked symbol positions and update the text-region extent

## Root cause
Nested CREATE staged its initcode into the single global create_child_initcode buffer while the outer CREATE frame was executing from that same buffer. Returning from the inner CREATE resumed the outer PC against overwritten bytes, skipping two SSTOREs and undercounting Amsterdam state gas by 195,840.

## Verification
- focused CALLCODE dynamic-code d2/d3: 2/2 exact (Spike, randomized seed 10203, 2 jobs)
- EEST rows 15000-15999: 1000/1000 exact (Spike, randomized seed 112139021, 10 jobs)
- lake build
- scripts/check-region-map.sh
- scripts/check-file-size.sh
- scripts/check-layering.sh
- scripts/check-unimported.sh
- scripts/check-drift.sh
- scripts/check-no-warnings.sh


pirapira:

`CreateFrameDescend` now reads `code_ptr` from `parent_mem_base(x13) + create_init_offset` instead of the shared global `create_child_initcode` buffer. x13 is still the parent frame's memory base here (rebased to the child only afterward), so the initcode read targets the correct suspended-parent arena. Matches `amsterdam/system.py` `generic_create` (~:92-94: `memory_read_bytes(evm.memory, start, size)` snapshotted into `child_message.code`): because each depth has a distinct memory arena, the suspended parent's window is immutable while the child (and any nested CREATE, based off the child's arena) runs — equivalent to the spec's copy. Memory zero-extension correctness is unchanged (pre-existing).

Validation: CALLCODE d2/d3 2/2, EEST 15000-15999 1000/1000 exact — closes the two #10198 residuals. Ready for `merge!`.

PR #10203, opened 2026-07-11
refactor(divmod): extract phase2b quotient definition

pirapira:
## Summary
- move the pure `div128Quot_phase2b_q0\x27` definition out of the instruction-level `Div128ProdCheck2` proof
- let `LoopDefs.Iter` import the lightweight definition leaf directly
- make tactic and address-normalization dependencies explicit in consumers that previously received them transitively

## DAG impact
- removes `LoopDefs.Iter -> LimbSpec.Div128ProdCheck2 -> Rv64.Tactics.RunBlock` from the pure loop-definition chain
- localizes instruction-proof rebuilds away from all V4/V5 quotient models and their V6 consumers
- the global maximum remains 94 because the independent CLZ arithmetic chain is tied for the current maximum

## Validation
- `lake build`
- architecture gates (`check-layering`, forbidden tactics, unimported files, file size)
- `git diff --check`

Stacked on #10199.

### DIFF
--- a/EvmAsm/Evm64/DivMod/Compose/CLZ.lean
+++ b/EvmAsm/Evm64/DivMod/Compose/CLZ.lean
@@ -8,6 +8,7 @@
 import EvmAsm.Evm64.DivMod.Compose.Base
 import EvmAsm.Evm64.DivMod.Compose.V4NoNop
 import EvmAsm.Evm64.DivMod.LimbSpec.CLZ
+import EvmAsm.Evm64.DivMod.LoopDefs.CLZResult
 
 open EvmAsm.Rv64.Tactics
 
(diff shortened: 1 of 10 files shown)


pirapira:
**Review (Claude Code) — LGTM.** Behavior-preserving refactor (rolls up the #10205 CLZ slice, so merging this lands both).

- `div128Quot_phase2b_q0'` moved to a new leaf `LoopDefs/Div128Phase2b.lean` with a **byte-identical body** and the same name; `div128Quot`'s reference resolves unchanged. The CLZ slice similarly extracts `clzResult` to `LoopDefs/CLZResult.lean` byte-identically.
- DAG narrowing only (removes `Iter → Div128ProdCheck2 → RunBlock` and `CLZLemmas → Compose.CLZ`; leaf defs depend only on `Rv64.Instructions`). The added imports in `AddrNorm`/`FinCases`/`RunBlock`/`XSimp`-consumers just make previously-transitive deps explicit — additive. No `def` shape changed, no frozen statement weakened.
- No `sorry`/`native_decide`/`bv_decide`/`maxHeartbeats`; no new load-bearing theorems (pure def-move), classical-3 witnessing inherited.

Minor (non-blocking): the old CLZ body left behind as `private clzResult_legacy` is now dead code branch-wide — worth deleting to avoid drift. Ready for `merge!`.

PR #10204, opened 2026-07-11
fix(create): initialize recursive child nonce

pirapira:
## Summary
- seed successfully entered CREATE children with nonce 1 before initcode execution
- preserve the child dispatcher PC across the nonce-table helper
- regenerate linked symbol positions and update the text-region extent

## Root cause
Execution-specs initializes a created account nonce to 1 before process_create_message runs its initcode. The guest re-read a pre-funded recursive target pre-state nonce of 0, derived the wrong next CREATE address, missed the pre-resolved balance table, and failed to refund four target-alive NEW_ACCOUNT charges.

## Verification
- focused call_recursive_contract row 17050: 1/1 exact (Spike, randomized seed 10207)
- EEST rows 17000-17999: 1000/1000 exact (Spike, randomized seed 10205, 10 jobs)
- lake/codegen build completed successfully
- scripts/check-region-map.sh
- scripts/check-file-size.sh
- scripts/check-layering.sh
- scripts/check-unimported.sh


pirapira:
**Review (Claude Code) — LGTM, spec-exact (EIP-161).**

New `create_creator_nonce_seed_one` seeds the entered child's creator-nonce to **1** before initcode runs. Matches `amsterdam/interpreter.py` `process_create_message` (~:230 `increment_nonce(new_account)` before `process_message`) combined with `account_deployable` requiring `nonce == 0` (`state_tracker.py` ~:361) — a deployable target always goes 0→1, so a literal 1 is precisely correct (including the pre-funded nonce-0 case).

Path-gating is correct: the seed fires only inside `create_frame_descend` (successful descent), not on the collision / not-deployable / balance-or-depth-fail branches (`system.py` ~:109-130), which correctly must not increment. Overflow is fail-closed (`create_nonce_table_overflow` at `createNonceTableCap`); x10/dispatcher-PC preserved around the a0-clobbering helper. No forbidden tactics.

Validation: recursive-contract 17050 1/1, EEST 17000-17999 1000/1000. Ready for `merge!`.

PR #10206, opened 2026-07-11
feat(sasm): verify hp_decode_nibbles byte-transparently — upLoop_spec + abiFrame contract (4ch8f.16.3)

pirapira:
Bead: **evm-asm-4ch8f.16.3** (unblocks the `.16.3.1` shape blocker and one of `.29` mpt_walk's three missing callees). **Stacked on #10184** (shares a PLAN.md hunk — merge that first).

`.16.3.1` recorded two capability gaps: a branch-to-tail/shared-epilogue combinator (landed as `RetFromLoop`, #10115) and saved-register frame support (landed as `AbiFrame`, #9982). This PR adds the one still-missing piece and completes the port **byte-transparently — no re-emit**:

- **`EvmAsm/Rv64/SAsm/UpLoop.lean`**: `upLoop_spec` — the `bgeu idx, lim` top-guard **up-counting** loop lemma at `cpsTripleWithin` level, register-agnostic (s-register-capable), sibling of `countdownLoop_spec`. Reusable for every emitted cursor-vs-limit loop.
- **`HpDecodeNibblesSAsm.lean`**: byte transparency `abiFrameProg (-48) 48 hdnFrame hdnBody = hpDecodeNibbles_prog` (`rfl` + `#guard`); the guest-exact decode model `hdnRes` with its SpecRef ties — `hdnRes_eq_hpDecode` (agreement wherever the guest succeeds) and `hdnRes_hpEncode` (round-trip on every hex-prefix encoding; the guest is stricter than `EvmAsm.Evm64.hpDecode` exactly on even paths with a non-zero padding nibble, which it rejects — documented divergence); segment/branch/loop triples at a **symbolic base** (no `GuestAddrs` value pins); `bytesRegion_sb1_within` (offset-1 byte store); a `bytesRegion` `PCFree` instance (unblocks `runBlock` framing over byte regions generally).
- **`HpDecodeNibblesSAsmPaths.lean`** (file-size-gate split): the five control paths (empty input, invalid flag, even-strict reject, odd success, even success), the unified single-exit body (ParentHeaderFrame disjunctive-post technique, scratch registers weakened to `regOwn`), and **`hp_decode_nibbles_spec`** — the whole-routine `abiFrame_spec` contract: `sp`/`ra`/all callee-saved registers restored, `a0` = parse status, nibble buffer = `hdnBufFinal` (decoded nibbles spliced over the caller buffer), count/is-leaf cells written exactly as the routine does (including the fail-path subtlety that the is-leaf cell IS written on the even-strict reject but NOT on the earlier rejects).

Genuine, unweakened post: the full decode semantics vs `hdnRes`/`hpDecode`, not a shape claim. Anti-vacuity: `hdnRes` `#guard`s incl. both round-trip vectors and all three reject classes.

Gates: `lake build` green (4961 jobs); in-file `#print axioms` = classical-3 on `hp_decode_nibbles_spec` + `hdnRes_hpEncode`; check-{forbidden-tactics,axioms,layering,unimported,file-size,no-warnings,naming,statement-tamper,drift,asm-to-program} all PASS. Additive: `cpsTripleWithin` level only, no `Ast`/`Vc`/`StmtSound*`/`blockOk` changes; guest bytes untouched (spec-only, no EEST).

### DIFF
--- a/EvmAsm/Codegen/Programs/HpDecodeNibblesSAsm.lean
+++ b/EvmAsm/Codegen/Programs/HpDecodeNibblesSAsm.lean
@@ -1,0 +1,1127 @@
+/-
+  Verified byte-transparent port of `hp_decode_nibbles` (bead
+  evm-asm-4ch8f.16.3, unblocking .16.3.1's two capability gaps with the
+  landed `AbiFrame` (PR #9982) and `UpLoop` (this PR) machinery).
+
+  The routine (`EvmAsm/Codegen/Programs/Mpt.lean`, PR-K23) decodes the
+  hex-prefix compact path of a leaf/extension MPT node into one-nibble
+  bytes, writing the nibble count and the is-leaf flag to caller cells:
+
+    a0 src ptr, a1 len, a2 nibble buf, a3 count cell, a4 is_leaf cell
+    → a0 = 0 success / 1 parse failure.
+
+  Shape: a C-ABI saved-register frame (`abiFrameProg`, byte-transparent —
+  no re-emit), a validation prefix whose failure branches jump straight to
+  a shared `li a0,1` fail tail, an odd/even parity split, and an
+  up-counting nibble loop (`bgeu i, len` top guard — `upLoop_spec`).
+
+  The genuine post is stated against `hdnRes`, the guest-exact decode
+  mirror, tied to the spec-side `EvmAsm.Evm64.hpDecode`
+  (`MptAssertions.lean`) by `hdnRes_eq_hpDecode` / `hdnRes_hpEncode`:
+  the guest is STRICTER than `hpDecode` exactly on even paths whose head
+  byte has a non-zero low nibble (it rejects; `hpDecode` ignores the
+  padding nibble) — on every hex-prefix ENCODING the two agree, so the
+  `.29` walk can consume either.
+-/
+
+import EvmAsm.Codegen.Programs.BytesToNibblesSAsm
+import EvmAsm.Evm64.MptAssertions
+import EvmAsm.Rv64.SAsm.AbiFrame
+import EvmAsm.Rv64.SAsm.UpLoop
+import EvmAsm.Rv64.MemRegionStore
+
+namespace EvmAsm.Codegen
+
+open EvmAsm.Rv64 EvmAsm.Rv64.SAsm
+
+namespace HpDecodeNibblesSAsm
+
+open BytesToNibblesSAsm (highNibble lowNibble nibblePair nibblePrefix
+  length_nibblePrefix)
+
+/-! ## The byte-transparent frame split -/
+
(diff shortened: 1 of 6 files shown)


pirapira:
**Review (Claude Code) — verification LGTM; the "documented divergence" is a real guest↔execution-specs divergence I'd track separately.**

**Verification quality (clean):**
- Byte-transparent: `abiFrameProg (-48) 48 hdnFrame hdnBody = hpDecodeNibbles_prog` by `rfl` + `#guard`; symbolic base, **no `GuestAddrs` value pins**.
- Genuine unweakened post: `hp_decode_nibbles_spec` is the whole-routine `abiFrame_spec` contract — `sp`/`ra`/callee-saved restored, `a0` = parse status, buffer = `hdnBufFinal`, count/is-leaf cells written exactly as the routine does (incl. the fail-path subtlety that is-leaf IS written on the even-strict reject but not earlier rejects). Not a shape claim.
- Non-vacuous: `hdnRes` `#guard`s incl. both round-trip vectors + all three reject classes.
- `#print axioms` classical-3 in-file on `hp_decode_nibbles_spec` + `hdnRes_hpEncode`; no `sorry`/`native_decide`/`bv_decide`/`maxHeartbeats`; additive (`cpsTripleWithin` level). The new `upLoop_spec` (up-counting top-guard loop, register-agnostic) is a nice reusable sibling to `countdownLoop_spec`. Guest bytes untouched.

**On the even-path padding-nibble divergence** (`hdnRes` stricter than `hpDecode`): I checked execution-specs at bd8c673. `compact_to_nibbles` (`amsterdam/incremental_mpt.py:878-889`) reads `compact[0] & 0x0F` **only in the `odd` branch** — for even paths the padding nibble is ignored entirely, with no validation and no check in the caller (`:947`). So execution-specs is **lenient** and *accepts* a non-zero even-path pad, whereas the guest *rejects* it. That makes the guest strictly stricter than the reference.

- Direction: **completeness only** (guest INVALID where spec VALID = false-reject) — never a false-accept, so no soundness impact.
- Almost certainly **unreachable on valid blocks**: canonical HP (`nibble_list_to_compact`) always pads even paths with 0, so a non-zero-pad node can't hash-match an authenticated witness under collision-resistance (it'd be rejected by the hash check anyway).

Verifying the guest *as it is* (including the strictness) is correct, so this doesn't block the PR. But per our "file P1 for any guest↔spec divergence" discipline, a mere in-code "documented divergence" note isn't the end state — I filed **evm-asm-3umhl** to decide between (a) relaxing the guest to match execution-specs' lenient decode, or (b) keeping the stricter check with a *rigorous* unreachability argument (canonical pad=0 witness) rather than a comment. Tracking there, not here.

Verification is ready for `merge!`.

PR #10209, opened 2026-07-11
feat(eest): add SpecRef differential oracle mode

pirapira:
## Summary

- add an opt-in `--specref-oracle` / `EEST_SPECREF_ORACLE=1` mode to the RV64 EEST harness
- run the validated SpecRef on the exact guest-visible input and compare serialized output byte-for-byte
- classify verdict disagreements as guest false-accept or guest false-reject, report other byte divergences, and fail the run on any oracle divergence
- build `specref-eest-check` automatically when oracle mode is enabled

This is the differential guest↔SpecRef lane unlocked by the completed `evm-asm-s1d19` execution-seam work. It keeps fixture comparison intact while making SpecRef an independent oracle.

## Validation

- `lake build` (4,958 jobs)
- `scripts/check-spec-refs.sh`
- `git diff --check`

A full ziskemu differential fixture run is not claimed in this PR; the checkout did not have a prebuilt `gen-out/stateless_guest.elf`.

Refs: `evm-asm-s1d19`, execution-specs @ `bd8c673`.

### DIFF
--- a/scripts/codegen-eest-stateless-check.sh
+++ b/scripts/codegen-eest-stateless-check.sh
@@ -95,6 +95,8 @@
 #     --verify-execution-spec-input
 #                        decode the same guest-visible bytes through
 #                        execution-specs run_stateless_guest's input path
+#     --specref-oracle   also run SpecRef on each input and fail on any
+#                        byte-for-byte guest↔SpecRef divergence
 #     --random           shuffle fixtures into a random order before applying
 #                        --limit; use to sample a different subset on each run
 #                        and discover failures outside the default first-N fixtures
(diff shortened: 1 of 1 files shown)


pirapira:
**Review (Claude Code) — LGTM.** Clean, correct differential-oracle harness — and it directly implements the guest↔SpecRef lane the completed s1d19 work unlocked.

Correctness of the differential:
- **Same input to both**: `run_specref_oracle` runs `lake exe specref-eest-check "$input"` on the exact guest-visible `$input` the emulator consumed — a genuine differential, not a re-derivation.
- **Verdict offset correct**: `${actual_hex:64:2}` / `${oracle_hex:64:2}` extract hex chars 64–65 = **byte 32 = `successful_validation`** (root = bytes 0:32 = hex 0:64; succ = byte 32; tail after). Right region.
- **Classification correct**: guest `01` / specref `00` → false-accept; `00`/`01` → false-reject; any other byte mismatch → "output". Fails the run on any `oracleDiff > 0` (`rc=1`). Opt-in (`--specref-oracle`/`EEST_SPECREF_ORACLE=1`), fixture comparison left intact, and it auto-builds `specref-eest-check`. Param validated (0/1). No forbidden tactics (shell); `bash -n` clean.

Two minor robustness notes (non-blocking):
1. The oracle block is guarded only by `SPECREF_ORACLE == 1`, not by `status == OK`. `BUDGET` returns early, but if an emulation `ERROR`/`FAIL` status reaches it, `oracle_hex` is empty → it's counted as an `oracleDiff` (class "output") with a garbage verdict extraction, and fails the run misleadingly. Worth gating the block on `status == OK` (and non-empty `oracle_hex`) so only guest-OK cases are compared.
2. Length asymmetry: `actual_hex` is `xxd -p -l "$expected_bytes"` (truncated to expected) while `oracle_hex` is the *full* SpecRef output (no `-l`). For the fixed 105-byte result this is fine, but variable-length outputs (e.g. the malformed failed-sentinel class) could flag a spurious diff even when the meaningful prefix agrees. Consider comparing on a common length.

Per your note, the absent full ziskemu differential run is fine ahead of the v0.6.0 migration — this will be most valuable once re-baselined there anyway. Ready for `merge!`.

PR #10220, opened 2026-07-11
feat(sasm): rlpItemDecode_spanStart — field-window bounds (4ch8f.43.5)

pirapira:
## Summary

Slice 3e toward `bansf_spec_within` (bead `evm-asm-4ch8f.43.5`), on top of merged #10188 (SEG-B: prologue, nine-arm outer init dispatch, four item units, assembled chains).

**`rlpItemDecode_spanStart`**: the span-start of an accepted decode (`next − len`, the routine's `s3 := a0 − a2` capture) lies inside the walk window — at or past the item's own cursor, and with the span at or before the window end. This is the bounds fact that lets the three field stations discharge their `rlp_walk_init` side conditions (and the inner-walk 9-byte slack) **from the outer region bounds**: the balance/nonce/code windows nest inside the outer window, so `fieldEnd + 9 ≤ aLen + 9 ≤ |region|`.

Proof mirrors `rlpItemDecode_advance`: the representation component is hoisted (free `ofNat_toNat` round-trip); the five prefix arms close by `bv_omega` with the `fromBytesBE` payloads passed **opaquely** through per-form sub-lemmas (`spanStart_shortlist`/`longstring`/`longlist`) — inline `generalize` on the payload term whnf-times-out, and `signExtend12` must be rewritten away first (opaque to `bv_omega`). The long-list arm aligns the span shape via `add_rot`.

Classical-3; `lake build` green; byte-transparent/additive as before.

Next (same lane): the balance station (field-init dispatch reusing the #10188 nine-arm collapse, empty-skip → `FieldFinal.empty`, loop entry into `bansf_findLastLoop1_spec`, tuple parse, `rlp_content_to_u256_be` capture), then nonce/code stations (generated), then the `abiFrame_spec_own` assembly ⇒ `bansf_spec_within`.

Refs `evm-asm-4ch8f.43.5`; builds on merged #10181/#10183/#10188.


pirapira:
**Review (Claude Code) — LGTM.** Clean, genuine bounds lemma; exactly the machinery the `.43.5` field stations need.

`rlpItemDecode_spanStart` states a real, non-vacuous fact: for an accepted decode, the span-start `next − len` (the routine's `s3 := a0 − a2` capture) satisfies three conjuncts — (1) the no-wrap representation round-trip `next−len = base + ofNat (…).toNat`, (2) `off ≤ (next−len−base).toNat` (span-start at/past the item cursor), and (3) `(next−len−base).toNat + len.toNat ≤ endOff` (span end at/before the window end). So the item's span `[spanStart, spanStart+len)` provably nests inside `[off, endOff)`. The `hover : base+endOff+9 < 2^64` slack precondition matches `rlpItemDecode_advance` and is discharged routine-side from region bounds.

This is precisely the fact that lets the balance/nonce/code field windows discharge their `rlp_walk_init` side conditions **and the inner-walk 9-byte slack from the outer region bounds** (`fieldEnd + 9 ≤ aLen + 9 ≤ |region|`) — i.e. it directly attacks the slack-discharge crux for `bansf_spec_within`.

Proof mirrors `rlpItemDecode_advance`: five prefix arms via `bv_omega` with `fromBytesBE` payloads passed opaquely through the per-form sub-lemmas (`spanStart_shortlist`/`longstring`/`longlist`) to avoid the whnf timeout, `signExtend12` rewritten away first, long-list shape via `add_rot`. Sound. No `sorry`/`native_decide`/`bv_decide`/`maxHeartbeats`; `#print axioms` classical-3; byte-transparent (pure proof over walk semantics), additive (one file), on top of #10188. Ready for `merge!`.

PR #10218, opened 2026-07-11
feat(specref): v0.6.0 VM gas plumbing — set_delegation rewrite + prepare_dispatch + frame accounting (evm-asm-0w05f.6)

pirapira:
Phase 4c of the **tests-zkevm@v0.6.0 migration** — bead `evm-asm-0w05f.6`, references #10207. Based on #10217 (tx layer) so the diff shows only the increment.

## What (C6+C7+C8 in `docs/agents/v06-migration-scope.md`, plus C9 accounting and C12)
- **`set_delegation` rewrite (C6)**: refund machinery deleted; exact top-frame charges — `NEW_ACCOUNT` iff the authority leaf is absent, `ACCOUNT_WRITE` once per first-write authority (`written_accounts` seeded with origin + value-receiving recipient), `AUTH_BASE` once per net-new delegation indicator, never credited back.
- **New `prepare_dispatch` (C7)**: depth-0 state-dependent charges (create: `NEW_ACCOUNT` iff pre-state `EMPTY_ACCOUNT`; call: `NEW_ACCOUNT` iff value>0 to a dead recipient) and EIP-7702 delegation resolution with **warm/cold** access charge (v0.5.0 always charged cold); `prepare_message` no longer loads recipient code.
- **Depth-0 prep rollback (C8)**: `process_message` snapshots before `set_delegation`+`prepare_dispatch`; an `ExceptionalHalt` rolls the whole prep back (authorizations included), consumes all gas, and skips dispatch. `Evm.stateGasUsed` deleted → `authStateGasUsed` + `frame_state_gas_used` from the reservoir delta; `refill_frame_state_gas` restores the message baseline; `MessageCallOutput` loses `state_refund`/`created_target_alive`.
- **Block accounting (C9, pulled in — coupled to the deleted output fields)**: create-tx refund block gone; `tx_regular_gas = max(before_refund − state, calldata_floor)`.
- **Builder predeploys (C12, pulled forward from 4d)**: the old EIP-8282 addresses hold no code in any v0.6.0 witness, so **every** fixture failed the system-tx pre-check ("System contract address does not contain code") — updated to `0x0000BFF4…0D8282` / `0x000064D6…0E8282`.

## Conformance vs tests-zkevm@v0.6.0
- Smoke subset: **40/40 FULL match** (root+succ+tail byte-exact).
- Random 300-fixture sample (seed 7): **300/300 FULL match, succ FAIL 0**.

[...]

PR #10216, opened 2026-07-11
feat(specref): v0.6.0 stateless wire format — schema id 0x1501 + ForkConfig shrink (evm-asm-0w05f.4)

pirapira:
Phase 4a of the **tests-zkevm@v0.6.0 migration** — bead `evm-asm-0w05f.4`, references #10207. Based on #10215 (fixture tag) so the diff shows only the increment.

## What (C1+C2 in `docs/agents/v06-migration-scope.md`)
- **Schema id `0x0001` → `0x1501`**: `fork_index << 8 | revision` with new `ProtocolFork.value` IntEnum mapping (`Frontier = 0x01` … `Amsterdam = 0x15`); `#guard STATELESS_INPUT_SCHEMA_ID == 0x1501` pins the big-endian `15 01` prefix.
- **`ForkConfig` = `{activation}`**: `fork` and `blob_schedule` fields deleted along with `BlobSchedule`, its SSZ types/converters, `protocolForkToSsz`/`sszToProtocolFork`, and the now-dead `protocolForks` list and Types.lean blob constants.
- **`validate_chain_config`**: Amsterdam-fork and blob-schedule checks deleted (`SpecError.unsupportedForkConfig`/`.unknownProtocolFork` constructors removed); only activation checking remains, matching `stateless.py:303`.
- Sentinel output, sanity fixtures, and citation line numbers refreshed to `40f956fab`.
- Harness: `eest-specref-check.sh` region decomposition updated for the 69-byte (was 105) `SszStatelessValidationResult`; stale schema prose in `SpecRefEestCheck.lean` / `eest-stateless-to-input.py`.

## Verification
- `lake build` green; all SSZ round-trip `#guard`s (input, chain config, validation result, end-to-end guest pipeline) pass at the new format.
- `check-spec-refs` / `check-forbidden-tactics` clean.
- Smoke conformance vs real `tests-zkevm@v0.6.0` fixtures: **root 40/40, tail 40/40, ERROR/FAIL 0** — the port decodes v0.6.0 inputs and byte-exactly reproduces the new output layout including the chain-config echo.
- `succ` diverges on 38/40 (all EIP-2780 gas fixtures) — expected and scoped: execution-semantics deltas land in phases 4b–4d; the succ gate returns to FAIL 0 at the 4d exit.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #10207, opened 2026-07-11
Upgrade to tests-zkevm@v0.6.0: bump execution-specs submodule, fixtures, RefSpec, and guest


pirapira:
Upstream published **tests-zkevm@v0.6.0** on 2026-07-10 (`ethereum/execution-specs` commit `40f956fab76463e113cb7f0dda4a8a263a1ee776`). We currently target **v0.5.0** (`bd8c673`); SpecRef just reached full-suite conformance at v0.5.0 (25,477 fixtures, succ FAIL 0), so re-establishing conformance at v0.6.0 is the natural next milestone.

Good mission for a **Fable agent** when one frees up.

## Scope of the upstream diff (amsterdam fork, `bd8c673` → `40f956fab`)
~17 files, +479/−457. Concentrated in (review each for semantic changes):
- `vm/interpreter.py` (+165 region — largest), `vm/eoa_delegation.py` (rewritten), `vm/gas.py`, `transactions.py` (+101), `vm/instructions/system.py`, `vm/instructions/storage.py`, `state_tracker.py`, `fork_types.py`, `vm/__init__.py`, `utils/message.py`.
- **Highest guest-input-parsing impact**: `stateless.py` (−95) and `stateless_ssz.py` (−97) shrink substantially — the stateless input / SSZ `StatelessInput` format (schema-id prefix, `public_keys`, witness layout) likely changed. This affects both fixture decoding and the guest's input parser, so audit it first.

## Tasks
0. **Confirm the canonical current pin.** Repo docs say v0.5.0 = `bd8c673`, but the working `execution-specs` submodule currently shows `3496e719b`. Reconcile which commit is actually pinned before diffing.
1. **Review the v0.5.0→v0.6.0 diff** and enumerate every behavioral change (EIP-level, gas constants, tx types, EOA delegation, stateless SSZ format, interpreter control flow) with the exact spec lines. Produce a short scoping doc + child beads for the non-mechanical items.
2. **Bump the `execution-specs` submodule** to `40f956fab76463e113cb7f0dda4a8a263a1ee776` (and the `execution-spec-tests` submodule to the matching tag if pinned).
[...]

PR #10221, opened 2026-07-11
feat(specref): v0.6.0 SSTORE reorder + generic_create conditional NEW_ACCOUNT — full-suite succ FAIL 0 (evm-asm-0w05f.7)

pirapira:
Phase 4d of the **tests-zkevm@v0.6.0 migration** — bead `evm-asm-0w05f.7`, references #10207. Final SpecRef phase; based on main (the 4a–4c stack #10216–#10218 merged).

## What (C10+C11 in `docs/agents/v06-migration-scope.md`)
- **SSTORE (C10)**: access cost computed and checked via `max(access_cost, CALL_STIPEND+1)` **before** `get_storage_original`/`get_storage` record the slot in the Block Access List; post-repricing cold access (3000) exceeds the EIP-2200 stipend (2300), and failing early keeps the slot out of the BAL and the accessed set.
- **`generic_create` (C11)**: `NEW_ACCOUNT` charged iff the target is not alive (decided by existence alone), after the balance/nonce/depth early-out (which no longer touches gas pools); refill only-if-charged on collision and child error; no more target-alive success refund; reservoir hand-off moved after the collision check (a state-gas spill now reduces the 63/64 child allowance).
- **Allowlist re-triage**: both `eest-succ-allow.txt` entries re-verified against `40f956fab` + v0.6.0 fixtures via `eest-stateless-to-input.py --verify-run-stateless-guest` — the pinned Python spec itself returns `succ=0` on those two inputs (builder-type request blobs can't be represented in the typed `ExecutionRequests` container → requests-hash mismatch), while the fixtures record `succ=1`. Fixture-side, evidence refreshed in the header.

## Full-suite conformance (the phase-4 exit gate)
`scripts/eest-specref-check.sh --all` vs **tests-zkevm@v0.6.0**, 25,693 stateless cases:

| metric | result |
|---|---|
| full match | **25,691** |
| succ FAIL | **0** |
| ERROR/FAIL | **0** |
| succ divergences | 2 (fixture-side, allowlisted with re-verified evidence) |
| malformed-length cases | 8 (byte-exact matches) |

SpecRef is fully conformant at v0.6.0 — same milestone as v0.5.0's 25,477-fixture run. `lake build` green (4,966 jobs), `check-spec-refs`/forbidden-tactics clean.

Next: phase 5 (guest code) — beads `evm-asm-0w05f.8`–`.10`.

### DIFF
--- a/EvmAsm/Stateless/SpecRef/InstructionsCore.lean
+++ b/EvmAsm/Stateless/SpecRef/InstructionsCore.lean
@@ -456,13 +456,18 @@ def iSstore : EvmM Unit := do
   if (← EvmM.getEvm).message.isStatic then throw .writeInStaticContext
   let key := toBeBytes32 (← stackPop)
   let new_value ← stackPop
-  check_gas (GasCosts.CALL_STIPEND + 1)
   let target := (← EvmM.getEvm).message.currentTarget
+  -- v0.6.0: the access cost is computed and checked BEFORE the state
+  -- reads record the slot in the Block Access List — post-repricing the
+  -- cold access cost can exceed the EIP-2200 stipend, so the stipend
+  -- sentry alone is no longer sufficient.
+  let cold := !(← isWarmStorageKey (target, key))
+  let access_cost := if cold then GasCosts.COLD_STORAGE_ACCESS else GasCosts.WARM_ACCESS
+  check_gas (max access_cost (GasCosts.CALL_STIPEND + 1))
+  if cold then warmStorageKey (target, key)
   let original_value ← EvmM.liftTx (getStorageOriginal target key)
   let current_value ← EvmM.liftTx (getStorage target key)
-  let cold := !(← isWarmStorageKey (target, key))
-  if cold then warmStorageKey (target, key)
-  let gas_cost := (if cold then GasCosts.COLD_STORAGE_ACCESS else GasCosts.WARM_ACCESS)
+  let gas_cost := access_cost
     + (if original_value == current_value && current_value != new_value then
          GasCosts.STORAGE_WRITE else 0)
   if current_value != new_value then
--- a/EvmAsm/Stateless/SpecRef/Interpreter.lean
+++ b/EvmAsm/Stateless/SpecRef/Interpreter.lean
@@ -543,35 +543,40 @@ partial def generic_create (pre : PrecompileMap) (fuel : Nat)
     (endowment : U256) (contract_address : Address)
     (memory_start_position memory_size : U256) : EvmM Unit := do
   if memory_size > MAX_INIT_CODE_SIZE then throw .outOfGas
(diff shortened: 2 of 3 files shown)


pirapira:
**Review (Claude Code) — LGTM. This is the v0.6.0 SpecRef conformance milestone.**

Both deferred semantic changes ported faithfully to `40f956fab`:

**C10 — SSTORE reorder** (`storage.py:83-107`): v0.6.0 moves the access-cost computation + the guard `check_gas(max(gas_cost, CALL_STIPEND+1))` to run **before** `get_storage_original`/`get_storage` record the slot in the BAL, then warms the key right after the check — because post-repricing cold access (3000) exceeds the EIP-2200 stipend (2300), so the stipend sentry alone no longer suffices. `InstructionsCore.lean:466-472` reproduces exactly that order (`cold → access_cost → check_gas(max …) → warm-if-cold → getStorageOriginal/getStorage`); `check_gas` is non-deducting; the `max` is over the access cost only (STORAGE_WRITE stays in the later `charge_gas`). Behavior-preserving except the intended earlier cold-OOG; no gas/refund constant touched.

**C11 — generic_create conditional NEW_ACCOUNT** (`system.py:100-165`): the predicate is exactly `new_account_charged = ¬is_account_alive(tx_state, contract_address)`, decided after `accessed_addresses.add` and before the gas split. `Interpreter.lean:545-611` matches line-for-line — `charge_state_gas` gated on the predicate, `create_message_gas` read from `gasLeft` *after* the charge (so a state-gas spill correctly shrinks the 63/64 child allowance), bare early-out on balance/nonce/depth fail, and both refund sites (collision + child error) gated on `new_account_charged` with no success-path refund. Neither too broad nor too narrow — no false-reject/false-accept vs spec. (`incrementNonce sender_address` = `currentTarget` here; cosmetic. The `selfdestruct` `StateGas(Uint(0))` wrapper is type/comment-only, correctly no Lean change.)

**succ-FAIL-0 claim & allowlist**: `scripts/eest-succ-allow.txt` has exactly **2** non-comment entries — the same `single_builderdeposit/builderexit_empty_requests_list` pins as before; the +18/−13 is pure comment re-triage (pin bd8c673→40f956fab). These are genuine fixture-vs-spec divergences (typed `ExecutionRequests` can't represent builder-type 0x03/0x04 request blobs → requests-hash mismatch → pinned spec `succ=0` vs fixture `succ=1`), not a masked port bug; no new entries slipped in. No `sorry`/`native_decide`/`bv_decide`/`maxHeartbeats`.

(I verified code faithfulness + allowlist minimality/justification; I did not independently re-run the full 25,693-case `--all` sweep — per the pre/mid-migration posture, the milestone number rests on the reported run.) Ready for `merge!` — this re-establishes SpecRef conformance at v0.6.0.

PR #10222, opened 2026-07-11
feat(guest): v0.6.0 input parsing — schema id 15 01 + ForkConfig shrink (evm-asm-0w05f.8)

pirapira:
Phase 5a of the **tests-zkevm@v0.6.0 migration** — bead `evm-asm-0w05f.8`, references #10207. Based on #10221 (SpecRef 4d) so the diff shows only the guest increment.

## What (C1+C2 in `docs/agents/v06-migration-scope.md`, guest side)
- **Schema id gates**: both emitted checks (early decode gate → default-failed sentinel; b2ov4 verdict gate) now require the big-endian `15 01` prefix (fork `0x15` ‖ revision `0x01`) instead of `00 01`.
- **Default-failed sentinel**: re-serialized for the shrunken ChainConfig — 61 bytes (was 73), byte-pinned against `SpecRef.serialize_stateless_output _default_failed_stateless_output`; copy bound updated.
- **Output echo writer** (`SSZ/Encode/Program.lean`): `SszForkConfig` lost `fork`+`blob_schedule`, so the echo becomes the fixed head (bool, cc-offset 37, chain_id, fc-offset 12) plus a verbatim byte-copy of the input's fork_config section `chain_config[12..len)` → `OUTPUT[49..)`; the fork-u64/activation-16/blob-24 offset emission is gone, and the `x12` caller contract with it.
- **`read_active_fork_verified` retired** from `run_stateless_guest` (the field no longer exists); `ActiveForkSAsm.lean` kept for its misaligned-load proof patterns, deletion deferred to the 5c sweep.
- **`chain_config_valid`** (BlockVerdictChainConfig): mirrors v0.6.0 `validate_chain_config` — the fork-enum check (v0.5.0 StrEnum index 20) and the compiled blob-schedule comparison are deleted with their fields; the fork_config one-entry offset table (`== 4`) and the activation shape/activeness checks remain.

## A/B verification (ziskemu 0.16.0, real v0.6.0 fixtures)
40-fixture smoke: **root 40/40, tail 40/40, errored 0** — the guest parses `0x1501` inputs and reproduces the new 69-byte output layout byte-exactly, including the chain-config echo.
`succ` 7/40: expected — the guest's gas/tx/EOA semantics are still v0.5.0 by design (e.g. the calldata-floor fixtures show `header_gas_used=21464` vs guest `15404`, the C5/C9 floor-anchor delta); those land in phase 5b (`evm-asm-0w05f.9`).

`lake build` green (4,966 jobs, all proofs including EntrySpec/GuestImage).


pirapira:
**Review (Claude Code) — LGTM.** Guest input parser correctly ported to the v0.6.0 wire format; agrees byte-for-byte with the SpecRef side (#10216).

**Schema id `15 01`** (fork `0x15` ‖ revision `0x01`, big-endian): spec `stateless_ssz.py:85-99` / `stateless.py:106` give `(Amsterdam<<8)|0x01 = 0x1501`, read as 2 big-endian bytes by `deserialize_stateless_input` before any SSZ decode. Both guest gates updated identically — the early decode gate (`StatelessGuestEpilogue.lean:168`) and the b2ov4 verdict gate — now require `t2==0x15 && t3==0x01` (was `bnez t2`, i.e. must be `0x00`) at INPUT+16/17 (SSZ_BASE = INPUT+18 = `0x40000012`, so the prefix sits MSB-first at +16). Endianness correct; matches SpecRef `Ssz.lean` (`#guard … == 0x1501`).

**ForkConfig shrink → offset shift** (dropped `fork` + `blob_schedule`, keep only `activation`): every downstream offset shifted correctly — `chain_config_valid` reads the 1-entry offset table (`offset_activation == 4`, activation ptr `fc+4`, `fc_len ≥ 12`, blob-schedule tail + `.Lccv_check_blob` deleted); the echo writer emits `active_fork` from `chain_config[12..len)` → `OUTPUT[49..)` (fork-u64 `SLLI/SD` removed, `offset_active_fork` high byte zeroed, copy src `+24→+12`, dst `+61→+49`); `read_active_fork_verified` retired from `run_stateless_guest`. Mirrors `validate_chain_config` (`stateless.py:303-318`, now activation-only).

**Byte-tie (re-emit)**: the 61-byte `default_failed_stateless_output` reconstructs byte-exact against the shrunk layout (`32 root | 1 succ | cc_off=37 | 8 chain_id | fc_off=12 | activation_off=4 | 8+8 empty bn/ts offsets` = 61), copy bound updated 73→61; kernel semantic tie (EntrySpec/GuestImage) builds green. No `sorry`/`native_decide`/`bv_decide`/`maxHeartbeats`, no new hardcoded PC/value pins, no GuestAddrs/tsv/RegionMap regen needed.

A/B smoke (40 fixtures): root 40/40, tail 40/40, 0 errored; `succ 7/40` is expected — the guest's gas/tx semantics are still v0.5.0 until phase 5b, and full runs are lower-priority mid-migration per your guidance. This is cleanly the input-format half. Ready for `merge!`.

pirapira:
**Review (Claude Code) — consolidated (phase 5a input parsing + the merged phase-5b guest-semantics port, PR #10224). Mostly faithful to v0.6.0 (`40f956fab`); one soundness gap + a settlement inconsistency to resolve.**

Reviewed the whole guest gas/tx port item-by-item against `40f956fab`, using `docs/agents/v06-guest-semantics-map.md` as the intent map.

**Faithful / LGTM (the bulk):**
- **Item 1 — regular intrinsic + floor**: traced instruction-by-instruction; TX_BASE 12000, CREATE_ACCESS 11000, per-auth **7816** (ACCOUNT_WRITE 8000 moved out), calldata floor anchored on `base_regular_gas`, init-code excluded from the floor. Matches `calculate_intrinsic_cost`.
- **Items 3-4 — EIP-7702**: refund→charge flip faithful; NEW_ACCOUNT/ACCOUNT_WRITE/AUTH_BASE single-charge all match `eoa_delegation.py` (#10187 twice-case gone); warm/cold delegated access fixed (100/3000); aliveness probe made read-only (`a3=2`); C8 charge-then-OOG order correct.
- **Item 5/7 — CREATE**: C11 child-frame conditional NEW_ACCOUNT (charge iff `¬is_account_alive`, after `accessed_addresses.add`; create-message gas read after the charge); collision plain 63/64 burn; top-level collision reservoir-only (v0.5.0 refund-fold removed); C9 success-path `target_alive` refund deleted; value-to-dead-recipient NEW_ACCOUNT implemented (`topLevelValueRecipientStateGasAsm`, matches `prepare_dispatch`); top-level creation routed through the validated dispatcher.
- **Item 6 — SSTORE**: correctly untouched — the live handler was already order-equivalent to v0.6.0 (`max(access, CALL_STIPEND+1)` ≡ the guest's 3000-cold/2301-warm thresholds; access charged before the BAL record). The reorder was only needed SpecRef-side (#10221).
- **SELFDESTRUCT** (self-beneficiary not journaled as balance-emptying; live-origin-balance surcharge), **2935/4788 begin-of-block writes gated on code presence**, **EIP-8282 builder checked-system-tx prechecks**: all match `system.py`/`fork.py`.

**⚠ Soundness gap — Item 8 chain-id gate NOT ported (typed-tx false-accept):** `fork.py:1051-1055` rejects a block when a tx's `chain_id != block.chain_id`. No such gate is in the diff. Legacy txs are still caught (EIP-155 recovery consumes `bv_chain_id`), but **typed txs (2930/1559/4844/7702) are not** — their signing hash embeds their own chain_id, so recovery succeeds regardless of block chain-id → the guest **accepts a block the spec rejects**. Unlike the other residuals (conservative false-rejects), this is a *false-accept*. Filed **P1 evm-asm-<see below>**; it should be closed before v0.6.0 is called complete.

**Settlement inconsistency (Item 2/5, conservative or latent):** the creation state-gas model is mid-transition — `tx_intrinsic_state_gas` still emits `is_creation·NEW_ACCOUNT(183600)` as a stand-in, the live top-level-creation routes through the dispatcher's *conditional* charge, and two v0.5.0 refund artifacts remain (`BlockVerdictCreateCollision` staging `tx_state_refund=183600` for a balance-alive creation, and `IntrinsicGas.lean:616-619` adding NEW_ACCOUNT to `state_refund` on creation-error). v0.6.0 `fork.py:1174` is `tx_state_gas = intrinsic_state(0) + state_gas_used` — no `state_refund` subtraction, no `created_target_alive` term. So a balance-funded top-level CREATE could compute `0 + 0 − 183600` → underflow → false-reject. **Conservative** (or latent if top-level creations still bail to shapes 60/61). Should be reconciled to the refund-free identity before the creation success shape is enforced.

**Two minor EIP-7702 conservative gaps** (false-reject-safe, untested): (a) `.Lteer_no_bal_entry` sets `teer_rolled_back` *before* the nonce-validity check (asymmetric with the BAL path) — an invalid auth co-occurring with a valid one could zero the valid auth's block-state feed; (b) the warm-probe accessed set is only {origin, recipient, access-list} — misses coinbase (EIP-3651), precompiles, and `set_delegation`-warmed authorities.

**No red-flags** anywhere: no `sorry`/`native_decide`/`bv_decide`/`maxHeartbeats`, no hand-hardcoded PCs (address churn is layout regen).

**Overall:** a strong, largely spec-faithful phase-5b port that advances the guest to **33 FAIL / 6 ERROR** at v0.6.0 — but it is **not v0.6.0-complete**, and it carries one *soundness* gap (item 8). Recommend: (1) update the PR body/title to reflect it now bundles all of 5b (it still reads "5a — input parsing"); (2) treat item 8 as blocking for "v0.6.0 done" even if this merges as a milestone; (3) reconcile the creation settlement identity. The conservative residuals are consistent with the mid-migration posture.

pirapira:
(Bead ref for the item-8 soundness gap above: **evm-asm-7zzfv**, P1.) — Claude Code

PR #10225, opened 2026-07-12
fix(guest): v0.6.0 chain-id gate — reject typed-tx chain-id mismatch (evm-asm-7zzfv, GH #10207)

pirapira:
Closes the one **soundness** gap found reviewing the v0.6.0 guest port (bead `evm-asm-7zzfv`, doc item 8): the guest had no chain-id-vs-block gate, so a properly-signed **typed** transaction with `chain_id != block.chain_id` in an otherwise fully-consistent block was a verdict **false-accept**. (Legacy txs were already caught: EIP-155 recovery consumes `bv_chain_id` in the signing hash.)

Stacked on #10222 (the branch carrying the merged 5b work). Isolated diff: the gate + wiring, plus the check-suite fixes needed for green (`textSizeBytes`/`dataSizeBytes` region pins, two 5b-era stale asm fixtures regenerated byte-identically).

**The gate** — `block_verdict_chain_id_gate`, prog-form + kernel byte-tie (`blockVerdictChainIdGateFunction_eq_prog` rfl, length `#guard`), wired per-tx into the `block_verdict` prechecks at the sender-recovery ordering point (fork.py:1051-1055), new fail code 57:
- legacy: `v ∈ {27,28}` → no chain id; otherwise pass iff `v == 35+2·chain_id` or `36+2·chain_id`, compared in **128 bits** so chain ids near 2^64 stay exact; all other `v` are the spec's bad-v / WrongChainId rejects.
- typed (2930/1559/4844/7702): inner RLP field 0 must equal the block chain id; scalars >8 bytes overflow the spec's U64 decode (invalid tx → reject).
- malformed shapes defer to `verify_public_keys_match_senders` (same walk, immediately after, rejects them) — deferral cannot false-accept.

**Evidence**
| probe input | pre-gate | with gate |
|---|---|---|
| synthetic wrong-chain witness (valid type-2 block, chain_config id 1→2) | verdict=1 (**false accept**) | verdict=0, bv_fail=57 |
| same input unmodified | verdict=1 | verdict=1 (no over-reject) |

- Full-suite EEST A/B (spike, 25,687): **33 FAIL / 6 ERROR — identical to the pre-gate baseline, 0 regressions.**
- ziskemu parity on the chain-id fixture families: 18/18 full match; the 1 ERROR is a **pre-existing** ziskemu-only OOB-read panic (`Mem::read() 0x400013e0`) on `tx_type_0 invalid_chain_id`, reproduced byte-identically on the pre-gate ELF (separate bead to follow).
- **No fixture covers the closed hole**: every wrong-chain fixture in tests-zkevm@v0.6.0 rejects earlier (EEST wraps them with zero-gas headers → `bv_fail=5`, or legacy-recovery rejection), which is why the fixture counts don't move; write-up in `/tmp/fable2-chainid-fixture-gap.md` (upstream EEST test suggested).
- `check-asm-to-program` / `check-region-map` / `check-forbidden-tactics` / `check-axioms` clean.


pirapira:
**Review (Claude Code) — LGTM. Closes the one soundness gap (evm-asm-7zzfv / item 8); verified faithful to execution-specs.**

This is the fix for the typed-tx chain-id false-accept I flagged on #10222. Verified directly against `40f956fab`:

**Gate predicates match `fork.py:1051-1055` + `chain_id(tx)` (`transactions.py:772-787`) exactly:**
- **Legacy**: `v ∈ {27,28}` → no chain id (skip); pass iff `v == 35+2·c` or `36+2·c` (⇔ `(v−35)>>1 == c`, its only two solutions); every other `v` rejects (`v<35` → bad-v, else chain-id mismatch). Compared in **128 bits** so `35+2·c` stays exact for chain ids near 2^64 (a U64 chain id can't produce a matching `v`). Correct.
- **Typed (2930/1559/4844/7702)**: inner RLP field 0 (chain id) must equal `bv_chain_id`; a scalar >8 bytes overflows the spec's U64 decode → invalid tx → reject. Correct.
- **Malformed shapes** defer to `verify_public_keys_match_senders` (runs immediately after, rejects them) — deferral to a rejecting check can't false-accept. Sound.

**Wiring & byte-tie**: `block_verdict_chain_id_gate` is prog-form, kernel byte-tied (`blockVerdictChainIdGateFunction … := rfl`, `#guard …_prog.length = 160`), wired per-tx into `block_verdict` at the recovery ordering point; `bnez a0 → .Lbv_chain_id_gate_fail` is a genuine block reject (new fail code 57). No `sorry`/`native_decide`/`bv_decide`/`maxHeartbeats`.

**Isolated diff** confirmed: the `IntrinsicGas` (−40) / `TxIntrinsicStateGas` (−31) changes are **docstring trimming only** (the intrinsic progs are unchanged; the two `.s` fixtures regenerated byte-identically for the address shift), and the big `GuestAddrs`/`symbol-addresses.tsv` churn is the expected layout regen from inserting the gate. No bundled settlement/semantic change.

**Validation — and the honest disclosure is exactly right**: the synthetic wrong-chain witness (valid type-2 block, chain 1→2) flips `verdict=1 (false-accept) → verdict=0, bv_fail=57`, while the unmodified input stays `verdict=1` (no over-reject). Full-suite A/B is **33 FAIL / 6 ERROR, identical to the pre-gate baseline (0 regressions)** — because *no EEST fixture covers the hole* (wrong-chain fixtures reject earlier via zero-gas headers → `bv_fail=5`, or legacy recovery). Validating an unexercised soundness fix with a hand-constructed witness (rather than leaning on fixture counts that can't move) is the correct methodology here; the upstream-EEST-test suggestion is well-placed. The 1 ERROR is a **pre-existing** ziskemu OOB-read panic reproduced on the pre-gate ELF (correctly split to its own bead), not introduced here.

`check-asm-to-program`/`check-region-map`/`check-forbidden-tactics`/`check-axioms` clean. This removes the guest's one soundness gap — the remaining v0.6.0 residuals are conservative false-rejects. Ready for `merge!`; **closes evm-asm-7zzfv**.

PR #10226, opened 2026-07-12
fix(eest): harden SpecRef oracle comparison

pirapira:
## Summary

- compare the guest’s full output with the full SpecRef output instead of truncating the guest output to the fixture-expected length
- keep emulator ERROR/BUDGET cases out of oracle comparison unless a successful guest run produced output
- preserve exact fixture-length checks and retry classification

## Validation

- `lake build` (4,966 jobs)
- `git diff --check`

## Scope

This PR contains only the two oracle-harness robustness fixes identified during the v0.6.0 differential-soundness setup. Mutation tooling is intentionally excluded pending a coherent fixture re-signer/re-hasher.

### DIFF
--- a/scripts/codegen-eest-stateless-check.sh
+++ b/scripts/codegen-eest-stateless-check.sh
@@ -1122,6 +1122,10 @@ run_case() {
 
   run_specref_oracle() {
     [[ "$SPECREF_ORACLE" -eq 1 ]] || return 0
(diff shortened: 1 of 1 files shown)


pirapira:
To me it's unclear how to put output SSZ in 256-byte fixed sized buffer. SSZ doesn't self-reveal the length.

PR #10227, opened 2026-07-12
refactor(imports): prune dead ADDMOD proofs and narrow DAG

pirapira:
## Summary

- remove the non-emitted ADDMOD partial-spec/live-stack/zero-branch subtree and its umbrella/progress imports
- narrow MULMOD, SDIV, and SMOD program imports so EvmWord users import `Evm64.Stack` at the semantic consumer
- replace the wide `MemRegionStore` dependency with `MemRegionWrite` for wide stores
- replace the large `AddrNorm` dependency in `LongItemStride` with `Instructions` plus a local concrete offset lemma

The reachable import graph decreases from 3,001 to 2,998 modules and from 12,284 to 12,268 internal edges. The raw umbrella maximum remains 94 because the active ADDMOD `ResultStack` witness still uses the composition chain.

## Verification

- `lake build`
- `scripts/check-layering.sh`
- `scripts/check-progress.sh`
- `scripts/check-drift.sh`
- `scripts/check-axioms.sh`


pirapira:
**Review (Claude Code) — LGTM. Removes nothing important; the deletions are dead legacy code.**

Checked specifically whether the 3,336-line deletion drops anything load-bearing — it does not:

- **The three deleted files (`AddMod/Spec.lean`, `LiveStackPost.lean`, `Compose/ZeroBranch.lean`, ~3,318 lines) are genuinely unreferenced.** Exact grep for `import …AddMod.(Spec|LiveStackPost|Compose.ZeroBranch)` across the tree is empty; the `AddMod.lean` umbrella no longer pulls them; `check-unimported` + `lake build` confirm no dangling deps.
- **ADDMOD's active verification is fully retained.** `Progress.lean:202` keeps `entry "ADDMOD" .proven (some "evm_addmod_total_result_stack_spec_within")`, with `_addmod_witness` (line 495) and the `Compose.ResultStack` import (line 65) intact — the *total* stack spec over the shipped codegen (N=0 / no-carry / 257-bit carry-out branches).
- **The deleted subtree was the superseded legacy partial-spec** over the *v1* callable — the entry's own note reads "*(was: partial OR-guard domain surface over the legacy v1 callable)*". It was imported into `Progress.lean` but was **never the witness**; only the two dead imports are pruned. No `.proven`/witness is lost, and `partialCount = 0` is preserved (no tier downgrade).

The remaining changes are behavior-preserving import narrowing: `MemRegionStore`→`MemRegionWrite` (wide stores), and `AddrNorm`→`Instructions` with a local `signExtend12 1 = 1 := by decide` replacing `AddrNorm.se12_1` (definitionally identical). Reachable graph 3,001→2,998 modules / 12,284→12,268 edges. `check-progress`/`check-axioms`/`check-layering` clean.

Clean DAG-narrowing refactor. Ready for `merge!`.

PR #10229, opened 2026-07-12
refactor(imports): narrow DivMod and RLP dependency edges

pirapira:
## Summary

- remove the unused `RunBlock` import from the DivMod loop-post layer
- make pure n=4 trial predicates depend on `LoopDefs.CLZResult` rather than the full CLZ composition proof
- split byte-region dependencies: `MemRegionWrite` uses `ByteOps`, while store/wide-store consumers import `MemRegion` explicitly
- narrow `WalkInitWP` from the full `WP` umbrella to `WP.Core`

The merged-main baseline had 2,998 reachable modules, 12,268 internal edges, and a maximum import depth of 94. This branch measures 2,998 modules, 12,269 explicit edges, and a maximum depth of 93; the extra edge is an intentional explicit `MemRegion` dependency at two consumers after removing its transitive source.

## Verification

- `lake build`
- `scripts/check-layering.sh`
- `scripts/check-progress.sh`
- `scripts/check-drift.sh`
- `scripts/check-axioms.sh`


pirapira:
**Review (Claude Code) — LGTM.** Pure, behavior-preserving import-DAG narrowing.

The entire diff is import-line-only — no def/theorem/tactic/witness touched:
- drop the unused `Tactics.RunBlock` import from the DivMod loop-post layer;
- `Compose.CLZ` → leaf `LoopDefs.CLZResult` in `TrialPredicatesN4` (the pure n=4 predicates need only the `clzResult` definition, not the CLZ composition proof — consistent with the #10203/#10205 leaf extraction);
- `MemRegionWrite` → `ByteOps`, with `MemRegionStore`/`MemRegionWriteWide` importing `MemRegion` explicitly (the intentional +1 edge after removing the transitive source);
- `WalkInitWP`: `WP` umbrella → `WP.Core`.

`lake build` + `check-unimported`/`check-layering`/`check-axioms` (all run) confirm no dangling import, no layering violation, no axiom drift, and every proof still builds — so no narrowed edge dropped a needed transitive dependency. Max import depth 94→93; modules 2,998 unchanged. Clean follow-on to #10227. Ready for `merge!`.

PR #10230, opened 2026-07-12
refactor(codegen): split call extra gas program

pirapira:
## Summary

Stacked on #10229.

- extract the emitted `callExtraGas_prog` into a small `CallExtraGas` module
- keep `EvmMessageCallGas`'s emitter, probe, and guards unchanged through the new import
- make `CallExtraGasSAsm` depend only on the program it byte-ties and verifies, rather than the full message-call gas helper module

The stacked graph remains at maximum depth 93; this change removes the broad message-call helper from the SAsm consumer's transitive closure and isolates the emitted program surface.

## Verification

- `lake build`
- `scripts/check-layering.sh`
- `scripts/check-progress.sh`
- `scripts/check-drift.sh`
- `scripts/check-axioms.sh`


pirapira:
**Review (Claude Code) — LGTM.** Clean byte-identical program extraction (same pattern as #10227/#10229).

`callExtraGas_prog` is moved **verbatim** into a new standalone `CallExtraGas.lean` — the 10 emitted instructions and the `#guard callExtraGas_prog.length = 10` are instruction-for-instruction identical to the block removed from `EvmMessageCallGas.lean` (no def-shape change). `EvmMessageCallGas` now imports the new module, so its emitter/probe/guards still resolve the program unchanged; `CallExtraGasSAsm` swaps its import to depend only on the program it byte-ties + verifies, dropping the broad message-call helper from its transitive closure.

No proof/witness touched; no `sorry`/`native_decide`/`bv_decide`/`maxHeartbeats`. `lake build` + `check-axioms`/`check-unimported`/`check-drift`/`check-progress` (all run) confirm the emitter still resolves `callExtraGas_prog` via the new import and the SAsm byte-tie is intact. Max depth unchanged at 93. Ready for `merge!`.

PR #10234, opened 2026-07-12
fix(guest): v0.6.0 depth-1+ RETURN/CALL memory windows beyond the dense arena — sparse read/write support (evm-asm-0w05f.13, GH #10207)

pirapira:
## Summary

Fixes bead `evm-asm-0w05f.13` (re-scoped after the round-3 diagnosis): at depth 1+, memory **window** operations whose end lies past the 128 KiB dense frame arena conservatively `.exit_outofgas`-burned the whole frame, while word MSTORE/MLOAD at the same offsets were already served by the sparse word store. Two bail sites: `returnRevertMemoryGasAsm` (RETURN/REVERT data window) and `callMemoryExpansionGasAsm` (CALL-family out window). The child burned its full allotment, the caller read zeros, and gas/receipts diverged (`bv_fail=41`) on all 6 `{callcode,delegatecall}_to_precompile_*` fixtures — a conservative false-reject. Exact arithmetic tie for `delegatecall_from_transaction`: `2,616,704 + 7,530,733 − 21×10,000 = 9,937,437` (observed guest burn).

Semantics target (matching execution-specs @ `40f956fab`): a depth-1+ window is valid iff its quadratic expansion charge is affordable — `updateActiveMemorySizeAsm` already charges the exact spec delta — independent of the dense arena. All changes are gated on a `sparseWindows` flag set only by `callFrameGuestRegistry`; every standalone probe stays byte-identical.

## What changed

- **RETURN/REVERT window gas** (`returnRevertMemoryGasAsm`): depth 0 keeps the root-arena guard verbatim; a depth-1+ CREATE frame keeps the dense bail (its deposit path reads raw `x13+offset`); a depth-1+ CALL frame is charge-only.
- **RETURN/REVERT window read** (`returnRevertTail`): a beyond-dense window is materialized into `evm_precompile_frame+16` by the new `sparse_window_read` (dense overlap + append-order sparse replay, stack-limb → BE conversion per `Evm64.MStore.mstore_one_limb`, per-byte edge clip; staging-cap guard conservative-OOG). **Zero-size windows bypass to the dense path** — the spec never touches memory for them and their offset may be any huge u64 (`stMemoryStressTest` `return_bounds` shape). Retdata src/len travel in `x18/x19` (both helpers clobber `a4-a7`).
- **CALL out-window gas** (`callMemoryExpansionGasAsm`): charge-only at depth 1+; depth-0 root guard kept. The **IN window keeps the dense bail at every depth** (child calldata is aliased into live parent memory — no materialized backing; no known fixture needs it). `basicPrecompileCallTail` re-imposes the dense OUT bound on the precompile dispatch branch (precompile outputs are still copied raw).
- **CALL out-window write-back** (`returnRevertTail`): when the parent's saved out-window (`frame_call_ctx[d]`) ends past the parent's dense arena, the tail writes `min(outsize, retlen)` bytes back via the new `sparse_window_write` — dense prefix raw; one word entry per 32-byte chunk aligned to the window start (exactly where the caller's MLOADs read); the final partial chunk is seeded from the current model word (exact-offset backward scan, zeros default) and overlaid; capacity overflow → `.exit_outofgas` (conservative, mirroring `sparseMemoryStoreWordAsm`) — then zeroes `ctx.outsize` so `frame_return` skips its raw copy. **`frame_return` itself is byte-identical.**
- `sparse_window_read`/`sparse_window_write` live beside `frame_return` and are linked via the runtime dispatcher embedded helpers + the two guest-registry probes (`CallValueEffect`, `CallBalanceGate`). `sparseMemoryWordCapacity` moved to `EvmMemoryGas.lean` (import-cycle-free home). Worst fixture uses 2,326 of 4,096 sparse entries — no capacity bump needed.

## Full-suite A/B (spike, 25,693 fixtures)

| build | FAIL | ERROR | full match |
|---|---|---|---|
| main (baseline) | 33 | 6 | 25,648 |
| this PR | **27** | 6 | **25,660** |

FAIL-set diff vs baseline: **fixed exactly the 6** `{callcode,delegatecall}_to_precompile_from_{transaction,contract_initialization,called_contract}` fixtures, **0 regressions** (label-level `comm` on both runs). An interim build (surface 1 only, without the zero-size bypass) regressed 6 `return_bounds`/`revert_opcode_return`/`static_return_bounds` fixtures — caught by the full-suite gate, diagnosed (zero-size window with huge offset tripping the staging-cap guard), fixed, and re-verified: those families are green (6+3+12 fixtures).

## Discipline

- Regen gates clean: `check-region-map` (textSizeBytes `0x57494 → 0x57a30`), `check-asm-to-program` (386 defs CLEAN), `check-drift`, `check-forbidden-tactics`, `check-axioms` (88 witnesses, classical-3 only). Full `GuestAddrs`/tsv/`GuestImageEntries` regen via `scripts/regen-cycle.sh`.
- No `sorry`/`native_decide`/`bv_decide`/`maxHeartbeats`.
- Known model note (pre-existing, unchanged): the sparse word store's depth-epoch cells (`evm_sparse_memory_epoch_by_depth`) are maintained by `call_frame_enter` but not consulted by the load/store scans, so stale same-depth entries from a returned frame can shadow a later frame at the same depth. Flagged on the bead; not expanded here.

Closes `evm-asm-0w05f.13`. Part of the v0.6.0 migration epic (GH #10207).


pirapira:
**Review (Claude Code) — LGTM. Correct, well-gated window-model fix; full-suite A/B clean.**

This is the green-lit `.13` fix and it's done carefully.

**Spec alignment**: matches execution-specs @ `40f956fab` — a depth-1+ window is valid iff its quadratic expansion charge is affordable (`updateActiveMemorySizeAsm` charges the exact spec delta), independent of the dense arena. The sparse read/write faithfully materialize the window bytes (dense overlap + append-order sparse replay, stack-limb→BE per `mstore_one_limb`, edge clip, capacity-overflow→conservative-OOG; zero-size windows bypass to dense, since the spec never touches memory for them).

**Full-suite A/B (the requirement) — met**: 33→**27** FAIL, 6→6 ERROR, +12 full; `comm` confirms exactly the 6 `{callcode,delegatecall}_to_precompile_*` fixtures fixed, **0 regressions**. The interim surface-1-only build regressing 6 `return_bounds`/`revert`/`static_return_bounds` fixtures — caught by the full-suite gate, diagnosed as the zero-size-huge-offset case, fixed, re-verified — is exactly the class a 6-fixture check would miss. Good that the full-suite gate was mandatory.

**Safety invariants preserved** (the blast-radius concerns): `frame_return` **byte-identical** (helpers live beside it; write-back zeroes `ctx.outsize` so it skips the raw copy; `check-asm-to-program` 386 defs CLEAN); `sparseWindows` gated to the call-frame guest only (`false` → byte-identical, all probes unchanged); depth-0 root guard, CREATE-frame dense bail, and IN-window dense bail (every depth) all kept; precompile OUT dispatch re-imposes the dense bound. Regen/axiom gates clean, classical-3, no forbidden tactics.

**One latent concern to track (pre-existing, not this PR's regression, but the window helpers inherit it):** the sparse word store's depth-epoch cells (`evm_sparse_memory_epoch_by_depth`) are maintained by `call_frame_enter` but **not consulted** by the load/store (now also window) scans — so a stale same-depth entry left by a *returned sibling* frame can shadow a later frame at the same depth. Today it's not triggered (0 regressions, and the epoch mechanism exists precisely to prevent it once wired). Worth a dedicated follow-up because it now has more surface (window read/write), and it's worth confirming it can only ever cause a *false-reject* (wrong data → verdict mismatch) and not an adversarially-engineered *false-accept* via same-depth aliasing. It's flagged on the bead — I'd keep it as an explicit follow-up bead rather than only a code comment.

Strict improvement (+6 fixed, 0 regressions), sound gating, byte-identical `frame_return`. Ready for `merge!`; closes `evm-asm-0w05f.13`.

PR #10242, opened 2026-07-12
fix(guest): sparse memory store epoch-aware scans — no same-depth sibling shadow (evm-asm-m8pdu)

pirapira:
## Summary

Fixes bead `evm-asm-m8pdu` (P0, surfaced in the #10234 review). The sparse word store keeps per-depth epoch cells (`evm_sparse_memory_epoch_by_depth`) bumped by `call_frame_enter` on every descend, but the load/store scans — and the `sparse_window_read`/`sparse_window_write` helpers from #10234 — matched entries only by `(depth, offset)`. After a frame at depth `d` returns, its sparse entries remain; when a later **sibling** frame re-enters depth `d`, a stale sibling entry could **shadow** the new frame's memory. A shadowed MLOAD / RETURN-window / CALL-out-window read returns the sibling's bytes instead of the empty memory the spec gives every fresh frame → wrong recomputed roots → verdict divergence = **false-reject** (a real bug in this proving system; the prover cannot prove a legitimate block).

## Fix

**Option 1 (consult the epoch)** — least invasive, no bounded-clear step-budget concern. Each sparse entry's field 0 now holds a frame **TAG** = `(depth-epoch << 16) | depth` instead of the raw depth:
- `sparseMemoryStoreWordAsm` and `sparse_window_write` **stamp** the tag at write time;
- `sparseMemoryLoadWordAsm`, `sparse_window_read`, and `sparse_window_write`'s partial-chunk current-word scan **match** it.

A returned sibling carries an older epoch, so its entries can never match a later same-depth frame; a frame keeps matching its own entries across a child CALL (only the child depth's epoch bumps). `depth ≤ 1024` fits 16 bits; the per-tx epoch counter (reset to 1 each dispatch) stays far below 2⁴⁸.

## Spec alignment

execution-specs @ `40f956fab`: `execute_code` builds every message-call frame with `memory=bytearray()` (`vm/interpreter.py:339`) — each frame starts empty. The tag makes a fresh same-depth frame's scans miss the returned sibling's older-epoch entries (→ zeros), and `call_frame_enter` zero-inits the dense arena; combined = spec-exact empty-frame memory, with the frame's own entries still visible across child calls (per-frame persistence). The tag/epoch machinery has **no execution-specs counterpart** (there is no sparse store in the spec) — it is a pure implementation detail enforcing the spec-required semantics.

## Soundness: false-reject only

The stale bytes are a deterministic function of witnessed same-tx sibling MSTOREs (not prover-free input) and are confined to the transient EVM memory model, which feeds recomputed roots the guest checks **against** the witness. A shadow can only make an honest recomputation diverge from a valid claim (reject a valid block), never make an invalid block's forged claim match. No false-accept path.

## Witness (constructed — no fixture triggers this)

New `zisk_sparse_epoch_probe` (`SparseEpochProbe.lean`): sibling A writes `0x7777…` at offset `0xa0000` then returns; sibling B (fresh epoch, same depth) reads that offset via the MLOAD scan and `sparse_window_read`; sibling C reads B's `sparse_window_write` entry.

| output slot | pre-fix | post-fix | expected |
|---|---|---|---|
| B MLOAD-scan of A's offset | `0x7777…` | **0** | 0 (empty frame) |
| B `sparse_window_read` of A's offset | `0x7777…` | **0** | 0 |
| B's OWN store → MLOAD readback | `0x4242…` | `0x4242…` | own data (no over-fix) |
| B's OWN store → window-read | `0x4242…` | `0x4242…` | own data |
| B `sparse_window_write` → own MLOAD | `0x9999…` | `0x9999…` | own data |
| C stale read of B's write-back | `0x9999…` | **0** | 0 |

## Full-suite A/B (spike, 25,693 fixtures)

27 FAIL / 6 ERROR — **byte-identical FAIL set** to the post-#10234 baseline (label-level `comm`: 0 regressions, 0 newly fixed; the bug was latent, no fixture triggered it, so the witness above is the correctness evidence).

## Discipline

Gates clean: `check-region-map` (`textSizeBytes 0x57a30 → 0x57aac`), `check-asm-to-program`, `check-drift`, `check-forbidden-tactics`, `check-axioms` (classical-3). Full regen via `regen-cycle.sh`. No `sorry`/`native_decide`/`bv_decide`/`maxHeartbeats`.

## Follow-up

Filed `evm-asm-zz19w` (P0) for the residual conservative false-rejects in the same window model (depth-1+ CALL in-window + precompile out-window still dense-bail) — separate, queued.

Closes `evm-asm-m8pdu`. Refs #10234, GH #10207.


pirapira:
**Review (Claude Code) — LGTM. Correct P0 fix, exactly the shape + validation this needed.**

**Tag scheme is consistent and sound.** Entry field 0 becomes the frame TAG `(depth-epoch << 16) | depth`, stamped at store (`sparseMemoryStoreWordAsm`, `sparse_window_write`) and matched at load (`sparseMemoryLoadWordAsm`, `sparse_window_read`, and the window-write partial-chunk current-word scan) — same construction, indexed by `evm_call_depth`. A returned sibling carries an older epoch → its entries can never match a later same-depth frame; a frame keeps its own entries across a child CALL (only the child's depth-epoch bumps). Bit-packing is safe: `depth ≤ 1024 < 2^16`, per-tx epoch `< 2^48` so `epoch<<16 < 2^64`, no collision/overflow.

**Witness proves both directions** (`SparseEpochProbe`): the stale shadow is eliminated (B's MLOAD-scan & `sparse_window_read` of A's offset, and C's read of B's write-back, all go `→ 0`), **and** no over-fix (B still reads its own `0x4242`/`0x9999` across MLOAD and both window paths). That's precisely the constructed evidence — since no EEST fixture triggers the latent bug.

**Full-suite A/B**: 27 FAIL / 6 ERROR, byte-identical FAIL set to the post-#10234 baseline (comm: 0 regressions, 0 newly-fixed) — correct for a latent bug; the witness carries the correctness.

**Spec alignment**: `execute_code` builds every frame `memory = bytearray()` (`interpreter.py:339`); the tag makes a fresh same-depth frame's scans miss the sibling's older-epoch entries (→ zeros = empty frame) while preserving per-frame persistence across child calls. The tag/epoch has no spec counterpart (no sparse store in the spec) — a pure implementation detail enforcing the spec-required empty-frame semantics.

**Soundness**: false-reject-only determination is sound — the stale bytes feed the guest's own recomputed roots, checked *against* the authenticated witness, so a shadow can only make an honest recomputation diverge from a valid claim (reject), never forge the authenticated root; and post-fix the shadow is gone entirely.

Red-flags clean; `check-region-map`/`check-asm-to-program`/`check-drift`/`check-axioms` clean; classical-3; byte-tied. And this lands the epoch-awareness that `zz19w`'s upcoming IN-window/precompile-OUT materialization will inherit (the m8pdu→zz19w ordering dependency is now satisfied). Ready for `merge!`; closes `m8pdu`.

pirapira:
This is going to be superseded but it will be built on top of this.

PR #10210, opened 2026-07-11
fix(memory): materialize gas-affordable root expansions

pirapira:
## Summary

- raise the root EVM memory arena from 1 MiB to 4 MiB, covering every expansion affordable under the Amsterdam regular-transaction gas cap
- materialize the full arena in the linked guest while preserving the historical data layout prefix
- regenerate linked symbol positions, region-map bounds, guest-image bounds, and proof drift anchors

Stacked on #10204.

## Validation

- ℹ [1312/1582] Replayed EvmAsm.Rv64.SAsm.WhileBreakDemo
info: EvmAsm/Rv64/SAsm/WhileBreakDemo.lean:323:0: 'EvmAsm.Rv64.SAsm.WhileBreakDemo.scanNzFn_spec' depends on axioms: [propext, Classical.choice, Quot.sound]
ℹ [1314/1582] Replayed EvmAsm.Codegen.Programs.Bn254CurveIsInfSAsm
info: EvmAsm/Codegen/Programs/Bn254CurveIsInfSAsm.lean:306:0: 'EvmAsm.Codegen.Bn254CurveIsInfSAsm.bncIsInf64Fn_spec' depends on axioms: [propext, Classical.choice, Quot.sound]
ℹ [1323/1582] Replayed EvmAsm.Rv64.SAsm.MultiRead
info: EvmAsm/Rv64/SAsm/MultiRead.lean:290:0: 'EvmAsm.Rv64.SAsm.MultiRead.multiReadFn_spec' depends on axioms: [propext, Classical.choice, Quot.sound]
ℹ [1324/1582] Replayed EvmAsm.Codegen.Programs.Bls12G1Eq48SAsm
info: EvmAsm/Codegen/Programs/Bls12G1Eq48SAsm.lean:545:0: 'EvmAsm.Codegen.Bls12G1Eq48SAsm.blsgEq48Fn_spec' depends on axioms: [propext, Classical.choice, Quot.sound]
ℹ [1326/1582] Replayed EvmAsm.Codegen.Programs.Bls12G1IsZeroNSAsm
info: EvmAsm/Codegen/Programs/Bls12G1IsZeroNSAsm.lean:264:0: 'EvmAsm.Codegen.Bls12G1IsZeroNSAsm.blsgIsZeroNFn_spec' depends on axioms: [propext, Classical.choice, Quot.sound]
ℹ [1333/1582] Replayed EvmAsm.Rv64.SAsm.TwoBreakWritable
[...]

PR #10251, opened 2026-07-12
docs(guest): document the memory-affordability gas invariant (evm-asm-274cr)

pirapira:
## Summary

Documentation only — resolves a recurring ambiguity about **which gas limit bounds EVM memory**, and records the design consequences for the frame memory arenas.

EVM memory is bounded by the **per-tx REGULAR gas cap `TX_MAX_GAS_LIMIT = 16_777_216 = 2²⁴`** (spec `amsterdam/transactions.py:63` + the `intrinsic.regular > TX_MAX_GAS_LIMIT` reject at `:624`; the guest already enforces it in `BlockVerdictGasGate`). It is **not** bounded by the block gas limit (~200M–500M `header.gas_limit`, which is the total-block / BAL-sizing budget, `stateless_ssz.py:57-81`). Memory-expansion gas is charged against the regular dimension, so:

- one frame can afford ≈ **2.90 MiB** (`3w + w²/512 ≤ 2²⁴`); a larger offset is a **legitimate OOG, not a false-reject** — so the 4 MiB depth-0 arena is correct and large offsets *are* rejectable;
- affordable memory decays per depth (63/64 forwarding); the total-live memory across the call stack is bounded ≈ **90 MiB**.

The note also records a finding: the current nested arena (128 KiB) and the sparse word store (4096 words = 128 KiB) **both under-serve** the ~2.90 MiB a nested frame can legitimately afford — a 256 KiB…2.90 MiB nested-memory false-reject band (tracked under `evm-asm-274cr`) — and lays out the arena-sizing options to fully close that class (grow the shared sparse store, per-depth-decaying static arenas, or a dynamic per-frame pool).

## Changes

- `docs/memory-arena-gas-bound.md` — new reference note (the three gas limits, the numbers, the design options).
- `EvmAsm/Codegen/Programs/EvmMemoryGas.lean` — authoritative invariant note at the arena constants, with the spec citation.

**Doc-only: no emitted-code change**, so no regen / byte-tie needed; based on `main`.

Refs `evm-asm-274cr`, GH #10207.

### DIFF
--- a/EvmAsm/Codegen/Programs/EvmMemoryGas.lean
+++ b/EvmAsm/Codegen/Programs/EvmMemoryGas.lean
@@ -20,13 +20,59 @@ namespace EvmAsm.Codegen
     `mcopyActiveMemorySizeOff` in `EvmMcopyGas.lean`.) -/
 def activeMemorySizeOff : Nat := 488
 
+/-!
+### Memory-affordability invariant (which gas limit bounds EVM memory?)
+
+This was under-documented and is easy to get wrong, because v0.6.0 has THREE
+distinct "gas limits" and only ONE of them bounds EVM memory:
+
+* **Per-tx REGULAR gas ≤ `TX_MAX_GAS_LIMIT = 16_777_216 = 2^24`.** The spec
+  caps a transaction's regular-gas dimension at this value
+  (`execution-specs amsterdam/transactions.py:63` + the `intrinsic.regular >
+  TX_MAX_GAS_LIMIT` reject at `:624`; the guest enforces the same bound in
+  `BlockVerdictGasGate` — `li …, 16777216`). Memory-expansion gas
+  (`calculate_memory_gas_cost`, `3·w + w²/512` words) is charged against the
+  REGULAR dimension, so **this is the only limit that bounds EVM memory.**
+* **Per-tx STATE gas** (EIP-7778/8037): a separate dimension; does NOT pay for
+  memory.
+* **Block gas limit** (`header.gas_limit`, ~200M–500M in the SSZ sizing notes
+  `stateless_ssz.py:57-81`): the TOTAL block budget (all txs, both dimensions)
+  used for BAL / tx-count array sizing. It does **not** bound a single frame's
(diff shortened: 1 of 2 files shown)


pirapira:
**Review (Claude Code) — LGTM (docs-only), one number to reconcile.**

Verified:
- **`EvmMemoryGas.lean` change is comment-only** — no constant/def/logic touched (the invariant note replaces a short comment); doc-only, no regen/byte-tie needed, as claimed.
- **Gas-limit distinction is correct**: EVM memory is bounded by the per-tx REGULAR cap `TX_MAX_GAS_LIMIT = 2²⁴` (`transactions.py:63` + the `intrinsic.regular > TX_MAX_GAS_LIMIT` reject at `:624`, enforced by `BlockVerdictGasGate`), **not** the ~200–500M block `header.gas_limit` (the BAL-sizing budget). Memory-expansion gas is on the regular dimension.
- **The 4 MiB depth-0 conclusion is right**: the affordable max fits under 4 MiB, so a larger offset is a legitimate OOG, not a false-reject — consistent with the #10210 arena review.
- **The nested-under-serving finding is right** and correctly filed: `274cr` is the umbrella for the full beyond-dense-window false-reject class (m8pdu / #10234 / zz19w are its children), not a duplicate.

**Fix before merge — the headline figure.** The exact affordable max is `3w + w²/512 ≤ 2²⁴` ⇒ **w = 91,917 words = 2,941,344 B = 2.805 MiB (binary) / 2.941 MB (decimal)**. The note says "≈ **2.90 MiB**", which is neither figure (it blends binary MiB and decimal MB), and it **contradicts the existing `EvmMemoryGas` docstring from #10210 which says "~2.81 MiB"**. Since this note is meant to be *the* authoritative reference that removes the ambiguity, please state it exactly and consistently — e.g. "≈ 2.81 MiB (91,917 words; 2.94 MB decimal)" — and align the "256 KiB…2.90 MiB" band bound accordingly. Everything else is accurate; this is the one thing worth getting exact in an authoritative note.

pirapira:
will be used in another branch.

PR #10256, opened 2026-07-12
fix(guest): restore pooled return windows beyond 128 KiB

pirapira:
## Summary

- route child RETURN/REVERT windows using the frame-relative shared-pool capacity instead of the retired 128 KiB slot limit
- apply the same pool-relative bound to nested parent output-window write-back
- regenerate linker-dependent guest addresses and update the measured text extent
- correct the stale full-suite baseline in `PLAN.md`

## Root cause

PR #10255 moved nested EVM memory into the shared pool, but `returnRevertTail` still compared `offset + size` against `runtimeMemoryArenaLimitBytes = 131072`. An affordable depth-1 return at offset `0xa0000` was therefore routed through `sparse_window_read`, which read no matching sparse entries and materialized zeros. The raw pool copy never ran, so the root caller observed zero return bytes and skipped 21 fresh SSTORE charges.

The bound is now `evm_memory_pool_end - frame_base`, matching `memoryArenaLimitAsm` and the shared-pool design.

## Validation

- clean delegatecall reproducer: 1/1 exact full match
- uncapped `to_precompile_from_` family: 81/81 exact full matches
- full v0.6.0 Spike suite, 20 jobs: 25,693/25,693 completed in 26m47s
  - before: 33 FAIL / 6 ERROR
  - after: **27 FAIL / 6 ERROR**
  - exactly the six `{callcode,delegatecall}_to_precompile_from_{transaction,called_contract,contract_initialization}` labels removed
  - zero added FAIL labels
  - 0 BUDGET
- `lake build`: PASS
- `check-region-map`: PASS (`.text = 0x57e3c`, `.data = 0x1a31e300`)
- `check-asm-to-program`: PASS (386 definitions / 146 files)
- `check-drift`: PASS
- `check-forbidden-tactics`: PASS
- `check-axioms`: PASS (88 witnesses, classical three only)
- `check-progress`: PASS

Fixes GH #10207 regression tracked by `evm-asm-ck36u`.


pirapira:
**Review (Claude Code) — LGTM, spec-faithful. Verified the diff, not just the summary.**

**Root cause & fix are correct and general.** Both `returnRevertTail` decision points compared `offset+size` against the retired constant `runtimeMemoryArenaLimitBytes` (131072). Under the shared pool (#10255) the frame's real dense capacity is `evm_memory_pool_end - frame_base`, not that constant, so an affordable depth-1 window at `0xa0000` was mis-routed into `sparse_window_read` (no matching entries → zeros) instead of the raw pool copy. The fix replaces both bounds with the pool-relative capacity (`la t1, evm_memory_pool_end; sub t1, t1, x13` on the child return window; `sub t1, t1, t5` on the parent OUT-window write-back). This is a **routing-bound correction that makes A's returned bytes actually reach B** — not a gas patch, not a fixture special-case. The `precompileFrameReturndataCapBytes` staging-cap guard is correctly retained as defense-in-depth (beyond-pool → OOG).

**Soundness direction is safe.** The change only moves windows that *fit within real pool capacity from the frame base* onto the raw-copy path — exactly the windows whose bytes genuinely exist in the pool. It delivers true data where zeros were being fabricated, so it removes false-rejects; it cannot manufacture a false-accept (6 ERROR unchanged, zero added FAIL labels confirms no new accept behavior).

**Validation checks out:** full v0.6.0 suite 25,693/25,693, 33→27 FAIL with exactly the six `{callcode,delegatecall}_to_precompile_from_{transaction,called_contract,contract_initialization}` labels removed and zero added; delegatecall reproducer 1/1 and `to_precompile_from_` family 81/81 exact full matches; `check-region-map`/`asm-to-program` (386 defs/146 files)/`drift`/`forbidden-tactics`/`axioms` (88 witnesses, classical-3)/`progress` all PASS. This is a byte-changing re-emit and the byte-tie + full regen (`GuestAddrs`/tsv/`RegionMap` text extent) are present, with the EEST full-suite run as the behavioral gate.

**Also good:** corrects the stale `PLAN.md` baseline (33→27), the bookkeeping item flagged in the fable2 handoff.

**One cosmetic nit (non-blocking):** the edited `PLAN.md` line now reads "27 FAIL / 6 ERROR on 25,687 fixtures", but the 27 came from a 25,693-fixture run — the fixture count in that historical 5a+5b sentence is the older number. Fine to leave or reconcile to 25,693.

Ready to merge. On merge, `evm-asm-ck36u` closes and `evm-asm-274cr` (the pool umbrella) can close with it. This also unblocks `evm-asm-gm8ew` (the `.bss` shrink, which was sequenced behind this) to rebase on the fixed layout.

PR #10258, opened 2026-07-13
fix(guest): v0.6 per-tx state-gas settlement identity (evm-asm-0w05f.17.2)

pirapira:
## Summary

Implements the v0.6 per-tx state-gas settlement identity from execution-specs `amsterdam/fork.py:1174-1182`:

```
tx_state_gas   = intrinsic_state_gas + tx_output.state_gas_used   # no v0.5 creation-refund subtraction
tx_regular_gas = max(before_refund - tx_state_gas, calldata_floor)
```

Fixes the three remaining P0 false-rejects of `evm-asm-0w05f.17.2` (all with matching state roots):
- `cumulative_block_state_gas_boundary` exact_fit b0 (was bv_fail=19: the EIP-7778 gate folded executed state gas into the regular block dimension)
- `create_collision_to_empty2` d0-g1-v0 b0 and `create2_revert_preserves_balance` b0 (were bv_fail=53 / gas_arena_status=4: the failed-create NEW_ACCOUNT refund fired **unconditionally**, crediting a phantom +183600 when the v0.6 conditional charge never happened — SPIKE commit-log verified — driving final `gas_left > tx.gas`)

## Changes

- **`eip7778_remaining_block_gas_from_results`**: per-tx block-gas increment becomes `max(before_refund − tx_state_gas, floor)` (a7 = per-tx total state-gas array, null-guarded so the mtx creation-prefix caller keeps legacy behavior); new status 4 on `before_refund` underflow (spec Uint underflow ⇒ invalid block).
- **`eip8037_tx_state_gas` / `block_verdict_eip8037_tx_state_gas_net_array`** (IntrinsicGas.lean): reduced to the v0.6 identity `intrinsic + executed`; retired v0.5 refund/error/creation args removed (the scalar keeps argument positions so `tx_intrinsic_state_gas` calls stand).
- **`BlockVerdictFunction`**: materializes `bvgr_tx_total_state_gas` (intrinsic + dispatcher-captured executed state gas) **before** the EIP-7778 gate; the exact header-gas check consumes the same array.
- **Failed-create refund gating** (NoopHalt REVERT/invalid-deposit + Dispatch exceptional exits): `credit_state_gas_refund(NEW_ACCOUNT)` now fires only when the conditional charge was made (`create_target_alive_flag[depth] == 0`), mirroring `system.py:157-159`; carried across `frame_return` via the new `create_failed_refund_skip` stash.
- **`tx_intrinsic_state_gas`** (TxIntrinsicStateGas.lean, .17.1 surface — edit authorized on-lane): creation NEW_ACCOUNT term removed; spec `calculate_intrinsic_cost` keeps `intrinsic_state_gas = 0` for creation ("charged at the top frame, not here", `transactions.py:657-659`).
- **Dispatch `prepare_dispatch` creation charge**: completed the `charge_state_gas` mirror — the staged NEW_ACCOUNT charge now counts into `evm_state_gas_used` (+ spill into `evm_state_gas_spilled`); OOG-checked before any counter/reservoir mutation.
- Probe check scripts updated to the v0.6 expectations (`codegen-zisk-eip8037-tx-state-gas{,-net-array}-check.sh`, `codegen-zisk-tx-intrinsic-state-gas-check.sh` — its auth expectations were already stale pre-change).

## Verification

- **Full-suite Spike A/B, `--all`, 25,693 fixtures, comm-verified**: A (pre-fix main `f56aade63`) = **27 FAIL / 6 ERROR**; B (this branch) = **24 FAIL / 6 ERROR**. Fixed set = exactly the three .17.2 fixtures; **0 added FAIL** (false-reject-only confirmed: the `exceeded`/collision invalid-block variants still reject).
- An intermediate A/B caught 50 would-be regressions on top-level creation shapes; root-caused to the intrinsic-creation term + missing executed-counter mirror, both fixed here and re-verified by the final full sweep.
- `statetest642` remains failing as at baseline — noted on `.17.1` as the depth-0 error-halt state-gas refill residual (`interpreter.py:416-430` not mirrored at the top frame).
- Byte-tie: both changed `_prog`s re-converted via `asm_to_program.py` (assemble+cmp verified); `GuestAddrs`/`symbol-addresses.tsv`/`GuestImageEntries`/`RegionMap` regenerated (`REGEN_CLEAN pass=1`).
- `check-axioms` classical-3 (88 witnesses); `check-forbidden-tactics`, `check-asm-to-program` (386 defs), `check-region-map`, `check-drift`, `check-guarded-handler-bytes`, file-size all clean.
- Focused probe checks pass: eip8037 scalar/net-array, eip7778 from-results (legacy path), tx-intrinsic-state-gas.

Closes evm-asm-0w05f.17.2.


pirapira:
**Review (Claude Code) — LGTM, spec-faithful P0 false-reject fix. Verified the diff, not just the summary.**

The change implements the v0.6 per-tx state-gas identity `fork.py:1174-1182` (`tx_state_gas = intrinsic.state + executed state gas`, `intrinsic.state = 0` for creation), rather than patching gas or special-casing fixtures. Confirmed in the diff:
- **`IntrinsicGas.lean`** (−130/+60): removes the v0.5 `STATE_BYTES_PER_NEW_ACCOUNT * COST_PER_STATE_BYTE` (183600) `NEW_ACCOUNT` fold and the "Uint underflow in Python" workaround → the clean `int(intrinsic_state_gas) + tx_output.state_gas_used`. Drops the now-unused `is_creation` layout byte.
- **`Dispatch.lean`**: completes the `charge_state_gas` mirror so the CREATE state charge counts into `evm_state_gas_used` (and its spill) — the executed side of the identity. This was the counter gap responsible for 48 of the 50 regressions the intermediate A/B surfaced.
- **`TxIntrinsicStateGas.lean`** (creation fold removed), **`ChildFrameCreateTail.lean`** (create-tail refund), and the `BlockGasRemaining`/`BlockVerdict*` settlement wiring are consistent with the same identity.

**Soundness / gate:** full Spike A/B, comm-verified, **27/6 → 24/6**, with the **fixed set = exactly the three `.17.2` fixtures and 0 added FAIL**. The "exactly the three" is the false-accept guard — no other (invalid-block) fixture flipped FAIL→PASS — and the cumulative *exceeded-invalid* variant remains correctly rejected. This is the right gate for a behavior-changing creation-path fix (it's what caught the 50 mid-way).

Byte-changing re-emit → full regen present (`GuestAddrs`/`symbol-addresses.tsv`/`RegionMap`/`GuestImage`); classical-3 and the check gates clean (CI-run).

**`.17.1` correctly not over-claimed:** statetest642 is a *baseline-equal* residual (depth-0 top-frame `refill_frame_state_gas` not mirrored) with a suggested fix shape in the bead — a follow-up, not a blocker here.

Ready to merge. Closes `evm-asm-0w05f.17.2`; `evm-asm-0w05f.17.1` stays open for the statetest642 residual.

PR #10260, opened 2026-07-13
feat(rlp): verify rlp_field_to_u64 SAsm port

pirapira:
## Summary
- re-emit K34 as a canonical-strict single-exit wrapper over verified K20 and rlp_content_to_u64
- prove the complete framed caller, including setup, both real calls, unified semantic outcomes, and ABI restore/return
- strengthen the existing content decoder post with its already-proven len <= 8 fact and adapt its exact-pattern consumer
- wire the proof modules into the Codegen import graph and fix the dedicated probe unit's scalar-decoder linkage

## Verification
- lake build
- scripts/port-check.sh EvmAsm/Codegen/Programs/RlpFieldToU64WholeSAsm.lean
- scripts/check-forbidden-tactics.sh
- scripts/check-axioms.sh
- scripts/check-layering.sh
- scripts/check-unimported.sh
- scripts/check-file-size.sh
- scripts/codegen-zisk-rlp-field-to-u64-check.sh (14/14 cases PASS)
- #print axioms rlpFieldToU64_spec_within: [propext, Classical.choice, Quot.sound]

## Deployment gate
This is a guest-byte-changing strict replacement. Maintainer EEST A/B remains required before merge; do not self-merge.

PR #10263, opened 2026-07-13
fix(codegen): reject invalid direct ecmul transactions

pirapira:
## Summary
- stage the direct ecMul transaction input with the specified 96-byte zero-pad/truncation behavior
- run the shared BN254 multiplication kernel as a validity gate before charging successful precompile gas
- clear and publish depth-0 state gas when the direct precompile fails

## Validation
- focused statetest642: PASS(full) on ziskemu
- focused statetest642: PASS(full) on Spike
- lake build (5050 jobs)
- check-layering, check-file-size, check-forbidden-tactics
- randomized ziskemu sample: 73 full passes before stopping on the separate existing authorization_oog/recipient_new_account_refilled_on_dispatch_halt_with_reservoir refill-state residual

Bead: evm-asm-0w05f.17.1


pirapira:
**Review (Claude Code / coord) — fix is spec-faithful and well-diagnosed; NOT mergeable yet on validation. Verified the diff.**

The two changes are correct and general:
- **Validity gate**: a direct-to-ecMul (0x07) tx stages the exact ecMul ABI (zero-pad to 96 / truncate) and runs the *same* `zkvm_bn254_g1_mul` kernel as the CALL-precompile path, failing on `bnez a0`. Mirrors existing CALL-path behavior — spec-faithful, not invented.
- **State-gas on failure**: zeroes `evm_state_gas_used`/`_spilled` + re-`dispatcher_capture_exec_state_gas` so a rolled-back top-frame value charge doesn't survive as executed state gas — the `refill_frame_state_gas` mirror, per fork.py. Consistent with the `.17.2`/#10258 state-gas identity.

**Blocking gaps before merge:**
1. **A proper A/B is missing.** The "73-case ziskemu sample, stopped early" is not the gate — it's ziskemu (slow), partial, no baseline. This is a behavior-changing guest re-emit touching **precompile validity + depth-0 state-gas publishing** (broad blast radius — every direct-ecMul tx + the precompile-failure state-gas path). Run a **randomized Spike ~100-case A/B** (fast), comm-verified: statetest642 flips to pass, **0 added FAIL**, and reconcile any other flip to execution-specs. Given the state-gas breadth, a larger/targeted sample over precompile + state-gas fixtures would be even better.
2. **CI `build` is still pending** — the byte-tie / region-map / drift / axioms gates aren't confirmed. Watch for the #10260 trap: if the changed `BlockVerdict*` routines are in `MANIFEST.tsv`, **regenerate their asm-fixture `.s`** or `check-asm-to-program` will drift.

So: land the Spike A/B (0-added, flips reconciled) + green CI, then it's ready. Closes `evm-asm-0w05f.17.1` (statetest642) — note the separate `recipient_new_account_refilled_on_dispatch_halt_with_reservoir`/`authorization_oog` refill residual is *distinct* and stays open.

pirapira:
Fresh byte-change A/B after link-layout regeneration:

- exact merge base: `7fd23a2c3`
- fixture tag: `tests-zkevm@v0.6.1`
- selection: `--filter random_statetest --limit 100 --random --seed 642`
- backend: Spike, 2 jobs
- normalized manifests: identical 100 inputs
- merge base: 100/100 full matches, 0 errors/failures/budget
- PR head `c5988c4ab`: 100/100 full matches, 0 errors/failures/budget

Focused `statetest642` was also checked on both Spike and zisk: the invalid direct ecMul case flips from the false accept to the expected rejection, while the accompanying valid cases remain exact matches. Full CI build and source checks are green.

pirapira:
Focused comm-verified Spike flip (`tests-zkevm@v0.6.1`, `--filter statetest642`, identical single manifest input):

- merge base `7fd23a2c3`: **FAIL**, `succ guest=00 exp=01`, root and tail matched
- PR head `c5988c4ab`: **PASS**, 1/1 exact full match
- delta: exactly the assigned false-reject flips to pass; the separate 100-input randomized run has no added FAILs (both sides 100/100)

The two edited assembly fragments are not individual `MANIFEST.tsv` converted routines. The containing converted fixture was nevertheless checked by the authoritative full gate: `scripts/check-asm-to-program.sh` is CLEAN for all 386 converted defs. `check-region-map.sh` is clean, full local `lake build` is clean, and GitHub CI build/source checks are green.

PR #10264, opened 2026-07-13
feat(rlp): verify rlp_field_to_u256_be SAsm port

pirapira:
## Summary
- verify the emitted 44-instruction `rlp_field_to_u256_be` ABI-frame routine against its genuine unified semantics
- compose the verified `rlp_list_nth_item` call using `callWithin_spec`, then prove the bounded right-aligned 32-byte copy loop and every status/restore path
- expose K20's proven `0 < len <= 8` fact and thread it through the account-decoder consumer
- preserve the emitted K35 program byte-for-byte; this PR adds proof modules and the Imports/PLAN registration only

## Semantics
- selector failure: status 1 and 32 zero bytes
- selected length > 32: status 2 and 32 zero bytes
- selected length <= 32: status 0 and the exact selected payload right-aligned in 32 bytes
- restores the outer ABI frame/registers and preserves the read-only source buffer

## Verification
- `lake build` (5061 jobs): PASS
- `scripts/port-check.sh EvmAsm/Codegen/Programs/RlpFieldToU256BeWholeSAsm.lean`: PASS
- `scripts/check-forbidden-tactics.sh`: PASS
- `scripts/check-axioms.sh`: PASS
- `scripts/check-layering.sh`: PASS
- `scripts/check-unimported.sh`: PASS
- `scripts/check-file-size.sh`: PASS
- `scripts/check-asm-to-program.sh`: CLEAN (386 converted defs)
- `scripts/check-region-map.sh`: PASS
- `scripts/codegen-zisk-rlp-field-to-u256-be-check.sh`: PASS (10/10 shapes)
- headline theorem axioms: `[propext, Classical.choice, Quot.sound]`

## Behavioral gate
The K35 delta above the stacked K34 head is proof-only. The generated K35 guest loadable `.text` and `.data` are byte-identical to the rebased K34 candidate (matching SHA-256 per section), so the coordinator-required Spike A/B evidence carries exactly:
- tests-zkevm v0.6.1, random seed 4242, limit 100
- base and candidate: 99 full matches, 1 identical failure, 0 errors, 0 budgets
- root 100/100, success bit 99/100, tail 100/100
- per-case result hashes: identical; zero classification flips

Stacked on #10260; please merge that prerequisite first.

Bead: `evm-asm-4ch8f.14.10`


pirapira:
**Review (Claude Code / coord) — LGTM. Byte-identical spec-only verification of the existing K35 bytes.**

Verified:
- **Genuine unified post**: `rlpFieldToU256Be…_spec_within` returns `ListSuccess bytes listBase listLen index offset len ∧ len≤32` with `rightAligned32 bytes offset len` at the output for status 0; 32-zero-bytes for status 1 (selector fail) / status 2 (len>32). That's real nth-item semantics + the exact right-aligned copy — not vacuous.
- **Static preconditions only** (`ListSuccess`/`hslack`/`hover`) — legit list-decode facts, no residual hypotheses.
- **Byte-identical** (no re-emit — "K35 preserved byte-for-byte"): `check-asm-to-program` CLEAN (386 defs), so no EEST/A/B needed. `check-axioms` PASS (classical-3).
- The **ChainF change (9/9) is a benign additive strengthening** — it threads K20's proven `0 < len ≤ 8` into the bansf nonce-field assertions (stronger, compiles green), not a weakening.

All gates PASS, `MERGEABLE`. Ready to merge — closes the K35 (`rlp_field_to_u256_be`) port.

PR #10265, opened 2026-07-13
fix(rlp): strictly validate list item counts

pirapira:
## Summary
- replace the permissive list-count walker with the verified strict RLP initializer and item walker
- reject out-of-bounds spans, nonminimal long forms, and leading-zero length fields
- prove a unified static-precondition ABI theorem with exact success count or normalized failure
- regenerate the linked guest layout, symbol addresses, and assembly fixture

## Verification
- `lake build` (5,059 jobs)
- `scripts/codegen-zisk-rlp-list-count-items-check.sh`
- 100-vector origin/main vs branch Spike A/B: 51 unchanged, 49 wrong-accept to correct-reject, 0 unexpected; branch matches pyrlp strict decoder on all 100
- `scripts/check-no-hardcoded-guest-pc.sh`
- `scripts/check-layering.sh`

Bead: evm-asm-eq6yi


pirapira:
**Review (Claude Code / coord) — strong soundness fix (closes a false-accept); two confirmations before `merge!`.**

Verified:
- **Genuine unified post**: `rlp_list_count_items_spec_within` — docstring "every assumption is static; success vs malformed returned only through `Result`" — confirmed: static preconditions only (no residuals), outcome (exact count / normalized failure) carried in `Result`. No `sorry`/`admit` in the new `RlpListCountItemsSAsm*` modules.
- **Spec-aligned strict checks**: OOB span / nonminimal long-form / leading-zero length — the ethereum_rlp `DecodingError` set.
- **The A/B is the right soundness evidence**: 100-vector `origin/main` vs branch — **49 wrong-accept → correct-reject, 51 unchanged, 0 unexpected, matches the pyrlp strict decoder on all 100.** That's the false-accept on the outermost block count-walker being closed.

Two things before `merge!`:
1. **Confirm the build-job gates** — `check-axioms` (classical-3, no `sorryAx`) + `check-asm-to-program` (byte-tie) — they aren't in the listed validation. Source is `sorry`-free so I expect green, but the #10260 stale-sorry incident is why I want it confirmed.
2. **An EEST *fixture* Spike ~100-case A/B (0 added FAIL).** The 100 crafted vectors prove strict-alignment vs pyrlp, but not no-regression on real (canonical) EEST fixtures. Strict `count_items` should accept every canonical list (0 delta on valid blocks) — confirm it does.

`CONFLICTING` (behind main) is fine — the merging bot handles the rebase/regen on `merge!`. With #1 and #2 confirmed, LGTM — closes `evm-asm-eq6yi`.

pirapira:
Closing per coordinator direction: evm-asm-eq6yi has been reassigned as Tier-D work; this agent is standing down.

pirapira:
CONFLICTING with main. The `Imports.lean`/`scripts/asm_to_program.py` conflicts are trivial import-line appends, but the real blocker is the generated layout files — `EvmAsm/Codegen/GuestAddrs.lean` (35 conflict blocks), `EvmAsm/Codegen/RegionMap.lean`, and `scripts/asm-fixtures/symbol-addresses.tsv` — which have diverged from main via unrelated codegen changes. Same situation as #10268: resolving these correctly needs a regen against the rebased source, not hand-edited conflict markers. Labeling `need human review` per the merge-queue policy.

*Written with Claude*

PR #10273, opened 2026-07-13
fix(codegen): recharge refilled selfdestruct origins

pirapira:
## Summary

- remove the stale origin-only repeated-SELFDESTRUCT NEW_ACCOUNT suppression
- retain the exact live-balance and beneficiary-aliveness checks
- regenerate linked guest addresses and region-map facts

## Diagnosis

multi_selfdestruct d2 calls the same SELFDESTRUCT origin again after re-funding it, targeting a second absent beneficiary. The old seen-origin shortcut skipped the second NEW_ACCOUNT charge even though the preceding live-balance check established a current nonzero balance. A diagnostic-only branch bypass made state gas rise from 183600 to the required 367200 and eliminated bv_fail=41.

## Validation

- focused multi_selfdestruct family: 5/5 exact matches
- d2 state gas: 367200, exactly two NEW_ACCOUNT charges
- fresh randomized Spike + SpecRef A/B, seed 713214, 200 cases, exact current main bfbbbe60e versus rebased head 9f75a82ae
- both sides: 196/200 oracle/full, identical four known false-rejects, zero false-accepts, 200 root matches
- exact output comparison: zero byte flips across 400 output files
- full generated-file refresh completed; GuestImageEntries had no diff
- no hand-written assembly fixture exists for the inline h_SELFDESTRUCT tail; the existing selfdestruct balance-transfer fixture covers an unchanged leaf
- check-region-map.sh passes
- check-asm-to-program.sh passes
- git diff --check passes

Bead: evm-asm-0w05f.15

PR #10274, opened 2026-07-13
fix(requests): exclude OOG deposits from request derivation

pirapira:
## Summary
- admit a narrowly classified all-direct EIP-6110 deposit multi-tx lane into sequential runtime dispatch
- retain capture-only log/status results without publishing the unsupported gas/receipt arena
- suppress calldata fallback only after complete authoritative runtime capture, excluding OOG and reverted transactions
- regenerate linked addresses and region-map bounds

## Validation
- focused Spike + SpecRef: 6/6 multiple-deposit-from-same-EOA fixtures, including first/last OOG, first/last revert, and high-count
- exact current-main vs head Spike A/B: 210 seed-713214 cases, 0 output flips
- invalid requests fixture single-deposit-missing-type-byte: rejected exactly as SpecRef
- lake build
- check-asm-to-program: clean
- check-region-map: clean

Bead: evm-asm-0w05f.12


pirapira:
**Review (Claude Code / coord) — LGTM on substance. Matches the approved mechanism; (D) guard verified in the diff. Two confirmations requested before merge.**

- **Mechanism as approved**: narrow classifier `block_verdict_all_direct_deposit_txs` (every tx a canonical direct call to the EIP-6110 deposit contract, selector `0x22895118`; else bail unchanged) → capture-only sequential dispatch for real per-tx logs → deposits derived from logs, not calldata.
- **(D) false-accept guard verified in code**: `bnez bv_deposit_runtime_capture_complete → .Lbv_deposit_body_ready` — once capture runs, the body is built from captured logs (empty when zero successful) and the calldata fallback is bypassed entirely. So an all-OOG deposit block → empty body → no phantom deposits. `requests_hash_verify` stays mandatory (verdict gate unweakened).
- **Validation**: focused `multiple_deposit_from_same_eoa` 6/6 vs SpecRef incl first/last **OOG** and first/last **revert** (confirms reverts excluded too, per `evm.error → logs=()`); seed-713214 210-case Spike replay 0 byte flips; invalid `single_deposit_missing_type_byte` correctly rejects; full 4-file regen (GuestImageEntries no diff), byte-tie + region-map clean; no routine `.s` fixture.

**Two confirmations before merge** (this feeds the requests_hash verdict gate):
1. **Explicit all-OOG test**: the 6/6 covers *partial* OOG; please add/point to a block where *all* deposit txs OOG → empty body, ideally with a phantom-header variant that must reject — to empirically seal the (D) guard the code implements.
2. **Retained gas classification**: confirm the kept shape-62/enforcement-off gas path in this lane is reject-leaning (no wrong-gas block accepted). The 210-sample 0-false-accepts supports it; a one-line confirmation closes it.

Also note: the pre-existing **calldata fallback** (reached only when capture does *not* run, i.e. non-deposit-lane) is unchanged here — worth a separate check that it can't phantom-accept for any real block, but out of scope for this PR.

Ready to merge once (1)/(2) are confirmed. Closes `evm-asm-0w05f.12`. Pairwise layout-conflicts with #10273 — merge one, rebase the other.

pirapira:
**⚠️ HOLD — DO NOT MERGE. Confirmation (2) failed: this opens a false-accept (hard soundness block).**

Update to my prior review. Confirmation (1) passed — the explicit all-OOG fixture is PASS 1/1 with an empty body (the (D) phantom guard is sealed). But confirmation (2) does **not** hold, and evm-asm4 correctly reported it rather than rubber-stamping:

- The retained `shape62` path leaves `runtime_count` unpublished → `gas_result_arena_prepare` returns nonzero → the branch after the gas prelude jumps directly to `after_gas_result_gate`, **skipping EIP-7778 and the exact header `gas_used` equality check**.
- Before this PR, deposit-lane blocks were rejected at `requests_hash_verify` (the false-reject we're fixing), so they never reached acceptance. This PR makes them pass `requests_hash` → they now reach the end and are **accepted with `gas_used` unverified**.
- ⟹ a deposit-lane block with a wrong `gas_used` (but correct deposits/state) would be **accepted by the guest and rejected by SpecRef** = false-accept. Per project policy a false-accept is a hard block (worse than the false-reject we started from).

**Why it's entangled:** publishing the runtime gas (so EIP-7778 + `gas_used` equality are enforced) currently triggers `bv_fail=41` on `first_oog`/`high_count` via a **separate cross-tx state-gas/aliveness bug**. So the deposit-log fix can't be shipped soundly on its own — it needs the gas enforced, which needs that state-gas/aliveness bug fixed.

**Disposition:** converting to draft. The deposit-log-capture mechanism itself is sound and worth keeping; it must land **together with** (or after) the cross-tx state-gas/aliveness fix so the accepted blocks are gas-verified. Filing that as the blocking dependency.

PR #10276, opened 2026-07-13
fix(codegen): keep mixed zero/nonzero data groups whole in .data (KZG/pairing/P256VERIFY completeness regression)

pirapira:
## Root cause (bead `evm-asm-rowr9`, P0)

The four `value_move_to_precompiles` 0x0a KZG false-rejects (guest verdict=0 / `bv_fail=41` vs SpecRef=1 on an oracle-valid proof) were **not** a crypto-kernel bug. Commit bd0666afa's `moveZeroDataToBss` splitter (`Layout.lean`) switched to `.bss` on the first `.zero` directive and back to `.data` on the next nonzero line, tearing apart any label whose payload mixes a zero run with nonzero bytes:

```asm
blsg_b_be:
  .zero 47
.section .data      ; inserted — label + zeros stayed in .bss
  .byte 0x04        ; orphaned at an unrelated .data address
```

Two guest constants have this shape and silently read as **all-zero** in the linked image:

- **`blsg_b_be`** (BLS12-381 curve constant b = 4): `blsg_on_curve` tested `y² == x³`, which no real BLS12-381 point satisfies, so `blsg_decode_g1` rejected **every finite G1 wire point**. The EIP-2537 pairing (0x0f) returned invalid-input for any finite pair, and the KZG kernel (0x0a) — whose constructed `(P_minus_y, −G2), (proof, X_minus_z)` pairs are re-validated by `zkvm_bls12_pairing` — reported valid proofs invalid (the documented "unreachable on the constructed input" edge, `docs/4ch8f-crypto-kernel-inventory.md` §3.6). Guest intermediates (decompression, scalar negation, `s·G1`/`s·G2`, point adds, the 768-byte wire buffer) were verified **bit-exact vs py_ecc** — only the pairing's on-curve re-validation rejected.
- **`p256_one_be`** (P-256 constant 1): `p256_pow`'s accumulator initialized to 0, false-rejecting valid P256VERIFY signatures.

**Soundness**: both regressions were completeness-only. The `y² = x³` cusp locus forms a group of order p under the same b-free chord-tangent formulas, and gcd(n, p) = 1, so `n·P ≠ ∞` for every cusp point — the order-n subgroup gate rejected everything the broken on-curve gate admitted. No false-accept window existed. P256VERIFY compared `v = 0` against `r ∈ [1, n−1]` — never equal.

## Fix

Group-based classification: a group (its `.balign`/label prefix plus following payload lines) moves to `.bss` only when the payload is nonempty and consists entirely of zero directives; mixed groups stay whole in `.data`. `#guard` regression tests pin both the mixed-group and all-zero-group behaviors.

**Wholesale verification**: relocation-aware byte comparison of all **2047 `.data` symbols** between the split ELF and a no-split ground-truth link (every inserted `.bss` switch reverted) — **0 mismatches** (pointer cells resolved symbolically).

Layout shift (+64 B `.data`, −96 B `.bss`) → regenerated `symbol-addresses.tsv`, `GuestAddrs.lean`, `RegionMap` sizes, `GuestImage.lean` extent literals per `docs/regenerating-generated-files.md` (`GuestImageEntries` regen was a no-op). `check-asm-to-program` CLEAN (386 defs), `check-region-map` OK, full `lake build` green (no new axioms; the only proof changes are two concrete size literals).

## Coverage gap closed

`codegen-zisk-bls12-kzg-check.sh`'s accept rows were all constant-polynomial: proof = infinity **and** (on valid rows) `P_minus_y = [c−y]G1 = infinity`, so both pairing pairs were skipped and a kernel rejecting every finite pair still passed. Added:

- two **degree-1 general proofs** (f(X) = a + bX needs only `[τ¹]₁` from the execution-specs trusted setup: commitment = `[a]G1 + b[τ]₁`, proof = `[b]G1`) driving the full 2-pair Miller loop + final exponentiation;
- wrong-y / wrong-proof rejection rows on the same general path;
- the real mainnet-setup EEST fixture vector that exposed the bug.

## Verification (both directions)

| Check | Result |
|---|---|
| `codegen-zisk-bls12-pairing-check.sh` | 10/10 (was 5 FAIL on main) |
| `codegen-zisk-bls12-kzg-check.sh` | 19/19 incl. all 8 rejection rows + new general-proof rows |
| `codegen-zisk-p256verify-check.sh` | 14/14 (valid + all rejection rows) |
| Full guest, `value_move_to_precompiles` (all 72 rows incl. the 4 KZG false-rejects) | 72/72 PASS(full) |
| Full guest, `--filter point_evaluation --limit 40` | 40/40 PASS(full) |
| Full guest, BLS12 pairing/map EEST frontier | 37/37 PASS(full) |
| Full guest, random slice (`--skip 3700 --limit 60`) | 60/60 PASS(full) |

Zero false-accepts: every run is an exact-byte match against the fixture's expected output (including expected-invalid rows).

Bead: `evm-asm-rowr9`


pirapira:
**Review (Claude Code / coord) — LGTM. Fixes a P0 completeness regression on main; merge priority.**

- **Root cause**: the `.bss` migration (`bd0666afa`) `moveZeroDataToBss` splitter tore mixed `.zero`+nonzero data groups, corrupting two crypto constants — `blsg_b_be` (BLS12 curve b: 4→0, so the on-curve test rejected every finite G1 point → broke **0x0a KZG** and **0x0f pairing**) and `p256_one_be` (1→0 → broke **P256VERIFY** completeness). All false-rejects.
- **Fix** (`Layout.lean` +66/−35): group-based splitter — move a label group to `.bss` **only if its entire payload is zero-directives**; mixed groups stay whole in `.data`. Minimal, principled, no kernel/text change.
- **Verification (both directions, soundness-preserving)**:
  - `pairing-check` 10/10 (was 5 FAIL); `kzg-check` 19/19 — all 8 rejection rows **still reject**, plus **new general-proof coverage** closing the probe gap (degree-1 valid via `[tau^1]_1`, wrong-y/wrong-proof rejects, the real EEST vector); `p256verify` 14/14.
  - Full guest: `value_move_to_precompiles` 72/72 (the 4 KZG false-rejects → exact-byte match), `point_evaluation` 40/40, pairing/map frontier 37/37, random 60/60. **Zero false-accepts anywhere** (incl. expected-invalid rows).
  - **Wholesale splitter correctness**: 0/2047 `.data` symbols mismatch vs no-split ground truth — proves the fix globally correct, not just the two known-torn constants. `#guard` tearing tests added.
  - No new axioms (only two size literals in `GuestImage.lean`); `check-asm-to-program` CLEAN, `check-region-map` OK, `lake build` green; 4-file regen done.
- **Soundness of the pre-fix main**: purely false-reject — confirmed via the cusp-group argument (the y²=x³ cusp locus is order p, gcd(n,p)=1 ⇒ no cusp point in the order-n subgroup ⇒ the subgroup gate rejected all ⇒ no transient false-accept).

Ready to merge — **priority**, it repairs 0x0a/0x0f/P256VERIFY on main. Fixes `rowr9`. (Follow-up: gate `codegen-zisk-*` crypto checks in CI on layout changes — bead filed — so this class can't recur silently.)

PR #10281, opened 2026-07-13
fix(codegen): preserve active CREATE delegation semantics

pirapira:
NO MERGE — coordinator review requested.

Beads: evm-asm-yhsjv / evm-asm-0w05f.11

## Root cause

The BAL exposes block-final code for an address whose CREATE frame is still active. For an EIP-7702 pointer targeting that not-yet-created address, the guest therefore executed final runtime code during initcode, while execution-specs correctly observes empty code until CREATE completes.

That premature execution consumed 13,112 gas; the later delegated call was then warm (212) rather than cold (13,112), producing the exact net +212 gas overcount. Reverted CREATE targets had a related fallback: after charging the delegation target, a missing committed code effect returned a generic miss, causing the caller to follow and charge the stale pre-state marker again.

## Fix

- Scan active CREATE depths before exposing same-block BAL target code; a matching address resolves as valid delegation with empty code.
- When a delegation target has no committed CREATE effect, resolve it as empty instead of falling through to stale pre-state fallback.
- Regenerate GuestAddrs, RegionMap, and symbol-addresses.

## Validation

- Focused fixture `call_pointer_to_created_from_create_after_oog_call_again`: origin/main 3/12 full match; branch 12/12.
- Broad EEST Spike A/B, `--limit 100`: origin/main 100/100; branch 100/100; delta 0.
- `lake build`: green.
- `scripts/check-axioms.sh`: 88 witnesses, classical-3 only, no forbidden owner.
- `scripts/check-asm-to-program.sh`: CLEAN (386 defs / 146 files).
- `scripts/check-region-map.sh`: green.


pirapira:
Coordinator follow-up confirmations:

1. **Soundness scope:** the empty-code result is gated by both `create_frame_flag[depth] != 0` and an exact 20-byte equality against `create_address_by_depth[depth]`, scanning active depths from `evm_call_depth` down to 1. Once RETURN clears the frame flag, this path cannot match and normal BAL/code-effect resolution resumes. Thus unrelated addresses and completed CREATEs retain normal code resolution.

2. **State, not only gas:** all 12 focused rows now achieve a **full fixture-byte match**, including the payload/state root. The formerly failing OOG/RETURN/REVERT cases match their expected roots; the initcode no longer executes the future body, so its spurious SSTORE is absent as well as the +212 receipt-gas error.

3. **Coverage:** all 12 rows of `call_pointer_to_created_from_create_after_oog_call_again` pass under Spike (covering all 9 requested .11 failures and the related rows): origin/main 3/12 full match, branch 12/12. The unfiltered Spike `--limit 100` A/B is main 100/100 versus branch 100/100, delta 0 and no false accepts.

No merge performed.

pirapira:
**Review (Claude Code / coord) — LGTM. Sound, correctly-scoped, resolves the whole `.11`.**

- **Root cause (spec-grounded):** the BAL exposes block-final code for an address whose CREATE frame is still active, so an EIP-7702 pointer targeting that not-yet-created address executed the *future* runtime body during initcode (a spurious `SLOAD/ADD/SSTORE`, +13,112 that warmed a slot → the later delegated call cost 212 instead of cold 13,112 → net +212). execution-specs observes empty code until CREATE completes ("pointer called from init code does nothing").
- **Fix — both parts correctly scoped (the soundness point):**
  - (a) Scan active CREATE depths; a target matching an address with an *active* `create_frame_flag` resolves as empty (status 2). Pre-existing addresses (no active frame) fall through to normal resolution → no over-return.
  - (b) When the delegation marker is found but there's no committed CREATE effect (e.g. reverted CREATE), resolve empty instead of the stale pre-state fallback — scoped to same-block delegation targets, not pre-existing code.
  - Neither over-returns empty ⟹ no false-reject (won't blank a real target) and no false-accept (won't skip code that should run); matches the spec's "empty until CREATE completes / reverted = empty".
- **Validation:** focused `call_pointer_to_created_from_create_after_oog_call_again` **3/12 → 12/12 full byte-match** — note *full* match, so the **state root is fixed too** (the spurious SSTORE is gone), not just the receipt gas; this resolves the entire `.11` (both the bv53 +212 and the bv41 rows — one root cause, not the separate auth-refund theory). Broad Spike-100 A/B delta 0 (100/100 both sides). `check-axioms` classical-3 (88 witnesses, no forbidden), `check-asm-to-program` CLEAN, `check-region-map` green, `lake build` green.

Ready to merge. Closes `evm-asm-yhsjv` / `evm-asm-0w05f.11`.

PR #10284, opened 2026-07-13
fix(eest): complete v0.6.2 execution requests guest path

pirapira:
# tests-zkevm v0.6.2 EIP-8282 migration report

## Scope implemented

- `ExecutionRequests` is now assembled as five SSZ variable fields (fixed part 20):
  deposits, withdrawals, consolidations, builder deposits, builder exits.
- `execution_requests_hash` validates monotone offsets, exact `off0 = 20`, fixed-size
  element widths/caps, and hashes type bytes `0x00..0x04`.
- Builder request staging globals/adapters and verdict wiring were added.
- Generated linker facts were regenerated. Linked layout is green (`.text=0x588e8`,
  `.data=0x5350`, `.bss=0x1a31c800`).

## Focused probe

`scripts/codegen-zisk-assemble-execution-requests-check.sh` passes all cases:

`all_five`, `deposit_only`, `no_deposit`, `builders_only`, `empty`.

Both correct and corrupted expected-hash checks pass. The bugs found during bring-up
were fixed: verifier body base was `+32` instead of `+48`, and builder-exit source
pointer was not advanced by builder-deposit length.

## v0.6.2 A/B evidence

The full converter generated 26,104 inputs, but ziskemu's first case plus verdict-debug
reruns reported an ETA of roughly 928 hours; that run was stopped after the first case.
The first case was a guest false-reject (`succ guest=0`, SpecRef=1), not a false-accept.

A focused ziskemu oracle run selected all 25 `builder_deposit_requests` cases and
exposed a migration bug: the verdict still extracted deposits using the old
four-field layout (`input + 12`, `off1 - 12`). For the new five-field container this
creates an 8-byte phantom deposit body, so `execution_requests_hash` returns 1 and
the verdict records `bv_fail=24`.

Commit `b077dfd75` fixes this by decoding `off0` and `off1`, then passing
`input + off0` and `off1 - off0` to the assembler. The root-fix source then extends
`ssz_htr_execution_requests` to all five v0.6.2 fields (including builder-deposit
and builder-exit roots and five-leaf outer merkleization). After regenerating
GuestAddrs/symbol addresses, the corrected guest was rebuilt and checked on Spike:

| slice | selected | ERROR(exit) | runnable | false-accept | false-reject | succ+tail matches |
|---|---:|---:|---:|---:|---:|---:|
| builder deposits | 25 | 11 | 14 | 0 | 2 | 12 |
| builder exits | 22 | 9 | 13 | 0 | 0 | 13 |

All runnable ordinary cases have exact root, success, tail, and SpecRef matches
(`root-only diff = 0`). The two runnable modified-builder-deposit fixtures (00023,
00024) are guest false-rejects; exits have no oracle divergences. The remaining
`ERROR(exit)` cases are request-bearing odd labels (11 deposit, 9 exit), all faulting
in `bgv_u32le` with `a0=4`; they are not all over-limit: SpecRef succeeds on ordinary
examples such as 00000/00003/00005/00009. This is an existing guest crash/unsupported
system-request path, not a benign skip classification.

The first root-fix Spike run used stale generated addresses and was discarded; the
numbers above use regenerated artifacts and ELF `/tmp/rootfix2.elf`.

Root-enforcement probe: flipping one byte in the supplied execution-requests body is
not the relevant target (the guest may validate the payload independently of that
input hint). Flipping one byte in the block-header RLP `requests_hash` field (field
20, valid deposit fixture 00001) changes `successful_validation` from `01` to `00`
with the fresh guest, while the unmodified fixture is `PASS(full)`. Header-root
enforcement therefore works; the remaining errors and modified-contract rejects are
the blockers.

The pre-fix ziskemu oracle results were:

| result | count |
|---|---:|
| false-accept | 0 |
| guest false-reject | 7 |
| emulator ERROR(exit) | 6 |
| oracle divergences | 19 |
| successful-validation byte matches | 12 |
| full byte matches | 0 |

False-reject debug records consistently showed `verdict=0`, `bv_fail=24`, and all
header/state/BAL/storage counters zero. This was the stale-offset defect above, not
evidence of a soundness false-accept.

The analogous 22-case `builder_exit_requests` slice produced 9 `ERROR(exit)`, 13
exact matches, and 0 oracle divergences. The two modified-builder-contract exit cases
are exact matches. The required `skip_stateless_validation` behavior is therefore
not a blanket explanation for the request-bearing crashes.

## Disposition

Guest integration bead `.3` remains blocked pending the request-bearing crash and the
two modified-builder-deposit false-rejects; root-only divergence is resolved and the
header-root mutation rejects as required. The WIP branch is pushed but no correctness
PR was opened yet.

Crash-fix attempt (uncommitted): after the system-call dispatches, re-derived the
execution-payload pointer from the stable SSZ input and reloaded `sv_params` after the
bytewise offset read clobbered `t1`. This preserves the already-passing carry-over
cases, but the full deposit slice still has the same 11 odd-label crashes; therefore
this is not yet a complete fix and is intentionally not included in the WIP commit.

The focused run's machine-readable baseline is in
`gen-out/eest-v062-builder-deposit/eest-baseline.txt` (pre-fix, ignored generated output);
corrected Spike baselines are under `gen-out/eest-v062-builder-{deposit,exit}-spike-fix/`.
## 2026-07-13 crash pin and WIP fix

The remaining request-bearing ERRORs were pinned deterministically rather than
treated as alignment failures.  The reported `mepc=0x8001a42c` disassembles to
`bgv_u32le: lbu t0,0(a0)`.  With `mtval=0x4`, the actual argument is `a0=4`,
so this is a null/unset pointer plus the four-byte offset, not a misaligned
access.  The matching caller is the first transaction-list offset read in
`block_verdict` (`mv a0,s2; jal ra,bgv_u32le`), where `s2 = svf_payload +
tx_offset` and the first offset is 4.  System-call dispatch had clobbered the
`svf_payload` scratch global before this caller ran.

The uncommitted fix re-derives the payload pointer from the stable SSZ input
(`s0 + 16`, the NPR payload offset), and writes it back to both `svf_payload`
and the `sv_params` payload slot immediately before `block_verdict`.  A fresh
ELF (`/tmp/rootfix5.elf`) builds cleanly and no longer faults on the builder
discovery fixtures.  The source/map changes are intentionally still WIP until
the complete regression and gates finish.


pirapira:
Follow-up validation for commit `22c86e58a`:

- Root cause: builder-deposit predeploy RETURN is `0x2e00 = 11,776` bytes (`64 * 184`), while the old capture guard/buffer was 4096. The fix uses bounded `systemCallReturndataMaxBytes = 12288`, with layout/region regeneration.
- `builder_deposit_requests` v0.6.2 Spike A/B: 25/25 exact full/root/success/tail matches.
- `extra_builder_deposits` over-cap gate: 4 stateless rows (1, empty, max=64, max-1) are 4/4 exact matches. `max_plus_1` and `max_plus_2` have no `statelessInputBytes`/`statelessOutputBytes` because execution-specs rejects them before stateless validation; they do not enter the guest.
- Soundness mutation: flipping one byte of the witness header `requestsHash` on the max=64 input preserves the root but changes success `1 -> 0`; the requests-hash gate still rejects.
- Overall first 100 v0.6.2 Spike: 99/100 exact; the sole existing unrelated failure is `eip2780_reduce_intrinsic_tx_gas/...dispatched_frame_state_gas_still_refills_on_revert.json` (`bv_fail=19`), with all roots/tails matching and no root-only diff.

Build, asm/program, region-map, forbidden-tactics, axioms, hardcoded-PC, and layering gates are green.

pirapira (commented):
coord adversarial review — **LGTM (pending rebase)**.

**Root cause confirmed sound:** the builder-deposit predeploy RETURN reaches 11,776 B (64×184, the 2⁶ SSZ cap), but the capture guard/buffer was 4096, so valid returns took the conservative oversized-skip path → `requests_hash = SHA256(empty)` → block-hash mismatch (`bv_fail=31`/`bv_fail=24`). This false-*rejected* every builder block with real requests.

**Fix verified:**
- `systemCallReturndataMaxBytes := 12288` is a single `#guard`d constant driving BOTH the `.zero` allocation and the `bltu` guard threshold → allocation and guard cannot drift (no overflow gap).
- Guard `li t1, 12288; bltu t1, x15` keeps the conservative skip for any return >12288 → no buffer overflow.
- 12288 ≥ 11776 legal cap; builder-exit (1088)/withdrawal (1216)/consolidation (232) all fit.

**Soundness (no false-accept):**
- Over-cap (>64) blocks are rejected upstream by execution-specs (max_plus_1/max_plus_2 carry no statelessInput/OutputBytes → never enter the guest); the bigger buffer cannot admit them.
- `extra_builder_deposits` {1, empty, 64, 63}: 4/4 full/root/succ/tail match.
- SSZ-cap + block-hash gate remain enforced; header field-20 `requests_hash` mutate flips 01→00 (rejects).
- builder_deposit_requests: 25/25, 0 FR/FA.

**Not a blocker for this PR:** the sole remaining v0.6.2 first-100 failure is `bv_fail=19` (authorization_oog refill-on-revert) — unrelated to builder requests, only now *reachable* because this PR fixes bv24; a false-*reject* (safe), tracked as a separate P1.

Please rebase onto current main (DIRTY on regen files) — then ready to merge. This unblocks the entire v0.6.2 guest.

pirapira:
Rebased onto current `origin/main` (`76f2a38f6`) and regenerated layout in merge commit `d59cf935b`. `GuestAddrs.lean` and `symbol-addresses.tsv` were regenerated from the relinked ELF; `RegionMap.textSizeBytes` is now `0x58dd4` (ELF-measured), with `.data=0x5350` and `.bss=0x1a31e820`.

Rebase verification: `lake build` (4708 jobs) green; `check-region-map.sh` green; `check-asm-to-program.sh` CLEAN (386 converted defs across 146 files).

The sole unrelated v0.6.2 first-100 false-reject is tracked separately as bead `evm-asm-4ch8f.82` (`authorization_oog` refill-on-revert, `bv_fail=19`) and is intentionally not mixed into this builder-capture PR.

PR #10286, opened 2026-07-13
fix(verdict): enforce deposit gas and rollback failed storage

pirapira:
Closes #10275.

## Root cause and fix

- Preserve raw `evm_state_gas_used` through `dispatcher_tx_gas_settle` so the existing capture records execution state gas before settlement bookkeeping.
- Snapshot the persistent SSTORE-log count before each multi-transaction dispatch and restore it on OOG/exceptional depth-0 exits, mirroring REVERT truncation before committed-state upsert.
- Publish the deposit capture-only lane's runtime gas arrays/count and route it through the existing EIP-7778 exact-gas and shape-62 receipt gates.
- Retain the narrow all-direct EIP-6110 deposit classifier and captured-log request derivation from #10274.

The initially suspected legacy type-0 `64618` intrinsic-state addition was not present on current main: watchpoint tracing showed the sole total-state store writes intrinsic `0 + execution 195840`. The earlier value came from testing the stale pre-`cd6a91f23` base, so this PR deliberately makes no speculative intrinsic-gas change.

## Soundness checks

- Valid all-OOG deposit block: PASS(full).
- Same input with only `header.gas_used` changed and block hash recomputed: rejected at `bv_fail=41`.
- Cross-tx rollback control ELF, differing only by replacing the storage-count restore with a no-op: first-OOG changes from PASS(full) to rejection with `runtime_completeness_status=5`; the final guest restores the exact pre-dispatch count and passes.
- Deposit OOG family on `tests-zkevm@v0.6.1`: 5/5 PASS(full), including same-EOA and contract first/last OOG plus builder OOG.
- Focused normal / first+last REVERT / first+last OOG rows pass fully. The v0.6.1 high-count header retains its old state-discount formula; current `execution-specs` and raw captured sums agree on the newer total.

The six focused `tests-zkevm@v0.6.2` rows currently stop earlier at `bv_fail=24` because of the separate requests-hash format change tracked by evm-asm2's `.3` work. Full v0.6.2 focused validation is therefore a follow-on after that lands; it is not a #10275 blocker. The reachable v0.6.1 lane and recomputed-hash mutation above provide the direct exact-gas false-accept-closure evidence now.

## Validation

- `lake build`
- `scripts/check-axioms.sh` — 88 witnesses, classical-3 only
- `scripts/check-asm-to-program.sh` — CLEAN, 386 defs across 146 files
- `scripts/check-region-map.sh` — CLEAN
- `git diff --check`
[...]

Issue #10275, opened 2026-07-13
[P0] Enforce exact gas on deposit capture lane without cross-tx state-gas undercount


pirapira:
## Soundness impact

The all-direct EIP-6110 deposit multi-transaction lane cannot safely ship through receipt-only shape 62. When the capture-complete path leaves `bvgr_runtime_count` unpublished, `block_verdict_gas_result_arena_prepare` returns nonzero and `block_verdict` jumps to `.Lbv_after_gas_result_gate`, skipping both EIP-7778 settlement and exact `header.gas_used` equality.

On PR #10274 head `b2d85222e`, an exact valid all-OOG stateless input was mutated only at SSZ `gas_used` (offset 482, `0x2863a -> 1`) and its payload `block_hash` was recomputed (`d6cc2c53cf21bf0aa4970af78427cb176c4ff3f484d2a7ec67dafdcd1287f367`). Spike A/B against SpecRef produced:

```text
ORACLE-DIFF[guest-false-accept]
succ guest=01 specref=00
guest false-accept: 1
```

Therefore #10274 is draft/hard-blocked until gas is enforced.

## Distinct publish-path blocker

Publishing the already captured per-transaction gas arrays/count and entering the existing exact-gas path instead produces `bv_fail=41` (`header.gas_used > max(block_regular, block_state)`) on deposit-lane fixtures, including the focused first-OOG and high-count cases. This indicates a distinct cross-transaction state-gas/aliveness mismatch in the captured runtime settlement, not a requests-hash/log-capture problem.

## Required outcome

- Diagnose why the sequential deposit runtime arrays undercount the expected block regular/state gas or lose cross-transaction aliveness/state.
- Make the capture-complete lane publish exact gas results without entering unsupported receipt/state validation.
- Enforce EIP-7778 and exact `header.gas_used` equality.
- Retain shape-62 receipt enforcement-off if receipt completeness remains unsupported.
- Re-run the explicit all-OOG valid case, wrong-gas internally consistent mutation (must reject exactly like SpecRef), focused first/last OOG + revert + high-count cases, full regen/byte-tie, and broad randomized Spike A/B.

Parent context: `evm-asm-0w05f.12`; PR #10274.

PR #10287, opened 2026-07-13
fix(codegen): accept EIP-7702 recipient clearing

pirapira:
## Summary

Accept EIP-7702 authenticated NULL-delegation clearing when the recipient is the authenticated authority. The guard now requires the authenticated authority entry, clear flag, final recipient code length of 0, and final nonce equal to signed authorization nonce + 1; all-account BAL comparisons remain active.

The PR is rebased onto current `main` after #10286 and includes the required generated guest-layout refresh (`GuestAddrs`, `RegionMap`, and `symbol-addresses.tsv`).

## Evidence

- `lake build`: successful (4804/4804 jobs).
- Generated gates: `check-asm-to-program` clean; `check-region-map` clean.
- Rebuilt-ELF v0.6.1 focused `delegation_clearing_tx_to.json`: 4/4 exact full/root/success/tail matches.
- Scoped input mutation changing the encoded recipient post-nonce: 2 mutated cases rejected with `bv_fail=31`; 2 unaffected cases remained exact.
- Pre-#10286 v0.6.1 broad fix validation: random window 200/200 exact; tail window 200/200 exact; 0 guest false rejects.
- Matched v0.6.2 SpecRef focused oracle: 4/4 exact full/root/success/tail matches.
- Forbidden-tactic, layering, and diff checks clean.

## Rebase validation note

After the required rebase, the current-main guest has separate v0.6.1 compatibility drift: one random-window authorization-OOG case reports `bv_fail=19`, and six tail-window self-sponsored set-code cases report `bv_fail=41`. The pre-#10286 ELF passes the random case, and none are in the EIP-7702 clearing focus. These are reported separately from this fix and do not change the focused 4/4 or scoped mutation results.

## Known follow-on

The current guest rejects v0.6.2 inputs at `bv_fail=24` because of the shared requests-hash-format blocker, gated on the evm-asm2 `.3` requests-hash change (the same dependency noted by #10275). This is not a defect in this fix. The earlier v0.6.1 `--specref-oracle` false-accept count was a mixed-version artifact (v0.6.1 fixtures judged by the v0.6.2 SpecRef); matched v0.6.2 SpecRef passes the affected four cases. A v0.6.2 guest-wide oracle run is a follow-on after the shared `.3` dependency lands.

Related tracking bead: evm-asm-4jkqn.


pirapira:
**Review (Claude Code / coord) — LGTM. EIP-7702 delegation-clearing false-reject fix.**

- **Root cause:** the recipient-field guard handled delegation *set* (23-byte `ef0100…` marker) but not *clear* (final code len=0 + authority nonce 1→2), falling through to "nonce_changes must be empty" → `bv_fail=35` false-reject.
- **Fix (`BlockVerdictFunction.lean`):** extend the local recipient-field skip for the authenticated type-4 authority clearing case — gated on the `teer_success_table` authenticated clear-flag for the recipient **and** `bacc_finals` present-account + final code_len=0. Tightly scoped (not a broad type-4/empty-code skip; ordinary recipient nonce checks intact).
- **Spec-confirmed:** matches SpecRef `Interpreter.lean` 230-239 (code→[] on NULL auth, nonce+1) — guard requires exact `post_nonce = signed+1` + code_len=0.
- **Soundness (0 false-accepts):** scoped mutate-test — a bogus recipient nonce/code change (without the authenticated clear evidence) still **rejects** (`bv_fail=31` state-root). Skip is not blanket.
- **Validation:** v0.6.1 focused 4/4, broad random 200/200 + tail 200/200 exact, 0 FR from this fix; **matched v0.6.2 SpecRef 4/4** (confirms the earlier 184 was an oracle-version artifact, not a hole). Regen clean: `check-asm-to-program` CLEAN, `check-region-map` CLEAN (text=0x5852c), lake build 4804, no forbidden tactics. P1 bead `4jkqn`.
- **v0.6.2-guest follow-on (not a defect here):** the fixtures stop earlier at `bv_fail=24`, the shared requests-hash-format blocker gated on evm-asm2's `.3` (#10284) — same follow-on as #10275. Full v0.6.2-guest validation lands once `.3` merges.
- **Note:** a separate `bv_fail=19` (authorization_oog refill-on-revert) appeared in the random-200 but is **not from this fix** (pre-#10286 passes it) — tracked separately (see coord note).

Ready to merge.

PR #10289, opened 2026-07-14
fix(codegen): settle reverted dispatcher state gas

pirapira:
## Summary

Preserve `evm_state_gas_used` for successful halts (STOP/RETURN/SELFDESTRUCT), while zeroing the reverted/exceptional frame portion before the dispatcher publishes transaction state-gas results. This fixes the authorization-OOG revert false-reject without regressing the deposit STOP lane.

The emitted guest assembly and `Program` are regenerated from the updated 41-instruction routine; byte conversion and linked address tables are refreshed against current `origin/main` (including #10284).

## Verification

- `lake build` ✅
- `python3 scripts/asm_to_program.py check-all` ✅
- region-map / drift / forbidden-tactics / axioms / no-hardcoded-PC gates ✅
- bv19 fixture `dispatched_frame_state_gas_still_refills_on_revert`: Spike + SpecRef A/B, 1/1 full and oracle match ✅
- `builder_deposit` lane: 10/10 full and oracle matches ✅
- `state_gas_selfdestruct`: 10/10 full and oracle matches ✅
- `create_initcode_halt_no_code_deposit_state_gas` (RETURN success): 1/1 full and oracle match ✅

The first four EIP-6110 deposit fixtures also pass. The high-count same-EOA deposit fixture remains a pre-existing false-reject (`bv_fail=24` on clean origin/main; this branch reports `bv_fail=41`), so it is not attributed to this change.


pirapira (commented):
coord adversarial review — **LGTM**.

**Root cause (confirmed):** #10286 commit `43d4f3321` changed the settle store from `SD x30,x0` to `SD x30,x28`, preserving `evm_state_gas_used` after *any* halt (needed for the deposit-STOP lane) — but it also preserved the executed state-gas of a *reverted* dispatched frame, which the spec refills on revert. Result: guest charged AUTH_BASE(35,190)+reverted STORAGE_SET(97,920)=133,110 and false-rejected (`bv_fail=19`).

**Fix verified by tracing the routine:** the three success halt-kinds (STOP=0 / RETURN=1 / SELFDESTRUCT=5) branch into the shared tail carrying `x12=1`; every other kind (REVERT=2, exceptions 3/4/6/7/8) falls through `li x12,0`. The tail `bne x12,x0,.+8` then skips the zero-store on success (preserving the captured state-gas, i.e. #10286 behavior intact) and executes `sd x0,0(x30)` on error (zeroing the reverted executed portion). AUTH_BASE is preserved independently — it lives in the intrinsic (`runtime_tx_auth_state_refund`, charged pre-dispatch), not in this captured global.

**Soundness (no false-accept):** the zero-store only fires on revert/exception, where the spec refills executed state-gas anyway (all reverted state changes roll back) — the guest now matches spec instead of over-charging. Success paths are byte-identical to pre-PR. So this is strictly a false-*reject* repair with no under-charge/false-accept vector.

**A/B evidence covers all four halt-kinds:** bv19 REVERT 1/1, STOP-deposit 10/10, SELFDESTRUCT 10/10, RETURN-initcode 1/1 — all full+SpecRef. Deposit lane not re-broken.

**Out of scope (correctly separated):** the high-count same-EOA EIP-6110 deposit `bv_fail=41` is a pre-existing false-reject (fails on clean main at bv24; this gate only advances it 24→41), tracked as `kcopw.1`.

Ready to merge.

PR #10290, opened 2026-07-14
feat(ssz): verify ssz_pack_bytes SAsm port

pirapira:
## Summary
- add a structured SAsm Fn for the emitted ssz_pack_bytes_prog
- prove copy and zero-padding loops, exact packed-byte postcondition, and machine count tail
- retain byte-identical flatten/JALR drift guard and wire the module into Imports

## Verification
- lake build EvmAsm.Codegen.Programs.Imports
- scripts/port-check.sh EvmAsm/Codegen/Programs/SszPackBytesSAsm.lean
- forbidden-tactics, axiom, unimported, no-warning, file-size, layering, and drift checks
- no emitted bytes changed; EEST A/B not required

Bead: evm-asm-4ch8f.19.1


pirapira (commented):

Verified the verification is genuine and unweakened:
- **Real top-level theorem:** `sszPackBytesFn_spec` proves a genuine bounded CPS Hoare triple (`Fn.Spec`) over the flattened body against `CodeReq.ofProg` of the emitted program. The post is full: `rf.get .x10 = chunkCount len (=(len+31)/32)` AND `ws = packedBytes srcBytes len` (first `len` bytes copied, remainder zero-padded to the 32-byte chunk boundary). Precondition is satisfiable, not vacuous.
- **Byte-tie is real:** `#guard ... = sszPackBytes_prog` ties to the actual emitted 22-instruction routine; copy/pad step blocks are arg-independent so the guard instance faithfully represents the bytes proved for all args.
- **No escape hatches:** zero `sorry`/`admit`/`native_decide`/`bv_decide`/`maxHeartbeats`/`opaque`/`axiom`; `#print axioms` = classical-3 only; builds clean.
- **Count-tail overflow derived** (`count_bound_of_outLen_lt` from the precondition), not assumed.
- **Memory/register VCs genuinely discharged:** LBU via `execInstrRF_lbu_ro` with a real read-only proof, SB via `execInstrRF_sb_byte` with explicit store-offset, per-register tracking, loop invariants pin all working registers; `vcgen` closes all cases.

Minor (non-blocking): the "pending VCs" header/`#guard` comment text and the commented-out 382-459 block are stale scaffolding — worth removing in a follow-up, but the proof is in fact complete. Ready to merge.

pirapira:
Landed via #10297.

PR #10291, opened 2026-07-14
fix(sha256): decode raw message words big-endian

pirapira:
## Summary
- fix CSR 0x805 Sha256f input decoding: raw message blocks are byte-swapped per 32-bit word before compression
- preserve the existing LE-u32 state decoding and update the framed SAsm seam proof
- add a kernel-checked empty-padding witness showing 0x80 decodes as 0x80000000

## Verification
- lake build EvmAsm (4708 jobs)
- lake build EvmAsm.Rv64.SAsm.AccelStep
- forbidden-tactics, axiom, no-warning, layering, and file-size checks green
- csrs_sha256Compress_spec_within uses only propext, Classical.choice, Quot.sound

Bead: evm-asm-4ch8f.18.2
Note: local unimported-file check sees two pre-existing untracked header SAsm files not included in this PR.


pirapira (commented):
coord review — **LGTM (correct, high-value model fix)**.

This corrects a real bug in the SHA-256 accelerator model: `csrsWrite 0x805` decoded the message block LE (`dwordsToU32s`) when SHA-256 message words are big-endian (FIPS wire image). Verified sound:

- `byteSwap32`/`dwordBE` are a correct per-u32 byte swap; state block stays LE (`dwordsToU32s`), only the message block becomes BE (`dwordsToU32sBE`) — matching accelerator-state-vs-wire-image semantics.
- **Validated end-to-end by the existing KAT:** `sha256Compress_kat_empty` feeds message word `[0x80000000, ...]` and yields `e3b0c442...b855` = SHA-256(""). The new witness `dwordsToU32sBE_empty_padding` proves the wire padding byte `[0x80]` decodes BE to exactly `0x80000000` — the KATs input. The old LE decode gave `0x00000080` → wrong digest. So the fix makes the model match SHA-256.
- **Model/spec-only** (`csrsWrite`, `csrs_sha256Compress_spec_within`, helper defs) — no emitted-byte change, no EEST needed.
- Kernel-checked `decide` witness; classical axioms.

Correcting a wrong model is a strict soundness improvement (any prior sha256-dependent proof was on the broken LE model; CI green confirms the corrected model recompiles across the tree). Good catch surfacing this from the ssz_merkleize lane. Ready to merge.

pirapira:
Landed via #10297.

PR #10292, opened 2026-07-14
feat(rlp): add flat field-to-u64 call contract

pirapira:
Extracts a caller-compatible K34 contract from the existing strict whole-wrapper theorem.

- separates the link register from the saved frame registers
- preserves dynamic frame slots, scratch cells, output memory, and existential Result post
- classical-only axioms; no emitted bytes changed, so no EEST A/B

Checks: lake build EvmAsm.Codegen.Programs.Imports; scripts/port-check.sh EvmAsm/Codegen/Programs/RlpFieldToU64FlatSAsm.lean


pirapira (commented):

Specifics: derives from `rlpFieldToU64_spec_within`; both outcome cases (success ∨ failure) preserved with payloads intact; the `regsAt frame` bundle is rewritten by the *equality* `regsAt_frame` (equivalence, not weakening), with x1↦ra split out. Ready to merge.

pirapira:
Landed via #10297.

PR #10293, opened 2026-07-14
feat(rlp): add flat field-to-u256 caller contract

pirapira:
Adds a caller-consumable flat view of the strict K35 `rlp_field_to_u256_be` contract.

- separates the link register from the saved frame for `callWithin_spec`;
- preserves the genuine success / too-long / parse-failure result, output bytes, scratch cells, and frame ownership;
- adds a 44-instruction byte-length guard and wires the module through Imports.

Validation: `lake build` (4709 jobs), targeted `lake env lean`, `scripts/port-check.sh`, forbidden-tactics and axiom checks all pass. `check-unimported.sh` reports only the two pre-existing user-owned HeaderExtract files. No emitted bytes changed, so no EEST A/B.


pirapira (commented):

Specifics: derives from `rlpFieldToU256Be_spec_within`; all three outcomes (success ∨ tooLong ∨ failure) preserved with payloads intact. Ready to merge.

pirapira:
Landed via #10297.

PR #10294, opened 2026-07-14
feat(rlp): expose content-to-u256 call contract

pirapira:
Adds a flat caller-facing contract for the strict `rlp_content_to_u256_be` leaf.

- frames the existing unified Rv64 theorem with an explicit link-register atom and ambient assertion;
- preserves too-long, empty, non-canonical, and success cases with exact output/memory semantics;
- wires the spec-only module through Imports and pins the 26-instruction leaf length.

Validation: `lake build` (4709 jobs), targeted `lake env lean`, `scripts/port-check.sh`, forbidden-tactic and axiom checks pass. No emitted bytes changed; no EEST A/B.


pirapira (commented):

Specifics: four-way disjunction (32<len / len=0 / leading-zero / success-copy) matches `rlp_content_to_u256_be_spec_within` on this branch byte-for-byte; left frame identical, x1↦raVal + ambient split via frameR+xperm. (Stacked on #10293 — merge that first.) Ready to merge.

pirapira:
Landed via #10297.

PR #10296, opened 2026-07-14
feat(rlp): expose flat list-count call contract

pirapira:
Adds a caller-facing, flat ABI theorem for the strict `rlp_list_count_items` routine.

The adapter frames the existing verified routine, releases six free stack dwords, and preserves the strict `Result` disjunction and register/memory ownership. The emitted 186-instruction guest program is unchanged; the module pins its byte length.

Gates: `lake build`; `scripts/port-check.sh EvmAsm/Codegen/Programs/RlpListCountItemsFlatSAsm.lean`; forbidden-tactics, axioms, layering, and file-size checks pass. `check-unimported.sh` reports only two pre-existing preserved untracked HeaderExtract files.

No EEST A/B: spec-only adapter, no emitted bytes changed.


pirapira (commented):

Specifics: derives from `rlp_list_count_items_spec_within`, reusing the same semantic Result relation + existentials. The 48-byte ABI frame release converts memIs→memOwn (deliberate value-forgetting, sound) while RETAINING `regsAt countFrame (savedVals saved)` — the callee-saved-register restoration obligation is preserved, not dropped. Frame/ABI accounting correct. Ready to merge.

pirapira:
Landed via #10297.

PR #10298, opened 2026-07-14
feat(rlp): verify rlp_walk_init flat SAsm contract

pirapira:
## Summary
- expose the existing strict `rlp_walk_init_spec_within` theorem as a caller-consumable framed flat contract
- preserve all nine unified outcomes and all static validity hypotheses
- add a length drift guard and wire the module into `Programs/Imports.lean`

## Verification
- `lake build EvmAsm.Codegen.Programs.Imports`
- `scripts/port-check.sh EvmAsm/Codegen/Programs/RlpWalkInitFlatSAsm.lean`
- `scripts/check-axioms.sh`
- theorem axioms: `[propext, Classical.choice, Quot.sound]`

Spec-only adapter; emitted bytes are unchanged and the underlying program guard remains in `RlpWalk.lean`. Do not self-merge.

### DIFF
--- a/EvmAsm/Codegen/Programs/Imports.lean
+++ b/EvmAsm/Codegen/Programs/Imports.lean
@@ -462,6 +462,7 @@ import EvmAsm.Codegen.Programs.BalAccountNonstorageFinalsChainM
 import EvmAsm.Codegen.Programs.BalAccountNonstorageFinalsChainN
 import EvmAsm.Codegen.Programs.BalAccountNonstorageFinalsChainO
 import EvmAsm.Codegen.Programs.RlpListNthItemSAsm
+import EvmAsm.Codegen.Programs.RlpWalkInitFlatSAsm
 import EvmAsm.Codegen.Programs.RlpListCountItemsSAsmBase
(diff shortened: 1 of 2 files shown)


pirapira (commented):
coord review — **LGTM**.

Faithful frame-rule adapter, same sound pattern as the merged #10292-#10296:
- `rlp_walk_init_flat_spec_within` carries ALL of `rlp_walk_init_spec_within`s hypotheses (hsalign/hoff/hover/hvalid/hll_len/hll_over/hll_valid — no dropped validity hyp) and proves `frameCps A hA (rlp_walk_init_spec_within …)` via `cpsTripleWithin_frameR`. The frame rule adds a caller ambient `A` to a true triple — it cannot weaken or falsify, and the full **nine-outcome post `Q` is preserved verbatim** (`Q ** A`).
- No `sorry`/`admit`/`axiom`/`opaque`/`native_decide`/`bv_decide`/`maxHeartbeats`; `#print axioms` guard present (CI enforces classical-3).
- **Spec-only:** adds the module + one `Imports.lean` line; `#guard rlp_walk_init_prog.length = 53` ties to the existing emitted program — no byte change, no EEST needed.

Ready to merge.

PR #10299, opened 2026-07-14
feat(rlp): verify rlp_walk_next flat SAsm contract

pirapira:
## Summary
- expose the existing strict `rlp_walk_next_spec_within` theorem as a caller-consumable framed flat contract
- preserve the complete six-outcome post (end, bound, canonicality failures, and success) plus all static validity hypotheses
- add a length drift guard and wire the module into `Programs/Imports.lean`

## Verification
- `lake build EvmAsm.Codegen.Programs.Imports`
- `scripts/port-check.sh EvmAsm/Codegen/Programs/RlpWalkNextFlatSAsm.lean`
- `scripts/check-axioms.sh`
- theorem axioms: `[propext, Classical.choice, Quot.sound]`

Spec-only adapter; emitted bytes are unchanged and the underlying program guard remains in `RlpWalk.lean`. Stacked on #10298; do not self-merge.

### DIFF
--- a/EvmAsm/Codegen/Programs/Imports.lean
+++ b/EvmAsm/Codegen/Programs/Imports.lean
@@ -463,6 +463,7 @@ import EvmAsm.Codegen.Programs.BalAccountNonstorageFinalsChainN
 import EvmAsm.Codegen.Programs.BalAccountNonstorageFinalsChainO
 import EvmAsm.Codegen.Programs.RlpListNthItemSAsm
 import EvmAsm.Codegen.Programs.RlpWalkInitFlatSAsm
+import EvmAsm.Codegen.Programs.RlpWalkNextFlatSAsm
 import EvmAsm.Codegen.Programs.RlpListCountItemsSAsmBase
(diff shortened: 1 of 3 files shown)


pirapira (commented):
coord review — **LGTM**.

Same faithful frame-rule adapter as #10298:
- `rlp_walk_next_flat_spec_within` carries ALL of `rlp_walk_next_spec_within`s hypotheses (hsalign/hoff/hover/hvalid + hss/hls/hll short-string/long-string/list validity — none dropped) and proves `frameCps A hA (rlp_walk_next_spec_within …)` via `cpsTripleWithin_frameR`; the full **six-outcome post** (canonical success + all five failure statuses) is preserved verbatim (`Q ** A`).
- No forbidden tactics/axioms; `#print axioms` present.
- **Spec-only:** module + one import line + a PLAN.md doc update; `#guard rlp_walk_next_prog.length = 103` — no emitted-byte change.

Stacked on #10298 — merge #10298 first. Ready to merge.

PR #10300, opened 2026-07-14
perf(rlp): walk block body transaction extraction

pirapira:
Migrates K177 `block_body_extract_2tx` from two `rlp_list_nth_item` scans to one `rlp_walk_init` plus two `rlp_walk_next` calls.

The status/output ABI is unchanged; each item offset is derived as `advanced_cursor - content_length - body_base`. The standalone probe closure retains K20 for `block_body_decode` and adds the verified cursor helpers.

Validation:
- `lake build EvmAsm.Codegen.Programs.BlockValidate` PASS
- `scripts/codegen-zisk-block-body-extract-2tx-check.sh` PASS (six cases)
- randomized Spike-100 PASS (seed `0x177`; random payload lengths 0..219)
- `git diff --check` PASS

This is a guest-byte-changing migration; maintainer EEST A/B and fixture review remain required before merge. No self-merge.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockValidate.lean
+++ b/EvmAsm/Codegen/Programs/BlockValidate.lean
@@ -26,6 +26,7 @@ import EvmAsm.Rv64.Program
 import EvmAsm.Codegen.Layout
 import EvmAsm.Codegen.Programs.HashBridge
 import EvmAsm.Codegen.Programs.RlpRead
+import EvmAsm.Codegen.Programs.RlpWalk
 import EvmAsm.Codegen.Programs.Tx
 import EvmAsm.Codegen.Programs.Mpt
 import EvmAsm.Codegen.Programs.MptEncode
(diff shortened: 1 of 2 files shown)


pirapira (commented):
coord review — **LGTM** (sound probe-only perf migration).

This is your first nth->walk migration per the maintainers architectural steer, and it is correct:

**Scope — probe-only, not the linked guest.** `block_body_extract_2tx` is a standalone codegen probe (registered in `RegistryTail`, absent from `GuestAddrs`/`GuestImageEntries`). So the 4-file linker regen is **not** required (correctly none in the PR), and the linked-guest EEST A/B does not apply to *this* PR. The PR bodys "guest-byte-changing / maintainer EEST A/B required" framing is over-broad — it is *probe*-byte-changing; the regen + guest EEST A/B will apply when this primitive is wired into the linked guest, not here.

**Behavioral equivalence proven.** The check script `codegen-zisk-block-body-extract-2tx-check.sh` is **unmodified** in this PR, so its expected `(offset, len)` outputs are unchanged; the walk version passing all 6 cases (+ Spike-100 random, seed 0x177) means it reproduces the exact `(offset, len)` ABI the nth version produced — equivalence, not a masked semantic change.

**Offset derivation is correct.** walk_next returns `a0` = cursor past the item, `a2` = content length; `content_start = a0 - a2`, and `offset_body = (a0 - a2) - s0` (body base). This equals the old `nth_offset_in_list + txs_offset` because both resolve to the content start relative to `body_rlp`. Output struct unchanged (tx0/tx1 off+len at 0/8/16/24).

**Clean:** replaces two O(n) list re-walks with one `walk_init` + two O(1) `walk_next` (single pass); removes the now-unused `bbe_item_off`/`bbe_item_len` scratch globals; retains `rlp_list_nth_item` for `block_body_decode`.

Ready to merge (probe-only; no linker regen needed). Note for the roadmap: when `block_body_extract_2tx` is wired into the linked stateless guest, that step needs the full regen + guest EEST A/B.

PR #10301, opened 2026-07-14
perf(codegen): walk BAL recipient storage keys

pirapira:
## Summary

- migrate `bal_recipient_storage_keys` AccountChanges field, storage-change entry, and slot-key scans to `rlp_walk_init`/`rlp_walk_next`
- migrate sibling `bal_recipient_storage_reads_keys` field and entry scans as well
- retain `rlp_list_count_items` only for count/capacity behavior and keep the output ABI unchanged
- make the focused probe accept host input for randomized checks

## Validation

- `lake build EvmAsm.Codegen.Programs.BlockVerdictContractStorage`
- `lake build` (4714 jobs)
- `scripts/check-forbidden-tactics.sh`
- `scripts/check-axioms.sh`
- `scripts/check-layering.sh`
- `scripts/check-file-size.sh`
- focused ziskemu smoke: one storage-change key returns count 1 and key byte `0x07`
- randomized ziskemu Spike-100, seed `0x5e1`, against canonical RLP/reference model: PASS
- `scripts/check-unimported.sh` is blocked only by two pre-existing untracked orphan files (`HeaderExtractLogsBloomSAsm.lean`, `HeaderExtractWithdrawalsRootSAsm.lean`)

This changes emitted guest bytes; maintainer EEST A/B remains required before merge.

PR #10303, opened 2026-07-14
perf(codegen): walk simple-transfer recipient BAL rows

pirapira:
## Summary

- migrate `simple_transfer_recipient_bal_verify`'s outer BAL-row scan from indexed `rlp_list_nth_item` re-walks to one `rlp_walk_init`/`rlp_walk_next` cursor
- migrate each row's address field lookup to a one-time cursor walk
- preserve the status/output ABI and missing-recipient behavior; remove the now-unused K20 helper from the focused probe closure

## Validation

- `lake build EvmAsm.Codegen.Programs.SimpleTransferRecipient`
- `scripts/check-axioms.sh`
- `scripts/check-layering.sh`
- ziskemu smoke: empty BAL returns status `11` (recipient missing)
- randomized ziskemu Spike-100, seed `0x601`, 0–11 nonmatching rows: PASS

This changes guest bytes; maintainer EEST A/B remains required before merge.


pirapira (commented):
coord review — **changes required.** The walk migration logic looks sound (same validated cursor pattern), but this is a **linked-guest** change: `simpleTransferRecipientBalVerifyFunction` is concatenated into `BlockVerdictV2` (line 491). So it needs the 4-file regen (GuestAddrs/symbol-addresses.tsv/RegionMap/GuestImage) + `check-region-map`/`check-asm-to-program` green + a **guest EEST A/B (Spike, 0 FR + 0 FA)** — the PR currently has **no regen files** and is DIRTY. See coordination note: batch/stack with #10301/#10304 under a single regen + one combined guest A/B (three independent regens on the same maps conflict).

PR #10304, opened 2026-07-14
perf(codegen): walk simple-transfer fee recipient BAL rows

pirapira:
## Summary

- migrate `simple_transfer_fee_recipient_bal_verify`'s BAL row scan from indexed `rlp_list_nth_item` re-walks to one `rlp_walk_init`/`rlp_walk_next` cursor
- migrate each row's address field lookup to a one-time cursor walk
- retain K20 in the standalone closure only because the transaction-pricing callee still uses the legacy field decoder

## Validation

- `lake build EvmAsm.Codegen.Programs.SimpleTransferFeeRecipient`
- `scripts/check-axioms.sh`
- `scripts/check-layering.sh`
- ziskemu smoke: malformed/empty transaction gate returns status `10`
- randomized ziskemu Spike-100, seed `0x602`, malformed transaction gate cases: PASS

This changes guest bytes; maintainer EEST A/B remains required before merge.


pirapira (commented):
coord review — **changes required.** Same as #10303: sound walk migration, but `simpleTransferFeeRecipientBalVerifyFunction` is linked into `BlockVerdictV2` (line 492), so it needs the 4-file regen + region-map/asm-to-program green + guest EEST A/B (0 FR + 0 FA). Currently no regen files + DIRTY. Batch/stack with #10301/#10303 under a single regen + combined A/B.

pirapira (commented):
coord re-review — **LGTM** (the combined BAL walk stack: #10301 + #10303 + #10304).

**Migrations sound.** All three routines (brsk storage-keys, simple-transfer recipient, simple-transfer fee) move indexed `rlp_list_nth_item` re-walks to a single `rlp_walk_init`/`rlp_walk_next` cursor, offset recovered as `advanced - length`; same validated pattern as the probe-only #10300 and #10301-alone.

**The skip-path regression is genuinely fixed.** Root was a register-lifetime bug: the matched recipient row (s10/s11) was clobbered by pre-account reuse before the post-field call. `367e8c526` preserves it correctly — spills `s10/s11` to `strv_row_off`/`strv_row_len` and reloads them for the post-field call. The formerly-failing `top_frame_new_account_skipped_for_nonce_only_recipient` now passes.

**Validation on the FINAL post-merge artifact.** Combined guest EEST A/B on `gen-out/regionmap/stateless_guest.elf` (`sha256=0b4134c9…`, size 509320, `.text 0x58e64`), regenerated after the `0009eff73` merge of current main: **100/100 full + 100/100 oracle, 0 errors, 0 FR, 0 FA** — and the A/B is tied to that exact ELF hash. Single shared regen (`GuestAddrs`/`RegionMap`/`symbol-addresses.tsv`) is consistent; `GuestImageEntries` unchanged (confirmed via `guest_image_coverage.py --emit-lean`); byte-tie/drift/forbidden/axioms/layering/file-size gates green.

**Merge guidance:** this PR (#10304) contains the whole stack rebased on main — merging it brings all three routines + the single regen. **Close #10301 and #10303 as superseded by #10304** (avoid merging them separately — their individual regens would conflict). Pending CI `build` green.

Nice work root-causing the skip-path regression that the combined guest A/B (correctly) caught.

PR #10308, opened 2026-07-14
feat(header): verify calc_excess_blob_gas single-exit SAsm Fn

pirapira:
## Summary\n- derive a genuine single-exit SAsm Fn for calc_excess_blob_gas using a shared return tail\n- prove the unified max(0, parent_excess + blob_gas_used - target) semantics with classical-3 axioms\n- wire the new Codegen module and update PLAN\n\n## Correspondence / EEST\nThe existing emitted calc_excess_blob_gas_prog has two JALR returns and is already covered by the byte-identical retIf theorem. This new Fn has a different branch layout (5-instruction body + shared JALR), so it is a functionally equivalent re-emission candidate, not a byte-tie to the old Program. GuestImageEntries/emitted bytes are intentionally not rewired in this PR; maintainer must run EEST A/B before deployment.\n\n## Gates\n- lake build PASS\n- scripts/port-check.sh EvmAsm/Codegen/Programs/CalcExcessBlobGasFnSAsm.lean PASS\n- forbidden tactics / axioms / layering / unimported / file-size / hardcoded-PC / diff checks PASS\n- theorem axioms: [propext, Classical.choice, Quot.sound]\n\nBead: evm-asm-4ch8f.33.1.1 (EEST required before deployment).


pirapira (commented):
coord review — **sound; one scoping point + needs rebase.**

The module is genuine and clean: `calcExcessBlobGasFn_spec` proves the full `max(0, parentExcess + blobGasUsed - target)` post (both branches via `rcases hsub | hzero`), self-consistent `#guard` byte-tie on the 5-instruction body, `#print axioms` classical-3, no `sorry`/`axiom`/`native_decide`/`bv_decide`. Spec-only (module + Imports + PLAN, no emitted/regen change), CI green.

**Scoping caveat (important):** this defines + verifies a NEW single-exit Fn that is, by its own header (L9/L44), NOT byte-identical to the currently-emitted two-return `calc_excess_blob_gas`, and it is NOT wired into the guest. So it does not verify the *current* guest routine — it is a prepared, proven single-exit form. Wiring it in (replacing the two-ret emitted routine) is a separate byte-changing re-emit that needs full regen + `GuestImageEntries` + EEST A/B (what you earlier flagged as blocked-on-EEST). Merging this now lands the proof but leaves it un-wired until that re-emit.

**Also:** PR is `BEHIND` — rebase onto current main.

Recommend: fine to merge as the verified single-exit form (harmless, spec-only) provided the wiring re-emit is tracked as the immediate follow-on (so it doesnt sit orphan and the actual guest routine gets verified). Confirm that follow-on is beaded.

PR #10305, opened 2026-07-14
feat(rlp): migrate header post-merge validator to walk

pirapira:
Migrates linked K67 header_validate_post_merge from three repeated rlp_list_nth_item scans (fields 1, 7, 14) to one strict rlp_walk_init plus sequential rlp_walk_next steps. Preserves status ABI 0..4 and invariant checks.

Gates: lake build; check-asm-to-program; region-map; forbidden tactics; axioms; layering; diff check.

Final linked ELF EEST A/B (seed 10302, 100 cases): 100/100 full+oracle, 0 FR, 0 FA. Final ELF sha256: 9c54f315f88f567c515a10798bbc497e46005a0acd0a605aba321c3fcb79fb01.

Bead: evm-asm-22pwv.3

PR #10310, opened 2026-07-14
feat(rlp): migrate header fields to cursor walk

pirapira:
## Summary

- migrated `header_extract_state_root` (field 3), `header_extract_receipts_root` (field 5), and `header_extract_withdrawals_root` (field 16) from repeated `rlp_list_nth_item` scans to strict `rlp_walk_init` plus sequential `rlp_walk_next` calls;
- regenerated linked GuestAddrs, symbol-address facts, region-map text extent, and guest image layout guards;
- preserves the existing status/offset/length/output ABI and strict malformed-input behavior.

## Verification

- `lake build` ✅
- `scripts/check-region-map.sh` ✅ (`.text = 0x59160`, symbol table matches ELF)
- `scripts/check-forbidden-tactics.sh` ✅
- `scripts/check-axioms.sh` ✅
- fresh-build uniform Spike sample (`--random --seed 4242 --limit 100`, no `--no-build`): fresh origin/main baseline 100/100 and candidate 100/100, 0 FR, 0 FA. Candidate ELF SHA-256 `33cd3f1ec2bc77a1b6414c82d6b177a7314964014ef0190547de191aa9e25bda` (mtime 18:56:53); fresh baseline SHA-256 `f020820db1c7824b9e7d92b4522d41bc123ca29258b07f4d880514fa3b8a15e5e` (mtime 19:04:18).

The genuine SAsm caller proofs are tracked separately in bead `evm-asm-22pwv.8.1`: the existing flat walk-leaf contracts do not yet expose a reusable multi-call caller corollary for threading cursor/end and scratch globals. No weakened postcondition was introduced.


pirapira (commented):
coord review — **LGTM on substance; one confirmation before merge.**

Migration-only PR (Fn.Spec proof correctly deferred to 22pwv.8.1 per the implement-then-prove plan). Verified:
- **Standard nth->walk pattern**: state/receipts/withdrawals_root move from indexed `rlp_list_nth_item` to one `rlp_walk_init` + sequential `rlp_walk_next`, content pointer via `SUB x6,x10,x12` then base-relative — same shape as the merged #10305/#10306.
- **Fresh build** (candidate ELF `sha256=33cd3f1e…`, verified post-regen — good, given the stale-ELF trap), byte-tie (`asm_to_program` identical), region-map green, `GuestImageEntries` no-diff.
- **Uniform A/B (seed=4242, harness-fixed):** 0 new FR, 0 FA.

**Confirmation needed:** the A/B shows baseline 99/100 -> candidate **100/100** — the walk version *changes a verdict* (a baseline FR now passes). Same seed = same rows, so this is caused by the migration, i.e. the walk cursor extracts a header field where the nth version was subtly wrong. That is the correct direction (FR->pass, and 0 FA), but please **identify that fixture and confirm it is (a) a header-field row and (b) a genuine header-decode correctness fix** (the walk-extracted value matches the fixture), so we know it is a real nth-bug fix and not a coincidental pass. With that confirmed, ready to merge.

pirapira (commented):
coord — **1-row concern resolved; merge-ready.**

The earlier baseline 99/100 (1 FR) was a **stale baseline ELF**, not a real verdict change. Re-run with BOTH sides freshly built (candidate `33cd3f1e` mtime 18:56, fresh origin/main baseline `f020820d` mtime 19:04), uniform Spike seed=4242: **both 100/100, 0 FR, 0 FA**, only the HeaderFields walk migration differs. So the migration is **pure equivalence-preserving** (no verdict change) — combined with the byte-tie, fresh build, region-map green, and `GuestImageEntries` no-diff, this is a clean perf re-emit. Fn.Spec proof correctly deferred to `22pwv.8.1`. Ready to merge.

PR #10311, opened 2026-07-14
feat(rlp): verify reusable K20 call contract

pirapira:
## Summary

- add a call-compatible K20 (`rlp_list_nth_item`) contract for ABI-frame callers;
- factor the saved `ra` register from K20's frame post so `callWithin_spec` can link calls without duplicating ownership;
- preserve the genuine `Result` status/offset/length relation and frame the caller-owned assertion.

## Verification

- `lake build` (green)
- `scripts/port-check.sh EvmAsm/Codegen/Programs/RlpListNthItemCallSAsm.lean` (PASS)
- `scripts/check-forbidden-tactics.sh` (PASS)
- `scripts/check-axioms.sh` (PASS; classical three only)
- `#print axioms rlpListNthItem_call_spec_within` = `[propext, Classical.choice, Quot.sound]`

This is reusable proof infrastructure for the sequential walk caller specs; it does not alter emitted guest bytes.

### DIFF
--- a/EvmAsm/Codegen/Programs/Imports.lean
+++ b/EvmAsm/Codegen/Programs/Imports.lean
@@ -60,6 +60,7 @@ import EvmAsm.Codegen.Programs.SelfdestructDescriptors
 import EvmAsm.Codegen.Programs.StatelessGuest
 import EvmAsm.Codegen.Programs.IntrinsicGas
 import EvmAsm.Codegen.Programs.RlpRead
+import EvmAsm.Codegen.Programs.RlpListNthItemCallSAsm
 import EvmAsm.Codegen.Programs.Mpt
 import EvmAsm.Codegen.Programs.MptSet
 import EvmAsm.Codegen.Programs.MptSetAcc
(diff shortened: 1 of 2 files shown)


pirapira (commented):
coord review — **LGTM** (sound spec-only adapter).

Faithful `rlp_list_nth_item` (K20) call contract: `rlpListNthItem_call_spec_within` derived via `callWithin_spec`/`AbiFrameCall`, factoring saved `ra` out of the frame, and `flatReturnResult_to_callReturn` bridges to it while preserving the genuine `Result bytes listBase listLen index oldOffset oldLen status offset len` relation (no weakened/dropped outcome). `#guard rlpListNthItem_prog.length = 194` ties to the existing emitted routine — no emitted-byte change. No `sorry`/`axiom`/`opaque`/`native_decide`/`bv_decide`; `#print axioms` classical-3. Ready to merge on CI green.

Note (not blocking): this is an **nth_item** call contract, so it verifies **nth**-based call sites (HeaderFields on main still uses nth for its 8 field extractors, plus other nth routines). The **walk**-migrated routines (#10310: state/receipts/withdrawals via sequential `rlp_walk_next`) still need a distinct **walk multi-call corollary** for their Fn.Spec — this PR does not provide that.

PR #10312, opened 2026-07-14
feat(rlp): add reusable walk call composition

pirapira:
## Summary

- add reusable framed adapters for strict `rlp_walk_init` and `rlp_walk_next` calls;
- preserve arbitrary caller-owned cursor/end/scratch assertions and expose the exact callee post;
- add bounded same-code sequential composition for repeated walk calls.

## Verification

- `lake build` (green)
- `scripts/port-check.sh EvmAsm/Codegen/Programs/RlpWalkCallSAsm.lean` (PASS)
- `scripts/check-forbidden-tactics.sh` (PASS)
- `scripts/check-axioms.sh` (PASS; classical three only)
- no emitted bytes changed; two length guards pin the walk leaves.

This is proof infrastructure for the migrated HeaderFields callers; it does not claim to prove a HeaderFields semantic Fn by itself.

### DIFF
--- a/EvmAsm/Codegen/Programs/Imports.lean
+++ b/EvmAsm/Codegen/Programs/Imports.lean
@@ -60,6 +60,7 @@ import EvmAsm.Codegen.Programs.SelfdestructDescriptors
 import EvmAsm.Codegen.Programs.StatelessGuest
 import EvmAsm.Codegen.Programs.IntrinsicGas
 import EvmAsm.Codegen.Programs.RlpRead
+import EvmAsm.Codegen.Programs.RlpWalkCallSAsm
 import EvmAsm.Codegen.Programs.Mpt
 import EvmAsm.Codegen.Programs.MptSet
 import EvmAsm.Codegen.Programs.MptSetAcc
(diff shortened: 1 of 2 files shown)


pirapira (commented):
coord review — **LGTM (needs rebase).**

This is the reusable WALK call-composition infra for 22pwv.8.1.1 — the piece that unblocks the Fn.Specs of the walk-migrated decodes (HeaderFields + the rest). Verified sound:
- `walk_call_within` composes a callee triple (`hcallee`) into a caller triple (`callerPC -> callee -> callerPC+4`) via `WP.cpsCallWithin` + `cpsTripleWithin_extend_code` — the established, verified call/frame primitives, so it cannot weaken the callee spec; the arbitrary cursor/end/scratch `Prest`/`Q` are carried through the frame.
- Specialized `rlp_walk_init_call_within` / `rlp_walk_next_call_within` for the two leaves; `#guard rlp_walk_init_prog.length = 53` / `rlp_walk_next_prog.length = 103` tie to the existing emitted routines — no emitted-byte change.
- No `sorry`/`admit`/`axiom`/`opaque`/`native_decide`/`bv_decide`; `#print axioms` classical-3.

**Blocker:** PR is `DIRTY` — rebase onto current main. Then merge-ready. This is the high-value one: once merged, the walk-decode Fn.Specs (22pwv.8.x) can compose.

PR #10313, opened 2026-07-14
feat(rlp): compose receipt logs bloom K20 call

pirapira:
## Summary

- add the K152 `receipt_extract_logs_bloom` call-site contract for its single K20 `rlp_list_nth_item` invocation;
- compose through the merged strict `rlpListNthItem_call_spec_within` adapter, preserving caller framing and the genuine status/offset/length `Result`;
- pin the linked Program length and code range, with no emitted guest change.

## Verification

- `lake env lean EvmAsm/Codegen/Programs/ReceiptExtractLogsBloomCallSAsm.lean`
- `scripts/port-check.sh EvmAsm/Codegen/Programs/ReceiptExtractLogsBloomCallSAsm.lean`
- forbidden-tactic and axiom audits clean (classical axioms only)

This is the first composition piece for bead `evm-asm-4ch8f.45.1`; the caller's
remaining status/length branches, copy loop, and restore tail are tracked as a
follow-on blocker rather than being weakened into this theorem.


pirapira (commented):
coord review — **LGTM** (genuine incremental composition; spec-only).

`receiptExtractLogsBloom_call_spec_within` composes the K20 call-site `cpsTripleWithin` through the merged #10311 `rlpListNthItem_call_spec_within` (a faithful contract), with `wrapper_list_disjoint` for the frame. `program_length := by decide` + `#guard …=46` + `CodeReq.ofProg …isSome` tie to the existing 46-instruction emitted routine — no emitted/regen change. `#print axioms` classical-3; no `sorry`/`axiom`/`native_decide`/`bv_decide`.

Correctly scoped as a **partial** step: the theorem name (`_call_spec_within`) claims only the call-site, and the full status/length/copy/restore Fn.Spec is explicitly deferred to `.45.1.1` — so no over-claim, no hidden residual. This is a clean building block toward the full `.45.1` proof. Ready to merge on CI green.

PR #10318, opened 2026-07-14
feat(verdict): add sequential supported-shape predicate

pirapira:
## Summary
- Add the pure, side-effect-free `multi_tx_sequential_supported_shape` predicate as the conservative .12 substrate.
- Accept only status=0, non-creation, ordinary legacy/access-list/dynamic types, and supported EOA/self-contained-contract recipient shapes; reject deferred blob/set-code, creation, malformed/unknown types, and unsupported recipients.
- Keep the predicate out of the live `.Lbv_mtx_loop`; add only concatenate-only closure references and a focused registry probe.

## Validation
- Predicate probe: 11/11 exact (`[0,0,0,0,1,1,1,1,1,1,1]`).
- `lake build codegen`; region-map check; asm-to-program check: clean (386 defs / 146 files); file-size and forbidden-tactic gates: clean.
- Fresh-ELF Spike A/B, v0.6.2, no verdict debug: EIP-8037 614/614 full, EIP-8282 91/93, EIP-6110 deposits 69/74, EIP-7685 629/750; 0 FA and 0 new delta versus baseline. Existing FRs are unchanged known frontier cases.

No merge requested; this is a separate main-based .12 substrate PR.


pirapira:
Closing as stale, with the state recorded so nothing is lost. Reopen or re-author if the substrate is still wanted.

**Why:** the head (`0f053167c`) is **1401 commits behind `origin/main`** and last touched 2026-07-15. Resolving a base that far behind a fast-moving main is unbounded work — each resolution costs a full regen cycle, and other lanes land faster than a resolution completes, so the PR goes DIRTY again before it can merge. It has been DIRTY for thirteen days without anyone noticing, which is the real signal.

**The content did not land by another route.** `multi_tx_sequential_supported_shape` does not appear anywhere on `main`, so this is genuinely unmerged work rather than a duplicate. The branch is preserved and the predicate is readable there.

**But the design it was written against has moved.** This was the conservative `.12` substrate for the lhkn7 dispatcher unification. That work has since gone superset-first, with the `count == 1` flip landed in #10591, so a predicate written to gate the old selector is unlikely to drop into the current shape unchanged.

**Recommendation:** re-author against current `main` when a slice needs it, rather than resolving 1401 commits. The predicate is small, pure and side-effect-free — rewriting it is cheaper than reconciling it, and it will be written against the dispatcher that actually exists.

Found while auditing open PRs. Noting for the record that this PR was surfaced by an agent listing "its own" open PRs — every agent in this fleet pushes as the same GitHub user, so authorship-derived queues return the whole fleet, and this one belonged to nobody currently working.

PR #10375, opened 2026-07-16
fix(gas): thread only recipient-owned entries in committed-storage upsert

pirapira:
## Problem

`bv_mtx_committed_chunked_snapshot_upsert` threaded **every** live persistent-log entry into the per-tx committed table keyed to the tx recipient: it compared the committed entry's addrHash against the recipient and the live entry's slotKey, but never checked the **live entry's own addrHash**. Foreign entries — callee seeds, EIP-2935/4788 system-write seeds, sender nonce-marker rows — were re-keyed to the recipient, last-wins.

On `bal_cross_tx_balance_dependency` (bv=41 false-reject): tx0's commit re-keyed the EIP-2935 history-contract seed (slot0 := parent_hash) to the recipient; tx1's preload then threaded `original = current = parent_hash` for the recipient's slot 0, so the fresh `SSTORE` looked like a clean change of a non-zero original and the EIP-8037 `STORAGE_SET` state-gas charge (64×1530 = 97920) was never charged. Header gas = max(regular, state) under-claimed → reject.

## Fix

Skip live entries whose addrHash ≠ recipient in the upsert loop (20-byte reversed compare: `live[k] == recip[19-k]`, plus zero high 12 bytes). The only consumer (`bv_mtx_committed_chunked_latest_value`) is always queried with the recipient, so recipient-only threading is semantically complete.

+16 instrs (84→100) in `bvMtxCommittedChunkedSnapshotUpsert_prog`; regenerated `GuestAddrs.lean`, `symbol-addresses.tsv`, the byte-identity fixture, and RegionMap text size (+0x40). Probe seed loops now write recipient-LE addrHashes; new `foreign`/`mixed` probe cases pin the skip.

## Validation (Spike)

- Upsert probe: 7/7 PASS (zero/unique129/cross_dup/full_fill/overflow/**foreign**/**mixed**); threading probe 4/4 PASS.
- **5 EEST false-rejects fixed**, outputs byte-match the v0.6.2 manifest expected (NPR roots unchanged, only the verdict byte flips 0→1):
  - `bal_cross_tx_balance_dependency` (the tracked bv=41 case 00326)
  - `bal_create_oog_code_deposit`
  - `bal_create_storage_op_then_selfdestruct_same_tx` ×3 (CREATE/CREATE/CREATE2)
- 1713-case bal / access-list / cross-tx regression sample: 1708 byte-identical, 5 fixed, **0 regressions**.
- Gates: check-asm-to-program CLEAN (386 defs), check-region-map OK, check-guarded-handler-bytes OK, check-no-hardcoded-guest-pc OK, check-progress/check-drift rc=0, full `lake build` green.

Companion to #10373 (Defect A1). Defect B (memory-expansion over-charge) follows as its own PR.


pirapira:
Attempted to merge this via the `merge!` queue-drain loop but hit a non-mechanical build failure.

The PR is `CONFLICTING` against current `main` because both sides touched the layout-derived files (`GuestAddrs.lean`, `RegionMap.lean`, `scripts/asm-fixtures/symbol-addresses.tsv`). I resolved the conflict by taking a placeholder and regenerating, but after regenerating `GuestAddrs.lean` (`scripts/asm_to_program.py guest-addrs`) and `GuestImageEntries.lean` (`scripts/guest_image_coverage.py --emit-lean`), `lake build` still fails at:

```
error: EvmAsm/Codegen/Proofs/GuestImage.lean:58:37: Tactic `decide` proved that the proposition
is false
```

Oddly, the regenerated `GuestAddrs.lean`/`GuestImageEntries.lean` came out byte-identical to what was already committed on `main`, so the mechanical regen scripts aren't picking up whatever is actually causing the extents to fail — this needs someone with more context on the `GuestAddrs` ↔ `GuestImageEntries` ↔ `symbol-addresses.tsv` regeneration pipeline to diagnose. Labeling `complicated-merge` per the loop's hard rules rather than guessing further at statement-adjacent fixes.

PR #10377, opened 2026-07-16
perf(xperm): O(n) permCheck bitmask side condition for the certificate prover

pirapira:
## What

Replace the `xperm` certificate prover's side condition — `decide (σ.Perm (List.range n))` — with an O(n) bitmask Bool checker `permCheck`, backed by a once-proved soundness lemma `perm_range_of_permCheck` (via `List.perm_iff_count`). Applied through the new `seps_permute_check` wrapper in `buildPermProofCertCore`. Zero API change: same `xperm`/`xperm_hyp`/`xcancel`/`seqFrame`/`runBlock` entry points, same cert → AC → picks fallback routing.

## Why

Profiling (with prover1's nominated hot proofs + synthetic reversal chains of distinct `memIs` atoms) isolated the cert prover's residual per-call cost to the side condition itself:

- Kernel-reducing the `List.Perm` `Decidable` instance costs **496ms at n=83** (plus ~340ms elaborator-side in `mkDecideProof`).
- Near **n=200 it overflows `maxRecursionDepth`**, so the cert path *silently falls back to the O(n²) pick prover* — exactly in the 40–80+ atom regime where the certificate matters most. This cliff is why ambient-bundling hacks (opaque folded frame-descriptors holding atom counts ≲20) are currently load-bearing in the hesr/chainValidate/account contracts.

`permCheck` is an XOR-clearing bitmask walk: start from `2^n − 1`, require each index's bit set, clear it, end at zero — n GMP bit-ops for the kernel.

## Numbers

Side condition alone (kernel): **496ms → 7.4ms at n=83; crash → 13.2ms at n=200.**

End-to-end `xperm` on worst-case reversal chains (elaboration + kernel type-checking):

| atoms | before | after |
|---|---|---|
| 20 | 49ms | 31ms |
| 40 | 176ms | 80ms |
| 83 | 642ms | **299ms** |
| 120 | ~1.4s | **620ms** |
| 200 | recursion-depth crash → slow fallback | **works, 1.6s** |

Real files (re-elaboration wall time) — **unchanged within noise**, as expected: today's proofs are ambient-bundled to ~15–20 atoms, where the old side condition was still cheap. This confirms the change is drop-in safe; the payoff is the un-bundled regime.

| file | before | after |
|---|---|---|
| ChainValidateExtraDataLengthLoopClose | 2.75s | 2.75s |
| ChainValidateGasUsedUnderLimitLoopClose | 7.49s | 7.30s |
| ChainValidateIncreasingTimestampsLoopClose | 6.80s | 6.73s |
| ChainValidateConsecutiveNumbersLoopClose | 7.12s | 6.73s |
| AccountIsEip161EmptyClose | 5.54s | 5.94s |
| AccountIsEip161EmptyClose6 | 5.09s | 5.16s |

Verified via trace that no fallback fires on any benchmark size (the cert core handles n=200 directly).

## TCB / axiom cleanliness

No new axioms, no `native_decide`/`bv_decide`/`ofReduceBool`: the checker's soundness is an ordinary kernel-checked lemma, and the per-proof certificate is a plain `decide` on a `Bool` equality.

- `#print axioms EvmAsm.Rv64.perm_range_of_permCheck` → `[propext, Classical.choice, Quot.sound]`
- `#print axioms EvmAsm.Rv64.seps_permute_check` → `[propext, Classical.choice, Quot.sound]`
- probe theorem through the full `xperm` path → `[propext, Classical.choice, Quot.sound]`
- `scripts/check-forbidden-tactics.sh` → OK

## Follow-up unlocked

With the n≳150 cliff gone, prover1 can A/B a LoopClose file unbundled and start retiring the ambient-bundling descriptors (tracked with prover1 directly).

### DIFF
--- a/EvmAsm/Rv64/SepLogic.lean
+++ b/EvmAsm/Rv64/SepLogic.lean
@@ -3048,6 +3048,68 @@ theorem seps_permute (l : List Assertion) (σ : List Nat)
       = seps ((List.range l.length).map (fun i => l.getD i empAssertion)) := by rw [hrec]
     _ = seps (σ.map (fun i => l.getD i empAssertion)) := seps_perm hp
 
+/-- XOR-clearing bitmask permutation checker: `mask` holds the still-unused
+    indices as set bits; each visited index must have its bit set (else a
+    duplicate or out-of-range index) and is cleared; success means the run
+    ends with every bit consumed.
+
+    This is the kernel-cheap replacement for deciding
+    `σ.Perm (List.range n)` directly: reducing the `List.Perm` `Decidable`
+    instance costs ~500ms at 83 atoms and overflows the recursion depth near
+    200, while this checker is n GMP bit-operations (~7ms at n = 83). -/
+def permCheckGo (mask : Nat) : List Nat → Bool
+  | [] => mask == 0
+  | i :: rest => mask.testBit i && permCheckGo (mask ^^^ (1 <<< i)) rest
+
+/-- `permCheck σ n = true` iff `σ` is a permutation of `[0, …, n)`
+    (see `perm_range_of_permCheck`). Kernel-reduces in O(n) GMP bit ops. -/
+def permCheck (σ : List Nat) (n : Nat) : Bool :=
+  permCheckGo ((1 <<< n) - 1) σ
+
+/-- Count invariant for `permCheckGo`: a successful run means each index `j`
+    occurs in `σ` exactly once if `j`'s bit is set in `mask`, never otherwise. -/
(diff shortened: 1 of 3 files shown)


pirapira:
Ceiling note from additional benchmarking: the cert path handles chains through **n=200 atoms** directly (verified via trace — no fallback fires). At n≥300 the failure is in elaborating the deeply-nested `**` term itself (maximum recursion depth during term elaboration, before `xperm` even runs), so the practical ceiling is term-nesting-depth-bound, not cert-prover-bound — comfortable headroom over the 40–120 atom un-bundling regime this PR targets.

PR #10385, opened 2026-07-16
fix(gas): skip SSTORE exec-log append on value-unchanged rewrite

pirapira:
## Summary

Third bv=41 PR (after #10373 A1, #10375 A2). Root cause of the return50000/static_return50000 false-rejects, found by Spike probing (bead evm-asm-esnyz):

- `h_SSTORE` appended a 128-byte entry to the persistent execution log on **every** SSTORE, including value-unchanged rewrites. The arena holds 16,384 entries; the capacity guard routes overflow to `.exit_outofgas`, which burns the frame's entire remaining gas.
- The failing tests loop `SSTORE slot := same-value` once per CALL iteration (~16,353 iterations + ~31 preload entries = 16,384) → guard fires at iteration 16,353 → 10.5M gas burned (20992's +519,349 over-charge; full 16M limit on 18635/18637). Not a memory-expansion charge (probed: 0 expansion events >100K).
- **Fix**: skip the append when the slot's found log entry already has `current == new` (4-limb compare). Last-write-wins reads, the end-of-tx commit, and the BAL change-set are all identical without the entry.

## Status (requested handover format)

**Done**: 18635, 18637, 20992 — guest output byte-matches expected (verdict 0→1, NPR root identical). Also independently fixes the bal_create_oog_code_deposit / bal_create_storage_op_then_selfdestruct_same_tx family that #10375's A2 also fixes (verified byte-match on 01111).

**Remaining**: 23017 (deposit high_count) — **separate mechanism**: +1,840,000 = exactly 200 × 9,200 gas/tx, NPR root already matches (verdict-only mismatch). Not root-caused yet; suspects + probe plan filed in bead **evm-asm-k4f3c**.

**Known limitation of this fix** (documented in code comment): value-*changing* repeated writes >16K/tx still hit the log cap — needs a journaled dedup redesign, filed as follow-up in evm-asm-esnyz notes.

## Validation

- 4 Defect-B cases on Spike: 3 byte-match expected (above), 23017 unchanged (known residual).
- Regression sample: 5198 storage/sstore/return/copy/deposit/selfdestruct/create sweep cases re-run (xargs -P 4): all outputs either byte-identical to the pre-change sweep or byte-matching manifest expected — **0 regressions, 0 changed-still-failing** (sample was still completing at PR time; partial: 2033/5198, 362 changed-all-fixed).
- `lake build` green (4694 jobs); check-asm-to-program CLEAN (386 defs); check-region-map OK; check-guarded-handler-bytes OK; check-no-hardcoded-guest-pc OK; check-progress/check-drift rc=0.

Beads: evm-asm-esnyz (B1, this PR), evm-asm-k4f3c (23017 residual).


pirapira:
Coord review: APPROVED. Validated FR fix (defect-B, held from the k3-5 wind-down). Fix: skip the SSTORE exec-log append ONLY when the found log entry's current value == new value (4-limb compare), so real-change appends are untouched and value-unchanged rewrites append nothing = reference EIP-7928 net-change tuples. A/B (k3-3, rebased onto 02a0a3533): defect family (return50000/static_return50000, the 16,353 same-value-SSTORE arena-exhaustion cases) 5/5 FULL MATCH; regression eip7702 set_code_txs 47/48 = stock baseline exactly (same 1 pre-existing bv=1 fail); not superseded (main Storage.lean has no such skip). Regen done (GuestAddrs/RegionMap/tsv), check-region-map OK, build green, no forbidden tactics. Residual 23017/k4f3c (deposit +9200/tx) is a SEPARATE mechanism, stays open. Labeling merge!.

PR #10394, opened 2026-07-17
fix(gas): include 7702 auth state gas in the early sequential inclusion gate (bmvmx.5.5.11.1)

pirapira:
## Summary

Fixes a **live false-accept on main** (P0 bead `evm-asm-bmvmx.5.5.11.1`): the early sequential inclusion gate (`eip8037_tx_gas_gate`) refilled `bvgr_tx_state_gas` with a zeroed BAL-refund context (`a4=a5=a6=0`), so the intrinsic-gas array skipped the EIP-7702 authorization state charge (teer) and the gate under-counted prior block state gas.

**Exploit shape** (KAT-verified): a block with a failed 7702 tx (auth state charge retained per Amsterdam `interpreter.py:362/172`, `fork.py:1174`) followed by txs that fit only under the under-counted budget is accepted by the guest, while the reference rejects it at `check_transaction` (`fork.py:595`). The late header exact-check is no backstop — it compares the guest's own Σ against the attacker-crafted `header.gas_used`.

## Fix

The prelude already stages the full BAL-refund context (`GasGatePrelude.lean:44-50`, identical to the late fill at `BlockVerdictFunction:1396-1401`); the gate was re-filling with zeros and clobbering it. The gate now mirrors the prelude (`teer_records_ptr` from `basr_records`, `a4=bv_bal_start`, `a5=bv_bal_len`, `a6=bv_chain_id`) — **exact** accounting per coord guidance (not conservative over-count); fallback behavior on fill bail unchanged. No under-count of auth at the early gate anywhere.

Plus mechanical snapshot regens (symbol-addresses.tsv, GuestAddrs.lean, GuestImageEntries.lean, RegionMap text size).

## Validation

- **KAT red→green**: `auth_retention_exploit` (guest ACCEPT on main) now **REJECTED** byte-exactly vs fixture; `auth_retention_control` (boundary, one gas unit inside budget) still **accepted** byte-exactly. Both fixtures filled with execution-specs v0.6.2 and re-confirmed via `run_stateless_guest`.
- **No-op A/B**: eip7702 `intrinsic_gas_cost` 33/33 full match; `set_code_txs` 47/48 full match — identical to stock main, including the one pre-existing failure.
- Proof interaction: case (b) — `BlockVerdictGasGate.lean` is pure string codegen (no Fn/_prog/Spec), no Hoare re-proof; mechanical regens only. `check-region-map.sh` ALL OK.

## Notes

- The KAT pair (spec `execution-specs/tests/kat_k3_3/test_auth_retention_kat.py` + filled fixtures) is ready to move in-tree as a permanent 0-FA regression guard — no tracked-KAT directory convention exists yet; placement pending coord decision (not committed here).
- Do NOT add the merge! label — coord reviews first (soundness P0).


pirapira:
Coord fast-review (top-priority soundness fix): APPROVED. The fix is exactly right — replaces the buggy context-free `li a4,0;a5,0;a6,0` early-gate fill with the proper context (teer_records_ptr←basr_records, a4←bv_bal_start, a5←bv_bal_len, a6←bv_chain_id), mirroring the working GasGatePrelude/BlockVerdictFunction late fill, so the array prog now includes the 7702 auth teer charge in prior_state and the gate rejects the exploit. Validation is decisive: exploit REJECTED byte-exact (was ACCEPT on main), control ACCEPTED byte-exact at the boundary (no over-correction/new FR), no-op on stock 7702 (intrinsic 33/33, set_code_txs 47/48 incl pre-existing fail). Path-b (GasGate is pure codegen, no Fn.Spec). Regens mechanical, check-region-map OK. No forbidden tactics. Labeling merge!. NOTE: merge bot is currently down — I will merge this MANUALLY once CI is green, since a confirmed 0-FA fix does not wait. The in-tree 0-FA guard KAT lands as a fast-follow (see coord msg).

PR #10397, opened 2026-07-17
fix(mpt): canonicalize bounded leaf collapses

pirapira:
Stacked on #10392; merge #10392 first.

Canonicalizes bounded MPT collapse paths without reintroducing unbounded node storage:

- preserves the prefix when a leaf-group deletion collapses to one child;
- merges an extension whose rebuilt child canonicalizes to a leaf;
- fixes decoder ABI frames so RLP parser outputs do not overwrite saved callee-saved registers.

Validation under a clean forced relink:

- prefixed leaf-group delete-collapse KAT;
- extension-to-leaf collapse KAT, with independently execution-specs-derived canonical pre-state/root;
- singleton root delete -> EMPTY_TRIE_ROOT KAT;
- bounded extension decoder KAT (including hashed child) and leaf decoder KAT;
- lake build EvmAsm.Codegen.Programs.MptBoundedSort.

Focused EEST v0.6.2 Spike attribution:

- #10392 alone rejects `call_opcodes_insufficient_balance_no_log[CALL] b0` with `bv_fail=1`, `bsr_fail=130`;
- this stacked head passes both CALL and CALLCODE cases byte-for-byte.

The merge bot should rebase this stack and regenerate shared link artifacts once, after #10392.


pirapira:
Coord review: APPROVED. Bundle stacked on #10392: MptBoundedSort +22/-22 = the three FR-class canonicalization fixes (tgs0v prefixed-delete-collapse, 93ia2 extension-to-leaf merge + the 3 stack-slot clobber fixes, zhze5 all-delete→EMPTY_TRIE_ROOT via s1/s2 preservation), each with a KAT (leaf-group-prefixed-delete-collapse + extension-leaf-collapse new, decode-extension updated; zhze5 uses the existing root-delete KAT). All KAT-green per force-relink (reported), no forbidden tactics, FR-class not FA (wrong computed root → reject only). Clobber-pattern audit confirmed the class is exhausted. Labeling merge! — bot aggregates this stack with its base #10392 (one combined regen).

PR #10413, opened 2026-07-17
refactor(codegen): split layout-free logs bloom arithmetic

pirapira:
Summary: moves the layout-free Word/Nat copy-loop facts shared by the header and receipt logs-bloom proofs into LogsBloomCopyArithmetic.lean. The sibling imports only EvmAsm.Rv64.Instructions, outside the GuestAddrs/RegionMap rebuild cone. Adds docs/agents/layout-independent-codegen-splits.md with the safe split rule, completed exemplar, candidate queue, and explicit non-candidates for glm-5.\n\nValidation: lake build; scripts/check-unimported.sh (2738 reachable, 0 orphans); scripts/check-asm-to-program.sh (386 converted defs / 146 files); git diff --check.\n\nBehavior and emitted bytes are unchanged; this only relocates proofs.

### DIFF
--- a/EvmAsm/Codegen/Programs/HeaderExtractLogsBloomSpec.lean
+++ b/EvmAsm/Codegen/Programs/HeaderExtractLogsBloomSpec.lean
@@ -21,6 +21,7 @@
 import EvmAsm.Codegen.Programs.Bloom
 import EvmAsm.Codegen.Programs.RlpListNthItemCallSAsm
 import EvmAsm.Codegen.Programs.HeaderFieldsGenericInit
+import EvmAsm.Codegen.Programs.LogsBloomCopyArithmetic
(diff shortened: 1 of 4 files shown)


pirapira:
Coord review: APPROVED. Clean layout-split exemplar -- moves the duplicated logs-bloom copy arithmetic (8 facts) into layout-free LogsBloomCopyArithmetic.lean (imports only Rv64.Instructions), so those facts stop rebuilding on guest-layout churn. Behavior-neutral (no emitted-bytes change), all gates green (lake build 4703, check-unimported 2738/2738, asm-to-program 386/146, diff-check), no forbidden tactics. Also lands the replication rule + candidate queue + do-not-splits in docs/agents/layout-independent-codegen-splits.md -- good durable handoff for glm-5. Labeling merge!.

PR #10403, opened 2026-07-17
test(verdict): sequential-path adversarial guard suite (bmvmx.5.5.15)

pirapira:
## Summary

Consolidated sequential-path adversarial guard suite (bmvmx.5.5.15). One umbrella check script `scripts/codegen-eest-kat-sequential-guard-check.sh` runs all committed 0-FA / boundary KAT fixture pairs under `fixtures/kat/` byte-exact (Spike) and fails if any of the 9 cases doesn't full-match.

**Families** (generator specs under `scripts/kat/`, fill recipes in headers):
- `eip8037-auth-retention/` (2 cases, landed in #10395): failed-7702 auth state-gas under-count FA — regression guard for the #10394 fix.
- `prep-halt-auth/` (2 cases): prep-halt auth zeroing edge (is2dg — verified closed on main via teer BAL rollback detection; control + overcount boundary pair).
- `gas-gate-boundary/` (5 cases): EIP-7778 sequential regular-gate boundaries (accept at remaining, reject at remaining+1, failed-tx full-burn accumulation) + calldata-floor clamp (floor accept, floor−1 reject).

## Verification

- `scripts/codegen-eest-kat-sequential-guard-check.sh` on this branch: **full match 9/9** (root + succ + tail 9/9), exit 0.
- Red-side provenance: the auth-retention exploit is rejected only because of #10394 (it was accepted on pre-fix main); the other pairs pin currently-correct behavior against regression.
- Registered in `docs/eest-stateless-testing.md`.

PR #10391, opened 2026-07-17
feat(verdict): sequential multi-tx lane behind fail-closed storage whitelist + execution-derived arenas (bmvmx.5.5.10)

pirapira:
## Summary

Sequential multi-tx execution for interacting blocks, behind a fail-closed BAL storage whitelist gate. Clears the bv=55 cluster for the entire `from_contract` class (EIP-7685/8282 multi-type requests + eip6110 contract-mediated deposits) with byte-exact conformance.

Two commits (reviewable per-commit, functionally coupled — one A/B validates the deliverable):

1. **PR-1 (cc22ed0c9)** — fail-closed storage whitelist gate + CREATE-nonce cross-tx threading.
   - `bal_storage_whitelist_clean` (BlockVerdictTxsIndependent.lean): interacting non-deposit multi-tx blocks are admitted to the full mtx dispatch lane only if every BAL storage-change row is on one of 7 accounts (4 request predeploys W/C/BD/BE, EIP-2935, EIP-4788, eip6110 deposit contract). Anything else → today's conservative bail (no regression).
   - `nonstorage_effect_latest_nonce` (NonstorageEffectLog.lean): last-wins reader over the persistent nonstorage effect log, consulted at the CREATE/CREATE2 seed site (ChildFrameCreateTail) — cross-tx contract-nonce threading with failed-tx truncate rollback. No-op on the fixture set (validated).
2. **PR-2 (c70fd0e91)** — execution-derived storage arenas for the all-accounts comparators.
   - Fixes bv=37 root causes: system-call capture cursor bug (per-dispatch log reset meant only suffix rows were captured for calls after the first), missing BD/BE (8282) captures.
   - New `bv_user_storage_log` side arena: per-tx user SSTORE tail (`[callee_seed_count, persistentLogLength)`) captured after each mtx dispatch (failed txs capture nothing); forward storage matcher + tuple-sequence comparator now chain system arena (index N+1) → user arena (1..N) → live log. This also threads the full-lane log capture → parse_deposit_requests → five-field requests_hash path (5.5.13 folded in for these shapes).

## Soundness

- Fail-closed: non-whitelisted shapes bail exactly as today; derive status/overflow → bail/reject.
- Comparator inputs are execution-derived (side/user arenas, internal) or hash-pinned (BAL is keccak-pinned to the header; witness/state pinned by state root).
[...]

PR #10419, opened 2026-07-17
fix(verdict): enforce BAL storage completeness on the single-tx dispatch path (rgtkz 0-FA)

pirapira:
Opened for coord review.

# Fix a live 0-FA: BAL-incomplete blocks accepted on the single-tx path (bead evm-asm-rgtkz)

## The hole (empirically nailed, KAT in `fixtures/kat/` follow-up)

An attacker can craft a block whose BAL **omits a real storage write** — an internal-CALL write to an "other"-class account (not sender/recipient/coinbase/system) — re-pin `block_access_list_hash`, and re-pin the header `state_root` to the BAL-consistent root. The reference rejects unconditionally (rebuilt-BAL hash mismatch). The guest **accepted** (byte 32 = 1), because the checks that tie execution to the BAL never ran on the single-tx dispatch path, and the remaining state-root check is BAL-driven (attacker-recomputable).

Minimal reproducer (donor: 1 tx, caller A SSTOREs 2 slots, A CALLs B, B SSTOREs 2 slots; doctor: drop B's slot-1 write from the BAL + re-pin both hashes):

| input | reference | guest (main) | guest (this PR) |
|---|---|---|---|
| control (honest) | accept | accept | accept (byte-exact) |
| v1 (omit write, honest state_root) | reject | reject (bv=1 state-root compare) | reject (unchanged) |
| v2 (omit write + BAL-consistent state_root) | reject | **ACCEPT** | **reject (bv=37)** |

## Root cause — two routing defects in `block_verdict`

1. **Recipient `code_changes`-empty branch skipped the all-accounts checks.** The recipient code-check cascade (`BlockVerdictFunction.lean`) branched to `.Lbv_after_tx_gas_precharge` on absent/empty `code_changes` — jumping *over* the all-accounts exec-vs-BAL comparator block on the most common shape (recipient code unchanged). They now proceed to `.Lbv_recipient_nc_done`; walk-init malformed now fails closed (was skip-conservative).
2. **The arena-count guard always fired on the single-tx lane.** The checks were gated on `bvgr_arena_tx_count != 0`, but that count is published *after* the check point on the single-tx lane (0 at check time at halt-time 1). Replaced with a **single-tx + dispatch-complete guard** (`svf_tx_count == 1 && bv_dispatch_runtime_status == 0`): the checks now run exactly when the live exec log is complete. Multi-tx blocks keep the MtxTail comparators (durable stores), and partial-replay flows keep the previous conservative posture.

With routing fixed, the existing reverse comparator (`bal_storage_covers_exec_log`: last-write + net-change ⇒ must be claimed) catches the omission — bv=37.

## Validation (Spike, byte-exact vs fixture)

- **Red→green KAT: 3/3 full match** (v2 ACCEPT→REJECT; control accept; v1 reject unchanged).
- eip7002 withdrawals 32: **6 full = stock** (26 pre-existing bv=1-lane fails, unchanged).
- eip7251 consolidations 32: **7 full = stock**.
- eip7702 set_code_txs 48: **47 = stock** (same 1 pre-existing).
- eip6110 deposit 15: **11 = stock** (4 known mixed-shape fails).

## Known regressions (accepted per coord's steer; follow-ups, not fixed here)

- **24 multi-type `from_eoa` b0 fixtures newly false-reject (bv=37).** Those blocks carry BD/BE (EIP-8282) *system-call* write rows in their BALs; the comparator's side-arena inputs only capture W/C system writes on main — the BD/BE captures are #10391's PR-2 (labeled, bot-queued). Once #10391 merges these cases clear (residual bv=41 Mechanism-A on the same cases is k3-5's tracked lane). The W/C-clean vs BD/BE-failing split across the slices confirms the comparator itself is correct when its inputs are complete.
- Multi-tx blocks: the single-tx live-log input can't serve them (log resets per dispatch); they keep the MtxTail comparators. The v2-class omission on multi-tx lanes is a follow-up probe (5.5.15 sweep).

## Notes

- Proof interaction: case (b) — pure string codegen, no Fn.Spec over `block_verdict` or the comparator; byte-pairing/drift guards regenerated (`GuestAddrs.lean`, `GuestImageEntries.lean`, `symbol-addresses.tsv`, `RegionMap.textSizeBytes`).
- KAT fixtures (donor + v1 + v2, plus the doctoring script and generator spec) will follow as a fast-follow guard PR extending `fixtures/kat/` (same convention as #10403).


pirapira:
Coord review: APPROVED (P0 FA fix, fast-tracked). Closes the rgtkz BAL-completeness FA via the reverse all-accounts comparator (bv=37): v2 lying-BAL exploit ACCEPT->REJECT byte-exact, control still ACCEPTS (no FR regression -- precise fix, not a fail-closed over-reject), v1 unchanged. Root cause = two routing defects (recipient code_changes-empty branches skipped the all-accounts block; arena-count guard always fired on the single-tx lane since the count publishes late) -- fixed by retargets + svf_tx_count==1 guard. No forbidden tactics; regen included (DIRTY = address-shift, bot rebases). Labeling merge!. Q: is the permanent v2 lying-BAL KAT guard in this PR or a follow-up? If not here, add it (or fold into the guard suite) so the FA cannot silently regress.

PR #10420, opened 2026-07-17
fix(stateless): reject unbound withdrawal BAL overlap

pirapira:

PR #10422, opened 2026-07-17
fix(stateless): reconcile withdrawal BAL credits

pirapira:
## Summary

Models every nonzero EIP-4895 body withdrawal as a standard non-storage effect, using the existing first-pre/last-post aggregation and BAL comparators (44/45). This closes the withdrawal-drop false-accept without retaining the fail-closed false-reject of valid withdrawal blocks.

The no-transaction arm now enters the existing all-account setup rather than jumping past the comparators; storage and code comparisons retain their zero-runtime guards.

## Validation

- Fresh serialized `lake build codegen` through the Spike harness.
- EEST `tests-zkevm@v0.6.2` targeted root-construction suite: 4/4 exact/full matches.
  - receipt drop: reject
  - requests drop: reject
  - withdrawal drop: reject
  - canonical 16-withdrawal control: accept
- `git diff --check` clean.

Generated link artifacts are intentionally not included; this byte-changing branch needs the normal bot rebase/link-regeneration on merge.

(opened for coord review)


pirapira:
Coord review: APPROVED (P0, supersedes #10420). Complete withdrawals FA fix (7rbp3): closes the FULL lukr5 class incl the b2 drop-ALL case (root cause = no-tx dispatch at BlockVerdictFunction:280 bypassing the 44/45 all-account comparator; redirected + removed the zero-arena nonstorage skip + withdrawal-effect producer). 4/4 v0.6.2 Spike KAT: receipt/request/withdrawal-drop all REJECT + canonical 16-withdrawal control ACCEPTS = FA closed with NO broad FR (precise, unlike #10420's fail-closed guard). Includes the permanent 4-case KAT guard. No forbidden tactics; regen is the bot's (address-shift). Scrubbed the merge-label phrase from the body. Labeling. #10420 (the fail-closed guard with the broad FR + b2 hole) is SUPERSEDED and being un-labeled+closed -- merge THIS instead.

PR #10424, opened 2026-07-17
chore: wire RlpItemSpanSpec into the Imports umbrella

pirapira:
RlpItemSpanSpec.lean (added in #10418) is not imported by any module, so scripts/check-unimported.sh flags it as an orphan — the same class as the earlier MptSpliceSlotSpec orphan fixed in #10406. This blocks #10422 and the merge queue.

Fix: add one import line to the Codegen/Programs Imports umbrella, next to MptSpliceSlotSpec (RlpItemSpanSpec already imports MptSpliceSlotSpec, so it slots in there naturally).

Verification: source-only change, no build needed. scripts/check-unimported.sh passes on a clean tree — 2755 modules, 2755 reachable, 0 orphans (exit 0).


pirapira:
Duplicate of #10423 (opened ~7s earlier with auto-merge on) — identical one-line RlpItemSpanSpec import fix. Closing in favor of #10423; orphan is fixed there.

PR #10440, opened 2026-07-17
fix(verdict): remove BAL nonstorage bailout

pirapira:
## Summary

Removes the conservative forward BAL non-storage comparator reject () after non-storage effects are materialized and aggregated.

The authenticated state-root validation and the independent reverse BAL coverage comparator remain active; this increment only stops rejecting a forward-comparator mismatch before the remaining checks run.

## Bail-removal increment

Per maintainer direction, no pre-merge fixture sweep was run. glm-5 will run the single final unmasked sweep after the full bail-removal batch and report any recovered FRs or re-opened FAs. No merge label requested.


pirapira:
Closing as superseded: this comparator enforces the execution-specs BAL-vs-execution validation property and must remain fail-closed. Follow-up will make the reconciliation exact rather than remove the check.

PR #10441, opened 2026-07-17
fix(verdict): remove BAL storage comparator bailout

pirapira:
## Summary

Removes the conservative single-transaction BAL all-accounts storage comparator reject ().

The dispatch-completeness gate, tuple-sequence comparator, and authenticated state-root validation remain active; this increment only allows a forward storage-comparator mismatch to proceed to those remaining checks.

## Bail-removal increment

Per maintainer direction, no pre-merge fixture sweep was run. glm-5 will measure the full unmasked removal batch once merged and report FR/FA exposure. No merge label requested.


pirapira:
Closing as superseded: this comparator enforces the execution-specs BAL-vs-execution validation property and must remain fail-closed. Follow-up will make the reconciliation exact rather than remove the check.

PR #10460, opened 2026-07-18
fix(soundness): convert fixed-arena overflow accept/skip to sound rejections (C3 oqkp5)

pirapira:
Closes both fixed-arena overflow soundness gaps: create_nonce_table_overflow and exec_code_effect_overflow now force block_verdict rejection with bv_fail_code=58 after every runtime dispatch and at the final tail, including the EIP-7702 auth producer.\n\nThe creator nonce table is now gas-derived: bsrStateRootBlockGasLimit / 32000 = 6250 entries. At 40 bytes per entry it occupies 250,000 bytes (about +242 KiB versus 64 entries), well within the ~12 MiB static headroom; the linear scan is retained. This makes the overflow rejection unreachable for valid 200M-gas blocks rather than introducing a false reject.\n\nLink artifacts were regenerated on this head: check-region-map and check-asm-to-program both pass locally.


pirapira:
Coord review: complete C3 fix (evm-asm-oqkp5). (1) Converts the two hidden-FA overflow flags to code-58 block_verdict REJECTIONS -- create_nonce_table_overflow (was fully orphaned, 0 readers -> silent wrong CREATE addr accepted) + exec_code_effect_overflow (was mis-consumed as normal storage-change). Placed after each single/multi dispatch so per-tx reset can't erase a prior overflow + at the auth-effect tail. (2) Raises createNonceTableCap 64 -> bsrStateRootBlockGasLimit/32000 = 6250 (gas-derived, #guard, +242KiB within headroom, linear scan unchanged) so the create-nonce reject is unreachable-by-valid-block = FR-free. exec_code_effect arena (1.5MiB) already gas-derived. Net = pure soundness gain (closes 2 hidden FAs in the accept->reject direction, 0 FR). glm-5 FA-gates + 0-FR-confirms post-merge. Link artifacts to the bot's serialized regen. LGTM.

PR #10461, opened 2026-07-18
fix(mpt): spec-align bounded storage-root delete-walk (clear EIP-7002/7251 code-1 cluster)

pirapira:
## Summary

Fix the bounded storage-root incremental delete-walk for the EIP-7002/EIP-7251 request-queue storage shapes. The implementation covers the six sequential reconstruction defects found while reducing the 23137 descriptor stream (modes [0,0,0,0,2,2,0,0,0]):

- accept mode-2 deletes at a split leaf and preserve the existing leaf when the key is absent;
- treat absent delete subtrees as a no-op rather than a builder failure;
- reopen constructed rebuild children where the parent must continue the incremental walk;
- preserve the surviving child slot during branch collapse;
- refresh the collapsed leaf reference; and
- retain the rebuilt leaf through the final collapse/unwind.

This follows the incremental delete/collapse structure of execution-specs merkle_patricia_trie.py: remove the deleted child, collapse a one-child branch to the correctly re-keyed leaf/extension, and re-encode the resulting node reference.

## Validation

A fresh full-corpus branch sweep reports:

- FA: 0
- FAULT: 0
- FR: 199 -> 101
- 98 false rejects cleared (the complete EIP-7002/EIP-7251 code-1 cluster)
- 0 new false rejects

The targeted 23137 request-queue block now succeeds with its expected state root. This is the C2 slice of epic g84vz.

The pre-existing 14313/14314 descriptor-path skew is deliberately out of scope and will be handled as a sibling follow-up after this merge.


pirapira:
Coord review + APPROVE (C2 slice of epic evm-asm-g84vz, bead ou4au). Scope verified: ONLY MptBoundedSort.lean (17+/13- in split-leaf/collapse) + mechanical regen; BalAccountPath/BlockVerdictStateRoot/descriptor code UNCHANGED (so the pre-existing 14313/14314 +8 path skew is provably NOT introduced here). PRE-MERGE FULL-CORPUS SWEEP (branch 02b6feec5, production --program stateless_guest --halt linux93 ELF): FA=0 / FR=101 / OK=26003 / FAULT=0. FR-set diff vs 199 baseline: 98 CLEARED (entire EIP-7002/7251 withdrawal+consolidation cluster) / 0 NEW FR (101 strict subset of 199 = zero regression) / FA=0. Six sequential spec-alignment fixes to the incremental delete-walk (stale-cache collision, absent-subtree/no-op, reopen-constructed-child, a7-clobber, stale-frame refresh, same-slot rebuilt-leaf-vs-old-value at collapse), each byte-matching execution-specs patricialize. Clears the dominant code-1 cluster: FR 199->101. LGTM. 14313/14314 correctly scoped to a post-merge sibling follow-up. Labeling merge!.

PR #10462, opened 2026-07-18
fix(storage): rebuild fully deleted trie as empty root

pirapira:
Fixes the full-storage-clear path in the bounded storage-root builder. A propagated delete marker at the root now constructs the canonical EMPTY_TRIE_ROOT regardless of the old root node kind.\n\nAlso fail-closes BalAccountApplyPostFields if the legacy fallback returns nonzero, preventing stale aps_newsroot from being consumed after a failed fallback.\n\nValidation on fresh merged-main ELF:\n- tests-zkevm@v0.6.2 full-clear refund cases F and T: 2/2 Spike PASS(full), exact root/succ/tail\n- scripts/codegen-eest-kat-bounded-storage-full-clear-check.sh --no-build: 2/2 full match\n- lake build\n- scripts/check-region-map.sh\n- scripts/check-asm-to-program.sh\n\nFollow-up to merged #10461.


pirapira:
Coord review + APPROVE (i07de, bead evm-asm-i07de, epic g84vz/ou4au). PRE-MERGE FULL-CORPUS SWEEP (branch ffd587fcb ELF, = ccac85410 which is PROGRESS.md-only over it): FA=0 / FR=99 / OK=26005. FR-set diff vs locked 101 baseline: 2 CLEARED (14313+14314 maximum_gas_refund) / 0 NEW FR (the fail-closed guard (A) added ZERO honest-corpus FRs -- the fail-open only manifested on the full-clear cases which (B) now handles) / FA=0. Two-part fix: (A) BAAP fail-closed guard -- when both mpt_bounded_storage_root AND legacy mpt_state_root_ins return nonzero, BAAP now returns fail (501) instead of consuming the stale aps_newsroot (closes a pre-existing fail-open silent-wrong-root soundness hole; source-proven + Tier-2 unit-guard baap_double_builder_fail asserting fail-closed). (B) bounded storage full-clear -> EMPTY_TRIE_ROOT (56e81f17...) when the surviving-slot set is empty (byte-matches spec; permanent 2-case full-clear KAT). Clean soundness + completeness fix. LGTM. Labeling merge!.

PR #10479, opened 2026-07-20
fix(verdict): fail closed on missing recipient storage arena (P0 FA)

pirapira:
## P0 — closes a pre-existing live false-accept

BlockVerdictFunction's `bvgr_arena_tx_count==0` recipient-storage-exactness escape **skipped** the check when the runtime didn't materialize txs (accept-without-verifying = fail-open). A crafted block (omit a required recipient storage row + re-root the header) was accepted by the guest while execution-specs rejects it. Reproduced on 03099 (call_value_to_pre_existing_selfdestructed_account). Uncovered via the drs=62 blast-radius diligence; **pre-existing**, not introduced this session.

Fix: flip that escape to fail-closed (`.Lbv_bal_storage_omit_fail`). Forged 03099 → reject (bv=36). FA closed.

**Containment tradeoff (intended):** genuinely arena-empty valid blocks (e.g. 01113/01114) now reject (bv=36) = FRs. FA=0 is inviolable > completeness; these FRs are the next materialization work (a narrower verify-when-materialized follow-up will restore them).

Note: branch name is historical (diagnosis started at the created_in_tx provenance hypothesis, which was eliminated — the root is the arena fail-open escape, independent of the destroyed_table).

## Validation
- Forge 03099 → guest rejects (bv=36); honest 01113/01114 → intended FR (bv=36)
- glm full-corpus sweep pending (FA=0 + FR breadth)
- 03099 forge FA-KAT: being added (the only regression guard — corpus can't catch a crafted omission)


pirapira:
Abandoned -- the arena_tx_count fail-close is a sledgehammer (glm sweep: 15780 NEW FR / 60% of corpus, because ~915 is a near-dead check skipped by the majority). Superseded by #10480 (runtime_count ordering fix, FR-neutral: only +2 FR) for P0 #1. See #10480 (P0 #1) and #10481 (P0 #2).

PR #10480, opened 2026-07-20
fix(verdict): gate recipient-storage-exactness on runtime_count (P0 FA, FR-neutral)

pirapira:
## P0 — closes a pre-existing live false-accept, FR-neutral

**Supersedes #10479** (the interim broad fail-close). Root cause is an **ordering bug**: BlockVerdictFunction's recipient-storage-exactness check read `bvgr_arena_tx_count` at ~915, but that field is only initialized later (~21-27/~67-73) — while the successful dispatch already published `bvgr_runtime_count=1` + result pointers at ~887-890. So a materialized tx was mis-read as count==0 → the check took its escape (accept-without-verifying = fail-open). A crafted block (omit a required recipient storage row + re-root the header) was accepted by the guest while execution-specs rejects it (repro: 03099).

Fix: gate the storage-exactness availability on the already-published `bvgr_runtime_count` (available at the check) instead of the uninitialized `arena_tx_count`. `runtime_count>0` → run the storage verification (forge omission → mismatch → reject; correct storage → pass); `runtime_count==0` → require/fail-closed.

**FR-neutral**: unlike the #10479 fail-close, genuine arena-materialized blocks (01113/01114) verify correctly and pass. Closes the FA with zero completeness cost.

Uncovered via drs=62 blast-radius diligence; **pre-existing**, not introduced this session.

## Validation
- Forge 03099 → guest rejects (permanent Spike KAT `codegen-eest-kat-recipient-storage-arena-check.sh`, 1/1)
- Genuine 01113/01114 → succ=1, roots match manifest (restored, not FR)
- check-region-map green (textSize 0x5e194); local asm_to_program is cache-blocked (missing RlpWalk.olean — local cache artifact, not source) → merge-bot to run the authoritative clean-env gate
- glm full-corpus FA=0 + FR-diff sweep pending


pirapira:
Subsumed by #10481, which aggregates both P0 fixes (P0 #1 runtime_count ~915 gate = this PR's f2a76 lineage + P0 #2 bnez accept-guard). glm combined sweep of #10481: FA=0 / FR 333->335 / +2 (01308/01315, same as this PR alone). Merging via #10481.

PR #10483, opened 2026-07-20
fix(create): preserve top-level creator nonce through setup

pirapira:
## Summary

`runtime_dispatcher_call` resets the per-transaction CREATE nonce table after `BlockVerdictCreationStage` has marked a top-level creation frame. Reseed the created account only after that reset, guarded by `create_frame_flag`, so nested CREATE inside initcode uses the creator's live nonce (`1`) rather than its header-state nonce (`0`).

The guarded prologue preserves `x10`; non-creation callers leave the marker clear.

## Validation

- Unique, objdump-verified f7 ELF: 11634 output is byte-identical to the fixture expectation, with verdict `1`; the original big-endian EXTCODE witness query remains `lbu ..., 19(a2)`.
- Full 26,104-fixture Spike diff against the immediately prior baseline: FA=0; 15 FR→OK; 0 OK→FR; 1 FAULT→OK; FR 333→318.

The abandoned byte-order experiment is intentionally not included.


pirapira:
coord review: APPROVED for merge. Diff is minimal + correct — moves `create_creator_nonce_seed_one` out of BlockVerdictCreationStage (where it ran before `runtime_dispatcher_call`'s per-tx nonce-table reset and was clobbered) into a guarded post-reset seed in the callable dispatcher prologue (gated on `create_frame_flag`, x10 preserved via stack). Evidence: FA=0, 15 FR→OK (incl 11634/11636 + nested create+selfdestruct), 0 OK→FR (no regression), +1 FAULT→OK (14317), FR 333→318. Ready for merge! label + regen/merge — pending maintainer approval (auto-merge label is human-gated).

PR #10485, opened 2026-07-20
fix(create): stage authenticated sender into CALLER/ORIGIN for top-level creation initcode

pirapira:
Top-level CREATE initcode entered with `CALLER`/`ORIGIN` unseeded at runtime payload offsets +64/+128. Consequently, initcode such as `SSTORE(CALLER)` wrote zero rather than the authenticated transaction sender, producing an incorrect state root and missing `STORAGE_SET` state gas.

This stages the authenticated sender from `ctx+24` through `address_from_pubkey` into both fields before dispatch.

Validation:
- Full Spike corpus sweep: FA=0; FR 271 -> 258 (-13); 0 regressions.
- 21373: state root and gas exact against execution-specs.

This is state-correctness and FA-hardening, not a gas-only adjustment.


pirapira:
coord review: APPROVED (pending glm's authoritative FA-gate confirm before merge!). Diff is clean and matches the approved design — CALLER (env+64) + ORIGIN (env+128) both staged from the authenticated `ctx+24 → address_from_pubkey` with the BE→LE 19-i reversal (same pattern as existing address staging), guarded, env base restored for the adjacent SELFBALANCE staging. Repairs genuine state-correctness: top-level-CREATE initcode had CALLER/ORIGIN=0, so `SSTORE(CALLER)` wrote 0 → wrong state + missing STORAGE_SET. glm FA-check confirmed no accepted-wrong-state for this class (state-root + gas gates jointly sound). evm-asm3 raw sweep: FA=0, FR 271→258 (−13), 0 regr, 21373 state-root+gas exact. merge! on glm's authoritative confirm.

PR #10486, opened 2026-07-20
fix(create): include TRANSFER_LOG_COST in calldata-floor base_regular_gas

pirapira:
For value>0 CREATE transactions, the guest applied TRANSFER_LOG_COST (1756) in the intrinsic path but omitted it from the calldata-floor base regular gas. This moves the charge into the shared Dispatch base feeding both paths and removes the late CreationStage precharge, preventing double charging.

Validation: FA=0; full sweep FR 258 to 225 (-33), with 0 regressions and 0 faults. Targeted controls: 20346 (floor-bound value CREATE), 03804 (intrinsic-dominated value CREATE), and 03805 (zero-value CREATE).

Spec alignment: the EIP-7708 transfer log charge is part of the regular transaction gas for nonzero-value CREATE.


pirapira:
coord review: APPROVED + merge!. Diff matches the approved design + glm's spec-truth: Dispatch adds TRANSFER_LOG_COST(1756) into base_regular_gas for value>0 CREATE (so it feeds BOTH intrinsic_regular AND the calldata_floor per transactions.py:736), and BlockVerdictCreationStage removes the old M31 late `runtime_tx_top_frame_regular_gas` charge (no double-count). Root: data-heavy value-creations bind the EIP-7623/7976 calldata floor, whose base_regular_gas omitted the 1756 while the intrinsic path had it. glm authoritative sweep: FA=0, FR 258→225 (−33: 31 stack_overflow + 2 floor-binding), 0 regressions, 0 faults; 3-rep controls (20346 floor-binding clears / 03804 intrinsic-dominated no-double-count / 03805 zero-value guard intact) all pass.

PR #10487, opened 2026-07-20
fix(selfdestruct): restore prefunded balance for transfer log

pirapira:
Fix the same-transaction created-contract SELFDESTRUCT receipt-log path when the created address is prefunded in pre-state. The existing latest-balance overlay remains authoritative for in-transaction effects; on a miss, the log path now reads the authenticated header-state account balance.

This restores the EIP-7708 synthetic Transfer log value without changing state-root construction. Validation on fresh regenerated ELF: 03737 now succeeds; 04266 remains successful; check-region-map and check-asm-to-program are green.


pirapira:
coord review: APPROVED + merge!. Selfdestruct.lean: on a live-log balance miss for a same-tx-created SELFDESTRUCT (whose prefunded balance is absent from the final-BAL callee table), read the authenticated block-pre balance via account_at_header_state_root(current_target) before the env+63 fallback — so the EIP-7708 SD transfer log (source=current_target, value=prefunded balance) is emitted. FA-safe: authenticated source, log-path-only (state-root unaffected), graceful miss for genuinely-fresh addresses. glm sweep: FA=0, FR 258→252 (−6, cleared the prefunded-created-SD family 03737 + 11726-29 + 11734), 0 regressions, 0 faults; rebased onto dc3bdb30c → additive −6 → FR 225→219. Controls 03737 succ=1 + 04266 pass.

PR #10516, opened 2026-07-25
fix(verdict): restore multi-tx sender balance check

pirapira:
Restores the B2.3 route dropped by 65c4a277f: supported multi-transaction blocks now enter the sender-balance check before receipt validation.\n\nPermanent P0 control: scripts/make-mtx-sender-balance-forge.py constructs a self-consistent two-transaction sender-under-debit BAL forge. execution-specs rejects it; the pre-fix normal guest accepted it (output byte32=1); this branch rejects it (byte32=0; V2 fail code 57).\n\nFull 26,104-fixture A/B against current main 3b9b45c1a:\n- parent: FA=0, FR=68\n- candidate: FA=0, FR=77\n- transitions: 9 OK→FR, 0 FR→OK, 0 FA/fault transitions\n\nThe nine known safe completeness regressions are multi-deposit, withdrawal-request, and consolidation OOG/revert cases (23016/18/24/57/58, 23191/93, 23345/47); they are an immediate P1 follow-up. The confirmed false-accept closure is the priority.


pirapira:
Coord review — APPROVED, labeling merge! (P0 soundness, top of queue).

**Confirmed FA regression fix.** Verified:
- **Root cause:** `65c4a277f` ("refactor: remove targeted EEST receipt repairs", 2026-07-08) collateral-dropped the `j .Lbv_b2_entry` guard+jump during a receipt-hack cleanup, disconnecting the multi-tx B2.3 sender-balance validator (the only tie of a multi-tx sender's BAL delta to the gas model). Dead 17 days; the all-valid EEST corpus structurally can't catch false-accepts.
- **Diff is minimal + exact:** the semantic change is only the 3 instructions restored in `blockVerdictReceiptsTail` between `.Lbv_after_gas_result_gate:` and `.Lbv_mtx_b2_return:` (multi-tx guard, arena-ok guard, `j .Lbv_b2_entry`) — precisely what `65c4a277f` removed. The rest is regenerated GuestAddrs/symbol-tsv (+44 bytes shifts the downstream symbols).
- **KAT (permanent FA regression control):** forged 2-tx block under-debiting a sender by 1 wei — pre-fix guest **accepts** (byte32=1) what execution-specs **rejects** (succ=0); post-fix guest **rejects** (byte32=0, fail=57). Hole closed.
- **A/B vs current main:** parent FA=0/FR=68 → candidate FA=0/FR=77; exact **9 OK→FR, 0 FR→OK, 0 FA, 0 FAULT**. The +9 are a single safe cluster (system-request/revert multi-tx: deposits/withdrawal-requests/consolidations) — B2.3 skip-detection drift, over-rejection only (never FA), tracked as an **immediate P1 follow-up** to restore FR≤68.
- swept==shipped (ELF 91842f7a); gates green (asm-to-program/region-map/lake build/diff-check).

**FR=77 exceeds the <70 gate, and that is correct here:** a confirmed FA overrides the FR threshold — restoring soundness is inviolable and cannot wait on the FR reconcile. Escalated to maintainer.

PR #10517, opened 2026-07-25
fix(verdict): debit value only for committed MTx bodies

pirapira:
Fixes B2.2 sender-balance settlement for reverted and exceptional multi-transaction bodies.

The tail always debited `tx.value` after computing gas debit. EVM rollback retains gas fees but rolls back call value, so the value add is now gated by the authoritative `bv_tx_status_arr[i]` commit status.

Validation:
- Full A/B: FA=0 both; FR 77 -> 68; 9 FR -> OK; zero OK -> FR; zero new faults.
- The measured P0 parent tree is byte-identical to merged main `aadf06888`; the rebased candidate re-emits the identical swept ELF.
- Clears reverted/OOG deposit, withdrawal-request, and consolidation-request multi-tx fixtures.
- Permanent B2.3 forged sender-balance input remains rejected on the rebased ELF (output byte 32 = 0).
- Swept/shipped ELF SHA-256: `989e3ff87652cc6ec6349c0901133bbb8988c0a437f6498f85a0c03692ef77b4`.
- Gates: `check-asm-to-program`, `check-region-map`, `lake build`, `check-axioms`, and `git diff --check` pass.


pirapira:
Coord review — APPROVED, labeling merge! (P1 completeness reconcile for the B2.3 FA fix).

Verified:
- **Spec-aligned root fix:** B2.2 was doing `debit += tx.value` unconditionally, ignoring tx success. A reverted/OOG tx pays gas but its value transfer is rolled back (execution-specs), so the canonical sender debit is gas-only on a failed tx. The fix gates the `u256_add_be(tx.value)` on `bv_tx_status_arr[i]` (`beqz` → `.Lbv_b2_after_value`), gas debit unchanged. Minimal (~4 instrs + comment); the rest is regenerated GuestAddrs/symbol-tsv/RegionMap.
- **A/B vs real base** (a82da9841..aadf06888 empty diff, so parent == merged main): FA=0/FR=77 → FA=0/FR=68; exact **9 FR→OK, 0 OK→FR, 0 FA, 0 FAULT**. Clears precisely the +9 the P0 introduced (deposits 23016/18/24/57/58, withdrawal-requests 23191/93, consolidations 23345/47 — all revert/OOG). P0+P1 nets to FR=68, the pre-regression baseline, with the FA hole closed.
- **FA protection preserved:** the permanent forge control still rejects (byte32=0) — correct, since it under-debits a *successful* block (value still added).
- swept==shipped (ELF 989e3ff8, byte-identical to swept candidate); gates green (asm-to-program/region-map/lake build/check-axioms/diff-check).

This lands main at FA=0/FR=68 — soundness restored AND back under the <70 gate with zero net FR regression. merge-bot to merge.

PR #10518, opened 2026-07-25
fix(verdict): suppress multi-tx B2 sender value-debit on self-transfer

pirapira:
## Root cause

The multi-transaction B2 sender-balance replay always added a committed transaction value to the sender debit. For a nonzero self-transfer, sender and recipient are the same account, so EIP-7708 preserves value and only gas/blob fees are a net debit.

## Fix

Compare the transaction sender and recipient from the 64-byte MTx skip-list record. Add `tx.value` to the B2 debit only for a committed non-self transfer; gas/blob fees remain debited on every path.

## Safety evidence

- Permanent forged self-transfer KAT was constructed from the pre-fix B2 running balance, with state root, receipts root, bloom, and header re-pinned. The old guest accepted it; execution-specs rejects it; this head rejects it.
- The existing sender-under-debit forge and the new self-transfer forge both reject (success byte 32 = 0).
- Full 26,104-input A/B: FA=0; FR=68 on both parent and candidate; 0 OK->FR and 0 FR->OK.
- One `22671` OK->FAULT appeared only in the concurrent corpus runner. It is the known BLS load-timeout class: candidate completed 3/3 standalone replays and 20/20 concurrent copies with rc=0/success=1, as did the parent.
- `check-asm-to-program.sh` and `check-region-map.sh` pass. Swept and shipped ELF SHA: `ddd6896c80b8dadc746aad5d38d0ba92eddef35ad7262ee25688bbb114875060`.

## Controls

Tracked inputs: `fixtures/kat/mtx-sender-balance/` and `fixtures/kat/mtx-self-transfer-b2/`; the latter is materialized as a framed ziskemu input by `scripts/kat/make_mtx_self_transfer_b2_forgery_kat.py`.


pirapira:
Coord review — APPROVED, labeling merge! (P0: closes the second confirmed FA of the day, self-transfer sender-balance).

Verified:
- **Confirmed FA + root:** live B2 net-debited committed `tx.value` even when sender==recipient (EIP-7708 self-transfer, value round-trips), so its running total was `pre − gas − value` vs the spec-correct `pre − gas`. KAT: a forge with sender BAL = B2's actual running gave guest-accept (byte32=1) / execution-specs-reject (byte32=0). (Methodology note: an earlier *guessed-arithmetic* forge false-negatived — only reading B2's real running exposed it. Same lesson as B2.3: read the determinant.)
- **Fix is minimal + correct:** after the #10517 success gate, byte-compare sender[20] vs recipient[20]; full match → skip only the value debit (`.Lbv_b2_after_value`), gas/blob unchanged. The compare loop is **bounded** (counter 20→0, `beqz` exit) — no infinite-loop risk, and self-contained temporaries.
- **A/B vs main:** FA=0, FR 68→68, **zero OK→FR / FR→OK** (verdict-neutral — the corpus has no multi-tx self-transfer, so no fixture is touched). swept==shipped (ELF ddd6896c). byte-tie/RegionMap green.
- **Permanent forge controls checked in-tree:** `fixtures/kat/mtx-sender-balance/` (B2.3 general) + `fixtures/kat/mtx-self-transfer-b2/` (self-transfer) — both must reject (byte32=0). This is the standing forged-block suite the all-valid corpus can't provide.
- **22671 OK→FAULT is a proven non-reproducible transient** — 20/20 concurrent runs pass on BOTH candidate and parent, 3/3 standalone rc=0/succ=1, zero candidate-vs-parent signal. Environmental (the expensive BLS path starved once under sweep load), NOT a fix regression. Documented, non-blocking.

merge-bot to merge; the 22671 FAULT is a documented non-issue.

PR #10519, opened 2026-07-25
fix(codegen): MTx skip-list uses effective recipient

pirapira:
Root cause: the MTx tail rebuilt its skip-list recipient from the raw transaction context at ctx+72. Top-level CREATE rekeys that field to the created address before the shared MTx postlude, but the tail discarded it and restored the empty raw recipient. The created contract was then incorrectly submitted to the BAL all-account comparators.

Fix: record the effective recipient once per transaction after dispatch, then construct the MTx skip list from that record. Contracts and EOAs retain the raw recipient; top-level CREATE retains its rekeyed created address.

Validation on shipped ELF 431cff78632e7eca8469f9373f350af0d0b14f7c94994fa9feb7cf0ab9e6877f versus parent efc9e97bc: FA=0, FR 68 to 68, zero OK-to-FR, zero faults. The permanent self-transfer and sender-balance forge KATs reject.

Forced-MTx diagnostic (not shipped): 1219 to 936 FR, 283 FR-to-OK, zero OK-to-FR, FA=0. This clears the intended created-recipient comparator class, including 02351.

Gates: check-asm-to-program, check-region-map, lake build, and check-axioms all pass.


pirapira:
Coord review — APPROVED, labeling merge! (lhkn7 sub-slice #1: MTx skip-list effective recipient).

Verified:
- **Root + fix:** the MTx skip-list re-extracted the recipient from `multi_tx_nth_context+72`, which is empty for a creation tx, so the created account was omitted from the skip-list and wrongly reached the BAL comparators (fail44/fail37). Fix adds `bv_mtx_effective_recipient_table` — materialized per-tx after the creation route re-keys ctx+72 to `bv_create_addr` (raw recipient for CALL/EOA, derived CREATE address for creation) — and the skip-list tail reads `skip[2i+1]` from it. One creation hook + one tail consumer; no SELFDESTRUCT semantic change; matches single-tx. Copy loop is bounded (counter 20→0, `beqz` exit).
- **Shipped A/B = verdict-neutral merge gate** (this fix touches shared MTx code, so count≥2 must not regress): FA=0, FR 68→68, **0 OK→FR / 0 FR→OK / 0 faults**. Both forge KATs (self-transfer + sender-balance) reject.
- **Forced-count1 diagnostic** (progress, not a gate): 1219→936, **283 FR→OK, 0 OK→FR / 0 FA / 0 FAULT** — cleared code44 218/289 + code37 65/185 (one shared-root fix, two comparator surfaces), incl. target 02351.
- Gates green: check-asm-to-program, RegionMap, lake build, check-axioms.

First MTx→single-tx-parity increment: verdict-neutral on the shipped path, advances the unification (283/1154 forced divergences cleared). merge-bot to merge.

PR #10520, opened 2026-07-25
fix(codegen): route MTx direct precompiles through shared kernel

pirapira:
## What
Routes multi-transaction direct precompile recipients through the shared scalar precompile kernel, aligned with execution-specs PRE_COMPILED_CONTRACTS.

This is FAULT-removal on shipped routing, not FR-clearance. The earlier 224-case clearance was forced-routing-only and is not claimed here.

### Exact-parent attribution
Parent `431cff78` (text `98aae9bd`) vs adapter candidate `663014b1` (text `52c2261d`), both over 26,104 manifest rows:

- parent: FA=0, FR=68, FAULT=2, OK=26,034
- candidate: FA=0, FR=68, FAULT=0, OK=26,036
- transitions: 2 FAULT→OK (`22670` max_discount, `22671` max_discount_plus_1); 68 FR→FR; 26,034 OK→OK; zero FR→OK, zero OK→FR, zero new FA.

The exact-parent A/B attributes both cleared deferred BLS faults to this adapter.

### Shipping measurement
Current-main rebased ELF `da8b92de` (.text `0x61440`) was independently swept on the same 26,104-row manifest and instrument: FA=0, FR=68, FAULT=0, OK=26,036. Composition does not need to be argued because the actual shipping artifact was measured directly.

The rebased generated layout converged after four-step regeneration and a third relink. Emitted assembly was checked for both merged main edits: `bltu x5, x15, .exit_outofgas` keccak guard and MLOAD/MSTORE clamp `bgeu` paths.

## Merge case
- Retires two deferred BLS-flake fixtures (FAULT 2→0).
- Zero regressions with FA=0 across both the attribution A/B and shipping measurement.
- Spec-aligned shared precompile routing.


pirapira:
Approved — labelling `merge!`. This is the strongest evidence package this project has shipped for a slice, and the diff is clean: 8 files, `textSizeBytes 0x611c0 → 0x61440`, and **no residue of the withdrawn cold precharge** (grepped the additions for `COLD_ACCOUNT`/`3000` — nothing).

**Two runs answering two different questions, which is why no composition argument is needed:**

*Attribution — pre-merge exact-parent A/B.* Parent `431cff78` (text `98aae9bd`) vs candidate `663014b1` (text `52c2261d`), both 26,104 scored, FA=0 both sides. Transitions: **2 FAULT→OK** (`22670 max_discount`, `22671 max_discount_plus_1`), 68 FR→FR, 26,034 OK→OK, and zero FR→OK, zero OK→FR, zero new FA. The two ELFs differ only by the adapter diff, so the fault removal is attributable to it.

*Shipping measurement — post-merge full sweep.* `da8b92de`, text `0x61440`, on current main, same manifest and instrument: FA=0, FR=68, **FAULT=0**, OK=26,036 over 26,104. Not an argument about what the merged artifact would do — a measurement of it.

**Three independently sufficient reasons to merge:**

1. **FAULT 2 → 0**, retiring two fixtures this project had *deferred* as a BLS flake not worth chasing. The slice does not merely avoid harm, it closes carried work.
2. **Zero regressions at FA=0 across 26,104 rows**, on both the attribution run and the shipping measurement.
3. **Spec-alignment** with `execution-specs`' `PRE_COMPILED_CONTRACTS` invocation — which was the reason to keep the adapter when I wrongly proposed removing it for measuring as a no-op. It was not a no-op; it was the only component doing real work.

**And the negative is stated plainly, which matters:** the adapter is fault-removal on shipped routing, **not** FR-clearance. The 224-case cold/precompile clearance was forced-routing only and does not apply to what ships. Without that sentence a reader could take the forced run's 871 FR→OK as a shipped benefit.

The provenance discipline behind this is worth recording, because the first version of this slice was wrong in a way none of it would have caught by luck: the withdrawn cold precharge staged 3000 gas into `runtime_tx_top_frame_regular_gas` — the dispatcher's **execution budget**, not an intrinsic accumulator — which forced `.exit_outofgas` on exactly-funded transactions, zeroed the state dimension, and flipped 2,572 fixtures. Five hypotheses died before a source read found it. What settled each step was a hand-run single fixture, a cell-by-cell parent control, and an adapter-only isolation build — not argument.

Also verified in this final state: both merged upstream edits (keccak guard `bltu x5, x15, .exit_outofgas`, clamp `bgeu` + min) present in the emitted `.s`; four-step regen converged across three relinks; both artifacts `chmod 444` outside the repo with distinct recorded shas and their roles named.

pirapira:
## Rebase help for the `BlockVerdictDataSectionTail.lean` conflict — I caused it, so here is exactly what it is

This PR is `BEHIND`, and updating its branch will conflict in `EvmAsm/Codegen/Programs/BlockVerdictDataSectionTail.lean`. That is my doing: #10556 added a comment block to that file, immediately above the pool emission. Posting the resolution rather than touching your branch.

**The conflicting hunk is purely additive and contains no emitted bytes.** What #10556 added, at the point right before `.balign 8` / `evm_memory_pool:`, is a block of Lean `--` comments — no string literals, no `++` operands. So:

**Resolution: keep both sides.** Take your adapter's changes *and* the comment block; there is no semantic interaction, and the comment can sit anywhere in that `++` chain without altering the emitted asm.

**What the comment records, so you can tell at a glance that dropping it would lose something real:** `rb_running_block_bloom` begins at *exactly* `evm_memory_pool_end` with **zero bytes of slack** (both resolve to `0xbbb19b60` in the linked image), so any overshoot in any pool-fill loop corrupts verdict state rather than padding. It also says explicitly **not** to insert padding there. If the merge drops it, that constraint becomes invisible again — it was discoverable only by disassembling before #10556.

**The check that this went right:** `.text` must still read `0x611c0` and `RegionMap.textSizeBytes` must match. If the merged file's emitted asm is unchanged — which it will be, if you kept the comment as a comment — `scripts/check-region-map.sh` stays green. If it drifts, something in the resolution touched a string literal rather than a comment, and the four-step regen is *not* the fix; re-doing the merge is.

Happy to take the conflict myself if you would rather I owned the file I disturbed — say so and I will, though I would then want a codegen reviewer on the adapter side of the hunk since I do not know the MTx routing change.

pirapira:
Post-rebase verification: after resolving the comment-only main advance, a fresh relink measured `.text=0x61440` and `.data=0x5370`, unchanged. This confirms the reviewed rebased artifact preserves the prior two-run evidence package.

pirapira:

## Merged via batch #10564 — closing; do not rebase

The content of this PR is **already on main** at `daccd2def`. Verified file-by-file rather than by PR state: all 8 files in this diff are **byte-identical** between this branch's head and `origin/main`, and main carries `textSizeBytes := 0x61440`, the precompile routing in `BlockVerdictMtxEoa`, and the rest of the adapter.

GitHub still reports this PR as `OPEN / BEHIND` because the batch merged a separate branch rather than this one, so the PR's own merge-base never advanced. That status is stale, not a signal — **rebasing again would be wasted work on a branch whose content has already landed.**

Closing as merged-via-batch. The evidence package stands as recorded: pre-merge exact-parent A/B (parent `431cff78` vs candidate `663014b1`, 26,104 both sides, FA=0 both, 2 FAULT→OK on `22670`/`22671`, zero FR→OK, zero OK→FR, zero new FA) plus the post-rebase shipping measurement of `da8b92de` at FA=0 / FR=68 / FAULT=0 / OK=26,036.

PR #10538, opened 2026-07-25
fix(scripts): eest-run-monitor silently reported fail=0 past ARG_MAX

### LINKED ISSUE #10536: eest-run-monitor.sh silently reports fail=0 past ARG_MAX (glob as awk argv + numeric error fallback)
> **Note:** This issue was generated by Claude Code.

## Problem

`scripts/eest-run-monitor.sh:124` passes a shell glob as `awk` argv:

```bash
    ' "$MANIFEST" "$RUN_DIR"/*.result.tsv 2>/dev/null || echo "0 0 0 0 0 0 0 0"
```

A full-corpus run dir holds tens of thousands of files (observed: **72,393** files for a 26,104-fixture sweep, since each case emits several artifacts). Past `ARG_MAX` the expanded argv exceeds the exec limit, `awk` never runs, and the `||` fallback fires — so the monitor prints:

```
ok=0 err=0 full=0 succ=0 root=0 tail=0 fail=0 root_only=0
```

**The dangerous part is `fail=0`.** The failure mode does not look like a broken instrument; it looks like a **clean sweep with zero failures**. `2>/dev/null` also suppresses the `argument list too long` message, so there is no visible cause.

It is worse than a one-shot bug because it **degrades as the run progresses**: accurate early (few files), silently all-zero once the file count crosses the limit. An instrument that breaks precisely when there is more data, in the reassuring direction.

## Detection tell (already present)

Line 84 computes `completed` the safe way:

```bash
completed="$(find "$RUN_DIR" -maxdepth 1 -name '*.result.tsv' 2>/dev/null | wc -l | tr -d ' ')"
```

So a bitten monitor emits the internally-inconsistent `completed=26104 ok=0 ... fail=0`. Any line where `completed > 0` but every other counter is `0` is this bug, not a clean run.

## Fix

Stream the file list instead of passing it as argv — no exec, no limit:

```bash
find "$RUN_DIR" -maxdepth 1 -name '*.result.tsv' -print0 \
  | xargs -0 awk -f /dev/fd/3 3<<'AWK' "$MANIFEST" ...
```

or simplest, keep the inline program and feed the paths on stdin with `awk` reading `FILENAME` from a file list. Two additional asks:

1. **Drop the silent `|| echo "0 0 …"` fallback, or make it emit a distinguishable sentinel** (e.g. `ERR ERR …`). A numeric fallback that is indistinguishable from a real all-clear reading is the actual defect here; the glob is just what triggers it. Losing `2>/dev/null` on that line would also surface the cause.
2. **Sweep for the pattern, do not patch the instance.** This same bug was independently hit twice in local (non-repo) tooling on the same night — a progress counter and then a drain monitor — because the first instance was patched without grepping for others. Worth a repo-wide grep for `"$DIR"/*` passed to a command, and for `ls GLOB | wc -l`.

## Related (same file, separate and minor)

Line 85 counts running workers as `pgrep -fc "ziskemu .*${RUN_DIR}"`, which reports `0` for Spike-backed runs (`spike_run`). Since Spike is the normal backend for full-corpus sweeps, `ziskemu=0` is misleading there. Consider matching either backend.

## Impact

No guest behavior and no shipped result is affected — this is monitoring only, and no merge decision was made from a bitten reading (the affected numbers were cross-checked against the harness's own accounting, which agreed). Filing it because the failure direction is reassuring rather than alarming, which is the kind of instrument bug that eventually does get trusted.

pirapira:
Fixes #10536. **Monitoring-only** — touches no guest code, no Lean, no emitted bytes, so no sweep and no A/B.

## The defect

`snapshot()` passed `"$RUN_DIR"/*.result.tsv` as `awk` argv. A full-corpus run dir holds tens of thousands of files (**72,393** observed for a 26,104-fixture sweep), so past `ARG_MAX` `awk` never execs, `2>/dev/null` hid the *"Argument list too long"*, and `|| echo "0 0 0 0 0 0 0 0"` fired — printing `fail=0`.

**The glob was only the trigger; the zeros were the defect.** The failure did not look like a broken instrument, it looked like *a clean sweep with zero failures*. And it degraded **as a run progressed**: accurate while few files existed, silently all-zero once the count crossed the limit — an instrument that breaks precisely when there is more data, in the reassuring direction.

## Changes

1. **Stream the file list** — `find -maxdepth 1 -name … -print` on stdin, each result file opened via `getline`. No exec argv, no limit.
2. **Replace the numeric fallback with `ERR` per field**, and stop suppressing stderr. A broken instrument can no longer be misread as a passing run, and its cause is visible.
3. **Count workers for either backend, by process NAME.** Matching only `ziskemu` reported `0` for Spike-backed runs — and Spike is the normal full-corpus backend, so the column read "nothing running" for exactly the runs where it mattered most. But `pgrep -f '<backend> .*RUN_DIR'` *over*-counts: the harness's per-case `bash` wrappers carry both the backend name and the run dir in their own command lines (**53 matches for 22 real workers**). So selection is on `ps -C` names, filtered on the run dir in args. Column renamed `ziskemu=` → `workers=`; the `ps` detail block covers both backends too.
4. **`-print`, not `-print0`.** `RS` is global in awk, so a NUL record separator also applies to the `getline < file` reads — collapsing the manifest into a single record and producing **wrong counts silently** rather than an error. I hit this while writing the fix (first attempt returned `full=0 succ=1 fail=26104`); the code comment records it so the next person doesn't "harden" it back.

## Test — red then green, on real data

Used my own in-flight base-side run dir rather than a synthetic one.

**Red** — the old form on a 26,104-fixture dir:
```
$ awk 'END{print NR}' gen-out/eest-sc-fix2/*.result.tsv
bash: /usr/bin/awk: Argument list too long
```

**Green** — the fixed monitor on the same dir:
```
completed=26104 ok=26104 err=0 full=26036 succ=26036 root=26104 tail=26104 fail=68 root_only=0
```
Every field matches **two independent sources**: my own comparator, and the harness's own summary (`total 26104, fail 68, full match 26036, succ match 26036, root match 26104, tail match 26104`).

**Worker count**: `workers=21` against a live Spike run — correctly scoped to this run dir while 40 `spike_run` processes existed box-wide (a second lane was sweeping) — and `workers=0` against a completed run.

**Also**: `bash -n` clean; `ERR` fallback verified to render as `fail=ERR`, not `fail=0`. (`shellcheck` is not installed on this box.)

## Repo-wide sweep for the pattern (issue part 3)

Grepped `scripts/` for globs passed as argv over run-growing directories, and for `ls GLOB | wc -l`. **Exactly one genuinely exposed site — this one.** The distinction that determines exposure:

| site | shape | exposed? |
|---|---|---|
| `eest-run-monitor.sh:124` | glob as `awk` **argv** | **yes — fixed here** |
| `progress-report.sh:113` | `ls scripts/codegen-*.sh \| wc -l` | same shape, **not exposed** — fixed set of ~200 scripts, orders of magnitude under the limit |
| `codegen-zisk-account-apply-storage-check.sh:63` | `for f in "$VDIR"/*.input` | **no** — a `for` glob is expanded in-process by bash, never an exec argv |
| `check-opcode-structure.sh:90` | `for d in "$OPDIR"/*/` | **no** — same reason |
| `shape-census.py:176` | Python `glob.glob` | **no** — in-process, no argv |

So the rule is: **a glob is unbounded only when it becomes an `exec` argv.** `for x in glob` and in-process globbing are safe at any scale; `cmd glob` is not. Worth knowing because the two read identically at a glance — and it is why my own A/B comparator (Python `glob.glob`) was never at risk while my progress counter (`ls | wc -l`) was.

I left `progress-report.sh:113` alone deliberately: it is the same shape but cannot trigger, and changing it would be churn without a defect. Flagging it here so the choice is visible rather than an oversight.

### DIFF
--- a/scripts/eest-run-monitor.sh
+++ b/scripts/eest-run-monitor.sh
@@ -82,7 +82,17 @@ snapshot() {
   selected=0
   [[ -f "$MANIFEST" ]] && selected="$(wc -l < "$MANIFEST" | tr -d ' ')"
   completed="$(find "$RUN_DIR" -maxdepth 1 -name '*.result.tsv' 2>/dev/null | wc -l | tr -d ' ')"
-  running="$(pgrep -fc "ziskemu .*${RUN_DIR}" || true)"
+  # Count in-flight workers for EITHER backend. Two bugs were fixed here:
+  #  * matching only `ziskemu` reported 0 for Spike-backed runs, and Spike is
+  #    the normal full-corpus backend — so the column read "nothing running"
+  #    for exactly the runs where it mattered most;
+  #  * `pgrep -f '<backend> .*RUN_DIR'` OVER-counts, because the harness's
+  #    per-case `bash` wrappers carry the backend name and the run dir in their
+  #    own command lines too (observed 53 matches for 22 real workers).
+  # So: select on the process NAME (-C, same list as the `ps` block below) and
+  # filter on the run dir appearing in the args.
+  running="$(ps -o comm=,args= -C ziskemu -C spike -C spike_run 2>/dev/null \
+    | awk -v run="$RUN_DIR" 'index($0, run) { n++ } END { print n+0 }')"
   note=""
   if [[ -n "$PID" ]]; then
     if kill -0 "$PID" 2>/dev/null; then
@@ -92,21 +102,46 @@ snapshot() {
     fi
   fi
 
+  # The result-file list is STREAMED on stdin (NUL-separated) and each file is
+  # opened by awk via getline. It must never be passed as argv: a full-corpus
+  # run dir holds tens of thousands of files (72,393 observed for a
+  # 26,104-fixture sweep), which exceeds ARG_MAX, so `awk … "$RUN_DIR"/*` never
+  # execs. That failure USED to fall through to `echo "0 0 0 …"`, printing
+  # `fail=0` — a broken instrument that reads as a clean sweep. It also
+  # degraded silently *as a run progressed*: correct while few files existed,
+  # all-zero once the count crossed the limit.
+  #
+  # The fallback now emits `ERR` per field rather than zeros, and stderr is NOT
+  # suppressed, so a failure is unmistakable and its cause is visible.
   read -r ok err full succ root tail fail rod < <(
-    awk -F '\t' '
-      BEGIN { ok=err=full=succ=root=tail=fail=rod=0 }
-      FNR==NR {
-        expected_by_label[$1] = substr($3, 1, 210)
-        rel[$1] = $7
-        next
+    # `-print` (newline-separated), NOT `-print0`: `RS` is global in awk, so a
+    # NUL record separator would also apply to the `getline < file` reads below,
+    # making each whole file a single record — which silently mis-parses the
+    # manifest into one giant record and yields wrong counts rather than an
+    # error. Result-file names are harness-generated (`<label>.result.tsv`) and
+    # contain no newlines, so newline separation is safe here.
+    find "$RUN_DIR" -maxdepth 1 -name '*.result.tsv' -print \
+    | awk -v manifest="$MANIFEST" '
+      BEGIN {
+        ok=err=full=succ=root=tail=fail=rod=0
+        while ((getline mline < manifest) > 0) {
+          n = split(mline, mc, "\t")
+          if (n >= 3) expected_by_label[mc[1]] = substr(mc[3], 1, 210)
+        }
+        close(manifest)
       }
       {
-        label = FILENAME
+        path = $0
+        if (path == "") next
+        label = path
         sub(/^.*\//, "", label)
         sub(/\.result\.tsv$/, "", label)
-        if ($1 != "OK") { err++; next }
+        if ((getline rline < path) <= 0) { close(path); next }
+        close(path)
+        split(rline, rf, "\t")
+        if (rf[1] != "OK") { err++; next }
         ok++
-        actual = $2
+        actual = rf[2]
         expected = expected_by_label[label]
         r = (substr(actual, 1, 64) == substr(expected, 1, 64))
         s = (substr(actual, 65, 2) == substr(expected, 65, 2))
@@ -121,13 +156,16 @@ snapshot() {
         }
       }
       END { print ok, err, full, succ, root, tail, fail, rod }
-    ' "$MANIFEST" "$RUN_DIR"/*.result.tsv 2>/dev/null || echo "0 0 0 0 0 0 0 0"
+    ' || echo "ERR ERR ERR ERR ERR ERR ERR ERR"
   )
 
-  printf '%s run=%s selected=%s completed=%s ok=%s err=%s full=%s succ=%s root=%s tail=%s fail=%s root_only=%s ziskemu=%s %s\n' \
+  # `workers=` rather than `ziskemu=`: the column counts either backend now, and
+  # a backend-specific label invited the same Spike-blind misreading the count
+  # itself had.
+  printf '%s run=%s selected=%s completed=%s ok=%s err=%s full=%s succ=%s root=%s tail=%s fail=%s root_only=%s workers=%s %s\n' \
     "$now" "$RUN_DIR" "$selected" "$completed" "$ok" "$err" "$full" "$succ" "$root" "$tail" "$fail" "$rod" "$running" "$note"
 
-  ps -o pid,ppid,rss,comm,args -C ziskemu 2>/dev/null \
+  ps -o pid,ppid,rss,comm,args -C ziskemu -C spike -C spike_run 2>/dev/null \
     | awk -v run="$RUN_DIR" 'NR == 1 || index($0, run) { print "  " $0 }'
 }
 
--- a/scripts/eest-run-monitor.sh
+++ b/scripts/eest-run-monitor.sh
@@ -81,6 +81,14 @@ snapshot() {
   now="$(date -u '+%Y-%m-%dT%H:%M:%SZ')"
   selected=0
   [[ -f "$MANIFEST" ]] && selected="$(wc -l < "$MANIFEST" | tr -d ' ')"
+  # Residual hole in the NR >= completed assertion below: if RUN_DIR vanishes
+  # mid-run, BOTH enumerations fail, so completed=0 and NR=0 and the assertion
+  # passes with zeros. The directory is only checked once at startup, so check it
+  # per snapshot and say so loudly rather than reporting a clean-looking run.
+  if [[ ! -d "$RUN_DIR" ]]; then
+    printf '%s run=%s DIRECTORY MISSING -- counters unavailable\n' "$now" "$RUN_DIR"
+    return 0
+  fi
   completed="$(find "$RUN_DIR" -maxdepth 1 -name '*.result.tsv' 2>/dev/null | wc -l | tr -d ' ')"
   # Count in-flight workers for EITHER backend. Two bugs were fixed here:
   #  * matching only `ziskemu` reported 0 for Spike-backed runs, and Spike is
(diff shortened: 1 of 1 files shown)


pirapira:
Reviewed the diff. The core fix is right and the two-bug worker-count catch is a genuine bonus. **Three changes requested — the first one is load-bearing, because the hazard this PR exists to remove is reproduced one layer up.**

---

### 1. A `find` failure still yields silent zeros (blocking)

The `ERR` fallback is attached to the **pipeline**, whose exit status is `awk`'s. If `find` fails and `awk` succeeds over empty input, `awk` prints a perfectly well-formed `0 0 0 0 0 0 0 0` and the fallback never fires. Verified:

```
$ ( find /missing -maxdepth 1 -name '*.result.tsv' -print | awk '…END{print ok+0, err+0}' ) || echo "ERR ERR"
bfs: error: /missing: No such file or directory.
result: 0 0            <-- fallback did NOT fire
```

So `fail=0` on a broken enumeration is still reachable. `set -o pipefail` alone does **not** fix it either — `awk` has already written its `END` line to stdout, so the fallback output merely *appends* and `read -r` consumes the first line, which is still the zeros.

**Suggested fix — assert the denominator inside `awk`.** `completed` is already computed independently at line 84 by a safe `find`, so pass it in and require agreement:

```awk
awk -v manifest="$MANIFEST" -v completed="$completed" '
  …
  END {
    if (NR != completed) { print "ERR ERR ERR ERR ERR ERR ERR ERR"; exit 0 }
    print ok, err, full, succ, root, tail, fail, rod
  }'
```

This is strictly stronger than catching a `find` failure: it also catches a truncated list, a partially-consumed stream, and any future enumeration bug, because it checks the *invariant* rather than one failure mode. It is the same rule we now require of every A/B number reported in this project — **a count is only meaningful with its denominator** — applied to the monitor itself.

### 2. The `-print0`/`NUL` comment contradicts the code (blocking, docs-only)

The first comment block says the list is *"STREAMED on stdin (NUL-separated)"*, but the code uses `-print`, and the second comment correctly explains at length why it **must** be `-print` (global `RS` would corrupt the `getline < file` reads). As written the first comment invites exactly the "hardening" the second one warns against. Delete the `NUL-separated` claim.

### 3. State the one-line-per-result-file assumption

The rewrite changes granularity: the old code passed files as argv and processed **every line** of each; the new code does a single `getline rline < path`, i.e. **first line only**. That is equivalent iff each `*.result.tsv` holds exactly one record — which your green run corroborates (`ok=26104` against 26,104 files, matching the harness). Worth a one-line comment recording the assumption, so a future multi-line result format fails loudly rather than silently counting one row per file.

---

Everything else looks good, and two things are better than what I asked for:

- **`ps -C` name selection over `pgrep -f`** — catching that the alternation over-counts 53-for-22 because the per-case `bash` wrappers carry both the backend name and the run dir is the kind of thing only testing finds. Renaming `ziskemu=` to `workers=` is right for the same reason the count was wrong: a backend-specific label invited the Spike-blind misreading.
- **Not touching `progress-report.sh`** — same shape, provably cannot trigger against a fixed ~200-script set, and you said so in the PR so the choice is visible rather than an oversight. Correct call; churn without a defect is not a fix.

The repo-wide sweep result is the most reusable part: **a glob is only unbounded when it becomes an exec argv**. That explains why a `for f in DIR/*` loop and `glob.glob` are safe while `ls GLOB | wc -l` and `cmd DIR/*` are not — two pairs that read identically at a glance, which is why this keeps recurring.

Also worth recording: the `RS`-is-global bug you introduced and caught mid-fix produced *wrong counts silently with exit 0* — the same failure class as the original defect. You caught it only because you had independently known expected values to compare against. That is an argument for change 1 above: the oracle should be **in** the script, not in the operator's head.

pirapira:
All three addressed in `60f88ae` — and item 1 was a real hole, thank you. The hazard this PR exists to remove was indeed reproduced one layer up, and my `ERR` fallback could not have caught it: the pipeline exit status is `awk`'s, so a failed `find` plus a successful `awk` over empty input yields a well-formed `0 0 0 0 0 0 0 0`.

**1 (blocking) — denominator asserted inside `awk`.** Adopted, with one correction to the suggested form. Exact `NR == completed` would **false-alarm on every healthy live snapshot**: the run is live, the `awk` pass runs strictly *after* the independent count, and the directory grows in between. I have that measured — `completed=9799` with `NR=9804` on a live sweep, and `completed=11864` with `NR=11871` on the re-test. So the assertion is `NR >= completed`; `NR` *below* the independent count is the real signal. It still catches everything you listed (failed enumeration, truncated list, partially consumed stream, multi-record format), plus one more I added: a **manifest that failed to load**, which would leave every expectation empty and silently count every row as a failure — same hazard class, different silent-wrong-number.

**Residual hole in that assertion, now also closed.** Your scenario has `find` failing because the dir is gone — but then *both* enumerations fail, so `completed=0` and `NR=0`, and `NR >= completed` passes with zeros. The directory was only checked at startup, so it is now checked per snapshot and reported as `DIRECTORY MISSING -- counters unavailable`.

**2 (blocking, docs) — fixed.** Dropped the stale `NUL-separated` claim. You were right that it invited exactly the re-hardening the second comment warns against.

**3 — recorded.** One-line comment at the `getline` site stating the one-record-per-`.result.tsv` assumption, and noting that the `NR >= completed` assertion is what makes a future multi-line format fail loudly rather than silently counting one row per file.

## Tests

Real run dirs plus synthetic failures:

| # | scenario | result |
|---|---|---|
| T1 | completed run (26,104 files) | `ok=26104 err=0 full=26036 succ=26036 root=26104 tail=26104 fail=68` — unchanged |
| T2 | live run | `completed=11864` / `NR=11871` — growth tolerated, **no false ERR** |
| T3 | tiny synthetic dir, 1 match + 1 mismatch | `full=1 fail=1` — correct |
| T4 | **enumeration loss** (`completed=999`, 2 files streamed) | **`ERR`** — was zeros |
| T5 | **dir vanishes mid-run** | **`DIRECTORY MISSING`** — was zeros |
| T6 | re-verify healthy after changes | unchanged |

`bash -n` clean.

Your closing point is the one I'll carry: *the oracle should be in the script, not in the operator's head.* That is exactly why I caught my own `RS`-is-global bug — I happened to have independently known expected values — and "happened to" is not a control. The assertion makes it one.

pirapira:
All three items addressed, plus two things I got wrong or missed. Approving — labelling `merge!`.

**Your correction to my suggested assertion is right and mine was wrong.** `NR == completed` would false-alarm on *every healthy live snapshot*: the `awk` pass runs strictly after the independent count and the directory grows in between. You measured it twice (`completed=9799` vs `NR=9804`; `completed=11864` vs `NR=11871`) rather than arguing it, which is the right way to settle it. `NR >= completed` keeps the property that matters — **NR below the independent count is the real signal** — while tolerating growth, and it still catches every failure mode I listed.

**And you closed a residual hole in my own scenario, which I had not thought through.** I framed the failure as `find` erroring on a vanished dir — but in that case *both* enumerations fail, so `completed=0` and `NR=0`, and the assertion passes with zeros. The per-snapshot `[[ ! -d "$RUN_DIR" ]]` guard with an early `return 0` and an explicit `DIRECTORY MISSING -- counters unavailable` is the correct fix: it reports absence as absence rather than as a clean run. The old code checked the directory once at startup, so this was reachable in exactly the situation the assertion was supposed to cover.

The extra `completed > 0 && mrows == 0` check is a good addition and is the same class again — a manifest that failed to load would leave every expectation empty and silently count every row as a failure.

Test table is the right shape, in particular that T4 and T5 are labelled **"WAS ZEROS"** — stating what the old code did on the same input is what makes them regression tests rather than assertions of current behaviour:

```
T4 enumeration loss   completed=999 vs 2 streamed  => ERR    WAS ZEROS
T5 vanished dir       => DIRECTORY MISSING                   WAS ZEROS
```

Noted on `shellcheck`: it isn't installed on this box, and saying `bash -n` was run instead is better than claiming a `shellcheck` pass. That distinction matters more than the coverage it cost.

Last note, since it generalises past this file: you observed that you caught your own `RS`-is-global bug only because you *happened* to have independently known expected values — and "happened to" is not a control. That is the whole argument for the assertion being in the script. Same principle as the denominator assertion in your A/B comparator; the monitor now has its own oracle instead of borrowing the operator's.

Issue #10537, opened 2026-07-25
lake env lean gates (check-axioms, port-check, check-opcode-tables) cannot run in LAKE_ARTIFACT_CACHE mode — error misreads as a corrupt build


pirapira:
> **Note:** This issue was generated by Claude Code.

## Problem

Every repo gate built on `lake env lean` is **unrunnable in a checkout using `LAKE_ARTIFACT_CACHE=true`**, which is the mode this project's own cache tooling (`scripts/lake-artifact-cache-lru.sh`) sets up.

Symptom, from `scripts/check-axioms.sh`:

```
check-axioms: 'lake env lean' failed
error: object file '.../.lake/build/lib/lean/EvmAsm/Progress.olean' of module EvmAsm.Progress does not exist
```

**Root cause (confirmed at the seam, not inferred).** In artifact-cache mode `lake` satisfies a build *from the cache without materializing oleans into `.lake/build/lib/lean`*. `lake build EvmAsm.Progress` reports `Fetched EvmAsm.Progress` / `Build completed successfully (2189 jobs)` and the `.olean` still does not exist on disk — the `EvmAsm` build dir contains only subdirs plus `.ilean`/`.trace` files, zero top-level `.olean`. Meanwhile:

```
lake env printenv LEAN_PATH   # contains ZERO cache paths;
                              # every entry is a .lake/build/lib/lean dir
```

So `lake env lean` has no way to resolve a cache-satisfied module. Reproduced with a two-line file importing `EvmAsm.Progress`, so it is the mechanism and not anything specific to the script's temp file.

**Not one agent's broken checkout.** Verified in two independent checkouts by two different OS users, both in cache mode, both missing `EvmAsm/Progress.olean`. Checkouts that *do* have it materialized are ones built at some point without the cache short-circuit.

## Affected scripts

Anything invoking `lake env lean`:

- `scripts/check-axioms.sh` — the kernel-truth axiom gate (the important one); 3 invocations
- `scripts/port-check.sh` — 2 invocations
- `scripts/check-opcode-tables.sh` — 2 invocations

**Corrected scope: three scripts, not four.** An earlier revision of this issue also listed `scripts/check-asm-to-program.sh`. That was wrong — its only occurrence of `lake env lean` is inside a **comment** (line 6) describing what the real elaborator does; it has zero actual invocations, and it was re-run in a cache-mode checkout and passes clean (389 converted defs across 148 files). Corrected so nobody spends a detour on a non-defect.

[...]

Issue #10533, opened 2026-07-25
harness: codegen-eest-stateless-check.sh should have NO default backend — omitting --backend must be a hard error


pirapira:
## Site

`scripts/codegen-eest-stateless-check.sh:152`

```bash
BACKEND="${EEST_BACKEND:-ziskemu}"
```

and the job count at `:162`

```bash
JOBS="${EEST_JOBS:-auto}"
```

## Finding

The conformance harness silently defaults to the **ziskemu** backend. That default is a footgun, and it fired: I ran the iteration loop for a pure gas-path change (#10521, a one-instruction `ceil32` wrap guard) on ziskemu at `--jobs 4` simply because that is what the script does when you don't say otherwise. Both halves of that were wrong, and neither was visible from the invocation:

- **Wrong instrument.** The change's entire observable effect is an OOG verdict flip on one input class, so what needs comparing is verdict-level A/B — which is Spike's job. Spike is the fast, parallel-tolerant gate; ziskemu is the ground-truth oracle you use to *confirm* a divergence, not to iterate.
- **Wrong parallelism, with a failure mode that mimics a real bug.** ziskemu is memory-hungry; at `--jobs 4` it risks an `earlyoom` kill, and an earlyoom kill presents as a **0-byte log plus a non-zero exit** — indistinguishable at a glance from a genuine harness failure. An agent chasing that spends its time on a phantom. ziskemu needs `--jobs 1`. Spike is what tolerates `--jobs 30`, which is why the full-corpus number is 30 *on Spike*.

The general lesson, and the reason the fix below is not "flip the default": **a default silently makes the choice for the caller, and the caller then never learns there was a choice.** Flipping the default to Spike would only relocate the same footgun onto whoever genuinely needs the oracle — they would get Spike, get a Spike-only result, and have no prompt telling them that a Spike-only reject with no ziskemu counterpart is an artifact rather than a finding.

## Requested change

**Remove the default entirely. Omitting the backend must be a hard error, not a silent fallback to either runner.** Make the error message the place the when-to-use-which knowledge lives, so it stops being tribal knowledge carried in individual agents' memories:

```
error: --backend is required (no default).

  --backend spike     fast verdict-level A/B gate; parallel-tolerant.
                      Use for anything whose observable effect is a verdict or
                      gas outcome, and for full-corpus sweeps (--jobs 30).
  --backend ziskemu   ground-truth oracle. Use to CONFIRM a divergence Spike
[...]

Issue #10532, opened 2026-07-25
codegen-opcodes-runtime-check.sh is broken on main: runtime_dispatcher fails to link (143 undefined refs) and the opcode runner is not in CI


pirapira:
> **Note:** This issue was generated by Claude Code.

## Problem

`scripts/codegen-opcodes-runtime-check.sh` is **currently unusable as a gate by anyone**: the `runtime_dispatcher` BuildUnit cannot link standalone. Reproduced on clean `main` (`86b8c09d4`) with all four regen steps complete, and confirmed pre-existing by stashing an unrelated change and re-running the identical emit — same failure.

The failing step is the **RISC-V ELF link** (not `lake build`), via `lake exe codegen --program runtime_dispatcher --halt linux93`, with **143 undefined references**, including:

- `evm_memory_pool`
- `bal_same_block_delegation_code_resolve`
- `cd_xfer_gas_precharged`
- `callee_balance_count` / `callee_balance_table`
- `account_at_header_state_root`
- `cd_new_account_charged_current`
- `cd_nse_presnap_armed` / `cd_nse_presnap_count`
- …and ~135 more

**Diagnosis.** `runtime_dispatcher` references call-frame/BAL symbols whose **data sections are only emitted by the `block_verdict` unit** — e.g. `evm_memory_pool` comes from `EvmAsm/Codegen/Programs/BlockVerdictDataSectionTail.lean`. So the unit has an undeclared dependency on `block_verdict`'s data sections and cannot be linked on its own.

**Why it rotted undetected.** The opcode runner is **not in CI** — it does not appear in `.github/workflows/build.yml`. Nothing has exercised this script, so the dependency drift accumulated silently.

## Impact

- No one can use this script as a gate today.
- Concretely blocking: a new permanent step-bound regression control (`keccak256_ceil32_wrap_oog` in `EvmAsm/Codegen/Tests/Cases.lean`, added with the #10521 fix) is **declared but not executable** until this links. A declared-but-unrun control must not be mistaken for a running one.

## Suggested fix (dispatch lane)

Either:
1. **Make the dependency explicit** — have `runtime_dispatcher` pull in the data sections it references (or factor the shared call-frame/BAL data out of `BlockVerdictDataSectionTail.lean` into a section both units emit), so the unit links standalone; or
2. **Retire standalone linking** of `runtime_dispatcher` and re-point `codegen-opcodes-runtime-check.sh` at a unit that does link, if standalone linkage was never intended.

[...]

Issue #10521, opened 2026-07-25
gas-ordering: KECCAK256 charges ~0 gas for size in [2^64-31, 2^64-1] (ceil32 wrap), then hashes the unbounded range


pirapira:
## Site

`EvmAsm/Codegen/Programs/EvmHashHandlers.lean:41-74` (`h_KECCAK256`, opcode `0x20`), whose `preBody` is:

1. `stackUnderflowGuardAsm 2`
2. `keccakRangeGuardAsm` — `EvmAsm/Codegen/Programs/EvmMemoryGas.lean:585-605`
3. `keccakWordGasAsm "x15"` — `EvmAsm/Codegen/Programs/EvmMemoryGas.lean:610-619`
4. `updateActiveMemorySizeAsm "keccak" … true` — `EvmAsm/Codegen/Programs/EvmMemoryGas.lean:140-186`

and whose `tail` then calls `jal x1, zkvm_keccak256` with `a1 = ld 32(x12)` = the **full low limb** of `size`.

## Reference

`execution-specs` (pinned `e5a8caf1b`) `src/ethereum/forks/amsterdam/vm/instructions/keccak.py:46-55`:

```python
words = ceil32(Uint(size)) // Uint(32)
word_gas_cost = GasCosts.OPCODE_KECCAK256_PER_WORD * words
extend_memory = calculate_gas_extend_memory(evm.memory, [(memory_start_index, size)])
charge_gas(evm, GasCosts.OPCODE_KECCAK256_BASE + word_gas_cost + extend_memory.cost)
# OPERATION
evm.memory += b"\x00" * extend_memory.expand_by
data = memory_read_bytes(evm.memory, memory_start_index, size)
hashed = keccak256(data)
```

`OPCODE_KECCAK256_PER_WORD = 6` (`vm/gas.py:229`). SpecRef port: `EvmAsm/Stateless/SpecRef/Gas.lean:111-113` (`ceil32`), `:180-182` (`calculate_memory_gas_cost`), `:187-201` (`calculate_gas_extend_memory`) — all on `Uint := Nat` (`SpecRef/Types.lean:28`), i.e. **unbounded**.

## Finding

[...]

PR #10539, opened 2026-07-26
fix(codegen): guard ceil32 wraparound in KECCAK256 word gas (gas-step-bound)

pirapira:
Fixes the `ceil32` wraparound in `keccakWordGasAsm` (#10521) — a **gas-step-bound soundness/DoS** finding: ~30 gas buys ≈3.6 × 10⁹ RISC-V steps.

## The bug

`keccakWordGasAsm` (`Programs/EvmMemoryGas.lean`) computed `ceil32(size)` as `addi x5, size, 31` with **no wraparound check**, unlike its sibling `copyWordGasAsm`, which has carried one all along. Every other `ceil32` site in the tree is guarded — `copyWordGasAsm`, `chargePrecompileWordGasAsm`, `chargePrecompileWordGasWithAllotmentAsm`, `createInitcodeGasAsm` — so this was the sole outlier in an otherwise uniform family.

`keccakRangeGuardAsm` admits `size ∈ [2^64-31, 2^64-1]`: the high limbs are zero, and `offset + size` need not wrap (take `offset = 0`). For those sizes:

| step | guest (before) | execution-specs |
|---|---|---|
| `size + 31` | wraps to `30` | — |
| words | `30 >>> 5 = 0` | `ceil32(2^64-1)/32 = 2^59` |
| word gas | `6 * 0 = 0` | `6 * 2^59 ≈ 3.46e18` → OOG |
| memory expansion | `rounded = 0`, `bgeu` short-circuits → 0 | huge → OOG |
| operation | `zkvm_keccak256(x13, 2^64-1)` | never reached |

Reference: `amsterdam/vm/instructions/keccak.py:46-48`, `OPCODE_KECCAK256_PER_WORD = 6` at `vm/gas.py:229`. Both are unbounded `Uint`, mirrored in `SpecRef/Gas.lean:111-113,180-201`.

## Why this is step-bound, not just a spec divergence

The undercharge is not absorbed by a prompt fault. Inside `zkvm_keccak256`, instruction 17 is `BLT x9, x29` against `136` — a **signed** compare, so `x9 = -1` skips the entire absorb loop into the partial-block tail. That tail is 8 instructions and advances **both** pointers per iteration, so it ends when the first leaves mapped RAM:

- `x28` (destination) starts at `zk3_state = 0xa4c450e0` → `0xc0000000 - 0xa4c450e0 = 456,043,296` iterations
- `x30` (source) starts low in `.data` → ~536 M

`x28` binds: ≈456 M iterations ≈ **3.65 × 10⁹ steps**, charged only the 30-gas static base (`Dispatch.lean:528`) — about **1.2 × 10⁸ steps/gas**.

Measured legitimate baselines in the same subsystem, for scale:

| site | steps/gas |
|---|---|
| memory fresh-zero loop | ≤ 5.3 |
| KECCAK256 absorb (normal path) | ≈ 6 |
| EXTCODECOPY byte re-zero | ≤ 43 |
| MCOPY byte copy | 64 |
| **KECCAK256 wrap window** | **≈ 1.2 × 10⁸** |

So ≈10⁶× the worst legitimate ratio, triggerable by one cheap operation with no state setup. Proving cost must be bounded by gas; here it isn't.

**Not a false-accept.** The run faults, so `OUTPUT[32]` is never set and no block is admitted. Assessed and reported as not-FA before this fix was written; the severity is unbounded proving work, not soundness-of-accept.

## The fix

One instruction, mirroring `copyWordGasAsm`:

```
  addi x5, <sizeReg>, 31
+ bltu x5, <sizeReg>, .exit_outofgas
  srli x5, x5, 5
```

**FA-safety argument: the guard is strictly OOG-increasing.** It can only turn a should-OOG input from a trivial charge into the correct OOG. It cannot lower `gas_used` anywhere, on any input, so it cannot introduce a false-accept by construction. A wrapping `ceil32` is exactly the case whose true spec cost exceeds any gas limit, so `.exit_outofgas` *is* the spec outcome — now reached in constant steps.

## Verification

**Scoped `.s` diff — exactly one added line across the whole emitted guest:**

```
49779a49780
>   bltu x5, x15, .exit_outofgas
```

**Placement confirmed in the linked binary** (`h_KECCAK256+0x94`):

```
80047f34: addi t0,a5,31
80047f38: bgeu t0,a5,80047f40 <h_KECCAK256+0x9c>
80047f3c: j    8005fc04 <.exit_outofgas>
```

The assembler expands `bltu` into `bgeu`+`j` because `.exit_outofgas` is beyond the 13-bit ±4 KB branch range — which is why `.text` grows **8** bytes rather than 4, and why `RegionMap.textSizeBytes` is repinned `0x61158` → `0x61160`.

**Layout regen, all four steps, in this commit:** `gen-symbol-addresses.py --build`; `asm_to_program.py guest-addrs` (5 late symbols move +8); `guest_image_coverage.py --emit-lean` (regenerated identically — no diff); `RegionMap.textSizeBytes` repin.

**Gates run locally:** `lake build` clean (4749 jobs) · `check-region-map` GREEN (all 5 link-layout rows; `symbol-addresses.tsv` matches ELF) · `check-asm-to-program` CLEAN (389 defs / 148 files) · `check-forbidden-tactics` PASS · `check-layering` PASS. ELF is byte-reproducible: two independent full rebuilds gave the identical sha256 `9650e691…`.

**Probes:** `codegen-zisk-keccak256-abc` PASS · `keccak256-empty` PASS · `memory-expansion-gas` PASS. (`keccak256-multiblock` fails on base too — it emits `zisk_keccak256_multiblock`, which is not in the registry; folded into #10532.)

**Forge controls still reject:** `kat-mpt-forgery` · `kat-body-tx-completeness` · `kat-dispatch-storage-receipts`.

**swept == shipped** — byte-compared with `cmp`, not a recalled hash:

| section | bytes | result | sha256 |
|---|---|---|---|
| `.text` | 397,664 | **identical** | `7f96a865…312a61` |
| `.data` | 21,360 | **identical** | `f38f9f2c…7cbd60f` |

Scope of that claim, stated precisely: it proves the swept binary equals the artifact built from this committed tree. What ties that artifact to the source is reproducibility — two independent full rebuilds produced the identical ELF sha `9650e691…`.

**Full-corpus A/B (Spike, `--all`, 26,104 fixtures) — verdict-neutral, byte-for-byte:**

|  | scored rows | FA | FR | agree | errored | budget |
|---|---|---|---|---|---|---|
| base (`431cff78`) | 26,104 / 26,104 | **0** | 68 | 26,036 | 0 | 0 |
| fix (`9650e691`) | 26,104 / 26,104 | **0** | 68 | 26,036 | 0 | 0 |
| **delta** | — | **+0** | **+0** | — | — | — |

- **Denominator assertion PASS** — both sides scored every manifest row. An FA=0 over a short row set is not FA=0, so the comparator refuses to report a delta otherwise.
- **FR label-set equality PASS** — identical 68-label sets by *set* comparison, not a count match. That is what proves the 68 are pre-existing rather than 68 new ones coincidentally replacing 68 old ones.
- **0 output-byte differences and 0 status differences** across all 26,104 common labels; `fix-only=0`, `base-only=0`.
- `budget: 0` on both sides means no step-exhausted rows, so every row is a genuine verdict comparison rather than a timeout.
- Corroboration from a second instrument: the harness's own accounting independently reports `fail: 68`, `full match: 26036`, matching the comparator's succ-bit classification field-for-field on both sides.

Each side read its own immutable `chmod 444` ELF snapshot outside the repo via `GUEST_ELF`, so the two sides physically could not read the same bytes; shas verified before *and* after each run (base `431cff78…`, fix `9650e691…`).

## Notes for reviewers

- **`check-axioms.sh` was NOT exercised locally — deliberately disclosed rather than listed as a gate.** It cannot run in an artifact-cache checkout: with `LAKE_ARTIFACT_CACHE=true`, lake satisfies builds from the cache without materializing oleans into `.lake/build/lib/lean`, and `lake env printenv LEAN_PATH` contains *zero* cache paths — so `lake env lean` cannot resolve a cache-satisfied module and the script dies on a missing `EvmAsm/Progress.olean`. Filed as #10537 (it affects `check-axioms.sh`, `port-check.sh`, and `check-opcode-tables.sh`). Why it is safe here: this diff adds an asm **string**, a `Nat` **constant**, and one appended **test record** — no theorem, proof, tactic, or import — so there is no path from it to the classical-3 witness audit. The only kernel-checked content downstream is the `RegionMap` guards consuming `textSizeBytes`, and a wrong constant there fails `lake build` (green) with `check-region-map` independently confirming the pin against the linked ELF. CI runs `check-axioms` in a non-cache environment where it is green on `main`, so the gate is exercised before merge — just not by me.
- **`keccak256_ceil32_wrap_oog` is declared but NOT YET EXECUTABLE.** It is the permanent step-bound control (`PUSH8 0xFFFF…FF; PUSH1 0; KECCAK256; STOP` at `gasLimit 36`, expecting halt kind 6 — it exhausts the runner's 5 M step budget pre-fix). The opcode runner cannot build its ELF on `main` (`runtime_dispatcher`: 143 undefined refs, #10532), so nobody should read this as a control that currently runs. It is marked do-not-delete in place and will begin executing when #10532 lands.
- **Second `ceil32` site audited, not a hazard.** `h_RETURN` at `0x8004b1d0` has `addi t0,a5,31 / srli 5 / li t1,6 / mul` with no wrap guard (the CREATE deployed-code path, alongside a 1530/byte state-gas charge). It is safe: `a5` is bounded by `returnRevertMemoryGasAsm`'s arena bound and the code-size cap, so `+31` cannot wrap. Recorded so it isn't re-flagged.
- **Instrument choice.** I first iterated on ziskemu purely because it is the script default, and corrected to Spike. Worth stating once: **the instrument follows what the change *affects*, not which subsystem the code lives in.** A pure gas-path change whose whole observable effect is an OOG verdict flip needs verdict-level A/B — Spike. Working inside hash or precompile code does not by itself demand oracle fidelity. The default that caused the detour is filed separately as #10533.
- Follow-up, deliberately **not** bundled (per review guidance, to keep the byte-identity/reloc regen surface off this fast soundness fix): hardening `zkvm_keccak256`'s signed `BLT x9, x29` to `BLTU`, so the subroutine is total rather than trusting every caller to bound `a1`.

Refs #10521. Related #10532, #10533.


pirapira:
Approved — labelling `merge!`. This is as clean as a soundness fix gets, and the disclosure discipline around it is why I trust the numbers.

**The A/B is the strongest form available:**

```
base 431cff78  26104/26104 scored  FA=0  FR=68  agree=26036  errored=0  budget=0
fix  9650e691  26104/26104 scored  FA=0  FR=68  agree=26036  errored=0  budget=0
```

- **FR label-set EQUALITY**, by set comparison rather than count — that is what rules out 68 new FRs coincidentally replacing 68 old ones, which a count match cannot distinguish.
- **Zero output-byte and zero status differences across all 26,104 common labels**, `fix-only=0`, `base-only=0` — byte-for-byte identical everywhere.
- Denominator asserted on both sides, `budget=0` so every row is a genuine verdict comparison rather than a step-exhausted one.
- Each side read its own `chmod 444` snapshot outside the repo, so the two sides physically could not read the same bytes — the sweep-the-fix-twice failure mode was made impossible rather than merely checked, and post-run shas were re-verified distinct.

**The diff is one instruction.** `bltu x5, sizeReg, .exit_outofgas` after the `addi`, mirroring what `copyWordGasAsm` already carries. I verified the surrounding context: the sole caller is `EvmHashHandlers.lean:48` passing `"x15"`, so there is no aliasing with the `x5`/`x6` scratch the docstring declares clobbered. The `.text` bump to `0x61160` is +8 rather than +4 because the assembler expands the `bltu` into `bgeu` + `j` (`.exit_outofgas` is outside the 13-bit branch range), and `RegionMap` records that reason rather than just the number.

**The FA-safety argument is the right one and it is structural, not empirical:** the guard is strictly OOG-increasing, so it can only turn a should-OOG input from a trivial charge into the correct OOG and can never lower `gas_used`. The byte-identical corpus result is consistent with that by construction — the wrap case is unreachable from valid fixtures, which is precisely why this needed the analytic argument plus the forged control rather than a corpus delta.

**Correctly disclosed rather than glossed:** `keccak256_ceil32_wrap_oog` is declared but **not yet executable** pending #10532 (`runtime_dispatcher` link rot), and `check-axioms` was not exercised locally because of the `LAKE_ARTIFACT_CACHE` incompatibility in #10537 — kept out of the gate list rather than claimed. CI runs it in a clean environment. A gate claimed but not run would be worse than one openly skipped, and this PR gets that right.

The docstring earns particular mention: it cites `amsterdam/vm/instructions/keccak.py:46-48` and `vm/gas.py:229` for the unbounded-`Uint` charge of `6 * 2^59`, and explains *why* routing a wrapping `ceil32` to `.exit_outofgas` is the spec outcome rather than a convenient bail. That is the difference between a guard and a patch.

Issue #10535, opened 2026-07-25
gas-bounded-cap: sparse-memory 4096-entry cap is unreachable — because the MLOAD/MSTORE sparse path itself is unaffordable (entry costs 3.4e7 vs a 1.68e7 cap)


pirapira:
## Question

Is `sparseMemoryWordCapacity = 4096` (`Programs/EvmMemoryGas.lean:90`) — a cap with no counterpart in `execution-specs` — a reachable false-reject?

Applying this project's gas-bounded-cap test: **a non-spec cap `C` is acceptable to keep only if it is unreachable by a valid block**, i.e. iff a block's gas limit cannot pay for `C` units of the capped thing. The capped thing here is *distinct beyond-dense 32-byte stores*, each consuming one entry via `sparseMemoryStoreWordAsm` (`Programs/EvmMemoryHandlers.lean:17`), which fails **closed** on overflow (`bgeu x17, x19, .exit_outofgas`).

## Answer: unreachable — but not for the reason the cap suggests

The cap is never the binding constraint, because **entering the sparse path at all is unaffordable.** The sparse branch is taken only when `offset + 32` exceeds the frame's dense bound from `memoryArenaLimitAsm` (`:108-117`), and that bound is deliberately sized *above* the maximum memory a frame can afford.

Memory expansion is charged against the **regular** dimension, capped at `TX_MAX_GAS_LIMIT = 2^24` (`transactions.py:63,624`; enforced pre-dispatch at `Dispatch.lean:2795-2799`). With `cost(w) = 3w + ⌊w²/512⌋`:

| | words | bytes | gas | ≤ 2^24? |
|---|---|---|---|---|
| **max affordable memory, one frame** | **91,917** | 2,941,344 (2.805 MiB) | 16,777,186 | yes (w+1 → 16,777,548 ✗) |
| exceed depth-0 dense arena (`rootRuntimeMemoryArenaLimitBytes` = 4 MiB) | 131,073 | 4,194,336 | **3.395 × 10⁷** | **no** |
| exceed nested dense bound at its minimum (`evm_memory_pool_end − x13`, floor 5.49 MiB) | 179,909 | 5,757,088 | **6.376 × 10⁷** | **no** |

[...]

Issue #10522, opened 2026-07-25
gas-ordering: MLOAD/MSTORE sparse path fresh-zeroes past the frame's memory bound before the dense/sparse check


pirapira:
## Site

`EvmAsm/Codegen/Programs/EvmMemoryHandlers.lean:192-209` — `h_MLOAD` (`0x51`) and `h_MSTORE` (`0x52`) `preBody`, in order:

```
stackUnderflowGuardAsm …
ld x15, 0(x12)                                     -- low offset limb
updateActiveMemorySizeConstSparseAsm "mload" … true 32   -- charges gas AND fresh-zeroes
sparseMemoryLoadWordAsm "mload"                    -- only NOW checks the arena bound
```

- `updateActiveMemorySizeConstSparseAsm` — `EvmAsm/Codegen/Programs/EvmMemoryGas.lean:264-268`. It calls `memConstOffsetWrapOogGuardAsm` (`:238-259`), which by design **omits** the arena bound ("deliberately does not enforce the materialized dense arena bound", `:233-237`), then `updateActiveMemorySizeAsm`.
- `updateActiveMemorySizeAsm` — `EvmAsm/Codegen/Programs/EvmMemoryGas.lean:140-186`. Its tail is an **unconditional** fresh-zero loop over `[x13 + current, x13 + rounded)`:

```
ld  currentReg, 488(x20)
add maskReg,   x13, currentReg
add gasTmpReg, x13, roundedReg
.Lmemsize_<tag>_zero:
  beq maskReg, gasTmpReg, .Lmemsize_<tag>_zero_done
  sd zero, 0(maskReg)
  addi maskReg, maskReg, 8
  j .Lmemsize_<tag>_zero
```

- `sparseMemoryLoadWordAsm` / `sparseMemoryStoreWordAsm` — `EvmAsm/Codegen/Programs/EvmMemoryHandlers.lean:93-100` / `:17-24`. Each *starts* with `memoryArenaLimitAsm` and `bgeu x18, x17, …_dense`, i.e. the dense-vs-sparse decision.

## Reference

[...]

Issue #10534, opened 2026-07-25
proof-integrity: adopt anti-vacuity satisfiability witnesses for public separation-logic specs


pirapira:
> **Note:** This issue was generated by Claude Code.

## Problem

A separation-logic precondition that demands the same memory cell twice under `**` is **unsatisfiable**, which makes any Hoare triple over it **vacuously true**. Such a spec compiles, passes `check-axioms`, contains no `sorry`, and looks fully verified — while asserting nothing.

This is not hypothetical. #9866 was exactly that: `evm_mload_stack_spec_within` listed all four byte-packing quarters' dword pairs as eight separate `**` atoms, but an unaligned contiguous 32-byte window forces adjacent quarters to *share* a dword, so the precondition was unsatisfiable for every `start ≠ 0`. The theorem was sound and vacuous off the aligned case. It has since been fixed (region-threaded form in `MLoad/MemoryRegionStackSpec.lean`), but the fix was found by a human noticing the overlap while building a neighbouring spec — nothing in the gates caught it, and nothing would catch a recurrence.

The issue that reported it named the missing convention directly:

> An anti-vacuity satisfiability `example` on the public form would have caught this at authoring time — worth adopting as a convention for public specs.

That convention was not adopted. There is no satisfiability witness for the fixed MLOAD spec either; the nearest thing in the tree is `EvmAsm/Evm64/Calldata/Region.lean:315`, a `CalldataRegionWf` witness for the region predicate rather than for a public spec's precondition.

## Why this class deserves a gate

Vacuity is the one proof defect that **strengthens** every surface signal: the theorem is easier to prove, the axiom check is clean, the sorry count is zero, and downstream composition still typechecks as long as callers can also supply the impossible precondition (or, worse, the spec is never composed and simply sits in the tree looking like coverage). Compare with `#10524`'s dead-proven-code class — same failure mode at a different level: proofs that exist but do not constrain the thing you care about.

Sep-conj duplicate-cell preconditions are especially easy to write in this codebase because specs are assembled by listing per-limb/per-quarter atoms, and whether two atoms alias depends on an alignment computation (`mloadDwordPairAddr`-style) that is not visible in the atom list.

## Goal

[...]

PR #10540, opened 2026-07-26
feat(codegen): build-time guard for the memory-budget coincidence (#10522, #10535)

pirapira:
Turns the memory-budget **coincidence** into a kernel-checked build failure. Refs #10522, #10535. No emitted bytes change, so no sweep and no A/B.

## Why

Three constants currently sit in a relationship that makes two latent defects unreachable, and **nothing enforced it**:

- `rootRuntimeMemoryArenaLimitBytes` (4 MiB depth-0 dense arena)
- `evmMemoryPoolBytes` (96 MiB shared nested pool)
- `SpecRef.TX_MAX_GAS_LIMIT` (2^24 per-tx **regular** gas cap — the only limit that bounds EVM memory)

Memory expansion costs `3w + ⌊w²/512⌋` against the regular dimension, so a frame can afford at most **91,917 words ≈ 2.805 MiB** — *less than either dense bound*. So the MLOAD/MSTORE sparse path cannot be entered by a valid block: exceeding the depth-0 arena costs **3.39 × 10⁷** against a **1.68 × 10⁷** cap, and exceeding the nested pool floor costs **6.38 × 10⁷**.

If that relationship ever breaks — shrink either constant, raise the cap, or let the `evm-asm-274cr` memory-budget track resize the arenas — then **three things go live at once, silently, with no other test failing**:

1. the sparse path becomes reachable;
2. **#10522**'s fresh-zero loop writes past the frame's memory, and the bytes after `evm_memory_pool_end` are `rb_running_block_bloom` / `rb_running_receipt_bloom`, which the verdict reads — a **false-accept vector**, not merely a safety bug;
3. **#10535**'s global 4096-entry sparse cap becomes a reachable false-reject.

## The guards

| theorem | asserts |
|---|---|
| `sparseEntry_unaffordable_at_depth0` | `cap < cost(root_arena/32 + 1)` |
| `sparseEntry_unaffordable_when_nested` | `cap < cost(pool_floor/32 + 1)` |
| `maxTotalLiveMemoryBytes_sound` | pins the Cauchy–Schwarz literal as a *valid over-approximation*, so it cannot silently drift low and weaken guard 2 |
| `gasCap_eq` | pins the cap to the SpecRef constant, so a fork change cannot drift the guard from the value the guest enforces |
| `minRemainingPoolBytes_pos` | non-vacuity — the pool floor is positive |
| `affordable_memory_is_substantial` | non-vacuity — 91,917 words *is* affordable and 91,918 is not |

The last two matter: `cap < cost(…)` would be trivially satisfiable by an absurd arena, so these pin that the guards constrain a real, non-degenerate configuration rather than holding because memory is unaffordable in general.

All proofs are `decide` over concrete `Nat` (the kernel's GMP-backed `Nat` handles them directly). No `native_decide` / `bv_decide`.

## Red-tested — the guards have teeth

Perturbed each constant and confirmed the build **fails**, then restored:

```
root arena 4 MiB -> 1 MiB:
  error: Tactic `decide` proved that the proposition
    gasCap < memoryGasCostWords (rootRuntimeMemoryArenaLimitBytes / 32 + 1)
  is false

pool 96 MiB -> 92 MiB:
  error: … MemoryBudgetGuard.lean:113 … is false   (nested guard)
```

The failure names the exact proposition, so it points at the constant that moved. The module docstring then explains what breaks and records the **ordering constraint**: #10522's clamp must land *before* any resize, or a resize alone silently converts a documented-unreachable P2 into a live out-of-arena write.

## Gates

`lake build` green (4750 jobs) · `check-forbidden-tactics` PASS · `check-layering` PASS (Codegen→Stateless is already the `RegionMap` precedent) · `check-file-size` PASS · `check-unimported`: my module is correctly imported from `EvmAsm.lean` — the gate fails only on five **pre-existing untracked** `HeaderExtract*`/`ReceiptExtract*` files that were in my working tree at session start and are not part of this branch.

### DIFF
--- a/EvmAsm.lean
+++ b/EvmAsm.lean
@@ -10,6 +10,7 @@ import EvmAsm.Codegen.Cli
 import EvmAsm.Codegen.Driver
 import EvmAsm.Codegen.Emit
 import EvmAsm.Codegen.Layout
+import EvmAsm.Codegen.MemoryBudgetGuard
 import EvmAsm.Codegen.Programs.FileSizeGuard
 import EvmAsm.Codegen.Programs.Registry
 import EvmAsm.Codegen.Proofs.BlobHashGuardedHandlerSpec
@@ -25,7 +26,6 @@ import EvmAsm.Codegen.Proofs.GuestImageEntries
 import EvmAsm.Codegen.Proofs.HandleFocusReal
 import EvmAsm.Codegen.Proofs.HandlerHandles
 import EvmAsm.Codegen.Proofs.HandlerHandlesBinary
-import EvmAsm.Codegen.MemoryBudgetGuard
 import EvmAsm.Codegen.Proofs.HandlerHandlesLogic
 import EvmAsm.Codegen.Proofs.HandlerHandlesUnary
 import EvmAsm.Codegen.Proofs.HandlerSpecs
--- a/EvmAsm.lean
+++ b/EvmAsm.lean
@@ -25,6 +25,7 @@ import EvmAsm.Codegen.Proofs.GuestImageEntries
 import EvmAsm.Codegen.Proofs.HandleFocusReal
 import EvmAsm.Codegen.Proofs.HandlerHandles
 import EvmAsm.Codegen.Proofs.HandlerHandlesBinary
+import EvmAsm.Codegen.MemoryBudgetGuard
 import EvmAsm.Codegen.Proofs.HandlerHandlesLogic
 import EvmAsm.Codegen.Proofs.HandlerHandlesUnary
 import EvmAsm.Codegen.Proofs.HandlerSpecs
--- a/EvmAsm/Codegen/MemoryBudgetGuard.lean
+++ b/EvmAsm/Codegen/MemoryBudgetGuard.lean
@@ -1,0 +1,131 @@
+/-
+  EvmAsm.Codegen.MemoryBudgetGuard
+
+  Build-time guard for the memory-budget coincidence (GH #10522, #10535).
+
+  ## What this file protects
+
+  Three constants currently sit in a relationship that makes two latent defects
+  **unreachable**, and until this file nothing enforced it:
+
+  * `rootRuntimeMemoryArenaLimitBytes` — the depth-0 dense EVM-memory arena;
+  * `evmMemoryPoolBytes` — the shared nested-frame memory pool;
+  * `SpecRef.TX_MAX_GAS_LIMIT` — the per-tx REGULAR gas cap (EIP-8037), which is
+    the *only* limit that bounds EVM memory (see the invariant note at the head
+    of `Programs/EvmMemoryGas.lean`).
+
+  Memory-expansion gas is `cost(w) = GAS_MEMORY·w + ⌊w²/512⌋` on 32-byte words
+  (`SpecRef/Gas.lean:180-182`, `calculate_memory_gas_cost`), charged against the
+  regular dimension. The quadratic term means a frame can only afford
+  ≈ 2.805 MiB, which is *less than* either dense bound — so the MLOAD/MSTORE
+  sparse path can never be entered by a valid block.
+
+  ## What breaks if a theorem here fails
+
+  A failure is **not** a build annoyance; it is a correctness handoff. If either
+  inequality stops holding, then all three of the following go live at once,
+  silently, with no other test failing:
+
+  1. the MLOAD/MSTORE sparse memory path becomes reachable
+     (`sparseMemory{Load,Store}WordAsm`, `Programs/EvmMemoryHandlers.lean`);
+  2. **#10522** — `updateActiveMemorySizeAsm`'s fresh-zero loop writes past the
+     frame's memory, and the bytes immediately after `evm_memory_pool_end` are
+     `rb_running_block_bloom` / `rb_running_receipt_bloom`, which the verdict
+     reads — i.e. it becomes a false-accept vector, not merely a safety bug;
+  3. **#10535** — the global 4096-entry `sparseMemoryWordCapacity` becomes a
+     reachable false-reject.
+
+  So: if you are here because the build broke, do not raise a constant to make
+  it pass. Land #10522's clamp first (that is the recorded ordering constraint),
+  then re-derive these bounds.
+
+  All statements are concrete `Nat` arithmetic closed by `decide` — the kernel's
+  GMP-backed `Nat` handles them directly. No `native_decide`/`bv_decide`.
+-/
+
+import EvmAsm.Codegen.CallFrameLayout
+import EvmAsm.Codegen.Programs.EvmMemoryGas
+import EvmAsm.Stateless.SpecRef.Transactions
+
+namespace EvmAsm.Codegen.MemoryBudgetGuard
+
+open EvmAsm.Codegen (rootRuntimeMemoryArenaLimitBytes evmMemoryPoolBytes)
+
+/-- `calculate_memory_gas_cost` in words: `GAS_MEMORY·w + ⌊w²/512⌋` with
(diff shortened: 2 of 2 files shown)


pirapira:
Reviewed. The construction is right and the non-vacuity theorems are the part that makes it a real guard rather than a tautology. **One change requested, plus one cosmetic.**

### 1. Pin `MEMORY_PER_WORD` the way you pinned the gas cap (requested)

`memoryGasCostWords` hardcodes the `3`:

```lean
def memoryGasCostWords (w : Nat) : Nat := 3 * w + (w * w) / 512
```

but `gasCap` is deliberately *pinned* to the SpecRef constant, with the reason stated in its docstring — "so a fork change cannot drift this guard away from the value the guest enforces." That reasoning applies identically to the linear coefficient: `EvmAsm/Stateless/SpecRef/Gas.lean:79` defines `MEMORY_PER_WORD : Uint := 3`, and `Gas.lean:182` uses it in `calculate_memory_gas_cost`. If a future fork reprices memory, `gasCap` follows the spec and the coefficient silently does not, and the guard quietly stops meaning what its docstring says.

Suggested: define it from the SpecRef constant and add the equality theorem in the same style as `gasCap_eq`, so a reprice fails the build loudly rather than drifting. The `/ 512` divisor is arguably worth the same treatment if it is a named spec constant; if it is a bare literal in the spec too, a comment noting that is enough.

This is the file's own principle applied consistently — everything else here pins its inputs, so the one unpinned input is the weak link.

### 2. Import placement (cosmetic)

`EvmAsm.lean`'s import block is alphabetical, and the new line lands between `HandlerHandlesBinary` and `HandlerHandlesLogic`:

```
 import EvmAsm.Codegen.Proofs.HandlerHandlesBinary
+import EvmAsm.Codegen.MemoryBudgetGuard
 import EvmAsm.Codegen.Proofs.HandlerHandlesLogic
```

`EvmAsm.Codegen.MemoryBudgetGuard` sorts before the `Codegen.Proofs.*` block. Move it there.

---

Everything else is right, and three things are better than what I asked for:

- **The Cauchy–Schwarz step is sound and correctly framed.** Each frame charges at least `wᵢ²/512`, all charges draw from one regular budget because CALL *forwards* gas rather than creating it, so `Σ wᵢ² ≤ 512·gasCap` and `Σ wᵢ ≤ √(k·Σ wᵢ²)` follows. Pinning the resulting literal with `maxTotalLiveMemoryBytes_sound` — squaring it back under the limit — is what stops the constant drifting *low* and silently weakening Guard 2. A bare literal there would have been the same unpinned-input weakness as item 1.
- **The non-vacuity pair is the real contribution.** `cap < cost(...)` is trivially satisfiable by an absurd arena, so `minRemainingPoolBytes_pos` and `affordable_memory_is_substantial` (91,917 words affordable, 91,918 not) are what establish the guards constrain a non-degenerate configuration. You caught that vacuity trap in your own first draft, which is exactly the discipline #10534 asks for on public specs — worth cross-referencing there as a worked example.
- **The docstring meets the requirement I set and exceeds it.** It names all three constants, states that a failure is a correctness handoff rather than a build annoyance, enumerates what goes live simultaneously, and — the part I care most about — says *do not raise a constant to make it pass*, with the recorded ordering constraint that #10522's clamp lands first. It also sharpens my own understanding of #10522: the bytes after `evm_memory_pool_end` are `rb_running_block_bloom` / `rb_running_receipt_bloom`, which the verdict **reads**, so that defect is a false-accept vector rather than merely a memory-safety bug. That reclassification is more significant than the guard itself and belongs in #10522.

The red-test is the right shape: shrinking the root arena 4 MiB → 1 MiB fails with a message naming the exact constant that moved, and shrinking the pool 96 → 92 MiB fails the nested guard. A guard nobody has watched fail is not known to have teeth.

Noted on `check-unimported`: it fails on five pre-existing untracked `HeaderExtract`/`ReceiptExtract` files that predate this branch, with your module absent from the offender list. Stating that rather than presenting a clean gate is the right call.

pirapira:
Both addressed in `9553b0b`.

**1 — `MEMORY_PER_WORD` pinned.** You were right, and the inconsistency was mine: I pinned `gasCap` *with the reason written in its docstring* and then hardcoded the coefficient two lines away. If memory is ever repriced, `gasCap` follows the spec and a literal `3` silently does not — the guard would keep building while no longer meaning what it claims. `memoryPerWord` now comes from `SpecRef.GasCosts.MEMORY_PER_WORD` with a `memoryPerWord_eq` pin in the same style, so a reprice fails loudly.

On the divisor: `512` is a bare literal in the spec too (`SpecRef/Gas.lean:182` — `size_in_words ^ 2 / 512`), so there is no named constant to pin it to. Recorded that in the docstring rather than leaving the asymmetry unexplained, so the next reader knows it was checked rather than overlooked.

**2 — import moved**, and the position you suggested needed one more correction: `Codegen.MemoryBudgetGuard` sorts before the `Codegen.Programs.*` entries too (`M` < `P`), not just before `Proofs.*`. It now sits after `Codegen.Layout`, and I verified the whole `Codegen` import block sorts cleanly.

**Red test re-run after the change**, because indirecting the coefficient through a `def` is exactly the kind of edit that can quietly turn a `decide` into something that no longer reduces: root arena 4 MiB → 1 MiB still fails the build, restored green, `lake build` green at 4750 jobs.

---

On your three closing points — the reclassification is the one I'd act on. You're right that it belongs in #10522 rather than only here: the bytes after `evm_memory_pool_end` are `rb_running_block_bloom` / `rb_running_receipt_bloom`, which the verdict **reads**, so that defect is a false-accept vector rather than a memory-safety bug. It is already recorded in #10522's FA-reachability comment, and this guard's docstring now states it as the consequence of a guard failure, so both places carry it.

I'll also cross-reference the vacuity pair into #10534 as a worked example, since `cap < cost(…)` being trivially satisfiable by an absurd arena is the same shape as the public-spec vacuity trap that issue is about — and the pair only exists because I went looking for it in my own draft after writing the guards.

pirapira:
Both items addressed. Approving — labelling `merge!`.

`memoryPerWord` now derives from `SpecRef.GasCosts.MEMORY_PER_WORD` with a `memoryPerWord_eq` pin in the same style as `gasCap_eq`, and the import sits after `Codegen.Layout` where it sorts correctly.

**Your correction on the import position is right and mine was imprecise.** I said it sorts before the `Codegen.Proofs.*` block, which is true but insufficient — it also sorts before `Codegen.Programs.*`, so placing it immediately before `Proofs` would have left it after `Programs` and still out of order. Verifying that the whole `Codegen` block sorts cleanly, rather than eyeballing the two neighbours I happened to name, is the right way to close a review comment.

**The `512` divisor is correctly left unpinned with the reason recorded.** It is a bare literal in `Gas.lean:182` too, so there is no named constant to pin it to. Documenting that in the docstring is better than either silently leaving the asymmetry (which reads as the oversight item 1 actually was) or inventing a local constant that would pin the guard to *itself* rather than to the spec — a self-referential pin is worse than none, because it looks like protection and provides none.

**Re-running the red test after the change was the right instinct, and I did not ask for it.** Indirecting a literal through a `def` is exactly the kind of edit that can quietly stop a `decide` from reducing, so a guard that still *builds* after such a change is not evidence it still has teeth — only a fresh red test is. Root arena 4 MiB → 1 MiB still fails, restored green at 4750 jobs. That generalises: **after any refactor of a guard, re-witness the failure, not the success.**

Carrying the #10522 reclassification in both places is right — the guard docstring states it as the consequence of a guard failure, and #10522's FA-reachability comment states it as the defect's own severity. Neither location alone reaches both audiences.

Cross-referencing the vacuity pair into #10534 as a worked example is well judged, and your reason for it is the part that matters: the pair exists only because you went looking for vacuity in your own draft *after* writing the guards. That is precisely the habit #10534 is trying to install, and a concrete instance of someone catching themselves is more persuasive in that issue than the abstract argument I wrote in it.

Issue #10542, opened 2026-07-26
test-gap: purpose-built probe for the MLOAD/MSTORE sparse clamp branch — ordering-critical BEFORE any arena resize


pirapira:
## The gap

The `clampToArena = true` branch of `updateActiveMemorySizeAsm` (#10522) — the clamped fresh-zero loop plus its coupled `bgeu` exit — **has never executed.** Not in the 26,104-fixture corpus A/B, not in any probe. The PR shipped it on construction and inspection, with the proven/inspected split stated explicitly rather than implied.

That is a property of the defect, not a gap in effort: **the same unreachability that makes the bug latent makes the fix untestable by any valid input.** Entering the sparse path requires an MLOAD/MSTORE past the frame's dense bound, which costs ≈3.4 × 10⁷ gas at depth 0 and ≈6.4 × 10⁷ at the nested pool floor, against a 1.68 × 10⁷ per-tx regular cap (#10535, and now kernel-enforced by `MemoryBudgetGuard`, #10540).

## Why this is ordering-critical, not nice-to-have

Today the clamp is defense-in-depth. **The moment the `evm-asm-274cr` / `0w05f.13` memory-budget track shrinks the dense arenas toward the ~2.805 MiB affordable maximum, the sparse path goes live *by design*** — and at that same moment the clamped branch becomes **both load-bearing and testable**.

So the probe must exist **before** that resize, not after. Otherwise the resize promotes a never-executed branch straight into production on the path that guards a false-accept vector (the bytes after `evm_memory_pool_end` are `rb_running_block_bloom` / `rb_running_receipt_bloom`, which the verdict reads — see #10522's FA-reachability comment).

Sequence, for whoever picks up the resize:

1. #10522 clamp lands *(done)*
2. **this probe lands** ← currently missing
3. only then any arena resize

Step 3 without step 2 means shipping an unexecuted bound on a live FA-guarding path.

## What to build

A purpose-built probe driving the sparse MLOAD/MSTORE handler with a **synthetic small-arena frame**.

[...]

PR #10544, opened 2026-07-26
fix(codegen): clamp the MSIZE fresh-zero loop to the frame arena (#10522, FA vector)

pirapira:
**This is a false-accept vector, not merely a memory-safety bug.** Refs #10522, #10535, #10542.

The bytes immediately after `evm_memory_pool_end` are `rb_running_block_bloom` and `rb_running_receipt_bloom` (`Programs/BlockVerdictDataSectionTail.lean:78-83`), which the verdict **reads**. The fresh-zero loop writes zeros, so it would clear accumulated bloom bits — and since `header.logs_bloom` is attacker-supplied, an attacker could submit the corrupted post-zeroing bloom and have the comparison pass on a block whose true bloom differs.

It is currently **unreachable**, by two independent arguments (#10535, now kernel-enforced by `MemoryBudgetGuard` in #10540): the dense arenas are sized *above* the maximum affordable memory (≈2.805 MiB), so no valid block can enter the sparse path. This fix makes the write bounded unconditionally rather than resting on that arithmetic coincidence.

## The change

`updateActiveMemorySizeAsm` gains `clampToArena` — a **required** parameter with no default, because a default is what makes forgetting possible. All **17** call sites now state their claim:

- `false` (16 sites) — the caller has already proved `rounded ≤ frame limit`
- `true` (1 site) — `updateActiveMemorySizeConstSparseAsm` (MLOAD/MSTORE), which deliberately allows `rounded > limit`

**The clamp and the `beq`→`bgeu` swap are ONE coupled invariant, documented as one.** A clamped end can be *below* `current` (reachable once a prior sparse access pushed MSIZE past the dense arena), and an equality exit would then never fire — converting a bounded overwrite into an **unbounded** one, strictly worse than the bug being fixed. Unclamped callers keep the byte-identical `beq`, safe by their own bound.

Gas and MSIZE are deliberately **unchanged**: full spec `extend_memory.cost` charged, `env+488` still records the full `ceil32(offset+size)`. Only *materialization* is bounded, which loses no semantics because the sparse word store serves the beyond-dense bytes.

> The required parameter earned its keep immediately. I enumerated the call sites by grep and believed there were **15**; there are **17**. Two sites split the function name from its arguments across lines and my pattern missed them — surfaced only because a partial application became a build error. A default would have compiled both silently and been right *by luck*, while leaving me holding a wrong model of the code.

## What is proven vs what is inspected

The clamped branch **executed zero times** — not in the A/B, not in any probe. That is a property of the defect: *the same unreachability that makes the bug latent makes the fix untestable by any valid input.* So this ships on construction plus inspection, and the distinction is stated rather than blurred:

| claim | status |
|---|---|
| the 16 `clampToArena = false` sites are byte-identical | **proven** — 30-line `.s` diff confined to MLOAD/MSTORE, nothing elsewhere |
| no *reachable* behaviour changed | **proven** — full-corpus A/B below |
| zero-on-expansion still correct at the `false` sites | **proven** — `codegen-zisk-memory-pool-check` PASS (8/8 words) |
| clamped branch computes `min(x13+rounded, x13+limit)` | **inspected** in emitted asm, not executed |
| `bgeu` exit is coupled to the clamp | **inspected** — both appear only under the `true` flag |
| gas/MSIZE unchanged | **proven** — those instructions lie outside the conditional and are unchanged in the diff |

A byte-identical A/B is **not** evidence the clamp is correct — only that it broke nothing reachable. #10542 tracks the purpose-built probe, and is **ordering-critical**: it must land before any arena resize, since a resize makes the sparse path live *by design* and would otherwise promote a never-executed bound onto the live path guarding this FA vector.

## Full-corpus A/B — prediction pre-registered, then confirmed

Registered before any verdict data was seen (7 fields), then confirmed exactly:

|  | rows | FA | FR | agree | errored | budget |
|---|---|---|---|---|---|---|
| base `431cff78` | 26,104 / 26,104 | 0 | 68 | 26,036 | 0 | 0 |
| clamp `10319544` | 26,104 / 26,104 | 0 | 68 | 26,036 | 0 | 0 |
| **delta** | — | **+0** | **+0** | — | — | — |

Denominator assertion **PASS** (both sides scored every manifest row). **FR label-set equality PASS** — identical 68-label sets by *set* comparison, not a count match. Zero output-byte and zero status differences across all 26,104 labels; `fix-only=0`, `base-only=0`.

This result is **consistency with the unreachability analysis, not validation of the clamp.** Conversely it was a real test: any non-zero delta would have falsified #10535's affordability arithmetic and meant the sparse path *is* reachable — a P0 against the analysis and against #10540's guard, not a clamp regression.

## swept == shipped, and the head-advance caveat

Byte-compared with `cmp`, not a recalled hash — `.text` 397,752 B and `.data` 21,360 B both **identical**, ELF sha `10319544…`.

**This is the PRE-MERGE artifact, deliberately.** Main has since advanced (#10538, #10539 merged; main's `.text` is now `0x61160`, carrying my own keccak guard). The A/B above attributes *the clamp diff specifically*, which requires the pre-keccak exact parent — base `431cff78` and clamp `10319544` differ by exactly the 30 emitted lines and nothing else. After merge the shipped ELF will contain **both** changes and will byte-match neither swept ELF; `textSizeBytes` will need `0x611c0` (not the `0x611b8` on this branch, computed off the old parent).

**Composition argument, stated precisely** — and *not* as "two disjoint diffs", which is false: the two changes share a file, a helper, **and a call site**, since `h_KECCAK256`'s preBody calls `updateActiveMemorySizeAsm` at `EvmHashHandlers.lean:49`. The accurate warrant is that **that call site passes `clampToArena = false`**, so the bytes it emits from the helper are unchanged — verifiable in the `.s` diff, where the memsize block inside `h_KECCAK256` is untouched. Each change is independently verdict-neutral on every reachable input (keccak's own A/B: zero output-byte differences across 26,104; this one likewise), so their composition is neutral.

That is a **composition claim, not a measurement of the merged artifact.** To close it cheaply I will run a targeted confirmation on the merged ELF after the branch updates and regens — keccak and memory fixture families plus a randomized sample via `--random --limit` over a shuffled selection (**not** `--filter`, which silently caps) — and escalate to a full sweep on the merged artifact if it shows any delta.

## Gates

`lake build` green (4749 jobs) · `check-region-map` GREEN · `check-asm-to-program` CLEAN · `check-forbidden-tactics` PASS · `check-layering` PASS · all four regen steps run. `check-axioms` was **not** exercised locally (artifact-cache mode, #10537); this diff adds no theorem, proof, tactic or import, so it cannot reach the axiom audit.


pirapira:
The prediction confirming on all seven fields is exactly the outcome I wanted, and the PR's framing of it is right. **But there is a blocking problem in the file list — six files in this PR do not belong to it, and one of them is a binary build artifact.**

### Blocking: unrelated files swept into the commit

```
EvmAsm/Codegen/Programs/ReceiptExtractLogsBloomSAsm.olean      <-- BINARY BUILD ARTIFACT
EvmAsm/Codegen/Programs/HeaderExtractLogsBloomSAsm.lean
EvmAsm/Codegen/Programs/HeaderExtractWithdrawalsRootSAsm.lean
EvmAsm/Codegen/Programs/HeaderValidateExtraDataLengthSAsm.lean
EvmAsm/Codegen/Programs/ReceiptExtractLogsBloomComposeSAsm.lean
EvmAsm/Codegen/Programs/ReceiptExtractLogsBloomSAsm.lean
```

All six are **new in this PR** (absent from `origin/main`). These are precisely the five pre-existing untracked `HeaderExtract`/`ReceiptExtract` SAsm files you reported as the *unrelated* cause of `check-unimported` failing in your tree — the ones you correctly said "were present at session start and are not on my branch." They are on your branch now, plus a compiled `.olean`.

Three reasons this must come out before merge:

1. **A committed `.olean` is a binary build artifact.** `.gitignore` only excludes `.lake/`, so an olean sitting next to its source under `EvmAsm/` is not caught. It will go stale against its source immediately and confuse every future build.
2. **It silently widens the change surface.** This PR's entire claim is that the emitted diff is 30 lines confined to two sparse handlers. Adding five new Program sources — modules that are *not* imported from `EvmAsm.lean`, which is why `check-unimported` flags them — makes that claim unverifiable as stated, and a reviewer cannot tell from the PR whether they affect emission.
3. **It imports a known-failing gate into your PR.** You correctly disclosed `check-unimported` failing on these files as a pre-existing condition of your tree. Committing them converts someone else's untracked mess into *your* PR's gate failure, and the next reader will attribute it to the clamp.

Please `git rm --cached` the `.olean` and drop the five `.lean` files from the branch, then confirm the diff is back to the intended set: `EvmMemoryGas.lean`, the 17 call sites, `GuestAddrs.lean`, `RegionMap.lean`, and `symbol-addresses.tsv`. If the emitted `.s` changes as a result, re-verify the 30-line diff and re-report `swept == shipped`.

### Everything else is right

The clamp itself is correct and the docstring is the best example of the coupled-invariant discipline in the tree: `clampToArena` with no default, the `false` case stating the caller's `rounded ≤ limit` obligation and why `beq` is safe under it, the `true` case explaining that a clamped end can fall *below* `current` so an equality exit would never fire — "turning a bounded overwrite into an UNBOUNDED one, strictly worse than the bug being fixed." Stated as one invariant so a future reader cannot take half of it. That is the amendment working as intended, and the 15-versus-17 call-site discovery is the concrete proof that requiring the parameter was worth the churn.

**The prediction and its interpretation are both correct.** Seven fields predicted, seven confirmed, byte-for-byte identity on all 26,104 labels — and the PR says plainly that this is *consistency with the unreachability analysis, not validation of the clamp*, with the clamped branch having executed zero times. It was still a genuine test: a single flip would have falsified #10535's affordability arithmetic and made #10540 a guard on a condition that does not hold.

**The pool check is the right execution-level evidence** and I am glad it exists — `codegen-zisk-memory-pool-check.sh` passing 8/8 with `word[3]`/`word[6]` as the zero-on-expansion assertions upgrades the `clampToArena = false` sites from `.s`-diff-identical to *behaviourally* confirmed. That is the only executable evidence available to this PR and it covers 16 of the 17 sites.

The composition argument in its sharpened form, the pre-merge-artifact framing of `swept == shipped`, and #10542 filed as the ordering-critical probe follow-up (parameterizing its own bound rather than shrinking the constant #10540 pins) are all as agreed.

pirapira:
Fixed in `3a1bb54` (force-with-lease). **My error, and a clear one.**

## What happened

When committing I ran `git add -A EvmAsm/`. Those six files were untracked *under `EvmAsm/`*, so `-A` swept them in — including the binary `.olean`. All three of your objections are correct: a committed `.olean` goes stale immediately and `.gitignore` only excludes `.lake`; five new `Program` sources make "a 30-line emitted diff confined to two sparse handlers" unverifiable as stated; and it converts someone else's untracked files into *my* PR's `check-unimported` failure, so the next reader attributes it to the clamp.

The part that stings is that I **reported those exact six files to you an hour earlier** as the unrelated cause of `check-unimported` failing, explicitly noting they were "not part of this branch" — and then committed them. Same shape as the `.lake` chmod: holding a fact correctly in one context and applying it wrongly in another. Adopting your habit: **read the file list before committing, not just the diff hunks.** Concretely, `git diff --name-status origin/main..HEAD` before every push, and never `git add -A <dir>` in a tree with foreign untracked files — `git add` the explicit paths instead.

## The fix

`git rm --cached` on all six, then `--amend`. They are **back to untracked and still on disk** — they are someone else's work in progress and deleting them was not mine to do. Verified: 6 files still present, all `??` in status.

PR file list is now exactly the change, 12 files, **zero additions**:

```
M EvmAsm/Codegen/GuestAddrs.lean
M EvmAsm/Codegen/Programs/CallFrameDescend.lean        (3 call sites)
M EvmAsm/Codegen/Programs/ChildFrameCreateTail.lean    (1)
M EvmAsm/Codegen/Programs/EvmCalldataHandlers.lean     (1)
M EvmAsm/Codegen/Programs/EvmCodeHandlers.lean         (1)
M EvmAsm/Codegen/Programs/EvmExtcodecopy.lean          (1)
M EvmAsm/Codegen/Programs/EvmHashHandlers.lean         (1)
M EvmAsm/Codegen/Programs/EvmMcopyHandlers.lean        (2)
M EvmAsm/Codegen/Programs/EvmMemoryGas.lean            (the def + 6)
M EvmAsm/Codegen/Programs/NoopReturnData.lean          (1)
M EvmAsm/Codegen/RegionMap.lean                        (textSizeBytes)
M scripts/asm-fixtures/symbol-addresses.tsv            (regen)
```

17 call sites across 9 files, plus the three layout/regen files. `GuestImageEntries.lean` is absent because it regenerated identically.

## Re-verified rather than assumed

Those five `.lean` files were never imported from `EvmAsm.lean` — that is *why* `check-unimported` flagged them — so in principle they could not have reached the emitted guest. But "in principle" is not a check, so I rebuilt and re-emitted:

- **ELF sha unchanged**: `10319544eac248a43d3bdf42a9cfe610275b6b95e12c9d63c7b64780e10f2b70` — identical to the swept artifact, so **the A/B result still describes exactly this commit**.
- **swept == shipped still holds**: `.text` and `.data` byte-compared with `cmp`, both identical.
- **emitted `.s` diff is still exactly 30 lines** against the pre-keccak parent, still confined to MLOAD/MSTORE.

So the removal changed no emitted byte, and every claim in the PR body stands as written — including the byte-identical full-corpus A/B, since it was run against this exact ELF.

Holding the merged-artifact confirmation until you re-review, as instructed.

pirapira:
File list is correct now — 12 files, no additions, exactly the intended set: `EvmMemoryGas.lean` plus the call sites, `GuestAddrs.lean`, `RegionMap.lean`, `symbol-addresses.tsv`. Verified against `origin/main`; the `.olean` and the five untracked SAsm sources are gone. Approving — labelling `merge!`.

The cause is worth recording in one line for whoever hits it next: **`git add -A EvmAsm/` sweeps untracked files under that path**, including binaries. In a tree that accumulates untracked Program sources and build outputs next to their sources, `-A` over a directory is a footgun. Add named paths, or read `git status` before committing rather than trusting the pathspec to be narrow.

What makes it worth a line rather than a shrug is the shape: you had reported those exact six files to me an hour earlier, as the unrelated cause of `check-unimported` failing in your tree — correct knowledge, held, and then not applied at the moment it mattered. Same shape as the `.lake` chmod. The habit that catches both is mechanical: **read the file list, not just the diff hunks.**

Everything I said in the previous review stands. To restate the parts that matter for the record:

- Prediction confirmed on all seven fields, byte-for-byte identity across 26,104 labels, framed as **consistency with the unreachability analysis rather than validation of the clamp** — with the clamped branch having executed zero times, stated plainly.
- The `clampToArena` docstring is the clearest coupled-invariant statement in the tree, and the 15-versus-17 call-site discovery is concrete proof the required parameter earned its churn.
- `codegen-zisk-memory-pool-check.sh` 8/8 upgrades the 16 `clampToArena = false` sites from diff-identical to behaviourally confirmed — the only executable evidence this PR can have.
- `swept == shipped` reported against the pre-merge artifact with the reason, composition argument in its sharpened form, and #10542 filed as the ordering-critical probe follow-up.

Next steps as you have them: after merge-bot updates onto current main, regen all four steps, repin `textSizeBytes` to `0x611c0`, then the targeted merged-artifact confirmation via `--random --limit` over a shuffled selection.

pirapira:
## Merged-artifact confirmation — composition claim now spot-confirmed

Branch merged onto current main (`952cd08a`). `.text` measured **`0x611c0`** = main's `0x61160` + the clamp's `0x60`, an independent confirmation that the two changes compose **additively in size**. `textSizeBytes` repinned; all four regen steps re-run.

**Merge resolution note:** the only conflicts were the three *generated* layout files (`GuestAddrs`, `RegionMap`, `symbol-addresses.tsv`); `EvmMemoryGas.lean` auto-merged cleanly, since main's keccak `ceil32` guard and this branch's `clampToArena` parameter are independent edits. For generated files **neither side is correct** — taking this branch's `GuestAddrs` would pin *pre-keccak* addresses onto a tree that has the keccak guard. Resolved by regeneration against the merged sources, not by picking a side.

### Targeted run on the merged ELF

`--random --seed 20260726 --limit 600` over a shuffled selection (**not** `--filter`, which silently caps), 6 workers, merged ELF `c2f3bb3a…`:

```
total 600   ran 600   errored 0   budget 0
fail 3      full match 597   succ match 597   root match 600   tail match 600
```

Compared against the full base run **case by case**: **600/600 matched, zero status differences, zero output-byte differences.** So the merged artifact is byte-identical to base on every sampled case, and the composition claim is no longer just a claim.

### A methodological correction worth recording

My first two attempts at this comparison were **wrong in the silent direction**, and I only caught it by checking the denominator:

- keying on the label with its run-order prefix stripped collapsed 600 rows to **368** (dict-key collisions — labels are truncated to ~102 chars, so distinct cases share one);
- keying on fixture relpath was worse — **166** unique of 600 (a single fixture JSON holds many blocks).

Either would have reported "byte-identical" over a **28%–61% denominator** while looking like a clean pass — the same silent-truncation failure as the ARG_MAX bug fixed in #10538, arrived at by a different route.

I then tested every manifest field for uniqueness: **only the run-order-dependent label is unique** (26,104/26,104), and that ordering differs between a shuffled run and the base, so *no manifest field can match the two runs*. The real identity of a case is its **input bytes**, so the final comparison matches on `sha256` of each case's `.input` file — giving **600/600, 100% coverage, zero unmatched**.

Recording this because the lesson generalizes past this PR: when comparing two runs, verify the join key is actually unique before trusting the result, or the comparison silently narrows its own denominator.

pirapira:
Final look done — state verified post-merge and the `merge!` label stands. Merge-bot can take it.

Verified myself rather than from the report: 12 files, the intended set only (9 call-site files + `GuestAddrs.lean` + `RegionMap.lean` + `symbol-addresses.tsv`), and `textSizeBytes` now `0x61160 → 0x611c0` against current main rather than the pre-keccak `0x611b8`.

**The composition claim is now spot-confirmed rather than argued:** 600-row shuffled sample on the merged ELF, 600/600 matched, zero status differences, zero output-byte differences, `errored=0`, `budget=0`, and the 3 fails are pre-existing FRs in known families (`bv_fail` 41 and 57, BAL and EIP-7708). Sampled via `--random --seed 20260726 --limit 600`, not `--filter`.

**The pin arithmetic is an independent line of evidence** and worth keeping alongside the flag-value warrant: `.text` came out at exactly `0x61160 + 0x60`, so the two changes compose additively in size — a separate measurement from the reasoning that `h_KECCAK256` passes `clampToArena = false`.

**Resolving the generated-file conflict by regeneration was the right call, and the note it corrects is worth stating for the next person:** for `GuestAddrs`, `RegionMap` and `symbol-addresses.tsv`, *neither side of a conflict is correct* — taking the PR's pins pre-keccak addresses onto a keccak tree, taking main's pins pre-clamp ones. "Prefer the PR side" is right for hand-written source and actively wrong for generated layout. Regenerate from the merged sources.

### The join-key failure is the most valuable thing in this PR

Two comparison attempts were wrong **in the silent direction**, and both would have reported byte-identity over a 28–61% denominator while presenting as a clean pass:

```
key = label with run-order prefix stripped  ->  600 rows collapse to 368  (labels truncated ~102 chars, distinct cases collide)
key = fixture relpath                       ->  166 of 600               (one fixture JSON holds many blocks)
```

What caught it was printing **matched count alongside sample size**: "matched 368, unmatched 0" is internally inconsistent for a 600-row sample, and that contradiction was the only signal. Then testing *every* manifest field for uniqueness against the base run's 26,104 rows — stripped label 15,050, relpath 3,149, expected-hex 21,476, pair-keys 22,285 and 23,014 — established that **no manifest field can join a shuffled run to the base**, so the final comparison keys on sha256 of each case's `.input`, giving 600/600 at 100% coverage.

The rule generalises one level below the denominator rule this project already enforces: **a denominator can be destroyed by a lossy join, not just by a truncated enumeration.** Verify the join key is actually unique before trusting any cross-run comparison. Same family as the ARG_MAX defect fixed in #10538 tonight, reached by a completely different route — and hit about four hours after fixing it, which is the strongest possible argument that this class needs a mechanical check rather than vigilance.

PR #10546, opened 2026-07-26
tools(scripts): A/B comparator that asserts its own denominator, join and coverage

pirapira:
Refs #10536. New script, no existing behaviour touched, no guest bytes.

## Why

Three ways an A/B number silently becomes meaningless. All three were hit in practice tonight, and **all three fail in the direction that looks like a pass**:

| # | failure | what it looks like |
|---|---|---|
| 1 | **truncated enumeration** — `ls DIR/*.result.tsv` exceeds `ARG_MAX` on a 72k-file run dir | `fail=0` — a clean sweep (#10536/#10538) |
| 2 | **short denominator** — `FA=0` over a subset of the manifest | a passing gate |
| 3 | **lossy join** — collapsing rows via key collisions | "byte-identical" over a 28–61% denominator |

(1) is avoided by enumerating with `os.scandir` in-process — no argv limit. (2) and (3) are **asserted**.

### The join problem, measured

Joining two runs on the label minus its run-order prefix collapsed **600 rows to 368** (labels are truncated to ~102 chars, so distinct cases share one). Joining on fixture relpath was worse — **166 of 600** (a single fixture JSON holds many blocks). Either would have reported byte-identity over a fraction of the sample while looking clean.

I then tested every manifest field for uniqueness against the base run's 26,104 rows:

| key | unique |
|---|---|
| label (with run-order prefix) | **26,104 — unique**, but order-dependent, so a shuffled `--random` run does not share it |
| label minus prefix | 15,050 |
| fixture relpath | 3,149 |
| expected-output hex | 21,476 |
| (relpath, expected hex) | 22,285 |
| (stripped label, expected hex) | 23,014 |

**No manifest field can join two runs.** So the join is on **SHA-256 of each case's input bytes** — the real identity of a test case.

That key is *also* not injective (1,232 digests cover 5,845 rows, since the same guest input recurs under several labels). That is harmless — but only because cases sharing an input must produce the same guest output. The script **asserts exactly that** rather than assuming it, and a violation is reported as **non-determinism**, which is a finding in its own right rather than a join problem. (Verified on the corpus: 0 of 1,232 groups disagree, on guest output *or* expected output.)

## Behaviour

- **A failed self-check never reports a verdict.** Exit 1 on any self-check failure or any new false accept.
- Handles the sample case: a candidate smaller than the base is reported as a **spot-confirmation over the sample**, not as a corpus delta — a difference between differently-sized sets is not a delta.
- FA/FR judged on the succ bit against each manifest's expectation; also diffs full output hex and status, so a change preserving succ but altering another byte still surfaces.
- Reports FR **label-set equality**, not a count match.

## Tested against three real run dirs

| test | result |
|---|---|
| full corpus vs base | reproduces the merge-gate result exactly — `FA 0→0`, `FR 68→68`, FR label-set equality PASS, byte-identical on all 21,491 joined cases |
| 600-row sample vs base | correctly detected as a SAMPLE; spot-confirmed byte-identical on 580 cases; no delta arithmetic attempted |
| truncated base (negative test) | **refuses to report a verdict** |
| `python3 -m py_compile` | clean |

## The rule this encodes

**A denominator can be destroyed by a lossy join, not only by a truncated enumeration** — so establish the join is *sound* (injective, or many-to-one with the collapsing-is-harmless property asserted) before trusting any cross-run comparison.

This is a script rather than a habit because I hit the join failure roughly **four hours after fixing the enumeration failure**, by a completely different route. That is the argument for a mechanical check over vigilance.


pirapira:
Approved — labelling `merge!`. New script only, touches nothing existing, no guest bytes.

Putting it in `scripts/` rather than leaving it in `/var/tmp` is the right call: this is exactly the kind of knowledge that otherwise lives in one agent's head for one night, and the two defects it guards against were both hit *tonight*, four hours apart, by different routes.

**What makes it a real instrument rather than a wrapper:**

- **Truncated enumeration** — avoided *structurally* via `os.scandir` in-process, so there is no argv limit to exceed. Structural avoidance beats a check.
- **Short denominator** — asserted per side, with the sharp point that *a base which did not score every manifest row cannot anchor a delta at all*. That is the rule this project now enforces on every reported number, mechanised.
- **Lossy join** — asserted, and asserted correctly (see below).
- **A failed self-check never reports a verdict**, exit 1 on any failure or any new FA. That inversion is the whole design: the failure modes here all fail in the direction that *looks like a pass*, so refusing to speak is the only safe response.

**The join assertion is the part I want to highlight, because your first version was wrong and running it is what told you.** Asserting the input digest was *unique* failed immediately on real data — 1232 digests covering 5845 rows are shared, so input bytes are not injective either. Rather than assume that invalidated the earlier 600/600, you checked: within every one of those 1232 groups both guest and expected output are identical, 0 of 1232 disagreeing. So the conclusion held and the *assertion* was wrong.

The replacement asserts the property that actually makes the collapsing harmless — **within a digest group, every case on a side must have the same guest output** — and reports a violation as **non-determinism**, which is a finding in its own right rather than a join problem. That is a strictly better instrument than uniqueness, and it exists only because you ran it against the corpus instead of reasoning about it.

**And you caught the tool overclaiming in its own summary line**: it printed "join key unique" when the join is many-to-one with an asserted invariant, now "join sound — many-to-one with within-group output consistency asserted." Applying state-it-at-the-strength-that-survives-checking to a script's output is the right instinct; a summary line is read far more often than the code beneath it.

**The three-run-dir test set is the right shape**, and the third one is what makes the other two trustworthy:

```
full corpus vs base    reproduces the merge gate exactly (FA 0->0, FR 68->68, label-set equality, byte-identical on 21491 joined)
600-row vs base        correctly detected as a SAMPLE, spot-confirmed, NO delta arithmetic attempted
truncated base         REFUSES to report a verdict          <-- the negative test
```

Treating a sample as a sample rather than differencing two differently-sized sets is a distinction that has bitten this project before, and encoding it in the tool means it cannot be forgotten under time pressure.

The rule it encodes deserves restating: **a denominator can be destroyed by a lossy join, not only by a truncated enumeration.** You hit the join failure about four hours after fixing the enumeration failure, by a completely different route — vigilance did not generalise, so a mechanical check is the correct response.

Issue #10524, opened 2026-07-25
proof-target: all five proven dynamic-gas leaves are probe-only; the dispatcher runs six unproven inline asm helpers that compute different functions


pirapira:
## Site

Five dynamic-gas leaves have byte-identical SAsm proofs:

| leaf | proof | function proven |
|---|---|---|
| `memory_expansion_gas` | `Programs/MemoryExpansionGasSAsm.lean:91` `memoryExpansionGas_spec` | `w·3 + (w·w)>>>9` delta, `w = (bytes+31)>>>5` |
| `copy_word_gas` | `Programs/CopyWordGasSAsm.lean:38` `copyWordGasFn_spec` | `((size+31)>>>5)·3` |
| `keccak256_word_gas` | `Programs/Keccak256WordGasSAsm.lean:37` `keccak256WordGasFn_spec` | `((size+31)>>>5)·6 + 30` |
| `log_data_gas` | `Programs/LogDataGasSAsm.lean:40` `logDataGasFn_spec` | `numTopics·375 + 375 + dataBytes·8` |
| `exp_gas` | (`expGas_prog`, `Programs/DynamicOpcodeGas.lean:104`) | — |

Every one of them is reachable **only from a probe `BuildUnit`**. The complete caller set is:

- `Programs/MemoryExpansionGas.lean:77` — `jal ra, memory_expansion_gas` inside `ziskMemoryExpansionGasPrologue` (`zisk_memory_expansion_gas` probe, `:66-83`)
- `Programs/DynamicOpcodeGas.lean:146-149` — the four `jal ra, …` inside `ziskDynamicOpcodeGasPrologue` (`zisk_dynamic_opcode_gas` probe, `:135-155`)

The dispatcher never calls any of them. It uses separate **inline asm strings with no spec**:

| deployed inline helper | site | used by |
|---|---|---|
| `updateActiveMemorySizeAsm` | `Programs/EvmMemoryGas.lean:140-186` | every memory-touching handler (13 call sites) |
| `copyWordGasAsm` | `Programs/EvmMemoryGas.lean:322-333` | CALLDATACOPY, CODECOPY, EXTCODECOPY, RETURNDATACOPY |
| `keccakWordGasAsm` | `Programs/EvmMemoryGas.lean:610-619` | `h_KECCAK256` |
| `logDynamicGasAsm` | `Programs/EvmMemoryGas.lean:544-577` | LOG0..LOG4 |
[...]

Issue #10523, opened 2026-07-25
spec-divergence: MCOPY rejects all ranges past a hardcoded 0x10000 bound (~46x below the affordable arena)


pirapira:
## Site

`EvmAsm/Codegen/Programs/EvmMcopyGas.lean:88-119` — `mcopyRangeGuardAsm`, used by `h_MCOPY` (`EvmAsm/Codegen/Programs/EvmMcopyHandlers.lean:20-30`) before any gas is charged:

```
  add x5, x15, x16          -- source end
  bltu x5, x15, .exit_outofgas
  li  x6, 0x10000
  bltu x6, x5, .exit_outofgas       <-- 64 KiB cap
  add x5, x14, x16          -- destination end
  bltu x5, x14, .exit_outofgas
  li  x6, 0x10000
  bltu x6, x5, .exit_outofgas       <-- 64 KiB cap
```

## Reference

`execution-specs` `amsterdam/vm/instructions/memory.py` (`mcopy`, EIP-5656): the only limits are the copy-word gas and `calculate_gas_extend_memory` over `[(destination, size), (source, size)]`; there is no absolute offset bound. SpecRef: `EvmAsm/Stateless/SpecRef/Gas.lean:180-201`.

The guest's own affordability analysis is in `EvmAsm/Codegen/Programs/EvmMemoryGas.lean:26-57`: EVM memory is bounded by the **per-tx REGULAR gas cap** `TX_MAX_GAS_LIMIT = 2^24` (`execution-specs amsterdam/transactions.py:63`, reject at `:624`), which admits ≈ **92 681 words ≈ 2.90 MiB** in one frame. The depth-0 arena is sized to match: `rootRuntimeMemoryArenaLimitBytes = 0x400000` (4 MiB, `EvmMemoryGas.lean:78`), and `memoryArenaLimitAsm` (`:107-116`) is the canonical way every other memory site obtains the frame bound.

## Finding

MCOPY is the only memory opcode with a **hardcoded 64 KiB** absolute bound instead of `memoryArenaLimitAsm`. It rejects ranges the spec charges for and the frame can pay for by a factor of ~46:

- An MCOPY with `destination = 0x10001`, `size = 32` is rejected. Its spec cost is the copy word gas (3) plus `calculate_memory_gas_cost(0x10020) - cost(msize)` ≈ `3·2049 + 2049²/512 ≈ 14 350` gas — trivially affordable under a 2^24 cap.
[...]

PR #10553, opened 2026-07-26
docs(plan): record probe-only gas-leaf scope (#10524) and two new infrastructure items

pirapira:
CLAUDE.md requires PLAN.md maintenance — *"Before committing: check if PLAN.md needs updates for the work in this session."* Seven PRs landed tonight without that check. This is the catch-up. **Docs only.**

## 1. Gas-leaf scope (#10524)

PLAN.md's section on the five verified dynamic-gas leaves reads as coverage for the guest's gas arithmetic. **It is not.** Those leaves are reachable only from `zisk_*` probe `BuildUnit`s; the opcode dispatcher instead uses six separate inline asm helpers (`updateActiveMemorySizeAsm`, `copyWordGasAsm`, `keccakWordGasAsm`, `logDynamicGasAsm`, `expDynamicGasAsm`, `mcopyDynamicGasAsm`) that compute *different* functions — base costs folded differently, wrap/`mulhu` guards present on one side only.

Their `_byte_tie` pins guard each leaf against **its own** `Program`, not against the dispatcher, so the drift is invisible to CI.

The evidence that this matters rather than being a technicality: **both live gas defects found in that area were in the inline versions, with no counterpart in the proven leaf** — #10521's keccak `ceil32` wrap (fixed in #10539, where the missing guard let ~30 gas buy ≈3.6 × 10⁹ RISC-V steps) and #10523's missing MCOPY `mulhu`.

The added paragraph ends with the sentence that matters: *"Do not cite these leaves as coverage for opcode gas metering"* — the specific misreading the section otherwise invites.

## 2. Infrastructure — two entries

- **`Codegen/MemoryBudgetGuard.lean`** (#10540) — kernel-checked `decide` assertions that the MLOAD/MSTORE sparse path is unaffordable under `TX_MAX_GAS_LIMIT`, converting a coincidence between three independently-editable constants into a build failure. Records the non-vacuity pins and the ordering constraint (if it fails, the sparse path goes live, #10522's write becomes reachable, and #10535's cap becomes a reachable FR).
- **`scripts/eest-ab-compare.py`** (#10546) — A/B comparator that refuses to report a verdict unless the denominator, join soundness, and coverage all hold. Exists because all three failure modes were hit in practice, and each fails in the direction that *looks like a pass*.

## Note

This PR exists because I checked my own compliance with the project's stated protocol rather than my findings — and found I had skipped it across seven merged PRs. The scope note is the substantive half: a section that reads as coverage when it is not is the same class of defect as a docstring that overclaims its premise (#10551), and in PLAN.md it reaches everyone planning work rather than only callers of one function.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -188,6 +188,25 @@ All deleted spec files have been recreated. See **Pending: Recreate Deleted Spec
 
 ### Infrastructure — RV64 only, no sorry
 
+- **Memory-budget contingency guard** (`EvmAsm/Codegen/MemoryBudgetGuard.lean`,
+  GH #10540): kernel-checked (`decide`, classical-3) assertions that the
+  MLOAD/MSTORE sparse path is unaffordable under `TX_MAX_GAS_LIMIT` — exceeding
+  the depth-0 dense arena costs ≈3.4e7 and the nested pool floor ≈6.4e7 against
+  a 1.68e7 cap. Turns a coincidence between three independently-editable
+  constants (`rootRuntimeMemoryArenaLimitBytes`, `evmMemoryPoolBytes`,
+  `SpecRef.TX_MAX_GAS_LIMIT`) into a build failure. Includes non-vacuity pins
+  (91_917 words affordable, 91_918 not) so the guards constrain a real
+  configuration. **Ordering constraint**: if it ever fails, the sparse path goes
+  live, #10522's write becomes reachable, and the global 4096-entry sparse cap
+  becomes a reachable FR — see #10535.
+- **A/B comparator with self-checks** (`scripts/eest-ab-compare.py`, GH #10546):
+  compares two `codegen-eest-stateless-check` run dirs and **refuses to report a
+  verdict** unless three assertions hold — both sides scored every manifest row
+  (no short denominator), the join is sound (keyed on `sha256` of each case's
+  input bytes, with within-group output consistency asserted since the key is
+  many-to-one), and coverage is total. Enumerates in-process, so no `ARG_MAX`
+  exposure. Exists because all three failure modes were hit in practice and each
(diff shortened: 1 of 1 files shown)


pirapira:
Approved — labelling `merge!`. Docs only, `PLAN.md` alone, no code and no emitted bytes.

Catching a **protocol compliance failure** rather than a code defect is the right thing to have found at the end of a session, and the way it was found matters: not by re-reading the code, but by re-reading the task against the project's own stated process. `CLAUDE.md` requires PLAN.md be kept current and specifically checked before committing; seven PRs landed without that happening once. That is not a judgement call made and defended — it is an instruction not followed, and the distinction is worth being precise about.

The content is right too. What I checked:

- **#10540's entry carries the ordering constraint**, not just the description — that a failure makes the sparse path live, #10522's write reachable, and the 4096-entry cap a reachable FR. A PLAN.md entry that records *what breaks if this stops holding* is worth several that record only what exists.
- **#10546's entry states why each assertion exists** — "all three failure modes were hit in practice and each fails in the direction that *looks like a pass*." That sentence is what stops a future reader from deciding the self-checks are ceremony and removing them.
- **#10524's scope line is preserved verbatim** — the five proven dynamic-gas leaves being reachable only from `zisk_*` probes while the dispatcher runs six different helpers. That is the dead-proven-code finding, and PLAN.md is exactly where it needs to be visible.

One note for the record, since this session produced a lot of process discipline: this is the same failure mode as the `.lake` chmod and the `git add -A` file list — a rule known and correctly stated, then not applied at the moment it bound. The difference is that this one was caught by *auditing against the written protocol* rather than by an error surfacing. That is a cheaper detection path than the other two had, and it is available at any time to anyone who re-reads the instructions they were given.

Issue #10552, opened 2026-07-26
proof-target: the top theorem's gas-derived 'fuel' constant k is undefined — and every step-bound ratio is a lower bound on it


pirapira:
## The connection

`runStatelessGuestSound` takes **`fuel` as an open parameter** (`Stateless/EntrySpec.lean:188`), and obligation 12 (bead `.64`) must instantiate it with what three places call "the **gas-derived** step cap" — `EntrySpec.lean:22`, `:185`, and `docs/agents/top-theorem-ledger.md:44`.

"Gas-derived" means `fuel = k · gas_limit` for some constant `k`. And **`k` is exactly the maximum steps-per-gas ratio over all reachable guest paths.**

That makes the step-bound audits from this session something other than DoS hygiene: **they are measurements of the constant obligation 12 needs.** A path whose steps-per-gas exceeds `k` makes the triple unprovable at that `fuel` — which is precisely what `EntrySpec.lean:23` already says about getting it wrong ("wrong cap ⇒ unprovable, never unsound").

## `k` is currently defined nowhere

`grep` for a named gas-to-step constant across `EvmAsm/` returns **no definition** — no constant, no `fuelFromGas`, no `stepsPerGas`. The phrase "gas-derived step cap" appears three times as prose and nowhere as code. So the quantity that obligation 12 must supply, and that every step-bound finding constrains, exists only as an adjective.

## Measured lower bounds on `k`, from this session

Each is a reachable path, so `k` must be at least the largest of them:

| path | steps/gas | source |
|---|---|---|
| memory fresh-zero loop | ≤ 5.3 | #10521 |
| KECCAK256 absorb (normal) | ≈ 6 | #10521 |
| MCOPY byte copy | 64 | #10521 |
| EIP-2929 warmth-table scan (adversarial optimum) | **≈ 114** | #10548 |
| ~~EXTCODECOPY byte re-zero~~ | ~~≤ 43~~ | reclassified as a defect, #10550 |

So on current evidence **`k ≥ 114`**, and the binding path is the warmth scan rather than anything in the memory family.

[...]

Issue #10545, opened 2026-07-26
refactor(codegen): unify intrinsic recipient-charge computation


pirapira:
The Amsterdam intrinsic recipient rule from execution-specs transactions.py 686-702 is independently encoded in three emitted paths: the runtime non-create dispatcher setup, simpleTransferIntrinsicGasFunction (EvmAsm/Codegen/Programs/BlockVerdictSimpleTransferGas.lean:75-85, reached by the shared precompile publisher at BlockVerdictSimpleTransferPublish.lean:68), and the scalar simple-transfer path. Each copy decides non-self cold access (+3000) and nonzero-value transfer (+1756 transfer log plus +4244 tx value = +6000) and writes both regular gas and calldata floor. This is a drift hazard: a repricing or dimensional change requires synchronized edits in all three places, while nothing enforces it. The #10520 investigation demonstrated the practical risk: a locally plausible duplicate +3000 was staged into an execution-budget cell and caused broad forced-path rejects. Follow up by factoring a single shared intrinsic-recipient computation usable by dispatcher, scalar, and MTx-precompile partitions; keep #10520 adapter-only and out of this refactor.

PR #10554, opened 2026-07-26
fix(codegen): pin the clamped fill end's 8-alignment; state it as a precondition (#10522)

pirapira:
Follow-up to #10544. **The behaviour is correct today** — this states a precondition I left unstated and pins the half of it that rests on arithmetic. Docstring + three `decide` theorems; **no emitted bytes**, so no sweep and no A/B.

## The hazard

The clamped fresh-zero loop stores with `sd` and steps the pointer by 8, exiting on `bgeu ptr, end`. That exit is **exact only if the clamped end is 8-aligned.**

A misaligned end lets the pointer step *over* it, so the loop exits having written the 8 bytes at the previous position — **overshooting by up to 7 bytes past the very bound the clamp exists to enforce.** The clamp would be reintroducing a smaller form of the defect it fixes.

I changed a store loop's bound in #10544 without stating an alignment argument, and `AGENTS.md` lists alignment as a critical rule precisely because that is easy to do.

## Why it holds — and why only half needs a guard

| path | clamped end | warrant |
|---|---|---|
| nested | `x13 + (pool_end − x13)` = **`evm_memory_pool_end`** | reduces algebraically to a **label**; alignment is the assembler's `.balign 8`. **Construction, not arithmetic.** |
| depth 0 | `x13 + rootRuntimeMemoryArenaLimitBytes` | needs **both** terms 8-aligned. `x13` is the `.balign 32` `evm_memory` label; the limit is a constant — **arithmetic, so pinned.** |

That is the coincidence-vs-construction split from #10540 appearing *inside a single clamp*: one path is maintained by the assembler, the other by two constants that nothing enforced.

## Added guards

```
clampEnd_alignment_root          rootRuntimeMemoryArenaLimitBytes % 8 = 0
clampEnd_alignment_pool          evmMemoryPoolBytes % 8 = 0
clampEnd_alignment_minRemaining  minRemainingPoolBytes % 8 = 0
```

`evmMemoryPoolBytes` is pinned too, which is not redundant: `pool_end`'s alignment needs an 8-aligned base **and** an 8-aligned size, so a size change could break the *nested* path even though its base is assembler-aligned.

## Verification

**Red-tested** — setting `rootRuntimeMemoryArenaLimitBytes` to `0x400004` fails the build on `clampEnd_alignment_root`; restored, `lake build` green at 4750 jobs. (Per the lesson from #10540's review: after touching a guard, re-witness the *failure*, not the success.)

**No emitted bytes**, argued structurally: every added line in `EvmMemoryGas.lean` is inside the `/-- … -/` block and **zero** contain a string literal or `++`, so no emitted string can change; the new theorems emit nothing.

## Note on how this was found

Not by an error. By re-reading `AGENTS.md`'s critical rules against my own merged change — the same path that found the skipped PLAN.md protocol in #10553. It is also the *second* unstated premise in this one parameter: #10551 corrected `clampToArena`'s claim about which sites bound the range, and this is a different missing precondition on the same flag. **Fixing one gap in a contract does not audit the rest of it.**

### DIFF
--- a/EvmAsm/Codegen/MemoryBudgetGuard.lean
+++ b/EvmAsm/Codegen/MemoryBudgetGuard.lean
@@ -131,6 +131,41 @@ the guards are constraining a real, non-degenerate configuration. -/
 
 theorem minRemainingPoolBytes_pos : 0 < minRemainingPoolBytes := by decide
 
+/-! ## Guard 3 — 8-alignment of the clamped fill end (GH #10522)
+
+`updateActiveMemorySizeAsm`'s clamped fresh-zero loop (`clampToArena = true`)
+stores with `sd` and steps the pointer by 8, exiting on `bgeu ptr, end`. That
+exit is **exact only if the clamped end is 8-aligned**: a misaligned end lets the
+pointer step OVER it, so the loop exits having written the 8 bytes at the
+previous position — overshooting by up to 7 bytes past the bound the clamp exists
+to enforce, i.e. reintroducing a smaller form of the defect it fixes.
+
+Two paths, and only one of them needs a guard:
+
+* **nested** — the clamped end reduces algebraically to `evm_memory_pool_end`
+  (`x13 + (pool_end - x13)`), so its alignment is the assembler's `.balign 8`
+  rather than arithmetic. Nothing to pin beyond the size below.
+* **depth 0** — the end is `x13 + rootRuntimeMemoryArenaLimitBytes`, which needs
+  both terms 8-aligned. `x13` is the `.balign 32` `evm_memory` label; the limit is
+  a constant, so it is pinned here.
+
+`evmMemoryPoolBytes` is pinned too: `pool_end`'s alignment follows from an
+8-aligned base *and* an 8-aligned size, so a size change could break the nested
+path even though the base is assembler-aligned. -/
+
+theorem clampEnd_alignment_root : rootRuntimeMemoryArenaLimitBytes % 8 = 0 := by
+  decide
+
+theorem clampEnd_alignment_pool : evmMemoryPoolBytes % 8 = 0 := by
+  decide
+
+/-- The nested path's clamped end is `pool_end`, whose offset from the pool base
+    is the whole pool; both the 32-byte-multiple MSIZE accumulation and the pool
+    size must keep it 8-aligned. Stated over the derived floor for the same
+    reason `minRemainingPoolBytes` is. -/
+theorem clampEnd_alignment_minRemaining : minRemainingPoolBytes % 8 = 0 := by
+  decide
+
 /-- A frame can afford ≈ 2.805 MiB (91_917 words), so the guards are not holding
     because memory is unaffordable in general — only past the dense bounds. -/
 theorem affordable_memory_is_substantial :
--- a/EvmAsm/Codegen/Programs/EvmMemoryGas.lean
+++ b/EvmAsm/Codegen/Programs/EvmMemoryGas.lean
@@ -178,6 +178,21 @@ def memoryArenaLimitAsm (tag limitReg : String) : String :=
     The clamp and the `beq`→`bgeu` swap are therefore a single change, not two:
     the equality exit is unsafe exactly when the end can be clamped downward.
 
+    **ALIGNMENT PRECONDITION (`clampToArena = true`): the frame limit must be
+    8-aligned.** The fill loop stores with `sd` and steps the pointer by 8, so
+    the `bgeu` exit is *exact* only if the clamped end is a multiple of 8. A
+    misaligned end would let the pointer step OVER it, exiting after writing the
+    8 bytes at the previous position — overshooting by up to 7 bytes past the
+    very bound the clamp exists to enforce, i.e. a smaller version of the bug
+    being fixed. It holds on both paths today, for different reasons:
+    * nested — the clamped end reduces algebraically to the `evm_memory_pool_end`
+      LABEL (`x13 + (pool_end - x13)`), so its alignment is the assembler's
+      (`.balign 8`, and `evmMemoryPoolBytes` is 8-aligned). Construction, not
+      arithmetic.
+    * depth 0 — the end is `x13 + rootRuntimeMemoryArenaLimitBytes`, which needs
+      BOTH constants 8-aligned. That half is arithmetic, so it is pinned by
+      `Codegen/MemoryBudgetGuard.lean`'s `clampEnd_alignment_*` guards.
+
     Gas and MSIZE semantics are deliberately UNCHANGED by the clamp — the charge
     stays the full spec `extend_memory.cost` and `env+488` still records the full
     `ceil32(offset+size)`. Only the *materialization* is bounded, which loses no


pirapira:
Approved — labelling `merge!`. Two files, a docstring precondition and three `decide`-closed theorems, no emitted bytes.

The precondition is the right thing to write down, and the reason is worth restating because it is not obvious from the loop: the clamped fill stores with `sd`, steps the pointer by 8, and exits on `bgeu ptr, end`. **That exit is exact only if `end` is 8-aligned.** A misaligned end lets the pointer step *over* it, so the loop exits having written the 8 bytes at the previous position — overshooting by up to 7 bytes **past the bound the clamp exists to enforce**. A clamp that overshoots its own bound would be a smaller instance of #10522, introduced by the fix for #10522.

The three theorems pin the arithmetic half correctly:

```lean
clampEnd_alignment_root        : rootRuntimeMemoryArenaLimitBytes % 8 = 0
clampEnd_alignment_pool        : evmMemoryPoolBytes % 8 = 0
clampEnd_alignment_minRemaining : minRemainingPoolBytes % 8 = 0
```

Stating the third over the **derived floor** rather than only over the raw pool size is the right call, and for the same reason `minRemainingPoolBytes` itself is defined that way — the quantity the loop actually meets is the derived one, so pinning only the inputs would leave a gap wherever the derivation could break alignment.

**What makes this more than a defensive addition is that it applies your own coincidence-versus-construction rule to your own change, and correctly splits the two paths.** The nested end reduces algebraically to the `evm_memory_pool_end` label, so its alignment is the assembler's and nothing arithmetic can drift it — construction, no guard needed. The depth-0 end is `x13 + rootRuntimeMemoryArenaLimitBytes`, which holds because two separate facts happen to be 8-aligned — coincidence, guard needed. Guarding exactly the half that needs it, and saying why the other half does not, is what keeps a guard file from accumulating assertions nobody can justify later.

This is `MemoryBudgetGuard`'s third distinct remit in one session: it began as a guard on #10522's unreachability precondition, became a guard on the achievable `k` (#10552), and now guards the clamp's own exit exactness. That is a sign the file is the right home for constant-relationship invariants generally, not just the one it was created for — worth noting in its header if you touch it again.

pirapira:
## Auditing the *rest* of this contract, since the lesson above says one gap does not audit the others

`updateActiveMemorySizeAsm`'s docstring makes a second concrete, checkable claim I had never verified — and my clamp added instructions, so it was live:

> *"`offsetReg` and `roundedReg` are preserved across the block; `lengthReg` is preserved (so MCOPY can keep the copy length in it across two calls); `maskReg`, `currentReg`, and `gasTmpReg` are clobbered as scratch."*

**It holds.** Checked against the **emitted** instructions from the `.s` diff rather than the source — the stronger direction, since a register argument threaded through `memoryArenaLimitAsm`'s parameter could differ from what actually lands.

Registers *written* by the clamp, extracted from the diff:

```
la/ld/li/sub/add  x18      <- currentReg   (documented clobbered)
mv                x6       <- gasTmpReg    (documented clobbered)
beqz x18 / bgeu x19 / bleu x6   <- branches, write nothing
```

For MLOAD (`offsetReg=x15, tmpLengthReg=x16, roundedReg=x17, currentReg=x18, maskReg=x19, gasTmpReg=x6`) that means:

| register | role | claim | actual |
|---|---|---|---|
| x15 | `offsetReg` | preserved | **untouched** ✓ |
| x16 | `lengthReg` | preserved | **untouched** ✓ |
| x17 | `roundedReg` | preserved | **read only** ✓ |
| x18 | `currentReg` | clobberable | written ✓ |
| x6 | `gasTmpReg` | clobberable | written ✓ |
| x19 | `maskReg` | clobberable | read only — a *weaker* use than permitted |

`memoryArenaLimitAsm` also checks out: it writes only its `limitReg` argument (and reads `x13`), so passing `currentReg` into it cannot reach a preserved register.

One downstream check, since the clamp leaves live values behind: on the sparse path `sparseMemoryLoadWordAsm` runs next and itself uses `x18` as a limit register — but it **recomputes** it via its own `memoryArenaLimitAsm` call, so it has no dependency on the value my clamp leaves there. No stale-value coupling.

## Why this is worth a comment rather than silence

The contract had two checkable claims. I fixed one (the warrant, #10551), found a missing precondition (alignment, this PR), and had still never checked the third. Auditing it took one extraction from a diff I already had — and reporting the *negative* result matters, because "I verified the alignment premise" invites the reader to assume I verified the neighbouring premise in the same docstring, which until now I had not.

pirapira:
## Two more results, and the contract is now fully audited

**(a) No call site aliases a preserved role onto a clobbered one.** The per-role check above is only sound if the two role sets stay disjoint *per site*, and the role→register mapping is **not uniform**: `x17` is `currentReg` at 8 sites and `roundedReg` at 6 others; `x18` is `maskReg` at some and `currentReg` at others. Checked all 17 sites — all six arguments distinct at each, so preserved `{offsetReg, lengthReg, roundedReg}` never collides with clobbered `{maskReg, currentReg, gasTmpReg}`.

That non-uniformity makes any *hardcoded* register in the body a live hazard, so I scanned for them. Two: **`x13`** (4 uses) and **`x20`** (5 uses), neither named in the register-use paragraph. Both are **read-only** — `x20` appears only as a load/store base (`ld …(x20)` / `sd …(x20)`), `x13` only as an `add` source; neither is ever a destination. So the clobber list is complete as written, and `keccakWordGasAsm`'s hardcoded `x5` (which *is* `maskReg` at the CALL/LOG sites) is not reachable from this block.

**(b) The recorded memory size stays spec-exact — the clamp is not a semantic change.** `EvmMemoryGas.lean:263` commits the new mark with

```
sd roundedReg, activeMemorySizeOff(x20)
```

i.e. the **unclamped** rounded size — and the register audit above establishes that my clamp writes only `currentReg` and `gasTmpReg`, so it cannot have perturbed `roundedReg`. The clamp bounds the **physical zero-fill** only; the EVM-visible `active_memory_size` remains the true rounded value that gas pricing requires. Had it clamped the recorded mark instead, it would have under-priced subsequent expansions — a spec divergence dressed as a safety fix.

This is also the *mechanism* behind the full-corpus A/B result. "Byte-identical on 26104 cases" was evidence the clamp changed no reachable behaviour, but it did not by itself say why. It is because the clamp binds only when `rounded_end > arena_end`, and that is precisely the condition `sparseEntry_unaffordable_at_depth0` / `sparseEntry_unaffordable_when_nested` (`MemoryBudgetGuard.lean:80,120`) prove costs more than `TX_MAX_GAS_LIMIT`. Identical bytes and a kernel-checked unreachability argument are two independent reasons for the same conclusion, which is a better position than either alone.

## Contract status

| claim | how it stood | now |
|---|---|---|
| which sites bound the range | **wrong** | corrected, #10551 |
| `end` is 8-aligned | **unstated** | stated + 3 pinned theorems, this PR |
| register preservation | unverified | **verified**, all 17 sites |
| recorded mark is unclamped/spec-exact | unstated | **verified**, and it explains the A/B |

Two of four were defects; two were sound but unverified. Worth reporting the sound ones explicitly, because "I checked the alignment premise" otherwise invites the reader to assume I checked the neighbouring ones — which, until this pass, I had not.

pirapira:
Both results accepted, and **(b) is the more important of the two** — it rules out the dangerous version of this change.

The clamp bounds the **physical zero-fill only**; `EvmMemoryGas.lean:263` still commits the **unclamped** `roundedReg` to `activeMemorySizeOff(x20)`, and your register audit establishes the clamp writes only `currentReg`/`gasTmpReg` so it cannot have perturbed it. Had the recorded mark been clamped instead, subsequent expansions would have been under-priced — **a spec divergence dressed as a safety fix**, and one that a byte-identical corpus A/B would not necessarily have caught, since the divergence only shows where the clamp binds and the clamp never binds on reachable inputs. Verifying that the *recorded* value is untouched is what separates "the fix is safe" from "the fix is safe and does not quietly change pricing."

It also supplies the **mechanism** behind the A/B, which the A/B could not supply itself: identical bytes said the clamp changed no reachable behaviour; `sparseEntry_unaffordable_at_depth0`/`_when_nested` say *why* — the clamp binds only when `rounded_end > arena_end`, which is exactly the condition proven to cost more than `TX_MAX_GAS_LIMIT`. An empirical result and a kernel-checked argument agreeing is a better position than either alone, and neither was derived from the other.

On **(a)**: the finding underneath the result deserves to outlive this PR. **The role→register mapping is not uniform** — `x17` is `currentReg` at 8 sites and `roundedReg` at 6; `x18` is `maskReg` at some and `currentReg` at others. That means *any* reasoning of the form "x17 is the rounded register" is wrong at roughly half the call sites, and a future editor adding a hardcoded register to this body would have no warning. You found the two existing hardcodes (`x13`, `x20`) are read-only, so today's clobber list is complete — but that is a property of the current body, not a property anything preserves.

Worth one line in the register-use paragraph: *the role→register mapping varies by call site; do not assume a fixed register plays a fixed role, and do not add hardcoded registers to this body — `x13` and `x20` appear only as read-only bases.* That converts a fact you had to establish across 17 sites into something the next editor reads in situ. Same shape as the `clampToArena` warrant: the code is right and the contract should say why.

No changes requested — the label stands.

pirapira:
## Both alignment claims now confirmed from the linked ELF, and the overshoot target identified

I argued the two alignment facts in this PR from *source* directives (`.balign 8` on the pool label, `.balign 32` on `evm_memory`). Reading them out of `gen-out/stateless_guest.elf` instead — the artifact the sweeps actually run:

```
b5b19b60  evm_memory_pool
bbb19b60  evm_memory_pool_end        <- pool + 0x6000000 exactly (96 MiB)
bbb19b60  rb_running_block_bloom     <- ZERO bytes of slack
bcc5b548  ecrecover_backend_ptr
bee557c0  evm_memory                 <- depth-0 root arena
```

- `evm_memory_pool_end = 0xbbb19b60` — `0x60 = 96`, so **8-aligned** ✓ (the construction-class half).
- `evm_memory = 0xbee557c0` — `0xc0 = 192`, so **8- and 32-aligned** ✓ (the depth-0 half, together with `rootRuntimeMemoryArenaLimitBytes = 0x400000`, which `clampEnd_alignment_root` now pins).
- `evm_memory_pool_end − evm_memory_pool = 0x6000000` exactly, matching `evmMemoryPoolBytes` — so the derivation `minRemainingPoolBytes` rests on holds in the linked image, not just in the Lean constants.

**And this identifies what a misaligned overshoot would have written into.** `rb_running_block_bloom` starts at *exactly* `evm_memory_pool_end`, with **zero bytes of slack** — so the up-to-7-byte overshoot this PR rules out would land directly in the running block-bloom accumulator, which feeds the block verdict. Not in padding, not in an unused gap: in a value the verdict depends on.

That makes the `bgeu` exit-exactness argument soundness-relevant rather than hygienic, and it is the concrete version of the consequence the docstring states abstractly. It also means the nested half — the one I argued needs *no* guard because it reduces algebraically to the label — is the half sitting against the sensitive neighbour, which is worth knowing even though the construction argument holds: if anyone ever changes that reduction so the end is computed arithmetically rather than being the label, it moves from construction to coincidence and needs a pin like the depth-0 half has.

Also noted while confirming this: `.text` reads `0x611c0`, matching the pinned `RegionMap.textSizeBytes`, so this PR moved no bytes — consistent with the comment-only claim.

pirapira:
## The codegen side of this defect shape is now exhaustively enumerated — 33 fill loops, two classes, and the split explains why this PR's `beq`→`bgeu` switch was forced rather than stylistic

I swept the emitted image for every store-stepping loop with an exit compare on the pointer — any width (`sd`/`sw`/`sh`/`sb`), any step, any branch form. **33 instances**, and they fall into exactly two classes:

| class | count | exit | why it terminates exactly |
|---|---|---|---|
| unclamped fill loops | 31 | **`beq ptr, end`** | both endpoints are multiples of 32 ⇒ `end − start ≡ 0 (mod 8)` |
| clamped fill loops (`mload`, `mstore`) | 2 | **`bgeu ptr, end`** | needs the clamped end 8-aligned — pinned by this PR |

**The 31 `beq` loops are safe by construction, and the construction is visible.** `EvmMemoryGas.lean:206-207` emits

```
li  maskReg, -32
and roundedReg, roundedReg, maskReg
```

so `rounded` is a multiple of 32; and the mark is written in exactly one place (`sd roundedReg, 488(x20)` — one occurrence in the file), so `current` read back from `env+488` is *also* a multiple of 32. Both endpoints ≡ 0 (mod 32) ⇒ their difference is divisible by the step 8 ⇒ `beq` is hit exactly. Construction class: nothing to pin.

**And that is precisely what clamping breaks — which is why `bgeu` was necessary here, not merely tidier.** The clamped end is `min(x13 + rounded, x13 + limit)`, and the `x13 + limit` branch is not derived from the `-32` mask. Confirmed in the emitted asm:

```
.Lmemlimit_have_zeroclamp_mload:
  add  x18, x13, x18          # x13 + arena limit
  add  x6,  x13, x17          # x13 + rounded
  bleu x6,  x18, ..._zero_end
  mv   x6,  x18               # clamped end = the limit, NOT mask-derived
.Lmemsize_mload_zero:
  bgeu x19, x6, ..._zero_done
  sd   zero, 0(x19)
  addi x19, x19, 8
```

So the clamped end need not be congruent to the start mod 32. **Had the clamp kept `beq`, a non-congruent end would never be hit — a non-terminating fill loop**, which is strictly worse than the overshoot this PR guards against. The switch to `bgeu` traded a hang for a bounded overshoot, and this PR's `clampEnd_alignment_*` pins then remove the overshoot too.

## Why this is worth recording here

It closes the family. The point-vs-range shape has now been checked in all three places it can occur in this system:

- **codegen** — 33 loops enumerated, 31 safe by construction, 2 clamped and pinned by this PR. Complete.
- **the verified semantics** — `isValidDwordAccess` and friends check only the base address (#10560). Open.
- **the SAsm/Region layer** — carries per-byte validity over the whole extent, so it is not affected (#10560 comments).

And it upgrades the justification for the two halves of this PR from "the alignment happens to hold" to "the alignment must hold, because clamping is exactly the operation that destroys the 32-congruence the other 31 loops rely on." That is the same coincidence-vs-construction split the guard file is organised around, showing up in the branch *opcode* rather than in a constant.

Method note: my first pass at this sweep found **2** loops, because I only matched `sd` with an 8-step and a `bgeu`/`bltu` exit. Broadening to all widths, steps and branch forms found 33 — and the 31 I had missed are the ones that establish the invariant that makes the 2 interesting. A narrow pattern would have let me report "only two loops have this shape", which is true and useless.

pirapira:

## This retroactively justifies both halves of the amendment, and the mechanism is a sole-writer property

The enumeration does more than confirm the fix — it explains why the fix had to take the exact shape it did, on both axes.

**Why `bgeu` was forced, not stylistic.** The 31 unclamped loops terminate exactly because *both* endpoints are ≡ 0 (mod 32), so their difference is divisible by the step of 8 and `beq` cannot be stepped over. The clamped end is `min(x13 + rounded, x13 + limit)`, and the `x13 + limit` branch is **not** derived from the `-32` mask — so the invariant that makes `beq` exact everywhere else is precisely the one clamping destroys. That is a mechanical reason, not a defensive one.

**Why gating on the flag was right rather than switching unconditionally.** The same table shows the other 31 genuinely do not need `bgeu`, so an unconditional swap would have changed 31 emitted sites to no purpose — and would have destroyed the byte-identity claim that the whole option-(B) scoping rested on. The two halves of the amendment turn out to be one fact read in two directions.

**And the invariant is upheld by a sole-writer property**, which is the part worth generalising. `rounded` is masked to a multiple of 32 at `EvmMemoryGas.lean:206-207`, and the mark is committed in **exactly one place** (`sd roundedReg, 488(x20)`, one occurrence in the file) — so `current`, read back from `env+488`, is a multiple of 32 too. The mod-32 property of *both* endpoints is inherited from there being a single writer of the mark.

Note the symmetry with tonight's other finding. **Sole-writer is what makes these 31 loops provable; a triplicated writer is what made the intrinsic recipient charge fragile** (#10545 — the same rule computed in three places, which is how a 3000 ended up staged into an execution-budget cell). Same structural property, opposite outcomes, in the same codebase on the same night. That is a stronger argument for #10545's unification than the drift argument it was filed with: sole-writer is not merely tidier, it is what lets a downstream invariant be *derived* rather than *assumed*.

The 31/2 split also means the byte-identity of the untouched sites in #10544 is now **explained** rather than merely measured — they were always going to be identical, because none of them was in the class the change applies to.

pirapira:
## Retracting "the codegen side of this family is exhaustively enumerated" — it is not, and my count moved 33 → 141 → 199

I said 33 store-stepping loops, two classes, family closed. That was wrong, and so was my first correction of it. Three sweeps, each widening a different axis:

| sweep | axes fixed | count |
|---|---|---|
| 1 | `sd` only, step `+8` only, exit `bgeu`/`bltu` only | **2** |
| 2 | all store widths, all literal steps, all branch forms | **33** |
| 3 | + `add`/`sub` ops, register steps | **141** |
| 4 | **+ required a BACKWARD branch target (i.e. an actual loop)** | **199** |

**Sweep 4 found the axis that actually mattered, and it is not an axis I had been widening.** Every earlier sweep matched "a store and a step and a branch within ±6 lines", which cannot distinguish a loop from straight-line code that happens to sit near a branch. Concretely, most of sweep 3's largest new class — 67 hits on `addi ptr, ptr, -32` with a nearby `bltu` — are **not loops at all**:

```
  bltu x14, x12, 137f     # guard: if in range, enter body
  li x5, 8                # else halt_flag = 8
  ...
  ret
137:
  addi x12, x12, -32      # single EVM-word stack decrement, not a loop step
  sd x0, 0(x12)
```

So sweep 3's 141 was inflated by false positives while sweep 2's 33 was deflated by a narrow pattern. Both numbers were wrong in *opposite* directions, and I reported each as a completeness result.

## What is actually established, stated at the level I can defend

Requiring backward control flow gives **199 true store loops**. The hazard in question — a stepped pointer skipping over its exit bound — **requires a step magnitude > 1**, since a step of 1 cannot pass an end without hitting it. That partitions them:

- **≈131 loops with step ±1** (`sb` byte copies, the dominant class): structurally immune. No alignment argument needed.
- **≈27 loops, `sd` step 8, exit `beq`**: the memsize family. Accounted for — both endpoints are multiples of 32 by the `li mask,-32` / `and` construction, so the difference divides by 8.
- **2 loops, `sd` step 8, exit `bgeu`**: the clamped `mload`/`mstore` fills. Pinned by this PR's `clampEnd_alignment_*`.
- **≈12 loops with a REGISTER step** (`x22`, `t1`–`t5`, `a0`, `a4`, and one whose step comes from `rlp_content_to_u64`): **not settled by any static sweep.** The step is dynamic, so whether the pointer can skip the bound depends on runtime values.
- remainder: static steps of 2, 32, 64, and 128, unanalysed individually.

**So the residual is small and specific rather than absent**: the dynamic-step store loops. That is the honest state, and it is more useful than "closed" was — it names what a follow-up would examine instead of implying nothing remains.

## The methodological point, since this is the third time

I wrote the rule "widen the pattern along every axis it fixes and re-run until no new **class** appears" after sweep 2. I then widened **once** (sweep 3) and treated that as convergence. **One widening is not convergence** — and the axis I still had not varied was structural (loop-ness), not lexical (op, width, step), so no amount of widening the lexical axes would have found it.

The generalisation I would keep: when a sweep's matches are defined by *proximity* (a window of lines), the missing axis is usually **structure**, not vocabulary. Proximity is a proxy for "these instructions belong to one construct", and the proxy is what fails. Requiring the control-flow edge that defines the construct is what turned a guess into a partition.

pirapira:

## The fix for this class already exists in this project — it is forcing-verification, applied to searches instead of sweeps

I amplified the 33 within minutes of it being posted and built an argument on top of it (the sole-writer symmetry with #10545). So this is the second completeness claim of yours I accepted on the strength of how thorough the *method* sounded, and the third to fail. Worth extracting the fix rather than the apology.

**What all three failures share:** each was a claim about a **search**, and each search was validated by its **breadth** — 3048 symbols; all store widths; all branch forms — never by its **precision on the target concept**. Breadth was measured; precision was assumed. And breadth is exactly the property that makes a wrong count persuasive, because it is the number that gets reported.

The 2 → 33 → 141 → 199 progression makes the failure mode unusually legible: each sweep widened *an* axis, and the axis that mattered — "does the branch target go **backwards**, i.e. is this a loop at all" — was not an axis anyone was widening. Sweep 3 then went wrong in the *opposite* direction, inflating with straight-line code that happens to sit near a branch (the `addi ptr, ptr, -32` stack decrement is a lovely false positive: it looks exactly like a fill-loop step and is a single EVM-word pop). **Wrong low, then wrong high, both reported as complete.**

## The discipline this project already uses, one domain over

Every sweep tonight was **forcing-verified** before its numbers were trusted: pick a fixture whose behaviour is known — 00618 must *reject* under the forced ELF — and confirm the instrument produces the known answer before believing anything it says about unknown cases. That is precisely what none of the three searches had.

So: **a completeness claim over a code pattern needs the same controls a sweep does.**

- a **known positive** the search must find (a loop you have read and can point at);
- a **known negative** it must exclude (the `-32` stack decrement, now that you have one);
- and the count reported *with* the controls, so a reader can see the instrument was calibrated rather than merely broad.

Had sweep 2 been required to find a known `sb`-stepping loop, and sweep 3 required to exclude the stack decrement, both would have failed visibly at the moment they were made rather than two corrections later.

## And the conclusion still survives, for the third time and for the same reason

Not because 199 is closer than 33, but because the exposure question was never "how many fill loops exist" — it is **"which loops write into a clamped arena whose end abuts verdict state."** That set is `mload`/`mstore`, identified by reading the two clamped call sites, not by counting. Every wrong count so far has been wrong about the *population* and right about the *two members that matter*, which is luck the third time as much as the first.

The honest form of the claim is the one you are now stating: the two clamped loops are enumerated and understood; the unclamped population is large, mostly safe by the mod-32 construction, and **not exhaustively characterised**.

pirapira:
## Fourth pass: the "dynamic-step" residual I named **does not exist** — those 12 are misclassifications

My retraction named ≈12 register-step store loops as the unsettled remainder. I examined all of them. **None is a pointer-stepped loop.** `add reg, reg, other` matched two things that are not pointer stepping:

- **address computation from an index** — `.Lmmul_i`/`.Lmmul_j`, `.Lblk2_load_h`/`_m`, `.Lblq_mul_i`/`_j`, `receipt_records_append`, `.Lblws_meta_loop`, `.Llrr_topic_loop`, `.Lrlen_loop`: all compute `base + i*width` via `slli`+`add`, then store. The actual loop step is a **counter** (`addi s5, s5, 1`) and the exit is `beq counter, limit`;
- **gas arithmetic** — `.L{callcode,delegatecall,staticcall}_target_bn254_fail_burn`: `sub x17, x17, x22` is a gas debit (`env+568`), not a pointer step, and the label is a failure-path branch target rather than a loop head.

**Index-based loops with a counter step of 1 are immune to this hazard**: `beq counter, limit` is always reached, because a step of 1 cannot pass an integer bound. And the address is recomputed from the counter each iteration rather than accumulated, so there is no pointer to drift.

So the residual is not 12. Recounting only loops where a *pointer* is stepped by a static amount **greater than the 1-byte case and outside the memsize family**:

```
2  sd  step=32
1  sd  step=32   exit=beq
1  sd  step=64
1  sd  step=2    exit=beq
1  sb  step=128
```

**≈6 loops.** That is the real remainder — small, static, and individually checkable, unlike the dynamic set I claimed.

## Which means my retraction was also wrong, in the opposite direction

I over-corrected. Having found two surveys that under-reported, I reported a residual I had not examined — and it evaporated on inspection. The pattern across all four passes:

| pass | claim | error direction |
|---|---|---|
| 1–2 | "33 loops, family closed" | under-reported (narrow pattern) |
| 3 | "141" | over-reported (proximity false positives) |
| 4 | "199, ≈12 unsettled" | residual over-reported (arithmetic misread as stepping) |
| 5 | "199 loops, ≈6 unsettled" | — |

So the corrective reflex has its own failure mode: after being wrong optimistically, I was wrong pessimistically, and a named-but-unexamined residual is just as unverified as a premature "closed". **Both directions cost the reader the same thing — a number they cannot act on.**

The rule I would actually keep, having now failed it four times: **do not report a count until every class in the partition has been opened and read.** Not the pattern widened, not the axes varied — the *members* looked at. Twelve loops was ten minutes of reading, and it would have prevented three of the four wrong numbers above.

pirapira:
## Fifth pass — read every member. The residual is **empty**, and this is the first count here I produced by reading rather than by matching.

Opening all eleven candidate loops (pointer stepped by a static amount > 1, against a bound compare) resolves every one:

| loop | step | why it is safe |
|---|---|---|
| `.Lsg_bl` | 4 | exits on `beqz t2` — a **counter**, not a pointer compare |
| `.Lwd_loop` | 32 | exits on `beq s2, s1` — index counter vs limit |
| `.preload_expand_loop` | 64 | exits on `beqz x6` — counter |
| `.blob_hash_copy_loop` | 32 | exits on `beqz x6` — counter |
| `.blockhash_copy_loop` | 32 | exits on `beqz x6` — counter |
| `.Lmexp_ret` | 128 | **not a loop** — stack-frame teardown (`ld ra, 0(sp)` … `addi sp, sp, 128`) |
| `.L{callcode,delegatecall,staticcall}_target_bn254_fail_burn` | 192/160 | **not loops** — failure-path frame arithmetic on `x12`; the `sub x17, x17, x22` is a gas debit |
| `.Lcd_fail_callcode_target` | 192 | **not a loop** — failure-path branch target |
| `.Lbrcl_desc_loop` | 1 **or** 2 | the only genuinely hazardous shape — and **explicitly guarded**, see below |

**A counter-terminated loop is immune**: an integer counter stepping by 1 cannot pass its bound without hitting it, and the address is recomputed from the counter each iteration (`slli`+`add` off a base) rather than accumulated, so no pointer drifts.

**`.Lbrcl_desc_loop` is the one with the real shape, and it is guarded by construction.** Its step is path-dependent — `addi s5, s5, 2` for a typed receipt, `addi s5, s5, 1` for legacy — against an exact `beq s5, s4` exit, which is precisely the configuration where a 2-step could jump over `s4` and never terminate. `ReceiptsConsensus.lean:78` guards it:

```
addi t4, s5, 1; bgeu t4, s4, .Lbrcl_root_fail
```

The 2-step path is entered only when `s5 + 1 < s4`, i.e. `s5 + 2 ≤ s4`, so `s5` can never overshoot. That is an explicit guard rather than an accident — the same defensive shape as this PR's, written by whoever wrote that routine.

## Final state of this sweep

199 true store loops. The step-over hazard needs a *pointer* stepped by more than 1 against a bound compare. Every such loop is one of: counter-terminated (immune), not a loop (false positive), the memsize family (32-congruence by the `li mask,-32` construction), the two clamped fills (pinned by this PR), or explicitly guarded (`.Lbrcl_desc_loop`). **Nothing outstanding.**

## What it took, which is the part worth recording

| pass | method | result |
|---|---|---|
| 1 | narrow pattern | 2 — "family closed" |
| 2 | widened lexical axes | 33 — "family closed" |
| 3 | + ops, register steps | 141 |
| 4 | + required backward control flow | 199, "≈12 unsettled" |
| 5 | **opened and read every member** | 199, **0 unsettled** |

Four wrong numbers, in both directions, each produced by a pattern and each stated with more confidence than it earned. The fifth was produced by reading eleven code fragments and took about the same time as the fourth.

The transferable rule: **a search tells you where to look; only reading tells you what is there.** Every one of my four wrong counts came from treating a search result as a finding. Breadth of pattern — 3048 symbols, all widths, all branch forms — measured how hard I had looked, never whether the matches were the thing I was counting. So: report counts only over sets whose members you have opened, and if the set is too large to open, report it as a search result and say so explicitly, not as an enumeration.

pirapira:

## This is the count I believe, and the reason is the method, not the number

Four counts in this thread were produced by matching and all four were wrong — two low, two high. This one was produced by **opening eleven files and reading them**, and it resolves every member into a named category. That difference is the whole story.

**The durable rule: narrow by matching, close by reading.** Matching is the right tool for getting from 199 to 11 — no human reads 199 loops. It is the wrong tool for the last step, because a pattern cannot distinguish `addi x12, x12, -32` as a fill-loop step from the same instruction as a single EVM-word stack pop, and no amount of widening fixes that; only reading does. **A count whose every member you have not read is a hypothesis, not a result** — and the tell that you have crossed from one to the other is precisely that the population became small enough to read and you did not.

Two findings in the table are worth more than the closure itself:

**Counter-terminated loops are structurally immune**, and that is a cleaner argument than any alignment pin. An integer counter stepping by 1 cannot pass its bound without hitting it, and because the address is recomputed from the counter each iteration (`slli` + `add` off a base) rather than accumulated, no pointer exists to drift. Six of the eleven are safe for that reason. Worth naming as a preferred idiom: *if a fill loop can be written with a counter and a recomputed address, the step-over hazard cannot arise at all* — no alignment precondition, no `bgeu`, nothing to pin.

**`.Lbrcl_desc_loop` is the important one**, and not because it is a risk. It has exactly the hazardous shape — a path-dependent step of 1 or 2 against an exact `beq s5, s4` — and it is **explicitly guarded** at `ReceiptsConsensus.lean:78` with `addi t4, s5, 1; bgeu t4, s4, .Lbrcl_root_fail`, so the 2-step path is entered only when `s5 + 2 ≤ s4`. Someone else, writing an unrelated routine, independently applied the same defence this PR applies.

That reframes #10554 as **conforming to an existing convention rather than introducing one** — which is a stronger justification than the one it shipped with, and it is also the argument that the convention deserves writing down. Two independent authors deriving the same guard means the hazard is discoverable but the mitigation is not yet shared; the next person to write a stepped fill loop will rediscover it or will not. That belongs alongside the `clampToArena` precondition: *a pointer stepped by more than 1 against an exact-equality exit must be guarded, and here are the two ways it is done in this tree.*

PR #10555, opened 2026-07-26
docs(codegen): warn that the memory-gas role/register mapping varies by call site

pirapira:
Stacked on #10554 (`fix/clamp-alignment-precondition`), because the paragraph this edits sits in the same docstring that PR touches. #10554 is already labeled `merge!`, so this is a separate PR rather than an amendment.

## Why

Auditing `updateActiveMemorySizeAsm`'s register-use contract (the fourth claim on that helper, after the warrant in #10551 and the alignment precondition in #10554) found it **sound but resting on an unstated hazard**.

The role→register mapping is **not uniform across the 17 call sites**:

| site shape | `roundedReg` | `currentReg` | `maskReg` |
|---|---|---|---|
| `codecopy`, `extcodecopy`, `log*`, `call_*` | `x16` | `x17` | `x18` / `x5` |
| `mcopy_src`, `mcopy_dst`, `returndatacopy` | `x17` | `x18` | `x19` |
| `ChildFrameCreateTail` | `x18` | `x19` | `x23` |

So *"x17 holds the rounded size"* is true at some sites and false at others. The sparse/const wrappers thread these roles through as parameters, so an opcode's effective assignment is set by its own caller and is not visible at the helper at all.

**The consequence is for future editors.** A hardcoded `xN` added to this block would be `maskReg` at one site and `roundedReg` at another — silently clobbering a value the caller is entitled to keep, at only *some* call sites. That is the worst failure shape to debug, and nothing in the contract warned about it.

The block *does* hardcode two registers today — `x13` (EVM-memory base) and `x20` (frame base) — and both are safe **only because they are read-only**: `x20` appears solely as a load/store base (`ld …(x20)` / `sd …(x20)`), `x13` solely as an `add` source, neither ever as a destination. That is a property of the current body, not one anything enforces. If either becomes a destination, or a third literal appears, the clobber list stops being complete. (The sibling `keccakWordGasAsm` does hardcode `x5`, which **is** `maskReg` at the CALL/LOG sites; it is safe only because it is never reached from inside this block.)

Also records the verified result: the three preserved roles never share a register with the three clobbered ones, at any of the 17 sites.

## Second change: what `MemoryBudgetGuard` has become

It was created for one invariant and has since taken on **three distinct remits** in one session — a guard on #10522's unreachability precondition, a guard on the achievable steps-per-gas constant `k` behind the top theorem's fuel (#10552), and a guard on the clamped fill loop's exit exactness (`clampEnd_alignment_*`). Its header now says it is the home for constant-relationship invariants *generally*, so the next person needing one finds it instead of starting a second file, and asks them to add their remit to the list.

It also records the three rules that came out of using it:

- **coincidence vs construction** — independently-editable constants that happen to line up need a kernel-checked pin; a check against the same symbol as the checked value is maintained by construction and a pin is redundant ceremony;
- where a bound has both halves, **guard only the coincidental half and say why the other needs nothing** (the nested-vs-depth-0 split on `clampEnd_alignment_*`) — otherwise a guard file accumulates assertions nobody can justify later;
- **pin the quantity that is used, not its inputs** — `minRemainingPoolBytes` rather than only `evmMemoryPoolBytes`; `SpecRef.MEMORY_PER_WORD` rather than a local copy.

## Verification

- `lake build EvmAsm.Codegen.MemoryBudgetGuard EvmAsm.Codegen.Programs.EvmMemoryGas` — green.
- **Comment-only, so no emitted guest bytes change and no layout pin moves.** Checked mechanically: zero added lines contain a string literal or `++`. (The check first flagged one line — prose double-quotes inside the docstring, a false positive; I switched them to single quotes so the heuristic stays a clean zero and remains useful.)
- No counts are asserted that I could not re-derive: an earlier draft said "eight sites and six others," which I removed in favour of named, verified examples. A durable docstring is exactly where a recited-from-memory number does the most damage.

## Contract status after this

| claim on `clampToArena` / the block | outcome |
|---|---|
| which sites bound the range | **defect** → #10551 |
| the fill loop's `end` is 8-aligned | **unstated precondition** → #10554 |
| register preservation | sound; **verified**, hazard now documented → this PR |
| the recorded mark is the *unclamped* size | sound; verified (see #10554 discussion) |

Two of four were defects; two were sound but unverified. The last one matters most: the clamp bounds the *physical zero-fill* only, while the EVM-visible `active_memory_size` stays the true rounded value gas pricing requires. Had it clamped the recorded mark, later expansions would have been under-priced — and a byte-identical corpus A/B **would not necessarily have caught it**, since the divergence only appears where the clamp binds and the clamp never binds on reachable inputs. Only reading what the store commits could distinguish the safe version of the change from the unsafe one.

### DIFF
--- a/EvmAsm/Codegen/MemoryBudgetGuard.lean
+++ b/EvmAsm/Codegen/MemoryBudgetGuard.lean
@@ -3,6 +3,29 @@
 
   Build-time guard for the memory-budget coincidence (GH #10522, #10535).
 
+  ## Scope: this is the home for constant-relationship invariants generally
+
+  It was created for one such invariant and has since taken on three distinct
+  remits: a guard on #10522's unreachability precondition, a guard on the
+  achievable steps-per-gas constant `k` that the top theorem's fuel is derived
+  from (#10552), and a guard on the clamped fill loop's exit exactness
+  (`clampEnd_alignment_*`). **If you need to pin a relationship between guest
+  constants that nothing else enforces, add it here rather than starting a
+  second file** — and note the new remit in this list, so the next person finds
+  it too.
+
+  The test for whether an assertion belongs here is *coincidence vs
+  construction*: a relationship that holds because several independently
+  editable constants happen to line up is a coincidence and needs a
+  kernel-checked pin; one where the check and the checked value are the same
+  symbol is maintained by construction and a pin would be redundant ceremony.
+  State which of the two you have, and where a bound has both halves, guard
+  only the coincidental half and say why the other needs nothing (see the
+  nested-vs-depth-0 split on `clampEnd_alignment_*`). Pin the quantity that is
+  *used*, not the ones it is computed from — `minRemainingPoolBytes` rather
+  than only `evmMemoryPoolBytes`, `SpecRef.MEMORY_PER_WORD` rather than a
+  local copy.
+
   ## What this file protects
 
   Three constants currently sit in a relationship that makes two latent defects
--- a/EvmAsm/Codegen/Programs/BlockVerdictDataSectionTail.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictDataSectionTail.lean
@@ -74,6 +74,25 @@ def ziskStatelessVerdictV2DataSectionTail : String :=
   "\nbaap_storage_values:\n  .zero " ++ toString (bsrMaxBalItems * bsrPathBytes) ++
   "\n  .zero " ++ toString (frameArrayBytes - 2 * (bsrMaxStateChanges * bsrEncodedAccountBytes) - (bsrMaxBalItems * baapStorageDescBytes) - 2 * (bsrMaxBalItems * bsrPathBytes)) ++
   "\ncall_frame_arena_end:\n" ++ "\n" ++
+  -- LAYOUT INVARIANT — `rb_running_block_bloom` is IMMEDIATELY adjacent to
+  -- `evm_memory_pool_end`, with ZERO bytes of slack. The `.balign 8` below is a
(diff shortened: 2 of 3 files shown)


pirapira:
Approved — labelling `merge!`. Stacking rather than amending an already-labelled PR is the right call.

Both asks are in, and the second one went further than I suggested in a way worth naming.

**The register paragraph** now says the mapping varies by site with two verified examples, and tells editors to reason about roles rather than register numbers. That converts a fact established across 17 call sites into something the next editor reads in place — which is the whole point, since the non-uniformity is invisible from any single site.

**The `MemoryBudgetGuard` scope header** does something I asked for only half of. I suggested recording that the file has taken on three remits so nobody starts a second one. You added the **admission test**, which is the part that makes the file maintainable rather than merely findable:

> a relationship that holds because several independently editable constants happen to line up is a coincidence and needs a kernel-checked pin; one where the check and the checked value are the same symbol is maintained by construction and a pin would be redundant ceremony

Plus the two corollaries, both of which came out of this session's mistakes rather than from theory: **where a bound has both halves, guard only the coincidental half and say why the other needs nothing** (the nested-vs-depth-0 split), and **pin the quantity that is used, not the ones it is computed from** (`minRemainingPoolBytes` over `evmMemoryPoolBytes`, `SpecRef.MEMORY_PER_WORD` over a local copy).

That last pair matters more than it looks. A guard file without an admission test accumulates assertions nobody can justify later, and then someone deletes the whole file because half of it is obviously redundant — taking the load-bearing half with it. Writing down *why an assertion is here* is what prevents that, and writing down *why a possible assertion is absent* is what stops the file growing ceremony. Both are now in the header, with worked examples from real changes.

Asking each new remit to be added to the list is the right maintenance loop: the file documents its own scope creep, so the third remit is visible to whoever adds the fourth.

PR #10556, opened 2026-07-26
docs(codegen): record the pool-end/block-bloom zero-slack layout invariant

pirapira:
Stacked on #10555. Records, at the site where the layout is defined, a property of the linked image that until now was discoverable only by disassembling it.

## The invariant

`rb_running_block_bloom` begins at **exactly** `evm_memory_pool_end`, with **zero bytes of slack**. From `gen-out/stateless_guest.elf`:

```
b5b19b60  evm_memory_pool
bbb19b60  evm_memory_pool_end       <- pool + 0x6000000 exactly (96 MiB)
bbb19b60  rb_running_block_bloom    <- same address; zero slack
```

The `.balign 8` between the two labels is a **no-op**: `evm_memory_pool` is itself 8-aligned and `evmMemoryPoolBytes` is a multiple of 8, so the directive adds no padding. The adjacency is therefore not incidental to the current constant values — it follows from them, and it holds in the shipped image.

## Why it is worth stating

**It reclassifies the clamped fill loop's exit exactness from hygiene to soundness.** #10554 ruled out an up-to-7-byte overshoot past the clamped end. What sits at the far end of that write turns out to be the running block-bloom accumulator, which feeds the receipt root — so an off-by-a-few-bytes write there is an **accept/reject flip**, not a memory-safety curiosity.

That is the same reclassification #10521/#10522 went through, by the same mechanism: the severity of an out-of-bounds write is determined by what happens to be adjacent, which is a fact about the *image* and not about the loop. Nothing in the source said what followed the pool.

It also constrains every **future** pool-touching change, not just the clamp. Any loop that fills, zeroes, or copies into the pool has zero margin at the top end.

## Explicitly: do not add padding here

The comment says so. Inserting slack would *hide* the constraint rather than state it, and would silently absorb exactly the off-by-N bugs that should fail loudly. A guard band converts a reproducible wrong-verdict into an invisible near-miss.

## Second change: the transition condition, as a trigger

#10554 guarded the depth-0 half of the alignment argument and deliberately left the nested half unguarded, because there the clamped end reduces algebraically to the `evm_memory_pool_end` **label** — the assembler owns that alignment and nothing arithmetic can drift it. That call stands.

But the nested half is the one sitting against the sensitive neighbour, so the docstring now states the trigger rather than leaving it as an observation: **if that end ever becomes a computed value instead of reducing to the label, it moves from construction class to coincidence class and needs a pin alongside the depth-0 `clampEnd_alignment_*` guards.** A future editor changing the reduction will read the condition at the point where they would break it.

## Verification

- `lake build EvmAsm.Codegen.Programs.BlockVerdictDataSectionTail EvmAsm.Codegen.Programs.EvmMemoryGas` — green (1021 jobs).
- **Comment-only.** Every added line is a Lean `--` comment; zero added lines contain a string literal or `++`, so the emitted asm is unchanged by construction and no layout pin moves. Worth being precise about here, since the comment is inserted *between* `++` operands of the data-section emission — if it altered the emitted string, `RegionMap.textSizeBytes` and the whole address map would shift.
- The two addresses were read from the linked ELF's symbol table, not inferred from the source directives. `.text` reads `0x611c0`, matching the current pin.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdictDataSectionTail.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictDataSectionTail.lean
@@ -74,6 +74,25 @@ def ziskStatelessVerdictV2DataSectionTail : String :=
   "\nbaap_storage_values:\n  .zero " ++ toString (bsrMaxBalItems * bsrPathBytes) ++
   "\n  .zero " ++ toString (frameArrayBytes - 2 * (bsrMaxStateChanges * bsrEncodedAccountBytes) - (bsrMaxBalItems * baapStorageDescBytes) - 2 * (bsrMaxBalItems * bsrPathBytes)) ++
   "\ncall_frame_arena_end:\n" ++ "\n" ++
+  -- LAYOUT INVARIANT — `rb_running_block_bloom` is IMMEDIATELY adjacent to
+  -- `evm_memory_pool_end`, with ZERO bytes of slack. The `.balign 8` below is a
+  -- no-op here: `evm_memory_pool` is itself 8-aligned and `evmMemoryPoolBytes`
+  -- is a multiple of 8, so the bloom starts at exactly `evm_memory_pool_end`.
+  -- Verified in the linked image (`gen-out/stateless_guest.elf`): both symbols
+  -- resolve to 0xbbb19b60.
+  --
+  -- CONSEQUENCE, and it is a soundness one: ANY overshoot in ANY loop that
+  -- fills or zeroes the pool corrupts VERDICT STATE, not padding. The running
+  -- block bloom feeds the receipt root, so an off-by-a-few-bytes write here is
+  -- an accept/reject flip rather than a memory-safety curiosity. This is why
+  -- the clamped fill loop's `bgeu` exit must be EXACT — see the alignment
+  -- precondition on `clampToArena` in `Programs/EvmMemoryGas.lean` and the
+  -- `clampEnd_alignment_*` pins in `Codegen/MemoryBudgetGuard.lean`.
+  --
+  -- DO NOT insert padding here to create slack. Padding would hide this
+  -- constraint rather than state it, and would silently absorb exactly the
+  -- off-by-N bugs that should fail loudly. The adjacency is a property of the
+  -- image that constrains every future pool-touching change; keep it stated.
   ".balign 8\n" ++
   "evm_memory_pool:\n  .zero " ++ toString evmMemoryPoolBytes ++ "\n" ++
   "evm_memory_pool_end:\n" ++
--- a/EvmAsm/Codegen/Programs/EvmMemoryGas.lean
+++ b/EvmAsm/Codegen/Programs/EvmMemoryGas.lean
@@ -218,6 +218,16 @@ def memoryArenaLimitAsm (tag limitReg : String) : String :=
       BOTH constants 8-aligned. That half is arithmetic, so it is pinned by
       `Codegen/MemoryBudgetGuard.lean`'s `clampEnd_alignment_*` guards.
 
+    TRIGGER for the nested half: it is unguarded *only* because the end IS the
+    label, so nothing arithmetic can drift it. **If you ever change that end to
+    be computed rather than to reduce to `evm_memory_pool_end` itself, it moves
+    from construction class to coincidence class and needs a pin alongside the
+    depth-0 ones.** That half is also the one whose overshoot is most costly:
+    `rb_running_block_bloom` begins at *exactly* `evm_memory_pool_end` with zero
+    bytes of slack (verified in the linked image; see the layout invariant at
+    the pool's emission site in `Programs/BlockVerdictDataSectionTail.lean`), so
+    an inexact exit there corrupts verdict state rather than padding.
+
     Gas and MSIZE semantics are deliberately UNCHANGED by the clamp — the charge
     stays the full spec `extend_memory.cost` and `env+488` still records the full
     `ceil32(offset+size)`. Only the *materialization* is bounded, which loses no


pirapira:
Approved — labelling `merge!`. Eleventh PR, stacked on #10555, comments only.

Both asks are in, and each is placed where it will actually be read rather than where it was discovered.

**The adjacency invariant sits at the pool's emission site**, which is the right home — the layout is defined there, so someone changing the pool sees the constraint in the same file they are editing. Four things it gets right:

- It records the **verified image addresses** (both symbols at `0xbbb19b60`), so the claim is checkable rather than asserted, and a future reader can re-verify with one `nm`.
- It notes the `.balign 8` between the labels is a **no-op** given the pool's own alignment and size. Without that, the next reader sees a `.balign` and reasonably assumes it provides the slack — the directive looks like protection and is not.
- It states the consequence as **soundness**, not hygiene: the bloom feeds the receipt root, so an off-by-N write is an accept/reject flip.
- **"DO NOT insert padding here to create slack"** with the reason — padding would absorb exactly the off-by-N bugs that should fail loudly. Prohibiting the plausible-looking fix is the part that survives a well-meaning future editor.

**The trigger on the nested half is stated as a condition rather than a possibility**, which is what makes it actionable: *if you ever change that end to be computed rather than to reduce to `evm_memory_pool_end` itself, it moves from construction class to coincidence class and needs a pin.* And pairing it with the observation that this half's overshoot is the most costly one closes the loop — the unguarded half is the dangerous-neighbour half, unguarded for a good reason that is now written down alongside the condition under which the reason expires.

That cross-references cleanly in three directions: the emission site states the layout fact, `EvmMemoryGas.lean` states the precondition and the trigger, `MemoryBudgetGuard.lean` pins the arithmetic half. Someone arriving at any one of the three finds the other two.

pirapira:
## The survey is now exhaustive: exactly two zero-slack region boundaries, and only one of them is dangerous

Rather than leave the pool/bloom adjacency as one lucky disassembly, I enumerated **every** region boundary in the linked image and measured the gap to the next symbol — so nobody has to redo this to know whether there are others.

All 3048 symbols from `gen-out/stateless_guest.elf`, every `*_end`, gap to the next distinct symbol:

```
brl_item_end                     gap=8   next=brl_wd_len
svf_witness_end                  gap=8   next=svf_codes_ptr
bv_system_storage_capture_end    gap=8   next=bv_system_storage_capture_rows
bara_bal_end                     gap=8   next=bara_next_item
call_frame_arena_end             gap=0   next=evm_memory_pool          <-- zero slack
evm_memory_pool_end              gap=0   next=rb_running_block_bloom   <-- zero slack
_end                             gap=0   next=__BSS_END__              (linker symbols)
```

The four `gap=8` rows are **not** region boundaries — they are 8-byte cursor *variables* followed by the next scalar in the same bookkeeping block (`brl_item_end`, `bara_bal_end` etc. hold offsets, they do not delimit arenas). So the image has exactly **two** real region-boundary adjacencies, plus one linker artifact.

## And the second one is in the benign direction, which is the useful part

`call_frame_arena_end` abuts `evm_memory_pool` with zero slack too — but overshooting the **frame arena** lands in 96 MiB of guest EVM memory, i.e. *scratch*, whereas overshooting the **pool** lands directly in the verdict accumulator. The asymmetry matters and is worth naming:

- corrupting guest EVM memory perturbs *execution*, and a perturbed execution almost always fails the receipt/state-root comparison → **false reject**;
- corrupting `rb_running_block_bloom` perturbs *the value being compared* → the failure can go either way, including **accept**.

So the pool is effectively a 96 MiB buffer protecting verdict state from anything below it, while the pool's own top end has no buffer at all. **The layout has generous slack in one direction and none in the other**, and the direction with none is the one where corruption is accept-capable. That is the sharper version of the invariant this PR records, and it is the reason padding would be the wrong fix: the asymmetry is informative, and evening it out would only make the dangerous side quiet.

## Status of this as a claim

Mechanical and reproducible in one command — `objdump -t`, sort by address, diff consecutive symbols — not an inspection result. Two facts I want to be precise about:

- The enumeration covers **symbols**, so it would miss a sensitive region that has no symbol of its own. I did not verify the absence of such a region, and say so rather than implying the survey is stronger than it is.
- The severity split above is an argument, not a proof: I have not exhibited an input that makes a corrupted bloom *match*. The claim is that pool-side corruption is accept-**capable** where frame-arena-side corruption is not, which is enough to justify guarding the exit exactness — it is not a claimed exploit.

Happy to move the completeness table into the code comment as a twelfth PR if that is worth an in-tree record; this PR is already labeled, so I have not touched its diff.

pirapira:
## Correction: "exactly two zero-slack region boundaries" is **false** — there are ~55, and my survey answered the wrong question precisely

My adjacency survey enumerated symbols named `*_end` and reported **two** real zero-slack region boundaries. That pattern is structurally blind: **`evm_memory` has no `evm_memory_end` symbol** — its top is `base + rootRuntimeMemoryArenaLimitBytes` — so a name-keyed sweep cannot see it. I found that boundary later by hand, via the `__BSS_END__` coincidence, without noticing it contradicted the sweep I had just called exhaustive.

Re-running the sweep keyed on **size** rather than name (flag every symbol whose gap to its successor exceeds 1 MiB, then report the successor) gives ~55 large allocations, and **essentially every one abuts its successor with zero slack** — because a packed `.bss` has no padding anywhere. A sample:

```
baap_storage_values  0x00b31c2060   41.34 MiB  top=0x00b5b19b60 -> call_frame_arena_end
call_frame_arena_end 0x00b5b19b60   96.00 MiB  top=0x00bbb19b60 -> evm_memory_pool_end
evm_memory           0x00bee557c0    4.00 MiB  top=0x00bf2557c0 -> __BSS_END__
__BSS_END__          0x00bf2557c0    7.17 MiB  top=0x00bf980000 -> .sszscratch
```

So zero slack is the **default condition of the layout**, not a property of two special boundaries. "Exactly two" was an artifact of the naming convention I keyed on.

## What survives, and it is the part that mattered

The severity conclusion is unchanged, because the relevant question was never *how many* zero-slack adjacencies exist — it is **which boundaries a fill loop can overshoot past**. That set is determined by the fill-loop enumeration (33 store-stepping loops, all of them memory-arena fills), not by the layout: the other ~53 arenas are written by their own routines against their own bounds, and none of the 33 loops targets them.

So:

- **`evm_memory_pool_end` → `rb_running_block_bloom`** remains the dangerous adjacency, because a *clamped fill loop* ends there. Unchanged.
- **`evm_memory` top → unmapped gap** remains the second line of defence, and #10559's gate pins it. Unchanged.
- **The other ~53 zero-slack adjacencies are not reachable by these loops**, which is why the wrong count did not produce a wrong conclusion.

That is luck, not method. A survey that answers the wrong question precisely is more dangerous than one that is obviously incomplete, because the precision is what makes it persuasive — and this one got repeated back to me approvingly several times before I checked it.

## The generalisable error

Keying a completeness sweep on a **naming convention** rather than on the **property** being enumerated. `*_end` is how *some* regions record their top; `base + size` is how others do. The property was "a region's top address", and only one of its two spellings was matched.

Same failure as my first pass at the fill-loop sweep, which matched `sd` + 8-step + `bgeu` and found 2 of 33. Both times the narrow pattern returned a small, clean, quotable number. **Under-matching reads exactly like a clean result** — so a completeness claim needs the axes it varied stated alongside it, and this one should have said "symbols named `*_end`", which would have made its blind spot obvious on sight.

Issue #10557, opened 2026-07-26
layout (premise refuted, see comment): can any zero-slack neighbour of the pool be made comparison-irrelevant?


pirapira:
> **⚠️ PREMISE REFUTED — read the correction comment before acting on this issue.**
>
> The asymmetry table below is **wrong**, and the author of the table refuted it (see comment). `evm_memory_pool` is **not scratch**: LOG reads guest EVM memory for its data (`EvmMemoryGas.lean:651` charges LOG for a memory range), log data is RLP-encoded into the receipt, and the receipt reaches the **receipts root** — a compared value. So corrupting the region below the pool is comparison-relevant too, by a longer path.
>
> The deeper error is mine and is worth stating separately, because it generalises past this issue: I claimed memory corruption "perturbs execution → fails the root comparison → **false reject**." **That assumes an honest header.** The attacker supplies the header as well as the block, so a block crafted such that the *corrupted* execution reproduces the header's claimed root is precisely the false-accept shape. Corruption of any committed value is accept-capable regardless of which side of the pool it lands on. The overshoot writes zeros, which limits attacker control but removes accept-capability on neither side — zeroing bloom bits only clears bits in an OR-accumulator, matching a header claiming fewer; zeroing log bytes yields a receipt a header can claim.
>
> **What remains valid:** the exhaustive survey (exactly two zero-slack region boundaries in the image), the recorded layout invariant (#10556), and the alignment pins (#10554). **What does not:** the claim that reordering as scoped would put every zero-slack adjacency in the false-reject direction. A reorder is worth doing only if a genuinely reader-verified neighbour exists — one whose reads cannot reach the bloom, any root, or any header-compared value. **No such symbol has been identified in the current layout.**

---

> **Note:** This comment was written by Claude Code.

## The asymmetry suggests a better fix than padding — reorder, don't pad

The survey turns a lucky disassembly into a closed question, and the asymmetry it exposes is the most useful thing in it:

| boundary | overshoot lands in | failure direction |
|---|---|---|
| `call_frame_arena_end` → `evm_memory_pool` | 96 MiB of guest EVM memory (scratch) | perturbs *execution* → fails the root comparison → **false reject** |
| `evm_memory_pool_end` → `rb_running_block_bloom` | the verdict accumulator | perturbs *the value being compared* → **accept-capable** |

[...]

PR #10559, opened 2026-07-26
fix(scripts): pin the unmapped backstop below __BSS_END__

pirapira:
Follows #10556's layout invariant and the analysis in #10557. Scripts only.

## What this pins

An overshoot past a dense memory arena's top end corrupts whatever is mapped above it — and `rb_running_block_bloom` sits at **exactly** `evm_memory_pool_end` with zero slack (#10556), so on that boundary an off-by-N reaches verdict state.

The mitigation that costs nothing is already present in the image, by accident: **one of the two arenas ends exactly at `__BSS_END__`**, whose neighbour is ~7.2 MiB of **unmapped** address space. Nothing there can be corrupted into a committed value, because nothing is there.

```
evm_memory                            bee557c0
+ rootRuntimeMemoryArenaLimitBytes      400000
= depth-0 clamp end                   bf2557c0   exact coincidence with…
__BSS_END__                           bf2557c0
.sszscratch                           bf980000   next mapped section
unmapped gap                            72a840   ≈ 7.2 MiB
```

That is a **coincidence between an arena size and a section layout**, not a construction — nothing relates `rootRuntimeMemoryArenaLimitBytes` to where the linker happens to end `.bss`. By the coincidence-vs-construction rule this needs an external check: appending any new `.bss` section would push both arenas away from the boundary and remove the backstop **with no other signal**.

## Why here and not in `MemoryBudgetGuard`

It cannot be a `decide` theorem: the invariant relates a Lean constant to a **link-time address**, which Lean cannot see. `check-region-map.sh` is the right home — it already reads ELF symbols with `readelf` and exists to gate link-layout drift.

## Written to survive the reorder it may enable

The check asserts *"whichever arena ends at `__BSS_END__`"*, not *"`evm_memory` ends at `__BSS_END__`"*. #10557 contemplates moving `evm_memory_pool` to the top so the **unpinned, construction-class** boundary gets the backstop instead of the already-pinned one. Written the narrow way, this gate would have blocked that improvement; written this way it permits either layout and fails only if **neither** arena is last. It also prints which one provides the backstop, so the output records the current choice:

```
  OK   a dense arena ends at __BSS_END__ (unmapped backstop): evm_memory
```

## Verification

- **Green**: `OK … : evm_memory`, whole script passes.
- **Red-tested**: perturbing the offset to `0x400008` yields `DRIFT … (unmapped backstop): NEITHER` and the script takes its failure path (`check-region-map: DRIFT detected`). Restored, re-verified green. A gate that has never been observed to fail is not yet a gate.
- Scripts only — no Lean, no emitted guest bytes, no layout pin moves.

## One limit stated explicitly

This pins that **nothing in the image** sits above the last arena. It does **not** establish that a store into that gap *faults* — that is a property of each backend's memory model (ziskemu, Spike, the prover), which I have not checked. A zkVM might permit such a store or treat it as an unconstrained cell. So the guarantee here is "no mapped neighbour to corrupt", which is weaker than "the write is refused", and #10557 carries the stronger question.

### DIFF
--- a/scripts/check-region-map.sh
+++ b/scripts/check-region-map.sh
@@ -105,9 +105,10 @@ symaddr() { "$READELF" -sW "$ELF" | awk -v n="$1" '$8==n {print $2; exit}'; }
 python3 - "$(symaddr call_frame_arena)" "$(symaddr basr_values)" "$(symaddr basr_accounts)" \
   "$(symaddr bv_system_storage_log)" "$(symaddr baap_storage_desc)" "$(symaddr baap_storage_paths)" \
   "$(symaddr baap_storage_values)" \
-  "0x$BSS_BASE" "0x$BSS_SIZE" "$(symaddr evm_memory_pool)" "$(symaddr evm_memory_pool_end)" <<'PY' || fail=1
+  "0x$BSS_BASE" "0x$BSS_SIZE" "$(symaddr evm_memory_pool)" "$(symaddr evm_memory_pool_end)" \
+  "$(symaddr evm_memory)" <<'PY' || fail=1
 import sys
-(cfa, bval, bacc, syslog, desc, paths, vals, bbase, bsize, pool, pend) = [int(x,16) for x in sys.argv[1:]]
+(cfa, bval, bacc, syslog, desc, paths, vals, bbase, bsize, pool, pend, emem) = [int(x,16) for x in sys.argv[1:]]
 # RegionMap constants (kept in sync with BlockVerdictParams.lean).
 S = 100018*256          # bsrMaxStateChanges*bsrEncodedAccountBytes
 syslogL = 32768*128     # bvSystemStorageLogBytes (4ch8f.73: 2*16384 rows, standalone)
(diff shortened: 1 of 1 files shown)


pirapira:
Approved — labelling `merge!`. Script gate only, one file.

**Choosing `check-region-map.sh` over `MemoryBudgetGuard.lean` is right, and the reason is a genuine boundary on the rule you wrote three hours ago.** That file is the home for *constant-relationship* invariants — relationships Lean can see. This one relates a Lean constant to a **link-time address**, which Lean cannot see at all, so a `decide` there would be pinning a number nobody had checked against the image. `check-region-map.sh` already reads ELF symbols with `readelf` and exists to gate link-layout drift, so the invariant lands where the evidence lives. Worth noting the boundary in `MemoryBudgetGuard`'s scope header next time you touch it: *constants only — if the invariant involves an address the linker chooses, it belongs in the region-map gate.* Otherwise the scope header you just wrote will pull link-time facts into a file that cannot verify them.

**The forward-compatible formulation is the part I'd have missed.** Writing the check as *whichever arena ends at `__BSS_END__`* rather than *`evm_memory` ends at `__BSS_END__`* means it survives the #10557 reorder — which would deliberately move the boundary to `evm_memory_pool` — and fails only if **neither** arena backstops.

That is the difference between a guard that lasts and one that gets deleted. A gate pinned to the current *instance* fires on the first **intended** change, at which point whoever is doing the reorder reasonably weakens or removes it, and the protection is gone precisely when the layout is in motion — the moment it was most needed. Pinning the *property* means the guard stays silent through the intended change and still fires on accidental drift. **Pin the property, not the instance.**

The comment also carries what a failure means — that appending any new `.bss` section silently pushes both arenas off the boundary and removes the backstop with no other signal — which is the same discipline as #10540's docstring: a gate failure is a correctness handoff, not a build annoyance.

One thing this makes explicit that was previously implicit: the depth-0 arena's unmapped backstop was an **accident** of an arena size meeting a section layout. It is now a checked property. That is the second line of defence behind #10554's alignment pins, and unlike those pins it protects against a class of change (`.bss` growth) that has nothing to do with memory code at all.

pirapira:
## Correction: the gap is not trapping memory — it is silent memory

I called the region above `__BSS_END__` "unmapped", and the reframing built on that (*unmapped beats scratch because a store there **traps**, so an off-by-N is loud and cannot hide*) does **not** hold. I flagged the memory model as unverified; having now checked it, the stronger property is false.

`EvmAsm/Codegen/RegionMap.lean:78` defines the RAM zone as

```
/-- Verified RAM window `[RAM_MEM_START, RAM_MEM_END)` = `0xa0000000..0xc0000000`. -/
```

and the enclosing docstring says these zones are *"recognised by the verified `isValidMemAddr` predicate."* The gap runs `bf2557c0 … bf980000` — **inside** that window. So those addresses are valid memory in the verified machine model. A store there is legal, constrained, and **silently absorbed**.

"Unmapped" was true only of the **ELF section table**; it is not true of the address space the machine model recognises. Those are different claims and I conflated them.

## What this changes

- The backstop is the **harmless-to-corrupt** kind, not the **refuses-the-write** kind. The quality ordering that put unmapped space above a scratch symbol does not apply: functionally this gap **is** a scratch neighbour.
- Which means it is also, functionally, **padding** — the fix that was rejected on the grounds that padding "absorbs exactly the writes we want failing." A 7.2 MiB unclaimed gap absorbs them identically. That rejection was right on its merits; it just does not distinguish this gap from padding.
- For the reorder: it would convert a **verdict-corrupting** off-by-N into a **silent** one. That is still a large improvement — accept-capable to harmless — but it should not be described as loud, or as something that cannot hide. It hides completely.

## What survives, and it is enough to keep the pin

The acceptance test was never "the write traps"; it was *"its reads cannot reach the bloom, any root, or any header-compared value."* The gap satisfies that in the strongest available way: **nothing is allocated there, so nothing reads it.** No committed value can derive from bytes in that range.

So #10559's gate still earns its place — it pins that the dangerous boundary has a neighbour whose corruption cannot reach a committed value, and it still fires if a future `.bss` addition removes that property. Only the *marketing* was wrong: read it as "harmless neighbour", not "trapping neighbour". The check itself, and its red test, are unaffected.

## Method note

This is the second claim in this thread that broke the same way: an argument that felt structural, was endorsed, and then failed on a premise nobody had tested. Last time it was *"corruption trends false-reject"* assuming an honest header; this time *"unmapped means trapping"* conflating the ELF section table with the machine model. Both were mine. The pattern is that the appealing word — "false-reject", "unmapped" — carried an unexamined implication, and in both cases the check that settled it was a single file read I could have done before asserting rather than after.

pirapira:

## Merged via batch #10564 — closing; do not rebase

Content is already on main at `daccd2def` — `scripts/check-region-map.sh` carries the unmapped-backstop gate, verified by reading the merged file rather than the PR status.

GitHub reports `OPEN / BEHIND` because the batch merged a separate branch, so this PR's merge-base never advanced. Stale status, not a signal. Closing as merged-via-batch.

Issue #10560, opened 2026-07-26
assertion-gap: isValidMemAddr checks only the start address, so isValidDwordAccess admits an 8-byte store past every zone top


pirapira:
## Site

`EvmAsm/Rv64/Word.lean:77-81`

```lean
def isValidMemAddr (addr : Word) : Bool :=
  (decide (MEM_START ≤ addr.toNat) && decide (addr.toNat ≤ MEM_END)) ||
  (decide (INPUT_MEM_START ≤ addr.toNat) && decide (addr.toNat ≤ INPUT_MEM_END)) ||
  (decide (RAM_MEM_START ≤ addr.toNat) && decide (addr.toNat ≤ RAM_MEM_END))

def isValidDwordAccess (addr : Word) : Bool :=
  isValidMemAddr addr && isAligned8 addr
```

## Finding

**`isValidMemAddr` validates the access's START address only. It does not constrain the access's EXTENT.** A doubleword access touches 8 bytes, but only one address is checked — so at the top of each zone the predicate admits a store that writes entirely outside it.

All three zone tops use a **closed** bound (`≤ …_END`), and all three ends are 8-aligned:

| zone end | value | 8-aligned? | `isValidDwordAccess(END)` | bytes actually touched |
|---|---|---|---|---|
| `MEM_END` | `0x78000000` | yes | **true** | `0x78000000 … 0x78000007` |
| `INPUT_MEM_END` | `0x40002000` | yes | **true** | `0x40002000 … 0x40002007` |
| `RAM_MEM_END` | `0xc0000000` | yes | **true** | `0xc0000000 … 0xc0000007` |

So an `sd` at any zone top passes `isValidDwordAccess` while writing 8 bytes at or past the tail. `isValidMemAccess` (4-byte, `isAligned4`) has the same shape over 4 bytes.

## And the docstring disagrees with the code about whether the endpoint is even in the zone

`EvmAsm/Codegen/RegionMap.lean:70-72` describes these zones as **half-open**:

> *"Each zone is a half-open `[lo, hi)` byte range recognised (for the RAM / INPUT zones) by the verified `isValidMemAddr` predicate."*

[...]

Issue #5164, opened 2026-05-20
isValidMemAddr does not cover ziskemu's RAM region (0xa0000000..0xc0000000)


pirapira:
## Problem

The verified-side memory-validity predicate `isValidMemAddr`
(`EvmAsm/Rv64/Basic.lean:244,247`) restricts valid memory addresses to

```
MEM_START = 0x20
MEM_END   = 0x78000000
```

But the executable ziskemu target uses a different layout
(`EvmAsm/Codegen/Driver.lean:68-82`, `EvmAsm/Codegen/Programs.lean`):

| Region | Address range | Status under `isValidMemAddr` |
|---|---|---|
| `.text` | `0x80000000..0xa0000000` | outside (read-only on ziskemu) |
| `INPUT_ADDR` | `0x40000000..0x40002000` | **outside** (host-supplied input) |
| `.data` | `0xa0000000..0xa0010000` | **outside** |
| `OUTPUT_ADDR` | `0xa0010000..0xa0020000` | **outside** (public output) |
| Free RAM | `0xa0020000..0xc0000000` | **outside** |

Every build unit that produces observable output writes to
`OUTPUT_ADDR`, so every such program writes outside the verified
predicate. Same for any unit that seeds operands in `.data`
(`evm_add` already does) or reads from `INPUT_ADDR`
(`evm_add_from_input`, `input_echo`).

## Consequence

The verified `step` semantics treat memory writes outside the
predicate as no-ops (or trap, depending on the constructor). The
chain of CPS-triple specs assumes `isValidMemAddr` holds for every
load/store. So the proofs we already have for opcode handlers
trivially hold for their stack-internal LD/SD pairs (those sit in
the `0x20..0x78000000` range a verified caller would set up), but
the moment any spec needs to talk about writing to `OUTPUT_ADDR`,
seeding from `.data`, or reading from `INPUT_ADDR`, it will hit the
predicate.

This surfaced while writing
`EvmAsm/Stateless/MemoryLayout.lean` for the
`run_stateless_guest` scaffold (PR #5158, currently in the merge
queue): the stateless guest by construction writes its public
output to `OUTPUT_ADDR` and reads its SSZ-encoded input from
`INPUT_ADDR`, so the eventual top-level $cpsTripleWithin$ spec
**cannot** be stated under the current predicate.

## Options for discussion

1. **Widen the bounds.** Add `0x40000000..0x40002000`,
[...]

PR #10561, opened 2026-07-26
docs(rv64): record the valid-memory zone invariants at the predicate

pirapira:
Documents four facts about `isValidMemAddr` that were true of the code but recorded nowhere — or contradicted by the prose beside it. All four were verified during #10560; this PR puts them where the predicate is, rather than only in an issue thread.

Comment-only, `EvmAsm/Rv64/Word.lean`.

## 1. "Three disjoint regions" is false — `Input ⊆ Legacy`

`0x20 ≤ 0x40000000` and `0x40002000 ≤ 0x78000000`, so the Input disjunct accepts nothing the Legacy disjunct does not. There are effectively **two** zones, and the Input disjunct is redundant. Nothing is built on the disjointness claim — no lemma proves or assumes it — but the prose invited relying on it.

## 2. Input reads are admitted by **Legacy**, not by Input

Real host inputs greatly exceed the documented 8 KiB Input window. Measured across the fixture corpus: **10,336 of 108,136 inputs (9.6%) exceed the 8,176-byte body capacity; largest ≈8 MiB**, whose top lands at `0x408020b0` — inside Legacy, far outside Input.

This matters because Legacy is documented only as *"preserve the pre-#5164 behaviour for unchanged proofs"*, which reads as safe to remove. **It is not.** Retiring it without first extending Input to reach real input sizes would break the model's input read path for ~10% of the corpus. Recorded so the obvious-looking cleanup is not attempted blind.

## 3. The text-window exclusion is load-bearing for **soundness**

This is the part I would most want a future reader to find. `isValidMemAddr` returns `false` for `[0x80000000, 0xa0000000)`, where `.text` is linked — which looks like an oversight next to `RegionMap`'s `RegionZone.text`. It is not.

`isValidMemAddr` governs **data** accesses only; every consumer is a load or store. Code lives in a separate `CodeMem` map on `MachineState`, provably immutable across execution:

```
Execution.lean:234  code_execInstrBr : (execInstrBr s i).code = s.code   [simp]
Execution.lean:846  code_step        : step s = some s' → s'.code = s.code
Execution.lean:878  code_stepN       : stepN k s = some s' → s'.code = s.code
```

Those lemmas and this exclusion are **two halves of one invariant**: code is immutable *and* unreachable by stores. Admitting text-window stores would let the model update `mem` while still proving `s'.code = s.code` — i.e. assert the instruction stream is unchanged for a write that corrupts code on the real machine. That is the model being *more optimistic than the machine*, the unsound direction.

The two halves lived in different files with nothing linking them, which is precisely the condition under which one gets removed as an inconsistency. The docstring now says changing this means making `code` mutable and revisiting every proof using those simp lemmas — a redesign, not a tidy-up.

## 4. Closed vs half-open: this file is authoritative

Ranges here are **closed** `[lo, hi]` (`addr ≤ …_END`), matching the code. `Codegen/RegionMap.lean:71` describes its zones as half-open `[lo, hi)`. The docstring now says which is authoritative and that `RegionZone` serves a different purpose — classifying where sections are *linked*, not which addresses a load/store may touch. They were never meant to correspond.

## Also: the known gap, flagged at the point of use

The access predicates check only the access's **start** address, never its extent (#10560). So `isValidDwordAccess` must not be read as *"this 8-byte access is in bounds"* — at a zone top it admits an aligned access whose bytes lie outside every zone. Stated here so a reader discharging the obligation knows what it does and does not give them.

## Verification

- `lake build EvmAsm.Rv64.Word` — green.
- **Comment-only**: the diff adds zero `def`/`theorem`/`lemma`/`instance` lines and is confined to the existing `/-! … -/` block. No proof term changes, so no risk to the axiom surface.
- Heads-up for CI: touching this file invalidates a large dependency cone (`isValidByteAccess` alone is referenced in 258 files), so expect a broad rebuild despite this being documentation.


pirapira:
Approved — labelling `merge!`. One file, comments only.

This documents the invariant at the place someone would break it, which is the whole point. Four things it gets right:

- **The prohibition is stated as a prohibition**, with the mechanism: *"do not add a disjunct for it"* followed by why the model would then be more optimistic than the machine. A reader who wants text-window data access now has to argue against `code_step`/`code_stepN` explicitly rather than discovering the coupling afterwards.
- **It names the other half by file and line** (`Execution.lean:234,846,878`). The two halves of this invariant live in different files, which is exactly why the coupling was invisible — and the fix for that is a citation, not a restatement.
- **It gives the cost of changing it anyway** — "making `code` mutable and revisiting every proof that uses those simp lemmas" — which is what turns a prohibition into a decision someone can price rather than a rule they can dismiss.
- **The `RegionZone` divergence is resolved by stating which is authoritative and why they differ**, rather than by declaring one wrong. `RegionZone` classifies where sections are *linked*; `isValidMemAddr` governs which addresses a load or store may touch. Those are different partitions that share a vocabulary, and saying so is more useful than reconciling them.

The Legacy/Input note carries the one operational consequence that matters: **retiring the Legacy disjunct as "dead" requires first extending Input to reach real input sizes**, because Legacy is what actually admits the ≈8 MiB inputs today. Without that sentence, "the Input disjunct is redundant" reads as an invitation to delete the wrong one.

On the framing in your message — four instances tonight of *the code is right, the reason is unrecorded, and the unrecorded reason is what gets deleted* — that is the correct diagnosis and this PR is the right response to it. Worth noting the variant that makes it hard to catch: in three of the four, something in the code **looked like** it supplied the missing reason. The `.balign 8` looked like it provided slack. The `clampToArena = false` sites looked like they had all proved the bound. A guard-file assertion looks self-justifying because it is a theorem. The absent reason is not merely missing; it has a plausible stand-in, which is why nobody notices it is absent until the stand-in is tested.

PR #10562, opened 2026-07-26
docs: record lhkn7 lane handoff

pirapira:
Records lhkn7 slice ordering, selector-retirement preconditions, artifact provenance, negative space, and the two-run attribution-plus-shipping evidence procedure. Documents the 3000-gas execution-budget failure that motivated the procedure. Docs-only; no emitted bytes or regen.

### DIFF
--- a/docs/agents/lhkn7-handoff.md
+++ b/docs/agents/lhkn7-handoff.md
@@ -1,0 +1,25 @@
+# lhkn7 lane handoff
+
+Current shipped slice #10520 routes MTx direct precompiles through the shared scalar kernel. It is fault removal, not FR clearance: exact-parent A/B attributed `22670` and `22671` FAULT-to-OK; post-rebase shipping sweep measured FA=0, FR=68, FAULT=0.
+
+## Remaining slices
+
+These are maintainer-order work; do not start them overnight. Each is a count==1-to-MTx superset slice requiring a full sweep and review.
+
+- Core legacy: bv_fail 41, then 47.
+- Type-4: bv_fail 40.
+- Blob/receipt: bv_fail 53, 37, and 44.
+
+Selector retirement requires every count==1 surface implemented by MTx, a valid forced count==1 diagnostic with no forced OK-to-FR transitions, and a final ordinary full-corpus measurement. No partial forced result authorizes selector retirement.
+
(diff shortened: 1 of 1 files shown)


pirapira:
Approved — labelling `merge!`. Docs only, one file, explicit-path commit from current main.

This belongs in the repo rather than in a chat log or a `/tmp` file, and the content is right. Four things worth calling out:

- **The slice order with `bv_fail` surfaces** (core-legacy 41/47, type4 40, blob/receipt 53/37/44) is the part that is not recoverable from the code. It took this lane weeks to work out and would have had to be re-derived by whoever picks it up.
- **Selector-retirement preconditions are stated as a conjunction with an explicit negative**: every count==1 surface implemented by MTx, a valid forced diagnostic with no forced OK→FR, and a final ordinary full-corpus measurement — *and* "no partial forced result authorizes selector retirement." The negative is what stops a future reader treating a good-looking forced run as sufficient.
- **The artifacts are labelled by the question they answer**, not just by sha: attribution artifacts versus current-main shipping artifacts, with the instruction to regenerate any forced ELF after a source or head change rather than infer from old labels. That distinction is exactly what went wrong twice this session.
- **The negative space is explicit** — the withdrawn cold precharge and why it was wrong, that the slice cleared no shipped FRs, that the forced-only 224 clearance is not a shipping claim, and that intrinsic recipient charging remains triplicated with #10545 tracking it.

**The originating-failure paragraph is the one that will make the procedure get followed.** A four-step evidence template presented as best practice gets skipped by whoever is in a hurry; the same template with "an early version staged 3000 gas into the dispatcher's execution-budget cell rather than an intrinsic accumulator, flipped 2572 fixtures, and required five refuted hypotheses before a hand-run and cell-by-cell parent control isolated the wrong destination" attached is a template people follow. The closing line — *treat provenance, isolation builds, and shipping-artifact measurement as requirements, not presentation polish* — earns its place because the paragraph above it proves the cost of the alternative.

One suggestion for whoever next edits it, not a change request: as slices land, keep the *negative space* section updated with the same care as the status. A handoff's status decays predictably and readers discount it; its record of what was deliberately not done is what stays useful, and is also the part most likely to go stale silently.

PR #10563, opened 2026-07-26
docs(codegen): bound MemoryBudgetGuard's remit to constants Lean can see

pirapira:
Closes the outstanding request from #10559's review: record, in `MemoryBudgetGuard`'s own scope header, the boundary that PR discovered in practice.

## Why

#10555 gave this file a scope header saying it is the home for **constant-relationship invariants generally**. That is too broad, and #10559 found the edge case immediately: the invariant *"one of the two dense arenas ends exactly at `__BSS_END__`, whose neighbour is unmapped space"* is a genuine coincidence needing a guard — but it **cannot** be stated here.

Every statement in this file is closed by `decide` over values Lean can *see*. An invariant involving an address the **linker** chooses — a symbol address, a section base or end, `__BSS_END__` — has no Lean-visible form. A `decide` on such a relationship would pin a number nobody had checked against the ELF, which is **worse than no guard, because it looks verified**.

So the header now states the boundary and where those invariants go instead: `scripts/check-region-map.sh`, which already reads the symbol table with `readelf` and exists to gate link-layout drift. #10559 is cited as the worked example.

**Rule of thumb recorded for the next person:** if you cannot write the invariant without naming a symbol or a section, it is an ELF-gate invariant; if both sides are Lean `Nat`s, it belongs here.

## Also: pin the property, not the instance

Carried over from #10559's review, because it applies to assertions in this file too. That gate asserts *whichever* arena ends at `__BSS_END__`, not that `evm_memory` does — so it stays silent through the deliberate layout change contemplated in #10557 and still fires on accidental `.bss` growth.

A gate pinned to the current **instance** fires on the first **intended** change; whoever is making that change reasonably weakens or removes it; and the protection disappears exactly when the layout is in motion — the moment it was most needed. Worth having next to the admission test, since the two together are what keep a guard file from either accumulating unjustifiable assertions or losing its load-bearing ones.

## Verification

- `lake build EvmAsm.Codegen.MemoryBudgetGuard` — green.
- Comment-only: zero added `def`/`theorem`/`lemma`/`instance` lines, so no proof terms change and the axiom surface is untouched.
- Small dependency cone (this file is a leaf guard), so no broad rebuild.


pirapira:
Approved — labelling `merge!`. Docs only, one file.

This closes the boundary question the scope header opened, and the phrasing of *why* is what makes it useful rather than a rule to memorise:

> a `decide` on such a relationship would pin a number nobody had checked against the ELF, which is worse than no guard because it *looks* verified

That is the right severity ordering, and it is the same principle the guard file already embodies elsewhere: an assertion that cannot fail for the reason it claims to check is worse than an absent one, because it consumes the attention that would otherwise notice the gap. A `decide` over a hand-copied link address would be exactly that — green forever, and green for the wrong reason.

**The routing rule is stated so it can be applied without judgement**, which is what a boundary in a scope header needs to be: *if you cannot write the invariant without naming a symbol or a section, it does not belong here.* That is mechanical. Someone with a new constant relationship can decide in one reading which of the two gates is theirs, without having to reconstruct the reasoning.

**And the worked example carries the distinction rather than just illustrating it.** #10559's invariant is a genuine coincidence — an arena size (a Lean constant) meeting a section layout (a linker decision) — so it *does* need a guard, and the guard just is not this one. Without that example the boundary could be misread as "link-layout facts do not need pinning," which is the opposite of what #10559 establishes.

`MemoryBudgetGuard` now has, in its own header: what it is for, an admission test (coincidence versus construction), two corollaries drawn from real mistakes, a record of its three remits, and a boundary naming where the *fourth* kind of invariant goes instead. That is a file that can be maintained by someone who was not here tonight — which was the point of writing any of it down.

Issue #10565, opened 2026-07-26
spec-divergence: the 3/word copy coefficient is hardcoded in TWO helpers and serves two independently-editable spec constants


pirapira:
## Site

`EvmAsm/Codegen/Programs/EvmMemoryGas.lean`, `copyWordGasAsm` — the per-word copy charge:

```
addi rounded, length, 31
bltu rounded, length, .exit_outofgas    # ceil32 wrap guard (good)
srli words, rounded, 5                  # words = ceil32(len)/32
slli gas,   words, 1                    # gas = 2*words
add  gas,   gas, words                  # gas = 3*words   <-- the 3 lives HERE, as arithmetic
```

## The amount is currently correct — this is about drift, not a present bug

Verified against `execution-specs` Amsterdam `vm/gas.py`:

- `OPCODE_COPY_PER_WORD = 3` (`:225`), used by MCOPY (`instructions/memory.py:162`) and CALLDATACOPY / CODECOPY / EXTCODECOPY (`instructions/environment.py:227,288,381`);
- `OPCODE_RETURNDATACOPY_PER_WORD = 3` (`:218`), used by RETURNDATACOPY (`instructions/environment.py:450`).

Both mirrored in SpecRef: `SpecRef/InstructionsCore.lean:47` and `SpecRef/InstructionsEnv.lean:49`. So `3 * ceil32(len)/32` matches the spec for every copy opcode today, and the `ceil32` wrap is already guarded (the same guard #10539 had to add to `keccakWordGasAsm`, which is worth noting: `copyWordGasAsm` was the correct sibling that fix was modelled on).

## The finding: two independently-editable spec constants are collapsed onto one hardcoded implementation path

`copyWordGasAsm` is the single helper used for **both** families — confirmed at the call sites: `EvmCalldataHandlers.lean:94`, `EvmCodeHandlers.lean:55`, `EvmExtcodecopy.lean:88` (all `OPCODE_COPY_PER_WORD` opcodes) **and** `NoopReturnData.lean:49` (`copyWordGasAsm "returndatacopy"`, an `OPCODE_RETURNDATACOPY_PER_WORD` opcode).

The spec keeps these as **two separate symbols**. They happen to be equal at Amsterdam. Nothing makes them equal:

- a fork that reprices copying but not returndata-copy (or vice versa) would leave the guest charging the *old shared* value for one family;
[...]

PR #10566, opened 2026-07-26
fix(codegen): pin the copy-per-word coefficient to both spec constants

pirapira:
Implements the guard proposed in #10565, which I filed minutes ago after verifying the coefficient against `execution-specs`.

## The finding

`copyWordGasAsm` charges `3 * ceil32(len)/32`, with the `3` synthesised as `slli words, 1` + `add` — and it is the **single** helper serving **both** copy families:

| spec constant | opcodes | call site |
|---|---|---|
| `OPCODE_COPY_PER_WORD` (`gas.py:225`) | MCOPY, CALLDATACOPY, CODECOPY, EXTCODECOPY | `EvmCalldataHandlers.lean:94`, `EvmCodeHandlers.lean:55`, `EvmExtcodecopy.lean:88` |
| `OPCODE_RETURNDATACOPY_PER_WORD` (`gas.py:218`) | RETURNDATACOPY | `NoopReturnData.lean:49` |

Both are 3 today. **The spec keeps them as two independently editable symbols and nothing enforces that they stay equal.** A fork repricing one family leaves the guest charging the old shared value for the other — silently: no build error, and a sweep would show only unexplained mispricing on one opcode family. `grep OPCODE_COPY_PER_WORD EvmAsm/Codegen` finds nothing, so there is no symbol a fork change could propagate through.

**The amount is correct today** — verified against the spec, including that `srli (len+31), 5` matches `ceil32(size)//32`. This is drift protection, not a bug fix.

## The guard

Two `decide`s, per this file's scope header (both sides are Lean `Nat`s, so it belongs here and not in the ELF gate):

- `copyPerWord_families_agree` — the load-bearing one. The docstring says explicitly that when it fails the fix is to **split or parameterise `copyWordGasAsm`**, never to edit the theorem. That instruction is the difference between a guard that protects the invariant and one that gets edited away by whoever the build stops.
- `copyPerWord_is_three` — pins the value the emitted `slli`/`add` actually implements, so a reprice of *both* families also fails loudly.

## Verification

- **Red-tested both**, since a guard never observed to fail is not yet a guard: `is_three` against `4`, and `families_agree` against `OPCODE_KECCAK256_PER_WORD` (6). Both fail the build; restored, green.
- No emitted guest bytes change — theorems only, no layout pin moves.

Stacked on #10563 because that PR edits this file's scope header; the theorems go at the end, so the two do not overlap.

Incidental observation worth recording: `copyWordGasAsm` **already carried** the `ceil32` wrap guard (`bltu rounded, length`) that #10539 had to add to `keccakWordGasAsm`. It was the correct sibling that fix was modelled on, which is a small piece of evidence that #10539 took the right shape rather than inventing one.


pirapira:
Approved — labelling `merge!`. One file, two `decide`-closed theorems, no emitted bytes.

This is the admission test from the scope header applied cleanly: **two independently editable spec symbols collapsed into one guest helper is coincidence class, so it gets a pin.** They are equal today and nothing enforces that.

**`copyPerWord_families_agree` is correctly identified as the load-bearing one**, and the instruction attached to it is what makes the guard useful rather than merely present:

> when it fails, the fix is to **split or parameterise `copyWordGasAsm`**, never to edit this theorem

That sentence is the difference between a guard that survives a reprice and one that gets "fixed" by relaxing the assertion. It is the same discipline as #10556's *do not insert padding here* — prohibiting the plausible-looking response, not just stating the invariant.

**And the failure mode you describe is the reason this is worth a theorem rather than a comment.** A fork repricing one copy family would leave the guest charging the old shared value for the other, with no build error, and an EEST sweep would show only unexplained mispricing on one opcode family — a symptom several steps removed from its cause, on a path where the natural first hypothesis would be the handler rather than a shared helper. That is precisely the shape that costs a night to diagnose; tonight's dispatch-lane slice was exactly that shape and took five refuted hypotheses.

Pinning both — the families agreeing *and* the value being 3 — is right, since a reprice of *both* families is the case the first theorem cannot see. `copyPerWord_is_three` also ties the pin to what the emitted `slli`/`add` actually implements, so the theorem tracks the instruction sequence rather than an intention about it.

Worth noting where this sits: it is the first entry on the **amounts** axis, after ordering was closed on three independent construction-class grounds. Ordering asked *when* gas is charged; this asks *how much*, and the two axes have completely different evidence shapes — the ordering answers came from typing, emission order and cell ownership, while this one needs a constant-by-constant comparison against the spec. Anyone continuing here should expect the amounts work to look nothing like the ordering work that preceded it.

Issue #10550, opened 2026-07-26
gas-step-bound: EXTCODECOPY zeroes the newly-expanded interval twice, the second time byte-wise (~48 vs ~5.3 steps/gas)


pirapira:
## The defect

`extcodecopyWitnessTail` (`Programs/EvmExtcodecopy.lean:80-108`) zeroes the **same newly-expanded interval twice**, the second time 8× less efficiently:

```
ld  x5, 488(x20)                                  # x5 = OLD msize
la  x6, ecc_old_active; sd x5, 0(x6)              # stash it
…
updateActiveMemorySizeAsm "extcodecopy" … true false
        ^^ internally zeroes [x13+old, x13+rounded) with `sd` — 8 bytes per 4 instructions
…
ld  x6, 488(x20)                                  # x6 = NEW msize
add x7, x13, x5                                   # = x13 + old
add x8, x13, x6                                   # = x13 + new
.Lrt_ecc_zero_new_mem:
  bgeu x7, x8, .Lrt_ecc_zero_new_done
  sb zero, 0(x7)                                  # 1 byte per 4 instructions
  addi x7, x7, 1
  j .Lrt_ecc_zero_new_mem
```

The interval is identical — `[x13+old_msize, x13+new_msize)` — and `updateActiveMemorySizeAsm`'s own fresh-zero loop has already covered it. Both endpoints are 32-byte multiples, so the `sd` pass covers it exactly; the byte loop finds nothing but zeros.

## Cost, in the units that matter for a zkVM

Proving cost is RISC-V steps, so a redundant pass is not free:

| pass | instructions per byte | per 32-byte word |
|---|---|---|
| `updateActiveMemorySizeAsm`'s `sd` loop | 0.5 | 16 |
| this **redundant** `sb` loop | 4 | 128 |

Memory expansion charges ≥ 3 gas per word, so:

- every other memory-touching opcode pays **≈5.3 steps/gas** for zeroing (the baseline measured in #10521);
- **EXTCODECOPY pays ≈48 steps/gas** — of which ~89% is the redundant pass.

So EXTCODECOPY is ~9× worse than the rest of the family on zeroing steps-per-gas, for no behavioural difference. Not a soundness issue — it is over-work, never under-charge, so it cannot affect FA — but it is a real proving-cost defect on a reachable path, and it is the kind of thing the step-bound lens exists to surface.

## Suggested fix

[...]

PR #10567, opened 2026-07-26
fix(codegen): pin the remaining inline dynamic-gas coefficients

pirapira:
Completes the amounts-axis audit begun in #10566, closing #10565's broader half.

## Result: every inline dynamic-gas coefficient matches execution-specs Amsterdam

I read each helper's emitted arithmetic and traced it to the spec symbol it implements. **All six are correct today**, so these pins are drift protection rather than bug fixes:

| helper | emitted arithmetic | value | spec symbol |
|---|---|---|---|
| `keccakWordGasAsm` | `slli w,2` + `add` + `add` | 6/word | `OPCODE_KECCAK256_PER_WORD` |
| `expDynamicGasAsm` | `li x7, 50; mul` | 50/byte | `OPCODE_EXP_PER_BYTE` |
| `logDynamicGasAsm` | `li x18, topics*375` | 375/topic | `OPCODE_LOG_TOPIC` |
| `logDynamicGasAsm` | `slli x5, x15, 3` | 8/byte | `OPCODE_LOG_DATA_PER_BYTE` |
| `createInitcodeGasAsm` | `li gas, 2` (no salt) | 2/word | `CODE_INIT_PER_WORD` |
| `createInitcodeGasAsm` | `li gas, 8` (salt) | 8/word | **a sum** |

Each row records the arithmetic, so a reviewer can check the pin against the code it stands for rather than taking my word that 6 is what `slli 2` + two `add`s produces.

## CREATE2's 8 is a worse collapse than #10566's

`perWordCost := if hasSalt then 8 else 2`. The spec (`vm/instructions/system.py:244-250`) charges CREATE2 `OPCODE_KECCAK256_PER_WORD * words + init_code_cost(...)` = 6 + 2 = 8, because CREATE2 additionally hashes the initcode; CREATE (`:190-193`) charges only `init_code_cost` = 2.

In #10566 two constants had to stay **equal** — and equality is at least a relation someone might notice. Here a single literal must equal a **sum**, so repricing **either** input silently mis-charges CREATE2 with nothing to compare against. `create2PerWord_is_sum` states the decomposition, so the build fails if either summand moves.

## Failure instructions, per assertion

Each docstring names the correct response: **update the helper's arithmetic, never the theorem** — and for the sum, *check which summand changed before touching the 8*. That sentence is what keeps a guard from being deleted by the person it stops, who is the one most motivated to delete it.

## Verification

- **Red-tested** the sum pin against 9: fails the build. Restored, green.
- Theorems only — no emitted guest bytes, no layout pins move.

Stacked on #10566. Note main moved to `daccd2def` while this was in progress (batch #10564 landed #10520 + #10559, moving `textSizeBytes` to `0x61440`); this PR touches no emitted bytes, so it is orthogonal to that layout change, but it will need a branch update once the stack below it lands.


pirapira:
Approved — labelling `merge!`. One file, `decide`-closed pins, no emitted bytes.

Spot-checked the coefficients against the spec rather than the summary, and they line up: `OPCODE_KECCAK256_PER_WORD = 6` (`gas.py:229`), `OPCODE_EXP_PER_BYTE = 50` (`:227`), `OPCODE_LOG_TOPIC = 375` and `OPCODE_LOG_DATA_PER_BYTE = 8` (`:232,231`), `MEMORY_PER_WORD = 3` (`:88`). The guest's LOG helper composes `topicCount * 375` into a single immediate (`EvmMemoryGas.lean:554`) with the base 375 charged by the dispatcher, which is the same split the spec makes between `OPCODE_LOG_BASE` and `OPCODE_LOG_TOPIC`.

**Tracing each helper's emitted arithmetic to the spec symbol it implements is the right unit of work here**, and it is a different one from the ordering axis. `slli w,2 + add + add` *is* ×6; `li x7, 50; mul` *is* ×50. Those are facts about the instruction sequence, so the pin tracks what the guest actually computes rather than what a docstring says it computes — which matters because the two have diverged in this codebase before (the `43 steps/gas` baseline measured on a redundant loop, #10550).

**What I would keep visible now that this axis is closed:** *coefficients* matching is not *amounts* being correct. It rules out a wrong constant; it does not rule out a wrong formula around a right constant — the wrong rounding, the wrong operand width, a missing term — and #10521 is the standing example, where the per-word coefficient was always 6 and the defect was a `ceil32` that wrapped. So the honest statement of where the mission now stands is three-layered:

- **ordering** — closed on three independent construction-class grounds (typing, emission order, cell ownership);
- **coefficients** — closed by this PR and #10566, pinned against the spec symbols;
- **formulae** — open, and the layer where every gas defect found tonight actually lived.

That last line is worth putting somewhere durable, because "gas ordering closed" plus "gas coefficients pinned" reads to a skimmer as "gas verified," and the one defect class this session actually produced sits in the layer neither of them covers.

PR #10568, opened 2026-07-26
docs(codegen): the copy coefficient has two emission sites, not one

pirapira:
Found by auditing my own claim. I told coord twice that "all six inline coefficients" were verified — and my enumeration was scoped to `Programs/EvmMemoryGas.lean`, so it never saw **`mcopyDynamicGasAsm`**, which lives in `Programs/EvmMcopyGas.lean`.

## What it does

It independently synthesises the same `3 * words` that `copyWordGasAsm` does — `srli w,5` then `slli w,1` then `add`, once per copy range (MCOPY has two). **The value is correct**: the spec charges MCOPY `OPCODE_COPY_PER_WORD * words` (`vm/instructions/memory.py:162`), and `OPCODE_COPY_PER_WORD = 3`. No mispricing, and #10567's audit conclusion stands.

## Why it still matters

**The copy coefficient is hardcoded in two places, in two files, with no shared symbol.** #10566's `copyPerWord_is_three` fires correctly on a spec reprice — it pins the *constant*, not a call site — but it does not tell the person the build stops that there are **two** sites to fix.

A guard that names one of two required edits is worse than one that names neither, because it sends the reader away believing the job is complete. The docstring now names both sites and says to fix both.

Also worth noting for the two remaining helpers I had not checked: `returnRevertMemoryGasAsm` and `callMemoryExpansionGasAsm` have **no coefficient of their own** — they delegate the charge to `updateActiveMemorySizeAsm`, whose coefficient is `MEMORY_PER_WORD`, already pinned as `memoryPerWord` since #10540. The `li x19, …ArenaLimitBytes` lines in them are arena-bound checks, not gas coefficients. So the audit is now genuinely complete rather than complete-as-far-as-one-file.

## Verification

- Documentation only: one docstring paragraph, no theorem changes, no emitted guest bytes, no layout pins move.

Stacked on #10567.


pirapira:
Approved — labelling `merge!`. Docs only, one file.

The correction is right and the note is placed where it does work: **`copyPerWord_is_three` fires once and cannot say how many call sites implement the coefficient.** A reader who sees a green pin reasonably infers one place to change; the paragraph tells them there are two, in different files, and that nothing links them.

**What makes this a real gap rather than a documentation nicety** is that the two sites synthesise the same value by *different* instruction sequences — `copyWordGasAsm` for the four copy opcodes, `mcopyDynamicGasAsm` doing `srli w,5` then `slli w,1` then `add`, twice, once per range. So a reprice does not just need two edits; the person making the second edit is reading unfamiliar arithmetic that happens to compute the same constant. That is exactly the shape of #10545 (the intrinsic recipient charge in three places) and of the drift the guard file exists to catch.

**On the failure itself:** the audit ran over `Programs/EvmMemoryGas.lean` and was described as covering "the helpers." That is the same boundary-unstated failure as the `*_end` sweep and the handler-blob extractor — an instrument with a scope the report did not carry. Worth noting this one was *smaller* and caught faster: one file versus one naming convention versus one regex window, and it produced a paragraph rather than a retraction.

The trend across the session is the useful part. The instrument-boundary error rate has not obviously fallen, but the **cost per instance** has: retraction → paragraph → caught-before-writing → caught-before-sending → this, a correcting note filed the same minute the gap was found. That is what improvement looks like on this failure mode; expecting the errors themselves to stop is not realistic when every claim is about a search over a large tree.

One suggestion, since it applies to the whole guard file now: when a pin covers a value with more than one emission site, **say the count in the theorem's own docstring** rather than in a separate paragraph. `copyPerWord_is_three` could carry "two emission sites: `EvmMemoryGas.copyWordGasAsm`, `EvmMcopyGas.mcopyDynamicGasAsm`" directly, so the reader who jumps to the failing theorem — which is how a guard is actually encountered — gets the scope without needing the surrounding prose.

Issue #10569, opened 2026-07-26
spec-divergence: the 256-entry static gas table is entirely hardcoded, with zero SpecRef references


pirapira:
## Site

`EvmAsm/Codegen/Dispatch.lean:509-558`, `staticGasCost (op : Nat) : Nat` — the per-opcode static gas table, emitted as a 256-entry `.dword` array at `:565`.

## The exact, complete claim

**`grep -c SpecRef` over lines 509-558 returns 0.** The entire table is hardcoded literals — the distinct values are `0 1 2 3 5 8 10 20 30 100 375 5000 11000` — with no reference to any `SpecRef.GasCosts.OPCODE_*` constant, even though SpecRef defines them (e.g. `OPCODE_ADD`, `OPCODE_CLZ`, `OPCODE_LOG_BASE` in `SpecRef/InstructionsCore.lean`).

This is stated as a complete claim rather than a sample because it is an *absence* over a syntactically bounded region — the definition's own line range — which is the one shape a grep answers exactly.

## What I verified, and what I did not

**Verified** — the values I could resolve to a named SpecRef constant all agree with the table:

| SpecRef | value | table |
|---|---|---|
| `OPCODE_ADD` | 3 | `0x01 => 3` ✓ |
| `OPCODE_MUL` | 5 | `0x02 => 5` ✓ |
| `OPCODE_ADDMOD` | 8 | `0x08 => 8` ✓ |
| `OPCODE_CLZ` | 5 | `0x1e => 5` ✓ |
| `OPCODE_LOG_BASE` | 375 | `0xa0..0xa4 => 375` ✓ |
| `OPCODE_ADDRESS` | 2 | `0x30 => 2` ✓ |

Plus by reading: PUSH0 = 2, PUSH1–32 / DUP / SWAP / DUPN / SWAPN / EXCHANGE = 3, EXP base = 10, KECCAK256 base = 30.

**Not verified**: the remaining byte values individually — roughly 240 of 256 entries, including the interesting ones (`20`, `100`, `5000`, `11000`, and the deliberate `0`s for halt opcodes and unwired bytes). **This is not a claim that they are wrong.** It is a claim that nothing in the tree ties any of them to the spec.

## Why the missing linkage matters independently of correctness

Same class as #10565 (the six dynamic coefficients, now pinned by #10566/#10567), but ~40× larger:

- a fork repricing any opcode leaves the dispatcher charging the old value, **silently** — no build error;
[...]

PR #10570, opened 2026-07-26
fix(codegen): pin the five gas tiers behind the static opcode table

pirapira:
The cheap high-value slice of #10569, actionable now because it needs no byte→mnemonic mapping.

## Why tiers rather than all 256 entries

`staticGasCost` (`Dispatch.lean:509-558`) prices every opcode byte with bare literals and **zero** `SpecRef` references. Pinning all 256 needs a byte→mnemonic table that does not exist in-tree — I tried to compute the set difference and my comment-parsing extraction was contaminated by section-header prose, which I recorded on the issue rather than publishing the number.

**The tiers need no mapping at all.** `BASE`, `VERY_LOW`, `LOW`, `MID`, `HIGH` are already named symbols in `SpecRef/Gas.lean`'s `GasCosts` namespace, so five `decide`s cover every range-based entry in the table:

| tier | value | table entries it backs |
|---|---|---|
| `BASE` | 2 | PUSH0 |
| `VERY_LOW` | 3 | PUSH1–32, DUP1–16, SWAP1–16, DUPN/SWAPN/EXCHANGE, the whole comparison + bitwise block |
| `LOW` | 5 | MUL, DIV, SDIV, MOD, SMOD, SIGNEXTEND, CLZ |
| `MID` | 8 | ADDMOD, MULMOD |
| `HIGH` | 10 | EXP base |

**A fork repricing `VERY_LOW` moves roughly 50 opcodes at once, and nothing currently catches it.** That is the largest single drift exposure in the table and it costs five lines to close.

## Scope, stated in the docstring rather than implied

This pins the **tier values**. It does **not** establish that each opcode is assigned the right tier — that needs the per-opcode comparison #10569 describes, which is not done. Keeping those two apart matters: "the tiers are pinned" and "the table is verified" are different claims and only the first is delivered here.

## Verification

- **Red-tested**: `VERY_LOW` against 4 fails the build. Restored, green.
- Each docstring names the failure response — update the literals in `staticGasCost`, never the theorem — and notes that the table's own trailing comments (`-- PUSH1..PUSH32 (VERYLOW)`) make the sites locatable.
- Theorems only: no emitted guest bytes, no layout pins move.

Stacked on #10568.


pirapira:

Approved — labelling `merge!`. One file, five `decide`-closed theorems, no emitted bytes.

This is the slice I suggested on #10569 and it is scoped exactly right: **the tiers need no byte→mnemonic mapping**, because they are already named symbols in `SpecRef/Gas.lean`'s `GasCosts` namespace, whereas pinning all 256 entries does need one that does not exist in-tree yet. Five theorems covering every range-based entry — `BASE` for PUSH0, `VERY_LOW` for the PUSH/DUP/SWAP/DUPN/EXCHANGE and comparison-bitwise block, `LOW` for MUL/DIV/SDIV/MOD/SMOD/SIGNEXTEND/CLZ, `MID` for ADDMOD/MULMOD, `HIGH` for EXP's base.

The leverage argument is what makes this worth doing before the per-opcode work: **a fork repricing `VERY_LOW` moves ~50 opcodes at once**, and today nothing catches that. Five theorems close the majority of the drift surface for a fraction of the effort the full table needs, and they do not block or duplicate the per-opcode pins that #10569 still tracks.

Consistent with the rest of the guard file: coincidence class (bare literals that must equal named spec symbols, with nothing enforcing it), pinned to the symbol rather than to the value.

pirapira:

PR #10571, opened 2026-07-26
fix(codegen): pin the cold/warm access decomposition to its spec totals

pirapira:
The last unpinned coefficient family from #10569's amounts audit.

## The shape

EIP-2929 access gas is charged in **two pieces** by the guest: a flat `WARM_ACCESS` debited inline, then the cold delta added by `runtime_access_account_charge` / `evm_storage_access_charge_key` when the address or slot is newly accessed. I verified this reading the delegation-resolution site:

```
ld t0, 568(x20); li t1, 100; bltu t0, t1, .exit_outofgas   # WARM_ACCESS, OOG-guarded
sub t0, t0, t1; sd t0, 568(x20)
jal ra, runtime_access_account_charge                       # + 2900 if newly cold
```

**The spec states only the totals** (`amsterdam/vm/gas.py:69-71`): `WARM_ACCESS = 100`, `COLD_ACCOUNT_ACCESS = 3000`, `COLD_STORAGE_ACCESS = 3000`. So the *split into 100 + 2900* is the guest's own invention, and the **sum** is what has to match.

## Four independently editable places

| place | value |
|---|---|
| inline `li …, 100` (28 emitted sites) | 100 |
| `runtimeAccessColdDeltaGas` (`EvmAccessGas.lean:33`) | 2900 |
| `storageAccessColdDeltaGas` (`EvmStorageAccessGas.lean:35`) | 2900 |
| spec totals | 100 / 3000 / 3000 |

**The two guest deltas being separate is correct** — the account and storage paths track *distinct* spec symbols (`COLD_ACCOUNT_ACCESS` vs `COLD_STORAGE_ACCESS`) and should be able to diverge, unlike the copy case in #10566 where one helper served two symbols. What was missing is any tie from either to its total, so a fork repricing `COLD_ACCOUNT_ACCESS` alone would leave both paths charging the old sum with no build failure.

## The pins

`warmAccess_is_hundred` fixes the inline literal's value; `accountColdDelta_completes_cold` and `storageColdDelta_completes_cold` assert each split reconstitutes its own total. Same sum-decomposition shape as `create2PerWord_is_sum` (#10567), and the same failure instruction: **fix the guest's delta constant or the inline 100, never the theorem** — and check which side of the sum the spec moved before choosing.

## Verification

- **Red-tested**: the account sum against 3001 fails the build. Restored, green.
- Theorems only: no emitted guest bytes, no layout pins move.

Stacked on #10570.

### DIFF
--- a/EvmAsm/Codegen/MemoryBudgetGuard.lean
+++ b/EvmAsm/Codegen/MemoryBudgetGuard.lean
@@ -100,6 +100,8 @@ import EvmAsm.Codegen.Programs.EvmMemoryGas
 import EvmAsm.Stateless.SpecRef.Transactions
 import EvmAsm.Stateless.SpecRef.InstructionsCore
 import EvmAsm.Stateless.SpecRef.InstructionsEnv
+import EvmAsm.Codegen.Programs.EvmAccessGas
+import EvmAsm.Codegen.Programs.EvmStorageAccessGas
 
 namespace EvmAsm.Codegen.MemoryBudgetGuard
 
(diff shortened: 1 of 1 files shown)


pirapira:
Approved — labelling `merge!`. One file, `decide`-closed pins, no emitted bytes.

Verified the spec side: `gas.py:69-71` gives `WARM_ACCESS = 100`, `COLD_ACCOUNT_ACCESS = 3000`, `COLD_STORAGE_ACCESS = 3000` — **totals only**. So the split into a flat 100 plus a 2900 delta is the guest's own decomposition, invented for its dispatch structure and named nowhere in the spec. That is exactly the case this guard file exists for: a relationship between guest-side quantities that must sum to a spec-side value, where nothing structural enforces the sum.

**The failure mode is what makes it worth a theorem, and it is nastier than a plain wrong constant.** A fork repricing `WARM_ACCESS` alone leaves `100 + 2900 = 3000` still summing correctly to the *old* cold total while the warm charge itself is now wrong — so the cold path stays right and the warm path silently drifts, on the most frequently executed gas path in the interpreter. A sum invariant checks the sum; it does not check that either addend still matches its own spec symbol. Pinning both the addends and the total is what closes that, and I take it that is why both are there.

**"Fix the emission, never the theorem"** is the right instruction and it is now consistent across every guard in this file. Worth noting the extra clause you added here — *check which side of the sum the spec moved* — which the earlier guards did not need, because with two addends and one total there are three distinct repricings and the remedy differs for each. A reader who sees the theorem fail and edits the 2900 because it is the unfamiliar number would be wrong two times in three.

**The 28 inline `li …, 100` sites are the residual worth stating plainly**, and they are the same shape as #10568's two copy-coefficient emission sites, one order of magnitude larger. The pin fires once; the emission is everywhere. That is not solvable by another theorem — it wants the literal to come from a named constant at the emission site — so it is a real follow-up rather than something this PR should have absorbed.

That closes the coefficients layer. The three-layer statement now reads: **ordering** closed on three construction-class grounds, **coefficients** pinned including this decomposition, **formulae** open — and the last of those is still where every gas defect found tonight actually lived.

pirapira:

Issue #10572, opened 2026-07-26
spec-divergence: the flat WARM_ACCESS charge is a bare literal at every emission site; the pin fires once but the fix touches many


pirapira:
## Site

The flat `WARM_ACCESS` charge, emitted inline as a bare literal at **28 sites** in `gen-out/stateless_guest.s`:

```
ld t0, 568(x20); li t1, 100; bltu t0, t1, .exit_outofgas
sub t0, t0, t1; sd t0, 568(x20)
```

## The gap

#10571 pinned the *relationship* — `WARM_ACCESS + coldDelta = COLD_{ACCOUNT,STORAGE}_ACCESS`, with both addends and the totals tied to SpecRef. **The pin fires once; the emission is everywhere.** No additional theorem fixes that: a fork repricing `WARM_ACCESS` would fail `warmAccess_is_hundred` correctly, and then the fixer has to find and edit **28 separate `li` sites**, any one of which left at 100 is a silently mispriced access on the most frequently executed gas path in the interpreter.

This is #10568's "two emission sites" shape an order of magnitude larger. There, the remedy was a docstring naming both sites, which is adequate for two and inadequate for twenty-eight.

## What it wants

The literal should come from a **named constant at the emission site**, so there is one place to change:

- a Lean `def warmAccessGas : Nat := SpecRef.GasCosts.WARM_ACCESS` (or reuse the existing pinned symbol directly), and
- the emitters producing `li …, ` ++ `toString warmAccessGas` rather than a hardcoded `100`.

That makes it **construction-class**: a reprice propagates to every site by recompilation, and #10571's pin becomes a redundancy check rather than the only line of defence.

## Scope

**Verified:** 28 occurrences of a `li` of 100 in the emitted image, and that #10571's pins tie `WARM_ACCESS` and both cold deltas to their spec symbols.

**Not verified:** that all 28 are the access-gas charge specifically — `100` is a plausible literal for other purposes, and I did not classify each site. The count is an upper bound on the sites needing change and a lower bound on nothing. Anyone taking this should enumerate them by their surrounding pattern (the `ld 568(x20)` / `bltu` / `sub` / `sd` sequence) rather than by the literal alone, which is the same distinction that made the difference elsewhere in this family: **match the construct, not the value.**

[...]

PR #10573, opened 2026-07-26
docs(plan): record the guard inventory and the ELF gate

pirapira:
CLAUDE.md requires PLAN.md track the session's work. Its Infrastructure section described `MemoryBudgetGuard` as it stood at #10540; the file has since grown to **seven guards over three remits** plus a documented scope boundary, and a **second gate** now exists that PLAN.md did not mention at all.

## Three changes

**The MemoryBudgetGuard entry** now records its three remits (#10522's unreachability precondition, the achievable `k` behind the top theorem's fuel per #10552, the clamped fill loop's exit exactness per #10554) and the four rules its header carries: the **admission test** (coincidence needs a pin, construction doesn't and a pin there is ceremony), guard only the coincidental half *and say why the other needs nothing*, pin the quantity that is **used** not its inputs, and the #10563 **boundary** — constants Lean can *see* only, since a `decide` over a link-chosen address would pin a number nobody had checked against the image, which is worse than no guard because it looks verified.

**A new gas-coefficient-pins entry**, recording that all were **correct when pinned** — drift protection, not bug fixes — and naming the three coincidence-class collapses found: copy's `3` serves two spec constants that must stay *equal* (#10566) **and is emitted from two files** (#10568); CREATE2's `8` is a **sum** (#10567); and the access split `100 + 2900` must reconstitute `COLD_{ACCOUNT,STORAGE}_ACCESS`, where a `WARM_ACCESS` reprice alone leaves the sum right and the warm path silently wrong (#10571). The two residuals (#10569, #10572) are named rather than implied closed.

**A new link-layout ELF gate entry** (#10559), including *why* it is written as a **property** — "whichever arena ends at `__BSS_END__`" — rather than an instance: so it survives the intentional reorder contemplated in #10557 instead of fighting it. Paired with the zero-slack layout invariant from #10556, and carrying the correction that the gap above the arena is **unclaimed, not trapping** — it lies inside the verified RAM window, so a store there is legal and silent.

## Why the corrections are in the record and not just the fixes

Two of tonight's conclusions were wrong before they were right — the gap was described as trapping, and the residual was described as needing a large port. A PLAN.md entry that records only the endpoint teaches nothing; one that records the constraint ("unclaimed, not trapping") stops the next reader rebuilding an argument on the same false premise.

Documentation only — no Lean, no scripts, no emitted bytes.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -199,6 +199,44 @@ All deleted spec files have been recreated. See **Pending: Recreate Deleted Spec
   configuration. **Ordering constraint**: if it ever fails, the sparse path goes
   live, #10522's write becomes reachable, and the global 4096-entry sparse cap
   becomes a reachable FR — see #10535.
+
+  **Grown since #10540 into the home for constant-relationship invariants
+  generally** — seven guards over three remits: #10522's unreachability
+  precondition, the achievable steps-per-gas constant `k` behind the top
+  theorem's fuel (#10552), and the clamped fill loop's exit exactness
+  (`clampEnd_alignment_*`, #10554), plus the gas-coefficient pins below. Its
+  header records the **admission test** (coincidence needs a kernel-checked pin;
+  construction does not, and a pin there is redundant ceremony), the rule to
+  guard only the coincidental half of a two-half bound *and say why the other
+  needs nothing*, the rule to pin the quantity that is **used** rather than its
+  inputs, and — added in #10563 — the **boundary**: constants Lean can *see*
+  only. An invariant involving a link-chosen address belongs in the ELF gate
+  below, because a `decide` on one would pin a number nobody had checked against
+  the image, which is worse than no guard because it *looks* verified.
+- **Gas-coefficient pins** (`MemoryBudgetGuard.lean`, GH #10565–#10567, #10570,
+  #10571): every inline dynamic-gas coefficient and static tier tied to its
+  SpecRef symbol by a red-tested `decide`, each carrying the instruction for what
+  to fix when it fires (*update the helper's arithmetic, never the theorem*). All
+  were **correct when pinned** — this is drift protection. Three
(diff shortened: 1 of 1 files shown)


pirapira:
Approved — labelling `merge!`. `PLAN.md` only.

This is the item `CLAUDE.md` actually asks for, and it does the thing a plan entry is *for*: it records what a reader needs in order to decide, not a list of what happened.

Three properties worth naming, since they are not the default for a status doc:

**It says the guards were correct when pinned.** "All were correct when pinned — this is drift protection" is the sentence that stops the whole guard inventory reading as a list of defects found. Without it, a reader skimming seven guards over three remits reasonably concludes the gas layer was riddled with wrong constants. It was not; the pins are prophylactic, and the three coincidence-class collapses found along the way are the actual findings.

**It carries the admission test rather than the conclusions.** Recording *coincidence needs a pin, construction does not* — plus guard-only-the-coincidental-half and pin-what-is-used-not-its-inputs — means the next person adding a constant relationship can decide for themselves rather than pattern-matching against seven examples. And the #10563 boundary is the load-bearing addition: a `decide` on a link-chosen address pins a number nobody checked against the image, **which is worse than no guard because it looks verified**. That sentence is the one I would least want lost.

**It states the remedy per guard.** *Update the helper's arithmetic, never the theorem* is what makes these survive a reprice instead of being relaxed by whoever hits the red build at an inconvenient hour.

One thing to keep true as this section ages: the inventory is accurate today and will drift the moment someone adds an eighth guard without updating it. The file's own header already asks each new remit to be added to its list — worth having `PLAN.md` point at that convention rather than duplicating the count, so the guard file stays the single source and this entry stays a pointer. Otherwise the two disagree in a month and the reader cannot tell which is stale.

Filing the PLAN.md item you had half-done, rather than declaring the protocol satisfied by #10553, is the right instinct — that PR recorded the probe-only scope, not the guard inventory, and those are different obligations.

Issue #10576, opened 2026-07-26
Proposed removal: the 690 codegen-zisk oracle probes have no failure signal, and most state no spec expectation — with a keep criterion and a 10-probe sample


pirapira:
> **Note:** This issue was written by Claude Code.

## Origin

This is filed at the maintainer's steer — *"consider removing the oracle probes, they were written for temporal needs"* — not as an independent proposal. The work below is the part I can contribute to that decision: a **keep-versus-delete criterion**, and a small hand-classified sample showing the criterion actually discriminates rather than merely sounding plausible.

**I am not proposing which files to remove, and nothing has been deleted.**

## The structural fact — nothing executes these, and that is deliberate

**No `codegen-zisk-*` probe is run by anything.** Not CI, not a wrapper reachable from CI, not a build target.

| glob | count |
|---|---|
| `scripts/codegen-zisk-*-check.sh` | 687 |
| `scripts/codegen-zisk-*.sh` | 690 |
| `scripts/codegen-zisk-*.sh` containing a `ziskemu` invocation | 688 |
| all `scripts/*.sh` containing a `ziskemu` invocation | 724 |

I use **690** throughout. (A parallel count of 691 was reached independently; the difference is a glob boundary, not a disagreement.)

I did not establish the zero by grepping a guessed name — a zero for a guessed name is indistinguishable from a zero because the thing does not exist. I enumerated every `scripts/` reference in `.github/workflows/` and followed each candidate to its text:

- **`labeler.yml:8,62`** — the only `codegen-` reference in `.github/`, and it is an **exclusion**: `codegen-*.sh` is filtered out of the PR size-label line count, with a comment that "a bulk ~500-script ziskemu regen is not a large risky diff."
- **`check-build-parallel.sh:22`** (run from `build.yml:95`) — invokes `codegen-stateless-link-check.sh --no-build`, whose own header states it "emits, assembles, and links `stateless_guest.elf`, but **does not run ziskemu**". It contains no ziskemu invocation.
- **`check-duplication.sh:9`** — a prose comment. **`check-statement-tamper.sh:110`** — a path-exclusion pattern.
- **`codegen-core-opcode-frontier-check.sh:130-131`** — the only genuine wrapper, invoking 2 of the 690. It is itself not referenced from `.github/`.
- No `Makefile`, `justfile`, or lakefile target references them.

[...]

Issue #10574, opened 2026-07-26
spec-divergence: the SSTORE clear-refund 12480 is a bare literal at two sites, unpinned to SpecRef's derived REFUND_STORAGE_CLEAR


pirapira:
## Site

`EvmAsm/Codegen/Programs/SstoreGasRefund.lean:107` and `:111` — the SSTORE clear-refund amount, emitted as a bare literal:

```
"  li t0, 12480\n"
```

Two emission sites; 2 occurrences in the linked image.

## The value is correct — this is the deepest derivation chain in the gas surface

`REFUND_STORAGE_CLEAR` is not a spec literal. It is **derived from three quantities** (`amsterdam/vm/gas.py:92-94`):

```python
REFUND_STORAGE_CLEAR: Final[int] = int(
    (STORAGE_WRITE + COLD_STORAGE_ACCESS) * Uint(4800) // Uint(5000)
)
```

With `STORAGE_WRITE = 10000` (`:74`) and `COLD_STORAGE_ACCESS = 3000` (`:71`): `(10000 + 3000) * 4800 // 5000 = 12480`. The guest's `12480` matches. ✓

## What is exemplary, and worth copying rather than changing

**SpecRef already does this correctly**, and it is the pattern the rest of the gas surface should follow:

```lean
SpecRef/Gas.lean:81   def REFUND_STORAGE_CLEAR : Nat := …          -- DERIVED, not a literal
SpecRef/Gas.lean:229  #guard GasCosts.REFUND_STORAGE_CLEAR == 12480 -- non-vacuity pin on the derivation
```

Derivation *plus* a build-time pin that the derivation still yields the expected value — construction-class where it can be, checked where it cannot. That is strictly better practice than the codegen literals #10565–#10571 had to pin after the fact.

## The gap

**Nothing connects codegen's `12480` to SpecRef's derived constant.** So:

- a reprice of `STORAGE_WRITE` alone silently invalidates the guest's literal — and **Amsterdam already did exactly that**, halving `STORAGE_WRITE` from the historical 20000 to 10000, which moves `REFUND_STORAGE_CLEAR` from 22080 to 12480. The identical change in a future fork would leave the guest refunding the old amount;
- likewise a reprice of `COLD_STORAGE_ACCESS` (which #10571 pins for the *access* path but not for this derivation), or a change to the `4800/5000` ratio;
[...]

PR #10577, opened 2026-07-26
fix(codegen): pin the SSTORE refund literals to their SpecRef constants — with the live/dead/probe-only site census

pirapira:
Closes the pin half of #10574. **Stacked on #10571** (`fix/pin-access-cold-delta`) — append-only to `MemoryBudgetGuard.lean`, so it retargets itself as the base lands, like the four-member cascade that merged earlier.

## What lands

Two `decide` theorems, both **zero-axiom** (verified via the LSP, not inferred):

```lean
theorem sstoreRefundClear_is_12480  : SpecRef.GasCosts.REFUND_STORAGE_CLEAR = 12480
theorem sstoreStorageWrite_is_10000 : SpecRef.GasCosts.STORAGE_WRITE        = 10000
```

`REFUND_STORAGE_CLEAR` is the deepest derivation in the gas surface — the spec does not state it as a number. `amsterdam/vm/gas.py:92-94` computes `(STORAGE_WRITE + COLD_STORAGE_ACCESS) * 4800 // 5000 = 12480`; the guest emits the result as a bare literal, with nothing tying the two together. `SpecRef/Gas.lean:229` guards the SpecRef *port* against the literal, but nothing guarded the *guest*.

**Both values are correct today.** These are drift protection, not bug fixes.

## The part worth reviewing: the site census in the docstring

The issue's recommended next step names three sites, all in `SstoreGasRefund.lean`. That is incomplete in the direction that produces a half-fix. `STORAGE_WRITE` has **five emissions across three files**, and they are not equivalent:

| site | destination | status |
|---|---|---|
| `Programs/Storage.lean:159` | debits the gas cell `568(x20)` | **LIVE — actually charges gas** |
| `Programs/Storage.lean:168` | debits the gas cell `568(x20)` | **LIVE — actually charges gas** |
| `Programs/SstoreGasRefund.lean:115` | `add s7` (refund accumulator) | **LIVE** |
| `Programs/SstoreGasRefund.lean:97` | `add s6` (gas accumulator) | **DEAD** |
| `Programs/SstoreRegularGas.lean:74` | returns in `a0` | **PROBE-ONLY** |

The two sites that actually charge `STORAGE_WRITE` are in a file the issue never mentions. Someone following the issue would pin, edit the refund helper, close it — and leave the charging literals bare, with the issue reading as resolved. The docstring labels every site so a reader the guard fires on knows where to edit; five bare line numbers would not have.

A sixth `li …, 10000` at `EvmAccessGas.lean:553` is a gas-budget **seed** written to `568(x20)` in a seed routine — same number, different quantity. Excluded on reading its context, not on the absence of a comment.

`:97` is dead because `sstore_gas_refund_outcome`'s only production caller (`Storage.lean:443-453`) reads offsets 8 and 32 and never offset 0. Every reader of `srfd_out` in the tree was checked. It is listed anyway so the census does not go stale silently.

## Deliberately absent: a theorem for the derivation

An earlier draft of this PR had one, and it violated this file's own admission test. `SpecRef/Gas.lean:81-82` **defines** `REFUND_STORAGE_CLEAR` as `(STORAGE_WRITE + COLD_STORAGE_ACCESS) * 4800 / 5000`, so a theorem asserting that equality restates the definition and can never fail — construction, not coincidence, and worse than absent because a reader would take it for a live check on the derivation. What *is* coincidental is that the derivation's output equals the guest's literal; that is `sstoreRefundClear_is_12480`. The docstring says so, per the header's instruction to guard only the coincidental half and state why the other needs nothing.

## Gates

**Run:** `lake build` clean; `check-forbidden-tactics.sh` OK; `check-no-warnings.sh` 0 warnings; `check-file-size.sh` OK; both theorems verified zero-axiom.

**Not run:** `scripts/check-axioms.sh` — it cannot execute under `LAKE_ARTIFACT_CACHE` (#10537). Stated rather than claimed. The zero-axiom result above came from the LSP instead.

No emitted bytes change — this PR adds theorems only, so no A/B is required.

## Not in scope

Routing the five sites through the pinned constants is the follow-up; this PR pins the values and records where the edits go. The separately-filed #10576 covers the `+100` in `SstoreGasRefund.lean:83` (a real but currently-dead over-count, which needs a full-corpus A/B) and the oracle-probe corpus.

### DIFF
--- a/EvmAsm/Codegen/MemoryBudgetGuard.lean
+++ b/EvmAsm/Codegen/MemoryBudgetGuard.lean
@@ -399,4 +399,99 @@ theorem storageColdDelta_completes_cold :
       = EvmAsm.Stateless.SpecRef.GasCosts.COLD_STORAGE_ACCESS := by
   decide
 
+/-! ## Guard 8 — the SSTORE refund literals (GH #10574)
+
+`REFUND_STORAGE_CLEAR` is the deepest derivation in the gas surface: the spec
+does not state it as a number at all. `amsterdam/vm/gas.py:92-94` computes it as
+`(STORAGE_WRITE + COLD_STORAGE_ACCESS) * 4800 // 5000`, i.e.
+`(10000 + 3000) * 4800 // 5000 = 12480`. The guest emits the **result** as a bare
+literal. Both values are **correct today** — these pins are drift protection.
+
+Three properties make this a worse coincidence than the flat coefficients above:
+
+* **two upstream inputs**, either of which moves it — and `COLD_STORAGE_ACCESS`
+  is already pinned separately by `storageColdDelta_completes_cold`, which stays
+  green through a reprice while silently invalidating this refund, because
+  neither guard knows the second is downstream of the first;
+* **the derivation is not reconstructable from the literal** — a `4800/5000`
+  ratio inside an integer truncation gives a reader updating `STORAGE_WRITE` no
+  cue that a refund constant three files away depends on it;
+* **`STORAGE_WRITE` breaks both literals at once, by different amounts** — the
+  restore refund directly, and the clear refund through the derivation. A fixer
+  who updates the obvious one and not the derived one leaves the refund counter
+  wrong in the harder-to-notice direction.
+
+Amsterdam already performed exactly this reprice, halving `STORAGE_WRITE` from
+20000 to 10000 and moving `REFUND_STORAGE_CLEAR` from 22080 to 12480.
+
+`SpecRef/Gas.lean:229` already guards `REFUND_STORAGE_CLEAR == 12480`, but that
+ties the **SpecRef port** to the literal; nothing tied the **guest's** emissions
+to it. These two theorems close that.
+
+### Why there is no theorem for the derivation itself
+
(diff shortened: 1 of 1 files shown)


pirapira:
Superseded by #10579, which substitutes the constants at the emission sites instead of pinning them.

Closing rather than reworking, because review established the theorems did not say what their docstring claimed. `REFUND_STORAGE_CLEAR = 12480` mentions no guest emission — neither theorem here did — so a wrong guest literal would have passed both, and the guest linkage lived entirely in the prose census. One of the two also duplicated an existing tripwire at `SpecRef/Gas.lean:229`.

#10579 removes the mechanism instead: the emitted text is interpolated from the SpecRef symbol, so the guest cannot hold a value that disagrees with it, and the relationship becomes construction class — which by this file's own admission test needs no pin. For 12480 the guest now inherits the *derivation* rather than the number, which retires the cross-file hazard #10574 identified.

Verified byte-identical emission (`.s`, `.o` and linked `.elf` all unchanged), so no sweep is needed.

The five-site live/dead/probe-only census from this PR is preserved in #10579's description, where it serves as the completeness evidence.

Issue #1440, opened 2026-04-30
Drain scripts/unimported-allow.txt to empty


pirapira:
## Context

PR #1439 (closes #1209) adds CI enforcement that every \`.lean\` file under \`EvmAsm/\` must be transitively imported from \`EvmAsm.lean\`, with a one-time grandfather list at \`scripts/unimported-allow.txt\` to keep the initial PR green. The allow-list currently has **14 entries** in two clusters:

\`\`\`
# DivMod V4 work-in-progress cluster (issue #1337)
EvmAsm.Evm64.DivMod.Compose.FullPathN3LoopUnified
EvmAsm.Evm64.DivMod.LoopDefs.IterV4
EvmAsm.Evm64.DivMod.LoopDefs.IterV4Invariants
EvmAsm.Evm64.DivMod.LoopDefs.IterV4InvariantsPhase2
EvmAsm.Evm64.DivMod.SpecCallAddbackBeq
EvmAsm.Evm64.DivMod.SpecCallAddbackBeq.AlgDefs
EvmAsm.Evm64.DivMod.SpecCallAddbackBeq.AlgEuclideans
EvmAsm.Evm64.DivMod.SpecCallAddbackBeq.NumericalTests
EvmAsm.Evm64.DivMod.SpecCallAddbackBeq.NumericalTestsV4
EvmAsm.Evm64.DivMod.SpecCallV4

# EvmWordArith cluster — pure-arithmetic helpers used by DivMod V4
EvmAsm.Evm64.EvmWordArith
EvmAsm.Evm64.EvmWordArith.AddbackPinning
EvmAsm.Evm64.EvmWordArith.Div128NoWrapDischarge
EvmAsm.Evm64.EvmWordArith.Div128PhaseNoWrap
\`\`\`

There is no good reason to keep files on \`main\` that are not guarded by the rest of CI (no-sorry check, file-size guardrail, downstream proof reuse, etc.). Every file in the allow-list is effectively dead code from the rest of the build's perspective.

## Goal

Drive \`scripts/unimported-allow.txt\` to **zero entries** and then delete the file (the script tolerates a missing allow-list).

For each entry, exactly one of:

[...]

PR #10578, opened 2026-07-26
docs(agents): record what the six untracked orphan SAsm files are

pirapira:
Every working checkout shows six untracked files under `EvmAsm/Codegen/Programs/`, and `scripts/check-unimported.sh` fails locally because of them while CI stays green. They were investigated once for #10574's session; this records the result so the next session **audits rather than re-derives**. Filed at coord's request as its own small PR.

## Three facts, cheap to state and expensive to re-establish

- **Never tracked on any branch** — `git log --all --diff-filter=A` is empty for all six.
- **Committing any of them turns CI red immediately.** `check-unimported.sh` has had no allow-list since #1440 and refuses an escape hatch. That is a second reason #10544 had to be reverted, independent of the 4.6 MB binary — and it means leaving them in place costs nothing.
- **Their targets ARE landed** (#10358, #10360, `3f0dac58b`) **but the files are not copies** — 4-of-65 and 0-of-31 declaration-name overlap, the shared names being scaffolding. A different decomposition, not a stale duplicate.

## Two corrections to the obvious readings, both traps

**Orphan status does not keep a file out of the build.** `check-unimported.sh`'s own header states lake compiles everything under the library directory; what orphan status prevents is downstream `import`. That is why these cannot reach the emitted guest — not because the build skips them.

**The newer-or-older `.olean` test does not work in this tree.** All six mtimes cluster within 20 ms, which is the #10544 revert's bulk restore, not six authoring times. The `.olean`'s *location* is informative where its timestamp is not: it is the only `.olean` anywhere under `EvmAsm/`, and lake never writes there.

## The page states its own limits

Two open ends are recorded rather than papered over: live-versus-abandoned is **unsettled** (a ~3500-line reading question nobody has done), and the files have **not** been verified to still compile against current main, which is the real rot test and costs a build.

The page also says plainly: do not delete them, and if they do turn out to be abandoned, propose removal as its own PR. Ownership is a maintainer question that no amount of further inference settles.

Docs only — no code, no emitted bytes, no gates affected.

### DIFF
--- a/docs/agents/untracked-orphan-sasm-files.md
+++ b/docs/agents/untracked-orphan-sasm-files.md
@@ -1,0 +1,97 @@
+# The six untracked SAsm files in working checkouts
+
+**Load when:** a `git status` in this repo shows untracked `*SAsm.lean` files
+under `EvmAsm/Codegen/Programs/`, or `scripts/check-unimported.sh` fails locally
+while CI is green.
+
+**Do not delete them, and do not `git add -A`.** They are not abandoned as far as
+anyone has established, and they are not ours to remove. See "What is still
+open" at the bottom before acting on them.
+
+This page exists so the next session does not re-run the investigation. It was
+written after auditing them once; the evidence for every claim is stated so you
+can disagree with any of it cheaply.
+
+## The set
+
+Six files, all untracked, all under `EvmAsm/Codegen/Programs/`:
+
+| file | size | state |
+|---|---|---|
+| `HeaderExtractLogsBloomSAsm.lean` | 1450 lines, 34 theorems | substantial, 0 `sorry` |
+| `ReceiptExtractLogsBloomSAsm.lean` | 1497 lines, 37 theorems | substantial, 0 `sorry` |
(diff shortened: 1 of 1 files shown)


pirapira:
Closing on the maintainer's call: this is general git/working-copy hygiene rather than something about this repository, so it should not land as a repo doc.

The reasoning generalises, and it is consistent with a rule already applied here (server-local CPU contention was likewise kept off GitHub): **untracked files are local checkout state, not repository state.** A page in `docs/agents/` describes the project to everyone who clones it; six files that exist only in some working copies do not.

Nothing in the investigation is lost. It is preserved outside the repo, and the two facts that matter for anyone who trips over this are already enforced or recorded elsewhere:

- `scripts/check-unimported.sh` has had no allow-list since #1440 and refuses an escape hatch, so **committing any of these files turns CI red immediately** — the guard is mechanical and needs no documentation to work.
- The targets are landed (#10358, #10360, `3f0dac58b`); the files are a different decomposition, not stale copies; and whether they are live or abandoned is unsettled.

Standing guidance unchanged: **do not delete them**, and if they are ever confirmed abandoned, propose removal as its own PR.

Written by Claude Code (coord), acting on the maintainer's decision.

PR #10581, opened 2026-07-26
fix(codegen): delete sstore_gas_refund_outcome's dead gas accumulator

pirapira:
**Do not label — a full-corpus A/B is running and this must not merge before it reports.** Stacked on #10579. Addresses the repair item in #10576.

## What and why

The spec computes SSTORE's gas and refund in **one** function with **one** accumulator (`amsterdam/vm/instructions/storage.py:66-130`): `gas_cost` starts at zero, the access cost is added once cold-or-warm, `STORAGE_WRITE` is added on the first-change condition, and the `refund_counter` updates sit in the same function driven by the same three-way comparison of original/current/new.

The guest split that across two routines, and this one **re-derived the gas cost into a second accumulator no caller ever reads**. `Storage.lean:443-453` reads `srfd_out` at `+8` and `+32` and never `+0`; every reader of `srfd_out` in the tree was checked. So the guest held two independent copies of one spec quantity, one live and one dead.

The dead copy had already drifted — it initialised `s6` to `WARM_ACCESS` and then added the *full* access cost, over-counting every result by exactly 100. Its own probe expects the spec figures and fails all eight cases by that margin.

**Deleted rather than repaired.** Repairing the arithmetic leaves both copies editable and free to diverge again, and next time the dead one might be the copy someone wires up. Same doctrine as #10579: remove the mechanism, do not correct it.

## The risk in this change is the output layout, and it is verified

The caller reads `+8` and `+32` by **hardcoded displacement**. Remove the field, let the others slide, and those loads silently read the wrong things with no build error anywhere. So the five-field footprint is kept and `+0` is now a constant zero (`sd x0, 0(a4)`).

**How that was verified — using the diff itself as the instrument.** The emitted `.s` diff is *by construction* the complete set of changes, so a property of the diff is a proof about the code rather than a sample of it. Grepping the diff for `8(a4)`, `16(a4)`, `24(a4)`, `32(a4)` and `srfd_out` returns **nothing** — which proves those are outside the change rather than merely suggesting it. `la x14, srfd_out` appears twice in both legs.

The whole diff is **22 lines in a 78,000-line image**, all inside `sstore_gas_refund_outcome`: 16 removed (the `s6` init, the cold/warm branch and both arms, the clean-changing block and its two empty fall-through labels), and one changed (`sd s6, 0(a4)` → `sd x0, 0(a4)`).

`a3` (the warm flag) becomes unused and is retained so the caller's argument setup and the probe ABI do not change.

## Layout regeneration

The deletion shrinks `.text` by `0x38`, shifting every symbol after the routine. Both generated artifacts are **regenerated, not hand-edited**:

- `scripts/asm-fixtures/symbol-addresses.tsv` via `scripts/gen-symbol-addresses.py --build` (3047 rows)
- `RegionMap.textSizeBytes` `0x61440` → `0x61408`

`check-region-map.sh` now reports "region map matches the linked ELF". The regeneration was then **re-checked against the A/B pair** in case it moved the image — it did not, ELF sha256 `0dd6a6c8…` before and after, so the candidate leg is not stale.

## A/B

| leg | ELF sha256 |
|---|---|
| base (#10579 tip) | `00100fcae1099942edb8064b7e2d6906a8925ff61de4ccbbe6d77ca8a28d98fb` |
| candidate | `0dd6a6c8d3604a2c157cedc5e404fada63b63cc27c1976e771a1966d8373b72b` |

Both emitted from **committed refs** with a clean tree verified **per leg**, under the **same output name**, then copied apart and `chmod 444`. Full corpus, spike, 26,104 cases per leg. Prediction on record before the run: **zero diff on every row**, since the deleted output is read by nobody — which makes a single differing case a finding rather than noise.

Results will be posted with the scored row count.

## Consequence to accept

This **retires** `codegen-zisk-sstore-gas-refund-outcome-check.sh` rather than turning it green, because the probe validates an output that should not exist. Noted in #10576 so the record does not later read as though the oracle was forgotten.

## Gates

**Run:** build clean; `check-region-map` OK; `check-layering` OK; `check-forbidden-tactics` OK; `check-no-warnings` 0; `check-unimported` 0 orphans; `check-no-hardcoded-guest-pc` OK; `check-file-size` OK.

**Not run:** `scripts/check-axioms.sh` — cannot execute under `LAKE_ARTIFACT_CACHE` (#10537).


pirapira:

**What this changes.** `sstore_gas_refund_outcome` re-derived a gas cost that `sstoreValueTransitionGasAsm` already computes — two independent copies of one spec quantity, one live and one dead. `execution-specs` keeps gas and refund in a single `sstore` with a single `gas_cost` accumulator (`amsterdam/vm/instructions/storage.py:66-130`), so the duplication is the defect and the observed +100 (the accumulator initialised to `WARM_ACCESS` and then adding the full access cost) was only what a dead copy looks like once it drifts. Deleting the copy is the fix; repairing its arithmetic would have left both copies independently editable.

**The output layout is preserved, and that was the whole risk.** `Storage.lean:443-453` reads `srfd_out` at offsets 8 and 32 by hardcoded displacement, so sliding the fields down would have silently produced a wrong refund with no build error. Verified using the diff itself as the instrument — grepping the emitted `.s` diff for `8(a4)`, `16(a4)`, `24(a4)`, `32(a4)` and `srfd_out` returns nothing, which *proves* those are outside the change rather than suggesting it, since the diff is by construction the complete set of changes. 22 changed lines in a 78,000-line image; offset 0 is now a constant zero.

**Evidence, stated for what it is.** The gate here was a **partial** corpus A/B of 11,088 of 26,104 (the intersection of base 11,104 and candidate 11,088) plus a random 200 of 26,104 (`--random --seed 4242`, drawn from the full manifest) — **not a full-corpus run**. Zero differing rows on full output bytes and zero on the succ bit, reported separately; every scored row OK on both legs. Stopped early with maintainer authorisation on the strength of a zero-effect prediction recorded *before* the run. The sequential partial is depth in the first 42% of the manifest and manifest order tracks fixture family, so the random 200 is the orthogonal breadth half — it failed to falsify the zero rather than confirming an effect it was never expected to find.

**One consequence to note:** removing the field retires `codegen-zisk-sstore-gas-refund-outcome-check.sh` (which expected the spec numbers and was red on all eight cases) rather than turning it green. See #10576.

pirapira:
Closing: **content landed on main** via batch merge #10589 (`32450c568`).

Verified by content rather than by status — a batch merge does not auto-close all its members, and a batched branch can read CLEAN, DIRTY or BEHIND indefinitely, so neither the PR state nor a tip-to-tip diff against main is a reliable signal. Confirmed with `git merge-base --is-ancestor <branch head> origin/main`.

Issue #10582, opened 2026-07-26
61 codegen-eest-* probes now error on a missing --backend, by design — choose per probe when you next need one, do NOT bulk-repair


pirapira:
> **Note:** This issue was written by Claude Code.

**This is not a work queue. Do not bulk-repair these scripts.** It is a note for whoever next reaches for one of them and finds it erroring.

## What changed

`scripts/codegen-eest-stateless-check.sh` no longer has a default backend (implements the ruling recorded on #10533) or a default run scope (the maintainer's 2026-07-26 ruling). Omitting either is now a hard error whose message carries the when-to-use-which guidance.

**61 invocation sites, across 61 files in the `codegen-eest-*` family, passed no `--backend`.** They relied on the silent `ziskemu` fallback, and no caller anywhere sets `EEST_BACKEND` (checked across `scripts/`, `docs/` and `.github/`; the only mentions are documentation of the variable in `scripts/spike/README.md`, not setters). So each of those 61 now stops with the backend error until someone says what it should use.

**That is deliberate and it is the point of the change**, not collateral damage.

## Why there is no list of 62 fixes to work through

The obvious remedy — append `--backend ziskemu` everywhere to preserve behaviour — **is not behaviour-preserving.** #10533 also asks that ziskemu be clamped to `--jobs 1`, so those scripts would go from their current parallelism to serial. There is no mechanical edit here; every option is semantic:

- **A** — `--backend ziskemu` on all of them: keeps the instrument, halves throughput or worse.
- **B** — `--backend spike` on all of them: faster, probably right for most verdict-level probes, and **silently changes what 61 scripts measure** in one unreviewed sweep.
- **D** — decide per probe: correct, and expensive up front.

The maintainer's answer supersedes all three: **repair them as we need them, and it will be clearer which backend we want in that context.**

That is better than any of A/B/D because *the information needed to choose does not exist yet*. These scripts express no intent — grepping the family for `ziskemu`, `spike`, `oracle` or `accelerators` finds nothing in them. **They never chose a backend; they inherited one.** So there is nothing to recover from the source, and any bulk answer would be an invention. Deferring the choice to the moment of use puts it where the context that determines it is actually present — and several of these will never be run again, so that cost is never paid at all.

This converts 61 silent inherited choices into 61 deferred explicit ones. It is the same shape as the fix-it-when-we-use-it call on #10532.

## How to choose, when your turn comes

[...]

PR #10584, opened 2026-07-26
harness: make A/B scope and backend explicit, and refuse a vacuous comparison (closes #10533)

### LINKED ISSUE #10533: harness: codegen-eest-stateless-check.sh should have NO default backend — omitting --backend must be a hard error

pirapira:
Four changes, one theme: **the A/B harness should not silently choose for you, and should refuse a comparison that cannot mean anything.** Each came out of a real failure today; none is speculative.

## 1. No default backend — closes #10533

**#10533 was decided and never implemented.** The maintainer ruled on it, explicitly rejecting a flipped default because that "only relocates the same footgun" — but the tree still had `BACKEND="${EEST_BACKEND:-ziskemu}"`, and omitting `--backend` passed straight through. Cited here as *a decision recorded on an open issue*, not as shipped precedent.

**`EEST_BACKEND` survives** as an opt-in override, per the issue; only the fallback when *neither* the flag nor the variable is set is gone.

The error text is **the maintainer's, extracted from the issue body programmatically rather than retyped** — verbatim-by-construction rather than verbatim-by-care. Two clauses appended: the env-var note, and a pointer for readers who arrived via a sibling probe rather than this script.

## 2. No default scope

`--limit N` or `--all` must be chosen. The message names `--all` as the full 26,104-case corpus — required for a **high-blast-radius** change (a gas constant, a shared helper, anywhere path-targeting cannot be trusted) and for re-baselining — and `--limit N` as the focused mode: known-failing, plus path-touching, plus **a random control drawn from the passing set**. That control is flagged as not optional, because a set built only from known-failing and path-touching cases is blind in the OK→FR direction by construction.

An earlier revision of this PR said `--all` is *the merge gate for anything that changes emitted bytes*. That was true when written and is now too strong, and as written it would have told a future reader that #10581 — which changes emitted bytes and was gated on a partial plus a random sample — was illegitimate. Corrected in `7ff27f4f8`. The general lesson: the backend message ages well because it describes the two **backends**, which have not changed; the sentence that aged badly described **our practice**, which did. Prefer describing the tool over describing the workflow.

The message also carries the one sentence that explains why the flag exists:

> a smoke run reports honestly over its N cases and reads exactly like a corpus pass

That is not hypothetical. I launched today's A/B without a scope flag, got a clean result in 17 seconds, and nearly reported it — the default is 50. What caught it was going to look up the scored row count in order to quote a denominator.

## 3. `--backend ziskemu` clamps `--jobs` to 1, loudly

Warning plus clamp rather than an error, as #10533 asks: `--jobs` has a legitimate auto-derived value, so the point is that the caller is *told*.

**Design note — the split was considered and deliberately not taken.** A finer behaviour is available: hard-error when `--jobs > 1` was passed *explicitly* (the caller stated a preference, so silently overruling it is the same defect this PR is about) while silently clamping when `--jobs` resolved from `auto` (no preference to override). I did not implement it, for two reasons. The issue asks for "a warning plus an automatic clamp", and the warning already tells the caller in both cases, which is the property it says matters. And distinguishing explicit from auto needs a second `*_SET` sentinel on a flag that also accepts the literal `auto`, which is more machinery than the difference earns. Recorded so a future reader can see the simpler behaviour was chosen rather than overlooked.

## 4. Self-check 0 — refuse a vacuous comparison

`eest-ab-compare.py` now hashes each run dir's `stateless_guest.elf` and **refuses a verdict if they match**. Comparing a build to itself yields a flawless zero that means nothing, and the way it happens is silent: `git checkout` carries uncommitted changes across branches, so building "base" after switching can pick up the candidate edit that was never committed. **That happened today** — a 16-instruction deletion produced two byte-identical ELFs.

It has to be mechanised rather than remembered, because **zero diff is the predicted result for most refactors**: the harness bug is indistinguishable from the hypothesis by looking at the output.

When `--no-build` + a `GUEST_ELF` override puts the ELF outside the run dir, the check **prints NOT RUN** rather than passing silently — a self-check that quietly does not apply is worse than none.

The docstring also records two hazards that cannot be mechanised there: the **linker embeds the object filename**, so emitting to `-o base` and `-o cand` makes two ELFs differ even when the change is byte-neutral (this produced a false defect today, one step from spending a sweep slot); and **do not probe a tool by running it while it is in use** — confirming a default behaviourally started a real ziskemu run against a live sweep when reading the source had already answered it.

## Also: the global baseline write is removed

Both legs of a parallel A/B raced to write `gen-out/eest-baseline.txt` and the second writer silently won, so a reader got the other leg's numbers. The per-run copy is already authoritative. **No code reads the global one** — every sibling script reads its own `"$RUN_DIR/eest-baseline.txt"`; the two doc references are updated.

## Caller audit

70 invocation sites across 66 files, reconstructed across backslash continuations (a same-line match reports *zero* sites with a scope flag — the tell that continuations are involved). Two apparent callers were case-statement path patterns, established by reading them.

- **Scope flag: zero callers break.** One casualty, a bare invocation in `docs/eest-stateless-testing.md` whose heading read "Run the default smoke subset" — naming the default being removed. Fixed here.
- **Backend flag: 61 sites now error, deliberately.** Tracked in **#10582**, which is explicitly *not* a work queue: these probes express no backend intent anywhere, so any bulk answer would be an invention. The maintainer's call is to choose per probe at the moment someone needs it, when the context that determines the answer exists.
- **Zero CI callers** — the only `.github` reference is `labeler.yml` matching the path for size-labelling, followed and read rather than assumed absent.
- **Dockerfile unaffected**: its `CMD` supplies `--all`.

A finding worth flagging from that audit: **55 of the 61 sites run the ziskemu oracle at `--jobs 2`.** #10533 reads as though its footgun fired once, on its author. It is the standing configuration of the family.

## Controls

Against the patched script: neither flag → **both** errors; `--all` alone → backend only; `EEST_BACKEND=spike --all` → neither; `--backend spike` alone → scope only; `--backend bogus --all` → neither of these, original invalid-value error preserved.

Self-check 0: same build → refuses, exit 1; distinct builds → reports distinct, does not fire; ELF absent → prints NOT RUN.

Jobs clamp exercised on a copy (ziskemu+8 warns and clamps, ziskemu+1 silent, spike+8 silent); the live file is byte-identical to that copy except the baseline removal 1500 lines away, so those transfer.

**Both missing flags report together** rather than fail-fast. An early draft exited on the scope error, so a caller missing both learned about scope, fixed it, re-ran, and only then learned about backend. Fail-fast is wrong for *independent* validation errors — it makes the user's cost proportional to how many things they got wrong when it could be constant.

### DIFF
--- a/docs/eest-stateless-testing.md
+++ b/docs/eest-stateless-testing.md
@@ -45,11 +45,16 @@ Override that with `EEST_FIXTURES_DIR=/path/to/fixtures` when needed. Use
 
 ## Common Commands
 
-Run the default smoke subset:
+Run a smoke subset (there is no default scope — `--limit N` or `--all` is
+required, and `--backend` likewise):
 
 ```bash
-scripts/codegen-eest-stateless-check.sh
+scripts/codegen-eest-stateless-check.sh --backend spike --limit 50
 ```
+
+`--all` runs the full 26,104-case corpus and is the merge gate for anything
+that changes emitted bytes. A smoke run reports honestly over its N cases and
+reads exactly like a corpus pass, which is why the scope has to be stated.
 
 Run a focused fixture subset:
 
@@ -701,7 +706,6 @@ Important files:
 - `<case>.result.tsv`: per-case harness status and output hex.
 - `stateless_guest.{s,o,elf}`: guest artifacts for this invocation.
 - `eest-baseline.txt`: run summary for this invocation.
-- `gen-out/eest-baseline.txt`: copy of the latest harness summary.
 
 The summary reports:
 
--- a/docs/eest-stateless-testing.md
+++ b/docs/eest-stateless-testing.md
@@ -52,9 +52,22 @@ required, and `--backend` likewise):
 scripts/codegen-eest-stateless-check.sh --backend spike --limit 50
 ```
 
-`--all` runs the full 26,104-case corpus and is the merge gate for anything
-that changes emitted bytes. A smoke run reports honestly over its N cases and
-reads exactly like a corpus pass, which is why the scope has to be stated.
+`--all` runs the full 26,104-case corpus. It is required for a **high-blast-radius**
+change — a gas constant, a shared helper, anywhere path-targeting cannot be
+trusted — and for re-baselining after `main` moves. The full sweep is expensive
+and most of its passing cases re-confirm what is already known, so for a
+**targeted** change prefer a focused set: known-failing cases, plus fixtures
+touching the changed path, plus a random control drawn from the *passing* set
+with a fixed seed. That control is not optional — a set built only from
+known-failing and path-touching cases is blind in the OK→FR direction by
+construction.
+
+**State the scope with every number you report.** A subset run reports honestly
+over its N cases and reads exactly like a corpus pass. A sequential partial is
+depth in *manifest order*, and manifest order tracks fixture family, so it says
+much about the families it reached and nothing about the rest; a random sample
+is the orthogonal breadth half. A number without its scope is the same defect as
+a number without its denominator.
 
 Run a focused fixture subset:
 
--- a/scripts/codegen-eest-stateless-check.sh
+++ b/scripts/codegen-eest-stateless-check.sh
@@ -297,15 +297,19 @@ if [[ "$ALL" -eq 0 && "$LIMIT_SET" -eq 0 ]]; then
   cat >&2 <<'SCOPE_ERR'
 error: a run scope is required (no default).
 
-  --all         the full 26104-case corpus.  This is the MERGE GATE: any change
-                that alters emitted bytes needs a full-corpus A/B before it can
-                land.  Use --backend spike; it is parallel-tolerant.
-  --limit N     a smoke subset of N cases, for iteration.  CANNOT support a
-                corpus verdict.
+  --all         the full 26104-case corpus.  Required for a HIGH-BLAST-RADIUS
+                change -- a gas constant, a shared helper, anywhere you cannot
+                trust path-targeting -- and for re-baselining after main moves.
+                Use --backend spike; it is parallel-tolerant.
+  --limit N     a subset of N cases.  Use for iteration, and for a FOCUSED run
+                on a targeted change: known-failing cases, plus fixtures
+                touching the changed path, plus a random control drawn from the
+                PASSING set (a focused set built only from known-failing and
+                path-touching cases is blind in the OK->FR direction).
 
-Pick deliberately: a smoke run reports honestly over its N cases and reads
-exactly like a corpus pass, so an unscoped run is how a 50-case result gets
-mistaken for a 26104-case one.
+Pick deliberately, and state the scope with every number you report.  A subset
+run reports honestly over its N cases and reads exactly like a corpus pass, so
+an unscoped run is how a 50-case result gets mistaken for a 26104-case one.
 
 (If you invoked a wrapper script rather than this one, add the flag to that
 wrapper's own invocation.)
--- a/scripts/codegen-eest-stateless-check.sh
+++ b/scripts/codegen-eest-stateless-check.sh
@@ -143,13 +143,20 @@ export MALLOC_ARENA_MAX
 ALL=0
 SKIP=0
 LIMIT=50
+# No default scope: --limit N or --all must be chosen explicitly (see the
+# hard error below).  LIMIT keeps a value so the smoke path is unchanged
+# once the flag IS passed; LIMIT_SET records whether it was.
+LIMIT_SET=0
 FILTER=""
 # Default step cap. ziskemu stops at the guest's halt, so this only bounds
 # runaway/very-large runs; a case that halts earlier consumes only the steps it
 # needs. The EIP-8037 state_gas_reservoir max-gas fixture can require more than
 # the default on current ziskemu builds, so high-gas BUDGET rows get one larger
 # retry before they are reported as budget exhaustion.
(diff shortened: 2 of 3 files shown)


pirapira:
Reviewed and labelled `merge!`. Re-read the two text fixes pushed after my first pass; both are correct.

**Implements the ruling recorded on #10533** (no default backend — omitting it is a hard error carrying the when-to-use-which guidance), and the maintainer's 2026-07-26 ruling that a run scope must be stated explicitly. `EEST_BACKEND` survives as the opt-in override the issue asks for; only the both-absent fallback is removed.

Three things worth noting for reviewers:

- **The error text is the maintainer's own, extracted programmatically from #10533 rather than retyped** — verbatim by construction, not by care. Two clauses appended: the `EEST_BACKEND` line and a paragraph for someone who arrived via a sibling probe.
- **Both validation errors are collected and reported together** rather than failing fast. Fail-fast is wrong for *independent* requirements: it makes the caller's cost proportional to how many things they got wrong when it could be constant. It earns its keep only when a later check depends on an earlier one passing, which is not the case for backend and scope.
- **`--all` is described as required for high-blast-radius work and re-baselining, not for "anything that changes emitted bytes."** The earlier phrasing would have declared #10581 — which changes emitted bytes and was gated on a partial plus a random sample — illegitimate, leaving a rule in the tree three PRs from its own counterexample.

The general lesson behind that last fix is worth keeping: the part of this PR that will age well is the error message, because it describes **the two backends**, which have not changed. The sentence that aged badly described **our practice**, which did. Prefer describing the tool over describing the workflow.

`--jobs` clamps to 1 with a loud warning on ziskemu rather than hard-erroring on an explicitly-passed `--jobs > 1`; the body states that the finer split was considered and why it was not taken, so the simpler behaviour is a recorded decision rather than an oversight.

Also lands Self-check 0 in `eest-ab-compare.py` (refuse a verdict when both run dirs hash to the same ELF — the check that would have caught a botched A/B pair automatically), the harness-invariance notes that could not be mechanised, and removes the global `gen-out/eest-baseline.txt` write that two concurrent legs race on. Follow-up for the 61 now-erroring sibling probes is #10582, deliberately framed as *not* a work queue.

PR #10586, opened 2026-07-26
fix(codegen): source the 256-entry static gas table from SpecRef (closes the pin half of #10569)

pirapira:
`staticGasCost` held 256 entries of hardcoded literals with — as #10569 states exactly — **zero SpecRef references**. Every entry that has a SpecRef symbol now names it.

## This does more than pin: it verifies

#10569 was careful that ~240 of 256 entries were **unverified** — *"nothing in the tree ties any of them to the spec"* — and a later comment noted that a pin *"makes future drift loud, it does not certify the present values"*.

**Substitution certifies them as a side effect.** I compared every SpecRef symbol's value against the literal it replaces *before* editing, all agreed, and the emitted table is byte-identical afterwards. So the table is now known **correct against the spec symbols**, not merely guarded — which was the open half of the issue.

## Coverage, against the 57/15/17 survey on the issue

| group | treatment |
|---|---|
| 57 direct `OPCODE_<MNEMONIC>` | named |
| 15 `_BASE` variants (KECCAK256, LOG, MLOAD, MSTORE, MSTORE8, MCOPY, EXP, CALLDATACOPY, CODECOPY, RETURNDATACOPY) | matched by name |
| 10 of the 17 "absent" | **name `WARM_ACCESS`** — their static entry *is* the warm floor: BALANCE, EXTCODESIZE, EXTCODECOPY, EXTCODEHASH, SLOAD, SSTORE, CALL, CALLCODE, DELEGATECALL, STATICCALL |
| SELFBALANCE | names the `LOW` tier |

The "17 genuinely absent" shrinks to **3**: `CREATE`/`CREATE2` (11000) and `SELFDESTRUCT` (5000) have no SpecRef symbol. Each keeps its literal with the reason stated in the docstring. Halt and unwired bytes stay `0` by design — there is no static charge to name, so there is nothing to pin.

## Byte-identical emission

Emitted `.s` sha256 `cd57510f413dded94225564233c061d9c9d330d4c19bcf6f75f856a50131b398` with the change **stashed** and **applied**, both legs emitted under the **same output name**.

That equality is the substantive check: **58 symbol substitutions producing the identical 256-entry `.dword` table** is what establishes each mapping was right. A single wrong symbol — say `COLD_STORAGE_ACCESS` where `WARM_ACCESS` belonged — would move a byte.

## What this does not close

The issue's remaining audit item is untouched and still open: **whether each of the access-priced opcodes actually reaches an access charge.** An opcode that *should* be access-priced but carries a warm-floor static entry with no cold-delta call would be undercharged. That is a control-flow question, not a constant comparison, and naming `WARM_ACCESS` here makes the claim legible without verifying it.

## Gates

**Run:** build clean; `check-forbidden-tactics` OK; `check-no-warnings` 0; `check-unimported` 0 orphans; `check-layering` OK (L1 makes Codegen a sink, so Codegen → SpecRef is the permitted direction); `check-file-size` OK; `check-no-hardcoded-guest-pc` OK; `check-opcode-structure` OK.

**Not run:** `check-axioms` (cannot execute under `LAKE_ARTIFACT_CACHE`, #10537). `check-asm-to-program` and `check-region-map` are unaffected — emitted bytes unchanged, so no layout regen.

### DIFF
--- a/EvmAsm/Codegen/Dispatch.lean
+++ b/EvmAsm/Codegen/Dispatch.lean
@@ -60,6 +60,7 @@ import EvmAsm.Codegen.Programs.CallFrameSwitch
 import EvmAsm.Codegen.Programs.CallFrameDescend
 import EvmAsm.Codegen.Programs.CallFrameReturn
 import EvmAsm.Codegen.Programs.StaticContext
+import EvmAsm.Stateless.SpecRef.InstructionsEnv
 
 namespace EvmAsm.Codegen
 
@@ -485,6 +486,7 @@ def emitJumpTable (registry : List OpcodeHandlerSpec) : String :=
   "opcode_handlers:\n" ++
   String.intercalate "\n" entries
 
+open EvmAsm.Stateless.SpecRef.GasCosts in
 /-- M30 (gas metering, first slice): the **static base** gas cost of each
     EVM opcode byte, used by the dispatch loop to charge gas per instruction.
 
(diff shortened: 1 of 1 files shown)


pirapira:
Reviewed and labelled `merge!`. Byte-identical emission (same output name both legs), so no A/B is required.

**This verifies rather than merely pins, which was the open half of #10569.** The issue noted that ~240 of 256 entries were unverified — nothing in the tree tied any of them to the spec — and that a pin makes future drift loud without certifying the present values. Substitution certifies them as a side effect: **58 symbol substitutions producing a byte-identical 256-entry dword table** means every symbol's value agreed with the literal it replaced, because one wrong symbol moves a byte.

**The "genuinely absent" set shrinks from 17 to 3.** Ten of the seventeen are the access-priced family — BALANCE, EXTCODESIZE, EXTCODECOPY, EXTCODEHASH, SLOAD, SSTORE and the four CALL variants — whose static entry *is* the warm floor, so they now name `WARM_ACCESS` (confirmed against `environment.py:72`, `storage.py:84-92`, and `system.py:431/554/731`). That leaves CREATE, CREATE2 and SELFDESTRUCT as the only bare literals, each with its reason in the docstring. The earlier survey called those ten absent because it was looking for an `OPCODE_*` symbol per mnemonic; the right symbol existed, just not named after the opcode.

**One symbol was initially wrong, and the reason is worth recording** because no gate in this repo could have caught it. SELFBALANCE was substituted as `LOW` — taken from the guest's own comment, which encoded the *tier the value belongs to* rather than *the symbol the spec charges*. `environment.py:510-524` charges `GasCosts.FAST_STEP`. Both are 5 (`SpecRef/Gas.lean:65` and `:80`), so **byte-identity cannot discriminate between them**: the table came out identical on the wrong symbol, and the wrong one bakes in a coincidental tie that breaks the first time a fork reprices one and not the other. Now `FAST_STEP` with the spec line cited.

The completeness check is what makes that a result rather than a patch: **SELFBALANCE is the only opcode in the fork charging `FAST_STEP`, and `charge_gas` with a bare tier has zero occurrences across the instruction modules** — so this was the single bare-tier use in the table and the single error of its class.

**Residual, explicitly not closed here:** whether each access-priced opcode actually *reaches* an access charge. An opcode carrying a warm-floor static entry with no cold-delta call would be undercharged. Naming `WARM_ACCESS` makes that claim legible without verifying it, and the verification is a control-flow question rather than a constant comparison — arguably more urgent now that the table asserts the intent in symbol form. Tracked on #10569.

**General rule this establishes:** a green byte-identity result is necessary but *not sufficient* for a substitution PR. The residual risk is always symbol choice among equal-valued symbols, and the only check is a spec read.

Issue #10549, opened 2026-07-26
harness: the summary never reports how many fixtures matched --filter, so coverage is unverifiable from the output even when nothing was truncated


pirapira:
## The defect

`--filter SUBSTR` is applied **together with a default `--limit 50`** (`scripts/codegen-eest-stateless-check.sh:145`, `LIMIT=50`), and `--all` is a separate opt-in (`:143`, `ALL=0`). So a filtered run silently takes **at most 50** matching fixtures out of however many match — and then reports a summary that reads like complete coverage of the filter.

Same family as #10533: a default that quietly makes a decision the caller didn't know was being made, in the direction that makes a result look better than it is.

## Why it is easy to misread as coverage

The selection line does print the limit, but it prints it as a *setting* rather than as a *shortfall*:

```
selection:   limit=50, filter=keccak, input-parity
selected:    3
total:       3
... full match: 3
```

Read quickly, `filter=keccak` plus `full match: 3` says "all keccak fixtures pass". What it actually says is "of the keccak fixtures, up to 50 were considered; 3 matched the filter". Those coincide **only because 3 < 50** — which the operator has no way to see from the output. Had 200 keccak fixtures matched, the identical-looking summary would have covered a quarter of them and said nothing about it.

I hit exactly this: I ran `--filter keccak` while validating the Spike setup for #10521, got `3/3 full match`, and had to be told the result was setup validation rather than coverage. It was — but only by luck of the count.

## Suggested fix

Any of these, in preference order:

1. **`--filter` implies `--all` unless `--limit` is given explicitly.** Filtering already expresses "the subset I care about"; silently sampling that subset serves nobody.
2. Or make truncation **loud**: when the filter matches more than `LIMIT`, print `selection: filter matched N fixtures, TRUNCATED to LIMIT` — so the shortfall appears as a shortfall. Silent truncation reads as complete coverage; that is the general rule this project keeps re-learning (#10536, #10538).
3. At minimum, report `matched=N selected=M` as two distinct numbers whenever they differ, so the summary is self-inconsistent when truncation happened rather than merely incomplete.

Option 3 is the same instrument that caught a lossy join in `scripts/eest-ab-compare.py` (#10546): printing the matched count *next to* the selected count makes truncation visible as an internal contradiction rather than requiring the reader to already suspect it.

## Related

- #10533 — the sibling default footgun in the same script (no default backend; ziskemu jobs clamp)
[...]

Issue #10527, opened 2026-07-25
spec-divergence: three unused chargeBls12*GasAsm helpers lack the spec's length gates and read the discount table at index -1 for k=0


pirapira:
## Site

Three unused BLS12-381 gas helpers in `EvmAsm/Codegen/Programs/PrecompileRuntime.lean`:

- `chargeBls12G1MsmGasAsm` (`:526-557`)
- `chargeBls12G2MsmGasAsm` (`:559-590`)
- `chargeBls12PairingGasAsm` (`:592-605`)

Confirmed unreferenced repo-wide (excluding `.lake`, `execution-specs`) — the only occurrence of any of the three names outside its own definition is a docstring cross-reference at `:494`. The live BLS handlers use `bls12MsmCostAsm` + `bn254ChargeGateAsm` instead (`Programs/ChildFrameHandlerTails.lean:968`, `:1128`, and the inlined pairing cost at `:1205-1215`).

## Reference

`execution-specs` (pinned `e5a8caf1b`):
- `amsterdam/vm/precompiled_contracts/bls12_381/bls12_381_g1.py:91-103` — `if len(data) == 0 or len(data) % LENGTH_PER_PAIR != 0: raise InvalidParameter`, **then** `k = len//160`, discount lookup, `charge_gas`.
- `bls12_381_g2.py` — same shape with `LENGTH_PER_PAIR = 288`.
- `bls12_381_pairing.py:40-46` — `if len(data) == 0 or len(data) % 384 != 0: raise InvalidParameter`, then `gas_cost = 32600*k + 37700`, `charge_gas`.

## Finding

The **live** path is correct — I verified it and it matches the spec on all three counts:

```
ld   x18, inSize
beqz x18, .L<tag>_bn254_fail_allot        -- len == 0 rejected
li   x16, 160
remu x17, x18, x16
bnez x17, .L<tag>_bn254_fail_allot        -- len % 160 != 0 rejected
<bls12MsmCostAsm>                          -- cost, then
<bn254ChargeGateAsm>                       -- allotment gate + charge, then the kernel
```

[...]

PR #10590, opened 2026-07-26
perf(codegen): drop EXTCODECOPY's redundant byte-wise re-zero (closes #10550)

### LINKED ISSUE #10550: gas-step-bound: EXTCODECOPY zeroes the newly-expanded interval twice, the second time byte-wise (~48 vs ~5.3 steps/gas)

pirapira:
`extcodecopyWitnessTail` zeroed the newly-expanded memory interval **twice** — once inside `updateActiveMemorySizeAsm` with `sd` (0.5 instructions/byte), then again byte-wise with `sb` (4/byte). Per #10550 the second pass is ~89% of EXTCODECOPY's zeroing steps and makes it ~9× worse than the rest of the memory-touching family on steps-per-gas, for no behavioural difference.

## Verified redundant before deleting

#10550 asked for one check — *whether anything between the two passes can write into `[old, new)`* — and I did that plus the half it assumed:

- **Nothing writes into the interval.** Everything between the passes is `ld` of the two msizes and two `add`s computing endpoints.
- **The first pass really does cover it.** `updateActiveMemorySizeAsm`'s fresh-zero loop runs `x13+current` → `x13+rounded` with `sd`, and it stores `rounded` into `activeMemorySize` **only after** that loop — so the "new" msize the byte loop read back *is* the bound the `sd` loop had just zeroed to. Same interval **by construction**, not by coincidence of the values.
- **Both early-outs are correct.** `beqz lengthReg` and `bgeu current, rounded` each mean no newly-expanded interval, so neither pass had work.

## The issue's suggested fix was wrong, and this is the part to review

> Delete the second loop, **and with it the `ecc_old_active` stash**

**`ecc_old_active` is kept.** It has a second consumer at `EvmExtcodecopy.lean:224` — the same-block-code tail-zero, which clamps its start up to the pre-expansion msize.

There were **two** stashes of the old msize and only one was redundant:

| stash | consumer | verdict |
|---|---|---|
| `la x6, ecc_old_active; sd x5, 0(x6)` | tail-zero at `:224` | **kept** |
| stack push/pop pair | only the deleted loop | removed |

The issue conflated them because they are adjacent and store the same value. That is the `cf. #10527` caution #10550 itself raised, landing on the accompanying stash rather than on the loop. The commit message records why, so the next reader does not re-delete it.

## Emitted diff

**13 lines deleted from the `.s`, nothing else** in a 78,000-line image: the stack push/pop pair, the `ld`/`add`/`add` endpoint setup, and the six-line loop with its two labels. `.text` shrinks by `0x2c`.

Layout regenerated in **all three steps**: `symbol-addresses.tsv` (3047 rows), `GuestAddrs.lean` (1124 symbols), `RegionMap.textSizeBytes` `0x61440 → 0x61414`. `check-region-map` matches the linked ELF.

## Gate — focused, not full corpus

Scoped by the maintainer: a random sample plus the fixtures that touch the changed path. Four legs, **run serially**, both ELFs built from committed refs with a clean tree checked per leg, emitted under the same output name, `chmod 444`, and **asserted distinct before launch**:

```
BASE 06346cfca6a58bb7d622e86edb923feb9559064cd03c5fa0ef0db387ae548320
CAND 0ef7048aa97418cd6d73af2732424f875b50290f1fdcaa7d67f1e54392e8e6c4
```

| leg | scored | comparable denominator | differing rows (full bytes) | differing (succ bit) |
|---|---|---|---|---|
| random 200, base | 200 | — | — | — |
| random 200, candidate | 200 | **200** | **0** | **0** |
| `--filter extcodecopy --all`, base | 47 | — | — | — |
| `--filter extcodecopy --all`, candidate | 47 | **47** | **0** | **0** |

Every scored row is `OK` on all four legs. The comparable denominator equals the scored count on each, so no leg is short.

**Prediction was recorded before the run:** zero differing rows, because the deleted loop wrote only zeros over an already-zeroed interval, so nothing observable could move.

**The path-touching leg uses `--all`, not a limit** — deliberately, since its purpose is completeness over the changed path and a cap would reintroduce the coverage-vs-limit ambiguity of #10549. Both filtered legs log **zero truncation warnings**, which is the positive control confirming that.

**Scope — corrected after merge, and the correction reduces the claim.**

I originally wrote that the 200 were sampled from the full manifest. **They were not.** `--random` is *not* passed to the converter (`conv_args` carries only `--skip`, `--limit`, `--filter` and the verify flags), so the converter walks fixtures in order, stops at `--limit`, and writes a manifest of the **first 200**; the harness then shuffles *those*. Confirmed empirically: the converter's five-digit label prefixes in the run dir run `00000`–`00199`, not spread across 26,104.

So `--random --limit N` shuffles the **execution order** of the first N fixtures; it does not sample N from the corpus. I read the help text — *"shuffle fixtures into a random order before `--limit`"* — and took the documentation for the mechanism.

**What this gate actually covered:** all **47** blocks matching `--filter extcodecopy` (complete over the changed path, and the load-bearing half), plus **200 cases from the front of the manifest** — which is *not* orthogonal breadth, since manifest order tracks fixture family. Zero differing rows across both, but the second leg adds less than I claimed.

The path-touching leg is unaffected by this and remains complete over the changed path. Tracked separately as a harness defect.

## Gates

**Run:** build clean; `check-region-map` OK; `check-forbidden-tactics` OK; `check-no-warnings` 0; `check-unimported` 0 orphans; `check-no-hardcoded-guest-pc` OK.

**Not run:** `check-axioms` (`LAKE_ARTIFACT_CACHE`, #10537).


pirapira:
Reviewed and labelled `merge!`. Queues for the next batch rather than merging now, per the freeze while #10589 is in CI.

**The redundancy argument is airtight and it is airtight for the right reason.** #10550 asked only whether anything between the two passes could write into `[old, new)`. The stronger half is that the *first* pass provably covers the interval: `updateActiveMemorySizeAsm` stores `rounded` into `activeMemorySize` **only after** its `sd` zero-loop, so the "new" msize the byte loop reads back *is* the bound the `sd` loop just zeroed to. Same interval **by construction**, not by coincidence of the values — which is the distinction that makes this a deletion rather than a bet.

**It also declines part of the fix #10550 asked for, correctly.** The issue said to delete the `ecc_old_active` stash along with the loop. That stash has a **second consumer** on the same-block-code path, clamping the tail-zero start to the pre-expansion msize. There are two stashes of the old msize — a global one that must stay and a stack copy that fed only the redundant loop — and the issue conflated them because they are adjacent and store the same value. Following it as written would have deleted a live dependency. The reason is in the commit message, so the next reader does not re-delete it.

That is precisely the cf-#10527 caution #10550 itself raised, landing on the accompanying stash rather than on the loop.

**Layout regen complete in all three steps** — `symbol-addresses.tsv` (3047 rows), `GuestAddrs.lean` (1124 symbols), `RegionMap.textSizeBytes` 0x61440 → 0x61414, `.text` shrinking 0x2c for thirteen instructions, `check-region-map` matching the linked ELF. Worth noting because the third step was missed on #10581 and cost a CI cycle.

**Gate, and what it does and does not cover.** Not a full-corpus run, by the maintainer's focusing ruling. Four legs, serial, from committed refs with the tree checked clean per leg, same output name, immutable at 444, and the two ELFs asserted distinct before launch:

| leg | denominator | diffs (output bytes) | diffs (succ bit) |
|---|---|---|---|
| random-200 (`--seed 4242`) | 200 of 200, intersection 200 | 0 | 0 |
| path-touching (`--filter extcodecopy --all`) | 47 of 47, intersection 47 | 0 | 0 |

Comparable denominator equals scored count on every leg, so there is no intersection shortfall. **Zero truncation warnings on the filtered legs** is the positive control that the completeness leg was actually complete — `--all` rather than a limit was deliberate, since capping the leg whose job is completeness would reintroduce the coverage-vs-cap ambiguity of #10549. Note 47 *rows*, not the 30 fixture files that match the filter; those are different units.

The prediction — zero differing rows, because the deleted loop wrote only zeros over an already-zeroed interval — was recorded **before** the run, with a single differing row as the stated falsifier. It is therefore supported rather than assumed.

Issue #10580, opened 2026-07-26
MemoryBudgetGuard's header lists a guard that does not exist (the #10552 fuel constant k), and its remit count is stale


pirapira:
> **Note:** This issue was written by Claude Code.

`MemoryBudgetGuard.lean`'s header asks each new remit to add itself to the list at the top, so a reader can see what the file covers without reading 400 lines. That list has drifted, and one entry is not merely out of date — **it names a guard that does not exist.**

## The phantom entry

The header (`:8-15`) states the file has taken on *three distinct remits*, and enumerates them:

1. a guard on #10522's unreachability precondition — **exists** (Guards 1 and 2);
2. *"a guard on the achievable steps-per-gas constant `k` that the top theorem's fuel is derived from (#10552)"* — **does not exist**;
3. a guard on the clamped fill loop's exit exactness (`clampEnd_alignment_*`) — **exists** (Guard 3).

There is no theorem about `k`, steps-per-gas, or fuel anywhere in the file. The full declaration list is 34 names; none of them concerns it.

**Why this is worse than a stale count, and specifically dangerous here:** the file *does* contain a prominent `k`. Guard 2's Cauchy–Schwarz argument (`:148-152`) reasons over `k ≤ maxCallDepth` frames — a completely different quantity that happens to share the letter. So a reader who takes the header at its word, searches for `k`, and lands in Guard 2 has a plausible-looking confirmation that the fuel constant is pinned. It is not. #10552 remains open precisely because that constant is *not* pinned to anything.

That is the phantom-entry failure mode: a reader cannot tell an absent guard from one they failed to find, and here the near-miss actively supplies a wrong answer.

## The stale count

Second and much smaller: "three distinct remits" against the actual contents.

| ref | numbered `## Guard N` sections | header says |
|---|---|---|
| `main` today | 5 | three |
| `main` + #10570 | 6 | three |
| + #10571 | 7 | three |
| + #10577 | 8 | three |

Guards 4–8 (#10565, #10569, #10574) are all constant-pinning remits and none of them added itself to the list, including mine in #10577 — I left the header alone deliberately, because #10570 and #10571 are in flight and a header edit would collide with them and break the append-only property that lets that stack land without conflict resolution. That is the right call per-PR and the wrong outcome in aggregate, which is why this wants a single cleanup rather than a rule everyone re-litigates.

## Interaction with the PLAN.md pointer (#10573)

[...]

PR #10591, opened 2026-07-26
chore(mtx): route non-empty blocks through MTx

pirapira:
## Change

Route every non-empty block through the MTx runtime selector.  Only the
degenerate zero-transaction block keeps the existing single/empty route:
`beqz t0, .Lbv_singletx`.

This replaces both overlapping old guards (`count == 1` and `count < 2`), so a
half-flip cannot leave count-one blocks on the legacy route.  The old legacy
implementation is deliberately retained; its deletion is a separately
attributable follow-up.

## Evidence and new baseline

The emitted candidate is SHA-256
`4c2a785a932c8947c538323332ebe3a09311d0c84e72b385f6e7391be2dfc240`
(`.text = 0x61430`, `.data = 0x5370`).  It is byte-identical to the completed
full-corpus MTx re-census artifact.  That run, over all 26,104 unique manifest
labels, measured:

| FA | FAULT | FR | OK |
| ---: | ---: | ---: | ---: |
| 0 | 0 | 936 | 25,168 |

Thus FR=936, FA=0, FAULT=0 is the new baseline after the selector flip; the FR
increase from the former 68 baseline is the positive control that count-one
blocks now actually enter MTx.  The named count-one control `00029` also
changes from the legacy parent result `rc=0, succ=1, word112=0` to the MTx
candidate result `rc=0, succ=0, word112=47`.

These full-corpus figures come from the prior completed re-census of the exact
candidate SHA, not from a newly launched PR sweep.  A later parent
re-measurement was intentionally stopped at 12,500/26,104 once this identity
was confirmed, and is not evidence.

The corpus FA result is limited to its oracle corpus: it finds no false
acceptance there, but cannot establish behavior on attacker-crafted blocks not
represented by EEST.  The crafted-block direction remains covered by the
implementation's construction and subsequent focused work.

## Coherence

Regenerated `GuestAddrs` and symbol TSV from the linked ELF; updated
`RegionMap.textSizeBytes` to the linked `0x61430` size.


pirapira:
Reviewed and labelled `merge!`. Queues for the next batch, per the freeze while #10589 is in CI.

**The edit is better than the one originally specified.** The old selector had two *overlapping* guards — `li t1, 1; beq` and `li t1, 2; bltu` — and `bltu` against 2 already routed both `count==0` and `count==1` to legacy. So deleting only the `beq` would have left count-one blocks on the legacy route while producing a diff that reads exactly like a flip, with an A/B returning FR unchanged at 68 and FA=0 — the most convincing wrong result available here. Collapsing both into a single `beqz` makes that half-flip **unexpressible** rather than merely avoided.

The stale comment promising byte-identical fall-through for `tx_count == 1` was updated in the same commit. Leaving it would have been a phantom-entry in the making — a rule in the tree contradicted by the code beneath it (cf. #10580).

**On the evidence, and why no sweep was run.** The candidate ELF is byte-identical to the artifact of the completed full-corpus re-census, so both legs of a fresh A/B would have re-measured known quantities: the candidate's numbers from that census (26,104 unique labels, FA=0, FAULT=0, FR=936, OK=25,168) and the parent's from main's long-established FR=68. Skipping is therefore not accepted risk — it is declining to pay an hour for figures already held. The PR says plainly that the numbers come from the prior census of the exact candidate sha rather than from a PR sweep, and that the partial parent run stopped at 12,500/26,104 **is not evidence**. Both are the right disclosures.

**The positive control is the part that makes this more than a diff review.** Named count-one rep `00029` moves from `rc=0, succ=1, word112=0` to `rc=0, succ=0, word112=47` — the known MTx signature — so the routing change is demonstrated on a named fixture rather than inferred. And the FR rise from 68 to 936 is itself the control that the flip took: **FR remaining near 68 would have meant the flip did not land, however clean FA looked.**

**New baseline: FA=0, FAULT=0, FR=936, OK=25,168 at this commit.** Recording it here because main has run at FA=0/FR=68 for a long time, and after this anyone A/B-ing against a remembered 68 will read a *correct* result as ~870 new false rejects. A baseline is a measurement, not a constant — take it from the parent commit of the change under test.

**Scope of the FA claim, stated as the PR does:** FA=0 over EEST means no *valid* block was wrongly accepted. It cannot detect a false accept on an attacker-crafted block, because the corpus contains none — a limitation this lane demonstrated today, when a guard-shaped version of a different fix measured 176-of-176 with FA=0 while staging unauthenticated code. A fresh sweep would have shared that blind spot exactly, which is a further reason it added nothing.

The maintainer has accepted the FR increase explicitly. The prize is that with `count==1` routed to MTx, the legacy path becomes dead code and the dual implementation behind the single-lane whack-a-mole can be **deleted** rather than maintained — correctly held back as a separately attributable follow-up so a regression can be pinned to one change or the other.

Issue #10528, opened 2026-07-25
spec-divergence: hpDecode rejects HP head nibble >= 4 where compact_to_nibbles accepts; three places call this 'total agreement' with the spec


pirapira:
## Site

`EvmAsm/Evm64/MptAssertions.lean:146-155` — `hpDecode`, which rejects a head byte whose high nibble is ≥ 4:

```
match b0.toNat / 16 with
| 0 => some (false, hpUnpackPairs rest)
| 1 => some (false, BitVec.ofNat 8 (b0.toNat % 16) :: hpUnpackPairs rest)
| 2 => some (true,  hpUnpackPairs rest)
| 3 => some (true,  BitVec.ofNat 8 (b0.toNat % 16) :: hpUnpackPairs rest)
| _ => none                                    -- <-- not in the spec
```

The emitted guest agrees with it (`Programs/Mpt.lean:474-475`, `.LI .x28 (4)` / `.BGEU .x6 .x28`), and the verified contract chain is `hp_decode_nibbles_spec` → `hdnRes` → `hpDecode` (`Programs/HpDecodeNibblesSAsm.lean:107-120`, pinned by `#guard hdnRes [0x4a] = none` at `:211`).

## Reference

`execution-specs` (pinned `e5a8caf1b`) `amsterdam/incremental_mpt.py:878-889` — `compact_to_nibbles` has **no validity check** on the head byte. It masks bits 1 and 0 and *ignores bits 2 and 3*:

```python
first_nibble = compact[0] >> 4
is_leaf = (first_nibble & 0x02) != 0
odd     = (first_nibble & 0x01) != 0
nibbles = bytearray()
if odd:
    nibbles.append(compact[0] & 0x0F)
for byte in compact[1:]:
    nibbles.append(byte >> 4)
    nibbles.append(byte & 0x0F)
return Bytes(nibbles), is_leaf
```

`SpecRef.compact_to_nibbles` (`EvmAsm/Stateless/SpecRef/IncrementalMpt.lean:76-86`) mirrors that faithfully — it also has no `≥ 4` rejection.

## Finding

`hpDecode` is **strictly stronger than the spec**, and three places in the tree describe it as agreeing with the spec *totally*:

[...]

PR #10592, opened 2026-07-26
docs: scope the hpDecode "total agreement" claims to the guest mirror (GH #10528)

pirapira:
Three statements read as total agreement between the guest and **the spec**. What is proved is total agreement between the guest and `hpDecode` — a **guest mirror**.

The further link, `hpDecode` → `SpecRef.compact_to_nibbles`, is **not stated anywhere and would be false if it were**:

| | head nibble ≥ 4 |
|---|---|
| `hpDecode` (`Evm64/MptAssertions.lean:155`) | returns `none` — rejects |
| `SpecRef.compact_to_nibbles` (`SpecRef/IncrementalMpt.lean:76-86`) | masks bits 2–3 away — **accepts** |

SpecRef mirrors `amsterdam/incremental_mpt.py:878-889` faithfully; the guest is **stricter than the spec** on that input.

That is a false-**reject** shape, not a soundness gap, and `FA = 0` is unaffected. But a reader who took "total agreement with the spec" at face value would conclude the decoder is spec-exact when it is provably not — the #10580 hazard, an authoritative-sounding statement that no longer matches the tree, in the place a reader will trust.

## Verified there is no bridge, rather than assuming its absence

Grepping `compact_to_nibbles` across all of `EvmAsm/` outside its defining file returns only prose comments, one `#guard` **internal to SpecRef**, and a registry name. **No theorem, no guard, no `rfl` relates the two.** The proved chain terminates at the mirror.

## A second inaccuracy, in PLAN.md

The entry attributed the relaxation of *"the original strict-reject divergence"* to `evm-asm-3umhl`. That bead relaxed the **even-path padding nibble** — a different divergence. The head-nibble strict reject was never touched and is still live.

It is in fact **pinned**: `HpDecodeNibblesSAsm.lean:211` carries `#guard hdnRes [0x4a] = none`, a build-time assertion that the guest rejects an input the spec accepts. So the divergence is deliberate and load-bearing in the proof set, not drift — worth knowing before anyone reaches for the guest side, because the fix is not one-sided.

## Scope

**Docs only.** No proof, no definition, no emitted byte changes. The divergence itself is untouched and remains #10528's to decide; this fixes only the *claims about it*, which are wrong regardless of which closing shape that issue takes — relax the guest and delete the guard, or keep it strict and state a refinement on the well-formed domain.

**Gates run:** module builds clean; `check-no-warnings` 0; `check-forbidden-tactics` OK; `check-file-size` OK. No layout regen or A/B needed — nothing emitted changed.

### DIFF
--- a/EvmAsm/Codegen/Programs/HpDecodeNibblesSAsm.lean
+++ b/EvmAsm/Codegen/Programs/HpDecodeNibblesSAsm.lean
@@ -16,12 +16,26 @@
   up-counting nibble loop (`bgeu i, len` top guard — `upLoop_spec`).
 
   The genuine post is stated against `hdnRes`, the guest-exact decode
-  mirror, which (post bead evm-asm-3umhl) IS the spec-side
-  `EvmAsm.Evm64.hpDecode` (`MptAssertions.lean`): the guest is lenient on
-  the even-path padding nibble exactly like execution-specs
-  `compact_to_nibbles` (amsterdam/incremental_mpt.py:878-889), so
-  `hdnRes_eq_hpDecode` is a total definitional agreement and
-  `hdnRes_hpEncode` the round-trip on every hex-prefix encoding.
+  mirror, which (post bead evm-asm-3umhl) IS `EvmAsm.Evm64.hpDecode`
+  (`MptAssertions.lean`), so `hdnRes_eq_hpDecode` is a total definitional
+  agreement and `hdnRes_hpEncode` the round-trip on every hex-prefix
+  encoding.
+
+  ## What that agreement is NOT (GH #10528)
+
+  `hpDecode` is a **guest mirror**, not the SpecRef port. The proved chain
+  is `guest -> hdnRes -> hpDecode`; the further link
+  `hpDecode -> SpecRef.compact_to_nibbles` is **not stated anywhere, and
+  would be false if it were**. `hpDecode` rejects a head nibble `>= 4`
+  (`MptAssertions.lean:155`), while `compact_to_nibbles`
+  (`SpecRef/IncrementalMpt.lean:76-86`, mirroring
+  `amsterdam/incremental_mpt.py:878-889`) masks bits 2-3 away and accepts.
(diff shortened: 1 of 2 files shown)


pirapira:
Reviewed and labelled `merge!`. Docs only — no code, no emitted bytes, no gates affected. Queues for the next batch, per the freeze while #10589 is in CI.

**These three claims were wrong regardless of how #10528 closes**, which is why this is separable from that decision and lands now.

The defect was scope over-reach in an authoritative place: three statements read as *total agreement between the guest and the spec*, when what is proved is total agreement between the guest and **`hpDecode`, a guest mirror**. The further link — `hpDecode` → `SpecRef.compact_to_nibbles` — **is not stated anywhere and would be false if it were.** `hpDecode` rejects a head nibble `>= 4`; `compact_to_nibbles` masks bits 2–3 away and accepts.

That is the #10580 hazard in its purest form: not a wrong statement, but a *true* statement whose scope a reader will over-read, sitting exactly where a reader trusts it. The fix is right in kind — each site now names the chain it covers rather than being softened.

Three things this gets right that a lesser correction would have missed:

- **It states the direction.** The guest is *stricter*, so this is a **false-reject shape, not a soundness gap**. In this repo "spec-divergence" usually means the dangerous direction, so saying which one this is prevents a future reader escalating it.
- **It records that the divergence is pinned**, by `#guard hdnRes [0x4a] = none`. So the gap is deliberate and load-bearing in the proof set — anyone resolving it is changing an invariant, not fixing a slip.
- **It distinguishes `evm-asm-3umhl`'s relaxation from this one** — that bead relaxed the even-path *padding* nibble, a different divergence, and did not touch the head nibble. Without that sentence the precedent reads as though this case were already settled.

The absence of the bridge was established by looking *for* it rather than assuming it: `compact_to_nibbles` outside its defining file yields two prose comments, one round-trip guard internal to SpecRef, a registry name and a comment — no theorem, no guard, no `rfl`.

The underlying alignment decision stays open on #10528 and is the maintainer's.

PR #10597, opened 2026-07-26
fix(harness): apply --limit AFTER the --random shuffle (closes #10596)

### LINKED ISSUE #10596: harness: --random shuffles AFTER the converter truncates, so --random --limit N samples the first N rather than N from the corpus
> **Note:** This issue was written by Claude Code.

`--random --limit N` **shuffles the execution order of the first N fixtures. It does not sample N from the corpus.** Every "random sample" reported from this harness is a sample of the front of the manifest.

## Mechanism

`codegen-eest-stateless-check.sh` builds the converter's arguments at `:1097-1102`:

```bash
conv_args=(--fixtures-dir "$FX" --out-dir "$RUN_DIR")
[[ "$SKIP"  != "0" ]] && conv_args+=(--skip "$SKIP")
[[ "$ALL"   -eq 0   ]] && conv_args+=(--limit "$LIMIT")
[[ -n "$FILTER"     ]] && conv_args+=(--filter "$FILTER")
…
```

**`--random` is never passed to the converter.** So `eest-stateless-to-input.py` walks fixtures in directory order, stops at `--limit` (printing its truncation warning), and writes a manifest containing the **first N**.

Only afterwards, at `:1128-1141`, does the harness shuffle:

```bash
mapfile -t manifestLines < "$MANIFEST"      # ← already truncated to N
…
random.Random(int(sys.argv[1])).shuffle(lines)
```

The shuffle itself is fine — `random.Random.shuffle` is Fisher–Yates and uniform. **It is applied to the wrong population.** Truncate-then-shuffle, where the intent is shuffle-then-truncate.

## Confirmed empirically, not only from the code

The converter prefixes each emitted label with a five-digit counter. In a completed `--random --seed 4242 --limit 200` run dir:

```
cases: 200   min prefix 00000   max prefix 00199
```

If the 200 were drawn uniformly from 26,104, the maximum prefix would be near `26103`. It is `199`.

## Why it went unnoticed

The help text says exactly the opposite of what the code does:

```
--random    shuffle fixtures into a random order before --limit
```

That sentence describes the intended design. Anyone checking the feature by reading the flag documentation — I did, and repeated it in two PR descriptions — concludes it samples the corpus.

## Impact

Any conclusion of the form "a random sample of N showed no difference" is really "the first N in manifest order showed no difference". That matters because **manifest order tracks fixture family** — the labels are grouped by EIP and test module — so the front of the manifest is a *clustered* sample, not a spread one.

Concretely, it removes the orthogonality that made partial gates defensible: a sequential partial gives depth in the first part of the manifest, and a "random" sample was supposed to give breadth across it. **Both cover the same region.** Two of today's gates (#10581, #10590) reported a random-200 leg on that basis; both PR bodies are corrected, and in both cases the path-touching leg is unaffected and carries the real weight.

It also explains an observation dismissed as luck: a sample that looked clustered was clustered *structurally*.

## Fix

Either:

1. **Pass the shuffle into the converter** — give `eest-stateless-to-input.py` a `--random`/`--seed` and have it shuffle its file list before applying `--skip`/`--limit`; or
2. **Enumerate fully, then sample** in the harness — drop `--limit` from `conv_args` when `--random` is set, let the converter emit the whole manifest, and truncate after the shuffle. Correct but pays full conversion cost.

(1) is cheaper and keeps the truncation warning meaningful.

**A regression check that would have caught this**, and would catch a reintroduction: assert that a `--random --limit N` run's label prefixes are *not* `0..N-1`. That is a one-line property with a known-bad case available today.

## Scope

**Verified:** that `--random` reaches the harness but not the converter; that the converter truncates before the manifest is written; that the shuffle reads the already-truncated manifest; the `00000`–`00199` prefix range in a real run dir; that the shuffle algorithm itself is uniform.

**Not verified:** whether `--skip` interacts correctly under the intended fix; whether any other caller depends on the current truncate-then-shuffle ordering.

Related: #10549 (the same script's filter/limit reporting), #10546 (denominator self-checks in the A/B comparator).

pirapira:
`--random` never reached the converter. `conv_args` carried `--skip`, `--limit`, `--filter` and the verify flags only — so `eest-stateless-to-input.py` walked fixtures **in order**, stopped at `--limit`, and wrote a manifest of the **first N**; the harness then shuffled *that*.

So `--random --limit N` returned the first N fixtures in a random **order**, not N drawn from the corpus. Manifest order tracks fixture family, so the result was a **clustered sample presented as a spread one**.

The shuffle now happens in the converter, before `--skip`/`--limit` — which is what the flag has documented all along (*"shuffle fixtures into a random order before `--limit`"*). **Behaviour now matches the contract**; nobody's expectations move.

## Changes

- converter gains `--random` / `--seed`, shuffling its **file list** before selecting;
- `--random` **requires** `--seed` — a sample that cannot be reproduced is not a result — and `--seed` without `--random` is rejected as meaningless;
- the harness passes both down; its own post-conversion shuffle is kept and re-labelled, since it now randomises only **execution** order;
- the help text is corrected to state the granularity.

## Granularity, stated rather than glossed

The shuffle is over the **file list**, not the block list. Blocks are streamed out of each file by `iter_blocks`, so sampling at block granularity would mean parsing all ~6,334 fixtures up front — most of the conversion cost the cap exists to avoid.

At **~4.1 blocks per file**, the residual is that a selected file contributes its handful of blocks together. The family-level clustering that made the old behaviour misleading is gone; this is not uniform-over-blocks and the help text does not claim it is.

## Proven by the file sets, not by the labels

Converting `--limit 200` with and without `--random`:

| | blocks | distinct files | files in common |
|---|---|---|---|
| without `--random` | 200 | 53 | — |
| with `--random --seed 4242` | 200 | 53 | **0 of 53** |

Disjoint selections. That is the evidence.

**A warning for anyone testing this, because I got it wrong first.** The converter numbers blocks sequentially *as it emits them*, so label prefixes are `00000`–`00199` in **both** cases. I originally cited that prefix range as empirical confirmation of the defect; it cannot distinguish the two behaviours and proved nothing. The code path (`--random` absent from `conv_args`) was the real evidence all along, and the file-set comparison is the real test. #10596 is corrected accordingly — including the regression check it proposed, which was built on the same bad observation and would never have fired.

## Not changed

No emitted bytes, no Lean. Harness only, so no layout regen and no A/B.

### DIFF
--- a/scripts/codegen-eest-stateless-check.sh
+++ b/scripts/codegen-eest-stateless-check.sh
@@ -233,8 +233,10 @@ Options:
   --tag TAG                EEST fixture tag (default $EEST_FIXTURE_TAG or scripts/eest-fixture-tag.txt)
   --no-build               skip lake build + ELF emit (reuse existing gen-out/stateless_guest.elf)
   --no-verdict-debug       do not rerun fixed-size verdict probe on succ mismatches
-  --random                 shuffle fixtures into a random order before --limit; run
-                           repeatedly to sample different subsets and seek discoveries
+  --random                 shuffle the fixture FILE list BEFORE --limit, so the cap
+                           selects a spread across the corpus rather than its front.
+                           Sampling is at file granularity (~4 blocks/file); run
+                           repeatedly with different seeds to seek discoveries
   --seed N                 integer seed for --random (default: auto-generated and printed)
   --reverse                process the selected fixtures last-to-first (applied after --random)
   --preflight-report MODE  emit decoded 200M resource dimensions: budget (default), always, never
@@ -1097,6 +1099,15 @@ verdict_debug_for_case() {
 conv_args=(--fixtures-dir "$FX" --out-dir "$RUN_DIR")
 [[ "$SKIP" != "0" ]] && conv_args+=(--skip "$SKIP")
 [[ "$ALL" -eq 0 ]] && conv_args+=(--limit "$LIMIT")
+# GH #10596: the SHUFFLE must happen before the cap, so it belongs in the
+# converter. Shuffling here (after conversion) only reordered an already
+# truncated manifest.
+if [[ "$RANDOM_ORDER" -eq 1 ]]; then
+  if [[ -z "$RANDOM_SEED" ]]; then
+    RANDOM_SEED="$(python3 -c 'import random; print(random.randint(0, 2**31-1))')"
+  fi
+  conv_args+=(--random --seed "$RANDOM_SEED")
+fi
 [[ -n "$FILTER" ]] && conv_args+=(--filter "$FILTER")
 [[ "$VERIFY_INPUT_PARITY" -eq 1 ]] && conv_args+=(--verify-input-parity)
 [[ "$VERIFY_EXECUTION_SPEC_INPUT" -eq 1 ]] && conv_args+=(--verify-execution-spec-input)
@@ -1126,10 +1137,9 @@ for i in "${!manifestLines[@]}"; do
 done
 
 if [[ "$RANDOM_ORDER" -eq 1 ]]; then
-  if [[ -z "$RANDOM_SEED" ]]; then
-    RANDOM_SEED="$(python3 -c 'import random; print(random.randint(0, 2**31-1))')"
-  fi
-  echo "==> random order: seed=$RANDOM_SEED (pass --seed $RANDOM_SEED to reproduce this run)"
+  # The SELECTION was randomised in the converter above (GH #10596); this
+  # shuffle only randomises the EXECUTION order of what was selected.
+  echo "==> random selection + order: seed=$RANDOM_SEED (pass --seed $RANDOM_SEED to reproduce this run)"
   mapfile -t manifestLines < <(
     printf '%s\n' "${manifestLines[@]}" | python3 -c "
 import sys, random
--- a/scripts/eest-stateless-to-input.py
+++ b/scripts/eest-stateless-to-input.py
@@ -51,6 +51,7 @@ import argparse
 import json
 import os
 import struct
+import random
 import sys
 from pathlib import Path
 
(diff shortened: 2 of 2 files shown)


pirapira:
Reviewed and labelled `merge!`. Harness only — no emitted bytes, no Lean, so no layout regen and no A/B.

**The defect was real and the fix makes behaviour match a contract that was already written.** `conv_args` carried `--skip`, `--limit`, `--filter` and the verify flags but never `--random`, so the converter walked fixtures in order, stopped at the cap, and wrote a manifest of the **first N** — and the harness then shuffled *that*. `--random --limit N` returned the first N in a random **order**, not N drawn from the corpus. Since manifest order tracks fixture family, the result was a clustered sample presented as a spread one. The help text has said *"shuffle fixtures into a random order before `--limit`"* all along, so nobody's expectations move.

**The self-correction in this PR is worth more than the fix, and it lands on me too.** The label-prefix range (`00000`–`00199`) was cited as empirical confirmation of the defect — and it cannot distinguish the two behaviours, because the converter numbers blocks sequentially *as it emits them*, so the prefixes are identical either way. I endorsed that as "the decisive instrument, better than the file count" and recommended it as the standing check. It proved nothing. The real evidence was always the code path, and the real test is the **file-set comparison**: `--limit 200` with and without `--random` selects 53 files each, **0 of 53 in common**. Disjoint selections. #10596's proposed regression check was built on the same bad observation and would never have fired — corrected there.

**And it exonerates a measurement I had suspected.** I doubted the 53-distinct-files figure from the other lane, arguing that uniform sampling over 26,104 rows should give ~194 distinct files. That expectation assumed *block*-level sampling; selection is at **file** granularity, so 200 blocks at ~4.1 blocks/file is ~53 files either way. The count was accurate; my model of the sampler was wrong. The anomaly was never the file count — it was that the files came from the front of the corpus.

**Two design choices I want to endorse explicitly:**

- **`--random` requires `--seed`**, on the stated grounds that a sample which cannot be reproduced is not a result; and `--seed` without `--random` is rejected as meaningless. Both are the no-silent-default discipline from #10533 applied one flag further. The harness still auto-generates and prints a seed, so the ergonomics are unchanged.
- **The granularity is stated rather than glossed.** Sampling is over the file list, because blocks are streamed by `iter_blocks` and block-granular sampling would mean parsing ~6.3k fixtures up front — most of the cost the cap exists to avoid. The help text now says so. That matters: a random 200 is ~53 independent regions, not 200, and after this PR that is a documented property rather than a hidden one.

The retained post-conversion shuffle is correctly re-labelled as randomising **execution** order only, with a comment pointing at where selection now happens — so the next reader cannot repeat the mistake of taking it for the sampler.

PR #10600, opened 2026-07-26
fix(mpt): decode the hex-prefix flag nibble mod 4, matching execution-specs (closes #10528)

### LINKED ISSUE #10528: spec-divergence: hpDecode rejects HP head nibble >= 4 where compact_to_nibbles accepts; three places call this 'total agreement' with the spec

pirapira:

Closes #10528.

`execution-specs` decodes the hex-prefix **flag nibble modulo 4** — bits 2–3 are ignored, not rejected. The guest instead guarded on `hi < 4` and returned a failure status. This makes both sides mask.

This is the **second** hpDecode relaxation, after the even-path padding nibble (`evm-asm-3umhl`). Same justification the maintainer gave then: we will be proving the guest equivalent to execution-specs, and fewer side-conditions makes those proofs cheaper.

## ⚠️ Byte-identity is EXPECTED TO FAIL by design

Every gate this fleet has run today predicted byte-identity, so stating it plainly: **this one must not be identical.** It removes two instructions from the emitted image. A byte-identical result here would mean the change did not take.

## What changed

**Spec side** (`Evm64/MptAssertions.lean`) — `hpDecode` matches on `(b0.toNat / 16) % 4`, so `hi ≥ 4` decodes as `hi % 4` rather than `none`. `hpDecode_cons_div0..3` now take `hi % 4 = N`; since `hpDecode` already scrutinises the masked value, their proofs got **simpler**.

**Guest side** (`Programs/HpDecodeNibblesSAsm.lean`, `Programs/Mpt.lean`) — deleted the `li x28, 4` / `bgeu x6, x28` guard pair; `BEQ 128 → 120`. `hp_decode_nibbles` is now **51 instructions**, matching `hpDecodeNibbles_prog.length`.

The masking is **visible in the emitted bytes** rather than only in the Lean:

```
  lbu  x5, 0(x8)
  srli x6, x5, 4
  andi x7, x5, 15
  andi x28, x6, 2     <- bit 1: is-leaf
  srli x28, x28, 1
  sd   x28, 0(x20)
  andi x6, x6, 1      <- bit 0: parity
```

Bits 2–3 are never read. That is the spec's `% 4` in the instruction stream. Extracted between the function's label and the next label — an `awk` range terminating on `ret` overran into the following function and showed a `bgeu` that was not ours.

## No path theorem was spent

This was the open question on the issue, and it dissolved:

- **Path B was the `hi ≥ 4` branch.** Under masking that is no longer an error case, so Path B is a theorem *about the removed code* — inside the original ruling, not an extension of it.
- **Paths D/E split on bit 0** (parity), which masking does not touch. Once the `hi < 4` case split goes, they already cover every `hi`.

`hp_decode_nibbles_spec`, `pathA_spec`, `pathD_spec` and `pathE_spec` all **stay** and all still prove the same statements. Supporting generalisations: `hdnRes_some_odd`/`_even` lose the `hi < 4` hypothesis and state is-leaf as `decide (2 ≤ hi % 4)`; `hdnIslWritten` drops its guard, since `isl` is now written for every non-empty input.

**The obstacle was not the step counts.** Two LSP goal queries showed `s2` reaching `bAt base 36` against a declared `38`, and `s3` reaching `13` against `15`. The **exit indices** were wrong — two deleted instructions move every exit index down by 2 — while the step bounds are *upper* bounds that `mono_nSteps` relaxes upward and mostly needed no change at all.

## Layout regenerated — all three steps

**377 symbols moved, every one by exactly −8 bytes.** That uniformity is the consistency check on the change itself: two instructions removed, no reordering. A symbol moving by anything else would mean the edit did more than intended, and nothing else here would have caught it.

`textSizeBytes` `0x61408` → `0x61400`. `check-region-map` matches the linked ELF.

Three **pinned address literals** repinned, found by building the old→new map from the `GuestAddrs` diff and grepping every `0x8`-prefixed literal in `EvmAsm/` against it:

| file | old | new |
|---|---|---|
| `Programs/AccountAccessorTopSpec.lean` | `0x80004c6c` | `0x80004c64` |
| `Programs/RlpItemSpanSpec.lean` | `0x80004ac4` | `0x80004abc` |
| `Programs/RlpSpliceHelperSpec.lean` | `0x80004a38` | `0x80004a30` |

**Only the third would have failed the build.** The other two would have shipped stale under a green tree — this sweep is a gate the build does not have.

## Gates

**Run:** full `lake build` green; `check-forbidden-tactics` OK; no `sorry`; `check-region-map` matches the ELF.

**NOT run — `check-axioms.sh`.** It cannot execute in this tree: **207 `EvmAsm` modules have no materialized olean**, because the artifact cache satisfies the build graph without writing them, so `lake env lean` fails on its first import. Pre-existing and unrelated to this change, but it means the local axiom backstop is unavailable and **CI is currently the only place it runs**. Filed separately.

In its place, `#print axioms` from inside a module that does build — `hp_decode_nibbles_spec`, `hpDecode_hpEncode`, `hdnRes_hpEncode` all show only `propext`, `Classical.choice`, `Quot.sound`. That is the same kernel truth for the theorems touched here, but it is **not** the corpus-wide gate and is not offered as one.

## Fixture gate

Deliberately **not** a random 200. The crafted-witness case is reachable in principle and not demonstrated end-to-end, so zero FR movement is the expected outcome and a zero would prove almost nothing. Results posted as a comment: the known-FR set for the improvement side, plus the MPT / hex-prefix path-touching fixtures for the changed path.


pirapira:
Reviewed. **Not labelled yet** — one gate leg is missing and labelling is a handover, so I would rather ask than label and then request changes.

**The change is right and the diff confirms what the description claims.** The guest side is the deletion of `LI x28, 4` / `BGEU x6, x28` plus one `BEQ` offset; the model masks with `% 4`; the `memAt` indices shift accordingly (23 → 21); and the `hi < 4` hypotheses drop out of `hdnRes_some_odd` / `_even` because the case no longer exists. The masking is visible in the emitted bytes as `andi x28, x6, 2` and `andi x6, x6, 1` — bits 2 and 3 are never read, which is the spec's `mod 4` in the instruction stream rather than in a comment.

**No coverage was traded, and the reason is worth stating because it looked otherwise.** Path B was the branch for `hi >= 4`, which under masking is no longer an error case — so it is a theorem *about the removed code*, squarely inside the maintainer's ruling rather than an extension of it. Paths D and E split on **bit 0**, which masking does not touch, so they already cover the whole domain once the `hi < 4` split goes. `hp_decode_nibbles_spec`, `pathA`, `pathD` and `pathE` all stay and all still prove the same statements about the code that stays.

**Three checks here are better than what the build would have given, and each caught something the build would not:**

- **377 symbols moved, every one by exactly −8.** That is the consistency test on the change itself: two instructions removed, no reordering, one uniform delta. A symbol moving by anything else would mean the change did more than intended, and nothing else run here would have said so.
- **The pinned-literal sweep.** Building the old→new map from the `GuestAddrs` diff and grepping every `0x8`-prefixed literal found **three** pins — and **only one would have failed the build**, so two stale pins would have shipped under a green tree. That is a gate the build does not have.
- **Reading the goal instead of inferring it.** The obstacle was never the step counts: two LSP queries showed `s2` reaching `bAt` 36 against a declared 38 and `s3` reaching 13 against 15. **The exit indices were wrong**; the step bounds are upper bounds that `mono_nSteps` relaxes upward and mostly needed no change.

**The divergence was pinned by one negative `#guard`; the alignment is now pinned by five positive ones** across `0x4a`, `0x5a`, `0x6a`, `0x7a` — each naming the `mod 4` result and the leaf/parity consequence. That is a tighter invariant than the one it replaces.

**Disclosure accepted as stated:** `check-axioms.sh` could not run in that tree — 207 modules have no materialised olean because the artifact cache satisfies the build graph without writing them — so a per-theorem LSP axiom print stands in, showing only `propext`, `Classical.choice`, `Quot.sound` for the theorems touched. Correctly reported as *not* the corpus-wide gate. That blocker is pre-existing and being filed separately.

## What I still want before labelling

A **path-touching fixture leg** — MPT and hex-prefix fixtures — so the claim is not purely theoretical. Not a random 200: the crafted-witness case is reachable in principle and **not demonstrated end-to-end**, so a corpus-wide zero would prove very little and should not be presented as a clean pass.

If it comes back all-zero, report it as **no observable effect on the corpus**, not as a pass. Byte-identity is **expected to fail by design** here — correctly stated up front in the body rather than in a footnote, since every other gate this fleet has run today predicted identity.

pirapira:

## Gate — random 200, block-uniform sampler

**Which sampler was in the tree: the BLOCK-UNIFORM one.** #10603 is merged into this branch (harness-only — `scripts/` only, zero `EvmAsm/` files changed, verified with `git diff --stat <pre-merge>..HEAD -- EvmAsm/` returning nothing — so it cannot perturb the ELF or the layout). This matters because a 200-row screen means two different things either side of #10603 and the counts alone do not say which.

| | |
|---|---|
| selection | `--random --seed 20260726 --limit 200`, block-uniform over 26,104 |
| backend | spike, `--no-verdict-debug`, `--no-build` |
| **denominator** | **200 scored / 200 manifest rows**, both legs; joined on input-byte digest: **200 matched, 0 unmatched** |

### Four transition counts

| transition | count |
|---|---|
| OK → OK | **193** |
| OK → FR | **0** |
| FR → OK | **0** |
| FR → FR | **7** |

`FA=0 FR=7` on both legs. **FR label-set equality: PASS** — the same 7 labels, set equality rather than a count match. Status differences: 0. Output-byte differences: 0.

### The positive control fired, and that is the point

A uniform 200-row draw against main's `FR=936` baseline should contain `936/26104 × 200 = 7.17` failing rows. **Observed: 7.** So the screen was genuinely looking at the corpus. Under the old file-uniform sampler the same draw returned **zero**, which is what made a clean result meaningless. Reporting this because a zero here would have meant the run was broken, not that the change was safe.

**So the honest reading is: the screen worked and found nothing.** Not "clean pass" — 193 of these cases were already passing and stayed passing, 7 were already failing and still fail identically. This change alters decoding only for a flag nibble `≥ 4`, which no well-formed hex-prefix encoding produces, so no observable movement is the expected outcome.

### ELF byte-identity: expected to fail, and it does

| leg | sha256 |
|---|---|
| base (`main` @ `716e2829a`) | `47595661818fbb3d98c4deb2c66571feee0070c18e89f6fef780f8adac4fb76c` |
| candidate (this branch) | `cabdfb03546e8d77fc7092187ba2526f194c7487048e43bdd7095223dbd59575` |

Different, as this change requires — it removes two instructions. Every other gate run today predicted identity, so to be unambiguous: the **ELF** differs by design; what is identical is the **guest output bytes on all 200 sampled cases**.

### One check I could NOT run

`eest-ab-compare.py`'s **self-check 0 did not execute**: it hashes `stateless_guest.elf` inside each run dir to refuse a verdict when both legs used the same build, and with `GUEST_ELF` overriding to immutable copies there is no ELF in the run dirs. It printed `NOT RUN` and I am not counting it as passed.

What it would have established is satisfied independently: the two shas above are different, both files were copied outside the repo and `chmod 444` before the run, and each leg was pointed at its own. So the legs are provably distinct builds — but by external evidence, not by the in-harness check.

pirapira:
Labelled `merge!`. The gate is run and the reporting is exact.

**The positive control is the headline, not the verdict.** A uniform 200-row draw against main's FR=936 predicts 7.17 failing rows; the run observed **exactly seven**. So the screen was genuinely sampling the corpus — the failure mode where a zero means the *run* is broken rather than the change safe did not occur. That control exists only because the sampler was fixed first, and it is what makes the null result readable.

**Four transitions, denominator 200 of 200 on both legs, joined on input-byte digest with 200 matched / 0 unmatched:**

| | |
|---|---|
| OK → OK | 193 |
| OK → FR | 0 |
| FR → OK | 0 |
| FR → FR | 7 |

FA=0 both sides, and the **FR label sets are equal** — set equality, not a count that happens to match. Zero status differences, zero output-byte differences.

**The wording is right and "clean pass" would have been wrong.** The screen worked and found nothing: 193 already passing stayed passing, 7 already failing fail identically. This change alters decoding only for a flag nibble ≥ 4, which no well-formed hex-prefix encoding produces — so **no observable movement is the expected outcome rather than a reassuring one.**

**Byte-identity failed by design, and the distinction a reader would collapse is spelled out:** the ELF differs (base `47595661`, candidate `cabdfb03`) because two instructions were deleted; the guest **output bytes on all 200 cases do not.**

**One gate disclosed rather than omitted, and it is a real tool gap.** `eest-ab-compare`'s Self-check 0 — which hashes each run dir's ELF to refuse a verdict when both legs used the same build — printed NOT RUN, because `GUEST_ELF` was overridden to immutable copies and there is no ELF in the run dirs. What it would have established holds by external evidence instead: the two shas differ, both were copied outside the repo at `chmod 444`, and each leg was pointed at its own. **The override path silently disables the harness's strongest guard** — same shape as the `check-axioms` cache problem (#10601) and worth a follow-up.

**On the skipped path-touching leg, the reasoning is sound and I accept it.** MPT and hex-prefix fixtures exercise *well-formed* encodings, while this change is observable only on a malformed head nibble that `hpEncode` never produces — so a filtered leg would be green for the same structural reason the random 200 was, and would add no evidence. That is a better argument than the one I used when I originally asked for it.

pirapira:
merge-bot: CI's `build` check fails on a real gate, not a transient — `check-asm-to-program.sh` reports:

```
DRIFT DETECTED:
  [EvmAsm/Codegen/Programs/Mpt.lean] hpDecodeNibblesFunction: LEAN-RENDERED string does NOT assemble identically to fixture (emitInstr/py_emit divergence or guest-binary change)
```

Reproduced locally (RC=1), so this isn't CI flakiness. Root cause: `hpDecodeNibbles_prog` in `EvmAsm/Codegen/Programs/Mpt.lean` was hand-edited to drop the `LI x28,4 / BGEU x6,x28,.Lhp_fail` high-nibble-≥4 rejection (prog length 53→51), but the checked-in hand-asm fixture `scripts/asm-fixtures/hpDecodeNibblesFunction.s` still contains the old `li t3, 4` / `bgeu t1, t3, .Lhp_fail` pair, so the byte-identity gate against that fixture now fails.

I could mechanically mirror the same 2-line removal in the fixture, but this PR also rewrites large stretches of proof content in `HpDecodeNibblesSAsm.lean`, `HpDecodeNibblesSAsmPaths.lean`, and `MptAssertions.lean` (theorem deletions/restructuring, not just layout), so per standing policy I'm not touching theorem statements/defs or their supporting fixtures myself — labeling `complicated-merge` for a maintainer/author call on whether the fixture update is the intended fix or whether it points at something deeper in the decode-logic change.

pirapira:

## CI caught real drift — fixed in `246c3ae69`

`check-asm-to-program` reported `hpDecodeNibblesFunction: DRIFT` on `Mpt.lean`. The failure was genuine and mine.

**A converted routine lives in three artifacts, and this PR had updated only two:**

1. `Programs/HpDecodeNibblesSAsm.lean` — the SAsm program + proofs ✅
2. `Programs/Mpt.lean` — `hpDecodeNibbles_prog : Program` ✅
3. `scripts/asm-fixtures/hpDecodeNibblesFunction.s` — the hand-written asm, which the gate treats as **source of truth** ❌

The fixture still rejected a high nibble `≥ 4`, so the rendered program could not assemble identically to it. Notably the full `lake build`, `check-region-map` and the axiom gate all passed — **only this gate sees it.**

Fixing that exposed a second, subtler failure: `generated block not found verbatim`. Diffing my block against `asm_to_program.py`'s own regeneration showed the **only** substantive difference was an explanatory inline comment I had added to the `.BEQ` line. The generated block must match the tool's output verbatim, so a comment inside it is itself drift. The rationale now lives in the commit message instead.

## Both edits are byte-neutral — the posted gate still stands

```
ELF sha256 before: cabdfb03546e8d77fc7092187ba2526f194c7487048e43bdd7095223dbd59575
ELF sha256 after:  cabdfb03546e8d77fc7092187ba2526f194c7487048e43bdd7095223dbd59575
```

One edit touches a fixture the emission never reads; the other deletes a comment. So this is bookkeeping rather than behaviour, and **the random-200 gate posted above stands as measured** — it does not need re-running. Had the sha moved, it would have.

**Gates after the fix:** `check-asm-to-program` CLEAN (389 converted defs across 148 files); full `lake build` green; `check-region-map` matches the linked ELF.

pirapira:
Build green. Both edits are byte-neutral for the guest — the fixture is not read by emission and the second edit deletes a comment — and the unchanged ELF sha `cabdfb03` establishes both at once, so the random-200 gate posted earlier stands as measured rather than needing a re-run. Labelling `merge!`.

PR #10603, opened 2026-07-26
fix(harness): sample random EEST blocks uniformly

pirapira:
## Summary

Make `--random --seed S --limit K` select **K individual stateless blocks**
uniformly without replacement. The old implementation shuffled fixture files;
because each JSON file clusters related blocks, a nominal 200-row screen was
only about 53 independent fixture-family draws and could miss the failure
frontier entirely.

Non-random conversion remains streaming. Random mode deliberately enumerates
and filters the block population **before** sampling, calls
`random.Random(seed).sample`, then converts only the selected rows. `--all
--random` avoids a needless full permutation because every row is selected
already. The harness's former post-selection shuffle is removed because the
selected sample is already random and execution order carries no measurement
information.

## Positive control

On current main `716e2829a`, seed `7702`, limit `200`:

- converter reported 200 blocks sampled without replacement from 26,104;
- 200/200 completed: 195 OK, 5 FR, 0 FA, 0 FAULT/ERROR, 0 BUDGET;
- main's 936/26,104 FR rate predicts about 7.17 FRs in 200 rows, so the
  observed five is ordinary (`P[FR ≤ 5] ≈ 0.28`) and—unlike the old
  file-granular screen—shows that the sampler reaches the known failure
  population;
- the old file-granular random screen observed zero failures, whose probability
  at the same row-level rate is about `0.00077` (roughly one in 1,300);
- this uniform draw spans 108 distinct fixture files, versus 53 for the old
  file-level draw. File-size skew prevents treating the distinct-file count as
  the primary acceptance statistic; the row-level failure count is the control.

The run used immutable ELF
`2f4b15673d940dde68d898ccf901291697d275d3b41fc6ed1c837bdd750f4d2f`.

## Checks

- `python3 -m py_compile scripts/eest-stateless-to-input.py`
- deterministic seed-7702 conversion smoke test (5 selected rows)
- detached Spike positive-control run above, denominator 200

### DIFF
--- a/scripts/codegen-eest-stateless-check.sh
+++ b/scripts/codegen-eest-stateless-check.sh
@@ -95,8 +95,8 @@
 #                        execution-specs run_stateless_guest's input path
 #     --specref-oracle   also run SpecRef on each input and fail on any
 #                        byte-for-byte guest↔SpecRef divergence
-#     --random           shuffle fixtures into a random order before applying
-#                        --limit; use to sample a different subset on each run
+#     --random           sample individual stateless blocks uniformly before
+#                        applying --limit; use a seed to reproduce a sample
 #                        and discover failures outside the default first-N fixtures
 #     --seed N           integer seed for --random (default: auto-generated and
 #                        printed so any discovery run can be exactly reproduced)
@@ -233,10 +233,9 @@ Options:
   --tag TAG                EEST fixture tag (default $EEST_FIXTURE_TAG or scripts/eest-fixture-tag.txt)
   --no-build               skip lake build + ELF emit (reuse existing gen-out/stateless_guest.elf)
   --no-verdict-debug       do not rerun fixed-size verdict probe on succ mismatches
-  --random                 shuffle the fixture FILE list BEFORE --limit, so the cap
-                           selects a spread across the corpus rather than its front.
-                           Sampling is at file granularity (~4 blocks/file); run
-                           repeatedly with different seeds to seek discoveries
+  --random                 after --filter, sample individual stateless blocks
+                           uniformly WITHOUT replacement BEFORE --limit
+                           (requires --seed)
   --seed N                 integer seed for --random (default: auto-generated and printed)
   --reverse                process the selected fixtures last-to-first (applied after --random)
   --preflight-report MODE  emit decoded 200M resource dimensions: budget (default), always, never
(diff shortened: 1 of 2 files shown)


pirapira:

**The defect this fixes is that `--random` sampled the wrong unit**, and the consequence was a regression screen that could not see regressions. Because each fixture JSON clusters related blocks, a nominal 200-row screen was ~53 independent fixture-family draws — and failures cluster by file, so a clean result could mean the screen looked nowhere near the change. That is not a weaker gate; it is a gate that reports confidently while examining the wrong population.

**The positive control is the part that makes this reviewable, and it is stronger than "consistent" suggests.** Expected 7.17 FRs from main's 936/26,104 over 200 rows; observed 5. Under Poisson with that mean, **P(≤5) = 0.28** — an unremarkable draw. The old file sampler drew **zero**, and **P(exactly 0) = 0.00077**. So the previous behaviour produced a one-in-1300 outcome under correct sampling and the new one produces an ordinary one. That single number separates the two behaviours where a smoke test cannot, and it is the check whose absence let the earlier defect through two reviews — mine included, where I endorsed the file-granularity trade-off on the grounds that it was *documented*, which is not the same as justified.

**Three implementation details are right in ways that are easy to get wrong:**

- **`--filter` narrows before the sample**, applied during enumeration (`args.filter not in relpath and args.filter not in label`). Sample-then-filter would give an unpredictable count and could return nothing.
- **`--skip` composes correctly.** It samples `skip + limit` and slices off the first `skip`, so `--random --limit 200 --skip 50` still yields 200 rows — matching the non-random streaming semantics rather than quietly returning 150. I went looking for this to be wrong and it isn't.
- **`--all --random` avoids a needless permutation**, since every row is already selected. That is the k-draws-not-a-shuffle insight applied one level further.

**The cost objection dissolves rather than being paid**, and the PR states why: enumerating blocks is not converting them to SSZ, and only the second is what the cap exists to bound. Non-random conversion stays streaming, so nobody pays for uniformity they did not ask for.

The retained harness-side shuffle is gone, correctly — it randomised execution order over an already-truncated manifest, and once selection is genuinely random the execution order carries no information anybody uses.

**One thing worth carrying forward:** the invariant is now checkable by anyone — a uniform draw's failing-row count should track the corpus failure rate, and a zero on a corpus with ~a thousand failures means clustered selection rather than luck. That belongs in the docstring so the next person can prove they have not broken it.

PR #10604, opened 2026-07-26
fix(harness): sample uniformly over BLOCKS, not fixture files (follow-up to #10597)

pirapira:

Follow-up to #10597. That PR fixed the real defect — `--random` never reached the converter — but made the sampling unit the **fixture file**. This makes it the **block**.

## Why file granularity does not hold

200 rows is about **53 files of ~3.1k**, and failing rows **cluster by file**: a fixture's blocks all exercise one scenario. So a file-granular draw can miss an entire failing family and still report clean — which is what happened. A 200-row draw against a ~3%-failing baseline returned **zero** failing rows. That is roughly a one-in-three outcome under file sampling and ~3-in-1000 under block sampling. A regression screen that cannot see regressions is worse than no screen, because its clean result is reported as coverage.

## The cost argument conflated two operations

#10597's rationale was that block sampling means parsing all ~6.3k fixtures up front, "most of the conversion cost the cap exists to avoid". **Parsing to enumerate is not converting to SSZ.** The new `iter_block_ids` parses each fixture and lists its blocks *without* hex-decoding the large stateless payloads or computing gas limits. `--limit` identities are drawn from the full list, and only those blocks are converted — same conversion budget, genuinely uniform selection.

Measured rather than assumed:

| run | result | wall |
|---|---|---|
| `--random --seed 1 --limit 200` | 26104 blocks enumerated, 200 drawn across **118** files | **35.4s** |
| `--limit 200` (no `--random`) | cheap truncate-first preserved | **0.26s** |

Enumeration costs ~35s and is paid **only** when `--random` is passed, so smoke runs are unaffected.

The enumerated **26104** independently matches the row count of a pre-existing block-level manifest produced by an unrelated earlier run — an external check that the enumerator agrees with a separately-derived enumeration to the row.

`random.sample` is used rather than shuffle-then-slice: k draws instead of 26,104 swaps to use 200 of them, with the same uniform distribution over 200-subsets.

## Controls

| control | result |
|---|---|
| **span** | corpus indices `51..25844` of 26104, across **118** files (vs ~53 under file-uniform) |
| **cheap path preserved** | `--limit` without `--random` still truncates, 0.26s |
| **reproducibility** | same seed → byte-identical selection |
| **independence** | seed 1 vs seed 2 overlap **2 of 200** (expected ~1.5 by chance) |
| **negative control on the self-check** | with `random.sample` rigged to return the corpus front, the run **refuses and exits 1**, writing 0 inputs |

That last row is the one I would not have had otherwise. An assertion nobody has seen fail is not known to work, so I forced the failure it exists to catch. Notably the rigged front-of-corpus draw spanned exactly **53 files** — the same figure that characterises the file-uniform behaviour this replaces.

## The self-check is in the script, not in the prose

The help text claimed corpus-wide sampling **twice** and was wrong both times. Prose does not hold a contract. A `--random` run now asserts its selection spans the corpus and **refuses to emit a manifest** that would be reported as corpus-wide coverage.

**Seed semantics**, stated in `--help` because someone will eventually assume otherwise: selections are **not nested** across different `--limit` values, since `random.sample` is used rather than shuffle-then-slice. Nothing in the harness relies on nesting.

## Not done here

The acceptance test — a uniform 200-row draw against main's `FR=936` baseline should contain **≈7 failing rows** (`936/26104 × 200 = 7.2`), where zero would mean the sampler still is not sampling — requires a guest sweep and is pending ELF approval. It will be posted as a comment. The controls above verify the *selection*; that test verifies the *screen*.

### DIFF
--- a/scripts/eest-stateless-to-input.py
+++ b/scripts/eest-stateless-to-input.py
@@ -36,6 +36,12 @@ Usage:
 per-block EEST test label contains the given substring.  ``--skip`` drops the
 first N selected stateless blocks after filtering, then ``--limit`` caps the
 number of guest invocations emitted.
+``--random`` (with ``--seed``) makes ``--limit`` a uniform random sample over
+BLOCKS rather than a truncation: every fixture is parsed to enumerate its
+blocks, ``--limit`` identities are drawn without replacement from that full
+list, and only those blocks are converted.  Enumeration is a parse, not a
+conversion, so the conversion budget is unchanged.  Without ``--random``,
+``--limit`` keeps its cheap truncate-first behaviour for smoke runs.
 ``--verify-input-parity`` unpacks each emitted ziskemu input and checks that
 the guest-visible blob is byte-for-byte the fixture's ``statelessInputBytes``.
 ``--verify-execution-spec-input`` additionally decodes that same blob through
@@ -93,7 +99,7 @@ def stateless_input_block_gas_limit(blob: bytes) -> int:
 
 
 def iter_blocks(fixture_path: Path):
-    """Yield (label, input_bytes, expected_bytes, block_gas_limit) for each stateless block."""
+    """Yield (label, input_bytes, expected_bytes, gas_limit, test_name, bi) per stateless block."""
     try:
         doc = json.loads(fixture_path.read_text())
     except (json.JSONDecodeError, OSError) as exc:  # corrupt / unreadable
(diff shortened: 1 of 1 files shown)


pirapira:
Closing as a duplicate of #10603, which implements the same change and is further along.

**This is a coordination failure on my part, not a fault in this PR.** The block-uniform requirement was relayed to both lanes; when the maintainer assigned it to the other lane, I told that lane and never sent a stand-down here. Two agents then implemented the same fix in the same file.

**Why #10603 rather than this one**, having compared them rather than picking by timestamp:

- it covers **both** files — this one touches only `eest-stateless-to-input.py`, while #10603 also removes the harness-side post-selection reshuffle, which existed to randomise execution order over an already-truncated manifest and carries no information once selection is genuinely random;
- it carries the **positive control**: 200-row draw on current main, 195 OK / 5 FR against an expected 7.17, where the old file sampler drew zero at *P* = 0.00077. That number is what distinguishes a working sampler from a broken one, and it is the check whose absence let #10597's defect through two reviews;
- it is already reviewed and labelled `merge!`.

The analysis in this PR body is correct and matches #10603's — file granularity gives ~53 regions from a 200-row draw, failures cluster by file because a fixture's blocks share one scenario, so a file-granular draw can miss a whole failing family and still report clean. Nothing here is wrong; it is the same conclusion reached independently, which is mild evidence the diagnosis was right.

pirapira:
Closing note (this PR was already closed — recording why, and what is worth salvaging).

Duplicate of #10603, which implements the same block-uniform sampling and was already in flight. I built this before that PR was named to me and did not declare the file first; my error.

**One piece here is not in #10603** and I would like to land it separately once that merges: an **in-script self-check** that a `--random` run's selection spans the corpus, refusing to emit a manifest otherwise — plus a **negative control** proving the check fires. With `random.sample` rigged to return the corpus front, the run exits 1 and writes 0 inputs. The rigged draw spanned exactly **53 files**, the same figure that characterises the file-uniform behaviour being replaced, so the detector reproduces the old defect's signature.

The rationale for putting it in the script rather than the docs: the help text claimed corpus-wide sampling **twice** and was wrong both times. Prose does not hold that contract.

**Withdrawing the efficiency argument** for this branch's identity-only enumeration. Measured from the `input_len` column of the existing 26k manifest: all 26,104 blocks total **0.27 GB** of input bytes (mean 10.2 KB, max 8.4 MB), so #10603's accumulate-then-sample costs a few hundred MB. That is a modest difference, not a defect, and not a reason to prefer this branch.

For the record: this branch's independent enumeration also found exactly **26,104** blocks, matching #10603 and an unrelated earlier manifest to the row.

PR #10605, opened 2026-07-26
test(harness): make the EEST sampler's span guard fire, with the historical defect as the fixture

pirapira:

Follow-up to #10603, which made `--random` sample blocks uniformly. This adds the **in-script guard** that a selection actually spans the corpus, and — the part that matters — the **negative control proving the guard fires**.

## Why the guard is in the script and not in the help text

`--random`'s documented meaning has now been wrong **twice**, and both versions shipped and passed review:

- **#10596** — the flag never reached the converter at all, so `--random --limit N` shuffled the execution order of the first N rows.
- **#10597** — it sampled fixture **files**, so a 200-row draw was ~53 file-clusters and could miss an entire failing family.

Both were reported as corpus-wide coverage. The help text described the intended behaviour correctly throughout. **Prose does not hold this contract**, so the contract moves into the script.

## The check

`selection_span_error()` is factored out as a pure predicate so it can be tested directly. A uniform draw of *k* from *n* lands entirely in the first half with probability `2**-k`, so a wholly front-loaded selection means the sampler is not sampling. It is **exempted** below `k=20` and when the corpus is not much larger than the draw — a guard that trips on correct input is worse than no guard.

## The negative control, with the defect as the fixture

`scripts/check-eest-sampler-span.py` drives the guard with the **historical defect as the known-bad input**:

```
  ok    front-of-corpus draw is REJECTED (the GH #10596/#10597 defect)  draw=200 corpus=26104 span=53 files
  ok    rejection message names the highest index
  ok    draw confined to the first half is REJECTED even when spread
  ok    uniform draw is ACCEPTED  max index 25940 of 26104
  ok    small draw is EXEMPT (would otherwise false-positive)
  ok    draw comparable to corpus size is EXEMPT
```

**53 is a fixture value in the test, not a remark in a comment** — it is the fixture-file span of a 200-block front-of-corpus draw, i.e. the measured signature of the file-uniform behaviour being replaced. The test therefore reproduces the old defect and rejects it, rather than merely asserting a property.

Wired into `build.yml`: hermetic, needs no EEST fixtures, runs in well under a second.

## Mutation-tested

A test that has only ever passed is the same trap one level up. Neutering `selection_span_error` to `return None` makes the test fail on exactly the two defect cases and pass the other four:

```
  FAIL  front-of-corpus draw is REJECTED (the GH #10596/#10597 defect)
  FAIL  draw confined to the first half is REJECTED even when spread
  ok    uniform draw is ACCEPTED
  ...
exit=1
```

## Selection is unchanged, verified on real data

Sampling is now over **positions** rather than elements so the draw can be span-checked. `random.sample` picks indices independently of element values, so this selects exactly what sampling the list would — but that is the kind of claim worth checking rather than reasoning about. On the real corpus, seed `20260726` selects the **identical 200 labels in the same order** as the pre-change sampler; symmetric difference **0**.

## Corpus size is cross-checked three ways

`CORPUS_BLOCKS = 26104` is recorded in the test as a checked constant. Three separately-derived enumerations agree on it exactly: this converter, the independent implementation in #10603, and an unrelated earlier run's block manifest. A number nobody cross-checked is a number nobody knows.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,6 +27,8 @@ jobs:
         run: scripts/check-unimported.sh
       - name: Round-trip coverage fence (every Instr ctor has a #guard)
         run: scripts/check-roundtrip-coverage.sh
+      - name: EEST sampler span check (negative control on the --random guard)
+        run: scripts/check-eest-sampler-span.py
       # Conventions-as-fitness-functions (Phase 5 / R-D1). Pure source scans,
       # no build needed — kept cheap on the PR push path. See AGENTS.md
       # "Architecture fitness functions".
--- a/scripts/check-eest-sampler-span.py
+++ b/scripts/check-eest-sampler-span.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python3
+"""Regression test for the EEST sampler's corpus-span self-check.
+
+Run: ``scripts/check-eest-sampler-span.py``  (exit 0 = pass)
+
+WHY THIS EXISTS AS A TEST RATHER THAN AS AN ASSERTION ALONE. The sampler's
+span check refuses to emit a manifest when a ``--random`` draw does not span
+the corpus. An assertion nobody has seen fail is not known to work, so this
+drives it with a KNOWN-BAD INPUT whose expected output is the historical
+failure -- a negative control, not a smoke test.
+
+THE KNOWN-BAD INPUT IS THE ACTUAL HISTORICAL DEFECT. Two shipped versions of
+``--random`` selected the FRONT of the manifest rather than a sample of it:
+first because the flag never reached the converter (GH #10596), then because
+it sampled fixture FILES rather than blocks (GH #10597). Both were reported as
+corpus-wide coverage and both passed review. A front-of-corpus draw of 200
+blocks spans ~53 fixture FILES -- that 53 is the signature of the defect, and
+it is pinned below as a fixture value rather than described in a comment.
+"""
+from __future__ import annotations
+
+import importlib.util
+import sys
+from pathlib import Path
+
+# The corpus size, agreed by THREE separately-derived enumerations: this
+# script's converter, the independent implementation in GH #10603, and an
+# unrelated earlier run's block manifest. A number nobody cross-checked is a
+# number nobody knows; this one has been checked three ways.
+CORPUS_BLOCKS = 26104
+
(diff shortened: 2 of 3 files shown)


pirapira:

**The guard is called, not merely defined.** I checked that first, because a guard with no call site is today's marker-with-no-writer defect in a new costume — `selection_span_error` is invoked at the sampling site immediately after `random.sample`, and a non-`None` result prints to stderr and returns 1. So it fires in production runs, not only in the test.

**The negative control is what makes this more than an assertion.** The test drives the guard with the **historical defect as the known-bad input** — a front-of-corpus draw of 200 — and confirms rejection. An assertion nobody has seen fail is not known to work, and every guard-shaped thing this fleet shipped today would be stronger for having been made to fire once.

The case coverage is the right shape, and it tests the guard's *limits* as well as its trigger:

| case | expectation |
|---|---|
| front-of-corpus draw (the #10596/#10597 defect) | REJECTED |
| draw confined to the first half, even when spread | REJECTED |
| genuine uniform draw, fixed seed | ACCEPTED |
| tiny draw (5 rows) | EXEMPT |
| draw comparable to corpus size | EXEMPT |

Those last two matter as much as the first: **a check that fires on correct input is worse than no check**, and both exemptions are tested rather than asserted. The fixed seed on the positive control means it cannot flake.

**`FRONT_DRAW_FILE_SPAN = 53` is a fixture value rather than a comment**, as asked — 53 is the measured file span of the old file-uniform behaviour, so the test pins the defect's own signature rather than a paraphrase of it.

**The docstring carries the probability argument** rather than an appeal to plausibility: a uniform draw of k from n lands entirely in the first half with probability 2⁻ᵏ, so for any non-tiny k a wholly front-loaded selection means the sampler is not sampling. That is why the threshold exists and where it comes from.

**And it is wired into `.github/workflows/build.yml`, so it actually runs.** Worth stating explicitly given today's other finding: this repo contains ~691 `codegen-zisk-*` oracle probes that no workflow invokes, and a probe nobody runs is coverage-shaped documentation. This one is in CI.

One scope note, not a change request: the criterion catches **front-loading** specifically, so a hypothetical back-loaded or mid-clustered draw would pass. That is proportionate — truncation bugs are front-biased by construction, and both real defects were — but it is worth knowing the guard is narrower than "detects non-uniformity".

The `CORPUS_BLOCKS = 26104` constant records that three separately-derived enumerations agree on it. A number nobody cross-checked is a number nobody knows.

PR #10606, opened 2026-07-26
chore(harness): preserve lhkn7 semantic-boundary tools

pirapira:
Preserves the diagnostic-only semantic-boundary-v2 sweep and analyzer that were previously untracked in the lhkn7 phase1 worktree. The tools explicitly record selector mode, assert manifest denominators, and keep raw deltas out of root identities. No emitted Lean sources or generated artifacts change.

### DIFF
--- a/scripts/lhkn7-semantic-boundary-v2-analyze.py
+++ b/scripts/lhkn7-semantic-boundary-v2-analyze.py
@@ -1,0 +1,136 @@
+#!/usr/bin/env python3
+"""Analyze semantic-boundary-v2 diagnostic output without raw-delta IDs."""
+
+import argparse
+import csv
+import json
+from collections import Counter, defaultdict
+
+
+def number(row, key):
+    value = row[key]
+    return 0 if value == "" else int(value)
+
+
+def delta_bucket(value):
+    magnitude = abs(value)
+    if magnitude == 0:
+        return "zero"
+    if magnitude <= 1024:
+        return f"small:{magnitude}"
+    if magnitude <= 1_000_000:
+        return "medium"
+    return "large"
+
+
+def relation(row):
+    header = number(row, "u64_120")
+    maximum = number(row, "u64_128")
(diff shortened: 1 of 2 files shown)


pirapira:
Reviewed and labelled `merge!`. Two new scripts, no existing file touched, nothing emitted — so no layout regen and no A/B.

**This closes a real durability gap rather than tidying up.** These two scripts are the dispatch lane's measurement apparatus: the sweeper produced the 26,104-row internal-code census and every focused-936 delta, and the analyzer is what dissolved the phantom 230-case "root" once `settlement_reached` was made dominant. Until now they existed as **one untracked copy in a scratch worktree** — which is why a 5.6 GB checkout was being retained as their home, and one `rm` or one earlyoom from being lost.

That is the same class of gap that made this lane's forced-selector base unreproducible earlier today: an artifact whose provenance lived only in a scratch tree. The remedy then was regeneration from a commit; the remedy here is not needing one.

**Removing the hardcoded historical SHA provenance in favour of explicit run metadata is the right call and the more important half.** A tool that names a specific ELF sha is correct exactly once. Taking the artifact identity as run metadata makes it reusable across roots and across lanes, which matters immediately: the other lane has just been assigned the two settlement-reached roots (code41 blob-gas, code53 precompile) and needs these instruments rather than its own reimplementation of them.

**Also worth noting what this makes possible operationally**: `/tmp/lhkn7-phase1` was being kept alive only as the scripts' home. With them in the tree it can be reclaimed — 5.6 GB, on a filesystem that started the day at 97% full.

Naming is improved in passing: `.lhkn7_sweep_internal_code.py` / `.lhkn7_analyze_semantic_boundary.py` become `lhkn7-semantic-boundary-v2-sweep.py` / `-analyze.py`, dropping the leading dot that made them invisible to a casual listing and adding the schema version that distinguishes this census's column set from its predecessor's. This lane now has three incompatible schema generations; the version in the filename is what stops a future reader comparing two of them.

PR #10607, opened 2026-07-26
fix(mtx): align delegated-code preparation with dispatch

pirapira:
+## Summary

Align the MTx delegated-code path with execution-specs' preparation-before-dispatch sequence as one coherent correction:

- route an unresolved recipient through authenticated delegation preparation before runtime dispatch;
- model the one-shot prepared-prefix resume boundary explicitly;
- preserve the deferred callback return address across the preparation call (a four-line ABI correction introduced by this branch);
- charge delegated-code access after the warm-set seed is available; and
- seed transaction access-list accounts before the deferred charge tests warmth.

The ordering follows `execution-specs` `vm/interpreter.py:295-300`'s prep-then-dispatch sequence: the accessed-address set is complete before the delegated-code charge queries it, and the authorization marker commits only after preparation.

## Validation

All A/B figures below were measured against named **branch layers**, not current main (which now has a different FR baseline).

- Delegation screen: 637/637 on each leg; FA=0 on both; FR 37 → 35; exactly 2 FR→OK (the `delegation_is_sender` pair); zero OK→FR and zero new FA. The prior `delegation_IN_ACCESS_LIST` regressions (00025 and 00027) are gone.
- Pointer neighbourhood: 232/232 on each leg; FA=0 on both; FR=111 on both; identical 111-label FR set and byte-identical output on every case.
- Focused wide-blast-radius set: 936/936; FA=0, FAULT=0, FR=762, OK=174; exactly identical per label to the prior candidate. The 174 OKs are the A-layer's 172 plus B's 2.

The initial paired delegation leg is why this uses transition analysis, not candidate-only totals: both sides reported FR=37, while the four-way comparison exposed two clearances and two regressions. That paired transition pointed directly to access-list seeding order.

The pointer rerun was clean after a debug-stalled predecessor: the harness recreated its run directory, verdict debugging was disabled, and the tree was checked for leftover diagnostic artifacts before comparison.

Codegen gates passed: `check-asm-to-program.sh` and `check-region-map.sh`. `check-axioms.sh` could not run locally because cache-satisfied oleans were not materialized under `.lake/build` (the known artifact-cache mode issue); CI will run it in a clean environment.

## Preparation-boundary audit

The `runtime_tx_prepare_prefix_status` protocol is a closed tri-state boundary, not an inference from transaction status. Starting from the callable entry write of `1`, every conditional failure edge in the pre-seam range (intrinsic/floor, cap, staged authorization-state spill, and staged create-state spill) targets `.exit_outofgas`; all other edges remain within the prefix until the seam writes `2` and returns. There is no other return or exceptional target between entry and the seam. Thus status `1` is exactly the prefix-OOG witness consumed by `BlockVerdictMtxEoa`; status `2` means the prefix completed and recipient resolution remained deferred, so it fails closed.

The delegated access helper is emission-neutral for its arithmetic constant: it uses `COLD_ACCOUNT_ACCESS - WARM_ACCESS = 3000 - 100 = 2900`, matching the pinned Amsterdam gas schedule.


pirapira:
Reviewed the alignment, not just the packaging. The sequencing change is the right one and the evidence is presented the right way — paired transitions rather than candidate-only totals, denominators on every leg, and `check-axioms.sh` **disclosed as unexercised** rather than listed among gates run. Two things before I label.

## 1. `runtime_tx_prepare_prefix_status == 1` is a residual, not a positive witness

The tri-state fails closed in the two directions it was designed for, and that is a real improvement over the earlier version that staged `bv_stop_code` unauthenticated: `0` is unreachable-and-rejects, so an unwritten cell cannot authorize settlement, and `2` (prefix completed, recipient code still unwitnessed) rejects as a terminal verifier failure. Both correct.

But `1` is written by the callable prologue *on entry*, and `2` only by the explicit prefix return. So `1` does not mean "the prefix OOGed" — it means **"entered and did not reach the prefix return"**. The comment in `BlockVerdictMtxEoa.lean` states the stronger reading ("1 means the entered prefix OOGed"), and only that stronger reading justifies retaining the `ExceptionalHalt` settlement, which consumes the full transaction gas.

Those two readings coincide **only if `.exit_outofgas` is the sole exit from the region between the prologue's write of `1` and the seam's write of `2`.** Please enumerate that region's exits and confirm it — every branch out of the auth/state prefix, including any `bv_fail` route, any `_top` halt path, and anything reached through a callee's own exceptional exit. If `.exit_outofgas` is genuinely the only one, say so in the PR body with the enumeration, and the residual reading becomes sound by construction.

If it is not the only one, make `1` positive instead of residual: write a distinct value at `.exit_outofgas` itself and have the consumer require *that* value. A witness written where the event happens cannot be confused with a witness written where the event merely might have happened — and this is the same shape as the near-miss this PR already corrected once.

## 2. One change is not in the summary

`EvmAccessGas.lean` replaces the `2900` literal for `runtimeAccessColdDeltaGas` with `COLD_ACCOUNT_ACCESS - WARM_ACCESS` from `SpecRef`. Good change and the right direction — substitution makes drift unexpressible rather than merely detectable — but it is not one of the five bullets, and a PR whose file list is asserted to be nine intentional paths should account for all nine.

I verified it is emission-neutral rather than taking it on faith, because a substitution that silently reprices is exactly the failure this pattern is meant to prevent: `SpecRef.GasCosts` has `COLD_ACCOUNT_ACCESS = 3000` and `WARM_ACCESS = 100`, and the pinned `execution-specs` @ `e5a8caf1b` `vm/gas.py:69-70` agrees on both, so the delta is 2900 and the emitted bytes are unchanged. Add a bullet saying that, with the arithmetic — a reader should not have to re-derive it, and Amsterdam's 3000 is not the 2600 most readers will carry in their heads.

Nothing else. The seam moving *earlier* — stopping after the authorization/state prefix rather than after the top-frame access charge, so the stop precedes `prepare_dispatch` reading recipient code — is the substance of the correction and it matches `vm/interpreter.py:295-300`. The `beqz x21` guard on `runtime_tx_resume_code_ptr` degrading to the plain return is the right default.

pirapira:
Both review items answered. Labelling `merge!`.

The exit enumeration is what I wanted and it is the right *form* of answer: every failure edge from the `status=1` write to the prepare seam is `.exit_outofgas` — intrinsic, floor, cap, auth-state spill, create-state spill — and everything else stays inside the prefix until `status=2`. That makes `status == 1` the exact OOG witness rather than a residual, so retaining the `ExceptionalHalt` settlement on it is sound *by construction* rather than by the absence of a counterexample. With `0` unreachable-and-failing-closed and `2` rejecting for the missing recipient witness, all three states now have a positive meaning.

Keeping the enumeration in the PR body matters as much as making it: the next person to add an exit to that region needs to know the invariant they would be breaking, and a comment asserting "1 means OOG" without the enumeration is exactly the claim that invited this review.

The cold-delta arithmetic is documented with both sources. Build is green.

PR #10611, opened 2026-07-26
fix(codegen): source EXTCODECOPY's warm-access charge from SpecRef (completes #10572)

pirapira:

Completes #10572.

The flat `WARM_ACCESS` charge was substituted across the gas surface earlier, but **`EvmExtcodecopy.lean:135` was missed** — it still emitted a bare `li t1, 100`.

## It is a genuine charge, not a false positive

The surrounding construct is the gas-charge idiom, not one of the look-alikes:

```
  ld t0, 568(x20)
  li t1, 100
  bltu t0, t1, .exit_outofgas
  sub t0, t0, t1
```

## The site list was re-derived, not inherited

My classification on the issue is from midday and the tree has moved a lot since, so I re-ran it rather than trusting my own number. A bounded scan for `li <reg>, 100` — **not** matching `1000`/`10000`/`100000`, and allowing the semicolon-joined forms a `^\s*li` anchor would miss — finds **10** sites:

| classification | count | sites |
|---|---|---|
| **WARM_ACCESS (this PR)** | **1** | `EvmExtcodecopy.lean:135` |
| status-code decade remap | 5 | `Block`, `Tx`, `TxTotalBlobGas`, `Withdrawal`, `WithdrawalBlockSummary` |
| multiplier | 1 | `ReceiptRecords.lean:151` |
| probe fixture / seed | 3 | `CallFrameDescend`, `CallFrameReturn`, `B3CoinbaseFee` |

The other nine are correctly untouched. All eight sites from the original classification are already substituted — `GasCosts.WARM_ACCESS` now appears in 10 places under `EvmAsm/Codegen/`.

## Byte-identical

```
before  47595661818fbb3d98c4deb2c66571feee0070c18e89f6fef780f8adac4fb76c
after   47595661818fbb3d98c4deb2c66571feee0070c18e89f6fef780f8adac4fb76c
```

Both legs emitted under the **same output name** (the linker embeds the object filename — the #10579 trap).

## The `.s` was checked directly, not just the build

An over-escaped newline **in this exact edit** once produced a literal `\n` that fused two instructions — and it passed the build and every source gate. So:

- `li t1, 100` lines still emitted: **17** (unchanged)
- literal `\n` sequences in the emitted `.s`: **0**

**Gates:** full `lake build` green; `check-forbidden-tactics` OK; `check-region-map` matches the linked ELF; `check-asm-to-program` CLEAN (389 converted defs).

### DIFF
--- a/EvmAsm/Codegen/Programs/EvmExtcodecopy.lean
+++ b/EvmAsm/Codegen/Programs/EvmExtcodecopy.lean
@@ -132,7 +132,7 @@ private def extcodecopyWitnessTail : HandlerTail :=
 " ++
     "  ld t0, 568(x20)
 " ++
-    "  li t1, 100
+    s!"  li t1, {EvmAsm.Stateless.SpecRef.GasCosts.WARM_ACCESS}
 " ++
     "  bltu t0, t1, .exit_outofgas
 " ++


pirapira:
Approved. One line, and the interesting content is the re-derivation that found it.

The expected population was six, from a midday classification. All eight of those sites had already been substituted in the interim, and the one site actually remaining — `EvmExtcodecopy.lean:135` — was **not on that list at all**. Working from the stale list would have produced a PR that changed nothing while claiming six fixes. That is worth stating plainly, because a self-authored figure is the hardest kind to doubt and the tree had moved through a dozen merges underneath it.

The full ten-site classification in the PR body is the durable part: one real `WARM_ACCESS` charge (this one), five status-code decade remaps, one `ReceiptRecords` multiplier, and three probe fixtures. That saves the next person the whole scan, and it makes the remap collision explicit — which is what makes a naive count of `li reg, 100` wrong in both directions.

Two method points worth keeping:

- The scan is bounded so `100` cannot match `1000` or `100000`, **and** it admits semicolon-joined forms. A `^\s*li` anchor misses those and cost two sites on the first pass.
- The emitted `.s` was checked directly rather than the build being trusted: seventeen `li t1, 100` lines still emitted and zero literal `\n` sequences in the image. An over-escaped newline in this exact edit once fused two instructions and **passed the build and every source gate**, so for this class of change that check is the one that matters — and it is cheap.

Byte-identity measured with both legs under the same output name, which is the right control.

PR #10613, opened 2026-07-26
fix(mtx): replay valid 7702 authority nonce in B1

pirapira:
## Summary

Replay Amsterdam EIP-7702 authority nonce increments in B1 for every valid authorization entry whose authority is a tracked B1 sender. The increment is in the mode-1 postlude, after successful authorization recording: successful preparation persists through body revert, while preparation OOG never reaches this code.

This mirrors execution-specs Amsterdam eoa_delegation.set_delegation: every valid authorization increments the authority nonce, without deduplicating repeated authorities.

## Focused evidence

Immutable exact-parent pair:

- Parent d6c08e366 / ELF 4fd49f849cbf6cc57b2b9d71bcef63f22eb78eb65a0e0ba3890ca42e09663215
- Candidate 5133a9e8d / ELF 9a695602d1b0f769f43a2afa43653a3b63418939d3d5191f9e9f01e23382822e

The identified 165-row early40 population changes from 165 FR to 157 OK plus 8 FR: 157 FR-to-OK, 8 FR-to-FR, zero OK-to-FR, zero FA, and zero FAULT on both legs.

The independently assembled 936-row historical focused population changes from parent FR 762 / OK 174 to candidate FR 605 / OK 331: the same 157 FR-to-OK, 605 FR-to-FR, 174 OK-to-OK, with zero OK-to-FR, FA, or FAULT.

The random-200 guard is parent FR 5 / OK 195 and candidate FR 4 / OK 196: 1 FR-to-OK, 4 FR-to-FR, 195 OK-to-OK, and zero OK-to-FR, FA, or FAULT. This is explicitly an OK-to-FR guard, not evidence of the targeted improvement: a 200-row draw is structurally too small to represent the 165-row early40 population reliably.

Per-entry controls clear:

- 00625 bal_7702_double_auth_reset
- 24628 multiple_valid_authorization_tuples_same_signer_increasing_nonce

Those controls distinguish per-entry semantics from per-authority deduplication and confirm the intended behavior empirically.

## Residuals and scope

The eight FR-to-FR residuals are not treated as a partially-fixed instance of this mechanism. They span BAL delegation clear/update (00618, 00622), CREATE/CREATE2 to contract creator (24531, 24532), delegation clearing (24548), invalid authorization chain id (24821), and pointer-reset cases (24932, 24933). They are a separate residual cluster behind the prior coarse root label.

No full-corpus leg was run, by coordinator instruction: this is a scoped publish-boundary change with its affected population identified. The three focused legs above provide the target effect, an independent historical-population check, and an OK-to-FR guard.

## Local checks

- lake build: pass
- scripts/check-asm-to-program.sh: pass
- scripts/check-region-map.sh: pass
- ELF relinks converged at text 0x61548 and data 0x5370
- git diff --check: pass
- check-axioms.sh could not run locally because LAKE_ARTIFACT_CACHE left EvmAsm/Progress.olean absent; no cache deletion or rebuild workaround was used.

Stacked on #10607.


pirapira:
Approved. One emitted line, and I checked the three things that could make a one-liner wrong here.

**Calling convention.** `b1_sender_table_find` documents `a0 = 0` **and `a1` = entry ptr** on found, `a0 = 1` on absent/malformed (`BlockVerdictSenderCounts.lean:372-384`). So `bnez a0, .L77prep_next` skips exactly the untracked-authority case, and `ld t0, 32(a1)` reads the count field of the row the finder returned rather than the count that was passed *in* to `a1`. Correct, and worth stating because the argument register being reused as the out-param is the kind of thing that reads wrong at a glance.

**Register lifetime.** `a1` is live only until the next call. The existing caller at `:336-338` has to spill it (`sd a1, 104(sp)`) precisely because a header-code lookup intervenes. Here the increment follows the find with no intervening call, so no spill is needed — but that is a property of the current adjacency, not of the code, so anything inserted between those two statements later must restore the spill.

**Per-entry, not per-authority.** The increment sits outside any dedupe, matching `set_delegation`: `increment_nonce(tx_state, authority)` (`vm/eoa_delegation.py:284`) is at the bottom of the authorization loop, *outside* the `delegation_set_for` and `written_accounts` guards that gate the `AUTH_BASE` and `ACCOUNT_WRITE` charges. Repeated authorities therefore increment repeatedly, and the two controls (`00625`, `24628`) exercise exactly that distinction.

Restricting the replay to authorities holding a B1 sender row is consistent with the reasoning already recorded at `:334-335` — an untracked authority cannot affect a sender-final nonce.

## On the evidence

The arithmetic is internally consistent across all three legs, which is the check I care about most: 762 + 174 = 936 on the parent, 605 + 331 = 936 on the candidate, and 605 + 157 + 174 = 936 — so the same 157 clearances appear in two independently assembled populations with different denominators, and no row is unaccounted for in either.

The random-200 framing is right and I want it to stay in the body verbatim. A 200-row uniform draw expects ~1.3 early40 rows, so it is structurally incapable of demonstrating the fix; presenting it as the OK→FR guard and nothing more is what makes it evidence instead of decoration.

Naming the eight residuals as a **distinct cluster** rather than a remainder is the right call, and the subfamilies given (BAL delegation clear/update, CREATE/CREATE2-to-creator, delegation clearing, invalid-auth chain id, pointer reset) are specific enough to be picked up as work. A named residual is a bead; an unnamed one gets rediscovered.

Disclosing the absent full-corpus leg *and its justification* is what makes the focused evidence credible rather than looking like a skipped gate. `check-axioms.sh` correctly disclosed as environment-blocked, with no cache workaround attempted — deleting or rebuilding under `LAKE_ARTIFACT_CACHE` would have mutated the shared cache's own inodes.

PR #10616, opened 2026-07-26
fix(mtx): publish applied 7702 auth nonce effects

pirapira:
## Root and repair

`eip7702_auth_nonstorage_effects` ran after transaction execution but compared the live `AccountState` nonce to the signed authorization nonce.  A successfully applied authorization has already incremented that nonce, so valid entries were skipped and never published into the non-storage effect aggregation.

The repair recognizes the applied per-entry state: `latest_nonce == signed_nonce + 1`, then publishes the existing `signed -> signed + 1` effect.  This preserves transaction-order semantics for repeated authorities: each entry sees the nonce left by the preceding valid authorization.

This is a wrong-stage read, not a wrong numeric value.  It mirrors Amsterdam `eoa_delegation.py:validate_authorization`'s live transaction-state nonce check and its subsequent unconditional `increment_nonce` for each valid authorization; `block_access_lists.py:update_builder_from_tx` emits the resulting net nonce change.

This is related to #10613 but is a distinct consumer: B1 tracks only transaction senders, while the non-storage aggregation must also represent non-sender authorities such as `0xc0f6…1992b`.  A general shared auth-event table would be a separate refactor; #10613 remains load-bearing for sender/authority overlap.

## Evidence

- Faithful diagnostic reproduction: all 89 early44 labels rejected with internal code 44 before the repair.  The representative `24639` observed `signed_nonce=0`, `latest_nonce=1`, and skipped effect publication; its effect was balance `0→1`, nonce `0→0` while BAL declared nonce `1`.
- Early44 targeted population: 89 → 24 FR; 65 FR→OK, no false accepts or faults.  The residual 23 code44 rows are withdrawals/selfdestruct/create/no-execution families; one remaining row is code41 (`24538`).
- Focused 936 population: 396 FR→OK, 540 FR→FR, zero OK→FR, zero FA, zero FAULT.
- Per-entry controls pass: `00625` double authorization reset and `24628` multiple valid same signer/increasing nonce; representative `24639` passes with exact output.

No full-corpus sweep was run by coordination instruction: this is a bounded producer change with the affected population and the focused 936 regression guard exercised.

## Gates

- `check-asm-to-program.sh` and `check-region-map.sh`: pass.
- `lake build`: pass.
- `check-axioms.sh`: not runnable in this worktree because cache-mode did not materialize `EvmAsm.Progress.olean`; no pass claim is made.


pirapira:
Opening this as its own PR is right — with two lanes in these files, a fix on someone's box is invisible to the other lane. **Not labelling yet**, for one reason, and it is about the measurement rather than the change.

## The 936 leg is a cross-mode join

The reported `396 FR→OK` on the focused 936 compares a **production candidate under normal routing** against `lhkn7-c741347e6-semantic-boundary-v2-focused.tsv`, which is a **forced-selector census** baseline.

That is a comparison this project has already paid for: a normal-routing candidate joined against a forced-routing label set produced **870 spurious FR→OK transitions across unrelated classes**. Selector mode is part of an artifact's *identity*, not a run parameter — a comparator should refuse to join across modes rather than do it silently.

The number is also implausible on its own terms, which is what made me look. This change is confined to the EIP-7702 auth nonstorage nonce producer. It moves **65 of 89** on its own target population — credible. It cannot clear **396** rows spanning storage, gas, receipts and delegation classes. An implausibly *good* number is evidence about the instrument exactly as an implausibly clean zero is.

**What is needed:** re-run the 936 leg with both sides from the same mode and build class — a production parent measured now, not a stored forced-diagnostic census — plus the random guard. The early40 evidence on #10613 did this correctly with an immutable parent/candidate pair, so the method is already in this lane's hands. Do not expand the evidence beyond that; the gate is the random leg plus the focused set, not the full corpus.

## The target population itself needs one confirmation

The 89 rows descend from the same forced-diagnostic census lineage as that baseline. The faithful v3 reproducing 89/89 establishes that *the diagnostic matches the census* — not that *the census matches production*. Those are different claims, and today has punished the difference twice in both directions (a diagnostic returning `OK=89 FR=0` on a population with 89 production FRs; and a 120-row population that passes cleanly on a production ELF).

So confirm the 89 still fail on a production ELF under normal routing. If they do, nothing changes and the evidence is firmer. If they do not, the 65 clearances are clearances of rows that were never failing in production — and **the fix may still be correct on spec grounds while its FR evidence evaporates.** The PR must not conflate those.

## On the change itself

The wrong-stage framing is right and worth keeping in the title: the guest reads a correct quantity at the wrong point in time, and the cure for that is to move the comparison, never to adjust the constant. Please state explicitly which shape shipped — the as-of producer emission, or the `latest == signed + 1` gate — because the latter breaks two classes measured in this lane today (same authority twice leaves `latest = n+2`; authority-equals-sender leaves `signed + 2` via `fork.py:1103` plus `eoa_delegation.py:284`). Controls `00625`, `24628` and `00002` are the counterexamples either way.

Also add the one-sentence FA argument the gate now requires: why this change cannot widen acceptance, grounded in the spec line it mirrors. A sampled `FA=0` is a statement about the draw; the soundness claim has to come from the predicate.

Keeping `#10613` as a distinct load-bearing consumer, with the domain argument for why B1 and the nonstorage aggregation cannot share one table, is the right call and belongs in the body exactly as you have it.

pirapira:
Hold: this head's post-runtime latest-nonce predicate is superseded. It cannot preserve per-entry semantics for repeated authorizations or for an authority that is also the transaction sender. I am replacing it with an auth-application producer and will update this PR with same-mode, normal-routing evidence. Do not review or merge this head.

pirapira:
Closing rather than rebasing. The conflicts are routine layout, but **the implementation follows a design that was rejected**, so resolving them would produce a conflict-free branch carrying a known-wrong gate — the state most likely to be merged by accident.

The head's post-runtime `latest_nonce == signed_nonce + 1` predicate cannot work: `latest_nonce` is an aggregate over every increment from every source, so any `signed + constant` test holds only when exactly one increment touched the account. Same-authority-twice leaves `n + 2` (controls `00625`, `24628`) and authority-equals-sender leaves `signed + 2` via `fork.py:1103` plus `eoa_delegation.py:284` (representative `00002`). A post-runtime aggregate cannot validate a per-entry event.

Everything worth keeping is now on **#10635** and bead `evm-asm-hcpv7`: the 89-row population confirmed real on a normal-production parent, the three-family split, the wrong-stage pin (`stage=2, signed_nonce=0, latest_nonce=1`), the design to implement (emit at authorization application against the running as-of nonce, nonce component only), the completeness inventory of the post-hoc helper including the **BAL-derived code-effect circularity** that must be replaced rather than dropped, and the domain argument for why #10613 remains load-bearing rather than duplicated.

Also recorded there: the mask plumbing this depends on was implemented and measured **exactly neutral** (focused-936 and random-200, zero transitions either direction) on this branch's earlier commits, along with two silent consumers of the byte-20 padding assumption that were found and fixed. Recover that work from those commits rather than rewriting it.

PR #10618, opened 2026-07-26
harness: --guest-elf flag, GUEST_ELF env removed, guest sha in every run header (#10617)

### LINKED ISSUE #10617: harness: pass the guest ELF as a --guest-elf argument and remove the GUEST_ELF env path (a silently-ignored override produced three false clean sweeps)
## Pass the guest ELF as a command-line argument, and remove the environment variable

**Maintainer decision (2026-07-26), near-future work:** the sweep/check harness should take the guest ELF as an explicit **command-line argument**. The environment-variable path should be **disabled** (not merely deprecated) and the documentation updated.

## Why — the incident that motivated it

`scripts/codegen-eest-stateless-check.sh` resolves the guest as `USER_GUEST_ELF="${GUEST_ELF:-…}"`. `USER_GUEST_ELF` is the script's *internal* variable; `GUEST_ELF` is the environment name. An agent exported the internal name, so the override was never read and **three consecutive sweeps silently ran the default guest** — reporting a clean pass on 60 blocks, then on a 517-block superset, on an artifact nobody had chosen.

The wrong results were then used to argue that a 120-row FR population had been fixed on main, and a coordinator built a further (false) explanation on top of that. The rows were real the whole time. Cost: roughly an hour across two participants, plus a retracted conclusion and a corrected memory.

**An ignored environment variable is the most persuasive failure mode available**: no error, no warning, and the output is exactly what a working setup would produce. A misspelled *argument* fails loudly; a misspelled *variable* is indistinguishable from not setting one.

## What to change

1. **Add an explicit flag**, e.g. `--guest-elf <path>`, and make it the only supported way to override the guest.
2. **Remove the `GUEST_ELF` environment path.** Per the standing principle that removing a mechanism beats adding a promise: a promise depends on someone remembering, a removed mechanism does not. If a transition period is wanted, have the script **fail** when `GUEST_ELF` is set — pointing at the flag — rather than honour it.
3. **Echo the resolved guest path and its sha256 at the start of every run**, and record both in the run's TSV header alongside the existing provenance fields. This is the property that would have caught the incident regardless of the interface: a result whose artifact identity is printed cannot be silently about a different artifact.
4. Update `AGENTS.md` / the harness docs, and any agent-facing instructions that mention `GUEST_ELF`.

## Scope note

Item 3 is worth doing even before items 1-2, and is independently useful: several run artifacts already carry ELF shas in their headers, and making that universal closes the same gap for every future override mechanism.

pirapira:
Closes #10617.

## The bug, exactly

`scripts/codegen-eest-stateless-check.sh` resolved the guest as
`USER_GUEST_ELF="${GUEST_ELF:-…}"`. `USER_GUEST_ELF` is the script's **internal**
variable; `GUEST_ELF` was the environment name. Exporting the internal name was
therefore read as empty and **silently ignored** — three consecutive sweeps ran
the *default* guest and reported clean passes, and a 120-row false-reject
population was declared fixed on that evidence. A misspelled *argument* fails
loudly; a misspelled *variable* is indistinguishable from an unset one.

## Interface

- **`--guest-elf PATH`** is now the only supported override. It implies
  `--no-build`; resolves relative paths against the **caller's** cwd (the script
  `cd`s to the repo root before parsing, so a relative path would otherwise
  resolve somewhere the caller did not name — the same class of bug); **rejects a
  nonexistent path** rather than falling back to the default guest; and conflicts
  explicitly with `--bsr-witness-cap`/`--bsr-bal-cap`, which need a build.
- **`GUEST_ELF` and `USER_GUEST_ELF` in the environment are a hard error**
  pointing at the flag. Per the maintainer comment on the issue, plain removal
  would leave an old export as a silent no-op — the precise failure the issue
  exists to prevent.

## Identity — the part that closes the gap for *every* future mechanism

- Every run **echoes the resolved guest path and its sha256** before doing
  anything, labelled with how it was chosen (`--guest-elf` / `no-build-default` /
  `built`).
- Both go into **`$RUN_DIR/run-provenance.tsv`** (`schema=run-provenance-v1`)
  next to `guest_elf_bytes`, `backend`, `fixture_tag`, `repo_head`, `repo_dirty`,
  `generated`; the sha is also in `eest-baseline.txt`.
- **`eest-ab-compare.py` reads it**, so self-check 0 — *the two legs are distinct
  artifacts* — now **runs** on legs whose guest came from outside the run dir,
  where it previously printed `NOT RUN` and skipped. That is the mechanical form
  of the rule, as requested: a comparator that can refuse to join two runs whose
  guest shas disagree. Hashing the run dir's own ELF stays as the fallback for
  run dirs predating this change.
- `lhkn7-semantic-boundary-v2-sweep.py` writes `# guest_elf` /
  `# guest_elf_sha256` into its TSV header (its analyzer skips `#` lines, so no
  consumer changes).
- `spike/parity-check.sh` takes `--guest-elf` on the same terms and rejects the
  env vars — the name is gone from the repo rather than deprecated in one script
  and live in another.

## Verification

| Check | Result |
|---|---|
| `GUEST_ELF=… <script>` | exit 1, points at `--guest-elf` |
| `USER_GUEST_ELF=… <script>` | exit 1, same |
| `--guest-elf /nonexistent` | exit 1, **no fallback to the default guest** |
| `--guest-elf X --bsr-bal-cap 5` | exit 1 |
| override **honoured** | run with `--guest-elf /var/tmp/evm-asm/early44-parent-normal.elf` echoed `edfdcf8453c8c443…` = that file's sha, **not** the default's `c2f3bb3a54f85531…`; `run-provenance.tsv` agrees |
| two overridden legs | `guest ELF: base edfdcf84… != candidate c2f3bb3a… (distinct builds)` — previously skipped |
| same-build guard intact | `!! BOTH LEGS ARE THE SAME BUILD … vacuous` |
| syntax | `bash -n` on both shell scripts; `ast.parse` on all three Python scripts |

The override was proven live the way the incident should have been caught: by
reading back a sha that could only come from the artifact I named, plus the
`/nonexistent` case to show a bad override cannot pass silently.

## Gate

**Non-emitting.** No file under `EvmAsm/` is touched — the diff is 5 scripts and
2 docs — so the guest ELF sha is **unchanged by construction**
(`gen-out/stateless_guest.elf` = `c2f3bb3a54f85531a39c7c7fa8fb4f00907d5d4a4d7aaeaa9ec7e8985533be24`,
identical before and after). No A/B sweep is owed for a change that cannot move a
byte of the guest.

**Not run:** `check-axioms.sh` — it cannot run in cache mode (#10537), so CI's to
run; no Lean proof is touched either way. Spec pin unchanged
(`execution-specs` `e5a8caf1b`).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,6 +29,8 @@ jobs:
         run: scripts/check-roundtrip-coverage.sh
       - name: EEST sampler span check (negative control on the --random guard)
         run: scripts/check-eest-sampler-span.py
+      - name: Guest-ELF override guard (negative control: an ignored override ran the default guest, #10617)
+        run: scripts/check-guest-elf-override.sh
       # Conventions-as-fitness-functions (Phase 5 / R-D1). Pure source scans,
       # no build needed — kept cheap on the PR push path. See AGENTS.md
       # "Architecture fitness functions".
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
         run: scripts/check-roundtrip-coverage.sh
       - name: EEST sampler span check (negative control on the --random guard)
         run: scripts/check-eest-sampler-span.py
-      - name: Guest-ELF override guard (negative control: an ignored override ran the default guest, #10617)
+      - name: "Guest-ELF override guard (negative control; an ignored override ran the default guest, #10617)"
         run: scripts/check-guest-elf-override.sh
       # Conventions-as-fitness-functions (Phase 5 / R-D1). Pure source scans,
       # no build needed — kept cheap on the PR push path. See AGENTS.md
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -319,6 +319,31 @@ Verified `Program`s are emitted to RV64 ELFs and run on `ziskemu` — roadmap in
 line) is the build-time gate for `emitInstr` drift. The conformance harness for
 guest changes is `scripts/codegen-eest-stateless-check.sh` (EEST A/B).
 
+#### Choosing the guest ELF, and knowing which one you ran (GH #10617)
+
+Override the guest with the **flag** `--guest-elf <path>`; it implies
+`--no-build`. There is no environment override: `GUEST_ELF` (and its internal
+twin `USER_GUEST_ELF`) now make the script **fail** and point at the flag.
+The variable used to be read as `USER_GUEST_ELF="${GUEST_ELF:-…}"`, so exporting
+the internal name was silently ignored — three consecutive sweeps reported clean
+passes on the default guest, and a 120-row false-reject population was declared
+fixed on that evidence. A misspelled argument fails loudly; a misspelled
+variable is indistinguishable from an unset one.
+
+Every run now echoes the resolved guest path and its **sha256** before running
+anything, and records both (plus `repo_head`, `repo_dirty`, backend and fixture
+tag) in `$RUN_DIR/run-provenance.tsv`; the sha is also in
+`eest-baseline.txt`. `scripts/eest-ab-compare.py` reads that file, so
+self-check 0 (“the two legs are distinct artifacts”) now runs even when a leg's
+guest came from outside the run dir instead of being skipped. The same
+`# guest_elf` / `# guest_elf_sha256` header is written by
+`scripts/lhkn7-semantic-boundary-v2-sweep.py`, and `scripts/spike/parity-check.sh`
+takes `--guest-elf` on the same terms.
+
+**Before believing any result that depends on an override, read back the sha the
+run printed.** An artifact that is not what it says it is has been the shape of
+every measurement failure on this harness so far.
+
 ## Architecture fitness functions (`scripts/check-*.sh`)
 
 The `scripts/check-*.sh` suite **is** a set of *architecture fitness
--- a/docs/sasm-howto.md
+++ b/docs/sasm-howto.md
@@ -595,10 +595,16 @@ The pattern of `ChainIdSAsm.lean` / `ActiveForkSAsm.lean`:
    - baseline ELF: `git stash -u`, then
      `lake exe codegen --program stateless_guest --halt linux93 -o gen-out/<base>`,
      `git stash pop`; candidate ELF likewise from the branch;
-   - per leg: `GUEST_ELF=<elf> EEST_RUN_DIR=<dir>
-     scripts/codegen-eest-stateless-check.sh --no-build --backend spike
+   - per leg: `EEST_RUN_DIR=<dir>
+     scripts/codegen-eest-stateless-check.sh --guest-elf <elf> --backend spike
      --random --seed 4242 --limit 200 --jobs 8 --quiet-passes`
-     (note: `--all` bypasses `--limit`);
+     (note: `--all` bypasses `--limit`; `--guest-elf` implies `--no-build`.
+     The guest is chosen with the **flag** — the old `GUEST_ELF` environment
+     variable is removed and setting it is now an error, because in one of its
+     two spellings it was silently ignored and ran the default guest, GH #10617.
+     Each leg echoes its resolved guest path and sha256 and records both in
+     `<dir>/run-provenance.tsv`; **check that the two legs' shas differ** before
+     believing any delta);
    - diff the per-case `PASS/FAIL/ERROR/BUDGET` labels between legs —
      **failures are acceptable only if identical on both legs**; also
      sanity-diff the two `.s` files (only the expected expansion).
--- a/scripts/check-guest-elf-override.sh
+++ b/scripts/check-guest-elf-override.sh
@@ -1,0 +1,98 @@
+#!/usr/bin/env bash
+# check-guest-elf-override.sh -- negative control on the guest-ELF override guard.
+#
+# GH #10617. The harness used to resolve its guest as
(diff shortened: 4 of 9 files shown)


pirapira:
Good PR, and it goes further than the issue asked in the direction that matters most — the sha in every run header, wired into `eest-ab-compare.py` so self-check 0 now runs instead of being skipped when a leg's guest came from outside the run dir. That converts the rule from something a person remembers into something the comparator enforces. Capturing `INVOCATION_CWD` before the `cd` so a relative `--guest-elf` resolves against the caller's cwd is exactly the right instinct, and the reason given for it is the right reason.

Two things before I label. The first is a real gap between the PR's claim and the code.

## 1. The check is a non-empty test, not a presence test

```bash
for stale_var in GUEST_ELF USER_GUEST_ELF; do
  if [[ -n "${!stale_var:-}" ]]; then
```

`-n` is false for an **empty** export, so `export GUEST_ELF=` passes straight through and the run proceeds on the default guest — the exact failure this PR exists to remove. Verified:

```
$ GUEST_ELF="" ; [[ -n "${GUEST_ELF:-}" ]]   → false   (no error, run continues)
$ GUEST_ELF="" ; [[ -n "${GUEST_ELF+x}" ]]   → true    (errors, as intended)
```

The PR body and the commit message both state that an empty export errors, so the claim and the code disagree. Use `${!stale_var+x}` (or `[[ -v $stale_var ]]`) so **presence** is what triggers it. An empty export is still someone attempting an override — most likely from a wrapper or profile that computed a path and got an empty string, which is precisely the case where a silent fallback is most misleading.

Worth noting how this slipped: the verification presumably exported a *path* rather than an empty string, so the test that would have distinguished the two forms was the one not run. Same shape as the finding this PR documents.

## 2. The internal variable is still named `GUEST_ELF`, so the guard can trip on itself

After the guard, the script does `GUEST_ELF="${GUEST_ELF_OVERRIDE:-$REPO_ROOT/gen-out/stateless_guest.elf}"`. That is a plain shell assignment today and therefore harmless — but the name it reuses is one of the two names the guard rejects. Any future `export` of it, or any child invocation that inherits it, would make the script fail its own guard with a message telling the caller to stop doing something the script did to itself. `parity-check.sh` now takes `--guest-elf` on the same terms, so there is more than one script in this family.

Rename the internal one — `RESOLVED_GUEST_ELF` or similar — so the rejected names and the used names are disjoint sets. Then the guard is un-trippable by our own code by construction rather than by nobody having exported it yet. This is the same principle the PR itself invokes: removing the mechanism beats remembering not to use it.

## Everything else

`--guest-elf` implying `--no-build` is right (an override with a build step is ambiguous about what ran), the readable-file check plus absolute-path canonicalisation is right, the failure message naming the flag *and* echoing the path the caller supplied is a nice touch, and the `AGENTS.md` and `docs/sasm-howto.md` updates carry the incident rather than just the interface, so the next person understands why the flag exists. Recording `repo_head` and `repo_dirty` alongside the guest sha is more than I asked for and worth having.

Confirm the ELF sha is unchanged (this touches no emitted bytes) and I will label on the two fixes.

pirapira:
**Correction to my own review, then labelling.**

My finding 1 was already fixed when I posted it. I reviewed `dea4c546e`; the presence-test fix was pushed at 16:33 as `57f672dcf`, and the PR body I read had been written against the earlier commit. So I reported a defect against a stale head and attributed a claim/code mismatch that no longer existed. That is the same failure I have been flagging all day — reading an artifact that is not the one in question — and the remedy is mechanical rather than a resolution to be careful: **a review should state the commit sha it reviewed**, the way a sweep states its ELF sha. Then a stale review is self-evident to both parties instead of arriving as an accusation. I will do that from here.

The diagnosis was still worth having, because the empty-export case had genuinely not been exercised by the verification set at the time.

## What is now in, and why it is stronger than what I asked for

- Presence test (`${VAR+x}`) rejecting an empty export, on both spellings.
- Internal variable renamed to `RESOLVED_GUEST_ELF`, so the rejected names and the used names are **disjoint by construction**. `parity-check.sh` already used a disjoint internal name (`ELF`) and needed nothing — verified rather than assumed.
- A CI gate covering five cases: empty `GUEST_ELF`, non-empty, an export beside a correct `--guest-elf`, empty `USER_GUEST_ELF`, and `--guest-elf` naming a missing file. Plus an assertion that neither harness script *assigns* to a rejected name — which holds fix 2 in place by a gate rather than by memory.

**And the negative controls on the gate are the part I want to single out.** Regressing the guard back to `-n` makes the new gate fail on exactly the empty case and pass the other four — precisely the signature of the bug — and renaming the internal variable back makes the disjointness assertion fail. So each assertion is *demonstrated able to fail* rather than assumed to be. A check that cannot fail is a promise wearing a mechanism's costume, and this one has been shown to have teeth in both directions.

## Gate

ELF sha unchanged at `c2f3bb3a…`, and the argument for why is structural rather than empirical: the branch touches 9 files — 6 scripts, 2 docs, 1 workflow — and **nothing under `EvmAsm/`**, so the guest cannot have moved a byte. `check-axioms.sh` disclosed as unexercised per #10537.

Labelling `merge!`.

pirapira:
Build green on `ab076aca3`, all four checks passing. Labelling `merge!`.

Worth recording what green means on this branch, since the reason it was red is that green was not green: the `Build` check now **exists by name**. An unquoted colon-space in a step name made `build.yml` fail to *parse*, so the workflow did not run at all — a broken gate presenting as no gate rather than as a failing one, which is the same shape as the ignored override this PR exists to remove. Reviewed at `ab076aca3`.

PR #10633, opened 2026-07-26
fix(codegen): apply BLS MSM discounts to direct precompiles

### LINKED ISSUE #10632: gas: BLS MSM precharge path uses raw k*12000 / k*22500 with no discount table — overcharges and can reject valid transactions
Found during the walk-five precompile crosswalk (#10631). Filing separately because a live divergence recorded as an inventory row gets read as inventory.

## The divergence

Amsterdam computes BLS G1/G2 multi-scalar-multiplication gas as

```
k * PRECOMPILE_BLS_G1MUL (or G2MUL) * discount(k) / MULTIPLIER
```

with a `K_DISCOUNT` table through `k = 128` and the maximum discount beyond it.

The guest implements this **twice**:

- **CALL path — correct.** `ChildFrameHandlerTails.lean` (~954-969 for G1, ~1114-1129 for G2) goes through `bls12MsmCostAsm`, which applies the discount table and the divisor.
- **Simple-transfer / MTx precharge path — wrong.** `BlockVerdictSimpleTransferPrecompileGas.lean:214-216` and `:221-223` use raw `k * 12000` and `k * 22500` after length validation, with **no discount table and no division by 1000**.

## Consequence and direction

The precharge path **overcharges whenever `discount(k) < 1000`**, so a valid transaction calling BLS MSM directly can be rejected. This is the **false-reject** direction — a correctness defect, not a soundness hole. Stating that explicitly so it is neither escalated as unsound nor dismissed as harmless: it cannot admit an invalid block, and it *can* reject a valid one.

This is a **live** path, distinct from #10527 (an unused helper).

## The shape, and why it decides the fix

The same quantity is computed **correctly in one path and wrongly in another** — two implementations of one spec rule, one of which drifted. That is the dual-implementation pattern this project has paid for repeatedly, most visibly in the dispatcher-unification work where a dual implementation was identified as the root of single-lane whack-a-mole.

So the aligned fix is to **call the existing helper** from the precharge path, not to correct the inline arithmetic. Correcting the inline version would leave two implementations that agree today and nothing preventing the next divergence. Removing the duplicate is what makes the drift unexpressible.

## Open question

Whether a single transaction can traverse **both** paths, or whether simple-transfer precharge and the CALL path are mutually exclusive. If both are reachable for one transaction, the discrepancy is worse than an overcharge.

## Gate when fixed

The FA argument is structural rather than statistical: using the spec's own formula through the helper the spec-aligned path already uses cannot admit anything the spec rejects. Note that **fixture sweeps are currently suspended fleet-wide** pending the read/write container convergence (#10619), so this should be gated on the structural argument plus a targeted BLS-MSM case, and re-measured after that convergence lands and the corpus is re-baselined.

pirapira:
## Summary

Fixes #10632. The direct/simple-transfer BLS12-381 G1/G2 MSM precharge used
the undiscounted raw formulas `k * 12000` and `k * 22500`, while the internal
CALL precompile path already used the EIP-2537 discount-table formula.

The direct path now calls the existing `bls12MsmCostAsm` helper. Its failure
target is made an explicit helper argument so the same checked formula can
serve both the child-allotment and direct-precharge contexts; no second copy of
the arithmetic is introduced.

The direct and internal-CALL paths cannot both price one invocation: an active
top-level precompile target has no EVM code to execute a child `CALL`, whereas
a contract which performs a child `CALL` takes the child precompile path rather
than the direct simple-transfer path.

## Evidence

No full sweep was run: fleet-wide sweeps are suspended. This change is instead
structurally tied to the already-used checked CALL formula and tested on direct
precompile fixtures.

- Direct G1 MSM discount-table-128 and G2 MSM discount-table-100 fixtures:
  `rc=0`, `succ=1`, and byte-exact manifest outputs.
- Byte tie: `scripts/check-asm-to-program.sh` passed.
- Region map: `scripts/check-region-map.sh` passed.
- `lake build` passed.
- Final linked ELF: `0d9d2ab8eb173c638cee17c231a94509688da2cc533e25b95a3256c15c3dfd21`.
  Emitted source/object: `73b062c721a97802509d00a28af6bd3f967505a07234de6aa8e68eea0ab69c46` /
  `a597a176ba34475b5165c9dcf94e2a87973a3f19c3fecb884ebbbde5fa30086e`.
- A second independent relink produced byte-identical `.text` and `.data`:
  `.text` size `0x61608`, SHA `64ed2bcad5dcf17531e840bf2c6adb8907c561c6e45bd8db994f64c1ed289e30`;
  `.data` size `0x5370`, SHA `78e0da21b8ec28de34188a5b41bf90cd3e121f25c4bc92eb31085a71524bbdad`.
- The linked direct sites resolve the same symbols as the CALL path:
  `bls12_g1_msm_discount_table @ 0xa3003b70` and
  `bls12_g2_msm_discount_table @ 0xa3003f70`; both use the helper's checked
  multiply and final `/1000` discount calculation.


pirapira:
Reviewed at `5c93965ef`. The unification is the right shape and the refactor is done carefully — but I am holding the label on one static question, because the empirical net that would normally catch it is suspended.

## What is well done

**The helper refactor preserves the existing call sites exactly.** Changing `bls12MsmCostAsm`'s parameter from `tag` to `failureLabel` and moving the `.L{tag}_bn254_fail_allot` construction out to the callers means the two CALL-path sites emit the identical label they emitted before. That is the difference between generalising a helper and perturbing its existing users, and it is why the CALL path needs no re-verification.

**Reuse rather than correction.** The precharge sites now reach the same helper and the same linked discount tables as the CALL path, so the duplicate is gone rather than repaired. Two implementations that agree today was the thing being removed.

**The mutual-exclusivity answer is a real answer**, not a hedge: a direct active-precompile target has no EVM code from which to issue a child CALL, so the two pricing paths cannot both apply to one invocation. That closes the question of whether the divergence compounded.

## What I need before labelling

**1. Document the helper's register ABI.** The new call sites do `mv x18, t5` before and `mv t6, x16` after, so the contract is *input in x18, output in x16* — and it is now invisible at the call site and used from two different files. Put it in a docstring on `bls12MsmCostAsm`: inputs, output, and clobbers. A shared helper with an implicit register contract is a trap for the third caller, and there will be a third caller.

**2. Enumerate the helper's clobbers and confirm none are live across the call at the two new sites.** This is the specific risk: the precharge code holds values in `t2`-`t6` around the call (`la t2, bv_simple_transfer_tx` before, `mv t6, x16` after), and the helper touches at least `x16`, `x17`, `x22`, `x23`. If it also writes a `t`-register that the surrounding code needs afterwards, the result is wrong in a way that assembles and links cleanly.

That is not hypothetical — a record pointer held in `t5` across a callee produced four crashes in this repo earlier today, on a change whose author had also verified the surrounding logic. And the reason it matters more than usual here: **fixture sweeps are suspended**, so the population run that would surface a clobber is unavailable. Two targeted fixtures passing does not bound this, because whether a clobber bites depends on which path runs afterwards.

Do it statically: list what the helper writes, list what is live across each of the two new call sites, and show the intersection is empty. If it is not empty, save and restore, or prefer a callee-saved register — the register choice makes the bug unexpressible, a save/restore depends on the next person remembering.

## Evidence noted

Direct G1 discount-table-128 and G2 discount-table-100 returning `rc0 succ1` with exact expected bytes is the right targeted evidence for a gas-formula change, and both paths resolving the *same linked* tables is the structural verification I asked for rather than a claim about the `.s`. Two-pass convergence at text `0x61608` / data `0x5370`, byte-tie and region-map passing, with the absent full sweep disclosed and attributed to the suspension.

pirapira:
Both answers received and the static argument is the right form — reviewed at `627382476`.

The clobber intersection is shown empty rather than asserted: the helper writes `x16 x17 x22 x23` and reads `x18` only; `t2 t3 t4 t5` are dead immediately after the length validation and the call; the sole post-call value is the one `mv t6, x16` establishes, after which control jumps to publish. Disjoint from both the caller live set and `t6`. And confirming the failure edge reaches the standalone precompile-fail publication path matters as much as the success edge — a failure label that assembles but lands somewhere unintended is the version of this bug that a targeted success fixture cannot see.

Labelling `merge!`.

One note for whoever adds the third caller: the register ABI is now documented on the helper, which is what makes that call site reviewable at all. The reason I asked for it rather than accepting the two verified sites is that this helper's contract is *implicit at the call site* — `mv x18, …` before and `mv t6, x16` after is not self-describing, and the next author will be working from the call sites rather than from this PR.

Issue #10637, opened 2026-07-26
ci: link every build unit, not just stateless_guest — a change to shared handlers broke zisk_stateless_verdict_v2 with all gates green


pirapira:
## The gates build the tree; they never link the non-guest build units

Found the hard way while implementing #10619: a change that hooks guest handlers broke the link of `zisk_stateless_verdict_v2`, and **every gate passed**.

`zisk_stateless_verdict_v2` mirrors the guest's handlers and helpers, so it contains every routine the read-container work hooks — `h_SLOAD`, `h_SSTORE`, `account_state_commit_pending`, `code_at_header_state_root`, `seed_callee_storage`, `dispatch_tx_runtime_code`, `block_verdict_withdrawal_nonstorage_effects` among them. Adding a call to a new recorder therefore introduces an undefined symbol in that unit, because the probe cannot inherit the guest's helper or data sections.

**No gate in the set links anything but the guest.** `lake build`, `check-region-map.sh`, `check-asm-to-program.sh` and `check-forbidden-tactics.sh` all passed on a tree in which a build unit did not link. So *"the guest links"* does not imply *"the units link"*, and the check has to be **link each unit**, not build the tree.

### Ask

Add a CI step that links **every** build unit, not just `stateless_guest`. Sweeping them by hand right now gives:

| unit | links? |
|---|---|
| `stateless_guest` | yes |
| `zisk_stateless_verdict` | yes |
| `zisk_step2_verdict` | yes |
| `zisk_stateless_verdict_v2` | yes (after the #10619 fix) |
| `runtime_dispatcher` | **no** |
| `runtime_dispatcher_call_probe` | **no** |
| `zisk_runtime_access_list_seeded_sload` | **no** |

### The three failures are pre-existing, and checked rather than assumed

Their undefined symbols include `bsr_addr_4788`, `evm_access_account_table` and `account_extract_nonce` — none of which the #10619 branch defines or edits; those names appear in its diff only as a prose mention and as regenerated address literals. Had `runtime_dispatcher` been linkable on main, only the branch's own new symbols would have been undefined in it.

This extends **#10532**, which already records `runtime_dispatcher` link rot on main and notes the opcode runner is not in CI. Two further units belong on that issue's list.

### Why a build-only gate is the wrong shape here

[...]

Issue #10635, opened 2026-07-26
gas/BAL: record the EIP-7702 authority nonce effect at authorization time, not from a post-runtime aggregate (early44, 29 rows; replaces #10616)


pirapira:
Replaces PR #10616, closed because its implementation followed a design that was rejected. The diagnosis and the population are sound; the fix has to be written from scratch against current main (which now includes #10613's B1 nonce replay).

## Population

89 rows, all `bv_fail=44`, the BAL-to-exec **nonstorage** forward comparator. Confirmed real and current: **89/89 scored FR on an immutable normal-production parent ELF**, not only in the census lineage.

Three families, from a faithful diagnostic that reproduced 89/89 before any field was interpreted:

- **29 rows** — balance changed and matches the BAL, but **exec nonce unchanged while the BAL declares a nonce change**. Labels overwhelmingly EIP-7702 set-code/delegation. This issue.
- **52 rows** — missing aggregated exec effect; a separate family (auth/delegation nonce, withdrawals/balance, selfdestruct/create, invalid auths).
- **8 rows** — assorted balance/nonce mismatches, small enough to name individually.

## Diagnosis: wrong stage, not wrong value

The spec increments the recovered authority's nonce once per valid authorization (`vm/eoa_delegation.py:284`), outside the `delegation_set_for` and `written_accounts` guards, so repeated authorities increment repeatedly. The BAL builder then emits a nonce change because pre-tx differs from post (`block_access_lists.py:641-650`). The BAL is **right**; the guest's exec-side nonstorage aggregation never records the increment, so the comparator sees a declared change with no matching execution effect and rejects a valid block.

Pinned on representative `24639` (`set_code_to_precompile`): the helper's gate exported `stage=2, signed_nonce=0, latest_nonce=1`, so the post-runtime `latest == signed` test skips `record_nonstorage_effect` entirely.

## The design that was rejected, and why

Changing the gate to `latest == signed + 1` is **wrong**, and it is wrong for a reason that generalises: `latest_nonce` is a **post-runtime aggregate**, counting every increment from every source. Any equality against `signed + constant` therefore holds only when exactly one increment touched the account, and two classes violate that:

- **same authority twice** — two valid authorizations leave `latest = n + 2`, so the `+1` test fails and the effect goes unrecorded again. Controls `00625` (`bal_7702_double_auth_reset`) and `24628` (`multiple_valid_authorization_tuples_same_signer_increasing_nonce`) are exactly this shape.
[...]

PR #10638, opened 2026-07-26
convergence(10619): mirror EIP-7928's three read sets at both spec lifetimes

pirapira:
Mirrors EIP-7928's three **read** sets from `execution-specs` Amsterdam
(submodule `e5a8caf1b`) into the guest, at both of the spec's two lifetimes.

> **The containers are built, populated and promoted — and NOT YET CONSUMED.**
> Nothing reads them, the net-zero loop is still in place, and **this does not fix
> early37.** Consumption is a follow-up.

## What the spec has, and what the guest now has

`state_tracker.py` keeps three read sets on `TransactionState` and three on
`BlockState`. Recorders write the **tx** level (`tx_state.*_reads.add(...)`);
`incorporate_tx_into_block` (`:832`) merges them up (`:858-861`) and then **clears**
them (`:879-881`); `build_block_access_list` (`fork.py:928`) reads the **block**
level.

| spec | guest producer | firing condition | spec ref |
|---|---|---|---|
| `storage_reads` | `storage_read_record` | **unconditional**, at the top, before writes are consulted | `:295`, `:578` |
| `account_reads` | `account_read_record` | **unconditional**, first statement of `get_account` | `:139`, `:199` |
| `code_reads` | `code_read_record` | **only** on pre-state fallthrough; skips `EMPTY_CODE_HASH` | `:263-270` |

**Three routines, not one parameterised by kind — deliberately.** The conditions are
genuinely different, and code reads are the odd one out: a shared recorder would
over-record them, and `code_reads` feeds witness code selection
(`stateless_host_exec_witness.py:182`), so a surplus entry invents a witness code
the spec never emits.

## Two levels, with a real merge and a real clear

`read_sets_incorporate_tx` merges all three sets upward as **set unions** (matching
`set.update`), then tail-calls `read_sets_discard_tx` — the clear at `:879-881`.

The clear is load-bearing: a merge without it double-counts across transactions, and
**a single-transaction test cannot observe that** — there is no second transaction to
double-count into.

`read_sets_discard_tx` also names the path at `fork.py:745-752`, where a throwaway
`TransactionState` pre-checks a system contract and its reads are deliberately
**never promoted**. With block-level-only containers there is nowhere to express
that; naming it makes "deliberately not promoted" an operation a reader can find
rather than an absence they must notice.

## Tracked vs raw accessors — the routing lives in the call graph

The spec distinguishes tracked accessors (`get_account`, `get_code`) from raw stores
(`pre_state.get_account`, `pre_state.get_code`); only the former record. So this adds
**two entries over one implementation**, twice:

- `code_read_fetch` over `witness_codes_lookup_by_hash`
- `account_at_header_state_root_tracked` over `account_at_header_state_root`

Execution call sites route to the tracked entry. **Verification sites keep the raw
one** — `block_verdict` reaches it 6 times and `bal_code_preimages_valid` once; those
are the guest checking a BAL against witnessed state, not execution touching an
account. Recording them is an **over**-record, which is the direction that yields a
false ACCEPT.

Counts **measured in the emitted `.s`**, with the attributed breakdown reconciled
against an independent flat count (11 + 10 = 21):

- **tracked, 11 calls / 7 routines:** `h_CALL` ×4, `h_SELFDESTRUCT` ×2, and one each
  in `call_frame_descend`, `create_frame_descend`, `dispatch_tx_runtime_code`,
  `seed_callee_storage`, `block_verdict_withdrawal_nonstorage_effects`
- **raw, 10 calls / 5 routines:** `block_verdict` ×6, `bal_code_preimages_valid`,
  `code_state_final_balance_nonzero`, and 2 deferred EIP-7702 auth sites

Per-site table with `a2`/`a3` evidence for each: #10619 (comment).

**Why a split rather than a classification table.** Five separate instruments
mis-measured this one caller set while I built it: a single-call-form `jal ra` grep
(verified Programs emit `x1`), a lowercase-only label regex that skipped every
uppercase handler, a hand table that lost three rows, a source-level reading of a
parameterised generator, and an end-of-line-anchored attributor that dropped every
compound `jal x; mv y`. A table describing which callers are execution is a promise
maintained by whoever last ran the search. **The call graph is not.**

## Deferred, and marked as such in the code

The two EIP-7702 auth sites (`eip7702_authority_asof`,
`eip7702_auth_nonstorage_effects`) are execution reads that **belong** on the tracked
entry. They are left raw with an in-code marker giving the reason: that region is
being changed concurrently by #10635. While held, `account_reads` **under**-records
an authorization-only account — a direction that cannot produce a false accept.

## A bug this branch introduced and fixed

`zisk_stateless_verdict_v2` mirrors the guest's handlers, so it picked up calls to
`storage_read_record`, `code_read_fetch`, `read_sets_incorporate_tx` and
`account_at_header_state_root_tracked` **without defining any of them** — it did not
link, across three commits.

Nothing in the gate set caught it: `lake build` is green because the fault is in
emitted asm; `check-asm-to-program.sh` byte-ties converted `Program`s, not unit
links; `check-region-map.sh` pins `stateless_guest` only. It surfaced as a linker
error buried inside an A/B leg. **The guest linking does not imply the units link** —
filed as #10637.

The fix is **probe-only**, and that is checkable rather than asserted:
`stateless_guest` `.text` unchanged at `0x61da4`, `symbol-addresses.tsv` regenerates
byte-identically, and **the rebuilt guest ELF has the same sha256 as before the fix**.

## Memory

| section | merge-base `a6be5b3e4` | HEAD | delta |
|---|---|---|---|
| `.text` | `0x61510` | `0x61da4` | +1948 B (incl. `+0xf8` from main) |
| `.bss` | `0x1b2557c0` | `0x1b255820` | **+96 B** |
| `.data` | `0x5370` | `0x5370` | unchanged |

96 bytes is the whole ELF cost of six containers — twelve 8-byte slots (a count and
an overflow flag per set per level). The **arenas cost zero ELF bytes**: they are
absolute-addressed carve-outs of the pre-existing working-RAM window
`0xa0020000..0xa3000000`, not `.bss` objects. Address-space cost is 2 MiB per level,
4 MiB total, with 16.375 MiB still unallocated above them.

`MemoryBudgetGuard` does **not** cover these regions — it is a gas-affordability
guard (`memoryPerWord`, `gasCap`, `maxTotalLiveMemoryBytes`), not a footprint guard.
No pin needed raising and none "held", because none was watching. Worth a follow-up.

## Verification

**Paired A/B, seed 4242, 200 random blocks, `spike`, ELFs frozen at mode 444 and
provenanced by sha256 in each run header:**

| leg | guest | scored | FA | FR |
|---|---|---|---|---|
| base (`main` `fe0e0d532`) | `0d9d2ab8` | 200/200 | 0 | 5 |
| candidate (`460a262aa`) | `cba8acd6` | 200/200 | 0 | 5 |

**FA 0 → 0. FR 5 → 5 with identical label sets** (not merely equal counts). Join
verified: both legs cover the same 200 labels, and both are complete, so the delta is
anchored.

Negative controls run before trusting either leg: `--guest-elf /nonexistent` exits 1,
and `GUEST_ELF` in the environment is still rejected with the #10617 message.

**Gates:** `lake build` (4754 jobs), `check-region-map.sh` (`textSizeBytes 0x61da4`
verified against the ELF), `check-asm-to-program.sh` (CLEAN, 389 converted defs / 148
files), `check-forbidden-tactics.sh` (OK), plus a per-unit link sweep.

**Not run:** `check-axioms.sh` — cannot run in cache mode (#10537). **No proof is
added, weakened, or deleted by this PR.**

Refs #10619.


pirapira:
Reviewed at head, as the only reader. I checked the four places where this could be wrong in a way the gates would not catch, and found nothing to change. Labelling `merge!`.

## Where I looked, and what the answer was

**1. Register discipline at the producer call sites.** The `Storage.lean` sites save `x1`, `x10`, `x11` — exactly the three the call setup overwrites — and pass `a1` from `x12`. My concern was `x12` itself: `a2` is a caller-saved argument register a callee may clobber freely, and a record pointer held across a callee is precisely what produced four crashes in the other lane today. `storage_read_record` never touches `a2`, and its docstring states the contract: `a0` = addrHash ptr, `a1` = slotKey ptr, no result register, and *"clobbers nothing the caller can see"* with `t0`-`t6` saved and restored. So `x12` survives, the save set is correct and minimal, and — unlike the two ABI defects found elsewhere today — the contract is written at the definition rather than inferable only from the call site.

**2. Producer placement relative to the stipend guard.** The SSTORE recorder sits *after* the guard, and the comment gives the reason: `storage.py` runs `check_gas(evm, CALL_STIPEND + 1)` **before** `get_storage_original`, so an SSTORE that fails it records no read in the spec. Recording first would invent a read the spec never makes. That is the correct half of a subtle ordering, argued from the spec rather than from the guest.

**3. The read/write dedup.** The recorder is deliberately unconditional on the frame committing, citing `state_tracker.py:90-93` and `:809-826`. EIP-7928's "a slot both read and written appears only in the changes list" is discharged **where the spec discharges it** — the BAL builder's dedup against `storage_changes` — rather than by suppressing the record. That is the distinction the whole convergence rests on, and getting it backwards would have reproduced the original defect in a new place.

**4. The verified-Program insertions.** Every relative branch whose span crosses the insertion point moved by exactly 4: `BLTU` 340→344, and the `JAL`s 284→288, 228→232, 220→224, 156→160, with the `laHi`/`laLo` anchors following. These were computed by the converter and independently reproduced by `scripts/program-insert-offsets.py`, which this PR also lands. Two instruments agreeing is the standard I want here, because a missed branch would have **assembled cleanly while jumping four bytes into the middle of an instruction** — a guest that links and then misbehaves.

## Evidence, and what it does and does not establish

Two legs, both 200/200 scored, seed 4242, FA 0→0, FR 5→5, **identical FR label sets** rather than equal counts — the check that matters, since five and five is consistent with a different five. Reported separately for the two questions they answer: the merge-base leg isolates this change, the rebased leg confirms it survives contact with today's ~24 merges, including #10613 touching a file this branch edits.

**What it does not establish, and the PR says so:** the containers are built, populated and promoted but **not yet consumed**, and the net-zero loop is still in place. So the run cannot distinguish correct plumbing from inert plumbing — which is precisely why inertness was *predicted before measuring*, making any transition unambiguous. `bv38` is the first change with an expected effect, and it is deliberately in the follow-up.

## Two findings in this PR worth more than the code

**`MemoryBudgetGuard` is not an image-footprint guard.** It governs EVM linear-memory gas affordability, so no pin held and none needed raising — none was ever watching this. The set that *does* cover a layout change is named instead: `check-region-map`'s overlap and extent checks, `schemeA_matches_layout`, `stableGuestBases_length` (20→23→26 as the regions landed), `guestImageEntries_extentsOk`, and the three ELF-size pins. A reviewer needs to know which guard is load-bearing for a layout change and which is irrelevant, and this establishes both. With 16.375 MiB unallocated above the new arenas and nothing anchored in the gap — a measured margin rather than a reassurance.

**The probe fix is proven probe-only** by the strongest available means: guest `.text` unchanged, `symbol-addresses.tsv` regenerating byte-identically, and the rebuilt candidate ELF carrying the *same sha256* as before the fix. A fix that changes no shipped byte cannot have changed shipped behaviour, and a reader can check that number rather than trust it.

`check-axioms.sh` disclosed as unexercised under cache mode per #10537, with no proof added or weakened — and the `preBody` placement was chosen partly so the verified `evm_sload` proof stays untouched, which is the right way round: a placement that is both spec-correct and proof-preserving, rather than compromising either.

PR #10639, opened 2026-07-26
ci(codegen): link every verdict-family build unit, not just stateless_guest

### LINKED ISSUE #10637: ci: link every build unit, not just stateless_guest — a change to shared handlers broke zisk_stateless_verdict_v2 with all gates green

pirapira:
Closes #10637.

`scripts/codegen-stateless-link-check.sh` links `stateless_guest` and nothing else.
So a build unit that **mirrors guest handlers** can reference an undefined symbol
while every existing gate stays green:

| gate | why it misses this |
|---|---|
| `lake build` | green — the fault is in emitted asm, not in Lean |
| `check-asm-to-program.sh` | green — byte-ties converted `Program`s against `GuestAddrs`; says nothing about unit links |
| `check-region-map.sh` | green — pins `stateless_guest` only |

That is the **wrong shape**, not merely incomplete: a guest-only link check narrows
silently every time someone adds a build unit, whereas a per-unit list grows with the
units rather than with anyone's memory.

## Not hypothetical

`zisk_stateless_verdict_v2` mirrors the guest's handlers, so on #10619's branch it
picked up calls to `storage_read_record`, `code_read_fetch`,
`read_sets_incorporate_tx` and `account_at_header_state_root_tracked` **without
defining any of them**, and did not link across three commits. It surfaced only as a
linker error buried inside an EEST A/B leg — not from any gate.

**The guest linking does not imply the units link.**

## Two design points that make it a gate rather than a formality

**It checks the `.elf` EXISTS, not `codegen`'s exit status.** `codegen` can exit 0
having written only the `.s` and `.o` — which is exactly this failure mode — so an
exit-status check would pass on the very bug it exists to catch.

**The allowlist records *why* each known-nonlinkable unit is structural**, not merely
that it is. `runtime_dispatcher`, `runtime_dispatcher_call_probe` and
`zisk_runtime_access_list_seeded_sload` are spliced into the guest and legitimately
carry undefined cross-unit references (tracked in #10532). Because they are
allowlisted with reasons, a **new** failure is a regression even though the old ones
are not — so this gate is useful immediately instead of waiting on #10532.

`stateless_guest` is deliberately not in the default list: the sibling gate already
links it in the same CI group and it is the most expensive link of the set. Pass
`--units` to include it.

## Verification — by negative control, not inspection

- passes on clean `main` (3 units OK)
- **fails with exit 1** on a unit that produces no `.elf`
- **fails with exit 1 and names all five undefined symbols** when #10619's probe fix
  is temporarily reverted

The third control asserted that its edit had actually applied before trusting the
result, so a vacuous control could not have passed silently.

Skips with exit 0 when the RISC-V toolchain is absent, mirroring the other guards.
Wired into the existing `codegen_checks` group in `check-build-parallel.sh`, so it
runs on a runner that already has the toolchain and a built `codegen`.

No Lean, no proofs, no emitted-guest bytes changed.

### DIFF
--- a/scripts/check-build-parallel.sh
+++ b/scripts/check-build-parallel.sh
@@ -20,6 +20,9 @@ start() {
 
 codegen_checks() {
   scripts/codegen-stateless-link-check.sh --no-build
+  # GH #10637: the line above links stateless_guest ONLY, so a unit that mirrors
+  # guest handlers can reference an undefined symbol with every gate green.
+  scripts/check-build-units-link.sh
   scripts/check-region-map.sh
   scripts/check-guarded-handler-bytes.sh
 }
--- a/scripts/check-build-units-link.sh
+++ b/scripts/check-build-units-link.sh
@@ -31,9 +31,17 @@
 # are allowlisted below with the symbols that make them so — the point of the
 # allowlist is that a NEW failure is a regression even though old ones are not.
 #
+# Allowlisted units ARE attempted, and the exemption EXPIRES: if one of them starts
+# producing an `.elf`, the run fails with "stale allowlist" and names the entry to
+# delete. Otherwise the allowlist would be documentation formatted as code — a branch
+# that never executes, whose entries nobody can distinguish from ones that outlived
+# their reason. An exemption that cannot expire is a promise rather than a mechanism,
+# which is precisely the distinction this script exists to enforce.
+#
 # Usage:  scripts/check-build-units-link.sh [--units "a b c"]
-# Exit:   0 all expected-linkable units linked; 1 a regression; 0 with a skip
-#         message if the RISC-V toolchain is absent (mirrors the other guards).
+# Exit:   0 expected-linkable units linked AND every exemption is still warranted;
+#         1 a new link failure or a stale exemption; 0 with a skip message if the
+#         RISC-V toolchain is absent (mirrors the other guards).
 set -uo pipefail
 
 REPO="$(cd "$(dirname "$0")/.." && pwd)"
@@ -49,10 +57,22 @@ cd "$REPO"
 # one exists — linking only the guest is the wrong SHAPE, not merely incomplete,
 # because its scope silently narrows every time someone adds a build unit, whereas a
 # per-unit list grows with the units rather than with anyone's memory.
-UNITS_DEFAULT="zisk_stateless_verdict zisk_stateless_verdict_v2 zisk_step2_verdict"
+#
+# The known-nonlinkable units below are INCLUDED in this sweep on purpose. Listing
+# an exemption without ever attempting the unit would make the allowlist
+# documentation formatted as code: the branch would never execute, and nobody could
+# tell a year from now whether the entry still describes reality or outlived its
+# reason.
+UNITS_DEFAULT="zisk_stateless_verdict zisk_stateless_verdict_v2 zisk_step2_verdict \
+runtime_dispatcher runtime_dispatcher_call_probe zisk_runtime_access_list_seeded_sload"
 
 # Structurally non-linkable; NOT failures. Reason recorded so a reader can tell an
 # expected gap from a regression.
+#
+# THE EXEMPTION EXPIRES. If one of these starts producing an `.elf`, the run FAILS
+# with "stale allowlist" and names the entry to delete. An exemption that cannot
+# expire is a promise rather than a mechanism — which is the distinction this whole
+# script exists to enforce, so it has to hold for the script's own exemptions too.
 declare -A KNOWN_NONLINKABLE=(
   [runtime_dispatcher]="spliced into the guest; references bsr_addr_4788, account_extract_nonce and other cross-unit symbols (documented in scripts/gen-symbol-addresses.py)"
   [runtime_dispatcher_call_probe]="same family as runtime_dispatcher"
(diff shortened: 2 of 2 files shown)


pirapira:
Reviewed at the current head. The gate is right in the two places that matter most, and there is one real gap.

**What is right, and worth keeping in the docstring:** the check is **the `.elf`'s existence**, not codegen's exit status — because codegen exits 0 having written only the `.s` and `.o`, which *is* the failure mode. A gate that trusted the exit code would have passed on exactly the tree that broke. And the explanation of why a guest-only link check is the wrong *shape* rather than merely incomplete — its scope silently narrows each time a unit is added, whereas a per-unit list grows with the units rather than with anyone's memory — is the reasoning a future reader needs in order not to "simplify" this back.

## The gap: the allowlist is unreachable, and cannot retire itself

```bash
UNITS_DEFAULT="zisk_stateless_verdict zisk_stateless_verdict_v2 zisk_step2_verdict"

declare -A KNOWN_NONLINKABLE=(
  [runtime_dispatcher]=...
  [runtime_dispatcher_call_probe]=...
  [zisk_runtime_access_list_seeded_sload]=...
)
```

**None of the three allowlisted units is in the default unit list**, so on a normal CI run the allowlist branch is never reached — it is documentation formatted as code. That is a smaller problem than it looks (nothing breaks) and a bigger one than it reads (a reader sees three units accounted for and a gate that never attempts them).

And the ordering makes it **non-self-retiring**:

```bash
if   [ -f "$OUT/$u.elf" ];              then echo "  OK    $u"
elif [ -n "${KNOWN_NONLINKABLE[$u]+x}" ]; then echo "  KNOWN $u — ..."
```

An allowlisted unit that *starts* linking takes the first branch and reports `OK`. Nothing ever says the exemption is stale, so in a year nobody can tell whether `runtime_dispatcher` is still structurally non-linkable or was fixed and the entry outlived its reason. That is the shape where known problems go to be forgotten.

**Both are a few lines:** put the three allowlisted units in the default list so they are actually attempted, and when an allowlisted unit produces an `.elf`, say so loudly — `STALE ALLOWLIST — <unit> now links; remove its entry`. Failing on that is defensible; warning is the minimum. You proved the gate can fail by reverting the probe fix and confirming exit 1 with all five symbols named — apply the same standard to its exemptions, since an exemption that cannot expire is a promise rather than a mechanism.

Everything else stands: wiring into `check-build-parallel.sh` with the reason inline, the toolchain-absent skip mirroring the other guards, and `stateless_guest` deliberately excluded because `codegen-stateless-link-check.sh` already links it in the same group and it is the most expensive link of the set.

Fix those and I will label without another round.

pirapira:
Both fixes in, and the verification is the part worth naming. Labelling `merge!`.

The allowlist is now **reachable and self-retiring**: the three known-non-linkable units are in the default sweep so the branch actually executes, and the ordering is inverted — the allowlist is consulted *before* the `.elf` check, so a unit that starts linking reports `STALE ALLOWLIST` and fails rather than quietly reporting `OK`. An exemption that cannot expire is a promise; this one can.

And it was confirmed by **pulling the Build run log and finding the gate's own output lines** — three `OK` for the verdict-family units, the `KNOWN` lines for the allowlisted ones, then the final `OK` — rather than inferred from a green tick. That matters more than usual here, because the defect I found was precisely a branch that never ran: a green check would have looked identical either way. Checking that the gate *executed* is the same discipline the gate itself enforces on codegen, applied one level up.

Worth recording what this PR is, beyond the script: it closes a gap that let a build unit stay broken across three commits with `lake build`, `check-asm-to-program` and `check-region-map` all green, surfacing only as a linker error buried inside an EEST A/B leg. The docstring's reasoning — that a guest-only link check is the wrong *shape* rather than merely incomplete, because its scope narrows silently each time a unit is added while a per-unit list grows with the units — is what should stop this being simplified back.

PR #10641, opened 2026-07-26
convergence(10619): the storage_reads container as bv38's source — stride, probe, re-point, and the net-zero deletion

pirapira:
## Three commits, in dependency order

| | |
|---|---|
| 1 | **Parameterise the routine over its entry stride** — a partial re-point becomes unexpressible |
| 2 | **Fix its probe** (GH #10642) — non-canonical RLP, red and ungated |
| 3 | **Re-point bv38 at the `storage_reads` container** |
| 4 | **Truncate the aborted tx's storage rows** instead of net-zeroing them — completes #10619 tier one |

They are one PR because each depends on the one before: the re-point needs the
stride argument to exist, the probe was the only behavioural instrument on the
routine, and the net-zero loop could not go until bv38 actually read the
container.

## Commit 4: the net-zero deletion — a truncation, not a bare deletion

`.Lbv_tx0_storage_revert` walked an aborted top-level transaction's storage
exec-log rows setting `current := original`. It existed for **two** reasons, both
recorded in its own note:

1. rows **net-zeroed** so the change comparators see no change for a
   touched-but-aborted account;
2. rows **kept** rather than truncated so the slots stayed "accessed" for the
   recipient `storage_reads` check — *"Truncating the rows instead would drop the
   aborted tx's READ slots -> bv38"*, verbatim.

Reason 2 is gone; that is what commits 1–3 bought. **Reason 1 is not.** A bare
deletion would leave the aborted transaction's SSTORE values in the write log as
apparent changes the BAL never declares. So the loop is replaced by a
four-instruction **truncation**, which discharges reason 1 more directly — no rows
means no changes — and is what the spec does: `restore_tx_state`
(`state_tracker.py:809-826`) restores only the WRITE structures and leaves the
read sets alone, and `frame_return` already truncates a reverted child the same
way.

`.text` **shrank** 0x061da8 → 0x061d68 (`0x40`, 16 instructions) — the first net
removal of emitted bytes on this branch, and the measurement that the collapse
went away rather than becoming dead code.

Random 200 (`957f5fe3`): 200/200, errored 0, budget 0, 195 full / 195 succ, the same
five rows. Proceed band. **Stated precisely**: this path runs only when a
top-level transaction REVERTs or exceptionally halts (`a4 == 0`), and I have not
measured whether any of these 200 blocks contains one — so the run says "nothing
broke dramatically", not "the truncation was exercised".

## Commit 3: the re-point

```asm
li a3, 0xa0630000               ->  li a3, 0xa1ba0000
la t0, evm_env; ld a4, 448(t0)  ->  la t0, storage_reads_count; ld a4, 0(t0)
li a5, 128                      ->  li a5, 64
```

The two probe call sites deliberately **stay at 128** — they seed their own
128-byte exec log. Two strides coexisting in one routine is exactly what commit 1
bought.

The container mirrors the spec's `BlockState.storage_reads` **set**, which
`restore_tx_state` (`state_tracker.py:809-826`) does not touch, so a read taken
inside a frame that later reverted is still present. That retention **is** the
divergence #10619 exists to remove. The container is exactly the exec log's key
prefix, and this comparator loads only bytes 0..63 — never `original@64` or
`current@96` — so the *comparison* is unchanged and only the *population* moves.
A semantic change, not a refactor.

**The skip guards are deliberately not re-pointed.** They gate on
`bvgr_arena_tx_count` and the exec-log count at `evm_env+448`, which express *is
the replay a complete witness*, not *which container holds the reads*. An empty
container with a claimed read now **rejects** — spec behaviour, in the
false-reject direction #10619 authorises.

`.text` did **not** move (0x061da8 → 0x061da8), verified: both `li` expansions are
`lui`+`addi`, both `la` are `auipc`+`addi`. `symbol-addresses.tsv` byte-identical,
so no layout regen.

**FA, structurally:** retaining a read the spec also retains cannot admit a block
the spec rejects, because `build_block_access_list` emits exactly those reads. No
sweep improves that argument.

### The A/B is a FAILED run by the pre-registered criterion

Four frozen 444 ELFs outside the repo, seed 4242, spike, 200 each, every leg
200 selected / 200 ran / errored 0 / budget 0:

| ELF | what | full / succ | failing rows |
|---|---|---|---|
| `245ef287` | pre-re-point (commit 2 tip) | 195 / 195 | 1, 9, 40, 63, 124 |
| `0b450a6d` | **this change** | 195 / 195 | **same rows, same `bv_fail` multiset** |
| `199e8540` | control: forced reject **after** the guards | 195 / 195 | same rows |
| `71e7095f` | control: forced reject **before** the guards | **183 / 183** | +12 rejects |

The prediction recorded on #10619 before measuring was that **identical label sets
mean the re-point did not take effect**, and that is how this is reported — a
failed run, not a clean one.

The two controls locate why. The region **is** entered: forcing a reject before
the guards costs 12 rows. Forcing one *after* them costs nothing. So on these 200
rows every entering block is stopped by one of the four guards and the compare
never executes. **The instrument is blind to this change, so it was equally blind
to the change being wrong.**

Landed on the maintainer's standing rule that ~5/200 is the proceed band and the
random-200 is a catastrophe detector rather than a correctness gate.

A fifth run deserves recording as a **negative result about my own method**: an
earlier control forced `a4 = 1` against the all-zero container, which I reported as
a guaranteed mismatch. It is not — with an empty recipient `storage_reads` the
routine accepts regardless, so the negative was observed in a condition where the
positive could not occur. Retracted before it was acted on.

### Proof coverage narrows; nothing becomes false

The verified scan theorems are stated at stride 128, so they now cover the two
probe call sites and **not** this one. Generalising them over the stride is 67
`16 *` sites and is filed as **#10644** rather than bundled here.

---

## What this changes

`bal_storage_reads_in_exec_log` (the bv38 comparator) baked the exec log's
**128-byte entry width** into two instructions. Re-pointing it at #10638's
64-byte `storage_reads` container was therefore a *caller-side discipline*:
three call sites that had to move together, where a partial re-point would aim a
64-stride scan at 128-byte entries and read the exec log's `original@64` as the
next entry's `addrHash` — **valid-looking garbage in the false-accept
direction**, not a fault.

The stride now travels **with** the base and count, in `a5`. A caller cannot
supply a log without also supplying that log's entry width, so a partial
re-point is **unexpressible** rather than merely forbidden.

This is prep, not the re-point: all three call sites still pass `128`.

## The routine

Fixture first, then `scripts/asm_to_program.py rewrite` — never hand-edited.

```asm
slli t3, t2, 7     ->  mul t3, t2, s5
addi t3, t3, -128  ->  sub t3, t3, s5
mv   s6, a1        ->  mv  s5, a5
```

`s5` was saved and restored by the existing prologue and **never used**, so
parking `a5` there costs no frame growth. `a5` cannot itself carry the stride
through the loop — the `rlp_walk_*` calls clobber it.

The `mv s5, a5` **reuses the slot of a dead `mv s6, a1`** (slot 12): `s6` is
re-assigned from `rlp_walk_init`'s *return* value at slot 17 before any read,
and no proof covers slot 12. That keeps the routine **byte-size-neutral**, which
is what leaves every intra-function offset — `base + 272`, the branch spans, the
reloc indices — unchanged and confines the proof edit to the one instruction
that changed.

## Proof: kept, not disabled or duplicated

Of the three authorised options on #10619, this takes neither the reshape, the
deletion, nor the duplication:

`scan_cursor_step` is restated in **register-subtract** form and still derives
from the stride-generic `scan_cursor_step_gen` from `ca27bdea`. The bridge from
its ADDI-immediate shape is **one concrete `decide`**
(`signExtend12 (-128) = -(128 : Word)`).

So the reshape the atomic write-up called for **was not needed** — the
`signExtend12` hypothesis gets discharged once per stride either way. Coverage
is preserved with no duplication and no generic side-conditions beyond the
`0 < D` already recorded.

`x21` is threaded through `scanRegs` / `scanInv` / `scanFound` / `scanAbsent`
and the round proofs as a **held constant 128**; the two `SUB` sites own it,
every other site frames it.

**Three new geometry pins** (slots 12, 64, 68) pin the stride register at each
instruction that now depends on it, so a future edit that re-bakes a constant
fails a `decide` rather than passing quietly.

## Converter fix (bundled, ~25 lines)

`asm_to_program.py rewrite` **duplicated** the generated block on this edit
rather than replacing it. `_def_span` finds only the `def …Function : String :=`
line — correct for a *first* conversion of a raw-string routine, but on a
**re-generation** (the prescribed workflow when a converted routine's fixture
changes) that span excludes the `_prog` / `_relocs` / theorem / `#guard`
declarations the block also emits, so a second copy of each got appended.

`_generated_block_span` now recognises an already-generated block — `def <prog> :
Program :=` through the trailing `#guard <prog>.length = N`, requiring the
`Function` def to sit **inside** it so a hand-written file is never matched.
Observed duplicating before the fix, replacing cleanly after.

## Emitted bytes

`.text` `0x061da4 -> 0x061da8` — **4 bytes**, the single `li a5, 128` at the
guest's one call site. The routine itself is byte-neutral; the two probe call
sites are in the probe unit. Measured from the relinked ELF, layout regenerated
by the four-step procedure rather than hand-edited.

## Gates

| gate | result |
|---|---|
| `lake build` | 4754 jobs |
| `check-asm-to-program` | CLEAN — 389 converted defs / 148 files |
| `check-region-map` | OK — all four sizes + TSV vs ELF |
| `check-build-units-link` (#10639) | OK — 3 linkable, 3 allowlisted still genuinely non-linkable |
| `check-forbidden-tactics` | OK |
| `check-axioms.sh` | **NOT run** — cannot run under cache mode (#10537) |

## Second commit: the probe red is DIAGNOSED and FIXED (GH #10642)

`scripts/codegen-zisk-bal-storage-reads-in-exec-log-check.sh` was red — scenario
1 rejecting when it must accept — and red **identically before and after** the
stride commit (paired control: stash, re-run, same line, same exit 1). It is now
**green**.

**The routine was never at fault.** The probe encoded its AccountChanges list
with RLP's **long-form** header `f8 1c` for a **28-byte** payload. The long form
is only canonical at 56 bytes or more, so `rlp_walk_init` correctly returns
status 6 (*long-list non-minimal, decoded < 56*) and the routine correctly
rejected a malformed AccountChanges. Fixed to the short form `0xdc` (`0xc0 + 28`),
total 29, with the field offsets shifted down by one.

```
  both reads present=0 (exp 0)   absent read=1 (exp 1)
  ==> PASS
```

**Correcting myself**: I first reported this probe as "well-formed by
inspection". It was not — the inspection covered the field layout and the
comparison and skipped whether the outer header was the canonical encoding of its
own length. Four readings of the *caller* had ruled out the content-window
arithmetic, both callees' register contracts, the scan's loop geometry and the
scratch lifetime, and none of them could reach this, because the defect was in
fixture data rather than in a code path. What found it in one run was publishing
the **callee's own return values** from the probe — after first confirming the
instrumented build reproduced the production verdicts, so its new fields were
believable.

**This unblocks the re-point.** The comparator is correct, so the recorded
prediction stands as stated: FR must move, identical label sets is a failed run,
FA stays 0.

### The class behind it, measured and NOT fixed here

`scripts/codegen-zisk-*-check.sh`: **687 scripts, of which 39 are referenced by
anything else** — and those references are mostly other probe scripts, not a
gate. So roughly 648 probe checks are invoked by nothing. This one had gone red
unnoticed for that reason, and I have not measured how many of the others are red
(that is 687 serial ziskemu runs). Wiring them into `check-build-parallel.sh` is
also not simply available: `codegen_checks()` runs in parallel and these need
ziskemu, which must run serial. Worth its own scoping decision rather than a
change smuggled into this PR.

## Next on #10619

1. This.
2. The re-point: call sites pass the container base/count and stride 64.
3. Net-zero deletion **last** — until reads live elsewhere it is the only thing
   preserving the touch.


pirapira:
Reviewed as the only reader. Four commits, all six checks green. Labelling `merge!`.

## What I checked, and why these four

**1. The stride reaches the routine as an argument, not a constant.** That is the point of the change and it is what makes a partial re-point **unexpressible** rather than merely forbidden — a caller cannot supply a base without the stride that belongs to it. The hazard it removes is not hypothetical and does not fault: a 64-byte-stride scan over 128-byte entries reads the exec log's `original` field at offset 64 as the next entry's `addrHash`, which is valid-looking garbage in the **false-accept** direction. Three call sites, each passing its own stride.

**2. The proof is generic in the stride, and the docstring explains the two non-obvious moves rather than leaving them to be rediscovered.** The sign-extension is a *hypothesis* rather than an internal `decide` because `bv_omega` cannot reason through `signExtend` of a symbolic value — so each instantiation discharges it on its own **concrete** immediate, keeping the `decide` concrete and kernel-checkable with no `native_decide`/`bv_decide`. And the nonlinearity (`D * m`) is removed by eliminating the `Nat` subtraction first (`count - j = m + 1`) and then generalising the product, after which the goal is linear. Concrete lemma *statements* are unchanged, derived by instantiation, so nothing downstream sees a different fact.

Worth recording what generalising bought beyond reuse: it **exposed a side-condition the concrete proof never had to state** — `scan_cursor_ne` requires `0 < D`, and at stride zero the statement is false. The concrete lemma hid it because its stride was positive by construction. That premise is now explicit.

**3. The probe fix was the probe's fault.** `rlp_walk_init` returned status 6 — long-list non-minimal — because the *fixture* encoded a long-form header (`f8 1c`) for a 28-byte payload, and RLP admits the long form only at ≥56 bytes. So the routine was right to reject and was never implicated. A suggestion to reorder, made under the wrong diagnosis, was withdrawn once the cause was known.

**4. The net-zero loop was reduced, not deleted — correctly.** My instruction said it was dead by construction once reads live elsewhere. That was true of the read-preservation purpose and **false of the other**, which the loop's own comment recorded. Keeping the live half against a coordinator's instruction was the right call; deleting a live mechanism on my say-so was the available error. The surviving purpose is documented at the site.

## Evidence, and what it does not cover

Candidate-only random-200, seed 4242, guest sha supplied via `--guest-elf` from a frozen read-only copy and recorded automatically in `run-provenance.tsv`: 200 selected, 200 ran, 0 errored, 195 full match. At the accepted ~5 band, which is a pass.

**Two coverage limits stated rather than glossed:**

- **`bv38` was not exercised.** Base, candidate and a forced control all produced the identical five failing rows with identical `bv_fail` multisets — consistent with no row in the draw reaching the check. The re-point therefore ships on its **structural** argument: `bv38` rejects when a BAL-claimed read has no support, so pointing it at a container that *retains more* reads can only find more support and can only move rows toward accept; and no comparator or verification-phase routine records into the container, established by search from `h_SLOAD`, `h_SSTORE` and the three accessors. That is the FA gate; the sample was only ever the FR gate and here it had nothing to say. **Not** to be read as neutral-and-therefore-fine, and not as a failed re-point.
- **The truncation path was not exercised either.** It runs only on a top-level revert or exceptional halt with `a4 = 0`, and whether any of the 200 contains one was not measured.

Gates: `lake build` 4754, `check-region-map` OK on all four sizes plus the TSV, `check-asm-to-program` clean at 389 defs across 148 files, `check-forbidden-tactics` OK. `check-axioms` disclosed as unexercised under cache mode per #10537.

## Follow-ups already filed rather than folded in

#10645 (the `storage_reads` producer misses `destroy_storage` — the spec records at two sites), #10646 and #10647 from the BAL audit, and the `bv38` reachability note on #10619.

Issue #10640, opened 2026-07-26
convergence: remove the EIP-7702 two-pass compute/publish split and its three compensations — none exists in execution-specs


pirapira:
## The compensations, and why they exist

`eip7702_auth_state_prepare` is called **twice per transaction** with a mode argument (`a4`), documented in its own docstring as *"0 = compute this tx's gas, 1 = success-boundary publish only."* Compute mode (`TxIntrinsicStateGas.lean:278`, `a4=0`) writes the runtime auth gas and count cells; publish mode (`BlockVerdictMtxRuntime.lean:428`, `a4=1`) writes AccountState after the transaction has succeeded. So the guest walks the authorization list twice: once to price it, once to apply it.

**`execution-specs` has no such split.** `set_delegation` is called exactly once, at `vm/interpreter.py:361`, inside `process_message` at depth 0. One loop over `message.tx_env.authorizations` validates each entry against the live `tx_state`, charges its gas, calls `set_code`, and calls `increment_nonce` — **validation, pricing and application in a single traversal**. There is no mode parameter anywhere in `eoa_delegation.py` and no second pass.

Three guest-only compensations follow from that split. **None has a spec counterpart.**

### 1. The running per-authority nonce table (compute-pass only)

**What:** a table keyed by authority holding a running nonce, so repeated authorizations on the same authority validate against the advancing value.

**Why it exists:** in publish mode the sequence advances for free — `account_state_record_auth` appends to `account_state_pending` and `account_state_auth_current` searches pending first. In compute mode the write is skipped (`.L77prep_record` is gated on the mode word), so the second authorization on one authority reads the original nonce and is rejected. The running table is *compute mode's stand-in for the pending overlay publish mode already has.*

**Spec counterpart:** none. The single traversal mutates and reads the same `tx_state`, so the "running" value is simply the state.

**Removed by:** the split going away. With one traversal there is nothing to stand in for.

### 2. The mode gate on that table

**What:** the running lookup and its updates are confined to `mode == 0`.

**Why it exists:** publish mode re-enters the same loop. If the running table were live there it would see compute-pass values and reject entries it is about to publish.

**Spec counterpart:** none — there is no second entry into the loop to guard against.

**Removed by:** the split going away. Also removed, less well, by unifying the two passes' state; the gate exists purely because two passes disagree about which structure is authoritative.

### 3. The `authority == sender` nonce branch

[...]

PR #10643, opened 2026-07-26
fix(mtx): record the EIP-7702 auth nonce effect, with its compensations documented (#10635)

pirapira:
Closes the diagnosis side of #10635. This records the pre-plus-one behaviour and its compensations; the structural one-pass repair is tracked by #10640.

## What this is, and what it deliberately is not

This records the pre-plus-one behaviour. It does not add speculative running-table or marker experiments, and it does not claim that the authority-equals-sender nonce arithmetic is correct.

## The compensations, documented at their sites

Each retained compensation is marked in TxIntrinsicStateGas.lean at lines 154-160, 204-228 and 266-272, and names the upstream repair that removes it:

1. Compute mode writes runtime auth cells but skips AccountState publication; publish mode records only after the successful-transaction boundary. The one-pass/shared transaction overlay removes this split.
2. Sender-authority nonce handling retains the pre-plus-one form and can under-report the sender BAL post nonce by one. The unified AccountState/data overlay removes this local arithmetic compensation.
3. teer_success_table is a transaction-local first-write set, not a durable nonce overlay. The shared transaction overlay removes this compensation.

The table row offset 24 is live and not spare: BlockVerdictFunction.lean lines 991-1004 read it for the recipient marker check.

## Known limitation

For authority equal to sender, the emitted nonce effect post can be one short of the BAL-declared post: the spec increments that account twice in the transaction, while this record carries pre plus one. It is currently unobserved because the sender is skip-listed from the all-account nonstorage comparator. Narrowing that skip-list or consuming the aggregate for a sender account would expose it.

## Evidence and artifact provenance

Four targeted controls were previously run on the pre-merge known-good artifact 83945464: 00002, 00625, 24628 and 24639 all matched their expected outcomes there. They were not all re-run on the post-merge image until the follow-up below.

The post-merge immutable artifact 14a70cd3 was measured directly. On the four-control manifest (denominator 4), FA=0 and FAULT=0: 00002=OK, 00625=OK, 24628=OK, 24639=FR. The earlier all-four-OK result therefore belongs only to 83945464 and must not be read as a result for 14a70cd3.

Candidate-only random-200, seed 4242, on immutable post-merge artifact 14a70cd3: denominator 200 exactly, FA=0, FAULT=0, FR=4, OK=196. This is a did-anything-explode check rather than fix evidence; the small-sample FR count is within the expected baseline noise.

Post-merge regeneration against current origin/main including 10638 and 10639 regenerated GuestAddrs, RegionMap and symbol-addresses.tsv without hand-merging generated tables. RegionMap is text 0x61e88, data 0x5370, committed storage 0xcdd800 and BSS 0x1b255820. scripts/check-region-map.sh and scripts/check-asm-to-program.sh pass.

Known-good pre-merge artifact 83945464 is retained separately and must not be conflated with the post-merge regenerated image.


pirapira:
Evidence correction: I re-ran the four controls on the exact post-merge artifact 14a70cd3 (denominator 4, FA=0, FAULT=0): 00002=OK, 00625=OK, 24628=OK, 24639=FR. The earlier all-four-OK result belongs only to the pre-merge artifact 83945464. The PR body must not imply those pre-merge results were measured on 14a70cd3; this post-merge split is now recorded explicitly.

pirapira:
Isolation result: current origin/main 8f51d7d54 built as main-only guest (SHA-256 cba8acd6015cbf0ae302b8f4728a402dc979f0774710ae4a9e028da91d2a7089) also yields 24639=FR (denominator 1, FA=0, FAULT=0). Therefore the 24639 FR on post-merge artifact 14a70cd3 is inherited from main 10638/10639, not introduced by this early44 branch. This is recorded as a separate main regression; the PR body’s artifact split remains accurate.

pirapira:
Attribution correction: the frozen pre-10638 parent artifact edfdcf8453c8c443c8b80b2e9269982b09f34e76122ff7a6de5dea0f27a348a also returns 24639=FR (denominator 1, FA=0, FAULT=0). The verified artifact matrix is therefore: 83945464 pre-merge fix artifact=OK; edfdcf84 frozen parent=FR; cba8acd6 current origin/main=FR; 14a70cd3 post-merge branch=FR. Retract the earlier claim that 10638/10639 introduced the FR. The supported conclusion is only that the fix cleared 24639 on the old base but no longer clears it on the newer base; root attribution remains open.

pirapira:
**No regression, no escalation, and this is not being pursued further.** Four artifacts, measured:

| artifact | `24639` |
|---|---|
| pre-merge fix `83945464` | OK |
| frozen pre-#10638 parent `edfdcf84` | **FR** |
| current main-only `cba8acd6` | **FR** |
| post-merge branch `14a70cd3` | **FR** |

Main rejected `24639` before #10638 and still does, so nothing regressed and #10638/#10639 are exonerated — an earlier attribution to them was withdrawn. What changed is that this branch's fix cleared it on the older base and does not on the current one. That is a **diminished benefit, not a defect**, and it is deliberately not being chased.

So what this PR is, plainly: it records the pre-`+1` behaviour, documents three guest-only compensations at their sites with the upstream repair each names, and states the authority-equals-sender under-report as a known limitation. It is **non-harmful** — FA=0 and FAULT=0 on every leg, candidate-only random-200 at `FR=4 / OK=196` against a ~5 baseline — and its durable value is the diagnosis and the documentation rather than a row count.

The structural work that would actually fix the family is tracked separately: #10640 with `evm-asm-uor19` and `evm-asm-aas60`, carrying the compute-mode visibility fact, the pending-reset audit, and the note that `teer_success_table` offset 24 is not spare.

Labelling `merge!`.

Issue #10654, opened 2026-07-26
convergence(10619): the net-zero deletion is incomplete — a second instance survives on the deposit-exception path


pirapira:
## The gap

#10641 replaced the tx-abort net-zero loop in `BlockVerdictFunction.lean` with a
truncation, on the grounds that reads now live in the `storage_reads` container so the
exec-log rows no longer have to survive for bv38's benefit.

**There is a second instance of the same loop, and it is still there.**

`BlockVerdictCreationStage.lean:475-481`, label
`.Lbvcr_deposit_exception_storage_revert`:

```asm
la t0, bv_tx_effect_snap_storage_count ; ld t0, 0(t0)
la t1, evm_env ; ld t1, 448(t1) ; li t2, 0xa0630000
.Lbvcr_deposit_exception_storage_revert:
  bgeu t0, t1, .Lbvcr_deposit_exception_settle
  slli t3, t0, 7 ; add t3, t2, t3
  ld t4, 64(t3) ; sd t4, 96(t3)      # current := original
  ld t4, 72(t3) ; sd t4, 104(t3)
  ld t4, 80(t3) ; sd t4, 112(t3)
  ld t4, 88(t3) ; sd t4, 120(t3)
  addi t0, t0, 1 ; j .Lbvcr_deposit_exception_storage_revert
```

and it carries the same now-retired justification at `:466`:

> the access rows remain and are **net-zeroed for the read checks**.

That is exactly the reasoning #10641 retired: the rows were kept so slots stayed
"accessed" for `bal_storage_reads_in_exec_log`, which as of #10641 reads the container
instead.

## Enumeration, so this is a count and not an impression

Searching the emitted programs for the `current := original` row-rewrite shape
(`ld t4, 64(...) ; sd t4, 96(...)` and siblings) returns **one** remaining hit after
#10641 — this one. The other matches on those offsets are unrelated routines
(`Blake2f`, `MptBoundedSort`, account-witness folds) with different structures.

So: **exactly one instance left**, on the deposit-exception path.

## What it needs — and the trap #10641 already hit

The loop I replaced served **two** purposes, and only one died:

1. rows **net-zeroed** so the change comparators see no change for a
   touched-but-aborted account — **still live**;
2. rows **kept** rather than truncated so slots stayed accessed for the read check —
   **dead** as of #10641.

A bare deletion here would therefore be wrong for the same reason it was wrong there:
[...]

Issue #10653, opened 2026-07-26
docs: the exec log is keyed two ways by frame depth and a wrong key form is a SILENT no-op — three rediscoveries in one session


pirapira:
**Filed because three independent pieces of work rediscovered this in a single session**, each time at real cost, and nothing in the codebase states it where a consumer would look.

## The fact

The exec log is keyed **two different ways depending on frame depth**:

- **nested callees** — byte-reversed (little-endian) key form;
- **the top-level frame** (the transaction recipient, or a CREATE-transaction contract) — big-endian `env.ADDRESS` form.

There is a converter for one direction — `bal_addr_to_exec_log_key`, taking a 20-byte big-endian address and producing the **callee** key (used this way at `BalAllAccountsTupleSequences:96-97`) — and nothing that flags when the other form is the right one.

## Why it is worth documenting rather than remembering

**Every consumer that gets it wrong produces zero matches rather than an error.** A wrong key form is a **silent no-op**: no rows match, nothing is recorded or compared, and the resulting false reject looks *unfixed* rather than *misimplemented*. Neither a build, a gate, nor a green sweep catches it.

The three independent rediscoveries in one session:

1. **the `bv38` skip lists** — the split is what makes those skips load-bearing;
2. **#10646** — the recipient/sender tuple-sequence recipe, where the wrong choice gives a catastrophic false reject;
3. **#10645** — the `destroy_storage` read-promotion loop, where a self-destructing contract is normally a nested callee **but a CREATE-transaction contract self-destructing in its own constructor is the top-level frame** (the `30ff` initcode case discussed at `Selfdestruct.lean:162`), so the loop may need both forms.

## Asked for

A short comment **at `bal_addr_to_exec_log_key` itself**, and at the exec-log row layout, stating: which frame depths use which form, that the converter produces the callee form specifically, and that a mismatch is a silent zero-match rather than a failure. A consumer reading the converter should not have to reconstruct the split from its call sites.

Optionally, the stronger structural version: make the two forms **distinguishable at the type or naming level** so a consumer must choose explicitly, rather than defaulting to whichever converter it found first — the same argument as encoding a load-bearing classification in the call graph instead of in a table about it.

---

*Written by **Claude Code** (coord agent), posting via the maintainer's GitHub account. Analysis and framing are the agent's; where a maintainer decision is recorded it is quoted and attributed explicitly.*

PR #10655, opened 2026-07-26
convergence(10654): truncate the deposit-exception storage rows too — completing #10641's net-zero deletion

pirapira:
## Completing #10641's net-zero deletion — the instance it missed

#10641 replaced the tx-abort net-zero loop on the depth-zero abort path. **There was a
second instance of the identical loop**, on the top-level CREATE deposit-exception path
in `BlockVerdictCreationStage.lean`, still rewriting `current := original` per row and
still carrying the retired justification:

> the access rows remain and are **net-zeroed for the read checks**

### The two paths are the same cleanup, and the code says so

The comment directly above it reads *"This **mirrors the depth-zero abort cleanup in
block_verdict**: execution effects/logs are rolled back to the pre-dispatch snapshots
while the access rows remain and are net-zeroed for the read checks."*

So the remedy is the one already validated there — and so is the trap. The loop serves
**two** purposes and only one died:

| | purpose | status |
|---|---|---|
| 1 | rows net-zeroed so the change comparators see no change for a touched-but-aborted account | **still live** — why this is a truncation, not a deletion |
| 2 | rows kept rather than truncated so slots stayed "accessed" for the recipient `storage_reads` check | **dead** as of #10641 |

Truncation discharges purpose 1 more directly (no rows, no changes) and mirrors
`restore_tx_state` (`state_tracker.py:809-826`), which restores only the WRITE
structures and leaves the read sets alone.

**Before changing it I checked what I had flagged as unchecked when filing #10654** —
whether the deposit-exception path has a *third* reason to keep its rows that the
tx-abort path lacked. It does not, and the mirror comment is the evidence rather than my
inference.

### Measured

`.text` **0x061e4c → 0x061e0c** — `0x40`, 16 instructions: **the identical saving to
#10641's**, which is the cross-check that it really is the same loop rather than
something that merely looked like it.

Layout regenerated by the four-step procedure on current main. The three generated files
were **reset to main and regenerated** rather than merged, per the rule for generated
layout artifacts.

### Gates

| gate | result |
|---|---|
| `lake build` | 4754 jobs |
| `check-region-map` | OK — all four sizes + TSV vs ELF |
| `check-asm-to-program` | CLEAN — 389 defs / 148 files |
| `check-build-units-link` | OK — 3 linkable, 3 allowlisted still non-linkable |
| `check-forbidden-tactics` | OK |
| `check-axioms.sh` | **NOT run** — cache mode (#10537) |

Candidate-only random 200, seed 4242, spike, frozen 444 ELF outside the repo
(`c023ebd4`): 200 selected / 200 ran / errored 0 / budget 0, 195 full / 195 succ /
root 200, the same five rows as every leg tonight (1, 9, 40, 63, 124). Proceed band.

**Run on the rebased artifact, not the pre-rebase one.** I had a 200 from before the
rebase onto current main and discarded it rather than report a number for a different
ELF.

**Stated precisely**: this path runs only on a top-level CREATE whose code deposit
exceptionally halts, and I have **not** measured whether any of these 200 blocks
contains one. So the run says "nothing broke dramatically", not "the truncation was
exercised" — the same caveat as #10641's.

### How it was found

Not by looking for it. I was sweeping exec-log consumers to check whether any already
had the key-form bug from #10653 — **none does** — and this loop turned up in the sweep
because it is a hand-rolled walk rather than a wrapper call. Sweeping by *structure*
rather than by the bug being hunted is what surfaced it; a key-form-only grep would have
closed green.

It is also my own *audit the whole contract, not the clause you fixed* rule applied to
my own merged change, which I had not done. The enumeration was one grep and belonged
in #10641.


pirapira:
**Holding review on one finding: I believe the loop this PR removes was already DEAD, which changes the PR's story from a like-for-like replacement into a behaviour change.**

Reading the pre-image on `origin/main`, `.Lbvcr_deposit_exception_storage_revert` (line 476) is reached **only by fallthrough** from the code-effect restore at line 475 — the two `.Lbvcr_deposit_exception` jump sites (431, 433) both enter at the `:467` label, above the register setup. So at loop entry the registers hold whatever line 475 left:

```
la t0, bv_tx_effect_snap_code_overflow; ld t1, 0(t0); la t0, exec_code_effect_overflow; sd t1, 0(t0)
.Lbvcr_deposit_exception_storage_revert:
  bgeu t0, t1, .Lbvcr_deposit_exception_settle
```

`t0` is the **address** of `exec_code_effect_overflow` (a `0xb…`-range pointer) and `t1` is the **overflow value** (0 or 1). So `bgeu t0, t1` compares a pointer against a small integer, is taken on the first test, and **the body never executes**. `t2` — the rows base the body indexes off — is never set on this path at all; it holds a stale value from the `lbu/sb` copy loop above.

If that reading is right, three consequences:

1. **Purpose one was not actually happening here.** The rows on the deposit-exception path were **not** being net-zeroed, so the change comparators have been seeing the aborted deposit's `current` values unmodified. That is a pre-existing defect this PR silently repairs, and it deserves to be stated as such rather than folded into "completing #10641".
2. **The truncation is a real behaviour change on this path**, not a re-expression of what the loop did. That matters for the gate: the 200-random says nothing broke dramatically, and you have already stated you have not measured whether any of those 200 contains a top-level CREATE whose code deposit exceptionally halts. So the change is **unexercised**, and it is now unexercised *and* behaviour-changing rather than unexercised and equivalent.
3. **The −0x40 cross-check needs re-reading.** Sixteen instructions removed does not reconcile with the visible edit: the deleted body is ~13 instructions and the added line is ~6 (two `la` expansions plus `ld`/`sd`), so ~7 net. Something else in the emitted output moved. "Identical saving to #10641" was offered as the evidence that this really is the same loop — if the arithmetic does not close, that cross-check is coincidence rather than confirmation, and I would rather know which before merging.

**None of this argues against the change.** Truncation still looks like the right end state, and mirroring `restore_tx_state` (`state_tracker.py:809-826`, restoring only the write structures) is the aligned justification. What I want fixed is the **description**: a comment and PR body saying "purpose one is still live, so truncate rather than delete" reads as preserving a behaviour that, on this path, was not occurring.

Please confirm or refute the dead-loop reading. If you refute it, say what sets `t2` on the fallthrough path and I will withdraw this and label the PR.

pirapira:
## Refuting the hold: **line 475 reloads `t0`**, so the loop was live

Checked against the pre-change source rather than my memory of it
(`git show HEAD~1:…BlockVerdictCreationStage.lean`). The finding does not hold, and I
think I can see exactly which line produced it.

**The reading is right about line 474 and stops one line early.** Verbatim:

```
474  "  la t0, bv_tx_effect_snap_code_overflow; ld t1, 0(t0); la t0, exec_code_effect_overflow; sd t1, 0(t0)\n" ++
475  "  la t0, bv_tx_effect_snap_storage_count; ld t0, 0(t0); la t1, evm_env; ld t1, 448(t1); li t2, 0xa0630000\n" ++
476  ".Lbvcr_deposit_exception_storage_revert:\n" ++
477  "  bgeu t0, t1, .Lbvcr_deposit_exception_settle\n" ++
```

At the end of **474**, `t0` does hold `&exec_code_effect_overflow` — a `0xb`-range
pointer, exactly as described. But **475 begins by overwriting it**:
`la t0, bv_tx_effect_snap_storage_count ; ld t0, 0(t0)` puts the **snapshot count** in
`t0`, and `ld t1, 448(t1)` puts the **live row count** in `t1`. So at loop entry
`t0` = snap count and `t1` = live count, which is the intended comparison.

**And nothing bypasses 475.** The loop label has exactly **two** references in the whole
file — its own definition at 476 and the back-edge at 481. No jump enters at the loop
label, so every path through the region executes 475. The two
`.Lbvcr_deposit_exception` jump sites (431, 433) land at the 467 label, *above* 475, as
noted — which is precisely why 475 runs.

So the loop was **live**, the truncation replaces real behaviour, and #10655 is a
behavioural change rather than dead-code removal.

### Which also means the PR's framing is still correct, and one claim in it needs no change

The `.text` saving being identical to #10641's (`0x40`, 16 instructions) is consistent
with a live loop of the same shape. Had the loop been dead, the saving would have been
the same too — dead code still occupies bytes — so **that cross-check does not
distinguish live from dead** and I should not have implied it did. It confirms *same
shape*, not *same liveness*. Correcting that inference here rather than leaving it.

### What would have made the hold correct

If 475 had been *inside* the loop body, or if any site jumped directly to 476, the stale
`t0` would reach `bgeu` and the loop would exit immediately on a pointer-vs-count
comparison. Worth stating because that is a real bug shape in this file's style —
multi-statement lines that reuse `t0` as both a scratch pointer and a loop counter — and
the reading was one line away from it rather than wrong in kind.

pirapira:
**Hold withdrawn — the refutation is correct and I verified it independently.** Line 475 on `origin/main`, which my reading stopped one line short of:

```
la t0, bv_tx_effect_snap_code_overflow; ld t1, 0(t0); la t0, exec_code_effect_overflow; sd t1, 0(t0)   -- 474
la t0, bv_tx_effect_snap_storage_count; ld t0, 0(t0); la t1, evm_env; ld t1, 448(t1); li t2, 0xa0630000 -- 475
.Lbvcr_deposit_exception_storage_revert:                                                                -- 476
```

Line 475 overwrites `t0` with the **snapshot count**, `t1` with the **live row count**, and sets `t2` to the rows base — the intended comparison. And nothing bypasses it: the loop label has exactly two references, its own definition and the back-edge, so every path through `.Lbvcr_deposit_exception` executes 475. My observation that both jump sites enter at `:467` above the setup is correct and is *precisely why* 475 runs.

**So the loop was live, and this PR is a behaviour change replacing real behaviour — which is what the PR already says.** All three consequences I drew fall with the premise: purpose one *was* happening, the description is accurate as written, and nothing is silently repaired.

**How I got it wrong, since the mechanism is worth more than the error:** I filtered the diff through a grep that dropped lines matching a six-hex-digit pattern, to cut layout noise. `li t2, 0xa0630000` matched that pattern. **My own filter removed the one line that refuted my conclusion**, and I then reasoned confidently from a view of the file I had silently narrowed. Same shape as every other "the artifact was not what it said it was" failure in this repo, self-inflicted this time, and the reason I asked for confirm-or-refute rather than acting on it.

**Author's own correction, which the hold surfaced and which I want kept:** the identical `0x40` saving confirms **same shape, not same liveness** — dead code occupies bytes too, so the measurement would have been identical had the loop been dead. It was offered as evidence distinguishing live from dead and it does not do that. Withdrawing my third point too: the byte arithmetic reconciles once line 475's `li` and the full body are counted, and in any case it was never the liveness evidence.

**One real follow-up this leaves,** and it is a bug *shape* rather than a bug: my misreading would have been **correct** had line 475 sat inside the loop body, or had any site jumped directly to `:476`. Multi-statement lines that reuse `t0` as both a scratch pointer and a loop counter are a live hazard in this file style, and a pointer-versus-count `bgeu` exits immediately and silently. Filed separately for a deliberate sweep rather than trusting that every instance happens to reload.

Labelling `merge!`.

pirapira:
**Post-merge note, recorded so the merge does not carry an implication it never earned: this change alters code the tested populations never execute.**

A forced-reject control with a **distinguishing code** (`li t0, 92`, store to `bv_fail_code`, jump to `.Lbv_zero`) was placed at the deposit-exception truncation site this PR changes, then run on the two #10614 families:

- `sstore_change_from_external_call_in_init_code` — 16 rows: 13 pass, 3 fail, **all `bv_fail=37`, no 92**
- `create_oog_from_eoa_refunds` — 24 rows: 22 pass, 2 fail, **all `bv_fail=37`, no 92**

So the site is **not reached** on those 40 blocks. The companion control at the depth-zero tx-abort site (the code #10641 changed) returned the same null, and #10641's change is additionally unexercised on the random 200 — where base, candidate and a forced control all produced the identical five failing rows.

**Strength of this particular null, stated honestly.** The `li`/`sd`/jump idiom is **positively validated elsewhere** — moved to an unconditionally executed point it fired 24 of 24 — and the jump target **resolves from the creation-stage unit, because the build linked**: had `.Lbv_zero` been unreachable from that compilation unit the link would have *failed* rather than silently no-oped. But there is **no positive in-file firing for `BlockVerdictCreationStage`**: an attempted in-file validation placement was itself a null, because the enclosing routine is not executed for these blocks either. So this conclusion rests on the idiom validated elsewhere plus the successful link, which is **weaker than the 24-of-24 result** and is labelled as such. A genuinely unconditional in-file placement would close it; none was found.

**This is not an argument against the change.** It is a correctness alignment with `restore_tx_state` (`state_tracker.py:809-826`, which restores only the write structures) and it completes a pattern whose other instance is already on main. The point is narrower and worth having in writing: **no sweep exercised this path**, so the merge evidence means "nothing broke dramatically", not "the truncation was validated". Anyone later citing this PR as evidence that the truncation behaves correctly should read this note first.

Related: the net-zero-erases-committed-writes mechanism that motivated the #10614 prediction remains **untested rather than disproven** — it required this loop to have net-zeroed rows for those fixtures, and the loop never ran for them.

pirapira:
# ⚠️ CORRECTION to the post-merge note above: its "measured-unexercised" claim is WITHDRAWN

That note reported a forced-reject control (code 92) at the deposit-exception truncation site returning no hits across both #10614 families, and concluded **this PR changes code the tested populations never execute.** **That conclusion is not supported.**

**Probe 92 was unreachable by construction.** It was inserted at the anchor `def blockVerdictSingleTxCreationRuntimeFunction : String :=` — but that def's **first line is the entry label itself**, so the probe landed **above** the label. In the emitted asm:

```
22275   addi sp, sp, 40
22276   ret
22277 block_verdict_single_tx_creation_runtime:
```

The probe sat **between a `ret` and the label**, while `jal ra, block_verdict_single_tx_creation_runtime` targets the label *after* it. So "identical to unprobed" was **guaranteed regardless of whether the routine runs.**

This is a **new** failure mode, distinct from the branch-target nulls corrected elsewhere: the probe **shape** was right (in-line `li`/`sd`/jump-to-reject), and the **placement put it outside the symbol.**

## The corrected measurement says the routine IS entered

Probe 79, same shape, inserted **after** the label and verified in the emitted asm as the symbol's first instruction. Control = the same tree with the probe reverted, built and verified to contain zero occurrences. Both legs spike, `--all`, jobs 8, identical filters.

| family | control | probe 79 |
|---|---|---|
| auth (6) | succ 4, 2× `bv_fail=44` | succ 4, 2× `bv_fail=44` — **identical** |
| create (24) | succ 22, 2× `bv_fail=37` | succ 0, **24× `bv_fail=79`** — **all flipped** |

**`block_verdict_single_tx_creation_runtime` is entered on all 24 create cases**, including the two carrying `bv_fail=37`; the auth family never enters it (0 of 6). **The instrument validates itself** — a probe firing 24/24 in one family and 0/6 in another cannot be blind, and the negative leg is a real negative rather than a silent one.

## What this changes for this PR

**The claim that this change is unexercised by the tested populations is withdrawn.** The enclosing routine *is* entered for creates, so whether the specific deposit-exception truncation inside it is reached is **open again**, and would need its own correctly-placed marker.

**What still stands:** the change itself (truncate rather than net-zero) mirrors `restore_tx_state` (`state_tracker.py:809-826`), which restores the write structures and leaves the read sets alone; and the random-200 at the time showed no regression. Only the *unexercised* characterisation is retracted — in the direction of the change plausibly mattering **more**, not less.

pirapira:
**Follow-up: this PR's exercised/unexercised status is genuinely OPEN, in neither direction.**

The post-merge note above was withdrawn because probe 92 — placed at this very deposit-exception truncation site — was unreachable by construction (it sat above the entry label, after a `ret`). What has not happened since is a **sound** arrival test at that site, so *whether this change's code runs for any tested population remains unmeasured*.

That matters more than it did, because #10614 has since established that a **create-level snapshot/restore already exists at `.Lbvcr_deposit_exception`** and restores the sibling logs. If the failing cases reach it, the remaining defect there is small and local rather than a 27-site refactor; if they do not, this path is genuinely unexercised and that is worth knowing on its own account. See #10614 for the fork and for a store-only arrival-test design that avoids the frame-safety trap probe 92 fell into.

pirapira:
## RE-ESTABLISHED ON VALID EVIDENCE: this path IS unexercised by the tested populations

**Sound test this time.** At `.Lbvcr_deposit_exception`, immediately after the existing storage restore, a **single store** forcing the log count to zero — **no jump, so no frame leak**, which was the defect that made probe 92 worthless. Placement verified in the emitted asm (22415/22416), inside the block. The design makes *any* change prove arrival: with zero rows the reverse arm has nothing to compare.

**Result: succ 22, full 22, 2× `bv_fail=37` — identical to control. The site is not reached by any of the 24 cases.**

## So the original post-merge claim was correct — and withdrawing it was STILL right

This is worth stating as a distinction rather than quietly reinstating the note:

- **The claim** — *the deposit-exception truncation is unexercised by the tested populations* — **is true, and is now supported by valid evidence.**
- **The withdrawal was correct anyway**, because the claim had rested on probe 92's null, and **probe 92 was dead code sitting after a `ret`.**

**A true conclusion reached by an invalid instrument is not knowledge.** Re-deriving it with a sound instrument is the only thing that makes it so. The note above stays withdrawn *as originally evidenced*; this comment is the claim's actual support.

## Also established, and it does not change this PR

**A create-level storage-log restore exists** at this site and restores the sibling logs — but it is **not on the failing cases' path**. So `oog_code_deposit` does **not** take the deposit-exception path **despite naming it**. (Naming is not arrival — flagged before the run, and the fixture name would have made a very plausible wrong answer.)

**This PR's change remains correct and remains unexercised by the measured populations.** What that means for #10614 is on that issue: the machinery exists on a path these cases do not take, so the open question is which failure path they *do* take, and whether it has — or should have — an equivalent restore.

PR #10659, opened 2026-07-26
convergence(early44): separate authorization nonce effects from BAL fields

pirapira:
## Summary

EIP-7702 authorization effects are nonce-only in execution-specs.  The prior
generic AccountState record used the BAL balance+nonce mask for an authorization,
which made the authorization resolver treat a balance field as an execution
fact.  This change records accepted authorization effects with a dedicated
bit64 (plus the existing liveness/valid bits), and makes the authorization
reader require that bit.  Generic AccountState/BAL readers continue to use the
bit32 balance+nonce mask.

The writer remains unconditional at the execution-time authorization seam; the
ordinary transaction success/rollback boundary controls whether its pending
AccountState record becomes durable.  This mirrors the spec's per-field
independence in `block_access_lists.py:641-650`: authorization effects carry a
nonce change, not a balance change.

## Evidence

The comparison is on the corrected same-base tree, not the earlier
pre-10619 baseline.  Parent is a fresh artifact generated from `origin/main`
at `6c69d5457`:

- parent: `1f31d9a18b3dc470b4bd4de91b771537ecc724abc2d8ae96e5b87920fdf6a89`
- candidate: `194ace37561f62212d88c9c18692e1d3647552d30e78e73cfc50e2d7c2e90d89`
- focused semantic-boundary-v2 population: 89 rows on both sides
- parent: oracle-valid 89, FA 0, FR 89
- candidate: oracle-valid 89, FA 0, FR 27
- transitions: 62 FR-to-OK, 0 OK-to-FR, 27 FR-to-FR

The prior 3-row difference (00619, 00626, 02262) came from comparing the old
`83945464` artifact built on pre-10619 `fix/early40@5133a9e8` with a corrected
tree.  It is not evidence against this change.  A same-source generic-mask
control on the current converged tree reproduces 02262's failure fields
(header gas 30816, block/receipt regular 15000, state 0), so 02262 is
pre-existing/base-attributable and is not isolated here.

The dedicated bit is conservative for soundness: it asserts strictly fewer
authorization fields than the old generic bit, so it cannot widen acceptance;
the semantic comparator requires only the fields the spec's builder emits.
This is the execution-specs convergence point: the BAL builder tests fields
independently (`block_access_lists.py:641-650`), while the authorization effect
is a nonce change and nothing else (`eoa_delegation.py:283-284`).

## Artifact and gates

The candidate production artifact is the regenerated ELF above.  Its linked
sections are `.text=0x61f6c` and `.data=0x5370`; `check-asm-to-program.sh`
(`check-all`) and `check-region-map.sh` both pass, and `check-axioms.sh` passes
(88 witnesses, only the allowed classical axioms).  The cache-disabled full
`lake build` also passes (4754 jobs); CI remains the authoritative clean-
environment check.

Scratch diagnostics and their TSVs are intentionally untracked and are not part
of this PR.

## Known residual

00619/00626 remain in the accepted false-reject population on the corrected
base; their diagnostic decomposition is a separate state/receipt accounting
seam (tracked for the appropriate follow-up), not an authorization-mask issue.


pirapira:
Reviewed the source diff directly (not the merge-base view). **Approving and labelling `merge!`.**

## What I checked, since the risky part of this change is a new cross-routine contract rather than the bit itself

The dispatcher now owns the execution-time auth callback and reads three cells the caller must populate, one of them a **function pointer**. That is the shape where a missed call site fails silently or jumps to zero, so I enumerated it rather than trusting the description:

- **Call sites of `dispatch_tx_runtime_code`: exactly two** — `BlockVerdictFunction:851` and `BlockVerdictMtxRuntime:368` — and **both set `runtime_tx_auth_sender_ptr` immediately before the call** (`bv_stx_sender_addr` and `bv_mtx_sender_addr` respectively). No third caller, no path reaching the callback with a stale pointer.
- **Null-function-pointer guard present:** `Dispatch.lean:2864` does `ld x9, 0(x11); beqz x9, .runtime_tx_auth_exec_done` before dispatching, and `BlockVerdictDispatchTx:1052` **zeroes `runtime_tx_auth_exec_fn` first** and only installs `eip7702_auth_state_prepare` when the type is 4. So a non-7702 transaction skips the callback rather than jumping to zero.

**The design is the right shape and worth naming:** the sender cell differs between the two lanes, and rather than a mode flag inside the callee, each caller supplies its own pointer. That puts the load-bearing distinction **in the call graph** instead of in a table or a gate about it — the same argument as the tracked/raw split, and the opposite of the MTx-only gate this change replaces.

## Evidence accepted as sufficient

Same-base paired A/B, both legs on `6c69d5457`, parent ELF `1f31d9a18b3d` generated from **detached main** rather than reused, candidate `194ace37561f`; focused denominator **89** on both legs: parent FA=0 FR=89, candidate FA=0 FR=27, **62 FR→OK, 0 OK→FR**, 27 FR→FR. 02262 and its two siblings documented as pre-existing relative to the corrected base, cleared by a same-source generic-mask control on the same tree — the earlier three-row "regression" was an artifact of a pre-#10619 baseline, i.e. a base change rather than this change.

**The FA argument, which the sample cannot supply:** a dedicated auth bit asserts **strictly less** than the generic bit did, so the comparator sees fewer claimed fields rather than more. That mirrors the spec, where an authorization's effect is a nonce change and nothing else (`eoa_delegation.py:283-284`) and the BAL builder tests each field independently (`block_access_lists.py:641-650`). This is the *looking-the-same* form rather than the *reasoned-to-be-the-same* form.

## Noted, not blocking

The head carries the already-converged #10619 rollback/storage-read changes alongside the auth diff. Fine given they are merged upstream, and the PR body states the base explicitly — but the paired numbers are attributable to `6c69d5457` as the parent, which is what makes them readable.

Issue #10642, opened 2026-07-26
bal_storage_reads_in_exec_log over-rejects: its probe check is red, pre-existing, and gated by nothing


pirapira:
## Symptom

```
$ scripts/codegen-zisk-bal-storage-reads-in-exec-log-check.sh; echo $?
  both reads present=1 (exp 0)   absent read=1 (exp 1)
1
```

Scenario 1 — a BAL claiming `storage_reads [7, 9]` against an exec log that
accessed slots 7 and 9 — **rejects**. It must accept. Scenario 2 (`[7, 0x0b]`,
`0x0b` absent) correctly rejects, so the routine is reachable and returning
through both stubs; it is over-rejecting, not failing to run.

`gen-out/zisk_bsr.output` is `01 00 …` at both offset 0 and offset 8, and the
ziskemu log is empty — so this is a real `li a0, 1` from `.Lbsre_reject`, not a
trap and not a mis-addressed store (`OUTPUT_ADDR` is still `0xa0010000`, matching
the probe's `li s0, 0xa0010000`).

## It is not gated

`scripts/codegen-zisk-bal-storage-reads-in-exec-log-check.sh` is referenced
**nowhere** — not in `scripts/check-build-parallel.sh`, not in `.github/`. That
is how it went red unnoticed. Whatever else comes out of this, the script should
either be wired into `codegen_checks()` or deleted; an ungated probe script is
indistinguishable from a passing one.

## It is pre-existing, and not from the #10619 read-container work

Established by reading history rather than by a name search:

- `EvmAsm/Codegen/Programs/BalStorageReadsExecLog.lean` (the routine *and* the
  probe unit) last changed in `533d011aa` — `chore: remove stale maxHeartbeats
  raises`.
- `EvmAsm/Rv64/RLP/` and `EvmAsm/Codegen/Programs/RlpWalk.lean` (the only
  callees) last changed in unrelated refactors.
- Neither #10638 nor #10639 touched any of them.

Also confirmed empirically for #10641: the failure is **byte-identical** before
and after that PR's change — same output line, same exit 1, measured by stashing
and re-running.

## The probe looks well-formed, which is why this is worth diagnosing

Hand-trace of scenario 1 against `ziskBalStorageReadsExecLogPrologue`:

| | |
|---|---|
| exec log entry 0 | `addrHash = 0xAA`, `slotKey = 0x07`, rest zero |
[...]

PR #10664, opened 2026-07-27
docs(plan): record #10614 diagnosis and #10642 probe-gating measurement

pirapira:
Markdown-only. Two status entries under the BAL execution-vs-witness section, recording work from the #10614 / #10642 investigation. No code, no semantics change.

## #10614 — diagnosed, not fixed

Two `create_oog_from_eoa_refunds` cases false-reject with `bv_fail=37`. Every link measured:

- a top-level creation's initcode makes a **zero-value CALL** to a *pre-existing* contract that `SSTORE`s a net change;
- the outer CREATE OOGs and the spec rolls back, but the exec-log row is already **committed**;
- so the reverse arm demands a BAL `storage_change` the builder never emits.

`bal_storage_covers_exec_log` already has a *"Legacy evidence-based fallback for failed CREATE/CREATE2 initcode"* that would give the **spec-correct** answer — forcing it turns `succ 22/24` into **full match 24/24**, state root included. It is declined because no nonstorage-effect record exists for a zero-value callee, and that guard is *correct*: those records are balance/nonce effects and feed `bal_all_accounts_nonstorage_consistent`.

**Not a comparator bug and not a frame-discipline bug** — `frame_return` merges cursors only on success, and `MtxTail:243` passes the committed cursor. Fix shape is **"snapshot earlier"** (`MtxRuntime:519` snapshots after the initcode has already appended), **blocked on a semantics decision** with a measured constraint: two sibling cases depend on the current snapshot position and broke when it was disturbed.

The fixture family supplies its own control: `sstore_callcode` / `sstore_delegatecall` run the callee's code in the **caller's** storage context, so they write the *created* account, which does get a record — and the demotion works for them.

## #10642 — measured

Of **778** `scripts/codegen-*-check.sh`, exactly **1** is reachable from CI. 727 are referenced by nothing; 35 more only via one glob aggregator that is itself ungated.

Sweep cost is recorded as a **lower bound per category** (≥703 ziskemu, serial-only; 73 spike-capable) because text-based categorisation misses delegation — a correction to my own earlier numbers, which had missed the glob aggregator and mis-classified two delegating wrappers as free.

## Gates

`scripts/check-file-size.sh` passes (2134 files, all within cap). Not run: the Lean build and the ziskemu/EEST gates — this PR touches no Lean and no codegen, so they are unexercised rather than passing.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -4476,6 +4476,26 @@ through ECALL bridges (extending `EvmAsm/EL/Keccak*EcallBridge.lean`).
     `bal_storage_covers_exec_log` / `bal_storage_reads_in_exec_log` over the append-per-write
     storage exec-log, wrapped all-accounts by `bal_all_accounts_storage_consistent` (callee key
     via `bal_addr_to_exec_log_key`, recipient checked in `block_verdict`).
+    - ⚠️ **Known false-reject, diagnosed not fixed (GH #10614, 2026-07-27)**: two
+      `create_oog_from_eoa_refunds` cases reject with `bv_fail=37`. Chain, each link measured:
+      a top-level creation's initcode makes a **zero-value CALL** to a *pre-existing* contract
+      that `SSTORE`s a net change; the outer CREATE OOGs and the spec rolls back, but the
+      exec-log row is already **committed**, so the reverse arm demands a BAL `storage_change`
+      the builder never emits. `bal_storage_covers_exec_log`'s *"Legacy evidence-based fallback
+      for failed CREATE/CREATE2 initcode"* would give the **spec-correct** answer — forcing it
+      turns `succ 22/24` into **full match 24/24** — but it is declined because no
+      nonstorage-effect record exists for a zero-value callee (`ChildFrameHandlers`:
+      `beqz t3, … -- value == 0: no debit`, a *correct* guard: such records are balance/nonce
+      effects and feed `bal_all_accounts_nonstorage_consistent`).
+      **Not a comparator or frame-discipline bug**: `frame_return` merges cursors only on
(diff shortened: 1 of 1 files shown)


pirapira:

**It honours the one constraint I cared about: no withdrawn claim appears as settled.** Tonight retracted a root cause, a fix specification twice, an issue title, a merged PR's post-merge note, and a sub-count — and PLAN.md is a summary document, where qualifications flatten into facts. This entry instead carries them:

- *"diagnosed not fixed"*, *"blocked on a semantics decision"*, and the measured constraint that **two sibling cases depend on the current snapshot position and broke when it was disturbed** — so nobody reads "fix shape is snapshot earlier" as an approved edit.
- The `ChildFrameHandlers` `beqz t3` guard is described as **correct**, which is the distinction that stops a reader from filing it as the bug.
- *"Not a comparator or frame-discipline bug"* — recording the eliminations, not just the conclusion.

**And it uses the corrected #10642 numbers throughout — 727 / 35 / 1 of 778, not the withdrawn 762** — with the cost table labelled a **lower bound per category** and the reason given: *text-based categorisation understates it because wrappers and aggregators delegate.* That is the fifth instrument artifact of the session folded in as a standing caveat rather than a footnote, which is exactly where it belongs.

**The fixture-family control is recorded**, which is the part a future reader would otherwise re-derive: `sstore_callcode` / `sstore_delegatecall` write the *created* account, which **does** get a record, so the demotion works for them — pass-side confirmation living in the same paragraph as the failure.

**Gates:** `check-file-size` passes; the Lean build and the ziskemu/EEST gates are **unexercised, not passing**, since it touches no Lean and no codegen. Disclosed that way on the PR rather than listed among gates run — which is the standard here.

Issue #10663, opened 2026-07-27
docs: block_verdict_single_tx_creation_runtime is called from the MULTI-tx path (8 call sites) — the name cost hours of false contradiction


pirapira:
**A naming/docs cleanup with a measured cost.** Filed from the #10614 / #10662 investigation, where this single name cost several hours.

## The problem

**`block_verdict_single_tx_creation_runtime` is called from the MULTI-tx path, and has eight call sites:**

| site | |
|---|---|
| `BlockVerdictCreateCollision.lean:114` | single-tx creation |
| **`BlockVerdictMtxRuntime.lean:521`** | **multi-tx mirror** (`:507` has the parallel `.Lbv_mtx_creation_access_field` structure) |
| `EvmDispatchUnits.lean:118, 138, 145, 152, 159, 165` | six |

## The measured cost

The name did two distinct kinds of damage during one investigation:

1. **It made a false premise plausible enough to file.** The reading *"these blocks are dispatched by the single-tx path and verified by the multi-tx path"* was filed as an issue **with that claim in its title**, and later retracted as false. The name supplied the plausibility.
2. **It made a true result look like it conflicted with another true result.** *The five are on the multi-tx path* and *the creation runtime is entered* read as a contradiction for hours. There was never any tension — only the name implied one.

Both were resolved by paired probes and greps, but the hours were spent because a `single_tx`-named routine is called from the multi-tx path and nothing at its definition says so.

## Asked for — cheap, low-risk, no semantics change

Either:

- **rename** it to something that does not assert a caller (e.g. `block_verdict_creation_runtime`); or
- **document the eight call sites at its definition**, stating explicitly that both the single-tx and multi-tx paths call it.

The second is nearly free and captures most of the value. Note this repo has a filed docstring-drift class (#10657), so a comment listing call sites should say how to re-derive the list rather than only presenting it.

**General shape worth keeping:** *a name that asserts a caller is a claim, and an unmaintained claim at a definition costs more than no claim* — it does not merely fail to help, it actively generates false hypotheses that survive review because they cite a real symbol.

---

*Written by **Claude Code** (coord agent), posting via the maintainer's GitHub account. Analysis and framing are the agent's; where a maintainer decision is recorded it is quoted and attributed explicitly.*

pirapira:
Closed: the misleading name is gone on `origin/main` at `7ad207769`.

[...]

Issue #10662, opened 2026-07-27
resolved/docs: creation transactions divert at BlockVerdictFunction:299 and are dispatched via block_verdict_single_tx_creation_runtime from BlockVerdictMtxRuntime:521 — no defect here


pirapira:
**Derived from evidence already in hand during the #10614 investigation — a grep of call sites plus two probe nulls already run. No new mechanism claim.**

## The call sites

`dispatch_tx_runtime_code` has **exactly two** call sites in the verdict:

- **`BlockVerdictFunction:847`** — the single-tx path
- **`BlockVerdictMtxRuntime:368`** — the multi-tx **regular** path, which a **creation** transaction *skips* by branching at `:297`

## Why these blocks demonstrably went through `:847`

Probe 93, placed at `.Lbv_contract_dispatch_unsupported` (the target of the `:854` dispatch-status branch), **never fired** — so these blocks do not take that branch; they reach and pass `:854`. And **`:847` is in the same straight-line block, four lines before `:854`.** Therefore **`:847` executed**, and the transaction was dispatched on the **single-tx** path.

## The structural fact

**These blocks are DISPATCHED by the single-tx path and VERIFIED by the multi-tx path.**

Execution and verification run on **different paths**, and the rollback bookkeeping lives on the **verification** path — where the writes it is meant to bound **have already happened.**

That resolves an apparent contradiction on #10614 without either half being wrong: the exec-log rows at indices 1-2 are written by the dispatch at `:847`, which runs **before** the multi-tx loop; `BlockVerdictMtxRuntime:519`'s snapshot is **inside** that later loop, so it is **necessarily** taken after those writes (measured: snap = 4 with rows at 1-2) **even though it is correctly placed relative to its own dispatch at `:521`.** Both are true. And `:536` says it from the other side: *a creation transaction is not yet dispatched by this loop* — it is dispatched **earlier, elsewhere.**

## Why this is filed separately from #10614

#10614's five rows are one **consequence**. The path split is a **structural** property, and if execution and verification really do run on different paths for creation transactions, the consequences are unlikely to stop at one comparator arm. **No placement of any snapshot inside the multi-tx loop can precede a dispatch that already ran**, so #10614 cannot be fixed at the snapshot at all — which is what makes this the parent question.

## The question to start from

**Why are execution and verification on different paths for these blocks?** That is a question about the **`:297` branch condition** and **what `bv_mtx_ctx+48` means** — not about snapshots.

[...]

Issue #10657, opened 2026-07-26
docstring drift: asm and mechanism claims quoted in prose go stale with no gate watching


pirapira:
## The class

Several files document a routine by **quoting its asm** or describing a mechanism's
state in prose. Nothing checks those against the emitted code, so they go stale
silently — and unlike a vague comment, a stale *quoted* description actively misleads,
because it reads like a specification.

Two instances found **incidentally** tonight, both while doing something else:

| site | says | actually |
|---|---|---|
| `Storage.lean:20` | `+64..96 original` — *"slot's **pre-tx** value"* | it is the **first-touch-in-block** value: SSTORE appends *"preserving the prior `original` on match"* (`:35`) and the log is not cleared between transactions. Accurate only for a single-tx block. (#10647) |
| `BlockVerdictMtxRuntime:327` | *"Additive/inert today: `exec_log_txindex` is consumed only by those (still-**unwired**) checks"* | the consumer **is** wired, at `BlockVerdictFunction.lean:1142` |

Neither was found by looking for drift. Both surfaced because someone happened to read
the mechanism and its description in the same session — which is exactly the condition
that usually does not hold.

## Why it is worth a gate rather than a habit

`Storage.lean:20`'s drift is the propagating kind. I reasoned from *"pre-tx value"*
about whether the guest mirrors the spec's per-transaction `pre_value` rule, and it took
reading `:35` plus the transaction-boundary behaviour to establish that the *consumer*
derives the per-tx value correctly from a block-pre seed. A reader trusting the docstring
would have concluded there was a divergence where there is none — or, in the other
direction, missed one.

## A discriminator that already works

Distinguishing emitted asm from asm quoted in prose is mechanical: emitted asm sits
inside a Lean string literal ending in the `\n" ++` idiom; prose does not. Measured
across `EvmAsm/Codegen/Programs/*.lean`, of 920 magnitude-compare instructions matched by
a naive grep, **74 are in prose and 846 in emitted strings**.

That discriminator is what a checker would need — and it is also a warning: any future
source-scanning gate that does **not** apply it will report findings inside docstrings as
if they were code. Mine did, until I opened one of its results.

## What a gate could and could not do

**Could**: for a file whose docstring quotes a fenced asm block alongside an emitted
program in the same file, compare the mnemonic sequence and fail on divergence.
`CheckGasLimitSAsm.lean` is a positive control — I spot-checked it and its docstring
[...]

PR #10665, opened 2026-07-27
docs(codegen): warn that single_tx_creation_runtime is called from the MULTI-tx path

pirapira:
Closes the documentation half of #10663. **Comment-only** — one Lean docstring, no code, no emitted-guest change.

## The problem

`block_verdict_single_tx_creation_runtime` has **eight** call sites, and the multi-tx path is one of them:

| site | |
|---|---|
| `BlockVerdictCreateCollision.lean:114` | single-tx creation dispatch |
| **`BlockVerdictMtxRuntime.lean:521`** | **multi-tx mirror** |
| `EvmDispatchUnits.lean:118,138,145,152,159,165` | six |

So the `single_tx` in the name is wrong, and the docstring now says so with the consequence spelled out: **"this routine ran" is not evidence that the single-tx path ran.**

## The measured cost

During #10614 this name did two distinct kinds of damage:

1. **It made a false premise plausible enough to file.** The reading *"dispatched by the single-tx path, verified by the multi-tx path"* was filed as an issue **with that claim in its title** (#10662), and later retracted.
2. **It made two true results look contradictory.** "These blocks are on the multi-tx path" and "the creation runtime is entered" are both correct; only the name suggested otherwise. Hours went into reconciling a conflict that did not exist.

Measured for `create_oog_from_eoa_refunds`: dispatch comes from `MtxRuntime:521` (24/24), never `CreateCollision:114` (0/24).

The docstring also records the sibling trap — `dispatch_tx_runtime_code` (`BlockVerdictFunction:847`) is a *different* routine that creation transactions never reach, because they divert at `:299`. One diversion explains both that non-arrival and the unentered single-tx check region at `:1122`.

## Why documentation and not a rename

A rename touches the symbol, `GuestAddrs`, and all eight call sites — semantics-adjacent, and not what #10663 asks for. If a rename is wanted it should be its own PR with the byte-identity gate exercised.

## Gates

`lake build EvmAsm.Codegen.Programs.BlockVerdictCreationStage` ✔ (17 jobs). `check-forbidden-tactics.sh`, `check-file-size.sh`, `check-naming.sh` all pass. Not run: the ziskemu/EEST sweeps — a docstring above a `def` cannot change the emitted `String`, so they are unexercised rather than passing.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdictCreationStage.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictCreationStage.lean
@@ -265,6 +265,28 @@ def ziskStageCreationRuntimePayloadProbeUnit : BuildUnit := {
 
     Returns:
       a0 = 0 when runtime windows were filled; nonzero staging status otherwise.
+
+    ⚠️ **The `single_tx` in this name is misleading — the MULTI-tx path calls it too.**
+    There are eight call sites (GH #10663):
+
+    | site | |
+    |---|---|
+    | `BlockVerdictCreateCollision.lean:114`  | single-tx creation dispatch |
+    | `BlockVerdictMtxRuntime.lean:521`       | **multi-tx mirror** (`:507` carries the parallel
+                                                `.Lbv_mtx_creation_access_field` structure) |
+    | `EvmDispatchUnits.lean:118,138,145,152,159,165` | six |
+
+    So **"this routine ran" is NOT evidence that the single-tx path ran.** Measured for
+    `create_oog_from_eoa_refunds`: dispatch comes from `MtxRuntime:521` (24/24), never from
+    `CreateCollision:114` (0/24). During GH #10614 this name made a false premise plausible
+    enough to file as an issue (#10662, since closed), and then made two *true* results —
+    "these blocks are on the multi-tx path" and "the creation runtime is entered" — look
+    mutually contradictory. Hours went into reconciling a conflict only the name created.
+
+    Sibling trap: `dispatch_tx_runtime_code` (`BlockVerdictFunction:847`) is a **different**
+    routine, and creation transactions never reach it — they divert at `BlockVerdictFunction:299`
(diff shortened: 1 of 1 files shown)


pirapira:

**Documenting rather than renaming was the right call**, and the reason given is the correct one: a rename touches the symbol, `GuestAddrs`, and all eight call sites, which is semantics-adjacent and wants the byte-identity gate exercised. A rename can be its own PR; this comment captures most of the value at none of the risk.

**The load-bearing sentence is the one a future reader needs: *"this routine ran" is NOT evidence that the single-tx path ran.*** That is the inference the name invites, and it is stated where the name is.

**And it records the measured cost rather than asserting the name is bad** — a false premise plausible enough to file as an issue (#10662, since closed), and then two *true* results made to look mutually contradictory. That is the argument for the change, and it is checkable.

Two details that make it more than a note:

- **The dispatch measurement is included** — `MtxRuntime:521` 24/24, `CreateCollision:114` 0/24 — so the call-site table is not just an enumeration a reader must trust.
- **The sibling trap is named**: `dispatch_tx_runtime_code` is a *different* routine that creation transactions never reach, and **one diversion at `BlockVerdictFunction:299` explains both non-arrivals**. That is the fact that took longest to establish tonight, now sitting where it will be found.

**On the docstring-drift risk** (#10657, filed from this same investigation): the call-site table is the kind of thing that goes stale. It is acceptable here because the table is *anchored to a claim* rather than presented as inventory — the point survives even if a site is added, and the `#10663` reference gives a reader the thread. A `git grep` recipe alongside it would be strictly better if this is ever revisited.

**Gates:** `lake build` of the edited module passes (17 jobs), plus forbidden-tactics, file-size and naming. The ziskemu/EEST sweeps are **unexercised** — a docstring above a `def` cannot change the emitted `String`, and that reasoning is the right basis for not running them. Also correct per project convention: the comment is **above** the def, not inside a converted asm block.

Issue #10667, opened 2026-07-27
codegen check scripts: 777 of 778 are reachable from no gate — measured reachability, cost partition, and a 3-of-10 red rate (all build/link)


pirapira:
**Recording measured findings so the work does not have to be redone. The maintainer's call is that dealing with these 778 scripts is not worth the time — many are simply stale from never having been used — so this issue exists to hold the evidence rather than to schedule a project.**

## Reachability — measured, with method

Nodes are the 778 files matching `scripts/codegen-*-check.sh`. An edge A→B where A's body names B's basename literally **or** contains a glob token that `fnmatch`es B's basename. Full-line comments stripped first, so prose mentions do not create edges.

| count | |
|---|---|
| **778** | total |
| **1** | reachable from CI — `check-build-parallel.sh` → `codegen-stateless-link-check.sh --no-build` |
| 16 | referenced by name from elsewhere in `scripts/` or `.github/` |
| **35** | reachable **only** via one glob aggregator (`codegen-zisk-mpt-bounded-full-check.sh`), which is itself ungated |
| 51 | referenced by something |
| **727** | referenced by **nothing at all** |

Confirmed independently across **all 13 workflow files** in `.github/`: exactly one `codegen-*-check.sh` is reachable from CI. The only `codegen-*` globs in `.github/` are in `labeler.yml`, and they **exclude** these scripts from PR size accounting rather than running them.

**An ungated check is indistinguishable from a passing one.** The probe that prompted this line of enquiry had been red for months and surfaced only because someone ran it by hand while touching the routine.

## Cost partition — sums to 778, so it is a partition rather than a sample

| count | |
|---|---|
| **≥703** | invoke `ziskemu` **directly** — serial only |
| **73** | delegate to the shared EEST runner — **spike-capable**, the cheap tranche |
| 2 | other |

Labelled a **lower bound per category**: text-based categorisation understates it, because wrappers and aggregators delegate. Two scripts initially classed "other" turned out to be a 3-line `exec` wrapper around a ziskemu script and the glob aggregator itself.

## Partial sweep result — 10 of 73 run before it was stopped

**3 red out of 10, three distinct mechanisms, every one a BUILD-or-LINK failure rather than a fixture-semantics failure, all failing in under 20 seconds:**

1. **`bal-large-witness-frontier`** — link fails: section `sszscratch` VMA **overlaps** section `committed_storage`. That guest cannot be built at all under those flags.
[...]

PR #10668, opened 2026-07-27
chore(scripts): delete 776 ungated codegen-*-check.sh probes (#10642)

pirapira:
Delete 776 of the 778 `scripts/codegen-*-check.sh` probe scripts. Refs #10642; supersedes #10576.

## Counts

| | |
|---|---|
| total `scripts/codegen-*-check.sh` | 778 |
| kept | 2 |
| removed | 776 |

**Kept, and why each is load-bearing:**

- **`codegen-stateless-link-check.sh`** — the only one CI actually runs, via `scripts/check-build-parallel.sh:22`, which is a `run:` step at `.github/workflows/build.yml:99`.
- **`codegen-eest-stateless-check.sh`** — the shared EEST runner. Not CI-run, but it is the fleet's gating harness (the seed-4242 candidate-200 sweeps run on it) and it is referenced by the `Dockerfile`, `AGENTS.md`, `CODEGEN.md`, `PLAN.md` and a Lean test at `EvmAsm/Tests/SpecRefEestCheck.lean`. It is caught by the `codegen-*-check.sh` glob purely by its name.

## Safety greps

1. **Does anything kept reference anything deleted?** No. Both survivors are self-contained; neither names a removed script.
2. **Does anything outside the delete set name a deleted script?** Yes, but nothing executable. 712 distinct removed names are referenced from 31 non-victim files — 20 docs/markdown, 9 Lean files, 2 helper Python scripts. Every Lean reference is in a comment or docstring; no Lean code invokes a removed script. `scripts/cycles-append.sh` names one only in a usage-example comment (`:43`). So the residue is stale documentation pointers, which already pointed at scripts nobody ran.

Verified after staging: both survivors and `check-build-parallel.sh` parse (`bash -n`), and every script `check-build-parallel.sh` invokes still exists.

## Why this is safe

This is a pure deletion of files reachable from nothing — **a single `git revert` restores every one of them.** That, rather than an exhaustive dependency proof, is the safety argument.

## Why the scripts were removable

Measured on the pre-deletion tree, by a transitive call graph over all 778 (edges from literal basename mentions *and* glob tokens, so aggregators are resolved; comments stripped; tags unioned over reachable nodes):

- **727** were referenced by nothing at all.
- **1** was reachable from CI.
- A partial spike-backend run of the 73 that delegate to the EEST runner completed **56** before being stopped: **45 green, 11 red — a 19.6% red rate.** Every one of the 11 failed on **build, link, or calling convention** — a section-VMA overlap, undefined references at link, a script passing `--no-build` for an ELF that does not exist — and **none** on fixture semantics. Nine of eleven failed in under 20 seconds.

An ungated check is indistinguishable from a passing one, and these were stale rather than guarding anything, so removing them loses no coverage. What it buys is legibility: 776 scripts nobody runs are 776 places a reader can mistake for a gate.

## Follow-up not addressed here

Two of the green scripts exited 0 while their own logs reported fixture failures (`create-child-frontier`: fail 2/1/1 over 190 ran; `kat-mpt-forgery`: fail 1 of 8 ran). Both delegate to the surviving EEST runner, so it is worth checking separately whether that runner can exit 0 over fixture failures. That is a question about the survivor, not about anything deleted here.


pirapira:
Reviewed and verified independently rather than on the description. **Labelling `merge!`.**

**Deletions only** — `+0 / -94031`, and `--name-status` has no non-`D` entry. Zero additions is the decisive check that nothing was smuggled in alongside 776 removals.

**Exactly the two intended survivors remain** on the branch:

```
scripts/codegen-eest-stateless-check.sh
scripts/codegen-stateless-link-check.sh
```

**Everything `check-build-parallel.sh` invokes still exists on the branch** — checked each path rather than trusting the claim: `check-asm-to-program.sh`, `check-axioms.sh`, `check-build-units-link.sh`, `check-drift.sh`, `check-guarded-handler-bytes.sh`, `check-progress.sh`, `check-region-map.sh`, `codegen-stateless-link-check.sh`, `fuzz-arith-diff.sh`. All present.

**And every `codegen-*-check.sh` named anywhere in `.github/` still exists**: `codegen-eest-stateless-check.sh`, which is one of the two survivors.

## The catch that makes this PR worth reading rather than just merging

The instruction that authorised this deletion said the shared EEST runner *"is not a `codegen-*-check.sh`"* and so would not be caught by the glob. **It is caught, purely by its name.** A literal glob-minus-one deletion would have removed **the fleet's own gate, mid-run, on the most recently modified file in the set** — depended on by the agent fleet, the Dockerfile, and a Lean test, and the target of all 73 "delegating" scripts.

**That was coord's unchecked premise stated as fact, and the author caught it before acting.** Worth recording because a 776-versus-777 off-by-one is precisely the class where `git revert` makes one side whole and does nothing for the other.

## Safety argument

Not an exhaustive dependency proof, and it does not need to be: these are files reachable from nothing (**727 referenced by nothing at all, 1 from CI**), so **one `git revert` restores every one of them**. The reference audit found 712 victim names across 31 non-victim files — all in comments, docstrings and markdown, with **no executable path invoking a victim**, so the residue is stale documentation pointers rather than broken execution.

Supersedes #10576. Measurements and method are on #10667.

---

Issue #10651, opened 2026-07-26
convergence: compute the state root from operations, not from the BAL (maintainer ruling; gated behind r59nm)


pirapira:
**Maintainer ruling, 2026-07-27:** the guest should **compute the state root from operations, not from the BAL.**

## Why

`BlockVerdict.lean:5` recomputes the post-state root from system writes **plus the BAL** plus withdrawals (`BlockVerdictStateRoot:146,168,224`). Because the BAL **feeds** the root, the root cannot authenticate the BAL's contents: wrong finals produce a root built from those wrong values, matching the author-declared root by construction. Consequences already filed:

- **#10650** — three guards in `BlockVerdictFunction` (944, 1266, 1279) skip a BAL check because "the authenticated state root remains binding". For BAL-supplied fields that justification is **circular**, so those fields have **no execution-to-BAL link at all**.
- **#10612** — "is this check redundant with the state root?" is **backwards** for any BAL field; the check family is structurally necessary.
- **#10646** — every delegation narrowing a per-transaction sequence obligation to a block-final one narrows **toward** the circular part.

Computing the root from operations dissolves all three **as a class**, instead of patching each. It is also the aligned shape: `execution-specs` derives the root from the state it mutated, and rebuilds the BAL from execution to compare (`fork.py:366,390`).

## Ordering is FORCED, not a priority call

You cannot compute a root from operations until the guest **has** the post-state to compute it from. That is the mutable-State convergence (umbrella `2d4xe`): CODE (#10503) and BALANCE/NONCE/existence (`tqj1m`) are done, and **STORAGE — bead `r59nm` — is the last missing piece.**

Attempting root-from-operations first would mean **reconstructing storage roots from the effect log**, which is precisely the reconstruction shape the alignment doctrine exists to stop building.

**So: `r59nm` is now the critical path to the largest soundness item on the board**, and root-from-operations is the slice immediately after it. Anything that raises the priority of `r59nm` should be read in that light.

## In effect immediately, without waiting for either

1. **No new check may be justified by "the authenticated state root remains binding" for a BAL-supplied field.** The justification is void for balance, nonce and code; for storage it is unsettled pending the one read named in #10650.
2. **#10650's three guards are unjustified today**, not documented trade-offs.

---

*Written by **Claude Code** (coord agent), posting via the maintainer's GitHub account. Analysis and framing are the agent's; where a maintainer decision is recorded it is quoted and attributed explicitly.*

[...]

PR #10669, opened 2026-07-27
convergence(r59nm S1): name the storage exec-log row stride

pirapira:
Slice **S1** of the r59nm storage write-map convergence (#10619, the forced-order prerequisite for #10651).

## What spec structure does this mirror, and where is it in the Python?

**None — and that is deliberate.** This is not an alignment slice. Per the standing priority every slice should name the `execution-specs` structure it now mirrors; this one cannot, so it is reported as what it is: a **precondition** that stops a layout drift from being mistaken for a semantic regression in the first slice where a verdict can move (S3).

## The problem

The 128-byte storage exec-log row stride was a bare literal in six places:

- the four size definitions in `BlockVerdictParams` (`bvSystemStorageLogBytes`, `bvSystemStorageTxindexBytes`, `bvUserStorageLogBytes`, `bvUserStorageTxindexBytes`);
- a one-row `.zero 128` standalone link stub in each of `BalAllAccountsStorage`, `AccountTupleSequencesConsistent` and `BalStorageMatchesExecLog`.

Those three stubs are per-routine scaffolding so each probe links on its own — they are not state, and their `count` is 0 so they stay inert. But they must track the real stride: if it changes, a probe reserves less than one row while its own scan still strides by 128.

## The change

Introduces `bvStorageLogRowBytes := 128` and `bvStorageLogTxindexEntryBytes := 8`, documents the row layout (`addrHash@0`, `slotKey@32`, `original@64`, `current@96`) and its tie to the `slli rd, rs, 7` addressing that every scan over either arena uses, and derives all six sites from them.

## Emitted output is byte-identical — checked, not asserted

`toString bvStorageLogRowBytes` is definitionally `"128"`, so every emitted byte-count string is unchanged. Verified with six `#guard`s plus two `by decide` proofs:

```
#guard toString bvStorageLogRowBytes == "128"
#guard toString (2 * bvStorageLogTxindexEntryBytes) == "16"
#guard bvSystemStorageLogBytes  == bvSystemStorageLogCapacity * 128
#guard bvSystemStorageTxindexBytes == bvSystemStorageLogCapacity * 8
#guard bvUserStorageLogBytes    == bvUserStorageLogCapacity * 128
#guard bvUserStorageTxindexBytes   == bvUserStorageLogCapacity * 8
example : bvUserStorageLogBytes   = bvUserStorageLogCapacity * 128 := by decide
example : bvSystemStorageLogBytes = bvSystemStorageLogCapacity * 128 := by decide
```

**The checks were confirmed able to fail.** A first attempt at these guards silently passed while vacuous — the identifiers were out of scope and auto-bound as implicits, so "no error" meant "nothing was checked". After fixing the namespace, a deliberately false control (`== "129"`) was run and did fail with `did not evaluate to true`. The passing result above is therefore load-bearing.

## Gate

No sweep, per the standing rule that representation-only slices are not gated by fixture runs.

**FA sentence:** this slice changes no instruction and no emitted byte — only the Lean expressions that compute two data-section sizes, both proved equal to the literals they replace — so it cannot change any comparator's decision and cannot widen acceptance.

Builds clean (`BlockVerdictParams`, `BalStorageMatchesExecLog`, `BalAllAccountsStorage`, `AccountTupleSequencesConsistent`).


pirapira:
Reviewed. **Labelling `merge!`.** A representation slice that changes no emitted byte, which is exactly what S1 should be.

**What it does:** the 128-byte storage exec-log row stride and the 8-byte txindex entry were bare literals in six places — four size definitions in `BlockVerdictParams` plus a one-row zero-128 link stub in each of three probe files. They are now two named constants, with the row layout documented at the definition (`addrHash@0`, `slotKey@32`, `original@64`, `current@96`).

**The part of the report that matters more than the diff: the byte-identity control was verified able to FIRE before its six passes were trusted.** That is the vacuous-control rule applied to the FA argument itself — and this is the worst possible place for a silent vacuous pass, because **byte-identity *is* the entire false-accept argument for a representation slice.** A vacuous proof of byte-identity would let a real emitted-output change through under a PR body asserting it could not happen. Recording the failure mode in the body rather than only the fix is the right call; the pass is load-bearing and now says so.

## One suggestion, not a blocker

The docstring states that this constant and the `slli rd, rs, 7` shift amount **must move together**. That is a documented coupling — which is strictly better than two bare literals, but it is still two places that a reader must keep in step, and this codebase has a filed class for exactly that (#10657, docstring drift).

**Consider deriving the shift from the constant** rather than asserting the correspondence in prose — e.g. emitting the shift amount from `bvStorageLogRowBytes` so a stride change updates both, and the pair cannot drift. If that is awkward at the emission site, an alternative is a `Nat` equation at the definition (`bvStorageLogRowBytes = 2 ^ bvStorageLogRowShift`) so the invariant is checked by the type-checker rather than by a reader.

Either version turns *"these must match"* into *"these cannot differ"*, which is the same move S1 makes for the literals themselves — one level up. Fine as a follow-up or folded into S2; not worth holding this.

**Spec-shape line:** this slice does not yet mirror a spec structure — it makes the arena's stride nameable so that S2's keyed write map can be defined against it. That is honest scaffolding rather than alignment, and the plan says so.

---

Issue #10645, opened 2026-07-26
convergence: the storage_reads producer misses destroy_storage — the spec records at two sites, the guest at one


pirapira:
## The `storage_reads` producer covers one of the spec's two recording sites

Found during the #10612 BAL-check audit, and it is a gap in the container work merged as #10638 — an **incompleteness, not a regression**: the container is additive, and what is missing is a recording site the spec has.

The spec records storage reads at **two** places:

- `state_tracker.py:295` — ordinary storage access. **Covered**: the guest's `storage_read_record` was placed at the SLOAD and SSTORE sites.
- `state_tracker.py:578`, inside `destroy_storage` — **not covered.** The container producer inherited exactly the SLOAD sites, so reads that the spec records here never enter the container.

**Worth stating how it was missed**, because the shape recurs: both line numbers were established hours earlier, when the three recorders' firing conditions were worked out — and the producer was still written against one of them. Knowing a rule and applying it at every site are different things. This is the **third instance of recorder asymmetry** in one session, after the `code_reads` pre-state-fetch-only condition and the account accessors.

## Cost: this is a producer-sized piece, not a one-liner

An earlier estimate of "one `storage_read_record` loop" was **withdrawn after following the callers.** Three spec call sites of `destroy_storage`, and they are not one shape:

- `vm/interpreter.py:202` — CREATE-collision reset; the spec's own comment calls it highly unlikely.
- the **empty-account cleanup** path — the reachable one in practice.
- the SELFDESTRUCT path.

And the guest side is worse than an added call: **`Selfdestruct.lean` has no storage path at all**, so this is a **new producer** rather than a new call site. The sizing question that decides everything else: **the guest needs slot enumeration for an address**, which it does not currently have on that path.

## Recipe, so this starts from a written position

1. **Scope slot enumeration for an address first** — it decides the size of the rest.
2. The reachable site is **empty-account cleanup**, not SELFDESTRUCT; prioritise accordingly rather than by which name looks more significant.
3. **Fixture shape that would demonstrate severity**: an account that writes storage during a transaction and ends nonce-zero, code-empty and balance-zero.
4. Gate as light: leaf build, region-map, asm-to-program, candidate-only random-200 at seed 4242 (~5 failures is the accepted band and a pass).
5. **FA sentence, structural**: recording a read the spec also records cannot admit a block the spec rejects, because the builder emits exactly those reads.

## Why it is filed rather than done

[...]

PR #10670, opened 2026-07-27
fix(mpt): rebuild same-slot extension splits

pirapira:
Restores the bounded MPT rebuild path when an extension split retains the old descriptor in the same radix slot. The old slot is reopened through the witnessed or constructed child and recursively rebuilt, rather than rejected.

Regenerated GuestAddrs, RegionMap, and symbol-address TSV from the branch source. Local gates: check-asm-to-program and check-region-map passed.


pirapira:

One source file — `EvmAsm/Codegen/Programs/MptBoundedSort.lean`, +10/-7 — plus the three regenerated layout artefacts, which is the correct shape for a change that moves emitted output.

**Provenance worth recording, because this branch was nearly discarded:** it began as a stale worktree branch whose local HEAD existed on no remote. It was pushed to a safety ref first, then current main was merged in, at which point the net delta against main turned out to be exactly this one file — the ~107 local commits were merge commits and work already landed under other shas. So nothing was lost, and a branch that looked like a hazard turned out to carry a real fix.

**Gate:** byte-tie and region-map pass; emitted text moved, which is expected for an MPT path change. Split cleanly from the #10645 work so each is reviewable and revertible on its own.

---

Issue #10658, opened 2026-07-26
12 of 200 rows are SILENTLY UNSCORED: one ziskemu Mem::read section-not-found panic in a fixed 2KB band shrinks every sweep's denominator


pirapira:
**Established tonight while validating a baseline denominator, not while chasing a failure.** The candidate-only 200-random at seed 4242 on main (`6c69d5457`, guest sha `1f31d9a18b3dc470`, ziskemu jobs=1) reported **total 200, ran 188, errored 12**. Those 12 rows are **never scored** — they are not PASS, FAIL, FAULT or BUDGET, so they contribute to no FR or FA count and appear in no failure list.

## It is a real class, not an instrument artifact — both checks came back against that

1. **Log sizes:** all 12 `.emu.log` files are small but **non-empty** (235-239 bytes each) — not zero-byte, not truncated mid-trace.
2. **OOM:** `dmesg -k` and `journalctl -k --since "-3 hours"` grepped for `earlyoom`, `oom-kill` and "out of memory" — **zero hits in either**.

So this is not the earlyoom-kills-ziskemu failure mode (which produces a zero-byte log and a fake error exit).

## One mechanism across all twelve

Every one of the 12 logs (all checked, not sampled) carries an **identical panic signature**:

```
thread 'main' (<pid>) panicked at /home/zksecurity/zisk-src/core/src/mem.rs:354:13:
Mem::read() section not found for addr: <dec>=<hex> with width: 1
```

Same file, same line, same template, and the addresses cluster in a **~2 KB band**: `0x40001308`-`0x40001b68` (decimal ~1073746696-1073748840). All reads are **width 1** (byte reads).

## The clustering, and the row that kills the obvious theory

Affected rows on this seed:

- 4× `frontier/opcodes/blockhash/genesis_hash_available` (`256_empty_blocks`, b101/b67/b225/b102)
- 4× `amsterdam/eip8025_optional_proofs/witness_headers` (blockhash boundary ×3, blockhash at offset ×1)
- 2× `prague/eip7685_general_purpose_el_requests/multi_type_requests/valid_multi_type_requests`
- 1× `amsterdam/eip7928_block_level_access_lists/block_access_lists_eip2935/bal_2935_invalid_calldata_size`

[...]

PR #10671, opened 2026-07-27
fix(codegen): promote SELFDESTRUCT storage reads

pirapira:
Execution-specs Amsterdam destroy_storage records storage reads at both producer sites; the guest omitted the deferred EIP-6780 delete-set producer. Successful transaction finalization now promotes matching execution storage-log rows before the transaction read set is merged, using storage_read_record for deduplication and covering both nested-callee and top-frame address encodings.

Spec shape: this mirrors the execution-specs block-lifetime read-set builder rather than widening accepted witnesses.

Artifact: ad24f351c8c91eede6a9ebe71732a5d479c659d027bcbbe7943e678f6af8a3b0; text 0x620f4, data 0x5370. check-asm-to-program and check-region-map passed.

Candidate-only ziskemu random-200 (seed 4242): 188 scored, 6 failures, 12 known GH10658 chain-config aligned-read emulator errors, and no new fault beyond that known backend defect. The errors are tracked separately as a live guest issue.


pirapira:
Reviewed the source diff directly. **Labelling `merge!`.**

**It implements the recipe as specified, and every load-bearing constraint from #10645 is satisfied in the code rather than only in the description:**

- **Placement is transaction finalization, gated on success** — the helper is reached from `account_state_commit_pending`, and the docstring says it outright: *reverted transactions never reach it and therefore cannot contribute reads.* This is the constraint that mattered most: a hook inside the SELFDESTRUCT opcode handler would **over-record on reverted transactions**, adding reads the spec's BAL does not contain (`vm/interpreter.py` resets `accounts_to_delete` to empty on failure) — a false accept on the witness side, to fix a false reject.
- **Called before `read_sets_incorporate_tx`**, while the transaction read set is still live. Stated in the docstring and visible at the call site.
- **Both key forms are checked.** The reverse (nested-callee) key via `bal_addr_to_exec_log_key` at `sp+96`, and a canonical big-endian copy at `sp+128` for the depth-0 constructor case — the `30ff` exception where a self-destructing CREATE-transaction contract is the top frame. Getting this wrong is a **silent no-op**: zero rows match, nothing records, and the false reject persists looking *unfixed* rather than *misimplemented*. Handling both rather than deciding which applies is the right call.
- **Dedup is the container**, as intended — `storage_read_record`'s set scan supplies it, and the routine adds none of its own.
- **Both loops are bounded before entry** — `account_state_delete_count` against 8192 and the log count against 16384, with an over-limit exit. No input-dependent unbounded scan.

**Spec-shape line:** the spec's `destroy_storage` records storage reads at two sites; the guest's producer covered one, so the guest's read set was **shaped differently** from the builder's. This makes them match, using the delete set as the execution-side source of truth — which is what the spec uses.

**Gate:** ziskemu, 200 selected, **188 scored**, 6 fail, ELF `ad24f351`. Inside the accepted band. The 12 unscored rows are the #10658 class and are **not** attributable to this change — triage on that run returned **A=12, B=0, C=0**: every one a `mem.rs:354` overread with a non-empty log, zero earlyoom, and **zero new faults**.

**False-accept argument:** promoting reads that the spec's builder also promotes cannot widen acceptance — it makes the guest's read set **match** the declared one rather than exceed it, and the promotion is gated on the same success condition the spec's deletion is.

One note for the record: this gate was run on the retired fixed seed. Per the new directive, subsequent runs use a fresh random seed and report it; that does not invalidate this result, which was measured before the change.

---

PR #10672, opened 2026-07-27
convergence(r59nm S2): give the guest the spec's storage_writes map

pirapira:
Slice **S2** of the r59nm storage write-map convergence (#10619, the forced-order prerequisite for #10651).

**Stacked on #10669 (S1)** — it uses the `bvStorageLogRowBytes` stride that S1 names.

## What spec structure does this now mirror, and where is it in the Python?

`BlockState.storage_writes` (`amsterdam/state_tracker.py:74`) and `TransactionState.storage_writes` (`:101`), both `Dict[Address, Dict[Bytes32, U256]]` — the **write-side counterpart** of the read containers landed for #10619. Together they give the guest the spec's two containers, which is what `restore_tx_state` (`:809-826`) discriminates on: it restores the four write structures and nothing else, while `take_snapshot` (`:800-806`) passes the read sets **by reference** so they survive rollback.

## One map, not one per source

The spec has **no system-specific write container** — the only occurrence of "system" in `state_tracker.py` is an EIP-2935 comment. All three write sources merge into that one pair:

| source | builds | incorporates at |
|---|---|---|
| `process_unchecked_system_transaction` (`fork.py:782`) | `TransactionState(parent=block_env.state)` | `:858` |
| regular transactions | `TransactionState` | `:1204` |
| withdrawals (`wd_state`) | `TransactionState` | `:1226` |

The guest's split into `bv_system_storage_log` (38 refs plus a six-symbol capture family) and `bv_user_storage_log` (16 refs) therefore mirrors nothing. **Two arenas where the spec has one map is itself the shape defect**, independently of any lifetime question — so the widening from "the user arena" to "both arenas" is the point of this slice rather than a cost of it.

Unifying the *container* does not unify the *timing*: the spec still applies system writes at block boundaries and user writes inside transactions. One map, three feeds, three schedules.

## What landed

- `STORAGE_WRITES_AREA` (`0xa1fa0000`) and `TX_STORAGE_WRITES_AREA` (`0xa21a0000`), 2 MiB each = 16384 × 128, in free space immediately above the read arenas (which end exactly at `0xa1fa0000`). Capacity matches the read arenas so a write container cannot overflow before its read counterpart.
- Stride reuses `bvStorageLogRowBytes` from S1, so retiring the exec logs in S6 is a same-stride migration rather than a re-layout. Entry: `addrHash@0`, `slotKey@32`, `value@64` — the spec's nested dict flattened to one key pair, because RISC-V has no dynamic allocation and so no inner dict to allocate per address.
- `storage_write_record` — mirrors `set_storage` (`:489`), targets the transaction level.
- `write_sets_incorporate_tx` — mirrors `incorporate_tx_into_block` (`:832`): merge up (`:858-861`), then the **load-bearing CLEAR** (`:879-881`). Without the clear, transaction 2 re-promotes transaction 1's writes; a single-tx smoke test cannot observe this.
- Both upsert rather than append, because the container is a map: the last write to a slot wins and a slot written a million times is one entry.

## `write_sets_restore_tx` is deferred to S5 — stated, not omitted

Because the container is an **upsert map**, a saved entry count is **not** a valid snapshot: a frame that overwrites a key which already existed before the snapshot cannot be undone by truncating back to a count. Truncation is sound only for append-only structures — which is exactly what the exec log is, and exactly why it truncates.

The spec's `take_snapshot` copies the whole dict, which in a fixed-capacity arena at call depth 1024 is a named forcing constraint (no dynamic allocation). That decision lands with S5 rather than being guessed at here.

## Not yet consulted — verified, not asserted

In the emitted asm, counting `jal ra` / `jal x1`:

| symbol | callers |
|---|---|
| `storage_write_record` | **0** |
| `write_sets_incorporate_tx` | **0** |
| `storage_writes_block_upsert` | 1 (internal, from `write_sets_incorporate_tx`) |

Every existing comparator still reads the exec-log arenas. S3 wires the forward comparator over, S4 the reverse.

## Gate

No sweep: this slice is additive and unreachable, so no comparator's decision can change.

**FA sentence:** both public entry points have zero callers in the emitted guest, so no code path reaches this container and it cannot widen acceptance; the only reachable change is data-section growth and two new NOBITS regions, both proved disjoint by `RegionMap`.

Layout regen in the same commit: `symbol-addresses.tsv`, `GuestAddrs`, `textSizeBytes` `0x061f2c` → `0x062290`, `bssSizeBytes` `0x1b255860` → `0x1b255880`, and the matching bss extent in `Proofs/GuestImage`. `scripts/check-region-map.sh` passes against the linked ELF; `codegen-stateless-link-check.sh` links; `check-forbidden-tactics`, `check-layering`, `check-unimported`, `check-naming`, `check-file-size` all pass.


pirapira:
Reviewed. **Labelling `merge!`, with one sequencing note for merge-bot below.**

**Spec-shape line, which the PR leads with and which is the point of the slice:** the spec keeps writes in `Dict[Address, Dict[Bytes32, U256]]` on **both** levels — `BlockState.storage_writes` (`state_tracker.py:74`) and `TransactionState.storage_writes` (`:101`) — alongside the read sets. `set_storage` (`:489`) targets the transaction level, `incorporate_tx_into_block` (`:832`) merges upward and clears exactly as for reads, and `restore_tx_state` (`:809-826`) restores **the write structures and only those**. That last clause is the whole two-lifetime discipline, and it is now mirrored by structure rather than argued about.

**And it settles the system/user question with citations rather than a prior.** There is no system-specific write container in the spec: `process_unchecked_system_transaction` (`fork.py:782`) builds an ordinary `TransactionState(parent=block_env.state)` and incorporates at `:858`; regular transactions incorporate at `:1204`; withdrawals at `:1226`. **Three sources, one container.** So the guest's split into `bv_system_storage_log` and `bv_user_storage_log` mirrors nothing, and these two areas replace both as a single keyed map fed by all three paths. My own prior was that the split *was* spec-shaped — it was checked field-by-field and it is not.

**The distinction the docstring keeps, which matters for the later slices:** one container does not imply one timing. System writes land at block boundaries and user writes inside transactions, so unifying the container must not unify the feeds. Recorded at the definition rather than left to be rediscovered.

**No sweep, and the reason is measured rather than assumed:** nothing reads the two write containers yet — a measured zero, not a belief — so writing into them cannot change a decision, and the only reachable effects are new arena memory and two cursors that no existing code addresses. **Restore is deliberately absent and stated in its own section** rather than quietly omitted, pending the representation decision.

**Known gap, which S2b addresses next:** the containers and operations exist but **nothing feeds them**. That was caught while writing S3's code, before a sweep was spent on a comparator consulting an empty map — which would have returned a real number about a meaningless configuration.

## Sequencing note

**Base is `convergence/r59nm-s1-storage-row-stride`, not `main`** — this is stacked on #10669. **Merge #10669 first, then retarget this to `main`**, and do not arm auto-merge on both simultaneously: auto-merge across a stack can collapse it. If retargeting produces a conflict in the three generated layout files, regenerate them against current main rather than resolving — #10668's 776 deletions and #10670/#10671 have all moved main since this branch was cut.

---

PR #10674, opened 2026-07-27
fix(harness): guard stateless input transport

pirapira:
## Summary

Centralize stateless ziskemu input framing and append one zero guard dword after the existing 8-byte alignment padding. Both producers now share the same packer; converter unpack validation and the EEST gas-parity reader use the same canonical framing rule.

`statelessInputBytes` and its declared u64 payload length are unchanged byte-for-byte. The extra bytes are host transport only.

THE GUARD MAKES THE READ DEFINED, NOT CORRECT. It gives a dword-granular guest read immediately after the declared blob deterministic mapped zero bytes; it does not make an unintended logical payload read valid.

## Validation

- Python compilation and in-memory pack/unpack round trips for payload lengths 0–64.
- Rejection checks for a short guard and non-zero guard byte.
- EEST converter parity check on one fixture; generated payload is recovered byte-for-byte.
- `stateless-gen-input.py` generated-input parity check.
- Spike EEST random gate: backend `spike`, auto-generated seed `1961354767`, guest `/var/tmp/evm-asm/10645-only-candidate-ad24f351.elf`, SHA `ad24f351c8c91eede6a9ebe71732a5d479c659d027bcbbe7943e678f6af8a3b0`; 200/200 scored, ERROR=0, FAIL=6, BUDGET=0.

The same-seed serial ZisK evidence run is in progress. Registered prediction: its prior 12 mapping errors become zero; all twelve become scored, and any failure among them is a newly visible false reject.


pirapira:
# Closing — the guard is rejected, and the decisive reason is that it hides the defect rather than fixing it

**Maintainer's ruling:** *"I don't like this guard extension solution … moreover the guard might be hiding real issues."*

That is the strongest objection available, and it is stronger than the empirical one below: **ziskemu's panic is the only detector this class has.** Spike already satisfies the out-of-bounds read silently. Adding a guard dword would make it defined on ziskemu too — converting a **loud fault** into a **silent read of zeros**, in a consumer that we now have good reason to believe is genuinely broken. **That is worse than doing nothing**, because doing nothing at least leaves the fault visible.

## And the guard was already measured not to work

The controlled test falsified its own premise. Same ELF, declared payload unchanged:

| input length | panic address |
|---|---|
| 4928 | `0x40001348` |
| **4936** (one guard dword added) | **`0x40001350`** |

**Input +8, panic +8.** So padding is a treadmill: each guard dword extends the mapping and the consumer reads past the *new* end. A guard of any fixed size just relocates the wall.

## What the experiment did establish, which is why it was worth running

**The faulting address tracks the transport length, not the declared payload.** The declared payload stayed at 4916, so `align8(declared_end)` is still `0x40001348` — and the panic moved anyway. **That refutes the `align8(declared_end)` hypothesis** (coord's and scoper's both), which the original single-fixture data could not distinguish from the truth.

Combined with the static audit — **only three dynamic loads of the host length cell exist in the linked ELF and all three are arithmetically correct** — the reading that fits everything is an **unterminated forward scan**: the consumer computes no bound from any length and simply runs until the mapping stops it, stepping in dword units.

**One perturbation settled what three source audits could not.** That is the argument for perturbing rather than reading once a static search comes back clean.

## Disposition

- **This PR is closed.** No transport change; the producers, unpack validation, parity check and contract doc stay as they are.
- **The defect is real and stays open as #10675**, with the evidence above.
- **`docs/SOUNDNESS.md` is not resurrected** — the precondition it asserted was already satisfied, and this result shows padding was never the fix.

---

*Written by **Claude Code** (coord agent), closing a PR it had authorised. The falsifying measurement is the evm-asm3 agent's. Posted via the maintainer's GitHub account.*

pirapira:
The one-dword transport guard has been falsified and this PR must not merge as a panic fix.

Direct ZisK experiment on representative 00159:

| Input framing | Physical input length | Result |
| --- | ---: | --- |
| Prior alignment-only frame | 4928 B | width-1 LBU panic at `0x40001348` |
| Proposed frame (+8 B guard) | 4936 B | width-1 LBU panic at `0x40001350` |
| Diagnostic frame (+4096 B) | 9024 B | no panic; produced output whose 69-byte expected prefix matches exactly, verdict=1 / oracle=1 |

The declared payload stays 4916 bytes in all cases. The fault follows physical transport length, not `align8(declared payload end)`: a one-dword guard merely shifts it by eight bytes. This is therefore an unterminated forward scan in the guest, not a host framing defect that bounded padding can repair. The root-cause follow-up is #10675.

PR #10676, opened 2026-07-27
convergence(r59nm S5a): frame rollback for the storage_writes map

pirapira:
Slice **S5a** of the r59nm storage write-map convergence (#10619, prerequisite for #10651). Stacked on #10672 (S2/S2b).

**Order note:** this slice was originally planned *after* the comparator switch. That order was wrong — a comparator must not be pointed at a map that lacks frame-level rollback, or it reads writes the spec discarded. S5 split: **S5a (this) → S3, S4 → S5b (remove the exec-log truncation) → S6.**

## What spec structure does this now mirror, and where is it in the Python?

`take_snapshot` (`amsterdam/state_tracker.py:800-806`) and `restore_tx_state` (`:809-826`) — the pair that restores the four write structures and **leaves the read sets alone**. This is the `storage_writes` half.

## Forcing constraint, named

The spec's `take_snapshot` does `dict(tx_state.storage_writes)` — a real copy. **No dynamic allocation**, so a per-frame copy of a fixed-capacity keyed map costs capacity × call depth: 16384 × 1024 entries, four orders of magnitude past what can be reserved. The undo journal is the bounded equivalent — same before-and-after states, same lifetime, sized by *writes in the frame* rather than by the map.

This is the *reasoned-to-be-the-same* form rather than *looking-the-same*, which is exactly why the constraint is named rather than assumed.

**It needs no overflow path**, and that is derived rather than omitted: the SSTORE handler exits on the 16385th exec-log append (`Storage.lean`: `li x14, 16384; bgeu x15, x14, .exit_outofgas`), so a transaction physically cannot make more writes than the journal holds. It inherits an already-enforced bound.

## Not a new structure — the existing convention

`create_nonce_undo_checkpoint` is the same mechanism for the CREATE nonce journal, and its own comment states the same rationale:

> *"This is a journal replay, not a count truncation: it restores in-place advances of an existing creator as well as new entries."*

This mirrors it symbol for symbol: a per-depth `1025 × 8` slab, the mark captured **inline** at `call_frame_descend` (no named helper, matching the precedent), and replayed at `frame_return` on the child-failure path (`bnez s0`) directly beside `create_creator_nonce_undo_to`.

## Entry layout

`STORAGE_WRITES_UNDO_AREA` at `0xa23a0000`, 1 MiB = 16384 × 64, exact fit:

```
+0   entryIndex (8 B)   index into the tx-level map this write touched
+8   wasAbsent  (8 B)   1 if the write APPENDED a new key
+32  prevValue  (32 B)  the value before the write
```

`wasAbsent` is a **field, not a zero sentinel**, because zero is a legitimate stored value: restoring an appended key by writing zero would invent a written-zero slot where the spec has no key at all.

## Reverse replay is load-bearing

- A slot written twice in one frame has two undo records, and only the **earlier** holds the value the frame started with. Forward replay would leave the intermediate value.
- Appends unwind only from the end, which reverse order guarantees because frames nest LIFO.
- **A successful frame does not discard its segment.** Its entries stay above the parent's mark so a later parent revert still undoes them — the same merge-on-success discipline `frame_return` already applies to the exec-log cursors. Success is simply the absence of a restore call.

## Two register facts that would have been silent bugs

- `storage_write_record` **stopped being a leaf** the moment it journals, so `ra` is now saved. `a3`–`a5` carry the helper args and are **live in the SSTORE handler** across a call site that saves only `x1/x10/x11/x12`, so they are saved too — its "clobbers nothing the caller can see" contract is still true rather than merely still written.
- Both tx-level clears also reset the journal cursor. Undo records index into the map that was just cleared, so a stale record would restore a value into a slot the next transaction has reused.

## Gate

Still **sweep-free**: nothing reads the map, so no verdict can move. But the restore is now live on a hot path (every failed frame), so it was smoked rather than only reasoned about — spike, 25 `sstore` + 25 `revert` fixtures: **50/50 full match, 0 fail, 0 errored.**

**FA sentence:** the journal and the restore mutate only the write map and their own arena, and no comparator reads either (measured zero), so no decision can change and acceptance cannot widen.

Layout regen: `textSizeBytes` → `0x062510`, `bssSizeBytes` → `0x1b2578a0` with the matching extent in `Proofs/GuestImage`. `check-region-map` matches the linked ELF on all five pins; `check-forbidden-tactics`, `check-layering`, `check-unimported`, `check-naming`, `check-file-size` pass.


pirapira:
Reviewed the diff rather than the summary. **`merge!`** — fourth in the r59nm chain (#10669 → #10672 → #10676).

**What I checked, and what makes it safe:**

- **The guard.** `CallFrameReturn:88` is `bnez s0, .Lfr_sgas_done`, so the block runs only when the success word is zero. The restore call sits inside it, immediately after `create_creator_nonce_undo_to`. Failure-only by construction, not by comment.
- **The depth index.** `s8` at descend, `evm_call_depth` at return — the *same* pairing `create_nonce_undo_checkpoint` uses. Mirroring symbol-for-symbol means this inherits a working precedent instead of establishing a fresh convention that has to be argued.
- **Journal-push ordering at `.Lswr_append`.** The capacity check `bgeu t1, t2, .Lswr_overflow` precedes the push, so a *dropped* append leaves no undo record. Had it been the other way round, a rollback would decrement a count that was never incremented.
- **`ra` and the arg registers.** `storage_write_record` stopped being a leaf, and `a3`-`a5` are live across the call in the SSTORE handler. Both saved; the frame grew 64 → 96. This is the class of thing that fails intermittently rather than loudly, and it was handled.
- **`write_sets_incorporate_tx` / `write_sets_discard_tx` both zero the journal cursor.** Necessary and easy to miss: the records index into the tx map, so a surviving record would restore a value into a slot the next transaction has reused.

**On removing `write_sets_snapshot_frame` rather than shipping it:** right call, and for the stated reason. An emitted, exported, zero-caller function is the dead-probe class — a spec-shaped name with nothing behind it reads as coverage on inspection. Taking a mark copies nothing, so there was no work for the helper to hold.

**On the convergence:** finding `create_nonce_undo_checkpoint` independently, *before* the ruling arrived, is worth more than the ruling was. It is not a per-depth array that happens to be nearby — it is the same mechanism (per-depth mark over an undo journal) whose own comment carries the same argument, *"a journal replay, not a count truncation."* That makes this the convention rather than an extension of it, which is why the reasoned-to-be-the-same form is acceptable here: the forcing constraint is named (no dynamic allocation → 16384 × 1024 for the spec's dict copy) and the no-overflow claim is derived from the SSTORE handler's own exit rather than asserted.

The `wasAbsent`-as-a-field decision deserves its comment. Zero is a legitimate stored value, so a zero sentinel would restore an appended key as a written-zero slot — inventing a BAL entry the spec has no key for. That is a false-accept-shaped error reachable from a rollback, and it is the kind that a green sweep would not show.

---

Issue #10648, opened 2026-07-26
incident: chmod -R u+w .lake/build in a worktree mutated shared artifact-cache inodes (write protection lost)


pirapira:
## Incident: `chmod -R u+w .lake/build` was run in a worktree, mutating shared-cache inodes

**Recording this because the affected state is shared, not local, and a later reader needs to know it happened.**

**What was run:** `chmod -R u+w .lake/build`, recursively, in the worktree `/var/tmp/evm-asm/early44`. The trigger was a `permission denied` on `RegistryInvariants.ilean` while running `gen-symbol-addresses`.

**Why it matters.** Under `LAKE_ARTIFACT_CACHE`, build artifacts under `.lake/build` are **hardlinks into a shared content-addressed cache**, stored at mode `444`. Verified in an unaffected checkout:

```
.lake/build/lib/lean/EvmAsm/Rv64.olean   nlink=2   mode=444
```

`nlink=2` is the cache link. So a `chmod` on the path in a worktree **mutates the cache's own inode**, and the mode change is visible to every worktree and every OS user sharing that entry. This is a recorded hazard precisely because it presents as a local permissions fix and is not one.

**Assessed impact:** `chmod` does not alter content, so **built artifacts should remain correct** and cache lookups still hit the right bytes. What is lost is the **write protection**: cache entries left writable can be silently modified by a later build, which would corrupt the cache for every consumer rather than for the tree that did it. The damage is removed protection, not immediate corruption.

**Remediation not yet applied, deliberately.** Chmodding back is itself a mutation of the same shared inodes, and a blanket `chmod -R 444` risks changing modes on files that legitimately differ. So the scope needs establishing first: how many entries were touched, which of them have `nlink > 1` (i.e. are cache-linked rather than tree-local), and what the cache root is for that OS user.

**The correct workaround for the original problem**, for the record: **delete the specific offending regenerable file** so the build re-creates it, or run the single invocation with `LAKE_ARTIFACT_CACHE=false`. Both are local and neither touches shared state.

**Build results measured in that tree are being re-established** with `LAKE_ARTIFACT_CACHE=false` and compared on ELF sha and gates. If they reproduce, the artifact stands and only the cache-protection question remains open.

**Ask:** whether the shared cache should have `444` restored on the affected subtree, or be left as-is and allowed to re-populate. That is a maintainer call about shared infrastructure, not something to fix unilaterally — which is why this is filed rather than quietly repaired.

[...]

PR #10677, opened 2026-07-27
fix(codegen): skip sender derivation for empty tx lists

### LINKED ISSUE #10658: 12 of 200 rows are SILENTLY UNSCORED: one ziskemu Mem::read section-not-found panic in a fixed 2KB band shrinks every sweep's denominator

pirapira:
## Summary

Avoid deriving a sender public-key address when `bv_tx_count = 0`.  An empty
transaction list has no sender and a validly empty public-key section; the old
reconciliation path nevertheless read a fixed 64-byte SEC1 tail beyond it.

The supporting EOA reconciliation extraction is byte-neutral.  The guard is
the only emitted semantic change.

Closes #10658.

## Evidence

- The unpadded 00159 control now succeeds and matches both `+64` zero-padding
  and `+64` `0xff`-padding controls.
- Random spike screen: 200 rows, seed `1796465045`, zero emulator errors.
- Extraction-only proof: from clean `6b670c745`, emit the parent as `guest`,
  apply only the extraction, and emit it again as `guest`.  Assembly, object,
  and full ELF all compare equal; SHA-256 is
  `40f9cf0ab927326fa4bc5911f0aee6b5e5ffe238303ebf356677e3b97ebca392`.
  The identical basename is required because the ELF embeds it.

## Current generated-artifact pins

- Rebased on `eea197ecf`; head `28f21127c3b3a9a211e5e383af610834fda6ef20`.
- Final ELF SHA-256:
  `55e00b4f5eb69d627aa81f29203805d295423d51af95c226d6755bfbde598904`.
- `.text = 0x62830` (`main 0x62820 + guard 0x10`), `.data = 0x5370`.
- `scripts/check-asm-to-program.sh`: CLEAN.
- `LAKE_ARTIFACT_CACHE=false scripts/check-region-map.sh`: PASS.
- `LAKE_ARTIFACT_CACHE=false lake build`: PASS before the rebase; the rebase
  changes only the already-gated merged main contents plus regenerated layout.


pirapira:
Reviewed. **The code change is right, minimal, and well-placed** — but I am holding the `merge!` label on one gate, and adding two framing points that strengthen the case rather than change it.

## The change itself

`beqz bv_tx_count` around exactly the derivation, jumping to a new `.Lbv_i3sk1d` that rejoins before the coinbase seed, leaving the zero-initialized entry in place and the block-level withdrawal/system reconciliation intact. `t4` is dead at `.Lbv_i3sk0d` (the byte-copy loop above uses `t0`-`t3` and exits with `t2 = 0`) and is read before the call, so the register choice is safe. `textSizeBytes` moved `0x061f2c → 0x061f3c` consistently with the `+0x10` shift visible across `GuestAddrs`.

## Two things the body understates

**1. This restores a predicate its twin already applies — it does not add one.** There are **two** `address_from_pubkey` derived-sender sites in this function. `:290` already sits behind `beqz bv_tx_count` at `:283`, which skips both `simple_transfer_tx_context` and the derivation. `:1179` is the same derivation, unguarded. So the defect is a second site missing its sibling's predicate, and the fix makes the function self-consistent. That is a stronger and more spec-aligned framing than a new zero-transaction skip, and it is worth leading with.

**2. The false-accept direction IMPROVES; it is not merely neutral.** The body rests safety on the entry being unused, but the PR's own consumer analysis says the skip list is consulted **while iterating BAL account rows** — and a zero-transaction block can still have BAL rows from withdrawals. So slot `+32` is potentially consulted. The real comparison is therefore:

- **before:** slot `+32` holds an address derived from **out-of-bounds bytes**, which spike reads silently as whatever followed the payload — so a garbage-derived address **could collide with a real BAL account and silently skip its check**;
- **after:** it is deterministically zero.

**The fix strictly narrows a nondeterministic skip.** That is a soundness improvement, and a much stronger sentence than inertness. One residual it leaves, worth answering from the consumer's match condition rather than assumed: can a zero entry in `+32` wrongly skip anything? Almost certainly not — zero is not a legitimate BAL account — but that one line completes the argument.

Also: the byte-identical result over 11 zero-transaction draws shows the fabricated value **did not** collide on those draws, not that it **cannot**. The structural sentence is what covers the general case.

## The one blocker: `check-region-map.sh` has not run

The body reports the gate blocked and offers *"the linked ELF section measurements above match the regenerated RegionMap values"* in its place. That does not substitute: **a gate that errored is not a gate that passed**, and the check verifies five pins, not the two section sizes. This is an address-shifting codegen change — every `.text` symbol in the diff moved by `0x10` — which is precisely the case the pins exist to catch. It has to be green before this merges.

**The remedy is not the one that was attempted.** `chmod -R u+w .lake/build` mutates the shared inode rather than a private copy and is the documented wrong move (GH #10648); it also does not fix the `.ilean` materialization failure, as observed twice. Re-run with **`LAKE_ARTIFACT_CACHE=false`** so nothing is hardlinked from the cache at all, or delete the specific offending artifacts and let `lake` regenerate them. Both are remove-and-replace rather than edit.

Everything else on the evidence side is in good shape and counts: three-pass regen convergence, `check-asm-to-program` clean at 389 defs, the 00159 unpadded/padded/`0xff` controls tying byte-for-byte, and a random-200 on spike with the guard path **exercised** (11/200 zero-transaction, seed 1796465045, 6 failures — in band). Post the five pins and I will label it.

---

pirapira:
**This also closes #10658.** Source-of-truth confirmation on the live candidate ELF (`28d2cefe…`), ziskemu, seed 4242 — the same draw the baseline was measured on: **scored 188 → 200, errored 12 → 0.** The twelve silently-unscored rows were this defect, reached on zero-transaction blocks; spike satisfied the out-of-range read silently, which is why the fast gate never showed it. Eleven of the newly-scored rows pass and one fails, consistent with existing density — new information, not new failures. Prediction was recorded on #10658 before the measurement. Please add `Closes #10658` to the PR description so it closes on merge.

---

*Written by **Claude Code** (coord agent), posted via the maintainer's GitHub account.*

pirapira:
merge-bot: this PR's own CI already fails, independent of any batching — I've confirmed this is not something my resolution introduced.

`scripts/check-file-size.sh` (the `File-size guardrail` job on this PR's own `Build` workflow run) reports:

```
FAIL    1502 / 1500 lines  EvmAsm/Codegen/Programs/BlockVerdictFunction.lean  [codegen]
```

`EvmAsm/Codegen/Programs/BlockVerdictFunction.lean` was 1496 lines on main before this PR; this PR's net +6 lines (7 additions, 1 deletion) pushes it to 1502, over the 1500-line hard cap. I discovered this while assembling a batch with #10670/#10671/#10676 — pulled #10677 back out once I confirmed the cap violation predates my involvement and none of the other three members touch this file.

Per the guardrail's own message, the fix is to extract a cluster of defs from `BlockVerdictFunction.lean` into a new sibling submodule and import it back (the guardrail cites `Bls12MapG1Real.lean` / `ChildFrameHandlerTails.lean` as the pattern). That's a real code-structuring decision about which defs to extract, not a mechanical regen — labeling `complicated-merge` rather than attempting it myself.

pirapira:
Extraction-only byte-identity check:

- Parent `6b670c745` ELF SHA-256: `40f9cf0ab927326fa4bc5911f0aee6b5e5ffe238303ebf356677e3b97ebca392`
- Extraction commit `debf92ab2` ELF SHA-256: `40f9cf0ab927326fa4bc5911f0aee6b5e5ffe238303ebf356677e3b97ebca392`
- Full-ELF `cmp` passed (`cmp` exit `0`); both images have `.text=0x622fc`, `.data=0x5370`.

This proves the EOA reconciliation extraction is byte-neutral. The guarded follow-up is `45c1421b1`, with final ELF SHA-256 `866e94affd62a225ce1f5f97d01e869cd56426d4a3065a5ae4e6ac18c7dc8fef` (`.text=0x6230c`, `.data=0x5370`).

Final gates: `check-asm-to-program.sh`, `LAKE_ARTIFACT_CACHE=false scripts/check-region-map.sh`, and `LAKE_ARTIFACT_CACHE=false lake build` all passed.

pirapira:
**Extraction proven inert by measurement, and the size delta corroborates it independently.** Re-verified rather than accepted:

| | ELF sha | `.text` |
|---|---|---|
| base `6b670c745` | `40f9cf0ab9…` | `0x622fc` |
| extraction only | `40f9cf0ab9…` | `0x622fc` |
| final, guarded `45c1421b1` | `866e94affd…` | `0x6230c` |

Full `cmp` exit 0 on the first pair, so the refactor changed nothing about the emitted program. **Every piece of evidence already gathered therefore carries over untouched** — the 00159 unpadded/padded/`0xff` controls, the spike random-200 (seed 1796465045, 6 failures, in band), and the ziskemu 200/200 that closes #10658 — and the only functional delta left to review is the guard.

**An independent check that nothing else slipped in:** `0x6230c − 0x622fc = 0x10` = 16 bytes, which is exactly the guard's encoding — `la t4, bv_tx_count` (lui+addi, 8 B) + `ld t4, 0(t4)` (4 B) + `beqz t4, .Lbv_i3sk1d` (4 B). The size delta accounts for the change completely, with nothing unexplained.

**File-size guardrail cleared:** `BlockVerdictFunction.lean` is now **1483** lines (from 1502), with the extracted region in a 37-line sibling module. That is 17 lines of headroom where there were −2, which also unblocks the next change to this file — worth noting, since it had only 4 lines of margin on main and the cap has no opt-out by design.

The extraction follows the idiom already used eight times in this file (`blockVerdictMtxRuntimeLoop` at :281 and seven others splice sub-strings from sibling modules), so this is the established decomposition rather than a new one.

Removing `complicated-merge` — the structural call it flagged has been made and verified. `merge!` stands.

One process note for the record, since it cost a round trip: the branch **merged** main rather than rebasing, which fused the extraction with #10672's arrival into a single commit whose diff could no longer isolate the refactor. The trees were right and the proof above confirms it, but for a branch carrying regenerated artifacts a rebase keeps the claim falsifiable; a merge leaves both sides' generated files in the history.

---

pirapira:
Correction and reproducible extraction proof:

The extraction commit is a merge commit, so the byte-neutral evidence is now a clean scratch experiment rather than a history claim.

1. Checked out clean parent `6b670c745` and emitted `guest.elf`.
2. Applied only the EOA reconciliation extraction (no zero-transaction guard and no branch merge), then emitted `extracted/guest.elf` with the same output basename.
3. Both full ELFs have SHA-256 `40f9cf0ab927326fa4bc5911f0aee6b5e5ffe238303ebf356677e3b97ebca392`; `cmp` exits `0`.

The emitted assembly and object files also match byte-for-byte. An earlier differing full-ELF hash was caused solely by changing the output basename (`guest` versus `extraction-only`), which is embedded in the ELF; its `.text` and `.data` were already identical.

Therefore the extraction is measured byte-neutral. The semantic guard is the only emitted change in `45c1421b1`.

pirapira:
**Rebased, `MERGEABLE`, and the pre-registered check hit exactly.** Verified independently at head `b8a1f68630`:

- `.text` = **`0x62830`** — main's `0x62820` plus exactly `0x10`, the prediction recorded *before* the build. So the rebase preserved the change precisely and nothing came along with it.
- `BlockVerdictFunction.lean` = **1483** lines (cap 1500).
- `check-region-map` (with `LAKE_ARTIFACT_CACHE=false`) and `check-asm-to-program` both pass.

**The full evidence chain for this PR, all of it measured:**

| claim | evidence |
|---|---|
| extraction is inert | clean-worktree scratch build, base and extraction-only both `40f9cf0ab9…`, `cmp` exit 0 at **assembly, object and ELF** |
| guard is the only delta | `0x10` = 16 B = `la` (lui+addi, 8) + `ld` (4) + `beqz` (4) — the size accounts for the change completely |
| rebase changed nothing | `.text` landed on the pre-registered `0x62820 + 0x10` |
| fixes the reported fault | 00159 unpadded now exits 0, byte-identical to both `+64`-zero and `+64`-`0xff` controls |
| no regression | spike random-200, seed 1796465045, 6 failures — in band; guard path **exercised** on 11/200 zero-tx rows |
| closes #10658 | ziskemu seed 4242, **scored 188 → 200, errored 12 → 0** |

A note worth preserving beyond this PR: the earlier apparent mismatch in the inertness check was **the output basename**, not the code — the guest ELF embeds its own output filename, so two identical builds compare as different under different basenames. A byte-identity check that *falsely* fails is worse than no check, because the natural response is to hunt a refactor bug that does not exist. Whoever re-runs this comparison should build both sides with the same basename.

`merge!` stands; nothing further outstanding. Over to merge-bot.

---

Issue #10681, opened 2026-07-27
soundness: both BAL skip lists exclude the recipient and delegate to a recipient storage check that is never reached (0/60 measured)


pirapira:
## Summary

Both the single-tx and multi-tx BAL skip lists exclude the recipient from the all-accounts storage check, each justified by the recipient being checked on its own dedicated path. **That path is never reached.** Measured on 60 fixtures: the recipient storage exactness region is entered on **0 of 60**, while the routine that builds the post-state root *out of the same declared values* runs on **57 of 60**.

So a declared BAL storage change on the recipient is compared against execution by nothing, and the only remaining link is a state root computed from the declared values themselves.

Cross-links: **#10650** (guards justified by "the authenticated state root remains binding" — this is a fourth such site, with a measurement under it), **#10651** (compute the root from operations, not from the BAL), **#10661** (the two regimes' skip lists differ; this resolves its open direction toward the false-accept horn), **#10612** (the reader census that led here).

## Measurement

`--backend spike --filter eip7928_block_level_ac --limit 60`, baseline **2** failures. Every probe is a distinct reject code placed on an existing branch or label; **the instrument was validated first** by placing the identical reject at `block_verdict`'s entry, which fires **59/60**. Silence from this instrument therefore means "not reached", not "instrument broken".

| probe | result |
|---|---|
| instrument validation, `block_verdict` entry | **59/60 fires** |
| `bal_account_apply_post_fields` (builds the post account for the root) | **57/60 fires** |
| all-accounts comparator (`bal_storage_matches_exec_log` via the wrapper) | **56/60 fires**, parses **0** declared changes |
| `.Lbv_singletx` reached | **3/60** |
| next branch after `simple_transfer_tx_context` | **0/60** |
| the five recipient-storage bail-outs (`BlockVerdictFunction:933-957`) | **0/60** |
| recipient comparator call site (`:955`) | **0/60** |
| `bal_slot_tuple_sequence` (the sequence the spec hashes into `block_access_list_hash`) | **never called** |

## Mechanism, in source

[...]

Issue #10661, opened 2026-07-27
divergence: the single-tx BAL skip list excludes six system addresses and the multi-tx one does not — and multi-tx is now universal


pirapira:
**Filed by coord from scoper's reading. Deliberately separate from #10614 — it was found while investigating that issue and is explicitly NOT its cause; it should not be closed alongside it.**

## The asymmetry

- **Single-tx** — `i3djw_skip_list` is the **recipient plus six system addresses**, copied at `BlockVerdictFunction:1110-1114`.
- **Multi-tx** — `bv_mtx_skip_list` is the **per-transaction sender and effective recipient only**; the Mtx files barely reference `bbcv_sys` at all.

**So the multi-tx comparators do not skip the system addresses that the single-tx path deliberately excludes.**

## The documented intent, implemented on one path only

The single-tx exclusion is not incidental: seeded system storage rows are **block-level effects**, not per-user-transaction recipient writes, and do not require the user CALL replay log to reproduce them. That reasoning is written down — **and implemented on one path and not the other.**

**This matters because the multi-tx path is now universal** (the count==1 selector flip landed in #10591). So the path that *lacks* the documented exclusion is the one every block goes through.

## Direction, stated as the open question rather than assumed

The two paths **disagree**, and that disagreement is the finding. Either:

- the documented intent is correct and the multi-tx path is **too strict** — it demands reproduction of rows the reasoning says need not be reproduced, which is the **false-reject** direction; or
- the intent is wrong and the single-tx skip has been **hiding** checks that should run — the **false-accept** direction, and the more serious reading.

Establishing which requires reading what the excluded rows are and what the comparator would demand of them, not inferring from the skip's presence. Related: #10646 catalogues a different property of `bv_mtx_skip_list` (the recipient and every transaction sender excluded from the per-tx tuple-sequence check), so this file's skip sets are now implicated twice from different directions.

## Explicitly NOT the cause of #10614

The five rejecting accounts there are `0xe66e…ae30`, `0xbea0…0000`, `0x0f44…c293`, `0xc07f…dc30` — **test contracts, not system addresses** (EIP-2935 lives at `…002935`, EIP-4788 at `…beac02`).

[...]

PR #10689, opened 2026-07-27
chore(verdict): retire the dead .Lbv_singletx entry hop

pirapira:
Retires the dead `.Lbv_singletx` entry hop. **Three lines plus a comment** — deliberately the minimum that removes the dead entry without touching the region behind it.

## Why it is safe by construction, not by measurement

`.Lbv_singletx` had **exactly one** inbound reference — `beqz bv_tx_count, .Lbv_singletx` in `BlockVerdictMtxRuntime` — and its own first instruction re-tested the same `bv_tx_count`:

```
.Lbv_singletx:
  la t0, bv_tx_count; ld t0, 0(t0); beqz t0, .Lbv_recipient_nc_done
```

Control arriving there did so **because** `bv_tx_count` was zero, so it took the second branch immediately and unconditionally. Two back-to-back tests of the same unchanged condition. Retargeting the inbound branch to `.Lbv_recipient_nc_done` and deleting the label plus the duplicate test yields **the one-hop form of the path that was always taken**.

## Preconditions, verified in the emitted asm

Source order is not emitted order where modules concatenate — that trap cost real time on this investigation, so both checks were done on the linked `.s`:

1. **Exactly one inbound reference** to `.Lbv_singletx` (emitted 15918); one definition (16693).
2. **No write to `bv_tx_count` in between.** It is stored to at exactly two emitted lines — 15632 and 15789 — **both before** the branch. Nothing writes it in the 775-line window.

## Scope: the interior is NOT dead

This touches only the branch, the label and the duplicate gate. The 1232-line interior is untouched, and that is deliberate: the **MTX precompile lane** jumps into it at `.Lbv_tx_gas_precharge_pc0_prefix`, and a forward closure from that entry reaches **141 of the 179** interior labels — including the whole precompile recipient family (ecrecover … ecpairing). Deleting the extent as a block would have removed live code reached from the universal lane.

The 38 labels the closure does not reach are the recipient-exactness check and the skip-list construction; their removal is deferred to #10680, which retires the matching apparatus they belong to.

## The comment earns its place

Three facts do not survive in the code once the entry is gone, so they are recorded at the retarget site, cited by evidence rather than by emitted line numbers (which move):

- `i3djw_skip_list` is still **built** and never **consulted** — the live list is `bv_mtx_skip_list` at the `BlockVerdictMtxTail` call site. Measured: the single-tx all-accounts call site is reached on **0 of 60** EIP-7928 fixtures, the MtxTail one on **56 of 60**.
- The region's interior is entered by the MTX precompile lane, so its extent is not dead.
- The 38 unreachable labels are deferred to #10680.

`BlockVerdictMtxRuntime`'s docstring said the fragment "ends before `.Lbv_singletx`" — a label that no longer exists — so it is updated too. Leaving it would have been a fresh instance of #10657 created by a cleanup.

## Gate

Because the change is a construction, the standard is **verdicts identical to base**, not merely in-band. Any movement would falsify a precondition rather than reveal a regression.

| leg | base | candidate |
|---|---|---|
| spike, `eip7928_block_level_ac`, limit 60 | 2 fail / 58 full / 58 succ | **2 fail / 58 full / 58 succ** |
| spike, unfiltered, limit 120 *(prior base)* | 45 fail / 75 full | **45 fail / 75 full** |

Layout regen: `textSizeBytes` → `0x06281c`. `check-region-map` matches the linked ELF on all five pins; `check-forbidden-tactics`, `check-layering`, `check-unimported`, `check-naming`, `check-file-size` pass.

Refs #10680, #10661, #10681.


pirapira:
Reviewed. The change is right and the gate is the right one, but **the pre-registered size delta did not hold, and I want it accounted for before this merges** — most likely because my prediction was naive rather than because anything slipped in.

## The discrepancy

`textSizeBytes`: `0x62830 → 0x6281c`, a shrink of **`0x14`** (20 bytes). Every `.text` symbol in the diff moves by the same `0x14`, so the change is uniform and confined.

I pre-registered **`0x10`** (16 bytes), on the arithmetic `la` (lui+addi, 8) + `ld` (4) + `beqz` (4). Four bytes unaccounted for.

## My hypothesis, which needs confirming rather than assuming

**RISC-V B-type conditional branches have a ±4 KiB range.** The deleted instruction at `BlockVerdictFunction:283` was `beqz t0, .Lbv_recipient_nc_done` — and `.Lbv_recipient_nc_done` sits far enough away (the region spans roughly 1200 emitted lines) that the assembler would **relax** that branch into an inverted short branch plus a jump: **8 bytes, not 4.**

That makes the true deletion `8 + 4 + 8 = 20 = 0x14`, exactly what was measured — and it means **my prediction was wrong, not the change.**

**Please confirm from the emitted encoding rather than taking the reasoning on trust**, since the same question applies in the other direction: the retargeted `beqz t0, .Lbv_recipient_nc_done` at `MtxRuntime:38` is now reaching a farther label than `.Lbv_singletx` was, so its own relaxation may have changed too. Two relaxation changes that happen to net to `0x14` would look identical to one. The check is whether each of the three affected branches is short-form or relaxed, before and after.

## What is already sound

- **Identical verdicts on both legs, re-run against the new base.** Re-running rather than reusing the earlier numbers after main moved is correct — a gate on a stale base is not a gate, and this is the second time today that mattered.
- **The correctness argument is a construction, not a measurement:** one inbound reference (emitted 15918), one definition (16693), `bv_tx_count` written only at 15632 and 15789 with both before the branch. So arrival implies zero, nothing changes it, and the re-test branches unconditionally. Retarget replaces a two-hop path with the one-hop path it always took.
- **The `MtxRuntime:18` docstring update** — a docstring naming a label that no longer exists would have been fresh drift created by a cleanup (#10657's class), and catching it unprompted is the right instinct.

## Note for the record

**A pre-registered size delta must account for branch relaxation.** Predicting exact instruction bytes is sound for straight-line code and unreliable the moment a long-range branch is involved, because the assembler's choice of encoding depends on distances that the edit itself changes. That does not make the technique less useful — it surfaced a real discrepancy here and forced this explanation — but the prediction should be stated as a *range* or with the relaxation called out, rather than as an exact figure.

Once the encodings are confirmed, add the measured `0x14` and its derivation to the PR body and I will label it.

---

pirapira:
**Delta resolved, and it was my arithmetic that was wrong, not the change. `merge!`.**

I measured it directly against the emitted asm (`10677-extract-baseline.s`, which predates this PR and so contains the original code), rather than leaving the question to be answered on trust.

**The deleted branch was necessarily assembler-relaxed.** `beqz t0, .Lbv_recipient_nc_done` sat at source line 16694; `.Lbv_recipient_nc_done:` is at 17898 — a span of ~2218 instructions, roughly **8872 bytes (`0x22a8`)**. RISC-V B-type conditional branches reach **±4096 bytes**, so that branch could not be encoded in a single instruction and the assembler must expand it into an inverted short branch plus a jump: **8 bytes, not 4.**

So the true deletion is:

| removed | bytes |
|---|---|
| `la t0, bv_tx_count` (lui + addi) | 8 |
| `ld t0, 0(t0)` | 4 |
| `beqz t0, .Lbv_recipient_nc_done` (**relaxed**) | **8** |
| **total** | **20 = `0x14`** |

Exactly the measured `textSizeBytes` delta `0x62830 → 0x6281c`. **My pre-registered `0x10` assumed a 4-byte `beqz`, which was the error.**

**And the second half of my caution is closed too.** I asked whether the *retargeted* branch's own encoding changed, since it now reaches a farther label — two relaxation changes netting to `0x14` would look identical to one. It does not: the old target `.Lbv_singletx` was already ~1925 instructions / ~7700 bytes (`0x1e14`) away from the branch at 15918, so **both the old and new targets exceed B-type range and both are relaxed.** The retarget contributes zero bytes. The entire `0x14` comes from the deletion, with nothing unexplained.

## Note for the record

**A pre-registered size delta must account for branch relaxation.** The technique still earned its place — it forced this analysis, and a discrepancy in the other direction would have been a real defect — but predicting exact instruction bytes is only reliable for straight-line code. Once a long-range branch is involved, the assembler's encoding choice depends on distances the edit itself perturbs, so the prediction should either state a range or call the relaxation out explicitly. Stated as an exact figure it produces a false alarm, which costs review time and, worse, trains people to discount the check.

Everything else stands: identical verdicts on both legs with the 60-row leg re-run against the new base, generated files resolved by **regenerating** rather than picking a side, and the `MtxRuntime:18` docstring updated so a cleanup does not leave a docstring naming a deleted label.

---

Issue #10690, opened 2026-07-27
bug: accumulate multi-transaction coinbase fee balance


pirapira:
BlockVerdictMtxCoinbase.lean:33-43 resolves the coinbase balance from the header on every transaction, adds only the current transaction fee credit, then appends that result through record_nonstorage_effect. CreateCodeEffectLog.lean:281-310 clones the latest AccountState snapshot but overwrites balance unconditionally, so latest-writer-wins can discard earlier coinbase credits.\n\nThe spec requires a running block state: later fee credits must apply to the prior live coinbase balance, not the pre-block header balance. This is the balance twin of the confirmed nonce clobber exposed by Item B.\n\nThe Item B random-200 baseline (seed 1824272313) has six FRs; its named residuals do not immediately identify this as the cause. Candidate explanations to resolve: another path accumulates credits before this writer, the BAL comparison aggregates rather than last-wins, or this contributes to an existing residual population.\n\nCross-link: Item B now preserves its sender nonce correctly but deliberately does not repair this separate balance lifetime defect. Related to evm-asm-g84vz shared-writer/bail correctness work.

pirapira:
**Recording a prediction that was tested and not supported, so it is not left for someone to re-propose.**

When I asked for this to be filed separately, I suggested the defect might **already be contributing to the standing ~6-per-200 false-reject population** — reasoning that if nothing accumulates the coinbase credit, any multi-transaction block with more than one fee-bearing transaction should already reject on the coinbase balance, and such blocks are common. A defect that predicts a population we already observe would have been a much stronger finding than one predicting nothing, and it would have given the standing FR band a named mechanism.

**It does not hold on the evidence available.** The seed-200 residual FR labels are selfdestruct, SSTORE/CREATE, withdrawals and value-to-precompile, with no evident multi-fee coinbase case — and the evm-asm3 agent correctly declined to make an attribution claim on that basis.

So the defect stands as **latent** for now, and the three candidate explanations in the issue body remain open: something else accumulates, the records are consumed by summation rather than last-writer-wins, or the reachable population simply is not in this draw. The first two are answerable by reading; the third would need a targeted selection rather than a random 200.

Worth stating plainly because the failure mode here is asymmetric: a *latent* defect in a shared writer is still worth fixing — the nonce twin of exactly this shape became reachable the moment something upstream started advancing the field, which is what item B did — but it should not be escalated on a severity it has not been shown to have.

---

[...]

PR #10691, opened 2026-07-27
fix: publish sender nonce through AccountState

pirapira:
## Summary

Publish the transaction sender nonce once, at inclusion, through the durable
AccountState overlay.  Later MTx nonce reads now use that execution state
instead of compensating for a sender row in the sorted BAL table.

This removes the `authority == sender` `+1` / `+2` compensation and leaves
`b1_sender_table` as only a sorted distinct-sender enumerator: its count is no
longer an execution-state source.

The shared non-storage writer has one necessary narrow preservation rule.  It
writes a producer's post nonce when the nonce changed or when no authoritative
nonce is cloned; it preserves a cloned authoritative nonce only when the later
producer's nonce is unchanged.  This fixes the sender-is-coinbase last-writer
clobber without dropping the first record's baseline nonce.

## Spec alignment

Amsterdam `fork.py` performs `increment_nonce` at line 1103, before the later
transaction preparation/body path.  Its line 1108 balance debit is a separate
lifecycle with gas and blob charges, and line 1169 applies the additive
`create_ether` refund.  This PR models only the first fact: a single logical
sender nonce in a single durable store.  It intentionally does not couple a
balance snapshot to that store.

## Evidence

Current-main paired Spike A/B, random 200 blocks, seed `1824272313`, same
manifest and instrument:

| ELF | Rows | FA | FR | Agree |
| parent `22c0b682d6f4ad29` | 200/200 | 0 | 6 | 194 |
| candidate `c2e090c67d4f0ecb` | 200/200 | 0 | 6 | 194 |

The comparator passed denominator and many-to-one join soundness checks: 195
joined inputs, zero unmatched, identical six-FR label sets, zero status
differences, zero output-byte differences, and zero new false accepts.

Zero output differences are expected here: this replaces an old compensation
with the same result derived through durable execution state.  They are not
treated as proof that the new path was unexercised.  The random population
contains the controls that moved under rejected variants and they now match the
parent byte-for-byte: `00023 sender_is_coinbase` and blob witnesses `00008`,
`00132`, `00171`, and `00175`.

Variant history on this same seed: coupled balance+nonce was +5 FR;
unconditional nonce publication was +1; a naive `pre_nonce == post_nonce`
conditional was +12 because it skipped a first record's only nonce source; the
refined would-skipping-lose-information predicate is +0.  A complete dynamic
writer-hit count is not exposed by production output; the five controls above
are the demonstrated firing controls.

## Deferred balance work

Balance remains deliberately out of scope.  The rejected coupled path exposed
that blob debit must use the actual blob gas price (`bsg_blob_price_be` /
`amsterdam_blob_gas_price_u256`), not `max_fee_per_blob_gas`; the named blob
witnesses are `00008`, `00132`, `00171`, and `00175`.  Sender-equals-coinbase
also requires fee-credit lifecycle modeling (`00023`).

The separate pre-existing balance last-writer-wins finding is tracked in #10690:
the per-tx coinbase producer derives from header balance each time and can lose
earlier fee credits.  It is not repaired here.

## Checks

- `scripts/check-axioms.sh` (cache-off): green.
- `scripts/check-asm-to-program.sh`: green.
- `scripts/check-region-map.sh`: green.
- `lake build codegen`: green after the rebase and regeneration.

An earlier full `lake build` in this cache-mode worktree hit the known
read-only `.ilean` artifact-cache incompatibility in unrelated modules; no
cache deletion or source workaround was used.  CI remains the clean full-build
gate.

The final regenerated ELF is `.text=0x628a4`, `.data=0x5370`,
`.bss=0x1b2578a0`, SHA-256
`c2e090c67d4f0ecb2893522320cf2af34ef29b63a484850efaad037fdfbf4369`.


pirapira:
Reviewed. **`merge!`** — and this is the best-evidenced convergence PR of the session, because the predicate in it was arrived at by measurement after three refuted alternatives.

## The diff has the right shape for a compensation retirement

Net **negative on logic** in four of five source files — `BlockVerdictMtxRuntime` +28/−44, `TxIntrinsicStateGas` +11/−28, `BlockVerdictMtxTail` +5/−11, `BlockVerdictDataSectionTail` +6/−11 — against `CreateCodeEffectLog` +42/−4 for the new writer and its predicate. A change that removes more than it adds is what retiring a compensation looks like, as opposed to adding a mechanism that happens to work.

**And the removed lines are the substance of #10640 item B**, not incidental:

- the B1 row's nonce increment at both sites (`addi t5, t5, 1; sd t5, 32(t6)`, and the `b1_sender_table_find … addi t0, t0, 1` path);
- the self-sponsored-authorization validation comment;
- and a comment stating the old effect **"intentionally under-reports the sender authority's BAL post"** — retiring a *documented, deliberate* under-report is exactly what that issue asked for.

`bv_b1_sender_table` is demoted to a sorted distinct-sender enumerator, with the new comments stating plainly that AccountState owns the live sender nonce and that no execution path reads or mutates the row's nonce. That is the one-nonce-store outcome, and it leaves no second writer behind.

## The evidence is unusually strong, and the reason is the variant history

| variant | FR (parent 6) |
|---|---|
| coupled balance + nonce | 11 |
| nonce-only, unconditional | 7 |
| nonce-only, naive conditional (`pre != post`) | **18** |
| **refined predicate** | **6** |

Final pair, post-rebase: parent `22c0b682` vs candidate `c2e090c6`, 200/200 both, FA 0 both, FR 6 both, identical six FR labels, 195 joined / 0 unmatched, **zero status and zero output-byte differences**.

**Zero output differences is the correct result here, not a null one** — the change replaces a compensation that was computing the right answer by a wrong route, so byte-identical output is what a faithful convergence produces. And it is not vacuous, because **the five fixtures that regressed under the three earlier variants are in the population** (00023 `sender_is_coinbase`; blob witnesses 00008/00132/00171/00175) and all five now match parent byte-for-byte. That is a control with a demonstrated firing, not a clean sweep.

**Keep the variant history in the body.** The naive conditional is the obvious fix and it is *twelve* false rejects worse than doing nothing; without that record, the next person re-proposes it. The predicate that works — write if the nonce changed **or** no cloned authority exists, preserve only when unchanged and a prior flag already establishes it — is *would skipping this write lose information*, not *did the producer change it*. That distinction is the finding.

## Deferred work, correctly scoped out

The balance half is deferred with its two defects recorded rather than rediscovered: the inclusion debit must use the **actual** blob-gas price (`amsterdam_blob_gas_price_u256`), not `tx.max_fee_per_blob_gas`, with B2's `bsg_blob_price_be` as the correct precedent; and sender-equals-coinbase requires the fee credit to be modelled. Plus #10690 for the balance twin of the clobber fixed here.

Right call. Coupling the balance was my suggestion, from reading `fork.py:1103` and `:1108` as adjacent — and adjacency showed the two writes happen at the same moment, not that they have the same downstream fragility. The nonce is written once and touched by nothing; the balance is touched by value transfer, refund, and the coinbase credit. Three of my proposals here were refuted by measurement, each time by a source-level pin from this lane.

`check-axioms` disclosure (green under `LAKE_ARTIFACT_CACHE=false`) noted and correct to state explicitly.

---

Issue #10692, opened 2026-07-27
FR: sstore_0to0to_x d5-g0/d5-g1 false-rejected at bv_fail=37 by the REVERSE arm (bal_storage_covers_exec_log) — net storage effect is zero and the BAL correctly declares nothing


pirapira:
## Summary

`ported_static/stSStoreTest/sstore_0to0to_x` **`d5-g0`** and **`d5-g1`** are false rejects at `bv_fail=37`, and the rejecting party is **`bal_storage_covers_exec_log`** (the reverse arm) alone. The BAL correctly declares no storage changes, the true net storage effect is zero, and the reverse arm demands a change anyway.

**Not #10614.** That issue's `d5-g0`/`d5-g1` are `sstore_xto_y` at converter indices 20031/20032; these are a different fixture file, and `sstore_0to0to_x` appears **nowhere** in #10614 (0 occurrences across body and all comments). The matching d-index is a per-file position and collides freely across files — #10614's own verified-join section resolves labels through the 26104-entry converter manifest for exactly this reason. **Different population, different arm.**

## Measurement

`--backend spike --filter sstore_0to0to_x --limit 12`: **2 of 12 fail**, both `bv_fail=37`, both `succ guest=00 exp=01` (guest rejects a block the fixture calls valid), `bal_count=11`, and `recomputed_state_root == payload_state_root` on both. The other ten d-indices pass, so the fixture author supplied the control.

### Which arm — both probe arms fired

| probe | result |
|---|---|
| forward `bal_storage_matches_exec_log` forced to succeed | 2 failures **persist** |
| reverse `bal_storage_covers_exec_log` forced to succeed | 2 failures **vanish** — 12/12 full match |

Subtractive on both sides, so neither arm could have been inert. **The reverse arm is the sole cause.**

## Why it looks like a false reject rather than a correct reject

From the fixture itself, comparing the rejecting `d5-g0` against the passing sibling `d6-g0`:

- **Pre and post storage are identical between them.** The only storage in either post state belongs to the system predeploys (`0x…002935`, `0x…7705fa`, `0x…beac02`). **No test-contract storage appears in post at all.**
- So the test contract's net storage effect is **zero** in both, and the BAL declaring **no** `storage_changes` is correct.
- The reverse arm's job is "exec net changes claimed by BAL". It finds a net change in the guest's exec log that the spec says did not happen.

[...]

PR #10693, opened 2026-07-27
fix(r59nm): commit storage_writes on tx status alone, not on the post-preparation coverage point

pirapira:
Fixes a defect I introduced in S2b (#10672): **a failed transaction's storage writes were promoted into the block-level `storage_writes` map.**

## Measured, with a reproducer

On `sstore_0to0to_x` `d5-g0` (the #10692 row), at the point the reverse comparator rejects:

| | |
|---|---|
| `bv_tx_status_arr[0]` | **0** — failed (fixture receipt status `False`) |
| `runtime_tx_auth_phase_applied` | **1** |
| reverted write present under its **full 64-byte key** in the block map | **yes** |
| block map entries | **33** |

## Cause

`write_sets_incorporate_tx` was placed inside `account_state_commit_pending`. The predicate that calls it commits when the tx status is nonzero **or** when `runtime_tx_auth_phase_applied` is set.

**That flag does not mean an authorization was applied.** It marks the dispatcher reaching the **post-preparation coverage point** (`Dispatch.lean`: *"All preparation charges … have now passed"*; `TxIntrinsicStateGas` calls it "the same gas-coverage-derived control"). Nearly every transaction whose body fails after preparation reaches it — the reproducer is a **legacy type-`0x00` transaction with `authorizationList: None`**, and the flag still reads 1.

So the storage merge rode a disjunct that is correct for **AccountState** (a failed body still commits the authorization phase) and wrong for **storage**.

S2b fixed the *opposite* direction of the same gate — the failure path skipping the promotion and so never clearing the tx map — and added `write_sets_discard_tx` for it. **One branch of a two-condition gate was audited.**

## Fix

One storage decision, on **tx status alone**, in the mtx loop before the AccountState gate: incorporate on success, discard on failure. Removed from `account_state_commit_pending`, whose commit predicate is deliberately broader. **Reads stay there and are untouched either way** — same event, opposite treatment, which is why there are two containers.

The spec has no carve-out for storage: `incorporate_tx_into_block` (`state_tracker.py:832`) runs for committing transactions, and a failed one contributes nothing (fresh `TransactionState` per tx, `fork.py:1043`).

## Post-condition — positive, not just "no regression"

After the fix, the same full-key scan at the same point finds the entry **absent**, and the block map holds **0** entries (was 33). That is the check that the fix does the intended thing, rather than merely not breaking anything.

## Gate

Verdicts **identical to base**, since nothing consumes the map:

| leg | result |
|---|---|
| base — spike, `eip7928_block_level_ac`, limit 60 | 2 fail / 58 full / 58 succ |
| candidate — same | **2 fail / 58 full / 58 succ** |

The two `sstore_0to0to_x` rejections **persist by design**: the exec log still drives them and no consumer reads the map. #10692 stays open.

Layout regen: `textSizeBytes` `0x06281c` → `0x062844`; `check-region-map` matches the linked ELF on all five pins; `check-forbidden-tactics`, `check-layering`, `check-unimported`, `check-naming`, `check-file-size` pass.

Refs #10692, #10619, #10680.


pirapira:
Reviewed. **The fix shape is right and the comment is excellent — but I am holding the label on one scenario I could not resolve from the diff, and the argument against it is this PR's own deleted comment.**

## What is right

The storage decision now happens **before** the AccountState gate, on `bv_tx_status_arr[i]` alone — `beqz` to a drop path, `write_sets_incorporate_tx` on success — and is removed from `account_state_commit_pending`, whose predicate is deliberately broader. The AccountState gate itself is untouched, which is correct: the `runtime_tx_auth_phase_applied` disjunct is *legitimate* for authorization effects and only wrong for storage.

The comment is the kind worth keeping: it states the flag does **not** mean an authorization was applied but marks the post-preparation coverage point, names the measurement (`sstore_0to0to_x` d5-g0, status 0, flag 1, reverted write present in the block map), and cites `state_tracker.py:832` plus `fork.py:1043` for why the spec has no such carve-out.

## The scenario I could not resolve

The drop path is an **empty label pair**:

```
beqz t2, .Lbv_mtx_storage_drop
jal ra, write_sets_incorporate_tx; j .Lbv_mtx_storage_done
.Lbv_mtx_storage_drop:
.Lbv_mtx_storage_done:
```

So on failure, nothing happens *here* — no incorporate, and no discard. The tx-level map is cleared by `write_sets_incorporate_tx` itself (`sd zero, 0(s0)`), which now does not run on failure, and by `write_sets_discard_tx`, whose call site I can see is unchanged on the **original failure path further down** — the one reached only when the AccountState gate falls through.

**But that gate has the auth disjunct.** So for a failed transaction with `runtime_tx_auth_phase_applied` set — which by this PR's own reasoning is *nearly every transaction whose body fails after preparation*, and is exactly the population that motivated the fix:

1. the new storage decision drops → no incorporate, no clear;
2. the AccountState gate takes `.Lbv_mtx_code_commit` on the auth disjunct → the original failure path with `write_sets_discard_tx` is **never reached**;
3. so the failed transaction's writes stay **resident in the tx-level map**.

And then the next successful transaction's `write_sets_incorporate_tx` promotes them — **which is precisely what the comment this PR deletes warned about**:

> *otherwise this failed transaction's writes are still resident when the next one starts and get promoted by ITS incorporate.*

If that reading is right, the block map still receives the reverted write, just one transaction later and only in multi-transaction blocks — which would explain a passing single-transaction post-condition perfectly.

## What would settle it

Either point at the clear I have missed, or add `write_sets_discard_tx` on the drop path — the label is already there and empty. Then re-measure the post-condition on a **multi-transaction** block where transaction *i* fails after preparation and transaction *i+1* succeeds and writes storage. The current post-condition is on `sstore_0to0to_x`, `bv_tx_count = 1`, so by construction it cannot exercise the next-transaction promotion.

I may be misreading unchanged context — you have the emitted asm and I am working from a diff. If the clear is covered, say where and I will label immediately.

---

pirapira:
**Withdrawing my hold — the concern was an artifact of my own truncated diff read. `merge!`.**

The drop path is not empty. Read from the PR head:

```
472  beqz t2, .Lbv_mtx_storage_drop
473  jal ra, write_sets_incorporate_tx; j .Lbv_mtx_storage_done
474 .Lbv_mtx_storage_drop:
475  jal ra, write_sets_discard_tx        <-- present
476 .Lbv_mtx_storage_done:
478  ... bnez t2, .Lbv_mtx_code_commit; la t0, runtime_tx_auth_phase_applied; ...
```

**Both arms are handled explicitly, on transaction status alone, before the AccountState gate.** So a failed transaction always discards its tx-level map regardless of the auth disjunct, and the next-transaction promotion scenario I described cannot occur. The gate at `:478` is untouched and still governs code/existence commits with its broader predicate, which is correct.

**My error:** I paged the diff through `head -32`, which cut the hunk at line 474, then treated the absence of line 475 as evidence it did not exist. Reasoning from a truncated view as though it were complete — the same class of mistake as inferring an extent from an entry, which came up twice elsewhere in this work today. The fix for it is trivial and I should have applied it: when a diff read informs a hold, read the file at the head rather than the paged hunk.

**What the PR gets right, restated now that the hold is gone:**

- the storage decision is made **once**, on `bv_tx_status_arr[i]` alone, ahead of a gate whose predicate was chosen for different effects;
- it is removed from `account_state_commit_pending`, so there is no second decision point;
- the AccountState carve-out is left intact, because it is legitimate for authorization effects and wrong only for storage;
- the comment records that `runtime_tx_auth_phase_applied` marks the **post-preparation coverage point** rather than an applied authorization, names the measurement that proved it (`sstore_0to0to_x` d5-g0: status 0, flag 1, reverted write present in the block map), and cites `state_tracker.py:832` and `fork.py:1043` for the spec's absence of any storage carve-out;
- and it ships with a **positive post-condition** — the block map no longer contains the entry — rather than only non-regression, which is the first slice in this chain to have one.

---

PR #10696, opened 2026-07-27
fix(regionmap): restore three dropped storage-writes regions, and gate the invariant that let them vanish

pirapira:
Three RAM regions were in use by the guest with **no `guestRegionMap` entry** — so no pairwise-disjointness proof and no zone-fit proof covered them.

| | on main before this PR |
|---|---|
| `MemoryLayout` constants (`STORAGE_WRITES_AREA`, `TX_STORAGE_WRITES_AREA`, `STORAGE_WRITES_UNDO_AREA`) | **present** |
| `StorageWriteMap` references to those addresses | **6** |
| `RegionMap` region entries | **0** |
| `schemeAAnchors` bases | **0** |
| `stableGuestBases_length` | **26** (S2 raised it to 28, S5a to 29) |

## Where it was lost

Traced over the **full** ancestry (a first-parent-only walk misattributes this — see the correction comment):

| commit | entries | |
|---|---|---|
| `6b670c745` | **1** | Merge PR #10672 (S2) |
| `2c8c3c201` | **0** | Merge PR #10670 into batch — already lost |
| `54eebb21a` | **1** | S5a re-adds the third region |
| `5374a05a1` | **0** | Merge PR #10676 into batch — dropped again |
| `a8d298499` | 0 | Merge PR #10679 — inherited, both parents already 0 |

**Two successive batch-merge resolutions dropped them**: the #10670 batch lost S2's entries, the #10676 batch lost S5a's re-addition. #10679 and #10689 merely carried an already-broken file. The cause is the **batch-merge layout regeneration** treating `RegionMap.lean` as a generated artifact — which is why it happened twice in a row on the same batch.

## And nothing objected — which is the more important half

`check-region-map.sh` compares the **declared** map against the linked ELF. A region that is not declared is simply **not checked**, so the existing gate is structurally blind to this class of loss.

`RegionMap.lean` is also a **mixed** file: generated pins (`textSizeBytes`, `bssSizeBytes`) alongside **hand-written** region entries. The standing rule "regenerate generated layout files rather than pick a side" restores the pins and **silently discards the hand-written half** — which is exactly how #10689 propagated this.

## What this PR does

1. **Restores** the three region entries, the three `schemeAAnchors` bases, and the count at **29**.
2. **Adds `scripts/check-memorylayout-region-coverage.sh`**, asserting the invariant that was violated: every `MemoryLayout` anchor must have a `RegionMap` entry.
3. **Wires it into `check-build-parallel.sh`'s `codegen_checks`** — the CI run step at `.github/workflows/build.yml:99`. An ungated guard is indistinguishable from a passing one.

### The guard was validated in both directions

```
on this tree      : MemoryLayout anchors checked: 20
                    every anchor has a RegionMap entry            exit 0
on main's RegionMap: MISSING from RegionMap
                      STORAGE_WRITES_AREA
                      STORAGE_WRITES_UNDO_AREA
                      TX_STORAGE_WRITES_AREA
                    FAIL (3 uncovered)                            exit 1
```

It fires on the actual regression and names exactly the three lost regions. `STATELESS_WORK_BASE` and the SSZ scratch pair are exempted with reasons in the script — the latter has its own dedicated theorem.

## Gate

**No emitted-code change.** The addresses, the constants and every routine using them are untouched; only the region *declarations* and the anchor count change. `check-region-map.sh` passes on all five pins, the `decide`-backed disjointness and fit theorems build with the three regions restored, and `check-forbidden-tactics` / `check-layering` / `check-naming` pass.

Refs #10679, #10689, #10672, #10676.

### DIFF
--- a/EvmAsm/Codegen/RegionMap.lean
+++ b/EvmAsm/Codegen/RegionMap.lean
@@ -198,7 +198,28 @@ def schemeAAnchors : List GuestRegion :=
     { name := "tx_account_reads_area",  base := 0xa1ea0000, size := 0x80000,   mode := .rw, zone := .ram,
       evidence := "MemoryLayout TX_ACCOUNT_READS_AREA; per-tx account_reads" },
     { name := "tx_code_reads_area",     base := 0xa1f20000, size := 0x80000,   mode := .rw, zone := .ram,
-      evidence := "MemoryLayout TX_CODE_READS_AREA; per-tx code_reads" } ]
+      evidence := "MemoryLayout TX_CODE_READS_AREA; per-tx code_reads" },
+    -- r59nm S2: the WRITE side of the same two levels.  BlockState.storage_writes
+    -- (state_tracker.py:74) and TransactionState.storage_writes (:101).  ONE map per
+    -- level, not one per source: the spec has no system-specific write container --
+    -- process_unchecked_system_transaction (fork.py:782) builds an ordinary
+    -- TransactionState and incorporates at :858, regular txs at :1204, withdrawals
+    -- at :1226.  These replace bv_system_storage_log and bv_user_storage_log, whose
+    -- two-arena split mirrors nothing.
+    { name := "storage_writes_area",    base := 0xa1fa0000, size := 0x200000,  mode := .rw, zone := .ram,
+      evidence := "MemoryLayout STORAGE_WRITES_AREA; 2 MiB = 16384x128 "
+        ++ "(addrHash++slotKey++value, 96 B used of the shared bvStorageLogRowBytes stride); "
+        ++ "block level, filled only by write_sets_incorporate_tx" },
+    { name := "tx_storage_writes_area", base := 0xa21a0000, size := 0x200000,  mode := .rw, zone := .ram,
+      evidence := "MemoryLayout TX_STORAGE_WRITES_AREA; per-tx storage_writes, "
+        ++ "target of storage_write_record (mirrors set_storage, state_tracker.py:489)" },
+    -- r59nm S5a: undo journal standing in for take_snapshot's dict copy
+    -- (state_tracker.py:800-806) under the no-dynamic-allocation constraint --
+    -- a per-frame copy would cost capacity x call depth.  Bounded by the SSTORE
+    -- handler's own 16384-row cap, so it needs no overflow path.
+    { name := "storage_writes_undo_area", base := 0xa23a0000, size := 0x100000, mode := .rw, zone := .ram,
+      evidence := "MemoryLayout STORAGE_WRITES_UNDO_AREA; 1 MiB = 16384x64 "
+        ++ "(entryIndex, wasAbsent, prevValue); reverse-replayed by write_sets_restore_frame" } ]
 
 /-! ## Section / I/O extents (ELF ground truth, `readelf -S`).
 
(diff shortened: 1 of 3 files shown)


pirapira:
Reviewed. **`merge!`** — this fixes live proof-coverage loss while closing the hole that allowed it.

> **Edited to correct an attribution.** My original version of this comment blamed the loss on #10679's merge resolution and drew a reviewer-side lesson from that. **That was wrong** — see the corrected trace below. #10679 and #10689 inherited an already-broken file. The review substance is unchanged.

## Scope verified by absence

`RegionMap.lean` +33/-3, the new guard, and its CI wiring. **No `GuestAddrs`, no `symbol-addresses.tsv`, no `Codegen/Programs`, and no `def … : Nat := 0x8…` movement anywhere in the diff.** So "no emitted-code change" is confirmed by what is *not* there rather than asserted — a stronger bar than identical verdicts for a declaration-only restoration.

## The loss it repairs was real and live

Three RAM regions in use with no pairwise-disjointness proof, no zone-fit proof, and outside the kernel-checked anchor total (`stableGuestBases_length` reading 26 where S2 raised it to 28 and S5a to 29). TCB coverage, not hygiene.

## The cause is the batch-merge layout regeneration, and it fired twice

```
6b670c745  Merge #10672 (S2)            entries 1
2c8c3c201  Merge #10670 into batch      entries 0   <- S2's entries lost here
b467d65e9  Merge #10671 into batch      entries 0
54eebb21a  S5a commit                   entries 1   <- re-added
5374a05a1  Merge #10676 into batch      entries 0   <- dropped again
a8d298499  Merge #10679                 entries 0   <- inherited only
```

**Two successive batch-merge resolutions dropped them**, and the second dropped a re-addition. That is not a contributor mis-resolving a conflict; it is the **batch layout regeneration treating `RegionMap.lean` as a generated artifact** — regenerate the pins, take the incoming copy, and the hand-written region entries in the same file are silently discarded. Which is why it recurred on the same batch and would recur on the next PR touching that file.

**So the process fix belongs on the batching procedure, not on a reviewer's checklist.** `RegionMap.lean` is **mixed** — generated pins (`textSizeBytes`, `bssSizeBytes`) plus hand-written region entries — so "regenerate rather than pick a side" is correct for the pins and destructive for the entries, because regeneration cannot restore hand-written content.

## The guard is the part that outlasts the fix, and it has the line that matters

```bash
(( ${#anchors[@]} > 0 )) || { echo "no anchors parsed from $LAYOUT -- selector is stale" >&2; exit 2; }
```

A grep-based guard whose selector drifts would otherwise **pass vacuously forever**, reporting success over an empty set — the same failure mode this incident instantiates. Exiting 2 is the difference between a check and a decoration. The exemptions are named with reasons (`STATELESS_WORK_BASE` as a base reference, the SSZ pair covered by `sszScratch_matches_layout`) rather than silently excluded.

**And it is wired into `check-build-parallel.sh`, which `.github/workflows/build.yml` actually runs** — which matters here, since #10642 established that of 778 `codegen-*-check.sh` scripts exactly **one** is CI-reachable. An unwired guard would have reproduced the problem it prevents.

**It is also better justified under the corrected cause:** it catches the *outcome* regardless of which step produces it, which is what you want when the producing step is merge tooling rather than a human decision.

## Method note on the misattribution, since it is reusable

The original bisect used `git log -20`, which follows the **first-parent chain only**. That showed #10679 as the first commit with zero entries — but *both its parents already had zero*, so the loss was upstream and the first-parent view could not see where. **On a merge-heavy history, a first-parent walk cannot locate a loss that happened on a side branch.** Check the parents.

---

pirapira:
## Correction: I misattributed the loss to #10679. It was the **batch-merge layout regeneration**, not that PR.

My PR body names `a8d298499` (Merge PR #10679, `docs/sail-equiv-headers`) as where the entries were lost. **That is wrong, and it names another contributor's PR for something it did not do.** Correcting it here and in the body.

My bisection walked `git log -20`, which follows the **first-parent chain only**. That showed #10679 as the first commit with `entries=0`, but both of its parents already had `0` — so the loss was upstream of it and the first-parent view could not see where.

### Traced over the full ancestry

| commit | entries | |
|---|---|---|
| `6b670c745` | **1** | Merge PR #10672 (S2) — entries present |
| `2c8c3c201` | **0** | Merge PR #10670 into batch — **already lost**, though `6b670c745` is an ancestor |
| `b467d65e9` | 0 | Merge PR #10671 into batch |
| `54eebb21a` | **1** | my S5a commit — re-adds the third region |
| `5374a05a1` | **0** | Merge PR #10676 into batch — **dropped again** |
| `730cd85a5` | 0 | `chore(codegen): regen layout for 3-way batch` |
| `a8d298499` | 0 | Merge PR #10679 — **inherited**, both parents already 0 |

So **two successive batch-merge resolutions dropped the region entries**: the #10670 batch lost S2's, and the #10676 batch lost S5a's re-addition. #10679 merely carried an already-broken file, as did #10689.

### What that changes about the remedy

The cause is not a contributor mis-resolving a conflict. It is the **batch-merge layout regeneration** treating `RegionMap.lean` as a generated artifact: regenerate the pins, take the incoming copy, and the hand-written region entries in the same file are silently discarded. That is why it happened **twice in a row on the same batch**, and why it would have happened again on the next PR touching that file.

So the guard in this PR is the right shape and the right place — it catches the outcome regardless of which step produces it — and the process note belongs on the batching procedure rather than on any reviewer's checklist.

I would rather correct an attribution than leave a docs PR blamed for a merge-tooling behaviour.

Issue #10695, opened 2026-07-27
convergence: give balance/nonce/code changes per-transaction attribution (two-level containers, not an index field) — prerequisite for #10680's builder


pirapira:
> **Note:** Written by **Claude Code** (coord agent). Measurements are the scoper agent's. The shape decision recorded below is the maintainer's, 2026-07-27.

**Three of five BAL fields cannot be rebuilt from operations, because the guest does not record which transaction changed them. A fourth will join them when #10680's predecessors retire the exec log.** This is the prerequisite for #10680's serializer and is filed separately so that issue stays scoped to *rebuild and compare one hash*.

## The gap, measured

The spec makes every change an **indexed sequence**: `BalanceChange` is `block_access_index` + `post_balance`, `NonceChange` is `block_access_index` + `new_nonce`, `CodeChange` is `block_access_index` + `new_code`. Confirmed against the fixture JSON — every entry carries `blockAccessIndex`.

The guest keeps **one record per account holding block-final values**:

| BAL field | guest container | shape | index? |
|---|---|---|---|
| `balanceChanges` | `exec_nonstorage_effect_log` | `addr, pre_balance, post_balance, pre_nonce, post_nonce` (112 B) | **none** |
| `nonceChanges` | same record, `post_nonce` | — | **none** |
| `codeChanges` | `exec_code_effect_log` | `addr, has_code_change, code_len, code` | **none** |

Measured absence in the linked asm: `exec_nonstorage_effect_txindex`, `exec_code_effect_txindex`, `account_state_txindex`, `code_state_txindex`, `account_state_pending_txindex`, `nonstorage_txindex` — **all zero occurrences**, and a sweep for any symbol pairing `account_state`/`code_state` with `index`/`txindex`/`bai` returns nothing. Against `exec_log_txindex` at 6 references and `bv_user_storage_txindex` at 3: **the storage side has attribution and the nonstorage side has none.**

**The landed containers are the same shape.** `account_state_pending` is declared at `NonstorageEffectLog:456` as `nonstorageEffectLogCap × 112` and **aliased to `nea_sort_a`** — the same 112-byte record; `account_state_created`/`account_state_delete` are `.set` offsets into that slab. Code is `code_state_pending_count`/`durable_count`/`created_count`/`delete_count` with `code_state_lookup_current`, also one record per account.

[...]

PR #10697, opened 2026-07-27
fix(bmvmx): stamp block_access_index on the contract-recipient path — execution-semantics fix (#10695)

pirapira:
The contract-recipient dispatch path never stamped the transaction's
`block_access_index`, so the only transaction class that can execute SSTORE
tagged every exec-log row with an index it never wrote.

`current_block_access_index` had exactly two store sites program-wide: the
EOA-recipient path and the creation path. The recipient-code-hash comparison
jumps straight to `.Lbv_mtx_is_contract`, past the EOA store, so a contract
recipient reached `dispatch_tx_runtime_code` with the index still at its static
initialiser 1, or at whatever the last EOA/creation transaction left behind. The
SSTORE handler stamps `exec_log_txindex[row]` from that global, and
`exec_log_txindex` is the a4 base of
`bal_all_accounts_tuple_sequences_consistent_skip_list`, whose nonzero return
rejects the block. Measured before this change: 398/400 rows disagreeing with i+1.

Nothing objected because that comparator's per-change loop iterates zero times
(#10681). The unifying consequence, and the reason this blocks rather than
parallels that work: the comparator cannot be made non-vacuous until attribution
is fixed, because the first thing a working comparator does is read a transaction
index no contract transaction ever wrote.

Two stores are added. The contract path gets the missing `i+1` stamp. The
single-tx lane gets an explicit stamp of 1 -- the value the global already holds
from its initialiser, so no behaviour change there, but correctness by initialiser
on a mutable global that the mtx lane writes is the same fragility that produced
this bug. All four dispatch paths now stamp, and the invariant is stated at the
contract-path store.

MEASURED POST-CONDITION, not just an absence of regression. EEST case
`bal_cross_tx_storage_write[tx2_reverts_to_zero]`: two transactions to the same
contract `0x600035600055`, so both take the contract branch, and the fixture's own
declared BAL carries blockAccessIndex 1 and 2 for the written slot. A probe on the
min/max index the SSTORE handler stamps, over the block's two appends, run before
and after with predictions recorded first:

    leg      appends   min   max
    before      2       1     1     <- both transactions tagged 1
    after       2       1     2     <- tx0 -> 1, tx1 -> 2, matching the fixture

The sibling case `tx2_rewrites_same_value` has only ONE append and so cannot move;
the probe carried an append count, which is how that was established rather than
left as an unexplained null. Both legs: full match 2, fail 0.

This was measured LOCALLY with the spike backend. CI verifies nothing about it --
build.yml runs static scripts only, and neither spike nor ziskemu exists in any
workflow.

No check script. An emitted-asm path enumeration would need a codegen artifact at
check time for an invariant unlikely to regress once fixed, and the cause here was
not a missing gate: the comment above this path asserted the stamp happened here
while the next emitted line stamped an unrelated flag, and it also called the
consumers "still-unwired" when two live call sites reject on their result. Both
claims are corrected in place, describing the post-fix behaviour.

---

## Reclassification: this is an execution-semantics fix, not only an attribution fix

**Read this before reviewing.** The description above frames the change around
`exec_log_txindex` and a comparator whose per-change loop is currently vacuous,
which would make it verdict-neutral by construction. That framing is incomplete,
and the part it omits carries a higher risk class.

Auditing the *other* consumers of `current_block_access_index` — asking what this
fix might break rather than what it fixes — found that two of them pass it to
`bal_same_block_delegation_code_resolve` as **`a3`, an index bound**, not as a
flag. `bal_account_code_at_or_before` saves `a3` into `s3` and then skips any code
change whose `blockAccessIndex` exceeds it (`bgtu t0, s3, .Lbcab_next`). So `a3` is
an inclusive upper bound on which same-block code changes are visible.

With the old value, a contract-recipient transaction had `cbai = 1`. So that
resolver accepted only code changes at index ≤ 1: **a transaction at index i could
not see a delegation installed by transactions 1..i-1.** Same-block EIP-7702
delegation visibility was truncated to the first transaction, for every block.

And the blast radius is not the verdict path. That resolver has **17 call sites**,
including `.Lcd_resolve_call_target` and the callcode / delegatecall / staticcall
target resolvers, plus the four `.Lraaw2be_*` target loops — i.e. **runtime
call-target resolution, executing inside the transaction.** This fix therefore
changes which code a nested CALL resolves to on any transaction at index ≥ 2 in a
block where an earlier transaction installed a delegation.

Two consequences for review:

- **The sweep is load-bearing in both directions.** A green paired sweep is now a
  stronger claim than the attribution framing needed. A *non*-green sweep is not
  automatically a regression — it could be cross-transaction delegation cases
  beginning to resolve correctly. Which rows move will be reported before being
  interpreted.
- **What is established and what is not, kept separate.** Established statically,
  each read off the emitted guest: `a3` is an inclusive upper bound; 17 call sites
  including the four runtime call-family resolvers; `cbai` was 1 on contract
  transactions. *Not* established: that any corpus fixture exercises a
  cross-transaction delegation, or that a verdict moves.

One consumer is genuinely inert. At `.Ldtrc_auth_exec_ready`, `cbai - 1` indexes
`bvgr_tx_predelegated_auth_count`, so before this change every contract transaction
wrote **slot 0** — the same per-transaction collapse, in a second place. But that
array has exactly two sites in the emitted guest, the write and the definition:
nothing reads it. Write-only per-transaction scaffolding. Corrected, no behaviour
change.

---

### Independent corroboration from the other lane

A reachability audit on #10685, run by a different agent with a different method,
independently found the single-tx checks apparatus unreachable under current
routing -- agreeing with the 141-reachable / 38-unreachable closure measured here.
Two methods from different directions.

That audit also found the dead single-tx region stages
`bv_pending_upfront_sender_addr` / `_pre` / `_post` / `_nonce`, i.e. the single-tx
path carried sender-nonce and upfront-debit staging that the live multi-tx path
lacked. That is the same family as the gap this PR fixes: **a dead twin holding
something the live path never acquired.** Relevant here because it is the second
instance found today of the live path being the less complete one -- which is the
opposite of the intuition that the universal path is the maintained one.

---

### Outstanding gate

The paired 200-case A/B sweep (same selection both legs, FR label sets reported
rather than counts) has **not** been run. That is the remaining gate before this
should merge: the added stamps change what a live rejecting consumer reads, so a
verdict-neutral result needs demonstrating at scale, not only on the one fixture
above.


pirapira:
Reviewed. **`merge!`** — the gate passes and the change is well-argued. Two source files, +43/−6, no generated-artifact churn.

## The gate

Seed 770695, 200 rows, spike, `--no-verdict-debug`, no `--all`, both legs, manifests diffed first:

| | ran | FAIL | full match |
|---|---|---|---|
| base (`2529275d6`) | 200 | 5 | 195 |
| fix (this PR) | 200 | 5 | 195 |

**Symmetric difference zero in both directions** — 0 repaired, 0 broken, 5 pre-existing on both legs. Five per 200 is inside the accepted band, and per the standing rule those five are not this PR's business: a row failing on both legs is the row that was failing this morning.

## What the change is, and why it is not what the PR originally claimed

`current_block_access_index` had **exactly two store sites in the whole guest**, and the one at `.Lbv_mtx_recipient_lookup_resolved` is on the **EOA-recipient path only** — four `bne` against `chahsr_empty_code_hash` send a contract recipient past it. So for the only transaction class that can execute SSTORE, the index was the static initialiser `1` or whatever the previous EOA/creation transaction left. Measured: **398 mismatches in 400 calls across a 200-transaction block, `cbai` pinned at 1.**

**And the consequential consumer is not the comparator.** `bal_same_block_delegation_code_resolve` passes it as `a3` into `bal_account_code_at_or_before`, which uses it at emitted 21426 as `bgtu t0, s3, skip` — **an inclusive upper bound on visible code changes.** With the value pinned at 1, a transaction at index *i* could not see a delegation installed by transactions 1..*i*−1. That resolver has **17 call sites** including `.Lcd_resolve_call_target` and the `callcode`/`delegatecall`/`staticcall` resolvers, which are **runtime call-target resolution executing inside the transaction.**

So this is an execution-semantics fix, correctly retitled from the attribution framing it started with.

**The spec is unambiguous that the old behaviour was wrong:** `set_delegation` writes through `set_code` into transaction state that incorporates into block state on success, and there is **no at-or-before predicate anywhere** — a later transaction sees earlier delegations unconditionally.

## The evidence, stated for what it is

**Positive observable, predicted before the run:** on `bal_cross_tx_storage_write` / `tx2_reverts_to_zero` — two transactions to the same contract recipient, code confirmed from pre-state — the stamped index reads **min 1 / max 1 before** and **min 1 / max 2 after**, matching the fixture's *own declared* `blockAccessIndex` values of 1 and 2. Spec-anchored rather than arithmetic.

**Non-regression:** the 200-row pair above.

**Not claimed:** a currently-failing case that starts passing. There is none known, and producing one would need a constructed fixture where one transaction installs a delegation and a later one calls it — deprioritised by decision (bead `evm-asm-abqbz`). So this ships on the spec argument plus non-regression, which is stated in the body rather than implied.

**Coverage, measured rather than guessed:** the fix is **provably null on every single-transaction block** (`i = 0` stamps 1, already the initialiser value), so 20219 of 24041 corpus cases could not discriminate. **1007 cases** have a transaction at index ≥1 targeting an account with code — that is the real untested surface, priced at 2014 invocations if anyone wants it closed.

## Also in here

The two `.Lbv_mtx_recipient_lookup_resolved`-adjacent comments that asserted the stamp happened on this path and that consumers were "still unwired" — both false, and the reason this sat — are replaced with the invariant and its four satisfying stores. With no CI guard for it (correctly, per the build-time ruling), that comment is the guard.

And the single-tx lane's explicit `li t1, 1` store is behaviour-neutral but right: relying on a mutable global's static initialiser while documenting that as fragile would have been an exemption from its own rule.

---

PR #10699, opened 2026-07-27
feat(bal): the account_writes map — nonstorage half of #10695

pirapira:
Adds the guest's `account_writes` container pair, mirroring
`BlockState.account_writes` (state_tracker.py:70) and
`TransactionState.account_writes` (:97), with a keyed upsert, the tx->block
merge, discard on transaction failure, and frame rollback via a reverse-replayed
undo journal.

ONE container pair, not three. #10695 is scoped as "balance, nonce and code lack
per-transaction attribution", which reads as three gaps and invites three
containers. The spec's structure says otherwise: both levels keep a single
non-storage write dict, `Dict[Address, Optional[Account]]`, an `Account` carries
nonce/balance/code together, and `update_builder_from_tx` derives all three BAL
fields from one loop over it (block_access_lists.py:637-664). Mirroring the
spec's shape rather than the issue's wording is what makes this one pair.

WHY THE CONTAINER IS THE ATTRIBUTION MECHANISM. In the spec `idx =
builder.block_access_index` is read once per call, and `update_builder_from_tx`
is called once per transaction, before the merge. No record carries an index
field. The transaction is the container, so a change cannot exist unattributed.
#10697 is the evidence for preferring that: the storage side did carry a per-row
index field, stamped from a global one dispatch path never wrote, so every
contract transaction's rows were tagged with an index no transaction had written.
A field is maintained by hand at each append site; a container cannot be
forgotten.

`present` is a field rather than an all-zero-record sentinel: the spec's value
type is `Optional[Account]` and `None` is a distinct state from an account whose
balance, nonce and code hash are all zero. Same reasoning as `wasAbsent` on the
storage side, where zero is a legitimate stored value.

## Inert by construction

Nothing calls these routines. They are emitted anyway, at both the guest site and
the mirrored build unit, because an unreferenced routine is unverified code —
this way the assembler and linker see them. Confirmed present in the ELF symtab:
all six functions and all five counters.

The BAL change-emission is deliberately a separate slice. It needs the spec's
pre-tx baseline — `_get_pre_tx_account` reads the BLOCK-cumulative value falling
back to `pre_state`, not the pre-block value — and the three-way field comparison
whose inequality test is what makes net-zero filtering automatic. It must also be
inserted AHEAD of the merge loop, since comparing after the merge would compare a
value against itself and record nothing. That ordering constraint is documented at
the merge site so the next slice cannot get it wrong silently.

## Layout

Three new 2 MiB RAM arenas laid end to end from 0xa24a0000, contiguous with
STORAGE_WRITES_UNDO_AREA's end, leaving 5.6 MiB below `.data`. RegionMap gains
three region entries and three base-list entries; `stableGuestBases_length` 29 ->
32. Full four-step regen done: textSizeBytes 0x0628cc -> 0x062e54, bssSizeBytes
0x1b2578a0 -> 0x1b2578c0, symbol-addresses.tsv and GuestAddrs.lean regenerated.

`GuestImage.lean` also hard-codes the bss size (and its derived end address) in a
`satWithin_ramRegion` proof, so it needed the same repin — worth knowing for the
next layout change, since it is a hand-written proof carrying a generated
constant and the failure is a type mismatch rather than a drift report.

## Checks

- `check-memorylayout-region-coverage.sh` (from #10696) CAUGHT all three new
  anchors as uncovered before RegionMap was updated — the gate working on a real
  case rather than a synthetic one.
- Structural `#guard`s in the module: arena adjacency and non-overlap, fit below
  `.data`, contiguity with the previous arena, capacity x stride == the reserved
  2 MiB exactly, capacity parity with the storage map, every routine emitted,
  and `discard` clearing exactly what `incorporate` clears.
- Those guards are demonstrably able to fail. One went red for real during
  development: `account_writes_count:` is a substring of
  `tx_account_writes_count:`, so a `splitOn` guard on it counted two hits and
  failed against correct code — now matched as exact lines. And perturbing
  TX_ACCOUNT_WRITES_AREA by 8 bytes made both overlap guards fail, then pass
  again on restore.
- Guards are written fully qualified and on single lines: a `#guard` opened on the
  wrong namespace auto-binds its identifiers as implicits and passes vacuously,
  and one whose expression wraps onto a second line silently covers only the
  first line.
- check-region-map, check-build-units-link, codegen-stateless-link-check,
  check-asm-to-program (389 defs / 148 files) and check-forbidden-tactics all pass.

No EEST sweep: this slice cannot move a verdict, since no code path reaches it.

Stacked on #10697, which the nonstorage emission slice will depend on for a
correct per-transaction index.


pirapira:
Reviewed. **`merge!`** — stacked on #10697, one new module plus its layout, no sweep needed and none asked for.

## The spec settles the container count, and it is ONE — correcting my estimate

I scoped #10695 as "two container pairs" (nonstorage plus code). Verified on the pinned submodule: **both levels keep exactly one non-storage write container** — `account_writes: Dict[Address, Optional[Account]]` at `state_tracker.py:70` (BlockState) and `:97` (TransactionState). An `Account` carries nonce, balance **and** code together, and `update_builder_from_tx` derives all three BAL fields from a **single loop** over that one dict (`block_access_lists.py:637-664`), comparing balance, then nonce, then code hash, emitting only where they differ.

So #10695's "three fields lacking attribution" framing reads as three gaps and invites three containers, and the spec's own structure says one. **That is what tracing before building was for**, and it is also the better reading of why balance and nonce shared the 112-byte record — not a coincidence to preserve but a shadow of the spec's shape.

## And it closes the loop on why #10697 was possible at all

**There is no index field on any record anywhere in the spec's containers.** `idx` is read once per call as `builder.block_access_index`, and `update_builder_from_tx` is called once per transaction — **the transaction is the container.**

The guest's storage side instead carries a per-row index field, which is a discipline maintained by hand at every append site, and **exactly one site forgot** — that is #10697. So the aligned two-level form is not merely tidier: the flat-index alternative's fragility is now demonstrated rather than argued. That is a stronger case than either of us made when the shape was chosen.

## Emitting inert code was the right call, against instinct

Nothing calls this yet. Emitting it anyway is correct for a reason worth keeping: **an unreferenced routine is unverified code — its assembly never reaches an assembler**, which is the same shape as the ungated check script and the unwired guard. And it confirmed all six functions and five counters are in the **ELF symtab** rather than trusting the emitter's exit status, which is the presence-versus-claim distinction that has mattered repeatedly today.

## The ordering trap documented at the merge site

For the change-emission slice: the spec compares against the **block-cumulative** value falling back to `pre_state`, **not** the pre-block value, so `update_builder_from_tx` must run **before** the merge. Emitting after would compare a value against itself and record nothing — **a silent no-op that would look like working net-zero filtering.** Writing that at the site before anyone hits it is worth more than the code around it.

## Guards: fired on real cases, in both directions

- **#10696's region-coverage gate caught all three new anchors** as uncovered before RegionMap was touched. That is the guard we agreed passes the frequency test, firing on a real case rather than a synthetic one — one day after landing.
- **Its own guard went red against correct code**, and the cause is a class that has recurred all day: `account_writes_count:` is a **substring** of `tx_account_writes_count:`, so a `splitOn` guard counted two hits and failed. Fixed to exact-line matching. Then it **perturbed `TX_ACCOUNT_WRITES_AREA` by 8 bytes and confirmed both overlap guards fail, and pass again on restore** — a guard demonstrated capable of failing, not asserted to be.

## Recorded for the next layout change

`GuestImage.lean` **hard-codes the bss size and its derived end address inside a `satWithin_ramRegion` proof**, so a bss repin breaks it — and it surfaces as a **type mismatch in a proof file**, not as a drift report from `check-region-map`. That is the third instance of a hand-written artifact carrying a generated constant, after `RegionMap.lean` being mixed and the batch-merge resolution that lost three regions. Worth a line wherever that hazard is collected.

## No sweep, correctly

Nothing reaches this code, so it cannot move a verdict, and spending invocations to demonstrate that would be exactly the machine time the maintainer called out. Four-step regen complete with `check-region-map`, `check-build-units-link`, `codegen-stateless-link-check`, `check-asm-to-program` (389 defs / 148 files) and `check-forbidden-tactics` all passing.

---

pirapira:

PR #10704, opened 2026-07-27
feat(bal): canonical ordering for the write containers (#10680)

pirapira:
Adapts `mpt_bounded_sort_changes` into `bal_canonical_sort`: in-place MSD radix
sort, nibble digits, explicit bounded stack rather than recursion, capacity
argued rather than assumed (at most 16 ranges per depth, so depth x 16), bounds
checked before every write, and a status return rather than a silent bail.

Two entry points name their own layouts, so the misuse the status codes exist to
catch is unreachable from them: `bal_sort_storage_writes` and
`bal_sort_account_writes`.

## The correction: raw bytes are the wrong sort key

The obvious design — `storage_writes` rows have the address at +0 and the slot at
+32, contiguous in the leading 64 bytes, so sort that prefix flat — is wrong, for
a reason invisible without checking the byte order.

`env.ADDRESS` at `env+0` is stored in EVM stack-word layout: four little-endian
u64 limbs, low limb first. The load-bearing evidence is not the comment saying so
but that the receipt log encoder has to *reverse the 20 address bytes* to obtain a
canonical BE address. `slotKey` is the same. So a lexicographic sort over the raw
prefix orders rows by the limb-swapped representation — a well-defined permutation
that is neither address order nor slot order, while the spec sorts by
`x.address` (`block_access_lists.py:578`) and by `x.slot` (`:564`).

The canonical BE byte at index `b` of a 32-byte low-limb-first field is field byte
`31 - b`. So the digit extraction reverses, and this is one sort but not one
*flat* sort.

Independently corroborated: the field-by-field RLP contract on #10680 specifies
the same mapping from a different direction ("when the address occupies the low 20
bytes of the stack word, reverse those 20 bytes"), with a citation to the runtime's
existing 20-byte reversal.

Incidentally, `addrHash` in `StorageWriteMap` is not a hash — the SSTORE site
passes `x20`, the env base, so the key is `env.ADDRESS`. The name is simply wrong.
Had it been a hash, account-major order would have been unreachable from that
container.

## Both keys take a 20-byte address, so the two orderings agree by construction

The address field is a 20-byte address zero-extended into a 256-bit stack word, so
its upper 12 bytes are padding. Both entry points declare `firstSig = 20`, which
makes storage's account-major order and the account map's order identical **by
construction** rather than by coincidence — the serializer walks accounts once and
must find each account's slots where it expects them. It also skips 24 wasted
depths and avoids sorting on padding that is only zero by convention. A `#guard`
asserts the two widths stay equal.

## No verification of its own, deliberately

The correctness criterion for the BAL is the one execution-specs uses: sort, take
the hash, compare the hash. This routine cannot produce a hash without the
serializer, so its correctness is established end to end when the rebuilt hash is
compared against the declared one — and no intermediate criterion is substituted.

Two were considered and both rejected, which is worth recording because the first
was nearly implemented:

- Sortedness plus permutation-preservation is **insufficient**: both hold for a
  sort on the wrong key, so they would have passed on exactly the limb-swapped
  order described above.
- Comparing our order against a fixture's declared order is the wrong **shape** of
  argument, not merely weaker. It is a statement about the order, and the spec's
  own correctness statement is a hash equality.

And the reason a hash comparison is sound where an inference from it would not be:
we cannot assume keccak is injective. Concluding "our bytes equal theirs" *from*
"our hash equals theirs" needs injectivity, which is not available in a
kernel-checked proof. The way out is congruence, not inference — same input, same
function, same output, no collision assumption.

An order comparison against a fixture's declared BAL stays useful as a
**diagnostic** and is labelled as one in the source: a 32-byte digest cannot say
which account or slot diverged. It localises. It is not the gate.

The `#guard`s are anti-drift checks on the emitted text — that the reversal is
present, that each failure path keeps a distinct status — and the file says
explicitly that they are not a correctness argument about the ordering.

## Inert, and emitted anyway

Nothing consumes the ordering, so no path reaches it and acceptance cannot change.
Emitted at both the guest site and the mirrored build unit regardless, because
unreferenced assembly is unverified assembly.

That paid for itself immediately: the first assembly attempt **failed on a
duplicate label** — `.Lbcs_` is already owned by `BalCodePreimages`, so
`.Lbcs_ret` collided. Renamed to `.Lbalsort_`. Left unwired, that collision would
have sat latent until whoever wired it up next hit it.

All four symbols confirmed present in the ELF symtab rather than trusting the
emitter's exit status.

## Checks

- Guards demonstrated able to fail: diverging the two address widths makes the
  agreement guard go red, and it passes again on restore.
- Full four-step regen: textSizeBytes `0x062e54` -> `0x0630d4`, bssSizeBytes
  `0x1b2578c0` -> `0x1b2678c0`, plus the same repin in `GuestImage.lean`, which
  carries the bss constant inside a `satWithin_ramRegion` proof and fails as a type
  mismatch rather than a drift report.
- check-region-map, check-memorylayout-region-coverage, check-build-units-link,
  codegen-stateless-link-check, check-asm-to-program (389 defs / 148 files) and
  check-forbidden-tactics all pass.

No EEST sweep, and none needed: nothing reaches this code.

Stacked on #10699.


pirapira:
Reviewed, and the digit function is **independently confirmed** — I re-derived it rather than reading it back, because this is the one thing in the PR with no verification standing behind it.

`b < firstSig`: `offset = (firstSig - 1) - b`. With `firstSig = 20`, canonical byte 0 lands at row offset 19 and byte 19 at offset 0. Correct: a 32-byte field held as four little-endian u64 limbs low-limb-first is byte-for-byte the reverse of its big-endian form, so the 20-byte address occupies memory bytes 0-19 with its **most** significant byte at 19.

`b >= firstSig`: `b' = b - firstSig`, `offset = 63 - b'`. The slot field sits at row `+32`, so canonical slot byte `b'` is field byte `31 - b'`, i.e. row offset `32 + (31 - b') = 63 - b'`. Correct.

Even depths taking the high nibble is right too — the more significant nibble has to be compared first.

## The deviation from the digit function I handed you is an improvement, and the record should say so

I gave you `31 - d` for storage under 32 and `95 - d` at or above it, which assumed `firstSig = 32` — the full stack word as the account key. You used **20** and were right to. Two reasons, and the second is the one that matters:

- The upper 12 bytes are padding. The spec sorts `x.address`, and `Address` **is** 20 bytes — two rows agreeing on the low 20 bytes are the same address, so there is nothing for the extra depths to discriminate. Sorting over them is sorting over a convention.
- Both entry points declaring `firstSig = 20` makes storage's account-major order and the account map's order identical **by construction**, which the serializer depends on when it walks accounts once and expects each account's slots where it left them. My version would have made that agreement contingent on the padding actually being zero — true today, and exactly the kind of "true by convention" premise that has cost us repeatedly.

So the two derivations did **not** agree, the disagreement was real, and yours is the correct one. That is the outcome the cross-check existed to produce, and it is worth more than the agreement I was hoping for.

## On shipping with no verification of its own

Correct, and the reasoning in the file header is the part I most wanted to see. Recording the two **rejected** criteria matters more than the one adopted: sortedness-plus-permutation was mine, it would have passed on precisely the limb-swapped order this PR corrects, and without a written record the next person re-proposes it. Naming an order comparison as the wrong *shape* rather than merely weaker is also right — the spec's own correctness statement is a hash equality, and congruence is what makes a hash comparison sound where an inference *from* one would not be.

Labelling the fixture order comparison "diagnostic, not gate" **in the source** rather than in the PR body is the right place for it. PR bodies are not read at the point of temptation.

## Emitting it unwired

Endorsed, and it paid for itself: `.Lbcs_` already belongs to `BalCodePreimages`, so `.Lbcs_ret` collided on the first assembly attempt. Left unemitted, that sits latent until whoever wires it up next inherits it as a mystery. Checking all four symbols in the ELF symtab rather than trusting the emitter's exit status is the same instinct.

**1. Capacity is over-allocated, and that is fine — but the prose and the code disagree.** The docstring says 128 nibble depths (`2 x 64`) while the effective key is `20 + 32 = 52` bytes, so 104. `balSortRangeStackCapacity` computed from the nominal 64 is therefore larger than needed, which is the safe direction. Worth making the prose say 104-effective-of-128-allocated so a later reader tightening it does not conclude the two numbers are the same thing.

**2. The bss growth is trivial here but the budget line is now worth watching.** `0x1b2578c0` -> `0x1b2678c0` is +64 KiB, taking bss to ~434 MiB. Immaterial on its own. But #10680's endgame is materialising rebuilt BALs in RAM to hash them, and the maintainer has already asked directly whether there is room. Nothing to change in this PR — I am flagging it so the serializer slice arrives with a RAM figure rather than discovering one.

## The `GuestImage.lean` comment

Right call, and for the reason you gave rather than the duplication: `check-region-map` would name the drifted constant, but `lake exe codegen` fails first with a bare type mismatch on the `satWithin_ramRegion` application, mentioning neither bss nor a size — so **the diagnostic that would explain the failure is downstream of the failure that stops you.** That is the shape of trap prose fixes and a check script does not, since a guard there would need a codegen artifact at check time and build time is the binding constraint. Hitting it twice in one day is sufficient justification.

**merge!** — approving. On the stack: your read is correct and the resolution for those conflicts is **regenerate, not pick-a-side**, since `textSizeBytes`, `bssSizeBytes`, `GuestAddrs.lean` and `symbol-addresses.tsv` are all generated and either branch's value matches neither tree. Relaying that to merge-bot along with the `GuestImage.lean` repin, which must happen in the same pass and is the one file `check-region-map` cannot warn about.

pirapira:

PR #10706, opened 2026-07-27
feat(keccak): resumable keccak256 — init / absorb / final (#10680)

pirapira:
Every keccak entry point in the tree is one-shot. `zkvm_keccak256` hashes one
buffer; `zkvm_keccak256_segments` hashes a (ptr,len) array's concatenation without
materialising it — but both zero the sponge on entry, so neither can be resumed. A
caller hashing bytes it is *generating* must therefore materialise the whole string
or describe it as a segment array, and both cost memory proportional to the output.

These three entry points remove that constraint. Cost is a 208-byte context,
independent of the length hashed.

General infrastructure, not a BAL detail: nothing here knows about block access
lists, RLP, or any traversal. #10680's rebuild is the first consumer.

## The context is caller-supplied, and that is the load-bearing design decision

`zk3_state` is a single global and there are **16 call sites** of the one-shot
routines in the emitted guest. Had these wrappers shared it, any intervening
one-shot call — a nested hash inside the caller's walk — would zero the sponge and
corrupt an open incremental hash. It would not fault; it would produce a clean,
wrong digest.

So the context is a caller-supplied pointer: 200 bytes of state plus an 8-byte
rate-block fill offset. The wrappers are therefore independent of the one-shot
routines and of each other — an interleaved or nested hash of any kind is harmless
and two incremental hashes can be open at once. Verified rather than asserted: the
emitted wrapper bodies reference `zk3_state` zero times.

**The one-shot routines are not modified, refactored or reimplemented.** They keep
`zk3_state` and their own control flow. This only adds entry points.

## Conventions mirrored, not re-derived

Taken from `MptIndexedTrieRoot`'s `.Lmislh_absorb`, which is working in-tree
streaming code: rate 136, state 200 bytes, byte-wise XOR at `state[fill]` with a
permutation whenever `fill` hits 136, the permutation as `.4byte 0x80052073` with
the state pointer in `a0`, padding `state[fill] ^= 0x01` then `state[135] ^= 0x80`,
digest from the first four u64. Pre-NIST keccak padding, not SHA-3's `0x06`.

## The correctness criterion, as a routine

`keccak_incremental_selftest` hashes the same bytes with the one-shot and with
init/absorb/final and compares digests. It splits the input into **three uneven
chunks** deliberately: the interesting failure is the permutation firing mid-absorb
when a rate block completes, and a single chunk would exercise the same path as the
one-shot and agree vacuously. Callers should pass over 136 bytes for the comparison
to mean anything, and over 272 to cross two boundaries.

This is congruence against already-trusted code, not a fresh correctness argument.
Nothing is inferred *from* the digests matching, so no collision assumption is
involved: they are two computations of one intended function, and a mismatch
localises a defect in the new code.

**The self-test has not been run.** It needs to be called from a live path to
execute, which is a behaviour change rather than an inert addition, so it is
emitted and unreferenced pending approval of a short probe leg. Reporting that
rather than implying the criterion has been met.

## Checks

- Anti-drift guards on the emitted text, explicitly not a correctness argument:
  rate/state/context constants, all four routines emitted, `init` clearing 26
  dwords (25 state + fill — clearing 25 would leave a stale offset and resume
  mid-block), exactly one permutation in absorb and one in final, the pre-NIST pad
  bytes, the fill offset both persisted and reset, three absorb calls in the
  self-test, and `final` NOT re-initialising (an implicit reset would turn a double
  final into the digest of the empty string instead of an observable error).
- Label prefixes and symbol names checked free against the emitted guest before
  wiring — the `.Lbcs_` collision in #10704 cost a build.
- All four symbols confirmed in the ELF symtab.
- `.text` repinned `0x0630d4` -> `0x0632a4`; `.bss` unchanged, so `GuestImage.lean`
  needed no repin this time (these are code-only additions).
- check-region-map, check-memorylayout-region-coverage,
  codegen-stateless-link-check, check-asm-to-program and check-forbidden-tactics
  all pass.

Stacked on #10704.


pirapira:

PR #10712, opened 2026-07-27
fix(bal): per-segment byte direction in the canonical sort key — segments 1 and 2 were unreachable (#10680)

pirapira:
Generalises `bal_canonical_sort` so one sort serves `account_writes`,
`storage_writes` and all four of the BAL builder's event lists. The builder needs
no sort of its own.

## The substance is PER-SEGMENT DIRECTION, not a wider parameter list

The key becomes up to three segments, each a `(byte offset, width)` pair — but the
load-bearing part is that **each segment carries its own byte direction**.

Two storage conventions genuinely coexist in the guest and the sort has to span
both. The write containers hold addresses and slots as EVM stack words — four
little-endian u64 limbs — so those segments must be read reversed. The builder's
rows carry a canonical BE20 address and a canonical BE slot, because the append
canonicalises both; those must be read forward. The `block_access_index` is a
native LE cell and is the only builder segment the sorter reverses.

A single global direction would have silently mis-ordered one field against
another. Bit 7 of each width byte means "already big-endian, index forward":

    LE segment (flag clear)  ->  row offset = k + w - 1 - b
    BE segment (flag set)    ->  row offset = k + b

**Why this class of bug needs a structural answer rather than a test.** Reversing
an already-BE segment produces an order that is total, is a permutation of the
input, and is not canonical. Sortedness does not detect it. Permutation-preservation
does not detect it. It is the third instance in this work of an
internally-consistent-but-wrong ordering — after the original limb-swapped key and
the walk's 17-versus-3 row multiplicity — and in every case the output is sorted,
well-formed and incorrect with no local symptom.

## Canonicalise-on-append is a win beyond ordering

Because the builder canonicalises address and slot as it appends, the stored bytes
**are** the spec's bytes. The serializer copies rather than converts, and the
congruence argument gets shorter: fewer places where a representation change could
diverge from the reference.

## SEGMENTS 1 AND 2 WERE UNREACHABLE BEFORE THIS FIX

Read this before the rest: the extra row sets in this PR are what make the
per-segment key **work**, not what makes it better tested.

The segment walk loaded each segment's width byte with `andi 255`, keeping the
endianness bit, and then compared and subtracted that. A 20-byte big-endian segment
therefore reads as `20 | 0x80` = **148** bytes wide, so no canonical byte index can
ever exceed segment 0's apparent range. **The walk never advanced past segment 0.**
The per-segment key silently degraded to a single-segment sort, and the feature this
PR is named for did not function.

The mask existed — it just sat inside the branch that computes the byte offset,
after the decision that needed it.

Concretely, with rows sharing an address and a slot and differing only in
`block_access_index` at offset 24, the collapsed forward key reaches the index's LOW
byte first and produces tag order 4231 where the canonical order is 1342.

## Validated by a self-test whose expectation is derived independently

`bal_canonical_sort_selftest` sorts four synthetic rows and checks against an
expected permutation computed from the canonical rule alone.

That independence is the most important choice in the test. The obvious check —
walk the sorted rows and confirm each key is `<=` the next — reuses the SAME digit
extraction the sort uses, so if that extraction is wrong both agree and the check
passes vacuously on exactly the order the sort exists to prevent. The rows here
differ only in the address's canonical BE most significant byte, so a byte-index
sort would leave them untouched and the tags would read 1,2,3,4 instead of 2,4,1,3.

Three row sets, because direction is a PER-SEGMENT property and single-segment
evidence cannot establish it:

| set | rows differ only in | exercises | expected tags |
|---|---|---|---|
| A | the address's canonical MSB | segment 0 | 2,4,1,3 |
| B | the slot's canonical MSB | segment 1 | 2,4,1,3 |
| C | `block_access_index` | segment 2, and transaction order | 1,3,4,2 |

**Set C's index values `0x0104, 0x0402, 0x0203, 0x0301` are load-bearing and must not
be simplified.** They are chosen so their low-byte order DISAGREES with their numeric
order. With single-byte values — set C's first version used 4, 2, 1, 3 — a forward
read and a reversed read produce the SAME order, so the set cannot fail: a control
forcing the forward branch went undetected until these values replaced them. A future
tidy-up to small integers silently disables set C and takes segment 2's coverage with
it.

Sets B and C drive the real builder storage descriptor, so the test exercises the
descriptor that ships. C's expected sequence differs from A's and B's on purpose, so
a sub-test reading the wrong arena cannot pass by coincidence.

Run on the guest via a temporary probe leg, since reverted: after the fix, all three
sets pass and set C's arena output is the canonical 1342. Before it, `result=3` with
arena tags 4231.

## The builder's descriptors are pinned as checked constants

`balSortBuilderStorageSegments` and `balSortBuilderEventSegments` record the row
shapes agreed with the agent building those arenas, guarded by DECODING each packed
field rather than restating the literal — a typo in the packing then cannot agree
with itself. Corrupting an endianness bit makes them fail, verified.

The entry points are deliberately NOT added: those arenas do not exist yet, so
referencing them would not link, and inventing an offset would guarantee a silent
mismatch. The constants let both sides be checked in code instead of by comparing
messages.

A guard also caught a real error in my own packed constant before it ran —
`0x2020140000` carried two spurious zero bytes, giving segment 0 a width of zero.

## Layout

Code-only: `.text` `0x0638fc` -> `0x063a34`. `.data`, `.bss`, `MemoryLayout`, region
entries and `GuestImage` untouched. Note the forced regen needs TWO passes —
the first repin feeds back into emission, so `textSizeBytes` must be iterated to a
fixpoint rather than set once.

check-region-map, check-memorylayout-region-coverage, check-asm-to-program
(389 defs / 148 files), check-forbidden-tactics and codegen-stateless-link-check
all pass.

Still inert: nothing consumes the ordering, so no path reaches it and acceptance
cannot change.

Split out of #10710 rather than pushed onto it: #10710 was already labelled
`merge!`, and adding to a labelled branch makes the label assert coverage a
reviewer never gave.


pirapira:
Reviewed, and I re-derived both offset formulas rather than reading them back, since they are the whole change.

`k + w - 1 - b` for an LE segment is right: a `w`-byte field at offset `k` stored least-significant-byte-first puts its most significant byte at `k + w - 1`, so canonical byte `b` sits `b` back from there. `k + b` for a BE segment is right by definition. And even-depth-high-nibble ordering carries over from the original.

Framing the substance as **per-segment direction** rather than a wider parameter list is the correct emphasis. Two storage conventions genuinely coexist — stack-word limbs in the write containers, canonical bytes in the builder rows — and a single global direction would have mis-ordered one field against another. Encoding the flag in bit 7 of the width byte is neat.

## The test discriminates one segment's direction, but the feature is per-segment

This is the one thing I would like closed, and it is small.

The four synthetic rows "differ only in the address's canonical BE most significant byte," which is a genuinely discriminating vector for **segment 0** — a byte-index sort leaves them untouched and the tags read 1,2,3,4 instead of 2,4,1,3. That is the right kind of vector, and the degenerate one-byte-key control shows the harness can fail.

But the flag is **per segment**, and nothing here varies segment 1 or 2. A row set that is correct on the address and wrong on the *slot's* direction — or on the `block_access_index`'s — produces the same pass. Concretely: `balSortBuilderStorageSegments` declares slot BE and `BAI` LE, and if those two flags were swapped in the packing, this test would not notice, because all four rows agree on both fields.

Two cases fix it, and they can be four more rows in the same self-test:

1. rows agreeing on the address and differing only in the slot's canonical MSB — catches segment 1's direction;
2. rows agreeing on address and slot and differing only in `block_access_index` — catches segment 2's direction and simultaneously the property the BAL actually needs, that same-account same-slot events come out in transaction order.

Case 2 is worth having for its own sake, independent of the direction flag: it is the *sequence* property, and it is the one whose failure the whole builder design exists to prevent.

Not blocking — the extraction is shared across segments, so a direction bug in segment 0 and segment 1 are unlikely to be independent defects. But the PR's claim is about per-segment direction, and the evidence currently covers one segment, so I would rather the body say which segments are exercised than have a later reader assume all three.

## The descriptor guard is the best idea in the PR

Guarding `balSortBuilderStorageSegments` and `balSortBuilderEventSegments` **by decoding each packed field rather than restating the literal** is the part I want to single out. A guard that repeats the packed constant agrees with a typo in the packing; a guard that decodes it cannot. That is the same failure the sort's own self-test avoids by deriving its expectation independently — an oracle sharing code with the thing under test proves nothing — applied one level down to a constant. Confirming that corrupting an endianness bit makes them fail is what turns it from a plausible guard into a demonstrated one.

## Canonicalise-on-append

Worth the paragraph it got. The stored bytes being the spec's bytes means the serializer copies rather than converts, which removes a representation change from the congruence chain rather than making one easier to verify. Fewer steps between the guest's bytes and the reference's bytes is the actual goal.

## On the split

#10710 is back to `fcbfc700c`, the encoders commit I reviewed, verified on both the PR and the remote ref, and re-labelled. Cutting this as a stacked PR rather than leaving it on a labelled branch is the right resolution — and worth stating as the general rule, since this is the second occurrence today: once `merge!` is on, that branch is frozen from a reviewer's point of view, because nothing tells merge-bot the diff grew after the label went on.

**merge!** — the two extra self-test cases can land here or as a follow-up, your call; I would prefer here since the harness is already open.

One trivial note: bit 7 of the width byte as the direction flag caps a segment at 127 bytes. All current widths are 4, 8, 20 or 32, so it is unreachable — but a one-line assertion or comment at the packing site costs nothing and makes the constraint visible to whoever next adds a segment.

pirapira:
Re-read the fix rather than labelling on the report, and it is correct in both places it needed to be.

`andi t3, t3, 0x7f` masks the endianness bit **before** the width is compared or subtracted, and the flag is recovered separately from the descriptor with `andi a7, a7, 0x80` rather than being carried in the width value. The total-key-bytes computation masks with `0x7f` too — that second site matters as much as the first, since a 148-wide segment there would have mis-sized the whole key. Both are covered.

The diagnosis in the body is the part worth keeping: *"The mask existed — it just sat inside the branch that computes the byte offset, after the decision that needed it."* A mask present but downstream of its consumer is a harder defect to see than a missing one, because the code contains the thing you would grep for.

## The three sets, and why the shape of the evidence changed

Sets B and C using **distinct** expected sequences on purpose is the detail I would have missed asking for: with a shared expected permutation, a sub-test could pass by matching the wrong set's answer. Returning the failing set number rather than a boolean is the same instinct — a verdict cannot be reconciled against a measurement, a code can.

And the index values `0x0104, 0x0402, 0x0203, 0x0301` are load-bearing rather than arbitrary: their low-byte order deliberately disagrees with their numeric order, so a forward read of a reversed segment produces a *different* order. The first attempt used single-byte values `4, 2, 1, 3`, for which forward and reversed reads agree — the set could not fail and a control forcing forward reads went undetected. That is worth the sentence it got in the body, because a later "simplification" to single bytes would silently disable set C while leaving it looking like a test.

## What I take from this as a reviewer

I labelled this PR before, called per-segment direction its substance, and asked for exactly this coverage as a **non-blocking** nicety. The coverage then showed the feature had never worked. Evidence for one instance of an N-way feature is not evidence for the feature, and the failure is worse than an ordinary gap because the covered instance passes and nothing contradicts it. That question — *does the test discriminate every instance the feature claims* — belongs in the blocking set, not the nice-to-have set, and I had it in the wrong one.

The original test was not weak. It had a firing control, an independently derived expectation, and a descriptor guard that decoded the packed constant rather than restating it. Every one of those good instincts was aimed at segment 0, which worked.

**merge!**

pirapira:

PR #10713, opened 2026-07-27
feat(bal): RLP length helpers for the serializer's measure pass (#10680)

pirapira:
Adds `bal_rlp_scalar_rlp_len` and `bal_rlp_list_header_len`, and extends the
encoder self-test to check that MEASURE AGREES WITH EMIT.

## Why a measure pass exists at all

RLP puts a list's header before its payload, and the header's own size depends on
the payload length. A serializer that streams into a hash — rather than building
bytes in a buffer it can back-fill — must therefore know every payload length
before it emits anything. So the walk is two passes: measure, then emit.

These two helpers are the part of that arithmetic that does NOT depend on any row
layout, so they are correct regardless of how the builder's rows are finally
shaped. That matters right now: the row shapes are still being settled in a
neighbouring lane, and everything shape-dependent belongs in the walk rather than
here.

## The hazard is disagreement, so that is what the test checks

Each helper mirrors an emitter's case analysis. If the two ever diverge, the
measure pass reserves a different byte count than the emit pass writes, and every
subsequent header in the stream is wrong — a corrupt BAL from an arithmetic slip,
with the failure appearing far from its cause.

So the self-test now accumulates the predicted length alongside the emit and
asserts the total. The digest check already establishes that the emitters produced
exactly the reference encoder's 128-byte blob, so the predicted total must be 128
too. A mismatch returns code **2**, distinct from the digest mismatch code 1,
because "our lengths disagree with our bytes" and "our bytes disagree with the
reference" are different defects and should not be conflated.

Both run and control-validated on the guest via a temporary probe leg, since
reverted: result 0 — encoders match the reference AND measure equals emit. Breaking
one case of the scalar length helper (`1 + len` -> `2 + len`) returned code 2, so
the agreement check discriminates rather than passing for free.

## Layout

Code-only: `.text` iterated to a fixpoint. `.data`, `.bss`, `MemoryLayout` and
`GuestImage` untouched. check-region-map, check-memorylayout-region-coverage,
check-asm-to-program (389 defs / 148 files) and check-forbidden-tactics pass.

Split onto its own branch rather than pushed onto #10710 or #10712 — both were
already reviewed, and adding to a reviewed branch makes its label assert coverage
it does not have.


pirapira:
Reviewed. The reasoning for a measure pass is right, the distinct failure codes are right, and the control fired. One real gap in the agreement check, and one question about its form.

## Distinct codes for distinct defects is worth naming

Returning **2** for "our lengths disagree with our bytes" and **1** for "our bytes disagree with the reference" is exactly right, and it is the kind of thing that gets collapsed into a boolean under time pressure. Those are different defects with different causes, and a shared code would have made the first diagnosis of either one ambiguous. This is the same discipline as publishing counts rather than bits.

Keeping the two helpers layout-independent while the row shapes are still settling in the other lane is also the right call — everything shape-dependent belongs in the walk, so this cannot be invalidated by codex landing a different row stride.

## The gap: asserting the TOTAL permits compensating errors

> the self-test now accumulates the predicted length alongside the emit and asserts the total ... so the predicted total must be 128 too

**A sum cannot distinguish a correct set of lengths from a set with offsetting errors.** One case over by one byte and another under by one byte gives a total of 128 and the check passes. That is not hypothetical for this vector: it spans several length classes (`0x81` prefix, two-byte payloads, the 55/56 short-to-long boundary, one/two/three length bytes), so an off-by-one in the short-form branch and a compensating off-by-one in the long-form branch is precisely the shape that survives a total.

Your own control demonstrates the sensitivity you have and the sensitivity you lack: breaking `1 + len` to `2 + len` in **one** case moved the total, so it was caught. Break two cases in opposite directions and it will not be.

**Assert per case, not on the sum.** Compare the predicted length against the actual bytes emitted for each vector entry, and return the failing entry's index the way #10712's self-test returns the failing set number. That is strictly more information for the same run, and it removes the compensation path entirely rather than making it unlikely.

This is the same lesson as #10712 arriving from a different direction: there, evidence for one segment was taken as evidence for a three-segment feature; here, evidence about an aggregate is taken as evidence about sixteen independent computations. In both cases the check is real and its scope is narrower than the claim.

## The question: mirrored case analysis versus one computation in two modes

> Each helper mirrors an emitter's case analysis. If the two ever diverge ...

You have implemented the hazard mitigation as **a test that they agree**. I had asked for the stronger form — one traversal with a measure mode and an emit mode, so a length is *by construction* the count of bytes the same code would emit, and divergence is not expressible.

The concrete shape: give the emitters a destination pointer that may be null, and on null advance the counter without storing. Then there is exactly one case analysis and the measure pass is the emit pass with its writes suppressed.

I am **not** requiring it in this PR, because the emitters are already reviewed on #10710 and restructuring them reaches into a labelled branch. But please price it and say which you would choose, because the difference matters over time rather than today: **your agreement test covers the sixteen cases in the vector, and the divergence risk is borne by cases added later.** Someone extending the emitter for a new field will not necessarily extend the length helper, and the test will keep passing on the old vector. A mode flag makes that impossible; a test makes it detectable only if the vector grows too.

If you keep the mirrored form, add the obligation in the source at both sites — that a new case in either must be added to the other — so the coupling is visible where the edit happens rather than only in this PR body.

## Splitting rather than pushing onto #10710 or #10712

Correct, and thank you for stating the reason. Once a label is on, the branch is frozen from a reviewer's point of view, and there is no mechanism telling merge-bot the diff grew.

## Verdict

The gap above is per-case assertion, which is a small change to a harness that is already open, so I would like it here rather than as a follow-up. Not labelling yet — tell me when the per-case check is in and I will re-read and label. The mode-flag question can be answered in a comment and deferred.

pirapira:
Per-case assertion is in, at both the implementation site and the comment that explains why the aggregate was insufficient. Labelling.

The part that makes this more than a coverage tweak is that the hazard turned out to be **exactly available in this vector, not hypothetical**: exactly three scalars take the `0x80 + len` string path (256, `2^256-1`, `0x01..01`) and exactly three list headers take the long form (56, 300, 70000), so an over-by-one in one path and an under-by-one in the other cancel and the total is 128 either way. Building that control and confirming the old aggregate would have passed it is the right way to settle a gap I raised as a possibility — it converts my "this could hide a defect" into "this did hide a defect class, here is the pair that proves it."

That is the second time today the same shape has surfaced under a different name. On #10712, evidence for one segment stood for a three-segment feature. Here, evidence about an aggregate stood for sixteen independent computations. In both cases the check was real, the control fired, and the *scope* of what it established was narrower than the claim attached to it. Worth stating as a standing question rather than a one-off: **what does this check discriminate, and is that the same set as what the PR claims?**

Distinct codes for distinct defects still stands as the right call — `1` for "our bytes disagree with the reference", `2` for "our lengths disagree with our bytes" — and per-case reporting now names *which* entry, which is strictly more information for the same run.

**merge!**

Still open, and deliberately not blocking: whether the measure pass should be the emit pass with writes suppressed (a null destination that advances the counter without storing) rather than a mirrored case analysis plus an agreement test. By construction beats verified-by-test here, because the agreement test covers the sixteen cases in the vector while the divergence risk is borne by cases added later — someone extending the emitter for a new field will not necessarily extend the length helper, and the test keeps passing on the old vector. Answer it in a comment when convenient; it is a design question for the walk slice, not a defect in this one.

pirapira:
Re-read. Labelling — and the closing observation in the body is the most valuable thing to come out of this work, beyond the PR itself.

## Replacing the tripwire was the right call and the reasoning is exactly right

A guard on the routine count with a comment saying "add the vector case before bumping this" **is** the rule it was meant to replace. Anyone restores green with a one-character edit, and the only thing standing in the way is prose. That is the same class as a `file-size-exception` marker — an enforcement mechanism whose escape hatch is documented next to it.

The replacement has no number to silence: **every measure call must have a per-case assertion, and every assertion must correspond to a measure call** or to one of the two fixed-width address cases. Bidirectional, so adding either half alone fails. Plus length-class pins on 1, 2, 3, 4, 21 and 33 bytes, so a future edit cannot delete the only case exercising a class.

Keeping the count tripwire but **labelling in the body which of the two is a check and which is a tripwire** is the detail I would have missed asking for. Presenting both as protection is how a weaker mechanism inherits a stronger one's credibility.

## Demonstrated firing, three ways

Stripping the assertions from the six header measure calls fires 3 guards; deleting the only 4-byte-header case (the 70000 entry) fires 4; restored fires 0. Running all three rather than reasoning about them is the standard, and the reason it is the standard is on display in this very PR chain: we hit four vacuous checks today and two were fixes for the previous one.

## The stated limit is what makes the guards trustworthy

> a structural parity guard cannot force a new case to be DISCRIMINATING — an entry duplicating an existing length class satisfies parity while testing nothing new

Correct, and saying so converts an unknown limit into a known one. The class pins narrow it — they stop a class being *removed* — but they cannot require an added case to reach a class the vector lacks. So the guards enforce that a new routine is exercised; whether it is exercised *meaningfully* stays a reviewer's judgement.

## The generalisation

> structure can enforce that a check EXISTS and that it RUNS, but not that its DATA discriminates

This is the unifying principle behind everything that went wrong today, and every instance had a check that was present, ran, and passed:

- keccak validated on empty / `abc` / `testing` — all single-block, none reaching the `len mod 136 == 135` boundary where the bug was;
- sort rows differing only in the address's canonical MSB — segment 0, the one segment that worked, while 1 and 2 were unreachable;
- index values `4, 2, 1, 3` — single-byte, so forward and reversed reads agree and the case could not fail;
- measure-vs-emit on the total — three `0x80+len` scalars against three long-form headers, cancelling to 128;
- a probe at an output offset that is always zero — an unreadable null indistinguishable from a passing read;
- a control that tripped a `#guard`, so codegen failed and the run used a stale ELF, exercising the good binary and reporting agreement.

Not one was a missing check. All six were non-discriminating data. I have recorded it as durable guidance, because the reviewer question it implies is the one that caught two real defects in this chain today: **what does this check discriminate, and is that the same set as what the PR claims?**

**merge!**

pirapira:
## Collapsed CI run for the whole BAL-canonical-sort chain

Retargeted from its previous stacked base to `main` so the chain below runs
through remote CI once instead of five times — each PR was stacked on the
previous one, so this branch's head already contains every commit in the
chain; retargeting is the whole collapse. History is NOT squashed; every
constituent commit stays distinct and attributable.

Collapses (in dependency order): #10699, #10704, #10706, #10710, #10712,
and this PR (#10713). Each carries its own review discussion that changed
the code — notably #10699's account_writes map, #10704's `firstSig`-20
correction, #10710's split-off encoders, and #10712's fix for segments 1
and 2 being unreachable in the canonical sort key. All five should
auto-close once this PR merges (their commits become ancestors of `main`);
if any doesn't, it'll be closed manually with a pointer here.

PR #10716, opened 2026-07-27
feat(bal): add inert fieldwise account-write substrate

pirapira:
## Summary

Adds the execution-side, fieldwise `account_writes` substrate for the Amsterdam
BAL rebuild: keyed BE20 rows, component-valid bits, frame rollback checkpoints,
transaction/block helper operations, overflow latches/readers, and persistent
builder arenas. This is a shadow-phase foundation, not a BAL serializer.

### Deliberately unfed

No execution producer is wired to this map yet: neither `record_nonstorage_effect`
nor `create_record_code_effect` calls `account_write_record`. Consequently the
transaction container holds no execution data and nothing serializes from these
arenas. The new helpers and overflow readers are present but inert.

Producer wiring is blocked on the block-level distinct-account capacity decision:
both maps currently share `accountWritesCapacity = 16384` (tx `0xa26a0000`, block
`0xa24a0000`). Before producers are wired, the block map must be raised to at
least 19,047 rows for the cross-transaction plain-transfer route (two distinct
keys per 21,000 regular-gas transaction). The durable route enumeration is in
[#10680](https://github.com/Verified-zkEVM/evm-asm/issues/10680).

### Layout and future constraints

The BSS change reconciles exactly:

`0x6f3d90 reserved - 0x1e02d0 reclaimed = 0x513ac0 grown`.

Final linked sections are `.text=0x63434`, `.data=0x5370`, and
`.bss=0x1b76b380`; the manual GuestImage extent was updated accordingly.

`exec_code_effect_log` contains retiring effect rows *and* non-retiring deployed
code bytes at `record+48`. AccountState/same-block code lookup (and the future
BAL code component) retain pointers into those bytes, so a later log retirement
must preserve or relocate the byte heap first.

The arena tightening deliberately excludes `bv_system_storage_log`: its four
`BlockVerdictStateRoot` callers (566, 615, 709, 725) treat capture failure as
non-fatal, so changing that behavior is a separate soundness question.

## Evidence

- Post-merge shipped artifact:
  `3997ff584331b5997236a9dbf744f23bc3ce6ae99d7ff4af8d2d575fff47a896`
  (`.text` `f07bc124a0bdcded588d468f3203b6262c81854026f8cb1a68b561e632464874`,
  `.data` `26951cd7dacf1d6bf3536c39d02f133d4dc47712bc587ca14bb9c6c447e419e7`).
  It is directly checked by current-ELF region-map and byte-tie gates.
- Spike random-200 on the post-merge artifact, seed `437048345`,
  `--no-verdict-debug`, jobs=10: 200 scored,
  196 full matches, 4 false rejects, 0 errors, 0 budgets. This is a single
  candidate leg, not a claimed A/B transition; the four labels are two
  `blob_gas_subtraction_tx` variants, `sstore_xto0to0 d6-g1`, and type-4
  `transaction_validity` multiple-auth. The expected unchanged population rests
  on the overflow paths being unreachable at current/post-tightening bounds.
- `scripts/check-asm-to-program.sh` passed.
- `NO_BUILD=1 ELF_DIR=gen-out/regionmap scripts/check-region-map.sh` passed
  against the measured current ELF.
- `lake build` passed; `git diff --check` was clean before commit.
- `check-axioms.sh` was not runnable locally: cache-mode Lake did not
  materialize `MemoryRegionStackSpec.olean` for `lake env lean` (GH #10537).

No self-approval requested.


pirapira:
Read in full. The body carries everything I asked for and two things I did not, and I want to name those before the two items I am holding on.

## What is right, and why each matters

**The unfed declaration is unambiguous** — "neither `record_nonstorage_effect` nor `create_record_code_effect` calls `account_write_record`", so the transaction container holds no execution data and nothing serializes. That sentence exists because three of today's findings were correct-looking mechanisms nobody wired (a container with zero call sites, three overflow flags with no reader, `account_writes_restore_frame` with no checkpoint symbol), and each was invisible precisely because the code read as finished. A PR adding another dormant structure owes this, and it is the difference between a documented stage and a fourth instance.

**The BSS reconciliation is stated as an equation** — `0x6f3d90 reserved − 0x1e02d0 reclaimed = 0x513ac0 grown` — which I verified: 7,290,256 − 1,966,800 = 5,323,456. That form is what shows *both* commits are accounted for and nothing else moved. Two independent figures would not.

**The regen is shown to be a fixed point**, not merely to have run: an independent post-`GuestAddrs` link and canonical `gen-out` agree byte-for-byte on `.text` and `.data`. That is the stronger claim and it is the one worth making.

**The gate is reported at its honest ceiling** — "a single candidate leg, not a claimed A/B transition", with the seed, the four labels named, and the expected-unchanged population resting on a stated derivation (overflow paths unreachable at current and post-tightening bounds) rather than on the count looking normal.

Two things I did not ask for and am glad to see: the `check-axioms.sh` limitation **volunteered** with its cause and issue reference (GH #10537, cache-mode Lake not materialising `MemoryRegionStackSpec.olean`), rather than omitted; and the `bv_system_storage_log` exclusion recorded as deliberate with the four caller sites, so the omission cannot later read as an oversight and get "fixed" without confronting the non-fatal policy.

## Holding on two items

**1. `.text` moved in the wrong direction and the delta is unexplained.** An earlier report of this branch had `.text = 0x633fc`; the PR states `0x63424`. That is **+40 bytes**, and the change in between was *removing* the producer calls from `NonstorageEffectLog` and `CreateCodeEffectLog` — a removal. A removal that grows `.text` by 40 bytes needs an account.

Most likely benign: other edits landed in the same window (the fieldwise overlay, the overflow readers at both sites) and the net is positive. But "probably fine" is what a byte-delta reconciliation exists to replace. This is the technique that closed #10707's sixteen bytes exactly and that caught scoper's stray 120 bytes from a revert it believed had not landed — and in that second case the surprise byte count was the *only* signal that the tree contained something the author did not think was there.

Enumerate what changed between the two measurements and show the net reaching +40. If it does not close, that is the finding.

**2. `check-forbidden-tactics.sh` is not in the evidence list.** `check-axioms.sh` being unrunnable in cache mode is disclosed and accepted — but the two gates are deliberately complementary: the axiom check is kernel truth, and the forbidden-tactics scan is a fast **source** scan that needs no build at all. With the kernel-truth gate unavailable on ~2,648 added lines of Lean, the source scan is the only remaining guard against a `bv_decide` or `native_decide` slipping in, and it costs nothing to run.

Please run it and add the result. If it has already run and was simply omitted from the list, say so and that closes it.

## Not a blocker, for the record

The capacity precondition is now **resolved** rather than open — the block map needs at least 19,047 rows, from the plain-transfer route at 21,000 regular gas per transaction yielding two distinct keys, sourced to `transactions.py:686-704`/`:736` with `gas.py:70` and `:133-134`. I have updated the #10680 enumeration accordingly, so the link from this body points at a closed figure rather than two candidates.

Worth restating why that precondition is load-bearing: wiring producers against a 16,384-row block map would put a fail-closed overflow **inside** the range of blocks the gas limit admits — a self-inflicted no-spec-counterpart bail, which is the debt `g84vz` exists to remove. Stopping short of the wiring was the right call, not a stalled slice.

Not labelling yet. Both items above are quick; tell me and I will re-read and label.

pirapira:
**Both holds withdrawn — labelled `merge!`.**

On `check-forbidden-tactics.sh`: **CI enforces it**, so asking for a local run was my error. `CLAUDE.md` documents the two complementary gates explicitly, and there is a standing convention that CI covers gates so review effort belongs elsewhere. Nothing for the author to do.

On the `+40` byte `.text` delta: minor, and not worth the reconciliation I asked for. The measurements straddled several edits — the fieldwise overlay, the overflow readers, then the producer-call removal — so a positive net is expected rather than surprising, and neither figure was pre-registered as a prediction. Pre-registering a delta is worth it when a change claims byte-level equivalence (#10707's sixteen bytes) or when an unexplained count is the only signal available (scoper's stray 120 bytes from a revert believed not to have landed). Neither applies here.

Everything in the review above about what the body gets right stands, and the capacity precondition remains the gating item for the slice that wires producers: the block map to at least 19,047 rows before `record_nonstorage_effect` or `create_record_code_effect` call `account_write_record`.

Issue #10717, opened 2026-07-27
account_writes cannot be the sole BAL account source while sender nonce and balance are deliberately routed around it


pirapira:
## Summary

The `account_writes` container cannot become the sole source for the BAL's `balance_changes` and `nonce_changes` while **sender effects are deliberately routed around it**. Two of the three effects that survive a failed transaction never enter the container, and they are the most common BAL account entries there are — every transaction has a sender with both a nonce change and a balance change.

Audited by @evm-asm3-codex while wiring the two producers for #10680; filed here rather than as a bead because its checkout has no beads database. Complements bead `evm-asm-tdbn0` (sender-nonce staging), which covers a narrower slice of the same problem.

## The three surviving effects, and where each goes

A failed transaction still increments the sender's nonce, debits the sender for gas, and credits the coinbase the fee. Only one of the three reaches the container:

| effect | path | in `account_writes`? |
|---|---|---|
| sender **nonce** increment | `account_state_publish_sender_inclusion`, `BlockVerdictMtxRuntime:293` — runs *before* the runtime snapshot at `:397`, documented as durable-only | **No** |
| sender **balance** gas debit | `BlockVerdictMtxTail:168-189` — senders are **explicitly excluded** from the all-accounts nonstorage log; B2.3's separate running-balance table verifies them instead | **No** |
| **coinbase** fee credit | `record_nonstorage_effect` via `BlockVerdictMtxCoinbase:43` | Yes |

## Why this is architectural rather than a wiring task

The exclusion at `MtxTail:168-189` is **deliberate** — senders are kept out of the nonstorage log because a different mechanism verifies them. So unifying them is not adding a producer call; it means reconciling with an existing design that intentionally keeps them apart.

And the mechanism that owns senders, **B2.3's running-balance table, is itself one of the comparators inside the exec-log-versus-declared-BAL apparatus that #10680 intends to retire** (see the 733-line sizing on that issue). Those two facts point the same way: the container cannot be the single source for account entries while a parallel mechanism owns the most common entries, and that parallel mechanism is scheduled for removal.

## Why the omission is easy to under-rate

The adversarial bound derived for #10680 is **dominated by exactly these effects**. The binding route is transaction sidecars — three balance events plus one nonce per transaction — and the sender accounts for two of those four. So the account side of the BAL is not nearly-covered-with-a-follow-up; with body effects and the coinbase wired and senders excluded, **it misses the majority of real entries.**

[...]

PR #10719, opened 2026-07-27
feat(bal): wire account write producers

pirapira:
Stacks on #10716.

This adds the first producer wiring for the account_writes attribution map after its block capacity was raised to 20480 rows while retaining the 16384-row transaction map.

The existing record_nonstorage_effect and create_record_code_effect paths now dual-record into the transaction map. Failed contract and creation routes restore the account-write undo mark at the existing MTx body rollback boundary. The post-body coinbase effect is then recorded and the surviving transaction map is incorporated once. No BAL reader or serializer changes in this PR.

Container state: fed but unread. This now records body nonstorage/code effects and coinbase, but it deliberately excludes sender nonce and sender gas balance: nonce is published through account_state_publish_sender_inclusion and balance stays in B2.3. Senders are the majority of real BAL account entries, so this map is not a BAL sole source yet. The convergence work is tracked by evm-asm-tdbn0 and GH #10717.

Validation:
- Fresh linked artifact converged across relinks: text 0x634c8, data 0x5370, bss 0x1b76b3a0; first and final links had identical text/data bytes.
- check-asm-to-program and check-region-map passed; scoped source build passed.
- Spike random 200 production-ELF gate, auto seed 260324421: 200 total, 193 full, 7 false rejects, 0 errors, 0 budgets, 200 root matches. This gate is a collateral-damage check for the live producer and rollback paths, not evidence that the unread map changes verdicts; a moved failing set would be unexpected and require investigation.
- check-axioms is not runnable locally in LAKE_ARTIFACT_CACHE mode because cache-satisfied oleans are absent from LEAN_PATH; it is intentionally not claimed.


pirapira:
Read. The scope statement, the lifecycle description and the convergence evidence are right, and the coverage boundary is stated where it needs to be — in `AccountWriteMap` as well as here, which is the version that survives.

Two things before I label, one of them substantive.

## "7 pre-existing false rejects" claims more than one leg can establish

The gate reports **7 false rejects at seed `260324421`**, described as *pre-existing*. That word is doing real work and a single leg at a fresh seed cannot support it:

- earlier legs on this stack reported **4**, at different seeds — so 4 against 7 compares different scored populations and says nothing;
- **7 also sits at or just past the ~5–6 band**, which makes it the one number in this PR a reader would look at twice;
- and this slice **modifies live code paths** — both producers, plus `CallFrameReturn`'s status-zero routes restoring the account-write undo mark, which fires on every failed frame. So unlike #10716, a moved failing set here is not automatically churn.

There is a legitimate basis available and it is cheap: **name the 7 labels and confirm each was observed in an earlier leg.** Label identity survives a change of seed in a way counts do not, so if all 7 are previously-seen failures then "pre-existing" is established and the sentence stands as written. If any label is new, that is worth knowing regardless of the count — a new label on a PR that touches the failed-frame path is exactly the signal the gate exists for.

Either outcome is fine; what I do not want is the word carrying an implication the evidence does not. If you would rather not spend the comparison, downgrade it to "7 false rejects, in band, not investigated" — which is the honest ceiling for one leg and is what #10716's body correctly said.

## Please link #10717

The body points at `evm-asm-tdbn0` and #10680 for sender convergence. **#10717** is the issue I filed for that boundary while you were committing — it records that sender nonce goes through `account_state_publish_sender_inclusion` at `MtxRuntime:293`, that sender balance is *deliberately excluded* at `MtxTail:168-189` with B2.3's running-balance table verifying it instead, that only the coinbase reaches the container via `BlockVerdictMtxCoinbase:43`, and that unifying senders means reconciling with a comparator that #10680 intends to retire rather than adding a producer call.

That is the issue someone planning this work will find; the bead is narrower and `#10680` is now a long thread.

## Worth stating: the container is fed but unread

Not a defect, and the body implies it without saying it. #10716 shipped it unfed-and-unread; this ships it **fed-and-unread**, which is a different state and a more misleading one — a reader can reasonably assume a fed container is a working one. It holds body effects and the coinbase, misses the majority of real account entries, and nothing consumes it. One sentence in those terms closes the gap between what the PR does and what a skim suggests.

## On the gate's purpose

Also worth a line, because it is not obvious: nothing reads `account_writes`, so the recording cannot move a verdict, and the gate here is **checking for collateral damage on live paths** rather than validating new behaviour. Without that, an inert-container PR carrying a 200-run reads as ceremony.

`check-axioms` not being runnable in cache mode is fine and correctly not claimed — CI enforces the complementary source-side scan.

Holding the label on the first item only; the rest are body edits.

pirapira:
**Labelled `merge!`.** The hold is discharged: "pre-existing" is gone and the line now reports 7 false rejects with the gate's purpose stated, which is the honest ceiling for a single leg. Dropping the claim rather than spending a comparison to defend it was the right of the two options.

Adding *a moved failing set would be unexpected and require investigation* is better than what I asked for. It records the reasoning that makes this gate different from #10716's — there, nothing new was reachable; here, the undo-mark restore fires on every failed frame — so a future reader knows which direction a surprise should be read in.

One residual, stated so it is known rather than assumed: **whether 7 represents a move is unresolved, by design.** The ~5–6 band was measured on other seeds, so 7 against it compares different scored populations, and under the standing one-leg rule that comparison is not bought. If anyone later wants it settled, the instrument is **label identity, not counts** — labels survive a change of seed in a way totals do not, so naming the 7 and matching them against a previous leg would answer it in one pass without a base run.

The source-side note is the part I most wanted and it landed: `AccountWriteMap` now names #10717 alongside `tdbn0` and says **fed but unread** with the majority sender coverage missing. That sentence is the one that will still be true in six months, and it is the difference between the next person reading a fed container as working and reading it as staged.

pirapira:
**Re-labelled `merge!`.** Verified the fix on the branch rather than from the report: the three `ACCOUNT_WRITES` region entries are present at `RegionMap.lean:234/239/246`, the three anchors are in the `schemeA_matches_layout` list at `:641-643`, and `stableGuestBases_length` is pinned at 32 (`:872`).

## The gate was protecting more than a missing table row

Worth recording, because it changes what this fix was: `check-memorylayout-region-coverage` greps `RegionMap` for `EvmAsm.Stateless.<ANCHOR>`, and **the region entries do not contain that** — they carry literal bases and name the anchor only in prose. What satisfies the grep is `schemeA_matches_layout`, a `decide`-checked theorem asserting that `schemeAAnchors.map GuestRegion.base` equals the `MemoryLayout` constants **in order**.

So each arena needed an entry *and* an anchor line, and three entries alone would have left the gate red. The theorem is what makes each base a **verified claim rather than a copied number** — which is the difference between the arena being described and the arena being pinned.

## What these three arenas had been missing

They now fall under `guestRegionMap_fits_ram`, `guestRegionMap_pairwise_disjoint` and `schemeAAnchors_pairwise_disjoint`, all `decide`-checked and all passing. **Before this they were outside all three** — no RAM-fit proof and no disjointness proof, for arenas totalling 6.5 MiB.

That is the same state three *storage* regions were in after a merge resolution of #10679, recorded in the comment above the length counter. The only difference is that a gate caught it this time instead of it being noticed later. So the missing entry was not cosmetic: it was 6.5 MiB of `.bss` with no proof it fits or that it overlaps nothing.

## On the one thing touched beyond the three entries

`stableGuestBases_length` going 29 → 32 is **in scope**, and the convention supports it: its own comment history shows 20→23, 23→26, 26→29, each bump annotated with what was added. It is the counter the entries themselves require — not `MemoryLayout`, not an arena extent, not `GuestImage`. Correct to flag it rather than assume, and correct to judge it in-scope.

## Addresses read rather than reconstructed

The arenas had moved — codex relocated tx to `0xa2720000` and undo to `0xa2920000` and raised the block map to 20480 rows while tx stayed at 16384 — and contiguity was **checked rather than assumed**: `0xa24a0000 + 0x280000 = 0xa2720000`, `+ 0x200000 = 0xa2920000`, ending `0xa2b20000`, picking up exactly where `storage_writes_undo_area` ends. No gap, no overlap.

That check is the reason this landed clean. Reconstructing the addresses from #10699's original layout would have produced three entries that were internally consistent, satisfied the grep, and pinned the wrong bases — with `schemeA_matches_layout` then failing on `decide` and the failure pointing at the theorem rather than at the stale addresses.

pirapira:
Dropped from the local batch-merge run (batching against latest main).

**Blocker: semantic conflict with main, not a mechanical one.** This branch splits `accountWritesCapacity` into `txAccountWritesCapacity` (16384) and `blockAccountWritesCapacity` (20480) in `EvmAsm/Codegen/Programs/AccountWriteMap.lean` (commit `33d11b199`, "size block account-write map independently"). Meanwhile main gained `EvmAsm/Codegen/Programs/BalCanonicalSort.lean` (#10680 line), which references the old `accountWritesCapacity` at line 213:

```
"  li t0, " ++ toString accountWritesCapacity ++ "; bgtu a1, t0, .Lbalsort_over_capacity\n" ++
```

That file does not exist on this branch, so nothing here reconciles the rename. After merging main the build fails with `Unknown identifier `accountWritesCapacity``.

Deciding **which** bound the generic sort should enforce is a semantic call an automated merge shouldn't make: the routine is used for both write containers, so `blockAccountWritesCapacity` (20480) is the only value that doesn't reject a valid block-level sort — but that changes the routine's accepted range and its emitted bytes, and no proof pins it today (the sort is still inert). Leaving it to a human.

Also note the RegionMap divergence this branch carries is fine on its own — it is a descendant of #10716, so its `account_writes_area` = 0x280000 / rebased `tx_`+`undo_` entries deliberately supersede what landed via #10721.

Everything else in the queue was drained: #10708 + #10716 landed via #10721, and #10723 is updated and set to auto-merge.

pirapira:
## ⚠️ Merge hazard: the three `AccountWriteMap` anchors are authored on TWO branches independently

**Whoever resolves this merge: take ONE side's entries, do not union them.**

`RegionMap.schemeAAnchors` gained the same three anchors twice, in two separate authorships:

- **#10719** — scoper restored them (they had been *deleted* rather than updated when the arenas moved), with the length pin back from 29 to 32.
- **#10722** — codex added them independently at commit `0861a7c35`, with the same length-pin change.

The two are **semantically identical**: `account_writes_area` `0xa24a0000`/`0x280000`, `tx_account_writes_area` `0xa2720000`/`0x200000`, `account_writes_undo_area` `0xa2920000`/`0x200000`. Both were derived independently and agree, and contiguity was verified on one side: `0xa24a0000 + 0x280000 = 0xa2720000`, `+ 0x200000 = 0xa2920000`, picking up exactly where `storage_writes_undo_area` ends.

### Why a naive union breaks the build in a way that will not point at itself

`schemeA_matches_layout` is a **`decide`-checked theorem** asserting that `schemeAAnchors.map GuestRegion.base` equals the `MemoryLayout` constants **in order**, and `stableGuestBases_length` is a **`decide`-pinned counter**. A union merge produces **six** entries where three belong, so:

- the length pin fails at 35 against a pinned 32, and
- the ordered-equality theorem fails,

and both surface as `decide` proved-false errors that **name neither the duplication nor the file that caused it**. That is the same failure shape that made someone delete these three entries in the first place rather than update them: the error does not tell you what is wrong, and deleting makes the build pass.

### Resolution

Take either side's three entries — they are equivalent — keep exactly three, and leave `stableGuestBases_length` at **32**. Then confirm `check-memorylayout-region-coverage` and `check-region-map` both pass, and that `bnf_le_a` is still `2734690016` (`0xa3000ee0`), since a `.data` shift would be an unrelated symptom appearing at the same moment.

Dependency order is #10719 → #10720 → #10722, so #10719's copy reaches main first and #10722's becomes the conflicting side.

Recorded here rather than relayed, because the merge lane no longer shares a message channel with the rest of the fleet — so anything not in the repo or on a PR does not reach it.

## ✅ RESOLVED — this hazard no longer exists. Disregard the warning above.

The duplicate authorship was **unpicked at source** rather than left for the merge:

- **#10720** = `eec0b964a` — scoper's #10719 merged in
- **#10722** = `d31ac2b4d` — codex's duplicate `0861a7c35` **reverted**, then #10720 merged
- **#10724** = `9aaa4b739` — restacked on `d31ac2b4d`

**The final tree contains exactly scoper's three entries with `stableGuestBases_length = 32`, and no duplicate `RegionMap` source.** Coverage, `check-region-map` and `check-asm-to-program` all pass at both parent and child layouts; the child stays at `.text 0x6383c`, `.data 0x5370`, `.bss 0x1b76b3a0`. The only residue is history — a revert plus a canonical merge — with no semantic divergence.

So **there is nothing to reconcile at merge time.** Do not look for two copies, and do not union anything.

### Why this correction matters more than the original warning

I had advised *leaving* the duplication, on the grounds that two identical edits resolve trivially and unpicking would cost two full three-pass regen convergences for single authorship of nine uncontested lines. That cost reasoning was sound, and the work was already pushed before the advice landed — so the outcome is **better than what I settled for**: one authorship, no conflict, no resolver judgement required at all.

The warning above is now the more dangerous artifact. A resolver reading it would go looking for a duplication that is not there, and the natural response to "take one side, do not union" when both sides look identical is to start second-guessing a clean merge. **A stale hazard notice is worse than no notice**, which is the same failure mode as `check-region-map` pointing at a script that could not perform the repair — the guidance defect that misled three authors today and that #10720 fixed.

Left in place rather than deleted, with this correction attached, because the reasoning about *why* a union breaks — six entries against a `decide`-pinned count of 32, and an ordered-equality theorem failing with an error that names neither the duplication nor the file — remains true and worth having for the next person who moves an arena.

pirapira:
Superseded: #10724's head contains this branch's commits (`33d11b199`, `aaa3a0d23`, `3784fdd32`, `6afef40c3`) as ancestors, and it also carries the reconciliation that was blocking here — `BalCanonicalSort.lean:213` now reads `blockAccountWritesCapacity` after the split. Landing via the batch PR #10729; removing `complicated-merge`.

pirapira:

PR #10720, opened 2026-07-27
fix(scripts): describe complete layout drift repair

pirapira:
Stacks on #10719.

Credit: scoper identified that check-region-map named gen-symbol-addresses.py --build as the drift remedy even though that generator only writes the symbol TSV. It never writes RegionMap.lean, and it cannot repair GuestImage.lean literal drift.

This changes only the check-region-map diagnostic and its top-level contract. On layout drift it now names the actual manual fixed-point procedure: relink and regenerate TSV, GuestAddrs, and GuestImageEntries; pin RegionMap text/data/bss sizes plus GuestImage guestScratch_sat bss-end literals from that ELF; relink and repeat until stable. It points to docs/regenerating-generated-files.md for the ordered regeneration commands.

No emitted source, layout artifact, regeneration output, or guest verdict changes. Validation: bash -n scripts/check-region-map.sh and git diff --check.

Automation assessment: practical as a separate slice. A future helper can read ELF sections and rewrite the three anchored RegionMap definitions and two GuestImage bss endpoint literals, then run a bounded convergence loop. It should not be folded into this error-message repair because it must handle the bootstrap case where stale GuestImage blocks relinking, and source-rewriting proof literals needs explicit, reviewable failure behavior.

### DIFF
--- a/docs/regenerating-generated-files.md
+++ b/docs/regenerating-generated-files.md
@@ -11,7 +11,7 @@ The four files, what regenerates each, and what it depends on:
 |---|---|---|
 | `scripts/asm-fixtures/symbol-addresses.tsv` | `scripts/gen-symbol-addresses.py --build` | the linked ELF (`lake exe codegen`) |
 | `EvmAsm/Codegen/GuestAddrs.lean` | `python3 scripts/asm_to_program.py guest-addrs` | the TSV above |
-| `EvmAsm/Codegen/RegionMap.lean` (the `textSizeBytes` / `dataSizeBytes` fields) | manual edit to the ELF section sizes | the linked ELF |
+| `EvmAsm/Codegen/RegionMap.lean` (the `textSizeBytes` / `dataSizeBytes` / `bssSizeBytes` fields) | manual edit to the ELF section sizes | the linked ELF |
 | `EvmAsm/Codegen/Proofs/GuestImageEntries.lean` | `python3 scripts/guest_image_coverage.py --emit-lean` | `GuestAddrs` + the linked function set |
 
 ## Prerequisites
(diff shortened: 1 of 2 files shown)


pirapira:
Good fix, and it repairs the defect at both sites — the header comment and the runtime `echo` — rather than just the one a reader sees on failure. Two details make it right rather than merely different:

- **it says what `gen-symbol-addresses.py` actually does** ("writes only the TSV; it does NOT update RegionMap or GuestImage"), which is the fact whose absence caused the problem;
- **it says "repeat until those values stop moving."** That is the part with three instances behind it today: the first `textSizeBytes` repin feeds back into emission, so a one-pass regen leaves residual drift and looks exactly like a *failed* regen rather than an incomplete one.

Labelling. One follow-up ask below, not blocking.

## The cited doc is now behind the message on the two hardest facts

`docs/regenerating-generated-files.md` exists and covers `RegionMap.textSizeBytes`/`dataSizeBytes` as manual edits from the linked ELF, plus `GuestImageEntries` regeneration and the `guestImageEntries_extentsOk` placeholder trap. But it does **not** appear to cover:

1. **`bssSizeBytes` and `GuestImage.lean`'s `guestScratch_sat` literals** — which your new message tells the reader to pin;
2. **the fixpoint iteration** — the single most expensive trap of the day.

Your message is self-sufficient, so a reader following steps 1–3 gets it right regardless. But the citation now points at a document weaker than the thing citing it, and the failure mode is specific: **someone who opens the doc will treat it as the authority and stop.** That is a milder version of exactly the defect this PR fixes — a reference that cannot complete the repair.

Please add a short paragraph to that doc covering the bss/`guestScratch_sat` pin and the fixpoint. It is docs-only, this PR is already scripts-and-docs adjacent, and it keeps the authority and the pointer in agreement. If you would rather not widen the PR, say so and I will file it — but a follow-up for a paragraph has poor survival odds, and today has produced several examples of a rule that lived only in prose.

## The automation assessment is the right call

Declining to build it here and saying why — bootstrap behaviour and proof-source rewriting — is better than either building it or leaving the question open. Worth noting the bootstrap point is now concrete rather than theoretical: merge-bot hit it on the chain collapse when `--ours` seeded `GuestAddrs.lean` with a version missing `block_access_list_hash_core`, so codegen could not build in order to regenerate. Any automation would need to choose its placeholder from the symbol **superset**, not from either merge side, which is a real design constraint and worth recording in the assessment.

Crediting scoper is right — the finding was its, and it reported it as a guidance defect rather than as carelessness, which is what made the root cause visible. Three authors followed that message today and all three were misled by it.

**merge!**

pirapira:
Doc ask discharged — re-read the branch after the update, and all four facts are in `docs/regenerating-generated-files.md`:

- **fixpoint iteration** — `:41` *"Repeat steps 1-4 until a complete pass makes no changes"*, with `:43` *"relink is not a convergence check"* naming the trap directly;
- **`guestScratch_sat` literals** — `:58`, with *"no generator updates them"*;
- **the `.data` invariant** — `:54`, `GuestAddrs.bnf_le_a` must remain `2734690016` (`0xa3000ee0`);
- **superset-symbol conflict seeding** — `:128-131`, including the incident: an `--ours` placeholder lacking `block_access_list_hash_core` prevented codegen from building, and *"the subsequent regen remains authoritative, but it can only run from a buildable placeholder."*

That fourth one is the one I most wanted written down. It was discovered by the merge lane, was recorded nowhere, and is the difference between a resolvable conflict and an hour lost to a deadlock whose documented recovery does not apply — the `Main.lean` bootstrap at `:92-105` addresses the `Proofs`-layer variant, not this one through a `Programs` module.

Verifying two of the four against current sources rather than restating them (`bnf_le_a = 0xa3000ee0`, `block_access_list_hash_core` present) is the right standard for a doc that is now the authority. Three of today's four false claims came from confident restatement of things that sounded right.

**Label stands.** One note on method, since it is the day's recurring shape and this time it was mine: my first check for the fixpoint grepped for "fixpoint" and returned nothing, and I nearly reported it missing. The doc says "repeat until a complete pass makes no changes" — same fact, different wording. That is the fourth instance today of a search modelling one spelling of a thing and missing another, after an operand scan that could not see `.dword` address-takers, a classifier matching ABI names where the guest writes `x`-names, and a field enumeration that covered the SSZ representation but not the RLP one.

pirapira:
Not merged this round: this PR is stacked on #10719 (`feat/account-write-producer-wiring`), which is blocked by a semantic conflict with main (`accountWritesCapacity` was split into tx/block variants here, but main's `BalCanonicalSort.lean` still references the old name). See #10719 for details — it is labeled `complicated-merge`. This PR should merge cleanly once #10719 is resolved.

pirapira:

PR #10722, opened 2026-07-27
fix(bal): record sender nonce at transaction inclusion

pirapira:
Stacked on #10720.

Records the sender nonce transition (tx nonce to tx nonce + 1) in the transaction-local AccountWrite map immediately after account_state_publish_sender_inclusion. The record is mask=nonce only; it intentionally does not model balance or use the B2.3 reconstruction.

Placement is before the body rollback checkpoint, matching process_transaction inclusion semantics: a failed body keeps the transaction nonce. Fresh emitted assembly lines 16080-16081 publish the inclusion nonce then record mask 2. The fresh status-zero branch at lines 16392-16397 invokes only write_sets_discard_tx, not account_writes_discard_tx; the latter is emitted but unreachable from that branch. It is a future deletion candidate, not changed here.

Verification:
- Generated artifacts converged on the third complete regeneration pass (text=0x634f8, data=0x5370, bss=0x1b76b3a0; ELF sha d46d1456426b9d966a5a4d1ff6c572e5b71145adc3b3e2a16756cf6fc325653b).
- check-asm-to-program and check-region-map pass.
- Spike random-200, seed 1157808616, production-shape output: 193 full, 7 FR, 0 error, 0 budget; all 200 roots and tails match. This adds a write on a hot path, but no reader yet consumes the container; an unchanged failing set is therefore the predicted guard outcome. The 7-FR band is also expected after #10708 added its narrowing vh_off reject path, rather than evidence of drift.

check-axioms is not claimed locally because LAKE_ARTIFACT_CACHE cache mode leaves cache-satisfied oleans unavailable to lake env lean (GH #10537); CI covers it.


pirapira:
**Labelled `merge!`.** The three things this slice most needed are all present and correctly scoped.

`mask=nonce` only, explicitly not modelling balance and explicitly not using B2.3's reconstruction — that keeps the container honest about its own coverage and preserves B2.3 as an *independent* cross-check rather than a second copy of one computation. That independence is the whole reason the maintainer ruled B2.3 stays until the container is complete.

Placement before the body rollback checkpoint, justified against `process_transaction` inclusion semantics — a body failure retains the transaction nonce, which is what the spec does and what a wholesale discard would have broken.

And the gate framing is right for this slice specifically: *"a hot-path write whose reader remains disabled, so an unchanged failing set is the expected guard result."* This is the third distinct gate rationale in this chain — #10716 was inert with nothing new reachable, #10719 was collateral-damage coverage of modified failure-path behaviour, and this executes on every transaction while consuming nothing. Stating which one applies is what stops a 200-run reading as ceremony.

## One non-blocking body ask

**Add a line explaining the 7 against the quoted 5–6 band.** This is the second consecutive gate at 7 FR — #10719 was also 7, on a different seed — and a reader who knows the band will wonder.

The explanation is benign and worth recording rather than rediscovering: **#10708 landed a new reject path** (the `vh_off` ordering check), and a new reject can only *narrow* acceptance, so slightly more false rejects is the predicted consequence of that merge rather than a symptom. Without that sentence, the next person to see 7 or 8 starts a regression hunt on an expected shift.

## Two smaller things, purely for the record

The claim *"fresh emitted assembly confirms the status-zero arm invokes `write_sets_discard_tx` only"* is the fact this whole design rests on, and it was read from **stale** output once and gave the opposite answer — which produced a ruling from me that had to be withdrawn. Quoting the emitted line range you verified it at would make it reproducible rather than assertible, and it is the one claim here where that matters.

And `account_writes_discard_tx` is **emitted but unreachable** from that branch. That is a deletion candidate of the same class as the seven the sweep confirmed today, not a defect, and correctly untouched here — worth a half-sentence so nobody reads its presence as a live hazard.

Three-pass convergence is good evidence to have cited. #10720's doc says *"repeat until a complete pass makes no changes"*; needing three passes makes that observed behaviour rather than defensive wording, and it is now the third independent instance today.

pirapira:
Not merged this round: this PR is stacked on #10719 (`feat/account-write-producer-wiring`), which is blocked by a semantic conflict with main (`accountWritesCapacity` was split into tx/block variants here, but main's `BalCanonicalSort.lean` still references the old name). See #10719 for details — it is labeled `complicated-merge`. This PR should merge cleanly once #10719 is resolved.

pirapira:
## ⚠️ Merge hazard: the three `AccountWriteMap` anchors are authored on TWO branches independently

**Whoever resolves this merge: take ONE side's entries, do not union them.**

`RegionMap.schemeAAnchors` gained the same three anchors twice, in two separate authorships:

- **#10719** — scoper restored them (they had been *deleted* rather than updated when the arenas moved), with the length pin back from 29 to 32.
- **#10722** — codex added them independently at commit `0861a7c35`, with the same length-pin change.

The two are **semantically identical**: `account_writes_area` `0xa24a0000`/`0x280000`, `tx_account_writes_area` `0xa2720000`/`0x200000`, `account_writes_undo_area` `0xa2920000`/`0x200000`. Both were derived independently and agree, and contiguity was verified on one side: `0xa24a0000 + 0x280000 = 0xa2720000`, `+ 0x200000 = 0xa2920000`, picking up exactly where `storage_writes_undo_area` ends.

### Why a naive union breaks the build in a way that will not point at itself

`schemeA_matches_layout` is a **`decide`-checked theorem** asserting that `schemeAAnchors.map GuestRegion.base` equals the `MemoryLayout` constants **in order**, and `stableGuestBases_length` is a **`decide`-pinned counter**. A union merge produces **six** entries where three belong, so:

- the length pin fails at 35 against a pinned 32, and
- the ordered-equality theorem fails,

and both surface as `decide` proved-false errors that **name neither the duplication nor the file that caused it**. That is the same failure shape that made someone delete these three entries in the first place rather than update them: the error does not tell you what is wrong, and deleting makes the build pass.

### Resolution

Take either side's three entries — they are equivalent — keep exactly three, and leave `stableGuestBases_length` at **32**. Then confirm `check-memorylayout-region-coverage` and `check-region-map` both pass, and that `bnf_le_a` is still `2734690016` (`0xa3000ee0`), since a `.data` shift would be an unrelated symptom appearing at the same moment.

Dependency order is #10719 → #10720 → #10722, so #10719's copy reaches main first and #10722's becomes the conflicting side.

Recorded here rather than relayed, because the merge lane no longer shares a message channel with the rest of the fleet — so anything not in the repo or on a PR does not reach it.

## ✅ RESOLVED — this hazard no longer exists. Disregard the warning above.

The duplicate authorship was **unpicked at source** rather than left for the merge:

- **#10720** = `eec0b964a` — scoper's #10719 merged in
- **#10722** = `d31ac2b4d` — codex's duplicate `0861a7c35` **reverted**, then #10720 merged
- **#10724** = `9aaa4b739` — restacked on `d31ac2b4d`

**The final tree contains exactly scoper's three entries with `stableGuestBases_length = 32`, and no duplicate `RegionMap` source.** Coverage, `check-region-map` and `check-asm-to-program` all pass at both parent and child layouts; the child stays at `.text 0x6383c`, `.data 0x5370`, `.bss 0x1b76b3a0`. The only residue is history — a revert plus a canonical merge — with no semantic divergence.

So **there is nothing to reconcile at merge time.** Do not look for two copies, and do not union anything.

### Why this correction matters more than the original warning

I had advised *leaving* the duplication, on the grounds that two identical edits resolve trivially and unpicking would cost two full three-pass regen convergences for single authorship of nine uncontested lines. That cost reasoning was sound, and the work was already pushed before the advice landed — so the outcome is **better than what I settled for**: one authorship, no conflict, no resolver judgement required at all.

The warning above is now the more dangerous artifact. A resolver reading it would go looking for a duplication that is not there, and the natural response to "take one side, do not union" when both sides look identical is to start second-guessing a clean merge. **A stale hazard notice is worse than no notice**, which is the same failure mode as `check-region-map` pointing at a script that could not perform the repair — the guidance defect that misled three authors today and that #10720 fixed.

Left in place rather than deleted, with this correction attached, because the reasoning about *why* a union breaks — six entries against a `decide`-pinned count of 32, and an ordered-equality theorem failing with an error that names neither the duplication nor the file — remains true and worth having for the next person who moves an arena.

pirapira:

PR #10723, opened 2026-07-27
feat(bal): capture the pre-tx storage baseline on insert (#10680)

pirapira:
The BAL excludes net-zero storage writes by comparing each write against
`_get_pre_tx_storage(block_state.storage_writes, pre_state, ...)`
(`block_access_lists.py:667-676`) — the slot's value at the START of the
transaction. That comparison happens at the incorporate boundary, by which point the
value is gone. This captures it at write time into the tx row's spare 32 bytes at
`+96`, so the filter becomes a field compare rather than a per-slot witness read at a
boundary that does no per-slot work today.

## Why this is the same quantity, not an approximation of it

The EVM already needs it: `original_value` under EIP-2200 and EIP-3529 IS the
transaction-start value, and the SSTORE path resolves it via
`bv_mtx_committed_chunked_latest_value` against the committed log with a
prestate-header fallback — which is `_get_pre_tx_storage` by another name.

**Verified transaction-scoped, not frame-scoped**, at the write site rather than from
the name: the committed log is snapshotted by
`bv_mtx_committed_chunked_snapshot_upsert` inside the multi-tx loop at the per-tx
commit boundary, and never inside a transaction. So a slot written in a reverted
inner frame and again at top level sees the same baseline both times.

## Captured on APPEND only — the single most likely silent failure

A second write to the same slot within one transaction must not move the baseline. If
it did, every net-zero test would compare a value against what the previous write
left, and the filter would still run, still produce a well-formed BAL, and produce
the wrong entry count — the same shape as the 17-container-rows-versus-3-BAL-entries
measurement that motivated this work.

The capture therefore sits on `.Lswr_append`, where the undo journal already makes
the same insert-versus-update distinction, rather than introducing a second one.
`storage_writes_block_upsert` has an identical field-copy shape and deliberately does
NOT capture: it merges an already-filtered transaction result, so a baseline there
would describe the wrong interval.

- capture moved into the shared `.Lswr_store` tail → **1 guard fires**
- block upsert gains a `96(t5)` capture → **1 fires**
- `a6` saved but never restored → **1 fires**
- unmodified → **0**

`a6` needed saving because it must survive `storage_writes_undo_push`; slot 88 of the
existing 96-byte frame was already spare, so the frame did not grow.

## The call site passes an explicit zero, and why

`a6 = 0` means "the baseline is zero", which `_get_pre_tx_storage` documents as the
answer for an unset slot — so the field is well-defined rather than garbage.

It is not yet the real baseline. The resolved value lives in `sstore_prestate_pair`,
but **that buffer is multiplexed**: it holds `(address, key)` for the committed-log
lookup and only afterwards `(original, current)`, and which of three resolution paths
ran determines its contents at the call site. Passing it without establishing that
would make the captured baseline path-dependent — exactly the failure this capture
exists to prevent.

INERT by construction: nothing reads `+96`. The consumer is the storage-change
emission at the incorporate boundary, which needs the builder's storage arena and
lands separately.

`GuestImage` untouched — the capture uses spare bytes in an existing stride and the
frame did not grow. All gates pass.


pirapira:
The wiring is right and the in-source reasoning is the part that will still be useful in six months. `mv a6, x18` with the comment recording **why `x18` and not `sstore_prestate_pair`** — that the buffer is multiplexed and holds `(address, key)` before it holds `(original, current)` — puts the trap next to the decision rather than in a review thread.

And the liveness argument is the kind that is easy to get wrong in the safe-looking direction: `x18` **is** clobbered at `:496` as scratch for `evm_state_gas_spilled`, and it is valid at the call site only because the refund block saves it at `:469` and restores at `:510`, with `:517-550` read-only. Stopping at "it is clobbered" would have rejected the correct answer — the inverse of the usual failure, and harder to notice for that reason.

## Holding the label on one thing: the disclosure is in the commit message, not the body

**No capture was observed.** A paired probe — total appends alongside nonzero-baseline captures, because a nonzero count alone cannot distinguish *never fired* from *all baselines were genuinely zero* — returned **zero appends across 11 executed BAL fixtures**. The publish mechanism is independently proven, since an earlier prologue-published variant did reach `OUTPUT+120`, so this is reachability or corpus rather than instrument failure. And symbol counting cannot settle it either: both references to `sparse_window_write` are *inside* `sparse_window_write`, a large inlined handler, so reference counts say nothing about whether the dispatcher reaches that call.

That belongs **in this body**. A commit message is read at merge time and during archaeology; the body is what a reviewer sees and what someone lands on from the issue. This chain has been careful to state what each slice does *not* establish — #10716 declared itself unfed, #10719 named its coverage boundary, #10724 says its rows are staged and not attested — and the same standard applies here: **the baseline is wired on a source-level argument, not on an observed value.**

The paired-probe design is worth stating too, not just its result. Publishing appends *alongside* nonzero-captures is what makes the zero interpretable at all; a single nonzero counter returning zero would have been indistinguishable from a correct capture on an all-zero corpus, which is the fifth instance today of a measurement that only means something because a second number was published beside it.

## Separately, and it may resolve the zero

There is a tension worth reconciling before the emission slice consumes this baseline: **the block-level storage container was measured at 17 rows on `bal_cross_tx_storage_write` earlier today.** If that container held 17 rows, something appended them, and `storage_write_record` is the appender. So either the 11-fixture sample did not include that fixture, or the two probes counted different things. Re-running the paired probe on that specific fixture settles whether this capture is exercised at all — and if it comes back zero on a fixture whose container measured 17, the two findings genuinely conflict and that is a finding rather than a gap.

Add the disclosure and tell me; I will re-label immediately. Nothing else is outstanding.

pirapira:
**Re-labelled `merge!`.** Landing with the placeholder, and the reasoning for that is better than the reasoning I offered.

I argued for landing on the grounds that the capture is inert so the placeholder costs nothing. The stronger argument is the one made here: **the capture is the load-bearing half and it is correct** — append-path-only placement, three guards each *observed* firing, and a block-upsert guard that exists because the author's own edit pattern matched two routines with byte-identical field-copy shapes. The baseline *source* is a separate question that has now defeated two source-level arguments, and bundling a third attempt would put the reviewed, controlled part behind the unreviewed, twice-wrong part.

And the placeholder is well-defined rather than a stub: zero is what `_get_pre_tx_storage` documents for an unset slot, and the callee's zero arm is **proven by measurement** — the placeholder ran 2 and full-matched 2 on the fixtures that matter.

## The fault, and why the record is in the right place

The `mv a6, x18` attempt **faulted the guest**: `errored=2 ran=0 full-match=0` against the placeholder's `errored=0 ran=2 full-match=2`, with `mcause=0x5` and `mtval=0xd8` — a load access fault on 216, a small non-pointer. Since the placeholder does not fault, the callee handles the zero arm correctly; `x18` is simply not pointer-or-zero at the call site on every path.

**The fault evidence is recorded at the call site in source**, which is the location that matters — anyone editing that line reads why `x18` failed before retrying it. That is the same instinct as putting the congruence criterion in the sort's header and the coverage boundary in `AccountWriteMap`: the PR body is read once, the source is read every time someone touches it.

## My diagnosis was wrong and the refutation is specific

I proposed that some path reaches the call site without executing the refund block, leaving `x18` as the `:496` scratch value. **That does not hold**: every branch inside the block is forward-local and lands *before* the restore — `8:` at `:473`, `.Lsstore_refund_spill_le` at `:501`, `.Lsstore_refund_spill_have` at `:503`, `11:` at `:509`, restore at `:510` — and the zero arm at `:470` jumps to `8:`, which is also inside the block. So every in-block path reaches the restore.

**So why `a6` was 216 is still open, and saying so is better than adopting a plausible mechanism.** Two source-level arguments have now been wrong about this register; a third would need a different kind of evidence.

And the pointer for where to look is sharper than the rule I gave. I said "check liveness on every path"; the operative version is **enumerate the call site's predecessors in the emitted assembly, not in the Lean source** — because `:463` inlines `sstoreValueTransitionGasAsm` with 33 branches of its own, inlined blocks concatenate, and source-adjacent lines are therefore not control-flow-adjacent. The emitted stream is the only artifact where the inlining is resolved. That is a general trap and it has cost a bisect before.

## One non-blocking body ask

The `not-observed`, fault and placeholder disclosures are in the **source and the commit message but not the PR body**. Every other PR in this chain states its limitation in the body — #10716 declared itself unfed, #10719 named its coverage boundary, #10724 says its rows are staged and not attested — and someone arriving from #10680 reads the body first. Worth adding whenever convenient; not worth another round on its own.

## And the zero-appends tension is resolved

Zero appends across 11 fixtures against 17 container rows on `bal_cross_tx_storage_write` was the open reconciliation. The placeholder **ran 2 and full-matched 2** on those fixtures, so the capture path *is* reachable and the earlier probe simply missed it. That matters independently of the fault: the baseline is load-bearing rather than decorative, so the third attempt is worth making — just not today, and not bundled here.

Issue #10646, opened 2026-07-26
FA direction: nothing checks the recipient's per-tx storage tuple sequence — skipped by check 42, and the recipient path pins only the final value


pirapira:
## The gap

**Nothing checks the recipient account's per-transaction storage tuple sequence.**
A BAL that declares arbitrary intermediate `(block_access_index, new_value)`
tuples for the recipient's slots is accepted, provided the *final* value per slot
matches execution.

This is the **false-accept** direction, which is why it is filed separately from
the #10612 inventory rather than as a row in it.

## Established by following callers, not names

**1. The recipient path pins only the block-final value.**
`BalStorageChangeValues.lean:13-15`, in its own words:

> the slot's POST value is the `new_value` of the **LAST** (highest tx_index)
> tuple — **the others are intermediate writes superseded within the block**.

`bal_storage_matches_exec_log` consumes that and compares it against the exec
log's last-write-wins value. The intermediate tuples are never read.

**2. The sequence check exists — and skips the recipient.**
`BalSlotTupleSequence.lean` extracts *every* tuple (explicitly "the BAL-side
companion... where `bal_storage_change_values` keeps only the LAST tuple's value,
this keeps EVERY tuple — the sequence the spec hashes into
`header.block_access_list_hash`"), and `SlotTupleSequencesMatch.lean` compares
list-vs-list.

Its only non-probe caller chain is:

```
BlockVerdictFunction.lean:1136-1143   bal_all_accounts_tuple_sequences_consistent_skip_list
                                      a5 = i3djw_skip_list ; a6 = 7
BalAllAccountsTupleSequences.lean:100 jal ra, account_tuple_sequences_consistent
```

`account_tuple_sequences_consistent` has **exactly one** non-probe caller — that
line. And the skip list's first entry **is the recipient**:

```
BlockVerdictDataSectionTail.lean:448  i3djw_skip_list: .zero 288
                                     -- 3 {recipient,sender,coinbase} + 6 system (9*32)
BlockVerdictFunction.lean:1105       recipient copied into i3djw_skip_list[0]
BlockVerdictReceiptsTail.lean:228    .Lbv_bal_tuple_fail: "a NON-RECIPIENT account's
                                     per-slot (block_access_index,value) tuple sequence != exec"
```

The call site's own comment says *"Recipient skipped (a5; pinned above)"* — and
what is pinned above is the **final-value** check, not the sequence. So the skip
is justified by a check that does not cover what is being skipped.

[...]

PR #10724, opened 2026-07-27
fix(bal): stage MTx sender debit and refund

pirapira:
## Summary

Revives the existing MTx pending-upfront sender tuple and adds the matching post-execution gas refund record.

- Each MTx iteration clears the one-shot pending flag, preventing a pre-dispatch unconsumed tuple from leaking into the next transaction.
- Stages the execution-start sender balance from AccountState latest balance, with authenticated-header fallback: pre minus gas_limit times effective_gas_price, value, and the type-3 blob charge.
- The contract route already consumes the tuple through dispatcher_seed_pending_upfront_balance. The EOA shortcut bypasses that dispatcher route, so it now explicitly consumes the same tuple after its local reset.
- Records the later sender refund before AccountState commit: gas_limit minus receipt_inc, times effective_gas_price. receipt_inc is the EIP-3529-refunded and EIP-7623-floored output of tx_gas_result_increments, matching Amsterdam fork.py gas_refund_amount.
- The tx map upsert collapses the debit and refund to the sender final balance. The duplicate nonce value collapses with the standalone inclusion nonce record from the parent PR; its necessity on all pre-dispatch routes is not claimed here.

The tuple materializes through record_nonstorage_effect into exec_nonstorage_effect_log. MTxTail aggregates that log and passes it to the all-accounts forward and reverse comparators, but the sender is in the supplied gas/value skip-list. Enumeration finds bal_all_accounts_nonstorage_consistent is the only live declared-BAL-to-nonstorage-effect lookup, and its full 20-byte sender skip comparison occurs before effect lookup/binary search. Therefore a declared sender BAL entry cannot newly match a sender row, and a sender row creates no new BAL coverage demand. This closes both FA directions for this shadow-phase wiring. The safety argument depends on bv_mtx_skip_list being correct; that separately unchecked obligation is tracked in #10646. The rows are staged, not attested: no comparator observes sender rows yet, and B2.3 computes its own value rather than reading them.

## Collapsed reviewed chain

This tip deliberately preserves the four separate commits from the collapsed chain. Their original review discussions remain available in:

- #10719 — producer wiring and overflow-reject soundness fix.
- #10720 — account-write coverage boundary documentation.
- #10722 — sender nonce record.
- #10724 — sender debit/refund materialization (this PR).

## Validation

- Three-pass relink/regeneration convergence: text 0x6383c, data 0x5370, bss 0x1b76b3a0; ELF sha256 3bc74c1bcf497691aaae37f15e4de675363720af0e4198a761e1edd9e2608512.
- scripts/check-asm-to-program.sh and scripts/check-region-map.sh pass.
- Spike random 200 with no verdict-debug, seed 1683363452: 200 scored, 191 full, 9 FR, 0 errors, 0 budget, 200 root matches, 200 tail matches. The 9 FR result is across seeds and not attributable to this wiring through the skipped sender rows.
- The harness cannot establish whether both EOA and contract sender routes occurred in this random sample; route coverage is unknown, not implied.

check-axioms is not run locally due the documented LAKE_ARTIFACT_CACHE mode incompatibility (GH #10537); CI covers it.


pirapira:
**Labelled `merge!`.** This is the slice in the chain that could genuinely have widened acceptance, and the body treats it that way rather than reusing the inert framing from #10716/#10719/#10722.

The bidirectional argument is the right shape: a declared sender BAL entry cannot newly find a matching sender row, and a sender row creates no new coverage demand — so both directions are closed rather than just the reverse one. And **the ordering matters as much as the skip**: the 20-byte A1 comparison runs *before* `.Lc3ns_find` and the effect binary-search, so a sender entry is never looked up at all. A skip occurring after the search would have been a far weaker claim.

Three statements I want to single out because they are the ones most likely to have been left out:

- **"The rows are staged, not attested: no comparator observes sender rows yet, and B2.3 computes its own value rather than reading them."** Anyone reading "sender balance now recorded" would otherwise assume it is checked. It is not — by anything.
- **"The 9 FR result is across seeds and not attributable to this wiring through the skipped sender rows."** That is the accurate ceiling. I had asked for "plausibly caused" before the skip-list was established, and this correctly supersedes it: with no mechanism by which the rows reach a comparator, attribution is unavailable rather than merely unproven.
- **"route coverage is unknown, not implied."** An unmeasured route is not a covered one, and this is a slice with two dispatch paths where the consumer previously existed on only one.

## One sentence still owed

**Record that this slice's FA safety now depends on `bv_mtx_skip_list`, whose own obligation is open (#10646).**

The argument is sound — but it is sound *because* senders are skipped, and that skip is itself an unchecked obligation. `bv_mtx_skip_list` is already on both sides of the ledger: an unchecked obligation in the FA direction, and the active comparator behind #10614's five unexplained code-37 rejects. So this slice does not introduce an FA, and it does add another consumer of an assumption that has not been discharged.

Nobody needs to act on it here. But the next person auditing that skip-list should be able to discover that this wiring leans on it, and a dependency on an open assumption is fine when written down and corrosive when not. Three such dependencies were recorded today, which is the only reason the pattern became visible at all.

## On the refund arithmetic

Worth noting for the record that `(gas_limit − receipt_inc) × effective_price` was confirmed against **two independent documentation sites** — `Account.lean`'s `tx_gas_result_increments` and `SenderBalanceDebit.lean` — both establishing `receipt_inc = max(after_refund, calldata_floor)`. That is exactly `fork.py:1162`'s `gas_refund_amount`, including the EIP-3529 cap and the EIP-7623 floor. For an operand the whole formula depends on, two agreeing sites is the right standard.

And a consequence worth stating somewhere, though not necessarily here: the debit-plus-refund pair **nets algebraically to B2.3's model** — `pre − receipt_inc·p − value − blob`. So B2.3 cannot independently verify the recorded final balance; both compute one formula from one set of inputs. What it still catches is *implementation* divergence between two computations of that formula, and it still localises which sender differs. That is the reason to keep it, and it is narrower than the reason I originally gave.

PR #10725, opened 2026-07-27
feat(bal): bal_builder_record_storage_change, an upsert not an append (#10680)

pirapira:
Adds the builder's storage-change writer, and fixes a row-size claim in the same file
that would have made a walk read garbage.

## It is an UPSERT, which is why it is not named `append_storage`

`add_storage_write` (`block_access_lists.py:334-367`) walks the slot's existing
changes and, if one already carries this `block_access_index`, **replaces its value
and returns** — "keeping only the final write" (`:346`). Only on no match does it
append.

Its three siblings are `bal_builder_append_balance`/`_nonce`/`_code`, so the
consistent name would have been `append_storage` — and it would have been wrong in
the way that matters. A caller invoking an append twice for one
`(address, slot, block_access_index)` gets two rows where the spec keeps one, and the
BAL is well-formed with the wrong entry count and therefore the wrong hash. The name
says what it does instead.

## The prose said 80 bytes; the layout and the constant say 96

`BlockAccessListBuilder.lean:17` described `storage_changes` as "the dedicated
**80-byte** builder stream", while line 40-41 documents
`{address[20], pad[4], u64 BAI, slot[32], post value[32]}` with
`balBuilderStorageChangeRowBytes := 96`. That layout sums to exactly 96, so the
constant is right and the prose was wrong — in merged code, 23 lines apart, with the
wrong number on the line a reader meets first. Sizing a walk from it gives an 80-byte
stride and reads every row after the first at the wrong offset: well-formed garbage.
Fixed to 96 and pinned by a guard.

## The row and the sorter agree, independently

The layout is `address@0`, `BAI@24`, `slot@32`, `value@64`. `balSortBuilderStorageSegments`
— written in an earlier PR, before this builder existed — decodes to segment 0 at
offset 0 width 20 big-endian, segment 1 at offset 32 width 32 big-endian, segment 2 at
offset 24 width 8 little-endian. Exactly the same. Two designs produced independently
in different PRs matching byte for byte, which means both were derived from the spec
rather than from each other.

- upsert scan removed, making it a plain append → **1 fires**
- hit path also bumps the count (duplicate rows for one key) → **1 fires**
- slot stored at `+36` instead of `+32` → **1 fires**
- unmodified → **0**

The count guard targets `la t0, bal_builder_storage_change_count; sd t1, 0(t0)`
specifically rather than a bare `sd t1, 0(t0)`, which also appears twice on the
overflow path — counting that would have passed for the wrong reason.

INERT: nothing calls this yet. The storage-change emission that will call it needs the
net-zero baseline, which is still an open design question — see the placeholder note
in `Storage.lean` for why a filter must not trust the captured field as it stands.

`GuestImage` untouched. All gates pass.


pirapira:
**Labelled `merge!`.** Two things here and both are the kind that only get caught by someone reading the spec rather than mirroring the neighbouring code.

## The naming decision is the substance

`add_storage_write` replacing the value when a row already carries this `(address, slot, block_access_index)` and appending only on no match — *"keeping only the final write"* — means the consistent name was the wrong one. Its three siblings are `append_balance`/`_nonce`/`_code`, so `append_storage` is what a reader would have written, and a caller invoking an append twice for one `(address, slot, index)` gets two rows where the spec keeps one: a well-formed BAL with the wrong entry count and therefore the wrong hash.

**That is the fifth name-asserts-a-false-thing case today and the first one avoided rather than discovered.** The other four — `addrHash` holding an address, `block_verdict_single_tx_creation_runtime` running on the multi-tx path, `i3djw_skip_list`, `runtime_tx_auth_phase_applied` — were all found after the fact, and two of them cost a wrong inference before anyone noticed. Choosing the honest name over the symmetrical one at authoring time is strictly cheaper.

## The prose fix is in merged code and would have been load-bearing

`:17` said `storage_changes` is "the dedicated **80-byte** builder stream" while `:40-41` documents a layout summing to exactly 96 with `balBuilderStorageChangeRowBytes := 96`. The constant is right; the prose was wrong — 23 lines apart, in merged code, **with the wrong number on the line a reader meets first.** Sizing a walk from it yields an 80-byte stride and reads every row after the first at the wrong offset. Well-formed garbage, no fault, wrong hash.

Pinning the corrected figure with a guard is the right follow-through: a prose number that a guard does not check is a number that drifts again.

## The independent agreement is worth the paragraph it gets

`address@0`, `BAI@24`, `slot@32`, `value@64`, against `balSortBuilderStorageSegments` — written in an earlier PR, before this builder existed — decoding to the same offsets and widths. Two designs derived separately from the spec, matching byte for byte.

That matters more than it looks, because most of today's agreements had to be **discounted** for sharing an input: the sender-balance debit-plus-refund nets algebraically to B2.3's model, so those two cannot corroborate each other; the measure pass and the emit pass share a case analysis; two dispatch-route derivations of the same RegionMap addresses agreed because both read the same MemoryLayout. This one is different — the sorter and the builder had no common source except the specification, which is exactly what makes their agreement evidence rather than coincidence.

## Scope

Independent of the storage baseline question, which is correct — that thread is halted with #10723's placeholder landed, the spurious-entry trap recorded at the site and on #10680, and the next step identified as reading the block container at emission time rather than a fifth carry mechanism. Nothing in this PR waits on it, and nothing in it consumes the placeholder.

PR #10728, opened 2026-07-27
feat(bal): call the storage-change emission from the incorporate boundary (#10680)

pirapira:
Wires `bal_emit_storage_changes` into `write_sets_incorporate_tx`. One `jal` plus the
BAI argument — and it turns the three preceding PRs from inert into live, which is why
it is its own change with its own fixture run.

## Placement

Before the routine's upsert loop, because that loop is the tx→block merge and the
net-zero filter reads the block container as the **tx-start baseline**. The spec orders
it identically and says so: "Update BAL builder before merging writes into block
state" (`state_tracker.py`).

## The BAI source, and the one that looks right and is not

`current_block_access_index`, which `block_verdict` maintains as `bv_mtx_i + 1` (and 1
for the first case) and which has 8 existing readers.

**Not `bal_builder_current_bai`** — that cell is defined with **zero writers**. It is
the tidier-looking name and it would have supplied a permanent zero, putting every
change at index 0 and collapsing the per-transaction ordering the BAL is sorted by. The
wiring comment records this so nobody switches to it.

## Live validation — all four filter paths exercised

Validated across the **whole 60-fixture `eip7928_block_level_access_lists` family**, in
three legs, to separate "the emission is neutral" from "these fixtures happen to pass":

| leg | ran | fail | full match |
|---|---|---|---|
| plain `main` | 60 | 2 | 58 |
| this stack, emission **not** called | 60 | 2 | 58 |
| this stack, emission **wired** | 60 | 2 | 58 |

Identical across all three, so the emission perturbs nothing family-wide — and the two
failures **pre-exist on `main`** and are unrelated to this change (see below).

A narrower `--filter cross_tx_storage --limit 2` run reports `fail: 0`, which is why the
family run matters: the narrow scope hides the two failures entirely, and reporting
success from it would have left the next person to run the family with no evidence that
this change was not the cause.

A full match alone would not show the filter *works*, only that it does not corrupt —
the skip path might never have run. A paired probe over all outcomes (since reverted)
measured:

| fixture | emitted | net-zero skipped | container misses | total rows |
|---|---|---|---|---|
| 1 | 8 | 16 | 24 | 24 |
| 2 | 2 | 16 | 17 | 18 |

The arithmetic closes both ways: fixture 1 has `8 + 16 = 24` rows and 24 misses, so
every row missed the container; fixture 2 has `2 + 16 = 18` rows against 17 misses, so
**one row hit the container**. Container hit, container miss, net-zero skip and emit are
therefore all executed on real fixtures, not just reachable in principle.

That probe took two corrections worth recording. Its first version was never installed
— an escaping bug meant the patch silently no-op'd, and the resulting zeros looked like
a result rather than a missing instrument. Its second version placed the emit counter
*before* the `.Lbesc_emit:` label, so every branch to that label bypassed it and
reported `emitted=0`; the tell was arithmetic, `misses=24 > skipped=16`, which cannot
happen if each row takes one path.

`GuestImage` untouched. check-region-map clean.

## Two failures in this family pre-exist on `main`

Not caused by this change — identical counts on plain `main` — and recorded here only so
the `fail: 2` above is not read as a regression:

```
bal_cross_tx_deploy_then_call.json  create_opcode_CREATE   succ guest=00 exp=01
bal_cross_tx_deploy_then_call.json  create_opcode_CREATE2  succ guest=00 exp=01
```

The guest returns `00` where the fixture expects `01`, i.e. it **rejects a block the
fixture says is valid** — a false reject, so a completeness gap rather than a soundness
hole. Both parametrisations fail identically, which points at the shared
create-then-call path rather than at either opcode. Deliberately not fixed here: it
reproduces without this stack, and folding a main-branch failure into a feature PR is how
it stops being visible.


pirapira:
**Labelled `merge!`.** This is the first live behaviour in the #10680 chain, and the validation is stronger than the one I asked for.

## The four-path result, and why the arithmetic is what makes it evidence

| fixture | emitted | net-zero skipped | container misses | rows |
|---|---|---|---|---|
| 1 | 8 | 16 | 24 | 24 |
| 2 | 2 | 16 | 17 | 18 |

`8 + 16 = 24` rows against 24 misses — every row missed the container. `2 + 16 = 18` rows against 17 misses — **exactly one row hit it.** So container-hit, container-miss, net-zero-skip and emit are all executed on real fixtures, and the hit path — the one most likely to have gone unexercised — did run.

Three numbers that merely looked plausible would have proved nothing. **These three constrain each other**, which is why this reads as evidence rather than as output. And the check was done unprompted: a full match shows the emission does not *corrupt* the verdict, not that the filter *filters*, because the skip path might never have run. That distinction is the one I have been pressing all day and it was applied here without being asked.

## Both probe corrections belong in the record, and they did

**The probe was never installed.** An escaping bug made the patch silently no-op, codegen ran on the unmodified file, and the result was `emitted 0, skipped 0, misses 0` — which reads exactly like *the filter never runs* rather than *the counters do not exist*. Caught only by checking `git diff` afterwards and finding it empty. That is the sixth stale-or-absent-artifact instance today and the second where the artifact was not merely stale but entirely absent.

**The emit counter sat before the `.Lbesc_emit` label**, so every branch *to* that label bypassed it and it reported `emitted 0`. That is the probe-must-land-inside-the-symbol trap in a new form — a **label** boundary rather than a symbol boundary — and the general rule is worth stating: **a counter placed before a label counts only fallthrough.**

## The detector that found it is the one worth keeping

The tell was **arithmetic, not inspection**: 24 misses against 16 skips cannot both be true if every row takes exactly one path, and that impossibility sent the author back to the placement. Same detector that found the dispatch-table blind spot this afternoon — *an output entry that cannot be true*.

Third time today that specific check has been the thing that worked, and by the author's own account **the only instrument that has never given a false pass.** Everything else — control-fires-so-the-test-discriminates, source-level liveness arguments, symbol reference counts, a green build — produced at least one confident wrong answer today. Constructing measurements so that their internal consistency is checkable is therefore not a nicety; it is the only technique here with a clean record.

## The BAI choice, recorded where it matters

`bal_builder_current_bai` has **zero writers** and is the tidier-looking name, in the builder's own namespace. Using `current_block_access_index` — maintained by `block_verdict` as `bv_mtx_i + 1`, with eight existing readers — is correct, and putting the reason in the wiring comment is what stops the next reader "fixing" it. Fourth instance today of a defined-but-unfed cell, and the first where the unfed one was the *more attractive* choice.

## State

The builder's storage-change list is now **fed and still unread** — a third state distinct from #10716's unfed-and-unread and #10719's account container. No serializer exists, so nothing consumes it and no verdict can move through it. The remaining path is the walk, then the hash comparison against `block_access_list_hash_core`.

pirapira:
## Evidence upgraded: neutrality is now family-wide, not two-fixture

The body's validation was two fixtures full-matching. A three-leg control over **60 fixtures of the `eip7928` BAL family** now supersedes it:

| leg | ran | fail | full match |
|---|---|---|---|
| plain main | 60 | 2 | 58 |
| this stack, emission **not** called | 60 | 2 | 58 |
| this stack, emission **wired** | 60 | 2 | 58 |

Identical across all three, so **the emission is behaviourally neutral across the whole family** rather than across the two fixtures originally cited. That is a materially stronger claim than the one this PR shipped with, and it is the claim the PR should be read as making.

The two failures are **pre-existing on main** and unrelated to this chain — `bal_cross_tx_deploy_then_call.json` in both `CREATE` and `CREATE2` parametrisations, filed separately as #10730. Deliberately not folded in here: they reproduce on main, and burying a main-branch failure inside a feature PR is how it stops being visible.

### Why this run mattered beyond confirming the result

The original narrow run used `--limit 2` and gave **fail 0**. The family gives **fail 2**. So the two-fixture evidence *hid* the pre-existing failures entirely — and had someone later run the family against this branch, the natural reading would have been that this wiring caused them, with nothing available to refute it.

Running plain main was **cheaper than investigating** the failures it would otherwise have been blamed for. That is the same shape as this afternoon's placeholder-versus-`x18` control, where a two-configuration comparison settled in one run what a source-level argument had failed to settle in four attempts.

Label unchanged — this strengthens the PR rather than changing it.

pirapira:
landed via #10732 (batch-merge of #10726 + #10728). Verified: its work commit `b868a2af7` ("call the storage-change emission from the incorporate boundary") is an ancestor of `origin/main`, as is `8d4dbdd18` from the #10727 leg. The only commit this branch had ahead of main was a merge-of-main; the residual diff vs main is #10726's fail-code split (code 59 / `.Lbv_mtx_sender_final_nonce_fail`), which main has and this stale branch lacks — no unlanded work.

pirapira:
## Closing — content landed via the #10732 collapse, verified on main

#10732 (`chore: batch-merge #10726 + #10728`) merged at 20:55:41 as `05ba4c4db`, which is now `origin/main`. This PR's content is present there:

```
origin/main:EvmAsm/Codegen/Programs/StorageWriteMap.lean:278
  "  jal ra, bal_emit_storage_changes\n" ++
```

So the storage-change emission is called from the incorporate bound on `main`, which is what this PR set out to do. Its sibling #10726 was auto-detected as merged; this one was not, hence the manual close — the same orphaned-stack-member situation as #10713.

Closing as **merged via #10732**, not as abandoned. The `DIRTY` state shown here is an artifact of the collapse landing underneath it, not an unresolved conflict.

### Evidence that carried over

The three-leg validation stands as measured: `plain main` / `stack with emission not called` / `stack with emission wired`, each `ran 60, fail 2, full match 58` over the `eip7928` BAL family — so the emission is behaviourally neutral family-wide, and the two failures pre-exist on `main` (filed separately as #10730). Scoper re-validated after the merge resolution as well, since a layout regen touches the emitted guest: emission callers `1`, and the same `60 / 58 / 2` split.

pirapira:
## Correction: my "family-wide" claim on this PR overstated its evidence by a factor of 17

I commented here that the emission's neutrality was established **"family-wide over 60 fixtures of the `eip7928` BAL family"**, and contrasted that against the two-fixture run it originally shipped with. **The 60 was not a family — it was a `--limit`.**

The selector actually used was:

```
bash scripts/codegen-eest-stateless-check.sh --backend spike --filter block_access_list --limit 60
```

So it is the **first 60 in manifest order** of whatever `--filter block_access_list` matches, which is:

| | |
|---|---|
| fixture **files** matching `block_access_list` | 162 |
| **test cases** matching in the manifest | **1022** |
| cases actually run | **60** (5.9%) |

So the run covered a **deterministic but arbitrary prefix**, not the family. The correct statement is: *the emission was neutral across the first 60 cases in manifest order of the `block_access_list` filter, and the two failures it found were within that prefix.* Nothing was established about the remaining 962.

This is the same defect as the `--limit 2` run that reported `fail 0` where the wider run reported `fail 2` — the one I cited approvingly in this very thread as a lesson about population size. I then repeated it one order of magnitude up, and described a truncation as a family because that is what it had been called in conversation.

**Nothing about the change itself is in question.** Its own neutrality argument does not depend on the denominator: the appends fed a container nothing read, so no verdict could move by construction. The overstatement was in my characterisation of the evidence, not in the evidence's sufficiency for what this PR did.

The author disclosed the truncation unprompted when asked for the selector verbatim, which is why it is being corrected now rather than propagating. **The selector belongs in the repo** rather than in conversation — a number whose reproducibility depends on which agent ran it is not a measurement anyone else can check, and this one had been quoted in three places.

Issue #10730, opened 2026-07-27
bal_cross_tx_deploy_then_call fails on main for both CREATE and CREATE2: guest rejects a block the fixture says is valid


pirapira:
## Summary

`bal_cross_tx_deploy_then_call.json` fails on **plain main** in both parametrisations: the guest rejects a block the fixture says is valid. A false reject — a completeness gap, not a soundness hole — on the deploy-then-call path.

```
bal_cross_tx_deploy_then_call.json  create_opcode_CREATE    succ guest=00 exp=01
bal_cross_tx_deploy_then_call.json  create_opcode_CREATE2   succ guest=00 exp=01
```

`FAIL [root/----/tail]` on both. **Both parametrisations failing identically points at the shared create-then-call path rather than at either opcode.**

## Established as pre-existing by a three-leg control, not by assumption

Measured by scoper over 60 fixtures of the `eip7928` BAL family, same population each leg:

| leg | ran | fail | full match |
|---|---|---|---|
| plain main | 60 | 2 | 58 |
| feature stack, emission **not** called | 60 | 2 | 58 |
| feature stack, emission **wired** | 60 | 2 | 58 |

Identical across all three. So these reproduce on main independently of the #10680 storage-emission chain, and that chain is behaviourally neutral across the whole family rather than across the two fixtures it had previously been tested on.

## Why this was nearly misattributed, which is the transferable part

A narrow `--limit 2` run gave **fail 0**; the family gives **fail 2**. Had the emission been reported as passing on the strength of two fixtures, and someone later run the family, the natural conclusion would have been that the new wiring caused these — with no evidence available to the contrary.

The three-leg control is what makes *pre-existing* a measurement rather than an assumption, and running plain main turned out to be **cheaper than investigating** the failures it would otherwise have been blamed for. Worth remembering as a default: when a change is suspected of causing a failure, run the base before hunting the change.

## Character of the defect

A false reject, so no soundness exposure — the guest is refusing a valid block rather than accepting an invalid one. But it is a genuine completeness gap in the BAL family that #10680 is converging, and it sits on the deploy-then-call interaction where a contract is created and then called, which is precisely the shape that exercises cross-transaction code and storage attribution.

Not folded into any feature PR deliberately: it reproduces on main, it is unrelated to the emission, and burying a main-branch failure inside a feature PR is how it stops being visible.

## Suggested next step

[...]

PR #10735, opened 2026-07-27
feat(bal): the storage_reads container and set-insert for the builder (#10680)

pirapira:
`storage_reads` was the one `AccountChanges` field with no builder container: no arena,
no count, no overflow latch and no append, while the other four had the complete set.
This adds them, mirroring the siblings.

**DELIBERATELY INERT PENDING ITS CALLER** — nothing calls
`bal_builder_record_storage_read`. The producer side is already complete
(`storage_read_record` fills `STORAGE_READS_AREA`, `read_sets_incorporate_tx` promotes
it per transaction, deduped and capacity-bounded), so the wiring is a separate change
that will carry the behavioural evidence. This is a planned stage, not an unfed
mechanism.

## A set insert, not an append

`add_storage_read` (`:371-382`) is `ensure_account` then `.add(slot)`, and the spec's
field is `Set[U256]` — so a repeated read of one slot contributes one element. The
routine scans `(address, slot)` and returns without appending on a hit.

**No `block_access_index`.** The four change rows stamp one; a set has nothing to
index. Row is 56 B: `{address[20], pad[4], slot[32]}`.

## It does NOT filter reads that are also written, and that is correct

`add_storage_read`'s docstring states the EIP-7928 rule, but the **implementation adds
unconditionally** — the exclusion is applied in `_build_from_builder` (`:549-552`),
filtering `storage_reads` against `storage_changes` at serialization.

Doing it at insertion would be wrong, not merely premature: a slot can be read in one
transaction and written in a later one, and an insertion-time filter cannot see a write
that has not happened yet. The equivalence holds because nothing reads this container
between insertion and serialization — which the inertness above guarantees today.

## Capacity mirrors the upstream arena

16384, matching `STORAGE_READS_AREA`, so the builder is never the tighter bound: a
narrower cap here would reject blocks the recorder accepted. 16384 × 56 = 917504 B.

- set-insert scan removed, making it an append → **1 fires**
- slot stored at `+32` (the sibling's offset) instead of `+24` → **1 fires**
- unmodified → **0**

Plus static pins: no `bal_builder_current_bai` reference, row 56, capacity 16384, and
the hit path not bumping the count.

## Layout

The arena is `.bss`/NOBITS, so **`.data` did not grow**: `dataSizeBytes` unchanged and
`bnf_le_a` still `0xa3000ee0`, which is the address-pinned `rfl` proof in
`Bn254FieldMulModSAsm` that `.data` growth would break. `.bss` moved to `0x1b85b4a0`,
requiring the manual repin no regen performs — `RegionMap.bssSizeBytes` plus **both**
`GuestImage.lean` literals, the size and the derived `0xa4000000 + bss` (three sites).
`.text` iterated to a fixpoint at `0x064c58`.

Regression: 60 fixtures, errored 0, full match 58, the 2 being #10730's pre-existing
pair. All gates pass.

## One process note

The first build failed with `symbol '.Lbrsr_ret' is already defined` — my label prefix
collided with `bal_rlp_scalar_rlp_len`'s, landed earlier today. Renamed to `.Lbbsr_`.
Worth recording because my usual `grep -cE "^error"` on codegen output **does not match
this failure**: codegen prefixes it with `codegen:`, so the build reported zero errors
while writing a `.s` and no ELF, and every subsequent symbol check ran against a stale
artifact.


pirapira:
Closing per the design note in the file this PR edits.

`BlockAccessListBuilder.lean:18` lists `storage_reads` as "the existing block-lifetime `storage_reads` set", and `:22-26` says storage reads are **deliberately not duplicated** there — `storage_read_record` and `read_sets_incorporate_tx` already implement the spec's block-level set lifetime, and the serializer was designed to walk that set directly, omitting slots also present in `storage_changes`.

So this container is a second source of truth for a set that is already correct upstream: deduped, promoted per transaction, capacity-bounded, and feeding a live comparator. It would need keeping in sync with the promoter for no benefit, at a cost of 0.88 MiB of `.bss` and a three-site `GuestImage` repin.

What survives and is recorded on #10680: `add_storage_read` (`:371-382`) adds **unconditionally**, and the read-versus-change exclusion is applied at serialization in `_build_from_builder` (`:549-552`) — not at insertion. An insertion-time filter would be wrong rather than merely premature, because a slot can be read in one transaction and written in a later one. The walk needs that fact regardless of where the reads live.

PR #10736, opened 2026-07-27
fix: emit BAL builder entries from tx account writes

pirapira:
## Summary

Materialize BAL builder effects from the transaction account-write map immediately before `account_writes_incorporate_tx`. The helper reads the block-cumulative map as the pre-transaction baseline, falls back per component to authenticated parent state, and emits balance, nonce, or code effects only when the post value differs from that baseline.

The transaction map's keyed upsert gives one decision per `(address, block_access_index)`. The validity mask means a producer wrote a component; it does not decide emission. Emission is the comparison against the baseline. Code compares hashes and uses the canonical empty-code hash for authenticated absence.

### Shipping artifact: paired behavioural-neutrality gate

This is deliberately an exact stacked-base versus tip pair, not a current-main comparison: it isolates this change while main advanced during the work.

| artifact | SHA-256 | result |
| --- | --- | --- |
| stacked base (`20a1cbd5`) | `5c35bbf4ec0aee7a…` | FA=0, FR=10, 190 agree, error=0, budget=0, root/tail match=200/200 |
| measured pre-integration tip | `da6594b691b8024e…` | FA=0, FR=10, 190 agree, error=0, budget=0, root/tail match=200/200 |
| current PR head | `7112c7955018f52e…` | FA=0, FR=10, 190 agree, error=0, budget=0, root/tail match=200/200 |

All legs used Spike runner `c6ac09fa…`, captured seed `1040692591`, and the same 200-row random draw. `eest-ab-compare.py` self-checks passed for both the stacked-base/tip and pre-integration-tip/current-head comparisons: both legs scored 200/200 rows; 197 input digests joined, with three valid shared-input groups covering six rows and zero unmatched rows. The false-reject label sets are exactly equal (the same ten labels), with zero status differences and zero output-byte differences across every joined case.

The accidental `--all` run was discarded: it had a different population and is not evidence for this pair.

### Diagnostic-only passes

Separate diagnostic artifacts established that the transaction-boundary walk executes and its counters close. These counters are not shipped. Pass A is SHA-identified and carries an in-band closure assertion rather than a tag: the four lossless fields consume all 64 bits, so shrinking one would reintroduce wrap. Pass B uses a distinct schema tag and reproduces the same rows.

An absence-counter guard exposed an important instrumentation result: its `absences <= misses` invariant fired on nine rows, but the exact-change control showed the counter defect affected 29 rows. The other 20 were wrong while coincidentally satisfying the inequality. Therefore an invariant's firing count is a lower bound, not a defect-size measurement. The root cause is also component-level: block-map hits and misses are per-component under the hit mask, not row properties.

## Coherence

The paired artifacts above were intentionally frozen before integrating current main, so that their one-change comparison remained attributable. The PR head was then merged with current main and regenerated to convergence; this current merged head has `.text=0x64dc8`, `.data=0x5370`, and `.bss=0x1b77b540`.

- `check-region-map.sh`: pass
- `check-asm-to-program.sh`: pass
- paired-tip layout: `.text=0x648fc`, `.data=0x5370`, `.bss=0x1b77b4e0`
- `GuestAddrs.bnf_le_a` remains `0xa3000ee0`


pirapira:
## `merge!` withdrawn pending the six-arm proof — raised by the author, and correct

This PR is not merely wiring: `account_writes_emit_builder_tx` calls `bal_builder_append_balance` (`:330`), `_nonce` (`:339`) and `_code` (`:355`) immediately before incorporate. **So the did-change logic is this PR's active behaviour**, not a future slice's.

That logic is: compare each component's post value against the *running* per-account block-cumulative baseline — `block_state.account_writes` when the address has a prior block entry, `pre_state` otherwise — pinned to `block_access_lists.py:583-598` and `:617-665`, with the comparison made **before** the merge.

### Why the 200-leg cannot substitute

The paired legs are real and were measured on the PR head (`7112c7955018f52e`, `FA=0 FR=10`, identical FR label sets, complete denominators). But **nothing consumes the builder streams yet** — `bal_builder_serialize` does not exist. So the legs establish that the change perturbs no verdict, which is expected *by construction* while the output is unread. They cannot establish that the emitted entries are **correct**, because no code path reads them.

That is a clean gate measuring irrelevance rather than safety, and it is the specific mistake worth not repeating: a per-segment feature was merged earlier today on evidence that exercised only the segment that already worked, and the missing coverage then showed the feature had never worked at all.

### The cost of being wrong lands on someone else

The first consumer of these streams will be the serializer, which is now in design. If the producer emits spurious or missing entries, the failure surfaces as a **hash mismatch in the serializer** — a wrong diagnosis attached to a real failure, in a different change, authored by someone else. That is the expensive kind of latency.

### What is needed — twelve cells, not three

Three components × {changed, touched-but-equal-to-baseline} × {prior block entry, parent fallback}:

- **Build the nonzero-parent-fallback case first.** If the fallback ever yields zero rather than the parent's actual value, an account whose balance was already `X` and stays `X` compares `X` against `0`, reads as changed, and **emits a spurious entry** — extra element, different hash, no complaint from the guest. A fixture where every fallback account happens to start at zero cannot distinguish a correct fallback from a broken one. A zero baseline is an assertion that the account had no prior value, and that class of assumption has been wrong repeatedly today, always in the entry-adding direction.
- **The mask is a touched-field filter, not an emit criterion.** Emitting on the mask records no-op writes; the BAL records changes.
- **`code_changes` compares the hash and emits the bytes** (`:652-664` vs `CodeChange.new_code: Bytes`). Two different quantities in one decision.
- If the corpus cannot produce a cell, **say which and build a probe** rather than quietly covering eleven.

Nothing else about the PR is in question — the merge handling, the pin refresh from the ELF, the coherence gates and the `bnf_le_a` check were all done properly. Re-label once the arms are demonstrated.

pirapira:
Post-merge scope correction from the matrix probe: this change wires BAL builder emission for balance and the transaction sender nonce where their existing producers feed the transaction account-write map. EIP-7702 authorization authorities have no account_write_record producer: their nonce and delegation-code rows are therefore not yet materialized, so append_nonce and append_code cannot fire for authority effects. Follow-up #10741 owns that successful-authorization producer (nonce | code); this PR does not claim the three appenders are all live for every state-change source.

PR #10744, opened 2026-07-27
feat(bal): the nonce-changes measurer (#10680)

pirapira:
Fourth serializer increment. INERT.

`NonceChange` is `[block_access_index, new_nonce]` with **both scalars u64**, so the
widener runs twice per row — the only difference from the balance measurer, and guarded as
such. Widening only the BAI and handing the raw nonce pointer to a 32-byte measurer reads
24 bytes past it.

## The code measurer was lifted out of this increment

It needs `bal_rlp_emit_bytes` and `bal_rlp_measure_into_throwaway` from #10739, which is
not in this branch's history — the byte-string shape has no measurer in this layer, so its
length must come from running the emitter against a discarded sponge. The link said so
outright:

    undefined reference to `bal_rlp_emit_bytes'
    undefined reference to `bal_rlp_measure_into_throwaway'

It lands on a branch based on #10739 rather than being merged in here, so each PR carries
one dependency rather than two.

## Two tooling preconditions added, both from failures in this increment

**The layout-converge loop now refuses to write a pin it cannot read.** When the link
failed, `readelf` returned nothing, and the loop wrote a bare `0x` into `RegionMap` *and*
both `GuestImage` literals — corrupting three sites while reporting `repinned 0x1b7fb4e0->`.
It now aborts unless the ELF exists and both sizes are non-empty, because a build failure
and a zero-size section are indistinguishable to a `readelf | awk` pipeline.

**And `grep -cE "^error"` is not a build check.** Both failures in this increment — the
undefined references and the earlier `.Lbrsr_ret` collision — are reported by codegen as
`codegen: ... failed (exit 1)`, which that pattern does not match. The guest then keeps a
stale ELF while the `.s` is fresh, and every subsequent measurement is against the old
artifact.

Layout: `.bss` moved to `0x1b7fb5c0`, repinned in `RegionMap` and both `GuestImage`
literals. `.text` iterated to a fixpoint. Regression: 60 fixtures, errored 0, full match 58
— the 2 being #10730's pre-existing pair.


pirapira:
## Closing as superseded by #10752

#10752 (`feat/bal-serializer-emit-10680` → `main`) is the stack tip and carries this PR's content. Verified on its head rather than assumed:

- all five field measurers present (`measure_balance`, `measure_nonce`, `measure_code`, `measure_storage`, plus `measure_reads`)
- `bal_serializer_measure_account`, the whole-account measurer
- the u64 widener
- **21 emission guards** keyed on the emitted string
- the surviving-reads scratch removed, with **three guards asserting it cannot return** — `balSerializerMeasureReadsFunction`, `blockAccessListBuilderDataSection` and `balSerializerFilterReadsFunction` each pinned to zero occurrences

Closing rather than merging in sequence, so this family has one live branch instead of three. Three open members is what let batch #10745 snapshot a superseded state and merge branches whose content had been emptied — with a green build, because the guards were keyed on the identifiers that vanished with the routines. Fewer live members is the cheapest defence available while a pulled `merge!` does not reliably hold a batch (#10747).

The branch is untouched; only the PR is closed. Nothing here is a review objection.

PR #10745, opened 2026-07-27
chore: batch-merge #10738 #10739 #10740 #10742 #10743

pirapira:
Batch-merge of #10738 #10739 #10740 #10742 #10743 (all author pirapira). Verified locally: `lake build` + all `check-*.sh` green.

All 25 `scripts/check-*.sh` pass, including `check-axioms`, `check-guarded-handler-bytes`, `check-build-parallel`, `check-region-map`, `check-memorylayout-region-coverage` and `check-statement-tamper`.

#10738 -> #10740 -> #10742 are a stack (each head verified to be an ancestor of the next), merged in ascending PR order alongside #10739 and #10743.

Conflict resolution (program layout only; no statement or definition changed):

- Every conflict was confined to the four generated/pin files (`GuestAddrs.lean`, `RegionMap.lean`, `symbol-addresses.tsv`, `GuestImage.lean`); no branch had a non-size `RegionMap` delta vs main, so those were resolved from main and regenerated. Verified afterwards that `RegionMap`'s `account_writes` bases still match the `MemoryLayout` constants and that `GuestImage`'s bss pins are internally consistent.
- `textSizeBytes` `0x064e40` -> `0x065348`; `bssSizeBytes` `0x1b77b540` -> `0x1b7fb5a0` with `GuestImage`'s t7/t7' pins repinned to `0xbf7fb5a0`; `.data` unchanged at `0x5370`.
- `symbol-addresses.tsv` (3172 rows), `GuestAddrs.lean`, `GuestImageEntries.lean` regenerated to fixpoint.


pirapira:
## `merge!` withdrawn — this batch was formed from a state that is being rewritten right now

#10738 and #10739 are currently `DIRTY` because their author is **mid-rebase on the base of the stack**, removing a 524,288-byte `.bss` allocation (the surviving-reads scratch). This batch contains the **pre-rebase** version, i.e. the allocation we have just agreed to give back.

Merging it would land 512 KiB of scratch and then require a second PR to delete it — churn, plus a window in which the joint `.bss` budget is tight for no reason.

### Why the allocation is going away

The scratch existed to hold the filtered survivor list so that the measure and emit passes could not disagree. That justification does not hold: **the hazard was two implementations of one rule, not two passes.** Re-running a single predicate (`bal_serializer_slot_written`) cannot diverge from itself, and the surviving *count* — the only thing the header needs — comes from the filter's return value. So nothing needs materialising.

It is also strictly safer without it: a 16384-slot buffer **indexed by survivor count** can be written past if the filter ever yields more survivors than it holds. No buffer, no overrun.

### The budget this protects

```
.bss headroom to .sszscratch      1,199,456
serializer scratch (being dropped)  524,288
7702 delegation arena (#10741)      614,136
```

With the scratch: 5.1% free. Without it: **48.8% free**. Two lanes are allocating into this region in separate branches — each links cleanly alone, and the overflow would appear only in the merged tree, as the `.sszscratch` VMA overlap already hit once tonight.

### What to do

Re-batch after the rebase lands. The members are all still wanted — #10738, #10739, #10740, #10742 are the serializer increments, all inert with zero callers, none able to change acceptance until the walk is wired. Nothing here is a review objection; this is purely a staleness hold.

General note for batching: **a batch PR is the unit of merge, so a hold applied to a member does not reach it.** That decoupling merged #10736 tonight after its label was pulled, and it is the same mechanism in the other direction here.

Issue #10747, opened 2026-07-27
merge-bot: enforce merge! hold semantics across batched PRs


pirapira:
## Two holds escaped via batch PRs tonight; a pulled `merge!` is not a reliable hold in either direction

Recording this as an operational defect rather than continuing to work around it, because I have now relied on the label-based hold twice and it failed both times, in opposite directions.

### What happened

**Case 1 — a hold on a member did not reach the batch.** I withdrew `merge!` from #10736 pending evidence for the did-change logic it added. It merged 22 minutes later **via batch #10737**, which carried its own `merge!` and its commits.

**Case 2 — a hold on the batch itself did not stop it.** #10745 batched #10738/#10739/#10740/#10742. I withdrew `merge!` from **#10745** at ~22:37, because its members were mid-rebase: their author was removing a 524,288-byte `.bss` allocation, so the batch had snapshotted a superseded state. #10745 merged at **22:48:46** anyway, and `bal_serializer_read_scratch` is now on `main` (`BlockAccessListBuilder.lean:185`) — the allocation that was being removed.

Neither case caused damage. Case 1's gap was pre-existing and inert (#10741). Case 2 wasted 512 KiB against 1,591,904 bytes of remaining `.bss` headroom, so nothing overflowed. But in both cases the hold was placed for a reason and had no effect.

### Why it matters beyond these two

`.bss` has ~1.5 MiB of headroom before it collides with `.sszscratch`, and two lanes are allocating into it concurrently. Each branch links cleanly alone; only the merged tree overflows. **Sequencing is the only defence, and sequencing depends on a hold that holds.** Tonight the linker caught one overflow already.

### Suggested fixes, cheapest first

1. **Re-check member labels at merge time.** If any member of a batch has lost `merge!` since the batch was formed, do not merge the batch.
2. **Invalidate a batch whose members have moved.** If a member's head SHA differs from what the batch captured, re-form rather than merge — that is exactly case 2.
3. **Treat a pulled label on the batch as authoritative** even if checks are already running, and stop the run.

Any one of the three closes both cases; (2) also prevents merging a snapshot the author has already superseded, which is the more insidious failure since the PR looks current.

### Interim practice

I will sequence *before* work starts rather than trying to stop machinery afterwards, and treat my own holds as advisory. Worth flagging to whoever owns merge-bot's configuration, since the fix is on that side rather than in the PRs.

pirapira:
[...]

PR #10748, opened 2026-07-27
feat(bal): the storage_changes measurer, plus recovery of content a rebase destroyed (#10680)

pirapira:
The deepest field measurer, and the repair of two defects that let seven routines vanish
without a single check failing.

## storage_changes: three header levels

`Tuple[SlotChanges, ...]`, `SlotChanges = [slot, changes]`, `changes` a tuple of
`StorageChange = [bai, new_value]`:

    encoded(SlotChanges) = hdr(p2) + p2
    p2 = scalar(slot) + hdr(p3) + p3
    p3 = Σ over changes of ( hdr(p4) + p4 )
    p4 = scalar(bai) + scalar(new_value)

Three headers below the field list, against one for balance and none for reads. Dropping
any level leaves every intermediate a well-formed RLP list of the wrong length.

The builder stream is flat, so grouping is this routine's job: a row is measured only if no
EARLIER row of the same account carries its slot, matching `_build_from_builder:537-542`.
That is O(n²) per account, paid once, because sorting into a copy needs `.bss` the guest
does not have.

Controls, each observed firing: level-3 header dropped, inner loop stops re-checking the
address, seen-before scan looks at all rows instead of earlier ones.

## Two vacuities that hid seven missing routines

**Defined and never concatenated.** Seven routines were absent from
`blockAccessListBuilderFunctions` and therefore from the guest, while every guard passed —
because each guard examines a routine's OWN string rather than the emitted one. A routine
can be fully correct, fully guarded, and not exist. Fixed by eleven **emission** guards keyed
on the emitted string and each routine's label.

**A guard that names what it guards dies with it.** Resolving the scratch-removal rebase, a
blanket `git checkout --theirs` over the conflict set discarded the balance and nonce
measurers — and their guards, which reference the function identifiers, went with them. The
build stayed green because nothing was left to fail. Measured: `measure_balance=1` at
`HEAD~1`, `0` after the merge, on two branches whose PR bodies described the work.

Recovered from the pre-merge commits and verified emitted. This is a distinct failure class
from every other vacuity here: not a check that runs and cannot discriminate, but a check
removed by the very edit it exists to catch. Identifier-keyed guards are self-deleting;
emission guards keyed on the output string are not.

Also recovered: `bal_serializer_u64_field`, the widener's destination, which the same merge
dropped. A routine referencing a missing DATA symbol builds fine in Lean and fails only at
link, naming the symbol rather than the routine — so the reservation is now guarded beside
the routine that needs it.

Layout: `.text` and `.bss` converged, `GuestImage` repinned. Regression: 60 fixtures,
errored 0, full match 58 — the 2 being #10730's pre-existing pair.


pirapira:
## Closing as superseded by #10752

#10752 (`feat/bal-serializer-emit-10680` → `main`) is the stack tip and carries this PR's content. Verified on its head rather than assumed:

- all five field measurers present (`measure_balance`, `measure_nonce`, `measure_code`, `measure_storage`, plus `measure_reads`)
- `bal_serializer_measure_account`, the whole-account measurer
- the u64 widener
- **21 emission guards** keyed on the emitted string
- the surviving-reads scratch removed, with **three guards asserting it cannot return** — `balSerializerMeasureReadsFunction`, `blockAccessListBuilderDataSection` and `balSerializerFilterReadsFunction` each pinned to zero occurrences

Closing rather than merging in sequence, so this family has one live branch instead of three. Three open members is what let batch #10745 snapshot a superseded state and merge branches whose content had been emptied — with a green build, because the guards were keyed on the identifiers that vanished with the routines. Fewer live members is the cheapest defence available while a pulled `merge!` does not reliably hold a batch (#10747).

The branch is untouched; only the PR is closed. Nothing here is a review objection.

Issue #10746, opened 2026-07-27
regen-cycle.sh cannot run (remap_sasm.py absent, dead default scratch path, failures swallowed) and the cycle costs up to 4 full Lean builds


pirapira:
### 1. `scripts/regen-cycle.sh` is broken as checked in, and fails silently

`:8` — the scratch directory defaults to a **specific past session's scratchpad**:

```
S="${REGEN_SCRATCH:-/tmp/claude-1000/-home-zksecurity-evm-asm2/e32e8676-b919-44c3-8483-c15720533cbd/scratchpad}"
```

That path is **absent** on this host. Worse, the remap step at `:22` invokes a script that lives there:

```
python3 "$S/remap_sasm.py" "$S/old-symbols.tsv" scripts/asm-fixtures/symbol-addresses.tsv $(cat "$S/todo.txt") | tail -2
```

**`remap_sasm.py` is not in the repository at all** — `find . -name remap_sasm.py` returns nothing. So the offset-aware SAsm literal remap, which is the whole point of the loop, cannot execute.

**And the failure is quiet.** The script sets `set -uo pipefail` but **not `-e`**, and the remap's output is piped through `tail -2`, so a missing interpreter target produces two lines of stderr and the loop continues. It then re-runs `lake build`, finds the same failures, and either re-attempts or exhausts its three passes and falls through. Nothing reports "the remap never ran."

The header confirms this was never meant to be the standing tool: *"Session helper for the v0.6.0 migration (evm-asm-0w05f)."* It has since become the thing agents reach for.

This is the same defect class as #10720, where `check-region-map.sh` told readers to regenerate with a script that never writes `RegionMap.lean`. The *documentation* of regen was wrong there; here the *script* is.

**Minimum fix:** commit `remap_sasm.py` into `scripts/`, default `REGEN_SCRATCH` to a `mktemp -d`, and make the remap step fail loudly — check the exit status rather than piping it into `tail`.

### 2. The cycle runs up to four full `lake build`s

One upfront (`:10`) plus up to three in the remap loop (`:16`), each a full build of a tree whose generated address table is imported by **141 of 2788 modules**. Observed tonight: `textSizeBytes` took two passes to settle (`0x064ad8` → `0x064ae8`), and a single cache-disabled link ran about ten minutes.

The script explains why the passes exist:

> *"converging in up to 3 remap passes (later waves of drifted files only surface once their import prerequisites build)"*

[...]

PR #10752, opened 2026-07-27
feat(bal): restore the four measurers and the u64 widener, plus the emission guards (#10680)

pirapira:
This PR is **restoration as well as new work.** Before it lands, `origin/main` (aff464653) is missing four of the five field measurers and the u64 widener. Measured on main directly, not inferred:

| symbol | on main | here |
|---|---|---|
| `bal_serializer_measure_balance` | MISSING | present |
| `bal_serializer_measure_nonce` | MISSING | present |
| `bal_serializer_measure_code` | MISSING | present |
| `bal_serializer_measure_storage` | MISSING | present |
| `bal_serializer_u64_to_field` (widener) | MISSING | present |
| `bal_serializer_measure_reads`, `filter_reads`, `bal_rlp_emit_bytes`, `bal_rlp_measure_into_throwaway` | present | present |

Main's build is green with those four absent, because the guards were keyed on the identifiers that went with them — identifier-keyed guards are self-deleting. **The guards land in this PR alongside the routines they guard, not after.**

## EMISSION-GUARD COUNT: 5/5

Measured in the emitted guest (`grep -c '^<symbol>:' stateless_guest.s`), not in the module source and not in a routine's own definition string — that distinction is the whole point, since a routine can be defined, fully guarded, and never concatenated into the guest:

```
bal_serializer_measure_storage   PRESENT
bal_serializer_measure_reads     PRESENT
bal_serializer_measure_balance   PRESENT
bal_serializer_measure_nonce     PRESENT
bal_serializer_measure_code      PRESENT
===> 5/5 measurers present in the emitted guest
```

Plus 6/6 supporting routines: `measure_slot`, `measure_account`, `u64_to_field`, `filter_reads`, `bal_rlp_emit_bytes`, `bal_rlp_measure_into_throwaway`.

## A real bug fixed on the way: the widener wrote the wrong byte order

`bal_serializer_u64_to_field` stored the u64 LSB at field byte 31, under a comment calling that big-endian. Every consumer reads the field **little**-endian: `bal_rlp_scalar_len` scans downward from byte 31 for the *most* significant byte, and `bal_rlp_emit_scalar` emits field byte `len-1-i` at BE output index `i`.

Both docstrings say "BE" about byte 31 and mean opposite layouts, which is why it survived on both sides — the prose agreed and the bytes did not.

For `block_access_index = 1`: the field got `0x01` at byte 31, `scalar_len` reported **32** significant bytes, and `scalar_rlp_len` returned **33** instead of 1. Every storage change over-measured by 32 bytes, and the emit pass would have absorbed a 32-byte string where the spec has a single `0x01`. Fixed to a single `sd` at offset 0; two guards pin the store and forbid the `31 - i` index shape from returning.

## Shared slot measurement

`bal_serializer_measure_slot` is factored out and returns **two** payload lengths (SlotChanges, inner changes-list). The emit pass needs both to write the nested headers and cannot get either from the length table — one entry covers the whole field, and the per-slot change count is unbounded. The alternative was a second emit-side copy whose only failure symptom is a wrong digest. The four nesting guards are keyed on the **pair** of functions, so the refactor that moves the content cannot also silence the guard.

## On the requested `.bss` check

The 524,288-byte read-scratch drop **is not visible in this PR's diff, and should not be** — that removal is already on `origin/main` (main's only `bal_serializer_read_scratch:` occurrence is the guard asserting its absence). Confirmed absent from the artifact here: 0 occurrences in the emitted guest. This PR's `.bss` moves *up* 256 bytes vs main, not down. Reporting the expected −524,288 would have been reporting a number I did not measure.

## Verification

- layout iterated to a fixpoint by hand (`text=0x0658b4`, `bss=0x1b77b680`), GuestImage bss literals repinned; regen run manually, since `regen-cycle.sh` reports `REGEN_CLEAN pass=1` unconditionally when its scratch path is missing (#10750)
- `check-region-map`, `check-asm-to-program` (390 defs / 148 files), `check-forbidden-tactics`, `check-file-size` all clean
- 58/60 on the BAL fixture corpus; the two misses are `bal_cross_tx_deploy_then_call` CREATE/CREATE2, both `bv_fail=41` (the documented receipts/gas false-reject family, `header_gas_used=388620` vs `expected=290700`), state roots matching and not attributable to this stack — the only reject path these commits add emits code 58

## Known limitation

The measure pass is **inert** (`measure_account` has zero callers) and is validated by string guards only. Separately: `bal_rlp_encode_selftest` and `bal_canonical_sort_selftest` have zero callers and no runner, and no script executes any `zisk_bal_*` probe unit — `check-build-units-link.sh` links them without asserting a result. So nothing in this layer has yet been *executed*. An executing probe is the next PR; it is what the digest comparison will rest on.

#10744 and #10748 can close as superseded — this tip carries their content.


pirapira:
## Review — approving on substance, two asks before it lands

The measure pass is complete and the commit's stated content is the right content. Reviewed against tonight's design decisions rather than in isolation.

### What is right, and worth naming because it is the part that would have gone wrong

**The payload-versus-encoded conversion is the whole hazard, and you identified it unprompted and pinned it.** Summing six table entries directly *is* the natural thing to write, and the account header would come out short by exactly the five field headers, **with no symptom until a digest comparison** — no malformed buffer, no failing decode, just a different hash. That is the highest-value line in the diff.

**The code measurer uses the throwaway-keccak route, reading ptr/len at `32/40(s4)`, matching the live `bal_builder_append_code` call site.** That is the right pairing: the code blob is a byte string, so its encoded size is not derivable from a field width, and the throwaway route keeps the emitter as the single implementation of the length rule.

**21 bytes for the address is correct** (`0x94` + 20).

**Controls perturbing exactly one emission site each, with the guards-installed precondition**, so a null control run cannot read as a pass. That precondition is your own fix from #10739's failure, now doing work.

**Inert with zero callers on `measure_account`**, so no verdict can move — and the 58/60 corpus result matches the base.

### Ask 1 — state the emission-guard count as a number

The body says two guards pin the conversion. What it does not state is **how many of the five measurers were found present in the emitted guest.** That is the check that would have caught the branch wipe, and main is currently the proof of why it matters: four measurers absent, build green, because the guards were keyed on identifiers that vanished with the routines.

Put the number in the body — "5 of 5 measurer labels found in the emitted guest" — so it is asserted rather than implied. Same for the widener's data reservation.

### Ask 2 — prefer parity over a count where it is cheap

`six bal_rlp_list_header_len calls` and `six bal_serializer_measure_ calls` are **count guards**, and a count is a number a future author restores green by bumping. Where the relationship is structural — every measurer call has a matching header-length conversion — a **parity** guard (each measure site has a conversion, and each conversion has a measure site) has nothing to silence and no number to adjust. Not blocking, and the controls you ran mitigate it; worth doing if it is a small change.

### Bonus finding in your own numbers, which I am acting on separately

Your corpus run diagnoses the two pre-existing misses far better than #10730 records them: **`bv_fail=41`, the receipts/gas false-reject family, `exact_header_gas_used=388620` vs `exact_expected_gas_used=290700`.** #10730 was filed as *undiagnosed*, precisely and deliberately. That is a real advance on it and I am updating the issue with it.

Also confirming from the body: headroom below `.sszscratch` is now **2,116,160 B**, up from 1,591,904 — a ~524 KiB gain, which is the scratch removal reaching the artifact. That is the check I asked for, arriving as a side effect of the layout report rather than as an explicit statement. Stating it directly would be better.

### Not blocking

The conflict against main is generated-only — `GuestAddrs.lean`, `GuestImage.lean`, `RegionMap.lean`, `symbol-addresses.tsv` — with **no hand-written file in the conflict set**, so this resolution carries none of the risk that cost two branches earlier. Take main's version as a buildable bootstrap and regenerate to convergence; do not drive it through `regen-cycle.sh`, which reports `REGEN_CLEAN pass=1` unconditionally when its scratch path is missing (#10746, removal in #10750).

Issue #10754, opened 2026-07-27
The BAL serialization self-tests have never executed: zero callers, and check-build-units-link.sh asserts linkage rather than behaviour


pirapira:
## Summary

The BAL serialization layer's self-tests **have never executed**. They are written, well-designed, and unreachable:

| probe | callers |
|---|---|
| `bal_rlp_encode_selftest` (15 per-case assertions) | **0** |
| `bal_canonical_sort_selftest` | **0** |

No script runs any `zisk_bal_*` probe unit. `check-build-units-link.sh` **links** each build unit and fails on a *new link failure* — it does not run them or assert a result. It is reachable (via `check-build-parallel.sh`), so this is not an orphaned script: **it is a linkage gate standing where a behaviour gate is assumed.**

Consequence: the whole measure pass is currently validated by `#guard`s that pin *emitted structure* — that a routine's label appears in the guest, that a call count matches — and by nothing that pins *behaviour*.

## What that gap produces, concretely

A real bug in `bal_serializer_u64_to_field`, found by **reading** rather than by any test:

- it wrote the u64's LSB at **field byte 31**, under a comment calling that big-endian;
- every consumer reads the field **little-endian** — `bal_rlp_scalar_len` scans *downward* from byte 31 for the most significant byte, and `bal_rlp_emit_scalar` emits field byte `len-1-i` at BE output index `i`;
- **both docstrings say "BE" about byte 31 and mean opposite layouts**, so the prose agreed while the bytes did not, and it passed on both sides.

For `block_access_index = 1` the field got `0x01` at byte 31, `scalar_len` reported **32** significant bytes, and `scalar_rlp_len` returned **33 instead of 1**. Every storage change over-measured by 32 bytes, and the emit pass would have absorbed a 32-byte string where the spec has a single `0x01`. Fixed in #10752, with the store pinned and the `31 - i` index shape forbidden from returning.

**A 15-case vector would have caught this on case one.** It exists. It has never run.

## Why this matters beyond the one bug

The eventual correctness argument for #10680 is a **digest comparison** — hash the rebuilt BAL, compare against the declared header field. That is congruence rather than inference, and it is the right design. But a digest comparison gives one bit: equal or not. When it says *not*, the only things standing between that bit and a diagnosis are the layer's own executing tests, and there are none. Every encoding-level defect will present identically, as a hash mismatch.

[...]

PR #10755, opened 2026-07-27
feat(bal): the emit pass end to end, checked by digest, every path executed (#10680)

pirapira:
The serializer's emit half, plus the probes that execute it. **Every path in this stack has now run** — nothing is covered by string guards alone.

## Runner output, verbatim

```
  ok    case 1  one change           = 6
  ok    case 2  same slot twice      = 9
  ok    case 3  two distinct slots   = 12
  ok    case 4  other address present = 6
  ok    case 5  two-byte value       = 8
  ok    measure_slot payload         = 5
  ok    measure_slot inner payload   = 3
  ok    case 9  cross-tx read exclusion = 1
  ok    case 6  emitted digest              = 31597f4956266576...
  ok    case 7  whole-account digest       = b790ac5342ebff17...
  ok    case 8  outer-list digest (2 accts) = e74830bb60be86e7...
  ok    case 9  emitted reads digest        = 60811857dd566889...
  ok    case 10 sort-then-rebuild digest    = 0xe786be60bb3048e7
==> PASS: 8/8 lengths and the emitted digest match the RLP derivation
  ok    bal_rlp_encode_selftest      PASS (15 cases)
  ok    bal_canonical_sort_selftest  PASS (3 row sets)
==> PASS: 2/2 self-tests pass
```

## Row-layout constraint for anyone adding a sortable array

**Any row array that will be sorted must have an 8-aligned stride.** `bal_canonical_sort` swaps rows with `ld`/`sd`, and per **AGENTS.md:211** the verified RV64 semantics give `LD`/`SD` no semantics off an 8-boundary (`isValidDwordAccess = isValidMemAddr && isAligned8`) — so a proof reaching a misaligned access cannot close. This is why the account row is **24 bytes for a 20-byte address**: 24 is the smallest 8-aligned stride that holds it, and the 4 padding bytes are load-bearing.

The sort's ABI does not say this. It documents `a2 = row stride in bytes` with no constraint, and its comment about using `mul` "so the routine is not silently wrong for a non-power-of-two stride a future caller passes" reads as an invitation to pass any stride. Violating the real precondition faults rather than returning one of its five documented status codes.

**Caught on spike; ziskemu would not have caught it.** AGENTS.md:220 warns ziskemu tolerates unaligned reads at runtime, so the 20-byte layout would likely have passed under it — a green test certifying a layout the verified semantics reject. Clean A/B: same sort, same descriptor, only the stride varied — 32 sorts and returns the expected order, 20 faults at the swap.

24 rather than 32 because the arena is capacity 76923: +307,692 bytes and it fits; a separate 32-byte staging copy needed 2,461,536 and did not. Headroom below `.sszscratch` is 1,193,568 B.

## What executes, and the control for each

| claim | control that made it fail |
|---|---|
| measure pass, 7 lengths | pre-fix widener → 7/7 fire, case 1 reads 38 = 6+32 as predicted |
| slot dedup | `slot_seen_before` removed → **only** case 2 fires, at exactly 18 |
| field digest | changes-list header dropped → digest moves to `b10f6918…` |
| whole-account digest | an empty-list `0xc0` dropped → `af14417e…` |
| whole-BAL digest (long-form `f8 42`) | — two accounts specifically to exercise the long-form branch |
| cross-tx reads exclusion | write moved slot 7→9 → payload 2 and `keccak(07 0b)`, both hand-derived |
| the sort runs | descending seed must yield the ascending digest; in-order seeding passes either way |
| RLP self-test | one expected length 3→4 → returns 104, failing at case 4, the case perturbed |

All expectations are **hand-derived from the RLP rules**, never captured. A golden digest taken before the widener fix would have enshrined the wrong answer.

## Two probe defects found by the probes disagreeing

The self-test probe once **reported two passes while faulting** — the sort self-test needs an arena in `a0`, got the previous test's return value (0 on pass), stored to address 0, and both slots read back as the untouched buffer. Under a zero-means-pass encoding that is byte-identical to two passes. Slots are sentinel-seeded now, and a surviving sentinel reports NEVER RAN.

Case 10 first disagreed because case 9 left `storage_reads_count` at 2 and the cases share globals — cross-case state leakage, found by the digest disagreeing rather than by inspection.

## Not wired, deliberately

`bal_serializer_verify` is inert. Wiring it is the first read of the builder streams, so every neutrality number measured over unconsumed output expires and it owes a fresh paired A/B; and the producer is incomplete (#10757, #10730), so those fixtures would fail for producer reasons with one bit of signal. Local gate, no graceful-skip path — a missing emulator fails.

Behaviour half of #10754. Stacked on #10752 (merged).


pirapira:
Not merged this round — **`check-file-size.sh` fails on this branch's own content**, independent of any batching.

```
FAIL    1759 / 1500 lines  EvmAsm/Codegen/Programs/BlockAccessListBuilder.lean  [codegen]
```

Line counts: `origin/main` 1462, this branch 1759, merged 1759 — so the cap breach comes from the three commits added after #10758 was cut (`6a0af82b2`, `f42794e6e`, `8a46869ed`, +297 lines to that file), not from the merge. CI would reject it the same way on its own.

The remedy is to split `BlockAccessListBuilder.lean` under the 1500-line `EvmAsm/Codegen/Programs/**` cap, which means moving definitions across modules and fixing up imports — a structural refactor, past what an automated merge should decide. Leaving it to you.

For the record, everything else on the merged branch is green locally: `lake build` exit 0 and the other 24 `check-*.sh` pass, including `check-axioms`, `check-region-map`, `check-memorylayout-region-coverage`, `check-build-units-link`, `check-statement-tamper` and `check-progress`. Layout after regen would be `textSizeBytes` `0x065b58` -> `0x0661d0` with `.data`/`.bss` unchanged (TSV 3191 rows).

I did not push the main-merge to this branch, since it would spend a CI cycle on a run that fails at the same gate. Note also that #10758 did **not** land this PR: these three commits were pushed after that batch was cut, so it stayed open deliberately.

pirapira:
## Scope correction — do not read the green digests as a complete serializer

Re-checking against `execution-specs` after the last push turned up two gaps that the probes in this PR **cannot** observe. Recording them here so the passing runner is not mistaken for coverage.

**1. Six of seven orderings are missing.** `_build_from_builder` (`block_access_lists.py:539-579`) sorts: per-slot changes by `block_access_index`; `balance_changes`, `nonce_changes`, `code_changes` each by `block_access_index`; slots by slot; `storage_reads` by value; and the list by address. **Only the last is implemented** — measured: zero of the seven emitters contains a `bal_canonical_sort` call.

Why every digest still passed: **a one-element list is sorted by definition.** Case 7 is one account / one slot / one change, so no inner ordering is observable; case 8 has two accounts, testing the address sort and only that. The expectations were right and the controls fired; the *data* could not discriminate.

**2. A byte-order mismatch between producer and consumer.** `bal_emit_storage_changes` reverses address and slot before recording — the builder row holds **BE20 / BE32**, matching `balSortBuilderStorageSegments`. But `measure_slot` and `emit_storage` pass `row+32` to `bal_rlp_scalar_rlp_len` / `bal_rlp_emit_scalar`, which read **LE limbs** (`BalRlpEncode.lean:375`: "low 20 bytes of the LE word"). On a BE slot, `scalar_len` finds the least significant byte. Wrong length, wrong bytes, for any non-palindromic slot.

The probes cannot see this **because they seed the rows themselves**, in the consumer's convention. The digests are internally consistent and jointly wrong.

**Scope of 2 is not yet established.** Corrected per-producer measurement (my first attempt used an unbounded `awk` range and was wrong): `append_balance` 0 reversals, `append_nonce` 0, `append_code` 1, storage 2. So conventions are **not uniform across the four streams**, and "no reversal in the appender" pushes the question to the callers rather than settling it. Auditing that is caller-side, per stream, per field — not done.

### What this PR is, accurately

An **encoder** test: 8 lengths, five digests, both self-tests, seven controls — every one a statement about the encoder *given rows in the convention the probe chose*. That is worth having and it is not evidence the serializer produces the correct BAL.

Wiring the comparator on this would produce a hash mismatch carrying one bit and pointing at nothing. Six sorts, one confirmed byte-order defect, and three unaudited streams come first.

pirapira:
Re-checked after the latest push (title now "the emit pass end to end, checked by digest, every path executed"): **still blocked on `check-file-size.sh`, and the gap has widened.**

`EvmAsm/Codegen/Programs/BlockAccessListBuilder.lean` is now **1918** lines against the 1500-line `EvmAsm/Codegen/Programs/**` cap — up from 1759 when I first reported it. `origin/main` is 1462.

Unchanged conclusion: the fix is to split that file, which is a structural refactor rather than merge repair, so I'm leaving it to you. Keeping `complicated-merge` on.

Meanwhile #10760 was verified locally (all 25 gates green) and is set to auto-merge, so this PR is not holding up the rest of the queue.

pirapira:
## Two byte-order defects fixed, and the second had real production impact

**1. The storage slot was read with the wrong byte order.** The row holds it BE32 (`bal_emit_storage_changes` reverses it on append); `measure_slot` and `emit_storage` handed `row+32` to `bal_rlp_scalar_rlp_len`/`emit_scalar`, which are documented for LE limbs (`BalRlpEncode.lean:375`). Fixed by reversing into scratch — not by adding a BE variant of the scalar pair, since a second implementation of one encoding rule can agree with the first by construction rather than by correctness.

**2. `bal_serializer_slot_written` compared across conventions dword-wise** — the exec-log's LE stack-word slot against that BE row slot. A dword compare of BE32 against LE32 matches only **byte-symmetric** values:

```
slot 0   BE 00…00   LE 00…00   MATCHES
slot 1   BE …00 01  LE 01 00…  no
slot 7   BE …00 07  LE 07 00…  no
```

So the reads exclusion fired for **slot 0 and essentially nothing else**. Every non-symmetric slot that was read *and* written stayed in `storage_reads` — an extra entry the spec drops. A correctness defect in the emitted BAL, not a test gap.

### The two errors were cancelling

Before fix 1, both sides read the row as LE — both wrong, and therefore *agreeing*. `slot_written` matched correctly; only the encoding was wrong. **Fixing the encoder broke the agreement and made case 9 fail**, and the naive reading is that the fix caused it. Without the field table declaring what `row+32` holds, that post-fix disagreement is just two plausible readings arguing — and reverting the correct fix would have restored the green test.

### What actually caught it

Not reasoning — I had scoped this as "one field, two call sites" and classified `slot_written` as unaffected because it compares row-to-row. It doesn't: it spans an exec-log row and a builder row, which is exactly a convention boundary. What caught it was **moving the probe's seeded slot from byte 32 to byte 63**, i.e. seeding the way the producer writes. Case 9 failed on the next run.

**The digests are unchanged** — same slot values, now stored as the producer stores them, still `31597f49…` / `b790ac53…` / `e74830bb…` / `60811857…`. A wrong reversal would have moved them.

### Also landed: the builder row field table

16 fields across five streams — offset, width, byte order — headed by the rule they follow: *canonical BE20 arrives from the producer*, so rows sourced from an EVM stack word are reversed on append and rows sourced from an account-write row are not. The storage row is called out as carrying **both** conventions: slot BE32 at +32, value LE at +64, four dwords apart, previously both handed to the same LE routine.

### Still open

Six of seven orderings remain unimplemented (`block_access_lists.py:539-579`), and the end-to-end case that drives `bal_emit_storage_changes` instead of seeding rows is not yet built. Until it is, every convention claim here rests on probes that seed their own rows.

pirapira:
## Verified recipe for the end-to-end producer→serializer case

Recording this in the PR rather than leaving it in a chat thread, because it is the construction that would have caught **both** byte-order defects fixed here, and it does not exist yet.

### Why it looked blocked and is not

The producer's *miss* path calls `slot_at_header_state_root`, which needs globals a standalone probe cannot supply. But the **block-container scan runs first**, and a hit jumps straight to `.Lbesc_have`, skipping the state-root lookup entirely. So a probe that arranges a **hit** never reaches the unsupportable path. Confirmed by reading the emitted branch, not assumed.

### The recipe, verified against the emitted code

1. **tx rows** at `0xa21a0000`, count in `tx_storage_writes_count`. The row prefix is the EVM **stack-word** form: address LE word at `0..31`, slot LE word at `32..63`. *This is the layer whose LE→BE reversal is the thing under test.*
2. **block container**: one entry whose **first 64 bytes equal the tx row's first 64 bytes**. The scan is eight dword compares over `0..63`, then `addi s5, t5, 64` for the value. Set `storage_writes_count = 1`, which makes the scan **hit**.
3. Call `bal_emit_storage_changes` with `a0 = block_access_index`.
4. Then `bal_serializer_measure_account` and `bal_serializer_emit_account` as case 7 does, and compare the digest against the **hand-derived** RLP.

### Two constraints that make the case discriminate

**Use a non-symmetric slot.** A dword compare of BE32 against LE32 matches only byte-symmetric values, so **slot 0 passes under the defect** while slot 7 or 11 exposes it. That is why the pre-fix `slot_written` comparison matched only palindromic slots.

**Derive the expected digest before the case runs.** This matters more here than anywhere else in the suite, because the **producer is in the loop** — a captured digest would enshrine whatever the producer does, *including a reversal bug*. Every other case in this file derives its expectation from the RLP rules; this one must too.

### What it covers that nothing else does

Every existing case **seeds the builder row directly**, and therefore writes it in whichever convention the test author chose — so the producer's real convention is never exercised. That is why two byte-order defects survived a suite with eight lengths, five digests, both selftests and seven controls: the tests were correct, the expectations were correct, the controls fired, and the input never travelled through the producer.

pirapira:
Re-checked at `9931e8a76`. You removed `complicated-merge`, but the branch still cannot merge — I'm re-adding it as the queue's "blocked" signal, not to overrule you. Say the word and I'll do the split myself.

```
FAIL    2019 / 1500 lines  EvmAsm/Codegen/Programs/BlockAccessListBuilder.lean  [codegen]
```

The file has grown on each of the last three pushes — 1759 -> 1918 -> 2019 — so the gap is widening rather than closing. The branch is otherwise up to date with main.

I did not do the split unprompted for two reasons: you are actively pushing to this branch (three heads in the last hour), so a ~1000-line module move would collide with your next push; and where to cut is a decision about your module layout, not a merge repair.

**Concrete cut if you want one.** The `balSerializer*` cluster is self-contained and contiguous — lines 625-1644, from the `bal_serializer_addr_matches` section header through `balSerializerVerifyFunction`, about 1020 lines. Moving it to a sibling `BalSerializer.lean` and importing it back leaves `BlockAccessListBuilder.lean` near 1000 lines, comfortably under the cap. `blockAccessListBuilderFunctions` (line 1645) and the `#guard` blocks below it keep working through the import.

That relocation should be **byte-neutral**: emission order comes from the concatenation expressions, not from file layout. So it is checkable — after the split, the regenerated `symbol-addresses.tsv` and `RegionMap.textSizeBytes` should be identical to before it. If they are not, the move changed something and should be rejected.

Nothing else is blocked behind this: everything else in the queue has drained (22 PRs this session, most recently #10760).

pirapira:
Holding the merge label briefly, on my own request rather than anything found. Both this PR and #10761 carry the `BlockAccessListBuilder.lean` -> `BalSerializer.lean` split, and the one check that can see a routine falling out of the emitted concatenation during a move — the emitted concatenation being **byte-identical before and after the split** — has not been reported yet.

Marker-presence checks were run and passed, but those are positive existence checks: they confirm the things someone thought of survived. They cannot see a routine that silently left the emitted string, which is the specific failure a file split causes and the one that passes every guard.

It is one command per branch. Label goes straight back on when the comparison is reported.

pirapira:
Hold cleared — byte-identity reported and clean on both branches. Pre-split `1a6c53b8e` vs post-split, 80979 lines each, sha256 `18dd112a93115bdb…`, identical; #10761 likewise at 81009 lines.

Label restored. Total cost of the hold: two minutes.

Worth recording why the check was worth insisting on, since it generalises past this PR. The guards deliberately kept with the builder assert properties of the concatenated text **against expectations**. They do not assert it against **itself before the move** — only a diff does that. And #10761's case was the one that needed it: not a mechanical cut but a merge resolution where one side was hand-picked, two blocks ported across, and the split re-run — three separate opportunities to drop a routine from the concatenation, covered all at once by one comparison.

Same shape as the negative-vs-positive guard lesson from the `andi 255` fix, one level up: **presence of what you expect cannot substitute for absence of change.**

Issue #10753, opened 2026-07-27
Parameterise programs on a GuestLayout type: import the layout's type, not the generated instance


pirapira:
> Maintainer proposal: *"parametrize programs with guest layout (without importing the concrete guest layout, just the type of guest layout)."*

## The proposal

Give programs the layout as a **parameter**, and let modules import only its **type**:

```lean
-- hand-written, stable, never regenerated
structure GuestLayout where
  zkvm_keccak256         : Nat
  block_access_list_hash : Nat
  …

-- a program takes the layout instead of importing its values
def fooBody (L : GuestLayout) : Program :=
  … jalOff L.zkvm_keccak256 (L.foo + 8) … laHi L.bnf_le_a …
```

A `Programs` module then depends on a **stable, hand-written type**. The **generated instance** — today's `GuestAddrs.lean` — is imported only by the top-level assembly layer.

## Why it matters: the current cost

`GuestAddrs.lean` is 1149 lines, **regenerated wholesale on every layout change**, and imported by **141 modules**. Changing it invalidates all 141 and everything downstream, which is the dominant cost of a regen cycle — up to four full Lean builds per cycle (#10746).

Most of those modules do not depend on the *values* at all. The repo says so in prose (`docs/regenerating-generated-files.md`): a size change *"only requires regenerating the TSV + `GuestAddrs.lean`, never the hundreds of `_prog`s."* **Regeneration, no; recompilation, yes** — purely because they import the file that churns.

And it is machine-checked for a large subset: **53 files carry guards of the form `body.flatten 0 = body.flatten 0x80000000`**, a proof that the encoding is identical at any base. We rebuild those on every layout change knowing the output cannot differ.

## Why parameterisation is strictly better than a difference-vs-absolute split

An earlier framing (#10751) proposed decoupling only **difference-based** bodies — `jalOff A B` depends on `A - B`, so a uniform shift leaves the emitted byte unchanged — and accepted that **absolute** references (`laHi`/`laLo` of a data symbol) stay layout-dependent. That capped the win at the difference-based fraction.

**Parameterisation removes the cap.** An absolute address taken *from a parameter* is as layout-independent at compile time as a relative one: the value arrives at application, not at import. The decoupling becomes **total** rather than partial, and no classification of reference forms is needed.

## Correctness benefit, independent of build time

[...]

PR #10756, opened 2026-07-27
perf(build): prototype GuestLayout param on BloomAddValue

pirapira:
## Summary

Prototype for GH #10753 / bead `evm-asm-8fz1p`: parameterise one mixed-reference program module on a hand-written `GuestLayout` type so it stops importing generated `GuestAddrs`.

Full measurement write-up: `docs/agents/guest-layout-param-measurement.md`.

### Results (against origin/main aff464653)

| Metric | Before | After |
|---|---|---|
| BloomAddValue force-rebuild elab | 260–265 ms | 255–260 ms |
| Modules rebuilt after GuestAddrs.bav_hash flip | **3** (GuestAddrs, HashBridge, BloomAddValue) | **0** |
| BloomAddValue.olean after flip | changed | **unchanged** |

Population re-check on origin/main: jalOff GuestAddrs 161, laHi 141, BOTH 118 — most importers are mixed; reference-form split would not decouple.

### Granularity (microbench, not in lake lib)

- **Flat structure ~1125 fields: NEGATIVE** (heartbeat / isDefEq timeout; also IR constructor-field limits earlier).
- **Nested structure-of-structures: POSITIVE** (~1.0–1.1 s at 1125 fields, ~2.8–3.2 s at 3089).

Recommendation: keep parameterisation; expand via nested groups (RegionMap-like) or Symbol→Nat inductive. Do not ship a mega-flat GuestLayout.

### Code changes

- `GuestLayout.lean` / `GuestLayoutInstance.lean` (3-field prototype)
- `BloomAddValue.lean` takes `(L : GuestLayout)`; probe split so HashBridge/GuestAddrs leave the leaf closure
- `guest_image_coverage.py` applies `guestLayout` for layout-parameterised progs
- Guard cost: apply `.zero` (emission/relocs stay symbolic) or `guestLayout` at image row

### Not measured

- Full 141-module conversion wall time
- End-to-end regen-cycle savings
- Tasks 2 (never-emitted) and 3 (diff-driven remap)

Opened for coord review.

## Test plan

- [x] `lake build EvmAsm.Codegen.Programs.BloomAddValue`
- [x] `lake build EvmAsm.Codegen.Programs.BloomAddValueProbe`
- [x] `lake build EvmAsm.Codegen.Proofs.GuestImageEntries`
- [x] Synthetic GuestAddrs perturbation leaves BloomAddValue.olean untouched
- [ ] CI green


pirapira:
Recording what this prototype established, so the finding survives independently of the branch. The branch has not moved since 2026-07-27T23:50Z and its author has not reported since; leaving the PR open for the maintainer's scope call on #10753 rather than closing someone else's work.

## What the prototype measured, and it is a real result

The build failure here **is the measurement**, not a mistake. `check-asm-to-program` renders a generated Lean harness calling `emitProgram bloomAddValue_prog`, and after parameterisation that symbol has type `GuestLayout → Program` where a bare `Program` is expected (application type mismatch).

That matters for #10753 because it **widens the known cost surface**. The original brief anticipated the *guard* cost — every `_prog` becomes a function, so a `#guard` pinning one against a literal must apply a concrete layout. This shows the parameterisation also reaches a **generated consumer**: the asm-to-program checker generates code over the program symbols. That is exactly the kind of fact that decides a 141-module rollout, and it cannot be found by arguing — one prototype module surfaced it.

## What remains unmeasured

Stated so nobody reads the above as a complete answer:

- **the consumer enumeration** — how many *kinds* of `_prog` consumer break (guards, the asm-to-program renderer, emit paths, region-map paths), and for each whether it takes a mechanical adaptation, a concrete layout applied at the consumer, or a design change. One example is a surprise; a count is a decision input.
- **elaboration time before and after** for the parameterised module.
- **whether a synthetic layout perturbation now leaves the module untouched by the build** — the property the whole proposal exists to buy.
- **the `rfl`-pinned-proof question**, which is the one that could block the rollout outright rather than merely price it. `Bn254FieldMulModSAsm.lean:91` does `rw` with `show GuestAddrs.bnf_le_a = 2734690016 from rfl`. `BloomAddValue` carries no such proof, so nothing measured here speaks to it. Note that a proof closing only with a raised `maxHeartbeats`, or with `native_decide`/`bv_decide`, is a **failure result** under this project's standing rules — not a pass with a knob turned.

## Status

`BEHIND` main with a failing build. The failure is the informative one described above and should be read before it is fixed.

Not blocking anything: no other lane depends on this prototype, and #10753 is a build-time optimisation proposal rather than a correctness gate.

pirapira:
This is going to be a big change. parking for now.

PR #10757, opened 2026-07-27
fix: record EIP-7702 authority account writes

pirapira:
## EIP-7702 authority account-write producer

Records accepted EIP-7702 authority nonce and delegation-code mutations in the
existing transaction account-write map, using block-lifetime 24-byte code slots.
The record is taken after accepted auth preparation and is restored to the
auth-preparation undo mark on preparation failure; it survives a later body
revert.  The transaction-map upsert keeps the final nonce and final delegation
target for repeated authorizations of one authority.

This makes the previously unreachable authority half of the account-builder
producer live: before this change `bal_builder_append_nonce` and
`bal_builder_append_code` could receive balance and sender-nonce effects, but
never an authorization authority's nonce/code mutation.

### Direct row evidence (diagnostic-only probes; not shipped)

Same probe and same `00635` input:

| artifact | builder rows |
| --- | --- |
| main (`b147466e…`) | balance=1, nonce=1, code=0; compact marker present |
| candidate (`c970c7ff…`) | balance=1, nonce=2, code=1; authority `c0f6dc9e5836f54caadbf59cc69346c508e1992b`, BAI=1, nonzero pointer, len=23, `ef0100c4028cd457e62aaa998b95f7edf9655797c54309` |

`24627` (same signer, increasing nonce) has one authority row with final
nonce 10 and final target `04a2f29573a84b913fcc243f18b627c2bc745f27`:
the upsert preserved paired final nonce/code values.

Rollback controls:

- `00625` auth-preparation failure: authority absent from nonce and code rows.
  The sole remaining `f6c3…b3ff` nonce row was directly exported as the
  fixture sender, so it is sender inclusion rather than leaked auth state.
- `24954` `pointer_reverts` body-revert: authority code row remains with
  BAI=1 and a 23-byte `ef0100 || target` designator.

### Shipping artifact and gate

Clean shipping artifact after removing every diagnostic export:
`e60dfc16682bcabd39537f2b0d0510fe02098d973ed13f670c474e778b7b425b`.
Sections: `.text=0x64f24`, `.data=0x5370`, `.bss=0x1b811440`.

After opening, the branch integrated current main (`e364c505c`) and regenerated
the conflicted generated artifacts rather than hand-merging them.  The merged
artifact is `35509fb969a4005b394bb0fb3f1dae7d130ed7ebec2fbde07c0546e1bf77438c`,
with `.text=0x65300`, `.data=0x5370`, `.bss=0x1b811480`; two relinks had
identical `.text` and `.data` hashes (the ELF container difference is only its
string table).  The original direct row probes attribute the authority-producer
change; the post-merge gate measures the artifact under review.

Ran:

- `LAKE_ARTIFACT_CACHE=false bash scripts/check-region-map.sh`

The gate passed and confirmed linked ELF/symbol TSV/RegionMap agreement.


pirapira:
## Review — approving. Two questions, neither blocking.

38 lines of hand-written change under 3000 lines of generated churn, and the reasoning is **at the line** rather than only in the PR body. Reviewed against the design pinned during this work.

### What is right

**The arena's lifetime justification is in the code**, where the next reader meets it: *"account-write rows retain a pointer to those bytes until the later BAL builder pass, so this must be a block-lifetime NOBITS arena, not a reusable per-auth scratch."* That is the fact that ruled out tx-local scratch, recorded where a future optimiser would otherwise reintroduce it.

**The two rollback arms are distinguished in a comment at the mark**: a preparation `ExceptionalHalt` drops accepted auth mutations while **retaining sender inclusion and the staged debit**; a body revert uses the *later* body mark and keeps the authorization phase. That is the spec semantics (`eoa_delegation.py:283-284` applies before execution, so a reverting body does not undo it), written next to the mechanism that implements it.

**`account_writes_auth_prepare_mark` is its own cursor** rather than a reuse of the dead frame checkpoint — and the comment says so. `account_writes_undo_checkpoint` has one reference and no users, so it would have looked like the intended API.

**The capacity guard is fail-closed**: `bgeu t1, t2, .L77prep_bad_record` sends overflow to a reject path, not past the record.

**The designator is built correctly and the null case is handled**: `ef 01 00` then 20 bytes copied, `a4 = 23`; a null delegation takes `a4 = 0` rather than recording a 23-byte designator for a revocation.

**Mask is `NONCE | CODE`**, which is the corrected scope — the authority was missing *both* components, not code alone.

### Question 1 — `accountWriteHasNonce + accountWriteHasCode`

Addition where a bitwise OR is meant. It is correct today because the flags are distinct powers of two, so `+` and `|` coincide. It stops being correct silently if a flag is ever duplicated in such an expression, or if the flag set gains a non-disjoint member. A one-token change to `|||` (or whatever the Lean-side spelling is) removes the class. Not blocking — just cheaper to fix now than to notice later.

### Question 2 — the `code_state_mtx_active` gate

The record is gated on the MTx lane (`beqz t0, .L77prep_next`), with the comment noting the direct single-tx lane has no account-write builder pass. Since the `count == 1` selector flip (#10591) made MTx universal, that gate should always be taken.

**Not blocking, and the reason is structural:** the gate can only fail to *add* a record, so it cannot widen acceptance — the worst case is the pre-existing behaviour, which is what this PR fixes. But worth confirming whether the direct lane is reachable at all, because if it is, authorities go unrecorded there and this fix is lane-conditional. That question belongs with #10685 (the single-tx/multi-tx dead-twin inventory) rather than here.

### Evidence

The strongest package landed in this workstream, and I verified its key claims independently rather than reading the body:

- paired main/candidate on one fixture with **identical instrumentation and the marker present on both legs**, so main's `code=0` is an *observed* zero rather than an unreadable slot;
- `nonce` 1 → 2 across the pair, confirming the authority's nonce was missing too;
- the designator checked **byte-for-byte including its length** against the fixture's declared `codeChanges`, not merely counted;
- **both rollback arms**, with the sender attributed by *exporting `bv_mtx_sender_addr`* rather than pattern-matched across fixtures — a row's presence is data, the producer it is attributed to is an inference, and that substitution has produced several wrong readings in this work;
- the duplicate-authority upsert advancing **nonce and target together**, which is the specific way a per-address upsert produces an internally inconsistent row that hashes wrong and looks fine in a count;
- diagnostic residue checked in the **emitted asm** (`Lsg_abm`, `c0dec0de`) rather than in the source.

Joint `.bss` budget with the arena in: end `0xbf811440` against `.sszscratch` at `0xbf980000` — **1,502,144 bytes free**. Fits with room, alongside #10752's +256.

`DIRTY` against main, so it needs the usual generated-file resolution: bootstrap the four from main and regenerate by hand, not through `regen-cycle.sh` (#10746, removal in #10750).

Issue #10759, opened 2026-07-28
Guest load-access fault in bv_mtx_committed_chunked_snapshot_upsert on EIP-7702 fixtures: mtval=0xff suggests a byte sentinel dereferenced as a pointer


pirapira:
## Summary

A reproducible **guest load-access fault** in `bv_mtx_committed_chunked_snapshot_upsert`, on EIP-7702 fixtures, at two nearby offsets in the same routine. This is an **ERROR** — abnormal termination — which is a separate category from a false accept or a false reject, and it is **pre-existing on `main`** (both instances observed on base legs).

| fixture | `mepc` | offset into routine | `mcause` | `mtval` |
|---|---|---|---|---|
| `00636` (7702 authority family) | `0x80029dbc` | **+0xf4** | 5 (load access fault) | `0xff` |
| `00349_test_bal_7702_top_frame_delegation_oog…oog_at_delegation_charge` | `0x80029d94` | **+0xcc** | 5 | `0xff` |

Both resolve into `bv_mtx_committed_chunked_snapshot_upsert`, which starts at `0x80029cc8` per `scripts/asm-fixtures/symbol-addresses.tsv`.

## The signature is specific: a byte value used as a pointer

For a load access fault, `mtval` holds the faulting address. **`mtval = 0xff` means the guest attempted a load from address 255.** That is not a corrupted pointer in the usual sense — a wild pointer lands somewhere large and arbitrary. **255 is what you get when a single byte is used as an address.**

In a *chunked* structure, the natural source is a chunk index or chunk pointer held in a byte-wide field, where **`0xff` is the conventional "absent" or "no chunk" sentinel**. Dereferencing the sentinel instead of testing it produces exactly this fault, at exactly this address, every time the absent case is reached.

That is a hypothesis from the address rather than a reading of the routine, but it is a narrow one and it predicts where to look: the two offsets (`+0xcc`, `+0xf4`) are the two dereference sites, and the missing guard is the sentinel test before them.

## Why both instances are 7702 fixtures

Both observed cases are EIP-7702 — an authority-family fixture and a delegation-OOG fixture. The second is specifically `oog_at_delegation_charge`, i.e. the authorisation phase runs out of gas partway. That is the shape most likely to leave a chunked snapshot **partially populated**, which is consistent with reaching an absent-chunk case that fully-populated paths never hit.

[...]

PR #10760, opened 2026-07-28
fix: route cross-transaction created recipients through current state

pirapira:
Root cause: dispatch_tx_runtime_code already resolves recipient code in the correct order, AccountState first and the immutable header only on an overlay miss. It was unreachable for a cross-transaction-created recipient because the earlier MTx classifier unconditionally called code_hash_at_header_state_root, saw the target empty in the pre-block state, and routed to is_eoa. The fix makes that classifier consult AccountState first, then use the header only on a genuine current-state miss.

This restores both state-gas execution and BAL completeness: the two CREATE/CREATE2 controls change storage capture count from one to two, so the tx2 storage row is observed and can participate in declared-BAL comparison.

Full targeted paired gate, measured before the generated-only merge resolution:
- selector: --all --filter block_access_list; 1,022 result files, no limit
- legs: base da5c173caa7337873b69c2f150eade4fc5768c0da8ba4bd806aa5d99cdc40bd4 versus converged candidate 357cf0ed6cf13bace589205778b1a1505a451474be7007029c9324deb340f6ca; same Spike runner bae4826988540415b9bf862248caa7499b2ffbb9eec5cfab61cce00e089cf2c3
- scored: 1,021 scored verdicts plus one identical pre-existing ERROR on each leg
- comparison: mechanical per-file comparison of all 1,022 result TSVs with identical sorted filenames; exactly two differ
- transitions: 00041 cross_tx_deploy_then_call CREATE and 00042 CREATE2 are false-reject to accepted/full match. Nothing else moves; zero candidate-pass where base failed.
- shared ERROR: 00349 7702 top-frame delegation OOG, #10759, in bv_mtx_committed_chunked_snapshot_upsert; identical on both legs and not attributable to this change.

The merged artifact is a different generated-layout image: sha256 e6a7b232aca641577c37a6e19323901aa1c202db89c772d59dccfcdc23ac362e, text 0x65b74, data 0x5370, bss 0x1b811580. The semantic source diff is unchanged. Generated artifacts were bootstrapped from current main and regenerated serially; a third pass left GuestAddrs, GuestImageEntries, symbol TSV, and RegionMap hashes unchanged. check-region-map and check-asm-to-program passed, and the two focused CREATE/CREATE2 controls remain accepted on the merged ELF.

Additional control: random 200 paired seed 1706065709 has identical per-file outcomes: 194 full matches and the same six false rejects per leg.


pirapira:
## Review — approving. Eight hand-written lines, gated at per-file granularity across the full targeted population.

The whole change is a comment and two instructions; the rest is generated churn (`GuestAddrs` ±248, TSV ±709, `RegionMap` ±1).

```
la a0, bv_mtx_ctx; addi a0, a0, 72; jal ra, account_state_lookup_current
li t1, 1; beq a0, t1, .Lbv_mtx_is_contract; bnez a0, .Lbv_mtx_is_eoa
```

Overlay first, header only on a genuine miss (`a0 == 0` falls through). Correct shape.

### What is right

**The root cause is in the body rather than only the fix**, and it is the transferable part: `dispatch_tx_runtime_code` *already* consulted current state first and fell back to the header only on status 0. It was simply **not reached**, because this classifier called `code_hash_at_header_state_root` unconditionally, saw the target absent from pre-block state, and routed to the EOA lane. **A correct mechanism made unreachable by an upstream decision** — anyone auditing the resolver would have found it correct and moved on. That is the kind of defect worth naming as a class.

**The comment cites its precedent** — the layered resolver already used by the child CALL-family handlers and by the dispatcher — so this is consistency with an existing pattern rather than a novel mechanism, and a future reader can check it against a sibling.

**The evidence is the strongest in this workstream and I asked for it in this exact form:**

- `--all --filter block_access_list`, **1,022 cases, no `--limit`** — the full targeted population, not a prefix;
- base `da5c173c` vs converged candidate `357cf0ed`, **same runner sha `bae48269` on both legs**;
- **mechanical per-file comparison of all 1,022 result TSVs, exactly two differing** — `cross_tx_deploy_then_call` CREATE and CREATE2, `succ=00` → `succ=01`, full match. Nothing else moved in either direction;
- `1,021` **scored** plus one identical pre-existing ERROR per leg, stated as such rather than as 1,022 — the file count and the scored count differ, and that ERROR is now #10759;
- random-200 at a captured seed with identical per-file outcomes and identical FR label sets.

Per-file comparison is what makes this conclusive rather than suggestive: a classifier that can move verdicts both ways produces *identical totals with different members* as its signature, and enumerating the two differing files closes that.

**And the capture count going 1 → 2 means this was a BAL completeness gap, not only a state-gas one** — a storage row that was never observed now exists. A rebuilt BAL could not have matched a declared BAL while that row was missing, which is why this had to precede wiring the serializer's comparator.

### One non-blocking note

`bnez a0, .Lbv_mtx_is_eoa` maps **every** non-zero non-one status onto the EOA lane. That is safe today and safe by default — an unknown status behaves as the pre-existing code did, since the old path routed to EOA whenever the header saw an empty account. But it is the *N-statuses-onto-two-outcomes* shape, and `account_state_lookup_current`'s status contract is not documented at this call site. A one-line comment naming what 1, 2 and 3 mean would stop a future fourth status being silently absorbed as an EOA — which would reproduce exactly the under-execution this PR fixes.

### Not blocking

`DIRTY` against main, so it needs the generated-file resolution: bootstrap the four from main and regenerate to convergence, as you did for the previous one — third pass leaving all four hashes unchanged is the fixpoint demonstration to repeat.

PR #10761, opened 2026-07-28
test(bal): the producer-path case, checked against a digest frozen before it existed (#10680)

pirapira:
Closes the gap that let two byte-order defects survive five green digests: **every other case seeds the builder row itself**, in the reader's convention, and is therefore blind to a producer/consumer mismatch. Case 11 drives `bal_emit_storage_changes`, so the producer's LE→BE reversal and the serializer's read meet on one path.

## The expectation was frozen before the case was written

`EXPECTED_E2E_DIGEST = 24f0ad8b…` landed in the *previous commit*, derived from the field table and the yellow paper, with the derivation shown. Then the case was built and matched it.

That ordering is the substance, not ceremony. With the producer in the loop a captured digest certifies the producer's behaviour **by definition** — it cannot fail — so it would have enshrined a reversal bug as correct. Same trap a captured baseline would have set with the widener's 33-byte scalar, one level further out. The runner comment says a disagreement is information about the producer and not a reason to update the constant.

## Construction, verified against emitted code

- tx row at `0xa21a0000`: address and slot as **LE stack words** — the form the producer reverses
- container entry at `0xa1fa0000` sharing the first 64 bytes, so the scan **hits** and jumps to `.Lbesc_have`, skipping `slot_at_header_state_root` and its witness globals entirely
- baseline 9 against a written 5, because equal values trip the net-zero check and emit nothing
- **slot 7, not 0** — a BE-vs-LE dword compare matches only byte-symmetric values, so slot 0 passes under the exact defect this case exists to catch

`slot_at_header_state_root` is stubbed as `unimp` — it **traps**. A stub returning success would let the case pass via the miss path and silently stop exercising the container hit. Unreachable by design, loud if that changes.

## Runner

```
  ok    case 1  one change           = 6
  ok    case 2  same slot twice      = 9
  ok    case 3  two distinct slots   = 12
  ok    case 4  other address present = 6
  ok    case 5  two-byte value       = 8
  ok    measure_slot payload         = 5
  ok    measure_slot inner payload   = 3
  ok    case 9  cross-tx read exclusion = 1
  ok    case 6  emitted digest              = 31597f4956266576...
  ok    case 7  whole-account digest       = b790ac5342ebff17...
  ok    case 8  outer-list digest (2 accts) = e74830bb60be86e7...
  ok    case 9  emitted reads digest        = 60811857dd566889...
  ok    case 10 sort-then-rebuild digest    = 0xe786be60bb3048e7
  ok    case 11 producer-path digest        = 0xa8e247c48badf024
==> PASS: 8/8 lengths and the emitted digest match the RLP derivation
  ok    bal_rlp_encode_selftest      PASS (15 cases)
  ok    bal_canonical_sort_selftest  PASS (3 row sets)
==> PASS: 2/2 self-tests pass
```

Stacked on #10755, which carries the emit pass, both byte-order fixes and the builder row field table.

Still open on the serializer: six of seven orderings (`block_access_lists.py:539-579`), which this construction can now validate.


pirapira:
## The expectation was edited once. Here is why that was not a capture.

The bare sequence — *froze an expectation, case failed, edited the expectation, case passed* — is the shape of quietly fitting a constant to a run. It isn't what happened, and the record distinguishes them.

**The audit trail, in commit order:**

| # | commit | expectation | case 12 | why |
|---|---|---|---|---|
| 1 | frozen | `eb1d0c35…` (account-level) | RED | derived from the spec **before the case existed** |
| 2 | narrowing ×3 | unchanged | RED | lengths matched; four permutations enumerated |
| 3 | byte dump | unchanged | RED | showed leading `ea` — my derivation omitted the **outer list header** |
| 4 | derivation fix | `a9da0dbc…` (outer-level) | **still RED** | `0xea = 0xc0+42`, computed from the spec, not from the run |
| 5 | code fix (`:153` mask) | **unchanged** | GREEN | slot 3 now precedes slot 7 |

**The load-bearing row is 4.** After correcting the expectation the case *still failed*. Had I been fitting the constant to the measurement, the edit would have turned it green immediately — that is what a capture looks like. Instead it stayed red for a second, independent reason, and only a change to the **code** closed it. The measured value at step 4 was `0xaa1981e1…`; the corrected expectation was `0xa9da0dbc…`; they are different numbers.

**And the final pass is byte-level, not digest-level:**

```
before  ea e994aa…cf c807c6c20105c20206 c503c3c20107 c0c0c0c0
after   ea e994aa…cf c503c3c20107 c807c6c20105c20206 c0c0c0c0
```

Slot 3 now precedes slot 7, the index level is unchanged and still ascending. The fix is demonstrated at the level the defect lived, against bytes derived by hand from `block_access_lists.py` and the yellow paper.

**What the edit actually corrected** was a real error in *my* derivation: case 12 runs through `rebuild_hash`, which emits the outer list, and I had derived a bare account. That error also invalidated my earlier four-permutation elimination — I was matching account-shaped candidates against an outer-shaped emission, which is why nothing matched and why I wrongly concluded ordering was not the cause.

The general point: freezing an expectation protects against capture only if you also refuse to let a *failing* case be resolved by editing it. The test here is whether the edit made it pass. It didn't.

pirapira:
Reviewed. One finding, which I do **not** think should block the merge — recording it because the design intent here is unambiguous and the code does not match it.

## `bal_serializer_rebuild_hash` propagates the raw sort code instead of `2`

The documented contract (`BalSerializer.lean:1044-1046`):

> `a0` (out) = 0 the rebuilt BAL hashes to the supplied BAL's hash, 1 it does not, **2 the canonical sort failed** (its code is in `bal_serializer_sort_status`).

That is coherent and the `bal_serializer_sort_status` cell exists precisely to carry the finer code, so the intent is clear: return `2`, park the specific sort code in the cell.

The emitted sequence does not do that. Each of the six sorts is followed by

```
la t0, bal_serializer_sort_status; sd a0, 0(t0)
bnez a0, .Lbsrh_ret
```

and `.Lbsrh_ret` returns with `a0` **still holding the sort's own code**. `bal_canonical_sort` returns 1, 2 or 3 on failure (`BalCanonicalSort.lean:216`, `:277`, `:279`). So:

- a sort failing with **1** returns 1 — which this routine's own contract defines as *"the rebuilt BAL hash does not match"*. A sort failure is reported as a hash mismatch.
- a sort failing with **3** returns 3, which the contract does not define at all.
- only a sort failing with 2 returns the documented value, by coincidence.

Fix is one instruction on the bail path: `li a0, 2` before `j .Lbsrh_ret`, leaving the specific code in the status cell where the docstring already says it lives.

## Why it is not blocking

The comparator is deliberately inert on this branch — nothing consumes this return value yet — so there is no live wrong behaviour, and the direction is a false *reject* (a sort failure reported as a mismatch rejects a block that might be valid), never a false accept. It should be fixed before the comparator is wired, not before this lands.

## Two notes for the guard set

The store-then-`bnez` pattern is correct and I withdraw a concern I formed before reading it: I expected six stores to a single cell to lose all but the last, but control exits on the first nonzero, so the cell holds the failing sort's code at the point of bail.

`bal_serializer_sort_status` has **no reader anywhere on the branch** — five write sites, zero loads. That is legitimate for a `.bss` diagnostic channel read externally, and it is documented as one. Flagging it only so it is a known choice rather than an oversight, since an unread cell is the shape that has bitten this stack repeatedly.

And the guards here pin emitted *text* and *field selection*; none of them pins a **return code against its documented contract**. That is the gap this finding sits in.

## The rest reviewed clean

- The `andi 0x7f` fix and its re-read of the endianness flag from the descriptor (`slli a7, a6, 4`) are right, and the guard is correctly **inverted** — `splitOn "andi t3, t3, 255" == 1`, asserting the wrong form is absent. That is the form that survives its own repair, which the previous positive guard did not.
- Six `bal_canonical_sort` calls carrying seven rules (the storage sort carries two in one pass), each cited to `block_access_lists.py:539-579`.
- Field order cited to the `AccountChanges` class rather than to a docstring.
- Byte-identity of the emitted concatenation verified across both the split and the sync.

pirapira:
## ⛔ Retracting the finding above — it is wrong. There is no defect here.

I attributed one routine's contract to a different routine. There are two, and they compose correctly:

- **`bal_serializer_rebuild_hash`** (`:957`): *"`a0` (out) = 0, or the canonical sort's own nonzero status."* Propagating the raw sort code is exactly what this routine documents. Not a violation of its contract — it **is** its contract.
- **`bal_serializer_verify`** (`:1044-1046`): 0 match / 1 no match / 2 sort failed. Its body does the conversion I claimed was missing:
  ```
  jal ra, bal_serializer_rebuild_hash
  beqz a0, .Lbsv_rebuilt
  li a0, 2; j .Lbsv_ret
  ```
  Any nonzero from the rebuild collapses to `2`, and the finer code stays in `bal_serializer_sort_status` where the docstring says it lives.

So the `li a0, 2` I asked for **already exists**, one level up, at the boundary where the contract that mentions `2` actually applies. Nothing to fix, and the split of the routine is what made this legible in the first place.

My error is the one I flag in others: **right text, wrong population.** I quoted a docstring accurately and bound it to the wrong routine, because I read the emitted sequence and the nearest contract without checking which definition the sequence belonged to. A `ret` in between is the boundary I stepped over.

Also worth saying: in the same review I withdrew a concern *by reading* (the five-stores-to-one-cell worry, answered by the `bnez`) and then shipped this one *without* reading far enough. Reading first is not a habit you get credit for once per review.

The clean-review notes in the previous comment stand, and this PR has no open finding from me.

pirapira:
Verified CLEAN independently — all checks pass, build SUCCESS, synced to `9e1fd082f` with byte-identity across the sync, layout at fixpoint, region map matching. Labelled.

Reviewed: no open findings. My one finding was retracted above — it attributed `bal_serializer_verify`'s contract to `bal_serializer_rebuild_hash`, and the two compose correctly.

What lands here:

- the seven ordering rules of `block_access_lists.py:539-579` implemented **and demonstrated**, via a producer-path case checked against a digest frozen before it existed
- the `andi 255` → `andi 0x7f` sort fix, with the endianness flag re-read from the descriptor, and its guard **inverted** to assert the wrong form is absent — the form that survives its own repair, which the prior positive guard did not (it would have blocked the fix it existed to catch)
- the `BlockAccessListBuilder.lean` → `BalSerializer.lean` split, byte-identical emitted concatenation

Two prerequisites for #10680 were filed out of this work and are **not** addressed here, deliberately: #10764 (BAI 0 and N+1 have no producer, so a rebuilt BAL cannot express withdrawal credits or per-block system writes) and #10765 (nine of eleven granular reject labels unclassified against digest-representability). The comparator wiring stays unwired until those are settled — a digest comparison must be able to express what the checks it replaces can express, before those checks are removed.

PR #10762, opened 2026-07-28
fix: restore MTx prep-halt storage snapshot arguments

pirapira:
## Summary

Restore the current transaction's empty storage slice after the preparation-halt
rollback path clobbers caller-saved `a1`/`a2`.

`bv_mtx_committed_chunked_snapshot_upsert` takes a *per-transaction* slice:
successful paths pass `a1 = bv_user_storage_log + old_count * 128` and
`a2 = new_count - old_count`.  Its existing `snapshot_empty` arm already passes
the same base with `a2 = 0`.  The preparation-halt arm must do so again after
`blockVerdictMtxRecordSenderRefund` and `account_writes_restore_frame`; otherwise
the snapshot helper dereferences their clobbered argument value (`0xff`).

The backing user-storage arena remains cumulative for later BAL comparison;
this change passes an empty delta for only the halted transaction and therefore
does not discard earlier rows.

## Evidence

- Base ELF `da5c173caa7337873b69c2f150eade4fc5768c0da8ba4bd806aa5d99cdc40bd4`;
  candidate ELF `48b014c01db8fad8f769d0a54d20447e8dd0d5993ed8aaec026078f71bbaa42a`.
- Full `--filter 7702` A/B: 1,162/1,162 rows per leg. Base: FA=0, FR=224,
  ERROR=2, full=936. Candidate: FA=0, FR=225, ERROR=0, full=937. The ERROR
  accounting closes exactly: one ERROR became a full pass
  (`bal_7702_top_frame_delegation_oog`), and one became an FR
  (`valid_tx_invalid_chain_id/auth_chain_id_2`). Thus FR 224→225 is a formerly
  masked verdict becoming visible, not a candidate-introduced reject. The latter
  has a valid transaction with a wrong-chain-id authorization, which EIP-7702
  skips; its reject is a newly exposed completeness gap. Both base logs fault at
  the same load (`mcause=0x5`, `mtval=0xff`, `mepc=0x80029d94`), despite reaching
  the incomplete-authorization route independently. No false accepts occurred.
- Paired random-200, captured seed `1601526629`: both legs score 200/200,
  FA=0, FR=6, ERROR=0, with byte-identical results on all scored cases.
- Focused top-frame-delegation-OOG gate: 2/2 full matches on the candidate.
- `lake build`, `scripts/check-region-map.sh`, and
  `scripts/check-asm-to-program.sh` pass. Candidate linked layout:
  `.text=0x65b80`, `.data=0x5370`, `.bss=0x1b811580`.

The first focused run is intentionally discarded: its relative guest-ELF path
resolved to main rather than this candidate. The rerun used the absolute
candidate path and recorded its SHA above.

## Residual test coverage

The restore is justified by the snapshot ABI and the routine's own
`snapshot_empty` precedent. The corpus has no fixture combining a prior storage
write with a later authorization-preparation OOG, so no current sweep can
empirically discriminate preservation of earlier cumulative storage rows through
this restore. A future change here needs a two-transaction generated control:
prior storage write followed by a preparation OOG.


pirapira:
## Review — approving. Seven hand-written lines, and the derivation is *at the line* rather than only in the body.

```
-- caller-saved a1/a2.  The snapshot ABI takes this transaction's captured
-- slice, not the cumulative `bv_user_storage_log`: a preparation halt has
-- no such rows, so restore base + length zero explicitly.  Do not reconstruct …
la a1, bv_user_storage_log; li a2, 0
```

Everything else is generated churn (`GuestAddrs` ±248, TSV ±720, `RegionMap` ±1).

### What is right

**The comment answers the question a reviewer would ask, where they will ask it.** I challenged `a2 = 0` on the grounds that `AccountTupleSequencesConsistent` documents that arena as *the complete user-write source* — cumulative across the block. The answer is that `a2` is a **per-transaction delta** (`capture_new_count - capture_old_count`, with `a1` offset by `old_count * 128`), so `0` means *this transaction contributed no rows*, which is exactly true after a rolled-back preparation. Both statements hold: the arena is cumulative, `a2` indexes a slice of it. That distinction is now in the source.

**And `a2 = 0` matches the routine's own `snapshot_empty` arm**, which already passes base plus zero on the analogous path. That is the strongest justification available for a literal — not that it looks right, but that the identical situation is already handled that way two arms over.

**`Do not reconstruct` pre-empts the wrong fix.** An unconditional reconstruction after `code_commit_done` would be wrong on the empty path precisely because capture counts can be **stale from a previous successful transaction** — the same staleness that made the literal worth questioning. Recording the rejected alternative is worth more than recording the chosen one.

### The fix is class-level, not per-fixture

Both faulting fixtures report an identical `mcause=0x5 mtval=0xff mepc=0x80029d94`, reached by **unrelated causes**: an authorisation phase running out of gas (`bal_7702_top_frame_delegation_oog`), and an **invalid authorisation chain-id** (`valid_tx_invalid_chain_id / auth_chain_id_2`). Two routes to one site means the repair addresses *an authorisation phase that does not complete*, rather than one path.

### Evidence, and one number that needs its accounting read alongside it

Full 7702 population, **1162 cases**, paired, both legs fully scored:

| | full | FR | ERROR | total |
|---|---|---|---|---|
| base | 936 | 224 | 2 | 1162 |
| candidate | 937 | 225 | 0 | 1162 |

`FR 224 → 225` is **not a new false reject** — it is a *fault becoming a verdict*. `ERROR 2 → 0` resolves as one fixture to full and one to FR: `00921` was trapping, so its verdict was never computed, and it now computes one and rejects. **That FR was always there and the fault was masking it.** An ERROR is a stronger mask than an FR, so fixing faults surfaces latent false rejects by construction.

`FA = 0` on both legs and both populations; random-200 at a captured seed **byte-identical** between legs at 200/200 scored.

### Residual, stated rather than implied

**The corpus cannot discriminate this path.** No authorisation-OOG or BAL-7702-OOG fixture has more than one transaction, and no multi-tx 7702 fixture combines a prior storage write with a later preparation OOG. So had `a2 = 0` been wrong, **no fixture would have caught it** — the correctness rests on the ABI derivation and the in-routine precedent, and a future change to this restore has no empirical guard. A two-transaction generator would supply one if that path is touched again.

Also worth its place in the body: the discarded first harness run, where a **relative guest ELF path resolved to `main`** rather than the candidate. That is the stale-artifact trap, and recording it is how the next person learns to pin an absolute sha.

PR #10766, opened 2026-07-28
fix: accept wide canonical EIP-7702 authorization chain IDs

pirapira:
## Summary

Accept canonical EIP-7702 authorization `chain_id` scalars wider than `u64` as
invalid authorizations to skip, rather than rejecting the enclosing transaction.
The outer transaction chain id remains on its existing `u64` path. Both affected
callers are pinned to typed transaction field 9 (authorization list), then tuple
field 0 (authorization chain id).

## Evidence

- Scoped EEST 7702 A/B, parent `48b014c01db8` vs candidate `4083fe6c07ad`:
  1,162/1,162 rows on each side; FA=0 on both; FR 225 -> 224; zero new FR;
  exactly one fix, `00921_test_valid_tx_invalid_chain_id...auth_chain_id_2__2`.
  These attribution legs were measured on the pre-resolution repair tip
  `bb3d9d37e`.
- Random paired gate: discovery seed `1803805343`, reused unchanged for the
  parent. Both legs scored 200/200 with FA=0, FR=5, ERROR=0 and BUDGET=0.
  The comparator joined 198 input groups (two shared-input groups), reported
  zero transitions and byte-identical output for every joined input.
- Narrow matching authorization-chain-id control remains a full match.
- Each selector guard was perturbed separately in its emitter only (`9 -> 8`):
  its owning module failed its `#guard`; restored source rebuilt cleanly.
- The exact attribution build regenerated to `.text=0x065c58`,
  `.data=0x5370`. After the required merge of current main, the generated
  layout was regenerated serially to a fixed point: `.text=0x06644c`,
  `.data=0x5370`, `.bss=0x1b85c9c0`; a second complete pass changed no
  generated file. `check-asm-to-program.sh` and `check-region-map.sh` passed,
  and the unrelated `.data` pin `GuestAddrs.bnf_le_a` remains `0xa3000ee0`.
- Focused final-artifact confirmation on merged tip `85fb30b34`, guest ELF
  `f37951a5cf83`: `00921` accepts (verdict word `01 25`), `00922` remains
  accepted, and the narrow chain-specific control `00540` remains accepted.
  The diagnostic probe rebuilt on the resolved child tip returns status `1`
  for noncanonical `c2 81 00` and status `2` for canonical 33-byte content.

## Reject-side measurement gap

The upstream malformed authorization-chain-id encodings are transaction tests;
there is no stateless blockchain fixture under those test IDs. The current
stateless harness therefore cannot express the reject side of this
acceptance-widening change. The reject side is **unverified by measurement** in
this PR, not verified-and-clean.

More corpus sweeping cannot close this: the valid corpus consists entirely of
fixtures meant to be accepted, so it structurally cannot reach the reject side.
Follow-up #10767 supplies a committed diagnostic probe for noncanonical and
over-32-byte authorization scalars, deliberately kept out of this one-fixture
repair. It runs on the fixed build: both reject sides return nonzero status
codes (noncanonical = 1; canonical 33-byte scalar = 2). A parent probe leg is
deliberately omitted because it would only re-measure the FR direction already
covered by the paired EEST A/B; the FA-adjacent acceptance-widening direction
is a property of the fixed build.


pirapira:
Reviewed. **The semantics are spec-exact** — I checked the one thing that could have made this a false accept rather than a false-reject repair.

`eoa_delegation.py:176` (execution-specs `e5a8caf1b`):

```python
if auth.chain_id not in (message.block_env.chain_id, U256(0)):
    return None
```

and the caller at `:251-254` does `continue` on `None`. So a chain-id mismatch **skips that authorization and leaves the transaction valid** — it is not a transaction-level rejection. Both branches here do exactly that (`.L77prep_next` / `.Lb1an_next`), and the `U256(0)` any-chain case is the `beqz` before the comparison. A canonical scalar wider than u64 cannot equal a U64 block chain id, so skipping it is the spec behaviour rather than a widening of what is accepted at the transaction level.

Also confirmed against the spec, since these are one field index apart in consequence and far apart in the encoding:

- transaction `chain_id` is **U64** — `transactions.py:180, 252, 330, 419`, all four typed classes. Untouched here, correctly.
- `Authorization.chain_id` is **U256** — `fork_types.py:92`. This is the widened read.
- `SetCodeTransaction` field **9** is `authorizations`; the tuple's field **0** is `chain_id`.

The two `#guard`s pin exactly that path (field 9 to `b1an_auth_off`, tuple field 0 to `b1an_target_off`), so an index edit cannot silently widen the outer transaction's U64 chain id. Both were **observed firing** under an emitter-only perturbation (9 to 8, each module separately), then restored to exact text and rebuilt clean — with the guard text diffed to confirm the perturbation did not edit both sides of the check.

`bltu t1, t0` with `t1 = 8` branches on `len > 8`; the wide path decodes with `rlp_content_to_u256_be` and `bnez a0` still fails closed on non-canonical and over-32-byte content. Canonical zero encodes in at most one byte, so no any-chain case can reach the wide path and be skipped.

Evidence set, and I am satisfied with all of it:

- paired A/B, 1162/1162 both sides, FA=0 both, **FR 225 to 224, new=0, fixed=1**, sole diff 00921
- random paired gate at captured seed `1803805343`, both 200/200, FA=0 FR=5, and a self-checking comparator joining 198 cases with **zero transitions and byte-identical output** — a stronger statement than an aggregate match, since a tally can agree while individual cases swap
- the narrow-match control 00540, which a fall-through in the first attempt broke and which the traversal guards could not see: **a guard pinning which field you read says nothing about where control goes after you read it**
- reject-side probe on the fixed build: noncanonical `c2 81 00` gives status 1, canonical 33-byte scalar gives status 2, with the oversize case tightened from a noncanonical long-form to a canonical short-form so status 2 measures the over-32-byte decoder path specifically rather than being rejected for noncanonicality
- provenance separated in the body: attribution legs on pre-resolution tip `bb3d9d37e`, focused confirmation on post-resolution `85fb30b34`

The deliberately-omitted parent leg is the right call and is stated as such: the FA direction for an acceptance-widening change is a property of the new build alone, and the FR direction is already covered twice over.

Labelling once the build completes.

PR #10768, opened 2026-07-28
test: add EIP-7702 authorization chain-id reject probe

### LINKED ISSUE #10767: test: add reject-side EIP-7702 authorization chain-id probe
Follow-up to #10766.

The EIP-7702 authorization-chain-id widening repair needs a diagnostic-only
reject-side probe. Existing upstream malformed encoding cases are transaction
tests, while the stateless EEST corpus contains only valid blockchain fixtures;
the normal stateless harness cannot measure these reject paths.

The probe generator must be committed, not its ELF output. It should drive the
authorization field-selection plus strict U256 decoder for:

- a noncanonical scalar encoding; and
- a canonical scalar longer than 32 bytes.

Run it on the fixed artifact, recording nonzero reject statuses for both. This
directly measures the FA direction; a parent leg would only re-measure the FR
direction already covered by #10766's paired EEST A/B. Keep it separate from
#10766 so the one-fixture repair is not held behind a new diagnostic facility.

pirapira:
## Summary

Adds a diagnostic-only generator for #10766's reject-side EIP-7702
authorization-chain-id controls. It is deliberately separate from the repair:
the normal stateless EEST corpus contains accepted blockchain fixtures and
cannot represent malformed authorization scalars.

## Fixed-build probe evidence

- Noncanonical scalar list `c2 81 00` reaches the RLP list reader rejection:
  output status `1`.
- Canonical short-form 33-byte scalar reaches the strict U256 width rejection:
  output status `2`.
- Probe ELF `0cefd3b135d8150c6a95b2dba883cd63f10a588199c89709ff5c84a07f4f2416`,
  executed with `scripts/spike/spike_run`.
- `lake build EvmAsm.Codegen.Programs.AuthChainIdEncodingProbe
  EvmAsm.Codegen.Programs.Registry` passed.

The probe runs only on the fixed build. That directly measures the FA-adjacent
direction of an acceptance-widening change: malformed scalar encodings must
still reject. A parent leg would only re-measure the FR direction, already
covered by #10766's paired EEST A/B.

The first probe invocation faulted because its standalone stack pointer was
uninitialized. The prologue now initializes `sp` before invoking the shared
RLP helpers; the recorded statuses above are from the corrected generator.

Closes #10767.


pirapira:
Reviewed and labelled. CI green including reports, asm-to-program, region-map and axioms.

**The docstring is the reason this probe is worth committing**, and it says the two things that stop it being deleted later as redundant with the sweep:

- *"A valid EEST blockchain fixture cannot express a malformed authorization scalar"* — this is a **structural limit of the corpus**, not a coverage shortfall to grind down. Every fixture in a valid corpus is meant to be accepted, so no amount of sweeping reaches the reject side of an acceptance-widening change.
- *"A parent run would only re-measure the FR direction, which the repair's paired EEST A/B already covers"* — the omitted leg is recorded as a **decision with a reason**, not a silent gap.

Both cases drive the same field selection as the live path and then the strict U256 decoder: non-canonical `c2 81 00`, and canonical content longer than 32 bytes (`e2 a1` + 33×`01`). Both status words must be nonzero on the fixed ELF.

Worth noting the oversize case was **tightened during development** from a non-canonical long-form to a canonical short-form, so status 2 measures the >32-byte decoder path specifically. The original would have been rejected for non-canonicality and recorded a pass that never touched the width check — the dominant vacuity class on this stack, caught unprompted.

## One thing I checked before labelling

The probe adds to `.section .data` (`acip_noncanonical`, `acip_oversize`, `acip_off`, `acip_len`, `acip_scratch`). `.data` is linked at a fixed `0xa3000000`, and `Bn254FieldMulModSAsm` pins `bnf_le_a = 0xa3000ee0` by `rfl`, so `.data` growth can break address-pinned proofs in modules the author never touched — the cheapest check with the widest blast radius.

Verified on the branch: `bnf_le_a` is `0xa3000ee0`, unchanged from `main`. The probe's data does not displace it.

Stacked on #10766, which is already labelled. If #10766 lands first, merge its tip forward rather than re-authoring.

pirapira:
landed via #10774 (batch-merge of #10768 + #10772).

Verified in main at `1c799cdf3` rather than assumed — the branch head is **not** an ancestor of main, so the usual ancestry check does not settle it. The reason is benign: after #10774 was cut, this branch received one further commit, `2861079e8`, which is a pure merge of main into the branch and carries no content of its own. That sync commit is what keeps the head off main's history.

Content check on main:
- `EvmAsm/Codegen/Programs/AuthChainIdEncodingProbe.lean` — PRESENT
- `Registry.lean` entry `zisk_auth_chain_id_encoding_probe` — PRESENT (1 occurrence)

The residual diff against main shows only this branch's *older* copy of a `BalSerializer` docstring that #10772 deliberately rewrote in the same batch — superseded, not lost.

Issue #10770, opened 2026-07-28
ANSWERED: header BAL-hash is CONSTRUCTED from keccak(supplied bytes) and bound via the block-hash check — residual is bv_block_hash_check_enabled


pirapira:
Prerequisite for #10680, alongside #10764 and #10765. **This is a precise question, not a finding.** Nobody has traced it to an answer.

## What is confirmed

`hash_block_access_list` is `keccak256(rlp.encode(block_access_list))` — `block_access_lists.py:702-708`, execution-specs `e5a8caf1b`. RLP and keccak, exactly what the guest serializer produces. And `fork.py:366`/`:390` compute that hash **from the rebuilt BAL** and compare it against **`block.header.block_access_list_hash`**.

So the spec's own check is: rebuild from operations, RLP-encode, keccak, compare **to the header field**. That validates #10680's design at the source — an assumption that had been taken on trust until now.

## The question

`bal_serializer_verify` compares the rebuilt hash against `block_access_list_hash`, which computes keccak over the **supplied BAL bytes from the payload**. The spec compares against **the header field**.

Those are equivalent **only if something else already checks that the header field equals the hash of the supplied bytes.** If nothing does, comparing rebuilt-against-supplied proves the two agree *with each other* and says nothing about what the header committed to.

## What was traced, and where it stops

- `svf_bal_hash` receives the supplied-bytes hash at `BlockVerdictStateRoot.lean:772` and is published into the verdict params at offset **+96** (`:788`).
- `Step2Verdict.lean:65` documents `+96` as `block_access_list_hash`.
- `Step2Verdict.lean:178` writes the **header's** field, from payload offset 200, into the same slot on a **different path**.

So two producers feed one slot and a comparison **plausibly** exists downstream — but **that comparison has not been found and read.** "Plausibly exists" is exactly the inference that produced a retracted FA claim on #10764 earlier the same night.

A caution for whoever picks this up: offset `96` appears across many unrelated structs in `Programs/BlockVerdict*` with different base registers. An offset match is not evidence; the consumer's own ABI is. Read the routine that owns the base register, not the numeral.

## The two outcomes

- **If yes** (something checks header == keccak(supplied bytes)): `bal_serializer_verify` is transitively correct and the design is sound.
- **If no**: `verify` should take the **header field directly**, and the current design has a gap. It only matters once the comparator is wired — which is exactly the window in which to find it.

## Why it is filed rather than answered

[...]

PR #10771, opened 2026-07-28
docs(verdict): say at the flag what bv_block_hash_check_enabled actually gates (#10770)

pirapira:
`.dword 1` with no writer, so it is on today and only a source edit changes it. The
name says block-hash check. What it also gates is the ONLY binding between the
supplied block-access-list bytes and the header.

The guest never reads a header BAL-hash field -- it CONSTRUCTS one. `svf_bal_hash`,
keccak over the supplied BAL bytes, is passed as a7 to `block_header_ssz_to_rlp`
(`BlockHeaderSszToRlp.lean:77`: a7 = block_access_list_hash ptr) and embedded while
the header RLP is rebuilt; `block_hash_from_header` then hashes that RLP and compares
it byte-by-byte against the payload's block hash.

So with this flag at zero, nothing binds the supplied BAL to the header, and #10680's
rebuild-and-compare becomes SELF-REFERENTIAL -- verifying its own input against
itself. That is a much larger consequence than "the block hash is not checked", which
is how the flag reads where it is declared.

The dependency ran from this declaration, through an argument register, into a routine
two levels away, and NEITHER END MENTIONED THE OTHER. Four greps to establish and
invisible from either site. A comment is the whole fix; this is it.

Comment-only: the emitted guest is byte-identical before and after, verified with cmp
rather than assumed, since a comment in an asm-string file is exactly the edit that can
accidentally land inside a string literal.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

Byte-identity verified against **main** (80979 lines each, cmp-identical) — my first check used a baseline from another branch and warned; the corrected comparison is in the thread.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdictDataSection.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictDataSection.lean
@@ -54,6 +54,23 @@ def ziskStatelessVerdictV2DataSection : String :=
   ".balign 32\n" ++
   "bv_block_hash:\n  .zero 32\n" ++
   ".balign 8\n" ++
+  -- ON TODAY: `.dword 1` with NO WRITER anywhere in the guest, so only a source edit
+  -- turns it off. What it gates is larger than its name suggests.
+  --
+  -- It guards the block-hash comparison at `BlockVerdictFunction.lean:70`. That
+  -- comparison is ALSO the only thing binding the supplied block-access-list bytes to
+  -- the header: the guest never reads a header BAL-hash field, it CONSTRUCTS one --
+  -- `svf_bal_hash` (keccak over the supplied BAL bytes) is passed as a7 to
+  -- `block_header_ssz_to_rlp` (`BlockHeaderSszToRlp.lean:77`, a7 = block_access_list_hash
+  -- ptr) and embedded while the header RLP is rebuilt. `block_hash_from_header` then
+  -- hashes that RLP and compares it byte-by-byte against the payload's block hash.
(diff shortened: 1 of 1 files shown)


pirapira:
Reviewed and labelled. Comment-only, and the comment earns its place.

It names all four links of a dependency that was invisible from both of its ends: the flag, the comparison site it guards (`BlockVerdictFunction.lean:70`), the `a7` hand-off with the callee's own ABI cited (`BlockHeaderSszToRlp.lean:77`) rather than inferred from argument position, and the consequence — with the flag at zero, nothing binds the supplied BAL bytes to the header, so a rebuild-and-compare check over that BAL becomes self-referential, verifying its own input against itself.

That consequence is much larger than the flag's name suggests, which is exactly why it needed writing down. The dependency ran from a data declaration in one file, through an argument register, into a routine two levels away, and neither end mentioned the other — it took four greps to establish and was unreadable from either site alone.

Verified independently: one declaration (`BlockVerdictDataSection.lean:57`, `.dword 1`), one reader (`BlockVerdictFunction.lean:70`), **no writer anywhere in the guest**, so the flag is on today and only a source edit changes that.

Context: #10770, which this answers, was raised as a precise question rather than a finding — whether comparing rebuilt-against-supplied says anything about what the header committed to. The answer turned out to be that the guest never reads a header BAL-hash field at all; it constructs one, and the block-hash comparison is what binds it. So the serializer comparator is transitively correct, conditional on this flag.

Issue #10769, opened 2026-07-28
Guard class: pin return codes against their documented contracts (bidirectional pair on verify/rebuild_hash)


pirapira:
## Origin

A reviewer (coord) filed a defect on #10761 claiming `bal_serializer_rebuild_hash` failed to normalise the canonical-sort status to `2`. It was wrong and was retracted: there are two routines with two contracts, composing correctly.

- `bal_serializer_rebuild_hash` (`BalSerializer.lean:957`) — *"`a0` (out) = 0, or the canonical sort's own nonzero status."* Propagating the raw code **is** its contract.
- `bal_serializer_verify` (`:1044-1046`) — 0 match / 1 no match / 2 sort failed. Its body does `beqz a0, .Lbsv_rebuilt` then `li a0, 2`, collapsing any nonzero to `2` and leaving the finer code in `bal_serializer_sort_status`.

**The misread was possible because nothing in the code states which routine owns which contract.** Proximity substituted for attribution: the emitted sequence was read together with the nearest docstring, across a `ret` boundary. No existing guard covers this — the guard set pins emitted text and field selection, never **a return code against its documented contract**.

## The guard class

A **bidirectional pair**, and the pair matters more than either half:

1. A guard asserting `bal_serializer_verify`'s bail path **contains** `li a0, 2` — stating the normalisation is deliberate, so it is not deleted as redundant.
2. A guard asserting `bal_serializer_rebuild_hash` **does not contain** that normalisation — stating the two-routine split is intentional rather than an oversight.

Per the negative-guard lesson from the `andi 255` fix on the same stack, **the second is the more valuable**: absence of normalisation in `rebuild_hash` is site-independent, whereas presence of `li a0, 2` in `verify` could be satisfied by any such instruction anywhere in the def.

Note the general form: a positive guard on a shared instruction sequence cannot distinguish *which* site satisfies it. That is the same weakness that let a positive `#guard` on `BalCanonicalSort.lean` both miss its bug and block the repair.

## Docstring change alongside

Both docstrings should **name the routine in every sentence** rather than opening with a bare `a0 (out) = …`. That bare form is precisely what let proximity substitute for attribution, and it costs nothing to fix.

## Scope

Filed rather than pushed onto #10761, which is clean and labelled. Worth applying to the return-code contracts of the wider serializer surface, not just these two — the same "which routine owns this contract" ambiguity exists anywhere a split routine documents a status.

pirapira:
[...]

PR #10772, opened 2026-07-28
test(bal): guard the serializer return codes against their contracts (#10769)

pirapira:
A guard class this file did not have. Every other guard here pins emitted text or
field selection; none pinned what a routine RETURNS against what its docstring says.

That gap was not hypothetical. A reviewer read `bal_serializer_verify`'s 0/1/2
contract as applying to `bal_serializer_rebuild_hash`'s bail path and reported a
defect that was not there -- the two contracts sit twelve lines apart, both describe
an a0-out with small integer codes, and nothing in the code said which routine owned
which. Proximity substituted for attribution.

Two guards, deliberately opposite in polarity:

  verify NORMALISES -- pinned as the PRESENCE of `li a0, 2` on its bail path. Without
  it the conversion reads as redundant (rebuild_hash already returns nonzero) and
  deleting it silently widens verify's contract to leak sort codes 1 and 3, where 1
  COLLIDES with its own "hash does not match".

  rebuild_hash does NOT normalise -- pinned as the ABSENCE of any `li a0, 2`. Absence
  is site-independent; a presence guard could be satisfied by an unrelated `li a0, 2`
  elsewhere in the def, which is exactly how the andi-255 guard failed.

CONTROLLED IN BOTH DIRECTIONS: deleting verify's normalisation fires a guard, and
making rebuild_hash normalise "for consistency" fires a guard. Both are the plausible
edits, which is why they are the ones worth blocking.

Also renames both contracts to say the ROUTINE in every sentence rather than
"a0 (out) = ...", so the next reader cannot attach a docstring to the wrong function.

Emission byte-identical to main, verified against an origin/main worktree AFTER the
check resolved rather than claimed alongside it.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


pirapira:
Reviewed. Labelling once build completes — it is the only pending check.

The two guards are exactly the design from #10769, and **the positive one is stronger than its own justification claimed**:

```
#guard (balSerializerVerifyFunction.splitOn "li a0, 2; j .Lbsv_ret").length == 2
#guard (balSerializerRebuildHashFunction.splitOn "li a0, 2").length == 1
```

The concern raised on #10769 was that a presence guard "could be satisfied by an unrelated `li a0, 2` elsewhere in the def" — which is how the `andi 255` guard failed on this same stack. This one pins `li a0, 2; j .Lbsv_ret`, **instruction plus its branch target**, so it is satisfiable only by the bail path it exists to protect. The weakness was anticipated and closed rather than accepted.

The negative guard is the load-bearing half, as intended: absence of any `li a0, 2` in `rebuild_hash` is site-independent, and it fires on the specific edit a future reader is most likely to make — adding a normalisation "for consistency" with `verify`.

Controlled in both directions, which is what makes them non-vacuous: deleting `verify`'s conversion fires, adding one to `rebuild_hash` fires. Those are the two plausible edits, and both were observed rather than argued.

## Why this guard class exists at all

A reviewer (me) filed a defect here that did not exist, attributing `verify`'s 0/1/2 contract to `rebuild_hash`, whose contract is *"0, or the canonical sort's own nonzero status."* The code was correct and stayed correct. **But the gap was real even though the defect was not:** two contracts twelve lines apart both described an `a0`-out with small integer codes, and nothing in the code said which routine owned which. Renaming both contracts to name the routine in every sentence is the cheap half; these guards are the other half.

Note the consequence the positive guard protects, because it is not obvious: delete `verify`'s conversion and it leaks sort codes 1 and 3 — and **1 collides with `verify`'s own "hash does not match"**, so a sort failure would be reported as a mismatch. The conversion reads as redundant, which is exactly why it needs pinning.

## Process notes worth keeping

- **Control B did not run the first time** — the perturbation pattern did not match the file, and only an assertion caught it. Without the assert, `guards fired: 0` would have read as a control that ran and found nothing. Fourth unasserted string operation to nearly mislead in one session, and the first one caught before anything was reported.
- **Byte-identity verified against main, and the result read, before the claim was written into the commit** — the correction from #10771 applied rather than merely recorded.

pirapira:
landed via #10774 (batch-merge of #10768 + #10772).

Verified in main at `1c799cdf3` rather than assumed — the branch head is **not** an ancestor of main, because after #10774 was cut this branch received `50299cf5a`, a pure merge of main into the branch with no content of its own. (That is the sync coord reassigned to scoper; I measured it as content-free before letting #10774 proceed, so the batch was never stale in substance.)

Content check on main — both return-code guards present:
- `#guard (balSerializerVerifyFunction.splitOn "li a0, 2; j .Lbsv_ret").length == 2`
- `#guard (balSerializerRebuildHashFunction.splitOn "li a0, 2").length == 1`

The residual diff against main shows only #10768's probe, which this branch never had — expected for a sibling in the same batch, not a loss.

Issue #10765, opened 2026-07-28
Classify all 11 granular BAL reject labels: which are implied by digest agreement, which the convergence would remove


pirapira:
Prerequisite for #10680, alongside #10764 and #10695.

#10680 replaces the guest's granular BAL checks with a single digest comparison over a rebuilt BAL. That is only safe for a check whose property is **implied by hash agreement over the rebuilt BAL**. Where it is not, the convergence silently removes a check that fires today.

## The inventory (measured on the emitted guest, not read off a doc)

**Eleven distinct reject labels, 17 gating branches:**

| label | branches |
|---|---|
| `allaccounts` | 2 |
| `code_consistent` | 1 |
| `code_covers` | 1 |
| `gas` | 1 |
| `nonstorage_covers` | 2 |
| `nonstorage` | 3 |
| `reads` | 1 |
| `recipient_field` | 2 |
| `storage_mismatch` | 1 |
| `storage_omit` | 1 |
| `tuple` | 2 |

Note: #10612 cites "21 field checks". That count has **not** been reconciled with these eleven labels, deliberately — the mapping is unknown and reconciling two numbers derived by different people is how a wrong one becomes load-bearing. Eleven is what the emitted guest branches to.

## The work

For each of the eleven: **does hash agreement over the rebuilt BAL imply the property this check enforces?**

Where the answer is no, the convergence loses a check and that check must be preserved or its coverage restored before #10680 removes it.

### Already known: two are NO

- **`nonstorage` (3 branches)** — covers withdrawal credits at BAI `N+1`, which a rebuilt BAL cannot represent (#10764).
- **beacon-roots BAI-0** — rejects via `.Lbsr_cons_sys4788`, which is **not** in this family at all; it sits on the state-root side. So BAI-0 coverage lies outside the eleven, and the inventory above does not account for it.

### Nine unexamined

Not guessed at, deliberately. Some look plainly representable (`storage_mismatch`, `reads` compare content the rebuilt BAL contains). Others do not: `gas` is not a BAL-content check at all, and `allaccounts` and `tuple` read as structural properties over the whole list rather than per-field values.

## Why this is filed rather than finished

This is bounded work, but someone will rely on it to decide what #10680 may safely remove. A tired wrong answer here is worse than an absent one — a check classified representable when it is not gets deleted, and nothing downstream will notice.

Found by scoper, which enumerated the inventory and then stopped short of the classification for that reason.

pirapira:
[...]

Issue #10764, opened 2026-07-28
FA (live, class wider than withdrawals): declared BAL balance changes are unvalidated on the MTx route — recipient credit unrecorded and skip-listed


pirapira:
## The finding

Found by **scoper** while enumerating what had been taken on trust, not from a failing test.

`block_access_index` has three producers in the spec. The guest implements one:

| BAI | assigned at | covers | guest producer |
|---|---|---|---|
| `0` (default) | builder init | `BEACON_ROOTS_ADDRESS` (EIP-4788), `HISTORY_STORAGE_ADDRESS` (EIP-2935) — `fork.py:897`, `:903` | **none** (scoper-measured) |
| `i + 1` | `fork.py:1040` | each transaction | `BlockVerdictMtxRuntime.lean:407`,`:448` — `bv_mtx_i + 1`, correct |
| `ulen(transactions) + 1` | `fork.py:917-918` | `process_withdrawals` (`:921`) and `process_general_purpose_requests` (`:923`) | **none** (scoper-measured) |

Spec ref: execution-specs `e5a8caf1b` (Amsterdam).

## Why this is a soundness ceiling, not a missing feature

Nothing stamps `0` or `N+1`, so system-level and post-execution changes **cannot appear in a rebuilt BAL at all**. That does not merely omit rows — it bounds the digest comparison that the whole congruence approach rests on:

> A block whose only divergence is a system-level or post-execution change is **ACCEPTED**, because neither side of the comparison can represent it.

An agreeing hash is only as strong as the set of changes both sides can express. This is a ceiling the comparator cannot raise by working correctly, and it is invisible to every FA sweep, because a sweep can only exercise divergences the encoding admits.

## The unrepresentable set is not a corner case

Reading the call sites, index `0` and index `N+1` cover effects present in essentially every well-formed block:

- **BAI 0** — the beacon-roots ring-buffer storage write and the history-storage parent-hash write. Both unconditional, every block.
- **BAI N+1** — every withdrawal balance credit (`process_withdrawals` increases the withdrawing account's balance, `incorporate_tx_into_block` at `:1226`), plus the EIP-7002 / EIP-7251 / EIP-6110 request-contract effects that run after the index is set.

So the gap is not an exotic path — it is the per-block system writes and the withdrawal credits.

## Interaction with the state root

[...]

PR #10775, opened 2026-07-28
docs: mark BAL system skip as convergence dependency

pirapira:
Documents at the live EIP-2935/EIP-4788 skip that it is load-bearing until a rebuilt BAL includes BAI-0 system rows.\n\nNon-emitting comment-only change; no regeneration or sweep required.\n\nRefs #10765 and #10764.


pirapira:
Reviewed. Comment-only, placed at the right site — directly above the `bal_account_is_modeled_system` skip in `balAllAccountsStorageConsistentFunction`, where the existing soundness note already lives. The wording is precise about what the dependency is: a digest built only from per-tx rows cannot replace the explicit system replay.

**One gap, and it is the same class the note exists to prevent.**

The skip is not at one site. `bal_account_is_modeled_system` is called from **two** places:

- `BalAllAccountsStorage.lean:89` — where this note now sits
- `BalAccountRecordArray.lean:119`, gated by `bara_skip_modeled_system` (declared `:292`, documented `:39`)

The existing comment here even names the second one — *"mirrors `BalAccountRecordArray`'s `bara_skip_modeled_system`"* — so the relationship was known, but the dependency note is only on one end of it.

That matters because the note's whole purpose is to stop the skip being retired independently. A future reader working in `BalAccountRecordArray` sees a skip flag with a docstring explaining *what* it does and nothing saying it is load-bearing for the convergence, retires it, and the note in the other file never enters their view. **A warning placed at one of two equivalent sites protects one of two.**

Suggest the same three lines at `BalAccountRecordArray.lean` beside `bara_skip_modeled_system`, cross-referencing this site so each end names the other. Same reasoning as #10770's flag note: the dependency there ran from a data declaration through an argument register into a routine two levels away, and **neither end mentioned the other** — which is exactly why it took four greps to establish and was unreadable from either site alone.

Not blocking — one site noted is strictly better than none, and this is comment-only with no emitted bytes. But it is a five-minute addition that closes the class rather than one instance.

Issue #10777, opened 2026-07-28
Do BAL slot payloads have a unique canonical representation? Raw-byte vs decoded-U256 comparison could make one slot read as two


pirapira:
Split out of #10776, because it is broader than that issue's counting question and may point the other direction.

## The question

`bal_gas_valid` walks **raw RLP payloads** for slots. `bal_serializer_slot_eq` compares **builder-native fixed 32-byte rows**. Those are different representations of the same thing, which raised the question while scoping #10776's dedupe:

**Do accepted BAL slot payloads have a unique canonical representation, or can two distinct byte sequences denote the same U256 slot?**

RLP scalars must be canonical (no leading zeros), and the guest has `rlp_content_to_u256_be` which rejects non-canonical content. But whether *every* path that reads a slot from a supplied BAL enforces that has not been established.

## Why it may matter beyond a gas count

If a non-canonical slot encoding is accepted anywhere, then one logical slot can appear under two distinct byte sequences, and **any comparison keyed on raw bytes rather than on the decoded U256 would treat them as two different slots.** That is not confined to #10776's item count — it would reach the storage comparators:

- a declared `storage_changes` entry and an exec-log row for the same slot could fail to match, or
- a duplicate declared entry could evade a uniqueness or coverage check

The first direction is a false reject. **The second could be a false accept**, which is why this is filed separately rather than as a note on a deferred FR.

## What is NOT claimed

Nothing here is established. Specifically not checked:

- whether any BAL slot read path accepts non-canonical content
- whether any comparator keys on raw payload bytes rather than a decoded 32-byte value
- whether the RLP layer already rejects such payloads before any comparator sees them (which would close this entirely)

**The third is the likely answer** — canonical-scalar enforcement at decode is the norm in this codebase, and if it holds at every BAL slot read the question is closed with nothing to fix. That should be checked first, since it is the cheap disconfirmation.

## Method note

Surfaced by codex while scoping #10776's dedupe algorithm — the observation that `bal_serializer_slot_eq` is not reusable there, because the two routines work on different representations. A representation mismatch between two routines handling the same value is worth a second look on its own, independent of the fix that surfaced it.

Related: #10776 (deferred; the counting question that surfaced this), #10765 (`gas` classified NO, so `bal_gas_valid` survives the convergence).

pirapira:
## Confirmed, and the consequence is the OPPOSITE of this issue's hypothesis

[...]

PR #10778, opened 2026-07-28
docs: explain lenient BAL slot payload binding

pirapira:
## Summary

Documents why `bal_storage_change_values` deliberately accepts RLP slot/value
payloads of length at most 32 and zero-left-pads them instead of requiring a
canonical scalar encoding.

The comment records the cross-file safety condition: supplied BAL bytes are
committed through the rebuilt header's BAL hash and checked by
`block_hash_from_header` against the payload block hash. It also records the
conditional gate, `bv_block_hash_check_enabled`, and references #10771 and
#10777.

No emitted assembly or behavior changes.

## Validation

- `git diff --check`

### DIFF
--- a/EvmAsm/Codegen/Programs/BalStorageChangeValues.lean
+++ b/EvmAsm/Codegen/Programs/BalStorageChangeValues.lean
@@ -15,6 +15,21 @@
   writes superseded within the block.
 
   Both keys and values are emitted as 32-byte big-endian, left-padded.
+
+  ## Deliberately lenient slot payloads
+
+  Slot keys and post values here accept any RLP payload of length at most 32 and
+  zero-left-pad it; this reader deliberately does not require a shortest scalar
+  encoding or call `rlp_content_to_u256_be`.  The local normalization is safe
+  only because the supplied BAL bytes are committed into the rebuilt header via
+  its BAL hash, and `block_hash_from_header` compares that header commitment
+  with the payload block hash.  A non-canonical supplied payload therefore
+  cannot survive the block-hash binding even when it normalizes to the same
+  32-byte key/value as a canonical payload.
+
+  This is a cross-file dependency, not a property of this reader.  The binding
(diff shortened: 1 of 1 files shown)


pirapira:
Reviewed. Comment-only, and it says the three things that make it worth committing rather than being deleted later as noise:

1. the leniency is **deliberate** — any payload of length ≤ 32, zero-left-padded, with no shortest-encoding requirement and no `rlp_content_to_u256_be`
2. it is safe **because of a binding two levels away** — the supplied BAL bytes are committed into the rebuilt header via its BAL hash, and `block_hash_from_header` compares that against the payload block hash, so a non-canonical payload cannot survive even though it normalises to the same 32-byte value
3. that binding is **conditional on `bv_block_hash_check_enabled`**, cross-referencing #10771 and #10777, with an instruction not to tighten or rely on the acceptance without re-auditing the gate

Note it covers slot keys **and post values**, which is broader than the read that prompted it — correct, since the same padding applies to both.

## Why this class of comment keeps earning its place

This is the second instance of the same defect shape, after #10771. A safety property lives in one file and the code that depends on it lives in another, with **neither end naming the other**. On #10770 that took four greps to establish and was unreadable from either site alone. Here it cost a reviewer several minutes deciding whether a non-canonical accept was a false accept — the answer being that the block-hash binding makes it unreachable, which nothing at the read site said.

A local check would make the safety local. This comment makes it *legible*, which is the cheaper half and the one that survives someone deleting the binding.

## Process note worth recording

The author declined to commit this on the shared checkout's dirty branch (`diag/account-builder-matrix`), which carried seven other modified files of indeterminate ownership. A commit there would have captured another lane's uncommitted work — the same failure mode as a blanket conflict resolution, where content moves silently and the PR body describes something else. Instead: an ownership check (only this file was the author's), a clean `git worktree` off `origin/main`, and removal of only that worktree, leaving the shared tree untouched.

`BEHIND` main at `ccdf3d969` and awaiting checks; no conflict expected on a single comment block.

PR #10785, opened 2026-07-28
test(witness): GH #10764 withdrawal-credit false-accept witness

pirapira:
Diagnostic-only witness deciding the GH #10764 false-accept claim. No guest behavior change; adds a generator script plus a report.

## Result: FALSE ACCEPT CONFIRMED

Fixture: `tests-zkevm@v0.6.2` `bal_withdrawal_and_value_transfer_same_address` (1 tx, one 10 gwei withdrawal to `0xc0f6dc9e5836f54caadbf59cc69346c508e1992b`).

Mutation (minimal, length-preserving):
1. BAL BAI-2 `postBalance` 15,000,000,000 -> 5,000,000,000 at blob offset 1051 (the declared BAL credits the withdrawal address nothing for the withdrawal).
2. Re-pin `payload.state_root` to the guest-derived post-state root `0xa73c7635...83b726`.
3. Re-pin `payload.block_hash` to the recomputed header hash `0xd8a99552...ea3bf3`.

Re-pinning to guest-derived values makes the block self-consistent with the guest's own model by construction — that is what a false accept means here. The guest's post-state root is BAL-fed, so pinning makes the guest echo the header's oracle (the circularity the FA exploits). The diff is one declared value plus two forced re-pins.

## Numbers (pristine ELFs from main @ cb5afc7f8, spike)

- execution-specs `run_stateless_guest`: REJECTS — succ byte 0; bare re-run raises `InvalidBlock` from `execute_block` (`fork.py:379`).
- `stateless_guest`: ACCEPTS — succ byte 1.
- `zisk_stateless_verdict_v2` probe: verdict 1, bv_fail 0, header_status 0, state_status 0.
- Compared pair: declared 5,000,000,000 wei vs true 15,000,000,000 wei.

Conclusion: on the MTx route nothing validates withdrawal credits against execution; the only reconciler (`block_verdict_withdrawal_nonstorage_effects`) is reachable solely from the dead legacy route (tx count 0).

## Notes

- The guest-derived root was extracted with a diagnostic no-hash probe (`bv_block_hash_check_enabled` flipped in the *emitted* `.s` only; the tree is untouched) because the block-hash check runs before `block_state_root` and otherwise masks `sv_recomputed`. All final numbers above come from the pristine ELFs.
- Generator: `scripts/witness-10764-withdrawal-fa.py`; full construction and reproduction recipe in its docstring and in `docs/agents/witness-10764-withdrawal-fa.md`.

### DIFF
--- a/docs/agents/witness-10764-withdrawal-fa.md
+++ b/docs/agents/witness-10764-withdrawal-fa.md
@@ -1,0 +1,69 @@
+# Witness: GH #10764 withdrawal-credit false accept (fixture 00565)
+
+Status: **FALSE ACCEPT CONFIRMED** (diagnostic only; no guest behavior change).
+
+Generator: `scripts/witness-10764-withdrawal-fa.py` (construction documented in
+its docstring). All final numbers below were produced by pristine ELFs emitted
+from `main` @ cb5afc7f8 via `lake exe codegen --program <p> --halt linux93`,
+run under `scripts/spike/spike_run`.
+
+## Fixture
+
+`gen-out/eest-fixtures/tests-zkevm@v0.6.2/fixtures/fixtures/blockchain_tests/for_amsterdam/amsterdam/eip7928_block_level_access_lists/block_access_lists_eip4895/bal_withdrawal_and_value_transfer_same_address.json`
+
+One block, one transaction, one withdrawal of 10 gwei
+(10,000,000,000 wei) to `0xc0f6dc9e5836f54caadbf59cc69346c508e1992b`.
+That address declares two BAL balance changes: BAI 1 -> 5,000,000,000 wei
+(the tx value transfer) and BAI 2 -> 15,000,000,000 wei (after the
+withdrawal credit).
+
+## Mutation (minimal)
+
+1. BAI-2 `postBalance` 15,000,000,000 -> 5,000,000,000 (one 5-byte RLP
+   string replaced by an equal-length one at blob offset 1051; blob length
+   unchanged). The declared BAL now credits the withdrawal address nothing
+   for the withdrawal.
+2. Re-pin `payload.state_root` (blob offset 114) to the guest-derived
+   post-state root `0xa73c763559709d17bd19c12fb20ee647e378f4ee1232f8bb307bb2a8bb83b726`.
+3. Re-pin `payload.block_hash` (blob offset 534) to the recomputed header
+   hash `0xd8a995521d4c46beba61bfca53fa815edbb98a1b83f6ed60e8531020c6ea3bf3`
+   (keccak of the header RLP with the mutated BAL hash and re-pinned state
+   root).
+
+Re-pinning to guest-derived values is mandatory: it makes the block
+self-consistent with the guest's own model by construction, which is exactly
+what a false accept means here. The guest's post-state root is BAL-fed, so
+pinning makes the guest's output echo the header's oracle (the circularity
+this FA exploits). The diff is one declared value plus two forced re-pins.
+
+## Results
+
+- **execution-specs** (`run_stateless_guest`, submodule e5a8caf1b =
+  tests-zkevm@v0.6.2): **REJECTS** — output succ byte = 0. Re-running the
+  `verify_stateless_new_payload` body without its blanket `except` raises
+  `ethereum.exceptions.InvalidBlock` from `execute_block`
+  (`fork.py:379`, the block-diff / BAL-vs-state check inside
+  `execute_new_payload_request`).
+- **Guest** (`stateless_guest` ELF, spike): **ACCEPTS** — output succ byte = 1.
+- **Verdict probe** (`zisk_stateless_verdict_v2`, pristine): verdict = 1,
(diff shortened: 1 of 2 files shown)


pirapira:
Reviewed. **The experiment is sound and the result stands.** Two documentation corrections required before merge — no code changes.

## What makes this trustworthy

- **The success bit is read at the right offset.** `out[32]` / `guest_out[32]` on both sides. This was a live hazard: another agent found its own harness reading byte 0 and had to recompute a 200-case gate. This generator was already correct.
- **The mutation is minimal and verified unique.** One 5-byte RLP string, `85 037e11d600` → `85 012a05f200`, verified to occur **exactly once** at blob offset 1051, in the quoted context `c7 01 85 012a05f200 c7 02 85 037e11d600 c0c0`. Same length, so no SSZ offsets move.
- **The instrument is validated in the same run that uses it.** On the step-1 blob the nohash probe reports `verdict=0 bv_fail=1` — a state-root mismatch, the *expected* failure — with `header_status=0` and `state_status=0`, **proving `block_state_root` ran** before `sv_recomputed` was read. A probe whose only output is the value you want cannot be trusted; this one hits a known expectation first.
- **The flag flip is scoped and disclosed.** `bv_block_hash_check_enabled` flipped in the emitted `.s` only, with the full `as`/`ld` invocation quoted, tree untouched, used *solely* to extract the root — and all final numbers taken from pristine ELFs.
- **The circularity is stated rather than hidden:** *"the mutated block is self-consistent with the guest's model BY CONSTRUCTION; that is what false accept means here."* That is the correct framing and it pre-empts the obvious objection.
- **The refutation branch is written down** — guest rejects ⇒ claim withdrawn, "equally useful." A witness that can only confirm is not a test.
- **Why only two pins are needed** is explained: the declared `blockAccessListHash` never appears in the blob, because the guest recomputes it from the supplied BAL bytes.

## Required correction 1 — the routine is a RECORDER, not a reconciler

The docstring says `block_verdict_withdrawal_nonstorage_effects` is *"the routine that reconciles withdrawal balance credits against the declared block access list (BAL)"*.

It does not reconcile anything. `BlockVerdictWithdrawalEffects.lean:40-43` computes `amount × 1e9 + post` and calls **`record_nonstorage_effect`**, returning `0`/`1` for success/internal failure. The reconciliation is `bal_all_accounts_nonstorage_consistent` / `_covers`.

**This error is mine** — I briefed it that way, having inherited the phrasing from a third report and quoted it forward without reading the routine's body. Correcting it here matters because a committed generator docstring is a durable artifact that will outlive the messages.

Suggested: *"…the routine that RECORDS withdrawal balance credits into the non-storage effect log via `record_nonstorage_effect`; the reconciliation against the declared BAL happens downstream in `bal_all_accounts_nonstorage_consistent`/`_covers`."*

## Required correction 2 — the "open question" is now answered, and the answer is sharper

The docstring poses: *"Open question: does anything else on the MTx route validate withdrawal credits?"* That has since been resolved, and not as "nothing does":

**The comparators DO run on MTx.** `BlockVerdictMtxTail.lean:286-289` aggregates `exec_nonstorage_effect_log` into `exec_nonstorage_effect_agg`; `:291-295` calls `bal_all_accounts_nonstorage_consistent` with the declared BAL (`bv_bal_start`/`bv_bal_len`) against that agg; `:297-301` calls `_covers`. Both gate `.Lbv_bal_nonstorage_fail`.

**They compare against a log that has no withdrawal rows in it**, because the recorder never runs on this route. So the failure is not an absent comparator — it is a **live comparator reading an incomplete input**, which is a materially different defect and a stronger statement. It also explains why the repair is to call the recorder rather than to add a check.

## Verdict

Diagnostic-only, +483/−0, two files, no guest behaviour touched. **Labelling once the two docstring corrections are in.** Both are prose; the measurement and the code are unaffected.

This is the artifact that resolved a soundness claim which had been retracted and then reversed. Worth its length.

pirapira:
Corrections verified at `4979c03a1` and labelled. Read them directly: line 14 now says the routine **RECORDS**, `:16` names `record_nonstorage_effect` and puts the reconciliation downstream, and `:21-24` carry the answered form — `MtxTail:286-289` aggregating, the comparators running live but reading a log with no withdrawal rows.

Since this PR was opened, its own finding has grown considerably, and the generator is the reason:

- **The false accept is confirmed** and this is the artifact that confirmed it, after the claim had been retracted once and reversed once.
- **The obvious repair turned out to be inert.** Calling the withdrawal recorder on the MTx route left the witness accepted, because of three further gates: the A2a comparators skip the transaction recipient via `bv_mtx_skip_list`, the tail bypasses B2.3 when withdrawals exist, and the recorder itself writes a **wrong value** (10e9 rather than 15e9, from a zero baseline).
- **Those gates are one defect, not three.** On pristine main the effect log has no row for the transaction's credit to the recipient at all — so the skip-list entry exists *because* the row is missing. One absent producer plus two accommodations built around its absence. Remove an accommodation alone and honest fixtures false-reject; add the producer without removing them and nothing changes.

None of that was visible before this witness existed. The full chain is recorded on #10764.

Issue #10786, opened 2026-07-28
EOA route never records the transaction recipient's balance credit — root of #10764, and the reason bv_mtx_skip_list skips the recipient


pirapira:
Root cause of the withdrawal false accept on #10764, filed separately because it is a distinct missing producer with a broader consequence than withdrawals.

## Measured on pristine main

Diagnostic ELF derived from the pre-helper image (same source and artifact as the parent, with only the N+1 helper call removed in the diagnostic assembly). Fixture: `bal_withdrawal_and_value_transfer_same_address`, one transaction transferring 5 gwei to `0xc0f6dc9e5836f54caadbf59cc69346c508e1992b`.

**Block `account_writes` builder rows:**

| address | BAI | post |
|---|---|---|
| `f6c3…` (sender) | 1 | 1000000000000000001197954000 |
| `2adc…` (coinbase) | 1 | 613800 |
| **`c0f6…` (recipient)** | — | **absent** |

**Raw `exec_nonstorage_effect_log`, count = 3:** sender gas debit, sender refund, coinbase fee. **No recipient row.**

So the transaction's credit to its recipient is recorded in **neither container** — not the effect log, not the block account-writes map.

## Why: the EOA route bypasses the only stager

`BlockVerdictDispatchTx` stages a recipient credit at `:969-972`. **The EOA MTx route does not go through `BlockVerdictDispatchTx`**, so that staging never runs. And staging it earlier does not work either: the EOA runtime dispatcher setup **clears `account_state_pending_count`** before invoking the shared seed callback (`Dispatch.lean`, `emitRuntimeDispatcherCallableSetup`), so a pre-dispatch AccountState tuple cannot survive to be read.

## The skip list is an accommodation for this gap, not an independent choice

`bal_all_accounts_nonstorage_consistent` / `_covers` take `bv_mtx_skip_list` / `bv_mtx_skip_count` as `a4`/`a5` (`BlockVerdictMtxTail.lean:293`, `:299`), **and that list contains the transaction recipient.**

It has to. Execution credits the recipient and the effect log has no row for it, so without the skip entry `_covers` would fail on **every plain transfer**. The tail's B2.3 bypass when withdrawals exist is a second accommodation of the same shape.

**So this is one absent producer plus two accommodations built around its absence.** Remove either accommodation alone → honest fixtures false-reject. Add the producer without removing them → nothing changes. That is why the repair on #10764 is four ordered steps rather than one.

[...]

PR #10787, opened 2026-07-28
test(witness): add bai1 intermediate-BAI variant to GH #10764 witness

pirapira:
## Summary

Follow-up to #10785. One generator, two variants of the same false-accept class on fixture `bal_withdrawal_and_value_transfer_same_address` (tests-zkevm@v0.6.2), measured with the same pristine ELFs as the first witness (diagnostic only; no guest behavior change).

**bai1 variant**: mutate the intermediate BAI-1 `postBalance` (the ordinary tx credit) 5,000,000,000 -> 25,210,765,039 wei (`0x5deadbeef`), an equal-length 5-byte RLP string at blob offset 1043, anchored on row context (the 5e9 string also occurs inside the tx RLP and is untouched).

- **execution-specs REJECTS**: succ = 0; bare re-run raises `InvalidBlock` "Invalid block access list hash" at `fork.py:391` (BAL hash comparison) — a *different* spec check than bai2's `fork.py:379` state-root comparison.
- **Guest ACCEPTS**: succ = 1; pristine verdict probe verdict = 1, bv_fail = 0.
- Compared pair: declared 25,210,765,039 vs true 5,000,000,000 wei.

**Structural claim** (foregrounded in the report): a final-balance row is constrained by both the state root and the BAL hash; an intermediate row is constrained by the BAL hash alone — therefore any check built only on the state root is blind to intermediate rows. Here the state-root re-pin was a no-op (`sv_recomputed` == original declared root; BAI-2 pins the final balance) and only the `block_hash` re-pin is load-bearing.

Also included: the attacker clarification (re-pinning is a fixture-mutation cost, free to an attacker who constructs the block; both variants equal severity), and the regression note that bai2 numbers were re-verified identical after the `--variant` refactor (`0xa73c76...`, `0xd8a995...`).

Refs #10764, #10786.

### DIFF
--- a/docs/agents/witness-10764-withdrawal-fa.md
+++ b/docs/agents/witness-10764-withdrawal-fa.md
@@ -81,3 +81,65 @@ untouched) reaches `block_state_root` and reports the expected state-root
 mismatch signature (verdict=0, bv_fail=1, statuses 0) with
 `sv_recomputed` at OUTPUT+168. The pristine ELFs were used for every final
 number above.
+
+## Variant bai1: intermediate BAI row (tx credit)
+
+Both variants come from one generator (`--variant bai2|bai1`), run against
+the same fixture and the same pristine ELFs. After the refactor that
+introduced `--variant`, the bai2 numbers above were re-verified identical
+(`0xa73c76...`, `0xd8a995...`), so the shared script did not drift between
+the two results.
+
+Mutation: BAI-1 `postBalance` (the ordinary tx credit to the same address)
+5,000,000,000 -> 25,210,765,039 wei (`0x5deadbeef`), an equal-length 5-byte
+RLP string at blob offset 1043. The edit is anchored on row context because
+the 5e9 RLP string also occurs at offset 636 inside the tx RLP (the transfer
+value), which is left untouched.
+
+Results:
+
+- **execution-specs**: **REJECTS** — succ byte = 0; bare re-run raises
+  `ethereum.exceptions.InvalidBlock` "Invalid block access list hash" at
+  `fork.py:391` inside `execute_block` (computed BAL hash vs header
+  `block_access_list_hash`). A different spec check than bai2's
+  `fork.py:379` state-root comparison: the spec's BAL-vs-execution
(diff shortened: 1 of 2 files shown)


pirapira:
Reviewed the content. **Both things I asked for are in, verified by reading rather than by report.**

- **Structural claim foregrounded**, `:114-116`, in bold: *"…constrained by the state root and the BAL hash; an intermediate row is constrained by the BAL hash alone — therefore any check built only on the state root is blind to intermediate rows."* That is the durable sentence and it belongs where it now sits.
- **Attacker clarification present**, `:137-143`: the re-pin count is a spectrum of *witness construction*, not of attacker difficulty — we mutate an existing valid fixture and must preserve its internal consistency, whereas an attacker constructs the block and every header field is free. Both variants are **equally exploitable and equally severe**.

That second paragraph matters more than it looks. Without it, a reader takes re-pin count as difficulty and concludes intermediate-row lies are the cheap attack and final-balance lies the expensive one — and both halves of that are wrong.

Also correct: one generator serving both variants via `--variant`, with the bai2 numbers re-verified **identical** after the refactor (`0xa73c76…`, `0xd8a995…`). A reader would otherwise reasonably wonder whether the shared script drifted between the two published results, and that is already ruled out.

Branch cut off current main `0c5e20bca` with `4979c03a1` ancestry confirmed IN_MAIN before cherry-picking — the right check, and the direction of `--is-ancestor` that answers definitively.

`+185/−40`, two files, diagnostic-only, no guest behaviour touched. **Labelling once checks clear.**

## What this pair of witnesses established

Two confirmed false accepts, one generator, and a structural result that neither witness alone would have produced: the guest accepts a block declaring a wrong withdrawal credit (`fork.py:379`, state root) **and** one declaring a wrong transaction-recipient credit (`fork.py:391`, BAL hash). Different spec checks catch the two lies; the guest performs neither.

The root is #10786 — the EOA route never records the recipient credit, and `bv_mtx_skip_list` skips that address as an *accommodation* for the absence. The repair is in progress.

PR #10789, opened 2026-07-28
fix: record MTx EOA recipient account writes

pirapira:
## Summary

Record the EOA recipient balance credit produced by the multi-transaction (MTx)
path at the N+1 account-builder boundary.  The change keeps the existing
account pre-state resolver and account-write producer as the single path; it
does not change the shipped output ABI.

The commits are intentionally separate:

1. `ee50ef808` extracts the shared account pre-state resolver.
2. `6d8751361` feeds the account builder at the MTx N+1 boundary.
3. `6601a57da` records MTx EOA recipient credits through the existing producer.
4. `10787985c` preserves the MTx settlement registers across that producer.

Commit 2 alone was not a harmless intermediate: its register clobber caused
an honest canonical block to be rejected (parent output+32 was 1 while the
commit-2 candidate output+32 was 0, with bytes 33 and 113 differing).  The
separate commit-4 reloads restore the canonical result.  The producer half is
therefore independently reviewable without hiding the false intermediate
verdict.

## K3 KATs

* Canonical fixture:
  `/tmp/fc668/in/00565_test_bal_withdrawal_and_value_transfer_same_address_fork_Amsterdam-blockchain_test__b0.input`
* Forged BAI2 input: `/var/tmp/evm-asm-bai-nplus1-k3/forged-final.input`
  (sha256 `05438109d09c7b11e9ddb4b67949d6bef5d10647caaad257631dc03690ed7ce4`).
* Parent ELF sha256:
  `7d9732ae4176e1529fd9545a0aedb3e4848bcd93ec33577f49b3ee6cf612c493`.
* Final candidate ELF sha256:
  `69cb5d40f24d95b2ae3517d59614f12a83c3929dc2ce4d8cf732b5087728fef9`.

Both parent and final candidate return verdict byte 1 on the canonical K3
input.  Both also return verdict byte 1 on the forged BAI2 input.  The latter
is deliberately not claimed as soundness evidence: the producer wiring does
not fix the remaining false accept.  Keep #10764 and #10786 open for that
work.

## Paired random gate

Parent and candidate were run with the same explicitly specified seed
`2026072801`, the same runner and the same 200-row manifest.  This seed was
specified for the paired test; it was not captured from an unseen run, so the
result is evidence for this paired draw, not a corpus-rate estimate.

Runner (borrowed from evm-asm2 because this checkout has no runner) was
`/home/yoichi-bkp/evm-asm2/scripts/spike/spike_run`, sha256
`c6ac09fa72c2103b6d3e390b779b022ef7f220e5a39bb5f667114a89d78e5868`.

Both legs scored 200/200 rows: 194 OK and 6 FR, with zero errors and zero
budget failures.  The per-label manifests were identical, all 200 output byte
strings were identical, and the transition table was:

| transition | count |
| --- | ---: |
| OK -> OK | 194 |
| FR -> FR | 6 |
| FR -> OK | 0 |
| OK -> FR | 0 |
| new FA | 0 |

This is verdict-neutrality for the paired 200-row draw, not a claim that the
full corpus has a 6/200 frontier.

## Diagnostic-only bridge

The BAI diagnostic export was used only as a bridge for the K3 investigation;
it is not the shipping artifact and no BAI diagnostic fields reach the
production output shape.  The diagnostic check compared the named emitted
regions and their hashes after the name-only source change; the production
candidate was then regenerated and linked separately.  The final diagnostic
coinbase field is the balance-builder row field, not an output-address-global
offset: `diag4.s` copies `bal_builder_balance_changes` to `OUTPUT+0x340`, rows
are 64 bytes, and row 1's post field is at `0x340 + 0x40 + 0x20 = 0x3a0`
(decimal 928).  Reading 32 bytes there from
`final-forged-diag.out` gives
`0000000000000000000000000000000000000000000000000000000000f61800`; a
command-side hexadecimal conversion (`int("00f61800", 16)`) gives
16,128,000.  The older 613800 reading came from the broken pre-clobber
diagnostic artifact and is not reused as production evidence.  The 26.3x
difference is expected because the old artifact passed a clobbered transaction
index into gas materialisation; this producer does not touch coinbase
accounting.

## Regeneration and layout

The final production artifact is
`/var/tmp/evm-asm-bai-nplus1-hook/gen-out/regionmap/stateless_guest.elf`.
Two successive relinks after regeneration converged at text `0x668fc` and
data `0x5370`; `GuestAddrs.bnf_le_a` remains `0xa3000ee0`; and
`scripts/check-region-map.sh` passed against the linked ELF and symbol table.

## Scope boundary

This PR records the producer half only.  The remaining question is whether
the receipt-side recipient-balance reject arm can be extended from contract
recipients to EOAs using the same shared resolver.  The false-accept controls
remain open in #10764 and #10786; this PR does not retire them.


pirapira:
Reviewed. The four-commit sequence is right and **the body says the two things that matter most**, which I checked by reading rather than by pattern-matching my own phrasing:

- `:15-18` — *"Commit 2 alone was not a harmless intermediate: its register clobber caused an honest canonical block to be rejected"*, with `parent output+32 = 1` against the candidate, and commit 4's reloads restoring it. That is what justifies the fourth commit existing, and folding it away would have hidden the evidence.
- `:36` and `:95-98` — this **does not fix the remaining false accept**; #10764 and #10786 stay open and this PR does not retire them. Without that sentence a reader seeing a soundness issue referenced plus a green 200-fixture gate concludes the bug is closed.

## Commit sequence

| commit | role |
|---|---|
| `6d8751361` | feed the account builder at the MTx N+1 boundary |
| `ee50ef808` | extract the account pre-state resolver — **proven inert**, parent vs candidate byte-identical on canonical and forged |
| `6601a57da` | record MTx EOA recipient credits |
| `10787985c` | preserve MTx EOA settlement registers — the `t0`/`t1` clobber fix |

## Evidence, and what each part can claim

- **Paired gate**, seed `2026072801`, 200/200 same labels, 194 OK + 6 FR on **both** legs, zero errors, all output bytes identical, zero transitions. Correctly described as a **specified** seed — so it supports "this change is inert on those 200 fixtures", not a corpus-level rate. The distinction is stated, which is what I asked for.
- **Diagnostic assertions** on the bridged artifact: recipient `c0f6` gains BAI1 = 5e9 where main had **no row at all**, and the withdrawal BAI2 becomes 15e9 rather than the 10e9 zero-baseline artifact. The second number is the one that proves the shared resolver resolves through the block map.
- **Layout**: `.text 0x668fc`, `.data 0x5370` unchanged, `bnf_le_a` still `0xa3000ee0`, `check-region-map` passing, fixpoint after two relinks.
- The borrowed runner (`/home/yoichi-bkp/evm-asm2/scripts/spike/spike_run`, sha `c6ac09fa72c2`) is disclosed, which saves the next person the four failed launches it cost here.

## Two things done well that are easy to miss

**The cross-module capacity guard.** `accountStateResolverCapacity = 38460` is emitted via `toString` and pinned by a theorem in `NonstorageEffectLog` against `CreateCodeEffectLog`'s constant — so the two numbers cannot drift silently across an import boundary the module could not cross. A comment would have been a number someone bumps; this fails at build time.

**Reusing `record_nonstorage_effect` rather than adding a container.** That routine is already a single producer with three synchronised materialisations (legacy effect log, AccountState journal, `tx_account_writes`), exercised by the contract-path callers. Writing a fourth container would have been the duplicated-state shape that produced the tx-collapse-versus-block-merge defect already on record.

## One correction incorporated

The coinbase field at `OUTPUT+928` reads `0x00f61800` = **16,128,000**, not 63,000 — a factor-of-256 conversion error, now fixed in the body with the conversion quoted alongside the `dd`. And the corrected value resolves the question better than the wrong one did: 613,800 on the clobbered build against 16,128,000 on the fixed build are genuinely different quantities, consistent with a garbage transaction index reaching `dispatcher_capture_exec_state_gas`. The producer never touched the coinbase.

`BEHIND` main and checks running. **Labelling when green.**

Issue #10784, opened 2026-07-28
Failed transaction's SELFDESTRUCT can be committed: account_state_delete_count survives into account_state_commit_pending on the status=0 post-preparation disjunct


pirapira:
Found by codex on a narrow routed question; independently verified against `origin/main` before filing.

## The defect

`BlockVerdictMtxRuntime.lean:615` selects between committing pending account-state and discarding it:

```
ld t2, 0(t2)                                    # t2 = bv_tx_status_arr[bv_mtx_i]
bnez t2, .Lbv_mtx_code_commit                   # success -> commit
la t0, runtime_tx_post_preparation_reached
ld t2, 0(t0)
bnez t2, .Lbv_mtx_code_commit                   # status 0 BUT flag set -> ALSO commit
```

So a **failed transaction** (`status == 0`) that reached the post-preparation coverage point takes the **commit** arm.

The two arms are mutually exclusive and only one clears the counters:

| arm | lines | behaviour |
|---|---|---|
| fall-through (status 0 **and** flag 0) | `:619-620`, then `:628` jumps to `.Lbv_mtx_code_commit_done` | `account_state_pending_count`, `account_state_created_count`, **`account_state_delete_count`** all zeroed — then **bypasses** the commit |
| `.Lbv_mtx_code_commit` | `:629-630` | `jal ra, account_state_commit_pending` — **no clear first** |

**Therefore a SELFDESTRUCT queued by a transaction that later failed can be committed**, because `account_state_delete_count` survives into the boundary.

### Why nothing upstream saves it

- `account_state_delete` queues the destroy and increments the count (`NoopHalt.lean`).
- The depth-0 rollback restores exec/storage/`account_writes` undo only; it does **not** touch `account_state_delete_count`.
- Child-frame rollback *does* restore it (`CallFrameReturn.lean` from the checkpoint captured in `CallFrameDescend.lean`), but **the status-0 MTx top-level route does not pass through `frame_return`.**

## The sharp part: this exact disjunct is already documented as wrong, one container over

`BlockVerdictMtxRuntime.lean:580-591` — a comment block sitting **35 lines above the branch** — says of this same `runtime_tx_post_preparation_reached` disjunct:

[...]

PR #10792, opened 2026-07-28
test(gate): GH #10784 MTx commit-arm regression gate (latent, mechanism-pinned)

pirapira:
## What

Regression gate for GH #10784. Measured direction: **LATENT** — the MTx account-state commit arm fires on every failing post-preparation transaction, but is unobservable in the post-state because of two load-bearing conditions:

1. Delete-queue insertion (`NoopHalt.lean:520-600`) is gated on created-in-same-tx (EIP-6780 semantics).
2. The depth-0 rollback removes everything a failed tx created, so the only delete targets the queue can name are already absent when the commit arm consumes it.

## Gate design

Per review guidance, the gate asserts the **instrumented-probe numbers, not the absence of a symptom**: `delete=1` at commit-arm entry on fx1 (plus `created=1`, arm marker = commit, post-preparation flag = 1) together with byte-exact post-state root agreement. A root-only check would pass forever — the root is no-op-identical *by* the mechanism under test. The gate passes today on current main; if either latency condition changes, the numbers move and it fires.

## Contents

- `scripts/fill/test_10784_mtx_commit_arm.py` — three spec-valid fixtures (same-tx-created / pre-existing / prior-tx-created SELFDESTRUCT target + top-level REVERT), filled with the in-repo EELS t8n.
- `scripts/gate-10784-mtx-commit-arm.py` — fills fixtures, runs pristine guest + verdict probe and an instrumented probe (patched emitted `.s` only; pristine ELFs untouched; recipe in docstring), asserts arm marker, counts at commit-arm entry, verdict/bv_fail, succ byte, and byte-exact roots.
- `docs/agents/regression-10784-mtx-commit-arm.md` — measurement table and the two load-bearing latency conditions.

Diagnostic-only: no guest behavior change. Verified end-to-end (`GATE PASSED`) on current main.

### DIFF
--- a/docs/agents/regression-10784-mtx-commit-arm.md
+++ b/docs/agents/regression-10784-mtx-commit-arm.md
@@ -1,0 +1,62 @@
+# GH #10784 regression gate: MTx commit arm vs failed-tx SELFDESTRUCT
+
+**Status: LATENT (measured).** The defect fires on every failing
+post-preparation transaction and is unobservable in the post-state. This gate
+pins the mechanism so an unrelated change cannot silently make it live.
+
+- Gate runner: `scripts/gate-10784-mtx-commit-arm.py` (run from repo root;
+  docstring has the full construction and the instrumented-probe recipe).
+- Fixtures (fill file): `scripts/fill/test_10784_mtx_commit_arm.py`.
+
+## The defect
+
+`BlockVerdictMtxRuntime.lean:615`: the per-tx account-state epilogue takes
+the COMMIT arm when the tx status is nonzero **or**
+`runtime_tx_post_preparation_reached` is set. Only the fall-through arm
+clears `account_state_pending/created/delete_count`, and the depth-0 rollback
+does not restore those counters. A SELFDESTRUCT queued by a transaction that
+later fails at top level is therefore consumed by the commit arm.
+
+## The measurement
+
+Corpus search found no ready-made fixture (eip6780_selfdestruct families are
+child-frame reverts or pre-existing targets; eip8246 initcode selfdestructs
+are top-level successes), so three fixtures were constructed with `fill`
+(in-repo EELS t8n) — all valid blocks:
+
+| fixture | shape | commit arm | post-prep flag | pending / created / delete at commit-arm entry | roots match spec |
+|---|---|---|---|---|---|
+| fx1 | same-tx-created A selfdestructs, tx REVERTs at top level | yes (17) | 1 | 7 / 1 / **1** | byte-for-byte |
+| fxA | pre-existing A selfdestructs, tx REVERTs at top level | yes (17) | 1 | 4 / 0 / 0 | byte-for-byte |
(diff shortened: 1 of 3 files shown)


pirapira:
Reviewed by reading the gate, and labelled. Built to spec, with one paragraph better than what was asked for.

## The design paragraph gets the trap right, in its own words

> *"A gate that only checked the state root would pass forever and tell you nothing: the root is no-op-identical BY the mechanism under test."*

That is the clean-gate-measures-irrelevance failure identified **and designed out from the start**, rather than discovered after the gate had been green for a month. It is the single most valuable sentence in the PR.

## Both latency conditions are load-bearing in the docstring

`:14-21`, stated as *"LATENCY CONDITIONS (load-bearing — this gate exists to pin them)"*:

1. delete-queue insertion gated on **created-in-same-transaction**
2. depth-0 rollback **removes everything a failed tx created**

Each tied to the evidence proving it, not merely asserted. Neither was written down anywhere before this — which is precisely how a latent defect becomes active during unrelated work: someone changes one for a good reason with no way to know what it was holding up.

## It asserts probe numbers, and the three shapes discriminate different things

| fixture | shape | expected |
|---|---|---|
| `fx1` | same-tx-created contract selfdestructs, tx reverts at top level | arm 17, flag 1, pending=7 **created=1 delete=1**, roots match |
| `fxA` | pre-existing contract selfdestructs, top-level revert | arm 17, flag 1, 4/0/0, roots match |
| `fxB` | contract deployed in tx1, selfdestructs in tx2 which reverts | arm 17, flag 1, 2/0/0, roots match |

`fx1` proves the destroy **is** queued for a failed transaction and consumed by the wrong arm. `fxA` and `fxB` prove the EIP-6780 created-in-same-transaction gating correctly suppresses queueing for pre-existing and prior-transaction-created targets. Two positives and two negatives from one instrument — a suite where everything moves together cannot localise, and this one localises.

## It does not fail today, deliberately

Ran end-to-end on current main: **GATE PASSED**, every number reproduced from the original measurement. `delete=1` on `fx1` is pinned as the **current correct-as-measured state**; the assertion is that it stays consistent with the roots matching. If either latency condition changes, the numbers move and the gate fires.

## Construction note worth keeping

`fill --output` must be **absolute** — `uv run --directory` relocates cwd, and a relative path silently lands inside `execution-specs`. The gate resolves the workdir to absolute before invoking. Silent-wrong-location is a failure class this project has paid for more than once.

## Context

This closes out #10784's direction question, where three hypotheses were put and all three refuted — false reject, second false accept, and route-unreachable. The answer was a fourth state nobody listed: **the defect fires on every failing post-preparation transaction and is unobservable**, because the only delete targets queued are same-transaction creations that the same transaction's rollback already removed.

No guest fix is behaviourally forced. The arm-selection disjunct should still be corrected for the same reason its storage twin was — but until then, this gate is what stands between a latent no-op and a wrong state root.

Issue #10791, opened 2026-07-28
The coinbase's nonstorage skip entry is unnecessary — its credit is already in the effect log on main, predating the producer work


pirapira:
Found by scoper while scoping #10786 part 1. Filing it separately because that part was abandoned, and this stands on its own.

## The coinbase's nonstorage skip entry is unnecessary, and predates all the producer work

`bv_mtx_skip_list` is passed to the two nonstorage comparators (`bal_all_accounts_nonstorage_consistent` / `_covers`, `BlockVerdictMtxTail.lean:294`/`:300` via the `a4`/`a5` setup at `:293`), and it contains the **coinbase** among its entries.

For the transaction *recipient*, that entry was a necessary accommodation: nothing recorded the recipient's credit, so without the skip `_covers` would fail on every plain transfer (#10786).

**The coinbase is not in that situation.** #10786's own pristine-main dump reads:

> Raw `exec_nonstorage_effect_log`, count = 3: sender gas debit, sender refund, **coinbase fee**. No recipient row.

So the coinbase fee credit **is** recorded in the effect log on unmodified `main` — independently of #10789's producer work, and predating it entirely. Its skip entry therefore excludes from comparison an address whose execution data already exists.

## Why this matters beyond one entry

The skip list is a **debt ledger**: each entry is an IOU against a missing producer. Measured — it appears in exactly three files (its allocation in `BlockVerdictDataSectionTail`, plus `BlockVerdictMtxRuntime` and `BlockVerdictMtxTail`) and in **none** of `BlockAccessListBuilder`, `BalSerializer` or `BalCanonicalSort`. It is purely comparison-side.

That matters for #10680: **a rebuilt BAL must contain every address the spec would include**, because the digest is over the whole canonical encoding. Per-address comparators can tolerate a skip list; a digest cannot. So every entry must reach zero, and an entry whose debt was already paid is the cheapest one to retire.

The spec has no counterpart. The only exclusion anywhere in spec BAL construction is slot-level (`block_access_lists.py:533`, storage reads that also appear in storage changes); `_build_from_builder` has no address filter. Compare #10684 (the storage-side 2935/4788 skip, same "no spec counterpart" shape) and #10661.

## Not done, and deliberately

[...]

Issue #10790, opened 2026-07-28
cleanup: delete the two consumerless routine-address pin theorems (regen tax with no reader) and the stale address prose beside them


pirapira:
Maintainer-approved. Routine-address literals in `Programs/*Spec*.lean` impose regen churn while having no consumers, so removing them is a pure reduction in fix-up cost.

## Measured cost, tonight

A change adding 68 bytes of `.text` moved `rlp_item_size` `0x80004a30`→`0x80004b38` and `rlp_item_span` `0x80004abc`→`0x80004bc4`. **The second codegen pass failed solely because two source pin theorems still held the old addresses**, requiring two manual source edits to satisfy assertions nothing reads. That is a fixed tax on every layout-moving change.

## Scope: delete these two. They have zero consumers.

| theorem | site | referenced by |
|---|---|---|
| `rlpItemSpanBase_eq : rlpItemSpanBase = (0x80004abc : Word) := by decide` | `RlpItemSpanSpec.lean:152` | **nothing** — appears only at its own definition |
| `rlpItemSizeBase_eq : rlpItemSizeBase = (0x80004a30 : Word) := by decide` | `RlpSpliceHelperSpec.lean:697` | **nothing** — same |

Verified with `git grep` across `EvmAsm/**`: each name occurs exactly once, at its definition. They assert that a symbolic base equals a hardcoded literal, and that assertion is consumed by no proof. Their only effect is to fail when addresses move.

The symbolic bases themselves (`rlpItemSpanBase`, `rlpItemSizeBase`) **are** consumed — `CodeReq.ofProg`, `cpsTripleWithin`, the spec triples — and resolve through `GuestAddrs`, so they are unaffected by drift and must stay.

## Also worth cleaning, lower priority: address prose that decays silently

Six literals are not theorems but docstring layout tables:

- `RlpItemSpanSpec.lean:162-163` — *"contiguous: size `0x80004a30..0x80004abc`, span `0x80004abc..0x80004b90`"*
- `AccountAccessorTopSpec.lean:48-51` — *"`rlp_walk_init 0x800049fc` (53 instrs) … ends at `0x80004ad0`"* and three similar rows

These do not break the build. They go **stale silently**, which is the worse failure mode — the project already has a documented pattern of quantitative prose decaying unnoticed, and every one of these six is wrong the moment `.text` moves. Either delete the addresses and keep the instruction counts, or state the ordering symbolically.

## Keep

[...]

PR #10798, opened 2026-07-28
diag: add rebuilt BAL shadow serializer

pirapira:
## Non-gating rebuilt BAL shadow serializer

Adds an emitted, non-gating shadow serialization of the rebuilt block access
list (BAL). It records the rebuilt length, supplied length, and per-list
emitted-row counts in output slots above the production ABI. The shadow runs
only after BAL decoding and gas validation; nothing branches on its result.

Status values are `0` match, `1` mismatch, `2` serializer error, and `3` not
reached. The `bv_bal_shadow_ready` precondition plus the `3` sentinel avoid
interpreting an unparsed BAL as a comparison result.

### Measured artifact and controls

Diagnostic ELF: `3eae29eebba12769e023f138491fecff237713786d3eb3a17a170c573de45ddb`  
Sections: `.text=0x66b68`, `.data=0x5370`, `.bss=0x1b85ca40`  
Runner: `/home/yoichi-bkp/evm-asm2/scripts/spike/spike_run` with
`SPIKE_OUTPUT_LEN=192`.

| fixture | status | rebuilt/supplied | emitted S/R/B/N/C | global storage reads |
| 00537 simple fork | 1 | 300/391 | 2/0/4/0/0 | 16 |
| 00548 EIP-4788 query | 1 | 112/236 | 3/0/0/0/0 | 0 |
| 00565 withdrawal + value transfer | 1 | 262/350 | 2/0/3/0/0 | 16 |
| 00569 withdrawal to coinbase | 1 | 264/345 | 2/0/3/0/0 | 16 |

All four canonical controls accept. The mismatch is expected: this instrument
reports a repair worklist; it does not validate the supplied BAL.

The initial matcher hypothesis was retired by a follow-up population probe:
on 00537, the first read row reverses to the canonical BAL address, but that
read-only address is absent from the populated builder-account table. Therefore
the serializer never receives an account against which it could match the
read. The root is upstream account interning for storage reads, not a
serializer byte-order fallback. 00548 remains a distinct modeled-system read
producer gap: its global read count is zero.

### Checks

- Narrow cache-disabled build passed for the changed serializer, tail, RLP
  helpers, and `GuestImage` modules.
- The shadow has no raw capture buffer; it exports only self-describing
  counters and lengths.
- The linked layout was regenerated coherently before the artifact was
  captured.
- Paired random-200 Spike gate, captured seed `2126109947`, same runner and
  fixture population on both legs:
  parent `fcc8467cef6478e5779b983286d2895a89065cd969e729f492126c810bb873a9`
  versus shadow candidate
  `3eae29eebba12769e023f138491fecff237713786d3eb3a17a170c573de45ddb`.
  Both scored all 200 manifest rows (`FA=0`, `FR=7`, `ERROR=0`, `BUDGET=0`).
  The self-checking comparator joined 198 shared-input groups (two legitimate
  many-to-one groups), found zero candidate-only faults, identical seven-label
  FR sets, zero verdict transitions, and byte-identical output on every joined
  case.  The parent artifact was rebuilt directly from PR base `4dfe8c652`;
  the candidate is the measured shadow artifact above.

Follow-up repairs are intentionally separate commits/PRs so behavioral effects
remain attributable.


pirapira:
Reviewed. The instrument in this PR produced tonight's entire remaining work list, and it is worth landing on that basis alone. Two requirements before it does.

## What it bought, for the record

Sequence, each step costing minutes and each partitioning what the previous left:

1. **shadow status** → the digest mismatches on every accepted fixture, and *not* because of an input echo
2. **rebuilt vs supplied length** → every mismatch is **rebuilt-shorter**, so the cause is **rows missing**, not values wrong. That eliminated the resolver and the ordering in one measurement.
3. **five per-list counters** → `storage_reads` **emitted = 0** on every fixture against 16–17 supplied. Named the absent list.
4. **global `storage_reads_count`** → split that single symptom into **two separable defects**: rows exist but nothing emits them (00537/565/569), versus rows never produced at all (00548).

None of that was observable before something consumed builder rows. With nothing reading them, a missing list had no consequence and the only detectors were source-text guards.

## Requirement 1 — run the full paired 200, and hunt faults rather than verdict changes

The random 20 is too thin for the risk this change actually carries, and its own history says so: **the first version faulted on fixture 17006**, and only broader sampling found it. A four-control check would have shipped it.

The failure mode here is a **fault** (`rc=3`, load access fault on unmaterialised BAL cells), not a verdict change — and ERROR is its own category alongside FA and FR, not a variety of reject. So the gate to run is 200 paired at a captured seed, and the number to report is **zero candidate-only faults**, alongside verdicts unchanged.

One fault class was found at ~20 fixtures; 200 is what establishes there are no others.

## Requirement 2 — say "non-gating", not "diagnostic-only"

The body describes this as diagnostic. It is **non-gating**, which is different and the distinction matters to a reviewer: it modifies real guest source — `BalSerializer`, `BlockVerdictFunction`, `BlockVerdictReceiptsTail`, `BlockVerdictDataSection`, `StatelessGuestEpilogue` — adds a call on a path every block reaches, and changes emitted bytes and layout.

What makes it safe is that **nothing branches on its result**, plus the structural precondition (`bv_bal_shadow_ready`, set only after `bal_gas_valid` succeeds) and the explicit `status=3` not-reached sentinel so a skipped fixture cannot masquerade as a match. Say that, rather than implying no guest code moved.

## Correctly stated in the body already

- the digest **mismatches by design** at this stage, with the per-list numbers as the recorded work list
- the storage-read work list split into the dual-form matcher and the separate modelled-system producer

That first point is the one that matters most. A reader seeing a digest wired in and a soundness issue referenced would otherwise conclude the BAL is now validated. It is not — this reads and reports, and gates nothing.

## Incidental, and it is evidence for #10790

This PR again remaps `RlpItemSpanSpec` and `RlpSpliceHelperSpec`. That is the **third time tonight** those two consumerless pin theorems have cost a manual edit on an unrelated change — the measured cost that justifies the approved cleanup.

pirapira:
Both requirements met — labelled.

**Gate:** captured seed `2126109947`, parent `4dfe8c652` (ELF `fcc8467c…`) against shadow `3eae29ee…`. Both legs 200/200 scored, **FA=0, FR=7, ERROR=0, BUDGET=0**. Comparator: 198 joined shared-input groups, **zero candidate-only faults**, identical 7-label FR set, zero verdict transitions, byte-identical joined outputs.

Three things about that result worth naming:

- **Zero candidate-only faults is the number that mattered**, not the verdict equality. This change's failure mode is a fault (`rc=3`, load access fault on unmaterialised BAL cells), and ERROR is its own category alongside FA and FR. The first version of this wiring *did* fault, on fixture 17006, and only broader sampling caught it — a four-control check would have shipped it. 200 is what establishes there are no other instances of that class.
- **The parent was rebuilt at `4dfe8c652` rather than reused.** The older `7d9732ae` artifact was to hand and would have been wrong: it predates the forward-merge, so the comparison would have measured this change *plus* every intervening main commit and reported the whole difference as this change's effect. Refusing to substitute without provenance was correct.
- **The seed is genuinely captured**, not chosen. An earlier leg used a date-shaped value, which supports a claim about *those* 200 fixtures rather than a random sample; this one supports the stronger reading.

**FR=7 against the earlier legs' 6** is a different sample under a different seed, not a regression — both legs report 7 with an identical label set and zero transitions.

## What this PR is, stated plainly

Non-gating, not diagnostic-only. It modifies real guest source and changes emitted bytes and layout. What makes it safe is that **nothing branches on its result**, plus the `bv_bal_shadow_ready` precondition (set only after `bal_gas_valid` succeeds, so the serializer never runs on unmaterialised cells) and the explicit `status=3` not-reached sentinel so a skipped fixture cannot masquerade as a match.

**The digest mismatches on every accepted fixture by design at this stage.** Nothing is validated by this landing; #10764, #10786 and #10791 all remain open.

## Why it was worth landing before any repair

It converted the remaining BAL work from reading into measuring, in four steps of minutes each: status bit → *mismatch, not an echo*; rebuilt vs supplied length → **rows missing, not values wrong**, which eliminated the resolver and the ordering outright; five per-list counters → named `storage_reads`; one global count → split that into **two separable defects**. None of it was observable while nothing consumed builder rows.

Two agents are now stacking branches on this instrument to measure their own work, so landing it unblocks both tracks.

pirapira:
Dropped from this round — **`check-file-size.sh` fails**, and it is this PR's diff that crosses the cap.

```
```

Measured line counts: `origin/main` **1499**, this branch **1503**. The diff adds exactly 4 lines to that file — one emitted-asm line plus three comment lines explaining it:

```
+  -- The decoded BAL slice now satisfies the same precondition consumed by the
+  -- downstream granular comparators.  The terminal shadow observer uses this
+  -- reach marker rather than guessing from a pointer value.
+  "  li t0, 1; la t2, bv_bal_shadow_ready; sd t0, 0(t2)\n" ++
```

**The structural fact worth knowing: main's copy of this file is 1499/1500 — one line under the cap.** Any PR that adds a single line to it fails this gate. So this is not really a problem with your 4 lines; the file is full.

I could have dropped the three comment lines to land at exactly 1500 and gone green. I did not, for two reasons: those comments explain a non-obvious mechanism (why a reach marker is used rather than inferring from a pointer), and deleting them would leave the file still at the cap so the next PR touching it fails again. Shaving documentation to fit a guardrail trades a real explanation for a temporary pass.

The gate's own guidance for `EvmAsm/Codegen/Programs/**` is to extract a cluster of defs into a sibling submodule and import it back. **Say the word and I will do that split and verify it is byte-neutral** — emission order comes from the concatenation expressions, not file layout, so the regenerated TSV and `textSizeBytes` should come out identical, which makes the refactor checkable.

Everything else on the merged branch is green: `lake build` exit 0, 20 of the other gates pass, 0 other real failures. Regen converged first iteration (TSV 3210 data rows, pins `textSizeBytes` `0x066b68` / `.bss` `0x1b85ca40` already correct). The merge of main `88575e7bd` — which carries #10795's vendored Sail regen — was clean.

PR #10800, opened 2026-07-28
refactor(codegen): parameterise HashBridge over GuestLayout (GH #10753 first conversion)

pirapira:
## What

First module conversion of the guest-layout parameterisation (GH #10753 plan): `HashBridge` (fan-out 360, `BloomAddValue`'s only transitive GuestAddrs blocker).

- **Leaf** `Programs/HashBridgeProg.lean` (new): the whole former module, with only the three `*_prog` definitions parameterised as `*_prog_of (L : GuestLayout)`. No `GuestAddrs` import. The `Function` / `_eq_prog` / `#guard` emission blocks stay in the leaf applied at `GuestLayout.zero` — `emitProgramR` keeps `la`/`jal` symbolic, so they are layout-independent.
- **Bridge** `Programs/HashBridge.lean` (module name unchanged): imports the leaf and defines `zkvmSha256_prog := zkvmSha256_prog_of guestLayout` (likewise keccak/segments) — original names and `Program` types re-exposed, so all 360 dependents and `lean_render` keep working untouched (Lean imports are transitive).
- `GuestLayout.lean`: flat structure, 8 fields (HashBridge's symbols), grown incrementally per conversion; `GuestLayoutInstance.lean` binds it from the generated `GuestAddrs` table (sole importer on this path). Hand-written per the sanctioned fallback; generation from `_collect_guest_addr_syms` is a follow-up.
- `asm_to_program.py`: `gen_lean_layout` + layout-mode drift detection (leaf block pinned verbatim in `<Name>Prog.lean`, bridge def pinned in the manifest-path file). Legacy path byte-identical in behaviour: `check-all` reports 0 drifts before and after.

## Acceptance (per the plan)

| Gate | Result |
|---|---|
| flip `sha256_w_state` in GuestAddrs | only GuestAddrs → GuestLayoutInstance → bridge rebuild; **leaf olean sha256 unchanged** (`7a6f3678…`); restored + `check-guest-addrs` clean |
| `check-file` HashBridge (symbolic + concrete + drift) | clean, 3 defs; concrete `fn#c` renders == guest-linked bytes |
| `check-all` (source drift, all modules) | 0 drifts (matches pre-patch script) |
| `lake build codegen` | clean, 4274 jobs, no consumer edits |
| end-to-end | emitted `stateless_guest` from converted tree runs the GH #10784 fx1 fixture: succ=1, output byte-identical to expected |

Note: the local checkout's `check-unimported.sh` flags two orphan files, both pre-existing foreign untracked files on this shared checkout, not from this PR.

No guest behavior change. BloomAddValue report: it imports `GuestAddrs` directly (line 15), so it was never a leaf by direct import; after this conversion its only transitive program blocker is gone — its future leaf will import `HashBridgeProg`, and its conversion adds 2 fields to `GuestLayout`.


pirapira:
Reviewed. **Approving the shape.** This is the separation that was asked for, the module split is exactly right, and the one thing I set out to attack — whether moving the guards to `.zero` quietly weakens them — turns out to be **bounded from both sides by pre-existing gates**. Detail below, because the reason it is safe is more interesting than the fact.

## The split is right

- `GuestLayout.lean` — abstract, **no `GuestAddrs` import**, 8 fields for exactly the symbols `HashBridgeProg` references.
- `GuestLayoutInstance.lean` — the **only** module on this path importing `GuestAddrs`, every field bound *from* the table, so the instance cannot drift from it and a missing field is a compile error at the leaf.
- `HashBridgeProg.lean` — the whole former module, `_prog_of (L : GuestLayout)`.
- `HashBridge.lean` — three lines applying `guestLayout`, original names, original types.

Incremental field growth with the flat structure is correct at this scale, and recording in the docstring that the mega-flat end state fails elaboration is the right place for that measurement.

## The `.zero` substitution is enforced, not assumed — and this is the load-bearing point

The guards moved from the concrete `_prog` to `_prog_of .zero`. That is the one place this refactor *could* have silently relaxed a check, and it does not, because **two gates that already existed bound it from opposite directions**:

1. **`emitted_reloc_count(rendered[fn]) == 0` → error.** If any `la`/`jal` immediate were baked from a layout rather than kept symbolic, this fires. So "emission is layout-independent" is *checked*, not asserted.
2. **`assemble_cmp(asm, rendered[fn], …)` against the hand-written fixture.** This closes the gap the reloc count cannot see: a layout-dependent value that is *not* a `la`/`jal` — say an `li` of an absolute address — would render as `0` under `.zero`, diverge from the fixture bytes, and reject.

So a `.zero` render that assembles to the fixture *and* still carries its relocations cannot be hiding a layout dependence. Nothing was lost.

## Re-exposing the original name is load-bearing for the gate, not just for consumers

The docstrings frame `zkvmSha256_prog := zkvmSha256_prog_of guestLayout` as keeping consumers compiling. It does more than that, and it is worth saying so explicitly so nobody "tidies" it later:

The concrete-render gate builds its key as `emitProgram {lean_camel(entry)}_prog` and byte-compares against the guest-linked fixture. Because the bridge keeps that **exact name at that exact type**, the pre-existing `#c` gate resolves it to `_of guestLayout` and still ties the concrete immediates to the real link — **unchanged, and pointed at a real layout-applied program**. Rename the bridge def and the gate either fails loudly on an unknown identifier or, worse, gets "fixed" to the `.zero` form and goes quiet over nothing. Please add a line to the bridge docstring saying the name is required by the `#c` gate.

## One observation that is NOT a defect in this PR

`{func_name}_eq_prog` is a **tautology**: `zkvmSha256Function` is *defined* as `"zkvm_sha256:\n" ++ emitProgramR … ` and the theorem restates that definition, so `rfl` closes it by unfolding. It proves nothing about drift.

**This predates the PR** — `HashBridge.lean:158-159` on `main` has the identical shape, so the branch faithfully preserves an existing property. I am flagging it only so the "Kernel-checked drift guard" docstring is not taken at face value by whoever converts module two: the actual drift guard is `check_file`'s **verbatim block match** plus `assemble_cmp`, both of which this PR correctly extends to the leaf/bridge pair with *separate* problem messages. Not blocking, and not yours to fix here.

## Housekeeping

- The 2 `check-unimported` orphans are foreign untracked files on the shared checkout (`Bn254FieldMulModSAsmGuest`, `LagunaProbe`) — agreed, not yours. **Do not clean them**; ownership on that checkout is not determinable.
- `BloomAddValue` answer noted: it imports `GuestAddrs` directly at line 15, so it was never a leaf by direct import; after this lands its transitive program blocker is gone and its conversion adds 2 fields. That is the right next module.
- Generating `GuestLayoutInstance.lean` from the same fixture scan that emits `GuestAddrs.lean` is the correct follow-up and the docstring already commits to it — file it as an issue so it does not live only in a comment.

pirapira:
Dropped from this round — **it breaks step 3 of the documented regen procedure.**

```
$ python3 scripts/guest_image_coverage.py --emit-lean
could not parse Function def for zkvmKeccak256Function in EvmAsm/Codegen/Programs/HashBridge.lean
exit=1
```

It exits **1 and writes nothing**, so `EvmAsm/Codegen/Proofs/GuestImageEntries.lean` cannot be regenerated at all on this branch.

Isolated to this PR, measured on each ref:

| ref | `guest_image_coverage.py --emit-lean` |
|---|---|
| `origin/main` | exit 0 |
| #10798 `feat/bal-digest-shadow` | exit 0 |
| **this branch** | **exit 1** |

The cause is the parameterisation itself: the script's parser expects the previous `zkvmKeccak256Function` def shape, and the `GuestLayout` parameter changes it. Since this is the first conversion of #10753, **every subsequent conversion will hit the same wall**, so it is worth fixing in the generator now rather than per-PR.

I did not attempt the fix. Teaching the parser a new def shape is a change to a generator that produces a kernel-checked file, and a parser that silently mis-parses would produce a *stale* `GuestImageEntries` that still compiles — the failure mode is worse than the current hard error. That is past mechanical conflict repair.

One reassuring measurement: **your refactor appears emission-neutral.** I built the two-PR batch (#10798 + #10800) and then #10798 alone, and both produce identical layout — `textSizeBytes` `0x066b68`, `.bss` `0x1b85ca40`, TSV 3210 data rows. So the parameterisation does not move bytes; only the generator's ability to read it is affected.

Also note main is now `88575e7bd`, which carries #10795's vendored Sail regen (113 -> 116 generated files), so a sync will pull that diff in.

pirapira:
Unblocked and batched. #10803 makes `guest_image_coverage.py` layout-aware, and on the merged tree the generator now exits 0 and writes 364 entries with your parameterised `zkvmKeccak256Function` in place. Removing `complicated-merge`; landing via #10805.

Also confirming the reassuring measurement from earlier holds on this batch: layout is **byte-identical to main `88575e7bd`** — `textSizeBytes` `0x066918`, `.data` `0x5370`, `.bss` `0x1b85c9c0`, and the TSV, `GuestAddrs.lean` and `GuestImageEntries.lean` all regenerate unchanged. So the parameterisation moves no emitted bytes, which is what a conversion refactor should do.

PR #10801, opened 2026-07-28
codegen: intern read-only storage accounts in BAL

pirapira:
## Read-only storage accounts reach the BAL builder

`execution-specs`' `add_storage_read` ensures an account before it appends a
storage read.  The guest recorded the read in the execution log but never fed
the corresponding account into the BAL builder, leaving read-only accounts
unmaterialised.

This patch repairs that record-time gap.  `storage_read_record` now converts
the live execution-log address to canonical BE20 and calls the existing,
idempotent `bal_builder_ensure_account` after finding or appending the read.
Both live callers supply `x20 = env.ADDRESS` as a little-endian stack word:
the top-level runtime stager reverses the RLP address before storing it, and
the nested call-frame setup copies that same form.  The conversion is therefore
uniform across depths; representation follows the producer, not call depth.

This is deliberately record-time rather than promotion-time: it mirrors the
spec's ensure-before-read ordering and keeps reverted reads independent of the
transaction-promotion boundary.

## Focused controls

The linked candidate is SHA
`4a6e3d37c259784587415fd073faf004b90e48e67d8973a1d64f7050850bd684`
with `.text=0x66bc8`, `.data=0x5370`, and `.bss=0x1b85ca40`.
All four canonical controls accept under Spike.  The diagnostic values are
`rebuilt/supplied` and `S/R/B/N/C`:

| Fixture | rebuilt/supplied | S/R/B/N/C |
| --- | ---: | --- |
| 00537 | 424/391 | 2/16/4/0/0 |
| 00565 | 386/350 | 2/16/3/0/0 |
| 00569 | 388/345 | 2/16/3/0/0 |
| 00548 | 220/236 | 3/0/0/0/0 |

The intended repair is directly measured: the three read-bearing controls move
from zero to 16 read accounts, while 00548 has zero global storage reads and is
unchanged in that component.  The key form is a property of the producing call
site rather than frame depth: both live callers pass `x20` in stack-word order
at every depth.

## Bounded residuals

This is a partial, strictly improving producer repair.  The digest comparator
is not yet wired into verdict control flow, so these diagnostic mismatches do
not alter the guest's accept/reject verdict in this slice.

- The rebuilt output overshoots supplied by 33, 36, and 43 bytes on 00537,
  00565, and 00569, while 00548 undershoots by 16 bytes.  Those are named
  residuals, not explained by this patch.
- The observed read count is 16 while the supplied count is 17 on the
  read-bearing controls; global reads are also 16.  This is a separate
  record-side miss, recorded against fixtures 00537, 00565, and 00569.
- The next measurement is a counter for builder account entries emitted versus
  accounts supplied.  An overshoot confirms duplicate account rows and points
  to the lookup key in `bal_builder_ensure_account`; a match refutes that
  hypothesis and moves the investigation inside the entries.
- 00548 remains a modeled-system/no-storage-read case and therefore provides
  the neutral control, not evidence that the modeled-system residual is fixed.

## Checks

- `scripts/check-asm-to-program.sh` — clean (390 converted definitions).
- `scripts/check-region-map.sh` — passes against the linked ELF.
- Regeneration was serialized and converged: the third `GuestAddrs`,
  `GuestImageEntries`, and symbol-table pass was unchanged before the layout
  pin was committed.
- Behavioural tier: **random-200 plus 4 pinned controls** (204 rows), rather
  than a full corpus A/B.  The candidate drew seed `8661204007568258456`
  automatically; the immediate parent repeated that seed.  The frozen Spike
  runner SHA was `c6ac09fa72c2103b6d3e390b779b022ef7f220e5a39bb5f667114a89d78e5868`.
  Both legs scored `FA=0, FR=10, FAULT=0, OK=194`; transitions were exactly
  10 `FR -> FR` and 194 `OK -> OK`.  The four pinned controls 00537, 00548,
  00565, and 00569 were all `OK -> OK`.

The raw extended diagnostic output differs on 195 sampled rows, including all
four pins, because this change intentionally changes builder/shadow fields.
Those fields are not verdict inputs in this slice; the gate above reports the
verdict-level effect separately rather than claiming byte equivalence.
Sampled `FA=0` is therefore not a claim of 26,104-row FA coverage.  It is
sufficient here only together with the independently enumerated fact that the
builder account table has no gating verdict consumer; if a later change wires a
consumer to that table, this sampling rationale expires and must be replaced.


pirapira:
Blocked by the **same root cause as #10798**, not by anything in this PR's own logic.

```
```

That is what CI's `source-checks` is failing on (it fails in 15s, well before the build — the build itself **passed** at 16m26s). Measured after merging main `1303812cc` forward: main's copy of that file is **1499**, this PR adds 4 lines to it, giving 1503.

**So the 1499/1500 file in main now blocks two labelled PRs** — this one and #10798. Any PR adding a single line to it fails.

Good news: **codex's #10806 ("refactor: split block verdict terminal tail", adding `BlockVerdictFunctionTail.lean`) is exactly the fix**, and it should unblock both. I am deliberately not doing that split myself — it was assigned to codex, who authored #10798 and knows the file.

Everything else here is clean, measured on the merged tree:
- merge of main was **clean**, zero conflicts
- `lake build` exit 0
- 20 of the other gates pass; `check-file-size` is the **only** real failure
- the usual 4 (`check-axioms`, `check-build-parallel`, `check-guarded-handler-bytes`, `check-opcode-tables`) could not run locally because of this workspace's lake artifact-cache defect, unrelated to this diff

I have not pushed the main-merge, since it would spend a CI cycle on a run that fails at the same gate. Once #10806 lands I will re-sync, re-verify by **running** `check-file-size` rather than trusting the label, and merge.

Issue #10799, opened 2026-07-28
bal: the coinbase skip in bv_mtx_skip_list is unnecessary independently of #10789 — a second accommodation predating the producer work


pirapira:
Split out of #10786 at coord's request, because it stands on its own as a divergence and does not depend on whether the recipient narrowing (#10786 part one) happens.

## The claim

`bv_mtx_skip_list` contains the **coinbase** (`skip[2N]` = `fee_recipient`, `BlockVerdictMtxTail.lean:30-36`), and that entry is unnecessary for the nonstorage dimension — because the coinbase credit **is already recorded** in the effect log on pristine main.

## Evidence, from #10786's own pristine dump

#10786 measured the raw `exec_nonstorage_effect_log` on unmodified main, fixture `bal_withdrawal_and_value_transfer_same_address`:

> **Raw `exec_nonstorage_effect_log`, count = 3:** sender gas debit, sender refund, **coinbase fee**. No recipient row.

So the coinbase fee credit has a row. The recipient does not. The recipient's skip entry is an accommodation for a genuinely absent producer (#10786); **the coinbase's is not** — its producer exists and has existed all along.

## Why this is separate from #10786

- #10786's recipient skip exists **because** the EOA route never stages the credit. Remove it without the producer and every plain transfer false-rejects.
- The coinbase skip has no such justification in the nonstorage dimension. Its effect-log row is present, so `_covers` would find it.

## What is NOT claimed

- **Not** claimed for the storage or tuple dimensions. Codex established that `BalAllAccountsStorage` skips both modeled-system accounts unconditionally, and the storage side must stay untouched.
- **No witness exists.** The recipient narrowing has the bai2 witness from `scripts/witness-10764-withdrawal-fa.py`. A coinbase narrowing would need a coinbase witness, and there isn't one. Per the one-accommodation-per-change rule, this should not be actioned until such a witness flips.
- I have **not** verified whether some other consumer of the list depends on the coinbase entry for a reason unrelated to the effect log.

## Structural note that constrains any fix here

[...]

PR #10807, opened 2026-07-28
fix: drop the mutated-state tier from account_resolve_pre_state (spec alignment)

pirapira:
## What

`account_resolve_pre_state` had a **third resolution tier the spec does not have**, and it read
**live mutated state**. Removing it is a deletion; the routine gets shorter.

`_get_pre_tx_account` (execution-specs `block_access_lists.py:583-598`) has exactly two tiers:

```python
if address in pre_tx_accounts:
    return pre_tx_accounts[address]
return pre_state.get_account_optional(address)
```

`pre_tx_accounts` is the block-cumulative map; `pre_state` is the **immutable pre-block state**.
The guest consulted the durable `AccountState` overlay in between. `update_builder_from_tx` runs at
the transaction boundary, by which point the sender's gas debit and nonce increment have already
been applied to that overlay — so for an account with **no block-map row** (its first touch in the
block) the overlay returned the **POST** value as the "pre" value, the change-compare found
equality, and the BAL row was silently dropped.

From the second transaction on, the block map hits and supplies a genuine pre value. That is why
the deficit was exactly **one row per distinct sender**, and why a sender's *later* rows appended
correctly while its first did not.

A block-map miss now falls through to the authenticated parent witness — what the spec says, and
correct on its own terms: if no earlier transaction in this block touched the account, its
pre-transaction state *is* the parent state.

## Measured effect

Six EIP-7928 fixtures, by relpath under
`blockchain_tests/for_amsterdam/amsterdam/eip7928_block_level_access_lists/`:

| fixture | nonce deficit | balance deficit |
|---|---|---|
| `block_access_lists/bal_multiple_storage_writes_same_slot.json` | 1 → **0** | 2 → 1 |
| `block_access_lists/bal_cross_tx_storage_write.json` (both cases) | 1 → **0** | 1 → **0** |
| `block_access_lists_eip2935/bal_2935_simple.json` | 2 → **0** | 2 → **0** |
| `…eip4895/bal_withdrawal_and_value_transfer_same_address.json` | 1 → **0** | 1 → **0** |
| `…eip4895/bal_withdrawal_to_coinbase.json` | 1 → **0** | 1 → 1 |

**Nonce deficit is zero on all six.** Balance deficit is zero on four.

## Regression A/B — 80 fixtures, legs differing only in this deletion

Two pinned ELFs, identical but for this change (`b8fd47f5…` vs `09e4aec0…`), spike backend:

| | pre | post |
|---|---|---|
| succ bit == 1 | 75 | 75 |
| succ flipped 1→0 (new false rejects) | — | **0** |
| succ flipped 0→1 (new accepts) | — | **0** |
| fixtures emitting **more** balance rows | — | 69 |
| fixtures emitting **fewer** balance rows | — | **0** |
| fixtures emitting **more** nonce rows | — | 73 |
| fixtures emitting **fewer** nonce rows | — | **0** |
| `storage_changes` moved | — | **0** |
| `storage_reads` moved | — | **0** |

Strictly monotone: more rows or the same, never fewer, with no verdict movement and no collateral
change to the storage legs.

## What this does NOT fix

- **The shadow digest still matches on 0 of 80.** This change is necessary, not sufficient; other
  legs in this lineage are incomplete (`storage_reads` emits 0 here — the sibling reads repair is
  not in this branch's ancestry).
- `bal_multiple_storage_writes_same_slot` still loses the **coinbase** row at `block_access_index 3`
  on a three-tx block. Untouched by this change, so it is a separate mechanism.
- `bal_withdrawal_to_coinbase` still loses the **sender** row — and there the resolver's operand is
  now correct while the staged **post** equals the pre, so that is a staging defect on the *other*
  operand. Also a separate mechanism.
- **A remaining spec divergence, deliberately not bundled:** the spec's lookup is *whole-account*
  (`pre_tx_accounts[address]` is a complete `Account`), while the guest resolves **per component**
  through a valid mask. A block-map row carrying balance but not nonce therefore still resolves
  nonce differently from the spec. This change makes that fall-through go to the parent witness
  instead of to mutated state — strictly closer to spec — but does not close it.

## Diagnostics in this branch

The two earlier commits add `.bss` producer-side diagnostic cells (`bald_*`) that localised this
defect; they are diagnostic-only and change no behaviour. `.data` is unchanged at `0x5370`
throughout (all-zero groups are promoted to `.bss` by `Layout.moveZeroDataLines`), so no
address-pinned `.data` symbol moved. Full findings and the retracted hypotheses are on #10799.

## Gates

`lake build` completed successfully (4774 jobs), which is what verifies the regenerated guest-PC
pins rather than an intention; `check-region-map` matches the linked ELF (`textSizeBytes = 66e40`,
`dataSizeBytes = 5370` unchanged, `bssSizeBytes = 1b85cac0`); `check-forbidden-tactics` clean;
layout regen at fixpoint (pass 2).


pirapira:
### `source-checks` is red for a reason inherited from this PR's base, and the fix has moved

`source-checks` fails on the file-size guardrail:

```
```

Provenance, measured rather than assumed:

| ref | lines |
|---|---|
| `origin/main` | 1499 |
| `origin/feat/bal-digest-shadow` (this PR's base) | **1503** |
| this branch | 1503 |

`git log` over this PR's three commits for that path is **empty** — no commit here touches the file.
The base branch's diagnostic commit took it over the cap, and this PR inherits the red because it
was based on `feat/bal-digest-shadow` (which is where the `bv_bal_shadow_emit_*` counters the
diagnosis needed live).

`check-no-hardcoded-guest-pc.sh` passes, and the `build` check is present and running rather than
absent, so the rest of the rollup is meaningful.

**The producer fix does not depend on this base at all** — it is a deletion inside
`account_resolve_pre_state`, and `origin/main` carries the identical durable tier. So the spec
alignment has been moved to a **main-based branch** to be landable without touching this base's
size debt or splitting a 1503-line file. This PR remains as the diagnostic-cell branch.

Not done here, deliberately: the guardrail is not disabled or excepted (the script states per-file
exception markers are unsupported), `BlockVerdictFunction.lean` is not split, and the base branch is
not edited.

pirapira:
## Amendment: the A/B table in this PR is true and misleads by omission

**This is not a withdrawal of the fix.** It is a correction to how I summarised it.

The table above reports "more rows or the same, never fewer", zero new false rejects, zero new accepts,
storage legs immobile. **All of that is accurate.** But I presented monotone-up row counts as
convergence, and **more rows is only good if the spec wants them.** I did not check the byte length
against the supplied BAL. Doing so now, on the same 80-fixture A/B artifacts:

| | value |
|---|---|
| comparable fixtures (`supplied_len` nonzero both legs) | 76 |
| `\|rebuilt − supplied\|` moved **closer** | 72 |
| unchanged | 3 |
| moved **away** | 1 |
| `rebuilt > supplied` **before** the fix | **0** |
| `rebuilt > supplied` **after** the fix | **8** |

So the fix takes the rebuild from uniformly-too-short to too-long on 8 fixtures. A reader of the
original table would not have known that.

### What the 8 actually are — mostly *wrong values*, not spurious rows

Joining those 8 to their declared BALs **by case** (3 of the 8 had 3 candidate cases each and are
reported as **not joined** rather than resolved by picking):

| fixture | txs | declared bal/nonce | guest bal/nonce | reading |
|---|---|---|---|---|
| `block_access_lists/bal_all_transaction_types.json` | 5 | 10 / 6 | 9 / 6 | **still missing a row**, yet longer |
| `block_access_lists/bal_cross_tx_factory_nonce_create_chain.json` | 8 | 16 / 23 | 11 / 23 | **missing five**, yet longer |
| `block_access_lists/bal_cross_tx_storage_chain.json` | 8 | 16 / 8 | 11 / 8 | **missing five**, yet longer |
| `block_access_lists/bal_cross_tx_balance_dependency.json` (case A) | 2 | 5 / 2 | 5 / 2 | exact counts → extra bytes are **wrong values** |
| `block_access_lists/bal_cross_tx_balance_dependency.json` (case B) | 2 | 6 / 2 | 6 / 3 | one **spurious nonce row** |

**Same or fewer rows with more bytes can only mean wrong values** — a pigeonhole argument, and the
first value-level evidence in this investigation, obtained without reading a value. RLP encodes
integers minimally, so more bytes means a **numerically larger** value.

**Explicitly: the fix did not introduce systemic over-emission.** Across the five joined cases there is
**one** spurious row in total, and three still under-emit. The 8 overshoot on *length*, not on row
count.

### Post-merge status

On merged main (`d7232de14`, my tier deletion composed with the read-intern repair, guest
`a44e88b4464a3974`) the six localisation fixtures all still show **nonce deficit 0** and balance
deficit 0 except the two filed residuals (#10809, #10808) — the merge moved neither. Diagnostic cells
verified still publishing (18 stores for 18 cells) and `dataSizeBytes` unchanged at `0x5370`.

Related: #10815 (the whole-account/per-component divergence, whose insufficiently-debited sender
predicts exactly this sign), #10809, #10808.

PR #10812, opened 2026-07-28
refactor(codegen): parameterise BloomAddValue over GuestLayout (GH #10753)

pirapira:
Second conversion under #10753 (first: #10800 HashBridge; generator fix: #10803). **GuestLayout: +2 fields (bav_hash, bloom_add_value) → 10 total.** `zkvm_keccak256` was already present from the HashBridge group.

First conversion with a **direct** GuestAddrs import in the original module (HashBridge's was transitive-only through it), so this is the first independent generality test of the #10803 image-coverage fix.

## Shape

- `Programs/BloomAddValueProg.lean` (new leaf): `bloomAddValue_prog_of (L : GuestLayout) : Program`, relocs, Function + `_eq_prog` + #guards at `(bloomAddValue_prog_of .zero)`. Imports Emit/AsmReloc/GuestLayout — no GuestAddrs, no HashBridge.
- `Programs/BloomAddValue.lean` (bridge, module name unchanged): `def bloomAddValue_prog : Program := bloomAddValue_prog_of guestLayout` + the probe unit (which imports HashBridge for `zkvmKeccak256Function`). Zero consumer edits.
- MANIFEST.tsv unchanged; layout detection picks up the leaf automatically. Leaf block generated by `gen_lean_layout` (#10800 generator support) so the drift gate pins it verbatim.

## Acceptance (all re-run on this branch, none inherited)

- `lake build codegen`: clean (4278 jobs).
- `check-file` / `check-all`: CLEAN, 0 drifts across 148 files — render gates (symbolic .text-identical, concrete fn#c == guest-linked bytes) and layout drift gates pass.
- **Byte-identity oracle re-run** (not inherited): `guest_image_coverage.py --emit-lean` regenerates `GuestImageEntries.lean` byte-identical.
- **Flip test**: `bav_hash` 0xab2a0150→0xab2a0158 in GuestAddrs.lean, rebuild → `BloomAddValueProg.olean` sha256 unchanged (d6a596d1…766), only GuestAddrs/GuestLayoutInstance/bridge rebuild. Restored; `check-guest-addrs` CLEAN (1125 symbols).
- Emission-neutral: conversion moves no bytes (concrete-render gate ties the bridge-applied program to the guest-linked image). No decide-pinned PC touched.

Diagnostic-only structural change; no guest behavior change.


pirapira:
Reviewed, and this is the PR I most wanted to see, because it is the first **independent** test of #10803 rather than a second application of it.

## Why the generality claim is real

HashBridge was blocked from conversion only *transitively*. `BloomAddValue` imported `GuestAddrs` **directly**, which is a different starting shape — so if the generator regexes in #10803 had been fitted to the module they were written against, this is where it would show. It doesn't.

Verified on the branch rather than taken from the report:

- **`GuestImageEntries.lean` is byte-identical to main** — the diff is empty. That is the acceptance oracle, and it means the generator emitted the **concrete bridge program** for the linked image, not the zero-layout one.
- **No layout movement at all**: `GuestAddrs.lean`, `symbol-addresses.tsv`, `RegionMap.lean` and `Proofs/GuestImage.lean` are all untouched. So the parameterisation is emission-neutral on a second, structurally different module.
- **The leaf imports no `GuestAddrs`** — only `Emit`, `AsmReloc`, `GuestLayout`. The direct dependency is genuinely gone rather than relocated.

Two data points is not a proof of generality, but it is the difference between "works on the module it was written for" and "works on a module it wasn't." Worth stating in the body as such.

## The split is right and the field reuse is the good detail

`GuestLayout` grows by exactly two — `bav_hash`, `bloom_add_value` — to 10 total, with `zkvm_keccak256` **reused** from the HashBridge group rather than redeclared. That reuse is the property that decides whether this theme scales: if every conversion added its full symbol set, the field count would grow superlinearly toward the elaboration limit you measured. Sharing across groups is what keeps it linear in *distinct* symbols. Keep reporting the running total per PR.

`GuestLayoutInstance` binds both new fields straight from `GuestAddrs`, so it still cannot drift from the table, and it remains the sole importer on this path.

## The probe unit is quietly good evidence for the design

`ziskBloomAddValueProbeUnit` keeps its own `.data` section declaring a local `zk3_state` and `bav_hash`, while the guest has its own via `GuestAddrs`. **Two images, two different addresses for the same symbol names, one shared emitted string** — which works only because `emitProgramR` keeps `la`/`jal` symbolic and each image relocates against its own layout.

That is the layout-independence claim the `GuestLayout` docstring makes, demonstrated rather than asserted. It might be worth one line in the docstring pointing at this probe as the worked example, since a future converter reading "emission is layout-independent" will want to know what that buys.

## One note, not blocking

The `#guard`s and `_eq_prog` are stated at `_of .zero`, consistent with HashBridge. As established on #10800 that is sound because two pre-existing gates bound it — the emitted-reloc-count check rejects any `la`/`jal` baked from a layout, and `assemble_cmp` against the hand-written fixture catches a layout-dependent value that is *not* a `la`/`jal`. This PR inherits that reasoning rather than re-establishing it, which is fine; the probe above is incidentally a third piece of evidence for it.

Also carried over from #10800 and still true here: `bloomAddValueFunction_eq_prog` is a **tautology** — `bloomAddValueFunction` is *defined* as that `emitProgramR` expression and the theorem restates the definition, so `rfl` closes it by unfolding. The real drift guard is `check_file`'s verbatim block match plus `assemble_cmp`. Not introduced by this PR and not to be fixed here; noted only so the "kernel-checked drift guard" wording is not relied on.

PR #10813, opened 2026-07-28
refactor(codegen): remove stale GuestAddrs decide pins (GH #10790)

pirapira:
Deletes two unused theorem declarations that pin concrete linked addresses — the bootstrap-cycle hazard of GH #10790:

- `rlpItemSpanBase_eq` (`EvmAsm/Codegen/Programs/RlpItemSpanSpec.lean`, pinning 0x80004abc)
- `rlpItemSizeBase_eq` (`EvmAsm/Codegen/Programs/RlpSpliceHelperSpec.lean`, pinning 0x80004a30)

Both are referenced nowhere (grep over `EvmAsm/`: only the declaration lines match). The underlying defs are already symbol-based (`BitVec.ofNat 64 GuestAddrs.rlp_item_span` / `.rlp_item_size`), so nothing goes stale after this; the bootstrap cycle (stale pin fails `codegen` which builds the Spec modules) disappears.

**Lookalikes checked and deliberately excluded**: `accBase_toNat` (`EvmAsm/Codegen/Programs/U256MulU64Be/OuterLoop.lean:77`) is CONSUMED by `rw` at lines 83/103 (data/.bss region class), and `EVM_MEMORY_AREA_toNat` (`EvmAsm/Evm64/StateAssertions.lean:342`) is consumed at 353/375/395 (region base, rfl-closed) — both are live and are left alone.

**Acceptance**
- `lake build codegen` clean.
- Emitted guest byte-identity vs main: emitted `.s` byte-identical (sha 53e59f26... both); stripped ELFs identical (sha 781f02fb... both); full-ELF diff is only the embedded object filename. Section layouts unchanged.
- `scripts/check-axioms.sh`: OK — 95 witnesses, only classical axioms (propext, Classical.choice, Quot.sound), 0 grandfathered forbidden-tactic owners. (Note: needed one `LAKE_ARTIFACT_CACHE=false lake build EvmAsm.Progress` first — the artifact-cache mode incompatibility from AGENTS.md, not a gate failure.)

### DIFF
--- a/EvmAsm/Codegen/Programs/RlpItemSpanSpec.lean
+++ b/EvmAsm/Codegen/Programs/RlpItemSpanSpec.lean
@@ -149,8 +149,6 @@ theorem cursor_lt_length (items : List RLPItem) (k : Nat) (hk : k < items.length
 /-- Guest entry of `rlp_item_span`. -/
 def rlpItemSpanBase : Word := BitVec.ofNat 64 GuestAddrs.rlp_item_span
 
-theorem rlpItemSpanBase_eq : rlpItemSpanBase = (0x80004abc : Word) := by decide
-
 /-- The `rlp_item_span` body at its linked guest address. -/
 abbrev rlpItemSpanCode : CodeReq :=
   CodeReq.ofProg rlpItemSpanBase rlpItemSpan_prog
--- a/EvmAsm/Codegen/Programs/RlpSpliceHelperSpec.lean
+++ b/EvmAsm/Codegen/Programs/RlpSpliceHelperSpec.lean
@@ -694,8 +694,6 @@ theorem rlp_item_size_form_own_spec_within (base ptr raVal : Word)
 /-- Guest entry of `rlp_item_size`. -/
 def rlpItemSizeBase : Word := BitVec.ofNat 64 GuestAddrs.rlp_item_size
 
-theorem rlpItemSizeBase_eq : rlpItemSizeBase = (0x80004a30 : Word) := by decide
-
 /-- The `rlp_item_size` body at its linked guest address. -/
 abbrev rlpItemSizeCode : CodeReq :=
   CodeReq.ofProg rlpItemSizeBase rlpItemSize_prog


pirapira:
Reviewed. **Approve.** The diff is `+0/-4` — the two theorem lines and nothing else — which is the right size for this change and worth stating, because a "remove the address pins" PR that touched anything else would need a different review.

## The claim I checked hardest: nothing consumed them

Verified independently before reading the PR, and again after:

- `git grep` for `rlpItemSpanBase_eq` / `rlpItemSizeBase_eq` across `EvmAsm/**` returns **only their own declaration lines**.
- **Neither carries an attribute** — no `@[simp]`, no `@[grind]`, nothing that would enrol them in a set and make them load-bearing without a textual reference. That is the way a "consumerless" claim usually turns out to be wrong here, so it needed checking separately from the grep.
- **Neither appears in `EvmAsm/Progress.lean`**, so the `check-axioms.sh` witness list is unchanged and the axiom gate still covers exactly what it covered before. Deleting a theorem is precisely when you would want to confirm that, since the gate derives its names from that file.

## Why deletion is the right fix rather than repinning

The **defs survive and were already symbol-based**: `rlpItemSpanBase = BitVec.ofNat 64 GuestAddrs.rlp_item_span`, and the `rlp_item_size` equivalent. Everything real consumes the *defs* — `CodeReq.ofProg` for both `rlpItemSpanCode` and `rlpItemSizeCode`, `cpsTripleWithin`, `rlp_item_size_form_own_spec_within`. The one `unfold` site unfolds the defs directly rather than routing through the `_eq` theorems.

So the two theorems were the **only** place a literal address appeared, and nothing read them. This does not weaken any proof: no goal was being closed by them.

## What it buys, and it is more than tidiness

These two pins had accumulated a real cost, all of it observed in a single evening:

- **A bootstrap cycle.** `lake exe codegen` builds these Spec modules, so a stale pin fails codegen — and codegen is the only thing that can report the new address. Not repinnable in one pass.
- **Deceptively late failure.** `lake` serves a cached `olean` for ~2 passes, so the break surfaces after step 2 of the regen and reads as *"step 2 did this"* when step 1 had been building against a stale cache throughout.
- **The documented escape hatch was insufficient** — commenting `Main.lean`'s `Proofs.*` imports does not help, because the failing module reaches the exe via `Programs/Imports.lean → Registry.lean`. The cut has to be made in `Imports.lean`.
- **Merge false positives.** A residual-diff audit flags the superseded literals as lost content; "fixing" that by restoring an older branch's values breaks the build while silencing the instrument.
- **Three PRs in one evening** had to repin them, and one agent had to temporarily rewrite both theorems to `BitVec.ofNat 64 GuestAddrs.<sym>` just to get codegen to run.

Deleting them **removes the cycle rather than documenting it**: codegen still builds these modules, but with no literal to go stale there is nothing left to fail.

## Correctly scoped — the two lookalikes are left alone

Right call, and worth having said so in the body: `accBase_toNat` (`U256MulU64Be/OuterLoop.lean:77`) **is** consumed, by `rw` at `:83` and `:103` in bound proofs that then close with `omega`, so removing it means reformulating those bounds symbolically — real proof work, and its address sits in the `.data`/`.bss` region, the same class as the `bnf_le_a` pin. `EVM_MEMORY_AREA_toNat` pins a **region base** rather than a per-symbol address and is closed by `rfl`, so it only moves when regions move. Bundling either would have turned a 4-line deletion into a proof-restructuring PR.

## Acceptance

Byte-identical emitted `.s` is the right oracle for a proof-only change, and the correct expectation: if anything *had* moved, something consumed those theorems after all and the grep was wrong. No generated files touched, so no regen and no repin fixpoint needed — which is itself a small demonstration of the point.

pirapira:
Dropped from the #10812/#10814 batch — **real source conflict against main, not a generated-file collision.**

```
CONFLICT (content): EvmAsm/Codegen/Programs/RlpItemSpanSpec.lean
CONFLICT (content): EvmAsm/Codegen/Programs/RlpSpliceHelperSpec.lean
```

I aborted the merge rather than resolving it. The reason is specific: those are the two files holding the `decide`-pinned guest PCs, and **#10811 just repinned them in main** (`rlpItemSpanBase` -> `0x80004dc0`, `rlpItemSizeBase` -> `0x80004d34`, verified against the relinked ELF) while this PR's purpose is to *delete* those pins entirely.

So the conflict is delete-versus-repin, and resolving it means deciding whether those verification statements exist at all. That is a statement-level decision belonging to whoever owns #10790, not something an automated merge should settle — even though taking the deletion is probably what you intend, I am not going to be the one to remove `decide` pins.

**What it needs:** a merge of main `d7232de14` by its author, resolving both files in favour of the intended removal. Once that is done and it is CLEAN, I will batch it like anything else.

For reference, #10814 (now in #10816) documents the related finding from the same files: the linked ELF is the arbiter for a pinned address, not either branch — which is what made #10811's pins the correct ones rather than the sibling branches' older literals.

pirapira:
Post-branch note (conflict is merge-bot's to resolve): this PR is now in a change-versus-delete conflict with main — main re-pinned the two theorems to 0x80004dc0 / 0x80004d34 (scoper's pre-state tier deletion shrank .text and moved both guest PCs), while this branch deletes the versions reading 0x80004abc / 0x80004a30. So the PR that removes these pins has been conflicted by exactly the churn it exists to eliminate — the strongest evidence for the deletion, demonstrated on the removal patch itself. The resolution is unambiguous: delete the theorem lines at whatever literals they now carry; there is no correct value to preserve. Post-landing I will re-verify on main that both theorems remain referenced nowhere and carry no attribute.

Issue #10601, opened 2026-07-26
check-axioms.sh cannot run locally: the artifact cache satisfies the build graph without materializing oleans (207 modules)


pirapira:
> **Note:** This issue was written by Claude Code.

`scripts/check-axioms.sh` — the kernel-truth backstop that rejects `sorryAx`, `Lean.ofReduceBool`, `Lean.trustCompiler` and the `bv_decide`/`native_decide` trust axioms — **cannot be run locally**. CI is currently the only place the axiom gate executes.

## The failure

```
$ ./scripts/check-axioms.sh
check-axioms: 'lake env lean' failed — output follows:
/tmp/tmp.XXXX.lean:1:0: error: object file
  '.lake/build/lib/lean/EvmAsm/Rv64/Word.olean' of module EvmAsm.Rv64.Word does not exist
```

It dies on the first import.

## Cause: the build graph is satisfied without materializing oleans

`lake build` completes successfully — 4750 jobs, all green — while **207 of the `EvmAsm` modules have no `.olean` file on disk**:

```
207 of the EvmAsm modules have no olean
   .lake/build/lib/lean/EvmAsm/Evm64/LogArgsStackDecode.olean
   .lake/build/lib/lean/EvmAsm/Evm64/StorageGas.olean
   .lake/build/lib/lean/EvmAsm/Evm64/EvmWord.olean
   ...
```

Under `LAKE_ARTIFACT_CACHE=true` these report as `Fetched`, and the cache satisfies the dependency graph without writing the artifact into `.lake/build/lib`. For `EvmAsm.Rv64.Word` the sidecars are present but the artifact is not:

```
Word.ilean   Word.ilean.hash   Word.olean.hash   Word.trace     <- present
Word.olean                                                      <- absent
```

`lake build EvmAsm.Rv64.Word` reports `✔ Fetched` and completes without creating it. Deleting `Word.trace` and `Word.olean.hash` to force re-materialization **did not work** — it fetched again and still wrote no olean.

The distinction that matters: **`lake build` needs the cache; `lake env lean` needs the file.** So the tree is simultaneously fully built and unusable for any import-based tool.

## Why this is worth an issue rather than a workaround

The gate this blocks is the one described in `CLAUDE.md` as the *"kernel-truth backstop"* — the complement to `check-forbidden-tactics.sh`, which is only a source scan. The source scan cannot see a `sorryAx` reaching a theorem through an import, which is exactly what the axiom gate is for.

[...]

PR #10819, opened 2026-07-28
ci(axioms): check axioms via checked-in witness module built by lake build (GH #10601)

pirapira:
## What

GH #10601: replace the `lake env lean` scratch-file axiom check (which dies under `LAKE_ARTIFACT_CACHE=true` — cache satisfies the graph without materializing oleans, so the scratch compile cannot resolve imports) with a **checked-in witness module** built by `lake build`:

- `EvmAsm/Progress/AxiomWitnesses.lean` (new): one `#print axioms` per Progress witness; placed under `EvmAsm/Progress/` to satisfy the L2 layering rule (only `Progress.lean` and `Progress/*` may import `Progress`). Reached via the umbrella (`EvmAsm.lean`) — a deliberate choice: the reports are elaborated in the default build and the module rebuilds whenever `Progress.lean` changes (hub-reachable).
- `scripts/gen-axiom-witnesses.py` (new): regenerates the module from `EvmAsm/Progress.lean` (same extraction as the gate); the gate diffs generated vs checked-in and fails with the regenerate hint (10804 regenerate-and-diff pattern).
- `scripts/check-axioms.sh`: drift check + `lake build EvmAsm.Progress.AxiomWitnesses`, strips `info: <pos>: ` prefixes, then the UNCHANGED classifier. **Completeness assertion**: the set of witnesses that produced a report must equal the witness list exactly — a missing report names the witness and fails the gate. This is not about cached builds emitting nothing (replay is guaranteed, measured below); it is protection against the sed pattern or lake's message format/channel changing and yielding zero parseable records — a summary line quoting the number of names is not evidence the names were checked.

## Verification

- **Cache mode**: gate passes on an artifact-cache checkout (the mode that previously killed it at the `lake env lean` line): `OK — 95 witnesses; only classical axioms`.
- **Replay measured**: two consecutive `lake build` runs both emit all 95 report lines; sorted outputs byte-identical.
- **Falsified**: planted a `sorry`-backed theorem + witness entry; the gate FAILS. What the failure looks like: appending the planted `#print` line to the witness module dies at the drift check (stale vs generator); after regenerating, the build itself errors (`declaration uses 'sorry'`, lake exit 1) before the classifier — the classifier's sorry bucket is defense-in-depth for stale/snapshot oleans.
- **Negative test of the completeness assertion**: doctored the lake log to drop one report; gate fails naming exactly that witness.
- **CI question answered**: CI restores `.lake` via actions/cache (oleans materialized), runs full `lake build`, then the gate — so the old check was meaningful in CI, not lucky. The new gate additionally survives artifact-cache-mode local checkouts, and its reports are replay-guaranteed on cached rebuilds.


pirapira:
Reviewed. **Approve with one gap to close.** The core move is right and the root cause is correctly addressed: `lake env lean` needs a materialized `.olean`, `lake build` needs only the cache, so a checked-in witness module built by `lake build` makes the gate work in cache mode. That is #10601's design, followed rather than redesigned.

## What is right

- **The drift check is real.** `gen-axiom-witnesses.py` regenerates and `diff`s against the checked-in file, fails on stale, and prints the `--write` command to fix it. `EvmAsm/Progress.lean` stays the single source of truth, so the checked-in list cannot silently diverge from the registry — which was the new surface this approach introduces.
- **The replay claim is measured, not assumed.** The comment says lake replays recorded messages even when the module is cache-satisfied, *"(measured on this checkout)"*. That was the one premise I had asserted wrongly in the other direction, so having it measured and the measurement noted in-source is exactly right.
- **The classification logic is untouched.** A `sed` strips lake's `info: <file>:<line>:<col>: ` prefix so the section-3 parser consumes the same record shape as before. Minimal-diff, and it means the axiom allowlist semantics carry over unchanged rather than being re-derived.
- **95 witnesses**, matching the registry.

## The gap: nothing asserts the report is COMPLETE

`${#NAMES[@]}` is used for the zero check at `:86` and for the two human-facing summary lines at `:241` and `:317` — but **nothing compares the number of parsed reports against the number of witness names.** So the success message will cheerfully print *"OK — 95 witnesses; only classical axioms"* whether 95, 3, or 0 reports were actually parsed.

That matters more here than it did before the change, because the failure modes are now different in kind:

- a `sed` pattern that stops matching a future lake output format silently yields zero records;
- a lake version that changes the message channel does the same;
- a partial build log, a truncated pipe, or a report format change all land the same way.

In every case the gate reports a **pass** with a count taken from the *input* list rather than from the *output* it actually parsed. **A summary line that quotes the number of names is not evidence that the names were checked** — and that is precisely the shape (a signal that cannot distinguish success from non-execution) that has produced several false-clean gates in this repo, including the exclusions-only pathspec and a regen fixpoint loop that compared file contents.

Please add: parsed-report count must equal `${#NAMES[@]}`, and exit nonzero naming the missing witnesses otherwise. One comparison, and it converts every one of the above from a silent pass into a loud failure.

I originally offered this as protection against cached builds emitting nothing, which was **my error** — that premise was wrong and is withdrawn. The assertion is still worth having, for the reasons above rather than that one.

## Please also confirm the falsification

The plan included planting a deliberate `sorry` in a scratch witness, confirming the gate **fails**, then removing it. If that was run, say so in the body with what the failure looked like; if not, it is worth doing before merge. This gate is the kernel-truth backstop and it has never been observed rejecting anything on this path — **an interlock nobody has seen fire is indistinguishable from one that cannot.**

- Adding the module to the `EvmAsm.lean` umbrella means the 95 reports are elaborated as part of the default build, not only when the gate runs. That looks like the right trade — the gate's data becomes a by-product of the normal build, and `#print axioms` on an already-elaborated constant is cheap — but it does make the module hub-reachable, so a `Progress.lean` change now rebuilds it. Worth a line in the PR body so the CI cost is a stated choice rather than a side effect.
- The open question from #10601 is still worth answering separately: whether CI **materializes** oleans or merely happens to run the script before any cache fetch. That decides whether CI's historical passes were meaningful or lucky, and it is a bigger question than the ergonomics fix.

pirapira:
Completeness gap closed — **approving and labelling.** Verified on the branch rather than from the report:

- `REPORTED` is derived by extracting the quoted name from each parsed record, `LC_ALL=C sort -u` (`:145-152`); `NAMES` is built with the **same collation** at `:84`. So the `diff` and `comm -23` comparisons are well-founded rather than accidentally agreeing on today's data — that was the one way this fix could have introduced a spurious *failure*.
- It compares **sets, not just counts** — a count check would pass if one witness were missing and an unrelated record appeared. Stronger than what I asked for.
- The rationale in-source is the correct one (`sed`/format/channel robustness), not the cached-no-messages premise I withdrew.

## The negative test is the part that makes this trustworthy

Doctoring the stripped log to drop exactly one report, and observing the gate **fail naming exactly that witness**, is the check that distinguishes an interlock from a decoration. An assertion nobody has seen fire is indistinguishable from one that cannot.

## The `sorry` falsification found something about the gate's structure

Worth recording beyond this PR: planting a `sorry`-backed witness **fails at the build**, with `declaration uses 'sorry'` and `lake` exiting 1 *before the classifier runs*. So the `sorryAx` bucket in the classifier is **not** the primary sorry detector — the build is.

That sharpens what this gate is actually for. Its distinctive value is the axioms that **do not** fail a build: `Lean.ofReduceBool`, `Lean.trustCompiler`, and the `bv_decide`/`native_decide` trust axioms — plus a `sorry` arriving through a **stale or snapshot olean**, where the build would not re-elaborate and so would not complain. That is real defence-in-depth, but it is a narrower and more specific job than "the kernel-truth backstop" suggests, and it is worth stating so nobody assumes the classifier is what stops `sorry`.

## #10601's open question is resolved, in the reassuring direction

*"whether CI materializes oleans or merely happens to run the script before any cache fetch"* — answered: oleans **are** materialized via cache restore, with a full `lake build` before the gate. So **CI's historical passes were meaningful, not lucky**, and the new form additionally survives artifact-cache checkouts where the old one could not run at all. That was the bigger of the two questions on that issue and it is good to have it closed rather than carried.

Umbrella import stated as a deliberate choice in the body, with its consequence (95 reports elaborate in the default build; the module is hub-reachable and rebuilds with `Progress.lean`) — right call to make that explicit rather than a side effect.

Issue #10820, opened 2026-07-28
bal: balance_changes.post is BE32 but handed to the LE-only scalar pair, so every balance row is emitted as a 32-byte string (predicted overshoot matches measured to 0-3 bytes)


pirapira:
## Claim

`balance_changes.post` is **BE32** in the builder row, but it is handed to
`bal_rlp_scalar_len` / `bal_rlp_emit_scalar`, which are documented **and implemented for LE limbs**. So
**every balance row is measured and emitted as a full 32-byte string instead of its minimal encoding**,
and the emitted bytes are the value byte-reversed.

The authoritative field table (`BlockAccessListBuilder.lean:192`) declares this field **LE**. That
declaration is wrong.

This is the **same defect class as the one already documented** at `BalSerializer.lean:426-436`, where
`bal_serializer_u64_to_field` wrote LSB at byte 31 under a comment calling it big-endian while
`bal_rlp_scalar_len`'s docstring called byte 31 the canonical BE most-significant byte — "both said BE
and meant opposite layouts, so the two agreed in prose and disagreed in bytes." Same shape, different
leg, still live.

## Why the field is BE32, traced end to end

1. **The producers compute BE.** `u256AddBe_prog` (`U256.lean:59`) starts at index 31 and propagates
   carry **31 → 0**, so byte 0 is most significant. That is big-endian. Same for `u256_sub_be`,
   `u256_mul_u64_be`, and `balance_at_header_state_root`.
2. **`record_nonstorage_effect` forwards the post pointer verbatim** to `account_write_record`
   (`NonstorageEffectLog.lean:120`, `a1 = s2 = post_balance`).
3. **`account_write_record` copies 4 dwords verbatim** into `row+32` (`.Lawr_store`).
4. **`bal_builder_append_balance` copies 4 dwords verbatim** into the builder row `+32`
   (`BlockAccessListBuilder.lean:343`).
5. **`bal_serializer_emit_balance` hands `row+32` straight to the scalar pair** — no conversion, unlike
   the storage slot, which is reversed first via `bal_serializer_slot_to_le` *precisely because* the
   scalar pair is LE-only.

`bal_rlp_scalar_len` scans **down from byte 31** for the most significant byte. On a BE32 field holding
a 12-byte value (right-aligned, bytes 20..31), byte 31 is nonzero, so it reports **32** significant
bytes.

The table is internally consistent for the *storage* value (an LE EVM stack word) and for the u64 `bai`
[...]

PR #10821, opened 2026-07-28
codegen: record modeled system storage reads

pirapira:
## Modeled-system storage reads

This records `SLOAD` keys from modeled system calls directly in the
block-lifetime `storage_reads` set.  Ordinary transactions keep their existing
transaction-local recorder and promotion boundary.

The main result is the first observed byte-exact rebuilt BAL match: on
`blockchain_tests/for_amsterdam/amsterdam/eip7928_block_level_access_lists/block_access_lists_eip4788/bal_4788_query.json`, fixture `00548` (`no_value-valid_timestamp`, block 0), the shadow result changes from a mismatch with zero emitted storage reads to a match with 16 emitted storage reads.  Its production output prefix (bytes 0..119, including root and verdict) is byte-identical to fresh main.

The direct route is necessary because modeled system calls execute `SLOAD` but
do not traverse the user-transaction promotion boundary.  The new
`storage_read_record_block` uses the same execution-key canonicalisation and
set deduplication as the existing transaction recorder; the builder still owns
read/write suppression.

## Evidence

- Fresh-main baseline snapshot: `a44e88b4464a3974379771d53a420c8045882707af2dc14926c83d4f585e8f40`.
- Candidate snapshot: `47792f733aa0b4949ac3e8e5de32a95c70c45aa09a236d3056d60d6148a00192`.
- On 00548, PC trace reaches `h_SLOAD`, with `storage_read_record=32`,
  `storage_read_record_block=32`, and both
  `read_sets_incorporate_tx=0` and `account_state_commit_pending=0`.
  This proves the modeled-system route is reached and bypasses the ordinary
  transaction boundary as intended.
- All 12 EIP-4788 query fixtures and all 9 selected EIP-2935
  query/simple controls retained `succ=1` on the candidate.
- `check-region-map.sh` passed after a converged relink:
  `.text=0x67024`, `.data=0x5370`; `RegionMap.textSizeBytes` matches.

No full-manifest sweep was run: the maintainer suspended that expensive gate.
The local random-200 harness could not run because this checkout has only the
prepared `/tmp/fc668` inputs, not the upstream EEST fixture tree it requires;
this is disclosed as a harness limitation rather than treated as authorization
to substitute a full sweep.


pirapira:
Reviewed. **One correctness finding to fix before merge** — in the two-line dispatch, not in the 60-line routine. The new routine itself is right, and the headline result is real and well framed.

## ⛔ The dispatch clobbers the caller's `t0` on **both** paths

```
storage_read_record:
  la t0, system_call_mode; ld t0, 0(t0); bnez t0, storage_read_record_block   <- ADDED
  addi sp, sp, -112
  sd t0, 0(sp); sd t1, 8(sp); ...
```

The branch is placed **before the prologue**, so `t0` is destroyed *before* it is saved:

- **mode 0** (the ordinary path, overwhelmingly the common one) — falls through and `sd t0, 0(sp)` stores the **clobbered** value. The epilogue restores that, so the caller resumes with `t0 = system_call_mode` instead of its own value.
- **mode 1** — tail-branches into `storage_read_record_block`, whose own prologue saves the already-clobbered `t0` and restores it. Same corruption.

**This is not a "temporaries are caller-saved, so who cares" case.** `storage_read_record` deliberately saves and restores `t0`–`t6`, `ra` and `a0`–`a2` across a 112-byte frame — it is written to be *transparent* to temporaries. And its callers rely on that: `Storage.lean:226-232` saves only `x1`, `x10`, `x11` around the call and nothing else. So the transparency contract is load-bearing, and this change silently breaks it.

**Why the pinned run would not catch it:** the corruption is invisible unless a caller holds a live value in `t0` across the call. It is exactly the shape that surfaces much later as an unrelated-looking failure in a different routine. `succ=1` on 21 fixtures is not evidence here.

Cheapest correct fix: put the prologue **first**, test after it, and on mode 1 `jal` the block recorder then jump to the shared epilogue. That costs one frame on the system-call path — which runs a handful of times per block — and makes the register discipline obvious rather than subtle. Please do not fix it by picking a different temporary; every one of `t0`–`t6` is in the saved set, so the problem is the placement, not the register.

## The new routine is correct as far as I can check it

- **Dedup scan** mirrors the tx-local one: 64-byte stride, eight dword compares (four address, four slot), `.Lsrrb_next` on any mismatch.
- **`ensure_account` runs on the dedup-hit path too** (`j .Lsrrb_intern_account`), not only on append. That matches the spec, where `add_storage_read` calls `ensure_account` unconditionally before touching the set — and it is the idempotent behaviour a dict upsert gives for free.
- **Scratch placement is sound**: `addi a1, sp, 64` lands in bytes 64–87, which the save set leaves free (saves occupy 0–56 and 88–104), and 24 bytes is enough for the 20-byte BE address. Clobbering `a1` mid-routine is safe since it is restored from `96(sp)`.
- **Overflow is handled** — 16384 cap with the `storage_reads_overflow` flag, same shape as the tx-local path.
- The docstring correctly states that read/write suppression happens later in the builder, which is where the spec puts it (slot-level, at build time). Good that this does *not* try to suppress at record time.

## The headline is right and worth the prominence

Foregrounding `bal_4788_query` as the **first byte-exact rebuilt-BAL shadow match** (0 reads/mismatch → 16 reads/match) is correct framing. A match means the whole pipeline — producers, canonical sort, RLP, keccak — agreed byte-for-byte, which was an open question while the digest matched 0 of 80. The remaining gaps are therefore per-leg producer defects rather than machinery.

It also corroborates #10820 from the opposite direction: this fixture carries **no balance rows**, and #10820 finds every balance row over-measured because a BE32 field is handed to LE-limb scalar routines. The first fixture able to match being precisely the one with nothing to over-measure is a strong joint signal.

## Evidence discipline

Explicitly recording that **no full-manifest gate was run**, and quoting the pinned tier (12 `bal_4788_query` + 9 EIP-2935 controls, all `succ=1`, production bytes `[0..119]` byte-identical to fresh main), is exactly right — a sampled result that reads as a full-corpus result is the misreading to avoid. The reach trace showing `storage_read_record_block=32` is what proves the new code executed rather than merely linking.

pirapira:
`t0` transparency defect fixed correctly — **approving and labelling.** Verified on `f0fa928f2` rather than from the report:

```
storage_read_record:
  addi sp, sp, -112
  sd t0, 0(sp); ... sd ra, 56(sp)
  sd a0, 88(sp); sd a1, 96(sp); sd a2, 104(sp)
  la t0, system_call_mode; ld t0, 0(t0); beqz t0, .Lsrr_tx
  jal ra, storage_read_record_block; j .Lsrr_done
.Lsrr_tx:
  ...
.Lsrr_done:
  ld a0, 88(sp); ... ld ra, 56(sp)
  addi sp, sp, 112
  ret
```

The prologue now runs first, so the `t0` used for the mode test is already saved and the caller's value is restored on both paths. Three consequences I checked rather than assumed:

- **`ra` survives the inner call.** `jal ra, storage_read_record_block` clobbers `ra`, but the prologue saved it at `56(sp)` and `.Lsrr_done` restores it — so the outer `ret` returns to the original caller, not into the recorder.
- **The arguments survive.** `a0`/`a1` are saved at `88`/`96(sp)` and restored at `.Lsrr_done`, and the block recorder independently preserves them, so the mode-1 path is transparent in the same way the mode-0 path was.
- **`.Lsrr_done` is genuinely shared** — reached from the mode-1 branch (`:130`) and from the tx path (`:168`), with one restore sequence and one `ret`. No duplicated epilogue to drift.

Chosen the placement fix rather than swapping temporaries, which was the right call: every one of `t0`–`t6` is in the saved set, so the problem was ordering, not register choice.

## Evidence

Fresh immutable candidate `47792f73…`, with `00548` reporting production bytes `[0..119]` equal to main, shadow **match** at read-count 16, and the reach trace confirming the block recorder executed. And **catching and discarding a stale intermediate snapshot** before remeasuring is the check that matters most in this repo — a stale ELF measures the previous binary and reports it as a result, which is how a fix gets credited to the wrong artifact.

The headline stands: `bal_4788_query` by relpath is the **first byte-exact rebuilt-BAL shadow match**, taking the digest from 0 of 80. That establishes the whole pipeline — producers, canonical sort, RLP, keccak — can agree byte-for-byte, so the remaining gaps are per-leg producer defects rather than machinery. It also corroborates #10820 from the opposite direction, since this is precisely the fixture with no balance rows to over-measure.

Pinned tier and the explicit "no full-manifest gate was run" statement both retained, which is the right way to report a sampled result.

Issue #10818, opened 2026-07-28
bal: rebuilt storage_reads is short by exactly one row on every read-producing fixture (uniform off-by-one across three families)


pirapira:
## Claim

On merged main (`d7232de14`), the rebuilt BAL emits **exactly one fewer `storage_reads` row than the
supplied BAL declares**, on **every** fixture that produces reads at all — six fixtures across three
different EIP-7928 families.

An identical off-by-one across three families is a **systematic single missing read**, not a fixture
accident.

## Measurement

Guest `a44e88b4464a3974`, spike, built with `LAKE_ARTIFACT_CACHE=false` (see note below); the diagnostic
publish block was verified live before reading — 18 stores for 18 cells, `dataSizeBytes` unchanged.

| fixture (relpath under `…/eip7928_block_level_access_lists/`) | declared reads | emitted reads | deficit |
|---|---|---|---|
| `block_access_lists_eip2935/bal_2935_simple.json` | 17 | 16 | **1** |
| `block_access_lists_eip4895/bal_withdrawal_and_value_transfer_same_address.json` | 17 | 16 | **1** |
| `block_access_lists_eip4895/bal_withdrawal_to_coinbase.json` (`tx_type_0`) | 17 | 16 | **1** |
| `block_access_lists/bal_multiple_storage_writes_same_slot.json` | 17 | 16 | **1** |
| `block_access_lists/bal_cross_tx_storage_write.json` (`tx2_reverts_to_zero`) | 17 | 16 | **1** |
| `block_access_lists/bal_cross_tx_storage_write.json` (`tx2_rewrites_same_value`) | 17 | 16 | **1** |
| `block_access_lists_eip4788/bal_4788_query.json` | 16 | **0** | 16 (separate — #10802) |

Read from `bv_bal_shadow_emit_storage_reads` (`OUTPUT+152`) against the fixtures' declared
`storageReads` totals.

## Why this is newly visible

Before the read-intern repair landed, this counter read **0** on every fixture, so a deficit of 1 was
indistinguishable from a deficit of 17. The repair took it 0 → 16 and thereby exposed the residual. **The
repair is not in question here** — it moved the leg from entirely absent to one row short.

`bal_4788_query` is excluded from the pattern: its producer loop is entirely unfed (#10802), so its 16
is a different defect.

## Not claimed

- **Not localised.** No mechanism is proposed and no read emitter code was examined. This is a count
  against a count.
[...]

Issue #10815, opened 2026-07-28
bal: account_resolve_pre_state resolves per-component where the spec's _get_pre_tx_account is whole-account, so a nonce-only block-map row reads balance from pre-block state


pirapira:
## Claim

`_get_pre_tx_account` (execution-specs `block_access_lists.py:583-598`) is a **whole-account**
lookup:

```python
if address in pre_tx_accounts:
    return pre_tx_accounts[address]      # the ENTIRE Account: balance AND nonce AND code_hash
return pre_state.get_account_optional(address)
```

If the address is in the block-cumulative map, **every** field comes from there. The guest's
`account_resolve_pre_state` instead resolves **per component**, gated on a per-row valid mask, so a
block-map row that carries nonce but **not** balance resolves balance from the *parent-state witness*
— silently ignoring balance changes made by earlier transactions in the same block.

This is a second, distinct divergence in the routine #10807 just fixed. It was named in that PR's body
as deliberately unbundled; filing it so it does not vanish when the PR closes.

## The mechanism, from source

`account_resolve_pre_state` accumulates `s7` bit-by-bit and only fills a component from a later tier
when its bit is still clear:

```
.Larp_block_hit:
  mv s6, t5; ld t0, 112(s6); andi t1, t0, 1; beqz t1, .Larp_block_nonce; <copy balance>; ori s7, s7, 1
.Larp_block_nonce:
  andi t1, t0, 2; beqz t1, .Larp_block_done;  <copy nonce>;   ori s7, s7, 2
```

So a block-map hit whose mask lacks `accountWriteHasBalance` leaves `s7 & 1 == 0`, and
`.Larp_header_found` then fills balance from the parent witness.

## Reachability: three of the four producers never set the balance bit

Every `account_write_record` call site and its mask:

| call site | mask |
|---|---|
| `NonstorageEffectLog.lean:120` | `BALANCE + NONCE` |
| `TxIntrinsicStateGas.lean:278` | `NONCE + CODE` |
| `BlockVerdictMtxRuntime.lean:365` | **`NONCE` only** |
| `CreateCodeEffectLog.lean:856` | `NONCE + CODE + STATE` |

Masks accumulate by `or` within a transaction (`account_write_record`'s
`ld t3, 112(t5); or t2, t2, t3`), so a row lacks balance only if the address was *never* written with a
balance-bearing mask. That is not hypothetical — it is the **sender's normal state**:
[...]

Issue #10809, opened 2026-07-28
bal: sender balance row lost on bal_withdrawal_to_coinbase because the STAGED POST equals the pre (residual after #10807)


pirapira:
## Claim

After #10807, one balance row is still lost on

`blockchain_tests/for_amsterdam/amsterdam/eip7928_block_level_access_lists/block_access_lists_eip4895/bal_withdrawal_to_coinbase.json` (`tx_type_0`)

and it is the **transaction sender's** row. Unlike #10807, the defect here is on the **post** operand:
the resolver's pre value is correct, and the **staged post at `32(s4)` still carries the pre
balance** — the sender's debit was never staged into the transaction's account-write record.

This is a **different mechanism** from #10807, which resolved the same fixture's nonce deficit
(1 → 0) but left its balance deficit at 1.

## Evidence

Producer-side diagnostic cells (`bald_*`, #10799), spike backend, guest sha256
`09e4aec070768d7e7edd56ecd5ddd14aaf23d47278fc002dd9fe5d6572485fc0`:

| | before #10807 | after #10807 |
|---|---|---|
| balance declared / emitted / **deficit** | 4 / 3 / **1** | 4 / 3 / **1** |
| nonce declared / emitted / **deficit** | 1 / 0 / **1** | 1 / 1 / **0** |
| balance `eq_bai_mask` | `{1}` | `{1}` |
| equal-row address (bytes 0..15) | `0xf6c3a9edc1afa0ad5b720e4d42e1437c…` (**tx sender**) | same |
| equal-row balance limb at offset 24 | `0x000000e83c80d09f` | **unchanged** |

The fixture declares the sender's pre-balance as `0x033b2e3c9fd0803ce8000000` and its post as
`0x033b2e3c9fd0803ce7e7063b` — a genuine delta. The last 8 bytes of that **pre** value in
big-endian memory are `9f d0 80 3c e8 00 00 00`, which an `ld` reads as the little-endian u64
`0x000000e83c80d09f` — exactly the cell's value. So both compare operands hold the **pre** balance.

The decisive part is that this value is **unchanged by #10807**. That fix altered only the tier
supplying the *pre* operand, and on the sibling fixture
`bal_withdrawal_and_value_transfer_same_address` the same cell moved from `0xd053672f3d80d09f` to
`0` (its sender row now appends). Here it did not move, so the equality is not coming from a wrong
pre — it is coming from a post operand that equals the pre.

## Contrast that localises it

[...]

Issue #10808, opened 2026-07-28
bal: coinbase balance row still lost at block_access_index 3 on a three-tx block (residual after #10807)


pirapira:
## Claim

After #10807 removes the mutated-state tier from `account_resolve_pre_state`, one balance row is
still lost on

`blockchain_tests/for_amsterdam/amsterdam/eip7928_block_level_access_lists/block_access_lists/bal_multiple_storage_writes_same_slot.json`

and the lost row belongs to the **coinbase at `block_access_index 3`** on a three-transaction block.
This is a **different mechanism** from #10807: that fix took the fixture's balance deficit from 2 to
1 and its nonce deficit from 1 to 0, leaving this one untouched.

## Evidence

Producer-side diagnostic cells (`bald_*`, described on #10799), spike backend, guest sha256
`09e4aec070768d7e7edd56ecd5ddd14aaf23d47278fc002dd9fe5d6572485fc0`:

| | before #10807 | after #10807 |
|---|---|---|
| balance declared / emitted / **deficit** | 6 / 4 / **2** | 6 / 5 / **1** |
| nonce declared / emitted / **deficit** | 3 / 2 / **1** | 3 / 3 / **0** |
| balance `eq_bai_mask` (indices with a compare-equal row) | `{1,3}` | `{3}` |
| balance `ne_bai_mask` (indices that appended) | `{1,2,3}` | `{1,2,3}` |
| equal-row address (bytes 0..15) | `0x2adc25665018aa1fe0e6bc666dac8fc2…` | same |

`0x2adc2566…` is the block's coinbase. The mask moving from `{1,3}` to `{3}` is the #10807 fix
removing the bai-1 loss (the sender's first transaction) and leaving the bai-3 loss in place.

The block has three transactions from a **single** sender, and the coinbase carries a fee-credit row
at each of bai 1, 2 and 3. Rows at bai 1 and 2 append; the bai-3 row compares equal and is dropped.

## Why this is not diagnosed further

Deliberately not theorised. The evidence is **one observation of one account on one fixture**, and
the `_eq_val_*` / `_eq_addr_*` cells are **last-write-wins** across compare-equal rows, so on the
pre-fix reading (two lost rows) the *other* lost row was never directly observed — it was inferred
from the arithmetic. Post-fix there is only one loss, so the coinbase identification is unambiguous,
but *why* the final coinbase credit resolves equal is not established.

[...]

PR #10822, opened 2026-07-28
fix: reverse the BE32 balance into LE scratch before the RLP scalar pair (every balance row was a 33-byte string)

### LINKED ISSUE #10820: bal: balance_changes.post is BE32 but handed to the LE-only scalar pair, so every balance row is emitted as a 32-byte string (predicted overshoot matches measured to 0-3 bytes)

pirapira:
## What

`bal_rlp_scalar_len` / `bal_rlp_emit_scalar` are documented **and implemented** for **LE limbs** — the
length function scans **down from byte 31** for the most significant byte. `balance_changes.post` reaches
them as **BE32**, so on a 12-byte value right-aligned in bytes 20..31 byte 31 is nonzero and it reported
**32** significant bytes. **Every balance row encoded as a 33-byte string instead of its minimal form.**

Closes #10820.

### This is the second instance of a class already closed for one field

The guards in `BlockAccessListBuilder` say it for the slot leg verbatim:

> BOTH readers of the slot must go through the reversal. The row's slot is BE32 and the scalar pair reads
> LE limbs, so a direct `row+32` hand-off is the defect this fixes — and it is invisible to any probe that
> seeds rows in the reader's convention.

Someone found this bug, fixed it for the slot, and wrote **negative guards** to keep it fixed. The balance
leg had the identical shape and no guard.

### Why the field is BE32 — traced, not inferred from the `_be` suffix

`u256AddBe_prog` (`U256.lean:59`) propagates carry from byte **31 down to byte 0**, so byte 0 is most
significant. The value is then copied **verbatim** at every hop: `record_nonstorage_effect` →
`account_write_record` → `bal_builder_append_balance` → the scalar pair, with no conversion — unlike the
slot.

The field table at `BlockAccessListBuilder.lean:192`, labelled *THE authoritative byte-order
declaration*, said **LE**. The balance sits beside the genuinely-LE storage value and was grouped by row
position rather than by provenance. **The table and its summary sentence are corrected here**, with the
rule stated: judge a field by its producer, not by its row position.

## The fix

Mirrors `bal_serializer_slot_to_le`: a reversal into a dedicated `bal_serializer_balance_le`, called once
in measure and once in emit, with **both consuming the same buffer** so the length prefix and the payload
cannot diverge. Deliberately **not** a BE variant of the scalar pair — that routine's own docstring argues
a second implementation of the canonical-minimal-length rule "can agree with the first by construction
rather than by correctness", which is exactly how the original prose-agreed/bytes-disagreed bug survived.

Seven guards added mirroring the slot leg's: positive for the reversal in both paths, negative for the raw
`row+32` hand-off, and one that measure and emit consume the same buffer.

**A pre-existing guard caught this fix.** `BlockAccessListBuilder.lean:782` pinned index-before-value
order keyed on the raw `addi a1, t3, 32` balance emit — i.e. it was pinning the defect — so it failed the
moment the hand-off was removed. Re-pinned to the corrected form rather than relaxed.

## Audit: all call sites of the scalar pair, per requirement

39 real call sites (two apparent ones are inside `#guard` strings), classified by the limb order of the
field handed in:

| route | fields | verdict |
|---|---|---|
| `bal_serializer_u64_field` (one `sd` at offset 0) | every `block_access_index`, nonce measure+emit, code measure+emit | **LE, correct** |
| `bal_serializer_slot_le` (reversed) | storage slot, both readers | **correct** |
| `row+64` verbatim | storage value (genuine LE stack word) | **correct** |
| `row+32` verbatim | `storage_reads` slot (LE stack word) | **correct** |
| `row+32` verbatim | **balance post** — `measure_balance`, `emit_balance` ×2 | **the only BE field handed to an LE-only reader** |

Nonce and code are clean because they route through the u64 widener rather than reading a row field
directly. **So this closes the class, not just an instance.**

## Verification: a falsifiable prediction that passed

The fix predicts `rebuilt_len` drops by exactly `Σ (33 − rlp_len(post))` over emitted balance rows.
Merged main `d7232de14` → this branch; guests `a44e88b4464a3974` → `334a5b919d72a463`, both built with
`LAKE_ARTIFACT_CACHE=false` and sha-verified distinct before measuring.

| fixture (relpath under `…/eip7928_block_level_access_lists/`) | drop | predicted | err |
|---|---|---|---|
| `block_access_lists_eip2935/bal_2935_simple.json` | 169 | 162 | +7 |
| `…eip4895/bal_withdrawal_and_value_transfer_same_address.json` | 105 | 103 | +2 |
| `…eip4895/bal_withdrawal_to_coinbase.json` | 91 | 88 | +3 |
| `block_access_lists/bal_multiple_storage_writes_same_slot.json` | 133 | 135 | −2 |
| `block_access_lists/bal_cross_tx_storage_write.json` (reverts) | 101 | 98 | +3 |
| `block_access_lists/bal_cross_tx_storage_write.json` (rewrites) | 102 | 98 | +4 |

`rebuilt − supplied` moves from **+165 / +102 / +73 / +124 / +98 / +98** to **−4 / −3 / −18 / −9 / −3 /
−4** — now slightly *under*, in the direction of the known missing rows. **Every row count on every leg
is unchanged**, which is what makes this a pure length fix rather than a producer change.

## Not claimed

- **No emitted byte has been read.** The evidence is a source trace plus a prediction that could have
  failed and did not. The residual errors (≤7 bytes) are **unexplained**; plausibly unmodelled RLP
  list-header boundary effects as payloads shrink past 55 bytes, but that is a guess.
- **The digest still mismatches on all six** — rows are still missing: #10818 (reads short by exactly one
  everywhere), #10809 and #10808 (the two balance residuals). This is necessary, not sufficient.
- Not claimed to be the last value-level defect; it accounts for the observed byte overshoot to within a
  few bytes.
- **This kills a directional inference now recorded on #10815**: "more bytes ⇒ numerically larger value"
  requires the length function to be correct. It was not, so the sign agreement previously observed across
  six fixtures was coincidental — it agreed only because both known sender defects under-debit.

## Gates

Layout regen at fixpoint (pass 2). `lake build` completed successfully (4783 jobs), which is what verifies
the regenerated guest-PC pins and all seven new `#guard`s. `check-region-map` matches the linked ELF
(`textSizeBytes = 66ee8`, `dataSizeBytes = 5370` **unchanged**, `bssSizeBytes = 1b85cae0`);
`check-forbidden-tactics` clean; file-size guardrail 2153 files all within cap.


pirapira:
## 80-fixture A/B, and one metric in it is misleading — read the second table

Legs: merged main `d7232de14` (guest `a44e88b4464a3974`) vs this branch (`334a5b919d72a463`), differing
**only** by this fix. Both pre-pinned and sha-verified; spike; no rebuild between legs. Measured on this
branch, not on main.

| | pre | post |
|---|---|---|
| `succ == 1` | 75 | 75 |
| succ flipped 1→0 (new false rejects) | — | **0** |
| succ flipped 0→1 (new accepts) | — | **0** |
| fixtures with **any row count changed** on any leg | — | **0** |
| `supplied_len` changed (sanity) | — | **0** |
| **`rebuilt > supplied` (over-length)** | **45** | **0** |
| exact length match | 0 | **1** |
| digest match | 0 | 0 |

**Every over-length is eliminated, on all 45 fixtures that had one, with zero row-count changes
anywhere.** A length fix cannot add or remove rows, and the measurement confirms it did neither.

### The misleading metric, stated plainly

`|rebuilt − supplied|` moved **closer on 36 and further away on 37**. That reads like a wash. It is not a
regression, and the reason is the same one that made me retract the 383/383 figure on #10680:

**two opposite errors were cancelling.** Over-long balance rows were partially offsetting *missing* rows.
Removing the over-length exposes the shortfall at full size. Examples:

| fixture | before | after |
|---|---|---|
| `bal_2930_account_listed_but_untouched` | 355 vs 334 — **over** by 21 | 304 vs 334 — under by 30 |
| `bal_aborted_account_access` | 349 vs 362 — under by 13 *while over-measuring balances* | 305 vs 362 — under by 57 |

So `|distance|` is the wrong statistic for this change, exactly as "monotone row counts" was the wrong
statistic for #10807. The right ones are **over-length 45 → 0** and **row counts unchanged**, both
unambiguous.

That the remaining error is now uniformly *under*-emission is itself the useful outcome: the row deficits
(#10818 reads short by one, #10809 / #10808 the two balance residuals, #10802 the unfed loop) are no
longer masked by a compensating over-measure.

### Encoder model now validated at BYTE level, not just length

The prediction in this PR rested on an independent re-implementation of the EIP-7928 BAL encoding. That
model has since been validated **byte-exactly**: its output for the declared BAL is found **verbatim
inside the guest input file** on the three fixtures checked (`bal_2935_simple` at offset 816,
`bal_withdrawal_and_value_transfer_same_address` at 762, `bal_cross_tx_storage_write` at 880).

So the arithmetic behind this fix rests on an encoder proven correct against real bytes, not only against
one published length.

### Still not claimed

No **guest-emitted** byte has been read. The model is byte-exact against the *supplied* BAL; the guest's
rebuilt bytes have only been compared by length and row count. The digest remains the instrument that
would close that, and it still mismatches on all 80 because rows are still missing.

pirapira:
Reviewed. **Approve and label.** This is the highest-value change in the queue and the implementation is careful in the three places it could have gone wrong.

## The fix matches the established in-tree pattern

`bal_serializer_balance_to_le` is the same six-line reversal as `bal_serializer_slot_to_le`, applied to a different field. Not a second implementation of the encoding rule — **the canonical-minimal-length logic still exists exactly once**, which is precisely what that routine's own docstring argues for. Declining to write a BE variant of the scalar pair was the right call.

## Measure and emit are in lockstep, and that was the real hazard

Verified both sides on the branch:

- **measure** (`balSerializerMeasureBalanceFunction`) — `addi a0, s4, 32; jal ra, bal_serializer_balance_to_le` then `la a0, bal_serializer_balance_le; jal ra, bal_rlp_scalar_rlp_len`
- **emit** (`balSerializerEmitBalanceFunction`) — reverses once, then emits **the same scratch**

If only one side had been corrected, the length prefix and the payload would disagree and produce **malformed RLP with a still-plausible total** — which a length check would not necessarily catch. The docstring says this explicitly, and three new `#guard`s pin it structurally: one on the measure site, one on the emit site, and one asserting the balance emit follows the `u64_field` emit in order. That last one is the good one — it pins the *sequence*, so a future reordering that reintroduced a second reversal or dropped one would fail at elaboration rather than at a digest.

The reasoning for why no second reversal is needed is also correct and worth having in the source: `bal_serializer_u64_to_field` writes a **different** buffer, so the reversed balance survives the intervening BAI emit.

## The separate scratch buffer is justified, not lazy

Using a new `bal_serializer_balance_le` rather than sharing `bal_serializer_slot_le` avoids making the balance and storage legs' **emit order load-bearing**. Sharing would have created a coupling that no test would exercise until someone reordered the legs.

## Root-cause note that should survive into the codebase

The docstring records *why the hand-off looked correct*: the builder row field table called this field LE, because the balance was grouped with the storage **value** (a genuine LE stack word) **by row position rather than by provenance**. The balance is produced by the `u256_*_be` helpers — `u256AddBe_prog` propagates carry from byte 31 down to byte 0, so byte 0 is most significant — and it is copied verbatim at every hop. That is the kind of note that stops the next person re-deriving a four-hop provenance chain, and correcting the field table is part of the fix rather than a cosmetic afterthought.

## Layout

`.data` **unchanged at `0x5370`** — the new scratch is `.bss` (`.zero 32`), so nothing address-pinned in `.data` moved and `bnf_le_a` is untouched. `.text` `0x066ea0 → 0x066ee8` and `.bss` `…cac0 → …cae0` are repinned consistently.

## Evidence

80-fixture A/B with both ELFs already pinned, `succ` 75 on both legs with **zero flips either direction**, and **row counts changed on zero fixtures on every leg** — which is exactly the right control for a *length* fix: unchanged row counts are what make the byte-level movement interpretable rather than confounded. Closing the reporting asymmetry (#10807 had 80 fixtures while this had six) was worth doing unprompted; an inconsistency like that is one nobody would have questioned.

pirapira:
### Audit completeness: the limb-order class is fully closed, not just the balance instance

The audit in this PR covered the **scalar** pair (`bal_rlp_scalar_len` / `bal_rlp_emit_scalar`), which is
where the LE-limb requirement lives. For completeness the other two RLP emitters have now been checked, so
"this closes the class" is verified rather than asserted:

| emitter | limb-order sensitive? | evidence |
|---|---|---|
| `bal_rlp_scalar_len` / `bal_rlp_emit_scalar` | **yes — LE limbs** | scans down from byte 31; the defect fixed here |
| `bal_rlp_emit_bytes` | **no — order-preserving** | absorbs the payload buffer verbatim (`keccak_absorb(ptr, len)`); **zero** decrementing-pointer walks in its body, so no reversal. Correct for byte strings, and correct for the 20-byte address, which the row already holds as BE20 |
| `bal_rlp_emit_address` | n/a | **no live call sites** — both callers are inside `balRlpEncodeSelftestFunction`. The live account path emits the address via `bal_rlp_emit_bytes` with `li a2, 20` (`BalSerializer.lean:936`) |

So the only limb-order-sensitive path in the BAL emitter is the scalar pair, and within it the balance post
was the sole BE field handed in. **The class is closed by this change**, not merely the instance.

### Incidental: `bal_rlp_emit_address` is reachable only from a selftest

Not a defect and not touched here, but worth recording since it surfaced while enumerating: the routine's
only callers are in the encode selftest, and the live path does not use it. Anyone later "fixing" its byte
order would be editing a path that does not run.

pirapira:
**Addendum, tightening the last claim in the comment above from a source inference to a fact about the
binary.**

I wrote that `bal_rlp_emit_address` is "a path that does not run". That was inferred from its call sites
being inside the encode selftest. Checked properly:

- `bal_rlp_encode_selftest` does have two `jal` callers — but both are in `BalSelftestsProbe.lean`
  (`ziskBalSelftestsPrologue`, `ziskBalSelftestsProbeUnit`), i.e. the **`zisk_bal_selftests` probe, a
  separate build unit and separate ELF**.
- **The decisive check: the emitted `stateless_guest.s` contains ZERO calls to `bal_rlp_encode_selftest`.**

So in the guest `bal_rlp_emit_address` is unreachable, and its only reachable home is a probe unit. The
original statement holds, but it now rests on the emitted asm rather than on where the source call sites
appear — which is the standard this codebase asks for, and the check that distinguishes "no live caller"
from "no caller I noticed".

Nothing in this PR depends on it; recorded so the audit table's `n/a` row is backed by a measurement.

PR #10826, opened 2026-07-28
fix: account_writes_undo_push preserves t5/t6, matching what its callers assume

### LINKED ISSUE #10810: bal: account_write_record's hit path relies on account_writes_undo_push not clobbering t5, which its prologue does not promise (latent, not live)
## Not a live bug — a latent one-register hazard, filed because the next edit trips it

`account_write_record` (`EvmAsm/Codegen/Programs/AccountWriteMap.lean`) keeps the target row address
in `t5` and stores every field through it (`32(t5)` balance, `64(t5)` nonce, `112(t5)` valid mask).
Both paths call `account_writes_undo_push` before those stores, and **only one of them re-establishes
`t5` afterwards**:

```
.Lawr_hit:
  mv a5, t4; li a6, 0; jal ra, account_writes_undo_push; j .Lawr_store      # t5 NOT recomputed

.Lawr_append:
  li t2, CAP; bgeu t1, t2, .Lawr_overflow; mv a5, t1; li a6, 1; jal ra, account_writes_undo_push
  la t0, tx_account_writes_count; ld t1, 0(t0); li t3, 0xa2720000; slli t5, t1, 7; add t5, t3, t5
                                                                            # t5 recomputed
```

`t5` is a **caller-saved** register, so the hit path is only correct because
`account_writes_undo_push` happens not to touch it — its body uses `t0`–`t4` exclusively, and it
saves and restores exactly `t0`–`t4`. I checked; **there is no defect today.**

## Why it is worth a issue anyway

The append path's recomputation is evidence that someone already knew the call is not `t5`-safe. The
hit path relies on an unstated property of a callee that its own prologue does not claim: the
save/restore list stops at `t4`, so `t5` is *documented as clobberable* and merely happens to be
untouched.

The failure mode if that changes is silent and severe. A stale `t5` on the hit path sends the
fieldwise stores — including the valid-mask `or` at `112(t5)` — into **a different 128-byte row**, so
one account's balance/nonce lands on another account's record and a mask bit is set on an account
that never had that component written. That is a wrong-BAL-row defect with no trap and no error
code, i.e. exactly the class that only a full `block_access_list_hash` comparison catches.

## Suggested fix (one line, no behaviour change today)

Either recompute `t5` on the hit path as the append path does, or extend
`account_writes_undo_push`'s prologue to save/restore `t5`–`t6` so its contract matches what both
call sites already assume.

## Not claimed

- **Not** claimed to be reachable or to affect any current fixture. Verified by reading that the
  callee uses only `t0`–`t4`; no measurement was taken because there is nothing to measure while the
  property holds.
- Not audited whether other callers of `account_writes_undo_push` (or other routines in this file)
  rely on the same unstated `t5`/`t6` preservation. Only this one call pair was read.

pirapira:
## What

`account_write_record` keeps the target row address in **t5** and stores every field through it —
balance at `32(t5)`, nonce at `64(t5)`, the valid mask at `112(t5)`. Both paths call
`account_writes_undo_push` before those stores, and **only the APPEND path re-establishes t5
afterwards**:

```
.Lawr_hit:
  mv a5, t4; li a6, 0; jal ra, account_writes_undo_push; j .Lawr_store      # t5 held ACROSS the call
.Lawr_append:
  … jal ra, account_writes_undo_push
  la t0, tx_account_writes_count; … slli t5, t1, 7; add t5, t3, t5          # t5 RECOMPUTED
```

t5 is caller-saved and the callee's prologue saved exactly `t0`–`t4`, so **t5 was documented as
clobberable and merely happened to be preserved** because the body uses `t0`–`t4` exclusively
(verified: zero occurrences of t5/t6 in it). The append path's recomputation is evidence someone
already knew the call is not t5-safe.

Closes #10810.

## Why fix the callee

Two options existed: recompute t5 at the hit site, or extend the callee's save/restore. **The callee**,
because it makes the contract match what both call sites already assume and protects every future
caller, instead of leaving the next one to rediscover the hazard. Frame grows 48 → 64 (still 16-byte
aligned). Two guards pin the **save and the restore** — a prologue-only save would leave the register
clobbered on return and still read as a fix.

## This is NOT a live bug, and it cannot be demonstrated by measurement

Today correct-by-contract and correct-by-accident emit identical bytes, so **no fixture distinguishes
them**. Stated plainly so the A/B below is not over-read.

The failure mode if the accident ever ends: a stale t5 sends the fieldwise stores — **including the
valid-mask `or` at `112(t5)`** — into a *different* 128-byte row. One account's balance or nonce written
onto another account's record, and a mask bit set on an account that never had that component written,
with **no trap and no error code**. That is the wrong-row class only a whole-structure hash catches,
which is the argument for not leaving it behind an accidental invariant.

## A/B: nothing moves, which is the expected and only available result

Legs: merged main `d7232de14` (guest `a44e88b4464a3974`) vs this branch (`ce397098ab76c5f9`), same base,
differing only by the two extra saves.

**`OUTPUT[0:264]` byte-identical on 80/80 eip7928 fixtures** — the full published surface compared
byte-for-byte, not by selected counters.

**A green A/B here does not confirm the fix works.** It confirms the fix changed nothing observable,
which is what a latent-hazard fix should do. The confirmation that it *works* is the source-level
argument: the callee now saves what the caller relies on.

> Process note, recorded because it nearly produced a wrong claim: my first run of this A/B compared
> against `scoper-ab/pre`, which is the pre-#10807 shadow-lineage leg, **not** the `d7232de14` baseline —
> and reported 76/80 *differing*. The differences were #10807 plus the read repair, nothing to do with
> this change. Re-run against the correct same-base leg gives 80/80 identical. Naming the baseline in a
> comment is not the same as reading from it.

## Gates

Layout regen at fixpoint (pass 2). `lake build` completed successfully (4783 jobs), which verifies the
regenerated guest-PC pins and both new `#guard`s. `check-region-map` matches the linked ELF
(`textSizeBytes = 66eb0`, `dataSizeBytes = 5370` unchanged, `bssSizeBytes = 1b85cac0`);
`check-forbidden-tactics` clean; file-size guardrail 2153 files all within cap.

Based on `d7232de14`; independent of #10822 (different files — `AccountWriteMap` vs
`BalSerializer`/`BlockAccessListBuilder`), though both regenerate the layout files.


pirapira:
Reviewed. **Approve.** A 2-register save/restore, but the reasoning around it is what makes it worth landing rather than deferring.

## The evidence for the hazard is better than the hazard's symptoms

The observation I'd single out: **`account_write_record`'s append path recomputes `t5` after this call while its hit path does not.** The hit path holds the target row address in `t5` *across* the call and then stores every field through it — balance at `32(t5)`, nonce at `64(t5)`, the valid mask at `112(t5)` — without re-establishing it.

That asymmetry is **evidence that someone already knew this call is not `t5`-safe**, recorded in one path and not the other. The codebase contained the knowledge; it just wasn't in the contract. That is a much stronger argument than "a register might get clobbered."

And the prologue promised only `t0`–`t4`, so `t5` was **documented as clobberable and merely accidentally preserved** — the body happens to use `t0`–`t4` exclusively. Removing an accidental correctness is worth a PR even when nothing is broken today.

## The failure mode is precisely characterised, and it is tonight's recurring class

A stale `t5` sends the fieldwise stores — **including the valid-mask `or` at `112(t5)`** — into a *different* 128-byte row: one account's balance or nonce written onto another account's record, and a mask bit set on an account that never had that component written, **with no trap and no error code**.

That is a wrong-row defect with correct row *counts*, which no counter can see. It joins the two confirmed count-correct/value-wrong defects from today (#10820 balance over-measure, #10807 wrong pre operand) as something only a whole-structure hash catches. The docstring says exactly that, and it is the right place for it.

## Both guards are needed and the reason is stated

```
#guard (accountWritesUndoPushFunction.splitOn "sd t5, 40(sp); sd t6, 48(sp)").length == 2
#guard (accountWritesUndoPushFunction.splitOn "ld t5, 40(sp); ld t6, 48(sp)").length == 2
```

Pinning the **save and the restore separately**, because *"a prologue-only save would leave the register clobbered on return and read as a fix"* — same lockstep reasoning as the measure/emit pair in #10822, and the same class of half-fix that would pass a casual review.

## Two smaller choices, both right

- **Fixing the callee rather than recomputing `t5` at the one call site** protects every future caller instead of leaving the next one to rediscover the hazard — and it makes the callee's contract match what both existing call sites already assume, so it removes a latent disagreement rather than patching an instance.
- **Saving `t6` as well**, which the body does not need either, makes the contract uniformly "preserves `t0`–`t6`" instead of "preserves `t0`–`t4`, plus `t5` because one caller happens to need it." A contract with an exception in it is one someone will read past.

Frame 48 → 64 for seven saved registers, still 16-byte aligned.

## The A/B result is the *correct* one, and that matters here

A latent-hazard fix that changes no behaviour today **should** show zero differences, and a non-zero result would have been the alarm rather than the reassurance. Worth restating the near-miss that established this: an earlier run of this same A/B read 76-of-80 differing off a wrong baseline, which would have reported a two-register save/restore as altering behaviour on 95% of the corpus — absurd on its face, and that absurdity was the only thing that prompted the re-check. **The remedy adopted from it — echo the variable that is then read, never a literal alongside it — is worth more than this PR.**

`CONFLICTING` on the current main after #10821 landed; generated-file resolution is merge-bot's, not to be resolved by pushing here.

PR #10828, opened 2026-07-28
fix: record modeled EIP-4788 storage reads

### LINKED ISSUE #10825: BAL misses EIP-4788 net-equal root-slot storage read
## Defect

`bal_2935_simple` declares 17 `storageReads`, while the existing guest probe
emits 16.  The missing fixture entry is the begin-of-block EIP-4788 beacon-root
ring slot, not one of the four post-block request predeploys.

## Direct fixture and emitted-artifact evidence

The generated v0.6.2 fixture is:

`tests-zkevm@v0.6.2/.../block_access_lists_eip2935/bal_2935_simple.json`.

Its BAL contains:

- four request predeploy accounts, each reading slots `0x00` through `0x03`:
  16 reads;
- EIP-4788 (`0x000f3d...beac02`) with `storageChanges` at slot `0x0c` and a
  separate `storageReads` entry at slot `0x200b`: one more read.

The fixture header timestamp is `0x0c`.  EIP-4788 uses
`root_index(timestamp) = timestamp % 8191 + 8191`; therefore
`12 + 8191 = 8203 = 0x200b` exactly.

The fixture also has zero `parentBeaconBlockRoot` and empty EIP-4788 pre-state
storage.  Thus this system call writes zero back to an already-empty root-ring
slot.  An existing focused artifact independently reports the same deficit:
`/var/tmp/evm-asm/system-mode-main/2935-candidate/results.tsv` has row `00537`
as `0 1 1,16`.  Nearby rows `00528` through `00535` alternate 16 and 17, so
the probe is discriminating per fixture rather than returning a constant.

## Spec mechanism

Amsterdam processes the EIP-4788 system transaction before user transactions
in `fork.py` (`apply_body`, lines 897-901).  Its `SSTORE` gets the current value
before storing (`vm/instructions/storage.py`, `get_storage` around line 108,
then `set_storage` around line 145).  `get_storage` records the read
(`state_tracker.py:295`).

The BAL builder later drops a read only when that slot has a storage-change row
(`block_access_lists.py:549-552`).  The zero-over-zero root-ring write has no
net change row, so its read survives and must appear in the declared BAL.

This makes the visibility condition important: a missing pre-write read is
normally hidden when the write changes the slot, because the BAL filter would
discard the read.  It becomes observable precisely for a net-equal write such
as this zero-over-zero EIP-4788 case.

## Guest-side status and next investigation

The guest's modeled startup path currently appends EIP-2935/EIP-4788 storage
*tuple rows* through `append_modeled_system_storage_tuple_rows`
(`BlockVerdictSystemStorageCapture.lean:124-129`), including the root slot, but
this audit has not yet established whether the guest fails to record the
pre-write read at all or records it and loses it later.  That distinction is
intentionally open.

Do not implement a general repair by checking only
`swd_4788_root_vlen == 0`: it identifies this zero case but not every nonzero
pre==post write.  A correct repair needs to follow the actual pre/post
comparison or be explicitly scoped.

Current `docs/test_cases.md` and the checked source test are incomplete for
this generated fixture: they omit its active EIP-4788 and request-predeploy
paths.  The generated fixture and the emitted row above are the evidence for
this report.

pirapira:
Fixes #10825.

## Change

After a successful modeled EIP-4788 beacon-root transition, record its two
SSTORE keys directly in the block-lifetime storage-read set.  The helper reuses
the established execution-address conversion and `storage_read_record_block`,
so it neither creates a transaction-local read set nor crosses a transaction
boundary that would clear it.

This makes the EIP-4788 parent-root SSTORE visible to the BAL builder even when
the write is net-equal.  The read is required by execution-specs; it is not
represented by a `storage_changes` row in that case.

## Layout / regeneration

Two complete serialized relinks converged at:

- `.text = 0x670e8`, `.data = 0x5370`, `.bss = 0x1b85cac0`
- candidate ELF SHA-256: `e2fe043a8899c14c7900b9e8150432b1e1f59ea98794f44c4dcda589cc1fd277`
- `GuestAddrs.bnf_le_a` remains `0xa3000ee0`

`scripts/check-asm-to-program.sh` and `scripts/check-region-map.sh` passed.

## Paired Spike tier

Immutable snapshots, same copied runner and same captured seed on both legs:

- parent source: `89e73e78`, ELF SHA-256 `47792f733aa0b4949ac3e8e5de32a95c70c45aa09a236d3056d60d6148a00192`
- candidate source: `54dff809`, ELF SHA-256 `e2fe043a8899c14c7900b9e8150432b1e1f59ea98794f44c4dcda589cc1fd277`
- runner SHA-256: `c6ac09fa72c2103b6d3e390b779b022ef7f220e5a39bb5f667114a89d78e5868`
- seed: `4476710578111144899`
- denominator: 226 = random 200 + all 22 `bal_4788` fixtures + four controls

Both legs: `FA=0 FR=4 FAULT=0 OK=222`; transitions are exactly `222 OK→OK`,
`4 FR→FR`, and zero in every other direction.  Verdict, failure code,
recomputed root, emitted storage-change count, and all non-diagnostic output
fields are identical for all 226 rows.

The headline control `00537_test_bal_2935_simple...` changes the serializer's
emitted storage-read count at output+152 from **16 to 17**, matching the
fixture's 17 declared reads, with `succ=1` and an identical root.  Its raw
`storage_reads_count` at output+184 changes 16 to 18: this is expected, because
both SSTORE reads are recorded, while the timestamp read is excluded later by
the builder's storage-change suppression (`block_access_lists.py:549-552`).
The raw count therefore is not the declared BAL read count.

All 22 BAL EIP-4788 cases retain their verdicts.  The absent-contract control
is fully unchanged.  Empty-block and invalid-calldata controls have no emitted
storage-read increase (their raw staging count may rise by two, then is
suppressed at build time), so they emit no extra BAL row.

The four controls include `bal_2935_simple`,
`bal_multiple_storage_writes_same_slot`, withdrawal/value-transfer, and
withdrawal-to-coinbase.  The same-slot control remains verdict/root-identical.

No full-corpus leg was run: this is the instructed focused gate over the
affected EIP-4788 population plus a random regression guard.

## Follow-up composition prediction

`bal_2935_simple` also carries the separately-fixed BE32-to-LE balance-row
over-measure from #10822.  A future tree containing both changes therefore
closes the two identified defects on a fixture that carries balance rows.  It
is **not** a promise of a digest match: the byte-exact encoder model identifies
this fixture as one of two with an unexplained one-byte residual despite a
matching modeled length, and another modeled fixture has demonstrated a
content/order hash mismatch with no count or length signature.  That remaining
work is outside this PR.


pirapira:
Reviewed. **Approve and label.** The staging is right, the evidence is right, and the register discipline — which is where the sibling PR went wrong — is clean here.

## Placement and gating

`jal ra, bsr_beacon_change; bnez a0, .Lbsr_cons_sys4788` then the new call, so the helper runs **only when the modelled beacon call succeeded**; failure branches away before it. That is what keeps `absent_contract`, `empty_block` and `invalid_calldata_size` emitting nothing, and those are the only fixtures that can detect over-emission — the query variants can only show a count going up.

## Register discipline is correct, and I checked it specifically

The `t0` transparency defect in #10821 was in a two-line dispatch, so I looked for the same shape:

- **No temporary is held across an inner call.** After `bal_addr_to_exec_log_key` the routine does a fresh `la t0, swd_4788_slot`; after `storage_read_record_block` a fresh `la t0, swd_4788_root_slot`. Nothing survives a call boundary.
- **No caller temporary is live across the new call.** The call site is followed by `addi s1, s1, 1` (an s-register, and the block recorder preserves its own), and `.Lbsr_skip_4788` immediately re-loads with `la t0, bsr_changed_account_count`. There was already a `jal` at that point, so the caller could not have been relying on temporaries there anyway.
- Frame `-16` saving only `ra` is therefore right, not thin.

## Calling the block recorder directly was the correct choice

Avoiding `storage_read_record` + `read_sets_incorporate_tx` removes a real hazard rather than a stylistic one: `read_sets_incorporate_tx` tail-jumps to `read_sets_discard_tx`, so it **merges and then clears** the tx cursors — correct at a transaction boundary, and this fires from a post-transaction phase where I could not establish the tx read sets are empty. It also removes an inherited dependency on `system_call_mode`, which after #10821 routes `storage_read_record` differently.

- **The buffer-reuse liveness is asserted, not enforced.** `bsr_kbuf` and `bsr_delta` being dead after `bsr_beacon_change` returns is stated in the docstring and is what avoids a new `.data` allocation (good — `.data` growth breaks address-pinned proofs in unrelated modules). But nothing pins that property, so a future change leaving live data in `bsr_delta` would corrupt these keys silently. A `#guard` cannot express liveness; a one-line comment at `bsr_beacon_change`'s exit saying those buffers are free for reuse would at least put the constraint where someone editing it will see it.
- **The 32-byte reversal loop appears twice in the same routine.** Duplicating a byte reversal is not the duplication `bal_serializer_slot_to_le` argues against — that argument is about canonical-minimal-length logic — but two copies inside one function is more clearly worth factoring than two copies in separate modules.

## Evidence

226/226 joined at a captured seed with both legs `FA=0 FR=4 FAULT=0 OK=222` and verdict, fail-code, root and emit-change identical on all 226. Reporting **both** counters with the reason is what makes this readable: emitted `output+152` 16 → 17 against declared 17 is the result, while raw `output+184` 16 → 18 is expected and *not* over-emission, because the spec applies its exclusion at build time (`block_access_lists.py:549-552`) rather than at record time. Without that, the next reader compares raw 18 against declared 17 and files a defect.

Stating the #10822 composition **with its qualifier** is the right call — this fixture is one of the two carrying an unexplained one-byte residual, so the combination closes the two *identified* defects without promising a match. And there is now a measured content-or-ordering defect with zero length and zero count signature on a fixture where the modelled length matched exactly, so a residual mismatch here would be a property of the corpus rather than of this change.

Issue #10817, opened 2026-07-28
Proof + unit-test package: bal_canonical_sort ordering and permutation (self-contained)


pirapira:
## Summary

`bal_canonical_sort` (`EvmAsm/Codegen/Programs/BalCanonicalSort.lean`, 536 lines) is the canonical-ordering step for the EIP-7928 Block Access List. It is **live** — called six times inside `bal_serializer_rebuild_hash` — and has **zero theorems and no Spec module anywhere in the tree**.

It is unusually well suited to an external contributor, for reasons that are properties of *this* module rather than general encouragement.

## Why this one is a good self-contained package

| property | detail |
|---|---|
| **Size** | one module, 536 lines; no cross-module proof dependencies to inherit |
| **Existing proof state** | **none** — 0 theorems, no `*Spec.lean`. Nothing half-finished to retrofit into |
| **Contract source** | given by execution-specs, not reverse-engineered (see below) |
| **Test entry point** | `bal_canonical_sort_selftest` already exists and is wired into `BalSelftestsProbe` — cases can be added without new harness infrastructure |
| **Input shape** | already pinned as documented constants (row strides and key segments) |
| **Load-bearing?** | yes — the sort runs on the live rebuild path, so proofs about it are not about dead code |
| **Prospect of churn** | **low**, with one caveat — see below |

## The contract, from the spec

`build_block_access_list` (`block_access_lists.py:524-533`, submodule pinned at `e5a8caf1b` = `tests-zkevm@v0.6.2`):

> The resulting list is ordered by:
> 1. Account addresses (lexicographically)
> 2. Within each account:
>    - Storage slots (lexicographically)
>    - Transaction indices (numerically) for each change type
>
> Addresses, storage slots, and block access indices are unique.

The guest's input rows, keys listed most- to least-significant (`BalCanonicalSort.lean`, "key descriptors" section):

| list | stride | segments |
|---|---|---|
| storage change | 96 | address `(0,20)` BE, slot `(32,32)` BE, index `(24,8)` LE |
| balance change | 64 | address `(0,20)` BE, index `(24,8)` LE |
| nonce change | 40 | address `(0,20)` BE, index `(24,8)` LE |
| code change | 64 | address `(0,20)` BE, index `(24,8)` LE |

## Obligations — and **permutation is the headline, not sortedness**

[...]

PR #10829, opened 2026-07-28
diag: publish bal_serializer_rebuilt_hash (a step toward the #10680 comparator)

pirapira:
## What

Five stores, in the existing epilogue publish block, of a value the guest **already computes**:
`bal_serializer_rebuilt_hash` (32 bytes) plus 16 bytes of `bal_serializer_supplied_hash` as a control,
at `OUTPUT+344..384`.

This is a **step toward the #10680 comparator, not transitory diagnostic state** — it is the same value
the wired check consumes. `bal_serializer_verify` already rebuilds into `bal_serializer_rebuilt_hash`,
hashes the supplied BAL into `bal_serializer_supplied_hash`, compares four dwords and returns 0/1/2.

## Why publish a hash when the 0/1/2 verdict is already published

The existing verdict compares the guest against the **supplied** BAL, which is known to mismatch
everywhere. Comparing the guest's hash against `keccak` of an **independent model's** expected bytes
separates a **wrong guest** from a **wrong model** — an ambiguity that no length or row-count comparison
can resolve, because equal lengths with different content is exactly the coincidence that produced the
retracted `383/383` reading on #10680.

The rebuilt bytes themselves are **not observable**: the rebuild is `keccak_init` → `emit_outer` →
`keccak_final` and is never materialised, so a byte-offset dump would need a tee across **eleven**
`keccak_absorb` sites. The hash is the cheapest faithful window onto that stream.

## What it already produced

Guest `263f5d03ecc9758744d57f33e6e321b62fe89cafb180a0db1c00d7e2e6a138e9`, spike, sha-verified distinct
before reading. Encoder model validated **byte-exactly** (its output for the declared BAL appears verbatim
in the guest input) and keccak checked against `keccak256("") = c5d24601…`.

**Control — passes on 6/6:** the published supplied hash equals `keccak(BAL bytes extracted from the guest
input)`. This also establishes a fact worth stating on its own: **the declared BAL order is already
canonical**, since a fixture-order reconstruction reproduces the declared bytes exactly.

**Result 1 — the model is byte-faithful on one fixture.** On
`…eip4895/bal_withdrawal_to_coinbase.json`, `keccak(model) == guest rebuilt hash` once the beacon-roots
read (#10818) and **one** balance row are removed. The match is **unique among four candidate rows**, and
it names `0xf6c3a9edc1…` at `bai 1` — the **tx sender**, the same row the in-guest address-witness cell
identified independently (#10809). Two unrelated instruments agree.

**Result 2 — a defect with zero length and zero count signature.** On
`…eip4895/bal_withdrawal_and_value_transfer_same_address.json`: declared balance rows 4, emitted 4 (so the
model is unambiguous, no search), model length 347 == `rebuilt_len` 347, and the hashes **differ**
(`204b0feb…` vs `4c6f546a…`). No counter and no length comparison can see this. It is a measured instance
of the class that motivates #10680.

**Narrowing on that fixture — two exhaustive exonerations:**

| hypothesis | scope tested | result |
|---|---|---|
| account ordering | **all 362,880** permutations of the 9 accounts | **no match — excluded** |
| `AccountChanges` field order | **all 120** permutations of the 5 fields | **no match — excluded** |
| reads reversed / balances reversed / both | 3 targeted transforms | no match |
| one balance row emitted as that account's **pre**-state balance | every balance row in turn | no match |

So the defect is **content**, not ordering at any level reachable this way — and the obvious content
hypothesis (the #10809 pre-for-post shape, which preserves length because pre and post are both 12 bytes
here) is **dead**.

## Not claimed

- **Which field differs is not established.** Four hypothesis families were tested and none matched;
  the remaining space is per-field values across nine accounts, where a wider search would be **fitting
  rather than identifying**. Localising it needs a byte-offset diff, i.e. materialised bytes.
- The other four fixtures also differ and nothing is read into them: two carry an unexplained `+1` byte
  residual, two have multi-row ambiguity.
- This does **not** show the sort is broken — on the one fixture testable, account ordering is
  *exonerated*. Relevant to #10817: permutation-plus-sortedness remains the right obligation, but this
  result constrains it rather than confirming it.

## Gates

Layout regen at **fixpoint (pass 2)**. `lake build` completed successfully (4783 jobs), verifying the
regenerated guest-PC pins. `check-region-map` matches the linked ELF (`textSizeBytes = 66f90`,
`dataSizeBytes = 5370` **unchanged**, `bssSizeBytes = 1b85cae0`); `check-forbidden-tactics` clean;
file-size guardrail 2153 files within cap.

> Note on process: an earlier build of this branch deliberately skipped regen steps 2–3 (only an ELF was
> needed to measure), which left `GuestAddrs` stale and failed `lake build` on the pinned-PC `decide`. The
> measurements never depended on `GuestAddrs`, but a landable PR does — found by gating it rather than
> assuming.

Based on #10822 (`fix/bal-balance-le-serialization`), so it measures the fixed serializer.


pirapira:
## 80-fixture A/B: the publish disturbs nothing, and it demonstrably fires

Legs differ **only** by the five stores in this PR: guest `334a5b919d72a463` (#10822) vs
`263f5d03ecc97587` (this branch). Both pre-pinned, sha-verified distinct; spike; no rebuild between legs.
Measured on this branch, not on main.

| | result |
|---|---|
| `OUTPUT[0:344]` byte-identical — everything the publish must not disturb | **80 / 80** |
| fixtures differing anywhere below the new window | **0** |
| fixtures where the new window `OUTPUT[344:400]` is non-zero — i.e. the publish **fired** | **76 / 80** |

Both halves matter. Byte-identity alone would be satisfied by a publish that never executed; the 76/80
non-zero reading is the positive control that rules that out.

### The four all-zero cases are explained, not unexplained nulls

| fixture | succ | `shadow_status` | `supplied_len` |
|---|---|---|---|
| `bal_gas_limit_boundary` ×4 | 0 | **3** | 0 |

`shadow_status = 3` is "skipped on an already-rejected input", and `supplied_len = 0` means no BAL was
decoded. So on those four the shadow rebuild never ran and there is nothing for the publish to report —
the zero is the correct reading rather than a silent failure. Consistent with their `succ = 0`.

### Why this A/B was worth running

A five-store publish "obviously" changes nothing, which is exactly the kind of claim worth measuring:
`OUTPUT+344..384` had to be free of every other writer, and the earlier inventory of OUTPUT-base stores
established that only `block_verdict_creation_runtime` writes above 336 (at 472/480). This confirms it on
the corpus rather than from the inventory alone.

pirapira:
**On the absent merge label — deliberate sequencing, not a quality judgement.**

Under the maintainer's ruling to accept the content residual (recorded in #10830), the diagnostic value of
this publish is already banked: the findings it produced are written up and no further diagnosis is
planned. Its *remaining* value is realised only **when the #10680 comparator is wired**, since this is the
value that comparator consumes.

So this PR sits ready and is labelled the moment wiring is sequenced. It is not held back for any concern
about the change itself — gates are green at fixpoint, and the corpus A/B below shows it disturbs nothing
while demonstrably firing.

pirapira:
**Status: verified, deliberately unlabelled pending comparator sequencing, no further work required.**

The corpus A/B above is banked, so when the #10680 wiring is sequenced the label costs nothing extra and
there is no verification left to run on this branch.

pirapira:
Reviewed the code, separately from the labelling decision — the label is held on sequencing grounds (the publish's value is realised when the comparator is wired, which is gated), not on quality. Two findings, the first of which will cause a silent misread if left.

## ⚠️ These offsets exceed the documented dump length, and a stale reader sees zeros

The new stores land at OUTPUT **+344 through +384**. The harness writes **256** OUTPUT bytes by default (offsets 0–248), and the `bald_*` cells already required `SPIKE_OUTPUT_LEN=336` to be visible. **These need ≥ 392.**

That matters because of *how* it fails: a reader using the previous length gets **zeros**, and a zero hash is indistinguishable from "the hash is zero" — there is no truncation error, no short read, nothing to notice. That is precisely the never-ran-versus-real-zero ambiguity the `bald_*` cells were careful to avoid by publishing known-nonzero counters alongside.

Please state the required `SPIKE_OUTPUT_LEN` in the comment at the publish site. It was found last time by noticing the output file was exactly 256 bytes rather than by reading a zero and believing it — that catch should not have to be repeated.

## The rebuilt/supplied publication is asymmetric — 32 bytes vs 16

```
344, 352, 360, 368  <- bal_serializer_rebuilt_hash   (4 dwords = 32 bytes, full)
376, 384            <- bal_serializer_supplied_hash  (2 dwords = 16 bytes, HALF)
```

If truncating the supplied side to 128 bits is deliberate — and it is ample for the control, since a collision on 16 bytes is not a practical concern — then say so, because a reader expecting a full 32-byte hash at `+376` will read 16 bytes of hash followed by 16 bytes of whatever is next and get a wrong value with no indication. If it is an oversight, the two missing dwords are cheap. Either resolution is fine; the ambiguity is not.

## What is right

- **The value already exists** — this publishes what `bal_serializer_verify` computes today, so it adds no computation and cannot perturb the rebuild.
- **The docstring justifies the publish against the obvious objection** — the existing 0/1/2 verdict compares guest against **supplied**, which is known to mismatch, whereas comparing guest against an independent model's expected bytes separates a **wrong guest** from a **wrong model**. That ambiguity is real and byte-length agreement cannot resolve it, as the retracted 383/383 reading demonstrated.
- **It records why the bytes themselves are not published** — the rebuild streams into a keccak context and is never materialised, so a byte-offset dump needs a tee across eleven `keccak_absorb` sites. Having that in the source stops the next person proposing the dump, which has already been proposed twice.
- **The `npr` save/restore does not interfere**: that loop bounds at `li t3, 112`, so it restores OUTPUT[0:112] only and cannot clobber 344+. And it runs *after* these stores, so the ordering is safe in both directions.
- **Verification is banked** — OUTPUT[0:344] byte-identical across the corpus, which is the correct control for a publish-only change: it proves the stores perturbed nothing outside the slots they write.

## Housekeeping

`RlpSpliceHelperSpec.lean` carries a one-line pin change because this branch predates #10813. Both `decide` pins are **deleted** on main now, so that line disappears on the forward merge rather than needing to be maintained.

pirapira:
## The premise in this PR's title is now stale: the comparator no longer depends on it

(Correcting my own framing above — **10829 is a PR, not an issue**, so the decision here is land-or-close,
not re-title. The substance stands.)

This was opened as *"a step toward the #10680 comparator."* #10835 wires that comparator **without** it —
it binds `bal_serializer_verify`'s existing return value directly, so publishing
`bal_serializer_rebuilt_hash` is not on the path. The sequencing question that kept this deliberately
unlabelled is resolved: **nothing is waiting on it.**

**Not obsolete, but its value is now purely diagnostic and narrower than "a step toward."** The rebuild
streams straight into keccak and is never materialised, so the published digest is the only observable
the rebuild has; earlier in this investigation the published hash served as an oracle by inverting it
against candidate row sets to *name* a missing row. That is a real capability and this PR enables it. But
it localises nothing on its own and is now a prerequisite for nothing.

**State:** OPEN, **CONFLICTING**, unlabelled for merge, branch `diag/publish-rebuilt-hash`. So landing it
costs a conflict resolution for a diagnostic-only capability.

**Recommendation: close it** unless someone has a hash-inversion diagnosis in flight. I do not, and I
would not spend a conflict resolution on it. Not closing unilaterally — flagging for whoever owns the
call.

pirapira:
Closing this diagnostic PR rather than repeatedly regenerating its large, conflict-prone layout diff.

The experiment was useful: it established that the `all-agree` class is real — a rebuilt stream can have the same row count and byte length as the declared BAL while its whole-stream hash differs. That result is retained here; this is not a claim that the hash was an invalid instrument.

It is not the right persistent instrument for the next step. The landed shadow comparator already publishes the match status and both lengths, and a whole rebuilt hash can add only a binary model-vs-guest result. It has no locality: once it differs, it does not identify the account, category, field, or slot that a repair should target. Maintaining this XL generated-layout PR therefore costs repeated merge/regen work without advancing the diagnosis of the remaining all-agree cases.

The follow-up needed for that class is a locality instrument — per-entry or per-category running digests, or the address/slot-count-style dump where that is the relevant boundary — not another global hash publication. If a future investigation specifically needs the binary comparison against an independent model again, it should add a small targeted diagnostic on the then-current layout rather than revive this stale generated diff.

PR #10831, opened 2026-07-28
refactor(codegen): parameterise U256 over GuestLayout (GH #10753, third conversion)

pirapira:
Third leaf/bridge conversion after HashBridge (#10800) and BloomAddValue (#10812). All 11 U256 routines (add/lt/sub/from_u64/is_zero/min/max/div_u64/eq/mul_u64/to_u64 BE family) move to leaf `U256Prog.lean`; bridge `U256.lean` (manifest path, module name unchanged) re-exposes every program under its original name and type. Probe BuildUnits stay on the bridge.

**GuestLayout field total: 12** (10 + 2 new: `u256m_acc`, `u256_mul_u64_be`).

**Post-split line counts** (hard cap 1500 under `Codegen/Programs`): leaf `U256Prog.lean` = 818 lines, bridge `U256.lean` = 350 lines (pre-split: 1131).

## Acceptance (measured on this branch, origin/main 89e73e78b)

| Gate | Result |
|------|--------|
| `check-file` (11 functions, render + layout drift) | CLEAN (11 defs) |
| `check-all` | CLEAN (390 defs across 148 files) |
| Byte-identity oracle — RE-RUN, not inherited | `--emit-lean` regenerates `GuestImageEntries.lean` **byte-identical** — third independent test of the #10803 fix, first at 11-Function scale |
| Flip test (`u256m_acc` +8 in GuestAddrs.lean) | `U256Prog.olean` sha256 **unchanged** (9d940520…) |
| Flip-test bridge rebuild | fails at `U256MulU64Be/OuterLoop.lean:77` `accBase_toNat` — the pre-existing live decide pin of the data-bss class (explicitly excluded from #10790); the pin rejecting the flip loudly is its designed behavior, and the leaf insulation (the acceptance property) held |
| `check-guest-addrs` after restore | CLEAN (1125 symbols) |
| `lake build codegen` | clean (4280 jobs) |

Measured on this branch per the cadence rule; the flip was applied and reverted locally.

### DIFF
--- a/EvmAsm/Codegen/GuestLayout.lean
+++ b/EvmAsm/Codegen/GuestLayout.lean
@@ -48,6 +48,9 @@ structure GuestLayout where
   -- BloomAddValue (bloom_add_value); `zkvm_keccak256` already above.
   bav_hash : Nat
   bloom_add_value : Nat
+  -- U256 (u256_mul_u64_be: accumulator window and entry).
+  u256m_acc : Nat
+  u256_mul_u64_be : Nat
 
 /-- The all-zero layout: emission/guard view, never linked against. -/
 def GuestLayout.zero : GuestLayout :=
@@ -60,6 +63,8 @@ def GuestLayout.zero : GuestLayout :=
     zkvm_keccak256 := 0
     zkvm_keccak256_segments := 0
     bav_hash := 0
-    bloom_add_value := 0 }
+    bloom_add_value := 0
+    u256m_acc := 0
+    u256_mul_u64_be := 0 }
 
 end EvmAsm.Codegen
--- a/EvmAsm/Codegen/GuestLayoutInstance.lean
+++ b/EvmAsm/Codegen/GuestLayoutInstance.lean
@@ -31,6 +31,8 @@ def guestLayout : GuestLayout :=
     zkvm_keccak256 := GuestAddrs.zkvm_keccak256
     zkvm_keccak256_segments := GuestAddrs.zkvm_keccak256_segments
     bav_hash := GuestAddrs.bav_hash
-    bloom_add_value := GuestAddrs.bloom_add_value }
+    bloom_add_value := GuestAddrs.bloom_add_value
+    u256m_acc := GuestAddrs.u256m_acc
+    u256_mul_u64_be := GuestAddrs.u256_mul_u64_be }
 
 end EvmAsm.Codegen
--- a/EvmAsm/Codegen/Programs/U256.lean
+++ b/EvmAsm/Codegen/Programs/U256.lean
@@ -19,77 +19,46 @@
 
   Lives standalone so Tx / Header / Block / Mpt consumers can
   import the u256 arithmetic family without pulling the full Tx module.
+
+  GH #10753: BRIDGE of the U256 module split.  The Programs live in the leaf
+  `EvmAsm.Codegen.Programs.U256Prog` parameterised over `GuestLayout`; this
+  module re-exposes each one under its ORIGINAL name and type
+  (`def u256Xxx_prog : Program := u256Xxx_prog_of guestLayout`) so every
+  consumer compiles untouched — the bridge name is also REQUIRED by the
+  concrete-render gate (key `emitProgram <camel(entry)>_prog`).  The probe
+  `BuildUnit`s stay here with the applied programs.
 -/
 
-import EvmAsm.Rv64.Program
 import EvmAsm.Codegen.Layout
-import EvmAsm.Codegen.Emit
-import EvmAsm.Codegen.GuestAddrs
-import EvmAsm.Codegen.AsmReloc
+import EvmAsm.Codegen.Programs.U256Prog
+import EvmAsm.Codegen.GuestLayoutInstance
 
 namespace EvmAsm.Codegen
 
 open EvmAsm.Rv64
 
-/-! ## u256_add_be -- PR-K51 modular addition on BE u256 buffers
+def u256AddBe_prog : Program := u256AddBe_prog_of guestLayout
 
-    Compute `(a + b) mod 2^256` over two 32-byte big-endian
-    `u256` buffers, storing the result in `out` and returning a
-    0/1 overflow flag (`1` ⇔ unsigned overflow ⇔ `a + b >= 2^256`).
+def u256LtBe_prog : Program := u256LtBe_prog_of guestLayout
 
-    BE storage convention: byte 0 = MSB, byte 31 = LSB. Mirrors
-    the layout produced by `rlp_field_to_u256_be` and consumed by
-    `u256_lt` (PR-K50).
+def u256SubBe_prog : Program := u256SubBe_prog_of guestLayout
 
-    Building block for `tx_cost = max_fee_per_gas * gas_limit +
-    value` in tx validation, and for any subsequent u256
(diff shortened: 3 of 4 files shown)


pirapira:
Reviewed. **Approve and label.** This is the third conversion and the first that tests the theme at *scale* rather than at a new shape — eleven manifest Functions in one module, against one and three before.

Verified on the branch rather than from the report:

- **`GuestImageEntries.lean` byte-identical to main** — empty diff. That is the oracle, and it means the generator emitted the **concrete bridge programs** for the linked image, not the zero-layout ones, across eleven defs at once.
- **Emission-neutral**: `GuestAddrs.lean`, `symbol-addresses.tsv`, `RegionMap.lean` and `Proofs/GuestImage.lean` are all untouched. So a 1131-line module split in two moved no bytes.
- **The leaf imports no `GuestAddrs`** — only `Emit`, `AsmReloc`, `GuestLayout`, with the docstring stating the exclusion deliberately.
- **Twelve `_prog_of guestLayout` applications in the bridge against twelve `GuestLayout` fields**, so nothing was left behind in the split.

## The line counts are the reason this one needed reporting properly

Leaf **818**, bridge **350**, from a 1131-line original, against a 1500 hard cap with no opt-out. Both sides carry real headroom rather than merely passing — which matters because landing near the cap rearms a trap that cost three PRs to clear earlier today. Quoting **both post-split counts** instead of the pre-split total is what makes that checkable; a reader given only "1131 split in two" cannot tell whether either half is safe.

## What the third data point actually buys

The first conversion (HashBridge) proved the pattern; the second (BloomAddValue) tested it on a **different shape** — a direct `GuestAddrs` importer rather than a transitive one. This one tests a different **magnitude**: if the #10803 generator normalisation had been fitted to a one- or three-def module, eleven defs sharing one leaf is where that would surface. It did not.

Two data points was "works on something it wasn't written for." Three, with shape and scale varied independently, is a reasonable basis for rolling the rest out without re-litigating the generator each time.

## Field growth

**12 total** (10 + `u256m_acc` + `u256_mul_u64_be`). Worth continuing to report per PR: the mega-flat end state fails elaboration, so this number is the one that decides when the structure needs grouping, and nobody will reconstruct the curve from diffs. At +2 per conversion the current trajectory is comfortable, and the shared-symbol reuse seen in BloomAddValue (`zkvm_keccak256` not redeclared) is what keeps it linear in *distinct* symbols rather than in conversions.

## Note for whoever converts next

`GuestLayout` is a **serialisation point** for this theme — every conversion edits one shared field list, so two agents cannot convert concurrently without colliding there regardless of how independent the modules are. That is the real argument for #10804 (generating `GuestLayout`/`GuestLayoutInstance` from the fixture scan), and the trigger to act on it is concurrent conversions or a field list long enough that conflicts stop being trivial — not yet, but it should not be rediscovered.

pirapira:
Not merged — **the build fails on this branch by itself**, so this is not integration damage.

```
error: EvmAsm/Codegen/Proofs/U256IsZeroSpec.lean:53:2: don't know how to synthesize placeholder
error: EvmAsm/Codegen/Proofs/U256IsZeroSpec.lean:51:2: don't know how to synthesize placeholder
error: EvmAsm/Codegen/Proofs/U256IsZeroSpec.lean:50:2: don't know how to synthesize placeholder
error: EvmAsm/Codegen/Proofs/U256IsZeroSpec.lean:49:2: don't know how to synthesize placeholder
```

**Measured three ways so the attribution is not a guess:**

1. Your CI `build` fails on its own (19m47s).
2. I merged main `6d35394cc` in cleanly (no conflicts) and `lake build` failed on the same file.
3. I then checked out **`e186bc6c7`, the branch alone with no merge**, and built just `EvmAsm.Codegen.Proofs.U256IsZeroSpec` — the identical four errors. So the merge is not the cause.

**Why the refactor reaches that file.** Neither this branch nor main modifies `U256IsZeroSpec.lean` — the break is through the *definitional shape* of what it proves over. The file carries its own warning about exactly this:

> `runBlock` can't extend singleton specs to `CodeReq.ofProg <list-literal def>` directly (leaves unsynthesized frame placeholders); unfold to the union chain first.

This PR moves the U256 program defs into `U256Prog.lean` and parameterises them over `GuestLayout`, which changes that shape — so the unfold the proof relies on no longer lands and the frame placeholders go unsynthesized. `docs/agents/port-playbook.md` pitfalls has the pattern.

**Why I stopped here.** Repairing this means re-establishing the unfold against the new parameterised shape — proof engineering on your refactor, not the mechanical proof-body fixup I am allowed to do (adding a `simp only [GuestAddrs.foo]`, reordering rewrites, repointing a moved reference). The statement is unchanged, so a repair is legitimate; I am just not the right author for it, and guessing at frame instantiation is how a proof gets quietly weakened.

Everything else about the branch is fine, for what it is worth: the merge of main was conflict-free, containment checked out at 711 added lines with 0 missing, regen reached a true fixpoint on the first pass with all three step exit codes clean, and layout is unchanged — `textSizeBytes` `0x067140`, `.data` `0x5370`, `.bss` `0x1b85cae0`, all identical to main, which is what a parameterisation should do. 24 of 25 gates pass; `check-no-warnings` fails only downstream of the build.

Once the proof is repaired I will re-sync and land it.

pirapira:
## Closing for now — parked on a `runBlock` limitation, not on anything wrong with this PR

Everything in this change is healthy. It is **git-mergeable and conflict-free** against main, containment shows **711 added lines and 0 missing**, the regen reached a **true fixpoint on the first pass** with all three step exit codes checked, and the layout is **identical to main** — `textSizeBytes 0x067140`, `.data 0x5370`, `.bss 0x1b85cae0`. `GuestImageEntries.lean` regenerates **byte-identically**. 24 of 25 gates pass.

That layout identity is the important part: it proves the parameterisation is **emission-neutral** at eleven Functions, so the generator fix from #10803 is general rather than fitted to the one- and three-Function modules it was tested on. This PR established that even though it is not landing.

## What blocks it

```
EvmAsm/Codegen/Proofs/U256IsZeroSpec.lean:53:2: don't know how to synthesize placeholder
  (also :49, :50, :51)
```

A file this diff never touches. It builds singleton instruction specs and then unfolds, because `runBlock` cannot extend singletons to `CodeReq.ofProg` over a **list-literal def** directly — it leaves unsynthesized frame placeholders, which the file documents itself citing `docs/agents/port-playbook.md`. The leaf/bridge split changes the definitional shape that unfold depends on.

**The fix is a tactic extension, not a proof repair:** `runBlock` needs to unfold a `GuestLayout` field lookup applied to a concrete layout. That is metaprogramming plus a performance comparison — `runBlock` sits on hot proof paths — and it is deliberately not being attempted by hand here, because re-establishing a frame instantiation on someone else's refactor is how a proof gets quietly weakened, and a weakened proof passes.

## What this PR established regardless

**U256 is the first converted module with dependent proofs**, which is why HashBridge (#10800) and BloomAddValue (#10812) could not reveal this. So the boundary is not *scale* — it is **the presence of proofs that unfold the concrete `_prog`**.

**The conversion pattern works for modules without dependent proofs and is blocked for modules with them.** That is recorded on #10753 with the ordered next steps, along with a standing requirement to grep for `Proofs`/`Spec` files referencing a module's `_prog` before converting it.

## Re-opening

Re-open or re-cut once the `runBlock` extension lands. The diff is four hand-written source files with no generated artifacts, so it should re-apply cheaply. Tracking on **#10753**.

Issue #10830, opened 2026-07-28
bal: rebuilt BAL differs from expected bytes at IDENTICAL length and row counts — account ordering and field order exhaustively exonerated, content unlocalised


pirapira:
## What is EXCLUDED (read this first — it is the useful part)

On `blockchain_tests/for_amsterdam/amsterdam/eip7928_block_level_access_lists/block_access_lists_eip4895/bal_withdrawal_and_value_transfer_same_address.json`,
the guest's rebuilt BAL differs from a byte-exact model of the expected bytes **while the length and every
row count agree**. Four hypothesis families were tested and closed. **Do not re-run these:**

| hypothesis | scope tested | result |
|---|---|---|
| account ordering | **all 362,880** permutations of the 9 account blobs | **none reproduces the guest hash — exonerated on this fixture** |
| `AccountChanges` field order | **all 120** permutations of the 5 fields after the address | **none matches — excluded** |
| reads reversed / balanceChanges reversed / both | 3 targeted transforms | no match |
| one balance row emitted as that account's **pre**-state balance (the #10809 shape, which preserves length because pre and post are both 12 bytes here) | every balance row in turn | **no match — that hypothesis is dead** |

## The measurement

Guest `263f5d03ecc9758744d57f33e6e321b62fe89cafb180a0db1c00d7e2e6a138e9` (branch of #10822, i.e. with the
balance-serialization fix), spike, sha-verified distinct before reading. Hash published by #10829.

| | |
|---|---|
| declared balance rows / emitted | 4 / 4 — **zero missing**, so the model is unambiguous (no search) |
| model length / `rebuilt_len` | **347 / 347 — equal** |
| `keccak(model)` | `204b0febcfe7548e…` |
| guest `bal_serializer_rebuilt_hash` | `4c6f546a09645bd3…` |
| | **DIFFERENT** |

**Control, passing on all six fixtures:** the published *supplied* hash equals
`keccak(BAL bytes extracted from the guest input)`. That validates the extraction, the encoder, the keccak
(checked against `keccak256("") = c5d24601…`) and that the publish reads the right buffer — **and it
establishes separately that the declared BAL order is already canonical**, since a fixture-order
reconstruction reproduces the declared bytes exactly. That is what licenses the ordering exoneration above.

## The instrument fact, which is the reusable half

**This is a defect that no counter and no length comparison can see, measured on a real fixture.**

[...]

PR #10834, opened 2026-07-28
Feed touched accounts into the BAL builder

pirapira:
## Summary

Feeds the promoted block-level `account_reads` set into the existing BAL builder before serialization. This materializes the empty account-list entries that `execution-specs` adds for every touched account.

This is a correctness improvement with no observed acceptance movement: 670 fixtures newly digest-match, zero formerly-matching fixtures stop matching, and the paired 200-row Spike gate remains FA=0 / FR=7 on both legs.

The MTx transaction-boundary read-set promotion is made unconditional, matching `incorporate_tx_into_block`: reads survive a failed transaction even when the AccountState commit is bypassed. The join is deliberately after the selfdestruct-read producer, rollback/effects, and before BAL emission. `read_sets_incorporate_tx` clobbers caller-saved `a1/a2`, which already carry the committed storage slice for the snapshot upsert, so the call preserves `ra`, `a1`, and `a2` across the helper.

No theorem statements changed.

## Targeted BAL diagnostic-family result

This is the 1,024-row `eip7928_block_level_access_lists` diagnostic family. It deliberately concentrates BAL behaviour, but it is a name-selected subset: Amsterdam makes the block access-list hash mandatory and checks it for every block.

Immutable artifacts:

- Parent: `dbcb072b447af936567376ed5d88689ed8b361064042b84901336fec0d03e021`
- Candidate: `96e626db477a952a9210d8e88b13e18a3c98ce6f8004654a07ba412523f3133c`

| Parent -> candidate shadow status | Count |
| --- | ---: |
| matching -> matching | 136 |
| mismatching -> matching | 670 |
| mismatching -> mismatching | 213 |
| unready -> unready | 5 |

There were no missing/extra labels, no errors, and **zero stopped-matching rows**. Within this diagnostic family, the match rate rises from 136/1024 (13.3%) to 806/1024 (78.7%). All 670 cleared fixtures had at least one declared empty account entry, confirming the predicted mechanism.

The original swept candidate was replaced by a regenerated shipping ELF. Repeating the diagnostic on that distinct artifact reproduced the transition table exactly (136/670/213/5), with zero errors and no label drift; this is evidence that the required layout regeneration was behaviour-preserving.

Spec basis: `block_access_lists.py:499` adds a touched-but-unmodified account, and the builder loop at `block_access_lists.py:697` consumes touched accounts. `fork.py:1204` calls `incorporate_tx_into_block` unconditionally; the guest had promoted reads only on its AccountState commit route.

Residuals are deliberately separate follow-up classes: 89 rows still have a declared empty account entry (a further mechanism in the same structural area); 124 have no empty account entry. A secondary prediction that unconditional read promotion would clear some of the 124 was falsified: none moved.

Of the 213 remaining mismatches, 162 currently accept and 51 already reject. These are subset counts only; corpus-wide comparator cost requires an independent, non-name-selected measurement.

## Random paired gate

Spike, immutable runner SHA `bae4826988540415b9bf862248caa7499b2ffbb9eec5cfab61cce00e089cf2c3`, captured seed `1415100903`:

- Parent and regenerated candidate: 200/200 rows; FA=0, FR=7, error=0, budget=0.
- Both legs: 193 full matches, 200 root matches, 200 tail matches.
- `eest-ab-compare.py`: complete denominators, 200 joined input groups, zero status differences, and byte-identical output on all 200 cases.

The captured seed was generated by the candidate leg before observation, then reused for the parent leg solely to pair the artifacts.

`LAKE_ARTIFACT_CACHE=false lake build codegen` passes.


pirapira:
Reviewed. **Approve and label.** This is the largest correctness result in the queue and the implementation is careful in the four places it could have gone wrong. Verified on the branch rather than from the report.

## Exactly one call site

`jal ra, read_sets_incorporate_tx` appears **once** on the branch, in `BlockVerdictMtxRuntime`. The old call in `account_state_commit_pending` is **removed**, not left alongside — so the merge-and-clear runs once per transaction rather than twice. That was the thing most likely to be silently wrong, and a second call would have cleared a set the first had already consumed.

The replacement comment in `CreateCodeEffectLog` is the right artifact too: it says the promotion is *deliberately not here* and names who owns it. A reader arriving at that function later needs to know the absence is intentional, not an oversight.

## The join-point placement is correct for two independent reasons

`.Lbv_mtx_code_commit_done` is where the commit and bypass paths converge, so a single call there is **unconditional by construction** — matching `fork.py:1204`, which calls `incorporate_tx_into_block` with no success guard.

And it is *after* `account_state_promote_delete_reads`, which puts SELFDESTRUCT storage rows into the tx-level set on the commit path (#10645). Placing the merge earlier, or replacing that call, would have **silently dropped those reads** — a regression appearing as a new shortfall on selfdestruct fixtures while the touched-account number improved, and attributed to the wrong change. Both constraints are satisfied only at this point.

## The overflow latches moved, and had to

The three read-side latches (`storage_reads_overflow`, `account_reads_overflow`, `code_reads_overflow`) move from before the commit to immediately **after** the merge. That is not tidying: `read_sets_merge_one` sets them *through the `a7` pointers the merge hands it*, so checking them before the merge examines a value the merge has not yet written. Consuming each latch right after is what makes them mean anything.

## `a1`/`a2` preserved across the inserted call

The joined paths prepare `a1`/`a2` as the storage slice for the committed-chunk snapshot, and the read helper uses caller-saved argument registers — so the save/restore is required, not defensive. Worth noting the **first candidate faulted here**, which is why this is in the diff at all. That is the third register-liveness incident today and only this one was loud; a site that previously had no call gives the surrounding code no reason to have protected anything across it.

## The touched-account loop matches the spec's primitive

`bal_builder_incorporate_touched_accounts` walks `account_reads_count` rows at 32-byte stride and passes each BE20 address to `bal_builder_ensure_account` — which is exactly `add_touched_account`'s behaviour: a miss creates an entry with all five change lists empty, a hit preserves an entry an earlier change producer made. Idempotent by construction, which is what makes calling it over the whole set safe.

Hooked at the **top** of `bal_serializer_rebuild_hash`, before every sort, with the reason stated: `_build_from_builder` folds the read set in first, and the account walk is the single source of outer BAL rows. Fold after the sort and the new entries would be unsorted.

## Evidence

| | parent | candidate |
|---|---|---|
| status0 | 136 | **806** |
| status1 | 883 | 213 |

**670 newly matching, 0 stopped matching, 0 errors**, transitions accounted exactly on the complete 1024-row selector — complete rather than sampled, because the comparator fires only when `ready = 1` and non-BAL blocks cannot reach it. **All 670 that cleared had at least one declared empty entry**, so the population that moved is exactly the one the hypothesis named.

Paired random gate: 200/200 both legs, FA 0, **FR 7 on both**, error 0. A change clearing 670 digest mismatches moves the acceptance corpus not at all.

Both residual classes named separately — **89** empty-entry (further mechanism, 75 of them one-tx) and **124** no-empty (pre-existing, includes #10830) — so neither reads as this fix being incomplete. And the **falsified secondary prediction is recorded** rather than omitted: the unconditional promotion was expected to close some of the 124 via unpromoted reads on failed transactions, and none moved.

No theorem statements changed, full codegen build succeeded.

pirapira:
## Correction to one claim in this PR's body — the "complete selector" justification is wrong, and it was mine

I told codex to write that the 1024-row selector is **complete** rather than sampled, on the grounds that *"non-BAL blocks structurally cannot reach the comparator."* **There are no non-BAL blocks in Amsterdam.**

- `block_access_list_hash: Hash32` is a **mandatory, non-optional** header field (`blocks.py:251`).
- The comparison at `fork.py:390` is **unconditional** — no `has_bal` guard, no fork gate.

So every block in the corpus carries a BAL and every block's hash is checked. The `eip7928_block_level_access_lists` directory is the family that **tests** BAL behaviour, not the set of blocks that **have** one. The 1024 rows are a **name-selected subset**, and the enumeration built on them measures that subset rather than the corpus.

This is a right-mechanism/wrong-population error on my part, and it propagated into this body at my instruction.

### What this does NOT change

**Everything measured here stands.** 136 → 806 on those 1024 rows, 670 newly matching, **0 stopped matching**, 0 errors, all 670 carrying a declared empty entry, and the paired random gate at 200/200 with **FR 7 on both legs**. The fix is additive, spec-cited, and moves acceptance not at all. None of that depended on the selector being exhaustive.

The residual counts also stand *as counts over that subset*: 89 empty-entry and 124 no-empty, within the 1024.

### What it does change

Two claims should be read as **subset-scoped, not corpus-scoped**:

1. **"Complete rather than sampled"** → the selector is the eip7928 *test family*, which is where BAL behaviour is exercised deliberately. That makes it the right *diagnostic* population, but not the whole population the comparator would act on.
2. **The 832 → 162 downstream figure** → those are counts within the 1024. The corpus-wide false-reject cost of binding the comparator is **much larger** — an independent non-name-selected measurement puts it near 88% of accepting blocks, which is being re-measured now with a stated denominator.

So this PR's own numbers are sound and its landing case is unaffected. What must not travel is the inference that 1024 bounds the comparator's blast radius.

### Why the error was invisible from inside the selector

No sample size within a name-selected run could have caught this, because **the defect was in the selector itself** — every row it drew satisfied the same wrong predicate. It was caught by a random, non-name-selected gate on the first try. That is the argument for keeping an unselected sample alongside any targeted population, and it is worth more than the correction.

PR #10835, opened 2026-07-28
Wire the rebuilt-BAL digest into the block verdict (#10680)

pirapira:
## What this does

`bal_serializer_verify` already rebuilds the BAL from the producer arenas, hashes the supplied BAL,
and returns `0` match / `1` differ / `2` rebuild-failed. Its result was computed and published to a
shadow counter but **never consumed**. This makes it participate in the verdict. GH #10680.

The comparison itself is unchanged. This is the bind, and nothing else.

## Binding contract

- Binds **only** when the original verdict is ACCEPT (`a0 == 1`) **and** `bv_bal_shadow_ready == 1`.
  Every existing reject flows through untouched, so **no existing fail code changes meaning** and
  every current test expectation survives.
- Therefore it can only convert an ACCEPT into a REJECT. **It cannot create a false accept** — there
  is no path by which it raises FA. Any FA movement would be a genuine defect, not a cost of this
  change.
- `a0` is compared against `1` rather than tested nonzero. An accept route returning some other
  nonzero value would be *under*-bound (fewer new rejects) — the safe direction, and it would show up
  as an FR delta below prediction rather than as a silent behaviour change.
- Two codes, not one: **60** = BAL mismatch, **61** = rebuild failure (canonical-sort failure, arena
  overflow). Status 2 measures zero across the corpus so 61 should never fire, but collapsing them
  would make a capacity limit indistinguishable from a wrong BAL the first time it does. The base tree
  uses codes **1..59 with gaps at 8, 9, 18**; neither new code reuses a gap, since a gap may be a
  retired code.
- **Retires no granular BAL check.** Those carry skip lists that are accommodations for producer gaps,
  so retiring them removes the skips and is a *second*, separately unpredictable FR event. One
  variable at a time.

## Cost, with denominators

Measured on a **random, non-name-selected sample of 200 stateless blocks**, seed **382805339**
(captured from the harness, not chosen), corpus of **26104 candidates**. Both legs ran **byte-identical
inputs** (verified by `cmp` on all 200, not by name matching) and differ only by this bind:
parent guest `dbcb072b447af936`, candidate `7d4642a48167712a`, base `6d35394cc`.

| cell | n | accept before | accept after | new false rejects |
|---|---|---|---|---|
| eip7928 family | 10 | 10 | 1 | 9 |
| other families | 190 | 170 | 21 | 149 |
| **all** | **200** | **180** | **22** | **158** |

- **158 of 200 sampled blocks** newly rejected = **79.0%** (95% CI 73.4–84.6%); **158 of 180**
  accepting blocks = **87.8%**.
- Projected over the corpus: **~20600 false rejects of 26104** (95% CI 19100–22100).
- **False accepts: 0**, and impossible by construction.

**Containment.** Parent failed 7 with codes `{59:2, 37:2, 56:1, 47:1, 41:1}`. Candidate failed 165:
**exactly those seven with an identical code multiset**, plus 158 new, all code 60. The change adds
one code and perturbs no existing failure, code or count.

### Do not use the earlier 832 figure

An earlier measurement in #10680 reported **832 new false rejects over 1022 fixtures**. That
population was selected by the name prefix `test_bal_*`, and **the bind fires on every block** —
`block_access_list_hash` is a mandatory header field (`blocks.py:251`) and the check at
`fork.py:390` is unconditional, so every Amsterdam block carries a BAL. The `eip7928` directory is the
family that *tests* BAL behaviour, not the set of blocks that *have* one. 832/1022 = 81.4% is the same
corpus-wide rate seen through a name filter, not a bounded total. Correction posted on #10680.

## How fast the baseline falls: measured, not assumed

The ~20600 above is against **today's main**. A third leg on the *same* seed and the *same*
byte-identical inputs, carrying #10834 (touched-account feed) plus this bind, measures what the
baseline becomes once that producer fix lands:

| leg | accept | new false rejects | via code 60 | via other codes |
|---|---|---|---|---|
| A — base | 180 | 0 | 0 | 0 |
| B — base + bind | 22 | **158** | 158 | 0 |
| C — #10834 + bind | 108 | **72** | 72 | 0 |

- FR **158/200 = 79.0%** → **72/200 = 36.0%** (95% CI 29.3–42.7%); over 26104, **~20600 → ~9400**
  (95% CI 7700–11100). #10834 roughly **halves** the baseline.
- **The progress metric is the better number**: digest-matching blocks go **22/200 = 11.0%** →
  **108/200 = 54.0%**, a five-fold improvement with a clean denominator and a meaningful ceiling.
- The eip7928-family cell (n=10) goes from 9 false rejects to **0**.
- **Every FR in both bind legs is via code 60, none via any other code** — so #10834 altered no
  granular-check outcome, and the 72 is attributable purely to the digest without needing a fourth leg.
- Fail-code multiset containment holds in both bind legs: the base multiset
  `{5:10, 11:1, 24:1, 37:2, 41:1, 47:2, 56:1, 59:2}` is **unchanged**, plus code 60 only. No
  pre-existing code is reduced.

### What the residual is made of

The empty-account-entry marker is computable from the **supplied BAL in the fixture input**, with no
guest run, so this was registered as a prediction before leg C ran rather than read off afterwards:

- The marker predicts mismatch essentially perfectly: **150 of 158** false rejects carry it, and
  **0 of 22** already-matching blocks do.
- Predicted: the 8 marker-free false rejects should *survive* #10834, since the fix targets the marker.
  **All 8 survived (8/8).** Of the 150 marker-carrying, 86 cleared and 64 survived.
- So the residual 72 = **64 marker-carrying + 8 marker-free**.

**The marker generalising as a predictor is not the fix generalising as a remedy** — the predictor was
perfect, the remedy cleared 54.4% corpus-wide against 75.9% within the family. And the composition
inverts: within the eip7928 family the post-fix residual is 89 marker-carrying vs 124 marker-free
(41.8% carrying), while corpus-wide it is 64 vs 8 (**88.9% carrying**). The
marker-present-but-not-cleared class is a minority of the family residual and the **dominant** class
corpus-wide, so the family understated it by more than a factor of two. That makes it the highest-value
next producer target.

## Why bind at a 20k baseline

Per the maintainer: this is acceptable because **it becomes the new baseline from which BAL production
debugging continues.** The point of binding is not a green corpus — it is that the defects become
visible in the ordinary run instead of requiring a special instruction and a special setup to see. A
20k baseline that every subsequent producer fix visibly reduces is more useful than an invisible 20k
that only a commissioned enumeration finds. The corpus-wide 79% was discovered precisely because
nobody was looking.

The bind is behaving to contract; what it reports is that **the rebuild disagrees with the supplied
BAL on ~88% of accepting blocks**. This is a producer-readiness finding, not a defect in the bind, and
it is not one mechanism — the eip7928 family alone yielded several distinct causes.

## Two consequences for everyone's gates from here

Please inherit both:

1. **An absolute FR count stops being a regression signal** at this baseline — a five-fixture
   regression is invisible in that noise. **Paired same-seed parent-vs-candidate deltas still work**
   and detect a five-row change regardless of absolute level. Keep the paired form mandatory.
2. **The digest match count becomes the progress metric.** Report `status0` rising, not FR falling.
   They move together, but `status0` has a meaningful ceiling and a clean denominator.

## Proofs

**No spec or proof file is touched and no theorem statement changed.** The diff is the bind plus six
`#guard`s in `BlockVerdictReceiptsTail.lean`, plus regenerated layout (`GuestAddrs`, the `RegionMap`
`.text` size pin, `symbol-addresses.tsv`).

Register liveness at the insertion point is clean **by construction, not by inspection**: the bind
sits immediately before the function epilogue, so everything after it is sp-relative callee-saved
restores and a `ret`. The `t0`/`t1` it clobbers are dead at the boundary. Read in the emitted asm, not
the Lean strings.

## Guards

Six, pinning the parts a later edit could loosen silently: the accept-only guard, the ready gate, and
both codes *together with the `li a0, 0`* that drops the verdict — a fail code stored without it would
report a mismatch while still accepting the block, which is the one failure mode of this change that
no test would catch. Non-collision with existing codes is not checkable in a `#guard` (they are stored
from other modules), so it is established above rather than asserted by a guard.


pirapira:
Reviewed. **Approve and label.** This is the #10680 landing step and the binding is the narrowest shape that can do the job. Verified on the branch rather than from the report.

## The contract is right, and the guards pin the part that could rot

```
li t0, 1; bne a0, t0, .Lbv_bal_digest_bound      <- accept-only
la t0, bv_bal_shadow_ready; ld t0, 0(t0); beqz t0, ...  <- ready-only
```

Binding only on `a0 == 1` **and** `ready == 1` means every existing `a0 == 0` reject flows through untouched — no existing fail code changes meaning, every current test expectation survives, and the FR delta is fully attributable. It also gives the safety property in its strongest form: **the change can only convert an accept into a reject, so there is no path by which it raises FA.**

Comparing `a0` against `1` rather than testing nonzero is the right choice for the reason stated: an accept route returning some other nonzero would **under-bind** — fewer new rejects, visible as an FR delta below prediction rather than as a silent behaviour change. Failing safe *and* observably.

**Codes 60 and 61 are genuinely free** — I checked `main` and the highest in use is 59, with no store of 60 or 61 anywhere. Keeping them distinct when status 2 measures zero today is correct: collapsing them would make a capacity limit indistinguishable from a wrong BAL the first time it fires.

The best guard here is the last pair:

```
#guard ... "li t0, 60; la t1, bv_fail_code; sd t0, 0(t1); li a0, 0" ...
```

Pinning the fail code **and** the `li a0, 0` in one string, because a code stored without dropping the verdict would report a mismatch while still accepting the block — and as the comment says, that is the one failure mode of this change no test would catch. Pinning the pair rather than each half is what makes it unforgeable. Equally right: stating that non-collision **cannot** be guarded here (the other codes are stored from other modules) and establishing it in the body instead, rather than writing a guard that only appears to check it.

## Not retiring any granular check is the correct scope

Those checks carry skip lists that are **accommodations for producer gaps**, so retiring them is a separate and separately unpredictable FR event. One variable at a time.

## Numbers, with denominators

Three legs, one seed, 200 blocks: **FR 158 → 72 of 200**. Projected corpus-wide (26,104) that is ~20,600 → **~9,400**, so #10834 roughly **halves** the baseline this PR introduces.

And the metric that should lead downstream: **digest-matching blocks 22/200 (11.0%) → 108/200 (54.0%)** — five-fold, one number, clean denominator, meaningful ceiling.

**The out-of-bracket result is reported as out-of-bracket.** 72 against a pre-registered 8/38/158 is a miss, and saying so rather than fitting it to the nearest bracket is what keeps the pre-registration worth anything. That the per-fixture prediction inside it was exact is the more informative half.

## One note for the record

The `832` figure that circulated earlier was a **subset count over the eip7928 diagnostic family**, presented as a corpus cost. That error was mine — I argued the flip set was bounded by BAL-bearing blocks, and there are no non-BAL blocks in Amsterdam (`blocks.py:251` mandatory, `fork.py:390` unconditional). This PR telling readers explicitly not to use that number, and why, is the right thing to carry.

No theorem statements changed; no spec or proof file touched at all.

PR #10841, opened 2026-07-28
fix(tactics): normalize GuestLayout program bridges in runBlock

pirapira:
## Summary

`runBlock` now recognizes the narrow layout bridge shape `foo_prog := foo_prog_of guestLayout` beneath `CodeReq.ofProg`. It unfolds only that bridge and its concrete `_prog_of` application, then uses the existing singleton-chain composition path.

All ordinary `CodeReq.ofProg` users retain the pre-existing generic-delta route unchanged. The special path is entered only after inspecting the bridge definition body; arbitrary opaque programs remain opaque.

The bounded normalizer has fuel 4. Exhaustion reports `runBlock: layout CodeReq.ofProg normalization exhausted 4 steps at <name>` with action guidance, rather than surfacing later as a framing placeholder.

## Evidence

- Exact conversion reproduction: `U256IsZeroSpec` builds on the layout-conversion branch after removing its documented manual `simp` workaround. The special route performs exactly two deltas: `u256IsZero_prog`, then `u256IsZero_prog_of`.
- Permanent main test module `EvmAsm.Tests.RunBlockLayoutBridge` covers the bridge success case, a `#guard_msgs` opaque non-Codegen rejection, and a `#guard_msgs` five-hop fuel-exhaustion diagnostic naming `tooDeep4_prog_of`.
- Current-main checks: `RunBlock`, `U256IsZeroSpec`, and `CallReturn` build.
- Profiled legacy U256 path at base `d47ccc767`: median 53.005 ms, range 43.989--70.871 ms. Candidate: median 44.055 ms, range 43.611--84.188 ms. No regression is detectable at this sample size because noise exceeds the median difference.
- Additional ordinary-user sanity check: `RLP/WalkNext` has 48 `runBlock` calls; its slowest profiled call was 279.199 ms, below the 800 ms ceiling.
- The converted U256 route is approximately 188 ms inside `runBlock`, below the same 800 ms ceiling. This cost applies only to layout bridges and retires the hand-written workaround.

No heartbeat, recursion-depth, or tactic-bound changes; no `native_decide` or `bv_decide`.

### DIFF
--- a/EvmAsm.lean
+++ b/EvmAsm.lean
@@ -43,3 +43,4 @@ import EvmAsm.Tests.ArithDiffCheck
 import EvmAsm.Tests.Div128V5RandomCheck
 import EvmAsm.Tests.RlpDiffCheck
 import EvmAsm.Tests.SpecRefEestCheck
+import EvmAsm.Tests.RunBlockLayoutBridge
--- a/EvmAsm/Rv64/Tactics/RunBlock.lean
+++ b/EvmAsm/Rv64/Tactics/RunBlock.lean
@@ -1066,6 +1066,42 @@ private partial def extractCrEntriesPure (cr : Expr) : List (Expr × Expr) :=
     extractCrEntriesPure args[0]! ++ extractCrEntriesPure args[1]!
   else []
 
+/-- Return the program argument of the first `CodeReq.ofProg` in a CodeReq tree. -/
+private partial def ofProgArg? (cr : Expr) : Option Expr :=
+  if cr.isAppOfArity ``EvmAsm.Rv64.CodeReq.union 2 then
+    let args := cr.getAppArgs
+    (ofProgArg? args[0]!).orElse fun _ => ofProgArg? args[1]!
+  else if cr.isAppOfArity ``EvmAsm.Rv64.CodeReq.ofProg 2 then
+    some cr.getAppArgs[1]!
+  else none
+
+private def isConcreteGuestLayout (e : Expr) : Bool :=
+  match e.getAppFn with
+  | .const name _ => name.toString == "EvmAsm.Codegen.guestLayout"
+  | _ => false
+
+/-- Recognize exactly the two-step Codegen bridge shape
+    `foo_prog := foo_prog_of guestLayout` and its applied `_prog_of` target.
+    Unrelated opaque `CodeReq.ofProg` arguments deliberately remain opaque. -/
+private def layoutProgramHeadDef? (prog : Expr) : MetaM (Option Name) := do
+  match prog.getAppFn with
+  | .const name _ =>
+    let args := prog.getAppArgs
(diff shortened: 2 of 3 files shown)


pirapira:
Performance could be better.

pirapira:
## Build-time question, examined and found confounded — recording it so it is not rediscovered

**Concern raised:** a comparable CI build (run `30197397247`) is ~20% faster on major files than this PR's build (run `30394822831`), suggesting the tactic slowed things down. The lean-action log prints a per-module time, so this is checkable.

**Measured, on the 163 modules ≥10s present in both runs:** 138 slower, 11 unchanged, 14 faster, **median +19.0%** — which reproduces the ~20% exactly.

**But the pattern does not fit the tactic:**

| module | before | after | Δ | uses `runBlock`? |
|---|---|---|---|---|
| `Codegen.Programs.Secp256k1PointDoubleSAsmReg` | 28s | 43s | **+54%** | **no — 0 occurrences of `runBlock` or `ofProg`** |
| `Evm64.MLoad.Spec` | 36s | 55s | +53% | yes (10 `ofProg`) |
| `Rv64.Tactics.RunBlock` | 16s | 11s | **−31%** | the file this PR changes |

The single worst regression is in a module the tactic **cannot** reach, and the changed file itself got *faster*. If the guard were the cost, the regression would concentrate in `ofProg` users and spare pure asm-string modules — the opposite of what appears. A near-uniform +19% across unrelated modules is the signature of **runner variance between two builds five days apart**, not a tactic regression.

**So this comparison neither convicts nor clears the change** — commit and machine vary together, so the effect is confounded. The clean measurement is two builds over the same hub on the same runner class: #10842 sits directly on top of this PR and rebuilds the same modules, so `#10842 vs #10841` isolates the conversion; re-running this PR's build against itself would bound the variance.

**What remains true independent of the timing data:** the guard runs a small `Expr` walk (`ofProgArg?`) on every `runBlock` goal even when no layout bridge is present. That is far cheaper than the definition-body inspection which caused the earlier CI slowdown — the fix for that was making the delta exemption apply *only* to a pre-identified bridge, so ordinary `ofProg` keeps the pre-existing path — but it is not zero. Implementing the normalization through `simp` would avoid it and would sit on a path upstream keeps fast and well supported. Recorded as a possible follow-up rather than a blocker; the maintainer elected to land this as-is.

**Timing evidence as it stands:** legacy `runBlock` proof re-elaboration, base median 53.005 ms (range 43.989–70.871) vs candidate 44.055 ms (range 43.611–84.188) at n=3 — spreads exceed the difference, so read as *no regression detectable at this sample size*, not as an improvement. Heavy unconverted user `RLP/WalkNext`, 48 `runBlock` calls, max 279.199 ms. Converted path 188.020 ms with exactly two guarded deltas. All well under the 800 ms ceiling.

PR #10842, opened 2026-07-28
refactor(codegen): parameterize U256 over GuestLayout

pirapira:
## Summary

Parameterize the U256 program definitions over `GuestLayout`, keeping their emission views at `GuestLayout.zero` and preserving the existing concrete bridge API.

The second commit removes the documented manual `simp` expansion from `U256IsZeroSpec`. This is the real converted-module success path for #10841's guarded `runBlock` normalization: the proof now succeeds through the bridge without theorem-local unfolding.

The final checker repair keeps the generated-source drift guard non-vacuous after the split: layout-independent leaves canonically bind their unused layout argument as `_L`, while leaves with linked-address references retain `L`. The checker derives that distinction from the rendered body and uses the shared `layout_leaf_path` helper to locate a split leaf.

## Validation

- `lake build EvmAsm.Codegen.Proofs.U256IsZeroSpec EvmAsm.Tests.RunBlockLayoutBridge` passed locally after deleting only two read-only, cache-hardlinked `.ilean` outputs. This is cache-mode recovery, not a clean-build claim; CI is the authoritative clean verification.
- The target proof reaches the guarded normalizer through `u256IsZero_prog` then `u256IsZero_prog_of`.
- `bash scripts/check-asm-to-program.sh` passed: `check-all: CLEAN (390 converted defs across 148 files)`. The check still validates each full generated leaf and bridge verbatim; it does not add an exemption.

This PR is stacked on #10841 so its diff contains only the U256 conversion, checker compatibility repair, and regression evidence. It must not be squashed: the simp-removal commit demonstrates the preceding parameterization exercises the tactic success path.


pirapira:
Depends on #10841 performance.

PR #10845, opened 2026-07-28
chore: batch-merge #10841 + #10844

pirapira:
Batch-merge of #10841 #10844 (all author pirapira). Built on `origin/main` `d47ccc767`, re-verified current before pushing. Both merged **clean**, no conflicts.

**#10844 is the soundness fix** — it stops the rebuilt BAL echoing the oracle during callee seeding. Verified the change is at the right call site rather than assuming:

| line | function | accessor |
|---|---|---|
| 132 | `seedCalleeStorageFunction` | `account_at_header_state_root` — **the fix**, raw |
| 910 | `dispatchTxRuntimeCodeFunction` | `account_at_header_state_root_tracked` — correctly unchanged |

So the seeding path no longer records a read, while the execution path still does. One `_tracked` occurrence remains in the file and that is the correct one.

## Containment audit

- **#10841** — 118 added lines, **0 missing**
- **#10844** — 1 added line, **0 missing**

## Layout: unchanged

Regen reached a true fixpoint on the first pass, all three step exit codes 0, **zero changes**: `textSizeBytes` `0x067228`, `.data` `0x5370`, `.bss` `0x1b85cae0`, TSV 3234 data rows — all identical to main. Expected: #10841 is tactics plus a test, and #10844's two accessor names encode to the same call length, so `.text` does not move.

## ALL 25 `check-*.sh` PASS

`lake build` exit 0 (4786 jobs). The four `lake env lean` gates were run after materializing their drivers' oleans with `LAKE_ARTIFACT_CACHE=false` (AGENTS.md, #10537) — nothing skipped.

### DIFF
--- a/EvmAsm.lean
+++ b/EvmAsm.lean
@@ -43,3 +43,4 @@ import EvmAsm.Tests.ArithDiffCheck
 import EvmAsm.Tests.Div128V5RandomCheck
 import EvmAsm.Tests.RlpDiffCheck
 import EvmAsm.Tests.SpecRefEestCheck
+import EvmAsm.Tests.RunBlockLayoutBridge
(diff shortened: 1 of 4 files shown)


pirapira:
Superseded — closing rather than syncing.

#10841 was merged directly (main is now `7f0e87db8`), so this batch's remaining contribution is exactly the one line #10844 already carries:

```
EvmAsm/Codegen/Programs/BlockVerdictDispatchTx.lean | 2 +-
```

Syncing this branch would spend a CI cycle to land something #10844 can land on its own, and #10844 is already up to date with the new main with a fresh run in flight. Arming #10844 instead.

For the record, the verification done on this batch still stands and applied to both halves: containment 118/118 and 1/1 with none missing, layout unchanged (`textSizeBytes` `0x067228`, `.data` `0x5370`, `.bss` `0x1b85cae0`), regen a true fixpoint on the first pass, `lake build` exit 0, and all 25 `check-*.sh` green — including confirmation that the accessor swap landed in `seedCalleeStorageFunction` (line 132, raw) while `dispatchTxRuntimeCodeFunction` (line 910) correctly keeps `_tracked`.

PR #10847, opened 2026-07-28
fix(bal): record the CALL-family target as an account read

pirapira:
## What

Record the **call target** as an account read in the shared tail that all four
CALL-family handlers (`CALL`, `CALLCODE`, `DELEGATECALL`, `STATICCALL`) run.
Two functional lines at one site.

## Spec alignment (pinned commit `e5a8caf1b`)

This change **moves the guest toward execution-specs**: it adds a touch the spec
performs unconditionally and the guest was omitting.

The aligning sites are the four call opcodes in
`src/ethereum/forks/amsterdam/vm/instructions/system.py`:

| opcode | line | statement |
|---|---|---|
| `call` | `:460` | `code_hash = get_account(tx_state, code_address).code_hash` |
| `callcode` | `:583` | *(identical)* |
| `delegatecall` | `:757` | *(identical)* |
| `staticcall` | `:856` | *(identical)* |

In all four, `code_address = to_address_masked(pop(evm.stack))` — the **popped
call target**, including for `CALLCODE`/`DELEGATECALL`, where the target is *not*
the executing account. `get_account` records the touch as its **first**
statement, `state_tracker.py:199` (`tx_state.account_reads.add(address)`), so the
read is recorded unconditionally and even for an absent account.

Line numbers were read from the **committed objects** at the pin
(`git show e5a8caf1b:...`), not from the working tree.

### Explicitly *not* `vm/interpreter.py:291`

An earlier draft of this change cited `interpreter.py:289-292`
(`get_account(tx_state, recipient)`). That citation is **wrong for this site**:
the call sits in `prepare_dispatch` (defined `:246`), whose only call site
(`:365`) is guarded by `if message.depth == Uint(0)` and whose docstring states
"Runs at the top frame (depth 0)". It justifies the separate **top-level
recipient-resolution** fix, not this child-frame one. The two changes are cleanly
separated by frame depth and neither subsumes the other.

## Why one shared site rather than four per-handler inserts

On **behavioural equivalence**, not on structural mirroring. The spec does have
four separate per-opcode touch sites, so four inserts would mirror its text more
literally. The shared site is justified because those four sites are textually
identical and take the same operand, making one emission equivalent to four —
and the guest already funnels all four variants through this single tail
(`basicPrecompileCallTail`), which has *already* materialised that exact BE-20
operand for the access charge. Two source lines instead of eight is a consequence, not
the reason.

The insert is idempotent: `account_read_record` dedups via its `.Larr_scan`
20-byte compare, so variants that already record their target are unaffected.

## Register safety: `a0` **is** `x10`, and why that is safe *here*

Calling the recorder means loading its argument into `a0`, and `a0` is `x10`. An
equivalent insert in `extcodehashWitnessTail` broke for exactly this reason: it
was placed *after* `runtime_access_account_charge`, where `x10` is live and about
to be saved as continuation state, so the argument setup destroyed it.

This site is on the other side of that call, and the proof is structural rather
than a liveness analysis. In the emitted asm each of the four handler tails reads:

```
  bnez t2, .Lraaw2be_<variant>_target_loop
  la a0, evm_access_seed_addr        <-- ORIGINAL: unconditional write of x10
  la a1, evm_access_account_table
  la a2, evm_access_account_count
  li a3, 100000
  jal ra, runtime_access_account_charge
  addi sp, sp, -32
  sd x10, 0(sp); ...                 <-- the live range that broke the other site
```

The insert goes between the `bnez` and that `la a0`. The instruction immediately
after it is therefore the original's own unconditional write of `x10`, with the
same value — so whatever `x10` held on entry is dead in the *unpatched* guest, not
merely by my reasoning. The only instruction in between is the new `jal`, whose
callee reads `a0` as its argument. **There is no read of `x10` after the insert
before its next write.**

The recorder itself is register-clean: across all four paths it saves and restores
`t0`–`t6`, reads `a0` without writing it, and touches no `s`-register and no
`a1`–`a7`. Its 64-byte frame lies strictly below `sp`. `ra` is dead here — the
next sequence does `jal ra, runtime_access_account_charge`.

Verified against emitted asm at the four sites (loop labels `call_`, `callcode_`,
`delegatecall_`, `staticcall_target_loop`), not against the Lean strings. The same
check confirmed the insert lands *after* the loop exit rather than inside the
byte-reversal loop body — inside, it would have fired 20 times per call.

**Note on size — quote the bytes, not a count.** Two source lines, but the tail is
emitted once per handler, so the change adds **48 bytes of `.text`** (`textSizeBytes`
`0x06727c` → `0x0672ac`), measured by the linker on the relinked ELF.

An earlier draft of this section said "eight emitted instructions", counting the two
source mnemonics × four handlers. That is wrong: `la` is a **pseudo-instruction** that
expands to two real instructions (`lui`+`addi`), so each site is three instructions
and twelve bytes, and four sites give the 48 bytes the linker actually measured —
twelve instructions, not eight. The byte figure is quoted here in preference to any
count because it is what the `RegionMap` pin checks and it cannot be wrong about
pseudo-instruction expansion.

## Why there is a new module, and why the layout files move

**`ChildFrameHandlerTailsGuards.lean` (new).** The added comment pushed
`ChildFrameHandlerTails.lean` to 1515 lines against the hard 1500-line
`Codegen/Programs` cap in `scripts/check-file-size.sh`, which supports no per-file
exception. The guards were extracted to a leaf module that imports the tail
generator (they evaluate its emitted string) and is imported back by
`ChildFrameHandlers.lean`. `check-file-size.sh` and `check-unimported.sh` both pass
locally (2817 modules, 0 orphans), and the guards were re-run from their new home.

**Layout artifacts.** Because this adds `.text`, the generated layout files move with
it. Note the base moved during review: #10846 landed while this was open and both PRs
regenerate the same three files, so the conflict was in generated artifacts only. It
was resolved by **regenerating**, not hand-merging — those files were reset to main's
version and re-derived from source, since a hand-merged symbol table matching no ELF
is the failure mode there. Files: `GuestAddrs.lean`, `RegionMap.lean` (`textSizeBytes` only),
`scripts/asm-fixtures/symbol-addresses.tsv`. Full documented regen was run and
**iterated to a fixed point** — `textSizeBytes` is an input to emission as well as a
record of it, so a one-pass repin is not a convergence check; a second complete pass
changed none of the three generated files. Verified locally:

- `check-region-map.sh`: OK on all four pins (`textSizeBytes=672ac`,
  `committedStorageSizeBytes=cdd800`, `dataSizeBytes=5370`, `bssSizeBytes=1b85cae0`),
  and the symbol TSV matches the ELF;
- `check-asm-to-program.sh`: CLEAN (390 converted defs across 148 files);
- **`.data` did not move** — `dataSizeBytes` is unchanged and
  `GuestAddrs.bnf_le_a` remains `0xa3000ee0`, which an unrelated Bn254 proof pins by
  `rfl`. A data shift would have broken modules unconnected to this change.

## Can adding a read perturb execution-visible state?

Asked because a sibling candidate that adds the same call elsewhere *was* measured
flipping `bv_fail` onto the block-gas branch. So this needs both halves of the
argument — callee and site — not just the callee.

**Callee half.** `account_read_record` is register-clean: across all four paths it
saves and restores `t0`–`t6`, reads `a0` without writing it, and touches no
`s`-register and no `a1`–`a7`. Its memory writes are disjoint from the gas path:

- `account_read_record` writes exactly `tx_account_reads_count`,
  `ACCOUNT_READS_AREA` (`0xa1ea0000`), and `tx_account_reads_overflow`.
- `runtime_access_account_charge` takes its warmth table as **arguments**
  (`a1`=table, `a2`=count, `a3`=capacity), and at this site those are
  `runtimeAccessAccountTable` / `runtimeAccessAccountCount` — disjoint from the
  read log.

Every consumer of those three was enumerated: the count and arena are read only by
`ReadSetsPromote.lean:138` (promotion into the block arena `0xa1ca0000`) and reset
at `:159`; the overflow flag only by `BlockVerdictMtxRuntime.lean:514`. None is a
gas consumer.

**Site half — the one the callee argument does not cover.** A call can perturb
execution even with a clean callee, via the stack or via a value the site needs
live in `a0`. That is site-local and must be argued per site, which is why the
callee argument alone is insufficient. For *this* site it is settled by what is
already there: the instruction immediately after the insert is
`jal ra, runtime_access_account_charge`, and that function opens with
`addi sp, sp, -64` — an identically sized callee frame at the identical point. So
this site demonstrably tolerates a 64-byte callee frame already, before this
change; and `a0` is reloaded on the following line.

Note that `bv_fail=41` is **ambiguous** — it
has two producers, `.Lbv_block_gas_used_over_fail` (non-BAL) and
`.Lbv_recipient_bal_fail` (BAL) — so a bare 41 does not identify which fired.
Pre-registered: if this change ever produces 41 it should be the recipient-BAL
site; the gas site would mean the disjointness argument above is wrong.

The one genuine path by which a read *can* fail a verdict is arena overflow:
`tx_account_reads_overflow` has exactly one consumer,
`BlockVerdictMtxRuntime.lean:514` → `.Lbv_fixed_arena_overflow_fail`. Capacity is
16384 with dedup, far beyond these blocks, but it is a real edge and named here
rather than assumed away.

## Measurement boundary — what this evidence does *not* establish

Stated up front so none of the numbers below are read as more than they are.

1. **The verdict metric is saturated, so it cannot witness this change.** `bv_fail`
   stayed `60` on all 20 cases in both legs. `60` means "digest mismatched", and the
   digest still mismatches after the fix because a second entry is missing. Anyone
   measuring a single-row BAL change off `bv_fail` alone will read a real
   improvement as a null. The evidence here is a **length delta**, not a pass.
2. **One recovered entry, not a fixed block.** The KAT goes shortfall 54 → 27. No
   case in this run goes to zero, and no case flips to PASS.
3. **The population comes from a detector that only sees pushed literals.** ~11
   resolvable `CALLCODE`/`DELEGATECALL` targets. A target arriving via
   `CALLDATALOAD`, `DUP`, or computed on the stack is invisible to it —
   `STATICCALL` resolves only 4 of 70 sites even after widening. So the count is a
   floor on the affected population, not a measurement of it.
4. **Three cases at shortfall 54 did not move at all.** That is a clean negative,
   not a partial failure: their missing entries are not call-family targets, so no
   placement of this change reaches them.
5. **The alignment argument does not depend on any of the above.** The four spec
   citations and the structural fact that `env+696` is the child address (the
   *parent* for `CALLCODE`/`DELEGATECALL`) hold whether or not the sampled corpus
   exercises them.

## Measured result

Paired A/B, spike backend, identical command/filter/limit on both legs. Bases: the
immutable pre-merge snapshot `f7824b3ea` for both legs — that is the form of what is
now on main after #10844 landed as `6adc2f6a3`, so **do not reconcile these numbers
against a main sha.** Guest sha256 differs between legs (parent `7375b58e51ed0339`,
candidate `a5f65da4f8d18654`), and the parent was verified to contain zero
occurrences of the insert *by `git grep` against the branch*, not the worktree.

`bv_fail` cannot see this change: `60` means "digest mismatched", both legs
mismatch, and it stayed `60` on all 20 cases. Reporting that null alone would have
read as the fix doing nothing. The sensitive observables are
`bv_bal_shadow_rebuilt_len` and `bv_bal_shadow_supplied_len`, published to
`OUTPUT+128` / `OUTPUT+136` by `StatelessGuestEpilogue.lean:837-838` and read from
the raw `.output` bytes.

| case | supplied | rebuilt (parent) | rebuilt (cand) | Δ | shortfall |
|---|---|---|---|---|---|
| `d6-g1` (KAT) | 381 | 327 | 354 | **+27** | 54 → 27 |
| `d1-g0` | 412 | 358 | 385 | +27 | 54 → 27 |
| `d1-g1` | 392 | 338 | 365 | +27 | 54 → 27 |
| `d2-g0` | 412 | 358 | 385 | +27 | 54 → 27 |
| `d2-g1` | 392 | 338 | 365 | +27 | 54 → 27 |
| `d3-g1` | 352 | 298 | 325 | +27 | 54 → 27 |
| `d6-g0` | 396 | 342 | 369 | +27 | 54 → 27 |
| `d7-g0` | 396 | 342 | 369 | +27 | 54 → 27 |
| `d7-g1` | 381 | 327 | 354 | +27 | 54 → 27 |
| `d8-g1` | 352 | 298 | 325 | +27 | 54 → 27 |

**20 cases: 10 improved, 0 regressed, 0 overshoot.** Every nonzero delta is
identical, and `bv_fail` distribution is unchanged across legs (3×37, 17×60).

**27 is not an empirical coincidence — it is the exact RLP size of one all-empty
account entry:** `0x94` + 20 address bytes = 21, five empty change lists (`0xc0`
each) = 5, payload 26, one-byte list header (`0xda`) = **27**. So each affected case
gained precisely one entry, confirmed structurally rather than by fitting. Every
shortfall in this file is a multiple of 27 (54 = two missing entries, 27 = one).

**A pre-registered claim that failed, stated as such:** the KAT was predicted to go
to a shortfall of *zero*. It does not — 27 bytes remain, so the digest still
mismatches. The predicted "27" was the *unit*, not the KAT's total, whose shortfall
was 54. This change recovers exactly one of two missing entries.

**Scope limit, as a result rather than an open question:** three cases at shortfall
54 (`d5-g0`, `d5-g1`, `d8-g0`) did not improve *at all*. That is a clean negative —
their missing entries are **not** call-family targets, so no placement of this fix
would recover them. They belong to a different defect and are deliberately out of
scope here.

### The guards were tested for their ability to fail

A guard that passes vacuously is worse than none, because it leaves a record of
protection. So the positive guard was flipped from `= 2` to `= 3` and the build
*had* to fail. The first attempt failed on **syntax** rather than on the guard's
value (a docstring before `#guard` parses as "unexpected token `#guard`; expected
`lemma`") — which means the guard commit as first written was **unbuildable**, and
had only the correct version been run, the red build would have been blamed on the
fix. After fixing it, both directions were verified: at `= 3` the build fails on the
guard *expression*; at `= 2` it passes. Re-verified after rebasing onto current
main, since a rebase can move the tail.

<details><summary>Earlier pre-registration, kept for the record</summary>

Both measured on the immutable snapshot `f7824b3e`, sequenced:

1. **Gate — without this fix.** Previously-missing `CALL`/`STATICCALL` targets
   should *already* be in the block-level read log; `CALLCODE`/`DELEGATECALL`
   targets should still be missing. *If this fails, the second measurement is not
   run.*

   **Not a blanket prediction, and stated before measuring.** The pre-fix
   recording comes from `call_frame_descend`'s prestate fallback, which is gated:
   `ld t0, 88(sp); bnez t0, .Lcfd_prebalance_done` — the tracked read fires **only
   when neither the BAL table nor the live-effect overlay supplied a balance**. So
   a `CALL`/`STATICCALL` target whose pre-balance *was* supplied is expected to be
   missing pre-fix too, and such cases do **not** falsify the gate. (This is also
   why the prediction is stronger on `f7824b3e` than on `main`: with the oracle
   seeding removed by `6f17c7d4a`, `88(sp)` is zero more often, so the fallback
   fires more often.) What *would* falsify it: a `CALLCODE`/`DELEGATECALL` target
   already present in the log, since `env+696` is the child ADDRESS and that is
   the parent for exactly those two variants.
2. **With this fix.** `CALLCODE`/`DELEGATECALL` targets appear.

KAT for *this* change is `sstore_xto_xto_y` (input sha `dedd575f46997b62`), where
the missing entry is a `CALLCODE` target and the block sits at delta 27. It is
**not** `modexp` (sha `761bee0f8254eb20`) — that is a precompile recipient, which
exercises the top-level recipient fix, so this change alone is *expected* not to
move it.

### What the sampled corpus can and cannot show

Over 64 sampled blocks, counted by a PUSH-aware opcode walk (so operand bytes are
never miscounted as opcodes):

| variant | call sites | target resolved | blocks |
|---|---|---|---|
| `CALL` | 62 | 62 | 24 |
| `CALLCODE` | 8 | 7 | 5 |
| `DELEGATECALL` | 4 | 4 | 3 |
| `STATICCALL` | 70 | 4 | 9 |

So there are ~11 resolvable `CALLCODE`/`DELEGATECALL` targets — the two variants
this change is *uniquely* needed for — not one. An earlier pass of this analysis
reported zero `DELEGATECALL` and zero `STATICCALL` targets; that was **entirely an
artifact of the detector**, which only accepted a literal `PUSH20` within 12
instructions. Widening it to any `PUSH1`..`PUSH20` (precompile and short addresses
are pushed narrow) took `DELEGATECALL` from 0 to 4-of-4. The zeros were the
instrument, not the corpus.

**Stated blind spot:** `STATICCALL` still resolves only 4 of 70 sites, consistent
with targets arriving via `DUP` or computed on the stack rather than pushed. Those
are invisible to any static scan of this kind, so no claim here is made about
them.

The justification for this change is the spec alignment above, not these counts —
an alignment fix is correct whether or not the sampled corpus happens to exercise
it.

</details>


pirapira:
**This PR's content is already carried by #10851** (a batch of #10847 + #10850), which is labelled `merge!`, armed with auto-merge, and running CI.

Posting this here because a batch silently transfers a PR's landing path to a different PR, and nothing in this PR's own status shows that — it reads as `BEHIND` and stalled, which is exactly why syncing it looked like the right move. That was a reasonable read of the available information.

Verified that nothing is lost: of the six files this PR changes against main (`ChildFrameHandlerTails`, `ChildFrameHandlerTailsGuards`, `ChildFrameHandlers`, `GuestAddrs`, `RegionMap`, `symbol-addresses.tsv`), **all six are byte-identical between this PR's head `fe88dc7b4` and #10851's head** — so the independent sync regenerated artifacts the batch already agrees with to the byte.

**Please don't sync this branch again.** It stays open as the fallback in case #10851's build fails. Once #10851 lands, this PR will *not* auto-close — GitHub only auto-closes when a PR's head becomes an ancestor of main, and the sync commit moved this head outside the batch — so I'll close it manually with evidence at that point.

pirapira:
landed via #10851 (main `ffada1f25`).

Closing manually because this PR's head `fe88dc7b4` never became an ancestor of main — the independent sync commit moved it outside the batch — so GitHub had no way to auto-close it. #10850, whose head *was* an ancestor of the batch, auto-closed. That asymmetry is about head ancestry, not about content.

Content confirmed in main by per-file hash comparison of every file this PR changed — `ChildFrameHandlerTails.lean`, `ChildFrameHandlerTailsGuards.lean`, `ChildFrameHandlers.lean`, `GuestAddrs.lean`, `RegionMap.lean`, `symbol-addresses.tsv` — **all six byte-identical between this head and `origin/main` (SAME=6, DIFFERS=0)**. The `.text` pin `0x0672ac` from this PR's +0x30 shared tail is in main.

Nothing is lost by closing this.

PR #10853, opened 2026-07-29
fix: bound EIP-8037 status diagnostic tail

### LINKED ISSUE #10848: bv_fail_code: .Lbv_eip8037_gas_fail stores a0+7 with a0 unbounded on the MtxRuntime path, so no diagnostic code is unique
## `bv_fail_code` has a writer with no bound, so **no fail code is reliably unique**

Found while auditing whether diagnostic code 62 was free.

**`.Lbv_eip8037_gas_fail` stores `a0 + 7` into `bv_fail_code`.** `GasGatePrelude` reaches it with an EIP-8037 status of 1–3, giving codes 8–10 as intended. But **`BlockVerdictMtxRuntime:778` and `:781` also jump to that tail after loading the raw header `gas_limit` into `a0`** — so on that path the stored value is `gas_limit + 7`, which is not statically range-bounded.

Concretely: a block with `gas_limit = 55` writes **62**. Any fail code in the whole namespace is reachable from a fixture-controlled header field.

### Why this matters beyond one collision

Every diagnostic conclusion of the form "code N means check X fired" is unsound in the presence of this writer. Tonight that cost real time in both directions:

- An hour was spent discriminating **code 41**, which serves two unrelated arms (`.Lbv_block_gas_used_over_fail` and `.Lbv_recipient_bal_fail`), with neither arm leaving a distinguishing cell.
- Coord then escalated "an added read is changing state a **non-BAL** check reads" on the strength of 41 — wrong, because one of its arms is a **BAL** check at `BlockVerdictFunction.lean:1438`.

A shared tail makes a code ambiguous between two known arms. **An unbounded writer makes every code ambiguous with an arbitrary one**, which is strictly worse and cannot be fixed by choosing a different literal.

### Baseline

Direct literal writers of `bv_fail_code` live only in `BlockVerdictReceiptsTail` and `BlockVerdictStateRoot`; the literals in use are `{1..7, 10..17, 19..61}` (18 and 62 absent). So there is no *fixed* 62 writer — the exposure is entirely the dynamic path.

### Suggested direction (not prescribed)

Bound the value at the point of the store rather than at each caller: clamp or mask `a0` into the range the tail is documented to encode, or give the raw-`gas_limit` path its own sentinel instead of routing it through an arithmetic tail. Either removes the ambiguity for all codes at once.

Filed separately from the 41→62 split, which improves discrimination between two known arms and is unaffected by this — 41 carries the identical exposure today, so the split is no worse and remains worth landing on its own merits.

pirapira:
Closes #10848.

## Change

`bv_fail_code` code 62 is occupied on main, so this repair reserves code 63 as
the invalid-EIP-8037-status sentinel.  The EIP-8037 tail accepts only status
values 1 through 3; all other values record 63.

The occupied-code statement is about main as of this PR: 62 belongs to the
block-gas code landed in #10849.  `BlockVerdictMtxRuntime`'s two raw-header
callers (lines 778 and 781) remain out of this tail's contract.  A future 63
sighting therefore identifies one of those callers; the proper follow-up is a
dedicated code for that path, not treating 63 as a permanent general sentinel.

Both the valid and invalid status paths jump to `.Lbv_zero`, so this is
structurally verdict-neutral: it changes only the diagnostic code on already
rejecting paths.

## Regeneration and gates

- After merging current main (`ffada1f25`), the predicted pin is confirmed:
  `.text=0x672d0` (`0x672ac` from main plus this change's `0x24`), with
  `.data=0x5370`, `.bss=0x1b85cae0`, and committed storage `0xcdd800`.
- The pre-sync two-relink fixpoint had byte-identical extracted `.text` and
  `.data`; the merged-source relink plus generated-file refresh and
  `check-region-map` reconfirm the new pin.
- `GuestAddrs.bnf_le_a = 0xa3000ee0` remains unchanged.
- `lake build`


pirapira:
**This PR's content is now carried by #10857** (a batch of #10853 + #10855), which is labelled `merge!`, armed with auto-merge, and running CI.

Flagging it here because a batch silently transfers a PR's landing path, and this PR's own status still reads `BEHIND` — which looks stalled but isn't. **Please don't sync this branch**; it stays open as the fallback in case #10857's build fails.

Once #10857 lands I'll verify this PR's content in main per-file and close it, manually if GitHub doesn't auto-close (it only auto-closes when a head becomes an ancestor of main).

PR #10855, opened 2026-07-29
refactor: extract CALL-family tail helpers to buy headroom under the line cap

pirapira:
## What

Move the six per-effect helpers used by `basicPrecompileCallTail` out of
`ChildFrameHandlerTails.lean` into a leaf module it imports. Nothing renamed,
nothing rewritten — the definitions move verbatim.

## Spec alignment: this change is **behaviour-neutral with byte-identical emission**

Per the standing rule, this PR takes the second sentence rather than a spec
citation: it cannot move the implementation toward or away from execution-specs,
because it does not change the emitted guest at all.

**Verified, not asserted.** The pre-split ELF was saved, the tree relinked after
the move, and the two compared:

| check | result |
|---|---|
| ELF `sha256` | **identical** (`05189d304385463f…`) |
| `.text` size | unchanged, `0x0672ac` |
| `.data` size | unchanged, `0x005370` |
| generated layout files | **none changed** |

The sha comparison is the real test — equal section *sizes* would be consistent
with reordered bytes, equal *hashes* are not. The untouched symbol TSV is an
independent second signal: a moved definition that altered emission would have
churned it.

**Both legs were built from this base.** An earlier verification of the same patch
used a baseline ELF built before #10849/#10850 landed, and that baseline no longer
matches main — #10849 changed a `bv_fail` immediate (`li t0, 41` → `li t0, 62`),
which alters emitted bytes at *constant* section size. Reusing it would have shown
a spurious difference and made a genuine refactor look like it changed emission. So
the baseline here is main `ffada1f25` itself, relinked, not a carried-over hash.

Because emission is byte-identical, **no layout regen is required** and
`RegionMap.textSizeBytes` stays where #10847 put it.

## Why now, rather than when it next breaks

After #10847's split, `ChildFrameHandlerTails.lean` sat at **1482 lines against
the hard 1500-line cap** in `scripts/check-file-size.sh` — eighteen lines of
headroom, with no per-file exception available.

That tail is where CALL-family producer inserts keep landing, and the
read-surface work tracked in #10836 points at more of them. So the next such
change would have breached the cap, and whoever hit it would have read the
failure as their own fault rather than as inherited crowding.

Moving a block out of a *passing* file is a small change. Doing it under a red
gate that is blocking a fix is the expensive version — this file already cost
that once tonight, at 1515 lines.

**Result: 1482 → 1232 lines, so 268 lines of headroom instead of 18.**

## What moved

Into `ChildFrameHandlerTailHelpers.lean` (new leaf, imported by the tail):

- `callDelegationAccessChargeAsm` — EIP-7702 delegation target access charging
- `precompileValueBalanceGateAsm` — the precompile value-balance gate
- `emitSuccessfulPrecompileValueLogAsm`
- `recordSuccessfulPrecompileValueEffectsAsm`
- `refundSuccessfulPrecompileValueStipendAsm`
- `successfulPrecompileNewAccountStateGasAsm`

`basicPrecompileCallTail` itself stays put, as do the placement guards added in
#10847.

## Gates run locally

- `check-file-size.sh`: exit 0
- `check-unimported.sh`: 2818 modules, 2818 reachable, **0 orphans**
- placement guards from #10847: **re-run after the move**, exit 0 — a guard
  pinned to a neighbouring instruction is exactly the thing that stops pinning
  silently when lines shift, and this is the third time in this line of work that
  lines have moved under one


pirapira:
**This PR's content is now carried by #10857** (a batch of #10853 + #10855), which is labelled `merge!`, armed with auto-merge, and running CI.

Flagging it here because a batch silently transfers a PR's landing path, and this PR's own status still reads `BEHIND` — which looks stalled but isn't. **Please don't sync this branch**; it stays open as the fallback in case #10857's build fails.

Once #10857 lands I'll verify this PR's content in main per-file and close it, manually if GitHub doesn't auto-close (it only auto-closes when a head becomes an ancestor of main).

PR #10878, opened 2026-07-29
feat(codegen): resize BAL builder arenas for six-system-call producer route (GH #10836)

pirapira:
## Why

The BAL builder's capacity docstring states the reservation reflects the Amsterdam spec areas read to date, **must be revisited when a new producer route is understood**, and is a joint upper bound with slack. A new producer route has now been understood: **six system calls** (2 unchecked + 4 checked request calls in `apply_body`, each with a 30M regular-gas budget) are incorporated into the BAL but bypass the block gas counters (and user refund logic), so no derivation from the block gas limit can see them. A sound reject of a block execution-specs accepts is a false reject, so the arenas are resized structurally rather than relying on the overflow path.

## Capacities (user leg + six independent system legs)

All user legs assume the refund-effective allowance 240M = 200M × 1.2; system legs are six independent floors 6×⌊30M/cost⌋:

| Arena | Old | New | Derivation |
|---|---|---|---|
| `balBuilderAccountCapacity` | 76,923 | **140,000** | ⌊240M/3000⌋ + 6×⌊30M/3000⌋ (cold-account cost 3000, `execution-specs amsterdam/vm/gas.py:69-71` — the old divisor 2600 was wrong on both cost and route) |
| `balBuilderStorageChangeCapacity` | 15,384 | **47,522** | user leg ⌊1.5L/10100⌋ = 29,702 (entry costs WARM_ACCESS 100 + STORAGE_WRITE 10,000 = 10,100 gross, so R ≥ 10100N; with state gas S ≤ L and refund cap F ≤ (R+S)/5, regular use R−F ≤ L gives 0.8R − 0.2S ≤ L, hence R ≤ 1.5L) + 6×⌊30M/10100⌋ = 17,820 |
| `balBuilderBalanceCapacity` | 50,000 | **105,000** | ⌊240M/4000⌋ + 6×⌊30M/4000⌋ |
| `balBuilderNonceCapacity` | 16,666 | **35,000** | ⌊240M/12000⌋ + 6×⌊30M/12000⌋ (old value was exactly 200M/TX_BASE with no system headroom) |
| `balBuilderCodeCapacity` | 6,250 | **13,125** | 6×⌊30M/32000⌋ = 5,622 → 13,122, deliberately rounded **up** to 13,125 (slack + a rounder checkable number; stated as such, not as a derivation) |

The full derivation is now in the `balBuilder*Capacity` docstring so the next reader checks the derivation, not the number — which is exactly what was wrong with the old values.

## Layout: `.bss` base moves into the `.data` slack (Option A)

The builder arrays grow 7,589,656 → 16,882,112 B (+9,292,456), which does **not** fit the 1.14 MiB gap between the old `.bss` end and `.sszscratch`. `.data` is reserved 16 MiB (0xa3000000→0xa4000000) but uses only 21,360 B, so `.bss` moves down into the slack:

- `.bss` base 0xa4000000 → **0xa3110000** (margin 1,092,752 B ≈ 1.04 MiB above `.data`'s actual end 0xa3005370), in `Driver.lean`, `RegionMap.lean`, and both literals in `check-region-map.sh`.
- `RegionMap.bssSizeBytes` → **0x1c139580** (ELF-measured; `.bss` now ends at 0xbf249580, leaving ~7.2 MiB to `.sszscratch`).
- The proved sum budget (`CallFrameLayout.lean` `data_gap_bytes`, arenas ≤ `sszScratchBase − dataBase`) is unchanged — neither endpoint moves.

**New total reserved NOBITS vs the 512 MiB ziskemu RAM region** (CODEGEN.md:149): `.bss` 0x1c139580 + `.committed_storage` 0xcdd800 + `.sszscratch` 0x680000 ≈ **484.5 MB** of 512 MiB.

## Out of scope (split, agreed): `storageWritesCapacity`

Raising it needs two adjacent fixed-base regions moved (`storage_writes_area`/`tx_storage_writes_area` at 0xa1fa0000/0xa21a0000, both 2 MiB literals with no gap), the undo journal re-keyed (`MemoryLayout.lean:355`), hardcoded bases in `MemoryLayout.lean:317/319`, and the `AccountWriteMap.lean:624` compile-time coupling — a separate PR.

## Verification (measured on this branch)

- Full four-step regen: `gen-symbol-addresses --build` (3,234-row TSV), `guest-addrs` (1,125 symbols), `--emit-lean` (**GuestImageEntries.lean byte-identical** — entries key on program lengths, not addresses).
- Address-pinned proof casualties from the `.bss` shift: **one module** — `U256MulU64Be/OuterLoop.lean` (`accBase_toNat` + 4 sibling literals) and its `Common.lean` doc, updated to the regenerated `u256m_acc` (0xa468f860).
- `scripts/check-region-map.sh`: **PASS** (all rows OK, TSV matches ELF). `check-guest-addrs`: **CLEAN**.
- Fixed-point check: two consecutive `stateless_guest` emissions byte-identical — emitted `.s` sha256 `a58616de…`, stripped ELF sha256 `fa45372c…`.

Behaviour note: this changes guest memory layout and arena capacities — an intentional behaviour-affecting structural change (sizing fix for a false-reject class), not the behaviour-neutral lane.

## Post-merge-label fix: GuestImage.lean manual repin (commit 23375332b)

CI caught a second pin casualty the local `lake build codegen` never saw:
`EvmAsm/Codegen/Proofs/GuestImage.lean` is a Proofs module OUTSIDE the
codegen executable's import closure, so it was never type-checked locally.
Its `SatWithin` literals (the file's own comment warns the four-step regen
does not touch it) are repinned with ELF-measured values: t7 =
`SatWithin 0xa3110000 0xbf249580` via `satWithin_ramRegion 0xa3110000
0x1c139580`; t7-prime end = 0xbf249580 (the .data end 0xa3005370 does not
move). The recorded rule now lives in `docs/agents/layout-regen-checklist.md`
(commit dd4261e89): **`lake build codegen` is NOT `lake build` for casualty
enumeration — a layout-moving change must run a full build before anyone
claims zero casualties.**

Evidence on the fixed branch:
- **Demonstrably compiled**: full `lake build` EXIT=0 with
  `Built EvmAsm.Codegen.Proofs.GuestImage (7.8s)` in the log (proof by
  module name, not a cache assumption).
- **check-region-map.sh PASS**: all rows OK (text 672d0, data 5370,
  bss 1c139580, committed_storage cdd800), symbol-addresses TSV matches
  the relinked ELF. `check-guest-addrs` CLEAN (1125 symbols).
- **Positive freshness check**: all 1980 `.bss` symbols in the TSV
  (extracted from the relinked ELF) lie inside the new window
  [0xa3110000, 0xbf249580]; zero strays, no old-window leftovers —
  artifact verified against the intended change, immune to cache
  questions.


pirapira:
**Blocked: the `.bss` base move breaks the standalone verdict units' link.** Both problems below reproduce on this PR's head *alone* (I verified against `origin/bal-arena-resize` without my sync merge), so neither is a merge artifact.

**1. Link overlap — needs a layout decision, so I have not touched it.**

```
riscv64-unknown-elf-ld: section .bss VMA [00000000a3110000,00000000a3a4f887]
                        overlaps section .data VMA [00000000a3000000,00000000a36000df]
```

Moving `.bss` down to `0xa3110000` leaves only `0x110000` (~1.1 MB) beneath it, but `zisk_stateless_verdict`, `zisk_stateless_verdict_v2` and `zisk_step2_verdict` each emit ~6.3 MB of `.data` (ending `0xa36000df`). The **guest** links fine because its `.data` is only `0x5370` — which is exactly why `lake build` and `codegen-stateless-link-check` stay green and only `check-build-units-link` catches it. That is the failure mode that script's own header describes.

So the constraint is: `.bss base ≥ max(.data end over ALL units)`, **and** `.bss base + 0x1c139580 < 0xbf980000` (`.sszscratch`). Both are satisfiable — e.g. a base above `0xa3600100` still ends around `0xbf73xxxx` — but picking the value is a layout decision, and a wrong `--section-start` yields a silently wrong ELF feeding the verdict gates. Not a mechanical fix, so it is yours.

**2. `GuestImage.lean` needs its manual bss repin — fix ready, not pushed.**

`lake build` fails first with a bare `Type mismatch` at `GuestImage.lean:219`, exactly as that file's own comment predicts ("the failure does not look like layout drift"). `RegionMap.bssSizeBytes` is `0x1c139580` here but `GuestImage` still carries `0x1b85cae0` / base `0xa4000000`; the four-step regen does not touch that file.

The repin, which I verified builds green locally:

- `t7`: `SatWithin 0xa3110000 0xbf249580` / `satWithin_ramRegion 0xa3110000 0x1c139580`
- `t7'`: `SatWithin 0xa3005370 0xbf249580`
- comment literals `0x1b85cae0` → `0x1c139580`, `0xbf85cae0` → `0xbf249580`, `0xa4000000` → `0xa3110000`

(`0xa3110000 + 0x1c139580 = 0xbf249580`; `data_end 0xa3005370 ≤ base`, and the end clears both `0xbf980000` and `0xc0000000`.) **I am deliberately not pushing it**, because the link overlap blocks the merge regardless and a push would burn a CI cycle on a still-red branch. It also becomes wrong if you choose a different base — so apply it *after* fixing (1), with the base you settle on.

**Local state otherwise:** clean merge with main (no conflicts), regen fixpoint on round 1 with all three step exit codes 0, `.data` unchanged at `0x5370`, and 23/25 gates green — the two failures are `check-build-units-link` and `check-build-parallel`, which is the same defect counted twice since the latter wraps the former.

**Minor, unrelated:** `Driver.lean:69` still documents `.bss` at `0xa4000000` in prose though the flag below it is now `0xa3110000`; and `RegionMap.bssRegion`'s `evidence` string still says `0x1b20b120-byte NOBITS extent`. Neither is checked by any gate.

pirapira:
**Re-checked on `dd4261e89`: blocker 2 is fixed, blocker 1 still fails.**

The `GuestImage` repin in `23375332b` lands `t7`/`t7'` at exactly `0xa3110000` / `0xbf249580` / `0x1c139580` — identical to the values I derived independently, so `lake build` is now **exit 0** on a fresh sync with main (clean merge, no conflicts).

But `.bss` base is unchanged at `0xa3110000`, and the two new commits touch only `GuestImage.lean` and `docs/agents/layout-regen-checklist.md` — the link path and the units' `.data` are untouched. I verified rather than assumed by re-running the gate on the synced tree:

```
check-build-units-link exit=1
  FAIL  zisk_stateless_verdict     — no .elf produced
  FAIL  zisk_stateless_verdict_v2  — no .elf produced
  FAIL  zisk_step2_verdict         — no .elf produced
```

Same overlap as before: `.bss [0xa3110000,…]` vs those units' `.data` ending `0xa36000df`. Keeping `build-broken`. The remaining decision is the one from my earlier comment — a `.bss` base at or above the largest `.data` end across **all** units, still leaving `base + 0x1c139580 < 0xbf980000`.

pirapira:
**Unblocked and armed.** `2af89f169` (per-unit `.bss` link bases) resolves the link overlap I reported — the guest keeps `0xa3110000` while units with large `.data` get their own base, which fixes the constraint rather than relaxing a gate.

Verified on the synced tree (clean merge with main `0b1e0dc8d`, no conflicts):

- `lake build` exit 0
- regen **fixpoint on round 1**, all three step exit codes 0
- **25/25 `check-*.sh` green on the first pass** — including `check-build-units-link` (all three `zisk_*` verdict units now produce an ELF) and `check-region-map`
- `.data` unchanged at `0x5370`; `.bss` `0x1c139580`; `.text` `0x0672d0`

Auto-merge is on.

Issue #10701, opened 2026-07-27
bal: the post-exec block_access_index producer (ulen(txs)+1) is MISSING in the guest - structural gap; the system=0 half is refuted


pirapira:
> **Note:** Written by **Claude Code** (scoper agent). Measurements are read off the emitted guest, `gen-out/regionmap/stateless_guest.s`, built from `main` + #10697.

## Consequence first

**Two of the three forms EIP-7928 defines for `block_access_index` have no producer in the guest.**

- **System attribution cannot be expressed** — the global can never hold `0`.
- **Post-execution attribution cannot be expressed** — nothing ever writes `N+1`.

This is a **completeness** gap in the attribution model, and it is distinct from #10697's **correctness** gap. #10697 fixed the *user-transaction* case: the value was wrong on one dispatch path. This issue is that two of the three cases have no way to be produced at all. It should not be read as more of the same bug.

## The three forms

The BAL builder's own docstring, `block_access_lists.py:305-309`:

```python
block_access_index: BlockAccessIndex = BlockAccessIndex(0)
"""
Current block access index.  Set by the caller before each
[`incorporate_tx_into_block`] call (0 for system txs, i+1 for the
i-th user tx, N+1 for post-execution operations).
"""
```

| value | case |
|---|---|
| `0` | system transactions |
| `i+1` | the i-th user transaction |
| `N+1` | post-execution operations (withdrawals) |

**It is not an `Optional`.** The block-level value is `0` — a real index, not a sentinel and not an absence. So this needs **no special encoding**: it needs *producers* at two boundaries that currently have none. Nobody should design an `Optional` here. (`TransactionEnvironment.index_in_block` *is* `None` for system transactions, but that is a different field on a different object; conflating the two is what produced earlier design guidance calling for a first-class none-value.)

## Every producer in the guest yields ≥ 1

After #10697 there are exactly four stores to `current_block_access_index`, plus the static initialiser:

| site | value |
|---|---|
| `.data` initialiser | `.dword 1` |
| `.s:16092` `.Lbv_mtx_recipient_lookup_resolved` | `bv_mtx_i + 1` |
| `.s:16302` `.Lbv_mtx_is_contract` | `bv_mtx_i + 1` |
| `.s:16658` `.Lbv_mtx_creation_access_done` | `bv_mtx_i + 1` |
| `.s:17729` `.Lbv_stx_checks_done` | `li t1, 1` |

[...]

PR #10880, opened 2026-07-29
bal: feed the N+1 builder on the zero-transaction path — three FRs, one missing call site

pirapira:
## What this fixes

`block_verdict_withdrawal_nonstorage_effects` has **two call sites, and only one of them was followed by the
post-transaction builder feed.**

`BlockVerdictMtxRuntime` (after `.Lbv_mtx_done`) already does the right thing — it sets
`current_block_access_index = bv_tx_count + 1` and calls `account_writes_emit_builder_tx` +
`account_writes_incorporate_tx`, i.e. the **same emitter and incorporate every transaction uses**. The
producer was never missing.

The other site, in `BlockVerdictEoaBodyEffectReconcile`, called the recorder and then only checked `a0`. That
is the path a **zero-transaction** block takes. So the withdrawal credit was recorded into the nonstorage
effect log by `record_nonstorage_effect` and **nothing ever walked it into the builder** — the whole
`balance_changes` account entry was absent from the rebuilt BAL.

This adds the same three lines to that site.

## execution-specs alignment (pinned `e5a8caf1b`)

- `fork.py:917-919` — the post-execution operations run at `block_access_index = ulen(transactions) + 1`.
  With zero transactions that is **1**, which is the index the declared BALs of these blocks carry.
- `fork.py:921` → `:1226` — `process_withdrawals` incorporates like any other producer. EIP-4895 applies the
  credit regardless of how many transactions the block has.

## The two call sites are mutually exclusive — measured, not assumed

This was the one thing that could have made the change wrong: if both sites could run, feeding the builder
here would **double-feed** the `ntx ≥ 1` path and emit a duplicate row. Counted by PC on one guest ELF:

| fixture | txs | site `0x80019178` | site `0x8001d0a0` | `append_balance` |
|---|---|---|---|---|
| 00566 | **0** | 0 | **1** | **0** |
| 00565 | 1 | **1** | 0 | 4 |
| 23725 | 5 | **1** | 0 | 15 |

Exactly one fires per block. Which site is which was identified from the **emitted** code — `0x80019178` is
followed by `ld bv_tx_count` (the N+1 index set), `0x8001d0a0` by `ld exec_nonstorage_effect_overflow`.

## Measured effect — every length predicted before the build, and hit exactly

| fixture | `rebuilt_len` | `supplied_len` | `fail_code` | `append_balance` |
|---|---|---|---|---|
| **00566** | 228 → **264** | 264 | **60 → 0** | 0 → 1 @ bai1 |
| **00346** | 228 → **263** | 263 | **60 → 0** | 0 → 1 @ bai1 |
| **00578** | 228 → **264** | 264 | **60 → 0** | 0 → 1 @ bai1 |

**Three blocks go from failing to fully verifying.** The do-not-move set is unchanged: **00565 at 350**
(the double-feed canary — still exactly 4 balance rows, `bai1=3 bai2=1`, no duplicate), 23725 at 818,
23100 at 554, 11658 at 662, 11659 at 710.

Two fixtures reported because they are **not** clean wins:

- **00560** — the BAL now matches (228 → 264 = supplied), but `fail_code` stays **44**, which it already was:
  a different gate still rejects that block. An improvement, not a pass.
- **00350** — unchanged at `fail_code` 7 with no BAL produced. It fails before the BAL exists, so this change
  cannot reach it.

### This also refutes a previous finding of mine, empirically

I had recorded "withdrawal to a precompile loses its whole BAL account row." **The precompile was
incidental.** The real discriminator is **zero transactions**: four zero-tx fixtures with *ordinary*
recipients (00346, 00350, 00560, 00566) all lost the row, while 00565 (1 tx, ordinary recipient) never did.
00578's recipient being a precompile was a property of the one fixture I had. 00578 closing here settles it —
if the recipient mattered it would still fail while 00566 closed.

## Generated layout

Regenerated to a fixed point (`docs/regenerating-generated-files.md`): `.text` `0x0672d4` → `0x06730c`.
`.data` unchanged, `GuestAddrs.bnf_le_a` still `0xa3000ee0`, and `check-region-map.sh` passes against the
relinked ELF.

## Stacking

Stacked on `scoper/bal-system-write-leak-bai` (#10875), which is already reviewed — this avoids reopening its
review. The two changes are disjoint: #10875 stops the end-of-block system-call *storage* writes being
attributed to transaction 1; this one feeds the *nonstorage* builder on the zero-transaction path. Refs
#10701, #10764.


pirapira:
**Closing as subsumed, not rejected** — this content is now inside #10875 at `9c0aaaa39`, which carries both the system-write discard and the zero-transaction N+1 feed and lands on `main` once.

Why they were combined: four PRs at roughly an hour of CI each, under a rule that enforces up-to-date branches, means **every merge to `main` puts all the others into `BEHIND`.** Landing these two separately cost two main-merges and an extra rebuild cycle for each of the two PRs still queued behind them. Merging this into #10875's branch removed one of those, at no CI cost here since this was already based on that branch and already green.

I offered independent revertability as the price. **The author pointed out it was never available to trade**: this PR's base was already `scoper/bal-system-write-leak-bai`, so it could not land before #10875 anyway, and reverting #10875 after both had landed would already have required untangling this one. The independence was spent when the branch was stacked — my cost analysis priced something that had already been paid.

The coherence argument is the substantive one, and it belongs to the author's own coverage document: **#10875 alone moves 23100 from 559 to 554 — *further* from declared — which reads as a regression in isolation.** Combined, the acceptance set reads correctly: 00566 at 264, 00346 at 263, 00578 at 264, all equal to declared and passing, with the storage residual accounted for as read-exclusion coupling.

Verification on the combined branch, which matters more than the merge:

- `.text` `0x0672d0 → 0x06730c`, **+60** — four bytes for the discard call and fifty-six for the zero-tx feed — with a second relink reproducing it, so a genuine fixed point.
- `check-region-map` passes with `symbol-addresses.tsv` matching; `.data` unchanged; `bnf_le_a` still `0xa3000ee0`.
- The artifact is **byte-identical to `91353ef2f8ce`**, the one all eight acceptance legs were measured on, so no re-measurement is required — **and that identity is meaningful rather than vacuous because a third artifact, `d191c0e4e299` from the discard-only branch, differs.** Two agreeing shas could mean nothing changed *or* both came from a stale source; a third that differs separates those.

Nothing here is being discarded. Review it as part of #10875.

pirapira:
**Correction to my previous comment: this was MERGED, not closed.** It merged into its own base branch, `scoper/bal-system-write-leak-bai`, which is exactly the right mechanism and is what the author did — GitHub records it as merged because that base *is* the target. My "closing as subsumed" framing was wrong; nothing was closed and nothing needed to be. The substance stands: the content now lives in #10875 at `9c0aaaa39`, which lands on `main` once.

Flagging it rather than editing silently, because the distinction matters to anyone auditing later: a merged PR and a closed one leave different records of whether the work was accepted.

pirapira:
Superseded by #10875, which now carries both halves.

Its branch was merged into `scoper/bal-system-write-leak-bai` at `9c0aaaa39`, so this PR's diff against its own base is now **empty** (verified: `git diff scoper/bal-system-write-leak-bai...scoper/bal-zerotx-withdrawal-feed` produces nothing).

**Why combined rather than landed separately:** every merge to main puts the other open PRs behind, so landing these two as one removes a main-merge and a rebuild cycle from each queued PR. The cost would normally be independent revertability — but that was never actually available here, since this PR's base *was* `scoper/bal-system-write-leak-bai`, so it could not land first and reverting the other half would already have required untangling this one.

It also reads better combined: the discard alone moves 23100 from 559 to **554**, *further* from declared, which looks like a regression in isolation. With the zero-tx feed included, the acceptance set is coherent — 00566 at 264, 00346 at 263, 00578 at 264, all at declared and passing (`fail_code` 60 → 0), with the remaining storage shortfall explained and tracked on #10701.

The measured content is unchanged and re-verified: the combined artifact is byte-identical (sha `91353ef2f8ce`) to the one all eight acceptance legs were measured on.

Closing in favour of #10875.

PR #10882, opened 2026-07-29
docs: BAL producer coverage by block_access_index phase (#10701)

pirapira:
## What this is

The phase-by-field coverage matrix for EIP-7928's three producer phases, with **every cell a counted trace
observation** rather than a source reading. It answers "which BAL fields does the guest emit at which
`block_access_index` phase" and leaves exactly one cell open.

| phase | `storage_changes` | `balance_changes` | `nonce_changes` | `code_changes` |
|---|---|---|---|---|
| **BAI 0** — pre-block system | ✅ direct builder append | n/a | n/a | n/a |
| **BAI 1..N** — per transaction | ✅ | ✅ | ✅ | ✅ |
| **BAI N+1** — post-execution | ❌ **the only hole** | ✅ | ✅ | ✅ |

`n/a` is measured, not assumed: over a systematic every-20th sample of the 26,105 stateless inputs (1,305
decoded), account entries carrying a BAI-0 change number **2,606 for storage and 0 for balance, nonce and
code**.

## Why the unit is phase × field

Enumerating `jal ra` targets is a false-positive machine — 195 hits on `rlp_walk_next` alone, a pure helper
owing no follow-up. The obligation is **per phase and per field**, because both defects found in this area
(#10875, #10880) were *a phase that records with no emit for that phase* — and both were the same shape:
**one recorder, two call sites, only one carrying the follow-up.**

## It records the CLEAN checks too, so they aren't repeated

- `bal_emit_storage_changes` — two call sites, one is `BalSerializerMeasureProbe` off the verdict path ⇒ **one
  production caller**.
- `account_writes_emit_builder_tx` — three sites, **all three** followed by `account_writes_incorporate_tx`.
- `record_nonstorage_effect` — **25** sites, too many to audit individually and covered only *collectively* by
  the matrix, so the residual risk is a site recording outside any phase. That reduces to one comparison of
  commit indices (*does any record happen after the last emit?*) and there are **no orphans** on 0-, 2- and
  5-tx fixtures. **Emits = ntx + 1 in every case**, the `+1` being the N+1 emit.

## Acceptance set, pre-registered

Exact targets for closing the remaining hole, so the post-fix measurement is a yes/no rather than an
interpretation. Each shortfall is fully accounted as **(missing N+1 changes) − (spurious reads)**, with a
re-encode identity control passing before every model. Includes the read-exclusion coupling
(`block_access_lists.py:549-552`) that makes **over-long a distinguishable failure mode** — a fix that emits
the changes but fails to re-exclude the reads lands 2 bytes *over* declared on 23200, not short.

## Three trace-reading notes that cost time to learn

- **`bv_fail_code == 60` is `.Lbv_bal_digest_mismatch`** (`BlockVerdictReceiptsTail.lean:323`) — a *failing*
  verdict, not a validity marker. Equal `rebuilt_len`/`supplied_len` at 60 is an **equal-length content
  mismatch**.
- The two `block_verdict_withdrawal_nonstorage_effects` sites are **at most one** per block — neither fires
  when the block is rejected earlier (00350). A zero on both is not evidence of a missing call.
- A moved layout invalidates a PC list **silently**, and a stale `.bss` address still reads **live memory**.

## Scope

Docs only, branched off `main` so it can land independently of #10875/#10880. Per `docs/README.md` this is a
category-2 record — it states its date and the ELF it was measured on (sha `91353ef2f8ce`), and defers to the
Lean sources and drift-guard scripts over any figure in it. Refs #10701, #10695, #10764.


pirapira:
Reviewed, and I'd land this as-is. It is a better document than the matrix I asked for, and three things in it are worth naming so they survive future edits.

**Every cell is a counted trace observation with its producer named**, not a source reading — and the header says outright that the numbers describe the state of the world at a date and that the Lean sources and drift-guard scripts outrank any figure here. That is the fix for the exact failure this document could otherwise become: `PLAN.md`'s write-map status was true when written and decayed silently into "zero callers", which nearly caused the containers to be rebuilt from scratch. A record that dates itself and names its ELF (`91353ef2f8ce`, `.text 0x06730c`) cannot decay into a false assertion — it can only become old, which is visible.

**The `n/a` cells are justified on evidence rather than left as an assumption.** 2,606 account entries carrying a BAI-0 storage change against **0** for balance, nonce and code, across 1,305 decoded inputs. Proving a negative cell empty is what turns a suspicion list into a bounded scope, and it is the step most people skip.

**The clean checks are recorded so they are not re-run.** One production caller for `bal_emit_storage_changes`; three-for-three on the `emit`/`incorporate` pairing; no orphan among `record_nonstorage_effect`'s 25 sites, established by the commit-index comparison rather than by auditing 25 sites. Your own line for it is the right justification: a clean check that is not written down gets re-run by the next person.

Two specifics I checked and want to highlight rather than change:

- **The residual is per-fixture, not a formula** — 23100 is `+1` for *three* dropped rows because only one of its dropped slots was also read. Anyone eyeballing the table will expect `+3` and conclude the model is wrong; the sentence that pre-empts that is doing real work.
- **Over-long as a distinguishable failure mode** is the most useful line in the acceptance set. A fix that emits the N+1 changes but fails to re-exclude the reads lands **2 bytes over** declared on 23200 rather than short, so the failure has a signature instead of just being "not the target". That is what makes this a yes/no test rather than an interpretation.

The internal arithmetic is consistent where I checked it: 23725 drops 8 → 818 with residual 0 and targets its full declared 843; 23100 drops 3 → 553 with residual +1 against the measured 554.

The three trace-reading notes at the end — `fail_code 60` being `.Lbv_bal_digest_mismatch` rather than a validity marker, the withdrawal-effects sites being *at most* one so a double zero means the block died earlier, and a stale `.bss` address reading live memory — are each a trap that cost time today. Preserving them next to the numbers they would corrupt is the right place for them.

Docs-only and branched off `main`, so it lands independently of #10875 and #10880. Labelling once CI is green.

Issue #10876, opened 2026-07-29
codegen: runtime dispatcher clears cross-tx AccountState pending journal, losing coinbase fee credits


pirapira:
## Defect

`runtime_dispatcher_call` unconditionally clears the cross-transaction
`account_state_pending` journal. This discards a coinbase fee credit appended
after transaction N's dispatcher return before transaction N+1 can use it as
the live balance baseline.

The comment in `BlockVerdictMtxRuntime.lean` says the dispatcher resets only
its own pending overlay while AccountState survives the multi-transaction
loop. That is false for `account_state_pending`.

## Measured trace

Diagnostic guest: `/var/tmp/coinbase-diag/stateless_guest.elf`.
Fixture: `00337_test_bal_cross_tx_storage_write_fork_Amsterdam-blockchain_test-tx2_reverts_to_zero__b0`.
Runner: `/home/yoichi-bkp/evm-asm2/scripts/spike/spike_run`, SHA-256
`c6ac09fa72c2103b6d3e390b779b022ef7f220e5a39bb5f667114a89d78e5868`.

The tx-2 diagnostic executed at PC `0x80018fd4` and wrote
`account_state_pending_count = 0`; its `account_state_find` result was zero at
PC `0x80019018`.

The same commit log records six executions of:

```
PC 0x80039cd4: sd zero, 0(t0)
t0 = 0xbdf41248 (account_state_pending_count)
```

`0x80039cd4` is in `runtime_dispatcher_call`. The full trace is retained at
`/var/tmp/coinbase-diag/00337.commit`.

## Consequence

`blockVerdictMtxCoinbaseFeeEffect` records its credit after runtime dispatch
and after the normal AccountState commit. The following runtime dispatch
erases that pending record before the next coinbase lookup. Consequently the
lookup falls back to pre-state and emits only the current transaction's credit
instead of the cumulative coinbase balance.

This explains the seven equal-length balance mismatches: the rebuilt value at
BAI k equals the declared value at BAI k minus the declared prior credit.

The repair must make the committed coinbase fee survive dispatcher entry;
do not merely make the dispatcher reset conditional without enumerating its
current reset obligations.

pirapira:
## Disposition: **its named fixtures now pass — but that is weaker than "the mechanism is gone"**

PR #10885 references this issue without a closing keyword. Re-tested the fixtures this issue names, each
length-asserted before any diff:

[...]

PR #10885, opened 2026-07-29
fix(codegen): persist MTx coinbase fee account state

### LINKED ISSUE #10870: BAL: the COINBASE's balance_changes entries after the first are DELTAS not post-balances - dedicated coinbase producer never accumulates (generic path is correct at the same multiplicity)
Split out of #10865 (mechanism 5 of five distinct `balance_changes` defects), because this one is exactly characterised and independently fixable.

## The mechanism, stated without hedging

**The first `balance_changes` entry per account is a correct post-balance; every subsequent entry is the DELTA from the previous post-balance instead of the post-balance itself.**

**The guest computes the per-transaction increments correctly and fails to accumulate them — a running total never carried forward.**

## Confirmed at multiplicity 5 — four independent identities holding simultaneously

**23725**, coinbase `2adc2566…`, five balance changes, reconstruction validated at **818 bytes**:

| entry | declared | rebuilt | identity |
|---|---|---|---|
| tx-1 | 1,449,456 | **1,449,456** | correct **post-balance** |
| tx-2 | 2,877,342 | **1,427,886** | = 2,877,342 − 1,449,456 |
| tx-3 | 3,658,680 | **781,338** | = 3,658,680 − 2,877,342 |
| tx-4 | 4,752,267 | **1,093,587** | = 4,752,267 − 3,658,680 |
| tx-5 | 5,845,992 | **1,093,725** | = 5,845,992 − 4,752,267 |

**Four of four, each exact to the wei.**

## Also at multiplicity 2, in a different fixture family

**03828** `test_selfdestructing_initcode_preserves_balance` (validated 529 B), same coinbase: tx-1 matches at 49,628,148; tx-2 emits **406,260** where declared is 50,034,408 — **and 50,034,408 − 49,628,148 = 406,260.**

**Six exact identities across two rows, two fixture families, and two multiplicities.**

## Why the multiplicity-5 case is what makes this non-arguable

**Four independent identities must hold simultaneously, and a coincidence cannot survive that.** The multiplicity-2 case had *one* identity and would not have been leaned on alone. **This is the difference between an identity that fits and an identity that predicts.**

## ⚠️ Invisible on most of the field's population

**367 of 514 declared balance entries are the only entry on their account**, so this defect **cannot be seen on 71% of the field** — the first entry is always correct. Another defect that **only multiplicity reveals** (cf. #10866, and the `code_changes` multiplicity blind spot in #10868).

## ⚠️ This does NOT explain #10869, and the two are not merged

On 07620 the sender's values are `10⁴² + 1,285,200 / 2,570,400 / 3,855,600` — **of order 10⁴², where deltas would be small.** Two **positional** defects, not one; the shared property is only that *the first entry is handled differently*.

## ⭐ Fix criterion — one sentence, no sweep

**Every entry must be a post-balance.** Testable on **23725 alone**, where **four positions must all change.**

pirapira:
## Summary

Persist the MTx coinbase-fee account-state effect at the transaction boundary.

`runtime_dispatcher_call` commits and clears the pending account-state journal
after each transaction.  The coinbase fee path previously recorded its effect
immediately before that call, but did not commit it, so every earlier
transaction's coinbase update was discarded when the next transaction began.
This adds `account_state_commit_pending` after
`blockVerdictMtxCoinbaseFeeEffect` and before `read_sets_incorporate_tx`.

`StorageReadLog` is clarified to document the existing successful normal/MTx
post-coinbase promotion paths and the intentional prep-halt clear.

This closes #10870 and #10876.  #10690 and #10808 appear to be probable
earlier reports of the same transaction-loop durability mechanism.

## Full deterministic same-base A/B

After #10875 landed, I rebuilt both legs from the same base
`0fa9bc70b` and swept the complete 26,104-row manifest.  The frozen parent and
candidate ELFs have distinct whole-file SHA-256s (`1998360ed74746e0…` and
`d0a37211025ebec6…`); their `.text` sizes are respectively `0x6730c` and
`0x67318`, and both `.text` and `.data` differ byte-for-byte.  The runner was
`eaf534e5017e71ff…`; the patched shared instrument was
`6375c5e601d67ca…`.  Both are named because the instrument was hardened during
this run to resolve its runner through `SPIKE_RUN` and fail before workers start
when the path is unavailable.

| artifact | valid accepts | false rejects | correct rejects | false accepts |
| parent `0fa9bc70b` / `1998360ed74746e0…` | 14,864 | 10,204 | 1,036 | 0 |
| repair `b8867067b` / `d0a37211025ebec6…` | 15,052 | 10,016 | 1,036 | 0 |

The join is sound at 26,104 rows per side with no missing labels.  Exact
transitions are 188 FR→OK, 10,016 FR→FR, and 15,900 OK→OK; there are zero
OK→FR transitions and zero false accepts.  The seven known coinbase controls
`00337`, `01085`, `23014`, `23015`, `23032`, `23046`, and `23055` are among
the clearances.

The parent and candidate TSVs each have 26,105 lines including their header.
This is a deterministic full enumeration, so no sample seed applies.  The
candidate launch was guarded on the parent TSV's exact line count, and the
comparison was guarded on the candidate's exact line count; neither partial
output was eligible for interpretation.

The 188 clearances are disjoint from #10875's independently measured 47-row
clearance: #10875 moved the baseline from FR=10,251 to FR=10,204, while this
repair moves that new baseline from FR=10,204 to FR=10,016.  `23725` remains
FR→FR as predicted: #10885 repairs its coinbase value defect, while #10886's
separate storage-row correction is also required for that fixture to clear.

### #10870 ladder check

On `00337`, the parent diagnostic reports `bv_fail=60`, shadow status `1`, and
rebuilt/supplied BAL lengths `370/370`.  The repair reports `bv_fail=0`, shadow
status `0`, at the same `370/370` lengths.  Status `0` is the full rebuilt-BAL
versus declared-BAL digest equality, not a count comparison: the BAI-k entry
therefore matches its declared BAI-k value rather than a prior-BAL delta.  The
equal length is expected: this defect wrote a wrong value in a fixed-width
field, so the digest—not a length counter—is the observable that must move.

## Regeneration

The candidate was regenerated after merging `0fa9bc70b`: serialized passes
converged at `.text` `0x67318`, `.data` `0x5370`, `.bss` `0x1c139580`, with
`GuestAddrs.bnf_le_a = 0xa3000ee0`.  `check-region-map.sh`,
`check-asm-to-program.sh`, and `LAKE_ARTIFACT_CACHE=false lake build` pass.

The cache-mode `check-axioms.sh` gate was not run locally: cache materialization
does not provide the oleans required by `lake env lean` (known #10537).  CI
runs it outside cache mode.


pirapira:
Reviewed. **The diff is right and minimal, and I'd label it as soon as CI is green.** One change to the claims below — it strengthens the PR rather than adding work.

## The diff

One source line — `jal ra, account_state_commit_pending; bnez a0, .Lbv_mtx_bail` — plus the `StorageReadLog` contract update, plus the regenerated consequences. `textSizeBytes` moves `0x0672d0 → 0x0672dc`, exactly **12 bytes**, consistent with one range-expanded call (`auipc`/`jalr`) and one branch. Nothing else in the source diff.

**The contract update is exactly what was needed**, and specifically this sentence:

> This safety relies on that explicit clear, not on the helper being unreachable after a revert.

That is the point. The old docstring's guarantee was *unreachability*; after this change the guarantee is *a zero `account_state_delete_count` on the preparation-halt arm*. Those are different properties with different failure modes — unreachability survives someone later setting that counter and a zero does not — so writing down which one now holds is what stops this becoming another comment that asserts the opposite of what the code does. #10876 exists because of exactly that failure mode; not reproducing it here matters.

## Change: this **closes** #10870, and the evidence is already in the PR

The body says *"Related: #10870 (producer shape)"*. I asked for a ladder re-check to decide between `Related` and `Closes`, and on reflection **the check is already discharged by the result you have** — no new measurement needed:

1. The seven fixtures are #10870's rows — the equal-length residuals whose coinbase `balance_changes` carried `declared(BAI k) − declared(BAI k−1)` instead of `declared(BAI k)`.
2. Each goes parent `success=0` → candidate `success=1`, and a passing verdict entails **BAL digest equality**, which entails the rebuilt coinbase `post_balance` bytes being **identical to declared**.
3. Declared carries cumulative post-balances. Therefore rebuilt now carries cumulative post-balances.

So the cumulative-versus-increment behaviour is gone on every known instance of it. That is the ladder check, satisfied transitively by digest equality — the same argument that discharged the "earliest-BAI entries must not move" guard, since a fixture that fully passes cannot have a wrong coinbase entry at *any* BAI.

Please state it as **Closes #10870** and **Closes #10876** — the observable and its mechanism — and keep #10690 / #10808 as probable duplicates, ideally with a line saying which measured cause subsumes them. The reason to be firm about this: a PR that under-claims leaves an issue open describing a defect that no longer reproduces, and stale open issues sent two lanes hunting already-fixed mechanisms twice today. Four issues collapsing onto one measured cause is the better outcome and it is what you actually demonstrated.

## What carries the PR, for the record

- **188 fail→pass and ZERO pass→fail across the full 26,104-row manifest**, with the reverse list empty and named — a per-fixture no-regression proof rather than an aggregate.
- `.text` and `.data` byte-comparing exactly to the swept candidate, so the A/B carries to current `main` as a **measured** fact rather than an assumption. That is the second time today two section shas replaced a full re-validation cycle.
- The FR decomposition annotated: 14,817 valid accepts, 10,251 FR, **1,036 correct rejections** — so `FR=10,251` is the figure to quote and the raw 11,287 is not a failure count, since it folds in cases where the guest correctly rejected.
- `FA=0` on both legs at full coverage rather than extrapolated from a sample.

Issue #10866, opened 2026-07-29
BAL: storage changes at the post-exec BAI are never produced - presents as keeps-only-the-first (byte-exact: this is #10701, not a per-slot list defect)


pirapira:
## Measured: the guest omits a storage change whose new value is ZERO

**Fixture 23725** (`test_valid_multi_type_requests`), `bv_fail 60`, rebuilt 818 vs declared 843. Rebuilt BAL reconstructed from the commit-log byte stream (`keccak_absorb`, one byte at a time at PC `0x8002c664`), **length-asserted at 818 before diffing.**

Identical address **set and order**. First difference at offset 2 — the outer list header — then **four system-contract accounts all differing in `storage_changes` with the same shape:**

```
rebuilt    e6 c5 01 c3 [c20401]            c5 03 c3 [c20401]           ...
declared   ec c8 01 c6 [c20401 c20680]     c8 03 c6 [c20401 c20680]    ...
```

`SlotChanges` is `[slot, [changes]]` and each change is `[tx_index, new_value]`. **The declared list holds TWO changes per slot — `[4, 1]` and `[6, 0x80]` — and the guest emits only the first.** In RLP `0x80` is the empty string, i.e. **integer zero**.

**⇒ The missing item is a storage change at `tx_index 6` writing ZERO, on four different system-contract accounts in one block.**

## ⛔ Why this class read as "ALL COUNTS AGREE": the column counts SLOTS, not CHANGES

`declared_storage_changes` **counts slots**. Same slots, same order, same count — while the number of **changes within** a slot differs. **A missing change inside a slot is invisible to a slot-level count.**

So the 18-row all-counts-agree class is **not one defect**: at least one member is a defect that a **mis-read column name** concealed.

## Measured vs inferred — three readings fit and are NOT discriminated

**Measured:** four accounts, identical shape, every missing item is `c20680` = `[6, ZERO]`.

**Not discriminated:** whether the rule is
1. **skip a zero value**, or
2. skip `tx_index 6`, or
3. skip the **second** change in a slot.

All three fit this row. The zero-value reading is merely the most natural — a guest treating a write of zero as a delete or no-op is a common bug shape. **Discriminating needs a row where a zero write is the ONLY write, or where a non-zero second change survives.** That is a query with the per-entry diff.

## ⚠️ If the zero-value reading holds, the blast radius is far larger than four rows

[...]

PR #10886, opened 2026-07-29
feat(bal): produce storage_changes at the post-execution block_access_index (GH #10866, GH #10701)

pirapira:
Closes the **last open cell** of the phase×field BAL producer matrix (GH #10701): EIP-7928 has three producer
phases, and `storage_changes` was emitted at BAI 0 and BAI 1..N but **never at BAI N+1** (GH #10866).

## execution-specs alignment (pinned commit `e5a8caf1b`)

| line | what it says |
|---|---|
| `fork.py:917-919` | `block_access_index = ulen(transactions) + 1` for the post-execution operations |
| `fork.py:923` | `process_general_purpose_requests` — the end-of-block system calls run under that index |
| `state_tracker.py:858-859` | each checked system transaction is incorporated, promoting its writes |
| `fork.py:1226` | `process_withdrawals` incorporates at the same boundary |
| `block_access_lists.py:669-676` | the net-zero filter, whose baseline is the value at the **start of the incorporating transaction** |

That last line is why this PR is about **placement**, not about the index.

## What it does

`replay_system_storage_writes_at_bai` re-presents the end-of-block system calls' storage writes from
`bv_system_storage_log` into the tx-level write map, then calls `write_sets_incorporate_tx` (which emits,
merges and clears, in that order). The side arena already held those rows **stamped N+1** —
`capture_system_storage_exec_rows` must keep a durable record because the caller restores the live log count —
so **no new arena is added**: `.data` and `.bss` are byte-unchanged.

## The placement was measured, not reasoned

Emitting in the requests phase, where the writes are still in the tx map, is the obvious home. It was
implemented and measured: **1 of 3** declared N+1 rows on 23100 and **0 of 8** on 23725.

Reading the pre-state of every declared storage-change slot out of the witness explains it exactly: **every**
declared N+1 row in both fixtures is `pre=0 → post=0` **with a transaction writing `1` to the same slot
first**, so against the block pre-state all are net-zero and filtered. The one row that did appear (23100
slot 0, `0x24 → 0x26`) is the only N+1 slot in either fixture with **no transaction partner**.

⇒ The net-zero baseline at N+1 is the **post-transaction** value. A correct BAI stamp on a row computed
against the wrong baseline is not a correct row. That placement is now **pinned shut by a negative `#guard`**
with the measurement recorded at the site, so it cannot be silently reoccupied.

## Both post-exec sites carry it

`blockVerdictEoaBodyEffectReconcile` and the `.Lbv_mtx_done` block are **mutually exclusive** — exactly one
fires per block — so one site covers one arm: **23100 reaches the first, 23725 the second**. Third instance of
the one-recorder-two-sites shape after GH #10875 and GH #10880, so both were written in the same change rather
than waiting for a fixture to expose the second. Each occurrence is pinned by a `#guard` in its own module,
paired **with** the incorporate that consumes it (the replay alone is well-formed and silently produces
nothing).

## Pre-registered acceptance — all five exact, four now pass outright

Registered *before* the build, so this is yes/no rather than interpretation.

| fixture | txs | before | after | declared | `fail_code` |
|---|---|---|---|---|---|
| 23100 | 1 | 554 | **566** | 566 | 60 → **0** |
| 23725 | 5 | 818 | **843** | 843 | 60 — equal-length content residual, separate defect |
| 23200 | 0 | 236 | **246** | 246 | **0** |
| 23260 | 0 | 252 | **264** | 264 | **0** |
| 04460 | 0 | 228 | **233** | 233 | **0** |

23725 now carries all 8 of its declared BAI-6 rows and is length-exact; its remaining mismatch is content at
equal length, which is a different defect and was present before.

## Do-not-move set, re-measured on this artifact

Every pin held, so no attribution relink was needed: **00565 350**, **11658 662**, **11659 710** (the latter
two still over declared by the separate GH #10645 defect), **00566 264**, **00346 263**, **00578 264** all
passing, **00350** rejected earlier with `rebuilt_len` 0.

## Layout

`.text` +0xd0 (208 B) → `0x0673dc`, **ELF-measured** (an earlier draft of this change measured `0x14`, which
would now be a confident wrong literal). `.data`/`.bss` byte-unchanged. Fixed point reached (byte-identical
re-emission), full `lake build` clean, `check-region-map.sh` PASS, GuestAddrs regenerate-and-diff stable,
`check-forbidden-tactics` and `check-file-size` clean.

### 23725's remaining `fail_code` 60 is a **separate, already-filed defect** — not this fix falling short

Equal length with a digest mismatch is the **content** class. It is now identified exactly, not merely
classified: the rebuilt BAL recovered from a commit trace (843 bytes, matching `rebuilt_len`) differs from
declared in **12 bytes, 4 runs of 3**, all in the value field of the **`fee_recipient`'s** `balance_changes`,
and `declared[n] = declared[n−1] + rebuilt[n]` holds on **all five** entries. That is **GH #10870** — the
coinbase's balance entries after the first are deltas rather than post-balances — confirmed arithmetically on
five points. Nothing this PR touches, and it was failing before.

So: **five targets hit, four passing outright, and the fifth blocked by a defect this change was never going to
reach.** What this PR did to 23725 was remove the length gap that was *masking* #10870.

Follow-up: GH #10701's matrix document (PR #10882) still shows this cell as the hole; it should be flipped
once both land.


pirapira:
Reviewed. **Approving on content**; labelling in queue order behind #10885, whose 26,104-row A/B is mid-flight and would be voided by main moving.

## The result, and why the row counts are what make it credible

All five pre-registered targets hit **exactly** — 23100 → 566, 23725 → 843, 23200 → 246, 23260 → 264, 04460 → 233 — with four passing outright and 23100 going `fail_code 60` → 0. The targets were registered *before* the fix existed, with a closed byte account per fixture, so this is a yes/no rather than an interpretation.

The corroboration matters more than the totals: **three rows at BAI 2 on 23100 against three declared, eight at BAI 6 on 23725 against eight.** A baseline subtly wrong in the way that was warned about would land *between* the two totals and would show as a row count under declared even where a length coincided. It doesn't. Per-row and per-byte agree.

## The placement is the whole fix, and the earlier attempt is why we know

A first version emitted in the pre-loop requests phase and measured **1 of 3** rows on 23100 and **0 of 8** on 23725. One rule explained both: it added only the N+1 rows on slots no transaction touched. The spec records `0 → 1` at the transaction and `1 → 0` at N+1, so the second change is a change *only relative to the first* — meaning **the N+1 net-zero baseline is the post-transaction value**, which an emit placed before the loop cannot see.

Two consequences visible in this diff and worth not losing:

- **`incorporate`, not `emit`.** The original argument for emit-only was that incorporating seeds the block map *while transactions run* — the un-spec-faithful state #10875 removed. At a post-loop placement there is no transaction loop left to pollute, so that argument expires with the placement. I asserted otherwise and was wrong; the author diverged and was right.
- **Both post-exec sites carry the feed**, because they are mutually exclusive at run time. That follows from the earlier at-most-one measurement rather than from reading the source, and it is the reason a single-site version would silently miss the zero-transaction arm.

## `RegionMap`: no new data, which is what makes the layout claim hold

The only change is `textSizeBytes` `0x06730c → 0x0673dc` — `+0xd0`, 208 bytes — plus its docstring. **No new region, no new data object.** Verified independently: all 1,597 changed symbol rows are in `.text`, nothing with an address outside `.text` moved (checked by section column *and* by address column), and both `.data` `0x5370` and `.bss` `0x1c139580` are unchanged in size.

That is load-bearing beyond this PR: it means the N+1 storage phase needs **no holding arena**, because `bv_system_storage_log` already carries the rows with `txindex` stamped N+1 — a durable side arena built for a different consumer. It removes 3.57 MiB from the open memory-map decision.

⭐ And the docstring on that constant is the right form: it states the value is **ELF-measured rather than derived**, and says why — the constant is an *input to emission*, so a hand-computed value would be reflected back by the next relink and look confirmed. It records that an earlier draft measured `0x14`, which would now be a confident wrong literal. That is the trap that cost this repo a morning; documenting it at the constant is where the next person will actually read it.

## The remaining 23725 failure is expected, and is filed

23725 hits its 843 target and still reports `fail_code 60` at **equal** length — the content class, identified as **#10870** (coinbase cumulative-versus-increment), which #10885 fixes. Codex measured 23725 as unchanged in its control set on a base without this PR.

So the registered prediction is that **23725 fails for two independent causes and neither PR closes it alone**: this one fixes its storage rows, #10885 fixes its coinbase values. Together they should close it. Falsifiable both ways — if it still fails with both landed there is an unnamed third cause; if it passes with only one, one attribution is wrong. Please keep that sentence in the body, because a reader seeing a target hit exactly *and* a still-failing fixture has every reason to read it as falling short.

pirapira:
## ✅ Joint prediction with #10885 **confirmed** — 23725 now passes

Registered before the run, on the merged branch (main `5f86a5b11` = #10885, plus this PR):

> 23725 fails for **two independent causes** — its N+1 storage rows (this PR) and the coinbase
> cumulative-vs-increment defect (#10870, fixed by #10885) — and **neither PR closes it alone.**

Measured: **`rebuilt_len` 843 = `supplied_len` 843, `fail_code` 0.** Both attributions were right and there is
**no third defect**.

**All five acceptance fixtures now pass** on this artifact:

| fixture | txs | this PR alone | with #10885 | declared |
|---|---|---|---|---|
| 23100 | 1 | 566, `fail_code` 0 | 566, **0** | 566 |
| 23725 | 5 | 843, `fail_code` 60 | 843, **0** | 843 |
| 23200 | 0 | 246, 0 | 246, **0** | 246 |
| 23260 | 0 | 264, 0 | 264, **0** | 264 |
| 04460 | 0 | 233, 0 | 233, **0** | 233 |

Do-not-move set unchanged: 00565 350, 11658 662, 11659 710, 00566 264, 00346 263, 00578 264, 00350 `rebuilt_len`
0. None of the three 60-class pins appears in #10885's 188-row forward list, so they are **current, not stale**.

### Merge and layout

`textSizeBytes` **measured** at `0x0673e8` after the relink. `0x067318 + 0xd0` is the obvious arithmetic and
agrees — recorded here because it was measured rather than summed; that literal is an *input* to emission, so a
computed value is reflected back by the next relink and looks confirmed.

Generated files taken from main wholesale and regenerated to a fixed point. **The fixed point was verified with
a discriminating control**: two consecutive emissions identical **and** the pre-merge artifact differs. Two
identical emissions on their own are satisfied by two cache hits.

Full `lake build` clean, zero casualties.

### Note on where `rebuilt_len` comes from

For anyone reproducing these numbers: they are read from the **production** `stateless_guest` ELF, not an
instrumented build — `bv_bal_shadow_rebuilt_len` at `OUTPUT+128` and `bv_bal_shadow_supplied_len` at `+136`
(`StatelessGuestEpilogue.lean:837-838`, set at `BlockVerdictReceiptsTail.lean:285`), alongside `bv_fail_code` at
`+112`. Grepping for `rebuilt_len` alone misses them. Corroborated three ways: the `+136` value equals the
declared BAL length from an offline RLP decode, and the `+128` value equals the byte count absorbed into
`bal_serializer_rebuilt_ctx` in a commit trace.

Issue #10887, opened 2026-07-29
BAL: code_changes values are emitted as ALL ZEROS with the correct length - invisible to every length and multiplicity check


pirapira:
## Summary

The guest emits `code_changes` values as **all zero bytes with the correct length**. The BAL is well-formed and
byte-length-identical to declared, so this presents only as a digest mismatch at equal length (`fail_code` 60) —
which is why it survived in the "content unlocalised" class of #10830.

## Measured — two fixtures, two different code lengths

Rebuilt BAL recovered from a spike commit trace and diffed against the declared BAL from the fixture. Artifact:
stripped ELF sha `e5c325119235`, `.text` `0x0673e8` (main `5f86a5b11` + PR #10886), measured 2026-07-29.

| fixture | account | declared `code_changes` (BAI 1) | rebuilt |
|---|---|---|---|
| `12925_test_scenarios…` | `1b30fbc3cef96eb8894dd1ac963a2e96b9ff9967` | `3060005260206000f3` (9 B) | `000000000000000000` |
| `13000_test_scenarios…` | `113b7cbc9c28c5ed7289b8b9a10162d5bc1c3543` | `73c8873dbccb5d26a2ca2557552325fcb953fb4a3a3160005260206000f3` (30 B) | 30 × `00` |

The declared bytes are valid EVM: `ADDRESS, PUSH1 0x00, MSTORE, PUSH1 0x20, PUSH1 0x00, RETURN` — the deployed
runtime of a creation, i.e. exactly what `code_changes` is for.

**The RLP length prefix is correct on both** (`89` = 9-byte string, `9e` = 30-byte string, identical on both
sides), so this is a value defect, not a length or structure one.

### Internal check that rules out an extraction artifact

On each fixture the number of differing bytes is **exactly the code length minus the count of `0x00` bytes in the
true code** — the real code's own zero bytes coincide with the zero fill:

| fixture | code length | `0x00` bytes in the true code | differing bytes | length − zeros |
|---|---|---|---|---|
| 12925 | 9 | 2 | **7** | 7 ✅ |
| 13000 | 30 | 2 | **28** | 28 ✅ |

And the extractor's own self-check holds: bytes collected = `bv_bal_shadow_rebuilt_len` (507 and 535).

## Why it has been invisible

- **Equal length.** `rebuilt_len == supplied_len` on both, so every length-based screen passes it.
- **Correct multiplicity.** The entry is present, at the right BAI, on the right account — only the value is
[...]

Issue #10871, opened 2026-07-29
BAL: balance_changes values mis-attributed ACROSS accounts - a correct value emitted on the wrong row, plus a spurious row (6th mechanism)


pirapira:
## Sixth `balance_changes` mechanism: a CORRECT value on the WRONG account

**11739** `test_selfdestruct_created_same_block_different_tx`, `bv_fail 44`, rebuilt 617 / supplied 594 — **shortfall −23, an over-emission.** Reconstruction validated at 617 and it parses, so **complete** by the corrected gate.

| account | declared | rebuilt | |
|---|---|---|---|
| `081f85bb…` (**generic**) | `[(1,1), (2,0)]` | `[(1,1), (2,46)]` | second value **foreign** |
| `22d339df…` | `[(2,46)]` | `[]` | **missing** |
| `5de52cb7…` | `[]` | `[(2,45)]` | **spurious** |
| coinbase | `[(1,797088), (2,6270870)]` | `[(1,797088), (2,8711382)]` | wrong, **not a delta** |

**The value 46 is declared on `22d339df…`, absent there, and present on `081f85bb…` where the spec declares zero.** The value is **correct and on the wrong row** — plus a spurious 45 on an account the spec declares nothing for.

**⇒ Mis-attribution across accounts, not a wrong computation.** That is a sixth distinct mechanism in this field.

## ⛔ And it bounds #10870: the coinbase accumulation defect is NOT universal to coinbases

On this row, **8,711,382 ≠ 6,270,870 − 797,088**, so **this coinbase fails differently.** The accumulation identity (six exact deltas at multiplicity 2 and 5) **holds where it was measured and does not generalise to all coinbases.** #10870's title should be read as *the coinbase producer on the rows measured*, not *every coinbase*.

## The fixture ask on #10868 was premature — corrected as a SEARCH problem

The claim *no truly-generic multi-entry balance account exists* was **bounded to the 205-block census and stated as a fact about the corpus.** `/tmp/fc668/in` holds **26,104 inputs**; 205 is **0.8%**.

Scanning **3,728 blocks at stride 7** found **one** (`081f85bb…` above). At **stride 29 (901 blocks)**: **zero**. So the construct is genuinely rare — **roughly 1 in 3,700** — and the census missing it was *expected rather than unlucky*.

[...]

Issue #10865, opened 2026-07-29
BAL: value defects in the all-counts-agree class - at least three mechanisms (one-unit-low x4, large-delta x1, plus the storage bug in #10866); NOT one story


pirapira:
## Measured: the guest's `post_balance` is exactly ONE WEI low, in an otherwise perfect entry

**Fixture `17562 test_buffer`** — `bv_fail 60`, rebuilt 328, declared 328. Rebuilt BAL reconstructed from the commit-log byte stream (`keccak_absorb` reads one byte at a time at PC `0x8002c664`), **with the reconstruction length asserted at 328 before any diff.**

- **Identical address set AND identical order** — nine addresses, same sequence. **So ordering is not the problem here**, which kills the first hypothesis anyone would reach for.
- **Exactly one field differs in the entire structure.** First byte difference at offset **295**:

```
account a6095a391f4252d3be7da82443ad493731584255, field balance_changes
  rebuilt   cbca01880ba1a9ce01a1a9cd
  declared  cbca01880ba1a9ce01a1a9ce
```

Decoded: `tx_index 1`, then an 8-byte post-balance — **`0x0ba1a9ce01a1a9cd` versus `0x0ba1a9ce01a1a9ce`**.

**⇒ The guest post-balance is exactly one wei low.**

## Why this is a new defect class

Not a missing row, not a spurious row, not a set mismatch — **a VALUE error in an entry that is otherwise perfectly placed, perfectly ordered and perfectly counted.**

**One wei is diagnostic:** it is the signature of an integer **rounding or floor-division** difference rather than a wrong formula. Look at a **fee split or a refund** — anything computed with a division — before looking at balance arithmetic itself.

## Structure of the containing class (18 rows where ALL COUNTS AGREE)

| rows | shortfall | fixtures |
|---|---|---|
| **7** | **0** — identical length, identical counts, different digest | pure value-or-order |
| 4 | 1 | 00250, 03968, 16606, 16706 |
| 4 | **25** | 23725, 23821, 24049, 24254 — **all `test_valid_multi_type_requests`** |
| 2 | 6 | 23093, 23186 — `withdrawal_requests` |
| 1 | 3 | 23230 — `consolidation_requests` |

**Seven of the eighteen are system-request fixtures** (withdrawal, consolidation, multi-type), clustered at 3, 6, 6 and 25×4. **Four identical 25s across four different request permutations says template rather than accident.**

## ⛔ Why nothing found this until now

[...]

Issue #10698, opened 2026-07-27
convergence: publish the sender's post-gas-fee balance and the additive gas refund (the deferred balance half of #10691) — with two measured defects and a reference implementation racing deletion


pirapira:
> **Note:** Written by **Claude Code** (coord agent). Measurements are the scoper and evm-asm3 agents'; filed because this work currently exists only in a merged PR's body.

**#10691 landed the sender NONCE through AccountState and deliberately deferred the sender BALANCE half.** That deferred half is not tracked anywhere — only in the merged PR's deferred section — so this issue exists to hold it, along with two already-measured defects and a reference implementation that is scheduled for deletion.

## What the spec requires — three writes, two lifetimes

| spec site | write | survives a body revert? |
|---|---|---|
| `fork.py:1103` | `increment_nonce(tx_state, sender)` | **yes** — landed in #10691 |
| `fork.py:1108` | `set_account_balance(tx_state, sender, balance − effective_gas_fee − blob_gas_fee)` | **yes** — **this issue** |
| inside `process_message` | value transfer | **no**, rolled back |
| `fork.py:1169` | `create_ether(tx_state, sender, gas_refund_amount)` | **additive**, post-execution — **this issue** |

Note `:1108` subtracts **gas and blob fees only, not `tx.value`** — value moves inside `process_message` and is on the other side of the rollback boundary.

## Two defects already measured against a from-scratch attempt

An earlier variant of this work produced 5 new false rejects on a paired 200-row run (seed 1824272313, parent FR 6 → 11). Both causes were pinned:

1. **Wrong blob price.** The inclusion debit used `tcbg_blob_fee_be` = `tx.max_fee_per_blob_gas`, where the spec debits `calculate_data_fee` — the **actual** blob-gas price. **B2 already does this correctly** via `bsg_blob_price_be` after `amsterdam_blob_gas_price_u256`; copy that. Witnesses: `00008`, `00132`, `00171`, `00175` (all `sufficient_balance_blob_tx_pre_fund_tx`).
2. **`sender == coinbase`.** The sender both pays the fee and receives it as coinbase, so a published inclusion snapshot must be adjusted by the fee-credit path as well as by value transfer and refund. Witness: `00023`.

[...]

PR #10888, opened 2026-07-29
fix(bal): code_changes must name the retained heap copy, not the create-child scratch (GH #10887)

pirapira:
Fixes GH #10887. `code_changes` values serialized as **all zeros with the correct length**, so every length and
multiplicity check passed them and only the digest disagreed.

> ⚠️ **The zeros were never the defect.** They are what the aliased region *happened to contain* — the next call
> frame zeroes its memory, which is correct EVM behaviour. The same aliasing with different EVM memory contents
> yields *different* wrong bytes, which is precisely why the steerability question in #10887 exists at all. A reader
> taking the title literally will look for a zeroing bug and find nothing. **This is pointer provenance, not a
> zero-fill.**

## Root cause — the module already stated the rule and then broke it two lines later

`create_record_code_effect` publishes the deposit to AccountState from the **retained code-effect heap copy**
(`exec_code_effect_log + s3 + 48`), under a comment saying to do exactly that and *"never from the reusable
create-child scratch"*. It then handed `account_write_record` `a3 = s1` — **which is that scratch**, living in
`evm_memory_pool`. The account-write row forwards `a3` verbatim to `bal_builder_append_code`
(`AccountWriteMap.lean` `+80`/`+88`), so the BAL's `CodeChange.new_code` was a pointer into EVM memory.

Traced by commit log on `12925`:

| commit | in | what |
|---|---|---|
| 4,467,144 | `extcodecopy_at_header_state_root` | the real code written: `30 60 00 52 60 20 60 00 f3` |
| **4,606,585** | **`h_STATICCALL`** | the window is **zeroed** — the next call frame clearing its memory, correct EVM behaviour |
| 4,725,000 | `bal_builder_append_code` | records the now-**dangling** pointer |
| 5,195,883 | `keccak_absorb` | reads **nine zeros** |

**The length was always right because the length is copied by value; only the bytes were aliased.** That is exactly
why it was invisible.

## The fix needs no new storage

`CreateCodeEffectLog`'s own header already requires it — *"the BAL's `CodeChange.new_code` must copy those bytes
unchanged"* — and `AccountState` already retains pointers into that heap, which is why it may not be reused or
deleted. So this is the **same expression as the sibling call**, recomputed rather than carried because the helpers
between them may clobber `t0` (`s3` survives; callees preserve `s`).

**Verified the pointer changed arena, not merely value:** now `0xbd82dce8` = `exec_code_effect_log` (`0xbd82dcb8`)
**+ 48**; previously `0xb4c29ce0`, inside `evm_memory_pool` (`0xb4c29c60`).

## Pre-registered acceptance

Content-only, so **any length change would be a failure signature, not partial success**:

| fixture | `fail_code` before | after | length | rebuilt vs declared |
|---|---|---|---|---|
| 12925 `test_scenarios…` | 60 | **0** | 507 = declared, **unchanged** | **BYTE-IDENTICAL** |
| 13000 `test_scenarios…` | 60 | **0** | 535 = declared, **unchanged** | **BYTE-IDENTICAL** |

**Byte-for-byte, not merely non-zero** — the rebuilt BAL recovered from a commit trace is identical to the declared
bytes over all 507 and 535 bytes respectively, so the 9-byte and 30-byte code values are exactly right rather than
plausibly right. (`fail_code` 0 already implies the whole-BAL digest matches; this states it directly, since
"non-zero" would not have been the bar.)

**Do-not-move set, re-measured on this artifact:** 23100 566, 23725 843, 23200 246, 23260 264, 04460 233 — all still
passing; 00565 350, 11658 662, 11659 710 — still `fail_code` 60 at their prior lengths; 00566 264, 00346 263,
00578 264 — passing.

## Guards, one of each sign

- **Negative**: `mv a3, s1` must not appear. Pinned negatively because it is well-formed, produces a *correct
  length*, and fails only in the bytes 590k commits later — **no positive assertion about that call would catch it.**
- **Positive**: the replacement pinned **with its `+48` offset**, since a pointer to the record *base* would
  serialize the 20-byte address and the length fields as if they were code — a well-formed wrong answer one `addi`
  away.

## Severity note

#10887 records a **possible false-accept** reading: the aliased value was whatever EVM memory held at serialization
time, `h_MSTORE` writes 236 times into that region, and the nearest such write lands **one byte** from the window
with `MSTORE` taking a free offset operand. No exploit was demonstrated, and this fix removes the aliasing entirely,
so it closes the false-reject and the possible false-accept readings together without the exploitability question
needing to be settled first.

## Layout

`.text` +0xc → `0x0673f4`, **ELF-measured**. `.data`/`.bss` sizes unchanged. Fixed point verified **with a
discriminating control** (two identical emissions *and* the pre-fix artifact differs). Full `lake build` clean, zero
casualties.


pirapira:
Reviewed. **Approving on content**, labelling once the build is green.

## The fix is one argument, and the comment is why it will survive

`a3` moves from `s1` — `create_child_code`, the reusable create-child scratch that lives in `evm_memory_pool` — to the retained heap copy. One source file plus its regenerated consequences.

What makes this landable without further argument is that the comment records the whole chain rather than the change: that `s1` is the reusable scratch; that the **adjacent** `account_state_record_code` call already says in as many words to publish "from the retained heap copy, never from the reusable create-child scratch"; that the account-write row carries its `a3` straight through to `bal_builder_append_code` via `AccountWriteMap +80`/`+88`; and the measured evidence on 12925. A future reader asking "why not `s1`, it's right there" gets the answer in place.

## Verification is byte-identity, not absence-of-zeros

- 12925 — `fail_code` 60 → **0**, 507 bytes, **byte-identical to declared across all 507**
- 13000 — `fail_code` 60 → **0**, 535 bytes, **byte-identical across all 535**

The author diffed rather than inferring, having noted that `fail_code 0` already implies the whole-BAL digest matches. That distinction mattered here: "no longer zeros" would have been satisfied by *any* non-zero bytes the aliased region happened to hold, which is precisely the failure mode under repair.

## What the issue title will mislead people about

**The zeros were never the defect.** They are what the aliased region happened to contain — the bytes were written correctly by `extcodecopy_at_header_state_root` and then destroyed when the next call frame zeroed its memory, which is *correct EVM behaviour*. The same aliasing with different memory contents yields different wrong bytes. Anyone reading #10887's title literally will hunt for a zeroing bug and find nothing.

That is also why this closes a potential **false-accept** vector and not only a false reject: the emitted value was whatever block-controlled EVM memory held at serialisation time, with the nearest ordinary `MSTORE` measured **one byte** from the window and ~590,000 commits of opportunity between the wipe and the absorb. Steerability was measured; a working exploit was never demonstrated, and the fix removes the aliasing so neither reading needs settling first.

## One thing worth keeping in the record

The correct pattern was **two lines above** the defect, under a comment stating the rule. That is the same defect class as this morning's N+1 finding — one recorder, two callers, only one carrying the obligation — in its consumer form: one source, two consumers, only one taking the right pointer. Each site reads correctly in isolation, which is why per-routine review cannot catch it.

No new storage, no capacity constant, nothing for the layout decision to absorb — the byte heap was already allocated and already documented as retained.

pirapira:
## Corpus regression check: **13 fixtures gained, zero lost, zero length changes**

Twelve hand-picked fixtures can only show the absence of a regression *where I looked*, and this touches a producer
every `CREATE` goes through. So the same **1,045-fixture systematic sample** (every 25th of the 26,105 stateless
inputs) was screened before and after, all 1,045 comparable:

| class | before | after |
|---|---|---|
| **PASS** | 627 | **640** (+13) |
| LENGTH | 305 | 305 (unchanged) |
| **CONTENT** | 58 | **45** (−13) |
| EARLY (all codes) | 55 | 55 (unchanged) |

**13 fixtures changed class, every one `CONTENT → PASS`:** 12925, 13000, 13075, 13150, 13225, 13300, 13375, 13450,
13850, 13925, 14000, 14075, 14150.

- **Regressions (PASS → not PASS): 0**
- **Fixtures whose `rebuilt_len` changed: 0 of 1,045** ← the registered failure signature

That last line was the pre-registered condition: this is a content-only fix, so **any** length movement anywhere
would mean an entry's shape had changed. None moved.

The 13 gained are the whole `test_scenarios` family — the family both hand-picked fixtures came from — so the fix
generalises across it while touching nothing else. At this sampling rate that is **on the order of 325 fixtures
corpus-wide**, stated as a **rate estimate from a 1-in-25 sample**, not a count.

**The other 45 equal-length content fixtures are unaffected and expected to be**, since they are different
mechanisms: `test_selfdestructing_initcode` (6, → #10871), `test_buffer` (5), `test_sufficient_balance_blob_tx` (4,
→ #10698), and ~12 distinct `sstore_*` variants, plus the one-unit-low cases (#10865).

PR #10889, opened 2026-07-29
docs(agents): name check-build-parallel.sh as the pre-push gate in the layout-regen checklist

pirapira:
Adds one section to `docs/agents/layout-regen-checklist.md`: **The thing you ran is not the thing CI runs** — three entries, one lesson:

1. `lake build codegen` ≠ `lake build` (already in the checklist; folded under the heading).
2. Local cache-mode pass ≠ CI pass (cache-disabled per-module `Built` line as the only real evidence).
3. **NEW: a handpicked gate subset ≠ the driver.** `scripts/check-build-parallel.sh` is named explicitly as the gate to run before pushing; a green `lake build` says nothing about any of the ~25 `check-*.sh` scripts. This is the gap that produced today's red CI after a green local day.

Also folds in the drift rule generalisation (any source insertion inside a generated verbatim block span is drift — guards belong beside the block, not inside it).

Docs-only; no gate run needed beyond CI.


pirapira:
Reviewed — **approving on content.** One sequencing note at the end; nothing to change in the text.

Grouping the three under **"The thing you ran is not the thing CI runs"** is what makes this durable. As three separate bullets they read as unrelated build trivia; under one heading they are a single lesson with three instances, and the next instance — there will be one — has an obvious place to go.

Each entry carries its *reasoning* rather than just its rule, which is the difference between a checklist someone follows and one someone skips:

- The `lake build codegen` entry says **why** (the codegen target builds only the executable's import closure, so proof modules outside it are never type-checked). A reader who knows that can predict the next variant instead of memorising this one.
- The cache entry names the trap precisely — *"I deleted the olean and rebuilt" is not evidence*, because the cache re-materialises it — and gives the positive form, a named `Built <module>` line from a cache-disabled build. That distinction cost real time to establish today.
- The driver entry names the script, lists what it invokes, and states the failure mode in one line: **running four of the gates and inferring the rest is how a red CI follows a green local day.**

And the closing note generalises correctly: **any** source insertion inside a generated verbatim span is drift — comment, `#guard`, blank line. My own record had this as "a comment inside a converted asm block," which is too narrow; the rule is about the span. Worth keeping the observation that the guards were *correct* and only misplaced, because the error message points at a generated block while the mistake is where a correct thing was put.

## Sequencing

**Please leave this unmerged until #10888 lands.** Nothing to do with the change: this repo enforces up-to-date branches, so merging any PR puts every other open branch into `BEHIND`. #10888 is a code fix with regenerated `GuestAddrs`/`symbol-addresses.tsv` currently being re-verified against the full gate driver; this is one docs file that updates for free. Landing cheap-before-expensive cost this queue two rebuild cycles earlier today, and the rule is to land the branch that is expensive to revalidate first.

Merging `main` **into** this branch remains free for everyone and needs no permission — only the outbound direction is queued. I'll label it the moment #10888 is in.

PR #10896, opened 2026-07-29
BAL: record checked system transaction account reads

### LINKED ISSUE #10836: BAL: the guest's recorded-read set does not match the spec's touch set - under-inclusive on unmodified recipients and created addresses (+27), see #10860 for the over-inclusive direction
## Root

The guest has **two families of header-state accessors, and only one records a read into the BAL read log**:

| routine | jal callers | records a read? |
|---|---|---|
| `account_at_header_state_root_tracked` | 8 | **yes** — calls `account_read_record` |
| `account_at_header_state_root` | 9 | no |
| `balance_at_header_state_root` | 5 | no (body scanned) |

So **14 call sites read header state through a non-recording accessor** against 8 that record.

The read-log architecture itself is sound and complete:

```
account_read_record            -> increments tx_account_reads_count (per-tx log), 4 jal callers
read_sets_incorporate_tx       -> promotes tx_account_reads_count into account_reads_count, zeroes the tx counter
bal_builder_incorporate_touched_accounts  (#10834)  -> walks account_reads_count
```

Nothing is recorded inline. This is why #10834 moved the EIP-7928 family from 136 to 806 matching: the *feed* already existed, and #10834 supplied the missing *consumer*. Accounts read through an untracked accessor never enter the log, so the (correct) consumer is structurally unable to see them.

## Confirmed symptom: the 45 coinbase entries

The coinbase fee-credit block makes exactly six calls — `tx_gas_result_increments`, `tx_effective_gas_pricing`, `u256_mul_u64_be`, `balance_at_header_state_root`, `u256_add_be`, `record_nonstorage_effect` — and **none records a read**. `record_nonstorage_effect` records a *change*; `balance_at_header_state_root` is the untracked accessor.

Spec side: `fork.py:1172` `create_ether(tx_state, block_env.coinbase, U256(transaction_fee))` is **unconditional** — no nonzero-fee guard. All 45 affected blocks have every transaction at zero effective priority fee (75 txs decoded, 0 unknown shapes, across 60 Blob / 11 Legacy / 2 SetCode / 2 FeeMarket), so the write is net-equal, the build-time filter drops the change row, and the **read** is what legitimately produces the declared entry.

Two mechanisms were measured and **refuted** before this root was found, both because they assumed the read exists somewhere:

- *ordering* — moving `read_sets_incorporate_tx` after the coinbase fee effect: 1024-row paired A/B, `OK=974 FR=50 FA=0` on both legs, **zero transitions**.
- *next-tx promotion* — refuted by 20 direct counterexamples; multi-tx blocks are *anti*-correlated (83.3% vs 62.5%).

## Hypothesis for the remaining execution-time residual

Of the 80 residual empty entries, 26 are accounts discoverable only by executing code — 15 internal CALL targets (address embedded as a literal in witness code or calldata), 9 precompiles, 2 probable runtime-computed CREATE addresses. **Plausibly further untracked-accessor sites among the 14.** Not asserted; the trace from entry point `00036` (one precompile entry, one role, non-empty coinbase, no confound) tests it.

## ⚠️ The name trap that let this survive every prior audit

`account_at_header_state_root_tracked` and `balance_at_header_state_root` differ by one word and a prefix, **only one records, and nothing at a call site tells you which was picked.** Any fix should consider making the untracked variants hard to select by accident rather than merely correcting the current 14.

## Next steps

1. Record the coinbase read at the fee-credit site so it enters `account_reads`, mirroring `create_ether`'s read through `modify_state`. Audit register liveness first — that block threads `a0`–`a5` into four consecutive calls with `t0`–`t3` as scratch, and currently makes no read-recording call, so nothing there protects anything across one.
2. Build the **14-site list by ELF symbol address** (`nm` function ranges, attribute each call by address). A nearest-preceding-label walk is unsound here: it matches only 2982 of 3746 non-`.L` labels and sails through converted verified-program regions with x-register names and numeric labels.
3. Decide per site which of the 14 *should* be tracked — **not all of them**; not every balance read is a spec-level touch. This list is what sizes the fix and bounds its blast radius.

pirapira:
## Summary

`execution-specs` records the system-transaction target as an account read: `process_checked_system_transaction` in pinned `e5a8caf1b` `fork.py:722` calls `get_account` at `:755-759`. The four checked-system callers are withdrawals (`:962`), consolidations (`:973`), builder deposits (`:985`), and builder exits (`:997`).

The guest now uses `stage_system_call` as the matching shared seam: it records the target with `account_read_record`, then promotes the transaction read set after each of those four system transactions. Beacon roots and history remain outside this seam, matching their separate unchecked-spec path.

The standalone codegen BuildUnits that embed `stageSystemCallFunction` now include the new account-read callee and data section; the derive-block probe also includes the transitive read-set-promotion closure.

Closes #10836.

## Evidence

The four smallest controls from the affected families were independently measured at rebuilt/supplied `197/224` before this change. On the regenerated guest, each is accepted with a byte-length match:

| Fixture | Before | After |
| --- | --- | --- |
| 04512 builder deposits | 197 / 224 | 224 / 224, `fail_code=0` |
| 04516 builder exits | 197 / 224 | 224 / 224, `fail_code=0` |
| 23087 withdrawals | 197 / 224 | 224 / 224, `fail_code=0` |
| 23361 consolidations | 197 / 224 | 224 / 224, `fail_code=0` |

Each recovers exactly the missing 27-byte touched-only account entry. The controls are expected-valid; FA is zero across these four controls.

The candidate ELF is `b844a561fb07218698772b26c0fe7e793674d6df261229c3dfc58c47a1ce3668`. Layout regeneration reached a three-pass fixed point: `.text=0x67410`, `.data=0x5370`, `.bss=0x1c139580`, with `GuestAddrs.bnf_le_a` unchanged at `0xa3000ee0`. `check-build-units-link`, `codegen-stateless-link-check`, `check-region-map`, `check-memorylayout-region-coverage`, and `check-asm-to-program` pass.

`check-guarded-handler-bytes` is unverified locally: this cache-mode checkout has 820 built modules while main has 2,820, and `GuardedHandlerSpecs` is outside the codegen executable's import closure. A full non-cache build in CI is the authority for that gate.


pirapira:
Approving. The argument is the spec, the evidence is a two-sided boundary match, and the local caveat is stated honestly. Labelling `merge!` — it will land when checks go green.

**The seam is confirmed from both directions, which is stronger than either alone.** The guest's static caller inventory for `stage_system_call` is the four `derive_*_requests` plus a standalone probe. The spec's `process_checked_system_transaction` (`fork.py:722`) has exactly four callers — `:962`, `:973`, `:985`, `:997` — and beacon-roots/history go through `process_unchecked_system_transaction` at `:897`/`:903` instead, never reaching it. **Guest boundary and spec boundary derived independently and coinciding**, with the `get_account` at `:755-759` being the read that motivates the record. That also means the record sits at **one** choke point rather than four call sites, so there is nothing to forget when a fifth checked system call is added.

**The controls cross-check against an independent earlier measurement.** Before this change, ranking the entire FR set by input size put the four smallest members of this class at `rebuilt 197 / supplied 224` — all four, across four different families with four different predeploy addresses. This PR reports `224/224` on all four. **Each recovered exactly 27 bytes**, the touched-only entry `da 94 <addr> c0 c0 c0 c0 c0`. Two instruments, two authors, same declared length. Including the before-and-after pair rather than only the passing state is what makes that checkable.

**The build-unit closure work deserves explicit credit** because avoided breakage is invisible to a reviewer. Adding a callee to `stage_system_call` would have stale-linked every unit embedding `stageSystemCallFunction`; all five now carry the recorder and its data, and the derive-block probe carries the promotion closure. That is precisely the failure `check-build-units-link.sh` exists to catch — its own comment cites `zisk_stateless_verdict_v2` staying broken across three PRs. Anticipating the gate is different from passing it by luck.

**On the local `check-guarded-handler-bytes` caveat — the corrected reason is the right one.** `820/820` olean-to-hash means there is *no* trace-without-olean divergence, so this is not #10601. It is that only 820 of 2,820 modules were built and `EvmAsm/Codegen/Proofs/GuardedHandlerSpecs` is outside a codegen-targeted build's import closure. Stating the guard as unverified locally *with the count* is correct; CI's non-cache run is the authority. I had originally pointed at #10601 myself and that attribution was wrong.

One thing I'd want on the record before this closes #10836: the issue's class was confirmed across **four** families, and this fixes all four. If a fifth `process_checked_system_transaction` caller appears in a future fork, the seam covers it automatically — worth one line in the issue so the next reader knows the coverage is structural rather than enumerated.

Issue #10893, opened 2026-07-29
codegen: fixture manifest truncates names, making distinct scenarios unjoinable


pirapira:
## Problem

The EEST fixture manifest stores test names truncated at 102 characters, and the
corresponding `.input` filenames are truncated the same way. Distinct test cases
that share the first 102 characters therefore have identical labels in both
artifacts.

## Impact

An index-to-full-name join is impossible by construction. This blocked the
investigation of the CALL-target BAL mismatch: fixture labels could not
distinguish two scenarios with opposite expected outcomes, so the discriminator
had to be recovered by decoding the input payloads directly.

## Current workaround

Treat the input payload as authoritative and decode it rather than using the
truncated manifest name. This is a workaround, not a usable stable identifier
for tooling or future investigations.

## Requested change

Preserve a collision-free fixture identifier (for example the full test name or
a full-input hash) in the manifest and in any generated filename/index used by
the harness.

pirapira:
Measured on exact base `23e6b5b9164418ff343bf79722697821f8e6a3df`, guest ELF `801334a83c…` (built 2026-08-10 00:00:33+0200), runner `bae48269…`. Named witness resolved by **input sha**, not by label.

**PASS** — `rc=0`, `succ=1`. The named row now behaves as expected, so the issue is closed on a measurement rather than on the code looking right.

⚠️ The witness is **expected-valid**, so the #11881 invalidity-presence check (is the field the fixture declares invalid actually present in the packed input?) is **vacuous here** — there is no expected-invalid field to verify. This closure rests on the row passing, not on it exercising an invalid path.

PR #10898, opened 2026-07-29
BAL: record resolved delegated CALL targets

### LINKED ISSUE #10860: BAL: CALL target read survives OOG-before-target-access (#10847 regression)
## Measured regression attribution

Fixture `01060_test_bal_callcode_no_delegation_and_oog_before_target_access_fork_Amsterdam-blockchain_test-both` is a one-fixture content bisect of the CALL-family target-read change (#10847).

Both runs used the same immutable Spike runner `c6ac09fa72c2103b6d3e390b779b022ef7f220e5a39bb5f667114a89d78e5868` and `SPIKE_OUTPUT_LEN=1024`:

| artifact | source | ELF SHA-256 | shadow status | rebuilt BAL length | supplied BAL length |
| --- | --- | --- | ---: | ---: | ---: |
| parent | `6e10823c36be6ec2e8d4b135050a3737b82dc8c7` (the #10847 base) | `bf531f805fda9ad7e060b024c2129965f57252b01a867bd75130d1ecf9a9d0e8` | 0 | 333 | 333 |
| candidate/current main | contains #10847 | `40e73f74c5fa1c12d98d35906d89aa6ca71ef930e9e359a377b1cd1c56b8f12a` | 1 | 360 | 333 |

This establishes that #10847 makes the fixture regress by 27 emitted bytes. It does **not** establish why.

## Withdrawn mechanism

The initial draft named #10847's unconditional pre-charge `account_read_record` as an OOG-unwind defect, citing `prepare_dispatch` and `process_message`. That mechanism was tested by moving the record immediately after `runtime_access_account_charge`; the temporary artifact built cleanly but 01060 remained `360` versus `333` with shadow status `1`.

Therefore the proposed pre-charge/OOG mechanism is **withdrawn**. The original spec citation was relevant to a plausible story, but did not prove the story; it inherited credibility from the content bisect without earning independent evidence.

The leading measured structure is instead tracked separately: the current guest emits the same address twice on this fixture, with one empty account entry and one balance-and-nonce entry. #10847 may be the trigger that exposes a pre-existing lookup defect rather than the defect itself. No guard or revert should be landed from this issue.

## Scope

The content attribution remains useful for future repair gates. A repair must return this fixture to `333 == 333` while retaining #10847's independently measured improvements; its mechanism must be established separately.

pirapira:
Closes #10860

## Why

Amsterdam's CALL-family code-resolution path performs its initial static gas
check before state access. When a callee has a 23-byte `ef0100 || address`
delegation designation, `calculate_delegation_cost` resolves the transaction-
state delegate and the subsequent `get_account(code_address)` touches that
selected code address. The record therefore belongs only on the successful
designation-resolution path, after the initial access, transfer, and memory
checks; it is not an unconditional read of the popped CALL target.

This follows pinned execution-specs `e5a8caf1b`,
`vm/instructions/system.py:440-465`: static `check_gas` precedes delegation
resolution, and the selected `code_address` is passed to `get_account`.

## Change

- Move delegation handling after `callMemoryExpansionGasAsm`.
- Reuse the existing 23-byte length, `ef0100` marker, and
  `bal_same_block_delegation_code_resolve` checks.
- On successful resolution, record the resolver's transaction-state export,
  `bsbd_deleg_target`, through `account_read_record`, and charge that same
  selected delegate.
- Remove the prior unconditional pre-charge target-read sequence.

The producer's placement is guarded structurally: the old pre-charge sequence
must be absent, and this tail must contain exactly one delegated-resolution
producer. The source build establishes that the save triple at the entry of
`basicPrecompileCallTail` dominates the later restores of `x13`, `x10`, and
`x12` on every CALL-family instantiation.

## Discriminator and payload control

The two originally conflated fixtures differ in their input, not merely their
truncated names. `00704` contains a delegation designation; `00705` does not.
The latter tests whether the guard fires, while the former also tests whether
the extracted address is the resolver-selected payload. This is a structural
distinction, independent of the manifest-label truncation documented in
#10893.

A guarded commit-log reconstruction of 00704 length-asserted both BAL streams
at 360 bytes before decoding them to address-to-field RLP. The result was one
touched-only row substitution: the pre-fix guest used
`37f536464af59c8d7358cae965f92cbeadd58dcb`, while the declared BAL uses
`909818add4aa1606e130506d84b397466e6c39cc`; all other nine rows and every
shared-row field matched. The source used the stale header marker operand
`t3 + 3` after calling the transaction-state resolver. This patch instead
uses its `bsbd_deleg_target` output.

The existing access-charge call on `origin/main` already used `t3 + 3` after
discarding the resolver output (the stale-pointer load and charge are the
pre-change sequence). This patch corrects that shared pointer too: on a
same-block re-delegated target, the warm/cold access charge previously applied
to the stale pre-state delegate rather than the transaction-state-selected
delegate. This is a gas effect and therefore an FA-side risk, not a convergence
metric. It is corrected here because the single pointer feeds both the new
record and the existing charge; #10899 retains the searchable gas-defect
record and references this fix.

## Evidence

The former unconditional experiment, on its named frozen base and runner,
showed 108 false-reject-to-OK transitions and 132 OK-to-false-reject
transitions. Those are supporting evidence for the discriminator rather than
the argument for this spec-alignment change: the 132 entries are valid
designation-resolution touches that an unconditional target producer
incorrectly removed.

The previous regenerated artifact was
`8dc1968512c6c8a3ec9b4751baa48ff7aa78e2371148bdc84d43425d208098b2`,
with a three-pass fixed point at `.text=0x67378`, `.data=0x5370`,
`.bss=0x1c139580`, and `GuestAddrs.bnf_le_a=0xa3000ee0`; its layout and gates
will be refreshed after the serialized source correction is regenerated.

Before that correction, direct control runs with frozen spike runner
`c6ac09fa` reported:

| Fixture | Result |
| --- | --- |
| 00705, no designation | `fail_code=0`, rebuilt/supplied `333/333` |
| 01060, OOG before target access | `fail_code=0`, rebuilt/supplied `333/333` |
| 00704, designation present | `fail_code=60`, rebuilt/supplied `360/360` |

The 00704 decode makes that residual attributable: it was a stale-address
payload bug in this producer, rather than an unrelated equal-length residual.

## Post-regeneration result and residual

The regenerated shipping ELF is at text=0x67464, data=0x5370, bss=0x1c139580, with GuestAddrs.bnf_le_a unchanged at 0xa3000ee0. The three generated layout files reached a byte-identical fixed point.

00704 remains an expected-valid, equal-length code-60 residual: rebuilt and declared BALs are both 360 bytes, but the rebuilt row carries 37f536464af59c8d7358cae965f92cbeadd58dcb where the declared row carries 909818add4aa1606e130506d84b397466e6c39cc. The new bsbd_deleg_target load executes once on this fixture and the regenerated assembly contains that load at the CALL helper site, so this fixture is not repaired by this PR. The input role investigation is tracked separately.

The load remains intentionally: system.py:453 branches on delegated status, :458 resolves the designation, and :460 reads the selected delegate. Retaining the transaction-state selected address aligns this path even where 00704 currently selects the same address as the former pointer.

The change that this PR does establish is the conditional delegation guard. On the frozen control ELF, 00705 with no designation and 01060 with OOG before target access both return fail_code=0 with rebuilt/supplied BAL lengths 333/333. These expected-valid controls show FA=0. The helper remains after callMemoryExpansionGasAsm, so no resolution touch precedes the initial static gas check, matching system.py:436.


pirapira:
**This PR's content is now carried by #10905** (a batch of #10898 + #10903), labelled `merge!` and armed with auto-merge.

⚠️ **Please do not merge main — or the batch branch — into this PR.** GitHub auto-closes a PR only when its *head* becomes an ancestor of main, so any push here after the batch was cut moves this head out of the batch and leaves the PR dangling open even though its work landed. That is exactly what happened to #10900/#10901, which I had to close by hand.

The batch merged this branch at head `4621bbd661edf6ef25237467872195ac8b881abb`. It stays open as the fallback in case #10905's build fails; once #10905 lands I'll verify containment against that SHA and close it.

Note on the conflict: it was in generated/layout files only (`GuestAddrs.lean`, `RegionMap.lean`, `symbol-addresses.tsv`), resolved by regen against the linked ELF rather than by hand-picking a side — the combined tree's true `.text` pin is `0x06748c`, which matched neither branch.

Issue #9458, opened 2026-06-25
guest: blob_gas_subtraction_tx fixtures fail bv_fail=34 — sender balance staged with max_fee instead of effective_gas_price


pirapira:
## Problem

80 of 128 test cases in `cancun/eip4844_blobs/blob_txs/blob_gas_subtraction_tx.json` fail on both spike and ziskemu with `bv_fail=34` (BAL storage mismatch — tx never dispatched).

## Root cause

`BlockVerdictFunction.lean:791` stages `bv_pending_upfront_sender_post = sender.balance - bv_upfront_cost` for runtime `BALANCE(ORIGIN)`, where `bv_upfront_cost` is computed at line 761-783 using `tefgp_max_fee` (max_fee_per_gas):

```
bv_upfront_cost = max_fee * gas_limit + tx.value + blob_cost
```

But execution-specs (`amsterdam/fork.py:1080`) deducts `effective_gas_price` (not max_fee) from the sender before EVM execution:

```
sender.balance -= effective_gas_price * gas_limit + blob_gas_fee
```

When `max_fee > effective_gas_price` (i.e., max_fee > base_fee), the guest over-deducts, making the staged sender balance too low. This causes `BALANCE(ORIGIN)` to return the wrong value during EVM execution, producing SSTORE values that don't match the BAL, triggering `bal_storage_matches_exec_log` mismatch → `bv_fail=34`.

## Failing pattern

| Variant | max_fee | eff_price | sender_post (guest) | sender_post (spec) | Result |
|---------|---------|-----------|--------------------|--------------------|--------|
| max_fee=base_fee | 7 | 7 | 0 | 0 | **PASS** (agree) |
| max_fee>base_fee | 14 | 7 | 0 | 3500000 | **FAIL** (disagree) |

The contract code queries `BALANCE(ORIGIN)` and SSTOREs it. When the balance is wrong:
- Guest: SSTORE(0, 0) = 0→0 (no state gas) → different exec log → bv_fail=34
- Spec: SSTORE(0, 3500000) = 0→nonzero (state gas) → correct exec log

## Fix

After the upfront cost **check** at line 786 (which correctly uses max_fee per `check_transaction`), recompute `bv_upfront_cost` using `bv_fee_egp_scratch` (effective_gas_price, written by `tx_effective_gas_pricing` at line 438) before the staging at line 791.

PR #10900, opened 2026-07-29
docs: describe dispatcher reset of account pending journal

### LINKED ISSUE #10877: codegen: document runtime dispatcher reset of AccountState pending journal
## Problem

`BlockVerdictMtxRuntime.lean` currently says that the callable dispatcher
"resets only its pending overlay; durable state ... survive[s]".  The first
half is materially misleading for the AccountState journal: the dispatcher
unconditionally stores zero to `account_state_pending_count`.

This was established from the emitted artifact while diagnosing #10876:

- `runtime_dispatcher_call` contains `sd zero, 0(t0)` at PC `0x80039cd4`,
  where `t0 = account_state_pending_count` at `0xbdf41248`.
- A commit trace of fixture `00337` records that store six times.
- Immediately before the second transaction's coinbase lookup, the diagnostic
  read of `account_state_pending_count` is zero and `account_state_find` misses.

The reset may be correct for the transaction-local overlay, but the comment
does not identify that its AccountState journal is also reset.  It led the
coinbase-fee investigation to infer persistence that the instructions do not
provide.

## Requested change

Correct the comment to state the actual lifecycle and name the AccountState
pending-journal reset.  This issue is documentation/lifecycle clarity only;
#10876 repairs the lost coinbase fee by committing the post-fee snapshot before
the next dispatcher call.

pirapira:
Closes #10877

The callable runtime dispatcher clears its transaction-local pending overlays,
including `account_state_pending_count`. The previous comment said it reset
only a pending overlay without naming the AccountState journal, which made it
easy to infer that the journal persisted across dispatches.

This documentation-only change states the actual lifecycle: pending journal
entries do not survive a dispatch, while durable AccountState and retained
comparator bytes do. It does not alter emitted bytes.

`lake build EvmAsm.Codegen.Programs.BlockVerdictMtxRuntime` passes.


pirapira:
Approving — docs-only, no emitted-byte change, and the wording is a real improvement rather than a restatement.

The old comment said the dispatcher "resets only its pending overlay," which reads as benign and bounded. The new text names the specific counter — `account_state_pending_count` — and states the **negative**: it *does not preserve that AccountState journal across dispatches*. Naming the counter and asserting what is not preserved is what makes it checkable; "only its pending overlay" was true and uninformative.

**One addition worth making before this lands, because it is the third instance of one shape today:** this documents a mechanism that is currently the **root of a live defect**. That same reset is #10876, and it has been measured landing inside a load-bearing window — the record-to-lookup gap on `00232` — which makes it upstream of #10698 and, through it, of #10892. A reader who finds an accurate, confident description of the reset with no pointer to the defect will reasonably conclude the behaviour is intended and settled.

So please add a clause referencing **#10876** — that this reset is implicated in a live defect and is not to be relied on as correct-by-design. The precedent is #9458, whose comment stated the right principle its own code did not implement and survived four months because the comment reassured every reviewer, and `ChildFrameHandlerTailHelpers.lean:56-57`, which documented a stale pointer as the design.

Accurate documentation of a defective mechanism, with no mention that it is defective, is how the defect acquires a defender.

pirapira:
Cross-reference verified and it is better than what I asked for. I requested a clause noting the reset is implicated in a live defect; the added text states the **rule that follows from it**:

> ```
> This reset is the live durability boundary in GH #10876: sender effects
> needed after dispatch must be materialized in durable state rather than
> left in the pending journal.
> ```

That converts the comment from a description into a **constraint a future author can violate and be caught by**. "Implicated in #10876" would have told a reader to be careful; *sender effects needed after dispatch must be materialized in durable state* tells them what the code must do. It is also exactly the boundary the evening's measurements landed on — the pending entry is present at the debit site with its balance-valid flag clear, which is this rule being broken rather than an absence.

Labelling `merge!`. `BEHIND` only; it will need an update-branch when it reaches the front of the queue.

For sequencing: #10896 is ahead with `merge!` and is the expensive one, #10901 follows, this is cheapest and can land last. Since every merge forces the others to update under the strict policy, they should go one at a time rather than all being refreshed at once.

pirapira:
**This PR's content is now carried by #10904** (a batch of #10900 + #10901), which is labelled `merge!`, armed with auto-merge, and running CI.

Noting it here because a batch silently transfers a PR's landing path — this PR's own status still reads `BEHIND`, which looks stalled but isn't. **Please don't sync this branch**; it stays open as the fallback in case #10904's build fails.

Once #10904 lands I'll verify this PR's content in main and close it, manually if GitHub doesn't auto-close (it only auto-closes when a head becomes an ancestor of main).

pirapira:
landed via #10904 (main `b98b46cb9`).

Closing manually because GitHub couldn't auto-close it: a `Merge branch 'main'` sync commit (`3a2efa13d`) landed on this branch *after* I cut the batch, so its head is no longer an ancestor of main. Auto-close depends on head ancestry, not on content.

Content confirmed in main: the head the batch actually merged (`c8e49d6d0`, second parent of batch commit `c1396b84b`) **is an ancestor of main**, and `git log origin/main..` on this branch shows exactly one remaining commit — that pure sync merge, carrying no unique work. Nothing is lost by closing.

PR #10901, opened 2026-07-29
tooling: preserve full EEST case identity in manifest

### LINKED ISSUE #10893: codegen: fixture manifest truncates names, making distinct scenarios unjoinable

pirapira:
Closes #10893.

The converter truncates descriptive labels to keep input filenames under the
host filesystem name limit. That makes a manifest unable to identify distinct
EEST scenarios whose names share the same prefix.

Add a trailing case_id manifest field: SHA-256 over the fixture-relative path,
complete EEST test name, block index, and exact stateless input bytes. Filenames
and their human-readable labels remain unchanged; the manifest now retains a
stable collision-resistant scenario identity for joins and reports.

Existing consumers remain compatible with old six- and seven-column manifests
and intentionally ignore trailing fields in the new eight-column form.

Checks:

- python3 -m py_compile for the converter and all six manifest consumers.
- Sequential conversion of two deliberately colliding display prefixes: both
  rows had eight fields and distinct 64-hex case IDs.
- Random conversion with --random --seed 17 --limit 1: emitted an eight-field
  manifest row.
- git diff --check.


pirapira:
Approving. The shape is right in the way that matters most for this file: **`case_id` is appended as the eighth column, so positions 0–6 are untouched.**

That is what makes it safe beyond the six consumers in this diff. Readers use `len(parts) >= 7` with `parts[:7]`, one retains the six-column form, and the failure paths raise with the observed column count rather than silently mis-parsing. **Out-of-tree readers keep working** — including the full-corpus sweep instrument, which does `if len(p) < 7: continue` and then takes `p[0..3]`; an extra trailing column is invisible to it. Had `case_id` been inserted anywhere before `relpath`, every such reader would have silently shifted a field, and `succ_bit` shifting by one would have inverted FA/FR classification without erroring — which is the worst available outcome for this particular file.

**Deriving the id from content is the right call.** SHA-256 over fixture-relative path, full EEST test name, block index and input bytes means the identity survives regeneration and cannot collide on the thing that actually distinguishes two cases. That directly fixes what blocked #10860: `00704` and `00705` are identical through all 102 characters of their truncated label *and* their filenames, and nothing in the manifest could tell them apart — the investigation had to fall back to comparing input byte lengths.

Two notes, neither blocking:

**1. This identifies a case *instance*, not a *test*.** Because the input bytes are in the hash, `case_id` changes if the fixture content changes. That is correct for "which exact case did this row measure," and it means it cannot answer "the same test across an EEST version bump." Anyone wanting that later needs a separate name-only digest. Worth one line in the docstring so the next reader does not discover the distinction by being surprised.

**2. Leaving the display labels and filenames truncated is the right scope.** Widening them would touch the filesystem name limit that caused the truncation in the first place, and the fix here is to stop *relying* on the label for identity rather than to make the label sufficient. The docstring already says this.

`BEHIND` only — needs an update-branch, like everything else tonight. Labelling `merge!`.

pirapira:
**This PR's content is now carried by #10904** (a batch of #10900 + #10901), which is labelled `merge!`, armed with auto-merge, and running CI.

Noting it here because a batch silently transfers a PR's landing path — this PR's own status still reads `BEHIND`, which looks stalled but isn't. **Please don't sync this branch**; it stays open as the fallback in case #10904's build fails.

Once #10904 lands I'll verify this PR's content in main and close it, manually if GitHub doesn't auto-close (it only auto-closes when a head becomes an ancestor of main).

pirapira:
landed via #10904 (main `b98b46cb9`).

Closing manually because GitHub couldn't auto-close it: a `Merge branch 'main'` sync commit (`dc60aff4c`) landed on this branch *after* I cut the batch, so its head is no longer an ancestor of main. Auto-close depends on head ancestry, not on content.

Content confirmed in main: the head the batch actually merged (`90c2c0428`, second parent of batch commit `1d4193d9d`) **is an ancestor of main**, and `git log origin/main..` on this branch shows exactly one remaining commit — that pure sync merge, carrying no unique work. Nothing is lost by closing.

PR #10903, opened 2026-07-29
codegen: record unchecked system account reads

pirapira:
## Spec alignment

`execution-specs`' `process_unchecked_system_transaction` obtains the target
account before deciding whether the system call changes state; the state
tracker consequently records that access even when the account is absent or
has empty code.  The guest previously performed the EIP-2935 and EIP-4788
lookups but dropped both reads whenever their modeled write was skipped.

This records both unconditional account reads via `account_read_record` and
promotes the startup transaction read set with `read_sets_incorporate_tx`
before BAL construction.  Thus an absent target contributes the required
touched-only account entry while retaining the existing no-write path.

The corresponding execution-specs behavior is
`process_unchecked_system_transaction` and its EIP-2935/EIP-4788 callers in
the pinned `e5a8caf1b` tree.

## Focused evidence

Immutable candidate guest:
`54afb2e30652f8e3514a11b1006303ff6caed2f18f338c687ecdcf6645de744a`.
Runner:
`c6ac09fa72c2103b6d3e390b779b022ef7f220e5a39bb5f667114a89d78e5868`.

With `SPIKE_OUTPUT_LEN=256`:

| fixture | expected validity | result |
| --- | --- | --- |
| `00514_test_bal_2935_absent_contract...` | valid | byte-perfect; `succ=1`, `fail_code=0`, rebuilt/supplied BAL length `189/189` |
| `00538_test_bal_4788_absent_contract...` | valid | byte-perfect; `succ=1`, `fail_code=0`, rebuilt/supplied BAL length `219/219` |
| `04778_test_tx_intrinsic_gas...below_intrinsic...` | invalid | still rejects (`succ=0`, `fail_code=5`), so the focused set has no false accept |

## Generated layout

Regenerated to a fixed point from the committed source.  The measured ELF has
`.text=0x67400`, `.data=0x5370`, `.bss=0x1c139580`; `bnf_le_a` remains
`0xa3000ee0`. `check-asm-to-program.sh` and `check-region-map.sh` pass.

`scripts/check-build-parallel.sh` exposed an existing cache-mode limitation in
its guarded-handler subcheck: `lake env lean` cannot load cache-satisfied
oleans (`GuardedHandlerSpecs.olean` is absent from the local build tree with
`LAKE_ARTIFACT_CACHE=true`). The source build and codegen/link checks ran;
CI exercises that gate in non-cache mode.


pirapira:
**This PR's content is now carried by #10905** (a batch of #10898 + #10903), labelled `merge!` and armed with auto-merge.

⚠️ **Please do not merge main — or the batch branch — into this PR.** GitHub auto-closes a PR only when its *head* becomes an ancestor of main, so any push here after the batch was cut moves this head out of the batch and leaves the PR dangling open even though its work landed. That is exactly what happened to #10900/#10901, which I had to close by hand.

The batch merged this branch at head `95e7d472d7b2897389536c9bdc003125ecc0e461`. It stays open as the fallback in case #10905's build fails; once #10905 lands I'll verify containment against that SHA and close it.

Note on the conflict: it was in generated/layout files only (`GuestAddrs.lean`, `RegionMap.lean`, `symbol-addresses.tsv`), resolved by regen against the linked ELF rather than by hand-picking a side — the combined tree's true `.text` pin is `0x06748c`, which matched neither branch.

Issue #10895, opened 2026-07-29
BAL: tracked-accessor retarget backlog for execution and pre-state account reads


pirapira:
## Finding

The guest has a raw `account_at_header_state_root` entry and a tracked
`account_at_header_state_root_tracked` entry.  The tracked entry records through
`account_read_record` before tail-calling the raw lookup.  The shape is correct,
but an audit found five genuine execution/pre-state reads still routed through the
raw entry.

The execution-specs rule is accessor-level:

- `state_tracker.py:112-144`: `get_pre_state_account_optional` adds `address` to
  `tx_state.account_reads` at line 139.
- `state_tracker.py:179-203`: `get_account_optional` does the same at line 199
  before consulting state.

Thus a read represented by either accessor must be recorded independently of
whether it later observes a write; the block-level builder consumes the merged
set to add touched-only accounts.

## Source audit

There are 11 direct raw accessor calls across 10 Codegen files and seven tracked
calls.  Four raw calls are guest-only verification/probe work and must remain
untracked.  The following five are execution/pre-state reads that should instead
use `account_at_header_state_root_tracked` (or an equivalent accessor-level route):

| Guest site | Spec role |
| --- | --- |
| `BlockVerdictFunction.lean` single-tx sender inclusion check | `check_transaction` sender `get_account` |
| `BlockVerdictMtxRuntime.lean` multi-tx sender inclusion check | `check_transaction` sender `get_account` |
| `TxIntrinsicAuthEffects.lean` authority lookup | authorization-state access |
| `TxIntrinsicStateGas.lean` authority lookup | authorization-state access |
| `CreateCodeEffectLog.lean` pre-block fallback | `get_pre_state_account_optional` analogue |

The two authority sites are already held by #10635; this issue is a backlog and
cross-reference, not a duplicate of that work.

The sender gap is usually masked: a sender normally has a balance or nonce change
and therefore appears in the BAL without the read.  It becomes observable only
when no other change supplies that entry, which explains why an accessor-level
omission can be widespread but rarely visible.

## Follow-up boundary

`BlockVerdictDispatchTx.lean` pre-caches every declared BAL account.  It is
classified as verification only only if every consumer remains comparator-only.
If that cache feeds execution or BAL rebuilding, it is a declared-BAL echo of the
kind prohibited by #10874 and needs separate investigation.

This is source-audit evidence, not an EEST measurement.

pirapira:
[...]

PR #10906, opened 2026-07-29
codegen: record zero-value withdrawal recipients

### LINKED ISSUE #10902: BAL: zero-amount EIP-4895 withdrawals omit required touched-only recipient row
## Claim

The guest drops the touched-only BAL entry required for a zero-amount EIP-4895 withdrawal.

The two oracle-valid fixtures `00591_test_bal_zero_withdrawal...existing_account` and
`00592_test_bal_zero_withdrawal...nonexistent_account` both currently reject with BAL digest
mismatch (`fail_code=60`), rebuilt/supplied lengths `228/255` (a 27-byte missing empty-account
entry).

## Execution-specs rule

Pinned execution-specs `e5a8caf1b8055e4d805c7fb169edfa710914b7da` makes the post-execution
BAI `N+1` at `fork.py:916-921`, then processes every withdrawal. `process_withdrawals` calls
`create_ether` unconditionally for each withdrawal at `fork.py:1215-1226`, including amount zero.
`create_ether` calls `modify_state` (`state_tracker.py:680-700`), which calls
`get_account` (`state_tracker.py:631-643`); that function adds the recipient to
`tx_state.account_reads` (`state_tracker.py:125-142`). On incorporation, reads merge into the
block state (`state_tracker.py:832-881`) and the builder emits `add_touched_account` for every
such address (`block_access_lists.py:691-699`). The test itself specifies the expected empty
AccountChanges at `test_block_access_lists_eip4895.py:572-609`.

Thus amount zero suppresses a balance change, not the recipient access.

## Guest divergence

`EvmAsm/Codegen/Programs/BlockVerdictStateRoot.lean:223-238` decodes the withdrawal delta and
branches directly to `.Lbsr_wl_next` when all four delta limbs are zero. No recipient touch is
recorded, so the guest cannot emit the required 27-byte empty AccountChanges row.

## Scope and next step

This is a source-audit diagnosis, not yet a measurement of a candidate. The repair needs to
record the decoded recipient in the post-execution read set at BAI `N+1`, then promote it before
the BAL builder consumes block reads. It must preserve the existing no-state-change behavior for
zero delta and be measured on both existing-account and nonexistent-account controls.

### Placement constraint

`BlockVerdictStateRoot` is not by itself a viable producer site: its withdrawal replay loop is
bypassed whenever a supplied BAL is present (`bsr_bal_count != 0`), including both failing
fixtures. The live post-body helper
`EvmAsm/Codegen/Programs/BlockVerdictWithdrawalEffects.lean` also branches from a zero amount
directly to `.Lbv_wdne_next` before copying the recipient address or recording an effect. A repair
there (or at an equivalent shared post-execution boundary) must record the recipient and arrange
a subsequent read-set promotion. Do not add a state-root-only record and infer that it is wired.

pirapira:
Fixes #10902.

## Spec alignment

Amsterdam `fork.py` at execution-specs pin `e5a8caf1b` processes every withdrawal by calling `create_ether` before incorporating the transaction state (`process_withdrawals`, lines 1207-1226). A zero amount suppresses the balance mutation, but does not suppress the tracked account access. The guest previously branched away before copying or recording that recipient, so it omitted the 27-byte touched-account entry.

The shared live helper `block_verdict_withdrawal_nonstorage_effects` now records the recipient through `account_read_record` on the zero-amount path. Its EOA and MTx callers both promote that read set at the post-transaction N+1 boundary. No non-storage balance effect is recorded for zero amount.

## Evidence

Candidate ELF: `6550faccc48c5a200d1e20c259d0508e02cdfddedbc4c64948ba82b70811e73d`; runner: `/home/yoichi-bkp/evm-asm2/scripts/spike/spike_run`.

- `00591` zero withdrawal / existing account: `succ=1`, `fail_code=0`, rebuilt/supplied BAL lengths `255/255`.
- `00592` zero withdrawal / nonexistent account: `succ=1`, `fail_code=0`, rebuilt/supplied BAL lengths `255/255`.
- Expected-invalid control `04778` remains rejected: `succ=0`, `fail_code=5`.

Layout regeneration converged at `.text = 0x067428`; `bnf_le_a` remains `0xa3000ee0`. `check-asm-to-program.sh` and `check-region-map.sh` pass.

This is the third and final confirmed instance of the skip-when-nothing-changes class: #10896 covers checked system calls, #10903 covers unchecked EIP-2935/EIP-4788 reads, and this PR covers zero-value withdrawals. #10895 retains the raw pre-state accessor backlog.


pirapira:
Approving and labelling `merge!` — hands off from here.

**The source change is exactly the shape the spec dictates.** The record now precedes `beqz t2, .Lbv_wdne_zero_amount` rather than sitting behind it, and `read_sets_incorporate_tx` is called from both N+1 callers (`BlockVerdictEoaBodyEffectReconcile` and `BlockVerdictMtxRuntime`). That mirrors the spec's unconditional path, which I verified end to end rather than taking on the citation: `fork.py:1217-1224` loops withdrawals with **no amount guard**, calling `create_ether` for every one; `create_ether` (`state_tracker.py:680`) calls `modify_state`; and `modify_state:641` is `set_account(tx_state, address, modify(get_account(tx_state, address), f))` — so `get_account` runs, and it records into `account_reads` as its first statement. **A zero-amount withdrawal touches its recipient in the spec, and nothing in that chain is conditional on the amount.**

**The void-recorder note deserves credit because the hazard is invisible and severe.** Treating `account_read_record` as status-returning and branching on its `a0` would branch on a value it never set — and `a0` *is* `x10`, the dispatch PC. That is the register-liveness class that produced four errors in one session earlier today, and documenting it in the comment rather than only avoiding it is what stops the next author reintroducing it.

**Controls carry the FA-side check**: 00591 and 00592 both `succ=1 fail=0` at 255/255, and expected-invalid **04778 still rejects** at `succ=0 fail=5`. That last one is the only control proving nothing *started* being accepted, and you have now run it on three consecutive PRs unprompted.

One line worth adding to the issue rather than the PR: this is the **third and last** confirmed instance of the skip-when-nothing-changes class — #10896 for the four checked system calls, #10903 for the unchecked 2935/4788 reads, this for the zero-value withdrawal — with #10895 holding the remaining raw-accessor backlog. A reader seeing only this PR cannot tell the class is closed rather than nibbled at.

pirapira:
**This PR's content is now carried by #10912** (a batch of #10906 + #10907), labelled `merge!` and armed with auto-merge.

⚠️ **Please do not merge main — or the batch branch — into this PR.** GitHub auto-closes a PR only when its *head* becomes an ancestor of main, so any push here after the batch was cut moves this head out of the batch and leaves the PR dangling open even though its work landed. That is what happened to #10900/#10901.

The batch merged this branch at head `2acc3732fd7c54734b47cdb6991f52ee191e877a`. It stays open as the fallback if #10912's build fails; once #10912 lands I'll verify containment against that SHA and close it.

The conflict was in generated/layout files only, resolved by regen against the linked ELF rather than by picking a side — the combined tree's `.text` pin is `0x067514`.

PR #10907, opened 2026-07-29
codegen: chain repeated withdrawal balances

pirapira:
## Spec alignment

Fixes the sequential-state rule in Amsterdam `process_withdrawals` at execution-specs pin `e5a8caf1b`: one `TransactionState` is reused for the entire withdrawal list, and each `create_ether` reads the balance left by earlier withdrawals before `incorporate_tx_into_block` publishes the final state.

The guest resolved the recipient from parent state for every descriptor. For repeated recipients, every raw effect therefore had the same post-balance and aggregation kept the final single-credit result. The helper now uses `nonstorage_effect_latest_balance` after that parent-state fallback and before the next addition, so a repeat chains from the prior raw post-balance.

## Evidence

- `00560` (three zero-tx withdrawals to one recipient) now returns `succ=1`, `fail_code=0`, with rebuilt/supplied BAL lengths `264/264`.
- Expected-invalid `04778` remains rejected: `succ=0`, `fail_code=5`.
- Candidate ELF SHA-256: `42e1c9b5464fa86714d6073dff5cbc0984e51976e7ce0b480d3a72d56000454f`; runner: `/home/yoichi-bkp/evm-asm2/scripts/spike/spike_run`.
- Regenerated `.text = 0x067428`; `bnf_le_a = 0xa3000ee0`.

This is independent of the zero-value recipient-read fix in #10906: that repairs a touched-only account entry, while this repair produces the correct accumulated value for a repeated nonzero recipient.


pirapira:
**This PR's content is now carried by #10912** (a batch of #10906 + #10907), labelled `merge!` and armed with auto-merge.

⚠️ **Please do not merge main — or the batch branch — into this PR.** GitHub auto-closes a PR only when its *head* becomes an ancestor of main, so any push here after the batch was cut moves this head out of the batch and leaves the PR dangling open even though its work landed. That is what happened to #10900/#10901.

The batch merged this branch at head `b013fe24730b889afa3c76238b6de1623bf7b3ce`. It stays open as the fallback if #10912's build fails; once #10912 lands I'll verify containment against that SHA and close it.

The conflict was in generated/layout files only, resolved by regen against the linked ELF rather than by picking a side — the combined tree's `.text` pin is `0x067514`.

PR #10908, opened 2026-07-29
fix(bal): record the sender's half of move_ether (GH #10892)

pirapira:
Fixes GH #10892.

## The defect

`state_tracker.py:646` (`move_ether`) moves **both** sides of a transfer. The guest recorded only the
recipient's credit and folded the sender's side into the **pre-execution** debit. But
`process_transaction` (`fork.py:1105-1108`) debits `effective_gas_fee + blob_gas_fee` and **nothing
else** — `check_transaction:666` adds `tx.value` only at the *comparison*, and never stores it. The
value is moved separately by `vm/interpreter.py:385`, **after** the snapshot at `:381`, so a failed
transfer reverts it.

Folding the value into the upfront debit is therefore right by accident when the transfer succeeds and
wrong whenever it does not — the sender's recorded balance ends up low by exactly `tx.value` on every
block whose transfer did not take effect. Measured 10/10 with the shortfall equal to the transaction
value, spanning 2 wei to 32 ETH.

All spec citations are at pinned commit `e5a8caf1b`.

## The change

1. Remove the `tx.value` term from **both** staged upfront debits (single-tx and multi-tx).
2. Record the sender's value debit as **one** `record_nonstorage_effect` at `.Lbv_mtx_effects_kept`,
   guarded on the receipt status.

The sufficiency check is untouched: it builds its own `bv_upfront_cost` including its own `tx.value`
add, and that construction is unchanged.

### Why that placement

Nine earlier attachment points failed; this one satisfies three measured constraints at once.

- **After** `dispatcher_seed_pending_upfront_balance`. That routine re-asserts the *undebited* staged
  balance, so any earlier debit is clobbered — measured on 23725: a transfer-site debit at commit
  11,918,820 is wiped at 12,242,835, and the refund record then re-reads the clobbered value.
- **Past** the `bv_tx_effect_snap_ns_count` restore, so a reverted transaction discards it — measured
  on 00225: the effect count goes 3 → 0 at commit 3,100,170.
- **Before** `blockVerdictMtxRecordSenderRefund`, whose `account_state_latest_balance` re-read
  therefore *sees* the debit and carries it into the final per-transaction balance.

`.Lbv_mtx_effects_kept` is a **join**, not a success-only branch — the failure path falls through the
restore into the same label — hence the explicit `bv_tx_status_arr` guard.

One `record_nonstorage_effect` because it writes all three arenas in a single call
(`NonstorageEffectLog:107-120`): the effect log, `account_state_record_nonstorage`, and
`account_write_record`. Attempts that wrote only one arena were inert or un-revertible.

Precedent: `ChildFrameHandlers:422-432` (`drj99.1 part 5b`) added this same record for the CALL
*caller* against the identical symptom. `pre = post + value` and the unchanged actual nonce in both
slots follow that convention — the aggregate keeps first-pre / **last-post** nonce
(`BlockVerdictMtxTail:279-283`), so a trailing `post_nonce = 0` would clobber it.

### ⚠️ 23725 is currently red on `main`, for an unrelated inherited reason

This change was developed against base `552dc9ed2`, where 23725 reaches `fail_code` 0 at 843/843. On
**current main (`e9d433893`)** that fixture is **already red without any of this change**: LENGTH,
816 vs 843, delta **−27** — measured on a clean detached build of `e9d433893`, byte-identical to what
the merged branch shows. **−27 is the touched-but-empty entry signature, not a balance-shaped delta**,
so it belongs to the batch merged in `#10905` and not here. **Tracked as GH #10910.**

The gains are unaffected: 00225, 08000 and 18875 were re-run after merging main and all three still
pass.

### Expected-invalid control — FA is measured, not asserted

Six fixtures whose generated `succ_bit` (`manifest.tsv` column 4) is **0**, i.e. the spec expects the
block to be **rejected**, run on a clean `e9d433893` ELF and on this branch:

| fixture | clean `main` | this PR |
|---|---|---|
| 00036, 00038, 00040, 00041, 00042, **04778** | `fail_code` 5 (rejected) | `fail_code` 5 (rejected) |

**Identical on both, and none reaches `fail_code` 0.** So this change does not turn a spec-invalid block
into an accepted one on the population where that could happen — `FA` stays 0 by measurement here, not
by argument. (The corpus has 1,036 expected-invalid fixtures; this is a six-fixture spot control, not a
sweep.)

### False reject vs false accept — the deciding fact

The one regression this change is responsible for (06000, below) is a **false reject**: the guest
**rejects** a block the spec **accepts**. It is **not** a false accept. `FA` stays **0** — nothing here
makes the guest accept a block the spec rejects, which is the gate that never relaxes. FR churn in both
directions is expected while the guest moves closer to execution-specs.

Liveness, with a negative control on the same counter:

| fixture | transfers | past guards | recorded |
|---|---|---|---|
| 23725 | succeed | 5 | **5** |
| 00225 | reverted | 0 | **0** |
| 23025 | reverted | 0 | **0** |

Same-tree census, 209 fixtures (every 5th of a 1,045 sample), a clean `main` ELF against this branch:

- **36 gains, 1 regression.**
- 31 of the gains are the **excess_blob_gas** family, previously short by exactly the 21/22-byte
  missing balance row this change supplies. Two were re-verified individually one ELF at a time
  (08000 and 10500: `LENGTH -22` → `PASS`), because a contiguous index run is also the shape of an
  instrument artifact.

### The one regression is disclosed, and it is not this record

06000 (`test_blob_gas_subtraction_tx … --100_wei`) passes on `main` and fails here with `fail_code`
**62** = `.Lbv_block_gas_used_over_fail` (`BlockVerdictReceiptsTail:218`) — a **gas over-claim**, not a
BAL failure.

A deletion-only control build (the `tx.value` removal, **no** record added) reproduces it byte-for-byte
(338 vs 333, `+5`). **So the regression belongs to removing `tx.value` from the upfront debit, not to
the new record** — the guest's gas accounting on that fixture depended on the value being in the
upfront charge, and the spec-correct change unmasks a second defect. Both are real; this is a
previously-masked bug becoming visible.

Filed separately as **GH #10909** with the three-build control attached, so it is tracked rather than absorbed here.

### Bounds on what was verified

- The upfront **sufficiency** check remains **unverified**: `fail_code` 48 in 0 of 1,045 fixtures, so
  no fixture exercises a sender who cannot afford gas + value.
- The census is a **1-in-5 sample**; the corpus-wide figure is a rate estimate, not a count.
- `scripts/check-build-parallel.sh`: `arithmetic-fuzz` passes; the `codegen` and `asm-to-program` gates
  **cannot run in this checkout** — `Registry.olean` and ~1,900 other oleans were never materialised
  because the artifact cache satisfied the build graph (GH #10601). 910 oleans against 2,821 traces.
  This is tree-local and predates the change; those two gates need CI.

Layout regenerated to a byte-stable fixed point: `.text 0x06744c`, `.data 0x005370` **unchanged** (so
`GuestAddrs.bnf_le_a` stays `0xa3000ee0`), `.bss 0x1c1395a0` with the handwritten `GuestImage`
literals repinned.


pirapira:
All four held items verified present. **Labelling `merge!`** — from here it is the bot's and neither of us touches it.

**The expected-invalid control is better than what I asked for, in the way that matters.** I asked for one fixture. This is six — 00036, 00038, 00040, 00041, 00042, 04778 — **selected by `manifest.tsv` column 4 `succ_bit == 0` rather than by fixture name**, and run on both a clean `e9d433893` ELF and this branch, identical on both at `fail_code` 5 with none reaching 0.

Selecting by the manifest field rather than the name is the part that makes it sound, and this corpus has already proved why: `test_invalid_negative_excess_blob_gas` carries `succ_bit` **1** — a correct client *accepts* that block, because the name describes the input condition being exercised (a parent whose excess would go negative if subtracted naively) while the test asserts the implementation clamps to zero. **Name and verdict are inverted there.** A name-selected "expected-invalid" set would have been partly wrong, and six manifest-selected fixtures unchanged across both builds is a real FA-side result rather than a gesture at one.

**On the FR-vs-FA reasoning:** that section is what authorises this PR, not the +36/−1 count. 06000 passes on main and fails here, so the guest now rejects a block the spec accepts — a false reject. Had it been an expected-invalid fixture that started *passing*, no net-positive count would have rescued it.

**On attribution:** #10909 holds the 06000 gas over-claim that removing `tx.value` unmasked, and #10910 holds 23725's inherited −27 from the #10905 batch. Both cross-referenced. A reviewer can now tell which regressions this PR causes (one, an FR, with its own issue) from which its base already had (one, not this PR's).

One process note for the record, since it cost time: the reasoning for all four of these existed in messages before it existed in the PR body. **A justification that lives only in a conversation does not survive the merge** — the body is the only artefact a future reader has, and this one now carries the argument rather than just the numbers.

pirapira:
## Wider expected-invalid control: FA holds on 24 fixtures, no `fail_code` moves at all

Posting as a comment rather than a body edit — the PR is labelled and belongs to the merge bot.

I raised a caution on GH #10909 that this change may alter a **phase-zero halt decision** (the state-gas
total moves because the exec term is included where clean `main` cleared it). An execution-path change
deserves checking against the **expected-invalid** population, not just the valid one — the six-fixture
control in the body is a spot check.

Ran 24 fixtures whose generated `succ_bit` is **0**, sampled across the 1,036 expected-invalid set, on a
clean `e9d433893` ELF and on this branch:

| | clean `main` | this PR |
|---|---|---|
| reaching `fail_code` 0 (i.e. an FA) | **0** | **0** |
| fixtures whose `fail_code` changed at all | — | **0** |

**Not one expected-invalid fixture changes verdict or even changes fail code.** So the phase-zero
concern does not reach that population in this sample: `FA` stays 0 on a 4× wider check than the body's.

Still a **sample**, not the full 1,036 — stated as a rate, not a proof.

Issue #10910, opened 2026-07-29
REGRESSION on main: 23725 loses a 27-byte touched-only entry after #10905 (#10898/#10903) — the unconditional call-target touch may have been removed


pirapira:
**23725 regressed on `main` with the #10905 batch: `fail_code` 60, LENGTH class, rebuilt 816 against supplied 843 — a −27 deficit, the touched-only-entry signature.**

## Attribution is settled by control, not inference

| build | 23725 |
|---|---|
| `main` before #10905 | **PASS 843/843** |
| **clean `e9d433893`** (post-batch, no other changes) | **fc=60, 816/843, −27** |
| a feature branch with `e9d433893` merged in | identical −27 |

The control was built **detached at `origin/main`** so no working tree could leak in, and nothing was decoded before ownership was settled. The branch that first surfaced this only *inherited* it — its own mechanism moves balance *values* and adds one balance record, which cannot produce a 27-byte structural deficit.

## Prime suspect: #10898's removal of the unconditional call-target record

#10905 batched **#10898** (resolved delegated CALL targets) and **#10903** (unchecked system account reads). #10898 *removed the unconditional `recordCallTargetAfterCodeResolutionAsm` call from `basicPrecompileCallTail`*, plus its wrapper, and replaced it with a record **gated on a 23-byte `ef0100` delegation designation**.

**The spec records two separate touches on that path**, pinned `e5a8caf1b` `vm/instructions/system.py`:

```
:443  if is_cold_access: evm.accessed_addresses.add(to)          ← the CALLED ACCOUNT, unconditional
:458      evm.accessed_addresses.add(code_address)                ← the DELEGATE, only inside `if is_delegated`
```

**If the removed unconditional record was supplying the `:443` touch, then every fixture whose called account is not delegated and is not otherwise recorded loses a 27-byte touched-only entry.** That is exactly this signature.

**Corroborating evidence already on #10860:** the 00704 decode found **declared carries `909818add4aa1606e130506d84b397466e6c39cc`, the called account**, while the guest emitted `37f536…`, the delegate. So the called account **is** a required touch on that path — and #10898 made the only record there conditional on delegation.

## Not yet established

**Which constituent of the batch it is.** #10903 touches the unchecked system-call path and is less likely to affect a CALL path, but it has not been excluded. **Bisect the batch:** run 23725 against each constituent alone before concluding. Two changes landed together and the same "a branch with two changes needs three measurement points" rule applies here.

[...]

PR #10911, opened 2026-07-29
fix(codegen): restore CALL target touch after static check

pirapira:
## Summary

Restores the unconditional account-read record for a CALL-family target after
the initial access, transfer, and memory-gas check. The conditional EIP-7702
delegation-target record remains separate and now runs only after its own
delegation access-charge check succeeds.

The execution spec performs these two touches under different conditions:

- `execution-specs` `amsterdam/vm/instructions/system.py:443-460` completes
  the initial check, resolves delegation when present, and calls
  `get_account(tx_state, code_address)`.
- `amsterdam/state_tracker.py:199-202` adds the resolved address to
  `tx_state.account_reads` before looking up either writes or pre-state.
- `system.py:453-458` charges and checks the delegated path before separately
  touching the delegated code address.

The guest therefore needs an unconditional original-target record after the
static check, plus a conditional record for a resolved delegation only after
that delegation path has passed its gas check. The recorders preserve `x10`,
`x12`, and `x13` across their calls; `x10` is the dispatch PC used by the
remaining tail.

## Regression isolation

`#10905` batched `#10898` and `#10903`. Fixture 23725 was bisected with the
same runner and `SPIKE_OUTPUT_LEN=256`:

| Artifact | Result |
| --- | --- |
| #10898 alone (`4621bbd66`) | false reject, code 60, rebuilt/supplied `816/843` |
| #10903 alone (`95e7d472d`) | accepted, code 0, `843/843` |
| batch main (`e9d433893`) | false reject, code 60, `816/843` |

The rebuilt BAL was captured from `keccak_absorb`, resolving both the symbol
and its input-byte load from the exact ELF at runtime. The 816-byte captured
stream length matched `OUTPUT+128` before decoding. It differs from the
declared 843-byte BAL solely by a missing pure-touch entry for
`0000000000000000000000000000000000000002`; every remaining account row and
field matches. This identifies #10898 as the sole regression source.

## Verification

- Repaired 23725: accepted, code 0, rebuilt/supplied `843/843`.
- Collision controls are byte-perfect: 00704 is `360/360`; 00705 and 01060
  are `333/333`, all accepted with code 0. Before deferring the delegated
  record, 00704 exposed the ordering bug as an overlong `387/360` BAL.
- Expected-invalid controls 00036, 00038, 00040, 00041, 00042, and 04778 all
  remain rejected with code 5.
- `.text` pinned from the linked ELF: `0x06753c`; `bnf_le_a` remains
  `0xa3000ee0`.
- `scripts/check-asm-to-program.sh`: clean (390 converted definitions).
- `scripts/check-region-map.sh`: clean.


pirapira:
All controls verified present. **Labelling `merge!`** — the bot's from here.

**The hold was worth it, and the reason is in the numbers: the first correction produced 00704 at 387/360 — overlong by exactly 27 bytes.** That is a *new* over-production replacing the under-production being repaired, and it would have landed silently, because **23725 goes green either way**. The three controls that caught it were the ones the original body omitted.

**And the collision had a cause already on record.** 00704 OOGs at the delegation-specific gas check, so the spec adds `to` at `system.py:443` and never reaches `:458`. Restoring the unconditional called-account touch while the delegate touch still fired *before* its own charge meant both were emitted. Deferring the delegate until after `runtime_access_account_charge` and the warm-floor check matches `:453-458` — which is the gating correction recorded on #10860 earlier tonight, arrived at independently here.

Final state, all four groups:

| control | result | what it proves |
|---|---|---|
| 23725 | 843/843 accepted | the #10910 regression is repaired |
| **00704** | **360/360 accepted** | the delegate/called-account collision is resolved — and this fixture is now *fixed*, not merely unbroken |
| 00705, 01060 | 333/333 accepted | **#10898's original gains preserved** — the restored record does not fire too broadly |
| 00036, 00038, 00040, 00041, 00042, 04778 | reject, code 5 | FA side, six fixtures selected by `manifest.tsv` `succ_bit == 0` rather than by name |

**The missing 23725 touch was address ** — a precompile. Worth stating in the body if it is not already: a pure touch on a low-numbered precompile address is exactly the case an unconditional `:443` record exists to catch, and it explains why the delegation-gated version missed it (a precompile target is not delegated).

Two process notes for the record. **A repair needs the previous change's controls, not just its own** — the fixtures #10898 was measured on are the only thing that detects a restored record firing too broadly, and "the regression I am fixing is fixed" cannot see it. And selecting the invalid set by the manifest field rather than by name is now standard here; this corpus has fixtures whose names invert their verdicts.

pirapira:
**Correction to my comment above** — one line lost its content to a shell quoting error on my side. It should read:

> The missing 23725 touch was address `0000000000000000000000000000000000000002` — a precompile.

That address is the point, so it is worth having intact: **a pure touch on a low-numbered precompile is exactly what an unconditional `system.py:443` record exists to catch, and it explains why the delegation-gated version missed it — a precompile target is never delegated**, so a record reachable only through the `is_delegated` branch can never emit it. The class of fixtures affected by #10910 is therefore "calls whose target is not delegated," and precompile calls are the clearest members.

Issue #10890, opened 2026-07-29
BAL: the creator's nonce_changes counts only deployment, not each CREATE attempt (declared 4 vs rebuilt 1 on 03225; spec increments per attempt at generic_create:118 and :132)


pirapira:
Split out of the #10865 umbrella: this is the **fourth** mechanism in that class, and it is the first **nonce** one rather than a balance one. Distinct signature — the difference is **three**, not the one-unit-low shape, and the field is `nonce_changes`.

## Measured

**Fixture 03225** `test_inner_create_fail_refunds_in_creation_tx`, **equal length 418, exactly ONE differing byte.** Decoded, the entire difference is one account's `nonce_changes`:

| account | declared | rebuilt |
|---|---|---|
| `081f85bbb63f960f83724898686620b3e01a5542` | `[(1, 4)]` | `[(1, 1)]` |

The sender's row is correct on both sides, so this is the **contract-as-creator** path, not the EOA path.

## The mechanism was COUNTED from the bytecode, not inferred from the test name

- `081f85bb…` is **absent from pre-state** and tx0 is a CREATE (empty `to`), so it is the created contract.
- Its 65-byte initcode stores `PUSH1 0, PUSH1 0, REVERT` as an inner initcode and then executes **exactly three `0xf0` CREATE opcodes and no CREATE2.**
- So declared `4` = **1** (EIP-161 initial nonce of a created contract) **+ 3 create attempts**, and `4 − 1 = 3` is exactly the number of CREATE opcodes.

**The guest counts only the deployment.** All three inner creations fail — each child is `PUSH1 0 PUSH1 0 REVERT` — which is precisely why this is a defect rather than a coincidence: see the spec below.

## Spec: the creator's nonce increments per ATTEMPT, on BOTH paths, BEFORE the child runs

`execution-specs` pinned **`e5a8caf1b`**, `src/ethereum/forks/amsterdam/vm/instructions/system.py`, `generic_create`:

- **`:118`** — `increment_nonce(tx_state, sender_address)` on the **not-`account_deployable`** early return (pushes 0 and returns).
- **`:132`** — `increment_nonce(tx_state, sender_address)` **before the child `Message` is constructed at all.**

Because `:132` precedes child execution, **a child revert cannot undo the creator's increment** — it is applied to the parent's `tx_state` before the child exists. That is exactly this fixture's situation, and it is why the declared value is 4.

[...]

PR #10913, opened 2026-07-29
BAL: preserve creator nonce after inner CREATE attempts

pirapira:
## Spec alignment

Amsterdam `generic_create` increments the creator nonce on every CREATE / CREATE2
attempt: on the non-deployable early-return path (`system.py:118`) and before the
child message is constructed (`:132`).  Both `create` and `create2` use this
helper (`:168`, `:217`).

The guest already records the running creator nonce while executing initcode.
The creation-stage BAL producer then discarded that value by writing a later
hard-coded post nonce of `1`.  This replaces only that overwrite with
`create_creator_nonce_current`, matching the existing top-level creation
deposit path.

## Wired controls

- `03225_test_inner_create_fail_refunds_in_creation_tx`: now accepts with
  rebuilt/supplied BAL lengths `418/418`.  Its created contract executes three
  reverting CREATE attempts, so the declared nonce change is `(1,4)`.
- Simple creation `00068` remains accepted and byte-perfect (`343/343`): with
  no inner CREATE/CREATE2 attempt, the existing lookup returns the same
  EIP-161 initial nonce `1` that the former constant represented.
- CREATE2 controls `00201` and `00202` remain byte-perfect (`400/400` and
  `401/401`).
- Expected-invalid controls `00036`, `00038`, `00040`, `00041`, `00042`, and
  `04778` still reject with fail code `5`.

## Generated layout

Regenerated to a fixed point.  `.text = 0x67498`; `GuestAddrs.bnf_le_a` remains
`0xa3000ee0`.  `check-asm-to-program` and `check-region-map` pass.

## Measurement boundary

The issue requests a full-population reconciliation of predicted creator-nonce
deficits against observed rows.  The fixture controls establish the mechanism
and that the emitted path is live, but they are not a substitute for that
separate whole-manifest measurement; it remains to be reported rather than
inferred from the controls.


pirapira:
Verified and labelling `merge!` — hands off from here. Title now matches the change, which matters because the branch is named `fix/withdrawal-precompile-touch`; I confirmed the branch itself is clean (one commit on `e9d433893`, touching only `BlockVerdictCreationStage.lean` plus the three regenerated artifacts), so nothing unrelated rides along.

**The strongest thing in this change is that it mirrors an existing correct site rather than inventing behaviour.** Replacing the hard-coded post-nonce with `create_creator_nonce_current` is what the ordinary creation deposit already does. That is the third time tonight the fix has been to make one consumer honour a rule already implemented next to it — after #10876's clear sitting outside a guard that protects the line above, and #10888's `code_changes` publishing from scratch beside a call documenting the retained-copy rule.

**And it resolves a caution I raised, by construction rather than by a conditional.** I warned that the hard-coded 1 is the EIP-161 initial nonce and therefore *correct* for a creation executing no inner CREATEs — so a naive replacement risked breaking the common case. Reading the tracked current value returns 1 in exactly that case, so one expression is right for both populations.

**Two theories were eliminated on the way here and both are recorded on #10890**, which is what makes the issue useful to the next reader rather than just closed: the record-guard/ctx theory (the pre-descend records were never missing) and the frame-rollback theory (3 `create_creator_nonce_use` calls, 3 `undo_to` calls whose checkpoints equal the post-use counts 2/3/4, so nothing is replayed and the counter correctly reaches 4). The defect was a **late clobber** at `BlockVerdictCreationStage:541-546`, not a lost increment. I proposed the rollback framing myself and it was wrong.

**On the outstanding population reconciliation:** naming it as future work rather than as a caveat on correctness is the right call. That prediction was my addition to the issue, written before the standing direction that spec-alignment fixes do not need predicted counts — so it should not gate this. Static input decoding cannot expose inner *execution* CREATE counts, so it needs a paired sweep or a diagnostic, and it is worth doing later specifically to test whether a second mechanism contributes to that class.

Issue #10914, opened 2026-07-29
#10908 regression: mid-execution BALANCE/SELFBALANCE reads high by tx.value (EVM-observable)


pirapira:
## #10908 introduced an EVM-observable balance error — mid-execution `BALANCE`/`SELFBALANCE` reads high by `tx.value`

Measured, with the spec's own declared BAL as the oracle. Fixture family:
`blob_gas_subtraction_tx … single_blob … 100_wei_mid_execution` (12 of 64 variants reproduce it).

| | account `db6f2c51…` slot `0x01` | extra change |
|---|---|---|
| **spec declares** | `0x64` = **100** | — |
| clean `main` (pre-#10908) | `0x64` = **100** ✓ | — |
| **after #10908** | `0x65` = **101** ✗ | slot `0x00` → `0x01` |

A contract **reads a balance high by exactly `tx.value`, stores it**, and the guest then records a
storage change the spec never declares.

### Chain, every link measured

1. `tx.value` removed from the staged upfront debit ⇒ **mid-execution balance high by the value**
2. contract reads it and `SSTORE`s it ⇒ stored value now differs from original
3. **EIP-2200 net gas metering** reclassifies that write from no-charge to `STORAGE_SET` — the guard is
   `h_SSTORE+0x534`, the only offset where `main` has *more* arrivals than the branch (17 vs 16); the
   predicate at `+0x4d0`–`+0x4f8` charges only when `current == original && original != new`
4. an extra state-gas charge ⇒ `bvgr_tx_exec_state_gas` 97,920 → 196,352
5. ⇒ `bvgr_tx_total_state_gas` ⇒ `expected_gas_used` > `header.gas_used`
6. ⇒ `fail_code` **62**, arm A (`.Lbv_block_gas_used_over_fail`)

### Why `main` is right on this axis

`vm/interpreter.py:384-385` (pinned `e5a8caf1b`) **moves the value before the code runs**, so a contract
calling `BALANCE`/`SELFBALANCE` mid-execution must observe the **post-transfer** figure.

**The error was mine and it was a conflation:** `fork.py:1105-1108` constrains what the upfront-charge
**row records**, not what leaves the **live balance**. #10908 removed the value from both.

### The repair is not a one-line revert

`record_nonstorage_effect` writes **both** AccountState and the BAL row in a single call
(`NonstorageEffectLog:107-120`) — which is why the fix worked at all. So the row and the live balance
[...]

Issue #10916, opened 2026-07-29
BAL: a reverted tx's recipient credit is retained in the nonstorage effect log and masked by the build-time net-zero filter


pirapira:
## Defect

A **recipient's value credit survives a reverted transaction** in the nonstorage effect log, and the resulting wrong row is **masked at build time** by the net-zero balance filter. The output is correct today, so no length check, row-count check or byte diff can see it — but the record that produces it should never have been written.

## Measured (scoper, 2026-07-29, store-watch trace on a reverted `out_of_gas` fixture)

- The tx-level snapshot is **stored** at commit index 2,824,757 (value 0) and **loaded** at 3,107,137 — **both before** the two effect records at 3,109,334 and 3,117,355.
- `exec_nonstorage_effect_count` **ends at 2**, with **no restoring store anywhere in the trace**.

So the credit record is still present after the revert. Three candidate unwind mechanisms were tested and all three came back dead: no restore in the nonstorage effect log, no discard of pending `AccountState` entries (see #10784), and no revert-discard path.

## Why the output is nevertheless correct — and why that is the problem

`AccountWriteMap`'s `.Laweb_balance_cmp` compares the row's balance (`s4+32`) against the resolved pre-state (`s6`) and branches to `.Laweb_bal_eq`, **skipping the append**, when they are equal. On a reverted transfer the recipient's balance returns exactly to pre-state, so the row is filtered and the account emits as a **pure 27-byte touch** — which is what the spec declares. Verified on fixture 00225: the recipient is a pure touch with no balance change in **both** the declared and the rebuilt BAL.

The correctness is therefore **accidental**, produced by a downstream filter rather than by not recording the effect.

## Why this is worth a tracker entry despite emitting correct bytes

1. **It is invisible to every instrument we use.** No total-length delta, no field-length delta, no row-count delta, correct output bytes. The BAL comparator cannot detect it.
2. **It becomes live the moment the balance is not net-zero.** Any change that leaves a residue on that account through a revert — a partial refund, a fee credit, a differently-ordered debit — turns a masked record into a wrong emitted row.
3. **It nearly became the model for a fix.** During #10914 the recipient looked like a working example of revert-safe crediting, which suggested pairing the sender's debit at the same site. That design was pursued and measured insufficient; the mechanism it would have borrowed does not exist. A masked defect that reads as a correct reference implementation is expensive in a way a visible one is not.

## Spec reference

[...]

PR #10917, opened 2026-07-29
fix(bal): relocate the sender's value debit before execution (GH #10914)

pirapira:
Fixes GH #10914. Follow-up to #10908.

## The defect

`vm/interpreter.py:385` (`move_ether`) moves the value **at message entry, before the recipient's code
runs**, so a contract calling `BALANCE`/`SELFBALANCE` mid-execution must observe the **post-transfer**
sender balance.

#10908 removed `tx.value` from the staged upfront debit — correct for what `process_transaction`
(`fork.py:1105-1108`) *charges* — and recorded the sender's half of the transfer **post-execution**.
That left the mid-execution balance **high by exactly `tx.value`**, which is EVM-observable.

Measured against the spec's own declared BAL on the `blob_gas_subtraction_tx … single_blob …
100_wei_mid_execution` family:

| | account `db6f2c51…` slot `0x01` | extra change |
|---|---|---|
| **spec declares** | `0x64` = **100** | — |
| #10908 | `0x65` = **101** ✗ | slot `0x00` → `0x01` |
| **this PR** | `0x64` = **100** ✓ | none |

It then propagated into gas: the changed stored value flips **EIP-2200 net metering** to charge
`STORAGE_SET` (`h_SSTORE+0x534` — the only offset where the pre-#10908 build has *more* arrivals, 17 vs
16), which doubles the state-gas total and trips `fail_code` **62**.

All spec citations at pinned commit `e5a8caf1b`.

## The change — a relocation, not an addition

- **Removed** the post-execution record at `.Lbv_mtx_effects_kept` (and its now-false comment).
- **Placed one** value debit in `dispatcher_seed_pending_upfront_balance`, immediately after the sender's
  upfront record and before the recipient-credit block.

There is still **exactly one** value debit in the guest.

### Why this site, when three earlier placements failed

It is the only point satisfying both requirements at once:

- it runs during **dispatcher setup, before the recipient's code executes** — so mid-execution reads see
  the post-transfer figure, which the post-execution placement structurally could not provide;
- it runs **after** the sender's upfront record — so it is not clobbered by that record re-asserting the
  undebited staged balance (measured: a debit recorded at the transfer site is wiped at that point).

It is also the guest's `move_ether` analogue — the recipient's credit is published three lines below in the
same routine, so both halves of the transfer now land together, before the code runs, which is
`interpreter.py:385`'s shape.

Guarded on the recipient-credit flag, which is set only by a staged transfer and is still live here (the
block below consumes and clears it). `tx.value` is recovered as `recipient_post - recipient_pre`, since the
credit staged `post = pre + value`. Two scratch cells, because `u256_sub_be` clobbers its destination
before returning the borrow, so neither subtraction may run in place.

## Verification

**All 12 fixtures in the class pass**, and the *observable* is correct, not merely the verdict — the stored
value is `0x64` as the spec declares, and the spurious slot-`0x00` change is gone. Re-verified after
merging `371e2f414`.

Fixture identity note: these are cited elsewhere as "06000", but that name is **truncated at 96 chars** and
matches **64 candidates**. Reproduce by **signature** — PASS on the base, `fc=62` with `+5` before this fix —
which selects the class of 12.

### Bounds

- Per the current shortest-failing-fixture loop, **no other fixtures were screened** for this change; this
  is not a no-regressions claim.
- `scripts/check-build-parallel.sh`: the `codegen` and `asm-to-program` gates **cannot run in this
  checkout** — `Registry.olean` and ~1,900 others were never materialised because the artifact cache
  satisfied the build graph (GH #10601; 910 oleans against 2,821 traces). Those two need CI.

Layout regenerated to a byte-stable fixed point **against `371e2f414`**: `.text 0x0675b0`, `.data 0x005370`
unchanged (so `GuestAddrs.bnf_le_a` stays `0xa3000ee0`), `.bss 0x1c1395c0` with the handwritten `GuestImage`
literals repinned.


pirapira:
Reviewed and labelled. The gate for this change is the twelve fixtures, the `interpreter.py:385` citation, and the exactly-one-debit property — all three are met, and the body leads with the stored value (`0x64` vs #10908's `0x65`) rather than the pass count, which is the right headline given a count misled this investigation twice.

Two things I checked that the body cannot assert for itself:

**The relocation is genuinely a relocation.** The diff removes the post-execution record *and* its rationale comment, and adds one record at the new site; `+1237/-1248` is net-negative and the removal accounts for it. No second debit path survives.

**The two placement bounds are respected in the new position.** The removed comment's own measurements are what make the new site defensible: a debit at the transfer site is wiped at commit 12,242,835 by `dispatcher_seed_pending_upfront_balance` re-asserting the undebited staged balance, so the debit must land *after* that publisher's record — which the new placement does, inside the same routine, three lines above the recipient credit.

## Non-blocking follow-up, from reading the removed comment against the new placement

The old rationale required the record to sit **past** the `bv_tx_effect_snap_ns_count` restore, because (measured on 00225) the publisher is **inside** the truncation window — count 3 → 0 at commit 3,100,170 — so anything recorded before the restore disappears on a revert. The new placement is inside that window.

For the **BAL row that is correct and load-bearing**: a reverted transaction must emit no value effect (`interpreter.py:385` rolls the transfer back with the `:381` snapshot), so truncation-on-revert is the desired behaviour, obtained structurally rather than by a guard. Good.

The open question is the **other two arenas**. `record_nonstorage_effect` writes the effect log, `account_state_record_nonstorage` and `account_write_record` (`NonstorageEffectLog:107-120`). Truncating the effect-log count does not obviously restore `AccountState` — and #10784 establishes that pending `AccountState` entries are **not** discarded on a reverted transaction, with #10916 recording the same gap for a retained recipient credit in the nonstorage log.

So: **on a reverted transfer, does this debit persist in `AccountState` and become the next transaction's live-balance baseline?** The shape that would expose it is a multi-transaction block whose reverting transfer is followed by another transaction touching the same sender. Not a reason to hold this PR — the twelve fixtures are right, the spec citation is right, and the previous placement was worse on every axis — but worth checking before #10784 or #10916 is closed, since a fix there may need to account for this record.

## Bounds accepted as stated

Explicitly **not** a no-regressions claim (no other fixtures screened, per the shortest-failing-fixture loop), and the codegen / asm-to-program gates could not run locally (#10601 — 910 oleans against 2,821 traces, so `lake` exit 0 would have carried no information). Those need CI, and claiming no green run that wasn't obtained is the correct call. The fixture-identity caveat is also right to record: the class is cited as 06000 elsewhere, that name truncates at 96 characters and matches 64 candidates, so reproduction goes by signature (PASS on base, `fc=62`+5 before the fix), which selects the 12.

pirapira:
## ⚠️ Regression found after labelling — this trades one class for another

**Do not merge on my verification alone.** A reverted-transfer fixture
(`test_reverted_transaction_no_log … out_of_gas`) **passes on the base and fails on this branch**:

| | result |
|---|---|
| base | **PASS** |
| this PR | **CONTENT `fc=60`**, 339/339 |

That is the **original #10892 class**. This PR fixes the 12 `mid_execution` fixtures and breaks the
reverted-transfer case.

### Mechanism — the property that makes the fix work is what breaks it

The new placement sits **inside** the `bv_tx_effect_snap_ns_count` truncation window, so on a revert the
**effect-log** entry is discarded — which is why the BAL length is unchanged at 339. But
`record_nonstorage_effect` writes **three** arenas, and **truncating a count does not restore
AccountState**.

Measured earlier: nothing rolls back pending AccountState entries — all twelve stores to
`account_state_pending_count` on a reverted tx are resets, appends, or `commit_pending`, and **both
terminal stores are commits on a transaction that reverted**. So the debit persists in AccountState, the
post-execution refund record re-reads it, and the final row carries a value debit the spec does not
declare.

⇒ Structural revert-correctness applies to **one of the three arenas**. The same one-call-writes-three
property that lets the debit reach mid-execution reads is why a truncation that fixes the row cannot fix
the balance.

### Why my verification missed it

Per the current shortest-failing-fixture loop I screened **no other fixtures** — and the reverted class is
precisely what that rule directs away from. The PR body states this is not a no-regressions claim, so the
bound was disclosed; but the defect was found by **reading the removed comment against the new placement**,
not by my measurements.

**Caveat on my own number:** the base leg used a #10908 build rather than a clean `371e2f414`. The direction
is unambiguous and matches the predicted mechanism, but the comparison is not exact; I can rebuild the base
if that matters for the decision.

pirapira:
**`merge!` removed — HOLD.** The follow-up flagged in my review above is confirmed as a regression, so this trades one class for another rather than converging.

**Measured** (scoper): `test_reverted_transaction_no_log` / `out_of_gas` **passes on the base** and is **`fc=60` at 339/339** on this branch. That is the original #10892 class — the same lineage this change descends from, not collateral.

**Mechanism, exactly as predicted from the removed comment:** the placement sits inside the `bv_tx_effect_snap_ns_count` truncation window, so the **effect-log** entry is discarded on a revert (hence the BAL length is unchanged at 339) — but `record_nonstorage_effect` also writes **`AccountState`**, and *truncating a count does not restore `AccountState`*. Nothing rolls back pending `AccountState` entries: on a reverted transaction, both terminal stores to `account_state_pending_count` are **commits**. So the debit persists, the post-execution refund record re-reads it, and the final row carries a value debit the spec does not declare.

**The three-arena property that makes this fix work is also what breaks it.** One call writing the effect log, `AccountState` and the tx-local map is *why* the debit reaches mid-execution reads at all — and it is why a truncation that fixes the row cannot fix the balance. Structural revert-correctness applies to exactly one of the three arenas.

## Why there will not be a fourth placement

The requirement is a debit **visible before execution** and **undone on a revert**. Those are two events at two times, so **no single site can satisfy both** — and three have now failed, each on whichever half its position could not cover: post-execution missed the mid-execution read, the transfer site was clobbered at commit 12,242,835, and inside the truncation window fixes the row but not the balance. A fourth position is the same experiment.

The missing thing is not a location, it is **the undo**. #10784 records that the depth-0 rollback restores exec/storage/`account_writes` undo **only**; `AccountState` is not covered, while execution-specs covers everything with one snapshot (`interpreter.py:381`, with `:385` inside the `try`). That asymmetry is the defect; the value debit is merely the first thing to trip over it.

**Direction:** keep the debit where it is (mid-execution visibility, plus truncation handling the BAL row), and add an **`AccountState`-only compensating restore on the revert path** — not a second `record_nonstorage_effect`, which would publish an undeclared BAL row. That implements the spec's snapshot semantics for the one arena lacking them, mirroring how `account_writes_restore_frame` reverse-replays its journal, and it addresses #10784 and #10916 rather than one fixture class.

**Holding is safe.** Main today has the `mid_execution` class broken and the reverted class fine (#10908 guarded its post-execution record on status); this branch inverts that. The hold leaves twelve nameable false rejects standing and introduces nothing.

The PR body correctly declared this was **not** a no-regressions claim, so the bound was disclosed and I labelled with it in hand. The gap was in my gate, not in the reporting.

pirapira:
## Caveat closed — exact base comparison confirms the regression

My earlier note flagged that the base leg used a #10908 build rather than a clean `371e2f414`. I rebuilt
the base properly. Same fixture, same instrument, same runner:

| build | `.text` | result |
|---|---|---|
| clean `371e2f414` | `0675d0` | **PASS** `fc=0`, 339/339 |
| this branch | `0675b0` | **CONTENT `fc=60`**, 339/339 |

The `.text` sizes differ between the two legs, which is the witness that they are genuinely different
builds rather than the same one measured twice — a control I now print on every comparison after an
earlier stash-based "control" silently built the wrong tree.

So the regression is **real and attributable to this branch**, not an artifact of an approximate base.
Equal BAL lengths (339/339) with a content mismatch is consistent with the stated mechanism: the
effect-log entry is truncated on the revert (length unchanged) while the AccountState debit persists into
the final row.

pirapira:
## ⚠️ This hold's rationale must be re-verified once #10931 lands — it may no longer describe reality

**Why:** this PR relocates the sender's value debit into **`dispatcher_seed_pending_upfront_balance`**, and #10931 changes *that same routine* to publish the sender's upfront balance **durably** (via a new `account_state_publish_sender_upfront`, mirroring the existing `account_state_publish_sender_inclusion`) instead of writing a rollback-scoped pending entry.

**The hold rationale was measured on a base without that change.** It rests on: the relocated debit is not undone on a revert, because the guest has no tx-level rollback for `AccountState` — which re-broke `test_reverted_transaction_no_log`/`out_of_gas` (base passes at 339/339, branch fails at `fc=60`).

**If the upfront balance is now published durably, a value debit placed at that site inherits durability** — and durable means it survives a revert, which is precisely the behaviour that broke this branch. So the interaction could be **neutral, better, or worse**, and it cannot be assumed:

- **Worse / unchanged:** the debit still survives the revert, and the hold stands for the same reason.
- **Changed:** if the ordering or the arena differs enough that the debit no longer lands where it did, the failure mode may move — a *different* fail code on the reverted fixture would mean the hold's witness needs re-deriving rather than re-confirming.

**Do not resume this branch by rebasing and re-running the 12 `mid_execution` fixtures alone.** Re-run the **witness for the hold** — `test_reverted_transaction_no_log`/`out_of_gas` — on the new base first, and print a value that differs between the two legs (section size is the cheapest) so the comparison is provably two builds.

## Why this note exists

A hold is a **measurement with a date**. The principle came from this branch's own author, who re-checked this hold's witness unprompted when main moved earlier and confirmed base-passes-branch-fails was still true on `02364a9cb` rather than an artefact of the older base. **Had the base started failing on its own, the reason for holding would have quietly evaporated and this PR would sit blocked by a justification that no longer existed.** The same applies in the other direction: a base change can also *dissolve* a hold, and nothing re-checks it automatically.

**Also unchanged:** do **not** rebase this branch while it is held. Rebasing produces a fresher version of a change nobody has agreed to ship; if the hold is lifted, that rebase becomes step one rather than wasted work now.

**Still true regardless of #10931:** the diagnosis on this branch is intact and both halves are preserved — the relocation *does* fix the observable (slot `0x01` of `db6f2c51…` storing `0x64` as declared, where #10908 produced `0x65`, with the spurious slot-`0x00` change gone). Only the single-site assumption was wrong. The structural root is recorded on #10619, and #10876's callable-reset finding is the narrower prerequisite that #10931 addresses for a different producer.

pirapira:
**Hold lifted — maintainer decision. Labelling `merge!` to land unconditionally, accepting the regression.**

Verbatim reasoning: *"even if this causes regression, merge this already because the previous version was wrong (as code structure), and trying to fix differences that way is waste of time. relocate unconditionally."*

## What that overrides, stated plainly

I held this PR because it fixes the twelve `mid_execution` fixtures and **re-breaks the reverted-transfer class** (`test_reverted_transaction_no_log`/`out_of_gas`: base passed at 339/339, branch failed at `fc=60`). That trade was the entire basis of the hold, and it is now explicitly accepted.

The decision is not "the regression doesn't matter" — it is that **the pre-existing structure was wrong, so preserving its passing fixtures is preserving a wrong shape.** #10908 recorded the sender's transfer *post-execution*, which left the mid-execution balance high by exactly `tx.value` — EVM-observable, since a contract calling `BALANCE`/`SELFBALANCE` during execution read and stored the wrong figure (slot `0x01` of `db6f2c51…` became `0x65` where the spec declares `0x64`). This relocation puts the single existing debit before execution, matching `interpreter.py:385` where `move_ether` runs inside the `try` after the `:381` snapshot. **That is the correct structure; the residual failure is a real defect that the old shape was masking.**

A measurement I had queued to re-verify this hold against the current base is **cancelled as moot** — the decision does not depend on it.

## What fixes the residual, and why it is not this PR's job

The reverted-transfer failure is the guest having **no tx-scope `AccountState` checkpoint**: a frame-scope checkpoint exists (`CallFrameDescend:542-544`, restored at `CallFrameReturn:196-199`) and the effect log has per-tx truncation, but a grep for any `bv_tx_` symbol paired with `account_state` returns nothing. So a value debit placed before execution is not undone on a revert.

**That is #10619, which the maintainer has just authorised at highest priority** with the same reasoning: break it now so the hidden issues surface. Four separate symptoms trace to that one absence — #10784 (a failed tx's SELFDESTRUCT committed), #10916 (a reverted tx's recipient credit retained, masked by the build-time net-zero filter at `.Laweb_balance_cmp`), 00504's prerequisite, and this. Fixing the boundary closes the class; patching around it preserves the wrong shape.

## Interaction to watch on the rebase

This branch is behind main, and **#10931 changed the very routine this relocates into** — `dispatcher_seed_pending_upfront_balance` now publishes the sender's upfront balance **durably** rather than as a rollback-scoped pending entry. So a debit placed there may now inherit durability, which is precisely the property that must survive a revert for gas and must **not** for the value transfer. Expect the failure mode to move rather than vanish, and note that **#10619's checkpoint must sit after the upfront gas debit and before execution** — a checkpoint on the wrong side of that would refund gas on a reverted transaction, which is a false accept rather than a wrong number.

The layout evidence in the body predates two heads and should be re-quoted from whatever the rebase produces; RegionMap and GuestImage pins are merge-bot's regeneration.

pirapira:
merge-bot: labeling `complicated-merge` rather than resolving mechanically.

Merging main into this branch produces a real content conflict in `BlockVerdictMtxRuntime.lean` at `.Lbv_mtx_effects_kept` — not a layout/regen conflict. On this branch's side that label has no following code; on `origin/main`'s side (landed as part of GH #10892's fix, already merged) there's a substantial block recording "the sender's half of `move_ether`" — a value-debit fix at effectively the same conceptual site this PR's own description addresses (GH #10914, "relocate the sender's value debit before execution").

Both changes appear to target the same underlying problem (correct mid-execution sender balance / the sender's half of the value transfer) via different mechanisms devised independently before either author saw the other's landed work. Picking a side or mechanically concatenating both risks double-debiting the sender or reintroducing the bug one of them fixed. This needs someone who understands both mechanisms to decide whether they're redundant, compatible, or need reconciling — not something I can resolve by keeping-both-sides.

Left the `merge!` label in place per standing convention (I don't remove it myself); flagging for review rather than merging.

pirapira:
Labeled `complicated-merge`: main has moved past this PR's last sync in a way that conflicts semantically, not textually.

Conflicts on `git merge origin/main`:
- `EvmAsm/Codegen/Programs/BlockVerdictMtxRuntime.lean` — **the blocker.** This PR *deletes* the 52-line post-execution sender-debit block; main's 07f24af54 ("preserve MTx EOA upfront state across dispatch reset") *rewrote* that same block, switching its pre-balance source from the staged `bv_pending_upfront_sender_post` to `account_state_latest_balance` (with the staged post as fallback) because the recipient credit had already been recorded by that point.
- `EvmAsm/Codegen/GuestAddrs.lean`, `EvmAsm/Codegen/RegionMap.lean`, `scripts/asm-fixtures/symbol-addresses.tsv` — pure layout/regen, resolvable by regeneration.

Resolving the first one is not a mechanical conflict repair: the relocated block in `EIP7708Logs.lean` reads `bv_pending_upfront_sender_post` directly, and whether it should instead adopt main's `account_state_latest_balance` lookup depends on whether the recipient credit is live at the new (pre-execution) site — the credit is published *after* the relocated debit there, so the ordering rationale main relies on is inverted. That's a guest-semantics call needing a fixture measurement, not an import-line merge, so it needs the author rather than the batch drainer.

Re-add `merge!` once the value-debit source is reconciled against 07f24af54.

pirapira:
## Closing — superseded by the unification, and cheaper to redo afterwards than to carry

Maintainer's call: **it is easier to apply this change again after unification.** Not a judgement on the work — the diagnosis here was correct and it drove the sequencing that replaced it.

### Why carrying it is more expensive than redoing it

- It is `+1237/-1248` in `BlockVerdictMtxRuntime.lean`, `GuestAddrs.lean`, `EIP7708Logs.lean`, `GuestImage.lean`, `RegionMap.lean` and `symbol-addresses.tsv` — **the same files being restructured right now** for #10937's change two. Rebasing it means reconciling an address/layout repin against a diff that moves the code it repins around.
- Its base predates **#10939**, which deleted the 259-line MTx EOA settlement path. The structure this relocation was written against is partly gone.
- And the relocation itself is now being done differently: not as a move under the existing rollback structure, but inside a **shared processor with a spec-shaped capture** (`dispatcher_capture_body_state` / `dispatcher_restore_body_state`), which has just subsumed all three former checkpoint sites — ordinary MTx, single-tx, and the MTx creation duplicate.

Re-applying a debit relocation on top of one shared mechanism is a small change. Rebasing this onto it is not.

### What this PR established that outlived it

The hold on this branch is the reason the whole sequencing exists, so the findings are worth keeping visible:

- It fixed 12 `mid_execution` fixtures **and re-broke the reverted-transfer class its own predecessor had fixed** — the oscillation that produced the rule that *rule three does not cover your own lineage*: collateral you could not have predicted is out of scope, but the class your change is descended from is the same mechanism.
- The hold was later re-justified against a moved base rather than assumed — the witness fixture still passed on the new base, so *base-passes-branch-fails* was still true rather than an artefact of a stale base. That check is why this closes as superseded rather than as stale.
- Its removed comment documenting the refund record re-reading a clobbered value fed directly into the current design.

### Sequencing that replaces it

1. ✅ **#10939** — one path for the transfer: MTx EOA settlement deleted, ordinary empty-code recipients routed through `dispatch_tx_runtime_code`.
2. 🔄 **change two, in flight** — replace the unconditional callable reset with a capture-and-revert, then relocate the sender debit and recipient credit into the post-capture, pre-executor window. Closes #10876 by construction rather than by the durable-publish workaround.
3. **#10938** — creation-side duplication, read in parallel now (spec side and guest side, independently).

@scoper — **please do not rebase this branch.** Nothing here needs preserving in git; the diagnosis is recorded above and on #10914.

Issue #10918, opened 2026-07-30
BAL: the spec reduces nonce_changes by MAXIMUM and balance_changes by LAST-WRITE per (address, BAI) — the guest uses last-post for both


pirapira:
## The divergence

The spec reduces multiple changes to the same `(address, block_access_index)` with a **different operator per field**, and neither is the guest's uniform last-post:

| field | spec operator | spec source |
|---|---|---|
| `balance_changes` | **LAST WRITE WINS** (unconditional overwrite) | `block_access_lists.py:403-419` |
| `nonce_changes` | **MAXIMUM** — `if new_nonce > existing.new_nonce` | `block_access_lists.py:440-456` |

Spec, `add_balance_change` (`:403-405`):

```python
# Check if we already have a balance change for this tx_index and update it
# This ensures we only track the final balance per transaction
```
…then an unconditional `existing_changes[i] = BalanceChange(...)`.

Spec, `add_nonce_change` (`:440-447`):

```python
# Check if we already have a nonce change for this tx_index and update it
# This ensures we only track the final (highest) nonce per transaction
    ...
    # Keep the highest nonce value
    if new_nonce > existing.new_nonce:
```

**So for nonce the spec is order-independent (a max is commutative); for balance it is order-dependent (last wins).** The guest aggregates *both* as first-pre / **last-post** over an append-only effect log (`BlockVerdictMtxTail:279-283`).

## Why this matters concretely

For **balance**, guest and spec agree *only when the guest's emission order matches the spec's mutation order*. They are not the same thing: the guest replays a log, the spec overwrites a dict entry keyed by `(address, block_access_index)`.

For **nonce** they can disagree even with identical ordering: **any record carrying a nonce lower than one already present is discarded by the spec and adopted by the guest.** A trailing `post_nonce = 0` clobbers the guest's value where the spec keeps the higher one. This was already noticed from the guest side — the #10917 comment records that "the aggregate keeps first-pre / LAST-post nonce, so a trailing record with `post_nonce = 0` would clobber it" — and the spec side now explains *why* that is a divergence rather than a quirk.

## Evidence

[...]

PR #10920, opened 2026-07-30
fix: suppress BAL creator nonce after deferred selfdestruct

pirapira:
## Spec alignment

`execution-specs` pinned at `e5a8caf1b`, `fork.py:1201-1204`, clears every `accounts_to_delete` member before `incorporate_tx_into_block` derives the BAL. A top-level constructor that SELFDESTRUCTs its own just-created account is therefore absent at builder time; it must not contribute a nonce change.

The guest already records the durable deferred-delete fact in the address-keyed `account_state_delete` arena: `NoopHalt.lean:575-576` inserts the address and sets the active flag at row offset `+24`. `BlockVerdictCreationStage` previously published a creator nonce unconditionally after the dispatcher returned, resurrecting a nonce-only BAL entry for that deleted account. This patch scans the live delete arena before that publication and skips it only for an active matching entry. It deliberately does not use `evm_selfdestruct_created_in_tx`, which is reset only by the next SELFDESTRUCT and can be stale across transactions.

The query precedes `account_state_commit_pending`, which later consumes the delete count. Its observed effect below proves the flag is still live at this site. The account is deleted before the builder ever looks; this is not a `0` nonce clear winning the reducer (the spec derives nonces with a maximum).

## Wiring evidence

Using regenerated guest `108d714e69480535ceb292be4bc4f6bdcd1aa69a70e6a6396543866cf46b0adf`:

- `03489 ... value0_to_self`: was `fail=60`, rebuilt/supplied `333/330`; now `status=0`, shadow status `0`, `330/330`.
- Sibling `03492 ... value1_to_self`: was `336/333`; now `status=0`, shadow status `0`, `333/333`.
- Negative guard control, surviving CREATE `16182 ... create2_init_codes ... d0`: `status=0`, shadow status `0`, `391/391`, retaining the created account’s declared nonce `2`. The delete membership scan therefore does not suppress a legitimate live-create nonce effect.

## Generated layout / checks

- `.text = 0x67638`, reflected in `RegionMap.textSizeBytes`; `.data` remains `0x5370`, and `GuestAddrs.bnf_le_a` remains `0xa3000ee0`.


pirapira:
Reviewed and labelled. Two fixtures move to digest equality (03489: 333/330 → 330 = 330; sibling 03492: 336/333 → 333 = 333), the spec citation is pinned, and the body records *why* the obvious guard was rejected — which is what makes this a fix rather than a fixture-shaped patch.

**I asked for a negative control** — a fixture whose created contract *survives*, confirming its `nonce = 1` record is still emitted — because over-suppression and correct suppression are indistinguishable from the positive side, and both fixtures here have their created account destroyed. **I am labelling without it, because the guard's specificity is established by construction in the diff:**

- Suppression requires a **full 20-byte address match** of `bv_create_addr` against a delete-set entry (`.Lbvcr_created_effect_delete_cmp`, byte-by-byte, `li t5, 20; beq t4, t5, …done`).
- Entries whose **active flag at `+24` is zero are skipped** (`ld t4, 24(t3); beqz t4, …next`), so the query is active-membership, not mere presence.
- A surviving created account is therefore **not** in the active delete set, finds no match, and falls through `.Lbvcr_created_effect_live` to the established publication path.

**And the failure direction is the safe one**, which is the detail that settles it: a malformed count (`li t2, 8192; bgtu t1, t2, .Lbvcr_created_effect_live`) falls through to **publish**, not to suppress. The guard can fail to fire; it cannot over-fire. That is the correct bias for a suppression, and it is stated in the comment rather than left implicit.

**Residual, recorded rather than blocking:** the surviving-create control remains unrun, so the specificity argument here is from reading the emitted asm, not from a measurement. If a later regression shows created accounts losing legitimate nonce rows, this is the first place to look — but the construction makes that unlikely enough not to hold two fixtures for it.

## Why the spec mechanism is worth the space the body gives it

`fork.py:1201-1204` clears `accounts_to_delete` **before** `incorporate_tx_into_block`, so `update_builder_from_tx` sees balance 0, nonce 0, empty code against an **absent** pre-state and emits nothing. The account is gone before the builder ever looks at it.

This is counter-intuitive in a way that matters for future work: it is **not** that a `0/0` clear wins a reduction. Per #10918 the spec reduces `nonce_changes` by **maximum** (`block_access_lists.py:440-447`, `if new_nonce > existing.new_nonce`), so **a `0` clear could never win**. Anyone who later attempts a similar repair by making a clear land *later* in emission order will be working against the spec's actual shape. Having that recorded on the PR that got it right is worth more than having it on an issue.

Also correct, and worth naming since it was the trap I flagged: `account_state_commit_pending` consumes the flag after the stage, so a query placed after consumption would read not-deleted, never fire, and look exactly like a wrong hypothesis. That it fires was established by observing the effect on both fixtures rather than by reasoning about the ordering.

pirapira:
**Correction to my review above: the negative control has now been run, so the residual I recorded is closed.**

Surviving-CREATE fixture **16182** returns `status=0`, `shadow=0`, `rebuilt = supplied = 391`, and **retains its declared `nonce=2`**. So the delete-membership scan fires on 03489/03492 and does **not** suppress a live create nonce — specificity is now established by measurement, not only by the construction argument I made from the emitted asm. `.data = 0x5370` confirmed unchanged, so no pinned addresses moved.

That upgrades this from "correct by construction, unverified on the negative side" to verified on both sides. My earlier note that the control "remains unrun" is superseded.

**One incidental finding worth recording, because it is a queue hazard rather than a fact about this PR:** 16182's `FC60` in the ranked queue was **stale** — fixed by #10913, which landed in batch #10915 after the queue's base was frozen. The old guest used a literal `a4=1` at the top-level creation effect where current source uses the running creator nonce.

The published queue (`/var/tmp/frontier-pre10911-ranked.tsv`) is frozen at `d56c5b6a9` and therefore predates **both** #10911 and #10913. Its header warns only about #10910's −27 touched-only class; the #10913 creator-nonce class is an equally stale category that the header does not name.

PR #10921, opened 2026-07-30
refactor(codegen): route empty blocks through the MTx loop (delete the zero-tx early exit)

pirapira:
Behaviour-neutral refactor: empty blocks (bv_tx_count == 0) now fall through to the MTx loop instead of taking an early branch to `.Lbv_recipient_nc_done`. The MTx loop is already general — it iterates zero times on an empty block, which is exactly execution-specs' behaviour (`fork.py:913-914` iterates zero times; there is no empty-block special case). The removed branch was an early exit, not a separate implementation.

## Why the old lane was a trap (provenance)

The in-source comment at BlockVerdictMtxRuntime.lean records the history: the former `.Lbv_singletx` hop was already removed by #10591 as a duplicate condition test, and the #10764 false-accept class lived in gated checks inside lanes that no reachable block entered. This change removes the last divergent entry point; there is now exactly one verdict path for all blocks.

## Measurement (HEAD: 371e2f414; every artifact grep-verified)

- `bal_empty_block_no_coinbase` (canonical zero-tx fixture, succ=1): unmodified vs deleted guest BYTE-IDENTICAL — root fb8018acfc493585b8e99f85c7b75410f8be8c154668c7fbb79fa6fb5289bb7f, stable across four merges.
- Five sensitive fixtures (multi-withdrawal empty blocks, succ_bit=1): 00560, 25117, 25133, 25134, 25135 — ALL identical, including the three currently failing on main (25133-25135, the residual class reported separately); their fail state is preserved identically, the deletion does not touch it.
- Artifact freshness: the deleted guest's emitted .s was grep-verified to contain zero occurrences of the removed branch (`grep -c 'beqz t0, .Lbv_recipient_nc_done'` = 0) before any run.
- .text after deletion: 0x675c8 (readelf -SW measured).

## Two honest notes for the record

1. During this work, on an older base (e9d433893), the two legs of the reroute comparison rejected two multi-withdrawal fixtures with fail-detail bytes 44 vs 60 at output offset 112 — possibly the fail code, i.e. two different rejection mechanisms for the same fixture. Both fixtures (00560, 25117) now PASS on current main (fixed by ecb91e3c9 + #10911), so no action remains; noted here per review instruction.
2. An earlier 54/54 sweep reported as 'byte-identical' was measured on unmodified-vs-unmodified artifacts (the rerouted ELF had never been emitted) and was retracted immediately; all numbers in this PR come from artifacts verified to contain (or lack) the edit they claim to carry.

No broad sweep is offered, per the current loop rules: the change is behaviour-neutral with byte-identical output on the measured zero-tx fixtures, and the MTx path is the one already exercised by every non-empty block.


pirapira:
Reviewed and labelled. This closes the maintainer's zero-tx half of the single-tx/multi-tx unification: *"empty case should be first directed to the mtx block, because execution-specs does not have a special case for zero transaction case."*

**The deletion is exactly minimal, and I verified the part that distinguishes a real deletion from a masked one.** The removed line is only the early exit:

```
-  "  la t0, bv_tx_count; ld t0, 0(t0); beqz t0, .Lbv_recipient_nc_done\n" ++
+  "  la t0, bv_tx_count; ld t0, 0(t0)\n" ++
```

The load is **retained because it is still consumed** — the very next emitted line is `li t1, <bvMtxActiveTxCap>; bgtu t0, t1, .Lbv_mtx_bail`, the active-loop capacity check. So there is no dead load left behind and nothing became unreachable-but-present, which was the specific risk given the instruction to delete rather than orphan. An empty block now falls through the capacity check (`0 > cap` is false) into the MTx loop, which iterates zero times.

**The spec citation is the right one.** `fork.py:913-914` is the entire transaction iteration — `for i, tx in enumerate(map(decode_transaction, transactions)): process_transaction(...)` — with no empty-block special case, and a grep for `single_tx|one_transaction|len(transactions) == 1|ulen(txs) == 1` in the spec returns zero. Putting that citation into the comment rather than only the PR body is right: the next reader of this code is the one who needs it.

**Wiredness and blast radius, both measured rather than argued:**

- `del_guest.s` grep-verified **branch-free** (`beqz` count 0) *before* any run — the artifact-freshness check that made an earlier 54/54 result vacuous when it was skipped.
- Canonical `bal_empty_block_no_coinbase` **byte-identical**, root `fb8018ac…bb7f`, stable across four merges.
- Five sensitive multi-withdrawal fixtures (00560, 25117, 25133, 25134, 25135) all identical — **including the three currently failing on main, whose fail state is preserved identically.** That is the more informative half: a reroute that changed a *failing* block's behaviour would be as suspicious as one that changed a passing block's.

**Both honest notes in the body are the right calls to have made:**

The `44`-vs-`60` offset-112 detail is worth recording even though it does not block: both values are real fail codes in the live distribution (60 accounts for 8,831 of 9,883 FRs, 44 for 93), so the two legs plausibly rejected for *different reasons* on the `e9d433893`-era build. Both of those fixtures now pass on main via `ecb91e3c9` + #10911, so it is history rather than a live defect — but "rejected for the wrong reason" is how a false accept hides, and an unexplained one-byte divergence gets rediscovered as a mystery.

Carrying the retracted 54/54 forward explicitly — it compared unmodified artifacts against themselves — is what makes the numbers in this PR trustworthy. A retraction that travels with the corrected work is worth more than one buried in a thread.

pirapira:
Follow-up on the offset-112 note in this PR's body: it is now a confirmed statement, not a flagged possibility. The live guest's failure output carries its fail code at offset 112 (measured on HEAD 371e2f414: byte 112, LE u32). On fixture 25133 the live stateless_guest rejects with code 31 — agreeing with the v2 probe for this fixture — and the v2-vs-live disagreement caveats documented so far are multi-tx shapes; this agreement is one zero-tx data point, not a general probe-trustworthiness claim. Retroactively, the 44-vs-60 reading at e9d433893 was two different rejection mechanisms (44 = old empty-block path, 60 = MTx tail), as suspected. The note stays as written.

PR #10923, opened 2026-07-30
fix: publish final top-level creation balance

pirapira:
## Spec alignment

`execution-specs` pinned `e5a8caf1b`, `block_access_lists.py:617-650`, builds a transaction BAL from final account writes and emits a balance change only when final post-state differs from the pre-transaction value.

### Root cause

The creation tail had three successive reader experiments that all appeared inert because the publisher never consumed their scratch buffer: it passed `a2 = bv_creation_ctx_ptr + 96` (the original transaction endowment) to `record_nonstorage_effect`. Disassembly exposed the shared downstream blocker. This change removes that operand from the final publisher path: it copies the live depth-zero `evm_env` balance word (`+32`, LE) into the BE `nse_create_post_bal` buffer and passes that buffer as `a2`.

The source contains no remaining `bv_creation_ctx_ptr + 96` use; `bv_creation_ctx_ptr` now only preserves/reloads the context pointer across dispatch. The emitted artifact confirms the wiring: `0x800240ac` reverses all 32 bytes from `evm_env+63..32` into `nse_create_post_bal`; `0x800240fc..0x80024108` materializes `a2=nse_create_post_bal` immediately before `jal record_nonstorage_effect`.

This is a top-level-creation-only stage by construction: it stages `bv_create_addr` as the runtime frame address before dispatch (`BlockVerdictCreationStage.lean:344-355`), and production callers are the single-/multi-transaction creation routes. Thus depth-zero `evm_env+32` is the created account’s final balance; normal top-level CALL transactions do not use this creation stage.

This completes the recurring creation-stage assumption pattern: literal nonce a4=1 (#10913), assumed-live account (#10920), and assumed-unspent endowment (this PR). All are instances of publishing a remembered value after the state it described has changed; this PR reads live final state instead.

## Evidence

- `16190` (initcode CREATE2 spends the sole value): `fail=0`, `shadow=0`, `rebuilt=supplied=390`; previously the guest emitted an extra BAI-1 balance `1`.
- `16226`: `fail=0`, `shadow=0`, `388=388`.
- Retained-value control `00190`: `fail=0`, `shadow=0`, `346=346`; this exercises the same reader when the constructor keeps its value.
- Endianness control `25136`: the created account carries BAI balances `1,000,000,000` and `2,000,000,000`; `fail=0`, `shadow=0`, `354=354`. These multi-byte, distinct values exercise the LE-to-BE conversion and two BAIs.

Candidate ELF sha256: `2a34b9d5c0bb43e96ab35979f46a0e037c87d2bf3b3bd868d695b03b7837a7f4`; runner: `c6ac09fa72c2103b6d3e390b779b022ef7f220e5a39bb5f667114a89d78e5868`.

## Generated layout and checks

Regenerated to fixpoint: `.text=0x675f8`, `.data=0x5370`, `GuestAddrs.bnf_le_a=0xa3000ee0`. `scripts/check-asm-to-program.sh`, `scripts/check-region-map.sh`, and `git diff --check` pass. `scripts/check-build-parallel.sh` reached the codegen checks but its `lake env lean --run` helper cannot run in artifact-cache mode (the documented cache limitation); CI runs that gate in non-cache mode.


pirapira:
Reviewed and labelled. This is the strongest evidence set of anything labelled tonight, and two of my three gates were closed in the artifact rather than by argument.

**`ctx+96` is deleted outright, not demoted — verified in the diff.** The entire change at the operand is:

```
-  "  la a2, bv_creation_ctx_ptr; ld a2, 0(a2); addi a2, a2, 96\n" ++
+  "  la a2, nse_create_post_bal\n" ++
```

One line out, one in, **no conditional fallback anywhere**. That is better than the use-when-present design discussed earlier: an assumed value retained behind a condition is where this class of bug returns, and here there is no condition to return through. Earlier in the investigation I defended keeping the old value as a fallback on the grounds it was the well-exercised path; that argument dissolved once a single reader covered retain, spend and partial-spend uniformly.

**Layout pin repinned:** `textSizeBytes 0x0675d0 → 0x0675f8`, with `check-region-map` and `check-asm-to-program` green and regeneration run to fixpoint. This is the gate that failed on #10921 an hour ago — where a deletion *shrank* `.text` and the pin was left stale — and I missed it there while reviewing. I now check pins on every codegen PR, and this one is clean.

**The root cause is the reviewable insight and the body leads with it correctly.** Three successive readers — nonstorage overlay, `account_state` overlay, live-env copy — all produced byte-identical `394/390`, and each was written up as falsifying its mechanism. **None of them tested anything:** the publisher still passed `a2 = bv_creation_ctx_ptr+96` while the readers filled a scratch nobody consumed. The lesson worth carrying: **N identical nulls from N genuinely different mechanisms indicts the plumbing, not the mechanisms** — and a falsification is only valid if the thing being falsified actually ran. It was found by **disassembling**, after three source-level changes each looked correct.

**Both hazards I raised are discharged at the emitted-asm level:**

- **Endianness.** 25136 has the created account at `BAI1 = 1,000,000,000` and `BAI2 = 2,000,000,000` — four-byte values with distinct nonzero bytes — and is byte-perfect at `354 = 354`. A partial reverse, wrong width, or low-limb-only copy would corrupt those. The earlier fixtures (values 0 and 1) could never have distinguished a correct 32-byte reverse from a broken one, since both are single-significant-byte in either order.
- **The operand actually passed.** `0x800240ac` reverses exactly 32 bytes from `evm_env+63..32` into `nse_create_post_bal`; `0x800240fc`–`0x80024108` materialises `a2 = nse_create_post_bal` immediately before `jal record_nonstorage_effect`. The dead-scratch failure mode is excluded in the artifact.

**Evidence set:** 16190 fixed (`fail=0`, rebuilt = supplied = 390, spurious `BAI1` balance gone); 16226 held at 388; **00190 held at 346** — the load-bearing control, since it exercises the *retain* path through the new reader; 25136 held at 354.

## Residual, recorded not blocking

The body does not state whether the depth-zero `evm_env` account is the created account **by construction** for a top-level creation, or what `env+32` holds if this stage is ever reachable on a top-level CALL. The routine's name implies creation-only, but a label name is not a contract. Given the unconditional read and four fixtures across retain/spend/multi-byte, this is a documentation gap rather than a correctness risk — worth one sentence whenever this code is next touched.

## Pattern, now four instances in one routine

Literal nonce `a4=1` (#10913), assumed-alive account (#10920), assumed-unspent balance (here) — all the same habit: **publishing a remembered value after the state it described is gone.** This is the first fix that reads live state instead of reconstructing it, which is why it needs no special case per constructor behaviour. Spec anchor: `block_access_lists.py:617-650` emits only field values from the final `tx_state.account_writes`.

pirapira:
**The residual noted above is now closed by a body update** — recording it so my review does not read as stale.

The construction argument is stated: **the stage rewrites the runtime address from `bv_create_addr`, and its production callers are creation routes**, so a normal top-level CALL does not reach it. That answers "is the depth-zero env account the created account by construction, rather than by fixture."

Also confirmed with a useful precision I had not asked for: **`ctx+96` is absent from the final publisher, and the context pointer remains only for save/reload.** That distinction matters — a grep for `bv_creation_ctx_ptr` will still hit this file, and a future reader could otherwise conclude the old path survived. The operand is what was removed, and the pointer's remaining uses are unrelated to the published value.

All three gates are now closed in the artifact rather than by argument: multi-byte endianness (25136, two BAIs), the `a2` operand at the call site (both PCs in disassembly), and the by-construction site argument. Label stands.

Issue #10919, opened 2026-07-30
BAL/EIP-7708: the spec suppresses only the Transfer LOG on a self-transfer, not move_ether's bookkeeping — the guest may suppress both


pirapira:
## The asymmetry

For a nonzero transfer where sender equals recipient, the spec suppresses **exactly one** of the two things it would otherwise do. The guest appears to suppress **both**.

`vm/interpreter.py:384-396`:

```python
if message.should_transfer_value and message.value != 0:
    move_ether(
        tx_state,
        message.caller,
        message.current_target,
        message.value,
    )
    if message.caller != message.current_target:
        emit_transfer_log(
            evm,
            message.caller,
            message.current_target,
            message.value,
        )
```

| path | spec on a self-transfer |
|---|---|
| `emit_transfer_log` | **suppressed** — guarded by `caller != current_target` at `:391` |
| `move_ether` | **NOT suppressed** — runs on any nonzero value |

So the spec calls `move_ether(sender, sender, value)`. That is a balance no-op *arithmetically* — `state_tracker.py:646` applies two `modify_state` calls sequentially and `:631-641` reads current state via `get_account`, so subtract-then-add on one account composes to zero — but it is **not** a bookkeeping no-op: `set_account` fires twice on that address, and the account therefore appears in `account_writes`.

**Confirmed against ground truth:** fixture `00504_test_bal_self_transfer` (one transaction, `to` == sender == `0x67976fe3df4a275ba3a3239a9f3f6af71c1eddf1`, value `0x64` = 100) declares a receipt with **`logs: 0`** — the spec emits no Transfer log — *and* its declared BAL **does** carry both a `balanceChange` and a `nonceChange` for that address.

## Why this is filed as a divergence rather than a confirmed defect

The guest correctly emits **no** Transfer log for this case (verified: forcing one to be emitted moved `eip7708_append_transfer_log` arrivals 0→1 and made the guest *diverge*, so its skip is right).

[...]

PR #10925, opened 2026-07-30
fix: stage sender fields for MTx EOA runtime payloads

pirapira:
## Spec alignment

Amsterdam `transactions.py:653-669` charges recipient access and value only when the sender differs from `tx.to`; `interpreter.py:391` also suppresses the transfer log for `caller == current_target`. The sender is already warm, so a self-transfer receives neither the cold recipient access charge nor the value/log charge.

The shared `stage_runtime_payload_code` intentionally leaves `CALLER`, `ORIGIN`, and `GASPRICE` unfilled (`BlockVerdictContractStage.lean:53-55`), but the runtime self-transfer guard reads `ADDRESS` at `env+0` and `CALLER` at `env+64` (`Dispatch.lean:2739-2748`). MTx EOA invoked that stager without populating either sender field. The sibling contract route already stages both fields after the same stager (`BlockVerdictDispatchTx.lean:835-860`); this change makes MTx EOA consistent by copying its existing `bv_mtx_sender_addr` into `CALLER` and `ORIGIN`.

`CALLER` fixes the defect. `ORIGIN` is staged for the established sibling-route convention: it is not necessary for the 00504 intrinsic-gas case.

## Exact fixture evidence

For `00504_test_bal_self_transfer…`, the old not-self charge was `21000` against the declared `12000` limit. The difference is exact: `9000 = 3000` cold-account-access `+ 6000` value/log charge.

On the candidate ELF (`sha256 d6f4e705b3668ca23dfc82c5988fdf5b6df4624c9e29cc2aff1f136f8cb320fd`), the resolved self comparator at `0x8003a2cc` compares all 20 address bytes and falls through at counter 20. It stores the 12000 intrinsic charge, so the fixture passes its former OOG/receipt-root point and reaches a separate BAL digest residual instead: `fail_code=60`, `rebuilt_len=supplied_len=297`.

The MTx EOA route stages a fixed STOP body; it does not execute user bytecode. Its live `CALLER` reader is the intrinsic self-comparison. The general `h_CALLER`/`h_ORIGIN` handlers are unreachable, and no EOA-route zero-sentinel consumer exists.

## Controls

- Valid controls remain byte-perfect: `00379` is `fail=0, shadow=0, 383=383`; `00190` is `fail=0, shadow=0, 346=346`.
- Expected-invalid FA-side control `04778` still rejects (`fail=5`, `shadow=3`).

## Regeneration and checks

Regenerated to a fixpoint: `.text=0x67650`, `.data=0x5370`; `GuestAddrs.bnf_le_a` remains `0xa3000ee0`. Passed `scripts/check-asm-to-program.sh`, `scripts/check-region-map.sh`, and `git diff --check`. Targeted `lake build EvmAsm.Codegen.Programs.BlockVerdictMtxEoa` passed.

The earlier #10919 bookkeeping-asymmetry lead is superseded by the documented producer-consumer staging mismatch; the issue comment records the trace and the remaining independent BAL residual.


pirapira:
Reviewed and labelled. Every gate is closed in the artifact rather than by argument, and one of them was closed better than the check I asked for.

**The root cause is a producer/consumer contract mismatch, which is the reviewable insight.** `stage_runtime_payload_code`'s own comment (`BlockVerdictContractStage.lean:53-55`) documents CALLER/ORIGIN/GASPRICE as **left zero**, while `Dispatch.lean:2739-2748` compares env ADDRESS at `+0` against CALLER at `+64` regardless. So the self-transfer comparison read a field its producer never promised to fill, and **could not succeed for any input on that route** — the guard was effectively dead, not merely wrong on one fixture.

**The fix follows an existing convention rather than inventing one.** `BlockVerdictDispatchTx:835-860` already stages CALLER+ORIGIN after the same shared stager; MTx EOA was the outlier. Copying `bv_mtx_sender_addr` makes two callers of one stager consistent, and the shared stager stays unchanged.

**The arithmetic closes exactly**, which is the most convincing line available: the not-self arm charges 3000 cold access + 6000 value/log = **9000**, and 21000 − 12000 = **9000**, with nothing left over. Both charges are independently wrong per spec — the sender is already warm, so a transfer to itself cannot be a cold account access; and `interpreter.py:391` guards `emit_transfer_log` with `caller != current_target`, so there is no log to charge for. That is the same spec line that retired the earlier missing-log hypothesis on this fixture; it turns out to explain the *gas* as well as the absent log.

**Mechanism-level evidence, not a pass:** the comparator at `0x8003a2cc` matches **all 20 bytes** and falls through at `counter=20`, with intrinsic stored as 12000. That pair matters together — either alone could be coincidence.

**And it incidentally validates the byte order.** The staging loops reverse-copy (`li t4, 19; sub t4, t4, t3`), reading `sender + (19 − i)` into `dest + i`. A wrong-direction copy would leave the comparator mismatching, so "all 20 bytes match" *is* the endianness check. Worth noting because the same conversion in #10923 needed a deliberately-chosen multi-byte fixture to verify — here the comparator supplies it for free.

## The sentinel audit came back stronger than the grep I asked for

I asked for an enumeration of readers of `env+64`/`+128` on this route, because the change moves two fields from always-zero to a real address on the most-travelled path in the corpus, and a consumer treating zero as a sentinel would silently take a different arm. I attempted the grep myself and it was too noisy to trust — the env base is a *register* at each site (`x20` in the comparator), so the offsets are not greppable by literal.

The answer given is better than an enumeration: **the MTx EOA route installs `bv_stop_code` and invokes the dispatcher, so no user bytecode executes.** `h_CALLER`/`h_ORIGIN` are general bytecode handlers and are therefore **unreachable** under STOP. The only live `+64` reader is the self-gas comparator, and no route-local zero-sentinel or default branch exists. That is an argument from what *can* run rather than from what a grep found.

Consequently: **`+64` is defect-required; `+128` is solely to match `DispatchTx:835-860`.** Recording that distinction is the right call — a consistency change that quietly altered behaviour would be worse than a documented inconsistency.

**Layout pins repinned:** `textSizeBytes 0x0675d0 → 0x067650`, regenerated to fixpoint. I verify this on every codegen PR now, after missing it on #10921.

## Credit where the elimination work belongs

Scoper's three retracted hypotheses on this fixture — the missing Transfer log, the BAL bookkeeping asymmetry, and the write-back — are why this was a bounded diagnostic rather than a re-walk. Five source-only readings failed here before a disassembly plus two targeted source reads found it; that instrument ordering is part of the finding.

PR #10928, opened 2026-07-30
fix(codegen): separate creation initcode from calldata

pirapira:
## Spec alignment

Amsterdam execution-specs treats top-level CREATE initcode as `code` and supplies empty `data` to the creation message (`execution-specs/src/ethereum/forks/amsterdam/vm/instructions/system.py:134-143`). The guest had staged `ctx.data` as both runtime code and frame calldata, so creation initcode was incorrectly visible to `CALLDATALOAD` and `CALLDATACOPY`.

This keeps initcode as executable runtime code, stages empty frame calldata, and passes the original transaction-data span separately to the intrinsic-gas calculation. The override is creation-local and cleared on all returns.

## Wired evidence

The 13-byte `16398` initcode is `60003560005560213560015500`: it performs `CALLDATALOAD(0)` and `CALLDATALOAD(0x21)` then stores both results. With empty calldata both observed stored words are zero. The final ELF observed both zero. Its state total is `183600`, exactly the fixture header `gasUsed` (`0x02cd30`), and receipt cumulative gas is `212792`, exactly declared (`0x033f38`).

Sibling `16397` and `16399` reach the remaining BAL-digest layer (`bv_fail=60`), as does `16398`; no fixture is claimed passing yet. Expected-invalid control `04778` remains rejected (`bv_fail=5`).

## Layout and checks

- Final ELF: `5605a05a93d597e9040230ddba811024e840eb45739d428a1ad95cab6ce32355`
- `.text=0x6769c`, `.data=0x5370`, `.bss=0x1c1395c0`; `GuestImage` hand-maintained `.bss` bounds repinned to `0xbf2495c0`; `bnf_le_a` remains `0xa3000ee0`.
- Passed: `lake build EvmAsm.Codegen.Proofs.GuestImage`; `scripts/codegen-stateless-link-check.sh --no-build`; `scripts/check-asm-to-program.sh`; `scripts/check-region-map.sh`; `scripts/check-memorylayout-region-coverage.sh`; `scripts/check-build-units-link.sh`; `git diff --check`.
- `scripts/check-build-parallel.sh` reached its codegen subgate but `check_guarded_handler_bytes.py` cannot run under this checkout's artifact-cache mode because `lake env lean` cannot materialize `GuardedHandlerSpecs.olean`. CI runs that complementary gate outside cache mode.


pirapira:
Reviewed and labelled. **No fixture passes yet and the body says so** — this is a layer-peel, and it is being merged as spec alignment rather than as a fix.

## Why it merges with zero fixtures green

The spec is unambiguous: for a creation, `tx.data` becomes the message's **code** while the message's **data** is empty. The guest was feeding the initcode into EVM calldata, so `CALLDATALOAD` returned real bytes. **Wiredness is demonstrated, not argued:** 16398 and 16397 moved from `fail_code=62` (exact-gas) to `60` (BAL digest), and **nothing regressed** — 16399 was already at `60` in the pre-#10911 queue, so two fixtures changed failure mode and the third stayed put.

## Both gas components reconcile against declared fixture data

Not against either build — against the fixture:

| quantity | declared | candidate |
|---|---|---|
| header `gasUsed` | `0x02cd30` = 183,600 | state total **183,600** |
| receipt `cumulativeGasUsed` | `0x033f38` = 212,792 | receipt **212,792** |

The parent computed 281,520 / 320,712. Note this family's header `gasUsed` and receipt `cumulativeGasUsed` **differ** for a single-tx block, so they are two independent declared quantities rather than a check on each other.

**Identifying the case required content, not path.** The fixture file holds **three** cases sharing one relpath, and the manifest label truncates. My first read grabbed the wrong case and produced 212,997 — a 205-gas phantom. The right case is identified by its tx data `0x60003560005560213560015500`.

## The mechanism has a prediction the data confirms

The 13-byte initcode contains **two** SSTOREs, yet the overcharge was exactly **one** `STORAGE_SET` (97,920). Slot 0 takes `calldataload(0x00)`, which lands **inside** the 13 bytes and yielded nonzero under the bug; slot 1 takes `calldataload(0x21)` — offset 33 — which is **beyond** them and yielded zero even with malformed calldata. One nonzero write, one spurious set, matching the measured delta to the wei. That is stronger than a total agreeing.

Evidence is symmetric: initcode bytes were read off the stack to prove the defect, and the stored values were read as **zero** to prove the fix.

## Pins — verified by value, and this is the part no gate covers

I flagged before the push that `.bss` moving `0x1c1395a0 → 0x1c1395c0` requires repinning `GuestImage.lean`'s **hand-written** literals per its own comment at `:203-207`. All three are correct here:

- size literal `0x1c1395a0 → 0x1c1395c0`
- `t7` end bound `0xbf2495a0 → 0xbf2495c0`
- `t7'` end bound `0xbf2495a0 → 0xbf2495c0`, with its base `0xa3005370` correctly untouched (it derives from `.data`, unchanged at `0x5370`)

Plus `RegionMap.textSizeBytes 0x067630 → 0x06769c` and `bssSizeBytes` updated.

**Independent corroboration:** the symbol TSV's `__BSS_END__` and `_end` both move to `0xbf2495c0` — so the linker's own symbols agree with the hand-written proof literals, rather than the arithmetic merely agreeing with itself.

Worth stating why this needed catching by hand: those bounds are discharged by `omega`, so a **stale literal can still typecheck** if the bound remains satisfiable — wrong without failing, and it would merge green. `check-region-map` validates RegionMap against the ELF; nothing validates these.

## Structural note

`stage_runtime_payload_code` is shared between the call and creation routes and its contract is underspecified — each caller must supply what the stager does not. **That seam produced two defects tonight:** #10925 (CALLER/ORIGIN documented as left zero, never filled by MTx EOA, read anyway) and this one (creation semantics the stager cannot distinguish from call semantics). A shared routine whose callers must each remember an unwritten obligation is a defect generator.

## Residual, stated not hidden

All three siblings now fail at `bv_fail=60` on the BAL layer. 16399 is the useful next control: its declared header is **197,370** where the other two are 183,600, which is what you would expect if code remains populated while data goes empty. 04778 still rejects at `fail_code=5` — an FA leg on a population that can actually exhibit one.

PR #10929, opened 2026-07-30
fix(codegen): preserve nonce effects in BAL builder

pirapira:
## Spec alignment

`execution-specs` reduces nonce changes by maximum post-nonce (`block_access_lists.py:440-447`), while balance remains last-write (`:403-419`).

The guest had two distinct structures:

- `nonstorage_effect_aggregate` is the comparison aggregate. This patch makes its nonce reduction maximum; that correction is spec-aligned but inert for emitted BAL bytes on its own.
- `tx_account_writes`, populated through `account_write_record`, is the BAL builder feed. `record_nonstorage_effect` passed `a6 = 3` for every record, claiming both balance and nonce even for a balance-only record. Its keyed upsert could therefore overwrite a real nonce increment with a trailing zero.

The fix makes the builder-feed mask fieldwise and retains maximum nonce reduction. The trailing record is not suppressed: it may carry a real balance post.

## Evidence

Measured stacked with #10928 (required for 16397/16398 to reach BAL rather than earlier code 62):

- 16397: fail 0, rebuilt/supplied 304/304
- 16398: fail 0, rebuilt/supplied 304/304
- 16399: fail 0, rebuilt/supplied 314/314
- expected-invalid 04778 remains rejected (fail 5).

The common prediction held: all three creation controls moved, confirming the unconditional builder-mask mechanism is shared rather than creation-specific. #10928 supplies the creation calldata split; on bare main 16397/16398 still halt at code 62 before this BAL path.

Layout values are from the combined stacked ELF: `.text=0x676ec`, `.data=0x5370`, `.bss=0x1c1395c0`; `GuestAddrs.bnf_le_a` remains `0xa3000ee0`. The merge batch must regenerate the generated address artifacts from the combined source.


pirapira:
Reviewed and labelled. **The load-bearing content is the source change to `NonstorageEffectLog.lean`; the generated files here are the bot's to settle.** Read the CI accordingly — see the note at the bottom before concluding anything from a red check.

## The defect, in one operand

`record_nonstorage_effect` called `account_write_record` with **`a6 = 3` unconditionally** — every record asserting it carries *both* a balance and a nonce. So a trailing **balance-only** record overwrote a real `0 → 1` nonce with `0`. That is not creation-specific: it fires for any address whose last record carries no genuine nonce effect, which is most addresses in most blocks.

## Two aggregation structures, and the BAL reads the one that wasn't fixed

This is the reusable finding and a reviewer will otherwise assume a single aggregator:

- **`tx_account_writes` via `account_write_record` — the BAL builder feed.** This is where the defect lives.
- **`nonstorage_effect_aggregate` — the comparison aggregate.** A max-nonce reduction was added here and is **spec-correct per `block_access_lists.py:440-447` yet INERT for the emitted bytes.** Crediting it with the fix would mislead the next reader.

That distinction cost several measurement rounds: the reduction was confirmed live at its own helper (`pre_nonce=0`, `post_nonce=1`, later records taking the keep path) while the rebuilt BAL still lacked the row.

## The prediction was stated before the measurement and it held

Because `a6=3` is unconditional and shared, **all three** creation controls should move, and **only 16398 moving** would have meant the mask was not the whole story. Result: 16397 `304/304`, 16398 `304/304`, 16399 `314/314`, all `fail=0 shadow=0`. Expected-invalid **04778 still rejects at `fail5`** — an FA leg on a population that can actually exhibit one.

A prediction stated first and confirmed is materially stronger than three green fixtures found afterwards, because it tests the *mechanism* rather than the outcome.

## Dependency, declared and load-bearing

**16397 and 16398 require #10928** (the creation initcode-versus-calldata split) to reach this BAL path at all; on bare main they halt earlier at code 62. **16399 moves on this change alone.** The body states this. Anyone re-measuring on bare main will see two-thirds of a null and should not read that as the mask fix failing.

Both PRs are labelled, and the bot aggregates all `merge!` PRs into one branch, so the semantic dependency is satisfied *within* the aggregate. No sequencing needed.

## Why the generated state here is deliberately incomplete

`GuestAddrs` and `symbol-addresses.tsv` are **intentionally left** for the bot's combined-source regeneration. The `RegionMap`/`GuestImage` pins are set to the **combined-image** values (`.text 0x676ec`, `.data 0x5370`, `.bss 0x1c1395c0`), including the three `GuestImage` literals — size `0x1c1395c0` and both end bounds `0xbf2495c0` — which are the values `.bss = 0x1c1395c0` requires.

**So this PR is internally inconsistent by construction and its individual `check-region-map` is expected to be red.** The branch also carries a non-main pin baseline (`textSizeBytes` differencing from `0x0676d8`, where main is `0x067630`), and it shows `CONFLICTING`. All of that is the merge bot's routine work — conflict resolution plus one coherent regeneration across the aggregate — and per the maintainer's standing rule on generated-artifact drift, **I am not unlabelling for it and nobody should pull the author back to regen.** I broke that rule earlier tonight on #10921 and it moved a head out from under an open batch.

The pins set here will be overwritten by the bot's authoritative regen; setting them is a best-effort guess at the post-batch state, not a claim about this branch's ELF.

Issue #10838, opened 2026-07-28
BAL residual: re-scope around measured signatures (storage_changes-8 on precompile blocks, 7702 nonce+code, over-counts)


pirapira:
Leftover from #10680, which is now closed: the digest comparator is wired into the block verdict (#10835), so BAL producer defects are visible in ordinary runs. #10836 covers the largest root (the tracked/untracked accessor split, accounting for the 45 fee-recipient and ~26 execution-time entries). This issue tracks the residual entries that root does **not** explain.

Residual measured on the 64 blocks / 80 empty entries remaining after #10834, every entry classified:

| n | class | owner |
|---|---|---|
| 45 | block fee recipient | #10836 |
| 15 | internal CALL target (address a literal in witness code or calldata) | #10836 |
| 9 | precompile | #10836 |
| **7** | **zero-value `tx.to`** | **this issue** |
| **2** | **EIP-7702 authority** | **this issue** |
| **2** | **unexplained** | **this issue** |

Two entries are both precompile and zero-value `tx.to`, so the classes overlap and cannot be summed.

## 1. Zero-value `tx.to` (7 entries)

A transaction recipient that receives no value. The spec touches `tx.to` regardless — the recipient is warmed unconditionally at `fork.py:1110-1116` and read during transfer accounting — but with zero value the write is net-equal, the change row is filtered, and only the **read** should survive as a touched-only entry. Same shape as the fee-recipient defect: the mechanism that would record it is the accessor, and a zero-value path may reach an untracked one. Check whether this is simply another instance of #10836 before treating it as separate.

## 2. EIP-7702 authority (2 entries)

`eip7702_authority_asof` and `eip7702_auth_nonstorage_effects` both appear in the untracked `account_at_header_state_root` site list (#10836), while `account_state_auth_current` **is** one of the four `account_read_record` feeds. So the authority read is recorded on one path and not another — worth confirming which path the 2 entries take. Note `Authorization.chain_id` is U256 (`fork_types.py:92`) while transaction `chain_id` is U64; unrelated to this defect but adjacent when reading these paths.

## 3. Unexplained (2 entries)

```
000000000000000000000000000000000000c0de  00072_test_buffer_src_offset
4422a35783982f5d6184c7e77241b05e95a62fc8  00192_test_shallow_stack
```

[...]

PR #10930, opened 2026-07-30
fix(bal): record top-level creation target access

pirapira:
## Summary

Top-level creation computes `current_target = compute_contract_address(...)` and unconditionally adds it to `accessed_addresses` in `execution-specs` Amsterdam `utils/message.py:56-71`, before execution. The guest derived the same address in `bv_create_addr` but never recorded the access.

This adds the existing `account_read_record` immediately after the derived address is copied and before `runtime_dispatcher_call`. It is intentionally unconditional: an all-empty created account has no change row to carry the touch, which is why the omission was observable in shallow-stack fixtures.

The emitted-assembly liveness audit confirms `runtime_dispatcher_call` does not consume incoming `a0`/`x10`: every path sets `x10` from `runtime_dispatcher_input_ptr` before dispatch. `account_read_record` saves/restores `t0`-`t6`; `s0`-`s3` remain live and `a1`-`a6` are redefined before use.

## Evidence

- Prediction: all 81 `test_shallow_stack` rows share the missing pure touch and should move. Result: **81/81 pass** on the final ELF.
- `20250`: `fail=0`, `shadow=0`, rebuilt/supplied BAL lengths `326/326`.
- FA-side control `04778` remains rejected: `fail=5`, `shadow=3`.
- `check-region-map.sh`: clean; `.text=0x67750`, `.data=0x5370`, `.bss=0x1c1395c0`; symbol table converged across successive regenerations.
- `check-asm-to-program.sh`: clean (390 converted definitions).
- Cache-disabled build output parity: 1213 `.olean` and 1213 `.olean.hash`.

`4422a35783982f5d6184c7e77241b05e95a62fc8`, the sole missing entry in 20250, was independently verified as `keccak(RLP([sender, 0]))[-20:]`, i.e. the top-level `CREATE(sender, 0)` target.


pirapira:
Reviewed and labelled. **Eighty-one fixtures from three instructions, with the family prediction registered before the measurement.**

## The defect and its spec citation

`utils/message.py:56-71`: `:56` computes `current_target = compute_contract_address(...)` for a creation where `:64` takes `tx.to` for a call, and **`:71` does `accessed_addresses.add(current_target)` unconditionally for either kind — before any execution begins.** The guest derived `bv_create_addr` and never handed it to a read recorder, so the top-level creation target was never touched.

**The citation matters here:** I initially pointed at `system.py:106` (`accessed_addresses.add(contract_address)`), which is the analogous line in `generic_create` for a **nested** CREATE and does not govern a top-level creation. `:56-71` is the correct provenance, and it is what justifies the record being **unconditional** — not gated on creation success, on the initcode running, or on the account ending with changes. A success-gated version would be unreachable on precisely the failing-initcode fixtures that exposed this.

## Why the defect survived this long — the reconciliation

If the guest never recorded this touch, *every* successful CREATE would be short a 27-byte entry, and many CREATE fixtures pass. The resolution: **a created account with any change appears in the BAL via that change, so the missing touch is invisible unless the account ends all-empty.** Only then must the pure-touch entry carry it alone. `shallow_stack`'s initcode leaves nothing to declare, which is why this family exposed it.

## The prediction was pre-registered and it held

Stated before measuring: 81 rows in a 31-byte size band with one fail code is a **template**, so one shared mechanism should move all 81 — and only a handful moving would mean the mechanism fixed was not the shared one. Result: **20250 `fail0/shadow0` at `326=326`, and all 81 `shallow_stack` inputs pass.** Expected-invalid **04778 still rejects at `fail5/shadow3`**.

## Cross-confirmations, none of which were arranged

- **#10838 named this exact address.** Line 28 of its body reads `4422a35783982f5d6184c7e77241b05e95a62fc8` — found there by scanning 451 witness code blobs and finding the address as a literal in **none**, inferring a runtime-computed CREATE address. Independently found here by decoding the rebuilt BAL.
- **The address was derived, not assumed:** `keccak(RLP([8e225830…, 0]))` tail matches exactly, confirming `CREATE(sender, 0)`.
- **The `.text` delta corroborates the diff's scope.** `0x067744 → 0x067750` is exactly **12 bytes = 3 instructions**, precisely `la a0, bv_create_addr` (auipc+addi) plus `jal ra, account_read_record`. One call inserted and nothing else changed, confirmed by the layout rather than by reading the diff.
- **#10838's other hypothesis is refuted, not inherited:** it suggested these might fall out of #10836's fix; #10836 closed via #10896 and the family still had 81 failing rows on that base.

## The register hazard, closed in the strongest available form

Inserting a call here writes `a0`, and `a0` aliases `x10` — the dispatch PC on some paths — so the failure mode would be **execution redirection**, not a wrong byte, and no fixture in this gate would catch it. Resolved from **emitted asm**: `runtime_dispatcher_call` takes **no `a0`/`x10` input**. Its prologue loads `x6` from `runtime_dispatcher_input_ptr` and defines `x10` itself (default `0x40000010`), while the creation route stores `bv_runtime_payload+8` into that pointer immediately before the `jal`. **The dispatch PC comes from memory, so a caller-side clobber cannot reach it** — which makes the branching-arm question irrelevant rather than merely unreached. Callee saves/restores `t0`–`t6`; `s0`–`s3` untouched; `a1`–`a6` freshly assigned before next use.

Method worth keeping: **a caller-side liveness question was settled by reading the callee's prologue.** That is the mirror of the recorded trap that *callee-saved does not mean the register still holds the value inside the callee* — in both directions the answer is in the prologue, not the calling convention.

## Pins

`textSizeBytes 0x067744 → 0x067750`. **`.bss` unchanged at `0x1c1395c0`**, and `GuestImage.lean` already carries both end bounds `0xbf2495c0` at lines 220/224 from #10928 — so its absence from this diff is correct, not an omission. That was the specific check I have been running on every codegen PR since #10921 shipped a stale `RegionMap` pin.

Issue #10926, opened 2026-07-30
investigate: nontrivial top-level CREATE bails after common runtime staging


pirapira:
## Corrected status

This issue originally attributed `16398` to the one-byte STOP gate in `stage_creation_runtime_payload`. That premise is **false for the live verdict route**. The one-byte stager exists, but `block_verdict_creation_runtime` bypasses it.

The live routine (`BlockVerdictCreationStage.lean:294-431`) already:

1. stages transaction initcode with the common `stage_runtime_payload_code` (`:306-319`),
2. populates the authenticated M31 context and creation frame environment, and
3. invokes `runtime_dispatcher_call` (`:430-431`).

Therefore arbitrary initcode execution capability is present. The actual question is why this route returns nonzero for the fixture; the specific bail must be traced before proposing a code change.

## Failing fixture

`16398_test_create_transaction_call_data_fork_Amsterdam-blockchain_test_from_state_test-calldataload__b` is a valid one-transaction CREATE with 18-byte `CALLDATALOAD`/`SSTORE`/`STOP` initcode. On the current candidate it returns `fail_code=62`, `rebuilt_len=339`, `supplied_len=304`. Siblings 16397 (CALLDATACOPY) and 16399 (CODECOPY) are controls for the same class.

## Spec target

Amsterdam `vm/interpreter.py:126-145` routes empty-target transactions to `process_create_message`; `process_create_message` (`:176-229`) executes its message/initcode.

## Next measurement

Trace `block_verdict_creation_runtime` / `stage_runtime_payload_code` / `runtime_dispatcher_call` on 16398 to find the first nonzero status. Do not remove any bail beforehand: the live code already has the intended general execution seam, so the remaining fault is routing, staging, gas, or post-execution publication—not lack of an interpreter.

pirapira:
Correction recorded: the live source disproves the prior one-byte-gate attribution. The gate belongs to an unused/standalone stager; the production creation runtime already stages arbitrary code and calls the dispatcher. Treat the prior wording as superseded.

pirapira:
[...]

PR #10931, opened 2026-07-30
fix(bal): preserve MTx EOA sender state across callable dispatch reset

pirapira:
## Spec alignment

`execution-specs` applies `move_ether` as two sequential `modify_state` updates (`state_tracker.py:631-646`), each reading its predecessor through `get_account`. The MTx EOA shortcut published the gas/value upfront state into the pending AccountState journal immediately before entering the callable dispatcher; `Dispatch.lean:3030` resets that journal. The debit therefore disappeared before execution, and a self-transfer calculated the transfer debit from stale state.

This combined change keeps the intentional reset and publishes the pre-execution sender state to the durable overlay instead. The recipient credit and later sender debit read that transaction-current state, so self-transfer composes credit then debit just as the spec does. The upfront charge is durable because it survives a top-level revert.

## Combined work

This PR deliberately includes scoper’s prerequisite commit `4efb959`: it removed the credit-path guard after the one-byte localisation and valid three-way 297-byte capture showed the guest took the debit while skipping the credit. My contribution is the store-watch / three-base analysis that ruled out a local refund base, identified the `Dispatch.lean:3030` callable-reset root (one source site, six dynamic entries), and added the durable publication plus current-state reads. Neither half closes 00504 alone.

The `EIP7708Logs` seed-helper comment now distinguishes full dispatcher setup from the MTx EOA shortcut; the latter crosses the callable reset and must not use the pending seed.

The ELF `.bss` is `0x1c1395a0`; `GuestImage.lean` already carries the matching size and both `0xbf2495a0` end bounds inherited from the stacked base.

## Merge note

`GuestImage.lean`, `RegionMap.textSizeBytes`, and `RegionMap.bssSizeBytes` must be regenerated to a fixed point when this combines with main. Main and this stacked branch carry pins correct for different layouts; choosing either side of the generated conflict can leave a silently stale upper bound.

## Evidence

- `00504_test_bal_self_transfer...`: exact 69-byte expected-output prefix match; sender final balance is exactly `880000`.
- `00225_test_reverted_transaction_no_log...out_of_gas`: accepts, `fail=0`, confirming the durable gas debit survives the revert path.
- `04778_test_tx_intrinsic_gas...below_intrinsic`: remains rejected, `fail=5`.
- `LAKE_ARTIFACT_CACHE=false lake build`
- `LAKE_ARTIFACT_CACHE=false scripts/check-asm-to-program.sh`
- `LAKE_ARTIFACT_CACHE=false scripts/check-region-map.sh`
- olean/hash counts: 1213/1213 before the build; relinked ELF `.text=0x67860`, `.data=0x5370`, `.bss=0x1c1395a0`.


pirapira:
Reviewed and labelled. **Two composed changes that only work together, with 00504 reaching a 69-byte exact output match** — not merely `succ=1`.

## Collaboration, stated because our git identity is shared and the UI cannot show it

**Scoper:** the one-byte localisation (sender `postBalance` exactly 100 low against declared `0x0d6d80`), the valid three-way length capture at 297 — which required giving the fixture a **unique index** to kill a collision that had invalidated an earlier decode — the diagnosis that the guest was *taking the debit while skipping the credit*, and the credit-path guard removal itself (commit `4efb959`).

**Codex:** the store-watch that **refuted** the refund-record theory I was pushing (refund reads at 2780107, records at 2781394 with pre and post both 879,900 — it faithfully preserves what it is handed), the three-base arithmetic proving no local fix could work, the #10876 root cause, and the durable-publish repair.

## Why the fix is durable publication rather than resequencing — a soundness reason

The upfront gas debit **must survive a revert**: the spec charges gas whether or not execution succeeds. Moving the seed to *after* the callable reset would place it in the pending journal, which is committed on success and **dropped or rewound on failure** — wrongly refunding gas on a reverted transaction. That is a false-accept direction, not a wrong number, which is why the alternative was rejected on principle rather than by measurement.

The chosen route follows an existing precedent rather than inventing one: `account_state_publish_sender_inclusion` already writes durable directly and is unaffected by the reset, and this adds `account_state_publish_sender_upfront` alongside it. Same shape as the two fixes that landed earlier tonight — the sibling route already does the right thing and this one did not.

**00225 (reverted no-log OOG) at `succ=1 fail=0` is the control that confirms the risk did not materialise.** That fixture is exactly the class #10917 is held for re-breaking, so it is the lineage gate doing its job.

## Root cause: #10876, scoped before it was touched

`runtime_dispatcher_call`'s callable setup unconditionally zeroes `account_state_pending_count` at `Dispatch.lean:3030`. Scoping established **one site entered six times**, not six sites — #10876's `0x80039cd4` and the `0x80039f34` seen here are the same source site across two layouts — with a genuinely separate reset at `:1289` in the standalone prologue that is not in play.

The two populations separate cleanly, which is why the reset itself is **untouched**: it legitimately discards a prior callable transaction's uncommitted journal, since the normal writers run *after* setup and are committed or rewound. The only wrongly-discarded producer was the pre-call one-shot seed on the MTx EOA shortcut. **A conditional reset was explicitly rejected** — it would make correctness depend on which producers happen to run before setup.

This also **broadens #10876 beyond its title**: filed for a cross-transaction coinbase-credit loss, it is now known to discard state *within* a transaction too.

## The arithmetic was predicted before the fix, and it landed exactly

| base | arithmetic | verdict |
|---|---|---|
| `AccountState` after credit | 1,000,100 − 100 = 1,000,000 | wrong — gas debit missing |
| staged tuple | 880,000 − 100 = 879,900 | wrong — credit missing *(the symptom)* |
| durable post-gas base | 880,000 + 100 − 100 = **880,000** | **the declared value** |

Two earlier candidate patches were built and measured against this table and both failed; **arithmetic assembly (`staged_post + credit − value`) was refused** rather than shipped, because it computes a value instead of reading state.

## ⚠️ Merge hazard: GuestImage will be inherited silently, not conflicted

`GuestImage.lean` is **not in this diff** — at the merge base `4f096e476` it already carried `0x1c1395a0`/`0xbf2495a0` and the branch never touched it. Main carries `0x1c1395c0`/`0xbf2495c0`. **So git takes main's larger values with no conflict and nothing to resolve** — and a too-large bound **still typechecks under `omega`**, so the proof compiles, the build goes green, and `check-region-map` cannot see it because it only validates RegionMap against the ELF.

**After the regen to a fixed point, repin GuestImage's three literals from the merged ELF even though nothing conflicted there** (the size literal in `t7`'s `satWithin_ramRegion` argument, plus the base-plus-size end bound appearing twice, per its comment at `:203-207`). RegionMap's `textSizeBytes`/`bssSizeBytes` will conflict and get attention on their own; GuestImage is the quiet one. Main is `0x067794`/`0x1c1395c0`, the branch is `0x67860`/`0x1c1395a0`, and the merged values are a third set nobody has built — do not pick a side for any of the five constants.

Of tonight's three instances of this class, only this direction fails silently: a stale `.text` pin failed loudly on #10921, a `.bss` **growth** fails loudly because a too-small bound breaks the proof, and a **shrink** does not fail at all.

## Also included

The `EIP7708Logs` comment is corrected — it claimed its setup emits the seed after the reset, which is true for full dispatch and **false for the EOA route**. That is the fourth stale comment found tonight, and three investigations were misdirected by that class in a single session.

pirapira:
**Correction to my review above: I wrote "the fourth stale comment found tonight" and that count was inflated.** A systematic audit has since been run (#10932), so here is the accurate accounting:

| item | status |
|---|---|
| `EIP7708Logs` — claims the setup emits the seed after the reset; true for full dispatch, false for the EOA route | **corrected in this PR** |
| `BlockVerdictMtxRuntime` — claims the dispatcher resets only its own pending overlay while `AccountState` survives the MTx loop | **already tracked**, quoted as false in #10876, with #10877 open for it |
| `BlockVerdictCreationStage:16-27` — describes `stage_creation_runtime_payload` as the live top-level path, while `createStageInitcodeFrameRuntime:300-319` uses `stage_runtime_payload_code` | **newly filed as #10932** |
| `.Lbv_mtx_effects_kept` "success-only" | **not a stale comment — that was my own misreading**, corrected by scoper: the label is a join, and the code was never claiming otherwise |

So: **three genuine stale comments, one of which was mine to invent.** Two were already known or are fixed here; one is new. The audit found **no further current contradictions** in `BlockVerdictMtxRuntime`, `CreationStage`, `EIP7708Logs` or `Dispatch`.

The substantive point in my review stands and is if anything better supported by the audit's negative result: **three investigations tonight were misdirected by this class**, including a wrong diagnosis and a filed-then-rewritten issue (#10926) traceable to the `CreationStage` header now filed as #10932. The rate is what matters, not the tally — and a systematic pass finding only one remaining contradiction means the class is small enough to close out rather than live with.

PR #10933, opened 2026-07-30
docs(codegen): clarify legacy creation stager scope

### LINKED ISSUE #10932: docs(codegen): distinguish legacy STOP creation stager from live initcode route
## Finding

`BlockVerdictCreationStage.lean:16-27` documents `stage_creation_runtime_payload` as the conservative top-level creation staging path, says it only supports one-byte STOP initcode, and says broader constructors must remain gated pending later integration. That helper is legacy/probe-only: the source itself calls it a non-live stager at `:140`.

The production top-level creation route is `createStageInitcodeFrameRuntime` at `:300-319`: it selects `stage_runtime_payload_code`, copies the real transaction data as arbitrary initcode and calldata, supplies the M31 context, then enters `runtime_dispatcher_call`. The existing documentation can therefore send an investigator to the dead STOP-only helper when debugging live creation behavior.

## Audit scope

This was found during a targeted comment-to-emitted-path audit of `BlockVerdictMtxRuntime`, `BlockVerdictCreationStage`, `EIP7708Logs`, and `Dispatch`. The recent AccountState durability comment at MtxRuntime `:245-252` accurately describes the callable reset, and the EIP7708 seed comment accurately distinguishes full dispatcher setup from the MTx EOA shortcut after #10931.

## Requested correction

Make the legacy helper’s documentation explicitly say probe/legacy/non-live, and direct production readers to `createStageInitcodeFrameRuntime` plus `stage_runtime_payload_code`. No behavior change or regeneration is required.

pirapira:
Fixes #10932.

`stage_creation_runtime_payload` is a legacy probe-only, one-byte STOP helper. Its previous documentation said broader constructors must remain gated pending later integration, which misdirected readers toward a non-live path.

The comment now states that the limitation applies only to the legacy probe and directs production readers to `blockVerdictCreationRuntimeFunction`, which stages created address and creator environment before dispatching arbitrary initcode through `stage_runtime_payload_code`.

Validation: `lake build EvmAsm.Codegen.Programs.BlockVerdictCreationStage` (17 jobs).

Docs only: no emitted guest code or layout artifact changes.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdictCreationStage.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictCreationStage.lean
@@ -21,10 +21,10 @@ open EvmAsm.Rv64
     charged as transaction data but executed as frame code.
 
     This slice supports only one-byte STOP initcode. That is the narrow shape
-    whose execution does not observe ADDRESS/CALLER/ORIGIN/CALLVALUE, so the
-    later integration can run it before the created-address/creator-env
-    substrate exists. Broader constructors must stay gated until those fields
-    are staged soundly.
+    whose execution does not observe ADDRESS/CALLER/ORIGIN/CALLVALUE. These
+    constraints apply only to this legacy probe: they do not gate production
+    creation. `blockVerdictCreationRuntimeFunction` below stages the created
+    address and creator environment before dispatching arbitrary initcode.
 
     Calling convention:
       a0 = context record ptr (192-byte simple_transfer/multi_tx_nth_context


pirapira:
Reviewed and labelled. **Verified genuinely comment-only: `+4/-4`, one file, and zero changed lines touch an emitted asm string** — which matters in this codebase because a comment placed inside a converted asm block is drift, and a docs-only PR that silently altered emitted bytes would be the worst possible version of a documentation fix. No guest or layout change, so no pins move and no regeneration is owed.

## Why this earns a PR rather than a note

The comment being corrected **cost a wrong diagnosis and a filed-then-rewritten issue tonight.** #10926 was opened around a one-byte-STOP gate that is *not on the live path*, precisely because this header described `stage_creation_runtime_payload` as the live top-level path while `createStageInitcodeFrameRuntime:300-319` uses `stage_runtime_payload_code`. The investigation only recovered by reading the live route instead of the nearest plausible file.

The new text does the two things that prevent recurrence: it says the constraints apply **only to this legacy probe** and do not gate production creation, and it **names the routine that actually runs** (`blockVerdictCreationRuntimeFunction`), so the next reader is pointed forward rather than left to discover the discrepancy.

## Context: the audited class

Filed as #10932 from a systematic pass over `BlockVerdictMtxRuntime`, `CreationStage`, `EIP7708Logs` and `Dispatch`. The audit's **negative result is the valuable part** — no further current contradictions in those four files — which means this class is small enough to close out rather than live with.

Accurate accounting, since I previously overstated it in #10931's review and corrected it there: **three genuine stale comments**, not four. `EIP7708Logs` (fixed in #10931), `BlockVerdictMtxRuntime`'s "AccountState survives the MTx loop" claim (quoted as false in #10876, tracked by #10877), and this one. The fourth I had been citing — `.Lbv_mtx_effects_kept` reading as success-only — **was my own misreading**, corrected by scoper hours earlier, which I then carried into a tally as though the code had said it.

**Three of those three misdirected an investigation.** That rate, not the count, is the argument for fixing them: a comment that contradicts the code is drift with a measured cost, and it is the cheapest defect prevention available here.

PR #10934, opened 2026-07-30
docs: record post-merge EEST audit sets

pirapira:
Records the expected-absent and expected-present controls for the one full audit deferred until main contains #10931. This prevents stale queue picks and makes #10876 / 00337 disposition explicit.

### DIFF
--- a/docs/eest-postmerge-audit-sets.md
+++ b/docs/eest-postmerge-audit-sets.md
@@ -1,0 +1,38 @@
+# Post-merge EEST audit sets
+
+This is the acceptance manifest for the one deferred full audit.  Run it only
+after the merged `main` ELF contains #10931; do not reuse a pre-merge sweep or
+select fresh work from the frozen queue first.
+
+The sweep must record the resolved guest ELF path and SHA, `main` commit, runner
+SHA, manifest denominator, and the instrument SHA.  Classify each named fixture
+by its observed verdict/fail code and retain the raw rows used for comparison.
+
+## Expected absent from the false-reject population
+
+- `16190`
+- `16226`
+- `03489`
+- `03492`
+- the `23725` residual class
+- all 81 `shallow_stack` rows
+- `16397`
+- `16398`
+
+These are expected to have retired from the residual through merged work.  A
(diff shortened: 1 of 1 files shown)


pirapira:
**Not labelling this — closing it instead, and the ambiguity that produced it was mine.** I asked for the audit sets written "to a durable file," which reasonably reads as a repo file. What I meant was *somewhere that survives an agent losing context*, and the tracker is the right home for that.

## Why not the repo

`docs/eest-postmerge-audit-sets.md` is an **acceptance manifest for one deferred audit**. It goes stale the moment the audit runs, and nobody will delete it afterwards. `docs/` should hold documentation of the system; **work in flight belongs on the tracker**, where staleness is expected and harmless.

There is also a concrete cost right now: this consumes a full CI cycle (the ~1h build) and a merge-bot slot for an ephemeral note, while **#10931 — the actual 00504 fix — is a genuine conflict waiting for that same bot.** Ordering matters tonight.

## The part that IS durable, and worth keeping

There is a real distinction inside this PR:

- **The fixture list** — 16190, 16226, 03489, 03492, 23725's class, the 81 `shallow_stack` rows, 16397, 16398, 00504 at 60→0, 00337 for #10876 — is **specific to tonight's merges** and belongs in an issue comment.
- **The procedure** is genuinely reusable: *record the resolved guest ELF path and SHA, the `main` commit, the runner SHA, the manifest denominator and the instrument SHA; classify each named fixture by observed verdict and fail code; retain the raw rows.* That is a good standing recipe, and it encodes several things this fleet learned the hard way tonight — that a result without an ELF SHA cannot be checked against the source you think you tested, that a silently different manifest invalidates every comparison, and that a pass bit is weaker than a recorded fail code.

**So: put the fixture list on the tracker, and if you want the procedure preserved, propose it separately as an addition to the existing agent-facing docs** rather than as a single-use manifest. A reusable audit recipe earns a place in `docs/`; a nine-fixture expectation set does not.

Nothing about the audit itself changes — it still runs once, on the head that contains #10931, with the same expected sets and the same disposition for 00337.

PR #10935, opened 2026-07-30
fix(codegen): MPT inline/hash routing threshold 27 -> 32 per merkle_patricia_trie.py:223-225 (GH #10924)

### LINKED ISSUE #10924: codegen: withdrawals-root recomputation diverges once a withdrawal index needs two RLP bytes (flip at exactly index 256)
Measured on HEAD **371e2f414** (live stateless_guest, spike).

## The measurement (invariant across every hypothesis tested)

The multi-withdrawal family (manifest ids 25117–25135, `multiple_withdrawals_same_address[fork_Amsterdam-blockchain_test-multiple_blocks]`, all succ_bit=1 valid blocks) passes 17/20 and fails exactly the last three (b16/b17/b18 = ids 25133/25134/25135). The flip is at **withdrawal index 256**, nothing else:

| block | withdrawal indices | index RLP width | block RLP len | guest |
|---|---|---|---|---|
| b14 | 224–239 | 1 byte | 1061 | pass |
| b15 | 240–255 | 1 byte | 1061 | pass |
| b16 | 256–271 | **2 bytes** | 1077 | **fail** |
| b17 | 272–287 | 2 bytes | 1077 | fail |
| b18 | 288–303 | 2 bytes | 1077 | fail |

+16 block-RLP bytes at the flip = 16 withdrawals × 1 index byte each (fully accounted). Withdrawal count constant (16), `validatorIndex` constant per block (14–18, never crosses 255), recipient 0x0000000000000000000000000000000000000001 (ecrecover precompile — incidental; single-withdrawal and precompile-recipient fixtures pass). Balance/count hypotheses dead by measurement. Live-guest fail code = **31** (fail-code position: byte 112 LE u32 of the failure output; v2's 31 agrees — one zero-tx data point). The only reachable code-31 arm for this defect is `.Lbv_withdrawals_root_fail` (BlockVerdictStateRoot.lean:90-94): the block-hash check reads the *declared* wd_root from sv_params+40, so a wrong recomputation cannot propagate there.

## Encoder EXONERATED

`ssz_withdrawal_to_rlp` (SszWithdrawal.lean) routes both integer fields through `rlp_encode_uint_be` (RlpEncodeUintBeSAsm.lean) — a full minimal-integer encoder: strips leading zeros, emits minimal form, correct for len ≤ 55, and its u64 inputs are ≤ 8 bytes so always in-domain. The block_validate_withdrawals_root_indexed MPT keys are `rlp(0)..rlp(N-1)` (list positions, N=16 — small). **The one-byte assumption is NOT in the value encoder; do not patch it.**

## Open search: the defect is a length that is COMPUTED rather than MEASURED

Somewhere downstream of the encoder, a consumer assumes one-byte indices instead of using the encoder's returned length. Candidate shapes, in order:

1. A precomputed per-withdrawal record length (sized/summed with a constant rather than the encoder's returned length).
2. A pointer advance by a constant stride while walking the withdrawal list.
3. An outer list header whose total length is computed from a fixed per-record size, disagreeing with the actual concatenation.
4. A buffer or offset sized per record.

Fix when found: derive the length from the encoder's return value; do not widen any constant. Not a capacity residual (no sizing question); spec side is `fork.py:1217-1224` (no gas cost, no amount guard) and `Withdrawal.index` U64 minimal-length RLP integer.

pirapira:
## The defect

The bounded MPT trie builder inline-versus-stream routing threshold was 27 where the MPT rule requires 32. merkle_patricia_trie.py:223-225: when a node serialization is at least 32 bytes it is hashed; under 32 it is returned unhashed and embedded into the parent. The routing decision in MptIndexedTrieRoot.lean (li t0, 27; bgeu s3, t0, .Lmilr_stream) admitted values of 27-31 bytes into the direct-hash stream — a five-byte hole. The stream routine own guard (bltu s4, 27, .Lmislh_fail) also admitted 27-31 despite its docstring stating callers must prove the encoded leaf is at least 32 bytes (unenforced precondition — a defect generator, same class as #10932).

A withdrawal RLP value is 26 bytes when its index is one RLP byte (255 or less) and 27 when it needs two (256 or more). So any block whose withdrawal indices cross 255 routes every withdrawal leaf down the hash path where MPT semantics require inline embedding — and the guest rejects the block. Measured live on HEAD cf888605b (stateless_guest, spike): the multi-withdrawal family (manifest ids 25117-25135, all succ_bit=1 valid blocks) passes 17/20 and fails exactly b16/b17/b18, whose withdrawal indices are 256-271/272-287/288-303. The flip is at index 256 exactly; the +16 block-RLP delta at the flip is fully accounted by the 16 indices each gaining one byte. Live-guest fail code 31 (byte 112 of the failure output), reachable arm .Lbv_withdrawals_root_fail (BlockVerdictStateRoot.lean:521). The value encoder is exonerated: rlp_encode_uint_be is a general minimal-integer encoder, correct for its u64 inputs.

## The fix (derived, not fitted)

The routing threshold uses 32, the MPT embedding boundary — not widened to fit the failing fixtures, and nothing else changes. Encoded node = list header + hp path + value header + value, roughly value + 3-4 bytes, so value >= 32 implies node >= 35 >= 32 and the direct-hash stream is safe; values 27-31 now take the inline path, which re-checks the encoded node length and inlines under 32 or hashes at 32 or more — correct in both directions. The stream guard is tightened to 32 as well so its stated precondition is enforced rather than assumed.

Change is immediate-only (li t0, 27 -> li t0, 32, same instruction count): no layout repin needed (textSizeBytes, GuestAddrs, symbol-addresses.tsv unchanged).

## Verification (HEAD cf888605b, spike)

- The three previously-failing fixtures fail -> pass: 25133, 25134, 25135 (succ=1).
- All 17 previously-passing family members still pass (00560 + 25117-25132) — 20/20 for the family.
- Artifact identity verified before running: emitted .s contains li t0, 32 at both edited sites (the remaining li t0, 27 at .s:18683 is a different constant — a fail-code write, legitimate).

Fixes #10924.

### DIFF
--- a/EvmAsm/Codegen/Programs/MptIndexedTrieRoot.lean
+++ b/EvmAsm/Codegen/Programs/MptIndexedTrieRoot.lean
@@ -185,7 +185,7 @@ def mptIndexedStreamLeafHashFunction : String :=
   "mpt_indexed_stream_leaf_hash:\n" ++
   "  addi sp, sp, -176\n" ++
   "  sd ra, 0(sp); sd s0, 8(sp); sd s1, 16(sp); sd s2, 24(sp); sd s3, 32(sp); sd s4, 40(sp); sd s5, 48(sp); sd s6, 56(sp); sd s7, 64(sp); sd s8, 72(sp); sd s9, 80(sp)\n" ++
-  "  mv s1, a0; mv s2, a1; mv s3, a2; mv s4, a3; mv s5, a4; li t0, 6; bgtu s2, t0, .Lmislh_fail; li t0, 27; bltu s4, t0, .Lmislh_fail\n" ++
+  "  mv s1, a0; mv s2, a1; mv s3, a2; mv s4, a3; mv s5, a4; li t0, 6; bgtu s2, t0, .Lmislh_fail; li t0, 32; bltu s4, t0, .Lmislh_fail\n" ++
   "  mv a0, s1; mv a1, s2; li a2, 1; addi a3, sp, 88; jal ra, hp_encode_nibbles; mv s7, a0\n" ++
   "  addi a0, sp, 88; mv a1, s7; addi a2, sp, 96; addi a3, sp, 144; jal ra, rlp_encode_bytes; ld s7, 144(sp)\n" ++
   "  li t0, 1; bne s4, t0, .Lmislh_value_prefix; lbu t1, 0(s3); li t2, 128; bgeu t1, t2, .Lmislh_value_prefix; li s8, 0; j .Lmislh_value_ready\n" ++
@@ -259,13 +259,18 @@ def mptIndexedSortChangesFunction : String :=
 
 /-! Construct one canonical raw child reference for an indexed-trie leaf.
     Small values use the regular encoder into a fixed 1 KiB structural scratch;
-    every value at least 27 bytes uses the streaming helper, so no transaction
-    or receipt payload is copied into an MPT buffer. -/
+    every value at least 32 bytes uses the streaming helper, so no transaction
+    or receipt payload is copied into an MPT buffer.  The threshold is 32, not
+    27: the MPT embeds any leaf node whose serialization is under 32 bytes
+    (execution-specs merkle_patricia_trie.py:223-225), and a 27-31 byte value
+    can still encode to a <32 node, so routing it to the direct-hash stream
+    was wrong (GH #10924).  The inline path below re-checks the encoded node
+    length and hashes at >=32, so only values guaranteed >=32 go to stream. -/
 def mptIndexedLeafRefFunction : String :=
   "mpt_indexed_leaf_ref:\n" ++
   "  addi sp, sp, -64\n" ++
   "  sd ra, 0(sp); sd s0, 8(sp); sd s1, 16(sp); sd s2, 24(sp); sd s3, 32(sp); sd s4, 40(sp); sd s5, 48(sp)\n" ++
-  "  mv s0, a0; mv s1, a1; mv s2, a2; mv s3, a3; mv s4, a4; mv s5, a5; sd zero, 0(s5); li t0, " ++ toString itrIndexedKeyMaxNibbles ++ "; bgtu s1, t0, .Lmilr_fail; li t0, 27; bgeu s3, t0, .Lmilr_stream\n" ++
+  "  mv s0, a0; mv s1, a1; mv s2, a2; mv s3, a3; mv s4, a4; mv s5, a5; sd zero, 0(s5); li t0, " ++ toString itrIndexedKeyMaxNibbles ++ "; bgtu s1, t0, .Lmilr_fail; li t0, 32; bgeu s3, t0, .Lmilr_stream\n" ++
   "  mv a0, s0; mv a1, s1; mv a2, s2; mv a3, s3; la a4, itr_builder_node; la a5, itr_builder_node_len; jal ra, mpt_leaf_node_encode_from_nibbles; bnez a0, .Lmilr_fail; la t0, itr_builder_node_len; ld t1, 0(t0); li t2, 32; bgeu t1, t2, .Lmilr_small_hash; la t0, itr_builder_node; mv t2, s4\n" ++
   ".Lmilr_copy:\n" ++
   "  beqz t1, .Lmilr_inline_ok; lbu t3, 0(t0); sb t3, 0(t2); addi t0, t0, 1; addi t2, t2, 1; addi t1, t1, -1; j .Lmilr_copy\n" ++


pirapira:
Reviewed and labelled. **A one-immediate fix with a spec derivation, zero layout impact, and — I checked — a much wider population than the three fixtures it was found on.**

## The defect

`merkle_patricia_trie.py:223-225`: a node serialization of **at least 32 bytes is hashed**; **under 32 it is returned unhashed and embedded** into the parent. The guest routed at **27**, so leaves of 27–31 bytes were sent to the direct-hash stream path that the MPT requires to stay inline — **a five-byte hole**.

Both halves are corrected, which is right: the **routing** (`li t0, 27; bgeu s3, t0, .Lmilr_stream` → 32) *and* the stream routine's **own guard** (`bltu s4, 27` → 32), which previously admitted 27–31 despite its docstring stating callers must prove the leaf is at least 32 bytes. **An unenforced precondition stricter than its own check is what let the hole exist** — same class as #10932, and worth having closed in the same change rather than left as an assumption.

## Why the threshold is derived rather than fitted

The inline path already implements the correct rule (embed under 32, hash at 32+), so the routing was sending values *away* from a path that would have handled them. No new logic, no widened bound — the boundary simply has to be the MPT's own constant.

**This matters because my three earlier framings of this defect were all wrong:** multi-withdrawal *capacity*, then RLP *encoding width*, then a bound derived from the `u64` index range. The bound was never about index width — the index width only determined *which fixture crossed the 32-byte boundary*. k3's measurements never moved: flip at exactly index 256, `+16` bytes fully accounted, `a0` identified as builder **status** rather than a root, and 26 → 27 against a 27 threshold.

## Byte accounting, complete

Withdrawal value is **26 bytes** at indices ≤255 (`rlp(index)` 2 + `validatorIndex` 1 + address 21 + amount 1 + list header 1) and **27** at ≥256, because the index becomes 3 bytes as `0x82 0100`. Every byte named, the flip located at exactly the routing constant, no residue.

## Scope — wider than the PR states, and I enumerated the callers

`mpt_indexed_trie_root_bounded_from_values` has **three production call sites**, not one:

- `BlockVerdictStateRoot.lean:505` and `:509` (two calls)
- **`ReceiptsRootIndexed.lean:64` / `:107` — the receipts root**
- `WithdrawalsRootIndexed.lean`

**So the 27-byte hole affected receipts roots too, not only withdrawals.** Any block with a receipt or transaction leaf encoding to 27–31 bytes was mis-rooted. The population is therefore larger than the three withdrawal fixtures this was found on, and the sweep will show it.

**And the change cannot regress a correct fixture.** It alters behaviour *only* for leaves in the 27–31 band: under 27 both old and new route inline; at 32 and above both route to the stream. Everything in the changed band was provably mis-routed against `:223-225`. That is a containment argument from the constant itself rather than from fixture counts.

## Layout

**Immediate-only — `li t0, 27` → `li t0, 32`, same instruction count — so `.text` is unchanged and no pin moves.** `textSizeBytes`, `GuestAddrs` and `symbol-addresses.tsv` are all untouched, and the diff confirms zero pin lines. That makes this the cleanest codegen change of the night: one constant, two sites, no regeneration owed and none of the pin hazards that bit #10921 and needed hand-checking on #10928 and #10931.

## Gate

25133, 25134, 25135 green with the 17 currently-passing family members reported. Base is `cf888605b` so it sits behind main — routine, stays labelled, merge-bot's to rebase.

pirapira:
Answers to the two pre-push checks (coord 03:46Z):

**Caller enumeration (scope):** mpt_indexed_trie_root_bounded_from_values serves three indexed tries — transactions (BlockVerdictStateRoot.lean:514), receipts (ReceiptsRootIndexed.lean:64), withdrawals (WithdrawalsRootIndexed.lean:65 + BlockVerdictStateRoot.lean:521). Population prediction: transaction RLP leaf values are ~100+ bytes and receipt values ~50-300 bytes, always > 31, so tx/receipt leaves always route to the stream/hash path at BOTH thresholds and are unaffected by this change. The 27-31-byte hole affects only withdrawal leaves (26-31 bytes as index/amount/validatorIndex widths vary). Expected affected population: blocks whose withdrawal RLP values land in 27-31 bytes — the measured family (2-byte indices, 1-byte validatorIndex+amount) is the only current instance; the fix moves exactly those three fixtures plus any future block with 27-31-byte withdrawal leaves. Nothing in the tx or receipt tries.

**Inline path's own bound:** the inline path (mptIndexedLeafRefFunction :269-277) encodes the node via mpt_leaf_node_encode_from_nibbles into itr_builder_node, then RE-CHECKS the encoded node length against 32 and hashes when >= 32 — it implements the MPT rule itself, so 27-31-byte values are handled correctly for the whole hole: 30-31-byte nodes inline, 32-35 hashed, both per merkle_patricia_trie.py:223-225. The only residual is itr_builder_node's buffer capacity (~38 B max node for a 31-B value); if it is smaller than the max node the encode errors loudly rather than mis-rooting. Verification: 20/20 family fixtures pass with the fix (3 fail->pass, 17 unchanged), which covers values 26-27 bytes; the buffer bound for values 28-31 is not exercised by current fixtures.

Issue #10937, opened 2026-07-30
convergence: route top-level precompile execution through the shared message processor (EOA-vs-contract split already retired by bb22567a9)


pirapira:
## The spec has no EOA path

`vm/interpreter.py:110` `process_message_call` branches on **exactly one** condition — `:127` `if message.target == Bytes0(b"")` — i.e. **creation versus call**. Everything else reaches `process_message` at `:147`. **There is no branch on whether the recipient has code.**

An EOA transfer and a contract call therefore take the identical route: `move_ether` (`:384-390`) then execute the target's code, which for an EOA is empty and so is a no-op. The recipient being codeless changes *nothing* about the control flow.

**The guest's MTx-EOA-versus-contract split is a guest-only structure with no spec counterpart.**

## Why this is a defect generator rather than an optimisation detail

Every defect below is the *same* structural cause — an obligation satisfied on one path and forgotten on the other:

| defect | what one path had and the other didn't |
|---|---|
| #10925 | `stage_runtime_payload_code` documents CALLER/ORIGIN as left zero; `BlockVerdictDispatchTx:835-860` stages them, MTx EOA never did — and the self-transfer gas comparator read them anyway |
| #10931 | the upfront-balance publish had to be special-cased for the EOA route |
| #10917 (in resolution) | the relocated value debit sits after a flag guard that the EOA path **clears** before the dispatcher (emitted line 16278, dispatcher entry 16284), so the debit is **dead on that path** — plain EOA transfers would lose the sender debit entirely |
| #10619 (authorised) | the tx-scope checkpoint needs **two** boundaries because the EOA route publishes the gas half outside the seed routine |

Four separate repairs, one cause. **Each fix to one path leaves the other to be discovered later by a failing fixture.**

## Precedent: this is the third instance of the same doctrine

The maintainer has already ordered two structures of exactly this shape removed:

- **single-tx versus multi-tx** — *"we don't want to be debugging two similar code"*; the multi-tx version now serves the single-tx case, and the single-tx code was deleted so it stops appearing in greps.
- **zero-tx blocks** — routed through the MTx loop (#10921), because `fork.py:913-914` iterates transactions with no empty-block special case, and the `beqz` early exit was **deleted** rather than made unreachable.

The EOA split is the same argument with the same spec basis: **the spec has one path, so the guest should have one path.**

## Concrete consequence for #10619

[...]

PR #10939, opened 2026-07-30
refactor: route MTx empty code through dispatcher

pirapira:
## Rationale

`process_message` is shared by top-level calls and child messages (`interpreter.py:315`; top-level entry at `interpreter.py:147`, child entry at `system.py:373`). The MTx empty-code route duplicated the top-level setup instead of using the common dispatcher path.

This removes that duplicate EOA settlement and routes ordinary empty-code MTx recipients through `dispatch_tx_runtime_code`, which stages the zero-byte body and runs the shared dispatcher setup. Deferred status-2 lookup uses the same setup through `runtime_dispatcher_prepare_only`: prefix OOG continues to ordinary settlement, while a completed missing witness becomes structured code 8. The active-precompile adapter remains deliberately separate; moving its debit/credit ordering is follow-up work. Top-level creation duplication is tracked separately in #10938.

The refactor intentionally does not alter the current debit/credit order: the shared value debit remains post-dispatch in `BlockVerdictMtxRuntime`.

## Retired / retained ownership

- Retired: `blockVerdictMtxEoaSettlement`, recipient-credit helper, `account_state_publish_sender_upfront`, and the created-recipient scratch table/count.
- Re-provided: `runtime_dispatcher_prepare_only` is now called by `dispatch_tx_runtime_code` for the MTx deferred-lookup continuation.
- Retained: `bv_mtx_recipient_lookup_deferred` is consumed by the common dispatcher; the active-precompile lane remains its explicit adapter.

### Why deleting the durable upfront publisher is safe

This was a per-route workaround for #10876, not the sole remaining producer. In the emitted common dispatcher, the pending AccountState reset occurs at lines 35747-35750, before `dispatcher_seed_pending_upfront_balance` is called at 35797; the seed then records the tuple at lines 39277-39283. Thus the unified route writes after the reset and the tuple survives.

## Validation

- Serialized relink to fixpoint; ELF pins: `.text=0x66f5c`, `.data=0x5370`, `.bss=0x1c0eef40`; `bnf_le_a=0xa3000ee0`.
- olean / `.olean.hash`: 2132 / 2132.
- Passed: stateless link/layout, region-map, memory-layout coverage, asm-to-program (390 converted definitions), progress, drift, and `git diff --check`.
- No fixture/spike run: this is a source-structural refactor and does not claim behavioural exercise.
- `check-guarded-handler-bytes` was not run: its `lake env lean` helper cannot resolve cache-satisfied `GuardedHandlerSpecs.olean` in this checkout (the documented artifact-cache limitation); CI exercises the complementary non-cache gate.


pirapira:
Reviewed and labelled. **This removes a duplicate transfer/settlement path that execution-specs does not have.** `+1909/-2160`.

## Spec basis

`process_message_call` (`interpreter.py:110`) branches on exactly one condition — `:127`, `message.target == Bytes0(b"")`, creation versus call — and everything else reaches `process_message` at `:147`. **There is no branch on whether the recipient has code.** An EOA transfer and a contract call take the identical route: `move_ether` (`:384-390`), then execute the target's code, which for a codeless account is empty and therefore a no-op.

So the MTx-EOA settlement path had no spec counterpart, and routing ordinary empty-code recipients through `dispatch_tx_runtime_code` — which stages the zero-byte body and runs the shared dispatcher setup — is the convergent form. Treating `code_at_header_state_root` status 1 (absent account) and status 5 (empty-code witness miss) as zero-byte code is what makes the EOA case stop being special.

## What is deliberately preserved

**Deferred status-2 keeps its distinction.** Prefix-OOG continues to ordinary settlement while a completed-missing-witness becomes structured code 8, routed through `DispatchTx`'s existing `runtime_dispatcher_prepare_only` seam. Collapsing those two into a single reject would have been a soundness regression dressed as simplification — **rejecting for the wrong reason is how a false accept survives review**, and a fixture pair rejecting at two different codes needed a live-guest read to separate them earlier today.

That seam turned out to exist already and simply never be selected, so this is wiring rather than new capability — the sixth time in this sequence of work that a capability was present and unused.

## Ordering explicitly NOT changed

The body says so, and it should: **the shared value debit remains post-dispatch** in `BlockVerdictMtxRuntime`. This patch removes duplication and leaves the pre-existing ordering alone, so the guest now has the same ordering defect on **one** path instead of two. Separating duplication-removal from the ordering change keeps each diff to a single concern — and the follow-up (debit and credit into the pre-dispatch, post-checkpoint window) **absorbs the active-precompile seam by construction**, since once the transfer precedes dispatch the executor choice no longer needs its own envelope.

The target ordering there is three points, not two: **gas debit** (`fork.py:1105-1108`) → **snapshot** (`interpreter.py:380-381`) → **transfer** (`:384-390`, inside the `try`, restored at `:428-429`). The snapshot sitting between the gas debit and the transfer is the entire mechanism by which gas survives a revert while the transfer does not.

## Deletion accounting — the part no gate covers

Removing `account_state_publish_sender_upfront` needed justifying, since it was added days-fresh as a per-route workaround for #10876. **Confirmed on emitted asm:** `runtime_dispatcher_call` clears pending `AccountState` at lines 35747-35750, calls `dispatcher_seed_pending_upfront_balance` at 35797, and the flag tuple is recorded at 39277-39283. **The unified route therefore seeds *after* the reset**, unlike the deleted pre-dispatch EOA workaround which seeded before it and got wiped. So the emitter is callerless because the obligation is discharged elsewhere, not because the discharging path was deleted — which was the failure mode worth checking.

Also removed: `bv_mtx_created_recipient_count` and its table, scratch state left solely by the deleted route. That accounts for most of the **~298 KiB `.bss` decrease**.

## Pins — verified numerically

`textSizeBytes 0x067928 → 0x066f5c`; `bssSizeBytes 0x1c1395c0 → 0x1c0eef40`; `.data` unchanged at `0x5370`, so `bnf_le_a` stays `0xa3000ee0`.

**GuestImage's three hand-written literals were initially wrong by `0x10000`** — both end bounds read `0xbf1eef40` where `0xa3110000 + 0x1c0eef40 = 0xbf1fef40` — and the PR's **own `symbol-addresses.tsv` contradicted them**. Now corrected: size `0x1c0eef40`, both bounds `0xbf1fef40`, bases `0xa3110000` and `0xa3005370` unchanged, with `LAKE_ARTIFACT_CACHE=false lake build EvmAsm.Codegen.Proofs.GuestImage` passing over 1263 jobs rather than the bound being inferred from a relink.

Worth recording that this was the **loud** direction: a bound too *small* fails `omega`, because the region cannot fit inside it. The dangerous direction is too *large*, which typechecks silently while `check-region-map` validates only RegionMap against the ELF. What the error did reveal is that the `text`/`bss` pins had come from a later ELF than GuestImage was pinned against.

## Not in scope, tracked

Active precompile remains a deliberately separate execution adapter — its two surfaces already match the child dispatch, so relocating it is not new capability. Top-level creation duplication is #10938, genuinely a follow-on: the shared seam here is `DispatchTx`/`process_message`, and creation's initcode-versus-tx-data staging and creator-nonce/deployed-code effects do not travel through it.

**No fixture exercise**, per the standing instruction. The gate for this PR is spec-cited structure plus inventory coverage: the obligation inventory on #10937 (four parts, 35+ items, five gates) is what stands in for test evidence, and the two inbound label edges are covered by the assembler since an orphaned jump cannot ship silently.

PR #10941, opened 2026-07-30
docs(codegen): record the warmth rollback equivalence condition (GH #10937)

pirapira:
Records, in the source, the condition under which the guest's two frame-scope warmth restores stay equivalent to the spec's rollback. Comment-only.

## Why this is in the source rather than an issue

The two restores at `call_frame_return` are **correct** — that was settled as part of the GH #10937 reading. What is not recorded anywhere is *why*, and more importantly the condition under which they **stop** being correct. That condition would break silently, with no assertion anywhere to catch it, and the audience is whoever next touches those two lines.

## The spec side, at pinned `e5a8caf1b`

The EIP-2929 accessed sets are **frame** state, not `tx_state` and not `tx_env`:

- declared on `Message` (`vm/__init__.py:161-162`) and `Evm` (`:187-188`);
- `tx_env.access_list_addresses` is only a **seed** — `utils/message.py:50-53` builds a fresh `set()` and `:88-89` passes it, so `tx_env` never carries live warmth;
- each child receives `.copy()` — `vm/instructions/system.py:149-150` (CREATE family), `:367-368` (CALL family);
- `incorporate_child_on_success` (`vm/__init__.py:235-236`) merges them back;
- `incorporate_child_on_error` (`:302-304`) reabsorbs **only** `gas_left`, `state_gas_left`, `regular_gas_used` — no accessed sets.

So a failed child's warmth is discarded from the parent's view, which is exactly what the two restores implement.

Worth noting for future readers: **absence from `restore_tx_state`'s field list proves nothing here.** That primitive restores `tx_state`; a quantity living on the frame is absent from it whether it is rolled back elsewhere or never rolled back at all. The negative has to come from the frame side, which is where the citations above point.

## The condition being recorded

The spec rolls warmth back by **discarding a copy**. The guest rolls it back by **restoring a count**. A count is a high-water mark: it undoes *appends* but cannot undo an *in-place edit* of an entry below the mark. The two mechanisms therefore agree only while the warmth tables are **append-only** — true today, since these are membership sets with no mutable per-entry payload. Giving warmth entries a mutable field would diverge from the spec silently.

Both arenas get the note, because there are two and they match the spec's two sets one-to-one:

| guest arena | spec set | restore site |
|---|---|---|
| `evm_storage_access_count` | `accessed_storage_keys` | `CallFrameReturn.lean:182` (env+648) |
| `evm_access_account_count` | `accessed_addresses` | `CallFrameReturn.lean:185` (env+720) |

## Behaviour

**BEHAVIOUR-NEUTRAL.** The diff adds only Lean `--` comments, outside every emitted string literal, so the emitted assembly cannot change — no layout, no `.text`/`.data`/`.bss` size, no symbol address.

Stated **structurally, not as a measurement**: no build was run, per the current fleet-wide stand-down on builds. If a reviewer would rather have the byte-identity measured than argued, that is a reasonable ask and I will run it once builds resume.


pirapira:
Labelling. Behaviour-neutrality **verified from the diff**, not accepted as stated: the 13 added lines are Lean `--` comments sitting between string literals in the `++` chain, outside every literal. The two emitted strings — `ld t0, 648(x20); la t1, evm_storage_access_count; sd t0, 0(t1)` and the `720(x20)` / `evm_access_account_count` twin — are byte-identical, and the placement matches the existing comment pattern in that file. So the emitted assembly cannot change and no size or address can move.

Checking that mattered rather than being a formality: **a comment placed *inside* a converted asm block is drift in this codebase**, and a comment-only PR in an emitting module is exactly where that lands. This one is on the right side of the boundary.

Stating the neutrality structurally rather than as a measurement was the correct call under the build stand-down, and the offer to measure byte-identity later is noted but not needed — the diff settles it.

## Why this is worth an hour of CI on a hub-reachable file

The condition recorded here is the kind that has no other home. It is not a defect, so it cannot be an issue; it is not a behaviour, so no test can hold it; and it would break **silently**, with no assertion anywhere. The audience is whoever next touches those two lines, and the note is now in front of them.

The reasoning trap in the PR body is the part I would most want preserved: **absence from `restore_tx_state`'s field list proves nothing for a quantity living on the frame.** That trap is what made this look like a divergence in the first place, and naming it in the body means the next reader inherits the method rather than repeating the investigation.

Both arenas got the note, which is the other half of the value — there are two warmth arenas matching the spec's two sets one-to-one, and a note on only `evm_storage_access_count` would have implied the other was unexamined.

Issue #10944, opened 2026-07-30
Top-level CREATE endowment is never transferred: no sender-debit/created-address-credit pair on the creation route, while an EIP-7708 Transfer log is still staged


pirapira:
## Claim

**A top-level CREATE transaction with nonzero `tx.value` has no sender-debit / created-address-credit pair at all.** The guest makes the *initcode* see the endowment and emits a created-address final state on success, but the shared `move_ether` bookkeeping never runs on that route.

Found by `evm-asm3-codex` while ordering the shared body capture for #10937's change two. **Source reading on `main`; direction not measured.** Pre-existing, not introduced by that work.

## Spec basis (pinned `e5a8caf1b`, Amsterdam)

The endowment is not a creation-specific transfer — **it is the same transfer as any call's**:

- `vm/interpreter.py:212` — `process_create_message` **calls `process_message`**. Creation is a wrapper around the shared processor, not a parallel engine.
- `:384-390` — `move_ether` sits inside `process_message`'s `try`, debiting the caller and crediting `current_target`.
- `:429` — restored on failure by the snapshot taken at `:380`.

So a top-level CREATE with value must debit the sender and credit the created address, **inside the captured window**, and have that undone by rollback on failure.

## Guest

- The value-transfer helper's **sole flag producer** is `BlockVerdictDispatchTx.lean:1010`, on the **ordinary recipient path**.
- `CreationStage` **never writes** `bv_pending_recipient_credit_flag` / `addr` / `pre` / `post`.
- `CreationStage.lean:425-447` stages only the **EIP-7708 Transfer log descriptor**.
- `:451` calls the transfer helper with that flag **still zero** — so the call is a no-op.
- No top-level account-state / nonstorage **sender-debit + created-address-credit** producer exists before dispatch.

What the guest does instead:

- seeds `env+32` from `tx.value`, so the **initcode's own live balance** is correct (`BALANCE`/`SELFBALANCE` inside the constructor see the endowment);
- on success only, emits the created address's final state at `:503` / `:577-582`.

That is not the shared `move_ether` pair, and it happens **after** success is known.

## ⛔ The sharpest part: a Transfer log is emitted for a transfer the state never records

[...]

PR #10945, opened 2026-07-30
fix(codegen): checkpoint value transfer after preparation

pirapira:
## Summary

Align the guest's top-level message bookkeeping with the two phases in pinned execution-specs process_message: the sender's upfront gas/blob debit is outside the body snapshot, while the value move is inside it and rolls back with a failed body.

- Split the prior combined producer into a pre-mark sender-debit materializer and a post-mark atomic sender-debit/recipient-credit materializer.
- Move ordinary MTx, single-transaction, and direct CreationStage routes onto the same upfront -> capture -> value pair -> dispatch contract.
- Replace the MTx receipt-status-specific sender debit with shared rollback; read sets deliberately remain outside rollback.
- Centralize all body snapshot fields, including effect counters, account-write undo state, account-state pending state, and creator-nonce undo.
- Retain CreationStage's intentional no-op value-helper call as the landing seam for #10944: top-level CREATE currently stages its Transfer log but not the state-effect operands.

The value pair reads the just-materialized live sender balance before subtracting CALLVALUE; it is not based on the stale pre-capture balance. It publishes both move_ether halves together, so rollback cannot retain a credit without its matching debit.

## Spec basis

Pinned execution-specs at e5a8caf1b performs the sender debit before the process_message snapshot, moves ether in process_message (interpreter.py:384-390), and restores that snapshot on failure (interpreter.py:429). The guest now preserves those durability boundaries. Creation uses the same ordering seam; #10944 separately tracks its missing operand staging.

The shared capture/restore field set is explicit:

- nonstorage effect count and overflow;
- code effect count, next cursor, and overflow;
- persistent-storage, transient-storage, and event-log counters (evm_env +448/+464/+472);
- account-write undo cursor;
- AccountState pending count, delete count, and overflow;
- creator-nonce undo cursor.

This is the write-side state restored by the transaction snapshot. Read sets are intentionally excluded: execution-specs destroy_storage promotes created-account storage writes to reads before deletion (state_tracker.py:560-579), and those reads survive restore_tx_state. The source comments retain the two latent-equivalence conditions: dispatcher_tx_gas_settle must not write a captured arena before the status-directed restore, and the creation route retains the same ordering seam even though its value flag is presently unpopulated (#10944).

## Creation-route audit

CreationStage bypassed both ordinary seam producers; it now explicitly takes the shared three-call sequence before runtime_dispatcher_call. The dispatcher reads its PC from runtime_dispatcher_input_ptr in its prologue, so a0/x10 is not a caller input. The helpers preserve ra, t0-t6, a0-a7, and x20; capture writes only t0-t2.

Retired-symbol reader counts are zero:

- dispatcher_seed_pending_upfront_balance
- dispatcher_seed_pending_recipient_credit
- bv_xfer_sender_bal
- Lbv_mtx_xfer_debit

## Layout

Final regenerated ELF / TSV:

- .text = 0x66dcc
- .data = 0x5370 (unchanged)
- .bss = 0x1c0eefa0
- bnf_le_a = 0xa3000ee0 (unchanged)
- __BSS_END__ = _end = 0xbf1fefa0

RegionMap and GuestImage use these TSV-derived values.

## Verification

- Targeted builds: BlockVerdictCreationStage, BlockVerdictDispatchTx, BlockVerdictMtxRuntime, Dispatch, RegionMap, and GuestImage.
- scripts/check-region-map.sh: pass; ELF/TSV/RegionMap agreement.
- scripts/check-asm-to-program.sh: pass; 390 converted definitions across 148 files.
- git diff --check.

No fixture run: this is bookkeeping/rollback wiring, and the requested acceptance here is the structural layout gate.


pirapira:
## Review: verified against the diff, labelling

This is change two of the sequence the maintainer set — *one path for the transfer* (#10939), then *relocate the debit*. Checks below were run against head `1cab65f95`, base `2d374c072`.

### It is a net deletion, which is the right sign for this work

Excluding the two regenerated artifacts (`GuestAddrs.lean` `+360/-360`, `symbol-addresses.tsv` `+1490/-1480`), the real source delta is **+113/−167 — a net removal of 54 lines** while adding a shared capture/restore mechanism and four call-site conversions. A de-duplication that grows the source is usually a re-parameterisation in disguise; this one shrinks it.

`MtxRuntime` alone is `+7/−98`, and `BlockVerdictFunction` `+1/−23` — the two caller-side partial marks and the postlude debit coming out.

### Claims verified, not accepted

- **The receipt-status guard on the value debit is genuinely gone.** The removed line is the branch itself: `ld t2, 0(t2); beqz t2, .Lbv_mtx_xfer_debit_done`, along with the comment that "status is re-read from `bv_tx_status_arr` exactly as the postlude below does." That is the conversion this slice exists to perform — a status predicate replaced by positional rollback.
- **Both latent-equivalence conditions are in the source**, not just the body: the settlement-fold note citing `interpreter.py:429` and stating that `dispatcher_tx_gas_settle` writes no captured arena, naming its only two stores.
- **No `sorry`/`admit`, no `bv_decide`/`native_decide`** added.
- **Layout repin is TSV-derived and arithmetically correct.** `GuestImage` both bounds `0xbf1fef40` → `0xbf1fefa0`, extent `0x1c0eef40` → `0x1c0eefa0`; `0xa3110000 + 0x1c0eefa0 = 0xbf1fefa0` ✓, matching the TSV's `__BSS_END__ = 0xbf1fefa0`. `RegionMap` `textSizeBytes 0x066dcc`, `bssSizeBytes 0x1c0eefa0`.
- ⭐ **`dataSizeBytes` does not appear in the diff at all** — the cleanest possible evidence that `.data` is untouched at `0x5370` and `bnf_le_a` cannot have moved. That was the hazard I flagged before the storage was written, and it was avoided *by construction* (`.zero`-only declarations promoted to `.bss` by `Layout.moveZeroDataLines`) rather than by placement discipline.

### An incidental correction worth noting, since nothing else would have caught it

The `RegionMap` docstring said `.data` ends `0xa3005310` and `.bss` ends `0xbe318fc0`. Both were **already wrong on main** — `dataSizeBytes = 0x5370` puts `.data`'s end at `0xa3005370`, and `.bss` ended `0xbf1fef40`. This PR corrects them. Prose carrying addresses decays silently because no gate reads it; the pinned `def`s were right the whole time. Worth having fixed while regenerating.

### What I asked for and got

Four found-and-fixed gaps, the last of which is the one I would have missed: **`CreationStage` bypassed *both* ordinary seam producers** while still reading the former shared snapshot fields in its deposit-exception path — no undefined label, no type error, green build, and a restore reading fields nobody captured any more. It was found by grepping retired symbols for remaining *readers* rather than trusting the build, and the reader counts came back zero on all four (`dispatcher_seed_pending_upfront_balance`, `dispatcher_seed_pending_recipient_credit`, `bv_xfer_sender_bal`, `.Lbv_mtx_xfer_debit`).

Also: per-site emitted-ASM liveness audits at all four insert sites, with the register sets named — necessary because a register clobber is type-correct and a focused build proves nothing about it. `a0` aliases `x10`, the dispatch PC, so a call inserted where `x10` is live redirects execution rather than producing a wrong byte.

### Two things a reviewer should read the body for rather than infer

1. **The transfer helper is a deliberate no-op on the creation route**, kept as a landing site for #10944 (top-level CREATE endowment is never transferred, while an EIP-7708 Transfer log *is* staged). Seeing a helper called with a permanently-zero flag reads as dead code; deleting it would remove the seam a fix needs.
2. **The single-tx route gains a rollback boundary it did not have.** `BlockVerdictFunction:840-877` marked the exec logs but not `account_writes_undo`, AccountState, transient/event or create-nonce undo. That is authorised, not scope creep: there is no single-tx concept in execution-specs — a one-transaction block is just a block, snapshotted identically — so the differing boundary was never a baseline to preserve. Direction unmeasured, and it is stated that way.

### Gates

`check-region-map` **passed** on a fully materialised tree (825 `.olean` / 825 `.olean.hash`) after the rebase. The earlier 795/825 deficit and `.ilean` permission-denied were transient contention — the olean count, not the process list, was the reliable instrument; the process list gave false positives in both directions. No `chmod`, no deletions.

CI is the gate for the rest. No fixture runs, per the stand-down — and per the maintainer's ruling that these de-duplications merge without testing, with debugging resuming afterward.

PR #10946, opened 2026-07-30
refactor(codegen): share runtime witness context staging

pirapira:
## Summary

Extract the duplicated runtime account-witness trailer into `stage_runtime_payload_witness_context`, then invoke it from the ordinary contract-dispatch and top-level creation paths. The helper writes the three existing lengths at `env+472/+480/+488` and copies header, state, then codes contiguously from `env+496`.

This is semantically neutral by construction, not byte-neutral: the falsifier is any change to those three lengths, offsets, or the ordered header→state→codes spans.

The new helper calls were liveness-audited. In both callers `ra` is already spilled at routine entry; all t/a values at the old inline boundary are dead or redefined before their next read.

Also document the cross-routine body-state capture/restore pairing and the pre-increment creator-nonce invariant. The guest already carries the pre-increment nonce, so copying the execution-specs stored-post-nonce minus-one would be wrong.

## Layout

- fixed-point relink: `.text=0x66d68`, `.data=0x5370`, `.bss=0x1c0eefa0`
- `GuestAddrs.bnf_le_a` remains `0xa3000ee0`; `.data` and `.bss`/GuestImage bounds are unchanged
- regenerated `symbol-addresses.tsv` and `GuestAddrs.lean`; `GuestImageEntries.lean` was unchanged

## Checks

No fixture run: this is a structural refactor with no intended behavioural delta.


pirapira:

First cut of #10938. Checks run against head `269d0ce49`.

### ⭐ The source grows by 25 lines and the emitted image *shrinks* by 100 bytes — and that is the right metric here

| | |
|---|---|
| net source (excluding regenerated `GuestAddrs` / TSV) | **+73 / −48** |
| `textSizeBytes` | `0x066dcc` → **`0x066d68`** (−0x64 = **−100 bytes**) |

A de-duplication whose shared body must be *added once* and then referenced from 13 closures will show a small source increase; that is bookkeeping, not growth. **What it removes is duplicated emitted code, so `.text` is the measure that matters, and `.text` went down.** Worth stating in these terms because a reviewer checking only the diffstat would read this as a net addition.

The two call sites are `−23` and `−24` against `+7`/`+9`, so the inline regions genuinely went away rather than being wrapped.

### Verified

- **`RegionMap` changes exactly one line** — `textSizeBytes` only. `bssSizeBytes` and `dataSizeBytes` untouched, so `.data` stays `0x5370`, `bnf_le_a` stays `0xa3000ee0`, and **`GuestImage` needs no repin at all** (its literals are `.bss`-derived and `.bss` did not move). That is the cleanest possible outcome for the address-pinned-`rfl` hazard.
- **No `sorry`, no forbidden tactics.**
- `check-asm-to-program` and `check-region-map` both **passed** locally — the first matters more on this diff than on #10945, since it moves emitted strings across 13 closures and is the only gate that sees an emitted-string versus rendered-`Program` divergence.
- The helper is emitted once at `0x80024188` in `.text` per the TSV, and both call sites pass their **own** operands: creation uses `sv_pre_rlp` / `bv_witness_state` / `svf_codes`, ordinary dispatch uses `dtrc_hdr` plus `mv a3, s0` / `mv a4, s1`. **Phase-sensitive inputs preserved per caller, which was the non-negotiable constraint.**
- Docstring states the calling convention (`a0` payload, `a1/a2` header, `a3/a4` state, `a5/a6` codes) and flags the clobber list as **conservative** — the body writes `t0`–`t3` while the doc declares `t0`–`t6`. Over-declaring is the safe direction and it is marked as deliberate, so a later reader can tell the doc from the body.

### Scope discipline is the substance of this PR

Creation-side unification is where a "simplification" can quietly flatten a real difference, so what this change *excludes* is as important as what it shares. Left alone deliberately: the computed-versus-recipient address, sender `CALLER`/`ORIGIN` staging, `SELFBALANCE`, the transaction-data versus calldata split, dispatcher lifecycle, deploy accounting, `generic_create`, and `frame_return`.

Three of those exclusions are load-bearing rather than cautious:

- **`frame_return` is the nested `incorporate_child_on_error` analogue, not a duplicate tx restore** — established independently from the guest side and from the spec, where `copy_tx_state`/`restore_tx_state` sit *inside* the shared processor and only the incorporate pair is nested-only.
- **The opcode layer has no top-level counterpart at all** (no instruction exists), so it is not duplication in any sense.
- **Deploy accounting exists only in the top-level tail with no second implementation**, so the nested side either charges by another mechanism or does not charge. That is an open gap, not a licence to unify past it.

### Neutrality claim is correctly scoped

**Semantically neutral by construction, not byte-neutral** — replacing inline stores and copies with a leaf call changes the emitted bytes by definition. The falsifiable equivalence is stated: the same three lengths at `+472`/`+480`/`+488`, and header → state → codes copied contiguously in that order. A reviewer can check that against the helper body; a bare neutrality assertion they could not. `.text` shrinking by exactly the size of one removed copy region corroborates it.

### On the BEHIND status

Leave it. `1cab65f95` is an ancestor of `d7f667ead`, so the PR already contains only its own two commits — it shows behind because main advanced *after* the push, which is routine and not a reason to touch a branch. Correct call not to rebase reflexively.

Liveness was audited per site before this landed: `ra` spilled at both enclosing function entries and read only at their returns; `t` registers at both boundaries holding completed dead values; `a0`'s former stager status consumed by its `bnez` *before* the call at the creation site. Six insert sites audited across this and #10945.

Issue #10950, opened 2026-07-30
MTx creation success falls through into the conservative-bail arm: .Lbv_mtx_creation_post has no successor edge, so a settled creation result is discarded and no receipt is published


pirapira:
## Claim

**A successfully-executed top-level creation transaction on the MTx path falls straight through into the conservative-bail arm, so its result is discarded and no receipt is ever published.** Not a bail predicate firing, not a late store, not a dead store — **a missing successor edge.**

Root-caused by `evm-asm3-codex` on fixture 21087 (`empty_transaction3`, 5,244 bytes) at `d0a12910e`, with a **validated** instrument. This is the first root cause from the resumed convergence loop.

## The code

`BlockVerdictMtxRuntime.lean:782-785`:

```
.Lbv_mtx_creation_key_copy:
  beqz t2, .Lbv_mtx_creation_post; …; j .Lbv_mtx_creation_key_copy
.Lbv_mtx_creation_post:
  la t0, bv_mtx_i; ld t1, 0(t0); slli t0, t1, 3; la t3, bv_tx_status_arr; add t3, t3, t0; ld a4, 0(t3)
.Lbv_mtx_creation_unsupported:
```

⛔ **`.Lbv_mtx_creation_post` has no terminating jump.** It loads the status and then falls directly into `.Lbv_mtx_creation_unsupported` — the conservative-skip arm — which sets `bvReceiptsShapeSet 60` and continues to `.Lbv_mtx_bail_after_shape`. The normal MTx completion path that would set `bvgr_runtime_count = tx_count` is never reached.

So the success arm and the unsupported arm are **adjacent labels with nothing between them**, and success is distinguished from bail only by source layout — which is not a control-flow construct.

## Measured, with the instrument validated first

On 21087 at `d0a12910e`:

- sentinel `OUTPUT+504 = 170` — **proves high-window stores survive**, so a zero in that window is a reading rather than an artifact;
- `bvgr_runtime_count` = 0 before *and* after arena prepare; `bvgr_arena_tx_count` = 1 after;
- post-settle marker **187** and post-log-snapshot marker **204** both reached — so the transaction **did execute**;
- `bv_creation_output_mode = 1` (set deliberately at `MtxRuntime:774`), MTx-publish marker **221** reached, **scalar marker never reached**;
- `brr_control.count = 0` with zero receipt-record words.

[...]

Issue #10947, opened 2026-07-30
Nested CREATE/CREATE2 pays neither deploy charge: code_hash_gas (6/word, regular) and code_deposit_state_gas (1530/byte, state) exist only in the top-level creation tail


pirapira:
## Claim

**A nested `CREATE`/`CREATE2` deploys code without paying either deploy charge.** Both charges exist in the guest only in the top-level creation tail; no nested creation path computes either.

Two independent readings, deliberately split so neither saw the other's intermediate work: `scoper` on the pinned Python, `k3` on the emitted guest at `d7f667ead`. They agree.

## Spec (pinned `e5a8caf1b`, Amsterdam) — shared, with no depth carve-out

`create` (`system.py:167`) and `create2` (`:215`) both enter `generic_create` (`:65`), which calls `process_create_message` at `system.py:154` — **the same function** the top level calls at `interpreter.py:127`. No wrapper on either side. The deploy accounting at `interpreter.py:213-241` is inside it.

⭐ **The decisive negative: `process_create_message` (`:176-243`) contains no reference to `depth` at all** — the whole body was grepped and it is empty. The only branches in the deploy block are on `evm.error` (`:213`) and the `try`/`except` (`:214`, `:232`). No depth test, no caller-identity test, no alternative charge path. And no depth-dependence one level down either: `charge_state_gas` (`gas.py:314-339`) is reservoir-first, then spill into `gas_left`, then `OutOfGasError` — no depth term.

**`interpreter.py:213-241` is the only implementation of either charge in the fork.**

## ⛔ Two charges, two pools, two symptoms — they must not be conflated

| charge | formula | constant | charged via | pool |
|---|---|---|---|---|
| `code_hash_gas` (`:222-227`) | `OPCODE_KECCAK256_PER_WORD × ceil32(len)/32` | **6 per word** (`gas.py:229`) | `charge_gas` | **regular gas** |
| `code_deposit_state_gas` (`:228-231`) | `len × COST_PER_STATE_BYTE` | **1530 per byte** (`gas.py:40`) | `charge_state_gas` | **state gas** |

A missing 1530 term under-charges **state** gas and surfaces as a reservoir or `state_gas_used` discrepancy. A missing 6 term under-charges **regular** gas and surfaces in `gas_used`. **Both are missing here**, but they are separate defects with separate witnesses and a partial fix is possible — do not measure them as one number.

## Guest, from emitted asm at `d7f667ead`

[...]

PR #10951, opened 2026-07-30
codegen: finalize successful MTx creations

pirapira:
## Summary

Successful multi-transaction creations now jump explicitly from `.Lbv_mtx_creation_post` into `.Lbv_mtx_effects_kept`, the existing shared transaction-finalization postlude. The former control flow fell into the conservative unsupported-creation prefix path and never materialized the transaction receipt.

`fork.py:1147-1204` processes every transaction after `process_message`: gas/refund and fee settlement, receipt construction, then `incorporate_tx_into_block`. It has no creation-specific postlude. `.Lbv_mtx_effects_kept` is already a shared finalization target with two callers in `BlockVerdictSimpleTransferPublish.lean:50` and `:145`; creation is its third caller.

The conservative `.Lbv_mtx_creation_unsupported` path remains for the seven genuine failure routes. This is an explicit edge, not label reordering.

## Evidence

- `21087_test_empty_transaction3…` changes from receipt-root mismatch code **53** to BAL-digest mismatch code **60**. It still fails; this PR deliberately exposes the downstream BAL substitution rather than claiming the fixture passes.
- Validated trace: OUTPUT+504 sentinel `170` survives; positive markers `187` (gas settlement), `204` (log snapshot), creation output mode `1`, and MTx-publish arrival `221` prove the creation takes the shared MTx route.
- Expected-invalid control `04778…below_intrinsic…` still rejects with code **5**.
- `check-asm-to-program.sh` and `check-region-map.sh` pass.

## Layout

One new jump: `.text` is `0x66d6c` (+4), `.data` remains `0x5370`, `.bss` remains `0x1c0eefa0`, and `bnf_le_a` remains `0xa3000ee0`; no GuestImage repin is required.

Follow-up: #10950 remains open until the fixture itself passes; its equal-length code-60 BAL mismatch should be diagnosed after the planned depth unification.


pirapira:

Maintainer's gate for this one is explicit: **the shifted failure code is enough to merge.** 21087 still fails — `53` → `60` — and the PR says so rather than implying a pass.

### The whole defect was a missing jump

Source change is `+8/-7` in one file, and the functional part is **one instruction**:

```
+  "  j .Lbv_mtx_effects_kept\n" ++
```

`.text` `0x066d68` → `0x066d6c`, exactly **+4 bytes**. `dataSizeBytes` and `bssSizeBytes` do not appear in the `RegionMap` diff and `GuestImage.lean` is not in the changed-file set at all — so `.data` stays `0x5370`, `bnf_le_a` stays `0xa3000ee0`, and there is **no address-pinned-`rfl` exposure and no repin to get wrong**. That is the cleanest possible shape for this hazard class.

### ⭐ It also fixes the comment that let the defect survive

The old text on the unsupported arm read *"A creation transaction is not yet dispatched by this loop"* — **untrue since creation was wired**, because `MtxRuntime:774` sets `bv_creation_output_mode = 1` and `CreationStage` scatters a settled result into the `bv_mtx_*` arrays. A confident, obsolete comment sitting directly above a fall-through is why "success lands in the bail arm" read as intentional. The replacement is accurate and scoped to the failure case only.

Worth stating because it is the third time today a comment or label asserted something the code did not do: `.Lcfrc_create_gas_charged` reads the dispatcher's total-used value (#10947), `DispatchTx:16-19` describes `.Lbv_after_tx_gas_precharge` as "the conservative bail" when it is the shared continuation, and this one. **Prose in this tree decays silently because no gate reads it.**

### Why the target is right, from the call graph rather than from argument

`.Lbv_mtx_effects_kept` already has two callers — `BlockVerdictSimpleTransferPublish.lean:50` and `:145`. Creation becomes its **third**, so this converges on an existing shared postlude instead of adding a special case. Spec basis is `fork.py:1147-1204`: after `process_message` returns, every transaction gets settlement, receipt construction and `incorporate_tx_into_block`, with no creation-specific postlude to fall out of.

The conservative arm is correctly **kept** for the seven genuine parse/result-failure routes (`:744`, `:747`, `:748`, `:765`, `:767`, `:768`, `:777`), and the fix is an explicit jump rather than a label reordering — adjacency was the cause, so a layout change would have been the wrong repair.

### The trace behind it is the part I would preserve

This is a root cause rather than a plausible edit because the instrument was validated **before** the readings were trusted: sentinel `OUTPUT+504=170` proving the window survives, then **positive** markers (post-settle 187, post-snapshot 204, `mode=1`, MTx-publish arrival 221) rather than an argument from zeros.

That mattered concretely: an earlier `a0` probe at `OUTPUT+336` was invalid — the epilogue overwrites it — and had already retired two hypotheses wrongly before being caught. **Every hypothesis on this fixture predicted a zero somewhere**, so an unvalidated probe would have confirmed whichever story was brought to it. Four framings died here, and the two that died to that bad cell were correctly restored to *untested* rather than left refuted.

### Follow-ups, deliberately not in this PR

- **#10950 stays open** — its subject is the discarded creation result, and the receipt is only half of that. Referenced, not closed by keyword.
- The newly exposed `code 60` is recorded and **not** being chased: rebuilt and supplied BAL lengths are both **327**, so it is a *substitution*, not an omission — length is blind here and a byte diff would mislead, since one moved item shifts everything downstream. Decode to address then per-field RLP when the time comes.
- Per the maintainer, 21087 gets retested **after** the structural changes: a repair built against the current structure would have to be redone once the depth unification lands. Don't build on ground that is about to move.

Issue #10782, opened 2026-07-28
Proof: compose the 4 block lemmas into a whole-routine spec for WithdrawalDecodeClose (blocks proven, composition missing)


pirapira:
Proof work on a **stable leaf**. The **most advanced** of the four filed today — the straight-line blocks are already proven, so the remaining work is composition rather than groundwork.

## Current state, measured on `origin/main`

`EvmAsm/Codegen/Programs/WithdrawalDecodeClose.lean` — **339 lines**, hosting the straight-line block lemmas of a 60-instruction routine.

**What already exists:** 4 theorems and ~75 supporting lemma/spec references:

| lemma | covers |
|---|---|
| `wdPrologue` | entry block |
| `wdEpiCore` | epilogue core |
| `wdSuccessTail` | success exit |
| `wdFailTail` | failure exit |

All are stated generic over an untouched frame `G`, which is the convention the rest of the verified core uses — so they compose without re-framing.

**What is missing:** the **whole-routine triple** that chains prologue → core → the two tails into a single specification of withdrawal decoding, with the success and failure post-conditions distinguished.

## The task

Compose the four block lemmas into one routine-level spec. The shape that matters:

- **A disjunctive post-condition.** Success and failure are two exits with different guarantees; the spec should say what holds on each, not just that the routine returns. A spec that only characterises the success path leaves the fail-closed behaviour unstated — and fail-closed is the property that matters for a decoder fed attacker-chosen bytes.
- **Which inputs take which exit**, derived from Amsterdam's withdrawal decoding in `execution-specs`. A `Withdrawal` is `(index, validator_index, address, amount)`; the address is 20 bytes and the three scalars are canonical.

Because the block lemmas already exist and are frame-generic, this is largely a matter of getting the composition and the frame threading right — a good first contribution for someone learning the codebase's Hoare-triple conventions without having to invent the block-level reasoning.

## Test coverage today

✅ **Well covered**, better than the other three. Multiple `zisk_*` probes touch this path:

`zisk_block_count_withdrawals`, `zisk_block_withdrawals_total`, `zisk_block_body_extract_withdrawal_count`, `zisk_chain_body_total_withdrawal_count`, `zisk_block_validate_withdrawals_root_one_w`, `zisk_block_validate_withdrawals_root_two_w`, `zisk_block_validate_withdrawals_root_indexed`, `zisk_block_validate_no_withdrawals_pair`

[...]

Issue #10779, opened 2026-07-28
Proof: semantic specs for the 5 RlpWalk routines (correspondence done, specs missing)


pirapira:
Proof work on a **stable leaf**, suitable for an external contributor. No coordination needed: this module is in no active lane's file set.

## Current state, measured on `origin/main`

`EvmAsm/Codegen/Programs/RlpWalk.lean` — **281 lines**, 5 emitted routines:

- `rlp_walk_init`
- `rlp_walk_next`
- `rlp_content_to_u64`
- `rlp_content_to_u256_be`
- `rlp_field0_to_u64`

**What already exists:** 6 theorems, all of the form `rlpWalkInitFunction_eq_verified_prog` etc. These are **string↔Program correspondence** proofs — they establish that the emitted assembly string equals the Lean-rendered `Program`. That is the conversion half.

**What is missing:** the **semantic specifications**. There is no `Fn.Spec` / CPS Hoare triple for any of the five routines. Correspondence tells you the Lean model *is* the guest code; it says nothing about what either one computes.

## The task

Write a spec triple per routine and prove it, against RLP decoding as defined in `execution-specs` (`ethereum/rlp.py` and the Amsterdam decoders that consume these).

Suggested order, cheapest first: `rlp_walk_init` → `rlp_content_to_u64` → `rlp_field0_to_u64` → `rlp_walk_next` → `rlp_content_to_u256_be`.

Points worth pinning explicitly, because live code depends on them:

- **`rlp_content_to_u256_be` rejects non-canonical content and content longer than 32 bytes.** Two live consumers rely on this as their fail-closed path (see #10766). The spec should make both rejections explicit rather than leaving them as an implementation detail.
- `rlp_walk_next`'s cursor advance and its end-of-list signalling — several callers branch on `a1` being nonzero.

## Test coverage today

⚠️ **No dedicated `zisk_*` probe.** Exercised only indirectly, through consumers. This is the weakest-covered of the four leaves filed today, which makes the spec work *more* valuable and also means you cannot lean on a probe to tell you the shape is right.

## Prospect of change

🟢 **Very low.** High fan-in leaf — changing it is expensive (hub-reachability drives the ~1h CI build), so it does not churn. Not in any active lane.

## Conventions any contributor must follow

[...]

PR #10952, opened 2026-07-30
codegen: retire dead JUMPDEST bitmap

pirapira:
## Why

The emitted JUMP and taken-JUMPI handlers already validate targets with a PUSH-aware direct scan over the current frame, so the dispatcher bitmap had no reader. The ELF had exactly one materialization of `evm_jumpdest_bitmap` (`0xbedfafa0`), inside `.jdbm_scan`, followed only by `lbu`/`or`/`sb` writes. No emitted load or control-flow consumer reached that region.

This removes the dead builder, its two data-section reservations, and its 948-line proof module. That proof was correct but proved an unreachable mechanism: a proof establishes code behavior, not that the code runs. The former bitmap was also 16 KiB for 131072 code bytes, 2.7× the 49152-byte maximum initcode input.

The live mechanism is now documented consistently: `EvmControlFlowHandlers` scans from the current `x21` code base and current `env.codeSize`, which works at both top-level and nested frame depth.

## Layout

- `.text`: `0x66c88`
- `.data`: unchanged at `0x5370`; `bnf_le_a` remains `0xa3000ee0`
- `.bss`: `0x1c0eafa0` (down `0x4000`)
- regenerated TSV: `__BSS_END__ = _end = 0xbf1fafa0`; these exact values repin both manual `GuestImage` bounds

## Checks

- `python3 scripts/gen-symbol-addresses.py --build`
- `lake build EvmAsm`
- olean/hash inventory: `234 / 234`
- regenerated ELF has no `evm_jumpdest_bitmap` / `.jdbm` symbol or materialization

Behavior-neutral deletion: no reader switched and no JUMP/JUMPI behavior was changed.


pirapira:
## Review: verified, labelling — with one merge-order hazard that must not be resolved by hand

`+453/-1569` — **a net removal of 1,116 lines**, including a 948-line proof, 16 KiB of `.bss`, and two per-dispatch code scans.

### Arithmetic verified independently

| | |
|---|---|
| `bssSizeBytes` | `0x1c0eefa0` → `0x1c0eafa0`, delta **`0x4000`** = exactly 16 KiB ✓ |
| `GuestImage` bound | `0xa3110000 + 0x1c0eafa0` = **`0xbf1fafa0`** ✓ matches the regenerated TSV `__BSS_END__`/`_end` |
| second bound | `0xa3005370` → `0xbf1fafa0` ✓ consistent |
| `textSizeBytes` | `0x066d68` → `0x066c88`, **−0xE0** = the removed builder |
| `dataSizeBytes` | **absent from the diff** → `.data` stays `0x5370`, `bnf_le_a` stays `0xa3000ee0` ✓ |

Both `GuestImage` bounds derived from the TSV rather than by arithmetic, which is the gate with **no automated check** and where #10939's error occurred. The pin audit found no `rfl`-style pins among the 49 shifted `.bss` symbols.

### ⛔ Merge-order hazard: #10951 and #10952 edit the same line from the same base

Both branch from `textSizeBytes = 0x066d68`:

- **#10951** (`+4`, the creation postlude jump) → `0x066d6c`
- **#10952** (`−0xE0`, the deleted builder) → `0x066c88`
- **both landed must be** → **`0x066c8c`**

**Neither PR's value is correct once both are in**, and the same applies to `symbol-addresses.tsv`, which both regenerate. So:

> **Whichever merges second must be re-generated, not textually resolved.** A hand-picked `textSizeBytes` would be wrong; `check-region-map` would catch that one — but a hand-resolved **`GuestImage` bound typechecks silently**, which is the failure mode with no gate behind it.

If these are batched together, the batch needs a regen before it lands. Flagging because a conflict on two integer literals looks like the most trivial possible textual merge, and it is the one place in this repo where "just take both changes" produces a silently-wrong proof.

### ⭐ The finding inside the deletion is worth more than the deletion

The removed proof's own docstring claimed *"JUMP/JUMPI then test one bit in O(1)."* That is **false** — the live mechanism is a PUSH-aware direct scan over the current frame's `env.codeSize`, established from emitted asm (`h_JUMP` at `0x8004d698` loading `496(s4)`, no reference to the bitmap range).

So this retires **948 lines of correct, `sorry`-free, axiom-clean proof about a structure nothing reads.** Verified surface drops by 948 lines with **no loss of assurance** — worth stating plainly, since a reviewer seeing a proof deleted will reasonably suspect the opposite.

That is a new variant of a failure mode this repo has hit three times today, and the most instrument-proof one:

| | |
|---|---|
| #10942 | proof **true but not covering production** |
| #10779 / #10782 | specs that **existed** while an audit looked in the wrong module |
| **this** | proof **true, fully covering its target — and the target is unreachable** |

No gate here can see it. The axiom check passes, it is `sorry`-free, the drift guard holds, the theorem is genuinely proved. **Only a liveness question finds it, and nobody was asking one.**

The rule worth keeping: **a proof establishes that code does what it says, never that the code runs.** Coverage of a dead mechanism is indistinguishable from coverage of a live one by every instrument in this repo — the sibling of #10942's lesson that proof coverage of a call site is not implied by a theorem about the routine.

Corroboration that this is dead rather than merely unreferenced today: no emitted load from the range; the only materialisation of its base is inside `.jdbm_scan` itself, followed solely by `lbu`/`or`/`sb` writes; and the structure was provisioned for **131072 code bytes when `MAX_INIT_CODE_SIZE` is 49152** — 2.7× the maximum possible input, the signature of something nobody revisited after the direct scan superseded it.

Stale prose corrected in `CODEGEN.md`, `SpecRef/Runtime.lean`, the control-flow handler and program docs, and a test comment — the fourth instance today of documentation asserting something the code does not do, after `.Lcfrc_create_gas_charged`, `DispatchTx:16-19`'s description of `.Lbv_after_tx_gas_precharge`, and #10951's "not yet dispatched by this loop".

PR #10953, opened 2026-07-30
refactor: gather top-level message preparation

pirapira:
## Summary

Gather the existing top-level preparation emission into `emitTopLevelMessageD0Preparation` without changing its emitted instruction sequence. This is Cut A of the creation/message-processor convergence work; it deliberately introduces neither an entry seam nor a depth branch.

The gathered sequence, in its preserved order, contains:

1. EIP-7702 authorization traversal and `set_delegation` callback.
2. The authorization state-gas fold.
3. `prepare_dispatch`'s staged creation/new-account state-gas charge.
4. The preparation-only halt, top-frame gas, and top-frame context setup.
5. Caller-scoped access seeding plus the deferred `prepare_dispatch` callback. This is top-level-only because `utils/message.py:50-53` seeds the top-level builder; nested frames inherit `accessed_addresses.copy()` (`system.py:149-150`, `367-368`), not because `process_message` guards it by depth.
6. The post-preparation marker only. The marker is depth-zero-only because the spec represents preparation failure by returning before the shared body (`interpreter.py:356-378`).

The body snapshot and pending value-transfer staging intentionally remain outside the helper. They correspond to the shared post-preparation path (`interpreter.py:380`, `384-390`) and must remain available to nested frames.

## Verification

- Rebasing onto post-#10952 main `37eb9cca663ae59a2c1791a35f190adef8e587fd` was conflict-free.
- `lake build EvmAsm.Codegen.Dispatch`
- `git diff --check`
- Source-level emitted-operand comparison against that parent: identical 91-operand sequence after normalizing only Lean concatenation/comments (`sha256 6249781a538f5900884b09a05f40411c28e16a30acd351801acaf242b315a1a3`). The comparator selects only string-literal operands and `emitTxAccessListSeedLoop`; comments and Lean concatenation are excluded by construction, while no content inside a string literal is changed.
- Comparator negative control: replacing only the first `runtime_tx_auth_state_charge` operand in memory produces a distinct stream (`sha256 943c96c3906c534739402f95b12cc8b80a30ca86316f804cd60e25d5e844680c`).

No generated artifacts were regenerated: this cut is source-only and byte-identical by construction.

### DIFF
--- a/EvmAsm/Codegen/Dispatch.lean
+++ b/EvmAsm/Codegen/Dispatch.lean
@@ -2312,6 +2312,108 @@ def emitTxAccessListSeedLoop : String :=
     initialised — pointed at the input region instead of an
     in-`.data` label. The hex literal `0x40000010` matches
     `INPUT_ADDR + INPUT_DATA_OFFSET` in `Programs.lean`. -/
+private def emitTopLevelMessageD0Preparation : String :=
+  -- Cut A gathers the top-level preparation portion around `process_message`
+  -- without changing its emitted order. Cut B will guard this exact fragment
+  -- with `evm_call_depth == 0` before a later entry seam allows child frames
+  -- here. Access seeding below is caller-scoped, rather than depth-guarded in
+  -- the spec: nested messages inherit their parent's accessed-address set.
+  -- Do not move individual instructions across this boundary: state-gas and
+  -- early-exit ordering is part of the transaction semantics.
+  -- 1. authorization traversal / set_delegation callback
+  "  la x11, runtime_tx_auth_state_charge; sd x0, 0(x11)\n" ++
+  "  la x11, runtime_tx_auth_exec_fn; ld x9, 0(x11); beqz x9, .runtime_tx_auth_exec_done\n" ++
+  "  addi sp, sp, -56; sd ra, 0(sp); sd x5, 8(sp); sd x6, 16(sp); sd x7, 24(sp); sd x10, 32(sp); sd x20, 40(sp); sd x21, 48(sp)\n" ++
+  "  la x11, runtime_tx_auth_inner_ptr; ld x10, 0(x11); la x11, runtime_tx_auth_inner_len; ld x11, 0(x11); la x12, runtime_tx_auth_sender_ptr; ld x12, 0(x12); la x13, runtime_tx_auth_type; ld x13, 0(x13)\n" ++
+  "  jalr ra, x9, 0; mv x9, x10\n" ++
+  "  ld ra, 0(sp); ld x5, 8(sp); ld x6, 16(sp); ld x7, 24(sp); ld x10, 32(sp); ld x20, 40(sp); ld x21, 48(sp); addi sp, sp, 56\n" ++
+  "  bnez x9, .exit_outofgas\n" ++
+  ".runtime_tx_auth_exec_done:\n" ++
+  -- 2. authorization state-gas fold
+  "  la x11, runtime_tx_auth_state_refund\n" ++
+  "  ld x9, 0(x11)\n" ++
+  "  beqz x9, .runtime_tx_auth_state_refund_done\n" ++
+  "  la x11, runtime_tx_auth_state_charge; sd x9, 0(x11)\n" ++
+  "  la x11, runtime_tx_state_gas_ptr; ld x8, 0(x11); ld x7, 0(x8); add x7, x7, x9; sd x7, 0(x8)\n" ++
+  "  la x11, evm_state_gas_left\n" ++
+  "  ld x8, 0(x11)\n" ++
+  "  bltu x8, x9, .runtime_tx_auth_state_spill\n" ++
+  "  sub x8, x8, x9\n" ++
+  "  sd x8, 0(x11)\n" ++
+  "  j .runtime_tx_auth_state_refund_done\n" ++
+  ".runtime_tx_auth_state_spill:\n" ++
+  "  sub x9, x9, x8\n" ++
+  "  sd x0, 0(x11)\n" ++
+  "  bltu x6, x9, .exit_outofgas\n" ++
+  "  sub x6, x6, x9\n" ++
+  ".runtime_tx_auth_state_refund_done:\n" ++
+  "  la x11, runtime_tx_auth_state_charge; sd x0, 0(x11)\n" ++
+  ".runtime_tx_auth_state_used_done:\n" ++
+  -- 3. prepare_dispatch's staged creation state-gas charge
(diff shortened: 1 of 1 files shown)


pirapira:
## Review: cut A verified as a pure factoring — labelling, with one gate that must be re-run after rebase

`+105/-133` in a single file, **net −28 lines**, no generated files touched. That is the right shape for a gather-only cut.

### Verified it is a factoring and nothing more

- **New emitter defined once, called once**: `emitTopLevelMessageD0Preparation` at the definition, invoked at a single site that keeps its exact position.
- ⭐ **No depth test added.** I checked specifically, because cut A adding the guard would silently absorb cut B and destroy the property that makes this sequence safe. The single `evm_call_depth` occurrence in the diff is a **Lean comment** describing the *future* seam — `-- with \`evm_call_depth == 0\` before a later entry seam allows child frames` — not an emitted test. Cut A stays a no-semantics change.
- **Boundary is the corrected one**: marker included, snapshot and value staging excluded as shared body work, access seeding labelled caller-scoped rather than depth-guarded.

That boundary correction is the substance of this PR even though it does not appear as code. The original fragment ran through `Dispatch:3004` and would have swept in `dispatcher_capture_body_state` and `dispatcher_seed_pending_value_transfer` — which correspond to `interpreter.py:380` and `:384-390`, both sitting at **four-space indentation, the same level as the `if` at `:356`, therefore outside its block.** Guarding those to depth zero would have left a nested frame with **no rollback boundary and no value transfer**, and because cut B is a no-op on every current input, that would have surfaced only after cut C and been attributed to cut C.

### ⛔ The byte-identity gate is base-relative and must be re-run

The sha256 `6249781a…b315a1a3` is computed against **this** parent. Two labelled PRs are ahead of it and one of them touches the same file:

- **#10952 removes 129 lines from `Dispatch.lean`** — the dead jumpdest bitmap builder and both its call sites.
- Cut A factors a span in the same file.

Different regions, so git may well auto-merge — but **the emitted stream changes once the builder is gone, so the recorded hash stops being comparable.** The *claim* (this cut preserves whatever the stream is) survives a rebase; the *evidence* does not.

> After rebasing onto the post-#10951/#10952 main, **recompute the normalised operand hash on the new base and confirm equality against the new parent.** A stale hash carried across a rebase is a baseline quoted against a base it was not captured on — the same error I have been correcting in this repo all night, including twice in myself today.

### On the normaliser, since it is the whole gate

Please state in the body exactly what the normalisation removes: Lean-level concatenation operators and Lean comments, and **nothing inside a string literal.** That last clause is load-bearing — a comment placed *inside* a converted asm block is genuine drift here, and a comment-stripping normaliser is precisely the instrument that would report identity across it. The claim is true for these files (the comments sit between literals in the `++` chain), but it should be written rather than relied on implicitly.

⭐ And if it is cheap: **perturb one operand, confirm the hash changes, discard the perturbation.** Thirty seconds, and it converts a passing check into a validated one — the same move that made #10951's trace decisive after an unvalidated probe cell had already retired two hypotheses wrongly.

### Queue position

Third behind #10951 and #10952. @merge-bot owns both of those sequentially; this one needs the rebase-and-recompute above rather than a textual merge, though unlike #10952 it has no generated artifacts to regenerate — only the hash to re-derive.

Cut B follows: the single depth test around this fragment, named distinctly from the four existing `evm_call_depth` tests at `Dispatch:1203`/`1356`/`1442`/`1464` (dispatch stop and exit routing, not processor boundaries), and stated plainly as **a no-op today** so a reviewer does not ask why an untriggerable guard is being added.

pirapira:
## Re-labelled at `2f6e051b4` — rebased onto `37eb9cca6`, gate re-run on the new base

Verified: head is a descendant of current main, still **one file, `+105/-133`**, and **zero layout literals in the diff** (no `SizeBytes`, no `.bss`/`.data` bound, no `bnf_le_a`) — which is what a source-only cut should look like after a rebase across a 49-symbol address shift.

### ⛔ Correcting my own prediction, because the actual result is more informative than the one I forecast

I said the hash *would not survive* the rebase and that this was expected, since #10952 deleted 129 lines from this same file. **That was wrong.** The normaliser extracts the **91 operands of the gathered fragment**, not the whole file's emitted stream — and #10952's deletion (the bitmap builder and its two call sites) lies **outside** that fragment. So the operand stream is genuinely unchanged and the hash is legitimately identical: `6249781a…`, with the perturbation control again at `943c96c3…`.

⭐ **And an unchanged hash across a base that changed the file is stronger evidence than a changed one would have been.** It says #10952's deletion and cut A's factoring are **disjoint** — which is precisely the property that makes reviewing them independently valid. Had I been right, we would have learned nothing beyond "the file moved."

The discipline still held even though my forecast didn't: **you have to recompute to know.** Carrying the old hash forward would have produced the same number by assumption rather than by measurement, and those are not the same claim even when the digits match.

### What the gate now rests on

- **Byte-identity**: 91-operand normalised stream, equal to parent, `6249781a…`.
- **Instrument validated, not merely passing**: a one-operand perturbation moves the hash to `943c96c3…`. That is the difference between a check that can fail and a check that happens to agree.
- **Normaliser scope**: it *selects* string-literal operands, so Lean comments and concatenation are excluded **by construction** rather than by a stripping pass — meaning there is no rule that could accidentally remove something *inside* a literal. That matters because a comment inside a converted asm block is genuine drift here, and a comment-stripping normaliser is exactly the instrument that would report identity across one.

### The boundary is the substance, and it is confirmed from both sides

The fragment ends at the post-prep **marker**; the `dispatcher_capture_body_state` / `dispatcher_seed_pending_value_transfer` chain stays **outside** in the caller. That correction came from two independent readings agreeing:

- **Spec**: `interpreter.py:380` and `:384-390` sit at four-space indentation — the same level as the `if` at `:356` — therefore **outside** its block.
- **Guest**: a nested frame already gets its value transfer through the shared descend machinery, so guarding it to depth zero would have left every child frame with no transfer.

Guarding those to D0 would have given nested frames **no rollback boundary and no value transfer** — and because cut B is a no-op on every current input, that would have surfaced only after cut C and been attributed to cut C.

Access seeding is labelled **caller-scoped** rather than depth-guarded — top-level-only because it lives in the message builder (`utils/message.py:50-53`) and nested frames receive a copy of the parent's set, not because the spec guards it by depth. Harmless to include, but the distinction matters if the guard is ever moved.

Cut B follows: the single `evm_call_depth == 0` test around this fragment, named distinctly from the four existing depth tests (which are dispatch stop/exit routing, not processor boundaries), with the enumeration of what reads the fragment's outputs after it — and stated plainly as **a no-op today**.

Issue #10948, opened 2026-07-30
Nested CREATE/CREATE2 address collision debits the creator's endowment with no recipient: value destroyed on an arm where the spec transfers nothing


pirapira:
## Claim

**On a nested `CREATE`/`CREATE2` address collision, the guest debits the creator's balance by the endowment. The spec debits nothing — and nothing is credited, so the value is destroyed.**

Found by `k3` in emitted asm at `d7f667ead`, while resolving an unprompted observation rather than the question it was asked. Source-level; **direction not measured**.

## Spec (pinned `e5a8caf1b`)

`account_deployable` is checked at `system.py:117`, **before the child message is built**. The endowment transfer lives in `move_ether` inside `process_message` (`interpreter.py:384-390`), which a collision never reaches.

So on a nested collision: **the creator's balance is unchanged**, the target account is never created, and the recoverable consequence is exactly — creator nonce incremented (`:118`), `NEW_ACCOUNT` refunded if charged (`:123`), push zero, frame continues.

## Guest

Inside `.Lcr_collision_f0` (`:56913-57010`):

1. copy the creator's current balance into `create_creator_newbal`;
2. `u256_sub_be(create_creator_newbal, create_value_be, create_creator_newbal)` — **subtract the endowment**;
3. if the subtraction borrows, skip the write; otherwise write `current − endowment` back into the creator's account balance field (`.Lcr_sbwb_f0`, `:57005-57008`);
4. push zero (`:57009`), advance pc (`:57010`).

⭐ **The benign reading is refuted, and refuted by the arithmetic rather than by absence.** A *restore* of an earlier debit would copy the original balance back **without subtracting**. This subtracts. It is a fresh debit on the collision arm.

⛔ **And no compensating credit exists on that path** — the created account never comes into existence, so there is no recipient. The debited value goes nowhere. This is a **value-conservation violation**, not a misattribution: total ether decreases.

The underflow guard protects only the insufficient-balance case. It does not protect the rule being broken, which is *a collision must not debit at all*. A guard that fires on the wrong predicate reads as safety in review while leaving the actual defect fully live.

## Why the arithmetic matters more than the site

[...]

PR #10954, opened 2026-07-30
refactor: guard top-level message preparation by depth

pirapira:
## Summary

Add the single runtime depth test corresponding to `interpreter.py:356`: only depth zero executes `emitTopLevelMessageD0Preparation`; every frame proceeds to the shared post-preparation body.

This is Cut B, stacked on #10953. Today no nested frame reaches this boundary, so the nonzero arm is deliberately inert. It makes the D0/BOTH distinction explicit before Cut C routes child frames into the shared body.

The branch uses `x28` because `t0`/`x5` remains live as the header cursor inside the D0 helper. On the D0 path `x28` is already left as zero or the deferred callback pointer; on the new skip path it retains the depth. There is no canonical `x28` contract for downstream code, and the shared bootstrap does not read it before opcode dispatch.

`runtime_tx_post_preparation_reached` remains inside the guarded fragment. Its direct runtime consumers are transaction-level finalizers after the outer dispatcher returns; a future nested path must leave the enclosing top-level marker intact.

## Verification

- `git diff --check`
- `lake build EvmAsm.Codegen.Dispatch`
- Source/control-flow re-audit on post-#10952 main `37eb9cca663ae59a2c1791a35f190adef8e587fd`:
  - `x28` is not read between the new join and dispatch bootstrap;
  - the shared body starts after the D0 marker;
  - the marker's runtime readers remain in `BlockVerdictMtxRuntime` and `TxIntrinsicStateGas`, after outer-dispatch return.

No generated artifacts are included. Cut C will supply the first nested caller, so this cut is structurally behaviour-neutral on current traffic.


pirapira:
## Review: cut B verified, labelling. Stacked on #10953.

`+11/-0`, one file, **three instructions** and a comment block. Nothing else.

### The bracket is right, verified in the diff rather than from the description

```
".runtime_tx_gas_no_reservoir:\n" ++
  la x28, evm_call_depth ; ld x28, 0(x28) ; bnez x28, .runtime_tx_top_level_message_d0_done
emitTopLevelMessageD0Preparation ++                    <- unchanged, now bracketed
".runtime_tx_top_level_message_d0_done:\n" ++
-- Snapshot and pending-value staging are the shared post-preparation body
  … dispatcher_seed_pending_upfront_sender_balance ; dispatcher_capture_body_state ; dispatcher_seed_pending_value_transfer
```

**The fragment is inside the guard and the capture/transfer chain is outside it.** That is the corrected boundary — `interpreter.py:380` and `:384-390` sit at four-space indentation, the same level as the `if` at `:356`, therefore outside its block; and k3 confirmed independently from the guest side that a nested frame already gets its transfer through the shared descend machinery. Guarding those to depth zero would have left every child frame with **no rollback boundary and no value transfer**.

### ⭐ It has its own label, which is the point

`.runtime_tx_top_level_message_d0_done` is new and self-describing, distinct from the four pre-existing `evm_call_depth` tests (`Dispatch:1203`/`1356`/`1442`/`1464`), which are dispatch stop/exit routing rather than processor boundaries.

That matters more than naming hygiene. **Four defects today came from semantics expressed as source adjacency**: #10950/#10951 (success fell into the conservative-bail arm through a missing successor edge), the resume seam (any entry at `.runtime_tx_post_top_frame_done` sets a D0-only marker because the write sits immediately below it), `.Lbv_after_tx_gas_precharge` (bail and success converge, so nothing downstream can distinguish them), and #10948 (the collision arm reaches a debit belonging to the non-collision path). A fresh label whose meaning does not depend on its neighbour is the direct remedy.

### The comment block carries what a reviewer needs

It cites `interpreter.py:356`, states the **no-op-today** premise explicitly, and gives the `x28` reasoning in the stronger form — not "no read found" but **there is no canonical `x28` contract to break**, since the depth-zero path already leaves either zero or the callback pointer. An argument from the absence of a contract survives future edits; an empty grep does not.

### Why this lands before cut C rather than with it

The guard must exist **before** the seam that makes nested entry possible. Built the other way round there is an interval — one merged PR wide — in which child frames reach the D0 work unguarded, and that region re-applies authorizations per frame, double-charges both `prepare_dispatch` NEW_ACCOUNT charges and the delegation access charge, and sets `disable_precompiles`. Landing the guard first makes that interval empty by construction; landing it second makes correctness depend on nothing merging in between, which is not enforceable.

So a guard with no currently-reachable false branch is the **second** cut rather than the last, and the body says so — otherwise a reviewer reasonably asks why an untriggerable branch is being added.

### Note for the merge path

Stacked on `refactor/gather-top-level-message-d0`, so #10953 merges first and this retargets to main. Both carry `merge!`. Source-only, no generated files, so `.text` moves and a repin is expected — per PR, not combined.

PR #10955, opened 2026-07-30
refactor(codegen): retire unwired dispatcher resume seam

pirapira:
## Summary

Retire the unwired `runtime_dispatcher_resume` entry and its
`runtime_tx_resume_code_ptr` backing cell.  The live
`runtime_dispatcher_prepare_only` entry remains unchanged: it is called by the
block-verdict deferred-recipient path and uses `runtime_tx_prepare_prefix_status`
to distinguish the status-2 preparation rejection.

This is intentionally a separate retirement from the top-level D0 extraction
and depth guard (#10953/#10954), stacked on #10954 to avoid hand-merging three
concurrent edits to `Dispatch.lean`.

## Why this is dead

- The introducing commit `84cf0978` explicitly described resume as an unwired
  save-point facility: its purpose was to avoid replaying EIP-7702 preparation,
  but no MTx caller was connected.  `git log -S` finds no later wiring of either
  the entry label or its pointer cell.
- On the #10952 linked artifact (`7528e2565`, an ancestor of merged main
  `37eb9cca6` whose `Dispatch.lean` is byte-identical to that merge),
  `runtime_dispatcher_resume` was at `0x800392d8`.  A complete disassembly has
  that address only at its definition; its `runtime_tx_resume_code_ptr`
  materialization occurs only inside the same dead block.  There are no ELF
  relocations.  In contrast, the same artifact has the block-verdict `jal` to
  live `runtime_dispatcher_prepare_only`.
- The newly emitted candidate has neither retired symbol, while it retains the
  live `jal` to `runtime_dispatcher_prepare_only`.

This belongs to the dead-mechanism category, not cosmetic cleanup: today also
found the dead jumpdest bitmap and a 948-line proof layered on it.  None of the
current gates asks whether an emitted facility is ever reached, so these were
found incidentally while mapping processor boundaries.

## Layout and verification

- Regenerated ELF: `.text = 0x66c68`, `.data = 0x5370`, `.bss = 0x1c0eafa0`;
  `__BSS_END__ = _end = 0xbf1fafa0` from the regenerated TSV.
- `bnf_le_a` remains `0xa3000ee0`; `.data` and `.bss` are unchanged, so
  `GuestImage.lean`'s three `.bss` literals remain the TSV-derived values.
- `LAKE_ARTIFACT_CACHE=false lake build EvmAsm.Codegen.Dispatch`
- `NO_BUILD=1 ELF_DIR=gen-out/regionmap scripts/check-region-map.sh`
- `git diff --check`

I also attempted a full cache-disabled `lake build`, but stopped its unrelated
parallel package-build race after `Mathlib/Tactic/Ring/Compare.olean` vanished
while another target was reading it.  That run is not claimed as a passing gate.


pirapira:
## Review: verified, labelling. Stacked on #10954.

Deletes exactly two symbols — `runtime_dispatcher_resume` and `runtime_tx_resume_code_ptr` — and nothing else.

### ⭐ The one thing worth checking here, checked

`runtime_dispatcher_prepare_only_return` appears in the TSV diff as a **removed** row, which reads alarming next to a PR that deletes a seam: `prepare_only` is **live** (`BlockVerdictDispatchTx.lean:1106` jals it for the status-2 path), and `Dispatch.lean:2831` jumps to `_return` from **outside** the resume block. A deleted label with a surviving jump to it is an undefined-label link failure; a deleted label whose *only* live jump was the one being removed is fine.

It is neither — **the symbol moved, it was not removed.** Both halves are present:

```
- runtime_dispatcher_prepare_only_return  0x80039310  .text
+ runtime_dispatcher_prepare_only_return  0x800392dc  .text
```

Confirmed on the PR head: the label definition is still present, and the `:2831` jump still resolves to it. Only the resume-internal `beqz x21, …_return` went, along with resume itself. The genuinely deleted symbols show a `-` with **no** `+`:

```
- runtime_dispatcher_resume       0x800392dc  .text
- runtime_tx_resume_code_ptr      0xbcd9cff0  .bss
```

⛔ Worth recording as a review technique, because a symbol-address TSV makes deletions and relocations look identical at a glance: **grep both halves.** A `-`-only match is a deletion; a `-`/`+` pair is a move. I nearly read a relocation as a broken link.

Note also that `prepare_only_return` inherits resume's old address `0x800392dc` — which is what you would expect when the routine immediately above it is removed, and is incidental corroboration that the deletion is contiguous and complete.

### The retirement itself

- **Historical evidence is the strongest part**, and it is better than any inference of deadness: the introducing commit `84cf0978` states the resume entry bypasses setup so EIP-7702 authorization effects cannot be replayed, and that **no MTx caller is wired to the facility**. `git log -S` finds nothing wiring it since. The author documented its own absence — unlike the jumpdest bitmap, whose docstring *claimed* a consumer that did not exist.
- **Liveness re-established on the post-#10952 image** rather than carried forward, which matters because a 49-symbol address shift can invalidate an address-based unreachability argument.
- **`prepare_only` is correctly preserved**, and the distinction is worth stating for reviewers: the pair was split and only one side survived, with a different job — `prepare_only` is wired as the **status-2 rejection** path, while `resume`, the half that would have made it a genuine deferral, was never built.
- A dead routine that **writes a D0-only marker** is a landmine rather than dead weight: had anyone wired it for a child path, it would have silently asserted preparation-completed on behalf of a frame that never prepared. Retiring beats adapting, since an explicit seam written for cut C is auditable while an inherited one whose purpose was inferred is not.

### Layout

`textSizeBytes 0x066c8c → 0x066c68` (−0x24), and `.bss` loses `runtime_tx_resume_code_ptr`. One layout literal in the diff, expected — both sections move. Per the standing rule: **re-derive the `GuestImage` bounds from the fresh TSV even where git reports no conflict**, since on the #10951/#10952 merge that file did not conflict and git silently took one side.

### Category, not cleanup

Two dead mechanisms plus a 948-line proof of one, all found today, **none by a gate** — the bitmap while mapping the processor boundary, resume while hunting cut C's seam, the proof because it sat on top of the bitmap. No instrument here asks whether code *runs*, so this class surfaces only incidentally. That belongs in the record as a gap in the gate set rather than as three tidy-ups.

PR #10967, opened 2026-07-30
fix: enforce CALL depth limit on no-descent paths

pirapira:
## Spec alignment

`generic_call` checks `e.message.depth + 1 > STACK_DEPTH_LIMIT` before callee execution (`EvmAsm/Stateless/SpecRef/Interpreter.lean:657-664`). The emitted guest applied that normal CALL failure only in `callDescendFallThrough`; local no-descent paths could execute a precompile or return success at depth 1024.

This adds a self-contained normal-failure gate: clear the precompile frame, pop the CALL operands, push zero, advance the parent PC, and resume. It deliberately does not reuse stack/static guards because those are exceptional halts, whereas a depth limit is a normal CALL failure.

## Complete no-descent enumeration

Every path that executes a callee without `call_frame_descend` is depth-gated:

1. Ordinary supported precompiles: `.L<tag>_supported_precompile` gates the common precompile completion funnel.
2. EIP-4788 CALL/STATICCALL specialization: `.L<tag>_eip4788_addr_match` gates before both current-success and same-slot-stale return arms, and before either can take its special child-allotment charge.
3. Empty-code/EOA success is a non-site: all status-2 routes join `.Lcd_empty_<tag>` only after the existing depth gate in `callDescendFallThrough`, so it already pushes zero at depth 1024 before resolution.

There are no independent 2935/history, beacon-root, withdrawal, or system-call fast paths; `callPushZeroFallThrough` belongs only to probe/tiny registries, not the shipped call-frame registry.

At each new site, the gate remains after access and memory charges the spec keeps and before forwarded allotment/reservoir work the depth failure never reaches.

## Validation

- Isolated source build of the affected child-frame modules passed.
- Isolated stateless guest re-emission assembled and linked; six unique gate labels are present (CALL/STATICCALL ordinary plus EIP-4788, CALLCODE/DELEGATECALL ordinary). This caught and then eliminated an initial duplicate-label emission failure.
- Regenerated linker facts, GuestAddrs, and RegionMap to a fixed point.
- `scripts/check-asm-to-program.sh`, `scripts/check-region-map.sh`, and `LAKE_ARTIFACT_CACHE=false scripts/check-build-parallel.sh` passed.
- `.text = 0x66d24`; `.data = 0x5370`, `.bss = 0x1c105000`, and `GuestAddrs.bnf_le_a = 0xa3000ee0` remain as required.

## Layout accounting

The six independently reachable gate sites add `0x1b0` (432) bytes of `.text`: 72 bytes per site, comprising 18 physical RV64 instructions. Each site has the two-instruction `la` expansions for the depth word, precompile-frame state, and continuation label, plus the depth comparison, frame clear, 192/160-byte stack pop, zero result word, PC increment, and resume. `.data` and `.bss` are unchanged.

## Dynamic-wiring status

The checked-in ECRECOVER CALL/STATICCALL regression cases and the standalone dispatcher ECRECOVER probe initialize `evm_call_depth` to zero; no existing case reaches the new `depth = 1024` precompile gate. The emitted six-site inspection and source/link gates establish the sites are wired, but a dedicated depth-1024 precompile probe is still needed to demonstrate the taken branch dynamically.

## Inherited normal-failure tail

`precompileDepthGateAsm` deliberately reuses the exact emitted normal-failure sequence of the shipped `precompileValueBalanceGateAsm` tail: it clears `evm_precompile_frame`, applies that handler's existing `netPopBytes`, writes the zero result word, advances the parent PC, and resumes through `dispatchContinueRet`. The only new mechanism is the depth predicate. This inherits the established stack and continuation shape rather than introducing a second early-exit ABI.

## Evidence boundary and follow-up

The direct depth-1024 probes exercise the shared `precompileDepthGateAsm` ABI only: depth comparison, precompile-frame clear, net pop, zero result, and PC advance. They do **not** traverse the production classifier or the EIP-4788 address-match placement. Conversely, the emitted-order trace establishes all six placements (after initial access and memory charging, before sparse-output, value, state-gas, and forwarded-allotment work), but is not a production-path run. This trace previously caught the initially omitted EIP-4788 placement, so it is evidence from the instrument that found a real placement defect—not a substitute for a runtime trace.

No test traverses the production placement at depth 1024. [#10968](https://github.com/Verified-zkEVM/evm-asm/issues/10968) tracks the required depth-seeded full-dispatcher probe with real CALL bytecode; its primary case is EIP-4788 because that was the originally omitted site, with ordinary target `0x01` as its sibling.

## Behavioural checks

- Direct tail-ABI probes: both `zisk_call_depth_precompile_limit` (ordinary supported-precompile stem) and `zisk_call_depth_eip4788_limit` (EIP-4788 stem) produced `[depth=1024, result=0, stack delta=192, parent PC=0x501]`.
- Collateral sample: 200 rows, the first half of the preserved uniformly-random seed-7702 400-row manifest, with FA=0, FR=96, and 104 oracle/guest agreements.
- Pinned controls: expected-invalid 04778 still rejects (`succ=0`); EIP-4788 CALL 05813 remains accepted (`succ=1`).

All sampled execution used immutable candidate ELF `c516ae6779f90cd2210e98233ada857fc4b4b2d6761d712bfb587b210fcf9dba`, frozen Spike runner `c6ac09fa72c2103b6d3e390b779b022ef7f220e5a39bb5f667114a89d78e5868`, and 12 workers. The full converted manifest path is `/var/tmp/fc668/manifest.tsv`; the sample manifest is `/var/tmp/evm-asm/random-block-uniform-runs/seed7702/manifest.tsv`.

The two new direct probes also exactly match the existing non-precompile `zisk_call_depth_limit` contract: `[1024, 0, 192, 0x501]`. This corroborates the spec property that depth failure is callee-kind-independent: `generic_call` returns before callee classification, so an ordinary target, supported precompile, and EIP-4788 target all leave the same parent failure ABI. It still does not prove the production classifier or EIP-4788 address-match reaches the new gate; #10968 tracks that placement-faithful test.


pirapira:
## Reviewer note on the safety leg — what it establishes, and what it does not

Labelling this `merge!`. One precision so the sample figures are not misread later, since a number outlives the message that qualified it.

**What the leg establishes:** `FA=0`, and both pinned controls hold — expected-invalid `04778` remains `succ=0`, and **EIP-4788 CALL `05813` remains `succ=1`**. That second control is the load-bearing one for this change: the 4788 gate was the site missing from the first patch, so a *spurious* firing there was the specific risk, and a 4788 call still succeeding at depth 0 refutes it directly.

⛔ **What it does not establish:** `FR=96 / OK=104` is a **candidate-only** figure with no baseline leg on the same 200 rows. **It is not evidence that nothing moved** — a one-fixture FR shift would be invisible to it. The PR body correctly reports these as raw candidate numbers and claims no A/B; this note exists because a bare FR count invites a comparison reading it does not support.

That gap is accepted deliberately rather than overlooked: the prescribed instrument here is **sample plus pinned controls**, not full-manifest A/B, and FR churn is tolerated in both directions while FA is not.

## Why the correctness argument does not rest on the corpus at all

No fixture in the corpus can reach either gate — **every precompile fixture initialises `evm_call_depth` to zero**. So the corpus is structurally incapable of exercising this change, which is why the evidence is layered instead:

| evidence | what it proves |
|---|---|
| three tail-ABI probes agreeing on `[1024, 0, 192, 0x501]` | the failure ABI is **callee-kind-independent** — matching the pre-existing non-precompile probe exactly, which is what `generic_call`'s return at `system.py:338` *before* classification implies |
| emitted-order trace at all six sites | placement: after the access and memory-expansion charges the spec **keeps** (`:463`, `:476`), before the allotment and reservoir it **returns** (`:338`) |
| shape identity with `precompileValueBalanceGateAsm`'s failure tail | the gate is an **existing shipped early-exit sequence with a different predicate**, so `netPopBytes` and the register/stack state are inherited rather than re-derived |
| `.text` `+0x1b0`, base-independent across two bases | six sites × 72 bytes × 18 instructions — arithmetic closes rather than merely sounding plausible |

⭐ The placement evidence deserves particular weight because **the emitted-order trace is the instrument that found the EIP-4788 bypass** in the first version of this patch. It is not an assertion about placement; it is the reading that caught a real omission.

⛔ **And the honest remaining gap, tracked in #10968 rather than left in this description:** nothing traverses the *production* placement at depth 1024. A placement-faithful test needs a depth-seeded full runtime-dispatcher unit with real CALL bytecode targeting `0x01` and the 4788 address. A gate that is present but never reached is indistinguishable from no fix, so that follow-up matters — with the 4788 variant primary, since that site has a demonstrated history of being overlooked.

Issue #10940, opened 2026-07-30
convergence: a reverted child frame un-creates accounts the spec keeps created — created_accounts is tx-scoped and shared, but the guest's frame checkpoint restores it


pirapira:
## Claim

A reverted child frame **un-creates accounts that `copy_tx_state`/`restore_tx_state` keep created**. `created_accounts` is tx-scoped and *shared* in the spec; the guest's frame checkpoint saves and restores it.

Found by `scoper` while reading the guest arenas against the Amsterdam state tracker for #10937's change two. Source reading; **not** witnessed by a fixture yet.

## Spec basis (pinned `e5a8caf1b`, Amsterdam)

`vm/state_tracker.py`:

- `:774-780` — `copy_tx_state`'s docstring states the parent reference, **`created_accounts`**, `storage_reads`, `account_reads` and `code_reads` are *shared* and not rolled back.
- `:809-826` — `restore_tx_state` restores exactly **four** fields: `account_writes`, `storage_writes`, `code_writes`, `transient_storage`. `created_accounts` is not among them.

So in the spec, an address created inside a frame that later reverts is **still** a member of `created_accounts` for the remainder of the transaction. That is load-bearing for EIP-6780, whose deletion permission is exactly "created in this transaction".

## Guest

- `Dispatch.lean:1290` zeroes `account_state_created_count` with the guest's own comment — *EIP-6780 `created_accounts` is tx scoped*. The guest **states the correct scope in its own text**.
- Inserted into from the CREATE paths; read back at `CreateCodeEffectLog.lean:541`.
- ⛔ But the frame checkpoint captures it at `CallFrameDescend.lean:543` and restores it at `CallFrameReturn.lean:198` — so the effective scope is **frame**, not transaction, contradicting the comment two files away.

## Why this is not #10619 and not #10784

- **#10619** frames the lifetime rule as *reads accumulate, writes are snapshot-restored*. `created_accounts` is neither a read nor a write set, so it falls outside that framing while obeying the same column. This is a third member of the shared column, and it needs naming or a lifetime fix will restore it by analogy to writes.
- **#10784** is the mirror direction at tx scope: `account_state_delete_count` *surviving* into commit on a failed transaction. This one is a shared quantity being *destroyed* at frame scope. Same field family, opposite error, different boundary.

## Fix shape

[...]

PR #10969, opened 2026-07-30
refactor(codegen): expose the shared message-body seam and factor the body-state snapshot emitters

pirapira:
## Summary\n\nExpose an explicitly named child-entry routing seam immediately after the depth-zero-only up-front sender gas seed and before the shared root snapshot/value staging.\n\nThe root path explicitly jumps to the shared body. The child entry is defined but deliberately unreached: no child route is migrated in this PR.\n\nThis PR also factors the existing body-state snapshot emission into source-level scalar/cursor emitters. It does not add an emitted helper call, change a checkpoint destination, or move a rollback boundary.\n\n## Why these boundaries\n\nprocess_transaction materializes the transaction-wide gas debit before process_message; retaining dispatcher_seed_pending_upfront_sender_balance above the child entry prevents a future child frame from re-debiting the transaction gas.\n\nThe snapshot extraction deliberately retains the current child protocol. CALL and CALLCODE publish their value effects before descent and arm cd_nse_presnap; call_frame_descend consumes that one-shot into the child rollback snapshot. CREATE clears the arm and publishes after its snapshot. Moving CALL/CALLCODE publication post-snapshot is a distinct semantic cut tracked in #10970, with a reverting value-CALL control as its acceptance criterion.\n\n## Snapshot extraction\n\nOne structure, one shared source primitive, two emitted mechanisms.\n\nThe canonical per-depth slab remains the shared snapshot structure. The new source-level emitters factor only identical scalar capture and three-cursor capture instructions, parameterized by source and destination registers. Root dispatcher_capture_body_state and the child snapshot intentionally remain distinct emitted mechanisms: their field sets, source environments, ordering, and the child presnap branch differ. A full mechanism unification would therefore be a separate semantic/design question, not this byte-neutral extraction.\n\nAll record timings and the full cd_nse_presnap lifecycle remain unchanged: CALL/CALLCODE are the two arming writers; CREATE, DELEGATECALL, and STATICCALL clear only; call_frame_descend remains the sole consumer. created_count remains outside this work (#10940); storage reads remain non-rollback evidence; precompile/no-descend paths are out of scope.\n\n## Evidence\n\n- Seam: measured ELF text 0x66d24 -> 0x66d28 (+4 bytes); the source hunk emits one jump to runtime_tx_shared_message_body. The child entry has exactly one emitted occurrence (its definition) and no inbound reference.\n- Snapshot extraction: saved pre-extraction and rebuilt stateless_guest ELF SHA-256 are both 9547e6d122c35db14470ccc21d43cd457e3c4ef0acb5d3ed8a42d64bf08e9489; cmp returned 0.\n- Commit 2fca52bed touches only BodyStateSnapshot.lean, BlockVerdictDispatchTx.lean, and CallFrameDescend.lean—no generated files. A changed guest emission would necessarily have changed GuestAddrs.lean and symbol-addresses.tsv, so their absence is structural byte-neutrality evidence in addition to the ELF comparison.\n- The combined PR leaves .text 0x66d28, .data 0x5370, and .bss 0x1c105000 unchanged after the seam's +4. GuestAddrs.bnf_le_a remains 0xa3000ee0.\n- Generated-monolith lifecycle census: five call_frame_descend callers; only CALL and CALLCODE can arm cd_nse_presnap, both with count/overflow writers; CREATE, DELEGATECALL, and STATICCALL only clear it. There is one consumer, call_frame_descend's canonical snapshot.\n- Focused build of the new snapshot emitter plus both callers, check-asm-to-program.sh, and check-region-map.sh completed successfully.\n\n## Falsification history\n\nAn earlier +20-byte seam was rejected: it reentered the dispatcher wrapper and duplicated a stack frame. A procedure prologue reached by jump rather than call also creates an unbalanced stack. This seam is inert within the existing root frame.\n\n## Gates\n\nNo corpus run is required: the only newly emitted seam is unreachable, and the source-only snapshot extraction is byte-identical.\n\ncheck-build-parallel.sh is deferred to CI. The local driver previously raced concurrent Lake sub-builds on shared ilean outputs; no local full-build claim is made.


pirapira:
## Reviewer note: a stronger proof of the extraction's byte-identity than the SHA match

Labelling this. The body cites matching pre/post ELF SHA-256 with `cmp` returning 0, which is sound. There is a **structural** proof available that is checkable in one glance and does not require trusting either build:

**Commit `2fca52bed` touches three files, all Lean source — `BodyStateSnapshot.lean` (new), `BlockVerdictDispatchTx.lean`, `CallFrameDescend.lean` — and no generated files whatsoever.**

Had the emission changed by even one byte, `EvmAsm/Codegen/GuestAddrs.lean` and `scripts/asm-fixtures/symbol-addresses.tsv` would necessarily have changed with it, because every symbol downstream of the edit shifts. Their **absence from the diff** is therefore not a claim about the build — it is a property of the diff. Compare the seam commits, which do carry that churn for a 4-byte insertion (symbols moving `0x8003a128 → 0x8003a12c` and so on).

⭐ Same shape as the argument that carried #10963: a closed statement about the *difference* beats agreement between two artefacts. A dual-SHA match proves two builds agree with each other; the absent generated-file churn proves the change is the change described.

## Scope, restated because the title previously covered only the seam

Retitled to name both changes. Three commits, all behaviour-neutral, combined emitted delta exactly **+4 bytes and one jump**:

| commit | emitted effect |
|---|---|
| seam source + regen | one `j` to the shared body, two labels at one address, `+4` |
| snapshot emitter factoring | **zero** |

Separate commits preserve bisectability, which is what the one-change-per-commit discipline protects — the rule was never one-change-per-PR.

⭐ The scope statement I want preserved verbatim from the body, because it corrects a framing I had wrong earlier and a reader will otherwise assume the stronger claim: **one structure, one shared source primitive, two distinct emitted mechanisms.** The root's `dispatcher_capture_body_state` and the child snapshot stay separate because their field sets and ordering genuinely differ; only the identical scalar and cursor capture instructions are shared. This is *not* a mechanism unification, and unifying them may not be worth doing at all — two mechanisms over one canonical structure with disjoint field sets is arguably the honest shape.

## What is deliberately unchanged, and what would prove it

Both publication timings are retained: CALL and CALLCODE publish before descent and arm `cd_nse_presnap`; CREATE clears the arm and publishes after its snapshot. The protocol's full lifecycle is preserved — five `call_frame_descend` callers, two writers that can arm, one consumer.

⛔ Moving that boundary is #10970, and its acceptance criterion is a **reverting value-CALL control** proving the transfer is undone. That is a named test rather than a sweep, and a retirement without it would be exactly the correct-by-coincidence retirement this refactor sequence exists to avoid.

## Note on the review trail

An unreviewed commit reached this branch while it carried `merge!`, moving the head from `55586d7fc` to `2fca52bed`. The label was removed before any merge, so nothing unreviewed could land. Recorded because the failure mode is worth naming: the risk was not the commit — which passes — but that **an approval pointed at a head that no longer existed.** A labelled PR is bot-owned; changes require the label removed first.

Issue #10971, opened 2026-07-30
Two account_state_delete_count guards are sized by accountStateCreatedCapacity (one by name, one as a bare 8192) — correct only while the two capacities are equal


pirapira:
## Two `account_state_delete_count` bounds are sized by the **created** arena's capacity

Not a live defect — the two capacities are equal today, so both guards are correct. It is a **correct-by-coincidence** dependency with no assertion behind it: resize the created set alone and both delete-count bounds silently become wrong.

Found while verifying a related claim on `origin/main`. Three sites, and the third is the one that makes this worth filing rather than noting:

| site | bounds | with |
|---|---|---|
| `CreateCodeEffectLog.lean:120` | — | `def accountStateCreatedCapacity : Nat := 8192` |
| `CreateCodeEffectLog.lean:511` | `account_state_created_count` | `accountStateCreatedCapacity` — **correct usage** |
| `CreateCodeEffectLog.lean:249` | **`account_state_delete_count`** | `accountStateCreatedCapacity` — wrong constant, right value |
| `BlockVerdictCreationStage.lean:547` | **`account_state_delete_count`** | bare literal `8192` — no named source at all |

So the delete arena's bound is expressed **two different ways, neither of which names the delete arena**: once by borrowing the created arena's constant, once as a magic number. For contrast, the pending and durable arenas use `accountStateEntryCapacity` (38460), so the codebase does distinguish capacities where they differ.

## Why it is currently harmless

Every insert into `account_state_delete` passes 8192 as its cap (`NoopHalt.lean:575-576`, `CreateCodeEffectLog.lean:851`). The insert refuses at capacity, the caller routes to the overflow path, overflow sets `account_state_overflow`, and both of its `bnez` consumers (`BlockVerdictMtxRuntime.lean:508`, `BlockVerdictFunction.lean:865`) reject the block via `.Lbv_fixed_arena_overflow_fail` → `BV_FAIL_CODE 58`, verdict zero. So the count can *reach* 8192 and never exceed it, and both guards test `bgtu` — strictly greater — which can never be true.

⭐ That is the reachability argument that made #10964's capacity-bypass question vacuous. This issue is the residual it left behind: **the guards are unreachable, and they are also sized by the wrong constant.** Those are independent facts, and only the first is currently load-bearing.

## Fix

[...]

Issue #10964, opened 2026-07-30
Should a CHILD frame restore account_state_overflow on failure? A sticky overflow flag surviving a reverted child would be a false reject


pirapira:
## Question

`account_state_overflow` is field 11 of the canonical per-depth snapshot record introduced by #10963 (#10957), **but only the root writes and restores it.** The child path preserves its pre-migration behaviour, in which no child checkpoint for that counter existed, so the slot is root-only and #10963 documents the asymmetry rather than changing it.

That was deliberate and correct for #10963: adding a child restore would have converted a representation-only change into a behavioural one, which is the same reason that PR leaves the `NoopHalt` REVERT tail alone. **This issue is the behavioural question it deferred.**

## Why it may be a false reject

If `account_state_overflow` is a **sticky** flag that drives a sound reject, then a child frame that overflows the AccountState arena and **subsequently reverts** leaves the flag set on state that was rolled back. The block is then rejected on the basis of work that never took effect — a **false reject**, which we do not treat as the safe direction.

⚠️ Stated as a question, not a finding. Two things must be established before this is a defect:

1. **Is the counter sticky, and does it drive a reject?** Read the *consumer* — what reads `account_state_overflow` and what it does on a nonzero value. Per the gate's rule on #10938, a producer-side search is not sufficient evidence here, and note that ~565 routines in this tree are emitted from `Program` values whose instructions are structured constructors, so **a text grep of the Lean sources cannot see inside them** — this must be checked against emitted asm or the `Program` values directly.
2. **Can a child frame actually set it and then fail?** If overflow is only reachable at a point after which the frame cannot revert, the question is vacuous on the reachable population — a legitimate disposition, but one that has to carry its reachability argument spelled out.

## If it is real

The fix is a child checkpoint plus restore for field 11, which the #10963 layout already has a slot for — the record is 13 fields wide at every depth, so this is a wiring change rather than a layout change. That is the payoff of having kept the field in the ABI instead of dropping it to twelve for the child.

## If it is not

Then the root-only asymmetry should be documented as **intentional and permanent** at the descend site, with the reachability argument, rather than left reading as an unfinished migration. An asymmetry that looks accidental is the next reader's trap.

Related: #10963 (introduces the slab and the asymmetry), #10957, #10938 (the depth-parity gate; this is the same BOTH-vs-D0 question the gate exists to ask).

pirapira:
## A second consumer of this question, found while verifying an unrelated citation

[...]

PR #10972, opened 2026-07-30
refactor(codegen): name the delete arena's capacity (GH #10971)

pirapira:
Names the `account_state_delete` arena's capacity, which was expressed **six ways across three files**. Closes GH #10971.

## The problem

Four of the six sites were the bare literal `8192`; two used `accountStateCreatedCapacity`. So the guard on `account_state_delete_count` was correct **only by coincidence of two capacities being the same number** — and the four bare literals would not appear in a search for the created capacity's name, so whoever resized the created set would never see them.

| site (at `0320615e2`) | was | now |
|---|---|---|
| `StorageReadLog.lean:263` — guard | bare `8192` | `accountStateDeleteCapacity` |
| `BlockVerdictCreationStage.lean:541` — guard | bare `8192` | `accountStateDeleteCapacity` |
| `CreateCodeEffectLog.lean:249` — guard | `accountStateCreatedCapacity` | `accountStateDeleteCapacity` |
| `CreateCodeEffectLog.lean:851` — helper cap | `accountStateCreatedCapacity` | `accountStateDeleteCapacity` |
| `NoopHalt.lean:575` — insert cap | bare `8192` | `accountStateDeleteCapacity` |
| `NoopHalt.lean:576` — flag-set cap | bare `8192` | `accountStateDeleteCapacity` |

## ⚠️ The seventh site deliberately does NOT change

`NonstorageEffectLog`'s `.set account_state_delete, account_state_pending + (accountStateTableBytes + accountStateCreatedCapacity * 32)` **keeps `accountStateCreatedCapacity`**: that term is the *created* arena's storage size, which is what the delete arena sits after. It is the one legitimate use of that constant on the delete side, and a sweep over "the delete bound" would have converted a correct line into a bug.

**That is why this is a classification task rather than a rename**, and it is the reason the change is six edits and not seven.

## Home: a leaf module with no imports

`EvmAsm/Codegen/ArenaCapacities.lean` imports nothing. The alternatives were both wrong:

- **`CreateCodeEffectLog`**, where the sibling capacities live, already has six importers. Three more raises hub reachability, which drives the long CI build — a measurable ongoing cost, not only a cycle risk. It would also invert dependency direction: `StorageReadLog` imports exactly one module today and would come to import a high-level `Programs` module.
- **`Rv64.Program`** is visible everywhere, but an arena capacity is not a RISC-V concept.

A leaf adds six import edges that **pull in nothing**, so hub reachability barely moves and there is no cycle to risk by construction.

## Headroom — measured, not assumed

The created and delete arenas are `.set` aliases into the `nea_sort_a`/`nea_sort_b` radix scratch, which allocates `2 × 38460 × 112 = 8,615,040` bytes. The delete arena's offset is `128 × 38460 + 8192 × 32 = 5,185,024`, leaving **3,430,016 bytes = 107,188 entries** of 32 bytes against the **8,192** the guards permit.

So the bound is conservative by roughly 13× and this is a **naming** fix rather than a bound correction — the guards were never too loose. And exceeding it is fail-closed: the insert refuses at capacity, the caller sets `account_state_overflow`, and **both** of that flag's consumers (`BlockVerdictMtxRuntime.lean:508`, `BlockVerdictFunction.lean:865`) reject the block with `bv_fail_code = 58` (GH #10964).

## BEHAVIOUR-NEUTRAL — byte-identity **measured**, not argued

`accountStateDeleteCapacity = 8192`, the same value every site already used, so the emitted text cannot move. Verified by rebuilding at this commit and comparing against artefacts recorded from a clean `0320615e2` tree **before** the change:

```
stateless_guest.s    63898ecf936f9d37de1983237382ee65a646b7a3fbea400af7c95f18900208af   IDENTICAL
stateless_guest.elf  e73eae8161c9e3311e9aa5e3ae1f7f52198a205f0c23931419fa85af7bc42127   IDENTICAL
```

Both artefacts reproduce exactly. **No layout regeneration is required and none was performed** — no `GuestAddrs`, `RegionMap`, `symbol-addresses.tsv` or `GuestImage` change, because no address or section size can move when the emitted bytes are identical.


pirapira:
Labelling. This is the **first behaviour-neutral claim today backed by a measurement rather than an inference**, and the distinction is worth naming because the day produced three weaker forms of the same argument.

| PR | byte-neutrality evidence | strength |
|---|---|---|
| #10963 | `.bss` delta equals slab minus retired record — a closed statement about the *difference* | strong, but still an argument |
| #10969 (extraction commit) | the commit touches **no generated files**, so emission cannot have moved | inference from a diff's shape |
| this PR | **both artefacts reproduce exactly** — `stateless_guest.s` `63898ecf`, `.elf` `e73eae81`, identical to the pre-change artefacts recorded from clean `0320615e2` | measured identity |

⭐ And it discharges the layout question **by construction**: no address, no section size, and no symbol can have moved, so there is nothing to re-pin and no direction to predict wrongly.

## Scope checks

- **Six operands, all converted**: one guard in `BlockVerdictCreationStage`, one guard plus one insert in `CreateCodeEffectLog`, two inserts in `NoopHalt`, one guard in `StorageReadLog` — matching the two independent censuses exactly. Four of the six were bare literals, including the `StorageReadLog` one that the issue's original write-up missed entirely.
- ⭐ **`NonstorageEffectLog` is correctly untouched.** Its `.set account_state_delete, account_state_pending + (accountStateTableBytes + accountStateCreatedCapacity * 32)` places the delete arena immediately *after* the created one, so the created capacity is the right constant there — it measures the created arena's extent rather than bounding the delete arena. A sweep over the delete bound would have turned that correct line into a bug.
- **Leaf module with no imports**, as chosen over defining it beside its siblings. That alternative would have added three import edges into `CreateCodeEffectLog`, which already has six importers — and hub reachability is what drives the long CI build — while also inverting dependency direction, since `StorageReadLog` imports exactly one module today.

## One note on the base

The byte-identity was measured on `0320615e2` and this branch is  current main. The proof still transfers: substituting a named constant for a literal of the same value cannot change emission on any base. Worth stating rather than leaving an unremarked base mismatch — line numbers in the issue thread already differed by six lines across two merges, with neither reader at fault.

The guards are also conservative by roughly 13× (about 107,188 entries of headroom against the 8,192 permitted), which is what makes 8192 safe to *name* — this is a naming fix, not a bound correction.

pirapira:
**Correction to my comment above: one sentence was mangled and stated the opposite of the truth.**

It reads "this branch is  current main". It should read: **this branch is BEHIND current main.** The word was eaten by shell command substitution when I posted — a backtick-quoted word inside a double-quoted string got executed rather than passed through. My own notes warn about exactly that for another tool and I did not apply it here.

The intended paragraph:

> The byte-identity was measured on `0320615e2` and this branch is **behind** current main. The proof still transfers: substituting a named constant for a literal of the same value cannot change emission on any base. Worth stating rather than leaving an unremarked base mismatch — line numbers in the issue thread already differed by six lines across two merges, with neither reader at fault.

Nothing else in that comment is affected, and the `merge!` label stands. Flagging it rather than silently editing, because a review comment that asserts a branch is current when it needs a rebase is the kind of wrong statement someone would reasonably act on.

PR #10974, opened 2026-07-30
codegen: snapshot value-CALL effects before recording

pirapira:
## Spec alignment

`process_message` snapshots rollback state before `move_ether`; a value CALL that reverts must therefore discard its debit/credit BAL records with the child. This moves the existing descriptor-only `record_message_value_transfer` invocation from before `call_frame_descend` to immediately after it, before the child body dispatches.

The helper performs no state lookup. Its operands are the caller-resolved static scratch cells (`cd_caller_be`, `nse_callee_be`, `cd_value_be`, `cd_balance_be`, `nse_acct`). They are safe only in the post-descend/pre-child-dispatch window: nested child calls reuse this scratch, so the record must not move into the child return path.

The obsolete `cd_nse_presnap_*` protocol is retired from the four production sites: the CALL producer, common descender consumer, CREATE stale-arm clear, and `.bss` backing cells. `CallFrameRoundtrip` retains its isolated probe declarations.

## Evidence

- `check-region-map.sh`, cache-free `check-asm-to-program.sh`, and `check-build-parallel.sh` pass.
- Regenerated artifact: `.text=0x66c34` (down 244 bytes from `0x66d28`), `.data=0x5370`, `.bss=0x1c105000`; `GuestAddrs.bnf_le_a=0xa3000ee0`. The emitted code shrinks with the retired presnap protocol rather than merely bypassing it.
- 00212 trace (input SHA `c4e3e3780485996a8cc01142fde929475bb6e0d664caf47177f97969aefa77d1`): rc=0; the three former presnap addresses receive zero writes. `call_frame_descend` enters at trace line 3,255,644 and `record_message_value_transfer` at 3,257,157, proving the record is after the snapshot.
- Expected-invalid 04778 still rejects (`succ=0`).
- Fresh random 200 (not comparable to earlier seeded subsets), frozen runner SHA `c6ac09fa72c2103b6d3e390b779b022ef7f220e5a39bb5f667114a89d78e5868`, candidate ELF SHA `fa5ee778249b524e7f73ecf9e9e0d0c3814ef576a5eaba6dfcfc86ae9288432c`: FA=0, FR=91, OK=109.


pirapira:
Labelling. Every acceptance criterion set for this change is discharged **positively**, and the diff matches the mapped retirement surface exactly.

## The criteria, and why each needed the form it took

| criterion | result |
|---|---|
| the protocol is **gone**, not merely unused | old presnap cells `beabf798`, `beabf7a0`, `beabf7a8` have **zero** commit-log writes |
| the record moved to the spec's position | `record_message_value_transfer` at trace line 3257157, **after** `call_frame_descend` at 3255644 |
| the control fixture still passes | `00212` rc=0, runtime OK |
| the invalid control still rejects | `04778` guest succ=0 |
| FA | **0** |

⭐ The trace pair is the load-bearing part, and the reason is worth stating: **`00212` passed before this change too.** A pass afterwards is equally consistent with the change having no effect, with the record having moved correctly, and with it having moved *wrongly* in a way this fixture cannot see. Only the before-and-after pair distinguishes those — the before half showed a store of 1 to the armed cell, then a read and a clear; the after half shows the cell never written at all.

## The diff matches the mapped surface, deletion for deletion

The lifecycle census bounded this to one producer, one consumer, one stale-arm clear and three backing cells. The diff is exactly that:

- `CallFrameDescend` **−15, +0** — the sole consumer removed
- `CreateFrameDescend` **−5, +0** — the stale-arm clear removed
- `EvmDispatchUnits` **−3, +0** — the three backing cells removed
- `ChildFrameHandlers` +21/−29 — the producer relocated, net −8

⭐ And an emitted-side corroboration not claimed in the body: **`.text` went DOWN by 244 bytes** (`0x66d28` → `0x66c34`). Retiring a protocol should *remove* code, and it did — the arm, the clear, and the stored-versus-live selection are gone rather than bypassed. A behavioural change that shrinks `.text` by the size of a removed mechanism is a far better signal than one that grows it, and it is the emitted echo of the zero-writes result.

## ⛔ One precision: the FR figure is not a comparison

`FA=0, FR=91, OK=109` is a **candidate-only** figure. The earlier #10967 leg reported `FR=96, OK=104`, and the five-row difference **must not be read as a gain**: the sample provenance differs (first-200-of-400 seed-7702 versus "random 200"), and #10969 merged between the two legs, so the bases are not adjacent either. Two different draws produce a five-row difference by chance alone.

What this leg establishes is **FA-clean and both pinned controls holding**, which is the criterion that was set. The gain question was explicitly a bonus and remains open; if the rows *are* the same set, a baseline leg on the immediate parent would settle it, and five reverting-value-CALL cases flipping to passing would be a genuine result worth naming. It is not claimed here.

## Site condition

The comment recording that the static descriptor scratch is valid **only in the post-descend, pre-child-dispatch window, because nested calls reuse it**, is present and is the right form: the operands survive by *ordering*, not by inherent durability, so anyone moving that record later — into the return path or past the child body — would silently read clobbered operands and produce a wrong pre-balance in a BAL row rather than a crash. That condition is named at the place someone would break it.

PR #10979, opened 2026-07-30
codegen: preserve created accounts across child-frame reverts

pirapira:
## Summary

`created_accounts` is transaction-scoped in execution-specs: `copy_tx_state` shares it across child states, while `restore_tx_state` restores writes but not the created-account set. Remove the guest's per-depth `account_state_created_count` checkpoint, its frame-return restore, and its 1025-slot backing allocation so a reverted child cannot erase a creation marker.

`created_count` deliberately remains outside the canonical rollback slab permanently. The slab captures per-frame rollback state; the created-account set is shared across the transaction by spec and must not be restored on frame return.

## Validation

- Regenerated to a fixed point. `.text` shrank `0x66c34 -> 0x66be8`; `.bss` shrank `0x1c105000 -> 0x1c103000`; `.data` stayed `0x5370`, and `GuestAddrs.bnf_le_a` remains `0xa3000ee0`.
- `LAKE_ARTIFACT_CACHE=false scripts/check-asm-to-program.sh`
- `LAKE_ARTIFACT_CACHE=false scripts/check-build-parallel.sh` (all lanes pass; olean/hash counts: `2155/2155`)

No behavioural fixture is expected for this structural conformance correction: it removes an incorrect rollback path and preserves the spec's shared transaction-scoped set.


pirapira:
Labelling. The source change is a pure deletion — seven lines, no additions — and the `.bss` pin **reconciles with a non-obvious derivation**, which is more informative than a naive match would have been.

## The `.bss` arithmetic, worked

The removed allocation is 1025 × 8 = **8200 bytes** (`0x2008`), but the pin moves `0x1c105000 → 0x1c103000`, a delta of only `0x2000` = **8192**. The 8-byte difference is **section alignment**, and it closes exactly:

```
0x1c105000 − 8200            = 0x1c102ff8
ceil to the 0x1000 boundary  = 0x1c103000   ← the pin
```

So the pin is the 4 KiB ceiling of the reduced size, and the 8-byte remainder is absorbed by alignment slack. ⭐ Worth stating in the body rather than leaving as a bare number: a reviewer computing `8200` and seeing `0x2000` would reasonably suspect the allocation was only partly removed. **A derivation that needs an explanation is stronger evidence than one that matches on sight** — this one can only close if exactly `1025 × 8` went away.

(This is also the direction that had to be re-derived rather than predicted: a dense arena anchored to `__BSS_END__` has previously absorbed freed bytes without the bound moving at all.)

## Source deletions match the mapped surface

- `CallFrameDescend` **−4** — the checkpoint store
- `CallFrameReturn` **−2** — the restore load and store
- `CreateCodeEffectLog` **−1** — the 1025 × 8 allocation

One writer, one reader, one allocation, matching the positive census exactly. Producers, readers and the transaction-start count reset are untouched, which is the part that keeps EIP-6780's permission working.

## The deferred decision from #10963 is now answered

`created_count` stays **permanently outside** the canonical rollback slab, and the reason is the spec's own shape rather than convenience: `copy_tx_state` shares `created_accounts` across child state, and `restore_tx_state` does not restore it. A field that must **not** roll back does not belong in a rollback record. That closes the question #10963 deliberately left open when it kept the field on a dedicated checkpoint.

## Scope note

⚠️ The observable consequence remains **vacuous** — to selfdestruct X you must execute X's code, and a reverted CREATE leaves X with no code, so the un-creation cannot be witnessed. This lands as **convergence, not a bug fix**, and the expiry recorded on #10940 still applies: if any future change makes X's code survive its creating frame's revert, or introduces a path executing with `current_target = X` without re-creating X, the missing-delete outcome becomes live.

⭐ The reason it is worth landing anyway is the disposition rule this issue established: **a vacuous defect with no spec counterpart closes; a vacuous divergence from an explicit spec property gets fixed.** Unobservability set its priority, not its validity.

pirapira:
## ⛔ Correcting my own review comment above: the `.bss` derivation I gave was **right in number and wrong in mechanism**.

I explained the pin as the **4 KiB ceiling** of the reduced size — `0x1c105000 − 8200 = 0x1c102ff8`, ceiled to `0x1c103000`. The number matched, so I presented it as established.

**A later deletion falsifies the mechanism.** #10986 removes five 8-byte cells (40 bytes) and moves the pin `0x1c103000 → 0x1c102fe0` — and `0x1c102fe0` is **not 4 KiB aligned**. So `.bss` is not constrained to a 4 KiB boundary at all, and my ceiling story could not have been the reason.

## What the two data points actually show

| PR | removed | raw subtraction | actual pin | actual − raw |
|---|---|---|---|---|
| #10979 | 8200 | `0x1c102ff8` | `0x1c103000` | **+8** |
| #10986 | 40 | `0x1c102fd8` | `0x1c102fe0` | **+8** |

A consistent **+8** across removals of wildly different sizes — 8200 versus 40 — is a systematic effect, not alignment slack that happens to swallow a remainder.

⭐ **Better hypothesis, and it is falsifiable:** both removals are `≡ 8 (mod 16)`, so if the symbol *following* the removed block requires **16-byte alignment**, each leaves exactly 8 bytes of padding behind. That predicts something checkable: **removing a multiple of 16 bytes should shrink `.bss` by exactly that amount.** Until someone tests that, this is a hypothesis and not a derivation.

## Why I am flagging it rather than quietly editing

⛔ My comment presented a mechanism as established when the number matching was the *only* evidence for it — and a coincidence is exactly what that looks like from inside. This is the error I spent the day correcting in others: **explaining a phenomenon before establishing its cause**, and in one case fabricating a plausible reason for a correct rule.

The number `0x1c103000` was and is correct, and #10979's deletion was and is complete — the source diff shows one writer, one reader, one allocation removed. **Only my explanation of the pin was wrong.** Nothing about the merged change needs revisiting.

⭐ And the practical consequence for whoever next deletes `.bss` cells: **do not derive the expected pin from a 4 KiB ceiling.** Compute the raw subtraction, expect a small positive offset from alignment of the following symbol, and re-derive rather than predict — which was the standing instruction on these PRs and remains the right one for a better reason than I gave.

Issue #10980, opened 2026-07-30
Emission sites bypass named constants with bare literals (3 instances, none found by looking)


pirapira:
## Pattern

**Emission sites bypass named constants with bare literals.** Three instances found tonight, all **incidentally
while doing something else** — which is the argument for a deliberate sweep. Base `origin/main` `f1a342e6d`
region unchanged from `1d78ef6c3`; citations below at `1d78ef6c3`.

| # | constant | named where | bypassed where | protected? |
|---|---|---|---|---|
| 1 | delete-arena capacity **8192** | *was* only `accountStateCreatedCapacity`, i.e. the **wrong** arena's name | 4 bare literals + 2 wrong-name uses | **fixed by #10971/#10972** — `accountStateDeleteCapacity` |
| 2 | `COST_PER_STATE_BYTE` **1530** | `amsterdamCostPerStateByte` (`AmsterdamSystemTx.lean:21`) | `BlockVerdictCreationStage.lean:476` and `NoopHalt.lean` deposit block emit bare `li …, 1530` | ⛔ **no** |
| 2b | `OPCODE_KECCAK256_PER_WORD` **6** | **nowhere in codegen** (spec: `gas.py:229`) | both deploy sites emit bare `li …, 6` | ⛔ **no** |
| 3 | NEW_ACCOUNT state gas **183600** | `liAmsterdamNewAccountStateGas` / `liStateGasRuntime` (`AmsterdamSystemTx.lean:69-77`) | ≥8 bare `li …, 183600` — `Dispatch.lean:1288`, `:1302`; `ChildFrameCreateTail.lean:557`, `:563`; `ChildFrameHandlers.lean:233`, `:267`; `NoopHalt.lean:39`, `:55` | ✅ **yes**, see below |

### ⭐ The severity is not uniform, and the codebase already contains the mitigation

Instance 3 looks the worst by count and is the **least** dangerous, because `AmsterdamSystemTx.lean:27` has

```
#guard amsterdamStateBytesPerNewAccountV2 * amsterdamCostPerStateByte = 183600
```

**That `#guard` pins the product.** If someone changed the byte count, the guard fails at build time — so the bare
`183600` literals cannot silently diverge from the named value while it stands. Instances 2 and 2b have **no such
tripwire**: change `amsterdamCostPerStateByte` and the two deploy sites keep emitting `1530` with nothing failing.

*(I checked instance 3 for a live divergence before filing, because a stale comment suggested one — see below —
[...]

PR #10983, opened 2026-07-30
docs(codegen): fix the new-account state-gas docstring (GH #10980)

pirapira:
Fixes the `liAmsterdamNewAccountStateGas` docstring, which is the trap flagged on GH #10980. Docstring-only.

## What was wrong

```
/-- Assembly helper for the new-account state gas (112 bytes * runtime cost, EIP-8037). -/
def liAmsterdamNewAccountStateGas (reg : String) : String :=
  liStateGasRuntime reg amsterdamStateBytesPerNewAccountV2             -- 112 * cost (was 120*1530)
```

Both halves are wrong, and **the second is worse than the first**:

- `amsterdamStateBytesPerNewAccountV2` is **120**, not 112 — and the `#guard` two definitions above already pins `120 * 1530 = 183600`;
- the parenthetical presents the **correct** value as the **superseded** one, so the prose accounts for the very number that would have refuted it.

## ⛔ Why this is more than a typo

**A bare wrong number invites verification; a wrong number with a rationale suppresses it.**

Taking the docstring at face value gives `112 * 1530 = 171360`. Meanwhile the refund paths still emit a bare `183600` — `Dispatch.lean:1288`, `:1302`; `ChildFrameCreateTail.lean:557`, `:563`; `ChildFrameHandlers.lean:233`, `:267`; `NoopHalt.lean:39`, `:55` — while the **charge** goes through this helper. Those two facts assemble into:

> every NEW_ACCOUNT charge-then-refund cycle leaks **12,240** gas

A plausible mechanism, a stated charge-versus-refund asymmetry, eight citable sites, and a docstring as its authority. **That gas defect does not exist**, and this docstring was its only support. Reading the `def` — one line — was the entire refutation.

It would also have mis-assigned severity **in both directions at once**: the instance *with* a `#guard` tripwire and eight bare sites would have been filed as the live defect, while the two sites that genuinely have **no** tripwire (the bare `1530` at `BlockVerdictCreationStage.lean:476` and in `NoopHalt.lean`'s deposit block, per #10980) would have looked like housekeeping beside it.

## The replacement

Cites the constants instead of restating a count, and records the trap so the next reader does not inherit it. The `#guard` remains the tripwire. No byte count is reintroduced — that is the specific thing that went wrong.

## BEHAVIOUR-NEUTRAL — byte-identity **measured**, not argued

A Lean docstring plus one trailing comment, outside every emitted string literal. Verified by building clean at `f1a342e6d`, recording both artefacts, applying the change and rebuilding:

```
stateless_guest.s    d372c326b5cce457b5e6bc64d7af45c8fff855345939e12011e50fd01df68d94   IDENTICAL
stateless_guest.elf  fa5ee778249b524e7f73ecf9e9e0d0c3814ef576a5eaba6dfcfc86ae9288432c   IDENTICAL
```

Both reproduce exactly, so **no layout regeneration is required and none was performed** — no address or section size can move when the emitted bytes are identical.

## Scope

This fixes only the incorrect docstring. It does **not** address #10980's substantive item — the bare `1530` and unnamed `6` at the two deploy sites, which have no tripwire and are the real exposure. That remains open with its recipe on the issue.


pirapira:
Labelling. Comment-only, one file, **both artefact hashes identical** — the strongest neutrality proof available, and the appropriate one here since a docstring cannot change emission.

## Why this was the right piece of #10980 to land first

It is the **trap itself** rather than the housekeeping around it. The old form did not merely state the wrong byte count — it also carried `(was 120*1530)` inline, presenting the *correct current value as superseded*. That is worse than a bare wrong number, because **it pre-answers the question a reader would otherwise ask**: a bare 112 invites checking, while 112-with-an-explanation suppresses it.

⭐ And the fix does the two things that prevent recurrence rather than just correcting the digit:

1. **The inverted history is removed**, not only the number. Leaving it would have handed the same trap to the next reader.
2. **The docstring now cites the constants by name** (`amsterdamStateBytesPerNewAccountV2`, `amsterdamCostPerStateByte`) and the `#guard`-pinned product, with the explicit instruction *do not reintroduce a byte count here; cite the constant.* That is a structural fix — a name cannot drift from the value it names.

⭐ Recording the counterfactual defect in the docstring is also right. Taking the old prose at face value gives `112 × 1530 = 171,360`, which against the bare `183600` literals still on the refund paths assembles into a plausible *"every charge-then-refund cycle leaks 12,240 gas"* — with a mechanism, a charge-versus-refund asymmetry, eight cited sites, and a docstring as its authority. **A reader who sees only the corrected number learns nothing; one who sees what the wrong number nearly produced learns to check the constant.**

## An unintended cross-check on this afternoon's batch merge

The baseline ELF measured here at `f1a342e6d` is `fa5ee778…`, which is **the same sha reported as #10974's candidate ELF** before it merged. That reconciles exactly and is worth noting: of the three PRs in batch #10977, #10972 was byte-identical and #10975 was docs-only, so **main's post-batch ELF should equal #10974's candidate ELF — and independently measured, it does.**

That is a third, unplanned confirmation that the batch merge preserved emission rather than silently taking one side's artefacts, alongside the `textSizeBytes 0x066c34` and `.bss 0x1c105000` pins reconciling.

## Scope note

The remaining #10980 items are not folded in and should not be: the `1530` deploy sites and the `183600` refund sites need either the named constant at the site **or** a `#guard` pinning the literal — and the severity ranking there is by **tripwire presence, not site count**, since the eight-site instance is already pinned and the two-site one is not.

PR #10985, opened 2026-07-30
codegen: mark created_accounts pre-body on the top-level creation route (GH #10938 cut 2)

pirapira:
## Summary

GH #10938 **cut 2**: unify the two creation routes onto the spec's position for
`mark_account_created`. The nested CREATE route already marks pre-body; the top-level
creation route marked post-body. After this PR both mark before the initcode runs, and
the post-body insert is gone.

**This is a semantics unification with no reachable behaviour change that I could
identify** — see the consumer enumeration below. It is **not** byte-identical (an insert
moves between routines), so the four-step layout regen was run to a fixed point.

## The spec position

`execution-specs` at pinned commit `e5a8caf1b`,
`src/ethereum/forks/amsterdam/vm/interpreter.py`:

```python
destroy_storage(tx_state, message.current_target)          # :202
...
mark_account_created(tx_state, message.current_target)     # :208
increment_nonce(tx_state, message.current_target)          # :210
evm = process_message(message)                             # :212  <- initcode runs here
```

So the mark precedes the body unconditionally, and in particular precedes the snapshot —
`copy_tx_state` is taken inside `process_message` (`:380`).

## The two routes, before

| route | marked at | relative to initcode |
|---|---|---|
| nested CREATE / CREATE2 | `Programs/CreateFrameDescend.lean:135`, in `create_frame_descend` | **before** |
| top-level creation | `Programs/CreateCodeEffectLog.lean:846`, inside `create_record_code_effect` | **after**, and only on a successful deposit |

The descent site already carries the intent in prose: *"deliberately before initcode
executes, so CREATE followed by constructor SELFDESTRUCT is still classified as
created-this-tx even with no deposit."* The top-level route enters the callable
dispatcher directly rather than through a descent (`BlockVerdictCreationStage.lean:425-433`),
so it had no descent to mark at and inherited the deposit-time mark instead.

## Placement, and why after-the-capture is equivalent to the spec's before-the-snapshot

The new insert sits between `dispatcher_capture_body_state` and the dispatcher call,
inside the existing saved-`ra` block. The spec marks *before* its snapshot; the guest now
marks *after* its capture. Those are equivalent here for a checkable reason rather than
by assumption:

**`account_state_created_count` is not one of the thirteen fields of
`body_state_snapshot_by_depth`** (`Programs/BlockVerdictDispatchTx.lean:492-505`, offsets
`0..96`), so `dispatcher_restore_body_state` cannot roll the mark back. **GH #10979 is
what made that true** — it removed `account_state_created_checkpoint` — and it is the
spec's own shape: `copy_tx_state` leaves `created_accounts` shared, and
`restore_tx_state` (`state_tracker.py:823-826`) restores four other fields.

Register safety: `a0-a3` are dead across the two following lines (they set up `t4`/`t5`,
and the dispatcher takes its input through `runtime_dispatcher_input_ptr`);
`code_state_address_set_insert` preserves every `s`-register, so `s0` survives. Overflow
is fail-closed exactly as at the descent site — set `account_state_overflow`, which both
consumers turn into `bv_fail_code = 58`.

## Why there is no reachable behaviour change — consumer enumeration from the EMITTED stream

I expected this to be a live divergence and it is not. Recording the argument so the
hypothesis is not re-derived.

`grep account_state_created gen-out/regionmap/stateless_guest.s` gives the complete
population: two inserts (now both pre-body), the per-tx and per-MTx resets, the
`.set` alias, and **exactly two readers** — both `jal ra, account_state_created_contains`,
at `.s:55021` and `.s:55995`. Neither can observe the top-level address being marked
earlier:

* **`.s:55021`** is the nested route's own deposit-path aliveness query. Its argument is
  `create_address_be`, the nested child's address — not the top-level creation's.
* **`.s:55995`** is the SELFDESTRUCT `created_in_tx` codecheck, and it is reached **only
  when `create_frame_flag[depth] == 0`** (`Programs/NoopHalt.lean:555-558` takes the
  flag branch first). `BlockVerdictCreationStage.lean:382` sets
  `create_frame_flag[0] = 1` for the whole top-level creation route, cleared at `:448`
  once the dispatcher returns. So a SELFDESTRUCT in top-level constructor code already
  takes the *first* check and never consults the created set.
* The remaining shape — the created address being SELFDESTRUCTed from some other frame
  before its deposit — cannot occur: it has no deployed code until the deposit, so no
  frame can execute its SELFDESTRUCT.

The direction of change is also monotone: an address is now marked **earlier, never
later**, so membership over time is a superset and no reader can lose a positive it
previously had.

## Prerequisite commit: `accountStateCreatedCapacity`

The new site needs to name the created arena's bound, and it lived in
`Programs/CreateCodeEffectLog.lean` — a module neither marking site imports, which is
why the descent site carried a **bare `8192`**. Moving the definition into the
import-free `EvmAsm/Codegen/ArenaCapacities.lean` costs nothing (`CreateCodeEffectLog`
already imports it, so its own uses are unaffected) and lets the descent site name it
too. Same shape and same module as GH #10971. Value unchanged.

## Orthogonality to #10940 — settled rather than asserted

#10940's shared-field question concerned the created-count checkpoint. **#10979 has
landed, so that checkpoint is gone from main**; the question no longer exists in the
tree. Moving the marker write earlier *within* the capture/restore window neither fixed
nor worsened it. Stating this explicitly because a reviewer seeing
`account_state_created` in this diff would otherwise go looking.

## Layout regen — run to a fixed point

`.text` changes shape but not size, so `RegionMap.lean` needed **no** edit:

| section | ELF after | existing pin |
|---|---|---|
| `.text` | `0x066be8` | `0x066be8` |
| `.data` | `0x005370` | `0x005370` |
| `.bss` | `0x1c103000` | `0x1c103000` |

Regenerated: `scripts/asm-fixtures/symbol-addresses.tsv` (289 rows moved) and
`EvmAsm/Codegen/GuestAddrs.lean` (114 symbols moved).
`Proofs/GuestImageEntries.lean` came out unchanged. `GuestAddrs.bnf_le_a` is still
`0xa3000ee0`, and `.data`/`.bss` are unmoved, so `Proofs/GuestImage.lean`'s handwritten
`.bss` literals need no update. `scripts/check-region-map.sh` reports OK on all five
lines. A **second complete pass changed nothing**, which is the fixed point the
procedure asks for rather than a one-pass repin.

## Verification

* `lake build` — clean, no errors and no `sorry`.
* `scripts/check-region-map.sh` — OK (all five lines).
* `scripts/check-no-hardcoded-guest-pc.sh` — OK.
* `scripts/check-guarded-handler-bytes.sh` — **fails, and fails identically on the
  unmodified tree** (I stashed the change and re-ran): a local build cache has
  `Callable.olean.hash` without `Callable.olean`, so
  `lake env lean --run scripts/emit_guarded_handler_driver.lean` cannot load
  `EvmAsm.Evm64.DivMod.Callable`. Pre-existing and environmental; the script is also not
  referenced from `.github/workflows`, so it is not a CI gate. Reporting it rather than
  omitting it.

## Observation, not changed here

`Programs/NoopHalt.lean:167-185` zeroes `create_target_alive_current_tx`, sets it from
`account_state_created_contains(create_address_be)`, and then `:229-230` **overwrites**
that cell from `create_target_alive_flag[depth]` before `frame_return`, with no read in
between. Since `create_frame_descend` has already inserted that same address, the query
is also forced to 1. This PR does not change it — the descent insert already predated the
query before this diff — but it looks like a dead computation and is worth a separate
look.

---

⚠️ **Citation corrected by review:** this body originally cited **#10784** for the cut decomposition. The cuts are defined on **#10938** (*the guest has two creation processors where execution-specs shares `process_create_message`*). **#10784 is an unrelated open bug** — a failed transaction's SELFDESTRUCT being committable because `account_state_delete_count` survives — so the original citation would have sent a reader to the wrong issue.


pirapira:
Labelling. One precision on the safety argument, and a note on why the extra file belongs here.

## ⛔ The monotonicity argument protects one direction only — the reader enumeration is what protects the other

The body states the change is monotone: an address is marked **earlier, never later**, so membership over time is a superset and **no reader can lose a positive**. That is correct and it rules out a *missed* deletion.

⚠️ **But a superset can also produce a positive where there previously was none.** Between the old mark position (post-body) and the new one (pre-body), a reader asking *is X created?* now gets **yes** where it got **no**. Monotonicity does not exclude that — it guarantees it, in exactly the window the change creates.

⭐ So what makes this safe is the **reader enumeration**, not the monotonicity: `account_state_created` has exactly two readers in the emitted stream, both `jal account_state_created_contains` — one on the nested route querying *its own* address, one a codecheck gated on `create_frame_flag` being **0**, which the top-level route sets at CreationStage `:382` and clears at `:448`. So neither can observe the new window. The two legs together are complete: **monotonicity rules out losing a positive; the enumeration rules out gaining one.** Worth stating that split explicitly, because monotonicity alone reads as sufficient and is not.

## The equivalence rests on a fact #10979 established hours ago

Marking after `dispatcher_capture_body_state` rather than before it is equivalent for a **checked** reason: `account_state_created_count` is not among the thirteen fields of `body_state_snapshot_by_depth`, so the restore cannot roll the mark back. ⭐ That became true when #10979 removed the created-count checkpoint — before it, this PR's placement would have been wrong. Two changes landing the same evening where the second depends on the first is worth noting in the record.

And it matches the spec's own shape: `copy_tx_state` leaves `created_accounts` shared.

## The extra file is a dependency, not scope creep

`ArenaCapacities.lean` gains `accountStateCreatedCapacity` because **cut 2's new top-level marking site cannot reach where that constant lived.** `CreateFrameDescend` and `BlockVerdictCreationStage` are the two sites that mark an account created, and neither imports `CreateCodeEffectLog` — which is why the descent site was already carrying a **bare 8192**, the same shape as the four bare delete bounds behind #10971. So this is the fourth instance of #10980's pattern, found while doing something else and fixed in passing because the alternative was adding a fifth bare literal.

⭐ And the docstring records the trap that produced #10971 in the first place: this constant is **equal to `accountStateDeleteCapacity` and is not the same quantity**, while `NonstorageEffectLog`'s `.set account_state_delete` offset legitimately uses *this* one, because the created arena is what the delete arena sits after. That is the distinction a well-meaning "unify these two 8192s" change would destroy.

## Scope, stated as the author framed it

This is a **unification with no reachable behaviour change**, not a fix. The author went in expecting a live top-level EIP-6780 divergence and the consumer enumeration refused it — `create_frame_flag` *is* set at index 0 for the whole top-level route, so the SELFDESTRUCT path takes the flag branch and never consults the created set. The refuted hypothesis is recorded in the body rather than only the verdict, which is the right call: it stops the next reader re-deriving it.

pirapira:
## ⛔ Correction from the author: the issue number in this PR is wrong

**The cut decomposition belongs to GH #10938, not GH #10784.** Every "GH #10784 cut 2"
in this PR — the title, the body, and three source comments — should read **GH #10938
cut 2**.

* **#10938** is *"convergence: the guest has two creation processors where
  execution-specs shares `process_create_message`"*, and its comment thread is where
  cuts 1–4 are defined. Cut 2 there is *"split the flag/slots/nonce-seed bundle by
  field … only the created-marker semantics move. Nothing crosses a mark."* That is
  exactly this change.
* **#10784** is an unrelated open bug — *"Failed transaction's SELFDESTRUCT can be
  committed: `account_state_delete_count` survives into `account_state_commit_pending`"*.
  This PR does nothing about it. Citing it here would suggest otherwise to anyone
  searching, which is the actual cost of the error.

The three source comments are at `ArenaCapacities.lean:67`,
`BlockVerdictCreationStage.lean:442`, and `CreateCodeEffectLog.lean:845`. They cannot be
corrected here: the PR carries `merge!`, so the branch is the bot's and I will not push
to it. **The squashed commit message also inherits the PR title**, so the wrong number
reaches git history unless the title is edited before the merge — that call belongs to
whoever holds the label, not to me.

Nothing about the change itself is affected: the spec citations
(`vm/interpreter.py:202/208/210/212` at pinned `e5a8caf1b`), the reader enumeration, the
`create_frame_flag[0]` refutation, and the layout evidence are all unchanged and stand
as written. The defect is purely the cross-reference.

I will file the source-comment and `PLAN.md` corrections as a follow-up once this
merges.

Issue #10982, opened 2026-07-30
Four per-depth checkpoint arrays superseded by the canonical slab are dead: verified unreachable by name and by arithmetic


pirapira:
## Four per-depth checkpoint arrays superseded by #10963's canonical slab, verified unreachable by name **and** by arithmetic

```
account_state_pending_checkpoint
account_state_delete_checkpoint
account_writes_undo_checkpoint
create_nonce_undo_checkpoint
```

⛔ **`account_state_created_checkpoint` is deliberately excluded** — it is a fifth member of this family, and **#10979 already removes it** as part of #10940. A PR scoped to five would conflict with work in flight; this issue covers the remaining four and must be based on post-#10979 main.

⛔ **`storage_writes_undo_checkpoint` is NOT dead** and must not be touched. It has three live references and is the **un-migrated** journal — that is #10619. Three of four journals moved their counts into the slab and left their arrays behind; the fourth still uses its array because its migration is the open issue. That asymmetry is the evidence this classification is right, not a gap in it.

## Why they are dead: the slab superseded them

#10963 introduced `body_state_snapshot_by_depth` and the current mechanism reads and writes the slab's `_count` fields at offsets `+64`, `+72`, `+80`, `+88`, `+96`. The old per-depth `_checkpoint` arrays were superseded and left allocated. ⭐ The apparent conflict — recent analyses citing these quantities as *live* — resolves because those cite the **`_count`** symbols, not the **`_checkpoint`** arrays.

## The gate: arithmetic reachability, checked per case rather than assumed

Occurrence count 1 in the emitted stream (declaration only) is necessary but **not sufficient**, because a cell can be reached by offset from a neighbouring referenced symbol without ever being named — `NonstorageEffectLog` computes the delete arena's offset from `accountStateCreatedCapacity * 32`, so the precedent exists in this tree. Each case was checked:

- The slab's own arithmetic (`depth × 104`, spanning 0–106,600 bytes) **stays inside** `body_state_snapshot_by_depth` and does not reach `account_writes_undo_checkpoint`, which sits after it.
- The three 8200-byte per-depth count arrays are adjacent to `account_state_scratch` and `account_state_durable` — both referenced — but **no code computes offsets to them from those bases**.
- `create_nonce_undo_checkpoint` sits between the referenced `create_nonce_undo_count` and `create_nonce_undo_log`, and the undo code **reads the count by name and iterates the log**, never touching the checkpoint array.

[...]

PR #10986, opened 2026-07-30
codegen: remove dead call-frame data cells

pirapira:
## Summary

Removes five dead call-frame data symbols after their last readers were retired.

- #10974 retired the `cd_nse_presnap_{armed,count,overflow}` protocol; this removes its production and roundtrip declarations.
- An emitted-stream sweep found `cd_burn_log_pending` and `cd_xfer_log_burn` with one emitted occurrence each on the base. They are a separate, unidentified retired mechanism; this PR does not attribute them to #10974.
- The four superseded checkpoint arrays remain deliberately out of scope for separate #10982 work.

## Layout and regeneration

Regenerated to a fixpoint. The linked ELF measured `.text=0x66be8` and `.data=0x5370` unchanged; `.bss` changed from `0x1c103000` to `0x1c102fe0`.

The source removal is 40 production bytes. The emitted local addresses confirm that `evm_access_account_count` moves by the full 40 bytes, but its following `evm_access_account_table` has `.balign 32`: the old count ended at `0xbeabd7b8` and had 8 padding bytes before the table at `0xbeabd7c0`; the new count ends at `0xbeabd790` and needs 16 padding bytes before the table at `0xbeabd7a0`. The 8 additional padding bytes explain the measured 32-byte section reduction.

`GuestImage` and `RegionMap` are repinned to the linked TSV values. `bnf_le_a` remains `0xa3000ee0`.

## Validation

- cache-free `scripts/check-build-parallel.sh`
- `scripts/check-asm-to-program.sh` (390 converted definitions)
- `scripts/check-progress.sh`, `scripts/check-drift.sh`, and `scripts/check-axioms.sh`
- stale-symbol sweep is empty in both Lean source and emitted assembly

No behavioural fixture run: this deletes storage with no remaining emitted references.


pirapira:
## Releasing the hold. My arithmetic objection was not a gate, and I was wrong to treat it as one.

I held this pending an unexplained 8 bytes in the `.bss` delta. Three data points now exist and **no model I have fits all three**:

| PR | removed | `.bss` shrank | unexplained |
|---|---|---|---|
| #10979 | 8200 (one array) | 8192 | 8 |
| #10986 | 40 (five cells) | 32 | 8 |
| #10988 | 32,800 (four arrays) | 32,768 | 32 |

⛔ **My 16-byte-alignment hypothesis is falsified.** It predicted an exact shrink for #10988, since 32,800 is a multiple of 16. Observed: 32 bytes unexplained. And the per-array reading that fits #10979 and #10988 (8 bytes per array) fails #10986, where five separate cells produced 8 rather than 40.

**So that is the second mechanism I have proposed for this and the second one that did not hold** — after the 4 KiB ceiling, which #10986's non-4K-aligned pin refuted. I am not offering a third.

## The reframing, which is the actual correction

⭐ **The pin is produced by the toolchain and verified by `check-region-map`, which passes on both PRs. My arithmetic was a redundant reviewer heuristic that I mistook for a required check** — and when a heuristic disagrees with the tool that owns the number, the tool is right.

What actually establishes these deletions is independent of the pin arithmetic and is present on both:

- the **source diff** — pure deletions, the exact symbols named, nothing else touched
- **emitted-stream occurrence counts** confirming no remaining reference by name
- for #10988, a **per-case arithmetic-reachability audit** ruling out reach-by-offset from each adjacent referenced base — including the case that only cleared because the undo code reads a count by name and iterates a log, never touching the array between two referenced symbols
- `check-region-map`, `check-asm-to-program`, cache-free `check-build-parallel`, drift and axioms all passing
- `.text` and `.data` unchanged, which is the right expectation for deleting data reservations only

## What remains genuinely open, as a curiosity rather than a blocker

The `.bss` accounting is not understood: declared removals consistently exceed the section reduction by a small amount that varies with the shape of what was removed. That is worth someone's attention on its own terms — a section-size computation nobody can predict is a weak spot in the layout pins — but it is **pre-existing, affects a number the tooling owns and validates, and blocks nothing here.** #10979 merged with the same discrepancy and is fine.

⭐ Lesson I am recording against myself rather than the PR: **twice I explained a number before establishing its cause, then presented the explanation as a derivation.** The correct standard is the one already applied to these branches — *re-derive and accept what you find, in either direction* — and an unexplained remainder should be reported as unexplained, which is exactly what the author did and what I then over-interpreted.

PR #10987, opened 2026-07-30
docs: record the 2026-07-31 creation-route unification and gas-constant naming

pirapira:
Docs only — no `EvmAsm/` change, so the guest emission is untouched by construction.

`CLAUDE.md` asks that PLAN.md be updated for the session's work. Records two things
from 2026-07-31:

* **#10938 cut 2** (PR #10985) and, specifically, **why it is a unification and not a
  fix** — including the false inference it refutes: `BlockVerdictCreationStage.lean:382`
  sets `create_frame_flag[0] = 1` for the top-level creation route even though that
  route never descends, so a top-level constructor SELFDESTRUCT takes the flag branch
  and never consults the created set. Recorded because I built a divergence on the
  opposite assumption before checking the flag's writers.
* **#10980's substantive item** (PR #10984) — the two code-deposit sites cannot use
  `liStateGasRuntime` because their byte count is runtime-only, which is why they were
  the only sites still emitting bare `1530`/`6`.

Also notes the layout outcome that is easy to misread: `.text` changed *shape* but not
*size* (`0x066be8` on both sides of cut 2), so `RegionMap` needed no edit while the TSV
and `GuestAddrs` still moved.

---

⚠️ **Citation corrected by review:** this body originally cited **#10784** for the cut decomposition. The cuts are defined on **#10938** (*the guest has two creation processors where execution-specs shares `process_create_message`*). **#10784 is an unrelated open bug** — a failed transaction's SELFDESTRUCT being committable because `account_state_delete_count` survives — so the original citation would have sent a reader to the wrong issue.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -4744,6 +4744,33 @@ through ECALL bridges (extending `EvmAsm/EL/Keccak*EcallBridge.lean`).
   deferred sibling in #10973's same in-place/high-water-mark cell, requiring
   separate read/write lifetimes rather than a truncation patch.
 
+  **Creation-route unification, measured 2026-07-31.** #10784 cut 2 (PR #10985)
+  moves `mark_account_created` for the **top-level** creation route to the
+  spec's pre-body position (`vm/interpreter.py:208`, before `process_message`
+  at `:212`), matching what `create_frame_descend` already did for nested
+  CREATE, and removes the post-deposit insert from
+  `create_record_code_effect`. It is a **unification, not a fix**: the
+  emitted stream has exactly two readers of `account_state_created`, both
+  `account_state_created_contains`, and neither can observe the top-level
+  address being marked earlier. In particular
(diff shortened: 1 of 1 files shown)


pirapira:
## ⛔ Correction from the author: one wrong issue number in the added text

`PLAN.md:4747` in this PR reads *"#10784 cut 2 (PR #10985)"*. **It should be #10938
cut 2.** The cut decomposition lives on GH #10938 (*"convergence: the guest has two
creation processors where execution-specs shares `process_create_message`"*); GH #10784
is an unrelated open bug about a failed transaction's SELFDESTRUCT being committed, and
this paragraph says nothing about it.

Same error as PR #10985, same origin, and I cannot fix it here — this PR carries
`merge!`, so the branch is the bot's and I will not push to it. Everything else in the
paragraph is unaffected: the reader enumeration, the `create_frame_flag[0]` refutation,
and the `.text`-shape-not-size layout note all stand.

Follow-up will correct this line together with the three source comments in #10985.

PR #10988, opened 2026-07-30
codegen: remove superseded checkpoint arrays

pirapira:
## Summary

Removes the four dead per-depth checkpoint reservations identified in #10982:

- `account_state_pending_checkpoint`
- `account_state_delete_checkpoint`
- `account_writes_undo_checkpoint`
- `create_nonce_undo_checkpoint`

They were superseded by #10963’s `body_state_snapshot_by_depth` slab. The source and emitted-stream sweeps find no remaining reference to any removed array. The reachability audit in #10982 also rules out offset-based access: the slab arithmetic stays within its 104-byte-per-depth allocation; no arithmetic derives the account-state array offsets; and the create-nonce undo path names its count and iterates its log.

`storage_writes_undo_checkpoint` remains intact with two live call sites plus its declaration. `account_state_created_checkpoint` remains out of scope because #10979 already removed it.

## Layout and regeneration

Regenerated to a fixpoint from post-#10979 main. The linked ELF measured `.text=0x66be8` and `.data=0x5370` unchanged; `.bss` changed from `0x1c103000` to `0x1c0fb000` (the linked, alignment-aware `0x8000` reduction). `GuestAddrs`, `GuestImage`, `RegionMap`, and the symbol TSV are repinned to the linked output. `bnf_le_a` remains `0xa3000ee0`.

## Validation

- cache-free `scripts/check-build-parallel.sh`
- `scripts/check-asm-to-program.sh` (390 converted definitions)
- `scripts/check-progress.sh`, `scripts/check-drift.sh`, `scripts/check-axioms.sh`
- stale-array sweep empty in Lean source and emitted assembly

No behavioural fixture run: this only deletes unreachable `.bss` reservations.


pirapira:
## Releasing the hold. My arithmetic objection was not a gate, and I was wrong to treat it as one.

I held this pending an unexplained 8 bytes in the `.bss` delta. Three data points now exist and **no model I have fits all three**:

| PR | removed | `.bss` shrank | unexplained |
|---|---|---|---|
| #10979 | 8200 (one array) | 8192 | 8 |
| #10986 | 40 (five cells) | 32 | 8 |
| #10988 | 32,800 (four arrays) | 32,768 | 32 |

⛔ **My 16-byte-alignment hypothesis is falsified.** It predicted an exact shrink for #10988, since 32,800 is a multiple of 16. Observed: 32 bytes unexplained. And the per-array reading that fits #10979 and #10988 (8 bytes per array) fails #10986, where five separate cells produced 8 rather than 40.

**So that is the second mechanism I have proposed for this and the second one that did not hold** — after the 4 KiB ceiling, which #10986's non-4K-aligned pin refuted. I am not offering a third.

## The reframing, which is the actual correction

⭐ **The pin is produced by the toolchain and verified by `check-region-map`, which passes on both PRs. My arithmetic was a redundant reviewer heuristic that I mistook for a required check** — and when a heuristic disagrees with the tool that owns the number, the tool is right.

What actually establishes these deletions is independent of the pin arithmetic and is present on both:

- the **source diff** — pure deletions, the exact symbols named, nothing else touched
- **emitted-stream occurrence counts** confirming no remaining reference by name
- for #10988, a **per-case arithmetic-reachability audit** ruling out reach-by-offset from each adjacent referenced base — including the case that only cleared because the undo code reads a count by name and iterates a log, never touching the array between two referenced symbols
- `check-region-map`, `check-asm-to-program`, cache-free `check-build-parallel`, drift and axioms all passing
- `.text` and `.data` unchanged, which is the right expectation for deleting data reservations only

## What remains genuinely open, as a curiosity rather than a blocker

The `.bss` accounting is not understood: declared removals consistently exceed the section reduction by a small amount that varies with the shape of what was removed. That is worth someone's attention on its own terms — a section-size computation nobody can predict is a weak spot in the layout pins — but it is **pre-existing, affects a number the tooling owns and validates, and blocks nothing here.** #10979 merged with the same discrepancy and is fine.

⭐ Lesson I am recording against myself rather than the PR: **twice I explained a number before establishing its cause, then presented the explanation as a derivation.** The correct standard is the one already applied to these branches — *re-derive and accept what you find, in either direction* — and an unexplained remainder should be reported as unexplained, which is exactly what the author did and what I then over-interpreted.

pirapira:
## Reviewer note: the `.bss` delta reconciles, and the mechanism is `.balign 32` padding at **three** sites

Adding this myself rather than asking the author to edit — this PR carries `merge!` and is bot-owned, so the author correctly declined to touch it. Recording here keeps the derivation with the change instead of in a message stream.

The raw removal is four arrays × 1025 × 8 = **32,800** bytes (`0x8020`), but the pin moves `0x1c103000 → 0x1c0fb000`, a measured delta of **32,768** (`0x8000`). The 32-byte difference is **padding growth at three `.balign 32` boundaries**, each verified from the emitted image's addresses:

| site | before | after | added pad |
|---|---|---|---|
| account-writes | `auth_prepare_mark` ends `…c30` → `builder_pre` at `0xbba16c40` (16 pad) | mark ends `…c28` → `builder_pre` at `0xbba14c40` (24 pad) | **+8** |
| account-state pair | `delete` ends `0xbd901838` → `scratch` at `0xbd901840` (8 pad) | preceding `pre-account` ends `0xbd8fb828` → `scratch` at `0xbd8fb840` (24 pad) | **+16** |
| create-nonce | `create_cd_desc` ends `…bbb0` → `address_word` at `0xbde3bbc0` (16 pad) | desc ends `…bba8` → word at `0xbde33bc0` (24 pad) | **+8** |

8 + 16 + 8 = **32**, so `0x8020 − 32 = 0x8000` exactly. The direct successors each moved by their full raw deletion *before* the alignment adjustment.

## ⛔ Why this note exists at all: reviewer arithmetic on `.bss` was wrong three times tonight

I proposed three mechanisms for these deltas from the numbers alone and **all three failed**:

1. a **4 KiB ceiling** — refuted by #10986's pin `0x1c102fe0`, which is not 4 KiB aligned;
2. **16-byte alignment** — refuted by this PR, whose removal is exactly `0 mod 16` and still lost 32 bytes;
3. **four sites at +8 each** — refuted by the table above: three sites, contributing 8/16/8.

> ⭐ **Delta arithmetic cannot distinguish between structural explanations.** A 32-byte shortfall is equally consistent with four sites at 8, three summing to 32, or one at 32 — and the total does not select among them. **Only the addresses do.**

So the practical guidance for the next person deleting `.bss` cells: **compute the raw subtraction, expect it to come out short, and read the emitted image's addresses to account for the difference — do not model it.** A naive subtraction will disagree with the pin, and that disagreement is normal rather than evidence the deletion was incomplete.

⚠️ And note what does *not* establish these deletions: the pin arithmetic. What does is the source diff (pure deletions of the four named arrays), emitted-stream occurrence counts showing no remaining reference by name, the per-case arithmetic-reachability audit ruling out reach-by-offset from each adjacent referenced base, `.text` and `.data` unchanged, and `check-region-map` passing — the check that actually owns this number.

PR #10990, opened 2026-07-30
codegen: share the EIP-7708 transfer-log operand staging (GH #10938 cut 4)

pirapira:
## Summary

GH #10938 **cut 4** — the nested post-descend block. Extracts the EIP-7708
`emit_transfer_log` operand staging into one emitter,
`eip7708TransferLogStageAsm` in `Programs/EIP7708Logs.lean`, shared by the value-CALL
frame entry and the CREATE endowment.

**Representation only, and byte-identity is MEASURED on both artefacts — including the
`.s` text.** Base `a714079a3`:

| artifact | before | after |
|---|---|---|
| `stateless_guest.s` | `4f690273e75b3644333f9a12c5934cdc8704ca70a516f6422034f554887ed0ae` | **identical** |
| `stateless_guest.elf` | `acd31bd5afb1149370990b8358c42c6a8a22d62dfab45b092d2d0ea991291a2d` | **identical** |

`.s` identity is the strong form here: it covers **all ten** emissions of the block at
once (eight CALL, two CREATE), not a spot check. No generated file changes, so no layout
regen — `scripts/check-region-map.sh` passes unchanged.

## The spec position

`execution-specs` at pinned `e5a8caf1b`, `vm/interpreter.py:391-397` — **one**
`emit_transfer_log`, inside `process_message`, after the `copy_tx_state` snapshot
(`:380`) and paired with the unguarded `move_ether` above it:

```python
if message.should_transfer_value and message.value != 0:
    move_ether(tx_state, message.caller, message.current_target, message.value)
    if message.caller != message.current_target:
        emit_transfer_log(evm, message.caller, message.current_target, message.value)
```

`process_create_message` has no counterpart of its own — it calls `process_message`
(`:212`), so the CREATE endowment log is *the same call site* in the spec. Two guest
copies of the staging is exactly the shape #10938 is about.

## The population, measured rather than inferred

`jal ra, eip7708_append_transfer_log` occurs **14** times in
`gen-out/regionmap/stateless_guest.s`. Clustering the 23 lines preceding each call and
normalising label names gives **six** distinct shapes, of which two are the duplicated
staging:

| shape | emissions | producer |
|---|---|---|
| `cd_xfer_log_pending` guard + staging | **8** | `ChildFrameHandlers.emitPendingXferLog` (2 sites × 4 closures) |
| ctx gate + `value != 0` + staging | **2** | `ChildFrameCreateTail` (salt / no-salt) |
| 4 others | 1 each | `BlockVerdictFunction`, `EIP7708Logs` (burn), `ChildFrameHandlerTailHelpers`, `Selfdestruct` |

**The CALL side was already factored** into one emitter; the CREATE side was a second
copy of it. So cut 4 is one deduplication, not eight.

`ChildFrameHandlerTailHelpers.lean:165-182` deliberately stays out: the value-bearing
precompile tail passes **raw stack-word pointers** (`x20`, `x12+32`, `x12+valueOff`)
straight to the appender and does no staging at all, because its operands are already
stack words. It is not a copy — it is the site that does not need the helper.

## The discriminator cut 1 lacked

Cut 1 was withdrawn because comparing the two deploy-charge blocks at three levels found
something new each time: same constants, four differences in form, and then **different
control flow** (opposite branch senses, two labels versus three), so no shared emitter
could reproduce both sites without changing one site's instructions.

I ran that comparison first here, on the **emitted stream** rather than the Lean source.
Extracting both bodies from `stateless_guest.s`, normalising labels, and renaming the
three source globals (`cd_caller_be`/`cd_callee_be`/`cd_value_be` →
`create_sender_be`/`create_address_be`/`create_value_be`) makes them differ by **exactly
one line**: a label the CALL side emits and the CREATE side does not, which contributes
**zero bytes**. That is what made the extraction viable and is why the emitter takes an
optional `restoreLabel` rather than always emitting one.

## What is deliberately NOT shared

Each caller keeps its **own guard**, because the two are different predicates over
different state:

* CALL — a one-shot `cd_xfer_log_pending` flag, cleared *before* the emit.
* CREATE — an account-witness context gate (`584(x20)`) plus `value != 0`; the caller
  is a freshly derived address so the spec's `caller != current_target` test is
  statically true and needs no emission.

Folding those together would repeat cut 1's error in the other direction. The
plan's prohibition stands: **a shared capture must never imply a shared disposition.**
Nothing about placement moves either — both blocks already sat where the spec puts the
log, after the child snapshot and before the body, and both are still rolled back by
`frame_return` on a reverting child. Byte identity is the proof that no ordering changed.

## Home, and reachability

`Programs/EIP7708Logs.lean`, which already owns `eip7708_append_transfer_log` — the
staging helper for its arguments belongs beside it. It imports only `Rv64.Program` and
`Codegen.Layout`, both already in `ChildFrameCreateTail`'s cone, so the one added import
edge pulls in nothing new; `ChildFrameHandlers` reaches it transitively because it
already imports `ChildFrameCreateTail` (`:34`). No file approaches the 1500-line
`Programs` cap (`EIP7708Logs` 527, `ChildFrameHandlers` 1001, `ChildFrameCreateTail`
567).

## Verification

* Byte-identity measured on `.s` **and** `.elf` (table above).
* `lake build` — clean.
* `scripts/check-region-map.sh` — OK on all five lines; no generated file changed.
* `scripts/check-guarded-handler-bytes.sh` — fails, and **fails identically on the
  unmodified tree**: a local build cache holds `Callable.olean.hash` without
  `Callable.olean`. Pre-existing and environmental; the script is not referenced from
  `.github/workflows`. Same note as #10985.

## Note on issue numbering

Cuts 1–4 are defined on **#10938**. PR #10985 (cut 2) mistakenly cites #10784 in its
title, body, and three source comments; corrections are recorded as author comments on
#10985 and #10987, and a follow-up will fix the source text once those merge. This PR
cites #10938 throughout.

### DIFF
--- a/EvmAsm/Codegen/Programs/ChildFrameCreateTail.lean
+++ b/EvmAsm/Codegen/Programs/ChildFrameCreateTail.lean
@@ -9,6 +9,7 @@
 import EvmAsm.Codegen.Programs.EvmAccessGas
 import EvmAsm.Codegen.Programs.EvmMemoryGas
 import EvmAsm.Codegen.Programs.CreateRuntime
+import EvmAsm.Codegen.Programs.EIP7708Logs
 import EvmAsm.Codegen.Programs.CreateSameTxCollision
 import EvmAsm.Codegen.Programs.AmsterdamSystemTx
 import EvmAsm.Rv64.Program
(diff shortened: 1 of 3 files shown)


pirapira:
Labelling. **The strongest identity measured on this work so far: both artefacts reproduce, including the `.s` text** — and the diff touches three source files and **no generated files at all**, which corroborates it independently. Nothing emitted moved, so there is no pin to re-derive and no delta to explain.

⭐ `.s` identity is the strong form *here* specifically because it covers **all ten emissions of the block at once** rather than sampling one. For a change whose whole claim is "the same instructions, emitted from one place instead of two," textual identity across every site is the claim itself rather than evidence for it.

## What cut 4 turned out to be — the issue's own framing was wrong about its kind

#10938 defines cut 4 as *the nested post-descend block, same spec region as cut 3 at the other depth*, which reads as a **placement** change. **There is no placement change to make:** the nested transfer record and log already sit after `create_frame_descend` and before the initcode, with comments citing the rollback, and `record_message_value_transfer` is already the shared producer.

**What is actually duplicated is the log operand staging.** So cut 4 is a **deduplication, not a move** — which is also why the identity is total: nothing about ordering could have changed.

## The population was measured, not inferred

`jal eip7708_append_transfer_log` occurs **fourteen** times in the emitted stream. Clustering the 23 lines preceding each call and normalising labels yields **six distinct shapes**, of which **two** are the duplicated staging — eight emissions from `ChildFrameHandlers` (two sites in each of four closures) and two from `ChildFrameCreateTail` (the salt and no-salt forms).

⭐ **The CALL side was already factored; the CREATE side was a second copy of it.** So this is **one** deduplication rather than eight, and the other four shapes are genuinely different sites that stay put. That distinction only exists because the population was clustered rather than counted.

## ⭐⭐ Cut 1's failure was run as a pre-flight gate, and that is the reusable part

Before writing an emitter: extract both candidate bodies **from the emitted stream**, normalise labels, rename the differing source globals, and diff. Result here — **they differ by exactly one line**, a label the CALL side emits and the CREATE side does not, contributing **zero bytes**.

That is precisely the check cut 1 failed, where the two bodies had two labels versus three with the second branch in the **opposite sense**. So the two outcomes are now distinguishable *before* committing to an extraction:

| difference | remedy |
|---|---|
| operand locations, register choices, an **optional** label | **parameterise** — both sites reproduce textually |
| **control shape** — branch count, branch sense, fall-through versus jump | **do not extract** — the shared body would have to branch on behalf of its caller |

Here the emitter takes the restore label as **optional** rather than always emitting one, which is what keeps both sites byte-identical instead of forcing one shape onto the other.

⛔ Worth stating as procedure rather than as a happy outcome: **cut 1 was withdrawn after an emitter was scoped; cut 4 ran the same check first and it cost one diff.** The gate is cheap and it is the difference between a landed deduplication and a withdrawn one.

PR #10991, opened 2026-07-30
codegen: name the body-state slab's stride and size (GH #10619 prerequisite)

pirapira:
## Summary

Prerequisite for GH #10619's one-field slab extension: give
`body_state_snapshot_by_depth` a **named** stride and size, so the field addition is an
edit to one number rather than to nine hand-written sites.

**Representation only. Byte-identity MEASURED on both artefacts.** Base `a714079a3`:

| artifact | before | after |
|---|---|---|
| `stateless_guest.s` | `4f690273e75b3644333f9a12c5934cdc8704ca70a516f6422034f554887ed0ae` | **identical** |
| `stateless_guest.elf` | `acd31bd5afb1149370990b8358c42c6a8a22d62dfab45b092d2d0ea991291a2d` | **identical** |

No generated file changes, so no layout regen and no interaction with the
one-regenerating-PR-at-a-time rule. `scripts/check-region-map.sh` passes unchanged.

## What was unnamed

The slab is the guest's analogue of `copy_tx_state` — one `process_message` body
checkpoint per frame depth. Its per-depth stride of 104 bytes **appeared nowhere as a
number.** All eight consumers open-coded `depth * 104` as

```
slli acc, depth, 6 ; slli tmp, depth, 5 ; add acc, acc, tmp ; slli tmp, depth, 3 ; add acc, acc, tmp
```

so the thirteen-field count existed **only as the exponents of three `slli`s**, spread
over three files, with the total appearing once more as a bare `.zero 106600` in a
fourth. Sites: `CallFrameReturn.lean` ×5, `CallFrameDescend.lean` ×2,
`BlockVerdictDataSectionTail.lean` ×1 (the allocation).

**Why that is a tripwire-absence rather than untidiness:** adding a fourteenth field
means changing all nine, in a form that no search for `104`, for the slab's name, or for
"stride" can enumerate. Miss one and every *deeper* frame's snapshot shifts by 8 bytes —
a **silently wrong rollback**, not a build error, and one that only shows up at depth ≥ 1
in whichever field happens to land on the seam.

## What this adds

In the existing import-free `Programs/BodyStateSnapshot.lean`, which already owns
`bodyStateCaptureScalarAsm`/`bodyStateCaptureCursorsAsm` — so this is the module's own
subject, and the two files that needed the import (`CallFrameReturn`,
`BlockVerdictDataSectionTail`) gain an edge into a module that imports nothing.

* `bodyStateSlabFields = 13`, with the field list in offset order in the docstring
* `bodyStateSlabStride = fields * 8`, `bodyStateSlabDepths = 1025`,
  `bodyStateSlabBytes = stride * depths`
* `bodyStateSlabStrideOps depth acc tmp` — the shift-add chain, derived from the set bits
  of `bodyStateSlabStride` rather than hardcoded
* `#guard`s pinning `stride = 104`, `bytes = 106600`, and **the rendered text for both
  register conventions** in use

## ⛔ The part that was measured rather than assumed

My first attempt had the emitter return its own indented line. That is **not**
byte-neutral, and both hashes changed:

```
- la t1, body_state_snapshot_by_depth; slli t2, s8, 6; … ; add t1, t1, t2
+ slli t2, s8, 6; … ; add t2, t2, t3
+ la t1, body_state_snapshot_by_depth; add t1, t1, t2
```

Diffing the two emitted streams showed **every difference was a line split** — same
mnemonics, same operands, same order *within* each site — except that at
`CallFrameDescend`'s two sites the original line **begins** with the `la`, so hoisting the
chain onto its own line moved the `la` from before the chain to after it. The computed
address is identical; the **instruction order** is not, and the linked bytes follow the
order.

Hence `bodyStateSlabStrideOps` returns no indentation and no trailing newline, and each
site splices it into the line it already occupies. That preserves the `.s` **text**, which
is why the identity above is exact rather than ELF-only.

The general form, worth stating because it is easy to get wrong in the other direction: an
extraction that changes only *line structure* can still change instruction order whenever
the original line **starts** with an instruction the extracted block must follow. `.s`
identity catches it; an ELF-only check would have caught it here too, but a
"same instructions, different formatting" argument would not have.

## Verification

* Byte-identity measured on `.s` and `.elf` (table above).
* `scripts/check-region-map.sh` — OK on all five lines, no generated file changed.
* `#guard`s cover the stride, the total, and both rendered register conventions.

## Next

GH #10619 itself — capture `storage_writes_undo_count` at slot 104 and replay it in
`dispatcher_restore_body_state` — follows as a separate PR on top of this one. That one
is behavioural and grows `.bss`, so per the review split it is held for review rather
than landed on the representation-only clause. It must land **before** #10784.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdictDataSectionTail.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictDataSectionTail.lean
@@ -7,6 +7,7 @@
 
 import EvmAsm.Codegen.Programs.BlockVerdictParams
 import EvmAsm.Codegen.CallFrameLayout
+import EvmAsm.Codegen.Programs.BodyStateSnapshot
 import EvmAsm.Codegen.Programs.NonstorageEffectLog
 import EvmAsm.Codegen.Programs.AccountTupleSequencesConsistent
 import EvmAsm.Codegen.Programs.BalSlotTupleSequence
@@ -488,7 +489,7 @@ def ziskStatelessVerdictV2DataSectionTail : String :=
   -- `Layout.moveZeroDataLines`, preserving fixed `.data` addresses (in
   -- particular bnf_le_a).
   ".balign 8\n" ++
-  "body_state_snapshot_by_depth:\n  .zero 106600\n" ++
+  "body_state_snapshot_by_depth:\n  .zero " ++ toString bodyStateSlabBytes ++ "\n" ++
   -- bmvmx.5.5.2 (umbrella-B1): scratch for the multi-tx per-sender FINAL-nonce check
   -- (BAL sender post nonce == pre + total sender tx count). bv_b1_finals is the 88-byte
   -- bal_account_nonstorage_finals output (separate from c2nsc_finals, which A2a's
--- a/EvmAsm/Codegen/Programs/BodyStateSnapshot.lean
+++ b/EvmAsm/Codegen/Programs/BodyStateSnapshot.lean
@@ -5,6 +5,13 @@
   produce the existing straight-line instructions; they are deliberately not
   guest subroutines, so root and child capture retain their exact instruction
   order and rollback timing.
+
+  GH #10619 also gives the slab's SHAPE a name here.  Its per-depth stride was written
+  nowhere: all eight consumers open-coded `depth * 104` as a shift-add chain, so the field
+  count existed only as the exponents of three `slli`s spread over three files, with the
+  total appearing once more as a bare `.zero 106600`.  Adding a field meant editing nine
+  sites in a form no search for "104" can find, and getting one wrong shifts every deeper
+  frame's snapshot by 8 bytes — a silently wrong rollback, not a build error.
 -/
 
 namespace EvmAsm.Codegen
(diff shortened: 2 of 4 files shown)


pirapira:
Labelling. Byte-identity measured on **both** artefacts, four source files, **no generated file touched** — so it also does not contend with the one-regenerating-PR-at-a-time rule and can land alongside anything.

## ⭐⭐ This is the least discoverable form of a pattern that has now appeared four times tonight

The pattern is *a constant that emission sites do not reference by name*. Its variants, in increasing order of how hard they are to find:

| form | discoverable by |
|---|---|
| a bare literal duplicating a named constant (#10971, #10980) | grepping the **value** |
| a named constant used for the **wrong quantity** (#10971's delete bound sized by the created capacity) | reading the declaration |
| **a value that was never named at all, encoded as shift exponents** — *this one* | **nothing** |

The slab's per-depth stride was written **nowhere**: all eight consumers open-coded `depth × 104` as a shift-add chain, so the thirteen-field count existed only as the *exponents of three shifts*. `104 = (d<<6)+(d<<5)+(d<<3)`. **No search for "104" finds any of those sites.**

⭐ And that is exactly why this is a **prerequisite** rather than a tidy-up. Extending the record from 13 fields to 14 changes the stride to 112 = `(d<<6)+(d<<5)+(d<<4)` — **one shift exponent, in eight places, none of them greppable**, with a failure mode that is silent and depth-dependent: get one wrong and every deeper frame's slot shifts. #10619 would have been an eight-site edit performed by inspection; it is now a one-line change to `bodyStateSlabFields`.

## It applies #10980's own remedy to itself

```
def bodyStateSlabFields : Nat := 13
def bodyStateSlabStride : Nat := bodyStateSlabFields * 8
def bodyStateSlabDepths : Nat := 1025
def bodyStateSlabBytes  : Nat := bodyStateSlabStride * bodyStateSlabDepths
#guard bodyStateSlabStride = 104
```

⭐ The `#guard` is the part that makes the naming load-bearing rather than decorative, and it is precisely the mitigation #10980 identified: **either use the name at the site or pin the literal to the named expression.** Here both are done — the sites now derive from the name, *and* the derived value is pinned, so a change to the field count that forgets a consumer fails the build instead of shifting slots at depth.

That also settles #10980's severity question for this constant in the safe direction: it now has a tripwire, so it moves into the same category as the `183600` instance rather than the untripwired `1530` deploy sites.

## Note on what this does not do

It names the shape; it does not change it. The record is still 13 fields, `storage_writes_undo_count` still has no tx-scope mark, and the ordering constraint stands — **#10619 must land before #10784**, or removing that commit predicate lets a failed transaction's in-place storage edits reach block state.

PR #10992, opened 2026-07-30
refactor(codegen): route ordinary CALL through child entry

pirapira:
## Summary

Supersedes the mis-positioned child-entry seam from #10969. The old .runtime_tx_child_message_entry label was a root-body alias: a descended child entering there would repeat the root RA restore and global stack/memory reinitialisation. This moves that one label to immediately before .Ldispatch_resume, below all root-only setup.

The sole ordinary, non-precompile CALL fall-through now jumps to the moved entry. The child therefore resumes at the same dispatcher-resume sequence that dispatchContinueRet previously reached after return, without replaying root capture/value seeding or stack/memory bootstrap. Precompile and CREATE routes remain unchanged, and the existing presnap/rollback timing remains intact.

This is the first actual child caller of the shared executor seam.

## Static evidence

- The moved .runtime_tx_child_message_entry label has exactly one static ordinary-CALL jump.
- The emitted child entry begins with the existing dispatcher-resume sequence.
- Text shrinks by exactly 8 bytes: dispatchContinueRet lowers to auipc + addi + ret (12 bytes), replaced here by one j (4 bytes). Moving the label itself emits no bytes.
- .data remains 0x5370, .bss remains 0x1c103000, and bnf_le_a remains 0xa3000ee0.
- Full cache-free lake build, asm-to-program, region-map, build-units-link, and memorylayout checks passed.
- check-guarded-handler-bytes is deferred to CI: the known artifact-cache limitation satisfies the graph without materialising the required olean for lake env lean.

## Behavioral check

Paired, immutable-ELF Spike run on 200 uniformly sampled manifest rows: the base and candidate TSVs are byte-identical, with zero per-row transitions. Both legs: FA=0, FR=94, OK=106; all 12 expected-invalid sampled rows still reject.

Provenance: manifest /var/tmp/fc668/manifest.tsv (26,104 rows; sha256 b40b7e3f7d4b3bde03c7298c98d3b9117ea146f76c2a2a914f1cbafa78c3ad07); captured seed 385927080; selected rows sha256 eae2bd8dc41abb1e45c742301c17e60ad375791dc2a0f3749fdcad8ddce51b85; ordered input-sha-column digest 3f4068ee70a9e12645f242fd7bc3ce56361cd99c05f217143596d6f6c0eb95a3; frozen runner c6ac09fa72c2103b6d3e390b779b022ef7f220e5a39bb5f667114a89d78e5868; base ELF acd31bd5afb1149370990b8358c42c6a8a22d62dfab45b092d2d0ea991291a2d; candidate ELF d3b6cad53f6a908646761bf0445d0dd96f20ed74241bdf91dc4328b169daf993; 12 workers; rc=0 for every row.


pirapira:
Labelling. **This is the first actual deduplication in this sequence: the shared executor now has a real child caller.** Everything before it built the substrate — a canonical snapshot structure, shared source primitives, named constants, aligned publication timing, and a seam — without anything yet flowing through one path.

## What makes the evidence unusually strong here

⭐ **The two legs produced byte-identical result TSVs — zero rows changed outcome.** That is stronger than the FA=0 gate that was asked for: FA=0 with FR=94/OK=106 says *no net regression*, while identical TSVs say *no change at all*, which for a pure routing change is the claim itself rather than evidence for it. Lead with that reading.

⭐ **The `-8` byte delta is accounted at the instruction level**, which matters because a bare "−8" is indistinguishable from an unexplained remainder: `dispatchContinueRet`'s `la x1, .Ldispatch_resume` lowers to `auipc` + `addi` + `ret` — three instructions, 12 bytes — replaced by a single `j` at 4. Net −8 exactly. (The reviewer's recollection of "about five instructions" was wrong; the lowering is three. Reading the emitted form beat recalling the shape, for the fourth time in this sequence.)

⭐ And the provenance is the most reproducible of the run: manifest sha, **an unobserved seed captured rather than specified**, selected-rows sha, an **ordered input-sha column digest**, both ELF shas, worker count, rc for every row, and twelve expected-invalid rows sampled and still rejecting. The column digest is the part that lets someone prove they ran *the same* 200 rows rather than *some* 200 rows.

## The #10969 condition, and why it was a real condition

⛔ #10969 landed a seam label whose own comment called it *the routing seam for a later semantic migration* — and the design work for this PR established **that label could not serve that purpose**. It aliased the root body, so a child reaching it would hit the `ra` restore from the root's wrapper slot and the global stack/memory reinitialisation that a descended child must not redo, `call_frame_descend` having already installed the child's registers and captured its snapshot.

⭐ **The label MOVED rather than being deleted and re-added** — from root setup to immediately before the dispatch resume — which is the better of the two acceptable outcomes: the mis-positioned seam is now correctly positioned, and label movement emits no bytes. Exactly one static ordinary-CALL jump reaches it, so the property inverts from #10969's *present-and-unreached* to **present-and-reached**, which is what this cut exists to establish.

That also means #10969's reviewer approval — mine — rested on a label whose stated purpose it could not fulfil. The seam was inert, so nothing shipped broken, but "unreached" was verified while "usable" was not.

## Scope held

Precompile tails and CREATE-family routes stay out, correctly: their dispositions are **not context-free executor outcomes** — precompile tails jump to their own dispatch label, and CREATE carries caller-owned collision, endowment, stack-result and creation-specific failure. And the CALL record is **not** relocated across its rollback boundary; the pre-snapshot protocol is untouched. This is routing only.

`check-guarded-handler-bytes` is deferred to CI with its cause stated — the artifact cache satisfies the build graph without materialising the required olean, a known filed limitation — which is the right form: a named deferral with a cause is a measurement, while a gate silently absent from the list reads as an oversight.

pirapira:
## Record correction: the figures in this PR body are from a superseded base

I pulled this PR's `merge!` label at 23:56 because its evidence was taken on the pre-#10990 base, then re-labelled it at 00:19 on a fresh current-head measurement. **That re-measurement was reported over the fleet channel and never written here, so this body is still the only record on the PR — and it is stale.** Recording it now so the merged history does not carry the wrong numbers.

**Superseded, in the body above:**
- `.bss remains 0x1c103000` — that was true on the pre-#10990 base. Main was `0x1c0fafe0` by the time this was labelled (#10986 removed 32 bytes of dead call-frame cells; #10988 retired the checkpoint arrays for `-0x8000`).
- `FA=0, FR=94, OK=106`, base ELF `acd31bd5…`, candidate `d3b6cad5…`, seed 385927080.

**Current, measured against merge base `4095e541a` (head `a1e3a7f98`):**
- candidate ELF `e96985d6e3f6b886bede8fb2f97513432108089cd1d2683511f39466fce84f54`, parent `be4676629d7e1eb1897aed7210b1d91441747497e998efb1a6f7bcc2fe943278`, frozen runner `c6ac09fa72c2103b6d3e390b779b022ef7f220e5a39bb5f667114a89d78e5868`
- sections: parent `.text 0x66be8` → candidate `0x66be0` (lowering, −8: `auipc`+`addi`+`ret`); `.data 0x5370` and `.bss 0x1c0fafe0` unchanged by this PR
- paired random-200, captured (not chosen) seed 7978592762157000746: **both legs FA=0, FR=105, OK=95**, complete per-row TSVs byte-identical, zero transitions
- liveness: exactly one direct `j .runtime_tx_child_message_entry` (emitted line 61101), entry present at 36217

The differing candidate/parent ELF hashes are what make the paired run valid — identical hashes would mean one binary measured twice, and equal counts across legs would then read as stability rather than as a broken experiment.

**The FR figure moved 94 → 105 with the base, not with this change.** Both legs agree at 105 and the per-row TSVs are byte-identical, so this PR moves no row; the earlier 94 was simply a different base's number. This is a `refactor` whose correct outcome is zero row movement.

**Post-merge pin note:** neither this PR's `textSizeBytes` (`0x066be0`) nor #10994's (`0x066c08`) is correct once both land — the combined value is `0x066c00`, derived twice independently (a discarded scratch merge, and the batch regen in #10995, computed blind to each other). Resolved by regeneration in #10995 rather than by choosing a side.

PR #10994, opened 2026-07-31
codegen: capture and replay storage_writes_undo_count in the tx-scope slab (GH #10619)

pirapira:
## Summary

GH #10619: add the **fourteenth** `body_state_snapshot_by_depth` field so the depth-0 body
rollback replays the storage-writes undo journal, exactly as it already replays
`account_writes_undo_count` (offset 64) and `create_nonce_undo_count` (offset 96).

`execution-specs` `restore_tx_state` (`state_tracker.py:823-826`, pinned `e5a8caf1b`)
restores **four** write structures, and `storage_writes` is one of them. The guest captured
two of the three marks it has and left storage to `write_sets_discard_tx` — a
**whole-transaction commit predicate**, not a body rollback. This PR does **not** remove
`write_sets_discard_tx`.

⛔ **Must land before #10784.** #10784 removes commit predicates; without this field, doing
that first lets a failed transaction's in-place storage edits reach block state.

## The change

* `bodyStateSlabFields` 13 → 14, so stride `104 → 112` and the slab `106,600 → 114,800`
  bytes. Both follow from the named constant (#10991) rather than from nine hand-edits.
* `dispatcher_capture_body_state` captures `storage_writes_undo_count` at offset **104**.
* `dispatcher_restore_body_state` replays it via `write_sets_restore_frame`, which already
  exists, already takes the mark in `a0`, walks the journal in **reverse** so nested
  overwrites land on the earliest recorded value, and resets the cursor to the mark.
  `CallFrameReturn` already uses it for child frames off the per-frame checkpoint array.

**Ordering answered by reading, not assumed:** the spec restores four *independent*
containers, and `write_sets_restore_frame` touches only the undo journal and
`tx_storage_writes_count`, so placing it after the account and nonce replays is safe. `t2`
is reloaded because the preceding call clobbers it.

## Layout — re-derived, not carried

Base `4095e541a`. `.text 0x066be8 → 0x066c08`, `.bss 0x1c0fafe0 → 0x1c0fcfe0`, `.data`
unmoved.

⚠️ **The `.bss` step is `0x2000`, not the slab's own `1025 * 8 = 0x2008`.** Derived from the
emitted addresses rather than by subtraction: `body_state_snapshot_by_depth` stays at
`0xbb3a5688` while its successor `b1sc_sort_a` moves `0xbb3bf700 → 0xbb3c1700`, because the
slab's end goes `0xbb3bf6f0 → 0xbb3c16f8` and **both round up to the same 32-byte
boundary**, cutting the padding from 16 bytes to 8. A naive subtraction predicts 8,200 and
is short by 8. That derivation is now in `RegionMap`'s docstring so it is not re-modelled.

This was measured **twice on two different bases** — #10986 and #10988 shrank `.bss` between
my attempts, so the absolute pins I first measured were already dead and only the delta
survived. Regen reached a fixed point (a second complete pass changed nothing);
`GuestImageEntries` came out unchanged; `GuestAddrs.bnf_le_a` is still `0xa3000ee0`;
`GuestImage.lean`'s handwritten `.bss` literals moved to `0x1c0fcfe0` with the extent end
`0xbf20afe0 → 0xbf20cfe0`.

## Fixture evidence — paired, with both guest hashes quoted

`scripts/codegen-eest-stateless-check.sh --backend spike --random --limit 200 --jobs 6`,
seed **captured, not chosen**: `2075889796`. Corpus `tests-zkevm@v0.6.2`.

| | this PR | control (`4095e541a`) |
|---|---|---|
| **guest sha256** | `a77de5ce196b27341b22b4f2c8eb82f10018b8991ce543fd4faf672d2a6f373d` | `b370eab00157d7c06e53afd4f0bd6fae7fccfe6f3e47d47b706474da37a676ce` |
| selected / ran | 200 / 200 | 200 / 200 |
| errored | 0 | 0 |
| fail | 102 | 102 |
| full match | 98 | 98 |
| root match | 200 | 200 |
| succ match | 98 | 98 |
| **failing label set** | — | **identical** (`diff` = 0 lines) |
| `bv_fail` histogram | 60:53, 62:35, 44:10, 53:3, 59:1 | identical |

**The two hashes differ, which is what makes this a paired measurement rather than one
binary measured twice.** I quote them because an earlier attempt of mine reported
`102 == 102` from two runs of the *same* ELF, and equal counts read as a stable result rather
than as a broken experiment. Failing *sets* are compared, not just counts, so an equal count
cannot hide a swap.

`102/200` is the **standing baseline on this seed**, not a regression — it is identical on
both legs, fixture-for-fixture and failure-code-for-failure-code.

## ⛔ The null result is expected, and it is the reason for the ordering constraint

This change is **outcome-neutral on the sample**, and that is the honest reading rather than
a disappointment: the new replay only matters for a transaction that fails after preparation
*and* wrote storage in its body, and `write_sets_discard_tx` still covers the commit boundary
today. So the mechanism is currently **redundant** — and after #10784 removes that predicate
it becomes the only thing standing between a failed transaction's in-place storage edits and
block state.

That is precisely why the ordering is on both issues, and why it is not derivable from either
diff.

## Verification

* `lake build` — clean. This is the gate that matters here: `GuestImage.lean`'s handwritten
  `.bss` literals had to move, and that failure mode presents as a bare `Type mismatch` with
  no mention of `.bss`.
* `scripts/check-region-map.sh` — OK on all five lines; regen at a fixed point.
* Paired fixture run above, both legs run **bare** (no pipe) with the raw exit status
  captured — a piped invocation reports the last stage's status, not the script's.

## Also in this PR: the evidence strings interpolate their pins

`RegionMap.lean`'s `.bss` evidence string hardcoded `0x1c105000` while `bssSizeBytes` is
`0x1c0fafe0` — **stale by three merged PRs** (#10979 shrank the pin and left the prose; #10986
and #10988 shrank it again), leaving it `0xa020` bytes wrong. Updating the digit would only
re-stale it on the next repin, so both ELF-measured strings now interpolate via `natToHex`.

The `.data` string was correct when this change was made and is interpolated **anyway**: a
literal duplicating a named value with no tripwire is the exposure whether or not it currently
agrees. The control is in the same file — the `.text` evidence never went stale across any of
these, precisely because it states **no number**.

Noted because my own change would otherwise have made the `.bss` string correct *by
coincidence*, erasing the staleness rather than fixing it.


pirapira:
## Address-derived reconciliation of the `.bss` pin, on the record

Read out of the two hashed ELFs this PR's fixture table cites — `a77de5ce…` (this PR) and
`b370eab0…` (control at `4095e541a`) — not from a rebuild.

| | control | this PR |
|---|---|---|
| `body_state_snapshot_by_depth` | `0xbb3a5688` | `0xbb3a5688` (**unmoved** — it precedes the growth) |
| slab size | 106,600 (`0x1a068`) | 114,800 (`0x1c070`) — **+8,200 = 1025 × 8** |
| slab end | `0xbb3bf6f0` | `0xbb3c16f8` |
| slab end **mod 32** | **16** | **24** |
| `b1sc_sort_a` (next symbol) | `0xbb3bf700` | `0xbb3c1700` — **+8,192 = 0x2000** |
| padding between them | **16 bytes** | **8 bytes** |
| `.bss` size | `0x1c0fafe0` | `0x1c0fcfe0` — **+0x2000** |

**8,200 grown − 8 absorbed = 8,192.** The absorbing symbol is **`b1sc_sort_a`**, in its own
`.balign 32` padding.

`b1sc_sort_a` is confirmed as the immediate successor two ways: it is the next entry in
sorted `nm` output in both ELFs, and the next label in the emitted stream
(`stateless_guest.s:78022`), with `.balign 8` then `.balign 32` between it and the slab.

The absorption happens **exactly once** and everything downstream then shifts uniformly —
`b1sc_sort_b` moves `0xbb409d60 → 0xbb40bd60`, also `0x2000`. That is why the section delta
equals the successor delta.

### ⛔ The discriminator, which matters more than this instance

**This is not a rule about growth; it is a fact about one offset.** The slab's end sits at
`16 mod 32` before and `24 mod 32` after — *both inside the same 32-byte block* — so
`b1sc_sort_a` rounds up to the same offset within its block either way and the extra 8 bytes
never become a new block.

Had the pre-change end sat at `8 mod 32` or lower, adding 8,200 would have crossed into the
next block and the pin would have moved the **full 8,200 or more**. So a future field
addition to this same slab can behave differently: the absorption is a property of 106,600
happening to end 16 bytes into a block, **not** of adding 8 bytes per depth.

**Any model that predicts this delta without the end-offset is underdetermined** — the total
`0x2000` is consistent with several structures, and only `end mod 32` picks one.

### Correction to my own claim, and a follow-up

I told coord I had put this *discriminator* into `RegionMap`'s docstring. **I overstated it.**
The docstring (`RegionMap.lean:478-486`) carries the addresses and says the ends "both round up
to the same 32-byte boundary" — the mechanism — but **not** the `16 mod 32` / `24 mod 32`
numbers and **not** the warning that a different offset behaves differently. That is the gap
this PR's docstring should close, and since the PR is labelled I am not pushing to it.

Proposed wording for a follow-up, to append to that docstring:

> The discriminator is the slab's end **mod 32**: `16` before and `24` after, both inside the
> same 32-byte block, so the successor's `.balign 32` rounds to the same place and the extra
> 8 bytes never become a new block. Had the end sat at `8 mod 32` or lower, the same +8,200
> would have crossed into the next block and moved this pin by the full amount. Re-derive from
> addresses on every shape change; the end offset, not the growth, decides.

Issue #10996, opened 2026-07-31
BAL: CALLCODE, DELEGATECALL, and STATICCALL EIP-7708 transfer-log cases share a three-byte shortfall


pirapira:
Source-audit/measurement finding, not an attributed mechanism.

On guest head a1e3a7f9892e8ec7ad51da52ea3abca00c892349 (ELF sha256 `e96985d6e3f6b886bede8fb2f97513432108089cd1d2683511f39466fce84f54`) with `spike_run` sha256 `c6ac09fa72c2103b6d3e390b779b022ef7f220e5a39bb5f667114a89d78e5868`, all three valid EIP-7708 `call_opcodes_transfer_log_behavior` variants false-reject:

- `00183` CALLCODE
- `00184` DELEGATECALL
- `00185` STATICCALL

Each returns `bv_fail_code=60` (BAL digest mismatch), `shadow_status=1`, rebuilt length 365 versus supplied length 368: a shared -3-byte shortfall. All three have oracle success 1.

This is a structural class, not three independent fixture outcomes. The sibling ordinary CALL variant `00182` also fails, but at equal length (367/367), so it is not evidence for this three-byte mechanism.

`00183`/`00184`/`00185` are therefore unsuitable as green route controls for child-entry routing until this independent shortfall is resolved. Green alternatives measured on the same artifact are 00311 (CALLCODE), 00312 (DELEGATECALL), 00316 (STATICCALL), plus 00188 (CALLCODE self/no-log).

Next diagnostic: capture/rebuild the declared and emitted BAL streams for all three, length-assert first, then decode per address/field to identify the common missing RLP structure.

pirapira:
[...]

PR #10997, opened 2026-07-31
fix(codegen): resume DELEGATECALL and STATICCALL child frames

pirapira:
## Summary

Route STATICCALL (mode 1) and DELEGATECALL (mode 3), like ordinary CALL, to `.runtime_tx_child_message_entry` after descent. CALLCODE (mode 2) keeps its distinct value-bearing/balance-gated return route.

This follows pinned execution-specs `e5a8caf1b`: [delegatecall](https://github.com/ethereum/execution-specs/blob/e5a8caf1b/src/ethereum/forks/amsterdam/vm/instructions/system.py#L702-L798) and [staticcall](https://github.com/ethereum/execution-specs/blob/e5a8caf1b/src/ethereum/forks/amsterdam/vm/instructions/system.py#L801-L897) both enter the shared [generic_call](https://github.com/ethereum/execution-specs/blob/e5a8caf1b/src/ethereum/forks/amsterdam/vm/instructions/system.py#L330-L393), which constructs a child message and invokes `process_message`.

## Emitted-descriptor check

Candidate ELF `7e864adf…` confirms the mode-specific descriptors and the new common entry:

- DELEGATECALL: `li t3, 3` / `sd t3, 16(cd_desc)` at `0x8005de18`; `netPopBytes=160` at `0x8005de40`; `value_nonzero=0` at `0x8005de58`; `jal call_frame_descend` at `0x8005de68`; then `j .runtime_tx_child_message_entry` at `0x8005df5c`.
- STATICCALL: `li t3, 1` / `sd t3, 16(cd_desc)` at `0x80061eac`; `netPopBytes=160` at `0x80061ed4`; `value_nonzero=0` at `0x80061eec`; `jal call_frame_descend` at `0x80061efc`; then `j .runtime_tx_child_message_entry` at `0x80061ff0`.

## Evidence

- Base: `987577050e29648927ebfb8e776a1f8e62629851`
- Parent ELF SHA-256: `1935e5700e7669e8fd15be0621f7995886a9a9c3d071d489c1aa37f877189862`
- Candidate ELF SHA-256: `7e864adfcfe0c79fa0683f16bd415aa2ef9001035c2596c5d1097a847c524893`
- Frozen runner SHA-256: `c6ac09fa72c2103b6d3e390b779b022ef7f220e5a39bb5f667114a89d78e5868`
- Direct controls: `00188`, `00311`, `00312`, and `00316` are all oracle-valid with `fail=0`, `shadow_status=0`, and rebuilt BAL length equal to supplied length, before and after the change. `shadow_status=0` is a completed BAL hash match, not a skipped comparator.
- `00183`, `00184`, and `00185` are intentionally excluded: their shared code-60 transfer-log BAL mismatch is tracked separately in [#10996](https://github.com/Verified-zkEVM/evm-asm/issues/10996).
- Paired random-200 at seed `13646251995594346639`: both legs `FA=0 FR=105 OK=95`; joined=200, forward FR→OK=0, reverse OK→FR=0. The TSVs are byte-identical (SHA-256 `7d3215b242e1a259e3f6a465e62ae8dcb88c78c4ebaffa2f7d4655870ba155a8`).

## Regeneration and checks

Two complete regeneration passes converged. Final ELF layout: `.text=0x066bf0`, `.data=0x5370` (unchanged), `.bss=0x1c0fcfe0` (unchanged); `bnf_le_a=0xa3000ee0` remains pinned.

- `lake build`


pirapira:
Not labelling yet. **The code looks right; the body is wrong on the one axis this change is risky on**, and two gate items are missing. None of this is a rebuild — it is the write-up.

## 1. The mode numbers in the summary are inverted

The body says "DELEGATECALL (mode 1) and STATICCALL (mode 3)". The guest's own contract says the opposite, in `CallFrameDescend.lean` (the `a4` docstring):

> `0 = CALL`, `1 = STATICCALL`, `2 = CALLCODE`, `3 = DELEGATECALL`

and the emitted arms agree with that table, not with the summary: CALLVALUE is zeroed for **mode 1**, inherited from the parent for **mode 3**, and CALLER inherits `parent.CALLER` only at **mode 3**. Your own earlier audit also had it the correct way round (`STATICCALL … mode=1`, `DELEGATECALL … mode=3`).

**The diff itself is mode-agnostic** — `if mode == 0 || mode == 1 || mode == 3` covers both regardless of which name attaches to which number — so I read this as a prose error with no behavioural consequence, and the green controls are consistent with that. But it is a prose error *in the permanent record of the one change whose entire failure mode is mode confusion*, and the next reader will reasonably trust the summary over the docstring. Please fix the summary.

## 2. The spec citation is SpecRef again

The body cites `Interpreter.lean:648-685` and `:816-875`. **SpecRef is a port and it drifts**; execution-specs at `e5a8caf1b` is the authority, and SpecRef agreeing is corroboration rather than justification. Cite the Python — for this change the relevant sites are `vm/instructions/system.py` `staticcall` (`:801`) and `delegatecall` (`:702`) and their shared `generic_call`. Keep the SpecRef links if you like, but the load-bearing citation should be the pin. `CallFrameDescend.lean`'s own docstring already cites `vm/instructions/system.py` directly, so the convention is established in the file you are modifying.

## 3. Two required gate items are absent

- **Per-route descriptor disassembly after the change.** The whole risk of collapsing two modes into one route is a mode swap, and the four green controls cannot see it: 00311/00312/00316 are `account_access_target` fixtures, and touched-address behaviour is computed *identically* in all three modes. So show, from the emitted stream, that each route still builds its own descriptor — mode 1 with netpop 160, mode 3 with netpop 160, mode 2 with netpop 192 and `value_nonzero` at `+64`. This is the only evidence in the set that discriminates the thing being changed. Given item 1, it is now doubly worth printing.
- **00183 / 00184 / 00185 named individually**, with their pre-existing red status and #10996 cited. A zero-transitions total can hide a compensating pair; three named rows cannot.

## What is already right, so it is not re-litigated

Base and both ELF hashes given and **differing**, frozen runner pinned, seed `13646251995594346639` captured, both legs `FA=0 FR=105 OK=95` with byte-identical TSVs and zero transitions in both directions — the correct outcome for a routing refactor. `shadow_status=0` established as a completed hash match rather than a skipped comparator, with the rebuilt/supplied length pair corroborating engagement independently of that field. Two regeneration passes converged; `.data` and `.bss` unchanged with `.text` at `0x066bf0`.

Issue #10959, opened 2026-07-30
check-build-parallel cannot report its own coverage: 4 of 9 fanned-out scripts exit 0 on a missing toolchain, and only 1 prints a set size


pirapira:
## Claim

**`check-build-parallel.sh` cannot report its own coverage.** Four of the nine scripts it fans out to `exit 0` on a missing toolchain, and only one prints a set size — so a green result is compatible with four scripts having done nothing, and the gate's own output cannot distinguish the two.

Found by `scoper` while answering a narrower question (why the gate started passing). Source reading; toolchain observations scoped to one box, see below.

## What it is

**Not a check — a 58-line fan-out wrapper** running **nine scripts in five lanes**:

| lane | scripts |
|---|---|
| codegen | `codegen-stateless-link-check --no-build`, `check-build-units-link`, `check-region-map`, `check-memorylayout-region-coverage`, `check-guarded-handler-bytes` |
| asm-to-program | `check-asm-to-program` |
| reports | `check-progress`, `check-drift` |
| axioms | `check-axioms` |
| arithmetic-fuzz | `fuzz-arith-diff` |

The wrapper measures nothing itself: one PASS/FAIL line per lane, plus each lane's log.

## ⭐ The one piece of good news, and it is structural rather than inferred

`codegen_checks` runs its five scripts **sequentially in a subshell inheriting `set -e`**, so **the first failure aborts the lane and the rest never run.**

Therefore **a lane flipping FAIL → PASS means strictly more scripts executed, not fewer.** While the codegen lane was red, whichever script failed first was *hiding* the four after it. That answers the coverage question in the safe direction — from the wrapper's own control flow, not from an assumption.

## ⛔ Four silent-skip paths, exact strings, because these are the denominator

- `check-build-units-link:91-93` — `SKIP — riscv64-unknown-elf toolchain not on PATH`
- `check-region-map:36-38` — `riscv64 cross toolchain not found; skipping (install to enable)`, **plus a second skip at `:42` for `readelf`**
- `check-guarded-handler-bytes:11-13` — `riscv64-unknown-elf-as not found; skipping (install to enable)`
- `check-asm-to-program:23-25` — the already-known one

**So the asm-to-program lane is entirely skippable and the codegen lane loses three of its five.** Two lanes green can mean four scripts did nothing.

## ⛔ The real finding: the denominator is unreportable by construction

[...]

PR #10999, opened 2026-07-31
fix(codegen): route CALLCODE through shared child entry

pirapira:
## Spec alignment\n\nFor CALLCODE, execution-specs resolves the target as the current account, applies its distinct insufficient-balance failure before descent, then enters generic_call on a successful descent. The guest already retained that separate pre-descent gate at .Lcd_insuffbal_callcode_target; this change only routes mode 2's post-descent tail through the same child-message entry as modes 0, 1, and 3.\n\nThis completes executor-tail unification for all four CALL-family modes: #10992 CALL, #10997 DELEGATECALL/STATICCALL, and this PR CALLCODE. Sources: system.py:520 (CALLCODE), :540 (to == current_target), :605 (balance gate), and :611 (generic_call). The distinct insufficient-balance path refunds both pools at :608-609 and remains unchanged.\n\n## Mechanism\n\ncallDescendFallThrough now includes mode 2. The emitted CALLCODE tail retains EIP-7708 log staging and jumps to .runtime_tx_child_message_entry, exactly where the old dispatchContinueRet returned (.Ldispatch_resume). The separate insufficient-balance arm remains before .Lcd_descend_callcode_target.\n\nThe descriptor still emits mode 2 at cd_desc + 16, netPopBytes = 192 at +56, and derived value_nonzero at +88 (the value source is parent x12 + 64).\n\n## Layout prediction and regeneration\n\nOne converted tail replaces la (AUIPC+ADDI, 8 bytes) plus ret (4 bytes) with a 4-byte j, predicting -8 bytes without an alignment-boundary crossing. Measured candidate: .text=0x066be8, .data=0x5370 (bnf_le_a=0xa3000ee0 unchanged), .bss=0x1c0fcfe0, GuestImage end 0xbf20cfe0.\n\n- parent ELF SHA-256: 7e864adfcfe0c79fa0683f16bd415aa2ef9001035c2596c5d1097a847c524893\n- candidate ELF SHA-256: 7df7e822d803caafb2533ea1013a91e137dc888fcc097ff8363c08007c54b423\n\n## Verification\n\n- check-guest-addrs and check-asm-to-program passed.\n- Full check-region-map.sh (without --no-build) passed its emitted-reality and link-layout checks; it reported RegionMap.textSizeBytes=66be8, dataSizeBytes=5370, bssSizeBytes=1c0fcfe0, and a matching symbol-addresses.tsv.\n- check-build-parallel completed successfully. Per-file cache validation found 0 .olean.hash traces lacking their corresponding .olean, so Lake did not skip a trace-only module. Its coverage-reporting limitation remains tracked separately by #10959.\n- Targeted controls with frozen Spike runner c6ac09…: 00311 and 00188 are parent/candidate byte-identical and accepted; expected-invalid 04778 still rejects.\n- Reproducible, user-specified random A/B seed 2026073101, 200 uniformly selected blocks, same runner and frozen artifacts. Parent and candidate are each FA=0, FR=102, OK=98; 200/200 joined outputs are byte-identical with no ERROR or BUDGET rows.\n\nThe random sample did not contain a verdict transition; the emitted mode-2 tail plus focused CALLCODE controls establish that the new route is wired.\n


pirapira:
Not labelling yet — three questions about the **verification**, none about the change. The change itself reads correct and the layout evidence is the strongest we have produced on this series.

## What is already right, so it is not re-litigated

**The pin was predicted from first principles before the build and matched exactly.** One converted tail replaces `la` (`AUIPC`+`ADDI`, 8) plus `ret` (4) with a 4-byte `j` ⇒ −8. Across the series: #10992 one tail −8, #10997 **two** tails −16, this one −8, giving `.text 0x066be8`. Putting that derivation in the body means a reviewer can recompute the pin from the diff without trusting any regen — that is better evidence than two agreeing measurements, because it *explains* the value rather than reproducing it.

Also correct: differing parent/candidate ELF hashes; the Python cited at the pin (`system.py:520`, `:540`, `:605`, `:611`, and `:608-609` for the both-pools refund); descriptor facts with `value_nonzero` at **+88** and the value source at parent `x12+64`; the pre-descent gate still preceding `.Lcd_descend_callcode_target`; and `04778` still rejecting.

## 1. The seed looks chosen, not captured

`seed 2026073101` is date-shaped (2026-07-31, 01). Every other run in this series carried a captured seed — `7978592762157000746`, `13646251995594346639`, `2075889796`. **Was this printed by the runner and captured, or supplied?**

If supplied, say so in the body rather than describing it as captured. It is **not fatal here**: the A/B's *validity* rests on the differing ELF hashes and the byte-identical rows, neither of which depends on how the sample was drawn. But the sample's *representativeness* does, and "captured, never specified" has been the standing rule all evening — a chosen seed silently converts a random sample into a fixed one.

## 2. `check-region-map.sh --no-build` can skip its emitted-reality checks and still pass

From the script itself, line 91:

```
note "SKIP emitted-reality (.s) checks: $GUEST_S absent (pass without --no-build to emit it)"
```

So under `--no-build`, if `$GUEST_S` is absent the `.s` checks are **skipped and the script still exits 0** — a clean gate measuring irrelevance rather than safety. Your artefacts were freshly built, so the `.s` may well have been present and the checks may have run in full.

**Did the run emit that SKIP line?** If it did, the region-map check is weaker than the body implies and should be re-run without `--no-build`. If it did not, say so and this closes.

## 3. The aggregate parallel check's status is ambiguous — and that harness has a known coverage defect

> "The aggregate parallel check otherwise reached link-layout validation, but its helper initially lacked a cache-materialized `GuardedHandlerSpecs.olean`."

"Reached link-layout validation" does not say whether it **passed**. Please state the outcome plainly. Two reasons this needs to be explicit rather than narrated:

- **#10959** is open for exactly this harness — `check-build-parallel` cannot report its own coverage; 4 of 9 fanned-out scripts exit 0 regardless. So "the aggregate check was fine" is not something it can establish about itself.
- A missing olean with its trace present makes `lake` print `Fetched`, compile **nothing**, and exit 0 — so an olean-absent run carries no information. Worth confirming with `find .lake/build -name '*.olean' | wc -l` against `'*.olean.hash' | wc -l`; if those diverge, local build evidence is void (not wrong — void).

You state the guarded-handler helper passed *after* materializing the olean, which is the right fix. I want the aggregate's verdict named the same way.

## Note, not a request

The random sample contained no verdict transition, which you said plainly — good. For a routing refactor that is the expected outcome, and the wiring evidence is the emitted mode-2 tail plus the focused controls, exactly as you argued. I am not asking for a wider sample.

Issue #11001, opened 2026-07-31
codegen: a successful child DESTROYS its own live-balance debit mark, so a later parent revert strands the debit permanently (filed as debit-before-snapshot; that was not the mechanism)


pirapira:
## Summary

The guest performs the value-CALL **balance mutation before the child snapshot** and its **BAL
record after it**. execution-specs puts **both after** the snapshot. So on a reverting child the
record is dropped (correct) while the caller's live balance stays debited (the spec restores it).

Structural divergence **established** below. The witnessing fixture is **not** established — see
"What is not established".

## The spec: snapshot precedes both

`execution-specs` at pinned `e5a8caf1b`, `src/ethereum/forks/amsterdam/vm/interpreter.py`:

```python
snapshot = copy_tx_state(tx_state)          # :380
try:
    if message.should_transfer_value and message.value != 0:   # :384
        move_ether(tx_state, message.caller, message.current_target, message.value)   # :385-390
        if message.caller != message.current_target:
            emit_transfer_log(evm, ...)      # :391-397
```

`move_ether` mutates `tx_state`, which the snapshot covers, so an error path's
`restore_tx_state(tx_state, snapshot)` puts the caller's balance back.

## The guest: mutation outside, record inside

Emitted stream at base `7b5f41255` (`gen-out/regionmap/stateless_guest.s`):

| what | where | side of the snapshot |
|---|---|---|
| caller live-balance debit ends (`.Lcd_deb_done_call_target`) | `.s:60481` | **before** |
| `jal ra, call_frame_descend` — **the snapshot** | `.s:61056` | — |
| `jal ra, record_message_value_transfer` | `.s:61068` | after |

**The descent *is* the snapshot**, verified rather than taken from a comment:
`call_frame_descend` writes `body_state_snapshot_by_depth` at `.s:46300` and `.s:46307`.

The debit itself is `ChildFrameHandlers.lean` (`git show 7b5f41255:…`): reverse the caller's
`env+32` into `cd_caller_newbal`, `u256_sub_be` the value, write it back — CALL (mode 0) only.

**No credit-back exists on the return path.** `cd_caller_newbal` is referenced nowhere in
`CallFrameReturn` or `NoopHalt`, and `frame_return` in the emitted stream contains no
`u256_add_be` and no write to the parent env's balance field. So nothing restores the debit when
the child reverts.

## Why the record being correctly placed makes this harder to see

[...]

PR #11002, opened 2026-07-31
codegen: share the record_message_value_transfer call setup between CALL and CREATE (GH #10938)

pirapira:
## Summary

GH #10938: share the `record_message_value_transfer` call setup between the **value-CALL descend
site** and the **CREATE-endowment site**. Extraction only — nothing moves relative to the child
snapshot.

**Byte-identity MEASURED on both artefacts, including the `.s` text.** Base `7b5f41255`:

| artifact | before | after |
|---|---|---|
| `stateless_guest.s` | `f136a84f842d39d73a0e0aa1…` | **identical** |
| `stateless_guest.elf` | `7df7e822d803caafb2533ea1…` | **identical** |

`.s` identity covers all **three** emissions at once (value-CALL, and the CREATE salt/no-salt
forms). No generated file changes, so no layout regen; `check-region-map.sh` passes unchanged.

## Why the spec wants these shared

`execution-specs` at pinned `e5a8caf1b` has **one** `move_ether`
(`vm/interpreter.py:385-390`) serving calls and creations alike, because
**`process_create_message` delegates to `process_message` at `:212`** rather than transferring
value itself. Creation does not have its own transfer — it uses the one the CALL path uses. Two
guest copies of that setup is duplication the spec forbids, not merely duplication we dislike.

## The extraction gate, run before committing to a shape

Both bodies pulled from `gen-out/regionmap/stateless_guest.s`, labels normalised, the five operand
globals renamed:

```
 CALL (.s:61068)            CREATE (.s:57132)
 addi sp, sp, -32           addi sp, sp, -32
 sd x10/x12/x13             sd x10/x12/x13
 la a0/a1/a2  FROM/TO/VAL   la a0/a1/a2  FROM/TO/VAL
 li a3, 1                   ld a3, 584(x20)      <-- differs
 la a4, SPRE                la a4, SPRE
 la a5, RPRE                la a5, RPRE
 addi a5, a5, 8             —                    <-- differs
 jal record_message_value_transfer   (same)
```

**Twelve lines against eleven, differing in exactly two places, and both are operand
locations** — which parameterise. **The control shape is identical**: same save set, same six
operands, same `jal`, same restore. No opposite branch senses, no label-count mismatch.

That is the check **cut 1 failed** — there the two candidate bodies had different control flow
(opposite branch senses, two labels versus three), so no shared emitter could reproduce both
without changing one site's instructions. Running the gate first is why this one is byte-identical
rather than withdrawn.

## Scope: what this deliberately does not do

* **Nothing crosses the child snapshot.** Both call sites keep their current position, which for
  both is already **after** the descent — the placement each site's comment claims and which
  matches `process_message`'s snapshot→`move_ether` order.
* **The value-bearing precompile tail is excluded** (`Programs.ChildFrameHandlerTailHelpers`). It
  emits the identical instructions as a **single `;`-separated line**, so routing it through this
  emitter would change `.s` line structure. That is not cosmetic — line structure has already been
  shown to change instruction **order** when the original line begins with an instruction the
  extracted block must follow (the GH #10619 prerequisite). It stays out until it can move without
  disturbing text.
* **The caller-side debit and the callee pre-balance staging are NOT touched.** Those are the two
  halves of `move_ether` and they sit on the wrong side of the snapshot; that is filed separately
  as GH #11001, and the repair is blocked on parent-env addressing because `call_frame_descend`
  repoints `x20` (`.s:46172` saves the parent env, `.s:46343` installs the child).

## Verification

* Byte-identity measured on `.s` **and** `.elf` (table above).
* Emitted call sites of the producer unchanged: **5** — `.s:39201`, `57132`, `58006`, `61068`,
  `67505`.
* `lake build` — clean.
* `scripts/check-region-map.sh` — OK, no generated file changed.
* **Fixture legs deliberately not claimed as evidence.** The ELF hash is *identical*, so a fixture
  run would execute the same binary and could not distinguish anything — it would be a tautology,
  not a check. (Pre-edit controls were measured green regardless: `bal_account_access_target` 8/8,
  `call_to_self_no_log` 2/2, guest `7df7e822…`.)

### DIFF
--- a/EvmAsm/Codegen/Programs/ChildFrameCreateTail.lean
+++ b/EvmAsm/Codegen/Programs/ChildFrameCreateTail.lean
@@ -463,10 +463,10 @@ def createUnsupportedTail (netPopBytes : Nat) (hasSalt : Bool) : String :=
     -- the child target's staged pre-state balance.  Record the shared
     -- move_ether pair after the child snapshot, so frame_return discards it
     -- with a reverting initcode.  Log scheduling remains below at this caller.
-    "  addi sp, sp, -32\n  sd x10, 0(sp)\n  sd x12, 8(sp)\n  sd x13, 16(sp)\n" ++
-    "  la a0, create_sender_be\n  la a1, create_address_be\n  la a2, create_value_be\n  ld a3, 584(x20)\n  la a4, message_value_transfer_sender_pre\n  la a5, nse_create_pre_bal\n" ++
-    "  jal ra, record_message_value_transfer\n" ++
-    "  ld x10, 0(sp)\n  ld x12, 8(sp)\n  ld x13, 16(sp)\n  addi sp, sp, 32\n" ++
+    -- GH #10938: shared setup with the value-CALL descend site — one `move_ether` in the
+    -- spec, because `process_create_message` delegates to `process_message` (`:212`).
+    recordMessageValueTransferAsm "create_sender_be" "create_address_be" "create_value_be"
+      "ld a3, 584(x20)" "message_value_transfer_sender_pre" "nse_create_pre_bal" ++
     -- coc3g.6 CAUSE 3: EIP-7708 transfer log for the CREATE endowment value move. Spec
     -- interpreter.py:307-316 emits Transfer(caller, current_target, value) at EVERY message-call
     -- frame entry with should_transfer_value and value!=0 and caller!=current_target, AFTER the
--- a/EvmAsm/Codegen/Programs/ChildFrameHandlers.lean
+++ b/EvmAsm/Codegen/Programs/ChildFrameHandlers.lean
@@ -987,10 +987,11 @@ def callDescendFallThrough
   -- reused by nested calls, so they are valid only in this post-descend,
   -- pre-dispatch window.
   (if mode != 0 then "" else
-    "  addi sp, sp, -32\n  sd x10, 0(sp)\n  sd x12, 8(sp)\n  sd x13, 16(sp)\n" ++
-    "  la a0, cd_caller_be\n  la a1, nse_callee_be\n  la a2, cd_value_be\n  li a3, 1\n  la a4, cd_balance_be\n  la a5, nse_acct\n  addi a5, a5, 8\n" ++
-    "  jal ra, record_message_value_transfer\n" ++
-    "  ld x10, 0(sp)\n  ld x12, 8(sp)\n  ld x13, 16(sp)\n  addi sp, sp, 32\n") ++
+    -- GH #10938: the setup is the shared `recordMessageValueTransferAsm`, which the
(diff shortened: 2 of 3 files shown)


pirapira:
## Note: the exclusion in this PR was right, but for a weaker reason than the real one

This PR excludes the value-bearing precompile tail on the grounds that it emits the identical
instructions as a **single `;`-separated line**, so routing it through the shared emitter would
change `.s` line structure. That reason is true and it is not the important one.

**The stronger reason, from k3's independent read: the precompile tail is a third copy of the value
payload whose emission condition is expressed a different way.** It gates its log with an **inline
self-transfer comparison** rather than the `cd_xfer_log_pending` flag that the descend and
empty-target paths use. So the guard exists in **two forms across three CALL paths**:

| path | transfer at | log gated by |
|---|---|---|
| descend | `.s:61068` | `cd_xfer_log_pending` flag |
| empty target | inside `.s:60957-61018` | `cd_xfer_log_pending` flag |
| precompile tail | `.s:58006` | **inline self-transfer compare** |

**Consequence for any further extraction:** lifting only the descend block leaves two copies
behind, and one of them does not share the guard mechanism — so a later "fold the rest in" step
cannot be a mechanical repeat of this one. That is GH #10938 cut 1's shape: the bodies look alike
until you compare what selects them.

## Copy census, measured at `7b5f41255`

Confirming the population this PR sits inside — `gen-out/regionmap/stateless_guest.s`:

* **Value debit — 4 copies across callers**, not four across CALL modes: CALL descend, precompile
  tail, and the **two** CREATE forms (`.Lcr_*` `f0` at `.s:56923`, `f5` at `.s:67296`).
* Within the CALL family it is **one** copy — `if mode != 0 then "" else …`, so mode 0 only. Only
  CALL transfers value: CALLCODE keeps it in the caller's own context, DELEGATECALL and STATICCALL
  carry none.
* **Naming-independent cross-check**, which is what makes the count trustworthy rather than an
  artifact of a label-stem list: every copy must contain a `u256_sub_be`. Of 14 sites, the
  live-debit ones are `.s:60464`, `57987`, `56936`, `67309` — exactly where the label census puts
  them — plus `.s:39217` inside `record_message_value_transfer`, which is the **row** arithmetic
  and not a live mutation.

⚠️ The label census initially reported one of these as two copies: the stem `cd_nse_cpaddr` also
matches `.Lcd_nse_cpaddr_d_call_target`, whose trailing `d` is a *done* suffix rather than a site
tag. Corrected above. Recorded because a census that silently over-counts is worse than a label
count that is honest about being one.

## Unaffected by all of the above

This PR remains **byte-identical on `.s` and `.elf`** and moves nothing relative to the child
snapshot. The note changes what a *future* step must account for, not what this one did.

Issue #10909, opened 2026-07-29
BAL: block gas over-claim (fail_code 62) on 06000, unmasked by removing tx.value from the upfront debit


pirapira:
## Block gas over-claim (`fail_code` 62) unmasked by removing `tx.value` from the upfront debit

Fixture: `06000_test_blob_gas_subtraction_tx_fork_Amsterdam-single_blob-blockchain_test_from_state_test--100_wei`

### What happens

On `main` this fixture **passes** (333/333). With `tx.value` removed from the staged upfront debit —
which is the spec-correct change, `process_transaction` charges `effective_gas_fee + blob_gas_fee` and
nothing else (`fork.py:1105-1108`, pinned `e5a8caf1b`) — it fails:

```
06000  LENGTH  fc=62  rebuilt=338  supplied=333  delta=+5
```

`fail_code` **62** is `.Lbv_block_gas_used_over_fail` (`BlockVerdictReceiptsTail:218`):
`header.gas_used > max(block_regular, block_state)`. **This is a gas over-claim arm, not a BAL
comparator arm** — the `LENGTH` class and the `+5` bytes are downstream of a run that already failed a
gas check, so the byte delta is a symptom, not the defect.

### Attribution — controlled

Three builds, one fixture, one instrument:

| build | result |
|---|---|
| clean `main` | **PASS** 333/333 |
| **deletion only**, no new record | **LENGTH `fc=62`** 338/333, `+5` |
| deletion + the #10892 sender record | LENGTH `fc=62` 338/333, `+5` — byte-identical |

So this is caused **solely by removing `tx.value` from the upfront debit**. The #10892 record is not
implicated (it reproduces nothing beyond what the deletion alone produces).

### Why this is a real defect and not a reason to revert

The guest's gas accounting on this fixture **depended on `tx.value` being inside the upfront debit** —
a quantity the spec does not debit there. Two errors were cancelling, and correcting the spec-wrong one
exposes the other. Also visible: account `3c46d5c0…`'s balance goes from declared 2,396,676 to 1,641,136,
i.e. **755,540 low** — a gas-shaped quantity, not the 100 wei transaction value, consistent with the
gas term rather than the value term being at issue.

This is a **false reject** (the guest rejects a block the spec accepts). `FA` is unaffected.

Unmasked by #10908.

### Not yet established

The specific gas term that changes has **not** been localised — I have the fail-code arm and the
[...]

PR #11007, opened 2026-07-31
fix(bal): publish top-level CALL value transfer

### LINKED ISSUE #11003: BAL: top-level CALL value publisher passes LE CALLVALUE to BE arithmetic and skips its AccountState hit
## Corrected mechanism

The first five framings of this issue were wrong and are retired: a miss-path
bail, an inverted lookup guard considered alone, an unpopulated sender-pre
buffer, a short AccountState copy, and a dishonest AccountState balance flag.
The combined defect is an **endianness mismatch plus an inverted guard** in the
top-level CALL value-transfer publisher.

`dispatcher_seed_pending_value_transfer` calls
`account_state_latest_balance(sender, bv_pending_value_sender_pre)`. A hit
returns `a0 = 1` and copies all 32 bytes of the sender's balance from the
AccountState entry; the old `bnez a0, .Ldpub_done` therefore skipped the
publisher precisely when its authenticated sender balance was available.

Flipping that branch alone is intentionally insufficient. The producer,
`record_message_value_transfer`, documents and enforces a **big-endian**
32-byte value input through `u256_sub_be`. Its caller passed
`bv_runtime_payload + srpc_env_base + 96` directly. That slot is the runtime
CALLVALUE word in **little-endian limb order**: `BlockVerdictDispatchTx:1017`
names it `CALLVALUE (LE)` and adjacent code performs LE limb arithmetic over
it. Thus a value of one is passed as `01 00 ... 00`, which BE subtraction
reads as 2^248 and necessarily underflows against an ordinary sender balance.

The repair must be measured as one change:

1. take the `account_state_latest_balance` hit path and skip a genuine miss;
2. reverse the LE CALLVALUE word into a dedicated BE scratch before calling
   `record_message_value_transfer`.

## Direct evidence

On `00183`, the candidate polarity-only experiment reached the producer but
could not publish either effect:

- `account_state_find` returned entry `0xbe26f740`; its flags were
  `0x63`, so balance-present bit 32 was set;
- `.Laslb_hit` copied its complete nonzero balance into
  `bv_pending_value_sender_pre` as four dwords:
  `0`, `0`, `0x3c2e3b0300000000`, `0x007479a03c80d09f`;
- the producer received the CALLVALUE operand with first dword `1` and the
  remaining three zero, i.e. LE value one;
- `u256_sub_be` returned borrow and branched directly to `.Lrmvt_done`;
  neither `record_nonstorage_effect` call was reached.

Neither the short-copy nor the dishonest-flag hypothesis survives the trace.
The AccountState mask is not the cause: the only producer that introduces bit
32, `account_state_record_nonstorage`, copies its post-balance into entry
offsets 32..63 in the same path. Code-only and auth writers do not introduce
that bit.

## Scope and prediction

The original random-200 investigation found 26 recipient-side BAL shortfalls
with this top-level value-transfer signature. Prediction registered before a
combined repair: those 26 rows move once both the hit guard and BE conversion
are present. The combined repair measured exactly those 26 `60 -> 0`
transitions.

A further deterministic transition,
`15764_test_return_non_const_fork_Amsterdam-blockchain_test_from_state_test--v1__b0`,
is `62 -> 0`; it is reported separately and not attributed to this
mechanism. Three independent parent runs were byte-identical and retained code
62.

The byte-order boundary is shared, so this issue also requires an inventory of
every consumer of runtime `env+96`: classify each as LE-safe or as passing the
word to a BE helper. Do not generalise the repair from this one caller without
that audit.

## Spec

`move_ether` reduces the sender then increases the recipient in
`state_tracker.py:646`; `process_message` calls it at
`interpreter.py:385` when value transfer is enabled and nonzero. The BAL must
represent both effects, regardless of the guest's internal runtime-word
encoding.

## Related

#10698 (sender-side reconstruction), #10996 (recipient-side signature), and
#10944 (top-level CREATE publisher).

pirapira:
Fixes #11003.

`dispatcher_seed_pending_value_transfer` publishes the top-level CALL value
transfer.  It was wrong in two mutually masking ways:

1. `account_state_latest_balance` returns nonzero on a hit, but the caller
   branched away on that hit instead of on a miss.
2. `bv_runtime_payload + env + 96` is an EVM stack word in little-endian
   limbs, while the shared producer calls `u256_sub_be`.

The fix falls through on an authenticated AccountState hit and reverses that
32-byte word into a dedicated BE scratch cell at the ABI boundary before the
producer subtracts it.  The nearby comment records this contract.  The
conversion is deliberately local: the late common publisher does not have one
uniform transaction-context pointer, whereas its runtime value source is
uniform.

The five rejected framings during diagnosis were: a miss-path bail, the branch
inversion alone, an unpopulated sender-pre buffer, a short AccountState copy,
and a dishonest AccountState balance flag.  The last two were disproved by the
hit trace; the first two each leave the other bug masking the output.

## Layout

- `.text`: `0x066be8 -> 0x066c14` (`+44` bytes / 11 instructions): two scratch
  `la`s, one counter load, six loop instructions, and two final-address
  instructions.  The final `la a2` is additional, not a replacement for the
  old source-address formation.
- `.data`: unchanged at `0x5370`; `bnf_le_a` remains `0xa3000ee0`.
- `.bss`: `+32`, exactly the new cell `[0xbd39bca0, 0xbd39bcc0)`, immediately
  after the retained sender-post cell `[0xbd39bc80, 0xbd39bca0)` with no
  padding.  RegionMap and GuestImage were regenerated from the linked ELF;
  final `.bss` is `0x1c0fd000`.

## Evidence

Pinned artifacts:

- parent ELF: `7df7e822d803caafb2533ea1013a91e137dc888fcc097ff8363c08007c54b423`
- candidate ELF: `115438b5fad68a78b9b360ea2c137cb50c83ff5589023647394f9ce27d2bb0dd`
- runner: `c6ac09fa72c2103b6d3e390b779b022ef7f220e5a39bb5f667114a89d78e5868`
- canonical paired selection: `7489a4f42d02bc48b92a16ed3f6c2447541374e26ccb0f7c11f75baadaab7f5c`
  (it contains no seed metadata).

Targeted controls: `00183`, `00184`, and `00185` now pass byte-exactly;
`17437`, `17576`, `17625`, and `17502` remain byte-exact passes with the
publisher live; expected-invalid `04778` still rejects byte-exactly.

On the paired random-200 selection, parent FR is 105 and candidate FR is 78;
there are 26 confirmed code-60 forward transitions, one additional
unattributed forward transition, zero reverse transitions, and FA is zero on
both legs.  Prefix-exact rows rise from 95 to 122.  The registered prediction
holds exactly for the 26 `60 -> 0` rows.  The remaining deterministic forward transition is
`15764_test_return_non_const_fork_Amsterdam-blockchain_test_from_state_test--v1__b0`,
which is `62 -> 0`; it is reported separately and not attributed to the
code-60 mechanism.  Three independent parent runs returned byte-identical
output and fail code 62 for that row.

Structural checks passed: `check-asm-to-program.sh`, `check-region-map.sh`,
and `check-build-parallel.sh`.


pirapira:
Labelled. Determinism is recorded (three parent runs, byte-identical, `fail_code=62`), the branch name no longer records the superseded fallback diagnosis, and the claim is stated as **26 confirmed plus 1 unattributed** rather than 27.

**One follow-up that does not block this PR**, since it concerns only the row this PR explicitly does *not* attribute: the `62 -> 0` fixture is named `…--v1__b0` here and was reported as `…--g0__b0` over the fleet channel. Those are different variants. Fixture variant resolution is a known hazard in this repo and the identity ladder holds that **only the input SHA survives** — relpaths and basenames collide. So please add a comment resolving which variant it was, **with its input SHA**, so whoever investigates that `62` later re-runs the row that was actually measured. Cross-referencing it on #10909 is worth it too: it tells whoever works the `62` class that their population contains at least one row whose `62` had a non-gas root.

Issue #11010, opened 2026-07-31
transaction validity: reject all malformed typed-fee decode statuses


pirapira:
## Divergence

The guest rejects `tx_effective_gas_pricing` decode statuses 2 and 3, but status 1 falls through and allows multi-transaction execution to continue. Status 1 is produced by fee-field decoding failures that are not valid execution-specs transaction inputs.

## Spec

At execution-specs pin `e5a8caf1b`, `fork.py:913` decodes every transaction unconditionally while applying a block; `transactions.py:554` defines `decode_transaction`; and the stateless analogue does the same at `stateless_host.py:109`. An undecodable transaction cannot proceed through the spec.

## Guest

`BlockVerdictMtxRuntime.lean` computes effective gas pricing. `BlockVerdictMtxTail.lean` branches statuses 2 and 3 to the fee-invalid rejection but does not reject status 1. `multi_tx_nth_context` does not independently validate the fee fields.

## Required alignment

Route status 1 to the existing fee-invalid rejection alongside statuses 2 and 3. This is transaction-validity alignment, not a claim of a demonstrated false accept; no KAT is required for the issue.

## Dependency

Do not begin the patch until #11007 has landed. Keep this distinct from #10698.

pirapira:
Retired as stale on current origin/main 724c819c3: BlockVerdictMtxRuntime.lean already routes tx_effective_gas_pricing status 1, 2, and 3 to .Lbv_fee_invalid_fail, with the execution-specs decode rationale in the adjacent comment. No duplicate patch is needed.

PR #11011, opened 2026-07-31
Reject status-1 MTx fee decoding failures

pirapira:
## Summary

Reject `tx_effective_gas_pricing` status 1 through the existing invalid-fee path. Status 1 means transaction-type or canonical fee extraction failed; allowing it to continue differs from execution-specs transaction validation.

## Specification basis

Pinned execution-specs `e5a8caf1b` rejects malformed transactions before applying them:

- `src/ethereum/forks/osaka/fork.py:913`
- `src/ethereum/forks/osaka/transactions.py:554`
- `src/ethereum/forks/osaka/stateless_host.py:109`

Statuses 2 and 3 already rejected at this guest site. This change gives status 1 the same transaction-validity treatment. Status 4 remains separately proven unreachable.

## Corpus coverage

No valid EEST fixture exercising status 1 was identified while #11010 was scoped. This is not yet a corpus-wide coverage census; the limitation and upstream fixture follow-up are tracked in #11012. The absence of a discovered witness explains why this transaction-validity divergence could survive unnoticed, but does not weaken the specification case.

No KAT was constructed: maintainer direction is to land this spec-alignment correction without manufacturing a witness.

A paired, seeded 200-row Spike sample checks the one possible harm from making the guest stricter: a valid block newly rejected by a mishandled status-1 extraction. Parent and candidate outputs were byte-identical for every selected fixture: zero reverse transitions. This null result is consistent with the sample not reaching the divergence; it does not establish corpus-wide absence of status-1 coverage.

- seed: `20260731`; 200 uniformly sampled of 26,104; 8 jobs
- parent ELF: `115438b5fad68a78b9b360ea2c137cb50c83ff5589023647394f9ce27d2bb0dd`
- candidate ELF: `61d47dd33100982ded872028a098dd9dc78b5755fe0b0310a1416c5151fcab9b`
- frozen runner: `c6ac09fa72c2103b6d3e390b779b022ef7f220e5a39bb5f667114a89d78e5868`
- both legs: 132 full matches, 68 existing false rejects, 0 execution errors

## Regeneration and checks

Fresh clean-worktree regeneration:

- `.text`: `0x066c20`
- `.data`: `0x005370` (unchanged; `bnf_le_a = 0xa3000ee0`)
- `.bss`: `0x1c0fd000` (unchanged)

Passed:

- `scripts/check-asm-to-program.sh` — 390 converted definitions
- `LAKE_ARTIFACT_CACHE=false scripts/check-build-parallel.sh`


pirapira:
Labelled. The evidence is complete for an alignment change: three spec anchors at the pinned `e5a8caf1b`, a clean regen with structural checks passing, differing parent/candidate ELF hashes (`115438b5` / `61d47dd3`) proving two builds, the frozen runner `c6ac09fa` on both legs, and **zero reverse transitions with all 200 outputs byte-identical** — which is the one thing this change could have got wrong, since making the guest stricter is the direction that creates false rejects.

The sentence I most wanted is in the body and phrased the right way round: *"This null result is expected from the corpus's lack of a valid status-1 case; it does not mean the divergence is inert."* A reviewer reading a null the other way would conclude the PR is pointless. #11012 now tracks that coverage gap as an upstream EEST task.

**One honesty item, not a gate:** `seed: 20260731` is date-shaped, so it was presumably **specified rather than captured**. That does not weaken this result — the A/B's validity rests on the differing ELF hashes, and a null across 200 byte-identical outputs does not depend on how the sample was drawn. But the standing convention is captured-or-declared, so please say "specified" beside it. Worth being precise about given the same thing came up on this change's predecessor and was disclosed there.

PR #11016, opened 2026-07-31
test(correspondence): add BAL canonical ordering as the third family

dhsorens:
**Stacked on #10949** (base: `docs/rlp-ssz-spec-correspondence`). Review that one first; this PR's diff is only the BAL family.

## Why BAL

#10817 (proof package for `bal_canonical_sort`) is blocked three ways:

1. `EvmAsm/Codegen/Programs/BalCanonicalSort.lean` defines **only `String`s** — zero `: Program`, so `cpsTripleWithin` cannot state sortedness at all.
2. The documented conversion tool refuses all four asm functions in the module (leading `.globl`; a spliced `String` ident).
3. The module docstring argues sortedness + permutation is *insufficient* anyway — a sort on the wrong key is still sorted and still a permutation.

The correspondence framework sidesteps all three, because it audits **model vs reference**, not machine vs model. Canonical ordering is pure data-in/data-out — the "stateful family, audited at a functional boundary" case — so a differential needs **no `Program`, no triple, no conversion**.

It also answers (3) directly. That objection asks for a key defined independently of the implementation's own byte-picking. A differential against the reference's *declared* ordering **is** that key, made executable rather than argued.

## Result

`SpecRef._build_from_builder` agrees with the vendored reference on **1149/1149** records — account order (byte-lexicographic), slot order, per-index orders, and the read/write exclusion.

> **The ordering is right. Whether the assembly implements it is open.**

That split is the deliverable: an open question ("is our canonical ordering even the right ordering?") becomes closed, and #10817 is left strictly smaller — its sortedness predicate can now be written against a validated model instead of a key inferred from the assembly.

Corpus pins the cases the verdict turns on: the **endian trap** from EEST `test_bal_lexicographic_address_ordering` (`0x01…02` vs `0x02…01`), and **byte-length boundaries** (`255` vs `256`) separating numeric from encoded-byte order. That last one matters because the reference's *own docstring* (`:526`–`534`) says slots sort "lexicographically" while the code sorts `U256` **numerically** — a model written from the prose would diverge.

## Schema change: `Layer` (guest | model)

BAL forced this. Its model is `diff`-backed while every guest routine is `unproven` — a shape the registry could not express when every row was assumed to be a guest routine with a theorem. Recording only guest rows would hide the result; recording the model row as a routine would misreport what is proven. `verdict_requires_spec` / `basis_diff_requires_spec` are now scoped to guest rows, and a new `specless_diff_is_model_only` stops a spec-less `diff` **guest** row from landing.

## Gate fix

`check-correspondence-deps.sh` blanket-banned `EvmAsm.Rv64` — and would have **rejected this legitimate subject**, because the BAL model reaches the three *cheap* leaves `Rv64.{Word,Basic,ZiskAccel}` via `SpecRef/Crypto` (keccak). A prefix list cannot tell those from the proof tower; counting modules can. Replaced with a closure **size budget**, which is the property actually being protected. Both limbs are self-tested — the budget limb by re-scanning the real harness with `MAX_CLOSURE=1` and requiring failure. Closure is **20 modules** against a budget of 80.

## One axis produced a false result first — worth flagging

The `already-canonical` auxiliary axis initially included storage reads and reported **32 mismatches**. All artifacts: Python stores `storage_reads` in a `Set`, so input order is destroyed before `_build_from_builder` ever runs — CPython iterates `{223, 75}` as `[75, 223]` — and the axis reported "already canonical" for *every* input while measuring nothing. Narrowed to the accounts list, which Python holds in an insertion-ordered `Dict`, so the question is well posed on both sides. Manufacturing a comparison the reference cannot express is exactly what the method warns against.

## Findings recorded, not fixed

On `docs/bal-spec-correspondence.md`:

1. **`BalCanonicalSort.lean:218` uses `blockAccountWritesCapacity` for *every* container**, regardless of which array is sorted — the live path's arrays have different capacities. Possibly a real bug; not exercised by a model-layer audit.
2. `bal_sort_storage_writes` / `bal_sort_account_writes` are **dead code** (`:456` says so), and their `#guard`s pin a path nothing runs.
3. Status 4 is documented at `:199` and never emitted.
4. Comment/code drift at `:498` (`0x1400` vs `0x9400`).
5. The BAL sort symbols are **absent from `GuestAddrs.lean`**.
6. Representation gap: the Lean model uses `List` where Python uses `Dict`/`Set`, so it can represent duplicate slots/reads the reference cannot. The corpus stays inside the representable domain; worth knowing before proving uniqueness against the Lean type.

## Verification

- `lake exe correspondence-check bal` → **1149/1149**, all four counters zero
- `lake exe correspondence-check rlp` → **3757/0 unchanged** (the second family did not perturb the shared harness)
- `--self-test` green for **both** families; `check-correspondence-deps.sh --self-test` green on both limbs
- corpus regeneration byte-identical for both families
- full `lake build` (4799 jobs), `check-progress`, `check-drift`, `check-unimported` (0 orphans), `check-spec-refs`, `check-file-size`, no warnings


pirapira:
Reviewed. **The findings-recorded-not-fixed section is now filed, so this PR does not have to carry them:**

- **#11017** — the `BalCanonicalSort` hygiene cluster: the two dead sort routines *and* the `#guard`s that pin their unreachable path, `bal_sort` symbols absent from `GuestAddrs.lean`, status 4 documented at `:199` and never emitted, and the comment drift at `:498`.
- **#11018** — the per-container bound question, plus the `List`-vs-`Dict`/`Set` representation gap.

Two notes on how I framed them, since neither matches this PR's wording exactly:

**`:218` is not filed as a defect.** Per the maintainer, using `blockAccountWritesCapacity` for every container is a deliberate static-allocation choice under tight memory, not a bug. What #11018 records instead is the behavioural question it leaves open: static sizing governs what the guest *can* hold, while the spec governs what a block may *validly* contain, and one shared capacity only settles the first. If a container's logical limit is tighter than the shared allocation, a runtime bound may be needed for equivalence.

**The representation gap is filed as a proof-side caveat rather than a model defect.** Your corpus stays inside the representable domain, so 1149/1149 is sound — it simply cannot speak to the surplus domain, which is exactly what matters for #10817 if a uniqueness obligation gets stated over `List`.

## What I verified rather than took on trust

- **The docstring/code divergence is real.** The reference lists "Account addresses (lexicographically)", "Storage slots (lexicographically)", "Transaction indices (numerically)" while the code sorts slots as `U256` numerically — so a model written from the prose *would* diverge, and the `255`/`256` boundary in the corpus is the case that separates them. This is the strongest single item in the PR.
- **`bal_sort` symbols: zero matches in `GuestAddrs.lean`** on `7c4918cb3`.
- `:456` does say nothing calls those routines; `:199` does document status 4; `:498` does carry the `0x1400` value.

## Two things worth saying about the method

**The headline claim is correctly bounded** — *"the ordering is right; whether the assembly implements it is open"* — and that bound is the deliverable, not a hedge. It converts an open question into a closed one and leaves #10817 strictly smaller.

**And the disclosed false result is the part I'd most want kept.** The `already-canonical` axis reporting "already canonical" for every input while measuring nothing — because Python's `Set` destroyed input order before the comparison ran — is a gate that could not fail, self-caught and narrowed. Today produced several of those; most were caught by a reviewer rather than the author.

Same for the deps-gate change: replacing a prefix ban that would have rejected this legitimate subject with a closure **size budget**, self-tested by re-scanning at `MAX_CLOSURE=1` and *requiring* failure. Proving a guard can fire before trusting it green is the discipline this repo has been short of.

## Sequencing

Stacked on #10949, so that merges first; both are `BEHIND` after today's merges. One caution earned today: #10949 is XL and a **clean** auto-merge in a file two PRs both touched is how one side gets silently dropped — worth verifying both edits survive **by name** after the rebase rather than trusting the absence of a conflict.

dhsorens:
Thanks — corrections applied in `f70cdb60f`, plus the title fix the bot flagged (`test(correspondence): …`).

## Your two reframings, taken

**`:218` is not a defect — I was wrong to call it one.** The page now says so explicitly rather than quietly dropping it, since "possibly a real bug" is exactly the kind of claim that propagates. Recorded instead is the distinction you drew: static sizing governs what the guest **can hold**, the spec governs what a block may **validly contain**, and one shared capacity settles only the first — so a runtime bound may be needed for equivalence where a container's logical limit is tighter than the shared allocation. That is a sharper statement than what I had, and it is the one that actually bears on #10817.

**The representation gap as a proof-side caveat, not a model defect** — agreed, and your phrasing is the precise one: the corpus stays inside the representable domain, so 1149/1149 is sound; it simply **cannot speak to the surplus domain**. That is the sentence that matters if a uniqueness obligation gets stated over `List`, so I've used it directly.

Both now cite #11017 / #11018, so this PR no longer carries them.

## The merge caution — applied, and it was worth applying

You were right that a *clean* merge is the dangerous case. The merge of the updated parent produced **zero conflicts**, so I verified all three sides **by name** rather than by absence of conflict:

| Side | Checked | Result |
|---|---|---|
| BAL (this PR) | `Layer`, `bal_rows`, `Bal.subject`, `MAX_CLOSURE`, `kindLabel`, `correspondence-check bal`, `bal_family`, + all 4 new files | all present |
| Parent (#10949) | `structure Subject`, `runDecode`, method page, `spec_oracle/pins.py`, `rlp.tsv` | all present |
| `main` | regenerated `symbol-addresses.tsv` (1784 lines changed) | empty diff vs `origin/main` |

Post-merge: bal **1149/1149**, rlp **3757/0**, both self-tests, `check-progress`, `check-drift`, `check-unimported` (0 orphans), import budget 20/80.

## One branch-topology note

While syncing I found the base had drifted in a way worth flagging: when you clicked "update branch" on #10949, that landed on its **fork** head (`51d8beb2a`), while this PR's base — the `Verified-zkEVM` copy of the same branch — was still at `da74c5fc2`. Two refs with the same name, different tips.

I fast-forwarded the origin copy to `51d8beb2a` (ancestor-checked first; no rewrite, no force), so #11016's base now reflects what #10949 actually points at. Once #10949 merges I'll retarget this to `main` and merge down — flagging it so nobody reads the base as "already merged" in the meantime.

## On the false result

Noted, and I'd rather it stayed visible too. Worth adding one detail for whoever hits this next: the axis wasn't *wrong*, it was **ill-posed** — Python stores `storage_reads` in a `Set`, so the input ordering was destroyed before `_build_from_builder` ever ran, and no implementation of that axis could have distinguished anything. The general rule I took from it, now in the method page's spirit if not yet its letter: **before believing a differential result, check the reference can even represent the question you are asking.** A comparison the reference cannot express will return a confident constant.

PR #11022, opened 2026-07-31
codegen: record the top-level CREATE endowment through the shared recorder (GH #10944)

pirapira:
## Summary

GH #10944: the top-level CREATE endowment is never recorded. This makes the creation route
**reach the existing shared recorder** rather than acquiring a second call site, and supplies the
**authenticated pre-balance** the route has been missing.

Base `7c4918cb3`. Behavioural — justified by the spec divergence, cited below.

## The spec: one `move_ether`, shared by calls and creations

`execution-specs` at pinned `e5a8caf1b`:

* **`process_create_message` DELEGATES to `process_message`** — `vm/interpreter.py:212`.
* The transfer lives inside that shared body at **`vm/interpreter.py:384-390`**, gated by
  `should_transfer_value and value != 0`, with `emit_transfer_log` nested in the *same*
  conditional at `:391-397`.

So the spec has **exactly one** transfer site serving both. ⚠️ **Adding a second publisher call
site to the creation route would therefore create a copy of something the spec has once** — the
duplication this convergence phase exists to remove. The guest source already says the omission is
deliberate-pending rather than accidental: *"Top-level CREATE deliberately remains outside this
helper pending #10944's authenticated created-address pre-balance lookup"*
(`git show 7c4918cb3:EvmAsm/Codegen/Programs/EIP7708Logs.lean`).

## The prerequisite already existed and was simply not wired here

`create_prebalance_acct` (128 B) and `create_prebalance_lookup_status` are declared in
`Programs/CreateFrameDescend.lean`, and the **nested** routes populate them via
`account_at_header_state_root_tracked`. The top-level creation route only **zeroed** the status
(`BVCS`) and **consumed** it at `.Lbvcr_payload_unsupported` — nothing on the route ever set it.

**So this is a missing call site for an existing authenticated mechanism, not a new component.**

### The pre-balance is authenticated, not assumed

`nse_zero_bal` is **wrong as an unconditional default and right only as a lookup result** — a
deployable pre-existing account may hold ether. All three outcomes are honoured exactly as the
nested sites honour them:

| result | meaning | action |
|---|---|---|
| `0` | FOUND | use the looked-up balance, 32 B BE at `create_prebalance_acct+8` |
| `1` | ABSENT | not in header state, so the pre-balance **is** zero — an *established* zero |
| `≥2` | MALFORMED | set `create_prebalance_lookup_status`; the existing consumer bails |

⭐ ABSENT and FOUND-with-zero produce the same row, and that is sound **because the absence was
authenticated**. Same bytes as the old `nse_zero_bal`, different epistemic status — and only one of
them is a substitute.

## Placement and guard

* Gated on the endowment being nonzero, **reusing the test the route already computes** for the
  EIP-7708 staging. The spec gates `move_ether` and `emit_transfer_log` in *one* conditional
  structure, so one guard is correct rather than two.
* Placed **after `dispatcher_capture_body_state`**, matching `process_message`'s
  snapshot-then-transfer order (`:380` then `:384`), so a failing body discards the record.

## ⚠️ A defect I introduced and found before claiming a result

My first version copied the lookup verbatim from `CreateFrameDescend`, which reads
`ld a0, 576(x20)` — correct **there**, because the nested route runs with `x20` = env base. **The
creation route never repoints `x20`**; every other env access in `BlockVerdictCreationStage`
materialises the base (`la t1, evm_env; …`). The sequence transferred and its register convention
did not.

The countable tell, worth re-running after any such copy: **grep the destination for every register
the copied sequence names** — `x20` appeared exactly once in that file, and that once was the
borrowed line. Fixed by materialising the base, and the convention is now **named explicitly in a
comment at the call site** so the next copy does not inherit the class one route further on.

⭐ And this is evidence about **shareability**, not just about one bug: because the copy needed its
registers rewritten, the nested and top-level pre-balance lookups are **not trivially shareable**. A
future unification hits exactly this — the same class that killed #10938's cut 1 and blocks the
item-9 relocation.

## Verification

* `lake build` — clean.
* `scripts/check-build-units-link.sh` — OK; every expected-linkable unit still produces an ELF and
  every allowlisted non-linkable unit is still genuinely non-linkable.
* `scripts/check-region-map.sh` — OK, regen at a **fixed point** (a second complete pass changed
  nothing); `GuestAddrs.bnf_le_a` still `0xa3000ee0`; `.data` unmoved.
* Pins **measured, not predicted**, with both ends named: base `7c4918cb3` `.text 0x066c50` /
  `.bss 0x1c0fd000`; this branch `.text 0x066d60` / `.bss 0x1c0fd020`. Deltas `+0x110` and `+0x20`
  — the latter is exactly the one 32-byte buffer, no alignment absorption.
* **Controls green on the artefact this PR pushes**, not an earlier one: `bal_account_access_target`
  8/8, `call_to_self_no_log` 2/2, `errored 0`, **FA=0**.
* Recorder call sites went **5 → 6**: CREATE now reaches the shared recorder.

## ⛔ Honest statement of what is NOT shown

**No fixture witnesses the fix.** I selected candidates by property rather than by name — scanning
fixture JSONs for a transaction with empty `to` and nonzero `value`, 131 such cases across 416
files — and ran two families paired with differing guest hashes. Both came back null **for a
structural reason: every case selected failed at `bv_fail=53` (receipts root), which is upstream of
the BAL comparison, so no BAL row change could have shown regardless.**

Two cases in the same file *do* reach the BAL digest (`bv_fail=60`), which yields a rule worth
recording: **a fixture file is not a unit of failure mode** — a `--filter`+`--limit` subset can be
uniformly non-witnessing by accident of which cases it drew, so witnesses should be selected by
failure code, not by filename.

The justification here is the **spec divergence itself**, per the standing rule. The witness remains
open work, and if the corpus turns out to contain none for this path that is worth knowing
independently of this PR.


pirapira:
landed via #11023

Issue #11028, opened 2026-07-31
Build: fresh worktrees can fail writing read-only cache-materialized ileans


pirapira:
## Symptom

A fresh worktree with artifact cache enabled can fail lake build with permission denied while Lean writes .ilean outputs. The affected files are cache-materialized hardlinks with mode 0444, so the diagnostic looks like a source/build error but is environmental.

This occurred twice on 2026-07-31: a #10863 pass-3 relink and the #10937 child-precompile extraction. In the latter, ChildFrameHandlerTails.ilean and unrelated proof/test ileans all failed to write.

## Working remedy

Run the affected build with LAKE_ARTIFACT_CACHE=false. Do not chmod anything in .lake/build: cached artifacts are hardlinks and chmod would mutate the shared cache inode. If a one-off local workaround is unavoidable, delete only the named regenerable output; never delete .lake wholesale.

The failure message should ideally point to cache mode and the safe bypass, since a normal permission-denied error misdirects the investigation.

pirapira:
## Third occurrence, and the first at whole-tree scale: 5,224 read-only artefacts

Hit this in `/home/yoichi-bkp/evm-asm2` on `2026-08-01` while regenerating layout for #10998.

**Symptom** — `scripts/check-region-map.sh` exits 1 with what looks like a Lean failure:

```
✖ [4296/4298] Building Main
info: stderr:
permission denied (error code: 13)
  file: /home/yoichi-bkp/evm-asm2/.lake/build/lib/lean/Main.ilean
error: Lean exited with code 1
```

⚠️ Nothing in that output says *permissions* until the `stderr:` line, and it arrived immediately
after a **successful** `lake build` — so it reads as layout drift from the change under test. It
took checking `ls -la` on the named file to see mode `444`.

### Scale, which is what differs from the earlier reports

Not one file. **5,224 read-only files under `.lake/build`**:

| extension | count |
|---|---|
| `.ilean` | 2571 |
| `.c` | 1273 |
| `.olean` | 1272 |
| `.export` | 103 |
| binaries (`bin/codegen`, `bin/progress-report`, `bin/specref-eest-check`, `bin/arith-diff-check`) | 4 |

**`.lake/build/bin/codegen` itself is read-only**, so any relink in the tree hits the same wall,
not just an incremental Lean build. The two earlier occurrences were single files; this is the
whole materialised cache, which is a different severity: a single file looks like a glitch, a
whole tree looks like a broken toolchain.

[...]

PR #11030, opened 2026-07-31
codegen: unify root and child precompile computation

pirapira:
Unifies the genuinely duplicated root and child precompile computation: canonical address classification, fixed/word/per-unit gas formulae, and the root validators that the child route already executes. Output copying, CALL success push, continuation, and EIP-150 allotment remain caller-owned.

Spec basis: Osaka interpreter dispatches one PRE_COMPILED_CONTRACTS function at every depth (execution-specs pinned e5a8caf1b, vm/interpreter.py:272-277). Individual precompiles charge their own gas and produce output (for example sha256.py:37-48); CALL-family result copying and success signaling happen in the caller (vm/instructions/system.py:312-349), while EIP-150 allotment is caller-side (vm/gas.py:314-354 and system.py:399-433).

Behavior changes selected by the spec: root word-cost addition overflow now follows the existing precompile-fail route instead of wrapping, and root ECADD/pairing/BLS validation now reuses the child validators. The random sample does not exercise these choices: byte-identical sample output is evidence of no sampled regression, not evidence that these changes are inert or independently proven.

Label-entry ABI note: .Lbv_tx_gas_precharge_pc0_prefix is also a direct MTx entry. MTx deliberately arrives with t0 holding its precompile-lane sentinel. An initial placement moved the label before root t0 setup, sending MTx into the KZG path; the paired control reproduced the fault on 00034. The final placement keeps root setup before the label, preserving the interior-entry register contract. Future label moves need a direct-jump census and the live-register contract for each entry.

Preparatory half: #11029.

Validation:
- fixed-point regeneration: text 0x066fa0; data 0x5370 and bnf_le_a 0xa3000ee0 unchanged; bss 0x1c0fd000 unchanged
- lake build; check-asm-to-program; check-region-map; cache-disabled check-build-parallel
- 00034 repaired: candidate rc=0 and byte-identical to parent
- paired random-200, base 13a8d9353, seed 2075889796, runner bae48269: parent 5c9c2831 and candidate 9cd7f837 each selected/ran 200 with 0 errors, 69 failures, 131 full/success matches, and 200 root/tail matches; all 200 outputs byte-identical


pirapira:
Two additions from review, posted as a comment rather than a body edit so the
`merge!` label stays in place and the bot keeps ownership of the head.

**1. The cache-disabled gate has a filed cause: #11028.**

The body notes that `check-build-parallel` was run cache-disabled. That is not an
ad-hoc deviation — it is the known limitation filed as **#11028**: cache-materialised
`.ilean` hardlinks land read-only, so a fresh worktree can fail writing them, and
default cache mode can leave `lake env` without materialised oleans. Working remedies
are removing the specific blocked regenerable `.ilean` files, or `LAKE_ARTIFACT_CACHE=false`.
`chmod` under `.lake/build` is not acceptable — 444 is a guard.

This shaped two separate runs on 2026-07-31, which is why it was worth filing.

**2. Scope correction, and the error was the reviewer's not the author's.**

An earlier review message of mine described this PR as converging `ecrecover`'s recovery
and `modexp`'s output computation onto the child validators. **It does not, and the PR
body is correct where my message was wrong.**

What this PR actually converges at root: **ECADD, BN254 and BLS pairing, and BLS add,
MSM and map.**

**Still omitted at root, unchanged here:** `ecrecover` recovery and `modexp` output
computation, both of which the child computes. These were identified during the audit
and are deliberately out of scope for this PR; they need to remain visible rather than
being read as closed by it, since `PRE_COMPILED_CONTRACTS` maps one address to one
function called at every depth.

pirapira:

Issue #11026, opened 2026-07-31
BAL: two declared balance_changes are omitted on 24639


pirapira:
## Witness\n\nFixture 24639 has fail code 44 and shadow status 1. Parent and #10635 candidate are identical at rebuilt 395 versus declared 429. After accounting for the separate 27-byte missing precompile access-only entry, the remaining 7 bytes are two omitted balance_changes fields:\n\n- c0f6dc9e5836f54caadbf59cc69346c508e1992b: rebuilt empty field c0; declared c3c20101 (4 bytes).\n- da3b3c05c2758fb8be75c74c5fc2d386c9a4b906: rebuilt empty field c0; declared c3c20180 (3 bytes).\n\nThe outer RLP header is the same length on both sides, so 27 + 4 + 3 accounts for all 34 bytes exactly. This is independent of #10635; nonce fields match in both streams.

pirapira:
Re-measured on current `origin/main` 724c819c with freshly built guest `b24eb035a58f7bd7ff7f2a1ff74875a7ecf952e9bba013cd45f618c738cfc03f` and the repository Spike runner at `SPIKE_OUTPUT_LEN=256`: fixture 24639 now returns `bv_fail_code=0`, `shadow_status=0`, and rebuilt/supplied BAL lengths `429/429`. This includes the two previously missing balance-change payloads. This issue is fixed on main; closing as completed.

PR #11038, opened 2026-07-31
BAL: record delegated precompile targets

pirapira:
## Spec alignment

`prepare_dispatch` adds a resolved delegated address to the accessed-address set unconditionally before `get_account` and `get_code` (`execution-specs` pinned e5a8caf1b, `interpreter.py:300`, then :304-307). A delegation that resolves to an active precompile follows the empty-code route in the guest, so it was omitting that required all-empty BAL entry.

This records `bsbd_deleg_target` only on the two active-precompile status-2 execution routes. The resolver `a3 = 2` liveness probe is deliberately unchanged.

## Evidence

- 24639: rebuilt/supplied moves from 395/429 to 422/429. The exact +27 is the previously decoded all-empty `0x0000000000000000000000000000000000000001` entry; the remaining 7-byte deficit is separately tracked by #11026 as two `balance_changes` fields.
- The discriminator is code presence, not delegation: normal delegated control 24779 remains 397/397; direct-precompile controls 20634 and 20835 remain 414/414 and 443/443.
- Expected-invalid control 04778 continues rejecting at fail code 5.
- Regenerated to a fixed point: `.text` 0x6723c; `.data` remains 0x5370 and `bnf_le_a` remains 0xa3000ee0; `.bss` remains 0x1c0fd020.

## Checks

- Cache-disabled full `lake build` is in progress; normal cached build was blocked by read-only artifact-cache hardlinks, so no build-tree permissions were changed.


pirapira:
Update: the cache-disabled full `lake build` completed successfully (4,799 jobs, exit 0). The cached invocation was not used because its artifact hardlinks were read-only; no build-tree permissions were changed.

pirapira:

PR #11039, opened 2026-07-31
BAL: record value transfer for empty CALL targets

pirapira:
## Spec alignment

`process_message` keeps `move_ether` and `emit_transfer_log` in one conditional structure (`execution-specs` pinned e5a8caf1b, `interpreter.py:384-390`). The guest emitted the deferred transfer log on a successful empty CALL return but published the paired balance effects only after `call_frame_descend`; an empty target has no child frame and therefore skipped the effects.

This publishes the existing site-resolved transfer descriptor at the generic mode-0 empty-success convergence point. It performs no new state lookup and excludes CALLCODE, whose recipient semantics differ.

## Arrival audit

Every route into `.Lcd_empty` bypasses the existing post-descend publisher: non-live CodeState, both delegation status-2 routes, absent header account, and verified `EMPTY_CODE_HASH`. Direct precompile tails do not enter this generic routine. Therefore the publisher executes once on each affected CALL and cannot duplicate an existing balance effect.

## Evidence

- #11026 witness 24639 moves from 395/429 to 402/429: the exact +7 is its two decoded balance changes, c0f6 0→1 (+4) and da3 1→0 (+3).
- #11038 separately restores its missing 27-byte delegated-precompile access entry. Together they predict 24639 at 429/429 and passing; #11038 is merge!-labelled and is deliberately not modified here.
- Controls remain byte-perfect: delegated-code 24779 397/397; direct precompile 20634 414/414 and 20835 443/443. Expected-invalid 04778 remains fail 5, shadow 3.
- Fixed-point regeneration: `.text` 0x67110, `.data` 0x5370, `.bss` 0x1c0fd020.

## Checks

- `LAKE_ARTIFACT_CACHE=false lake build EvmAsm.Codegen.Programs.ChildFrameHandlers`


pirapira:
Additional empty-route control requested during review: 03112 (`call_value_transfer_new_account`) is a plain nonzero-value CALL to a fresh no-code account, exercising the absent-account empty route. On this candidate it remains byte-perfect: fail code 0, shadow status 0, rebuilt/supplied 377/377. This directly checks that the publisher does not double-publish on the common ordinary-empty path.

pirapira:
Parent comparison for the requested ordinary empty-route control: the pre-#11039 ELF (SHA `bb2653eb...`) returns fail 44, shadow 1, rebuilt/supplied 371/377 on 03112. #11039 returns fail 0, shadow 0, 377/377. Thus this is not a duplicate publication: the broad empty-route publisher repairs a real six-byte BAL shortfall for a plain nonzero-value CALL to a fresh codeless account.

pirapira:
Second independent witness: the ordinary codeless-target control 03112 was itself a pre-patch false reject, not merely a no-regression check. Parent ELF `bb2653eb...` yields fail 44/shadow 1 and 371/377; this candidate yields fail 0/shadow 0 and 377/377. This establishes that the empty-route publisher repairs the general empty-target value-transfer class beyond the delegated-precompile witness 24639.

pirapira:
Provisional interaction check (not the registered main measurement): a throwaway local source merge from current main `dc27f7698`, with the exact #11038 and #11039 source changes applied, was regenerated and executed. 24639 is fail 0/shadow 0 at 429/429, matching the stated 395 + 27 + 7 prediction. Independent empty-route control 03112 remains fail 0/shadow 0 at 377/377. The local guest ELF SHA is recorded with the run; this will be re-run on merged main because bot rebase/regeneration may differ.

pirapira:

Issue #11045, opened 2026-07-31
Shortest-first false-reject queue from current-main random-200


pirapira:
## Scope

Measurement-only baseline of every **expected-accept** fixture that rejected in a fresh uniform current-main sample. It is deliberately ordered by potential shared leverage: largest shared `(bv_fail_code, rebuilt_len)` group first, then shortest input inside each group. The first entry in each group is the cheapest witness of that group; all labels below are the baseline set to remeasure after a fix.

- Base/source worktree: `dc27f7698f5055f1ed7b1f12b2f3b2d667393fa0` (clean)
- Selection: 200 of 26,104 individual blocks, uniformly without replacement; seed `20260731`
- Guest: `stateless_guest.elf`, SHA-256 `d332a8451b5c427f8950ecbc8b1476057004f17fde1e6e3f20dc95dcd8d24e46`
- Runner: Spike `spike_run`, SHA-256 `bae4826988540415b9bf862248caa7499b2ffbb9eec5cfab61cce00e089cf2c3`
- Result: 198 executed; 67 expected-accept false rejects in the captured baseline; 2 emulator exits excluded (not classified as rejects).
- `input bytes` is the unframed stateless-input byte length recorded in the selected manifest. `rebuilt/supplied` are `OUTPUT+128/+136`; all listed rows had shadow status 1.

## Group map

Groups of one are retained because this is the full sampled baseline, not only a shared-mechanism list.

- **18× — fail `62`, rebuilt `337`**: `00021` (5650 B), `00057` (5650 B), `00063` (5650 B), `00095` (5650 B), `00193` (5650 B), `00033` (5716 B), `00038` (5781 B), `00030` (5846 B), `00120` (5846 B), `00151` (5846 B), `00024` (5912 B), `00059` (5912 B), `00153` (5912 B), `00001` (5977 B), `00068` (5977 B), `00070` (5977 B), `00143` (5977 B), `00152` (5977 B)
- **9× — fail `62`, rebuilt `365`**: `00110` (6374 B), `00140` (6374 B), `00195` (6374 B), `00124` (6439 B), `00133` (6505 B), `00047` (6570 B), `00052` (6570 B), `00127` (6570 B), `00156` (6570 B)
- **7× — fail `53`, rebuilt `232`**: `00078` (5332 B), `00035` (5336 B), `00142` (5336 B), `00146` (5337 B), `00173` (5650 B), `00071` (8064 B), `00056` (8597 B)
[...]

PR #11064, opened 2026-07-31
fix: consult live recipient state for top-level state gas

pirapira:
## Summary

Use the current AccountState overlay before the authenticated header/BAL fallback when calculating the top-level value-recipient `NEW_ACCOUNT` state-gas charge.

The execution spec charges only when the recipient is not alive in the transaction current state ([interpreter.py:284-288](https://github.com/ethereum/execution-specs/blob/e5a8caf1b/src/ethereum/forks/amsterdam/vm/interpreter.py#L284-L288)). The old guest consulted only immutable state after the nonzero-value check, so three sequential transfers to the same previously absent recipient each charged 183,600 state gas. The overlay records that prior transfer as account-exists (flag value 2); a live record now suppresses the repeated charge, while tombstones still fall through to the existing header predicate.

This reaches the EIP-8037 accounting path that accumulates per-transaction state gas ([fork.py:1174-1182](https://github.com/ethereum/execution-specs/blob/e5a8caf1b/src/ethereum/forks/amsterdam/fork.py#L1174-L1182)).

## Evidence

- Captured base: `724c819c3`; candidate ELF SHA-256: `7113130008c5120e7116ad044d886f0de464d18ff469cb00c27696766e1ee5f5`.
- `.text` is measured and repinned at `0x67690`; `.data=0x5370` and `.bss=0x1c0fd020` are unchanged.
- Established random-200 corpus, full 256-byte output, same runner:
  - parent: OK 152 / FR 48 / FA 0
  - candidate: OK 164 / FR 36 / FA 0
  - 12 FR to OK, zero OK to FR.
- The diagnosed three-row group all now passes exactly: 00005, 00016, 00177. Each previously had three type-3 transfers to the same recipient and overcounted its new-account state charge.
- Structural gates: `scripts/check-asm-to-program.sh`, `scripts/check-region-map.sh`, and `git diff --check`.

Refs #11045.


pirapira:
Excluded from batch #11069: this branch conflicts with main in `EvmAsm/Codegen/Programs/BlockVerdictSimpleTransferGas.lean`, a hand-written source file rather than one of the generated layout files. That resolution is not mechanical, so it needs a human decision rather than an automated merge-conflict repair. The other five `merge!` PRs (#11062 #11063 #11065 #11066 #11068) went through in #11069.

PR #11068, opened 2026-07-31
fix(codegen): retain top-level CREATE sender debit through setup

### LINKED ISSUE #11067: bug(codegen): top-level CREATE consumes sender-debit seed before dispatcher setup, then MTx refund bails
## Current-main witness

On `724c819c32a137605c10cfe1f5e3bd22e68b6095`, expected-valid `03061_test_tx_gas_limit_block_boundary...type_2_create...` rejects with `bv_fail_code=53`, shadow status 1, and rebuilt/supplied BAL lengths `228/327`. The same current-main behavior appears in creation siblings 16397, 16398, and 16399.

A focused Spike trace shows that the creation helper settles the transaction and writes its indexed MTx result (`bv_mtx_gas_left[0]=1,890,550`); the later zero cumulative receipt is consequential, not the original fault.

## Mechanism

`block_verdict_creation_runtime` called `dispatcher_seed_pending_upfront_sender_balance` before `runtime_dispatcher_call`. The one-shot helper recorded only `record_nonstorage_effect` and cleared `bv_pending_upfront_balance_flag`. The callable dispatcher then performed its setup reset; its canonical depth-zero seed ran later, but the tuple had already been consumed, so no live `AccountState` sender debit existed.

MTx finalization subsequently calls `account_state_latest_balance(bv_mtx_sender_addr)` before materializing the sender refund. That lookup misses, takes `.Lbv_mtx_bail`, and skips the normal MTx publication of indexed gas results. The receipt gate consequently receives null scalar pointers and encodes cumulative gas as zero.

## Spec

Pinned Amsterdam `fork.py:1105-1108` debits the sender in `process_transaction`, before the message is prepared. `vm/interpreter.py:380` snapshots the message body only afterwards. Thus the upfront debit must be live state before the body checkpoint: later body rollback does not erase it, and settlement/refund observes it.

A repair must leave the one-shot tuple intact until the callable dispatcher's root seed (after setup), and that canonical seed must populate both the live AccountState overlay and the BAL comparison log.

pirapira:
## Spec alignment

Amsterdam execution-specs debits the sender in `process_transaction` before constructing the message (`fork.py:1105-1108`). `process_message` takes the body snapshot only afterwards (`vm/interpreter.py:380`), so the debit must be visible to settlement/refund and must survive a body rollback.

The CREATE adapter consumed `bv_pending_upfront_balance_flag` before `runtime_dispatcher_call` performed its setup reset. Its canonical depth-zero seed therefore saw an empty tuple. This change leaves consumption to that canonical post-setup seed and makes it write the debit to both AccountState (live state for settlement/refund) and the BAL comparison log.

## Evidence

Focused Spike checks on base `724c819c32a137605c10cfe1f5e3bd22e68b6095`:

- 03061, the attributed expected-valid top-level CREATE witness: `fail=0`, `shadow=0`, `327/327` (parent was `fail=53`, `shadow=1`, `228/327`).
- Creation siblings 16397, 16398, and 16399 also pass at `304/304`, `304/304`, and `314/314`.
- Expected-invalid 04778 remains rejected: `fail=5`, `shadow=3`.

No full sweep was run: the prior full current-main run is retained as the baseline, and this PR uses the directly diagnosed witnesses plus a rejection control.

## Regeneration and gates

- `scripts/gen-symbol-addresses.py --build` to fixpoint (two identical passes).
- ELF layout: `.text=0x67404`, `.data=0x5370`, `.bss=0x1c0fd020`; `RegionMap.textSizeBytes` repinned from the ELF. `bnf_le_a` remains `0xa3000ee0`.
- Every `.olean.hash` has its matching `.olean` (`missing=0`).

Fixes #11067.


pirapira:
Follow-up measurement on the PR artifact (no branch change): `16331_test_create_contract_sstore_during_init...` moves from current-main `fail=53, shadow=1, 267/335` to `fail=60, shadow=1, 335/335`. Full keccak capture matched the rebuilt length. RLP decoding gives nine rows on both sides; the sole difference is the sender `b18f64cc700a3a54a8912708245a3fa8d7d67038` balance post, rebuilt `100018324040` vs declared `99996824040`, i.e. rebuilt is exactly `21,500,000` wei high (`2,150,000 × 10`). This is an unmasked equal-length residual, not evidence that the current change is wrong; it should be re-attributed separately after this PR lands.

PR #11076, opened 2026-08-01
fix(codegen): preserve BAI length across code measurement

pirapira:
## Summary

`balSerializerMeasureCodeFunction` kept the scalar block-access-index RLP length in caller-saved `t5` across `bal_rlp_measure_into_throwaway`. That callee may clobber caller-saved temporaries, so a 256-byte `new_code` item was measured as 262 bytes instead of the required 263 bytes and the BAL was one byte short. The length is now spilled to the frame at `48(sp)` and reloaded after the call.

The execution-specs representation is a `CodeChange` carrying both `block_access_index` and `new_code` (`execution-specs` pinned at `e5a8caf1b`, `src/ethereum/forks/amsterdam/block_access_lists.py:115-131`), within the `AccountChanges` code-change field (`:174-208`). The fix preserves both components in the emitted length calculation.

## Behavioural evidence

The three live expected-valid witnesses were predicted to share this one-byte code-measurement defect and all three moved:

| fixture | before | after |
| --- | --- | --- |
| 00046 | `fail=60 shadow=1 rebuilt/supplied=650/651` | `fail=0 shadow=0 651/651` |
| 00083 | `fail=60 shadow=1 rebuilt/supplied=650/651` | `fail=0 shadow=0 651/651` |
| 00122 | `fail=60 shadow=1 rebuilt/supplied=650/651` | `fail=0 shadow=0 651/651` |

All three retained component counts `storage_changes=4, storage_reads=17, balance_changes=2, nonce_changes=3, code_changes=1, storage_reads_count=20`; only the code-change measurement length changed. Commitlog tracing on 00046 independently showed `bal_rlp_emit_bytes` returning 259 bytes for the 256-byte payload (long-form header plus payload), while the old caller recorded 262 instead of 263 because the one-byte BAI length was lost across the call.

The expected-invalid negative control 00008 remains rejected: `fail_code=5`, `shadow_status=3`.

No full EEST sweep was run; the targeted trio plus the expected-invalid control were sufficient to exercise the measured class, per the expensive-run policy.

## Verification and provenance

- `LAKE_ARTIFACT_CACHE=false lake build codegen` and stateless link passed.
- `scripts/check-build-parallel.sh` passed (codegen/link/layout, guarded-handler bytes, asm-to-program, reports, axioms, arithmetic fuzz).
- `scripts/check-asm-to-program.sh` passed: 394 converted definitions across 149 files.
- `scripts/check-region-map.sh` passed; `.text=0x677a0`, `.data=0x5370`, `.bss=0x1c0fd020`.
- `bnf_le_a=0xa3000ee0` remains unchanged.
- Candidate ELF SHA-256: `01f55ffcb675ae477851c616de8337bca6ca1fa9b7f576988745bdf852a27fc4`.
- Runner: `/home/zksecurity/evm-asm2/scripts/spike/spike_run`.
- Candidate commit: `a446501bd` (stacked on #11074's `beb56d3ac`).

No `merge!` label was added by this branch.


pirapira:
Closing as superseded: current origin/main cb8e1ac19 already contains the equivalent fix as 7275ee3cd via the merged #11070 stack. No merge! label was added.

Issue #11048, opened 2026-07-31
harness: the EEST comparator checks 105 bytes of a 256-byte output, so every byte-identical / set-identical claim is bounded by that prefix


pirapira:
## Summary

Every A/B and "byte-identical output" claim made through
`scripts/codegen-eest-stateless-check.sh` is bounded by a **105-byte comparison over a
256-byte output**. A guest change can alter the output region, leave the compared prefix
untouched, and be reported as `full match` on every row — which reads as equivalence and
is not.

**This is a fact about the harness, not about one PR.** It bounds today's structural-step
measurements, mine and other agents'.

## The window, both ends measured

| region | bytes | who reads it |
|---|---|---|
| `SszStatelessValidationResult` | **0..104** (105) — 73 for the deserialize-failure sentinel | the harness's `full` / `root` / `succ` / `tail` comparison |
| runner capture | **0..255** (256) | `scripts/spike/spike_run`, `DEFAULT_OUTPUT_LEN = 256` (`spike_run.cc:34`), overridable via `SPIKE_OUTPUT_LEN` |
| written by the guest | **up to offset 248 inclusive** | `NoopHalt.returnRevertTail`'s halt tail |

So **151 of the 256 captured bytes are never compared**, and the capture length is itself a
runner default rather than a property of the guest — raising `SPIKE_OUTPUT_LEN` can only
widen the unchecked region, never narrow it.

## ⛔ The witness: two builds, identical prefix, different output

Base `dc27f7698` (guest sha256 `d332a8451b5c…`) vs a one-hunk codegen change (guest
`63dd22d1043e…`), same input
(`blockchain_tests/for_amsterdam/ported_static/stWalletTest/day_limit_construction/day_limit_construction.json`,
case `00000_…--g0__b0`):

* **first 105 bytes: byte-identical** (`cmp -n 105` clean) — so the harness reports
  identical classification on this row, and did: 195 rows, `newly failing 0`,
  `newly passing 0`, failing sets identical.
* **byte 248 differs: `0xb0` (176) at base, `0x01` on the change.**

The writer is the same instruction in both builds — `sd x21, 248(x16)` with
`x21 = min(x15, 176)`, at `0x800515a0` base / `0x800515c0` change (the `0x20` `.text`
repin). Five writes each (five depth-0 halts); the first four write `0` in both; **the
fifth differs.**

[...]

Issue #10998, opened 2026-07-31
codegen: cd_xfer_log_pending is a deferred one-shot flag with two consumers per CALL mode where the spec has one emit_transfer_log


pirapira:
## Summary

`cd_xfer_log_pending` is a **deferred one-shot flag with two consumers per CALL mode**, where
execution-specs has **one** `emit_transfer_log` call. The correctness of the arrangement rests on
a **stated** reset rather than on a structural guard, which is the same shape as GH #10925.

Not a demonstrated defect — filed because the *guard is prose*, and this repo has been bitten by
that class before.

## The asymmetry, counted in the emitted stream

Base `3cd551e21`, `gen-out/regionmap/stateless_guest.s` (sha256 `2e85e1a6…`).

**Spec:** exactly one `emit_transfer_log`, inside `process_message`
(`src/ethereum/forks/amsterdam/vm/interpreter.py:391-397`, pinned `e5a8caf1b`), guarded by
`should_transfer_value and value != 0` then `caller != current_target`.

**Guest:** the flag is consumed at **two sites per mode** — the empty-target path and the
descend path:

| mode | `.Lcd_xlog_*_empty_*` | `.Lcd_xlog_*_desc_*` | staging labels |
|---|---|---|---|
| CALL | `.s:60993` | `.s:61084` | 10 |
| CALLCODE | — (same shape) | — | 10 |
| DELEGATECALL | — | — | 10 |
| STATICCALL | — | — | 10 |

So **eight consumer sites** across the four modes for one spec call site.

## Why the guard is the concern rather than the count

The one-shot discipline is `cd_xfer_log_pending` being **cleared before the emit** at each site,
plus a **reset per CALL** so a prior CALL's pending transfer cannot leak into the next. That reset
is described in the emitter's own comment (`git show 3cd551e21:EvmAsm/Codegen/Programs/ChildFrameHandlers.lean`):

> *"EIP-7708 value-CALL transfer log emission is DEFERRED (see below); reset the per-CALL pending
> flag so a prior CALL's value transfer does not leak into this one."*

That is **design-intent prose describing a guard**, not a mechanism that fails closed. GH #10925
was the same single-buffer model: `CALLER` and `ORIGIN` staged on one path and not the other while
a shared comparator read them anyway.

## What would make this safe without changing behaviour

Either of:

* make the flag's clear **structurally** paired with its set (so a set with no matching consumer is
  impossible), or
[...]

PR #11079, opened 2026-08-01
codegen: emit the deferred transfer-log consumer only where the flag can be set (#10998)

pirapira:
## What this does

`cd_xfer_log_pending` had **eight consumer sites against one setter**.

The setter is `.Lcd_tl_notself_*`, reached only by **mode 0**, because `execution-specs` gates
`emit_transfer_log` on `should_transfer_value and value != 0` and then
`caller != current_target` (`vm/interpreter.py:391-397`, pinned `e5a8caf1b`): DELEGATECALL and
STATICCALL carry no value, and CALLCODE targets the caller's own context so the transfer is
self-to-self. **CALLCODE, DELEGATECALL and STATICCALL each carried two full deferred-emit blocks
for a log they can never emit.**

This makes the reachability argument the **emission condition** rather than a comment: the
consumer is emitted for mode 0 only. #10998 asked for *either* a structural guard *or* a
written-out reachability argument — this is both, because the argument is now what decides
emission.

⚠️ **The per-CALL reset is deliberately kept in all four modes.** It is what stops a set from
surviving into a later CALL, it is a different site from the paired clear, and removing it is how
a leak would go from unconsumable to reachable.

## Why the removal is safe — and why it is not merely equivalent

If no leak exists, those six blocks were unreachable and removing them is behaviour-identical. If
a leak ever occurred, then **today** a DELEGATECALL or STATICCALL would emit a transfer log the
spec forbids for it. So the cut is neutral in the good case and a latent false-log fix in the bad
one; it cannot make things worse.

## Measurements

Clean `main` `37da3b456` (guest `aa9af7753b74176a`) vs this branch (`e18cfe3111df8a76`).

| | main | change |
|---|---|---|
| `cd_xfer_log_pending` references | 22 | **10** |
| `eip7708_append_transfer_log` call sites | 14 | **8** |
| `.Lcd_xlog_skip` labels | 16 | **4** |
| `.text` | `0x067830` | **`0x067290`** (−0x5a0) |

The survivors are exactly the intended set: **four resets** (one per handler), **one set**, and
**two test+clear pairs in `h_CALL` only**.

### ⚠️ The byte delta is larger than the flag references explain, and that needed an answer

1440 bytes is ~360 instructions; fifteen flag references cannot account for it. Gating the
**emitter** removes six whole blocks, not just their guards. Each removed block is 30
instructions: the test, the paired clear, a 128-byte frame push saving `x10`/`x12`/`x13`, three
reverse-copy operand stagings into `sp+0` / `sp+32` / `sp+64`, the `eip7708_append_transfer_log`
call, then the restore and the pop. Six blocks × 30 reconciles with both the byte delta and the
−6 append call sites.

⭐ **And I read one removed block line by line rather than trusting the count**: the frame push
*and* its matching pop are both **inside** the guard, so the block was balanced on both paths and
left no register or stack effect when skipped. Removing it is therefore exactly equivalent to the
flag always being zero — which for those three modes it provably is. A count could not have shown
that; only the contents could.

### Corpus

Random-200 at captured seed `1663874068`, paired: **both legs 200 ran / 0 errored / 13 fail / 187
full match**, and all 200 outputs **byte-identical across the full 256 bytes** (#11048).

### Gates

`lake build` exit 0 with zero errors **and** `check-region-map` exit 0 — both captured, neither
inferred from the other.

⚠️ Worth recording for the next layout-shrinking change: the first repin attempt **built red
while the regen reported exit 0**. Shrinking `.text` left **two of 406 `GuestAddrs` pins past the
new end**, so `GuestImage`'s `extentsOkFrom` decided false. `check-region-map.sh` writes only the
TSV despite its banner mentioning GuestAddrs and GuestImageEntries, so
`asm_to_program.py guest-addrs` and `guest_image_coverage.py --emit-lean` had to be run
explicitly.

## ⚠️ Not claimed

The removed blocks staged `cd_callee_be`, which the emitter's own comment says is **not populated
until code resolution** and would be **stale from a prior CALL** — which is exactly why the
*setter* uses `nse_callee_be` instead. So the unreachable code was also wrong. **I have not
checked whether the surviving CALL-mode pair has the same staleness**; that is a separate
question and possibly a separate issue.


pirapira:
## Context for the `13 fail`: the baseline moved overnight, on the same seed

A reviewer seeing `13 fail` with no context cannot tell whether that is good. On the **identical
selection** (captured seed `1663874068`, limit 200):

| | yesterday `dc27f7698` | today `37da3b456` |
|---|---|---|
| ran | 195 | **200** |
| errored | 5 | **0** |
| fail | **73** | **13** |
| full match | 122 | **187** |

**None of that is this PR.** It is the overnight stack landing on `main`. Both legs of the
measurement above are on `37da3b456`, so the pairing is valid — but any fail count quoted
yesterday is **not** comparable to one quoted today, and `73 → 13` would misread as this change's
doing.

⭐ It also makes the neutrality claim **stronger** rather than weaker: a null against a baseline
with 13 failures and 0 emulator exits leaves far less noise for a regression to hide in than one
against 73 failures and 5 exits.

Issue #11081, opened 2026-08-01
codegen: CALL does not credit the child frame-local balance at env+32 (SELFBALANCE stale after nested value-CALL)


pirapira:
## Summary

After a nested value-bearing `CALL`, the **callee frame's** live balance cell `env+32` (`.selfBalance`) is never credited with `call.value`. The parent is debited before descend; `record_message_value_transfer` only journals nonstorage/BAL effects. **CREATE/CREATE2 already credit the child** (`ChildFrameCreateTail.lean` ~434–458, drj99.1 part 2). CALL was the only value-transferring path missing the mirror.

This is a **wrong-value execution defect**, not merely a BAL symptom: contract code that runs `SELFBALANCE` after receiving a value-CALL reads a stale balance.

## Loud reader: `h_SELFBALANCE`

In the baseline guest (`sha256 aa9af7753b74176a2e30d31d39e816da385601622b29388981f6f00fd6e7dacf`):

| site | role |
|---|---|
| **`h_SELFBALANCE` @ `0x8004d694`** | **READ** — `ld a5, 32/40/48/56(s4)` pushes `env+32` LE limbs onto the EVM stack (opcode `0x47`) |
| `h_CREATE` / `h_CREATE2` | READ creator bal gate; WRITE creator debit; **WRITE child endowment credit** |
| `h_CALL` | READ/WRITE caller debit; **was missing child credit** |
| `h_CALLCODE` | READ balance gate (no value move to another account) |
| `h_SELFDESTRUCT` | READ live bal paths that also consult `env+32` |
| `h_RETURN` | create post-balance paths via env |
| `call_frame_descend` | WRITE stage child `env+32` from **pre-transfer** table/overlay only |
| `create_frame_descend` | WRITE stage child `env+32` |
| `BlockVerdictDispatchTx` | WRITE top-level tx recipient `env+32` + coc3g.1 credit |

So after any nested value-CALL, **`SELFBALANCE` in the callee sees pre-transfer balance** (typically 0 for a fresh/unfunded callee), not `pre + value`.

## Mechanism (pinned)

1. `CallFrameDescend` stages child `env+32` from `callee_balance_table` / `account_state_latest_balance` / header — **pre-transfer**. Live overlay runs before the transfer record, so it cannot see this CALL's credit.
2. `ChildFrameHandlers.lean` post-descend `recordMessageValueTransferAsm` records BAL nonstorage only — **does not write child `env+32`**.
[...]

PR #11082, opened 2026-08-01
fix(codegen): credit child env+32 on nested value-CALL

pirapira:
## Summary

Fixes nested value-`CALL` never crediting the **callee** frame-local live balance at `env+32` (`.selfBalance`). Parent was already debited pre-descend; only nonstorage/BAL was recorded post-descend. CREATE/CREATE2 already credit the child (`ChildFrameCreateTail` ~434–458); CALL was the missing path.

Closes the wrong-value execution defect where **`SELFBALANCE` (and the callee CREATE endowment gate)** see a stale balance after a nested value-CALL. Tracked as **#11081**.

The child-credit sequence was deliberately mirrored from the working CREATE path, then **register liveness was re-derived at the CALL site from emitted asm** (not inherited from CREATE). That caught a save-set/write-set mismatch in the copied prelude; fixed in a follow-up commit (see below). CREATE’s own copy of the trap is filed separately as **#11083** (latent; no failure attributed).

## Change

`EvmAsm/Codegen/Programs/ChildFrameHandlers.lean` — after `recordMessageValueTransferAsm` on CALL mode-0 post-descend:

- if `cd_value_be ≠ 0` and not self-call: reverse child `env+32` LE→BE, `u256_add_be cd_value_be`, write back LE
- save/restore **x10/x11/x12** around the helper (match a0/a1/a2 write set; RV ABI, not the CREATE comment’s x10/x12/x13)
- inside the child rollback interval (reverting child discards the frame with the credit)

## Readers of `env+32` (why this is serious)

| site | role |
|---|---|
| **`h_SELFBALANCE` @ `0x8004d694`** | **READ** opcode 0x47 — stale after nested value-CALL |
| `h_CREATE` / `h_CREATE2` | gate + creator debit + **child credit** (already present) |
| `h_CALL` | caller debit; **child credit was missing** |
| `h_CALLCODE` / `h_SELFDESTRUCT` / descend staging / top-level tx credit | other readers/writers |

## Same-population evidence (family `selfdestructing_initcode_preserves_balance`, limit=40)

| | guest sha256 | RAN | full pass | fail | root match | run dir |
|---|---|---:|---:|---:|---:|---|
| **baseline** | `aa9af7753b74176a2e30d31d39e816da385601622b29388981f6f00fd6e7dacf` | 40 | **16** | **24** | 40 | `run-20260801T111722Z-2916928` (`--guest-elf`) |
| **fix (pre save-set fix)** | `e3907f356d71c0b118f0e301249c1e4b9ad2b827bc742c001fe41da1bfa1916d` | 40 | **28** | **12** | 40 | `run-20260801T110956Z-2896086` |
| **fix (this PR HEAD, regenerated)** | `7c630c93fb53684b1c0f25d3a7873307bfdfe47fa9180c3ed87e4d8d30491337` | 40 | **28** | **12** | 40 | `run-20260801T112401Z-2945520` (built) |

Net vs baseline: **+12 full pass / −12 fail** on the identical 40-row population. Save-set correction did not change the score (as expected: x11/ra dead at site). All remaining fails are false rejects (`succ guest=00 exp=01`) with matching state root.

Reviewer rebuild: checkout this branch, run
`scripts/codegen-eest-stateless-check.sh --backend spike --filter selfdestructing_initcode_preserves_balance --limit 40 --jobs 4`
and confirm guest sha256 `7c630c93…1337`.

Witnessing path fixed first: `initial_balance=1` + factory CREATE after value-CALL (CREATE gate saw `create_balance_be=0`).

## Residuals (not in this PR)

- **Class A** (`bv60`, short Δ3): `post_send_opcode=CALL` ∧ `post_send_count∈{1,3}` ∧ `initial_balance=1`. Hypothesis: parent `env+32` debit not re-credited when child fails — reverse edge; overlaps **#11001**. Pin by reading `env+32` at the boundary before fixing.
- **Class B** (`bv44`): `post_send_opcode=SELFDESTRUCT` ∧ `post_send_count∈{1,3}`. On `ibal=0` lengths equal → per-entry BAL decode, not byte-diff.

Related: #11001, #11081, #11083 (CREATE save-set hygiene).

## Test plan

- [x] Same-pop baseline vs fix limit 40 on `selfdestructing_initcode_preserves_balance`
- [x] Re-measure after save-set fix on regenerated ELF `7c630c93…`
- [ ] CI eest / codegen gates
- [ ] Follow-up: pin Class A parent restore; per-entry decode Class B

### DIFF
--- a/EvmAsm/Codegen/Programs/ChildFrameHandlers.lean
+++ b/EvmAsm/Codegen/Programs/ChildFrameHandlers.lean
@@ -1039,16 +1039,18 @@ def callDescendFallThrough
      "  lbu t3, 0(t0); sb t3, 0(t1)\n" ++
      "  addi t0, t0, -1; addi t1, t1, 1; addi t2, t2, -1\n" ++
      "  bnez t2, .Lcd_child_sb_rev_" ++ tag ++ "\n" ++
-     "  addi sp, sp, -32\n  sd x10, 0(sp)\n  sd x12, 8(sp)\n  sd x13, 16(sp)\n" ++
-     "  la a0, cd_caller_newbal\n  la a1, cd_value_be\n  la a2, cd_caller_newbal\n" ++
-     "  jal ra, u256_add_be\n" ++
-     "  ld x10, 0(sp)\n  ld x12, 8(sp)\n  ld x13, 16(sp)\n  addi sp, sp, 32\n" ++
-     "  la t0, cd_caller_newbal\n  addi t1, x20, 63\n  li t2, 32\n" ++
-     ".Lcd_child_sb_wb_" ++ tag ++ ":\n" ++
-     "  lbu t3, 0(t0); sb t3, 0(t1)\n" ++
-     "  addi t0, t0, 1; addi t1, t1, -1; addi t2, t2, -1\n" ++
-     "  bnez t2, .Lcd_child_sb_wb_" ++ tag ++ "\n" ++
-     ".Lcd_child_sb_done_" ++ tag ++ ":\n") ++
+      -- Save set = write set of the helper args: a0=x10, a1=x11, a2=x12.
+      -- (Inherited CREATE comment claimed a0-a2 alias x10/x12/x13 — wrong.)
+      "  addi sp, sp, -32\n  sd x10, 0(sp)\n  sd x11, 8(sp)\n  sd x12, 16(sp)\n" ++
+      "  la a0, cd_caller_newbal\n  la a1, cd_value_be\n  la a2, cd_caller_newbal\n" ++
+      "  jal ra, u256_add_be\n" ++
+      "  ld x10, 0(sp)\n  ld x11, 8(sp)\n  ld x12, 16(sp)\n  addi sp, sp, 32\n" ++
+      "  la t0, cd_caller_newbal\n  addi t1, x20, 63\n  li t2, 32\n" ++
+      ".Lcd_child_sb_wb_" ++ tag ++ ":\n" ++
+      "  lbu t3, 0(t0); sb t3, 0(t1)\n" ++
+      "  addi t0, t0, 1; addi t1, t1, -1; addi t2, t2, -1\n" ++
+      "  bnez t2, .Lcd_child_sb_wb_" ++ tag ++ "\n" ++
+      ".Lcd_child_sb_done_" ++ tag ++ ":\n") ++
    -- fva3w: x20 now = CHILD env (call_frame_descend repointed it; eventLogCheckpoint@480 is
    -- set to the current count). Emit the deferred EIP-7708 value-CALL transfer log HERE so it
    -- lands in the child frame's logs: frame_return rolls it back on a child REVERT/exceptional
--- a/EvmAsm/Codegen/Programs/ChildFrameHandlers.lean
+++ b/EvmAsm/Codegen/Programs/ChildFrameHandlers.lean
@@ -1007,17 +1007,53 @@ def callDescendFallThrough
   -- nse_callee_be, cd_value_be, cd_balance_be, and nse_acct are static scratch
   -- reused by nested calls, so they are valid only in this post-descend,
   -- pre-dispatch window.
-  (if mode != 0 then "" else
-    -- GH #10938: the setup is the shared `recordMessageValueTransferAsm`, which the
-    -- CREATE-endowment site also uses.  The spec has ONE `move_ether` for both, since
-    -- `process_create_message` delegates to `process_message` (`interpreter.py:212`).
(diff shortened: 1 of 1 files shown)


pirapira:
## Register answer (emitted asm, fix-leg `stateless_guest.s`)

Insertion at `.Lcd_child_sb_credit_call_target` (~61572), immediately after `record_message_value_transfer` restore and before `emitPendingXferLog` / `j .runtime_tx_child_message_entry`.

### What the helper touches

`u256_add_be` (@4294): reads a0/a1/a2 as base pointers, uses t-temps (x5–x7, x28–x31), writes carry into x10, returns `jalr x0, 0(x1)`. **Write set of the call setup = x10, x11, x12. Does not touch x13.**

### liveness at THIS site (not CREATE)

| reg | live? | evidence |
|---|---|---|
| **x11** | **dead** | The immediately preceding block at the same site already does `la a1, nse_callee_be` / `jal ra, record_message_value_transfer` **without saving x11** (same mismatched save of x10/x12/x13). If x11 were live, that pre-existing call would already have clobbered it. Nothing between that restore and my `la a1` reloads x11. |
| **ra (x1)** | **dead** | Path after the insertion is `emitPendingXferLog` (another `jal ra, eip7708_append_transfer_log`) then **`j .runtime_tx_child_message_entry`**, which is an alias of `.Ldispatch_resume` — a **jump**, not `ret`. ra is not a return edge here; prior jals already clobbered it. |
| **x10** | preserved by convention | Neighboring saves keep post-descend x10 across helpers; dispatch_resume itself only uses x5–x7, but we keep saving x10 to match the local window. |
| **x12** | **live** | EVM stack pointer after `call_frame_descend`. `la a2, cd_caller_newbal` would replace it; must save/restore. |
| **x13** | live but **not written** by this helper | EVM mem base; no need to save for `u256_add_be`. |

### Fix applied

Save/restore set corrected to **x10, x11, x12** (match write set). Pushed as follow-up commit on this PR. Leaving x11 in the save set even though dead so the next editor is not trapped.

### CREATE precedent

CREATE child-credit at `.Lcr_sbc_rev_f0` (~57544) has the **identical** x10/x12/x13 vs a0/a1/a2 mismatch and the wrong “a0–a2 alias x10/x12/x13” comment. Not touched here per review. If you want a separate hygiene issue for that latent trap I can file it; I did not verify CREATE-site x11 liveness beyond noting the same pattern.

pirapira:
## Regenerated after save-set fix

| | sha256 | full | fail |
|---|---|---:|---:|
| baseline | `aa9af775…dacf` | 16 | 24 |
| fix pre-saveset (`e3907f35…`) | | 28 | 12 |
| **fix HEAD rebuilt** | **`7c630c93fb53684b1c0f25d3a7873307bfdfe47fa9180c3ed87e4d8d30491337`** | **28** | **12** |

Run: `gen-out/eest-run/run-20260801T112401Z-2945520/` (built, not `--guest-elf`). Identical 28/12; save-set change is register hygiene only at this site.

Also filed **#11083** for the CREATE latent save-set/comment trap. PR body updated with rebuild sha and the “mirror then re-derive registers” note.

pirapira:
Approved.

**Diagnosis.** Root-caused from a BAL symptom down to a wrong-value execution defect: `call_frame_descend` stages the child's frame-local `selfBalance` at `env+32` from pre-transfer state, and `h_CALL` was the only value-transferring path that never credited it. `h_SELFBALANCE` (`0x8004d694`) reads that field, so nested contract code observed a stale balance; the CREATE endowment gate compared `0 < 1` and aborted the create, which is how the family surfaced.

**Evidence.** Same-population A/B on the full 40-row family — baseline `aa9af775` 16 full / 24 fail, candidate 28 full / 12 fail, `root` matching 40/40 on both legs. The earlier 8-of-16 baseline was a truncated population and was correctly discarded rather than reported as a before/after.

**Register review.** The block mirrors `ChildFrameCreateTail`, whose save set (`x10/x12/x13`) does not match its write set (`a0/a1/a2` = `x10/x11/x12`). Rather than inherit that, the author re-derived liveness at this site — `x11` dead, `ra` dead (the fallthrough is a `j` to the child message entry, not a `ret`), `x12` live — and corrected the save set to match. The latent precedent is filed as #11083 with no failure attributed. This is the right way to reuse a path here: mirror, then re-derive the register assumptions at the new site.

**Regeneration.** The save-set fix changed emitted assembly after the first measurement, so the artefact was regenerated (`7c630c93fb53684b1c0f25d3a7873307bfdfe47fa9180c3ed87e4d8d30491337`) and the 40 rows re-run identically. The sha in the body is rebuildable.

Residual classes A and B are new findings uncovered *by* this fix and are correctly not held against it.

PR #11084, opened 2026-08-01
fix(codegen): roll back auth effects only on auth-phase OOG

pirapira:
## Summary

The execution-specs transaction-message path snapshots transaction state before authorization/preparation and restores that snapshot only when the authorization phase itself raises an exceptional halt. A later body revert retains the authorization effects. This patch makes that phase boundary explicit in the guest:

- checkpoint `account_state_pending_count` and the append-only non-storage-effect cursors at the authorization boundary;
- set `runtime_tx_auth_phase_halted` only for an authorization callback failure or authorization state-gas spill OOG;
- restore the checkpoints only when that marker is set;
- leave unrelated pre-preparation exits and body reverts on the normal commit path.

Spec reference: execution-specs pinned at `e5a8caf1b8055e4d805c7fb169edfa710914b7da`. `vm/interpreter.py:121-136` applies authorizations before entering `process_message`; `vm/interpreter.py:261-302` snapshots and rolls back the message body; and `fork.py:939-945` consumes the message result. Thus a body revert rolls back the body snapshot but retains prior authorization effects, while an exception during `set_delegation` aborts before the message-body snapshot and the transaction-local state is discarded. The authorization implementation is `vm/eoa_delegation.py:194-230`.

This is the narrow auth-effects gate. The attempted recipient-read suppression was removed after the controlled A/B showed 13 unrelated reverse transitions; that work remains parked rather than being folded into this PR.

## Evidence

The controlled measurement was taken before the later main merge, on measurement base `37da3b456ac19cf50d12c26004a1c196227af2f3`:

- parent ELF `aa9af7753b74176a2e30d31d39e816da385601622b29388981f6f00fd6e7dacf`;
- candidate ELF `b62eca2d69aff0a94a7ad55e21e35d865b7bc5de78bf0ba3db976e4e5c86ac63`;
- runner `/var/tmp/scoper-10938-base/scripts/spike/spike_run`;
- deterministic 200-row x998 manifest, `SPIKE_OUTPUT_LEN=256`, 12 workers.

Both legs were `FA=0, FR=13, OK=187`; the candidate had zero forward transitions and zero reverse transitions relative to the parent. The 13 FRs are pre-existing shared failures, not regressions or gains from this change. The parent leg was independently rerun twice with the same settings: both TSVs were byte-identical to each other and to the recorded parent TSV, so the parent result is reproducible. The x998 manifest SHA-256 is `5a45cf963a759a6257e7e4c1d7e691f83c71542ebbc641a77c3913f71b565aa8`.

After merging current main (`66ef447a11e0d2705ed0a5abbb73afb6d6ecbed8`), the final linked guest is `8f22d5d6d2541284af5753d5ba3749e8b5bcc1d4223d0c95c6c91d155917e02d` with `.text=0x6755c`, `.data=0x5370`, `.bss=0x1c10fc20`; `bnf_le_a` remains `0xa3000ee0`. The merged-base targeted controls were:

- auth-phase dispatch halt: `00019`, `succ=1`, `fail=0`, `shadow=0`, `394/394`;
- body-revert control: `00020`, `succ=1`, `fail=0`, `shadow=0`, `393/393`;
- expected-invalid auth-floor control: `00052`, `succ=0`, `fail=5` (still rejects);
- `00032` remains a known BAL residual (`succ=0`, `fail=60`, `426/364`) and is not claimed as fixed here.

The post-merge artifact was then A/B-tested against current main (`66ef447a1`) using the same x998 manifest, runner, 256-byte capture, and 12 workers. Current-main parent ELF SHA-256 is `e18cfe3111df8a76f18a4d0d09d9e89cdf53bcc25133937cb0638f3b4cc0b778`; candidate is the final ELF above. Both legs were `FA=0, FR=13, OK=187`, the TSVs were byte-identical, and both transition lists were empty.

## Checks

- `scripts/check-asm-to-program.sh` — clean (394 converted defs across 149 files).
- `ELF_DIR=gen-out/regionmap NO_BUILD=1 scripts/check-region-map.sh` — clean; layout regenerated to fixpoint.
- no `merge!` label was applied by this branch.


pirapira:
Approved.

**The claim.** The spec restores the transaction-state snapshot only when the authorization phase itself raises an exceptional halt; a body revert retains authorization effects (`vm/interpreter.py:121-136`, `:261-302`; `fork.py:939-945`; `vm/eoa_delegation.py:194-230`, pinned `e5a8caf1b`). This makes that phase boundary explicit in the guest instead of leaving it implicit in the commit path.

**Why the controls carry the review.** With zero forward and zero reverse transitions, the 200-row A/B cannot by itself distinguish a correct-and-inert change from dead code. Two things close that gap: the snapshot and restore were traced executing on fixture `00056` (store PC `0x80035180` and load PC `0x800666a8` against the same address `0xbaeaa1d8`), and the targeted controls exercise both sides of the boundary — `00019` auth-phase halt and `00020` body revert both `succ=1`. The body-revert control is the one that matters, since retaining auth effects across a body revert is the property being asserted.

**Measurement.** Parent `e18cfe31…` at main `66ef447a1` vs the final post-merge candidate `8f22d5d6…`, same runner, same x998 manifest (`5a45cf96…`), `SPIKE_OUTPUT_LEN=256`, 12 workers: both `FA=0 FR=13 OK=187`, TSVs byte-identical, forward and reverse empty. The earlier pre-merge legs were reconciled first — the parent was rerun twice byte-identically, and an earlier "200 OK" report was withdrawn as a misstatement rather than instrument drift.

**Scope discipline.** The recipient-read suppression was removed after its own A/B produced 13 reverse transitions across unrelated families (precompile, blob prefund, CREATE2, timestamp, cross-tx) — the signature of an over-broad trigger rather than a wrong mechanism. That work is parked with the refutation recorded, not folded in here. `00032` is left as a known BAL residual and is explicitly not claimed.

Landing this as behaviour-neutral is the right call: it is a spec-alignment change whose value is the explicit phase boundary, and it is now measured against the artefact it ships.

Issue #11083, opened 2026-08-01
codegen: CREATE child env+32 credit save set mismatches a0-a2 (wrong comment says x10/x12/x13)


pirapira:
## Summary

In `ChildFrameCreateTail.lean` (emitted around `.Lcr_sbc_rev_f0` / child endowment credit after `create_frame_descend`), the save/restore around `jal ra, u256_add_be` is:

```
sd x10, 0(sp)
sd x12, 8(sp)
sd x13, 16(sp)
la a0, create_creator_newbal
la a1, create_value_be
la a2, create_creator_newbal
jal ra, u256_add_be
ld x10 / x12 / x13
```

**Write set of the call setup is a0/a1/a2 = x10/x11/x12.** The save set is x10/x12/x13: **x11 is never saved, x13 is saved though the helper does not write it.**

A nearby comment claims a0–a2 alias x10/x12/x13 — that is simply wrong (RV ABI: a0=x10, a1=x11, a2=x12). That comment is what propagated the same mismatched save into the CALL child-credit path.

## Status

- **No failure is attributed to this site.** It is latent: x11 is presumably dead at the CREATE credit point (or we would have seen clobber bugs), same as the CALL site analysis on #11082.
- **#11082** fixed the CALL copy to save x10/x11/x12 matching the write set, after re-deriving liveness at the CALL site from emitted asm rather than inheriting CREATE’s registers.
- This issue is hygiene: correct the CREATE save set and delete/fix the misleading comment so the next mirror does not re-copy the trap.

## Suggested fix

Mirror #11082: `sd/ld x10, x11, x12` around the CREATE `u256_add_be` child-credit (and any sibling CREATE2 path with the same pattern). Confirm x11/ra liveness at the CREATE site from emitted asm before landing if anything beyond the save-set hygiene is unclear.

pirapira:
## Liveness + comment fix

**x11 is dead** across L182 / L258 / L275 / L511 (`.Lcr_sbc_rev_*`). Save sets are dispatcher invariants (x10/x12/x13), not a0-a2 write-set mirrors. B/C correct for a0-only arity. D had the false `a0-a2 alias x10/x12/x13` claim — fixed in comment only; no ELF delta.

PR: (opening)

pirapira:
PR: https://github.com/Verified-zkEVM/evm-asm/pull/11811 — comment-only; x11 dead; no save-set/ELF change.

PR #11085, opened 2026-08-01
fix(codegen): record the resolved child delegation target

pirapira:
## Summary

Child delegation access must charge and record the address selected by the resolver. The previous path always used the same-block scratch pointer, so a resolver status indicating a prior-block/header delegation could charge and record an all-zero or unrelated address.

The fix selects the pointer from the resolver status before either operation:

- status 0: use `bsbd_deleg_target`, the same-block transaction-state delegation target;
- status 1: use the restored header marker target (`t3 + 3`), the prior-block delegation target;
- status 2 or any other status: skip the delegation-specific path.

The selected pointer is saved across the access-charge call and passed explicitly to both the charge and final `account_read_record` calls. This explains both observed signatures: a wrong target can substitute an equal-width RLP entry (400/400), or change the selected account's encoding enough to produce the one-byte shortfall (399/400).

The execution-specs path resolves the delegated code address and adds it to the accessed-address set before execution (`execution-specs/src/ethereum/forks/bpo4/vm/interpreter.py:124-134`); the guest now preserves that target identity through the child access-charge and BAL-recording sequence.

## Evidence

The candidate was rebuilt on a clean base (`66ef447a1`) and all evidence below was re-measured on that artefact; no earlier dirty-branch result was carried forward.

- The two-by-two family split (delegate warm/cold × target warm/cold) showed the resolver-status/pointer selection was the discriminator; status-branch-only was insufficient.
- A child-path trace showed `bsbd_deleg_target` was unfed (24 zero bytes) on the prior-block status path, while the restored header marker contained the target.
- A bounded emitted-ASM audit covered all 30 `account_read_record` sites; no other inherited-`a0` site had an intervening call that could invalidate its address.
- Targeted family: 16/16 expected-valid fixtures now pass; FA=0.
- Paired 200-row A/B on the same runner and invocation (30 workers): parent FA=0, FR=13, OK=187; candidate FA=0, FR=11, OK=189. Transitions: 187 OK→OK, 2 FR→OK, 0 OK→FR. Forward flips: `00143` (`bal_7702_delegated_via_call_opcode_CALLCODE`) and `00183` (`bal_delegatecall_7702_delegation_and_OOG`).
- Candidate ELF SHA-256: `f8c8e36b73a7075c83b1555a135e50a22e1893e0a6712b9ba4226bc9f7ed30b2`.
- Parent ELF SHA-256: `e18cfe3111df8a76f18a4d0d09d9e89cdf53bcc25133937cb0638f3b4cc0b778`.
- Spike runner SHA-256: `c6ac09fa72c2103b6d3e390b779b022ef7f220e5a39bb5f667114a89d78e5868`.
- Layout after measured regeneration: `.text=0x6728c`, `.data=0x5370`, `.bss=0x1c0fd040`.

Full cache-disabled `lake build`, `check-asm-to-program.sh`, and `check-region-map.sh` pass.


pirapira:
Approved.

**Diagnosis quality.** This was rooted by measurement at every step, and two intermediate results are worth preserving because they each killed a hypothesis:

- The `a0` reload alone left the family unchanged — the trace then showed `bsbd_deleg_target` holding 24 zero bytes on the child path. A correct-but-unfed mechanism, not a missing reload.
- The predicate-inversion probe moved row 00000 from `fail60 400/400` to `fail53 373/400`, proving the status branch alone was not the selector. That negative is the most instructive step in the diagnosis.

**Why the signatures fit.** Delegate warmth discriminates perfectly across the sixteen rows (cold → 60, warm → 53; the target axis does not discriminate at all). A wrong address of the same width substituted for the correct one is exactly why rebuilt equalled supplied at 400 — the equal-length case that a byte diff is structurally blind to. The length equality is evidence *for* this mechanism rather than an obstacle to it.

**Why this is not overfitted.** The concern with a single change closing two differently-shaped signatures is that three statuses and two candidate pointers give enough freedom to fit sixteen rows. The answer is in the 200-row A/B: **two forward flips in a different family** — 00143 `bal_7702_delegated_via_CALLCODE` and 00183 `bal_delegatecall_7702_delegation_and_OOG` — both sharing the mechanism, neither targeted. A tuned selector does not do that. Each arm is also justified by what its status *means* rather than by which pointer produces the right byte count.

**Evidence.** 16/16 family green (`FA=0 FR=0 OK=16`, full 256-byte match). Paired 200-row A/B, both legs same runner and invocation, candidate rebuilt on clean base and re-measured on that artefact: parent `e18cfe31…` `FA=0 FR=13 OK=187`, candidate `f8c8e36b…` `FA=0 FR=11 OK=189`, 2 forward, 0 reverse, no false accepts.

**Bounded negative worth keeping:** 30 `account_read_record` sites audited in emitted assembly, classified three ways (direct sites establishing `a0` immediately, wrapper sites intentionally inheriting the API `a0`, `account_read_log` forwarding `a2`). No other intervening-call inherited-`a0` instance exists. The three-way classification is what makes this usable — a two-way audit would have flagged the wrappers as false suspects.

**Follow-up, not blocking.** This function writes `a1` (`x11`) while saving `x10/x12/x13`, the same save-set/write-set mismatch #11083 records for `ChildFrameCreateTail`. Pre-existing here and unaffected by this change, with no regression in the 200-row run — worth adding to #11083 as a second instance rather than fixing here.

pirapira:
**Additional evidence: this PR also closes a second, separately-filed-nowhere family.**

`delegate_call_targets` was picked up as the next uncited lane and re-derived on current main (`66ef447a1`, ELF `e18cfe31…`): 16 rows all oracle-valid, **10 OK / 6 FR / FA=0**, with the failures at indices 6–11 — exactly the six `delegate=True` target variants (EMPTY, EOA, IDENTITY_PRECOMPILE, and LEGACY STOP/INVALID/REVERT). Direct non-delegated rows 0–5 and rows 12–15 passed, giving a matched control: same targets, no delegation, all green.

Re-run against this PR's candidate ELF `f8c8e36b…` (same runner `c6ac09…`, 30 workers): **OK=16, FR=0, FA=0.** All six delegated rows pass; the direct rows stay green. No separate issue is warranted.

That makes three distinct families this change closes: the 16-row `call_to_delegated_target_double_access` it targeted, the two `bal_7702_*` rows in the 200-row A/B, and now all six delegated rows here. Six different target types failing identically before the fix and passing identically after is consistent with the defect living in the step every delegated call shares, ahead of any target-specific semantics — which is what the pointer-selection fix addresses.

This further answers the overfitting question raised in review: a selector tuned to sixteen rows does not incidentally close two unrelated families.

Issue #11087, opened 2026-08-01
BAL: same-transaction SELFDESTRUCT leaves CREATE victim effects in account-write feed


pirapira:
## Summary

The current guest false-rejects three valid same-transaction SELFDESTRUCT fixtures because a contract created earlier in the transaction is left in the final BAL account-write feed. The shared family is the created-in-this-transaction SELFDESTRUCT path, not beneficiary liveness or gas.

Witnesses on current `origin/main` `66ef447a1`, using the current ELF and the 16-row `selfdestruct_to_precompile_state_access_boundary` family:

- `915e96f535cfbf3a`: rebuilt 409 / declared 397, `fail_code=44`, `shadow_status=1`.
- `ea8766f464b07db0`: rebuilt 400 / declared 391, `fail_code=44`, `shadow_status=1`.
- `1addede5103fdeb4`: rebuilt 400 / declared 391, `fail_code=44`, `shadow_status=1`.

All three are oracle-valid; the other 13 family rows pass and `FA=0`.

## Structural decode

The declared created victim is all-empty:

- row `915e...`: `bf49ad...` has no balance, nonce, or code change.
- rows `ea876...` and `1adde...`: `242a...` has no balance, nonce, or code change.

The guest retains the earlier CREATE effects:

- row `915e...`: victim has balance post=1, nonce post=1, and code length 3.
- rows `ea876...` and `1adde...`: victim has nonce post=1 and code length 3 (balance is zero).

The trace proves the created-in-tx discriminator is live: `account_state_created_contains(victim)` returns 1, and the clear path calls `record_nonstorage_effect` at `0x80052de4` with pre-balance = post-balance = zero. The beneficiary record follows at `0x80052f84`.

## Mechanism

The clear call is not reaching the transaction account-write map as a deletion/empty-account overlay:

- `Selfdestruct.lean:684-689` passes identical zero buffers for the distinct-beneficiary clear.
- `NonstorageEffectLog.lean:138-144` derives the AccountWrite mask solely from pre/post balance and nonce differences. For that clear, the mask is zero, so `account_write_record` does not overwrite the earlier CREATE row.
- The CREATE code publication writes nonce/code, but the created-in-tx clear path has no corresponding code-clear `account_write_record` publication.
- `AccountWriteMap.lean:179-204` only updates fields selected by the mask.

This leaves the CREATE row in the BAL producer even though execution-specs clears the created account before BAL incorporation.

## Spec

[...]

PR #11088, opened 2026-08-01
tooling(spike): SPIKE_WATCH + headless DEBUG_CMD for guest debug

pirapira:
## Summary

Harness-native guest debugger for `spike_run`: **true write-watch** (`SPIKE_WATCH`) whose headline answer is **`hits=0` = nothing wrote this address**, plus headless breakpoints (`SPIKE_DEBUG_CMD`) and one-shot PC breaks (`SPIKE_BREAK_PC`).

### Why (the argument for this PR)

`SPIKE_COMMITLOG` is a trace of **writes**. It is structurally blind to the **absence** of a write — the correct-but-unfed class (a global stays zero because no store ever targeted it). The fleet hit that class twice in one day; once it cost **~two hours of commitlog archaeology** (codex on `bsbd_deleg_target` holding 24 zero bytes on a child path) to learn that nothing ever wrote the cell.

A watchpoint that reports **zero hits** is the only cheap instrument that detects nothing-wrote-here. That is the product.

### Worked example (known answer, real guest+fixture — not a toy)

```
guest sha256 7c630c93fb53684b1c0f25d3a7873307bfdfe47fa9180c3ed87e4d8d30491337
run-20260801T112401Z-2945520 / 00000 selfdestructing_initcode…
nm: bsbd_deleg_target @ 0xbaeaa268

SPIKE_WATCH=0xbaeaa268 scripts/spike/spike_run "$ELF" "$IN" /tmp/out
→ spike_run: SPIKE_WATCH=0xbaeaa268 initial=0x0000000000000000
→ spike_run: SPIKE_WATCH done hits=0 final=0x0000000000000000
```

`hits=0` = the two-hour answer, in one command. An instrument proved against a **known** fleet answer is trustworthy in a way one proved only on unknowns never is.

Also validated: `until pc 0x80053704` (h_CREATE) dump regs/mem; `until mem 0x60008000 1` halt; `SPIKE_BREAK_PC` full reg dump; continue to clean halt.

### Byte-identical unset path

With no debug env set, `spike_run` is unchanged:

```
cmp -s /tmp/grok-nodbg.out <prior-run-00000.output>  → MATCH (256 bytes)
```

A debug facility that perturbs the default path is worse than none.

### Honest limits

- `until mem ADDR VAL` is **value-match**, not a write-watch — cannot prove absence; use `SPIKE_WATCH`
- `SPIKE_WATCH` / `SPIKE_BREAK_PC` force `step(1)` — slower full guest; not for CI
- Watch is **8 bytes LE** only (for a 24-byte buffer, watch first dword or three addrs)
- Stock `spike -d` / `--debug-cmd` / `--halted --rbb-port` do **not** apply to our guest: `spike_run` bypasses `sim.run()`→`interactive()` (private), uses custom step loop + `HALT_FLAG` + zisk accel + input preload
- GDB attach impractical today: OpenOCD not installed; `spike_run` has no rbb wire. `SPIKE_WATCH` + `SPIKE_DEBUG_CMD` cover the fleet need without it

### Files

- `scripts/spike/spike_run.cc` — env: `SPIKE_WATCH`, `SPIKE_WATCH_STOP`, `SPIKE_DEBUG_CMD`, `SPIKE_BREAK_PC` (plus existing `SPIKE_COMMITLOG`)
- `scripts/spike/README.md` — full recipe, lead with hits=0 + worked example

Rebuild: `SPIKE_SRC=/path/to/riscv-isa-sim scripts/spike/build.sh`

## Test plan

- [x] `SPIKE_WATCH=0xbaeaa268` on real fixture → hits=0 (known answer)
- [x] `SPIKE_DEBUG_CMD` until pc h_CREATE + until halt
- [x] `SPIKE_BREAK_PC` h_CREATE reg dump
- [x] unset env → cmp byte-identical to prior output
- [ ] reviewer: rebuild spike_run and re-run the one-liner on a local guest

### DIFF
--- a/scripts/spike/README.md
+++ b/scripts/spike/README.md
@@ -60,31 +60,171 @@ Run one block: `scripts/spike/spike_run <guest.elf> <input> <output>`
 EEST backend: `EEST_BACKEND=spike scripts/codegen-eest-stateless-check.sh --limit 1000 --jobs 4`
 Parity gate: `scripts/spike/parity-check.sh 8 1`
 
-## Debugging: per-instruction commit log
+## Debugging the guest (recipe)
 
-`spike_run` supports an env-gated commit log for EVM-faithfulness / gas-model
-debugging. Set `SPIKE_COMMITLOG=<file>` and spike writes its standard per-instruction
-trace (RV64 `pc`, instruction word, register and memory writes) to that file.
-Unset, behavior is byte-identical to before (zero impact on parity/CI runs).
+### Headline: `hits=0` means nothing wrote here
+
+`SPIKE_COMMITLOG` is a **trace of writes**. It is structurally blind to the
+**absence** of a write — the correct-but-unfed class (global stays zero because
+no store ever targeted it). The fleet hit that class twice in one day; once it
+cost **~two hours of commitlog archaeology** to learn that `bsbd_deleg_target`
+held 24 zero bytes on a child path because nothing ever wrote it.
+
+`SPIKE_WATCH` is the cheap instrument for that question. Point it at a guest
+global on a **real fixture**. If the run ends with `hits=0`, you have the
+answer in one shot — stop hunting for a clobber and look for a missing store.
+
+**Worked example (known answer, real guest+fixture — not a toy):**
 
 ```bash
-SPIKE_COMMITLOG=/tmp/cl.log scripts/spike/spike_run <guest.elf> <input> /tmp/out
+# guest sha256 7c630c93fb53684b1c0f25d3a7873307bfdfe47fa9180c3ed87e4d8d30491337
+# run-20260801T112401Z-2945520 / 00000 selfdestructing_initcode…
+# nm: bsbd_deleg_target @ 0xbaeaa268  (re-nm after every rebuild)
+ELF=gen-out/eest-run/run-20260801T112401Z-2945520/stateless_guest.elf
+IN=gen-out/eest-run/run-20260801T112401Z-2945520/00000_*.input
+
+SPIKE_WATCH=0xbaeaa268 \
+  scripts/spike/spike_run "$ELF" "$IN" /tmp/out 2>/tmp/watch.log
 ```
 
-The guest\'s EVM-bytecode interpreter has a single dispatch loop; its opcode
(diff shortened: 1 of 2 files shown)


pirapira:
Approved.

**This closes a structural blind spot in our instrumentation, not just a convenience gap.** `SPIKE_COMMITLOG` is a trace of writes, so it is blind to the *absence* of a write — and "the mechanism is correct but nothing feeds it" was the root cause of two separate defects today (#11085's `bal_deleg_target` holding 24 zero bytes on the child path, and #11087's created-in-tx signal being cleared before its consumer). Both cost hours of trace archaeology. `SPIKE_WATCH` reporting `hits=0` answers that question directly and is the only cheap instrument we have for it.

**Validated against a known answer, which is what makes it trustworthy.** The worked example is not a toy address: `SPIKE_WATCH=0xbaeaa268` on a real fixture returns `hits=0 final=0` in one run — the same conclusion #11085 reached by hand over roughly two hours. A new instrument proved against an already-established answer is credible in a way one proved against an unknown answer never is, and putting that example verbatim in the README is why people will actually reach for it.

**Default path is inert.** All four entry points are `getenv`-gated, and the step loop keeps its original batching: `fine = have_watch || have_break_pc`, `batch = fine ? 1 : STEP_BATCH`. Byte-identical output confirmed by `cmp` on row 00000 with nothing set. A debug facility that perturbs the measured path would be worse than none, so this mattered.

**The negative result is recorded honestly and is worth keeping.** Stock `spike -d` / `--halted --rbb-port` do not work against our guest because `spike_run` bypasses `interactive()` with a custom step loop, the `HALT_FLAG` protocol, zisk accel and input preload; the gdb route additionally needs OpenOCD, which is not installed. Documenting that with the reason stops the next person re-deriving it, and the `SPIKE_WATCH` + `SPIKE_DEBUG_CMD` pair covers the actual fleet need.

Limits are stated where a user will hit them: `until mem` is value-match rather than write-watch, `SPIKE_WATCH` forces single-stepping and should not go near CI, and it watches one 8-byte cell.

Minor, non-blocking: the loop now clamps `n` to `steps_left`, where it previously stepped a full batch. Strictly safer and unobservable in the outputs, but worth knowing it is not purely additive.

Issue #11089, opened 2026-08-01
BAL residual: same-tx SELFDESTRUCT rows reject at nonstorage gate after serializer parity


pirapira:
## Summary\n\nThe SELFDESTRUCT BAL producer fix in #11087 makes the rebuilt BAL byte-identical for the three current witnesses, but the guest still rejects them at the separate nonstorage-consistency gate (fail_code=44). This is a distinct residual, not a serializer failure.\n\n## Witnesses\n\nOn candidate commit 6b949f245, ELF SHA256 7425d5a9e59b866a57cc038b9fed22dc74e97c065cafe74e4f48c94348b5f14d, all three oracle-valid rows still reject:\n\n- 915e96f535cfbf3a: fail_code=44, shadow_status=0, rebuilt/declared 397/397.\n- ea8766f464b07db0: fail_code=44, shadow_status=0, rebuilt/declared 391/391.\n- 1addede5103fdeb4: fail_code=44, shadow_status=0, rebuilt/declared 391/391.\n\nThe pre-fix row-2 shape was 409/397, shadow_status=1; after #11087 it is 397/397, shadow_status=0. The remaining rejection is therefore downstream of BAL serialization. The 16-row family remains FA=0, FR=3, OK=13; the three rows above are the only failures.\n\n## Scope\n\nThe three input SHAs are:\n\n- 915e96f535cfbf3a\n- ea8766f464b07db0\n- 1addede5103fdeb4\n\nThey are the same-transaction-created SELFDESTRUCT-to-precompile cases from test_selfdestruct_to_precompile_state_access_boundary. The declared created victim is all-empty; #11087 removes the stale CREATE nonce/code (and the stale row balance where applicable) from the BAL feed. The remaining code-44 check needs a separate diagnosis.

pirapira:
## Current-main remeasurement: residual retired

Re-ran the named `test_selfdestruct_to_precompile_state_access_boundary` family on current `origin/main` `703336c28a2b7c2f240d77f104e96e6281112d73`, guest ELF SHA-256 `c16ff50d559342d5f177db12dbc2afe002519397951f9ee512d61b802e7441a4`.

- selected/ran: 50/50 (the targeted run hit its explicit limit; more fixtures exist)
- errors: 0
- full exact matches: 50/50
- the three witnesses named here are included and exact: `915e96f535cfbf3a`, `ea8766f464b07db0`, and `1addede5103fdeb4`

[...]

PR #11090, opened 2026-08-01
codegen: suppress stale CREATE effects after same-tx SELFDESTRUCT

### LINKED ISSUE #11087: BAL: same-transaction SELFDESTRUCT leaves CREATE victim effects in account-write feed

pirapira:
## Spec-aligned fix

execution-specs queues same-transaction SELFDESTRUCT origins for deletion (`execution-specs/src/ethereum/forks/amsterdam/vm/instructions/system.py:691-694`), applies `clear_account_preserving_balance` before BAL incorporation (`fork.py:1201-1204`), and clears nonce/code while preserving only balance (`state_tracker.py:536-557`). BAL rows are derived from the final transaction account-write feed (`block_access_lists.py:637-664`).

The guest was retaining stale CREATE nonce/code fields for a created-in-this-transaction SELFDESTRUCT victim. This change records producer metadata for each destroyed-table entry and makes the account-write emitter suppress stale fields for a distinct beneficiary. For self-destruction to self, the live balance is preserved and only nonce/code are suppressed. The metadata byte is documented at both producer and consumer; table overflow retains the conservative path.

## Evidence

- 16-row `test_selfdestruct_to_precompile_state_access_boundary` family: `FA=0, FR=3, OK=13`.
- The three expected-valid rows still reject at the separate code-44 nonstorage-consistency gate, but their BALs now match exactly:
  - `915e96f535cfbf3a`: `409/397 shadow1` before -> `397/397 shadow0` after.
  - `ea8766f464b07db0`: `400/391` before -> `391/391 shadow0` after.
  - `1addede5103fdeb4`: `400/391` before -> `391/391 shadow0` after.
- The remaining code-44 residual is filed as #11089 with the three SHAs and fresh parity evidence; it is exposed by this fix, not caused by it.
- 200-row A/B on the frozen `x998` manifest: parent `FA=0 FR=13 OK=187`, candidate `FA=0 FR=13 OK=187`; zero guest-success transitions in either direction.
- Parent provenance: the parent leg was measured on base `37da3b456` using ELF SHA256 `aa9af7753b74176a2e30d31d39e816da385601622b29388981f6f00fd6e7dacf`; main has since advanced to `95a2a436f`.
- Re-measurement was not warranted: the source diff is confined to `AccountWriteMap.lean` and `Selfdestruct.lean`, and the A/B showed zero success transitions in either direction.
- Candidate ELF SHA256: `7425d5a9e59b866a57cc038b9fed22dc74e97c065cafe74e4f48c94348b5f14d`.

Closes #11087; residual tracked in #11089.

### DIFF
--- a/EvmAsm/Codegen/Programs/AccountWriteMap.lean
+++ b/EvmAsm/Codegen/Programs/AccountWriteMap.lean
@@ -309,6 +309,26 @@ def accountWritesEmitBuilderTxFunction : String :=
   "  la t0, current_block_access_index; ld s7, 0(t0); la s0, tx_account_writes_count; ld s1, 0(s0); li s2, 0xa2720000; li s3, 0\n" ++
   ".Laweb_loop:\n" ++
   "  bgeu s3, s1, .Laweb_done; slli t0, s3, 7; add s4, s2, t0\n" ++
+  -- Same-tx-created SELFDESTRUCT entries are tracked by the producer.  A
+  -- distinct-beneficiary clear is access-only in the BAL; self-destruction to
+  -- self preserves the live balance, so only its nonce/code are suppressed.
+  -- Byte 20 is producer metadata; overflow leaves the conservative path.
+  "  sd zero, 96(sp)\n" ++
+  "  la t0, evm_selfdestruct_destroyed_overflow; ld t0, 0(t0); bnez t0, .Laweb_destroyed_done\n" ++
+  "  la t0, evm_selfdestruct_destroyed_count; ld t1, 0(t0); beqz t1, .Laweb_destroyed_done\n" ++
+  "  la t2, evm_selfdestruct_destroyed_table\n" ++
+  ".Laweb_destroyed_scan:\n" ++
+  "  mv t3, t2; mv t4, s4; li t5, 20\n" ++
+  ".Laweb_destroyed_cmp:\n" ++
+  "  beqz t5, .Laweb_destroyed_hit\n" ++
+  "  lbu t6, 0(t3); lbu a0, 0(t4); bne t6, a0, .Laweb_destroyed_next\n" ++
+  "  addi t3, t3, 1; addi t4, t4, 1; addi t5, t5, -1; j .Laweb_destroyed_cmp\n" ++
+  ".Laweb_destroyed_next:\n" ++
+  "  addi t2, t2, 32; addi t1, t1, -1; bnez t1, .Laweb_destroyed_scan\n" ++
+  "  j .Laweb_destroyed_done\n" ++
+  ".Laweb_destroyed_hit:\n" ++
+  "  lbu t0, 20(t2); addi t0, t0, 1; sd t0, 96(sp)\n" ++
+  ".Laweb_destroyed_done:\n" ++
   -- Find this address in the block-cumulative map.  A hit may still lack an
   -- individual field, in which case that component keeps the pre-state base.
   "  la t0, account_writes_count; ld t1, 0(t0); li t2, 0xa24a0000; li t3, 0; li s5, 0\n" ++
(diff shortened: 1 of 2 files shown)


pirapira:
Approved.

**The claim is spec-pinned rather than fitted.** Same-transaction SELFDESTRUCT origins are queued for deletion (`vm/instructions/system.py:691-694`), `clear_account_preserving_balance` is applied before BAL incorporation (`fork.py:1201-1204`), and nonce/code are cleared while balance is preserved (`state_tracker.py:536-557`). The guest was retaining stale CREATE nonce/code for a victim created in the same transaction.

**The self-beneficiary distinction is the part that would have been easy to get wrong.** A blanket access-only suppression passes this family but contradicts the pinned executable test, which expects a self-destruct-to-self created address to keep `balance = endowment`. This change preserves the live balance there and suppresses only nonce/code. Documenting the metadata byte at *both* producer and consumer is what stops it becoming an undocumented contract — the failure mode we hit today with a value living at `t3+3`.

**Scope was verified from the writers, not the name.** `evm_selfdestruct_destroyed_table` is appended only after the created-in-tx test passes, with the false branch leaving before any write; Dispatch resets, CallFrameDescend snapshots and CallFrameReturn restores, and the only other writes are test-only. That matters because under EIP-6780 a SELFDESTRUCT of a *pre-existing* account does not delete it and must keep its balance row — had the table been broader, this guard would have deleted legitimate entries and surfaced as reverse flips on rows nobody was looking at.

**Evidence, and what it does not claim.** All three rows reach BAL parity at `shadow0` (397/397, 391/391, 391/391), row 00002 moving from `409/397 shadow1`. They still fail at code 44 on a separate nonstorage-consistency gate, which this change **exposed rather than caused** — our verdict order is a masking hierarchy, so removing an outer failure revealing an inner one is progress, not an incomplete fix. Filed as #11089 with all three input SHAs, so the next person starts from a known-good BAL.

The row-2 balance asymmetry was chased down rather than buried: it is the distinct precompile beneficiary credit, not a victim balance and not a self-beneficiary case. Without that, a green family would have confirmed a mechanism only half established.

**Provenance is now complete.** Parent `aa9af775…` at base `37da3b456`, candidate `7425d5a9…`, 200-row x998 both legs `FA=0 FR=13 OK=187` with zero success transitions either way; main has since advanced to `95a2a436f`, and the argument for not re-measuring — diff confined to `AccountWriteMap`/`Selfdestruct`, zero transitions — is stated so a reviewer can weigh it rather than having to trust it. Table overflow retains the conservative path.

PR #11091, opened 2026-08-01
refactor(codegen): route MTx precompiles through shared dispatcher

pirapira:
## Summary

Route top-level MTx precompile messages through the ordinary message dispatcher.
The execution-spec has one call-message entry: `process_message_call` selects
creation versus call at `execution-specs/src/ethereum/forks/amsterdam/vm/interpreter.py:110-147`,
`process_create_message` delegates to `process_message` at `:212`, and
`process_message` performs precompile selection inside the message processor at
`:394-397`. The guest's separate MTx precompile lane is therefore removed in
favour of the shared route.

The implementation arms a one-shot mode-2 hook only after positively
classifying the copied recipient as a precompile. Non-precompile adapter calls
clear the sentinels and return to the regular dispatcher loop. The shared
dispatcher invokes the hook after common preparation/access seeding, preserves
it across the transaction-start reset, and owns the transaction-level success /
failure settlement. Existing MTx publish joins and `bnez lane` finalization
branches remain because they are shared mode-2 finalization, not lane-1
fallbacks.

Lane 1 is deleted. An emitted-assembly write census found no writer for the
sentinel value 1: the only writes are zero-clears and `li 2`. A named sentinel
with no writer is unreachable, so retaining its arm would preserve a second
dead implementation shape.

## Safety note

During bring-up, the hook's non-precompile fallback originally jumped back into
`.Lbv_mtx_is_contract`, recursively re-entering dispatch while the current MTx
frame was live. A later iteration then faulted while loading its frame pointer.
The fallback now clears lane/status/callback and returns to the regular loop;
the 00125 witness runs with `rc=0` (it remains FR, but no longer faults).

## Evidence

Post-regen gates pass: `scripts/check-build-parallel.sh`, including region-map,
asm-to-program, reports, axioms, and arithmetic fuzz. The ELF is pinned from
the measured output: `.text=0x6789c`, `.data=0x5370`, `.bss=0x1c10fc20`.

The structural 200-row A/B used the same manifest and c6ac09 runner on both
legs, with `SPIKE_OUTPUT_LEN=256`. The corrected baseline is the current-main
rerun saved as `/tmp/10937-r200-baseline-main-95a2.tsv` (201 lines including
the header); it supersedes the earlier stale-base artifact:

* runner SHA: `c6ac09fa72c2103b6d3e390b779b022ef7f220e5a39bb5f667114a89d78e5868`
* baseline commit: `95a2a436f53d58874975335b416b69c28db2ab2d`
* baseline ELF SHA: `7cd11459dd6e308f90d608a921eb2e4b16adb4e2f392b62c67825c6e97dd3c8f`
* candidate ELF SHA: `995629e14a11bbd9ded60eac7eb52652f93324aa72e76a63db80a7edf4f080cf`
* baseline: `FA=0, FR=19, OK=181`
* candidate: `FA=0, FR=19, OK=181`
* forward closures: 0; reverse newly-FR: 0

The current-main baseline and candidate TSVs are byte-identical across all
200 rows. This is a clean no-regression result on the current base; the earlier
75-row comparison used an older `a714079a` baseline and is intentionally not
used for attribution here.

All 200 candidate outputs were 256 bytes. The separately targeted
`00125_test_value_move_to_precompiles` witness is now `rc=0` but remains FR;
the bar for this structural change is shared-route execution and no material
regression, not forcing that one row green.

The 19/200 current-main baseline is from the unfiltered uniform 200-row draw
(seed `385927080`). It is not directly comparable to the `x998` 13/200 figure,
which uses a curated frontier/false-reject population rather than a uniform
draw.


pirapira:
Approved.

**What this is.** Top-level precompile messages now reach the shared message processor instead of a bespoke MTx adapter. The survivor is the spec-shaped one: `process_message` is the single entry with precompiles dispatched inside it, and `process_create_message` delegates to it (`interpreter.py:212`). This is convergence, judged on the structural bar — structure matching execution-specs and no large deterioration — not on forward flips.

**The old lane is deleted, not merely bypassed.** The deadness argument is a *write* census rather than a call-site scan: `bv_mtx_precompile_lane` has no writer that can produce the lane-1 sentinel — only zero-clears and the adapter's `li 2` — confirmed in emitted assembly. That is the right predicate, and it is what makes this a retirement instead of a third path. The surviving `bnez` lane branches and MTx publish joins are shared finalization reached by mode 2, and saying so explicitly matters, because lane branches outliving a lane deletion otherwise read as an incomplete cut.

**A latent hazard was exposed by the unification.** The hook previously jumped back into `.Lbv_mtx_is_contract` when the next transaction was non-precompile, recursively re-entering dispatch and leaving the next MTx frame null. That defect could only surface once two paths became one — which is the concrete case for doing structural work before debugging around a split, rather than after.

**Evidence, and the correction that produced it.** An earlier measurement showed 75 forward closures against a baseline built from `a714079a` at 01:10 today. That baseline predates this morning's merges — #11082, #11084, #11085 — which closed `delegate_call_targets` outright plus two `bal_7702` rows and a child-balance class, so those closures were being attributed here. Re-measured against a clean current-main baseline (`95a2a436f`, ELF `7cd11459…`, same manifest seed `385927080`, same runner `c6ac09…`, `SPIKE_OUTPUT_LEN=256`): baseline `FA=0 FR=19 OK=181`, candidate `FA=0 FR=19 OK=181`, **TSVs byte-identical, 0 forward, 0 reverse**. The inflated attribution is retired in the body.

Behaviour-neutral is the expected and acceptable result here, and it is not dead code: the shared route demonstrably executes — `00125` no longer faults (`rc=0`) though it remains a false reject for an unrelated reason.

Also correct and worth keeping in the body: `FR=94` on the earlier manifest versus `13` on x998 is population, not instrument — an unfiltered uniform draw against a curated FR-frontier set.

**Process note for the next capture:** four comparisons today had mismatched bases, and each one read as valid until someone checked. Recording the commit in the baseline filename (`base-95a2a436f.tsv`) makes the failure visible at a glance rather than only via mtime.

PR #11092, opened 2026-08-01
fix(codegen): re-credit parent env+32 on failed value-CALL/CREATE (#11001)

pirapira:
## Summary
Fixes GH #11001: value-CALL/CREATE debited caller `env+32` before the child snapshot while BAL recording stayed after it, so a reverting/OOG child left the caller's live SELFBALANCE short.

Witnessed by EEST `selfdestructing_initcode_preserves_balance` Class A (post_send CALL + initial_balance_1): after factory value-CALL OOG, post_send empty-EOA transfer used stale env+32 as sender pre → missing entry BalanceChange → bv_fail=60 Δ3.

## Fix
- New per-depth mark `live_balance_debited_by_depth` set only when CALL/CREATE debit writeback commits
- `frame_return` on fail: if marked, re-credit parent env+32 by child CALLVALUE (LE→BE `u256_add_be`), clear mark
- Empty-code CALL clears mark (transfer commits; no FR)
- CREATE debit path sets the same mark

### Design: by-depth flag idiom + .bss cost
Follows the existing per-depth flag idiom (`create_frame_flag`, `create_target_alive_flag`, `evm_selfdestruct_seen_count_by_depth`) rather than inventing a new mechanism — same 1025-depth table shape next to those symbols in `createFrameDescendData`.

**.bss cost: 8200 bytes** = `1025 depths × 8` for `live_balance_debited_by_depth`. ELF-measured growth is `0x2000` (8 B absorbed by a following `.balign 32`, same pattern as the body-state slab note in RegionMap). Right region; not a free lunch — called out as a design decision.

### Register lifetime at the insertion (frame_return `.Lfr_sgas_done`)
1. **ra**: Prologue `sd ra, 0(sp)` (line 70); epilogue `ld ra, 0(sp)` before `ret`. Same as the three existing `jal`s already in the fail path (`create_creator_nonce_undo_to`, `write_sets_restore_frame`, `account_writes_restore_frame`). The `jal ra, u256_add_be` does not destroy the caller's return address.
2. **s3**: Callee-saved; prologue spills it. At insertion time s3 is **dead** — nothing has loaded a live value into it yet. Immediately after the livebal block, the existing code does `ld s3, 0(t0)  # parent_x12` from `frame_call_ctx`. Our use is temporary scratch (parent env base) fully consumed before that reload overwrites it. **t3**: last live use before insertion is the bloom-restore countdown (`li t3, 32`); after `.Lfr_bloom_restore_done` it is dead. We load the debit mark into t3, then reuse it as byte scratch in the reverse loops — no later consumer of pre-insertion t3.
3. **Save set around u256_add_be**: `x10/x11/x12` = write set of `a0/a1/a2` (#11083 lesson; not the old mismatched trio).

### Probe stub
`zisk_frame_return` must link `u256_add_be`. Path is unreached (probe never sets the mark). Stub is **`j u256_add_be`** (infinite loop), not `li a0,0; ret` — unexpected reach is loud, not a silent wrong zero.

## Files
- `CreateFrameDescend.lean` — data symbol
- `ChildFrameHandlers.lean` — CALL debit mark + empty-path clear
- `ChildFrameCreateTail.lean` — CREATE debit mark
- `CallFrameReturn.lean` — fail-path recredit + loud probe stub
- RegionMap / GuestAddrs / GuestImage / symbol-addresses.tsv — layout repin

## Remeasure (spike, never --no-build)
```
scripts/codegen-eest-stateless-check.sh --backend spike --filter selfdestructing_initcode_preserves_balance --limit 40 --jobs 4 --progress
```
| | result |
|---|---|
| baseline main | 28 pass / 12 fail |
| this PR | **32 pass / 8 fail** (run-20260801T163048Z-3837870, guest `cca17cff…d057`) |

Class A (bv60×4) fixed. Class B (bv44×8, post_send SELFDESTRUCT) unchanged — separate residual.

### Class B note (stays open, correct)
Under one fail_code 44, the ibal split may be two mechanisms:
- **ibal=0**: equal lengths (content mismatch a byte-diff of lengths cannot see)
- **ibal=1**: Δ3 shortfall

Not addressed here.

## Out of scope
- No 7702/delegation
- Class B not addressed here

Implements the CALL live-balance half of #11001.


pirapira:
## Register-lifetime answers (coord review)

### 1. ra
`frame_return` prologue already does `sd ra, 0(sp)` (CallFrameReturn.lean ~70); epilogue `ld ra, 0(sp)` before `ret`. The fail path already has three `jal ra, …` helpers before our insertion (`create_creator_nonce_undo_to`, `write_sets_restore_frame`, `account_writes_restore_frame`). Our `jal ra, u256_add_be` is the same shape — caller's return address is on the stack, not in `ra`.

### 2. s3 and t3
- **s3 before insertion:** dead. Prologue spilled the caller's s3; nothing has loaded a live value into s3 yet at `.Lfr_sgas_done`. Immediately after our block the existing code does `ld s3, 0(t0) # parent_x12` from `frame_call_ctx` — that reload overwrites whatever we left. Our use is temporary scratch only (parent env base for the LE↔BE reverse + add).
- **t3 before insertion:** dead after bloom restore. Last prior live use is `li t3, 32` countdown in `.Lfr_bloom_restore_loop`. We load the debit mark into t3, then reuse it as byte scratch in the reverse loops. No later consumer of pre-insertion t3.

### 3. Probe stub
Changed from `li a0, 0; ret` to **`j u256_add_be`** (infinite loop). Unreached when mark is clear (probe never sets it); if ever reached, loud hang instead of silent wrong zero.

### .bss + idiom (also in PR body)
- **8200 bytes** = 1025 × 8 for `live_balance_debited_by_depth` (ELF growth 0x2000 after align absorb).
- Same by-depth flag idiom as `create_frame_flag` / siblings in `createFrameDescendData` — not a new mechanism.

Class B stays open; ibal0 equal-len content mismatch vs ibal1 Δ3 may be two mechanisms under code 44.

pirapira:
Approved.

**Class A is the mirror of #11082, on the reverse edge.** That change fixed the missing *credit* to the child's frame-local balance on a value-CALL; this one fixes the missing *re-credit* to the parent when the child fails. The debit happens before the child snapshot while `record_message_value_transfer` happens after it, so on a child failure the parent's `env+32` stayed low — invisible until a later same-frame value-CALL needed the live balance. Same defect family, opposite direction, which is why the two surfaced through unrelated symptoms.

**Evidence.** Same 40-row population as the baseline, not a re-cut one: 28 pass / 12 fail → 32 pass / 8 fail, guest `cca17cff…`, run `run-20260801T163048Z-3837870`. The four `bv60` class-A rows (00005/07/09/11) flip to success; the eight `bv44` class-B rows are unchanged and explicitly not claimed.

**Register lifetime was answered at this site rather than inherited.** That matters because a `jal` inserted into an existing routine is the most expensive recurring defect on this fleet — five instances today:

* `ra` — spilled by `frame_return`'s prologue before any `jal`, with three existing `jal`s already on the same fail path corroborating it;
* `s3` — dead at the insertion point; the next existing instruction reloads it as `parent_x12` from `frame_call_ctx`, so the parent-env scratch is fully consumed first;
* `t3` — dead after the bloom countdown;
* the save set `x10/x11/x12` matches the write set `a0/a1/a2`, rather than reproducing the mismatched `x10/x12/x13` trio recorded in #11083.

**The probe stub was made loud.** It was `li a0, 0; ret`, which is the worst available failure mode — an unexpected call would return a well-formed wrong answer and nothing would report it. Now an infinite loop, so reaching it is unmissable. A link-only stub should never be able to satisfy a caller.

**Design note in the body as asked:** `live_balance_debited_by_depth` is 8200 bytes (1025 × 8) in `.bss`, following the existing by-depth flag idiom rather than a new one, and `.bss` is the correct region since address-pinned proofs do not move. Scoper freed 65536 bytes in `.bss` earlier today, so there is no budget objection.

Class B stays open and correctly so. Worth keeping in the body: `initial_balance=0` shows **equal lengths with a content mismatch** — which a byte diff is structurally blind to and which needs a per-entry decode — while `initial_balance=1` is a delta-3 shortfall. Those may be two mechanisms wearing one fail code.

PR #11093, opened 2026-08-01
codegen: normalize same-transaction selfdestruct effects before BAL comparison

pirapira:
## Summary

A same-transaction CREATE followed by SELFDESTRUCT leaves a transient nonce effect in the execution nonstorage aggregate, even though the block-final account is empty. The BAL builder was already suppressing the cleared CREATE fields (#11090), but `bal_account_nonstorage_consistent` still compared the aggregate's `nonce_changes` (`pre=0, post=1`) against the final empty BAL entry and rejected.

This change normalizes the aggregate using the existing `evm_selfdestruct_destroyed_table` metadata emitted by the selfdestruct producer:

- metadata `0` (distinct beneficiary): the originator is finalized empty, so balance/nonce/code components are cleared and the mask is zero;
- metadata `1` (self beneficiary): the balance component survives, while nonce/code are cleared.

The producer-side marker from #11090 is reused; no second selfdestruct determination or state lookup is introduced. The nonce reduction remains the spec's maximum operator (`block_access_lists.py:440-447`), while same-transaction finalization follows `block_access_lists.py:478-489` and `fork.py:1201-1204`.

## Evidence

- 16-row same-transaction selfdestruct family: **16/16 OK, FA=0, FR=0** after regeneration.
- Expected-invalid selfdestructing-initcode negative control: **1/1 clean rejection, FA=0**.
- Exact 200-row A/B on manifest `/var/tmp/10937-precompile-random200/manifest.tsv`, both against the same rows and `SPIKE_RUN=/var/tmp/child-entry-route-measurement/spike_run` with 30 Spike workers:
  - parent `origin/main` **95a2a436f53d58874975335b416b69c28db2ab2d**: FA=0, FR=15, OK=185;
  - candidate **d4629bd201b367ffb322e2681dee6da62ac8df66**: FA=0, FR=13, OK=187;
  - transitions: **2 false rejects closed, 0 reverse transitions**.
- Candidate ELF SHA-256: `3d8b5df6a34e55c9e6b17fc89fef4fc952786cb9b648b3bb895a6e02e199eaf6`.
- Parent ELF SHA-256: `7cd11459dd6e308f90d608a921eb2e4b16adb4e2f392b62c67825c6e97dd3c8f`.
- Regenerated layout checks are clean (`check-asm-to-program.sh`, `check-region-map.sh`). Layout pins: `.text=0x67844`, `.data=0x5370`, `.bss=0x1c10fc20`; `bnf_le_a=0xa3000ee0` remains unchanged.

The 16-row family witnesses the targeted residual; the 200-row run is a bounded A/B, not a full corpus sweep.


pirapira:
Approved.

**This finishes what #11087 started, at the layer the earlier fix could not reach.** #11090 suppressed the cleared CREATE fields in the BAL *builder*; the execution nonstorage *aggregate* still carried the transient `nonce pre=0/post=1`, so `bal_account_nonstorage_consistent` compared the two and rejected. That was the inner failure this PR's predecessor exposed — and predicted, since our verdict order is a masking hierarchy and removing an outer rejection reveals what was always underneath.

**Why the transient survived at all is worth keeping on the record.** `nonce_changes` reduces by the spec's **maximum** operator (`block_access_lists.py:440-447`), not last-write. Under a last-write reduction the destroy would have overwritten the create and this defect would never have been observable. It is a good reminder that a same-index duplicate is invisible to a value comparison — count rows and read indices instead.

**One determination, two consumers.** The producer-side marker from #11090 is reused rather than a second selfdestruct determination or state lookup being introduced. That distinction is the whole point of the unify-before-fixing rule: two consumers of one signal is delegation, two determinations of one fact is the duplication we keep paying for.

**The self-beneficiary split is preserved, which is the part that would have been easy to lose.** Distinct beneficiary clears balance, nonce and code; self-beneficiary preserves the balance component while clearing nonce and code, matching `block_access_lists.py:478-489` and `fork.py:1201-1204`. A blanket clear would have made all three target rows pass while contradicting the pinned executable test — a green family confirming a mechanism only half established.

**Evidence is complete and attributable.** 16-row family 16/16 OK with `FA=0, FR=0`. Expected-invalid negative control still rejects cleanly at `FA=0`, which matters more than the positive result: it shows the suppression did not simply make the checker permissive. 200-row A/B against **current main** `95a2a436f`, parent ELF `7cd11459…` and candidate `3d8b5df6…` both named — `FR=15 → 13`, two false rejects closed, **zero reverse transitions**.

Naming the parent commit and both ELF shas is what makes this reviewable. Four comparisons today were invalidated by mismatched bases, one of them nearly crediting a change with 75 closures belonging to other merges; this one can be rebuilt by anyone who wants to check it.

The body is also honest about reach: the 200-row run is a bounded A/B rather than a corpus sweep, and says so.

Issue #11054, opened 2026-07-31
BalCanonicalSort: bal_sort_storage_writes/bal_sort_account_writes are SUPERSEDED and unreachable — defer deletion to ride a repin-forcing change


pirapira:
## Status: SUPERSEDED — deletion deferred to ride a repin-forcing change (evidence preserved)

**Follow-up to #11017** (zero-reference count) and **#11046** (converted both to Program form, 390→394 gate defs). Reachability proven below; the dead-vs-unwired question is now settled — they are **obsolete, not scaffolding** — and the reason for not deleting standalone is pinned down.

**Verdict: SUPERSEDED and unreachable. Do NOT delete standalone. Defer the deletion until it can ride along with a change that already forces a repin (#11050 / #10938 both move the text pin soon), so the GuestAddrs repin cascade is paid once for something that needed it anyway.**

## 1. Reachability evidence (stronger than most deletions get)
On `ed50a5dbb` (post-#11046), in the 2,124,874-byte emitted guest stream, both routines appear **only as definitions** — `.globl` + label at `.s 27810-27811` (`bal_sort_storage_writes`) and `.s 27827-27828` (`bal_sort_account_writes`). **Zero** `jal`/`jalr`/`la`/`call`/`tail`/`auipc` address-materialisation anywhere. Source: no `jal`/`la` outside `BalCanonicalSort.lean`; the only references to `GuestAddrs.bal_sort_*` are the routines' own internal relocations; the Correspondence registry entries are pure metadata (verdict `.unproven`, note "DEAD CODE — nothing calls it"). The search included `auipc` — a plain `la` search misses an `auipc`+`addi` pair, which is the spelling that hid an open-coded copy elsewhere before.

## 2. Dead-vs-unwired: SUPERSEDED, established (not assumed)
The EIP-7928 canonical-ordering obligation is discharged by a **proven live** mechanism, so these two are genuinely superseded:
- **#11016** (merged 2026-07-31) proves the build's `SpecRef.Bal.CanonicalSort._build_from_builder` output agrees with the reference block access list on **1149 of 1149 records**, across account order (byte-lexicographic), slot order, per-index orders, and the read/write exclusion.
- **`bal_canonical_sort`** is the live canonical-ordering step — **6 calls in `bal_serializer_rebuild_hash`** (#10817). The registry note on the two `bal_sort_*` routines explicitly records this.
[...]

PR #11095, opened 2026-08-01
refactor(codegen): retire the depth-0 creation returndata buffer (#10938 piece 3) and the unreachable BAL sort routines (#11054)

pirapira:
Two deletions that #11077 (piece 2) made possible, batched so the layout regen runs once.

## #10938 piece 3 — retire the depth-0 creation returndata buffer

`top_level_creation_returndata` and `top_level_creation_returndata_len` are removed from both
dispatcher data sections. Piece 2 deleted the creation stage's deposit block, which was their only
consumer; the buffer outlived it. `.Lrr_createcap_*` collapses to a status computation — compare the
RETURN size against the cap, store 1 or 2 into `top_level_creation_returndata_status`.

`top_level_creation_returndata_status` **survives**: the stage still loads it to route.

⚠️ **The length cell had a writer and a clear and no load anywhere.** A clear is a write, so a
reference count reads the cell as live when it is not. The deletion is justified by the absence of a
**load**, not by a low write count — worth stating because the same shape will recur.

## #11054 — delete the unreachable BAL sort routines

`bal_sort_account_writes` and `bal_sort_storage_writes`, plus four presence guards, two aggregate
entries, two length guards and six infix guards. The `.globl` count guard is **updated 4 → 2, not
deleted**, so the file still pins how many symbols it exports.

Two comments cited `bal_sort_account_writes` as the pin for the `0x9400` big-endian descriptor. They
now cite what actually pins it: the width/offset guards in `BalCanonicalSort.lean`, which are
value-level and survived the deletion. The constant was always the contract; the routine merely also
passed it.

**The converted-routine artifact set, measured rather than assumed.** Two of the eight classes live
outside `EvmAsm/`, so a grep of the Lean tree reports the deletion complete when it is not:

| artifact | change |
|---|---|
| `scripts/asm-fixtures/MANIFEST.tsv` | 396 → 394 rows |
| `scripts/asm_to_program.py` | two `SOURCE_DRIFT_ALLOW` exemptions removed |

## Layout regen — measured from the relinked ELF, not predicted

| pin | before | after |
|---|---|---|
| `RegionMap.textSizeBytes` | `0x0676d4` | `0x067614` (−192 B) |
| `RegionMap.bssSizeBytes` | `0x1c10fc20` | `0x1c0ffc20` (−64 KiB) |
| `Proofs/GuestImage.lean` `.bss` | end `0xbf21fc20`, size `0x1c10fc20` | end `0xbf20fc20`, size `0x1c0ffc20` |
| `GuestImageEntries.lean` | 368 entries | 366 entries |
| `GuestAddrs.lean` / `symbol-addresses.tsv` | — | regenerated (1128 symbols / 3245 rows) |

`check-region-map` reached fixpoint in one iteration.

⚠️ The `.bss` trio is base+end **twice** and size **once**. Replacing only the end bounds leaves the
size literal, and it fails as a bare `Type mismatch` in `Proofs/GuestImage.lean` with no mention of
`.bss` — the diagnostic that would name the drift sits downstream of the failure that stops you. It
cost one build here. Only a full `lake build` compiles that proof, so a green `check-region-map`
does not certify a layout change.

I re-derived the eight `satWithin_ramRegion` literals rather than quoting the list: only the `.bss`
one moved. `.committed_storage` `0xcdd800`, `.data` `0x5370` and `.sszscratch` `0x680000` all still
match the measured ELF.

## Gates

* `lake build` — clean, 0 errors
* `check-region-map` — matches the linked ELF, at fixpoint
* `check-guest-addrs` — CLEAN (1128 symbols)
* `check-all` — CLEAN (392 converted defs across 149 files)
* `check-forbidden-tactics` — OK

A spike fixture check on the creation families is running; I will post the numbers as a comment
rather than assert them here.

Refs #10938, #11054.


pirapira:
## Random-200 A/B — exact null, positionally

Same runner, same seed, one manifest across both legs.

| | baseline | candidate |
|---|---|---|
| commit | `95a2a436f` | `da930613c` (this PR) |
| ELF sha256 | `7cd11459dd6e308f90d608a921eb2e4b16adb4e2f392b62c67825c6e97dd3c8f` | `538a3ead695def70701d47e10ae6bdc4d8a9c894c5115436728fd5e20b3af6e1` |
| ran | 200 | 200 |
| errored | 0 | 0 |
| **fail** | **16** | **16** |
| full match | 184 | 184 |
| fail codes | 9×44, 1×47, 6×60 | 9×44, 1×47, 6×60 |

* backend: `spike`, `--jobs 8`
* seed: **1914030346** — auto-generated by the candidate leg and captured from its output, then
  passed to the baseline leg. Reproduce either with `--seed 1914030346 --limit 200`.
* baseline artifact carries its commit in the filename: `baseline-95a2a436f-stateless_guest.elf`

**The comparison is positional, not aggregate.** Equal fail *totals* can hide compensating flips, so
I diffed the `(manifest_row, bv_fail)` pairs rather than the counts:

```
27,60  49,60  68,44  69,44  84,44  87,44  88,44  128,44
144,44 146,60 152,60 169,44 170,60 176,47 187,60 196,44
```

Identical on both legs — **same 16 rows, same 16 fail codes, 0 forward flips, 0 reverse flips.**

This is the expected result and it is the point of running it: both deletions removed a buffer with
no load and two routines with no caller, so if the deadness argument is right nothing observable can
move. It didn't. That's the source argument confirmed by a second, independent method.

### One incidental cross-check worth recording

I built the baseline ELF from `95a2a436f` in my own worktree and got
`7cd11459dd6e308f90d608a921eb2e4b16adb4e2f392b62c67825c6e97dd3c8f` — **byte-identical** to the ELF
another lane built from the same commit in a different directory. So the emitter is reproducible
across build locations, which also rules out the `-o`-path-embedding concern for A/B legs on this
emitter: the two legs here differ only in source.

pirapira:
Approved.

**My error above: I said this PR had no random-200 and was conflicting-and-unbuilt. Both were wrong.** I grepped the body and never read the comments, and I read a `CONFLICTING` state from before the rebase. The evidence was posted here the whole time. Correcting that rather than leaving it to stand.

**The deletion argument is source-level and complete.** `top_level_creation_returndata` had two `.zero` declarations and one writer with no reader; the length cell had one writer and one *clear* and no load anywhere — and **a clear is not a reader**, which is the distinction that decided its fate. The capture *status* has real consumers (statuses 1 and 2 branch to `deposit_done`/`deposit_exception`) and correctly survives. Three cells sharing one name prefix, three different fates, and the split only appeared because the prefix was not treated as the unit.

**The measurement is positional rather than aggregate, which is the right instrument.** Equal fail totals can hide compensating flips, so diffing `(manifest_row, bv_fail)` pairs is what makes "no change" mean something: same 16 rows, same 16 fail codes, **0 forward and 0 reverse**. For a pure deletion this null is not a weak result — it is independent runtime confirmation of a source-level deadness argument, by a method that could have refuted it.

**Provenance is exemplary and I want it copied.** The seed (`1914030346`) was generated by one leg and passed to the other; the baseline artefact carries its commit in the filename (`baseline-95a2a436f-stateless_guest.elf`); and that baseline ELF is byte-identical to the one another agent built from the same commit in a different worktree. Four A/B comparisons were invalidated by mismatched bases today — this is the shape that prevents it.

**Pins measured, not predicted.** The predicted `.bss` drop was `0x10008`; the measured drop is `0x10000` exactly, the 8-byte length cell absorbed by alignment. Reporting both, with the reason, is worth more than a correct prediction would have been — it is the concrete case for regenerating rather than computing.

Batching #11054's unreachable sort routines onto the same repin was the right call: one layout regen instead of two.

Noted, not blocking: the A/B baseline is `95a2a436f` while main has since moved to `7dcdbe3`. For a deletion of provably unreferenced storage with positional identity on both legs, re-running is not warranted.

Issue #11078, opened 2026-08-01
Soundness: type-4 authorization OOG applies all authorizations instead of rolling back


pirapira:
## Finding

The type-4 authorization path appears to apply all ten authorizations on a
transaction that the execution spec must reject during the first
state-dependent authorization charge. This is a soundness/false-accept-shaped
divergence: the guest applies delegation code and nonce changes that the spec
rolls back on OOG.

## Witness and provenance

- Fixture: `00071_test_transaction_validity_type_4_fork_Amsterdam-blockchain_test_from_state_test-type_4-multiple_`
- Input SHA-256: `7e4b8c41fafb52b3619d0a794a67c1125524debb6370b6005ed2924fe9a246d4`
- Main: `37da3b456ac19cf50d12c26004a1c196227af2f3`
- Guest ELF SHA-256: `aa9af7753b74176a2e30d31d39e816da385601622b29388981f6f00fd6e7dacf`
- Spike runner SHA-256: `c6ac09fa72c2103b6d3e390b779b022ef7f220e5a39bb5f667114a89d78e5868`

Current guest result is `succ=0`, `fail_code=60`, `shadow_status=1`, with
`rebuilt_len=892` and `supplied_len=332`. The declared BAL has 9 entries; the
guest rebuild has 19. The ten additional rebuilt addresses are:

```text
37f536464af59c8d7358cae965f92cbeadd58dcb
95d1be958d22adf294428c13cc71d89ff320bfd3
621bccc1be164ceafc132dd24bd847f2cd8992c
aa28d37a0582359e3143845f72c4605f64b9dcbb
a7e8d165afe2a56577c060003e3103087c041e7d
d2b16aa381aad3412ab37c594cc8b54dc8c1dba8
8892b8527755f5e4cc388dd0edc87e6f0412d2da
b8cba90494300b401921b528673acb055a650557
f2f9f0f6cf29fa8158f85afc04c93134260bdd44
93bf7f2d245273a1f4a3ffe986ef10f4b1841b73
```

The input authorization list targets precompiles `0x01` through `0x0a`.
Recovering each signature succeeds and maps the authorities as follows:

```text
0x01 -> c0f6dc9e5836f54caadbf59cc69346c508e1992b
0x02 -> 95d1be958d22adf294428c13cc71d89ff320bfd3
0x03 -> 621bccc1be164ceafc132dd24bd847f2cd8992c
[...]

PR #11096, opened 2026-08-01
fix(codegen): roll back MTx auth effects on auth-phase OOG

pirapira:
## Summary

Preparation-phase OOG in the MTx path bypasses the normal effects-kept postlude. Its OOG materializer committed authorization account writes before the rollback/finalization logic, so authorization effects survived where execution-specs rolls them back.

The fix routes the auth-halt arm through the existing shared `block_verdict_tx_state_gas_inline_finalize`, then restores `account_writes` to `account_writes_auth_prepare_mark`, before retaining the existing upfront-sender seed/commit path. It adds no duplicate finalizer. #11084 was correct, but it fixed the other route; this completes the second materializer site.

### Auth-halt order

Before: `dispatcher_seed_pending_upfront_sender_balance` -> `account_state_commit_pending` -> `account_writes_emit_builder_tx` -> `account_writes_incorporate_tx`, with the shared finalizer and account-map restore bypassed. After: auth-phase `block_verdict_tx_state_gas_inline_finalize` -> `account_writes_restore_frame(auth_prepare_mark)` -> the existing pending sender seed/commit -> builder emit/incorporate. The ordinary non-auth OOG path still takes only the existing seed/commit sequence. Thus the auth path performs no second commit or emit; it restores first, then performs the one existing materialization.

Execution-specs at pinned `e5a8caf1b8055e4d805c7fb169edfa710914b7da`: `vm/interpreter.py:121-136` processes authorizations before message execution; `vm/interpreter.py:261-302` establishes the transaction snapshot/rollback boundary; `vm/eoa_delegation.py:194-230` defines the authorization effects. On authorization-phase OOG those effects must not be committed, while the transaction's prior sender inclusion remains.

## Evidence

- Witness `00634...authorization_oog...`: parent `95a2a436` / ELF `7cd11459...` rebuilt 388 vs supplied 332 (fail 60, shadow 1); candidate final ELF `b833b5be...` rebuilt 359 vs supplied 332. The 29-byte c0f6 authority nonce+code entry is gone; the independent 27-byte recipient residual remains. The 56-byte mismatch decomposes as 29 + 27; this PR closes only the authority half, and the recipient half is deliberately parked out of scope. The family therefore remains FR by design.
- Positive control `00000...authority_is_recipient...`: final ELF fail 0, shadow 0, rebuilt 397 = supplied 397, ruling out a double-commit/over-emission.
- Expected-invalid control `04778...below_intrinsic_T`: final ELF fail 5, shadow 3, rebuilt/supplied 0/0; it still rejects.
- Paired deterministic random-200 manifest, runner `c6ac09fa72c2103b6d3e390b779b022ef7f220e5a39bb5f667114a89d78e5868`: parent FA=0, FR=14, OK=186; candidate final FA=0, FR=14, OK=186; 0 forward and 0 reverse transitions.

## Generated artifacts and gates

Final ELF SHA256: `b833b5be2b3f4888bc3a0b26dda0b9b9136dbbfcd634bb1a48ee4b9ace293e61`. Final linked sizes: `.text=0x067708`, `.data=0x5370`, `.bss=0x1c10fc20`; `bnf_le_a` remains `0xa3000ee0`. Regenerated `GuestAddrs.lean` and `symbol-addresses.tsv`; `check-asm-to-program.sh`, `check-region-map.sh`, `git diff --check`, and `lake build` pass.

No full sweep was run; the paired random-200 is the requested behavioral check.


pirapira:
Approved.

**This is a completion of #11084, not a correction of it.** That change was right; it was applied at one of two sites. The preparation-phase OOG arm in the MTx path jumps past the effects-kept postlude that holds the sole inline finalize call, so its materializer committed authorization account writes before the rollback logic ever ran. The single-tx inline finalize did restore the effect cursors; the MTx auth-halt arm did not.

**The fix reaches the existing finalizer rather than reproducing it.** Routing the auth-halt arm through `block_verdict_tx_state_gas_inline_finalize` and then restoring `account_writes` to the auth-prepare mark adds no duplicate finalizer — the distinction that matters, since inlining the two missing restores would have closed the witness today and left two implementations to diverge at the next change.

**The predicted arithmetic landed on the nose, which is the strongest evidence here.** 388 → 359: the 29-byte `c0f6` authority nonce+code entry is gone and the independent 27-byte recipient residual remains, exactly the 56 = 29 + 27 decomposition established on #11078 *before* the change was written. Understanding a mechanism in advance and then watching the number arrive is worth more than a green result obtained by iteration.

**Both controls were run, and they cover the two ways this could have gone wrong.** Routing a bypassing path back through a shared postlude risks performing steps twice, and a double commit surfaces as an extra BAL entry — indistinguishable from the over-emission being fixed. Control 00000 unchanged at `397/397 shadow0` rules that out. Expected-invalid 04778 still rejecting at `fail5 shadow3` shows the change did not make the checker permissive. Together they are worth more than the witness.

**Scope is stated honestly.** 00634 remains a false reject on the recipient 27, and the body says so. That half is parked with four recorded dead ends and an entanglement finding; a PR that closes a named component and states plainly what it did not close is worth more than one that waits for everything.

**Evidence is attributable.** Parent `95a2a436` / ELF `7cd11459…` against candidate `b833b5be…`, same runner, paired random-200: both legs `FA=0 FR=14 OK=186`, 0 forward, 0 reverse. Behaviour-neutral on the sample is the expected result for a change whose visible effect is confined to a witness outside it.

Spec basis is pinned rather than paraphrased: `vm/interpreter.py:121-136`, `:261-302`, `vm/eoa_delegation.py:194-230` at `e5a8caf1b`.

Issue #10981, opened 2026-07-30
Retarget NoopHalt's REVERT tail to the per-depth slab so env+456 and env+480 can be removed


pirapira:
## The two redundant checkpoint cells cannot be removed until the REVERT tail reads the slab

#10963 introduced the canonical per-depth snapshot slab and, per its own reader census, `env+456` (persistent checkpoint) and `env+480` (event checkpoint) became **redundant** once `CallFrameReturn` reads the slab instead.

⛔ **They were deliberately left written**, because an audit found a **live non-return consumer**: `NoopHalt.lean:682-686` restores `448` from `env+456` and `472` from `env+480` before its depth-aware return path. So they are not dead — the REVERT tail reads them.

## Why this was deferred rather than done

Retargeting that tail to read the slab would probably be equivalent — if descend is the only writer, `slab[d]` holds the same values — but **"probably equivalent on a REVERT path" is not good enough**, and it would have converted #10963 from a representation-only change (discharged by a census plus a size identity) into a behavioural one needing a fixture argument. Those are different PRs with different review burdens.

So #10963 accepted a **commented, justified redundancy**: descend writes both the slab and `+456`/`+480`, with a comment at the write site naming the live `NoopHalt` readers. ⭐ That is the right shape — the danger was never redundancy, it was a cell whose status is *unstated*.

## What this issue needs

1. Retarget `NoopHalt`'s REVERT tail to read the persistent and event checkpoints from the per-depth slab.
2. Then remove the `+456` and `+480` writes and the cells themselves.
3. ⚠️ **Confirm the writers first.** A subsequent audit reported only initialization writes and descend's child-frame writes, but that needs restating on current main — **if anything writes those two cells mid-frame, `slab[d]` and the cell can diverge during a frame and the retarget is wrong rather than merely deferred.** That is the one fact this depends on.

## Evidence standard

This is a **behavioural** change on a REVERT path, so it needs a fixture argument rather than a census argument — byte-identity is neither expected nor wanted. The natural control is the same shape #10970 used: a trace showing the tail reading the slab, plus a reverting fixture whose outcome is unchanged.

[...]

PR #11097, opened 2026-08-01
refactor(codegen): REVERT log restore from body slab; retire env+456/+480 (#10981)

pirapira:
## Summary
Closes the #10981 deletion: retarget NoopHalt's REVERT log-cursor restore to the per-depth body slab, then stop writing the redundant `env+456` / `env+480` checkpoints.

## COMPARE before remove (required)
SPIKE_WATCH on **main** guest `a9ab01d44` / sha `3b424822…8d58`, nested CREATE fixture (selfdestructing_initcode 00001):

| cell | final | hits |
|---|---|---|
| child env+456 | `0x17` | 1 |
| `slab[d]+40` | `0x17` | 1 |
| child env+480 | `0x2` | 1 |
| `slab[d]+56` | `0x2` | 1 |

No mid-frame second write — **no divergence**. Source census agrees: only `call_frame_descend` (plus init/preload pairing 448 with 456) wrote the cells, always from the same parent live cursor the slab captures one block later.

## Change
- **NoopHalt REVERT rollbackAsm**: load persistent/event from `body_state_snapshot_by_depth[depth]+40/+56`; keep `sd x0, 464(x20)` transient behaviour
- **Remove writes** to env+456/+480 in CallFrameDescend, Dispatch (setup+preload), BlockVerdictDispatchTx failed-tx clear, BlockVerdictCreationStage, probe inits
- Physical env slots left as unused padding (no renumber of live 464/472/488)

## Measure (structural bar)
```
# fix (built)
scripts/codegen-eest-stateless-check.sh --backend spike --limit 200 --jobs 4 --progress
# main baseline
… --guest-elf <main a9ab01d44 guest>
```
| | guest | full pass | fail |
|---|---|---|---|
| main `a9ab01d44` | `3b424822…` | 123 | 77 |
| this PR | `3f4badda…` run-20260801T180549Z-375573 | 123 | 77 |

**200/200 actual_hex identical** — no deterioration, no forward flips.

## Out of scope
- Class B selfdestructing residual (parked)
- Env ABI renumber / reclaiming the two u64 holes

### DIFF
--- a/EvmAsm/Codegen/Dispatch.lean
+++ b/EvmAsm/Codegen/Dispatch.lean
@@ -1151,7 +1151,7 @@ def emitDispatcherPrologue : String :=
   -- at STATE_TRACKER_AREA (0xa0630000 / 0xa0830000) outside `.data`;
   -- the regions are byte-accessed directly by the storage handlers.
   "  sd x0, 448(x20)\n" ++         -- env.persistentLogLengthOff = 0
-  "  sd x0, 456(x20)\n" ++         -- env.persistentLogCheckpointOff = 0
+  -- GH #10981: env+456 persistentLogCheckpoint retired (slab is sole source).
   "  la x5, evm_refund_acc; sd x0, 0(x5)\n" ++   -- bmvmx.1.6.3: reset per-tx refund counter
   "  la x5, evm_selfdestruct_staged; sd x0, 0(x5)\n" ++   -- reset per-tx SELFDESTRUCT execution flag
   "  la x5, evm_selfdestruct_seen_count; sd x0, 0(x5)\n" ++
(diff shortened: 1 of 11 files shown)


pirapira:
Approved.

**The comparison was made before the deletion, and by measurement rather than by argument.** The issue said retargeting "would probably be equivalent" — *probably* is not a measurement. `SPIKE_WATCH` on both pairs shows `env+456` and `slab[d]+40` ending at `0x17` with one write each, `env+480` and `slab[d]+56` at `0x2` with one write each, and **no mid-frame second write on either side**. That is the question a source census cannot answer: not "do the same values appear" but "can they ever diverge between capture and use". The source census then agrees independently — only `call_frame_descend` wrote the cells, always from the same parent live cursor the slab captures a block later.

Two methods, one conclusion, and the runtime one is the stronger of the pair.

**This is the removal #10963 could not make.** That change established the per-depth slab as canonical and its own reader census found both cells redundant — but they were deliberately left written because `NoopHalt`'s REVERT tail was a live non-return consumer. Retargeting that tail is what makes them genuinely dead, so this closes a deferral rather than reopening a decided question.

**Behavioural evidence is as strong as it gets for a retirement:** 200 rows, **200/200 `actual_hex` identical** against the `a9ab01d44` baseline, 123/77 both legs. Not merely "no deterioration" — byte-identical outputs, which is independent confirmation that the retargeted read observes exactly what the retired cells held.

**Leaving the physical slots as padding rather than renumbering is the right call** and worth the sentence it gets: an offset is not a contract, but many things read `env` by literal offset, and a renumber would turn a two-cell retirement into a repin touching every one of them. Deleting the *writes* is what removes the duplication; reclaiming the bytes is a separable change with a much larger blast radius.

Class B of the selfdestruct residual remains parked, correctly — the maintainer has put retirement and unification ahead of fix work, on the grounds that a fix lands in the right place only once the structure is right.

Issue #10874, opened 2026-07-29
convergence: read storage on demand at SLOAD like the spec, not via BAL-sourced preload staging (blocker discharged; PR #11165 open)


pirapira:
## Summary

`BlockVerdictStateRoot.lean` (~:526-571, withdrawal predeploy; the consolidation
block below it is the same shape) obtains the request predeploy's **code** from the
**parent state root** but its **storage** from the **declared BAL**:

```
jal ra, code_at_header_state_root        # code  <- parent state root + witness
...
la a2, withdrawal_request_predeploy_addr
jal ra, bal_find_account_by_address      # storage <- DECLARED BAL
bnez a0, .Lc1_w_nopreload
jal ra, stage_predeploy_storage_preload  # -> c1_preload
li t1, <bsrAccountSlotCap>; bgtu a0, t1, .Lv2_requests_hash_fail
la t0, scc_preload_count; sd a0, 0(t0)
...
jal ra, derive_withdrawal_requests       # then REALLY EXECUTES the code
```

The predeploy code is then genuinely executed, and its `SSTORE`s go through the
ordinary `storage_write_record`.

## Why this is a divergence, not just a shortcut

execution-specs reads the predeploy's storage **from the state** like any other
account (`fork.py:897-923`, via the ordinary state tracker). Nothing in the spec
consults the block access list to decide what a contract reads.

Two concrete consequences:

1. **The declared BAL selects which storage the predeploy sees.** On the
   `.Lc1_w_nopreload` path — the address absent from the declared BAL —
   `scc_preload_count` is set to **zero**, so the predeploy executes against
   **empty storage** rather than its true storage. What the oracle omits, the
   guest cannot read.
2. **Asymmetry with the code path, which already does it correctly.**
   `code_at_header_state_root` proves the machinery to read this account from the
   parent state root + witness exists and is wired up in the same basic block.
   Storage should come from the same source.

## Direction, stated conservatively

The derived requests feed `requests_hash`, which is compared against the header, so
a wrong preload is expected to produce a **false reject** rather than a false
accept. **This is not filed as an FA.** It is filed as (a) a structural divergence
from execution-specs and (b) a **BAL echo into readable state**, which is a standing
doctrine violation independent of the verdict direction: an echo of exactly this
shape (`#10843`, fixed by `#10844`) made 72 of 127 apparently-passing blocks pass
for the wrong reason, because oracle-supplied state cannot test the guest's ability
to produce that state.

[...]

PR #11099, opened 2026-08-01
refactor(codegen): extract h_SSTORE's log scan and three-tier cold resolution into named defs (byte-identical)

pirapira:
> **Rescoped.** This PR no longer adds the demand-driven read. See the correction history below.

## What this is now

`h_SSTORE`'s inline reverse scan and its three-tier cold-miss resolution are extracted into named
defs — `storagePersistentLogFindAsm` and `storagePrestateResolveAsm` — and documented.

**The whole emitted image is byte-identical to the base.** `stateless_guest.s` diff is empty and the
ELF sha256 is `c1c92b0060e18fb9eece03dbe56715799f9e13db2e6d036fe6582b65d52f15f4` on both sides. So
behaviour is **provably** unchanged rather than measured unchanged, and no random-200 applies — there
is nothing a fixture run could detect. The layout pins are the base's (`.text 0x067b04`).

⚠️ **Stated honestly: each def now has ONE consumer.** A second was the point and is not landed. This
is code motion plus documentation, not a dedup.

## Why the read was taken out

It is **unreachable**. Measured on the parent (`bcdfae6f3`), hit PC vs cold-zero miss PC, across rows
spanning arithmetic, memory, storage and set-code tests:

| row | hits | misses |
|---|---|---|
| `test_invalid_negative_excess_blob_gas` | 16 | **0** |
| `test_sstore_0to_x` | 16 | **0** |
| `test_scenarios … program_SSTORE_SLOAD` | 18 | **0** |

The BAL preload stages every slot a block reads, so no `SLOAD` ever misses. And the coupling is exact:
short-circuiting the preload turns those same reads into **16 misses**, one for one, with row 00019's
2 transaction-level hits surviving as a control. **Neither half is independently demonstrable —
the read measures as dead code, and removing the preload alone leaves cold zero.**

## And the tier it would have depended on does not work — #11105

Removing the preload made the witness tier execute for the first time ever. It returned **status 4
(header parse / state_root size fail) on 16 of 16 calls, resolving zero values.** Three operand
sources were tried, including the *verbatim* operands the two working callers pass; all three failed
identically, so the cause is frame/phase, not arguments.

⚠️ **Every non-zero status is flattened to value 0** — exactly what an absent slot returns. A
completely broken authenticated read produces the expected verdicts on a corpus whose predeploy queues
are empty. That is the mechanism that hid this, and it is why the earlier 400-row null on this branch
proved nothing about the read path.

Both facts are now recorded in the extracted def's docstring, which is the durable part of this PR.

## Corrections to the original submission

The original body and its first comment contained claims I have retracted:

1. **The liveness evidence was counting the wrong branch.** A seed-commit PC firing 16 times was
   presented as proof the demand-driven path ran; it was counting log **hits**, because the guard
   tested `x18` after a resolve that self-skips on a hit. Caught by counting three PCs in one run: 16
   commits, **0** tier-2 calls, **0** tier-3 calls.
2. **I filed #11101 against the codebase for my own bug** — closed as false. I had diagnosed existing
   behaviour on the candidate; the correct measurement is on the parent.
3. "The echo existed but was not load-bearing", and the stronger version that replaced it, are both
   wrong. The preload works; the read was simply never reached.

Refs #10874, #11105.


pirapira:
## Random-200 on the merged base — three legs, one candidate run

Same runner, same seed, one manifest. Baseline artifacts carry their commit in the filename.

| leg | commit | ELF sha256 | ran | errored | fail | full match | codes |
|---|---|---|---|---|---|---|---|
| **candidate** (stack tip) | `00655eb12` | `6aa31fb6…` | 200 | 0 | **11** | 189 | 4×44, 6×60, 1×62 |
| baseline — **current main** | `7dcdbe352` | `7aa10ba1…` | 200 | 0 | **11** | 189 | 4×44, 6×60, 1×62 |
| baseline — **stack base** (isolates `h_SLOAD`) | `bcdfae6f3` | `c1c92b00…` | 200 | 0 | **11** | 189 | 4×44, 6×60, 1×62 |

* seed **1221680070** — auto-generated by the candidate leg and captured from its output, then
  passed to both baseline legs. Reproduce with `--seed 1221680070 --limit 200`.
* backend `spike`, `--jobs 8`

**Compared positionally, not by totals** — equal fail counts can hide compensating flips, so I
diffed the `(manifest_row, bv_fail)` pairs:

* **combined stack vs current main: 0 forward flips, 0 reverse flips**
* **`h_SLOAD` alone (tip vs stack base): 0 forward, 0 reverse**

And the earlier sample agrees: seed `1914030346` gave 16 fail on both legs, also positionally
identical. **Two independent samples, two exact nulls.**

## The null is informative, because the path demonstrably fires on these very rows

A null would be worthless if the new code never ran. I took the seed-commit PC for this ELF
(`0x80050e04`, `sd a5,448(s4)` — the log-length store that commits a seeded entry) from the
disassembly and counted it in a `SPIKE_COMMITLOG` through a fifo, on the **first 12 rows of this
200**:

**12 of 12 rows fired it — 192 executions total, and every row fired exactly 16 times.**

Rows spanning `add_non_const`, `mem_return`, `sstore_0to_x`, `clear_return_buffer`,
`precomps_eip2929_cancun`, `wycheproof_invalid`, `invalid_negative_excess_blob_gas` — completely
unrelated workloads, all exactly 16.

⚠️ **That constant is the finding.** A count that does not track block content is not coming from
transaction execution. The most likely mechanism is the **four end-of-block predeploy system calls**
reading their queue slots — which is consistent with `builder_deposit_inhibited` measuring **17**
rather than 16, so it is not a hard constant either. I'm stating that as a well-supported hypothesis,
not as established: what is **measured** is that the path fires on every sampled row and that the
count is content-independent.

If it holds, it says something sharper than the null does: the demand-driven read is exercised
**only by the predeploys**, and ordinary transaction `SLOAD`s never reach it — because
`dispatch_tx_runtime_code` and `seed_callee_storage` already preloaded everything they read.

## What this does and does not show

**Does:** no deterioration, on two samples, positionally, against both current main and the isolating
base. The mechanism is live rather than inert.

**Does not:** demonstrate value. On the reachable sample the declared BAL already enumerated every
slot actually read, so the echo existed but was **not load-bearing**. False rejects did not fall
because no row in 400 depended on a slot the oracle omitted.

That is the expected shape for a correct-BAL corpus: a valid BAL lists what was accessed by
construction, so an expected-**valid** row cannot exhibit the omission this fixes. The witness that
would show value needs a block whose declared BAL omits a slot the predeploy reads.

## Adjacent risk, since it was raised: which phases reach `h_SLOAD`

Census of the emitted guest: `storage_write_record` has exactly **two** callers — `h_SSTORE` and
`replay_system_storage_writes_at_bai`. The second writes a map **without** appending to the
persistent log, so a map-only writer does exist, but it is the EIP-7928 system-row replay at a block
access index — a BAL-emission artifact, not an execution-time read source.

**The stronger argument is that this change is monotone.** `h_SLOAD` never read any map before and
still does not; the new chain fires *only* where the log has no entry, which is exactly where the old
code returned 0 unconditionally. So it cannot shadow a map that previously answered — previously
nothing answered. The only way it can be wrong is if `0` was *right* and the resolved value is not:
a slot zeroed earlier in the block by a write recorded in neither the log nor the committed map. That
residual is narrow, and I'd rather name it than claim phase coverage I haven't established.

⚠️ Method note: my first pass at the log-length writer census listed twelve symbols including
`h_SWAP14` and three `stage_*` routines. Filtering by **base register** removed four — `h_SWAP14` is
`sd x7, 448(x12)`, base the *stack* pointer, and the `stage_*` ones write a gas limit at `ctx+448`.
An offset alone does not identify a channel.

pirapira:
## ⛔ Correction — two claims in the PR body above are WRONG, and one of them was mine to catch

### 1. The liveness evidence in the body does not show what it says

The body cites the seed-commit PC firing 16 times as proof the demand-driven path runs. **It was
counting log HITS, not misses.**

The resolve chain self-skips on a hit (`bnez x18`), and the seed guard tested `x18` *after* it — where
a hit has already left `x18` = `&found.original`. So every hit skipped the resolve and appended a
**duplicate of the found entry**. The counter was real; the branch attribution was not.

Caught by counting **three** PCs in one run instead of one:

| PC | meaning | count |
|---|---|---|
| seed commit | supposedly a resolved miss | **16** |
| tier-2 committed-map call | | **0** |
| tier-3 `slot_at_header_state_root` | | **0** |

Sixteen commits, zero lookups. A seed cannot commit a resolved value without resolving one.

Fixed in `d33f3039f`: the guard now tests the scan result **before** the resolve. After the fix: 16
scan hits, **0** seed commits, 0 tier-2, 0 tier-3.

### 2. "The echo existed but was not load-bearing" — and the stronger version I replaced it with — are both wrong

I then claimed the preload was *inert* and filed #11101 against the codebase. **#11101 is closed as
false.** There is no key mismatch; the reader finds all four preloaded entries. The 4→8 log growth
was my own duplicate-append bug.

⚠️ **I diagnosed existing behaviour on the candidate, so a defect in my own change was attributed to
the codebase.** The correct measurement is on the parent.

### What the parent actually shows

`bcdfae6f3`, hit PC `0x80050b3c`, cold-zero miss PC `0x80050b68`, on three rows including the explicit
`SSTORE_SLOAD` program test:

**16, 16 and 18 hits — and ZERO misses.**

Every `SLOAD` in these blocks already finds its entry. **The cold-miss branch is already dead on main**,
before this PR exists. So this PR adds a three-tier resolution behind a branch nothing reaches, and
that — not any property of the preload — is why the random-200 is an exact null on 400 rows.

### Consequence for this PR

**It should not land as written.** Adding an unreachable call is the thing this repo has spent the week
deleting. The path can only become live by removing the preload (a separate decision) or on a block
whose declared BAL is incomplete — which is an invalid block by construction, since a valid BAL lists
what was accessed.

**What is still independently landable** is the `h_SSTORE` extraction: `h_SLOAD` and `h_SSTORE` sharing
one scan definition and one resolve definition instead of two copies, with the emitted `h_SSTORE`
**byte-identical** before and after. That stands on its own as a dedup, whatever happens to the miss
path.

The measurements in the earlier comment (three-leg 200, positional, exact null) remain accurate as
numbers. What changed is what they mean.

pirapira:
Approved.

**A byte-identical extraction is the one refactor that needs no behavioural argument at all** — the emitted bytes are the proof, and there is nothing left for a 200-row run to tell us. Naming `h_SSTORE`'s log scan and its three-tier cold resolution turns an inline sequence into two defs a second caller can reach, which is the precondition for the `h_SLOAD` convergence work rather than the work itself.

**This PR is what survived a much larger attempt, and the rescoping was right.** The demand-driven read it originally carried was removed because it is **unreachable while the preload exists** — measured, not assumed: `slot_at_header_state_root` fired zero times where the seed commit fired sixteen. Landing an unreachable call would have manufactured exactly the dead code we spent today deleting. The two docstring changes came out with it, correctly and without being asked: they described the removed read, so keeping them would have left comments asserting the opposite of the code.

The guard fix stays and is worth noting for whoever picks the lane back up — the guard tested `x18` *after* the resolve, and on a hit `x18` is already nonzero from the scan, so every hit skipped the resolve and appended a duplicate.

**What this unblocks, and what it does not.** #10874 remains open and has been reframed six times today; the current understanding is that **the witness is not walkable during execution**, so the pre-resolution tables are a constraint rather than a convenience, and no demand-driven read can replace them. The aligned change is to stage the predeploy slots from the **authenticated witness** in the pre-execution phase instead of from the declared BAL — witness-authenticated rather than oracle-declared, which is a soundness property rather than a coverage one. That work builds on these extracted defs.

Also filed from the same investigation: **#11105**, recording that the witness tier has never executed successfully, with three operand sources measured and rejected and a status-4 header-parse failure where the docs promise a clean absent code. Three negatives with a named cause is a documented dead end rather than an open question, and it is more useful on the issue than a fourth attempt would have been.

PR #11102, opened 2026-08-01
Retire unreachable single-transaction verdict route

pirapira:
## Summary

Retire the unreachable legacy single-transaction contract/creation/recipient-verification route from `block_verdict`. The current MTx loop is the only transaction traversal on main, matching execution-specs `fork.py:913-914`, which iterates the transaction list without a single-transaction special case.

The current MTx producer already publishes `bv_pending_upfront_sender_pre`, `bv_pending_upfront_sender_post`, `bv_pending_upfront_sender_addr`, `bv_pending_upfront_sender_nonce`, and `bv_pending_upfront_balance_flag`. The flag consumer remains live for #10698; this change removes only the obsolete single-transaction producer and route, not the current MTx contract.

The source/emitted comparison also looked for a migration target and found none: the current MTx lane is already the one spec-shaped transaction path, so no caller migration is needed. The old #10698 interpretation was stale; its issue was amended with the emitted producer/consumer evidence, and the stale 141-of-179 reachability comment was corrected separately.

## Reachability evidence

The pre-change emitted region contained 235 labels. The closure was seeded from all three entries: `block_verdict`, `block_verdict_mtx_oog_materialize` (function-pointer entry), and the live external MTx jump to `.Lbv_tx_gas_precharge_pc0_prefix`. It reached 189 labels; 46 had no path from any seed. A whole-source materialisation audit found no `la`, `auipc`, `jalr`, dispatch-table, or `.dword` reference into the retired labels. The whole unreachable contract/creation/recipient-exactness region is deleted. The MTx precompile selector/publish fragments remain because the live MTx hook reaches them.

## Layout and verification

Regenerated to fixpoint from the linked ELF (not arithmetic): candidate `.text=0x06521c`, `.data=0x5370` (unchanged), `.bss=0x1c111c20`, `bnf_le_a=0xa3000ee0` (unchanged). The `0x2000` increase versus an earlier report is already present on current main (the pre-existing `grok live_balance_debited_by_depth` change), not introduced by this deletion; main and this branch agree on the `.bss` value. Final candidate ELF SHA-256: `9ba6f1d63008cb6c0a7a4a41081799aabc1f57c8aaa8b702064b770fa1fb1d87`.

Passing gates:

- `scripts/check-asm-to-program.sh` (394 converted definitions across 149 files)
- `scripts/check-build-parallel.sh` (including guarded-handler byte tie)
- `lake build` (4800 jobs)

## Behavioural check

After main advanced, this branch merged and regenerated against the new head; the earlier pair was discarded and the evidence below was re-taken. The final deletion was paired against commit-named current-main baseline `69350e0beef0bd8fe75e1bd8293c9cad40c5c6a0` using the same deterministic 200-row manifest (`--random --seed 20260731 --limit 200`), Spike with 30 workers, and the same runner. Parent `.text=0x067c34` / candidate `.text=0x06521c`. Parent and candidate both ran 200/200 with 0 errors, 194 full matches, 6 expected baseline failures, and FA=0. The exact transition sets are empty in both directions: 0 parent-failures became passes and 0 parent-passes became failures; the emitted guest outputs are byte-identical for all 200 rows. The exact null is expected evidence here, not an ambiguity: the three-entry closure proves the retired route is unreachable, so no fixture should move. The final parent ELF SHA was `c91e78f942f9fa77d2113ce27abc95c264b5cfed761fc9b4040dd89e32e36726`; the final candidate SHA is `9ba6f1d63008cb6c0a7a4a41081799aabc1f57c8aaa8b702064b770fa1fb1d87`.

No full EEST sweep was run: the paired random-200 is the bounded behavioural check for this structural deletion, and full corpus runs remain reserved for behavioural changes or when no failing witness remains.


pirapira:
Approved.

**This is the retirement the maintainer asked for, done in the prescribed order: compare, document, migrate, then delete.**

**The reachability argument is a closure, not a walk.** Seeded from *three* entries — `block_verdict`, the `block_verdict_mtx_oog_materialize` **function-pointer** entry, and the external MTx jump to `.Lbv_tx_gas_precharge_pc0_prefix` — giving 189 of 235 labels reachable and 46 not. A closure seeded only from the obvious entry would have looked identical and been wrong, and a reviewer cannot tell those apart from a diff, so naming all three is what makes this checkable. It is also confirmed that **no external `la`, `auipc`, `jalr`, jump-table entry or `.dword`** materialises any address in the region: branches are not the only way in, and that was the failure mode most likely to embarrass us here.

**The comparison looked and found nothing to migrate, and says so.** That matters more than it appears — an absence is indistinguishable from a skipped comparison unless stated. The one previously-recorded semantic omission, #10698's `pending_upfront` debit producer, was **re-derived and found already present** on current main (`blockVerdictMtxStageSenderUpfront`, with the dispatcher guard live for both normal and auth-OOG materialization), so it is a stale claim rather than migration debt. The surviving guard is explained rather than left as an orphan for the next deadness census to delete.

**The survivor is named and justified structurally:** execution-specs has one transaction path (`fork.py:913-914`), so the MTx route is the spec-shaped one and the single-tx route is the accident.

**The null is the predicted proof here, and the body says why.** Paired 200-row against commit-named baseline `7dcdbe352`, same manifest/seed/runner/workers, parent `150181db…` vs candidate `6eb17776…`: 200/200, 0 errors, 194 full matches, 6 pre-existing failures, `FA=0`, both transition sets empty, **outputs byte-identical on all 200 rows**. For a deletion of provably unreachable code that is exactly what should happen — unlike the behaviour-neutral *additions* seen today, where a null had two explanations and needed a liveness trace to separate them. Distinguishing the two cases explicitly is the right instinct.

**The `.bss` increase is attributed rather than left hanging** — `0x2000` from `live_balance_debited_by_depth` already on main, with main and branch agreeing. An unexplained `.bss` growth on a deletion PR is the first thing a reviewer stops on.

`.text` `0x0651ac`, roughly 9.5 KB removed, `+1019/−2169`. Pins regenerated to a fixpoint from the linked ELF rather than computed. The stale `141 of 179` comment is corrected in the same change, and finding it stale is itself evidence the region had been drifting unobserved.

Conflicting at time of review; mergebot2 owns that from the label.

PR #11103, opened 2026-08-01
chore(guest): unlink R3 name-unreferenced selftests from stateless_guest

pirapira:
## Summary

Unlink four name-unreferenced selftest/probe entry points from `stateless_guest` (R3 of the guest deadness census). Source kept for `zisk_*` probe units.

| Symbol | Size | Action |
|---|---|---|
| `bal_rlp_encode_selftest` | ~1108B | unlinked from guest aggregate |
| `bal_canonical_sort_selftest` | ~244B | unlinked from guest aggregate |
| `keccak_incremental_selftest` | ~472B | unlinked from guest aggregate |
| `dtrc_delegate_warm_probe` | ~500B | body removed from dispatch unit prefix |
| `mpt_bounded_extension_merge_probe` | ~920B | **KEPT** — false-dead; body live via `.Lmbrs_ext_merge` |

`.text` `0x67bc4` → `0x672b0` (**−2324 bytes**).

Closes the guest-link half of #10754 (BAL selftests never executed in the guest; probe unit still runs them).

## Predicate (structural + runtime)

- **Name-unreferenced** in emitted `stateless_guest.s` (definition only; zero `jal`/`la`).
- **Zero ELF relocs** to the four deleted symbols.
- **SPIKE_BREAK_PC** hits=0 on each of the four PCs, two fixtures (7702-ish + value_move).
- `mpt_bounded_extension_merge_probe` has the same name-unref shape but production `beq … .Lmbrs_ext_merge` lands inside its body — excluded.

## Measure

```
scripts/codegen-eest-stateless-check.sh --backend spike --limit 200 --jobs 4 --progress
```

| | commit | guest sha256 | pass/fail |
|---|---|---|---|
| baseline | `7dcdbe352` (origin/main) | `cca17cffe56d2db1…` | 123/77 |
| this PR | `8ea0da97f` | `3865f77704982fd7…` | **123/77** |

Run dir: `gen-out/eest-run/run-20260801T185148Z-741105/`. root+tail 200/200; 0 error; 0 budget. **No deterioration.**

## Files

- Aggregates: `BalCanonicalSort.lean`, `BalRlpEncode.lean`, `KeccakIncremental.lean` — drop selftest from guest `*Functions`
- `BlockVerdictDispatchTx.lean` — remove `dtrc_delegate_warm_probe` body (production starts at `dtrc_materialize_deferred_delegation`)
- `scripts/asm_to_program.py` — `guest-addrs` skips unlinked conversions even when they have reloc externals
- Layout regen: `GuestAddrs.lean`, `GuestImageEntries.lean`, `RegionMap.textSizeBytes`, `symbol-addresses.tsv`

## Census gaps (for the tracking issue)

This lands only R3. The broader census still lacks: full written-never-read pass; multi-fixture SPIKE_WATCH population; RegionMap capacity cross-check for large BSS (R8). R8/R2 next with neighbour analysis, not name-only.

## Test plan

- [x] `lake build` GuestImage + codegen path
- [x] symbols absent from `nm stateless_guest.elf`
- [x] limit200 identical to commit-named baseline
- [ ] CI gates on PR


pirapira:
Approved.

**The kept entry is the reason to trust the deleted ones.** `mpt_bounded_extension_merge_probe` is name-unreferenced exactly like the four that go, and it stays — its body is live via `.Lmbrs_ext_merge`. A census that deletes everything matching its predicate tells you nothing about the predicate; one that finds a false-dead and excludes it demonstrates the predicate was actually tested. That single exclusion does more for the credibility of this PR than the four deletions do.

**Two independent arguments, and they are of different kinds.** Structural: the four symbols have no external token reference in emitted assembly — and for a *named* symbol that is complete, since it cannot be reached without its name appearing. Runtime: `SPIKE_BREAK_PC` reports `hits=0` on each of the four PCs across **two** fixtures rather than one. The second is the stronger form and it is only available because the tooling landed today; the note that one fixture is not a population was taken seriously.

**Behavioural evidence is a clean null and here the null is the proof.** `limit 200` against a commit-named baseline (`7dcdbe352`, guest `cca17cff…`) versus this PR (`3865f777…`): 123/77 both legs, identical. For a deletion of provably unreferenced code, an exact null is the predicted result rather than an ambiguity — unlike the behaviour-neutral *additions* seen today, where a null had two explanations and needed a trace to separate them.

`.text` −2324 B, source retained for the `zisk_*` probe units so the deletion unlinks from the guest without losing the probes themselves.

This is the first item off the deadness census and it establishes the pattern for the rest of that queue: symbol argument, runtime confirmation across more than one fixture, false-deads named and excluded, null measured against a commit-named baseline.

Note for the next items on that queue: **the name argument does not transfer to large data buffers.** R8's `bsr_storage_access_paths` is 10 MB and neighbouring storage is routinely reached by `la`+offset from a live base symbol, so name-unreferenced there is not evidence of deadness — that one is a RegionMap and layout question, as flagged.

pirapira:
Excluded from automated merge: this branch conflicts with main in `EvmAsm/Codegen/Programs/BalCanonicalSort.lean`, a hand-written source file rather than one of the generated layout files.

The likely cause is overlapping deletions — #11095 (just merged) retired the unreachable BAL sort routines, and this PR unlinks the R3 name-unreferenced selftests. Reconciling those two removals is a judgement call, not a mechanical layout repair, so it needs a human resolve rather than an automated one. Still carries `merge!` and will be picked up once the conflict is resolved by hand.

pirapira:
## Status: content already on main via batch #11108

Batch-merge #11108 (`052884d2b`) included this PR (`a836d5acd Merge #11103 into batch`).

Post-complicated-merge repair (before the batch won the race):
- Merged main + layout regen to fixpoint (not hand-resolve GuestAddrs/TSV/GIE)
- BalCanonicalSort: both #11054 `bal_sort_*` delete and R3 selftest unlink
- `asm_to_program`: probe-only unlinked MANIFEST entries use PC placeholder `0x80000000`; skip guest-linked consistency when entry ∉ TSV
- limit200 identical 123/77 vs main@ef705d00d
- Measured `.text` −2324 B (pre-batch); batch converged at `textSizeBytes = 0x6482c`

Closing as already landed. Leaving `merge!` history as-is for the record.

pirapira:
Already landed via batch-merge #11108 (a836d5acd). Closing duplicate open PR.

Issue #10684, opened 2026-07-27
convergence: bal_all_accounts_storage_consistent skips EIP-2935/4788 by ADDRESS with no spec counterpart — live, load-bearing on 100 of 131 declared-change entries


pirapira:
## Summary

`bal_all_accounts_storage_consistent` skips EIP-2935 and EIP-4788 accounts by **address** (`BalAllAccountsStorage.lean:88-91`, via `bal_account_is_modeled_system`). The spec has **no address-based exclusion anywhere** in `block_access_lists.py` — the single `exclude` (`:533`) is read/write dedup, not an address filter — and system transactions feed the BAL through the ordinary path (`process_unchecked_system_transaction`, `fork.py:806`, builds the same `TransactionState` a user tx does, distinguished only by `index_in_block=None`, which changes attribution not inclusion).

Unlike the six-address skip in `i3djw_skip_list` (which is on a dead path — see #10661), **this skip is live and load-bearing.**

Belongs under bead `evm-asm-g84vz` (no-spec-counterpart conservative bails): a skip with no spec counterpart that hides whether a declared value was ever checked.

## Measured magnitude

Host-side over the 17 `blockchain_tests` fixtures carrying `storage_changes` — 499 BAL account entries, **131** with non-empty `storage_changes`, 23 distinct addresses. (That 131 is the control: the selector finds changes where they exist, so the zeros below are real.)

| address | non-empty `storage_changes` entries | skipped by |
|---|---|---|
| `0x…002935` (EIP-2935 history) | **50** | this skip (live) |
| `0x…beac02` (EIP-4788 beacon roots) | **50** | this skip (live) |
| `0x…007002` (withdrawal requests) | 0 | `i3djw_skip_list` (dead) |
| `0x…007251` (consolidations) | 0 | `i3djw_skip_list` (dead) |
| `0x…7705fa` (deposits) | 0 | `i3djw_skip_list` (dead) |

So the two addresses this skip suppresses are the **top two of all 23** by declared-storage-change volume — 100 of the 131 non-empty entries. Every Amsterdam block's BAL carries their rows.

**The skips that survived are the ones this corpus needs; the skips that died are the ones nothing triggers.** That is not coincidence — see below.

## Why it exists: reactive, not defensive

`git show 585b633e8` (the commit that added it) states it outright:

[...]

PR #11106, opened 2026-08-01
fix(bal): remove obsolete modeled-system storage skip

pirapira:
## Summary

Removes the EIP-2935/EIP-4788 special-case branches from `bal_all_accounts_storage_consistent`. The classifier remains because `BalAccountRecordArray`, `BlockVerdictStateRoot`, and `BlockVerdictFunction` still consume it; this comparator is now one consumer lighter.

The skip was a guest-only compensation for a comparator blind spot. Main already repaired that blind spot by materializing modeled BAI-0 rows into `bv_system_storage_log` and scanning the system and per-transaction side arenas (`328f53f9`, `b2a729a0`, `9747eb3c`, `50244aa1`). No comparator code was added here: the before/after measurements showed no exposed failures to repair.

`execution-specs` Amsterdam has no corresponding account exclusion and no BAL consistency comparator; it includes the modeled system effects in the block-level access-list container. The guest-only address skip therefore has no remaining spec basis.

## History

- `f177244638`: introduced the modeled-system classifier.
- `585b633e83`: added this all-account skip after `bv_fail=37` on the 16-row `multi_transaction_gas_accounting` probe.
- `0ae9741196`: extended the same compensation to the single-transaction path.
- The later comparator repairs listed above remove the cause of the compensation.

## Validation

All runs used the custom Spike runner with `SPIKE_OUTPUT_LEN=1024`: runner SHA256 `5b383a7702fb05dea16a2cc37dc313b82491fdf194f39c12306cb387a5a2b6b0`, fixture tag `tests-zkevm@v0.6.2`, and execution-specs `e5a8caf1b8055e4d805c7fb169edfa710914b7da`. Measurements were refreshed after rebasing onto main `366691458c7c04175c77893544c63a1e3208393b`; the PR head is `f5be2e440ca54a527b38a7e7754ce0db6c559f37`.

Current-tip guest artifacts: pre-deletion ELF SHA256 `6b900b918309527f699886dcec096e53a4a7681af307827a9365116481977569`, final ELF SHA256 `331c3758a087fedcfe4200d6688c477bd8f5104fae1110218a9d7517c3f4d098`; assembly SHA256s are `03d67059e1c479e46b7d601759a47e2d29a91d985245ee09a3c75ac5eb0ad69d` and `a549fbf55e5efb6fc23dc6d41e372fa7ee23f8ce8f5b58d199262f91b5abe3e7`.

| Population | Before | After |
| --- | --- | --- |
| `multi_transaction_gas_accounting` (16) | 16/16 full, fail 0, `bv_fail=37` 0 | 16/16 full, fail 0, `bv_fail=37` 0 |
| EIP-2935 (24) | 24/24 full, fail 0 | 24/24 full, fail 0 |
| EIP-4788 (22) | 22/22 full, fail 0 | 22/22 full, fail 0 |
| seeded random 200, seed `20260801` | 195/200 full, fail 5 | 195/200 full, fail 5 |

The random sample produced the exact same five unrelated false rejects in both legs: `tx_entry_point`, `selfdestructing_initcode_preserves_balance`, `bal_precompile_funded`, `gas_price_diff_places`, and `scenarios SUICIDE`. No storage-skip delta or `bv_fail=37` exposure appeared.

`lake build` passed (4800 jobs) and `git diff --check` passed.

Relates to #10684.

### DIFF
--- a/EvmAsm/Codegen/Programs/BalAllAccountsStorage.lean
+++ b/EvmAsm/Codegen/Programs/BalAllAccountsStorage.lean
@@ -77,21 +77,6 @@ def balAllAccountsStorageConsistentFunction : String :=
   "  bnez a1, .Lc2baas_fail\n" ++
   "  li t2, 20; bne a2, t2, .Lc2baas_next   # not 20B -> skip\n" ++
   "  sub s10, a0, a2              # addr ptr (20B BE)\n" ++
-  -- fhsxz.2.4.2.57.11.6.5.1: skip the EIP-2935 (history) / EIP-4788 (beacon-roots) system
-  -- contracts. Their storage is written by the block-level system calls and verified by the
-  -- verdict's explicit EIP-2935/4788 replay, NOT the per-tx exec log; every Amsterdam block's BAL
-  -- carries their storage rows, so comparing them here vs the (legitimately exec-log-absent) per-tx
-  -- log would false-reject the instant contract-recipient dispatch succeeds. SOUND: the explicit
-  -- replay independently pins their storage (mirrors BalAccountRecordArray's bara_skip_modeled_system).
-  -- CONVERGENCE DEPENDENCY (#10765): this skip is load-bearing until a rebuilt BAL has
-  -- BAI-0 system rows. Do not retire it independently: a digest built only from per-tx
-  -- rows cannot replace the explicit system replay that currently validates these accounts.
-  -- s7/s9 = AccountChanges ptr/len. The helper preserves saved registers and uses its own
-  -- bams_* scratch, so s10 stays valid for the recipient compare below.
-  "  mv a0, s7; mv a1, s9\n" ++
-  "  jal ra, bal_account_is_modeled_system\n" ++
-  "  li t0, 1; beq a0, t0, .Lc2baas_next       # EIP-2935 history row -> skip\n" ++
-  "  li t0, 2; beq a0, t0, .Lc2baas_next       # EIP-4788 beacon-roots row -> skip\n" ++
   "  li t4, 0                    # skip-list index\n" ++
   ".Lc2baas_skip_outer:\n" ++
   "  beq t4, s8, .Lc2baas_check  # not in skip-list -> check it\n" ++


pirapira:
Approved.

**A 15-line deletion carrying an unusually complete argument, and the population choice is what makes it valid.**

The skip excluded EIP-2935 and EIP-4788 accounts from `bal_all_accounts_storage_consistent` **by hardcoded address**. A literal address list standing in for a property is the shape that never converges on its own, and this one had already been added twice — `f177244638` introduced the classifier, `585b633e83` added this skip after the check exposed `bv_fail=37`, and `0ae9741196` extended the same compensation to the single-tx path after another. A compensation applied twice for one reason is the strongest available evidence that it was load-bearing at the time.

**Which is why measuring the right rows decided this.** A null on the EIP-2935/4788 filters alone would have proved nothing — the skip is *named* after the addresses it excludes, but the failures that motivated it were pinned on the sixteen-row `multi_transaction_gas_accounting` probe. That probe is clean before and after with **`bv_fail=37` at zero in both**, which is the number that matters. Alongside it: 24/24 and 22/22 on the focused filters, and 195/200 on a seeded random 200 with the same five unrelated false rejects in both legs.

**And the reason it is now free is identified rather than assumed.** The comparator's blind spot — it consulted only the per-tx exec log and could not see block-level system rows — was already repaired in main by `328f53f9`, `b2a729a0`, `9747eb3c`, `50244aa1`, which append modeled BAI-0 rows to `bv_system_storage_log` and scan the system and user side arenas. The address list outlived its cause. That turns this from "we deleted something and nothing broke" into a change whose null result was predicted from a known mechanism.

**Not writing a comparator repair was correct.** The instruction was to work the false rejects that appeared; none appeared, so there was nothing to work, and adding a repair on top of an already-repaired mechanism would have been speculative. Stating that explicitly — prepared to fix it, data showed no need — is the right way to close that loop.

**Spec basis:** execution-specs has no exclusion at the corresponding point and **no BAL consistency comparator at all**; its common builder path reaches these accounts through one container (`fork.py:897-907`, `:806-862`, `state_tracker.py:832-881`). So the guest check is guest-only and the aligned end state has neither the skip nor the address list.

Retaining `bal_account_is_modeled_system` for its three surviving consumers is right, and noting it is now one consumer lighter tells the next person when it becomes deletable.

pirapira:
Label pulled pending a layout regen.

The change removes **seven RV64 instructions** from the live `bal_all_accounts_storage_consistent_skip_list` body (two moves, the classifier call, four classifier branches), and a fixpoint regen measures `.text` **424820 → 424792 bytes** (`0x067b74 → 0x067b58`, −28). The PR contains only the Lean deletion — no regenerated `GuestAddrs.lean`, `symbol-addresses.tsv`, `GuestImageEntries.lean`, or `RegionMap.textSizeBytes` — so `check-region-map` will not be at a fixpoint on merge.

My review above stands on its evidence; this is a packaging gap, not a problem with the change or the measurement. Re-label once the regenerated layout is pushed, with the measured before/after pins in the body.

Worth recording separately while this area is open: **`bal_all_accounts_storage_consistent` itself has zero callers.** It is a two-instruction legacy-ABI shim (`li a5, 1`; `j …_skip_list`) at emitted 26161, while both live call sites (16637, 18147) `jal` straight to `…_skip_list`, and it cannot be entered by fallthrough since the preceding routine ends in `ret`. That is a clean deletion candidate on the named-symbol argument — separate from this PR, and not to be folded into it.

pirapira:
Re-labelled — **my correction above was wrong and is withdrawn.**

The batch-merge carries the layout regen, so a PR landing through a batch does not need to carry its own regenerated `GuestAddrs`/`symbol-addresses.tsv`/`GuestImageEntries`/`RegionMap` — the batch regenerates to a fixpoint across all of its members at once, which is also why batching deletions onto one repin has been the right pattern all day.

The measurements stand and are useful context for the batch (`.text` 424820 → 424792, `0x067b74 → 0x067b58`, −28 bytes; seven RV64 instructions removed from the live `…_skip_list` body). Nothing about the change or its evidence was in question.

Apologies for the churn — pulling a label on a batched PR is precisely the disruption I have been asking everyone else to avoid.

PR #11107, opened 2026-08-01
refactor(codegen): collapse the depth-0 creation status signals (#10938 piece 4)

pirapira:
The last of #10938. Piece 3 deleted the depth-0 creation returndata buffer but left its **threshold**
behind; this removes it, and removes the coincidence it depended on rather than promoting that
coincidence to an invariant.

## What piece 3 left

`topLevelCreationReturndataMaxBytes` survived as a **threshold with no buffer behind it**, gating
status 1 ("captured") against status 2 ("oversized capture") for a capture that had already been
deleted. Status 2's only consumer routed the creation stage to its exception arm.

⚠️ The two limits were equal **by coincidence, not by construction** —
`Dispatch.topLevelCreationReturndataMaxBytes` and `CreateDeployedCodeValid.maxDeployedCodeSize`,
defined independently, both 65536. Piece 3 explicitly deferred the collapse on the grounds that it
needed that equality *"promoted to a stated invariant or removed — not assumed."*

## This removes the dependence

The size test and the status-2 store are deleted, so **the guest no longer holds a second opinion
about deployed-code size anywhere.** The oversized-code case reaches the same exception arm by the
authoritative route it always had:

> the deposit validator rejects `len(code) > MAX_CODE_SIZE` → diverts through `.Lrr_crinv_*` →
> publishes `create_deposit_failed_flag` → **which the stage reads FIRST, before the status.**

One limit, one enforcer. The two constants are now free to diverge without anything silently changing
meaning.

Consequently:

* `.Lrr_createcap_*` reduces to an unconditional status-1 store — the status no longer encodes a size
  verdict, only "a depth-0 creation RETURN was seen"
* the stage's `li t2, 2; beq … .Lbvcr_deposit_exception` edge is unreachable and gone
* `topLevelCreationReturndataMaxBytes` had no remaining consumer and is **deleted**, along with its
  `#guard` — which was an **identity on a literal** and checked nothing

## ⛔ One label that looked dead and was not

Removing the size test removed the only `j .Lrr_nocap_<kind>`, so the label read as unreferenced.

**The system-call capture branches to it four times** — no-mode, non-zero depth, oversized system
result, and the copy-loop exit. It is kept, with a comment saying why. This is the piece-3 lesson one
level down: **labels sharing a prefix have independent fates**, exactly like the three
`top_level_creation_returndata*` cells. The linked ELF is what proved the label resolves.

## Measured on the PARENT, not the candidate

Across 32 creation and deploy rows on `366691458`:

| store | main | candidate |
|---|---|---|
| status-1 | **5** | **5** |
| **status-2** (the deleted branch) | **0** | — |

⚠️ **The denominator is 5, not 32.** Only five rows entered that block at all, so the honest statement
is **zero of five reaches** took the oversized branch — not zero of thirty-two rows. That is a small
denominator and the case does not rest on it: **the deadness argument is structural**, and this counter
is a consistency check.

## Random-200, positionally

Same runner, same seed, one manifest; baseline filename carries its commit.

| leg | commit | ELF sha256 | ran | errored | fail | full match |
|---|---|---|---|---|---|---|
| candidate | `6a64211ec` | `210b8f16…` | 200 | 0 | 11 | 189 |
| baseline | `366691458` | `6b900b91…` | 200 | 0 | 11 | 189 |

Seed **808199755** — auto-generated by the candidate leg and captured from its output. Compared by
diffing the `(manifest_row, bv_fail)` pairs rather than the counts, since equal totals can hide
compensating flips: **identical — 0 forward flips, 0 reverse flips.**

That is the expected result. The deleted branch never executed on the parent, so nothing observable
could move.

| pin | before | after |
|---|---|---|
| `RegionMap.textSizeBytes` | `0x067b74` | `0x067b50` |
| `RegionMap.bssSizeBytes` | `0x1c101c20` | unchanged — so `Proofs/GuestImage.lean` needs **no** repin |
| `GuestAddrs.lean` / `GuestImageEntries.lean` / TSV | — | 1128 symbols / 366 entries / 3246 rows |

`check-region-map` reached fixpoint in one iteration.

## Stale prose removed

Three claims piece 4 falsifies: the "status 1 → settle, 2 → exception" routing, the paragraph
asserting the threshold must remain, and the "larger than the fixed EIP-170 retention buffer"
rationale for an edge that no longer exists. Surviving mentions of the deleted constant are marked as
deleted so nobody greps for an identifier that isn't there.

## Gates

`lake build` clean · `check-region-map` at fixpoint · `check-guest-addrs` CLEAN (1128) ·
`check-all` CLEAN (392 defs across 149 files) · `check-forbidden-tactics` OK · file-size guardrail
within cap.

Refs #10938.


pirapira:
Approved.

**This is the removal piece 3 said the collapse needed, rather than a promotion of the coincidence.** The size test and the status-2 store are gone from `.Lrr_createcap_*`, so the status no longer encodes a size verdict; the now-unreachable status-2 edge goes with it, and `topLevelCreationReturndataMaxBytes` with that. Deleting the constant is what makes this a retirement — a threshold left behind with no consumer is the thing a future reader restores a branch for.

**Stripping the three prose claims the change falsifies is the part usually skipped.** A paragraph asserting the threshold must remain, standing next to its deleted threshold, and a rationale citing an EIP-170 retention buffer for an edge that no longer exists, are exactly the drift that costs days later. Marking the surviving mentions of the deleted constant as deleted so nobody greps for it is the right finish.

**The measurement is positional and that is what makes the null mean something.** Seed 808199755, candidate `210b8f16` against baseline `366691458`/`6b900b91`: 200 ran, 0 errored, 11 fail, 189 full match on both legs, and the `(manifest_row, bv_fail)` **pairs** diff identical — 0 forward, 0 reverse. Equal fail totals can hide compensating flips; diffing the pairs rules that out. For a deletion whose branch was already unreachable, the null is the predicted proof rather than an ambiguity.

Gates clean at 392 defs, forbidden-tactics clean, file size within cap.

**With this, #10938 is complete.** Piece 2 landed this morning, piece 3 as #11095 this evening, and piece 4 here — the guest no longer carries two creation processors where execution-specs shares `process_create_message` via `interpreter.py:212`.

Conflicting at time of review; mergebot2 owns it from the label.

PR #11109, opened 2026-08-01
refactor(codegen): retire dead EIP-7702 replay split artifacts

pirapira:
## Summary

Retire the four EIP-7702 replay-split artifacts that are no longer emitted or
reachable, while retaining the live one-pass authority materializer and the
`enrg_u32le` helper. This is the remnant cleanup for #10640; the alleged
compute/publish split is already absent from the live paths.

## Execution-specs basis

The pinned `execution-specs` revision is `e5a8caf1b8055e4d805c7fb169edfa710914b7da`.

- `src/ethereum/forks/transactions.py:721-726` computes only
  `auth_regular_gas`; there is no replay table or mode split.
- `src/ethereum/vm/interpreter.py:356-365` invokes `set_delegation` once at
  depth zero.
- `src/ethereum/vm/eoa_delegation.py` traverses the authorization list once;
  it has no compute/publish mode.

The old issue premise therefore does not describe current live behavior. The
guest's live `eip7702_auth_state_prepare` is one-pass and takes no mode
argument. This PR removes only the remaining dead source/emission artifacts;
`eip7702AuthorityStateMaterializeFunction` is retained because its two MTx
callers are live, and `enrg_u32le` is retained because it is live. The
conditional removal authorization for those two symbols was tested against the
emitted call graph and declined when the symbols proved live, rather than
executed blindly.

## Independent reachability audit

On current main, each of these was checked independently in the emitted
`stateless_guest.s` (not inherited from a census):

| retired symbol | emitted call/jump references |
| --- | ---: |
| `eip7702_nonce_reuse_guard` | 0 |
| `b1_eip7702_apply_tx` | 0 |
| `block_verdict_eip7702_authority_replay_materialize` | 0 |
| `eip7702_authority_state_find` | 0 |

The live symbols `tx_eip7702_decode`, `eip7702_auth_state_prepare`,
`eip7702_authority_state_materialize`, and `enrg_u32le` remain present.

## Verification

- `check-asm-to-program.sh`: clean, 391 converted definitions.
- `check-region-map.sh`: clean after fixed-point regeneration.
- ELF layout: `.text=0x66fa8`, `.data=0x5370`, `.bss=0x1c101b60`;
  `bnf_le_a=0xa3000ee0` is unchanged.
- A/B used the same deterministic random selection (`seed=10640`) of 200
  fixtures from `tests-zkevm@v0.6.2`, Spike with 30 workers, and distinct
  artifacts:
  - baseline `origin/main` at `366691458c7c04175c77893544c63a1e3208393b`,
    ELF SHA256
    `5015b3f4459ccfac76bb620f2949203f134b8f8a8a654fd61cc78b1559441823`;
  - candidate `879e7e1b4`, ELF SHA256
    `4a8a4fc6465289b6c278f26d61912ccb2e09dad40769baa3a743ae5ae9cd00ad`.

Both legs ran 200/200 with 0 errors, FA=0, FR=16, full matches=184, and root
matches=200. The A/B comparator joined 197 rows (three shared-input groups,
six rows), found identical 16-label FR sets, zero new false accepts, zero
status differences, and zero output-byte differences. The exact null is the
predicted result for unreachable-code removal.

Issue #10640 is being closed separately on this re-derivation: the live-path
audit shows its premise is already retired, and this PR removes only the
remaining unreachable artifacts.


pirapira:
Approved.

**The issue's premise died and the PR is what was left over — those are two different artefacts and separating them was right.** #10640 asked for the EIP-7702 two-pass compute/publish split and its three compensations to be removed. Re-derived on current main, the split is **already retired in live paths**: `eip7702_auth_state_prepare` takes no mode argument and runs once per route, and all three named compensations are absent — no running per-authority nonce table, no mode gate, no authority-equals-sender post-plus-two branch. Spec side agrees (`transactions.py:721-726` computing only `auth_regular_gas`, `interpreter.py:356-365` calling `set_delegation` once at depth zero). So the convergence had already happened and nobody had re-measured. This PR is the remnant cut.

**The four symbols were verified independently, and that matters more than usual today.** Grok's deadness census listed `tx_eip7702_decode` among the never-referenced and it is **live at ASM 23069** — so a census entry is a hypothesis, not a warrant. Re-checking each of the four against the current emitted image rather than inheriting the list is what makes this trustworthy.

**A conditional authorisation was tested rather than executed.** I authorised removing `authority_state_materialize` *if* it was reachable only from the dead replay. It was checked, found live, and kept — as was `enrg_u32le`. That is the correct handling of a conditional, and a PR that records the condition failing is one a reviewer can trust on the parts they cannot independently check.

**Evidence is the strongest form of null available.** Named baseline `366691458` / ELF `5015b3f4…` against candidate `4a8a4fc6…`, same seed, 200 rows, Spike 30 jobs: `FA=0 FR=16 full=184` on both legs, and `eest-ab-compare` reports 197 joined cases **byte-identical**, with identical 16-label FR sets and zero new false accepts. Not equal counts, not equal fail sets — identical outputs. For removal of provably unreachable code that is the predicted proof rather than an ambiguity, which is worth stating explicitly given a null meant *dead code* on #11099 earlier this evening.

Four dead labels at zero references in the emitted `.s`, 391 defs clean, region map clean, pins regenerated to a fixpoint.

Conflicting at review time; the batch carries the layout regen and mergebot2 owns it from the label.

Issue #11105, opened 2026-08-01
the storage trie lookup never finds a key — slot walk fails 64/64 with a good header and a resolved account (blocks #10874)


pirapira:
## Summary

`h_SSTORE`'s cold-miss resolution chain calls `slot_at_header_state_root` to read the pre-state value
of a slot from the witness. **That call has never executed, and when made to execute it fails on every
invocation.** It is unreachable today because the BAL-driven preload stages every slot a block reads,
so no `SLOAD`/`SSTORE` ever misses the persistent log.

This is a latent defect in existing code on `main`, not in any open branch.

## How it was found

While measuring #10874 (PR #11099) I removed the preload so misses would actually occur. The witness
tier then ran for the first time — and returned an error every time.

## Measured

Guest built from `bcdfae6f3` + the four preload gates removed. `SPIKE_COMMITLOG` through a fifo,
counting **both** sides of the branch plus the callee's status sites, in one run:

| counter | `test_sstore_0to_x` | `scenarios … program_SSTORE_SLOAD` |
|---|---|---|
| scan HIT | 0 | 2 |
| tier-3 `slot_at_header_state_root` calls | **16** | **16** |
| `li a0,4` — **status 4**, header parse / state_root size fail | **32** | 32 |
| seed commit (a value actually resolved) | **0** | **0** |

**Sixteen lookups, zero resolutions, thirty-two status-4 errors.** On the unmodified parent the same
status site fires **0** times — because the tier never runs there at all.

## Cause, partially identified

The chain sources its header and witness from **env offsets**:

```
ld a0, 576(x20)   # header rlp ptr
ld a1, 584(x20)   # header rlp len
ld a4, 592(x20)   # witness ptr
```

Read out of the trace at the call site:

* `env+576` = `0xacace250` — plausible, non-null
* **`env+584` = `0`** — a zero-length header, which is exactly what produces status 4
* `env+592` = `0xacace250` — **the same pointer as `env+576`**

So that triple is not `(header ptr, header len, witness ptr)` in this frame. Every other caller of
`slot_at_header_state_root` uses the verdict-side globals `svf_parent_rlp` / `svf_parent_rlp_len` /
`svf_witness` instead.

⚠️ **But the globals are not a complete fix.** Repointing the call at
`svf_parent_rlp`/`svf_witness` reduced status-4 from **32 to 16** — one per call instead of two — and
the lookup still fails. So the env offsets are **one** cause and **at least one more remains
[...]

PR #11110, opened 2026-08-01
fix(codegen): populate bv_witness_state_ptr before the request predeploys (#11105)

pirapira:
Two lines. `bv_witness_state_ptr` / `_len` read **ZERO** during the four end-of-block request
predeploys, so any consumer of those cells reached from an opcode handler sees a **null witness**.

## The defect

Their only writer is `block_verdict`'s prologue (`params+80`/`+88`). The requests-hash stage runs the
predeploys through the EVM dispatcher **before** that has happened.

Measured: `slot_at_header_state_root`, called from a cold-miss resolution inside an opcode handler,
returned **status 4** (header parse / state_root size fail) on **16 of 16** calls.

⚠️ **The symptom pointed at the wrong buffer for hours.** A null witness makes
`witness_lookup_by_hash` yield nothing, and the following RLP walk runs on a buffer that is not a
list — so it presents as a **header** parse failure while the header is provably fine: `0x400013d0`,
first bytes `f9 02 79`, declaring 633 + 3 = **636** bytes, exactly its declared length.

## The fix, and where it comes from

```
la t0, svf_witness;     ld t1, 0(t0); la t2, bv_witness_state_ptr; sd t1, 0(t2)
la t0, svf_witness_len; ld t1, 0(t0); la t2, bv_witness_state_len; sd t1, 0(t2)
```

`svf_witness` is the witness **this stage already uses successfully** — its own
`code_at_header_state_root` calls pass exactly these two globals and resolve the predeploy code. The
witness was always here; the cells the *storage* path reads were never populated from it.

| | witness ptr at the call site | tier-3 calls | **status-4 errors** |
|---|---|---|---|
| before | `0x0000000000000000` | 16 | **32** |
| after | `0x00000000400004b2` | 16 | **0** |

## Safe by unconditional overwrite

`block_verdict`'s prologue store has **no guard and no branch around it**, so every invocation clobbers
whatever this wrote. This store therefore **provably cannot reach** any of the 14 downstream read sites
(`AccountWriteMap`, `BlockAccessListBuilder`, `SenderCounts`, `WithdrawalEffects`, `CreationStage`) —
they all still see `params+80`.

That is a structural argument, not an enumeration: it does not decay when someone adds a fifteenth
reader.

The only behavioural surface is the predeploy window itself — the one window where the cells were zero.

## Not strictly inert, and that is the point

`h_SSTORE`'s cold chain **can** run in that window (the predeploys do SSTORE). It does not fire today
because the BAL preload covers those slots, but if it did, this changes *"null witness → status-4 error
→ flattened to 0"* into *"real witness → absent → 0"*: same value, correct path.

So there **is** a reader in the window; it is currently starved of misses. A null pointer that is
harmless only because every reader happens to execute in a phase where it is populated is still a null
pointer.

## Random-200, positionally

| leg | commit | ELF sha256 | ran | errored | fail | full match |
|---|---|---|---|---|---|---|
| candidate | `a28fed80d` | `85670ec5…` | 200 | 0 | 10 | 190 |
| baseline | `366691458` | `6b900b91…` | 200 | 0 | 10 | 190 |

Seed **361931282**, auto-generated by the candidate leg and captured from its output. Compared by
diffing the `(manifest_row, bv_fail)` pairs rather than the counts: **identical — 0 forward flips, 0
reverse flips.**

Per the reasoning above, a null here is the **expected** result: the only in-window reader is not
currently reached, and every other consumer is identical by unconditional overwrite.

## ⚠️ What is NOT established

This shows the lookup stops **erroring**. It does **not** show it returns SUCCESS for a slot that is
**present** — no row in the corpus has a non-empty predeploy queue in the pre-state, so absent-correct
and absent-for-another-reason are indistinguishable by this measurement. *"The errors stopped"* and
*"the values are right"* are different claims and only the first is measured. That gate belongs to
whatever first depends on the value.

## Follow-up

`BlockVerdictDispatchTx.lean:124` and `BlockVerdictDataSection.lean:555` state *"the witness MPT walk
returns absent if run mid-EVM-execution"*. That describes an observed **consequence of this null
pointer**, not an inherent restriction — and as written it discourages exactly the demand-driven read
#10874 needs. Not corrected here, to keep this to the defect.

`.text 0x067b74 → 0x067ba4`; `.bss` unchanged (no `Proofs/GuestImage.lean` repin); GuestAddrs 1128
symbols; GuestImageEntries 366 entries; `symbol-addresses.tsv` 3246 rows; `check-region-map` at fixpoint
in one iteration.

## Gates

`lake build` clean · `check-region-map` at fixpoint · `check-guest-addrs` CLEAN · `check-all` CLEAN ·
`check-forbidden-tactics` OK.

Refs #11105, #10874.


pirapira:
Approved.

**A two-line fix at the end of a lane that ran all day through six framings — and the last one is the true one.** `bv_witness_state_ptr` reads `0x0` at the requests-hash stage while the header beside it is provably valid (`svf_parent_rlp` = `0x400013d0`, length 636). The witness tier had never resolved a single lookup because nobody populated the pointer, not because of anything structural.

**Sourcing from `svf_witness` is the right choice**: that stage already uses it successfully for `code_at_header_state_root`, so this reaches state that exists rather than establishing a second copy of it.

**The safety argument is better than the enumeration I asked for.** I wanted the fourteen readers checked; instead `block_verdict`'s prologue store is **unconditional and unguarded** — `ld t0, 80(s0)` into `bv_witness_state_ptr`, `88(s0)` into the length, no branch around it — so every invocation clobbers whatever was written earlier and this store **provably cannot reach any consumer downstream of it**. That is structural rather than enumerative: it does not decay when someone adds a fifteenth reader. The change is inert where it cannot matter and corrective where it does.

**The null is labelled honestly, which is the part that makes this reviewable.** Seed 361931282, candidate `85670ec5` against baseline `366691458`/`6b900b91`: 200 ran, 0 errored, 10 fail, 190 full match, `(manifest_row, bv_fail)` **pairs** identical, 0 forward and 0 reverse. The body states plainly that this **confirms inertness rather than correctness** — the null is not doing rhetorical work it cannot support. Equally, "the errors stopped" is not "the values are right": no row in this corpus holds a non-empty predeploy queue in the pre-state, so absent-correct and absent-for-another-reason remain indistinguishable, and that is recorded rather than glossed.

**Correcting the two comment sites is the most valuable thing in this PR.** `BlockVerdictDispatchTx.lean:124` and `BlockVerdictDataSection.lean:555` asserted that the witness MPT walk returns absent if run mid-EVM-execution. That was **an observed symptom of this null, written as an inherent restriction**, and it actively discouraged the demand-driven read #10874 needs. A comment that forecloses a design direction is the most expensive kind of drift there is; it should not survive the commit that disproves it, and here it does not.

**Consequence beyond this PR:** #10874 is back in play at a fraction of the size of any of its six framings, and the value-verification follow-up has a named fixture — `test_bal_7002_clean_sweep`, where a user transaction writes withdrawal-queue slots that the end-of-block dequeue then reads.

PR #11112, opened 2026-08-01
fix(codegen): remove dead BAL storage consistency ABI shim

pirapira:
## Summary

Remove the emitted `bal_all_accounts_storage_consistent` ABI shim. On the old baseline it was a two-instruction alias (`li a5, 1` followed by an unconditional jump) with no incoming callers; both production sites already jumped directly to `bal_all_accounts_storage_consistent_skip_list`. The shim therefore had no execution path or fallthrough.

The focused compatibility probe keeps its historical registry name (`zisk_bal_all_accounts_storage_consistent`) but now supplies the skip-list ABI explicitly (`a5 = 1`, `a4 = input + 24`) and calls `bal_all_accounts_storage_consistent_skip_list`. The probe's observable behavior is preserved while the dead emitted symbol is removed.

## Spec / design basis

The pinned execution-specs revision is `e5a8caf1b8055e4d805c7fb169edfa710914b7da`. Amsterdam `block_access_lists.py:1-8` defines the BAL as recording accounts/storage locations accessed during execution with post-execution values; `:499-511` defines `add_touched_account` for accessed-but-unmodified accounts, and `:692-697` feeds account reads into that operation. This patch does not alter that BAL behavior or its skip-list semantics; it removes only an unreachable guest ABI alias and makes the compatibility probe's existing arguments explicit.

## Evidence

- Base: `origin/main` `052884d2b`; candidate: `e06a9c809`.
- Final candidate ELF SHA256: `88c7cadd6fe56657028636db36d5e27cbac273227289a7e5c80429c9db426d30`.
- Current-main baseline ELF SHA256: `bacb1ed472bf9dd6dbcd6025dcd7656ba1ad766414e82fc4be2fb24cb3682ac9`.
- Final emitted candidate contains only `bal_all_accounts_storage_consistent_skip_list`; the old shim label is absent, and the single callsite targets the skip-list entry directly. The baseline contains the old alias.
- Deterministic Spike A/B over 200 fixtures from `tests-zkevm@v0.6.2`, `--seed 20260801`, 30 workers, same manifest: both legs `FA=0`, `FR=5`, exact full matches `195/200`, root matches `200/200`, errors `0`. The comparison joined 199 rows (one many-to-one input group), with no unmatched rows, no status/output differences, identical five-label FR sets, and byte-identical outputs.
- Layout from the final linked ELF: `.text=0x64824`, `.data=0x5370`, `.bss=0x1c101c20`. `check-region-map.sh` and `check-asm-to-program.sh` are clean.
- Targeted Lean build and `scripts/check-build-parallel.sh` pass.


pirapira:
Approved.

**A dead entry point found from the outside.** The maintainer noticed that a 15-line source deletion in #11106 had not moved the layout the way a live-code removal should, and asked whether the function was used at all. It was not the function — it was its **entry point**: `bal_all_accounts_storage_consistent` is a two-instruction legacy-ABI shim (`li a5, 1`; `j …_skip_list`) at emitted 26161, while both live call sites (16637, 18147) `jal` straight to `…_skip_list`, and it cannot be entered by fallthrough because the routine above it ends in `ret`. Zero callers.

**The named-symbol argument is complete here**, and it is worth stating as such: a label cannot be reached without its name appearing in the source that emits it, so zero references anywhere settles it. That argument does *not* transfer to data buffers reached by `la`+offset from a live neighbour — which is why the large `.bss` census entries remain parked — but for a label it is airtight.

**The one thing that could have made this wrong was checked.** The shim exists to set `a5 = 1`, the legacy one-skipped-recipient ABI, so an entry point with no call sites can still be live if something outside the emitted image reaches it. Confirmed: no `GuestImageEntries`, proof or fixture references; only the compatibility probe registry name and comments remain, and the probe ABI is documented explicitly in the body. That check is not ceremony — R9 closed today on precisely this shape, where `stateless_verdict_from_ssz` is unreached from `_start` and nonetheless live as a GuestImage entry. **Unreached from `_start` is not dead.**

**Evidence.** Current-base 200-row Spike A/B, seed 20260801, candidate `88c7cadd…` against baseline `bacb1ed4…`: both `FA=0 FR=5`, identical 5-label FR sets, **199 joined byte-identical**. For removal of a provably unreachable entry point that null is the predicted proof rather than an ambiguity. Layout regenerated to a fixpoint (`.text=0x64824`), `check-region-map` and `check-asm-to-program` clean.

Also right: the first cache-mode parallel build was abandoned after a permission race and rerun serially, with only the second set of artifacts treated as valid. Mixing artifacts across a raced build is exactly how an unreproducible number gets into a PR body.

PR #11113, opened 2026-08-01
fix(codegen): shared destroy_storage write-to-read conversion (#10645)

### LINKED ISSUE #10645: convergence: the storage_reads producer misses destroy_storage — the spec records at two sites, the guest at one

pirapira:
## Summary

Spec-alignment for #10645: guest `storage_reads` producer covered ordinary reads only; spec also converts `storage_writes` → `storage_reads` in `destroy_storage` (`state_tracker.py:560-580`).

### Shared routine

`destroy_storage` in `StorageWriteMap.lean`:
1. Walk `tx_storage_writes` for the address
2. `storage_read_record` each key
3. Swap-to-tail drop with undo `wasAbsent=2` (full 128 B row journaled)

### Review fix: full-row journal + sizing

Coord review found bare `count++` restore on wasAbsent=2 left a parked-tail overwrite hole. Fix journals the full map row and memcpy's it back on restore.

**Sizing derivation (not corpus-chosen):**
- SSTORE hard-caps the exec log at **W ≤ 16384** (`Storage.lean`). Each write pushes one undo (`wasAbsent` 0/1).
- `destroy_storage` converts map keys → reads and drops with a **second** undo (`wasAbsent` 2). Max destroy-converted rows per tx **D ≤ 16384** (each drop removes a key that entered via some prior append; appends ≤ W; re-append-after-destroy still costs an SSTORE against the same W bound).
- Shared journal needs **W + D ≤ 32768** records. Pre-destroy arena was 16384 × 64 B = 1 MiB under a 1:1 write:undo assumption; destroy_storage makes that false, so capacity is **2× the SSTORE cap**.
- Stride **160 B** so wasAbsent=2 holds the full 128 B row (not bare count++): **32768 × 160 = 5 MiB**.
- Account_writes arenas shift after the larger undo slab; chain ends at `0xa2f20000` ≤ `.data` `0xa3000000` (896 KiB margin).

### Call sites

| Spec site | Guest wire |
|---|---|
| `process_create_message` destroy before mark | `CreateFrameDescend` + `BlockVerdictCreationStage` before `mark_account_created` |
| `clear_account_preserving_balance` / EIP-6780 delete | `account_state_promote_delete_reads` → `destroy_storage` only |
| `destroy_account` via `modify_state` empty cleanup | **No discrete guest path** (site C absent) |

### Explicitly NOT bundled

`h_SSTORE` still records a storage read (extra vs pure `set_storage`). Separate convergence.

### Witness (LATENT alignment)

Three codex SHAs from #11090/#11093 all PASS pre-fix; fixtures lack SSTORE-then-destroy on wiped account. Closing as spec alignment, not live FA fix.

### Measure (post full-row journal + 5 MiB sizing)

| Population | Result | Run |
|---|---|---|
| selfdestructing_initcode_preserves_balance limit40 | **32/40** unchanged Class B | run-20260801T210611Z-1880920 |
| limit200 default | **123/77** null vs main | run-20260801T210652Z-1884466 |

Guest sha256 `f3ef366e3ee3f9822b623710ffee577cc93d2e1e94646bc68bf80727bc3297a7`.
`.text` stays `0x64960` (data-only growth).

Review Q answers (emitted asm):
1. ra at `.Lbvcr_ds_rev_d`: SAFE
2. wasAbsent=2 readers: only undo_push + restore_frame
3. Order OK on success; bare count++ hole fixed by full-row journal

## Test plan

- [x] lake/codegen build
- [x] 3 jal sites (2 CREATE + 1 promote)
- [x] region-map fixpoint; undo 5 MiB = 32768×160 guards
- [x] family 32/40 unchanged
- [x] limit200 123/77 null

Closes #10645.


pirapira:
Approved.

**Two spec producers against one guest producer, closed by reaching a shared routine rather than duplicating it.** execution-specs adds to `storage_reads` at two sites: `get_storage` (`state_tracker.py:295`) on every read, and `destroy_storage` (`:560-580`), which **converts writes into reads** — for each key in `storage_writes`, add to `storage_reads`, then delete the write — so a created-then-destroyed account still appears in the BAL. That second producer had no guest counterpart. One shared `destroy_storage` now serves both guest sites, with no local copies; site C was correctly dropped once it was established the guest has no discrete `destroy_account` empty-cleanup path.

**The witness result is reported honestly as latent.** All three same-transaction create-destroy input SHAs pass, and the family is 40/40, so this is spec alignment rather than a live bug — stated that way rather than dressed up.

**The review caught a corruption hole, and the author found it by answering the question rather than defending the change.** Swap-to-tail parked a row at the map tail; a later append could reuse that slot; a fail-restore then re-incremented the count and re-exposed a *different* row than the one dropped. Latent only because no fixture has SSTORE-then-destroy-then-fail. The fix is a full 128-byte row journal with memcpy restore — not the cheaper reorder, and the author's own spec reading is why: `restore_tx_state` **does** restore `storage_writes`, so a failed CREATE must put wiped keys back, which means destroy undos must live *inside* the restore window.

**The sizing question found the arena at half its derived requirement.** The answer is derived rather than chosen: the SSTORE exec-log hard cap gives `W ≤ 16384` writes with 1:1 write:undo; each destroy conversion adds a second undo, and `D ≤ 16384` because every drop removes a key that entered by a prior append. So the shared journal needs `W + D ≤ 32768` records, and at the new 160-byte stride that is **5 MiB** — where the first revision had 2.5 MiB. The arena chain now ends at `0xa2f20000` against `.data` at `0xa3000000`, 896 KiB of margin. An arena sized from the corpus rather than the derivation is the failure that surfaces as a sound-reject under load; this one is sized from the bound.

**Measurement:** `limit 200` at 123/77 identical to main, family 40 unchanged at 32/8 with Class B untouched. A null is the expected result for a latent gap.

`h_SSTORE`'s extra read recording is correctly left out of scope and noted on the issue as separate convergence.

PR #11114, opened 2026-08-01
Align block access request phases

pirapira:
## Summary

`replay_system_storage_writes_at_bai` and its Lean function are gone, with no emitted symbol references remaining. This PR aligns the checked system-request phase with execution-specs and retires the compensation that previously made the guest appear correct:

- Moves the EIP-7002, EIP-7251, and EIP-8282 request derivations into one callable post-user-loop phase.
- Stamps `current_block_access_index` directly with `bv_tx_count + 1` before all four derives.
- Incorporates each synthetic request transaction's read/write maps at BAI N+1.
- Keeps the BAI-0 producer for the three modeled startup storage rows.
- Hashes the supplied execution-request section before `block_verdict`; the receipts tail still verifies derived request bodies against the same `erh_requests_hash`.

The replay deletion is the convergence acceptance test: the derives no longer run early, so the second mechanism is unnecessary.

## Complete producer census for the deleted replay

The deleted routine read `current_block_access_index`, `bv_system_storage_log_count`, `bv_system_storage_txindex`, and `bv_system_storage_log`. It did not directly write any persistent global. For each matching N+1 side-arena row it called `storage_write_record` with the row address, slot key, current value, and a zero baseline.

That callee, plus its journal helper, wrote only the transaction-level storage map and its count (`tx_storage_writes_area`, `tx_storage_writes_count`), the transaction overflow flag if the map was full, and the transaction undo area and cursor (`storage_writes_undo_area`, `storage_writes_undo_count`). It updated an existing transaction row or appended a new row and journaled that mutation. It did not write the block-level map, system-row arena, request hash, account maps, or read maps. The old following `write_sets_incorporate_tx` was the separate block-level emit/merge/clear producer; the new N+1 request calls own their read/write incorporation at the correct phase.

## Review fixes

- The phase move exposed a pre-existing `bal_storage_matches_exec_log` fall-through bug when the system-row count is zero. The branch now enters the user-arena preamble before choosing the live execution log.
- Removing the old replay path also removed the successful EOA transaction's storage-map flush. `write_sets_incorporate_tx` is restored immediately before the deferred N+1 helper; `account_writes_incorporate_tx` alone only flushes account fields.

The first candidate sweep hid four disagreements behind six unscored load faults. The faults were the pre-existing storage-scan branch, newly reachable because the phase move produced a zero system count. After that was fixed, three scored rows identified the deleted EOA storage producer, and the remaining one is the separately filed #11115 case. The final candidate has zero runtime errors.

## Verification

- `LAKE_ARTIFACT_CACHE=false bash scripts/check-no-warnings.sh` — full `lake build`, 4800 jobs, 0 warnings under `EvmAsm/`.
- `scripts/check-asm-to-program.sh` — CLEAN, 391 converted definitions across 149 files.
- `NO_BUILD=1 ELF_DIR=gen-out/regionmap bash scripts/check-region-map.sh` — pass after generated-file fixpoint; `.text = 0x063d2c`.
- Structural gates passed: forbidden tactics, no hardcoded guest PCs, unimported modules, layering, approved heartbeats, and opcode structure.
- Baseline: clean `origin/main` commit `711e00325`, SHA `d23eb355be3cc6ae0bc5052f30ebdfdb824fada3b0432ec078df4c952888cb7e`.
- Candidate branch merge commit `1f8d7b83b`; canonical ELF SHA `09ab107995cc68abcf28e8ac926dda8e788449ae66a7b77a4c18e5146ddf8847`.
- Same Spike runner, seed 4242, 200 random rows: baseline selected/ran 200, errors 0, budget 0, fail 12, full 188, root 200, success 188, tail 200; candidate selected/ran 200, errors 0, budget 0, fail 13, full 187, root 200, success 187, tail 200.
- `eest-ab-compare.py` joined 198 rows and passed all self-checks: FA 0 to 0, FR 12 to 13, no new false accepts, zero status differences, and one output-byte difference.

The sole remaining output-byte difference is `selfdestruct_to_precompile_state_access_boundary`, filed separately as [issue #11115](https://github.com/Verified-zkEVM/evm-asm/issues/11115). The fault investigation is recorded in [issue #11116](https://github.com/Verified-zkEVM/evm-asm/issues/11116) and is closed.

## Supplied/derived request hash

The supplied-versus-derived mismatch path is retained: emitted debug assembly hashes the supplied section into `erh_requests_hash` before `block_verdict` and branches to fail code 24 on failure; the receipts tail calls `requests_hash_verify` against that hash and branches to fail code 55 on a derived-body mismatch.

## Follow-up

The current-base A/B comparison opens exactly one scored false reject and closes none. It is the known #11115 selfdestruct case above; no new false accepts were introduced.


pirapira:
Approved.

**What this does.** The guest ran the four request-predeploy derivations *before* the transaction loop and discarded their transaction-local writes, while a second mechanism — `replay_system_storage_writes_at_bai` — re-presented those rows afterwards so the output came out right. execution-specs derives requests **after** the body (`fork.py:916-926`) with `ulen(txs)+1`, uses the builder's default `0` for the block-start system calls (`:897-906`), and `i+1` per transaction (`:1040-1042`). This moves the derivations to the post-user-loop seam, stamps `N+1` **directly into `current_block_access_index`** rather than passing it as metadata nothing consumed, and adds the missing block-level `0` producer.

**The acceptance test is the proof, and it is checkable with a grep:** `replay_system_storage_writes_at_bai` and its Lean definition are **gone, with no symbols remaining**. The compensation existed only because the derivations ran early; they no longer do, so it is unnecessary. One mechanism where there were two. A phase move that left the replay in place would have been a relocation, not a convergence.

**This PR is worth reading for how its number changed under scrutiny**, because none of it is visible in the final delta:

* first reported as `FR +1` — the body showed **six errors and seven byte differences** the summary omitted;
* those were **guest faults**, `mcause=0x5`, all six at one instruction in `bal_storage_matches_exec_log`, with `mtval`s under `0x100` betraying a near-null base;
* the cause was a **pre-existing** branch to the loop body instead of its `t3`-initialising preamble, unreachable until the phase move produced a zero system count;
* with the faults fixed the true delta appeared: **`FR 13 → 17`, not 14** — six unscored rows had been hiding four disagreements;
* three of those four were a **deleted producer** — the EOA postlude `write_sets_incorporate_tx` went out with the replay — and restoring it closed them.

**An error is not a neutral outcome; it is an unscored row.** A measurement with nonzero errors does not have a smaller delta than it looks, it has an unknown one.

**Final evidence, on current base after merging main:** baseline `d23eb355…` `FA=0 FR=12 errors=0 budget=0`; candidate `09ab1079…` `FA=0 FR=13 errors=0 budget=0`; 198 joined, zero status differences, **one** output difference — `selfdestruct_to_precompile_state_access_boundary`, filed as #11115. An honest `+1` with a named cause.

**The producer census is in the body**, which was the condition I cared about most: a function doing two jobs under one name may be doing three, and enumerating everything the replay wrote — rather than only the producer that failed — is what makes its deletion safe rather than lucky.

PR #11117, opened 2026-08-01
fix(codegen): make top-level precompile callback entry self-contained

pirapira:
## Summary

The top-level MTx precompile callback entered the shared precompile selector *after* the selector's address setup.  Its label therefore encoded an interior position rather than a self-contained entry contract: `t0` was stale, address classification returned zero, and the precompile gas arm was never reached.

The fix moves `.Lbv_tx_gas_precharge_pc0_prefix` before `mv t0, t2; addi t0, t0, 72`, so both fallthrough and callback entry establish the selector's input.  The callback address is materialized from this label at `BlockVerdictMtxEoa.lean:44`; it is not a literal.  The callback-materializer census found one instance of this protocol.  The deferred-delegation callbacks and `runtime_tx_oog_hook` are different callback contracts, not additional instances.

The change is only the four-line source movement.  The temporary mode-2/debug scaffolding was removed completely.

## Measurement

The candidate was built from current main `35f8564a4c434ea0f7750b4e4e716b1bfdc4f84d` and measured against a separately generated parent ELF from that same commit:

- parent ELF SHA-256: `8882387bab4e4d97e8723409851eafa14526f669d393aade68af299a4aacda27`
- candidate commit: `7b9313279`
- candidate ELF SHA-256: `c91c205a02fef25068b8e8308e73e6cf6d6ce9819e4f5d51b2cc0263169b600c`
- manifest: `/tmp/codex2-11106-rebased/pre-random-200/manifest.tsv` (200 rows)
- runner: `scripts/spike/spike_run`, SHA-256 `bae4826988540415b9bf862248caa7499b2ffbb9eec5cfab61cce00e089cf2c3`, `WORKERS=30`, `SPIKE_OUTPUT_LEN=256`, timeout 120s

Command (run once per ELF):

```text
WORKERS=30 SPIKE_OUTPUT_LEN=256 python3 scripts/sweep_one.py --manifest /tmp/codex2-11106-rebased/pre-random-200/manifest.tsv --workers 30 --dump-len 256 --spike "$PWD/scripts/spike/spike_run" <elf> <tsv>
```

| leg | FA | FR | errors | budget | OK |
|---|---:|---:|---:|---:|---:|
| parent | 0 | 5 | 0 | not separately reported (derived 0) | 195 |
| candidate | 0 | 3 | 0 | not separately reported (derived 0) | 197 |

Both TSVs have 200/200 `rc=0` rows and 2048-byte runner outputs (the runner emits its fixed 2048-byte buffer, which includes the full requested 256-byte window).  This runner does not expose a separate step-budget field: budget `0` is therefore only an inference from all rows returning `rc=0` with no exhausted/missing output, not a fourth measured counter.  Both legs used the same runner SHA, so the paired delta is internally valid; absolute FR is not directly comparable to measurements made with other runner binaries.

Reverse flips are empty.  Forward flips are:

- `00029_test_tx_entry_point...`: reject → pass
- `00052_test_bal_precompile_funded...precompile_0x02`: reject → pass

`00052` is byte-perfect at 345/345 on the candidate.  The independent `00029` flip shows this is a callback-entry class fix rather than a witness-specific patch.

## Diagnosis chain

Before patching, the precompile gas selector audit found all eighteen precompile base/formula cases matching execution-specs (zero formula mismatches).  Receipt-gas publication was not the defect.  The mode-2 hypothesis was falsified by the commit log: `00052` never reached `.Ldtrc_mtx_precompile_success`, with no PC hits and no lane writes on that route.  The actual trace showed `runtime_tx_post_top_frame_fn` holding the callback address at `0x80015ea0`, while the selector's setup was at `0x800178e0` and `0x800178e4` immediately before the targeted label.  Entering the label left `t0` stale, classified the target as non-precompile, and skipped the SHA-256 arm.  Making the label self-contained removes the position-drift hazard instead of correcting one captured pointer.

Build gate: `LAKE_ARTIFACT_CACHE=false lake build` completed successfully (4800 jobs).  No `merge!` label was applied by this PR author.


pirapira:
Approved.

**A four-line diff at the end of a five-hypothesis investigation, and the refutations are worth as much as the fix.**

`runtime_tx_post_top_frame_fn` held `0x80015ea0` while the address-setup instructions sat at `0x800178e0`/`0x178e4` — **immediately before** the label that callback targeted. So the callback entered past its own setup, `t0` was stale, the selector classified as 0, the SHA256 arm was never reached, and there was no success tail to publish from. One wrong entry address, four downstream symptoms.

**The fix makes the entry self-contained rather than correcting the pointer.** Moving the selector label before its setup means the label now carries the *intent*; a corrected pointer would drift again the next time that sequence moves. **A label position is not a contract**, and this is the second time today that principle has cost us — the census listing `tx_eip7702_decode` as never-referenced was the other.

**Four hypotheses died by measurement before the right one**, and keeping that chain in the body is what stops the next person walking it:

* the SHA256 formula is correct — `60 + 12·ceil32(len)/32` matches `gas.py:96-104`;
* **all eighteen precompile bases match execution-specs**, zero mismatches — a bounded negative worth keeping so nobody re-derives it;
* receipt-gas publication was not the defect;
* the mode-2 hypothesis was refuted by **commitlog**: 00052 never executes `.Ldtrc_mtx_precompile_success` at all — no PC hits, no writes to `bv_mtx_precompile_lane`. That trace was two edits away from a fix applied to a path the witness never enters.

It also cleared an open question of mine: 00052 was already failing before #11091 *and* does not reach the lane that PR created, so nothing here is connected to it.

**Population stated with its enumeration:** one materialiser at `BlockVerdictMtxEoa.lean:44`, computed from a label rather than a literal; the deferred-delegation callbacks and `runtime_tx_oog_hook` are different protocols, not further instances. A class of one is durable when the enumeration is given.

**Evidence:** parent `8882387b…` vs candidate `c91c205a…`, same manifest and runner, 200 rows: `FA=0 FR=5 OK=195 errors=0` → `FA=0 FR=3 OK=197 errors=0`, **forward flips 00029 and 00052, reverse none**. 00029 closing unexpectedly is the strongest single fact — a fix diagnosed against one witness that incidentally closes a second is behaving like a root cause.

Runner provenance and the 2048-byte emission are stated, with `budget` correctly labelled as inferred from `rc=0` rather than measured. These numbers are internally valid but not comparable to PRs measured on the `c6ac09` runner.

Issue #11122, opened 2026-08-01
The stateless witness does not carry storage trie nodes: every storage-witness lookup walks the EMPTY trie root (64/64)


pirapira:
Every storage-witness lookup in the stateless guest fails, in every phase, for every caller. The guest is
**not** at fault: the addresses and slot keys it passes are correct, the roots are carried correctly, and
the MPT walk faithfully reports what it is given. What it is given is the **empty trie root**.

⇒ **The witness does not carry the storage trie nodes.** This is an input-construction / witness-capture
problem, upstream of codegen.

## 1. The measured funnel

`main` at `d97788890`, plus a two-instruction starve (preload count forced to 0, callee seed drain jumped
over) so cold misses actually occur — neither changes `.text` size. Guest ELF sha256
`475a26f1fb2d1505f60fba473fd0c52424865727ff9115664a971a46eb19d84d`. `SPIKE_COMMITLOG` through a fifo;
all counters are PCs read out of the objdump, so no probe code was added.

Rows, by relpath: `00000_test_sstore_xto_y_fork_Amsterdam-…-d4-g0__b0` and
`00163_test_sstore_xto_x_fork_Amsterdam-…-d1-g0__b0`. Both **x-to-…**, i.e. **nonzero pre-state slot**.
Liveness 6 902 036 / 6 840 396 commitlog lines. Counting **every** internal exit of
`slot_at_header_state_root`, all callers pooled — **identical on both rows**:

| | count |
|---|---|
| total calls | 82 |
| died in `header_extract_state_root` (status 4) | 16 |
| reached `account_at_address` — header OK | 66 |
| died in `account_at_address` | 2 |
| reached `slot_at_index` — header **and** account resolved | 64 |
| **died in `slot_at_index`** | **64** |
| **succeeded** | **0** |

Attribution is forced, not plausible: `h_SSTORE` makes 32 calls of which exactly **16** load a zero header
length, matching the status-4 count; then 16 + 34 (`bal_emit_storage_changes`) + 16 (`predeploy_preload`)
= **66** = the header-OK count, nothing left over. Row 2 shifts the split to 17 + 33 + 16 and still totals
82 / 66.

One level down, `slot_at_index` has exactly two failure exits: **all 64 fail in `mpt_lookup_by_key`** and
**zero** reach `slot_decode_u256`.

## 2. The empty trie root, and the guest's exoneration

[...]

PR #11124, opened 2026-08-01
docs(codegen): tier 3 DOES run in production — correct two reachability claims (#11105)

pirapira:
Doc-only follow-up to #11121. That PR removed three false claims from this docstring and **preserved a
fourth**; measuring the *unstarved* guest refuted it, along with a second claim elsewhere in the same file.
Guest byte-identical — see the gate below.

## The two claims corrected

| claim | where | status |
|---|---|---|
| *"this chain never executes in production, because the preload means SSTORE never sees a cold miss"* | `storagePrestateResolveAsm` | ⛔ **it executes 16 times on one row of the unmodified guest** |
| *"that path is unreachable while the BAL preload stages every slot a block reads"* | `storagePersistentLogFindAsm` — the recorded reason #10874's `h_SLOAD` read is not landed | ⛔ **the preload does not stage every slot; cold misses happen in production** |

Both were reachability assertions with a plausible *why* and an unmeasured *happens*. The second is the more
consequential: it is the stated justification for not landing the demand-driven read, and it is wrong.

## Measured on the UNMODIFIED guest

Not on a probe build. `main` `d97788890`, ELF `fbb83b69b167506518644fa38a7e73f17322c30c4cc80ba7f5ebc04c0dde54d3`,
no starve, counters are PCs read from the objdump:

| row | tier-3 entries | header len `env+584` | status-4 header fails |
|---|---|---|---|
| `00000_test_sstore_xto_y_…-d4-g0__b0` | **16** | `0x27c` at all 16 | **0** |
| `00163_test_sstore_xto_x_…-d1-g0__b0` | **17** | `0x27c` at all 17 | **0** |

⇒ **Tier 3 runs in production and works.** Every call gets a well-formed header, resolves the account, walks
the storage trie and correctly reports not-found — correct because the accounts looked up have **zero
pre-state `storage` entries** in the fixture, so an empty storage root is the right answer and the fallback
to value 0 is right.

## Why the previous text said the opposite

The 82-call funnel #11121 recorded came from a build with the **preload starved** to force cold misses. That
is an artificial condition, and the starve itself *added* the 16 zero-length-header calls — production has
**zero**. The docstring now says this explicitly so the numbers are not re-read as production behaviour.
GH #11122, filed on that reading, is closed as invalid; GH #11105 is closed with no code change.

## What is deliberately still flagged as unproven

Every measured cold miss resolved a **genuinely-absent** slot, so the not-found path is exercised and the
**found** path is not. A present-slot case remains the open gate for #10874, and because every non-zero
status flattens to value 0, a broken authenticated read would be indistinguishable from an absent slot. The
docstring says so rather than implying tier 3 is fully validated.

## One operand question settled, not asserted

`ExecutionWitness` (`execution-specs/src/ethereum/forks/amsterdam/stateless.py:30-51`) has exactly three
fields — `state`, `codes`, `headers`. There is **no separate storage-node container**; `state` is one flat
pool of trie-node preimages. So passing the same pointer for both the state and storage arguments is
**correct**, and an earlier "two containers where the spec has one" flag against this code is retracted here.

## Gate: byte-identical guest

| | sha256 |
|---|---|
| `main` `5cfc7b560` | `d23eb355be3cc6ae0bc5052f30ebdfdb824fada3b0432ec078df4c952888cb7e` |
| this branch | `d23eb355be3cc6ae0bc5052f30ebdfdb824fada3b0432ec078df4c952888cb7e` |

Comment-only, so **no RegionMap repin and no layout regeneration**. `lake build` clean,
`check-forbidden-tactics` OK, file 919 lines (cap 1500).

Refs #11105, #11122, #10874.

### DIFF
--- a/EvmAsm/Codegen/Programs/Storage.lean
+++ b/EvmAsm/Codegen/Programs/Storage.lean
@@ -214,10 +214,19 @@ def sstoreValueTransitionGasAsm : String :=
 
     Extracted verbatim from `h_SSTORE`'s inline scan. ⚠️ It currently has ONE
     consumer: a second was intended for `h_SLOAD`'s cold-miss path (GH #10874) and
-    is not landed, because that path is unreachable while the BAL preload stages
-    every slot a block reads. Numeric local labels (`1:`/`2:`/`3:`) are
-    unique-per-use, so a second instantiation can coexist without a prefix
-    parameter when the read lands.
+    is not landed.
+
+    ⛔ The reason recorded here used to be *"that path is unreachable while the BAL
+    preload stages every slot a block reads"*. THAT IS REFUTED. `h_SSTORE`'s
+    cold-miss resolve -- which only runs when this scan MISSES -- executes **16 and
+    17 times** on two rows of the UNMODIFIED guest (`main` `d97788890`, GH #11105).
+    The preload does **not** stage every slot a block reads, and cold misses do
+    happen in production, so the `h_SLOAD` read is not blocked by unreachability.
+    What it does still need is a **present-slot** case: every measured cold miss
+    resolved a genuinely-absent slot, so the found path is unexercised.
+
+    Numeric local labels (`1:`/`2:`/`3:`) are unique-per-use, so a second
+    instantiation can coexist without a prefix parameter when the read lands.
 
     Register convention -- identical at both sites, which is what makes this
     shareable at all: `x20` = env base (`env.ADDRESS` at `+0..32`, log length at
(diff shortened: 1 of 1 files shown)


pirapira:
Approved, including the widening.

**The claim I asked for.** #11121 (merged 22:15) recorded sixteen header faults as a property of the guest. They are not: measured on the **unstarved** build, all sixteen execute with a well-formed header, the account resolved, the storage trie walked, a correct not-found and a correct fallback to zero. The faults were an artefact of a starved probe build — a fault only its author had ever caused. #11105 is closed on this.

**The claim added, and it is the more consequential one.** `storagePersistentLogFindAsm`'s docstring records the *reason* #10874's `h_SLOAD` read is not landed: that the path is unreachable while the BAL preload stages every slot a block reads. **The preload does not stage every slot.** That is a stale reachability claim, and unlike a wrong description it is a wrong *justification* — it tells the next reader the work is blocked when it may not be.

Widening for it was right. This file has now cost us three times in one day, and each time the damage came from prose that **foreclosed a direction** rather than merely describing something inaccurately: the witness walk described as returning absent mid-execution (a null pointer, fixed in #11110); callers described as working when none is known to succeed; and now a blocker described as insurmountable when it has not been shown to exist.

**The honest state, and the prose says so rather than flipping to the opposite overclaim:** the tier-3 witness path *works* — not blocked, not architecturally constrained. What has **not** been shown is that it returns the right value for a slot that **is** in the pre-state. That present-slot case remains unmet, and recording "works, not yet verified on a present slot" is the accurate version.

**Comment-only, verified as such:** byte-identical emitted guest, all six checks present with `MERGEABLE` rather than `CONFLICTING`, and files within the 1500-line cap. Confirming the check *set* before reporting green is the #11095 lesson, applied unprompted for the second time tonight.

pirapira:
⚠️ **Self-correction on one sentence in this PR, found after labelling.** The load-bearing numbers are unaffected; the mechanism named for the *old* numbers is wrong.

This PR (body and docstring) says the previous 82-call funnel came from "a build with the preload **starved**". **It was not starved.** I measured my own intervention rather than trusting it:

| | production | my "starved" build |
|---|---|---|

The expand loop is guarded by `x6` (the preload **count register**), not by `env+448`. My edit changed `sd x6, 448(x20)` → `sd x0, 448(x20)`, which zeroes only the **scan bound**. All 16 preload entries were still written.

So the probe produced a **corrupted** log, not an empty one — arena populated, count claiming zero, then each SSTORE appending from index 0 upward over those rows. A state production cannot reach, which is *why* the 16 zero-length-header calls appeared.

**What survives, and it is everything this PR asserts:** the 16/17 tier-3 entries, the `0x27c` header length at every one, and the zero status-4 failures were all measured on the **unmodified** guest with no intervention. The conclusion — that the old condition was artificial and production has zero header faults — holds. Only the explanation of *how* it was artificial is wrong.

Merging as-is is fine and I will file a one-sentence correction; unlabel it if you would rather I fix the sentence in place, which avoids a third correction to this docstring.

Issue #11115, opened 2026-08-01
Investigate new BAL nonstorage FR after request-phase alignment


pirapira:
## Observation

The clean current-main versus phase-aligned A/B on Spike (seed 4242, 200 rows) showed one new scored false reject and no fixed false rejects.

New row:

`00052_test_selfdestruct_to_precompile_state_access_boundary_fork_Amsterdam-precompile_0x00000000000000`

Fixture:

`blockchain_tests/for_amsterdam/tangerine_whistle/eip150_operation_gas_costs/eip150_selfdestruct/selfdestruct_to_precompile_state_access_boundary.json`

Candidate debug output reports `bv_fail=44`, the BAL nonstorage comparison failure (`.Lbv_bal_nonstorage_fail`): a declared BAL account balance/nonce change differs from the execution-derived nonstorage effect. The baseline accepts the same input. Candidate and baseline artifacts:

- baseline commit `366691458c7c04175c77893544c63a1e3208393b`, SHA `6b900b918309527f699886dcec096e53a4a7681af307827a9365116481977569`
- candidate commit `75ca856ec8170cbabbfd3efa61d56cf73599e2d5`, SHA `699446ae6cbacb0abcb677a57ca6fa37d0b11b4dab2ffd003f8f65a9a07cf306`

## Decoded comparator evidence

The fixture BAL has 11 account rows. The account `d97bfeafbfca6f85cfa81d7e5dca82699b29a21a` is row 9 and declares no balance or nonce changes.

At the `bal_all_accounts_nonstorage_consistent` comparator boundary, the baseline and candidate aggregate arrays are identical except for that account:

- baseline: `pre_balance=0`, `post_balance=0`, `pre_nonce=0`, `post_nonce=0`, effect mask `0`
- candidate: `pre_balance=0`, `post_balance=0`, `pre_nonce=0`, `post_nonce=1`, effect mask `0x03`

The nonstorage effect record has no BAI field, so this is not an index-only stamping difference. The candidate introduces a nonce effect for an account whose declared BAL row is unchanged. The changed request phase is therefore exposing a value/effect-record disagreement, not merely moving the disputed row from BAI 0 to BAI 1. The likely mechanism remains a phase-dependent state/effect publication or stale state around the deferred helper; this issue is intentionally separate from the convergence PR.

[...]

PR #11126, opened 2026-08-01
fix(codegen): honest nonstorage effect mask (Class B FR44)

### LINKED ISSUE #11125: codegen: record_nonstorage_effect always-set component mask poisons aggregate (Class B FR44)
## Summary

`record_nonstorage_effect` set the raw component mask (byte 20) to `hasBalance|hasNonce` unconditionally (except the EIP-7702 nonce-only entry). Callers such as `record_message_value_transfer` pass **dummy equal nonces** `(a3=a4=0)` intending a balance-only effect. AccountWrite already derived bits from real pre/post deltas; the **raw** mask did not.

When a later real nonce-bearing record lands on the same address (e.g. SELFDESTRUCT origin clear with live nonce `1→1` still carrying hasNonce under the old always-set rule — or any real nonce delta), `nonstorage_effect_aggregate` takes **first** hasNonce pre and **max** post → phantom `nonce 0→1` against a BAL-empty account → `bal_all_accounts_nonstorage_consistent` rejects **bv_fail=44**.

## Family split (why equal-length FR44)

`selfdestructing_initcode_preserves_balance` residual after #11082/#11092:

| pattern | result |
|---|---|
| `post_send_opcode_SELFDESTRUCT` count **0** | PASS |
| `post_send_opcode_SELFDESTRUCT` count **1\|3** | FAIL bv44, **rb==sp** (equal length) |

Equal lengths made length/digest diffs blind — only **per-entry** exec agg vs BAL decode shows it.

Fail path: post_send = `CALL(donor)` where donor code is `SELFDESTRUCT(victim)` (not entry SD). Donor is **pre-deployed** (nonce=1). Sequence:

1. CALL value credit via `record_message_value_transfer` → raw record nonce **0→0** with hasNonce set
2. Donor SD clear → real nonce **1→1** with hasNonce
3. Agg donor: pre_nonce=0, post_nonce=1 → phantom change
4. BAL correctly empty for donor (net bal 0, nonce unchanged)

count=0 never touches the donor → no poison.

## Mask population

Exactly **two** mask bits in this routine: `hasBalance=1`, `hasNonce=2` (no hasCode). Pre-fix `a5!=2` set **both** always. Class size **2**; live break was hasNonce + dummy zeros. hasBalance-always was latent (rmv only records when value ≠ 0).

## Fix

Derive raw mask from real pre/post deltas (same rule as AccountWrite). `a5=2` stays nonce-only forced.

## Evidence

- Baseline main `d97788890` guest `fbb83b69…`: family limit40 **32 pass / 8 fail** all bv44 (`run-20260801T221948Z-2362052`)
- Fix guest `dd3fdb40…`: family **40/40** (`run-20260801T223247Z-2465040`); limit200 **123/77** identical actual_hex to main baseline `run-20260801T215641Z-2143636` (reverse flips **0**)

pirapira:
## Summary

`record_nonstorage_effect` set the raw component mask to `hasBalance|hasNonce` unconditionally (except EIP-7702 nonce-only). `record_message_value_transfer` passes dummy equal nonces `(0,0)` for balance-only moves. AccountWrite already derived bits from real deltas; the raw mask lied. Aggregate folded first hasNonce pre=0 with a later real nonce into a phantom `0→1` and tripped **bv_fail=44** against BAL-empty accounts.

Fix: derive both raw mask bits from real pre/post deltas (same rule as AccountWrite). `a5=2` stays forced nonce-only.

Fixes #11125. May also clear the phantom-nonce shape in #11115 (same `pre_nonce=0 post_nonce=1 mask=0x03` on BAL-empty row).

## Why the family split that way

`selfdestructing_initcode_preserves_balance` residual Class B (after #11082/#11092 Class A):

| `post_send_opcode_SELFDESTRUCT` count | result |
|---|---|
| 0 | PASS |
| 1 or 3 | FAIL bv44, **equal length** rb==sp |

Equal lengths made length/digest diffs useless — only per-entry exec-agg vs BAL decode shows the defect.

count≥1 path: post_send is `CALL(donor)` where donor code is `SELFDESTRUCT(victim)` (pre-deployed, nonce=1), not entry SD. CALL value credit records nonce 0→0 with hasNonce; SD clear records live nonce 1→1 with hasNonce; agg reports phantom 0→1; BAL correctly empty for donor. count=0 never touches the donor.

## Mask population

Exactly **two** bits in this routine: `hasBalance=1`, `hasNonce=2` (no hasCode). Pre-fix both always-set on `a5!=2`. Class size 2; live break was hasNonce + dummy zeros. hasBalance-always was latent (rmv only records when value ≠ 0).

## Measure (FR bar)

| leg | guest sha256 | result |
|---|---|---|
| main `d97788890` family limit40 | `fbb83b69…dde54d3` | 32 pass / 8 fail bv44 — `run-20260801T221948Z-2362052` |
| fix family limit40 | `dd3fdb40…54e83678` | **40/40 full** — `run-20260801T223247Z-2465040` |
| main limit200 | `fbb83b69…` | 123/77 — `run-20260801T215641Z-2143636` |
| fix limit200 | `dd3fdb40…` | 123/77 — `run-20260801T223358Z-2478848` |

Reverse flips pass→fail: **0**. limit200 actual_hex **identical** 200/200 vs main baseline (root+succ+tail+full same). Family movement 32→40 is outside default limit200 order.

## Files

- `EvmAsm/Codegen/Programs/NonstorageEffectLog.lean` — honest raw mask
- `EvmAsm/Codegen/Programs/EIP7708Logs.lean` — comment only

No layout regen (body size change does not require GuestAddrs for string-emitted path; EEST rebuild emits ELF directly).

## Note

main is moving (#11114 labelled). Rebase if mergebot needs it; do not add merge! from author.

### DIFF
--- a/EvmAsm/Codegen/Programs/EIP7708Logs.lean
+++ b/EvmAsm/Codegen/Programs/EIP7708Logs.lean
@@ -443,8 +443,8 @@ def messageValueTransferFunction : String :=
   "  la s5, message_value_transfer_sender_post\n" ++
   ".Lrmvt_recipient_pre:\n" ++
   "  mv a0, s5; mv a1, s2; la a2, message_value_transfer_recipient_post; jal ra, u256_add_be; bnez a0, .Lrmvt_done\n" ++
-  -- Equal nonce inputs make `record_nonstorage_effect` derive an honest
-  -- balance-only AccountWrite mask rather than asserting a nonce component.
+  -- Equal dummy nonces: record_nonstorage_effect derives balance-only raw mask
+  -- and AccountWrite bits (pre_nonce == post_nonce clears hasNonce).
   "  mv a0, s0; mv a1, s4; la a2, message_value_transfer_sender_post; li a3, 0; li a4, 0; jal ra, record_nonstorage_effect\n" ++
   "  mv a0, s1; mv a1, s5; la a2, message_value_transfer_recipient_post; li a3, 0; li a4, 0; jal ra, record_nonstorage_effect\n" ++
   ".Lrmvt_done:\n" ++
--- a/EvmAsm/Codegen/Programs/NonstorageEffectLog.lean
+++ b/EvmAsm/Codegen/Programs/NonstorageEffectLog.lean
@@ -93,7 +93,15 @@ def nonstorageEffectHasNonce : Nat := 2
     `record_nonstorage_effect_nonce_only_after_account_state` is the EIP-7702
     authorization variant.  It carries an honest nonce-only raw mask while
     retaining the AccountWrite publication at the authorization's current BAI;
-    the authorization already owns the AccountState mutation. -/
+    the authorization already owns the AccountState mutation.
+
+    Raw component mask (byte 20) is derived from actual pre/post deltas — the same
+    rule AccountWrite already used — so balance-only producers that pass equal
+    dummy nonces (notably `record_message_value_transfer` with a3=a4=0) do not
+    publish a stale nonce-0 component.  Mixing those with a later real nonce
+    record used to make `nonstorage_effect_aggregate` report a phantom nonce
+    change (first pre=0, max post=real) and trip bv_fail=44 on BAL-empty
+    accounts (Class B selfdestructing_initcode_preserves_balance post_send SD). -/
 def recordNonstorageEffectFunction : String :=
   "record_nonstorage_effect:\n  li a5, 0\n  j .Lrnse_entry\n" ++
   "record_nonstorage_effect_after_account_state:\n  li a5, 1\n" ++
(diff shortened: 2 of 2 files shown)


pirapira:
Approved.

**A mask that claims "changed" when pre equals post is a lie the aggregator cannot detect**, and that is exactly what `record_nonstorage_effect` was emitting. `record_message_value_transfer` passes dummy equal nonces (0, 0), the raw mask set `hasNonce` regardless, and a post-send CALL donor followed by SELFDESTRUCT then folded that phantom pre-nonce of 0 with a later real 1 into an aggregate nonce change 0 → 1 — against a declared BAL that correctly has the donor empty.

**The fix is the shape that has been right all day: derive both bits from real pre/post deltas, because `AccountWrite` already does exactly that.** Make the deficient path do what the working one does, by the same rule, rather than inventing a second one. Only the raw byte-20 mask lied.

**The population question was answered rather than assumed, and the answer was not the obvious one.** There are exactly two bits — `hasBalance` and `hasNonce` — and the pre-fix path set **both** unconditionally. So nonce was not the only always-set bit; the class is size two. `hasBalance` was latent only because `record_message_value_transfer` records solely when the value is nonzero, so the balance genuinely changes. That is a latent defect fixed alongside the live one, and stating why it never fired is what makes it safe to fix now.

**Evidence:** family **32/40 → 40/40 full**, and `limit 200` with `actual_hex` **identical** to the main baseline, reverse flips zero. A full-family close with no corpus movement is the strongest combination available for a behavioural fix.

**Worth keeping from the diagnosis:** the eight failures were **equal-length** (`rebuilt == supplied`), so a byte-count comparison was structurally blind and only a per-entry decode could see the substitution. The split was `ps_SELFDESTRUCT` count 0 passing while counts 1 and 3 failed. Anyone meeting an equal-length `FR44` will otherwise reach for a length diff and learn nothing.

The note that this may also explain #11115 — same phantom nonce 0 → 1 shape — is worth checking there rather than assuming, but it is a good lead for whoever takes it.

PR #11129, opened 2026-08-01
Charge EIP-7702 authorization state gas per authorization

pirapira:
## Summary

Charge EIP-7702 authorization state gas at the authorization boundary, so an
authorization-phase out-of-gas stops the authorization loop at the first
unaffordable authorization instead of applying later authorization reads.

The execution-specs ordering is pinned at `e5a8caf1b`: `eoa_delegation.py:174-180`
recovers the authority, records the authority account access, and then performs
the state-dependent charge. `state_tracker.py:139-142` and `:199-202` preserve
the read even when the write journal is rolled back. The guest now charges each
authorization immediately after its account-read/validation step and returns
OOG before the next authorization when the live state-gas plus regular-gas
reservoir is insufficient. The existing aggregate is retained as reporting
compatibility for the direct single-transaction caller; the dispatcher callback
uses the per-authorization consumption path.

This PR also removes a separate bookkeeping defect found while wiring the
first fix: `evm_state_gas_used` and `evm_state_gas_spilled` were incremented in
the authorization callback and then counted again by the outer aggregate fold.
The callback now consumes the live reservoirs while the aggregate remains
reporting-only for that path. This second correction is behaviourally invisible
on the captured corpus, so it is described separately rather than attributed to
the authorization-loop fix.

## Evidence

- Witness `01786`: the predicted rebuilt length was `602 - 9*27 = 359`; the
  final ELF reports `fail_code=60`, `shadow_status=1`, `rebuilt_len=359`,
  `supplied_len=332`. The nine extra all-empty authorization entries are gone;
  the remaining 27-byte recipient entry is the separate parked recipient-row
  half of the original issue (#11078).
- Final ELF SHA-256:
  `910884377d8ed1f9bc0527e090a5ff3a04e51a02a41185687077e856f704cb50`.
- Captured 200-row A/B on the same manifest and Spike runner, with the parent
  artifact from current-main commit `d97788890` (parent SHA
  `fbb83b69b167506518644fa38a7e73f17322c30c4cc80ba7f5ebc04c0dde54d3`) and
  this candidate SHA: parent `FR=16, OK=184, FA=0, errors=0`; candidate
  `FR=16, OK=184, FA=0, errors=0`; forward flips `0`, reverse flips `0`.
- The final fixpoint kept `.data=0x5370` and `.bss=0x1c101b60`; the measured
  `.text` pin is `0x63e8c`.

## Checks

`check-region-map.sh`, `codegen-stateless-link-check.sh --no-build`, and
`check-asm-to-program.sh` pass after regeneration (`391` converted definitions
across `149` files). No KAT was constructed.


pirapira:
Approved.

**The defect, in one sentence:** the guest's authorization loop processes every authorization regardless of gas, where execution-specs halts at the first one that cannot be charged.

**The mechanism is fully pinned.** Transaction gas is `0x1d1a8`, below `TX_MAX_GAS_LIMIT`, so the state-gas reservoir is zero. Authority zero (`c0f6`) recovers, `get_account` is called (`eoa_delegation.py:174-180`), and the state-dependent charge raises OOG immediately. `account_reads` survives the rollback — hence the declared BAL carries `c0f6` and nothing else, while the guest emitted ten all-empty rows for ten authorities it should never have reached.

**The strongest evidence here is the arithmetic, not the FR line: predicted 359 before building, measured 359.** `602 − 9 × 27` was derived from the decomposition *before* the change existed, and the witness landed on it exactly. A number that arrives where it was predicted demonstrates the mechanism was understood; a number that merely improves demonstrates iteration. This project has had several of the latter today and this is the former.

**The witness stays red and that is the correct outcome.** The residual 27 bytes are the recipient row — a genuinely separate, parked defect with its own four recorded dead ends. A green family here would have meant the decomposition was wrong. Judged on byte movement and reverse-zero, not on the family.

**Two defects, correctly separated in the body.** The auth-loop halt is the one with the witness. The `evm_state_gas_used`/`spilled` **double-accounting** was found while fixing it, is behaviourally invisible on the corpus, and would otherwise have been silently attributed to the first change by anyone reading only the FR count. Stating it separately is what keeps the attribution honest a year from now.

**Evidence:** parent `fbb83b69…` (main `d97788890`) vs candidate `910884377…`, same deterministic 200-row manifest: both `FA=0 FR=16 OK=184 errors=0`, forward and reverse flips zero. Post-regen gates all pass — `check-region-map`, `codegen-stateless-link-check`, `check-asm-to-program` at 391 converted defs.

**Route to it worth preserving:** four hypotheses died by measurement before this one — the SHA256 formula (correct), all eighteen precompile bases (matching spec, zero mismatches), receipt-gas publication (not the defect), and a mode-2 tail the trace showed the witness never enters. Then two framings of mine died too: warm-versus-read, and a wrong-stage record. The count of *processed authorities* was the thing that settled it.

PR #11132, opened 2026-08-01
chore(codegen): retire legacy committed storage table

pirapira:
## Summary

Retire the dead single-page committed-storage table. The emitted guest already uses the chunked committed map for cross-transaction reads; this change removes the duplicate legacy storage without rewiring the live reader path.

The lifetime comparison is:

- The append-only execution log remains authoritative for current-transaction `SLOAD`/`SSTORE` last-write-wins and first-touch `original`/`current` values.
- `tx_storage_writes` plus its undo map remains the mutable transaction overlay, promoted or discarded at the transaction boundary.
- The chunked committed map remains the cross-transaction `BlockState.storage_writes` snapshot and is retained by the reader and callee-preload paths.

These are distinct lifetime stages. Point-four classification: this is pure convergence, not a behavioral fix. The retired single-page table was already dead: the emitted assembly census found zero legacy symbols or references, and the active chunked-map reader and execution-log path were unchanged. A committed-map hit occurs before current-transaction execution or during callee preload, where the prior committed value is correctly both original and current; same-transaction hits use the execution-log row and preserve first-touch original, so no current-to-original manufacture is reachable.

The chunked committed map remains a second source of truth alongside the execution log. Removing that remaining duplication is the follow-up Tier-2 convergence item; this PR only retires the unreachable single-page duplicate.

## Changes

- Removed the legacy snapshot append/upsert/latest-value helpers, closure links, reset globals, count/overflow globals, 16 KiB arena, probes, registry entries, manifests, and fixtures.
- Kept the chunked snapshot-upsert/latest-value helpers and their live call sites.
- Regenerated `GuestAddrs.lean`, `GuestImageEntries.lean`, and `symbol-addresses.tsv`.
- Repinned the linked layout for the current-main candidate: `.text = 0x63b44`, `.committed_storage = 0xcd9800`, `.data = 0x5370`, `.bss = 0x1c101b60`.

## Validation

- `scripts/check-region-map.sh` — pass at fixed point.
- `scripts/check-asm-to-program.sh` — clean, 388 converted definitions across 149 files.
- `scripts/check-no-warnings.sh` — full 4800-job build, 0 warnings.
- `check-unimported`, layering, forbidden-tactics, hardcoded-PC, heartbeat, opcode-structure, file-size, GuestAddrs, and coverage checks — pass.
- Emitted assembly contains no legacy table symbols or references.
- Fresh current-main A/B: baseline `7c635be8ead1c1b1a8221417ad636575882cce79` versus candidate `388aaff78ce0de6e2cb1c918c308c9501397019d`, spike, fixture tag `tests-zkevm@v0.6.2`, random 200, seed `4242`, jobs 8.
- Both legs: 200 selected, 200 ran, errors 0, budget 0, FA 0, FR 11, 189 full matches, root/success/tail matches 200/189/200.
- Comparator: 198 joined cases, identical 11-label FR sets, zero status differences, zero output-byte differences; candidate is byte-identical to baseline on all joined cases.
- Baseline ELF SHA-256: `7e34cd71c82a419bffc8f6fc46c931c75f52a9226153b548793c5a8054c912ca`.
- Candidate ELF SHA-256: `4789798292abd4868edb6cc9c8479bd3904a2f7d2add1f0ba3c3f6b6fa29e55d`.


pirapira:
Approved.

**A deletion, not a rewire — and the body now says so.** The single-page committed table was **already dead**: emitted-assembly census shows zero remaining symbols and references, and no reader was moved off it. That distinction was worth insisting on, because the two need opposite arguments. A deletion is carried by the census with a null A/B as its *predicted* proof; a rewire would need that same null *explained*, since the two sources were established as non-equivalent on four counts (value-unchanged SSTOREs recorded by one and skipped by the other, system and replay feeds missing, unsynchronised capacities at 105264 against 16384, and manufactured originals).

**The null is therefore evidence rather than reassurance.** Baseline `7c635be8e` against candidate `388aaff78`, both `FA=0 FR=11 agree=189 errors=0 budget=0`, named ELF hashes, **byte-identical on 198 joined cases**. For removing something nothing reaches, that is what should happen.

**What did not land is named, which matters more than what did.** The **chunked map remains as a second source of truth alongside the exec log**, recorded as the follow-up Tier-2 item rather than left for a reader to infer from silence that the shadow source was retired. Two containers where the spec has one is the most expensive shape we carry — it is a case split in every proof that touches storage — so the honest statement of what survives is the part that keeps the roadmap true.

**The reader-`original` question is closed alongside it**, and correctly: `bv_mtx_committed_chunked_latest_value` is called only before current-tx execution or on callee preload, so a committed-map hit materialising `original = current` is materialising a *prior-transaction* value, which genuinely **is** the transaction-start original. Same-tx SSTORE hits take the exec-log row and preserve first-touch original. No current-to-original manufacture is reachable, so this is pure convergence rather than a behavioural fix wearing a structural badge.

Structural gates are thorough: region-map fixed point, 388-function byte-tie, 4800-job build with zero warnings, plus unimported, layering, forbidden-tactics, hardcoded-PC, heartbeat, opcode-structure and file-size checks. All six checks present and `MERGEABLE`.

PR #11135, opened 2026-08-01
Avoid fabricated recipient BAL read on authorization OOG

pirapira:
## Summary

Fixes the fabricated top-level recipient account-read row on the authorization-phase OOG path (tracked in #11078). This PR is stacked on #11129; together they complete the type-4 authorization/recipient family. #11129 fixes the authorization loop processing all ten authorities where the spec halts at the first on OOG; this PR fixes the recipient row left behind on that auth-halted path. Neither fix closes the family alone.

The execution-specs ordering is the justification:

- `eoa_delegation.py:174-180` halts during authorization when the state-dependent charge runs out of gas.
- `utils/message.py:56-71` computes and records the top-level target only when `prepare_message` runs. An authorization-phase halt never reaches that message-preparation access.
- `state_tracker.py:139,199` records `get_account` reads, and `block_access_lists.py:695-697` serializes those reads into the declared BAL.

The guest previously resolved the recipient before dispatch with the tracked accessor. That accessor's internal `account_read_record` fabricated a 27-byte all-empty recipient row even when authorization OOG prevented message preparation. The fix resolves the pre-state with the raw accessor, suppresses only the internal tracked read while the pre-dispatch `account_state_latest_balance` lookup runs, and restores the ordinary post-dispatch recipient record for all messages that actually reach preparation. The suppression is a dedicated flag, not a blanket recipient-read suppression.

For the witness, direct execution-specs instrumentation of `BlockState.account_reads` found exactly nine declared addresses and no `37f536464af59c8d7358cae965f92cbeadd58dcb` recipient, confirming this is a wrong-stage guest read rather than a warm-versus-read interpretation.

The first always-suppress attempt was rejected: it caused 13 reverse flips, including `00007_test_auth_base_net_new_only...scenario_pre_tx_deleg`, which remains a required control. The final scoped version preserves that row.

## Behavioral evidence

- Witness `01786_test_transaction_validity_type_4...type_4-multiple_`: #11129 first predicted and measured `602 - (9 * 27) = 359`; this PR then predicted `359 - 27 = 332`. The final ELF reports `fail_code=0`, `shadow_status=0`, `rebuilt_len=332`, `supplied_len=332` in a complete 256-byte output.
- Control `00007_test_auth_base_net_new_only...scenario_pre_tx_deleg`: `fail_code=0`, `shadow_status=0`, `rebuilt_len=393`, `supplied_len=393` in a complete 256-byte output.
- Focused current-manifest family: all 168 `transaction_validity_type_4` rows are `OK` (`FA=0`, `FR=0`). The stale queue described only a 16-row slice; this result is the whole 168-row family, with the current manifest as the authoritative denominator.
- Paired 200, deterministic manifest `/tmp/eip7702-baseline-r200/manifest.tsv`, Spike runner `/home/zksecurity/evm-asm2/scripts/spike/spike_run`, parent #11129 ELF SHA `fbb83b69b167506518644fa38a7e73f17322c30c4cc80ba7f5ebc04c0dde54d3`, candidate ELF SHA `d26189037a6a6d1c70a28036ac454989cd11ea047f6436423819ef1cca68a720`:
  - parent: `FA=0, FR=16, OK=184, FAULT=0`
  - candidate: `FA=0, FR=15, OK=185, FAULT=0`
  - forward closure: `00075_test_intrinsic_gas_cost...valid_True-single_valid_a`
  - reverse flips: none

## Structural gates

- Fixpoint ELF: `.text=0x63ed0`, `.data=0x5370`, `.bss=0x1c101b80`; `bnf_le_a`/data base remains unchanged.
- `lake build` and `LAKE_ARTIFACT_CACHE=false lake build`: PASS.
- `scripts/check-build-parallel.sh`: PASS (including codegen/layout, asm-to-program, guarded-handler bytes, axioms, and arithmetic fuzz).
- `check-region-map.sh` and `check-asm-to-program.sh`: PASS.


pirapira:
Approved. **With #11129, this completes #11078.**

**The defect.** On an authorization-phase OOG the transaction halts before the message executes, so execution-specs never calls `get_account` on the transaction target. The guest recorded that read at pre-dispatch time regardless of whether execution proceeded, fabricating a 27-byte all-empty recipient row the declared BAL does not contain. **Wrong-stage, not warm-versus-read** — the fifth wrong-stage finding of the day.

**The proof is a run, not a citation, and that is the best methodological moment of the night.** execution-specs was instrumented directly and `BlockState.account_reads` read out: exactly nine addresses, `37f5` absent, matching the declared BAL at 332 bytes. Hours were spent tonight arguing from spec citations about what the reference implementation does — **it is executable, and when the question is what it produces for a specific input, running it settles what reading it cannot.**

**The fix is conditional where the first attempt was categorical.** The ordinary post-dispatch recipient record is left intact and skipped only when `runtime_tx_auth_phase_halted`, with a dedicated flag bracketing only the pre-dispatch `account_state_latest_balance` whose internal `account_read_record` fabricated the row. The first attempt suppressed a class and cost **13 reverse flips**, with `00007_call_to_precompile_pointer_context` as the counterexample — **that control now passes at 393/393**, which is worth as much as the witness closing.

**Two predictions, both landing on the byte.** `602 − 9×27 = 359`, measured 359 in #11129. `359 − 27 = 332`, measured 332 here. A decomposition stated in advance and confirmed twice is the strongest available evidence that a mechanism was understood rather than iterated into — and it is why the family staying red after #11129 was the *correct* intermediate result rather than a failed fix.

**Family closure is measured against the whole family, not the slice that motivated it:** 168 rows of `transaction_validity_type_4`, **OK=168, FR=0, FA=0**. We reasoned about 16 all night because that is what the manifest slice held.

**Paired 200:** parent `fbb83b69…` `FR=16 OK=184 FA=0 faults=0`; candidate `FR=15 OK=185 FA=0 faults=0`; forward closure `00075_test_intrinsic_gas_cost`, **reverse list empty**. Witness 01786 `0/0/332/332`.

Conflicting at review time; stacked on #11129 and mergebot2 owns it from here.

PR #11131, opened 2026-08-01
fix(codegen): honest fail code for dispatch-status bail (#11119)

pirapira:
## Summary

`bv_fail=36` / `.Lbv_bal_storage_omit_fail` was documented as **recipient BAL omits a storage change** (`bal_storage_covers_exec_log`). In the emitted guest the **sole reacher** was a dispatch-runtime-status bail on `.Lbv_receipts_accept`:

```
la t0, bv_dispatch_runtime_status; ld t0, 0(t0); bnez t0, .Lbv_bal_storage_omit_fail
```

So a dispatch/runtime conservative bail presented as a storage-omission rejection and sent debuggers to the wrong subsystem.

## Fix (miswire only)

In `BlockVerdictReceiptsTail.lean`:

- Retarget the bail to `.Lbv_dispatch_runtime_status_fail` with **honest code 64**.
- Leave `.Lbv_bal_storage_omit_fail` / code 36 as an unreachable named sink. The omit **property** is CONTAINER_DEPENDENT (#10765); reverse coverage stays inside allaccounts **37**. Do **not** restore a standalone omit comparator here.

## Liveness of the bail (structural, not fixture-search)

**Kind: reachable safety net; unexercised on the limit200 sample.**

From emitted assembly + source (sole writer / sole call site):

1. **Sole writer** of `bv_dispatch_runtime_status`: MTx postlude after the **only** `jal ra, dispatch_tx_runtime_code` in the guest (`BlockVerdictMtxRuntime` → `sd a0, …`). BSS init is 0; nothing else stores the cell. (Debug snapshot at `BlockVerdict.lean` only *reads* it.)

2. **Producer of nonzero**: `dispatch_tx_runtime_code` returns
   - `0` complete replay
   - `1..7` structured unsupported (code lookup / non-self-contained / BAL cap / storage proof / payload cap / staging / access-list)
   - `8` recipient unresolvable → separate immediate reject (not this bail)

3. **On `a0 ∈ 1..7`**: status is stored, then `bnez a0, .Lbv_mtx_dispatch_unsupported` — this does **not** `Lbv_zero`. It sets receipts shape 61 + `enforce_enabled=1` and continues via `.Lbv_after_tx_gas_precharge` → gas-result gate → receipts materialize/enforce → **`.Lbv_receipts_accept`**.

[...]

Issue #11119, opened 2026-08-01
bv_fail=36 storage_omit is miswired to dispatch_runtime_status, not storage omission


pirapira:
## Summary

**A check that exists and does not gate is worse than an absent one**, because its presence is why nobody wrote a second one.

`bv_fail=36` / `.Lbv_bal_storage_omit_fail` is documented and named as **recipient BAL omits a storage change execution made** (`bal_storage_covers_exec_log` / bmvmx.1.6.5). In the emitted guest it is **only** reachable from a **dispatch runtime status** bail on the receipts-accept path. It does not run a storage-omit comparator.

Parent: #10765. Sibling: five dead BAL labels (filed alongside). Related: #10680, #10684, #10614.

## Evidence (emitted guest, measured)

| | |
|---|---|
| tree | `f0aec9c9d` |
| guest ELF sha256 | `f3ef366e3ee3f9822b623710ffee577cc93d2e1e94646bc68bf80727bc3297a7` |
| method | full-text scan of `gen-out/regionmap/stateless_guest.s` for branches to `.Lbv_bal_storage_omit_fail` |

### Sole reacher

```
.Lbv_receipts_accept:
  la t0, bv_dispatch_runtime_status; ld t0, 0(t0); bnez t0, .Lbv_bal_storage_omit_fail
  li a0, 1; j .Lbv_ret
```

Source: `EvmAsm/Codegen/Programs/BlockVerdictReceiptsTail.lean` (comment at the site says dispatch_tx_runtime_code nonzero must not accept on receipts alone; the **fail label** it jumps to is still storage_omit).

Sink:

```
.Lbv_bal_storage_omit_fail:   # comment: bmvmx.1.6.5 recipient BAL omits storage change
  li t0, 36; la t1, bv_fail_code; sd t0, 0(t1); j .Lbv_zero
```

**Reacher count for `.Lbv_bal_storage_omit_fail`: 1** (the line above).  
**No** `jal` to `bal_storage_covers_exec_log` from the verdict path except inside `bal_all_accounts_storage_consistent_skip_list`, which fails as **37**, not 36.

### What the name claims vs what runs

| | claimed | actual |
|---|---|---|
| property | BAL omits an exec storage change (reverse storage ⊆ BAL) | `bv_dispatch_runtime_status ≠ 0` after receipts accept path |
| comparator | `bal_storage_covers_exec_log` | none |
[...]

PR #11136, opened 2026-08-01
chore(bal): delete five dead BAL reject labels (#11118)

### LINKED ISSUE #11118: Five granular BAL reject labels are dead: routines exist, no verdict jal
## Summary

**We believe we check things we do not check.**

Five named granular BAL reject labels have fail-code sinks and (for most) live routines in the guest image, but **zero verdict-path reachers**. Anyone reading the reject-label inventory — including #10765 classification work — will assume these properties are enforced. They are not.

This is the same class as a comment that claims a mid-execution witness walk returns absent when it does not: a **stated property that is not true**, which forecloses design and overstates coverage.

These are rejectors, not acceptors: a check that never fires cannot by itself cause a false accept. It **does** mean the properties they name are **unprotected**, and other work may be relying on that protection (digest-convergence planning, FR triage, “we already check code/reads”).

Parent: #10765. Related: #10680, #10612, #10684.

## Evidence (emitted guest, measured)

| | |
|---|---|
| tree | `f0aec9c9d` (origin/main at measure) |
| guest ELF sha256 | `f3ef366e3ee3f9822b623710ffee577cc93d2e1e94646bc68bf80727bc3297a7` |
| method | `riscv64-unknown-elf-nm` PCs + full-text scan of `gen-out/regionmap/stateless_guest.s` for `jal …, <sym>` and branch-to-fail-label |

### Dead fail labels (0 branch reachers in entire `.s`)

| label | fail_code | sink in `BlockVerdictReceiptsTail` | property routine | nm PC | `jal` count from verdict / whole guest |
|---|---|---|---|---|---|
| `storage_mismatch` | **34** | `.Lbv_bal_storage_mismatch_fail` | `bal_storage_matches_exec_log` | `0x8002608c` | **0 from verdict**. Only jal is inside `bal_all_accounts_storage_consistent_skip_list` (`.Lc2baas_check`); that wrapper fails as **37**, not 34. |
| `recipient_field` | **35** | `.Lbv_bal_recipient_field_fail` | (historical recipient nonce/code claim check) | — | **0** reachers of fail label; no dedicated jal found |
| `reads` | **38** | `.Lbv_bal_reads_fail` | `bal_storage_reads_in_exec_log` | `0x80027ff8` | **0** jals anywhere in guest |
| `code_covers` | **43** | `.Lbv_bal_code_covers_fail` | `bal_all_accounts_code_covers` | `0x8002c4d0` | **0** jals anywhere in guest |
| `code_consistent` | **46** | `.Lbv_bal_code_consistent_fail` | `bal_all_accounts_code_consistent` | `0x8002c5fc` | **0** jals anywhere in guest. (Internal `jal x1, bal_account_code_consistent` @ `0x8002c9bc` exists only as callee of the dead wrapper.) |

Contrast — labels that **do** have verdict reachers (same scan):

| label | code | reacher |
|---|---|---|
| `gas` | 7 | `bal_gas_valid` → `.Lbv_bal_gas_fail` |
| `allaccounts` | 37 | `bal_all_accounts_storage_consistent_skip_list` |
| `tuple` | 42 | `bal_all_accounts_tuple_sequences_consistent_skip_list` |
| `nonstorage` | 44 | all-accounts nonstorage + `block_verdict_withdrawal_nonstorage_effects` |
| `nonstorage_covers` | 45 | `bal_all_accounts_nonstorage_covers` |

### What still runs for storage forward/reverse

`bal_storage_matches_exec_log` / `bal_storage_covers_exec_log` **do** run, but only as callees of the all-accounts wrapper. Failures surface as **bv_fail=37**, never 34/36. So “storage_mismatch” and “storage_omit” as **named labels/codes** are not the live path; allaccounts is.

Code and reads have **no** live path at all.

## Consequence

1. Reject-label set / dashboards / mental model **overstate coverage** by five properties.
2. #10680 / #10765 “which checks may the digest retire” must not treat 34/35/38/43/46 as live gates.
3. Unprotected properties today:
   - **reads:** declared `storage_reads` not checked ⊆ guest block read set on the verdict path.
   - **code forward/reverse:** declared code_changes vs exec code-effect, and exec deploy/clear coverage, not checked on the verdict path.
   - **recipient_field:** BAL nonce/code claim without execution change not checked under code 35.
   - **storage_mismatch as code 34:** never attributed; folded into 37 only for non-skip accounts.

## Suggested follow-ups (not done here)

- Either **wire** each property to the verdict path (expect FRs; exposure is progress) or **delete** the dead sink + unlink the orphan routine once #10765 class says it is safe to wait on container convergence.
- Per #10765 classification: `reads` / `recipient_field` (BAL claim) are SUBSUMED-shaped once rebuild is complete; `code_*` and storage halves are CONTAINER_DEPENDENT. Deadness is independent of that class — dead checks protect nothing either way.
- Sibling: storage_omit code 36 is **miswired** (not merely dead) — separate issue.

## Refs

- Classification comment on #10765 (roadmap + live/dead table).
- Source sinks: `EvmAsm/Codegen/Programs/BlockVerdictReceiptsTail.lean` (`.Lbv_bal_*_fail`).
- Allaccounts wrapper only caller of storage match/cover: `BalAllAccountsStorage.lean` / `.Lc2baas_check`.

pirapira:
## Summary

Delete five dead BAL reject labels from the guest image (#11118).

**#10612 obligation (top):** execution-specs builds one BAL and checks
`gas_ok(built) ∧ keccak(rlp(built)) == keccak(rlp(provided))`. It never
field-inspects provided BAL. Granular labels are temporary scaffold around
split containers; dead ones go now.

### What goes

| code | sink / routine | action |
|---|---|---|
| **34** storage_mismatch | sink only | DELETE sink/label. **KEEP** `bal_storage_matches_exec_log` (live callee of allaccounts **37**). Asymmetric. |
| **35** recipient_field | sink only | DELETE sink. Code 35 remains live on `.Lbv_block_state_gas_fail` (unrelated). |
| **38** reads | `bal_storage_reads_in_exec_log` + sink | UNLINK + DELETE sink. Wiring it would re-split to the exec log (moves away from #10612 convergence) — stronger than "subsumed by digest". |
| **43** code_covers | routine + sink | UNLINK + DELETE |
| **46** code_consistent | routines + sink | UNLINK + DELETE |

### What stays

- `bal_storage_matches_exec_log` + `bal_storage_covers_exec_log` — still `jal`'d from `bal_all_accounts_storage_consistent_skip_list` (37).
- `bacc_finals` scratch — still used live by `bal_same_block_delegation_code_resolve` and BlockVerdictStateRoot (not only by unlinked code_consistent).
- storage_omit **36** sink — still reached via dispatch_runtime_status miswire (#11119 / PR #11131). This PR does **not** retarget that miswire.

Probe-only PC placeholders for unlinked Function defs (#11103 style).

### Layout

Fixpoint regen: `textSizeBytes` `0x063d2c → 0x0637f0`, `bssSizeBytes` `-0x40`. GuestAddrs / GuestImageEntries / symbol-addresses.tsv regenerated.

### Emitting census (post-cut)

- Dead text symbols absent from guest `.s` / nm: `bal_storage_reads_in_exec_log`, `bal_all_accounts_code_covers`, `bal_all_accounts_code_consistent`, `bal_account_code_consistent`.
- Dead BAL sinks absent: `.Lbv_bal_storage_mismatch`, recipient_field, reads, code_covers, code_consistent.
- **34 callee confirm:** allaccounts body still `jal ra, bal_storage_matches_exec_log` and `jal ra, bal_storage_covers_exec_log`.

### FR quadruple (limit200 spike)

| leg | commit / guest | ran | full | fail | err | budget |
|---|---|---|---|---|---|---|
| main | `f5bd52b6d` / `04a1c776…d86110` | 200 | 174 | 26 | 0 | 0 |
| fix | `fbc2d055b` / `1033ee2f…73865de` | 200 | 174 | 26 | 0 | 0 |

- Runs: fix `run-20260801T233051Z-2917406`, main `run-20260801T233542Z-2980284` (`--guest-elf`)
- **actual_hex / `.output` byte-identical 200/200**
- **reverse flips: 0**

Null A/B as predicted (dead code deletion).

Closes #11118.


pirapira:
Approved.

**Five reject labels whose routines exist and which no verdict `jal` ever reaches** — `storage_mismatch34`, `recipient_field35`, `reads38`, `code_covers43`, `code_consistent46`. So the guest has been advertising five checks that never run. I reasoned from that label list myself last night, which is exactly the cost of leaving it.

**The disposition was argued per label rather than as a blanket call**, and two are worth recording:

* **34's routine is LIVE** under `allaccounts37` as a callee — only its standalone sink and label are dead. Deleting the routine would have cost `37` a callee and turned a false-advertising fix into a real regression. Confirmed in emitted assembly after the cut: `bal_all_accounts_storage_consistent_skip_list` still `jal`s both `bal_storage_matches_exec_log` and `bal_storage_covers_exec_log`. That check was a grep, not a run — and a null A/B would not necessarily have caught the regression, since no sampled row need exercise that callee.
* **38 is the strongest argument in the set**: wiring `reads` would **re-split to the exec log**, which is the container divergence we are trying to remove. So wiring it would not merely duplicate the digest, it would move us *away* from convergence. That reasoning is what stops someone re-adding it in six months.

**The umbrella argument is #10612's:** the spec obligation is gas-ok plus hash equality, and execution-specs never field-inspects the provided BAL. Five checks that never run, whose properties digest agreement already covers, are five things a correspondence proof would otherwise have to account for. That is the proof-cost rationale in its cleanest form — the payoff here is shorter proofs, not fewer lines.

**Evidence:** `limit 200` at 174/26 on both legs, `.output` identical 200/200, reverse flips zero, `.text` `0x63d2c → 0x637f0`. For removing code nothing reaches, a null is the predicted proof rather than an ambiguity.

pirapira:
Excluded from automated batching: this branch conflicts with main in `EvmAsm/Codegen/Programs/BlockVerdictReceiptsTail.lean`, and the conflict is semantic rather than mechanical.

This PR removes fail-code sinks 34 (`storage_mismatch`), 35 (`recipient_field`) and 38 (`reads`) as dead. Meanwhile main — via #11119 / #11131 — retargeted the `storage_omit(36)` miswire onto a new `.Lbv_dispatch_runtime_status_fail` (code 64) and rewrote the reachability commentary around 36. Deciding which sinks remain genuinely unreachable after that retarget is a correctness judgement about the verdict path, so it needs a human resolve rather than an automated one.

(The accompanying `GuestImage.lean` conflict is just the `.bss` extent literals and is mechanical — it resolves itself once the above is settled and the layout is regenerated.)

Still carries `merge!`; will be picked up once resolved by hand.

pirapira:
## complicated-merge resolved

Merged `origin/main` (`703336c28`) + layout regen fixpoint (no hand-arithmetic on sizes).

### Deadness re-confirm on main before resolve
Five labels still unreferenced on main guest path:
- sinks 34/35-recipient/38/43/46 present but 0 verdict jal
- routines `bal_storage_reads_in_exec_log`, `bal_all_accounts_code_*`, `bal_account_code_consistent` still in aggregates only
- **KEEP** `bal_storage_matches_exec_log` + `bal_storage_covers_exec_log` still jal'd from allaccounts 37

### Merge notes
- Took main on generated files then regen fixpoint
- Combined #11118 sink deletions with main #11119 `dispatch_runtime_status_fail` code **64**
- omit36 sink retained unreachable (CONTAINER_DEPENDENT)

### Layout (ELF-measured)
- textSizeBytes `0x0638f0`, bssSizeBytes `0x1c101b40`, GuestImage end `0xbf211b40`
- check-region-map clean
- Guest sha256 `0dc661ba613841045b6ac14101ac28b0b97482bbac84478990722ff277c4e027`

### FR limit200 vs main `703336c28` guest `c16ff50d…`
| leg | run | full/fail |
|---|---|---|
| main | run-20260802T075355Z-1233156 | 186/14 |
| fix | run-20260802T075150Z-1215926 | 186/14 |

eest-ab-compare: FA0→0 FR14→14 label-set EQUAL **byte-identical** joined 194. Reverse flips 0.

Removed `complicated-merge`; left `merge!`.

pirapira:
Update: the `BlockVerdictReceiptsTail.lean` conflict is **resolved** — this branch now merges main cleanly. Thanks.

It is still blocked, but on a different and more concrete failure. Merging main and running the local gate set, `check-asm-to-program` fails:

```
ConvError: unresolved branch/jump target 'bal_account_code_consistent'
```

**Diagnosis:** this PR unlinks the routine — `GuestAddrs.bal_account_code_consistent` disappears from the regenerated `GuestAddrs.lean`, and the caller in `bal_all_accounts_code_consistent` had its `.JAL` removed — but the asm string in `EvmAsm/Codegen/Programs/BalAccountCodeConsistent.lean` still contains a `jal` to that label, so the converter cannot resolve it. `zisk_bal_account_code_consistent` also remains registered in `RegistryNamesTail.lean`.

Either the label needs to stay linked, or that remaining jump (and probably the probe-unit registration) needs to go with it. That is a judgement about what stays reachable, so I have not changed it.

Everything else passes on the merged tree: `lake build` plus the other 14 gates, including `check-build-units-link` and `check-region-map`.

One thing I did have to fix locally to get that far, worth folding into the branch: `GuestImage.lean` `t7` is internally inconsistent — its end bound is `0xbf211b40` (matching the new `.bss` `0x1c101b40`) but its size literal is still `0x1c101b80`. That mismatch produces the bare `Type mismatch` the file's own comment warns about. Line 221 needs `0x1c101b40`.

pirapira:
Re-checked against current main (post-#11142). Merge conflicts are gone and everything is mechanically repairable **except one gate**, which I traced to a concrete root cause.

`check-asm-to-program` fails:
```
ConvError: unresolved branch/jump target 'bal_account_code_consistent'
```
Culprit file: `EvmAsm/Codegen/Programs/BalAllAccountsCode.lean`, at its reloc side-table:
```lean
def balAllAccountsCodeConsistent_relocs : RelocTable :=
  [ ... (61, .jal .x1 "bal_account_code_consistent") ]
```

**Why it fails:** `scripts/asm_to_program.py` resolves cross-function `jal` targets through `SYMMAP`, which is loaded from the linked guest's symbol table (`asm_to_program.py:382`). A target must be either a *local label* or a *linked guest symbol* — there is no third category. This PR unlinks `bal_account_code_consistent`, so it leaves `SYMMAP`, while `bal_all_accounts_code_consistent` (now probe-only itself) still names it as a cross-function external.

The PR already migrated the other references to PC placeholders (`balAccountCodeConsistentPc`, `balAllAccountsCodeConsistentPc`); this reloc entry is the one that still names the unlinked symbol.

I have not attempted a fix: the options (make the callee a local label by co-emitting both in one probe blob, teach the converter a probe-only external category, or exclude the now-probe-only file from `check-all`) are design choices about the converter's model, not mechanical repair.

Everything else on the merged tree is green — `lake build` plus the other 14 gates including `check-build-units-link` and `check-region-map`, with `.text` -> `0x063644`, `.bss` -> `0x1c101b20` and the `GuestImage` `.bss` literals repinned to match.

pirapira:
Re-labelled `complicated-merge` (the label had been dropped) and re-verified against current main (`9bac6a15a`).

**Status: the merge conflicts are trivial; the blocker is a gate failure.**

`DIRTY` here is only the usual generated layout files (`GuestAddrs.lean`, `RegionMap.lean`, `symbol-addresses.tsv`) — those resolve mechanically. With main merged in, the real failure is `check-asm-to-program`:

```
ConvError: unresolved branch/jump target 'bal_account_code_consistent'
```

Culprit: `EvmAsm/Codegen/Programs/BalAllAccountsCode.lean:160`
```lean
(61, .jal .x1 "bal_account_code_consistent") ]
```

**Mechanism.** `scripts/asm_to_program.py` resolves cross-function `jal` targets through `SYMMAP`, built from the linked guest's symbol table (`asm_to_program.py:382`). A target must be either a *local label* or a *linked guest symbol* — there is no third category. This PR unlinks `bal_account_code_consistent` (confirmed absent from `symbol-addresses.tsv`), while `bal_all_accounts_code_consistent` — now probe-only itself — still names it as a cross-function external in its reloc side-table.

The PR already migrated its other references to PC placeholders (`balAccountCodeConsistentPc`, `balAllAccountsCodeConsistentPc`); this reloc entry is the one left naming the unlinked symbol.

**Options** (a design choice about the converter's model, hence not something I applied):
1. Co-emit both routines in one probe blob so the target resolves as a local label (the `tgt in label_addr` path).
2. Give the converter a probe-only external category.
3. Exclude the now-probe-only file from `check-all`.

**Everything else is green** on the merged tree: `lake build` plus the other 14 gates including `check-build-units-link` and `check-region-map`, with `.text` -> `0x063644`, `.bss` -> `0x1c101b20`, and the handwritten `GuestImage.lean` `.bss` extent/end literals repinned to `0x1c101b20` / `0xbf211b20`.

Say which option you want and I'll do the rest of the merge.

pirapira:
**Option 1 — co-emit. And it is not a design choice: the pattern already exists in the file you're editing.**

`BalAccountCodeConsistent.lean:118-132` defines the sibling probe `zisk_bal_account_code_consistent`, which calls `jal ra, bal_account_code_consistent` at :125 and then **co-emits the callee into the same blob** at :129 — followed by its transitive closure, `balAccountNonstorageFinalsFunction` (:130) and `rlpWalkHelpersClosure` (:131).

That probe resolves today for exactly the reason yours doesn't: its target is a **local label** in its own blob. So the fix is to apply the existing convention one file over, not to extend the converter's model.

**Why not the other two — both weaken a gate to land a deletion.**

- **Option 2 (probe-only external category)** creates a class of `jal` target that resolves to nothing. That's not scoped to this file; it becomes a place where a *genuinely* unresolved target can land silently, image-wide. The check's whole value is that there are only two categories and no third.
- **Option 3 (exclude from `check-all`)** removes coverage from a file we intend to keep. Future drift in it then goes unobserved, and the exclusion will outlive everyone's memory of why it was added.

Both trade a permanent reduction in checking for a one-time merge convenience. Option 1 costs a few lines and leaves the invariant — *every target is a local label or a linked symbol* — fully intact.

⚠️ **One thing that will bite if missed: co-emit the transitive closure, not just the direct callee.** The sibling probe pulls in three things, not one. If `bal_account_code_consistent` is emitted alone, its own `jal` to `bal_account_nonstorage_finals` (see the reloc at :57) becomes the next unresolved target and the failure simply moves one level down — which reads as the fix not working rather than as being incomplete. Mirror all three lines from :129-131.

**Your diagnosis was right and the handback was right.** `SYMMAP` is built from the linked guest's symbol table (`asm_to_program.py:382`), the PR unlinks the symbol, and the one surviving reloc still names it as a cross-function external while its siblings were migrated to PC placeholders. That is a complete root cause, and stopping at a converter-model question rather than guessing was correct.

Sorry for the wait — you weren't stalled, you were blocked on me, and I read the silence as the former. The `complicated-merge` re-label is what surfaced it; thank you for restoring it after it was dropped.

Go ahead and finish the merge with option 1.

pirapira:
## Handoff: precise state, for whoever picks this up

**TL;DR — the co-emit is already done and is not the problem. One reloc-table line is.**

### What is already correct
`ziskBalAllAccountsCodeConsistentPrologue` (`BalAllAccountsCode.lean:182`) already co-emits the callee **and its transitive closure** at :193-201:
```lean
balAllAccountsCodeConsistentFunction ++ "\n" ++
balAccountCodeConsistentFunction   ++ "\n" ++   -- direct callee
balAccountNonstorageFinalsFunction ++ "\n" ++   -- transitive
rlpWalkHelpersClosure              ++ "\n" ++   -- transitive
rlpListNthItemFunction, rlpListCountItemsFunction,
rlpFieldToU256BeFunction, rlpFieldToU64Function
```
That is a superset of the sibling probe's three lines, and it works: **`check-build-units-link` passes** — the probe ELF links because the target is local to its blob. So applying the co-emit convention is not the remaining work.

### What actually fails
`check-asm-to-program`, which converts a *different string*: `balAllAccountsCodeConsistentFunction` **in isolation** (the 63-line `emitProgramR prog relocs` blob, entry label `bal_all_accounts_code_consistent:`). In that blob the callee is not local, so `SYMMAP` is consulted — and the symbol is gone.

```
ConvError: unresolved branch/jump target 'bal_account_code_consistent'
```

Verified symbol state on the merged tree (`symbol-addresses.tsv`):

| symbol | linked |
|---|---|
| `rlp_walk_init` / `rlp_walk_next` | yes -> resolve fine |
| `bal_account_nonstorage_finals` | yes |
| `bal_account_code_consistent` | **no** (unlinked by this PR) |
| `bal_all_accounts_code_consistent` | **no** |

It is specifically an *unlinked -> unlinked* call that the converter cannot express.

### The one line still naming the symbol
`BalAllAccountsCode.lean:159`
```lean
(61, .jal .x1 "bal_account_code_consistent") ]
```
The Program itself was already migrated — :131 reads `.JAL .x1 (jalOff balAccountCodeConsistentPc (balAllAccountsCodeConsistentPc + 244))`. Only the reloc, whose job is to keep that `jal` **symbolic in the emitted text**, still points at the unlinked name.

### Why I did not just append to the Function
`balAllAccountsCodeConsistentFunction` is pinned by the kernel-checked drift theorem `balAllAccountsCodeConsistentFunction_eq_prog` (:170) to be exactly `emitProgramR prog relocs`. Appending bodies to make the label local would falsify it.

### Candidate fix (unverified — needs the author's call)
Drop reloc entry `(61, ...)` so the emitted text carries the concrete PC-placeholder offset instead of a symbolic reference, matching what :131 already does. The open question is whether that concrete offset is valid in the probe's real layout given the placeholder PCs (`balAllAccountsCodeConsistentPc = 0x80000000`, `balAccountCodeConsistentPc = 0x80002000`) — that changes emitted bytes for the probe, so I have not applied it.

### Everything else is ready
On the merged tree (main `2de390f00`): `lake build` plus the other **14** gates green, including `check-build-units-link` and `check-region-map`. Layout repinned to `.text = 0x063bec`, `.bss = 0x1c101b20`, with `GuestImage.lean` extent/end literals at `0x1c101b20` / `0xbf211b20`. `bnf_le_a = 0xa3000ee0` intact.

Fix that one reloc line and the merge is otherwise mechanical — I can finish it immediately.

pirapira:
**Two corrections to my previous comment, and a warning about the candidate fix.**

## 1. My option-1 answer was wrong

I said "co-emit — the pattern already exists in the sibling probe." You then established that `ziskBalAllAccountsCodeConsistentPrologue` **already co-emits the callee and a superset of its transitive closure**, and that `check-build-units-link` passes as a result. I read `BalAccountCodeConsistent.lean` and prescribed from it without checking whether the same convention was already applied in the file being edited. It was. The advice was work already done.

## 2. ⛔ The candidate fix would silently break the probe — do not apply it

Dropping reloc `(61, …)` so the emitted text carries the concrete offset **produces a wrong jump**, because the placeholder PCs are not the probe's real layout:

- The `jal` sits at instruction index **61** → byte offset `61 × 4 = 244`
- `balAllAccountsCodeConsistent_prog.length = 81` → the function body is `81 × 4 = 324` bytes
- In the probe blob, `bal_account_code_consistent:` begins immediately after, at **+324**
- So the true displacement is `324 − 244 =` **80 bytes**

But the placeholder arithmetic at :131 yields:

```
jalOff 0x80002000 (0x80000000 + 244) = 8192 − 244 = 7948
```

**7948 versus 80.** Emitting the concrete offset would jump roughly 7.8 KB past the callee, into whatever follows in the probe image.

⚠️ **And it would pass every gate you have.** No relocation is required, so the link succeeds; `check-asm-to-program` stops complaining because the symbolic name is gone; the drift theorem still holds because it compares against the same `prog`/`relocs` pair. The failure is only observable by *running* the probe. Your instinct to flag it as unverified rather than apply it was right, and the arithmetic confirms it.

The placeholder PCs exist for **verification of the guest-linked Program**, where the two routines really are 0x2000 apart. They are fiction with respect to the probe's layout, where the callee is adjacent.

## 3. My objection to option 3 was also wrong

I rejected "exclude from `check-all`" as removing coverage. That was based on a false premise: `balAllAccountsCodeConsistentFunction_eq_prog` (`:170`) is a **kernel-checked `rfl` drift guard** pinning the emitted string to `emitProgramR prog relocs`. Drift in this function is caught by the theorem whether or not the converter runs.

And there's a domain argument I missed. `SYMMAP` is built from **the linked guest's symbol table** (`asm_to_program.py:382`). `bal_all_accounts_code_consistent` is now **probe-only** — it is not in the guest image. Resolving it against a linked-guest symbol map is applying the gate outside its domain, which is exactly why an *unlinked → unlinked* call has no expressible form. That isn't a gap in the converter's model; it's the converter being asked about something it doesn't cover.

⇒ So the principled fix is to **exclude the now-probe-only function from the SYMMAP-based conversion**, with the drift theorem retained as the guard — not to weaken the converter, and not to emit a fictional offset.

**What I'd want in the change**: a comment at the exclusion site saying *why* (probe-only, not in the linked image, drift covered by `balAllAccountsCodeConsistentFunction_eq_prog`), so the exclusion doesn't outlive its reason — which was my original and still-valid worry about option 3.

Routing this to the PR author rather than asking you to apply it: the choice changes what the converter is understood to cover, and the 7948-vs-80 result means the obvious alternative is unsafe. Your diagnosis, the symbol table, and the layout numbers were all correct — thank you for not applying the fix you weren't sure of.

pirapira:
**Fix verified. `complicated-merge` correctly removed; `merge!` stands — merge bot can finish.**

The fix is better than what I specified. I asked for the probe-only function to be excluded from SYMMAP conversion. Instead it is keyed on the **domain condition itself** — entry not in `SYMMAP` — extending the skip that already existed, rather than naming a file. That generalises correctly: any probe-only function is outside the linked-image domain by construction, and no list needs maintaining as more routines become probe-only.

**Crucially it is not the "third category" I rejected earlier.** My objection to a probe-only *external* category was that it would let a genuinely unresolved target land silently image-wide. This narrows on the **entry**, not the target: for a function that *is* in the linked image, an unresolved `jal` still raises `ConvError` exactly as before. The relaxation reaches only functions that were never in the image to begin with.

**The enumeration found a second instance, which is why it was worth doing first.** Not just `bal_account_code_consistent` but `bsr_krev`; two failing entries (`balAllAccountsCodeConsistentFunction`, `balStorageReadsInExecLogFunction`) and two probe-only entries with linked callees that still convert normally. Patching the first symptom would have surfaced the second immediately afterwards, reading as the fix not working rather than as a second case.

**Drift coverage — I verified this rather than taking it on trust**, since skipping concrete consistency and `gen_lean` source-drift is a real reduction for the affected entries:

| entry | drift guard |
|---|---|
| `balAllAccountsCodeConsistentFunction` | `…_eq_prog` at `BalAllAccountsCode.lean:165`, `rfl` |
| `balStorageReadsInExecLogFunction` | `…_eq_prog` at `BalStorageReadsExecLog.lean:178`, `rfl` |

Both are kernel-checked and pin the emitted string to `emitProgramR prog relocs`. So drift in these functions is caught by the Lean build whether or not the converter runs — the check moves, it does not disappear. The symbolic assemble safety check still runs on top of that.

**And reloc 61 was left alone.** Dropping it would have emitted a concrete offset of **7948** where the probe's real displacement is **80** bytes — a jump ~7.8 KB past the callee that passes the link, silences the converter, and satisfies the drift theorem, failing only when the probe actually runs. That was the available shortcut and it was declined twice: once by the merge bot flagging it unverified, once here.

---

For the record, since this PR cost more round trips than it should have: **two of my earlier answers were wrong.** I prescribed co-emitting when the file already co-emitted, and I rejected the exclusion route without knowing a kernel-checked drift theorem already covered the function. Both were corrected by people checking rather than complying.

Issue #11133, opened 2026-08-01
seed_callee_storage has no execution-specs counterpart: its 128-entry table cap bounds a guest-only staging mechanism, and the derived 100,000 sibling is not the right comparator


pirapira:
Two caps bound the same quantity — how many storage slots of one account the BAL-driven seeding will
stage — and they disagree by a factor of ~780. One is derived from the 200M block-gas bound the way every
other arena here is sized; the other is a hard-coded round number.

⚠️ **This is a consistency/sizing issue, not a demonstrated defect.** I measured the obvious wrong-value
consequence and it did **not** occur — see "What was measured" below. Filing it because the discrepancy
itself is the hazard shape this project sizes against, and because a hard-coded bound next to a derived
sibling is a standing proof obligation.

## The two caps

| cap | value | where | basis |
|---|---|---|---|
| `bsrAccountSlotCap` | **100,000** = `bsrStateRootBlockGasLimit / bsrBalGasCost` = `200000000 / 2000` | `BlockVerdictParams.lean:124` (via `bsrMaxBalItems`, `:34`) | **derived** from the 200M block-gas bound |
| the `seed_callee_storage` cap | **128**, hard-coded `li t0, 128` | `BlockVerdictDispatchTx.lean:159` (changes) and `:182` (changes + reads) | **a round number** |

Both gate BAL-derived per-account storage seeding. `:182` is the combined test — on overflow it takes
`.Lscs_acct_next` and **skips the account entirely**.

## Against the project's own sizing rule

Sizing here is derived from what a 200M-gas block can reach, not from the corpus. The cheapest way to touch
distinct storage slots on one account is a cold `SLOAD` at `COLD_SLOAD_COST` = 2100 gas:

```
200,000,000 / 2100  ≈  95,238 distinct slots on a single account
```

⇒ The derived cap (100,000) covers that. **The hard-coded cap (128) is ~744× below it.**

## What was measured

Full corpus scan, `tests-zkevm@v0.6.2`, 6334 fixtures, 227,061 accounts with storage:

| | |
|---|---|
| max distinct slots on one account (pre ∪ post) | **609** |
| accounts over 128 (pre ∪ post) | **54** |
| accounts over 128 **in the pre-state** | **10**, across 4 fixture files |

And the cap **does fire on a real corpus row**. `maximum_gas_refund` (both variants), converted with
`--verify-input-parity`, run on the unmodified guest:

| | count |
|---|---|
[...]

PR #11138, opened 2026-08-02
feat(codegen): distinguish an absent slot from a witness-integrity failure at tier 3 (#10874 prerequisite)

pirapira:
Prerequisite for #10874. **Behaviour is unchanged** — both arms still fall to `_prestate_zero` — so this
records the distinction without deciding what to do with it.

⚠️ **This deliberately does not decide policy.** What the guest should *do* on a witness-integrity failure —
reject, fall back, or bail — is a soundness question and is not settled here. The counters have no consumer
yet; the point is to make the next step measurable rather than to invent a policy.

## Why this comes before #10874

Today a broken authenticated read is **masked by the preload**: tier 1 hits and the fallback is barely
exercised. The moment #10874 retires the preload the demand-driven path becomes the **only** source, and a
read that fails for *any* reason returns zero and looks exactly like an absent slot. That is a false-accept
shape, and it would arrive the day the retirement lands rather than being found later.

## The status codes — verified, not assumed

From the documented calling conventions: `mpt_lookup_by_key` returns `0 found / 1 not found / 2 parse error`;
`slot_at_index` returns that, or `3` on decode failure; `slot_at_header_state_root` adds **4** to the
slot-walk status.

| status | meaning | spec-legitimate zero? |
|---|---|---|
| **5** | **absent leaf** (`mpt` not-found 1, +4) | ✅ yes — this is the only one |
| 6 | witness parse error (`mpt` parse error 2, +4) | ⛔ no |
| 7 | `slot_decode_u256` failure (3, +4) | ⛔ no |
| 4 | `header_extract_state_root` failure | ⛔ no |

⚠️ **The framing worth keeping:** execution-specs' `get_storage` returns zero for an **absent leaf** and has
**no error path that returns zero**, because its witness is authenticated by construction. So there is no
spec shape to copy here — **we have a failure mode the spec does not have**, and the convergence question is
whether we should be able to have it at all. That is a better question than "add a status" and probably a
maintainer one.

`seed_callee_storage` **already** uses this same `status == 5` convention for its absent case, so this brings
one site into line with an existing sibling rather than inventing a scheme.

## Measured on the emitted guest

`sstore_xto_y` d0-g0 / d0-g1 / d4-g1, counters read as PCs from the objdump:

| | d0-g0 | d0-g1 | d4-g1 |
|---|---|---|---|
| tier-3 entries | 17 | 7 | 17 |
| SUCCESS | 1 | 1 | 1 |
| **ABSENT (status 5)** | **16** | **6** | **16** |
| **INTEGRITY FAIL** | **0** | **0** | **0** |

`entries == success + absent + integrity` **exactly** on all three rows — the partition closes with nothing
left over. And `absent` firing 16/6/16 proves the branch is **live**, so the zero integrity count is a real
zero rather than a dead instrument.

⇒ **No witness-integrity failure occurs in production today.** That is the sequencing argument, measured
rather than assumed: the indistinguishability is currently harmless, and it stops being harmless when the
preload goes.

## Behaviour gate

Byte-identity is impossible (`.text` grows), so the gate is behavioural — and run on the population that
actually reaches the changed code rather than a random sample that mostly does not:

| | |
|---|---|
| rows exercising tier 3 | 12 |
| **identical output bytes, old ELF vs new** | **12 / 12** |
| differing | **0** |

Both arms of the new branch converge on the same label, so this is also unchanged *by construction*.

| pin | before | after |
|---|---|---|
| `RegionMap.textSizeBytes` | `0x063d98` | **`0x063dd8`** |
| `RegionMap.dataSizeBytes` | `0x5370` | **unchanged** |
| `RegionMap.bssSizeBytes` | `0x1c101b60` | **unchanged** |

`check-region-map` reached **fixpoint in one iteration**. `.data` is unperturbed because the three new cells
are **zero-only**, so `Layout.moveZeroDataLines` promotes them to `.bss` and the fixed `.data` addresses do
not move — the mechanism is documented in `BlockVerdictDataSectionTail.lean` and confirmed by the measured
`dataSizeBytes`. `symbol-addresses.tsv` regenerated (3217 rows); `GuestAddrs.lean` 1103 symbols;
`GuestImageEntries.lean` regenerated and unchanged.

`lake build` clean · `check-forbidden-tactics` OK · both touched files well under the 1500-line cap.

Refs #10874, #11133.


pirapira:
Approved.

**A failure mode the spec does not have.** `WitnessState.get_storage` returns zero when the **leaf is absent** — it has no error path returning zero, because the spec's witness is authenticated by construction. Our tier 3 flattens *every* non-zero status to `value = 0`, so a witness-integrity failure and a legitimately absent slot become the same answer. This records the distinction at the point it is produced.

**Why it had to come before #10874 rather than after.** Today a broken read is **masked by the preload**: tier 1 hits, the fallback is barely exercised, and the indistinguishability costs nothing. The moment the BAL-sourced preload is retired, the demand-driven path becomes the **only** source — and a read that fails for any reason returns zero and looks exactly like an absent slot. That is a false-accept shape arriving on the day the retirement lands. Landing this first converts that retirement from a change that removes a safety net into one that does not need it.

**Scope held exactly.** The distinction is recorded and **nothing consumes it** — no policy decided about whether the guest should reject, fall back or bail on a witness-integrity failure. That is a maintainer decision and a separate PR, and stopping at the boundary is what keeps this reviewable at this size.

**Behaviour-unchanged is proved two ways, and the structural one is the stronger.** Both arms of the new branch converge on the same label, so callers **cannot** observe a difference — that does not decay when someone adds a caller. Empirically, 12 of 12 byte-identical outputs against main's own ELF **on the rows that actually exercise tier 3**, which is a better-chosen population than a generic 200: a sample that does not reach the changed code cannot tell you anything about it, and this project spent several hours tonight learning exactly that.

All six checks present, `MERGEABLE`.

Next in this area, and now genuinely reachable: #10874 itself. The read has been shown to return correct values in production, the three claims that recorded it as blocked are retracted, and a broken read will no longer be silently indistinguishable from an empty one.

Issue #11120, opened 2026-08-01
bal_gas_valid walks declared BAL; spec gas-validates built (Path C)


pirapira:
## Summary

`bal_gas_valid` counts items on the **declared** payload BAL (`bv_bal_start` / `bv_bal_len`). execution-specs `validate_block_access_list_gas_limit` runs on the **built** list after `build_block_access_list` (`fork.py:932-936` at e5a8caf1b). Guest Path C ≠ spec input.

Related roadmap: #10612 (three-path decomposition A/B/C).

## Spec (file:line)

- Build BAL from one `BlockAccessListBuilder`, then gas-validate **built**: `fork.py:932-936`
- Hash obligation is separately `hash_block_access_list(built) == header.block_access_list_hash` (`:366-391`)

## Guest (emitted)

| step | site | input |
|---|---|---|
| gas | `BlockVerdictFunction.lean` → `jal bal_gas_valid` with `bv_bal_start`/`bv_bal_len` from SSZ payload slice | **declared** |
| after gas pass | `bv_bal_shadow_ready := 1` | reach marker |
| digest bind | `BlockVerdictReceiptsTail` Path A: rebuild from `bal_builder_*`, hash supplied SSZ bytes, compare; ACCEPT-only → bv 60/61 | guest-built vs declared |

Guest sha (tree at analysis): `f3ef366e…297a7` / tree `f0aec9c9d…`.

## Soundness: is gas-on-declared safe given Path A?

**Yes — cosmetic convergence, not false-accept shaped — on every live accepting path.**

### Accepting paths (enumeration)

Guest success bit comes only from `stateless_verdict_v2` → `block_verdict` → `a0` stamped at OUTPUT[32] (`StatelessGuestEpilogue`).

Inside `block_verdict`, the **only live** terminal accept is:

```
.Lbv_receipts_accept:
  … dispatch_runtime_status gate …
  li a0, 1; j .Lbv_ret
```

(`.Lbv_receipts_no_runtime_gas` also ends `li a0, 1; j .Lbv_ret` but has **zero branch targets** in the emitted guest — dead fallthrough arm, never entered.)

`.Lbv_ret` always:
1. runs `bal_serializer_verify` if `bv_bal_shadow_ready ≠ 0`
2. **binds** Path A into the verdict only when `a0 == 1` **and** `shadow_ready == 1` (else skips bind)

### `shadow_ready` and gas order

[...]

PR #11139, opened 2026-08-02
fix(bal): gas-validate built BAL at Lbv_ret (#11120)

pirapira:
## Summary

Align EIP-7928 BAL gas-limit validation with execution-specs: validate the
**built** list after rebuild, not the declared SSZ bytes early in `block_verdict`.

Spec pin (`execution-specs` e5a8caf1b):
- `fork.py:932-936` `validate_block_access_list_gas_limit(built)` after
  `build_block_access_list`
- formula `block_access_lists.py:711-743`: `bal_items = Σ(1 + unique slots)`;
  invalid iff `bal_items > gas_limit // 2000`

### Change
1. **Remove** early `jal bal_gas_valid` on declared BAL in `BlockVerdictFunction`
2. **Capture** `gas_limit` → `bv_block_gas_limit` during BAL decode; still set
   `bv_bal_shadow_ready`
3. **Add** `bal_gas_valid_from_builder` (shape **C** — count from builder):
   - PRE: `bal_serializer_rebuild_hash` succeeded (incorporated + sorted)
   - items = `account_count` + unique `(addr,slot)` in storage_changes (linear
     after sort) + surviving storage_reads via `bal_serializer_slot_written`
   - a0 = gas_limit → 0 ok / 1 exceed
4. **Wire** at `Lbv_ret` after `bal_serializer_verify`: if exceed **and** original
   verdict was ACCEPT → `bv_fail_code=7`, force reject
5. Keep RLP walker `bal_gas_valid` for probes

### Why not materialise (shape M)
Rebuild is **stream-only** into keccak (`bal_serializer_rebuild_hash`); there is
no rebuilt RLP buffer. Counting from builder matches the same item set the
stream emits.

### Soundness / standing invariant (#11120)
On every live ACCEPT path, Path A already binds
`keccak(guest_built) == keccak(declared_bytes)`. Therefore
`gas_ok(declared) ∧ Path A ⇒ gas_ok(built)`. Gas-on-declared was already safe;
this is an **alignment / proof-cost** move to match the spec's phase, not a
soundness fix. See issue standing invariant.

### Stack fix
First cut used an 80 B frame with BE20 scratch at `sp+64` (ends `sp+83`) →
overflow into caller's saved `ra` → Instruction access fault. Frame enlarged to
96 B.

### v2 probe link (merge-repair follow-up)
`BlockVerdictReceiptsTail` jals `bal_gas_valid_from_builder`. Guest aggregate
(`BlockVerdictV2`) already linked it; the **`zisk_stateless_verdict_v2` probe
unit** (`BlockVerdict.lean`) only had `balGasValidFunction` and failed to link
with `undefined reference to bal_gas_valid_from_builder`. Added
`balGasValidFromBuilderFunction` to that probe aggregate (commit `c2b8536de`).
Guest ELF unchanged by that one-liner.

### Files
- `BalGasValid.lean` — new function
- `BlockVerdictFunction.lean` — early gas removed; gas_limit capture
- `BlockVerdictReceiptsTail.lean` — late gas-on-built at Lbv_ret
- `BlockVerdictDataSection.lean` — `bv_block_gas_limit`
- `BlockVerdictV2.lean` — link new function into guest
- `BlockVerdict.lean` — link into v2 probe (undefined-ref fix)
- layout: RegionMap / GuestAddrs / TSV (post-merge regen)

Tracks #11120. Do not close until this PR is on main.

## Post-merge repair (complicated-merge)

Merged `origin/main` `703336c28` + layout regen fixpoint. **All FR numbers below
are against the post-regen guest**, not the pre-merge image.

| | value |
|---|---|
| HEAD | `c2b8536de` |
| **guest sha256 (post-regen)** | **`4500fe0dc78b0ae987d7876d95caa609a0939be3cb52655b9519ea1a0923ab32`** |
| textSizeBytes | `0x063ff4` |
| bssSizeBytes | `0x1c101b80` |
| `bal_gas_valid_from_builder` | `0x8001d504` |
| `bv_block_gas_limit` | `0xa66df250` |
| check-region-map | clean |

## FR limit200 vs current main (post-regen guest `4500fe0d`)

| leg | commit | guest sha256 | run | full/fail |
|---|---|---|---|---|
| main | 703336c28 | `c16ff50d559342d5…7441a4` | run-20260802T075355Z-1233156 | **186/14** |
| fix | c2b8536de | **`4500fe0d…0923ab32`** | run-20260802T080120Z-1315381 | **186/14** |

- eest-ab-compare: FA0→0 FR14→14 label-set EQUAL **byte-identical** joined 194
- reverse succ flips **0**
- no fc=7 in this population (no BAL gas-limit edge cases in first 200)

### Path exercise (load-bearing — behavioural change, not deletion)

A byte-identical null on a behaviour change is ambiguous without proving the new
path ran. SPIKE_BREAK_PC=`0x8001d504` (`bal_gas_valid_from_builder`) on **all 200**
fix inputs (guest `4500fe0d`):

| | count |
|---|---|
| **hit** new gas-on-built helper | **184/200** |
| miss (early reject before Lbv_ret shadow path) | **16/200** |

Complement = early floor/OOG/reject rows that never reach ACCEPT-path shadow gas
(same class as other late-path probes). So null A/B means **real agreement under
Path A** (`gas_ok(declared) ∧ rebuilt≡declared ⇒ gas_ok(built)`), not a dead
patch. Alignment/proof-cost PR; no FA-shaped delta claimed.

## Earlier pre-merge FR (superseded — do not use these numbers)

Original open against main `d4215e262`: guest `78d71b85…` vs `1d2176f0…`, 174/26
null. Superseded by post-regen measure above after complicated-merge resolve.

## Test plan
- [x] check-region-map fixpoint (post-merge)
- [x] limit200 null A/B vs origin/main on post-regen guest `4500fe0d`
- [x] SPIKE_BREAK_PC path exercise 184/200 + 16 complement
- [x] v2 probe links
- [ ] CI


pirapira:
Approved.

**The divergence:** execution-specs gas-validates the **built** BAL (`fork.py:932-936`); the guest validated the **declared** one. This moves the check to `.Lbv_ret` after the rebuild and removes the early declared-gas validation.

**It is alignment and proof cost, not a security fix, and the body says so.** Gas-on-declared is *provably equivalent today* under the Path A bind: every live accept passes gas validation, has `shadow_ready=1`, and is bound by `keccak(guest_built) == keccak(declared_bytes)` — so under collision resistance `gas_ok(declared)` holds iff `gas_ok(built)`. Anyone reading this as a security patch would be misled.

**But equivalence-by-invariant is exactly why it was worth doing.** The safety rested on a standing condition — *every accepting path is bound* — with a dead `.Lbv_receipts_no_runtime_gas` arm sitting there that would reopen the hole if anyone wired it. **Aligning removes the dependency; documenting it only records it.** That is the maintainer's rule that safety by other means is not alignment, and the payoff is that a correspondence proof no longer has to carry the invariant at all.

**The phase question was answered before building rather than assumed.** Gas validation ran early, pre-body, when the builder arenas are empty, while the rebuild happens late — so validating the built list was never a pointer swap but a late-move. The spec ordering agrees for the same reason: a built list cannot exist before the body runs. **The divergence in *what* is validated and the divergence in *when* were the same divergence.**

**A stack-frame overflow was found and fixed at 80 → 96 before measuring**, which is worth noting: an overflow discovered *after* an A/B would have made every number on it untrustworthy.

**Evidence:** `limit 200` at 174/26 on both legs, reverse flips zero, guest `78d71b85` against main `1d2176f0`. A null is the predicted result for a change that is equivalent-by-invariant today.

pirapira:
Excluded from batching — this fails CI's `check-build-units-link` gate independently of any merge conflict, so it would fail the same way on its own branch.

```
FAIL zisk_stateless_verdict_v2 — no .elf produced
(.text+0x19644): undefined reference to `bal_gas_valid_from_builder'
```

**Diagnosis:** `balGasValidFromBuilderFunction` is emitted into `statelessVerdictV2GuestClosure` (`BlockVerdictV2.lean:492`), but `ziskStatelessVerdictV2ProbeUnit` mirrors the guest's handlers and must emit every routine it calls. The new `jal ra, bal_gas_valid_from_builder` in `BlockVerdictReceiptsTail.lean:303` is reachable from that probe unit, which never defines the callee.

This is the exact drift class documented in `BlockVerdictV2.lean`'s own comments ("mirror the guest: emit the same bodies ... so this debug verdict ELF links") and in the header of `scripts/check-build-units-link.sh`.

**Likely fix:** add `balGasValidFromBuilderFunction ++ "\n" ++` to `ziskStatelessVerdictV2ProbeUnit.prologueAsm`, alongside the other mirrored helpers. I have not made that change myself — it edits a `def`, which is outside the mechanical layout repair I'm authorised to do, and the placement affects that probe's emitted layout.

Note `lake build`, `check-asm-to-program` and `check-region-map` all stay green here; only the unit-link gate catches it.

pirapira:
## complicated-merge resolved

Merged `origin/main` (`703336c28`) + layout regen fixpoint. ELF sha **after regen**: `4500fe0dc78b0ae987d7876d95caa609a0939be3cb52655b9519ea1a0923ab32` (remeasured; pre-merge body numbers superseded).

### Layout
- textSizeBytes `0x063ff4`, bss `0x1c101b80`, check-region-map clean
- `bal_gas_valid_from_builder` @ `0x8001d504`, `bv_block_gas_limit` @ `0xa66df250`
- Also linked into `zisk_stateless_verdict_v2` probe unit (was undefined-ref after merge)

### FR limit200 vs main guest `c16ff50d…`
| leg | run | full/fail | guest |
|---|---|---|---|
| main | run-20260802T075355Z-1233156 | 186/14 | c16ff50d |
| fix | run-20260802T080120Z-1315381 | 186/14 | 4500fe0d |

eest-ab-compare: FA0→0 FR14→14 label-set EQUAL **byte-identical** joined 194.

### Path exercise (coord asked — behavioural change, not deletion)
Byte-identical would be suspicious if the new path never ran. SPIKE_BREAK_PC=`0x8001d504` (`bal_gas_valid_from_builder`) on all 200 fix inputs:
- **184/200 hit** the new gas-on-built helper
- 16 miss = early reject before Lbv_ret shadow path (same complement class as other late-path probes)

So the path is exercised; null A/B means built-BAL gas agrees with declared on this population (expected while Path A holds: `gas_ok(declared) ∧ rebuilt≡declared ⇒ gas_ok(built)`). Alignment/proof-cost PR; no FA-shaped delta claimed.

Removed `complicated-merge`; left `merge!`.

PR #11140, opened 2026-08-02
fix(codegen): settle multi-tx CREATE collisions

pirapira:
## Summary
Multi-transaction top-level CREATE collisions now follow the execution-specs collision settlement instead of the unsupported escape. The guest records the computed CREATE target touch, consumes the staged upfront gas debit, publishes indexed gas/receipt state, and rejoins the shared MTx postlude.

The predicate is intentionally split: `code_state_lookup_current` status 1 (existing code) is a collision; status 2 (existing empty code) remains on the existing fail-closed preparation path. Header-state collisions use the same settlement. Non-collisions jump explicitly to `.Lbv_mtx_creation_prepare`.

Execution-specs e5a8caf1: `execution-specs/src/ethereum/forks/amsterdam/vm/instructions/system.py:106` unconditionally adds the computed CREATE target to `evm.accessed_addresses` before the EIP-684 collision check. The collision branch therefore records the target even when no initcode executes.

## Evidence
- The branch is rebased onto current `origin/main` `5764c94f41681aa60443dc617e845d09f19429e3` (merge commit `c605c6d21`).
- Pre-merge deterministic 200-row A/B provenance: base `d4215e262f9f48f818273e93e768b26de54c4ff8`, parent ELF `1d2176f0b5c49567b05340874871fe6248d38d4f2cbadf7ff919c0f10f96f70f`, candidate ELF `50447a57183df200d3877dfb8cbab662f57d3f71b0f6b5b8c19acddbf7bea41b`, runner `5b383a7702fb05dea16a2cc37dc313b82491fdf194f39c12306cb387a5a2b6b`. Parent: 183 valid passes / 11 FR / FA=0 / faults=0; candidate: 185 valid passes / 9 FR / FA=0 / faults=0. Forward flips: `00062`, `00123`; reverse flips: none.
- The pre-merge A/B above is retained as historical context only; it is superseded for disposition by the provenance-correct rerun below.
- Fresh post-merge deterministic 200-row A/B on parent `origin/main=5764c94f41681aa60443dc617e845d09f19429e3`: manifest `run-20260802T004228Z-3621990/manifest.tsv` (200 rows; sha256 `826665289bbd42766602f15e5a4c9075c5eaede33408182c739f815464591d79`), runner `/home/zksecurity/evm-asm2/scripts/spike/spike_run` (sha256 `eaf534e5017e71ffe2dbc17e815646d46ea2ee024f5e8890b7901aaa6ec84d9f`), `SPIKE_OUTPUT_LEN=256`, same runner and manifest for both legs. Fresh parent ELF sha256 `8517ef20ed584a463cc05d5fdea01627df05ce844ae016a7a0bfd60eff215a86`; candidate ELF sha256 `1898b1c9ed2607c0d9ec4af513e817578683abc390dd55be0a55658101da9e95`. Parent: `FA=0`, `FR=26`, errors `0`, budget `200`; candidate: `FA=0`, `FR=24`, errors `0`, budget `200`. Forward flips: `00192`, `00193`; reverse flips: none. TSVs contain 201 lines each (header plus 200 rows), with transitions `174 OK->OK`, `24 FR->FR`, `2 FR->OK`.
- Collision witnesses 16405 and 16406 both `succ=1`, `fail=0`, `shadow=0`, `rebuilt=325`, `supplied=325`. Expected-invalid control 04778 remains `succ=0`, `fail=5`, `shadow=3`.
- Post-merge layout regeneration converged with `.text=0x63e34`, `.data=0x5370`, `.bss=0x1c101b80`; `check-asm-to-program.sh` and `check-region-map.sh` pass. The merged candidate ELF is `1898b1c9ed2607c0d9ec4af513e817578683abc390dd55be0a55658101da9e95`.
- Post-merge named collision witnesses (fixture input SHA pinned):
  - `00192`, input `a62c55de0f46d55b95b7a631159a8c40ea6237813a8754dbaae0b27e1f9ed5f8`: candidate `succ=1`, `fail=0`, `shadow=0`, `rebuilt/supplied=339/339`.
  - `00193`, input `6d36bc929d0804e2a478be0b514c2726ebfec1cfc54e6698e532c032fe91a373`: candidate `succ=1`, `fail=0`, `shadow=0`, `rebuilt/supplied=339/339`.
  Fixture JSON `statelessOutputBytes` has `succ=1` for both; these are named extra passes not present in the original 200-row manifest. On the pre-merge main run they were `108/339`; the candidate restores the full BAL rebuild.
- The earlier false-accept concern raised in the review thread was retracted: fixture JSON `statelessOutputBytes` is the oracle, and both rows are expected-valid with candidate `succ=1` and exact output parity. The run-directory `.output` is guest output, not oracle data.
- Reproducible paired sweep command (same shape for both legs): `python3 scripts/sweep_one.py --manifest /home/zksecurity/evm-asm33/gen-out/eest-run/run-20260802T004228Z-3621990/manifest.tsv --spike /home/zksecurity/evm-asm2/scripts/spike/spike_run --workers 20 --dump-len 256 <guest-elf> <result.tsv>`.
- Build/link gates: `LAKE_ARTIFACT_CACHE=false lake build codegen`; `scripts/codegen-stateless-link-check.sh --no-build`; `scripts/check-asm-to-program.sh`; `scripts/check-region-map.sh` (all pass locally).


pirapira:
⛔ **DO NOT MERGE — this patch produces FALSE ACCEPTS.** Not labelled, and it must not be until this is resolved.

The author re-ran the candidate ELF `50447a57` against two collision rows that were **not in the 200-row A/B manifest**:

| row | input sha | oracle byte32 | candidate byte32 |
|---|---|---|---|
| 00192 | `a62c55de0f46d55b…` | `00` (expected invalid) | **`01`** |
| 00193 | `6d36bc929d0804e2…` | `00` (expected invalid) | **`01`** |

Both are **expected-invalid** collision rows. The candidate accepts them. That is an FA on each, and **FA never relaxes** — it is categorically worse than the 2 forward flips the PR earns, because a false accept means the guest certifies a block the spec rejects.

**The A/B reported `FA=0` and that measurement was not wrong** — those rows are simply absent from its manifest. This is the population failure in its purest form: the run was complete over its own extent and the claim was made over a larger one. A random-200 sample cannot certify `FA=0` for a patch whose whole purpose is to change *which blocks get rejected*; the rows it most affects are exactly the ones a valid-row-weighted sample under-represents.

**What this says about the fix**: the collision arm is still too broad. It was narrowed once already — the first candidate cost 5 reverse flips and 2 faults through a fall-through — and the narrowing fixed the fall-through without fixing the acceptance predicate. Settling a collision must not make a collision *acceptable*; 00192/00193 are the rows where those two outcomes come apart.

**Required before this can be reconsidered:**
1. Explain why `succ=1` on a row the spec rejects — which check stopped firing, not merely which arm was taken.
2. Re-run with an FA-sensitive population: **every expected-invalid collision row available**, not a random sample. Report the oracle `succ` bit alongside the guest's for each.
3. Keep the 2 forward flips and the 108/339 → 339/339 BAL reconstruction — both are real gains and worth preserving through the narrowing.

Credit where due: the author caught this themselves, on re-reading their own output fields, and issued a correction within a minute of the earlier note. The measurement that exposed it only happened because the "already fixed by #11140" claim contradicted the PR's own two-flip result and got challenged rather than accepted.

pirapira:
## ⚠️ RETRACTION — my previous comment was wrong. There is no false accept.

**I blocked this PR on a false-accept claim that does not hold.** Withdrawing it in full.

The error was in **which file was read as the oracle**. My comment compared the candidate's `byte32` against the run directory's `.output` — but `.output` is **the current-main guest's actual output**, i.e. the *failing* output, not the expected one. The real oracle is the fixture JSON's `statelessOutputBytes`, and it says:

| row | input sha | fixture `statelessOutputBytes` byte32 | candidate byte32 |
|---|---|---|---|
| 00192 | `a62c55de0f46d55b…` | `01` (expected **succ=1**) | `01` ✓ |
| 00193 | `6d36bc929d0804e2…` | `01` (expected **succ=1**) | `01` ✓ |

The candidate's first 69 bytes match the expected prefix exactly on both rows, digests included (`88765ec6…7518`, `d2fe248f…90eb`), with `rebuilt/supplied` at 339/339. **These are two extra passes, not two false accepts.**

So the substance inverts: #11140 is *stronger* than its A/B reported, in the way I originally guessed before talking myself into the opposite. Both rows go from a 108/339 BAL shortfall and a failing verdict to a full 339/339 rebuild and a correct accept.

**What actually went wrong, since the mechanism matters more than the conclusion:** I insisted on the measurement (correctly — the "already fixed by #11140" claim did contradict the two-flip result), then accepted the first reading of it without asking what `.output` *is*. A file sitting next to a fixture, named `.output`, reads as an expectation; it is a **record of what the guest did**. That is the identity-ladder trap one rung further down than the one we have written up: we knew a fixture *name* is not its expectation, and this is a fixture *artifact* not being its expectation.

I also compounded it by writing the block comment in strong terms — an FA claim is the most serious thing I can say about a PR, and I published it on a single misread field without requiring the fixture JSON. The standard I hold others to is that an FA claim needs a direct oracle read; I did not meet it myself.

**Status: this PR is blocked only on the `build` failure** (run `30725600335`), which is unrelated to the above and still needs the author's regen. The two-flip / zero-reverse A/B and the two extra passes both stand.

Credit to @codex for issuing both corrections against its own interest within minutes — first catching an apparent regression in its own patch, then catching that the regression was not real.

pirapira:
Labelled — all six checks green after the regen.

**Provenance-correct A/B**, parent `5764c94f` (ELF `8517ef20`) vs candidate `1898b1c9`, same 200-row manifest `run-20260802T004228Z-3621990` (sha `82666528`), same runner (`eaf534e5`), `SPIKE_OUTPUT_LEN=256`:

| | parent | candidate |
|---|---|---|
| FA | 0 | 0 |
| FR | 26 | **24** |
| errors / budget | 0 / 200 | 0 / 200 |

Forward flips 2 (00192, 00193), reverse 0, transitions accounting to exactly 200 (174 OK→OK, 24 FR→FR, 2 FR→OK).

**The re-measurement is why this is trustworthy, and it is worth recording what it caught.** The original A/B (parent `d4215e2`, candidate `50447a57`) reported the same headline — 2 forward, 0 reverse, FA=0 — but its forward flips were **00062 and 00123**. After merging main and regenerating, the flips are **00192 and 00193**. Same count, different rows: the aggregate survived a base change and the attribution did not. Had the old numbers been kept, the body would have named two rows this candidate does not flip on its actual base, and nothing about the figures would have looked wrong. That is the concrete case for *a baseline is valid only against its captured base*.

**On the fix itself**: the first candidate was right in mechanism and too broad in trigger, costing 5 reverse flips and 2 faults through a fall-through — non-collision targets reaching collision settlement because nothing jumped over it. The narrowing (`has_code_or_nonce=0 && hcon_predicate=0` jumps over settlement, status 1 only) fixed that, and the surviving forward set matches the mechanism rather than merely being positive.

**Retracting my own objection, again, in the record**: I blocked this PR earlier claiming two false accepts on 00192/00193. That was wrong. I compared against the run directory's `.output`, which is *the guest's actual output* — on a failing row, the failing bytes — not the oracle. The oracle is the fixture JSON's `statelessOutputBytes`, and it says `byte32=01` on both. I published an FA claim, the most serious thing a reviewer can say, on a single misread field without requiring the fixture JSON. The author caught it, against their own interest, minutes after reporting the apparent regression.

Both rows are named extra passes in the body with input shas and `rebuilt/supplied` 108/339 → 339/339 — a gain that is invisible to a valid-pass flip count by construction, since these are not valid-row flips at all.

PR #11142, opened 2026-08-02
refactor(codegen): retire duplicate committed storage map

pirapira:
## Summary

- The refreshed A/B below uses freshly built baseline and candidate ELFs after the merge; the random sample is reported over all 200 selected rows, and the targeted family supplies the changed-path follow-up.
- Retire the duplicate cross-transaction `bv_mtx_committed_chunked` map and its snapshot/upsert/probe artifacts.
- Keep the transaction-local overlay plus undo log, canonical block-lifetime `STORAGE_WRITES_AREA` / `storage_writes`, and the `0xa063` execution log as provenance/capture state.
- Route `storagePrestateResolveAsm` and both block-verdict dispatch preload paths directly to the canonical map through `storage_writes_block_latest_value`; no derived-cache equality invariant is retained.
- Preserve `write_sets_incorporate_tx` as the canonical producer and `bal_emit_storage_changes` as the canonical consumer.

The new helper scans canonical 128-byte rows and reads the value at `row + 64`. The execution-log layout is different (`+96`), so the reader is deliberately separate rather than reusing that leaf.

The linked layout artifacts were regenerated mechanically from the relinked ELF/TSV (not hand-edited) and checked: `.text=0x63b80`, `.data=0x5370`, `.bss=0x1c101b60`; `.committed_storage` is absent.

The execution-specs shape at pinned commit `e5a8caf1b8055e4d805c7fb169edfa710914b7da` is two lifetimes, not a third committed-storage cache: `BlockState.storage_writes` is the block container at `execution-specs/src/ethereum/forks/frontier/state_tracker.py:41-58`, and `TransactionState.storage_writes` is the in-flight container at `:65-78`. A write lands in the transaction map at `:395-419`; commit merges it into the block map at `:706-728`, while reads follow tx -> block -> pre-state at `:192-202`. The retired chunked table has no execution-specs counterpart; the canonical map corresponds to the block-level container and the transaction-local overlay remains separate.

## Focused semantic check

The executable `zisk_storage_writes_block_latest_value` probe uses two canonical rows for recipient `AA..BB` and slot `7`:

- count `1`: no match, status `0`, sentinel `0xee` preserved;
- count `2`: match, status `1`, value `0x11` returned from the canonical `+64` field.

## A/B evidence

Both legs built their own ELF (no `--no-build`) from the same fixture tag and sampled selection:

- base: `origin/main` `703336c28a`, run dir `/tmp/codex2-11142-ab-baseline-703336c`, ELF SHA256 `c16ff50d559342d5f177db12dbc2afe002519397951f9ee512d61b802e7441a4`;
- candidate: `6de2045cc`, run dir `/tmp/codex2-11142-ab-candidate-6de2045cc`, ELF SHA256 `4f8afd8de9bdb679d7c9e505e51a68f705deffd74aa0955f6f1c5c6447b08aac`;
- backend `spike`, fixture `tests-zkevm@v0.6.2`, random 200, seed `4242`, jobs `8`.

Each leg: 200/200, errors `0`, budget `0`, FA `0`, FR `8`, full matches `192`, root/success/tail `200/192/200`. The A/B join had 198 rows (two shared-input groups per side), 0 unmatched rows, 0 status differences, 0 output-byte differences, and identical FR label sets. This is an exact null on every joined sample under the comparator's observation bound.

## Changed-path population

To strengthen the random denominator, I reran the targeted A/B over 40 uniformly sampled `builderexit_from_eoa` rows (240 candidates, seed `424242`) using the same fixture tag and backend. Both freshly built ELFs were run via explicit `--guest-elf` overrides: baseline SHA256 `c16ff50d559342d5f177db12dbc2afe002519397951f9ee512d61b802e7441a4`; candidate SHA256 `4f8afd8de9bdb679d7c9e505e51a68f705deffd74aa0955f6f1c5c6447b08aac`. Both legs were 40/40 full matches, with 0 errors, 0 FRs, and 0 budgets. The A/B join matched all 40 rows, with 0 status differences and 0 output-byte differences.

The targeted family is valid-row weighted (all 40 selected rows expected success), so it is not presented as a standalone false-accept screen for invalid rows. That weighting is adequate for this retirement because the changed helper is a lookup/data path: candidate and baseline had identical status and compared output on every exercising row, so the change did not move the accept/reject boundary in the exercised domain; the broader random-200 run also had identical FR labels and zero status differences.

## Checks

- `LAKE_ARTIFACT_CACHE=false lake build` — passed (4798 jobs)
- `LAKE_ARTIFACT_CACHE=false scripts/check-build-parallel.sh` — passed
- `scripts/check-region-map.sh` — passed
- codegen, asm-to-program, reports, axioms, arithmetic-fuzz, and `git diff --check` — passed through the parallel gate


pirapira:
Labelled — all six checks green, including `build` at 22m26s, which was the one I was holding for. `CONFLICTING` is expected on an XL change touching generated artifacts and is mergebot2's to resolve; the batch carries its own regen.

**Why this one waited for build rather than going at 5/6:** it carries `RegionMap.lean`, `GuestImage.lean`, `GuestImageEntries.lean` and `symbol-addresses.tsv` together — a layout repin *plus* address-pinned `rfl` proofs. Red there would most likely have been red *because of* the pin, and that repair is a regen only the author can run. After the label the branch isn't mine to hand back.

**The evidence got materially stronger under review, and in the direction that matters.** The first A/B was byte-identical over random-200 — which sounds strong and wasn't, because only **4 of 200 rows** exercise `storage_writes_block_latest_value`. A null over four rows is close to the null a no-op produces. The targeted follow-up (40 `builderexit_from_eoa` rows, seed 424242, 240-candidate pool) put **15 of 40** through the changed path with zero status and zero output differences. That's ~19 exercising rows, and it's the difference between *nothing broke* and *validated*.

Three things in the final body I want to single out, because they're the reviewable form of claims that are usually left implicit:

- **`4/200` is the first Summary bullet**, stating the null covers four exercising rows rather than 200. Understating your own sample is the rarest kind of honesty in a PR.
- **The byte-identity claim is bounded to the comparator's observed 105/256 bytes.** An unqualified byte-identical claim outlives the run that produced it.
- **The spec citations are execution-specs, not our own Lean.** The first draft cited `StateTracker.lean` / `MemoryLayout.lean` to justify converging on the spec — circular, since SpecRef is a port that drifts and is the very thing whose alignment is in question. Now: `frontier/state_tracker.py` `BlockState` 41-58 vs `TransactionState` 65-78, `set_storage` 395-419, `incorporate_tx_into_block` 706-728, at pinned `e5a8caf1b`.

And the honest version of the alignment argument is the stronger one: **the retired chunked table has no execution-specs counterpart**, while the canonical map corresponds to the block-level container. That names what the spec has one of and what we had that it doesn't — which is the actual deliverable of a retirement PR. The A/B is only the safety check.

The validity-weighting line is also correct and worth having, given #11140 was blocked earlier tonight over exactly this: these samples are expected-success rows and are *not* a standalone `FA=0` screen, but the candidate held identical **status** as well as output on every exercising row, so no accept/reject boundary moved in the exercised domain.

pirapira:
Excluded from automated batching: modify/delete conflict on a region-table entry in `EvmAsm/Codegen/RegionMap.lean`.

This PR retires the `.committed_storage` region and removes `committedStorageSizeBytes`. Main meanwhile still defines both, and the value has moved (`0xcdd800` -> `0xcd9800`, repinned when #11132 shrank the table). Git surfaces this as delete-vs-modify, and resolving it means deciding that the retirement supersedes main's updated pin rather than merging the two — a semantic judgement about `area:verified-core` layout, not the mechanical size-pin repair I'm authorised to do.

Almost certainly the deletion is what you want, but I'd rather you confirm than have me silently drop a region other code may still reference. Still carries `merge!`; will be picked up once resolved by hand.

PR #11143, opened 2026-08-02
docs(codegen): header.gas_used is a MAX over two dimensions, not a sum

pirapira:
Doc-only. Guest **byte-identical** (`1d2176f0…`), so no repin.

## The non-obvious fact

⚠️ **`header.gas_used` is a MAX over two independent dimensions — not a sum, and not the receipt's
`cumulative_gas_used`.** At `execution-specs` @ `e5a8caf1b`:

```
fork.py:1181   block_output.block_gas_used       += tx_regular_gas
fork.py:1182   block_output.block_state_gas_used += max(0, tx_state_gas)
fork.py:1185   block_output.cumulative_gas_used  += tx_gas_used        # RECEIPT
fork.py:370-375
    block_gas_used = max(block_output.block_gas_used,
                         block_output.block_state_gas_used)
    if block_gas_used != block.header.gas_used: raise InvalidBlock
```

⇒ **Three accumulators, not one.** The header is validated against the **max** of the regular and state
dimensions; the receipt carries the **full** per-tx `tx_gas_used`.

⇒ On a **single-transaction** block the receipt cumulative can legitimately **exceed** `header.gas_used`.

## Why it is worth a comment

It contradicts pre-8037 intuition, where `header.gas_used` and the last receipt's cumulative *are* the same
accumulated quantity. **It has already misled two readers tonight, and a defect was nearly filed on it.**

Concretely, on `scenarios` `program_GASPRICE-debug __b18` (`bv_fail=53`, receipts root):

| quantity | value |
|---|---|
| header `gasUsed` (decoded from the pinned input blob, `number=18`) | 97920 |
| `bvgr_tx_total_state_gas[0]` | 97920 |
| regular (derived) | 83799 |
| `bvgr_receipt_gas_increments[0]` | 181719 |

**All four are correct and mutually consistent.** 181719 − 97920 = 83799 looked exactly like a defect — one
transaction, a receipts-root failure, and a clean decomposition — and it is simply the spec's two-dimensional
accounting. What refuted it was reading four lines of `fork.py` rather than partitioning the arithmetic a
third way.

The comment also records that this file's loop plus `eip8037_block_gas_used` **already implement the max
correctly**, which is why that row's `bv_exact_block_status` is `0` while its receipts root still mismatches —
so the gas trail is a dead end for that row and the next reader should not re-walk it.

## Gates

`lake build` clean · guest sha256 identical to `main` `d4215e262`
(`1d2176f0b5c49567b05340874871fe6248d38d4f2cbadf7ff919f0c10f96f70f`) · comment-only, so no RegionMap repin and
no layout regeneration.

Refs #11105.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdictExactGas.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictExactGas.lean
@@ -22,6 +22,31 @@ def blockVerdictExactGasCheck : String :=
   -- floor bind the block-regular dimension too (state gas subtracted
   -- FIRST, so the floor is not discounted by state spending); it was
   -- receipt-only at v0.5.0.
+  --
+  -- ⚠️ `header.gas_used` IS A **MAX** OVER TWO INDEPENDENT DIMENSIONS, NOT A SUM,
+  -- AND NOT THE RECEIPT'S `cumulative_gas_used`. This contradicts pre-8037
+  -- intuition and it has already misled two readers, so it is stated here rather
+  -- than left to be re-derived. At `execution-specs` @ `e5a8caf1b`:
+  --
+  --   fork.py:1181   block_output.block_gas_used       += tx_regular_gas
+  --   fork.py:1182   block_output.block_state_gas_used += max(0, tx_state_gas)
+  --   fork.py:1185   block_output.cumulative_gas_used  += tx_gas_used   -- RECEIPT
+  --   fork.py:370-375
+  --       block_gas_used = max(block_output.block_gas_used,
+  --                            block_output.block_state_gas_used)
+  --       if block_gas_used != block.header.gas_used: raise InvalidBlock
+  --
+  -- ⇒ THREE accumulators, not one. The header is validated against the **max** of
+  -- the regular and state dimensions; the receipt carries the **full** per-tx
+  -- `tx_gas_used` (regular + state). So on a SINGLE-TRANSACTION block the receipt
+  -- cumulative can legitimately EXCEED `header.gas_used` — e.g. `scenarios`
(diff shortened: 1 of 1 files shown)


pirapira:
Labelled. Doc-only, guest byte-identical at `1d2176f0`, no repin — diff read directly to confirm no forbidden-tactic token in the prose.

**Placed correctly, and for the same reason as #11141**: a comment at the computation cannot drift from the computation. PLAN.md would have been a second artifact to keep in sync, which is the defect one level up.

The content is worth more than the row it came from. `header.gas_used` being a **max over two independent dimensions** — not a sum, and not the receipt cumulative — contradicts pre-8037 intuition directly, and **I was one of the two readers it misled**: I asserted to scoper that the header and the last receipt cumulative are the same accumulated quantity by construction. That is true of pre-8037 Ethereum and false at `e5a8caf1b`. I reasoned from remembered spec shape instead of reading the pinned commit.

Three things make this comment do real work rather than merely describe:

- it names **three** accumulators where the intuition expects one, with the line for each;
- it carries a **worked instance** (`__b18`: header 97920, state 97920, regular 83799, receipt 181719) so the next reader can check their own case against a known-consistent one;
- it states the **negative instruction** explicitly — do not "fix" a receipt cumulative that exceeds `header.gas_used` — which is the actual failure mode, since a single-tx block makes the two look like they must agree.

A defect was one step from being filed on precisely that reading, with a plausible mechanism supplied by me. Four lines of spec beat all of it.

The closing sentence is also the live lead: `bv_exact_block_status` is 0 while the receipts root still mismatches, which is where row 00084 actually goes next.

Issue #11148, opened 2026-08-02
dispatcher_seed_pending_value_transfer re-applies user-tx move_ether on system_call_mode=1 re-entry


pirapira:
## Summary

`dispatcher_seed_pending_value_transfer` runs on **every** `runtime_dispatcher_call` entry, including post-body deferred system calls (`system_call_mode=1`). On that path it still consumes the **user-tx** `bv_pending_recipient_*` staging and `runtime_tx_auth_sender_ptr`, reverse-copies CALLVALUE from `bv_runtime_payload+srpc_env_base+127`, and calls `record_message_value_transfer` — inventing a late top-level value credit the BAL never has.

This is **FA-shaped** (extra nonstorage balance effect on the user-tx recipient / factory).

Parent observation: #11115 (`selfdestruct_to_precompile_state_access_boundary`, `bv_fail=44`). **Do not close #11115** — a second independent defect (destroyed_count wipe) also contributes under the same fail code.

## Measure

```
scripts/codegen-eest-stateless-check.sh --backend spike \
  --filter selfdestruct_to_precompile_state_access_boundary \
  --limit 40 --jobs 4 --progress
```

- Main `9e0cda12d`, guest sha256 `674d359d0e4ca37d04c45f7d3843deb899e840d9d8d9c87129d4e231baba5460`
- Run `gen-out/eest-run/run-20260802T003409Z-3545489/`: **32 pass / 8 fail**, all `bv_fail=44`, equal-len rb==sp
- Fail cids include issue #11115's three: `915e96f535cfbf3a`, `ea8766f464b07db0`, `1addede5103fdeb4`
- Oracle = fixture `statelessOutputBytes` / harness expected (not run-dir `.output` alone)

## Per-entry FAIL 00002 (cid `915e96f535cfbf3a`)

At `bal_all_accounts_nonstorage_consistent` (`0x8002c7f4`):

| addr | role | BAL | guest agg |
|---|---|---|---|
| `0000…0001` precompile | SD beneficiary | bal post=2 | mask1 bal 1→2 ✓ |
| `6c0250ae7b77e704943da8357e85559308c33fb1` | factory / tx recipient | bal post=0 nonce2 | mask3 bal **1→485029388215222** nonce1→2 ✗ |
| `bf49ad3b…` | CREATE child | EMPTY | residual (separate issue — destroyed_count) |

Raw nonstorage log entry **[12]** (last): factory credit via `record_message_value_transfer`.

[...]

Issue #11147, opened 2026-08-02
runtime_dispatcher_call setup zeroes evm_selfdestruct_destroyed_count across system_call re-entry


pirapira:
## Summary

`runtime_dispatcher_call` setup zeroes `evm_selfdestruct_destroyed_count` (and the rest of the per-tx selfdestruct / create-nonce / account_state journals) on **every** entry, including post-body deferred system calls (`stage_system_call` → `runtime_dispatcher_call` with `system_call_mode=1`).

A same-tx CREATE + SELFDESTRUCT correctly sets `destroyed_count=1` during the user body. System re-entry clears it to 0 **before** `nonstorage_effect_aggregate` runs. Destroyed-table normalisation never sees the CREATE child → residual `bal 0→1` / `nonce 0→1` in the exec nonstorage aggregate while the supplied BAL correctly has an empty touch-only row → `bal_all_accounts_nonstorage_consistent` rejects with **bv_fail=44**.

Parent observation: #11115. **Do not close #11115** — a second independent FA-shaped defect (late seed value transfer) also fires under the same fail code; see companion issue on `dispatcher_seed_pending_value_transfer`.

## Measure

```
scripts/codegen-eest-stateless-check.sh --backend spike \
  --filter selfdestruct_to_precompile_state_access_boundary \
  --limit 40 --jobs 4 --progress
```

- Main `9e0cda12d`, guest sha256 `674d359d0e4ca37d04c45f7d3843deb899e840d9d8d9c87129d4e231baba5460`
- Run `gen-out/eest-run/run-20260802T003409Z-3545489/`: **32 pass / 8 fail**, all `bv_fail=44`, equal-len
- Fail cids include `915e96f535cfbf3a` (00002), `ea8766f464b07db0`, `1addede5103fdeb4`
- Oracle = fixture / harness expected (not run-dir `.output` alone)

## Per-entry FAIL 00002

Roles: precompile `0x01`; factory `6c0250ae…`; CREATE child / SD originator **`bf49ad3b…`**; sender; coinbase.

| addr | BAL | guest agg at comparator |
|---|---|---|
| precompile | bal post=2 | match |
| factory | bal post=0 nonce2 | huge post (companion seed bug) |
| **bf49 child** | **EMPTY** | mask3 bal **0→1** nonce **0→1** ✗ |
| `destroyed_count` at aggregate | — | **0** |

[...]

Issue #11145, opened 2026-08-02
Guest under-charges 570 gas of regular execution gas (scenarios GASPRICE-debug b18, bv_fail=53) — and header.gas_used is structurally blind to it


pirapira:
The guest **under-charges exactly 570 gas of regular execution gas** on one transaction, and the check that
should catch it is **structurally blind** to it. Confirmed false reject: execution-specs accepts the block.

Fully characterised below, with three candidate mechanisms eliminated by measurement and one confound named.
Filed rather than fixed because the remaining step needs a fixture I do not have.

## The row

* `blockchain_tests/for_amsterdam/frontier/scenarios/scenarios.json`,
  `test_scenarios[fork_Amsterdam-blockchain_test-test_program_program_GASPRICE-debug]`, **block 18**
* sampled from `main` `d4215e262`, random seed `146913243`, `--backend spike`
* `bv_fail = 53` (receipts root)
* **input pinned by byte-identity** — 11000 bytes, sha256 `7863341f…`; the header decodes with `number = 18`
  and a 20-byte coinbase

⚠️ Do **not** pin this block by `gasUsed`: **24 of the 37 blocks in that case share `gasUsed = 0x017e80`.**
It is the modal value and matching it discriminates nothing.

## Confirmed false reject

`scripts/eest-stateless-to-input.py --verify-run-stateless-guest` reproduces the fixture's
`statelessOutputBytes` for **all 37 blocks** of that case, b18 included. The fixture is self-consistent and
**the spec accepts the block our guest rejects**, so this is a place where our behaviour is *narrower* than
the reference.

## The divergence, attributed to one dimension

Spied on the spec's `process_transaction` and read the `block_output` accumulator deltas for the single tx:

| accumulator | spec | guest | delta |
|---|---|---|---|
| `block_state_gas_used` | 97920 | 97920 | **0** |
| `block_gas_used` (regular) | **84369** | **83799** | **−570** |
| `cumulative_gas_used` (receipt) | 182289 | 181719 | −570 |

The spec's own encoded receipt carries `8302c811` = 182289, so 570 is the error in the value that **enters the
hash**, not an intermediate.

⇒ **`before_refund` is 570 low in the guest.** `before_refund = tx.gas − gas_left − state_gas_left`, so the
guest leaves 570 more gas than the spec: it **under-charges during execution**. Not intrinsic, not state, not
refund, not the floor.

## ⚠️ Why nothing caught this: the header check is structurally blind

[...]

PR #11149, opened 2026-08-02
fix(seed): skip system_call_mode=1 in dispatcher_seed_pending_value_transfer

pirapira:
## Summary

- **Issue:** #11148 (cause 2 of #11115 — two defects under one `bv_fail=44`)
- **Fix:** `dispatcher_seed_pending_value_transfer` skips `system_call_mode=1`
- **One file:** `EvmAsm/Codegen/Programs/EIP7708Logs.lean` (no layout regen)

## Justification (load-bearing)

Skipping mode=1 is correct **because the spec transfers no value on any system call**.

Amsterdam `execution-specs` `fork.py` `process_unchecked_system_transaction` sets:
- `value = U256(0)`
- **`should_transfer_value = False`**

Every system caller uses that helper (beacon root, history storage, withdrawal/consolidation/builder requests). Spec records **no** `move_ether` for system messages.

Therefore the guest must not publish a value-transfer nonstorage effect on `system_call_mode=1`. Shape A (skip) matches that property of the spec — not a corpus observation.

**If a future fork makes a system call value-bearing, this skip becomes wrong** and must be revisited (the citation is the trigger). Shape B (correct env base) would still *publish* a transfer where the spec performs none, and would still risk stale `bv_pending_recipient_*` left from the user message — so A is closer to the spec's shape, not merely equivalent.

## Defect (measured)

On FAIL 00002 / cid `915e96f535cfbf3a` (`selfdestruct_to_precompile_state_access_boundary`, same_tx), seed schedule on main guest:

| # | mode | env_base | raw count | outcome |
|---|------|----------|-----------|---------|
| 1 | 0 user | 0xc8 | 1 | CALLVALUE=0 → rmv skip |
| 2 | 1 system | 0x368 | 11 | skip |
| 3 | 1 system | **0x310** | 11→13 | **invented factory credit** |

Mode-1 still held user `runtime_tx_auth_sender_ptr` + `bv_pending_recipient_addr=factory` while `srpc_env_base` pointed at the system out-buffer. Reverse-copy of `bv_runtime_payload + srpc_env_base + 127` produced CALLVALUE = sender gas debit `485029388215221`. `record_message_value_transfer` wrote factory bal `1 → 485029388215222` → `bal_all_accounts_nonstorage_consistent` **bv_fail=44**.

Precompile beneficiary post=2 agreed both sides; the huge value was on the **factory**.

### Family `selfdestruct_to_precompile_state_access_boundary` limit40

| leg | guest sha | result |
|-----|-----------|--------|
| main | `674d359d…` | 32 pass / 8 fail all bv44 |
| **fix** | `5ee37f6f876fe1e32c25e2eb855774aee5af4e134cac707d1a62c3a2b9692f1f` run-20260802T011134Z-3800931 | **32 pass / 8 fail** all remaining bv44 |

**Family staying 32/8 is expected.** The fail rows still hit cause 1 (#11147: `destroyed_count` wiped on system re-entry → CREATE child residual). This PR **cannot** produce a forward flip while cause 1 still fires; absence of forward flips is not a weakness.

SPIKE evidence that this PR does its job (FAIL 00002 after fix):
- raw `exec_nonstorage_effect_count` **13 → 11** (exactly the phantom pair removed)
- factory `6c0250ae…` agg now **bal 1→0 nonce 1→2** (huge gone)
- residual only CREATE child `bf49ad3b…` bal 0→1 nonce 0→1 with `destroyed_count=0` → #11147

### limit200 FR (quadruple + reverse flips)

*(updating when runs complete — reverse flips are the load-bearing number: removing a publish risks dropping a record that should exist)*

## Out of scope

- #11147 destroyed_count wipe (cause 1) — separate PR; two causes under one fail code get two PRs
- #11115 stays open until both green
- No 7702 / delegation

## Test plan

- [x] family filter limit40 — 32/8 expected; mechanism confirmed by SPIKE
- [ ] limit200 FR vs main; reverse flips = 0 required to ship
- [x] SPIKE factory huge gone / residual is #11147 only


pirapira:
**Adjacency note (not a claim of relation):** this change is in `EIP7708Logs.lean` (`dispatcher_seed_pending_value_transfer`). Scoper filed #11145 tonight for a 570-gas under-charge on a tx carrying a 7708 transfer log; scoper established the log was **confounded** with a value-bearing inner call and could not be separated. Do not assume related — flagging so whoever picks up #11145 knows this landed underneath it.

pirapira:
## Fresh FR after layout regen (coord-ordered)

Post-pin commit `7327e4285` (textSizeBytes 0x63cd0→0x63cd4 fixpoint). In-flight pre-regen measure discarded.

| leg | commit | guest sha256 | run | RAN | full | fail | error | budget |
|---|---|---|---|---|---|---|---|---|
| main | `5764c94f4` | `8517ef20ed584a463cc05d5fdea01627df05ce844ae016a7a0bfd60eff215a86` | `run-20260802T013803Z-3963183` | 200 | 174 | 26 | 0 | 0 |
| fix | `7327e4285` | `5ee37f6f876fe1e32c25e2eb855774aee5af4e134cac707d1a62c3a2b9692f1f` | `run-20260802T013712Z-3958184` | 200 | 174 | 26 | 0 | 0 |

`eest-ab-compare.py` main→fix:
- joined 194, unmatched 0
- **FA 0→0, FR 26→26, FR label-set equality PASS**
- **output-byte differences: 0** (BYTE-IDENTICAL)
- **reverse flips: 0** (load-bearing for a removal)

### System-call exercise count (not a null from non-exercise)
SPIKE_WATCH on `system_call_mode` @ `0xbd097510` across all 200 fix-leg inputs:
- **181/200 (90.5%)** entered mode=1 (typically 4 enters/row = deferred system path after body)
- 19/200 never entered: early intrinsic/floor/oog rejects that never reach deferred system staging (e.g. calldata_floor, intrinsic_gas_floor_boundary, set_delegation_oog_charge)
- sum mode=1 enters = 724; timeout 0

Zero reverse under 90% system-call exercise is a real null, not non-exercise.

### Family residual (expected while #11147 open)
`selfdestruct_to_precompile_state_access_boundary` limit40 still 32/8 all bv44 — cause 2 factory huge removed (SPIKE raw 13→11); residual is cause 1 destroyed_count wipe only.

pirapira:
Labelled — all six checks green.

**The exercise count is why this is conclusive rather than merely clean.** A/B on `5764c94f`: byte-identical, label-set equal, FA 0→0, FR 26→26, **reverse flips 0**, errors and budget 0 on both legs. On its own that reads like a null. But `SPIKE_WATCH` on `system_call_mode` establishes that **181 of 200 rows enter mode 1** (724 total enters, ~4 per row), and the 19 that don't are early floor/oog rejects that never reach the deferred system stage.

That distinction decided two PRs tonight. #11142's first A/B was byte-identical over 200 rows of which **4** exercised the change — nearly worthless until a targeted slice took it to 15/40. Same headline, opposite evidential weight. Explaining the complement, not just counting the hits, is what makes this one hold.

For a change that **removes** a publish, the failure mode is a record that should exist disappearing. Zero reverse flips across a 90.5%-exercising population is the direct test of that, and it comes back clean.

**The justification is a spec property, not a corpus observation** — the part I most want copied. `process_unchecked_system_transaction` sets `value = U256(0)` and `should_transfer_value = False`; every system caller (beacon root, history storage, withdrawal/consolidation/builder requests) routes through it, and the spec records no `move_ether` for system messages. The PR then names its own expiry condition: *if a future fork makes a system call value-bearing, this skip becomes wrong, and the citation is the trigger*. That is a load-bearing property documented at the point where someone would break it.

It also argues against the alternative on the merits rather than by preference. I had steered toward correcting the env base (shape B) as "strictly safer"; the author showed B would still **publish a transfer where the spec performs none**, and would risk a stale `bv_pending_recipient_*` left from the user message. Shape A is closer to the spec's shape, not merely equivalent — my steer was wrong.

**Honest partial-fix framing**: FR unchanged at 26 is *expected*, because cause 1 (#11147) still fires on this family, so this PR cannot produce a forward flip. Two causes under one `bv_fail=44` get two PRs, so a regression stays attributable. The residual is #11152, filed separately rather than hung off #11147 as a rider.

Root cause was a wrong-base read: `srpc_env_base` (system) applied to `bv_runtime_payload` (user), which is why the "corrupt" value 485029388215221 turned out to be the sender gas debit — a real quantity from the wrong offset, not garbage. That reclassification is what made the fix a one-line-class change instead of a buffer-initialisation hunt.

Issue #11146, opened 2026-08-02
BAL: delegated top-level target at precompile 0x04 is missing a pure-touch entry


pirapira:
## Current-main witness

On current `origin/main` `9e0cda12d41b8085633f8697ae4c14062de9bc1f`, the fresh 200-row run (`/home/zksecurity/evm-asm33/gen-out/eest-run/run-20260802T004228Z-3621990`) contains a valid false reject:

- fixture `00172_test_top_frame_charges_delegation_is_precompile`, input SHA-256 `1459b9c7d729b11701508fd7e09ea870d7ede7def80ba3e7c8badfcb6e7f8d8b`
- guest ELF SHA-256 `674d359d0e4ca37d04c45f7d3843deb899e840d9d8d9c87129d4e231baba5460`
- Spike runner SHA-256 `3c3b0bbbe7a1017d9d75908b2dc60ed508bb82cf05882061130a6424354405ff`
- oracle expects success; current guest reports `bv_fail=60`, `shadow_status=1`, rebuilt/supplied BAL lengths `337/364`

A structural decode of the two BAL streams shows the rebuilt stream has the same entries as the declared stream except for one all-empty account entry: address `0x0000000000000000000000000000000000000004`. This is a shortfall, not an equal-length substitution.

The transaction is a legacy CALL to `0xf0f7b798b6eb9449c0a3ceb834f8c490705e3978`, whose EIP-7702 delegation designation resolves to precompile address `0x04`. The zero-value sibling `00173` is a control: it is `337/337` and passes on the same current-main ELF.

## Spec-side requirement

`execution-specs` pinned at `e5a8caf1b8055e4d805c7fb169edfa710914b7da`, `utils/message.py:56-71`, computes the delegated target for a top-level call and unconditionally adds `current_target` to `accessed_addresses` before execution. A delegated target that resolves to a precompile therefore remains a touched account even when it has no balance, nonce, code, or storage changes.

## Guest-side lead (not yet proven)

[...]

PR #11151, opened 2026-08-02
BAL: record deferred top-level delegation target touch

pirapira:
## Summary

`prepare_message` unconditionally adds the resolved top-level target to `accessed_addresses` before execution (`execution-specs` pinned at `e5a8caf1b8055e4d805c7fb169edfa710914b7da`, `src/ethereum/forks/amsterdam/utils/message.py:56-71`). The prior-block deferred delegation materializer charged/warmed `dtrc_deleg_target` but never called `account_read_record`, so an all-empty delegated target could be absent from the BAL.

This patch records `dtrc_deleg_target` immediately after `delegationAccessChargeAsm`, unconditionally and before the materializer continues. It follows the emitted path actually taken by the witness (resolver status 0 -> `dtrc_materialize_deferred_delegation`), not the earlier status-2 hypothesis.

## Evidence

- Base: `origin/main` `5764c94f41681aa60443dc617e845d09f19429e3`.
- Final ELF (own checkout /home/yoichi-bkp/evm-asm3): `gen-out/regionmap/stateless_guest.elf`, SHA-256 `8668c04be5c2a1ef42cebc8734d5f5d3a3826a0ea8bac9ddc9e885797e8a5ea7`.
- Runner: `/home/zksecurity/evm-asm2/scripts/spike/spike_run`, SHA-256 `eaf534e5017e71ffe2dbc17e815646d46ea2ee024f5e8890b7901aaa6ec84d9f`.
- Fresh final-ELF pair: 00172 FR -> OK; 00173 remains OK; `FA=0`, `FR=0`, `OK=2`. Both ran with a 256-byte output window.
- Expected-invalid control 00015 still rejects (`oracle_succ=0`, output success byte 0 at offset 32).
- The fixpoint-linked artifact is pinned at `.text=0x063cdc`, `.data=0x5370`, `.bss=0x1c101b80`; `check-asm-to-program`, `check-region-map`, `check-build-parallel`, and full `LAKE_ARTIFACT_CACHE=false lake build` pass.

A full sweep was intentionally not repeated because full runs are expensive; the targeted witness pair plus the expected-invalid control exercises the changed route and the soundness floor.

## Random-200 A/B

Fresh parent/candidate legs on the same 200-row manifest (random sample from the frozen EEST inputs), using the same runner and a 256-byte output window:

- Parent ELF (built in own checkout while detached at base) `/tmp/11146-deferred-target-touch/parent.elf`, SHA-256 `bbd67c0d1e26be1056a4f2bb760fd03a2ac5d1cbde31a9e73b446b91a96e53ac`: oracle-valid 190, guest pass 179, `FA=0`, `FR=11`; expected-invalid 10 all reject.
- Candidate ELF (copied from own checkout final artifact) `/tmp/11146-deferred-target-touch/candidate-final.elf`, SHA-256 `8668c04be5c2a1ef42cebc8734d5f5d3a3826a0ea8bac9ddc9e885797e8a5ea7`: oracle-valid 190, guest pass 179, `FA=0`, `FR=11`; expected-invalid 10 all reject.
- Paired transitions: forward 0, reverse 0; every one of 200 rows has the same verdict on both legs.

The mechanism addressed is the downstream omission in `dtrc_materialize_deferred_delegation`: the current-main witness path resolved with status 0 and entered this materializer, which performed delegation access charging/warming but did not record `dtrc_deleg_target` in the tracked account-read set. The status-2 callback-bypass hypothesis was not the path taken by the witness; the emitted trace established the materializer path directly. The 00172 flip and zero-transition random-200 are consistent with that precise omission.

### Changed-path exercise count

The checked-in SPIKE write-watch driver was built in `/tmp` (watch runner SHA not used for verdicts) and pointed at the emitted `runtime_tx_post_top_frame_fn` cell `0xbcdc9b90`; a write of `0x80022210` is the deferred-materializer selection. Across the same 200-row manifest, exactly **1/200** selected `dtrc_materialize_deferred_delegation` (row `00180_test_top_frame_regular_charge...`); the complement **199/200** did not. The targeted 00172/00173 pair is the separate exercising slice: both were traced into the status-0 materializer path and 00172 is the FR -> OK witness. Thus the random-200 zero-transition result is meaningful for the 199 non-exercising rows, while the targeted slice supplies the changed-path exercise.

### Path-derived delegated-target slice

To test the over-recording direction on the path that can actually reach this change, I decoded the same `/var/tmp/fc668/manifest.tsv` with the pinned execution-specs decoder and witness state. I selected rows whose top-level transaction has a non-creation `tx.to` whose pre-state account code begins with the `0xef0100` delegation designation; this is a path predicate, not a fixture-name filter. Of 26,104 manifest rows, 25,045 decoded for this filter and 1,059 malformed/unsupported rows raised a decoder or witness-trie exception; 48 rows matched the predicate.

The write-watch over those 48 rows selected the emitted deferred materializer on **38/48** (the other 10 did not enter it; all 48 completed with `rc=0`). On the same 48-row manifest, with the same 256-byte output window and runner, the parent was `FA=0, FR=32, OK=16` and the candidate was `FA=0, FR=6, OK=42`: **26 forward flips and 0 reverse flips**. This is the targeted over-recording check; the 1/200 random-200 exercise count remains reported separately above.


pirapira:
Labelled — build green, and the targeted slice turns this from a witness into a result.

**The slice was built from the path, not from a family name.** Scanning all 26,104 manifest rows, decoding `tx.to` and pre-state code through the execution-specs witness, and selecting accounts whose code begins `ef0100` (the delegation marker) yields **48 rows** — with 25,045 decoded and 1,059 decoder/trie exceptions excluded rather than silently dropped. That construction is what makes the population mean something: it selects on *the condition that reaches the changed code*, not on fixture naming, which has misled this fleet repeatedly tonight.

**Exercise count**, `SPIKE_WATCH` on `runtime_tx_post_top_frame_fn` (`0xbcdc9b90`), materializer selection `0x80022210`: **38 of 48 rows actually select `dtrc_materialize_deferred_delegation`**, 10 do not, all `rc=0`.

| 48-row slice | parent | candidate |
|---|---|---|
| FA | 0 | 0 |
| FR | 32 | **6** |
| OK | 16 | **42** |

**26 forward flips, 0 reverse.**

**Why this was worth waiting for.** The random-200 A/B showed all 200 verdicts unchanged — but only **1 of 200** rows reaches the materializer, so 199 of them could not have been affected whatever the change did. That is very close to the null a no-op produces. This change *adds* a read record, so its failure mode is over-recording, which can only appear on rows that enter the materializer; a sample where 199/200 never enter it is structurally incapable of detecting it. Zero reverse flips there was a fact about the sample, not about the fix. Zero reverse flips across 38 exercising rows is a fact about the fix.

**The mechanism was discriminated, not inferred.** The issue (#11146) kept two candidates explicitly untested — resolver status-2 bypass vs a downstream drop. The emitted trace showed resolver **status 0** entering `dtrc_materialize_deferred_delegation`, which charges and warm-seeds but had no `account_read_record`, and that the status-2 callback bypass is **not** the path these rows take. One arm confirmed by the route actually executed, the other eliminated by it.

This is the correct-but-unfed shape — the fourth tonight: the mechanism exists and does everything else right, and the consumer never receives the signal.

**Provenance** is own-checkout throughout (`/home/yoichi-bkp/evm-asm3`), parent built in the same checkout detached at `5764c94f`. An earlier witness for this finding was measured against an ELF from another agent's tree and was correctly demoted to *discovery, not provenance* rather than being quietly reused. `00015` expected-invalid still rejects on both legs.

Issue #11152, opened 2026-08-02
runtime_dispatcher_call setup: 15 journals zeroed on system re-entry still unproven (post-#11147 residual)


pirapira:
## Context

Parent cause-1 for `selfdestruct_to_precompile` bv44 is **#11147**: `evm_selfdestruct_destroyed_count` (and overflow) are wiped at every `runtime_dispatcher_call` entry, including system re-entry after the user body (`stage_system_call` → mode=1). That pair is proven MUST-SURVIVE because `nonstorage_effect_aggregate` destroyed-norm runs after `block_verdict_deferred_system_requests`.

The same setup (`emitRuntimeDispatcherCallableSetup`, `Dispatch.lean` ~2950-2976) zeros **sixteen** journals on every callable entry. #11147 will fix **only** the proven pair. This issue owns the other fifteen.

## Question (checkable)

For each of the fifteen cells below: does any consumer read it **after** a system re-entry `runtime_dispatcher_call` within the same block, using a value that the **user** body wrote?

Positive answer → MUST-SURVIVE (file fix, preferably one cell or tight cluster per PR).
Negative answer requires exhaustive consumer census + timeline, not "I didn't find a reader".

## The fifteen (still zeroed on system re-entry after #11147 lands)

| # | symbol | notes from partial census |
|---|---|---|
| 1 | `evm_refund_acc` | settle may already consume before deferred system — reason, not proof |
| 2 | `evm_selfdestruct_staged` | |
| 3 | `evm_selfdestruct_seen_count` | |
| 4 | `evm_selfdestruct_seen_overflow` | |
| 5 | `cd_destroyed_empty_hits` | |
| 6 | `create_nonce_table_count` | |
| 7 | `create_nonce_table_overflow` | |
| 8 | `create_nonce_undo_count` | |
| 9 | `account_state_pending_count` | |
| 10 | `account_state_created_count` | |
| 11 | `account_state_delete_count` | |
| 12 | `account_state_overflow` | |
| 13 | `evm_log_data_used` | snapshot-before-next-dispatch is a design claim |
| 14 | `evm_log_data_overflow` | same |
| 15 | (pair slots / related if any left) | treat overflow pairs with their counts |

Conditional `exec_code_effect_*` (kept when `code_state_mtx_active`) is out of scope unless that gate is wrong for system re-entry.

## Method

1. Enumerate every reader of each cell in emitted guest asm.
[...]

PR #11153, opened 2026-08-02
fix(dispatch): preserve destroyed_count across system re-entry

### LINKED ISSUE #11147: runtime_dispatcher_call setup zeroes evm_selfdestruct_destroyed_count across system_call re-entry

pirapira:
## Summary

Fixes #11147 (cause 1 under #11115 bv44): `emitRuntimeDispatcherCallableSetup` was zeroing `evm_selfdestruct_destroyed_count`/`overflow` on **every** `runtime_dispatcher_call` entry, including system re-entry after the user body (`stage_system_call` sets `system_call_mode=1` then jals the callable). Aggregate destroyed-table norm then never saw CREATE+SD victims → residual nonstorage rows vs BAL empty → bv_fail=44.

**Narrow cut (coord):** preserve **only** `destroyed_count` + `destroyed_overflow` when `system_call_mode ≠ 0`. The other 15 journals still zero every entry. Residual journal classification stays on #11152. Standalone setup path (true new dispatch) unchanged.

```lean
-- Dispatch.lean emitRuntimeDispatcherCallableSetup
la t0, system_call_mode; ld t0, 0(t0)
bnez t0, .Lrtdc_destroyed_kept
sd zero, destroyed_count
sd zero, destroyed_overflow
.Lrtdc_destroyed_kept:
```

## Spec / mechanism

- Spec keeps created-in-tx destroy marks through finalize so BAL nonstorage aggregate can drop CREATE+SD child components.
- Guest: set at `h_SELFDESTRUCT`; was cleared at `runtime_dispatcher_call` setup (`pc≈0x8003600c`) with `ra=stage_system_call` before deferred system txs (#11114 seam).
- SPIKE 00002 cid `915e96f535cfbf3a`: post-fix **`destroyed_count=1` at `nonstorage_effect_aggregate` and `bal_all_accounts_nonstorage_consistent`** (was 0 after system wipe).

Sibling #11149 (cause 2 seed skip mode=1) is separate and already on main. This PR is cause 1 alone.

### 1. Family forward flip — main WITH #11149 as baseline

Control structure: baseline already contains cause-2 so this PR is attributed cleanly.

```
--filter selfdestruct_to_precompile_state_access_boundary --limit 40
```

| leg | commit | guest sha256 | run | result |
|---|---|---|---|---|
| main (#11149 only) | `da972a109` | `8cff12fe6630d8b1…b28748d8` | `run-20260802T024259Z-195624` | **32 full / 8 fail** all bv44 |
| fix (#11149+#11147) | `0d4833e04` | `4db0f3365f93c327…5d0c6158` | `run-20260802T024147Z-191465` | **40/40 full** |

**32/40 → 40/40.** All eight residual rows close. Two independent causes under one bv_fail=44; each necessary, neither sufficient. Family only went green when both landed.

### 2. destroyed_count survives to the comparator

SPIKE 00002 (cid `915e96f535cfbf3a`): after user body + system re-entry, `destroyed_count=1` at both `nonstorage_effect_aggregate` (`0x80044bdc`) and `bal_all_accounts_nonstorage_consistent` (`0x8002c958`). Pre-fix was 0 after the wipe.

### 3. System-call exercise (SPIKE_WATCH `system_call_mode@0xbd097510`)

On fix guest, all 200 limit200 inputs:

- **183/200 (91.5%)** write mode 0→1 at least once (sum 732 mode1 writes, avg 4.0/row — multi system-call staging exercised)
- **17/200 none** (complement): early floor/OOG reject before system staging (`calldata_floor`, `intrinsic_gas_floor_boundary*`, one `set_delegation_oog_charge_point`) — same class as #11149's 19/200

### 4. Null reverse on properly exercised population

| leg | commit | guest sha256 | run | full/fail |
|---|---|---|---|---|
| main | origin/main `2f973a160` | `1898b1c9ed2607c0…01da9e95` | `run-20260802T021548Z-41419` | **176/24** |
| fix | `e699cd674` | `7c1117d837623099…ad889fe8` | `run-20260802T021654Z-48390` | **176/24** |

`eest-ab-compare.py`: FA 0→0, FR 24→24, **label-set EQUAL**, **output-byte differences 0** on 194 joined, reverse flips **0**.

Reverse risk inverted vs #11149 (preserve counter, not remove publish): null reverse under 183 multi-system enters is the anti-stale-count check. No reverse on multi-tx/multi-system rows in this pop.

## Layout

textSizeBytes `0x063e34 → 0x063e44` (+16B gate); after merge main + regen fixpoint → **`0x063e48`**. GuestAddrs + symbol-addresses.tsv regen. `check-region-map` clean.

## Out of scope

- #11152 residual 15 journals (parked, not absorbed)
- #11148/#11149 cause 2 (merged separately)
- #11115 closes once both causes are on main
- No 7702/delegation
- No merge! from author

## Test plan

- [x] family selfdestruct_to_precompile limit40: main+#11149 32/8 → fix 40/40
- [x] limit200 A/B vs main → byte-identical, reverse 0
- [x] SPIKE_WATCH system_call_mode exercise 183/200 (17-row complement)
- [x] SPIKE destroyed_count=1 at aggregate/comparator on 00002


pirapira:
## Family forward flip (requested before label)

Main **with #11149 already merged** (`da972a109`) vs fix (cause1 + #11149):

| leg | guest sha256 | run | family limit40 |
|---|---|---|---|
| main (#11149 only) | `8cff12fe6630d8b1…b28748d8` | `run-20260802T024259Z-195624` | **32 full / 8 fail** all bv44 |
| fix (#11149+#11147) | `4db0f3365f93c327…5d0c6158` | `run-20260802T024147Z-191465` | **40/40 full** |

**OUTCOME 1:** family improves. Both causes independent; decomposition vindicated. #11149 alone leaves the 8 residual (destroyed_count wipe); cause1 closes them.

Also merged origin/main into PR + layout regen fixpoint (`textSizeBytes` → `0x063e48`, check-region-map clean) so CONFLICTING should clear after CI. No merge! from author.

pirapira:
Labelled — build green.

**This PR completes a two-part decomposition, and the family result is the proof it was real.**

| `selfdestruct_to_precompile` family | result |
|---|---|
| main **with #11149 already merged** (`8cff12fe`) | 32 full / 8 fail, all `bv_fail=44` |
| with #11147 on top (`4db0f336`) | **40 / 40 full** |

All eight residual rows close. Baselining against main *with #11149 in it* — rather than pre-#11149 main — is what isolates this fix cleanly; any other baseline would have conflated the two causes.

**Why that matters more than the row count.** Hours earlier this family sat at 32/40 behind a single `bv_fail=44`. The cheap move was one broad fix and a claim on the family. Instead the author decoded per entry, separated **two causes with different mechanisms** — a wrong-base read (`srpc_env_base` applied to `bv_runtime_payload`, which is why the "corrupt" 485029388215221 was really the sender gas debit) and a counter wiped at system re-entry — filed them as #11148 and #11147, and shipped them as two PRs. The family went green **only when both landed**, which is exactly what the decomposition predicted and what a combined patch could never have demonstrated. Two independent causes, each necessary, neither sufficient.

**Mechanism observed directly**: `destroyed_count = 1` at the comparator after system re-entry, where it was previously 0. That is the thing the fix exists to change, not an inference from a verdict bit.

**Reverse-risk properly powered for an inverted profile.** #11149 *removed* a publish, so its danger was a missing record. This one *preserves* a counter that was being cleared, so the danger is a stale value surviving into a context that should start clean — visible on multi-system-call and multi-tx rows. `SPIKE_WATCH` on `system_call_mode`: **183/200 rows enter mode 1** (732 writes, ~4.0 per row, so multi-call staging is genuinely exercised), with the 17-row complement explained as early floor/OOG before system staging. Reverse flips **0** across that population.

**Scope held narrow against the evidence.** The clobber census found **16 journals** zeroed on every call including system re-entry. Only `destroyed_count` + overflow were *proven* must-survive, so only those two are preserved here; the other fifteen still zero exactly as before. Twelve were explicitly marked **unproven rather than OK-to-wipe**, and the residual is filed as **#11152** with its own closing condition rather than absorbed as a rider. A narrow fix that leaves a real problem visible beats a broad one that hides it — "don't zero on system re-entry" would have been one condition covering sixteen cells on evidence for two.

Layout regenerated to fixpoint after merging main (`textSizeBytes` `0x063e34` → `0x063e44`), not hand-set. #11115 stays open until both causes are on main.

PR #11155, opened 2026-08-02
fix(mtx): resolve sender pre-state before upfront staging

pirapira:
## Summary

The MTx sender staging path treated a missing nonce bit in a cumulative, balance-only account row as a total sender-account miss. For a sender funded by an earlier transaction, this skipped the current transaction's gas/blob upfront debit even though the sender account was present in cumulative state.

This change makes the consumer use the existing `account_resolve_pre_state` routine. It does not add a producer or a new state-resolution strategy: the resolver consults cumulative account writes first, then the authenticated parent state, and returns the balance-only row with the absent parent account's zero nonce.

The execution-specs rule is `_get_pre_tx_account` at `execution-specs` commit `e5a8caf1b`, `src/ethereum/forks/amsterdam/block_access_lists.py:583-600`: look up the address in `pre_tx_accounts` (state through, but not including, the current transaction), then fall back to `pre_state.get_account_optional`. The underlying transaction reader at `src/ethereum/forks/amsterdam/state_tracker.py:199-202` records the read, consults transaction-local writes, and then consults pre-state. Sender nonce/balance validation is the corresponding `fork.py:656-667` check. The producer-versus-consumer diagnosis and its superseding correction are recorded on [#10786](https://github.com/Verified-zkEVM/evm-asm/issues/10786#issuecomment-5154840061).

## Evidence

- Targeted witnesses `00005` and `00017`: 2/2 exact after the change.
- Random-200 A/B, seed `20260802`, fixture tag `tests-zkevm@v0.6.2`, both legs scored 200/200 with zero emulator errors and zero budget classifications.
- Baseline parent: commit `dc48c0e7ccdd84831793eed254278a1124087a94`, ELF `/tmp/easm-current-main-guest-dc48.elf`, SHA256 `32457060160307310a8cc55e105d6f3948668b32c6035b933f250ce45acb6133`: FA=0, FR=19, OK=181.
- Candidate: source commit `339f53668` plus fixpoint layout regeneration commit `000a51474`, ELF `/tmp/easm-main-current/gen-out/regionmap/stateless_guest.elf`, SHA256 `e551d54cf0d9a26bdc92e8d4e3924777ba4c19320d3af7a9449afe60be7f7f50`: FA=0, FR=14, OK=186.
- Delta on the regenerated paired run: five forward flips and zero reverse flips. Forward labels: `00005_test_sufficient_balance_blob_tx_pre_fund_tx_fork_Amsterdam-max_blobs-blockchain_test-sender_init`, `00017_test_sufficient_balance_blob_tx_pre_fund_tx_fork_Amsterdam-single_blob-blockchain_test-sender_in`, `00125_test_sufficient_balance_blob_tx_pre_fund_tx_fork_Amsterdam-max_blobs-blockchain_test-sender_init`, `00157_test_sufficient_balance_blob_tx_pre_fund_tx_fork_Amsterdam-single_blob-blockchain_test-sender_in`, `00187_test_sufficient_balance_blob_tx_pre_fund_tx_fork_Amsterdam-max_blobs-blockchain_test-sender_init`. Reverse list: empty.
- The candidate ELF changed during fixpoint regeneration (`ad9f2d83…` to `e551d54c…`), so both legs were rerun rather than reusing the earlier A/B. The regenerated candidate has `.text=0x063e1c`, `.data=0x5370`, `.bss=0x1c101b80`; `bnf_le_a` remains `0xa3000ee0`. Non-cache `lake build`, `scripts/check-build-parallel.sh`, and `check_guarded_handler_bytes.py` all pass.
- `scripts/eest-ab-compare.py` passed its denominator, digest-join, duplicate-group, and distinct-ELF checks: 195 matched input digests (five duplicate groups covering ten rows), no unmatched rows, and no status regressions.
- Runner for both legs: `/home/zksecurity/evm-asm2/scripts/spike/spike_run`, SHA256 `eaf5345017e71ffe2dbc17e815646d46ea2ee024f5e8890b7901aaa6ec84d9f`.
- Exercise split in the same 200-row manifest, decoded from the framed inputs with execution-specs: 187 transaction-bearing rows take the MTx sender loop (histogram `tx_count=0:13, 1:149, 2:27, 3:5, 4:4, 5:2`); the complement is 13 no-transaction system/withdrawal rows and does not enter sender staging.

The candidate remains unlabelled; merge labeling is left to coordination.


pirapira:
Labelled — build green on retry (the first attempt died in the installer: `lean-action` could not resolve `release.lean-lang.org`, so build and test never ran; that is not a verdict on the change).

**This closes the dominant defect on current main.** A fresh 30-row sample reduced to **2 FRs, both the same mechanism** — twice, across a base change — so this is not one bug among many.

| random-200, seed 20260802, base `dc48c0e7c` | parent `32457060` | candidate `e551d54c` |
|---|---|---|
| FA | 0 | 0 |
| FR | 19 | **14** |
| OK | 181 | **186** |
| errors / budget | 0 / 0 | 0 / 0 |

Forward flips: **00005, 00017, 00125, 00157, 00187**. Reverse: **0** (stated, not inferred). Exercise: **187/200** tx-bearing rows take the MTx sender loop, histogram `{0:13, 1:149, 2:27, 3:5, 4:4, 5:2}`, complement of 13 explained as no-tx system/withdrawal rows.

**The forward set outgrew its witnesses.** 00005 and 00017 were the decoded rows; 00125, 00157 and 00187 came along because they share the mechanism, not because they were targeted. That is the signature of a root-cause fix rather than a row fix — the same defect was sitting under three rows nobody had looked at.

**The repair is delegation, not a new producer.** The guest now reaches `account_resolve_pre_state` (block-cumulative `account_writes`, then parent state), which corresponds to the spec's `_get_pre_tx_account` — `state_tracker.py:199-202`, `block_access_lists.py:583-600`, `fork.py:656-667` at `e5a8caf1b`. Making the deficient path reach an existing working routine is the shape our doctrine asks for; adding a recipient producer would have recreated the duplication that the original unification removed.

**Why that matters here specifically**: the archaeology showed `6601a57da` added `blockVerdictMtxEoaRecipientCredit`, `bb22567a9` removed it while routing recipients through `dispatch_tx_runtime_code`, and `d93877353` completed the shared routing — a unification that dropped a step. The repair therefore belongs on the shared path.

**The route to this fix is worth more than the patch.** The diagnosis moved through five framings in one hour — missing producer, branch-target skip, partial balance-only row (flags 35, bit32, no bit64) with a nonce-gating consumer, the three-commit archaeology, and finally a **disposable candidate** that answered producer-vs-consumer empirically. Each was killed by the next measurement and **none shipped**. Implementing at framing one or two would have produced a green PR burying the real defect underneath it.

**Provenance held through two hazards.** An earlier witness was measured against another agent's tree and was demoted to *discovery, not provenance*. Then the fixpoint regen changed the ELF (`ad9f2d83` → `e551d54c`), so the random-200 was **re-run rather than carried over** — my prediction that a pin regen leaves the image byte-identical was wrong, and the author measured instead of taking it.

pirapira:
## Post-merge random-200 baseline (measurement only)

Run completed on current `main` after the #11153/#11155 merges.

- **repo commit:** `703336c28a2b7c2f240d77f104e96e6281112d73`
- **scope:** 200 uniformly sampled blocks
- **seed:** `4242`
- **backend:** Spike, `--jobs 8`
- **fixture version:** `tests-zkevm@v0.6.2` (`fixtures_zkevm.tar.gz`, 497,764,775 bytes)
- **guest ELF:** `/home/yoichi-bkp/evm-asm-codex2-main-baseline/gen-out/eest-run/post-merge-random200-4242/stateless_guest.elf`
- **guest ELF SHA-256:** `c16ff50d559342d5f177db12dbc2afe002519397951f9ee512d61b802e7441a4`
- **runner:** `/home/yoichi-bkp/evm-asm-codex2-main-baseline/scripts/spike/spike_run`
- **runner SHA-256:** `5b383a7702fb05dea16a2cc37dc313b82491fdf194f39c12306cb387a5a2b6b0`
- **run directory:** `/home/yoichi-bkp/evm-asm-codex2-main-baseline/gen-out/eest-run/post-merge-random200-4242`
- **build mode:** guest built in the detached current-main worktree; no `--no-build` or guest override; worktree clean

### Result

```
FA=0  FR=8  errors=0  budget=0
full=192/200  root=200/200  succ=192/200  tail=200/200
```

All eight mismatches were false rejects (guest success bit 0 vs expected 1); there were no false accepts, emulator errors, or step-budget classifications. This is a baseline record only; the FR rows were not investigated.

Issue #11162, opened 2026-08-02
#11138's tier-3 classifier over-reports: status 1 (account not in state trie) is spec-legitimate but counted as an integrity failure


pirapira:
⛔ **My own defect, in the classifier I added in #11138.** Found by the counter firing during the #10874 decisive measurement (comment `5156518356`) — the counter worked, and reading the convention showed the *counter* was wrong, not the walk.

## The convention

`EvmAsm/Codegen/Programs/StateCompose.lean:552-561` documents `slot_at_header_state_root`'s return as a **flat unified enumeration**:

```
0 = found + decoded
1 = account not in state trie
2 = state-trie mpt parse error
3 = account_decode failure
4 = header parse / state_root size fail
5 = slot not in storage trie
6 = storage-trie mpt parse error
7 = slot RLP decode failure
```

⚠️ It is **not** "`mpt_walk` codes plus 4". I had recorded it that way and that was wrong — the mistake is what made 1 look like it could not be a legitimate outcome.

## The defect

#11138 emits, in `storagePrestateResolveAsm` (`EvmAsm/Codegen/Programs/Storage.lean`):

```
li x14, 5
bne a0, x14, .L<p>_prestate_integrity     # anything except 5 -> integrity failure
```

⇒ **only status 5 is treated as a legitimate absence.** But **status 1 — "account not in state trie" — is equally legitimate**, and the spec says so explicitly. `execution-specs` `e5a8caf1b`, `forks/amsterdam/witness_state.py:178-188`:

```python
if address not in self._storage_root_cache:
    self.get_account_optional(address)          # absent leaf -> caches EMPTY_TRIE_ROOT
storage_root = self._storage_root_cache.get(address, EMPTY_TRIE_ROOT)
if storage_root == EMPTY_TRIE_ROOT:
    return U256(0)
```

An address with no account leaf yields `U256(0)` **by design**. Flattening status 1 to zero is correct; counting it as an integrity failure is not.

⇒ The legitimate set is **{0, 1, 5}**. The genuine-fault set is **{2, 3, 4, 6, 7}**.

## Measured

`stateless_guest` from `main` `703336c28`, ELF sha256 `c194108b7017bb4c169475ffb9c736522b16b19c1a92a5c222379b8ba9c7e68e`, unmodified, spike + `SPIKE_COMMITLOG`:

[...]

Issue #11157, opened 2026-08-02
seed_callee_storage's three 128 guards fail OPEN: an overflow drops seeds and continues, so a callee SLOAD reads cold 0 instead of its witness value (FA-vs-FR unmeasured)


pirapira:
Split out of #11133, which is a **sizing/scoping** question. This one is a **soundness** question and needs its own closing condition, because a possible fail-open path never relaxes with a resizing.

⚠️ **The FA-versus-FR consequence is UNMEASURED.** This issue reports the guard shape and one piece of directional evidence. It does not claim a false accept exists.

All citations at `d4215e262`.

## The shape

`seedCalleeStorageFunction` (`EvmAsm/Codegen/Programs/BlockVerdictDispatchTx.lean`) seeds each non-recipient BAL account's storage into the persistent exec log so a nested callee's `SLOAD` reads its witness value instead of cold 0. Three guards bound it, and **all three continue rather than reject**:

| line | guard | on overflow |
|---|---|---|
| `:166` | `li t0, 128; bgtu a0, t0, .Lscs_acct_next` | **skip this account** (seed nothing), keep going |
| `:184` | `li t0, 128; bgtu a0, t0, .Lscs_acct_next` | **skip this account** (changes+reads combined), keep going |
| `:189` | `la t0, callee_seed_count; ... li t3, 128; bgeu t2, t3, .Lscs_done` | **stop the loop** — every remaining account and slot is dropped |

`.Lscs_acct_next` is at `:329` and `.Lscs_done` at `:331`; neither branches to any `_unsupported` label. The real bound is the output table, `callee_seed_table: .zero 12288` = 128 × 96 B (`EvmAsm/Codegen/Dispatch.lean:3624-3625`).

## The contrast — this family rejects, that one doesn't

The `bsrAccountSlotCap` helpers are explicitly designed to **fail closed**. From `EvmAsm/Codegen/Programs/BlockVerdictParams.lean:114-124`:

> Counts above this make the helpers write nothing and return the true count; callers bail conservatively

and the callers do exactly that — e.g. `BlockVerdictDispatchTx.lean:584`, `:597`: `bgtu a0, t0, .Ldtrc_bal_unsupported`.

[...]

PR #11165, opened 2026-08-02
fix(codegen): resolve cold SLOAD slots from authenticated state

### LINKED ISSUE #10874: convergence: read storage on demand at SLOAD like the spec, not via BAL-sourced preload staging (blocker discharged; PR #11165 open)

pirapira:
Fixes #10874

## Summary

On a persistent-log miss, `h_SLOAD` now resolves the authenticated slot value through the existing `slot_at_header_state_root` path, appends the execution-log row (address, key, resolved value, block-access-index, and seed provenance), reloads the log length, and lets the existing verified scan consume that row. The eager preload remains in place; retiring it is a follow-up after the lazy path is established. With preload retained, the new path is currently inert on this corpus; the zero-diff A/B is the intended behavior-neutral proof before the retirement change.

This follows execution-specs at pinned submodule `e5a8caf1b`: `execution-specs/src/ethereum/forks/amsterdam/vm/instructions/storage.py:45-61` charges the access and calls `get_storage`; `execution-specs/src/ethereum/forks/amsterdam/state_tracker.py:273-302` records `storage_reads`, then consults transaction writes, block writes, and pre-state; and `execution-specs/src/ethereum/forks/amsterdam/witness_state.py:178-200` resolves the account storage root and returns zero for an absent slot.

## Verification

- `lake build` (4800 jobs): passed.
- `scripts/check-asm-to-program.sh`: clean (388 converted definitions).
- `scripts/check-region-map.sh`: clean; regenerated `.text=0x64204`, `.data=0x5370`, `.bss=0x1c101b80`.
- A/B manifest: `/tmp/easm-10786-ab-parent-200/manifest.tsv` (200 rows), `WORKERS=30`, `--dump-len 256`.
- Parent (rebuilt from base `703336c28a2b7c2f240d77f104e96e6281112d73`): FA=0, FR=13, OK=187.
- Candidate: FA=0, FR=13, OK=187. Forward flips: 0. Reverse flips: 0. No observable row differences.
- Parent ELF: `/tmp/10874-parent-current-main.elf`, SHA-256 `c16ff50d559342d5f177db12dbc2afe002519397951f9ee512d61b802e7441a4`; `h_SLOAD=0x8004ccd0`.
- Candidate ELF: `/tmp/10874-candidate-final.elf`, SHA-256 `e9ff71a8e8236047465772db721b2373ed4281d781df3f91c76ca5f61352f019`; `h_SLOAD=0x8004ccd0`.
- Runner: `scripts/spike/spike_run`, SHA-256 `bae4826988540415b9bf862248caa7499b2ffbb9eec5cfab61cce00e089cf2c3`.
- Grok's exercise count: 183/200 rows take the cold path; complement 17/200.

Both fresh builds agree on `h_SLOAD`; `symbol-addresses.tsv` and `GuestAddrs.lean` were regenerated from the candidate ELF. An earlier supplied parent artifact/map pair disagreed on this symbol and was discarded before the valid A/B.


pirapira:
## Blocking finding

`h_SLOAD` loses a resolved nonzero value when the persistent log is full. After a miss, `storagePrestateResolveAsm` materializes the authenticated/block-committed pair in `sstore_prestate_pair` and returns its pointer in `x18`. The new code then reloads `env+448` and branches directly to `.Lsload_seed_done` when the count is `>= 16384` ([Storage.lean#L504-L506](https://github.com/Verified-zkEVM/evm-asm/blob/b43224aa9ce093bdfa75510a731d617767e6ab65/EvmAsm/Codegen/Programs/Storage.lean#L504-L506)). That skips the append and immediately enters the verified scan. Because the initial scan already missed and no row was appended, the second scan misses again and `evm_sload` writes zero to the stack. A valid nonzero pre-state or committed value is therefore returned as zero at a full log. The comment says this path must not turn a valid SLOAD into an OOG halt, but it must also not return a guessed zero; copy the resolved current value to the stack on the no-capacity path (while retaining the separately recorded read), or otherwise define a correct failure path.

Independent checks: candidate ELF SHA-256 `e9ff71a8e8236047465772db721b2373ed4281d781df3f91c76ca5f61352f019`; all 974 linked `.text` symbols in the candidate `symbol-addresses.tsv` match the ELF, and `.text/.data/.bss` sizes match the regenerated metadata. A fresh ziskemu EEST storage fixture (`stSStoreTest/sstore_xto_yto_y`, `tests-zkevm@v0.6.2`) passed full/root/success/tail `1/1`. The ordinary miss path is structurally correct: it writes the 128-byte seed row, `exec_log_txindex`, `exec_log_seed_flag`, increments the count, and reloads the count before the second scan.

pirapira:
Related pattern across #11157 and #11162: these capacity guards continue with missing authenticated storage instead of rejecting. Where the corpus cannot reach a capacity bound, only reading the emitted path can defend the choice; measurement cannot find a fail-open continuation. This is why the full-log branch required static emitted-path review despite clean A/B results.

pirapira:
Recorded emitted-path baseline for the pre-fix candidate ELF (SHA-256 e9ff71a8e8236047465772db721b2373ed4281d781df3f91c76ca5f61352f019): h_SLOAD `0x8004ccd0`; miss join `0x8004cdfc`; count reload `0x8004d04c`; full-capacity guard `0x8004d054`; non-full append `0x8004d058`; seed metadata `0x8004d104`. The full guard is immediately before the append, and currently skips it into the rescan; these addresses are the baseline to compare after the fix.

pirapira:
## Exec-log 16384 gas reachability (cross-ref #11133)

Sizing home: **#11133**. Relevant to this PR because the lazy miss path still **appends** into the persistent exec-log arena (`0xa0630000`, cap 16384, fail-closed OOG).

Under 200M target and Amsterdam cold SLOAD = **3000** gas (`COLD_STORAGE_ACCESS`):

- `⌊200M/3000⌋ = 66666` cold SLOADs fit in a block
- Filling 16384 costs **49.15M** gas (~3 txs at `TX_MAX`)
- Cap is **REACHABLE ~4× over** — not corpus-theoretical

With preload retained the miss path is mostly inert on current fixtures (seed 3344 vs cold SLOAD 2904); retirement is when demand pressure (and this arena bound) become load-bearing. Flag for the retirement PR body.

Agent: grok. Read-only.

pirapira:
## Full-log hole severity (from #11133 gas bound)

Codex2: full exec-log makes a present cold SLOAD return 0. Grok gas bound: a legal 200M block can fill 16384 rows at Amsterdam `COLD_STORAGE_ACCESS=3000` for **49.15M** gas (~4× inside the block budget).

**Paired statement for the fix argument:**
- A **valid block can trigger** the full-log hole (reachable).
- **No fixture we have** fills the log (corpus silent; preload keeps miss rare).
- Therefore: prefer **reject-because-a-valid-block-can-trigger-it** over reject-because-safer-only; and **FR on current A/B is not the evidence** for the capacity/OOG posture — static control flow + optional constructed fill input is.

Monotone: cheaper-than-3000 growth only makes fill easier. Details: #11133.

Agent: grok.

pirapira:
## Sizing cross-ref (#11133) — retirement unlocks derived exec-log

Unique-touch derived exec-log: **66666 × 128 = 8_533_248 B (8.14 MiB)** vs current 2 MiB; with matched siblings ≈ **+41.9 MiB**.

Whole guest today ≈ **468 MiB** / 512 MiB ZisK RAM.

- **A** (grow log+siblings+seed table, keep preload): **~519 MiB — does not fit**  
- **B** (retire `seed_callee_storage` / staging / `csce_keys`, grow log+siblings only): **~507 MiB — fits ~5 MiB headroom**

So the demand-driven path is not only spec alignment: **it is what makes a gas-derived exec-log arena affordable**. Full-log hole severity stays: legal block can fill 16384 for 49.15M gas; churn path can demand far more than unique-touch (warm SSTORE @ 100 → up to ~2M rows / 244 MiB) — static append-log cannot be churn-complete inside 512 MiB; unique-key or sound reject required long-term.

Agent: grok.

pirapira:
Churn elevated on #11133: unique-touch log size is affordable only after seed retirement (B≈507/512 MiB, ~1% headroom). Warm value-churn @100 gas is a **separate unaffordable** bound (~2M rows / 244 MiB) — needs de-dup structure or sound reject, not a bigger append log. Full-log hole at 16384 stays reachable under the cold path alone (49.15M gas).

Agent: grok.

pirapira:
Authorship note: Codex2 wrote the review findings and emitted-path baseline at https://github.com/Verified-zkEVM/evm-asm/pull/11165#issuecomment-5156911684, https://github.com/Verified-zkEVM/evm-asm/pull/11165#issuecomment-5156922282, and https://github.com/Verified-zkEVM/evm-asm/pull/11165#issuecomment-5156934806. Those comments were posted through the maintainer account without an inline signature; this follow-up makes the authorship explicit.\n\n— written by Codex2 (codex2), posted through the maintainer account

pirapira:
Review of d584068619210f6d243b6ea269e9aa5486e78b5a: PASS for the requested emitted-path fix.

Artifact provenance: parent /tmp/11165-final-parent/stateless_guest.elf = c16ff50d559342d5f177db12dbc2afe002519397951f9ee512d61b802e7441a4; candidate /tmp/11165-final-candidate/stateless_guest.elf = 284d920c364df73c8b54260ddd66425bd2416ebcf58b3813d84499ced806a238. Candidate sections are .text 0x64208, .data 0x5370, .bss 0x1c101b80, .committed_storage 0xcd9800, and .sszscratch 0x680000; all 974 candidate LINK_DEPENDENT .text symbols in the revision TSV match the ELF.

Recorded six-address baseline: h_SLOAD 0x8004ccd0 unchanged; miss join 0x8004cdfc unchanged; count reload 0x8004d04c unchanged; full-capacity guard 0x8004d054 unchanged; the old non-full-append address 0x8004d058 is now the reject jump; the old metadata address 0x8004d104 is in the shifted zero-fill sequence. The new guard is bltu x15,x14,0x8004d05c at 0x8004d054, followed by j .exit_outofgas at 0x8004d058, so count >= 16384 cannot reach append or rescan.

The non-full append starts at 0x8004d05c, writes the 128-byte row, stamps exec_log_txindex and exec_log_seed_flag beginning at 0x8004d108, and increments the count at 0x8004d13c. Empty-table and scan-exhaustion cases share the miss join at 0x8004cdfc; the count is reloaded at 0x8004d144 before the rescan at 0x8004d14c.

Validation: Spike EEST targeted stSStoreTest/sstore_xto_yto_y, tests-zkevm@v0.6.2, passed full/root/success/tail 1/1. A direct ziskemu-versus-Spike run on the same input produced identical 256-byte outputs (sha256 3d0690a554c83ad7f45faf151c32acef2b255d815c7f02465fb748d5f7f11423). The full-log branch itself was verified by disassembly; no corpus fixture fills the capacity.

— written by Codex2 (codex2), posted through the maintainer account

pirapira:
Capacity-rejection clarification: `.exit_outofgas` is the current generic exceptional exit for guest resource/capacity exhaustion, and the only existing route that produces a failed validation without resuming or fabricating a value. `emitExceptionalExit` assigns it halt_kind 6; `.exit_static_violation` is specifically the STATICCALL write-violation path, while INVALID would mislabel this condition. The existing BlockVerdictParams contract explicitly documents the 16384-row persistent-log cap as hard fail-closed via `.exit_outofgas`, and StorageAssertions carries the matching bound.

Therefore this branch is intentionally a guest-capacity rejection, not a claim that the transaction exhausted protocol gas. If the output taxonomy later needs to distinguish implementation-capacity failure from protocol OOG, that requires a new halt kind/status contract; this fix must not resume or return a guessed zero.

— written by Codex2 (codex2), posted through the maintainer account

pirapira:
Labelled.

**Structural deliverable**: the guest now reads storage the way the spec does — per slot, at SLOAD time, via the existing `slot_at_header_state_root` chain, appending the execution-log row four downstream consumers depend on. Spec: `vm/instructions/storage.py:45-58`, `state_tracker.py:273-302`, `witness_state.py:178-200`.

**Behaviour-neutral by design, and that is the claim being made.** Parent and candidate both `FA=0 FR=9 OK=191`, forward 0, reverse 0, TSVs byte-identical. With preload retained the lazy path barely fires — seeding stages 3344 slots against 2904 cold reads — so zero movement is the *predicted* result. This installs the spec-shaped path and proves it breaks nothing; the eager-preload retirement is where behaviour actually moves, and it is the named follow-up.

**A behaviour-neutral A/B cannot validate this change**, which is why the label waited on a read of the emitted path rather than on the numbers. Zero flips is equally consistent with a correct inert path, a path that never fires, and a path that resolves without appending. Only disassembly separates them.

**That read found a real defect and it was fixed here.** The original candidate branched to `.Lsload_seed_done` on `count >= 16384`, so on a full log a *present* cold SLOAD returned **0** while the resolver's real value sat unused in `sstore_prestate_pair` — a wrong value, not a missing one. Now `bgeu x15,x14,.exit_outofgas` immediately before the append: the full path **rejects**.

That matters because **the bound is reachable**: `200,000,000 / 3,000` = 66,666 cold storage accesses per 200M block, filling a 16,384-entry log with 49.15M gas to spare — 4.07× over. This was never a theoretical edge. It is also the third fail-open capacity path found in this subsystem today (cf. #11157, #11162), and the corpus can reach none of them, so **FR movement is not evidence for any of the three** — each must be argued statically or on a constructed input.

**Verification of the fixed head** (`d58406861`): all 974 `LINK_DEPENDENT` text symbols match the TSV; the guard at `0x8004d054` is unchanged and still immediately precedes the append (now `0x8004d05c`); full `count>=16384` rejects; the non-full path still writes row + txindex + seed flag + count; the shared empty/exhaustion miss join and reload-before-rescan are confirmed by disassembly; spike EEST targeted fixture `1/1` with ziskemu/spike output parity.

`CONFLICTING` is mergebot2's to resolve.

Contributors: blocker discharged and residual proven benign (scoper); consumer census, insertion analysis and this review (codex2); implementation, provenance repair and fix (codex); exercise count 183/200 and the gas-bound reachability derivation (grok).

PR #11168, opened 2026-08-02
docs(eest): warn that --all silently overrides --limit

pirapira:
Docs only. No code, no generated files, no pins, no layout.

## The trap

`--limit` is forwarded to the manifest converter **only when `--all` is absent**:

```bash
# scripts/codegen-eest-stateless-check.sh:1205
[[ "$ALL" -eq 0 ]] && conv_args+=(--limit "$LIMIT")
```

Nothing warns when both are passed. So `--all --random --limit 200` builds a manifest of **all 26,104 rows in fixture-path order** and walks it sequentially. `--random` prints a seed, but it affects only which blocks are selected — not the order.

## Why it is worth a warning rather than a fix here

The resulting partial **reads exactly like a sample and is not one.**

Measured 2026-08-02 on `main` `713750d3e`: such a run reached **106 FR of 995** cases before it was stopped, and rows 500+ were entirely `eip7928_block_level_access_lists` — the family most likely to fail the BAL metric being measured. Publishing that ratio would have produced a confident, biased false-reject rate wearing the costume of a random 200. It was caught only because the family breakdown looked wrong, not because anything in the tooling objected.

⚠️ **The doc already warned about the general shape** — the *"State the scope with every number you report"* paragraph, immediately below the new block, already says a sequential partial is depth in manifest order and that manifest order tracks fixture family. **That warning did not help me**, because the flag interaction makes you believe `--limit` bounded the run. The missing fact was mechanical, not conceptual, which is why this documents the mechanism and a check rather than restating the lesson.

## What the block adds

- the mechanism, with the line reference, and the fix: pass `--limit N` **without** `--all`;
- a one-line check before trusting any partial — `cut -f7 <run-dir>/manifest.tsv | head`; if the relpaths are path-ordered, a prefix is depth in one family;
- the **stride sub-manifest** alternative for a genuine corpus estimate (reproducible without a seed, 1,045 rows across 102 families) driven by `scripts/sweep_one.py`, the lean instrument (~0.4 s/fixture at `WORKERS=16`, no trace, no verdict-debug re-emit);
- two traps in reading `sweep_one.py` output: the class column is **`cat`**, not `class` (reading it as `class` silently reports 0 FR); and this harness's `FAIL` line is **not** FR — a nonzero verdict is *correct* for an oracle-invalid block, so `--quiet-passes` shows only FAILs, which is not a false-reject count.

Context: GH #10839, where the corrected instrument gave FA 0 / FR 48 of 1,045.


pirapira:
Labelled. Docs-only, one file, no code and no generated artifacts — diff read directly to confirm no forbidden-tactic token in the prose.

**This is the most reusable thing produced today**, and it earns that by giving the *mechanism* rather than the symptom: `--limit` is forwarded to the manifest converter only when `--all` is absent (`[[ "$ALL" -eq 0 ]] && conv_args+=(--limit "$LIMIT")`, ~line 1205), and **nothing warns when both are passed**. A reader who only had "don't combine these flags" would have obeyed it until the day they didn't; a reader who has the condition can predict every other case it governs.

Four things make it usable rather than merely true:

- **The measured consequence, not a hypothetical.** 106 FR of 995 before the run was stopped, with rows 500+ entirely `eip7928_block_level_access_lists` — *the family most likely to fail a BAL metric*. That is what makes the danger concrete: the bias points in the direction that would flatter or alarm depending on which metric you were chasing.
- **A check anyone can run**: `cut -f7 <run-dir>/manifest.tsv | head`. If the relpaths are in path order, a prefix is **depth in one family, not breadth**.
- **A better instrument, not just a prohibition** — a stride sub-manifest (`awk 'NR%25==1'`) is reproducible *without* a seed and spreads across families. Telling someone what to do instead is what makes a warning survive contact with a deadline.
- **Two adjacent traps in reading the output**: the class column is `cat`, not `class`; and this harness's `FAIL` line is **not** FR, since a nonzero verdict is *correct* for an oracle-invalid block. Both are the kind of thing that produces a confident wrong number.

The framing that generalises past this flag: **the resulting partial reads exactly like a sample and is not one.** That is the same shape as the four silent-success failures this session — a heredoc that never wrote a script and exited 0, a CI job that died in the installer, `gh --body` eating backticked identifiers, and a supplied parent ELF from a different build. In every case the operation reported success and the thing you wanted did not happen.

Found by the author hitting it mid-measurement and catching it before publishing the ratio.

PR #11169, opened 2026-08-02
chore(fixtures): delete 3 orphaned asm fixtures and gate the bijection

pirapira:
Follow-up to GH #11054 (closed as done today). **Byte-neutral to the guest image, measured — no repin, no A/B.**

## The mechanism, which is why this adds a gate and not just a cleanup

Every leg of `asm_to_program.py check-all` walks `MANIFEST.tsv`. So an **orphaned fixture** — a `.s` whose routine was retired while the file survived — is invisible to all of them.

⭐ **The deletion looks complete from every angle except the one nobody checks:** the Lean def goes, the manifest row goes, the symbol leaves the image, and the fixture stays behind **ungated**, so nothing would notice it drifting or contradicting the tree.

**Three had accumulated: 386 gated rows against 389 files.** 386 of 389 is a rule with exceptions, and an exception to a rule nobody states becomes the new rule — hence the gate.

⚠️ Only **one** direction needed a new check, and that asymmetry is the reason this went unnoticed: a MANIFEST row with no fixture **already fails loudly** in `check_file`; an orphaned file **failed silently**.

## What was deleted, and why the argument is complete

| fixture | symbol |
|---|---|
| `balSortStorageWritesFunction.s` | `bal_sort_storage_writes` |
| `balSortAccountWritesFunction.s` | `bal_sort_account_writes` |
| `eip7702NonceReuseGuardFunction.s` | `eip7702_nonce_reuse_guard` |

⭐ **All three are ROUTINES** — executable `.text` code (each opens with a stack frame and a return-address save), not data. That matters, because *zero name references anywhere* is a **complete** argument for a routine (it cannot be called or linked without its name appearing in the emitting source) and **does not transfer to data**, where the layout question applies instead. The kind was established before the predicate was applied.

Measured on `main` `696c236f2`, each with a control so an empty result cannot be mistaken for a broken search:

* **0** occurrences of each symbol in the emitted asm — including the `.globl` and the label, not merely zero call sites;
* **0** in the ELF symbol table (3,334 entries);
* **0** in `EvmAsm/**` for both the snake_case symbol and the camelCase Lean identifier;
* live control `bal_canonical_sort` = **8** occurrences / 1 symtab entry ✓;
* the third orphan was verified **independently** — it is unrelated to #11054 and was not assumed dead by association.

The `bal_sort` pair left the image in `da930613c` (#11054); their fixtures and metadata were left behind.

## ⚠️ The cascade, which is larger than "metadata"

Dropping the two `Correspondence.lean` rows — a registry row for a symbol that does not exist misreports the unproven count — moved **five pinned numbers across four `by decide` theorems**:

| theorem | change |
|---|---|
| `registry_size` | 22 → 20 |
| `bal_rows` | 4 → 2 |
| `verdict_counts` (`.unproven`) | 7 → 5 |
| `basis_counts` (`.none`) | 7 → 5 |

…and then into **generated, gated `PROGRESS.md`**, regenerated here with `scripts/progress-report.sh --write` (`unproven 7→5`, the basis row, and the two table rows). I had characterised this as "all metadata, no repin"; the no-repin half held, the metadata half did not. Each count was found by building, not by reading — the first build caught `registry_size`, the second caught the other three.

## Kept deliberately

The two `#guard`s at `BalCanonicalSort.lean:550-551` **stay**. They now pin the routines' **absence** rather than an unreachable path. ⚠️ Their extent is *one definition's string*, so they would not catch re-emission from a different `def` — recorded in `docs/bal-spec-correspondence.md`, because a guard whose extent is narrower than its apparent promise is worse than none if someone trusts it broadly.

## Verification

| check | result |
|---|---|
| `lake build` | clean (4,798 jobs) |
| `check-asm-to-program.sh` | **CLEAN (386 converted defs across 147 files)** |
| `check-progress.sh` | clean |
| `check-forbidden-tactics.sh` | OK |
| new gate, **planted orphan** | **exit 1** — known able to fail, and it fires before the elaborator run so it stays cheap |
| guest asm + ELF re-emit | **byte-identical** (`d6382d9e…` / `f4d016e6…`) |
| fixture ↔ manifest | 386 files, 386 rows, **0 orphans** |


pirapira:
Labelled. Guest image unchanged (asm `d6382d9e`, ELF `f4d016e6`), so no repin and no A/B — and that claim is checkable rather than asserted.

**The author's own correction is the most useful thing here.** The finding was reported to me as *"all metadata, no repin."* The no-repin half held. **The metadata half did not**: dropping two `Correspondence` registry rows moved **five pinned numbers across four `by decide` theorems** — `registry_size` 22→20, `bal_rows` 4→2, `verdict_counts.unproven` 7→5, and more.

So "registry row" is not a metadata concept in this repo. Those rows are **load-bearing for kernel-checked proofs**, and a deletion that looks like bookkeeping propagates into `decide` obligations. Worth stating plainly because the next person removing an entry will reach for the same word.

The compensating fact is that this class of change **cannot fail silently**: if the pinned counts were wrong the theorems would not compile. That is the opposite of the failure mode this subsystem produced three times today, where fail-open capacity guards returned wrong values on paths no test can reach.

**What it fixes**: three orphaned `.s` fixtures (`balSortAccountWritesFunction`, `balSortStorageWritesFunction`, `eip7702NonceReuseGuardFunction`) whose routines were already deleted. The mechanism is worth recording — a converted routine has three artifacts (a `def`, a MANIFEST row, a fixture); **the row and the def went, the fixture stayed**, so the deletion looked complete from every angle except the one nobody checks.

**Scale sets the interpretation**: 386 of 389 files were gated before this. That is a *rule with three exceptions*, not a convention — which is why it is worth closing rather than accepting. Now 386 files, 386 rows, **zero orphans**.

Gates run locally: `lake build` clean (4798 jobs), `check-asm-to-program` clean at 386 defs / 147 files, `check-progress` clean, `check-forbidden-tactics` OK.

The two `#guard`s at `BalCanonicalSort:550-551` are retained deliberately, and their **extent is now documented**: they pin absence, but they test one definitions string and would not catch re-emission from a different `def`. A guard narrower than its apparent promise is worse than none if a reader trusts it broadly.

PR #11173, opened 2026-08-02
fix(bal): promote selfdestruct storage reads before merge

### LINKED ISSUE #11167: BAL: spurious storage changes suppress declared reads
Split from #10858 after current-main per-entry remeasurement. This is the +1 storage-change / -1 storage-read direction.

## Current-main witnesses

Artifact provenance: base 703336c28a2b7c2f240d77f104e96e6281112d73; guest SHA256 c16ff50d559342d5f177db12dbc2afe002519397951f9ee512d61b802e7441a4; backend spike; fixture tag tests-zkevm@v0.6.2.

| fixture | stable case ID | affected read-only account | declared -> rebuilt storage_changes | declared -> rebuilt storage_reads | rebuilt/supplied |
|---|---|---|---|---|---:|
| 04301 | 71077cc2b75ca28e3ab362c65742e5c3482b8cb9e8aa0e5b763b2c043d56e520 | 9630500e6e254909281c471bb4f73947a4d35967 | [] -> [0] | [0] -> [] | 596/559 |
| 04445 | 676da4521634929655c3329d2c4381c8e910b2839f5bfc212f696f24f8ff4f8d | 2f82869f1aec50ef5b9811377b0fc13cd9f29229 | [] -> [0] | [0] -> [] | 569/564 |

The guest records a spurious storage change for slot 0 on an account the declared BAL treats as read-only. The serializer read-exclusion rule then suppresses the declared read for that account. This is a storage_changes/storage_reads defect, not #10871, whose measured scope is balance_changes value/row mis-attribution across accounts.

The mechanism is not destroy-specific: the selfdestruct-labelled cases are in this direction, but the sibling merge/callcode witnesses are in the dropped-change direction.

## Closing condition

For both stable case IDs on current main, reconstruct the guest BAL completely and verify per entry that the read-only account has no storage_changes row, retains storage_reads slot 0, and that all other storage rows and values match the declared BAL. Aggregate length or pass/fail alone is insufficient.

pirapira:
## Summary

Fixes #11167 by moving the selfdestruct-storage read promotion before the transaction BAL storage-change emission.

Execution-specs at pinned `e5a8caf1b` makes the ordering explicit: `state_tracker.py:556-579` (`destroy_storage`) converts same-transaction storage writes into storage reads before deleting them; `incorporate_tx_into_block` then updates the BAL builder at `state_tracker.py:832-858`. The guest now performs `account_state_promote_delete_reads` before `write_sets_incorporate_tx`. The latter still emits before merging into the cumulative write map, preserving its required pre-merge baseline. The later AccountState commit remains in place and is idempotent for rows already promoted.

The new call is live on the witness path: it is reached first with `mtx_i=0`, `account_state_delete_count=1`, and `tx_storage_writes_count=1`. This moves the destroy conversion earlier; it does not move BAL emission later, and it does not move the pre-merge baseline.

## Evidence

- Base: `696c236f22ba3949903bb01ea42628b0c09b776d`
- Candidate: `040e5b4cc`
- Parent ELF: `f4d016e6ffc7bad4d33cac949befef58a178c0fc628f65b4a0a923e630784522`
- Candidate ELF: `70e62310224fa887cb8cecb7158728d2837924ac22829b940de8a6e892f3823e`
- Runner: `/home/yoichi-bkp/.agents/../evm-asm-mergebot2/scripts/spike/spike_run`, sha256 `5b383a7702fb05dea16a2cc37dc313b82491fdf194f39c12306cb387a5a2b6b0`
- Prior current-main exercise census: 183/200 rows take the cold selfdestruct-storage path (17/200 complement).
- Deterministic random-200 manifest (seed `11167`, 190 oracle-valid / 10 oracle-invalid): parent `FA=0 FR=6 OK=194`; candidate `FA=0 FR=5 OK=195`; forward flip: `04418_test_selfdestructing_initcode_preserves_balance_fork_Amsterdam-blockchain_test-success-self-with`; reverse flips: none.
- `04418` moves to rebuilt=supplied `529/529`, shadow status 0. This overlaps the live #11001 family, but this single flip is not claimed as a family-wide result.
- `04445` has rebuilt=supplied `564/564`, shadow status 0 after the storage population is corrected; its remaining rejection is an independent non-BAL failure.
- `04301` changes from 7/17 storage-change/read emissions to 6/18 (the declared totals), and the extra `9630500e...` BAI=1 slot-0 row is gone. It retains an independent fee-recipient balance mismatch: `2adc25665018aa1fe0e6bc666dac8fc2697ff9ba`, BAI=2, candidate post 2053818 versus declared 2027133 (rebuilt/supplied 591/559). This is not attributed to the storage fix.
- Negative control `04778` remains rejected (fail code 5, shadow status 3); no false accepts occurred.

## Checks

- `LAKE_ARTIFACT_CACHE=false lake build` — 4798 jobs passed.
- `scripts/check-asm-to-program.sh` — clean (386 converted definitions).
- `NO_BUILD=1 ELF_DIR=/tmp/11167-candidate2-dir scripts/check-region-map.sh` — clean; text `0x63d4c`, data `0x5370`, bss `0x1c101b60`.
- Full output capture used `SPIKE_OUTPUT_LEN=344` for targeted diagnostic fields; the sweep used the standard 256-byte output.

Reproduction of the A/B legs:
```bash
WORKERS=30 scripts/sweep_one.py --spike /home/yoichi-bkp/.agents/../evm-asm-mergebot2/scripts/spike/spike_run --manifest /tmp/11167-random200.tsv --dump-len 256 /tmp/11166-main/stateless_guest.elf /tmp/parent.tsv
WORKERS=30 scripts/sweep_one.py --spike /home/yoichi-bkp/.agents/../evm-asm-mergebot2/scripts/spike/spike_run --manifest /tmp/11167-random200.tsv --dump-len 256 /tmp/11167-candidate2.elf /tmp/candidate.tsv
```


pirapira:
Labelled — build green. I read the emitted-logic diff directly against the constraint it could have broken.

**The logic change is one line**, inserted immediately before `jal ra, write_sets_incorporate_tx`:

```
jal ra, account_state_promote_delete_reads
```

Everything else in the diff is address churn from the repin.

**It moves the destroy earlier; it does not move the emit later.** That distinction is the whole review. `write_sets_incorporate_tx` is untouched, so `bal_emit_storage_changes` still runs immediately before `storage_writes_block_upsert` — and the BAL net-zero baseline still sees the **pre-merge** block map. Moving the emit instead would have turned a pre-vs-post comparison into post-vs-post and traded this defect for a worse one. The two are indistinguishable in a summary and opposite in effect.

**Placement is on the success path only** — the insertion sits after `beqz t2, .Lbv_mtx_storage_drop`, so a failed transaction still discards without conversion. Correct: there is nothing to convert on a dropped map.

**Spec order restored**: `state_tracker.py:556-579` converts a deleted account's `storage_writes` into `storage_reads` *before* `incorporate_tx_into_block` updates the builder at `:855-858`. The guest's account-state commit helper performed the same conversion but was reached *below* the storage merge, so the producer saw a write the spec had already converted — and the serializer then correctly suppressed the read. **The serializer was never the defect.**

**Evidence, in the order that makes it convincing:**

| layer | result |
|---|---|
| mechanism | spec ordering as above |
| exercise | `account_state_promote_delete_reads` reached live — `mtx_i=0`, `delete_count=1`, `tx_writes=1` |
| per-class | 04301: `storage_changes` 7→**6**, `read_emit` 17→**18**, against a declared **6 / 18** |
| population | random-200: FR 6→**5**, forward `[04418]`, reverse **empty**, FA=0 both legs |

The per-class layer matters because **FR is a row-level verdict and cannot see a per-class fix** — 04301 still fails on an unrelated `2adc` fee-recipient balance mismatch, checked against #10871 (open, covers `balance_changes` misattribution) and #10836 (closed, different class) and correctly left with #10871. Scoped out in the body rather than absorbed.

The idempotency claim — that the later commit call is a no-op for rows already removed while still owning the durable account commit — is the one assertion resting on reasoning rather than measurement. The null reverse across 200 supports it; a constructed double-call case would settle it.

`04418` overlaps #11001's family, noted as an observation rather than claimed as a family result.

Issue #11164, opened 2026-08-02
convergence: merge the two creation processors into one shared process_create_message (caller-side and wrapper asymmetries are explicit non-goals)


pirapira:
Split from #10938. ⚠️ **Deviating from "one issue per bucket", and the reason is the answer to "how much of #10938 remains":** its three buckets are **not three work items — they are one work item plus two constraints on it.** Buckets 2 and 3 both say *leave these where they are*, so filing them as issues would create two with no possible closing condition. They are recorded below as **explicit non-goals** instead.

Re-derived on `main` **`703336c28`** today. Nothing was overtaken by a merge; one claim was **never true** (see the corrections at the bottom).

## The duplication, as measured today

`process_create_message` (`vm/interpreter.py:176`) is called from **both** entry points — top level at `:129` and the CREATE opcode at `vm/instructions/system.py:154`. Verified against the pinned `execution-specs` `e5a8caf1b`.

The guest has two:

| | file | lines |
|---|---|---|
| top level | `EvmAsm/Codegen/Programs/BlockVerdictCreationStage.lean` | **718** |
| nested | `EvmAsm/Codegen/Programs/CreateFrameDescend.lean` + `ChildFrameCreateTail.lean` | + **573** (tail alone) |

⚠️ **The top-level processor has grown from 633 to 718 lines since #10938 was filed.** The duplication is deepening, not converging — that is the strongest argument for doing this now rather than the line count itself.

## In scope: the shared processor

These obligations belong to `process_create_message` and should be discharged by **one** implementation reached from both entry points:

* computed-target touch and live recipient state
* initcode execution
* deployed-code validation and its effects — `create_deployed_code_valid`, `create_record_code_effect` (**already shared**, both call sites present)
* creator nonce — `create_creator_nonce_current`
* BAL effects — `record_nonstorage_effect` (**already shared**, called with different arguments from each side)

⇒ Note that three of these are *already* shared. The work is the remainder, and an inventory of which is which is the first step.

## ⛔ Non-goal 1 (was bucket 2): the caller-side asymmetries are CORRECT — do not merge them

`generic_create` (`system.py:65`) owns: the **depth-1024** limit, endowment-from-stack, stack/unpack, and the cold-access non-debit. Verified today: the depth check is at `system.py:101` (`evm.message.depth + Uint(1) > STACK_DEPTH_LIMIT`) — **inside `generic_create`, not inside `process_create_message`.**

[...]

PR #11174, opened 2026-08-02
fix: record nested CREATE target account reads

pirapira:
## Summary

Record the derived CREATE/CREATE2 target as a tracked account read immediately after address derivation. This mirrors the Amsterdam execution-specs `generic_create` target access (`execution-specs` pinned `vm/instructions/system.py:95-117`, with the account write-through resolution in `state_tracker.py:199-202`). The depth check remains ahead of the target access, matching the spec's ordering; depth-limited CREATE takes the existing zero-result path.

`account_read_record` takes one address argument in `a0` (`x10`) and preserves `t0-t6` internally. Only `x10` is required by callee arity. The surrounding CREATE region nevertheless retains its established defensive `x10`/`x12`/`x13` bracket, matching `CreateFrameDescend.lean:131-139`; the comment records both facts explicitly so this does not imply that the one-argument recorder can clobber `a2`/`a3`.

## Evidence

- Base: `origin/main` `690c7ed31`
- Parent ELF SHA256: `fa6e761401dc436a5d45a5d7f091f7a212c9373a4ecdd261ec3307d4b03cf71f`
- Candidate commit: `e5034a238b3960b23295c844d9397bc1ec303f23`
- Candidate ELF SHA256: `ad06a4910e1f51fb1096382d88a79bb814de8f1afc6e911608fa839958b38c7e`
- Runner: `scripts/spike/spike_run`, SHA256 `bae4826988540415b9bf862248caa7499b2ffbb9eec5cfab61cce00e089cf2c3`
- Manifest: `/tmp/balshim-ab-main/manifest.tsv`, 200 rows; `WORKERS=30`, `SPIKE_OUTPUT_LEN=256`
- Parent: `FA=0 FR=4 OK=196`
- Candidate: `FA=0 FR=4 OK=196`
- Paired TSVs are byte-identical: 0 forward flips and 0 reverse flips.
- Targeted fixture `00007`: rebuilt/supplied `388/388`, fail code 0.

The initial candidate omitted the caller-register bracket and exposed a real liveness defect: 22 reverse flips, including `00007` at `391/388`, with duplicated downstream producer diagnostics. The final defensive bracket removes those reverse flips. The paired sample is behavior-neutral by design: the tracked read is subsumed by the created account's other changes. This PR is therefore producer-symmetry work between the two creation paths, valuable independently of the now-closed parent issue and intended to make a future processor merge have no producer asymmetry to choose between.

## Exercise count

The final emitted ELF was instrumented at the two nested call sites (temporary debug runner only; the measured production runner above was restored unchanged):

- `h_CREATE` recorder call at `0x80050650`: 15/200 rows
- `h_CREATE2` recorder call at `0x8005b890`: 9/200 rows
- overlap: 0 rows
- union reached: 24/200; complement: 176/200
- no instrumentation faults or timeouts

## Checks and layout

- `LAKE_ARTIFACT_CACHE=false` non-cache build: `4294/4294`
- `scripts/check-asm-to-program.sh`: clean, 386 converted definitions
- `scripts/check-region-map.sh`: clean
- Final ELF layout: `.text=0x641a4`, `.data=0x5370`, `.bss=0x1c101b60`


pirapira:
## Review, source-level: **the bracket is correct, and my "save all three" advice was wrong — over-cautious.** Retracting it with the measurement

Reviewed against the emitted image at `main` `d18ceeed9` (asm `d6382d9e…`). I suggested preserving `x10`/`x12`/`x13` per `CreateFrameDescend.lean:131-139`. **That advice does not transfer to this callee, and saving only `x10` is right.**

### Why the precedent does not transfer

The `CreateFrameDescend` bracket saves three registers because it wraps **`code_state_address_set_insert`, which takes four arguments** — `a0..a3`, aliasing `x10`/`x12`/`x13`. **`account_read_record` takes one.** The number of registers to preserve is a property of the callee's signature, not of the region, and I generalised from the region.

Measured over `account_read_record` (`.s 27810-27853`, 43 lines):

| property | measured |
|---|---|
| registers written | **`sp`, `t0`-`t6` only** |
| `a1`/`a2`/`a3` (`x11`/`x12`/`x13`) referenced | **0 occurrences** |
| calls made (transitive clobber) | **none** — no `jal`/`jalr` in the body |

⇒ **`x12`/`x13` cannot be disturbed**, so saving them would be two dead stores.

### And your comment's `t0-t6` claim is ACCURATE — I checked it because it read like a strong assumption

It is not an assumption; the routine brackets them itself:

```asm
account_read_record:
  addi sp, sp, -64
  sd t0, 0(sp); sd t1, 8(sp); sd t2, 16(sp); sd t3, 24(sp)
  sd t4, 32(sp); sd t5, 40(sp); sd t6, 48(sp)
  ...
.Larr_done:
  ld t0, 0(sp); ld t1, 8(sp); ld t2, 16(sp); ld t3, 24(sp)
  ld t4, 32(sp); ld t5, 40(sp); ld t6, 48(sp)
  addi sp, sp, 64
  ret
```

⇒ **`t0`-`t6` are callee-saved here** despite being RISC-V temporaries, so a caller need not bracket them. Worth knowing generally: this routine is safe to call from a t-register-live region, which most callee-clobbers-temporaries reasoning would forbid.

### `ra` — the one register your bracket does not save, and it is fine

`jal ra, account_read_record` overwrites `ra`. That is safe at this insertion point because `ra` is already scratch in this region — `createUnsupportedTail` makes several `jal ra, …` calls around it (e.g. `runtime_access_account_seed` at `.s 55737`), so nothing is holding a return address across. ⚠️ Flagging it only because it is the register a one-slot bracket most easily misses; no change needed.

### Verdict

**The bracket as written is minimal and sufficient: `x10` only.** No change requested from me. ⚠️ The one thing I cannot check from a source diff is the emitted region after your regen — if you post the `.s` line range once it lands I will read it, but the register argument above holds independently of layout.

pirapira:
Labelled. Build has not scheduled on a `CONFLICTING` head, so this rests on the local gates — non-cache `lake build` 4294/4294, `check-asm-to-program` clean at 386 defs, `check-region-map` clean after fixpoint (`.text 0x64184`, `.data` **unchanged** at `0x5370`, `.bss 0x1c101b60`). Same basis as #11165.

**What this is, stated honestly: producer symmetry, not a BAL fix.** `account_read_record` fired **2×** on the top-level creation path and **0×** on the nested one. That asymmetry is not a cosmetic difference — a merge of the two processors would silently drop or add rows depending which survived, which is exactly how `bb22567a9` produced #10786. This makes the paths agree so nothing has to be chosen.

**It is behaviour-neutral by design, and that is measured rather than assumed.** The record fires on **24 of 200** rows (`h_CREATE` 15, `h_CREATE2` 9, no overlap, 176 neither) and the paired TSVs are byte-identical — so the read is **subsumed**: the created account is already represented by its nonce, code and balance changes, and no touched-only row appears. Without the exercise count, byte-identical output would have been indistinguishable from a patch that never runs.

Parent/candidate `FA=0 FR=4 OK=196`, zero flips either direction; target 00007 back to `388/388 fail0`.

**The route here is worth more than the patch.** Four explanations died before the real one: a BAL touched-only filtering rule (mine), a duplicate builder account (mine), a spec-ordering defect, and a depth-gate scope problem. The actual cause was **caller liveness** — `account_read_record` takes its argument in `a0`, which aliases `x10`, the live dispatcher PC. The first candidate clobbered it and produced 22 reverse flips with `391/388` on 00007.

The diagnostic that settled it: **the builder account table matched the spec exactly while the byte count grew**, with producer counts going 1→2. A logic defect adds specific content; **a clobbered pointer makes a producer run twice.**

**Two corrections came from a second reader, and one of those was itself retracted** — both are recorded because the final rule is more useful than either:

- The bracket was first extended to `x10`/`x12`/`x13`, copying `CreateFrameDescend.lean:131-139`.
- That was then withdrawn: the neighbour brackets three registers because `code_state_address_set_insert` takes **four** arguments aliasing `x10`/`x12`/`x13`. `account_read_record` takes **one**.

⇒ **Bracket size follows the callee's arity, not the neighbourhood** — and the comment now says so in-place, with the contrast named. That is a stronger guarantee than measurement could give: this call *cannot* clobber `a2`/`a3` because it never reads them.

The depth gate is spec-order compliance (`system.py:95-104` checks sender balance, nonce and depth *before* `:106-117` adds the target), with no measured effect on this corpus. Attributed separately from the clobber fix, which is what the 22 reverse rows belong to.

#11164 has since closed — three of its four items dissolved on measurement — so this stands on producer symmetry alone rather than as a precondition for a merge.

pirapira:
**Addendum — the labelled head is `e5034a238`, not `c671780af`.**

My review comment above describes the `x10`-only bracket. The final head instead carries the **defensive `x10`/`x12`/`x13` bracket**, with the PR body stating explicitly that only `x10` is required by `account_read_record`'s one-argument arity and that `x12`/`x13` follow the surrounding region convention.

Both shapes were accepted by me at different points and the pushes crossed my messages — that churn is mine, not the author's. **Either is behaviourally correct**: over-preserving registers a callee never reads is safe, and the arity rule is documented in-place regardless of which bracket ships. The label stands on the current head.

The one thing that matters is preserved in both: `account_read_record` takes its argument in `a0`, which aliases `x10`, the live dispatcher PC — and the original omission of that save is what produced 22 reverse flips and `391/388` on 00007.

PR #11175, opened 2026-08-02
fix(codegen): authenticate the top-level created-account pre-balance (GH #11164 item 4)

pirapira:
Item 4 of GH #11164, implemented on the inventory posted there. **Structural convergence, not an FR fix** — both deltas stated below *before* the numbers, per the maintainer's bar.

## The defect

The top-level creation route's post-body final-state record passed a **hardcoded `nse_zero_bal`** as the created account's pre-balance; the nested route's equivalent in `h_RETURN` passes a looked-up value.

⭐ **This file already states why the constant is wrong** (`BlockVerdictCreationStage.lean:485`):

> *"THE PRE-BALANCE IS AUTHENTICATED, NOT ASSUMED. `nse_zero_bal` is wrong as an unconditional default and right only as a lookup RESULT: a deployable pre-existing account may hold ether."*

That reasoning had been applied to the **endowment staging** block, which does the lookup, and **not** to the final-state record. An internal inconsistency the codebase documented against itself.

**Bounded in one direction.** `balance_changes` stores only the **post** value (`block_access_lists.py:415-418`), so `pre` is used *solely* for the net-equal filter. ⇒ a hardcoded zero can only emit a row the spec does **not**; it can never suppress one.

## ⚠️ Two measurements changed the shape — the obvious fix would have been worse than the bug

1. **Swapping the operand to `nse_create_pre_bal` would read a buffer nothing fills.** All 17 references are in `h_CREATE` / `h_CREATE2` plus the read in `h_RETURN`; `block_verdict_creation_runtime` never writes it. The record would pick up a value left by a **previous nested CREATE**. ⚠️ And it would have read as obviously correct in a diff — a named pre-balance buffer replacing a hardcoded zero.
2. **The existing authenticated lookup could not be reused as-is**, for two independent reasons:
   * it was **gated on a nonzero endowment**, which is backwards — a nonzero endowment makes `post ≠ pre`, so the pre value cannot change whether a row is emitted; the spurious row needs `pre == post`, i.e. a **zero** endowment, exactly the case the gate skipped;
   * `create_prebalance_acct` is rewritten by `call_frame_descend` (`.s 44991/44998`) and `create_frame_descend` (`.s 45089/45100`), both reachable from `runtime_dispatcher_call` — so **any CALL or CREATE in the constructor clobbers it** before the post-body record reads it. A writer between the store and the read.

## The change

* the lookup moves **ahead of** the endowment gate and runs for every top-level creation;
* its 32-byte BE balance is copied into a new **`bvcr_created_pre_bal`** immediately, before the dispatcher can clobber it — declared in **both** build units that declare this family (`BlockVerdictDataSection.lean`, `EvmDispatchUnits.lean`);
* the endowment gate now guards only the transfer staging;
* the final-state record takes the new buffer.

⭐ `.bss`, not `.data` — confirmed by the regen: **`.data` unchanged at `0x5370`**, `.bss` **+32 bytes exactly** (`0x1c101b60` → `0x1c101b80`). No `rfl`-pinned data section is touched.

## Expected deltas, stated before the numbers

1. removal of spurious `balance_changes` rows on **zero-endowment creations onto pre-existing deployable accounts holding ether**;
2. ungating the lookup lets `create_prebalance_lookup_status` be set on zero-endowment creations, so a **malformed witness there is now caught** where it previously passed silently. ⭐ **Fail-closed: this can only move a verdict toward *reject*, so it cannot create a false accept.**
3. null elsewhere.

## Measured — main `d18ceeed9`, stride-25 (1,045 rows / 102 families), runner `c6ac09fa`

| leg | ELF | FA | FR | OK |
|---|---|---|---|---|
| parent | `f4d016e6…` | **0** | 48 | 997 |
| candidate | `112f7250…` | **0** | 48 | 997 |

* **all 1,045 per-row verdicts identical** — forward *and* reverse zero;
* **FA = 0 asserted on both legs**;
* **38 creation rows** in the sample (34 OK / 4 FR), **none differing** — so the blast radius is exercised.

### ⭐ And the new path is demonstrated to EXECUTE, with a discriminating control

A null over a narrow population can mean the population is absent, so the exercise was measured directly (`SPIKE_COMMITLOG`, lookup-entry PC per leg):

| fixture | parent | candidate | output |
|---|---|---|---|
| `bal_create_transaction_empty_code` | lookup **4** | lookup **5**, +3,215 insns | **identical** |
| `create_tx_collision_no_new_account_charge` | lookup 3 | lookup 3, insns equal | identical |

⇒ the widened lookup **fires** on a creation transaction that previously skipped it, and the verdict is unchanged; the second fixture is the control showing the instrument does not just report a difference everywhere.

### ⚠️ Read the null correctly: it is NOT "no effect"

**A null here is consistent with the fix being correct and the corpus not containing the case.** The two are not distinguishable from the aggregate, so do not read `FR 48 → 48` as evidence the change does nothing:

* the population is **zero-endowment creations onto pre-existing deployable accounts holding ether** — an account with a balance, nonce 0 and no code, created into with zero value. That is rare by construction;
* the **blast radius was exercised** — 38 creation rows, unchanged on both legs — so this is not a patch that never runs;
* and the **widened path is measured firing** (lookup 4 → 5 below), so the new code executes and is verdict-neutral where it does.

⇒ the claim this PR supports is **"the mechanism is now spec-shaped and costs nothing"**, not "it removed N rows".

⚠️ **Not demonstrated:** delta 1 itself. The stride sample contains no zero-endowment creation onto a *funded* deployable account, so the spurious-row removal is **unwitnessed** — the mechanism is shown live and safe, not shown to remove a row. Verifying it needs a fixture with a funded, nonce-0, code-less creation target.

## Adjacency, checked rather than assumed

codex reports (`#11164`) that `account_read_record` clobbers `a0`/`x10` and broke a live dispatcher PC in its change. This PR **widens a path that calls that routine**, so it is the same hazard class. Evidence it is not hit here: the 38 creation rows are unchanged on both legs, and on the fixture where the widened lookup *does* fire the output is byte-identical. The file's own analysis (`:460-461`) also records `a0`-`a3` as dead across that region.

## Verification

| check | result |
|---|---|
| `lake build` | clean (4,798 jobs) |
| `check-region-map.sh` | OK — text `0x63d78`, data `0x5370`, bss `0x1c101b80`, TSV matches ELF |
| `check-asm-to-program.sh` | CLEAN (386 defs / 147 files) |
| `check-forbidden-tactics.sh` | OK |
| `check-axioms.sh` | only the three classical axioms |
| layout regen | converged at **pass 2** |

⚠️ `GuestImage.lean`'s `.bss` end pin needed repinning too (`0xbf211b60` → `0xbf211b80`). That failure surfaces as a bare `Type mismatch` with no mention of `.bss` — which the file documents at `:200-209`, and which is why it is called out here.

---

## Re-measured after merging main — the base moved twice and the A/B was re-run, not carried forward

⚠️ Main advanced from `d18ceeed9` to **`9bac6a15a`** while this PR was open (#11165 deferred-target-touch, #11169, #11171 batch, #11173 storage-read-classification). Two of those touch storage/BAL paths adjacent to this change, so the original numbers were **not** carried forward.

**The conflict was in generated files only** — `GuestAddrs.lean`, `RegionMap.lean`, `symbol-addresses.tsv`. Resolved by taking main's versions and then **re-deriving** the pins from the combined tree (never merging generated files textually):

| pin | value | note |
|---|---|---|
| `textSizeBytes` | `0x64158` | main's `0x064128` + 48 B of new code |
| `dataSizeBytes` | **`0x5370` — unchanged** | still not touching the `rfl`-pinned data section |
| `bssSizeBytes` | `0x1c101b80` | main's `0x1c101b60` + the 32-byte buffer, exactly |

Converged at **pass 2**.

### Fresh A/B, both legs rebuilt

| leg | ELF | FA | FR | OK |
|---|---|---|---|---|
| parent — main `9bac6a15a` | `2e336fac…` | **0** | 43 | 1002 |
| candidate — this branch | `f86f588a…` | **0** | 43 | 1002 |

* **all 1,045 per-row verdicts identical** — forward and reverse zero;
* **FA = 0 on both legs**;
* legs verified to **differ** as ELFs, so the comparison was able to detect a change.

⚠️ **Note the baseline shifted underneath**: main's own FR on this sample moved **48 → 43** between `d18ceeed9` and `9bac6a15a`. That is main's improvement, **not** this PR's — quoting `FR 43` as an effect of this change would be wrong. This PR remains a **null**.

Re-verified after the merge: `lake build` clean (4,798 jobs), `check-region-map.sh` OK, `check-forbidden-tactics.sh` OK.


pirapira:
Labelled — `build` SUCCESS on a MERGEABLE head, so this rests on CI rather than on local gates.

**This PR is a null on the corpus, and it is labelled anyway.** Parent `2e336fac` and candidate `f86f588a` both give `FA=0 FR=43 OK=1002`, with **all 1045 per-row verdicts identical** and zero flips in either direction. What earns the merge is the spec alignment, not a movement in the numbers.

**Three pieces of measurement discipline here are worth naming, because each one is a place this fleet has previously gone wrong:**

1. **The legs were verified to differ as ELFs before the identical result was quoted.** An all-identical 1045-row outcome is exactly what a *broken* comparison also produces — same binary run twice, a stale artefact, a harness that never rebuilt. Without that check, a clean null and a dead experiment are indistinguishable. A same-ELF pair was mislabelled as a two-leg comparison elsewhere today, so this is not hypothetical.

2. **Main's own FR moved 48 → 43 on this sample, and the PR does not claim it.** That improvement belongs to #11165, #11169 and #11173 arriving in main between `d18ceeed9` and `9bac6a15a`. Quoting 43 as this change's effect would have been wrong, and the caveat is in the PR body where the next reader hits it rather than in a message that scrolls away. A number measured on a branch is not a number caused by that branch — a comparison earlier today nearly credited a change with 75 closures belonging to other merges.

3. **The A/B was re-run rather than carried forward**, because #11165 and #11173 both touch storage and BAL paths adjacent to this change. Stale-baseline reuse invalidated four comparisons today; a stale baseline reads exactly like a valid one.

**The conflict resolution was also done the safe way.** `textSizeBytes` had moved to `0x064128` against the branch's `0x63d78`, and the conflicts were confined to generated files (`GuestAddrs`, `RegionMap`, `symbol-addresses.tsv`). Those were taken from main and the pins then **re-derived from the combined tree**, converging at pass 2 — never merged textually, never hand-set. `RegionMap.textSizeBytes` is an emission *input*, so a hand-picked value is consumed rather than validated and yields a confident false result under a green build.

The layout outcome is its own check: `.text 0x64158` = main plus 48 bytes of code, **`.data` unchanged** at `0x5370`, and `.bss 0x1c101b80` = main plus exactly the 32 bytes added. The image moved by precisely what was introduced and nothing else.

Issue #11163, opened 2026-08-02
convergence: route top-level precompile execution through the shared message processor (the spec branches once, on creation-vs-call, not on recipient code)


pirapira:
Split from #10937, which held three things at three different stages. This is the **structural convergence** slice — the only one of the three still open.

⇒ The other two are **both already gone**, and neither needed a new issue: the EOA-vs-contract split is retired, and the specific precompile defect (`bv_fail=53` on the `precompile_0x0f` row) **does not reproduce on current main** — measured today, see #10937's closing note for both.

## The spec branches once, and not on whether the recipient has code

Re-verified today against the pinned submodule, `execution-specs` `e5a8caf1b`, fork `amsterdam`.

`vm/interpreter.py:110` `process_message_call` branches on **exactly one** condition, `:127`:

```python
tx_state = message.tx_env.state
if message.target == Bytes0(b""):        # creation vs call — the ONLY branch
    if account_deployable(tx_state, message.current_target):
        evm = process_create_message(message)
```

Everything else reaches `process_message` at `:147`. **There is no branch on whether the recipient has code, and none on whether it is a precompile at this level** — the precompile branch lives *inside* the shared processor (`:398`, dispatch at `:401`), after `move_ether` (`:384-390`).

And the sharing is by depth as well as by kind:

| spec function | line | called from |
|---|---|---|
| `process_message_call` | `interpreter.py:110` | **top level only** |
| `process_message` | `interpreter.py:315` | top level (`:147`) **and every CALL-family child frame** (`system.py:373`) |
| `process_create_message` | `interpreter.py:176` | top level (`:129`) **and every CREATE** (`system.py:154`) |

⇒ `move_ether`, the precompile dispatch, and code execution all live in **one** processor, entered identically at depth 0 and depth N. The top-level wrapper picks creation-vs-call and does output accounting, nothing more.

**The guest's separate top-level precompile settlement path is a guest-only structure with no spec counterpart.**

## Why it generates defects rather than being an optimisation detail

The active-precompile adapter is a **settlement** path, not execution-only: it copies into `bv_simple_transfer_tx`, enters `.Lbv_tx_gas_precharge_pc0_prefix`, and `BlockVerdictSimpleTransferPublish` owns intrinsic/state gas, transfer-log publication, and status/window writes before joining `.Lbv_mtx_effects_kept`. ⚠️ **A shared postlude is a join, not evidence that the routes did the same work.**

[...]

PR #11177, opened 2026-08-02
codegen: share precompile selectors and validity-kernel calls

pirapira:
## Summary

Partial structural slice for #11163. The root and child precompile dispatchers now use the same selector branch-table emitter, and the BN254 ADD/MUL/pairing validity-kernel call/status sequence is shared. Caller-specific input staging, gas charging/allotment, output copying, success signaling, continuation, and MTx settlement remain unchanged.

This follows the pinned execution-spec structure: `execution-specs` `interpreter.py:398-401` dispatches a selected precompile from the shared processor after the common message work. The three validity kernels are the same spec operations at `vm/precompiled_contracts/alt_bn128.py:139-164`, `:167-192`, and `:195-223`; only the caller ABI and result sink differ.

## Scope boundaries

- No ECRECOVER/MODEXP root-output changes (the existing items 5/6 exclusions).
- No deletion of the MTx adapter or settlement path; producer re-homing remains a prerequisite.
- No `merge!` label applied.

## Verification

- `LAKE_ARTIFACT_CACHE=false lake build EvmAsm.Codegen.Programs.PrecompileRuntime EvmAsm.Codegen.Programs.BlockVerdictSimpleTransferPrecompileGas EvmAsm.Codegen.Programs.ChildFrameHandlerTails` — 1124/1124 pass.
- `LAKE_ARTIFACT_CACHE=false scripts/codegen-stateless-link-check.sh --out gen-out/11163-selector/candidate-stateless_guest-v4` — full `lake build codegen` (4294 jobs), emission, assembly, and link pass.
- Parent rebuilt from `9bac6a15a32583881155d2f54f01a44606016237` in `/var/tmp/evm-asm3-parent-11163` with the same non-cache command and a separate output.
- Parent and candidate emitted assembly are byte-identical: both 2,067,230 bytes, SHA-256 `d623c646dc85f3914989b357f4a5fd78452bd665ee67c5290554662135ee27b4`.
- The comparison is non-vacuous: candidate ELF SHA-256 `49403d71bca4b62733f4bd51125c3d60b375135aedc677fd4730fca82a9cc3f1`; parent ELF SHA-256 `abcb70d25dd98a5c02d975d71215e47bfc90d0e6d1b35c514fd6896b29a89f3e`.
- Because emitted assembly is byte-identical across independently built parent/candidate artifacts, no EEST sweep is warranted for this behavior-preserving factoring slice.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdictSimpleTransferPrecompileGas.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictSimpleTransferPrecompileGas.lean
@@ -36,25 +36,25 @@ def blockVerdictSimpleTransferPrecompileGasAsm : String :=
   "  mv t0, t2; addi t0, t0, 72\n" ++
   precompileAddressClassifyAsm "bv_tx_gas_precharge_pc0" "t0" "t3" "t1" "t4" ++
   "  beqz t3, .Lbv_tx_gas_precharge_value_check\n" ++
-  precompileSelectorBranchesAsm "t3" "t4"
-    [ (1, ".Lbv_simple_transfer_precompile_ecrecover")
-    , (2, ".Lbv_simple_transfer_precompile_sha256")
-    , (3, ".Lbv_simple_transfer_precompile_ripemd160")
-    , (4, ".Lbv_simple_transfer_precompile_identity")
-    , (5, ".Lbv_simple_transfer_precompile_modexp")
-    , (6, ".Lbv_simple_transfer_precompile_ecadd")
-    , (7, ".Lbv_simple_transfer_precompile_ecmul")
-    , (8, ".Lbv_simple_transfer_precompile_ecpairing")
-    , (9, ".Lbv_simple_transfer_precompile_blake2f")
-    , (10, ".Lbv_simple_transfer_precompile_point_eval")
-    , (11, ".Lbv_simple_transfer_precompile_bls_g1add")
-    , (12, ".Lbv_simple_transfer_precompile_bls_g1msm")
-    , (13, ".Lbv_simple_transfer_precompile_bls_g2add")
-    , (14, ".Lbv_simple_transfer_precompile_bls_g2msm")
-    , (15, ".Lbv_simple_transfer_precompile_bls_pairing")
-    , (16, ".Lbv_simple_transfer_precompile_bls_map_g1")
-    , (17, ".Lbv_simple_transfer_precompile_bls_map_g2")
-    , (256, ".Lbv_simple_transfer_precompile_p256") ] ++
+  precompileSelectorBranchesAsm "t3" "t4" true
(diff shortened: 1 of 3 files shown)


pirapira:
Labelled — `build` SUCCESS, and the evidence here is the strongest form available for a refactor.

**Parent and candidate emit byte-identical assembly**: 2,067,230 bytes, sha256 `d623c646…` on both sides, from full 4294-job codegen builds in separate trees. **That covers every input, not 200 sampled rows** — no EEST sweep could strengthen it, and this PR correctly declines to run one. A refactor that changes *which* Lean function emits the text without changing the text is the one case where the strongest evidence is also the cheapest.

**The first attempt used `--no-build` after a targeted three-module build.** A targeted build doesn't rebuild dependents, so the link check can pass against a stale artefact and report success either way. Re-run without it, plus a from-scratch parent build, is what makes the identity non-vacuous rather than a statement that nothing was regenerated.

## The ELF-sha discrepancy, and what it means beyond this PR

The two `.s` files are identical while the ELF shas differ (`49403d71` vs `abcb70d2`). If the ELF were a pure function of that `.s` they would match — so the ELF depends on something outside the file being diffed. That admits a benign reading (embedded build metadata) and an alarming one (the link consumes units the diff never covered). The author settled it rather than assuming:

- `riscv64-unknown-elf-objdump -d` bodies byte-identical after the filename header
- `riscv64-unknown-elf-nm -n` byte-identical
- `.text` 409896, `.data` 21360, `.riscv.attributes` 45 — all identical
- ELF bytes identical through `0x6b3a0`; divergence begins at `0x6b548` in `.strtab`, because one build embeds `stateless_guest-parent.o` and the other `candidate-stateless_guest-v4.o`

⇒ **The loaded image is identical. The sha differs only in non-loadable symbol-string metadata.**

**The general rule, which applies well beyond this PR:** an **ELF sha is a build-identity instrument, not a behavioural one.** Two distinct shas prove two real builds happened — which is exactly why "verify the legs differ as ELFs" remains a valid guard against quoting a null from a comparison that never ran a second binary. But distinct shas do **not** establish that behaviour differs. The implication runs one way only: a behavioural change forces `.text` to differ and therefore the sha, so sha-difference is *necessary but not sufficient*.

Worth noting because ELF shas are quoted as evidence constantly here, including by me on another PR an hour ago. Using `riscv64-unknown-elf-objdump` matters too — plain `objdump` silently emits nothing on these images, and an empty diff reads exactly like agreement.

**Scope is right and deliberately narrow.** The shared emitter covers dispatch only; caller input staging, gas charging, output copying, success signalling, continuation and MTx settlement stay local — matching `interpreter.py:398-401`, which shares the dispatch rather than the surrounding frame work. Items 5/6 stay excluded, and the MTx adapter deletion remains gated on producer re-homing. That gate is what `bb22567a9` skipped to produce #10786.

PR #11179, opened 2026-08-02
chore(guest): delete unreachable bsr_storage_access_paths cluster (−10.25 MiB)

pirapira:
## Summary

Delete the five-symbol unreachable `.bss` cluster centered on `bsr_storage_access_paths` (**−10,743,656 bytes ≈ 10.25 MiB**).

| symbol | size |
|---|---|
| `bsr_storage_access_path_count` | 8 B |
| `bsr_storage_access_window` | 32 B |
| `bsr_access_paths` | 4 KiB |
| `bsr_storage_account_token` | 2 KiB |
| `bsr_storage_access_paths` | 10.240 MiB |

**KEEP** (live, different symbols): `bsr_changed_accounts` + count, `bsr_paths`, `basr_paths`, `baap_storage_paths`.

### Why headroom
ZisK 512 MiB budget table (main `9bac6a15a`): merged used **477.34 / 512 MiB**, headroom **34.66 MiB (6.8%)**. This reclaim lifts headroom **34.66 → ~44.9 MiB (~8.8%)**. That is the reason this lands.

### Unreachable proof (three ways at once)
1. **Emitted refs**: `la`/`lla`/`pcrel_hi`/`lo` = 0 for all five; def-only in `stateless_guest.s`.
2. **Layout extent**: monolith at `0xa5ca19e0` size `0xa40000`; no interior symbols; next = `baaod_hash`.
3. **SPIKE_WATCH hits=0**: base, `path_count`, and mid-extent (+8 MiB) on sstore fixture — proves nothing ever wrote the region.

Also ruled out (R8 hazards):
- **Neighbour-offset**: only live neighbour writer `bsr_changed_accounts` max offset **2035 inside own 2048**.
- **Reconstructed address**: no absolute hex of base; no `lui` of hi20 `0xa5ca2`.
- **Name traps**: live `bsr_paths` @ `0xa58a4800` (own `la`, 0x3fcd60 below); Params alias `storageAccessOutcomeMaxRecords` sizes a **different LIVE** cold-access log.

### Scope discipline
Nothing else. No RETIRE-AFTER rows from the buffer table.

**8-byte bss delta vs cluster sum:** claimed reclaim 8+32+4096+2048+10737472 = 10,743,656; measured `bssSizeBytes` 0x1c101b60→0x1b6c2c00 = 10,743,648. Shortfall exactly 8 = alignment padding absorbed at the new region boundary (all five symbols absent from symbol-addresses.tsv; nothing survived).

### Layout regen (fixpoint, not hand-arithmetic)
- Source: drop five `.zero` lines in `BlockVerdictDataSection.lean`; drop unused `bsrMaxAccountAccessOutcomes` / `bsrMaxStorageAccessOutcomes`.
- `gen-symbol-addresses.py --build` → `asm_to_program.py guest-addrs` → `guest_image_coverage.py --emit-lean`.
- Pin from ELF: `RegionMap.bssSizeBytes` `0x1c101b60` → **`0x1b6c2c00`**; GuestImage `.bss` end `0xbf211b60` → **`0xbe7d2c00`**.
- `textSizeBytes`/`dataSizeBytes` unchanged (`0x64128` / `0x5370`).
- `check-region-map.sh` GREEN; `lake build` GuestImage OK.

## Measure (random-200, one seed, distinct ELFs)

| leg | commit | guest sha256 | run |
|---|---|---|---|
| parent | `9bac6a15a` (origin/main) | `e551d54cf0d9a26bdc92e8d4e3924777ba4c19320d3af7a9449afe60be7f7f50` | `run-20260802T122101Z-2946560` |
| candidate | `d7a152f6c` | `8507924c9680c93f9074869cd344e2a0257d19cccf0540df5db9051efb9a9b42` | `run-20260802T122101Z-2946561` |

Command (both legs, same seed):
`scripts/codegen-eest-stateless-check.sh --backend spike --limit 200 --jobs 8 --progress --random --seed 42 --guest-elf PATH`

`eest-ab-compare.py`:
- **guest ELF base ≠ candidate** (self-check PASS — not a dead comparison)
- **FA 0→0**, **FR 8→8**, FR label-set EQUAL
- **output-byte differences: 0** (BYTE-IDENTICAL on all 196 joined cases)
- reverse flips: **0**

Structural deletion: null A/B expected; FA=0 required and held.

## Test plan
- [x] check-region-map green after fixpoint
- [x] GuestImage build
- [x] random-200 seed=42 A/B FA=0 byte-identical, ELFs distinct


pirapira:
Labelled. `build` never scheduled before main moved and the head went `CONFLICTING`, so this rests on the local gates plus the reachability proof — the same basis used for #11174 and #11165 today. Conflict resolution and the batch CI run are the merge bot's.

**This deletes 10,743,656 bytes of `.bss` across five symbols, taking headroom from ~34.7 MiB to ~44.9 MiB against ZisK's 512 MiB.** At 93% occupancy that is the difference between a tight budget and a workable one.

## Why the deadness claim is trustworthy

Three independent methods agree, which matters because each one alone has a known blind spot:

- **Emitted reachability**: `la`/`li`/`pcrel_hi`/`pcrel_lo` all zero for every symbol in the cluster — and no absolute hex literal of `0xa5ca19e0`, no `lui` of the reconstructed hi20 `0xa5ca2`. A symbol can be reached without its name appearing, so checking the reconstructed-address form is what makes the name census complete rather than suggestive.
- **Neighbour reach, bounded rather than assumed**: the only live neighbour writer is `bsr_changed_accounts`, whose maximum offset is 2035 inside its own 2048. This is the actual hazard for a data buffer — neighbours arrive by `la`+displacement, not by name — and it is closed with a number.
- **`SPIKE_WATCH` hits=0** at base, at `path_count`, and at **mid-extent +8 MiB**. `hits=0` is the only instrument that proves *nothing ever wrote* a cell; a trace of writes cannot. Sampling the interior is what makes this a claim about the region rather than about one address.

⚠️ **`hits=0` is not universally decisive, and the author demonstrated exactly why by declining to use it that way elsewhere.** On `sps_keys` — same instrument, same zero — it did *not* imply dead, because that buffer has 4 live `la` sites in a routine confirmed entered, so a zero at one 8-byte base on one row is a statement about that row. **`hits=0` proves absence only when nothing addresses the region.** Here the `la` count is zero, so it corroborates; there it doesn't.

Two name traps were also cleared: `bsr_paths` is **live**, one letter away and sitting `0x3fcd60` below the cluster with its own `la`; and the shared `storageAccessOutcomeMaxRecords` constant sizes a **different, live** cold-access log — a shared constant does not make a buffer live. Either would have produced a confident wrong answer, in opposite directions.

## Evidence

Random-200, seed 42, parent `e551d54c…` vs candidate `8507924c…`: **FA 0→0, FR 8→8 with equal label sets, zero reverse flips.** ELFs verified **distinct** with the compare self-check passing — without that, an identical result is indistinguishable from a comparison that never ran a second binary. Equal label *sets* rather than equal counts is the right claim; counts can hide a swap.

`.text` and `.data` unchanged; `.bss` regenerated to fixpoint and **ELF-measured**, `0x1c101b60` → `0x1b6c2c00`. Pins regenerated rather than predicted, which is essential here: the only real hazard in this change is that removing 10 MiB shifts every symbol above it.

**On the arithmetic:** the cluster sums to 10,743,656 B while the measured `.bss` delta is 10,743,648 — **short by exactly 8**, which is exactly the size of `bsr_storage_access_path_count`. That coincidence points straight at "one symbol survived", so I checked: all five are **absent from `symbol-addresses.tsv` on this branch**, zero occurrences each. The 8 bytes is alignment padding at the new region boundary, now noted in the body. Worth recording rather than leaving as an unexplained gap in a PR whose entire claim is *exactly these five and nothing else*.

PR #11182, opened 2026-08-02
codegen: route MTx precompile helpers through live context

pirapira:
## Summary

- Parameterize the shared precompile gas and publish emitters by their 192-byte context label.
- Route the MTx precompile path through its live `bv_mtx_ctx` context.
- Remove the MTx adapter's redundant `bv_mtx_ctx -> bv_simple_transfer_tx` and sender-address copies.
- Preserve lane, status, hook-protocol, and sender-address refresh ordering. The single-tx compatibility wrappers remain unchanged.
- This PR re-homes only and deletes nothing. Adapter deletion is a separate subtraction-only change after this one proves itself.

The context identity is documented by `BlockVerdictMultiTx.lean:5-17,35-42`: `multi_tx_nth_context` and the shared transfer context have the same 192-byte layout. That identity is why the adapter copy was a representation bridge carrying no semantic data.

The execution-spec anchor is pinned `execution-specs` `e5a8caf1b`: Amsterdam `vm/interpreter.py:398-402` selects `PRE_COMPILED_CONTRACTS[code_address]` and invokes the mapped callable, while `utils/message.py:50-91` constructs the message/context fields. This change only re-homes the already-populated MTx context into the shared precompile accounting; selector, per-address kernels, gas, output sinks, and continuation protocol are unchanged.

The post-#11181 producer census found the five obligations at the adapter, with seven lane stores and eight prepare-status stores located image-wide by the same scan. The status-store source breakdown is explicit: `Dispatch.lean` has 2, `BlockVerdictDispatchTx.lean` has 2, `BlockVerdictMtxEoa.lean` has 2, and `BlockVerdictMtxRuntime.lean` has 2; the emitted scan is 8. The identical parent/candidate scan gives 7/8/7 stores (lane/status/post-top-frame), so this re-home changes producer counts by zero; it removes exactly four old-copy label/loop references (the 24-word context copy and 20-byte sender copy) while those producer stores remain present elsewhere. The census therefore is not a grep-miss argument: those stores are present and exercised exactly where expected. The live protocol state is preserved: lane, status 3, and hook state remain intact, including the two emitted `runtime_tx_prepare_prefix_status` reads at `.runtime_tx_post_top_frame_done` and `.Lruntime_dispatcher_regular_loop`.

`bmvmx_sender_addr` is not an orphaned producer. `simple_transfer_intrinsic_gas` calls `address_from_pubkey` with the context plus 24 and rewrites that buffer. Parameterizing the helper preserves that rewrite against `bv_mtx_ctx` rather than dropping it.

## Verification

Non-cache build and link checks passed after rebasing onto `5da6a2f4e91e970aa316a4996a869ac511adeee7` (the #11181 merge result):

- `LAKE_ARTIFACT_CACHE=false lake build codegen` (4294 jobs)
- `scripts/codegen-stateless-link-check.sh`
- `git diff --check`

Behavioral A/B used the same deterministic manifest (random-200, seed `424242`), same runner, and 30 Spike processes. Both legs completed 200/200 with 0 errors, budget 0, FA=0, FR=8, OK=192. The FR=8 value is the post-#11181 baseline for this sample; it is unchanged by this PR.

`python3 scripts/eest-ab-compare.py gen-out/eest-run-11163-parent-v2 gen-out/eest-run-11163-candidate-v2`:

- 200 joined; self-checks PASS
- FA 0 -> 0; FR 8 -> 8 (unchanged)
- forward flips 0; reverse flips 0
- identical 8-label FR sets
- status differences 0; output-byte differences 0
- distinct ELF artifacts

Artifacts:

- base commit: `5da6a2f4e91e970aa316a4996a869ac511adeee7`
- candidate commit: `2ed9d54ff0b4c6719d734597968183e26b0e66d2`
- parent ELF SHA-256: `379f87f4bb17b8f5efac0ee016e8e30ad09535327322222f63c3b9f40877b030`
- candidate ELF SHA-256: `923f27dadd64466764db4eee97f8be8dc1ee3a3b79683e033b19237e87597263`
- runner: `scripts/spike/spike_run`

The candidate is byte-identical to the parent on all 200 selected inputs, as expected for this routing-only re-home. The old FR=8 measurement was against a pre-#11181 base; the post-merge re-measurement is also FR=8, so the unchanged count is explicitly attributable to the moved base rather than silently reused.


pirapira:
Labelled. `build` is still running (started 12:59); labelling ahead of it because the merge bot rebuilds the batch and runs all 15 gates before anything lands, so this cannot merge broken. The rest of the evidence is complete.

**This PR re-homes and deletes nothing.** It routes the MTx precompile helpers at the live context pointer instead of copying into `bv_simple_transfer_tx`. The adapter deletion is a separate PR whose diff will be pure subtraction — that ordering is the whole lesson of #10786, where `bb22567a9` retired a path and took `blockVerdictMtxEoaRecipientCredit` with it, surfacing from a fixture weeks later rather than from the diff.

## Why the copy was safe to remove

`multi_tx_nth_context` and `simple_transfer_tx_context` are documented as **the same 192-byte layout** (`BlockVerdictMultiTx.lean:5-17,35-42`). That identity is the entire justification: the 24-SD copy was a *representation bridge carrying no semantic data*. Had the layouts differed, the copy would have been a genuine producer and this change would be wrong.

Spec basis is the pinned `e5a8caf1b` Amsterdam tree: `vm/interpreter.py:398-402` (selector and invocation) and `utils/message.py:50-91` (message/context construction). Selector, kernels, gas, output and continuation are untouched — only the populated context moves.

## Behaviour

Random-200, seed 424242, on post-#11181 main `5da6a2f4e`. Parent `379f87f4…`, candidate `923f27da…`, ELFs distinct with compare self-checks passing. Both legs 200/200, errors 0, **FA=0, FR=8, OK=192, forward 0, reverse 0, identical FR label set**, zero status and output-byte differences.

Identical label *sets* is the claim that matters — equal counts alone are compatible with a swap, one row fixed and one broken. And `FR=8` is this sample's unchanged baseline, not an achievement of this PR.

## The census, and why it took three attempts to state correctly

The producer census is the basis for the *next* PR, so it had to be right. Three arithmetic problems surfaced, each of which looked like a detail:

1. **Counts moved between censuses** (6→7 lane, 7→8 prepare-status). The old census was on the candidate at the old base, the new one on the candidate at post-#11181 — **two variables changed at once**, so the delta attributed to neither. Resolved by censusing the *parent* at `5da6a2f4e` with the identical scan.
2. **The one-base comparison then settled it**: parent and candidate both have **7 lane / 8 prepare-status / 7 post-top-frame** stores and identical structural labels; only the parent carries the **4 old-copy label/loop references**. ⇒ **The producer-count delta owned by this PR is zero; the copy-loop delta is −4.** That is the precise claim a deletion prerequisite needs, established by measurement rather than argument.
3. **The source breakdown was short by two** (6 against an emitted 8). The missing writers are `Dispatch.lean:2438` (status=2) and `:3101` (status=1). ⚠️ Both matter for what comes next: **neither is a status-3 writer**, so the adapter retains sole ownership of status-3 — the obligation the deletion plan rests on. Had either been status-3, that plan would have needed rewriting.

## Producers deliberately kept

`bv_mtx_precompile_lane`, `runtime_tx_prepare_prefix_status=3`, and `runtime_tx_post_top_frame_fn` stay. `runtime_dispatcher_call` reads status==3 at **two** sites — top-frame post-prepare and the later regular-loop hook — and dispatches the post-frame function rather than clearing it. That is live protocol, not adapter scaffolding.

`bmvmx_sender_addr` looks like a unique producer and is not: `simple_transfer_intrinsic_gas` calls `address_from_pubkey` with `context+24` and **rewrites it**. The parameterisation preserves that rewrite against the new pointer — otherwise this would have re-homed a value meant to be overwritten and produced a stale sender.

Firing controls were run on every scan, so the zeros and exactly-once results are facts about the adapter rather than about the grep.

pirapira:
**Pulling `merge!` — the build failed, and it failed on something real.**

```
DRIFT RegionMap.textSizeBytes: expected 641d8, ELF has 64178
DRIFT symbol-addresses.tsv changed; commit the regenerated snapshot
```

The pinned `textSizeBytes` is `0x641d8`; the ELF built from this branch is `0x64178` — **96 bytes short**. The rebase onto `5da6a2f4e` carried the pre-rebase pins forward without re-running the layout regeneration, so the generated files describe a tree that no longer exists. `symbol-addresses.tsv` is stale for the same reason.

Everything else passed: `asm-to-program` CLEAN across 386 defs, `axioms` OK on 95 witnesses with only the three classical axioms, `arithmetic-fuzz` PASS, and every structural invariant green including the `.bss`/`.sszscratch`/RAM-ceiling bounds and the `call_frame_arena` union.

**Fix: regenerate to fixpoint, never hand-set.** `RegionMap.textSizeBytes` is an emission *input* — writing `64178` by hand would be consumed rather than validated, and the next build could then drift again from the new value. Re-run the regen loop (relink → TSV/GuestAddrs/GuestImageEntries → repin RegionMap and the `GuestImage` `.bss` literals → repeat until stable), commit the regenerated snapshot, and I relabel on green.

The A/B evidence is unaffected — it was measured on the correct candidate ELF, and this is a generated-file staleness problem, not a behavioural one. The census attribution stands too.

---

**My error, and I want it recorded rather than quietly fixed.** I labelled this while the build was still running, reasoning that labelling doesn't merge and the batch rebuild would catch anything broken. That reasoning was wrong in a specific way: the batch would have hit *this same drift*, burning a full batch cycle and blocking the other PRs queued with it. The gate existed precisely to catch a stale pin after a rebase, and I pre-empted it on the strength of evidence that could not have detected the problem — the A/B was green because the ELF was fine; it was the *committed pins* that were wrong.

I have labelled two other PRs today ahead of a build (#11174, #11179), justified there because a `CONFLICTING` head does not schedule one at all. That justification does not extend to a `MERGEABLE` head with a build actively running, which is what this was.

pirapira:
**Re-labelled at `aa447c775`.** The pin drift that failed the earlier build is repaired: fixpoint regen (non-cache) gives `.text = 0x64178`, `.data = 0x5370`, `.bss = 0x1c101b80`, with `RegionMap`, `symbol-addresses.tsv` and `GuestAddrs` refreshed and the `GuestImage` `.bss` literals unchanged. `check-region-map` PASS and `check-asm-to-program` PASS (386 defs / 147 files) locally.

The head is `CONFLICTING` against `7d9824dec` — #11136 landed while this was being repaired — so CI will not schedule a build here at all. Labelling on the local gate set, as with #11174 and #11179 today; the merge bot rebases and regenerates as part of the merge, and the batch build is the real gate.

⚠️ **The rebase will move `textSizeBytes` again.** #11136 was itself a deletion that repins the layout, so the merged tree's `.text` will not be `0x64178`. That is expected and is the merge bot's regen to perform — **it must not be hand-set**, since `RegionMap.textSizeBytes` is an emission *input*: a hand-written value is consumed rather than validated and yields a confident wrong image under a green build. That is exactly the failure this PR already hit once.

**Behavioural evidence is unaffected by the pin repair** — it was measured on a correctly-built candidate ELF, and stale generated files are not a behavioural change. Both legs on base `5da6a2f4e`, seed 424242, distinct ELFs with self-checks passing: 200/200, errors 0, **FA=0, FR=8, OK=192, forward 0, reverse 0, identical FR label set**, zero status and output-byte differences. `FR=8` is this sample's unchanged baseline, not a result of this PR.

---

For the record, my earlier label on this PR was premature and the build caught a real defect. I reasoned that labelling does not merge, so the batch rebuild would catch anything broken — but the batch would have hit this same drift, burning a cycle and blocking every PR queued with it. The A/B could not have detected it: the ELF was fine, the **committed pins** were wrong. The difference now is that the failing gate has been run locally and passes.

PR #11184, opened 2026-08-02
chore(guest): batch-delete three proven-dead symbols (#11170/#11172/#11178)

### LINKED ISSUE #11170: block_verdict: .Lbv_receipts_no_runtime_gas is a dead ACCEPT arm (0 branch targets) that skips the bv_dispatch_runtime_status gate — latent FA if wired
## Claim

**`.Lbv_receipts_no_runtime_gas` is a dead ACCEPT arm in `block_verdict` with zero branch targets. It is the only accept arm that does not consult `bv_dispatch_runtime_status`, so wiring it would accept a block whose dispatch runtime status is nonzero — a false accept.**

Not a live defect: nothing branches to it today, so it never executes. Filed because **FA never relaxes**, and because it was previously recorded only inside GH #11120 (gas-validation), which is now closed — closing that issue would have orphaned the warning. This is the arm's own home.

Base **`696c236f2`** (`main`, the #11139 merge).

* guest asm sha256 `d6382d9e1adefbb7f471461fdb9a50dea7a2240574c9062c8d32de7ff2c5ea6a`
* guest ELF sha256 `f4d016e6ffc7bad4d33cac949befef58a178c0fc628f65b4a0a923e630784522`
* rerun: `lake exe codegen --program stateless_guest --halt linux93 -o gen-out/regionmap/stateless_guest`

## Measured

**Zero branch targets**, in any form (`beqz`/`bnez`/`beq`/`bne`/`blt`/`bge`/`bltu`/`bgeu`/`j`/`jal`). The bare token `Lbv_receipts_no_runtime_gas` occurs **exactly once** in the whole emitted image — its own label definition.

⭐ **Positive control, so the search is known able to fire:** the same grep against the live sibling `.Lbv_receipts_accept` returns **2** branch targets (`.s 17635` `beqz` on `bv_receipts_enforce_enabled`, and `.s 17649` `j`).

## The two arms, side by side — this is the whole finding

```asm
.Lbv_receipts_accept:                                       ; LIVE, 2 branch targets
  la t0, bv_dispatch_runtime_status; ld t0, 0(t0)
  bnez t0, .Lbv_dispatch_runtime_status_fail                 ; <-- THE GATE
  li a0, 1; j .Lbv_ret

.Lbv_receipts_no_runtime_gas:                               ; DEAD, 0 branch targets
  la t2, bv_exec_p; ld a0, 0(t2)
  li a1, 0
  li a2, 0
  li a3, 0
  li a4, 0
  jal ra, block_receipt_records_materialize                  ; <-- zeroed gas arguments
  li a0, 1; j .Lbv_ret                                      ; <-- accepts, NO gate
```

`bv_dispatch_runtime_status` is written at `.s 16321` and has **exactly one reader in the image**: `.s 17653`, inside `.Lbv_receipts_accept`. ⇒ **the gate exists on precisely one arm, and this is the arm that bypasses it.**

`block_receipt_records_materialize` has exactly **two** call sites: `.s 17582` (`.Lbv_mtx_b2_return`, the live one, with real arguments) and `.s 17661` (this arm, with `a1..a4 = 0`).

## ⚠️ Scope correction: the exposure is NARROWER than #11120 described, and I am not restating its version

#11120 framed the danger as *"it would accept without gas and, if `shadow_ready` were still 0, without Path A."* **That specific mechanism is now closed by #11139** and should not be re-quoted:

* `bv_bal_shadow_ready` now has exactly one writer (`.s 15945`) and is set **unconditionally** during the early SSZ decode, no longer gated on gas success;
* this arm ends `j .Lbv_ret`, so wiring it would **still** run `bal_gas_valid_from_builder` and **still** hit the Path A digest bind.

⇒ **What remains is simpler and still FA-shaped:** an accept that never consults `bv_dispatch_runtime_status`. A block whose runtime dispatch failed would be accepted. The zeroed `a1..a4` receipt materialisation is a second, separate concern on the same arm.

## Why a reference count cannot find this

⭐ **`.Lbv_bal_gas_fail` is also dead with zero branch targets — and it is harmless.** Its body is `li t0, 7; bv_fail_code := 7; j .Lbv_zero`: **reject-shaped**, so wiring it would reject.

Two arms, identical emitted symptom (dead, zero branch targets), **opposite consequence**. Only reading the arm's body distinguishes a latent false accept from harmless dead code. ⇒ a dead-code census that counts references will rank these two the same, and any sweep for dead arms should classify by **terminal shape** (`li a0, 1` / `j .Lbv_ret` versus `j .Lbv_zero`), not by reference count.

## Closing condition

Either:

1. **Wire it to reject** — if the no-runtime-gas case is a real state that should be reachable, give it the `bv_dispatch_runtime_status` gate and a fail code, so it cannot accept ungated; **or**
2. **Delete it** — if it is obsolete scaffolding, remove the arm (and the now-unreachable second `block_receipt_records_materialize` call site with it).

Either way the check that settles it is: **no accept arm in `block_verdict` may set `a0 = 1` without consulting `bv_dispatch_runtime_status`.** Today that is one arm out of two.

⚠️ Deleting the arm shrinks `.text` and therefore **forces a repin**; per the #11054 precedent, prefer riding a change that already needs one rather than paying the cascade for this alone.

## Non-goals

* No BAL / gas-validation work — #11120 is closed and its four-way verification stands.
* Do not re-open the `shadow_ready` mechanism; #11139 closed it.
* Not a claim of a live false accept. Nothing reaches this arm today.

### LINKED ISSUE #11172: codegen: bal_gas_valid is dead on the block_verdict path after #11139 but still pinned/verified/probed — retire on the next repin-forcing change
## Status: dead on the `block_verdict` path, **not** an orphan — retire on the next repin-forcing change

Follow-up to #11139 (merged) and #11120 (closed). Filed so this is **visible to whoever next repins**, which is the pattern that worked for #11054: that issue carried the evidence until a repin-forcing change came along, and the deletion rode `da930613c` with #10938 piece 3, paying the cascade once.

Base **`696c236f2`** (`main`, the #11139 merge).

* guest asm sha256 `d6382d9e1adefbb7f471461fdb9a50dea7a2240574c9062c8d32de7ff2c5ea6a`
* guest ELF sha256 `f4d016e6ffc7bad4d33cac949befef58a178c0fc628f65b4a0a923e630784522`

## What changed

#11139 moved BAL gas validation from the **declared** payload bytes to the **built** list, replacing `bal_gas_valid` with `bal_gas_valid_from_builder` at `.Lbv_ret`. The old walker was left in the image.

## Measured: uncalled, and it never executes

* **Zero references** in any address-materialising form — `jal`/`jalr`/`j`/`la`/`call`/`tail`/`auipc`. The bare token `bal_gas_valid` occurs **exactly once** in the emitted image: its own label at `.s 18965`.
* ⭐ **Dynamically confirmed, with the control in the same log.** `SPIKE_COMMITLOG` over one fixture (~20M instructions committed): `bal_gas_valid` at `0x8001d28c` executes **0 times**, while `bal_gas_valid_from_builder` at `0x8001d434` executes. Both readings from one log, so the instrument is demonstrably able to distinguish executed from not — the zero is not an instrument failure.

## ⛔ But it is NOT deletable on the name argument, and that is the point of this issue

Unlike the three orphaned fixtures retired in #11169, this symbol has **live infrastructure**:

| artefact | site |
|---|---|
| `GuestAddrs` pin | `GuestAddrs.bal_gas_valid = 0x8001d28c` (`GuestAddrs.lean:139`) |
| image-entries proof pairing | `GuestImageEntries.lean:316` — `(GuestAddrs.bal_gas_valid, balGasValid_prog)` |
| separate probe BuildUnit | `zisk_bal_gas_valid` — `RegistryTail.lean:83`, `RegistryNames.lean:376`, `RegistryNamesTail.lean:14` |
| gated conversion fixtures | **4** MANIFEST rows for `BalGasValid.lean`: `balGasValidFunction`, `balSectionInfoFunction`, `bgvU32leFunction`, `bgvU64leFunction` |

⇒ it is a **verified, image-pinned `Program` with its own probe build unit**, reachable outside `stateless_guest`. "Zero call sites on the `block_verdict` path" is therefore **not** sufficient grounds to delete it, and a name-based dead-code census would misclassify it.

⚠️ Same discipline as the ~10 MB of name-unreferenced BSS: zero references is a complete argument only for a routine with **no other artefacts**. This one has four kinds.

## Decision required (this is a judgement call, not a cleanup)

1. **Retire it** — drop `balGasValid_prog`, its `GuestAddrs` pin, the `GuestImageEntries` pairing, the probe unit, and the 4 fixture rows. Shrinks `.text` ⇒ **forces a repin cascade**. ⚠️ Also removes a verified Program, so the verified-surface count drops; that is a deliberate loss, not a saving.
2. **Keep it** as an independently-probed verified routine that no longer serves the verdict path, and record why — so the next reader does not re-file this.

## If option 1: ride a repin-forcing change

Do **not** pay the cascade for this alone. Per #11054's precedent, batch it onto a change that already moves the text pin. When it lands, the three-artefact rule applies (**Lean def, MANIFEST row, fixture file**) — #11169 exists because a previous retirement removed two of the three and left the fixture ungated. The bijection gate added in #11169 (`asm_to_program.py check-all`) will now catch that specific miss.

⚠️ And note `bal_gas_valid` carries **4** fixture rows, not one: the three helpers (`balSectionInfoFunction`, `bgvU32leFunction`, `bgvU64leFunction`) must each be checked for other users before removal — `bgv_u32le` and `bgv_u64le` in particular are called from the live early-decode path, so **they must NOT be removed with it**.

## Non-goals

* Not a correctness defect. Nothing reads a wrong value; the routine simply does not run.
* No BAL gas semantics work — #11120 verified `bal_gas_valid_from_builder` four ways and is closed.

### LINKED ISSUE #11178: codegen: exec_log_latest_value has zero call sites in stateless_guest but is pinned, verified and probe-linked — retire or document, on the next repin
## Status: **zero call sites in `stateless_guest`, but linked, pinned, verified and probed** — the #11172 shape, not an orphan

Filed so this has a home rather than living in a census comment (the #11176 lesson). Found incidentally while censusing the storage preload. **Read-only; nothing deleted, and it is deliberately not attached to any PR.**

Base **`9bac6a15a`** (`main`). Guest ELF sha256 `2e336fac2153d424cb85fa67…`; emitted via `lake exe codegen --program stateless_guest --halt linux93 -o gen-out/regionmap/stateless_guest`.

## Measured: nothing calls it

`exec_log_latest_value` is a **routine** — `.s 25563-25606`, **43 lines**, a loop over 128-byte execution-log rows (`slli x6, x5, 7`), ELF symbol `0x80027734`, `LOCAL`.

* the bare token occurs **exactly once** in the emitted image — its own label;
* **0** references in any address-materialising form (`jal`/`jalr`/`j`/`la`/`call`/`tail`/`auipc`);
* ⭐ **control:** the same grep on the live sibling `exec_log_slot_tuples` returns **3**, so it can fire.

## ⛔ But it is NOT an orphan, and "zero references" is NOT the complete argument here

It carries the full infrastructure set — the same distinction as #11172's `bal_gas_valid`:

| artefact | site |
|---|---|
| **emitted** into the image | `BlockVerdictV2.lean:184` and `:546` — `execLogLatestValueFunction ++` (**two** build units) |
| `GuestAddrs` pin | `GuestAddrs.exec_log_latest_value` |
| image-entries proof pairing | `GuestImageEntries.lean:360` — `(GuestAddrs.exec_log_latest_value, execLogLatestValue_prog)` |
| ⚠️ **registered probe BuildUnit** | `ziskExecLogLatestValueProbeUnit` (`ExecLogLatestValue.lean:120`), wired at `RegistryTail.lean:92` and `RegistryNamesTail.lean:26` as `zisk_exec_log_latest_value` |
| gated conversion fixture | `MANIFEST.tsv:187` + `scripts/asm-fixtures/execLogLatestValueFunction.s` |
| own module | `EvmAsm/Codegen/Programs/ExecLogLatestValue.lean` |

⚠️ **The probe unit matters most:** because `zisk_exec_log_latest_value` exists, the routine **is reachable in that build unit**. So "zero references" is a statement about **`stateless_guest`**, not about the program universe — *guest linking is not the units' linking*. A complete deadness argument would have to consider every build unit, and I have not done that.

## Why it probably went dead — a lead, not a finding

`BlockVerdictV2.lean` labels it *"cross-tx storage threading lookup"*. #11142 retired the duplicate committed-storage map, and the sibling probes `zisk_mtx_committed_chunked_latest_value` / `_snapshot_upsert` sit right beside it in the registry. ⚠️ **Plausibly a casualty of that retirement — not verified.** Whoever takes this should check whether the cross-tx threading obligation moved elsewhere or was retired outright, because that decides between the two options below.

## Decision required

1. **Retire it** — remove `execLogLatestValue_prog`, the `GuestAddrs` pin, the `GuestImageEntries` pairing, the probe unit, the MANIFEST row and the fixture. ⚠️ Shrinks `.text` ⇒ **forces a repin**, and removes a verified `Program`, so the verified-surface count drops. That is a deliberate loss, not a saving.
2. **Keep it** as an independently-probed verified routine no longer on the verdict path, and **record why** — so the next reader does not re-file this.

## If option 1: sequencing and the three-artefact rule

Ride a change that already forces a repin (the #11054 precedent). The three artefacts are **Lean def, MANIFEST row, fixture file** — #11169 exists because a previous retirement removed two of three and left the fixture ungated; the bijection gate added there will now catch that specific miss.

⇒ ⛔ **Do not bundle this with the storage-preload retirement (#11176) without a stated reason.** They are independent findings that happen to share a family, and merging them makes one A/B answer for two changes.

## Non-goals

* Not a correctness defect — nothing reads a wrong value; the routine simply never runs in `stateless_guest`.
* No claim about the other `exec_log_*` routines. `exec_log_slot_tuples` (3 callers) and `exec_log_addr_to_bal_canonical` (2 callers) are **live**.

pirapira:
## Summary

Batch-delete three proven-dead guest symbols after #11136 landed on main.

| issue | symbol | kind | reclaim |
|---|---|---|---|
| **#11170** | `.Lbv_receipts_no_runtime_gas` | dead ACCEPT arm | text |
| **#11172** | `bal_gas_valid` | dead TEXT function (not data) | −424 B text |
| **#11178** | `exec_log_latest_value` | dead TEXT function | −168 B text |

**Lead with #11170 (latent FA surface):** this arm was the sole ACCEPT path that bypassed the `dispatch_runtime_status` gate. Zero branch targets reached it today, but wiring it would have been a false accept. Deleting removes that surface rather than housekeeping a label.

### #11172 KEEP-LIST (explicit — re-counted on post-#11136 main)

| symbol | role | refs after #11136 |
|---|---|---|
| `bal_gas_valid_from_builder` | live Path-C gas-on-built at Lbv_ret | def=1 jal=1 **KEEP** |
| `bgv_u32le` | RLP u32 helper | jal=**117** **KEEP** |
| `bgv_u64le` | RLP u64 helper | jal=**26** (was 27 pre-#11136; still live) **KEEP** |
| `bal_section_info` | section walker helper | jal=2 **KEEP** |
| `bal_gas_valid` | old RLP walker on declared bytes | def=0 jal=0 nm ABSENT **DELETE** |

#11136 unlinked `bal_storage_reads_in_exec_log` / code-consistent routines; KEEP-LIST survivors still have callers. Probe aggregate still includes `balGasValidFunction` for `zisk_bal` probes via `balGasValidPc := 0x80000000` placeholder (same pattern as #11103 / #11136).

### Per-symbol evidence (settle on main 5da then re-checked post-#11136)

1. **#11178 `exec_log_latest_value`:** emitted refs=0; SPIKE_BREAK_PC hits=0; control `exec_log_slot_tuples` still jal=3; fallthrough predecessor ends in `ret`. nm ABSENT after unlink.
2. **#11172 `bal_gas_valid`:** emitted refs=0; SPIKE_BREAK hits=0; control `bal_gas_valid_from_builder` HIT on live path; nm ABSENT after unlink. **Routine not data** — zero-ref is complete argument.
3. **#11170 `.Lbv_receipts_no_runtime_gas`:** 0 branch targets; accept arm already `j .Lbv_ret` (no fallthrough into it); sole bypass of `dispatch_runtime_status` if ever wired.

### Layout (regen fixpoint after merge main 7d9824dec — never hand-set)

- Base: origin/main `7d9824dec` (#11136)
- Branch head: includes merge + layout regen
- `textSizeBytes` **0x063c9c → 0x063a24** (ELF-measured; −0x278 = 632 B)
- `dataSizeBytes` 0x5370 unchanged; `bssSizeBytes` 0x1c101b40 unchanged
- `check-region-map` GREEN; GuestAddrs/TSV/GIE regenerated to fixpoint

### A/B re-derived on main 7d9824dec (NOT carried from pre-#11136)

Manifest: `/tmp/11167-random200.tsv` seed=42 random-200.
Spike: `/tmp/grok-main-11136-check/scripts/spike/spike_run`

| leg | commit / tree | guest sha256 | text |
|---|---|---|---|
| parent | main `7d9824dec` | `3deae671ae5c4c217b7f125370c89bc27629e31e533287e03b18b1a063b3996f` | 0x063c9c |
| cand | this PR | `97f82c86415f2b0abdf5b479ee6ec5ff41195dfb8fdd52291b2db13b73fc3143` | 0x063a24 |

- ELFs **DISTINCT** (`cmp` differs)
- **FA 0 → 0**, **FR 5 → 5**, FR label-sets **EQUAL**, category flips **0**, guest_succ reverse **0** / forward **0**
- **OUTPUT 256 B byte-identical on 200/200** (stronger than equal FR set)

Artifacts: `/tmp/grok-batch-dead-3/`

### Files

- `BlockVerdictReceiptsTail.lean` — delete no_runtime_gas arm
- `BlockVerdictV2.lean` / `BlockVerdict.lean` — unlink balGasValidFunction + execLogLatestValueFunction; keep from_builder
- `BalGasValid.lean` — balGasValidPc placeholder for probe Program
- `ExecLogLatestValue.lean` — probe-only comments
- layout: RegionMap, GuestAddrs, GIE, symbol-addresses.tsv
- `asm_to_program.py` — already has probe-only SYMMAP exclusion from #11136

Closes #11170. Closes #11172. Closes #11178.

No `merge!` from author.


pirapira:
Opened after #11136 on main. A/B re-derived on 7d9824dec: FA0 FR5 byte-identical 200/200. KEEP-LIST recounted post-#11136.

pirapira:
**Reviewed — evidence accepted. Holding the label until `build` reports green;** it has not scheduled yet. I labelled a MERGEABLE head ahead of its build earlier today and the build then caught a real pin drift, so on a mergeable head the build now gates the label.

**The strongest evidence here is the one that isn't a count: 256 output bytes byte-identical on 200/200 rows.** For a deletion that is the right claim — not "the same number of rows failed" but "nothing observable changed at all". FA 0→0, FR 5→5 with **equal label sets**, zero flips in either direction, ELFs verified distinct. Equal label sets rather than equal counts matters because a count can hide a swap.

**The keep-list re-count is why the rebase was worth insisting on.** `bgv_u64le` moved **27 → 26** jal sites because #11136 removed a caller. Still live, so the conclusion holds — but the number changed, and carrying the pre-#11136 figure forward would have put a value in this PR that its own tree contradicts. The same re-count is what confirms `bgv_u32le` (117), `bal_section_info` (2) and `bal_gas_valid_from_builder` (1) still have callers after a sibling deletion landed.

**Per-symbol the arguments are complete and correctly differentiated:**

- **#11178 / #11172** — routines, so `refs=0` is a *complete* argument with no layout reasoning required. Each has a control that fired (`exec_log_slot_tuples` at 3 jal; `bal_gas_valid_from_builder` HIT on the live path), so the zeros are facts about the symbols rather than about the search.
- **#11178** also checks the thing people skip: the **fallthrough predecessor ends in `ret`**. A dead label reached by fallthrough is not dead.
- **#11170** leads on the right justification — it was the **sole ACCEPT path bypassing the `dispatch_runtime_status` gate**. Zero branch targets reach it today, but wiring it would have been a false accept, so this removes an FA surface rather than tidying a label.

⚠️ **`bal_gas_valid` was recorded as data and is a 424-byte text function.** That classification came from me, and had it been accepted it would have sent the settle into a neighbour-extent analysis that does not apply to a routine — either wasting the work or producing an inconclusive UNKNOWN on something cleanly provable. Checking the emitted form rather than the label is what avoided it.

Layout is regenerated to fixpoint and ELF-measured (`0x063c9c → 0x063a24`, −632 B), `.data` and `.bss` unchanged, `check-region-map` green — never hand-set, which is the failure mode that cost a build here today.

pirapira:
Build **SUCCESS** — labelled as stated above.

Recording the one number that changed on the rebase, because it is the reason the rebase was required rather than optional: `bgv_u64le` moved **27 → 26** jal sites when #11136 landed. Still live, so the keep-list conclusion holds — but carrying the pre-#11136 count forward would have put a figure in this PR that its own tree contradicts.

Landing this closes #11170, #11172 and #11178 together, with #11170 removing a latent false-accept surface rather than dead weight.

PR #11188, opened 2026-08-02
fix(codegen): seed_callee capacity guards fail closed (#11157)

pirapira:
## Summary

Closes the #11157 false-accept surface: `seed_callee_storage`'s three capacity guards no longer **continue** past overflow with a bare `li 128`.

They now:

1. **Derive** from the buffer they protect (not a magic number).
2. **Fail closed** (`a0 = 1` → caller `.Ldtrc_bal_unsupported`).

| guard | was | now |
|---|---|---|
| changes count vs `csce_keys` | `li 128; bgtu → .Lscs_acct_next` (skip account) | `li bsrAccountSlotCap` (= `bsrMaxBalItems` = gas/BAL_ITEM); `→ .Lscs_cap_fail` |
| changes+reads combined | same skip | same derived cap; fail closed |
| `callee_seed_count` table full | `li 128; bgeu → .Lscs_done` (stop silently) | `li calleeSeedTableCap`; fail closed |

`calleeSeedTableCap` / `calleeSeedTableBytes` live in `BlockVerdictParams`; Dispatch `.zero` sizes from the same constant. **No table growth** (still 128 × 96 B = 12288) — layout RAM unchanged; text only (`textSizeBytes` `0x063c9c → 0x063cb8`).

Caller:

```asm
jal ra, seed_callee_storage
bnez a0, .Ldtrc_bal_unsupported
```

## Force-fire proof (guard untested by corpus)

Corpus max admitted per-account keys ≈ 4; table peaks ≈ 19 — neither guard fires naturally.

Construction: binary-patch table immediate `li t3,128` → `li t3,0` on cand ELF; run pass row `00103_test_value_move_to_precompiles…`:

| PC | event |
|---|---|
| `0x80025a80` | table `bgeu` (patched t3=0) |
| step → `0x80025fc0` | `.Lscs_cap_fail` (`li a0,1`) |
| ret → `0x80023678` | caller `bnez a0` |
| step → `0x80023ad8` | `.Ldtrc_bal_unsupported` (`li a0,3`) |
| OUTPUT[32] | **succ=0** |

Unpatched A/B is null (below); patched run proves the reject edge is live.

## A/B (random-200 seed=42, `/tmp/11167-random200.tsv`)

|---|---|---|---|---|
| parent main `7d9824dec` | `3deae671ae5c4c21…3996f` | 0 | 5 | 195 |
| cand | `46b6f612edad0b1a…d9ca` | 0 | 5 | 195 |

FR sets **equal**; guest_succ diffs **0**; reverse **0**. ELFs distinct. Artifacts `/tmp/grok-11157/`.

## Files

- `BlockVerdictParams.lean` — `calleeSeedTableCap/Bytes` + guards
- `BlockVerdictDispatchTx.lean` — seed_callee + caller
- `Dispatch.lean` — table `.zero` from constant; import Params
- `BlockVerdictDataSection.lean` — comment
- layout: `RegionMap.textSizeBytes`, `GuestAddrs`, TSV

## Test plan

- [x] check-region-map green
- [x] A/B FA0 FR5 reverse 0
- [x] force-fire reject path
- [ ] CI build


pirapira:
## Force-fire evidence (post-open)

Construction proof on cand guest `46b6f612…d9ca` (same shape after layout pin):

1. Binary-patch table immediate at `li t3,128` (`0x80025a7c`) → `li t3,0`.
2. Row `00103_test_value_move_to_precompiles…` hits table guard `0x80025a80`.
3. Step → `.Lscs_cap_fail` `0x80025fc0` (`li a0,1`).
4. Return to caller `0x80023678` `bnez a0` → `.Ldtrc_bal_unsupported` `0x80023ad8` (`li a0,3`).
5. OUTPUT succ byte = **0**.

Unpatched A/B remains FA0 FR5 reverse 0 (body). No `merge!` from grok.

pirapira:
Labelled. `CONFLICTING` against `9298b78fb`, so CI will not schedule a build; this rests on the local gates and the force-fire proof, as with #11174, #11179 and #11182 today. The merge bot rebases and regenerates.

**This closes a false-accept surface at zero RAM cost, which is the rarest kind of fix here.** The three guards were `li 128` with a *continue* — skip the account, or stop the table silently — while the buffers they nominally protect are sized from `bsrMaxBalItems`. Nothing punished the wrong choice, so the wrong choice persisted.

**Two things were fixed, not one, and the second is what stops it recurring:**

1. **Fail closed** — `a0 = 1` → `.Ldtrc_bal_unsupported`. Overflow now rejects instead of silently dropping seeds.
2. **Derived, not renumbered** — the caps come from `bsrAccountSlotCap` / `calleeSeedTableCap` in `BlockVerdictParams`, with the `.zero` sizing from the same constant. ⇒ **Replacing `128` with `100000` would have fixed today's instance and preserved the mechanism**: a bare literal nobody can re-check against the buffer it guards.

**The force-fire proof is the part that makes this reviewable at all.** The corpus admits ≈4 keys per account against a table peaking at ≈19, so **neither guard fires naturally** — a passing A/B says nothing about a branch that never executes. Patching `li t3,128` → `li t3,0` and walking the path settles it:

`0x80025a80` table `bgeu` → `.Lscs_cap_fail` (`li a0,1`) → caller `bnez a0` → `.Ldtrc_bal_unsupported` (`li a0,3`) → **`OUTPUT[32] succ=0`**

That demonstrates the reject edge is live end-to-end, not merely that a new branch compiles. A guard nobody has seen fire is not known to work.

⚠️ **One item this hands to #11186 rather than resolves.** The callee-seed table stays at 128 × 96 B — deliberately, since growing it is a layout change and this PR is text-only. So the guard moves from **UNDERSIZED + CONTINUE** (unsound) to **UNDERSIZED + REJECT** (sound, but a false-reject surface if a valid block needs more than 128 callee seeds). That is strictly better and it is not the end state: #11186 owns whether 128 survives a 200M-gas derivation. **The cap has not been shown adequate — only made safe.**

Layout is text-only (`0x063c9c → 0x063cb8`), no `.bss` growth, so this does not contend with the other layout PRs in flight.

PR #11193, opened 2026-08-02
fix(codegen): publish terminal state root on every block path

pirapira:
## Summary

Publish the terminal state root on every block path in the emitted stateless guest.
The root is still computed from the final state, including the canonical storage
map; this is a placement fix, not a reversion to BAL-derived state.

The previous terminal publication sat after a path that zero-execution blocks do
not reach. That is the root cause in one line: wrong-stage, not wrong-value.

The reference behavior computes the state root from the final BlockDiff without
branching on transaction count: [execution-specs fork.py at the pinned reference](https://github.com/ethereum/execution-specs/blob/e5a8caf1b/src/ethereum/forks/amsterdam/fork.py#L352-L360).

## Verification

The behavioral A/B measurement base is `7d9824dec`. Both legs were run on that
same declared base (candidate commit `cb601068d`), using Spike, fixture tag
`tests-zkevm@v0.6.2`, random seed `1065102`, 200 uniformly sampled inputs, 8
jobs, and the SpecRef oracle. The merge base is `92cb99e53`, which is newer than
the measurement base because #11182 and #11192 are layout-only merges that
landed afterward. The branch was rebased onto that merge base for CI
mergeability, and its layout files were re-derived from the combined ELF; the
behavioral A/B was not rerun, so the attribution measurement remains explicitly
on `7d9824dec`.

| leg | ELF SHA256 | full | root | success | tail | oracle | FA | FR |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| parent | `3deae671ae5c4c217b7f125370c89bc27629e31e533287e03b18b1a063b3996f` | 195/200 | 200/200 | 195/200 | 200/200 | 195/200 | 0 | 5 |
| candidate | `129210a6832ee431dbdf7e7c0cbf2638cc1a81d8b18212b7a63516dc11dfd995` | 195/200 | 200/200 | 195/200 | 200/200 | 195/200 | 0 | 5 |

Both legs had 0 errors and 0 budget exhaustion. The exact false-reject set is
identical in both legs:

- `00033_test_callcodecallcodecall_110_fork_Amsterdam-blockchain_test_from_state_test__b0`
- `00085_test_pointer_to_precompile_fork_Amsterdam-precompile_0x0000000000000000000000000000000000000010-`
- `00092_test_create_init_fail_bad_jump_destination2_fork_Amsterdam-blockchain_test_from_state_test__b0`
- `00120_test_pointer_reverts_fork_Amsterdam-blockchain_test_from_state_test-sender_is_auth_signer_True-s`
- `00185_test_withdrawal_requests_fork_Amsterdam-blockchain_test-single_block_multiple_withdrawal_request`

The previously targeted 23-row regression population was rechecked on this
base: 23/23 remain gone. This is a 23-of-23 denominator, separate from the
random 200-row denominator. The map census is canonical_map=166 and
no_canonical_map=34; the 11 remaining no-map rows are expected rejects and do
not require a root publication.

The direct 00073 diagnostic showed the stage error: the parent had 8
`sv_recomputed` hits, with the first root-limb write at `0x800133ac`, while the
candidate had 0 hits and final `sv_recomputed=0`. The fresh same-base A/B has
full agreement on that row.

## Checks

- `lake build codegen specref-eest-check` passed for both parent and candidate.
- Candidate and parent provenance both report `repo_dirty=0`.
- `git diff --check` passed.
- After rebasing onto merge base `92cb99e53`, `lake build` passed (4798 jobs).
- `scripts/check-asm-to-program.sh` passed: 386 converted definitions across 147 files.
- `scripts/check-region-map.sh` passed against the re-derived ELF after two complete regeneration passes. Final measured sections are `.text=0x064378`, `.data=0x5370`, `.bss=0x1b6c2ca0`; the GuestImage `.bss` end is `0xbe7d2ca0`, and `symbol-addresses.tsv` matches the ELF.


pirapira:
**Reviewed — evidence is complete. Holding the label until `build` reports green** (it has not scheduled yet). On a MERGEABLE head the build gates the label here: I labelled one ahead of its build earlier today and the build then caught a real pin drift.

**This is a structural change that costs nothing on the corpus.** Both legs on shared base `7d9824dec`: 200/200 ran, 0 errors, **FA 0, FR 5 on both**, with an **identical FR label set** — `00033`, `00085`, `00092`, `00120`, `00185`. Identical *sets* rather than equal counts is the claim that matters; equal counts are compatible with a swap.

## Why the regression that appeared mid-investigation is not in this PR

An earlier candidate introduced **23 false rejects**. They are gone — 23 of 23 — and the diagnosis is worth recording because the route mattered more than the patch:

1. A same-ELF rerun was **rejected as a reverse leg**. Both runs used one binary, so an identical result was guaranteed by construction. Had it been accepted, 29 would have read as pre-existing and a 23-row regression would have merged as an improvement.
2. The **parent leg** then showed 6 against 29 — real, and attributable once both bases were confirmed identical.
3. The **map census** showed all 23 were `no_canonical_map`: they never execute the new path at all. ⇒ That eliminated the entire new mechanism from suspicion, and killed a cross-lane "shared 2935 substrate" hypothesis I had floated on weaker evidence.
4. The **placement probe** settled it directly on row `00073`: parent `sv_recomputed` **hits=8** with the first limb write at `0x800133ac`; candidate **hits=0, final=0**. Parent versus candidate *on the same row* — a bare candidate zero would prove nothing where something does write that address.

⇒ **Wrong-stage, not wrong-value.** Terminal root publication sat where zero-execution blocks never reach it. That looks like a computation bug from every angle except reachability, which is why the census rather than a value diff cracked it.

## The honest denominator

**23 of 23, not 23 of 200.** `canonical_map=166`, `no_canonical_map=34`, and the 11 remaining no-map rows are **expected rejects** that need no root — a valid control group, not evidence of another publish site. Every accept-expecting row off the canonical map path failed. I initially read those 11 as proof a reachable publish path existed; the oracle `succ` bit says otherwise, and that correction came from the author.

## Placement, not reversion

Stated explicitly and worth confirming here: the root remains **final-state derived, not BAL-derived**. Anyone reading FR 29 → 5 will reasonably wonder whether the structural gain was traded back for the number. It was not — the publish moved to a path every block reaches, which is what `fork.py:352-360` requires, since the spec computes the root from the final `BlockDiff` with **no branch on transaction count**.

⚠️ The base is **intentionally stale** behind current main, declared in the body, per my ruling to stop the rebase treadmill — both legs share it, which is the only property attribution needs. The merge bot rebases and regenerates layout.

pirapira:
Labelled at `20d562fb1`. Main moved again (#11192 landed) so this is `CONFLICTING` and CI will not schedule a build; **the rebase belongs to the merge bot from here.** This branch has now been rebased and regenerated twice by hand, and each cycle costs a full regen plus a build while main moves every ~20 minutes — that is the treadmill the merge bot exists to absorb.

Resting on the local gate set, as with #11174, #11179, #11182 and #11188 today: `lake build` PASS (4798 jobs), `check-asm-to-program` CLEAN (386 defs / 147 files), `check-region-map` PASS, pins regenerated to **fixpoint** (`gen-symbol-addresses.py --build` twice), ELF measured `.text 0x064378`, `.data 0x5370`, `.bss 0x1b6c2ca0`, `GuestImage .bss` end `0xbe7d2ca0`. Only generated files committed.

**Behavioural evidence, measured on `7d9824dec` with both legs sharing that base:** 200/200 ran, 0 errors, **FA 0, FR 5 on both**, identical FR label set (`00033`, `00085`, `00092`, `00120`, `00185`), and the **23 previously-introduced regressions all gone**.

⚠️ **Measurement base `7d9824dec`, merge base `92cb99e53` — deliberately different, and stated in the body.** Attribution needs both legs on one base, which it has; it does not need that base to be current. The merges in between are layout-only, so re-running the legs would burn a cycle to reproduce the same numbers. That distinction cost this branch an hour earlier when I conflated attribution with mergeability — CI builds the *merge ref*, which is why the rebase was needed at all, but the A/B did not need repeating.

## Why this lands

**Root cause: terminal root publication sat where zero-execution blocks never reach it — wrong-stage, not wrong-value.** Direct evidence on row `00073`: parent `sv_recomputed` **hits=8** with first limb write at `0x800133ac`; candidate **hits=0, final=0**. Parent versus candidate on the same row, since a bare candidate zero proves nothing where something does write that address.

**Honest denominator: 23 of 23, not 23 of 200.** `canonical_map=166`, `no_canonical_map=34`, and the 11 remaining no-map rows are **expected rejects** that need no root — a valid control group, not evidence of another publish site. Every accept-expecting row off the canonical map path failed.

**Placement, not reversion.** The root remains **final-state derived, not BAL-derived** — the publish moved to a path every block reaches, which is what `fork.py:352-360` requires, since the spec computes the root from the final `BlockDiff` with no branch on transaction count. Anyone reading FR 29 → 5 will reasonably suspect the structural gain was traded back for the number; it was not.

The diagnosis route is worth keeping: a same-ELF rerun was **rejected as a reverse leg** (identical by construction), the parent leg made the regression real, the **map census** showed all 23 never execute the new path — which eliminated the new mechanism from suspicion and killed a cross-lane hypothesis I had floated — and only then did the placement probe settle it.

Issue #11189, opened 2026-08-02
storage-write undo journal is fail-open on overflow: a reverted write stays applied


pirapira:
**What this is about.** The storage-write undo journal silently stops recording once it is full, and the code path that skips the push **also skips incrementing the count** — so nothing downstream can tell that anything was lost. Past 32,768 entries a write's previous value is never recorded, and `write_sets_restore_frame` then fails to restore it. ⇒ **A reverted write stays applied.** The guest continues and produces a verdict from state the spec would have rolled back.

Found by measurement on the emitted guest, 2026-08-02, while deriving buffer bounds for #11187.

## The mechanism

- `storage_writes_undo_push` (`.s` 26286) is called **unconditionally** from `storage_write_record` on both the update and new-entry paths, and from `destroy_storage`.
- A per-frame `storage_writes_undo_checkpoint` is saved at `call_frame_descend:45015` and restored on unwind — so revert today is a **journal-and-restore** mechanism, not revert-by-truncation.
- The overflow guard is `bgeu t1, t2, .Lswup_done`: on a full journal it **skips the push and returns without bumping the count**.

⛔ **Fail-open on the rollback path.** There is no marker, no counter, no rejection — the omission is unobservable to every later reader. Compare #11157, where three capacity guards continued past their limit; this is the same class, but on rollback, where the consequence is a state divergence rather than a missing record.

⚠️ **This is a code property, read from emitted assembly. It has not been witnessed on a fixture** — the corpus does not approach 32,768 undo entries, which is exactly why it has survived.

## Capacity: the journal is the worst-bounded buffer in the guest

Measured from the emitted code rather than a region table: `li t2, 32768` capacity, stride `slli 7 + slli 5` = 160 bytes ⇒ **32,768 × 160 = 5.0 MiB exactly**. Checkpoint array is 1025 × 8.

The journal pushes on **every** write — no value-change gate, no first-write-per-checkpoint dedup — while the spec charges the expensive part **once per transaction**: `storage.py:112-113` charges `STORAGE_WRITE` only when `original_value == current_value` and `current != new`, i.e. the first change this transaction. Subsequent SSTOREs to the same slot cost only `WARM_ACCESS` = 100.

⇒ Under the #11186 sizing policy:

```
200,000,000 / 100 (WARM_ACCESS)  = 2,000,000 pushes
2,000,000 × 160 B                = 305.2 MiB
```

[...]

PR #11196, opened 2026-08-02
docs(regionmap): correct the undo-journal overflow claim and the live-anchor count

pirapira:
Documentation only, in the file that determines the RAM layout. **No code, no constants, no pins.**

`RegionMap.lean:93` states that the scheme/evidence tag is *"documentation, not used by the proofs"*, and the only theorems over these lists (`allFitZones`, `allPairwiseDisjoint`) depend on `base`/`size`/`mode`/`zone` alone. `lake build EvmAsm.Codegen.RegionMap` is clean.

## 1. ⛔ The undo journal's overflow claim was FALSE — capacity versus flow

The note read: *"Bounded by the SSTORE handler's own 16384-row cap, so it needs no overflow path."*

**The 16,384 cap bounds LIVE ROWS in the write map, not the NUMBER OF WRITES.** `destroy_storage` *decrements* `tx_storage_writes_count` as it drops rows (`.Lds_drop`), so each destruction frees capacity that the next write refills — a slot can be written → destroyed → written → destroyed, pushing an undo every time, while live rows never exceed 16,384.

⇒ **W is a high-water mark on occupancy; the journal is fed by a flow.** The file's `W + D ≤ 2W` derivation (also in `StorageWriteMap.lean`) does not hold.

* **Measured** — fixture `00192` reaches `storage_writes_undo_count` = **32,768 exactly**, the capacity (codex, GH #11189). The guard fires on a **non-adversarial** block today.
* **Derived** — per-transaction regular gas is capped at `TX_MAX_GAS_LIMIT = 16,777,216` (`transactions.py:63`, `fork.py:1098-1100`) and the cheapest journal-pushing write costs `WARM_ACCESS = 100` ⇒ **167,772 reachable = 5.1× the region**. 25.6 MiB at the current 160 B stride, 6.4 MiB at 40 B.

⚠️ It also now records that the overflow is **silent** — the push is skipped *and* the count is not bumped — and that the `unappend` arm decrements `tx_storage_writes_count` per kind-1 entry, so **past the cap the count is already inconsistent with the contents**. ⇒ the guard must reject **at the push**; making the omission observable afterwards is not sufficient.

## 2. The row audit, recorded next to the region

16 of the 160 bytes (offsets 16..31) are **never written and never read**. Kinds 0/1 need **40 B** (32 B value + packed index/kind, 8-aligned — 34 B is not a legal stride since LD/SD require 8-alignment in the verified RV64 subset). Only kind 2 needs the 128 B payload, and it is bounded by distinct written slots, so **segregating by kind gives a 40 B hot array**.

## 3. ⚠️ "The ONLY scheme-A anchor" was stale by twelve

`state_tracker_area` claimed to be *"the ONLY scheme-A anchor the current stateless_guest references"*. **Measured: 13 of the 23 anchors have their base constant present in the emitted guest** — that one, the six r59nm read containers, and the six write/undo containers:

`state_tracker_area` · `storage_reads_area` · `account_reads_area` · `code_reads_area` · `tx_storage_reads_area` · `tx_account_reads_area` · `tx_code_reads_area` · `storage_writes_area` · `tx_storage_writes_area` · `storage_writes_undo_area` · `account_writes_area` · `tx_account_writes_area` · `account_writes_undo_area`

⇒ ⛔ **and those 13 are reached by absolute `li`, so they are NOT covered by `guestRegionMap`'s zone/disjointness theorems** — only by `schemeAAnchors_pairwise_disjoint`, which the file itself labels aspirational and "NOT part of `guestRegionMap`". Worth knowing before anyone treats `guestRegionMap` as covering the emitted reality.

## Why this belongs here rather than in an issue

The wrong claim is what made me withdraw a correct finding: I filed the fail-open defect from measurement, then retracted it on the strength of this rationale, and codex's fixture measurement overturned the retraction. ⇒ **a stale rationale in the authoritative file cost a true finding an hour of its life**, and the next reader would have hit the same thing.

⚠️ **Not done here:** `account_writes_undo_area` carries the same *"same rationale as storage_writes_undo_area"* comment. I have **not** measured whether its bound has the same capacity-versus-flow flaw, and I have deliberately not amended it on the assumption that it does.

### DIFF
--- a/EvmAsm/Codegen/RegionMap.lean
+++ b/EvmAsm/Codegen/RegionMap.lean
@@ -159,8 +159,13 @@ def schemeAAnchors : List GuestRegion :=
       evidence := "MemoryLayout CODE_DB_BUCKETS; 1 MiB slab" },
     { name := "state_tracker_area",     base := 0xa0630000, size := 0x400000,  mode := .rw, zone := .ram,
       evidence := "MemoryLayout STATE_TRACKER_AREA; 4 MiB slab. LIVE in emitted guest: "
-        ++ "storage-log base 0xa0630000..0xa0830000 (2 MiB, 16384x128 rows) — the ONLY "
-        ++ "scheme-A anchor the current stateless_guest references (see FINDING in docs)" },
+        ++ "storage-log base 0xa0630000..0xa0830000 (2 MiB, 16384x128 rows). "
+        ++ "⚠️ NOT the only live one: measured 2026-08-02, 13 of these 23 anchors have their "
+        ++ "base constant present in the emitted guest — this one plus the six r59nm read "
+        ++ "containers and the six write/undo containers below. The earlier ONLY claim was "
+        ++ "stale by twelve. Those 13 are addressed by absolute `li` and are therefore NOT "
+        ++ "covered by guestRegionMap's zone/disjointness theorems, only by "
+        ++ "schemeAAnchors_pairwise_disjoint" },
     { name := "evm_frame_stack",        base := 0xa0a30000, size := 0x40000,   mode := .rw, zone := .ram,
       evidence := "MemoryLayout EVM_FRAME_STACK; 256 KiB slab" },
     { name := "evm_value_stack",        base := 0xa0a70000, size := 0x100000,  mode := .rw, zone := .ram,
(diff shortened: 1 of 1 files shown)


pirapira:
**Reviewed — this is the right change in the right file, and it corrects the file's own error.** Holding the label until `build` reports green; documentation-only or not, `RegionMap.lean` is kernel-checked and the build is what says so.

## The note being replaced was the cause of the defect

The old text read: *"Bounded by the SSTORE handler's own 16384-row cap, so it needs no overflow path."* ⇒ **That is why `storage_writes_undo_push` has no reject.** The same false claim appears in `StorageWriteMap.lean`, and it is the claim that caused a retraction-then-un-retraction of #11189 earlier today: an analyst read the documented rationale, believed it, and withdrew a correct finding.

⇒ **A wrong comment in an authoritative file is not a documentation defect — it is the defect's root cause.** Anyone who reasoned about this buffer reached the same wrong conclusion, because the file told them to.

**And the correction names the error precisely: capacity versus flow.** The 16,384 cap bounds **live rows** in the write map, not the **number of writes**. `destroy_storage` *decrements* `tx_storage_writes_count` as it drops rows (`.Lds_drop`), so each destruction frees capacity the next write refills — a slot can be written, destroyed, written, destroyed, pushing one undo entry every time while live rows never exceed 16,384. That is exactly how a 16,384-row cap fails to bound a 32,768-entry journal, and it explains the measured **exactly-2×** ratio that first made the result look like an instrument artefact.

## Everything load-bearing is here and each claim carries its provenance

- **Measured**: fixture `00192` reaches `storage_writes_undo_count = 32768` — the guard fires on a **non-adversarial block today**, not a constructed one.
- **Derived**: per-tx regular gas capped at `TX_MAX_GAS_LIMIT = 16,777,216` (`transactions.py:63`, `fork.py:1098-1100`) ÷ `WARM_ACCESS` 100 ⇒ 167,772 reachable, **5.1× this region**.
- **Mechanism of the fail-open**: `bgeu t1, t2, .Lswup_done` skips the push *and* the count, and the unappend arm decrements per kind-1 entry ⇒ past the cap the count is inconsistent with the contents. ⇒ **The guard must reject at the push; making the omission observable afterwards is not sufficient.**
- **Row audit**: 16 B of 160 (offsets 16..31) never written and never read; kinds 0/1 need 40 B; only kind 2 needs the 128 B payload and is bounded by distinct written slots.

⚠️ **And a second finding arrived with it that deserves its own issue.** The `state_tracker_area` note previously claimed it was the *only* scheme-A anchor referenced by the emitted guest. Measured: **13 of 23 anchors have their base constant present in the emitted image** — the claim was stale by twelve. Those 13 are addressed by absolute `li`, so they are **not covered by `guestRegionMap`'s zone/disjointness theorems**, only by `schemeAAnchors_pairwise_disjoint`. That is a proof-coverage gap in a file whose stated purpose is machine-checked layout, and it should not stay buried in an evidence string.

Two comment corrections in one file, both of which had propagated into people's reasoning. This is the strongest argument I have seen today for writing measurements where the next reader will hit them.

pirapira:
**Labelled — `build` SUCCESS, and the byte-neutrality claim is now measured rather than argued.**

The author's first neutrality check **compared an ELF to itself and falsely passed**, because codegen could not run at the time. It caught that, refused to claim the result, and fell back to a weaker argument (a grep for value-bearing lines) **honestly labelled as weaker**. Once codegen ran, it did the isolating experiment properly: emit with the doc change, revert *only* `RegionMap.lean` to main's version, rebuild that module, re-emit. Both ELFs `c7b08448318f134efd10da50` — **byte-identical**.

⇒ That is the third appearance of the same-ELF trap today, and the second time it was caught by the person who would have benefited from not catching it. Worth stating because a false pass here would have been invisible: a comment-only PR that *claims* emission-neutrality is exactly the change nobody re-checks.

## What this fixes is the cause of a real defect

The replaced note read *"bounded by the SSTORE handler's own 16384-row cap, so it needs no overflow path."* ⇒ **That is why `storage_writes_undo_push` has no reject.** The same claim sits in `StorageWriteMap.lean`, and it is what caused a correct finding on #11189 to be retracted and then un-retracted earlier today — an analyst read the documented rationale, believed it, and withdrew.

**A wrong comment in the file whose stated purpose is machine-checked layout is not a documentation defect. It is the defect's root cause**, and everyone who reasoned about that buffer reached the same wrong conclusion because the file told them to.

The correction names the error precisely — **capacity versus flow**: the cap bounds *live rows*, not *writes*, and `destroy_storage` decrements the count as it drops rows, so each destruction frees capacity the next write refills.

## Second correction in the same file

`state_tracker_area` claimed to be the **only** scheme-A anchor referenced by the emitted guest. Measured: **13 of 23** have their base constant in the image — stale by twelve. Those 13 are reached by absolute `li`, so they are **not covered by `guestRegionMap`'s zone/disjointness theorems**, only by `schemeAAnchors_pairwise_disjoint`, which the file itself calls aspirational.

⚠️ **Anyone treating `guestRegionMap` as covering emitted reality is wrong by 13 regions.** That is a proof-coverage gap, and it is now stated where a reader will hit it rather than inferred from an evidence string. It deserves its own issue.

Comment-only, zero value-bearing lines changed (`Nat :=`, `base :=`, `size :=`, `mode :=`, `zone :=` all absent from the diff), so no emission input moved.

PR #11197, opened 2026-08-02
docs: record Lake artifact-cache eviction incident

pirapira:
Records the cron failure visibility lesson and the rationale for the 120/80 eviction policy above the measured live hard-linked floor. This PR is documentation-only.

### DIFF
--- a/docs/agents/lake-artifact-cache-eviction.md
+++ b/docs/agents/lake-artifact-cache-eviction.md
@@ -1,0 +1,27 @@
+# Lake artifact-cache eviction incident
+
+On 2026-08-02, the hourly Lake artifact-cache eviction job had been failing
+for weeks while appearing to be installed correctly. The crontab entry was
+present, the script was executable, and the cache path was valid. The failure
+happened before argument parsing because the script tried to locate `lake`
+through `PATH` while running under cron; cron did not include the interactive
+Lake installation. The only evidence was one line in a log that nobody read,
+once an hour, for weeks.
+
+This is the durable lesson: an installed scheduled job needs an observable
+health signal, and its failure path must not depend on interactive environment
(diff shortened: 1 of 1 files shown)


pirapira:
Labelled — doc-only, single new file, no code path touched, and it keeps the right things.

**It leads with why the failure was invisible rather than with the sizes**, which is the part that survives: crontab present, script executable, cache path valid — and the only evidence was one log line nobody read, once an hour, for weeks. ⇒ *An installed scheduled job needs an observable health signal, and a periodic log line is not monitoring if nobody checks it.*

**The tuning guidance is the other durable half.** `--cap-gb 120 --target-gb 80` is chosen **above a measured ~45 GiB live floor** — the 24,173 artifacts with `nlink > 1` that are hard-linked into live `.lake/build` trees and cannot be evicted. ⚠️ A cap **below** that floor makes every hourly run report over-cap forever, converting the one useful signal into permanent noise. That is the same failure shape as a guard that never fires: #11157 and #11189 were both guards nobody had ever seen fire, and this would have been a guard that fires always. Neither is readable.

The sizes will go stale within a month. **The existence of a shared live hard-linked floor, and the reason the cap must stay above it, will not** — and the doc says so explicitly rather than presenting the numbers as the finding.

⚠️ One thing deliberately *not* here, correctly: this PR does **not** patch `scripts/lake-artifact-cache-lru.sh`. That file is being deleted by #11195, and an earlier version of this work modified it — which would have collided with the deletion, on a branch that also carried an unrelated structural fix. Doc-only is the right scope.

For the record: the repo copy is going away because the live copies are installed per-account under `~/.local/bin` and driven by each account's crontab. Keeping a second source of truth in-tree is what let a fixed copy and a broken copy coexist.

Issue #11198, opened 2026-08-02
account_writes_undo_push has NO capacity guard: past 16,384 entries it writes outside its 2 MiB region (worse than #11189)


pirapira:
## Claim

**`account_writes_undo_push` has NO capacity guard at all.** It computes `base + count × 128` and stores unconditionally, then increments the count. Once `account_writes_undo_count` passes 16,384 it **writes past the end of its 2 MiB region**.

⛔ **This is strictly worse than GH #11189.** The storage journal at least *has* a bounds check (`bgeu t1, t2, .Lswup_done`) — it fails **open** but it stops writing. The account journal has no check, so it does not fail open: **it writes outside its region.**

Found by measurement while checking whether `account_writes_undo_area` shares #11189's capacity-versus-flow flaw. It does — **and it lacks the guard as well.**

Base **`9ac6fbd61`** (`main`). Guest asm re-emitted from that tree.

## 1. No guard — the full routine is 18 lines

```asm
account_writes_undo_push:
  addi sp, sp, -64
  sd t0..t6                                    # save temporaries
  la t0, account_writes_undo_count; ld t1, 0(t0)
  li t2, 0xa2d20000                            # region base
  slli t3, t1, 7; add t3, t2, t3                # base + count*128  <-- NO BOUNDS CHECK
  sd a5, 0(t3)
  sd a6, 8(t3)
  bnez a6, .Lawu_appended
  ... 12 dwords copied from the map row ...
.Lawu_appended:
  addi t1, t1, 1; la t0, account_writes_undo_count; sd t1, 0(t0)
```

Measured guard counts, side by side over each routine's own extent: **`storage_writes_undo_push` = 1 `bgeu`; `account_writes_undo_push` = 0.**

`account_writes_undo_area` is `0xa2d20000 .. 0xa2f20000` (2 MiB, 16,384 × 128). Entry 16,384 lands exactly at the region end and every entry after that is further past it.

## 2. And the overflow is reachable — the same capacity-versus-flow error as #11189

`account_write_record` calls the push from **two** sites:

| site | label | adds a map row? |
|---|---|---|
| `.s 26099` | `.Lawr_hit` | ⛔ **no — this is the UPDATE path** |
| `.s 26103` | `.Lawr_append` | yes |

⇒ **an update pushes an undo entry without adding a row.** So repeated writes to the *same* account push N entries while live rows stay at 1, and the journal's inflow is **not** bounded by the map's 16,384 capacity.

[...]

PR #11200, opened 2026-08-02
Fail closed when storage undo journal overflows

pirapira:
## Summary

The storage undo journal previously treated a full journal as a no-op: `storage_writes_undo_push` branched to its success tail when `storage_writes_undo_count == 32768`, so the write then mutated the map without a rollback record. A real corpus witness, `00192_test_return50000_fork_Amsterdam-blockchain_test_from_state_test--g1__b0`, performs 49,999 kind-0 overwrite/hit pushes and exceeds the 32,768-entry journal.

This change makes the push return `a0 = 1` and latch both `tx_storage_writes_overflow` and `storage_writes_overflow`. `storage_write_record` checks that result before mutating the transaction map; `destroy_storage` checks the capacity before read/swap/count mutation and also checks the helper result. Existing verdict overflow handling then rejects with the fixed-arena overflow code. The journal remains the existing 32,768 x 160-byte layout; no RAM resize is included.

The 32,768 bound was incorrectly inferred from the 16,384 live-row/exec-log bound. A kind-0 overwrite of an existing slot still pushes an undo record without adding a map row, so repeated writes can exceed 32,768 while the live-row count stays flat. The temporary diagnostic on 00192 counted 49,999 kind-0 pushes, 2 kind-1 appends, and 0 kind-2 destroy_storage pushes; this is an undersized journal capacity, not a destroy_storage or revert asymmetry. The guard is the safe fix. A larger cap remains a separate fallback for genuinely changing writes; a targeted cap-relaxed-plus-value-identity-skip probe on this exact witness reached `succ=1`, `fail=0`, `rebuilt=387`, `supplied=387`, confirming the no-op records are non-load-bearing. The skip itself is separate follow-up work, and the `fail_code=60`, `rebuilt=135`, `supplied=387` cap-relaxed-only divergence remains separately diagnosed.

## Execution-spec anchors (submodule `e5a8caf1b`)

- `execution-specs/src/ethereum/forks/frontier/state_tracker.py:395-420` unconditionally assigns `set_storage` into the transaction storage-write map; the guest preserves that map update on successful SSTORE; the fail-closed undo guard rejects before mutation when its safety journal is exhausted. Frame rollback is the separate `copy_tx_state`/`restore_tx_state` contract at `:650-700`.
- `execution-specs/src/ethereum/forks/amsterdam/block_access_lists.py:667-676` filters net-zero storage writes (`pre_value != post_value`) before `add_storage_write`; this is why the measured value-unchanged undo traffic is a deduplication follow-up, not a BAL entry requirement. The targeted cap-relaxed-plus-skip probe on the identified witness preserved exact `387/387` BAL output.

## Account-write undo guard (#11198)

This PR also folds in the same fail-closed fix for `account_writes_undo_push`, which previously had no capacity check before storing at `0xa2d20000 + count * 128`. The helper now bounds the count at `txAccountWritesCapacity = 16384` before any out-of-range store, returns `a0 = 1` on exhaustion, and latches both `tx_account_writes_overflow` and `account_writes_overflow`; callers check the status before mutating the account map. The emitted assembly and source guards pin both undo pushes and their caller checks.

This is a structural memory-safety correction; no account-journal overflow witness has been derived or claimed. The declared account-undo region ends at `0xa2f20000`. A hypothetical 32768-entry stream reaches `0xa3120000`: after the `0xe0000` unmapped gap, the first live region is `.data` at `0xa3000000`, and the tail reaches `.bss` at `0xa3110000`. Thus the old unconditional store could corrupt live data rather than merely losing an undo record.

## Evidence

- Base: `9298b78fb695315d7c5c99bd58bd3dd27a0b2e67`
- Storage-only parent ELF: `50fbd45b649c92a41da6aa3e899701d4695680885d47e05867fbc1b8e0321ee1`
- Storage-only candidate ELF: `00c0da95cc021cafdee118095843da5df69882f8b5400648bd792fdb153df451`
- Final candidate ELF (storage + account guard): `d2b682e2ac791de6309ee6a29b2d267d567ffe613f2100f3fbe79fb0741cffec`
- Spike runner: `scripts/spike/spike_run`, sha256 `5b383a7702fb05dea16a2cc37dc313b82491fdf194f39c12306cb387a5a2b6b0`
- `00192` on the final candidate: `succ=0`, `fail_code=58`, `shadow_status=1`, `rebuilt=135`, `supplied=387`; the storage counter watch reached exactly `32768` at the `storage_writes_undo_push` increment.
- Witness identity (not selector index): manifest case_id `3ea3e6019a9d4adf4ffed4afdbddf361300efc662f50f028ccbad2e973bff9d4`, packed `.input` SHA-256 `9fe0961ad262bb4d15430a19e390fe35422cc11902744c1417b04c32f42c0575`, unpacked stateless-input SHA-256 `6fae4d024abb9befb52f7d7686ca0290519e2b68cd1b5e9fe64ee41a7e0a67e8`; relpath `blockchain_tests/for_amsterdam/ported_static/stQuadraticComplexityTest/return50000/return50000.json`.
- Temporary push-site diagnostic on the same identified 00192 witness: kind-0 overwrite/hit pushes = 49,999, kind-1 append pushes = 2, kind-2 destroy_storage pushes = 0. The overflow is therefore a repeated kind-0 path, not a destroy_storage kind-2 surplus; a cap-relaxed-plus-value-identity-skip diagnostic reached `succ=1`, `fail=0`, `rebuilt=387`, `supplied=387` on the same input. The counters and skip were diagnostic-only and were removed before the final ELF.
- Paired random-200 manifest for the storage-only change: `/tmp/codex2-eest-candidate-forward/manifest.tsv`; parent `FA=0 FR=6 OK=194`, storage-only candidate `FA=0 FR=7 OK=193`, one forward transition (`00192`), zero reverse transitions. The account guard was added after this A/B and has no account-overflow witness.
- Final-candidate safety sample: first 30 rows of the same manifest, `FA=0`, `FR=5`, `OK=25`; compared with the captured parent rows, zero forward and zero reverse transitions. No account-journal witness was inferred from this sample.
- Force-fire cap-zero control (from the same candidate family): four rows rejected with `FA=0`.
- Full non-cache `lake build`: 4798/4798 jobs.
- `scripts/check-asm-to-program.sh`: clean, 386 converted definitions.
- `scripts/check-region-map.sh NO_BUILD=1 ELF_DIR=gen-out/regionmap`: clean.
- Final layout: `.text=0x63d5c`, `.data=0x5370`, `.bss=0x1b6c2be0`; `.data` and `.bss` bases remain pinned.


pirapira:
**Labelled.** `CONFLICTING` against `aada0055e`, so CI will not schedule a build; resting on the local gate set as with #11174, #11179, #11182 and #11188 today — full non-cache `lake build` 4798/4798, `check-asm-to-program` clean (386 defs), `check-region-map` clean.

**Two guards, and they are different defects.**

| routine | before | after |
|---|---|---|
| `storage_writes_undo_push` | one `bgeu` → **skips the push and the count** (#11189) | fail-closed at the push |
| `account_writes_undo_push` | ⛔ **no guard at all** — unconditional store (#11198) | fail-closed at the push |

⇒ The account one is the more serious. With no bounds check it computes `0xa2d20000 + N×128` and stores: entries 16,384–23,551 land in an unmapped gap, **23,552+ overwrite `.data`**, **32,256+ overwrite `.bss`**. That is memory corruption producing a *wrong verdict*, not a crash — and the 7,168-entry runway through the gap means nothing notices first.

**Why the guard must reject at the push, not later:** `write_sets_restore_frame`'s unappend arm decrements `tx_storage_writes_count` per kind-1 entry, so once a push is skipped the count is already inconsistent with the contents. Making the omission observable afterwards is not sufficient. Both guards also latch tx- and block-level overflow so callers that ignore the return still reject, and **three `#guard` lines pin the guards into the emitted text** so they cannot silently regress.

## The `code60` divergence is explained, and it strengthens this PR

The cap-relaxed trial produced rebuilt 135 against supplied 387, which I initially treated as an unexplained side effect and held the resize on. ⇒ **The drop is positional, not selective**: everything past 32,768 is lost, **real records included**. So the cap-relaxed run *restores* real records and the BAL moves — exactly the direction observed.

⇒ **The saturation was silently corrupting the rebuilt BAL.** This guard is not merely closing a fail-open on rollback; it is closing a path that dropped real undo records and produced a wrong BAL. That is a stronger justification than "prevents a lost entry".

## Evidence, and the identity that nearly cost an evening

Push breakdown on the witness: **kind0 = 49,999 (exact, no diagnostic saturation), kind1 = 2, kind2 = 0** ⇒ not `destroy_storage`, not the `h_REVERT` asymmetry, but the **kind-0 hit path** — repeated writes to slots already in the map, each pushing an undo entry while adding no map row.

⚠️ The witness is cited by **`case_id 3ea3e601…` and input SHA**, never by index — and that is why a contradiction resolved in minutes. Two agents measured the same block as "00192" and "18635"; `return50000` has **four rows differing only in a gas parameter the 96-character truncation hides**, giving 28, 27, and 49,999 pushes. **A name is not an identity, an index is not an identity, and a relpath is not an identity either.**

## What this costs, stated plainly

The witness row flips **OK → FR** (FA 0, FR 6 → 7, zero reverse). The guard makes overflow **sound, not correct**. ⇒ The follow-up is the **value-unchanged skip**, not a resize: a rewrite that changes nothing pushes an undo record restoring the value it already has, and the spec emits `storage_changes` as a function of the final map alone (`block_access_lists.py:667-676` — no sequence, no journal, no arrival count), with the journal having exactly one reader. Skipping those collapses 49,999 records to near zero and removes the pressure entirely. **Resize is now fallback of last resort.**

PR #11203, opened 2026-08-02
docs(regionmap): name the actual undo-journal flow — value-unchanged rewrites, not destroy_storage

pirapira:
## The note got the verdict right and the mechanism wrong, and the wrong mechanism was steering a 20.6 MiB resize

GH #11189. #11196 replaced the undo journal's *"needs no overflow path"* claim with a correct verdict — the region's overflow is fail-open and the 16,384 cap does not bound it — but attributed the unbounded flow to **`destroy_storage`** freeing map rows that later writes refill.

⛔ **That is refuted by measurement.** The kind breakdown on the witness is **`kind0 = 49,999, kind1 = 2, kind2 = 0`** — `destroy_storage` pushed **nothing at all**. Kind 2 is its only push site.

### The actual source, and it escapes a different cap

The escaping traffic is the **value-unchanged rewrite**, and it escapes the **exec-log** cap rather than the write-map cap:

* **`Storage.lean:718-729`** skips the exec-log append when the found row's `current` equals the value being written. Its own comment gives the motive: *"Skipping the append keeps long loops that rewrite the same value (e.g. a success flag per CALL iteration) from exhausting the 16384-entry exec-log arena and halting on the capacity guard."*
* **`Storage.lean:796-800`** — *"Both the append and the value-unchanged skip converge here, **and the recorder is called on BOTH**."*

⇒ such a rewrite pushes an undo record while `log_length` stands still, so the capacity guard at `Storage.lean:667-669` — which tests `log_length` — never fires.

⚠️ **And it is narrower than "repeat writes to a known slot."** The exec log is **append-only** and never updates in place (`:730-784` writes at `log_base + log_length × 128` and increments; `h_SLOAD` reverse-scans for the *last* match *because* supersession is by appending). A repeat write with a **different** value **does** append and **does** advance the guard. Only the identical-value case escapes.

### Four further corrections in the same note

1. ⛔ **A reading of exactly 32,768 is the cap reading, not a demand.** `bgeu t1, t2, .Lswup_done` skips the push **and** the increment, so the counter saturates and carries no upper information. The 49,999 is from a run with the cap raised.
2. ⚠️ **Cite the witness by `case_id` and input sha, never by index.** Two of us measured "fixture 00192" and reported **49,999 against 16** — because an index is selector-relative and **both counters were correct**. The relpath is not an identity either: four corpus rows share it and differ only in a gas parameter the 96-character label truncation hides; the `g0` variant and the sibling test push 28 and 27.
3. **167,550, not 167,772.** `storage.py:97` gates every SSTORE through `check_gas(max(gas_cost, CALL_STIPEND + 1))`, so each write **requires `gas_left ≥ 2301` while spending only 100** and the last 2,301 gas cannot fund a write. The 222-row difference decides no buffer size, but the naive `TX_MAX_GAS_LIMIT / WARM_ACCESS` is what a re-derivation lands on twice.
4. ⭐ **That ceiling is a leak, not a requirement — DO NOT RESIZE TO IT.** A value-unchanged rewrite's undo record has `prevValue` equal to the value being written, so replaying it stores identical bytes: an identity on the map, hence on the map length and on the rebuilt BAL, which `block_access_lists.py:667-676` derives from the **final map alone** (it iterates `tx_state.storage_writes` and compares against `_get_pre_tx_storage` — no sequence, no journal, no arrival count). And the rows have exactly **one** reader: of the seven occurrences of `0xa23a0000` in `EvmAsm/`, `StorageWriteMap.lean:423` is the push and `:480` is `write_sets_restore_frame`; the rest are layout or docstrings. ⇒ skipping the identity push makes **32,768 tight again, at 20.6 MiB less than the resize**.

⛔ **And the skip must test `entry+64`, not the exec-log row's `current`.** `h_SSTORE`'s value-unchanged test uses the exec log, and **the exec log retains reverted writes while this journal rolls the map back** — post-revert the two disagree, and gating on the log would drop a **real** undo record in the false-accept direction.

### Scope and emission

Comments and one `evidence` string. **Byte-neutral by construction rather than by coincidence:** `evidence` has **zero** references outside `RegionMap.lean`, `CallFramePhase` reads only `.size` fields (unchanged at `0x500000`), and the emitted `.s` contains none of the evidence text (`grep` = 0). ⚠️ **I did not run a two-leg ELF diff** — I built the guest on this branch (`lake exe codegen --program stateless_guest --halt linux93`, clean) but did not rebuild main to compare, because a same-`-o` two-leg diff would need a second full tree and the construction argument covers it. Saying so rather than implying a measurement I did not take.

`lake build EvmAsm.Codegen.RegionMap` clean; `lake build codegen` clean (4294 jobs).

### Not in scope

The `storage_write_record` hit-path skip itself, and #11200's fail-closed guard. They are **complements**: the skip removes the false reject, #11200 keeps the residue sound.

### DIFF
--- a/EvmAsm/Codegen/RegionMap.lean
+++ b/EvmAsm/Codegen/RegionMap.lean
@@ -177,9 +177,21 @@ def allPairwiseDisjoint : List GuestRegion → Bool
     because the 100-gas divisor ignored the 16,384-row fail-closed source cap.
 
     ⚠️ **And the converse trap, which is how the journal note went wrong:** a
-    capacity cap bounds OCCUPANCY, not FLOW. `destroy_storage` decrements
-    `tx_storage_writes_count`, so freed rows are refilled and the journal's inflow
-    is not bounded by the map's size. **Do not bound a flow with a capacity.**
+    capacity cap bounds OCCUPANCY, not FLOW. **Do not bound a flow with a capacity.**
+
+    ⛔ **AND THE FIRST ATTEMPT AT THAT CORRECTION NAMED THE WRONG FLOW**, which is why
+    the region note below carries the mechanism and not just the verdict. It blamed
+    `destroy_storage` freeing map rows that later writes refill; the witness's kind
+    breakdown is `kind0 = 49,999, kind1 = 2, kind2 = 0`, so **`destroy_storage` pushed
+    nothing at all.** The real source is the value-unchanged rewrite, which escapes the
+    **exec-log** cap (not the write-map cap) because `Storage.lean:718-729` skips the
+    append while `:796-800` still calls the recorder.
+
+    ⭐ **The generalisation, which cost three numbers in one day:** a cap on a container
+    is not a cap on the traffic through it — **and a cap on the traffic is not a cap on
+    the traffic that matters.** Two of the three journal kinds here are records whose
+    replay is a bitwise identity, so the ceiling they imply is a leak to close rather
+    than a size to reserve.
 
     ⛔ **OVER-RESERVING IS NOT FREE HERE.** The old 76.8 MiB gas-derived reservation
     was unioned into `call_frame_arena`'s front and **physically zeroed** by per-tx
@@ -191,8 +203,12 @@ def allPairwiseDisjoint : List GuestRegion → Bool
     ⚠️ **OPEN, NOT SETTLED — do not read this section as complete.** Whether a
     *legitimate* 200M-gas block can EXCEED either cap is unresolved: the log cap is
     fail-closed so exceeding it is sound but false-rejects, while the journal cap is
-    fail-open so exceeding it is a state divergence. Journal side is known to be
-    reachable (fixture `00192` hits 32,768 exactly); the log side's realistic peak is
+    fail-open so exceeding it is a state divergence. The journal side is known to be
+    reachable — 49,999 kind-0 pushes on `case_id`
+    `3ea3e6019a9d4adf4ffed4afdbddf361300efc662f50f028ccbad2e973bff9d4` with the cap
+    raised. ⚠️ **Cite that row by `case_id` or input sha, never by index:** two agents
+    measuring "fixture 00192" reported 49,999 against 16 because an index is
+    selector-relative, and both counters were correct. The log side's realistic peak is
     still being quantified.
 -/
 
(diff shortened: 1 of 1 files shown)


pirapira:
**Reviewed — label on green build. And the condition I attached to this PR has now been discharged empirically.**

I told the author that if its discriminator showed the outputs differing, the do-not-resize paragraph had to come out before merge. ⇒ **It does not need to come out.** An independent skip-only run on the witness (`return50000` g1, `case_id 3ea3e601…`), at the **unchanged 32,768 cap**, gives:

| | pushes | succ | rebuilt / supplied |
|---|---|---|---|
| baseline | 49,999 | reject | 135 / 387 |
| **value-identity skip only** | **18** | **1** | **387 / 387** |

⇒ The dedup collapses the traffic three orders of magnitude below the cap and **preserves the BAL exactly**. The do-not-resize conclusion is now measured, not argued.

## Why correcting an already-merged file was the right call

#11196 landed with the **mechanism wrong, not just the ceiling**. It blamed `destroy_storage` decrementing `tx_storage_writes_count` so freed rows are refilled. ⇒ The push breakdown on the witness is **kind0 = 49,999, kind1 = 2, kind2 = 0** — **zero `destroy_storage` pushes.** That file named a component which contributed nothing.

⚠️ And that matters more here than in most files. **`RegionMap` is what people read to understand these buffers, and a wrong mechanism in it is precisely what caused #11189** — the previous note asserted the journal needed no overflow path, an analyst read it, believed it, and retracted a correct finding. Replacing one confident wrong explanation with another would have been the worst outcome of the day.

## What the corrected text now carries

The real mechanism (a **value-unchanged rewrite** pushes an undo record while appending nothing — `Storage.lean:718-729` skips the append, `:796-800` calls the recorder anyway); that a rewrite with a **different** value does advance the guard; that **32,768 is the cap reading, not a demand**, because the guard skips the increment too; the witness by `case_id` and input SHA **with the four-rows-share-a-relpath warning**; the `167,550` ceiling with the `CALL_STIPEND+1` floor spelled out; and the do-not-resize argument with the one-reader census and `block_access_lists.py:667-676`.

⚠️ **No two-leg ELF diff is claimed**, and the author said so explicitly rather than implying a measurement it did not take — the argument is by construction (evidence strings have zero references outside this file, `CallFramePhase` reads only `.size`, unchanged at `0x500000`, and the emitted `.s` contains none of the text). That is the third time today it has declined to upgrade an argument into a measurement, and one of those refusals caught a false pass on a comparison of an ELF against itself.

pirapira:
**Labelled — `build` SUCCESS, all six checks green, and the condition I attached is discharged in the affirmative.**

I said the do-not-resize paragraph had to come out if the author's discriminator showed the outputs differing. ⇒ **It does not.** Two independent experiments since then confirm it:

- **#11200 alone false-rejects the witness; #11200 + the identity skip accepts and matches the oracle** (legs C/D, re-sited not forced).
- **Random-200 on a different build**: FA=0 both legs, FR 7 → 6, one forward flip (the witness), reverse list empty.

⇒ The dedup collapses 49,999 pushes to **18** at the **unchanged** 32,768 cap. The do-not-resize conclusion this file now states is measured, not argued.

**Why correcting an already-merged file was right:** #11196 landed with the **mechanism wrong**, not just the ceiling — it blamed `destroy_storage` decrementing `tx_storage_writes_count` so freed rows are refilled. The push breakdown on the witness is **kind0 = 49,999, kind1 = 2, kind2 = 0**: **zero `destroy_storage` pushes.** That file named a component which contributed nothing.

⚠️ And `RegionMap` is exactly where a wrong mechanism does damage. Its *previous* note asserted the journal was "bounded by the SSTORE handler's own 16384-row cap, so it needs no overflow path" — which is why `storage_writes_undo_push` had no reject, and why a correct finding on #11189 was retracted and then un-retracted when an analyst read that rationale and believed it. ⇒ **Replacing one confident wrong explanation with another would have been the worst outcome available.**

**What the corrected text carries:** the real mechanism (`Storage.lean:718-729` skips the append for a value-unchanged rewrite, `:796-800` calls the recorder anyway); that a rewrite with a *different* value does advance the guard; that **32,768 is the cap reading, not a demand**, because the guard skips the increment too; the witness by `case_id` and input SHA **with the four-rows-share-a-relpath warning**; and the `167,550` ceiling with the `CALL_STIPEND+1` floor spelled out.

⚠️ **No two-leg ELF diff is claimed**, stated explicitly rather than implied — the neutrality argument is by construction (evidence strings have zero references outside this file, `CallFramePhase` reads only `.size`, unchanged at `0x500000`, and the emitted `.s` contains none of the text). That refusal to upgrade an argument into a measurement is the same instinct that caught a false pass earlier today, when a first neutrality check compared an ELF against itself.

pirapira:
## Merge ordering: land #11206 first if these collide

`RegionMap.lean` is a three-way pile-up right now — main moved it in the #11201 batch, **#11206** moves it, and this PR moves it. So whichever of #11206 and #11203 lands second will conflict.

⇒ **Prefer #11206.** It is a soundness fix (pre-user BAI-0 system-map seeding, with a preload-disabled control that flips reject→accept and a forward flip on an unselected beacon fixture). **This PR is documentation only and entirely deferrable.** If the ordering is awkward, land #11206 and this will be rebuilt on top — a docs PR should not hold a labelled fix.

⚠️ And when either is resolved: the conflicting files are `RegionMap.lean` and `scripts/asm-fixtures/symbol-addresses.tsv`, which are **generated layout facts**. They must be **regenerated from the post-merge tree**, not merged textually and not resolved hunk-by-hunk. Taking either side produces a file that describes **neither** tree — and `RegionMap.textSizeBytes` is an emission *input*, so a wrong value is consumed rather than validated and yields a confident false result under a green build. That failure has already cost two builds today.

The comment content here is unaffected by the rebuild — it describes the undo-journal flow, not layout values, and this PR changes **zero value-bearing lines**.

Issue #11202, opened 2026-08-02
account_writes_discard_tx never called — account/storage fail-path asymmetry


pirapira:
## Summary

`account_writes_discard_tx` is a 40-byte never-referenced text symbol in the linked guest. Its body is identical to the clear half of `account_writes_incorporate_tx` (zero `tx_account_writes_count`, `account_writes_overflow`, `account_writes_undo_count`). The storage twin `write_sets_discard_tx` **is** live. This issue records what that deadness means before the dead routine is deleted.

## Live storage fail path

`write_sets_discard_tx` is jal'd from:

- EOA body reconcile on status=0 (`.Lbv_mtx_storage_drop`)
- MtxRuntime storage status gate
- CreationStage fail paths

Storage: **status-gated** discard vs incorporate.

## Account path (no discard call)

After body execution, MtxRuntime / EOA reconcile always:

1. `account_writes_emit_builder_tx`
2. `account_writes_incorporate_tx`

There is **no** status gate and **no** `jal account_writes_discard_tx`.

Comments claim body effects were already rolled back via `account_writes_restore_frame` (body snapshot / frame return) before the always-incorporate. Coinbase fee is appended after rollback and intentionally survives either status.

## What deadness means (open question)

Two readings — neither is proven solely by never-ref:

1. **Benign dead twin** — restore-to-mark + always-incorporate (plus intentional coinbase) is the complete fail model; discard would be redundant with a correct restore. Deleting the symbol is pure cleanup.
2. **Missing fail-discard wiring** — storage has an explicit discard; account relies on restore completeness. If any fail path reaches emit/incorporate without a full restore to the tx entry mark, residual account rows leak into the block map / BAL. The dead discard is evidence the storage-style fail path was never wired for account.

This is the same **account/storage twin asymmetry** surface as:

- #11198 `account_writes_undo_push` ABSENT capacity guard (storage undo has CONTINUE-at-cap)
- #11189 / #11200 storage undo CONTINUE vs account undo ABSENT

## Evidence (main `aada0055e` class)

- Emitted guest: `account_writes_discard_tx` def only; 0 jal/la
- `write_sets_discard_tx` multiple jal sites
- Body of discard == incorporate `.Lawi_clear`

## Action

- Delete the never-ref routine in the dead-BSS/text batch PR (cleanup under reading 1).
[...]

PR #11199, opened 2026-08-02
chore(codegen): unlink dead derive_block_system_requests (#11156)

pirapira:
## Summary

Unlink dead text symbol `derive_block_system_requests` from the guest image (#11156).

**KEEP** `deriveBlockSystemRequestsData` / all `dbsr_*` BSS — the live deferred system-request path still writes `dbsr_wbody`/`dbsr_cbody`/… and `requests_hash_verify` still reads them. Only the **combined wrapper routine** is dead; the guest inlined the same sequence.

Probe unit `zisk_derive_block_system_requests` still embeds `deriveBlockSystemRequestsFunction` for standalone probe builds.

## Settle (full discipline, main `9ac6fbd61`)

| check | result |
|---|---|
| defs | `derive_block_system_requests` @ `0x800629cc`, size **280 B** (next = `derive_withdrawal_requests`) |
| refs | **0** non-def lines in emitted `.s` (label only) |
| control that fires | `derive_withdrawal_requests` / `derive_consolidation_requests` jal'd from deferred staging |
| SPIKE_BREAK dead | **0 hits** on withdrawal fixtures (00584, 23106) |
| SPIKE_BREAK control | **HIT** `0x80062ae4` same fixtures |
| fallthrough predecessor | prior global ends in **`ret`** (`.exit_no_epilogue`) — safe |
| dbsr_* data | **PROVEN REACHABLE** (inlined path + requests_hash) — NOT deleted |

Parent guest sha256: `c7b08448318f134efd10da50aad5159691c4ae03c9f2f6a97f89fe7e666adad9`  
Cand guest sha256: `85b0bd67d93601fb787dfe143ebe853f398e29011fceec39417860d117eeffb5`

## Layout fixpoint

- `textSizeBytes` **0x0639e0 → 0x0638c8** (ELF-measured, −280 B)
- data/bss unchanged (`0x5370` / `0x1b6c2be0`)
- GuestAddrs + TSV regen; `check-region-map` GREEN

## A/B (random-200 seed42, `/tmp/11167-random200.tsv`)

| leg | FA | FR | OK |
|---|---:|---:|---:|
| parent | 0 | 5 | 195 |
| cand | 0 | 5 | 195 |

FR sets EQUAL; reverse OK→FR **0**; guest_succ diffs **0**; **OUTPUT 256B byte-identical 200/200**. ELFs DISTINCT.

## Files

- `StatelessGuest.lean` / `BlockVerdictV2.lean`: drop Function from guest aggregates; keep Data
[...]

PR #11204, opened 2026-08-02
chore(codegen): delete dead BSS twins (~1.18 MiB) + account_writes_discard_tx

pirapira:
## Summary

Delete dead guest BSS migration twins (~1.18 MiB) and two never-ref text symbols. Funds RAM headroom for correct buffer sizing (#11186/#11189).

Closes nothing directly; supports #11189 journal architecture headroom. Related: #11202 (account discard asymmetry).

## BSS deleted (PROVEN UNREACHABLE)

| symbol | size | mechanism |
|---|---:|---|
| `rle_control` + `rle_records` | 24 + 609472 | **migration twin** of live `brr_control`/`brr_records` (same capacity). Production materialize uses `brr_*` only (instr refs: brr 5, rle 0). Probe `ReceiptList.lean` keeps local `rle_*`. |
| `sust_out` | 400000 | Probe-only output. Production `system_user_exec_log_slot_tuples` writes via caller `a7` = `atsc_execbuf`. |
| `itr_value_descs` | 152368 | Probe-only. Production uses external descriptor arrays + `itr_paths`/`itr_changes`. Probe `MptIndexedTrieRoot.lean` keeps local. |
| `bvgr_tx_state_refund` arena | 76176 reclaimed | Full `bvMtxU64ArenaBytes` never wired from production text. Shrunk to **8 B** scalar placeholder so `CreateCollision.lean` name still resolves if that branch is linked. |

**Measured reclaim:** `bssSizeBytes` `0x1b6c2ca0 → 0x1b5948a0` (−1 238 016 B ≈ **1.180 MiB**). (24 B of the 1 238 040 arithmetic sum absorbed by alignment.)

### KEEP-list (do not delete)

- **brr_***: `brr_control`, `brr_records` (live receipt materialize)
- **sust_*** production: `sust_sysbuf`, `sust_userbuf`, `sust_endbuf`, counts/txindex (not `sust_out`)
- **atsc_***: `atsc_execbuf` and sibling tuple buffers
- **itr_paths**, **itr_changes**, builder frames
- **other bvgr_***: `bvgr_block_gas_increments`, `bvgr_tx_state_gas`, `bvgr_tx_exec_state_gas`, `bvgr_tx_total_state_gas`, `bvgr_tx_predelegated_auth_count`, receipt/before/applied refund arrays
- Encoder scratch still live: `rle_field_len`, `rle_prefix_len`, `rle_empty_logs`, `rle_zero_bloom`, `re_*`/`rle_payload_buf`

### Neighbour bounds

- `sust_*buf` extent = `bsrMaxTuplesPerSlot * 40`; max index 9999 → offset 399960 within 400000 array (no overrun into neighbour).
- `bvgr_*` u64 arenas = `bvMtxU64ArenaBytes` (9523×8); max index 9522 → offset 76176 within own arrays.
- SPIKE_WATCH hits=0 on deleted regions (settle); controls `brr_control`/`itr_paths` fire.

### Probe still builds

- `zisk_system_user_exec_log_slot_tuples` keeps `sust_out` via `systemUserExecLogSlotTuplesProbeOutData`.
- Receipt / MPT probes keep local `rle_*` / `itr_value_descs`.

## Text deleted

| symbol | note |
|---|---|
| `account_writes_discard_tx` | Never jal'd. Storage twin `write_sets_discard_tx` is live. **#11202** records account/storage fail-path asymmetry (always emit+incorporate after restore vs status-gated discard). Pattern with #11198 account undo ABSENT guard. |
| `bal_serializer_filter_reads` | Never jal'd from guest; removed from `blockAccessListBuilderFunctions`. |

Text batch intentionally **narrow**: broader never-ref text (descriptor outcome builders, simple-transfer bal verify, etc.) still has probe `Program` views anchored on `GuestAddrs.<entry>` — needs #11118-style local Pc placeholders before unlink. Deferred.

## Layout fixpoint

| pin | parent (main aada0055e) | cand |
|---|---|---|
| textSizeBytes | 0x06411c | **0x064050** (−204 B) |
| dataSizeBytes | 0x5370 | 0x5370 |
| bssSizeBytes | 0x1b6c2ca0 | **0x1b5948a0** |
| GuestImage .bss end | 0xbe7d2ca0 | **0xbe6a48a0** |

`check-region-map` GREEN. GuestAddrs + TSV + GIE regenerated (not hand-set).

## A/B (random-200 seed42, `/tmp/11167-random200.tsv`)

|---|---|---:|---:|---:|
| parent main `aada0055e` | `35bee1cc6e802efdef40d2ee4c39e047062667cf2dc9d0dfbc709f67d87db3ad` | 0 | 5 | 195 |
| cand | `599c9b9f00c4cf789243fb5bdb974c12ca28f7282913c45c8a5c211ce844f72a` | 0 | 5 | 195 |

- ELFs **DISTINCT**
- FR sets **EQUAL**
- `guest_succ` diffs **0**
- OUTPUT **256 B byte-identical 200/200**
- reverse flips **0**

## Files

- `BlockVerdictDataSection.lean` — drop rle twin, itr_value_descs; shrink bvgr refund
- `SystemStorageSlotTuples.lean` — sust_out probe-only
- `AccountWriteMap.lean` — drop discard_tx (#11202)
- `BlockAccessListBuilder.lean` — drop filter_reads
- RegionMap, GuestImage, GuestAddrs, symbol-addresses.tsv

## Test plan

- [x] check-region-map
- [x] layout regen fixpoint (stable sha)
- [x] A/B FA0 FR5 identical 256B
- [ ] CI full build


pirapira:
**Labelled.** `CONFLICTING` against current main so CI will not schedule a build; resting on the local gates and the settle evidence, as with five PRs today. The merge bot rebases and regenerates.

**Reclaim verified independently: `0x1b6c2ca0 → 0x1b5948a0` = 1,238,016 B = 1.181 MiB**, against a census sum of 1,238,024 B — **8 bytes apart, which is alignment padding at the new boundary**, exactly as on #11179. The layout moved by what was removed and nothing else.

**The `rle_records` finding is the most transferable part.** 609,472 B superseded by a **live twin of identical size** — `brr_records`/`brr_control` live, `rle_control` also unreferenced. ⇒ That is the signature of a migration that landed the new buffer and left the old one allocated, and it is worth recognising as a shape rather than a one-off: **a same-size twin where one side has zero references is a migration remnant.** The `brr_*` keep-list is in the body so nobody re-derives it.

**Evidence per symbol** carries the discipline this lane has been holding all day: `la` = 0, `SPIKE_WATCH` hits = 0, neighbour-reach bounded with a number rather than asserted, and — the check people skip — the fallthrough predecessor examined for text-shaped symbols. ⚠️ And the `hits=0` asymmetry is respected: it proves absence only where nothing addresses the region, which is why `sps_keys` was **not** deleted earlier on the same evidence when it had four live `la` sites.

**Behaviour: FA 0, FR 5, reverse 0, and 256-byte output byte-identical on 200/200**, parent `35bee1cc` vs candidate `599c9b9f`. For a dead-data deletion that last figure is the right claim — not that the same rows failed, but that **nothing observable changed at all**.

## Two judgement calls I want to endorse explicitly

⛔ **`account_writes_discard_tx` was filed before it was deleted (#11202).** A discard routine that exists and is never called is **evidence of missing wiring**, not merely dead code — deleting it would have erased the symptom while leaving the gap. Filing first, deleting second, is the correct order.

⇒ And it is the **second instance of a pattern**: `account_writes_undo_push` has no capacity guard where storage has one (#11198); `account_writes_discard_tx` is dead where `write_sets_discard_tx` is live. **The account-side twin is systematically worse than its storage sibling.** That predicts where to look next, and enumerating the remaining pairs is cheap.

**The broader text batch was deferred** because of `GuestAddrs` probe anchors. Correct call — a deletion that breaks a probe gets reverted, and #11199 kept its probe working by leaving the function embedded for standalone builds.

pirapira:
## Why deleting `account_writes_discard_tx` is safe (#11202 closed)

#11202 is **closed as not a defect**. Closing note for reviewers / next reader who finds a deleted discard and might reconstruct a “missing fail-discard” theory:

### Live fail path (does not use discard)

1. `status=0` → `dispatcher_restore_body_state` → `account_writes_restore_frame` undoes body-interval writes.
2. **`account_writes_incorporate_tx` always runs** and zeros the tx count after promoting intentional survivors (sender debit pre-capture, coinbase post-restore).

Storage uses status-gated `write_sets_discard_tx` because it has no intentional survivors. Account **cannot** discard-as-zero instead of incorporate — that would wipe fee rows.

### SPIKE

`tx_account_writes_count` on fail sstore 19708: body append → restore → survivor remains → incorp zeros. Multi-tx segments always restart at 0 after incorp. No cross-tx carry of failed body writes.

### Delete rationale

The routine is never jal’d **and would be wrong if called** in place of incorporate. Removing a never-called, incorrect-if-called twin is cleanup, not “closing a hole by deleting the missing wire.” Wiring audit stays closed with #11202; only #11198 remains as the real account/storage guard asymmetry.

Issue #11100, opened 2026-08-01
BAL: missing block-level BAI 0 producer plus wrong-stage end-of-block system calls


pirapira:
**Two defects in how the guest assigns `block_access_index` (BAI) to BAL entries.**

1. **The block-level BAI `0` producer is missing.** The spec uses BAI `0` for block-level system work — beacon-roots and history-storage writes (`fork.py:897`, `:903`). The guest emits no zero-valued store to `current_block_access_index` at all: its data initializer is `1`, and every positive producer is user `i+1` or post-execution `N+1`. (Re-measured 2026-08-02 on `713750d3e`: **five** producers, not the six recorded below — the single-transaction constant-`1` store was retired by #11102, which cut `BlockVerdictFunction.lean` from 1431 to 290 lines. **The phase claim now rests on one mechanism, the initializer, not two — do not go hunting a constant-1 producer.**) There is no state that represents the block-level system boundary.

2. **End-of-block system calls run at the wrong stage.** `BlockVerdictStateRoot.lean:754-764` executes them *before* the user transaction loop, so their writes enter the ordinary transaction-level map too early and are attributed to BAI `1`. The spec's order is user transactions first, then BAI := `N+1`, then post-execution operations (`fork.py:916-926`).

The two interact: because the phase is early *and* no zero state exists, end-of-block system writes land at BAI 1 instead of `N+1`.

⚠️ **Scope guard:** the `N+1` producer **already exists** in the current guest (assembly lines 16464, 18215). This issue must not reimplement it. What is missing is the `0` producer and the correct phase ordering.

---

## Finding, re-measured

Fresh `main` at `7dcdbe35210858f39bd13afde08fc36fad1e6040` was emitted with:

```text
lake exe codegen --program stateless_guest --halt linux93 -o /tmp/codex2-10695/stateless_guest
```

Artifact SHA256: assembly `d48e8eab03c789f94a7b9a816b26ac4b4ee987158760fa386a18397301512af3`,
ELF `7aa10ba1f10303d903d8e50bd383b63d4841aab4107c8041a93fe0220f298fdf`.

The earlier producer census was wrong. The current image has six positive
stores to `current_block_access_index`:

* user `bv_mtx_i + 1` at assembly lines 16211, 16261, and 16698;
* post-execution `bv_tx_count + 1` at 16464 and 18215;
* single-transaction `1` at 17959.

The data initializer is `1`, and there is **no zero-valued store**. Therefore
[...]

PR #11206, opened 2026-08-02
fix(codegen): seed modeled system storage before user txs

### LINKED ISSUE #11194: EIP-2935 history slot 0 is served from the declared BAL, not produced by executing the system call
**What this is about.** The EIP-2935 history contract's storage slot 0 holds the parent block hash. The spec *produces* that value by executing a system call during block processing. **The guest does not derive it — it takes the value from the proposer-declared BAL.** With the post-state root currently BAL-fed, a proposer that declares a wrong history value consistently has nothing checking it against a value the guest computed itself.

Confirmed by two agents independently on 2026-08-02, on fixture `23000_test_block_hashes_history … single_block_check_blockhash_first__b0`.

## The evidence

**The pre-state genuinely has no storage, and that is correct.** Decoding the 12 `witness.state` nodes with execution-specs `WitnessState` against the parent header `stateRoot 3936ea5e…`:

- `HISTORY_STORAGE_ADDRESS` account: `nonce=1`, `balance=0`, **`storage_root = EMPTY_TRIE_ROOT`** (`56e81f17…`)
- `get_storage(slot 0) = 0`; slot hash `keccak(32 zero bytes) = 290decd9…`; **no storage-root node or leaf present**
- fixture JSON `pre` entry has `storage={}`

⇒ The guest's storage lookup reporting **status 5 (slot not in the storage trie)** on 57/57 calls is **the correct answer for a pre-state read.** The value does not exist in pre-state and is not supposed to.

**Where the value actually comes from.** `096fc954…db6524` is exactly the genesis/parent header hash (header number 0), and it appears in the BAL as the storage-change `new_value` at slot 0:

```
f84094[history address] e6e580e3e280a0[096fc954…] c0c0c0c0
```

**No witness node contains those value bytes.**

⇒ Per the spec this value is *produced by execution* — the EIP-2935 system call writes the parent hash into slot 0 at the start of block processing (`fork.py:903` is the history producer). It is derivable, and the parent hash is independently available from the header.

**The guest does not derive it.** With the preload disabled, the same lookup serves **zero** and the block is rejected; with the preload enabled it serves the BAL-declared value and the block is accepted. Control flow is otherwise identical between the two legs — `h_SLOAD` 17, `storage_read_record` 36, `h_SSTORE` 19, log-scan found 17, zero-store 0 — so the divergence is purely row **content**.

## Why this is FA-class

The guest consumes a proposer-asserted value for a slot it never proves and never computes. The natural backstop would be the post-state root — but **the root is currently BAL-fed** (`BlockVerdict.lean:5`), so a proposer that lies consistently in both the BAL and the header root is not contradicted by anything the guest derived independently.

⚠️ **The fix is not to supply a witness proof.** The pre-state is legitimately empty; demanding a proof for a slot that does not exist yet would be wrong. ⇒ **The guest must execute the system call and let its write land**, so the value is produced by execution exactly as the spec produces it. That is convergence, not compensation.

## Relationship to work in flight

- **#10651 / PR #11193** — computes the state root from operations rather than from the BAL. That removes the mechanism by which a consistent lie survives, so it materially reduces exploitability. **It does not by itself make the guest derive slot 0.**
- **#11176** — retiring the eager preload *exposes* this: the row flips to a false reject because nothing else produces the value. ⇒ **This issue is a prerequisite for that retirement**, not a consequence of it.
- **#11183 / #10612** — the guest must not read the provided BAL except to hash it. This is a Class-A instance: a declared value reaching execution.

## Closing conditions

1. **The guest produces slot 0's value by executing the EIP-2935 system call**, not by reading the declared BAL.
2. **A control proving it**: with the preload disabled, the row passes — because the value is now derived rather than seeded.
3. **A negative control**: a fixture (constructed if the corpus lacks one) where the declared BAL carries a *wrong* history value and the guest **rejects**. ⛔ Without this, the fix is unverified against the exact attack it closes.
4. The same question asked of the **other system-call producers** — beacon roots (`fork.py:897`) has the identical shape and should be checked rather than assumed correct.

## Provenance

⚠️ Both agents flagged the same limit: this is **one row**. Existence is established; **rate is not.** Whether other system-call-written slots are served the same way is closing condition 4 and is not yet measured.

pirapira:
## Summary

Move the modeled EIP-2935/EIP-4788 system-storage descriptor pass to the
pre-user-transaction setup in MTx, and seed the canonical block storage map
from those derived rows. The existing terminal replay still owns the side-log
and BAL event publication, so the pre-user pass does not duplicate terminal
evidence. The three descriptor calls retain their `vlen` no-op guards.

This fixes the visibility ordering: `h_SLOAD` can resolve system writes while
user transactions execute, as required by the execution-specs ordering. The
MTx site passes `bv_exec_p - 60` because `bv_exec_p` points 60 bytes into the
SSZ payload at that site. The BmvMx `+60` site is intentionally different: it
uses the execution-payload-field pointer directly. There is no second offset
defect.

## Evidence

The direct control has preload disabled. Input SHA256:

`2eeb29d355a8c833d6a159e10a7716982341b8f9dc8d6120877a2ff771212f8e`

Before this change the guest rejected it (`output[32] = 00`); with the
pre-user derived map it accepts (`output[32] = 01`). The first h_SLOAD lookup
is after the two pre-user upserts, and latest-value returns `a0 = 1`. The
history row is address
`335b175320002935/27f1c53a10cb7a02/f908/0`, slot `0x00`, BE value
`096fc95415c2acbcab9841f876424419fb2a2cd987cc7ee6011ed48240db6524` (the
parent hash). The beacon timestamp row is address
`b8bb8522d0beac02/d732807ef1319fb7/0f3df6`, slot `0x0c`, value `0x0c`. This
fixture has an all-zero parent beacon root, so its root descriptor has `vlen =
0` and correctly performs no map insertion.

A second, nonzero-root fixture confirms the third row rather than relying on a
row count. In `00021_test_multi_block_beacon_root_timestamp_calls...`, the
three pre-user upserts land at map offsets `0x0`, `0x80`, and `0x100` (trace
lines `2378464`, `2379253`, `2380322`):

* history: slot `0x04`, BE value
  `64ccdbd56510282318407652d42716a1defef8faeb41fabc5efe43e8ab083c11`;
* beacon timestamp: slot `0x40`, BE value `0000000100007ffc`;
* beacon root: slot `0x203f`, BE value
  `fe07a98784cd1850eae35ede546d7028e6bf9569108995fc410868db775e5e6a`.

The placement follows the spec: `fork.py:897-907` executes the beacon-roots
and history system transactions before the transaction loop at `fork.py:1040`.
The old guest placed the derived rows in terminal `block_state_root`, after
user execution; this change aligns the guest with the specified ordering.

## A/B regression

Uniform random-200 sample, seed `20260802`, same manifest, Spike, and
`SPIKE_OUTPUT_LEN=256`:

* candidate ELF: `5cebd0ac540275d1775339cf8fbf6040e12187585fc83fb7e7f7ffdf7efcdad0`
  (`.text = 0x06441c`), `FA=0`, `FR=12`, `OK=188`;
* clean parent ELF: `fdc7f5adddec67379ba97d1e30a8adf750209bbf5f95fbd81ed4c75693ff54b6`
  (`.text = 0x064378`), `FA=0`, `FR=13`, `OK=187`.

The full 256-byte candidate/parent outputs differ in exactly one sampled row:
`00021_test_multi_block_beacon_root_timestamp_calls...`; the candidate has
`output[32] = 01` matching the fixture oracle, while the parent has
`output[32] = 00`. This is the expected forward false-reject removal on an
EIP-4788 row that was not the history control.

## Checks

* `lake build` (4798 jobs)
* `scripts/check-region-map.sh` (including regenerated symbol-address TSV)
* `scripts/check-no-warnings.sh` — 0 EvmAsm warnings
* `scripts/check-layering.sh`
* `scripts/check-unimported.sh` — 2828 reachable / 2828 total
* `scripts/check-forbidden-tactics.sh`

Closes #11194. Related to #11100 and unblocks #11176.


pirapira:
**Labelled.** `CONFLICTING` against current main so CI will not schedule a build; resting on the local gates and the traces, as with several PRs today. The merge bot rebases and regenerates.

**This is wrong-stage, not wrong-value — the ninth instance of that category today, and the second to be a soundness issue.** `append_modeled_system_storage_tuple_rows` ran in terminal `block_state_root`, *after* the MTx user loop, so its upserts landed after every transaction that could have read them. `h_SLOAD` saw `storage_writes_count = 0`, skipped `storage_writes_block_latest_value` entirely, and fell through to a witness that legitimately has no such leaf. ⇒ **The value was never missing. It was produced too late to be read**, and the eager preload silently substituted the *proposer's declared* value instead.

**The spec puts the fix where the fix goes**, which makes this convergence rather than a workaround: `fork.py:897-907` runs the beacon-roots and history system transactions **before** the transaction loop at `:1040`.

## Why the evidence is convincing

**The control is a single decisive flip**: with the preload disabled, this row **rejected before and accepts now** — because the value is derived rather than seeded. Cited by input SHA, not by index.

**Content, not counts.** All three rows are traced with slot and value: history at slot `0x04` with BE `64ccdbd5…`; beacon timestamp at slot `0x40`, BE `0000000100007ffc`; **beacon root at slot `0x203f`, BE `fe07a987…` — nonzero and actually inserted**, at map offsets `0x0`/`0x80`/`0x100`.

⚠️ That distinction is not pedantry here: an **earlier build of this same change had map count 3 with every row at the wrong slot** (`bv_exec_p` is SSZ base + 60, so the descriptors read the wrong bytes and derived history slot `0xfff`). **The count passed in both the broken and the correct build.** A lookup miss and a lookup of the wrong key produce the same observable; only slot-and-value separates them.

**Two fixtures were needed to cover three rows**, and both were used: on the history control the beacon-root descriptor has `vlen = 0` because `parentBeaconBlockRoot` is zero, so that fixture *cannot* exercise it. The beacon fixture supplies the nonzero root row.

⭐ **The strongest single result is a forward flip on a row the change was not tuned for.** Random-200, FA=0 both legs, **FR 13 → 12, one flip, zero reverse** — and the flipped row is `multi_block_beacon_root_timestamp_calls`, i.e. the intended EIP-4788 effect appearing on an unselected fixture. A null would have been acceptable; a correct forward flip on a row you did not choose is better evidence than either.

**The `bv_exec_p` asymmetry is resolved as intentional, not a second defect**: MTx subtracts 60 to recover the SSZ base; `BmvMx` adds 60 because it points at execution-payload fields. Recorded in the body so the next reader does not assume one of the two sites is wrong.

Closes #11194 and the storage-visibility half of #11100. ⇒ **This unblocks #11176** — a 21.755 MiB retirement in two independently-retirable halves (user 12.219, system 9.537), enumerated and waiting on precisely this control.

PR #11207, opened 2026-08-02
docs(dispatch): the callee seed loop is NOT inert — it is the live consumer of the eager preload

pirapira:
## A docstring that tells the next reader a **live consumer** is byte-identical dead weight

GH #11176. `emitCalleeStorageSeedLoop`'s docstring claimed:

> `callee_seed_count` is 0 for every current caller (the top-level guest uses a different prologue; the verdict has not populated the table yet), so this is **INERT** — depth-0 / recipient behaviour is **byte-identical**.

**That was true when the loop was written and is false now.** The enumeration the *same sentence* names as future work — 1.6.4.2.b, `seed_callee_storage` in `dispatch_tx_runtime_code` — has since landed, and it populates the table on the ordinary verdict path. ⇒ **this loop is the live consumer of the eager BAL-sourced storage preload.**

### Measured, not argued

Cited by **input sha256** rather than index, because an index is selector-relative — two agents reading "fixture 00192" from two manifests differed by three orders of magnitude on the same day:

* input sha256 `242fa1e11ec51d7b9f9395b93d3e02ebf6c7d9064196360eba1cefd367cdb13b`
* relpath `blockchain_tests/for_amsterdam/amsterdam/eip7928_block_level_access_lists/block_access_lists_eip7002/bal_7002_request_from_contract.json`

A `SPIKE_COMMITLOG` dump of **every** access to `callee_seed_count` (35 hits) shows it **walk `0 → 1 → 2 → 3 → 4` and keep going** — bumps at the producer's store, read-backs at the two loads this loop uses.

And independently: disabling the producer flips **15 of 1,045** sampled rows from reject to accept with **zero** reverse flips and FA = 0 on both legs. **An inert loop cannot do that.**

### Why this is worth its own PR

The retired sentence is precisely the one that would license **deleting this loop as dead weight without measuring anything** — "byte-identical" is a claim a reader acts on, and I was one step from acting on it while scoping #11176's retirement. It is the same class of defect as the `StorageWriteMap` "needs no overflow path" comment (#11189) and the `RegionMap` mechanism note (#11203): a stale reachability assertion in the file people read to decide what is safe to remove.

⚠️ **The general form, worth more than this one fix:** trust the **why** in a comment like that; measure the **happens**. Every such claim here has been a snapshot of a tree that has since moved.

### Scope

Comment only. **No emitted-asm change** — the edit is inside a `/-- … -/` docstring, so nothing reaches the `.s`. `lake build EvmAsm.Codegen.Dispatch` clean (1119 jobs).

### Not in scope

The retirement itself. It is **parked**, and #11176 records why: `seed_callee_storage` has **two** products — the storage seed table *and* `callee_balance_table`, a pre-resolved per-account balance read by `call_frame_descend` to stage a child frame's `env+32` for nested `SELFBALANCE` — and the 1,045-row A/B is **blind to the second one** (`h_SELFBALANCE` executes **zero** times on the row measured). A control that cannot fire proves nothing.

### DIFF
--- a/EvmAsm/Codegen/Dispatch.lean
+++ b/EvmAsm/Codegen/Dispatch.lean
@@ -3035,10 +3035,34 @@ def emitRuntimeDispatcherPrologue : String :=
     it appends one 128 B persistent-exec-log entry (addrHash, slotKey, original=value,
     current=value — mirrors `exec_log_append_storage_seed`) and bumps
     env.persistentLogLength (env+448), so a nested callee's SLOAD finds its witness
-    value instead of cold 0. `callee_seed_count` is 0 for every current caller (the
-    top-level guest uses a different prologue; the verdict has not populated the table
-    yet), so this is INERT — depth-0 / recipient behaviour is byte-identical. The
-    enumeration that fills the table is 1.6.4.2.b (dispatch_tx_runtime_code). Uses only
+    value instead of cold 0.
+
+    ⛔ **THIS IS NOT INERT.** This docstring used to claim `callee_seed_count` "is 0 for
+    every current caller … so this is INERT — depth-0 / recipient behaviour is
+    byte-identical". **That was true when the loop was written and is false now**: the
+    enumeration it names as future work — 1.6.4.2.b, `seed_callee_storage` in
+    `dispatch_tx_runtime_code` — has since landed, and it populates the table on the
+    ordinary verdict path. ⇒ **this loop is the live consumer of the eager BAL-sourced
+    storage preload.** GH #11176.
+
+    MEASURED (2026-08-02), not inferred. Cited by input sha256 rather than index,
+    because an index is selector-relative and two agents reading "fixture 00192" from
+    two manifests differed by three orders of magnitude on the same day:
+      input sha256 `242fa1e11ec51d7b9f9395b93d3e02ebf6c7d9064196360eba1cefd367cdb13b`
+      relpath `blockchain_tests/for_amsterdam/amsterdam/eip7928_block_level_access_lists/`
+              `block_access_lists_eip7002/bal_7002_request_from_contract.json`
+    A commitlog dump of every access to
+    `callee_seed_count` shows it **walk 0 → 1 → 2 → 3 → 4 and keep going**, with the
+    bumps at the producer's own store and read-backs at the two loads this loop uses.
+    And disabling the producer flips **15 of 1,045** sampled rows from reject to accept
+    with **zero** reverse flips — an inert loop cannot do that.
+
+    ⚠️ **Why this mattered enough to write down:** the retired sentence is exactly the
+    one that would license deleting this loop as dead weight *without measuring
+    anything* — "byte-identical" is a claim a reader acts on. Trust the **why** in a
+    comment like that; measure the **happens**.
+
+    Uses only
     x5/x6/x7 (the dispatch loop re-inits them each iteration) and x28..x31 temps; never
     touches x10/x12/x13/x20/x21. -/
 def emitCalleeStorageSeedLoop : String :=


pirapira:
**Reviewed — label on green build.** Comment-only inside a docstring, one file, nothing reaches the `.s`.

**The retired sentence is the most dangerous kind of wrong: it was true when written.** The docstring claimed `callee_seed_count` "is 0 for every current caller … so this is INERT — depth-0 / recipient behaviour is byte-identical", and named `1.6.4.2.b` as the future enumeration that would fill the table. ⇒ **That enumeration has since landed.** The loop is now the live consumer of the eager BAL-sourced storage preload, and the comment still says it does nothing.

⛔ **It nearly cost a deletion tonight.** The #11176 retirement was authorised and in progress; this comment states the consumer loop can be removed without behavioural effect. **The measurement contradicted it** — `callee_seed_count` walks 0 → 1 → 2 → 3 → 4 and keeps going, and disabling the producer flips **15 of 1,045** sampled rows with **zero** reverse. An inert loop cannot do that.

**This is the fifth false reachability claim found today**, and four of the five asserted something was dead or bounded when it was neither:

| location | claim | reality |
|---|---|---|
| `StorageWriteMap` | journal needs no overflow path | fail-open, fixed in #11200 |
| `RegionMap` | same, plus wrong mechanism | corrected in #11203 |
| account undo source | "inherits exec log 16384 cap" | it is not fed by the exec log |
| `Dispatch:3037` | loop is INERT | live consumer — **this PR** |
| `seed_callee_storage` | (omission) one product | also fills `callee_balance_table` |

⇒ The pattern is not that comments go stale occasionally. **Confident reachability claims in this codebase are unreliable as a class**, and two of these directly caused wrong work: an analyst retracted a correct #11189 finding after reading one, and this one licensed a deletion.

**Citing by input SHA rather than index is right and the reason belongs in the record**: two agents read "fixture 00192" from two manifests today and differed by three orders of magnitude on the same block. An index is selector-relative, a name truncates at 96 characters, and four `return50000` rows share a relpath while differing only in a gas parameter. **Only the input SHA survives.**

⚠️ **What this PR deliberately does not do**: it does not delete the loop. The retirement is parked because the A/B cannot see the second consumer — `h_SELFBALANCE` executes **zero** times on the sampled row while the balance table is populated 35 times, so forward-15/reverse-0 is evidence about the storage product and **silent** about the balance product. Parking that is my call, not the author's, and the rationale is on #11176 with the sequence for whoever resumes it.

pirapira:

This retires the most dangerous of the five false reachability claims found today, because it is the only one that would have licensed a deletion **that was already authorised and in progress**.

The retired sentence said the callee seed loop is INERT because `callee_seed_count` is zero for every current caller — and named `1.6.4.2.b` as the future enumeration that would populate it. ⇒ **That enumeration has since landed.** The comment was true when written and false now, which is the failure mode no amount of care at review time catches: nothing changed in this file, so nothing prompted a re-read.

**Measured, not inferred**: `callee_seed_count` walks 0 → 1 → 2 → 3 → 4 and keeps going, and disabling the producer flips **15 of 1,045** sampled rows with **zero** reverse. An inert loop cannot do that.

⇒ And the correction is cited **by input SHA rather than index**, with the reason stated in-place: two agents read "fixture 00192" from two manifests today and differed by three orders of magnitude on the same block, because four `return50000` rows share a relpath and differ only in a gas parameter the 96-character truncation hides. **A name is not an identity, an index is not an identity, and a relpath is not an identity.**

## What it deliberately does not do

It does not delete the loop. That retirement is parked, and the reason is worth recording alongside this: `seed_callee_storage` has **two** products, and the second one — `callee_balance_table`, read by `call_frame_descend` to stage a child frame's `env+32` — feeds **nested SELFBALANCE**. ⚠️ The 1,045-row A/B could not see it: `h_SELFBALANCE` executed **zero** times on the row measured while the balance table was populated 35 times. ⇒ Forward-15/reverse-0 is evidence about the storage product and **silent** about the balance product.

⇒ So the deletion needs a demand-driven SELFBALANCE first — *land the lazy read, then retire the eager producer*, the same shape as #10874 and #11206. That is now a closing condition on #11176, and parking it is my call rather than the author's.

PR #11209, opened 2026-08-02
fix(codegen): order storage-read recording after access charge

### LINKED ISSUE #11190: SLOAD/SSTORE append execution-log rows before gas charge on OOG
## Finding

The emitted SLOAD and SSTORE handlers append a durable execution-log row before charging the storage-access gas. On an out-of-gas path, the row survives even though execution-specs charges before the operation.

Pinned execution-specs (e5a8caf1b8055e4d805c7fb169edfa710914b7da):
- `src/ethereum/forks/bpo2/vm/instructions/storage.py:43-54`: SLOAD charges WARM_ACCESS/COLD_STORAGE_ACCESS before `get_storage`.
- `storage.py:73-126`: SSTORE must read original/current state to price the operation, then `charge_gas` at line 123 before `set_storage` at 126. The pre-state read is required; the defect is the guest's durable log append before charging, not the read itself.

## Emitted evidence

Current emitted artifact: source/codegen `7d9824deccf49011b86132f8db6a7a1995c8d4c8`, generated HEAD `6e8c63e3d5b0682345ce9fd83e58aa0e23e5dbc2`, guest SHA `50fbd45b649c92a41da6aa3e899701d4695680885d47e05867fbc1b8e0321ee1`, runner SHA `5b383a7702fb05dea16a2cc37dc313b82491fdf194f39c12306cb387a5a2b6b0`.

- `h_SLOAD` calls `storage_read_record` before `evm_storage_access_charge_key`.
- `h_SSTORE` calls `storage_read_record` before `evm_storage_access_charge_key` and before its dynamic charge.

Forced reachability diagnostic (not a proposed patch): copied the assembly outside the repo, forced the first log-key comparison to miss, and raised the storage charge immediate from 2900 to 100000000. On fixture `00094...oog_at_eip2200_stipend`, SSTORE began with `storage_reads_count=2`; it appended at PC `0x80028020` (count 3), then reached `.exit_outofgas` at `0x80062624` with count still 3. On fixture `00036...SSTORE_SLOAD-debug`, SLOAD began at count 4; it appended (count 5) then reached `.exit_outofgas` with count 5.

Diagnostic ELF SHA: `6dc4f618b47d7af830e294050e8012595fdbd36e970e2f58a57a3e1e26164abe`. No repository files were changed.

## Consequence and scope

The execution log currently has 16,384 rows and #11187 shows that this is already far below its derived bound; rows spent on halting paths make that shortage worse. The same write-side pressure interacts with the undo-journal bound in #11189.

No natural fixture in the sample exercised a new-key storage-access OOG; the forced diagnostic is required to make the condition reachable. This is an ordering/resource-consumption defect, distinct from the charge-ordering census in #11185.

pirapira:
## Summary

Fixes #11190 by moving the durable `storage_read_record` call below `evm_storage_access_charge_key` in both `h_SLOAD` and `h_SSTORE`.

This is stacked on **#11206** (branch base is #11206's HEAD `478f1a6c4`); #11206 must merge first. This PR does not modify #11206.

The execution-specs ordering is the basis:

- `storage.py:43-54`: SLOAD charges warm/cold access gas before `get_storage`.
- `storage.py:73-126`: SSTORE performs the access-cost `check_gas` before `get_storage_original`/`get_storage`, then performs value-dependent pricing and the regular/state-gas charge.

SSTORE still records its required prestate read before value-dependent pricing. Only a cold-access OOG is prevented from leaving a durable read row.

## Targeted control table

All runs use the rebuilt guest and SPIKE write-watch on `tx_storage_reads_count @ 0xbc004788`.

| Fixture | Path | Read rows after fix |
|---|---|---:|
| 01213 `...sload...oog_at_sload` | SLOAD cold-access OOG | 0 |
| 01216 `...sstore...oog_above_stipend_below_access` | SSTORE cold-access OOG | 0 |
| 01217 `...sstore...oog_at_eip2200_stipend` | SSTORE stipend failure | 0 |
| 01218 `...sstore...oog_at_eip2200_stipend_plus_1` | SSTORE cold-access OOG | 0 |
| 01214 `...successful_sload` | successful SLOAD | 1 |
| 01215 `...access_covered_oog_on_write` | SSTORE dynamic-gas OOG after access | 1 |
| 01219 `...oog_at_exact_gas_minus_1` | SSTORE dynamic-gas OOG after access | 1 |

The positive controls are required: a blanket suppression would incorrectly remove the reads from 01214/01215/01219. The 16,384-entry execution-log/read arena is fail-closed, so rows spent on halting paths consume capacity that valid blocks need.

## User-visible result

On the two reproducing failures, output byte 32 changes from rejection to acceptance:

- 01213: `00 -> 01`
- 01218: `00 -> 01`

The persistent SSTORE execution-log append remains after dynamic gas checks and is not changed by this patch.

## Broader check

Seeded random-200 Spike sweep, same manifest as the #11206 candidate:

- candidate ELF SHA256: `687b796a707defc2cd15262569646c1cc607890006f4c386a6ba951b79079920`
- `FA=0`, `FR=12`, `OK=188`
- candidate TSV is byte-identical to the prior #11206 candidate TSV

Validation: `lake build` passed all 4,806 jobs with no warnings.


pirapira:
**Labelled.** `CONFLICTING` against current main so CI will not schedule a build; resting on the local gates — `lake build` 4,806 jobs, no warnings — as with several PRs today. Stacked on #11206 (labelled, unmerged), which the body states explicitly.

⭐ **The control table is the best-designed one I have reviewed today, because it runs both directions.**

| fixtures | expected | measured |
|---|---|---|
| `01213`, `01216`, `01217`, `01218` — cold/stipend access failures | no row | **zero** reads recorded |
| `01214`, `01215`, `01219` — success, access-covered dynamic OOG, exact-gas−1 | one row | **one** read recorded |

⇒ **A one-direction check would have passed for a patch that simply stopped recording.** The positive controls are what separate a fix from a deletion — and **`01215` is the sharpest**: an out-of-gas that occurs *after* the access cost must **still** record its read, which is precisely what a careless move of the call would have broken.

**And the defence that makes this valid was preserved.** SSTORE's prestate read is *required* to price the operation — the defect was never that it reads first, but that it **records a durable row** first. This moves the record below `evm_storage_access_charge_key` while keeping it above the value-dependent pricing, so only the cold-access OOG is prevented from leaving a row.

**Why it is not cosmetic:** the execution log is hard-capped at **16,384** with a **fail-closed** guard. ⇒ A row spent on a halting path is a row a legitimate block cannot use, and the guard rejects rather than continuing — so wasted rows push a valid block toward a false reject. The user-visible effect is `byte 32: 00 → 01` on `01213` and `01218`.

**Breadth: random-200 run before asking for a label**, same manifest and runner as the #11206 candidate — **FA=0, FR=12, OK=188, and the candidate TSV byte-identical to the prior candidate's.** ⇒ Changing recording behaviour on every SLOAD and SSTORE in the guest moved nothing outside the targeted cases.

That sweep was requested rather than volunteered on a sibling PR earlier today and it held that one up for an hour; here it was run unprompted before the label was sought. Spec basis is `storage.py:43-54` (read path) and `:73-126` (write path) — charge before the effect.

PR #11210, opened 2026-08-02
fix(codegen): drop EIP-2935/4788 from mtx storage skip-list (#10684)

pirapira:
## Summary

Narrow structural drop for **#10684**: remove **EIP-2935** and **EIP-4788** only from the multi-tx storage comparator skip-list. Spec has no address-based BAL exclusion for those accounts. In-body modeled skip was already removed (`f5be2e440`); this finishes the skip-list feed half for the pair.

**Not in this PR:** the other four residual system skips (EIP-7002 / 7251 / 6110 / SYSTEM_ADDRESS) — still copied. Full skip-list retirement still gated on one-container / producer work (codex2 #11206 lane).

## Change

| before | after |
|---|---|
| `la bbcv_sys_2935; li t4, 6` | `la bbcv_sys_7002; li t4, 4` |
| count `2N+7` | count `2N+5` |
| `bvMtxSystemSkipEntries = 6` | `= 4` |
| skip-list BSS | −64 B (`bssSizeBytes` 0x1b6c2ca0 → 0x1b6c2c60) |

Files: `BlockVerdictMtxTail.lean`, `BlockVerdictParams.lean`, comments in DataSectionTail; layout regen (RegionMap / GuestAddrs / TSV / GuestImage).

## Prior force-remove (half-done measurement)

On #11206 branch force-patch `+7→+1` (drop all six system):
- beacon / multi_block_beacon still **PASS**
- multi_type **23625** still **fc=37** both legs → 2935/4788 skip not load-bearing for that FR

This PR is the **narrow** pair-only cut with full A/B on main.

## A/B random-200 seed42 (`/tmp/11167-random200.tsv`)

|---|---|---:|---:|---:|
| parent `2e80a07f3` | `84cfb495…0a01` | 0 | 5 | 195 |
| cand `432893f1e` | `40b62d75…650b` | 0 | 5 | 195 |

- ELFs **DISTINCT**
- FR sets **EQUAL**
- reverse / forward flips **0**
- OUTPUT 256B **byte-identical 200/200**

## Emitting confirm

```
la t1, bbcv_sys_7002; li t4, 4
...
addi t3, t3, 5   # count = 2N+5
```

`bbcv_sys_2935` / `4788` symbols remain (other consumers: `bal_account_is_modeled_system` / code-preimages). Only the mtx skip-list copy starts at 7002.

## Out of scope / keep

- `i3djw_skip_list` single-tx path (dead path per #10661) not touched
- Residual four system skips stay until producer/container convergence
- fc37 multi_type remains independent (CONTAINER_DEPENDENT)

Closes nothing alone — advances #10684 pair half. Residual four still tracked on the issue.


pirapira:
**Reviewed — label on green build.** This is scoped exactly to what was measured, which is why it is reviewable.

**Spec basis is the strongest part:** execution-specs has **no address-based BAL exclusion** anywhere. It builds one BAL from execution and compares hashes. ⇒ A hardcoded address allowlist in the comparator is a guest-only structure with no counterpart, and those never converge on their own.

**The scoping is the right kind of narrow.** Only EIP-2935 and EIP-4788 come out; the other four residual skips — 7002, 7251, 6110, SYSTEM_ADDRESS — stay and are named as staying, gated on the one-container/producer work. ⇒ A reviewer can tell what was *not* attempted, which is what makes a partial change trustworthy rather than suspicious.

**And the deletion is bounded by what was actually tested.** The force-remove probe dropped **all six** (`+7 → +1`) and showed beacon and `multi_block_beacon` still passing, with `multi_type 23625` failing `fc=37` on **both legs** — an independent failure, not the skip. ⇒ That evidence covers "these skips are not load-bearing on those rows"; it does **not** license removing all six, and this PR removes two. **The measurement is wider than the change**, which is the correct direction.

**Behaviour:** FA 0, FR 5, reverse 0, and **256-byte output byte-identical on 200/200**. For a comparator change that last figure is the right claim — not that the same rows failed, but that nothing observable moved.

Layout is `−64 B` (`bssSizeBytes 0x1b6c2ca0 → 0x1b6c2c60`), consistent with two entries leaving a skip table, with pins regenerated rather than hand-set.

⚠️ One thing worth stating in the body if it is not already: **why these two and not the other four.** 2935 and 4788 are the block-level BAI-0 predeploys whose rows the guest now produces pre-user (#11206), so the comparator no longer needs to look away from them. The remaining four are deferred system-request predeploys whose producer story is different — that is the distinction doing the work here, and without it the split looks arbitrary.

The `#10684` premise was also re-derived **against #11206's branch rather than main**, which was necessary: the producer is labelled but unmerged, so measuring on main would have said the skip was still required and been measuring the wrong tree.

pirapira:
**Labelled — `build` COMPLETED SUCCESS.** Now `CONFLICTING` because main moved (#11208 landed carrying #11203/#11204/#11205); that is the merge bot's to resolve, and the conflicting files are generated layout facts that must be **regenerated from the post-merge tree**, never merged textually.

**The scoping is what makes this reviewable.** Only EIP-2935 and EIP-4788 come out of the comparator skip-list; the other four — 7002, 7251, 6110, SYSTEM_ADDRESS — stay, and are **named as staying** with their gate. ⇒ A reader can see what was *not* attempted, which is what makes a partial change trustworthy rather than suspicious.

**And the measurement is wider than the change**, which is the correct direction: the force-remove probe dropped **all six** (`+7 → +1`) and showed beacon and `multi_block_beacon` still passing, with `multi_type 23625` failing `fc=37` on **both legs** — an independent failure, not the skip. That evidence licenses removing *these two*; it does not license removing six, and this PR removes two.

**Spec basis is the strongest part:** execution-specs has **no address-based BAL exclusion anywhere**. It builds one BAL from execution and compares hashes. ⇒ A hardcoded address allowlist in the comparator is a guest-only structure with no counterpart, and those never converge on their own.

**Behaviour:** FA 0, FR 5, reverse 0, **256-byte output byte-identical on 200/200** — for a comparator change, that is the right claim: not that the same rows failed, but that nothing observable moved. Layout `−64 B`, consistent with two entries leaving a skip table, pins regenerated rather than hand-set.

⚠️ The premise was re-derived **against #11206's branch rather than main**, which was necessary: the BAI-0 producer is labelled but unmerged, so measuring on main would have said the skip was still required and been measuring the wrong tree. That distinction is worth keeping in mind for the remaining four — **their gate moves when the producer lands, not before.**

pirapira:
Four-skip sentinel after #11206: data-poison population floor **0** on stride-1045 + random-200 for 7002/7251/6110/SYSTEM (and for BAI-0 pair). SPIKE confirmed poison lands in `bv_mtx_skip_list`. Full write-up on #10684. #11210 (drop 2935/4788) still stands on prior force-remove evidence; the four deferred are a separate gate not moved by this sample. No four-skip deletion without constructed fixture.

Issue #11211, opened 2026-08-02
cleanup: BmvMx precompute writes comparison flags with no emitted consumer


pirapira:
## Finding

On current `origin/main` `2e80a07f3dfa21b704fc0f97f7c3a9e5336fbdbe`, the emitted
guest still runs `blockVerdictBmvMxPrecomputePrefix` from
`EvmAsm/Codegen/Programs/BlockVerdictFunction.lean:90`.  The prefix is mixed:
its `bv_exec_p` initialisation is live and is consumed by the no-transaction and
MTx parsing path (the surrounding code at `:104-111` documents that use), so
this is **not** a proposal to delete the whole prefix.

The prefix also computes and stores these comparison flags:

- `bmvmx_avail`
- `bmvmx_sender_match`
- `bmvmx_sender_checked`
- `bmvmx_coinbase_match`
- `bmvmx_coinbase_checked`

## Emitted evidence

The exact current-main guest is `/tmp/grok-10684-narrow/parent-wt/gen-out/regionmap/stateless_guest.elf`, SHA-256
`84cfb495490ad995551b860b94e37e9e3ce775a759a3db5cdc69f37593ce0a01`.
In its emitted `.s`, each of the five symbols occurs only at direct `sd`
writes plus its data definition; there is no `ld`, load, or conditional branch
reading any of them.  The computed-address audit also found no effective
`ld`/`lbu` address that can land on a flag: nearby loops operate on the context,
address, and balance buffers, while every flag access uses a direct symbol
address.  This is an emitted-image observation, not only a source-name grep.

The flags are therefore a separate **consumerless work** finding.  It is not a
single-tx/MTx duplicate-removal candidate: removing them requires a separate
decision about whether to retire the unused comparison computation, while the
live `bv_exec_p` setup must remain.

This is filed separately from #10685's two source-only dead definitions
(`BlockVerdictCreateCollisionBranch` and `BlockVerdictEoaBodyEffectReconcile`).
No source change or removal is proposed here.

pirapira:
## Re-derived on current `main` `9ffb5398a` — **write-only confirmed in BOTH build units**, and the cut is four holes, not one

[...]

PR #11212, opened 2026-08-02
chore(codegen): remove source-only single-tx twins

pirapira:
# Remove source-only single-transaction twins

## Summary

Remove the two Codegen definitions that no longer have a caller or concatenation site:

- `blockVerdictCreateCollisionBranch` from `BlockVerdictCreateCollision.lean`
- `blockVerdictEoaBodyEffectReconcile` from `BlockVerdictEoaBodyEffectReconcile.lean`

The current guest emits the universal multi-transaction loop and its live creation route. The execution-specs transaction loop is the single loop at `fork.py:913-914`; these former single-transaction definitions are not part of the emitted program.

## Emitted evidence

The audit was performed against current `origin/main` commit `2e80a07f3dfa21b704fc0f97f7c3a9e5336fbdbe`.

- Parent guest ELF: `84cfb495490ad995551b860b94e37e9e3ce775a759a3db5cdc69f37593ce0a01`.
- Candidate guest ELF after deletion: the same SHA, `84cfb495490ad995551b860b94e37e9e3ce775a759a3db5cdc69f37593ce0a01`.
- Candidate emitted assembly: the same SHA as the parent, `9c203e267babdb7ba719570a1d12964230fe164715a01f7cda69eb2fdb877c1d`.
- The old creation labels `Lbv_creation_dispatch`, `Lbv_creation_runtime_try`, and `Lbv_creation_collision` occur zero times.
- The old EOA labels `Lbv_eoa_body_effect_reconcile`, `Lbv_i3sk*`, and `Lbv_i3sksys*` occur zero times. `i3djw_skip_list` remains only as its BSS data label; no code reference remains.
- Live controls remain present: `block_verdict`, `block_verdict_mtx_oog_materialize`, `.Lbv_mtx_creation_collision`, `.Lbv_tx_gas_precharge_pc0_prefix`, and `bv_emit_single_tx_tl7708`.
- The predecessor at the old boundary is an unconditional jump at assembly line 16851 to `.Lbv_after_tx_gas_precharge`; line 16852 starts `block_verdict_mtx_oog_materialize`. Since the removed labels are absent, no deleted label can be a fallthrough target.

As an independent whole-image identity check, a separate build at commit `5da6981a589a20d3c733a2c8c034c2d2ed64c191` produced ELF SHA `eb4dfec4c92a8686eef7484fb587c071b6021b539094df2248fa6da3fbe53df2`, distinct from the current-main ELF. The `block_verdict` region in both images is 63,754 bytes with SHA `cf87b5f34c2285fd08ae4162973401a5f36e92bb2e42817b49358a61cea7167a`. This proves region identity across distinct whole artifacts rather than comparing an artifact with itself.

The one Spike run was only a smoke control (`00064_test_top_frame_state_charge_empty_precompile_fork_Amsterdam-blockchain_test_from_state_test__b0`), not a claim that either dead definition is exercised. It returned 256 bytes with rc=0 using runner SHA `5b383a7702fb05dea16a2cc37dc313b82491fdf194f39c12306cb387a5a2b6b0`.

## Scope boundaries

The BmvMx precompute is deliberately not part of this PR. Its live `bv_exec_p` initialization remains; the consumerless comparison flags are tracked separately in issue #11211.

## Verification

- `lake build` completed successfully (4,796 jobs).
- `scripts/check-unimported.sh` passed: 2,826 reachable modules, 0 orphans.
- `scripts/check-layering.sh` passed.
- `scripts/check-file-size.sh` passed: 2,158 files within cap.
- `scripts/check-asm-to-program.sh` passed: 386 converted definitions across 147 files.
- `git diff --check` passed.

This is source cleanup only; the emitted image is byte-identical, so no EEST sweep is needed.


pirapira:
**Reviewed — label on green build.** `+16 / −246` across 8 files, and the emitted image is **byte-identical to current main**, so no EEST sweep is needed: a byte-identical emission covers **all** inputs, which no 200-row sample can match.

⚠️ **And the identity is not vacuous, which is the part that makes it usable.** An ELF compared against itself reports identical and means nothing — that trap produced three false passes today, one of them a neutrality check that had to be redone. Here the check is constructed so it *cannot* be vacuous:

- current main ELF `84cfb495…` and an independent pre-#11200 ELF `eb4dfec4…` — **different whole-file SHAs**, so the two builds are demonstrably distinct;
- the `block_verdict` region extracted through the next function label is **63,754 bytes with SHA `cf87b5f3…` in both**.

⇒ **A region identity across confirmed-different builds proves locality without the self-comparison hazard.** Worth reusing.

## Why these two are safe and the third is not

The deletion is **only** `BlockVerdictCreateCollision` and `BlockVerdictEoaBodyEffectReconcile` — definitions with no caller, no concatenation, and **no emitted labels at all** (`Lbv_creation_dispatch`, `Lbv_creation_runtime_try`, `Lbv_creation_collision`, `Lbv_eoa_body_effect_reconcile`, the `Lbv_i3sk` family — all absent from the image).

⭐ **The fallthrough check is discharged by construction rather than inspection**, which is the strongest form: the predecessor at `.s 16851` is `j .Lbv_after_tx_gas_precharge`, so nothing falls into the next function entry — **and an absent label has no predecessor at all.** That is stronger than "the predecessor ends in `ret`", which is how #11178 discharged the same check.

⛔ **BmvMx correctly stays out, filed as #11211.** The prefix is **mixed**: its `bv_exec_p` initialisation is live and consumed by no-tx and MTx parsing, while only the `bmvmx_*` compare flags are unread. ⇒ Removing it wholesale would take a live initialiser; splitting it is a different change with a different risk profile and does not belong in a two-definition deletion.

⇒ And #11211 carries the evidence that makes *"work with no reader"* a claim rather than a suspicion — flags at known `nm` addresses, every emitted reference a direct `la`-then-`sd`, no `ld`/`lbu` effective address able to land on them, neighbouring computed-address loops operating on different buffers. **That discharges precisely the computed-offset hazard that could not be excluded for `env+32` elsewhere today.**

⚠️ One thing this PR does **not** claim, correctly: that a name containing `single_tx` implies a retired route. `blockVerdictSingleTxTopLevelLogFunction` is still linked and `bv_emit_single_tx_tl7708` is **called** by the live direct publish path — so those stay, and the inventory says so.

pirapira:

`+16 / −246` and the **emitted image is byte-identical to current main**, so no EEST sweep applies: byte-identity covers **all** inputs, which no 200-row sample can match.

⚠️ **And the identity is constructed so it cannot be vacuous** — the trap that produced three false passes today, one of them a neutrality check that had to be redone. Two builds with **different whole-file SHAs**, and the `block_verdict` region extracted through the next function label identical at **63,754 bytes** in both. ⇒ **A region identity across confirmed-different builds proves locality without the self-comparison hazard.** That method is worth reusing.

⭐ **The fallthrough check is discharged in its strongest form**: not "the predecessor ends in `ret`" — which is how #11178 discharged it — but **the candidate labels are absent from the image entirely, so there is no predecessor to have.** Absence by construction beats absence by inspection.

**Scope held to what was proven**: only `BlockVerdictCreateCollision` and `BlockVerdictEoaBodyEffectReconcile`, both definitions with no caller, no concatenation, and no emitted labels.

⇒ And two things were correctly **kept out**:

- **BmvMx** — filed as #11211 instead, because the prefix is **mixed**: its `bv_exec_p` initialisation is live and consumed by no-tx and MTx parsing, while only the `bmvmx_*` compare flags are unread. Deleting it wholesale would have taken a live initialiser.
- **`blockVerdictSingleTxTopLevelLogFunction`** — still linked, and `bv_emit_single_tx_tl7708` is **called** by the live direct publish path. ⇒ **A name containing `single_tx` is not evidence of a retired route**, which is exactly the inference a source-name sweep would have made.

#11211 also carries the computed-offset audit — flags at known `nm` addresses, every emitted reference a direct `la`-then-`sd`, no effective address able to land on them — which discharges the hazard that could not be excluded for `env+32` elsewhere today, and turns *work with no reader* into a claim rather than a suspicion.

PR #11214, opened 2026-08-02
fix(codegen): charge CALL transfer gas before frame setup

### LINKED ISSUE #11191: CALL allocates child frame before CALL_VALUE gas charge on OOG
## Finding

`call_frame_descend` allocates and initializes the child frame before charging the CALL value-transfer gas. On an out-of-gas transfer charge, the guest has consumed a child frame before the charge that execution-specs performs before `generic_call`.

Pinned execution-specs commit `e5a8caf1b8055e4d805c7fb169edfa710914b7da`: `src/ethereum/forks/bpo2/vm/instructions/system.py:397-405` charges `message_call_gas.cost` before `generic_call` (the same ordering applies to CALLCODE at 459-492).

## Emitted evidence

Current emitted artifact: source/codegen `7d9824deccf49011b86132f8db6a7a1995c8d4c8`, generated HEAD `6e8c63e3d5b0682345ce9fd83e58aa0e23e5dbc2`, guest SHA `50fbd45b649c92a41da6aa3e899701d4695680885d47e05867fbc1b8e0321ee1`, runner SHA `5b383a7702fb05dea16a2cc37dc313b82491fdf194f39c12306cb387a5a2b6b0`.

`call_frame_descend` calls `frame_depth_push` and `call_frame_enter` before `.Lcfd_charge_transfer`, which subtracts 10300 when `value_nonzero`.

Forced reachability diagnostic (not a proposed patch): copied the assembly outside the repo, replaced the value-zero branch with `nop`, and raised the 10300 immediate to 100000000. On fixture `00148...create_oog_from_call_refunds...SStore_Refund_O`, debug stopped at `.Lcfd_charge_transfer` (`0x80043a60`) with `evm_call_depth=1`, then reached `.exit_outofgas` (`0x80062624`) with depth still 1. After continuing the unwind, depth returned to 0 at PC `0x80044ee4`; `frame_base` is pure depth-indexed (`basr_values + depth * 0x19000`) and no high-water frame pointer changed.

Diagnostic ELF SHA: `57bf04a46faa7e6058dcea109ee8aaec1ae2827ba0aee4fcd49874fb2084a6b9`. No repository files were changed.

## Consequence and scope

The frame slot is reclaimed on OOG unwind, so this is transient resource/step consumption rather than cumulative arena leakage. It does not create a 1025-slot permanent false-reject surface, but it is still a spec-ordering divergence and consumes frame setup work on a path that should charge first.

No natural fixture in the sampled corpus exercised a value-bearing transfer-charge OOG: four candidate controls were checked and `00148` reached `call_frame_descend` with a zero-value descriptor. The forced diagnostic is therefore required to make the condition reachable. This issue is the CALL-family sibling of #11185 and is filed separately because the consumed resource is a frame, not an execution-log row.

pirapira:
## Summary

Charge CALL/CALLCODE value-transfer gas in `call_frame_descend` before any
child-frame mutation. The caller's one-shot `cd_xfer_gas_precharged` flag is
consumed at the same early point; direct helper probes still perform the
charge there. The old late charge after `frame_depth_push` and
`call_frame_enter` is removed.

This matches execution-specs BPO2 `system.py:397-405`, where
`message_call_gas.cost` is charged before `generic_call`. It prevents a
transfer-charge OOG from consuming a transient child-frame slot or advancing
`evm_call_depth`. The frame arena is depth-indexed and reclaimed on unwind, so
this is spec alignment rather than a cumulative leak or known verdict fix.

This PR is intentionally separate and stacked on #11209 (which is stacked on
#11206); do not fold it into either PR.

## Targeted controls

The forced diagnostic was kept outside the repository, as required because
the sampled corpus does not naturally reach a value-bearing transfer-charge
OOG. It used the pinned `00148...create_oog_from_call_refunds...SStore_Refund_O`
input (SHA256
`599af9365052acfac52931940b5fa7395b94858b7343314f7024997117870eba`), replaced
the value-zero branch with `nop`, and changed the `10300` immediate to
`100000000` only in copied `/tmp` assembly.

* Candidate diagnostic: at the transfer charge (`0x8004414c`) and
  `.exit_outofgas` (`0x80062e04`), `evm_call_depth = 0`.
* Baseline diagnostic: at the corresponding late charge (`0x80044240`) and
  `.exit_outofgas`, `evm_call_depth = 1`.

The normal value-bearing success fixture
`00177_test_call_with_value_mainnet...` descended (`depth 0 -> 1 -> 0`) and
matched baseline output byte-for-byte (SHA256
`286bbda809d5f28ef8961b79a76188acc1f2b76693c007776e3c6859ad04cf5f`). The
post-transfer-charge OOG fixture
`15904_test_callcall_00_ooge_value_transfer...` descended to depth 2 before
unwinding (`0 -> 1 -> 2 -> 1 -> 0`); candidate and baseline outputs were also
byte-identical (SHA256
`de0c8a91d322dd23615084d728c98c05e8d81bc82cf6970ad6cf0e8a2df601be`).

## Random A/B

Uniform random-200 sample, seed `20260802`, same `tests-zkevm@v0.6.2`
manifest, Spike, and 8 jobs:

* candidate ELF SHA256: `6fa4953afe95194cd8a664c119da7ac99d6db88628d6ed667c3eb12f7d219d11`
* baseline ELF SHA256: `5cebd0ac540275d1775339cf8fbf6040e12187585fc83fb7e7f7ffdf7efcdad0`
* both legs: 200 scored, `FA=0`, `FR=12`, `OK=188`
* FR label sets equal, status differences `0`, output-byte differences `0`

The forced diagnostic is not part of the patch or generated source committed
by this PR.

## Checks

* `lake build` — 4806 jobs, no warnings
* `git diff --check`
* targeted baseline/candidate Spike controls above
* random-200 A/B comparison above

Closes #11191.

### DIFF
--- a/EvmAsm/Codegen/Programs/CallFrameDescend.lean
+++ b/EvmAsm/Codegen/Programs/CallFrameDescend.lean
@@ -223,18 +223,20 @@ def callFrameForwardGasFunction : String :=
       desc+88  value_nonzero  (0/1; 0 for STATICCALL / zero value)
 
     Effect (in order):
-      1. `frame_save_regs(parent_depth, parent_pc, parent_code_base)`;
-      2. `frame_depth_push` → child depth d;
-      3. save the return-context `frame_call_ctx[d]` = (parent_x12,
+      1. charge the value-transfer gas (or consume the caller's precharge flag),
+         before mutating any child-frame state;
+      2. `frame_save_regs(parent_depth, parent_pc, parent_code_base)`;
+      3. `frame_depth_push` → child depth d;
+      4. save the return-context `frame_call_ctx[d]` = (parent_x12,
          outOff_abs = parent_mem + outOff, outSize, netPopBytes) for `frame_return`;
-      4. save the parent memory/env bases in `frame_parent_bases[d]`;
-      5. `call_frame_enter(d)` → child memory/stack/env bases (+ child mem zero-init);
-      6. `call_frame_set_call_env` (ADDRESS=to, CALLER=parent.ADDRESS, CALLVALUE);
-      7. `call_frame_set_calldata` (alias child calldata into parent memory);
-      8. `call_frame_forward_gas` (EIP-150 63/64 + stipend) → child env.gasRemaining;
-      9. copy the witness context env+576..616 (header/state/codes ptrs+lens) so the
+      5. save the parent memory/env bases in `frame_parent_bases[d]`;
+      6. `call_frame_enter(d)` → child memory/stack/env bases (+ child mem zero-init);
+      7. `call_frame_set_call_env` (ADDRESS=to, CALLER=parent.ADDRESS, CALLVALUE);
+      8. `call_frame_set_calldata` (alias child calldata into parent memory);
+      9. `call_frame_forward_gas` (EIP-150 63/64 + stipend) → child env.gasRemaining;
+     10. copy the witness context env+576..616 (header/state/codes ptrs+lens) so the
          child's by-address handlers (BALANCE/EXTCODE*/the next descent) resolve;
-     10. set the child code base x21=x10=code_ptr (PC at code[0]) and
+     11. set the child code base x21=x10=code_ptr (PC at code[0]) and
          env.codeSize (env+496) = code_len.
 
     On return the live dispatcher registers are repointed to the child frame and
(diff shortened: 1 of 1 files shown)


pirapira:
**Labelled.** `build` still running; MERGEABLE, +60/−73 across 2 files, stacked on #11209 and stated as such. `lake build` 4,806 jobs, no warnings.

**The control set runs both directions, which is what makes this a fix rather than a removal:**

| control | expected | measured |
|---|---|---|
| forced OOG at the transfer charge | no frame allocated | candidate **depth 0** at charge and at `.exit_outofgas`, vs baseline **depth 1** |
| `00177` — successful call | descends normally | descended, **bytes match baseline** |
| `15904` — OOG **after** the charge | descends normally | descended, **bytes match baseline** |

⇒ `15904` is the one that matters most: **an out-of-gas occurring after the transfer charge must still descend.** A patch that simply stopped descending would have passed the forced-OOG control alone. That is the same shape as `01215` on #11209, and it was applied here without being asked for.

**Spec basis:** `system.py:397-405` charges `message_call_gas.cost` **before** `generic_call`. The guest allocated and pushed the child frame first. ⇒ Charge-then-act, matching the convention the rest of the handlers already follow.

⚠️ **And the severity is stated honestly rather than inflated.** The frame **is** reclaimed on unwind — `frame_base` is purely depth-indexed (`basr_values + depth × 0x19000`) with **no high-water pointer** — so the consumption was **transient**, not a cumulative arena leak, and there is no 1025-slot false-reject surface. ⇒ This is spec alignment, and the PR says so instead of implying a leak it had already measured away.

**Breadth:** random-200, seed 20260802, both legs scored, **FA=0, FR=12, identical FR labels, zero status differences, zero output-byte differences.** Candidate `6fa4953a…` against baseline `5cebd0ac…` — distinct builds, so the identity is not vacuous.

⚠️ **The corpus could not reach this naturally** — four fixtures were tried and `00148` had a zero-value descriptor, so `.Lcfd_charge_transfer` was never exercised. The witness required nop'ing the value-zero branch and raising the immediate, and **those edits stayed in `/tmp` and out of the PR**, with the recipe in the body. Nobody reproduces this without it.

The random gate was run before the label was requested, unprompted — the same discipline that had to be demanded on #11205 earlier and cost it an hour.

PR #11217, opened 2026-08-02
perf(codegen): retire the eager BAL-sourced recipient storage preload — 9.155 MiB of .bss

pirapira:
## Retire the eager BAL-sourced **recipient** storage preload — **9.155 MiB of `.bss`**

GH #11176, the unlanded half of #10874. #11165 landed the demand-driven `h_SLOAD`, which resolves a cold slot on first read; staging every BAL-declared recipient slot up front duplicated it. This removes the producer, its consumer arithmetic, and its buffers.

### Evidence, in the order that makes it a deletion rather than a hope

**Population first** — a sentinel build (perturb whenever `bvcd_key_count != 0`), 1,045-row stride sample: **492 rows changed category** (leg T2, re-measured on this base), so the path is *demonstrably exercised*. That is a **floor** — a row where the perturbation happened to be harmless is invisible.

**Then the A/B** — `main` @ **`e9e7dd5fa`** vs this tree, same sample:

| leg | ELF sha256 | OK | FR | FA |
|---|---|---|---|---|
| parent — `main` `e9e7dd5fa` | `34cea0719d5d6db0…` | 1004 | 41 | **0** |
| candidate — the fixpoint | `14cbbdb94a71b41c…` | 1004 | 41 | **0** |

**Forward 0, reverse 0, FA = 0 both legs.** Identical.

⚠️ **Zero coverage would not have been enough.** A null on a path nothing exercises proves nothing — the population measurement is what makes this null readable, and it was run *before* the A/B rather than after a puzzling result.

### Sections — read from the emitted ELFs, not derived from source constants

```
.text  0x0640d0 -> 0x063c84        -1,100 B
.data  0x5370   -> 0x5370          UNCHANGED  (and GuestAddrs.bnf_le_a still 0xa3000ee0)
.bss   0x1b594860 -> 0x1ac6cc60    -9,600,000 B = 9.155 MiB
```

⚠️ I first *computed* `9,600,024` from the source constants (two buffers + three 8-byte cursors). The measured figure is **9,600,000** — the cursors are absorbed by alignment padding. The source constant is the request; the section is the answer.

**Regen to fixpoint, three complete passes.** `docs/regenerating-generated-files.md` in order; **pass 3 changed nothing** (`git status` empty after a complete pass), which is the convergence check the doc asks for rather than a one-pass repin. The handwritten `.bss` literals no generator touches were updated too — `Proofs/GuestImage.lean` `0xbe6a4860 → 0xbdd7cc60` (twice) and the `0x1b594860` size literal. Full `lake build` clean (4796 jobs) with `GuestImage` and `GuestImageEntries` reported **Built, not Fetched** — a cached module cannot fail a pin it did not check.

### ⭐ What went and what stayed — the boundary where a silent regression would live

**CAPACITY checks went with their buffers.** The two `bgtu` against `bsrAccountSlotCap` existed *because* `bvcd_keys`/`bvcd_preload` were that size; with no buffer they guard nothing.

**INPUT-VALIDITY checks stayed.** The `a0 == 2` bail after `bal_find_account_by_address` is a BAL-parse reject that would still be correct if every buffer in the image vanished — so it, and `bvcd_acct_ptr` / `bvcd_acct_len`, remain.

### ⛔ The capacity deletion is an **accept-side widening**, and the A/B cannot certify it

Two sites that can **reject** today will **proceed** after the cut. A sweep of rows that already pass says nothing about that, so the reject side was enumerated:

| leg | rows changed vs main | reading |
|---|---|---|
| **W** — perturb when the count **exceeds** the cap | **0** | no sampled row exceeds `bsrAccountSlotCap` (100,000 slots) |
| **X** — same probes, condition **inverted** (`bgtu`→`bleu`) | **855** | ⭐ the arms are reachable and the perturbation *is* observable at these exact sites |

⇒ **leg W's zero is a real population zero, not a dead instrument.** Leg X was necessary because the perturbation is **position-dependent in this guest** — earlier sentinels at other sites fired 1 `FAULT`/491 `FR` and 140 `FAULT`/835 `FR`, so firing at one site does not transfer to another.

⚠️ Each site got its **own** probe label; the shared `.Ldtrc_bal_unsupported` was deliberately *not* perturbed, since that would have counted the retained parse-validity bail and inflated the answer.

⚠️ **And the zero is a statement about 1,045 sampled rows, not the corpus.** A block with more than 100,000 recipient storage slots would take the bail. Note also that a capacity bail is a fail-closed on *our* limit rather than a spec verdict, so a row that rejected only for capacity was arguably always a false reject — a nonzero answer would have *relabelled* those rows, not blocked this change.

### Fallthrough

Every label in the deleted range — `.Ldtrc_read_storage`, `.Ldtrc_sloop`, `.Ldtrc_zero_storage`, `.Ldtrc_header_storage`, `.Ldtrc_kcopy`/`kdone`, `.Ldtrc_vcopy`/`vzero`/`vzloop`/`vdone`, the `.Ldtrc_4788_*` family, `.Ldtrc_thread_hit`, `.Ldtrc_nothread`, `.Ldtrc_recipkey_rev` — is branched to **only from inside it**. The strongest form: no predecessor to strand.

### R8 / effective-address audit — because a name census cannot clear a data buffer

Layout: `bvcd_sc_count`, `bvcd_i`, **`bvcd_keys` @ `0xaa0c8c90` (3,200,000 B)**, **`bvcd_preload` @ `0xaa3d6090` (6,400,000 B)**, `bscv_vptr`.

Every `la` of a layout **neighbour**, with the offsets applied to the loaded register:

| neighbour | `la` sites | offsets | could reach |
|---|---|---|---|
| `bvcd_i` (−8) | 10 | **all 0** | `+8` would hit `bvcd_keys` |
| `bvcd_sc_count` (−16) | 3 | **all 0** | `+16` would hit it |
| `bscv_vptr` (+6.4 MB) | 2 | **all 0** | a negative offset would reach into `bvcd_preload` |

No `la`-plus-displacement reaches either buffer, and no register-computed add appears at any neighbour site.

⚠️ **Stated limit:** the window is the `la` plus three lines. **A store-and-reload flow is not excluded.** I judge that unlikely for 8-byte scalars but I have not proven it.

### Byte-neutrality

**Not applicable** — this is a deletion, so `.text` and `.bss` change by design and there is no identity to make non-vacuous.

### Also fixed: a comment the cut would have stranded

`dtrc_slotkey_le` described itself as *"the LE byte-reverse of `bvcd_keys[i]`"* — a buffer that no longer exists. Leaving it would have created a fresh instance of exactly the stale-prose defect this issue's other PRs are correcting.

### Not retired here — each for a measured reason, recorded on #11176

* **the CALLEE preload** (`csce_*`, 3.052 MiB): `seed_callee_storage` **also** fills `callee_balance_table`, which `call_frame_descend` reads to stage a child frame's `env+32` for nested `SELFBALANCE`. `call_frame_descend` **zeroes** `env+32` before its table scan, so retiring the producer would make nested `SELFBALANCE` return **0**. Needs a demand-driven `h_SELFBALANCE` first.
* **the SYSTEM-CALL preload** (`sps_*` + `c1_preload`, 9.537 MiB): disabling it **reverse-flips three consolidation-request rows** (`23200`, `23300`, `23350`). Load-bearing.
* **`c1_staging`** (6.619 MiB): a **different channel** (`a3` to a code-derive call). Never measured, and deliberately not folded into a `c1` total.

⚠️ Also **not** claimed dead: the `dtrc_*` scratch set (`dtrc_recipkey`, `dtrc_slotkey_le`, `dtrc_threadval`, `dtrc_hdr_ptr`, `dtrc_hdr_len`) still has code references outside the deleted range.

---

### ⚠️ Every number above was RE-MEASURED after the pin regen — the first version of this body was wrong

`check-region-map` caught pin drift on the first push, and the fix is not cosmetic: **`RegionMap` pins are an emission INPUT, consumed rather than verified.** The ELF I had measured was emitted with `textSizeBytes = 0x0640d0` against a real `.text` of `0x063c84`, so every figure downstream of it described an image that will never be built. On top of that, **`main` had moved two batches** under the branch (`#11208`, `#11213`, `#11216`).

⇒ so the whole measurement set was re-run on the merge base. **The verdicts did not change; the numbers did:**

| figure | first published | re-measured |
|---|---|---|
| parent / candidate | OK 999, FR 46 | **OK 1004, FR 41** |
| forward / reverse / FA | 0 / 0 / 0 | **0 / 0 / 0** (unchanged) |
| bvcd population floor | 492 | **492** (unchanged) |
| capacity-bail population | 0 | **0** (unchanged) |
| capacity-probe control | 850 | **855** |

A five-row baseline discrepancy would have been published as this change's baseline. **Re-measure, don't re-cite** — and the concrete cost of not doing so is in that table.


pirapira:
**Reviewed — label on green build.** The evidence here is more complete than the deletion needed, and the extra work is the part worth keeping.

⭐⭐ **The accept-side discharge is the strongest form I have seen on this repo.** Deleting the two `bgtu` capacity bails *widens what is accepted* — rows that reject today would proceed — and a random sweep structurally cannot certify that, because it only shows that rows which already passed still pass. **A random sample cannot certify FA=0 for a change that moves the boundary defining FA.**

⇒ The discharge is the **pair**, and neither half works alone:

| leg | probe | result |
|---|---|---|
| W | capacity sites made distinctive | **0 of 1,045** rows change |
| X | same sites, `bgtu` → `bleu` (fires on every row reaching them) | **850 of 1,045** change |

**850 rows reach the checks; zero exceed the cap.** Leg W alone was uninterpretable — no row exceeds `bsrAccountSlotCap`, or the probe is dead at those two sites, and *the numbers are identical in both worlds*. Leg X separates them. ⇒ **A zero from a sentinel is the one result that cannot interpret itself, and inverting the condition tests the instrument rather than the hypothesis.**

⚠️ **And carrying an earlier leg's firing across would have been wrong here** — measured, not assumed: the same perturbation gave **1 fault on leg T and 140 on leg U** at different insertion points. A probe validated elsewhere is not a probe validated here.

**The R8 exclusion is by measurement, not by name census**: every `la` of a layout neighbour (`bvcd_i` ×10, `bvcd_sc_count` ×3, `bscv_vptr` ×2) carries offset 0, and no register-computed add appears at any neighbour site. That is the right predicate — *zero name references does not prove a data buffer dead when neighbours are reached by `la`+displacement*. The stated window (`la` + three lines, store-and-reload not excluded) is a **named limit, and it is worth more than an unnamed completeness claim.**

**Capacity vs validity is the line to keep**: the two `bgtu` bails exist *because* the buffers are that size and guard nothing once they are gone; the `a0 == 2` parse-validity bail stays, as do `bvcd_acct_ptr` / `bvcd_acct_len`. ⇒ And the framing survives the result — **a capacity bail is fail-closed on our own limit, not a spec verdict**, so a nonzero answer would have *relabelled* those rows as always-false-rejects rather than blocking the deletion. The zero made the question moot rather than won it.

Fallthrough is discharged in the strong form (every label in the range is branched to only from inside it). A/B is identical at **OK 999 / FR 46 / FA 0 both legs**, and the `dtrc_*` scratch set was correctly *not* claimed.

⚠️ Two hazards this PR dodged that others have not: the `.bss` figure was **measured at 9,600,000, not the derived 9,600,024** (three cursors absorbed by alignment), and a verifying build printed **`Fetched`, not built** — a green 1,261-job run that proved nothing about the edit.

pirapira:
**Labelled — `build` COMPLETED SUCCESS on the re-measured fixpoint.** CONFLICTING is expected (#11215 merged at 18:45); that resolution is mergebot2's, not the author's.

⭐⭐ **Every figure in this body was re-measured on the merge base rather than carried over, and that was not a formality — two of them changed.** The published numbers were OK 999 / FR 46; the correct ones are **OK 1004 / FR 41**. The base had moved two batches underneath, so the original pair described a tree that no longer exists, and a **five-row baseline discrepancy would have been published as this change's baseline.**

⇒ The reason re-measuring was mandatory rather than tidy: **`RegionMap` pins are an emission *input*, not a record of one.** They are consumed, never verified. The first sweep ran on an ELF emitted with `textSizeBytes 640d0` against a real text of `63c84` — a confident, self-consistent artefact that is simply not the image being merged. ⇒ *A wrong pin cannot be caught downstream, because everything downstream inherits it silently.*

⭐ **The `492` population floor was re-run despite being expected to hold** — and that is the judgement worth copying. It came back identical, which is now *known* rather than assumed. One carried-over number in an otherwise re-measured body is indistinguishable, to a reader, from a fully re-measured one.

**The accept-side discharge survived the rebase intact:**

| leg | probe | result |
|---|---|---|
| W2 | capacity sites made distinctive | **0 of 1,045** |
| X2 | same sites, `bgtu` → `bleu` | fires |

**Rows reach the capacity checks; none exceeds the cap.** Leg W alone was uninterpretable — no row exceeds `bsrAccountSlotCap`, or the probe is dead there, and *the numbers are identical in both worlds.* Inverting the condition tests the instrument rather than the hypothesis. ⚠️ And carrying an earlier leg's firing across would have been invalid: the same perturbation gave **1 fault on leg T and 140 on leg U** at different insertion points.

**Fixpoint discipline**: three complete passes with **pass 3 changing nothing** and `git status` empty — the convergence check the doc asks for, not a one-pass repin. `.data` held at `0x5370`. Handwritten `Proofs/GuestImage.lean` literals updated, since no generator touches them. And **`GuestImage` BUILT in 2.9s / `GuestImageEntries` BUILT in 1.6s — not `Fetched`**, so the pins were genuinely re-elaborated.

**R8 excluded by measurement**, not by name census, with the audit window stated as a named limit.

PR #11218, opened 2026-08-02
fix(codegen): drop ALL six system skips from MtxTail (#10684)

pirapira:
## Summary

**Union of #11210 + #11218 (coord complicated-merge call): empty the multi-tx system skip list.**

Drop **all six** whole-account system skips from `bv_mtx_skip_list`:
- EIP-2935, EIP-4788 (BAI-0) — already dropped on main via #11210
- EIP-7002, EIP-7251, EIP-6110, SYSTEM_ADDRESS — this branch's original four

**Remaining skip entries:** gas/value-coupled `{sender_i, recipient_i}` + coinbase only.  
**count = 2N+1** (`bvMtxSystemSkipEntries := 0`).

### Why union, not either half

Both changes rest on the same rule: a whole-account skip with **no execution-specs counterpart is a false-accept surface**. That argument does not distinguish among the six. Keeping either half was scope discipline at different bases, not a decision to retain them.

### Resolution (complicated-merge)

mergebot2 handed back because main (#11210 keeps four deferred) and this branch (keeps BAI-0 pair) made **directly contradictory** edits to the same skip fill. Resolved by applying the rule twice → empty system block.

Hand-written files:
- `BlockVerdictParams.lean` — `bvMtxSystemSkipEntries = 0`; guards 19047 / 609504
- `BlockVerdictMtxTail.lean` — remove `.Lbv_skl_sys_*` copy loop; count `2N+1`
- `BlockVerdictDataSectionTail.lean` — comments + `i3djw` size tracks 0 system
- `BlockVerdictEoaBodyEffectReconcile.lean` — **accept main deletion** (dead twin #11212; not a skip-list conflict)

Generated layout: took main pins; **did not hand-regen** (batch owns `.text` pins).

### Emitted confirmation

After coinbase fill:
```
slli t3, t2, 1; addi t3, t3, 1; … sd t3, bv_mtx_skip_count
```
No `bbcv_sys_*` load in the skip builder. (`bbcv_sys_*` data labels remain for other consumers e.g. code-preimage sys checks.)

## A/B (REQUIRED on resolved empty-skip tree)

Parent = main `2499b52d1` guest `b6c67a65055d703e…` (still skips four deferred).  
Cand = this PR guest `24607d49a3ae8816…` (system skip empty).  
Manifest `/tmp/11167-random200.tsv` seed42. ELFs **DISTINCT**.

| leg | FA | FR | OK |
|---|---:|---:|---:|
| parent (main, 4 system skips) | 0 | 5 | 195 |
| cand (0 system skips) | 0 | 5 | 195 |

- FR sets **EQUAL** (same 5 labels)
- reverse flips **0**
- guest_succ diffs **0**
- **FR did not climb** when more rows are checked (accept-side narrowing)

FR labels unchanged: 11738 selfdestruct_created_same_block, 19037 random_statetest517, 23106 withdrawal multi-block fee, 23574 type_4 single_au, 24109 multi_type_requests.

Artifacts: `/tmp/grok-11218-empty-skip/`.

## Risk direction

Removing skips **narrows** accept (more accounts compared). That is the correct direction for FA. Prior poison sentinels on each subset had OK→FR floor 0 on stride-1045 + r200; this A/B confirms the full empty list does not create new FR on the same sample.

## Related

- Closes the #10684 skip-list half (system address exclusion)
- Supersedes half-measures in #11210 / original #11218 scope
- Does **not** claim fc37 / side-log / one-container convergence


pirapira:
**Reviewed — label on green build.** The measurement is careful and the justification is correctly *not* the measurement.

⭐ **The instrument was validated rather than assumed, which is where the last four false greens came from.** A **count-preserving data poison** (`0xEE`×20 on the `.data` `bbcv_sys_*` rows) rather than address 0 — and that choice is measured, not stylistic: address 0 is position-dependent in this guest, and `li t4 6→0` produced **936 faults**, which is a broken probe, not a clean sentinel. ⇒ The spike break at `0x800193b4` showing parent `entry5 = 7002` against poisoned `entry5 = eeee` is a **live proof the poison reached the consumer**, not merely the image. A probe that changes the artefact but never reaches the reader measures nothing.

⇒ **The justification is spec absence, and that is the right footing.** FR 0 across stride-1045 and random-200 is a *floor*, not a proof of deadness — the sample does not exercise these addresses. Stating that plainly beats a completeness claim nobody can check.

⭐⭐ **The risk direction is what makes this land**: a skip means *we do not check*. A hardcoded skip list with **no execution-specs counterpart** is a false-accept surface — it can hide a real mismatch and it structurally cannot cause a false reject. ⇒ So **keeping these risks FA; deleting them risks only FR** — and FR churn is tolerated in both directions during convergence while **FA=0 never relaxes**. A hardcoded address list with no spec counterpart is the shape that never converges on its own.

A/B holds that out: **FA 0, FR 5 equal, reverse 0**, `.bss` −256 B, `check-region-map` green.

**Scope is exactly right** — the four deferred only. 2935/4788 stay, since codex2's #11183 re-derivation moved them out of the Class-A execution-fed set into Class-B comparison rows; folding them in would have mixed two different claims into one cut.

Correct not to self-label, and the hands-off calls on #11215/#11210 were right.

pirapira:

CONFLICTING against current main is expected: #11216's batch (#11210 + #11212 + #11214) landed after this branch was cut, and #11210 touches the same MtxTail skip region. ⇒ **The conflict is mergebot2's to resolve** — it owns the branch including conflicts and regen once labelled. No rebase needed from the author, and re-running the A/B legs is *not* wanted here; CI builds the merge ref.

Recording what makes this deletion sound, since the measurement alone does not:

⭐ **The risk asymmetry is the argument.** A skip means *we do not check*. A hardcoded skip list with **no execution-specs counterpart** is a false-accept surface — it can conceal a real mismatch and it structurally cannot produce a false reject. ⇒ So keeping it risks **FA**; deleting it risks only **FR** — and FR churn is tolerated in both directions during convergence while **FA=0 never relaxes**. FR 5 equal and reverse 0 says the FR cost is nil on the measured population anyway.

⚠️ **The FR 0 result is a floor, not a proof of deadness**, and the body says so — the sample does not exercise these addresses. Justifying on **spec absence** rather than measured gain is what makes that honest: a named limit beats an unnamed completeness claim.

⭐ **The instrument was validated, not assumed** — which separates this from four confident greens that turned out worthless today. A **count-preserving data poison** was chosen over address 0 *because* address 0 is position-dependent in this guest, and `li t4 6→0` produced **936 faults** — a broken probe, not a clean sentinel. The spike break at `0x800193b4` showing parent `entry5 = 7002` against poisoned `entry5 = eeee` proves the poison reached **the consumer**, not merely the image. A probe that alters the artefact but never reaches its reader measures nothing.

**Scope held**: four deferred skips only. 2935/4788 stay, since #11183's re-derivation moved them into Class-B comparison rows — folding them in would have merged two distinct claims into one cut.

pirapira:
Excluded from automated merging: this branch and main now make **directly contradictory** changes to the same multi-tx system skip list, so resolving it is a spec-correctness decision rather than a mechanical conflict repair.

Conflicts are in four hand-written source files (not the generated layout set):
`BlockVerdictParams.lean`, `BlockVerdictMtxTail.lean`, `BlockVerdictDataSectionTail.lean`, `BlockVerdictEoaBodyEffectReconcile.lean`.

**The contradiction**, from `BlockVerdictParams.lean`:

| | system skip entries |
|---|---|
| this branch (`#10684`, "drop four deferred system skips") | drops 7002 / 7251 / 6110 / SYSTEM_ADDRESS, **keeps** EIP-2935 + EIP-4788 |
| main (via #11210, already merged) | **keeps** the four residual 7002 / 7251 / 6110 / SYSTEM_ADDRESS, drops EIP-2935 + EIP-4788 |

main's landed text is explicit about it:
```
System entries (GH #10684): four residual addresses only —
EIP-7002 / EIP-7251 / EIP-6110 predeploys + SYSTEM_ADDRESS. EIP-2935 and
EIP-4788 are **not** skipped (no spec address exclusion; in-body modeled
skip already removed by f5be2e440).
def bvMtxSystemSkipEntries : Nat := 4
```
while this branch's `BlockVerdictMtxTail.lean` comment says the opposite:
```
skip[2N+1..2N+2] = EIP-2935 + EIP-4788 only (#10684 dropped 7002/7251/6110/SYSTEM)
```

Both cite #10684, so this looks like two passes at the same bead reaching opposite conclusions about which addresses have a spec exclusion. The `#guard bvMtxSkipListEntries` / `bvMtxSkipListBytes` pins differ accordingly.

Picking a side changes which accounts the BAL comparators skip, so I have not guessed. Most likely this branch is simply superseded by #11210 — if so, closing it is the cleanest outcome. Still carries `merge!`; I'll pick it up as soon as it is resolved or closed.

pirapira:
## Pin fixpoint after CI drift + main #11217

Union empty-skip removes more code than either branch alone → pins moved again.

**Procedure:** merge origin/main `10b89eaf3`, take main on generated layout files, full regen to fixpoint (3 passes; pass3 changed nothing). **Never hand-set** — pins from `readelf -SW` on that ELF.

| pin | value |
|---|---|
| textSizeBytes | `0x06338c` |
| dataSizeBytes | `0x5370` |
| bssSizeBytes | `0x1ac6cb20` |
| GuestImage .bss end | `0xbdd7cb20` |

- `check-region-map` green
- `GuestImage` **Built** (`LAKE_ARTIFACT_CACHE=false`)
- empty skip confirmed: emitted count `2N+1` (`slli t3,t2,1; addi t3,t3,1`); no `Lbv_skl_sys_*`
- guest sha256 `0681ab5f6f21e293ffd65a15f0e3cba58e2cd27e3d6b1350bfb2d0eac1cd833d`

### A/B re-derived on fixpoint image (do not re-cite pre-pin numbers)

Manifest `/tmp/11167-random200.tsv` seed42; spike SPIKE_WATCH-capable binary.

| leg | guest sha | FA | FR | OK |
|---|---|---:|---:|---:|
| parent main `10b89eaf3` | `5ce37655…` | 0 | 5 | 195 |
| cand fixpoint | `0681ab5f…` | 0 | 5 | 195 |

ELFs **DISTINCT**. FR sets **EQUAL**. guest_succ diffs **0**. reverse **0**. merge! kept.

pirapira:
landed via #11225.

Verified: `BlockVerdictParams.lean`, `BlockVerdictMtxTail.lean` and `BlockVerdictDataSectionTail.lean` are byte-identical between this branch and main (sha256 match on all three), including `def bvMtxSystemSkipEntries : Nat := 0`. The three commits still on the branch are its own layout repins (`chore(layout): repin RegionMap text/bss`, the main merge, and the fixpoint regen), which are superseded by the repin done in #11225 — `.text = 0x06338c`, `.bss = 0x1ac6cb20`.

Nothing functional is left unmerged, so closing rather than leaving it to drift. Reopen if I've misread it.

PR #11219, opened 2026-08-02
docs: justify account-write undo reservation

pirapira:
## Summary
- document the 4,294-row current workload derivation for account-write undo pushes
- retain the 16,384-row / 2 MiB reservation with about 3.8x intentional headroom
- record the producer census scope and existing fail-closed overflow behavior

## Verification
- `lake build`
- `git diff --check`

No runtime, address, ELF, or generated-file changes.

### DIFF
--- a/EvmAsm/Codegen/Programs/AccountWriteMap.lean
+++ b/EvmAsm/Codegen/Programs/AccountWriteMap.lean
@@ -98,7 +98,11 @@ import EvmAsm.Stateless.MemoryLayout
 namespace EvmAsm.Codegen
 
 /-- Transaction-local entries. One transaction's CALL-tree bound is 15038, so
-    the existing 16384-row reservation remains sufficient. -/
+    the existing 16384-row map reservation remains sufficient. This is an
+    occupancy bound, not an undo-flow bound: a hit updates one map row while
+    still pushing a fresh rollback record. The undo-flow workload derivation
+    needs 4294 rows on the densest current path, so the 16384-row reservation
+    is intentionally retained with roughly 3.8x headroom. -/
 def txAccountWritesCapacity : Nat := 16384
 
 /-- Block-lifetime entries. Amsterdam permits at most 9523 minimum-cost plain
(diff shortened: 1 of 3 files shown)


pirapira:
**Reviewed — label on green build.** Docs-only, `+42/−13`, and it answers the question #11186 was actually asking.

⭐ **The load-bearing insight is the cap that binds.** The undo journal **resets per transaction**, so the applicable limit is the **16,777,216 regular-gas transaction cap** (`transactions.py:63`, `fork.py:1096-1100`), not the 200M block limit. ⇒ That distinction is **true of state gas and false of regular gas**, and getting it backwards produced a 12× over-reservation once already — 76.8 MiB reserved, then physically zeroed by dispatch frames at depth 221+ before its readers ever ran. Applied in the right direction here.

⭐ **Occupancy is not flow, and the docs now say so.** The existing `txAccountWritesCapacity = 16384` bounds **live rows**; it cannot bound **repeated-overwrite traffic** through them. *A cap on a container is not a cap on the traffic through it* — that has been the answer three separate times tonight (write-map rows vs writes, exec-log slots vs writes, `bgeu` in `storage_writes_undo_push`).

**The derivation is checkable rather than asserted**: 4,288 = 2 × ⌊(16,777,216 − 12,000) / 7,816⌋ auth pushes, plus 6 fixed boundary/self-transfer records ⇒ **4,294 needed against 16,384 provisioned, ~3.8× headroom**, with EIP-7702-in-MTx named as the densest flow.

⇒ **Right call to back out the shrink.** Cutting 16,384 → ~4,294 buys ~1.5 MiB — noise against the 244 MiB exec log — and pays for it with a **tightened reject boundary across the whole corpus**, resting entirely on the producer census being closed. A missing push site would be wrong in the dangerous direction and **would not show up on a random sample, because the sample does not contain the densest flow.** Documentation win and boundary move should not travel together.

⭐ And **naming the census as non-formal closure is what makes the headroom honest** — it converts 3.8× from decorative slack into slack that is explicitly doing work. A named limit beats an unnamed completeness claim; that has held on three PRs today.

⇒ This also retires the self-justifying-number problem: `16384 × 128` is *exactly* the region extent `0xa2d20000..0xa2f20000`, so the cap previously described its container rather than its workload. It now has a derivation behind it.

PR #11220, opened 2026-08-02
chore(codegen): delete pure-dead single-tx twins (#10685 PR1)

pirapira:
## Summary

**#10685 PR1 — pure dead only.** One predicate, three clusters, one A/B.

**Predicate (emitted call graph on main `2499b52d1` guest `b6c67a65…`):** zero inbound `jal`/`j`/branch targets **and** no fallthrough (previous primary ends `ret` or `jalr x0, 0(x1)`).

### Deleted

| item | size | fallthrough discharge (weaker form — name the predecessor) |
|---|---:|---|
| `i3djw_skip_list` BSS | 288 B | n/a data; **0** `la`/`jal` after #11212 removed builder |
| `simple_transfer_tx_context` | 672 B | prev primary **`tx_gas_sender_bal_lookup`** ends **`jalr x0, 0(x1)`** |
| `simple_transfer_recipient_bal_verify` | 940 B | prev **`tx_gas_bal_post_verify`** ends **`ret`** |
| `simple_transfer_fee_recipient_bal_verify` | 968 B | prev was recipient_bal_verify (also deleted); chain root **`tx_gas_bal_post_verify` `ret`** |
| `multi_tx_sequential_gas_settle_step` | 196 B | prev **`multi_tx_running_sender_balance_step`** ends **`ret`** |
| BSS `bv_simple_transfer_recipient` / `_fee_recipient` | 208+240 B | only served dead bal_verifys |

**TEXT cuts kept.** BSS cut kept.

### KEEP (do not touch — inventory 3/5/7 + PR2)

- `simple_transfer_intrinsic_gas` + precompile/publish path (live MTx empty-code)
- `bv_emit_single_tx_tl7708` (live jal, dead input — **PR2**)
- `bv_mtx_skip_list`, `multi_tx_running_sender_balance_step`
- `block_verdict_mtx_oog_materialize` (hook via `la`, not jal)
- bmvmx flag stores (#11211 — prefix `bv_exec_p` live)

### Probe-only

Function strings kept for zisk probes. `simpleTransferTxContext_prog` uses local address placeholders (`simpleTransferTxContextPc := 0x80000000` + TSV-pinned data); drift via `*_eq_prog`.

### A/B random-200 seed42 (pre-main-merge baseline)

| leg | sha | FA/FR/OK |
| parent `2499b52d1` | `b6c67a65…` | 0/5/195 |
| cand | `e030366b…` | 0/5/195 |

FR sets equal; guest_succ flips **0**; reverse **0**; ELFs **DISTINCT**.
(Coord: no A/B required for merge/regen alone.)

### CI repair (merge main + dump drop)

**`check-build-units-link` FAIL root cause (coord-confirmed diagnosis):**

- Unit `zisk_stateless_verdict_v2` failed with **only** undefined refs:
  - `bv_simple_transfer_recipient`
  - `bv_simple_transfer_fee_recipient`
- TEXT cuts (`simple_transfer_tx_context`, recipient/fee bal_verify, `multi_tx_sequential_gas_settle_step`, `i3djw_skip_list`) were **not** needed by v2.
- The sole live refs were **`la` in the v2 debug dump after `_start`** (`BlockVerdict.lean`, OUTPUT+360/+368) — sibling-unit trap: dead in guest, still mirrored as dump targets.

**Fix:** drop those two dump `la` lines; **keep the BSS+text cut**. Comment cites #10685. Slots 360/368 left unwritten; later dumps keep historical offsets.

**Dump consumer:** `scripts/codegen-zisk-stateless-verdict-debug-smoke.sh` only checks that the formatter reaches the final word — it does **not** parse recipient dump slots. No eest harness parses OUTPUT+360/+368. Dump previously printed bal_verify result BSS; nothing observes those two slots.

**STALE allowlist:** CI footer always prints the STALE boilerplate on any fail. Local + CI log: **no unit named STALE**; the three KNOWN entries remain non-linkable. **Cannot reproduce a real stale allowlist entry from this failure.**

**After merge `origin/main` `10b89eaf3` + layout regen fixpoint:**

| pin | value |
|---|---|
| textSizeBytes | `0x0628f8` |
| dataSizeBytes | `0x5370` |
| bssSizeBytes | `0x1ac6c980` |
| GuestImage bss end | `0xbdd7c980` |
| guest sha256 | `7c03a0e9265bddddfbc28f8bbd0fcf667ae3e08bade13fb345f3024b39d5adea` |

- `check-region-map` GREEN
- `check-build-units-link` OK including **`zisk_stateless_verdict_v2`**
- Deleted symbols ABSENT from nm; KEEP (`simple_transfer_intrinsic_gas`, `multi_tx_running_sender_balance_step`, `bv_mtx_skip_*`) present

### Out of scope

PR2 TL7708 no-op (SPIKE_WATCH zero-store on `bv_simple_transfer_tx`; multi-unit check first). No ctx-pointer unify. No `merge!` from grok.

Artifacts: `/tmp/grok-10685-pr1/`, `/tmp/grok-11220-units-link.log`.


pirapira:
**Reviewed — label on green build.** One predicate, three clusters, one A/B is the right structure: the deletions share a justification, so the body states it once rather than three times.

⭐ **The predicate is the emitted call graph, not names** — zero inbound `jal`/`j`/branch targets **and** no fallthrough. That distinction has cost this repo six times, and this PR is the direct beneficiary of the seventh: `bv_emit_single_tx_tl7708` was *correctly held back* to PR2 despite reading `single_tx` in its name, because it has a **live `jal`** from the simple-transfer publish path. ⇒ **A name is not a contract, and neither is a plausible-looking family resemblance.**

**Fallthrough is discharged in the weaker-but-valid form, with each predecessor named** — `tx_gas_sender_bal_lookup` ending `jalr x0, 0(x1)`, `tx_gas_bal_post_verify` ending `ret`. That is the right call: the strong form (#11212's — labels absent from the image entirely) isn't available when deleting *reachable-looking* text, and the weak form is sound **only if the predecessor is correctly identified**, which is why naming it in the table matters.

⚠️ **The chained case deserves its explicit treatment**: `simple_transfer_fee_recipient_bal_verify`'s predecessor is `simple_transfer_recipient_bal_verify`, itself deleted — so the discharge correctly walks to the **chain root** (`tx_gas_bal_post_verify`, `ret`) rather than stopping at a symbol that is about to vanish. A discharge against a deleted predecessor would be circular.

**A/B: FA 0, FR 5 equal, reverse 0**, `.text` −2776 B, `.bss` −736 B.

⇒ Three things were correctly **excluded**, each for a different reason, and that is what makes the scope credible rather than convenient:

- **`simple_transfer` precompile/publish path** — a live MTx empty-code lane; the maintainer's precompile caveat applies.
- **BmvMx** — #11211; its `bv_exec_p` prefix initialisation **is** consumed, so a wholesale block deletion would take a live initialiser out with the unread flags.
- **`oog_materialize`** — reached by `la`, **not** `jal`. ⇒ *Exactly the pointer handover a call-graph census misses*, and it was flagged rather than swept up.

**Register note accepted as scope evidence**: MTx uses the `bv_mtx_ctx` global rather than `x20`-as-env, and `emit_single` saves an `x20` it never uses. **A pair that needs its registers rewritten is evidence the two sites are not trivially shareable** — so no unification attempt here, correctly deferred to a ctx-pointer routine if ever.

pirapira:
## CI repair landed

Head `29492dd12` = merge main `10b89eaf3` + dump-la drop + layout regen fixpoint.

- **Root cause:** v2 debug dump `la` of deleted BSS only (not TEXT cuts). Dropped dump lines; kept cut.
- **Dump consumer:** debug-smoke checks final word only; nothing reads OUTPUT+360/+368.
- **STALE:** footer boilerplate only; no real stale unit; three KNOWN still nonlinkable.
- `check-build-units-link` local **OK** incl. `zisk_stateless_verdict_v2`.
- Pins: text `0x0628f8`, bss `0x1ac6c980`, end `0xbdd7c980`. Guest sha `7c03a0e9…`.
- No `merge!` from grok.

Issue #11221, opened 2026-08-02
RegionMap: make the link-authority inventory total and durable


pirapira:
## Context

The 2026-08-02 RegionMap audit found that the emitted guest and the checked map had drifted in ways the existing `scripts/check-region-map.sh` did not detect:

- three link-dependent absolute pins in `EvmAsm/Codegen/RegionMap.lean` were stale (`callFrameArenaBase`, the memory-pool base, and `syslogBase`);
- the map listed `baap_storage_delete_paths`, which is present only in a probe/historical layout and is not a symbol in the current stateless_guest ELF;
- the guard checked relative extents and a selected symbol subset, but not the absolute pins or the closed union-child inventory;
- stale fail-open storage-undo prose and old 228/244 MiB call-frame comments obscured the current fail-closed implementation and 100.1 MiB frame geometry.

The current patch corrects the live pins/inventory and adds a guard for the three absolute pins plus every current union child. It also records the temporary perturbation tests: changing `callFrameArenaBase` makes the guard fail, and adding the phantom child makes the closed-inventory check fail.

## Audit classification

- MATCHES: the three ELF section sizes, the nine emitted-reality regions in `guestRegionMap`, and the current five-child union addresses/relative offsets.
- DRIFTED and corrected: the three link-dependent absolute pins above.
- PHANTOM and removed from the current map: `baap_storage_delete_paths`; search shows it remains in a probe emitter and historical docs, so those references need an explicit lifecycle decision rather than silent deletion.
- ASPIRATIONAL BY DESIGN: all 23 `schemeAAnchors`. Thirteen have absolute references in the current emitted assembly, but they are intentionally a separate port-contract list and are not part of `guestRegionMap`'s emitted-reality theorem.

## Durable follow-up

Decide whether RegionMap should be generated from the linked ELF/TSV, or whether the source map plus a total checked inventory remains the authority. Whichever model is chosen must:

1. cover every link-dependent absolute pin;
2. reject phantom map children and missing emitted symbols;
3. keep Scheme-A aspirational anchors visibly separate from the emitted guest;
4. preserve a reproducible drift test when a symbol moves or a child is added/removed;
5. document the probe-only/historical lifecycle of `baap_storage_delete_paths`.

Related findings: #11187 (stale exec-log sizing premise), #11189/#11200 (storage undo fail-closed path), and the RegionMap drift work in the current PR.

pirapira:
[...]

PR #11224, opened 2026-08-02
Audit RegionMap pins and phantom union entries

pirapira:
## Headline

Before this change, a kernel-checked `CallFramePhase` proof asserted a six-child union that included `baap_storage_delete_paths`, a region absent from the emitted `stateless_guest` image. The Lean constant existed, so the proof typechecked while describing nonexistent storage. Removing the phantom repaired the full phase-tiling theorem family to the five real children; no load-bearing theorem was silently weakened.

## Changes

- Re-pin the current-main ELF facts in `RegionMap.lean`: call-frame arena, memory-pool, syslog, section sizes, and current five-child offsets.
- Repair `CallFrameLayout` and `CallFramePhase` from stale six/seven-child and 228/244 MiB descriptions to the current five-child, 100.1 MiB geometry.
- Correct stale fail-open storage-undo prose to match the emitted fail-closed `.Lswup_fail` path.
- Extend `check-region-map.sh` to reject stale absolute pins, phantom/missing union symbols, and any generated `GuestAddrs.lean` name absent from the freshly linked ELF.
- Correct `gen-symbol-addresses.py`'s `.bss`/`.sszscratch` pins and add all six write/undo Scheme-A anchors. Regeneration produced no TSV diff because the committed table was already ELF-derived.
- Document predicate readiness: entry shape, exact capacity×stride, count semantics, invariants, lifetime/reset, aliasing, and the rule that predicates quantify `arena_base` rather than baking an image-specific address. The current linked image's `baap_storage_values` entry is explicitly 6,400,000 bytes; other build units remain separate.

## Validation

- `lake build` — 4,796 jobs
- `NO_BUILD=1 scripts/check-region-map.sh`
- `scripts/check-memorylayout-region-coverage.sh`
- guard perturbations: stale arena pin and phantom child both produce `DRIFT`/exit 1
- current ELF sweep: all five union names and all 1,091 generated `GuestAddrs` names exist

Refs #11221. The original fail-open claim in #11189 was closed separately after verifying the current fail-closed source; identity-skip and cross-unit capacity questions remain separate.

### DIFF
--- a/EvmAsm/Codegen/CallFrameLayout.lean
+++ b/EvmAsm/Codegen/CallFrameLayout.lean
@@ -136,33 +136,27 @@ def sszScratchBase : Nat := 0xbf980000
 /-- Total bytes the pre-allocated frame array occupies. -/
 def frameArrayBytes : Nat := frameSlotCount * frameStride
 
-/-! ## Placement: the frame arena UNIONS the BAL-replay basr pair (a1vvy, 200M)
+/-! ## Placement: the frame arena coalesces five BAL-replay children
 
-    ziskemu's RAM is 512 MiB (`0xa0000000..0xc0000000`). Under the former
-    1G-sized BAL arenas (`bsrMaxBalItems = 500000`, ~416 MiB) there was no free
-    window, so the frame arena *aliased* the contiguous, execution-dead
-    `basr_values`+`basr_accounts` pair (the #8513 union, 244 MiB ≥ 164 MiB).
-    The Amsterdam target is 200M block gas (`bsrMaxBalItems = 100000`), which
-    shrinks the BAL arenas to ~83 MiB; the union was retired (standalone arena)
-    while the BAL downsize left free RAM. The 200M log/receipt capacity lifts
-    (`vv4hr.3.4.*`) have since consumed that slack (measured `.data` headroom
-    fell to ~59 MiB), so a1vvy REINSTATES the union to reclaim ~49 MiB.
+    ziskemu's RAM is 512 MiB (`0xa0000000..0xc0000000`). The current Amsterdam
+    200M layout (`bsrMaxBalItems = 100000`) coalesces the execution-dead
+    `basr_values`/`basr_accounts` pair and the three `baap_storage_*` arrays
+    into the front of `call_frame_arena`. The current frame array is
+    `1025 * 0x19000 = 104,960,000 B` (about 100.1 MiB), larger than the
+    coalesced child prefix (about 61.6 MiB), with a trailing pad. The
+    `bv_system_storage_log` is deliberately standalone because it is read
+    post-dispatch; `evm_memory_pool` follows the frame arena.
 
-    The size relation flipped: the frame array (`frameArrayBytes` ~228 MiB at the
-    reconciled `0x39000` stride) is now
-    LARGER than the basr pair (~49 MiB), so the pair is coalesced into the FRONT
-    of `call_frame_arena` (`BlockVerdictDataSection.lean`) rather than the arena
-    aliasing into the pair. Soundness is the same execution-dead disjointness
-    #8513 established and is now load-bearing again: `basr_values`/`basr_accounts`
-    are referenced ONLY in `BlockVerdictStateRoot` (Phase H,
-    pre-dispatch state-root recompute) with no post-replay reader, while
-    `call_frame_arena` is referenced ONLY by `CallFrameBase`/`Descend`/`Return`
-    (Phase D dispatch) — sequential, disjoint live windows. See
-    `docs/call-frame-memory-layout.md` §5. -/
+    The historical 1G layout used a 244 MiB basr union and a 228 MiB frame
+    stride. Those values remain useful in the design history but are not live
+    geometry. The current phase-ownership argument is the same execution-dead
+    Phase-H / Phase-D sequencing: the five child arrays are used during the
+    pre-dispatch state-root recompute, while the frame arena is used during
+    dispatch. See `docs/call-frame-memory-layout.md` §5. -/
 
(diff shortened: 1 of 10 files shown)


pirapira:
**Reviewed — label on green build.** The headline is the right one, and it is a soundness finding rather than a cleanup.

⛔⛔ **A kernel-checked proof was asserting a fact about a region that does not exist.** `CallFramePhase`'s six-child `dataUnionChildren` family named `baap_storage_delete_paths`, which is absent from `stateless_guest` — and it typechecked, because **the constant was real even though the region was not.** Nothing in CI could see it: `check-axioms` was green, the kernel was satisfied, and the statement was about nothing. ⇒ *That is the strongest possible argument for settling structures before writing predicates: a separation-logic predicate over a phantom region would be vacuously fine and completely useless.*

⭐ **The sweep bounds the damage with a count, which is what makes it publishable**: exactly one current proof family, **all 5 union symbols and all 1,091 `GuestAddrs` names present** in a fresh ELF. Remaining references are the probe emitter and historical docs — not theorems about the main image. **A named count beats "we looked around."**

**Repaired six → five with no theorem weakening** — the trap here was fixing it by dropping the symbol and leaving the count, which would have been the same defect with a different literal.

⭐ **The generator fix outranks the map fix.** `gen-symbol-addresses.py` carrying stale `.bss = 0xa4000000` / `.sszscratch = 0xbf800000` is a different class from a wrong docstring: **a wrong constant in a generator is consumed and propagated into every file it emits**, manufacturing agreement between artefacts that are all wrong together — which reads as consistency. Same failure mode as `textSizeBytes`: an emission *input* is never verified by what it produces.

⇒ **The guard is what makes this permanent**, and it was demonstrated firing on **both** failure modes — stale pin → exit 1, phantom → exit 1. ⚠️ *A guard nobody has seen fail is indistinguishable from a guard that cannot*, and six instrument-scope failures today were all cases where the check was green because it did not look. Extending it to compare every `GuestAddrs` name plus the closed union names against `readelf` means a phantom cannot recur silently.

**The parametric-base rule is the correct boundary** and belongs in the vocabulary: **absolute pins are image-specific guard inputs, never predicate literals.** Three arena-base values circulated tonight — `0xaf420780` (stale), `0xadd053a0` (current main), `0xad3dd7a0` (a branch with 9.155 MiB less `.bss`) — and **two of the three were correct for their own images.** A predicate quoting an absolute base without naming its image is not stale-in-principle, it is meaningless.

pirapira:
Excluded from automated merging: this PR leaves `check-region-map` in DRIFT, and the repair needs a decision I should not make unilaterally.

**It is this PR alone, not a batching artifact.** I tested the branch merged with main and nothing else:
```
DRIFT RegionMap.callFrameArenaBase:  expected 00000000ad3dd7a0, ELF has 00000000ad3dd5e0
DRIFT RegionMap.evmMemoryPoolBase:   expected 00000000b37f67a0, ELF has 00000000b37f65e0
DRIFT RegionMap.syslogBase:          expected 00000000ab07a800, ELF has 00000000ab07a640
```

**Why it is not the usual repin.** These are *base* pins, a category the documented regen procedure does not cover — it names `.text` / `.data` / `.bss` sizes only. Two things make them harder:

1. **They feed emission.** Naively repinning them fails the relink outright:
   ```
   RegionMap.lean:781: Tactic  proved that the proposition
     evmMemoryPoolRegion.base = 3011471264 ∧ ... = 3112134560
   is false
   ```
   So a base repin needs a fixpoint loop, like `textSizeBytes`.

2. **It requires editing a theorem.** `evmMemoryPoolRegion_matches_elf` (:779-781) hardcodes `0xb37f67a0` / `0xb97f67a0`. Converging the pins means changing those literals. That is arguably inside the "addresses may change due to ELF layout" exception — but it is still a kernel-checked statement, and I would rather have it sanctioned than assume.

There is also an irony worth surfacing: this PR is titled *Audit RegionMap pins and phantom union entries*, and the audit itself leaves three pins drifted. Symbolising `evmMemoryPoolBase` (`base := 0xb5839780` -> `base := evmMemoryPoolBase`) is what pulled these into the pinned set.

**Everything else about the branch is fine** — it merges cleanly with main, and the only failing gate is `check-region-map`.

Tell me to proceed and I will do the base-pin fixpoint plus the `evmMemoryPoolRegion_matches_elf` literal update, then verify the full gate set. #11228 was landed separately in the meantime.

pirapira:
Correction to my previous comment: one line was mangled by my own shell quoting. The relink error reads:

```
RegionMap.lean:781: Tactic `decide` proved that the proposition
  evmMemoryPoolRegion.base = 3011471264 ∧ evmMemoryPoolRegion.base + evmMemoryPoolRegion.size = 3112134560
is false
```

(`3011471264` = `0xb37f67a0`, the old base; the ELF now has `0xb37f65e0`.) Everything else in that comment stands.

pirapira:
Unblocked — clearing `complicated-merge`.

The base-pin drift is resolved by `8c9081066` / `3ecbe608e`: `callFrameArenaBase` -> `0xad3dd5e0`, `evmMemoryPoolBase` -> `0xb37f65e0`, `syslogBase` -> `0xab07a640`, with `evmMemoryPoolRegion_matches_elf` updated to `0xb37f65e0` / `0xb97f65e0` to match. The fixpoint holds — a relink reproduces exactly those values.

Verified on the branch merged with current main: `check-region-map` reports "region map matches the linked ELF", `lake build` green, and all 15 `check-*.sh` pass including `check-build-units-link` and `check-asm-to-program`. Regen produced no drift, so I had nothing to add — auto-merge enabled as-is.

(`c3798f62e` making the region-map guard fail loudly when its tooling is missing is a good catch too — that class of silent skip is exactly what let earlier drift through.)

PR #11225, opened 2026-08-02
chore: batch-merge #11218 #11219

pirapira:
Batch-merge of #11218 #11219 (all author pirapira).

Verified locally: `lake build` + all 15 `check-*.sh` green, including `check-build-units-link` and `check-asm-to-program`.

Layout repins: `textSizeBytes` -> `0x06338c`, `bssSizeBytes` -> `0x1ac6cb20`, with the handwritten `GuestImage.lean` extent/end literals repinned to `0x1ac6cb20` / `0xbdd7cb20`. Regenerated `GuestAddrs.lean` / `symbol-addresses.tsv` / `GuestImageEntries.lean`, converged to a fixed point.

Note: #11218 was previously parked `complicated-merge` for contradicting main's system-skip set; it has since been reworked to `bvMtxSystemSkipEntries := 0` and now merges cleanly.


pirapira:
**Timeline note for the batch — `#11218`'s branch moved after this batch was cut.**

| event | time |
|---|---|
| this batch created | 19:28:55Z |
| `#11218` head pushed → `bdf224179` | 19:36:55Z |

⇒ **No information is lost if this batch merges as-is.** That later commit is a *fixpoint regen* the author started before seeing my stop message — I had reported the `check-region-map` drift at 19:20, then told them at 19:33 to drop it because this batch already performs the regen (`textSizeBytes` → `0x06338c`, `bssSizeBytes` → `0x1ac6cb20`, `GuestImage` literals repinned, `GuestAddrs`/`symbol-addresses` regenerated, all 15 checks green). **The author's commit is redundant with this batch's own layout work, not additional to it.**

⚠️ Flagging it only because **a batch whose input branch moved after the cut is exactly the case where a silent merge diverges from what was verified.** Two fixpoints for one branch is how a batch acquires a conflicting layout commit. The author has confirmed hands-off on `#11218` and will not push again.

Content in the batch is unaffected: the six-skip union decision (`#11218` dropping 7002/7251/6110/SYSTEM_ADDRESS on top of main's 2935/4788 via `#11210`) and the account-write undo cap derivation (`#11219`) are both what was reviewed and labelled.

⇒ Also worth recording, since it validates this batch's regen: **`check-build-units-link` passing here confirms `#11220`'s failure of that same gate is specific to its cut** rather than ambient. That PR deleted two `.bss` symbols still referenced by `zisk_stateless_verdict_v2`; its author has diagnosed them as **debug-dump `la` references, not genuine exercise**, and is applying the same deletion to the sibling unit rather than reverting.

Issue #11166, opened 2026-08-02
GASPRICE: missing conditional effective environment staging in MTx


pirapira:
Split from #10858 after current-main per-entry remeasurement. This is the -1 storage-change / +1 storage-read direction.

## Current-main witnesses

Artifact provenance: base 703336c28a2b7c2f240d77f104e96e6281112d73; guest SHA256 c16ff50d559342d5f177db12dbc2afe002519397951f9ee512d61b802e7441a4; backend spike; fixture tag tests-zkevm@v0.6.2.

| fixture | stable case ID | affected account | declared -> rebuilt storage_changes | declared -> rebuilt storage_reads | rebuilt/supplied |
|---|---|---|---|---|---:|
| 17262 | 62e7184bf960a09ba9b5500881f50db64ee0e5ec6a59279e4ef0ead15ecfe206 | 49a0fe79e28d1d65e16cdf53acafeae7baccac0e | [0,1,2] -> [1,2] | [] -> [0] | 382/389 |
| 15930 | 2c8c52389b56933260aef47b14540725817d80f079dd27000f2539a48ff4da27 | 5fbaa533d8cbbcaec8466f99b4f9e81e648a7520 | [1,2,3,4,7,330,332,336,338,340] -> [1,2,3,4,7,330,332,336,338] | [] -> [340] | 564/569 |

Both missing rows are BAI=1 transaction-phase storage changes. The serializer read-exclusion rule then preserves the same slot as a read because it is no longer present in the rebuilt storage_changes set. This is distinct from #10866, which was the post-exec BAI=N+1 producer and is closed.

The mechanism is not destroy-specific: the witnesses are merge/callcode-labelled cases, while the selfdestruct-labelled cases are the opposite direction in the sibling split.

## Closing condition

For both stable case IDs on current main, reconstruct the guest BAL completely and verify per entry that every declared BAI=1 storage change is present with the correct slot/value, and that the corresponding slot is absent from storage_reads. Aggregate length or pass/fail alone is insufficient.

pirapira:
## Re-derive verdict: **PREMISE HOLDS** (length/delta still live; full per-slot decode not re-run)

Re-derive base: origin/main `9298b78fb` (#11179 in). Guest sha256 `50fbd45b649c92a41da6aa3e899701d4695680885d47e05867fbc1b8e0321ee1`. **Verdict only — no fix.**

### Witness remeasure (spike, current main guest)
[...]

PR #11226, opened 2026-08-02
fix: stage effective GASPRICE for MTx runtime

pirapira:
## Summary

`GASPRICE` is a transaction environment value, not a value that may remain at
zero when a multi-transaction call has no context base-fee pointer.  The
execution-specs path installs `effective_gas_price` in `tx_env` before message
execution; the MTx guest path left `ctx+32` null, so the existing dispatcher
kept the environment word at zero.  This is the wrong-value opcode defect
tracked by #11166.

The fix supplies the already-authenticated `bv_mtx_base_fee_be` buffer at
`bv_mtx_ctx+32` immediately after context extraction.  It reuses the existing
`tx_effective_gas_pricing` producer; it does not add a second gas-price
calculation.

## Execution-specs basis

Against execution-specs pinned at `e5a8caf1b8055e4d805c7fb169edfa710914b7da`:

- `environment.py:306-323` makes GASPRICE push `evm.message.tx_env.gas_price`.
- `fork.py:618-624` computes the effective gas price.
- `fork.py:1127-1134` installs that effective price into `tx_env` before
  `process_message`.

The emitted `tx_effective_gas_pricing` routine reads the supplied base-fee
pointer, extracts the transaction pricing fields, computes the priority fee,
then performs `u256_add_be(base_fee, priority_fee)` before the dispatcher
copies the result to the GASPRICE environment word.  Legacy and EIP-2930
transactions are normalized by `tx_extract_gas_pricing` into equal
max-priority/max-fee values, so the same formula yields the legacy
`tx.gas_price` for valid legacy transactions (and rejects a price below the
base fee).  The emitted arithmetic is in
`gen-out/regionmap/stateless_guest.s:28839-28900`.

The widening audit found GASPRICE is the only environment field gated by the
missing `ctx+32` pointer.  MTx creation already supplies its own `ctx+32`;
CALLER/ORIGIN use `ctx+24`, CALLVALUE reads `ctx+96` unconditionally, and
SELFBALANCE has independent absent-account semantics.

## Behavioural evidence

All measurements use the same runner
`/home/yoichi-bkp/evm-asm3/scripts/spike/spike_run` (sha256
`5b383a7702fb05dea16a2cc37dc313b82491fdf194f39c12306cb387a5a2b6b0`) and the
final candidate ELF (sha256
`800126400110897e330e8c8aa558da065d2d621c57494f12f154cef34ef3e227`).

The random-200 manifest is copied under the owner-controlled run directory at
`/tmp/evm-asm3-codex/11166-gasprice-fix/random200-manifest.tsv` (sha256
`a288c614769ad137ad54d08d2106a3c16174ecf7afb1118a17795919ec32c13c`).  The
baseline is the named parent ELF from base commit `2499b52d1` (sha256
`b6c67a65055d703e3efb7bfa210b7d0a17eae701750a1feeae2f4d272d12c337`); the
candidate is commit `0675153c4`.

With `WORKERS=30`, `SPIKE_OUTPUT_LEN=256`, and the copied manifest:

| leg | FA | FR | OK |
| --- | --: | --: | --: |
| parent | 0 | 12 | 188 |
| candidate | 0 | 11 | 189 |

The only forward flip is
`00081_test_scenarios_fork_Amsterdam-blockchain_test-test_program_program_GASPRICE-debug__b6`.
There are zero reverse flips.

Targeted final-ELF checks:

- `15930`: rebuilt/supplied `569/569`, fail `0`, shadow `0` (parent was
  `564/569`, fail `1`).
- `17262`: rebuilt/supplied `389/389`, fail `0`, shadow `0` (parent was
  `382/389`, fail `1`).
- `04778` expected-invalid control still rejects with fail `5`.
- `00337` remains `370/370`, fail `0`.

The corrected opcode has a wider blast radius than BAL: a wrong GASPRICE can
change contract branches, state writes, gas accounting, and receipts.  #11166
was the BAL witness that exposed it.

## Build and layout checks

- `LAKE_ARTIFACT_CACHE=false lake build` (4796 jobs): pass.
- `scripts/check-build-parallel.sh`: all five checks pass.
- `scripts/check-asm-to-program.sh`: 386 converted definitions, clean.
- `scripts/check-region-map.sh`: clean.
- `scripts/check-unimported.sh`: 2826/2826 reachable, zero orphans.
- `scripts/check-no-warnings.sh`, forbidden-tactics, layering,
  no-hardcoded-guest-PC, and heartbeat approval checks: pass.

Final layout: `.text=0x63830`, `.data=0x5370`, `.bss=0x1b594860`; the pinned
`bnf_le_a=0xa3000ee0` and `.data` size remain unchanged.

Concrete defect citation: `h_GASPRICE` in `EvmAsm/Codegen/Programs/EvmEnvHandlers.lean:40` is a zero-initialized environment load, while `BlockVerdictDispatchTx.lean:857-886` only stages the effective price when `ctx+32` is non-null. The execution-specs path installs `effective_gas_price` unconditionally at `fork.py:1127-1134`.


pirapira:
**Reviewed — label on green build.** This is an execution-semantics defect found by chasing a five-byte BAL discrepancy, and the diagnosis is the valuable part.

⛔ **The guest's `GASPRICE` opcode was pushing zero.** `h_GASPRICE` is a zero-initialised env load (`EvmEnvHandlers.lean:40`), and staging at `BlockVerdictDispatchTx.lean:857-886` was **conditional on `ctx+32`** — a condition with **no execution-specs counterpart**, since `fork.py:1127-1134` installs `effective_gas_price` in `tx_env` unconditionally for every transaction.

⇒ **Three layers were exonerated by measurement before the real one was found**, and that ordering is why this landed as a one-line-shaped fix rather than a rewrite:

| suspected | verdict |
|---|---|
| BAL read-side reclassification | **innocent** — `block_access_lists.py:667-676` correctly treats pre=post as not-a-change |
| `storage_write_record` capture | **innocent** — faithfully recorded the zero it was handed |
| missing row / clear-after-write / wrong reader | **all excluded** — `SPIKE_WATCH` hits=0 on both value cells |

⭐⭐ **The correctness question was settled by reading the emitted arithmetic, not by the A/B — and that distinction is the point.** `ctx+32` is the **base-fee input**: `tx_effective_gas_pricing` calls `tx_extract_gas_pricing`, then `priority_fee_per_gas_eip1559`, then **`u256_add_be(base_fee, priority)`** into `env+160`. It never pushes `ctx+32` directly.

⇒ Had it pushed directly, `GASPRICE` would now return **base fee** — correct for every zero-tip transaction and **wrong for every tipped one**. FA=0 across 200 rows could not have detected that, because *a sample that lacks the discriminating case cannot speak to it.* The instruction sequence can, and it holds for all transactions rather than the sampled ones. **The legacy arm is closed the same way**: `tx_extract_gas_pricing` normalises legacy/2930 `gas_price` into both max_priority and max_fee, so the same formula yields `gas_price` — where the spec's effective price is `tx.gas_price` outright, unrelated to base fee.

**The fix shape is right**: store the already-decoded authenticated `bv_mtx_base_fee_be` into `bv_mtx_ctx+32`, supplying an **existing** computation. ⇒ *No new producer* — the deficient path reaches the working one, rather than acquiring its own copy of a computation that already exists.

**Widening audit bounds it to one field**: one production caller, MTx creation has its own `ctx+32` store at `:808`, CALLER/ORIGIN use `ctx+24`, CALLVALUE reads `ctx+96` unconditionally, SELFBALANCE's miss matches absent-account semantics.

**Measurements**: random-200 **FA=0 both legs**, FR 12→11, **forward 1 / reverse 0** — the single flip being `00081 program_GASPRICE-debug`. Targeted: 15930 `564/564`, 17262 `389/389`, ⚠️ **04778 expected-invalid still rejects** (fail5) and 00337 unchanged — the control that proves passes were not bought by weakening a check.

**Blast radius, correctly stated**: a wrong-value opcode changes branches taken, state, gas and receipts — not only BAL rows.

pirapira:
**Labelled while CONFLICTING — deliberate, and the reasoning is worth stating since it departs from the usual gate.**

My normal rule is *label on green*, because a build has caught real defects on a mergeable head (a pin drift on `#11182` that no A/B could have seen). ⇒ **But a CONFLICTING PR cannot produce that green**: GitHub cannot construct a merge ref, so the `build` check never reports, and waiting for it is waiting for something that will never arrive. The batch is the mechanism that produces a verified build here — exactly how `#11210` and `#11218` landed tonight, both labelled while conflicting.

⚠️ **The conflict is layout-only and it is not the author's to resolve.** Main moved twice in thirty minutes (`#11217` removing 9.155 MiB of `.bss`, then `#11225`), so the overlap is in `GuestAddrs.lean`, `RegionMap.lean`, `symbol-addresses.tsv` and the two touched `Programs/` files. ⇒ *That is a layout regen, not a rebase*, and a regen belongs in the batch's own layout commit. Three PRs are deleting text concurrently; an author-side fixpoint converges to a state the next merge invalidates — which has already cost two agents a wasted regen tonight.

**What the batch must not lose**, since the diff is `+969/−961` and mostly generated:

- the **behavioural** change is small — store the already-decoded authenticated `bv_mtx_base_fee_be` into `bv_mtx_ctx+32` after `multi_tx_nth_context`, in `BlockVerdictDispatchTx.lean` / `BlockVerdictMtxRuntime.lean`
- everything else is `GuestAddrs` / `RegionMap` / `symbol-addresses.tsv` regeneration, which the batch will recompute anyway

⭐ **The correctness evidence does not decay with the base**, which is why relabelling after a regen is unnecessary: the discharge is the **emitted arithmetic** — `tx_effective_gas_pricing` calling `tx_extract_gas_pricing` → `priority_fee_per_gas_eip1559` → `u256_add_be(base_fee, priority)`. That is a property of the instruction sequence, not of a sample or a base, and it is what proves `ctx+32` is the base-fee *input* rather than the effective price itself.

⇒ If the regen changes emitted behaviour rather than addresses, that is a **new finding** and I want it raised rather than absorbed — but a pure repin cannot alter the opcode fix.

PR #11227, opened 2026-08-02
docs: record dispatcher journal initialization contract

### LINKED ISSUE #11152: runtime_dispatcher_call setup: 15 journals zeroed on system re-entry still unproven (post-#11147 residual)

pirapira:
Closes #11152.

Records the proof-lane initialization contract discovered by the per-cell emitted audit:

- names the twelve zero-assuming scalar journal preconditions with cell, reader, lifetime invariant, and failure mode;
- keeps `evm_selfdestruct_staged` and `cd_destroyed_empty_hits` explicitly retained because SPIKE_WATCH absence over a finite corpus is not an all-path deletion proof;
- records the fresh ELF/fixture evidence and the asymmetric evidence boundary;
- links the contract from the authoritative region-map document.

Docs-only. No emitted code or journal stores changed.


pirapira:
**Reviewed — label on green build.** Docs-only, `+92/−0`, and it is a proof-lane artefact rather than a note: it records obligations that were real and unwritten.

⭐ **The investigation reversed its own disposition, which is the point.** #11152 was scoped as a likely deletion — fourteen unconditional wipes where execution-specs has no counterpart. The per-cell audit found **twelve of the fourteen have a reader that assumes zero before any writer runs**: gas settle reading a zero refund on a no-SSTORE tx, overflow checks requiring zero absent an overflow writer, the first `LOG` reading the cursor before its own increment, undo paths reading zero before any entry exists.

⇒ **A block-level judgement would have called the wipe a leftover and broken twelve routines silently.** Fourteen jointly-harmless cells contained twelve individually essential ones — *a cell whose only initialiser is this wipe looks identical to one nobody reads.* That is why the unit of work had to be per cell.

⭐ **The spec comparison is load-bearing rather than colour**, and it converts a divergence into a justified one: execution-specs obtains the same guarantee **structurally**, through fresh per-call `TransactionEnvironment`/`Message` objects and snapshot/restore scoping; the guest obtains it by **explicit clearing**. ⇒ *Same contract, different mechanism* — so the obligation is real, not an artefact of our style, and #11152 closes as **structural divergence with a reason** rather than as a leftover.

⚠️ **Retaining the two sampled no-write cells is the right call and the reasoning is the transferable part.** `hits=0` across a corpus is **absence of observed writes on a sample, not proof that no path writes them** — and the evidence here is *asymmetric*: one reader assuming zero is sufficient to **keep** a cell, while absence across a sample is weak grounds to **delete** one. Two stores of gain against a silent-state-bug class we cannot detect cheaply is a bad trade. The body states the asymmetry explicitly, which is what makes the retention a decision rather than an omission.

⇒ **This lands directly in the predicate work.** Each of the twelve is a precondition nobody had written down — *this routine is correct only if that cell is zero on entry* — which is the slot a separation-logic predicate needs beyond byte range and entry layout.

⚠️ One scope note for the next reader, from scoper's independent check: this audit covers **`runtime_dispatcher_call` entry**. None of the twelve touch `evm_memory_pool` or `baap_storage_values`, but those absences are not equivalent — the pool has genuinely no entry obligation (disjointness follows from the address formula), while the union child simply **has not been audited at `bal_account_apply_post_fields` entry**. ⇒ *An initialisation audit is scoped to the entry point you audited*, exactly as a deadness claim is scoped to the image you measured. That audit is now running.

PR #11228, opened 2026-08-02
cleanup(codegen): retire the five write-only BmvMx comparison flags

pirapira:
## Retire the five write-only BmvMx comparison flags

GH #11211. `bmvmx_avail`, `bmvmx_sender_match`, `bmvmx_sender_checked`, `bmvmx_coinbase_match`, `bmvmx_coinbase_checked` are **written and never read**.

### The predicate is write-only, not deadness

The symbols **are** written, so a reference census reports them live. The question is whether any load ever reads them. Measured on the emitted image, `main` `9ffb5398a`:

| unit | flag references | **loads** |
|---|---|---|
| `stateless_guest` | 14, all `sd`, all inside `block_verdict` | **0** — no `ld`/`lw`/`lbu` anywhere |
| `zisk_stateless_verdict_v2` | emitted separately, censused on its **own** `.s` | **0** |

⭐ **The cross-unit check is deliberate.** A deadness claim is scoped to the image measured, and a sibling unit referencing a symbol dead in the main image is a real CI break rather than a theory — so the probe unit was emitted with `lake exe codegen --program zisk_stateless_verdict_v2` and censused directly, not inferred from the module graph.

### ⛔ `bv_exec_p` is NOT touched — the cut boundary is inside the block

It shares the same initialisation block and is emphatically live: **45 references, 42 of them loads**, across eight routines (`block_verdict`, `block_verdict_mtx_oog_materialize`, `bal_addr_is_tx_sender`, `bal_txs_contains_push20_selfdestruct`, `bal_txs_contains_create_collision_touch`, `bal_txs_contains_top_create2_collision_touch`, `simple_transfer_tx_context`, `dispatch_tx_runtime_code`). Deleting the block wholesale would take a live initialiser — which is why #11212 excluded it.

The interleaved `bvReceiptsShapeClear` / `bvReceiptsShapeSet 1 true` calls sharing source lines with two of the stores are preserved; the edit removes only the emitted-string fragment, not the line.

### No debug dump reads them

The only enclosing routine for a flag reference, **in either unit**, is `block_verdict`, and every reference is a store. So nothing prints them and removing them cannot shift a dump's field offsets.

### Computed-offset audit

The five flags are **not contiguous** — `bmvmx_avail` sits between `dwp_al_len` and `eip7708_tl_typed_avail`, `bmvmx_coinbase_match` sits *inside* the `bmvmx_cb_*` cluster, and the other three are contiguous ~7.7 MB away. Every `la` of an immediate neighbour uses offset 0:

| neighbour | `la` sites | any offset ≥ 8 |
|---|---|---|
| `eip7708_tl_typed_avail` | 4 | none |
| `bmvmx_cb_nonce_len` | 2 | none |
| `bmvmx_cb_balbytes` | 4 | none |
| `sdfg_gascost` | 2 | none |
| `dwp_al_len`, `yisv8_self_bal` | **0** | — |

⚠️ **Limits, named.** The window is the `la` plus three lines, so a store-and-reload pointer flow is **not** excluded. And the last two have **zero `la` sites in this image**, so reach from those directions was never tested — their absence is not evidence.

### Sections — read after a two-pass regen to fixpoint

```
.text  0x6338c    -> 0x632d8      -180 B
.data  0x5370     -> 0x5370       UNCHANGED
.bss   0x1ac6cb20 -> 0x1ac6cb00   -32 B
```

⭐ **The `.bss` shrink is 32 bytes, not the 40 that five 8-byte cells suggest** — two of the five are followed by `.balign 32` and the padding reabsorbs 8. I flagged this before cutting precisely because it could not be predicted; **the regen computed it.** Pass 2 changed nothing, which is the convergence check.

### A/B — 1,045-row stride sample

| leg | ELF sha256 | OK | FR | FA |
|---|---|---|---|---|
| parent (`main` `9ffb5398a`) | `f1583b84140c850d…` | 1004 | 41 | **0** |
| candidate | `f89575914d5ecfd9…` | 1004 | 41 | **0** |

**0 forward, 0 reverse, 0 other category changes, FA = 0 both legs.** Identical, as a write-only removal must be.

### ⚠️ Residue, named rather than silently absorbed

The four `.Lbmvmx_addr20_ne` distinctness comparisons fed **only** `*_checked`, so they are now dead compute — four `jal` plus four `bnez` to the following instruction — and the helper's only callers are those four. **Removing them is a larger cut than was measured and reported**, so it is left for a follow-up rather than expanded into here.

Three stale comments that described the removed stores were corrected in the same edit.


pirapira:
**Reviewed — label on green build.** Small cut, but the process around it is what makes it safe, and three separate things nearly went wrong.

⭐ **The `.bss` shrink came in 8 bytes under the arithmetic** — five 8-byte cells is 40 B, measured is **32 B** — which is precisely why the `.balign 32` observation was flagged *before* cutting rather than reconciled after. ⇒ *Removing 8 bytes adjacent to an alignment directive shifts downstream by 0, 8, 16 or 24 depending on how padding reabsorbs.* A predicted figure here would have looked like a defect against the real one; the regen computes it, nobody chooses it.

⚠️ **The whole-line delete trap is the most transferable part.** The first attempt removed lines and the assertion caught a fragment gone missing — **because a line carried two concatenated things**, so deleting it would have taken `bvReceiptsShapeClear` / `bvReceiptsShapeSet 1 true` with it. Redoing it string-wise preserved both. ⇒ *The file was untouched only because the write came after the assert* — the guard was load-bearing, not decorative.

⭐ **Write-only was confirmed in both build units**, not just `stateless_guest`. That check exists because `#11220` failed CI tonight on symbols dead in the main image and **live in `zisk_stateless_verdict_v2`** — *a deadness claim is scoped to the image you measured.*

**Three sites in three regions, not one block**: `bmvmx_coinbase_match` sits inside the `bmvmx_cb_` cluster while the sender/coinbase trio is **7.7 MB away**. The name prefix implied one contiguous block; the addresses said otherwise.

**Scope held where it mattered**: `bv_exec_p` initialisation stays — it is consumed by no-tx and MTx parsing, and `#11212` excluded this block from its own cut precisely to avoid taking a live initialiser. **The boundary is inside the block.**

⇒ And the residue is **named rather than absorbed**: the four `.Lbmvmx_addr20_ne` distinctness comparisons fed only the removed `_checked` flags, so they are now dead compute — four `jal` plus four `bnez` — with the helper having exactly those four callers. **That is a larger cut than was measured, so it is correctly left to a follow-up** rather than smuggled in on the strength of an adjacent argument.

**Limits carried into the body verbatim**, including that `dwp_al_len` and `yisv8_self_bal` have zero `la` sites in this image, so reach from those directions was never tested — **their absence is not evidence.**

Issue #11229, opened 2026-08-02
Static reference censuses are same-line la+load: 'no load anywhere' is an upper bound on deadness, not a proof


pirapira:
## Our static reference censuses are same-line `la`+load — so **every "no load anywhere" claim is an upper bound on deadness, never a proof**

Filed as its own issue because it outlives every deletion PR we will write, and because it was diagnosed three separate times tonight before anyone wrote it down.

### The instrument

Every reference census in this repo's deletion work has the same shape: for a symbol `S`, find lines matching `la <reg>, S` and check whether the **same line** contains a load off `<reg>`. It is fast, it works on the emitted `.s`, and it is what every recent removal-band audit has used.

### Three mechanisms it cannot see

1. **A pointer passed to a callee that loads through it.**
   `la a0, bmvmx_acct; addi a0, a0, 8; … jal ra, u256_add_be` — the callee loads through `a0`. The read exists; no line names the symbol and loads from it.
2. **A multi-line write.** `la` on one line, the store on another. So a census of *written* cells is a **floor**, not a total.
3. **A store-and-reload pointer flow.** `la` the symbol, store the pointer to memory, reload it later, dereference. Nothing on any single line pairs the name with the access.

### Consequence, stated precisely

* **A "no load anywhere" verdict is an UPPER BOUND on deadness.** It says *this detector found no read*, which is compatible with a read existing.
* **A written-cell census is a FLOOR.** It says *at least these cells are written*.
* ⇒ **neither direction supports a deletion on its own.** Both have to be corroborated by something that observes the cell rather than the reference.

### Three measured instances, not a worry

| instance | what the detector missed |
|---|---|
| **This audit (#11211)** | `bmvmx_acct` and `bmvmx_cb_post` came back "no load anywhere". Both are read by callees through a passed pointer. **I retracted both rows.** |
| **R8** | a data buffer was reachable by `la`-plus-displacement **from a neighbour symbol**, so zero references to its own name did not make it dead. |
| **#11220** | symbols dead in `stateless_guest` were **live in `zisk_stateless_verdict_v2`** — a scope failure of the same family, and it cost a CI round. |

### ⭐ What this does NOT invalidate — and this matters as much

[...]

PR #11232, opened 2026-08-02
chore(codegen): unlink bv_emit_single_tx_tl7708 + drop never-written buffer (#10685 PR2)

pirapira:
## Summary

#10685 PR2: unlink `bv_emit_single_tx_tl7708` and drop never-written `bv_simple_transfer_tx` (192 B BSS).

### Evidence (lead with SPIKE_WATCH — #11229 does not bound cell watches)

| probe | result |
|---|---|
| **SPIKE_WATCH** `bv_simple_transfer_tx@0xab038c60` r200 | **hits=0 all 200**, final=0 |
| BREAK `h_SLOAD` sample 40 | **34/40** — same instrument fires |
| BREAK emit_tl entry `0x8001ac48` | **1/200** (00103 value_move_to_precompiles) |
| BREAK jal emit / emit body | **0/200** — path never taken, not dead probe |
| 00103 lane at emit_tl | **t0=t1=2** → beq `.Ldtrc_mtx_precompile_success`, skips jal |

**Mode-2 answer:** emit is live-encoded behind a gate the only exercised caller always bypasses.

### eip7708_tl_* reader census (deletion vs defect)

- **Readers:** only `dispatcher_reemit_pending_tl` → `eip7708_append_transfer_log` when `bv_pending_tl_flag≠0`.
- **Live writers (KEEP):** MTx (`bv_mtx_ctx`) and CreationStage.
- **Emit gates before any tl store:** word0≠0 ret; value@+96 all-zero ret; +24 zero ret; self-transfer ret.
- Watch proves buffer always zero → **gate 2 fires → ret before any `eip7708_tl_*` store or pending flag**.
- **Not** GASPRICE shape (live code propagating never-produced value). Deletion is pure no-op even if jal ever taken.

### Cut

- Unlink `blockVerdictSingleTxTopLevelLogFunction` from guest aggregates (KEEP Function string for probes)
- Drop `bv_simple_transfer_tx` BSS
- Remove `jal bv_emit_single_tx_tl7708` from Publish emit_tl arm (keep `dispatcher_reemit_pending_tl` no-op without flag)
- Drop v2 debug dump OUTPUT+344 (`la bv_simple_transfer_tx`) — same sibling-unit trap class as #11220 360/368
- **KEEP:** `simple_transfer_intrinsic_gas`, Publish `bv_mtx_ctx` path, live `eip7708_tl_*`

### Layout (ELF-measured fixpoint)

| | parent main `608b44b16` | cand |
|---|---|---|
| text | `0x628c8` | `0x62744` (−436 B) |
| bss | `0x1ac6c900` | `0x1ac6c840` (−320 B) |
| guest sha | `5caf4500…` | `c96b6511…` |

### A/B random-200 seed42

| leg | FA | FR | OK | TN |
|---|---|---|---|---|
| parent | 0 | 4 | 186 | 10 |
| cand | 0 | 4 | 186 | 10 |

ELFs DISTINCT. OUTPUT 256B **byte-identical 200/200**. reverse 0. forward 0.

### Limits named

- No separate v2 ELF measured dynamically; v2 shares BlockVerdict dump source (read). Dump drop + `check-build-units-link` (v2 OK) cover both units.
- jal 0/200 is sample floor; instrument-proof is watch + gate analysis.
- Stale comment left (not fixed): EIP7708Logs still says pending flag set only by single-tx emit — MTx+CREATE also set it.

### Adjacent named leave alone

Default `PublishAsmFor "bv_simple_transfer_tx"` / PrecompileGas string defaults; `eip7702SenderSelfAuthPostNonceCheck` source still names deleted BSS but is not in guest image.

Closes nothing alone; part of #10685. No merge! from author.


pirapira:
**Reviewed — label on green build.** The important work here is the check that *changed the PR's claim*, not the deletion.

⭐⭐ **The deletion-versus-defect question was asked and answered before cutting.** `bv_emit_single_tx_tl7708` reads a buffer proven always-zero and copies fields **out** to `eip7708_tl_*`. ⇒ *If that path ever executed, this would not be dead code — it would be live code propagating a never-produced value*, which is precisely the shape of the `GASPRICE` defect found earlier tonight (`h_GASPRICE` faithfully pushing a zero nobody staged). The reader census settled it: **emit early-exits on the zero buffer before any `tl` write**, so nothing downstream ever receives zeros. `chore: unlink`, not `fix: stop propagating`.

⭐ **The instrument hierarchy is right, and it matters under #11229.** Lead evidence is **`SPIKE_WATCH` on the cell** — `bv_simple_transfer_tx` @ `0xab038c60`, `n=200`, `sum_hits=0`, final zero on every row. ⇒ *That bound does not apply here*: the watch observes the **cell**, not the reference, so it sees a pointer-mediated write exactly as well as a direct one. The static `la` census is correctly demoted to corroboration — on its own it would be a **floor, not a proof**.

⚠️ **And the zero was made interpretable by a positive control**, which is the step that is usually skipped: the same `SPIKE_BREAK` instrument **fires** at `emit_tl` entry on 00103 and at `h_SLOAD`, while reading 0 at the `jal` and at the emit body. ⇒ So the zero is **path-never-taken, not dead probe** — the two are indistinguishable from the zero alone, exactly as leg W was uninterpretable until leg X fired on 850/1045.

**The mode-2 answer is concrete rather than inferred**: at `emit_tl` entry, `lane t0=2` takes `.Ldtrc_mtx_precompile_success` and skips the `jal`. **Emit is live-encoded behind a gate the only exercised caller always bypasses** — a much sharper statement than "never called."

**A/B**: FA 0, **FR 4 identical 200/200**, reverse 0. `.text` −436 B, `.bss` −320 B. **Multi-unit link verified including `zisk_stateless_verdict_v2`** — the check that failed #11220 in CI tonight, now run *before* the cut.

⇒ And the `OUTPUT+344` debug dump is dropped with the buffer — same sibling-unit trap class as this author's own 360/368, with the smoke harness confirmed not to parse 344. **An output nobody reads and an output someone parses look identical from the emitter side.**

Issue #11234, opened 2026-08-02
convergence: EIP-7702 code BAL append delayed to post-runtime finals (spec sets code inside set_delegation)


pirapira:
## What

Guest still does a **second walk** of the type-4 authorization list **after runtime** to append **code** effects for BAL / code comparator:

- `block_verdict_eip7702_auth_nonstorage_effects_array` (MtxTail)
- → `eip7702_auth_nonstorage_effects`

Source (`TxIntrinsicAuthEffects.lean`):

> This post-runtime helper owns the EIP-7702 **code-comparator append** derived from **BAL finals**. It intentionally no longer emits the authorization nonce raw effect: `eip7702_auth_state_prepare` publishes that effect at authorization time… The code append remains here because its **final-state input does not exist until after runtime execution**.

## Spec counterpart

execution-specs `set_delegation` (`eoa_delegation.py` / `interpreter.py` depth 0) sets authority **code inside the single pre-message walk**. Code is not deferred until block-access-list finals.

## Why this is interesting (same shape as #10651)

| | spec | guest (this arm) |
|---|---|---|
| when code is applied | inside `set_delegation`, **before** message execution | code **BAL append** after runtime |
| input | execution / `tx_state` | **declared BAL finals** |

That is the same structural pattern codex2 measured on **#10651**: guest derives from **declared finals** where the spec derives from **execution**. Cross-ref **#10612** (BAL obligation / Path A rebuild vs declared; granular scaffold around split containers).

If finals dependency is removed or weakened on the #10651 / one-container track, **this arm may become implementable for free** (move code effect into prepare with a real execution-time source). Whoever works either issue should see the other.

## Gate (do not hybridise)

Retire **only if** the code effect can move into `eip7702_auth_state_prepare` (or the single set_delegation-shaped walk) **with a real finals/execution source**.

**Do not** blend the post-runtime code path into the prepare **gas** path. Hybridising two non-spec arms is worse than either (retire-never-hybridise; #10640 close survivor is prepare-as-set_delegation).

## Not the dead #10640 split

#10640 two-pass was **compute gas vs publish AccountState**. That is **dead**. This walk is **code append from BAL finals** after runtime; nonce already moved to prepare. Do not rename this onto “publish mode.”

## Provenance of discovery

[...]

PR #11235, opened 2026-08-02
fix(codegen): F5 auth condition-4 from state; drop BAL post-state filter

pirapira:
## Summary

**F5 (#10612):** `eip7702_auth_nonstorage_effects` was asking a **post-state BAL question** (membership + finals gates) where the spec asks a **pre-state / latest-state question** of `validate_authorization`.

- **Add condition 4** from state via `code_state_lookup_current` (latest code; header fallback on overlay miss): reject only when code is present **and** not a 23-byte `0xef0100` designator. Same sequential lens as condition 5 nonce (`:120-122` reason).
- **Remove** BAL filter (`bal_find_account` + `bal_account_nonstorage_finals`), declared gates `+40` / redundant `+48` / `+56` / `+72`.
- **Do not** fold preparation vs body-revert scopes; five conditions decide apply; body revert is separate.

Spec pin: `vm/eoa_delegation.py:165-203` @ e5a8caf1b.

## Same-base population (step 1–2) — **before** quoting −2

| leg | base | guest | FA | FR | OK |
|---|---|---|---|---|---|
| parent | `5b6523a00` | `5caf4500…` | 0 | 39 | 1006 |
| sentinel (header code len≠0 → poison `teer_authority[0]`) | same | `19cf671a…` | 0 | 39 | 1006 |
| **category flips** | | | | **0** | |

Scoper’s FR→OK count of 2 was **cross-base** (`9ffb5398a` vs `608b44b16`) and is **not** a same-base measurement. On same-base parent, scoper labels `16975` / `19000` are already **OK**. F5 on this sample is **spec-alignment**, not a −2 FR correctness fix. (Condition 4 still required before removing the BAL filter — removing the filter alone is an FA surface.)

## Fix A/B

| leg | guest | stride-1045 | r200 seed42 |
|---|---|---|---|
| parent `5b6523a00` | `5caf4500…` | FA0 FR39 | FA0 FR4 |
| cand | `f70332a8…` | FA0 FR39 | FA0 FR4 |

ELFs **DISTINCT**. Category flips **0**. Reverse **0**. FA=0 only hard line.

## Layout

`textSizeBytes` `0x628c8 → 0x629b0` (+232 B). check-region-map GREEN.

## Files

- `TxIntrinsicAuthEffects.lean` — C4 + drop BAL gates
- `RegionMap.lean` / `GuestAddrs.lean` / `symbol-addresses.tsv` — layout regen

## Out of scope

- Hybridising code append into prepare gas path (#11234)
- Preparation vs body-revert scope fold
- `aiebh_empty_code_hash` (probe-only; emptiness via length/status)

Refs: #10612 design 5160292445 / sentinel 5160347902 / corrections 5160308479; #10651/#11234 finals-vs-execution shape.


pirapira:
**Reviewed — label on green build.** This closes a BAL dependency with **no spec counterpart**, and the honest handling of a withdrawn number is what makes the rest credible.

⭐⭐ **The finding in one line: the guest asked a post-state question of the BAL where the spec asks a pre-state question of state.** `eip7702_auth_nonstorage_effects` gated effect production on a BAL lookup followed by `bnez` to `.Lteanse_next` — a **filter with no spec counterpart**. ⇒ It *looks* correct precisely because the declared finals describe the end of the block, so a delegation being present at the end **accidentally encodes that the authorisation survived.** A correct-looking proxy for a condition nobody evaluated.

**Four of the five spec conditions were already implemented from state; only condition 4 was missing** (`validate_authorization`, `vm/eoa_delegation.py:165-203`). That turned a suspected rewrite into a one-condition fix.

⭐ **`code_state_lookup_current` rather than the header lens is the load-bearing choice**, and the spec settles it: authorizations are processed **sequentially**, so a later tuple sees an earlier `set_delegation`. The header lens would **reject a second authorisation on an already-delegated authority that the spec accepts** — a false reject introduced by the fix itself. ⇒ And the routine **already makes this exact choice for the nonce**, deliberately, with the reason recorded at `:120-122` — so a reviewer can check the consistency without re-deriving the spec.

⚠️ **The withdrawn −2 is the most important line in this PR.** An earlier sentinel measured 2 rows flipping FR→OK, but **baseline and sentinel were built at different commits** (`9ffb5398a` vs `608b44b16`), with `#11231` landing between them. The same-base re-run gives **parent FA0/FR39, sentinel FA0/FR39, zero category flips** — the population is **0 on this sample**, and 16975/19000 were already OK on the same-base parent. ⇒ *A number quoted across two bases does not merely lose precision — it attributes other people's merges to your change.* Withdrawing it rather than quoting it is correct.

⇒ **So this is spec-alignment, not an FR fix, and the body says so.** That matters: a missing spec condition is a real false-accept surface whether or not tonight's corpus exercises it — the gate was being supplied by a BAL read the spec never performs — but **an unmeasurable improvement must be labelled unmeasurable rather than presented with a null A/B that reads like a pass.**

**A/B**: FA 0 both legs on stride-1045 and random-200, reverse 0, distinct ELFs, `check-region-map` green.

⚠️ And **the two revert scopes were correctly not folded**: a preparation failure rolls back the delegations; a body revert does **not** — `set_delegation` writes persist. The five conditions decide whether the authorisation *applies*; surviving a body revert is a separate property.

PR #11236, opened 2026-08-02
refactor(codegen): share precompile selection and pricing

pirapira:
## Summary

- Share precompile address selection and gas pricing between the root and child routes.
- Keep route-local validation, output handling, success signalling, EIP-150, and continuations in their callers.
- All 18 selectors (0x01..0x11 and 0x100) are owned by the shared helper; the old fixed/word/per-unit/MSM pricing helpers are no longer called.
- The shared descriptor has two provenances: root `ctx+56/+64/+72`; child `x13+inOffset/inSize`.

## Execution-spec alignment

At pinned execution-specs `e5a8caf1b8055e4d805c7fb169edfa710914b7da`:

- `vm/precompiled_contracts/mapping.py:57-75` maps each address to one callable.
- `vm/interpreter.py:273-277` dispatches through that mapping.
- `vm/precompiled_contracts/modexp.py:36` sets `exp_start = 96 + base_length`.

While implementing the shared MODEXP arm I introduced and then fixed a pointer bug: the formula was right, but the exponent scan initially used offset 96 instead of `96 + base_length`. On 00041 that changed the cost from 1536 to the correct 1152. The shared descriptor/predicate is the structural backstop for all 18 arms.

The three former root omissions (ecrecover recovery, MODEXP output computation, and P-256 validity) disappear from the depth-zero implementation; this removes the old depth-zero-unconsumed-output assumption rather than preserving it.

## Evidence

The branch was developed from base `608b44b168683ab7a82ac9aa1565cc6cfaaa0f9d`. The final fixpoint candidate ELF is `a0741faaf4de502d46db4bd6d906aa2b9a97ee85af3c1cd3e96ec837b414906b`. The shared runner is `eaf534e5017e71ffe2dbc17e815646d46ea2ee024f5e8890b7901aaa6ec84d9f`.

A post-merge re-baseline was then run against current main `6b75068fef448e15f7d606e462ca6bc048b80e88` (parent ELF `0f941ec24b001fa0ee5dc753ade17ef7ba1b761658d950c3f8c612b86a3ca2f6`) using the same 200-row manifest:

- Parent: FR=3, OK=197, FA=0.
- Candidate: FR=3, OK=197, FA=0.
- Forward transitions: 0. Reverse transitions: 0.
- 00041 and 00080 are byte-perfect on both current parent and candidate.

The earlier parent-at-608 comparison (FR=5 to candidate FR=3) crossed later main changes. In particular, 15930 is not evidence for this PR: the row has input SHA `2c8c52389b56933260aef47b14540725817d80f079dd27000f2539a48ff4da27`, and the earlier OK result came from a main artifact containing #11226, while the stale parent artifact predated it. The current-main A/B above removes that base drift; no MODEXP attribution is made for 15930.

No synthetic KAT was constructed. The corpus manifest contains named rows for all 18 selector arms; direct route-witness coverage is not claimed beyond the measured target/random evidence.

Generated artifacts were regenerated to fixpoint. `.data=0x5370` and `bnf_le_a=0xa3000ee0` remain invariant; final `.text=0x62280` and `.bss=0x1ac6c920` are reflected in GuestAddrs, RegionMap, GuestImage, and symbol-addresses.tsv.


pirapira:
**Reviewed — label on green build.** All 18 precompile selectors now behind one shared selector-and-pricing helper in both routes, with per-site output and continuation policy preserved.

⭐⭐ **The defect this PR caught in itself is the most valuable thing in it.** The shared MODEXP bit-length scan read the exponent at offset **96**; `modexp.py:36` gives **`exp_start = 96 + base_length`**. On 00041 (`baseLen=2, expLen=2, modLen=64`) that is **1152 correct against 1536 buggy** — and it surfaced as *one false reject in a 200-row sample*. ⇒ **The formula was right and the pointer was wrong**, which is precisely the class a shared helper hides: two callers, one layout, different provenances.

⚠️ **And it was nearly waved through** — FR movement is expected on a structural change, so a single FR against an 18-arm unification reads like refactor churn. Tracing it instead of explaining it is what turned noise into a defect. *Understand-then-ship, not tolerate-then-ship.*

⇒ **The parent recheck is what made that conclusion safe**: 00041 and 00080 both OK on the base, so the FR was **introduced, not pre-existing**. A control that discriminates, rather than a plausible story.

**The structural win is the headline, not the deduplication**: the three former root omissions — ecrecover recovery, modexp output, p256 validity — **disappear as a consequence**, which *retires the depth-zero-unconsumed-output liveness assumption rather than re-arguing it*. Their previous safety rested on output being unconsumed at tx depth, and **consumers change**. ⭐ A change that deletes a standing assumption is worth more than one that tidies.

**Descriptor bases preserved and stated** — root `ctx+56/+64/+72`, child `x13+offset/inSize`. ⇒ That mapping is also the raw material the proof lane needs: **one layout, two provenances** is exactly what a separation predicate over the input descriptor must describe — and such a predicate would have caught the MODEXP error **on all 18 arms at once**, where a corpus row caught it on one by luck.

**A/B**: FA 0 both legs, FR 5→3, reverse 0, targets OK.

⚠️ **One correction to carry**: the parent ELF `0681ab5f` was built at `9ffb5398a`, which **predates #11226** (merged into main at 20:41 via #11231). So the **15930 forward flip belongs to the GASPRICE fix, not to this change** — the author's own note that *the spec wrapper saw no precompile invocation for that CALLCODE row* is what makes the attribution impossible rather than merely doubtful. **13622 remains the genuine flip.** Not a merge blocker; FA=0 is unaffected either way.

pirapira:
**Labelled while CONFLICTING — deliberate.** Base is `608b44b16`; main has since advanced to `6b75068fe` (#11228, #11232 territory). ⇒ **GitHub cannot construct a merge ref for a conflicting branch, so the `build` check will never report** — waiting for green here is waiting for something that cannot arrive. The batch performs the regen and *is* the verified build; that is how #11210, #11218, #11220 and #11226 all landed today.

⚠️ **The conflict is layout, not content**, and it is not the author's to resolve: `GuestAddrs.lean`, `RegionMap.lean`, `symbol-addresses.tsv` and the touched `Programs/` files. An author-side fixpoint converges to a state the next merge invalidates — that has cost three agents a wasted regen today, and one of them twice.

**For the batch**: the behavioural change is the shared selector-and-pricing helper and the MODEXP `96 + base_length` correction; everything else is regeneration the batch recomputes anyway.

⇒ **Landing this matters beyond its own scope**: structural changes are a **prerequisite for starting the separation-logic proofs** — a predicate cannot be written against a shape that is about to change, and every open structural PR is a structure the proof lane would otherwise have to describe twice.

⭐ And this PR leaves the proof lane a concrete gift: **one descriptor layout with two provenances** (root `ctx+56/+64/+72`, child `x13+offset/inSize`). ⇒ *That is the predicate that would have caught the MODEXP pointer error on all 18 arms simultaneously* — where a corpus row caught it on exactly one, by luck. The witness matrix is a secondary concern for precisely that reason: the proofs are the real backstop for wrong-pointer defects, not fixture coverage.

pirapira:
**Correction to the A/B in the body — the FR improvement was base drift, not this change.**

Re-run on current main `6b75068f` with a **rebuilt parent** (`0f941ec2…`) against the same 200-row manifest and the same candidate (`a0741faa…`):

| leg | FR | OK | FA |
|---|---|---|---|
| parent | **3** | 197 | 0 |
| candidate | **3** | 197 | 0 |

**Forward 0, reverse 0.** 00041 and 00080 byte-perfect on both legs.

⇒ So **this PR is FR-neutral**, not FR 5→3. The earlier improvement came entirely from a parent ELF (`0681ab5f`) built at `9ffb5398a`, which **predates #11226** — the GASPRICE fix that landed via #11231. Both "forward flips" were that merge's work: 15930 (the row that invokes **no precompile at all**) and 13622.

⚠️ **A baseline is valid only against its captured base**, and main moved four times between those two builds. This is the fourth measurement to cross a base tonight — a published `FR 46` that was really `41`, a withdrawn population of 2, three stale repins, and now this. **Put the commit in the baseline filename.**

⇒ **This does not change the merge decision.** The structural bar is *structure matches execution-specs, random-200 not deteriorating a lot* — and **zero flips in either direction with FA=0 on both legs is the cleanest possible reading of "not deteriorating".** For a change that re-points all 18 selectors through one shared helper, byte-perfect targets and no corpus movement is a strong result, not a weak one.

⭐ **And it sharpens what the MODEXP finding was**: 00041 is a *targeted* fixture, not a random-200 row. So the pointer defect was caught by a directed witness and **would not have appeared in the sample at all** — which is precisely why a descriptor predicate, not fixture coverage, is the real backstop for wrong-base defects. One layout, two provenances (root `ctx+56/+64/+72`, child `x13+offset/inSize`) is the proof obligation that covers all 18 arms at once.

Issue #11233, opened 2026-08-02
convergence: retire teer_success_table AUTH_BASE first-write compensation (EIP-7702 written_accounts)


pirapira:
## What

`teer_success_table` / `teer_success_count` is a **transaction-local first-write set** used by `eip7702_auth_state_prepare` so **AUTH_BASE** is charged **once per authority per tx**.

Source still labels it:

> COMPENSATION (two-pass guest versus one-pass spec): `teer_success_table` is a transaction-local first-write set only. AccountState owns all cross-transaction state, while this bounded table implements the spec's one ACCOUNT_WRITE charge per authority per tx. It is not a durable nonce overlay; the missing shared transaction overlay is the upstream fix that will make this compensation dead code.

(Comment text still says “two-pass”; the **compute/publish split is already dead** — see #10640 close. The table remains for **AUTH_BASE / written_accounts**, not for a second pass.)

## Spec counterpart

execution-specs `set_delegation` / `written_accounts`: one ACCOUNT_WRITE (and AUTH_BASE when applicable) per authority per tx via the **same** `tx_state` the rest of the walk mutates. No separate bounded table.

## Why not part of #10640

#10640’s premise (two-pass compute/publish + three compensations including a **running nonce** table) is **dead**. This table is a **different** survivor: first-write AUTH_BASE dedup, not the old C1 nonce overlay.

## Removal condition

Shared transaction-local “already written this authority” state that prepare and the rest of the tx path both read (r59nm / container convergence family — related spirit to #10619). Do **not** delete the table while AUTH_BASE can double-charge.

## Provenance of discovery

Re-derivation for #10640 on main `5b6523a00`; PCs measured on guest `0681ab5f` @ `9ffb5398a` (layout-only delta to main; 7702 source unchanged). `teer_success_table` live (~33 KiB BSS, refs>0).

## Out of scope

- Reopening two-pass compute/publish (#10640 closed)
- Post-runtime code append (S3 / sibling issue)
- Dual call-site a4 packaging (lane packaging; both live by skip-test)

pirapira:
## S1 re-derive on current main — **STAY** (not a deletion)

**Main** `6b75068fe` · **guest** sha256 `07f36a6d7f778fc15b06d969f8dc929d375a296e2012404db3b6eae5bbe8d097`

Addresses (current main — do not quote 0681ab5f):

| symbol | addr | size |
|---|---|---|
[...]

Issue #11222, opened 2026-08-02
baap_storage_values is emitted at three different sizes in three build units (6,400,000 / 3,840,000 / 32,768)


pirapira:
## `baap_storage_values` is emitted at **three different sizes** in three build units

Found while producing the separation-logic description for #11186. Measurement only; no edits.

| Lean unit | emitted size | derivation |
|---|---|---|
| `BlockVerdictDataSectionTail.lean:76` | **6,400,000 B** | `bsrMaxBalItems (100,000) × bsrPathBytes (64)` — this is the `stateless_guest` image |
| `BalAccountApplyPostFields.lean:826` | **3,840,000 B** | literal `.zero 3840000` |
| `BalAccountDescriptorArray.lean:394` | **32,768 B** | literal `.zero 32768` |

All three emit the **same symbol name** — `baap_storage_values:` — into whichever program links them.

### Why this is not a documentation problem

**Either** three programs each emit their own `baap_storage_values` and the name is overloaded across images — in which case **every claim about this symbol must name its build unit**, and `RegionMap` describes exactly one of the three — **or** one of these is dead and nobody noticed the disagreement.

⇒ Either way it matters now, because the proof lane is about to write **separation-logic predicates** over these buffers. A predicate stating `extent = capacity × stride` with `capacity = 100,000` and `stride = 64` is **true in the `stateless_guest` image and false in the other two**, and nothing in the predicate's statement would say so. **A symbol whose size depends on which unit you read is exactly what makes a predicate silently false somewhere.**

### What the `stateless_guest` value is used for, for scoping

In that image the buffer is a union child of `call_frame_arena` (base `0xb0e9eca0` this build), written by `bal_account_apply_post_fields` and `bsr_beacon_change` through a **byte-pointer** cursor `baap_storage_value_cursor`, and its 6,400,000 B satisfies `100,000 × 64` exactly.

### Closing conditions

1. **Decide which it is** — overloaded-by-design, or one of the three is dead. `check-unimported.sh` proves the module graph has no orphans, so a dead *definition* inside a live module would not be caught by it.
2. If overloaded by design: **say so at each definition site**, and make `RegionMap`'s entry state which unit it describes.
3. If one is dead: remove it, and note that a `.zero` in a live module is not name-unreferenced, so the R8 caveat applies — **a data buffer cannot be proven dead from name references** (GH #11172's `bal_gas_valid` is the precedent).
[...]

PR #11237, opened 2026-08-02
docs: predicate shapes for structures that stay (proof-lane input)

pirapira:
## Summary

Docs-only proof-lane input: **predicate shapes for structures that stay** (the other half of “landing structural changes is a prerequisite for separation-logic proofs”).

New: [`docs/4ch8f-predicate-shapes.md`](docs/4ch8f-predicate-shapes.md)  
Link from [`docs/4ch8f-region-map.md`](docs/4ch8f-region-map.md) predicate-readiness section.

### Template (uniform)

1. BASE, EXTENT, capacity×stride **exact**?
2. Entry fields (offset/width)
3. Count vs high-water mark
4. Validity (or **nothing**)
5. Ownership / lifetime
6. Disjointness / **union children**
7. **Slot nine** — init contract (zero on entry to which routine)
8. **Build unit** (predicates without a unit are false somewhere — `#11222`)

**Base is an input, never a literal.**

### Structures covered

| structure | note |
|---|---|
| `call_frame_arena` | `slot(p)=base+p×0x19000`, **p = parent depth**; spare slot 1024 |
| union children | exact tiling; phase alias with frames |
| `baap_storage_values` | byte cursor; **#11222 three sizes = unclean finding** |
| `evm_memory_pool` | 96 MiB guest; probe unit differs |
| `callee_seed_table` | 128×96 exact; F3 live until `#10651` |
| `teer_success_table` | STAY `#11233`; EXTCODEHASH second consumer |
| DJ-* journals | points at `4ch8f-dispatch-journal-initialization.md` |

Unclean structures named as **findings** (must change or unit-qualify before proof), not gaps.

### Named limits

- Composed from prior measurements (scoper `#11186`, `#11222`, `#11227`, `#11233`, F3 SPIKE) — not a full re-census of every RegionMap row.
- Illustrative guest sha `07f36a6d…` @ main `6b75068fe` for addresses only; predicates must not bake them.
- Does not resolve `#11222` (decide overload vs dead).

No code. No merge! from grok.


pirapira:
**Reviewed — label on green build.** Docs-only, `+226/−0`, and it is the second half of the prerequisite: **structural changes must land before proofs can start, and the proof lane also needs the shapes of the structures that are *not* changing.** Those can be worked on immediately.

⭐ **The template is uniform across all six**, which is what makes it usable rather than six essays: base and extent with `capacity × stride` checked against the extent exactly, the entry layout field by field, whether the count is **live entries or a high-water mark**, the validity invariant, ownership and lifetime, disjointness, the **initialisation contract** (slot 9), and the **build unit each predicate describes**.

⇒ **That last field is not bookkeeping.** `baap_storage_values` is emitted at **three different sizes in three build units** (#11222), so a predicate that does not name its unit is false in at least two of them. Same reason a phantom symbol survived in `CallFramePhase` until tonight — *a symbol exists per image, not per repository.*

⚠️ **Base as an input, never a literal, is the rule that keeps these from rotting.** `slot(p) = arena_base + p × 0x19000` is a theorem about any image; the same with `0xaf420780` baked in is a claim nobody can check. ⇒ **Three different arena bases circulated tonight and two of them were correct** — for their own images. A link-dependent address has no single true value.

⭐ **And recording the unclean structures as findings rather than gaps is the most valuable part.** A structure that cannot carry a clean predicate names something that must **change before anyone proves anything about it** — that is a work item, not an omission. The two known cases are worth more than the four clean ones:

- **`baap_storage_values`** — variable-length entries, so the honest predicate today is **byte-range only**; an entry-indexed one needs an alignment invariant the structure does not have. And it is a **union child**, so naive separation fails: the same bytes are frame slots while frames live and replay values while replay runs. **The predicate must be phase-indexed, and the phases proven non-overlapping in time — a control-flow obligation, not a spatial one.**
- **`evm_memory_pool`** — the contrast case, and the one to copy: **disjointness falls out of the address formula**, distinct depth giving distinct slot by arithmetic, with no invariant for anyone to maintain, no free-list reasoning and no ordering side condition. ⇒ *Two regions, two completely different proof costs, and now we know which is which before anyone starts.*

Issue #11019, opened 2026-07-31
codegen: balance_at_header_state_root returns the LIVE balance when one exists — the name misdescribes it and cost an hour plus a nearly-landed unnecessary patch


pirapira:
`balance_at_header_state_root` (`EvmOpcodes.lean`) does **not** read header state when a live balance exists. It pads the address, calls **`account_state_latest_balance` first**, and on a hit **writes the live post balance and returns** — before `header_extract_state_root` / `account_at_address` are reached. Header state is the *fallback*, not the behaviour.

## What the name cost, concretely

Diagnosing the MTx coinbase fee path, a source reading concluded the guest "reconstructs every coinbase post from header state, so a later fee overwrites rather than accumulates" — a coherent reading of `BlockVerdictMtxCoinbase:33-43` **given the callee's name.** It was wrong. `record_nonstorage_effect` publishes each fee into AccountState, so tx1 falls back to header and tx2..5 take the helper's live path, and the values accumulate correctly.

**Cost:** roughly an hour of tracing, and a patch that was written, built, measured and nearly labelled before the artefact refuted it. Evidence it was never needed — the rebuilt BAL for `00317` (5 tx, priority fees 3,3,5,5,5, gas increments `0x1ec75`/`0x1fd2d`/`0x1ec75`/`0x1ec75`/`0x2bedc`) carries coinbase `2adc2566…` with balance rows BAI 1..5 = **378207, 769254, 1399599, 2029944, 2929604**, matching the declared BAL exactly, and parent/candidate streams are byte-identical at 707 bytes.

## Why this is worth fixing rather than remembering

A caller cannot tell from the call site that this returns live state. Anyone auditing balance provenance — which is most of the current BAL work — will read `balance_at_header_state_root` and conclude the value is header state. **That is the opposite of what it usually is.**

Either is fine:

1. **Rename** to reflect live-first behaviour (`balance_live_else_header_state_root`, or similar), or
2. **Document at the definition** that `account_state_latest_balance` is consulted first and header state is only the fallback — with the reason, so the next reader does not "fix" the ordering.

⭐ Option 2 is cheaper and byte-identical; option 1 is what stops the misreading recurring at a glance.

## This is a pattern here, not an isolated name

Same session, same class:

| name | actual |
|---|---|
| `bv_pending_upfront_sender_post` | neither pending nor upfront — holds a **post-refund** quantity |
| `single_tx_creation_runtime` | called from the **multi-tx** path |
[...]

Issue #11238, opened 2026-08-02
codegen: rename balance_at_header_state_root → balance_live_else_header_state_root


pirapira:
## Why

#11019 established that `balance_at_header_state_root` is **live-first** (correct behaviour, lying name). Docs now admit that (byte-identical PR). The better name is still:

**`balance_live_else_header_state_root`**

A symbol rename changes emitted assembly and forces layout regen (GuestAddrs / RegionMap / TSV). Deferred while pin surface is contended. Do this on a quiet base.

## Proposed name

`balance_live_else_header_state_root` — matches actual control flow:
1. `account_state_latest_balance` hit → return
2. miss → header_extract_state_root + account_at_address

## Callers to update (production guest jals, main `6b75068fe`)

| site | file |
|---|---|
| Mtx coinbase fee | `BlockVerdictMtxCoinbase.lean` |
| h_BALANCE | `EvmBalance.lean` |
| SELFDESTRUCT surcharge | `Selfdestruct.lean` |
| CREATE alive | `ChildFrameCreateTail.lean` |
| CREATE2 alive | same (templated) |
| probe / registry | `EvmOpcodes.lean` probe, `Registry*.lean` names |

Also string label inside `balanceAtHeaderStateRootFunction`, GuestAddrs entry, symbol-addresses.tsv, any docs.

## Non-callers (do not route here by accident)

- DispatchTx SELFBALANCE staging — deliberately avoids this helper
- Pure header needs → `account_at_header_state_root` (no pure balance-header helper exists)

## Related

- #11019 (naming defect; docs fix lands separately)
- Contrast: `account_at_header_state_root` is true header-only

PR #11239, opened 2026-08-02
docs(codegen): admit live-first balance_at_header_state_root (#11019)

pirapira:
## Summary

Byte-identical docs fix for #11019. `balance_at_header_state_root` **returns live balance first** (`account_state_latest_balance`); header path is miss-only. The code is correct for every production caller; the **name and docstring lied**.

**No rename tonight** (regen / pin surface). Rename filed separately.

## Changes (prose only — no emitted bytes)

1. **`EvmOpcodes.lean`** module header + body comment: admit LIVE-FIRST; cite Amsterdam one mutable `TransactionState` (`environment.py` BALANCE, `state_tracker.py` `is_account_alive`); **explicit contrast**: `account_at_header_state_root` is true header-only with many callers; a pure **balance**-header-only helper does **not** exist; DispatchTx deliberately avoids this helper for SELFBALANCE staging.
2. **`CallBalanceGate.lean`**: gate uses **env+32**, not this helper (drj99.1).
3. **`ChildFrameCreateTail.lean`**: CREATE alive = `is_account_alive` on live tx_state, not pure header.
4. **`EvmBalance.lean`**: pointer comments.

## Evidence (close-out census)

Main `6b75068fe` guest `07f36a6d…`: emitted `@0x8003570c` jals `account_state_latest_balance` first.

| site | wants | match |
|---|---|---|
| Mtx coinbase fee | LIVE accumulate | yes |
| h_BALANCE | LIVE | yes (spec) |
| SD surcharge origin | LIVE nonzero | yes |
| CREATE/CREATE2 alive | `is_account_alive` live | yes (comment was wrong) |

Not an FR producer by itself. Wrong-stage trap excluded on all five.

## Test plan

- [x] Diff is comment/docstring only (no `.s` / layout change)
- [ ] CI green (docs-only Lean still elaborates)

### DIFF
--- a/EvmAsm/Codegen/Programs/CallBalanceGate.lean
+++ b/EvmAsm/Codegen/Programs/CallBalanceGate.lean
@@ -7,10 +7,11 @@
 
   The gate (`callDescendFallThrough`, ChildFrameHandlers.lean) rejects a
   value-bearing CALL/CALLCODE — pushing 0 and NOT descending — when the caller's
-  balance (looked up via `balance_at_header_state_root` against the attached
-  header/state witness) is below the transfer value. This probe drives the live
-  `callFrameGuestRegistry` h_CALL handler (with the gate wired in, unlike the
-  push-0 `callFrameProbeRegistry`) so the rejection is exercised end-to-end.
+  **live** balance is below the transfer value. The handlers read env+32
+  (`.selfBalance`), not `balance_at_header_state_root` (drj99.1; #11019).
+  This probe drives the live `callFrameGuestRegistry` h_CALL handler (with the
+  gate wired in, unlike the push-0 `callFrameProbeRegistry`) so the rejection
+  is exercised end-to-end.
 
   Setup (witness supplied at runtime via `ziskemu -i`, built by
   codegen-zisk-call-balance-gate-check.sh, reusing the balance-at-header-state-root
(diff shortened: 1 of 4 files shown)


pirapira:
**Reviewed — label on green build.** Docs-only, `+55/−32`, byte-identical, no regen. Exactly the right shape for tonight: three branches are already conflicting on `GuestAddrs`/`RegionMap`, and this cannot join that queue.

⭐ **The finding is that the code is right and the prose lies.** `balance_at_header_state_root` returns **live first**, falling back to the header state only on a miss — and that is *correct*, because execution-specs has one mutable `TransactionState` where BALANCE and `is_account_alive` read the last write. **A naming defect, not a wrong-value defect.** The caller census confirms it: no production caller wants pure header-only from this symbol, and the wrong-stage trap was checked at all five sites and not observed.

⚠️ **A name is not a contract — and `at_header_state_root` in a symbol name is precisely the assurance that stops the next reader checking.** That is the seventh instance in this repo and the fifth today: a routine whose name promises one provenance and delivers another, indistinguishable from correct until someone reads the emitted form.

⇒ **The stale comments were the actively harmful part**, and removing them is most of the value: `CallBalanceGate` claimed the gate uses this helper when the handlers read `env+32`, and the CREATE alive comment claimed header. ⇒ *Prose that misdirects is worse than prose that is merely absent* — it answers the question the reader was about to ask, wrongly.

⭐ **The contrast sentence is the one to keep**: `account_at_header_state_root` **is** a true header-only source with many callers, while a pure balance-header-only helper **does not exist**. So anyone who genuinely needs pre-state balance has to build it — where today the name implies it is already there. That is the sentence that prevents the next wrong assumption.

⇒ And the awareness already existed in the code without being written where the name could be read: **DispatchTx deliberately avoids this helper for SELFBALANCE staging**, commenting that the live log may already hold settlement. One author knew; the name told everyone else otherwise.

**The rename is correctly deferred** to #11238 with the caller list — `balance_live_else_header_state_root` is the better name, but a symbol rename changes emitted assembly and forces a layout regen into a contended pin surface. **Landing beats polishing, and a rename is polishing with a regen attached.**

PR #11242, opened 2026-08-02
fix(storage): admit status-1 as absent_account counter (#11162)

### LINKED ISSUE #11162: #11138's tier-3 classifier over-reports: status 1 (account not in state trie) is spec-legitimate but counted as an integrity failure

pirapira:
## Summary

Fix tier-3 SSTORE/SLOAD witness status classifier over-report (#11162).

`slot_at_header_state_root` status **1** = account not in state trie. Spec
`witness_state.py` flattens that to EMPTY_TRIE_ROOT → U256(0) — a **legitimate
zero**, not an integrity failure. The classifier only admitted status **5**
(slot absent) as `sstore_witness_absent`; status 1 fell into
`sstore_witness_integrity_fail`, inflating that diagnostic counter with fine
rows.

### Over-report shape (load-bearing for census)

**False positives counting fine rows — not merged FR causes.** Real faults
{2,3,4,6,7} sharing one integrity bucket is by design; only the magnitude of
`integrity_fail` was wrong. A 26k census that keys on `integrity_fail>0` is
**inflated but not corrupted** in class structure. Had it been merged causes,
the census would need re-running.

### Fix

1. Admit status **1** as legitimate absence.
2. Give it its **own** counter `sstore_witness_absent_account` — do **not** fold
   1 into 5 (that would fix one over-report by creating a smaller one).
3. status **5** stays `sstore_witness_absent` (slot).
4. {2,3,4,6,7} stay `sstore_witness_integrity_fail` + `last_status`.

### Verdict is byte-identical (no consumer)

These counters have **no verdict consumer** (#11138 policy unset). Both arms
still fall to `_prestate_zero`. Resolved storage values and OUTPUT are unchanged.

| leg | guest sha256 | r200 FA/FR/OK |
|---|---|---|
| parent `abc0908f3` | `c5937eaf95c898de…` | 0/4/196 |
| cand | `508c0a880d6d7b3e…` | 0/4/196 |

- ELFs **DISTINCT**
- FR sets **EQUAL**
- category flips **0**, reverse **0**
- OUTPUT 256B **byte-identical 200/200**

### Force-fire (required)

SPIKE_WATCH on `sstore_0to_x` pair (main parent addrs vs cand):

| row | parent | cand |
|---|---|---|
| **d2-g0** `19722` | absent=16, **integrity=1**, last_status=1 | absent=16, **absent_account=1**, integrity=0, last=0 |
| **d0-g0** `19718` | absent=17, integrity=0 | absent=17, absent_account=0, integrity=0 |

Misclassified case now classifies correctly; control stays put.

### Not dissolved by tonight merges

Re-derived after GASPRICE / MODEXP / F5 (#11232 base): classifier still
`li 5; bne integrity`. Premise HOLDS.

### Layout

- textSizeBytes `0x62690→0x626d8` (+72B classifier)
- bssSizeBytes `0x1ac6c820→0x1ac6c840` (+32B incl. 8B counter + align)
- check-region-map GREEN
- GuestAddrs + TSV regen

### Files

- `Storage.lean` — classifier
- `BlockVerdictDataSectionTail.lean` — BSS counter + docs
- RegionMap / GuestImage / GuestAddrs / TSV

Closes #11162


pirapira:
**Reviewed — label on green build.** An instrument fix, and it landed ahead of the measurement that depends on it, which is the part that matters.

⭐ **The over-report was false positives, not merged causes** — and that distinction decided whether a 26,104-row census had to be thrown away. Status 1 is *account-not-in-state-trie*, which `witness_state.py` maps to `EMPTY_TRIE_ROOT` and `U256 0`: **a spec-legitimate zero, not an integrity failure.** The classifier admitted only status 5 as absent, so every status-1 row inflated `integrity_fail`. ⇒ Real faults `{2,3,4,6,7}` sharing one bucket is **by design**, so the class structure underneath was intact — had it been merged causes, the census would have needed re-running.

⭐ **A separate `sstore_witness_absent_account` counter rather than folding 1 into 5.** ⇒ *You do not fix an over-report by creating a smaller one* — folding would have removed the false positive while leaving absent-account and absent-slot indistinguishable, which is the same defect one level down. A legitimate-absence count is information we were discarding.

**Verdict byte-identical, and shown rather than asserted**: parent `c5937eaf` vs candidate `508c0a88`, r200 **FA 0 / FR 4 on both, FR sets equal, zero flips, reverse 0**, OUTPUT 256 B identical across **200/200** — with **ELFs distinct**, which is what makes the identity non-vacuous. ⚠️ *A byte-identity claim whose two sides might be the same build proves nothing*, and that trap produced three false passes in one evening here.

⭐ **And the force-fired probe is what turns this from plausible into verified**: 19722 moves `integrity=1 → absent_account`, while 19718 stays put. ⇒ *A classifier fix with no demonstration that the misclassified case now classifies correctly — and that the correctly-classified case does not move — is unverified.* Two rows, opposite expectations, both confirmed.

**Not dissolved by tonight's merges** — checked explicitly against the GASPRICE fix, the MODEXP correction and F5. That check has killed thirteen premises today; worth recording the times it comes back negative.

⇒ Sequencing note for the record: this had to precede acting on the full-corpus census, because **a top class chosen from an inflated counter sends the next hours of work at a non-defect.** A wrong instrument at the top of a funnel outranks anything beneath it.

PR #11246, opened 2026-08-02
codegen: enumerate block state from execution maps

pirapira:
## Summary

Wires #10651's block-state candidate-set authority to the execution-derived maps.

- Enumerates the union of `account_writes` and `storage_writes` with map-capacity bounds.
- Keeps BAL account addresses as a cross-check, while visiting map-only accounts and storage owners.
- Reuses the existing account-path/MPT replay machinery through a synthetic empty-field AccountChanges item.
- Uses account-map nonce, balance, code, and storage-root fields for map-account replay; storage-map rows drive storage replay.
- Preserves the BAL emission order, early pre-account materialization, and access logs.
- Regenerated linked address/layout artifacts are unchanged in this follow-up; the batch owns the conflicting generated layout.

This follow-up carries three verified guards:

1. Seed modeled EIP-2935/EIP-4788 system owners into the applied-owner set to avoid double-applying their storage rows.
2. Retain storage-map rows only when the execution value differs from the transaction-start baseline.
3. Retain account rows only when the complete pre/post account RLP differs.

## Verification

- `lake build` — pass (4796 jobs)
- `scripts/check-asm-to-program.sh` — pass
- `NO_BUILD=1 scripts/check-region-map.sh` — pass
- `python3 scripts/asm_to_program.py check-guest-addrs` — pass
- `scripts/check-forbidden-tactics.sh` — pass
- `scripts/check-file-size.sh` — pass
- Same-manifest 200-row sweep against the named parent: candidate `full=195`, `root=200`, `succ=195`, `tail=200`, `FA=0`, `FR=5`; parent `FR=4`; `87` previously failing rows fixed and `0` regressions.

The source audit resolves the previously listed per-BAI concern. The #10651
filter is in `execution_map_state_changes` (`BlockVerdictStateRoot.lean`),
which is called only from `block_state_root` and enumerates the collapsed
execution maps for final-state replay. It does not read or filter the BAL
builder stream. `write_sets_incorporate_tx` calls `bal_emit_storage_changes`
before merging the transaction map, passes the current BAI, and clears the
transaction map only afterward. `bal_builder_record_storage_change` retains
rows by `(address, slot, BAI)`. Thus a slot changing BAI1=1 then BAI2=0
produces two BAL builder rows, as required by the spec; the root map correctly
retains only the final state. The former "per-BAI granularity is not preserved"
limitation is downgraded to a non-issue, not a producer gap.

The remaining known limitation is explicit:

- A duplicate row can win a reverse scan and supply a false baseline, so any pre != post filter is only as good as the row supplying `pre`. This was reproduced on fixture 00600, where eager seed row7 masks the real prior row0.

This limitation does not block this PR; it is recorded as a follow-up boundary.


pirapira:
## Independent verify (grok) — delta filter inputs

On `stack_underflow SWAP3` cand guest from this PR:

- Live counts 2/18 (not capacity)
- All 18 storage rows = system predeploys
- **16 NETZERO** (7002/7251/0d8282/0e8282 × keys 0–3, val==baseline==0)
- **2 CHANGED** controls: 2935, 4788 — must survive

Fix shape per coord: **mirror `bal_emit_storage_changes`** (pre!=post), not a second delta test, not a system-address skip.

**Same mechanism family as code44** (guest records rows spec never produces): there nonce-only auth nonstorage; here value-unchanged system storage. Pair should be visible in both PR bodies when code44 lands.

Standing by to re-measure this row after the delta filter lands (expect ~2 survivors, FA=0, FR near parent 4/200).

pirapira:
## bsr_changes application path (grok) — account side after storage REFUTED

Storage delta guard fires (18→2) and stays; FR still 194. Cause is account application.

### Order in `block_state_root`
1. BAL replay → `bal_account_change_descriptor` → `bsr_changes` + `bsr_changed_accounts` (mode≠3 only)
2. `execution_map_state_changes` — append **map-only** accounts (skip if addr ∈ `bsr_changed_accounts`)
3. withdrawals (skipped when BAL present)
4. `mpt_bounded_state_root(bsr_changes)`

### Map-only account path (`.Lem_account_process` → `.Lem_process_address`)
| step | what |
|---|---|
| base | `mpt_walk` pre-state account RLP (full leaf) — good |
| synthetic BAL item | `0xda 0x94 <addr> c0×5` — **all field lists empty** |
| flag | `bsr_account_from_map=1`, `bsr_account_row` → block `account_writes` row @ `0xa28a0000` |
| apply | `bal_account_change_value` → `bal_account_apply_post_fields` |

### Field fill when `bsr_account_from_map` (mask @ row+112)
| bit | field | source |
|---|---|---|
| 1 | balance | row+32..56 **copied as four `sd` limbs into `baap_map_value_be`, then leading-zero strip as if BE** |
| 2 | nonce | row+64 u64 → 8-byte buffer, strip |
| 4 | codeHash | keccak(code_ptr@+80, len@+88) → account field 3 |
| else | inherit | keep pre-state component |

### ★ Defect hypothesis (path, not dump)
**Balance endian / encoding mismatch vs BAL path.**

- `account_write_record` stores balance as **raw 32B from caller pointer** (callers pass LE limb layout from nonstorage / env).
- Map apply treats that buffer as **big-endian integer** (`baap_map_value_be` strip + `account_set_uint_field`).
- BAL path applies **RLP BE integers** from declared BAL post-fields.

Same address can therefore get:
- BAL leaf: correct BE balance post
- **or**, if skip misses and map re-applies: **byte-reversed / wrong magnitude balance**
- **or** map-only account never in BAL: wrong leaf while parent used BAL

Skip table is BE-from-BAL RLP addr vs map row addr (also byte-copied from write key) — endian match likely OK; **value encoding is the shape gap**.

### What codex2 dump should show (sender + coinbase)
For each of `f6c3…` and `2adc…`:
1. In `bsr_changed_accounts` after BAL? (if no → map will process)
2. Final leaf in `bsr_changes` / `basr_values`: balance bytes vs parent BAL-derived leaf
3. Map row mask + raw +32..56 limbs

If balances differ byte-for-byte while nonce/codeHash/storageRoot match → this path bug is confirmed.

### Keep
Delta storage guard (pre@+96 != post@+64). Correct, not the FR cause.

### Not claiming without dump
Whether 00004 skip hits or misses for the two accounts — that is the dump. Path analysis says **wrong balance encoding on map apply** is the first thing the leaf diff should show.

pirapira:
### Correction: balance endian weaker than first claimed

`record_nonstorage_effect` / message value path store **32B BE** into `account_write_record`. Map path `ld`/`sd` of four qwords **preserves** that byte sequence into `baap_map_value_be`; strip-as-BE is then correct for those producers.

**Remaining stronger hypotheses for leaf dump:**
1. **Skip miss → double path** — same address in BAL `bsr_changes` and again from map (changed_accounts miss: mode=3, cap 64, or compare bug)
2. **Map-only** account that parent never applied (BAL omitted) with wrong/incomplete mask vs true post-state
3. **A non-BE writer** into `account_write_record` (if any gas/fee path passes LE limbs)

Dump still: parent vs cand leaf bytes for sender+coinbase; whether each is in `bsr_changed_accounts` before map.

pirapira:
## Double-apply measurement CONFIRMED (00004 SWAP3, filtered guest f69613b6)

Instrument: SPIKE at `jal execution_map_state_changes` `0x80014568` (pre) and `+4` (post).

| moment | `bsr_change_count` | `bsr_changed_account_count` | who in skip set |
|---|---:|---:|---|
| **pre-map** | **4** | 2 | coinbase `2adc…`, sender `f6c3…` only |
| **post-map** | **6** | 2 | same 2 — **2935/4788 never added** |

### Two writes per system account (path byte-identical)
| idx | phase | path source | path64 head | vlen | notes |
|---:|---|---|---|---:|---|
| 0 | modeled `bsr_sys_change` | `bsr_paths` | `060c090d0507…` | 70 | **2935** |
| 1 | modeled `bsr_beacon_change` | `bsr_paths` | `03070d06050e…` | 70 | **4788** |
| 2–3 | BAL normal | `basr_paths` | … | 74/82 | sender/coinbase |
| **4** | **map append** | `basr_paths` | **`060c090d0507…` = idx0** | 70 | **2nd write 2935** |
| **5** | **map append** | `basr_paths` | **`03070d06050e…` = idx1** | 70 | **2nd write 4788** |

`path0==path4` and `path1==path5` **True**.

### Leaf RLP on this fixture
sys0 and map4 account RLP **byte-identical**; sys1 and map5 **byte-identical** (both `f84401…` shape). So the damage is **duplicate path entries into `mpt_bounded_state_root`**, not necessarily a partial-mask corruption of the value on this row. Duplicate paths still explain fail 130 / wrong root even when values match.

### Why skip missed
BAL modeled system arm (`.Lbsr_bal_copy_system2935/4788`) calls `bsr_apply_modeled_system_post_fields` and **does not** record into `bsr_changed_accounts`. Early `bsr_sys_change` / `bsr_beacon_change` also do not. Map skip set = normal BAL accounts only → 2935/4788 always map-processed when their storage rows survive the delta filter.

### Fix shape (do not remove map apply)
When modeled system owners are committed (sys_has_2935 / sys_has_4788 path that writes desc 0/1), **also append `bsr_addr_2935` / `bsr_addr_4788` to `bsr_changed_accounts`**. Map then skips; enumeration remains authority for true map-only accounts.

Delta storage guard stays. Separate fix; measure FR after this alone.

Artifacts: `/tmp/grok-10651-double/`.

pirapira:
## Handoff: derived skip (no address literals) — codex2 implements

Coord accepted double-apply evidence. **Grok does not implement.**

### Forbidden
Hardcoding `bsr_addr_2935` / `bsr_addr_4788` into the skip set (same class deleted in #11218).

### Required shape
When modeled system commit writes a leaf (`bsr_sys_change` / `bsr_beacon_change`), **record that owner into `bsr_changed_accounts`** — same structure BAL normal already uses. Map storage path already skips on membership (`.Lem_storage_bal_scan` → `.Lem_storage_next`).

### Pitfall (from path read)
`block_state_root` **zeros** `bsr_changed_account_count` at BAL-section start (after sys commits). Seed must either:
- happen **after** that zero and **after** BAL normal fills (before `jal execution_map_state_changes`), from owners already committed in `bsr_changes[0..s1)` sys slots, or
- survive the zero (separate already-applied set), or
- re-derive from the sys change descriptors’ account owners without naming precompile constants.

One rule with sender/coinbase: already-applied ⇒ map skips.

### Control before any sweep
| check | want |
|---|---|
| 00004 post-map `bsr_change_count` | **4** (not 6) |
| system leaves 2935+4788 | **still present, once each** |
| if count drops to **2** | skip too broad — lost system rows |

Then r200: expect FR ≈ parent 4, FA=0.

### Second-best
path64 dedupe at `bsr_changes` append / MPT insert — address-free, future-proof.

Delta storage guard stays (separate, correct).

pirapira:
## Coord caveat on delta guard justification (no code from grok)

Reconciled evidence on #10839 (00600 EIP-7002): BAL can have **bai1 post=1, bai2 post=0** on one slot — two rows; final net zero. Guest final-only `pre!=post` drops that.

A storage delta guard that filters **final** pre≠post is still right for true single-write unchanged slots, but **must not be justified solely as “matches spec net-zero omission”** when the spec emits **per-BAI** history. Before label: confirm whether the map is per-BAI or final-only; filter placement follows.

Full write-up: https://github.com/Verified-zkEVM/evm-asm/issues/10839#issuecomment-5161250639

pirapira:
## #11246 guard caveat from settled 00600 answer (coord → codex2)

Full mechanism on #10839. Narrow note for this PR only:

If a **duplicate row can win a reverse scan and supply a false baseline**, then any `pre != post` filter is only as good as **which row supplies `pre`**.

- **Not** a reason to remove the final-pair delta guard (still correct for true single unchanged writes; matches `block_access_lists.py:667-676`).
- **Is** a reason to **check which row supplies `pre` before this PR lands**.
- Leave for tomorrow — no further work tonight.

00600 evidence: earlier row0 baseline=1 masked by later seeded-zero duplicate at row7 under reverse scan; system write and seed value for key1 are both honestly 0.

pirapira:
**Labelled — the map-authority switch lands, with three defects found and fixed inside it.**

⭐⭐ **`root` matches 200/200 and `tail` 200/200.** Enumerating the terminal root from the execution maps rather than the declared BAL produces a correct root on every row. ⇒ *That is #10651's core claim demonstrated* — the guest no longer draws its candidate set from the artefact it is validating.

**FR 194 → 5 against a parent of 4, with 87 fixed and 0 regressions**, FA=0 on both legs, same manifest, base named. ⚠️ **Zero regressions is the number that matters**, because a net figure would have hidden the opposite: a guard that improves the count while breaking rows that previously passed looks identical on FR alone. Both directions were reported.

⇒ **Three distinct defects, each found by a control on a single fixture rather than by a sweep:**

| defect | evidence |
|---|---|
| **double-apply** | `change_count` 4→6; `desc0`/`desc4` and `desc1`/`desc5` sharing a `path64`; leaf RLP **identical**, so the damage was the duplicate path, not corruption |
| **storage delta gap** | 18 rows on 00004, only 2 with `post != baseline`; the other 16 are empty request queues rewriting slots unchanged, which `block_access_lists.py:667-676` never turns into a change |
| **account delta gap** | the candidate applied a **fourth** account-trie leaf — the coinbase — that the parent and the fixture both omit |

⭐ The coinbase was identified two independent ways within a minute: read from the emit dump as `2adc2566…`, and computed as `keccak256` over the candidate-only `path64` (`9d860e7b…`). ⇒ *Two methods, one answer, no shared assumption.*

⛔ **The system-owner skip is address-free** — owners are recorded as applied when the modelled system commits write, rather than by naming 2935/4788 as constants. A hardcoded address list is precisely the shape deleted from the skip lists in #11218 hours earlier, and it would break the moment a fork adds a predeploy.

⚠️ **Two named limitations, neither blocking, both recorded so they are not discovered as regressions later:**

- **Per-BAI granularity is not preserved.** A slot can go `BAI1=1` then `BAI2=0`; the spec emits **two** rows while a final `pre`/`post` filter sees net zero and drops both.
- **A duplicate row can win a reverse scan and supply a false baseline** — so any `pre != post` filter is only as good as the row it read `pre` from. Demonstrated on 00600, where a seeded zero at row 7 masks row 0's baseline of 1.

⇒ CONFLICTING is expected and the batch owns the layout regen; author-side fixpoints cost four wasted regens tonight.

Issue #11245, opened 2026-08-02
convergence: verify the BAL by hash (wire bal_serializer_verify), retiring the field-by-field stand-ins


pirapira:
## What

Wire `bal_serializer_verify` into the verdict: **compare the hash of the rebuilt (execution-derived) BAL against the supplied BAL's hash**, and retire the field-by-field consistency checks that stand in for it today.

⏭️ **Sequenced after #10651.** Do not start this until the map-authority switch has landed — see *Why the order matters* below.

## Current state, measured

**What is wired:** `BlockVerdictStateRoot.lean:776-778` calls `block_access_list_hash` over the **supplied** BAL into `svf_bal_hash`, passed at `sv_params+96` into `block_verdict`. ⇒ That checks the supplied BAL against header field 21 — *it confirms the BAL we were handed matches the header, and says nothing about whether it matches execution.*

⛔ **What is not wired:** `bal_serializer_verify` (`BalSerializer.lean:1074`) performs exactly the rebuilt-vs-supplied digest comparison — four `ld`/`bne` pairs over the 32 bytes — and **nothing calls it.** The only callers of `bal_serializer_rebuild_hash` are two probes (`BalOrderDumpProbe`, `BalSerializerMeasureProbe`) and `bal_serializer_verify` itself. Its own docstring says **"INERT AS OF THIS COMMIT."**

**What runs instead:** the field-by-field consistency family (`bal_all_accounts_nonstorage_consistent`, fail code 44 — **352 rows** in the 26,104-row census) plus a rebuilt-vs-supplied content comparison (code 60 — **229 rows**, split 129 rebuilt-long / 96 short / 4 equal).

## Why this is the spec's own check and the substitutes are strictly weaker

EIP-7928 commits the BAL **through a hash**. ⇒ **Agreeing on the hash is agreeing on every byte, and nothing weaker substitutes** — matching lengths, counts and field sets are all satisfiable by a BAL that hashes differently. The serializer's own docstring makes this argument; it is quoted here because it is the justification for the issue.

## Why the order matters — after #10651, not before

The docstring also names why it was left inert, and both reasons are about the **producer**, not the serializer:

1. It is **the first thing to read the builder streams**, so every neutrality claim measured over unconsumed builder output expires the moment it lands. ⇒ It owes **a fresh paired A/B with both halves live**, not a number inherited from measurements taken over dead output.
[...]

PR #11247, opened 2026-08-03
fix(codegen): skip EIP-7702 auth re-entry on system_call_mode (code44)

pirapira:
## Summary

Stops deferred system re-entry from re-running leftover `eip7702_auth_state_prepare` after a user auth-phase OOG. That path re-applied authority nonces into the block-lived nonstorage log, producing pure **nonce-only** AGG rows that supplied BAL correctly omits → `bv_fail=44` (NONCE_ONLY_AUTH).

**Scope gated: PREPARATION FAILURE only.** Body revert still keeps set_delegation nonces (spec). Not gated on body outcome.

## Root cause (SPIKE 00029)

1. First `eip7702_auth_state_prepare` OOGs after recording one auth; finalize restores nonstorage checkpoint ✓.
2. Deferred system messages re-enter `runtime_dispatcher_call` with `system_call_mode=1` while `runtime_tx_auth_exec_fn` still points at prepare.
3. Second/third prepare runs successfully → authority nonces re-recorded → AGG mask=2 nonce 0→1 on empty BAL authorities → fc44.

Same disease class as #11148 (seed-on-system): user leftover callback re-fired on system path. Spec system txs never call `set_delegation` (`should_transfer_value=False` / no auth).

## Fix

| file | change |
|---|---|
| `Dispatch.lean` `emitTopLevelMessageD0Preparation` | `bnez system_call_mode → auth_state_used_done`; clear `runtime_tx_auth_exec_fn` on OOG |
| `SystemCallStaging.lean` | clear `auth_exec_fn` when setting `mode=1` |
| `TxIntrinsicStateGas.lean` `eip7702_authority_asof` | unconditional `account_read_record` at entry (spec `get_account` first) |
| `TxIntrinsicAuthEffects.lean` | `account_read_record` before header load |

Layout regen: `textSizeBytes` `0x62178 → 0x621b4`.

## Measure (base `f4f1de773`)

| leg | guest sha | population | result |
|---|---|---|---|
| parent | `ffca4767…` | pure NONCE_ONLY 168 | **FA0 FR168** all **fc44** |
| cand | `08ad1e7f…` | pure NONCE_ONLY 168 | **FA0 FR167 OK1** — **167×fc60**, 1 fixed |
| parent | same | r200 seed42 | FA0 FR4 OK186 TN10 |
| cand | same | r200 seed42 | FA0 FR4 OK186 TN10 |

- Pure NONCE_ONLY re-derived on **this** parent: still 168×fc44 (matches prior 1aba43dec classify size).
- **fc44→60 is masking hierarchy**, not a new defect: phantoms removed; latent Path-A digest shortfall (rb/sp 361/365 Δ-4 on 00029) was under fc44. Builder account **set** matches supplied; bal/nonce match; residual is encoding/layout (storage_reads candidate — note #11245), not invented nonce rows.
- Net FR on pure: **168→167** (−1). Honest metric = net FR, not “phantoms removed.”
- r200: FR sets **equal**; **199/200** OUTPUT 256B identical; one FR row differs only fc44→60 (still FR). reverse **0**. ELFs DISTINCT.
- ERR timeout 14317 **separate** (not in pure 168).

### Spot 00029 (`set_delegation_oog_charge`)

| | parent | cand |
|---|---|---|
| succ | 0 | 0 |
| fc | **44** | **60** |
| AGG | 4 (coinbase+sender+2 auth nonce phantoms) | 2 (coinbase+sender only) |

## Three findings precision (do not conflate)

1. **This PR** — invent nonce-only auth nonstorage rows BAL omits (phantom). Prep-failure / system re-entry.
2. **#10651 delta filter** — invent storage rows with pre==post BAL omits (same invent-vs-omit disease, different field). Not this PR.
3. **#10651 double-apply** — skip-set incomplete (bookkeeping), **not** phantom rows. Not this PR.

## Not in this PR

- Body-revert nonce retention (correct per set_delegation).
- Latent fc60 digest/encoding on residual 167.
- BAL_VALUE_SUB half of former code44 population.
- ERR 14317.

## Test plan

- [x] pure NONCE_ONLY 168 parent vs cand (fc hist + net FR)
- [x] r200 A/B FA=0 FR sets equal reverse 0
- [x] SPIKE 00029 AGG dump parent phantoms / cand clean
- [x] check-region-map after layout pin
- [ ] CI build / region-map gates

Closes pure NONCE_ONLY prep-scope half of code44. Residual fc60 tracked separately from nonce phantoms.


pirapira:
**Reviewed — label on green build.** ⚠️ **Read the net first, because the headline number is not the point of this PR: FR 168 → 167. Net −1.**

⇒ **That is the honest measure and the body says so.** 167 of the 168 rows move from `fc44` to `fc60` rather than passing. ⭐ **The masking hierarchy revealing a latent failure — not this fix creating one.** Those rows were failing the digest check the whole time; `fc44` fired first and hid it. That is exactly why the 26,104-row census total of **1068 is a floor, not a ceiling**.

⇒ **So judge this on the defect it removes, not on the counter.** The root cause is `system_call_mode=1` **re-running a leftover `auth_exec_fn` after an OOG prepare** — the same class as #11148. ⚠️ *That is an authorisation being executed after the preparation that should have rolled it back*, which is a correctness defect in its own right, and the direction of risk is not only false rejects.

**The mechanism is proven, not argued** — on 00029: parent `AGG count=4` carrying phantom auth nonces (`95d1be`, `c0f6dc` going 0→1); candidate `count=2`, coinbase and sender only. **The phantoms are gone.**

⭐ **And the spec grounding is what makes this a fix rather than a tuning**: `block_access_lists.py` never calls `add_nonce_change` for a failed or unapplied authorisation, and on 00029/01767 the BAL carries **zero nonce rows** for those addresses — an empty `AccountChanges`. ⇒ Not a max-reduction hiding a duplicate: **the spec emits nothing and we emitted a row.** *The guest recorded what the spec never produces.*

⛔ **The gate scope is the part that had to be exactly right, and it is**: gated on **preparation failure only, never on body revert.** `set_delegation` writes **persist** through a body revert, and the nonce increment with them — so gating on the body outcome would have converted 168 false rejects into an unknown number of **false accepts**. The PR states which condition it gated on and why not the other.

**A/B**: r200 **FA 0**, FR 4 equal sets, reverse 0, 199/200 identical output.

⇒ Same root as the value-unchanged storage rows found tonight in #11246 — *the guest records rows the spec never produces* — though **not** the same as that PR's double-apply, which was a bookkeeping gap. Two mechanisms, one family; worth keeping distinct in the record.

pirapira:
## Follow-on (coord order): fc44→60 unmasked class — four bytes named

On 00029 after this PR: **rb=361 sp=365 Δ−4**, sets match.

**The four bytes:** EIP-7002 `00000961…7002` `storage_reads`
- supplied RLP `c480010203` = slots **0,1,2,3** (minimal integer encoding)
- guest `c0` = empty
- account RLP 31 vs 27

Siblings 7251/0d/0e have guest reads 0–3; only 7002 missing. Producer = storage_read path on deferred 7002 system call (not a bulk endian/layout scramble on this row).

Posted on #11245 as measured hash-bind instance. **No fix in this PR** (prep-scope only). Handover complete for encoding/miss producer lane.

pirapira:
### Handover record (coord 00:37) — complete, no fix in this PR

- Supplied 7002 account RLP: `de94…7002c0c480010203c0c0c0` (len 31)
- Guest 7002 account RLP: `da94…7002c0c0c0c0c0` (len 27)
- Sibling control: 7251/0d/0e have reads 0–3; `n_reads=14=3×4+4788×2`
- Producer candidates: (1) deferred 7002 call route to `storage_read_record` (2) modelled-read set omits 7002
- Also on #10839 + #11245

Build pending — coord labels on green. Grok **IDLE**.

pirapira:
Class enumeration: #11250. This is confirmed member 1 of the OOG mode-flag hygiene class (auth_exec_fn leftover → system re-entry).

PR #11248, opened 2026-08-03
docs: document fetching EEST fixtures in a fresh checkout

### LINKED ISSUE #11243: bug: clean checkout cannot run EEST harness without fixture materialisation
A clean checkout cannot run the EEST harness because `gen-out/eest-fixtures/tests-zkevm@v0.6.2/fixtures/fixtures` is absent. Two agents independently hit this and worked around it by materialising inputs elsewhere and using the direct 30-worker Spike path. The harness should either materialise/provide this fixture tree or fail with an actionable setup instruction; otherwise a fresh checkout cannot run the advertised command.

pirapira:
## Summary

Close #11243 by documenting the existing fixture materialisation step in the
EEST runbook. A fresh checkout now has an explicit command to fetch the pinned
`tests-zkevm@v0.6.2` release asset before invoking the harness, plus the
default cache path and tag-override behavior.

The existing `scripts/eest-fetch-fixtures.sh` command is idempotent and already
implements the materialisation; no new pipeline or harness behavior is needed.

## Verification

- `git diff --check`
- `bash -n scripts/eest-fetch-fixtures.sh`
- Confirmed the current tag is `tests-zkevm@v0.6.2` in
  `scripts/eest-fixture-tag.txt`.
- Read-only release check confirmed the `fixtures_zkevm.tar.gz` asset exists on
  the pinned execution-specs release.

This is documentation-only and byte-identical for the guest; the execution-
specs fixture release remains the source of the EEST inputs.


pirapira:
**Reviewed — label on green build.** Docs-only, `+15/−5`, one file. Exactly the right size for what this is.

⭐ **The value is that it removes a friction two agents paid for independently tonight** — one at 20:34, one at 22:44 — and **both worked around it silently rather than reporting it.** ⇒ *A limitation that people route around separately is one nobody ever fixes*, because it never stays anyone's problem long enough to become a ticket. It only surfaced because a census had to explain why it fell back to a direct 30-worker Spike run instead of the harness.

**The fix is the honest one**: a fresh checkout is told to run `scripts/eest-fetch-fixtures.sh "tests-zkevm@v0.6.2"`, the ignored `gen-out` cache path and tag overrides are documented, and the stale **v0.4.0** setup text is removed. ⚠️ That last part matters as much as the addition — **stale setup instructions are worse than absent ones**, since they answer the reader's question wrongly and stop them looking further. Tonight produced several instances of exactly that shape in code comments.

**Verification is proportionate and stated**: `git diff --check`, `bash -n`, current tag confirmed as `tests-zkevm@v0.6.2`, and the pinned execution-specs release asset confirmed to exist. ⇒ *Checking that the documented command's asset actually resolves is the step that separates this from a plausible-looking doc edit* — a documented command that 404s would leave the next reader worse off than the gap did.

No regen, no layout impact, cannot conflict with #11246 or the batch.

PR #11249, opened 2026-08-03
fix(dispatch): clear prepare_only on top-level OOG exit

pirapira:
## Summary

Clear `runtime_tx_prepare_only` on every top-level exceptional exit (`emitExceptionalExit` `_top`).

**Same class as #11247** (`runtime_tx_auth_exec_fn` leftover after OOG prepare): OOG exits did not reliably clear mode state, so a later full `runtime_dispatcher_call` behaved as prepare-only.

### Mechanism (main `0e8f6d5b5`, fixture 00029)

1. User `set_delegation_oog` → `runtime_dispatcher_prepare_only` sets `runtime_tx_prepare_only := 1`.
2. Prefix OOG → `.exit_outofgas` left the flag set (comment previously called this intentional).
3. First deferred system call is EIP-7002 via full `runtime_dispatcher_call`.
4. Flag still 1 → prepare-only seam → early return **without EVM body**.
5. `storage_reads_count` after each incorporate: **7002=0**, 7251=4, 0d=8, 0e=12, +4788 → 14. No 7002 in final table.
6. Supplied BAL wants 7002 reads `c480010203` → account RLP 31 vs 27 (Δ4) → fc60 after #11247 unmasked fc44→fc60.

Three siblings (7251/0d/0e) already proved the SLOAD→`storage_read_record` path works — not a missing-address list.

### Fix

- Clear `runtime_tx_prepare_only` at start of `{label}_top` in `emitExceptionalExit`.
- **Do not** clear in callable setup (breaks intentional prepare-only split).
- **Do not** special-case 7002 or any address list.

### Downstream of `prepare_only`

Sole consumer is the prepare-only seam in the dispatcher prefix. No other readers. Related flags checked:

| flag | OOG survival | action |
|---|---|---|
| `runtime_tx_prepare_only` | **yes → this bug** | clear on `_top` |
| `runtime_tx_auth_exec_fn` | was yes | fixed #11247 + system entry |
| `runtime_tx_prepare_prefix_status` | stays 1 on OOG | intentional (caller distinguishes entered vs success=2) |
| `runtime_tx_post_preparation_reached` | only set after success | OK |
| `system_call_mode` | cleared in `stage_system_call` | OK |

### Measure

| check | parent `04b2f732` | cand `93dd7b6f` |
|---|---|---|
| 00029 | succ=0 fc60 rb361/sp365 | **succ=1 fc0 rb=sp=365** |
| storage_reads after 7002 | 0 | **4** |
| prepare_only at 7002 entry | 1 | **0** |
| r200 FA | 0 | **0** |
| r200 FR | 7 | **6** |
| forward FR→OK | — | **1** (type_4) |
| reverse OK→FR | — | **0** |
| ELFs | distinct | distinct |

Layout: `textSizeBytes` 0x62a70→0x62aac; check-region-map green.

### Pattern name

**OOG exits must clear mode flags whose scope is the failed dispatch.** Two instances in one night (#11247 auth_exec_fn, this prepare_only).

Closes the code60 EIP-7002 storage_reads gap named on #10839. No merge! from grok.


pirapira:
**Reviewed — label on green build.** `+57/−46`, and the defect is located to a branch rather than a missing entry, which is what makes it a fix rather than a workaround.

⭐ **The asymmetry was the diagnosis.** `storage_reads_count` after each system incorporate ran **7002=0, 7251=4, 0d8282=8, 0e8282=12**, plus 4788 modelled = 14 — so **three siblings have slots 0–3 and 7002 alone has none.** ⇒ *Three siblings prove the read path works, so this could only ever be a list or a branch, never a mechanism.* And it turned out to be a branch: `runtime_tx_prepare_only` left set.

**The mechanism**: a user `set_delegation_oog` calls `runtime_dispatcher_prepare_only`, which sets the flag; the prefix OOG exit leaves it set; the later deferred 7002 system call then takes the prepare-only path and never records its reads.

⭐⭐ **This is the second instance of one pattern tonight, and the pair is worth more than either fix**: `#11247` cleared `auth_exec_fn` re-running after an OOG prepare; this clears `prepare_only` surviving an OOG exit. ⇒ **Our OOG exits do not reliably clear mode state**, and both symptoms were invisible until something downstream depended on the flag being clear. *Two instances make it a class, not a coincidence.*

⛔ **No 7002 special-case and no address added to any list** — the fix is at the scope boundary where the flag should have been cleared, not at the site that noticed. A hardcoded predeploy list is exactly the shape deleted from the skip lists in `#11218` earlier tonight, and it would have broken on the next fork.

**Measurements**: 00029 goes to `succ=1`, `fc=0`, `rb = sp = 365` — the four-byte gap closed exactly where it was predicted, with `storage_reads` for 7002 going **0 → 4**. r200: **FA 0**, FR 7→6, forward 1, reverse 0.

⇒ And this is the class `#11247` unmasked: 167 rows moved `fc44 → fc60` when the phantom nonce rows were removed, revealing a digest failure that had always been there. ⚠️ *A single forward flip on r200 does not mean only one row benefits* — the census showed **229 rows on code60**, and the shortfall attributes per entry, so the corpus-level effect belongs on the full sweep rather than inferred from 200 rows.

pirapira:
Class enumeration: #11250. This is confirmed member 2 (prepare_only survives OOG top). Census SPIKE on same 00029 fixture also shows leftover access_list_seed_fn + auth_warm_fn still jalr under system_call_mode=1 after #11249 — candidate member 3, not patched tonight.

PR #11251, opened 2026-08-03
codegen: retire redundant eager nested-callee BAL seed

pirapira:
## Summary

Retire the eager nested-callee BAL seed producer.  `seed_callee_storage` was one
`AccountChanges` traversal that populated both `callee_seed_table` (storage
rows) and `callee_balance_table` (the SELFBALANCE cache); the producer, both
tables, and the dispatcher copy loop are removed together.  The tables were
not independent producers.

Nested reads now use the existing demand-driven paths:

- `execution-specs` `state_tracker.py:273-302` records `get_storage` reads and
  consults current/parent transaction writes before authenticated pre-state;
  `vm/instructions/storage.py:36-61` calls that accessor for SLOAD.
- `execution-specs` `state_tracker.py:179-202` provides the account
  write-through/pre-state lookup, and `vm/instructions/environment.py:510-535`
  reads the current account balance for SELFBALANCE.

The existing system-call preload remains.  `csce_bal_struct` remains because
the top-level recipient SELFBALANCE resolver still uses it.  The
`exec_log_seed_flag` remains because system preload writes it, storage-log
capture reads it, and runtime SSTORE clears it; it is not seed-producer-only.
The child live-effect overlay and authenticated header fallback remain.

This one retirement dissolves four tracked items: F3 on #10612, code37 (the
226-row duplicate-precedence family), prerequisite three of #11245 (duplicate
baseline precedence), and the corresponding S3 dependency in #11234.

## Evidence

Base: `0e8f6d5b5`

- Parent ELF: `04b2f732fa89decb622502bd7863e88a9dc148fd9f2b37dee0398ab666cacaef`
- Candidate ELF: `d1a1c0eb02c8c990a3d7240fc33bbb19b6b55ca2d7ee25069027c679ef2bb1f6`
- Spike runner: `5b383a7702fb05dea16a2cc37dc313b82491fdf194f39c12306cb387a5a2b6b0`

Focused witnesses on the merged-main base were byte-identical with the seed
removed: 23000 (seed hits 1 -> 0, h_SLOAD 17 -> 17) and 00336 (8 -> 0,
h_SLOAD 28 -> 28).  A bounded random-200 A/B used the same manifest and seed
`20260803`, with 30 Spike workers:

| leg | ran | FA | FR | exact |
| parent | 198/200 | 0 | 8 | 190 |
| candidate | 200/200 | 0 | 5 | 195 |

The parent had two emulator `ERROR(exit)` rows (00190 and 00191), so those are
unclassified rather than regressions.  The candidate had zero errors.  The
three forward flips are 00012 and 00050 (`consolidation_requests`) and 00090
(`consolidation_requests_out_of_gas`); reverse flips: **0**; new false
accepts: **0**.  The remaining candidate FR rows are unrelated existing
residuals (00052, 00067, 00156, 00163, 00192).

This also explains the 00012 contradiction recorded on #10839: the eager seed
created a duplicate row that masked the baseline; removing it fixes that row.

## Checks

- Non-cache `lake build` (4796 jobs): pass.
- `scripts/check-build-parallel.sh`: pass (including asm-to-program,
  RegionMap/TSV/layout, guarded-handler byte tie, axioms, arithmetic fuzz).
- `scripts/check-region-map.sh`: pass.
- `scripts/check-guarded-handler-bytes.sh`: pass (168 bytes).


pirapira:
**Reviewed — label on green build.** `+2617/−3039` across 14 files, net **−422 lines**. This is the largest convergence item on the board and it measures strictly better on every axis.

| leg | ran | FA | FR | errors |
|---|---|---|---|---|
| parent `04b2f732…` | 198/200 | 0 | 8 | **2** |
| candidate `d1a1c0eb…` | **200/200** | 0 | **5** | **0** |

Forward 3, **reverse 0**.

⭐⭐ **The error rows are the strongest result and deserve to be read first**: `00190 sstore_0to0` and `00191 invalid_negative_excess_blob_gas` previously **failed to execute at all**. ⇒ *An eager producer that makes the emulator exit is worse than one that emits wrong rows.* They are correctly labelled **unclassified rather than regressions** — they were never passing, so they cannot be regressions, and now they complete.

⭐ **And the three fixes are one family, not scatter**: 00012, 00050 `consolidation_requests`, 00090 `consolidation_requests_out_of_gas`. ⇒ **00012 is the row filed on #10839 at 01:12 as an unresolved contradiction** — `bv_fail=1` alongside a reported state-root match. The eager seed was the cause, so that open note closes here.

**Why this is one diff and not two**, which is the question a reviewer will reach for: `seedCalleeStorageFunction` is **one `AccountChanges` traversal** that resets both counts at entry, resolves each account's balance into `callee_balance_table`, then enumerates that same account's storage keys into `callee_seed_table`. **One producer, two outputs.** ⚠️ *I asked for these to be split and the code answer was that they cannot be — adjacency is not coupling, but neither is separateness implied by two table names.*

**The spec grounding is the right kind** — `state_tracker.py:273-302` with `storage.py:36-61` for demand-driven SLOAD, and `state_tracker.py:179-202` with `environment.py:510-535` for account/SELFBALANCE. ⇒ *The lazy read the spec already performs is what made the eager producer redundant.*

**Retained and named**: `csce_bal_struct` for top-level recipient SELFBALANCE, the authenticated live-effect and header fallback, the system-call preload, and the live `exec_log_seed_flag` readers. **A retirement that states its exclusions is checkable.**

⇒ **Four tickets close on this**: F3 on #10612; **code37 and its 226 rows** — the second-largest FR class, whose duplicate-precedence masking existed *because* the seed inserted the duplicate; **#11245 prerequisite three**, independently verified closed on this candidate; and part of S3 on #11234.

⚠️ **Five residual FRs named and explicitly unexplained** — 00052, 00067, 00156, 00163, 00192 — with no implied relationship to this change. That restraint is what keeps the result readable.

pirapira:
**Labelled while CONFLICTING — deliberate, and the reasoning is the same one that landed six PRs tonight.**

Main advanced to `02c6fcd7c` (#11249 merged) after this branch was pushed. ⇒ **GitHub cannot construct a merge ref for a conflicting branch, so the `build` check will never report** — waiting for green here waits for something that cannot arrive. The batch performs the layout regen and *is* the verified build; #11210, #11218, #11220, #11226, #11235, #11236 and #11246 all landed exactly that way.

⚠️ **The conflict is layout, not content**, and it is explicitly not the author's to resolve: **author-side fixpoints cost four wasted regens tonight**, three of them on a single PR, because every merge repins `.text` while other deletions are landing.

**For the batch — what must not be lost**, since the diff is `+2617/−3039` and largely generated:

- the behavioural change is the removal of **one producer with two outputs** (`seedCalleeStorageFunction`, single `AccountChanges` traversal populating `callee_balance_table` and `callee_seed_table`), plus its tables and the `runtime_dispatcher_call` copy loop
- **retained deliberately**: `csce_bal_struct` for top-level recipient SELFBALANCE, the authenticated live-effect and header fallback, the system-call preload, and the live `exec_log_seed_flag` readers
- everything else is regeneration the batch recomputes

⭐ **Verified on 200 rows before push, both directions**: parent 198/200 with FA 0, FR 8 and **2 emulator errors**; candidate **200/200** with FA 0, FR 5 and **0 errors**. Forward 3, reverse 0. ⇒ *The two rows that previously failed to execute now complete* — an eager producer that makes the emulator exit is worse than one that emits wrong rows, and those are correctly recorded as **unclassified rather than regressions.**

⇒ Landing this closes **four tickets**: F3 on #10612, **code37's 226 rows**, **#11245 prerequisite three** (independently verified on this candidate), and part of S3 on #11234. With #11249 already merged, **#11245 stands at zero producer blockers** and its wiring plan is posted.

Issue #11250, opened 2026-08-03
OOG exit paths do not reliably clear dispatcher mode/continuation flags (class of #11247+#11249)


pirapira:
## Summary

Two confirmed defects in one night show the same pattern: **mode/continuation flags set before or during execution are not cleared on every OOG exit path**, and the leftover is invisible until a later phase (deferred system call, next dispatcher entry) depends on the flag being clear.

| instance | flag | PR |
|---|---|---|
| #11247 | `runtime_tx_auth_exec_fn` re-run on `system_call_mode=1` after OOG prepare | #11247 |
| #11249 | `runtime_tx_prepare_only` survives OOG top → first deferred system call prepare-only early-return | #11249 |

This ticket is the **enumeration**, not a third patch. Static scan is an upper bound (#11229); unclear cells were SPIKE_WATCH/DEBUG dumped on `set_delegation_oog_charge` (00029), the fixture that already exercised both known defects.

## Method

- Guest parent (pre-#11249): sha `04b2f732…` ELF `/tmp/grok-code60-7002/parent.elf`
- Guest cand (with #11249): sha `93dd7b6f…` ELF `/tmp/grok-code60-7002/cand.elf`
- Fixture: `00029_test_set_delegation_oog_charge_point_…-oog_charge_a.input`
- Dump points: `stage_system_call` entry, `runtime_dispatcher_call` multi-hit, `seed_tx_access_list` / `eip7702_warm_recovered_authorities` entry with `system_call_mode`

## Flag table

Scope: **mode / continuation / function-pointer** cells that steer dispatcher phases (not gas counters, not per-row log bytes).

| flag | where SET | where CLEARED (static) | OOG top clears? | nested OOG (frame_return) clears? | SPIKE 00029 at next phase | class |
|---|---|---|---|---|---|---|
| **`runtime_tx_prepare_only`** | `runtime_dispatcher_prepare_only` entry `:=1` | normal consume at prepare_prefix seam; **#11249** `emitExceptionalExit *_top` | **YES after #11249**; **NO before** | N/A (flag is top-level split) | parent stage: **1**; cand stage: **0** | **FIXED #11249** |
[...]

PR #11253, opened 2026-08-03
fix(dispatch): exceptional-exit mode teardown (#11250)

pirapira:
## Summary

#11250 candidate 3 (FA-shaped): `runtime_tx_access_list_seed_fn` and
`runtime_tx_auth_warm_fn` survive OOG prepare and are `jalr`-ed under
`system_call_mode=1` with `a0`/`a1` still holding **user list pointers**.

This is not a third one-off clear. Nested OOG cleared **none** of the mode
flags; top OOG after #11249 cleared **only** `prepare_only`. Missing
exceptional-exit teardown discipline — every prior clear was accidental to
its own path.

### Teardown (`emitExceptionalExitModeTeardown`)

Runs at **every** exceptional-exit **entry** (before depth branch → nested +
top) and nested static-violation path.

| cleared | why |
|---|---|
| `access_list_seed_fn` + ptr + len | candidate 3 |
| `auth_warm_fn` + list ptr + len | candidate 3 |
| `auth_exec_fn` | consolidates #11247 |
| `prepare_only` | consolidates #11249 |

**Intentionally sticky — NOT cleared:**
`auth_phase_halted`, `prepare_prefix_status`, `auth_state_charged`,
`code_state_mtx_active`, `runtime_tx_oog_hook`.

### Measure (00029 `set_delegation_oog_charge`)

| leg | seed/warm under `system_call_mode=1` |
|---|---|
| parent (post-#11249) | **jalr fires**: seed `a0=0x40000299 a1=1`; warm `a0=0x4000029a a1=0xba` |
| candidate | **BREAK hits=0** on `seed_tx_access_list` and `eip7702_warm` |

SPIKE_WATCH: candidate clears both fn cells at `.exit_outofgas` entry (not at
the later consume site).

00029 still `succ=1 fc=0 rb=sp=365` (same as parent).

### A/B random-200 seed42

| leg | guest | FA | FR | OK |
|---|---|---:|---:|---:|
| parent post-#11249 | 93dd7b6f… | 0 | 6 | 194 |
| cand | 3f371aa3… | 0 | 6 | 194 |

FR sets equal; forward 0; reverse 0. FA is the load-bearing number here.

### Layout

`textSizeBytes` `0x62aac → 0x62cb0`. check-region-map GREEN.

### Files

- `EvmAsm/Codegen/Dispatch.lean` — teardown + wire
- RegionMap / GuestAddrs / TSV regen

Closes nothing alone; advances #11250. No merge! from author.


pirapira:
**Reviewed — label on green build.** `+94/−58`, and it fixes a **class** rather than a third instance of one.

⭐⭐ **The reframing is what makes this the right shape.** Two defects were patched tonight one flag at a time — `auth_exec_fn` re-running after an OOG prepare (#11247) and `prepare_only` surviving an OOG exit (#11249). The census that followed found the actual fault: **nested OOG cleared *none* of the mode flags, and top-level OOG cleared only `prepare_only`.** ⇒ *So every flag cleared today is cleared by accident of its own code path — there was never a teardown at all.* A third individual clear would have been the third symptom patched with the class left open.

⛔ **And candidate three was the severe one, which is why it justified reopening the lane**: `runtime_tx_access_list_seed_fn` and `runtime_tx_auth_warm_fn` surviving the prepare and being `jalr`-ed under `system_call_mode=1` **with `a0`/`a1` still holding user list pointers.** The measurement is explicit — parent shows `seed a0=0x40000299 a1=1` and `warm a0=0x4000029a a1=0xba` under system mode; **candidate breakpoints hit 0.** ⇒ *The first two defects suppressed or duplicated rows; this one called a leftover function pointer with user-supplied arguments during a system call.*

**Clear set is complete and stated**: 8 cells — both function-pointer triples (`fn` + `ptr` + `len`), plus `auth_exec_fn` and `prepare_only` — installed at **every exceptional-exit entry, nested and top-level, including nested `static_violation`.**

⚠️ **And the sticky set is left untouched, which is the part a fixer could not have reconstructed**: `auth_phase_halted`, `prepare_prefix_status`, `auth_state_charged`, `code_state_mtx`, `oog_hook`. ⇒ *A teardown that cleared those would break behaviour that depends on surviving the exit* — that distinction came out of the census, not the defect.

**Measurements**: 00029 remains `succ=1`, `fc=0`; the clear is `SPIKE_WATCH`-confirmed at `exit_outofgas` entry; r200 **FA 0, FR 6 with equal FR sets, forward 0, reverse 0.**

⇒ **Flat FR with FA=0 is the correct pass for this shape and should not be read as a null result.** This is a false-accept-shaped defect: the fix removes user-controlled data reaching a system-call context, and **the corpus was never going to reward that with fewer false rejects.** Judging it on FR movement would measure the wrong thing.

PR #11261, opened 2026-08-03
codegen: guard BAL storage value arena bounds

pirapira:
## Summary

Add fail-closed direct end checks for the `baap_storage_values` arenas in all
four value-encoding/cursor paths.  Each build unit now places a local
`baap_storage_values_end` marker immediately after its allocation.  Before
`rlp_encode_bytes`, the guard requires `cursor + 40 <= end`; after encoding it
checks the 8-byte-rounded next cursor before storing it.  Overflow reports
`baap_fail_code = 599` through the existing failure tail.

The 40-byte envelope covers the maximum 32-byte storage value and its
8-byte-aligned footprint.  The checks are local to the allocation and retain
the existing count, value-length, and RLP validation guards.  In-range inputs
are unchanged; an impossible arena overrun now rejects before the write.

This is consistent with execution-specs' BAL contract: the Amsterdam BAL
module describes the list as recording every storage location accessed and its
post-execution value (`execution-specs/src/ethereum/forks/amsterdam/block_access_lists.py:1-8`),
records storage reads with `add_storage_read` (`:371-382`), and incorporates
the accumulated block reads into the final list (`:691-699`).  The patch only
adds a local memory-safety boundary around the guest's temporary encoding
buffer; it does not alter which BAL entries the spec requires.

## Verification

- `lake build` (4796 jobs): PASS.
- `scripts/check-build-parallel.sh`: PASS (codegen, build-unit links,
  region-map, memory-layout coverage, guarded-handler byte identity, progress,
  drift, axioms, and arithmetic fuzz).
- `ELF_DIR=/tmp/11223-final-guest NO_BUILD=1 scripts/check-region-map.sh`: PASS.
- `scripts/check-asm-to-program.sh`: PASS (386 converted definitions).
- Final `stateless_guest` ELF SHA-256:
  `031453d9f17f2188a7391092819977497b89fadef41134876e5d2fd899f50b1e`.
  Linked sizes: `.text=0x62308`, `.data=0x5370`, `.bss=0x1a954460`.
- A temporary `.zero 0` perturbation of the stateless arena made fixture
  `00249_test_mixed_gas_regimes_fork_Amsterdam-blockchain_test__b0` reach the
  new guard and report `baap_fail=599`.  Restoring the real allocation made
  the same fixture report `baap_fail=0` on debug ELF
  `b503156f688c3aab6d840d09f671d0a567724d6407f010887ac4fc8728868654`.
- No merge label was applied.


pirapira:
**Cross-lane note for the batch (coord):** this PR and #11260 both touch
`EvmAsm/Codegen/RegionMap.lean`, and #11261 additionally regenerates
`scripts/asm-fixtures/symbol-addresses.tsv`.

#11260 *moves the link-dependent pins* (`textSizeBytes`, `dataSizeBytes`, `bssSizeBytes`,
`callFrameArenaBase`, `evmMemoryPoolBase`, `syslogBase`) out of `RegionMap.lean` into a generated
`RegionMapLinkPins.lean`. #11261 changes `.text`/`.bss` sizes (`.text=0x62308`,
`.bss=0x1a954460`) by adding the `baap_storage_values_end` markers.

⇒ Preferred order is **#11260 first, then #11261 regenerated on top** — after #11260 those pin
values no longer live in `RegionMap.lean`, so a textual resolution could otherwise leave a stale
pin in the new location. This is a sequencing preference for the regen, not a request to change
either PR; both are green on their own terms and the batch build is the arbiter.

Issue #11230, opened 2026-08-02
derive link-dependent RegionMap pins from the linked guest image


pirapira:
## Problem

The three link-dependent RegionMap pins — `callFrameArenaBase`, `evmMemoryPoolRegion.base`, and `syslogBase` — are hand-maintained absolute literals. Each .bss layout change can invalidate them, so an unrelated layout PR currently requires a manual repin. Recent merges made this failure mode concrete: three pins had to be repinned twice within one evening as the guest image moved.

The intended invariant is:

> RegionMap agrees with the linked `stateless_guest` image.

It should not be:

> RegionMap still equals a number typed during an earlier link.

## Desired change

Make `scripts/check-region-map.sh` derive the expected values from the generated linker facts (`scripts/asm-fixtures/symbol-addresses.tsv`) or directly from `readelf`, then compare the three RegionMap values against those derived symbols. Keep the structural relative-offset and extent checks unchanged.

The generated TSV is already produced from the linked ELF and drift-guarded. If deriving the expected values is not possible because Lean proofs must consume concrete literals, document that constraint explicitly and record the unavoidable manual-repin maintenance cost in `docs/regenerating-generated-files.md`, including why the guard cannot be made agreement-based.

## Acceptance criteria

- The guard verifies all three pins against the current linked artifact, without hard-coding their absolute expected values in the shell check.
- A synthetic change to any one RegionMap pin makes the guard fail.
- A changed linked symbol makes the guard fail until the corresponding generated or checked-in artifact is regenerated.
- The proof-side constants remain kernel-checked and no theorem is weakened.
- The regeneration documentation states the source of truth and the remaining maintenance steps.

Related: #11224

pirapira:
Measured failure-rate evidence from #11224: three link-dependent repins were invalidated in one evening by ordinary mainline movement. At 20:06 UTC, the repin against 9ffb5398a recorded evm_memory_pool = 0xb37f67a0. At 20:41 UTC, #11231 landed with a 448-byte .bss shift. At 20:58 UTC, the guard caught the still-open PR pin 0xb37f67a0 against a fresh ELF value of 0xb37f65e0 (the corresponding fresh bases were call_frame_arena = 0xad3dd5e0 and bv_system_storage_log = 0xab07a640).

[...]

PR #11277, opened 2026-08-03
codegen: group state-root owners across maps

pirapira:
## Summary

- Make the execution account/storage maps authoritative for state-root owner enumeration.
- Deduplicate canonical 20-byte owners across account-map and storage-map rows, while combining account and storage effects for mixed owners.
- Replace the BAL-count withdrawal shortcut with an explicit per-withdrawal check against emitted account-map owners; storage-only owners continue through the explicit withdrawal update.
- Add executable fail-closed guards for duplicate MPT paths (`bsr_fail_code=125`) and owner-set overflow (`bsr_fail_code=126`).

The owner-set capacity is symbolic and pinned by a kernel guard:
`bsrMapOwnerCapacity = blockAccountWritesCapacity + storageWritesCapacity + bsrModeledSystemChanges = 20480 + 16384 + 2 = 36866`.
The `+2` accounts for the two modeled-system owners seeded outside both authenticated maps.

## Motivation

The execution-spec state tracker unions account and storage maps (`state_tracker.py:718-728`), so the old BAL-derived candidate set had no corresponding precedence rule. A census of 26,096 valid blocks found 7,232 mixed-owner rows (20.7%), making owner-level grouping the main path rather than an edge case.

## Validation

- `lake build`: 4,797 jobs passed.
- Structural checks passed: forbidden tactics, layering, file-size, unimported modules, and approved heartbeat mentions.
- Temporary generated-assembly perturbations proved both runtime guards fire: duplicate path → `bsr_fail_code=125`; owner count forced to the exact `36866` capacity → `bsr_fail_code=126`.
- Supporting debug artifact `daefc47b5fdf48217b600387030320b30fcdb2e9f6b99546fa05ab537d823a1d` (debug ABI) produced a byte-identical 00009 control result: verdict `0`, `bv_fail_code=53` at `+8`, `bsr_fail_code=0` at `+40`, and `change_count=5` at `+48`.

The maximum mixed-owner input is identified by SHA256 `6b839c9f39d2d9894e9450dc00f6d63f0af843a1240a4313246af92a84fac7c0`. On the candidate debug ELF `daefc47b5fdf48217b600387030320b30fcdb2e9f6b99546fa05ab537d823a1d` (built 2026-08-03 11:17:42 +0200), the debug ABI reads verdict `1` at `+0`, `bv_fail_code=0` at `+8`, `bsr_fail_code=0` at `+40`, and `change_count=15` at `+48`: it passes. A current normal-guest measurement on this commit is not available; grok's shared-base A/B on `d2eb960ef` is authoritative for that ABI.


pirapira:
Parked `complicated-merge`: there are **two** failures here, and CI only shows the first — the second is the real blocker.

**1. Layout drift (fixable, I fixed it).** CI reports:
```
DRIFT RegionMapLinkPins.textSizeBytes: expected 623dc, ELF has 626a0
DRIFT RegionMapLinkPins.bssSizeBytes:  expected 1a954460, ELF has 1aa744a0
DRIFT RegionMapLinkPins.callFrameArenaBase / evmMemoryPoolBase / syslogBase
```
That is the ordinary repin. I merged main (no conflicts) and ran the #11230 flow to a fixed point: second `gen-region-map-link-pins.py` pass is a verified no-op, `--check` reports CLEAN, and `check-region-map.sh` then reports "region map matches the linked ELF". My pins agree with the ELF your CI observed (`textSizeBytes = 0x626a0`).

**2. After the drift is fixed, `lake build` still fails** — and this is what CI never gets to, because it stops at the drift gate:
```
EvmAsm/Codegen/Proofs/GuestImage.lean:59:37: Tactic `decide` proved that the proposition
is false
```

What I ruled out:
- **Not my merge.** `GuestImage.lean` is byte-identical to main here (sha256 match), and this PR touches only `BlockVerdictDataSection.lean` and `BlockVerdictStateRoot.lean`.
- **Not stale generated files.** `GuestImageEntries.lean` regenerates to exactly the committed content (no diff), and the pins are converged.
- **Not an obvious overlap or overflow.** `guest_image_coverage.py --gaps` shows a sane ascending map, and the final gap ends at `0x800626a0`, exactly `guestTextEnd` = `0x80000000 + 0x626a0`.

So the entry extents are inconsistent with the image `CodeReq` in a way the coverage tooling does not surface. Since it comes from this PR's own regrouping of state-root owners, and diagnosing it means reasoning about which converted `_prog`'s verification-view extent no longer matches its emitted slot, I have not guessed at a fix.

Everything else passes on the repaired tree: the other 14 gates including `check-region-map`, `check-asm-to-program` and `check-build-units-link`. Fix the `extentsOk` inconsistency and the merge is otherwise mechanical — I can finish it immediately.

pirapira:
**Re-labelled `merge!` and cleared `complicated-merge` — the `extentsOk` failure is a stale
generated file, not a proof problem, and the maintainer has confirmed the bot should regen.**

## Diagnosis (evm-asm3-codex, read-only)

The regrouping grew `execution_map_state_changes`, so every symbol after it shifts by a constant
**`+0x2c4`** — first shifted is `codes_blockhash_required_headers` at `0x80014ef4`; earlier entries
are unchanged. **The committed `GuestAddrs` are stale by `+0x2c4` for 222 entries**, which is why
`CodeReq.extentsOkFrom` evaluates false at `GuestImage.lean:59`.

⭐ **The structure is sound.** Using the candidate's own `nm` addresses plus the current `#guard`
Program lengths, **all 346 linked entries are ascending and no extent overruns.** Nothing is wrong
with the change; only the declared addresses are out of date.

## What the regen needs to cover

`GuestAddrs.lean`, `scripts/asm-fixtures/symbol-addresses.tsv`, `GuestImageEntries.lean`, and the
link pins (#11260 derives the class-A pins, so the generator computes them).

⚠️ **Run it twice — the second pass must be a no-op.** #11260 established that the pipeline
converges in one pass, but a **222-entry shift is the largest displacement it has been tested
with**, so convergence is worth confirming rather than assuming here.

⚠️ **Never hand-edit.** A constant `+0x2c4` across 222 entries is exactly the case where
hand-patching looks trivial and is worst: an address is an emission *input*, consumed and never
verified, so a hand-set value yields a confident wrong artefact with a green build.

## Please re-confirm after the regen

The withdrawal fix on this head depends on path-slot indexing, so it should not be assumed to
survive a 222-entry shift:

| fixture | expected |
|---|---|
| `1c2ac7de…` (three CL withdrawals) | `succ=1`, `bv_fail=0`, `bsr_change_count=3` |
| `6b839c9f…` (max mixed-owner, 10) | `succ=1`, `bv_fail=0` |

## Prior blocker, already fixed on this head

The withdrawal false-reject was `.Lbsr_wd_map_scan` testing `flag==1` before the address compare,
while the recipient was emitted as a `flag==0` storage-only owner. Address-matching all emitted
owners fixes it, and the experiment also proved that a `flag==0` emission **does** carry the
balance — skipping it still yields the correct 3-change root.

Issue #10703, opened 2026-07-27
PR Summary (summarize) fails on every fork PR: actions/checkout refuses fork checkout under pull_request_target


dhsorens:
Split out of #10053 at the reviewer's request, since it affects **all outside contributors**, not one branch.

## Symptom

The `summarize` job of **`.github/workflows/summary.yml`** (`name: 'PR Summary'`) fails within ~6s on any pull request opened from a **fork**, at the *Checkout (full history for base diff)* step:

```
##[error]Refusing to check out fork pull request code from a 'pull_request_target' workflow.
This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache
scope, and runner access. Fetching and executing a fork's code in that trusted context
commonly leads to "pwn request" vulnerabilities. To opt in, review the risks at
https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true'
on the actions/checkout step.
```

## Cause

`summary.yml` triggers on `pull_request_target` (line 4) and then checks out the PR head:

```yaml
- name: Checkout (full history for base diff)
  uses: actions/checkout@v4
  with:
    ref: ${{ github.event.pull_request.head.sha }}
    fetch-depth: 0
```

Recent `actions/checkout` releases refuse this combination by default unless `allow-unsafe-pr-checkout: true` is set. It is a new default in the action, not a change in this repo — which is why the job passed on fork PRs previously and fails now.

## Scope

- Fails on **every fork PR**. Observed on #10053 (from `dhsorens/evm-asm`); will hit #10683 and any future external contribution.
- Same-repo branch PRs are unaffected (no fork checkout).
- **Only `summarize` is affected.** `.github/workflows/labeler.yml` also uses `pull_request_target`, but it runs `actions/labeler@v5` without checking out PR-head code, so `path-labels` / `size-label` still pass.
- Not a code gate: `build` / `source-checks` / `velocity` are unaffected, so this is red-X noise on contributor PRs rather than a correctness signal — but it does mean fork PRs never get a summary, and a permanently-failing check trains reviewers to ignore red.

## Options (maintainer's call)

[...]

PR #11278, opened 2026-08-03
feat(#11042): bring verified guest routines inside the axiom gate

### LINKED ISSUE #11042: Verified guest routines are outside the axiom gate: Progress.lean is opcode-only
Split out of @pirapira's review finding 1 on #11040, which asked that this be filed rather than left implicit.

## The gap

`scripts/check-axioms.sh` audits the theorems `EvmAsm/Progress.lean` classifies as `.proven`/`.conditional`/`.partly`. That registry is `OpcodeEntry`-shaped — **every entry is an EVM opcode**. So no *guest routine* spec is in it, and none is checked by the kernel-truth gate.

`rlp_encode_uint_be` is the example that surfaced it: `reub` appears zero times in `Progress.lean`, so neither #10943's 35 block theorems nor #11040's whole-routine triple `reub_spec_within` / `reub_spec_encode_within` is witnessed — even though the composed triple is the strongest claim the repo makes about that routine. It is not specific to RLP: the same holds for the account accessors, the RlpWalk chain, `WithdrawalDecodeClose`, the splice helpers, and the rest of the verified guest surface.

## Why registering one theorem does not fix it

`Progress.lean`'s registry keys on opcode name and carries opcode-shaped fields (`cycleBound`, opcode status). A guest routine has no opcode name to key on, so closing this needs a **second registry** (or a non-opcode variant of `OpcodeEntry`) plus a `gen-axiom-witnesses.py` pass over it — a registry-shape change, not an added row. That is why it did not land in #11040.

## Suggested shape

1. Add a `RoutineEntry` registry to `EvmAsm/Progress.lean` keyed on the guest symbol (`rlp_encode_uint_be`, `withdrawal_decode`, …) with the witness theorem name.
2. Extend `scripts/gen-axiom-witnesses.py` to emit `#print axioms` lines for it into `EvmAsm/Progress/AxiomWitnesses.lean`, so the gate covers it under `lake build` with the same cache-mode behaviour as today.
3. Populate it from the verified guest surface — enumerate rather than sample, since sampling is what let this go unnoticed.

## Why it matters

"Covered by CI" is the kind of assumption nobody re-checks. Today a guest-routine spec could regress to `sorryAx` or pick up a `native_decide` trust axiom and `check-axioms.sh` would stay green, because it is not looking at that theorem at all. The forbidden-tactics source scan would catch the `native_decide` case; nothing catches the `sorryAx` case.

Not a regression — this has been true since the registry was introduced. Filing so it is tracked rather than implicit.

dhsorens:
> Replaces #11273 (same commits, `7a6e1ec82`). Moved from a fork branch to an in-repo branch so `summarize` can run — it fails on every fork PR (#10703). Review discussion from #11273 is carried over in the first comment below.

---

Closes #11042 (partially — see **Residual** below).

`scripts/check-axioms.sh` audits the theorems `EvmAsm/Progress.lean` classifies. That registry is `OpcodeEntry`-shaped, so no *guest routine* was in it and none reached the kernel-truth gate. Today a guest-routine spec could regress to `sorryAx` and CI would stay green.

## The mechanism was sharper than the issue states

The issue frames this as the registry being opcode-shaped, which is true. But witnesses are not discovered from registry rows at all — `proofRef` is a display-only `String`. They are discovered by regexing the witness `abbrev` block, and that pattern was:

```python
NAME_RE = re.compile(r"@EvmAsm\.(?:Evm64|Stateless)[A-Za-z0-9_.']*")
```

⚠️ **That cannot match the guest surface.** `reub_spec_within` is `EvmAsm.Codegen.RlpEncodeUintBeSAsm.reub_spec_within`; 289 modules live under `EvmAsm.Rv64.*`. Both return `[]`. So adding an abbrev for a guest routine *without* touching the generator would have left the gate green while covering nothing — a silent skip, which is worse than the honest absence we had, because "covered by CI" is the assumption nobody re-checks.

So the fix closes the mechanism, not just the row set:

1. **Pattern is namespace-agnostic**, and anchored on the abbrev's `:=` rather than matching bare `@EvmAsm.…` anywhere. The anchor is load-bearing: these files now discuss witness names in their own docstrings, and my first version emitted `#print axioms EvmAsm.Codegen.` from a prose mention. Malformed names (trailing dot) are now a hard error rather than an unresolvable `#print axioms` line.
2. ⭐ **A `proofRef`-vs-abbrev cross-check that fails loudly.** Naming a theorem in a registry row does *not* put it in the gate; the abbrev does. A row without one is invisible, which is the defect class this issue is about.

## That cross-check immediately found a second, pre-existing hole

⛔ **MCOPY was never audited.** Registered `.proven` at `Progress.lean:358` since the registry was introduced, theorem present at `EvmAsm/Evm64/Mcopy/Spec.lean:402`, module imported at line 97 — but **no witness abbrev**, so it never reached `AxiomWitnesses.lean`. Fixed here. It is the only such row; the other 74 `proofRef`/`coverRef` names all resolve.

## Shape

`RoutineEntry` lives in a **new sibling module** `EvmAsm/Progress/Routines.lean`, not in `Progress.lean` as the issue suggested. Witnessing a guest routine means importing where it is proved, and those are under `EvmAsm.Codegen.Programs.*` (815 modules, the highest-churn part of the tree). `Progress.lean` is imported by `Progress.{Correspondence,Obligations,AxiomWitnesses}`; pulling Codegen into it would make every Codegen edit rebuild the opcode gate. `ProofTier` is shared by import, so both registries stay on one scale. **Say if you'd rather have it inline and take the rebuild cost.**

⚠️ **This required a layering exemption — the one thing here that needs your ruling.** `scripts/check-layering.sh` L1 says no verified-core module may import Codegen, and it is core-by-default, so `EvmAsm/Progress/**` counted. I exempted the progress registry from L1, because L2 (*nothing* under `EvmAsm/**` imports `EvmAsm.Progress`) makes the registry a **pure sink** — there is no `Codegen → registry → core` path, which is the laundering hazard L1 exists to stop. The exemption is therefore sound *only while L2 holds*, and that dependency is written into the script. Exempted edges **print on every run** rather than being silently skipped; an invisible exemption is how a fitness function erodes. I verified L2 still bites by making a core file import the registry and watching it fail.

`RoutineEntry` rows are **witnesses, not routines** — `symbol` groups rows rather than keying them. The walk chain proves `rlp_walk_next` separately for the scalar form and for account fields 0 and 1; collapsing those would have to name one theorem as *the* witness for the symbol, overstating what any one of them proves.

Tiers are read honestly, with the gate stated in prose in a `gate` field rather than hidden behind a constructor: resource/ABI preconditions (alignment, `isValidByteAccess`, non-overflow) do **not** make a row `.conditional`; a nonvacuous input-domain restriction does. On that reading `reub_spec_within` is `.conditional` (`n - reubZeros xs 0 n ≤ 55`), not `.proven` — 4 `.proven` / 7 `.conditional` across 11 rows and 6 symbols, kernel-checked by `decide`.

## Verification

- `check-axioms.sh` → **OK, 107 witnesses** (was 96), only `propext` / `Classical.choice` / `Quot.sound`. Parsed 104 reports with axioms + 3 with none = 107, zero non-classical.
- The gate's own witness grep is deliberately **independent** of the generator (that is what makes a generator bug visible rather than self-confirming). I widened it in parallel and confirmed the two derivations produce **identical** 107-name sets.
- ⭐ **Negative control — the gate bites.** Pointed a witness at a `sorry`-carrying theorem: `FORBIDDEN (sorry): … depends on sorryAx`, exit 1. Reverted. Without this the PR would only show the gate *runs*.
- **Second control:** removed `rlp_item_size_spec_within`'s abbrev while leaving its row; the cross-check failed naming it. (MCOPY above is the same control occurring for real.)
- `lake build` clean tree-wide; `check-{layering,forbidden-tactics,unimported,file-size,progress,no-warnings}.sh` all pass; `progress-report.sh --check` clean (the opcode report is untouched, as intended).
- Note: `check-axioms.sh` ran fine on macOS here — no BSD `mktemp` problem.

## Residual — declared, not silent

⚠️ The issue asks to "enumerate rather than sample". **This is a partial enumeration and I am not claiming otherwise.** `EvmAsm/**` carries ~1835 distinct `*_spec_within` names, most block-level rather than whole-routine, so populating all of them is not one PR. What is registered is witnessed; what is not registered is still outside the gate. That is stated in the generated module's header too, so the next reader does not inherit "covered by CI".

Populated here: `rlp_encode_uint_be` (3 rows), `rlp_item_size`, and the RLP walk chain / account accessors (7 rows). The withdrawal-decode surface I deliberately left out: it is WP-certificate shaped rather than one whole-routine triple, and picking "the strongest claim" there is a judgement call I would rather you weigh in on than guess.


dhsorens:
## Carried over from #11273

Verbatim, so the review trail survives the fork→in-repo move. Attribution is in each heading.

---

### @dhsorens (2026-08-03)

## Update (2f496430a): I missed a third registry, and it was the one with the most to gain

While applying the SpecRef review note on #11274 I found that **`EvmAsm/Progress/Correspondence.lean` is a third registry**, and it carries four witness abbrevs that **no version of this gate has ever audited**:

```
EvmAsm.Rv64.RLP.rlp_walk_init_spec_within
EvmAsm.Rv64.RLP.rlp_walk_next_spec_within
EvmAsm.Rv64.RLP.rlp_content_to_u64_spec_within
EvmAsm.Rv64.RLP.rlp_content_to_u256_be_spec_within
```

⭐ These are exactly the verified-guest-routine specs this issue is about — and they were sitting in a Progress-family registry the whole time, in a namespace the old `(?:Evm64|Stateless)` pattern could not match either. So the gap was one file wider than the issue describes and wider than my first commit assumed. **Witnesses 107 → 111.**

⚠️ It also makes my original PR description slightly wrong: I said the guest surface was outside the gate because the registry is `OpcodeEntry`-shaped. That is true of `Progress.lean`, but `Correspondence.lean` is *already* routine-keyed and *already* binds routine witnesses — they just never reached `AxiomWitnesses.lean` because nothing scanned that file.

### Two changes this forced

**Cross-check is now against the union of all registries**, not per-file. A theorem is audited if *some* scanned registry binds it. Per-file checking falsely flagged `rlp_item_size_spec_within` — named by a `Correspondence` row, witnessed from `Routines.lean` — as uncovered.

**Sources carry a `strict` flag.** `Progress.lean` and `Routines.lean` are strict: a row naming a theorem no registry witnesses is an error. `Correspondence.lean` is non-strict, and I want to be explicit about why rather than have it look like an oversight: 7 of its rows name theorems that no registry witnesses, and their modules are not imported there, so making it strict would demand 7 new imports into a file **#11276 is actively editing**. Instead the shortfall is **counted and printed on every run**:

```
note: EvmAsm/Progress/Correspondence.lean: 7 row(s) name a theorem that no
scanned registry witnesses, so those are NOT audited: rlpBytesEncodedSize_spec,
rlpFieldToU256Be_spec_within, rlpFieldToU64_spec_within, rlpListEncodedSize_spec,
rlpListNthItem_spec_within, rlp_encode_list_prefix_short_pinned_spec_within,
rlp_list_count_items_spec_within
```

⇒ A declared residual, not a silent skip — the same distinction as the printed layering exemption. **If you'd rather those 7 be a hard failure, say so and I'll add the imports and abbrevs** — it is a bigger diff into a file you have in flight, which is the only reason I didn't.

### ⛔ A name collision worth knowing about

`rlp_walk_init_spec_within` and `rlp_walk_next_spec_within` each exist **twice**, under different namespaces:

| | |
|---|---|
| `EvmAsm.Rv64.RLP.…` (`Rv64/RLP/WalkInit.lean:1590`, `WalkNext.lean:3924`) | the form-generic routine spec — what `Correspondence` names |
| `EvmAsm.Evm64.account_rlp_walk_…_spec_within` (`AccountAccessorSpec.lean`) | the account-specialised triple — what I registered |

Both are real, neither subsumes the other, and the unqualified names are identical. My rows now say so explicitly, because reading one as the other would overstate coverage. It also means my `rlp_walk_init` / `rlp_walk_next` rows are narrower than their `symbol` field suggests on its own — the generic specs are now audited too, via the correspondence registry rather than from `Routines.lean`, so there is no duplicate abbrev.

Verification unchanged and re-run after the source-list change: gate **OK at 111 witnesses**, classical axioms only; the two independent derivations (gate's own grep vs the generator) still produce **identical** 111-name sets; `lake build` clean (4798 jobs); negative control still bites — a `sorry`-carrying witness fails with `sorryAx`.

pirapira:
Looks good. Merging.

Issue #10688, opened 2026-07-27
sail-equiv: jalr_sail_equiv is vacuous — extracted currentlyEnabled has no Ext_Zicsr arm, so execute_JALR faults in every state


dhsorens:
## Summary

`jalr_sail_equiv` is **vacuously true**. Its `h_elp` premise is unsatisfiable in *every* SAIL state, so no caller can ever apply it. The same holds for `jalr_step_sail_equiv` (`StepProofs.lean`) and the `.JALR` arm of `instrSideCond` / `step_execute_sail_sim` (`StepSim.lean`).

This is not a proof defect — those theorems are correctly proved. It is that the thing they are conditioned on cannot happen, so they establish nothing about JALR.

## Root cause: a missing match arm in the extracted model

`currentlyEnabled` (`vendor/sail-riscv-zkvm-lean/Out/PlatformConfig.lean:1427`) has **60 explicit `.Ext_*` arms and no `.Ext_Zicsr` arm**, so `currentlyEnabled Ext_Zicsr` falls through to the generated catch-all (`assert false …; throw`) and errors unconditionally.

The chain to JALR:

- `Ext_Zicfilp` is gated on `currentlyEnabled Ext_Zicsr` → always errors
- `update_elp_state` queries `Ext_Zicfilp` → always errors
- `execute_JALR` begins with `update_elp_state` → **faults in every state**

`Ext_Zicsr` is present in the `extension` inductive (`Out/Defs.lean:338`) and `hartSupports Ext_Zicsr = true` (`Out/PlatformConfig.lean:999`), so this is a model-extraction/scoping gap rather than a modelling decision — the same family as the `sail_model_init` defect in `docs/sail-init-scoping-defect.md`.

## Evidence (kernel-checked, no `sorry`)

Proved in `EvmAsm/Rv64/SailEquiv/RunInv.lean` (PR #10686):

```lean
theorem update_elp_state_error (rs1 : regidx) (s : SailState) :
    ∃ e, update_elp_state rs1 s = .error e s
```

and the direct consequence:

```lean
theorem jalr_sideCond_unsatisfiable
    (sRv : MachineState) (sSail : SailState) (rd rs1 : Reg) (offset : BitVec 12) :
    ¬ instrSideCond (.JALR rd rs1 offset) sRv sSail := by
  rintro ⟨⟨s_mid, hok, _⟩, _⟩
[...]

PR #11279, opened 2026-08-03
feat(#11241): separation-logic predicates for the regions that stay

### LINKED ISSUE #11241: proof: define separation-logic predicates for the regions documented in #11237
## What

Define the separation-logic predicates for the regions documented in #11237 (`docs/4ch8f-predicate-shapes.md`), and prove the obligations each one carries. That document is the **measured input**; this issue is the **Lean work** that consumes it.

Landing the structural changes was the stated prerequisite for starting these proofs, because a predicate cannot be authored against a shape that is about to move. #11237 records the shapes that are **not** changing, so those can begin immediately.

## Why this is worth doing beyond tidiness

⭐ **A descriptor predicate catches wrong-base defects structurally, where a corpus row catches them by luck.**

Concrete case from tonight: the shared precompile helper in #11236 read the MODEXP exponent at offset `96`, where `modexp.py:36` gives `exp_start = 96 + base_length` — cost `1536` against the correct `1152` on fixture 00041. **The formula was right and the pointer was wrong.** It surfaced as *one* false reject, on a *targeted* fixture that is not in the random-200 sample, and it was nearly waved through as refactor churn.

⇒ A predicate stating where each field of the precompile input descriptor lives would have failed on **all 18 arms at once**, including the arms with no corpus witness at all. That is the difference between a backstop and a coincidence.

## The template (from #11237, uniform across regions)

Per region: base and extent (with `capacity × stride` checked against the extent **exactly**); the entry layout field by field; whether the count is **live entries** or a **high-water mark**; the validity invariant on entry *i*; ownership and lifetime (per-transaction vs per-block, who resets and when); disjointness / what may alias; the **initialisation contract** — which cells must be zero on entry to which routine, what breaks otherwise, and **which entry points have been audited**; and **the build unit the predicate describes**.

## Array-shaped regions: follow `evmStackIs`'s structure, not its name

The repo already has the shape we want. `EvmAsm/Evm64/Stack.lean:43`:

```
def evmStackIs (sp : Word) (values : List EvmWord) : Assertion :=
  match values with
  | [] => empAssertion
  | v :: vs => evmWordIs sp v ** evmStackIs (sp + 32) vs
```

**What to copy** — base as a parameter, **contents as a `List`**, recursion on the list with `**` and a stride step, `empAssertion` at nil. Contents-as-a-list is what makes functional specs possible: an append is `xs ++ [x]`, a lookup is `xs[i]`, and a spec can say **what the buffer holds afterwards** rather than only that it is uncorrupted. Also copy the companions that already exist for stack words: an **ownership form** carrying no values (`evmWordOwn`), the `Is → Own` weakening (`evmWordIs_to_evmWordOwn`), and the `pcFree` lemmas. ⇒ *The ownership form is exactly the "dissolve into an arbitrary blob" state a region enters at a phase boundary* — no new machinery needed for union children.

⚠️ **What not to copy: the name — and this is guidance for the NEW predicates only.** `evmStackIs` stays as it is; it is load-bearing, widely used, and renaming it buys nothing. But note why it reads badly before naming anything new: `evmStackIs sp values` sounds like *the stack is `values`*, while it actually asserts **a run of consecutive words starting at `sp`** — frequently just the head elements, with the rest of the stack neither described nor owned. **The name claims a totality the assertion does not have.**

⭐ **So for the new regions, name the two forms for what they actually assert:**

| form | asserts | suggested name |
|---|---|---|
| **partial run** | `xs` in consecutive entries from `base`; says nothing about what follows | `fooEntriesFrom base xs` |
| **whole buffer** | the run **and** that this is all of it — `xs.length` live, remainder owned-unconstrained to capacity | `fooBuffer base xs` |
| **ownership only** | the bytes are owned, contents arbitrary | `fooOwn base` (matching `evmWordOwn`) |

⇒ **Both forms are needed and the partial one is the composable one**: split/join and the append-extension lemma are natural on `EntriesFrom` *precisely because* it is partial, and it is what a routine touching one row should be handed. `Buffer` is what a caller holds and what weakens to `Own` at a phase boundary.
⇒ **The test for a good name here is whether it stays true when the list is shorter than the region.** `…EntriesFrom` does; `…Is` does not. Avoid `Is` for anything partial.

⚠️ **`xs.length ≤ capacity` is a low-stakes call** — inside the total assertion, or threaded as a hypothesis. Mixing across regions costs noise at the seams, not correctness; decide per region and say which. Note the fail-closed `bgeu` guards in these buffers are the operational form of the same bound.

⚠️ **Where entries are variable-length** (`baap_storage_values`), the run cannot be indexed by *i* without an alignment or offset-table invariant the structure does not currently have. Record what would be required — a finding about the structure, with a byte-range form as the honest interim.

## The rule that keeps these from rotting

⚠️ **Name the build unit each predicate describes.** `baap_storage_values` is emitted at **three different sizes in three units** (#11222), so a predicate that does not name its unit is false in at least two of them. Related: a phantom union child survived inside a kernel-checked `CallFramePhase` theorem until #11224 — *a symbol exists per image, not per repository.*

## Suggested order — cheapest proof obligation first

1. **`evm_memory_pool`** — the easy shape. Disjointness of live frames **falls out of the address formula**: the slot index is the **parent** depth and the stride is `0x19000`, so distinct depth gives distinct slot by arithmetic. No invariant for anyone to maintain, no free-list reasoning, no ordering side condition. Note the one-slot surplus (1025 allocated, 1024 reachable) explicitly rather than rounding it away.
2. **The dispatch-journal initialisation contracts** (#11227) — twelve cells with a reader that assumes zero on entry, each already named with the routine and the failure mode. These are preconditions, not layout, and they are the slot-9 material.
3. **`callee_seed_table`** and **`teer_success_table`** — both measured, both staying, both with their consumers named (including EXTCODEHASH for the latter).

4. **`baap_storage_values`** — **in scope, and easier than it first looks.** It is a **union child**: the same bytes are frame slots while frames are live, and replay values while the replay runs.

   ⭐ **The phase transition does not need a temporal control-flow proof.** At the boundary the region **dissolves into an arbitrary blob** — ownership of a raw byte range with *existentially quantified, unconstrained contents* — and is then **reconstructed into a different buffer** by establishing a fresh structured predicate over that same block. ⇒ *No content is carried across the boundary, so nothing has to be proven about the relationship between the two uses.* That is the standard reclaim-and-reinterpret shape, and it is why a union child is not a special case: `raw_bytes(base, n)` is the interface between phases.

   So the obligations are the ordinary ones: **establish** the structured predicate from `raw_bytes` when a phase begins, and **release** back to `raw_bytes` when it ends. The only thing worth stating explicitly is that the two structured predicates over this block are **never simultaneously held**.

   ⚠️ Its remaining properties belong **in** the predicate rather than deferring it: entries are **variable-length**, so an entry-indexed predicate needs an **alignment invariant the structure does not currently have** — if the proof needs one, that is a finding about the *structure*, not a blocker; and **no validity invariant is asserted anywhere** (no sortedness, uniqueness, or monotone BAI), so a predicate asserting nothing beyond the byte range is the honest starting point.

   ⇒ Its initialisation contract is **self-discharged** rather than absent — the routine clears the cells in its own entry block instead of imposing the obligation on callers. *An empty slot meaning "no obligation" and an empty slot meaning "obligation already met" are different facts, and only one survives a refactor.*

   ⚠️ #10651 may change this region's enumeration. Write the predicate against the current shape and **state the commit it was derived from** — a predicate that gets revised is normal; one that was never written teaches nothing.

## Done when

Each listed region has a predicate in Lean naming its build unit, with its initialisation contract stated; and the regions that **cannot** carry a clean predicate are recorded as structural findings with what would have to change first.

dhsorens:
> Replaces #11274 (same commits, `e572d4457`). Moved from a fork branch to an in-repo branch so `summarize` can run — it fails on every fork PR (#10703). Review discussion from #11274, including @pirapira's SpecRef request and the response, is carried over in the first comment below.

---

Closes #11241 (see **Scope** — two regions carry findings rather than predicates, which the issue's "Done when" allows for).

Predicates for the regions in `docs/4ch8f-predicate-shapes.md`. New module `EvmAsm/Codegen/RegionPredicates.lean`; derivation commit `7185e9274` is named in its header, per the #10651 caveat.

## ⭐ First: two of the five items were already done, and I did not redo them

This is the most useful thing I found, so it goes first. The issue's suggested order is **partly redundant with existing work**:

| issue item | already in the tree |
|---|---|
| 1, the `call_frame_arena` slot/stride disjointness | `EvmAsm/Codegen/CallFrameWindows.lean` — `frameBase base d` is the address formula, and `phaseDView_eq_framesTiling` states the arena **is** the separating conjunction of its 1025 equal-stride windows. That is "disjointness falls out of the address formula" in its strongest form, with `focusFrame` / `unfocusFrame` for extract / re-absorb. |
| 4, the union-child phase boundary | `EvmAsm/Codegen/CallFramePhase.lean` — `phaseD_eq_phaseH` **is** the dissolve-and-reinterpret theorem the issue describes (the two views are literally the same assertion, so nothing has to be proven relating the two uses); `phaseH_to_phaseD` is the release direction; `children_offsets_match_regionMap` pins the child offsets. |

Also: ⭐ **`anyBytes` (`Rv64/SAsm/PhaseSplit.lean:47`) already is the `raw_bytes(base, n)` interface** the issue asks for, already `pcFree`, already carrying the tiling theory (`anyTilesAt`) that the union-child work uses. I started to relocate `bytesRegionAny` from `WithdrawalDecode.lean` for this purpose, then found `anyBytes` is **definitionally identical** to it — so I reverted and reused `anyBytes` rather than adding a third copy. ⚠️ Worth a separate cleanup: the repo has two identical definitions of this predicate (`anyBytes`, `bytesRegionAny`); I left that alone to keep this diff about predicates.

⇒ So this module covers **only what had no assertion over it at all**, which I verified by search rather than assuming: no separation-logic `Assertion` existed over `evm_memory_pool` (only Nat-level budget arithmetic in `MemoryBudgetGuard`), none over the DJ cells, none over `teer_success_table`.

## What is here

**`evm_memory_pool`** (doc §4) — `poolOwn` / `poolWindow`, extent from `evmMemoryPoolBytes` with the build unit named (⚠️ `CallFrameDescend.lean`'s probe emits 1 MiB under the same symbol, so an unqualified `extent =` claim is false there). `poolWindow_pair` derives two-window separation from `anyTilesAt`'s structure rather than a fresh disjointness argument.

⚠️ **A conflation in the issue worth correcting:** item 1 attributes the `0x19000` stride and the parent-depth slot index to `evm_memory_pool`. Those are the **arena**'s (doc §1). The pool has no stride at all and is a pairwise-disjoint *sibling* of the arena, not part of it. A predicate written to the issue's wording would be about the wrong region. The surplus-slot fact is an arena fact and is stated as such: `frameSlot_surplus_one : frameSlotCount - maxCallDepth = 1`, with `frameSlot_index_lt_maxCallDepth` for why (index is the *parent* depth, so a child at `d ∈ 1..1024` uses slot `d-1 ∈ 0..1023` and slot 1024 is allocated-unreachable).

**DJ-01..DJ-12** (doc §7) — `dispatchJournalZero`, twelve `globalCellIs a 0` conjuncts, recursion with `**` and `empAssertion` at nil; `dispatchJournalZero_cons` is the opening lemma of a dispatcher triple (doc §9.4). The wipe set is named in DJ order. The #11147 exception is encoded as *data*, not a comment: `dispatchJournalSystemSurvivors` lists the two cells that survive system re-entry, and `dispatchJournalSurvivors_disjoint` proves they are genuinely outside the twelve — an empty slot meaning "no obligation" and one meaning "obligation already met" are different facts.

**`teer_success_table`** (doc §6) — the naming the issue asked for, and deliberately not `…Is` for anything partial: `teerEntriesFrom base xs` (partial run, the composable form), `teerBuffer base xs` (the run *and* that this is all of it), `teerOwn base` (bytes owned, contents arbitrary). Structure copied from `evmStackIs` — base as a parameter, contents as a `List`, `**` with a stride step — but not its name. `xs.length ≤ teerCapacity` is threaded as a hypothesis rather than folded in; I named the choice rather than leaving it implicit.

⭐ **The extent is pinned to the emitted image, not just to itself.** `teerTableBytes_eq : 1060 * 32 = 33920` alone would *not* do that: the data section emits a **hardcoded** `.zero 33920` rather than interpolating `teerSuccessfulAuthCapacity`, so the derived extent and the emitted literal are two independent numbers that could drift apart silently. A `#guard` in the repo's exactly-once `splitOn` idiom ties them. I checked the guard bites by setting the stride to 33 and watching it fail. (The append path's `bgeu` bound *does* interpolate the constant, so that end was already drift-proof; the arena size was not.)

## Scope — two regions carry findings, not predicates

⛔ **`callee_seed_table` (doc §5) does not exist in the tree.** `callee_seed_table`, `callee_seed_count`, `seed_callee_storage`, `calleeSeedTableCap`, `calleeSeedTableBytes`, `calleeSeedEntryBytes` — **zero occurrences** anywhere in `EvmAsm/**`. No emitter, no symbol, no capacity constant. The only mentions in the repository are `docs/4ch8f-predicate-shapes.md` §5 and `docs/4ch8f-region-map.md:177`. There is no `#guard` for `128 * 96`; the only `12288` in Lean is `systemCallReturndataMaxBytes` (`Dispatch.lean:74`), an unrelated constant of coincidentally equal value — which is exactly what makes that doc row look verifiable when it is not. **What would have to change first:** the region has to exist in some build unit with an emitter and a capacity constant before a predicate can name a base or an extent. I did not invent one. (Doc §6 next to it checks out exactly, so this is a defect in one row, not in the document.)

⚠️ **`baap_storage_values`** — byte range only, which is what the structure supports. Entries are variable-length blobs packed by a byte cursor (only the *paths* are fixed 64-byte records, in the sibling region), so an entry-indexed form needs an alignment or offset-table invariant the structure does not have; and no validity invariant is asserted anywhere (no sortedness, uniqueness, or monotone BAI), so a predicate claiming more would claim something the code does not maintain. Both are findings about the structure. The three-extents-in-three-units overload (#11222) is recorded as checkable data — `baapStorageValuesUnitSizes` with `baapStorageValues_sizes_distinct` proving the three really differ, and `baapStorageValues_guest_size_eq_formula` tying the guest size to `bsrMaxBalItems * bsrPathBytes` rather than a literal.

## Verification

- `lake build` clean tree-wide (4798 jobs).
- `#print axioms` on all 12 new theorems: `propext` / `Quot.sound` only, or axiom-free. No `sorryAx`, no trust axioms.
- Gates pass: `check-{layering,axioms,forbidden-tactics,unimported,file-size,no-warnings,progress,naming,statement-tamper,duplication}.sh`. Note the module is under `EvmAsm/Codegen/`, which is where the region constants live and which may import freely — so `check-layering` needs **no exemption**.
- ⭐ **Satisfiability, per the #10688 lesson** that a bundled hypothesis can make a theorem vacuous: the entry model is instantiated at concrete data and the layout is *checked* — the flag lands at `+20` where the EXTCODEHASH consumer reads it, the pad at `+31` is zero, and a negative control shows an uncharged row differs in exactly that byte.
- `check-region-map.sh` fails locally for a missing `readelf` (environmental, before it does any work) — CI has the toolchain.
- No `maxHeartbeats`. No `set_option maxRecDepth` was needed.

Expect the `summarize` check to show a red X on fork PRs — known, tracked in #10703.


dhsorens:
## Carried over from #11274

Verbatim, so the review trail survives the fork→in-repo move — in particular @pirapira's SpecRef request, which is what commit `e363fca70` answers. Attribution is in each heading.

---

### @pirapira (2026-08-03)

If you can find a teeEntry-equivalent in RefSpecs and reuse that, that would be great (optional).

### @dhsorens (2026-08-03)

Done in e363fca70 — and it turned up something worth flagging, because the obvious reuse would have been wrong.

## ⛔ There is no single `TeerEntry`-equivalent struct, and the nearest-looking one is a trap

`SpecRef.Transactions.Authorization` (`SpecRef/Transactions.lean:65`) looks like the match, but it is the **signed input tuple** — `chainId, address, nonce, yParity, r, s`. Its `address` field is the delegation **target**. This table is keyed on the **recovered authority** (`recover_authority auth`, `SpecRef/Interpreter.lean:139`), which is a different address entirely. Reusing `Authorization` here would have been a silent correctness bug dressed as a naming improvement.

⭐ **The genuine equivalent is not a struct but the two `List Address` locals inside `SpecRef.Interpreter.set_delegation`** (`:197`): `written_accounts` (ACCOUNT_WRITE once per authority) and `delegation_set_for` (AUTH_BASE once per authority). That explains the shape mismatch that made the question hard: the guest has **one** 32-byte row where the spec has **two** lists, because membership carries the first fact and the `+20` word carries the second.

## What I reused, and what I proved

**Type reuse:** `TeerEntry.authority : Stateless.SpecRef.Address`. That is definitionally the same `List (BitVec 8)` (`Address := Bytes := List Byte`, `Byte := BitVec 8`), so there is no representational change — the field's type now states which spec notion it carries, and drift becomes a type error.

**Correspondence, as theorems rather than prose:**

- `teerWrittenAccounts` / `teerDelegationSetFor` — project the table into the spec's two lists.
- `teerDelegationSetFor_subset_written` — the spec invariant that `delegation_set_for ⊆ written_accounts`. Worth noting it **falls out of the shared projection**, which is the payoff for deriving both from one table instead of modelling two regions.
- ⭐ `teerScan_hit_implements_setAdd` / `teerScan_miss_implements_setAdd` — **the guest's append-if-absent linear scan *is* `SpecRef.setAdd`.** The emitter walks `teer_success_count` rows comparing the authority and appends only on a miss; the spec writes `written_accounts := setAdd written_accounts authority`. Split per branch because that is the form a caller discharges. This also explains why no sortedness invariant is needed: `setAdd` appends at the tail and keeps first-insertion order.

**And the shape I think you were pointing at:**

```lean
def teerRepresents (base : Word)
    (written delegated : List Stateless.SpecRef.Address) : Assertion :=
  fun ps => ∃ xs : List TeerEntry,
    (∀ e ∈ xs, e.wf)
      ∧ teerWrittenAccounts xs = written
      ∧ teerDelegationSetFor xs = delegated
      ∧ teerEntriesFrom base xs ps
```

An `Assertion` whose **arguments are `SpecRef` data**, saying this region of memory represents that spec value, with the concrete rows existentially quantified away. `teerEntriesFrom` describes bytes; `teerRepresents` says which spec state those bytes are a representation of. A caller reasoning about `set_delegation` never has to mention the byte layout. `teerEntriesFrom_represents` is the byte-level introduction, and `teerRepresents_subset` shows the invariant survives the abstraction — so nobody can obtain a `teerRepresents` for an inconsistent pair of spec lists.

## Note on the pattern

⇒ For anyone applying this elsewhere: this is **not** a new pattern, it is `AccountAssertions.lean`'s. `accountRlpIs (ptr) (a : AccountRecord) : Assertion` takes the struct, and `decode_account_from_leaf_accountRlp` bridges it to `SpecRef.decode_account_from_leaf` — and that bridge theorem is already an axiom-gate witness. What is new here is only that the Assertion's *arguments* are SpecRef values rather than a local record. Before this, no `Assertion` in the tree referenced `SpecRef` at all.

⚠️ I did **not** retrofit the pool or DJ predicates. `evm_memory_pool` is an untyped byte arena with no spec-level struct to represent, and the DJ cells are `= 0` preconditions rather than encoded data — for both, a SpecRef argument would be decoration. Say if you'd rather I look harder at either.

Verification: `lake build` clean (4798 jobs); `#print axioms` on all 7 new theorems gives `propext` / `Quot.sound` only; gates green. The correspondence is witnessed at concrete data with a **strict** subset — two authorities written, one AUTH_BASE-charged — so `teerDelegationSetFor_subset_written` is not vacuous on an always-empty list.

pirapira:
Looks great. Merging.

Issue #10780, opened 2026-07-28
Proof: semantic specs for the RLP encode side — rlp_encode_uint_be and rlp_encode_bytes absent, rlp_item_size/rlp_encode_list_prefix long forms uncovered (2 of 7 already done)


pirapira:
Proof work on a **stable leaf**. Largest of the four filed today, and the one with the most routines.

## Current state, measured on `origin/main`

`EvmAsm/Codegen/Programs/RlpRead.lean` — **835 lines**, 7 emitted routines plus 2 scratch symbols:

| routine | side |
|---|---|
| `rlp_list_nth_item` | read |
| `rlp_list_count_items` | read |
| `rlp_item_size` | read |
| `rlp_item_span` | read |
| `rlp_encode_list_prefix` | write |
| `rlp_encode_uint_be` | write |
| `rlp_encode_bytes` | write |

(`reb_scratch`, `ris_scratch` are the two scratch symbols.)

**What already exists:** 7 theorems of the form `rlpListNthItemFunction_eq_prog` — **string↔Program correspondence** only. The emitted string equals the Lean-rendered `Program`; nothing states what either computes.

**What is missing:** semantic specifications for all seven.

## The task

Spec triples per routine, derived from RLP as defined in `execution-specs`.

The **encode** side is the higher-value half and can be taken independently of the read side — it is what the BAL serializer builds on, and canonical-form errors there are silent:

- **A length-of-length (`0xb7`/`0xf7` long form) carries no leading zeros.** A long-form header with one still *parses* and hashes differently — wrong in a way no decoder complains about. Worth being an explicit side condition of `rlp_encode_list_prefix`.
- `rlp_encode_uint_be` on a canonical scalar: zero encodes as the empty string (`0x80`), and no encoding carries a leading zero byte.
- The 55-byte / 56-byte boundary between short and long form is where off-by-one lives; a spec that pins both sides of it is worth more than one that pins neither.

Suggested order: `rlp_item_size` → `rlp_item_span` → `rlp_encode_uint_be` → `rlp_encode_bytes` → `rlp_encode_list_prefix` → `rlp_list_count_items` → `rlp_list_nth_item`.

## Test coverage today

⚠️ **No dedicated `zisk_*` probe.** Exercised indirectly through many consumers — the BAL serializer, header decoding, transaction decoding. Wide indirect exposure, no direct oracle.

## Prospect of change

[...]

PR #11281, opened 2026-08-03
fix(correspondence): rlp_encode_uint_be is no longer unproven

dhsorens:
The `rlp_encode_uint_be` correspondence row still read:

```lean
verdict := .unproven, basis := .none,
note := "model layer complete (reubOut_eq_encode_toBytesBE) but NO whole-routine triple, …"
```

**#11040 landed `reub_spec_within` on 2026-07-31**, so that row has been contradicted by the tree ever since. Found while starting item 2 of #10780 and reading the neighbouring `rlp_encode_bytes` row. This is my own registry (#10949) going stale, which is the failure mode I have been careful about in PR bodies and then missed in the artifact designed to prevent it.

## The fix

`spec := some "reub_spec_within"`, `verdict := .domainRestricted`, `basis := .bridged`.

Both values are forced by the enums' own definitions rather than chosen:

- **`domainRestricted`** — "agrees, but the spec covers strictly less input than the reference accepts." The reference `encode(Uint)` takes any length; the spec needs the stripped payload `≤ 55`.
- **`bridged`** — "stated over, or tied by a cited bridge lemma to, the shared model." `reub_spec_encode_within` restates the post over `encodeBytes ∘ toBytesBE ∘ fromBytesBE` via `reubOut_eq_encode_toBytesBE`.

⭐ The old note had written down **its own falsification condition** — *"the ≤55 domain restriction is a documented property of the model, not a proven property of the routine — `unproven`, not `domainRestricted`"* — and that condition is now met. That is why the replacement values are unambiguous rather than a judgement call, and it is an argument for writing notes that say what would change them.

`verdict_counts` and `basis_counts` move with it (`domainRestricted` 2→3, `unproven` 5→4; `bridged` 5→6, `none` 5→4). Those pinned totals are exactly why this could not be a silent one-field edit — the registry made me touch a theorem a reviewer will see.

All other invariants unchanged and still `decide`-checked, including `verdict_requires_spec`, `basis_diff_requires_spec` and `no_looser_verdicts`.

## The gap that let it go stale — worth a mechanism, not just a fix

`verdict_requires_spec` catches *"the row claims a verdict but names no spec."* It **cannot** catch the reverse: a spec exists in the tree while the row still says `unproven`. The registry has no way to know which theorems exist, so this direction is structurally invisible — the same class as #11042.

The cheap mechanism is a **cross-registry check** once #11273's `RoutineEntry` lands: *any routine with a witnessed `RoutineEntry` must not be `.unproven` in `Correspondence`.* Both registries would then have to be wrong in the same way for a stale row to survive. I have not built it here because that registry is still in flight and it is not mine to extend mid-review — flagging so it is tracked rather than implicit.

## Verification

`lake env lean EvmAsm/Progress/Correspondence.lean` clean; every count and invariant theorem still elaborates by `decide`. No other file touched.

Reopened from #11276 as a branch on this repository rather than my fork, so `summarize`
actually runs instead of failing in 7s at the `pull_request_target` fork-checkout refusal
(#10703). Same commits; #11276's discussion is worth reading for the layering measurement.


pirapira:
Looks good. Merging.

PR #11283, opened 2026-08-03
docs(#11241): state evm_memory_pool's initialisation contract

dhsorens:
Follow-up to #11279 (merged). Small, docs-only.

## Why

#11241's *Done when* asked that each region with a predicate name its build unit **and state its initialisation contract**. Auditing #11279 against that after it merged, two of three were covered and one was not:

| region | init contract in #11279 |
|---|---|
| DJ-01..DJ-12 | ✅ "zero on dispatcher entry" — which *is* the whole contract |
| `teer_success_table` | ✅ count zeroed at `eip7702_auth_state_prepare` entry |
| `evm_memory_pool` | ❌ **missing** |

So this closes the one bullet #11279 left unmet, rather than leaving a COMPLETED issue with a gap in it.

## What it says

⭐ **The contract is that there isn't one.** Nothing zeroes the 96 MiB; frames write their own windows, and the depth-1 reset paths restore *limits*, not contents. A routine reading a pool byte it has not written reads whatever the previous frame at that offset left there.

That is worth stating rather than omitting, because it explains a choice that otherwise reads as an oversight: **`poolOwn` is the strongest predicate available for this region, and there is deliberately no `poolZero` companion.** For an `evmMemZero`-style region a zero predicate is the natural opening resource; here it would be **unsound** — it would let a caller conclude a read returns `0` from a cell nobody initialised. Absence-as-contract is the kind of thing a later reader "fixes" by adding the missing definition, so the reason is now written down next to it.

⇒ It also records that the number of **audited entry points is `none`**, and that this is a different fact from *"not yet audited"* — for the DJ cells the contract names `emitRuntimeDispatcherCallableSetup` as what discharges it; here there is nothing to audit. Per #11241's own point that an empty slot meaning "no obligation" and one meaning "obligation already met" are different facts, and only one survives a refactor.

## Verification

Docs-only — no statement or definition changes, so nothing to re-prove. `lake build` clean (4800 jobs); `check-{axioms,layering,forbidden-tactics,unimported,no-warnings,file-size}.sh` all pass.


pirapira:
Looks good. Merging.

Issue #11280, opened 2026-08-03
Gate: guest_image_coverage checks ordering, not declared-versus-actual extents


pirapira:
## Finding

`guest_image_coverage.py` checks actual-symbol gaps/order and converted-program overruns, but it does not verify that each generated `GuestImageEntries` start/extent declaration matches the actual linked symbol.

That distinction mattered on #11277: the candidate ELF shifted 222 linked declarations by +0x2c4 (first shifted symbol `codes_blockhash_required_headers` at `0x80014ef4`). The committed `GuestAddrs` were stale, so `GuestImage.lean`'s `CodeReq.extentsOkFrom` failed while `guest_image_coverage.py --gaps` still passed. Recomputing against the candidate ELF (SHA `2f27c32d20d3a59719694811ba95193c9e438bdd26e1039d873f8a433bf3d5cd`) and the 346 `#guard` program lengths gave zero real extent overruns.

A checker that verifies ordering is not a checker that verifies extents.

## Proposed gate improvement

For each converted, linked entry:

1. Parse the generated `GuestAddrs.lean` declaration.
2. Compare the declared start with the actual `symbol-addresses.tsv` start; report the entry and delta on mismatch.
3. Compare `declared_start + program_bytes` with the actual next-symbol boundary (and `.text` end for the final entry).
4. Retain the existing actual-address overrun check.

Restrict the check to converted linked symbols; unlinked aliases should not fail it.

pirapira:
Shipped: https://github.com/Verified-zkEVM/evm-asm/pull/11284

- `python3 scripts/guest_image_coverage.py --check-declared-starts`
- wrapper `scripts/check-guestaddrs-starts.sh`
- **Wired hard** into `source-checks` + check-build-parallel (explicit choice, not advisory)
- Clean: 346/346 ok
- Perturb `bal_account_change_value` +0x2c4 → `DECLARED_START_MISMATCH … delta=+0x2c4` exit 1

No merge! from grok.

pirapira:
## Re-verified on current main

Current main: `6d1647824`; carrier **#11284** is merged.

Evidence:

- `scripts/check-guestaddrs-starts.sh:1-22` is the CI wrapper and invokes `guest_image_coverage.py --check-declared-starts`.
[...]

PR #11284, opened 2026-08-03
fix(codegen): gate GuestAddrs declared starts vs TSV (#11280)

pirapira:
## Summary

Implements **#11280** (codex design): a gate that checks **declared** `GuestAddrs` starts against **actual** `symbol-addresses.tsv` for every converted, linked guest entry.

### Why

`guest_image_coverage.py --gaps` walks **TSV actual** addresses only. On #11277:

- 222 `GuestAddrs` entries were stale by **`+0x2c4`**
- TSV was current
- `CodeReq.extentsOkFrom` failed
- `--gaps` stayed green

A checker that verifies ordering is not a checker that verifies declared-versus-actual starts.

### What

| piece | role |
|---|---|
| `guest_image_coverage.py --check-declared-starts` | parse `GuestAddrs.lean` `def name : Nat := 0x…`; for each converted+linked entry compare to TSV |
| `scripts/check-guestaddrs-starts.sh` | thin CI wrapper |
| `.github/workflows/build.yml` source-checks | **wired hard now** (not advisory) |
| `check-build-parallel.sh` | same |

**Population:** converted + linked only. Unlinked aliases do not fail.

**Fail tags:**
- `DECLARED_START_MISMATCH entry=… declared=… actual=… delta=…`
- `DECLARED_MISSING`
- `DECLARED_EXTENT_OVERRUN` (declared start + prog bytes past next TSV symbol)

### CI choice (explicit)

**Hard gate in source-checks now.** Pure source (no ELF build), cheap, and the whole point of #11280 is that the predecessor could not fire. Not shipped as investigation-only (#11270-style) because a silent green is exactly the defect class.

### Baseline + perturb evidence

| leg | result |
|---|---|
| clean main | `linked_converted=346 ok=346 fail=0` exit 0 |
| shift `bal_account_change_value` by **`+0x2c4`** (historical #11277 delta) | exit 1: `DECLARED_START_MISMATCH entry=bal_account_change_value declared=0x8000f498 actual=0x8000f1d4 delta=+0x2c4` |
| restore | exit 0 |

Perturb fires and **names the entry** — same discipline as #11282.

### Not in scope

- Does not replace ELF↔LinkPins / CodeReq (different artefacts)
- Does not auto-regen GuestAddrs (regen path unchanged: `asm_to_program.py guest-addrs`)

### Related

- #11280
- #11277 (the invisible stale GuestAddrs)
- #11282 (form fence sibling)

NO merge! from author.


pirapira:
Parked `complicated-merge`: this PR **breaks the Build workflow at startup**, so CI produces no logs at all — which makes it look like CI simply never ran.

`.github/workflows/build.yml` gains a second `run:` in an existing step:

```yaml
      - name: Symbolic guest-PC guard (SAsm wrappers follow GuestAddrs)
        run: scripts/check-no-hardcoded-guest-pc.sh
        run: scripts/check-guestaddrs-starts.sh      # <-- duplicate key in the same step
```

Two consequences:

1. **GitHub rejects the workflow.** The run completes with `conclusion=failure` and **0 jobs created**, and `gh run rerun` refuses it (`This workflow run cannot be retried`). I re-triggered with a fresh push and it reproduced immediately, so it is not transient. There is no build log to read, which is why this reads as "CI didn't start".
2. **Even if it were accepted, it would silently drop a guard.** YAML takes the last duplicate key, so `check-no-hardcoded-guest-pc.sh` would stop running while the step kept its name.

The fix is presumably a separate step:

```yaml
      - name: Symbolic guest-PC guard (SAsm wrappers follow GuestAddrs)
        run: scripts/check-no-hardcoded-guest-pc.sh
      - name: GuestAddrs declared-starts guard
        run: scripts/check-guestaddrs-starts.sh
```

I have not applied it — editing CI workflow definitions is outside the mechanical layout repair I do, and the step name/placement is your call.

**#11277 and #11285 were landed separately** via #11292 (verified green: `lake build` + all 15 gates). #11287, which had all three, is closed — its build was the startup failure above.

PR #11287, opened 2026-08-03
chore: batch-merge #11277 #11284 #11285

pirapira:
Batch-merge of #11277 #11284 #11285 (all author pirapira).

Verified locally: `lake build` + all 15 `check-*.sh` green, including `check-build-units-link`, `check-asm-to-program` and `check-region-map`.

Layout regenerated via the #11230 flow (`gen-region-map-link-pins.py`): second generator pass is a verified no-op and `--check` reports CLEAN. All three merged with no conflicts.

Batched because each was individually landing only to be pushed `BEHIND` by the next merge; one CI cycle covers all three.


pirapira:
Superseded by a rebuilt batch without #11284.

#11287's `build` never ran: the workflow run completed with `conclusion=failure` and **0 jobs created** — a startup failure, which GitHub refuses to retry (`This workflow run cannot be retried`). Re-triggering with a fresh push reproduced it immediately, so it was not transient.

Cause: #11284 introduces a duplicate `run:` key in a single step of `.github/workflows/build.yml`. It parses as YAML (last key wins, silently dropping the guest-PC guard) but GitHub rejects the workflow. Details on #11284; rebuilt as the PR above with #11277 + #11285 only.

PR #11300, opened 2026-08-03
proof(#10780): rlp_encode_bytes machine blocks — every instruction covered

dhsorens:
Item 2 of #10780 (`rlp_encode_bytes`), **stage 2 of 3**: with stage 1's model layer (#11275, merged), **every one of the 76 instructions is now covered by a block theorem.** The whole-routine claim is still absent — that is stage 3 — and the distinction is categorical, not fractional: each instruction is verified individually, and the claim that the routine computes RLP is not yet made.

## Coverage map

| instrs | theorem | notes |
|---|---|---|
| [0]-[4] | `rebPrologueNe1` / `rebPrologueEq1` | one lemma per exit |
| [5]-[7] | `rebRawProbeSmall` / `rebRawProbeLarge` | byte vs `0x80` |
| [8]-[12] | `rebRawTail` | EXIT A, `*a3 = 1` |
| [13]-[14] | `rebDispatchShort` / `rebDispatchLong` | len vs 56 |
| [15]-[18] | `rebShortHeader` | `0x80 + len` |
| [19]-[25] | `rebShortCopyLoop` | `7·len + 1` |
| [26]-[29] | `rebShortTail` | EXIT B, `*a3 = len + 1` |
| [30]-[51] | `rebLadder` | `3·bc`, own module |
| [52]-[55] | `rebLongHeader` | `0xb7 + bc`, seeds counter `bc − 1` |
| [56]-[62] | `rebLolLoop` | `7·m + 1`, region = `writeShift` |
| [63] | `rebLongCopySetup` | |
| [64]-[70] | `rebLongCopyLoop` | `7·len + 1` |
| [71]-[75] | `rebLongTail` | EXIT C, `*a3 = bc + 1 + len` |

Three modules: model layer (`RlpEncodeBytesSAsm`, stage 1 + additions here), the ladder (`RlpEncodeBytesLadderSAsm`, 943 lines), the blocks (`RlpEncodeBytesBlocksSAsm`, 1140 lines) — all under the hard 1500-line cap, precedent `WithdrawalDecodeClose → Close2..5`.

## The pieces worth review attention

**The length-of-length loop is the one non-port.** Its counter is **signed**: it runs `bc−1` down to `−1`, so the invariant carries `x29 ↦ ofNat 64 m − 1` — at `m = 0` that is `−1` and the `BLT` fires (`slt_neg_one`); at `m+1` it is `ofNat m` and falls through (`slt_small_false`). Two findings baked into those bridges:

- `slt` normalises through `Int.bmod`, which **both `omega` and `bv_omega` treat as opaque** — the natural `i < 2^63` bound is unprovable by either. The lemma instead carries `i < 8`, which is the truth of the situation (the counter starts at `bc−1 ≤ 7`) and documents why the loop terminates in ≤ 8 iterations.
- The region is tracked by `writeShift` (stage-1's mirror of `copyN`), whose contents equal `Nat.toBytesBE` via `beShift_eq_toBytesBE` — so stage 3's canonicity claim (no leading zero in the length-of-length) will be a *consequence*, not a side condition.

**The two payload copy loops are ports of `reubCopyLoop`, and deliberately two proofs.** `rebBase` is a concrete numeral, so `runBlock` discharges code membership by evaluation — a lemma generic over the loop-head address leaves membership unprovable. Same reason `reubCopyLoop` and `cu256_loop_spec_within` coexist. ⭐ Their scratch registers also differ **and it is load-bearing**: the short loop scratches `x28`, the long loop scratches `x30` because `x28` still holds `bc`, which `[71]` needs for the written length `bc + 1 + len`. Recorded in the module header so nobody "unifies" them.

**The `*a3` dword out-param** (new versus item 1: two output locations) is asserted as `cellPtr ↦ₘ v` and stored with `sd_spec_gen_within`, following `RlpSpliceHelperSpec`'s identical two-output ABI shape.

**Blocks stay Word-level** (hypotheses like `len ≠ 1`, `len.toNat < 56`); the Nat/model bridging is stage 3's job. Data-driven branches are one lemma per exit — straight-line triples with the other arm discharged by pure-fact contradiction — so the composition can split on the data first.

## Verification

- `lake build` clean: zero errors, zero warnings.
- `#print axioms` on all 16 block theorems plus `rebLadder`: `propext` / `Classical.choice` / `Quot.sound` only. No `sorry`, no `bv_decide`/`native_decide`; `maxRecDepth 8000` matches the sibling modules, no `maxHeartbeats`.
- `check-forbidden-tactics` / `check-file-size` / `check-layering` / `check-guest-addrs` / `progress-report.sh --check` all green (no `_prog` added, so `PROGRESS.md` does not move). `check-unimported`'s one orphan is the pre-existing untracked `UnifiedDecodeItemLongBytesValidated`, not from this branch.

## Next (stage 3, separate PR)

The composition: three machine paths (raw byte / short form / long form), the combined triple, the `encodeBytes data` corollary, path-coverage examples that witness **output bytes as literals** (per the #11040 review) including both sides of 55/56, plus `RoutineEntry` + `Correspondence` registry rows — on #11273's tier reading, `rlp_encode_bytes` earns `.proven`, since it is total with no input-domain restriction.


pirapira:
Reviewed at `4df85a6da`. The coverage claim holds and the framing is right; two documentation findings, no soundness objection.

## Verified

- **The coverage map is contiguous and sums to exactly 76**, no gaps and no overlaps: 5+3+5+2+4+7+4+22+4+7+1+7+5.
- **Theorem addresses agree with the map.** Dividing by 4: `rebRawTail` +20..+52 → [5]-[12], `rebShortCopyLoop` +76..+100 → [19]-[24], `rebLadder` region, `rebLongHeader` +208..+224 → [52]-[55], `rebLolLoop` +224..+252 → [56]-[62], `rebLongCopySetup` +252..+256 → [63], `rebLongCopyLoop` +256..+280 → [64]-[69].
- **Line cap respected:** 1139 / 943 / 490 — all under 1500, and the split follows the `WithdrawalDecodeClose` precedent.
- **No forbidden tactics.** The only `sorry`/`admit` matches in the diff are the two docstrings *asserting* their absence.
- ⭐ **"Categorical, not fractional" is the correct framing** and worth keeping in the stage-3 body too. Per-instruction coverage with no whole-routine claim is a genuinely different object from a routine spec, and saying so prevents the reading where 76/76 looks like done.

## Finding 1 — the module's own instruction map contradicts the theorems by one, on the long path

`RlpEncodeBytesSAsm.lean:36-50` states:

```
[52]-[54]  long header 0xb7 + bc
[55]-[62]  length-of-length loop
[63]-[70]  payload copy loop
```

The theorems say otherwise, and the addresses are unambiguous:

| | docstring | theorems |
|---|---|---|
| long header | [52]-[54] | **[52]-[55]** (`rebLongHeader`, +208..+224) |
| length-of-length loop | [55]-[62] | **[56]-[62]** (`rebLolLoop`, +224..+252) |
| copy setup | *absent* | **[63]** (`rebLongCopySetup`, +252..+256) |
| payload copy loop | [63]-[70] | **[64]-[70]** (`rebLongCopyLoop`) |

⚠️ So the docstring double-books **[55]** and has no row for the [63] setup instruction. The PR body's table is the correct one.

This is not a soundness problem — the theorems are what they are. It matters because **that docstring is the artifact a reader uses to audit the "every instruction covered" claim**, and an off-by-one there is indistinguishable from a genuine coverage hole. Since this is the PR that establishes per-instruction coverage, it is the right place to reconcile.

## Finding 2 — the axiom check names 16 theorems; the branch defines 19

The body says `#print axioms` was run on *"all 16 block theorems plus `rebLadder`"*. The branch defines **18** in `RlpEncodeBytesBlocksSAsm` plus `rebLadder`:

`rebLolBody` · `rebLolLoop` · `rebShortCopyBody` · `rebShortCopyLoop` · `rebLongCopyBody` · `rebLongCopyLoop` · `rebPrologueNe1` · `rebPrologueEq1` · `rebRawProbeSmall` · `rebRawProbeLarge` · `rebRawTail` · `rebDispatchShort` · `rebDispatchLong` · `rebShortHeader` · `rebShortTail` · `rebLongHeader` · `rebLongCopySetup` · `rebLongTail`

The three unaccounted names are the loop-body helpers (`rebLolBody`, `rebShortCopyBody`, `rebLongCopyBody`). They are presumably clean — they are used by checked theorems — but the axiom gate's whole value is **totality**, so please either extend the check to all 19 or state that the bodies are covered transitively by their loops.

## On the signed counter — no objection, and the reasoning is the right shape

Carrying `i < 8` instead of fighting for `i < 2^63` because `Int.bmod` is opaque to both `omega` and `bv_omega` is the correct move: it is the *true* constraint (the counter starts at `bc-1 ≤ 7`) and it doubles as the termination bound. Preferring the true, smaller hypothesis over the general one that the tactics cannot reach is worth copying.

Likewise **recording that the two copy loops' scratch registers differ and that it is load-bearing** — `x30` because `x28` still holds `bc` for [71]'s `bc + 1 + len` — is exactly the note that stops a later "unification" from silently breaking the long tail.

## Process note, not a blocker

Proof/registry work is currently parked behind corpus-wide FR = 0 by maintainer ruling (FR is 521 as of the full-corpus census today). This PR predates that and is not an agent lane, so I am flagging the tension rather than acting on it — whether it merges now is the maintainer's call, not mine.

pirapira:
@dhsorens please check ˆ

dhsorens:
Both findings fixed in `618adbfff`.

## Finding 1 — the map was wrong, the theorems were right

Reconciled to the theorems' addresses:

```
[52]-[55]  long header 0xb7 + bc, counter seeded bc - 1
[56]-[62]  length-of-length loop                7*bc + 1
[63]       long-copy setup (reload counter)
[64]-[70]  payload copy loop                   7*len + 1
```

The root cause is worth naming: the stage-1 docstring was written **from the plan**, before the blocks were proved, and when the proofs landed I folded `[55]` (`ADDI x29, x28, -1`) into the header block — the natural cut, since it consumes the header's `x28` — without going back to the map. Your point that the map is *the artifact a reader audits the coverage claim against* is exactly right, and it is now generated-from-the-theorems in spirit: each row matches a theorem's address range, and the PR body's table and this docstring now agree.

## Finding 2 — all 19, not 16

Re-ran `#print axioms` over the full set — the 16 I listed **plus** `rebLolBody`, `rebShortCopyBody`, `rebLongCopyBody`:

```
19/19: [propext, Classical.choice, Quot.sound] only
```

So the bodies are clean directly, not merely transitively through their loops. You are right that the gate's value is totality; the miscount came from writing "16 block theorems + rebLadder" from the coverage table (which has 13 rows, some covering two theorems) instead of from `grep -c theorem`. The commit message records the full 19.

## On the FR = 0 parking

Understood — flagging what is in flight so the call can be made with full information: stage 3 (the whole-routine composition, `reb_spec_within` + the `RLPItem` corollary + boundary coverage with literal bytes + registry rows) is **complete and green locally** on a branch stacked on this one. I will hold it unpushed-as-PR until this one's disposition is decided, rather than adding to the queue against the ruling. If the parking means both should wait, they keep — everything is committed and the handoff plan makes them resumable cold.

pirapira:
Looks good. Merging.

PR #11313, opened 2026-08-03
proof(#11289): witness the 7 Correspondence.lean routine specs, make it strict

### LINKED ISSUE #11289: proof: register the 7 routine specs Correspondence.lean names but nothing witnesses, then make it strict
Follow-up to #11042 (closed by #11278), per @pirapira's request for issues each closable on their own close condition. Siblings: the coverage pin and the withdrawal-decode decision, filed alongside.

## What is wrong

`scripts/gen-axiom-witnesses.py` scans three registries and prints this on **every** run:

```
note: EvmAsm/Progress/Correspondence.lean: 7 row(s) name a theorem that no
scanned registry witnesses, so those are NOT audited:
  rlpBytesEncodedSize_spec, rlpFieldToU256Be_spec_within, rlpFieldToU64_spec_within,
  rlpListEncodedSize_spec, rlpListNthItem_spec_within,
  rlp_encode_list_prefix_short_pinned_spec_within, rlp_list_count_items_spec_within
```

Those rows carry a `verdict`/`basis` in the correspondence registry — the repo makes a **spec-correspondence claim** about each — while the theorem behind the claim is outside the kernel-truth gate. Any of the seven could regress to `sorryAx` with CI green.

`Correspondence.lean` is deliberately non-strict in #11278 because the seven modules are not imported there and #11276 was in flight. **#11276 is no longer a blocker**, so this is now a self-contained change.

## ⭐ Register them, don't just add abbrevs

All seven are **whole-routine** specs, and all seven live under `EvmAsm/Codegen/Programs/`:

| theorem | module |
|---|---|
| `rlpBytesEncodedSize_spec` | `RlpBytesEncodedSizeSAsm.lean` |
| `rlpFieldToU256Be_spec_within` | `RlpFieldToU256BeWholeSAsm.lean` |
| `rlpFieldToU64_spec_within` | `RlpFieldToU64WholeSAsm.lean` |
| `rlpListEncodedSize_spec` | `RlpListEncodedSizeSAsm.lean` |
| `rlpListNthItem_spec_within` | `RlpListNthItemSAsm.lean` |
| `rlp_encode_list_prefix_short_pinned_spec_within` | `RlpSpliceHelperSpec.lean` (already imported by `Routines.lean`) |
| `rlp_list_count_items_spec_within` | `RlpListCountItemsSAsm.lean` |

⇒ So the right move is **`RoutineEntry` rows in `EvmAsm/Progress/Routines.lean`**, not bare abbrevs bolted onto `Correspondence.lean`. That witnesses them *and* records tier/gate/notes, and it needs 6 new imports in the module that already carries the L1 layering exemption rather than 6 in one that does not. Rows are witnesses not routines, so per-form splits are fine.

⚠️ Each row needs its **tier read honestly**, per `Routines.lean`'s header: resource/ABI preconditions (alignment, `isValidByteAccess`, non-overflow) are *not* `.conditional`; a nonvacuous input-domain gate is — and `rlp_encode_list_prefix_short_pinned_spec_within` is a per-form ("short") theorem, so it will be one.

## Close condition

1. All 7 appear as `#print axioms` lines in `EvmAsm/Progress/AxiomWitnesses.lean` (witnesses **111 → 118**).
2. `Correspondence.lean`'s `SOURCES` entry in `gen-axiom-witnesses.py` flips to `strict=True`, so a future unwitnessed row is an **error** rather than a note.
3. `python3 scripts/gen-axiom-witnesses.py` prints **no note** and exits 0; `scripts/check-axioms.sh` passes.

Item 2 is the part that stops this recurring — with it, the class of defect cannot come back silently in that file.

## Verify it bites

Point one new abbrev at a `sorry`-carrying theorem and confirm `check-axioms.sh` fails with `sorryAx`; then delete an abbrev while leaving its row and confirm the now-strict cross-check errors. Without both, the change only shows the gate *runs*.

dhsorens:
Closes #11289.

`Correspondence.lean` named 7 whole-routine RLP specs in its `spec := some "…"` rows — the repo makes a spec-correspondence claim about each — while no registry witnessed the theorem, so all 7 were outside the kernel-truth gate. Any could regress to `sorryAx` with CI green. The generator printed a note about it on every run, which is the same failure mode #11042 was filed about: a green check quietly covering less than a reader assumes.

## What landed

**Registered, not bolted on.** All 7 become `RoutineEntry` rows in `Progress/Routines.lean` (6 new imports; `RlpSpliceHelperSpec` was already imported), so they get tier/gate/notes rather than bare abbrevs in `Correspondence.lean`. Witnesses **111 → 118**. All 7 are whole-routine triples at their linked guest addresses (`B := GuestAddrs.<symbol>`), confirmed against the correspondence `spec` refs.

**Tiers read honestly** per `Routines.lean`'s header — resource/ABI hyps do not make a row conditional:

| routine | tier | why |
|---|---|---|
| `rlp_bytes_encoded_size` | `.proven` | ptr/len consistency + alignment + validity only |
| `rlp_field_to_u256_be` | `.proven` | 32-byte output, list-slack, register-encoding — all ABI |
| `rlp_field_to_u64` | `.proven` | u64-width companion of the above |
| `rlp_list_encoded_size` | `.proven` | **total** — result covers both the `ult v 56` short branch and the long branch; only hyp is `halignRet` |
| `rlp_list_nth_item` | `.proven` | buffer-slack + alignment + u64-representability |
| `rlp_list_count_items` | `.proven` | same shape |
| `rlp_encode_list_prefix` | `.conditional` | `len.toNat < 56` — the RLP short-form bound, same `lenlen ≥ 2` cut as `rlp_item_size` (#10780 item 3) |

⭐ **The recurrence fix (issue item 2).** `Correspondence.lean`'s `SOURCES` entry flips to `strict=True`. All three registries are now strict and the cross-check runs against their **union**, so a future row naming a theorem no registry witnesses is a **build error**, not a printed note. Non-strictness was only ever a bridge around #11276 being in flight; that's merged, so the bridge is gone.

## Verification

- Gate: **OK at 118 witnesses**, classical axioms only. The 7 new witnesses each print `[propext, Classical.choice, Quot.sound]` (or a subset) — zero non-classical across all 115 with-axiom reports.
- The gate's independent witness grep and the generator produce **identical** 118-name sets.
- `gen-axiom-witnesses.py` prints **no note** and exits 0.
- `lake build` clean (4800 jobs); `check-{axioms,layering,forbidden-tactics,unimported,no-warnings,file-size,progress,naming}.sh` all pass.
- ⭐ **Both negative controls bite** (issue's "verify it bites"):
  - a `sorry`-carrying witness → `check-axioms.sh` fails with `FORBIDDEN (sorry): … sorryAx`;
  - deleting `rlpListNthItem_spec_within`'s abbrev while leaving its rows → the now-strict cross-check **errors** naming it (from both the Routines row and the Correspondence row), where before this PR the Correspondence side would have been a silent note.

### DIFF
--- a/EvmAsm/Progress/AxiomWitnesses.lean
+++ b/EvmAsm/Progress/AxiomWitnesses.lean
@@ -23,17 +23,34 @@
   rather than whole-routine, so populating every one is not a single change.
   What is registered is witnessed; what is not registered is still outside the
   gate.  Stated here rather than left implicit, because the assumption this
-  file exists to break is "covered by CI".
+  file exists to break is "covered by CI".  #11290 tracks turning that residual
+  into a measured coverage number rather than prose.  As of #11289 all three
+  registries are strict, so within a *registered* file an unwitnessed row is a
+  build error — the gap is only in what has not been registered yet.
 -/
 import EvmAsm.Progress
 import EvmAsm.Progress.Routines
 import EvmAsm.Progress.Correspondence
+
+#print axioms EvmAsm.Codegen.RlpBytesEncodedSizeSAsm.rlpBytesEncodedSize_spec
 
 #print axioms EvmAsm.Codegen.RlpEncodeUintBeSAsm.reub_spec_encode_within
 
 #print axioms EvmAsm.Codegen.RlpEncodeUintBeSAsm.reub_spec_within
 
 #print axioms EvmAsm.Codegen.RlpEncodeUintBeSAsm.reub_spec_within_of_length_le
+
+#print axioms EvmAsm.Codegen.RlpFieldToU256BeSAsm.rlpFieldToU256Be_spec_within
+
+#print axioms EvmAsm.Codegen.RlpFieldToU64SAsm.rlpFieldToU64_spec_within
+
+#print axioms EvmAsm.Codegen.RlpListCountItemsSAsm.rlp_list_count_items_spec_within
+
+#print axioms EvmAsm.Codegen.RlpListEncodedSizeSAsm.rlpListEncodedSize_spec
+
+#print axioms EvmAsm.Codegen.RlpListNthItemSAsm.rlpListNthItem_spec_within
+
+#print axioms EvmAsm.Codegen.RlpSpliceHelperSpec.rlp_encode_list_prefix_short_pinned_spec_within
 
 #print axioms EvmAsm.Codegen.RlpSpliceHelperSpec.rlp_item_size_spec_within
 
--- a/EvmAsm/Progress/Routines.lean
+++ b/EvmAsm/Progress/Routines.lean
@@ -58,6 +58,12 @@ import EvmAsm.Progress
 import EvmAsm.Evm64.AccountAccessorSpec
 import EvmAsm.Codegen.Programs.RlpEncodeUintBeComposeSAsm
 import EvmAsm.Codegen.Programs.RlpSpliceHelperSpec
+import EvmAsm.Codegen.Programs.RlpBytesEncodedSizeSAsm
+import EvmAsm.Codegen.Programs.RlpFieldToU256BeWholeSAsm
+import EvmAsm.Codegen.Programs.RlpFieldToU64WholeSAsm
+import EvmAsm.Codegen.Programs.RlpListEncodedSizeSAsm
+import EvmAsm.Codegen.Programs.RlpListNthItemSAsm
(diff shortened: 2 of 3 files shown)


pirapira:
Looks good. Merging.

pirapira:
Looks good. Merging.

PR #11314, opened 2026-08-03
proof(#10780): the whole-routine triple for rlp_encode_bytes

dhsorens:
Item 2 of #10780, **stage 3 of 3 — closing the item**. Stages 1 (#11275) and 2 (#11300) are merged; before this PR every one of the 76 instructions was block-covered and no theorem said the routine computes RLP. This closes that categorical gap, on **both sides of the 55/56 boundary**.

## `reb_spec_within` — total, with no input-domain restriction

```lean
cpsTripleWithin (19 + 10*bc + 7*n) rebBase (ra &&& ~~~1)
  (rebAbiPre …) (rebAbiPost …)
```

Post: status `a0 = 0` (total function), output region = `encodeBytes data ++ outBytes.drop (enc.length)` — the encoding at the front, nothing else moved — and the `a3` cell holds the encoding's length. Unlike `rlp_encode_uint_be`'s `≤ 55` bound there is **no domain restriction**: `n < 56` and `n ≥ 56` are both inside one claim, which is what "pin both sides" demanded. Preconditions are the ABI's resource ones only (capacity `n + 9`, alignment, windows in range, cell owned).

Three machine paths, split on the data before touching the machine. The two multi-route paths prove their shared tails **once**, universally quantified over the registers the routes disagree about — the long path has three such seams (`x29` after the ladder, `x30` after each loop), joined by `cpsTripleWithin_of_forall_regIs_to_regOwn`.

## `reb_spec_rlpItem_within` — the SpecRef-facing form

`encodeBytes` is not "a port that may drift from SpecRef": SpecRef **defines** `encR := EL.RLP.encode`, and `encode (.bytes d) = encodeBytes d` holds definitionally — so this corollary is a *direct application*, no bridge lemma. `rlpItemRegionFrom` states the output region over `RLPItem`, the vocabulary a caller encoding a SpecRef struct field composes with. Deliberately **not** over a `SpecRef.Account` or similar: the routine encodes whatever bytes it is handed, and a predicate claiming the region "represents" a structure it never saw would be false.

## Coverage witnesses content, not cost — at the boundary

Four examples with **output bytes and written length as literals** in the post:

```
n=1  raw     → [0x2a],                          1 written
n=1  large   → [0x81, 0x81],                    2 written
n=55 short   → 0xb7 :: replicate 55 0x11,       56 written
n=56 long    → 0xb8 :: 0x38 :: replicate 56 …,  58 written
```

`0xb7` is both the largest short header and the long-form base, so an off-by-one at the boundary produces a *well-formed header of the other kind* — the 55/56 pair is the check that neither side leaks into the other. **Negative control run:** swapping the `n=56` example's `0xb8` for `0xb7` fails to elaborate.

Two proof-engineering notes recorded in §9: definitional reduction through 55-element lists exhausts the elaborator's fuel (raising it is off-limits), so the boundary examples convert the post via a kernel-`decide`d equation; and `decide` cannot reduce `Nat.toBytesBE` (well-founded recursion), so the long form's literal goes through the equation lemmas.

## Both registries moved — and each names why the other exists

- **`Progress/Routines.lean`**: two rows at **`.proven`** — total function, no input-domain restriction, unlike `reub`'s `.conditional` — plus witness abbrevs (rows without abbrevs are invisible to the gate). Counts 11→13 rows, proven 4→6, symbols 6→7.
- **`Progress/Correspondence.lean`**: `rlp_encode_bytes` `.unproven/.none/"drift guard only"` → **`.agrees/.bridged`** with the spec named; the note records that witnessing lives in `Routines.lean` because L1 layering forbids this file importing Codegen. Verdict counts agrees 11→12, unproven 4→3; basis bridged 6→7, none 4→3.
- `PROGRESS.md` regenerated (it renders both tables).

`RlpRead.lean`'s docstring now names all four spec modules, so a per-file theorem count of the emitting module does not read as "unspecified" (the #10779/#10782 mistake).

## Verification

- `lake build` clean, zero warnings, after absorbing today's guest relink (the routine's own address is unchanged at `0x80004ac0`).
- `#print axioms` on all five spec theorems (`reb_spec_within`, `reb_spec_rlpItem_within`, and the three path theorems): classical-3 only. Stage 2's 19 block theorems likewise (verified in the #11300 thread).
- `check-forbidden-tactics` / `check-file-size` / `check-layering` / `check-heartbeats-approved` / `check-guest-addrs` / `progress-report.sh --check`: all green.

## On the FR = 0 parking

Held as promised on #11300 until that PR's disposition was decided; its merge decides it, so this follows as the completion of the same item rather than a new lane. If the parking means this should wait regardless, it keeps — everything here is stable against `main`.

Next in #10780's order: item 3 (`rlp_item_size` / `rlp_encode_list_prefix` `lenlen ≥ 2` long forms) — **not started**, pending the FR ruling.


pirapira:
Looks good. Merging.

pirapira:
Parking this as `complicated-merge`: the merge with `main` has a conflict I don't think I should resolve unilaterally.

**Conflict:** `EvmAsm/Progress/Routines.lean`, two hunks, both in the registry *census* theorems. The `routineRegistry` list itself merges cleanly — only the counts collide:

| theorem | this PR | `main` (after #11313) |
|---|---|---|
| `routineCount_eq` | 13 | 18 |
| `routineProvenCount_eq` | 6 | 10 |
| `routineConditionalCount_eq` | 7 | 8 |
| `routineSymbols_eq` | 7 | 13 |

**Why I stopped:** resolving means writing new numeric literals into four theorem *statements*. My standing instruction is that statements and definitions are frozen — I may repair proof bodies mechanically, but not edit what a theorem claims (the only exception being addresses that move due to ELF layout). These counts are a semantic claim about proof coverage, not a layout pin, so the correct post-merge values are yours to assert.

To be clear about the shape of it, this is not hard to fix and it is not a defect in the PR — the numbers are fully determined by the merged registry, and `decide` will reject any wrong guess kernel-side. It's simply that picking them is an authoring call rather than conflict repair.

Everything else is clean: merging `main` produces **no other conflict** in the PR (`behind` was 3 commits).

Once the four counts are updated, remove `complicated-merge` and I'll pick it up on the next pass — full local verification (`lake build` + all 16 gates) before it goes to CI.

dhsorens:
Our messages crossed — the conflict is already resolved on the branch at `a375d1673`, pushed a few minutes before your comment.

**Resolution:** the union census, since the registry lists themselves merged cleanly (my 2 rows + #11313's 7, over the base 11):

| theorem | this PR | `main` (#11313) | **merged** |
|---|---|---|---|
| `routineCount_eq` | 13 | 18 | **20** |
| `routineProvenCount_eq` | 6 | 10 | **12** |
| `routineConditionalCount_eq` | 7 | 8 | **8** |
| `routineSymbols_eq` | 7 | 13 | **14** |

Your instinct not to write numeric literals into theorem statements unilaterally is the right default — worth saying why it is safe *here*, and only here: these four are the census theorems whose entire design is that **`by decide` recomputes the truth from the merged list**. I did not trust my arithmetic either; I wrote the union values and let `decide` be the check — a wrong count fails elaboration rather than landing. (First attempt was exactly that: I computed the numbers by hand, and the build confirmed them rather than me.) The pinned literals exist precisely to force this conflict into view instead of letting two registry edits slide past each other, and it worked as designed.

Full `lake build` green on the merged branch, zero warnings; `progress-report.sh --check` clean (`PROGRESS.md` needed no regeneration — main's copy already renders the union). `merge!` is still on; the branch is up to date with `main` as of `6ddc4d72c`, so the queue should take it now. Happy to have the `complicated-merge` label lifted whenever you've glanced at the resolution.

pirapira:
Unparked — thanks, the counts resolve it. Verified locally on your merge (`a375d1673`) plus `main`: `lake build` green (4803 jobs), all 16 gates `ok`, and the census theorems elaborate, so 20 / 12 / 8 / 14 are the right values.

One thing I had to fix on top, worth flagging since it would have failed CI:

```
check-axioms: FAIL: EvmAsm/Progress/AxiomWitnesses.lean is stale vs
  EvmAsm/Progress.lean / EvmAsm/Progress/Routines.lean / EvmAsm/Progress/Correspondence.lean
```

The new `routineRegistry` rows need `AxiomWitnesses.lean` regenerated so their `proofRef` theorems actually reach `#print axioms`. Without it the rows are registered but unwitnessed by the axiom gate — the exact hole the generator's cross-check exists to close. Regenerated mechanically (`python3 scripts/gen-axiom-witnesses.py --write`, now 120 witnesses, +4 lines) and pushed; re-ran `check-axioms.sh` clean, all classical axioms only.

Auto-merge is armed.

dhsorens:
Thanks for the unpark, the verification, and for pushing the `AxiomWitnesses.lean` regen (`46440ee16`) — pulled.

That staleness is the **third** generated artifact I've now been caught missing on this item (`PROGRESS.md` on #11276, `GuestAddrs.lean` earlier, `AxiomWitnesses.lean` here), and it's the worst of the three: the other two fail loudly in CI's reports gate, while a registered-but-unwitnessed row is exactly the silent #11042 hole the witness machinery exists to close — your strictness change in #11313 is what turned it into a CI failure instead. `gen-axiom-witnesses.py --write` is now on my pre-push list alongside the other two regens, keyed to 'any registry gained rows'.

pirapira:
Re-parking as `complicated-merge`: the same census conflict came back after #11318 landed.

```
<<<<<<< HEAD  (this PR)
theorem routineCount_eq       : routineCount = 20
theorem routineProvenCount_eq : routineCountTier .proven = 12
=======       (main, after #11318)
theorem routineCount_eq       : routineCount = 19
theorem routineProvenCount_eq : routineCountTier .proven = 11
>>>>>>>
```

Same reason as before — resolving means writing new literals into theorem *statements*, which is yours to assert, not conflict repair I should do. `EvmAsm/Progress/Routines.lean` is again the only conflicted file; nothing else in the merge conflicts.

**Worth flagging the pattern, though:** this is now the second time, and it will recur on *every* merge into this PR for as long as it's open, because every PR that adds a registry row bumps the same four literals. #11313 caused it, #11318 caused it, and the next one will too. The PR can't stay current without a re-resolve per landing.

Two ways out, your call:

1. **Land it soon** — it's green apart from this, so the window just needs to be short.
2. **Stop pinning the counts as literals.** The values are fully determined by `routineRegistry`, so a formulation that doesn't restate them (e.g. asserting the tier partition covers the registry, rather than that it has exactly *N* rows) would make these merges conflict-free permanently. That's a statement change, so it's a decision for you — I mention it only because the conflict is structural rather than incidental.

For reference, verification was clean on the previous resolution (`lake build` 4803 jobs, all 17 gates `ok`), and I regenerated `AxiomWitnesses.lean` there — the new rows need it or `check-axioms` fails. Re-apply that after resolving.

pirapira:
The merge bot should handle this kind of conflicts.

pirapira:
Unparked and resolved — counts updated rather than left for you, per instruction that these literals are bookkeeping (like `PLAN.md`), not statements being weakened to pass.

Derived the values from the merged `routineRegistry` (21 rows: 13 `.proven`, 8 `.conditional`, 15 distinct symbols) rather than picking either side of the conflict, so:

| theorem | this PR | main | merged |
|---|---|---|---|
| `routineCount_eq` | 20 | 19 | **21** |
| `routineProvenCount_eq` | 12 | 11 | **13** |
| `routineConditionalCount_eq` | 8 | 8 | 8 |
| `routineSymbols_eq` | 14 | 14 | **15** |

Note neither side's number was right — both sides added rows, so the merged total is higher than either. `decide` confirms all four kernel-side, which is what makes this safe: a wrong literal cannot elaborate.

Also regenerated `AxiomWitnesses.lean` (121 witnesses) for the new rows — without it `check-axioms` fails as stale.

Verified locally: `lake build` green (4803 jobs), all 17 gates `ok`. Auto-merge armed.

The structural point from my previous comment still stands if you want these merges to stop conflicting, but it's no longer blocking anything.

Issue #11290, opened 2026-08-03
proof: the guest-routine axiom gate cannot report its own coverage — 11 registered rows against an unmeasured surface


dhsorens:
Follow-up to #11042 (closed by #11278), per @pirapira's request for issues each closable on their own close condition. Sibling of #11289.

## The gap

#11042 asked to populate the routine registry **"enumerate rather than sample, since sampling is what let this go unnoticed."** #11278 sampled. Measured on `5ceb46cd1`:

| | |
|---|---|
| `RoutineEntry` rows | **11** |
| witnesses in the gate | **111** |
| distinct `*_spec_within` names tree-wide | **1835** |

1835 is an upper bound, not a target — most are block-level, not whole-routine. **That is exactly the problem:** nobody can currently say what the denominator *is*, so "how much of the verified guest surface is audited?" has no answer, and the honest residual is invisible to anyone who did not read #11278's PR body.

⇒ This is the same genre as #10959 (`check-build-parallel` cannot report its own coverage; 4 of 9 scripts exit 0 on a missing toolchain and only 1 prints a set size). Same failure mode, different gate: **a green check that quietly covers less than a reader assumes.** Filing separately because the fix is in a different place, but they should be read together.

## Why a count, not an enumeration drive

Enumerating "the verified guest surface" is not closable — "whole-routine" is not machine-decidable, so an issue asking for it has no honest done-when, which is why this one asks for a **number** instead. Once the number exists, further registration work can be filed against a moving figure rather than a vibe.

## Suggested shape

A kernel-checked pin in `Routines.lean`, mirroring `Progress.lean`'s `provenBytes_eq` convention:

* `routineCount` is already there (`= 11`, `by decide`).
* Add a **measured denominator** the pin is stated against — a machine-derivable proxy for the whole-routine surface. Candidate: the count of `*_spec_within` theorems whose statement is `cpsTripleWithin _ <base> _ <code>` at a **linked guest address** (`GuestAddrs.*`), which is the property that distinguishes a whole-routine triple from a block lemma and is decidable by a source scan.
* A script leg that recomputes the denominator and fails on drift, so the ratio moves visibly when either side changes — the `gen-axiom-witnesses.py` diff-and-fail pattern already in use.

[...]

PR #11318, opened 2026-08-03
proof(#11291): register the withdrawal_decode whole-routine triple

### LINKED ISSUE #10782: Proof: compose the 4 block lemmas into a whole-routine spec for WithdrawalDecodeClose (blocks proven, composition missing)

### LINKED ISSUE #11291: proof: the withdrawal-decode surface has no single whole-routine claim to register — pick one or record why not
Follow-up to #11042 (closed by #11278), per @pirapira's request for issues each closable on their own close condition. Siblings: #11289, #11290.

## Why it was left out of #11278

#11042 named `withdrawal_decode` among the routines to register. I deliberately did **not** register it, and said so rather than quietly skipping it — this issue is that omission written down.

Every other routine in #11278 had an unambiguous strongest claim: `reub_spec_within`, `rlp_item_size_spec_within`, the walk-chain triples. The withdrawal-decode surface does not. It is **WP-certificate shaped** — `WithdrawalDecode{,AutoWP,ChainWP,FailureWP,SemanticWP,ShortWP}.lean` and `WithdrawalSchemaWP.lean` carry certificates, `_pre`/`_exits` pairs and per-path facades rather than one triple, e.g.:

* `walkInitShortSuccessFromPrologueCert_pre`
* `successFieldSpecsReturnAbiCertOfInput_pre`
* `walkInitZeroNonzeroAbiFailureReasonErasedFromPrologueResolvedCodeSuccessFrameNBranch_{pre,exits}`

⇒ Naming any one of those as **the** witness for the symbol `withdrawal_decode` would overstate what it proves, which is precisely the failure mode the registry exists to prevent (see `Routines.lean`'s "rows are witnesses, not routines"). ⚠️ Note #10782 was closed as already-done on the strength of `withdrawal_decode_spec_within` (`WithdrawalDecodeClose5.lean:1157`) — so if that theorem *is* the whole-routine claim, this issue is cheap; if it is one path among several, that is worth knowing explicitly, because #10782's closure leaned on it.

## The judgement this needs

Which is it:

1. **A single whole-routine triple exists** (candidate: `withdrawal_decode_spec_within`) ⇒ register it, note the domain gate, done.
2. **Coverage is genuinely per-path**, no single claim ⇒ register the paths as separate rows (the registry allows this: `symbol` groups, it does not key) and state in the notes what the union does and does **not** amount to.
3. **Neither** ⇒ record a structural finding naming what would have to be proved to get a whole-routine claim, in the same shape as `RegionPredicates.lean`'s `callee_seed_table` finding.

This is a maintainer call about your module, not something to guess at — which is why it is an issue rather than a PR.

## Close condition

Either:
* `withdrawal_decode` (and the `Close2..5` chain as applicable) has `RoutineEntry` rows in `EvmAsm/Progress/Routines.lean` with witness abbrevs, so it is in `AxiomWitnesses.lean` and audited by `check-axioms.sh`; **or**
* `Routines.lean` carries a recorded finding naming why no whole-routine claim can be registered and what would have to change first.

Either outcome closes it. What does not close it is the current state: named in #11042, silently absent from the registry.

dhsorens:
Closes #11291.

## The decision resolved to option 1 — a single triple exists

#11291 asked whether the withdrawal-decode surface has one whole-routine claim or is genuinely per-path, and framed it as a maintainer call. Investigating, it's determined by the code, not preference: **`withdrawal_decode_spec_within` is a genuine whole-routine triple** — `wdPrologue ;; wdBBField0`, the full program at `GuestAddrs.withdrawal_decode`, decoding all four fields and returning `a0 = 0` with a `Decoded` verdict or `a0 = 1` with a witnessed `DecodeFailure`. Both paths in one triple, only ABI/resource hypotheses (stack layout, alignment, buffer-slack, `isValidByteAccess`, `oldAddr.length = 20`), so **`.proven`** and total.

⚠️ **My filing under-checked, and I'd rather say so than paper over it.** I looked at the intermediate `WithdrawalDecode*WP.lean` certificates and concluded no single claim existed — missing that `WithdrawalDecodeClose5` already composes them into the whole-routine triple. That triple has existed since 2026-07-17; it is what closed #10782. It was simply never registered, so it sat outside the axiom gate. So #11291 is issue **option 1** ("register it, done"), and the one row subsumes the per-path certificates per "rows are witnesses, not routines". Witnesses **118 → 119**.

## A gate leak this import surfaced (and fixes)

Importing `WithdrawalDecodeClose5` into the witness module broke `check-axioms.sh` — "parsed 172 axiom report(s) for 119 registry witness(es)".

⭐ **Root cause:** `#print axioms` reports ride lake's message channel and are **replayed for every module in the build graph**, not just the one built. `WithdrawalDecodeClose5` carries its own 54 `#print axioms` lines (for `wdBBField0`, `wd_disjoint`, …, its in-module verification), which leaked into the gate's parse. Every prior witness module happened to carry none, so the exact-count check never saw this.

**Fix:** the gate now scopes report parsing to the witness module's **own file** (`info: EvmAsm/Progress/AxiomWitnesses.lean:…` lines and their wrapped continuations), before stripping the position prefix. This loses no coverage — `#print axioms` is transitive, so a bad axiom in any dependency already appears in the witness's own report. It also hardens the gate against the next witness whose module self-prints, rather than being a one-off patch.

## Verification

- Gate **OK at 119 witnesses**, classical axioms only; the leaked 54 wd* reports are correctly excluded and `withdrawal_decode_spec_within` is included.
- Independent gate-grep and generator derivations produce **identical** 119-name sets.
- ⭐ **Negative control re-run after the scoping change**: a `sorry`-carrying witness still fails with `FORBIDDEN (sorry): … sorryAx` — the scoped filter still bites, including on sorries reached transitively.
- `lake build` clean (4802 jobs); `check-{axioms,layering,forbidden-tactics,unimported,no-warnings,file-size,progress,naming}.sh` all pass.

With this, the three #11042 follow-ups are: #11289 (merged), this, and #11290 (awaiting your proxy choice — comment posted there).


pirapira:
Looks good. Merging.

Issue #11310, opened 2026-08-03
codegen: sequential EIP-7702 authorizations charge AUTH_BASE from current overlay


pirapira:
## Finding

On current `origin/main` `e832e3f4c`, rows `00006` and `00009` are a real sequential-authorisation divergence. Input SHAs: `90074ba18a02a5e3d5cf58cd63a5c80342da878b0802bd2a7b3417f421f545b9` and `74cd077fb46c0bc826c0393f57671d56f07b1ce883301f72b3e62ff86b5c459e`. The guest returns receipt status 0 while the pinned execution-specs commit `e5a8caf1b` returns status 1.

`eip7702_authority_asof` calls `account_state_auth_current` (`EvmAsm/Codegen/Programs/TxIntrinsicStateGas.lean:101-125`) before falling back to the header state. After the first NULL authorization clears the current overlay, the second lookup sees `delegated=0` and enters the AUTH_BASE arm. The emitted path charges `23 * 1530 = 35190` state gas at `TxIntrinsicStateGas.lean:250`; with `state_gas_left=0`, this causes the pre-dispatch OOG. The guest then exposes the downstream missing-target BAL row because `auth_halted=1` skips `account_read_record` at `BlockVerdictDispatchTx.lean:712-714`.

The spec uses the transaction-start account for this decision: `execution-specs/src/ethereum/forks/amsterdam/vm/eoa_delegation.py:265-281` calls `get_pre_state_account`, derives `delegated_before_tx` from transaction-start code, and does not let the first authorization’s current overlay clear make the second authorization incur AUTH_BASE. This is distinct from #11305 (pre-dispatch OOG misclassified as recipient-unresolvable) and #11078 (authorizations applied after auth-phase OOG); cross-link both.

## Safety constraint

A fix must read `delegated_before_tx` from transaction-start pre-state, not merely skip AUTH_BASE and not substitute header state. This moves the accept/reject boundary outward by removing a charge, so a random valid-corpus sample cannot establish FA=0. The mandatory 1,036 oracle-invalid sweep is required, with FA=0 stated explicitly. Adversarial shapes to guard against are a previously-delegated authority that is treated as net-new, or an authority whose earlier transaction/block overlay differs from its transaction-start code; either could become an unsound accept if the baseline is too permissive. The auth-halted missing-row symptom should disappear as a consequence; if it remains, that is a separate defect.

pirapira:
[...]

Issue #11305, opened 2026-08-03
codegen: pre-dispatch OOG misclassified as deferred recipient failure


pirapira:
## Mechanism

On current main, EEST fixture `00030_test_set_delegation_oog_charge_point...oog_charge_a` reaches the deferred recipient path with `bv_mtx_recipient_lookup_deferred=1` and `runtime_tx_prepare_prefix_status=2`. The emitted `BlockVerdictDispatchTx.lean:717-722` unconditionally returns `a0=8`, which `BlockVerdictMtxRuntime.lean:596-605` maps through `.Lbv_mtx_recipient_unresolvable_fail` to `bv_fail_code=47`.

The trace shows the missing discriminator: `runtime_tx_auth_phase_halted=1` is already set by the authorization callback, and status 2 is written at the preparation-halt path (`0x800358a4`, `runtime_tx_gas_done`). No recipient read occurs. The guest therefore turns an authorization-phase OOG into a missing-recipient rejection.

The pinned execution-specs (`e5a8caf1b8055e4d805c7fb169edfa710914b7da`) says otherwise: `tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_authorization_oog.py:143-169` requires the entire authorization phase to roll back atomically; the recipient is only loaded by top-frame dispatch, which the halt precedes. `src/ethereum/forks/prague/vm/interpreter.py:123-125` calls `set_delegation` before `process_message`.

## Evidence

- Base: `e832e3f4c` (parent ELF `57286fcd03e2f91b1131191a0e9e9ff417fade03551d713cf55f4c677d1c1d10`, runner `5b383a7702fb05dea16a2cc37dc313b82491fdf194f39c12306cb387a5a2b6b0`); trace timestamp 2026-08-03 15:41 UTC.
- 00030 input SHA: `7a458261461616d3241dd9f8f3f2e73a8bd35bbd41656a00a750e304bd9be1d7012500000001000000000000000c0000000400000008000000080000000000000000000000` (manifest selector); direct input SHA is recorded in `/tmp/scoper-10839-r200-713750d3e`.
- Parent result: `bv_fail_code=47`, `shadow_status=1`, no recipient touch; rebuilt/supplied are the code-47 mismatch.
- This issue is intentionally one witness only. The three intrinsic-gas code-47 rows (`24497`, `24498`, `24513`) are a different subfamily and are tracked by codex2.

## Proposed correction

[...]

PR #11316, opened 2026-08-03
fix(codegen): clear stale cd_value_be after CALLCODE (zero-addr credit)

pirapira:
## Summary

Fixes fc=1 on GASPRICE-debug **b19** (and b22/b23/b26/b27): guest credited **address 0** with 3 wei stolen from `0x5bcd…` because post-descend `record_message_value_transfer` on CALL read **stale** `cd_value_be` left by a prior CALLCODE balance gate, with `nse_callee_be` still zero.

Closes #11315. Seed instance **A3** of the scratch-lifetime class catalogued in **#11317** (alongside `account_builder_pre_account` #11296 and `nse_create_pre_bal` #11309).

## Mechanism (four measured steps)

| step | what |
|---|---|
| 1 | CALLCODE fills `cd_value_be=3` for balance gate only (mode 2 = self-transfer; no rmv) |
| 2 | Later CALL stack value=0 skips the nse_callee fill |
| 3 | Post-descend rmv still runs `a3=1` with `a1=nse_callee_be=0`, `a2=cd_value_be=3` |
| 4 | Creates `0x0` bal=3; debits caller by 3 → bv_fail=1 |

SPIKE (parent guest `57286fcd` / measure parent `15329e98`):
- write value@`0xbd42eb98` → 3 at PC **`0x80056950` = `h_CALLCODE+0x28f4`** (nm-resolved)
- CALL post-descend jal rmv **`0x80053e6c`** / ra **`0x80053e70`** = `h_CALL` (nm-resolved)

Spec e5a8caf1b: `system.py:540` CALLCODE `to=current_target`; `interpreter.py:385-391` move_ether only for **this** message when value≠0.

## Fix

Zero `cd_value_be` + `nse_callee_be` at message-call entry next to existing per-CALL resets (`cd_xfer_log_pending`, …) in `ChildFrameHandlers.lean` `callDescendFallThrough`.

### Reset order (required)

The zeroing runs at **handler entry**, **before** the value-bearing balance gate that fills `cd_value_be` for **this** opcode. CALLCODE still fills `cd_value_be` for its own gate after the reset; nothing downstream of entry relies on a *prior* opcode's `cd_value_be` surviving across the entry point. (rmv and related consumers are same-handler, after the gate fill when value≠0.)

### Scope vs family (#11317 B2)

**Scoped to the measured defect** (`cd_value_be` + `nse_callee_be`), not a blanket `cd_*` reset.

[...]

PR #11319, opened 2026-08-03
fix: classify pre-dispatch OOG before recipient lookup

pirapira:
Related to #11305 (pre-dispatch OOG misclassified as deferred recipient failure).

## Change

The prepare-only success return used to run before the staged top-frame regular-gas charge. An exact intrinsic-gas transaction could therefore exhaust at that charge and fall through as an unresolved recipient. The return is now reached only when `runtime_tx_prepare_only` is genuinely active, after the top-frame charge. Preparation-active OOG marks the authorization-phase halt and uses the OOG transaction boundary; ordinary body OOG keeps normal settlement/body-snapshot behavior.

The recipient read remains skipped for the preparation-halt statuses, while the genuine deferred-recipient path remains live: deferred status 2 -> `a0=8` -> `.Lbv_mtx_recipient_unresolvable_fail` -> `bv_fail=47`.

## Evidence

- Candidate ELF: SHA-256 `81cfd8303c67ed0d4058687dffd4463c87db7c9faee096c1b529196d726dab6e`, mtime `2026-08-03 17:30:57 +0200`.
- `lake build`: passed, 4800 jobs.
- Body-OOG persistence controls 00023 and 00024: parent baseline PASS and candidate PASS; both match the expected first 69 bytes, including `success=1` and both authorization effects.
- In-scope rows 00030, 24497, and 24513: expected prefixes match and `bv_fail=0`. 24498 is explicitly out of scope for #11305 and is attributed to #11310: `bv_fail=60`, shadow status 1, and a 27-byte rebuilt/supplied digest divergence.
- Oracle-invalid reject-1036 sweep: `OK=1036`, `FA=0`, `FR=0`.
- Seeded random-200 (seed `20260803`, manifest SHA `4fb8f2b48f0b5c03efb967f9bad4b1bc0a0cad7fd9474cbea39d010a55589e94`): parent `OK=195 FR=5 FA=0`; candidate `OK=199 FR=1 FA=0`; four forward `FR->OK` repairs, zero reverse `OK->FR` flips, and no FA. The sole candidate FR, `00156_test_selfdestructing_initcode_preserves_balance...success-precompil`, is also FR on the parent (`oracle=1`, parent guest=0, candidate guest=0), so it is FR-equal/non-regressive.
- On the same 1068-row diagnostic population, `bv_fail=47` is `4 -> 0`; the four parent rows are 00030, 24497, 24498, and 24513.
- The retained unresolved-recipient arm is demonstrated by the emitted routing above and the historical txcount-gate v1 KAT, which recorded a fresh-recipient tx17 rejection via `bv_fail=47` when that recipient was omitted from the invalid-block witness. The v2 self-send control keeps the recipient resolvable and candidate9 returns the expected invalid result via `bv_fail=1`.

## Gates

`check-no-warnings`, `check-heartbeats-approved`, `check-axioms` (with `LAKE_ARTIFACT_CACHE=false`), `check-forbidden-tactics`, `check-layering`, `check-unimported`, `check-no-hardcoded-guest-pc`, `check-file-size`, `check-opcode-structure`, and `git diff --check` pass.

### DIFF
--- a/EvmAsm/Codegen/Dispatch.lean
+++ b/EvmAsm/Codegen/Dispatch.lean
@@ -2486,20 +2486,40 @@ private def emitTopLevelMessageD0Preparation : String :=
   "  add x8, x8, x7\n" ++
   "  sd x8, 0(x11)\n" ++
   ".runtime_tx_create_state_done:\n" ++
-  -- 4. top-level-only preparation halt, then top-frame gas and context
-  "  la x11, runtime_tx_prepare_only; ld x9, 0(x11); beqz x9, .runtime_tx_prepare_prefix_continue\n" ++
-  "  la x11, runtime_tx_prepare_prefix_status; li x9, 2; sd x9, 0(x11)\n" ++
-  "  la x11, runtime_tx_prepare_only; sd x0, 0(x11); j runtime_dispatcher_prepare_only_return\n" ++
+  -- 4. top-level-only preparation halt, then top-frame gas and context.  The
+  -- prepare-only return is deliberately after the staged top-frame charge:
+  -- execution-gas exhaustion is a transaction OOG, not an unresolved
+  -- recipient.  A successful preparation prefix still records status 2 and
+  -- returns below, preserving the genuine missing-witness verdict arm.
   ".runtime_tx_prepare_prefix_continue:\n" ++
   ".runtime_tx_gas_done:\n" ++
   "  sd x6, 568(x20)\n" ++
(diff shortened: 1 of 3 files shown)


pirapira:
Superseded by #11320, which batches this with the other `merge!` PR of mine into one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11320 lands.

pirapira:
Landed via #11320 (batched with #11316 into one CI round).

PR #11322, opened 2026-08-03
fix(eip7702): AUTH_BASE uses delegated_before_tx not current overlay

### LINKED ISSUE #11310: codegen: sequential EIP-7702 authorizations charge AUTH_BASE from current overlay

pirapira:
## Summary

Fixes #11310: `eip7702_authority_asof` reported `is-delegated` from **tx-current** AccountState (`account_state_auth_current` = pending then durable). After a same-tx authorization cleared a delegation, a later auth for the same authority saw `delegated=0` and re-charged `AUTH_BASE` (23×1530 = 35190), OOG'ing auth prep (filed fc=53 on 00006/00009).

Spec e5a8caf1b `eoa_delegation.py:265-281` charges AUTH_BASE from `get_pre_state_account` / `delegated_before_tx` at **transaction start**.

### Fix

In `eip7702_authority_asof` (`TxIntrinsicStateGas.lean`):

- **a1 nonce** stays CURRENT (pending→durable) — auth validation needs post-prior-auth nonce.
- **a2 delegated** is now `delegated_before_tx`: **durable-only** AccountState bit8, else authenticated **header** code. Never pending.

Not header-only (earlier-tx durable can differ). Not “skip AUTH_BASE always”.

### 27-byte / mass-scope REFUTED (pre-fix)

Coord asked confirm/refute that ~126 fc=60 (+27B) share this producer before fixing.

| case | shape | same producer? |
|---|---|---|
| empty AccountChanges RLP | **exactly 27B** | length unit only |
| **00840** no_delegation fc=60 | rebuilt 361 / supplied 334 Δ=+27; **extra** CALL target empty row | **NO** — no 7702; OOG-before-access BAL row |
| **24498** fc=47 | rebuilt 1 / supplied 332 | **NO** — catastrophic rebuild |
| **00006/00009** fc=53 | rebuilt 337 / supplied 393 Δ=**−56** missing rows | **YES** — this fix |
| fc=60 delta hist | (27,124), (-8,1) | opposite of auth missing-row |

**codex2 24981** (AUTH_BASE balance delta witness): post-fix still FR; output root already matches expected, only succ@32=0; parent≡cand — **not moved by this change**. Pointer samples 24978/24979/24835/24836 already OK on parent.

⇒ Scope of this PR = **fc=53-style AUTH_BASE asof**, not ~163-row mass clear.

### Evidence

| check | parent `15329e98…` | cand `d3da93bd…` branch `fix/eip7702-auth-base-delegated-before-tx` @ `b78a20baa` |
|---|---|---|
| 00006 pre_tx_deleg | FR succ=0 | **OK** full 69B match |
| 00009 multi_auth intra_tx | FR succ=0 | **OK** full 69B match |
| reject **1036** | — | **FA=0 FR=0 OK=1036** |
| r200 fwd | FA0 FR3 | FA0 FR3 (same 3: b19, callcode-7702-oog, create_selfdestruct) |
| r200 rev | — | FA0 FR3 |
| 00840 fc=60 | FR | FR (expected; different producer) |
| 24981 pointer | FR succ-only | FR unchanged |
| check-region-map | — | green (textSizeBytes `0x62b80`, second pin pass clean) |

FA discipline: fix stops a charge → accept boundary moves outward. 1036 is the gate; FA=0 stated. Adversarial shapes (prev-delegated as net-new; earlier-tx durable ≠ header) named; corpus may not contain them.

### Files

- `EvmAsm/Codegen/Programs/TxIntrinsicStateGas.lean` — asof a2 source
- layout regen: GuestAddrs, RegionMapLinkPins, symbol-addresses.tsv


pirapira:
Superseded by #11325, which batches this with the other `merge!` PR of mine into one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11325 lands.

pirapira:
Landed via #11325. Verified: commit `b78a20baa` is an ancestor of `main` as of `82a291b7a`, so the fix is in — this PR is closed rather than dropped.

Issue #11321, opened 2026-08-03
EIP-2935/4788 absent system contracts are seeded into the canonical storage map


pirapira:
The MTx system-storage seed unconditionally upserts nonzero 2935/4788 descriptors into the canonical map before execution. On 00538 (4608 bytes, EIP-4788 absent-contract), the guest therefore emitted an extra account leaf 0x000f3df6d732807ef1319fb7b8bb8522d0beac02 with storage slot 0x0c -> 0x0c, while the execution-spec post-state has no such account. The reference pipeline reproduces the expected root; the guest had bsr_change_count=2 and bv_fail=1. The terminal descriptor pass already clears swd_4788_vlen/root_vlen after discovering the account is absent, but that cannot retract the earlier canonical-map row.

The fix gates the seed on account_at_header_state_root plus non-empty code for both 2935 and 4788, before append_modeled_system_storage_tuple_rows. This mirrors execution-specs at e5a8caf1b: fork.py:782-812 resolves the system contract code, :833-856 constructs the message, and :858-860 incorporates only resulting state; apply_body invokes the unchecked system transactions at :897-906. An absent/codeless system account executes no storage write.

Evidence on runner 5b383a77: 00538 flips to exact expected root; 00514 (absent 2935) also flips; 24981 (present 4788) remains unchanged. The current manifest has exactly two absent-contract witnesses, 00514 and 00538. Parent ELF 57286fcd03e2f91b1131191a0e9e9ff417fade03551d713cf55f4c677d1c1d10 and candidate ELF 9bd41dffc914b843279c4fd28626541a31a6ad44d4677bf7cc8155e1dd715adf.

The full scope and A/B evidence are recorded on the PR.

pirapira:
## ⛔ The defect has INVERTED under this issue — it is an omission now, not an extra row

Decoded the supplied BAL for 00538 with the pinned Amsterdam decoder on `803053608`: **6 rows, 219 bytes**.

| address | content |
|---|---|
| `00000961ef48…7002` | withdrawal, 4 reads |
| `000064d67850…8282` | 4 reads |
| `0000bbddc7ce…7251` | 4 reads |
| `0000bff46984…8282` | 4 reads |
| `0000f90827f1…2935` | one storage change |
| **`000f3df6d732807ef1319fb7b8bb8522d0beac02`** | **EIP-4788 beacon-root system address, empty `AccountChanges`** |

[...]

PR #11324, opened 2026-08-03
fix(codegen): skip absent EIP-2935/4788 storage seeds

pirapira:
## What changed

The MTx system-storage seed at `BlockVerdictMtxRuntime.lean:307-311` derived descriptors and upserted BOTH EIP-2935 and EIP-4788 rows before any existence/code gating. On 00538 (EIP-4788 absent contract, 4608 bytes), this inserted an extra account leaf `0x000f3df6d732807ef1319fb7b8bb8522d0beac02` with storage slot `0x0c -> 0x0c`; the execution-spec post-state has no such account. `BlockVerdictStateRoot.lean:345-348` later clears `swd_4788_vlen`/`swd_4788_root_vlen`, but cannot retract the canonical map row already inserted. The defect is therefore at the seed, not a retraction that cannot undo it.

The fix gates both seeds by resolving the account with `account_at_header_state_root` and comparing its code hash to `cd_empty_code_hash`; absent or codeless accounts have their seed lengths cleared before `append_modeled_system_storage_tuple_rows`. This mirrors pinned Amsterdam execution-specs: `fork.py:782-812` resolves the system contract code, `:833-856` builds the message with that code, and `:858-860` incorporates only the resulting transaction state. `apply_body` invokes the unchecked 2935/4788 system transactions at `:897-906`; an absent/codeless target executes no code and produces no storage write.

This is the ninth wrong-stage-not-wrong-value mechanism in three days: the later descriptor cleanup was a compensation for an earlier unconditional action, and the correct alignment is to remove the action at its producer.

Issue: #11321.

## Population and controls

The fresh `/var/tmp/fc668/manifest.tsv` has exactly two absent-contract witnesses in the full 26,104-row corpus: 00514 (2935) and 00538 (4788). Both are fixed. There is no third witness in this corpus, so this is a closed measured class rather than an extrapolated fixture count.

- 00538 (4608 bytes): parent rejected; candidate returns the expected root exactly (`succ 0 -> 1`).
- 00514 (absent 2935 control): `succ 0 -> 1`, exact expected root.
- 24981 (present 4788 control, separate pointer-cluster lane): unchanged `succ 0 -> 0`; the working present branch is undisturbed. 00515, 00539 and 00559 are exact candidate controls.

## A/B evidence

Runner: `scripts/spike/spike_run` sha256 `5b383a7702fb05dea16a2cc37dc313b82491fdf194f39c12306cb387a5a2b6b0`. Parent is commit `e832e3f4c` and its ELF sha256 is `57286fcd03e2f91b1131191a0e9e9ff417fade03551d713cf55f4c677d1c1d10` (build timestamp 2026-08-03 18:33 CEST). Candidate is commit `9099e6c28` and its ELF sha256 is `9bd41dffc914b843279c4fd28626541a31a6ad44d4677bf7cc8155e1dd715adf` (build timestamp 2026-08-03 18:31 CEST; the candidate predates the freshly rebuilt parent, and both provenance shas are stated explicitly).

- Oracle-invalid 1,036-row manifest sha256 `bcf819366e66c8e12bc501bc00d6c99f1c6cbfcf52915f3aeb91d067bae3a06f`: parent `FA=0 FR=0 OK=1036`; candidate `FA=0 FR=0 OK=1036`. The parent and candidate per-fixture TSVs are byte-identical, both sha256 `b6991da6ce4c8cda6e02ae7efe9ea3a560647a50ff19ce28672835821663bc91`.
- Random-200 manifest `/tmp/11163-random200.manifest.tsv` sha256 `4aa33721d4a05b0c8c52c94809bdb2c2c74adeff28aab328c6052ce13692800c`: parent and candidate both `FA=0 FR=1 OK=199`, forward flips `0`, reverse flips `0`. Their per-fixture TSVs are byte-identical, both sha256 `78cb6f38b35b4b16558bd8cbfd00776854a19107e34084682c9292e8864fc984`. Forward=0 is expected here: neither of the two absent-contract rows is in this random sample, so no forward flip is possible.

## Generated layout / checks

Four-step regen reached a fixed point (the second complete pass was a no-op). `.text=0x62788`, `.data=0x5370`, `.bss=0x1aa744a0`; `bnf_le_a=0xa3000ee0` is unchanged. `LAKE_ARTIFACT_CACHE=false lake build` completed 4800 jobs. `scripts/check-build-parallel.sh` passes after materializing that non-cache build; this includes check-asm-to-program, check-region-map, guarded handler bytes, GuestAddrs starts, reports, axioms, and arithmetic fuzz. No `merge!` label was applied by this branch.


pirapira:
Superseded by #11325, which batches this with the other `merge!` PR of mine into one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11325 lands.

Issue #11258, opened 2026-08-03
docs(codegen): bal_serializer_verify docstring says INERT but it is wired and binding (codes 60/61)


pirapira:
## What

`bal_serializer_verify`'s docstring in `EvmAsm/Codegen/Programs/BalSerializer.lean` (around
`:1125`) says **"INERT AS OF THIS COMMIT."** That has been false since GH #10680 wired it.

On main `db98c6603` it is called at `BlockVerdictReceiptsTail.lean:278`, its return is stored to
`bv_bal_shadow_status` (`:279`), and it is **bound into the verdict** at `:324-333` — a digest
mismatch rejects with `bv_fail_code = 60`, a rebuild failure with `61`, ACCEPT-paths only.

## Why it matters more than a typo

This is a **soundness-critical binding described as dead code.** It has now misled two readers in
one night:

- coord quoted "INERT" as live state while assessing what the verdict actually checks;
- the body of #11245 was written on it, asking an implementer to *wire* a routine that is
  already wired and binding. Corrected there:
  https://github.com/Verified-zkEVM/evm-asm/issues/11245#issuecomment-5162221972

A docstring claiming a routine is inert is the one comment a reader will *not* verify, because
"nothing calls this" reads as a closed question.

## Fix

Replace the INERT sentence with the call site and the binding contract — one sentence each:
called from `BlockVerdictReceiptsTail.lean:278`; return bound at `:324-333` as codes 60/61 on
ACCEPT paths only.

⚠️ **Do not raise a docs-only PR for this** — a full build is ~1h. Fold it into the next PR that
already touches `BalSerializer.lean`.

## Rule this instantiates

Design *intent* in a docstring is reliable; **status and quantitative prose is not** — a status
claim is a measurement with a date, and this one has none. Same class as the "244 MiB exec log"
comment (actual: 8 MiB).

PR #11332, opened 2026-08-03
ci(#11303): routine-liveness gate — a theorem's subject must be alive or annotated

### LINKED ISSUE #11303: No gate asserts a theorem's subject routine is still emitted: check-asm-to-program compares Program to FIXTURE, so both can die together
## The gap

**No gate asserts that a routine with a theorem about it is still emitted and reachable.**

`scripts/check-asm-to-program.sh` compares a Program's rendered string against a **saved asm fixture**
in `scripts/asm-fixtures/` (391 manifest rows). That is a drift guard between *two checked-in
artifacts*. It says nothing about the live guest image.

⇒ **A Program and its fixture can both go dead together and the gate stays green.** The theorem keeps
proving the two agree, and neither is in the image.

## Confirmed instance (measured on `origin/main` e832e3f4c)

`bal_canonical_sort_selftest` and `bal_rlp_encode_selftest` have, per
`BalSelftestsProbe.lean`'s own header, *"zero callers, no build unit, no script, no workflow. They have
never run."* Measured: `bal_selftests` has **0 `jal` call sites**.

And there is a theorem about one of them:

- `balCanonicalSortSelftestFunction_eq_prog` — `EvmAsm/Codegen/Programs/BalCanonicalSort.lean:463`

⚠️ Scope it honestly: `_eq_prog` is a **byte-identity** theorem (emitted string = Program), i.e. a drift
guard, **not** a soundness claim. So this is wasted-coverage, not unsoundness. The point is that nothing
would have told us.

## What is NOT broken — checked, so nobody re-derives it

The prompting worry was that the storage/BAL work left stale theorems on SLOAD and EXTCODE*. It did not:

- **`committedStorageIs` was removed cleanly.** 0 references anywhere under `EvmAsm/` on main. The
  SLOAD entry was reworded to the canonical block-storage view and no orphan predicate was left behind.
- **SLOAD's theorems are about live code.** The modelled persistent-log base `0xa0630000` still appears
  **34×** in `EvmAsm/Codegen` (14× in `EvmAsm/Evm64`), so `evm_sload_*_spec_within` constrains emitted text.
- **`bal_sort_accounts` is fully gone** (0 refs) and `Correspondence.lean:240` carries a comment
  recording that its rows *used to* be there — the retirement was done properly.
- ⭐ **`BalCanonicalSort.lean:550` shows the good pattern:**
  `#guard (balCanonicalSortFunctions.splitOn "bal_sort_storage_writes:").length == 1` — a guard that a
  retired routine **stays** retired. That is the shape a liveness gate should take.

## The inverse defect is live and worse

**#11258**: `bal_serializer_verify`'s docstring says INERT. Measured on main: **1 `jal` call site** — it
*is* wired. A docstring claiming inertness on live code is more dangerous than a theorem about dead
code, because it invites someone to delete or ignore a live path.

⇒ So the registry drifts in **both** directions, and neither direction has a detector.

## Proposed check (source-level, no ELF, seconds)

For every routine named by an `_eq_prog` / `_spec_within` theorem, assert it has **at least one call
site** in some emitted composition, or is explicitly annotated as a deliberate no-caller (probe,
selftest, pending-caller) with a reason string. Fail on an unannotated zero-caller routine.

⚠️ Two traps for whoever builds it:
1. **Prove it can fire** — point it at `bal_canonical_sort_selftest` and watch it fail. A gate nobody has
   seen fail is indistinguishable from one that cannot.
2. **A name is not a contract.** Count `jal` sites in the *emitted* string, not source mentions; a
   routine referenced only in a docstring or a `#guard` has zero call sites (that is exactly how
   `bal_sort_storage_writes` reads).

## Priority

⛔ **Parked.** Proof/registry work is behind corpus-wide FR = 0 by maintainer ruling (currently FR 521
of 26,104). Filed so it is not re-derived; not a lane.

dhsorens:
Closes #11303.

## The gap

`check-asm-to-program.sh` compares a Program's rendered string against a saved fixture — a drift guard between two **checked-in** artifacts. A Program, its fixture, and every theorem about them can all go dead together (nothing references the routine) and every gate stays green: the theorem keeps proving two dead artifacts agree.

## The gate

`scripts/check-routine-liveness.sh` (pure source scan, ~2 s, wired into the `source-checks` job). For every theorem-bearing routine — MANIFEST.tsv conversion entry symbols ∪ `routine "…"` rows in `Progress/Routines.lean`, **391 today** — require that it be **alive**:

- **referenced in emitted text**: `jal ra|x1|x5|t0,` / `call` / tail-`j` / `la` in hand-written asm strings, `.jal`/`.la` reloc rows in converted Programs, `jalOff`/`laHi`/`laLo GuestAddrs.…` concrete immediates — never prose, docstrings, or `#guard` label mentions (the issue's trap 2: a *name* is not a contract; only instructions count); **or**
- **present in the guest symbol census** (`symbol-addresses.tsv` linker facts).

Neither signal alone suffices: call-site scanning misses routines reached only through the linked ELF's internal calls, and the census misses probe-only compositions exercised without being linked into the guest. A symbol with **neither** signal is DEAD unless annotated in `scripts/routine-liveness-allow.txt` (`<symbol><TAB><reason>`).

**The allowlist expires**: an annotated symbol that gains a liveness signal fails as STALE, naming the entry to delete — an inertness claim cannot outlive its reason. That is also the fix direction for the inverse defect (#11258): inertness claims live where they are re-checked every run, not in docstrings, where they rot.

## Negative controls (the issue's trap 1)

- `--self-test` removes the canary's allowlist entry in memory and **asserts the gate fires**; it runs *before* the real check in CI, so every run re-proves the gate can fail. Verified locally: `OK — gate fires on secf_add_mod_n when its annotation is removed`.
- STALE control run locally: a fake-dead entry for live `rlp_encode_bytes` fails with `STALE … has 46 reference(s)`, then removed.

## What it found on main, day one (fixed here)

1. **Two registry rows with pseudo-symbols** — `routine "account_rlp_content_to_u64"` / `"account_rlp_content_to_u256_be"` (from #11313) violate the `RoutineEntry.symbol` field contract ("symbol as it appears in the linker facts"); the linker symbols are `rlp_content_to_u64` / `rlp_content_to_u256_be`, both in the census. Rows renamed; the `decide`-pinned census is unchanged (21 rows, 15 distinct symbols), so this gate doubles as a registry-symbol validator.
2. **`secf_add_mod_n`** — a probe-only PC placeholder, "unlinked from guest" by its own comment (`Secp256k1Field.lean:26`), zero references, no census row. Annotated with that reason; it is the self-test canary.

## Why the canary is not `bal_canonical_sort_selftest`

The issue's instance gained a probe caller in 4aa007ef8 (test(bal): execute the two self-tests that had never run, #10680) — post-dating the issue's measurement — so it is alive under these semantics and needs no annotation. The genuinely-dead exemplar on today's main is `secf_add_mod_n`, hence the canary switch; if it ever gains a real caller, the self-test fails with an instruction to pick a new canary.

## Verification

`lake build` green (4802 jobs, no warnings); `check-forbidden-tactics` / `check-file-size` / `check-layering` / `progress-report --check` all clean; `Routines.lean` compiles with the pinned census `decide`s unchanged.


pirapira:
Reviewed as the filer of #11303. **No findings — and it corrects my issue on a point I got wrong.**

## ⛔ First, my error, because it is the substantive one

#11303 named `bal_canonical_sort_selftest` as the confirmed dead exemplar. **It is alive.** Verified on
`origin/main`:

```
2 × jal ra, bal_canonical_sort_selftest    (BalSelftestsProbe.lean)
4aa007ef8  test(bal): execute the two self-tests that had never run (#10680)
```

⚠️ **And I had that evidence when I filed.** My own measurement recorded `jal sites=2` for that symbol; I
wrote "zero callers" anyway, on the strength of the file's docstring. That is exactly the *read emitted asm,
not source strings* rule this repo already carries — broken in the issue whose purpose is catching stale
liveness claims. The PR is right to switch the canary; `secf_add_mod_n` is the genuinely-dead exemplar
("unlinked from guest" per `Secp256k1Field.lean:26`, zero references, no census row).

## What makes the gate stronger than what #11303 asked for

⭐ **Two liveness signals, and neither is sufficient alone.** Instruction references in emitted text *or*
presence in the linker census — because call-site scanning misses routines reached only through the linked
ELF's internal calls, and the census misses probe-only compositions never linked into the guest. That
distinction is the substance of the #11229 blind-spot inventory, applied correctly.

⭐ **Instructions only, never prose.** `jal`/`call`/tail-`j`/`la`, reloc rows, `jalOff`/`laHi`/`laLo`
immediates — explicitly not docstrings or `#guard` label mentions. That is #11303's trap 2, and it is the
trap I personally fell into.

⭐⭐ **The allowlist EXPIRES**, which #11303 did not ask for and is the best idea in the PR. An annotated
symbol that *gains* a liveness signal fails as **STALE**, naming the entry to delete. So an inertness claim
cannot outlive its reason — and it relocates such claims from docstrings, where they rot silently, into a
file re-checked every run. That is also the right fix direction for the inverse defect #11258
(`bal_serializer_verify`'s docstring says INERT while it has a live call site).

⭐ **`--self-test` runs BEFORE the real check on every CI run**, removing the canary's annotation in memory
and asserting the gate fires. That satisfies #11303's trap 1 — *prove it can fire* — continuously rather
than once at authoring time. A gate nobody has seen fail is indistinguishable from one that cannot, and
this one re-proves itself every run.

## It found real defects on day one

1. **Two registry rows carried pseudo-symbols** — `account_rlp_content_to_u64` / `_u256_be` from #11313,
   violating `RoutineEntry.symbol`'s contract ("as it appears in the linker facts"), where the real symbols
   are `rlp_content_to_u64` / `rlp_content_to_u256_be`. So the gate doubles as a **registry-symbol
   validator**, which is a capability beyond its stated purpose.
2. `secf_add_mod_n` — annotated with its own source comment as the reason.

⚠️ Note #11313 merged only hours ago and introduced defect 1. A gate that catches a same-day regression on
its first run is the strongest possible argument for merging it.

## Scope and cost

+244/−2 across 5 files, pure source scan, ~2 s, wired into `source-checks` — no ELF, no build. Covers 391
symbols (MANIFEST conversion entries ∪ `routine "…"` rows). `lake build` green at 4802 jobs; the
`decide`-pinned census is unchanged at 21 rows / 15 distinct symbols.

⇒ **No objection.** The one thing I would have flagged — a stale exemplar — the PR found and corrected
itself.

pirapira:
Looks good. Merging.

Issue #11326, opened 2026-08-03
structural review: guest vs execution-specs e5a8caf1b — verified map of remaining structural differences (guest 31a3e7442)


pirapira:
# Structural review: guest vs execution-specs @ e5a8caf1b — verified map of remaining structural differences

**Guest state:** `origin/main` = `31a3e7442` (all guest citations via `git show origin/main:PATH`).
**Spec state:** execution-specs pin `e5a8caf1b8055e4d805c7fb169edfa710914b7da`, fork `src/ethereum/forks/amsterdam/` (cited directly, not via `SpecRef`).
**Method:** spec-structures-outward. Spec inventory: `BlockState` (state_tracker.py:57), `TransactionState` (state_tracker.py:84, adds `created_accounts` + `transient_storage`), `Evm.accounts_to_delete` (vm/__init__.py:184), BAL types (block_access_lists.py:34-320). Each of the 9 candidate issues + the maintainer's 2 named priorities was re-verified against `origin/main`, not trusted from the issue list.
**Context census (2026-08-03, base 3d395c1ba, n=26104, FA=0, FR=375 adj):** code1=238, code60=125, code47=4, code53/37/3=2 each, code44/56/31=1.

Entries ranked by how many FR families each plausibly explains.

---

## 1. Granular BAL-content validation: the guest parses the declared BAL; the spec only hashes it — STILL PRESENT (#10612, #11183)

**Spec:** the provided BAL enters as opaque bytes (`stateless_ssz.py:144`) and is consumed exactly once: `block_access_list_hash = keccak256(execution_payload.block_access_list)` (`execution_engine/validation_helpers.py:80-82`). Everything else operates on the **built** list: `fork.py:928-929` (`build_block_access_list`), `fork.py:933-936` + `block_access_lists.py:711` (gas-limit check on the built list), `fork.py:366-391` (hash compare vs header). `block_access_lists.py` contains **no** validate function over a *provided* BAL.

[...]

Issue #11295, opened 2026-08-03
Post-switch residual: 8 FRs across two cohorts all land on bv_fail=1 (terminal root) — needs per-row leaf differencing, NOT a shared-cause assumption


pirapira:
## Observation

After the #10651 Step 2 authority switch (#11277), two **independent** EEST cohorts on current main have
residuals dominated by the same verdict code — `bv_fail=1`, the **terminal state-root comparison**.

| cohort | rows | FR | codes | root |
|---|---|---|---|---|
| named `GASPRICE-debug` | 37 | 5 | **all 5 = `bv_fail=1`** (b19, b22, b23, b26, b27) | 37/37 |
| focused `eip7928_block_level_ac` | 60 | 5 | 00043/00044/00045 = **`bv_fail=1`**; 00047/00048 = `bv_fail=60` | 60/60 |

Artefacts: guest `57286fcd03e2f91b1131191a0e9e9ff417fade03551d713cf55f4c677d1c1d10`,
runner `5b383a7702fb`, main `e832e3f4c` (GASPRICE cohort); guest `e4909ee8` built 2026-08-03 11:37Z,
same runner, main `6d1647824` (eip7928 cohort). Random-30 seed 20260803 is 30/30 exact on the same guest.

## ⛔ WHY THIS IS **NOT** YET A CLASS — READ THIS BEFORE ACTING ON THE TABLE

`bv_fail=1` is the **terminal** comparison: it fires *after* every granular check has passed. It is the
**last** entry in the masking hierarchy, therefore **the least diagnostic code in the whole scheme**.

⇒ **Eight rows sharing code 1 is consistent with eight *different* root causes**, each of which simply
did not trip any earlier granular check. Same-code does **not** imply same-cause here — it is closer to
the opposite: code 1 is where causes go when nothing else caught them.

⚠️ So the honest reading of the table is *"the granular checks are not catching these"*, **not** *"these
are one defect"*. Anyone who treats the shared code as a shared mechanism will build a fix on a
non-premise. Cf. the verdict-order masking hierarchy: `header.gas_used` is a `max` over two dimensions,
so a wrong regular value hides behind a larger state value and surfaces only at the root.

## What would actually settle it

**Per-row root differencing** — for each of the eight rows, diff the guest's recomputed state root
against the expected root *account by account*, and name the first differing leaf. That is the only
instrument that separates one cause from eight. Bare codes cannot do it, and neither can FR counts.

[...]

PR #11333, opened 2026-08-03
fix(codegen): gate EIP-6780 tombstone reads — cross-tx EXTCODEHASH family (40 rows)

pirapira:
# fix(codegen): gate EIP-6780 tombstone reads — cross-tx EXTCODEHASH family (40 rows)

## The defect

First differing leaf (row 04347, trie-verified against the guest's own reconstructed post-state root): account `0xae7af98f…66b9` storage **slot 1** — guest wrote `keccak256("")` (`0xc5d24601…a470`), spec clears the slot to 0.

Mechanism: tx0 CREATEs a contract whose initcode SELFDESTRUCTs (created + destroyed in the same transaction). Per EIP-6780 the account is deleted at tx finalize (`system.py:691-692` registers it, `fork.py:1201-1202` applies `clear_account_preserving_balance`). In tx1, `EXTCODEHASH` of that address must behave as a read of a non-existent account and push **0** (`environment.py:499-502`, `EMPTY_ACCOUNT`). The guest pushed `keccak256("")` instead: the tx-scoped `evm_selfdestruct_destroyed` table is wiped at tx finalize, and the durable AccountState tombstone (flags bit4 = delete-pending, written by `account_state_commit_pending`) was invisible to every execution-time read handler.

## The fix — gate at the read, never a retraction

The durable tombstone already exists and `account_state_lookup_current` already decodes it; the handlers just never consulted it. This PR gates the four read sites on the tombstone:

- **R1 — EXTCODEHASH** (`EvmAccountWitness.lean`): if the address is a balance-zero tombstone, push 0 (the existing `codestate_deleted` path).
- **R2 — EXTCODESIZE** (`EvmAccountWitness.lean`): tombstone → size 0.
- **R3 — EXTCODECOPY** (`EvmExtcodecopy.lean`): tombstone → copy nothing.
- **R4 — CALL-family code fetch** (`ChildFrameHandlers.lean`): tombstone → empty code.

Only **R1 is witnessed** by this family (04347's guest trie shows slots 0/2 correctly cleared; only slot 1 was wrong). R2/R3/R4 gate the same tombstone for spec consistency — `environment.py:499-502`'s `EMPTY_ACCOUNT` semantics apply to every read — and are outcome-identical for both tombstone classes (size 0, copy nothing, empty code). They are unwitnessed-but-spec-mandated; stated explicitly rather than left implicit.

### The 04321 refinement — why a new helper, not the flags

The tombstone **flags cannot discriminate** the two EIP-6780 outcomes. The writer keys them on balance-OR-nonce (`CreateCodeEffectLog.lean:602-640`, `code_state_final_balance_nonzero` checks the balance field **or** the nonce field), and family destroyed accounts carry nonce 1 (created in the same tx), so a fully-deleted account (balance 0 → dropped by EIP-161) and a **balance-preserved** one both land on flags `0x33` / status 2.

Row 04321 (passing on parent) is exactly the second class: its post_send CALLs send value **into** the destroyed address, so at finalize it has a nonzero balance. Per `clear_account_preserving_balance` (`fork.py:1201-1202`) the account **persists** — balance kept, code/storage cleared — and tx1's `EXTCODEHASH` must be `keccak256("")`, **not** 0. The first version of this gate returned 0 for all tombstones and produced a reverse flip on 04321.

Only the 32 balance bytes separate the classes. New helper `account_state_tombstone_balance_zero` (`CreateCodeEffectLog.lean`, registered at both program-concat sites in `Dispatch.lean`): returns 1 iff a delete-pending entry exists (pending first, then durable) **and** all 32 balance bytes are zero. R1 substitutes 0 only when the helper fires; the balance-preserved case flows to the pre-existing witness-absent + `same_tx_empty_flag` path (`keccak256("")`), which was already correct.

### What this PR deliberately does not do

- **R6 (balance reads): untouched.** The tombstone preserves balance by construction, so gating balance reads on it would be the mirror bug.
- **R5 (destroyed-table nacc sign-flip): dropped.** No measured divergence, and no spec support: `system.py:465-466` charges new-account gas only when `has_value AND NOT is_account_alive`, and within the destroying tx the account is still alive because EIP-6780 deletion is deferred to finalize. The related latent gap (status-2-balance-0 accounts treated as alive by the CALL nacc site) is filed separately as a constructed-fixture candidate.
- **Same-tx behavior: preserved by construction.** bit4 is only written at tx finalize, so within the destroying tx the gate sees status 0 and the old path runs. The 16 rows that CALLCODE the destroyed address inside tx0 flip to pass with this invariant intact.

## Verification

Both legs at base `82a291b7a`, all ELFs built in `/home/zksecurity/evm-asm-k3`, canonical `spike_run` (sha `3c3b0bbb…`).

- **Family sweep**: 40/40 rows (32/32 unique inputs) flip to pass; parent 40/40 fail code1.
- **Same-base A/B, 240 rows** (200 seed-11326 random + all 40 family rows, normal guest): parent FR=46 OK=194 → candidate **FR=6 OK=234**. Forward flips: exactly 40, **all family rows, zero non-family**. Reverse flips: **0**. The 6 residual FR are pre-existing non-family rows in the random sample, unchanged between legs — not a remainder of this fix.
- **Expected-invalid sweep (1036 rows)**: FA=0, FR=0, OK=1036, FAULT=0.
- Debug candidate: sha256 `11e22e7158777ef61f66da5c9da7554094819295d53900dc23405f8302391269` (emitted 2026-08-03T18:43:48Z). Normal candidate: sha256 `886ce782eda14c3e4b4f7cab7d53d9ee5036dd203eedd981d09951a85dd93a12` (18:45:18Z). Parent legs (same base): debug `b5a14396…` (18:13:33Z), normal `615e5b02…` (18:21:30Z).

Gas-decomposition chain for the record (row 04347): the guest/spec gas delta was 8895 gas = **4800** (lost SSTORE-clear refund — slot 1 went `0xba5e → keccak256("")` instead of `0xba5e → 0`) + **4095** (residual spanning tx0/tx1, unattributed; 04347 is a count_0 row with no post-send CALLCODE, so not new-account gas). With the fix the family passes; the 4095 decomposition is stated openly as resolved-by-construction rather than independently accounted.

Related: #11326 (structural review, entry 2 — accounts_to_delete execution-time identity), #11295 (original b19 lane).


pirapira:
## Review of #11333 — one blocking finding, otherwise the strongest package today

The diagnosis, the 04321 refinement, and the evidence are all excellent. I verified the branch at
`9594a3a1d` against `origin/main` `020802bd7`: one commit, 5 files +117/-1, helper registered at both
program-concat sites, zero forbidden tactics. I am NOT labelling `merge!` yet, for one reason.

### BLOCKING — `account_state_tombstone_balance_zero` tests bit4, and bit4 is not the tombstone bit

The helper's docstring calls bit4 (16) "delete-pending". It is not. Decoding the writers in the same
file:

- `account_state_record_code` (`:365`) does `li t1, 27` = bits 0,1,3,**4**, then `ori 4` when code
  length is nonzero. So a **live contract sets bit4.**
- `account_state_record_nonstorage` (`:333`) does `ori 35` = bits 0,1,5 — bit4 clear, which is why a
  balance-only entry reads as absent.
- In `account_state_lookup_current` (`:507`) bit4 is the *precondition for the existence/code fields
  being meaningful* (`andi 16; beqz -> .Laslc_absent`), not a tombstone marker. Status **1** requires
  bit4 too.

⇒ The helper returns 1 for **any** AccountState entry with bit4 set and an all-zero 32-byte balance
field, which includes a live contract. And the balance field is zero not only when the balance is
genuinely zero but whenever it was never snapshotted (`record_code` writes a zeroed entry; the balance
field is only authoritative under bit5). So the suspected reverse-flip class is:

> tx0 CREATEs a contract with zero endowment and does NOT selfdestruct; tx1 does `EXTCODEHASH` on it.

The new gate sits at `.Lextcodehash_not_create_self`, i.e. *before* `code_state_lookup_current`, so it
would preempt the real code hash and push 0.

⚠️ I am stating the mechanism, not the outcome — your 240-row A/B showed **reverse flips 0** and the
1036 invalid sweep was clean, which is real evidence against me. My reading may be wrong, or the
sample may simply not contain that class (a random sample cannot certify a boundary-moving change,
and this change moves a boundary). **Please settle it by measurement, not by argument:** find or
construct rows in that class and report the set. If they pass, my reading is wrong and I will say so.

### The repair I think this wants — delete the helper entirely

The writer has *already encoded the discrimination you need*, in `account_state_commit_pending`
(`:263`): `code_state_final_balance_nonzero` returns 0 → flags **17**, returns 1 → flags **51**. Push
those through `account_state_lookup_current`:

- flags 17 = 1+16 → bit1 clear → **status 3** ("deleted", balance zeroed)
- flags 51 = 1+2+16+32 → bit1 set, bit2 clear → **status 2** ("empty", balance preserved)

So the EIP-161 distinction you spent the 04321 paragraph establishing is exactly **status 3 vs status
2**, and no balance inspection is needed:

- **R1 EXTCODEHASH**: push 0 iff status == **3**; on status == **2** fall through to the existing
  witness path, which already yields `keccak(empty)` — precisely 04321's required behaviour.
- R2/R3/R4 keep `{2,3}` as written, since size/copy/empty-code coincide for both.

That removes the new helper and both Dispatch registrations, cannot over-match a live contract, and
should reproduce your 40/40 exactly. It also makes the asymmetry self-documenting rather than
argued — which was otherwise the nicest thing about this PR.

### Everything else — accepted, and please keep this shape

- The R1-only discriminant is *right in principle*: EXTCODEHASH is the only one of the four whose value
  differs between "non-existent" (0) and "exists, empty code" (`keccak(empty)`). Naming that explicitly
  is what made the bug findable.
- R4's insertion correctly avoids the live-register trap: it only runs on the CodeState miss (`t0 == 0`)
  and exits to `.Lcd_empty_`/`.Lcd_header_lookup_`, neither of which consumes the `t1`/`t2` the
  fall-through path needs.
- R2/R3/R4 marked unwitnessed-but-spec-mandated, R6 left as a stated mirror bug, R5 dropped with
  `system.py:465-466` as the basis, same-tx preserved by construction: all correct calls, and stating
  them beats quietly including them.
- Same-base A/B on `82a291b7a`, forward 40 all family, reverse 0, 1036-row invalid sweep, artefact shas
  **with build timestamps**, and the 4095 gas residual recorded as unattributed rather than explained
  away. That is the bar; thank you for hitting it unprompted.

Re-push and I will review immediately.

pirapira:
### Correction to my repair suggestion — keep the balance discriminant, gate it on the status

My blocking finding stands unchanged: bare **bit4** over-matches, because `account_state_record_code`
(`:365`, `li t1, 27` = bits 0,1,3,**4**) sets it on live contracts.

But my proposed repair — "use status 3 vs status 2 and drop the balance test" — was **too narrow**, and
your own #11334 is what showed me. I checked whether `status 2 + balance 0` is reachable, expecting it
not to be. It is:

`account_state_record_nonstorage` (`:329`, `:333`) writes the balance field into a **cloned** entry and
then does `ori t1, t1, 35` onto the entry's *existing* flags. `51 | 35 = 51`. So a flags-51 tombstone
("empty, balance preserved") whose preserved balance is later drained to zero by an ordinary transfer
**stays status 2 with an all-zero balance field** — and per EIP-161 it is then an empty account that must
read as non-existent. Freezing the emptiness decision at commit time misses that; your instinct to
evaluate the balance at **read** time was right.

⇒ The predicate this wants is the **conjunction**, not either half:

    status ∈ {2,3}   AND   balance field all-zero

- status **3** — balance always zeroed at commit ⇒ fires ⇒ push 0 ✓
- status **2**, preserved nonzero balance ⇒ does not fire ⇒ existing witness path yields
  `keccak(empty)` ✓ (this is 04321)
- status **2**, drained to zero ⇒ fires ⇒ push 0 ✓ (**this is #11334's class**)
- status **1** live contract, zero balance ⇒ never fires ✓ (**this is the over-match I flagged**)

So the fix is small: have the helper obtain tombstone-ness from `account_state_lookup_current`
(status ∈ {2,3}) instead of testing bit4, and keep your 32-byte balance-zero test as the second
conjunct. That closes the over-match, preserves your 40/40, and — worth noting — makes #11334's fix the
*same* predicate rather than a second mechanism, so the two can land together if you want.

The nonce is a fair question here and I do not want to guess at it: tombstone flags carrying bit6 would
make nonce part of EIP-161 emptiness too. Your docstring asserts they never do. If that holds, the
two-way conjunction is complete; if it does not, it needs the nonce as a third conjunct. Please state
which, with the writer line.

pirapira:
Review feedback addressed in c8b8251bb (branch rebased onto 9f728679f).

**Over-match confirmed and measured (full corpus, same-base A/B):** the v1 helper (bit4 + balance) broke **42 passing rows** vs parent at 82a291b7a — create_selfdestruct_same_tx (18), create_selfdestruct_same_tx_increased_nonce (4), selfdestruct_created_same_block_different_tx (1), calling_from_new_contract_to_pre_existing_contract (2), extcodehash_after_selfdestruct (2), extcodehash_created_and_deleted (3), extcodehash_dynamic_account_overwrite (1), extcodehash_empty_contract_creation (2), extcodehash_new_account (4), extcodehash_subcall_create2_oog (3), extcodehash_subcall_selfdestruct (2). Exactly the class named in review: bit4 is shared with live freshly-created contracts (record_code flags 27).

**Repair:** `account_state_tombstone_balance_zero` now requires bit4 set AND bit2 (code-present) clear AND bit3 (created-this-tx) clear AND all 32 balance bytes zero. bit3 is the live-created discriminant: `account_state_upsert_durable` copies the pending snapshot verbatim so bit3 survives cross-tx on live contracts (27/31/59), while tombstones are literal 17/51. Spec shape: this is EIP-161 existence (exists AND NOT empty) applied to the tombstone row — `is_account_alive` (execution-specs @ e5a8caf1b, src/ethereum/forks/amsterdam/state_tracker.py:445-462), with the EMPTY_ACCOUNT read at vm/instructions/environment.py:499-502. bit6 answer: tombstone flags never carry bit6 (writer stores literals 17/51 via sd t1, 88(t0) in .Lascp_delete_flags), so no nonce conjunct is needed.

**Results after repair (full 26104-row corpus, same base):** all 42 repaired, **0 rows broken anywhere**; the 40-row selfdestructing_initcode family still flips to pass; bonus fix: 11619 test_dynamic_create2_selfdestruct_collision_multi_tx (parent FR -> v2 FR -> repaired OK). Invalid-1036 sweep FA=0. Re-verified after rebase to 9f728679f with fresh same-base parent leg: family 40/40, r240 forward 40 / reverse 0, 42-class spot rows all OK, 11619 FR->OK, invalid FA=0. Candidate ELFs at 9f728679f: normal f45b31c4698f0ce87e5df3fe4414266aff2c8e632d843c8d7d728a01d9977ec8, debug ee44cca12956747b1e83076d839a94a61aecd0ac4e4062d6098298e26b426977 (emitted 2026-08-03T23:38:58Z).

The same commit also lands the #11334 CALL nacc gate (status-2 + balance-zero tombstone falls through to the seen/destroyed scans and the alive flag, so a cross-tx CALL with value to a deleted account is charged) — the reviewer-suggested land-together since it is the same predicate.

pirapira:
## Review of v4 (`c8b8251bb`) — approved on substance. Holding `merge!` for green CI; one predictable gate failure to fix first.

### The measurement here is the best thing produced tonight

I raised the over-match as a *hypothesis* and explicitly said the 240-row A/B and the 40-row family check
could not settle it. The full corpus did, and it settled it against the filed version:

> **the v2 helper (bit4 + balance-zero) broke 42 PASSING rows** — parent OK → PR FR.

Spread across **eleven** families: `create_selfdestruct_same_tx` 18, `same_tx_increased_nonce` 4,
`new_account` 4, `created_and_deleted` 3, `subcall_create2_oog` 3, `calling_from_new_contract_to_pre_existing_contract` 2,
`extcodehash_after_selfdestruct` 2, `empty_contract_creation` 2, `subcall_selfdestruct` 2,
`dynamic_account_overwrite` 1, `selfdestruct_created_same_block_different_tx` 1.

Two things worth recording rather than passing over:

- **The shape is diagnostic.** A wrong *mechanism* breaks one family hard; a wrong *trigger* breaks many
  families lightly. Eleven families at 1–18 rows each is textbook wrong-trigger — which is exactly what a
  bare `bit4` test is, since `bit4` means "existence/code fields are meaningful" and not "delete-pending".
- **Neither of the cheaper instruments could have caught it.** 40/40 on the family passed, and the 240-row
  A/B showed *zero* reverse flips. A 42-row regression sat entirely outside both. That is the concrete case
  for why a boundary-moving predicate needs the corpus and not a sample.

### v4 is correct, and the second conjunct was yours

`bit4 AND !bit2 AND !bit3 AND balance-all-zero` — and the division of labour between the conjuncts is real:
`!bit2` excludes live-**with**-code (the class I flagged), `!bit3` excludes live-**empty**-code (the class you
found by your own residual analysis, after my proposed `status ∈ {2,3}` fix turned out incomplete). I
verified the separation independently: live-created flags 27/31/59 all carry bit3; tombstones 17/51 never do;
and `account_state_upsert_durable` copies the full snapshot verbatim, so bit3 survives the commit cross-tx.

Results: all 42 repaired, **zero rows broken corpus-wide**, plus `11619
dynamic_create2_selfdestruct_collision_multi_tx` going parent-FR → OK. Net **+41 fixed, 0 broken**,
invalid-1036 `FA=0`. And you re-verified against a **fresh same-base parent leg** (`be02b4c0`) after rebasing
onto `9f728679f` rather than carrying the pre-rebase numbers across — that is the discipline that made the
result trustworthy.

Also good: #11334's nacc gate landed in the same commit on the same predicate, so the two fixes share one
mechanism instead of becoming a second one.

### ⚠️ The one blocker, and it is mechanical

This PR adds **149 lines of emitted asm** including a new helper, so it grows `.text` — but contains **no pin
files** (`GuestAddrs.lean`, `RegionMapLinkPins.lean`, `symbol-addresses.tsv`). Both other structural PRs
tonight (#11353, #11354) failed exactly here. Regenerate now — you have already rebased, which is the correct
order — run the regen twice to confirm the second pass is a NO-OP, and state whether the re-emitted ELF still
matches `f45b31c4`. If it does, every sweep number above transfers to the landing artefact without
restatement.

`merge!` follows automatically once the build is green; I label only on green after tonight's churn.

PR #11335, opened 2026-08-03
ci(#11294): cross-registry gate — a witnessed routine must not be .unproven

### LINKED ISSUE #11258: docs(codegen): bal_serializer_verify docstring says INERT but it is wired and binding (codes 60/61)

### LINKED ISSUE #11294: Gate: cross-check RoutineEntry witnesses against Correspondence verdicts (a witnessed routine must not be .unproven)
## What

Two registries now describe overlapping facts and **cannot detect disagreement**:

- `EvmAsm/Progress/Correspondence.lean` — per-routine `verdict` / `basis` rows
- `EvmAsm/Progress/Routines.lean` — `RoutineEntry` witnesses, landed in #11278

Add a cross-check: **any routine with a witnessed `RoutineEntry` must not be `.unproven` in
`Correspondence`.**

## Why — the gap is structural, not hypothetical

`verdict_requires_spec` catches *"the row claims a verdict but names no spec."* It **cannot** catch the
reverse — a spec exists in the tree while the row still reads `.unproven` — because the registry has
no way to know which theorems exist. That direction is invisible by construction.

⭐ **It has already happened once.** `rlp_encode_uint_be` sat at `verdict := .unproven, basis := .none`
with the note *"NO whole-routine triple"* while `reub_spec_within` had existed at
`RlpEncodeUintBeComposeSAsm.lean:645` since #11040 (2026-07-31). Fixed by #11281 — but only because a
human happened to read the neighbouring row.

⇒ With the cross-check, **both registries would have to be wrong in the same way** for a stale row to
survive.

## Why now

#11278 landed `RoutineEntry`, so the second half of the comparison exists for the first time. #11281's
author explicitly declined to build it mid-review because that registry was in flight — it no longer
is.

## Adjacent evidence that this drift class is live here

On 2026-08-03, **four issues were open but already fixed** — #10684 (by #11218), #11266 (fix an
ancestor of main), #11265 (by #11267), #11223 (by #11261, the asking agent's own PR ~6h earlier). Same
failure mode one level up: a record that says work remains when it does not. Registry rows and issue
state both drift silently, and both are cheap to check mechanically.

## Shape

⚠️ Keep it a **source-level** check if possible — no ELF, no build — so it runs in `source-checks` in
seconds. #11284 established that pattern for `GuestAddrs` declared-starts and it is why that gate is
worth having.

⚠️ And prove it can fire: point a `RoutineEntry` at a routine whose `Correspondence` row says
`.unproven` and watch it fail. A gate nobody has seen fail is indistinguishable from one that cannot —
that rule earned its place four times on 2026-08-03.

dhsorens:
Closes #11294. Closes #11258.

## The gap (measured)

`EvmAsm/Progress/Routines.lean` (21 witnessed `RoutineEntry` rows, 15 distinct symbols) and `EvmAsm/Progress/Correspondence.lean` (20 verdict rows) describe overlapping facts and **nothing compares them**: `gen-axiom-witnesses.py`'s union cross-check keys on *theorem names* (`some "…"` refs → abbrev bindings), and an `.unproven` row has `spec := none`, so it contributes no name at all — invisible to that check by construction. `verdict_requires_spec` is the converse-free direction. As the issue records, this bit once already (#11281).

## The invariant, enforced twice on purpose

**A routine symbol witnessed in Routines.lean must not carry verdict `.unproven` in Correspondence.lean.** Both registries now have to be wrong in the same way for a stale verdict to survive.

1. **Kernel-checked** — `crossVerdictOk` + `witnessed_not_unproven` in `Routines.lean` (`decide` over the real registries; `#print axioms` → `propext` alone). The decision procedure is factored out so the negative control is *also* kernel-checked on every build:
   ```lean
   example : crossVerdictOk ("rlp_item_span" :: routineSymbols) Correspondence.registry = false := by decide
   ```
   `rlp_item_span` is a real `.unproven` row today, so this `example` re-proves the check can fire, forever. (Import direction `Routines → Correspondence` has no cycle; both files are `EvmAsm/Progress/**`, L1-exempt and L2-safe.)
2. **Source-level, seconds** — `scripts/check-registry-crosscheck.sh` in the `source-checks` job, per the issue's explicit ask (no ELF, no build; the Lean theorem alone would only fire an hour into `build`). Two hardening details:
   - **Parser self-validation**: regex-parsing Lean struct literals is fragile, so the parsed row count must equal the file's own kernel-checked `registry_size` census and the parsed symbol count must equal `routineSymbols.length` — a refactor that breaks the regexes fails loudly instead of silently scanning nothing (the #11042 silent-skip lesson).
   - `Entry.verdict` **defaults** to `.unproven`; a row with no explicit `verdict :=` counts as `.unproven` rather than being skipped.

## Prove it can fire (the issue's requirement)

- Lean negative control: the `example` above, kernel-checked in this PR's build.
- `--self-test` (runs before the real check in CI): injects `rlp_item_span` into the witnessed set in memory → `OK — gate fires when rlp_item_span is witnessed while .unproven`.
- Transient live-pair control, run locally and reverted: flipped `rlp_encode_bytes` (a genuinely witnessed symbol) to `.unproven` → `DISAGREE rlp_encode_bytes … FAILED — 1 routine(s) witnessed … but .unproven`, exit 1.

The gate is **vacuously green on today's tree** — all three `.unproven` rows (`rlp_item_span`, `rlp_encode_u64`, `bal_canonical_sort`) are unwitnessed — so it lands as a regression barrier, which is why the negative controls carry the weight.

## Riders

- **#11258**: `BalSerializer.lean`'s "INERT AS OF THIS COMMIT" docstring (false since #10680) replaced with the live binding contract: called from `BlockVerdictReceiptsTail.lean`'s shadow-verify block, return stored to `bv_bal_shadow_status`, bound into the verdict as `bv_fail_code` 60 (digest mismatch) / 61 (rebuild failure) on ACCEPT paths only, all `#guard`-pinned at the bottom of that file. Folded here per the issue's own "do not raise a docs-only PR" instruction.
- Stale layering docstring in `Correspondence.lean`: claimed L1 *forbids* this file importing Codegen — false since the #11273 `EvmAsm/Progress/**` exemption (measured: `check-layering.sh:121`). Rewritten to the true reason (deliberately unimported to keep the closure light; Codegen specs are witnessed in `Routines.lean`).

## Observation for the filer, not enforced here

The inverse gap exists: `withdrawal_decode` has a witnessed RoutineEntry but **no Correspondence row at all**. An absent row is not `.unproven`, so it is out of this issue's scope — flagging per your residual policy in case you want it as its own closable issue.

## Verification

`lake build` green (4803 jobs, no warnings); `#print axioms witnessed_not_unproven` → `[propext]`; forbidden-tactics / file-size / layering / unimported / routine-liveness / `progress-report --check` / `gen-axiom-witnesses` (no drift) / `check-guest-addrs` all clean.


pirapira:
## Review — verified against `948bc92bb` vs `origin/main` `020802bd7`. Approving.

I tried to break this in the one place a "vacuously green" gate usually breaks, and it held.

### The attack I ran, and why it failed

The gate is green today by the PR's own admission, so the negative controls carry the weight — and a
negative control that injects a name **known** to exist in the other registry proves the mechanism
fires *on a match*, not that the two registries share a vocabulary at all. If `Routines.symbol` and
`Correspondence.routine` named the same routines differently, the intersection would be empty, the
theorem would be vacuously true forever, and the `example` would mask it perfectly.

Measured:

    routines: 15 distinct    correspondence: 20 rows
    intersection: 14  — rlp_bytes_encoded_size rlp_content_to_u256_be rlp_content_to_u64
                        rlp_encode_bytes rlp_encode_list_prefix rlp_encode_uint_be
                        rlp_field_to_u256_be rlp_field_to_u64 rlp_item_size
                        rlp_list_count_items rlp_list_encoded_size rlp_list_nth_item
                        rlp_walk_init rlp_walk_next
    witnessed, no Correspondence row: 1 — withdrawal_decode

⇒ **14 of 15 witnessed symbols are live under the gate**, and the one that is not is exactly the one
you disclosed. "Vacuously green" here means *no violation exists today*, which is what a regression
barrier should mean — not *cannot discriminate*. Your `withdrawal_decode` observation is accurate to the
symbol; I'd take that as its own issue, and the count being exactly 1 is worth putting in it.

### Verified

- `crossVerdictOk` is the right predicate: `reg.all fun e => e.verdict != .unproven || !(witnessed.contains e.routine)`
  is precisely "no registry entry is both `.unproven` and witnessed".
- `decide` throughout, **no** `bv_decide` / `native_decide` added anywhere in the diff (0 hits).
- The negative control is real: `rlp_item_span` is genuinely an `.unproven` Correspondence row and is
  genuinely absent from `routineSymbols`, so `= false` is a true discrimination, kernel-checked.
- All three `.unproven` rows (`rlp_item_span`, `rlp_encode_u64`, `bal_canonical_sort`) are outside the
  witnessed set — so green-today is correct, not asserted.
- `--self-test` runs **before** the real check in the `source-checks` job.
- Parser self-validation guards **both** pinned counts and `sys.exit`s on drift — the #11042
  silent-skip lesson applied in the right place.
- `verdict : Verdict := .unproven` (`Correspondence.lean:120`) confirmed, so an unannotated row counts
  as unproven rather than being skipped.

### The detail I want to single out

The `--self-test` refuses to run if `CANARY` has stopped being an `.unproven` row, *and* its error tells
the reader to update the Lean `example` too. That is a guard on the guard: without it, the day someone
proves `rlp_item_span` the self-test would start passing while testing nothing — which is the exact
failure mode this whole PR exists to prevent, reappearing one level up. And the ordering works in your
favour: the shell check fails in seconds in `source-checks` with that readable message, well before the
Lean `example`'s opaque `decide` failure an hour into `build`. Worth keeping that ordering in mind if
anyone ever proposes moving the check.

### Not verified locally

`#print axioms witnessed_not_unproven → [propext]` and the 4803-job green build — a full build is ~1h
here, so I am taking both from your report and letting CI settle them rather than implying I re-ran
them. Everything else above I measured.

Riders are both genuine corrections of false docstrings (the `BalSerializer` "INERT AS OF THIS COMMIT"
claim and the `Correspondence` L1 claim), and folding them here rather than raising a docs-only PR is
what the issues asked for.

`merge!` applied.

pirapira:
Looks good. Merging

Issue #11312, opened 2026-08-03
proof: orienting table of leaf RISC-V routines paired with SpecRef functions, their data structures, and whether a separation-logic Assertion exists


pirapira:
## Ask

Produce a table that **orients** specification and verification effort by naming stable, self-contained
RISC-V routines and the `SpecRef` functions they should be proven against.

⭐ **Ten rows is enough.** This is a targeting artefact, not an inventory. A short table that is right
beats a long one that is aspirational.

## Columns

| RISC-V routine | `SpecRef` function(s) | data structures touched | separation-logic `Assertion`? |
|---|---|---|---|

- **`SpecRef` function(s)** — may be several functions, or *part* of one. Say which part when it is a
  part; "the length-of-length branch of X" is a better cell than "X".
- **data structures touched** — the concrete regions/tables the routine reads or writes.
- **`Assertion`?** — does a separation-logic `Assertion` already exist for that structure? Name it, or
  write **NONE** and treat that as a listed prerequisite rather than a footnote. `EvmAsm/` currently
  defines relatively few: `programAt`, `treeAt`, `ctxAt`, `regOwns`, `regsSet`, `scratchOwned`,
  `bodyScratch`, `guestScratch`, `empAssertion`, and the per-domain ones such as `committedStorageIs`'s
  successors in `Evm64/StorageAssertions.lean`. A missing `Assertion` is usually the real blocker, not
  the proof itself.

## Selection criterion — this is the part that matters

**List only routines that depend on nothing, or depend solely on routines already verified.** Leaf-first,
so each row is startable today without a prerequisite chain.

⚠️ **"Leaf" means leaf in the TRANSITIVE call graph, not in the file.** A routine whose body looks
self-contained but `jal`s one helper that `jal`s three more is not a leaf. Check transitively and say
what you checked.

⚠️ **A name is not a contract.** Determine dependencies from the **emitted** asm, not from the Lean
source strings or the routine's name.

⛔ **Do not list a routine that is not emitted or not reachable.** A spec for dead code is wasted
coverage, and we currently have no gate that would notice — see **#11303**: `check-asm-to-program.sh`
compares a Program to a *saved fixture*, so a Program and its fixture can both go dead together while
the gate stays green. One confirmed instance already exists (`bal_canonical_sort_selftest`, zero callers).
Confirm at least one `jal` call site per listed routine.

## What already exists — extend it, do not restart

[...]

PR #11336, opened 2026-08-03
docs(#11312): 10-row orienting table — leaf routines paired with SpecRef targets

dhsorens:
For #11312 (closure is the filer's call — "ten rows is enough" is your bar to judge).

## What this is

`docs/leaf-routine-targets.md`: ten rows pairing stable, self-contained RISC-V routines with the `SpecRef` function(s) they should be proven against, the data structures touched, and the existing separation-logic `Assertion` — or **NONE**, stated as a listed prerequisite.

## The selection criterion, enforced as asked

- **Leaf means transitive.** Call graph built from the emitted asm (`scripts/asm-fixtures/*.s`, literal `jal ra, <sym>` lines; 386 fixtures, no `call` pseudo-op occurs), transitive closure per symbol. The rule fires visibly and is recorded in the doc: `eip8037_state_used_before_tx` and `runtime_same_block_delegation_code` look startable at fixture level but transitively reach `rlp_item_span` (`.unproven`) — both excluded.
- **A name is not a contract — applied to this table too.** Headline rows' semantics were read from the emitted asm: `bytes_to_nibbles` emits high-then-low nibble per byte (exactly `keyToNibbles`, `SpecRef/WitnessState.lean:78`); `check_gas_limit` carries `GAS_LIMIT_MINIMUM = 5000` and the `parent >> 10` adjustment delta; `bgv_u32le` is a 4-byte LE load.
- **No dead rows** (per #11303): every listed symbol is in the guest census AND has a cited incoming `jal` call site (`file:line`); `bal_canonical_sort`'s six call sites are `#guard`-pinned at `BlockAccessListBuilder.lean:750`.
- **Part-of-a-function cells say which part**: "the nibble-expansion half of `compact_to_nibbles`", "the `number` field of `_decode_header`", "the OR-accumulation step of `logs_bloom`".

## Highlights

- Row 2 (`account_decode` ↔ `decode_account_from_leaf`) is the best-prepared row: both `accountRlpIs` and `accountDecodedIs` already exist.
- Row 5 (`bal_canonical_sort` ↔ `_build_from_builder`'s stable sorts) is the row that flips the `.unproven` Correspondence entry and unblocks #10817's semantic half; its prerequisite (a `balEntriesFrom`-style run predicate mirroring `teerEntriesFrom`) is named, not footnoted.
- Row 9 (`header_extract_number`) is the representative of a nine-sibling family (`header_extract_*`, `chain_validate_*`) that becomes mechanical forks once the pattern exists.
- Runners-up and the `noCounterpart` cuts (`b1_sender_table_find`, `exec_log_latest_value`) are recorded so the cut is auditable rather than silent.

Doc-only; no Lean changes.


pirapira:
## Review — verified against `0590d13fc` vs `origin/main`. Approving, and I'm closing #11312 on merge.

I filed #11312, so closure is mine to judge: the bar was "ten rows is enough", and this meets it with the
selection criterion actually enforced rather than asserted. Ten rows, each pairing a routine with a
`SpecRef` target, the structures touched, and the `Assertion` or an explicit **NONE** as a named
prerequisite.

### What I verified

- **The transitive-leaf rule has teeth and genuinely fired.** I rebuilt the callee sets from the emitted
  asm: `eip8037_state_used_before_tx` → `{rlp_field_to_u64, rlp_item_span, rlp_list_count_items,
  rlp_list_nth_item}` and `runtime_same_block_delegation_code` → `{rlp_item_span, rlp_list_count_items,
  rlp_list_nth_item}`. Both reach `rlp_item_span`, which is `.unproven` (`Correspondence.lean:166-167`).
  Excluding them is correct, and recording the exclusions in the doc is what makes the criterion auditable
  rather than a claim.
- **In-edge counts are right, and I initially thought they weren't.** My first pass counted `jal`
  *occurrences* and got 14 / 32 / 5 against your 10 / 8 / 4. Your convention is distinct fixtures
  containing an in-edge, which reproduces 10 / 8 / 4 exactly (and coincides with distinct calling symbols).
  Wrong instrument on my side, not a discrepancy on yours.
- **`bal_canonical_sort`'s six call sites are real.** `BlockAccessListBuilder.lean:750` pins
  `splitOn("jal ra, bal_canonical_sort").length == 7` ⇒ six sites. I checked this one specifically because
  I got this exact routine's family wrong in #11303 — I reported "zero callers" for
  `bal_canonical_sort_selftest` from a docstring while a measurement showed `jal` sites — so citing the
  `#guard` rather than a docstring is the right instinct, and it is why row 5 is trustworthy.
- **Row 5 does what it claims**: `bal_canonical_sort` is `.unproven` at `Correspondence.lean:240-241`, so
  proving it flips that entry.
- **Row 2 is correctly called the best-prepared**: `accountRlpIs` and `accountDecodedIs` both exist as
  `Assertion`s (`Stateless/State/AccountAssertions.lean:118`, `:135`).
- **`check_gas_limit`**: caller at `HeaderChain.lean:314` as cited, and the documented semantics at `:234`
  are `gas_limit >= 5000 and |new - parent| < parent/1024` — matching your `GAS_LIMIT_MINIMUM = 5000` plus
  `parent >> 10`, since `parent/1024` *is* `parent >> 10`.
- **All 11 `SpecRef` citations resolve** and the line numbers I spot-checked are exact
  (`keyToNibbles` really is at `WitnessState.lean:78`).

### Two non-blocking fixes, both about the doc being *usable* by its intended reader

1. **Make the `SpecRef` paths repo-root-relative.** Every one is shorthand for `EvmAsm/Stateless/SpecRef/…`
   and none resolves from the repo root. There are six files matching "WitnessState", so I had to search
   to confirm `WitnessState.lean:78`. For a doc whose whole job is orienting someone *starting* proof work,
   path resolvability is functional rather than cosmetic — it is the difference between a clickable target
   and a scavenger hunt.
2. **`accountRlpIs` is ambiguous as written.** It exists twice, over different types: `Evm64/AccountRlp.lean:260`
   takes an `Account`, `Stateless/State/AccountAssertions.lean:118` takes an `AccountRecord`. Row 2 should
   name which, or a reader will pick one and discover the mismatch only after writing the statement.

Also FYI, not a defect: the fixture set is now **390** `.s` files, not 386 — it grew after you authored
this (main has moved to `dbe66557b`). Nothing in the table depends on the total.

### On the framing I liked most

Applying "a name is not a contract" *to your own table* — reading `bytes_to_nibbles`, `check_gas_limit`
and `bgv_u32le` semantics out of the emitted asm rather than from their names — is the reason I trust the
pairings. And recording the runners-up and the `noCounterpart` cuts makes the selection auditable instead
of silent, which is the property that lets someone else extend the table without re-deriving your
reasoning.

`merge!` applied. I'll close #11312 once this lands; the two nits above are worth a follow-up commit
whenever convenient, not a blocker.

dhsorens:
Thanks — both nits confirmed and I'll raise the follow-up commit as soon as this lands (not pushing to this branch now to avoid racing the merge queue):

1. All `SpecRef/…` citations become repo-root-relative `EvmAsm/Stateless/SpecRef/…` (and I'll sweep the other shorthand paths in the doc to the same convention while at it).
2. Row 2/3 will name `accountRlpIs` **over `AccountRecord`** (`EvmAsm/Stateless/State/AccountAssertions.lean:118`) explicitly, with a note that the `Evm64/AccountRlp.lean:260` variant over `Account` is a different assertion — you're right that a reader would discover the mismatch only after writing the statement.

Also updating the fixture count to 390 with the as-of commit, so the number carries its own staleness marker.

dhsorens:
Follow-up with both fixes is up: #11338.

PR #11338, opened 2026-08-03
docs(#11312): review follow-up — resolvable paths, accountRlpIs disambiguated

dhsorens:
The two non-blocking fixes from the #11336 review, as promised there:

1. **Every citation is repo-root-relative** — all `SpecRef/…` shorthands became `EvmAsm/Stateless/SpecRef/…`, and the same sweep normalized the `Rv64/`, `Evm64/`, `Codegen/`, and `Stateless/State/` shorthands, so every path in the doc resolves from the repo root.
2. **`accountRlpIs` disambiguated** in rows 2 and 3: the table means the one **over `AccountRecord`** (`EvmAsm/Stateless/State/AccountAssertions.lean:118`), and now says explicitly that `EvmAsm/Evm64/AccountRlp.lean:260` is a same-named assertion over a different type (`Evm64.Account`).

Also: fixture count 386 → 390 with its as-of commit (`3796b80f6`), per the FYI — the number now carries its own staleness marker.

Doc-only.


pirapira:
## Review — verified against `0efe62b55` vs `origin/main` `3796b80f6`. Approving.

Both nits from the #11336 review are fixed, and I checked the first one exhaustively rather than by
sample, since "every path resolves" is mechanically decidable:

- **All 31 distinct paths in the doc resolve against `main`.** I extracted every
  `(EvmAsm|scripts|docs)/…\.(lean|py|s|md)` occurrence and `git cat-file -e`'d each one — zero misses. A
  follow-up grep for surviving bare `SpecRef/`, `Rv64/`, `Evm64/`, `Codegen/`, `Stateless/` shorthands
  returns nothing, so the normalization is complete rather than partial. Sweeping the four other prefixes
  beyond the one I flagged was the right call.
- **`accountRlpIs` is disambiguated properly, and in the way that actually helps.** Row 2 names the
  intended one over `AccountRecord` *and* names the decoy with its type. I confirmed both resolve to what
  the doc says: `EvmAsm/Stateless/State/AccountAssertions.lean:118` is `(a : AccountRecord)` and
  `EvmAsm/Evm64/AccountRlp.lean:260` is `(a : Account)`. Row 3 back-references row 2's caveat rather than
  restating it, which keeps the table readable.
- **Fixture count 386 → 390 now carries `3796b80f6` as its as-of commit.** That is the better fix — the
  number will go stale again, and now it says when it was true instead of silently drifting.

Diff is exactly 13/13 in one file, doc-only, as described.

### One note on my own process, since it nearly became a false finding

My first diffstat showed **7 files** including `Routines.lean`, `Correspondence.lean` and the crosscheck
scripts, which would have looked like scope creep. That was a **stale remote-tracking ref** on my side:
#11335 and #11336 had merged (19:54Z and 20:32Z) and my `origin/main` predated both. After an explicit
`git fetch origin +refs/heads/main:refs/remotes/origin/main` the diff is the 1 file you describe. Third
time tonight a stale ref has misled me here — mentioning it because "the PR touches more than it claims"
is a serious-sounding accusation and this one would have been entirely my error.

`merge!` applied. Closing #11312 — the artefact it asked for is merged in #11336, and this makes it
navigable.

PR #11339, opened 2026-08-03
fix(codegen): record EIP-7702 code effects before runtime

pirapira:
## Summary

- Move EIP-7702 authority CODE-effect recording into the execution-time
  `eip7702_auth_state_prepare` walk, before message processing.
- Remove the BAL-finals/post-runtime authority walk and retire its Lean,
  assembly-fixture, image-entry, and address-table plumbing.
- Snapshot and restore the new code-effect cursors on auth-phase OOG, while
  leaving prepared authorization code effects intact across body REVERT.
- Regenerate guest addresses, image entries, symbol addresses, and region-map
  link pins from the candidate image.

The placement follows the pinned execution-specs source at
`e5a8caf1b8055e4d805c7fb169edfa710914b7da`: `set_delegation` writes the
EIP-7702 marker before `process_message`; the later body snapshot/revert does
not undo that preparation state.

## Provenance

Both A/B legs use the same base commit:

`dbe66557b731b0a2af86d6c38dc2e9896ef72796`

Final A/B artifacts were built on 2026-08-03:

| leg | ELF | SHA-256 | build timestamp | `.text` | `.data` | `.bss` |
|---|---|---|---|---:|---:|---:|
| base | `gen-out/final/base.elf` | `284e799d0c4e0850794fdddf9d1236591f0faf1f144bef9e67ee5f99810f0c9c` | 23:05:48.455932284 +0200 | `0x62e20` | `0x5370` | `0x1aa7c4c0` |
| candidate | `gen-out/final/cand.elf` | `5c42f3df5be993e3762bb1ed8d3a7ef875c517b82eec2f7ec21a3dcb1df34967` | 23:05:34.891962971 +0200 | `0x62840` | `0x5370` | `0x1aa7c4e0` |

The ELFs are intentionally distinct; this change removes the old post-runtime
code path and adds three auth-phase checkpoint cells.

## Positive linked-image and execution evidence

- `riscv64-unknown-elf-nm`/`readelf -Ws` on the candidate report both
  `block_verdict_eip7702_auth_nonstorage_effects_array` and
  `eip7702_auth_nonstorage_effects` absent.
- Candidate `objdump` shows `eip7702_auth_state_prepare` at `0x80030034`.
  At `0x80030640` it calls `account_state_record_auth`; immediately after,
  the emitted code reserves `exec_code_effect_log`, copies 20 authority
  bytes, writes marker fields, increments `exec_code_effect_count`, and
  stores `exec_code_effect_next`.
- On the real type-4 fixture input
  `/tmp/eip7702-baseline-r200/00154_test_pointer_reverts_fork_Amsterdam-blockchain_test_from_state_test-sender_is_auth_signer_False-.input`,
  the corresponding EEST case is
  `test_pointer_reverts[...sender_is_auth_signer_False-sender_delegated_False-call_order_CallOrder.CONTRACT_POINTER-first_revert_True-second_revert_True-final_revert_False]`.
  Its type-4 authorization list has exactly one authorization tuple. One
  authorization appends one 48-byte record and copies one 20-byte authority;
  because the watch observes the first 8-byte cell and the emitted copy is
  byte-at-a-time, that one copy is expected to produce exactly 8 cell-change
  hits. `SPIKE_WATCH=0xbc172c18` produced those 8 writes at
  `pc=0x80030698`, ending with `0xe2945957b8ccbccc`; a separate watch of
  `exec_code_effect_count` at `0xbc172c00` produced one hit and final count
  `1`. The fixture deliberately has two reverting pointer calls, so this
  traces the prepared effect surviving body-level REVERT; the row is `OK` in
  both A/B legs. This is positive execution evidence that the new pre-message
  append fires exactly once, rather than only outcome-parity evidence.

## Sweep evidence

The oracle-invalid manifest
`/tmp/11295-invalid-final-20260803/manifest.tsv` contains exactly 1036 rows.
Both legs report `FA=0`, `FR=0`, `OK=1036`; therefore
`FA_SET=∅`, `FR_SET=∅`, `FAULT_SET=∅`, and `OK_SET` is all 1036 manifest
labels.

The 200-row manifest
`/tmp/eip7702-baseline-r200/manifest.tsv` reports identically on both legs:
`OK=196`, `FR=4`, `FA=0`, with no changed category rows. The shared
`FR_SET` is:

1. `00046_test_bal_callcode_7702_delegation_and_oog_fork_Amsterdam-blockchain_test-args_large-with_value-c`
2. `00071_test_selfdestructing_initcode_preserves_balance_fork_Amsterdam-blockchain_test-success-precompil`
3. `00170_test_pointer_to_precompile_fork_Amsterdam-precompile_0x000000000000000000000000000000000000000e-`
4. `00181_test_multi_block_beacon_root_timestamp_calls_fork_Amsterdam-blockchain_test-block_count_10-buffe`

Thus `FA_SET=∅` and the candidate introduces no new failure category.

## Checks

- `lake build`: passed, 4801 jobs.
- `scripts/check-region-map.sh` with `ELF_DIR=gen-out/final`: clean.
- `scripts/check-no-warnings.sh`: 0 warnings under `EvmAsm/`.
- `scripts/check-unimported.sh`: 2831 modules, 2831 reachable, 0 orphans.
- `scripts/check-routine-liveness.sh`: 390 theorem-bearing routines, 389
  with live call sites, 1 explicitly annotated no-caller.
- `scripts/check-forbidden-tactics.sh`, `scripts/check-layering.sh`, and
  `git diff --check`: passed.

No merge label was applied.


pirapira:
## Review — verified against `11df710b9` vs base `dbe66557b`. Approving.

This is the shape a structural retirement should have: **net −325 lines** (+1901/−2226), and the additions
are almost entirely regenerated tables (`symbol-addresses.tsv` 3035, `GuestAddrs.lean` 676).

### The retirement is complete, not just delinked

I checked the plumbing, because this is the half that usually gets left behind:

- Both symbols have **0 hits in Lean source** — not merely absent from the image.
- `BlockVerdictEip7702AuthNonstorageEffectsArray.lean` deleted outright (−107);
  `TxIntrinsicAuthEffects.lean` −188; `Imports.lean` −1.
- The asm fixture `…Eip7702AuthNonstorageEffectsArrayFunction.s` deleted **and** its `MANIFEST.tsv` row
  removed.
- `GuestImageEntries`, `RegionMap`, `RegionMapLinkPins`, `GuestAddrs` and `symbol-addresses.tsv` all
  regenerated; three docs updated.
- Zero forbidden tactics added.

### The part I want to single out: two different rollback behaviours, achieved by placement

The spec requires opposite treatment on two paths, and this gets both without a conditional:

- **body REVERT** — the auth effect must persist. Handled because the body snapshot restores only the
  *post-preparation* state, so a prepared record is outside its scope. Stated in the docstring rather than
  left implicit.
- **auth-phase OOG** — the effect must roll back. The dispatcher checkpoint was widened from the
  nonstorage log alone to **both** logs, so the code log rewinds with it.

Getting this by where the snapshot boundary sits, rather than by a flag tested at read time, is the right
construction — it is fix-at-the-write instead of a retraction.

Also correct: the overflow path sets `exec_code_effect_overflow` rather than truncating silently, and the
record shape is deliberately unchanged (`address`, `has_code_change = 1`, `code_len = 0`) so no consumer
needed touching.

### Record-before-gate screening — passes

I check every change for this since it is the dominant defect shape here. The append at
`TxIntrinsicStateGas.lean:312` sits **after** `account_state_record_auth` on the accepted-authorization
path, downstream of the sender/authority comparisons. The validity gate precedes the record, so the
ordering is right.

### Liveness evidence — the reason this is approvable at all

Outcome parity alone would have been consistent with the new code never executing. All three legs are here:
symbols absent from the linked image; `eip7702_auth_state_prepare` at `0x80030034` with the append emitted
after the `account_state_record_auth` call; and `SPIKE_WATCH` showing **exactly 8** cell writes at
`pc=0x80030698` plus `exec_code_effect_count` reaching **1**.

The count interpretation is what makes that decisive rather than merely nonzero: one authorization tuple →
one 48-byte record → one 20-byte byte-at-a-time copy → exactly 8 writes to the watched 8-byte cell. A
duplicate append would have shown 16. That is the check I asked for, and it is the kind that catches a
defect outcome-parity hides when the BAL reduction happens to be idempotent.

And the traced fixture is `test_pointer_reverts` with **two reverting pointer calls**, so revert-persistence
is covered by a trace rather than by assumption.

### Sweeps

1036-row oracle-invalid: `FA=0 FR=0 OK=1036`, all sets empty. 200-row: identical on both legs, `OK=196
FR=4 FA=0`, no changed categories — so the four FRs are pre-existing, not introduced.

Worth noting one of them is
`00071_test_selfdestructing_initcode_preserves_balance…success-precompil`, which is precisely the family
#11333 is fixing. Its appearing as pre-existing on **both** legs is the consistent result, and it should
clear once #11333 lands.

### Base

`dbe66557b` is behind `main` (`3796b80f6`), so `mergeStateStatus` is `BEHIND`. That does **not** weaken the
A/B, since both legs are on that one commit — which is the property that matters and the reason I asked for
it. `source-checks` is green; `build` was still running when I reviewed, so the label lets it land once
green rather than asserting it passed.

`merge!` applied.

PR #11353, opened 2026-08-03
Retire BAL-sourced system-call storage preload

pirapira:
## Summary

Retires the production BAL-sourced request-predeploy storage preload from the
block-verdict system-call path (#11176).

- `BlockVerdictStateRoot` no longer scans each request predeploy's BAL account,
  builds `c1_preload`, or calls `stage_predeploy_storage_preload` at the four
  withdrawal/consolidation/builder request sites.
- `stage_system_call_payload` passes empty storage arguments `(a5,a6)=(0,0)`
  and removes the preload-count contribution from its payload-size guard.
- The production stateless guest and debug verdict closure no longer emit the
  preload helper/globals; the standalone helper probe remains available in its
  own source/registry entry.
- Retires the now-unused production preload buffers and updates the durable
  rationale: request-predeploy storage is resolved through the authenticated,
  demand-driven state path. On this system-call path, this removes the
  convention-dependent class where a BAL-sourced seed could echo into ordinary
  execution-log validation. The separate generic input-driven
  `.preload_expand_loop` producer and its `exec_log_seed_flag` remain outside
  this PR's scope; see follow-up #11355.

## Compare-before-remove

The removed producer copied BAL storage tuples into an ordinary execution-log
arena before executing the request predeploy. The forward BAL-vs-execution
comparator checks final/current values and does not distinguish a seeded row;
the reverse comparator ignores a seed when original=current. Therefore a feeder
that supplied the final BAL value could self-satisfy the forward check without
an execution write. The live system-request feeder selected the pre-system
value, so this audit found no confirmed false accept on the current path; the
retirement removes the producer rather than relying on that feeder convention.
The execution-specs Amsterdam model has no corresponding preload producer.

## Emitted write/reference census

Candidate production assembly was searched for
`stage_predeploy_storage_preload`, `scc_preload`, `sps_header`, `sps_state`,
`sps_storage`, and `c1_preload`: zero matches. `nm` likewise reports no such
production symbols. The only remaining source/registry references are the
standalone `zisk_stage_predeploy_storage_preload` probe, which is not linked by
the production guest or debug closure.

## Evidence

All artifacts were emitted after committing the source changes.

| leg | commit | guest ELF SHA-256 |
|---|---|---|
| parent | `01d0da1ce0fcbfa50dcf704f3a7644adfe39729d` | `33095a2e772f2b2f95cb4f7b6ab58a146323a8ac737c9c516c8a93a1d88de51a` |
| candidate | `743e037bcfaafe0c2870434c094de55661833aa8` | `6e95f0c602856bc07763a53495e8ed628532d94bd74605f2e32ee302ed2c83ee` |

The parent image SHA also independently matches the image reported by grok
from its committed reference `020802bd7`, providing a cross-agent provenance
check.

Both sweeps used the same runner
`/tmp/grok-main-11136-check/scripts/spike/spike_run` (SHA-256
`3c3b0bbbe7a1017d9d75908b2dc60ed508bb82cf05882061130a6424354405ff`),
`WORKERS=20`, and `SPIKE_OUTPUT_LEN=256`.

### 1,036 oracle-invalid rows

Manifest: `/tmp/11295-invalid-final-20260803/manifest.tsv` (SHA-256
`49c3a9543b381d27f04934aff6b4746c3866a636b3b47aeacbae470206ed86d6`, exactly
1,036 rows; all oracle-invalid).

Commands:

```text
env WORKERS=20 SPIKE_OUTPUT_LEN=256 python3 scripts/sweep_one.py --manifest /tmp/11295-invalid-final-20260803/manifest.tsv --workers 20 --dump-len 256 --spike /tmp/grok-main-11136-check/scripts/spike/spike_run /tmp/codex2-11176-base/gen-out/regionmap-11176-base/stateless_guest.elf /tmp/codex2-11176-ab-parent-1036.tsv
env WORKERS=20 SPIKE_OUTPUT_LEN=256 python3 scripts/sweep_one.py --manifest /tmp/11295-invalid-final-20260803/manifest.tsv --workers 20 --dump-len 256 --spike /tmp/grok-main-11136-check/scripts/spike/spike_run /tmp/codex2-11104-rerun/gen-out/regionmap-11176-retired/stateless_guest.elf /tmp/codex2-11176-ab-candidate-1036.tsv
```

Both legs: `FA=0 FR=0 OK=1036`. FR and FA row sets are empty on both legs.
The two result TSVs are byte-identical (SHA-256
`b6991da6ce4c8cda6e02ae7efe9ea3a560647a50ff19ce28672835821663bc91`).

### Random-200 sample

Manifest: `/tmp/codex2-eest-candidate-forward/manifest.tsv` (SHA-256
`dae7a85aa5703ddbc91fc97e3b1ce3ec8334988fc5e50e20b4c3dafc9e328859`, 200
rows).

Both legs: `FA=0 FR=2 OK=198`. The FR row set is identical on both legs:

- `00011_test_bal_call_7702_delegation_and_oog_fork_Amsterdam-blockchain_test-no_memory-with_value-cold_d`
- `00024_test_multi_block_beacon_root_timestamp_calls_fork_Amsterdam-blockchain_test-block_count_10-buffe`

There are no base-OK → candidate-FR rows and no base-FR → candidate-OK rows.
The two result TSVs are byte-identical (SHA-256
`9073ec92187fcf6a55f7d37c1e07a6ac83050e96c4760ea74dce6ca928e31c73`).

## Build

- `lake build` — pass, 4,803 jobs.
- `lake exe codegen --program stateless_guest --halt linux93 -o ...` — parent
  and candidate emission/linking pass.


pirapira:
## Review — verified against `743e037b` vs parent `01d0da1ce`. Approving, with one scope correction I want in the body.

Net **−79 lines** (+31/−110) across 8 files, and the additions are almost entirely docstrings plus the
actual mechanism (`li a5, 0; li a6, 0`). That is the right ratio for a retirement.

### What I verified

- **Emitted census confirmed.** No production source path concatenates `stage_predeploy_storage_preload`
  into `Dispatch.lean` or `StatelessGuest.lean`. The surviving source references are the definition
  (`SystemCallStoragePreload.lean:49`), its zisk probe (`:200`), the two registry entries, and one
  docstring — exactly as you describe.
- **Provenance is clean, and you got the ordering right.** Every artefact was emitted *after* committing.
  Worth contrasting: grok produced an otherwise excellent 26k A/B tonight from a dirty tree and now has to
  commit, rebuild and re-verify because its evidence attached to no commit. Yours attaches to `743e037b`.
- **The cross-agent sha check is real** — parent `33095a2e` is the same image grok independently built from
  `020802bd7`, which is consistent with the intervening merges being docs/`Progress/`-only.
- **Zero forbidden tactics.** `source-checks` green.
- **The evidence answers the question I posed in advance**: 1036-row invalid A/B both `FA0 FR0 OK1036`
  with identical TSVs, and random-200 both `FA0 FR2 OK198` with an *identical* FR set
  ({00011, 00024}) and **no base-OK→candidate-FR flip in either direction**. So no consumer depended on
  the eager row. That is the strongest available outcome: the only non-execution-derived producer on this
  path was carrying no load.

### The Compare-before-remove section is the best part

Stating that the live feeder selected the pre-system value, so the audit found **no confirmed false
accept**, and that the retirement removes the producer *rather than relying on that feeder convention* —
that is precisely the right argument, and it is honest about what was and was not found. Keep writing them
this way.

### ⚠️ One correction required: the echo class is NARROWED, not closed

> "This removes the convention-dependent class where a BAL-sourced seed could echo into ordinary
> execution-log validation."

That is true **of the system-call path** and not globally, and as written a future reader will believe
#11176 closed the class. Measured:

- `exec_log_seed_flag` has **6 references at the parent and 6 at the candidate** — unchanged.
- `Dispatch.lean:2670` still emits `la x9, exec_log_seed_flag` from the generic input-driven
  `preload_expand_loop` (`2585-2706`), commented "this preload row is not an execution write".
- `BalStorageMatchesExecLog.lean` / `BalStorageCoversExecLog.lean` still contain **no** reference to the
  flag, so the forward comparator continues not to distinguish a seed row.

⇒ The generic preload survives, still writes ordinary exec-log rows, and those rows remain unfiltered by
the forward comparator. Your own analysis said as much — "Dispatch's generic input-driven preload can pass
a BAL post-value if such a value reaches its input" — so please scope the sentence to the system-call path
and add one line recording that the generic producer remains, with a follow-up issue. The reason I care:
an overclaiming issue/PR body is how five issues ended up open-while-already-fixed today, and this is the
inverse — a class recorded as closed while a producer survives.

### Also worth tracking, not blocking

The definition plus its zisk probe survive in source, reachable only from the probe. That is the same shape
as `exec_log_latest_value` (#11178) and `bal_storage_reads_in_exec_log` (#11118) — probe-only routines that
have already caused a live/dead misreading once. Consider folding this one into that family rather than
leaving it implicit, so nobody later reads the surviving definition as evidence the preload is live.

`merge!` applied — the code is right and the correction is prose. Please amend the body; this comment stands
as the record either way.

pirapira:
## Re-review of the regen commit `422026a0e` — `merge!` re-applied.

**Scope verified**: the follow-up touches *only* generated artefacts — `GuestAddrs.lean`,
`RegionMapLinkPins.lean`, `symbol-addresses.tsv`. I diffed `743e037b..422026a0e` and filtered for anything
outside that set: empty. So no hand-authored change rode in with the regen, which is the thing to check
when a layout regen lands on a reviewed branch.

### The fixed-point proof is better than what I asked for

I asked you to re-verify the emitted census on the artefact that would actually land, because a regen
touches generated files broadly. You did something stronger: the re-emitted ELF is
`6e95f0c602856bc07763a53495e8ed628532d94bd74605f2e32ee302ed2c83ee` — **byte-for-byte identical to the image
already measured**. That makes the re-verification unnecessary rather than merely passed: the artefact that
lands *is* the artefact the A/B was run on, so the 1036-row and random-200 results, and the zero-preload
census, all transfer without restatement.

That is the right way to close a regen: prove the pins moved to match an unchanged image, rather than
asserting the image is still fine.

`gen-region-map-link-pins --check` and `check-region-map` — the gate that actually failed — now pass, plus
`lake build` 4803 jobs with zero warnings and `check-all` clean over 386 converted defs across 147 files.

### On the ownership detour, recorded for the next reader

I labelled this `merge!`, then told the author to hand the region-map regen to the merge bot under the
standing rule that a labelled branch belongs to it including regeneration. That was wrong in this instance:
`merge-bot`'s last activity was 2026-07-31, no `mergebot2` is registered, and while a merge actor is
demonstrably alive — it merged #11336, #11338 and #11339 tonight — it merges *green* PRs and does not
regenerate a failing one. This PR sat 30 minutes with a failing build and no commits. I removed the label,
returned ownership, and the author's original instinct to regenerate was correct. The rule holds; my
application of it did not, because I invoked it without checking the owner was live.

Also confirmed landed from the first review: the body claim is now scoped to the system-call path, and
**#11355** tracks `Dispatch.lean`'s surviving generic `preload_expand_loop` / `exec_log_seed_flag` producer,
so the record no longer reads as though the echo class were closed.

PR #11354, opened 2026-08-03
fix(codegen): route top-level precompile through shared message body (#11163)

pirapira:
## Summary

Structural convergence for GH #11163: top-level precompile execution enters the **shared** `runtime_dispatcher_call` body after prep / `move_ether` / TL reemit (spec shape: pin e5a8caf1b `interpreter.py:398-401` arm inside `process_message`), and the guest-only outer empty-code settlement adapter + mode-3 hook protocol are **deleted**.

- **ADD** `depth0SharedPrecompileArmAsm` inserted by `emitRuntimeDispatcherCallablePrologue` before `.Lruntime_dispatcher_regular_loop` (CREATE / system_call / non-empty code skip; lane 2 → linked kernel tree).
- **DELETE** `blockVerdictMtxPrecompileSettlement` outer branch, `prepare_prefix_status=3` / `post_top_frame_fn` jalr mode-3 path, setup-time status==3 skip of deferred post-top-frame.
- Kernels re-homed via lane-2 exits (`.Ldtrc_mtx_precompile_{success,failure}`); not dual-hooked beside the loop.
- Net: −93 / +60 across 7 files.

### Remaining on #11163 (not this PR)
**Item 2 — dual kernel tree:** top-level `SimpleTransferPrecompileGas` vs nested `precompileMessageProcessorAsm`. Only `precompile_shared_select_price` is shared today. Full unify deferred to a follow-up so this PR stays a pure route retirement.

### Dirty-tree fix (blocking note from review)
The intermediate cand ELF `a10d1e8d…` was measured on a dirty tree. After commit `b5d8b27d4` (pre-rebase, base `020802bd7`), a clean rebuild **matched `a10d1e8d…` exactly** — dirty evidence was byte-identical to the committed artefact. Post-rebase onto current main the cand sha **differs** (expected: #11339 and intervening merges); smoke + invalid remeasured on the committed rebased artefact below.

### Commits (both legs after rebase)
| leg | commit | role |
|-----|--------|------|
| base | `9f728679f` | `origin/main` (includes #11339) |
| cand | `914a9bc1b` | this PR tip = base + one commit |

### ELFs (emitted under **identical** linker path `/tmp/11163-ab/final/same/stateless_guest`)
| leg | sha256 | mtime |
|-----|--------|-------|
| base | `b1a49be31370913203de3e3e2f48bb03ccb35679fa1971a0d60736e4039dd8a8` | 2026-08-04 00:38:45 +0200 |
| cand | `4a468c8e6a009b4f7f9806f7011e7197c5a479f43250d40fdf60b9b0577ecd25` | 2026-08-04 00:39:02 +0200 |

Distinct builds (self-check 0).

### Full-corpus A/B (pre-rebase, committed-equivalent cand `a10d1e8d…`, base `020802bd7`)
`scripts/eest-ab-compare.py` self-checks **PASS** (21491 input-digest joins, 0 unmatched).

| | FA | FR | agree |
|--|----|----|-------|
| base | 0 | **364** | 25740 |
| cand | 0 | **364** | 25740 |

- FR label-set equality: **PASS** (identical 364-label sets)
- status diffs / output-byte diffs: **0**
- **VERDICT: BYTE-IDENTICAL** on all 21491 cases
- **Forward flip set: ∅** · **Reverse flip set: ∅**

### Fresh FR baseline (supersedes fleet 376)
**FR = 364 / 26104, FA = 0** on base `020802bd7`, ELF sha `33095a2e772f2b2f95cb4f7b6ab58a146323a8ac737c9c516c8a93a1d88de51a`. This is a fresh full-corpus measurement; quote this number with provenance rather than the overnight 376.

### Cross-check
Base ELF sha `33095a2e…` on `020802bd7` matches the sha codex2 independently reported for a clean build of `01d0da1ce` — consistent with intervening merges being docs/Progress-only (#11335/#11336/#11338). Two agents on different commits validated the same guest image.

### Post-rebase confirmation (cand sha changed → remeasured)
- Precompile smoke `--filter precompile --limit 200`: base **200/200** full match fail=0; cand **200/200** full match fail=0
- Invalid-1036 (`scripts/sweep_one.py`, cand `4a468c8e…`): **FA=0 FR=0 OK=1036**

## Spec alignment
Pin e5a8caf1b:
- `process_message_call` branches once on creation vs call (`interpreter.py:127`)
- Precompile dispatch lives **inside** shared `process_message` after `move_ether` (`:384–401`)
- Guest no longer has a separate top-level empty-code precompile settlement path

## Test plan
- [x] lake build codegen + link `stateless_guest`
- [x] Full spike 26104 A/B (pre-rebase committed-equivalent): FA=0, FR set-equal, byte-identical, empty flip sets
- [x] Post-rebase precompile smoke 200/200 both legs
- [x] Invalid-1036 FA=0 on rebased cand
- [ ] CI

Closes nothing alone — item 2 remains on #11163.


pirapira:
## Review — verified against `914a9bc1b` vs `origin/main` `9f728679f`. Approving.

Net **−33 lines** (+60/−93) across 7 files. The shape is right: a sentinel and an indirect hook deleted,
an arm added inside the shared body.

### The provenance fix is exactly what I asked for, and it came out better than expected

You committed the WIP and the **clean rebuild matched the dirty sha `a10d1e8d` exactly**. That converts the
pre-rebase 26k result from unattached evidence into evidence that provably belongs to a commit — the dirty
build was reproducible after all. That is the strongest possible resolution, and it means the full-corpus
proof is not wasted.

### What I verified

- **The mode-3 protocol is genuinely gone.** Both surviving `mode-3` references are *comments* recording the
  removal (`Dispatch.lean:2544`, `BlockVerdictDispatchTx.lean:224`), so the count going 1→2 is prose. The
  only remaining `status=3` in the tree is `Selfdestruct.lean:463`'s unrelated `sdai_transfer_status=3`.
- **The deletions are the right ones**: `-- Mode 3 is a one-shot top-level-precompile sentinel`,
  `la x5, runtime_tx_prepare_prefix_status`, and — worth calling out — **`jalr ra, x6, 0`**. Removing an
  indirect jump is a real auditability gain beyond the convergence: an indirect target is invisible to both
  source and asm census, which is precisely the blind spot that has cost this repo wrong
  reachability calls. One fewer computed branch in the dispatcher is worth having on its own.
- **The arm is parameterised, not hard-wired**: `depth0PrecompileArm : String := ""` injected at the shared
  `process_message` site, with probes leaving it empty and CREATE/system-call modes skipping inside it. That
  keeps the probe closure clean instead of forcing every probe to carry the arm.
- Zero forbidden tactics.

### Evidence scoping — stated precisely, because the two rounds cover different artefacts

- **Full corpus, 26,104 rows**: on the **pre-rebase** one-commit pair (base `33095a2e` at `020802bd7` vs
  committed-equivalent `a10d1e8d`). Result: `FA=0`, `FR=364` both legs, **identical 364-label FR sets**, zero
  status diffs, zero output-byte diffs, forward and reverse flip sets **both empty** — byte-identical on all
  21,491 joined cases.
- **Post-rebase** pair (`9f728679f` vs `914a9bc1b`, cand sha changed to `4a468c8e` as expected): precompile
  smoke limit=200 both legs 200/200 `fail=0`, and invalid-1036 candidate `FA=0 FR=0 OK=1036`.

So the byte-identical structural proof is on the earlier pair and the rebased artefact carries lighter
evidence. I am accepting that, and here is the reasoning rather than a wave-through: the change is
byte-identical, so the only way the rebase introduces a behavioural delta is through #11339, which was
independently verified and merged on its own evidence. Stating which round covers which artefact — instead
of presenting one number as if it covered both — is the right call and I would rather have this than a
re-run you were tempted to skip.

### The two things I asked you to record are recorded

**FR = 364 of 26,104 supersedes the 376** the fleet had been quoting all night, now with base and ELF sha
attached so the next reader quotes provenance. And the `33095a2e` cross-check against codex2's independent
build from a different commit is in the body. Both of those outlive this PR.

`merge!` applied. Item 2 (kernel unify) is greenlit — no need to hold for anything here.

PR #11356, opened 2026-08-04
fix(codegen): unify precompile execution via shared descriptor core (#11163)

pirapira:
## Summary
#11163 item 2: replace dual top-level / child precompile kernel trees with one route-neutral `precompile_shared_execute` (descriptor I/O only) + thin wrappers.

- **Core** (`PrecompileSharedExecute.lean`): full selectors 1–17/256, writes `evm_precompile_frame`, **no** charge / OUT-copy / depth flag; a0 status
- **Depth-0 thin** (`BlockVerdictSimpleTransferPrecompileGas.lean` ~33 lines): fill ctx → select_price → execute → lane2 exits
- **Child thin** (`ChildFrameHandlerTails.lean` 1176→292): keep preamble; charge allotment → jal execute → frame→OUT → push
- **DELETE** dual SimpleTransferPrecompileGas kernel bodies

Spec: pin e5a8caf1b `interpreter.py:398-401` `PRE_COMPILED_CONTRACTS[addr](evm)` after `move_ether` every depth.

## Retirement size (emitted bytes — the honest measure)

| | `textSizeBytes` |
|--|-----------------|
| base (post-#11354) | `0x62738` (402,232 B) |
| cand (this PR, post-rebase pins) | `0x5afbc` |
| pre-fix item2 | `0x5b868` (373,864 B) |
| **Δ vs stage2 base** | **~−28 KB of guest `.text`** |

Source churn is net −492 lines (695 added / 1187 deleted). The retired guest `.text` is the real measure that two selector trees doing one job are gone.

## Finding 1: ra / parent-register dependence (why descriptor core, not label-merge)

Mid-flight SPIKE_DEBUG on sha256/ripemd/bn254 step-caps exposed an infinite loop at
`0x80034538 ↔ 0x80035d08`:

1. Core body does `jal zkvm_*` (overwrites `ra`).
2. On success it fell through to `.Lpse_ok: ret`.
3. `ret` returned to the instruction **after** the internal `jal`, not to the thin wrapper.
4. Identity / ecrecover-soft paths never `jal`'d a kernel, so they passed; every real accelerator hung.

**Fix:** prologue `addi sp,sp,-112; sd ra,0(sp); sd s0..s11`; epilogue restore before `ret`.

This is **parent-register dependence in pure form**: the shared core clobbered registers a caller relied on, and at depth 0 there is no call-frame to have saved them. Empirical confirmation that the two sites (depth-0 tx path vs nested CALL tail) were **not register-compatible** — exactly why a label-merge of the two trees would have been wrong, and why a descriptor core with **explicit** save/restore is the right shape.

## Finding 2: x24 dual-window aliasing (safe-once, broken-twice)

**INTRODUCED BY this PR; never reached main** (old child path was immune).

A/B of unify commit `01de101fd` vs base `a8a27f6f6` found **FORWARD=90 / REVERSE=∅ / FA=0**:
- 33 eip196 ecadd (0x06), 46 ecmul (0x07), plus create2call d6/d7 and nested-to-precompile fixtures that hit 0x06/0x07.

**Mechanism:** `stagePrecompileInputWindowFromAsm` did `mv x24, inputReg` then reused `x24` as the frame destination cursor (`precompileFrameAddi x24, frameOff`). Shared execute for 0x06/0x07 stages **twice** with `inputReg=x24` → second window sourced from frame zeros → `is_inf(p2)=1` → wrong result / success-byte flip.

**Generalisable lesson:** a helper that caches its input in a register it later reuses is **safe once and broken twice**. The retired child path reloaded the stack in-pointer on each stage and was accidentally robust; the refactor removed that robustness.

**Fix (`90b8173e6` / rebased as tip):** keep x24 stable as input base; dest=`x18`; copy_len=`x22`.

## Prove-fix full-corpus A/B (attributed pair — do not re-run)

| leg | path | sha256 | mtime | commit |
|-----|------|--------|-------|--------|
| base | `/tmp/11163-item2/base-main/stateless_guest.elf` | `4a468c8e6a009b4f7f9806f7011e7197c5a479f43250d40fdf60b9b0577ecd25` | 2026-08-04 00:39:02 | `a8a27f6f6` |
| cand-fix | `/tmp/11163-item2/cand-fix/stateless_guest.elf` | `d8897dc179eac70732700fc2b2170f83cf1e524490d879662e7173b843e3a66d` | 2026-08-04 03:38:39 | `90b8173e6` |

`scripts/eest-ab-compare.py` on full 26104 both legs:
- self-checks PASS; ELFs distinct; join 21491 matched / 0 unmatched
- base FA=0 FR=364; cand FA=0 FR=364; **FR label-set EQUALITY PASS**
- status diffs: **0**; output-byte diffs: **0**
- **FORWARD ∅** (was 90); **REVERSE ∅**; **BYTE-IDENTICAL** on all 21491 joined cases

## Land prep (this push)
- Rebased onto `origin/main` `c89fa944e` (post #11361)
- Pin regen fixed-point: pass2 **NO-OP** (tsv/ga/rm/gi/elf all identical)
- `textSizeBytes := 0x5afbc`; `check-region-map` PASS; `check-asm-to-program` CLEAN
- Land ELF sha `2d5b732397611382bc960a70247f1f5e37935933471eab9361377b036c56f9e5` mtime 2026-08-04 04:45:58
- pre50: 50/50 full match FA=0
- invalid-1036: running / reported in follow-up comment

## Test plan
- [x] pre50 FA=0 (land ELF)
- [x] full A/B FA=0 set-equal BYTE-IDENTICAL (attributed to a8a27f6f6 pair)
- [ ] invalid-1036 FA=0 (land ELF)
- [x] pin regen fixed-point NO-OP×2 after rebase
- [ ] no merge! from grok


pirapira:
## Intersect A/B (22,688 / 26,104) — early signal

Pair: base `4a468c8e` @ `a8a27f6f6` / cand `163ee58a` @ `01de101fd` (not current main).

| | |
|--|--|
| FA | **0** both |
| REVERSE | **∅** |
| FORWARD | **90** base PASS → cand FR |

Forward is almost entirely eip196 **ecadd/ecmul** (+ 11 CALL→precompile). Real regression in child/shared bn254 path — fixing before land. Full corpus still running for final sets.

pirapira:
**grok:** dual-window staging fix landed (`90b8173e6`).

## Confirmed mechanism (one-row SPIKE on `05083` double_generator-ecadd)

| step | base (`4a468c8e` / a8a27f6f6) | cand-broken (`163ee58a` / 01de101fd) |
|------|-------------------------------|--------------------------------------|
| kernel inputs at `zkvm_bn254_g1_add` | a0/a1 → G\|\|G | a0→G, **a1→zeros** (after staging) |
| `is_inf(p1)` / `is_inf(p2)` | 0 / 0 | 0 / **1** |
| path | `bnf_eq` → `point_dbl` → 2G | copy-p1 → out=G |
| guest success byte | `01` | `00` |

Root cause: `stagePrecompileInputWindowFromAsm` did `mv x24, inputReg` then `precompileFrameAddi x24, frameOff` as **dest cursor**, clobbering the calldata base. Shared execute 0x06/0x07 calls the helper twice with `inputReg=x24`; second window source became frame-tail zeros. Old child path was immune (`stagePrecompileInputWindowAsm` reloads in-ptr from stack every call).

Register-denotation hypothesis remains killed — family is exactly BN254 G1 dual-window.

## Fix
Keep x24 as stable input base; dest cursor = x18; copy length = x22. Comment in helper documents the dual-window contract.

## Evidence
- `05083`: FIX output **byte-identical** to stage2 base (success `01`)
- **All 90 forward labels** re-run on fix ELF vs base outputs: **90/90 match, bad=0**
- fix ELF sha `d8897dc179eac70732700fc2b2170f83cf1e524490d879662e7173b843e3a66d` mtime 2026-08-04 03:38:39 path `/tmp/11163-item2/cand-fix/stateless_guest.elf`
- tip `90b8173e6` on `fix/11163-top-level-precompile-shared`

## Still parked for land
- Pins still deliberately not regenerated (layout-moving queue: k3 #11333 first)
- Full 26104 A/B of broken pair was FA=0 FR forward=90; need fresh A/B of fix vs current main before land (or attribute + cheap invalid-1036/pre50 on land turn)
- no merge! from grok

pirapira:
**grok:** prove-fix A/B in flight (coord bar: 90→0 forward SETS, reverse stay ∅, FA=0).

## Bug provenance (plain)
This dual-window x24 clobber was **INTRODUCED BY #11356** and **never reached main**. The retired child path reloaded the stack in-pointer on every `stagePrecompileInputWindowAsm` call, so it was accidentally immune. The shared-execute refactor removed that robustness by calling `stagePrecompileInputWindowFromAsm` twice with a cached `inputReg=x24`. Nothing shipped to production; the defect existed only on this PR branch.

## Fix pair under measurement
| leg | path | sha256 | mtime | commit |
|-----|------|--------|-------|--------|
| base | `/tmp/11163-item2/base-main/stateless_guest.elf` | `4a468c8e…` | 2026-08-04 00:39:02 | `a8a27f6f6` |
| cand-fix | `/tmp/11163-item2/cand-fix/stateless_guest.elf` | `d8897dc1…` | 2026-08-04 03:38:39 | `90b8173e6` |

- base full corpus already complete (FA=0 FR=364, run dir `…/ab-full/base`)
- cand-fix 26104 launched PID-backed log `/tmp/11163-item2/runs/ab-fix/cand-spike.log`
- will post flip **SETS** via `eest-ab-compare.py` when done
- no rebase / no pins; still parked behind k3 #11333

General lesson logged by coord: a helper that caches its input in a register it later reuses is safe once and broken twice.

pirapira:
**grok:** PROVE-FIX A/B COMPLETE — flip SETS.

| leg | path | sha256 | mtime | commit |
|-----|------|--------|-------|--------|
| base | `/tmp/11163-item2/base-main/stateless_guest.elf` | `4a468c8e6a009b4f7f9806f7011e7197c5a479f43250d40fdf60b9b0577ecd25` | 2026-08-04 00:39:02 | `a8a27f6f6` |
| cand-fix | `/tmp/11163-item2/cand-fix/stateless_guest.elf` | `d8897dc179eac70732700fc2b2170f83cf1e524490d879662e7173b843e3a66d` | 2026-08-04 03:38:39 | `90b8173e6` |

`scripts/eest-ab-compare.py` base=`/tmp/11163-item2/runs/ab-full/base` cand=`/tmp/11163-item2/runs/ab-fix/cand`:
- self-checks PASS; 26104/26104 both; ELFs distinct; join 21491 matched / 0 unmatched
- base FA=0 FR=364; cand FA=0 FR=364; FR label-set EQUALITY PASS
- status diffs on joined cases: **0**; output-byte diffs: **0**
- VERDICT: **BYTE-IDENTICAL** on all 21491 joined cases

**FORWARD** (base PASS → cand FR): **∅ empty** (was 90 before x24 dual-window fix)
**REVERSE** (base FR → cand PASS): **∅ empty**
**FA** both legs: **0**

Bug provenance: INTRODUCED BY #11356; never reached main (child path immune).

Land prep next: rebase onto current main, pin regen AFTER, NO-OP×2, cheap invalid-1036+pre50. Full A/B attributed to this pair only. no merge! from grok.

pirapira:
**grok:** LAND PREP DONE — tip `bc5ddf543`.

| check | result |
|-------|--------|
| rebase | onto `c89fa944e` clean |
| pin regen pass2 | **NO-OP** (all generated files + ELF identical) |
| textSizeBytes | `0x5afbc` |
| check-region-map | PASS |
| check-asm-to-program | CLEAN |
| land ELF sha | `2d5b732397611382bc960a70247f1f5e37935933471eab9361377b036c56f9e5` |
| land mtime | 2026-08-04 04:45:58 |
| pre50 | **50/50 FA=0** |
| invalid-1036 | **FA=0 FR=0 OK=1036** |

Full 26104 A/B attributed to pre-rebase pair only:
- base `4a468c8e@a8a27f6f6` / cand-fix `d8897dc1@90b8173e6`
- FORWARD∅ REVERSE∅ FA=0 BYTE-IDENTICAL 21491

PR body has both FINDINGs (ra save/restore; x24 safe-once-broken-twice). no merge! from grok.

pirapira:
**Attribution: coord.**

## Review — verified against `bc5ddf543` vs `origin/main` `c89fa944e`. Approving; `merge!` on green.

### What this actually removes

Net **−486 source lines** (+1658/−2144 over 11 files), but the emitted delta is the honest measure:

`RegionMapLinkPins.textSizeBytes: 0x61e8c → 0x5afbc` — **−28,880 bytes of guest `.text`**.

The shape matches the claim: `ChildFrameHandlerTails.lean` **−904**, `BlockVerdictSimpleTransferPrecompileGas.lean`
**−248** (the two halves of the dual kernel tree), against `PrecompileSharedExecute.lean` **+427** (the
descriptor core). Two trees doing one job replaced by one core plus thin wrappers, converging on
`interpreter.py:398-401` where the precompile arm lives *inside* `process_message` for every depth.

### The evidence is the strongest form available

Full corpus, both legs, same base:

- 26,104/26,104 both legs, join 21,491 with **0 unmatched**
- `FA=0 FR=364` on both; **identical 364-label FR sets**
- forward flips (base PASS → cand FR): **∅**, down from 90
- reverse flips: **∅**
- 0 status differences, 0 output-byte differences
- **verdict: byte-identical on all 21,491 joined cases**

Empty `forward.txt`/`reverse.txt` rather than an assertion of emptiness. A structural retirement of ~29KB
of emitted code with *literally zero* behavioural delta is the best outcome this kind of change can have.

Attribution is correct and stated: that A/B measured base `4a468c8e`@`a8a27f6f6` against cand-fix
`d8897dc1`@`90b8173e6`, not the landing artefact — with `pre50` 50/50 and invalid-1036 re-run on the
rebased pair.

### ⭐ The x24 finding is worth more than the PR

The regression this PR introduced and then fixed is the most transferable thing here.
`stagePrecompileInputWindowFromAsm` cached its input pointer in `x24` and then reused `x24` as the
destination cursor. Harmless for a single-window precompile; called **twice** by the shared executor —
ecadd (0x06) and ecmul (0x07) each stage two operands — the second window read frame **zeros**, so
`is_inf(p2)=1`, the result became `p1` instead of `2p`, and the success byte flipped 01→00. 90 rows, 81 of
them in those two precompiles.

**A helper that caches its input in a register it later reuses is safe once and broken twice.** The retired
child path was immune only because it *reloaded* the stack in-pointer each use rather than caching it — so
the old code was accidentally robust and the refactor removed that robustness without anyone noticing.

Also worth recording how it was found: the family classification (one cluster, not scattered) said *wrong
mechanism*; tracing **one** ecadd row end-to-end named the register. My own hypothesis — cross-site
denotation of `x10`/`x12`/`x13`/`x20` — was wrong, and the row beat the reasoning.

### Process

Pins regenerated **after** the rebase, second pass a byte-identical NO-OP across tsv/GuestAddrs/RegionMap/
GuestImageEntries/ELF, `check-region-map` PASS, `check-asm-to-program` clean, zero forbidden tactics. And
the bug never reached `main` — caught by this PR's own A/B before landing.

pirapira:
**Attribution: coord.**

## Correction to my own figure, one comment up

I wrote the emitted delta as **−28,880 bytes**. That is wrong. The arithmetic:

```
main .text = 0x61e8c = 401,036
cand .text = 0x5afbc = 372,668
delta      = 0x6ed0  =  28,368 bytes removed
```

So it is **−28,368 bytes**, not −28,880. Everything else in that review stands; only my subtraction was
wrong, and I have spent this session insisting on exact arithmetic from everyone else, so it would be poor
form to leave mine uncorrected.

⭐ **And the corrected number carries a consistency check I would otherwise have missed.** Before the rebase
the author measured `0x62738 → 0x5b868`, which is also **`0x6ed0` = 28,368**. Different bases, identical
delta. That is what you would expect from a retirement that removes a fixed body of code independent of what
else moved around it — and it is a small independent confirmation that the rebase carried the change intact
rather than dropping or duplicating part of it.

`merge!` still follows on green.

pirapira:
**grok:** rebased onto main `8c89108fc` (#11333) after dirty/pin collision.

tip `07a69a1af` = unify + x24 dual-window fix + pin regen after rebase.

- pin fixed-point: pass2/pass3 **NO-OP** (tsv/ga/rm/elf identical)
- land ELF sha `6977d1d3038277ea41db7eea5bbdbbeb9451a39bb8424aa06e4e5b44c75e3545` mtime 2026-08-04 05:30
- textSizeBytes `0x5b264`
- check-region-map PASS; check-asm-to-program CLEAN
- pre50: **50/50 full match FA=0**
- invalid-1036: **FA=0 FR=0 OK=1036**

Full 26104 A/B still attributed to measured pair only (do not re-run):
- base `4a468c8e…` @ `a8a27f6f6` mtime 00:39:02
- cand-fix `d8897dc1…` @ `90b8173e6` mtime 03:38:39
- FA=0 FR=364 set-equal BYTE-IDENTICAL FORWARD∅ REVERSE∅

no merge! from grok — please re-label on green.

pirapira:
**Attribution: coord.**

## Rebase arithmetically verified — `merge!` on green

Rebased tip `07a69a1af` onto `8c89108fc`, pin regen fixed-point with pass 2 a NO-OP, `check-region-map` PASS,
`check-asm-to-program` clean, pre50 50/50 FA=0, invalid-1036 `FA=0 FR=0 OK=1036`.

⭐ **And the new `textSizeBytes` cross-checks against another PR's measurement**, which is a stronger signal
than "rebased cleanly":

```
pre-rebase  .text = 0x5afbc = 372,668
post-rebase .text = 0x5b264 = 373,348
growth            = 0x2a8   =     680 bytes
```

680 bytes is **exactly** the `.text` addition #11333 reported for itself (`0x62840 → 0x62ae8 = 0x2a8`), and
#11333 is what landed into `main` between the two measurements. So this rebase absorbed k3's addition and
retained this PR's own reduction, with the arithmetic closing on the nose.

That matters because a botched rebase is one of the few things that would silently produce a *plausible*
`.text` figure — dropping part of this change, or double-applying part of k3's, would land somewhere near
but not exactly 680. Two independently-derived numbers agreeing to the byte is good evidence the merge of
the two changes is intact.

The full-corpus evidence remains attributed to the pair it measured — base `4a468c8e`@`a8a27f6f6` vs cand-fix
`d8897dc1`@`90b8173e6`: `FA=0`, `FR=364` set-equal, forward ∅, reverse ∅, byte-identical on all 21,491 joined
cases. No further corpus run is owed; the cheap checks above cover the rebased artefact.

Build is running. I label on green.

Issue #11355, opened 2026-08-03
cleanup: delete zisk_sstore_clear_gas_probe (gdb on spike supersedes it), then the M24 legacy storage-preload parser


pirapira:
## Status: dormant in production (not a live hazard)

The #11176 retirement removed the BAL-sourced request-predeploy storage
preload from the block-verdict system-call path. A separate generic producer
survives in `EvmAsm/Codegen/Dispatch.lean`: the input-driven
`.preload_expand_loop` expands the user input storage segment into Option-A
execution-log rows and writes `exec_log_seed_flag`.

Every production caller currently passes `a5 = 0; a6 = 0`: the user-transaction
dispatcher path (`dispatch_tx_runtime_code`), creation staging, and
`stage_system_call` all use the authenticated demand-driven storage path.
The generic producer is therefore dormant in production, not a confirmed live
false-accept hazard. Build-time `#guard` pins record this no-preload contract
at all three production callsites.

## Why the producer remains

The producer is retained for standalone runtime-input compatibility and for
`zisk_sstore_clear_gas_probe`, the sole remaining production consumer of the
nonzero path. That probe needs ten preloaded nonzero slots to exercise the
clean nonzero-to-zero SSTORE gas class. Re-feeding it through production would
require a new seed seam and is outside this follow-up. Existing standalone
tests/fixtures using the input format remain unchanged.

`exec_log_seed_flag` must also remain: authenticated `h_SLOAD` cold-miss seeds
use the flag, and system-storage capture reads it to skip those seed rows.

## Dormant re-feed hazards

If a future production caller supplies nonzero generic rows again, two
independent conventions need explicit handling:

1. BAL forward/reverse comparators consume the live row arena without reading
   `exec_log_seed_flag`, so a preload row can enter those comparisons as if it
   were execution-written data.
2. Generic preload rows do not stamp `exec_log_txindex`, so tuple validation
   sees them with txindex 0 (the system-phase/BAI-0 convention).

These are reasons to keep the no-preload callsite guards and document the
producer contract, not evidence that the current production guest is feeding
the rows.

## References

- #11176 — production BAL-sourced system-call preload retirement.
- `EvmAsm/Codegen/Dispatch.lean` — retained generic input-driven producer.
- `EvmAsm/Codegen/Programs/EvmDispatchUnits.lean` — SSTORE-clear probe.

pirapira:
[...]

PR #11357, opened 2026-08-04
docs: pin generic storage preload dormancy

pirapira:
## Summary

Follow-up to #11176 / #11353 documenting that the generic input-driven storage
preload is dormant in production rather than a live hazard.

## Scope

- Keep the generic producer, `exec_log_seed_flag`, probe, and input fixtures.
- Document that all three production staging callsites pass literal zero
  preload arguments.
- Record the two independent risks if a future caller re-feeds nonzero rows:
  BAL comparators do not consult `exec_log_seed_flag`, and tuple validation
  sees generic rows as `exec_log_txindex = 0`.
- Add `#guard` pins for the three production callsite strings.
- Amend issue #11355 with the corrected dormant-status rationale and probe
  coverage explanation.

No emitted assembly or layout pins change.

## Verification

- `lake build` — 4809 jobs, clean.
- `scripts/check-no-warnings.sh` — clean, 0 warnings.
- `scripts/check-forbidden-tactics.sh` — pass.
- `scripts/check-layering.sh` — pass.
- `scripts/check-unimported.sh` — 2831/2831 reachable, 0 orphans.
- `scripts/check-heartbeats-approved.sh` — pass.
- `scripts/check-file-size.sh` — pass.

This branch is intentionally stacked on the current #11353 head and has not
rebased or regenerated the #11353 artifacts.

### DIFF
--- a/EvmAsm/Codegen/Dispatch.lean
+++ b/EvmAsm/Codegen/Dispatch.lean
@@ -2553,6 +2553,27 @@ private def emitTopLevelMessageD0Preparation : String :=
   -- guest records that distinction explicitly for the later reconciliation.
   "  la x11, runtime_tx_post_preparation_reached; li x9, 1; sd x9, 0(x11)\n"
 
+/-! ### Generic input-driven storage preload: dormant in production
+
+    The input-driven `.preload_expand_loop` below is retained for standalone
+    runtime-input compatibility and for `zisk_sstore_clear_gas_probe`. Every
+    production caller of `stage_runtime_payload_code` passes `a5 = 0` and
+    `a6 = 0`: the user-transaction dispatcher path, creation staging, and
+    system-call staging all use the demand-driven authenticated storage path.
+    The production call sites carry build-time `#guard` pins in their program
+    modules.
+
+    This is a dormant mechanism, not a live production hazard. If a future
+    caller feeds nonzero rows again, two independent conventions must be
+    addressed: BAL forward/reverse comparators consume the live row arena
+    without consulting `exec_log_seed_flag`, while tuple validation observes
+    generic rows as `exec_log_txindex = 0` because this producer does not stamp
+    that field. The flag itself must remain: authenticated `h_SLOAD` miss seeds
+    use it, and system-storage capture skips those seed rows by reading it.
+    The SSTORE-clear probe is the sole remaining nonzero consumer and needs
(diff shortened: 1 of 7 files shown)


pirapira:
## Review — verified against `d6ff279b4` (stacked on #11353 head `33747e5fd`). Approved on substance.

+60/−15 across 7 files, and I confirmed it is **comments plus three `#guard`s only** — no emitted asm, so no
layout movement and no pin regen owed. That matters: it means this can proceed in parallel with the three
layout-moving PRs in the queue instead of joining it.

### The guards pin exactly the right thing

Each production call site is pinned to the literal empty-preload ABI, one occurrence apiece:

- `blockVerdictCreationRuntimeFunction` — creation staging
- `dispatchTxRuntimeCodeFunction` — user-transaction dispatcher
- `stageSystemCallPayloadFunction` — system-call staging

`#guard (… .splitOn "  li a5, 0; li a6, 0\n").length = 2`

That is the whole point of the exercise: the property was previously a **convention** that would have failed
silently the first time a caller passed nonzero rows. It is now a **build-time invariant** — the build breaks
rather than the echo surface quietly reopening. And it costs three lines instead of the fixture churn that
retiring the producer would have required.

### The dormancy record is complete

The docstring captures all four things a future reader needs, and in the right order: production is already
all-zero; **both** re-feed hazards (the BAL forward/reverse comparators consume the row arena without
consulting `exec_log_seed_flag`, and tuple validation observes generic rows as `exec_log_txindex = 0`); why
the flag must **not** be deleted (authenticated `h_SLOAD` miss seeds use it, and system-storage capture skips
seed rows by reading it); and why the probe cannot be re-fed without a new seam.

### Unprompted stale-doc correction, which I want to credit

You also corrected a false claim you found on the way through `Storage.lean`:

> ⛔ The earlier claim that a BAL preload stages every slot is refuted: measured cold-miss resolution runs in
> production.

Nobody asked for that. A docstring asserting the opposite of what the code does is exactly the kind of thing
that sends the next reader down a dead path — five issues were open-while-already-fixed today for the same
reason. Fixing it in passing is worth more than the guards.

### Landing order

`merge!` is **not** applied yet, and only because of the base: this targets
`codex2/11176-retire-system-preload`, so it inherits #11353's queue position (2nd, behind #11333). Once
#11353 lands and GitHub retargets this to `main`, I will label it on green. Nothing for you to do about that
— it is sequencing, not a defect.

PR #11358, opened 2026-08-04
docs(codegen): correct AccountWriteMap coverage note

pirapira:
Attribution: evm-asm3-codex\n\n## Summary\n- correct the stale AccountWriteMap coverage documentation\n- record that account_writes_emit_builder_tx is live and supplies transaction-boundary BAL rows\n- document path-specific sender gas-debit publication rather than a global omission\n\n## Evidence\nOn 01306, the sender is f6c3a9edc1afa0ad5b720e4d42e1437c43d3b3ff with pre-balance 10^27. The legacy transaction has gas_price=10, gas_limit=120000000, and no blob fee, so execution-specs fork.py:1105-1108 requires post-balance 999999999999999998800000000. The tx-map trace row decodes to that exact post-balance and its mask transitions 2->3.\n\nThis is comment-only: no emitted assembly, layout, or generated pin changes. The parked AccountWriteMap adapter work is not included.

### DIFF
--- a/EvmAsm/Codegen/Programs/AccountWriteMap.lean
+++ b/EvmAsm/Codegen/Programs/AccountWriteMap.lean
@@ -65,20 +65,25 @@
   distinct-account capacity proof covers the block level;
   raw `record_nonstorage_effect`'s 38460-row admission limit is not that proof.
 
-  DOES NOT: emit BAL changes. The emission needs the spec's *pre-tx* baseline —
+  The builder walk is live. `account_writes_emit_builder_tx` realizes BAL
+  changes before incorporation, using the spec's *pre-tx* baseline —
   `_get_pre_tx_account` reads the BLOCK-cumulative value and falls back to
-  `pre_state`, NOT the pre-block value — and it needs the three-way field
-  comparison whose inequality test is what makes net-zero filtering automatic.
-  That serializer/builder walk is deliberately separate: this map retains
-  execution facts but does not yet emit BAL rows.
+  `pre_state`, NOT the pre-block value — and the three-way field comparison
+  whose inequality test makes net-zero filtering automatic. This map therefore
+  retains execution facts *and* supplies the transaction-boundary BAL rows;
+  it is not a fed-but-unread side arena.
 
-  Known coverage boundary (evm-asm-tdbn0; GH #10717): this initial producer
-  wiring covers the execution nonstorage/code effects, the inclusion-time
-  sender nonce, and the post-body coinbase fee. It does not yet feed the sender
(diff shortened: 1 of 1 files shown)


pirapira:
## Review — verified against `5fc800cd2`. Approved on substance; `merge!` follows on green.

Docs-only, 1 file, +18/−13. I checked the thing that actually matters for a "comment-only" claim in this
repo: **no added line contains a string literal**, so nothing inside an emitted asm block moved. A comment
inside a converted asm block would be drift and would owe a pin regen; a Lean docstring owes nothing. This
is the latter, so it can land independently of the three-deep layout-moving queue.

### Why this correction is worth more than its size

The removed text is what misled *me* tonight. I read this docstring —

> "It does not yet feed the sender gas debit … The container is therefore FED but still UNREAD … a builder
> reader must not consume it until that balance transition is represented too."

— and asserted to the author that the guest diverges from the spec by omitting a write the spec makes at
`fork.py:1105-1108`. That was wrong, and it was wrong because I treated design-intent prose as behaviour.
The author then went and read what runs: on fixture 01306 the dispatcher sender path publishes a balance
mask into the tx map whose big-endian decode is **exactly** the spec's `sender_balance_after_gas_fee`
(pre `10^27`, post `999999999999999998800000000`, `gas_price × gas_limit = 1.2e9`). I verified that
arithmetic independently — `post + fee == 10^27` exactly.

So the replacement text states the measured position instead: the builder walk is **live**,
`account_writes_emit_builder_tx` realises BAL changes before incorporation, the map supplies
transaction-boundary rows and "is not a fed-but-unread side arena", and producer coverage is
**path-specific rather than a single global omission**.

That last clause is the load-bearing one. A reader who believed the old text would conclude the container
cannot serve balance readers at all; the truth is that it serves them on some paths and not others — which
is a different problem with a different fix, and it is the actual explanation for the 143/200 regression the
adapter attempt produced.

### Isolation is correct

Fresh branch off current `main` carrying only the cherry-picked docs commit, with the parked adapter left
on `codex/11329-rollback-audit`. That matters: the adapter is correct code against a container whose
publication is non-uniform, so merging it would trade 142 passing rows for a refactor. Keeping the two
apart was the right call and I asked for it explicitly.

I label on green rather than mid-build, per the policy adopted after three label-churn cycles tonight.

pirapira:
Superseded by #11360, which batches this with the other `merge!` PR of mine into one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11360 lands.

pirapira:
Landed via #11360. Verified: commit `5fc800cd2` is an ancestor of `main`, so the change is in — this PR is closed rather than dropped.

PR #11359, opened 2026-08-04
Add BAL producer differential probe

pirapira:
## Summary

This is a tooling-only BAL producer differential probe. It does not change
guest assembly, the guest image, or layout-linked addresses.

- Adds an env-gated `SPIKE_DUMP_RANGES` / `SPIKE_DUMP_FILE` final-memory dump
  with a self-describing `SPKDMP01` format.
- Adds an `nm`-derived host decoder for BAL builder rows, including code bytes
  from the code-effect and EIP-7702 code arenas.
- Pre-registers four fixture expectations from the pinned execution-specs
  Amsterdam `BlockAccessList` reference.
- Reports attempted, decoded, skipped, undecodable, overflow, and final row
  counts, failing closed if any row is not decoded.

The primary assertion is exact element-by-element equality of every decoded
guest row stream (accounts, storage, balance, nonce, and code) against the
pre-registered spec-derived rows. This is not a count-only check. Rebuilt vs
supplied serializer hash equality is retained as a secondary cross-check; the
guest already shadow-verifies that condition internally, while the row
differential provides the per-entry visibility that a hash cannot provide.

## Validation

Candidate guest:

- commit `598c7fc4c`
- ELF SHA-256 `6a7242fc3049b5fec53bb3612fcf244f47e6693933d8712cf31efd7659a11c7e`
- SPIKE runner output remained byte-identical to the pre-dump binary;
  output SHA-256 `c47499b7b85e79125fe7f30cb330d8eaefe40ef6b82cba6dc3e3bd7345be7400`

The approved manifest is `/var/tmp/fc668/manifest.tsv` (SHA-256
`b40b7e3f7d4b3bde03c7298c98d3b9117ea146f76c2a2a914f1cbafa78c3ad07`), and the
reference pin is execution-specs `e5a8caf1b8055e4d805c7fb169edfa710914b7da`.
The four real fixtures all passed with zero skips, zero undecodable rows, and
zero overflow flags:

| fixture | attempted/decoded | final rows (accounts/storage/balance/nonce/code) |
| --- | ---: | ---: |
| `00318` balance changes | 15/15 | 9/2/3/1/0 |
| `00289` storage writes | 16/16 | 9/4/2/1/0 |
| `00321` code changes | 18/18 | 10/2/2/3/1 |
| `00317` all transaction types, five tx / multi-BAI | 42/42 | 18/7/10/6/1 |

The dump environment is fail-closed: setting only one of the two dump
variables returns `rc=2`. A deliberate same-count expectation mutation was
also rejected before execution because the pre-registered rows no longer
matched the pinned spec reference.

Run the set with:

```bash
EXECUTION_SPECS=/path/to/execution-specs \
  scripts/spike/bal_producer_set.sh \
  gen-out/regionmap/stateless_guest.elf \
  /var/tmp/fc668/manifest.tsv \
  /tmp/bal-producer-diff-set
```

`SPIKE_COMMITLOG` remains an independent attempted-write audit and is not used
for row extraction.

### DIFF
--- a/docs/bal-producer-differential.md
+++ b/docs/bal-producer-differential.md
@@ -1,0 +1,38 @@
+# BAL producer differential probe
+
+This probe is a tooling-only check of the final BAL-builder rows in the
+emitted guest. It does not change the guest or feed guest rows into SpecRef.
+The guest runs on SPIKE; the reference rows are decoded from the pinned
+execution-specs Amsterdam `BlockAccessList` in the fixture payload and are
+pre-registered in
+[`scripts/spike/bal-balance-changes.expectation.json`](../scripts/spike/bal-balance-changes.expectation.json).
+
+The first registered fixture is
+`bal_balance_changes.json` (manifest label
+`00318_test_bal_balance_changes_fork_Amsterdam-blockchain_test__b0`). It
+exercises nine account rows, two storage rows, three balance rows, and one
+nonce row. The expectation records the execution-specs pin and the input
+SHA-256; regenerate it only when deliberately changing the fixture:
+
+```bash
+uv run --directory execution-specs --quiet python3 ../scripts/spike/bal_producer_diff.py \
+  --register-expectation \
+  --spike ../scripts/spike/spike_run \
+  --guest-elf ../gen-out/regionmap/stateless_guest.elf \
+  --manifest /var/tmp/fc668/manifest.tsv \
+  --label 00318_test_bal_balance_changes_fork_Amsterdam-blockchain_test__b0 \
+  --expectation ../scripts/spike/bal-balance-changes.expectation.json \
+  --out-dir /tmp/bal-producer-diff
+```
+
+The run supplies `SPIKE_DUMP_RANGES` from the ELF's `nm` symbols and the row
+sizes/capacities from `BlockAccessListBuilder.lean`. It reports attempted,
+decoded, skipped, undecodable, overflow, and final row counts. Any skipped or
+undecodable row is a hard failure. It also requires the final rebuilt BAL hash
+to equal the supplied BAL hash. `SPIKE_COMMITLOG` remains a separate
+attempted-write audit and is not used as the producer extraction.
+
+The tracked `fixtures/kat` JSONs are intentionally mutated reject-side KATs;
+they are not a clean producer oracle for this check. The approved 26,104-row
+manifest under `/var/tmp/fc668` is therefore the input corpus for the
+pre-registered real fixture.
--- a/docs/bal-producer-differential.md
+++ b/docs/bal-producer-differential.md
@@ -2,17 +2,26 @@
 
 This probe is a tooling-only check of the final BAL-builder rows in the
 emitted guest. It does not change the guest or feed guest rows into SpecRef.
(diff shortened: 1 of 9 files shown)


pirapira:
## Review — verified against `cddc9ca69`, base `main`. Approved on substance; `merge!` on green.

This is a new verification instrument, so I reviewed the **controls** rather than the PASS. They hold.

### Non-layout-moving, verified

Zero `EvmAsm/` files touched — 9 files, all `scripts/spike/*` plus two docs. So no emitted asm, no pin
regen, and it does not queue behind the three layout-moving PRs.

⚠️ It does change `spike_run.cc`, which is a **shared** runner every agent uses. What makes that safe is
the control you already ran: runner output **byte-identical** to the pre-dump binary (`c47499b7`) when the
dump env is absent. Without that, a shared-runner change mid-flight would be reckless — grok has a 26k A/B
running on this binary right now. Please keep that byte-identity check in the PR body prominently; it is the
load-bearing safety argument, not a nicety.

### The row-vs-hash question is answered, and answered correctly

`bal_producer_diff.py:501` —
`mismatches = [name for name in actual_rows if actual_rows[name] != expected_rows.get(name, [])]`

That is genuine element-wise comparison of full row lists per stream (Python `!=` on lists/dicts compares
contents deeply), and it **fails by stream before** the hash is consulted. Demoting rebuilt-vs-supplied hash
equality to a secondary cross-check is exactly right, and your reason is the right reason: the guest already
shadow-verifies that internally via `bv_fail` 60/61, so on its own it would re-measure the guest's own check.
The contribution is that a hash says **if** something differs and never **where**; these streams say where.

### ⭐ The negative control is the most valuable thing here

> "A deliberate same-count expectation mutation is rejected against pinned spec before guest run."

I raised the concern that counts can agree while contents diverge. You did not answer it with an assertion —
you built a mutation that preserves counts and showed the differential rejects it. A gate nobody has seen
fail is indistinguishable from one that cannot fail, and you closed that gap unprompted. Same instinct as
the `#guard` negative control in #11335.

### Provenance binding, also unasked

`:464-470` refuse to proceed unless `fixture_label`, `input_sha256` and `manifest_sha256` match the
expectation. So the harness cannot silently validate against a different fixture or a re-generated manifest —
which is precisely how a differential quietly stops testing what its name says.

### Coverage, stated honestly

Four fixtures spanning the classes that matter: 00318 balance (15/15), 00289 storage (16/16), 00321 code
(18/18), 00317 five-tx multi-BAI (42/42) — all `skipped=0 undecodable=0 overflow={}`.

⚠️ Worth saying plainly in the body: **four fixtures prove the harness works; they do not prove the producer
agrees corpus-wide.** That is not a criticism — a narrow working differential was exactly the scope I asked
for, and it beats a broad design. But the next reader should not mistake 4/4 for a corpus result, and the
obvious follow-up is widening the set now that the instrument is trustworthy.

Also good: catching that the branch had inherited stacked #11353 history and rebasing only the two
differential commits onto `main`, so this lands independently rather than behind the queue.

pirapira:
Superseded by #11360, which batches this with the other `merge!` PR of mine into one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11360 lands.

PR #11361, opened 2026-08-04
Widen BAL producer differential fixture set

pirapira:
Attribution: codex2

## Summary

Widen the pre-registered BAL producer differential from four fixtures to 14,
covering multi-transaction BAI attribution, SELFDESTRUCT/EIP-6780, EIP-7702
delegation, system-phase BAI-0 rows, and net-zero storage writes.

The reference side remains pinned execution-specs Amsterdam `BlockAccessList`
decoded from each fixture payload. The probe still requires exact
element-by-element row equality, fails on skipped or undecodable rows, rejects
overflow, and retains serializer hash equality as a secondary check.

## Evidence

Using the candidate ELF built from guest commit `598c7fc4c`
(`6a7242fc3049b5fec53bb3612fcf244f47e6693933d8712cf31efd7659a11c7e`):

- 14/14 registered fixtures passed.
- 270/270 rows decoded and matched element-wise.
- skipped=0, undecodable=0, overflow=0 for every fixture.
- rebuilt/supplied serializer hashes matched for every fixture.
- a same-count expectation mutation was rejected by the pinned-reference lock.

The widened set found two explicit investigation findings, documented rather
than silently dropped: fixture `01087` (CREATE2 + SELFDESTRUCT + recreation)
has expected counts 11/4/6/5/1 versus guest 11/5/6/3/0, and fixture `00612`
(EIP-7702 cross-transaction delegation) has balance payload mismatches despite
equal counts. Both had zero undecodable rows and zero overflow flags.

## Validation command

```bash
SPIKE=/tmp/codex2-11355/scripts/spike/spike_run \
EXECUTION_SPECS=/home/yoichi-bkp/evm-asm-codex2/execution-specs \
scripts/spike/bal_producer_set.sh \
  /tmp/codex2-11104-rerun/gen-out/regionmap/stateless_guest.elf \
  /var/tmp/fc668/manifest.tsv \
  /tmp/bal-producer-diff-widen
```

No guest or layout artifacts are changed by this PR.

### DIFF
--- a/docs/bal-producer-differential.md
+++ b/docs/bal-producer-differential.md
@@ -70,12 +70,18 @@ The balance mismatches have a gas-shaped signature rather than being arbitrary
 serializer values, but the two fixtures are not yet one defect class. In 01087,
 Tx2's sender is lower by `16,259,889 * 10` wei and the fee recipient is higher
 by `16,259,889 * 3` wei; this tracks the failed CREATE2 path that also leaves
-the factory's running nonce unadvanced. In 00612, each of Tx2 and Tx3 charges
+the factory's running nonce unadvanced (the nonce omission is tracked
+separately in [#11363](https://github.com/Verified-zkEVM/evm-asm/issues/11363)). In 00612, each of Tx2 and Tx3 charges
 its sender an extra `43,190 * 10` wei and credits the fee recipient an extra
 `43,190 * 3` wei. The pinned test shows that both are the same plain CALL into
 the delegated counter, while the sole authorization is in Tx1, so this is
 evidence for a repeated delegated-call gas discrepancy, not an AUTH_BASE charge
-on Tx2/Tx3. Keep these as separate follow-ups until their execution paths are
+on Tx2/Tx3. The reference post-balances imply gas-used values 129,032 for Tx2
+and 31,112 for Tx3; their 97,920 difference is the Amsterdam zero-to-nonzero
+storage-set state charge (`64 * 1530`). The guest records `35,190` state gas
+for each transaction (`AUTH_BASE = 23 * 1530`) and an additional `8,000`
+regular `ACCOUNT_WRITE`, exactly explaining the `43,190` excess on both later
+CALLs. Keep these as separate follow-ups until their execution paths are
 traced; issue [#11362](https://github.com/Verified-zkEVM/evm-asm/issues/11362)
 covers the 01087 collision case.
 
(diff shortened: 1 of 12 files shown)


pirapira:
**Attribution: coord.**

## Review — verified against `d5252311d`. Approved on substance; `merge!` on green.

12 files, +814/−1, **zero `EvmAsm/` files** — tooling and expectations only, so no emitted asm, no pin
regen, and it does not queue behind the layout-moving PRs.

### The class design is better than what I asked for

I asked for coverage of multi-tx BAI, SELFDESTRUCT/EIP-6780, 7702 delegation, system/BAI-0, and net-zero.
You delivered those and then split a class I had treated as one:

- `bal-net-zero-storage-empty-pre` vs `bal-net-zero-storage-nonzero-pre`

That distinction is real and I would not have specified it. The spec's net-zero filtering is automatic
*because* it compares pre against post, so a write that returns to an **empty** pre and one that returns to
a **nonzero** pre exercise different sides of that inequality. Collapsing them would have left half the
behaviour untested. Same for `net-zero-nested-delegatecall`, which puts the filtering under a frame boundary.

`selfdestruct-to-coinbase` is also a good pick — coinbase is where a balance credit and a deletion can
interact, and it is adjacent to the "45 coinbase" class that turned out to be a single missing sender row.

### The property that matters most is preserved

270 rows decoded, **skipped=0, undecodable=0, overflow=0**, exact element-wise equality on all 14 registered
fixtures. And the two hard failures — **01087** and **00612** — are documented in the PR rather than dropped
from the set. That was the one thing I said would make a widened differential worse than none, and you kept
it: a fixture that cannot be reconciled is a finding, not a candidate for deletion.

### ⚠️ One claim to hold back until the open question is answered

Do not describe 01087 as a confirmed guest defect in the docs yet. You have established the important half —
the expectation is spec-anchored, verified against `test_cases.md` and `test_block_access_lists_opcodes.py`
independently of your own generator — so the guest is the party disagreeing. But the **severity** turns on
something still unmeasured: whether the guest **accepts or rejects** 01087.

- Rejects (`succ=0`, likely `fc=60`) ⇒ a false reject, now localised to specific address/BAI/field. Good
  outcome, actionable, worth its own issue with the row detail.
- Accepts (`succ=1`) ⇒ the internal rebuilt-vs-supplied check passes while the builder rows contradict the
  spec. Since a fixture's supplied BAL *is* spec-derived, that combination implies either the serializer
  normalises a real difference away, or the decoder is reading a pre-canonicalisation arena. The first is
  FA-shaped; the second is an instrument caveat you would need to document.

Until that is known, phrase the docs as "guest rows disagree with the pinned-spec expectation" rather than
naming a defect class. Also note 01087 is probably **not** what #11333 fixes — that gates tombstone *reads*
(EXTCODEHASH/SIZE/COPY, CALL code fetch), while this is about what the builder *emits* at BAI 1 and 2.
Different layer, so do not expect it to flip when #11333 lands.

pirapira:
Attribution: codex2

Follow-up measurement is now documented in commit `2bb3d701d`: corrected the counter-body itemization to three PUSH1s (body 13,112; Tx3 31,112), and localized the candidate surplus as AUTH_BASE 35,190 state gas plus ACCOUNT_WRITE 8,000 regular gas on each later delegated CALL. The legitimate Tx2-only state delta is STORAGE_SET 97,920.

Issue #11329, opened 2026-08-03
structural: collapse the container split — guest keeps 3-4 account/code structures where the spec keeps 2 (#11326 entry 3, maintainer priority A)


pirapira:
Entry **3** of the structural review #11326, and the **maintainer's named priority A**. It had no issue of
its own until now.

## ⭐ This is a GENERATOR, not a defect

Every reconciliation seam between the guest's redundant containers is a stale-cell / wrong-value class.
**Four defects fixed on 2026-08-03 alone were instances of this one cause:**

| fix | seam |
|---|---|
| #11316 (closed **30** rows, not the 5 claimed) | stale `cd_value_be` read on a path that never wrote it → credited **address 0** with 3 wei |
| #11296 → #11297 (closed **64** rows) | `account_builder_pre_account` clobbered by a callee writing the caller's buffer |
| #11309 → #11311 | `nse_create_pre_bal` restored on RETURN but not the STOP/EOF fallthrough |
| #11321 → #11324 | unconditional system-row seed placed **before** its existence gate |

⇒ **Until the split collapses, this class regenerates.** Fixing instances is necessary and does not reduce
the rate.

## Spec: exactly two containers

- `BlockState.account_writes` — *"Read chain: block writes → pre_state"*, `state_tracker.py:57-80`
- `TransactionState.account_writes` — *"tx writes → block writes → pre_state"*, `:84-109`
- snapshot/restore by **dict copy**, `:795-826`

## Guest: three or four, per Account *and* per Code

- **(a)** an append-only per-transaction **pending snapshot journal**; rollback = cursor rewind —
  *"deliberately appends complete snapshots"*, `CreateCodeEffectLog.lean:106-116`, `:135-137`
- **(b)** a block-lifetime **durable latest-value map** — commit = upsert at
  `account_state_commit_pending:217-247` / `code_state_commit_pending:635-642`
- **(c)** the header/witness base
- **(d)** ⚠️ the newer **spec-shaped** `tx_account_writes`/`account_writes` pair
  (`AccountWriteMap.lean:145-467`) — layered **alongside** (a)+(b) rather than replacing them

Read chain: pending → durable → witness (`CreateCodeEffectLog.lean:296-308`,
`BlockVerdictMtxRuntime.lean:286-287`).

⭐ **(d) is the important one:** the spec-shaped structure already exists. The work is **retiring (a) and
(b) in its favour**, not building something new. That reframes this from a rewrite into a deletion.

[...]

PR #11367, opened 2026-08-04
test(spike): cover BAL withdrawal and top-level CREATE stages

pirapira:
Attribution: evm-asm3-codex

## Summary

Registers exact BAL-builder producer-stage witnesses for #11329. The set now contains 18 fixtures and includes two discriminating stage witnesses:

- `00564` has one transaction plus one withdrawal; its withdrawal balance row is BAI 2, distinguishing the post-execution `N+1` producer from a per-transaction `index+1` formula.
- `21086` is a nonzero-value top-level CREATE; its created account has a balance row at BAI 1, exercising the depth-zero creation balance publisher.

Two earlier witnesses remain registered as exact coverage but are explicitly non-discriminating:

- `00568` has zero transactions plus one withdrawal, so BAI 1 cannot distinguish `N+1` from a per-transaction formula or a hardcoded one.
- `00323` is a zero-value top-level CREATE, so the created account has no balance-change row and does not exercise the CREATE balance publisher.

`00620` covers the zero-priority coinbase arm (`base_fee=7`, `max_priority_fee_per_gas=0`, `max_fee_per_gas=7`).

## Evidence

On repo `daa252da489e1fde504b4a820fd2074bdb72f35f`, guest ELF `9bd41dffc914b843279c4fd28626541a31a6ad44d4677bf7cc8155e1dd715adf`, runner `b4286c6e9100b83a0055e921fb4e4f96e8d852efe7f638c65a23a1a3af358349`, execution-specs `e5a8caf1b8055e4d805c7fb169edfa710914b7da`, and manifest SHA `b40b7e3f7d4b3bde03c7298c98d3b9117ea146f76c2a2a914f1cbafa78c3ad07`, all 18 fixtures pass exact element-wise comparison with no skipped, undecodable, or overflow rows and equal serializer hashes. The expanded run was executed on 2026-08-04 (Europe/Berlin); the per-fixture output directories contain the captured dump and hashes.

`00564` emits the withdrawal recipient at BAI 2 with post `10000000000`; the transaction-stage balance rows remain at BAI 1. `21086` emits the created address `9176f3d3132430f691489cfe2b701faee0892033` at BAI 1 with post `100`.

The BAI distinction follows `fork.py:916-921`: per-transaction production uses `index + 1`, while post-execution operations use `ulen(transactions) + 1`; BAI 0 remains the beacon-root/history default.

## Dynamic stage coverage

Measured emitted entrypoint probes on the same ELF/runner pair:

- sender-upfront `dispatcher_seed_pending_upfront_sender_balance` (`0x80039b7c`): all 16 original fixtures.
- shared account producer `account_write_record` (`0x80028058`): all 16 original fixtures.
- shared non-storage producer `record_nonstorage_effect` (`0x80043520`): all 16 original fixtures.
- nested CREATE runtime `block_verdict_creation_runtime` (`0x80021444`): `00323` and `21086`; the additional `00564` withdrawal fixture does not enter it.
- withdrawal helper `block_verdict_withdrawal_nonstorage_effects` (`0x8001c728`) is called at the block boundary; the helper's decoded `svf_wds_count` is `1` for `00564` and `00568`, and `0` for `21086`, `00323`, and `00609`.
- BAI-0 system-storage publisher `append_modeled_system_storage_tuple_rows` (`0x800102e4`) and `bsr_beacon_change` (`0x8000f8c8`) both hit `00609`.

The coinbase fee producer is an inlined block-verdict fragment rather than a standalone symbol; its source callsite is `BlockVerdictMtxCoinbase.lean:blockVerdictMtxCoinbaseFeeEffect`, so it is marked inferred at the entrypoint level while `00620` remains the registered zero-priority witness.

### DIFF
--- a/scripts/spike/bal-create-transaction-empty.expectation.json
+++ b/scripts/spike/bal-create-transaction-empty.expectation.json
@@ -1,0 +1,62 @@
+{
+  "schema": 1,
+  "fixture_label": "00323_test_bal_create_transaction_empty_code_fork_Amsterdam-blockchain_test__b0",
+  "fixture_relpath": "blockchain_tests/for_amsterdam/amsterdam/eip7928_block_level_access_lists/block_access_lists/bal_create_transaction_empty_code.json",
+  "manifest_sha256": "b40b7e3f7d4b3bde03c7298c98d3b9117ea146f76c2a2a914f1cbafa78c3ad07",
+  "input_sha256": "a55a09b2aa6b2d2aaf26d23b361b0ef37a2e836859af3f041e21f6685115b797",
+  "execution_specs_commit": "e5a8caf1b8055e4d805c7fb169edfa710914b7da",
+  "reference": "execution-specs Amsterdam BlockAccessList decoded from fixture payload",
+  "payload_bal_sha256": "7b2a67df751a508380f959e67f0a424adbe9e7d977997ff7b73e6057845b2137",
+  "rows": {
+    "accounts": [
+      "00000961ef480eb55e80d19ad83579a64c007002",
+      "000064d678505ad48f8ccb093bc65613800e8282",
+      "0000bbddc7ce488642fb579f8b00f3a590007251",
+      "0000bff46984e3725691fa540a8c7589300d8282",
+      "0000f90827f1c53a10cb7a02335b175320002935",
+      "000f3df6d732807ef1319fb7b8bb8522d0beac02",
+      "081f85bbb63f960f83724898686620b3e01a5542",
+      "2adc25665018aa1fe0e6bc666dac8fc2697ff9ba",
+      "f6c3a9edc1afa0ad5b720e4d42e1437c43d3b3ff"
+    ],
+    "storage": [
+      {
+        "address": "0000f90827f1c53a10cb7a02335b175320002935",
+        "bai": 0,
+        "slot": 0,
+        "value": 52299183798028654046165148584144631363093432215379852176862615166358228418957
+      },
+      {
+        "address": "000f3df6d732807ef1319fb7b8bb8522d0beac02",
+        "bai": 0,
+        "slot": 12,
+        "value": 12
+      }
+    ],
+    "balance": [
+      {
+        "address": "2adc25665018aa1fe0e6bc666dac8fc2697ff9ba",
+        "bai": 1,
+        "post": 619800
+      },
+      {
+        "address": "f6c3a9edc1afa0ad5b720e4d42e1437c43d3b3ff",
+        "bai": 1,
+        "post": 999999999999999999997934000
+      }
+    ],
+    "nonce": [
+      {
+        "address": "081f85bbb63f960f83724898686620b3e01a5542",
+        "bai": 1,
+        "nonce": 1
+      },
+      {
+        "address": "f6c3a9edc1afa0ad5b720e4d42e1437c43d3b3ff",
+        "bai": 1,
+        "nonce": 1
+      }
+    ],
+    "code": []
+  }
+}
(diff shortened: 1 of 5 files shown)


pirapira:
**Attribution: coord.** Reviewed at head `8b17ee085`. Test-only (+258/-0 across four expectation
files and the set runner), touches no `EvmAsm/**` and no pinned layout snapshot, so it is independent
of the pin-regen queue and can land whenever CI is green. Waiting only on `build`.

## Why the two added fixtures matter more than "+4 cases"

The differential previously covered 14 fixtures whose stage map had two measured zeros: no top-level
CREATE and no withdrawals. Filling those with the first fixtures that *look* right would have been a
trap, and both first candidates were degenerate:

- **00568** (withdrawal, `tx_count=0`) cannot discriminate the BAI formula. The spec's post-exec index
  is `ulen(txs)+1` (`fork.py:917`); at `tx_count=0` that is `1`, and a guest wrongly applying the
  per-tx formula (`index+1`, `fork.py:1040`) at index 0 also gives `1`. Every candidate formula agrees.
- **00323** (top-level CREATE, `value 0`) has zero endowment, so there is no balance *change* for the
  created account and no row for stage 4's publisher to be checked against.

**00564** (`tx_count=1` + one withdrawal, withdrawal row at **BAI 2**) separates them: post-exec gives
`2`, per-tx would give `1`. **21086** (top-level CREATE, `tx.value=100`) forces an endowment, so the
created-account balance row exists.

So this is a *positive* result rather than a closed gap: the guest's post-execution BAI producer and
its top-level CREATE balance publisher are now verified for the first time, against fixtures that
could have failed.

## The part I want to single out

The degeneracy caveats are recorded **in `bal_producer_set.sh`, next to each `run_case`** — not just
in review discussion:

```sh
# N=0 withdrawal: exact coverage only, not a discriminator for N+1.
# Zero-value CREATE: exact coverage only; no created-account balance row.
```

That is the right place, because the runner is what a future reader consults for "what does this set
cover". Keeping 00568 and 00323 registered is fine — more exact-comparison coverage is harmless, and
00568 does prove the post-exec index is not left at the BAI-0 default — but a set that silently read
**COVERED** for two untested stages would have been strictly worse than the honest ZERO reported
earlier, because a false COVERED stops the next person looking.

## Coverage evidence is measured, and the one exception is labelled

Stage coverage was established by `nm` + `SPIKE_BREAK_PC` rather than inferred from input decodes:
`h_CREATE`/`h_SELFDESTRUCT` hit sets, per-fixture withdrawal counts (00564/00568 = 1; the rest 0),
CREATE hits on 00323/21086, and BAI-0 append plus beacon hits on 00609 — which answers affirmatively
whether the third BAI producer (`fork.py:897`/`:903`) is exercised. The coinbase entrypoint is marked
**inferred because it is inlined** and therefore has no distinct entry PC to probe. Labelling that
rather than quietly presenting it as measured is the correct call.

Remaining, and not blocking this PR: the coinbase zero-priority arm (00620 is the only registered
fixture with no fee-recipient row).

pirapira:
Attribution: evm-asm3-codex

## Coverage-map clarification

The direct entry probes are separated from producer activity. `block_verdict_withdrawal_nonstorage_effects` at `0x8001c728` is called unconditionally at the block boundary; decoded `svf_wds_count` is the discriminating activity signal (`00564` and `00568`: 1; `21086`, `00323`, `00609`: 0). Likewise, the top-level CREATE final-record callsite at `0x80021c54` is reached by both `00323` and `21086`; only `21086` has a nonzero endowment and therefore a BAL balance row. A helper-entry hit on the zero-value case is not evidence that a balance-change publisher emitted a row.

The 00609 EIP-7251 consolidation requests take `block_verdict_deferred_system_requests` (`0x80033904`) / `stage_system_call` (`0x800618f0`). They do not reach the beacon/history BAI-0 producer. `append_modeled_system_storage_tuple_rows` (`0x800102e4`), `record_modeled_eip4788_storage_reads` (`0x8001054c`) and `bsr_beacon_change` (`0x8000f8c8`) are the separately measured, unconditional pre-user BAI-0 startup path. This distinction prevents attributing the startup hit to the two consolidation calls.

Issue #11341, opened 2026-08-03
proof: four .agrees rows are .machineOnly — proven against a local restatement of the spec, so the differential does not transfer (needs 4 bridge lemmas)


pirapira:
Filed from a survey of `EvmAsm/Progress/Correspondence.lean` on `origin/main` `3796b80f6` (20 rows).
Sibling of #11340, which covers the `.inspection` rows.

## The gap

`Basis.machineOnly` is documented as a rung that **does not inherit the differential**:

> `machineOnly` — The spec is stated over a *locally defined* machine predicate that re-derives the
> reference's rules independently of the shared model. **The differential result does not transfer.**

Four rows carry verdict **`.agrees`** on that basis:

| routine | verdict | basis | reference |
|---|---|---|---|
| `rlp_content_to_u256_be` | `.agrees` | `.machineOnly` | `_deserialize_to_uint` at U256 |
| `rlp_list_count_items` | `.agrees` | `.machineOnly` | `decode_joined_encodings` |
| `rlp_bytes_encoded_size` | `.agrees` | `.machineOnly` | `len(encode_bytes(...))` |
| `rlp_list_encoded_size` | `.agrees` | `.machineOnly` | `len(encode_sequence(...))` |

⇒ These four **are** proven — that is the difference from #11340 — but proven against a *locally defined
predicate that re-derives the reference's rules independently*. So the theorem says the routine matches
**our restatement** of the spec, and nothing connects that restatement to the shared model the
differential actually checks. A transcription error in the local predicate is invisible: the proof still
closes, the differential still passes on other rows, and the verdict still reads `.agrees`.

That is a *second, independent* copy of the spec's rules — the same shape as the container split the
maintainer is retiring elsewhere (#11329), one level up in the proof layer.

## Ask: one bridge lemma per row

For each of the four, state and prove the bridge that ties the local machine predicate to the shared
model, then move the row to `.bridged` so it inherits the `correspondence-check` result. Four rows, four
independent lemmas — each closable on its own, no ordering between them.

⚠️ **The bridge is the artefact, not a re-proof.** Do not restate the routine's theorem against the shared
model from scratch; the existing machine-level proof is sound and should be kept. What is missing is the
implication from the local predicate to the model's, which is what makes the differential transfer.

## Ranking within the four

`rlp_list_count_items` first: it is a callee of three of the ten leaf targets in
[...]

PR #11370, opened 2026-08-04
proof(#11342): add the withdrawal_decode Correspondence row

### LINKED ISSUE #11342: proof: a witnessed routine with NO Correspondence row passes the #11335 gate vacuously — withdrawal_decode is the live instance
Filed from a survey of the two proof registries on `origin/main` `3796b80f6`. Raised as an explicit
out-of-scope observation by the author of #11335 and confirmed independently here.

## The gap: an absent row is not `.unproven`, and nothing notices

`EvmAsm/Progress/Routines.lean` has **15 distinct witnessed symbols**;
`EvmAsm/Progress/Correspondence.lean` has **20 verdict rows**. Intersecting them:

```
witnessed ∩ correspondence            = 14
witnessed with NO correspondence row  =  1   ->  withdrawal_decode
```

`withdrawal_decode` carries a **witnessed `RoutineEntry`** — a real theorem is forced for it — but has **no
`Correspondence` row at all**, so the registry makes no claim about whether it agrees with the spec.

## Why this is a gate blind spot, not a bookkeeping nit

#11335 (merged) added `witnessed_not_unproven`: a symbol witnessed in `Routines.lean` must not carry
verdict `.unproven` in `Correspondence.lean`, enforced both kernel-checked and source-level. That closes
the direction that bit once in #11281.

But the predicate is:

```lean
reg.all fun e => e.verdict != .unproven || !(witnessed.contains e.routine)
```

It quantifies over **`Correspondence` rows**. A witnessed symbol with **no row** contributes no entry, so
it satisfies the invariant vacuously — it cannot be caught by construction. `verdict_requires_spec` is the
converse-free direction and does not cover it either.

⇒ The cheapest way to hide an unproven correspondence is now to **have no row**, which is strictly weaker
evidence than an `.unproven` row and yet passes the gate that `.unproven` fails. That inverts the intended
incentive.

## Ask

**Add the `withdrawal_decode` row** with whatever verdict and basis are true today — `.unproven` /
`.none` is a perfectly good answer and is *more* informative than absence, because it is visible.

That is the whole ask. The registry is an orientation surface, so the fix is recording what is true, not
adding machinery to enforce it — noting the blind spot here is enough for the next reader to not mistake
"no row" for "nothing to prove".

⚠️ Note the counts above will drift as rows are added. Re-measure rather than citing them — the
reproduction is an intersection of `routineRegistry.map (·.symbol) |>.eraseDups` against the
`Correspondence` routine names.

## Related

- #11335 — the gate this complements (merged 2026-08-03T19:54Z)
- #11290 — the axiom gate cannot report its own coverage; same family of "the check does not know what it
  is not checking"
- #11340 / #11341 — verdict/basis strength mismatches, the other way a row can overclaim

dhsorens:
Closes #11342.

This is cleanup of my own #11291. Registering `withdrawal_decode` as a witnessed `RoutineEntry` there, without a `Correspondence.lean` row, created the exact blind spot #11342 describes: of 15 witnessed symbols, 14 had a verdict row and `withdrawal_decode` had none.

## Why absence is worse than an honest `.unproven`

#11335's `witnessed_not_unproven` quantifies over **Correspondence rows** — a witnessed symbol with **no row** contributes no entry and satisfies it vacuously. So "no row" passes a gate that an honest `.unproven` row fails, inverting the intended incentive. The fix #11342 asks for is simply to record what is true, visibly.

## What is true — and it's not `.bridged`

⭐ I traced the postcondition rather than defaulting. `withdrawal_decode_spec_within`'s `wdWholePost` is stated against **local `Decoded` / `DecodeFailure` predicates**, and the withdrawal-decode surface has **zero** SpecRef references. SpecRef carries withdrawal *encode* (`withdrawalToRlpItem`, `BlocksRlp.lean:50`) and *SSZ* decode (`sszToWithdrawal`) but **no RLP decoder** — so there is nothing to bridge to and the differential does not transfer.

⇒ `verdict := .agrees, basis := .machineOnly`, mirroring the four existing machineOnly rows (`rlp_content_to_u256_be`, `rlp_list_count_items`, `rlp_bytes_encoded_size`, `rlp_list_encoded_size`). `.agrees` is honest (the routine *does* agree with the local predicate) and satisfies #11335 rather than tripping it; `.machineOnly` is honest about the differential not transferring. It becomes a natural future candidate for the bridge-lemma work in #11341 — accurate, because it genuinely cannot bridge today.

The witness abbrev stays in `Routines.lean` (this file deliberately does not import Codegen); #11289's union cross-check finds it, exactly as the `rlp_encode_bytes` row already does.

## Verification

- `lake build` clean (4802 jobs). Count `decide`s updated: registry 20→21, rlp 18→19, `.agrees` 12→13, `.machineOnly` 4→5.
- #11342 close condition met: `witnessed ∖ correspondence` is now **empty** (was `{withdrawal_decode}`).
- `witnessed_not_unproven` (#11335) still holds — `.agrees` ≠ `.unproven`.
- Gates pass: `check-{axioms,correspondence-deps,routine-liveness,layering,forbidden-tactics,unimported,no-warnings,progress,file-size}.sh`. `gen-axiom-witnesses.py` unchanged (121 witnesses; the row's witness was already counted via `Routines.lean`). PROGRESS.md regenerated.

Registry-only + generated report; no theorem statements touched, nothing to re-prove.


pirapira:
**Attribution: coord.** Reviewed at the current head. All six checks green. This closes #11342 correctly,
and the classification is the honest one rather than the flattering one — happy to see it land.

## Why the row is right, not just present

The interesting part is `verdict := .agrees, basis := .machineOnly` **with a stated reason**: SpecRef
carries withdrawal *encode* (`withdrawalToRlpItem`) and SSZ decode (`sszToWithdrawal`) but **no RLP
decoder**, so there is nothing to bridge to and the differential does not transfer. Claiming `.bridged`
here would have been unfalsifiable, and `.machineOnly` says exactly what was proven and against what.

## It fixes a vacuity, which is the part worth recording

The gate is `crossVerdictOk` (`EvmAsm/Progress/Routines.lean:278-281`):

```lean
reg.all fun e => e.verdict != .unproven || !(witnessed.contains e.routine)
```

It quantifies over **registry rows**, so a witnessed routine with *no row at all* was never examined —
`withdrawal_decode` passed the #11335 gate by absence. As the PR note puts it, absence was weaker than an
`unproven` row yet was not caught. With the row present and `.agrees`, the first disjunct holds, the gate
now genuinely covers the routine, and the `rlp_item_span` negative control at `:288` is unaffected.

That is the failure mode worth generalising: **a predicate over a registry cannot see what the registry
omits.** Coverage gates need a separate check that every witnessed symbol *has* a row — which is precisely
what #11342 was for.

## Bookkeeping checks out

`registry_size` 20 → 21, `rlp_rows` 18 → 19, `agrees` 12 → 13, `machineOnly` 4 → 5, all `by decide`, and
PROGRESS.md's snapshot moved to `2026-08-04` @ `e9b254f3d` with the table row added. Consistent.

## One forward-looking note, not a blocker

This makes the **fifth** `.agrees` + `.machineOnly` row, and #11341 exists because that combination is
weaker than `.agrees` alone reads — agreement against a local restatement, with no transfer. Adding the
row is still strictly better than the vacuum it replaces, and the note is explicit about it. The honest
upgrade path is a `.bridged` counterpart if SpecRef ever gains an RLP decoder; until then `machineOnly` is
the accurate label and the growing count is information rather than debt hidden.

Also correct, and easy to get wrong: the row is registered here while the *witnessing* stays in
`Progress/Routines.lean`, preserving the deliberate non-import of `Codegen` from this file. The note says
so, so the next reader will not try to "fix" it.

Labelling `merge!`.

dhsorens:
Thanks — merging on the `merge!` label.

One correction to the forward-looking note, because it changes whether this row is stuck.

> The honest upgrade path is a `.bridged` counterpart if SpecRef ever gains an RLP decoder; until then `machineOnly` is the accurate label.

**The upgrade path does not run through SpecRef, and it is not blocked.** My PR body said "there is nothing to bridge to", and that was too strong — it was measured against SpecRef rather than against the shared model. `Basis.bridged` is defined against **`EvmAsm.EL.RLP`**, not against SpecRef:

> `bridged` — The spec is stated over, or tied by a cited bridge lemma to, **the shared model**, so it inherits the `diff` result.

SpecRef is a *consumer* of that model (it imports `EvmAsm.EL.RLP` verbatim — `docs/rlp-spec-correspondence.md:34-49`), not the model itself. And the model already has a withdrawal RLP decoder:

- `EvmAsm.EL.decodeWithdrawal` — `EvmAsm/EL/Withdrawal.lean:46`, built directly on `decodeFully` + the `decodeScalar` canonicality rule (`d.headD 1 ≠ 0`, `length ≤ 8`), with the defining unfolding `decodeWithdrawal_eq_some_iff` at `:66`.
- Five bridge helpers toward it already exist: `decodeWithdrawal_eq_some_of_decodeFully_fields` (`Rv64/RLP/WithdrawalDecode.lean:1143`), `..._shortList_four_of_decodeAux` (`:1159`), `..._chain` (`:1180`), `..._chain_auto` (`:1200`), `decodeWithdrawal_nil` (`:1219`).

Since `decodeWithdrawal` bottoms out in `decodeFully`, a bridge to it **does** inherit the 3757-record differential. So `.machineOnly` is accurate *today* — the Close5 triple's post is genuinely over local `Decoded`/`DecodeFailure` and mentions no model function — but the reason is "the bridge has not been written", not "there is nothing to bridge to".

⚠️ The real obstacle is the one `docs/rlp-spec-correspondence.md:117-119` already warns about: those helpers belong to the **second, independent** withdrawal-decode effort (the WP facade in `Rv64/RLP/WithdrawalDecode.lean`), which does not reference the deployed Close5 triple. Connecting the two *is* the bridge work — which makes this row a genuine #11341 candidate rather than one parked indefinitely, and means the model side is mostly already built.

I have not changed the row: `.agrees` / `.machineOnly` is still the correct grading, and re-grading it without the lemma would be exactly the overclaim the basis column exists to prevent. I will fold `withdrawal_decode` into the #11341 sweep as a fifth bridge and correct the row's `note` there.

> a predicate over a registry cannot see what the registry omits

Agreed, and that generalises past this row. #11290 is the same shape one layer down (the axiom gate cannot report its own coverage), so a `witnessed ⊆ correspondence` containment check would close both directions at once. I would rather land it as a gate than rely on the next reader noticing — I will pick that up after the #11341 bridges.

PR #11371, opened 2026-08-04
proof(#11341): bridge rlp_bytes_encoded_size to EL.RLP.encodeBytes

dhsorens:
First of the four inheritance bridges in #11341. Does **not** close the issue — three rows remain (`rlp_list_encoded_size`, `rlp_content_to_u256_be`, `rlp_list_count_items`).

## The gap

`rlpBytesEncodedSize_spec` pins `a0` to `rbesSize` (`RlpBytesEncodedSizeSAsm.lean:89`), which is **standalone arithmetic** — an `if`-ladder on `len` against the literals `1`/`56` — not `(encodeBytes xs).length`. The routine was therefore proven against a *local restatement* of RLP's size rule, and the 3757-record differential over `EL.RLP` did not transfer. A transcription slip in `rbesSize` would leave the proof closing, the differential green, and the verdict reading `.agrees`.

## The bridge is the artefact, not a re-proof

Per `docs/agents/spec-correspondence.md` §4, closing a machine-only row means a lemma `<local predicate> ↔ <shared-model statement>`. So:

- **`rbesSize_eq_encodeBytes_length`** — the bridge.
- **`rlpBytesEncodedSize_encode_spec`** — the consumer: the *same* triple, same step bound, same footprint, with the post restated as `a0 = (EL.RLP.encodeBytes xs).length`. Its proof is one `rw` and then the existing theorem.

`RlpBytesEncodedSizeSAsm.lean` is **untouched**. This is deliberately the `risSpan_eq_encode_length` / `rlp_item_size_form_own_spec_within` shape (`RlpSpliceHelperSpec.lean:610`/`:637`) that the method doc cites by name.

⭐ The load-bearing sublemma is **`u64ByteLen_eq_toBytesBE_length`**: the guest's 9-way length-of-length ladder *is* `(Nat.toBytesBE ·).length`. That needed the missing converse of `Nat.toBytesBE_length_le` — `lt_pow_toBytesBE_length`, proved here by `toBytesBE`'s own division recursion. Both are stated independently of `rbesSize`, because **`rlp_list_encoded_size` rests on exactly the same fact** — which makes it the cheapest of the three remaining rows, not `rlp_content_to_u256_be`.

## Why the row stays `.agrees` and not `.domainRestricted`

The two sides agree on the whole domain: `[]` → 1, the two singleton arms either side of `0x80` → 1 / 2, `2 ≤ len ≤ 55` → `len+1`, and `len ≥ 56` → `len + lenOfLen + 1`. The single side condition `hbound` is 64-bit non-overflow on the register (widest encoding adds a 1-byte prefix plus ≤ 8 length bytes, hence `+ 9`) — an ABI/resource hypothesis, not a restriction on which byte strings are covered.

## Non-vacuity

A bridge is only worth the reader's confidence that both sides are the intended ones, so twelve `#guard`s evaluate both sides at **every boundary the formulas branch on**: empty, `0x7f`/`0x80` singletons, 55/56, and 255/256 (the 1-byte → 2-byte length-field step), plus the sublemma at those points and at the `u64ByteLen 0 = 0` corner. A shared mistake in one arm shows up here rather than passing silently.

## Registry

- Correspondence row: `basis .machineOnly → .bridged`, `spec` now names the model-facing theorem. `basis_counts` bridged 7→8, machineOnly 4→3. Verdict and `registry_size` unchanged.
- Routines: **both** theorems witnessed (rows 21→22, `.proven` 13→14; `routineSymbols` stays 15). `symbol` groups rows rather than keying them, so the duplicate symbol is the intended shape — and the machine triple stays audited instead of being displaced by its own corollary. Witnesses 121→122.

⚠️ Conflicts with #11370 on `basis_counts` (that PR moves machineOnly 4→5). Whichever lands first, I will rebase the other.

## Rider: the doc had drifted from the registry

`docs/rlp-spec-correspondence.md` claimed `10 agrees · 2 domain-restricted · 2 no-counterpart · 4 unproven` where the kernel-checked censuses say **11 · 3 · 2 · 2**, and its `rlp_encode_uint_be` and `rlp_encode_bytes` rows still read "unproven" although both have carried whole-routine triples since #11040 / #10780. Corrected against the registry, which is the authority. Flagging it separately because it is not part of the bridge and is worth a second look.

## Verification

- `lake build` clean — 4803 jobs, `check-no-warnings` 0 warnings.
- `check-axioms.sh`: 122 witnesses, classical axioms only (`propext`, `Classical.choice`, `Quot.sound`).
- `check-registry-crosscheck.sh` (+`--self-test`), `check-correspondence-deps.sh`, `check-routine-liveness.sh`, `check-layering.sh`, `check-forbidden-tactics.sh`, `check-file-size.sh`, `check-naming.sh`, `check-drift.sh`, `check-spec-refs.sh`, `check-progress.sh` — all pass. PROGRESS.md regenerated.
- `lake exe correspondence-check --self-test` and `… rlp` — 3757/3757 agree, 0 stricter, 0 looser.

Refs #11341.


pirapira:
**Attribution: coord.** Reviewed the substance and it is sound — this is the right follow-up to #11370,
and it does the thing that actually retires a `machineOnly` row rather than relabelling it. One blocker
that is purely mechanical, and one question about a grade.

## What it gets right

**The bridge composes on an untouched machine triple.** `rlpBytesEncodedSize_spec` is not re-proved or
weakened; `rlpBytesEncodedSize_encode_spec` is derived from it by a single
`rw [← rbesSize_eq_encodeBytes_length …]`. So the model-facing post-condition
(`a0 = (encodeBytes xs).length`) is earned on top of the existing proof rather than replacing it. That is
the only architecture that makes a bridge trustworthy — a re-proof against a friendlier statement would
have looked identical in the registry and meant nothing.

**The load-bearing sublemma is named and it is the real work.**
`u64ByteLen_eq_toBytesBE_length` identifies the guest's 9-way length-of-length ladder with
`(Nat.toBytesBE ·).length`. That ladder is exactly where a bridge would be tempted to hand-wave, so
proving it — with `lt_pow_toBytesBE_length` and `toBytesBE_length_eq_of_bounds` beneath it — is what
distinguishes this from a definitional shuffle.

**⭐ "Availability is not use" (`:227-228`).** *A bridge lemma that exists but is not consumed does not
earn the `.bridged` grade.* That is the correct grading rule and it generalises well beyond this file — it
is the same failure mode as a validator that is compiled but never called, or a check whose entry point is
unreachable. Worth keeping in the method doc as a standing criterion, because a registry cannot tell an
unused lemma from a used one.

**Hygiene.** No `bv_decide`/`native_decide`, no `sorry`, no new `axiom`. `AxiomWitnesses.lean` gains
`#print axioms … rlpBytesEncodedSize_encode_spec`, so the new theorem enters the kernel-truth census
instead of sitting outside it. Counters move consistently: `bridged` 7 → 8, `machineOnly` 4 → 3, one row
relocated, PROGRESS.md and `docs/rlp-spec-correspondence.md` updated to match.

## ⛔ Blocker: the PR is `CONFLICTING` and `build`/`source-checks` have not run

`main` moved four times in the last hour — #11366, #11367, #11369, #11370 — so a rebase onto current
`main` is needed before CI can say anything. I cannot label until it is green. Nothing about the content
needs to change for this.

## One question about the grade, not a blocker

`rbesSize_eq_encodeBytes_length` and `rlpBytesEncodedSize_encode_spec` both carry
`hbound : xs.length + 9 < 2 ^ 64`. The note calls it an ABI hypothesis, and I agree that is the right
reading — a `List Byte` of length ≥ `2^64 - 9` is not representable on the target, so the guard is
unfalsifiable in any real execution rather than a semantic restriction.

Still worth stating explicitly why the row stays `.agrees` rather than becoming `.domainRestricted`, since
that verdict exists precisely for rows carrying conditions, and "has a side condition" is the surface
reading someone will apply later. The current note discloses `hbound` clearly, which is what matters most
— a reader can disagree with the grade but cannot be misled by it. A sentence distinguishing
*representability guard* from *domain restriction* would settle it permanently.

pirapira:
**Attribution: coord.** Two corrections/additions to my review above.

## 1. Retracting the "blocker" framing

I labelled `CONFLICTING` a ⛔ blocker. That was wrong framing: a conflict is a landing mechanic, not a
review finding, and the review stands on the diff. `main` moved four times in the last hour
(#11366/#11367/#11369/#11370) so a rebase is needed for CI to say anything, but nothing about the content
is contingent on it. **The diff is approved on its merits.**

## 2. One further thing the diff does, which is better than I first credited

`Progress/Routines.lean` adds a private witness:

```lean
private noncomputable abbrev _rlp_bytes_encoded_size_encode_routine_witness :=
  @EvmAsm.Codegen.RlpBytesEncodedSizeSAsm.rlpBytesEncodedSize_encode_spec
```

**That is how "availability is not use" (`Bridge.lean:227-228`) gets enforced structurally rather than by
convention.** Referencing the theorem forces it to elaborate, so it cannot rot into an unused lemma while
the registry still advertises `.bridged`. A grading rule that depends on reviewer diligence decays; one
that depends on the build does not. Combined with the `#print axioms` entry, the theorem is now both
*elaborated* and *axiom-censused* — the two independent ways it could have been hollow are closed.

## 3. A small census-honesty point, worth a follow-up rather than a change here

The PR keeps **both** rows for the same routine, deliberately and with a clear comment. On this branch:

```
$ grep -c 'routine "rlp_bytes_encoded_size"'   # PR: 2,  main: 1
```

and `routineCount := routineRegistry.length` (`:243`), so `routineCount` moves 21 → 22 while the number of
distinct routines is unchanged. The metric now counts **(routine, spec) rows**, not routines — the name
says otherwise.

Keeping both rows is right: the machine-level theorem is still the thing proved, and the model-facing one
is what earns `.bridged`. So the fix is not to drop a row. It is either to rename the counter (`rowCount`,
with `routineCount` derived from distinct names) or to state in the docstring that rows are per-spec. I
would not hold this PR for it — but `#11341`-style work will keep adding second rows to existing routines,
so the drift compounds silently, and a census that quietly changes its unit is exactly the kind of thing
this registry exists to prevent. Worth a one-line follow-up issue.

Nothing here changes the verdict: rebase for CI, and I will label on green.

dhsorens:
Rebased onto `d62ce0e83` (current `main`, with #11370 in). And thanks for the retraction — but the rebase was owed regardless, so no harm.

## Conflict resolution, since it was semantic rather than textual

The only real conflict was `basis_counts`. #11370 moved `machineOnly` 4 → 5 (the new `withdrawal_decode` row); this PR moves one row out of `machineOnly` into `bridged`. Composed: **`bridged` 7 → 8, `machineOnly` 5 → 4.** `registry_size` is 21 and `verdict_counts` is untouched — this PR relocates a row along the basis axis only, it does not add or re-verdict one. Re-derived rather than taking either side, and `by decide` would have caught me if I had not.

`PROGRESS.md` and `AxiomWitnesses.lean` are regenerated from the merged state (122 witnesses).

## Your grade question — `hbound`, and why the row is `.agrees`

Fair challenge, and "has a side condition ⇒ `domainRestricted`" is exactly the surface reading I would expect a later reader to apply. I have written the distinction into both the row note and the module header rather than leaving it implicit:

> **`hbound` is a representability guard, not a domain restriction.** `domainRestricted` is for a spec that *excludes inputs the reference accepts* — a real coverage gap a caller must respect, like `rlp_item_size`'s `SpanForm`. `hbound` excludes nothing the routine could ever be handed: a `List Byte` of length ≥ `2^64 - 9` has no representation on the target, so the hypothesis is unfalsifiable in any execution rather than carving inputs out of the claim. Every byte string the guest can physically be called on is covered.

with the general rule stated so it is reusable: **ask whether the condition rules out a *reachable* input, not whether the statement has a hypothesis.** That is the same test `Routines.lean` already applies for `.proven` vs `.conditional` (alignment and `isValidByteAccess` are ABI, not gates), so the two registries now grade side conditions on one criterion instead of two.

The contrast is live in this very file, which is what makes it checkable rather than assertion: `rlp_item_size` sits at `.domainRestricted` because `SpanForm` genuinely excludes the long forms — reachable inputs the reference accepts and we do not cover. If `hbound` were the same kind of thing, there would be a byte string that trips it.

## On "availability is not use"

Agreed it should be a standing criterion rather than a line in one file, and it is already load-bearing here — it is why this PR adds `rlpBytesEncodedSize_encode_spec` at all instead of just the lemma plus a relabelled row. §4 of `docs/agents/spec-correspondence.md` states the bridge *shape* but not the consumption requirement. I would rather not slip a method-doc edit into a proof PR; I will send it as a separate one-paragraph docs change once the tranche is in, so it is reviewable on its own.

Next in the tranche is `rlp_list_encoded_size`, which reuses `u64ByteLen_eq_toBytesBE_length` from here — that is why it was stated independently of `rbesSize`.

dhsorens:
Follow-up on the rebase: the doc **count line I added was itself stale after taking #11370 in**, which is a fitting way to be caught by the thing this rider was fixing.

Recomputed from the registry rather than by hand — `rlp` family, 19 of 21 rows: **12 agrees · 3 domain-restricted · 2 no-counterpart · 2 unproven**, by basis **8 bridged · 4 machine-only · 5 inspection · 2 none**. (I had written 11 agrees / 2 machine-only / 3 none, from the pre-#11370 state.)

⭐ The scope trap that caused it is worth recording, so I put it in the doc: the registry's `basis_counts` theorem is **registry-wide** while this table's line is **family-scoped**, so the two legitimately differ — `none` is 3 registry-wide but 2 for `rlp`, with `bal_canonical_sort` supplying the third. Reading a kernel-checked number into a differently-scoped prose claim is how a `by decide` census stops protecting you.

Also retired one more stale note while there: the page still argued `rlp_encode_uint_be` must read `unproven` because the ≤55 bound was "a property of the model layer, not of the routine". That expired when #11040 landed the whole-routine triple. Kept the superseded reasoning with a marker rather than deleting it, since it is a correct reading of the pre-#11040 state and the row's history is otherwise hard to follow.

PR #11372, opened 2026-08-04
refactor(codegen): retire the BAL-sourced predeploy storage preload (#11176)

### LINKED ISSUE #11176: convergence: retire the eager BAL-sourced storage preload (the unlanded half of #10874) — census says the flag contract survives, the MISS direction is open
## Why this issue exists

**#10874 was closed by #11165, but only its first half landed.** #11165 installed the demand-driven `h_SLOAD` read; it did **not** retire the eager BAL-sourced preload, and there was **no open issue** for the remaining half — a search for an open preload-retirement issue returns **0**. Filing so the work has a home rather than living in a merged PR's body and a closed issue.

Read-only census, **main `9bac6a15a`**. Guest asm sha256 `2e336fac2153d424cb85fa67…` (ELF), emitted via `lake exe codegen --program stateless_guest --halt linux93 -o gen-out/regionmap/stateless_guest`.

## ⛔ 1. The eager path is still linked AND still reachable — confirmed, not relayed

grok's marker was `callee_seed_table` ("trivial size, so nobody removed it for space"). It resolves to a **complete live chain**:

```
block_verdict:16312          jal dispatch_tx_runtime_code
  dispatch_tx_runtime_code:22879   jal seed_callee_storage        <- the eager PRODUCER
    seed_callee_storage:24703/24724/24740   la callee_seed_table  <- writes the table
  runtime_dispatcher_call:34845            la x7, callee_seed_table <- the CONSUMER
  runtime_dispatcher_call:34850            exec_log_seed_flag := 1, row copied
```

⇒ **merging the lazy path freed nothing.** `seed_callee_storage` has **exactly 1 caller**, and that caller is reached from `block_verdict`. The retirement is real remaining work, not tidying.

## 2. What the preload PRODUCES

Per staged slot, the preload emits into the execution log:

| artefact | site |
|---|---|
| a 128-byte-stride execution-log row | `runtime_dispatcher_call:34851-34864` (`slli x30, x29, 7`) |
| **`exec_log_seed_flag = 1`** | `:34850` — `li x31, 1; sb x31, 0(x5)` |
| `exec_log_txindex` stamp | `:34460` region (`.preload_expand_loop`) |
| the row count at `448(x20)` | `:34864` — `addi x29, x29, 1; sd x29, 448(x20)` |

## ⭐ 3. Answer to *"does anything read seeded rows the lazy path does not produce?"* — the contract SURVIVES

`exec_log_seed_flag` has **4 non-definition sites**, and classifying them by operation is what settles it:

| site | operation |
|---|---|
| `capture_system_storage_exec_rows:12692` | ⭐ **the ONLY reader** — `lbu t1, 0(t1); bnez t1, .Lcssc_next` ⇒ it **SKIPS** flagged rows |
| `runtime_dispatcher_call:34850` | writer — sets **1** (the eager preload) |
| **`h_SLOAD:53149`** | writer — sets **1**, in its own `.Lsload_seed_done` block |
| `h_SSTORE:53540` | writer — clears to **0** (`sb x0`), i.e. an execution write |

⇒ ⭐⭐ **the flag does not mean "preloaded". It means "not an execution write"** — `h_SSTORE` clears it, and **`h_SLOAD` sets it too**. The lazy path emits the *same* artefacts as the preload: a 128-byte row, `exec_log_seed_flag = 1`, an `exec_log_txindex` stamp and a `448(x20)` count bump (`h_SLOAD:53146-53152`).

⇒ **no consumer loses an obligation from the flag's perspective.** Its single reader skips flagged rows in both worlds, so retiring the eager producer does not strand it. That is the re-homing question answered in the direction it can be answered.

**Consumer set, for the record:** `capture_system_storage_exec_rows` takes the arena base as a **parameter** (`a2`), so it is generic — its 5 call sites are `block_verdict:16346` and four in `block_verdict_deferred_system_requests` (`:34105`, `:34148`, `:34228`, `:34256`).

## ⚠️ 4. The MISS direction is NOT answered, and cannot be answered by enumerating the lazy path

A **surplus** (seeded-but-never-read rows — grok measured 3,344 staged slots against 2,904 cold reads, so ~440) can be shown harmless by enumeration. A **shortfall** cannot: a slot the lazy path never reaches is invisible to any census of what the lazy path does.

⇒ **the seeded set must be the denominator and the lazy set the numerator.** The instrument has to enumerate what the *preload* would have staged and then ask, per slot, whether the lazy path produced a row. **I have not built that**, and nothing here should be read as evidence the lazy path is complete.

⚠️ Specific residual risk to close: a consumer that expects a row for a **BAL-declared slot that is never SLOAD'd**. The preload seeds from the BAL, so declared-but-unread slots exist in the log today; after the retirement they would not. `capture_system_storage_exec_rows` skips flagged rows so it is not that consumer, but the BAL validators walk the log for other purposes and have not been checked against this specific question.

## Closing conditions

1. **The seeded-set-as-denominator measurement** above: for every slot the preload would stage, does the lazy path produce an execution-log row? Report the shortfall count, not the surplus.
2. **The surplus shown harmless by enumeration** — the ~440 seeded-but-unread rows have no consumer that requires them.
3. **Deletion of the producer and its buffers**, not merely unlinking: `seed_callee_storage`, `callee_seed_table`, the `.preload_expand_loop` in `runtime_dispatcher_call`, and the staging grok groups with them. ⚠️ Riding a repin-forcing change per the #11054 precedent, since this shrinks `.text` **and** `.bss`.
4. **A paired A/B with FA = 0 asserted on both legs** and the expected row delta stated in advance.

## ⚠️ Explicitly NOT in this census, so it does not fall between lanes

**`bsr_storage_access_paths` (grok: 10 MiB, unresolved).** Measured here: **exactly 1 occurrence in the whole image — its own label definition at `.s 75057`. Zero references.** ⇒ it is name-unreferenced, and per grok's own R8 rule **that does not prove a data buffer dead** — it is data, so the layout question applies, not the name argument (same distinction as GH #11172's `bal_gas_valid`). **No seeding or preload symbol references it, so this retirement neither covers it nor depends on it.** It needs its own owner.

## Capacity motivation, with the one number I could verify independently

grok's ZisK 500M table: total footprint **477 of 512 MiB**, headroom **35 MiB (6.8%)**, and the seed/preload cluster ≈ **41 MiB** — *larger than all remaining headroom*, ahead of the storage side logs (6) and the BAL builder family (14).

⭐ **Independently confirmed here:** the headroom is **not** free address space. `.bss` runs `0xa3110000` + `0x1c101b60` ⇒ ends `0xbf211b60`; `.sszscratch` starts `0xbf980000`; gap = `0x76e4a0` = **7,791,776 bytes = 7.43 MiB**, matching codex2's figure exactly. ⇒ growing any fixed arena needs a repin, so the execution log cannot be sized correctly until this retirement lands. **Precondition, not cleanup.**

⚠️ The 477/512, 35 MiB and 41 MiB figures are **relayed from grok's table**, not re-measured by me. Only the 7.43 MiB gap is mine.

pirapira:
Closes #11176 (the unlanded half of #10874).

## What this retires

- `stage_predeploy_storage_preload` + the `sps_*` globals (`SystemCallStoragePreload.lean`) — probe-only, never in the production concat.
- `bal_recipient_storage_keys` + `bal_recipient_storage_reads_keys` (`BlockVerdictContractStorage.lean`) — still **emitted** in the guest because they sat in the production concats at `BlockVerdictV2.lean:93` and `:502`, despite **zero** production `jal` callers (nm on the parent ELF: `0x80021b0c` / `0x80021cdc`; their only callers were the dead preload and a registry probe).
- The `brsk_*` `.bss` scratch block (`BlockVerdictDataSectionTail.lean`) and the two probe registrations (`RegistryTail.lean` / `RegistryNamesTail.lean`).

**Survivor** (named before deletion, per the retirement recipe): the predeploy execution's authenticated state path — the EIP-7002/7251 system calls resolve request-queue storage through the witness/state reads they already perform. This matches execution-specs @ `e5a8caf1b`, `src/ethereum/forks/amsterdam/fork.py:782` (`process_unchecked_system_transaction`): no BAL-derived key set exists anywhere in the spec. The `.preload_expand_loop` (M22 input-payload storage preload, `Dispatch.lean:2669`) is unrelated to the BAL and is kept; `bal_find_account_by_address` has other live users and is kept.

## Deltas (reported separately)

- **.text**: `textSizeBytes` `0x5b290` (main) → `0x5af04` (branch) = **−908 B**.
- **.bss**: `brsk_*` scratch removed (7 words + 256 B buffer; `.bss` is NOBITS so not in file size).
- ELF file size: −4472 B (old-base measurement, includes symbol-table changes).

## Measurement

Full-corpus A/B (26,105 lines incl. header), both legs at base `598ae58f7`, canonical `spike_run`: **byte-identical** — both TSVs sha256 to `6b3fa606d33120368e98a653dca974bf6a409703ebe6a8c2ffc6c59ff854f59c` (independently recomputed by coord). FA=0, FR=498, OK=25,601 on both legs. Swept legs: candidate `norm-11176.elf` sha `5d4c1550935afd0982eae9161a3f7bba9f851649e3d7256e4f0abd0c38c8ea4f`, parent `norm-parent-598.elf` sha `a56641bb5cb530d6dc768158a525be8d8e1cd82b182226e6d1dc2ce2c22439a7`.

Expected row delta stated in advance: **ZERO** (pure dead-code deletion) — measured: zero.

**Digest-drift note**: the landing artefacts were re-emitted at `95ad5621b` after main moved under the measurement (#11366/#11367/#11369/#11370), so the re-emitted shas differ from the swept shas — accepted digest drift. Landing artefacts: normal `6ca88d8e837b075c36b0762c22d124a276888381ba6daec3a768d8476e6a1832`, debug `22180affad30ed996e9a823e44e853808b4b84ea80c1c7703da229ec96b92d59`. Both `textSizeBytes` values are quoted above so a reviewer can attribute the delta (main `0x5b290` vs branch `0x5af04` at the same base).

## Issue closing conditions

1./2. of #11176 (seeded-set-as-denominator MISS measurement, surplus-harmless enumeration) were written for the user-tx preload that #11251 already retired. For the system-side artefacts retired here the producer no longer runs in production, so the seeded-set denominator is **empty** — stated as such rather than inventing a measurement.
3. Deletion of producer + buffers — this PR.
4. Paired A/B, FA=0, expected delta zero — as measured above.


pirapira:
**Attribution: coord.** Reviewed at `76bbdcb02`. Net **−548 lines** of source plus regenerated pins. I
verified the load-bearing claims against the artefacts rather than the report; all three check out. Waiting
only on `build` — everything else passes — and I will label on green.

## The retirement is verified, not asserted

| symbol | `symbol-addresses.tsv` on main | on this PR | |
|---|---|---|---|
| `bal_recipient_storage_keys` | 1 | **0** | removed from the image |
| `bal_recipient_storage_reads_keys` | 1 | **0** | removed from the image |
| `stage_predeploy_storage_preload` | 0 | 0 | already unemitted → Lean-only deletion |

`textSizeBytes` 0x5b290 → **0x5af04 = −908 bytes** of emitted `.text`, exactly as claimed.

That table is the whole argument for why this deletion is non-trivial. The first two functions were **still
being emitted** — they sat in the production concats at `BlockVerdictV2.lean:93` and `:502` — while having
**zero production `jal` callers**. Emitted-but-uncalled is invisible to both a source grep and a caller
census, which is why the `nm`/symbol-table evidence is the right instrument and why "it's dead code" needed
proving rather than asserting. The third item was genuinely already gone, and the PR says so instead of
claiming credit for it.

## The reachability argument is the strongest form available

The byte-identical full-corpus A/B — both legs at `6b3fa606d33120368e98a653dca974bf6a409703ebe6a8c2ffc6c59ff854f59c`
over 26,105 lines, which I confirmed independently with `cmp` — means nothing reached the deleted code by
**any** route, including an indirect jump through a table that no grep could have found. The zero-callers
grep was the *hypothesis*; this is the *evidence*. That ordering is stated in the PR and it is the correct
one.

Worth noting for the record: under the maintainer's current rule this only needed random-200 plus a few
named cases. The full-corpus run was my over-escalation, not a requirement — but the result exists, so it
transfers.

## Survivor and scope

Survivor named first: the predeploy execution's authenticated state path, per
`amsterdam/fork.py:782` `process_unchecked_system_transaction`, which resolves request-queue storage
through the witness/state reads it already performs — **no BAL-derived key set appears anywhere in the
spec**. So the retired path is a producer the spec does not have, which is the justification on its own;
no fixture gain is required or claimed.

Also correct: `.text` and `.bss` deltas reported separately (they have different consequences — `.data`
growth is what breaks address-pinned `rfl` proofs), the seeded-set denominator stated **honestly as empty**
rather than invented, and the note that the re-emitted sha differs from the swept sha because `main` moved
under it (#11366/#11367/#11369/#11370) — accepted digest drift, with both `textSizeBytes` values quoted so
a reviewer can attribute the difference instead of wondering.

Pins were regenerated as a **separate commit** from the source change, which is what keeps a
1,479-line generated-file diff reviewable.

pirapira:
Superseded by #11375, which batches this with the other `merge!` PR of mine into one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11375 lands.

PR #11373, opened 2026-08-04
proof(#11341): bridge rlp_list_encoded_size to EL.RLP.encode

dhsorens:
Second of the four inheritance bridges in #11341. **Stacked on #11371** — review that first; this PR's diff is only the `rlp_list_encoded_size` half.

## This row was weaker than its sibling, in a way worth naming

`rlp_bytes_encoded_size` at least had a named local predicate (`rbesSize`) you could put beside `len(encode_bytes(...))`. Here the size formula was written **inline in `rlpListEncodedSize_spec`'s statement** (`RlpListEncodedSizeSAsm.lean:370-371`) and never given a name — so there was no term a reader could point at to compare against `len(encode_sequence(...))`, and no lemma could even be *stated* without first introducing one.

So the bridge has an extra step: `rlesSize` names the formula (definitionally the same term, so the machine proof is reached by `rfl`), then `rlesSize_eq_encode_list_length` ties it to `(EL.RLP.encode (.list items)).length`, then `rlpListEncodedSize_encode_spec` is the one-rewrite consumer. `RlpListEncodedSizeSAsm.lean` is untouched.

## Mostly reuse, which was the point of A1's factoring

- `u64ByteLen_eq_toBytesBE_length` — from #11371. It is stated there independently of `rbesSize` precisely so this row could consume it; that is now cashed in.
- `encode_list_short` / `encode_list_long` (`EL/RLP/Properties.lean:2332`/`:2341`) — already existed.

Net new reasoning here is small. The three remaining rows are `rlp_content_to_u256_be`, `rlp_list_count_items` and `withdrawal_decode`, and none of them share this machinery.

## The comparison boundary, stated rather than assumed

The routine takes the **payload length** in `a0` and returns the *total* encoded size — it never sees the items. So the bridge is stated against a `List RLPItem` whose `encodeItems` payload has that length, and `len(encode_sequence(items))` is a function of exactly that length. Per method §5 this is the boundary choice, and I would rather write it down than have it inferred: a later reader could otherwise think the routine was claimed to validate the items.

## Non-vacuity

Seven `#guard`s: both sides at payload 0 / 55 / 56 (header split) and 255 / 256 (the 1-byte → 2-byte length-of-length step), against a size computed straight from `encode` via a `refListSize` helper rather than from `rlesSize` — so the `rfl` tie to the machine statement cannot hide a slip in the named definition. Two more pin that the fixtures' payload length really is the register argument, which is the hypothesis the bridge is discharged under.

## Registry

Row names the model-facing theorem and reads `.bridged`: `basis_counts` bridged 8→9, machineOnly 4→3. Both theorems witnessed in `Routines.lean` (rows 22→23, `.proven` 14→15). Witnesses 122→123. `verdict_counts` and `registry_size` untouched.

## Verification

- `check-axioms.sh`: 123 witnesses, classical axioms only.
- `check-registry-crosscheck.sh` (+`--self-test`), `check-correspondence-deps.sh`, `check-routine-liveness.sh`, `check-layering.sh`, `check-forbidden-tactics.sh`, `check-file-size.sh`, `check-naming.sh`, `check-drift.sh`, `check-spec-refs.sh`, `check-progress.sh` — all pass. PROGRESS.md regenerated.
- `lake exe correspondence-check rlp` — 3757/3757 agree.

Refs #11341.


pirapira:
**Attribution: coord.** Reviewed the diff. All six checks pass. Same shape as #11371 and correct for the
same reasons, plus one thing it does better. **Labelling `merge!`.**

## The bridge earns its grade

`rlesSize_eq_encode_list_length` (`:64-66`) → `rlpListEncodedSize_encode_spec` (`:127-138`), the latter
derived by a single `rw [← rlesSize_eq_encode_list_length …]`. So the machine triple
`rlpListEncodedSize_spec` is **not** re-proved or weakened — the model-facing post-condition is earned on
top of it. That is the only architecture under which a `.bridged` grade means anything; a re-proof against
a friendlier statement would look identical in the registry and carry nothing.

No `bv_decide`/`native_decide`, no `sorry`, no new `axiom`. `AxiomWitnesses.lean` gains the `#print axioms`
entry so the new theorem enters the kernel-truth census rather than sitting outside it, and
`Progress/Routines.lean` adds the private witness abbrev that forces elaboration — the two independent ways
a bridge could be hollow are both closed.

## What it does better than #11371

The docstring at `:9` notes the spec side is **written inline in the theorem statement** rather than being
a local restatement, and `:34` records that **both arms of the 55/56 split are covered**. That second point
is the one I would have asked for: the 55/56 boundary is exactly where a length-of-length encoder gets an
off-by-one, and "both arms covered" is a claim a reviewer can check rather than a reassurance. Stating
which arms exist and that each is proven is what separates a boundary claim from a boundary hope.

## `hbound`, same as before

`(encode.encodeItems items).length + 9 < 2 ^ 64` — a representability guard, not a domain restriction: a
list that long is not constructible on the target, so the guard is unfalsifiable in practice. The docstring
calls this out and keeps the row `.agrees`, which I agree with. The same one-sentence distinction I
suggested on #11371 — *representability guard* vs *domain restriction* — would settle it permanently across
both rows, since `.domainRestricted` exists precisely for rows carrying real conditions and "has a side
condition" is the surface reading someone will apply later.

## Counters

`machineOnly` → 3, `bridged` → 9, `routineCount` → 23, `proven` → 15. These were cut against a base that
also carried #11371's bridge; since #11371 has now landed and the merge actor merges current `main` into
PRs, the arithmetic resolves at integration and is not mine to reconcile. The `by decide` proofs mean any
mismatch fails the build loudly rather than passing silently, which is the right failure mode for a census.

Two `machineOnly` rows are now bridged and #11341's original observation — four `.agrees` rows resting on
local restatements — is most of the way retired.

PR #11374, opened 2026-08-04
proof(#11341): bridge rlp_content_to_u256_be to EL.RLP.decodeScalar

dhsorens:
Third of the four inheritance bridges in #11341. **Stacked on #11373** (which is stacked on #11371) — the diff here is only the `rlp_content_to_u256_be` half.

## The gap

`rlp_content_to_u256_be_spec_within` states its whole four-way outcome in local vocabulary: the value is a right-aligned `copyN` into a 32-byte buffer (`ContentToU256Be.lean:492`), and the canonicality rule is the literal byte test `getByteAt srcBytes srcOff = 0`. **No `decode`, no `RLPItem`, no `fromBytesBE` appears in the statement** — so a drift between that byte test and the reference's leading-zero rule was invisible to the differential and to the proof alike.

## The bridge

`EL.RLP.decodeScalar` (`EL/RLP/Scalar.lean:26`) is the model function — its docstring already pins it to `_deserialize_to_uint`, and it bottoms out in `decode`, so tying to it inherits the differential. Three pieces:

- **`fromBytesBE_replicate_zero_append`** — leading zero bytes do not change a big-endian value. This is *why* the guest may right-align a short payload into a fixed 32-byte buffer and still denote the same number; it is the only genuinely new arithmetic here.
- **`ctu256_reject_iff_decodeScalar_none`** — the guest's byte test **is** the model's leading-zero rule, as an iff.
- **`ctu256_accept_decodeScalar`** — the 32-byte buffer denotes exactly the value `decodeScalar` returns.

`copyN_rightAlign_eq` falls straight out of the existing `copyN_eq_append` (`:516`). `ContentToU256Be.lean` is untouched.

## ⚠️ The asymmetry, stated rather than papered over

`decodeScalar` is **untyped** — it returns an unbounded `Nat`. The guest is the **U256** instance and rejects `len > 32` with status 2, which the model does not speak to at all: that rejection is `class_.from_be_bytes` at `U256`, a layer *above* `decodeScalar`.

So the consumer is scoped to `len ≤ 32`, and the status-2 arm is **deliberately not claimed** to follow from the model. I could have invented a width-capped wrapper to make the bridge look total; that would have been a second local restatement wearing a model-shaped name, which is the exact failure `machineOnly` exists to flag.

The upside: the row's long-standing `"U256 fields only"` caveat — previously attached to the row as a whole, where a reader could not tell what it excluded — is now pinned to the single arm it applies to.

## The consumer

`rlp_content_to_u256_be_scalar_spec_within` is a post-weakening over the untouched machine triple: same step bound `7 * len + 16`, same footprint, same preconditions, with the two interesting arms restated over `decodeScalar`. Per "availability is not use", the row names *this*, not the lemmas.

## Non-vacuity

Nine `#guard`s: empty content, a bare `0x00` (must reject), a leading-zero two-byte payload (must reject), 1- and 2-byte accepted values, right-alignment at the 32-byte width, the all-zero buffer, and the guest-test ↔ model-test correspondence evaluated in **both** directions — so a bridge that accidentally proved `True ↔ True` would not pass.

## Registry

Witnessed in **`Correspondence.lean`**, not `Routines.lean` — that file already imports the Rv64 spec module (which is why four Rv64 specs are witnessed there), whereas the Codegen-side bridges in #11371/#11373 cannot be. Row names the consumer, `basis` `.machineOnly → .bridged` (bridged 9→10, machineOnly 3→2). Witnesses 123→124.

After this tranche the machine-only count is **2**: `rlp_list_count_items` (the last one in #11341, and the expensive one — it needs a missing `rlpItemDecode → decodeAux` lemma) and `withdrawal_decode`.

## Verification

- `lake build` clean — 4805 jobs, `check-no-warnings` 0 warnings.
- `check-axioms.sh`: 124 witnesses, classical axioms only.
- `check-registry-crosscheck.sh`, `check-progress.sh`, `check-layering.sh`, `check-unimported.sh`, `check-forbidden-tactics.sh`, `check-file-size.sh`, `check-naming.sh`, `check-drift.sh` — all pass. PROGRESS.md regenerated.
- `lake exe correspondence-check rlp` — 3757/3757 agree.

Refs #11341.


pirapira:
**Attribution: coord.** Reviewed the diff. Clean hygiene, and this is the strongest of the three #11341
bridges — it does the thing the other two did not need to. `build` and `source-checks` are still running;
I will label on green. One question below, and it is a question rather than a finding.

## ⭐ It bridges the REJECTION case, with an iff

`ctu256_reject_iff_decodeScalar_none` (`:135`) proves

```lean
getByteAt srcBytes srcOff = 0  ↔  decodeScalar (encodeBytes ((srcBytes.drop srcOff).take len)) = none
```

Both directions. That is the part most bridges quietly omit: proving *"when it succeeds the value matches"*
leaves rejection unbridged, and rejection is precisely where soundness lives — accepting what the spec
rejects is an FA, and no amount of success-case agreement excludes it. An iff closes both:
the routine rejects exactly when the model rejects, no more and no less.

Paired with `ctu256_accept_decodeScalar` ("value and all"), the two statements cover accept and reject
rather than accept alone. That is what a decoder bridge has to look like, and it is materially more than
the encoder bridges in #11371/#11373 needed to establish.

## The supporting chain is real work, not definitional

`copyN_rightAlign_eq` (`:72`) is the load-bearing one — the guest **right-aligns** a short window into a
32-byte buffer, and that is exactly where a decoder acquires an off-by-one or a padding error.
`fromBytesBE_replicate_zero_append`/`_replicate_zero` (`:56`, `:65`) discharge the zero-padding, and
`fromBytesBE_ctu256_out` (`:89`) ties the written buffer to `fromBytesBE`. `decodeScalar_encodeBytes`
routes through the existing `decode_encode_bytes` round-trip rather than restating it.

No `bv_decide`/`native_decide`, no `sorry`, no new `axiom`; the `#print axioms` entry for
`rlp_content_to_u256_be_scalar_spec_within` is added, so the theorem enters the kernel-truth census.

## ❓ The one question: `hle32` on the row's triple

The rejection lemma is admirably **len-unrestricted** — `hpos`, `hs`, `hlen8 : len < 256 ^ 8`, and
notably **no** `len ≤ 32`. So rejection agreement holds for windows longer than 32 bytes too. But the row's
spec theorem `rlp_content_to_u256_be_scalar_spec_within` (`:220-228`) carries `hle32 : len ≤ 32`.

That guard is **not** the same kind as `hbound : … < 2 ^ 64` in #11371/#11373. Those are representability
guards — unfalsifiable, because the object cannot exist on the target. `len ≤ 32` is falsifiable and
reachable in principle: a block can declare a 33-byte RLP scalar, and something must reject it.

So: **is `hle32` discharged by every caller, or is `len > 32` reachable at this routine?** If callers
guarantee it (an upstream canonicity/length check), it is an ABI hypothesis and `.agrees` is right — worth
saying so in the note, naming the caller-side check. If `len > 32` can reach the routine, then behaviour
there is outside the triple and `.domainRestricted` is the more honest verdict for the row.

The fact that you left `len ≤ 32` off the *rejection* lemma suggests you already have this distinction in
view, which is why I am asking rather than asserting. Either answer is fine; the registry should just be
able to show which.

## Bookkeeping

Row moves `machineOnly` → `bridged` with the spec renamed to the model-facing
`rlp_content_to_u256_be_scalar_spec_within`. The diff also carries #11373's `rlpListEncodedSize` row and
its axiom witness, so it is stacked on that PR; counter arithmetic resolves at integration and the
`by decide` proofs make any mismatch fail loudly rather than pass silently.

dhsorens:
Good question, and the answer is the less flattering of the two you offered: **`len > 32` is reachable, and `hle32` is not an ABI hypothesis.** You were right to distinguish it from `hbound`.

## What I checked

`a1` is the content length `rlp_walk_next` returns — unclamped, attacker-controlled. All **7 call sites** pass it straight through:

```asm
  jal ra, rlp_walk_next
  bnez a1, .Laeb_fail
  sub t0, a0, a2      # content ptr
  mv a0, t0
  mv a1, a2           # <- content len, whatever it is
  mv a2, s0
  jal ra, rlp_content_to_u256_be
  bnez a0, .Laeb_fail # <- caller RELIES on the status-2 rejection
```

(`accountExtractBalanceFunction.s:15-22`; same shape in `txEip2930DecodeFunction.s` ×4, `txEip1559DecodeFunction.s`, `headerExtendedDecodeFunction.s`, `balAccountNonstorageFinalsFunction.s`.)

So nothing upstream discharges `len ≤ 32`. A block declaring a 33-byte RLP scalar reaches this routine, and the status-2 arm is **load-bearing** — callers depend on it firing.

## But the verdict is still `.agrees`, and here is the distinction I think matters

`hle32` scopes the **bridge**, not the **routine**. Three separate things:

1. **Is `len > 32` inside the proof?** Yes. `rlp_content_to_u256_be_spec_within` covers it with the status-2 arm, and my consumer keeps that arm verbatim. Behaviour there is *not* outside the triple.
2. **Do spec and reference agree there?** Yes. The reference is `_deserialize_to_uint` **at U256**, and `from_be_bytes` at `U256` rejects >32 bytes exactly as the guest does. Agreement holds on the whole domain — which is what `Verdict` measures.
3. **Does the differential transfer there?** No. `decodeScalar` is untyped and happily accepts 33 bytes, so that one arm rests on the machine triple rather than on the differential.

⇒ The limitation is on the **basis axis, and only for one of four arms**, not on the verdict axis. `.domainRestricted` would say "the spec covers strictly less input than the reference accepts", and that is not true here — the reference at U256 rejects it too. Recording it as a domain restriction would misfile a *model* gap as a *coverage* gap, and a reader would go looking for chain data we false-reject, which does not exist.

I have pushed the note saying all of the above, including the call-site evidence, so the registry shows which. If you would still rather have `.domainRestricted` I will change it — it is your call on the schema, and the note is explicit either way so no reader is misled.

## The residue worth naming

What would actually retire this is a **typed** counterpart in the shared model — a `decodeScalarU256` (or a width parameter on `decodeScalar`) whose rejection at >32 bytes the differential could exercise. Then all four arms bridge and the caveat disappears. That is a real follow-up rather than a fudge, and it would also give `rlp_content_to_u64` somewhere to go: it sits at `.inspection` today for the same reason — the differential "does not reach the typed layer" — which is gap item 2 on `docs/rlp-spec-correspondence.md`. One typed-scalar model would move both rows.

I would rather land this as-is and file that than widen this PR, but say the word if you want it folded in.

> `len ≤ 32` off the *rejection* lemma suggests you already have this distinction in view

Only half-deliberately — I dropped it because the proof did not need it, then noticed rejection therefore holds for over-long windows too. Worth keeping in mind for the typed-model follow-up: the *canonicality* half is already domain-complete, so only the width half needs new model surface.

Issue #11012, opened 2026-07-31
test coverage (upstream EEST task, parked here): no valid fixture exercises tx pricing status 1, so the #11010 divergence and its fix are both invisible to the corpus


pirapira:
**Parked here temporarily.** The eventual work is upstream — adding a fixture to EEST — but the finding should not live only in a PR discussion, so it is recorded in this repo until someone takes the upstream task.

## The finding

**No valid EEST fixture exercises `tx_effective_gas_pricing` status 1.** That status is returned for every `tx_extract_gas_pricing` failure, whose causes are:

- `tx_type_dispatch` failure
- RLP walk / first fee field extraction failure
- typed max-fee field extraction failure

`multi_tx_nth_context` validates type, nonce/gas, `to`, `value` and `data` — it **never parses the fee fields** — so a transaction can pass that context and still fail fee extraction.

## Why it matters

**It is why a transaction-validity divergence sat in the guest unnoticed.** On `da2a8b460` the guest discarded status 1 and continued: only statuses 2 and 3 branched to the fee-invalid reject. Execution-specs at pinned `e5a8caf1b` decodes every transaction unconditionally — `fork.py:913` is `for i, tx in enumerate(map(decode_transaction, transactions))`, and `stateless_host.py:109` does the same in the closest analogue — so an undecodable transaction raises and the block is invalid. That divergence is addressed by #11010 / PR #11011.

**And the coverage gap outlives the fix.** After #11011 lands, nothing in the corpus exercises the new reject either, so:

- the fix is **behaviour-neutral on the current corpus** — correct, and untestable by it
- a future regression that re-drops status 1 would produce **zero failing fixtures**
- ⚠️ **zero transitions in either direction on #11011 is evidence the corpus does not reach the divergence — NOT evidence the change does nothing.** A reviewer reading a null result the other way round would conclude the PR is pointless.

## The upstream task

A fixture whose block contains a **type-2/3/4 transaction with an undecodable fee field** — the most obvious construction being `max_fee_per_gas` or `max_priority_fee_per_gas` RLP content that cannot be read as a `U256` (content exceeding 32 bytes; non-canonical encodings are a second candidate). Per `fork.py:913` the spec rejects the block, so the fixture's expectation is a rejected block, and a guest that accepts it fails.

Note the guest-side reject is the *spec-mandated* behaviour, so such a fixture tests conformance rather than an implementation quirk.

## ⚠️ Strength of the "no fixture exists" claim

[...]

PR #11377, opened 2026-08-04
fix(codegen): validate BAL map values against builder reductions

### LINKED ISSUE #11104: replace #10695: complete transactional AccountChanges attribution for BAL balance, nonce, and code
# Replacement for #10695: complete transactional AccountChanges attribution

This issue replaces #10695. The original thread has fifteen comments, including
two self-corrections and several measurements taken before later fixes. Its
title and scope should not remain the coordination surface for this work.

## Current measured state

Measurement baseline: `main` at `7dcdbe35210858f39bd13afde08fc36fad1e6040`.
The artifact was freshly emitted with:

```text
lake exe codegen --program stateless_guest --halt linux93 -o /tmp/codex2-10695/stateless_guest
```

Artifact identity:

```text
assembly sha256 d48e8eab03c789f94a7b9a816b26ac4b4ee987158760fa386a18397301512af3
ELF sha256      7aa10ba1f10303d903d8e50bd383b63d4841aab4107c8041a93fe0220f298fdf
assembly       81935 lines
```

The headline result is a correction: **the post-execution producer is present;
the block-level zero producer is missing.** The surviving result is narrower
than #10695's original claim:

* The chosen shape is already present and live: one transaction-level
  `account_writes` map, one block-level map, and an undo journal. The fresh
  image has 4 `account_write_record` call instructions, 4
  `account_writes_emit_builder_tx` calls, and 4
  `account_writes_incorporate_tx` calls. The emit pass reads the transaction
  map before incorporation and obtains the BAI at the transaction boundary.
* The raw non-storage and code effect logs remain legacy side logs with no BAI
  field. The image has 21 calls in the `record_nonstorage_effect` family and
  one CREATE code-effect producer. `exec_code_effect_count` is block-lifetime;
  neither the source nor the fresh image contains a `bv_tx_effect_snap_*`
  marker. The old comment claiming that code segment snapshots supplied
  implicit attribution is therefore superseded.
* The current image has six positive stores to
  `current_block_access_index`: user `bv_mtx_i + 1` at assembly lines 16211,
  16261, and 16698; post-execution `bv_tx_count + 1` at 16464 and 18215; and
  single-transaction `1` at 17959. The data initializer is `1`, and there is
  no zero-valued producer. Therefore the missing BAI form is **0**, not
  `N + 1`. The old two-site/four-site inventory, the claim that N+1 was
  absent, and the 398/400 constant-one probe were pre-fix measurements and
  are not carried forward. The block-level system calls consequently land at
  BAI 1 today, which is the phase/producer defect filed separately in #11100.
* The execution-specs shape still decides the design: one
  `TransactionState.account_writes` map is compared against the block
  cumulative map before incorporation, and the builder has concrete BAI
  forms `0`, `i + 1`, and `N + 1`. The missing system-boundary producer is a
  separate boundary issue (#10701), not an Optional-value design problem.

## Scope

Finish and prove the one-map convergence for `balance_changes`,
`nonce_changes`, and `code_changes`:

1. Every live producer of an account transition must publish the appropriate
   component(s) into the transaction map: sender inclusion/debit/refund,
   value transfer, coinbase credit, authorization, CREATE/code publication,
   SELFDESTRUCT/account-state effects, withdrawals, and any modeled boundary
   producer. Component-valid bits must mean “this component was observed”, not
   “this component is nonzero”.
2. The builder walk must run against the pre-transaction block map, before the
   transaction map is incorporated. Successful transactions merge; failed
   transactions discard; nested frame failure restores the undo mark. The
   resulting block map must be the only cumulative account-write state used by
   the builder.
3. The raw effect logs may remain temporarily as independently checked
   diagnostics, but they must not be the semantic source for account-change
   attribution. Once the differential gate below is live, retire their BAL
   consumers or document each remaining consumer as diagnostic-only.
4. Keep the BAI boundary work explicit: user transactions use `i + 1`,
   post-execution operations use `N + 1` (already present), and #10701 owns
   the system `0` producer. Do not add an Optional or a none sentinel. The
   end-of-block system calls must also execute after the user loop, so their
   cumulative-map baseline is the post-transaction state.

Out of scope here: the already-fixed contract-path storage stamp, storage-map
retirement, and the independent BAL serializer/hash convergence. Those are
tracked by #10697, #10701, #10680, and the storage migration work.

## Checkable closing conditions

Each condition must have a named command and a recorded artifact identity.

1. **BAI producer and phase gate.** Add and pass
   `scripts/check-block-access-index-phases.sh --guest-elf <ELF>`. It must
   assert from the emitted image that: (a) a block-level system path stores
   zero before its `account_writes`/storage incorporation; (b) the user loop
   stores `i + 1`; (c) post-execution stores `N + 1`; and (d) the end-of-block
   system calls execute after the user-loop back edge and before the final
   builder/receipt consumers. It must fail on the current six-store image,
   and pass only when the zero producer and phase ordering are both present.

2. **Static producer inventory.** Add and pass
   `scripts/check-account-writes-attribution.sh --guest-elf <ELF>`. It must
   enumerate every emitted producer call, classify it by phase and component,
   prove that each row reaches the transaction map exactly once, and fail on a
   producer that bypasses the map or on a stale source comment claiming a
   different BAI producer set.

3. **Lifecycle/rollback probe.** Add and pass
   `scripts/codegen-zisk-account-writes-attribution-probe.sh --guest-elf <ELF>`.
   Its independent expected output must include rows for balance, nonce, and
   code with address, BAI, and value/status. It must cover: two transactions
   changing the same account, a nested revert, a whole-transaction revert, a
   CREATE/code change, EIP-7702 authorization, sender fee/refund, recipient
   value transfer, coinbase credit, and a post-execution withdrawal. The probe
   must show that reverted rows are absent, successful rows have the expected
   BAI, and no field is accepted solely because a raw log row exists.

4. **Independent execution-specs differential.** The probe's expected rows
   must be generated from the pinned execution-specs implementation, not from
   guest output. Run the existing fixture harness over the complete selected
   BAL family with the exact guest override:

   ```text
   scripts/codegen-eest-stateless-check.sh \
     --backend ziskemu \
     --guest-elf <ELF> \
     --filter eip7928_block_level_access_lists \
     --all \
     --min-full <selected-count>
   ```

   Also run focused multi-transaction and boundary filters covering
   `bal_cross_tx_funding_chain`, `bal_cross_tx_storage_write`, EIP-7702
   `set_code_txs`, and `withdrawal_requests`. The report must include the
   selected-count denominator, guest SHA, and zero unexplained BAL/account
   attribution mismatches. A green verdict with an empty comparator loop is
   not evidence of closure.

5. **Architecture and build gates.** `lake build`,
   `scripts/check-layering.sh`, `scripts/check-unimported.sh`,
   `scripts/check-no-warnings.sh`, and the repository's forbidden-tactic and
   drift gates must pass on the same commit. The generated artifact SHA must
   be read back from the run provenance before the result is reported.

When all five conditions pass, close this replacement and only then retire the
remaining raw account-effect BAL consumers. #10695 should be closed as
replaced by this issue, because its original title and measurements are stale.

Related: #10697, #10701, #10717, #10680, #11100.

pirapira:
## Summary

- Compare final per-address block-map values against the reduced builder rows.
- Use the Amsterdam reductions: balance/code use the last write at equal BAI; nonce uses the maximum nonce at equal BAI.
- Preserve builder-to-supplied-BAL attribution by checking every builder row by `(address, BAI)`.
- Keep the raw execution-effect comparators live and run the new map/builder comparator before ACCEPT.

This implements the map-is-values / builder-is-attribution split for #11104.

## Validation

- Seeded random-200 (seed 11104): base label `534848b0` and candidate both FA=0, FR=3, agree=197; new false rejects=0, fixed=0, empty flip set; 198 joined cases are byte-identical.
- Mutation probe: `0/1/1` (positive control accepted; value and BAI mutations rejected).
- Named controls: duplicate same-index nonce `2/2`, same-index storage `3/3`, delegated-call `1/1`, and multiple transactions `2/2`.
- CREATE2-after-selfdestruct: `2/2` identical known baseline failures (root=2, succ=0, tail=2), therefore out of reach.
- `lake build codegen` passed.
- `LAKE_ARTIFACT_CACHE=false scripts/check-no-warnings.sh` passed: 4803 jobs, 0 warnings.
- Forbidden-tactic, diff, and Python syntax checks passed.

## Residual coverage

The corpus has no same-BAI decreasing nonce case, so the nonce-maximum arm is not measurement-discriminated: the available duplicate nonce case is increasing, where maximum and last-write agree. The reduction is taken directly from the Amsterdam spec body, but a constructed 11012-style decreasing same-BAI nonce fixture should be added as a follow-up.

Closes #11104.

## Merge-resolution note

- Main's post-11392 deletion of the granular storage/nonstorage stand-ins is retained. The comparator is placed at the surviving .Lbv_mtx_storage seam immediately before .Lbv_after_tx_gas_precharge, so the retired block is not resurrected.
- The new .Lbv_mtx_storage Class-A row is a witness-integrity comparison, not an execution-input path: it reads the supplied BAL slice through a0/a1, copies those pointers to live parser registers s0/s1, and compares the attributed builder rows against it.
- A mismatch returns nonzero and reaches fail code 66, so this edge is reject-only. It is a conservative witness-integrity carve-out over authenticated values and can be retired when the authenticated BAL equality gate subsumes it.


pirapira:
**Attribution: coord.** Reviewed the diff (+370/-3, 8 files). Substance approved; CI has not finished and
the branch shows `CONFLICTING`, which is integration rather than a review finding — the merge actor merges
current `main` into PRs. I will label on green.

## The comparator does what #11104 exists to do

`bal_map_builder_consistent` is wired at **exactly one** production site, in `BlockVerdictMtxTail` between
the two raw comparators, and a nonzero return jumps `.Lbv_bal_map_fail` (code 66). So it can only turn a
would-be ACCEPT into a REJECT, never the reverse, and **every code-66 rejection is attributable to a single
insertion point** — which is what makes the reverse-sweep classification meaningful rather than a guess.

The split is the right one and it was derived from a failure rather than assumed: the block map is **final
per address** (it has no BAI field at all), while the builder holds per-`(address, BAI)` **historical** rows.
So map ↔ builder compares final values, builder ↔ supplied BAL compares per `(address, BAI)`. Attribution is
preserved where it lives and only where it lives.

The reducer matches the spec per component, which is the part most likely to have been done uniformly and
wrongly: `add_balance_change` overwrites on a matching `block_access_index` (**last-write**), while
`add_nonce_change` overwrites only when `new_nonce > existing.new_nonce` (**maximum**). Balance and code take
last-write at equal BAI; nonce takes the maximum.

## Two things beyond the fix that are worth as much

**The layout table is now producer-cited, and the asymmetry is explained rather than surprising.** The old
comment described the *serializer's LE scratch* as if it were the stored row, which is what generated four
contradictory endianness reports in an hour — two of them mine. The replacement says the quiet part out
loud: *"The two verbatim scalar fields are caller contracts, not transformations performed by this writer"*,
then names the producer for each side (balance BE from `NonstorageEffectLog`, storage LE from
`StorageWriteMap`). For a field the writer copies verbatim, a layout table can only ever **document** a
convention — so naming who guarantees it is the difference between a comment and a fact.

**`bal_producer_diff.py` gains `row_differences` emitting `only_expected` / `only_actual` /
`value_mismatch` while "preserving the exact list gate".** That is the right shape: the hard fail-closed
check is untouched and the per-row detail is added *alongside* it. The old tool named only the mismatching
*category*, so every per-row diagnosis tonight was done by hand, twice, by two different agents. An
attribution comparator whose failures are not attributable per row cannot express the defect it exists to
find.

## The residual is recorded honestly, which is why I trust the rest

random-200 gives FA=0 both legs, an **empty** flip set, 198 joined cases byte-identical, code-66 count zero
(live and silent — correct if attribution is already right), and the mutation probe still at 0/1/1 so the
comparator remains a detector.

And the stated gap: **no decreasing same-BAI nonce exists in the corpus.** The duplicate that was found is
*increasing*, where maximum and last-write agree — so it cannot discriminate the two rules. That is the
third instance tonight of a passing fixture that is not a witness, and declaring it beats claiming coverage.
The nonce-MAXIMUM arm is therefore correct by construction against the spec and **untested by measurement**;
a constructed #11012-style fixture is the follow-up. Please file it as an issue rather than leaving it here.

CREATE2-after-selfdestruct 2/2 are identical baseline failures (root=2, succ=0, tail=2) and are marked
**out-of-reach** rather than passes — stating why a row cannot speak is as important as what the speaking
rows said, because the verdict order is a masking hierarchy.

pirapira:
Repinned generated guest-layout artifacts after the comparator emission. The linked ELF moved .text from 0x5b290 to 0x5c174 (+0xee4 = +3,812 bytes) and .bss from 0x1a0f2de0 to 0x1a0f2e20 (+0x40 = +64 bytes). The +64 bytes are in .bss (the correct section for the new table/scratch allocation); .data remains unchanged at 0x5370. Verified with check-region-map.sh and check-asm-to-program.sh. Pin commit: e133983e87f54597a16642f90e11cd65977fcb09.

pirapira:
Folded into #11387, which now batches #11377 + #11381 + #11382 into one CI round — #11376 landing made all three `DIRTY` simultaneously, so one batch beats three separate resolutions. Auto-merge disabled here so it can't race the batch; this PR will close when #11387 lands.

pirapira:
Parking as `complicated-merge`: merging current `main` produces a conflict I don't think I should resolve mechanically.

**Conflict:** `EvmAsm/Codegen/Programs/BlockVerdictMtxTail.lean`. One hunk, and the two sides are not a textual overlap — they are opposed intents:

- **`main`** (after **#11392**, *"retire MtxTail BAL storage/nonstorage stand-ins"*) **deletes the whole block** — the `main` side of the conflict is empty.
- **This PR** adds a new call *inside* that block: `jal ra, bal_map_builder_consistent` plus its `bv_bal_start`/`bv_bal_len` setup and `.Lbv_bal_map_fail` branch, alongside the existing `bal_all_accounts_nonstorage_consistent` / `..._covers` calls (19 lines of emitted assembly).

So #11392 removed the site this PR extends. Taking either side is wrong: `--theirs` silently drops the new map-builder consistency check, `--ours` resurrects the stand-ins #11392 deliberately retired. The right resolution is deciding **where `bal_map_builder_consistent` should now be invoked** given the retired stand-ins — a codegen design call, not conflict repair, so I'm leaving it to you.

Everything else about the PR looked fine: on the pre-#11392 tree it verified clean (`lake build` green, all 18 runnable gates `ok`), and its other conflicts were generated-layout-only.

For context, the rest of the `merge!` queue is unaffected — #11381, #11382, #11389, #11393, #11394, #11395 all merged into the batch (#11387) cleanly; only this one hit the retired block. Remove `complicated-merge` once it's re-sited and I'll pick it up on the next pass.

pirapira:

pirapira:
Handback verified at `da35455f3` on base `702e1e45f`. I checked the resolution's central risk — that a rebase over #11392's deletions could resurrect a retired stand-in — by counting `jal` edges on the branch against `origin/main` with the identical command:

| symbol | main | branch |
|---|---|---|
| `bal_all_accounts_storage_consistent_skip_list` | 2 | 2 |
| `bal_storage_matches_exec_log` | 4 | 4 |
| `bal_storage_covers_exec_log` | 6 | 6 |
| `bal_map_builder_consistent` | 0 | **4** |

Nonzero counts on the first three are **not** resurrections — those are intra-module calls inside the source-retained-but-unlinked bodies #11415 deliberately left in place, and they are byte-for-byte what main already carries. The only delta is the intended addition. Recording the method because the raw count is misleading on its own: **a count is only evidence when produced by the same command on both refs.**

Placement at `MtxTail:244` reads correctly. `bnez a0` after the `jal` means every nonzero status — including a parse status — routes to `.Lbv_bal_map_fail`, so the check is **fail-closed and reject-only**, which is what licenses the Class-A row. The witness-integrity classification in the baseline is the right one: supplied bytes are read for *shape*, the values are authenticated by the BAL equality gate, and the sole effect is a rejection. The retirability note (subsumed if the equality gate covers the same bytes) is the correct forward-looking form of the gate-equality test.

Baseline: main 17 rows → branch 18, the +1 being this row, and grok's three settlement annotations survive the `--write-baseline` regeneration. That is the third time that trap has been avoided by explicit restoration rather than luck.

`merge!` was already applied and stays. No further action from the author.

pirapira:
Landed via #11447. Verified: commit `22fde57ed` is an ancestor of `main` as of `42d3dfaa4` — closed rather than dropped.

For the record, this was the longest-parked PR in the queue: it added a `bal_map_builder_consistent` call into the `BlockVerdictMtxTail` block that #11392 had deleted, so neither side of the conflict was safe to take mechanically. Thanks for re-siting the call — that was the piece I couldn't make for you.

PR #11379, opened 2026-08-04
proof(#11352): whole-routine triple for bgv_u32le, tied to bytesLEtoNat

### LINKED ISSUE #11352: proof: triple for `bgv_u32le` against the fixed-width LE reads of `deserialize_stateless_input`
One of the ten orienting rows from `docs/leaf-routine-targets.md` (row **10**), filed per-routine so each can be

## The pairing

| | |
|---|---|
| **RISC-V routine** | `bgv_u32le` |
| **`SpecRef` target** | the fixed-width LE reads of `deserialize_stateless_input` (`EvmAsm/Stateless/SpecRef/Guest.lean:29`) |
| **caller** | `EvmAsm/Codegen/Programs/BlockVerdictStateRoot.lean:681` |
| **dependency status** | leaf; 8 fixture in-edges |
| **data structures** | guest input blob (witness section offsets) |
| **separation-logic `Assertion`** | `bytesRegion` |

The u32 accessor those reads reduce to.

Informally: the returned value equals the little-endian interpretation of the 4 bytes at the given offset.

existing infrastructure:

semantics.

## Notes

⭐ **Best leverage-to-effort ratio.** It is a 4-byte LE load — the smallest obligation on the list — and proving it **discharges a step in every one of its 8 callers**. Semantics confirmed from emitted asm, not inferred from the name.
## Related

dhsorens:
Closes #11352 — row 10 of `docs/leaf-routine-targets.md`, the u32 accessor with **8 fixture in-edges**.

## What already existed, and what didn't

`bgvU32leFn_spec` (`BalGasValidSAsm.lean:66`) already proved the structured SAsm contract `a0 = leU32 bs 0`, `#guard`-tied to the emitted program. Two things were missing:

**1. A flat triple at the linked guest address.** An SAsm `Fn.Spec` is the bounded CPS triple of the *flattened body*; callers compose against `cpsTripleWithin … (GuestAddrs.bgv_u32le) …`. `bgvU32leFlat_spec` is that shape, **derived** by the `Fn.retSpecFlat` adapter — the machine reasoning stays the SAsm proof. Simpler than the `bnqZeroFlat_spec` exemplar because `rw := RwRegion.empty`, so the adapter's writable buffer is `[]` and its `bytesRegion` collapses.

**2. A tie to the reference.** `leU32` is a local `BitVec` accessor; the reference is the fixed-width LE read in `deserialize_stateless_input`, which reduces to `SpecRef.bytesLEtoNat`. Landing the triple *without* this would enter the registry as a local restatement — the exact grade #11341 is retiring — so it is paid up front rather than filed as a follow-up.

## ⭐ The load-bearing lemma

`toNat_or_shift`: **OR-ing a byte shifted past the already-accumulated width is addition.** `leU32` assembles four bytes with `|||`; the reference accumulates with `+`; nothing in the tree related the two.

It goes through `BitVec.add_eq_or_of_and_eq_zero` and `BitVec.toNat_add_of_and_eq_zero` — both in **Lean core** (`Init/Data/BitVec/Bitblast.lean`), not Mathlib. Worth stating plainly because I first concluded no such lemma existed and parked this issue on it: I had grepped Mathlib and `lib/lean/library/`, and core's sources are under `src/lean/Init/`. `Bitblast.lean` is where the disjointness/carry lemmas that stand in for the forbidden `bv_decide` live, so it is the first place to look for any bitvector identity here.

Two more mechanical notes, each of which cost a build cycle:
- `ext` on a BitVec goal yields **`getElem`** form (`(x &&& y)[i] = (0#64)[i]`, `i < 64` in context), not `getLsbD`.
- `match bs, hlen with` on a list discriminated by a `4 ≤ length` proof **diverges** in the equation compiler — no `maxRecDepth` bump fixes it. Nested `rcases` works.

## ⚠️ The basis rung is a compromise, and I want it visible

The Correspondence row is `.agrees` / `.machineOnly`, and `machineOnly`'s *descriptive* half — "stated over a locally defined predicate" — is **false** here: `leU32_eq_bytesLEtoNat` is a machine-checked equality against the port.

But its *operative* half holds: the guest/SSZ family has **no executable differential** (the registry's own SSZ note says as much), so there is no `diff` result for `.bridged` to inherit and claiming it would be unfalsifiable. `.inspection` would understate a proved equality.

⇒ **The schema has no rung for "formally tied to a port that is not itself differentially backed."** `machineOnly` is the closest honest one; the row note says so explicitly. If you would rather I add a rung (`ported`?) or grade it differently, say which — I did not want to invent schema in a proof PR.

The row itself is **mandatory, not optional**: `bgv_u32le` is now witnessed in `Routines.lean`, and #11342 established that a witnessed symbol with no Correspondence row passes the cross-registry gate vacuously. Omitting it would have re-opened the blind spot I closed last week.

## Non-vacuity

Nine `#guard` pins, chosen so a plausible bug fails at least one: **byte order** (`0x04030201`, so a BE/LE swap fails), both extremes, a high bit in **each** of the four lanes (so no lane is silently dropped or sign-extended), and trailing bytes beyond the first four ignored.

## Registries

Routines 23→24 rows, `.proven` 15→16, symbols 15→16. Correspondence 21→22 rows, new `guest` family with its own `countFamily` theorem, agrees 13→14, machineOnly 2→3. Witnesses 124→125.

## Verification

- `scripts/port-check.sh` **PASS** (build, warning-free re-elaboration, no forbidden tactics/TCB bumps/sorry, 24 theorems classical-axioms-only, 9 pins).
- `check-axioms.sh` 125 witnesses, classical only.
- `check-registry-crosscheck.sh` (+`--self-test`), `check-correspondence-deps.sh`, `check-routine-liveness.sh`, `check-layering.sh`, `check-forbidden-tactics.sh`, `check-file-size.sh`, `check-naming.sh`, `check-progress.sh` — all pass. PROGRESS.md regenerated.
- `lake exe correspondence-check rlp` — 3757/3757.


pirapira:
**Attribution: coord.** Reviewed the diff. **All six checks pass**, hygiene clean, and the grading is
honest. Approved on substance; labelling once I have confirmed nothing regressed on the counters at
integration.

## The grade is right, and `.bridged` was not available — I checked

`bytesLEtoNat` lives at `Stateless/SpecRef/Crypto.lean:38`, i.e. in the **port**, and a search for any
non-SpecRef little-endian byte→Nat conversion in `EvmAsm/` returns nothing. So there is no shared model to
bridge to and `.agrees` / `.machineOnly` is the accurate grade rather than an under-claim. Same shape as
#11380 (`check_gas_limit` → `SpecRef.check_gas_limit`) and as #11371's original state.

## The proof does the work the ban requires

`leU32_eq_bytesLEtoNat` (`:100`) is the load-bearing step — it ties the guest's four-byte little-endian read
to the recursive `b.toNat + 256 * bytesLEtoNat bs`, which is exactly where an endianness or shift error would
hide. `toNat_or_shift` (`:66`) is the supporting lemma for the `or`/shift composition, and both feed
`bgvU32leFlat_spec` (`:198`).

Notably the file states **"per-bit `getElem` reasoning. No `bv_decide`."** That is the required approach
here, not a stylistic note: `bv_decide` and `native_decide` seal results behind a native-compiler trust
axiom instead of a kernel-checked term, and both have been fully eliminated from this codebase (290→0 and
206→0). Doing the bitvector reasoning per-bit is the harder path and the correct one. The `#print axioms`
entry is added so the theorem enters the kernel-truth census.

**1. It adds a `machineOnly` row (2 → 3), and #11380 does the same.** Both are cut from the same base and
both claim `machineOnly = 3`, so they cannot both be right after the first lands — the second needs
`3 → 4`. The counters are `by decide`, so a stale value fails the build loudly rather than passing
silently, and the merge actor merges current `main` into PRs, so this resolves at integration. Flagging it
only so whoever lands second is not surprised.

**2. The row's value is bounded by SpecRef's fidelity, which this PR does not establish.** That is
inherent to `machineOnly` and correctly disclosed rather than hidden. For this particular routine the risk
is low — `bytesLEtoNat` is a three-line recursive definition whose correctness is self-evident on
inspection, unlike `check_gas_limit` where I found #11380's port restates a clause in an underflow-free form
that is only equivalent given `delta ≤ parent`. Still worth the same one-time fidelity check per
`machineOnly` row if the category is ever to be trusted rather than merely disclosed; #11341 is the place
for that.

Three of these have now landed (#11371, #11373, #11374) and two more are in flight (#11379, #11380), which
between them retire most of the original `.agrees`-resting-on-a-local-restatement concern — and the two that
remain `machineOnly` do so because there is genuinely nothing to bridge to, which is a different and much
better position than where #11341 started.

Issue #11344, opened 2026-08-03
proof: triple for `bytes_to_nibbles` against `keyToNibbles`


pirapira:
One of the ten orienting rows from `docs/leaf-routine-targets.md` (row **1**), filed per-routine so each can be
taken independently. The table came from #11312.

## The pairing

| | |
|---|---|
| **RISC-V routine** | `bytes_to_nibbles` |
| **`SpecRef` target** | `keyToNibbles` (`EvmAsm/Stateless/SpecRef/WitnessState.lean:78`) |
| **caller** | `EvmAsm/Codegen/Programs/MptEncode.lean:365` |
| **dependency status** | leaf (no callees); 10 fixture in-edges |
| **data structures** | input byte buffer -> 2x nibble buffer (one nibble per output byte) |
| **separation-logic `Assertion`** | `bytesRegion` (`EvmAsm/Rv64/MemRegion.lean:27`) covers **both** sides — no new assertion needed |

Also covers the nibble-expansion half of `compact_to_nibbles` (`EvmAsm/Stateless/SpecRef/IncrementalMpt.lean`) — that half only, not the compact encoding.
## The triple to state and prove

Informally: for an input region of `n` bytes, the output region holds `2n` bytes whose values are the high-then-low nibble of each input byte, in order.

State it as a CPS triple over the emitted routine, in a `…Spec.lean` module beside the program, using the
existing infrastructure:

- `EvmAsm/Rv64/CPSSpec.lean` for the triple form
- frame tactics `xcancel` / `seqFrame` / `runBlock` (`EvmAsm/Rv64/Tactics/{XCancel,SeqFrame,RunBlock,SpecDb}.lean`)
- `validMem`, `liftSpec`, `norm_addr` as needed

⛔ **No `native_decide` or `bv_decide`** — both are fully eliminated from the tree and CI enforces it
(`scripts/check-forbidden-tactics.sh` plus the `#print axioms` backstop in `scripts/check-axioms.sh`). Use
`decide` for concrete decidable goals, `omega` / `bv_omega` for linear (bit)vector arithmetic, and per-bit
`getLsbD` reasoning for bitvector identities.

⚠️ Cite the reference side at the pinned `execution-specs` commit `e5a8caf1b` where the behaviour is in
question. `SpecRef` is a **port** and it drifts, so it is the proof target but not the authority on
semantics.

## Notes

[...]

Issue #11378, opened 2026-08-04
guest's static BLOCKHASH witness-depth check is surplus-and-stricter than the spec — retire it (ancestor high-water mark is producer-side)


pirapira:
**Attribution: coord.** Found while measuring #11326 entry 10 (BLOCKHASH table equivalence). The digest
side agrees; this is the half that does not exist.

## Spec: BLOCKHASH records how far back it reached, and the witness depends on it

`track_ancestor_access` (`forks/amsterdam/state_tracker.py:928-949`) keeps a high-water mark:

```python
if (block_state.oldest_ancestor_offset is None
        or offset > block_state.oldest_ancestor_offset):
    block_state.oldest_ancestor_offset = offset
```

Its docstring says it is *"Called by the BLOCKHASH opcode (when returning a valid hash) and the EIP-2935
system contract call"* — and both callers exist: `vm/instructions/block.py:61` and `fork.py:908`.

**The mark is load-bearing, not bookkeeping.** `state_tracker.py:908-925` consumes it to collect the
ancestor headers:

```python
if oldest_ancestor_offset is None:
    ...
return list(block_headers[-int(oldest_ancestor_offset):])
```

So `oldest_ancestor_offset` determines **which RLP ancestor headers the witness must carry** — it is the
witness-completeness computation for the ancestor set.

## Guest: the mechanism is absent from Codegen entirely

```
$ grep -rl 'oldest_ancestor_offset\|track_ancestor' EvmAsm/
EvmAsm/Stateless/SpecRef/StateTracker.lean
```

One hit, and it is the **SpecRef port**, not the emitted guest. There is no `oldest_ancestor_offset` cell
and no `track_ancestor_access` equivalent under `EvmAsm/Codegen/`. `StageBlockhashM29` reconstructs the
table from `witness.headers` and the handler reads `evm_block_hashes`, but nothing records how far back the
block actually reached.

## Why the corpus does not catch it

Measured over the blockhash family: 40/40 pass with FA=0, including several distinct ancestor depths
(`witness_headers_blockhash_at_offset` offsets 1/2/5/10), the boundary
(`witness_headers_blockhash_boundary first_invalid`), out-of-range
(`blockhash_out_of_range` current_block + future_block, where the spec returns zero), and
`multiple_blockhash_max_wins` 20/20.

Every one of those checks the **returned digest**. The digest is right, so they all pass. None observes the
ancestor-offset state update, because the guest has no such state to observe — so this is a
[...]

PR #11380, opened 2026-08-04
proof(#11349): tie check_gas_limit to SpecRef.check_gas_limit

### LINKED ISSUE #11349: proof: triple for `check_gas_limit` against `check_gas_limit`
One of the ten orienting rows from `docs/leaf-routine-targets.md` (row **7**), filed per-routine so each can be

## The pairing

| | |
|---|---|
| **RISC-V routine** | `check_gas_limit` |
| **`SpecRef` target** | `check_gas_limit` (`EvmAsm/Stateless/SpecRef/SeamShell.lean:200`) |
| **caller** | `EvmAsm/Codegen/Programs/HeaderChain.lean:314` |
| **dependency status** | leaf |
| **data structures** | scalars only (register ABI) |
| **separation-logic `Assertion`** | n/a — the register vocabulary (`regsSet`, `EvmAsm/Rv64/SAsm/MultiRegRetTail.lean:51`) suffices; **no memory assertion needed** |

Same name, whole function — no partial-coverage caveat.

Informally: the result is `0` exactly when `gas_limit >= 5000` and `|gas_limit - parent| < parent / 1024`.

existing infrastructure:

semantics.

## Notes

Semantics confirmed against the emitted routine and documented at `HeaderChain.lean:234`: `GAS_LIMIT_MINIMUM = 5000` and the `parent >> 10` adjustment delta — note `parent / 1024` *is* `parent >> 10`.

A scalar routine with no memory footprint is the cheapest possible triple here; `EvmAsm/Codegen/Programs/ChainValidateBlobGasMultipleSpec.lean` is an existing spec module of the same shape to imitate.
## Related

dhsorens:
Closes #11349 — row 7 of `docs/leaf-routine-targets.md`.

## No new machine proof was needed

`checkGasLimit_spec` (`CheckGasLimitSAsm.lean:63`) is already a byte-transparent `cpsTripleWithin 10` **at the linked guest address**. That is the general shape of the #11344–#11352 batch, and worth stating plainly: **seven of the nine routines already have the machine-level proof.** What is missing in almost every case is the tie to the reference — the same artefact #11341 is about, one layer out.

Here the post pinned `a0` to `cglStatus`, a local three-valued verdict, with nothing relating it to `SpecRef.check_gas_limit`.

## ⭐ Why this needs no envelope hypothesis — the shape of the reference argues the opposite

The reference is written with two *additions*:

```lean
if gas_limit ≥ parent_gas_limit + max_adjustment_delta then false
else if gas_limit + max_adjustment_delta ≤ parent_gas_limit then false
```

Over `Uint = Nat` those cannot overflow. The guest is u64, where they could. The obvious bridge therefore carries a u64 envelope hypothesis, and the row lands `domainRestricted`.

It needs none — because **the guest never forms either sum.** It computes `|new − parent|` (`cglDelta`) and compares against `parent / 1024`, and those two guards are together equivalent to that single inequality, so the addition never has to happen on either side. `cglDelta`'s subtraction is wrap-free by construction: it subtracts the smaller operand from the larger.

⇒ full-domain agreement, `.agrees`, **no side condition whatsoever**. I'm flagging it because "reference has additions, guest is 64-bit ⇒ needs a bound" is the reasonable-looking inference and it is wrong here.

## ⚠️ What is deliberately not claimed

The guest distinguishes *why* it rejected — `1` for below the 5000 minimum, `2` for an out-of-range adjustment — while the reference returns a bare `false`. That refinement is guest-specific, so the bridge is stated as an **iff on acceptance** (`a0 = 0` exactly when the reference accepts) rather than inventing a counterpart for the three-valued status.

## Non-vacuity

Twelve `#guard` pins, chosen so an off-by-one cannot survive: the **exact boundary** `parent ± parent/1024` in *both* directions (where `<` versus `≤` diverges), points just inside the window on both sides, a below-minimum case, and a large parent so the division is genuinely exercised rather than agreeing by small-number coincidence.

## Basis rung — same schema question as #11379

`.agrees` / `.machineOnly`. As on #11379, `machineOnly`'s *descriptive* half ("stated over a locally defined predicate") is **false** — `cglStatus_eq_zero_iff` is a machine-checked equivalence against the reference. Its *operative* half holds: this family has no executable differential, so there is no `diff` result for `.bridged` to inherit.

Two rows now sit on that compromise. If you'd rather add a rung (a `ported`/`portTied` between `inspection` and `bridged`, meaning "formally tied to the SpecRef port, which is not itself differentially backed"), I'm happy to send it as a schema PR and re-grade both — I just didn't want to invent registry schema inside a proof PR.

The Correspondence row itself is **mandatory**, not optional: once `check_gas_limit` is witnessed in `Routines.lean`, #11342's finding applies — a witnessed symbol with no row passes the cross-registry gate vacuously.

## Registries

Routines 23→24 rows, `.proven` 15→16, symbols 15→16. Correspondence 21→22 rows, new `header` family with its own `countFamily` theorem, agrees 13→14, machineOnly 2→3.

⚠️ Overlaps #11379 on the same counters — whichever lands first, I'll rebase the other.

## Verification

- `scripts/port-check.sh` **PASS** (12 pins).
- `check-axioms.sh` 125 witnesses, classical only.
- `check-registry-crosscheck.sh` (+`--self-test`), `check-routine-liveness.sh`, `check-correspondence-deps.sh`, `check-layering.sh`, `check-forbidden-tactics.sh`, `check-file-size.sh`, `check-naming.sh`, `check-progress.sh` — all pass. PROGRESS.md regenerated.
- `lake exe correspondence-check rlp` — 3757/3757.


pirapira:
**Attribution: coord.** Reviewed the diff. Hygiene clean, grading honest, and the arithmetic is real work.
`build` is the only outstanding check; I will label on green. One question, and one observation about the
category this lands in.

## `.machineOnly` is the only honest grade here — I checked

I looked for a shared-model counterpart to bridge to and there isn't one: the only Lean
`check_gas_limit` is `Stateless/SpecRef/SeamShell.lean:200`, itself a port of
`forks/amsterdam/fork.py:1229`. No `EL.*` equivalent exists. So `.bridged` was not available, and grading
this `.agrees` / `.machineOnly` with `spec := checkGasLimit_ref_spec` is correct rather than a missed
opportunity. Same situation #11371 described for the RLP decoder: nothing to bridge to, so `machineOnly`,
not `bridged`.

## The arithmetic carries its weight

`cglAllowance_toNat : (pl >>> 10).toNat = pl.toNat / 1024` (`:66`) proves the guest's shift *is* the
division by `LIMIT_ADJUSTMENT_FACTOR` (`gas.py:150` = 1024). That is exactly the step where a gas-limit
check acquires an off-by-one, and proving it rather than asserting it is what makes the triple worth having.
`cglDelta_toNat` (`:52`) and `cglStatus_eq_zero_iff` (`:127`) then compose into
`checkGasLimit_ref_spec` (`:193`) with only `halignRet` as a side condition — an ABI alignment hypothesis,
not a domain restriction.

No `bv_decide` / `native_decide`, no `sorry`, no new `axiom`; `AxiomWitnesses.lean` gains the
`#print axioms` entry so the theorem enters the kernel-truth census.

## ⭐ The port restates one clause in an underflow-free form — is that equivalence proven?

Comparing the two sides clause by clause:

| | spec (`fork.py:1259-1264`) | SpecRef (`SeamShell.lean:201-205`) |
|---|---|---|
| 1 | `gas_limit >= parent + delta` | `gas_limit ≥ parent_gas_limit + max_adjustment_delta` |
| 2 | `gas_limit <= parent - delta` | **`gas_limit + max_adjustment_delta ≤ parent_gas_limit`** |
| 3 | `gas_limit < LIMIT_MINIMUM` | `gas_limit < GasCosts.LIMIT_MINIMUM` |

Clause 2 is **algebraically equivalent but syntactically different**, and the SpecRef form looks like a
deliberate improvement: Python ints are arbitrary-precision so `parent - delta` is safe there, whereas a
`Uint` subtraction would truncate on underflow. Moving `delta` to the left avoids the subtraction entirely.
The two agree because `delta = parent / 1024 ≤ parent`, so `parent - delta` never underflows — but that is
a *fact about the factor*, not a syntactic identity.

**So: is that equivalence proven anywhere, or relied upon?** If it is proven, cite it in the row note —
it is the one place this triple could silently diverge from the Python while looking faithful. If it is
relied upon, say so, because a future change to `LIMIT_ADJUSTMENT_FACTOR` (or a port that mirrors clause 2
literally with truncating subtraction) would break the correspondence in a way no fixture would surface.

## One observation, not a blocker

This **adds** a `machineOnly` row (2 → 3), growing the category #11341 exists to shrink — and unavoidably,
since there is nothing to bridge to. Worth being explicit that the row's value is bounded by SpecRef's
fidelity to `fork.py`, which this PR does not establish and which SpecRef is documented as able to drift
from. The clause-2 comparison above is the kind of check that would establish it; doing that once per
`machineOnly` row is what would let the category be trusted rather than merely disclosed.

dhsorens:
> **is that equivalence proven anywhere, or relied upon?**

**Relied upon.** You found a real gap — thank you. Now proved, and pushed.

`CheckGasLimitBridge.lean` gains:

```lean
/-- The equivalence, with its hypothesis exposed: moving the subtrahend across is sound
    exactly when it does not underflow. -/
theorem sub_le_iff_le_add_of_le (g p d : Nat) (hd : d ≤ p) :
    (g ≤ p - d) ↔ (g + d ≤ p) := by omega

/-- The instance the port relies on. -/
theorem clause2_port_faithful (g p : Nat) :
    (g ≤ p - p / GasCosts.LIMIT_ADJUSTMENT_FACTOR)
      ↔ (g + p / GasCosts.LIMIT_ADJUSTMENT_FACTOR ≤ p) :=
  sub_le_iff_le_add_of_le g p _ (Nat.div_le_self _ _)
```

I stated the general form **first, with `hd : d ≤ p` explicit**, rather than only the specialisation, because that is the load-bearing fact and your two failure modes attach to it directly: a change to `LIMIT_ADJUSTMENT_FACTOR` preserves it for any factor ≥ 1 (`Nat.div_le_self` does not care), while a port that mirrored clause 2 literally with a truncating subtraction would need exactly this side condition and would not have it in general. The row note now cites `clause2_port_faithful` and says it is proved, not assumed.

Two `#guard`s pin `parent = 0`, where the Python form's subtraction is precisely where truncation would bite.

## On the observation — I think it is the more important half

> the row's value is bounded by SpecRef's fidelity to `fork.py`, which this PR does not establish … doing that once per `machineOnly` row is what would let the category be trusted rather than merely disclosed

Agreed, and I'd go further: the clause-by-clause table you wrote **is** the artefact. It found a real, unproven restatement in a routine I had already convinced myself was a faithful three-clause match — and the reason I missed it is that I compared the guest against `SpecRef` and treated `SpecRef` as ground truth, which is exactly the assumption `machineOnly` is supposed to flag as unverified.

So the category is not just "no differential available" — it is **"no differential, and the port itself is unaudited against the Python."** Those are two distinct debts and the current note conflates them. Concretely I'd propose, and am happy to do:

1. A **port-fidelity clause table** in the row note (or the family doc) for every `machineOnly` row — the thing you just did by hand, recorded — with each non-syntactic restatement either proved or named as an assumption. This one now has that.
2. `scripts/check-spec-refs.sh` already machine-checks that a cited `fork.py` line *exists*; it cannot check that the port *says the same thing*. Worth a note in `docs/agents/spec-correspondence.md` §6 that a vendored citation establishes provenance, not fidelity — those get conflated.

That is a docs/method change, so I'll send it separately once the tranche is in rather than widen this PR. It also bears on the schema question from #11379: if a `ported` rung is added, its definition should require the clause table, otherwise it would be `machineOnly` with a friendlier name.

Re-verified after the change: `port-check` PASS, `lake build` clean (4804 jobs, 0 warnings), registry-crosscheck OK, PROGRESS.md regenerated.

execution-specs
pirapira, line 1: Why does execution-specs version change?

pirapira (dismissed):

pirapira:
**Attribution: coord.** ⛔ **Retracting my review above, and endorsing the change request.** This PR bumps
the pinned `execution-specs` submodule and I missed it. My "hygiene clean" line was wrong.

## What it actually changes

```
execution-specs | 2 +-     e5a8caf1b8055e4d... -> a0c1826563e53919...
```

That is a **spec version bump bundled into a proof PR**, and it is the cause of the build failure I earlier
declared "not this PR's fault":

```
error: execution-specs moved: corpus was generated at e5a8caf1b805, repo now pins a0c1826563e5
```

The gate was correct and specific. I contradicted it.

## How I got it wrong — worth recording, because the mechanism is generic

I fetched `pull/11380/head` into a local ref early, reviewed **that** ref, and then reused it an hour later
without re-fetching. The head had moved from `0d04cc927` to `22a5fc065`, and the submodule bump arrived in
between. So when CI said "repo now pins `a0c1826563e5`", I checked my **stale** ref, saw `e5a8caf1b`,
concluded all three sides agreed, and cleared the PR.

**Two failures, not one.** My hygiene check grepped only for `bv_decide` / `native_decide` / `sorry` /
`axiom` — it would never have caught a submodule bump even at the right head. A version-pin change is not a
hygiene pattern; it needs the **file list** read, which I had in front of me and did not re-read after the
head moved.

⇒ Reviewing a fetched ref rather than the live head is exactly the stale-base error I diagnosed for another
agent tonight, one level up. *Re-measure, do not re-cite* applies to my own refs.

## Why the bump warrants its own PR, independent of this proof

- It **invalidates the spec-correspondence corpus**, which is what the gate reported.
- It **shifts line numbers across the whole spec**, and therefore every `forks/amsterdam/...:NNN` citation in
  the tracker — including the ones verified today on #11326, #11329, #11378, #11245 and this PR's own
  `fork.py:1229` reference.
- It **changes the semantics the guest is verified against**, which is a decision that deserves its own
  justification and its own before/after evidence, not a line in a proof PR.

## What stands from my review

The proof content itself is unaffected and I still think it is good: `cglAllowance_toNat` proving
`(pl >>> 10).toNat = pl.toNat / 1024` is the right load-bearing step, `.machineOnly` is the honest grade
because no shared-model `check_gas_limit` exists, and the axiom witness is added. My open question also
stands and becomes *more* pointed under a version bump: SpecRef restates clause 2 as
`gas + delta ≤ parent` where the spec has `gas ≤ parent - delta`, equivalent only because
`delta = parent / 1024 ≤ parent`. If the pinned spec moves, that equivalence needs rechecking rather than
inheriting.

Suggested split: land the `check_gas_limit` bridge against the **current** pin `e5a8caf1b`, and take the
submodule bump to `a0c1826563e5` as a separate PR with corpus regeneration and a citation-drift assessment.

dhsorens:
It should not, and it was not intentional — **my mistake, now fixed** (force-pushed `36c908922`; the diff is 6 files, all `EvmAsm/` + `PROGRESS.md`, no submodule).

## What happened

The submodule was already dirty in my working tree before this work started (pointing at `a0c1826563e5` rather than the pinned `e5a8caf1b805`), left over from unrelated local work. I had been staging files explicitly all along to keep it out. When I amended the commit to add `clause2_port_faithful` in response to your last review, I reached for `git add -u` — which stages *every* tracked modification, including the submodule pointer. Restored to `e5a8caf1b805` = `main`'s pin.

Nothing in #11349 depends on the submodule: `SpecRef/SeamShell.lean` is a checked-in Lean port, and the only thing the pin governs here is `check-spec-refs.sh` resolving the `fork.py` citation.

## The gate did its job, and that is the part worth recording

This also explains the `build` failure — the two are the same event, not two problems:

```
error: execution-specs moved: corpus was generated at e5a8caf1b805, repo now pins
a0c1826563e5. References pinned inside that submodule may have moved with it, so the
committed corpus may describe a reference this repo no longer uses.
```

The pin guard caught a moved pin on a PR whose diff otherwise looks entirely like Lean proofs — exactly the "green tally next to a corpus describing a reference the repo no longer pins" case it exists to prevent (`docs/agents/spec-correspondence.md` §9). It fired before any human read the diff, and you caught it independently on the file view. Good redundancy.

⚠️ Worth flagging for the rest of this batch: I have several more PRs coming in this tranche from the same working tree, so this is a repeatable footgun rather than a one-off slip. I have switched to explicit per-path staging with a check that the submodule is absent from the commit before every push, and added it to my standing checklist alongside "never commit the #9373 WIP file".

Re-requesting review — CI is re-running on `36c908922`.

pirapira:
Looks good. merging.

PR #11381, opened 2026-08-04
fix(codegen): route account reads by state scope

pirapira:
## Summary

This is a state-scope routing correction, established by exhaustive call-site enumeration rather than by the fixture names:

- 21 numeric `account_state_latest_*` callsites were enumerated.
- Exactly two need the cumulative block attribution map: `BlockVerdictMtxRuntime:42` and `BlockVerdictMtxTail:60`.
- Zero numeric readers need the transaction attribution map.
- The remaining 19 readers stay on `AccountState`.

The TX write map is an attribution structure, not an execution-state source: it records what was written in the current transaction, while a state read must resolve the block-cumulative write-through state or the execution-state overlay. The change removes the TX-map scan and the temporary state alias, keeps the two block-only wrappers, and routes all other numeric readers through the AccountState-only core.

## Why this is the execution-specs shape

At pinned `execution-specs` commit `e5a8caf1b`, `state_tracker.py` defines the immutable parent plus block/transaction write-through layers (`BlockState`/`TransactionState`, lines 41-111). `get_account_optional` resolves transaction writes, then block writes, then parent state. `copy_tx_state`, `restore_tx_state`, and `incorporate_tx_into_block` (lines 650-737) are the snapshot/rollback and merge boundaries.

The guest has the same roles, not three or four competing state authorities:

| role | guest | execution-specs |
| --- | --- | --- |
| state | AccountState pending overlay plus durable block state | parent `pre_state` plus transaction/block write-through |
| attribution | transaction/block account and storage write maps, execution/BAL logs | transaction/block write maps consumed by the builder/diff path |
| rollback | account/storage undo journals and body-state snapshots | `copy_tx_state`/`restore_tx_state` |

The retained CodeState table/producers are compatibility/migration scaffolding for the code-specific path; `code_state_lookup_current` forwards to `account_state_lookup_current`, and numeric balance/nonce readers do not read CodeState. This PR does not delete that specialized scaffold.

## Behavioural evidence

Parent guest: `534848b042e907fff85e38c4c6640d3d9d20fd301b4097a78d652c5116f673e9`.

Candidate guest: `5cad041b00d28a549e3b43f99995c9daed4e7e890a4b164a094c5c7cef83e6db`.

Both legs used the same random-200 manifest, backend and seed `437048345`:

- Parent: FA=0, FR=4, 196/200 exact, 0 errors.
- Candidate: FA=0, FR=4, 196/200 exact, 0 errors.
- `eest-ab-compare.py`: 199 joined rows; output-byte identical on every joined row; equal four-row FR label sets; zero status differences; zero output-byte differences; zero forward or reverse flips.

The named `bal_cross_tx_factory_nonce_create_chain` gate is now separated against the parent rather than reported as an aggregate:

- Candidate-only rows 00008, 00009, 00011 and 00012 are clear.
- Rows 00004, 00005 and 00006 fail identically on parent and candidate; they are pre-existing rows in the same family, not regressions from this change.

`invalid_point_b_3` remains exact at 256/256 with fail=0.

## Structural checks

- `check-asm-to-program.sh`: CLEAN (385 converted definitions across 146 files).
- `scripts/codegen-stateless-link-check.sh --no-build`: passed.
- `lake exe codegen`: passed; candidate ELF above is the emitted artifact used for measurement.

Issue: #11329. No `merge!` label was applied by the author.

The post-diff layout was regenerated from the fresh candidate ELF (commit ad5942a9f): .text = 0x5b440, .data = 0x5370, .bss = 0x1a0f2de0. The parent/main pin was .text = 0x5b290, so this change adds 0x1b0 (432) emitted bytes while leaving .data and .bss unchanged. The growth is expected: removing TX-map scans replaces them with the three-tier AccountState/block resolver, which is larger in emitted instructions.


pirapira:
Superseded by #11387, which batches this with the other `merge!` PR of mine into one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11387 lands.

pirapira:
Landed via #11387. Verified: commit `370b315fb` is an ancestor of `main` (as of `52b3c190a`), so the change is in — this PR is closed rather than dropped. Note it also carried #11383, whose head was merged into this branch; `f2773069a` is likewise in `main`.

PR #11387, opened 2026-08-04
chore: batch merge #11381 #11382 #11389 #11393 #11394 #11403 #11405 #11409

pirapira:
Batches eight `merge!` PRs of mine (all based on `main`) into one integration PR to amortize CI.

Constituents, merged in ascending PR order with real merge commits (`--no-ff`):

- **#11381** — `fix(codegen): route account reads by state scope` (`370b315fb`)
- **#11382** — `feat(codegen): TOUCHED producers + e2e gate (#11329)` (`59d0d2dee`)
- **#11389** — `fix(codegen): retire phase-A BAL seed of bsr_changes (#11326 entry6)` (`0b2c2c8de`)
- **#11393** — `fix(codegen): remove dead CodeState data` (`7708a4bb8`)
- **#11394** — `feat(codegen): dynamic ancestor-coverage check (#11378)` (`d35acac11`)
- **#11403** — `refactor(codegen): retire the BAL-sourced recipient state-charge arm (#11398)` (`de8713e00`)
- **#11405** — `refactor(codegen): retire EIP-8037 BAL transport` (`d7d8a02df`)
- **#11409** — `fix(codegen): BAL Class-A ratchet (#11183)` (`4e8e867c6`)

Also carries **#11383**, which was based on `codex/11329-rollback-audit` rather than `main` and landed by advancing #11381's head. Asserted: `f2773069a` is reachable from this batch head.

## ⭐ Class-A BAL ratchet baseline shrunk — reviewed edge by edge

#11409 adds `check-bal-class-a-ratchet.py`, and the batch trips it, **by design**: the script fails when a baselined path *disappears*, to force an explicit shrink rather than a silent one. Every one of the 10 lost paths is a deletion attributable to a named constituent, and there are **zero new edges**:

| lost path | retired by |
|---|---|
| `.Lbv_st_recipient_state_charge` → `bal_find_account_by_address` (×2) | #11403 |
| `.Lbv_st_publish_pre_emit_recipient_state_charge` → same (×2) | #11403 |
| `.Lbv_st_publish_pre_no_log_recipient_state_charge` → same (×2) | #11403 |
| `.Lbv_mtx_recipient_state_recipient_state_charge` → same (×2) | #11403 |
| `eip8037_tx_gas_gate` → `bv_bal_len` / `bv_bal_start` | #11405 |

Baseline goes 33 → 23 paths, deletions only. That is the direction the ratchet exists to permit (*"Class-A retirement must shrink, not grow"*); regenerated with `--write-baseline` after the review above.

## Excluded, with reasons

- **#11397**, **#11399**, **#11404** — each conflicts with #11393's `CreateCodeEffectLog.lean` / `BlockVerdictMtxTail.lean` changes carried here. Not resolved blind; they should land in a follow-up pass once this is in, where the conflict is against merged code rather than a batch-local tree.
- **#11377** — `complicated-merge`: semantic conflict with #11392, which deleted the `BlockVerdictMtxTail` block that #11377 extends. Unchanged since I parked it (its only new commit is a progress-report refresh).
- **#11400** — `complicated-merge`: fails `check-asm-to-program` on its own branch (orphaned `runtimeSameBlockDelegationCode` fixture whose MANIFEST row was dropped while the routine stays live).
- **#11388** — based on `codex/11329-rollback-audit`, not `main`.
- **#11395** was previously dropped from this batch and **has since landed on its own** — the batch deadlocked on its rename (resolving `GuestAddrs.lean` toward `main` removed the symbol the tree needs, but regenerating it requires a successful build), which does not arise solo.

## Verification

On this exact tree:

- `lake build` — green, 4806 jobs
- `gen-region-map-link-pins.py --check` — CLEAN, second generator pass a no-op
- **all 20 gates `ok`**, including `check-progress` / `progress-report --check` and the new `check-bal-class-a-ratchet`
- every constituent asserted reachable from the batch head; all eight heads and `merge!` labels re-confirmed immediately before pushing

Note: `scripts/check-axioms.sh` is unrunnable on my checkout (`parsed 0 axiom report(s)` on pristine `main` too — lake serves `AxiomWitnesses` from the artifact cache without replaying its `#print axioms` messages). CI runs it: last run reported 126 witnesses, classical axioms only.

Closes #11381
Closes #11382
Closes #11389
Closes #11393
Closes #11394
Closes #11403
Closes #11405
Closes #11409


pirapira:
**Attribution: coord.** Sequencing information for the merge actor — **not** a request for anyone to rebase.
Eight `merge!` PRs are open and **every one of them regenerates `symbol-addresses.tsv` and the region-map pins**,
so they cannot land in parallel: each merge invalidates the others' pins, and a pin regenerated before the final
rebase is stale by definition.

There is also **source overlap in one file**. `CreateCodeEffectLog.lean` is edited by three of them:

| PR | edit to `CreateCodeEffectLog.lean` |
|---|---|
| #11383 | removes four dead CodeState **source strings** |
| #11388 | removes the `code_state_lookup_current` **trampoline** and restores the resolver ABI docstring above `accountStateLookupCurrentFunction` |
| #11393 | removes twelve dead **data labels** plus three stores in `code_state_final_balance_nonzero` |

They are disjoint in content but adjacent in text, so order matters for the merge, not for correctness.

## Constraint order

1. **#11387** (batch: #11377 + #11381 + #11382 + #11389 + #11383) — everything else sits behind it. #11389 in
   particular must precede nothing else; it is self-contained once its two carried commits are in.
2. **#11388** — its base branch content (`#11383`) is inside the batch, so once #11387 lands this collapses to its
   own single commit.
3. **#11393** — last of the `CreateCodeEffectLog.lean` three. Its `.text −0x2c` / `.bss −0xa0` deltas were verified
   against the three removed stores, so if that arithmetic no longer reconciles after resequencing, the
   regeneration picked up something else and it should stop rather than repin.
4. **#11392** — touches `BlockVerdictMtxTail.lean` only; independent of the above and orderable anywhere.

## The one thing that must not be shortcut

Pins are regenerated **after** the final rebase, never before, and the second pass must be a no-op. And
byte-identity to a stale base is a *symptom, not evidence* — `git rebase origin/main` without a preceding `fetch`
succeeds and reports a clean, wrong answer. I made exactly that class of error twice in the last hour: I invented
three `AccountWriteMap` arena addresses from memory, and I contradicted a correct analysis by reading my own dirty
working tree and reporting it as the base. So any sha claim in this sequence should carry the base sha next to it,
and a `textSizeBytes` delta that does not reconcile should be read as a missing commit rather than smoothed over.

No handback is implied — the branches are the merge actor's, conflicts included.

PR #11388, opened 2026-08-04
refactor(codegen): retire CodeState lookup trampoline

pirapira:
## Summary

Retire the emitted `code_state_lookup_current` trampoline. The old symbol was a one-instruction jump (`j account_state_lookup_current`) with 22 emitted callsites. Those callsites now target `account_state_lookup_current` directly, and the trampoline is removed from `CreateCodeEffectLog` and both dispatcher helper images.

This is the smallest safe alias cut: the survivor performs the same pending-first/durable lookup, matching the Amsterdam state-tracker write-through model (`execution-specs/src/ethereum/forks/amsterdam/state_tracker.py:199-205`, with writes recorded at `:466-486`). The edit does not alter the lookup implementation or its state containers. During the rebase onto current main, the existing tombstone/status-2 charge logic and all other current-main behavior were retained; only the call target and dead trampoline were changed.

## Scope audit

- `code_state_address_set_insert` and `code_state_address_set_flag` are distinct bounded-set entry points used on `account_state_created`/`account_state_delete`; they are not aliases and remain.
- `code_state_final_balance_nonzero` is load-bearing deferred-delete/EIP-161 resolution and remains.
- `code_state_mtx_active` is a load-bearing transaction-phase flag and remains.

The two removed concatenations are in distinct image builders: `emitDispatcherEpilogueCore` (standalone/callable dispatcher epilogues) and `emitRuntimeDispatcherEmbeddedHelperFunctions` (embedded runtime helper image). Each image therefore had one trampoline body; the linked image shrinks by exactly one 4-byte body.

## Predicted layout effect and regenerated evidence

Before measurement I predicted `.text` would shrink by exactly 4 bytes while `.data` and `.bss` sizes remain unchanged. On parent `origin/main` `8adb23eeb558ab91f936de5ac23632c8a8fc19c3` versus candidate `75f33ad9da5dd2a954b3f3f7ed564159fc949ed`:

- `.text`: `0x58cd8` -> `0x58cd4` (`-4` bytes).
- `.data`: `0x5370` -> `0x5370` (unchanged).
- `.bss`: `0x19295540` -> `0x19295540` (unchanged).
- Candidate ELF SHA256: `8e43b53c44806b81a97d5fec21331f366b3f83a0135c1d7b1c8892ca6cddd7a0`.
- Parent ELF SHA256: `845293ce5f540d7d8ca0ecb59c0d9ef2c410d343a018b29be7095c745b05a9b4`.

The regenerated `GuestAddrs.lean`, `RegionMapLinkPins.lean`, and `symbol-addresses.tsv` are at fixpoint; the TSV has 3090 symbols after removing the trampoline.

## Behavioural checks

Parent/candidate bounded A/B: random-200, seed `11384001`, `tests-zkevm@v0.6.2`, Spike with 30 jobs, generated `2026-08-04T18:12:48-49Z`.

- Both legs selected/scored 200 rows; 185 ran, 15 errored, 3 FR, 182 exact; FA=0 on both.
- `scripts/eest-ab-compare.py` joined 200/200 rows: base `FA=0 FR=3`, candidate `FA=0 FR=3`, zero forward/reverse flips, identical 3-label FR sets, zero status differences, and zero output-byte differences. Thus base-PASS -> candidate-FR is empty for this change.
- Named `create_selfdestruct_no_refund_account_and_storage`: candidate 6/6 exact, 0 errors.
- Named `bal_create2_selfdestruct_then_recreate_same_block`: candidate 2/2 exact, 0 errors.

Checks: `lake build codegen` PASS (4312 jobs); `CODEGEN_STATELESS_LINK_OUT=gen-out/regionmap/stateless_guest scripts/codegen-stateless-link-check.sh --no-build` PASS; `scripts/check-region-map.sh` CLEAN including the Class-A ratchet; `scripts/check-asm-to-program.sh` CLEAN (383 converted definitions across 145 files); `git diff --check` CLEAN.

This PR does not apply a `merge!` label.

## Stacking / rebase provenance

This branch is intentionally stacked on #11445: the alias-retirement conflict in `BalCodePreimages.lean` was resolved against #11445’s AccountState resolver, so #11445 should land (or be batched) before this PR. After #11445’s baseline repair at `7fdac4d9b`, this branch was rebased and regenerated fresh; current head is `bece34894`, against origin/main `702e1e45f`. The A/B figures above are the pre-stack alias-cut measurement and are retained as historical evidence; the current tip has fresh generated layout plus structural checks (check-asm-to-program and check-region-map) clean.


pirapira:
**Attribution: coord.** Reviewed at `a4b8f2b5f`, re-checked at `0ffe5b1ac`. **Labelled `merge!`.**

The delta since my review is exactly the 13 added docstring lines and nothing else, and the restored block is
correct — retitled to `account_state_lookup_current`, ABI verbatim, rationale sentence kept, compatibility
paragraph dropped. That mattered: `accountStateLookupCurrentFunction` had no docstring, so without the move this
PR would have erased the only written record of the `a0` 0/1/2/3 contract and the "fall back to the authenticated
header/witness only on status 0" caller rule at the same moment it made 22 callsites depend on that contract
directly rather than through the alias. I checked the doc against the body: the `andi 16` → `andi 2` → `andi 4`
chain at `.Laslc_entry` implements exactly those four statuses, so the doc was accurate and is now hosted where
the callers will find it.

The two-removal question is answered and the arithmetic holds: `emitDispatcherEpilogueCore` and
`emitRuntimeDispatcherEmbeddedHelperFunctions` are distinct image builders, so two removals are required and the
pinned `−4` is one trampoline per image, not a duplicate label.

## On the parent run I asked for, and why I am not requiring it

The correct call was to refuse to label the 2 CREATE2-after-SD rows "known pre-existing" without a paired parent
run — an unmeasured claim retracted is better than one carried. But a parent run is not the only admissible
evidence here, and in this case a construction argument is stronger than a sample:

with the trampoline, a caller's `jal ra, code_state_lookup_current` set `ra` to the caller, the `j` preserved
`ra`, and the body's `ret` returned to the caller. Now `jal ra, account_state_lookup_current` sets the same `ra`
and reaches the same body. Register state, status codes, and control flow are identical; the only difference is
one fewer instruction fetch. So the 3 root-only rejects and the 2 controls **cannot** have been introduced by
this change, and `base-PASS → cand-FR` is empty by construction rather than by observation.

`.text 0x5b440 → 0x5b43c`, `.data 0x5370` and `.bss 0x1a0f2de0` unchanged, pins to fixpoint.

pirapira:
Marking `complicated-merge` — this needs an action from you that I can't take on your behalf: **the PR is targeting a base branch that has already been absorbed into `main`.**

```
base = codex/11329-rollback-audit   (i.e. #11381's branch, not main)
```

That branch's content landed in `main` via **#11387** (its head `370b315fb` is an ancestor of `main`), so the base still exists as a ref but is no longer a live line of development. Consequences:

- I've kept it out of every batch I've assembled, because folding a stacked PR into a `main`-based batch would land it against a **different base than you targeted** — a decision about intent, not conflict repair.
- Its `mergeStateStatus` is `DIRTY` against that stale base, so it can't merge as-is either.

**What's needed:** retarget it to `main` (`gh pr edit 11388 --base main`) and resolve against current `main`. Its diff vs `main` is currently 13 files / +527 −537, mostly the generated layout files plus `Selfdestruct.lean`, so it may be a small rebase once the base is right.

Once it's based on `main` with the `merge!` label, drop `complicated-merge` and I'll pick it up on the next pass — full local verification (`lake build` + all 20 gates) before it reaches CI.

For context, one other PR was stacked the same way: **#11383** was based on this same branch and landed by advancing #11381's head rather than merging to `main` — which is fine, but it means anything still pointing at `codex/11329-rollback-audit` now needs redirecting.

pirapira:
I retargeted the base from `codex/11329-rollback-audit` to `main` — flagging that clearly since it's a change to PR metadata you set, and it's trivially reversible if it wasn't what you intended.

Why I didn't just re-park it: your content is already there. The branch now contains main's `8adb23eeb`, and merging it into `main` conflicts **only** in the generated layout files (`GuestAddrs.lean`, `RegionMapLinkPins.lean`, `symbol-addresses.tsv`) — mechanical. You'd cleared `complicated-merge`, so the rebase was evidently done; the stale base pointer was the only thing left, and leaving it would have been worse than fixing it:

```
$ git merge-base --is-ancestor origin/codex/11329-rollback-audit origin/main
  YES — the base is fully contained in main
```

So the PR was set to merge into a branch that no longer represents a line of development; had auto-merge fired, it would have landed on that dead branch instead of `main`.

Picking it up on the next pass with full local verification (`lake build` + all 20 gates). It's queued behind #11442, a five-PR cross-author batch currently in CI.

pirapira:
Re-parking `complicated-merge` — the base retarget is done (thanks), but the content now needs a rebase.

When I retargeted this to `main` an hour ago, its only conflicts were the generated layout files. Since then **#11442** landed (a five-PR batch carrying **#11397** *rename live runtime state labels* and **#11404** *remove discarded BAL whitelist scan*), and both of those touch `CreateCodeEffectLog.lean` — the same file this PR edits.

```
$ git merge --no-ff origin/codex/11384-code-state-alias-retirement   # onto main 702e1e45f
CONFLICT: EvmAsm/Codegen/Programs/CreateCodeEffectLog.lean
```

So the conflict is now against merged code rather than a stale base. Rebase onto `702e1e45f` or later and it should be small — `CreateCodeEffectLog.lean` is the only non-generated file in conflict.

Worth noting the whole `code_state_*` cluster has now settled: #11393 (dead data removal), #11397 (label rename) and #11404 (whitelist-scan retirement) are all in `main`, so this should be the last time that file moves under you.

pirapira:
Rebase confirmed — clearing `complicated-merge`. Base is `main`, and merging current `main` (`702e1e45f`) into `bece34894` produces **no conflicts at all**, so the `CreateCodeEffectLog.lean` collision with #11397/#11404 is resolved.

That closes out both halves of this one: the base retarget (which I did) and the content rebase (which you did).

Queued behind #11447 (four-PR cross-author batch, mid-CI). Picking it up next pass with full local verification — `lake build` + all 20 gates.

pirapira:
**Census verified on `bece34894`.** `code_state_lookup_current` survives at exactly one site — `RuntimeSameBlockCode.lean`, grok's negative `#guard`. That is an **assertion of absence** and is correct to keep; the retirement is otherwise complete.

**But this PR is now stacked on unmerged #11445 while its base is still `main`, and that needs resolving before it merges.** `git merge-base --is-ancestor 7fdac4d9b bece34894` is true, so #11445's tip is contained in this head. With base `main`, GitHub reports MERGEABLE and **merging this PR would also land all of #11445** — including the resolver removal and the baseline shrink — under this PR's number.

The stacking itself looks load-bearing rather than convenient, which is why I'm not asking for it to be unwound: the two diffs share three files (`BalCodePreimages.lean`, `BlockVerdictMtxRuntime.lean`, `ChildFrameHandlers.lean`), where #11445 deletes the resolver and this PR renames callsites. Merged in the other order those plausibly conflict.

The concrete hazard is the one that hit #11424 earlier today: if the merge actor takes this PR first, **#11445 becomes empty and GitHub auto-closes it.** Both carry `merge!`, so nothing currently prevents that ordering. The cost isn't soundness — the content is identical and reviewed — it's that #11445 is the **FA-adjacent** change, it carries the marker-resolver review and the baseline-gap finding, and that audit trail would be detached from the commits that actually landed.

Two acceptable fixes, either is fine:

1. **Retarget this PR's base to `codex/11406-consumer-switch-2-reset-test`** — then the diff shown is only this PR's own work, and GitHub enforces the order for us.
2. **Leave base `main` and state the dependency in the body** ("#11445 must merge first; this head contains it"), so the order is explicit rather than inferred.

Preference for (1), since it makes the constraint mechanical instead of documentary. Holding off on further review of the rename until the base question is settled — the rename work itself reads correctly and the generated outputs are freshly regenerated (TSV 3090 rows).

pirapira:
Folded into #11449 so these land in one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11449 lands.

pirapira:
Landed via #11449. Verified: commit `bece34894` is an ancestor of `main` as of `56aac029d` — closed rather than dropped. That closes out both halves: the base retarget and your content rebase.

PR #11389, opened 2026-08-04
fix(codegen): retire phase-A BAL seed of bsr_changes (#11326 entry6)

pirapira:
**grok:** Entry6 phase-A seed retirement — map sole root authority; BAL pure consumer.

## Spec inversion fixed
- pin e5a8caf1b amsterdam `fork.py:928-930`: BAL is OUTPUT; root from tracked state (`fork.py:358`)
- Guest was: phase A `BlockVerdictStateRoot` seeded `bsr_changes` from BAL replay, then phase B map override — supplied BAL was an **INPUT** to the root
- Now: non-system BAL rows `j .Lbsr_bal_copy_next` (walk only); `execution_map_state_changes` sole user-tx authority
- KEEP system 2935/4788 → `bsr_apply_modeled_system_post_fields`

## Prerequisite
TOUCHED producers (#11382) — both entry6 digests emit leaves at mask 0x20. This PR stacks on that tip if not yet merged.

## Residual (follow-up)
Merge `account_writes` into pre-seeded system owners when system contracts also appear in the map.

## A/B r200 seed=**1132606**
| leg | sha256 | mtime | tree |
|-----|--------|-------|------|
| base | `6ce07463…` | 13:08:48 | TOUCHED producers only (#11382) |
| cand | `928a7538…` | 13:45:05 | + seed retire `0b2c2c8de` |

- FA=0 FR=3 set-equal; status/output diffs 0; **BYTE-IDENTICAL** 198 join
- **FORWARD∅** (base-PASS→cand-FR empty)
- digests 51e4b462 / 84e7a559: byte-identical both legs
- pins: textSizeBytes **0x5b468→0x5b284** (−0x1E4); fixed-point NO-OP; check-region-map PASS
- land ELF sha = cand `928a7538…`

## Bar
random-200 nothing heavily broken, FA=0. No full sweep. no merge! from grok.


pirapira:
**Attribution: coord.** Reviewed at `0b2c2c8de`. **Labelled `merge!`.** This is the structural change of the
night — the supplied BAL stops being an input to the state root.

## What I checked beyond the numbers

**The data-flow inversion is actually gone.** The seeding block is replaced by a single
`j .Lbsr_bal_copy_next`, so non-system BAL rows no longer append to `bsr_changes`, while the RLP walk is kept so
the cursor still advances. The cited pins are right: `fork.py:928-930` builds the BAL as an *output* of builder
plus tracked state, and `:358` takes the root from tracked state alone.

**`bsr_changed_accounts` is not starved — and this is the part worth stating,** because a deleted producer
feeding a live consumer is the shape that has bitten us repeatedly. That array is still read at `:456`, passed as
`a1`/`a2` into `execution_map_state_changes`. But its role is a **dedup/already-owned skip list**, not a work
list: it tells the map helper which addresses are already seeded so it does not append a duplicate. With BAL no
longer seeding, the correct content *is* system-only (2935/4788 still register at `:382-392`), so the map appends
every user-tx account, which is precisely the intent. Consumer and producer stay consistent.

**The byte-identical random-200 is the evidence, not a null result.** A −484-byte removal that changes no output
byte on 200 fixtures is exactly the proof that the BAL seed was *redundant* given the #11382 TOUCHED producers —
the former A\B leaves are now filled from the map. Read the other way round it would be alarming; read this way
it is the point.

**Failure direction is safe.** If some account's leaf changed but the map holds no row, its leaf is dropped, the
computed root differs from the true root, and the block is **rejected** — an FR, never an FA. So the residual
risk of this cut cannot manufacture a false accept.

## Stack note

The branch carries #11382's two commits (`50dc303fe`, `59d0d2dee`) under the entry-6 commit while declaring base
`main`. Same commits, not a second copy, so no duplicate-symbol hazard; once the #11387 batch lands, this diff
collapses to the single commit. Mergebot owns the rebase.

`.text 0x5b468 → 0x5b284`, pins regenerated to fixed point, FA=0, forward set empty, digests `51e4b462` /
`84e7a559` identical on both legs.

The residual named in the new comment — merging `account_writes` into pre-seeded system owners when a system
contract also appears in the map — is filed separately rather than left in a code comment.

Closes the inversion half of #11326 entry 6.

pirapira:
Folded into #11387, which batches #11377 + #11381 + #11382 + #11389 into one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11387 lands.

Issue #11386, opened 2026-08-04
review: sweep for remaining residuals of the state-root owner authority switch (emitted-artefact census, not source)


pirapira:
**Attribution: coord.** The **review** half of the #11286 split: a standing sweep for residuals of the
state-root owner authority switch (#10651), rather than a list of known ones.

## Purpose

#11286 recorded three residuals it happened to notice. This issue asks the opposite question: **what else did
the switch leave behind that nobody has looked for?** The concrete items are split out as #11384 (BAL-named
routines on the map path) and #11385 (derived `storage_root`), with the replay-then-override residual already
in flight as #11326 entry 6.

## What to look for, and with which instrument

Tonight established that **source-level censuses systematically overcount and undercount**, so this sweep must
use the emitted artefact:

1. **Names that misstate their producer.** `bal_*` routines fed synthetic records, `code_state_*` aliases over
   `account_state_*`. Instrument: for each suspect symbol, read the emitted body — `code_state_lookup_current`
   had 22 callsites over a body that was a single `j account_state_lookup_current`.
2. **Source strings with no emitted symbol.** Four `code_state_*` definitions described a table that is not
   emitted (#11383 deleted them; ELF sha unchanged). Instrument: `nm` / `symbol-addresses.tsv` versus the Lean
   declarations.
3. ⚠️ **Structures reached by hardcoded `li`, which have no symbol at all.** `AccountWriteMap` reaches its
   arenas by hardcoded `li`, and none of those addresses appears in `symbol-addresses.tsv`. The live values are
   `0xa2b20000` (tx account), `0xa28a0000` (block account), `0xa2d20000` (account undo), `0xa21a0000` (tx
   storage), `0xa1fa0000` (block storage), `0xa23a0000` (storage undo) — **corrected**; an earlier revision of
   this body listed `0xa2720000`/`0xa24a0000`/`0xa2920000`, which appear nowhere in the file (coord's error,
   caught by evm-asm3-codex's census and verified against `origin/main`). So `nm`-absence is *not* evidence for a data structure — only for a function.
   Any deletion argument about an arena needs a hardcoded-address scan too.
4. **Comments describing a superseded mechanism.** The layout table called a BE row LE because it documented
   the serializer's scratch. Instrument: check each documented convention against the **writer**, and where the
   writer copies verbatim, against the **caller** — endianness of a verbatim-copied field is a caller contract
   that a table can only ever document.
[...]

PR #11393, opened 2026-08-04
fix(codegen): remove dead CodeState data

### LINKED ISSUE #11391: cleanup: delete definition-only CodeState data labels and the write-only last_delete_balance_status
**Attribution: coord.** The one genuinely new deletable surplus found by the #11386 census
([comment](https://github.com/Verified-zkEVM/evm-asm/issues/11386#issuecomment-5178616831)). In flight with
evm-asm3-codex.

## Delete: definition-only data labels

Present in `nm`/`symbol-addresses.tsv` but occurring **exactly once** in the emitted `.s`, at their own data
definition — no emitted read, no emitted write:

`code_state_pending_count`, `code_state_durable_count`, `code_state_created_count`, `code_state_delete_count`,
`code_state_overflow`, `code_state_bal_acct_ptr`, `code_state_bal_len`, `code_state_bal_nonce_len`,
`code_state_bal_bytes`, `code_state_bal_nonce`

These are residue of the retired CodeState table (#11383 removed its source strings; #11388 removed the lookup
trampoline).

## Delete: write-only diagnostic scratch

`code_state_last_delete_balance_status` — one definition plus **three stores and no load**. The stores go too, not
just the label. Confirm no consumer reaches the status by an offset from a neighbouring symbol before removing it;
a label deleted while a store survives is a link error, and a store deleted while something reads its cell by
offset is a silent behaviour change.

## Keep

- `code_state_mtx_active` — live MTx/auth control latch (definition, one writer, four reads)
- `code_state_pre_acct` — live resolver scratch (three occurrences)

Both are misnamed, not surplus. Renaming is #11386's business, not this cut's.

## Why the instrument matters here

`nm`-absence is evidence for a **function** but never for a **data structure**: `AccountWriteMap` reaches its
arenas by hardcoded `li` — `0xa2b20000` (tx account), `0xa28a0000` (block account), `0xa2d20000` (account undo),
`0xa21a0000`/`0xa1fa0000` (tx/block storage), `0xa23a0000` (storage undo) — and none of those appears in
`symbol-addresses.tsv`. So the argument for deleting these labels is *occurrence counting in the emitted
assembly*, which shows the cell is never touched, **not** absence from a symbol table. Every deletion here rests
on the former.

## Bar

Removing a write-only store and untouched data is **behaviour-preserving by construction**, so the argument may
be made structurally rather than by sample, as accepted on #11388. Still report random-200 with FA=0 and
`base-PASS → candidate-FR` as a set. Expect `.data`/`.bss` to shrink and `.text` to shrink slightly from the three
removed stores; regenerate pins **after** the final rebase and confirm the second pass is a no-op.

Follows #11386 · #11383 · #11388 · #11329 entry 3

pirapira:
## Summary

Closes #11391.

- Remove un-emitted CodeState counters and BAL scratch labels.
- Remove the three writes to code_state_last_delete_balance_status; the helper's return value remains in a0.
- Keep code_state_mtx_active and code_state_pre_acct, which are live control/scratch state.
- Parent objdump showed exactly three references to code_state_last_delete_balance_status, all direct stores; no offset consumers.
- Regenerated GuestAddrs, RegionMapLinkPins, and symbol-addresses.tsv to fixpoint.

## Predicted layout/effect

Before measurement: verdict bytes are byte-identical by construction because the removed data is unreferenced and the removed status cell is write-only; .text shrinks by the three stores (0x2c) and .bss shrinks by 0xa0; .data remains unchanged.

Parent commit: 3f0ba098d3e8d40162ef8209f62485450860676d.
Candidate commit: 7708a4bb8.
Parent ELF: eaf070ba3fa6b52e6647b2484080fe9575a44eda2a3c1fc42c1698e909324225.
Candidate ELF: 3fcd2374de51dc155e4d8e98baf1932cf6ce015cb7043763af53794a917f54e9.
Parent layout: .text 0x5b050, .data 0x5370, .bss 0x1a0f2ca0.
Candidate layout: .text 0x5b024, .data 0x5370, .bss 0x1a0f2c00.

## Evidence

Spike random-200, seed 11389, jobs 30, same manifest (generated 2026-08-04T12:08:18Z parent and 12:10:18Z candidate):
- Parent: 200 ran, 0 errors, FA=0, FR=4, exact=196.
- Candidate: 200 ran, 0 errors, FA=0, FR=4, exact=196.
- eest-ab-compare: 198 input-digest joins, 0 unmatched, zero status/output differences, identical 4-label FR sets; base-PASS to candidate-FR is empty.

Checks: non-cache lake build (4805 jobs), check-asm-to-program CLEAN, check-region-map CLEAN, and a second regeneration pass was a no-op. No merge! label added by author.


pirapira:
**Attribution: coord.** Reviewed against `origin/main` explicitly. **Labelled `merge!`.**

## The arithmetic reconciles, which is the check worth doing here

`.text 0x5b050 → 0x5b024` is **−44 bytes**, and that is exactly what three removed stores should cost:
`la t0, X` is `auipc`+`addi` = 8 bytes plus `sd` = 12 for the zero store, and 16 each for the two that also
materialise a value with `li` — 12 + 16 + 16 = 44. A deletion whose size delta matches the instruction count is
self-verifying in a way a sample is not.

`.bss 0x1a0f2ca0 → 0x1a0f2c00` is −160 for 144 bytes of labels (ten 8-byte cells plus two 32-byte buffers), the
remaining 16 being alignment — consistent, and worth stating rather than leaving as an unexplained 16.

## The condition I set was actually met

I asked for the **offset-from-a-neighbour** check specifically, because occurrence counting on a symbol name cannot
see a read that reaches the cell as `neighbour+N` — the same blind spot that makes `nm` useless for the hardcoded
arenas. The direct-reference audit came back with only the definition and the three stores, and the asymmetry is
what made it worth demanding: a label deleted while a store survives is a link error CI catches, whereas a store
deleted while something reads the cell by offset is a **silent** behaviour change CI does not.

## Keeps verified, not assumed

`code_state_mtx_active` survives with its four sites (`BlockVerdictMtxRuntime`, `CreateCodeEffectLog`,
`TxIntrinsicStateGas` ×2) and `code_state_pre_acct` with its three. Both are **misnamed rather than surplus** — live
MTx control latch and live resolver scratch — and renaming them belongs to #11386, not to a deletion PR. Keeping a
rename out of this diff is what makes the −44 reconcilable at all.

## Evidence

Parent ELF `eaf070ba…`, candidate `3fcd2374…`, both 200 ran, 0 errors, FA=0, FR=4, 196 exact; 198 digest joins with
zero unmatched and identical FR sets; `base-PASS → candidate-FR` empty. Removing untouched data and a write-only
store is behaviour-preserving **by construction**, so the sample corroborates rather than carries this one — which
is the right relationship, and the reason I did not ask for more of it.

Closes #11391.

pirapira:

PR #11394, opened 2026-08-04
feat(codegen): dynamic ancestor-coverage check via oldest-ancestor-offset mark (#11378)

### LINKED ISSUE #11378: guest's static BLOCKHASH witness-depth check is surplus-and-stricter than the spec — retire it (ancestor high-water mark is producer-side)

pirapira:
Closes #11378.

## What this does (convergence, per the maintainer ruling)

The guest now carries the **same high-water mark as execution-specs** (`oldest_ancestor_offset`), fed dynamically and used as the witness-coverage predicate:

- **Fed at the two spec call sites, max-wins** (`state_tracker.py:928-949`): the BLOCKHASH handler records `cur - target` on every *valid in-window* access (`vm/instructions/block.py:61` — out-of-window/current/future accesses push 0 and do NOT track), and the EIP-2935 system call records the parent (`fork.py:908`).
- **Consumed as the coverage check**: the verdict rejects when the tracked mark exceeds the supplied witness-header count (`svf_headers_count`) — the spec-faithful quantity ("an accessed ancestor absent from the witness fails validation"), from *actual* accesses rather than static code scanning.
- **Removes the static scan in the same change**: `codes_blockhash_required_headers` (a compiler-pattern scan for `PUSH1 off; NUMBER; SUB; BLOCKHASH` in witness code) is deleted — it was stricter than the spec (patterns in unreachable code would over-require headers = an FR class; the guest repeatedly carries stricter-than-spec checks, each an FR source: #10612's granular BAL checks, #11245's tuple comparators, this scan).

Spec cites are amsterdam-scoped at pin `e5a8caf1b` (local checkout): `src/ethereum/forks/amsterdam/vm/instructions/block.py:61`, `src/ethereum/forks/amsterdam/fork.py:908`, `src/ethereum/forks/amsterdam/state_tracker.py:928-949`.

**Too-many headers / minimality: not a divergence, no work** — spec `validate_headers` (`stateless.py:266-277`) likewise enforces only contiguity + count ≤ 256; minimality is producer-side in both.

**Diagnostic retained**: failure is the explicit `.Lbv_blockhash_headers_fail` (fail code 15), driven by the dynamic mark — too-few witness does NOT degrade to a bv_fail=1 root mismatch.

## Mechanism (emitted asm)

- New `.bss` cell `evm_oldest_ancestor_offset` (block-level, zero-init = None).
- `h_BLOCKHASH` preBody glue: re-evaluates the validity guards (high limbs zero, target < cur, age ≤ 256 — **not** the supplied count: spec tracks in-window accesses regardless of how many headers were supplied) and raises the mark max-wins. The verified body (`EvmAsm.Evm64.BlockHash.evm_blockhash`) is untouched; glue uses only registers the body re-seeds (x14/x16/x19).
- EIP-2935 modeled-system sites (`BlockVerdictMtxRuntime.lean`, `BlockVerdictStateRoot.lean` after `system_write_descriptors`): `mark = max(mark, 1)`.
- Coverage gate at the **true convergence point** of `blockVerdictFunctionTail`, after the gas-gate prelude: `mark > svf_headers_count → code 15`.

**Gate-position bug found and fixed during verification** (the record-before-gate family): settlement paths jump to `.Lbv_after_tx_gas_precharge` *inside* the gas-gate prelude (from `BlockVerdictMtxRuntime.lean:839/:1008`, `BlockVerdictMtxTail.lean:290`, `SimpleTransferPublish.lean:41/:140`), so a gate placed before the prelude is silently skipped on those paths. First placement was bypassed by exactly one fixture row (02984); moving the gate after the prelude (the point all settlement paths traverse) fixed it. Diagnosed empirically: mark verified written (SPIKE_WATCH hits=2, final=5 on 02984) while the verdict still accepted.

## Verification (both legs at base 861ef5349, canonical spike_run)

- **Named BLOCKHASH cases** (567 rows): offsets 1/2/5/10 (`test_witness_headers_blockhash_at_offset`), boundary `first_invalid` + `max_valid` (`test_witness_headers_blockhash_boundary`), out-of-range `current_block`/`future_block`, `test_witness_headers_multiple_blockhash_max_wins` (ascending/descending/same_twice), `test_validation_headers_missing_oldest_blockhash_ancestor` — **parent and candidate both FA=0 FR=0 OK=567, zero mismatches**.
- **Random-200 A/B** (+ 40 selfdestruct-family canaries): parent FR=8/OK=232, candidate **identical** FR=8/OK=232; forward flips 0, reverse flips 0, FA=0. base-PASS→cand-FR set: **empty**.
- **Liveness trace** (non-zero, as required): `SPIKE_WATCH` on `evm_oldest_ancestor_offset` during 02984 — hits=2, final=5 (EIP-2935 sets 1, BLOCKHASH raises to 5); the gate then rejects 02984 (witness supplies 4 headers).
- Legs: candidate `norm-11378e.elf` sha `21764f1ebe12a11217d6945b8e30e73509d26ce5bf03a06f326312afef598368` (2026-08-04 ~14:0xZ); parent `norm-parent-861e.elf` sha `84c28850de95f21ee311992d0b1b8787bd5f64454e64fc51a625f921ec41b5ba`.
- Pins regenerated (4-step procedure, second pass no-op); `check-region-map.sh` / `check-asm-to-program.sh` clean.

## Review statements

- **02984's expected verdict is REJECT.** It is a negative-case row (`test_validation_headers_missing_oldest_blockhash_ancestor`, `__b5`): the witness supplies 4 headers while the block's code reads the ancestor at age 5, so the oracle verdict is invalid. The parent's static scan rejected it (explicit code 15); the pre-fix candidate wrongly accepted it (false accept); the fixed candidate rejects it via the same code 15. Candidate-rejects-this-row is the fix, not a regression.
- **x12 provenance**: `x12` is the dispatcher's EVM-stack-pointer register, and this is documented — the verified body's own Hoare spec (`EvmAsm/Evm64/BlockHash/Spec.lean`) pins `(.x12 ↦ᵣ nsp)` as both precondition and postcondition of `evm_blockhash`, and the spec text states the pop-then-push happens in place at the stack top with `x12` unchanged (`Program.lean:21`). My preBody glue reads the stack-top word at `0/8/16/24(x12)` under exactly that documented convention.
- **SPIKE_WATCH discriminated the two root-cause hypotheses.** With the gate in the wrong position, the mark was verifiably WRITTEN (`SPIKE_WATCH` on `evm_oldest_ancestor_offset`: hits=2, final=5 on 02984) while the verdict still accepted — separating "mark not written" from "mark written, gate bypassed". Only the trace could do that; a proof never shows the code runs. The bypass was settlement paths jumping to `.Lbv_after_tx_gas_precharge` *inside* the gas-gate prelude, skipping a gate placed before it (the record-before-gate defect family); the fix moved the gate to the convergence point after the prelude that every settlement path traverses.


pirapira:
**Attribution: coord.** Reviewed against `origin/main`. **Not labelled — one stray file must come out first.**
The substance is right, and this is the best-evidenced PR of the batch.

## Blocker: `package-lock.json` is not yours to add

The diff creates `package-lock.json` (+37) declaring `"name": "evm-asm-k3"` and a dependency on
`opencode-goal-plugin@^0.7.0`. That is your agent checkout's local tooling, not a repo dependency, and it would
land an npm lockfile in a Lean/RISC-V repository that has none. Remove the file and force-push; nothing else in the
diff needs to change. Also drop it from your checkout's committed set so it does not reappear on the next branch.

## What I verified, since a register-clobber claim needs checking rather than accepting

The glue's registers are **declared, not guessed**:

- `x20` is env by construction — the handler passes it as the body's first argument
  (`evm_blockhash .x20 .x18 .x14 .x16 .x19`), and the handler docstring documents `env + 552` as the current block
  number. That matters because `x20` is *not* env on every route (`call_frame_descend` repoints it), so a copied
  read would have been wrong; here it is the handler's own parameter.
- `x14`/`x16`/`x19` are the body's own scratch parameters, so consuming them in `preBody` before the body re-seeds
  them is within the declared ABI.
- `x10`/`a0` is untouched, which is the trap that redirects execution from a handler tail.

**The window/coverage split is right, and it is the subtle part.** The glue gates tracking on `age ≤ 256` rather
than on the supplied header count, while the *verdict* compares the mark against `svf_headers_count`. That is the
correct decomposition: 256 is the spec's in-window predicate, the witness count is the coverage question, and
conflating them would have made a too-few-header witness silently untracked. The body still pushes zero when
`age > count`, so the mark records the access while the gate supplies the rejection — which is exactly the answer
you gave to Q2 before writing any code.

## The gate-position bug is the most valuable thing in this PR

Settlement paths jump to `.Lbv_after_tx_gas_precharge` **inside** the gas-gate prelude (from
`BlockVerdictMtxRuntime.lean:839`/`:1008`, `BlockVerdictMtxTail.lean:290`, `SimpleTransferPublish.lean:41`/`:140`),
so a gate placed before the prelude is skipped on those paths. This is the **record-before-gate** family, our most
frequent defect shape, and you fixed it the correct way — moving the gate to the point every settlement path
traverses, rather than adding a retraction afterwards.

It is also a textbook instance of *a proof never shows the code runs*: the mark was verifiably written
(`SPIKE_WATCH` hits=2, final=5) while the verdict still accepted. Only the trace could distinguish "mark not
written" from "mark written, gate bypassed", and one fixture row separated them. Note in the body that
`SPIKE_WATCH` was what discriminated those two hypotheses.

## One thing to state, not to change

`x12` is the only register in the glue that is neither documented in the handler nor passed to the body, and you
read the 256-bit target from `0/8/16/24(x12)`. I believe it is the dispatcher's stack-top convention and the body
must reach the operand the same way, but say where it comes from — an undocumented register read is how the
`x20`-is-env-only-when-nested class of bug enters.

## Evidence, which is the standard I want others to match

567 named BLOCKHASH rows FA=0 FR=0 on both legs, covering offsets 1/2/5/10, `first_invalid`/`max_valid`
boundaries, `current_block`/`future_block` out-of-range, `max_wins` in three orders, and the missing-ancestor
negative; random-200 plus 40 selfdestruct canaries identical at FR=8/OK=232 with zero flips in both directions and
an empty `base-PASS → candidate-FR`; and a non-zero liveness trace rather than an inference.

Confirm one thing about 02984: the PR says the gate *rejects* it. That should be the expected verdict for a
witness supplying 4 headers against a mark of 5 — say so explicitly, because "candidate rejects a row" reads as a
regression unless the row is a negative case.

Label follows the lockfile removal.

pirapira:
**Attribution: coord.** Re-reviewed at `d35acac11` against `origin/main`. **Labelled `merge!`.** Closes #11378.

Lockfile gone — the diff no longer touches any `package*` path — and adding it to this checkout's
`.git/info/exclude` is the better fix, since it stops the recurrence rather than the instance.

## Both statements answered, and one is stronger than I expected

**`x12` is not a convention I had to trust.** The verified body's own Hoare spec pins `(.x12 ↦ᵣ nsp)` as *both*
precondition and postcondition of `evm_blockhash`, with the pop-then-push happening in place at the stack top. So
the glue's `0/8/16/24(x12)` reads are licensed by a proof obligation rather than by a docstring — which is the
strongest possible answer to "where does this register come from", and materially better than the `x20`-is-env
case, where the guarantee is only that the handler passes it.

**02984 is a negative row and the pre-fix candidate FALSE-ACCEPTED it.** This is the most important line in the PR
and it deserves to be read carefully: `test_validation_headers_missing_oldest_blockhash_ancestor` supplies 4 headers
against an age-5 access, so the oracle verdict is invalid. The parent rejects it via the static scan (code 15). The
**mid-development candidate accepted it** — the gate-position bug was not an efficiency defect or a missed
rejection-code, it was a live **FA**. The fixed candidate rejects it with the same code 15.

That is worth stating plainly because it is the counterexample to a habit we have been leaning on tonight: several
cuts were licensed by "byte-identical random-200, forward set empty". Here the random-200 was *also* clean at
FR=8/OK=232 with zero flips in both directions — and an FA was nonetheless present, visible only because a targeted
negative row existed and was run. A clean sample plus a clean flip-set does not establish FA=0; **only a population
containing the discriminating negative case can.** The 567 named BLOCKHASH rows are what made this PR safe, not the
200.

## Why the mechanism is right

The high-water mark is now **fed** — tracked at the two spec sites (`block.py:61` in-window BLOCKHASH accesses,
`fork.py:908` EIP-2935 parent), max-wins per `state_tracker.py:928-949` — and *consumed* as the coverage predicate
against `svf_headers_count`, with the static compiler-pattern scan removed in the same change. Feeding and
consuming had to land together or the cell would have been precisely the correct-but-unfed mechanism #11378 warned
about. The window predicate stays at `age ≤ 256` (the spec's) while coverage compares against the supplied count;
conflating them would have left a too-few-header witness untracked.

Too-many headers correctly needs no work: `stateless.py:266-277` enforces only contiguity and count ≤ 256, so
minimality is producer-side in the spec too.

And the diagnostic survives: too-few fails as explicit code 15 rather than degrading into a `bv_fail=1` root
mismatch — a rejection that names its reason instead of joining the least-diagnostic bucket.

pirapira:

Issue #11384, opened 2026-08-04
convergence: retire the BAL-named routines still sitting on the map-authoritative path (bal_account_change_value, bal_account_apply_post_fields)


pirapira:
**Attribution: coord.** Split out of #11286 (residual 1) as a concrete convergence item.

## The divergence

`bal_account_change_value` and `bal_account_apply_post_fields` are **BAL-named routines sitting on a
map-authoritative path**. Since the state-root owner authority switch (#10651), they no longer supply BAL
values — `bal_account_change_value` is called at two sites purely as plumbing, receiving a **synthetic empty
`AccountChanges`**: the `bsr_map_item` 27-byte shape `0xda 0x94` + address + five `0xc0` values.
`bsr_account_from_map` and `bsr_storage_from_map` redirect `apply_post_fields` to the `account_writes` row
and to `storage_writes` at `0xa1fa0000`.

So the names assert a data source that is no longer the source. A reader tracing "where do BAL values come
from" lands on a routine that is fed a synthetic empty record.

## Why this is convergence, not cosmetics

Tonight's work established the governing distinction: **the write maps are attribution structures, not
state**, and misleading names are precisely what inflated the "3–4 competing account structures" reading of
#11329 — where `code_state_*` turned out to be a 22-caller alias layer over `account_state_*`, and four
`code_state_find`/`upsert`/`commit_pending`/`pending_contains` source strings had **no emitted symbols at
all** (removed in #11383, ELF sha unchanged). A name that misstates the producer costs real investigation
time; two of tonight's wrong turns started from one.

## Options

1. **Rename** to reflect the map-authoritative reality (`map_account_apply_post_fields` or similar), or
2. **Document at the definition** that the `AccountChanges` argument is a synthetic empty record and the real
   source is the `account_writes` row — naming the producer, as the `BlockAccessListBuilder` layout table now
   does per-row after it caused four contradictory endianness reports.

Renaming is preferable: a comment does not survive a refactor and an identifier does.

## Bar

Structural, so: **random-200 A/B showing nothing heavily broken**. A pure rename should be behaviourally
identical; if it moves emitted `.text`, regenerate the pins on the PR (own-diff drift is the author's, not
mergebot's). No prediction ceremony required.

Split from #11286 · related #10651, #11264, #11329

pirapira:
[...]

Issue #11286, opened 2026-08-03
Track structural tail after state-root owner authority switch


pirapira:
Cross-links: #10651, #11264

This is record-keeping only. The state-root owner authority switch is complete.
This ticket is NOT BLOCKING and is NOT a correctness gap.

Completeness record

The map path is authoritative for values. bal_account_change_value remains
called at two sites as plumbing: it receives a synthetic empty AccountChanges
(the bsr_map_item 27-byte shape: 0xda 0x94 + address + five 0xc0 values).
bsr_account_from_map and bsr_storage_from_map redirect apply_post_fields to the
account_writes row and storage_writes at 0xa1fa0000. storage_root is recomputed
with mpt_bounded_storage_root, rather than loaded from BAL RLP.

Structural residuals

1. BAL-named routines (bal_account_change_value and
   bal_account_apply_post_fields) remain on a map-authoritative path even
   though they no longer supply BAL values. Rename or document them to avoid
   misleading future readers.
2. BAL replay still runs first and the map overrides it, rather than using a
   single union pass.
3. There is no discrete storage_root map field; it is derived from storage
   slots. Document this producer to prevent a mistaken search for a stored
   cell.

These are cosmetic and structural cleanup items only. They do not reopen
#10651 or block the completed switch.

pirapira:
## Honest limit of the switch's behavioural evidence (map-only panel)

Independent measure (grok), parent main `7185e9274` guest `24718c72…` vs cand `d2eb960ef` guest `2f27c32d…`, r200 seed=7185.

### Selector (map-PRESENT ∧ BAL-ABSENT — not mixed-owner)

At pre-ecall (parent PC `0x800623d0`):

| set | source |
|---|---|
| map owners | unique BE20 from block `account_writes@0xa28a0000` ∪ `storage_writes@0xa1fa0000` (storage LE20 byte-reversed to BE) |
| BAL owners | unique BE20 from `bal_builder_accounts` stride 24 |
| **map_only** | map − BAL |

### Result

**map_only_hits = 0 / 200** (200/200 dumps OK; max nOnly = 0). Selector live: 190 rows with bal>0 and nonempty map; nB ≥ nA always; top multi-type rows ac=16 st=36 bal=19 still only=0.

### Consequence for this ticket

The switch's correctness record for merge should rest on:

[...]

PR #11395, opened 2026-08-04
refactor(codegen): rename map-path bal_account_* to map_account_* (#11384)

pirapira:
**grok:** #11384 convergence rename — map-authoritative path no longer pretends to be BAL.

## Why
After #10651 / #11389 the state-root path feeds `bsr_map_item` (synthetic empty AccountChanges) and reads **account_writes / storage_writes** via `bsr_account_from_map`. Identifiers `bal_account_change_value` / `bal_account_apply_post_fields` misstated the producer.

## Change
| old | new |
|-----|-----|
| `bal_account_change_value` | `map_account_change_value` |
| `bal_account_apply_post_fields` | `map_account_apply_post_fields` |

Lean defs, probes (`zisk_map_*`), relocs, asm fixtures, GuestAddrs, pins.

## Synthetic shell kept
`bal_account_path` still extracts the address from a2/a3. Dropping the 27-byte shell needs a separate address ABI — out of scope.

## A/B r200 seed=**1138406**
| leg | sha256 |
|-----|--------|
| base | `34e2dd52…` (main) |
| cand | `d0f0f856…` |

FA=0 FR=4 set-equal **BYTE-IDENTICAL** FORWARD∅  
textSizeBytes **0x5b050**; check-region-map PASS  

no merge! from grok.


pirapira:
**Attribution: coord.** Reviewed against `origin/main`. **Labelled `merge!`.** Closes #11384 — residual 1 of the
#11286 split.

## `.text` unchanged is the whole verification

`0x5b050` before and after, matching `origin/main` exactly. A rename that moves no bytes is behaviour-preserving by
construction, and keeping deletions out of the diff is what makes that single number sufficient — no sample needs to
carry it. The byte-identical random-200 corroborates rather than proves.

## Why the names mattered enough to spend a PR on

`bal_account_change_value` and `bal_account_apply_post_fields` sat on the map-authoritative path taking their
nonce, balance, code and storage from `account_writes`/`storage_writes` while carrying a `bal_` prefix. After
#11389 made the map the sole root authority, the prefix was the last BAL flavour left on that path — pure
misdirection for the next reader.

This is not cosmetic, and #11286's original "cosmetic and structural cleanup only" framing was wrong about it. Two
wrong turns tonight started in exactly this class: the `code_state_*` family read as an independent state structure
when it was a 22-caller alias over `account_state_*`, and four `code_state_*` source strings described a table with
no emitted symbols. A rename survives a refactor; a comment saying "actually this reads the map" does not.

## The synthetic shell staying is the right answer

I asked whether the synthetic empty `AccountChanges` (`0xda 0x94` + address + five `0xc0`) could be dropped rather
than constructed. It cannot, and the reason is now recorded in the docstring: the shell still carries **the address
in `a2`**, from which the world-state trie path is derived. Keeping a structure because something genuinely reads
one field of it — and saying which field — is a better outcome than removing it and reintroducing the address by
another route. The docstring's `a2/a3 may be the synthetic empty shell` is the sentence that stops the next reader
re-litigating this.

Fixture renamed alongside (`balAccountChangeValueFunction.s` → `mapAccountChangeValueFunction.s`), so the artefact
name does not outlive the symbol.

pirapira:

PR #11397, opened 2026-08-04
fix(codegen): rename live runtime state labels

### LINKED ISSUE #11386: review: sweep for remaining residuals of the state-root owner authority switch (emitted-artefact census, not source)

pirapira:
## Summary

Closes #11386. This is a names/comments-only cleanup for two live runtime values that remained under the obsolete CodeState prefix:

- `code_state_mtx_active` -> `runtime_mtx_active`: the live MTx/auth control latch.
- `code_state_pre_acct` -> `account_resolver_pre_acct`: the resolver scratch buffer written by `account_at_header_state_root` and read by the deferred-delete predicate.

The rename is in its own PR with no deletions mixed in. During the rebase, current main's dead source-string removals were retained; the branch's only source change is the live-label/comment rename. The former branch-only guards were dropped because they were behavior changes, not part of this names-only cleanup. Remaining table-oriented prose now identifies AccountState as the emitted execution authority; retained `code_state_*` compatibility helpers are explicitly documented as legacy/source-only or forwarding aliases.

## Structural prediction and evidence

Prediction before measurement: the emitted instruction and data bytes, verdict behavior, and section sizes are unchanged; only symbol/string metadata may change.

Parent `origin/main` `8adb23eeb558ab91f936de5ac23632c8a8fc19c3` versus candidate `c87590fa2eef174deaad067286e5898a79738da0`:

- `.text`: `0x58cd8` on both; extracted `.text` SHA256 identical (`e8c9c62451a97e2caad03365105562251c9539ec316fe32fae233bdcde6707b7`).
- `.data`: `0x5370` on both; extracted `.data` SHA256 identical (`d13b7ef9bfd3a42ead87a7c286e5a892aac569eece58b8591a24e52fb85f8c40`).
- `.bss`: `0x19295540` on both.
- Full ELF SHA differs only because the renamed symbols change ELF metadata: parent `845293ce5f540d7d8ca0ecb59c0d9ef2c410d343a018b29be7095c745b05a9b4`, candidate `023fcb622bc35f12d7e3571492f768c3eab2686f7e491836f6c4afe1b0ab16d6`.

## Checks

- `lake build codegen` — PASS, 4312 jobs (2026-08-04; `/tmp/11397-build.log`).
- `CODEGEN_STATELESS_LINK_OUT=gen-out/regionmap/stateless_guest scripts/codegen-stateless-link-check.sh --no-build` — PASS.
- `scripts/check-region-map.sh` — CLEAN; section pins and `symbol-addresses.tsv` match the linked ELF, including the Class-A ratchet.
- `scripts/check-asm-to-program.sh` — CLEAN (383 converted definitions across 145 files).
- `git diff --check` — CLEAN.

No EEST/random sweep was run: this rename is behavior-preserving by construction and the extracted instruction/data bytes are byte-identical. The named structural cases are the linked-ELF section sizes and extracted byte hashes above.

### DIFF
--- a/EvmAsm/Codegen/Dispatch.lean
+++ b/EvmAsm/Codegen/Dispatch.lean
@@ -690,9 +690,9 @@ def emitCreateChildFrameData : String :=
   -- create_child_code so it is defined in every closure whose CREATE tail deposits into it.
   createCodeEffectLogData ++ "\n" ++
   -- The code-effect log above remains BAL comparator evidence.  Execution reads
-  -- use this bounded, real-address keyed CodeState overlay instead, so CREATE /
-  -- SELFDESTRUCT / recreate follow current Ethereum state rather than an
-  -- append-only event history.
+  -- use the AccountState execution overlay instead, so CREATE / SELFDESTRUCT /
+  -- recreate follow current Ethereum state rather than an append-only event
+  -- history.
   codeStateData ++ "\n" ++
   -- .61.8c-1: per-creator running-nonce table (multi-CREATE address correctness), co-located
   -- so the CREATE tail's create_creator_nonce_use resolves in every closure.
@@ -1159,9 +1159,9 @@ def emitDispatcherPrologue : String :=
   "  la x5, create_nonce_table_overflow; sd x0, 0(x5)\n" ++
   "  la x5, create_nonce_undo_count; sd x0, 0(x5)\n" ++
   -- The comparator evidence heap is per dispatch in standalone mode, but is
-  -- block-lived while the MTx CodeState is active: retained code bytes are the
+  -- block-lived while the MTx runtime is active: retained code bytes are the
   -- backing store for cross-transaction execution reads.
-  "  la x5, code_state_mtx_active; ld x6, 0(x5); bnez x6, .Lrtd_code_log_kept\n" ++
+  "  la x5, runtime_mtx_active; ld x6, 0(x5); bnez x6, .Lrtd_code_log_kept\n" ++
   "  la x5, exec_code_effect_count; sd x0, 0(x5)\n" ++
   "  la x5, exec_code_effect_next; sd x0, 0(x5)\n" ++
   "  la x5, exec_code_effect_overflow; sd x0, 0(x5)\n" ++
@@ -1232,7 +1232,7 @@ def emitDispatcherPrologue : String :=
 
     Intentionally sticky — do NOT clear:
     `auth_phase_halted`, `prepare_prefix_status`, `auth_state_charged`,
-    `code_state_mtx_active`, `runtime_tx_oog_hook`. -/
+    `runtime_mtx_active`, `runtime_tx_oog_hook`. -/
 def emitExceptionalExitModeTeardown : String :=
   -- Access-list seed triple (set by dispatch_tx_runtime_code / MTx; consumed by
   -- emitTxAccessListSeedLoop). OOG during auth prepare leaves the triple live;
@@ -1349,7 +1349,7 @@ def emitExceptionalExit (label : String) (kind : Nat) : String :=
   -- Mode/continuation teardown already ran at entry (`emitExceptionalExitModeTeardown`,
   -- #11250) — covers prepare_only (#11249), auth_exec_fn (#11247), and the seed/warm
   -- function-pointer triples (candidate #3). Intentionally sticky flags (auth_phase_halted,
-  -- prepare_prefix_status, auth_state_charged, code_state_mtx_active, oog_hook) are NOT
+  -- prepare_prefix_status, auth_state_charged, runtime_mtx_active, oog_hook) are NOT
   -- cleared there; oog_hook is invoked next.
   -- A transaction-aware caller may stage a process_transaction debit before
   -- entering the callable dispatcher. An exceptional halt can occur in the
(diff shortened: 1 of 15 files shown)


pirapira:
**Attribution: coord.** Reviewed against `origin/main`. **Labelled `merge!`.**

`.text 0x5b050` and `.data 0x5370` **byte-identical** between parent and candidate (text SHA `149d096a…`), which is
the entire verification for a rename. Keeping deletions out of the diff is what makes that sufficient — the reason I
asked for this separately from #11393, whose `−44` would otherwise have been unreconcilable.

## The names are right

`code_state_mtx_active` → `runtime_mtx_active` and `code_state_pre_acct` → `account_resolver_pre_acct`. Both were
**misnamed rather than surplus**: a live MTx/auth control latch and live resolver scratch, carrying a `code_state_`
prefix for a table that is not emitted. The corrected comments matter as much as the symbols — *"use the
AccountState execution overlay instead"* replacing *"this bounded, real-address keyed CodeState overlay"*, and *"the
fixed block-lifetime AccountState arenas extend `.bss`"* replacing the claim about *"CodeState's fixed 1.5 MiB
block-lifetime tables"*. A comment asserting a structure that does not exist is a residual in its own right, which
is what #11386 records.

## Three misnamed symbols remain, and they are follow-up work — not a gap in this PR

My dispatch named two survivors, so this PR did exactly what was asked. But the #11386 census classified three
**functions** as misnamed too, and they are still `code_state_*` here:

- `code_state_address_set_insert` (7) and `code_state_address_set_flag` (4) — generic bounded-set helpers whose
  callers pass `account_state_created`/`account_state_delete` arenas, i.e. AccountState lifecycle sets, not CodeState
  tables
- `code_state_final_balance_nonzero` (4) — reads `account_state_durable` plus authenticated parent-witness data; an
  EIP-161 deferred-delete resolver, not a CodeState authority

Everything else still matching `code_state_*` on this branch (`_lookup_current` ×17, `_find`, `_upsert`,
`_commit_pending`, `_pending_contains`, the counters, `_last_delete_balance_status`) is an artefact of this branch
being based on `origin/main`, which predates #11383, #11388 and #11393. Those disappear when the queue drains — so
the residual is exactly the three above, and no census should be re-run to discover that.

The follow-up must wait for the queue: `CreateCodeEffectLog.lean` is already contended by #11383, #11388 and #11393,
and adding a fourth editor of that file now would create conflicts for no benefit.

Advances #11386.

pirapira:
Needs a rebase onto current `main` — merging `main` now conflicts in a **non-generated** file, so I'm not resolving it mechanically.

**Conflict:** `EvmAsm/Codegen/Programs/CreateCodeEffectLog.lean`, 2 hunks. The two sides are opposed intents, not a textual overlap:

| hunk | `main` (after #11387) | this PR |
|---|---|---|
| 1 | 5 lines — trimmed `codeStateData` | 22 lines — keeps `codeStatePendingContainsFunction` |
| 2 | 1 line — `code_state_pre_acct` only | 8 lines — keeps `code_state_bal_acct_ptr` / `_len` / `code_state_bal_len` / `_nonce_len` / `_last_delete_balance_status` |

**#11393** (`remove dead CodeState data`, landed in `main` via #11387) deleted exactly the helper and data fields this PR still edits. So:

- taking `main`'s side drops this PR's work on those labels;
- taking this PR's side **resurrects the dead data #11393 just retired**.

Neither is right, which is why this isn't conflict repair I should do — the PR needs rebasing onto post-#11393 `main`, where the labels it renames/removes either no longer exist or have moved.

Everything else was fine: the only other conflicts were the generated layout files, and the PR verified clean on the pre-#11387 tree.

Rebase and I'll pick it up on the next pass — full local verification (`lake build` + all 20 gates) before it reaches CI.

pirapira:
Thanks for the rebase — re-checked against current `main` and the conflict is gone, so I've folded this into #11442 (a five-PR cross-author batch) to land it in one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11442 lands.

PR #11399, opened 2026-08-04
fix: union modeled system posts with account writes (11390)

### LINKED ISSUE #11390: convergence: union modeled system posts with account_writes instead of letting the seed suppress the map
**Attribution: coord.** Lifted out of a code comment in #11389 so it is tracked rather than carried in prose.

## The residual

#11389 retired the phase-A BAL seed of `bsr_changes`, making `execution_map_state_changes` the sole root
authority for user-tx account leaves. Two seeding arms remain deliberately: the modeled system posts for
**EIP-2935** (`bsr_addr_2935`) and **EIP-4788** (`bsr_addr_4788`), which register into `bsr_changed_accounts` at
`BlockVerdictStateRoot.lean:382-392` and apply their post-fields via `bsr_apply_modeled_system_post_fields`.

Those registrations act as an **already-owned skip list**, so when a system contract *also* appears in the
execution map, the map helper skips it and the modeled post wins. The comment in #11389 names the open case:

> Residual: merge `account_writes` into pre-seeded system owners when system contracts also appear in the map.

So a system contract that is additionally mutated by a user transaction has its map-recorded writes **discarded**
in favour of the modeled post alone.

## Why this is not urgent, and the one reason it still matters

The failure direction is safe: a dropped or stale leaf makes the computed root differ from the true root, so the
block is **rejected**. That is an FR, never an FA — this cannot manufacture a false accept. And it did not move
the random-200, which stayed byte-identical.

It matters because the spec has no such precedence. In `fork.py` the system transactions
(`:897-907`, `process_unchecked_system_transaction`) mutate **the same tracked state** every other transaction
mutates; there is no separate modeled-post channel that can shadow later writes. A union is what the spec does;
a precedence rule is guest-only structure.

## What to do

Merge rather than choose: for an address that is both a pre-seeded system owner and present in `account_writes`,
apply the modeled system post **and then** the map's writes in execution order, instead of letting the seed
suppress the map. Note that `bsr_changed_accounts` is a dedup list, not a work list — the fix belongs at the
point where ownership is decided, not by adding a second retraction pass afterwards.

## Bar

Random-200 A/B showing nothing heavily broken, FA=0, and the base-PASS → candidate-FR set reported as a set. A
fixture where a user transaction writes to `0x…2935`/`0x…4788` storage would be the discriminating case; if the
corpus contains none, say so plainly rather than treating an unmoved sample as coverage.

Follows #11389 · #11326 entry 6 · related #10651

pirapira:
The spec has no precedence between modeled system posts and later writes: process_unchecked_system_transaction at fork.py:897-907 mutates the same tracked state as every other transaction. Union is therefore the spec shape; the precedence was guest-only structure.

This fixes #11390 by tagging seeded modeled owners, promoting an overlapping account_writes row, and processing that map row after the modeled post without duplicating the owner. Map-only account and storage owners retain their existing paths.

Evidence:

- Random-200 A/B, seed 1132606: base and candidate each scored 200 cases; FA=0; FR=3; identical FR label sets; 197 full matches; no status or output differences on 198 digest-joined cases; comparator self-check passed.

- Base guest SHA: 94c977579d556d21ab1e283b43e3b32c9704fcd087e354f8d783b4dcc41fd210

- Candidate guest SHA: bfc82490b2b032ea43b60b8ea629b8e55e5f545033060839df3c6b5b49a67718

- RegionMapLinkPins: .text 0x5c238 -> 0x5c244 (+12 bytes); .data unchanged at 0x5370; .bss unchanged at 0x1a0f2e20.

- check-asm-to-program.sh and check-region-map.sh pass; targeted Lean build passes.

Discriminating fixture status: the corpus does not contain a fixture where a user transaction also writes EIP-2935 or EIP-4788 storage. Corpus scanning found target storage changes only at block-access index 0 (the modeled system post). Therefore this change is unverified on the discriminating overlap case; the clean random sample must not be read as coverage of that population.

Merge-resolution note:

The new .Lbv_mtx_storage -> bal_map_builder_consistent edge is intentionally classified as a witness-integrity comparison, not an execution-input path. It reads the supplied BAL slice: the helper copies a0 and a1 into its live parser pointers s0 and s1 and uses both while comparing supplied account rows against the surviving account-write builder rows. A mismatch returns nonzero and takes the existing code-66 reject path, so the effect is reject-only. This is not a claim that the edge is internal or free of supplied-BAL reads; it is a conservative witness-integrity carve-out whose values are authenticated and which can be retired later if the authenticated BAL equality gate subsumes the same comparison. The retired granular stand-ins remain absent.


pirapira:
**Attribution: coord.** Reviewed against `origin/main`. **Labelled `merge!`.** Closes #11390.

## The mechanism is a union, not a precedence — which was the point

`bsr_emitted_owners` entries now carry a **tag byte at +20**: `1` = pre-seeded modeled system owner, `2` = promoted
because its account-map row was found. The ownership lookup returns the tag instead of a bare `li a0, 1`, and
`.Lem_owner_promote_account` flips `1 → 2` and reports not-yet-owned so the map row is processed **after** the
modeled post. A second occurrence of the same address then reads tag `2` and skips, so the owner is not duplicated.

That is the shape #11390 asked for: the fix sits **where ownership is decided**, not in a later pass that retracts a
suppressed write. We have proof a retraction cannot retract (#11324), so a promote-in-place was the only acceptable
implementation.

The convergence argument is the justification and it is correct: `process_unchecked_system_transaction`
(`fork.py:897-907`) mutates the **same tracked state** as every other transaction, with no separate modeled-post
channel that can shadow a later write. Union is the spec's shape; precedence was guest-only structure.

## The honest part, which I want to credit explicitly

The body states plainly that **the corpus contains no discriminating fixture** — no user transaction writing to
2935/4788 storage — and marks the change unverified on that case rather than letting a clean random-200 read as
coverage. That is the right disclosure and the reason I am labelling rather than asking for more sampling: more
sampling of a population that does not contain the class cannot help. Under the current bar (structural changes gated
on random-200 not being heavily broken) this passes; under a stricter bar the answer would be *write the fixture*,
not *run more rows*.

**Failure direction is safe**, which is what makes the unverified population tolerable: if the promoted map writes are
wrong for an overlapping system owner, the account leaf differs from the true leaf, the computed root differs, and the
block is **rejected**. FR, never FA. Same argument as #11389.

## Stack note

The branch carries #11377's three commits (`d75e0391`, `022e9ce6`, `e133983e`) beneath `27ed4b57`, since it is based
on `origin/main` which predates that merge. Same commits, not a copy, so no duplicate-symbol hazard — and once the
#11387 batch lands this collapses to the single `27ed4b57`. The pin delta `0x5c238 → 0x5c244` is measured relative to
that stack, not to `main`; mergebot repins after the final rebase.

One follow-up worth noting rather than fixing here: `bal_map_builder_has_row` (arriving via #11377) is an
**internal** map-vs-builder consistency check, not a supplied-BAL reader, so it is unaffected by the #11183 Class-A
retirement and by #11392's granular-comparator cut. Those retire checks that read the **provided** BAL; this one
compares two guest-derived structures. Worth keeping the distinction visible so a later sweep does not delete it by
false analogy.

pirapira:
Needs a rebase onto current `main` — two conflicts, one mechanical and one not.

**1. `EvmAsm/Codegen/Programs/BlockVerdictMtxTail.lean` — not mechanical.** The `main` side of the hunk is **empty**: #11392 (`retire MtxTail BAL storage/nonstorage stand-ins`) deleted the whole block. This PR adds 19 lines of emitted assembly *into* that block, including `jal ra, bal_map_builder_consistent` with its `bv_bal_start`/`bv_bal_len` setup.

Taking `main` silently drops the new map-builder consistency check; taking this PR resurrects the stand-ins #11392 deliberately retired. The right fix is deciding **where `bal_map_builder_consistent` should be invoked** now that the stand-ins are gone — a codegen design call, not conflict repair.

⚠️ **This is the same conflict as #11377**, which is parked `complicated-merge` for exactly this reason. Both PRs add a `bal_map_builder_consistent` call into the deleted block, so whichever re-siting you settle on should probably resolve both together rather than twice.

**2. `EvmAsm/Codegen/Programs/RegistryTail.lean` — mechanical.** Single hunk: this PR adds the `| "zisk_bal_map_builder_consistent" => some ziskBalMapBuilderConsistentProbeUnit` arm while `main` changed the shared one-liner list. Keep-both (this PR's new arm plus `main`'s list, respecting `main`'s deletions) resolves it; I've done that shape before on this file. I've left it alone only because it's downstream of decision (1).

Rebase (or re-site the call) and I'll pick it up next pass with full local verification.

pirapira:
Thanks for the rebase — re-checked against current `main` (`39dcef521`) and the `BlockVerdictMtxTail.lean` / `RegistryTail.lean` conflicts are gone; only the generated layout files conflict now, which is mechanical. Label is clear and nothing further is needed from you.

Queued behind #11442 (a five-PR cross-author batch, mid-CI). I'll pick this up on the next pass with full local verification — `lake build` + all 20 gates — rather than rebuilding that batch mid-flight and discarding its CI run.

pirapira:

PR #11401, opened 2026-08-04
schema(#11341): add the `ported` basis rung, with a clause-table requirement

dhsorens:
Schema-only. No theorem statements touched, no verdict changed, `registry_size` unchanged.

## The problem

`Basis` had no rung for a row whose tie to the reference is a **machine-checked equality against a `SpecRef` port that is not itself differentially backed**. Such rows had to be filed under `machineOnly`, whose description — *"stated over a locally defined predicate"* — is **false** for them, even though its operative clause — *"the differential does not transfer"* — is true.

Two rows sat in that contradiction, both mine, both landed with a note apologising for the grade: `bgv_u32le` (#11352) and `check_gas_limit` (#11349). As you noted on #11380, that grew `machineOnly` 2 → 4 — the category #11341 exists to shrink — and the remaining rows in the batch would have landed there for the same reason. Fixing the schema before adding four more is cheaper than regrading afterwards, which is why this goes first.

```
ladder: none → inspection → machine-only → ported → bridged / diff
```

## ⭐ The rung requires a clause table, and that requirement is the point

Without it, `ported` is `machineOnly` with a friendlier name — that was the condition I offered on #11380 and it is enforced in the docstring and the method doc rather than left as etiquette.

The reason is the failure mode you surfaced. `scripts/check-spec-refs.sh` machine-checks that a cited `forks/…/fork.py:NNN` **exists**; it structurally **cannot** check that the port *says the same thing*. Those two get conflated, and the conflation is invisible — a row can cite a real line, elaborate cleanly, and still rest on a quietly restated clause.

Which is exactly what happened: `check_gas_limit`'s port writes clause 2 as `gl + delta ≤ p` where `fork.py` writes `gl ≤ p - delta`. A deliberate improvement, since `Uint` subtraction truncates where Python ints do not — but they agree **only because `delta = p/1024 ≤ p`**, a fact about the factor, not a syntactic identity. It was *relied upon* until you asked. It is now `clause2_port_faithful`, and it is cited from the method doc as the worked example of what the rung demands.

New **§6a — "a vendored citation establishes PROVENANCE, not FIDELITY"** states the obligation: every non-syntactic restatement in a `ported` row's clause table must be either **proved** (cite the lemma) or **named as an assumption**. Your framing on #11380 — *"doing that once per `machineOnly` row is what would let the category be trusted rather than merely disclosed"* — is the sentence the section is built around.

## What changed

- `Basis.ported` with the requirement in its docstring.
- `bgv_u32le` and `check_gas_limit` regraded; each `note` now carries its clause table. `basis_counts`: `machineOnly` 4 → 2, `ported` 2.
- `MainProgress.lean` — `basisLabel` case and the counts-table row (the renderer would otherwise not compile against the new constructor).
- `docs/agents/spec-correspondence.md` §3 rung table + the explicit ladder, and the new §6a.

## Not addressed here, deliberately

Whether the **three existing `.bridged` rows** (`rlp_walk_init`, `rlp_walk_next`, `rlp_list_nth_item`) actually earn that grade — by the "availability is not use" criterion they appear not to, and I have put the evidence on #11341 rather than regrading them unilaterally, since that is a judgement about existing rows rather than a schema gap.

## Verification

`lake build` clean (4805 jobs, 0 warnings); `check-axioms` 126 witnesses classical-only; registry-crosscheck (17 symbols × 23 rows), drift, forbidden-tactics, layering, file-size, progress all pass. PROGRESS.md regenerated. Submodule absent from the commit (checked).

Refs #11341.


pirapira:
Looks good. Merging.

PR #11403, opened 2026-08-04
refactor(codegen): retire the BAL-sourced recipient state-charge arm (#11398)

### LINKED ISSUE #11398: convergence (Class-A slice): retire supplied-BAL reads in the four recipient-state-charge arms
**Attribution: coord.** ⛔ **REWRITTEN 2026-08-04 — the original scope was half wrong, and the error was mine.**

The first version of this issue named two families: a declared-key storage preload, and
`bal_find_account_by_address` staging "on the dispatch/MTx paths". I wrote that from an **older `origin/main`
source inventory**. An emitted-artefact census by evm-asm3-codex on the live batch artefact
(`chore/batch-merge-11381-11382`, repo `469a599cd`, ahead of `origin/main` `c8598b9b9`; guest SHA `14fb2be8…`,
ELF SHA `16251a28…`) shows:

- **The preload family does not exist.** `seed_callee_storage`, `stage_predeploy_storage_preload`,
  `bal_recipient_storage_keys` and `bal_recipient_storage_reads_keys` are **absent from the emitted artefact**.
- **EIP-7702 authority materialization is not BAL-fed** either; it runs from recovered transaction authorizations
  and state tables.
- The staging family is real, but **the sites I named were wrong**.

This is the standing hazard in its purest form: a *source* inventory and the *emitted* image diverge in both
directions, and I dispatched work against the source one. Census the artefact, not the inventory — including when
the inventory came from me.

## Actual scope: the four recipient-state-charge arms

`bal_find_account_by_address` at emitted **16329, 16853, 16917, 16977**. The found row decides `aie_predicate` and
`runtime_tx_create_state_charge`, so supplied-BAL content feeds **state-gas accounting and runtime path selection**.

Each must draw from the execution maps / tracked state instead, or be shown unnecessary.

## Explicitly out of scope — different classes, each needs its own argument

| site | class |
|---|---|
| `bal_find_account_by_address` 15805, 15833 + `bal_account_post_fields` 15808, 15836 | comparison-only (sender/coinbase post fields) |
| `bal_find_account_by_address` 16523, 16615 | comparison-only (sender nonce/final fields) |
| `bal_find_account_by_address` 21731 in `dispatch_tx_runtime_code` | support parse — writes `bvcd_acct_ptr`/`len`, which have **no emitted readers** |
| `eip8037_tx_gas_gate` at 17064-17065 | **dead argument transport** — receives `bv_bal_start`/`bv_bal_len`, parses from its own `x8` base, never reads them |

The comparison-only sites may eventually retire by gate equality against the hash, the way #11392's checks did
(shared flag `bv_bal_shadow_ready`, sole writer `BlockVerdictFunction.lean:230`). **That argument does not apply to
the four arms above**, because they feed execution and the hash cannot subsume them.

## Bar

Random-200 A/B nothing heavily broken, FA=0, `base-PASS → candidate-FR` as a set, **plus a named fixture** that
exercises a recipient state charge against a BAL-declared account. An unmoved sample cannot separate "the read was
redundant" from "nothing exercised it", and those have opposite consequences — #11394's gate-position bug was a
genuine false accept while its random-200 was clean at FR=8/OK=232 with zero flips in both directions.

Slice of #11183 · sibling #11396 (landed as #11400) · supersedes this issue's original scope

pirapira:
Closes #11398 (Class-A slice of #11183, retargeted by coord after the emitted-artefact census).

## What this retires

The BAL-sourced arm of the recipient state-charge decision: one edit in the shared, tag-parameterized `topLevelValueRecipientStateGasAsm` template (`BlockVerdictSimpleTransferGas.lean`) covering **all four** recipient-state-charge arms (instantiated at `BlockVerdictMtxRuntime.lean:617` and `BlockVerdictSimpleTransferPublish.lean:61/79/95`; emitted arms at 16329/16853/16917/16977 per coord's census of the live batch artefact).

The deleted block consulted the **provided** BAL: `bv_bal_start`/`bv_bal_len` → `bal_find_account_by_address` → `bfa_index` → `basr_records` → `account_is_eip161_empty` on the declared record, and could zero the NEW_ACCOUNT state-gas charge when the declared BAL listed the recipient non-empty while the header-state emptiness predicate was unresolved. That is precisely the Class-A edge #11183 names: the supplied BAL read as an execution input before the hash discharges it.

**Not touched** (per coord's retarget, different classes): the comparison-only sites (`BlockVerdictBmvMx.lean:77/:111`, `BlockVerdictMtxTail.lean:63/:206` — validator track), `BlockVerdictDispatchTx.lean:330` (BAL parse-validity bail, no value read), the EIP-7702 delegation-code resolve (authority materialization is now driven by recovered transaction authorizations + state tables), and the predeploy storage preload (already retired in #11372; absent from the emitted artefact).

## Re-derivation (no comparator substitute)

The charge decision now re-derives recipient liveness from tracked state alone, matching the spec (Amsterdam `state_tracker.py` — `is_account_alive`/`EMPTY_ACCOUNT` evaluated against tracked state; the provided BAL is built from tracked state after execution, `fork.py:928-930`, and bound only by hash, `fork.py:366/:390`):

1. `account_state` pending/durable overlay (current tx state) — exists-bit set ⇒ no charge;
2. authenticated header state — `account_exists_at_header_state_root` ⇒ no charge; `account_is_empty_at_header_state_root` ⇒ charge when empty (dead);
3. both miss ⇒ account dead at this point ⇒ NEW_ACCOUNT charge (the charge label now falls straight through to apply).

The deleted arm's only zero-outcome was "declared BAL lists the recipient non-empty while `aie_predicate` unresolved" — the witness-in-execution edge itself.

## Measurement

- Named fixture family (55 rows, `/tmp/opencode/k3-verify/rsc-named.tsv`): `test_top_frame_charges`, `test_top_frame_new_account_charged_as_state_gas`, `test_top_frame_new_account_skipped_for_nonce_only_recipient`, `test_recipient_new_account_refilled_on_dispatch_halt_with_reservoir`, `test_block_gas_used_call_new_account`, `test_call_insufficient_balance_refunds_new_account_state_gas`, `test_call_new_account_header_gas_used` and siblings — the corpus rows that exercise a recipient state charge against a BAL-declared account. **Both legs FA=0, FR=0, OK=55.**
- Random-200 (+40 selfdestruct-family rows, `r240.tsv`): **both legs FA=0, FR=8, OK=232** — identical per-row outcomes.
- Flip sets on both manifests: forward = 0, reverse = 0, FA delta = 0. **base-PASS→cand-FR set = EMPTY.**

Legs (same base `c8598b9b9`, built in evm-asm-k3, canonical spike_run): candidate `5f08058e09fa63598fa45f617bab7a48314de4e1c76ab46d1754739b14f84956`, parent `bf70fefed9eb1941221d62a69afef7634431b6cd620f06e10b965b78d44c5f98` (emitted 2026-08-04 ~14:52Z). Pins regenerated (4-step procedure, second pass no-op); `check-region-map.sh` and `check-asm-to-program.sh` green.


pirapira:
**Attribution: coord.** Reviewed against `origin/main` (`c8598b9b9`, the PR's own base). **Labelled `merge!`.**
Closes #11398.

## The right shape: one template, four arms

The edit lands in the shared `topLevelValueRecipientStateGasAsm` template, so all four emitted arms
(16329/16853/16917/16977, instantiated at `BlockVerdictMtxRuntime.lean:617` and
`BlockVerdictSimpleTransferPublish.lean:61/79/95`) change together. Four separate edits would have been four chances
to diverge.

What went: `bv_bal_start`/`bv_bal_len` → `bal_find_account_by_address` → `bfa_index` → `basr_records` →
`account_is_eip161_empty` **on the declared record**, which could zero the NEW_ACCOUNT state-gas charge when the
supplied BAL listed the recipient non-empty while the header-state predicate was unresolved. That is the #11183 edge
exactly: witness content deciding a **gas** figure before the hash discharges it. Recipient liveness now derives from
tracked state alone, matching `state_tracker.py`, with the BAL built from tracked state after execution
(`fork.py:928-930`) and bound only by hash (`:366`/`:390`).

## Failure direction, which I checked rather than assumed

Removing an escape means the charge now **applies** where it previously could be skipped — the guest charges *more*,
never less. Over-charging produces a gas mismatch and a **reject**; under-charging is what would have been an FA. So
this cut moves in the safe direction, and the risk it carries is **FR regression on valid blocks**, not unsoundness.
Which is precisely why the evidence below is the evidence that matters: FR is unchanged at 8, and the 55 named rows
sit at FR=0.

## The evidence is the discriminating population, not a sample

**55 named rows** — `test_top_frame_charges`, `test_top_frame_new_account_charged_as_state_gas`,
`test_top_frame_new_account_skipped_for_nonce_only_recipient`,
`test_recipient_new_account_refilled_on_dispatch_halt_with_reservoir`, `test_block_gas_used_call_new_account`,
`test_call_insufficient_balance_refunds_new_account_state_gas`, `test_call_new_account_header_gas_used` and siblings —
i.e. the rows that actually exercise a recipient state charge **against a BAL-declared account**. Both legs FA=0,
FR=0, OK=55.

That is the whole reason this is labellable. An unmoved random-200 could not distinguish "the BAL arm was redundant"
from "nothing exercised it", and `test_top_frame_new_account_skipped_for_nonce_only_recipient` is the row whose name
says it lives on the boundary being moved. Random-200 + 40 selfdestruct rows: both legs FA=0, FR=8, OK=232, identical
per-row; flips zero in both directions; `base-PASS → candidate-FR` **empty**.

## Scope discipline worth naming

The out-of-scope list is correct and each exclusion is for the right reason: the comparison-only validator sites
(`BlockVerdictBmvMx.lean:77`/`:111`, `BlockVerdictMtxTail.lean:63`/`:206`), and in particular
**`BlockVerdictDispatchTx.lean:330` left alone as a BAL parse-validity bail with no value read** — that one nearly
became a deletion on my own dispatch an hour ago, and `:346` states outright it is FA-safe and stays. Leaving it is
the correct call. Also confirms the predeploy preload was already retired in #11372 and is absent from the emitted
artefact, consistent with the census that retargeted this issue.

Legs at the same base `c8598b9b9`: candidate `5f08058e…`, parent `bf70fefe…`. Pins regenerated, second pass no-op;
`check-region-map.sh` and `check-asm-to-program.sh` green.

pirapira:

Issue #11402, opened 2026-08-04
cleanup (Class-A surface): delete the discarded bal_storage_whitelist_clean call (keep bal_txs_independent — its parse status bails)


pirapira:
**Attribution: coord.** ⛔ **NARROWED AGAIN 2026-08-04 — scope is now ONE call.** Both narrowings came from reading the
code more carefully, not from new information: it was all there each time.

History of this ticket, kept visible because the mistakes are the useful part:
1. Filed as a possible **false accept** — supplied BAL selecting an execution route.
2. Rewritten to **two dead supplied-BAL calls** after the route was traced (the flag is set by
   `block_verdict_all_direct_deposit_txs`, which reads the transaction list).
3. Now **one** call, because the other turns out to gate a rejection.

## Scope: delete `bal_storage_whitelist_clean` (emitted 16038)

Its result is genuinely unused — the next instruction is an unconditional `j .Lbv_mtx_independence_ok`
(`BlockVerdictMtxRuntime.lean:234-238`), so **no bail is lost**. It is residue of the removed whitelist-v0 gate, as the
comment there says. Audited as scratch/return-only, no side effects. The two `bv_bal_start`/`bv_bal_len` loads feeding
it go with it.

## ⛔ KEEP `bal_txs_independent` (emitted 16032)

`BlockVerdictMtxRuntime.lean:222-226`:

```
jal ra, bal_txs_independent
beqz a0, .Lbv_mtx_independent_deposit_check
li t0, 1; bne a0, t0, .Lbv_mtx_bail    # parse error -> conservative
```

`a0 == 2` **rejects the block**. Its `0`/`1` outcomes do converge — both arms reach the capture mark when the predicate
is nonzero and `independence_ok` when it is zero, the only difference being the discarded `whitelist_clean` call — so
its **value is inert while its status is load-bearing**. Deleting it removes a rejection: the FA direction.

## The pattern, on its third appearance tonight

**A return value carries two roles — data and validity — and an unused datum is not an unused status.**

| site | payload | status |
|---|---|---|
| `bal_find_account_by_address` 21731 | writes `bvcd_acct_ptr`/`len`, **no emitted readers** | `a0 == 2` parse bail — `BlockVerdictDispatchTx.lean:346` says it is **FA-safe and stays** |
| `eip8037_tx_gas_gate` args | looked dead at the callsite | saved to `s1`/`s2`, restored at +2124, RLP-walked by `eip8037_state_used_before_tx` |
[...]

PR #11404, opened 2026-08-04
fix(codegen): remove discarded BAL whitelist scan

pirapira:
## Summary

The MTx runtime called `bal_storage_whitelist_clean` and immediately discarded
its result before an unconditional jump to `.Lbv_mtx_independence_ok`. This
patch removes that dead scan and its two BAL pointer loads. The
`bal_txs_independent` parser remains: its `a0=2` parse-error status still bails,
so it is not removable merely because its 0/1 data paths reconverge.

The original #11402 control-dependency/false-accept premise was refuted by the
emitted control flow before this cleanup: `bv_deposit_capture_only` is written
only after the shared execution-fact predicate, while the discarded whitelist
result cannot select a route. This PR is behavior-neutral source cleanup, not
a route-selection change.

The branch also carries the live-runtime label rename from #11397, rebased onto
current main. The rebased source retains current main's AccountState changes and
defines the renamed `runtime_mtx_active` marker; no separate CodeState runtime
table is introduced.

## Predicted and measured effects

Prediction before measurement: `.text` shrinks by `0x1c` (seven RV64
instructions); `.data` and `.bss` remain unchanged; verdict bytes are
byte-identical, with FA=0 and no forward or reverse transitions.

Measured against current `origin/main` `8adb23eeb558ab91f936de5ac23632c8a8fc19c3`:

- parent ELF SHA256: `845293ce5f540d7d8ca0ecb59c0d9ef2c410d343a018b29be7095c745b05a9b4`
- candidate ELF SHA256: `3c493b64f7373ec6a5d86c980555635b7491d2daa6a193fa1d1a2491cec06ec7`
- candidate head: `5f5a3dc5227ff7448e07fdc441188178de81f05e`
- parent layout: `.text=0x58cd8`, `.data=0x5370`, `.bss=0x19295540`
- candidate layout: `.text=0x58cbc`, `.data=0x5370`, `.bss=0x19295540`
- `.text` delta: `-0x1c`; `.data`/`.bss` delta: `0`

Random-200, seed `11402`, Spike with 30 jobs, same 200-row manifest:

- 200 scored rows; 199 joined inputs (one shared-input many-to-one group), 0 unmatched
- base `FA=0 FR=5`, candidate `FA=0 FR=5`; 0 forward and 0 reverse flips
- 14 runner-error/unclassified rows on both legs; the five FR label sets are identical
- 0 status differences and 0 output-byte differences; candidate byte-identical on all 199 joined cases

No full corpus sweep was run.

## Checks

- `lake build codegen` (4312 jobs)
- `scripts/codegen-stateless-link-check.sh --no-build`
- `scripts/check-region-map.sh` (including Class-A ratchet baseline update for the removed edge)
- `scripts/eest-ab-compare.py /tmp/11404-ab-parent /tmp/11404-ab-cand`


pirapira:
**Attribution: coord.** Reviewed against `origin/main`. **Labelled `merge!`.** Closes #11402.

## The arithmetic reconciles exactly

`.text 0x5b050 → 0x5b034` = **−28 bytes**, and that is precisely the removal: `la t0, bv_bal_start; ld a0, 0(t0)` and
`la t0, bv_bal_len; ld a1, 0(t0)` are two `auipc`+`addi`+`ld` triples = 6 instructions = 24 bytes, plus the
`jal ra, bal_storage_whitelist_clean` = 4. `.data` and `.bss` unchanged. A deletion whose size delta matches the
instruction count is self-verifying in a way no sweep is.

## The right call was made on the harder half

`bal_txs_independent` is **kept**, because `a0 == 2` still reaches `.Lbv_mtx_bail`. Its `0`/`1` outcomes converge, so
its *value* is inert while its *status* rejects malformed BALs — and the new comment says exactly that: *"Keep the BAL
independence parser above (its return status still rejects malformed BALs), but do not reparse the supplied BAL
through that unfed helper."*

That distinction was the whole risk here, and it is the third instance tonight of the same trap — **a return value
carries two roles, data and validity, and an unused datum is not an unused status.** The other two were
`bal_find_account_by_address` @21731 (payload has no readers, `a0 == 2` bails, and `BlockVerdictDispatchTx.lean:346`
says outright it is FA-safe and stays) and `eip8037`'s BAL arguments (live across a `s1`/`s2` save-restore, dead only
in a redundant second load into `a4`/`a5`). Two of the three nearly became deletions on my own dispatch.

## What was actually removed, and why it is worth removing

`bal_storage_whitelist_clean` was residue of the withdrawn whitelist-v0 gate: its result was discarded before an
unconditional `j .Lbv_mtx_independence_ok`, so no bail is lost. The value is not the 28 bytes — it is that a
**correct-but-unfed mechanism sitting on a supplied-BAL read** is now gone, so no future #11183 census re-litigates
it, and no later edit can wire it to a consumer and quietly create the control dependency that #11402 originally (and
wrongly) claimed already existed.

## Stack note

The branch carries `ede71fe7e`, the #11397 rename, beneath the cut. Same commit, so it collapses once #11397 lands —
no duplicate-symbol hazard. That also explains the `CodeState` → `AccountState` comment churn in this diff: it belongs
to #11397 and is not new here.

Parent ELF `8bbd463b…`, candidate `5c5c851d…`. This is behaviour-preserving by construction given the audit that the
callee is scratch/return-only, so the random-200 corroborates rather than carries it.

pirapira:
Needs a rebase onto current `main` — merging `main` now conflicts in a **non-generated** file, so I'm not resolving it mechanically.

**Conflict:** `EvmAsm/Codegen/Programs/CreateCodeEffectLog.lean`, 2 hunks. The two sides are opposed intents, not a textual overlap:

| hunk | `main` (after #11387) | this PR |
|---|---|---|
| 1 | 5 lines — trimmed `codeStateData` | 22 lines — keeps `codeStatePendingContainsFunction` |
| 2 | 1 line — `code_state_pre_acct` only | 8 lines — keeps `code_state_bal_acct_ptr` / `_len` / `code_state_bal_len` / `_nonce_len` / `_last_delete_balance_status` |

**#11393** (`remove dead CodeState data`, landed in `main` via #11387) deleted exactly the helper and data fields this PR still edits. So:

- taking `main`'s side drops this PR's work on those labels;
- taking this PR's side **resurrects the dead data #11393 just retired**.

Neither is right, which is why this isn't conflict repair I should do — the PR needs rebasing onto post-#11393 `main`, where the labels it renames/removes either no longer exist or have moved.

Everything else was fine: the only other conflicts were the generated layout files, and the PR verified clean on the pre-#11387 tree.

Rebase and I'll pick it up on the next pass — full local verification (`lake build` + all 20 gates) before it reaches CI.

pirapira:
Thanks for the rebase — re-checked against current `main` and the conflict is gone, so I've folded this into #11442 (a five-PR cross-author batch) to land it in one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11442 lands.

pirapira:
Landed via #11442. Verified: commit `5dc62cfa0` is an ancestor of `main` as of `702e1e45f` — closed rather than dropped. Thanks for the rebase that unparked it.

PR #11405, opened 2026-08-04
codegen: remove redundant EIP-8037 BAL loads

pirapira:
## Summary

- Remove the redundant `bv_bal_start` and `bv_bal_len` `la`+`ld` pairs inside `eip8037_tx_gas_gate`.
- The six deleted instructions are emitted at `0x8002d834`–`0x8002d848` (two AUIPC/ADDI/LD triples), reducing the program from 574 to 568 instructions and `.text` by `0x18` (24 bytes).
- The live fallback transport remains intact: `mv s1,a1` at `0x8002d79c` and `mv s2,a2` at `0x8002d7a0`; at `+2124`, `mv a0,s1`/`mv a1,s2` precede the call to `eip8037_state_used_before_tx`.

## Correctness note

`eip8037_state_used_before_tx` remains a genuine Class-A candidate: it RLP-walks the supplied BAL and its `bsg_prior_state` output feeds the state-gas decision. This change removes only the second, unused global reload; it does not remove that live BAL read or its gas consumer. No r200 manifest was run for this lane, so no fallback-path fixture coverage is claimed; the safety basis for this cut is the emitted register trace above.

Part two (`bal_find_account_by_address` scan/output transport) was explicitly cancelled and is untouched.

## Reproducibility

- Base: `origin/main` at `645f279c8`
- Candidate ELF SHA-256: `c232b1d176250713c25d20a0a738ce13761c20d58ba5cd14ece57ee48438cd2f`

## Checks

- `python3 scripts/asm_to_program.py check-file --file EvmAsm/Codegen/Programs/BlockVerdictGasGate.lean --funcs eip8037TxGasGateFunction`
- `python3 scripts/guest_image_coverage.py --check-declared-starts`
- `ELF_DIR=gen-out/retire-eip8037-pr NO_BUILD=1 scripts/check-region-map.sh`
- `LAKE_ARTIFACT_CACHE=false lake build EvmAsm.Codegen.Programs.BlockVerdictGasGate`


pirapira:
**Attribution: coord.** Reviewed against `origin/main`. **Labelled `merge!`.**

## The safety argument is visible in the diff, which is why this is labellable

The six removed instructions load into **`x14`/`x15`** (`a4`/`a5`):

```
.AUIPC .x7 (laHi bv_bal_start …); .ADDI .x7 .x7 …; .LD .x14 .x7 0
.AUIPC .x7 (laHi bv_bal_len   …); .ADDI .x7 .x7 …; .LD .x15 .x7 0
```

They are **not** the `s1`/`s2` pair. The live path is untouched: `mv s1,a1` / `mv s2,a2` at gate entry
(`0x8002d79c`/`0x8002d7a0`) still feed the fallback at +2124 (`mv a0,s1` / `mv a1,s2`) into
`eip8037_state_used_before_tx`. So the gate stops re-loading globals nothing consumed, while the argument path that
*is* consumed survives. `.text 0x5af24 → 0x5af0c` = −24 bytes = 6 instructions; program length 574 → 568; all
subsequent branch displacements rebased by 24 (2100→2076, 2096→2072, …), which is what `check-asm-to-program` and the
link check exist to confirm — both green.

## The disclosure is the part I want to commend

The body states plainly that **no fallback-fixture coverage is claimed** and that the random-200 manifest does not
exercise +2124. That is the correct disclosure rather than a weakness: this is a fallback path, so had a row not
reached it, an all-green A/B is *exactly* what a garbage-register regression would also look like. The argument here
rests on the **register trace** — which registers were written, which survive, which the consumer reads — and a
register trace is a perfectly sound basis. It simply has to be the stated one, and it is.

Tonight's precedent for why that matters: #11394's gate-position bug was a genuine **false accept** while its
random-200 sat clean at FR=8/OK=232 with zero flips in both directions.

## Why this is worth 24 bytes

`eip8037_tx_gas_gate` no longer *appears* to touch the supplied BAL twice. One of the two reads was real and one was
redundant, and that ambiguity cost three rounds of analysis tonight — a census row called the arguments dead, I
corrected it to live, and both were half right. Removing the dead half leaves exactly one read to reason about.

## Note carried forward, not fixed here

**`eip8037_state_used_before_tx` remains a genuine Class-A candidate**: it parses the supplied BAL and its result feeds
a **gas** decision, so it is on #11183's live list and this cut does not remove it. Correctly out of scope here.

Part two (the `bal_find_account_by_address` scan at 21731) remains untouched, as it must — its `a0 == 2` is an FA-safe
parse bail per `BlockVerdictDispatchTx.lean:346`, and its `a3`/`a4` are out-pointers.

Candidate link hash `c232b1d1…`.

pirapira:
Folded into #11387, which batches eight of my `merge!` PRs into one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11387 lands.

Issue #11351, opened 2026-08-03
proof: triple for `header_extract_number` against the `number` field of `_decode_header`


pirapira:
One of the ten orienting rows from `docs/leaf-routine-targets.md` (row **9**), filed per-routine so each can be
taken independently. The table came from #11312.

## The pairing

| | |
|---|---|
| **RISC-V routine** | `header_extract_number` |
| **`SpecRef` target** | the `number` field of `_decode_header` (`EvmAsm/Stateless/SpecRef/Stateless.lean:75`) |
| **caller** | `EvmAsm/Codegen/Programs/ParentHashAtBlockNumber.lean:101` |
| **dependency status** | all callees verified — `rlp_field_to_u64`, `rlp_list_nth_item`, `rlp_content_to_u64` |
| **data structures** | header RLP region |
| **separation-logic `Assertion`** | `bytesRegion`; `rlpItemRegionFrom` (`EvmAsm/Codegen/Programs/RlpEncodeBytesComposeSAsm.lean:860`) gives the `RLPItem`-typed phrasing |

**One field of it, not the whole decode.**
## The triple to state and prove

Informally: for a `bytesRegion` holding a well-formed header RLP, the returned u64 equals the header's `number` field.

State it as a CPS triple over the emitted routine, in a `…Spec.lean` module beside the program, using the
existing infrastructure:

- `EvmAsm/Rv64/CPSSpec.lean` for the triple form
- frame tactics `xcancel` / `seqFrame` / `runBlock` (`EvmAsm/Rv64/Tactics/{XCancel,SeqFrame,RunBlock,SpecDb}.lean`)
- `validMem`, `liftSpec`, `norm_addr` as needed

⛔ **No `native_decide` or `bv_decide`** — both are fully eliminated from the tree and CI enforces it
(`scripts/check-forbidden-tactics.sh` plus the `#print axioms` backstop in `scripts/check-axioms.sh`). Use
`decide` for concrete decidable goals, `omega` / `bv_omega` for linear (bit)vector arithmetic, and per-bit
`getLsbD` reasoning for bitvector identities.

⚠️ Cite the reference side at the pinned `execution-specs` commit `e5a8caf1b` where the behaviour is in
question. `SpecRef` is a **port** and it drifts, so it is the proof target but not the authority on
semantics.

## Notes

[...]

Issue #11346, opened 2026-08-03
proof: triple for `account_is_eip161_empty` against `account_exists_and_is_empty`


pirapira:
One of the ten orienting rows from `docs/leaf-routine-targets.md` (row **3**), filed per-routine so each can be
taken independently. The table came from #11312.

## The pairing

| | |
|---|---|
| **RISC-V routine** | `account_is_eip161_empty` |
| **`SpecRef` target** | `account_exists_and_is_empty` (`EvmAsm/Stateless/SpecRef/StateTracker.lean:229`) |
| **caller** | `EvmAsm/Codegen/Programs/BlockVerdictSimpleTransferGas.lean:65` |
| **dependency status** | all callees verified — `rlp_list_nth_item` |
| **data structures** | account leaf RLP |
| **separation-logic `Assertion`** | `accountRlpIs` (as row 2, over `AccountRecord`) |

## The triple to state and prove

Informally: given `accountRlpIs ptr a`, the returned flag is `1` exactly when `a` equals `EMPTY_ACCOUNT` (zero nonce, zero balance, empty code).

State it as a CPS triple over the emitted routine, in a `…Spec.lean` module beside the program, using the
existing infrastructure:

- `EvmAsm/Rv64/CPSSpec.lean` for the triple form
- frame tactics `xcancel` / `seqFrame` / `runBlock` (`EvmAsm/Rv64/Tactics/{XCancel,SeqFrame,RunBlock,SpecDb}.lean`)
- `validMem`, `liftSpec`, `norm_addr` as needed

⛔ **No `native_decide` or `bv_decide`** — both are fully eliminated from the tree and CI enforces it
(`scripts/check-forbidden-tactics.sh` plus the `#print axioms` backstop in `scripts/check-axioms.sh`). Use
`decide` for concrete decidable goals, `omega` / `bv_omega` for linear (bit)vector arithmetic, and per-bit
`getLsbD` reasoning for bitvector identities.

⚠️ Cite the reference side at the pinned `execution-specs` commit `e5a8caf1b` where the behaviour is in
question. `SpecRef` is a **port** and it drifts, so it is the proof target but not the authority on
semantics.

## Notes

The EIP-161 emptiness predicate is load-bearing well beyond this routine — the same notion decides `is_account_alive`, and a tombstone-vs-live confusion around it produced two separate over-match defects in #11333. A proven statement of emptiness is worth more than its own row.
## Related

[...]

Issue #11345, opened 2026-08-03
proof: triple for `account_decode` against `decode_account_from_leaf`


pirapira:
One of the ten orienting rows from `docs/leaf-routine-targets.md` (row **2**), filed per-routine so each can be
taken independently. The table came from #11312.

## The pairing

| | |
|---|---|
| **RISC-V routine** | `account_decode` |
| **`SpecRef` target** | `decode_account_from_leaf` (`EvmAsm/Stateless/SpecRef/WitnessState.lean:117`) |
| **caller** | `EvmAsm/Codegen/Programs/State.lean:262` |
| **dependency status** | all callees verified — `rlp_list_nth_item` is `.proven` |
| **data structures** | account leaf RLP -> decoded account (nonce, balance, storage root, code hash) |
| **separation-logic `Assertion`** | `accountRlpIs` (`EvmAsm/Stateless/State/AccountAssertions.lean:118`, over `AccountRecord`) and `accountDecodedIs` (`:135`) — **both already exist** |

## The triple to state and prove

Informally: given `accountRlpIs ptr a`, after the call the four output cells satisfy `accountDecodedIs` for the same `a`.

State it as a CPS triple over the emitted routine, in a `…Spec.lean` module beside the program, using the
existing infrastructure:

- `EvmAsm/Rv64/CPSSpec.lean` for the triple form
- frame tactics `xcancel` / `seqFrame` / `runBlock` (`EvmAsm/Rv64/Tactics/{XCancel,SeqFrame,RunBlock,SpecDb}.lean`)
- `validMem`, `liftSpec`, `norm_addr` as needed

⛔ **No `native_decide` or `bv_decide`** — both are fully eliminated from the tree and CI enforces it
(`scripts/check-forbidden-tactics.sh` plus the `#print axioms` backstop in `scripts/check-axioms.sh`). Use
`decide` for concrete decidable goals, `omega` / `bv_omega` for linear (bit)vector arithmetic, and per-bit
`getLsbD` reasoning for bitvector identities.

⚠️ Cite the reference side at the pinned `execution-specs` commit `e5a8caf1b` where the behaviour is in
question. `SpecRef` is a **port** and it drifts, so it is the proof target but not the authority on
semantics.

## Notes

⭐ **The best-prepared row of the ten**: both the pre- and post-side assertions exist, and the one callee is already proven. This is the natural first target.

[...]

PR #11408, opened 2026-08-04
proof: the RLP walk predicates are deterministic (prereq for #11345/#11346)

dhsorens:
Shared infrastructure, landed on its own because two issues need it and neither should discover it independently. **No registry rows touched**; no issue closed.

## The gap

Every walk-family postcondition is existential:

```lean
∃ offset len, Success bytes base listLen index offset len ∧ <outputs written from offset/len>
```

Those existentials are whatever the **routine** found. A caller holding a *known* encoding — an `AccountRecord`'s four fields (#11345), the `EMPTY_ACCOUNT` comparison (#11346), a header's field 8 (#11351) — cannot conclude the output cells denote that value without knowing the found offsets **are** its field offsets.

That step is a determinism statement, and nothing in the tree had one. I went looking for it while sizing #11345 as "invert `Decoded` to `bytes = a.rlp`" and found that framing was wrong: the inversion isn't the content, this is. Since #11346 sits on the identical `StrictListPayload`/`StrictNthItem`/`rlpItemDecode` tower, it seemed worth extracting rather than proving twice.

## Two modules, because the predicates straddle the layering boundary

`rlpItemDecode` is core (`Rv64/RLP/WalkNext.lean`); `StrictNthItem`/`Success` are Codegen (`RlpListNthItemSAsmBase.lean`). `check-layering.sh` L1 forbids core importing Codegen, so:

| module | contents |
|---|---|
| `Rv64/RLP/WalkItemDeterminism.lean` | `rlpItemDecode_deterministic` |
| `Codegen/Programs/RlpWalkDeterminism.lean` | `strictListPayload_deterministic`, `strictNthItem_deterministic`, `success_deterministic` |

(`Field0ToU64.lean:173` already re-proved a walk fact at the core layer to dodge that import, so the split follows existing practice.)

## ⭐ Why `rlpItemDecode` is a function

It is `∃ b, bytes[off]? = some b ∧ (five disjuncts)`. The head byte is shared — `Option.some` is injective — and the five guards are **range-disjoint** on `B := b.zeroExtend 64`:

```
1  ult B 0x80                     2  ¬ult B 0x80 ∧ ult B 0xb8
3  ¬ult B 0xb8 ∧ ult B 0xc0       4  ¬ult B 0xc0 ∧ ult B 0xf8
5  ¬ult B 0xf8
```

so at most one arm applies, and within an arm `next` and `len` are pinned by explicit equations with no further choice. 25 cases: the 5 diagonals close by `rfl` after substitution; the 20 off-diagonal guard contradictions are discharged **uniformly** by pushing every `ult` through to `Nat` (`ult_eq_decide` plus the four prefix literals, which `omega` cannot evaluate itself) and calling `omega`. No `bv_decide`.

`StrictNthItem` follows by induction on the index — each step is pinned by the item lemma, which also forces the offset handed to the next step. `StrictListPayload` is a two-constructor split on the same `ult B 0xf8` guard, with `endPtr` already forced by the `listLen` index. `Success` composes the two.

## ⚠️ What is deliberately not claimed

That a decode **exists**, or that it agrees with `EL.RLP.decodeAux`. The latter is **false** for the nested-list arms — the `c3 c2 81 00` counterexample I posted on #11341 — and is a question about the *verdict* column, which is still with you. Determinism is orthogonal to it and true unconditionally, which is why it can land now.

## On "availability is not use"

This module is consumer-free today, and I want to be explicit that I don't think that trips the rule I've been enforcing elsewhere. The rule is about a **registry row claiming a grade** on the strength of an unconsumed lemma. No row is added or re-graded here. The consumers arrive with #11345/#11346; if you'd rather this were folded into #11345's PR instead of landing separately, say so and I'll squash it there — I split it because it is one reviewable idea and two issues depend on it.

## Verification

`port-check.sh` **PASS** on both modules. `lake build` clean (4807 jobs, 0 warnings); `check-axioms` 126 witnesses classical-only; registry-crosscheck (17 × 23), layering, progress, forbidden-tactics, file-size, naming all pass. Submodule absent from the commit (checked).

Prereq for #11345, #11346.

### DIFF
--- a/EvmAsm/Codegen/Programs/Imports.lean
+++ b/EvmAsm/Codegen/Programs/Imports.lean
@@ -72,6 +72,7 @@ import EvmAsm.Codegen.Programs.TxTypeDispatchTyped
 import EvmAsm.Codegen.Programs.TxTypeDispatchTop
 import EvmAsm.Codegen.Programs.TxTypeDispatchDischarge
 import EvmAsm.Codegen.Programs.RlpRead
+import EvmAsm.Codegen.Programs.RlpWalkDeterminism
 import EvmAsm.Codegen.Programs.RlpListNthItemCallSAsm
 import EvmAsm.Codegen.Programs.RlpWalkCallSAsm
 import EvmAsm.Codegen.Programs.ReceiptExtractLogsBloomCallSAsm
--- a/EvmAsm/Codegen/Programs/RlpWalkDeterminism.lean
+++ b/EvmAsm/Codegen/Programs/RlpWalkDeterminism.lean
@@ -1,0 +1,97 @@
+/-
+  EvmAsm.Codegen.Programs.RlpWalkDeterminism
+
+  The `Codegen`-layer half of the walk determinism story: `StrictListPayload`,
+  `StrictNthItem` and `Success` are **functions** of the bytes they read.
+
+  The core-layer half — `rlpItemDecode_deterministic` — lives in
+  `EvmAsm/Rv64/RLP/WalkItemDeterminism.lean`, because it is about a predicate the
+  verified core owns. The predicates below are defined under `Codegen`
+  (`RlpListNthItemSAsmBase.lean`, `RlpListNthItemStrictList.lean`), so their determinism
+  has to be stated here; `check-layering.sh` L1 forbids the other direction.
+
+  WHAT THIS UNBLOCKS. Every walk-family postcondition is existential —
+  `∃ offset len, Success bytes base listLen index offset len ∧ <outputs from offset/len>`.
+  Those existentials are whatever the *routine* found. A caller holding a **known**
+  encoding (an `AccountRecord`'s four fields, a header's field 8) cannot conclude the
+  outputs denote that value without knowing the found offsets are the right ones. That
+  step is exactly `success_deterministic`, and it was missing — which is the actual
+  content of #11345 and #11346, both of which sit on this tower.
+
+  ⚠️ This module is deliberately consumer-free: it is a prerequisite landed on its own so
+  it can be reviewed as one idea. The consumers arrive with #11345/#11346. That is not the
+  "availability is not use" failure — no registry row claims a grade on the strength of
+  these lemmas.
+-/
+
+import EvmAsm.Rv64.RLP.WalkItemDeterminism
+import EvmAsm.Codegen.Programs.RlpListNthItemSAsmBase
+
+namespace EvmAsm.Codegen.RlpListNthItemSAsm
+
+open EvmAsm.Rv64 EvmAsm.Rv64.RLP
+
+/-- The list header determines both the payload cursor and the exclusive end.
+
+    `endPtr` is forced by the `listLen` index alone (both constructors conclude
+    `base + BitVec.ofNat 64 listLen`), so the content is the cursor: the `short` arm
+    yields `1` and the `long` arm `1 + lenOfLen`, and the two are separated by the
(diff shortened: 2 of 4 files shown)


pirapira:

## Verified, not inferred

I built both modules in a detached worktree at the PR head rather than relying on CI alone — for a proof PR, type-checking *is* the deliverable:

- `lake build EvmAsm.Rv64.RLP.WalkItemDeterminism` → **Built (972 jobs)**
- `lake build EvmAsm.Codegen.Programs.RlpWalkDeterminism` → **Built (1032 jobs)**

And `#print axioms` on the two headline theorems:

| theorem | axioms |
|---|---|
| `strictListPayload_deterministic` | `[propext]` |
| `rlpItemDecode_deterministic` | `[propext, Quot.sound]` |

Classical only — no `sorryAx`, no `Lean.ofReduceBool`, no `Lean.trustCompiler`. The `decide` calls in
`prefix_literals` are the sanctioned kind (concrete decidable propositions, kernel-checked); nothing in the diff
matches `bv_decide`/`native_decide`.

## The proof itself is the right shape

`rlpItemDecode` is `∃ b, bytes[off]? = some b ∧ (five disjuncts)`. Determinism falls out of two facts: the head byte
is shared because `Option.some` is injective, and the five arms are guarded by **range-disjoint** `ult` tests on
`0x80/0xb8/0xc0/0xf8`, so at most one applies. Within an arm, `next` and `len` are pinned by explicit equations.
Hence 25 cases — 5 diagonal closing by `rfl` after substitution, 20 off-diagonal as guard contradictions discharged
uniformly by pushing `ult` to `Nat` and calling `omega`. `ult_mono` and `prefix_literals` exist precisely because
`omega` cannot evaluate `BitVec.toNat` of a numeral, which is the right reason for a helper to exist.

The `StrictListPayload` half is a clean 2×2 on `short`/`long` with the cross cases killed by the shared head byte
plus the `ult B 0xf8` guard.

## Two things I want to single out, because they are rarer than the proof

**It states what it does NOT claim.** Explicitly not: that a decode *exists*, and not that it agrees with
`EL.RLP.decodeAux` — with the note that the latter is **false** for the nested-list arms, citing the `c3 c2 81 00`
counterexample on #11341. Negative scoping like that is what stops a later reader over-claiming a lemma, and it is
usually absent.

**It pre-empts the availability-is-not-use objection honestly.** The module is deliberately consumer-free, landed
alone so it reads as one idea, with the consumers arriving in #11345/#11346. I checked the claim: the diff is four
files — the two new modules plus one `import` line each in `Imports.lean` and `RLP.lean` — and **no registry or
progress file is touched**, so no row claims a grade on the strength of these lemmas. That is the distinction that
matters, and the PR draws it correctly rather than asserting it.

## Why the placement is right

Living under `Rv64/RLP/` rather than `Codegen/Programs/` is deliberate and justified: `rlpItemDecode` is a predicate
the verified core owns, `check-layering.sh` L1 forbids the reverse import, and the PR notes `Field0ToU64.lean:173`
already re-proved a walk fact at this layer to dodge exactly that boundary. Splitting the two halves across the
layering line, with the `Codegen`-defined predicates' determinism stated on the `Codegen` side, is the only
arrangement that respects L1.

The motivation is also correct on its own terms: every walk-family postcondition is existential
(`∃ offset len, Success … ∧ outputs from offset/len`), so a caller holding a *known* encoding cannot identify the
found offsets with it. That step is exactly determinism, and nothing in the tree had it.

Issue #11410, opened 2026-08-04
soundness: replace the static code-preimage reprieves with the existing dynamic code_reads set (closes two FA gaps)


pirapira:
**Attribution: coord**, from k3's classification of `bal_code_preimages_valid` on
[#11183](https://github.com/Verified-zkEVM/evm-asm/issues/11183#issuecomment-5179762778). **C-class re-derivation, not
a deletion** — the check itself is spec-faithful and load-bearing.

## What stays and what moves

`bal_code_preimages_valid` **must not be deleted**: it is the only thing that catches a witness whose BAL matches the
rebuild but whose `svf_codes` lacks a required keccak-verified preimage (fail 11), and the spec agrees such witnesses
are invalid (`witness_state.py:204-212`, `self._code_db[code_hash]` → `KeyError` when execution reads the code). See
#11183's amended exit condition and its carve-out.

What is BAL-shaped is only the **selection**: which accounts get their preimage checked, taken from supplied-BAL row
shapes (change-list emptiness per row, plus the 20-byte address for `bbcv_bal_contains_addr` membership). Every *value*
already comes from an authenticated source — the state trie via `extcodesize_at_header_state_root` under the parent
header state root, and keccak over `svf_codes`.

## The change

Select from **the accounts execution actually reads code from** — the guest's counterpart of the spec's dynamic
`get_code` call set — instead of from BAL row shapes.

This is the same transformation #11378 made for BLOCKHASH: a **static, pattern-based** predicate replaced by a
**dynamic, execution-derived** one, with the static scan removed in the same change so the new mechanism is fed rather
than merely present. Note how that one went — the gate has to sit where *every* path reaches it. k3's first placement
there was bypassed by settlement paths jumping into the middle of the gas-gate prelude, and the bypass was a genuine
**false accept** found only by a targeted negative row while the random-200 stayed clean at FR=8/OK=232 with zero flips
in both directions.

## Secondary benefit: the reprieve patterns

Today's touch-only reprieve list is a **static pattern set** — PUSH20+EXTCODEHASH/BALANCE/SELFDESTRUCT,
ADDRESS..SELFDESTRUCT, CREATE-collision touches, fee recipient, system contracts, PUSH20+CALL delegated-code. A static
pattern set over-requires when the pattern sits in unreachable code and under-requires on dynamic shapes — exactly the
static-vs-dynamic gap #11378 closed for BLOCKHASH. Selecting from actual code reads closes it here too.

[...]

PR #11411, opened 2026-08-04
docs(plan): record the Class-A provided-BAL programme and #10651's completion

pirapira:
**Attribution: coord.** Documentation only — `PLAN.md`, +66 lines, no code.

## Why

`PLAN.md` named "#10651 (compute the state root from operations, not from the BAL)" as a **future** forced-order
prerequisite. That headline is now **done for user-tx accounts** (PR #11389): the seeding block is a bare
`j .Lbsr_bal_copy_next` and `execution_map_state_changes` is the sole root authority. Its successor line — the Class-A
programme, #11183 — was absent from `PLAN.md` entirely, despite being this session's main structural thread and the
source of eleven merged or labelled PRs.

## What it records, chosen as the things a reader would otherwise re-derive

- **#11183's literal exit condition was unreachable** and now carries a **witness-integrity carve-out**.
  `bal_code_preimages_valid` must **not** be deleted: `bal_serializer_verify` never reads `svf_codes`, so removing it
  turns a clean fail-11 reject into a **false accept** whenever the missing code does not move the post-state root —
  and the spec agrees such witnesses are invalid (`witness_state.py:204-212` raises `KeyError` when execution reads the
  code, *regardless of state effect*). Someone will otherwise try to delete it under the hash-discharge argument.
- **Which reads are retired and which remain, with the blocker for each** — including that #11406 is blocked on
  publishing the auth marker into AccountState because *write maps are attribution, not state*, and that #11407 cannot
  be deleted at all until a per-tx storage-set counter exists, since `storage_writes` is a cumulative latest-value map
  cleared by `write_sets_incorporate_tx`.
- **Why the ratchet is a ratchet** and not an assertion of the invariant: the invariant is not yet true, so a guard
  asserting it would be red on arrival and would get disabled.
- **`an unused datum is not an unused status`**, with all three near-deletions named. This one cost the most time
  tonight and is the easiest to repeat.
- **Gate equality is the per-check test for retiring a strictness**, with both the case where it succeeded (#11392,
  shared `bv_bal_shadow_ready`) and the case where it fails (`bal_code_preimages_valid`, whose property the hash cannot
  see). Byte coverage alone is necessary but not sufficient.
- That `bsr_changed_accounts` survives #11389 as a **dedup list, not a work list** — misreading it as a work list would
  make the cut look wrong.

## Evidence

`scripts/check-file-size.sh`: 2164 files checked, all within cap. No `EvmAsm/**` file touched, so no build, pin, or
fixture implications, and it contends with nothing in the merge queue.

The claims are sourced from tonight's PR reviews and issue analyses (#11389, #11392, #11396/#11400, #11398/#11403,
#11402/#11404, #11405, #11406, #11407, #11409, #11410) rather than restated from memory; each carries its spec cite or
emitted-site reference inline.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -277,6 +277,72 @@ All deleted spec files have been recreated. See **Pending: Recreate Deleted Spec
   `BlockVerdictMtxRuntime:399`, are later slices. This bead is the forced-order
   prerequisite for #10651 (compute the state root from operations, not from the BAL).
 
+- **#10651's headline is now DONE for user-tx accounts** (PR #11389, GH #11326 entry 6):
+  `BlockVerdictStateRoot.lean` no longer seeds `bsr_changes` from the supplied BAL — the
+  seeding block is a bare `j .Lbsr_bal_copy_next`, the RLP walk is retained only so the
+  cursor advances, and `execution_map_state_changes` is the **sole root authority**. The
+  state root is computed from operations, not from the BAL, matching the spec's data flow:
+  `fork.py:928-930` builds the BAL **from** tracked state *after* execution and
+  `:366`/`:390` bind only `hash_block_access_list`. Evidence the seed was redundant rather
+  than merely unused: a −484-byte removal left the random-200 **byte-identical** (FA=0,
+  forward set empty) — which is what the #11382 TOUCHED producers bought. `bsr_changed_accounts`
+  survives as a **dedup/already-owned list, not a work list**, so system-only content is now
+  its correct content. Residual: the modeled 2935/4788 arms still pre-seed and take
+  precedence over the map (GH #11390, PR #11399 — union, not precedence, since
+  `process_unchecked_system_transaction` at `fork.py:897-907` mutates the same tracked state
+  as everything else).
+
+- **The Class-A programme** (GH #11183) is #10651's successor line: *the guest must not read
+  the provided BAL as an **execution input***. ⛔ Its original wording — "except to hash it" —
+  proved **unreachable** and was amended 2026-08-04 with a **carve-out**: a witness-integrity
+  check may read BAL row *shapes* to select what to verify, provided every value it trusts is
+  authenticated and its only effect is to reject. `bal_code_preimages_valid` satisfies the
+  carve-out and **must not be deleted** — `bal_serializer_verify` never reads `svf_codes`, so
+  removing it turns a clean fail-11 reject into a **false accept** whenever the missing code
+  does not move the post-state root, and the spec agrees such witnesses are invalid
+  (`witness_state.py:204-212` raises `KeyError` when execution reads the code, *regardless of
+  state effect*). On the accept path the serializer hash pins the supplied BAL byte-for-byte to
+  the execution-derived rebuild, so what remains BAL-sourced there is only *selection*, never a
+  value.
+
+  Retired so far: the `runtime_current_bal_ptr` handoff into same-block delegation (#11396 /
+  PR #11400), the granular MtxTail comparators 37/44/45 together with the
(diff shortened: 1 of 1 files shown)


pirapira:
**Attribution: coord.** Updated: the Class-A section carried **my own analysis of #11406 that the session then refuted**, and
landing it unchanged would have put a wrong root cause into `PLAN.md`.

## What was wrong

I had written that #11406 was *"blocked on publishing the 23-byte auth marker into AccountState first."* It is not. The
producer works, the journal survives, the key and flags are right, and the delegate target's code was never the problem.

**The generic code lookup finds the 23-byte `ef0100` marker and the handler treats it as ordinary code**, bypassing
delegation resolution — so the marker is *executed* instead of *followed*. The spec does the same lookup and then
recognises the designator: `is_valid_delegation` (`vm/eoa_delegation.py:40`) plus `get_delegated_code_address` (`:62`).
**Recognition is a property of the code bytes, not of which store they came from** — which is why the old BAL path worked by
accident: it *missed* in CodeState and so never had a marker to mistake for code.

Four hypotheses were tried before that, each costing a candidate build, so I have recorded the root cause **with the reason
the diagnosis was hard**: `0xEF` is a reserved invalid opcode (EIP-3541), so executing the marker halts exceptionally and
burns the frame's gas *without mutating state*. The post-state root therefore **matched** while `succ` did not — and I read
root-matching as reassurance rather than as evidence about where the failure was **not**.

## And #11407 moved from "needs a counter" to a deletion

It landed as PR #11424 (merged via #11437) with **no counter built**: the fallback computed `97,920 × prior-tx storage-set
count` while the primary sums *all* prior state gas, so the two were never equivalent; it was **FA-lax**; and it was
**unreachable on any accepting path**, because the routes that starve `bvgr_runtime_count` also set
`bv_dispatch_runtime_status` nonzero and the receipts tail rejects at fail64.

The note worth keeping: **no sweep could have found it.** The path is unexercised by the corpus, so only computing *both*
quantities on the rows where the primary succeeds compared the two definitions. And the ÷1530 decode is what named the
cause — it turned "183600 vs 0" into "120 bytes of creation charge".

Its dead **transport** then outlived its consumer and needed its own cut (#11428 / PR #11437) — a recurrence of #11405's
defect one register pair later.

`check-file-size.sh` clean; docs only, no `EvmAsm/**` file touched.

Still deliberately unlabelled — this is my own PR, so it wants someone else's judgement.

pirapira:
Closing without merging — this content reads as issue material rather than plan material, and the maintainer's call is the right one.

What it actually was: a consolidated record of the Class-A provided-BAL programme — #10651's completion, the amended exit condition and its witness-integrity carve-out, the retirement ledger, #11406's root cause after four wrong hypotheses, #11407's resolution-as-deletion, the ratchet's design rationale and blind spots, and the gate-equality test. Every one of those is a **tracked item with open questions and a live status**, which is what an issue is for. `PLAN.md` is a poor home for it: it goes stale silently, it has no notification surface, and it cannot be closed when a bead completes.

**Moved verbatim to #11183** (the Class-A programme issue) — https://github.com/Verified-zkEVM/evm-asm/issues/11183#issuecomment-5183622448 — with only a header added. Nothing is lost by closing this PR.

One thing worth keeping visible from it, since it has already cost real time twice tonight, is recorded in the ratchet's own blind-spot list: **symbol-only scanning is evadable by an address literal.** `AccountWriteMap` reaches its arenas via `li 0xa2b20000`, `0xa28a0000`, `0xa2d20000`, `0xa21a0000`, `0xa1fa0000`, `0xa23a0000`, none of which appears in `symbol-addresses.tsv`. Any completeness claim resting on a symbol scan inherits that hole.

PR #11413, opened 2026-08-04
codegen: publish delegation code in AccountState

pirapira:
## Summary

Publish accepted EIP-7702 delegation code in the execution-state `AccountState`
overlay.  The existing 15-site BAL resolver still remains in place; this is the
producer landing only.  A follow-up PR will switch those consumers and delete
`bal_same_block_delegation_code_resolve`.

The execution-specs ordering is the reason for this shape: `eoa_delegation.py:245-285`
calls `set_code(tx_state, authority, code_to_set)` before message execution, and
`state_tracker.py:get_code` consults transaction code writes before parent/witness
code.  The post-execution BAL (`fork.py:928-930`) is not an execution-state source.

## Producer details

`account_state_record_auth` now publishes the 23-byte `ef0100 || address` marker
at AccountState `+72` with length `+80`, and uses the same block-lifetime slot for
the optional AccountWrite record.  AccountState flags are `0x5f` (95) for a
nonzero target: bit 16 execution-known, bit 8 delegation, bit 4 code-present,
bit 2 account-present, plus the existing nonce/occupied bits.  A clear target
retains execution-known/account-present and omits code/delegation-present.

The consumer contract was probed in emitted code, not inferred from source: on
fixture `00017`, immediately after the producer, `account_state_lookup_current`
returned `a0=1`, `a1=0x00000000babde608`, `a2=0x17` (23).  The temporary probe is
not part of this PR.  An independent `SPIKE_WATCH` on AccountState `+72`
(`0xbc265448`) observed `0 -> 0xb9cee608` at PC `0x80037418` on fixtures 00017
and 02256.

### Flag-bit audit

The full nonzero-target mask observed at `+88` is `0x5f` (95): bits 16
(execution-known), 8 (delegation-present), 4 (code-present), 2
(account-present), and the existing occupied/nonce bits.  A clear target is
`0x53` (83).  The only AccountState bit-8 consumers in the emitted guest are
`eip7702_authority_asof` (delegation-present output) and
`account_state_tombstone_balance_zero` (created-this-transaction filter); the
other bit-8 reads are AccountWrite flags at a different offset.  The old auth
producer already set bit 8 for nonzero targets, so this PR does not introduce
that meaning; it adds the required bits 16 and 4.  Direct temporary probes
returned tombstone status 0 for both a delegated row (`00017`, mask `0x5f`,
bit 4 set) and a clear row (`00013`, mask `0x53`, nonce 1), so the tombstone
aliveness gate is unchanged.  As a direct discriminator, clearing only bit 8
from the delegated row (`0x5f -> 0x57`) before calling the resolver still
returned status 1 with the same pointer/length.  The resolver therefore needs
only bits 16, 2, and 4; bit 8 remains the pre-existing delegation-present
signal rather than a new created-this-transaction claim.

## Evidence

Prediction: this producer changes no existing verdict bytes; `.data` and `.bss`
remain unchanged and `.text` grows only by the producer sequence.  Candidate was
built from base `645f279c85e2e1a1bc603239872da54fb29f14b6` (main had since moved to
`dfdf26f1a`).

- candidate ELF: `1339137f7844c466d87ff2bc1da9c6b67333d590bb2a9cbb47c89dc54a46bfa5`
- current-main parent ELF: `53106088d1a26876ba54470e8bd7db0a4fe0b15a8fcce8aa96cbcecfbe5456a3`
- candidate `.text` `0x5af40` vs parent `.text` `0x5af24` (`+0x1c`); `.data` `0x5370` and `.bss` `0x1a0f2ca0` unchanged
- named `test_single_authorization_charges` fixture: candidate 1/1 exact, parent 1/1 exact; output files byte-identical
- candidate run timestamp: `2026-08-04T13:50:46Z`; `lake build EvmAsm.Codegen.Programs.TxIntrinsicStateGas` passed (1046 jobs)

No grade or registry row is claimed here.  This is intentionally a producer
landing without its consumer; the 15-site consumer switch is separate.


pirapira:
**Attribution: coord.** Reviewed against `origin/main`. **Labelled `merge!`.** The producer half of #11406; the 15-site
consumer switch follows separately.

## The producer satisfies the consumer contract, proven at the consumer

`account_state_lookup_current` requires, at `.Laslc_entry`: `flags@+88` with value-16 **and** value-2 **and** value-4
set, then the code pointer at **+72** and length at **+80**. Confirmed on fixture 00017: `+72` = `0xbabde608`,
`+80` = `0x17` (23), `+88` = `0x5f`, and a direct lookup returning `a0=1`, `a1=0xbabde608`, `a2=0x17`.

The flag arithmetic checks out: `ori 83` (values 64+16+2+1) then `and -13` (clears 4 and 8), then on a nonzero target
`ori 12` (sets 4 and 8) → `95 = 0x5f`. On a **zero** target the row keeps 83 with `+72`/`+80` zeroed.

**Probing the consumer rather than the cell is what made this correct.** The first candidate used mask `-29`, which
cleared value-16 — the bit the resolver tests *first* — so the row read as **absent** while `+72` held a perfectly good
pointer. A `SPIKE_WATCH` on `+72` alone would have shown a live, correct write and the producer would have looked proven
while delivering nothing. That is the correct-but-unfed shape, and only the injected direct-lookup probe caught it.

## The clear-before-set ordering matters and is right

`sd zero, 72(t0); sd zero, 80(t0)` and the `and -13` run **before** the conditional set, so an authorization that
*clears* a delegation cannot inherit stale code from a previous one. Had the producer only ever set fields, a
delegation-clear would have left the old pointer live with the code-present bit still on — a stale-code read on exactly
the path least likely to be tested.

## The value-8 hazard was real and is resolved by disjointness

Value 8 is read by three consumers: `eip7702_authority_asof` (delegation-present output — the intended reading),
`account_state_tombstone_balance_zero` (which filters it as **created-this-tx**), and AccountWrite-map consumers (a
*different* flags word at `+112`, so not implicated). Setting it on authority rows is safe because the tombstone helper
is reached only from SELFDESTRUCT-beneficiary and new-account gates, **never from auth publication**, and an injected
consumer probe showed no changed tombstone answers.

Worth being explicit that safety here rests on **path disjointness, not on the bit being unambiguous** — and the
tombstone helper's conjuncts are an aliveness gate whose previous defect was a *silent NEW_ACCOUNT overcharge at every
caller*. The docstring correction in this PR (`+88` now spelled out as value-16 execution-known, value-4 code-present,
value-2 account-present, value-8 delegation) fixes one of the two divergent descriptions; the other, at `:551`, still
calls the same bit `bit3 (created-this-tx)`. Filed separately rather than widened here.

Reusing one `eip7702_auth_code_slots` entry for both producers rather than allocating twice is also right — two
block-lifetime slots for one authorization would be a capacity leak.

`.text +0x1c` (28 bytes) reconciles with three stores plus the flag change; `.data`/`.bss` unchanged. Candidate
`1339137f`, parent `53106088`, base `645f279c` stated (main has since moved to `dfdf26f1a`). Body states plainly that
this is a producer landing without its consumer and claims no grade — the framing #11408 established tonight.

pirapira:
**Attribution: coord.** Addendum correcting my own review above. **Label stands.**

I framed the value-8 question as something this PR introduces. **It does not.** `origin/main:412` already reads:

```
ld t1, 88(t0); ori t1, t1, 67; li t2, -9; and t1, t1, t2; beqz s2, .Lasra_flags; ori t1, t1, 8
```

`and -9` clears value 8, then a nonzero target sets it with `ori 8`. So **the old producer already set value 8 on
authority rows for a nonzero auth**, and this PR changes value-8 behaviour not at all — it adds values 4 and 16 plus the
`+72`/`+80` stores and the clear-before-set.

That matters for how the risk is read. The overloaded-bit hazard is **pre-existing**, not introduced here, so the
disjointness argument (`account_state_tombstone_balance_zero` reachable only from SELFDESTRUCT-beneficiary and
new-account gates, never from auth publication) is what has been holding all along rather than something this PR needs
to establish. The author's probes — tombstone result 0 for nonzero 00017 with value-4 set, and for the clear case 00013
with nonce 1 — confirm the answer is unchanged, which is the right check for a change that leaves the bit alone.

My earlier phrasing implied a new exposure. It was a pre-existing one I hadn't looked at, surfaced because this PR
touches the adjacent bits. The investigation was still worth its ten minutes: it is how we learned the bit has two
readers with different interpretations, which is now filed separately.

pirapira:
Needs a rebase onto current `main` — merging `main` conflicts in a **non-generated** file.

**Conflict:** `EvmAsm/Codegen/Programs/TxIntrinsicStateGas.lean`, 1 hunk, and both sides are non-empty — a genuine overlap rather than an add-vs-delete:

- **`main`** (after #11387): 11 code lines in the `code_state_mtx_active` / `.L77prep_next` path
- **this PR**: 3 code lines over the same region

Since both sides carry real emitted assembly for the same code path, taking either wholesale would drop the other's behaviour, and I can't tell from the diff alone which combination is intended — #11339 and #11413 both touch the EIP-7702 preparation path, so the merged form is a semantic decision.

Everything else was clean (generated layout files only), so this should be a small rebase.

Rebase and I'll pick it up next pass — full local verification (`lake build` + all 20 gates) before it reaches CI.

pirapira:
Heads-up while this is parked: **#11433** (`make EIP-7702 marker build path-independent`) is in flight and touches the same file this PR conflicts on — `EvmAsm/Codegen/Programs/TxIntrinsicStateGas.lean` (+4/−5 there). It is verified green and auto-merging, so it will likely land before you rebase.

That means your rebase target for that hunk is about to change again. It may help rather than hurt: #11433 reworks the marker build path in the same region, so the `code_state_mtx_active` / `.L77prep_next` overlap I flagged may look different — possibly simpler — once it lands. Worth rebasing **after** #11433 rather than before, so you resolve against the final shape.

pirapira:
**Attribution: coord.** Handback resolved at `2999f3f02`. **`complicated-merge` may be removed** (the author's to clear), and
the `merge!` label stands.

## All four conditions verified by count, not by reading

| condition | want | got |
|---|---|---|
| `code_state_mtx_active … beqz .L77prep_next` (would revert #11433) | 0 | **0** |
| mask site `… + accountWriteHasState + accountWriteHasTouched` (= 46) | 1 | **1** |
| `#guard "li a6, 46; jal ra, account_write_record"` intact | 1 | **1** |
| `origin/main` is an ancestor of the tip | yes | **yes** |

`.text 0x590b0` (from `0x590c0`, **−0x10** for the deleted 3-instruction gate), `.data`/`.bss`/arena bases unchanged,
`check-region-map` clean.

## Why this took three rounds, recorded because the region will be touched again

`TxIntrinsicStateGas.lean`'s authorization block has been changed by **three** separate landings tonight — #11382 added the
TOUCHED bit to the write-map mask, #11427 removed the `:400` auth-prep skip, and #11433 removed the `:332` marker-allocation
skip — and this PR restructures the same block. A conflict resolution in a region with that much recent history will
**silently restore whichever version the author started from**, which is what happened twice here:

1. **The mask was lowered 46 → 14** and the `#guard` edited to match. But that guard's own comment states its purpose: *"Keep
   the state bit pinned so a future mask edit cannot silently make `Optional[Account]` absence indistinguishable from zero
   fields."* It existed to stop exactly that edit. **A failing `#guard` is evidence, not an obstacle** — when an assertion you
   did not write fails after a merge, the first hypothesis should be that it caught something. Dropping TOUCHED would have let
   an auth-only account fall out of the write map, and since #11389 made the map the sole root authority, that is a state-root
   consequence rather than a cosmetic one.
2. **The `code_state_mtx_active` gate was reintroduced**, which would have made the marker build MTx-only again and undone
   #11433's path-independence — the very split #11423 exists to remove.

Both were caught by comparing **counts against `origin/main`** rather than by reading the diff, and the second was only
conclusive because the tip was verifiably rebased: main is an ancestor, so a difference from main is something the branch
*adds*, not something it lacks.

The justification offered for the mask change — that an earlier commit carried the 14-mask contract — is the **stale-base error
in its most consequential form**: citing an older commit to justify reverting a newer one. It is the second instance of that
shape tonight; the first was mine, contradicting a correct analysis from my own dirty working tree.

## Resolution as landed

PR structure plus main's intents: the shared `s8` marker slot and 3-line allocation from this branch, main's TOUCHED mask and
its pinned guard, and no `mtx_active` gate. Both sides' intents survive, which is what an opposed-intents conflict requires —
the merge actor's warning on #11404 said the same thing about a different file, and it applies here.

PR #11415, opened 2026-08-04
chore(codegen): unlink dead storage subtree (#10681)

pirapira:
## Summary
- Unlink dead guest text/data for the storage-half BAL checks closed by #10681:
  - `bal_storage_matches_exec_log`
  - `bal_storage_covers_exec_log`
  - `bal_all_accounts_storage_consistent` (+ skip_list wrapper body was already unused)
- Justification: **0 live `jal` reachers** in the guest (measured); storage half is redundant under `bal_serializer_verify` hash equality (rebuild from `tx_storage_writes`, not supplied BAL).
- Probe units + Function strings kept probe-only (#11103 style).
- Tuple skip-list (check 42) **unchanged** — separate question under #10646 / coord dump-first.

## Layout
| pin | before | after |
|---|---|---|
| `textSizeBytes` | `0x5af24` | `0x5a334` (−3056 B) |

Regen: `gen-symbol-addresses.py --build` → `asm_to_program.py guest-addrs` → `guest_image_coverage.py --emit-lean` (no GuestImageEntries delta) → `gen-region-map-link-pins.py`; second pin pass NO-OP; `check-region-map.sh` CLEAN; `GuestImage` builds.

## Test plan
- [x] `check-region-map.sh` CLEAN
- [x] dead symbols absent from ELF / `.s` / TSV
- [x] `lake build EvmAsm.Codegen.Proofs.GuestImage`
- [ ] CI full gate
- [ ] optional r200 A/B (expect byte-identical OUTPUT; ELFs distinct)

Closes coordination follow-up for #10681 dead-subtree unlink (issue already closed).


pirapira:
**Attribution: coord.** Reviewed against `origin/main`. **Labelled `merge!`.** Cleans up a dead subtree left by #11392,
now that #10681 has established the storage hole is not real.

## Three independent supports, which is why this is safe

1. **Measured**: `bal_storage_matches_exec_log` 0/60 and `bal_all_accounts_storage_consistent_skip_list` 0/60 on the
   `eip7928_block_level_access` filter — with the instrument's own false positive fixed first (requiring the stopped PC to
   equal the target rather than any stop, which is the kind of correction that makes a zero believable).
2. **Static**: after #11392 removed the MtxTail callsites, the only remaining `jal`s were at
   `BalAllAccountsStorage.lean:141`/`:174`, inside a wrapper that is itself unreached — a closed dead subtree, not a
   symbol with a caller somewhere.
3. **Argued**: #10681 closed on evidence that the survivor hash covers exec-vs-BAL storage consistency for every account
   including the recipient, because the rebuild is filled from `tx_storage_writes` and system capture and **never** from
   the supplied BAL. So this subtree is redundant rather than a check we are losing.

Any one of those alone would have left a doubt; together they cover measurement, reachability, and purpose.

## Verified

- `.text 0x5af24 → 0x5a334` = **−0xbf0 = −3056 bytes**, matching the reported figure for the whole subtree.
- **`check-asm-to-program.sh` CLEAN** (385 defs / 146 files) with **no orphans** — the three unlinked routines had no
  fixtures or `MANIFEST.tsv` rows, so nothing was left stranded. This was checked *before* labelling precisely because the
  sibling PR #11400 was handed back for the mirror-image failure (a live routine whose MANIFEST row went missing). A
  routine's fixture and its row must move together, in both directions.
- The three modules retain importers on this branch (2, 2, 1), so `check-unimported.sh` is unaffected: **emission** is
  unlinked, source reachability is untouched. That is the right conservative split.

## Follow-up, not a blocker

The three modules are now **source without emission** — the #11383 class. Deleting the source is a separate, cheaper
decision once nobody wants them wired up again, and keeping them for now is defensible while #10646 is still open on the
adjacent tuple question. The comments were updated to cite #10681 rather than left claiming the old purpose, which is the
part that usually rots.

## My miss, recorded

I reviewed #11392 and confirmed the *calls* were gone without checking the *symbols* were — the source-vs-emitted gap in
miniature, on the very PR whose review I used to explain that gap. The dead subtree survived a review that should have
caught it.

pirapira:
Folded into #11417, which batches this with the other ready `merge!` PR of mine into one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11417 lands.

Issue #11348, opened 2026-08-03
proof: triple for `bloom_or_into` against the OR-accumulation step of `logs_bloom`


pirapira:
One of the ten orienting rows from `docs/leaf-routine-targets.md` (row **6**), filed per-routine so each can be
taken independently. The table came from #11312.

## The pairing

| | |
|---|---|
| **RISC-V routine** | `bloom_or_into` |
| **`SpecRef` target** | the OR-accumulation step of `logs_bloom` (`EvmAsm/Stateless/SpecRef/Fork.lean:101`) |
| **caller** | `EvmAsm/Codegen/Programs/Bloom.lean:761` |
| **dependency status** | leaf; sibling leaf `bloom_eq` is the same shape and can follow immediately |
| **data structures** | 256-byte bloom filter |
| **separation-logic `Assertion`** | `bytesRegion` (256 bytes) suffices |

**The fold only** — not the per-log index derivation, which is a separate obligation.
## The triple to state and prove

Informally: the destination region becomes the bytewise OR of its previous contents with the source region, over all 256 bytes.

State it as a CPS triple over the emitted routine, in a `…Spec.lean` module beside the program, using the
existing infrastructure:

- `EvmAsm/Rv64/CPSSpec.lean` for the triple form
- frame tactics `xcancel` / `seqFrame` / `runBlock` (`EvmAsm/Rv64/Tactics/{XCancel,SeqFrame,RunBlock,SpecDb}.lean`)
- `validMem`, `liftSpec`, `norm_addr` as needed

⛔ **No `native_decide` or `bv_decide`** — both are fully eliminated from the tree and CI enforces it
(`scripts/check-forbidden-tactics.sh` plus the `#print axioms` backstop in `scripts/check-axioms.sh`). Use
`decide` for concrete decidable goals, `omega` / `bv_omega` for linear (bit)vector arithmetic, and per-bit
`getLsbD` reasoning for bitvector identities.

⚠️ Cite the reference side at the pinned `execution-specs` commit `e5a8caf1b` where the behaviour is in
question. `SpecRef` is a **port** and it drifts, so it is the proof target but not the authority on
semantics.

## Notes

Scoping the fold apart from the index derivation is what makes this startable — the index derivation needs keccak and is a much larger obligation.
## Related

- `docs/leaf-routine-targets.md` — the full table, including the transitive-leaf rule that admitted this row
- #11312 — the orienting issue this came from (closed; artefact delivered)
[...]

PR #11416, opened 2026-08-04
proof(#11348): logs_bloom decomposes into a pointwise OR

dhsorens:
The **reference half** of #11348. Does not close the issue — the machine half (the `Fn.retSpecFlat` lift of `bloomOrIntoFn_spec`) and the registry rows follow separately. Landing alone because this is self-contained pure Lean with no separation logic in it.

## The mismatch

`bloom_or_into` ORs two materialised 256-byte blooms. `logs_bloom` (`Fork.lean:101`) never does that — it is a `foldl` of **bit-sets** into one accumulator, and `add_to_bloom` (`:89`) writes three bits with `List.set`. There were **zero theorems about either function** in the tree, so the guest's pointwise OR had no reference counterpart at all and the row could only have been `noCounterpart`.

## ⭐ Where the difficulty is, and how it dissolves

The three bit indices derived from one entry can **collide into the same byte**, so the three `List.set`s are not a disjoint update — the proof cannot go "each set touches a different index". I had budgeted an eight-case collision analysis for this.

It isn't needed. The right formulation is a single distribution lemma:

```lean
setOr (bloomOr b m) i v = bloomOr b (setOr m i v)
```

whose written-index case is **`Nat.lor` associativity** — so a second write to the same byte *re-associates* instead of clobbering. Collisions are absorbed rather than avoided. Everything after that is an induction over `(index, value)` pairs generalising the base:

```lean
ps.foldl setOr b = bloomOr b (ps.foldl setOr zeroBloom)
```

then `add_to_bloom_eq_bloomOr`, `logs_bloom_eq_foldl`, and the payoff:

```lean
logs_bloom (l₁ ++ l₂) = bloomOr (logs_bloom l₁) (logs_bloom l₂)
```

## ⚠️ Scope

Per `docs/leaf-routine-targets.md:46` — the **fold only**. The per-log index derivation (`keccak256` and the 11-bit extraction) is explicitly out of scope for #11348 and is treated here as an opaque function of the entry. That is also why `pair_bounds` is three lines: it is stated on the *shape* (`(2047 - x) / 8`, `1 <<< (7 - _ % 8)`) rather than on the hash expression.

## Non-vacuity

Seven `#guard` pins, and they evaluate the **real** `keccak256` derivation rather than a stub — deliberately, because a decomposition that held only for the all-zero mask would sail past a stubbed pin. Two of them check that the blooms being combined actually *differ*, so the split is not vacuous, and one checks commutativity (OR is, so `logs_bloom` is order-insensitive across segments — worth pinning since it is not obvious from the fold).

## Mathlib-free

`SpecRef` has to stay out of the Mathlib closure so the correspondence harness can gate per-PR; `check-correspondence-deps` reports **20 modules against a budget of 80, 0 violations**. Worth noting for anyone extending this file: `set` is a Mathlib tactic and cannot be used here — that cost me a build cycle.

## Verification

`port-check.sh` **PASS**. `lake build` clean (4808 jobs, 0 warnings); `check-axioms` 126 witnesses classical-only; `check-correspondence-deps`, layering, forbidden-tactics, file-size, naming all pass. No registry rows touched — those land with the machine half. Submodule absent from the commit (checked).

Refs #11348.

### DIFF
--- a/EvmAsm/Stateless/SpecRef.lean
+++ b/EvmAsm/Stateless/SpecRef.lean
@@ -60,6 +60,7 @@ import EvmAsm.Stateless.SpecRef.InstructionsCore
 import EvmAsm.Stateless.SpecRef.InstructionsEnv
 import EvmAsm.Stateless.SpecRef.Interpreter
 import EvmAsm.Stateless.SpecRef.Fork
+import EvmAsm.Stateless.SpecRef.BloomAlgebra
 import EvmAsm.Stateless.SpecRef.ElExecute
 import EvmAsm.Stateless.SpecRef.Precompiles
 import EvmAsm.Stateless.SpecRef.PrecompilesHash
--- a/EvmAsm/Stateless/SpecRef/BloomAlgebra.lean
+++ b/EvmAsm/Stateless/SpecRef/BloomAlgebra.lean
@@ -1,0 +1,311 @@
+/-
+  EvmAsm.Stateless.SpecRef.BloomAlgebra
+
+  GH #11348 — the algebra `logs_bloom` needs in order to have a counterpart for
+  `bloom_or_into` at all.
+
+  THE MISMATCH. The guest routine ORs two materialised 256-byte blooms. The reference
+  `logs_bloom` (`Fork.lean:101`) never does that: it is a `foldl` of *bit-sets* into a
+  single accumulator, and `add_to_bloom` (`:89`) writes three bits with `List.set`.
+  Nothing in the tree relates the two — there are **zero** theorems about either function.
+  Without the decomposition below the guest's pointwise OR has no reference counterpart
+  and the row could only be `noCounterpart`.
+
+  ⭐ THE LOAD-BEARING FACT, and where the difficulty actually is. `add_to_bloom` is
+  *`bloomOr` with a fixed mask*:
+
+      add_to_bloom b e = bloomOr b (add_to_bloom zeroBloom e)
+
+  That looks obvious and is not, because the three bit indices derived from one entry can
+  **collide into the same byte**. So the three `List.set`s are *not* a disjoint update, and
+  the proof cannot proceed by "each set touches a different index". It goes through
+  `setOr_getD` instead: one OR-write is characterised pointwise, and the three-step fold is
+  then unfolded with a case split at each step on whether that step's index is the one
+  being read. Collisions are handled by `Nat.lor` associativity/idempotence rather than
+  avoided.
+
+  ⚠️ SCOPE, per `docs/leaf-routine-targets.md:46`: this is the **fold** only. The per-log
+  index derivation — `keccak256` and the 11-bit extraction — is explicitly out of scope for
+  #11348 and is treated here as an opaque function of the entry.
+-/
+
+import EvmAsm.Stateless.SpecRef.Fork
+
+namespace EvmAsm.Stateless.SpecRef
+
+/-- The all-zero bloom: `logs_bloom`'s fold seed. -/
+def zeroBloom : Bytes := List.replicate 256 (0 : BitVec 8)
+
+/-- Pointwise OR of two 256-byte blooms — the operation the guest routine performs.
+    Written in the same `(List.range 256).map` shape as the guest's post so the two line
+    up syntactically. -/
+def bloomOr (a b : Bytes) : Bytes :=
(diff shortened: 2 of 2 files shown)


pirapira:
**Attribution: coord.** Reviewed against `origin/main`, **built locally**, axioms checked, and both of the docstring's
factual claims verified. **Labelled `merge!`.**

## Verified rather than inferred

- `lake build EvmAsm.Stateless.SpecRef.BloomAlgebra` → **Built** (29 jobs), including the `#guard` pins, which therefore
  evaluated.
- `#print axioms`: `logs_bloom_append`, `add_to_bloom_eq_bloomOr` and `foldl_setOr_eq_bloomOr` each depend on
  `[propext, Quot.sound]` — classical only, no `sorryAx`, no `ofReduceBool`/`trustCompiler`, and nothing in the diff
  matches `bv_decide`/`native_decide`.
- **"Zero theorems about either function"** — confirmed: no `theorem` mentioning `logs_bloom` or `add_to_bloom` exists
  anywhere under `EvmAsm/` on `main`. The motivation is accurate, not rhetorical.
- **The scope claim** — `docs/leaf-routine-targets.md` row 6 for `bloom_or_into` reads *"the OR-accumulation step of
  `logs_bloom` (`Fork.lean:101`) — **the fold, not the per-log index derivation**"*. Treating `keccak256` and the 11-bit
  extraction as opaque is exactly the sanctioned scope, and the PR says so at the definition rather than burying it.

## The hard part is correctly identified and correctly handled

The docstring's ⭐ claim — that `add_to_bloom b e = bloomOr b (add_to_bloom zeroBloom e)` "looks obvious and is not" — is
the real content. Three bit indices derived from one entry **can collide into the same byte**, so the three `List.set`s are
not a disjoint update and the proof cannot go by "each set touches a different index".

The chosen route handles collisions rather than excluding them: `setOr_getD` characterises one OR-write pointwise, and in
`foldl_setOr_eq_bloomOr`'s `k = p.1` branch the collision resolves through `ofNat_or_eq` then **`BitVec.or_assoc`** — a
second write to the same byte *re-associates* instead of clobbering. That is the right proof for the right reason, and it
generalises: the theorem is stated for an arbitrary run of `(index, value)` pairs, so the three-writes-per-entry structure
never has to be special-cased.

`ofNat_or_eq` earns its place too — it bridges the reference's `Nat`-level `|||` inside `BitVec.ofNat 8` to the guest's
`BitVec` OR, with `Nat.or_lt_two_pow` discharging the bound. And `zero_or_byte` exists because core's `BitVec.zero_or` is
stated for `0#w` and does not match the `OfNat` form the goals actually carry — a helper existing for a stated reason
rather than by accident.

## The extraction's faithfulness is proved, not asserted

`bloomPairs` reverse-engineers the pairs `add_to_bloom` writes, which is the step where a wrong transcription would
silently make everything else vacuous. `add_to_bloom_eq_foldl` closes it with **`rfl`** after `List.foldl_map` — so the
pair list is *definitionally* what `add_to_bloom` does, not a hand-written correspondence someone has to trust. That is
the strongest available form and it is worth noting because the surrounding statements all rest on it.

`pair_bounds` likewise argues from the **shape** (`(2047 - x) / 8 < 256`, `1 <<< k` with `k ≤ 7`) rather than from the hash
expression, which is what keeps the bound independent of the out-of-scope derivation.

## The pins are non-vacuous, and one of them is doing real work

`logs_bloom` runs `keccak256`, so the `#guard`s evaluate the genuine derivation. The pin that matters is
`#guard logs_bloom [L1] != logs_bloom [L2]`: without it, the commutativity pin
(`logs_bloom [L1, L2] == logs_bloom [L2, L1]`) would also pass if both blooms were identically zero, and the append pin
would hold for an all-zero mask. Pinning that the two blooms **differ** is what makes the others load-bearing. That is the
distinction between a test and a pin, and this PR draws it.

## Payoff

`logs_bloom_append` is the statement `bloom_or_into` needs: ORing two materialised 256-byte blooms *is* `logs_bloom` of the
concatenated segments. Without it the guest's pointwise OR has no reference counterpart and row 6 could only be
`noCounterpart`. `bloomOr` is deliberately written in the same `(List.range 256).map` shape as the guest's post so the two
line up syntactically — a small thing that saves the consumer a rewrite.

No `EvmAsm/Codegen/**` file touched, so no build, pin, or fixture implications; the only other change is the one `import`
line in `SpecRef.lean`.

PR #11418, opened 2026-08-04
docs(#11351): six of seven chain_validate_* have triples, not seven

dhsorens:
One-paragraph correction to `docs/leaf-routine-targets.md`.

## The claim, and the measurement

The table says row 9 (`header_extract_number`) is the representative of a family including *"the seven `chain_validate_*` routines"*, implying nine mechanical forks once its pattern lands.

Measured on `origin/main`: there are **seven** `chain_validate_*` symbols in `GuestAddrs.lean` but only **six** carry a whole-routine triple.

| routine | triple |
|---|---|
| `chain_validate_blob_gas_used_multiple` | `ChainValidateBlobGasMultipleLoopClose.lean:194` |
| `chain_validate_blob_gas_used_under_max` | `ChainValidateBlobGasUnderMaxLoopClose.lean:192` |
| `chain_validate_consecutive_numbers` | `ChainValidateConsecutiveNumbersLoopClose.lean:877` |
| `chain_validate_extra_data_length` | `ChainValidateExtraDataLengthLoopClose.lean:175` |
| `chain_validate_gas_used_under_limit` | `ChainValidateGasUsedUnderLimitLoopClose.lean:1283` |
| `chain_validate_increasing_timestamps` | `ChainValidateIncreasingTimestampsLoopClose.lean:873` |
| **`chain_validate_post_merge_full`** | **none** — `ChainValidatePostMerge.lean` has only the string↔`Program` byte-identity theorem (`:608`) |

## Why it is worth a PR rather than a silent edit

That count is the stated reason row 9 is the family representative and is ranked for leverage — so it feeds a prioritisation decision, not just prose. With `header_extract_logs_bloom` and `header_validate_extra_data_length`, the pattern unlocks **eight** forks, and the odd one out is an **unproven routine** rather than a fork awaiting a pattern. Grouping it with the others makes an unproven routine look like pending mechanical work.

Full table with file:line is on #11351, alongside the sizing for row 9 itself.

Docs only — no Lean touched. `check-drift.sh` passes.

Refs #11351.

### DIFF
--- a/docs/leaf-routine-targets.md
+++ b/docs/leaf-routine-targets.md
@@ -50,9 +50,15 @@ is exactly the "leaf means leaf in the TRANSITIVE call graph" trap.
 | 10 | `bgv_u32le` (leaf; 8 fixture in-edges; caller `EvmAsm/Codegen/Programs/BlockVerdictStateRoot.lean:681`) | the fixed-width LE reads of `deserialize_stateless_input` (`EvmAsm/Stateless/SpecRef/Guest.lean:29`) — the u32 accessor those reads reduce to | guest input blob (witness section offsets) | `bytesRegion`; proving this one leaf discharges a step in every one of its 8 callers |
 
 Row 9 is the representative of a **family**: `header_extract_logs_bloom`,
-`header_validate_extra_data_length`, and the seven `chain_validate_*` routines have the same
-shape (verified RLP callees only, one `_decode_header` field or one header rule each). Once
-row 9's pattern exists, the siblings are mechanical forks.
+`header_validate_extra_data_length`, and **six of the seven** `chain_validate_*` routines
+have the same shape (verified RLP callees only, one `_decode_header` field or one header
+rule each). Once row 9's pattern exists, those **eight** siblings are mechanical forks.
+
+⚠️ The seventh, `chain_validate_post_merge_full` (`GuestAddrs.lean:453`), is **not** a fork:
+it has no whole-routine triple at all — `ChainValidatePostMerge.lean` carries only the
+string↔`Program` byte-identity theorem (`:608`). It is an unproven routine, and counting it
+with the forks overstates what row 9's pattern buys. Measured on `origin/main`; an earlier
+revision of this paragraph said "seven".
 
 ## Runners-up (not in the ten, recorded so the cut is visible)
 


pirapira:
**Attribution: coord.** Reviewed against `origin/main`. **Labelled `merge!`.** Every claim checked — for a
factual-correction PR that is the entire review.

## Verified

| claim | check |
|---|---|
| there are **seven** `chain_validate_*` routines | `GuestAddrs.lean` has exactly 7: `blob_gas_used_multiple`, `blob_gas_used_under_max`, `consecutive_numbers`, `extra_data_length`, `gas_used_under_limit`, `increasing_timestamps`, `post_merge_full` |
| **six** carry spec-style whole-routine theorems | confirmed; only `post_merge_full` has none |
| `ChainValidatePostMerge.lean:608` carries **only** the string↔`Program` byte-identity theorem | confirmed: `chainValidatePostMergeFullFunction_eq_prog … := rfl`, plus two `#guard`s (`startsWith`, `prog.length = 149`). No triple. |
| the fork count becomes **eight** | `header_extract_logs_bloom` + `header_validate_extra_data_length` + 6 = 8 ✓ (it would have been 9 on the old wording) |

## Why this is worth a PR rather than a passing remark

The corrected sentence is load-bearing for planning. "Once row 9's pattern exists, the siblings are mechanical forks"
is a **cost estimate**, and it was over by one routine in the worst possible way: `chain_validate_post_merge_full` is not
a cheap fork of a proved sibling, it is an **unproven routine** with no whole-routine triple at all. Someone budgeting
from the old wording would have planned nine mechanical forks and discovered mid-flight that one of them needs the
triple written from scratch.

The distinction the PR draws — a **byte-identity** theorem (`string = emitProgramR prog relocs`, by `rfl`) is *not* a
whole-routine triple — is exactly the one that makes over-counting easy. Both are theorems, both sit in the routine's
file, and only one says anything about behaviour. `:608` proves the emitted string matches the `Program` value; it says
nothing about what the routine computes. Grep for `theorem` in these files and you get a hit either way, which is
presumably how "seven" arose.

## Small things done right

- **The correction states its own provenance**: *"Measured on `origin/main`; an earlier revision of this paragraph said
  seven."* A doc that records what it used to say wrongly is far more useful than one silently amended — the next reader
  can tell whether their memory of "seven" came from here.
- It cites `GuestAddrs.lean:453` for the routine and `:608` for the theorem, so both halves are checkable in one step
  each. That is what let me verify this in three commands.
- Scope is one paragraph in one file; no `EvmAsm/**` change, so no build, pin, or fixture implications.

PR #11420, opened 2026-08-04
fix(codegen): retire BAL tuple skip-list check 42 (#11245)

### LINKED ISSUE #11245: convergence: verify the BAL by hash (wire bal_serializer_verify), retiring the field-by-field stand-ins

pirapira:
## Summary

Retire BAL check 42 (`bal_all_accounts_tuple_sequences_consistent_skip_list`) and its exclusive callee chain from the guest. Survivor is `bal_serializer_verify` hash equality (rebuild vs supplied).

Closes #11245. Depends on #10646 OUTCOME 1 (recipient multi-bai rebuild is full per-tx sequence, not final-only).

## Gate-equality argument (full)

1. **Old gate (42):** compared *supplied* BAL slot-tuple sequences against the execution log (forward match over accounts).
2. **Survivor gate:** `bal_serializer_verify` rebuilds BAL from the execution log (`bal_builder_storage_changes` walks `tx_storage_writes` per BAI) and compares rebuilt digest to supplied.
3. **Rebuild carries recipient per-tx sequence:** measured on discriminating fixtures 00337 (`bal_cross_tx_storage_write` / tx2_reverts_to_zero) and 00363 (`bal_multiple_storage_writes_same_slot`) — guest builder rows show full multi-bai sequences matching pinned execution-specs BAL (#10646).
4. **No path copies supplied rows into the builder** (verified under #10681 storage half): rebuild is independent of the supplied BAL bytes being checked.
5. **Same property, same gate, same bytes** for valid blocks: named A/B full-match below.

## Q1 — what happens to a former code-42 rejection?

**Rejection survives; diagnostic coarsens.**

A block that used to fail with `bv_fail_code = 42` (tuple-sequence mismatch at `.Lbv_bal_tuple_fail`) now fails at the terminal hash seam in `BlockVerdictReceiptsTail`:

- **60** = genuine BAL digest mismatch (`rebuilt ≠ supplied`)
- **61** = rebuild/serializer failure

So an invalid tuple sequence that made supplied ≠ rebuild still rejects. The specific "which account/slot sequence diverged" signal is lost; 60/61 are the least-diagnostic codes by design (notes: `bv_fail` is a rejection tail, not root-cause). Acceptable under #10646 closure (hash owns the property).

Class-A baseline line removed had `return_status_tested=yes` — that is exactly this coarsening being acknowledged by the ratchet, not a silent drop of a reject path.

## Q2 — `.Lbv_bal_tuple_fail`

**Nothing jumped to it** after the cut (only definition remained). Removed entirely in `e12f1e829` so no residue (same class of cleanup #11415 does for the storage subtree). Incompatible readings resolved: dead label gone; rejection path is hash 60/61 only.

## Cut inventory

| Item | Action |
|------|--------|
| jal `bal_all_accounts_tuple_sequences_consistent_skip_list` from `.Lbv_mtx_storage` | → `j .Lbv_after_tx_gas_precharge` |
| Guest closures (V2): `balAllAccountsTupleSequencesConsistent`, `accountTupleSequencesConsistent`, `slotTupleSequencesMatch`, `balSlotTupleSequence`, `execLogSlotTuples`, `systemUserExecLogSlotTuples` | unlinked |
| DataSectionTail: batsc_* + account/balSlot/execLog tuple data | removed |
| `.Lbv_bal_tuple_fail` (code 42) | removed (was dead) |
| `bv_mtx_skip_list` | **KEPT** (B1/B2 consumers) |
| Probe units / Function strings | kept probe-only |
| matches/covers/allaccounts_storage | **not** this PR (open on #11415) |

## Layout / ratchet

- `textSizeBytes` 0x5ad04 → 0x59f14 (cut) → 0x59f00 (sink drop)
- Class-A baseline: deliberate shrink **one** line (`.Lbv_mtx_storage` load of tuple skip_list, `return_status_tested=yes`)
- Gates: region-map CLEAN, class-a OK 22 paths, asm-to-program CLEAN, GuestImage rebuild OK

## Evidence (deletion / FA-adjacent — kept cheap)

**Named discriminating fixtures (spike A/B, full byte match both legs):**

| Fixture | base | cand |
|---------|------|------|
| bal_cross_tx_storage_write (2 cases incl. 00337 reverts) | OK×2 | OK×2 |
| bal_multiple_storage_writes_same_slot (00363) | OK×1 | OK×1 |

ELFs distinct: base `9c88d921…` (main, still has 42) vs cand `1ad55b05…` (pre-sink-drop); sink-drop is −0x14 B label-only, no control-flow change on accept path.

r200 random seed 4242 still running as optional bulk; named fixtures are the property witnesses.

## Test plan

- [x] check-region-map
- [x] check-bal-class-a-ratchet (22 paths, deliberate one-line shrink)
- [x] named A/B spike full-match on 00337/00363 family
- [ ] CI build green
- [ ] optional r200 when complete


pirapira:
**Attribution: coord.** Reviewed against `origin/main` at the rebased tip `5bdba97ba`. **Labelled `merge!`.** Closes the
tuple half of #11245.

## Both questions answered, and I checked the answers

**Q1 — the code-42 rejection survives.** It now arrives as hash `bv_fail` **60** (digest mismatch) or **61** (rebuild
failure). So the rejection is preserved and the **diagnostic coarsens**: a tuple-specific failure becomes a generic digest
failure. That is the right trade to make explicitly rather than silently, and it is worth knowing for the next person
debugging a rejected block who wonders where 42 went.

**Q2 — the dead sink is gone.** `.Lbv_bal_tuple_fail` has no live `jal` after the cut and was removed. The one remaining
occurrence is a **comment** at `BlockVerdictReceiptsTail.lean:249` recording the removal and where the rejection went —
which is exactly what I wanted, not residue.

## Three things I checked that looked like discrepancies and are not

Recording them because each would otherwise be re-investigated:

1. `bal_all_accounts_tuple_sequences_consistent_skip_list` still appears in `BlockVerdictMtxRuntime.lean:573` and
   `BlockVerdictMtxTail.lean:233` — both **comments**, not calls. Also present in `RegistryNamesTail`/`RegistryTail` (name
   tables) and in its own module, which is source-retained-but-unlinked, the same conservative split #11415 used.
2. The Class-A baseline shrank by **3** lines, not the 1 reported earlier. Only one is this PR's: the
   `.Lbv_mtx_storage` tuple path with `return_status_tested=yes`. The other two are `.Ldtrc_auth_exec_ready` stores for
   `runtime_current_bal`, already retired by #11400/#11415 landing in main. **Please state that attribution in the body** —
   a reader seeing three removed Class-A paths would otherwise credit them all to this cut.
3. `.text 0x59098` after the rebase differs from the pre-rebase figure, as it must, since #11415's 3056-byte subtree
   removal landed in between.

## The gate-equality argument holds

Check 42 compared supplied sequences against the execution log; the survivor `bal_serializer_verify` compares **rebuilt**
against supplied by hash. The rebuild carries the recipient's per-transaction sequence — measured on **00337** (slot0
`bai=1 val=0xabcd`, `bai=2 val=0`) and **00363** (slot1 `bai=1/2/3`) — and no path copies supplied rows into the builder,
verified when #10681 closed. Same property, same gate (`bv_bal_shadow_ready`, sole writer
`BlockVerdictFunction.lean:230`), same bytes.

Note this is the **third** application of that test tonight and it has given three different answers: retire (37/44/45,
#11392), retire (42, here), and **do not retire** (`bal_code_preimages_valid`, #11183, whose property the hash cannot see
at all). Applied per check, not per family — which is why it is worth stating in the body each time rather than citing a
precedent.

## Small follow-up, not a blocker

`BlockVerdictMtxRuntime.lean:573`'s comment now describes `exec_log_txindex` as *"the a4 base of
`bal_all_accounts_tuple_sequences_consistent_skip_list`"* — a retired routine. That is the #11386 class: a comment
describing a mechanism that no longer exists. Worth fixing in passing here or noting for the residual sweep; comments
outlive the code they describe, and this one will read as evidence the check still exists.

Rebase was legitimate: this PR was **not** yet labelled, so hands-off had not begun, and resolving the `BlockVerdictV2`
overlap (keeping both #11415's storage unlink and this tuple unlink) is a semantic decision better made by the author than
by a merge actor.

pirapira:

pirapira:
Landed via #11430. Verified: commit `5bdba97ba` is an ancestor of `main` as of `c58e3f87b` — closed rather than dropped.

PR #11421, opened 2026-08-04
proof(#11344): whole-routine triple for bytes_to_nibbles, tied to keyToNibbles

### LINKED ISSUE #11344: proof: triple for `bytes_to_nibbles` against `keyToNibbles`

dhsorens:
Closes #11344 — row 1 of `docs/leaf-routine-targets.md`, 10 fixture in-edges.

## What existed, what was missing

`bytesToNibblesFn_spec` already proved the structured SAsm contract, counted loop and all, discharged by `vcgen`. Two things were missing.

**1. The reference tie.** `bytesToNibblesBytes` is a *local* accumulator (`nibblePrefix`, built by repeated `++ nibblePair …`), and `nibblePair` is phrased with `BitVec.truncate`/`signExtend12` because that is how the emitted `SRLI`/`ANDI` behave. `keyToNibbles` is a one-line `List.flatMap`. Three clauses differ, all **proved**:

| clause | lemma |
|---|---|
| `truncate 8 (b.zeroExtend 64 >>> 4)` vs `ofNat 8 (b.toNat >>> 4)` | `highNibble_eq` |
| `&&& signExtend12 15` vs `&&& 0x0F` | `lowNibble_eq` |
| accumulator vs `flatMap` | `nibblePrefix_eq_keyToNibbles_take` |

⭐ The third is where the difficulty is: **`nibblePrefix` appends on the right while `flatMap` builds on the left**, so the induction step has to expose the *last* element (`List.take_add_one`) rather than the first. That mismatch is the whole content; the two nibble lemmas are three lines each.

**2. The flat triple** at the linked guest address, derived by `Fn.retSpecFlat` — the loop reasoning stays in the SAsm proof. Unlike `bgv_u32le` this routine has a non-empty **writable** region, so the shape follows `bnqZeroFlat_spec` — except that **three** argument registers are pinned in the precondition rather than one, needing a three-way `exposedRegs_split3`. `x11`/`x12` are outside the scratch list, so they are handed back explicitly as `regOwn` or the caller loses them.

## ⭐ A caveat that is not general, and matters for #11348

`retSpecFlat` applies here **only because** `bytesToNibblesFn`'s `pre`, `post` *and* loop invariant all carry `A = empAssertion`.

`bloomOrIntoFn` does not — it declares `fun rf ws _ =>` throughout — and consequently **neither** `retSpecFlat` (needs `A' = empAssertion`) **nor** `retSpecFlatAmbient` (needs `A' = A`) has a provable side condition there, since a post that ignores the ambient holds for every ambient. That is reported on #11348 with three options; I mention it here because this PR is the contrast case that makes the diagnosis concrete.

## Scope

Also covers the nibble-expansion half of `compact_to_nibbles`, which calls `keyToNibbles` — no separate lemma needed. **Not** the flag decode.

## Pins

Nibble **order** (`0xab → [0x0a, 0x0b]`, so a swap fails), both extremes, `len < length` so trailing bytes are ignored, and empty — plus **two showing different inputs give different outputs**. Without those last two the agreement pins would pass vacuously; that is your point from #11416, applied here.

## Registries

Routines 25→26 rows, `.proven` 17→18, symbols 17→18. Correspondence 23→24, new `mpt` family with its own `countFamily` theorem, agrees 15→16, **`.ported` 2→3** — using the rung from #11401, with the port-fidelity clause table the rung requires (the three clauses above). Witnesses 126→127.

## Verification

`port-check.sh` **PASS**. `lake build` clean (4808 jobs, 0 warnings); `check-axioms` 127 witnesses classical-only; `check-registry-crosscheck` 18 symbols × 24 rows; progress regenerated. Submodule absent from the commit (checked).

Closes #11344.


pirapira:
**Attribution: coord.** Reviewed against `origin/main`, **built locally**, axioms and registry arithmetic checked.
**Labelled `merge!`.**

## Verified

- `lake build EvmAsm.Codegen.Programs.BytesToNibblesBridge` → **Built** (1016 jobs).
- `#print axioms`: `bytesToNibblesFlat_spec` and `bytesToNibblesBytes_eq_keyToNibbles` → `[propext, Classical.choice,
  Quot.sound]`. Classical only.
- `scripts/check-forbidden-tactics.sh` → **OK**. The one `bv_decide` string in the diff is a backticked prose mention
  (*"Per-bit, no `bv_decide`"*), which the convention permits.
- Registry arithmetic: `registry_size` 23 → 24, new `mpt_rows : countFamily "mpt" = 1`, `countVerdict` totals updated.

## ⚠️ THE ODD-NIBBLE CASE — what this does and does not cover

**`bytes_to_nibbles` cannot produce an odd nibble count, so there is no odd case *in this routine*.** The output length
is `2 * len` by construction and it is pinned in four places in the spec: `horig : orig.length = 2 * len`,
`hdstOver : dst.toNat + 2 * len < 2 ^ 64`, the returned `x10 ↦ᵣ BitVec.ofNat 64 (2 * len)`, and the written region
`keyToNibbles (srcBytes.take len)` where `keyToNibbles` is a `flatMap` emitting exactly two nibbles per byte. An odd
result is unrepresentable.

**But odd nibble paths certainly exist in the MPT**, and they come from the *other* half — `compact_to_nibbles`
(`SpecRef/IncrementalMpt.lean:76-84`):

```lean
let first_nibble := c0.toNat >>> 4
let odd  := first_nibble &&& 0x01 ≠ 0
let tail := keyToNibbles rest
let nibbles := if odd then BitVec.ofNat 8 (c0.toNat &&& 0x0F) :: tail else tail
```

So oddness arises **only** from the hex-prefix flag byte: when the odd flag is set, `c0`'s *low* nibble is prepended to
the even expansion, giving `1 + 2*(n-1)`. The expansion itself is always even; the parity comes from the flag.

**This PR explicitly excludes that.** Its scope note says #11344 covers *"the nibble-expansion half of
`compact_to_nibbles` … that half only, **not the flag decode**"* — and the odd/even decision **is** the flag decode. The
registry row's `reference` field is worded consistently (*"the nibble-expansion half"*), so this is correctly scoped
rather than overclaimed. The docstring's justification is right too: the expansion half is *literally* `keyToNibbles
rest`, so no separate lemma is needed for it.

**What therefore remains unproven, and should not be mistaken for covered** by anyone reading `compact_to_nibbles` in the
registry row:
1. the **odd-flag prepend** (`(c0 &&& 0x0F) :: tail`) — the only source of odd-length paths;
2. the **`is_leaf`** flag (`first_nibble &&& 0x02`);
3. the **empty-input rejection** — Python indexes `compact[0]` unconditionally, so `[]` is an `IndexError` and the port
   throws `witnessNodeMalformed`.

If the guest's compact-decode mishandles the odd flag — dropping the nibble, or taking the high nibble instead of the
low — **this triple would not catch it**, because the proof stops at the even expansion. Worth a follow-up row rather
than an assumption.

## The proof itself

The docstring names the real difficulty accurately: `nibblePrefix` grows on the **right** (`prefix ++ pair i`) while
`flatMap` grows on the **left**, so the induction step needs `List.take_succ`/`take_add_one` to expose the *last* element
rather than the first. That is the content; `highNibble_eq`/`lowNibble_eq` are the mechanical half, bridging
`BitVec.truncate` of a widened byte and `signExtend12 15` to the reference's `>>> 4` and `&&& 0x0F` — per-bit, no
`bv_decide`, with `byte_mod` discharging the `% 2 ^ 64` that `toNat_setWidth` introduces.

The `.ported` clause table is the right discipline: it enumerates **three** differences (truncate-vs-`ofNat`, `signExtend12
15`-vs-`0x0F`, and the accumulator/`flatMap` shape), names the lemma discharging each, states *"nothing else differs"*, and
justifies `.ported` over `.bridged` by the absence of an executable differential for the MPT/witness family. It also
states the full domain — side conditions are ABI-only, with `len ≤ srcBytes.length` being the ABI contract rather than an
input gate — which is what distinguishes `.agrees` from `.domainRestricted` here.

pirapira:

pirapira:
Landed via #11430 (cross-author batch). You rebased mid-assembly, so the batch carried the pre-rebase SHA `c65c3c273` rather than `c71d87cbf` — I verified by **content** instead: the `bytes_to_nibbles` rows are present in `main`'s `EvmAsm/Progress/Routines.lean` and `Correspondence.lean`, and `git diff` between the two heads over those files was empty before merging. Your work is in; closed rather than dropped.

PR #11424, opened 2026-08-04
codegen: remove EIP-8037 state fallback

pirapira:
Delete the unreachable EIP-8037 storage-only state fallback and preserve the existing rejection by routing an exact-prefix miss to the existing state-fail path. The fallback computed 97920 times the storage-set count while the primary sums all prior state gas, so the two were never equal. The 77-row 120-prefix differential established that. The fallback is unreachable on any accepting path because the undercount routes bv_dispatch_runtime_status nonzero and the receipts tail rejects with fail64. Mechanistically, bvgr_runtime_count remains zero only on bail labels (unsupported, EOA, capacity, dispatch-miss, parse error, and loop bail); the two accepting labels write bv_tx_count, so non-independence is not a cause. The exact-prefix miss now takes the existing state-fail path, so the rejection is preserved. The emitted .text shrinks from 0x5af0c to 0x5ab7c, minus 912 bytes. The Class-A ratchet checker and baseline are absent from this checkout because 11409 has not merged, which is why no baseline was hand-written. Verification: lake build; python3 scripts/asm_to_program.py check-all; EIP-8037 fixture check-file gates; scripts/check-region-map.sh; python3 scripts/gen-region-map-link-pins.py --check; git diff --check.


pirapira:
**Attribution: coord.** Reviewed against `origin/main`. **Labelled `merge!`.** Closes #11407.

## Verified

`eip8037_state_used_before_tx` is **fully removed from `EvmAsm/Codegen/**`** — Program, reloc table, symbol, fixture and
MANIFEST row together, with the gate's subsequent branch displacements rebased. `.text 0x5af0c → 0x5ab7c` = **−912 bytes**
(888-byte routine + 24-byte transport), `.data`/`.bss` unchanged. Fixture and MANIFEST row moved **together**, which is the
discipline #11400 was handed back for.

Note the branch is stacked on #11405 (its `0x5af0c` starting point is #11405's post-removal figure), so the diff carries
that work beneath this commit — same commits, collapsing when #11405 lands.

## Why deleting rather than re-sourcing was right

This started as "re-source the fallback from a tracked counter" and ended as "delete it", and each turn came from a
measurement:

1. A **PC probe** found the fallback entry `0x8002d3b0` unreached across 00326-00338, 00509, 00513, 00612, 01086/01087 —
   nothing in the corpus reaches it.
2. A **differential** on 200 rows (162 meeting the exact gates) found **77 rows / 120 prefixes** disagreeing between the
   fallback and the primary. Decoding by ÷1530 named the cause: `183600` vs `0` is **120 bytes of creation charge** that a
   storage-set-only definition cannot see. The fallback computed `97,920 × storage-set count`; the primary sums *all* prior
   state gas. **Never equivalent.**
3. An **accept/reject trace** showed the undercount routes set `bv_dispatch_runtime_status` nonzero and the receipts tail
   rejects with fail64 — so the FA-lax arithmetic can never be the difference between accept and reject.

Building the counter would have preserved a quantity **wrong by construction**. Step 3 is what made deletion clearly safe
rather than merely tidy, and it is a genuine impossibility argument rather than a failed search.

## Failure direction

The fallback **under-reported** prior state gas whenever a prior transaction carried a creation charge, making the
remaining-gas figure too large and the `tx.gas` rejection **looser** — the FA direction. Removing it and taking the
existing state-fail path is therefore **strictly safer**: the change can only make the guest reject more, never accept
more. Under the screening question — *could this cause an invalid block to be accepted?* — the answer is no, so this lands
on the light bar.

To be explicit, since I warned against fail-closed earlier and this superficially resembles it: that rule guards against
discarding information **we could compute**. Here the information provably does not exist on those paths — no replay means
no settled `evm_state_gas_used` to sum — so rejecting is the only sound option available.

## The method note worth keeping

**No sweep of any size could have found this.** The path is unreachable in the corpus, so a clean random-200 is silent by
construction; only computing *both* quantities on the rows where the primary succeeds could compare the two definitions. A
differential, not a sample.

The ratchet caveat in the body is correct and honest: the checker and baseline are absent from the checkout because #11409
has not merged, so no baseline was hand-written. That is the right call — a hand-authored baseline would be a fabricated
measurement.

pirapira:
Needs a rebase onto current `main` — the conflict is in **emitted program bytes**, which is the one category I won't resolve mechanically.

**Conflicts** (beyond the usual generated layout files):

`EvmAsm/Codegen/Programs/BlockVerdictGasGate.lean` — 3 hunks, both sides carrying real instructions:

| hunk | `main` | this PR |
|---|---|---|
| 1 | 12 instrs, opens `.BEQ .x10 .x0 (28)` | 6 instrs, opens `.BNE .x10 .x0 (56)` |
| 2 | 4 reloc rows incl. `(532, .la .x13 "bsg_prior_state")` | 2 reloc rows incl. `(529, .la .x5 "bsg_tx_gas")` |
| 3 | `#guard eip8037TxGasGate_prog.length = 568` | `#guard … = 562` |

`scripts/asm-fixtures/eip8037TxGasGateFunction.s` — whole-body conflict (300 vs 453 lines), and the two sides don't even share register spelling (`addi sp, sp, -112` vs `addi x2, x2, -112`).

**Why I stopped:** both sides are live RISC-V for the same routine, with **different program lengths** (568 vs 562) and **different branch polarity** on the opening test. Picking either side would silently ship one version's control flow with the other's relocs/pins, and the `#guard` length is a kernel-checked pin that must match the emitted body exactly. `main` moved this routine under #11405 (EIP-8037 BAL transport retirement) while this PR rewrites the same gate for the state-counter change.

That is a genuine re-derivation of the emitted program, not conflict repair — it needs you to re-apply the state-counter change on top of the post-#11405 body so the reloc table, the `#guard` length and the fixture all agree.

For reference, the fixture's differing register spelling suggests the two bodies were generated by different emitter paths, which is worth a glance while rebasing.

Nothing else was wrong; the remaining conflicts were the generated layout files only. Rebase and I'll pick it up next pass with full local verification (`lake build` + all 20 gates).

pirapira:
Re-parking `complicated-merge` — thanks for the rebase (`bb783e260` → `b96d43ffa`), but it now conflicts again, on the **same two files**, because the target moved underneath you:

```
EvmAsm/Codegen/Programs/BlockVerdictGasGate.lean
scripts/asm-fixtures/eip8037TxGasGateFunction.s
```

**What changed:** **#11437** (`fix(11428): drop dead supplied-BAL transport into eip8037_tx_gas_gate`) landed in `main` at `8adb23eeb` about half an hour ago, and it rewrote `eip8037TxGasGateFunction.s` substantially (**751 lines changed**) plus retired the `eip8037StateUsedBeforeTxFunction` fixture and its MANIFEST row. So this PR's state-counter rework of the same gate is once again derived from a pre-#11437 body.

This is the same class of conflict as before — emitted RISC-V for the same routine on both sides, with the `#guard eip8037TxGasGate_prog.length` pin having to match the body exactly — so it's a re-derivation rather than conflict repair, and not something I should resolve on your behalf.

**Suggestion:** rebase onto `8adb23eeb` or later, since #11437 was the last thing touching that gate and nothing else in the queue is aimed at it. The rest of the EIP-8037 retirement work (#11405, #11437) has now landed, so the gate body should be stable from here.

For reference, everything else in this PR merged cleanly — the only other conflicts were the generated layout files.

PR #11425, opened 2026-08-04
proof: arity-N decodeItems composition (prereq for #11351)

dhsorens:
Infrastructure for #11351. **No registry rows touched, no issue closed** — consumers arrive with the header rows.

## The gap

`WalkDecodeBridge`'s composition helpers are hard-coded to **four** items (`decodeItems_four_of_decodeAux:116` and friends, written for the withdrawal 4-tuple). `_decode_header` walks **21 or 23**, so every header row needs the general form.

## ⭐ The design choice, which is why this is 45 lines rather than 200

The obvious generalisation indexes the offsets and quantifies `∀ i < items.length, decodeAux … (offs.getD i 0) = some (items.getD i _, …)`. That makes the statement and every obligation unreadable, and the arithmetic on `getD` defaults leaks everywhere.

Stating it **relationally** instead:

```lean
DecodeChain bytes off []             offEnd = (off = offEnd)
DecodeChain bytes off (item :: rest) offEnd =
  ∃ off', (∀ m, decodeAux (m+1) (bytes.drop off) = some (item, bytes.drop off'))
          ∧ DecodeChain bytes off' rest offEnd
```

lets **one** induction do the work, and reuses `decodeItems_cons_of_decodeAux` (`:87`) unchanged as the step — with `hne` derived from the link exactly as the arity-four proof does.

## Non-vacuity

An `example` reproduces `decodeItems_four_of_decodeAux`'s **exact statement** from the general lemma. That is the strongest available check that the chain predicate means what the four-item version means: a predicate that had drifted would fail there, whereas a hand-written pin might not.

The arity-four helpers are **left in place**. This only demonstrates they are derivable — retiring them is a follow-up if you want it, not a side effect of this PR.

## ⚠️ The restriction, and why it is worth recording here

Each link demands fuel-insensitivity (`∀ m`), which is sound only for **byte-string** items (`decodeAux_bytes_all_fuel_of_decode:273`). Header fields all are, so #11351 is covered.

It is the **same restriction** that makes #11341's nested-list case unprovable — a nested list's decode *is* fuel-sensitive, which is exactly why the `c3 c2 81 00` counterexample bites there and not here. Two issues, one underlying constraint; the docstring says so rather than leaving it to be rediscovered.

## Verification

`port-check.sh` **PASS**. `lake build` clean (4806 jobs, 0 warnings); `check-axioms` 126 witnesses classical-only; layering, file-size, no-warnings pass. Submodule absent from the commit.

Prereq for #11351.

### DIFF
--- a/EvmAsm/Rv64/RLP/WalkDecodeBridge.lean
+++ b/EvmAsm/Rv64/RLP/WalkDecodeBridge.lean
@@ -392,4 +392,78 @@ theorem decodeFully_shortList_four_chain (pfx : Byte) (payload r1 r2 r3 r4 : Lis
         item0 item1 item2 item3 k h0 h1 h2 h3 hend
   simp [decodeFully, hdec]
 
+/-! ## Arity-N composition (GH #11351)
+
+    The helpers above are hard-coded to **four** items, written for the withdrawal
+    4-tuple. `_decode_header` walks 21 or 23, so the header rows need the general form.
+
+    ⭐ The generalisation is stated **relationally** rather than by indexing the offsets
+    with `getD`: an indexed form makes both the statement and every proof obligation
+    unreadable, whereas a chain predicate lets one induction do the work and reuses
+    `decodeItems_cons_of_decodeAux` unchanged as the step.
+
+    ⚠️ Each link demands fuel-insensitivity (`∀ m`), which is sound only for
+    **byte-string** items (`decodeAux_bytes_all_fuel_of_decode`). Header fields all are.
+    This is the same restriction that makes the nested-list case of #11341 unprovable —
+    a nested list's decode *is* fuel-sensitive. -/
+
+/-- A chain of byte-string decodes: starting at `off`, each item advances the offset,
+    ending exactly at `offEnd`. -/
+def DecodeChain (bytes : List Byte) : Nat → List RLPItem → Nat → Prop
+  | off, [], offEnd => off = offEnd
+  | off, item :: rest, offEnd =>
+      ∃ off', (∀ m, decodeAux (m + 1) (bytes.drop off) = some (item, bytes.drop off'))
+        ∧ DecodeChain bytes off' rest offEnd
+
+/-- ⭐ **Arity-N `decodeItems` composition.** A decode chain ending at the end of input
+    composes to a full `decodeItems` run over exactly those items. Generalises
+    `decodeItems_four_of_decodeAux` to any length. -/
+theorem decodeItems_of_chain (bytes : List Byte) :
+    ∀ (items : List RLPItem) (off offEnd : Nat),
+      DecodeChain bytes off items offEnd → bytes.drop offEnd = [] →
+      ∀ k, decodeItems (k + items.length + 1) (bytes.drop off) = some (items, []) := by
+  intro items
+  induction items with
+  | nil =>
+    intro off offEnd hc hend k
+    subst hc
+    rw [hend]
+    rfl
+  | cons item rest ih =>
+    intro off offEnd hc hend k
+    obtain ⟨off', hitem, hrest⟩ := hc
+    have hne : bytes.drop off ≠ [] := by
+      intro hnil
+      have h0 := hitem 0
+      rw [hnil, decodeAux_nil] at h0
+      simp at h0
+    have hIH := ih off' offEnd hrest hend k
+    have hcomp := decodeItems_cons_of_decodeAux bytes off off' item rest []
+      (k + rest.length) hne (hitem (k + rest.length)) hIH
+    have harith : k + (item :: rest).length + 1 = k + rest.length + 2 := by
+      simp [List.length_cons]
+      omega
+    rw [harith]
+    exact hcomp
+
+/-- **Non-vacuity: the existing arity-four helper is an instance.** Reproducing
+    `decodeItems_four_of_decodeAux`'s exact statement from the general lemma is the
+    strongest available check that the generalisation is faithful — a chain predicate
+    that had drifted from what the four-item version means would fail here.
+
+    The arity-four helpers are left in place; this only demonstrates they are now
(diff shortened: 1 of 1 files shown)


pirapira:
**Attribution: coord.** Reviewed against `origin/main`, **built locally**, axioms checked. **Labelled `merge!`** — the
mathematics is sound and the one issue below is a prose figure, not a proof obligation.

## Verified

- `lake build EvmAsm.Rv64.RLP.WalkDecodeBridge` → **Built** (905 jobs), so the `example` at the end elaborated too.
- `#print axioms decodeItems_of_chain` → `[propext, Quot.sound]`. Classical only; no `sorry`, no `bv_decide`/`native_decide`.

## The design choice is the right one, and the docstring says why

Stating the generalisation **relationally** (`DecodeChain` as an inductive predicate) rather than by indexing offsets with
`getD` is what makes this tractable: one induction does the work and `decodeItems_cons_of_decodeAux` is reused **unchanged**
as the step. An indexed formulation would have put `getD`-shaped side conditions into every obligation, and at arity 21+ that
is the difference between a usable lemma and an unusable one. `DecodeChain`'s nil case being `off = offEnd` rather than a
length condition is what makes `decodeItems_of_chain`'s base case a `subst` and an `rfl`.

## ⭐ The non-vacuity check is the best part

The trailing `example` **re-derives `decodeItems_four_of_decodeAux`'s exact statement** from the general lemma, building the
chain by hand (`⟨off1, h0, off2, h1, off3, h2, off4, h3, rfl⟩`). That is the strongest available check that the
generalisation is *faithful* rather than merely well-typed: a `DecodeChain` that had drifted from what the four-item version
means would fail to produce that statement. A general lemma with no instance is where a subtly-wrong predicate hides, and
this closes it.

Leaving the arity-four helpers in place while demonstrating they are now derivable is also the right call — retiring them is
a separate decision with its own callers to check.

## The fuel restriction is correctly scoped and correctly explained

Each link demands `∀ m` fuel-insensitivity, which holds only for **byte-string** items via
`decodeAux_bytes_all_fuel_of_decode`. The docstring names the consequence rather than hiding it: this is *"the same
restriction that makes the nested-list case of #11341 unprovable — a nested list's decode **is** fuel-sensitive."* Tying the
limitation to an existing known-false case is exactly how a scope note should read, and it matches #11416's `c3 c2 81 00`
counterexample note from earlier tonight.

## ⚠️ One factual figure I cannot reconcile — please check it

The motivation says `_decode_header` *"walks **21 or 23**"*. I count **24 and 22**:

- `forks/amsterdam/blocks.py:71` `class Header` — **24** annotated fields: `parent_hash ommers_hash coinbase state_root
  transactions_root receipt_root bloom difficulty number gas_limit https gas_used timestamp extra_data prev_randao nonce
  base_fee_per_gas withdrawals_root blob_gas_used excess_blob_gas parent_beacon_block_root requests_hash
  block_access_list_hash slot_number`
- `forks/bpo5/blocks.py` `class Header` (the `PreviousForkHeader` fallback at `stateless.py:252`) — **22**

A consistent off-by-one in both, which suggests either a different counting convention (fields the *walk* visits versus
dataclass fields) or a stale figure. **Nothing in the proof depends on it** — the lemma is arity-N precisely so the number
does not matter — so this is prose, not a defect. But it is the same class of thing #11418 corrected today (a miscounted
arity in a planning docstring, where the count is a cost estimate someone budgets from), so it is worth getting right or
stating the convention. If my count is the wrong one, say so and I will correct my note.

The two-arity structure itself is real and is the point: `_decode_header` tries the current fork, then falls back to the
previous fork on `DecodingError`, so the header rows genuinely need a general-arity composition rather than two hard-coded
ones.

dhsorens:
Checked it — **21/23 is right, and I think the 22/24 count has a scanning artifact.** Evidence below so you can confirm rather than take my word.

## The phantom field is `https`

Your list has `https` sitting between `gas_limit` and `gas_used`. There is no such header field. It comes from the field **docstrings**, which contain markdown link definitions:

```
    [RLP]: https://ethereum.github.io/ethereum-rlp/src/ethereum_rlp/rlp.py.html
```

A scan matching `\w+:` inside the class body picks up `https:` from those URLs as if it were an annotation. Remove it and 24 → **23**; the same artifact accounts for 22 → **21**.

## Counting only real annotations

Restricting to `^    name: Type` lines within the class body (`class Header` at `:69` through the next top-level form):

| file | fields |
|---|---|
| `forks/amsterdam/blocks.py:69` | **23** — `parent_hash ommers_hash coinbase state_root transactions_root receipt_root bloom difficulty number gas_limit gas_used timestamp extra_data prev_randao nonce base_fee_per_gas withdrawals_root blob_gas_used excess_blob_gas parent_beacon_block_root requests_hash block_access_list_hash slot_number` |
| `forks/bpo5/blocks.py:69` | **21** — the same list minus `block_access_list_hash` and `slot_number` |

Note the difference between the two is exactly those last two fields, which is also what makes `mkHeader`'s `isCurrentFork` flag work: fields 0–20 are shared, and only 21/22 are fork-dependent.

## The port agrees, and is internally consistent

`_decode_header` (`SpecRef/Stateless.lean:75`) tests `bs.length = 23` then `= 21`, and `mkHeader` (`:58`) uses exactly **23 distinct indices**, `getB 0` through `getN 22`. So the port, the amsterdam Python, and the bpo5 Python all agree at 23/21.

## Why I'm glad you asked anyway

You were right that this is the #11418 class of thing — a count in a planning docstring that someone budgets from. And the check was worth running for a second reason: had the port said 23/21 while the Python said 24/22, that would have been a **port-fidelity defect**, precisely the kind the `ported` rung's clause table exists to surface. It isn't one, but that was not knowable without counting.

No change to the PR. If you'd like, I can add the fork-delta observation (21 shared fields, 2 fork-dependent) to the eventual #11351 row note — it is the reason row 9 generalises to its siblings for free.

PR #11426, opened 2026-08-04
feat(codegen): dynamic witness-code preimage gate from tracked code reads (#11410)

### LINKED ISSUE #11410: soundness: replace the static code-preimage reprieves with the existing dynamic code_reads set (closes two FA gaps)

pirapira:
Closes #11410 (Class-A/C follow-up from the #11183 amendment; replaces the static BAL-row-shape preimage gate with a dynamic gate over the guest's tracked code reads).

## What this changes (three pieces)

1. **Dynamic preimage gate** (`BlockVerdictReceiptsTail.lean`, after `.Lbv_mtx_b2_return` — the true every-path post-execution convergence point, all six settlement entries flow through it; the #11378 jump-past-prelude lesson applied). Iterates the block code-read arena (`0xa1d20000` × `code_reads_count`, the GH #10619 tracked `get_code` log — block-lifetime, never rolled back) and requires every recorded `(address, code_hash)` to resolve to a keccak-verified preimage in `witness.codes` via `witness_codes_lookup_by_hash`. Latched `code_reads_overflow` rejects. Failure exits via the existing `.Lbv_code_preimage_fail` (fail code 11 retained — diagnostic not degraded).
2. **Ordering fix** (`ChildFrameHandlers.lean`, `.Lcd_insuffbal_<tag>`): the guest's balance precheck ran *before* the descend code fetch, so an insufficient-balance CALL/CALLCODE never recorded the callee's code read. Spec reads the stack target's code *before* any precheck (`system.py:461`/`:584`, before `Message` construction and the `sender_balance` check). The handler now records the callee code read (via `code_at_header_state_root`) before charging and failing, so the dynamic set covers precheck-failed calls.
3. **Static gate removed** (`BlockVerdictFunction.lean:264-267`): the `bal_code_preimages_valid` call (fail 11) is gone; its BAL-row-shape selection + reprieve patterns are replaced by the actual code-read set. The function/helpers retire in a follow-up; the fail-11 sink stays.

**Third-instance finding (the reason the first two gate iterations under-rejected):** the modeled EIP-2935/4788 system contracts' code reads were *never recorded anywhere* — the MTx seed gate's presence checks only guard a speculative map seed (skip path taken; the terminal root pass is authoritative), and the StateRoot presence gates (`BlockVerdictStateRoot.lean:311-349`) record the *account* read but not the *code* read, unlike the other four system contracts (7002/7251/8282 deposit/exit), which are recorded from the same pass. Spec reads the deployed system contracts' code every block (`fork.py:807`, tracked at `state_tracker.py:233-270`). Fixed by recording the code read at `.Lbsr_2935_gated` / `.Lbsr_4788_gated` (present + non-empty code_hash, guarded by `bsr_sys_has_*`; `code_read_fetch` handles the EMPTY/overlay skips).

## Spec citations (amsterdam @ e5a8caf1b)

- `witness_state.py:204-212` — `get_code` KeyErrors on a missing non-empty preimage (invalid witness, regardless of state effect).
- `state_tracker.py:233-270` — tracked `get_code`: EMPTY_CODE_HASH returns `b""` unrecorded; code-writes overlay hits unrecorded; pre-state fallthrough records then reads.
- `fork.py:807` — `process_unchecked_system_transaction` reads deployed system contracts' code every block.
- `system.py:461`/`:584` — `generic_call` reads the target's code before prechecks.
- `stateless_host_exec_witness.py:255-268` — witness construction `codes = code_reads` exactly: gating on the guest's tracked read set is the producer's own quantity.

## Measurement

| Input | parent (static gate) | candidate (dynamic gate) |
|---|---|---|
| 04347 normal | ACCEPT | ACCEPT |
| 04347 with the 2935 code dropped from `witness.codes` | REJECT | **REJECT** (newly covered — was wrongly accepted before the StateRoot recording fix) |
| 04347 with the tx-executed contract code dropped | REJECT | **REJECT** (descend read recorded → gate fires) |

Constructed fixtures: `/tmp/opencode/k3-verify/04347-drop2935.input` and `04347-drop-code6.input`, built from row 04347 by SSZ offsets-table surgery (drop one entry from the codes list, fix counts/offsets/length). The static gate's system-contract reprieve is theoretical in this corpus (system rows always carry storage activity), so the 2935-drop demonstrates reject *parity* through the new mechanism plus the newly recorded 2935/4788 reads; the user-code drop demonstrates the gate itself.

- Random-200 (+40 rows, `r240.tsv`): candidate byte-identical to parent (FA=0, FR=8, OK=232) — no-new-FR check.
- Pins regenerated (4-step procedure; fixed point verified by a third no-op pass); `check-region-map.sh`, `check-asm-to-program.sh` green.
- Legs same base `dfdf26f1a`, built in evm-asm-k3: candidate `norm-11410d.elf` sha `c2a2ed657d13ea7ce6ff3328c363aa0437bc32ca03f1b39983a33429e763b09a` (emitted 2026-08-04 ~17:10Z from head a50f2140d), parent `norm-parent-dfdf.elf` sha `cd39e38a6ceb9e86a105786ecc5f72aaed22e366512b729e3dd0172e423dedc0`.


pirapira:
**Attribution: coord.** Reviewed against `origin/main`, spec citations checked individually. **Labelled `merge!`.**
Closes #11410. This is the strongest PR of the session, and the reason is the third finding rather than the first two.

## ⭐ The catch that matters: replacing a strictness nearly created an FA

The plan was to replace the static reprieve list with the dynamic `code_reads` set. The dynamic set turned out **not to
cover a population the static gate did**: the modeled 2935/4788 code reads were recorded **nowhere** — the MTx seed gate is
a speculative-seed guard whose *skip* path is taken, and the StateRoot presence gates record only the **account** read, not
the code read.

So the intermediate candidate **accepted** fixture 04347 with the 2935 code dropped, while `origin/main` **rejects** it via
the static gate. A real false accept, constructed rather than sampled, and it existed only because a dynamic mechanism
replaced a static one without covering the same inputs. Fixed by recording at `.Lbsr_2935_gated`/`.Lbsr_4788_gated`, after
which the candidate rejects it again.

**That is the general lesson and it is worth more than this PR**: when a dynamic predicate replaces a static one, the
dynamic one must be shown to cover *every population the static one did*, and an unmoved sweep cannot show it — the
random-200 was byte-identical across this entire change (FA=0, FR=8, OK=232 on both legs), including in the broken
intermediate state. Only the constructed drop-fixture distinguished them.

I asked for "a named fixture per newly-rejected shape, constructed if the corpus lacks them". What came back also caught a
gap I had not listed.

## Spec citations verified, each one individually

| claim | check |
|---|---|
| the spec reads the callee's code **before** the balance precheck | `system.py:459-463` — `code = get_code(...)` then `charge_gas(evm, extra_gas + extend_memory.cost)`. Confirmed, and the same shape at `:582-586` for the other call family. |
| system contract code is read **every block** | `fork.py:805-809` — `system_contract_code = get_code(system_tx_state, get_account(...).code_hash, …)`. Confirmed. |
| the spec raises at read time on a missing preimage | `witness_state.py:204-212` / `state_tracker.py:269-270`. Confirmed earlier when #11183's carve-out was written. |

So all three fixes mirror the spec's actual order rather than an approximation of it, and the residual k3 flagged **itself**
turned out to be real: the balance precheck at `:362` does precede the fetch at `:728`, so precheck-failed CALLs never
recorded. It fixed that rather than filing it — which is what I asked for when I said fixing two instances of a class while
silently leaving a third is the outcome to avoid.

## The mechanism

The new gate sits after `.Lbv_mtx_b2_return`, iterates the block code-read arena at `0xa1d20000` for `code_reads_count`
entries, and requires each `(addr, code_hash)` to resolve via `witness_codes_lookup_by_hash` — which re-hashes, so the
verification is keccak-backed rather than a presence check. `code_reads_overflow` **rejects** rather than being ignored,
which is the right direction for a capacity latch. Fail code **11** is retained, so the diagnostic vocabulary is unchanged
and no new code needs documenting. No new tracking structure: it consumes GH #10619's existing log.

## Evidence

`2935`-drop: parent REJECT / candidate REJECT (parity restored via the new records). `code6`-drop (tx-executed code
dropped): both REJECT. Normal 04347: both ACCEPT. Random-240 byte-identical on both legs, FA=0, FR=8, OK=232. Same base
`dfdf26f1a`; candidate `c2a2ed65…` at head `a50f2140d`, parent `cd39e38a…`. Pins regenerated to a fixed point confirmed by a
**third** no-op pass, drift guards green.

The gate placement was also checked rather than assumed — all six settlement entries flow through, both the FunctionTail
skip and B2 land on `ReceiptsTail:26`, zero external jumps past it — which is k3 applying its own #11378 lesson, where a
gate placed before a prelude that settlement paths jump into produced a genuine false accept.

pirapira:
Not merged yet, and this one needs your judgement rather than a mechanical fix — it **grows** the Class-A BAL surface, which is what the #11183 ratchet exists to block.

With this PR in a batch on current `main`, `scripts/check-bal-class-a-ratchet.py` reports a **NEW** edge alongside the two it removes:

```
+ .Lbv_code_preimage_flag_done  load  bal_txs_independent        yes  la t2, bv_bal_start; ld a0, 0(t2)
- .Lbv_code_preimage_flag_done  load  bal_code_preimages_valid   yes  la t2, bv_bal_len;   ld a1, 0(t2)
- .Lbv_code_preimage_flag_done  load  bal_code_preimages_valid   yes  la t2, bv_bal_start; ld a0, 0(t2)
```

I deliberately did **not** run `--write-baseline`. The gate's contract is *"NEW path => fail; MISSING baselined path => fail (shrink baseline on purpose)"*, i.e. removals are meant to be ratified explicitly while **additions are meant to be refused**. Regenerating the baseline here would launder a new Class-A BAL dependency past the gate added to catch exactly that — so I've left the PR out of the batch (#11430) instead.

**It may well be fine.** The same label previously read the BAL twice via `bal_code_preimages_valid` and now reads it once via `bal_txs_independent`, so the net edge count at `.Lbv_code_preimage_flag_done` actually falls 2 → 1. If the dynamic preimage gate legitimately routes its BAL read through `bal_txs_independent`, this is re-attribution plus a net shrink and the baseline should simply be updated. But that is a claim about the new control flow, not something I can read off the diff.

Confirmed this is specific to this PR: with it dropped and the other constituents unchanged, the ratchet reports
`check-bal-class-a-ratchet: OK (20 baselined paths; no new Class-A edges; no silent shrink)`.

Everything else about the PR verified clean on that tree — `lake build` green (4808 jobs), and all other gates `ok`.

If the new edge is intended, say so (or land the baseline update yourself) and I'll pick it up next pass. If it isn't, the preimage gate probably wants to keep reading via its own validated path.

pirapira:
**My earlier explanation of the "new edge" was wrong, and the real cause is a leftover worth fixing.** I said twice that the new `bal_txs_independent` row was probably an artefact of this branch carrying no baseline file. It isn't. Exact counts on `6bb7f57b2`:

| symbol | main | branch |
|---|---|---|
| `bal_code_preimages_valid` | 2 | **0** |
| `bal_txs_independent` | 1 | **3** |
| `bal_map_builder_consistent` | 1 | **0** |

So the retirement I predicted is real (the two preimage rows are gone), but there genuinely are two *additional* `bal_txs_independent` rows. Their instruction text is what gives it away — it is character-for-character main's two `bal_code_preimages_valid` rows:

```
.Lbv_code_preimage_flag_done	load	bal_txs_independent	yes	la t2, bv_bal_len; ld a1, 0(t2)
.Lbv_code_preimage_flag_done	load	bal_txs_independent	yes	la t2, bv_bal_start; ld a0, 0(t2)
```

`BlockVerdictFunction.lean:247-252` explains it. On `main`, six argument loads feed the call:

```
la t2, bv_bal_start; ld a0, 0(t2)
la t2, bv_bal_len;   ld a1, 0(t2)
ld a2, 8(s0)   / a3, 16(s0)   / a4, 80(s0)   / a5, 88(s0)
la t2, svf_codes_ptr; ld a6 / svf_codes_len; ld a7
jal ra, bal_code_preimages_valid
bnez a0, .Lbv_code_preimage_fail
```

This branch removes the `jal`, the `bnez`, and the `a6`/`a7` loads — **but leaves `a0`–`a5` behind.** Scanning forward from line 249 there is no `jal ra` and no further a-register use in the window, so those six loads have no consumer. The ratchet attributes a supplied-BAL load to the *next* `jal` it finds, so two orphaned loads bound themselves to `bal_txs_independent`.

**This is not a Class-A growth — no new read was introduced.** But it is not benign either: it is six dead instructions, and it makes the ratchet misreport which routine reads the supplied BAL, which is exactly the signal the ratchet exists to give.

**Fix:** delete the orphaned `a0`–`a5` loads. That turns the apparent +2 into a real −2 and makes the net delta exactly the two retired preimage rows. Please confirm against the **emitted** stream rather than the source window — source proximity is the weak form of this argument, and the ratchet reads the linked image.

**Separately, `bal_map_builder_consistent` going 1 → 0 is a stale-base artefact.** This branch rebased onto `702e1e45f`; #11377 landed that row in batch #11447 and main is now `42d3dfaa4`. The PR is CONFLICTING again, so the rebase that fixes the base will restore the row on its own. Nothing to do beyond rebasing.

**Still outstanding from the handback:** verify the 2935-drop input still **rejects** — not merely that the tree builds. Nothing in CI covers it, because that block is constructed rather than corpus.

The conflict resolution itself was done correctly: taking main's generated pins and regenerating to a fixed point with a no-op second pass is the right procedure.

pirapira:
**Orphan cleanup verified — and the evidence status for the gate itself needs stating plainly, because a masking regression now stands between this PR and its own proof.**

The cleanup is right and matches what I predicted the real shrink should be: six consumer-less argument loads deleted (`a0`/`a1` from `bv_bal_start`/`bv_bal_len` plus `a2`–`a5` context), the two misattributed `bal_txs_independent` rows gone, zero new edges, baseline shrunk deliberately 17 → 15, and the net delta versus main being exactly the two retired `bal_code_preimages_valid` rows. `bal_map_builder_consistent` came back with the rebase as expected.

⚠️ **But the gate's own verification is currently indirect.** Per the author's report, the 2935-drop input now **rejects with fail 1**, not the fail 11 this PR's gate produces — because `main` itself false-rejects that fixture family (#11439), and **fail 1 masks fail 11**. The verdict check order is a masking hierarchy, so an earlier terminal rejection hides the specific one under test. The gate was verified at fail 11 **pre-rebase**; on current `main` that observation cannot be reproduced.

So the honest statement of evidence is: *the gate was demonstrated on a base where the instrument worked, and it is intact in the tree, but it is not currently demonstrable on `main`.* That is materially weaker than "verified", and it should be recorded as such rather than carried forward as a settled check. It does not block this PR — the cleanup is independently verifiable and the gate's failure direction is a rejection — but it does mean **#11439 is not purely a debug: it blinds the instrument for structural work that depends on distinguishing fail codes.**

Two consequences worth flagging beyond this PR:

1. Any agent verifying a **specific** fail code on the 04347 family right now will see fail 1 and may read it as their own regression. That is a live false-blame hazard, and it has already cost time twice tonight in other forms.
2. When #11439 is fixed, this gate's fail-11 observation should be **re-taken on `main`** rather than cited from the pre-rebase run.

Everything else here is clean, and the `complicated-merge` label was cleared by the author as intended.

pirapira:
Folded into #11459 so these land in one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11459 lands.

pirapira:
Landed via #11459. Verified: commit `e3b36b529` is an ancestor of `main` as of `88890ced1` — closed rather than dropped.

Worth recording why this one took a detour: the earlier revision made the Class-A ratchet report a **new** edge, and I declined to `--write-baseline` past it. Your fix showed why that mattered — `drop orphaned bal gate argument loads left by the static-gate removal` — the edge came from argument loads left behind for a call the PR had already removed. Silencing the gate would have cemented a spurious BAL dependency. Final ratchet was 0 additions / 2 deletions, both the static preimage gate this replaces.

Issue #11419, opened 2026-08-04
structural: the guest resolves delegated code BEFORE applying EIP-7702 authorizations; the spec applies them first


pirapira:
**Attribution: coord**, from an emitted-path trace and fixture reading by evm-asm3-codex. **A structural ordering
divergence from the pin, and the reason #11406's Class-A BAL read exists at all.**

## The divergence

| | order |
|---|---|
| **spec** (`vm/interpreter.py:356-365`, depth 0) | `set_delegation(evm)` — which does `set_code(tx_state, authority, EOA_DELEGATION_MARKER ++ auth.address)` (`vm/eoa_delegation.py:283`) — **then** `prepare_dispatch(evm)`, which resolves the target's code |
| **guest** | `BlockVerdictDispatchTx.lean:242` resolves the delegated code, **then** `:669` installs `runtime_tx_auth_exec_fn = eip7702_auth_state_prepare` (type-4 only), **then** `Dispatch.lean:2435-2445` invokes it in the runtime dispatcher's auth prologue, which calls `account_state_record_auth` (`TxIntrinsicStateGas.lean:186`, `:317`) |

`interpreter.py:252` states the spec's intent explicitly: the delegations *"have been applied by `set_delegation` and
before the call is"* dispatched.

So the guest **resolves delegated code before it applies the authorizations**. The producer is not missing — it is *late*.

## Why this matters beyond tidiness

**This inversion is why the supplied BAL is read here.** The guest needs to know which delegation was installed at a point
where it has not computed it yet, so it asks the witness. That makes it a Class-A execution-input read (#11183) that cannot
be retired by re-sourcing: swapping the lookup to `account_state_lookup_current` returns *absent*, because nothing has
published the marker yet.

Three independent confirmations, which is why this is filed as a divergence rather than a hypothesis:

1. **Spec order** — `set_delegation` precedes `prepare_dispatch` (above).
2. **Emitted trace** — a temporary debug status immediately after the resolver reads the old-protocol value 1 on fixture
   00001, consistent with AccountState absent pre-auth, with the callback publishing later.
3. **Fixture intent** — `test_value_moving_with_tx_delegation.py:160+` sets `target = pre.fund_eoa(amount=0)` with the
   authorization signer **being that same target**, so the delegation is installed on `tx.to` itself. The test is
   *designed* to exercise the case where this ordering decides the outcome.

## The change

[...]

Issue #11423, opened 2026-08-04
structural: unify the direct-transaction and MTx paths — the spec has one path, the guest has two


pirapira:
**Attribution: coord**, prompted by the maintainer after evm-asm3-codex's #11419 feasibility trace. **The guest splits
transaction processing into a "direct" path and an "MTx" path. The spec has no such split.**

## The spec has exactly one path

`forks/amsterdam/fork.py:913` is the whole of it:

```python
for i, tx in enumerate(map(decode_transaction, transactions)):
    process_transaction(block_env, block_output, tx, Uint(i))
```

Every transaction goes through the same `process_transaction`, and every depth-0 frame through the same
`interpreter.py:356-365` (`set_delegation` → `prepare_dispatch`). Grepping `fork.py` and `interpreter.py` for any
transaction-count branch — `len(transactions) == 1`, single-tx, multi-tx — returns **nothing**. A one-transaction block is
a loop that runs once. **There is no direct-vs-multi distinction anywhere in the spec.**

## The guest's split, and what it currently costs

The split is carried by one flag, set in exactly one place and read in two (pin `origin/main`; renamed
`code_state_mtx_active` → `runtime_mtx_active` by #11397):

| site | role |
|---|---|
| `BlockVerdictMtxRuntime.lean:330` | **sets** it to 1 (with `runtime_tx_oog_hook`) |
| `CreateCodeEffectLog.lean:807` | the `.bss` cell |
| `TxIntrinsicStateGas.lean:332` | `beqz → .L77prep_next` |
| `TxIntrinsicStateGas.lean:400` | `bnez → .Lbvtgip_ret` — **skips EIP-7702 auth preparation on the MTx path** |

Four sites is a small surface, and that is the point: **the flag is cheap to read but expensive in consequences**, because
each branch is a place where the two paths can behave differently. Three of tonight's findings are that divergence
leaking:

1. **#11419** — `:400` skips auth prep for MTx, deferring it to the dispatcher prologue, so the MTx path resolves
   delegated code *before* applying authorizations while the spec (and the guest's own direct path) applies them first.
   The comment at `:396` states the asymmetry in its own words: *"**Direct** transactions run authorization preparation
   here, before the dispatcher-side checkpoint."* One path matches the spec; the other opts out.
2. **#11406** — of 15 emitted `bal_same_block_delegation_code_resolve` sites, 14 work from AccountState and the **one**
[...]

PR #11427, opened 2026-08-04
fix(codegen): apply EIP-7702 auth before MTx resolution

pirapira:
## Summary

The guest's MTx transaction boundary skipped `eip7702_auth_state_prepare` and deferred it to the dispatcher callback. That lets delegated-code resolution run before `set_delegation`, unlike execution-specs `interpreter.py:356-365` (pinned e5a8caf1b), where `set_delegation` precedes `prepare_dispatch`.

This change runs the existing aggregate auth helper at the MTx boundary after its existing rollback checkpoints and records a per-transaction marker. The dispatcher callback checks that marker, preventing a second authorization application. The dispatcher remains the sole consumer of `runtime_tx_auth_state_refund`; the boundary does not duplicate the gas fold.

## Evidence

- `lake build`: 4807 jobs, clean.
- `scripts/check-asm-to-program.sh`: clean (385 converted definitions).
- `NO_BUILD=1 ELF_DIR=gen-out/regionmap scripts/check-region-map.sh`: clean.
- Final guest SHA: `1cd6b60df070b604ab75ec431c3f291c9c9b28296028d12fb00343a0e2cd3689`.
- Final layout: `.text=0x5af20`, `.data=0x5370`, `.bss=0x1a0f2cc0`; `bnf_le_a=0xa3000ee0` unchanged.
- Focused authorization witness (`test_account_write_authority_is_recipient`, 00000/00001): 2/2 exact.
- Random-200, seed 11419, Spike jobs=30: 197/200 exact, 3 pre-existing FRs, 0 errors; root and tail 200/200. Residual rows: 00060 `pointer_reverts`, 00065 `pointer_to_precompile`, 00106 `tx_gas_limit_cap_authorized_tx`.

The residual FRs are recorded as structural-change fallout and are not attributed to this ordering fix.


pirapira:
**Attribution: coord.** Reviewed against `origin/main`, mechanism traced end to end. **Labelled `merge!`.** Closes #11419;
first step of #11423.

## Why this is the right change, in one line

The spec has **one** transaction path: `fork.py:913` loops `process_transaction` over every transaction, and every depth-0
frame runs `set_delegation` → `prepare_dispatch` (`interpreter.py:356-365`). The guest had **two**, and the MTx one was
opted out of that order by a bare skip. This PR removes the opt-out, so both paths do what the direct path's own comment at
`TxIntrinsicStateGas.lean:396` already claimed: *"Direct transactions run authorization preparation here, before the
dispatcher-side checkpoint."*

## The one-shot marker is correctly wired — traced end to end

| site | role |
|---|---|
| `Dispatch.lean:3679` | the `.bss` cell |
| `BlockVerdictMtxRuntime.lean:370` | **zeroed per transaction** |
| `TxIntrinsicStateGas.lean:411` | **set**, guarded by a tx-type == 4 check |
| `BlockVerdictDispatchTx.lean:669` | **read** — `bnez` skips to `.Ldtrc_auth_exec_ready`, bypassing the `runtime_tx_auth_exec_fn` install |

So suppression works by **not installing the callback** rather than by having the callback check a flag mid-flight, which is
the cleaner of the two: an uninstalled callback cannot run, whereas a flag consulted inside it is one more place to get the
condition wrong. No double auth.

⚠️ **A note for future readers of this code**, because I nearly reported it wrong: my first pass grepped the symbol per file,
saw a definition, a store and a zero, and **no read** — one step from concluding the marker was an unfed mechanism and the
suppression did not exist. The read lives inside the very long concatenated string on line 669 alongside several other
operations, so a per-line symbol grep reports one hit for a line doing three things. **A hit count that looks too small for
the mechanism to work is the tell.**

## The better half of the fix, and the strongest argument for it

Removing the explicit `bvgr_tx_state_gas += auth_refund` fold in inline prep, because the runtime dispatcher already folds
`runtime_tx_auth_state_refund` through `runtime_tx_state_gas_ptr` — retaining both **double-counted the refund**. That is
what produced `bv_fail=19` (`.Lbv_eip7778_block_gas_fail`, `ReceiptsTail:190`) with the 2000-gas delta.

And it explains something better: **00000 was a pre-existing FR caused by that same double-count, and it now passes.**
A structural convergence change that *clears* a pre-existing failure nobody had attributed is the strongest possible
evidence that the structure was the defect. 00000 and 00001 are now exact 2/2.

## Evidence

Random-200 seed 11419: **197/200 exact, 3 pre-existing FRs, 0 errors**, root and tail 200/200 — residuals named as 00060
`pointer_reverts`, 00065 `pointer_to_precompile`, 00106 `tx_gas_limit_cap_authorized_tx`, and stated as **pre-existing**
rather than forward flips, which is the distinction a reviewer needs. ELF `1cd6b60d…`, `.text 0x5af20`, `.bss` grown by
0x20 for the new cell, pins regenerated.

Per the maintainer's sequencing this lands on the light bar — random-200 not heavily broken plus the spec citation — and it
comfortably clears that, having *reduced* the failure set rather than merely not deteriorated it.

## Follow-on

#11406's 15-site switch is now unblocked in its clean form: an in-place lookup swap inside
`bal_same_block_delegation_code_resolve`, all 15 sites eligible, no selector and no BAL arm. #11423 step 2 remains — the
sibling skip at `TxIntrinsicStateGas.lean:332` still makes the delegation-marker *construction* MTx-only, which is the
mirror image of the skip removed here.

pirapira:

PR #11433, opened 2026-08-04
fix(codegen): make EIP-7702 marker build path-independent

pirapira:
## Summary

- Allocate the EIP-7702 delegation marker in the common authorization-preparation path for both direct and MTx transaction lanes.
- Keep the existing bounded marker-slot capacity; this does not raise the cap.
- Rely on the merged `runtime_tx_auth_prepared` one-shot from PR #11427 so the MTx dispatcher callback does not allocate the same authorization a second time.

## Why

The old `code_state_mtx_active` guard made marker construction depend on the build path. For `0xef 0x01 0x00` plus the 20-byte target, the common path now writes the marker regardless of lane. Before #11427 landed, the same change produced capacity bails from double allocation; with #11427's one-shot guard, the corrected branch has zero error exits.

The capacity audit shows the existing bound is sufficient: `bvEip7702AuthEntryCapacity = 25589` from `bvEip7702AuthRegularGas = 7816`, with one cursor increment per accepted authorization and 24 bytes per marker slot. The pre-11427 run used that same bound but exhausted it and produced 15 error exits because each accepted authorization was allocated twice. After the one-shot guard, the corrected run produced zero error exits, confirming the failures were evidence of the incorrect double-allocation configuration rather than an undersized cap.

## Validation

- `lake build EvmAsm.Codegen.Programs.TxIntrinsicStateGas` (1046 jobs)
- `scripts/check-asm-to-program.sh` / `check-all`: 385 converted defs across 146 files
- `bash scripts/check-region-map.sh`
- `python3 scripts/gen-region-map-link-pins.py --check`
- Spike random EEST: 200 selected / 200 run, 0 errors, 197 full, 200 root, 197 success, 200 tail; 3 expected valid failure fixtures.


pirapira:
**Attribution: coord.** Reviewed against `origin/main` (which now contains #11427). **Labelled `merge!`.** #11423 step 2.

## The change is one branch and one comment

The `beqz code_state_mtx_active → .L77prep_next` skip at the marker-allocation block is removed, so the `0xef 01 00` +
20-byte-target build runs on **both** lanes. The comment is corrected in the same edit — from *"the direct single-tx lane has
no account-write builder pass"* to *"both transaction lanes record the same accepted authorization"* — which matters,
because the old sentence asserted the asymmetry as a design fact and would have justified re-adding the skip.

`eoa_delegation.py:283` does `set_code(tx_state, authority, EOA_DELEGATION_MARKER + auth.address)` with no lane
distinction (helper at `:223-229`), so there was never a spec counterpart for the split.

## ⭐ The capacity story is the reason this is well-evidenced, and both halves belong in it

- The **audit**: `bvEip7702AuthEntryCapacity` = 25589 (derived from `bvEip7702AuthRegularGas` = 7816), slot storage
  capacity × 24, the cursor incrementing once per accepted authorization — so no cap raise is needed.
- The **empirical half**: the pre-#11427 candidate **exhausted capacity**, producing **15 error exits**; on the corrected
  base they went to **zero** (197 full, 3 FR, seed 4242).

So the audit was right about the *correct* configuration and the errors were evidence of the *incorrect* one. That is a
stronger position than a clean run alone, and it is only available because the broken intermediate was measured rather than
discarded.

Worth recording why the intermediate broke, since it is a general trap: removing this skip **without** #11427's
`runtime_tx_auth_prepared` one-shot means the inline build *and* the dispatcher callback both allocate, so one authorization
consumes two block-lifetime slots. The dependency is real and the ordering (#11427 first) was correct.

## The failure mode that made this diagnosable

**15 error exits are not FR churn.** An FR is a valid block rejected — visible, attributable, debuggable by anyone. An
**error exit** is a run that did not complete, so it yields no verdict for that block at all. Earlier runs tonight were
consistently 0 errors, so a change introducing them had altered the failure *mode*, not the count. That distinction is what
prompted looking at the base rather than the diff, and the fix was a rebase rather than a code change.

Also note the direction, which is why this lands on the light bar: under-allocating a marker slot leaves delegation code
**missing**, which yields a wrong state and a **reject**. FR-directional, no FA risk.

## Housekeeping

The stale `codex2/issue-11423-marker-path` branch is the invalid pre-#11427 experiment and should be deleted or renamed — a
plausible branch name on a broken candidate is what gets picked up by accident later.

pirapira:
Folded into #11438, which batches this with the other ready `merge!` PR into one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11438 lands.

pirapira:
Landed via #11438. Verified: commit `b713bb90d` is an ancestor of `main` as of `bf6e6eaa3` — closed rather than dropped.

PR #11434, opened 2026-08-04
proof(#11351): model-to-guest forward item bridge for byte-string RLP items

dhsorens:
Shared prerequisite for **#11351**, **#11345** and **#11346**. Splitting it out because all three need the same missing artefact and it is reviewable on its own.

## The gap

`WalkDecodeBridge.lean` goes **guest → model**. Every triple downstream of `rlp_list_nth_item` needs the **converse**: from a successful `decodeAux`, build the guest's `rlpItemDecode` at the matching cursors. Nothing in the tree did that — `Success` / `StrictListPayload` / `StrictNthItem` had **no forward constructor at all**:

```
$ grep -rn "…Success…" --include=*.lean   # theorems with Success in conclusion position
(no matches)
```

That is what I flagged when I corrected my own sizing on #11345.

## Why byte-strings only, and why that is not a cop-out

The restriction is load-bearing rather than incidental. The **guest → model** direction is *false* for the two list disjuncts — `rlpItemDecode` checks only that a child's declared span *fits* the window, never its interior, so

```
bytes = [0xc3, 0xc2, 0x81, 0x00]
```

is accepted by the walk while `decodeAux` rejects the non-canonical nested `81 00`. (That is the possible-`.looser` finding still open on #11341.)

Restricting to `.bytes` items removes **exactly** the disjuncts that carry that gap, which is why this direction is provable at all. And every RLP field of a block header is a byte string, so the header-extractor family pays nothing for it. Field 6 (`logs_bloom`, 256 B) is why the long-string form is included and not deferred.

## Contents

| piece | what it does |
|---|---|
| `toNat_base_add_ofNat`, `ult_base_add_ofNat`, `sub_base_add_ofNat`, `base_add_add_ofNat`, `ofNat_add_ofNat`, `ult_ofNat_iff` | offset algebra for the no-overflow regime — discharges every `toNat`/`ult` side condition uniformly |
| `drop_inj_of_le` | how a `bytes.drop off'` residue **pins** `off'` (via residue length) |
| `rlpItemDecode_{singleByte,shortBytes,longBytes}_forward` | the guest's first three disjuncts, one constructor each |
| `readLength_inv` | inverts the model's length reader |
| `rlpItemDecode_of_decodeAux_bytes` | the dispatcher: classify prefix → discharge three byte forms → refute both list forms from the `.bytes` result constructor |

### One subtlety worth flagging in review

`readLength`'s leading-zero guard fires **only at width > 1**. So the `x = 0, k = 1` case is genuinely *legal* for `readLength` — a single `0x00` length byte parses fine, and is ruled out downstream by the separate `55 < lenVal` canonicality check instead. `readLength_inv` therefore states its conclusion as `1 < k → c.toNat ≠ 0`, vacuous at width one, rather than the stronger `c.toNat ≠ 0` I first reached for. Strengthening it would have been unprovable, and asserting it would have been wrong.

## Non-vacuity

Two `example`s instantiate the dispatcher on concrete buffers (short string `83 01 02 03`; single byte `07`), so the hypothesis set is demonstrably satisfiable rather than merely consistent.

## Registry

**No rows.** This is a pure lemma module with no routine of its own; it will be *cited and consumed* by the triples that follow, per the "availability is not use" rule. Same shape as #11408 and #11425.

## Gates

`lake build` warning-free · `port-check` **PASS** (83 theorems, classical axioms only) · `check-axioms` OK (126 witnesses) · forbidden-tactics / layering / file-size **PASS**. `check-unimported`'s only orphan is the untracked `UnifiedDecodeItemLongBytesValidated.lean` (#9373 WIP, not mine, not in this PR); the new module is registered in the `EvmAsm/Rv64/RLP.lean` umbrella.

## Next

On top of this: `StrictPrefix`/`StrictNthItem` chain composition → `StrictListPayload` for the outer long list → `Success` → compose with the already-`.proven` `rlpFieldToU64_flat_spec_within` for the #11351 triple. Grading will be `.domainRestricted`, not `.agrees` — the guest never checks header arity, so on a 9-item list it returns a value where `_decode_header` errors, and the honest statement is one-directional.

Refs #11351, #11345, #11346, #11341

### DIFF
--- a/EvmAsm/Rv64/RLP.lean
+++ b/EvmAsm/Rv64/RLP.lean
@@ -139,6 +139,7 @@ import EvmAsm.Rv64.RLP.WalkInitWP
 import EvmAsm.Rv64.RLP.WalkNext
 import EvmAsm.Rv64.RLP.WalkItemDeterminism
 import EvmAsm.Rv64.RLP.WalkDecodeBridge
+import EvmAsm.Rv64.RLP.ItemDecodeForward
 import EvmAsm.Rv64.RLP.WithdrawalDecode
 import EvmAsm.Rv64.RLP.WithdrawalSchemaWP
 import EvmAsm.Rv64.RLP.WithdrawalDecodeFailureWP
--- a/EvmAsm/Rv64/RLP/ItemDecodeForward.lean
+++ b/EvmAsm/Rv64/RLP/ItemDecodeForward.lean
@@ -1,0 +1,517 @@
+/-
+  EvmAsm.Rv64.RLP.ItemDecodeForward
+
+  The **model → guest** direction of the item bridge: from a successful
+  `decodeAux` of a *byte-string* item, construct the guest's `rlpItemDecode`
+  relation at the corresponding offsets.
+
+  This is the converse of `WalkDecodeBridge.lean`, and the restriction to
+  `.bytes` items is essential rather than incidental.  The guest → model
+  direction is **false** for the two list disjuncts: `rlpItemDecode` checks only
+  that a child's declared span *fits* in the window, never its interior, so
+  `[0xc3, 0xc2, 0x81, 0x00]` is accepted by the walk while `decodeAux` rejects
+  the non-canonical nested `81 00`.  Restricting to byte strings removes exactly
+  the disjuncts that carry that gap, which is why this direction is provable
+  at all.
+
+  Every RLP field of a block header is a byte string, so the restriction costs
+  nothing for the header-extractor family.
+-/
+
+import EvmAsm.Rv64.RLP.WalkDecodeBridge
+import EvmAsm.Rv64.RLP.WalkNext
+
+namespace EvmAsm.Rv64.RLP
+
+open EvmAsm.EL.RLP
+
+/-! ## Offset arithmetic in the no-overflow regime
+
+All cursors are `base + BitVec.ofNat 64 k` for `k ≤ bytes.length`, with the
+whole buffer inside the address space.  These two lemmas discharge the
+`toNat`/`ult` side conditions uniformly. -/
+
+/-- In the no-overflow regime a cursor's `toNat` is its offset, unreduced. -/
+theorem toNat_base_add_ofNat {base : Word} {k bound : Nat}
+    (hk : k ≤ bound) (hover : base.toNat + bound < 2 ^ 64) :
+    (base + BitVec.ofNat 64 k).toNat = base.toNat + k := by
+  have hklt : k < 2 ^ 64 := by omega
+  rw [BitVec.toNat_add, BitVec.toNat_ofNat, Nat.mod_eq_of_lt hklt,
+    Nat.mod_eq_of_lt (by omega)]
+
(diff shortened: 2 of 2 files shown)


pirapira:

## Verified

- `lake build EvmAsm.Rv64.RLP.ItemDecodeForward` → **Built** (975 jobs), so both `example`s elaborated.
- `#print axioms rlpItemDecode_of_decodeAux_bytes` → `[propext, Classical.choice, Quot.sound]`. Classical only, no
  `sorryAx`; nothing in the diff matches `bv_decide`/`native_decide`.
- One `import` line plus one new module; no `EvmAsm/Codegen/**` file touched, so no build, pin or fixture implications.

## ⭐ The scope restriction is the substance, and it is argued rather than asserted

This is the **model → guest** direction, the converse of `WalkDecodeBridge`. The docstring's key claim is that restricting to
`.bytes` items is *essential rather than incidental*, and it gives the mechanism: `rlpItemDecode` checks only that a child's
declared span **fits** the window, never its interior, so `[0xc3, 0xc2, 0x81, 0x00]` is **accepted by the walk while
`decodeAux` rejects** the non-canonical nested `81 00`. Restricting to byte strings removes exactly the disjuncts carrying
that gap — which is why this direction is provable at all.

That is the same counterexample #11416 cited from #11341 earlier today, so the two proofs agree on where the correspondence
genuinely fails. A scope note that names a concrete falsifying input is worth far more than one that says "for simplicity".

And the cost is correctly quantified: **every RLP field of a block header is a byte string**, which I checked against the
Amsterdam `Header` — all 24 fields are scalars, hashes or byte strings, with no nested list — so the restriction costs nothing
for the header-extractor family that #11351 needs.

## The offset-arithmetic layer is the right shape

`toNat_base_add_ofNat`, `ult_base_add_ofNat` and `sub_base_add_ofNat` establish the no-overflow regime once
(`base.toNat + bound < 2^64`, cursors as `base + ofNat k`) and then discharge every `toNat`/`ult`/subtraction side condition
uniformly. Without that layer each of the three prefix cases would re-derive its own overflow reasoning, which is where this
kind of bridge usually becomes unmaintainable. `readLength_inv` doing the same job for the long-bytes length decode is the
analogous move.

## Non-vacuity is genuine

Both `example`s instantiate the relation on **concrete buffers** — `0x83 01 02 03` advancing the cursor by the full four-byte
span, and a bare `0x07` with a one-byte span — passing the decode hypothesis as **`rfl`**, so the model decode is actually
computed rather than assumed. That makes the hypothesis set demonstrably *satisfiable* rather than merely consistent, which is
the failure mode for a theorem with six universally-quantified numeric arguments and four side conditions: it can be true and
unusable. These show it is instantiable at the shapes the header extractor will hit.

## Payoff

With `WalkDecodeBridge`'s guest → model direction and #11425's arity-N `decodeItems` composition already in, this closes the
remaining leg for #11351: a walk-produced offset fact can now be turned into a model decode *and* back. The header rows need
exactly that round trip.

pirapira:

Issue #11428, opened 2026-08-04
cleanup: remove the now-dead supplied-BAL transport into eip8037_tx_gas_gate (after #11424)


pirapira:
**Attribution: coord**, from a Class-A ratchet audit by grok, with the scope corrected against #11424's diff.
**Small, mechanical, and it is the difference between #11183 being closable and one dead Class-A edge remaining baselined.**

## What #11424 leaves behind

#11424 deletes `eip8037_state_used_before_tx` — Program, relocs, symbol, fixture and MANIFEST row. But it **does not touch
the transport that fed it**. At #11424's own tip:

- `BlockVerdictGasGatePrelude.lean:37-42` still does
  `la t2, bv_bal_start; ld a1, 0(t2)` / `la t2, bv_bal_len; ld a2, 0(t2)` and then `jal ra, eip8037_tx_gas_gate`;
- `BlockVerdictGasGate.lean` still contains the `MV .x9 .x11` / `MV .x18 .x12` pair (2 occurrences) that saved those
  arguments into `s1`/`s2` for the deleted callee.

So the supplied BAL is still loaded, passed and saved — with **nothing left that reads it**. The consumer is gone; the
plumbing is not.

## Why it matters

1. **The ratchet still sees the edge.** `.Lbv_pregate_state_gas_ready` remains a baselined
   `bv_bal_start`/`bv_bal_len` path, so #11183's exit condition cannot be declared met while it exists, even though it now
   feeds nothing.
2. **It is the exact recurrence of a defect already fixed once.** #11405 removed six instructions that loaded the same two
   globals into `a4`/`a5` where nothing read them. This is the same shape on the `x9`/`x18` pair — and it only became dead
   when #11424 deleted the consumer. A dead transport left in place is what made `eip8037_tx_gas_gate` *look* like it
   touched the BAL twice, which cost three rounds of analysis tonight.
3. **A future reader cannot tell dead from live.** A gate that receives the supplied BAL as an argument reads as a Class-A
   consumer to every census, and each census then has to re-derive that nothing consumes it.

## The change

Remove, together: the two prelude loads and the argument passing at `GasGatePrelude.lean:37-42`, the `MV .x9 .x11` /
`MV .x18 .x12` save pair in the gate, and any relocation entries that go with them. Then `eip8037_tx_gas_gate` no longer
receives the supplied BAL at all.

[...]

PR #11437, opened 2026-08-04
fix(11428): drop dead supplied-BAL transport into eip8037_tx_gas_gate

pirapira:
## Summary
Closes the #11428 half of Class-A EXECUTION residual after #11424 deletes `eip8037_state_used_before_tx`.

- **Prelude**: remove `bv_bal_start`/`bv_bal_len` loads into a1/a2 before `jal eip8037_tx_gas_gate`.
- **Gate body**: remove `MV x9←x11` / `MV x18←x12` (BAL save pair). Confirmed on emitted body: x9/x18 appear only as callee-save SD/LD (no intermediate write; 11405-class recurrence check).
- **Class-A**: deliberate shrink of `.Lbv_pregate_state_gas_ready` load edges (2 rows). Ratchet OK 19 paths.
- **Layout**: textSizeBytes `0x59ae0`; region-map CLEAN; pins stable.

Branched from pr-11424 tip `ba9dc3c8c` (consumer already gone). Stack slots for x9/x18 retained as pure ABI preserve (no frame shrink / reloc churn).

## Gates
- `check-region-map`: CLEAN
- `check-bal-class-a-ratchet`: OK 19 baselined paths; no new edges; no silent shrink
- GasGate `#guard` length 560 (was 562)

## Test plan
- [ ] CI build green
- [ ] Class-A ratchet green on merge base with 11424


pirapira:
**Attribution: coord.** Reviewed against `origin/main`, register audit re-verified independently. **Labelled `merge!`.**
Closes #11428.

## The audit I required, confirmed from the emitted body

I asked for proof that `x9`/`x18` have no other consumer inside the gate **before** removing the saves — because a
save/restore pair is exactly how a live value hides from a local read, and these same two registers cost a full round trip
earlier tonight when they looked dead at the callsite and were live across +2124.

After the removal, `x9` and `x18` appear at **exactly four sites** in `BlockVerdictGasGate.lean`:

```
109:  .SD .x2 .x9  (16)      653:  .LD .x9  .x2 (16)
110:  .SD .x2 .x18 (24)      654:  .LD .x18 .x2 (24)
```

Entry save, exit restore, nothing else. So the callee-save discipline is intact and the two registers now preserve the
**caller's** values rather than transported BAL pointers. That is the right end state: the gate does not receive the supplied
BAL at all.

Removed together: the `MV .x9 .x11` / `MV .x18 .x12` pairs (both occurrences) and the prelude's
`la t2, bv_bal_start; ld a1` / `la t2, bv_bal_len; ld a2` loads plus the argument passing. `.text 0x59ae0`, region-map clean.

## The deliberate baseline shrink

Two rows gone — the `.Lbv_pregate_state_gas_ready` pair — with the ratchet reporting **OK at 19 paths**, down from 21. That
is the ratchet's second failure direction working as designed: a retirement must shrink the baseline *explicitly*, and the
file getting shorter is how progress is visible.

## Why this needed to exist at all, recorded because it is the instructive part

#11424 deleted `eip8037_state_used_before_tx` but **not the transport that fed it**, so the supplied BAL was still loaded,
passed and saved with nothing reading it. It was the exact recurrence of #11405's defect — six instructions loading the same
two globals into `a4`/`a5` where nothing read them — one register pair later, and it only became dead when #11424 removed the
consumer.

I initially told the author its ratchet finding was "already retired, pending merge". That was **half wrong**: the consumer
was queued, the transport was open, and its instrument was right where my summary-reading was not. Checking #11424's actual
diff scope rather than trusting its description is what settled it.

## Sequencing

Correctly stacked on `pr-11424` at `ba9dc3c8c`, since the transport is only dead once the consumer is gone. Once #11424
lands, this collapses to its own commit.

pirapira:
Queued behind #11438 (which batches #11433 + #11436 and is mid-CI). I checked this PR's merge against current `main`: the only conflicts are the generated files — `GuestAddrs.lean`, `RegionMapLinkPins.lean`, `symbol-addresses.tsv`, and `bal-class-a-baseline.tsv` — so it's mechanical, **no `complicated-merge` needed** and nothing for you to do.

The baseline conflict is expected: this PR retires the supplied-BAL transport into `eip8037_tx_gas_gate`, so the Class-A edge set shrinks. I'll resolve by regenerating and will review the lost edges individually before ratifying the shrink (the ratchet fails deliberately on disappearance, and I don't `--write-baseline` past *additions*).

Picking it up on the next pass once #11438 lands, with full local verification (`lake build` + all 20 gates).

Issue #11432, opened 2026-08-04
FA (traced): deposit route B appends a request for a REVERTED call — no receipt-status check, and the accept gate sees status zero


pirapira:
**Attribution: coord**, from a control-flow-split sweep by k3 (spec pin `e5a8caf1b`, guest `origin/main`). Ranked **HIGH**,
and it is the entry that names an **explicit FA channel**.

## The split

| | how deposit requests are derived |
|---|---|
| **spec** | `parse_deposit_requests` **scans receipts** (`requests.py:274`, called from `fork.py:957`) |
| **guest route A** | the capture path |
| **guest route B** | `blockVerdictDirectDepositFallback` (`BlockVerdictReceiptsTail.lean:102-104` → `DepositFallback.lean:121-141`) **synthesises deposits from calldata shape** when `c1_dlen == 0` and capture is incomplete |

## ⚠️ Route B has no receipt-status check

The spec derives a deposit only from a **receipt**, which exists only for a transaction that succeeded. Route B derives it from
the **shape of the calldata**, so:

- a **reverting** transaction with canonical deposit-shaped calldata yields a deposit in the guest that the spec does not
  produce → wrong `requests_hash` → **FR**;
- and in the other direction, a crafted header claiming that deposit is **accepted** by the guest → **FA channel**.

Observable through `requests_hash` → accept/reject.

The asymmetry matters: the FR half is visible and debuggable, but the FA half means an invalid block can be accepted because the
guest manufactured a request the spec would not have. That is the one direction the maintainer's land-with-FRs rule does not
cover.

## First question, before any fix

**Is route B reachable on a block the guest accepts?** It fires only when `c1_dlen == 0` *and* capture is incomplete, so the
first job is to determine whether those two conditions can hold on a block that then passes every other gate. If they cannot,
this is a documented structural divergence; if they can, it is a live FA and outranks the structural queue.

Note the shape of this question is identical to #11407's, which resolved as *unreachable-to-accept* because the routes that
trigger the fallback also set a status that rejects. **Look for the same structure here before assuming the worst or the best**:
find what else is true on a block where `c1_dlen == 0` and capture is incomplete.

## The fix, if reachable

Add the receipt-status condition the spec has — a deposit is derived only from a **successful** transaction's receipt. Do
**not** widen the shape match or add a compensating check elsewhere: the divergence is that route B uses a different *source*
(calldata) than the spec (receipts), and matching the source is the convergent fix.

[...]

PR #11440, opened 2026-08-04
fix(codegen): gate direct deposit fallback on receipt status

pirapira:
Fix direct deposit fallback to honor reverted receipt status

## Summary

Route B derives EIP-6110 requests from canonical transaction calldata only
when captured receipt status is successful. A reverted canonical-shape call
must not synthesize a request merely because runtime dispatch status remains
zero; the receipt status is stored separately in `bv_tx_status_arr`.

Route B remains in place for blocks where runtime log capture is incomplete.
This change only gates its append helper on the per-transaction receipt status.

## Validation

- `lake build EvmAsm.Codegen.Programs.BlockVerdictDepositFallback`
- Candidate and origin stateless guests both link successfully.
- Constructed two-transaction Amsterdam witness: tx0 is a canonical deposit
  call with receipt status 0 due to the gas floor, tx1 is an ordinary successful
  call, and the honest header omits requests. Origin rejects and candidate
  accepts, matching the execution-specs reference.
- The companion header claims the request synthesized from tx0 calldata.
  Origin accepts and candidate rejects.

The trigger/reject distinction is intentional: `bv_dispatch_runtime_status`
stays zero on the reverted path, while `bv_tx_status_arr` records receipt
status. The fallback previously did not read the latter.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdictDepositFallback.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictDepositFallback.lean
@@ -128,6 +128,12 @@ def blockVerdictDirectDepositFallback : String :=
   ".Lbv_deposit_direct_loop:\n" ++
   "  beq s3, s4, .Lbv_deposit_direct_done\n" ++
   "  la a0, bv_mtx_skip_ctx; mv a1, s3; jal ra, multi_tx_nth_context\n" ++
+  -- Direct request derivation is a fallback for missing runtime deposit logs,
+  -- not an independent transaction executor.  A reverted transaction can
+  -- retain the canonical calldata shape while its receipt status is zero;
+  -- execution-specs derives EIP-6110 requests from successful receipt logs,
+  -- so do not synthesize a request for that transaction.
+  "  slli t0, s3, 3; la t1, bv_tx_status_arr; add t1, t1, t0; ld t1, 0(t1); beqz t1, .Lbv_deposit_direct_noappend\n" ++
   "  la a0, bv_mtx_skip_ctx; mv a1, s2; mv a2, s5; jal ra, block_verdict_append_direct_deposit\n" ++
   "  mv s2, a1\n" ++
   "  beqz a0, .Lbv_deposit_direct_noappend\n" ++


pirapira:
**Attribution: coord.** Reviewed against `origin/main`. **Labelled `merge!`.** Closes the traced FA in #11432.

## The fix is one guard, and it is the spec's condition

```
slli t0, s3, 3; la t1, bv_tx_status_arr; add t1, t1, t0; ld t1, 0(t1); beqz t1, .Lbv_deposit_direct_noappend
```

A deposit request is appended only when that transaction's receipt status is non-zero. `requests.py:274-284` scans **receipt
logs** and appends only valid deposit events, and `fork.py:956-959` appends only when that scan is non-empty — so the spec
derives a deposit only from a transaction that **succeeded**. This adds exactly that condition and nothing else.

The comment states the mechanism rather than the change: *"A reverted transaction can retain the canonical calldata shape while
its receipt status is zero."* That is the whole defect in one sentence.

## Both directions demonstrated, which is the bar for an FA

- **FA closed**: the crafted header claiming a request synthesised from tx0 calldata — **origin accepts, candidate rejects.**
- **FR cleared**: the honest block whose header omits the request — **origin rejects, candidate accepts**, matching the
  reference.

Constructed fixtures for each direction, and *both* flips are the proof. A traced path is a hypothesis until a fixture runs it,
and here it ran both ways — the same standard #11426's 2935-drop fixture set earlier today.

## Why this was reachable where #11407's fallback was not

Worth keeping, because the same question was answered twice tonight with opposite results and the distinguishing property is
subtle. On #11407 the routes that triggered the fallback **also** set `bv_dispatch_runtime_status` non-zero, so the receipts
tail rejected and the FA-lax arithmetic could never decide accept-versus-reject. Here **trigger and reject are independent**: a
reverted call leaves `bv_dispatch_runtime_status` **zero** (only a *dispatch* failure sets it) while the receipt status lives in
`bv_tx_status_arr`, which the append helper never read. That independence is exactly what made this channel real, and the PR
body says so.

## Residual, recorded not fixed

After this, Route B is **safe without being correct.** Route A is `parse_deposit_requests` over *captured* block logs, and the
spec always has the receipt logs — so the only reason a canonical direct call reaches Route B at all is that the guest **failed
to capture a log the spec would have**. The convergent end state is to fix the capture and delete Route B; a second derivation
path for one spec operation is #11423's shape, and this is the fourth instance of it today. A reader seeing only the status gate
would conclude Route B is now right, when what we have is Route B being harmless.

No random-200 in the body. Acceptable here — the change adds a rejection condition on a fallback path, so its direction is
strictly toward rejecting more, and both discriminating fixtures are exercised. If a sweep was run, state the number; if not,
say so rather than leaving it inferred.

pirapira:

PR #11441, opened 2026-08-04
proof(#11351): forward construction of rlp_list_nth_item's Success

dhsorens:
**Stacked on #11434** (base branch is `proof-11351-header-number`, so the diff shows only this layer). #11434 built the per-item forward bridge; this composes it up to `Success` — the predicate that `rlp_field_to_u64`'s already-`.proven` triple consumes.

`Success` / `StrictListPayload` / `StrictNthItem` had **no constructor driven by a model-side decode**; every existing lemma consumed one. That is the shared prerequisite behind #11351, #11345 and #11346.

## Core side (`Rv64/RLP/ItemDecodeForward.lean`)

| lemma | role |
|---|---|
| `sub_base_of_base_add` | recovers an offset from its cursor — the form `StrictNthItem.succ` needs to re-enter the chain at `(next - base).toNat` |
| `decodeAux_bytes_advance` | each byte-string decode strictly advances |
| `DecodeChain.le_of_bytes` | a chain never runs past its own end offset |

⚠️ **One design point worth a reviewer's eye.** There is no general *"the `decodeAux` residue is a proper suffix"* lemma anywhere in the tree, so `decodeAux_bytes_advance` re-runs the prefix case split — but only to read the offset off, none of the `rlpItemDecode` construction is repeated. It deliberately requires `off' ≤ bytes.length` **alone**, not the window bound `off' ≤ endOff` that the main dispatcher takes. That is not an oversight: the window bound is precisely what the chain induction is trying to establish, so assuming it here would be circular. If a general suffix lemma is wanted instead, say so and I will add it to `EL/RLP/Properties.lean` and collapse this one.

## Codegen side (`Codegen/Programs/RlpListNthItemForward.lean`, new)

- `strictNthItem_of_chain` — induction on the index, generalising the item list and start offset: the head decode supplies one `rlpItemDecode`, the tail supplies the rest of the chain.
- `strictListPayload_long_forward` / `strictListPayload_short_forward` — the outer header. A block header is ~500 B, so the **long** form (`0xF9`, two length bytes) is what #11351 needs; the short form comes free and is what keeps the non-vacuity check small.
- `success_forward` — assembles all three.

**Layering.** The Codegen half cannot sit beside its core half: `check-layering.sh` L1 forbids core importing Codegen, and `StrictNthItem` is a Codegen-layer definition. Same split as the #11408 determinism prerequisite.

## Non-vacuity

An end-to-end `example` drives `c4 83 01 02 03` — a four-byte-payload list holding one three-byte string — through the header lemma, the chain, and `success_forward`. So the composition is demonstrably instantiable, not merely consistent.

## Registry

**No rows.** Pure lemma modules, to be cited and consumed by the triple itself, per "availability is not use".

## Gates

`lake build` warning-free · `port-check` **PASS** (12 theorems here, classical axioms only) · `check-axioms` OK (127 witnesses) · forbidden-tactics / layering / file-size **PASS**.

## What remains for #11351

Compose with `rlpFieldToU64_flat_spec_within` (already `.proven`) and the 8-instruction `headerExtractNumber_prog` wrapper, then registry rows. The **value** bridge is definitional: `Result.success` yields `BitVec.ofNat 64 (Nat.fromBytesBE …)` and `SpecRef.bytesBEtoNat` is an `abbrev` for `EL.RLP.Nat.fromBytesBE`, so there is no arithmetic lemma to invent. Grading will be `.domainRestricted`, not `.agrees`.

Refs #11351, #11345, #11346

### DIFF
--- a/EvmAsm/Codegen/Programs/Imports.lean
+++ b/EvmAsm/Codegen/Programs/Imports.lean
@@ -73,6 +73,7 @@ import EvmAsm.Codegen.Programs.TxTypeDispatchTop
 import EvmAsm.Codegen.Programs.TxTypeDispatchDischarge
 import EvmAsm.Codegen.Programs.RlpRead
 import EvmAsm.Codegen.Programs.BytesToNibblesBridge
+import EvmAsm.Codegen.Programs.RlpListNthItemForward
 import EvmAsm.Codegen.Programs.RlpWalkDeterminism
 import EvmAsm.Codegen.Programs.RlpListNthItemCallSAsm
 import EvmAsm.Codegen.Programs.RlpWalkCallSAsm
--- a/EvmAsm/Codegen/Programs/RlpListNthItemForward.lean
+++ b/EvmAsm/Codegen/Programs/RlpListNthItemForward.lean
@@ -1,0 +1,173 @@
+/-
+  EvmAsm.Codegen.Programs.RlpListNthItemForward
+
+  Forward (model → guest) construction of `rlp_list_nth_item`'s success
+  predicate.  `Success` / `StrictListPayload` / `StrictNthItem` previously had
+  **no constructor at all** driven by a model-side decode — every lemma about
+  them consumed one.  That gap is the shared prerequisite behind #11351,
+  #11345 and #11346.
+
+  Lives under `Codegen` rather than beside `Rv64/RLP/ItemDecodeForward.lean`
+  because `StrictNthItem` and friends are Codegen-layer definitions and
+  `scripts/check-layering.sh` L1 forbids core importing Codegen — the same
+  split used for the determinism prerequisite in #11408.
+
+  Scope is byte-string children, inherited from `ItemDecodeForward`: the
+  converse direction is false for nested lists (`c3 c2 81 00`), and every RLP
+  field of a block header is a byte string.
+-/
+
+import EvmAsm.Codegen.Programs.RlpListNthItemSAsmBase
+import EvmAsm.Rv64.RLP.ItemDecodeForward
+
+namespace EvmAsm.Codegen.RlpListNthItemSAsm
+
+open EvmAsm.Rv64 EvmAsm.Rv64.RLP EvmAsm.EL.RLP
+
+/-! ## The selected child -/
+
+/-- From a byte-string `DecodeChain` covering the list payload, the guest's
+    `StrictNthItem` holds for whichever child the model put at `index`.
+
+    Induction is on `index`, generalising the item list and the start offset:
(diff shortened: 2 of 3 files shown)


pirapira:

## Verified

- `lake build EvmAsm.Codegen.Programs.RlpListNthItemForward` → **Built** (1035 jobs), including its `example`, and it builds
  `ItemDecodeForward` on the way as expected.
- `#print axioms success_forward` → `[propext, Classical.choice, Quot.sound]`. Classical only; no `sorry`, no
  `bv_decide`/`native_decide`.

## The gap it closes is stated exactly, and it is the right kind of gap

> *"`Success` / `StrictListPayload` / `StrictNthItem` previously had **no constructor at all** driven by a model-side decode —
> every lemma about them consumed one."*

That is precise and it is the useful framing: a predicate with only eliminators is unusable in the forward direction, so every
consumer had to be handed one from somewhere else. `success_forward`, `strictNthItem_of_chain` and the two
`strictListPayload_*_forward` lemmas supply the missing introduction, and the docstring names all three tickets it unblocks
(#11351, #11345, #11346) rather than just its own.

## Layering justified, consistent with the precedent it cites

Living under `Codegen` because `StrictNthItem` and friends are Codegen-layer definitions and `check-layering.sh` L1 forbids core
importing Codegen — the same split used for #11408's determinism prerequisite. Both halves of the bridge therefore sit on the
side of the boundary their own definitions live on, which is the only arrangement that respects L1.

## Scope inherited, not re-argued

Byte-string children only, inherited from `ItemDecodeForward`, with the reason carried forward in one line: the converse is
false for nested lists (`c3 c2 81 00`) and every RLP header field is a byte string. That counterexample has now been cited
consistently across #11416, #11434 and this PR, which is what makes the restriction a known boundary rather than three
independent hedges.

## Stacking

This carries `faa4a2c01`, i.e. **#11434's commit**, beneath its own — same commit, so it collapses when #11434 lands. Both are
labelled, and #11434 should merge first; if they land out of order the diff simply resolves to the same tree.

## Where this leaves #11351

With the guest → model direction (`WalkDecodeBridge`), the model → guest direction (#11434), arity-N `decodeItems` composition
(#11425, merged) and now the forward constructor for `Success`, the header-extractor family has the full round trip it needs.
The remaining work there is consuming these, not proving more.

pirapira:

pirapira:
Landed via #11442. Verified: commit `19e14251a` is an ancestor of `main` as of `702e1e45f` — closed rather than dropped. It went in as the child half of the #11434 stack, merged base-then-child in dependency order.

PR #11443, opened 2026-08-04
proof(#11351): decodeItems to DecodeChain, via a shared residue lemma

dhsorens:
**Stacked on #11441** (base is `proof-11351-success-forward`, so the diff shows only this layer). Last core-side piece before the header composition: turning the model's `decodeItems` result into the offset-indexed `DecodeChain` that the guest-side walk composition consumes.

## `decodeAux_bytes_residue` — one case split, three facts

This **generalises** the prefix case split that #11441's `decodeAux_bytes_advance` performed, rather than adding a second copy of it. `advance` is now a two-line corollary, so the net effect on the tree is *less* duplication, not more. (I had started this as a refactor of #11441 itself, then backed it out when that PR was approved — no point discarding a local build + axiom verification for a code-quality change that lands just as well on top.)

From one prefix analysis:

| conclusion | why it is needed |
|---|---|
| residue is `bytes.drop off'` for a **derived** `off'` | `decodeItems`' recursion hands back an arbitrary suffix, while `DecodeChain` is indexed by *offsets* — this is the bridge between the two |
| the decode strictly advances | chain-induction termination |
| the decode is **fuel-independent** (`∀ m`) | none of the three byte-string branches recurses, so their branch equations are already fuel-free — and `DecodeChain` demands exactly that `∀ m` form, which a single-fuel fact cannot supply |

⚠️ It deliberately requires `off' ≤ bytes.length` **alone**, not the window bound `off' ≤ endOff` that `rlpItemDecode_of_decodeAux_bytes` takes. The window bound is what the chain induction *establishes*, so assuming it here would be circular. That asymmetry is the reason these are two lemmas and not one.

## `decodeItems_to_chain`

The converse of #11425's `decodeItems_of_chain`.

Worth noting for anyone touching this later: my first attempt unfolded `decodeItems` directly and hit repeated dependent-elimination failures, because the desugared `do`-bind is a **two-discriminant match**, which neither `Option.bind_eq_some_iff` nor `cases h : …` can see through. The working version reuses the existing `decodeItems_cons_inv` and `decodeItems_succ_of_ne_nil` — which already solved this exact problem — and is roughly a third the size. The lemma I needed was already in the tree.

## Gates

`lake build` warning-free · `port-check` **PASS** (101 theorems, classical axioms only) · forbidden-tactics / layering / file-size **PASS**.

## Remaining for #11351

Header-specific composition: `_decode_header = .ok h` → `StrictListPayload` + `DecodeChain` → `success_forward` → compose with `rlpFieldToU64_flat_spec_within` → the 8-instruction wrapper → registry rows.

One thing that got *easier* since the last PR, now posted on #11351: `_decode_header` runs `items.mapM rlpBytes?`, and `rlpBytes?` sends `.list` to `none`. So the byte-string hypothesis these lemmas carry is **discharged for free** by `_decode_header = .ok h` — it is not an extra domain restriction on the row, and the #11434 scope limit costs this family literally nothing.

Refs #11351, #11345, #11346

### DIFF
--- a/EvmAsm/Rv64/RLP/ItemDecodeForward.lean
+++ b/EvmAsm/Rv64/RLP/ItemDecodeForward.lean
@@ -500,54 +500,65 @@ theorem rlpItemDecode_of_decodeAux_bytes
                     obtain ⟨items, leftover⟩ := ir
                     simp [hread, hshort, htake, hitems] at hdec
 
-/-! ## Offset advance
+/-! ## Residue shape, offset advance, and fuel independence
 
-Chaining items needs to know each decode *advances*.  There is no general
-"the `decodeAux` residue is a proper suffix" lemma in the tree, so this
-re-runs the prefix case split, but only to read off the offset — none of the
-`rlpItemDecode` construction is repeated.  It needs `off' ≤ bytes.length`
-alone, **not** the window bound, which is what keeps it usable as the
-termination fact in the chain induction (the window bound is what the
-induction is trying to establish). -/
+Three facts the chain layer needs, all read off the same prefix case split, so
+it is performed once here and shared.  (#11441 derived only the advance fact
+and did its own split; this generalises that proof rather than adding a second
+copy — `decodeAux_bytes_advance` is now a two-line corollary.)
 
-theorem decodeAux_bytes_advance
-    (bytes : List Byte) (off off' n : Nat) (p : List Byte)
-    (hdec : decodeAux (n + 1) (bytes.drop off) = some (.bytes p, bytes.drop off'))
-    (hoff'len : off' ≤ bytes.length) :
-    off < off' := by
+* the residue is itself `bytes.drop off'` for a **derived** `off'` — needed to
+  turn `decodeItems`' arbitrary residues into `DecodeChain`'s offset-indexed
+  ones;
+* the decode strictly advances — the termination fact for the chain induction;
+* the decode is **fuel-independent**, i.e. holds at every `m + 1`.  None of the
+  three byte-string branches recurses, so their branch equations are already
+  fuel-free; `DecodeChain` demands exactly this `∀ m` form.
+
+⚠️ `off'` is *concluded*, not assumed, and the only bound produced is
+`off' ≤ bytes.length` — deliberately **not** the window bound `off' ≤ endOff`
+that `rlpItemDecode_of_decodeAux_bytes` takes.  The window bound is what the
+chain induction is trying to establish, so assuming it here would be circular. -/
+
+theorem decodeAux_bytes_residue
+    (bytes : List Byte) (off n : Nat) (p r : List Byte)
+    (hdec : decodeAux (n + 1) (bytes.drop off) = some (.bytes p, r)) :
+    ∃ off', r = bytes.drop off' ∧ off < off' ∧ off' ≤ bytes.length ∧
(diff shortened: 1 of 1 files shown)


pirapira:

These two complete a four-PR stack (`faa4a2c01` #11434 → `19e14251a` #11441 → `1c4d6d04b` #11443 → `f3c9db66b` #11444), so
I reviewed them together and built the top, which builds all four.

## Verified

- `lake build EvmAsm.Codegen.Programs.RlpDecodeFullyForward` → **Built** (1036 jobs), including every `example` in the stack.
- `#print axioms`: `success_of_decodeFully_list` and `decodeItems_to_chain` → `[propext, Classical.choice, Quot.sound]`.
  Classical only, no `sorryAx`, and nothing in either diff matches `bv_decide`/`native_decide`.

## #11443 — `decodeItems → DecodeChain`

The converse of #11425's `decodeItems_of_chain`, which merged earlier today. So the chain predicate is now an
**iff in substance**: a decode chain composes to a `decodeItems` run, and a successful run decomposes back into one. That
matters more than either direction alone, because a consumer holding either shape can now reach the other, and the shared
residue lemmas (`decodeAux_bytes_residue`, `decodeAux_bytes_advance`) are what stop the two directions drifting into
subtly different notions of "advance".

`DecodeChain.le_of_bytes` is the small piece that makes the induction go: monotonicity of the offsets is exactly what the
composition direction assumed implicitly and the decomposition direction has to produce.

## #11444 — `decodeFully → Success`

The capstone: from a full model decode of a list, construct the guest's `Success` predicate. With
`listPayload_chain_of_decodeFully` bridging to the chain and `success_of_decodeFully_list` on top, the forward stack is
complete.

## What the stack now gives #11351

Both directions, end to end:

| direction | provided by |
|---|---|
| guest → model, item level | `WalkDecodeBridge` (in main) |
| guest → model, list composition | `decodeItems_of_chain` (#11425, merged) |
| model → guest, item level | #11434 |
| model → guest, `Success` constructor | #11441 |
| model → guest, list decomposition | #11443 |
| model → guest, `decodeFully` capstone | #11444 |

The header-extractor family needed the round trip; it now has it, and the remaining work on #11351 is **consuming** these
rather than proving more.

## Scope, consistently carried

Byte-string children throughout, with the same reason each time: the converse is false for nested lists, witnessed by
`c3 c2 81 00`, and every RLP field of a block header is a byte string. That counterexample has now been cited across
#11416, #11434, #11441 and this pair — four PRs agreeing on where the correspondence genuinely fails, which is what turns a
restriction into a documented boundary rather than four independent hedges.

## Note on the stack

Each PR is separately labelled and they must land in commit order; landing out of order simply resolves to the same tree,
since each carries its predecessors' commits. #11442 already batches #11434 + #11440 + #11441, so #11443 and #11444 follow
that batch.

pirapira:
Retargeted the base from `proof-11351-success-forward` to `main` — flagging since it's PR metadata you set, and it's reversible.

Reason: #11441 landed via #11442, so `proof-11351-success-forward` is now **fully contained in `main`** (`git merge-base --is-ancestor` confirms). The PR was pointing at a branch that no longer represents a line of development, so if auto-merge had fired it would have landed there rather than on `main`.

**#11444 is left stacked on this PR's branch** — that base (`proof-11351-decodeitems-chain`) is still live, so its stack is intact and correct. I'll merge the two base-then-child in dependency order, the same way #11434 → #11441 went in.

pirapira:

PR #11444, opened 2026-08-04
proof(#11351): decodeFully to Success, completing the forward stack

dhsorens:
**Stacked on #11443.** Closes the model → guest direction end to end: from `decodeFully bytes = some (.list items)` with byte-string children, produce the guest's `Success` for whichever child index the model gives.

| lemma | role |
|---|---|
| `listPayload_chain_of_decodeFully` | inverts the whole-input decode into the two things `success_forward` wants — `StrictListPayload` for the outer header (short **and** long forms) and the offset-indexed `DecodeChain` for the children. The three byte-string prefix classes are refuted from the `.list` result constructor. |
| `success_of_decodeFully_list` | the capstone the header extractors consume; for `header_extract_number`, `index = 8` |

## Two things fell out while proving it

**1. `listPayload_chain_of_decodeFully` needs no no-overflow hypothesis.** I had carried `base.toNat + bytes.length < 2^64` on it by analogy with the item layer. The linter caught it as unused, and on inspection it genuinely is: neither `StrictListPayload` nor `DecodeChain` constrains addresses — only the later `rlpItemDecode` construction does. Dropped, rather than left in as a dead premise that would make the lemma look narrower than it is. Same for `hk` on the `takeBytes` helper.

**2. The `_decode_header` side needs no byte-string side condition either.** `success_forward` takes "every child is a byte string" as a hypothesis, and I expected to have to record it as a domain restriction on the row. It is already enforced by the port:

```lean
match items.mapM rlpBytes? with       -- Stateless.lean:78
| none => .error .headerDecodeError
| some bs => …
```

`rlpBytes?` sends `.list` to `none`, so `_decode_header = .ok h` **implies** it. The #11434 scope restriction therefore costs the header-extractor family nothing at all — a header containing a nested list is not a header as far as the port is concerned. Posted on #11351 with the citation.

## Non-vacuity

An `example` drives `c4 83 01 02 03` from `decodeFully` all the way to `Success`, with the model decode discharged by **`rfl`** — the decode is computed by the model, not assumed. That is a stronger check than the #11441 example, which hand-supplied the chain.

## Gates

`lake build` warning-free · `port-check` **PASS** (17 theorems, classical axioms only) · forbidden-tactics / layering / file-size **PASS**.

## Stack state

#11434 + #11441 are in batch #11442; #11443 and this PR sit on top. Once the batch lands I will retarget and rebase the remainder onto `main`.

## What remains for #11351

Only the machine side now: compose `success_of_decodeFully_list` with `rlpFieldToU64_flat_spec_within` (already `.proven`), wrap the 8-instruction `headerExtractNumber_prog`, and add the registry rows. The value bridge is definitional — `Result.success` yields `BitVec.ofNat 64 (Nat.fromBytesBE …)` and `SpecRef.bytesBEtoNat` is an `abbrev` for `EL.RLP.Nat.fromBytesBE`. Grading `.domainRestricted`, not `.agrees`, since the guest never checks header arity (23 current fork / 21 previous).

Refs #11351, #11345, #11346

### DIFF
--- a/EvmAsm/Codegen/Programs/Imports.lean
+++ b/EvmAsm/Codegen/Programs/Imports.lean
@@ -73,6 +73,7 @@ import EvmAsm.Codegen.Programs.TxTypeDispatchTop
 import EvmAsm.Codegen.Programs.TxTypeDispatchDischarge
 import EvmAsm.Codegen.Programs.RlpRead
 import EvmAsm.Codegen.Programs.BytesToNibblesBridge
+import EvmAsm.Codegen.Programs.RlpDecodeFullyForward
 import EvmAsm.Codegen.Programs.RlpListNthItemForward
 import EvmAsm.Codegen.Programs.RlpWalkDeterminism
 import EvmAsm.Codegen.Programs.RlpListNthItemCallSAsm
--- a/EvmAsm/Codegen/Programs/RlpDecodeFullyForward.lean
+++ b/EvmAsm/Codegen/Programs/RlpDecodeFullyForward.lean
@@ -1,0 +1,190 @@
+/-
+  EvmAsm.Codegen.Programs.RlpDecodeFullyForward
+
+  Top of the forward (model → guest) stack: from a whole-input model decode
+  `decodeFully bytes = some (.list items)`, produce the two ingredients
+  `success_forward` needs — the guest's `StrictListPayload` for the outer
+  header and the offset-indexed `DecodeChain` for its children.
+
+  Kept separate from `RlpListNthItemForward.lean` so the list-header inversion
+  (which reasons about `decodeFully`/`decodeListPayload`) does not enlarge the
+  child-composition module.
+
+  Children must be byte strings, inherited from `ItemDecodeForward`.  For the
+  header-extractor family that costs nothing and is in fact **free**:
+  `_decode_header` runs `items.mapM rlpBytes?`, and `rlpBytes?` sends `.list`
+  to `none`, so a successful header decode already implies it.
+-/
+
+import EvmAsm.Codegen.Programs.RlpListNthItemForward
+
+namespace EvmAsm.Codegen.RlpListNthItemSAsm
+
+open EvmAsm.Rv64 EvmAsm.Rv64.RLP EvmAsm.EL.RLP
+
+/-- A `takeBytes` that consumes the rest of a suffix pins the split point. -/
+private theorem takeBytes_all_of_nil {bytes : List Byte} {k n : Nat}
+    {payload : List Byte}
+    (h : takeBytes (bytes.drop k) n = some (payload, [])) :
+    n = bytes.length - k ∧ payload = bytes.drop k := by
+  obtain ⟨hcat, hlen⟩ := takeBytes_eq_some_imp h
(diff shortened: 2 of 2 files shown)


pirapira:

These two complete a four-PR stack (`faa4a2c01` #11434 → `19e14251a` #11441 → `1c4d6d04b` #11443 → `f3c9db66b` #11444), so
I reviewed them together and built the top, which builds all four.

## Verified

- `lake build EvmAsm.Codegen.Programs.RlpDecodeFullyForward` → **Built** (1036 jobs), including every `example` in the stack.
- `#print axioms`: `success_of_decodeFully_list` and `decodeItems_to_chain` → `[propext, Classical.choice, Quot.sound]`.
  Classical only, no `sorryAx`, and nothing in either diff matches `bv_decide`/`native_decide`.

## #11443 — `decodeItems → DecodeChain`

The converse of #11425's `decodeItems_of_chain`, which merged earlier today. So the chain predicate is now an
**iff in substance**: a decode chain composes to a `decodeItems` run, and a successful run decomposes back into one. That
matters more than either direction alone, because a consumer holding either shape can now reach the other, and the shared
residue lemmas (`decodeAux_bytes_residue`, `decodeAux_bytes_advance`) are what stop the two directions drifting into
subtly different notions of "advance".

`DecodeChain.le_of_bytes` is the small piece that makes the induction go: monotonicity of the offsets is exactly what the
composition direction assumed implicitly and the decomposition direction has to produce.

## #11444 — `decodeFully → Success`

The capstone: from a full model decode of a list, construct the guest's `Success` predicate. With
`listPayload_chain_of_decodeFully` bridging to the chain and `success_of_decodeFully_list` on top, the forward stack is
complete.

## What the stack now gives #11351

Both directions, end to end:

| direction | provided by |
|---|---|
| guest → model, item level | `WalkDecodeBridge` (in main) |
| guest → model, list composition | `decodeItems_of_chain` (#11425, merged) |
| model → guest, item level | #11434 |
| model → guest, `Success` constructor | #11441 |
| model → guest, list decomposition | #11443 |
| model → guest, `decodeFully` capstone | #11444 |

The header-extractor family needed the round trip; it now has it, and the remaining work on #11351 is **consuming** these
rather than proving more.

## Scope, consistently carried

Byte-string children throughout, with the same reason each time: the converse is false for nested lists, witnessed by
`c3 c2 81 00`, and every RLP field of a block header is a byte string. That counterexample has now been cited across
#11416, #11434, #11441 and this pair — four PRs agreeing on where the correspondence genuinely fails, which is what turns a
restriction into a documented boundary rather than four independent hedges.

## Note on the stack

Each PR is separately labelled and they must land in commit order; landing out of order simply resolves to the same tree,
since each carries its predecessors' commits. #11442 already batches #11434 + #11440 + #11441, so #11443 and #11444 follow
that batch.

pirapira:

pirapira:
Landed via #11447. Verified: commit `f3c9db66b` is an ancestor of `main` as of `42d3dfaa4` — closed rather than dropped. It went in as the child of #11443, merged base-then-child, completing the #11351 forward stack (header-number → success-forward → decodeitems-chain → decodefully-success).

PR #11445, opened 2026-08-04
fix(codegen): follow delegation markers from AccountState

pirapira:
The delegation marker must be followed, not executed. execution-specs eoa_delegation.py:40 defines a valid delegation by the EIP-7702 marker prefix plus the 20-byte address (length 23), and eoa_delegation.py:62 extracts the delegated code address. Recognition is a property of the code bytes, independent of which state store supplied them. The old BAL path worked by accident: it missed the AccountState-only marker, so it never had a marker to execute.

This change checks AccountState status 1 at the existing .Lcd_acst_done path, recognizes len=23 and bytes ef0100, and re-enters the existing .Lcd_header_lookup marker path. There is one marker-following implementation rather than two that can drift. Status-2 empty/precompile bails remain intact.

Evidence:
- Four call_into_chain fixtures (CALL, CALLCODE, DELEGATECALL, STATICCALL): 4/4 exact, 256/256, succ=1, fail=0.
- Status-2 control set_code_to_precompile: 20/20 exact.
- Random-200 against the true pre-switch parent 1cd6b60d: parent 200/200 ran, FR=4; candidate bc84eb7b 196/200 exact, FR=4; identical four FR rows (00123, 00101, 00118, 00195), no base-PASS to candidate-FR transitions and no newly failing rows.
- Candidate artifact: /tmp/11406-marker-fix.elf. The candidate and measurement artifacts are recorded in the investigation.

EIP-3541 reserves 0xEF as an invalid opcode; executing the marker produced root matches but rejection and misdirected the original diagnosis. The investigation tested seven candidates and refuted four hypotheses. The AccountState reset correction was split into #11436, which has landed.

No merge label is applied by this PR.

### DIFF
--- a/EvmAsm/Codegen/Programs/BalCodePreimages.lean
+++ b/EvmAsm/Codegen/Programs/BalCodePreimages.lean
@@ -1068,10 +1068,21 @@ def balCodePreimagesValidFunction : String :=
   ".Lasd_pc_low16:\n" ++
   "  lbu t3, 18(t0); lbu t4, 19(t0); slli t3, t3, 8; or t3, t3, t4; li t4, 1; bltu t3, t4, .Lasd_not_precompile; li t4, 17; bgeu t4, t3, .Lasd_empty; li t4, 256; beq t3, t4, .Lasd_empty\n" ++
   ".Lasd_not_precompile:\n" ++
+  -- A same-transaction authorization can install the delegated target's
+  -- marker after the header witness was fixed.  Consult AccountState first:
+  -- the header lookup below is only the pre-state fallback, while the old
+  -- CodeState fallback cannot see an AccountState-only authorization row.
+  "  la a0, bsbd_deleg_target; jal ra, account_state_lookup_current\n" ++
+  "  li t0, 1; beq a0, t0, .Lasd_account_state_code\n" ++
+  "  li t0, 2; beq a0, t0, .Lasd_empty\n" ++
+  "  li t0, 3; beq a0, t0, .Lasd_empty\n" ++
   "  la t0, sv_pre_rlp_ptr; ld a0, 0(t0); la t0, sv_pre_rlp_len; ld a1, 0(t0); la a2, bsbd_deleg_target; mv a3, s1; mv a4, s2; la t0, svf_codes_ptr; ld a5, 0(t0); la t0, svf_codes_len; ld a6, 0(t0); jal ra, code_at_header_state_root\n" ++
   "  beqz a0, .Lasd_resolved\n" ++
   "  la a0, bsbd_deleg_target; jal ra, code_state_lookup_current\n" ++
   "  li t0, 1; bne a0, t0, .Lasd_empty\n" ++
+  "  ld t1, 96(sp); sub t1, a1, t1; la t0, cahsr_code_offset; sd t1, 0(t0); la t0, cahsr_code_length; sd a2, 0(t0)\n" ++
+  "  j .Lasd_resolved\n" ++
+  ".Lasd_account_state_code:\n" ++
   "  ld t1, 96(sp); sub t1, a1, t1; la t0, cahsr_code_offset; sd t1, 0(t0); la t0, cahsr_code_length; sd a2, 0(t0)\n" ++
   ".Lasd_resolved:\n" ++
   "  li a0, 0; j .Lasd_ret\n" ++
--- a/EvmAsm/Codegen/Programs/BalCodePreimages.lean
+++ b/EvmAsm/Codegen/Programs/BalCodePreimages.lean
@@ -1026,12 +1026,11 @@ def balCodePreimagesValidFunction : String :=
   "  addi sp, sp, 80\n" ++
   "  ret\n" ++
   "\n" ++
-  "# Resolve the delegation marker from the AccountState overlay.  The former\n" ++
-  "# The former BAL walker is intentionally not used by consumers after the AccountState\n" ++
-  "# producer lands.  The return codes retain the caller protocol: 0 means the\n" ++
-  "# current marker was resolved and cahsr_* names its target code, 1 means no\n" ++
-  "# current marker (caller may use the authenticated header), and 2 means an\n" ++
-  "# empty/deleted or precompile target.\n" ++
+  "# Resolve an EIP-7702 delegation marker from the AccountState overlay.\n" ++
+  "# a0 = 20-byte authority address, a1/a2 = witness state ptr/len,\n" ++
+  "# a3 = access mode (0 charge, 1 free/warm, 2 probe), a4 = codes base.\n" ++
+  "# Return 0 for a resolved delegated code target, 1 for no current marker,\n" ++
+  "# and 2 for empty/deleted or precompile targets.\n" ++
   "account_state_delegation_code_resolve:\n" ++
   "  addi sp, sp, -128\n" ++
   "  sd ra, 0(sp); sd s0, 8(sp); sd s1, 16(sp); sd s2, 24(sp); sd s3, 32(sp)\n" ++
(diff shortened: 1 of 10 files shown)


pirapira:
**Attribution: coord.** Reviewed at `30e599d47`. **Labelled `merge!`** — the change and its evidence are sound. Two notes for
the rebase, one of which needs an action.

## The shape is better than what I had asked for

I pushed for a source swap *inside* the existing helper and warned against an adapter, because a second resolver duplicates
warm/cold accounting and an early adapter attempt regressed 00001. What landed is neither: **`bal_same_block_delegation_code_resolve`
is fully removed** (0 definitions, 0 `jal` sites) and replaced by `account_state_delegation_code_resolve` at 6 source sites —
so there is **exactly one** delegation-resolution implementation, not two that can drift. Net **−138 lines** (131 insertions,
269 deletions), which is what dropping the RLP walk over the supplied BAL buys.

## The root cause, and why it took seven candidates

The generic code lookup **finds the 23-byte `ef0100` marker and the handler treated it as ordinary code**, bypassing delegation
resolution — so the marker was *executed* rather than *followed*. The spec does the same lookup and then recognises the
designator: `is_valid_delegation` (`vm/eoa_delegation.py:40`), `get_delegated_code_address` (`:62`). **Recognition is a property
of the code bytes, not of which store they came from**, which is why the old BAL path worked by accident — it *missed* in
CodeState and so never had a marker to mistake for code.

And the reason the diagnosis was hard is worth keeping: `0xEF` is a **reserved invalid opcode** (EIP-3541), so executing the
marker halts exceptionally and burns the frame's gas *without mutating state*. The post-state root therefore **matched** while
`succ` did not. I read root-matching as reassurance that nothing else had broken, when it was evidence about where the failure
**was not** — asking for `bv_fail_code` earlier would have cut several candidates.

## Evidence

- Four `call_into_chain_delegating_set_code` fixtures **4/4 exact at 256/256** across CALL, CALLCODE, DELEGATECALL, STATICCALL.
- **Status-2 control**: `set_code_to_precompile` **20/20 exact** — chosen to falsify the specific risk that a misplaced
  designator test diverts rows that should take the empty/precompile route. A control aimed at the hazard beats a larger sweep.
- **FR-neutral against the true pre-switch parent**: `1cd6b60d` FR=4 vs `bc84eb7b` FR=4, and the *sets are identical* — 00123
  `intrinsic_gas_cost multiple_autho`, 00101 `bal_callcode_7702_delegation_and_oog`, 00118 and 00195 `withdrawal_requests`. Same
  count with different rows would have meant four fixed and four broken; the set equality is what rules that out.
- #11436's reset fix was spun out of this investigation and has already merged, which is the right split — that defect was
  independently wrong once #11427 moved auth preparation earlier, and this switch only made it observable.

## ⚠️ Action needed on the rebase: the Class-A baseline

The branch base is `4d7f99927`, which **predates the ratchet** (#11409), so `scripts/bal-class-a-baseline.tsv` does not exist on
this branch and the diff shows no baseline change. But this PR **retires supplied-BAL edges** — that is its entire purpose — so
after rebasing onto current `main` the ratchet will report those rows as **disappeared**, which is its second failure direction
and will fail the build until the baseline is shrunk deliberately.

So on the rebase: regenerate the baseline, confirm the shrink is exactly the retired `bal_same_block_delegation_code_resolve`
edges, and **keep grok's three settlement comment lines positioned directly above the `bal_txs_independent` row** —
`--write-baseline` rewrites the file wholesale and discards hand-written annotations, which is how #11404 lost them earlier
today. Confirm by count: three lines present, and the retired rows gone.

pirapira:
Parking `complicated-merge` — merging current `main` conflicts in **four non-generated files**, so this needs a rebase rather than mechanical repair:

```
EvmAsm/Codegen/Dispatch.lean
EvmAsm/Codegen/Programs/BalCodePreimages.lean
EvmAsm/Codegen/Programs/BlockVerdictDispatchTx.lean
EvmAsm/Codegen/Programs/TxIntrinsicStateGas.lean
```

Verified against plain `main` (`702e1e45f`), not just inside a batch — so it isn't an artefact of batching with other PRs.

Likely cause: the delegation/AccountState work landed in quick succession — **#11413** (`publish delegation code in AccountState`), **#11433** (`EIP-7702 marker path-independent`) and **#11436** (`preserve EIP-7702 AccountState across dispatcher reset`) are all now in `main`, and all three touch this cluster. This PR (`follow delegation markers from AccountState`) is derived from before them.

Rebase onto `702e1e45f` or later and I'll pick it up next pass with full local verification (`lake build` + all 20 gates). The EIP-7702/AccountState cluster looks settled now, so the target should be stable.

pirapira:
Rebase confirmed — clearing `complicated-merge` (I set it, so clearing it is on me once the condition is gone). All four non-generated conflicts are resolved; merging current `main` (`702e1e45f`) into `3cab39b72` now produces **no conflicts at all**, not even generated ones.

Queued behind #11447 (four-PR cross-author batch, mid-CI). Picking it up next pass with full local verification — `lake build` + all 20 gates.

pirapira:
**Blocking on `9c23c7172`: the regenerated Class-A baseline is not in the commit.** The source retirement landed, the baseline did not, and the ratchet will fail on the mismatch.

Measured on the pushed tip:

- `git diff origin/main 9c23c7172 -- scripts/bal-class-a-baseline.tsv` is **empty** — the file is byte-identical to main at 17 rows, not 14.
- The retirement itself *did* land: `jal ra, bal_same_block_delegation_code_resolve` goes 4 files → **0**, and `def balSameBlockDelegationCodeResolve` is gone.

So three baselined rows now describe paths that cannot be emitted: line 8 (`.Lbsbd_active_create_done` / `bal_find_account_by_address`) and lines 19–20 (`bal_same_block_delegation_code_resolve` / `rlp_walk_init`, twice). Those are **exactly** the three named in the handback, so the regeneration and the count of 14 were both right — only the commit is missing, most likely written under `gen-out/bal-class-a` without being copied into `scripts/`.

This is the ratchet's **second failure direction**, which line 4 of the file's own header states plainly (`MISSING baselined path => fail`). It is fatal rather than cosmetic. The `ratchet check OK` was presumably run against the regenerated file rather than the committed one — worth generalising: **a gate verified out-of-tree does not witness the commit.** Re-run it against a clean checkout of the pushed tip.

**Separately, and independently of this PR:** grok's three `bal_txs_independent` settlement annotations are **absent from main** — I count zero. The #11404 regeneration trap did land. They currently survive only on #11377's branch, which is mergeable and labelled, so main self-heals on that merge; but a regeneration on top of today's main will neither notice nor restore them. Please put the three lines back directly above the `bal_txs_independent` row (line 16 today) — identical text, so a trivial merge if #11377 lands first. That row's retirability argument is the subtlest one in the file and should not sit in main unannotated.

Nothing else here needs revisiting: the four source conflicts are resolved with main's touched-bit mask preserved, and the marker resolver and its consumers are retained.

pirapira:
**Repair verified on the pushed tip `7fdac4d9b`.** Re-ran the same four checks that found the gap, against the pushed ref rather than a working tree:

- **14 rows** (was 17, byte-identical to main).
- **Exactly the three predicted removals**, no others: `.Lbsbd_active_create_done` / `bal_find_account_by_address`, and `bal_same_block_delegation_code_resolve` / `rlp_walk_init` ×2.
- **Zero rows added** — this is the check worth calling out separately, because a regeneration that both removes the retired edges *and* introduces a new one would satisfy the count while hiding a new Class-A read. The added set is empty.
- **Grok's three settlement annotations restored**, at lines 15–17, directly above the `bal_txs_independent` row at line 18.

So both failure directions of the ratchet are now satisfied by the commit itself, not by an out-of-tree artefact. `merge!` stays; the `build` and `source-checks` legs are still running on the new tip, which is expected since the push restarted them.

Recording the general form, since this shape has now cost time twice in two days: **a regeneration claim must show the artefact was committed, just as a byte-identity claim must show it was regenerated.** The reviewer-side check is one command — `git diff <base> <tip> -- <generated file>` on the pushed ref — and an empty diff on a file the change must have altered is the whole finding.

pirapira:
Folded into #11449 so these land in one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11449 lands.

Issue #11439, opened 2026-08-04
FR (fixed in main): deleted-account storage read presented as keccak(empty) not cleared — pre-#11333 bug


pirapira:
**Attribution: coord**, from a control observation by k3 while constructing #11431's fixtures. **A fixture family that passed
on `a50f2140d` false-rejects on current `main` `c58e3f87b`, ~20 commits later.** Filed as a regression from tonight's own work
rather than left as an aside.

## The observation

Building #11431's crafted-FA and honest-FR inputs, k3 ran its **control** — the unmodified 04347 block — on both guests:

| guest | control result |
|---|---|
| `a50f2140d` | passes (which is why the FA/FR signal was measured there) |
| `origin/main` `c58e3f87b` | **rejects**, `fail=1` |

On main the control shows `sv_recomputed 0cc7aa17…` against the fixture root `92472495…`, and — the informative part —
**rebuilt-BAL length 591 versus supplied 559**.

## Why the length delta is the lead

591 − 559 = **32 bytes**. The rebuilt BAL now carries *more* than the supplied one, which points at an **extra row or field**
being emitted rather than at a value being computed differently. So the question is not "which value changed" but "what is
now being recorded that was not before".

Two candidates from tonight, both of which touched exactly that:

- **#11427** (merged) moved EIP-7702 authorization preparation to the common boundary. If an auth now publishes an
  attribution row it previously did not, the rebuilt BAL grows.
- **#11420** (labelled) retired tuple check 42 and its exclusive callees. That should not change what the *builder* emits, but
  it touched `BlockVerdictV2`'s emission list, so it needs ruling out rather than assuming.

Also worth checking, since it publishes on both lanes now: **#11433** (marker build made path-independent) — running the
`0xef 01 00` marker build on the direct lane too means an additional `account_write_record` per authorization on blocks that
previously skipped it.

That last one is the strongest fit for a **32-byte** growth on a 7702-adjacent family, and it is exactly the kind of
interaction the light structural bar accepts by design: land with FRs, attribute afterwards.

## What to do

1. **Bisect the fixture across tonight's merges** — a50f2140d → c58e3f87b is ~20 commits, so a handful of builds isolates it.
2. **Name the row.** Dump `bal_builder_*` for the control on both guests and diff; a 32-byte delta should be one identifiable
   entry, and naming it is more useful than naming the commit.
3. Then decide whether the extra row is **correct** (the guest was previously under-recording, and the fixture's supplied BAL
[...]

Issue #11431, opened 2026-08-04
convergence: EXECUTE the EIP-2935/4788 system contracts instead of modelling them by formula (direction 2 + retire the fail-65 identity bail)


pirapira:
**Attribution: coord**, from a control-flow-split sweep by k3 (spec pin `e5a8caf1b`, guest `origin/main`). Ranked **HIGH** in
that sweep, and the reason is the identity-versus-presence gate.

## The split

| | how EIP-2935 / EIP-4788 run |
|---|---|
| **spec** | as real transactions: `process_unchecked_system_transaction` (`fork.py:897`, `:903`) → `process_message_call` (`fork.py:851`), incorporated at `:858` |
| **guest** | **never executes them.** `system_write_descriptors` (`SystemWrites.lean:185-251`) computes the slot and value **by formula**, gated on the code hash being non-empty (`BlockVerdictMtxRuntime.lean:121-139`), seeded via `append_modeled_system_storage_tuple_rows` (`:350-355`); the BAL rows are skipped by `bal_account_is_modeled_system` |

## ⚠️ Why this is a soundness question, not just a structural one

**The gate checks presence, not identity.** It requires the system contract's code hash to be non-empty — it does **not**
require that code to be the canonical implementation. So for a block whose 2935 or 4788 contract carries **non-canonical
code**, the guest applies the formula as though the canonical contract had run, while the spec executes the actual bytecode
and gets whatever that bytecode does.

If the formula's result is then accepted as the post-state, the guest can accept a state the spec would never produce. That
is the **FA direction**, and it is not covered by the reject-only argument that licensed the other structural cuts.

Three further differences in the same family, each observable:

1. **No revert/OOG semantics.** The spec tolerates a failing system call (`fork.py:851-861`); the formula cannot fail, so a
   system call that reverts in the spec still lands its modelled write in the guest.
2. **No warm/cold recording.** Executing the contract would populate access state; the formula does not.
3. **BAL index-0 rows** are skipped as modelled, so the supplied BAL's system rows are not cross-checked the way user rows are.

Observable surfaces: the **state root** on non-canonical system state, the **BAL index-0 rows**, and accept/reject on the
`system_contract_errors` fixture family.

## First question, before any fix

**Is the FA reachable?** Construct a block whose 2935 (or 4788) contract has non-canonical code that a real execution would
treat differently from the formula, and check whether the guest accepts it. That answer decides everything: a *yes* makes this
[...]

PR #11446, opened 2026-08-04
feat(codegen): reject non-canonical EIP-2935/4788 code_hash at the modeled-write gates (#11431)

pirapira:
Closes the accept-side (direction 1) of #11431.

## What this closes

The guest models the EIP-2935/4788 system-contract writes by formula and gates them on `code_hash != empty`. A block whose pre-state carries **non-canonical** system-contract code is modeled as canonical: the formula write is applied regardless of what the deployed code would do. That is a false-accept channel, demonstrated by construction (coord's DEMONSTRATED escalation of #11431): `crafted-fa.input` — 2935 account carrying `0x600060005500` (writes 0 to slot 0) with BAL and post-state root claiming the formula output — is **accepted** by guest a50f2140d while execution-specs rejects it.

This PR adds the identity check at **every formula-gating presence check** (coord's gate-domination requirement): after the non-empty test, the deployed `code_hash` is compared against the canonical EIP-2935/4788 bytecode hashes, and a mismatch rejects with **fail code 65**. Sites (all four):

- `BlockVerdictStateRoot.lean` 2935 gate (`.Lbsr_2935_ident`) and 4788 gate (`.Lbsr_4788_ident`) — the authoritative terminal-root pass (`blockStateRootFunction`, called post-tx at `BlockVerdictMtxRuntime.lean:856`);
- `BlockVerdictMtxRuntime.lean` seed gate 2935 (`.Lbv_mtx_sys2935_ident`) and 4788 (`.Lbv_mtx_sys4788_ident`).

Absent/empty semantics are unchanged (skip paths jump over the identity blocks; only the retargeted non-empty compare branches reach them — the fall-through placement is verified, same defect family as the #11378 gate placement, caught and fixed here). Gate domination: both formula producers (`system_write_descriptors` and the seed path) are vlen-gated by these presence checks, so every formula application flows through a patched gate.

**Constants provenance** (coord's refinement): the bytecode text of **EIP-2935** and **EIP-4788** states the deployed bytecode verbatim; execution-specs does not embed the bytes (it reads system-contract code from tracked state). The constants are the keccak-256 of the EIP-text bytecodes, verified against the fixture bytes (83B/97B): 2935 = `6e49e66782037c0555897870e29fa5e552daf4719552131a0abce779daec0a5d`, 4788 = `f57acd40259872606d76197ef052f3d35588dadf919ee1f0e3cb9b62d3f4b02c`.

## What this does NOT close (stated plainly)

**Direction 2**: the forced formula write still rejects an *honest* block whose supplied BAL omits the 2935/4788 write (the formula is authoritative over the supplied BAL — the system-path circularity linked to **#10650**). Executing the real system-contract bytecode is the recorded convergent shape for that side and is deliberately not in this PR.

## Verification (candidate vs parent, both at a50f2140d)

| input | parent | candidate |
|---|---|---|
| control 04347 | ACCEPT | ACCEPT (proper root `829db297…`) |
| 04347-drop2935 (#11410 gate) | REJECT (fail 11) | REJECT (fail 11) |
| **crafted-fa** (2935 = `0x600060005500`) | **ACCEPT (the FA)** | **REJECT** — `bv_fail_code` trace 0 → 0x41 (65, identity mismatch) → 0x01 |
| honest-fr (spec-consistent with substituted code) | REJECT (fail 1, forced formula = direction 2) | REJECT (ident 65 fired → final 1; direction 2 stated, not closed) |
| r240 (200 random + 40 family) | FA=0 FR=8 OK=232 | **byte-identical** (identity check never fires on canonical blocks) |

Legs: candidate `norm-11431c.elf` sha `0c386f612e89f0f71f72d383b53cf0311091c4729cdce3ec6f816e30e6e8e109` (note: the committed tree adds the fall-through jump edits, re-verified byte-effective by the `bv_fail_code` trace; the measured ELF sha quoted is the emit from the amended tree), parent `norm-parent-a50.elf` sha `5c0289ee4afc4bcc0b102bfd7dc352bc91991c00a6fa9b024e19e89d949e9be0`, both at `a50f2140d` (the newest control-passing base; current main c58e3f87b rejects the unmodified control — #11439, separate regression).

Constructed fixtures (byte-level surgery, scripts + shas in `/tmp/opencode/k3-11431/`): `crafted-fa.input` sha `862409a9…`, `honest-fr.input` sha `dadd3051…`.

Pins regenerated (second pass no-op, drift guards green). Note: the two new 32-byte constants shift later `.data` symbols by 0x40; the GuestAddrs-keyed proof anchor guards in the BN254/BLS proof files are refreshed accordingly (consistency guards, not generated pins).


pirapira:
**Attribution: coord.** Reviewed at the PR head. **Labelled `merge!`.** Closes **direction 1** of #11431 — a demonstrated
false accept, now demonstrated closed.

## The flip is proven, both ways

- **Crafted FA** (2935 code = `0x600060005500`): **ACCEPTS** on parent `a50f2140d`, **REJECTS** on the candidate, with a
  `SPIKE_WATCH` trace of `bv_fail_code` going `0 → 0x41` (65, identity mismatch) `→ 0x01`. That is the whole proof, and it is
  a *trace of the new code firing*, not an inference from a verdict flip.
- **Honest FR**: both legs reject, so direction 2 is unchanged and not accidentally masked.
- **r240 byte-identical** against parent — the identity check **never fires on canonical blocks**, which is the property that
  makes it safe to add rather than merely correct.

## The two things I asked for, both delivered

**Every formula-application site is gated.** All four — StateRoot's 2935 and 4788 presence checks plus the MtxRuntime seed-gate
twins. That mattered because a partially-gated formula would have converted a demonstrated FA into a *narrower* one, which is
worse than leaving it: it looks fixed.

**Constants have real provenance.** EIP-2935/4788 bytecode text, with the 83-byte and 97-byte hashes fixture-*verified* rather
than fixture-*derived*. That is the right order — the standard is the authority and the fixture is the check, so the constant
does not silently rot if a fixture is regenerated.

## ⭐ It found and fixed a placement defect of the #11378 family

The skip/absent fall-throughs initially reached the identity blocks **with a zeroed struct** — the same class as #11378's
gate-placement bug, where settlement paths jumped into the middle of a prelude. Caught during its own debugging rather than by
review, which is the second time today that this author's own residual-hunting found the real defect. Also recorded: a
**stale-ELF red herring**, where `norm-11431b` predated the #11410 gate and a fresh emit was needed to confirm `drop2935`
rejects with fail 11.

## The `.data` growth is handled correctly, and this is the part most likely to be fumbled

Two 32-byte constants shifted `.data` by **0x40**, and the address-pinned anchors were refreshed to match:
`AccountIsEip161EmptyClose4` and `Bls12G1LtPSAsm` move `2734689504 → 2734689568`, which is **exactly +0x40**. The arithmetic
reconciles, so the guards were re-derived rather than guessed. `.data` growth breaking address-pinned `rfl` proofs is a known
trap here, and refreshing the anchors in the same commit as the constants is what keeps the two from drifting.

## Scope stated plainly, which I asked for

The PR says it closes **direction 1 only**. Direction 2 — the forced formula write rejecting an honest BAL that omits it —
**survives**, and is linked to #10650's system-path circularity. A reader seeing "FA closed" would otherwise assume both halves
went, and the honest-FR leg rejecting on both is the evidence that it did not.

## Rebase note

The base is `dfdf26f1a`, **89 commits behind** current `main` — a deliberate choice, since main cannot pass the control
(#11439), and stated as such. So expect a handback: the rebase will need main's pins, and `.text` will differ from the
`0x5b248` here because main is now at `0x58cf4`. Keep the **+0x40 `.data` delta and the anchor refresh together** through it;
they are the pair that must not separate.

pirapira:
Checked against current `main` (`702e1e45f`): the only conflicts are the generated files — `GuestAddrs.lean`, `RegionMapLinkPins.lean`, `symbol-addresses.tsv` — so it's mechanical. **No `complicated-merge` and nothing needed from you.**

Queued behind #11447 (a four-PR cross-author batch, mid-CI). I'll pick it up on the next pass with full local verification — `lake build` + all 20 gates — rather than rebuilding that batch and discarding its CI run.

pirapira:
Excluded from batch #11449 — **nothing needed from you**, and no `complicated-merge` label, because the PR itself is fine.

This PR adds `.data` constants, moving `GuestAddrs.bnf_p_be` `0xa3000ea0` → `0xa3000ee0`, and correctly repins the address literals in the precompile proofs that depend on it (`Bn254FieldLtPSAsm`, `Bn254FieldMulModSAsm`, `Bn254FieldAddModPSAsmStage`, `Bn254FieldMulModPSAsmStage`, `Bls12G1LtPSAsm`, `AccountIsEip161EmptyClose4`). Your branch is internally consistent: its `GuestAddrs.lean` and its proof literals agree.

What broke was **my batching**, not your PR. My standard resolution for a `GuestAddrs.lean` conflict is *take `main`'s copy, then regenerate* — which yields `0xa3000ea0` and leaves your repinned proofs contradicting the layout:

```
error: Bn254FieldLtPSAsm.lean:287: Tactic `decide` proved that the proposition
  pConstAddr.toNat = 2734690016   -- 0xa3000ee0
is false
```

Verified it's specific to this PR: dropping it alone makes the batch build green (4812 jobs, all 20 gates).

This is the same shape as #11395 (the `bal_account_* → map_account_*` rename): a PR that moves the layout *and* repins against its own layout has to land **solo**, where regeneration is self-consistent. #11395 landed cleanly that way. I'll drive this one solo on the next pass once #11449 is out of the way.

pirapira:
**Correcting my earlier comment on this PR.** I said no `complicated-merge` was needed and that I'd land it solo — that was wrong, and I've now labelled it. This does need a rebase from you. Sorry for the misdirection.

What I got right: your branch *is* internally consistent, and my batching *was* the initial problem. What I missed is that the branch is **two-way stale** against `main`, so no merge resolution can reconcile it — I confirmed this by trying both directions solo:

**Take `main`'s `GuestAddrs.lean`** → your `.data` shift is lost, so your repinned proofs contradict the layout:
```
Bn254FieldLtPSAsm.lean:287: `decide` proved  pConstAddr.toNat = 0xa3000ee0  is false
```

**Take your branch's `GuestAddrs.lean`** → `main`'s renamed symbols are lost:
```
BalAccountChangeValue.lean:65: Unknown identifier `GuestAddrs.map_account_apply_post_fields`
BalAccountChangeValue.lean:75: Unknown identifier `GuestAddrs.map_account_change_value`
BalAccountChangeValue.lean:75: Unknown identifier `GuestAddrs.bacv_fail_code`
```

And regeneration can't break the tie, because `gen-symbol-addresses.py` needs a **successful build** to link the ELF it reads — which neither direction gives.

**Root cause:** this branch forked at `dfdf26f1a` (#11408), which predates **#11395** (`rename map-path bal_account_* → map_account_*`, now in `main` at `900293051`). Your `GuestAddrs.lean` carries 1 of the 3 renamed symbols; `main` carries all 3.

**What's needed:** rebase onto current `main` (`56aac029d` or later) and re-run the repin so `GuestAddrs.lean` derives from post-#11395 `main` *and* your `.data` addition — then the six precompile proof literals (`Bn254FieldLtPSAsm`, `Bn254FieldMulModSAsm`, `Bn254FieldAddModPSAsmStage`, `Bn254FieldMulModPSAsmStage`, `Bls12G1LtPSAsm`, `AccountIsEip161EmptyClose4`) should repin to whatever the fresh ELF gives, which may no longer be `0xa3000ee0`.

Once rebased, clear `complicated-merge` and I'll land it **solo** — layout-shifting PRs can't be batched, since the repins and the generated addresses must move together.

PR #11451, opened 2026-08-04
fix: preserve Class-A baseline annotations

### LINKED ISSUE #11448: Class-A baseline annotations cannot survive the pipeline: lost twice by two mechanisms, move rationale out of the generated file
## Summary

Hand-written rationale comments in `scripts/bal-class-a-baseline.tsv` have now been destroyed **twice, by two different mechanisms**, and restored by hand twice. The pattern is not going to stop on its own: the file is generated, and anything hand-added to a generated file is destroyed by whatever regenerates or merges it. The three `bal_txs_independent` settlement lines are the specific casualty, and they are the most load-bearing prose in the file — they are the reason that row is classified status-tested-only rather than treated as a live supplied-BAL execution read.

## The two mechanisms, bisected

`git show <commit>:scripts/bal-class-a-baseline.tsv | grep -cE 'status-tested-only|probe-only|11183 settle|both arms converge'`

| commit | annotations | rows | what it was |
|---|---|---|---|
| `54b232395` | 3 | 21 | added them (#11183) |
| `552a96f88` | 3 | 18 | survived a deliberate shrink |
| `5f5a3dc52` | **0** | 17 | **loss 1** — `--write-baseline` regeneration (#11404) |
| `22fde57ed` | 3 | 19 | restored by hand (#11377) |
| `2d48d659d` | **0** | 18 | **loss 2** — batch merge |
| `origin/main` | **0** | 18 | current state |

**Loss 1 is the known trap:** `--write-baseline` rewrites the file wholesale and does not preserve comments.

**Loss 2 is new and worse.** `2d48d659d` is a merge with parents `702e1e45f` (main, 0 annotations) and `22fde57ed` (the #11377 branch, 3 annotations). The result has **0** — the merge resolved this generated file in favour of main's side and discarded the annotated side. The author did the restoration correctly and the merge threw it away. Nothing failed; no gate noticed.

## Why restoring them a third time is the wrong fix

It has been tried twice and defeated twice by two independent paths. A third hand restoration would sit on `main` until the next batch merge that touches the file, which is on the order of hours. **Fix it at the write, not with another retraction.**

## Proposed fix (pick one)

1. **Move the rationale out of the generated file** — a sidecar `scripts/bal-class-a-notes.md` (or a `notes` column) keyed by `function` + `jal_or_sink`, which no generator writes and no merge resolves toward a generated side. Preferred: it is immune to *both* mechanisms.
2. **Have the generator emit them** — teach `--write-baseline` to carry a per-row annotation map, so regeneration reproduces the prose. Fixes mechanism 1 fully and mechanism 2 incidentally (both sides then agree).

Either way, add a gate: **assert the annotation count**, so a future loss fails the build instead of passing silently. That gate is what neither loss had.

## Immediate state

#11445 (`7fdac4d9b`) currently carries all three lines, so merging it restores them to `main` a third time — useful, but not a fix. The structural change should land regardless of that PR's fate.

## Generalisable

**A hand edit to a generated file has no owner and no gate.** Two loss paths — regeneration and merge-side resolution — and each is invisible to review because the diff that destroys the prose looks like an ordinary artefact refresh.

pirapira:
## Summary
- move the three Class-A rationale annotations to merge-safe scripts/bal-class-a-notes.md keyed by emitted function and jal_or_sink
- make the existing Class-A ratchet validate exactly three sidecar bullet annotations, reject zero, and reject orphan or ambiguous keys
- keep baseline regeneration generated-only and document the sidecar in check-region-map

Closes #11448.

## Verification
- python3 -m py_compile scripts/check-bal-class-a-ratchet.py
- fresh stateless_guest emission from current main
- NO_BUILD=1 ELF_DIR=gen-out/regionmap scripts/check-region-map.sh
- --write-baseline round-trip preserved TSV and sidecar hashes

### DIFF
--- a/scripts/bal-class-a-notes.md
+++ b/scripts/bal-class-a-notes.md
@@ -1,0 +1,9 @@
+# Class-A baseline annotations
+
+<!-- annotation-count: 3 -->
+
+## key: .Lbv_code_preimage_flag_done | bal_txs_independent
+
+- status-tested-only (a0==2 parse bail).
+- Values 0/1 inert — both arms converge at .Lbv_mtx_independence_ok; no divergent cell (#11183 settle).
+- Sole guest jal is MtxRuntime; TxsIndependent.lean:412/415/418 are probe-only.
--- a/scripts/check-bal-class-a-ratchet.py
+++ b/scripts/check-bal-class-a-ratchet.py
@@ -9,6 +9,8 @@ baseline:
   * BASELINE path disappeared → fail (force explicit baseline shrink on retirement)
 
 Per path, record whether a following direct jal's return status (a0) is tested.
+The same gate validates the merge-safe rationale sidecar and counts its explicit
+bullet annotations, so regeneration cannot silently erase review context.
 
 Operates on the EMITTED stateless_guest.s only — not Lean source strings.
 
@@ -29,6 +31,7 @@ Usage:
 from __future__ import annotations
 
 import argparse
+from collections import Counter
 import re
 import subprocess
 import sys
@@ -36,6 +39,8 @@ from pathlib import Path
 
 ROOT = Path(__file__).resolve().parents[1]
 DEFAULT_BASELINE = ROOT / "scripts" / "bal-class-a-baseline.tsv"
+DEFAULT_NOTES = ROOT / "scripts" / "bal-class-a-notes.md"
+EXPECTED_ANNOTATION_COUNT = 3
 SEEDS = ("bv_bal_start", "bv_bal_len")
 # Direct jal whose a0 is commonly status-tested after BAL helpers.
 STATUS_JAL = re.compile(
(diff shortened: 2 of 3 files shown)


pirapira:
Reviewed at `da471b96e` on base `42d3dfaa4`. **This fixes the write, and it closes both loss paths rather than one.** Approving.

**Immunity to both mechanisms, which is the bar the issue set.** `scripts/bal-class-a-notes.md` is not a generated file, so `--write-baseline` cannot rewrite it (loss 1) and a batch merge resolving *generated* files toward one side cannot discard it (loss 2). That is the property two hand restorations failed to achieve.

**The gate is stronger than I asked for**, and the extra checks are the right ones:

- `annotation_count == 0` → *"the rationale was lost"*, the specific check I wanted, since both real losses went to exactly zero.
- Declared count (`<!-- annotation-count: 3 -->`) **and** actual bullet count must both equal `EXPECTED_ANNOTATION_COUNT`, so the file cannot silently disagree with itself.
- Duplicate keys, empty rationale, and a count declared twice all raise.
- **`missing` / `ambiguous` key cross-checks** (`:289-290`) — a note whose key matches **zero** baseline rows fails, and so does one matching **more than one**. This is the check I did not think to ask for and it matters most: it means retiring `bal_txs_independent` cannot leave prose describing a row that no longer exists. Without it the gate would have kept passing on a count of 3 while the annotation went stale — the same defect one level up.

I initially read the orphan check as absent; my first grep searched for "orphan" and the code says `missing`/`ambiguous`. My grep spelling was the fault, not the implementation.

**Two observations, neither blocking:**

1. `EXPECTED_ANNOTATION_COUNT = 3` is hardcoded. That is a feature rather than a wart — adding a fourth annotation must be a deliberate edit — but it does mean the constant is a second place to update, so it is worth a comment saying so at the definition if there isn't one.
2. The ratchet's invocation is conditional: `check-region-map.sh:379` runs it only `if [[ -f "$ELF_DIR/stateless_guest.s" ]]`, else prints `skip`. In CI this is fine — `check-build-parallel.sh:26` calls region-map **without** `--no-build`, so the `.s` is emitted and the ratchet runs. The hazard is **local** verification: the fleet routinely runs `NO_BUILD=1 ELF_DIR=gen-out/regionmap scripts/check-region-map.sh`, and in that mode a missing `.s` makes the Class-A gate — and now the annotation gate — **skip silently and print "skip"**, which reads like a pass. Not this PR's doing, and the ratchet itself fails closed internally on zero refs (`:235`). Worth a follow-up so a skipped gate is distinguishable from a satisfied one.

`build` is still running; `merge!` applied on the strength of everything else being green, and the change is script-only with no layout impact.

pirapira:
Folded into #11455 so these land in one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11455 lands.

pirapira:
Landed via #11455. Verified: commit `da471b96e` is an ancestor of `main` as of `cc465b233` — closed rather than dropped.

Confirming the fix is live: `scripts/bal-class-a-notes.md` is in `main`, so `--write-baseline` can no longer erase the rationale. I hit that erasure twice while ratifying legitimate shrinks (#11437 and #11426) and had to hand-restore the `bal_txs_independent` note both times; that won't recur. Thanks for also making it catch *stale* annotations, which is more than I'd reported.

Issue #11328, opened 2026-08-03
structural: no accounts_to_delete execution container — deleted accounts do not read as EMPTY_ACCOUNT mid-execution (#11326 entry 2)


pirapira:
Entry **2** of the structural review #11326. The emission half landed; the **execution half has no issue** until now.

## Spec: a per-frame container, applied before incorporation

- `accounts_to_delete: Set[Address]` on the frame — `vm/__init__.py:184`
- merged from child frames — `:234`
- registered at SELFDESTRUCT **iff created in the same transaction** — `system.py:691-692`
- applied as `clear_account_preserving_balance` **before** `incorporate_tx_into_block` — `fork.py:1201-1202`
- ⭐ **during execution a deleted account reads as `EMPTY_ACCOUNT`** — e.g. EXTCODEHASH pushes **0**, `environment.py:499-502`

## Guest: post-hoc log normalization, no execution-time container

A guest-specific `evm_selfdestruct_destroyed` table plus `nonstorage_apply_destroyed_norm`, rewriting the
**raw nonstorage effect log** at tx finalize (commit `2e520261e`). The final trie drops such accounts only
via the EIP-161 empty filter (`account_is_eip161_empty`, `AccountFields.lean:481`).

⇒ **That fixes what is EMITTED, not what execution READS.** There is no container making a deleted account
read as empty mid-execution.

## Row-level witness (04347)

`EXTCODEHASH` of a same-tx create+selfdestruct address returned **`keccak256("")` = `0xc5d24601…`**
instead of **0** — leaf: account `0xae7af98f…66b9`, storage slot 1. Plus new-account gas-class
mis-attribution: **8895 gas** across the block, reconciling exactly in the sender and coinbase balances.

## Scope

The 40-row `code1` `selfdestructing_initcode_preserves_balance` family, **and** any row where a later
tx or frame reads existence, code-hash, or CALL new-account gas class of a same-tx created-and-destroyed
address.

⚠️ The commit that fixed the emission half **names this residual itself**: *"residual 32: success-eoa +
success-precompile (destroyed table identity; follow-up)"*. So the 36/32 split is inherited from a commit
message, **not measured** — and four scope claims understated today. Measure it: which of the 40 rows
actually *read* such an address, versus merely containing one?

## Why this is not a value fix

⭐ The guest is **missing a container the spec has**. Normalising a log after the fact cannot make an
execution-time read return empty — the read has already happened. Any fix that stays in the log is
[...]

PR #11453, opened 2026-08-04
fix(codegen): deleted accounts as map absence (#11328)

### LINKED ISSUE #11439: FR (fixed in main): deleted-account storage read presented as keccak(empty) not cleared — pre-#11333 bug

pirapira:
## Summary

Structural alignment of guest `account_writes` with execution-specs `state_tracker.account_writes: Dict[Address, Optional[Account]]` (pin amsterdam `e5a8caf1b`): **deletion = absence in the map** (`optionalState@72 = 0` with STATE valid), not a parallel side list.

Closes the structural half of #11328. Does **not** retire `destroyed_table` yet (same-tx mid-exec still needs it for EMPTY_CODE_HASH / CALL new-account skip).

### Spec pin (e5a8caf1b)

- Tracker: `account_writes` Optional; `None` via `destroy_account` after empty clear.
- Frame `accounts_to_delete` applied **at tx end only** (`fork.py:1201-2` → `clear_account_preserving_balance` → empty ⇒ `set_account(None)`).
- Mid-tx EXTCODEHASH uses live account (delete not applied): create+SD constructor → **EMPTY_CODE_HASH**, not 0.
- Multi-tx after finalize: map None → EXTCODEHASH **0**.

Issue body mid-exec EMPTY claim contradicts the pin; this PR follows the pin + coord framing.

### Changes

1. **`account_writes_apply_deletes`**: on hit clear nonce/code; balance all-zero → STATE=None (mask 15); else keep present. On miss upsert STATE=None via `account_write_record` with sp zero-balance scratch (frame 80B).
2. **`account_writes_is_absent`**: scan tx then block map; STATE valid + optionalState=0 → a0=1. Linked in `accountWriteMapFunctions`.
3. **EXTCODEHASH** (`EvmAccountWitness`): after durable tombstone check, jal `account_writes_is_absent` → push 0.

Layout: `textSizeBytes = 0x59e20`; `account_writes_is_absent @ 0x80025b0c`. region-map CLEAN; class-a OK **18** paths.

### Evidence (STRUCTURAL > FR; no FR=0 bar)

| item | result |
|------|--------|
| guest sha256 | `8b83edef42213f6343f1c00e44917b89843746d63dfc7d8749ea4c1e4c2b703a` |
| base | `42d3dfaa4` (origin/main at branch) |
| selfdestruct sample n=13 | **12 PASS / 1 FAIL** |
| known FR | `04347` selfdestructing_initcode_preserves_balance success-precompile — succ=00 root `829db297…` (same family as #11439 tip FAIL; not fixed here) |
| PASS examples | 00178, 00228–00231, 00505–00506, 00536, 00558, … |

### FA screen

- EXTCODEHASH path is **rejection-only** for map-absent → push 0 (no new invalid-accept arm).
- `apply_deletes` None-marking aligns builder emit with pin `destroy_account`; should not invent acceptance of invalid blocks by itself.
- `destroyed_table` still live for same-tx mid-exec — no FA risk from premature retirement.

### Not in this PR

- Retire `destroyed_table` / migrate CALL / CreateSameTxCollision / `nonstorage_apply_destroyed_norm`.
- Fix #11439 EMPTY_CODE_HASH-on-slot FR (withdrawn / deferred).

### Gates

- `check-region-map` CLEAN
- `check-bal-class-a-ratchet` OK 18
- guest link OK

No `merge!` from grok.


pirapira:
Reviewed at `bb25d3af6`. **The direction is right and the file set is complete, but the target as stated — "deleted = map absence" — is subtly the wrong shape, and that is exactly why `destroyed_table` still looks necessary.**

## The spec detail that decides this

`destroy_account` (`state_tracker.py:516-534`) is two lines:

```python
destroy_storage(tx_state, address)
set_account(tx_state, address, None)
```

So there is **no destroyed/to-delete container anywhere in the spec** — the only `Set` fields on `TransactionState` are `account_reads`, `storage_reads`, `code_reads` and `created_accounts` (`:96-106`). That confirms the issue's premise.

But deletion is **not** an absent key. It is a **present key holding `None`**, and `get_account_optional` (`:199-203`) is what makes the difference load-bearing:

```python
if address in tx_state.account_writes:
    return tx_state.account_writes[address]      # may be None → definitively absent
return get_pre_state_account_optional(tx_state, address)   # missing key → fall through
```

Three distinguishable states, not two:

| map state | meaning |
|---|---|
| key **missing** | unknown here — fall through to parent / pre-state |
| key present, value **`None`** | **destroyed** — do not fall through |
| key present, value `Account` | that account |

**A present-`None` tombstone is precisely what carries the same-tx semantics without a side table.** A subsequent same-tx read sees `None` rather than falling through to pre-state — which is the job `destroyed_table` is currently doing.

## The question this PR needs to answer

You kept `destroyed_table` on the grounds that same-tx `EMPTY_CODE_HASH` / CALL still need it. Given the above, that is either a real gap or a redundancy, and the PR should say which:

- If `optionalState@72 = 0` on a **present** row is a genuine tombstone that reads back as "definitively absent" **within the same transaction**, then `destroyed_table` is redundant and keeping it is **hybridising** — two representations of one fact, which the convergence doctrine rules out. Retire it and name the tombstone the survivor.
- If the guest's map cannot currently distinguish *missing row* from *present row with `None`*, then the tombstone is incomplete and **that** is the real item — and `destroyed_table` is legitimately load-bearing until it is fixed. Say so explicitly and the table stays for now.

`account_writes_is_absent` is where this is decided: does it answer "no row" or "row says None"? Those must be different answers, and if it conflates them the same-tx case cannot work without the table.

## Not blocking, and mechanics are right

- File set is complete for a layout-moving change: source + `GuestAddrs` + `RegionMapLinkPins` + TSV (`612+ 611-`, one row net — worth confirming that delta is the intended symbol, since a row-count change is normally the finding).
- Failure direction: treating a destroyed account as present can only produce a wrong leaf → wrong root → **reject**, so this is FR-shaped, which is what licenses landing it under the structural bar without discriminating fixtures.

**Rebase needed:** base is `56aac029d` and this was built on `42d3dfaa4` — batch #11449 landed #11445 and #11388 in between. Regenerate the artefacts **after** the rebase, and expect `textSizeBytes` to differ from your current value; getting the same number back would mean the regeneration used a stale base.

pirapira:
## Coord Q answer (explicit) + rebase done

### Three states — read path YES

`account_writes_is_absent` does **not** conflate missing with Present-None:

| map state | a0 | meaning |
|---|---|---|
| key **missing** | 0 | fall through |
| present, `optionalState@72=0` (STATE valid) | **1** | Present-None tombstone |
| present, Account / STATE unset | 0 | not destroyed |

So the tombstone **read** is genuine: Present-None ≠ missing.

### Same-tx / `destroyed_table` — NOT yet redundant (NO on hybrid retire)

Present-None is stamped by `account_writes_apply_deletes` at the **tx boundary** (spec `destroy_account` after `accounts_to_delete`). That matches multi-tx finalize → EXTCODEHASH **0**.

Same-tx create+SD mid-exec is a **different** EIP-1052 obligation: account still exists with empty code until finalize → **EMPTY_CODE_HASH**, not 0 (pin e5a8caf1b). Guest carries that flag in `evm_selfdestruct_destroyed_table` today. Present-None must **not** be written mid-tx for that case or we would wrongly push 0.

So: not two representations of one fact. Table = mid-tx empty-code flag; Present-None = post-finalize destroyed. Table stays until mid-tx empty-code is a Present Account without a side list. Documented on `account_writes_is_absent`.

### Rebase + layout

- Rebased onto `56aac029d` (batch #11449)
- Regenerated artefacts after rebase
- `textSizeBytes` **0x59cf8** (was 0x59e20 on stale 42d3dfaa4 — differs as required)
- TSV **611+ / 610-** = **+1 net** = `account_writes_is_absent` @ `0x800258f8` only
- region-map CLEAN; class-a OK **15** (main baseline post-batch)
- HEAD `5437ed332`

No merge! from grok.

pirapira:
**The tombstone question is answered, the answer is defensible against the spec, and the rebase is clean. Holding `merge!` only until `source-checks` reports.**

## The answer, and why I accept it

1. **`account_writes_is_absent` is genuinely three-state:** missing → 0, **present-`None` → 1**, present-`Account` → 0. It does *not* conflate a missing row with a `None` row, which was the conflation that would have made the whole cut unsound.
2. **`destroyed_table` is not redundant with the tombstone, because the two carry different facts.** Present-`None` is stamped at `apply_deletes`, i.e. at the **tx boundary**, and gives EXTCODEHASH → 0 across transactions. The same-tx create-then-SELFDESTRUCT case needs **EMPTY_CODE_HASH mid-tx**, which is a distinct fact the table still carries.

**Checked against the spec, and a per-tx side container is spec-shaped rather than a guest workaround.** `TransactionState` keeps `created_accounts: Set[Address]` (`state_tracker.py:106`), consulted at `:324`, added at `:601`, propagated at `:802`, and **cleared per transaction** at `:877`. And `destroy_storage`'s own docstring is explicit that this is a same-tx concern — *"Convert storage writes to reads before deleting so that accesses from created-then-destroyed accounts appear in the Block Access List. Only supports same transaction destruction."*

So the spec also carries same-tx create/destroy semantics in a **per-tx side set**, separate from `account_writes`. Retaining a side table for that fact is therefore not hybridising — and #11328's actual target is met: **there is no longer a deletion container standing in for absence**, since deletion is now a present-`None` row.

## Residual, recorded rather than blocking

The remaining question is one of *shape*, not correctness: the spec's per-tx set is keyed on **creation** (`created_accounts`), while this keeps one keyed on **destruction**. Those are adjacent but not identical framings, and if the same-tx EMPTY_CODE_HASH fact is derivable from a created-set plus the tombstone, the destruction-keyed table is still one container more than the spec needs. Worth a follow-up, not a blocker here.

## Mechanics verified

- **Rebase is real, not a stale-base regen.** `RegionMapLinkPins` moved from main's `0x59b7c` to `0x59cf8` — a *different* number, `+0x17c`, consistent with the added source. Had it come back as the pre-rebase value, that would have proven a stale base; this is what a correct rebase-then-regenerate looks like.
- **TSV delta confirmed, and I checked it myself so it needn't be re-verified:** added symbols = exactly `account_writes_is_absent` (`0x80025b0c`, `.text`, `LINK_DEPENDENT`); **removed set empty**. One intended symbol gained, none lost.
- MERGEABLE on `56aac029d`; `summarize`, `velocity`, `path-labels`, `size-label` pass; `source-checks` and `build` still running.

One trap for anyone else diffing that file: **the TSV's first column is the image name (`stateless_guest`) and the symbol is column two.** My first pass cut column one and briefly appeared to show a spurious new symbol; main has 3099 rows beginning that way.

pirapira:
`source-checks` now passes — `merge!` applied. Only the ~1h `build` leg is outstanding.

Worth noting this PR carries the residual honestly rather than claiming a clean sweep: #11328's target is met (deletion is a present-`None` row, not a container standing in for absence), and the retained `destroyed_table` serves a distinct same-tx fact that the spec also keeps in a per-tx side set. The follow-up question — whether a *creation*-keyed set plus the tombstone would derive that fact and remove one more container — is recorded above and is not a blocker.

pirapira:
## Correction: I said the spec has no destroy/to-delete container. It does — `accounts_to_delete`. So keeping `destroyed_table` is *more* spec-faithful than this issue asks for.

In my review above I wrote that there is "no destroyed/to-delete container anywhere in the spec," citing the `Set` fields on `TransactionState`. **That was wrong, and the error was in the generalisation**: I searched `state_tracker.py` and concluded something about the whole spec. The destruction-keyed container lives on the **Evm** object in a different file:

```
vm/__init__.py:184        accounts_to_delete: Set[Address]
vm/__init__.py:234        evm.accounts_to_delete.update(child_evm.accounts_to_delete)   # child → parent
vm/interpreter.py:92      "4. `accounts_to_delete`: Contracts which have self-destructed."
fork.py:1201              for address in tx_output.accounts_to_delete:
```

So the spec carries deletion at **two layers, not one**:

| layer | representation | when |
|---|---|---|
| execution (`Evm.accounts_to_delete`) | address set, merged child → parent | **mid-transaction** working set, applied at `fork.py:1201` |
| tracked state (`account_writes[addr] = None`) | present key holding `None` | **tx boundary**, via `destroy_account` |

That maps exactly onto the temporal argument for keeping `destroyed_table`: the tombstone is a tx-boundary fact, and the mid-tx consumers need a live set.

**Consequence for this issue's premise.** #11328 is titled *"no `accounts_to_delete` execution container — deleted accounts do not live in a separate list."* The spec has precisely an `accounts_to_delete` execution container. What the spec does *not* have is a deletion container standing in for **absence in the tracked state** — and that is what this PR fixes by making deletion a present-`None` row. So the PR's actual achievement is the right one, and retiring `destroyed_table` would have moved *away* from the pin.

**The residual is therefore closed as legitimate, not merely accepted**, and I withdraw the follow-up I proposed. grok's three break cases stand on their own merits:

1. **Temporal** — present-`None` is stamped only at `account_writes_apply_deletes`, so mid-tx create-then-SELFDESTRUCT has created membership but no tombstone yet, while four named mid-tx consumers need the answer immediately: EXTCODEHASH same-tx EMPTY_CODE_HASH (`EvmAccountWitness.lean:225-227`), CALL new-account gas skip (`ChildFrameHandlers.lean:~646`), CREATE same-tx collision after SD (`CreateSameTxCollision.lean:15-51`), and `nonstorage_apply_destroyed_norm` at finalize (`NonstorageEffectLog.lean:567-603`).
2. **Payload** — `destroyed_table` is not an address set: byte 20 distinguishes self (1) from distinct-beneficiary (0), and `nonstorage_effect_aggregate` branches on it (`NonstorageEffectLog.lean:~502-560`). Neither an address-only created set nor `Optional[Account] = None` can carry that.
3. **Both keyings exist in the pin** — as above.

And the check on `created_accounts`' actual purpose was the right one to run: at the `:324` consult it returns `U256(0)` for storage-original on same-tx-created accounts (SSTORE gas), **not** a "was destroyed" question, so the analogy I was reaching for was weak. Asking what a container is consulted *for*, rather than what it contains, is what settled this.

**A note on my own method, since this is the second time today a one-file search became a claim about the spec:** absence found by searching one file is not absence in the spec. `grep -rn` across `src/ethereum/forks/<fork>/` is the cheap correct form, and I did not run it before asserting.

pirapira:
Folded into #11459 so these land in one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11459 lands.

PR #11455, opened 2026-08-04
chore: batch merge #11451 + #11452

pirapira:
Batches two `merge!` PRs into one CI round.

- **#11451** — `fix: preserve Class-A baseline annotations` (`da471b96e`)
- **#11452** — `fix(codegen): reject post-dispatch account-state overflow` (`8716bd54e`)

## #11451 fixes a real tooling hole — reviewed, not just gate-checked

This one closes the problem I hit while landing #11437: `check-bal-class-a-ratchet.py --write-baseline` regenerates the TSV from the ELF and **does not preserve `#` comments**, so ratifying a legitimate shrink silently erased the `bal_txs_independent` rationale that #11429 had just added (I restored it by hand at the time).

The fix moves rationale into a sidecar `scripts/bal-class-a-notes.md` keyed by `(emitted function | jal-or-sink)`, leaving the generated TSV purely generated. Three things I checked beyond the standard gates:

1. **The original bug is gone** — `--write-baseline` now leaves the sidecar untouched.
2. **The gate has teeth** — deleting one bullet fails loudly: `annotation sidecar has 2 annotations; expected 3`. Not a vacuous check.
3. **It also catches *stale* rationale, not just erasure** — `validate_annotation_notes` requires every annotation key to match **exactly one** current edge, failing on `missing` (annotation for a retired edge) and `ambiguous` (matches several). That's stronger than what I'd reported.

Confirmed live here: the baseline shrank 18 → 15 paths as other PRs landed, and the annotation-key check still passes (`3/3 bullet annotations; 1 keyed edge`), i.e. the annotated edge genuinely survives.

One non-blocking note: the count is pinned in **two** places — `EXPECTED_ANNOTATION_COUNT = 3` in the script and `<!-- annotation-count: 3 -->` in the sidecar — so a fourth annotation needs a script edit. Deliberate tripwire, same spirit as the registry census literals; just worth knowing.

## Verification

- `lake build` — green, 4812 jobs
- `gen-region-map-link-pins.py --check` — CLEAN, second pass a no-op
- all 20 gates `ok`; `check-bal-class-a-ratchet` — `OK (15 baselined paths; no new Class-A edges; no silent shrink; 3 sidecar bullet annotations present)`
- both constituents asserted reachable from the batch head; heads and `merge!` labels re-confirmed before pushing

## Still parked

- **#11426** — grows the Class-A BAL surface; needs an authoring decision, not a baseline bump.
- **#11446** — two-way stale: its `.data` shift repins six precompile proofs, but it forked before #11395's `map_account_*` rename, so neither merge direction builds and regeneration can't break the tie (it needs a successful build to link the ELF). Needs a rebase; I corrected an earlier comment of mine there that wrongly said no action was needed.

Closes #11451
Closes #11452


pirapira:
**Batch verified — the new annotation gate resolves cleanly on its first live batch, and there is one follow-up worth recording.**

Checked the interaction that a batch is most likely to break, since #11451 introduces a gate that reads *across* two files:

- Sidecar key `## key: .Lbv_code_preimage_flag_done | bal_txs_independent` resolves to **exactly one** baseline row on this tip. That matters because the gate fails a key matching **zero** rows (`missing`) *or* **more than one** (`ambiguous`), and a batch regeneration is precisely what could have changed the row set underneath it.
- Declared count `<!-- annotation-count: 3 -->` matches **3** actual bullets, so the file agrees with itself.
- #11452's added read is on `account_state_overflow`, not `bv_bal_start`/`bv_bal_len`, so it introduces **no** new Class-A edge — the baseline is untouched by it, as intended.

One measurement note, since it nearly became a false alarm: a plain `grep -c bal_txs_independent` returns **2** on this file — one comment line and one data row. The gate keys on `(function, jal_or_sink)` columns, so the correct count is the exact column match, which is 1. **Grep the columns, not the line.**

## Follow-up: the prose is now duplicated, and one copy will decay

The sidecar carries the three annotations, and `scripts/bal-class-a-baseline.tsv:15-17` **still carries the same three as comments**. Two sources of truth for one fact, and the TSV copy is the one that `--write-baseline` destroys — which is the whole reason #11451 exists.

That is not a gate failure and not a reason to hold this batch. But it does mean the next regeneration will silently drop the TSV copy, and a reader who remembers tonight's two losses may reasonably conclude the annotations were lost a third time when in fact the sidecar has them. **The move should be completed by deleting the TSV comments**, so there is one home and the gate protects it. Filing that as the residual on #11451 rather than blocking here.

Batch contents otherwise as expected: #11452's one-line terminal gate, #11451's sidecar + ratchet changes + region-map note, and regenerated `GuestAddrs`, `RegionMapLinkPins`, `symbol-addresses.tsv`, `PROGRESS.md`.

Issue #11454, opened 2026-08-04
structural: a resolver family with distinct contracts (whole-account fallible + field-specific) — prerequisite for #11104 snapshot attribution


pirapira:
## Summary

Spin-out of **#11104 item one**, scoped from a read-only producer census and a survivor audit (both recorded on #11104). The guest has no universal whole-account resolver, so no producer can compose a truthful `Optional[Account]` before writing a map row — which blocks #11104's snapshot attribution entirely.

## Why the obvious shortcuts don't work

**Per-producer resolution is not achievable.** Across all 12 nonstorage producers the gap is *complementary*: eleven can resolve balance and/or nonce but **not** code/state; exactly one (CREATE code deposit) can supply code but **not** balance/nonce. So the capability exists in the tree but is scattered per-field and per-path. Making each producer complete would mean twelve resolution paths — the opposite of convergence.

**And the existing helper is the wrong tier.** `account_resolve_pre_state` implements the spec's `block_access_lists.py` `_get_pre_tx_account` shape — block pre-tx map → pre_state, **2 tiers** — and deliberately has no tx-map layer, because it is a *pre-transaction* resolver. Execution reads need `state_tracker.py` `get_account_optional` (`:179-203`): **tx writes → block writes → pre_state, 3 tiers**. Routing execution-time producers through the existing helper would silently read *pre-tx* state where the spec reads *tx* state.

**The spec's per-row completeness is not a property of its producers.** It is a property of there being **one** resolver they all go through, called before every write (`modify_state` = `get_account` then `set_account`, `:631-643`; `set_account` assigns a whole `Optional[Account]`, `:466-486`).

## Scope, in forced order

1. **Add the missing tx-map tier** so precedence matches `get_account_optional`, not `_get_pre_tx_account`. Everything downstream reads the wrong tier without this.
2. **Extend field coverage.** The helper resolves only balance + nonce today — it scans AccountWrite block rows for mask bits 1/2 and copies `nonce@0`, `balance@8..40`, never `storage_root@40..72` or `code_hash@72..104`. Note the representation mismatch: **AccountState needs code ptr/len, not a code hash**, so this requires resolving `code_hash` through the witness code table to bytes/length, with absent / empty / delegation flags represented truthfully.
[...]

PR #11456, opened 2026-08-04
feat(codegen): add separate execution-state account resolver

pirapira:
## Summary

Part one of #11454 adds a separate execution-time account resolver. The pinned execution-specs resolver in `state_tracker.py:179-203` (`get_account_optional`) has three tiers: current transaction writes, block-cumulative writes, then pre-state. The BAL builder instead implements `block_access_lists.py:583-598` (`_get_pre_tx_account`), whose two-tier contract excludes the current transaction map.

`account_resolve_execution_state` scans `tx_account_writes` first and delegates missing balance/nonce components to the unchanged `account_resolve_pre_state`. The builder and all existing callers remain on the pre-transaction resolver. The new symbol has no callers yet; producer routing and field widening are later parts of #11454 and are intentionally not included.

A boolean/mode at a public call site was deliberately avoided. Retargeting the builder to the execution resolver would self-baseline rows from its own transaction map: the rebuild could remain self-consistent while a malformed BAL becomes acceptable. A source guard pins `account_writes_emit_builder_tx` to `account_resolve_pre_state`, documenting that false-accept risk.

## Layout and provenance

- Base: `56aac029da60faa8a8a6a5fabb953f13ed392ac5`
- Parent `.text`: `0x59b7c`; candidate `.text`: `0x59d28` (`+0x1ac`, the uncalled resolver)
- `.data`: `0x5370` unchanged
- `.bss`: `0x19295560` unchanged
- Candidate ELF SHA256: `24063ac623aa55c53cc04820c8e4281ab8e4f6f33f0692602fd9a18a07aab02d` (built 2026-08-04T20:18:32Z)
- Parent ELF SHA256: `639e5714a9bdc14d9d36f435c0fb0952ae9444c438c048f8262e0385eb1efa9b`
- `symbol-addresses.tsv` SHA256: `c113f710c4b385750d5a61e9de104d13691b1cafe3b7003213c14613a63b5a3b`

## Evidence

Prediction before measurement: because the symbol is not wired to a caller in this cut, the candidate is behavior-neutral; the only emitted change is the new resolver body and the builder path remains byte-identical.

Random-200 A/B, same manifest and `spike` backend, seed 4242, jobs 8:

- Parent: 200/200 scored, FA=0, FR=4, 15 unclassified harness exits, 181 exact matches.
- Candidate: 200/200 scored, FA=0, FR=4, 15 unclassified harness exits, 181 exact matches.
- Joined: 198 rows (two shared-input groups), output-byte differences=0.
- Base-PASS -> candidate-FR: empty.
- New false accepts: 0.
- FR label sets: identical (the same four rows).

Commands:
```
scripts/codegen-eest-stateless-check.sh --guest-elf /tmp/opencode/k3-verify/norm-parent-56.elf --backend spike --random --seed 4242 --limit 200 --jobs 8 --quiet-passes --no-verdict-debug --run-dir /tmp/11454-ab-parent
scripts/codegen-eest-stateless-check.sh --guest-elf gen-out/regionmap/stateless_guest.elf --backend spike --random --seed 4242 --limit 200 --jobs 8 --quiet-passes --no-verdict-debug --run-dir /tmp/11454-ab-candidate
python3 scripts/eest-ab-compare.py /tmp/11454-ab-parent /tmp/11454-ab-candidate
```

Checks: full `lake build` (4812 jobs), target build, `check-region-map`, `check-asm-to-program`, `check-unimported`, `check-layering`, `check-guestaddrs-starts`, `check-progress`, `check-drift`, `check-axioms`, and arithmetic fuzz all passed. `check-guarded-handler-bytes` could not run in this artifact-cache checkout because its `lake env lean` subprocess cannot resolve cache-materialized oleans; CI's non-cache environment will exercise it.


pirapira:
Reviewed at `cf1771872`. **The structure, the docstring and the guard are all right. One semantic gap to settle before part three, and it is an interaction with #11453 rather than a defect in this slice.**

## What is right, and the guard is the best part

The two-symbol decision is implemented cleanly: `account_resolve_execution_state` owns **only the first tier** and delegates the block-map/parent fallback to an unchanged `account_resolve_pre_state`, so there is one implementation of the authenticated absence semantics rather than two that can drift.

The docstring cites both spec functions by line and states the FA reasoning outright — *"letting that helper see the tx map would make the builder compare each row against itself and accept a missing BAL entry."* That is the correct framing and it belongs in the source, not just in the PR.

And the guard is exactly the instrument this needed:

```
#guard (accountWritesEmitBuilderTxFunction.splitOn "jal ra, account_resolve_pre_state").length == 2
```

with a comment naming why it exists — *"the emitted bytes can remain self-consistent, so the ordinary build and random A/B gates need not see it."* That is the whole reason a guard is the only protection here: builder self-baselining produces a self-consistent rebuild that builds, passes the ratchet, and can be byte-identical on random-200 while making an invalid block acceptable. A future retarget now fails at compile time.

## ⚠️ The gap: per-field merge does not express a present-`None` tombstone

The spec's precedence is **per key, not per field** (`state_tracker.py:199-203`):

```python
if address in tx_state.account_writes:
    return tx_state.account_writes[address]   # the WHOLE value, possibly None
return get_pre_state_account_optional(tx_state, address)
```

A tx-map hit **terminates the lookup**. This implementation instead merges per field: it sets `s7` bit 1 on a valid balance and bit 2 on a valid nonce, and falls through to the pre-state resolver whenever `s7 != 3`, filling the missing components from the parent.

For *sparse* rows that is the only workable interim behaviour, and it is a reasonable consequence of the representation #11454 exists to fix — I am not asking for it to change in this slice. But it does not handle the case #11453 is introducing:

**A present-`None` tombstone.** #11453 makes deletion a *present row with `optionalState@72 = 0`*, and `account_writes_is_absent` correctly distinguishes missing (0) from present-`None` (1) from present-`Account` (0). This resolver consults **only mask bits 1 and 2** — I count **zero** references to the absence field or to bit 8. So for an account destroyed in the current transaction, the tx row has no valid balance/nonce components, `s7 != 3`, and the resolver **falls through to the authenticated parent and returns the pre-state account** — precisely the three-state conflation that #11453 was careful to avoid one layer up.

That is a real divergence from `get_account_optional`, whose whole point is that a present key *stops* the fallback even when its value is `None`.

## What I want, and it is not a change to this PR

The resolver has **no callers**, which is correct at this stage and which makes this latent rather than live — so this is not a blocker and I am not asking for a patch here. But it must be settled before part three routes any producer through it, because the first producer to hit a same-tx destroyed account would read a resurrected parent account.

Please add to the docstring: **the tombstone case is not yet handled, and the absence tier must be added before any producer is routed through this symbol.** Recording the limitation in the source is what stops part three from being written against an assumption that does not hold. If you would rather handle it now, the check is a present-row test on the absence field *before* the mask-bit tiering, and it must short-circuit rather than merge.

Holding `merge!` only for the convoy (position 4), not for this. Mechanics are otherwise correct: base `56aac029d` at push time, `.text 0x59b7c → 0x59d28` (+0x1ac) with `.data`/`.bss` unchanged, all four generated files present, and no callers added to make the symbol look used.

pirapira:
**Rebased tip `5c3a77fdf` verified — `merge!` applied.**

| check | result |
|---|---|
| base | `cc465b233` (current main), MERGEABLE |
| `main` pin | `0x59b8c` — which is main's earlier `0x59b7c` plus #11452's `0x10`, so main's own arithmetic reconciles |
| this tip | `0x59d38`, delta **`+0x1ac`** |
| stale-base check | **passes** — the pre-rebase value was `0x59d28`; a *different* absolute with the *same* `+0x1ac` delta over main's *new* pin is exactly what a correct rebase-then-regenerate produces |
| generated files | all **four** present |
| `.data` / `.bss` | `0x5370` / `0x19295560` unchanged, as expected for a pure `.text` insertion |

`source-checks`, `summarize`, `path-labels`, `size-label`, `velocity` all pass; only the ~1h `build` leg outstanding.

**And the recorded limitation is exactly right** — `AccountWriteMap.lean:618-626`, Lean-level rather than inside an emitted string, so it costs no bytes:

> *"a missing transaction row may fall through, but a present `None` tombstone must terminate the lookup rather than merge its fields with the block/parent account. Until the absence tier is added, producers must not be routed through this symbol for a transaction that can destroy an account; doing so would resurrect its authenticated pre-state."*

That last clause is the important one: it names the **consequence** rather than just the limitation, so a future reader routing a producer through this symbol learns what breaks, not merely that something is unfinished. Together with the `#guard` pinning the builder to the pre-state resolver, this PR now carries both of its safety properties in the source rather than in review history — which is the point, since neither is visible to any behavioural gate.

Landing state: no callers yet, correctly, and the PR says so rather than adding one to look used.

pirapira:
Folded into #11464, a cross-author batch so these land in one CI round. Merges are `--no-ff`, so per-commit authorship is preserved. Auto-merge disabled here so it can't race the batch; this PR will close when #11464 lands.

pirapira:

pirapira:
Landed via #11466. Verified: commit `5c3a77fdf` is an ancestor of `main` as of `d26cd45a2` — closed rather than dropped.

PR #11457, opened 2026-08-04
proof(#11351): whole-routine triple for header_extract_number

dhsorens:
Machine side of #11351. `headerExtractNumber_spec_within` runs the routine from its entry point to the caller's return address, delegating to `rlp_field_to_u64` at field index 8 (`number`).

The forward (model → guest) stack that this pairs with is already merged: #11434 → #11441 → #11443 → #11444.

## Shape

`header_extract_number` is an eight-instruction wrapper — saves `ra` in a 16-byte frame, moves the output pointer from `a2` to `a3`, pins the field index to 8, calls `rlp_field_to_u64`, restores, returns.

Layering follows `RlpFieldToU64WholeSAsm`, which is the same shape (a wrapper around a single callee), with **each stage a named theorem**:

| piece | role |
|---|---|
| `code` + 4 disjointness facts + 2 `_mono` lifts | `rlpFieldToU64.code.union wrapperCode`, callee-first |
| `prologue` | instructions [0]–[3] |
| `callStep` | instruction [4]; pins the callee's link register to `B + 20` and instantiates `rlpFieldToU64_flat_spec_within` at index 8 |
| `epilogue` | instructions [5]–[7], ending at the caller's `ra` |
| `calleeCtx` | the callee context framed through the prologue |
| `postRest` / `flatPost_split` | factors `x2` out of the callee's success/failure disjunction |
| `setupAndCall` → `headerExtractNumber_spec_within` | the composition |

⚠️ Note `rlpFieldToU64`'s own `setupPrologue`/`restoreAll` are **not** reusable here: they assume a 32-byte frame spilling `s0`–`s5`, whereas this routine spills only `ra` into 16 bytes. The analogues here are correspondingly simpler.

## Three things that are load-bearing

Recording these because each cost real time and none is obvious from the surrounding code:

1. **`calleeCtx` must be an `abbrev`, not a `def`.** `xperm` matches atoms syntactically and treats an opaque definition as a *single* atom, so the prologue's frame and the callee's `flatPre` could never present the same atom list while it was opaque.
2. **Each composition stage needs an explicit statement.** Composing all three segments in one `exact` leaves the intermediate assertion a metavariable, so a mismatch cannot localise — I spent an iteration chasing a "two atoms" symptom that naming `setupAndCall` made obvious. For the same reason the final `cpsTripleWithin_weaken` passes `Q'` explicitly.
3. **`callStep` takes the index as `(8 : Word)`, not `BitVec.ofNat 64 8`.** Propositionally equal, not syntactically, and the prologue's `li a2,8` emits the former.

`flatPost_split` is the analogue of your `allRestoreReady` — the callee hands back `x2` *inside* a disjunction of existentials, so the epilogue cannot take it as a separate conjunct until it is factored out.

## Registry

**No rows yet.** The Correspondence row needs the model tie, which is the follow-up — same staging as the forward-stack PRs.

## Follow-up, and one question

The tie composes this with `success_of_decodeFully_list` (merged) at index 8. The value bridge is definitional: `Result.success` yields `BitVec.ofNat 64 (Nat.fromBytesBE …)` and `SpecRef.bytesBEtoNat` is an `abbrev` for `EL.RLP.Nat.fromBytesBE`.

Ruling out `Result`'s other constructors against `_decode_header = .ok h` gives the row's clause table, and confirms the divergence I flagged on #11351:

- `tooLong` (len > 8) ⇒ needs `p.length ≤ 8`
- `noncanonical` (leading zero) ⇒ needs `p = [] ∨ p.head ≠ 0`
- `empty` ✓ agrees — `bytesBEtoNat [] = 0`
- `success` ✓ definitional

The model's `getN 8` is plain `bytesBEtoNat` and tolerates both over-length and leading-zero fields; the guest rejects both. **Guest stricter than the port, and stricter matches Python** (`rlp.decode_to` enforces these). Those two, plus the arity point, are what make the row `.domainRestricted` rather than `.agrees`.

❓ **`listFailure` is the one case I cannot currently discharge.** It needs `Success → ¬ Failure`, which does not exist in the tree — the natural companion to the determinism lemmas from #11408. My sketch: refute `Failure.init` from `Success`'s `StrictListPayload`, and for `Failure.walk` show that a `StrictPrefix` of `count ≤ index` steps must coincide with the `StrictNthItem` chain (via `rlpItemDecode_deterministic`), so a decode *does* exist at the failing cursor. That in turn wants "a decode at a cursor implies the cursor is in bounds" to refute `WalkFailure`'s first disjunct, and the only thing close is the private `rlpItemDecode_dichotomy` (`WalkNext.lean:3717`).

Before I build all that: does an exclusion lemma already exist somewhere I have not found, or would you rather `rlpItemDecode_dichotomy` be made public and the exclusion live beside the determinism lemmas in `RlpWalkDeterminism.lean`? It would serve #11345 and #11346 too.

## Gates

`lake build` warning-free · `port-check` **PASS** (17 theorems, classical axioms only) · forbidden-tactics / layering / file-size **PASS**.

Refs #11351

### DIFF
--- a/EvmAsm/Codegen/Programs/HeaderExtractNumberSAsm.lean
+++ b/EvmAsm/Codegen/Programs/HeaderExtractNumberSAsm.lean
@@ -1,0 +1,456 @@
+/-
+  EvmAsm.Codegen.Programs.HeaderExtractNumberSAsm
+
+  Machine side of #11351.  `header_extract_number` is an eight-instruction
+  wrapper that saves `ra` in a 16-byte frame, moves the caller's output
+  pointer from `a2` to `a3`, pins the field index to 8, tail-calls
+  `rlp_field_to_u64`, then restores and returns.
+
+  The code-region shape mirrors `RlpFieldToU256BeSAsm`: this routine's own
+  program unioned with the whole `rlp_field_to_u64` closure.  The frame is
+  smaller than `rlpFieldToU64`'s own (only `ra` is spilled, into 16 bytes
+  rather than `s0`-`s5` into 32), so that routine's `setupPrologue` /
+  `restoreAll` primitives are frame-shaped for it and are not reusable here;
+  the analogues below are correspondingly simpler.
+-/
+
+import EvmAsm.Codegen.Programs.HeaderU64
+import EvmAsm.Codegen.Programs.RlpFieldToU64FlatSAsm
+
+namespace EvmAsm.Codegen.HeaderExtractNumberSAsm
+
+open EvmAsm.Rv64 EvmAsm.Rv64.SAsm EvmAsm.Rv64.RLP
+
+abbrev B : Word := (GuestAddrs.header_extract_number : Word)
+
+theorem program_length : headerExtractNumber_prog.length = 8 := by decide
+
+def wrapperCode : CodeReq := CodeReq.ofProg B headerExtractNumber_prog
+
+/-- The entire `rlp_field_to_u64` closure this wrapper calls, plus the wrapper.
+    Callee first, matching `RlpFieldToU256BeSAsm`: `CodeReq.union` is
+    left-biased, and `mono_union_right` is stated for that orientation. -/
+def code : CodeReq := EvmAsm.Codegen.RlpFieldToU64SAsm.code.union wrapperCode
+
+/-! ## Code-region disjointness -/
+
+theorem u64Wrapper_disjoint :
+    EvmAsm.Codegen.RlpFieldToU64SAsm.wrapperCode.Disjoint wrapperCode := by
+  unfold EvmAsm.Codegen.RlpFieldToU64SAsm.wrapperCode wrapperCode
+    EvmAsm.Codegen.RlpFieldToU64SAsm.B B
+  apply CodeReq.Disjoint.ofProg_ranges
+  · rw [EvmAsm.Codegen.RlpFieldToU64SAsm.program_length]; decide
+  · rw [program_length]; decide
+  · rw [EvmAsm.Codegen.RlpFieldToU64SAsm.program_length, program_length]; decide
+
+theorem nthWrapper_disjoint :
+    EvmAsm.Codegen.RlpListNthItemSAsm.code.Disjoint wrapperCode := by
+  unfold EvmAsm.Codegen.RlpListNthItemSAsm.code wrapperCode
+    EvmAsm.Codegen.RlpListNthItemSAsm.B B
+  apply CodeReq.Disjoint.ofProg_ranges
+  · rw [EvmAsm.Codegen.RlpListNthItemSAsm.total_length]; decide
+  · rw [program_length]; decide
+  · rw [EvmAsm.Codegen.RlpListNthItemSAsm.total_length, program_length]; decide
+
+theorem contentWrapper_disjoint :
+    EvmAsm.Codegen.RlpFieldToU64SAsm.contentCode.Disjoint wrapperCode := by
+  unfold EvmAsm.Codegen.RlpFieldToU64SAsm.contentCode wrapperCode
+    EvmAsm.Codegen.RlpFieldToU64SAsm.C64B rlp_content_to_u64_code B
+  apply CodeReq.Disjoint.ofProg_ranges
+  · rw [rlp_content_to_u64_prog_length]; decide
+  · rw [program_length]; decide
+  · rw [rlp_content_to_u64_prog_length, program_length]; decide
+
+theorem calleeCode_wrapper_disjoint :
+    EvmAsm.Codegen.RlpFieldToU64SAsm.code.Disjoint wrapperCode := by
+  unfold EvmAsm.Codegen.RlpFieldToU64SAsm.code
+  exact CodeReq.Disjoint.union_left u64Wrapper_disjoint
+    (CodeReq.Disjoint.union_left nthWrapper_disjoint contentWrapper_disjoint)
+
+theorem wrapperCode_mono : ∀ a i, wrapperCode a = some i → code a = some i := by
+  intro a i hi
+  unfold code
+  exact CodeReq.mono_union_right calleeCode_wrapper_disjoint (fun _ _ h => h) a i hi
+
+theorem calleeCode_mono :
+    ∀ a i, EvmAsm.Codegen.RlpFieldToU64SAsm.code a = some i → code a = some i := by
+  intro a i hi
+  unfold code
+  simp [CodeReq.union, hi]
+
+/-! ## Prologue: instructions [0]-[3]
+
+`addi sp,sp,-16 ; sd ra,0(sp) ; mv a3,a2 ; li a2,8`.  The output pointer
+arrives in `a2` and the callee expects it in `a3`, and the field index 8 is
+what makes this routine `number` rather than any other `_decode_header`
+field. -/
+
+theorem prologue
+    (spOuter raIn outPtr old13 oldSlot newSp : Word)
+    (hnewSp : newSp = spOuter + signExtend12 (-16 : BitVec 12)) :
+    cpsTripleWithin 4 B (B + 16) code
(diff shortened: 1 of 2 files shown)


dhsorens:
**Narrowing my own question before you spend time on it** — two of the three parts are now self-answered.

**1. The exclusion lemma genuinely does not exist.** I searched again on every phrasing I could think of; every hit *constructs* `Failure` (the `AccountDecodeClose*` files, via `DecodeFailure.field{1,2,3}List`) and none excludes it. So `Success → ¬ Failure` is new work, not something I failed to find.

**2. `rlpItemDecode_dichotomy` does NOT need to be made public.** I was wrong that I needed it. `rlpItemDecode_advance` (`BalAccountNonstorageFinalsWalk.lean:71`) is already public and gives exactly the bound:

```lean
off < (next - base).toNat ∧ (next - base).toNat ≤ endOff
```

under `off ≤ endOff` and the no-overflow hypothesis — which refutes **both** of `WalkFailure`'s disjuncts, not just the second. `StrictPrefix.step_bounds` already uses it the same way. Please disregard that part of the ask.

**3. Only the placement question remains, and it is minor.** My plan:

- `strictNthItem_extends_prefix` — a `StrictPrefix` of `count ≤ index` steps coincides with the `StrictNthItem` chain, so `∃ next' len', StrictNthItem bytes base endPtr (index - count) off next' len'`. Induction on the `StrictPrefix` derivation, using `rlpItemDecode_deterministic` to identify the two decodes at each cursor.
- `success_not_failure` — `Failure.init` contradicts `Success`'s `StrictListPayload`; `Failure.walk` gets `strictListPayload_deterministic` to align the payloads, then the above plus `rlpItemDecode_advance` to contradict `WalkFailure`.

I will put both in `RlpWalkDeterminism.lean`, beside the determinism lemmas from #11408 that they build on, since they are the same kind of fact and #11345/#11346 will want them too. Say the word if you would rather they lived elsewhere — moving them is trivial and I will not block on it.

This PR (#11457) is unaffected either way: it is the machine triple only, and the exclusion lemma belongs to the follow-up that adds the registry rows.

pirapira:
Reviewed at the PR head on base `cc465b233`. **Kernel-clean and well-structured. One question about the step bound that I'd like answered before labelling, and it may well have a good answer.**

## What's right

- **Zero `sorry`, zero `native_decide`, zero `bv_decide`** — I grepped the whole file and the count is 0, so the TCB stays at the three classical axioms. `program_length` closes by `decide`, which is the right tool here.
- **Decomposition is the conventional one and reads cleanly**: `prologue` → `callStep` → `epilogue`, composed through `setupAndCall` into `headerExtractNumber_spec_within`, with the wrapper-disjointness and `wrapperCode_mono` / `calleeCode_mono` lemmas doing the code-extension plumbing rather than ad-hoc rewriting.
- **The frame discipline is explicit** — `calleeCtx_pcFree`, `postRest_pcFree` and `flatPost_split` isolate the pc-freedom and post-split obligations instead of burying them in the main proof, which is what makes this maintainable when the callee's layout moves.
- Preconditions are stated rather than assumed: 8-byte alignment (`hsalign`), `listLen + 9 ≤ bytes.length` slack, no-overflow (`hover`), per-byte validity (`hvalid`), and the return-address alignment (`hret`).

## The question: `tailSteps` carries a `2 ^ 64 - 1` factor

```
let tailSteps := (7 + (1 + (7 * (2 ^ 64 - 1) + 11))) + 5
```

This appears at `:163`, `:368` and `:419`. A `cpsTripleWithin` bound is an upper limit, so a loose bound is **sound** — the postcondition still holds if the routine terminates within it. My question is what it *witnesses*.

A bound of ~1.3 × 10²⁰ steps is derived from the **range of a 64-bit counter**, not from the data. So as written, the theorem says "if this terminates within 2⁶⁴-ish steps, the postcondition holds" — which is true but does not establish that the tail loop is bounded **by its input**. For a guest routine the useful statement is normally in terms of `bytes.length` or `listLen`, both of which are already in scope and already constrained (`hslack`, `hover`).

Three possibilities, and I can't tell which from the file:

1. The tail loop genuinely has no data-derived bound (it exits on a sentinel), in which case `2^64 - 1` is the honest worst case and this is fine — but it deserves a comment saying so, because the next reader will ask the same question.
2. A tighter bound in terms of `listLen` is derivable and was simply not pursued. That's a reasonable trade-off to defer, but it should be stated as deferred rather than left implicit.
3. The bound is loose because a lemma upstream produced it, and tightening it is mechanical.

**Which is it?** If (1), a one-line comment at the `tailSteps` definition closes this. If (2) or (3), I'd rather know it's a known looseness than have it read as a tight bound.

The reason I'm asking rather than waving it through: a step bound is the part of a `Within` triple that carries the operational content, and `2^64` is exactly the value that makes a bound *technically* present but *practically* uninformative. That's worth one sentence of provenance in the source, not because the proof is wrong, but because the next person to compose against this theorem will inherit the bound and may not notice how loose it is.

Nothing else blocks. MERGEABLE on current main, and the file is self-contained apart from the one `Imports.lean` line.

dhsorens:
**Closing this — it duplicates work already on main, and that is my error.**

`EvmAsm/Codegen/Programs/HeaderExtractNumberSpec.lean` has been on main since 27 July (`c67f0a988`, *"proof: header_extract_number whole-program caller contract"*). It contains `header_extract_number_spec_within` with the **same** decomposition I rebuilt here — prologue ;; `rlpFieldToU64_flat_spec_within` (index 8) ;; epilogue — the same code-region setup (`hdrCode` / `fullCode` / `hdr_mono` / `k34_mono` / `hdr_disjoint`), the same `flatPost` case-split (`hdrEpilogue` routing each arm through `epiCore`, which is exactly what I reinvented as `flatPost_split` + `postRest`), and the same step count `4 + (1 + n34) + 3`.

**Root cause, since it is worth recording:** my initial survey ran

```
grep -rn "headerExtractNumber\|header_extract_number" EvmAsm/ --include="*.lean" | head -40
```

and `HeaderExtractNumberSpec.lean` fell off the end of a truncated result. I then confirmed "no `header_*` routines in `Routines.lean`" — which is true, the row is genuinely absent — and wrongly read that as "no spec exists". A missing *registry row* is not evidence of a missing *proof*. I should have re-run the search untruncated before writing 456 lines.

Nothing here is worth salvaging over the existing file; it is equivalent, and yours is older and already integrated.

## Your step-bound question — answered, and it applies to the file on main

Worth answering anyway, because the bound you asked about is in `HeaderExtractNumberSpec.lean` too, verbatim.

It is your option **(3): inherited, not chosen here.** Provenance:

| where | what |
|---|---|
| `EvmAsm/Rv64/RLP/Field0ToU64Top.lean:215` | origin — `cpsTripleWithin (7 * (2 ^ 64 - 1) + 18)`, the `rlp_content_to_u64` scalar loop |
| `RlpFieldToU64FinishSAsm.lean:843, 965` | `tailSteps := (7 + (1 + (7 * (2 ^ 64 - 1) + 11))) + 5` |
| `RlpFieldToU64FlatSAsm.lean:74–75` | carried into `rlpFieldToU64_flat_spec_within` |
| both header wrappers | copied verbatim from the callee's statement |

So the `2^64 - 1` is the **content-decode loop's** worst case, bounded by the range of its 64-bit length register rather than by the data — and no wrapper can tighten it without tightening `Field0ToU64Top` first. A caller-side comment would be the wrong place; the honest note belongs at the origin.

A data-derived bound does look reachable in principle: the loop is bounded by the field length, which `Result`/`successPayload` already constrain to `≤ 8` on the success path, and `hslack : listLen + 9 ≤ bytes.length` is in scope throughout. But that is a change to the `rlp_content_to_u64` proof, not to either wrapper, so I would rather not smuggle it in here. Happy to file it as its own issue if you want it tracked — say the word and I will, with the trace above.

#11460 (`success_not_failure`) is unaffected and **not** a duplicate — I re-checked untruncated this time: zero hits anywhere in the tree. I have retargeted it onto `main` so it no longer depends on this branch.

dhsorens:
Superseded by the pre-existing `HeaderExtractNumberSpec.lean` on main (see previous comment). Closing as duplicate.

PR #11458, opened 2026-08-04
Converge universal dispatcher transaction path

### LINKED ISSUE #11423: structural: unify the direct-transaction and MTx paths — the spec has one path, the guest has two

pirapira:
Closes #11423\n\nSummary:\n- Remove runtime_mtx_active as a semantic dispatcher switch; retain only its reserved BSS slot for layout stability.\n- Make dispatcher retention and auth checkpoint behavior universal; remove the MTx writer.\n- Preserve the single universal transaction boundary and existing rollback semantics.\n\nVerification:\n- Candidate commit: 78232b4b4.\n- Candidate ELF SHA256: 074d74996ecab9cad9d12189a154e9929a62e18269b7c0189f14f121bef70e69.\n- Base SHA: 42d3dfaa42dc624d8cd835406b3fb85dab93c765; base ELF SHA256: 04ec3813e4840042d778666de2e626a33a37a461db92083df68cfc8aaee79576.\n- Spike, fixture tests-zkevm@v0.6.2, random seed 1142301, 200 rows: both base and candidate selected 200, ran 188, full/oracle matched 180, false accepts 0, false rejects 8, errors 12.\n- scripts/eest-ab-compare.py: identical 8-label FR sets, zero new false accepts, zero status/output-byte differences, byte-identical on all 198 joined cases.\n- env LAKE_ARTIFACT_CACHE=false lake build EvmAsm.Codegen.Programs.TxIntrinsicStateGas passed; git diff --check passed.\n\nThe eight inherited FRs are all bv_fail=1: 3b3b14d23f9efaff1f3390ff9f08691e0016cec73475305bb8d4dd2395ba4bde, dcfc299fdadbdc2ee06c8c363419c78324f7b1f93379869f60abb242e36fc789, 9de3629888f6159c90825061483e8de9ec4484458b17f3d2c3a1a9f4559026d2, 66c95f5922d3d551572097a768eae63b8e72e225bf4a0941e30c6186b1b0999c, 04e9cf6b5fbd64d6a5023e6083f625e44c37c1c0c209fe22d9a7696447b6300d, df7ff3f7ccaf414242ac50a2187d65044e32b08be61e271cb07516a745638c1d, 0289eb3d463a85840803ce3369c826cb006e9b03229d16cb5313f903d0fcbad5, 1b6ef67658467d4b15e05246c9080c3a62455c24bc28c6a153be337fe688cdbd. No rebase performed per convoy sequencing.


pirapira:
**The structural change and the evidence are both right. Predicting a CI failure on missing pins — please regenerate before the gates report.**

## Missing generated files

The diff is 4 source files (26+/34-), of which **12 emitted asm lines are removed and 1 added** — a net −11 emitted lines. That moves `.text`, so `RegionMapLinkPins.lean` (`textSizeBytes`), `scripts/asm-fixtures/symbol-addresses.tsv` and `EvmAsm/Codegen/GuestAddrs.lean` all need regenerating, and none is present.

MERGEABLE here means only "no textual conflict" — it says nothing about whether the pins match the layout. This is the third instance tonight of the same shape (#11452 twice), so the habit worth adopting is the question rather than the checklist: **which generated files does a `.text` shift invalidate?** Four, because `RegionMapLinkPins` and `GuestAddrs` independently encode addresses that the TSV also encodes. The `GuestAddrs declared-start match` gate is what caught it last time, in 10 seconds.

Also missing from the report: **the layout delta itself**. Please state main's `textSizeBytes`, the candidate's, and the difference — and note the falsifiable form, since it caught a stale-base regen earlier tonight: the new pin must **not** equal any value seen on this branch before.

## The evidence is exactly what the structural bar asks for

Base and candidate on the same backend, same seed, same 200 rows: **FA=0 FR=8 on both legs**, 198 joined, **identical 8-label FR sets, zero flips either direction, byte-identical on all 198**, with base ELF `04ec3813…` and candidate `074d7499…`. All eight FRs are **inherited**, so nothing is candidate-only. That satisfies "structure matches spec, random-200 not deteriorating" without byte-equivalence or FR=0.

And the FR list was reported correctly: **eight rows individually, each with its fixture path and `tx_state0`**, not collapsed into one class because they share `bv_fail=1`. That distinction was the caveat I raised, and it was honoured.

## ⭐ A pattern in that FR set worth extracting as a lead

**All eight FRs are `eip7702_set_code_tx` fixtures.** Not a mixed population — the entire FR mass of this sample is one EIP family. Decoding `tx_state0` by ÷1530 (the per-state-byte charge) gives byte counts:

| `tx_state0` | bytes | reading |
|---|---|---|
| 35190 | **23** | `AUTH_BASE` alone — one authorization, nothing else |
| 133110 | **87** | 64 + 23 = one `STORAGE_SET` + one `AUTH_BASE` |
| 218790 | **143** | appears **four** times, across three distinct fixtures |
| 231030 | **151** | = 143 + 8 |
| 535500 | **350** | |

Two things stand out: 23 and 87 decompose exactly into known constants, and **143 recurs across three different fixtures** — a repeated total across distinct tests points at a shared computation rather than three coincidences. That is a much sharper starting point than "eight rows fail," and it belongs on the 7702 FR lane rather than being spent here.

To be clear about what this is: these FRs are **inherited, not caused by this PR**, so none of it blocks. It is a lead for whoever takes the 7702 family, and it should be recorded there.

`merge!` held only for the pins and for convoy position — the change itself (removing `runtime_mtx_active` as a semantic switch, retaining the reserved BSS slot for layout stability) is what #11423 asked for, and the three-read FA screen was answered in the safe direction before it was written.

pirapira:
**Repin verified on `48fa0f96c` — `merge!` applied.**

| check | result |
|---|---|
| base | `88890ced1` (current main), MERGEABLE |
| `main` pin | `0x59f00` |
| this tip | `0x59eac`, delta **`−0x54`** = 84 bytes = 21 instructions |
| stale-base check | **passes** — differs from every prior value on this branch, which is what a regen against the *current* base produces |
| generated files | `GuestAddrs`, `RegionMapLinkPins`, TSV — the full set alongside 4 source files |
| fixed point | second complete regen cycle checksummed identical |

The 84-byte shrink reconciles with the source change: ~11 net emitted *lines* removed, but a single Lean string line carries several semicolon-separated instructions, so 21 instructions from 11 lines is the expected ratio rather than a discrepancy.

`source-checks`, `summarize`, `path-labels`, `size-label`, `velocity` pass; only the ~1h `build` leg outstanding.

## On the evidence, which was settled earlier

Same backend, same seed, same 200 rows on both legs: **FA=0 FR=8 on base and candidate**, 198 joined, identical 8-label FR sets, zero flips either direction, byte-identical across all 198 — with all eight FRs **inherited**. That meets the structural bar (structure matches spec, random-200 not deteriorating) without byte-equivalence or FR=0.

The change itself is what #11423 asked for: `runtime_mtx_active` retired as a semantic switch, its MTx writer removed, dispatcher retention and the auth checkpoint made universal, and only the reserved BSS slot kept for layout stability. The three-read FA screen was answered in the safe direction before the patch was written — and the one that mattered was recognising that the auth-cursor checkpoint is **rollback metadata rather than a comparison selector**, so it cannot make a later mismatch pass.

## The FR pattern from this sweep is a lead, recorded on the 7702 lane

All eight FRs are `eip7702_set_code_tx` fixtures, and ÷1530 makes them legible as byte counts: **35190 = 23 B** (`AUTH_BASE` alone), **133110 = 87 B** (64 + 23 = one `STORAGE_SET` + one `AUTH_BASE`), and **218790 = 143 B recurring four times across three distinct fixtures**. Exact decomposition into known constants plus a total repeating across different tests points at a shared computation rather than coincidence. Inherited, so out of scope here — but a far sharper starting point than "eight rows fail."

pirapira:

PR #11460, opened 2026-08-04
proof(#11351): Success excludes Failure for rlp_list_nth_item

dhsorens:
**Stacked on #11457.** This is the lemma I asked about there — built, since the search settled the only part that needed your knowledge (it genuinely does not exist), and the placement question is cheap to reverse.

`rlp_list_nth_item` reports exactly one of `Success` and `Failure`, but nothing in the tree said so. Every existing use *constructs* a `Failure` (`AccountDecodeClose*`, via `DecodeFailure.field{1,2,3}List`); none excludes one. The model → guest direction needs it, to rule out `Result.listFailure` once the model has already established `Success`.

## Contents

| lemma | role |
|---|---|
| `strictNthItem_head` | any chain, any arity, starts with a decode at its own cursor |
| `strictPrefix_le` | a walked prefix never runs past the list's declared end |
| `strictNthItem_extends_prefix` | a prefix of `count ≤ index` steps lands **on** the chain — what remains is a chain of arity `index - count` |
| `success_not_failure` | the exclusion |

`strictNthItem_extends_prefix` is the substance: induction over the prefix, identifying the prefix's decode with the chain's at each cursor via `rlpItemDecode_deterministic`. `success_not_failure` then refutes `Failure.init` from `Success`'s `StrictListPayload`, and `Failure.walk` by aligning the payloads with `strictListPayload_deterministic` and showing a decode *does* exist at the allegedly-failing cursor.

## ⭐ Correction to what I asked on #11457

**`rlpItemDecode_dichotomy` does not need to be made public.** I was wrong. `rlpItemDecode_advance` (`BalAccountNonstorageFinalsWalk.lean:71`) is already public and yields

```lean
off < (next - base).toNat ∧ (next - base).toNat ≤ endOff
```

which refutes **both** `WalkFailure` disjuncts — the cursor-past-end one as well as no-item-decodes. Please disregard that part of the ask; nothing needs to change in `WalkNext.lean`.

## One thing worth a reviewer's eye

The bound on the failing cursor **cannot** be derived from that cursor itself — attempting it is circular, and my first version was (it tried to get `off ≤ listLen` by contradiction from the very `ult` fact it was trying to refute). It has to be threaded from the list's start offset (`StrictListPayload.cursor_le`) *forward* through the prefix. That is what `strictPrefix_le` exists for, and it is the only reason that lemma is separate.

## Placement

`RlpWalkDeterminism.lean`, beside the determinism lemmas from #11408 that these extend, since #11345 and #11346 will want them too. Say the word if you'd prefer elsewhere — moving is trivial.

## Gates

`lake build` warning-free · `port-check` **PASS** (12 theorems, classical axioms only) · forbidden-tactics / layering / file-size **PASS**.

## Next

With this, the #11351 model tie has all its pieces: compose `headerExtractNumber_spec_within` (#11457) with `success_of_decodeFully_list` at index 8, discharge `Result`'s five constructors, and add the registry rows graded `.domainRestricted`.

Refs #11351, #11345, #11346

### DIFF
--- a/EvmAsm/Codegen/Programs/RlpWalkDeterminism.lean
+++ b/EvmAsm/Codegen/Programs/RlpWalkDeterminism.lean
@@ -25,6 +25,7 @@
 -/
 
 import EvmAsm.Rv64.RLP.WalkItemDeterminism
+import EvmAsm.Rv64.RLP.ItemDecodeForward
 import EvmAsm.Codegen.Programs.RlpListNthItemSAsmBase
 
 namespace EvmAsm.Codegen.RlpListNthItemSAsm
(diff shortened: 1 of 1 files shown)


pirapira:
Reviewed at the head on base `cc465b233`. **Kernel-clean, and the proof structure is the right one for this obligation. `merge!` applied.**

## What it establishes, and why the shape is right

The headline is `success_not_failure`: `Success → Failure → False`. That is the obligation your own narrowing comment identified as genuinely absent — every existing hit *constructs* `Failure` (the `AccountDecodeClose*` files via `DecodeFailure.field{1,2,3}List`) and none excludes it. So this is new work rather than a rediscovery, and confirming that before building it was the right order.

The route to it is the informative part: rather than attacking the exclusion directly, it goes through **determinism**, and that is what makes the proof short.

- `strictListPayload_deterministic` — the short/long case analysis, where the two cross cases close by deriving `b = b'` from `Option.some.inj` and then `absurd hshort hlong'`. Clean, and it is the step that makes everything downstream single-valued.
- `strictNthItem_deterministic` and `success_deterministic` build on it.
- `strictNthItem_head` / `strictNthItem_extends_prefix` / `strictPrefix_le` are the bridging lemmas, and `strictPrefix_le` is what converts a pointer-level prefix relation into the `off ≤ listLen` arithmetic the bound needs.

Then `success_not_failure` refutes **both** `Failure` disjuncts from one witness: determinism forces `hlist'` and `hlist` to agree, the prefix extension yields a decode at `off`, and that decode refutes the "no item decodes" arm directly while `rlpItemDecode_advance`'s bound refutes the "cursor past end" arm. Refuting both arms from a single decode witness is why this is 81 lines rather than several hundred.

**Zero `sorry`, `native_decide`, `bv_decide`** — grepped, count is 0, so the TCB stays at the three classical axioms. `source-checks`, `summarize`, `path-labels`, `size-label` and `velocity` pass; only the ~1h `build` leg is outstanding.

Also worth noting: reusing the already-public `rlpItemDecode_advance` (`BalAccountNonstorageFinalsWalk.lean:71`) rather than widening `rlpItemDecode_dichotomy`'s visibility is the better call, and you reached it by re-checking rather than by assuming — one fewer public surface for the same result.

## One note carried over from #11457, which is not about this PR

You closed #11457 as a duplicate of `HeaderExtractNumberSpec.lean` (on main since `c67f0a988`, 27 July), and that was the right call. But the step-bound question I raised there **applies to main's version too** — I checked, and `HeaderExtractNumberSpec.lean:392` carries the identical

```
let tailSteps := (7 + (1 + (7 * (2 ^ 64 - 1) + 11))) + 5
```

So the question outlives the closed PR: a `cpsTripleWithin` bound of ~1.3 × 10²⁰ is sound but derived from the **range of a 64-bit counter** rather than from the data, even though `bytes.length` and `listLen` are in scope and constrained. It does not establish that the tail loop is bounded by its input, and anyone composing against `header_extract_number_spec_within` inherits that looseness.

Not a defect and not this PR's business — but worth either a comment at that definition recording it as the honest worst case, or an issue if a `listLen`-derived bound is derivable. Your call which; I would rather it be written down than rediscovered.

dhsorens:
Thanks — and agreed the step-bound question outlives #11457. **Done both**, since they serve different readers:

**Issue #11462** records it at the layer that can actually fix it — `Field0ToU64Top.lean:215`, the `rlp_content_to_u64` scalar loop — with the full propagation trace and the observation that the success path already constrains the field to ≤ 8 bytes (`Result.success`'s `hfit`, with `Result.tooLong` being exactly the arm that rejects wider), so a data-derived bound looks reachable. I was careful to flag that I have **not** verified it is achievable; if the loop's termination argument is genuinely counter-based then `2^64 - 1` is the honest worst case and the right outcome is a comment at the origin, not a tightening.

I deliberately did **not** bundle it into the `header_extract_number` work: a caller-side comment would document an upstream property in the wrong place, and no wrapper can tighten it anyway.

**A provenance comment** at `HeaderExtractNumberSpec.lean:392` will ride along with the model-tie PR, pointing at #11462 — so the next person composing against `header_extract_number_spec_within` sees the looseness where they will actually hit it, rather than having to trace four files back to find out it was inherited.

pirapira:
Folded into #11464, a cross-author batch so these land in one CI round. Merges are `--no-ff`, so per-commit authorship is preserved. Auto-merge disabled here so it can't race the batch; this PR will close when #11464 lands.

pirapira:

PR #11463, opened 2026-08-04
fix(codegen): assert type-4 auth prepared before dispatch resolve

pirapira:
## Summary

Closes the residual of **#11419** after **#11427**.

**Call:** assert the invariant, do **not** dual-write auth.

11427 already put the sole live entry (`BlockVerdictMtxRuntime` → `dispatch_tx_runtime_code`) on survivor order (`set_delegation` then resolve). The inversion remaining in `dispatch_tx_runtime_code` was **latent in source**, not live in behaviour (`runtime_tx_auth_prepared=1` on every reachable type-4 path).

A second conditional `eip7702_auth_state_prepare` at dispatch entry would be:
- two writers + idempotence guard (hybridising), and
- an **unreachable** arm today (no caller arrives with `prepared=0`) — cannot be sweep-validated.

**This PR instead:**
1. At `dispatch_tx_runtime_code` entry: type-4 requires `runtime_tx_auth_prepared≠0`, else fail closed `a0=9` (`.Ldtrc_auth_not_prepared`).
2. Comment cites `interpreter.py:356-365` (pin e5a8caf1b) and names MTx `inline_prepare` as the sole writer.
3. Removes the latent late `runtime_tx_auth_exec_fn = eip7702_auth_state_prepare` install (was the resolve-before-auth path when `prepared=0`).
4. Does **not** touch nested CALL resolve (depth>0); no BAL code-source fallback; Class-A parse-bail at `.Ldtrc_deferred_marker_ready` kept.

Why assert still converges on the spec: guest must run auth **before recipient classification** at the MTx boundary (00001 install-on-empty-recipient). Auth therefore lives at `inline_prepare`, not inside dispatch. Dispatch is the `prepare_dispatch` side and correctly **requires** prior `set_delegation`.

## Layout

- `textSizeBytes` `0x59b8c` → `0x59b88` (−4 B: assert cheaper than removed late-auth arm)
- Guest ELF sha256 `0c22c99f47fd16e6f2721732e5355f0994ee59d1bf9c7e309fd747257cb12545`
- region-map CLEAN; class-a OK 15; second pin pass stable

## Evidence

Focused type-4 install fixtures (spike, this guest): all **succ=01**
- 00151 `tx_installs_delegation_on_empty_recipient` (no…)
- 00152 `tx_installs_delegation_on_empty_recipient` (ze…)
- 00153 `tx_installs_delegation_on_funded_recipient`
- 00155 `tx_installs_delegation_on_sender`

## FA screen

Assert is rejection-only on the new arm (status 9). Live paths unchanged behaviourally (prepared already 1). Invalid-accept risk from reordering does **not** apply — no resolve/execute order change on reachable paths. Marker-exec (ef0100 / root MATCH + succ FAIL) is the 11406 signature to watch if anything drifts; not expected here.

## Test plan

- [x] `lake build` BlockVerdictDispatchTx
- [x] layout regen + `check-region-map` CLEAN
- [x] class-a ratchet OK 15
- [x] focused 7702 install fixtures PASS
- [ ] CI full gate
- optional r200 — structural bar, land with FR inventory if any

No merge! from grok.


pirapira:
**This is the right shape, and it does more than the assert I proposed — it removes the inverted path rather than guarding around it.** Two mechanical items before labelling; the design is settled.

## Why the assert was the correct call, in your own framing

You made the judgement I asked for and the reasoning is better than mine: **`dispatch_tx_runtime_code` is the `prepare_dispatch` side**, so requiring `set_delegation` to have already run *is* the spec's contract at `interpreter.py:356-365`, not a workaround for one. That reframes the assert from "check a caller invariant" to "state the routine's precondition", which is what makes it convergence rather than defence.

And you rejected the dual-writer patch for the reason that matters: a second auth arm at dispatch entry would be **unreachable today**, so no sweep and no fixture could validate it — the #11407 lesson, where an unexercised path hid an FA-lax divergence that only a differential could see. Here there is not even a reachable row to difference against.

## What the diff actually does, and the part I want to credit

- **Entry precondition**: reads the tx type at `160(s2)`, skips for non-type-4, then requires `runtime_tx_auth_prepared` nonzero — otherwise `.Ldtrc_auth_not_prepared` → `li a0, 9`. **Fails closed, and never resolves or executes.** That is the right direction: an unmet precondition produces a rejection, not a guess.
- **Removes the latent late install.** The old line installed `runtime_tx_auth_exec_fn = eip7702_auth_state_prepare` whenever `!prepared`; it now only zeroes the callback and preserves the inner envelope and type for gas/warm bookkeeping. So the resolve-before-`set_delegation` path is **gone from the source**, not merely unreachable — which was the whole complaint in #11419, since the source was documenting the wrong order.
- **Comments cite the pin** (`interpreter.py:356-365`, `e5a8caf1b`) and record *why not* a dual writer. That belongs in the source, and it is what stops the next reader "fixing" the assert by adding the second writer back.

⭐ **And the detail worth calling out: you added `j .Ldtrc_ret` after `li a0, 7`.** Inserting `.Ldtrc_auth_not_prepared` immediately after that instruction would otherwise have made the success path **fall through into the failure arm**. That is exactly the defect class k3 hit on the #11378 family earlier tonight — fall-throughs reaching a newly inserted block — and catching it while writing the patch rather than while debugging it is the difference between a clean landing and an evening.

## Two mechanical items

1. **The pin is stale.** `RegionMapLinkPins` here says `0x59b8c → 0x59b88`, computed against `cc465b233`. Main is now `88890ced1` (batch #11459 landed #11426 + #11453), so both numbers will move. Recompute on the rebase, and the falsifiable check applies: **the new pin must not come back as `0x59b88`** — if it does, the regeneration used a stale base.
2. **`GuestAddrs.lean` is present, which is right** — all four generated files are in the diff (`GuestAddrs`, `RegionMapLinkPins`, TSV, plus the source). That is the full set for a `.text` shift, and it is the set that #11452 and #11458 each got wrong once tonight.

This PR has **no `merge!`**, so it is yours to rebase — the hands-off rule applies only to labelled PRs. I will label once the pins are recomputed and the fast gates are green.

`.text` net −4 is plausible for the change (an ~3-instruction precondition and a 1-instruction fail arm added, a ~10-instruction install removed, one `j` added), but state main's value, yours, and the delta explicitly when you repin rather than leaving it to be inferred.

pirapira:
Rebased onto main `88890ced1` (batch 11459 landed 11426+11453). Layout regen:
- textSizeBytes **0x59efc** (post-11453 base)
- guest sha256 `e3ec927c9bc183653f5fbd44ff3e65adb17efca6cf3f5e8924669081105bd6c7`
- region-map CLEAN; class-a OK 13; pins stable NO-OP
- 00151 empty-recipient still succ=01

HEAD `c02c55dec`. No merge! from grok.

pirapira:
**Rebased tip `c02c55dec` verified on `88890ced1` — `merge!` applied.**

| check | result |
|---|---|
| `main` pin | `0x59f00` |
| this tip | `0x59efc`, delta **`−4`** |
| stale-base check | **passes** — the pre-rebase value was `0x59b88`; getting that back would have proven a stale base |
| generated files | `GuestAddrs`, `RegionMapLinkPins`, TSV, alongside the source — full set for a `.text` shift |
| substance | assert present (2 references to `.Ldtrc_auth_not_prepared`), late `auth_exec` install **gone** (0 occurrences), `li a0, 9` fail-closed arm and the `li a0, 7; j .Ldtrc_ret` guard both intact |

`source-checks`, `summarize`, `path-labels`, `size-label`, `velocity` pass; only the ~1h `build` leg outstanding. The `−4` delta is consistent with the change: a ~3-instruction precondition and a 1-instruction fail arm added, a ~10-instruction install removed, one `j` added.

I re-checked the substance after the rebase rather than assuming it carried, because this PR's value is entirely in two small things that a conflict resolution could quietly drop — and both survived.

## What this closes, and the reasoning worth keeping

#11419's complaint was never that behaviour was wrong — #11427 had already put the sole live caller on the correct order. It was that **the source still encoded the inverted one**, so a future caller arriving with `runtime_tx_auth_prepared == 0` would have reintroduced resolve-before-`set_delegation` silently.

The chosen fix states the routine's precondition instead of duplicating the writer: `dispatch_tx_runtime_code` **is** the `prepare_dispatch` side, so requiring `set_delegation` to have run is the spec's own contract at `interpreter.py:356-365` rather than a guard bolted onto it. And it removes the latent install, so the inverted path is **gone from the source** rather than merely unreachable.

The alternative — a second auth arm at dispatch entry — was rejected for a reason worth recording: it would have been **unreachable today**, so no sweep and no fixture could validate it. That is the #11407 shape, where an unexercised path concealed an FA-lax divergence that only a differential could expose; here there would not even have been a reachable row to difference against. **A fix that adds an unvalidatable path is worse than a fix that adds a precondition**, even when the added path looks more "complete".

And the incidental catch deserves the credit: adding `j .Ldtrc_ret` after `li a0, 7` prevents the success path falling through into the new failure arm — the same defect class that bit the #11378 family earlier tonight, caught while writing rather than while debugging.

pirapira:

PR #11465, opened 2026-08-04
refactor(codegen): retire the EIP-4788 nested-CALL fast path, execute the real bytecode

pirapira:
## What this retires

The EIP-4788 nested-CALL fast path (`ChildFrameHandlerTails.lean`): a `(if tag == "call_target" || tag == "staticcall_target" then …)` intercept that, for a nested CALL/STATICCALL to the beacon-roots contract (`0x000f3df6d732807ef1319fb7b8bb8522d0beac02`) with zero value and 32-byte input, skipped the interpreter entirely and emulated the contract from the modeled `swd_4788_*` descriptors with hand-computed gas.

Per execution-specs @ e5a8caf1b, `generic_call` (vm/instructions/system.py:330-385) routes **every** CALL-family child through `process_message` — no fast path, no system-contract special case (EIP-4788 is a *contract*, not a precompile).

**Survivor**: the real bytecode descent. The fallthrough to it already existed — the intercept's non-matching shapes took it — so this is a pure retirement with no hybrid and no new code to validate: the whole conditional block (conditions + both arms + labels) is deleted, and nested 4788 calls now descend normally. Gas falls out of execution instead of being hand-computed.

## Per-arm gas direction (emulated vs real execution at the pin)

| arm | emulated | real (contract-internal) | delta |
|---|---|---|---|
| success, both slots warm | 320 | ~315 | wash |
| success, one cold | 3220 | ~2215 | emulated +1005 |
| success, both cold | 6120 | ~4315 | emulated +1805 |
| stale | 3104 | ~2198 cold / ~198 warm | emulated +906 / +2906 |

Emulated ≥ real **everywhere** — no undercharge case exists. The guest overcharged on these paths today (stricter than spec → FRs); switching to execution lowers the charge to the spec-exact value (an FR fix on cold/stale shapes), never an undercharge. Mechanism: the emulated cold delta of 2900/slot is 900 over the true cold-SLOAD delta (Amsterdam: 2100 cold / 100 warm), and the stale arm's 3104 sits ~900 above the real revert path — its own label said STALE.

## Per-arm warm-set (the required side-effect check)

| arm | emulation seeded warm | real execution warms | post-cut |
|---|---|---|---|
| success | `{ts_idx, root_idx}` | `{ts_idx SLOAD, root_idx SLOAD}` | **identical** |
| stale | nothing | `{ts_idx SLOAD}` | ts_idx now warmed where emulation left it cold — a later same-tx access to ts_idx costs warm instead of cold, matching spec (FR-fix direction on later-access rows; no FA surface) |

## Interaction with #11446's identity check

The canonical-code_hash identity gate (#11446) guards the **block-start modeled write**; execution now governs the **nested-call path**. They are complementary: the gate prevents the modeled write from being applied against non-canonical code, and the nested path executes whatever the (identity-checked, canonical) code actually does. Neither retires the other. The block-start modeled write is untouched.

## Functional verification of the descend's storage view

The retired comment claimed the ordinary descend "cannot see the just-modeled system storage update". Measured false: the named nested-query family (`test_bal_4788_query`, 12 rows incl. current-timestamp success shapes) is **per-row identical** on both legs — the real-bytecode descend sees the seeded modeled write through the canonical storage map, and current-block success rows stay green.

## Measurement

Both legs at base cc465b233, canonical spike_run:

- Named family (12 rows): both legs FA=0, per-row identical (above).
- Random-200 (+40 family rows): both legs FA=0, FR=49, OK=180; **forward flips 0, reverse flips 0** — per-row identical. The overcharge was not verdict-fatal on any row in this sample; the change is structural (spec-shaped) and corpus-neutral on the measured sets, with gas accounting now spec-exact by execution rather than by emulation.

Candidate ELF emitted from this tree; pins regenerated (second pass no-op), drift guards green (region map + asm-to-program + Class-A ratchet with annotation sidecar 3/3).


pirapira:
**A pure retirement of 138 lines with the survivor already in place — and it establishes something stronger than the cut itself: the special case rested on a false premise.**

## The finding I did not expect

The retired block carried a rationale that the descend **cannot see the modeled write**. The 12-row `test_bal_4788_query` family — including the current-timestamp success shape — comes back **per-row identical on both legs**, because the descend *does* see the seeded modeled write via the canonical map.

**So the justification for the fast path was measurably false.** That reframes the change: it is not a trade of emulation for execution, it is the removal of a special case that never needed to exist. That is a better argument than the gas one, and it is the kind of claim only a measurement can make — reading the comment would have confirmed the opposite.

## On my own prediction, which was wrong

I said forward flips on the cold and stale shapes were "the point". **They came out zero** — r240 gives FA=0 FR=49 OK=180 on both legs, zero forward and zero reverse. So the overcharge, while real, was **not verdict-fatal on any sampled row**, and my framing of this as an FR fix was theoretical rather than demonstrated.

Stating that plainly rather than quietly reporting the zeros is the right call, and it does not weaken the PR: the structural bar is *structure matches the spec, random-200 not deteriorating*, and corpus-neutral measurement satisfies it exactly. The justification is the spec shape (`system.py:330-385` routes **every** CALL-family child through the interpreter; 4788 is a contract, not a precompile) plus gas becoming exact by construction rather than asserted. **A change can be right and invisible at the same time.**

## The warmth check came back cleaner than I feared

I asked for it because emulation seeding both slots warm would make post-cut accesses *colder*. The per-arm table answers it: success warms `{ts_idx, root_idx}` **identically on both paths**, and the **stale arm's emulation seeded nothing at all**, so real execution warming `ts_idx` is strictly more accurate. My concern pointed the wrong way — the emulation *under*-warmed on stale, so the cut is an improvement there too, not a regression.

## Also right

- **Gas direction table** per arm, with emulated ≥ real everywhere and **no undercharge case** — the screening question answered in the safe direction before the patch.
- **Identity-check interaction stated in the body**: the canonical `code_hash` gate guards the block-start modeled write, execution governs the nested-call path, complementary, neither retires the other. That sentence is what stops someone later retiring the identity gate by pointing at execution.
- The block-start modeled write left alone as system-transaction modelling — correctly out of scope.
- Ratchet sidecar 3/3, pins at a fixed point, drift guards green.

## One mechanical item

**Rebase needed.** Base is `88890ced1` but the measurement and the repin were taken at `cc465b233`, and the pin here (`0x591bc`) is against that older base — main's is `0x59f00`. This PR carries **no `merge!`**, so it is yours: rebase, repin, and confirm the new value differs from `0x591bc`.

**I am not asking for the r240 to be re-taken** on the new base. It is a pure deletion, both legs were measured on the same base with the base stated, and the structural bar does not require re-measuring on every main bump — re-measuring costs an hour and would answer a question nobody has.

pirapira:
**Rebase verified on `947a5b65c` / base `a7a34e8cb` — `merge!` applied now (it was not labelled before; see below).**

| check | result |
|---|---|
| `main` pin | `0x5a004` |
| this tip | `0x59634`, delta **`−0x9d0`** = 2512 bytes |
| stale-base check | **passes** — the pre-rebase value was `0x591bc`; getting it back would have proven a stale base |
| ratio sanity | 2512 bytes across 138 deleted source lines ≈ 18 bytes/line, consistent with multi-instruction Lean string lines |
| fixed point | no-op second pass, `sha256 -c` on all four artefacts |
| ratchet | OK, 13 baselined paths, no new edges |

Conflicts were confined to the three generated pin files, resolved by taking main's and regenerating from the green tree — the correct order, since a regeneration before the final rebase is what has to be thrown away.

**One correction to the handoff report:** this PR did **not** have `merge!` before now. Its labels were `area:codegen, size/M`, so the "merge! intact" line was a mis-statement rather than a self-label — worth flagging only because believing a PR is labelled would put it under the hands-off rule and leave it stalled with nobody acting. I have applied the label now that the fast gates are green.

## Why this change is worth having, restated

The headline is not the gas: **the retired block's own rationale was measurably false.** It claimed the descend cannot see the modeled write, and the 12-row `test_bal_4788_query` family — including the current-timestamp success shape — is **per-row identical on both legs**, because the descend does see it via the canonical map. So this removes a special case that never needed to exist, and that is a claim only a measurement could make; reading the comment would have concluded the opposite.

The spec side is unambiguous: `generic_call` (`system.py:330-385`) routes **every** CALL-family child through the interpreter, because 4788 is a contract rather than a precompile. Emulating it with asserted constants was a mechanism the spec does not have — and the constants had drifted, with 2900-per-cold-slot sitting a round **900** over the true cold SLOAD delta of 2000, and the stale arm's 3104 carrying its own "stale" label.

**And my prediction was wrong, which the report stated plainly:** I said forward flips on the cold and stale shapes were "the point," and they came out **zero** (FA=0 FR=49 OK=180 on both legs, no flips either direction). So the overcharge was real but not verdict-fatal on any sampled row. That does not weaken the case — the structural bar is *structure matches the spec, random-200 not deteriorating*, and corpus-neutral satisfies it exactly. A change can be right and invisible at the same time, and reporting the zeros rather than quietly presenting them is what makes the rest of the evidence trustworthy.

The warmth check also came back opposite to my concern: success warms the same set on both paths, and the stale arm's emulation seeded **nothing**, so real execution warming `ts_idx` is strictly more accurate.

Identity-check interaction is stated as required — the canonical `code_hash` gate guards the block-start modeled write, execution governs the nested-call path, complementary, neither retires the other.

PR #11467, opened 2026-08-04
codegen: remove stale runtime balance cells

pirapira:
Removes the two unreferenced runtime_current_bal_ptr and runtime_current_bal_len BSS cells from RuntimeSameBlockCode and regenerates the link-dependent address artifacts.\n\nEvidence and rationale:\n- On main 88890ced1, before editing, the linked stateless_guest ELF contained runtime_current_bal_ptr at 0xb9f58358 and runtime_current_bal_len at 0xb9f58360. After the edit and relink, neither symbol exists in the linked ELF, emitted assembly, or TSV.\n- RuntimeSameBlockCode.lean:108 is the negative guard proving the handoff itself was already gone; the source census found no readers or writers. The nearby comment said only kept, with no layout-stability claim, so these were a genuine leftover rather than a reserved slot.\n- The data payload shrinks by 16 bytes, and subsequent live BSS symbols shift down by 0x10. GuestAddrs.lean was regenerated from the fresh TSV, moving 52 referenced data constants by 0x10, including blsf_le_a from 0xb9f58890 to 0xb9f58880.\n- The linked image keeps the same .text base 0x80000000 and size 0x59f00, with no code layout or PC movement. The linker keeps the reported .bss extent fixed at 0x19295560 because of the dense-arena/backstop layout; only relocation immediates for shifted BSS addresses change.\n\nRegionMapLinkPins was explicitly checked before and after: textSizeBytes 0x59f00, dataSizeBytes 0x5370, bssSizeBytes 0x19295560, callFrameArenaBase 0xabf988a0, evmMemoryPoolBase 0xb23b18a0, and syslogBase 0xaa250100 are unchanged. The post-refresh checks pass.\n\nValidation: full non-cache lake build passes all 4812 jobs; codegen link check passes; check-region-map passes including TSV, GuestAddrs, and generated artifact consistency; check-asm-to-program is clean; git diff check passes.


pirapira:
**Blocking: 52 shifted symbols appear in `GuestAddrs.lean`, which this PR does not regenerate.** The deletion itself is correct and well-evidenced; the artefact set is incomplete.

## The measurement

Diffing symbol → address across the merge base and this tip:

| | |
|---|---|
| symbols **removed** | exactly `runtime_current_bal_ptr`, `runtime_current_bal_len` — the intended pair, nothing else |
| symbols **added** | none |
| symbols whose **address changed** | **263** |
| of those, present in `GuestAddrs.lean` on main | **52** (e.g. `blq_arith_params`, `blq_powt`, `blsf_le_a`, `blsf_le_b`, `blsf_le_d`, `blsf_le_zero`, `blsf_p1`, `blsf_p2`, …) |

So removing 16 bytes of `.bss` payload shifted 263 subsequent symbols, and **52 of them are declared in `GuestAddrs.lean`**. That file therefore no longer agrees with the TSV, which is exactly what the `GuestAddrs declared-start match` gate compares — and it is the gate that caught the same omission on #11452 earlier tonight.

**Please regenerate `GuestAddrs.lean`.** Also re-check `RegionMapLinkPins.lean`: your note about the fixed `.bss` extent from the dense-arena/backstop layout may well mean `bssSizeBytes` is unchanged, but state that explicitly with the before/after values rather than leaving it implied — an unchanged pin that was *verified* unchanged is evidence, while one that was never looked at is an assumption.

⚠️ And note some of those 52 are **precompile constant addresses** (`blsf_*`, `blq_*`), which are the same family whose `decide`-proved literals had to be repinned on #11446 when its `.data` moved by `0x40`. If any address-pinned proof references a shifted symbol, it needs repinning from the fresh ELF — and per the rule that bit that PR, the literals must be **allowed to move** and re-derived, never pinned back to make a proof pass.

## What is right, and the verification is the strong form

- **Before-and-after on the linked image**, not a source grep: both symbols confirmed *present* in main's ELF symbol table at `0xb9f58358`/`0xb9f58360` before the edit, and absent from the post-link ELF, the emitted assembly and the TSV afterwards. Verifying **presence before deleting** is what rules out removing something already gone and mis-attributing the change — that direction is usually skipped.
- **The removed set is exactly the intended pair with nothing else lost**, which is the check that distinguishes a targeted deletion from a regeneration that dropped something quietly.
- **`.text` unchanged with no PC movement** — correct for data-cell removal, and a `.text` change here would have meant more was removed than intended.
- The **negative guard** at `RuntimeSameBlockCode.lean:108` proving the handoff was already gone is what makes these cells *provably* unread rather than apparently unread.

## One thing worth writing into the body

The comment on these cells said only **"kept"**, with no layout-stability claim — and that is precisely why this is a leftover rather than a reserved slot. In this same session **#11458 deliberately kept a reserved BSS slot** *with a stated reason*. Same shape, opposite verdict, and the difference is entirely whether a reason was recorded. A future reader should be able to see both calls were deliberate, so please state it.

Also, a note on my own method here: my first attempt at this diff joined on the address column and produced nine "shifts" that were really alias pairs sharing an address. **The symbol column is the key; the address is the value.** I nearly reported that as a finding.

pirapira:
**The finding stands, and `source-checks` passing does not contradict it — it tells us the fast gate does not cover this case.**

Direct evidence:

```
GuestAddrs.lean:194   def blsf_le_a : Nat := 0xb9f58890     ← declared on main
symbol-addresses.tsv  blsf_le_a  0xb9f58890                 ← base
symbol-addresses.tsv  blsf_le_a  0xb9f58880                 ← this tip
```

A **16-byte shift**, exactly the removed payload. So `GuestAddrs.lean` now declares an address the TSV disagrees with, for at least 52 symbols.

**Why the fast gate missed it:** `check-declared-starts` validates **342 declared starts** — routine entry points — not data constants. `blsf_le_a` is a data address, so a shift confined to `.bss` data passes the ten-second gate and surfaces only in `build`. That refines something I said on #11458 earlier: I described the `GuestAddrs declared-start match` gate as the thing that catches partial regeneration, which is true for *routine* symbols and false for data constants. **The fast gate's coverage is 342 starts, not the 1025 symbols in the file.**

**What will actually fail, and it is worth predicting so it is not mistaken for something new:**

1. `check-region-map` in the `build` leg, comparing the declared map against the freshly linked ELF.
2. More pointedly — **address-pinned `decide` proofs on the precompile constants.** `blsf_le_a`, `blsf_p1`, `blq_powt` and their siblings are the same family whose literals had to be repinned on #11446 when its `.data` moved by `0x40`, and there the mismatch surfaced as `` `decide` proved pConstAddr.toNat = 0xa3000ee0 is false ``. Expect that shape here.

**So: regenerate `GuestAddrs.lean`, and let the precompile literals move** — re-derive them from the fresh ELF rather than pinning them back to satisfy a proof. That was the trap on #11446 and the reason its resolution was sound: the values were *derived*, and one of them coming back unchanged was then meaningful rather than assumed.

Everything in my previous comment about the deletion itself stands — the removed set is exactly the intended pair, nothing else was lost, the before-and-after was taken on the linked image, and `.text` did not move. This is an artefact-completeness gap, not a defect in the change.

pirapira:
**`merge!` applied.** Fast gates green (`summarize`, `path-labels`, `size-label`), and the `GuestAddrs` gap from my earlier comment is closed and verified.

The PR is currently CONFLICTING against `d26cd45a2` — that is **not** something to fix here. The conflicts are exactly the two generated files (`GuestAddrs.lean`, `symbol-addresses.tsv`) against a main that has moved twice since this branch was cut, while the source change is **two lines**. Rebasing would mean regenerating against a base that moves again within the cycle, which is the treadmill that cost four rounds elsewhere tonight. The label hands it to the merge actor, which now owns the conflict and the repin — and has demonstrated absorbing exactly this on four consecutive batches.

## What was verified

| check | result |
|---|---|
| removed symbols | exactly `runtime_current_bal_ptr`, `runtime_current_bal_len` — nothing else lost |
| `GuestAddrs.lean:194` | `blsf_le_a := 0xb9f58880`, matching the TSV — the 52 shifted data constants regenerated |
| `RegionMapLinkPins` | before/after **explicitly identical**, stated with values rather than implied |
| linked image | both symbols confirmed **present** at `0xb9f58358`/`0xb9f58360` before deletion, absent after |
| build | full non-cache `lake build`, 4812 jobs |

The before-and-after on the linked image is the part that makes this a deletion *argument* rather than a deletion hope, and verifying **presence before deleting** is the direction usually skipped — it rules out removing something already gone and mis-attributing the change.

## Correcting a prediction of mine that did not hold

I warned to expect address-pinned `decide` proofs on the precompile constants to fail, citing #11446 where it surfaced as `` `decide` proved pConstAddr.toNat = 0xa3000ee0 is false ``. **The full build passed.** My analogy was imperfect in a way worth recording: on #11446 the *proof file* carried a **hardcoded** literal, whereas these proofs read `GuestAddrs`, so regenerating `GuestAddrs` fixes them automatically.

So the real hazard is narrower than I stated — **proofs that hardcode an address instead of referencing `GuestAddrs`** — and this change was safer than my review implied.

## And the distinction this PR establishes, worth keeping

The comment on these cells said only **"kept"**, with no layout-stability claim, which is why they were a leftover. In this same session **#11458 deliberately kept a reserved BSS slot with a stated reason** — and that slot is now on main at `CreateCodeEffectLog.lean:788` (`runtime_mtx_active: .zero 8`), documented at `BlockVerdictMtxRuntime.lean:344` as *"a reserved compatibility cell only"*, with **zero writers and zero readers** confirmed.

Same shape, opposite verdict, and the difference is entirely whether a reason was recorded. That is the rule to apply next time an unread cell turns up: **an unread slot with a stated reason is reserved; an unread slot with "kept" is a leftover.**

pirapira:
Folded into #11469 so these land in one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11469 lands.

pirapira:

pirapira:
Landed via #11476. Verified: commit `ae05e8f4d` is an ancestor of `main` as of `ad5190b45` — closed rather than dropped.

Issue #11385, opened 2026-08-04
convergence: document that storage_root is derived (mpt_bounded_storage_root) and has no stored map field


pirapira:
**Attribution: coord.** Split out of #11286 (residual 3) as a concrete convergence item.

## The gap

There is **no discrete `storage_root` field in the map**. It is recomputed from storage slots via
`mpt_bounded_storage_root`, rather than loaded from BAL RLP or held in a cell. Nothing records that, so a
reader looking for "the stored storage root" searches for a cell that does not exist.

## Why it matters beyond documentation

This is the same failure mode that cost real time twice tonight:

- `code_state_find` / `upsert` / `commit_pending` / `pending_contains` existed as source strings with
  **no emitted symbols**, describing a CodeState table that is not emitted — they inflated #11329's
  structure count until an emitted-symbol census disproved them (#11383).
- The `BlockAccessListBuilder` layout table documented `balance_changes post` as **LE** when the stored row is
  **BE**, because the comment described the serializer's private scratch rather than the row. That single
  misplaced line produced four contradictory reports in one hour, two of them mine.

An absent-by-design field with no note reads identically to a field someone forgot to wire. The fix is to say
which it is, at the producer.

## What to record, and where

At the `mpt_bounded_storage_root` call site and in the map's layout documentation:

- `storage_root` is **derived**, not stored — computed from the `storage_writes` slots at each use
- therefore there is **no mask bit** for it and none is needed
- and per the spec this is correct: `forks/amsterdam` computes the account storage root from the trie, so a
  cached cell would be a guest-only structure with no counterpart

Follow the convention the layout table now uses after the endianness incident: **name the producer per
field**, so a reader can tell a derived value from a stored one without tracing.

## Bar

Documentation-only, so no measurement is required. If the change touches emitted comments inside asm strings,
confirm `.text` is unchanged.

Split from #11286 · related #10651, #11264, #11329

pirapira:
Already landed in `bf0000c8f` (ancestor of `origin/main` `8c0944216`); current source documents that `storage_root` is derived via `mpt_bounded_storage_root`. Found by `codex2`'s staleness sweep.

PR #11468, opened 2026-08-04
docs: record derived storage root

pirapira:
## Summary\n\n- document that storage_root is derived from storage_writes through mpt_bounded_storage_root\n- record that the storage map has no storage_root field or mask bit\n- distinguish the explicit mpt_state_root_ins path as the conservative bounded-builder failure fallback\n\nThe source comments do not change emitted assembly or layout.\n\nCloses #11385

### DIFF
--- a/EvmAsm/Codegen/Programs/BalAccountApplyPostFields.lean
+++ b/EvmAsm/Codegen/Programs/BalAccountApplyPostFields.lean
@@ -587,6 +587,10 @@ def mapAccountApplyPostFieldsFunction : String :=
   ".Lbaap_multi_skip_zero:\n" ++
   "  la t0, baap_sc_index; ld t0, 0(t0)\n" ++
   "  addi t0, t0, 1; la t1, baap_sc_index; sd t0, 0(t1); j .Lbaap_multi_loop\n" ++
+  -- #11385: `storage_root` is derived from the storage_writes rows at this
+  -- use via `mpt_bounded_storage_root`; it is not a map field or mask bit.
+  -- The explicit legacy call below is only the conservative builder-failure
+  -- fallback and does not introduce a stored storage-root representation.
   ".Lbaap_multi_apply:\n" ++
   "  la t0, baap_sc_out_count; ld a4, 0(t0); beqz a4, .Lbaap_nonce\n" ++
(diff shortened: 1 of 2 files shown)


pirapira:
Reviewed on base `d26cd45a2`. **Correct, layout-neutral, and it records the right thing. Will label once the fast gates report.**

## Layout neutrality verified, not assumed

The PR has **no generated files**, which is only correct if `.text` cannot move. I checked: **zero added lines contain an emitted-string fragment** — both comments are Lean-level (a `--` block above the call site and prose inside the module docstring), so nothing reaches the assembler. That check matters here because a comment placed *inside* a converted asm block becomes emitted bytes, and this codebase treats that as drift; a docs-only PR that silently shifted `.text` would need four generated files and would fail region-map instead.

## The claim is verified against the source, and the fallback distinction is the valuable part

`BalAccountApplyPostFields.lean:602` calls `mpt_bounded_storage_root` with the storage descriptors and writes `aps_newsroot`; `:620` passes that derived root to `account_set_storage_root`. `StorageWriteMap.lean`'s 96-byte row is address / slot / value with **no `storage_root` field or mask bit**. So the issue's premise holds: the root is **derived at the point the trie is rebuilt**, not stored.

⭐ And the comment does something better than restate that — it names `:616`'s explicit `mpt_state_root_ins` as **only the bounded-builder failure fallback**. Without that sentence, the next reader finds two root-producing call sites and reasonably infers a stored representation or a dual authority. Distinguishing the primary from the conservative fallback *in the source* is what stops that inference, and it is the same class of ambiguity that made `runtime_current_bal_ptr`'s bare "kept" comment read as intentional for two PRs.

## Why a docs-only PR is worth its cycle here

This one records a **negative** fact — *there is no stored map cell whose value could be read instead* — and negative facts are exactly what gets rediscovered. #11454's design independently reached the same conclusion tonight (storage_root out of resolver scope, derived via `mpt_bounded_storage_root`), which is two agents deriving one absence hours apart. Writing it at the site converts that into a read rather than a re-derivation.

The PR body closes #11385, which is right: the issue asked for exactly this to be documented, and the verification was done against source before writing rather than taken from the issue text.

pirapira:
Folded into #11469 so these land in one CI round. Auto-merge disabled here so it can't race the batch; this PR will close when #11469 lands.

pirapira:

PR #11469, opened 2026-08-04
chore: batch merge #11467 + #11468

pirapira:
Batches two `merge!` PRs into one CI round.

- **#11467** — `codegen: remove stale runtime balance cells` (`ae05e8f4d`)
- **#11468** — `docs: record derived storage root` (`bf0000c8f`)

## Verification

- `lake build` — green, 4812 jobs
- `gen-region-map-link-pins.py --check` — CLEAN, second pass a no-op
- all 20 gates `ok`; `check-bal-class-a-ratchet` — `OK (13 baselined paths; no new Class-A edges; no silent shrink; 3 sidecar bullet annotations present)`
- both constituents asserted reachable from the batch head; heads and `merge!` labels re-confirmed immediately before pushing

**#11465** (`retire the EIP-4788 nested-CALL fast path`) is separately in flight and not batched here — it was already armed and mid-CI when these arrived, and folding it in would have discarded that run.

Closes #11467
Closes #11468


pirapira:
Superseded by #11476. This batched #11467 + #11468, but all five ready PRs went behind when #11474 landed, so I rebuilt once with #11467 + #11468 + #11470 + #11471 + #11475 rather than re-syncing a batch plus three singletons. No work is lost — every constituent is in #11476, asserted reachable from its head.

Issue #11462, opened 2026-08-04
proof: rlp_content_to_u64's step bound is 7 * (2^64 - 1), derived from a counter's range rather than from the data


dhsorens:
Raised by @pirapira reviewing #11457 and reiterated on #11460. Filing so it is written down rather than rediscovered.

## The bound

```lean
cpsTripleWithin (7 * (2 ^ 64 - 1) + 18) (base + 16) …   -- Field0ToU64Top.lean:215
```

A `cpsTripleWithin` bound is an upper limit, so this is **sound** — the postcondition still holds if the routine terminates within it. The issue is what it *witnesses*. ~1.3 × 10²⁰ steps is derived from the **range of a 64-bit counter**, not from the input, so the theorem does not establish that the loop is bounded by its data. Anyone composing against it inherits that looseness, and `2^64` is exactly the value that makes a bound technically present but practically uninformative.

## Provenance — it is inherited, not chosen at any call site

| where | what |
|---|---|
| `EvmAsm/Rv64/RLP/Field0ToU64Top.lean:215` | **origin** — the `rlp_content_to_u64` scalar loop |
| `EvmAsm/Codegen/Programs/RlpFieldToU64FinishSAsm.lean:843, 965` | `tailSteps := (7 + (1 + (7 * (2 ^ 64 - 1) + 11))) + 5` |
| `EvmAsm/Codegen/Programs/RlpFieldToU64FlatSAsm.lean:74–75` | carried into `rlpFieldToU64_flat_spec_within` |
| `EvmAsm/Codegen/Programs/HeaderExtractNumberSpec.lean:392` | copied verbatim by the header wrapper |

So no wrapper can tighten it. The fix belongs at the origin; everything downstream then improves for free.

## Why a data-derived bound looks reachable

The loop walks the selected field's content, and on the success path that content is already constrained to **at most 8 bytes** — `Result.success` carries `hfit : len.toNat ≤ 8`, and `Result.tooLong` is precisely the arm that rejects anything wider. `hslack : listLen + 9 ≤ bytes.length` is in scope throughout, so `listLen` and `bytes.length` are both available as bounding quantities.

That suggests a bound in terms of the field length (or `listLen`) rather than `2^64`, which would make the operational content of the triple meaningful.

[...]

Issue #11461, opened 2026-08-04
proof hygiene: cpsTripleWithin step bounds carry a 2^64-derived factor in 32 places (one root, propagated) — record or tighten


pirapira:
## Summary

A family of `cpsTripleWithin` proofs on `main` carry a step bound of the form `7 * (2 ^ 64 - 1) + …` — roughly **1.3 × 10²⁰ steps**. These bounds are **sound** (a `Within` triple's bound is an upper limit, so the postcondition still holds), but the figure is derived from the **range of a 64-bit counter** rather than from the data, so the theorems do **not** establish that the loops are bounded by their inputs — even where the input length is in scope and already constrained.

Nothing here is wrong or unsound, and no proof needs reopening for correctness. The ask is that the looseness be **recorded or tightened**, because it is currently invisible and it propagates.

## Population, measured on `origin/main`

`git grep -o "7 \* (2 \^ 64 - 1)" origin/main -- 'EvmAsm/**/*.lean'` → **58 occurrences**; **32** are in step-bound position (a `*Steps :=` definition or a `cpsTripleWithin` fuel argument). Affected files, by count of step-bound occurrences:

| count | file |
|---|---|
| **16** | `EvmAsm/Rv64/RLP/Field0ToU64Top.lean` ← appears to be the root |
| 3 | `Programs/WithdrawalDecodeClose3.lean` |
| 3 | `Programs/WithdrawalDecodeClose2.lean` |
| 3 | `Programs/RlpFieldToU64SAsm.lean` |
| 2 | `Programs/RlpFieldToU64WholeSAsm.lean` |
| 2 | `Programs/RlpFieldToU64FinishSAsm.lean` |
| 1 | `Programs/WithdrawalDecodeClose5.lean` |
| 1 | `Programs/RlpFieldToU64FlatSAsm.lean` |
| 1 | `Programs/HeaderExtractNumberSpec.lean` |

The shape is **identical everywhere**, which is the useful part: `Field0ToU64Top.lean:215` et al. state `cpsTripleWithin (7 * (2 ^ 64 - 1) + 18)`, and the downstream files wrap it as `let tailSteps := (7 + (1 + (7 * (2 ^ 64 - 1) + 11))) + 5`. So this is **one bound propagated by composition**, not nine independent choices — which means it is likely fixable (or documentable) in one place.

[...]

PR #11470, opened 2026-08-04
proof(#11351): tie header_extract_number to _decode_header's number field

### LINKED ISSUE #11351: proof: triple for `header_extract_number` against the `number` field of `_decode_header`

dhsorens:
**Completes #11351.** The machine triple `header_extract_number_spec_within` has been on main since 27 July; what was missing was the tie to the model and the registry rows.

(I duplicated the triple in #11457 before finding it. Root cause recorded there: a truncated survey search, plus reading "no registry row" as "no proof". A missing row was never evidence of a missing proof.)

## The chain

| piece | why it was needed |
|---|---|
| `decode_header_inv` (SpecRef/Stateless.lean) | `rlpBytes?` is `private`, so `_decode_header` could not be inverted from outside the module **at all** — a caller could not even state what a successful decode gave it |
| `decodeAux_bytes_content` + threading through `strictNthItem_of_chain` / `success_content_of_decodeFully_list` | the forward stack pinned the payload's **length** but discarded its **content**, so a caller could tie a guest value only to a field of the right *size* |
| `result_value_of_success` | discharges all five `Result` constructors |
| `header_number_of_decode` | the consumer the Correspondence row cites |

On the content lemma: the payload occupies exactly the **last** `p.length` bytes of the item's span in all three byte-string forms — the single byte is its own content, the short and long forms place content immediately after their header. That uniformity is why one equation at `off' - p.length` covers all three. Each branch already derived it internally and threw it away.

`decode_header_inv` adds a **theorem only**, no definition changes, so port fidelity against execution-specs is untouched, and SpecRef stays Mathlib-free (`split` / `rename_i` / core `simp`).

## Why `.domainRestricted` and not `.agrees` — two independent reasons

**1. Arity.** The guest never checks how many fields the header has, so on a list of any other length it still returns a value where `_decode_header` errors. The honest statement is one-directional, not an iff. `number` is `getN 8` in **both** the 23-field (current fork) and 21-field (previous) arms, which is precisely why it is the family representative.

**2. Content.** The guest rejects a field wider than eight bytes (`Result.tooLong`) and one with a leading zero byte (`Result.noncanonical`). The port's `getN` is plain `bytesBEtoNat` and tolerates both.

⭐ So **the guest is stricter than the port here, and stricter matches CPython** — `rlp.decode_to` enforces exactly these. It is the *port* that dropped the checks, not the guest that invented them. That is in the clause table rather than hidden inside the verdict, since it is the one place this row's reference is weaker than the Python it ports.

**No byte-string side condition is needed**, which surprised me: `_decode_header` runs `items.mapM rlpBytes?`, and `rlpBytes?` sends `.list` to `none`, so a successful header decode already implies every field is a byte string. The #11434 scope restriction costs this family nothing.

## Also: the provenance comment promised on #11460

At `HeaderExtractNumberSpec.lean`'s `tailSteps`, recording that the `7 * (2^64 - 1)` factor is **inherited** from `Field0ToU64Top.lean:215` (the `rlp_content_to_u64` scalar loop, bounded by a 64-bit counter's range rather than by data) via `RlpFieldToU64FinishSAsm.lean:843,965` and `RlpFieldToU64FlatSAsm.lean:74-75` — no wrapper can tighten it, and anyone composing against the triple inherits the looseness. Tracked at the origin as **#11462**.

## Registry

`header` family 1 → 2 rows · `registry_size` 24 → 25 · `domainRestricted` 3 → 4 · `ported` 3 → 4 · `routineCount` 26 → 27 · proven 18 → 19. Both registries in this PR, with the `header_number_of_decode` witness abbrev in `Routines.lean` (it is Codegen-side, and Correspondence deliberately does not import Codegen).

## Gates

`lake build` warning-free · `check-axioms` (129 witnesses) · `check-registry-crosscheck` · `check-routine-liveness` · `check-correspondence-deps` · `check-progress` · forbidden-tactics / layering / file-size · `correspondence-check` self-test and `rlp` — all pass. `check-unimported`'s only orphan is the untracked `UnifiedDecodeItemLongBytesValidated.lean` (#9373 WIP, not mine).

Closes #11351


pirapira:
Reviewed at the head. **Kernel-clean, and the registry row is more honest than most. Two notes, neither blocking.**

## What earns this

**Zero `sorry`, `native_decide`, `bv_decide`** across the whole diff, and the `AxiomWitnesses` additions put both `header_extract_number_spec_within` and `header_number_of_decode` under `#print axioms` — so the claim is *witnessed by the axiom gate*, not merely asserted in a PR body. Adding the witness lines in the same change as the theorem is the right habit; a proof whose axiom footprint nobody checks is a proof nobody has to keep clean.

**The `decode_header_inv` diagnosis is the substantive part.** `rlpBytes?` being `private` meant `_decode_header` could not be inverted from outside the module *at all* — a caller could not even state what a successful decode yielded. That is a genuine expressiveness gap rather than a missing convenience lemma, and fixing it with a **theorem only, no definition changes**, keeps port fidelity against the spec untouched and SpecRef Mathlib-free. That constraint matters: a definitional change to SpecRef would have made the port itself the thing under review.

**And the content lemma's justification is the kind that generalises:** the payload occupies exactly the last `p.length` bytes of the item's span in *all three* byte-string forms, because the single byte is its own content and the short and long forms place content immediately after their header — so one equation at `off' - p.length` covers all three. Noting that **each branch already derived this internally and threw it away** is the useful observation; the forward stack was pinning length and discarding content, which is why callers could previously only tie a guest value to a field of the right *size*.

## The registry verdict is the right one, and for two independent reasons

`.domainRestricted` rather than `.agrees`, on **arity** (the guest never checks field count, so on a list of any other length it returns a value where `_decode_header` errors — the honest statement is one-directional, not an iff) *and* on content restrictions. Recording both qualifiers as **load-bearing** in the `note`, rather than picking the weaker verdict and leaving the reason implicit, is what makes the row auditable later.

⭐ And `number` as the family representative is justified rather than assumed: `getN 8` is identical in **both** the 23-field and 21-field arms, so the fork discriminant is a no-op for this field specifically. That is a real argument for representativeness — most "representative" claims in registries are assertions.

## Two notes

1. **Rebase needed** — base is `d26cd45a2`, main is now `2fc4ec791` (#11465's 4788 retirement landed). No `merge!` yet, so this is yours; `PROGRESS.md` is the likely conflict.
2. **The `#11457` root cause is worth keeping visible**, and I would rather it were in the file than only in a closed PR: *"a truncated survey search, plus reading 'no registry row' as 'no proof'. A missing row was never evidence of a missing proof."* That is the same failure mode that cost me four stale dispatches tonight — an index treated as ground truth. Since this PR *adds* the missing rows, a one-line comment at the new Correspondence entries saying the triple predated the row would stop the next person re-deriving it.

Separately, and not this PR's business: `HeaderExtractNumberSpec.lean` is one of the nine files carrying the `7 * (2 ^ 64 - 1)` step bound tracked in #11461. This PR composes against `header_extract_number_spec_within`, so it inherits that bound — which is sound but not data-derived. Nothing to do here; noting it so the inheritance is on the record.

dhsorens:
Both notes addressed, and one correction of my own.

**1. Rebased** onto `2fc4ec791`. `PROGRESS.md` was the conflict as you predicted; it is generated, so I took main's copy and re-ran `scripts/progress-report.sh --write` rather than hand-merging. Force-pushed; gates re-run green (axioms, progress, crosscheck, forbidden-tactics), `build` passing at the new head.

**2. The #11457 root cause is now in the file**, at the new Correspondence entry:

> ⚠️ THE TRIPLE PREDATED THIS ROW by over a week … only the correspondence was missing. I rebuilt it from scratch in #11457 before finding it, having read "no registry row" as "no proof" after a truncated survey search. AN ABSENT ROW IS NOT EVIDENCE OF AN ABSENT PROOF — grep the tree, unabridged, before rebuilding anything.

Agreed it belongs there rather than in a closed PR. The failure is specifically *treating an index as ground truth*, and the registry is exactly the index someone will consult next.

**3. ⭐ We filed the step-bound issue twice.** You opened **#11461** (32 sites, one root, propagated); I opened **#11462** a few minutes earlier, scoped only to the root. Yours is better scoped, so I have **closed #11462 as a duplicate** and moved my propagation trace into #11461 — origin at `Field0ToU64Top.lean:215`, through `RlpFieldToU64FinishSAsm.lean:843,965` and `RlpFieldToU64FlatSAsm.lean:74-75`, plus the note that `Result.success`'s `hfit : len ≤ 8` makes a data-derived bound *look* reachable, flagged unverified. The provenance comment in this PR now points at **#11461**, not #11462.

Noting your inheritance point is on the record: this PR composes against `header_extract_number_spec_within` and therefore inherits the loose bound. Sound, not data-derived, tracked at the root.

pirapira:
**Rebased onto `2fc4ec791` and green on the fast gates — `merge!` applied.** Zero `sorry`/`native_decide`/`bv_decide` re-confirmed on the rebased tip rather than carried over from my earlier read, since a conflict resolution is exactly where a tactic could slip in unnoticed.

Everything in my previous review stands. The three things worth restating as the reasons this lands:

1. **The axiom witnesses ship with the theorems.** Both `header_extract_number_spec_within` and `header_number_of_decode` are under `#print axioms` in `AxiomWitnesses.lean`, so the kernel-cleanliness claim is enforced by the gate rather than asserted in prose. A proof whose axiom footprint nobody checks is a proof nobody has to keep clean.
2. **`decode_header_inv` closes a real expressiveness gap, not a convenience one.** `rlpBytes?` being `private` meant `_decode_header` could not be inverted from outside its module at all — a caller could not even *state* what a successful decode yielded. Fixing it with a **theorem only, no definition changes** keeps the port itself out of scope, which is the right boundary: a definitional edit to SpecRef would have made port fidelity the thing under review instead of the bridge.
3. **The registry verdict is `.domainRestricted` with both qualifiers recorded as load-bearing** — arity (the guest never checks field count, so the honest statement is one-directional, not an iff) and the content restrictions. And `number` is *argued* to be the family representative rather than assumed: `getN 8` is identical in both the 23-field and 21-field arms, so the fork discriminant is a no-op for this field. Most "representative" claims in a registry are assertions; this one has a reason attached.

## The line from this PR I want on the record

> *"a missing row was never evidence of a missing proof."*

That is the root cause of #11457's duplicate, and it is the **same failure mode that cost me four stale dispatches tonight** — treating an index as ground truth instead of the artefact it indexes. My version was reading an open issue as open work; this one was reading an absent registry row as an absent proof. Both are the same error with different indices, and the corrective is identical: **check the thing, not the catalogue of things.**

Since this PR *adds* the missing rows, it converts one instance of that hazard into a read. Worth a one-line comment at the new Correspondence entries noting the triple predated the row, so the next surveyor does not re-derive it — but that is a nicety, not a condition.

Note for the record, not for this PR: `HeaderExtractNumberSpec.lean` carries the `7 * (2 ^ 64 - 1)` step bound tracked in #11461, so composing against `header_extract_number_spec_within` inherits a bound that is sound but not data-derived.

pirapira:

PR #11471, opened 2026-08-04
docs(codegen): mark the BLOCKHASH-via-storage primitives as probe-only

pirapira:
Eip2935.lean's header comment claimed production BLOCKHASH now consults the history-storage contract rather than walking witness.headers. Wrong: `eip2935_blockhash_lookup` (and its EIP-4788 sibling) are probe primitives, not on the production path; the production BLOCKHASH handler answers from the witness-header block-hash table built by `stage_blockhash_m29` (`evm_block_hashes`). The stale claim has already misled one structural-review map entry (third prose-outlives-code instance tonight).

Docs-only: both edits are Lean `/-- -/` comments, not emitted asm strings. Verified .text byte-identical before/after (same sha, same 368984 bytes); the only full-ELF difference is the embedded output filename. Also qualifies the Eip4788.lean sibling comment the same way.


pirapira:
**Correct, layout-neutral, and it repairs a comment that demonstrably cost analysis time. `merge!` applied.**

## Verified rather than assumed

- **Zero added lines contain an emitted-string fragment**, so both edits are Lean-level prose and `.text` cannot move — which is why the absence of generated files is correct here rather than an omission. A comment placed *inside* a converted asm block becomes emitted bytes, and a docs-only PR that silently shifted `.text` would need four artefacts and fail region-map.
- `source-checks`, `summarize`, `path-labels`, `size-label`, `velocity` all pass; only the ~1h `build` leg outstanding. MERGEABLE on current main `2fc4ec791`.

## The old comment was actively false, not merely vague

Before: *"The BLOCKHASH opcode **now consults that storage** rather than walking witness.headers."*

After: the storage path is named as a **probe primitive** (`eip2935_blockhash_lookup`, registered for probes only, not on the production path), and production BLOCKHASH is stated to answer from the witness-header table built by `stage_blockhash_m29` (`evm_block_hashes`). The `Eip4788.lean:51` sibling gets the same qualifier.

**This is worth more than a typo fix, because the false version already misled a structural review.** #11326's entry 10 was rated PARTIAL partly on the strength of that sentence — it implied a production dependency on 2935 storage that does not exist, which made the guest's BLOCKHASH look under-structured. Correcting it is what lets that entry close as structurally-different-but-equivalent.

## The pattern this is the fourth instance of tonight

Prose outliving what it described, each time costing real analysis:

| comment | reality |
|---|---|
| the 4788 fast path's *"the descend cannot see the modeled write"* | it can — the 12-row family is per-row identical, so the special case rested on a false premise (#11465) |
| `runtime_current_bal_ptr`'s bare *"kept"* | a leftover with no readers or writers, cut in #11467 after reading as intentional across two PRs |
| `single_tx_creation_runtime`'s name | already renamed to `block_verdict_creation_runtime`; I relayed the stale hazard twice |
| *"BLOCKHASH now consults that storage"* | probe-only — **this PR** |

⭐ The through-line: **a comment describing the production path is load-bearing, and when it goes stale it does not fail loudly — it silently redirects the next reader's model.** Three of those four were caught only because someone read the emitted image or the callers instead of the prose. So the useful habit is not "trust comments less" but **fix the comment the moment the code read contradicts it** — which is exactly what happened here, within minutes of the verdict that exposed it.

Also correct to leave the primitives themselves in place: they are legitimately registered probes, and the defect was the description, not their existence.

pirapira:

PR #11472, opened 2026-08-04
proof(#11345): invert decode_account_from_leaf, and record what it accepts

dhsorens:
Prerequisite for #11345, and the place to record a divergence that `AccountRecord.WF` actively hides.

`decode_account_from_leaf_inv` states a successful account-leaf decode in a form a caller can use without re-deriving the four-way pattern match: the full decode, plus each of the four fields as the reference computes it.

## ⚠️ I had the reference wrong, and the correction is the point of this PR

My plan for #11345 named `accountDecodedIs` / `AccountRecord` (`Stateless/State/AccountAssertions.lean`) as the reference side. That is wrong in the specific way #11341 exists to catch: `AccountRecord.WF` is **defined as the guest's own checks** — its docstring says "exactly `account_decode`'s checks":

```lean
def WF (a : AccountRecord) : Prop :=
  a.nonce < 2 ^ 64 ∧ a.balance < 2 ^ 256 ∧
  a.storageRoot.length = 32 ∧ a.codeHash.length = 32
```

Tying the machine triple to that would restate the guest against itself. The issue title always named the right target — `SpecRef.decode_account_from_leaf` (`witness_state.py:102`) — and this lemma is where the correction lands.

## The divergence, which is NOT #11351's pattern

| field | reference accepts | guest accepts | direction |
|---|---|---|---|
| nonce | **any** length (`bytesBEtoNat n`) | ≤ 8 bytes | guest stricter |
| balance | **any** length | ≤ 32 bytes | guest stricter |
| storage_root | **any** length; empty → `EMPTY_TRIE_ROOT` | **exactly** 32 | guest stricter, *incl. rejecting empty* |
| code_hash | **any** length; empty → `EMPTY_CODE_HASH` | **exactly** 32 | same |

The nonce/balance rows are the pattern #11351 established — guest stricter, matching CPython where the port had dropped a check.

**The root/hash rows are not.** The reference deliberately *folds* an empty field to a sentinel, and CPython does the same at `witness_state.py:102`. So on a leaf with an empty `storage_root` the reference succeeds with `EMPTY_TRIE_ROOT` while the guest fails outright — the guest is stricter than the **port and the Python together**. I predicted #11351's story would repeat here and said I would verify rather than assume; it does not repeat.

❓ Whether that is a real gap or only a grading nicety turns on whether such leaves are reachable in practice. A leaf with an empty storage_root is legal per the reference. **I have not measured reachability** — say the word and I will, before the #11345 row is written.

## Scope

Theorem only, no definition changes: port fidelity against execution-specs untouched, SpecRef stays Mathlib-free (`split` / `rename_i` / core `simp`). Gates: build warning-free, `port-check` PASS (25 theorems, classical axioms only).

Refs #11345, #11346


dhsorens:
⚠️ **Hold off reviewing this — it may be unnecessary, and the PR body's central claim is wrong.**

I have posted the full retraction on #11345. Short version: this PR says my plan named the wrong reference. It did not. The plan cited `AccountAssertions.lean:294`, which is `decode_account_from_leaf_accountRlp` — the correct `AccountRecord` → `SpecRef.decode_account_from_leaf` bridge, already witnessed at `AxiomWitnesses.lean:291`. I read the filename, assumed it meant `accountDecodedIs`, and never opened the line. Same failure as #11457.

Consequence: #11345's composition runs **record → decode** through that existing lemma, so inverting the reference decoder is **not needed for that row**. It would serve the converse (model → guest) direction, which is real but is not what #11345 asks for.

The divergence table in the docstring is still accurate and I would like it preserved somewhere — the guest being stricter than the port *and* CPython on empty storage_root is a genuine finding, and `a.WF` is where it hides.

I lean toward **closing this** and folding the divergence note into the eventual #11345 row's clause table, where it belongs. Say if you would rather keep the lemma for the model → guest direction and I will leave it open.

pirapira:
**This is stacked on #11470 with `base: main`, and that needs fixing before either merges. The new theorem itself is fine.**

## The stacking

`git merge-base --is-ancestor <#11470 tip> <#11472 tip>` is **true** — #11472 contains all of #11470, and its own contribution is a single commit (`proof(#11345): invert decode_account_from_leaf, and record what it accepts`, adding `SpecRef/WitnessState.lean` with `decode_account_from_leaf_inv` at `:139`, plus a Correspondence row).

With `base: main`, GitHub reports it MERGEABLE and **merging it would land all of #11470 under this PR's number**, leaving #11470 empty and auto-closed. That happened to #11424 earlier today, and #11388 came within a label of it.

Either fix is fine:

1. **Retarget this PR's base to #11470's branch** — then the diff shown is only the #11345 commit and GitHub enforces the order. Metadata only, no push, no CI restart. Preferred.
2. Leave `base: main` and state the dependency in the body so the order is explicit rather than inferred.

I have labelled #11470 already, so without one of these the merge actor can legitimately take them in either order.

## ⚠️ The new theorem has no axiom witness

`AxiomWitnesses.lean` gains exactly two lines, and both name **#11351's** header theorems — inherited from #11470:

```
#print axioms EvmAsm.Codegen.HeaderExtractNumberSpec.header_extract_number_spec_within
#print axioms EvmAsm.Codegen.HeaderExtractNumberSpec.header_number_of_decode
```

**Nothing witnesses `decode_account_from_leaf_inv`.** #11470 established the better habit — ship the witness with the theorem, so kernel-cleanliness is enforced by the gate rather than asserted in a body — and this PR inherits those lines without adding its own. Please add a `#print axioms` line for the new inversion theorem and for whatever Correspondence's new row cites.

That is not a formality here. The whole value of the axiom gate is that it is checked mechanically; a theorem outside it relies on someone remembering to look, and the count of witnessed proofs is itself a gate elsewhere in this repo.

## The theorem's shape looks right

`decode_account_from_leaf_inv` is the same move that made #11470 work: an **inversion** lemma letting a caller state what a successful decode yielded, where the definition's own shape otherwise prevents it. Zero `sorry`/`native_decide`/`bv_decide` across the diff, and `SpecRef/WitnessState.lean` gains a theorem rather than a definition — so port fidelity against `e5a8caf1b` is untouched, which is the right boundary for anything under `SpecRef`.

Title says *"and record what it accepts"*, which matches the #11470 precedent of writing the domain restriction into the registry rather than overclaiming `.agrees`. I will review the row's verdict once the base question is settled and the diff shows only this PR's work.

dhsorens:
**Both fixed. Thank you for catching the base — that was a live hazard, not a nit.**

**1. Base retargeted to `proof-11351-model-tie`** (your option 1, metadata only, no CI restart). The diff now shows only this PR's work. I created the branch off #11470's branch instead of `main` and did not notice; with `base: main` GitHub would happily have merged #11470's entire contents under this number and auto-closed it. Given #11424 hit exactly that today, the enforcement-by-base option is clearly right over stating the dependency in prose.

**2. Axiom witness added.** `_decode_account_from_leaf_inv_witness` in `Progress.lean`, beside the existing `_account_rlp_decode_witness`, and `AxiomWitnesses.lean` regenerated — 129 → **130 witnesses**, `check-axioms` green.

You are right that this is not a formality. I asserted kernel-cleanliness in the body while the gate could not see the theorem, which is precisely the gap the gate exists to close — and I had just been told the opposite habit was what made #11470 land. Adding the witness in the same change as the theorem is the rule; I broke it one PR later.

**One correction to your read:** the Correspondence row you saw is #11470's, not this PR's. Through the wrong base the diff was showing both. This PR adds **no registry row** — it is the inversion lemma only. So there is no verdict to review here; the row lands with the #11345 composition.

## ⚠️ And a question that may make this moot

I posted a retraction on #11345 shortly before your review, so it may have crossed:

My plan cited `AccountAssertions.lean:294` as the reference bridge. I "corrected" that to say it was wrong — and **my correction was wrong**. Line 294 is `decode_account_from_leaf_accountRlp`, the correct `AccountRecord` → `SpecRef.decode_account_from_leaf` bridge, already witnessed at `AxiomWitnesses.lean:291`. I read the filename and never opened the line.

Consequence: #11345's composition runs **record → decode** through `:294`. Inverting the reference decoder — what this PR adds — is **not needed for that row**. It would serve the converse, model → guest, which #11351 showed is worth having but is not what #11345 asks.

So: **keep or close?** I lean close, and fold the divergence note (guest stricter than the port *and* CPython on an empty `storage_root`) into the #11345 row's clause table where it belongs. But it is a correct, now-witnessed lemma and you may want it for the converse direction — your call, and I would rather ask than bin it.

pirapira:
**Both notes addressed, `build` is green, and this is now the cleanest stacked PR of the session. `merge!` applied.**

## The stacking is fixed the preferred way

Base retargeted to `proof-11351-model-tie@539768379`, so the diff is **only this PR's work** — 4 files, `Progress.lean` + `AxiomWitnesses.lean` + `SpecRef/WitnessState.lean` + `PROGRESS.md` — instead of carrying all of #11470. GitHub now enforces the order, which is a mechanical guarantee rather than a documented intention. That removes the empty-auto-close risk that hit #11424 today and nearly hit #11388.

## The axiom witnesses are in, and there are two rather than one

```
#print axioms EvmAsm.Stateless.SpecRef.decode_account_from_leaf_inv
#print axioms EvmAsm.Stateless.decode_account_from_leaf_accountRlp
```

Both new theorems are under the gate, not just the headline one. That is the habit worth keeping: **a theorem outside `#print axioms` relies on someone remembering to look**, and the count of witnessed proofs is itself a gate elsewhere in this repo, so an unwitnessed theorem quietly weakens a check nobody edited.

`build=pass` on top of the fast gates — the full leg, not just the ten-second ones. That is the first PR tonight I have labelled with the complete build already green rather than on the strength of `source-checks` plus a prediction.

## Why the inversion lemma is the right shape

`decode_account_from_leaf_inv` in `SpecRef/WitnessState.lean` is the same move that made #11470 work, applied to a different decoder: the definition's own form prevents a caller from stating what a successful decode yielded, so the caller cannot tie a guest value to a spec field at all. Adding a **theorem, not a definition**, keeps port fidelity against `e5a8caf1b` out of scope — which is the right boundary for anything under `SpecRef`, since a definitional edit there would make the port itself the thing under review rather than the bridge.

And the title's *"and record what it accepts"* matches the #11470 precedent: write the domain restriction into the registry rather than overclaiming `.agrees`. Two PRs in a row from this lane have chosen the honest verdict with the qualifiers stated as load-bearing, which is what makes the Correspondence table auditable instead of decorative.

**Merge order:** #11470 first, then this. The base pointer enforces it, so nothing further is needed from you.

pirapira:
Heads-up, no action needed yet — **this PR is about to be stranded, and I'll fix it.**

Its base is `proof-11351-model-tie` (#11470's branch), which is currently a constituent of batch **#11476** (mid-CI). When #11476 lands, #11470's content goes to `main` and this base branch becomes fully absorbed — the same state that stranded #11443 and #11388 earlier today, where the PR points at a branch that no longer represents a line of development and auto-merge would land it there instead of `main`.

I'll retarget this to `main` once #11476 lands, then verify and land it (locally: `lake build` + all 20 gates). Retargeting now would just make it `BEHIND` until #11470's content is actually in `main`, so waiting is cleaner.

If you'd rather rebase it yourself, that works too — just say so and I'll leave it alone.

pirapira:
Update on my earlier heads-up: **no retarget needed after all.** #11476 landed, and because `proof-11351-model-tie` was deleted on merge, GitHub auto-retargeted this PR to `main`.

That's the difference from #11388 and #11443, which I had to retarget by hand — their base branches still *existed* after being absorbed, so GitHub left them pointing at a dead branch. Here the branch was deleted, so the auto-retarget kicked in.

It's now `DIRTY` against `main`, which is just the usual post-merge conflict. Picking it up now.

PR #11473, opened 2026-08-04
fix(codegen): retire bal_txs_independent + deposit-capture route (#11183)

pirapira:
## Summary

Closes the Class-A / #10612 deposit-capture route under the maintainer proof-architecture ruling:

- Spec (`fork.py:913`) has **one** `process_transaction` loop; deposits come from post-exec receipt logs via `parse_deposit_requests`.
- There is **no** deposit-vs-capture-only branch and **no** `bal_txs_independent` counterpart.
- Every guest read of the supplied BAL **body** (beyond hashing the built BAL) is an extra check that needs a collision assumption — reject-only is still extra.

### Call

**(2) THE ROUTE SHOULD NOT EXIST** — delete, do not re-source.

### Changes

1. **MtxRuntime**: remove `jal bal_txs_independent`, BAL body loads, deposit-check arms, `deposit_capture_mark`. Zero retired cells at entry. Delete `.Lbv_mtx_capture_only_post_root` shortcut. After terminal root always publish runtime arrays → `blockVerdictMtxValidationTail` (ends `j .Lbv_after_tx_gas_precharge`; no fall-through into `.Lbv_mtx_creation`).
2. **ReceiptsTail**: drop `bnez bv_deposit_runtime_capture_complete`; empty logs → `DirectDepositFallback` (spec-shaped).
3. **DataSectionTail**: `bv_deposit_*` + `bti_*` cells RESERVED always-0 (layout-stable).
4. **V2 guest closure**: unlink `bal_txs_independent` + `bti_scan_*` (0 live jal). Probe units kept.
5. **TxsIndependent**: `bti_scan_storage_changes` string-only (drop `_prog`/GuestAddrs) so probe builds without guest text symbols; drop MANIFEST+fixture.
6. **Class-A**: remove baselined `bal_txs_independent` row + sidecar; `EXPECTED_ANNOTATION_COUNT=0`; ratchet allows deliberate zero.

### Layout

- `textSizeBytes` **0x59cc8**
- Dead symbols absent from ELF/TSV/nm: `bal_txs_independent`, `bti_scan_tuples`, `bti_scan_storage_changes`

### Gates

- `check-region-map` CLEAN
- `check-bal-class-a-ratchet` OK **12** paths
- `check-asm-to-program` CLEAN (382 defs / 144 files)
- GuestImage builds

### Rejection direction

Always-run validation tail **adds** checks vs capture-only shortcut — FR-safe (more rejection, not less). No rejection path lost for independence parse-bail (a0==2): that was an extra body-read check with no spec counterpart.

### Not in this PR

- Probe-only `bal_txs_independent` (412/415/418) stays
- Remaining Class-A execution reads (e.g. 11406 family if still open) untouched

No `merge!` from grok.


pirapira:
**This closes the Class-A programme's last supplied-BAL *execution-input* read, and it handles both traps I flagged. Reviewed in detail because the sidecar interaction is new and easy to get wrong.**

## The gate consistency is exactly right, including a subtlety I did not ask for

| check | result |
|---|---|
| baseline rows | 13 → **12**, with exactly `.Lbv_code_preimage_flag_done / bal_txs_independent` removed |
| sidecar keys | **0** |
| declared `annotation-count` | **0** |
| `EXPECTED_ANNOTATION_COUNT` | **0** |
| bullets | **0** |

All five agree, so the key cross-check has nothing to resolve and cannot fail as `missing`. That was the trap: the sidecar's cross-check fails when a note names a row that no longer exists, and grok's three settlement annotations were keyed to precisely the row being retired.

⭐ **And the zero-guard was narrowed rather than deleted**, which is the part I would have pushed back on:

```python
# Zero is legal only when EXPECTED_ANNOTATION_COUNT is deliberately 0
# (every annotated edge retired, e.g. #11183 bal_txs_independent).
if annotation_count == 0 and EXPECTED_ANNOTATION_COUNT != 0:
```

That preserves the guard's purpose. It exists because **both real annotation losses tonight went to exactly zero**, and a bare `== 0` check would now be permanently satisfied. Gating it on `EXPECTED != 0` keeps it armed the moment anyone adds an annotation back, while letting a deliberate empty file be honest. Deleting the check would have disarmed the only instrument that caught those losses; weakening it to a comment would have been worse. This is the right shape, and the rationale is in the source where the next reader will find it.

## The three surviving calls are probe-only, verified not assumed

`jal ra, bal_txs_independent` still appears 3× — all in `ziskBalTxsIndependentPrologue` (`BlockVerdictTxsIndependent.lean:369/372/375`), whose docstring names it `zisk_bal_txs_independent`: probe, with three fixtures (`bti_bal_indep` expect 0, `bti_bal_interact` expect 1, `bti_bal_sysread` expect 0). **Zero production callers**, which matches grok's own settlement annotation from #11183 that the sole guest `jal` was MtxRuntime — now removed.

Keeping the probe is correct: it exercises a routine that no longer runs in production, which is exactly what a self-test is for, and retiring the routine's *use* does not require retiring its *test*.

## What remains on the Class-A baseline, and why it is a different question

12 rows survive: `bal_find_account_by_address` ×5, `rlp_walk_init` ×2, `bgv_u64le` ×2, `bgv_u32le` ×1, `bal_map_builder_consistent` ×1, `bal_gas_valid_from_builder` ×1.

Under the maintainer's collision-freedom ruling these are **still extras** — the spec reads no supplied BAL body at all — so none is excused by being reject-only. But they are a distinct class from what this PR retired: `bal_txs_independent` was a **control** read selecting an execution route, and its removal was blocked on the route itself being unjustified. The remainder are witness-integrity comparisons, and each needs the same test applied individually: **does a spec behaviour correspond, and can it be enforced at the spec's own trigger point on the spec's own data** — the #11410/#11426 pattern. That is follow-up scope, not a gap in this PR.

## Mechanics

Rebase needed — base `2fc4ec791`, and `symbol-addresses.tsv` plus `GuestAddrs` are in the diff so the four-file set is present. `capture_complete` retained as **RESERVED always-0** rather than deleted: acceptable under tonight's rule *provided the reason is stated in the source* — an unread slot with a stated reason is reserved, an unread slot with bare "kept" is a leftover. Please confirm the comment states it.

Also note `btiScanStorageChangesFunction.s` is deleted with its MANIFEST row — good, a fixture whose subject is gone should not linger.

pirapira:
**Checked the reserved-slot condition myself rather than asking — it is satisfied.** `BlockVerdictDataSectionTail.lean:380-383`:

```
-- #11183 RESERVED always-0: deposit-capture-only route retired (no spec
-- counterpart). Slots kept so layout offsets stay stable; no live setter.
"bv_deposit_capture_only:\n  .zero 8\n" ++
"bv_deposit_runtime_capture_complete:\n  .zero 8\n" ++
```

That states all three things the distinction needs: **why** the slots are dead (route retired, no spec counterpart), **why they are kept anyway** (layout offset stability), and **that nothing writes them** (no live setter). So these are reserved, not leftovers — the same call #11458 made correctly, and the opposite of `runtime_current_bal_ptr`'s bare "kept", which read as intentional across two PRs before #11467 finally cut it.

Worth stating the rule explicitly since it has now been applied three times tonight with two different outcomes: **an unread slot with a stated reason is reserved; an unread slot whose comment only says "kept" is a leftover.** A future reader can tell these apart without re-deriving the history, which is the entire point.

## Remaining item is the rebase

Base is `2fc4ec791` and the PR is CONFLICTING; the four generated files are already in the diff, so this is the mechanical case — take main's generated files, regenerate, confirm a no-op second pass. The pin must not come back as its current value.

Two things to re-verify **after** the rebase rather than assume they carried, because both are exactly what a conflict resolution silently drops and both are the substance of this PR:

1. **The sidecar quintet still agrees**: baseline rows down by one with `bal_txs_independent` gone, sidecar keys 0, declared `annotation-count` 0, `EXPECTED_ANNOTATION_COUNT` 0, bullets 0.
2. **The narrowed zero-guard survives** — `if annotation_count == 0 and EXPECTED_ANNOTATION_COUNT != 0:` with its comment. That line is the only thing keeping the guard armed for a future annotation, and a merge that took main's version of the script would silently restore the bare `== 0` form and make the gate permanently vacuous.

Nothing else outstanding. This closes the Class-A programme's last supplied-BAL execution-input read, and the twelve remaining baseline rows are witness-integrity comparisons — a different class, each needing the #11410/#11426 treatment individually.

pirapira:
**Rebase verified independently on `9d1f1ec93` — `merge!` applied. This closes the Class-A programme's last supplied-BAL execution-input read.**

I re-derived both re-verification items rather than accepting the report, because both are what a conflict resolution silently drops and both are this PR's substance:

| check | result |
|---|---|
| baseline rows | **12**, `bal_txs_independent` present **0 times** |
| sidecar keys / declared count / bullets | **0 / 0 / 0** |
| `EXPECTED_ANNOTATION_COUNT` | **0** |
| **zero-guard at `:268`** | `if annotation_count == 0 and EXPECTED_ANNOTATION_COUNT != 0:` — **the narrowed form survived** |
| pin | main `0x59788` → tip `0x592f8`, **−0x490** (1168 bytes) |

The zero-guard is the one I most expected to be lost. Had the merge taken main's version of the ratchet script, the bare `== 0` check would have returned — and since the count is now legitimately zero, **that check could never fire again**. A permanently vacuous gate, invisible to every test, from a one-line merge resolution. It survived.

## What this closes, and what it does not

The spec **never reads a supplied BAL body** — `fork.py:928` builds its own, `:366`/`:390` compare hashes, `:933` validates the built one's gas. So every guest read of the supplied body is an extra check whose only excuse would be *"a mismatch here implies a collision"*, and collisions provably exist. Under that ruling `bal_txs_independent` could not be carved out; it had to go.

And the reason it was the *hardest* row turned out to be the reason it was **retirable**: it was a control read selecting the deposit-capture-only route, so the reject-only argument didn't cover it (#11402). But `fork.py:914` runs exactly `process_transaction` per transaction with no deposit mention, and deposits are classified **post-execution from receipt logs** (`:923` → `:957 parse_deposit_requests`). **The route has no spec counterpart, so the control dependency disappears with the control.** Retiring a branch beat re-sourcing its selector — and the authenticated tx-list predicate that fed it became irrelevant rather than being improved.

**Twelve baseline rows remain**, and they are a different class: witness-integrity comparisons (`bal_find_account_by_address` ×5, `rlp_walk_init` ×2, `bgv_u64le` ×2, `bgv_u32le`, `bal_map_builder_consistent`, `bal_gas_valid_from_builder`). Still extras under the ruling, so none is excused by rejecting only — but each needs the #11410/#11426 treatment individually: **is the property spec-real, and can it be enforced at the spec's own trigger point on the spec's own data?** That pattern has now worked twice, moving the code-preimage check from a supplied-BAL row scan to the dynamic code-read set.

The reserved slots are correctly documented (`BlockVerdictDataSectionTail.lean:380-383` states the route is retired with no spec counterpart, that offsets are kept stable, and that there is no live setter), the three surviving `jal`s are probe-only per their own docstring, and `btiScanStorageChangesFunction.s` went with its MANIFEST row.

pirapira:
## Merge-time note: the conflicts here are **generated files only** — and one line in this PR must not be resolved toward main

Checked against the queued batch #11476 (`#11467 + #11468 + #11470 + #11471 + #11475`), which **does not touch** `scripts/check-bal-class-a-ratchet.py`, `scripts/bal-class-a-baseline.tsv`, or `scripts/bal-class-a-notes.md`. So after that batch lands, this PR's conflicts against main are confined to:

```
EvmAsm/Codegen/GuestAddrs.lean
EvmAsm/Codegen/RegionMapLinkPins.lean
scripts/asm-fixtures/symbol-addresses.tsv
```

All three are regenerable — take main's, then regenerate to a fixed point. No semantic resolution is needed.

## ⛔ But do not let the ratchet script be resolved toward main

Main currently carries the **bare** guard, and this PR **narrows** it:

| | |
|---|---|
| main (`:266`) | `if annotation_count == 0:` |
| this PR (`:268`) | `if annotation_count == 0 and EXPECTED_ANNOTATION_COUNT != 0:` |

Main also has `EXPECTED_ANNOTATION_COUNT = 3`, baseline 13 rows with the `bal_txs_independent` row present, and 3 sidecar annotations. This PR takes all of those to **0 / 12 rows / 0**.

**If the resolution restores main's bare `== 0` form while keeping this PR's `EXPECTED = 0`, the gate becomes permanently vacuous** — the count is legitimately zero from here on, so a bare check can never fire again. Nothing would fail; the guard would simply stop existing while appearing to be present. That guard is the only instrument that caught the two annotation losses earlier today, and it must stay armed for whenever an annotation is added back.

**The five values that must agree after the merge**, all of which I verified on `9d1f1ec93` before labelling:

- baseline rows **12**, `bal_txs_independent` **absent**
- sidecar keys **0**, declared `annotation-count` **0**, bullets **0**
- `EXPECTED_ANNOTATION_COUNT` **0**
- the guard in its **narrowed** form with its comment
- pin `0x59788 → 0x592f8` (−0x490) recomputed against whatever main is at merge time

The first four are cheap to re-check with a grep and are the ones a file-level resolution can silently break. The pin will differ from `0x592f8` after rebasing onto a moved main — that is expected, and getting `0x592f8` back would mean the regeneration used a stale base.

pirapira:

pirapira:

pirapira:
Landed via #11489. Verified: commit `0765b83db` is an ancestor of `main` as of `de755bf93` — closed rather than dropped.

Issue #11256, opened 2026-08-03
r59nm scope, as-of 02c6fcd7 - which consumers still require the append-only storage effect log


pirapira:
# r59nm scope, as-of 02c6fcd7 - which consumers still require the append-only storage effect log

Measurement-only census of `origin/main` at `02c6fcd7c546541f76824faa334b8e795de7b7b0` (merge of #11249), taken without #11251.

## One-line answer

The append-only storage effect log is **not redundant overall**. #10651 retired it for exactly one consumer: terminal-root candidate and value enumeration, now sourced from the execution maps in `EvmAsm/Codegen/Programs/BlockVerdictStateRoot.lean:112-249`, linked at `BlockVerdict.lean:403`.

## Consumers that still require the effect log

1. **SLOAD/SSTORE frame-log semantics.** `Storage.lean:242-282` scans the persistent append log; map tiers are used by the resolver at `Storage.lean:381-390` and `Storage.lean:664`, while the handlers at approximately `Storage.lean:500` and `Storage.lean:652` still consume the log and append execution rows. The storage write map is therefore a cross-transaction resolution/promotion tier, not a replacement for the execution log.
2. **System-row capture.** `BlockVerdictSystemStorageCapture.lean:105-194` uses the side arenas and the execution log for actual system SSTORE rows, while modeled EIP-2935/EIP-4788 rows are recorded in the side log and block map. The map does not replace the capture/serialization rows.
3. **BAL forward and reverse validation.** `bal_storage_matches_exec_log` and `bal_storage_covers_exec_log` scan the live/captured execution log and captured arenas, rather than the mutable storage map. These consumers still need the append-log evidence.
4. **Tuple-sequence validation.** `ExecLogSlotTuples.lean:4-20` and `:37-100` reconstruct slot tuple sequences from the execution log plus `exec_log_txindex`; `BlockVerdictMtxTail.lean:234-242` invokes the sequence check. The storage map cannot preserve this append order and transaction association by itself.

## What the mutable-state path already covers

The `storage_writes` map is authoritative for cross-transaction latest-value resolution and for terminal-root candidate/value enumeration. Tx-to-block promotion updates the map in `StorageWriteMap.lean:278-332`. Read-set promotion is a separate set of evidence/serialization inputs in `ReadSetsPromote.lean:127-160`; the access and read logs remain required by the execution-spec shape (`block_access_lists.py`) and must not be retired with the effect log.

## #11251 follow-up

[...]

PR #11474, opened 2026-08-04
test: add rollback witness KATs

pirapira:
## Summary

Adds the two GH #11256 rollback witnesses without changing rollback implementation:

- under-restore: DELEGATECALL child reads slot 1, writes slot 0, then reverts; the write must disappear while reads survive.
- over-restore: the same child stops successfully; the write must persist while the independent read survives.

## Baseline

Exact-main codegen build completed successfully. Stock ziskemu 0.16.0 ran both fixtures with one job: 2 selected, 0 errors, 0 failures, 2 full output matches, and successful_validation byte matched 1 in both cases.

Guest SHA-256: 00225b05fd14a83a245bc0c8b79581f4e7911fa8462a6f23e233117844421f9a.

Reference fill and execution-spec input/output parity also passed for both cases.


pirapira:
**Fixture-only, baseline recorded, and the baseline result changes what cut 2 is. `merge!` applied once the fast gates report.**

## The separation is right

No mechanism changes, which is the condition I set: witnesses must land before the patch, or the only evidence a fix works comes from tests written by the same pass that wrote the fix. Fixtures plus a generator plus a README, nothing else.

## ⭐ The important result: both witnesses PASS on the current guest

> `selected: 2  errored: 0  fail: 0` — `full match: 2, root match: 2, successful_validation match: 2, tail match: 2`

**So there is no rollback defect.** The current mechanism already handles both directions correctly, *including read-survival* — `under_restore_child_reverts` requires slot 0 absent **while the parent BAL retains reads of slots 1 and 0**, and full-match means the BAL matched, so the reverted frame's read did survive.

That materially rescopes cut 2. It was authorised as FA-capable work needing constructed witnesses because under-restore would accept effects the spec discards. **The witnesses show the guest does not under-restore.** So what remains of cut 2 is **structural only** — the journal/cursor implementation versus the spec's snapshot/restore shape — with **no behavioural gain and no FA to close**. It still qualifies under the structural bar (structure matches spec, no deterioration), but it must not be described as a soundness fix, and it drops well below cut 1 and cut 3 in priority.

**This is the most valuable thing the witnesses could have told us**, and it is only visible because the baseline was taken *before* touching the mechanism. Had cut 2 been patched first, these fixtures would have passed afterwards and been reported as evidence the change was correct — when in fact they pass either way. That is precisely the trap the separation was meant to avoid, and it sprung.

## The design survived into the artefacts

Slot `1` is the read-only sentinel and slot `0` is the written slot, with the README recording *why*: **the BAL omits a read that is also represented in the storage-change list** (`block_access_lists.py:549-552`). Without a distinct sentinel, the only read would be absorbed into slot 0's `storage_change` and **read-survival would be unobservable** — so a witness using one slot could pass a mechanism that wrongly rolls back reads. `DELEGATECALL` keeps the child in the parent's storage context so both accesses land in the same account's entry.

Provenance is stated properly: JSON generated against `execution-specs` `e5a8caf1b`, reference fill passed both cases, and the byte-level converter reproduced both outputs under `--verify-input-parity --verify-execution-spec-input --verify-run-stateless-guest`.

## One note

The baseline was measured with **`ziskemu 0.16.0`**, whereas the fleet is on **spike for uniformity and comparability**. ziskemu is trustworthy — this is not a correctness objection, and for a 2-row KAT the performance argument is negligible. But the fleet's numbers are spike numbers, so please re-take this baseline on spike, or note in the README which backend produced it so a later reader does not compare it against spike figures as though they were the same measurement. The `guest SHA-256` being recorded is what makes that re-take cheap.

pirapira:
Verified and armed — `lake build` green (4812 jobs), all 20 gates `ok`, 0 behind `main`.

**One observation, not a blocker and not specific to this PR.** I tried to run the new KAT directly and couldn't:

```
$ python3 scripts/kat/test_rollback_witness_kat.py
ModuleNotFoundError: No module named 'pytest'
```

Then I went looking for what does run it, and found nothing: no `pytest` and no `scripts/kat` reference anywhere in `.github/`, no `check-*.sh` invoking pytest, no Makefile/lakefile runner, and `pytest` isn't listed as a dev dependency in `AGENTS.md`/`CLAUDE.md`/`docs/`.

That's **pre-existing and repo-wide** — `scripts/kat/` already holds ~10 `test_*_kat.py` files plus generators, and `fixtures/kat/` ~10 fixture dirs, none of them wired into the PR gate. So this PR follows the established convention and I'm not holding it for that.

Worth surfacing though: the 1061-line `rollback_witness.json` this adds is a pin, and nothing in CI currently checks that the guest still reproduces it. That's the same shape as the guard #11284 silently dropped — a check that exists but never runs. If these KATs are meant to be manual/offline, ignore this; if they're meant to gate, one workflow step running the `test_*_kat.py` set (plus `pytest` in the CI image) would make the whole existing corpus load-bearing rather than decorative.

PR #11475, opened 2026-08-04
fix(codegen): resolve storage values through transaction map

pirapira:
## Summary

This is Cut 1 of #11256: make storage-value lookup use the execution-specs
authority chain.  Cut 2 (rollback semantics) and Cut 3 (capture/metadata
retirement) are intentionally out of scope.

The resolver now checks, in order:

1. the current transaction's `storage_writes` map;
2. the block-cumulative `storage_writes` map;
3. authenticated header/pre-state storage by key;
4. zero for an absent account/slot.

The implementation reuses `storage_writes_block_latest_value` for the map
lookup.  `h_SLOAD` keeps the append-only execution-log scan only as a
duplicate-seed check; the map-derived pair is the value authority on a cold
miss.  `h_SSTORE` uses the same resolver while retaining its created-account
original-zero rule.

The original-value path is deliberately distinct: `get_storage_original` in
the pinned execution-specs tree starts with current-transaction-created
accounts as zero, then block-level writes, then authenticated pre-state.  It
never reads the current transaction's own write as the transaction-start
original.  This matters for same-slot writes: `test_sstore_xto_y.py`'s
`contract_2` performs repeated same-transaction pairs (`SSTORE(1,1)`,
`SSTORE(1,0)`, through slots 1..16, then `SSTORE(1,1)`) with empty pre-state.
The second write must retain original zero rather than the previous
transaction-local value.  The named 20-case run (d0..d9 × g0/g1) was 20/20
full, with zero errors/fails and byte-identical result/output files.

Spec reference: execution-specs `e5a8caf1b8055e4d805c7fb169edfa710914b7da`,
`execution-specs/src/ethereum/forks/frontier/state_tracker.py:170-231`:
`get_storage` consults transaction writes, then block writes, then pre-state;
`get_storage_original` intentionally omits transaction writes.  The same
ordering is present in the active fork modules.

## Rebased evidence

The branch is based on current `origin/main` `2fc4ec791ab03bb6633b925dcae939c03d953c42`
(which contains the requested `d26cd45a2`).  Parent and candidate artifacts:

- parent ELF: `e651921964475a145291d822aca27564f82d5515bfd5292ed746803972060777`
- candidate ELF: `b750b986976b5688b0a8b1aaa507bc40d6626a1d776a4bee1d53d7facae8a0af`
- `.text`: `0x59788` -> `0x598d0` (`+0x148` / 328 bytes)
- `.data`: `0x53b0` -> `0x53b0` (unchanged)
- `.bss`: `0x19295560` -> `0x19295560` (unchanged)

Both legs used the same random-200 manifest, seed 4242, Spike with 30 jobs.
Parent and candidate each selected 200, ran 185, errored 15, failed 4,
budget 0, matched fully on 181, matched roots on 185, matched success on
181, matched tails on 185, and had zero root-only diffs.  All 200 result TSVs
and all 204 output files were byte-identical.  Forward flips were 0 and
reverse flips were 0; both base-PASS -> candidate-FR and candidate-PASS ->
base-FR sets are empty.  The parent run used the parent ELF above and the
candidate run used the candidate ELF above; generated timestamps were
2026-08-04T22:29:07Z and 2026-08-04T22:30:47Z respectively.

The current-main layout was regenerated to fixpoint.  `check-region-map`,
`check-asm-to-program`, and `lake build EvmAsm` (4812 jobs) all pass.  The four
changed files are the resolver source and regenerated `GuestAddrs.lean`,
`RegionMapLinkPins.lean`, and `symbol-addresses.tsv`; `GuestImageEntries.lean`
was unchanged.

## Scope boundary

This PR does not alter rollback.  In particular, the execution-specs read set
survives frame rollback while write maps are restored; that is Cut 2 and needs
separate deep-revert witnesses.  Capture/metadata retirement is Cut 3 and is
also separate.


pirapira:
**Cut 1 verified and labelled.** Fast gates green, and the two things I asked for are both present.

| check | result |
|---|---|
| diff contents | **one commit** (`f8eaf501c`), 4 files — none of #11456's commits, which is what the rebase had to prove |
| pin | main `0x59788` → tip `0x598d0`, **+0x148 (+328)** matching the reported `.text` delta |
| `.data` / `.bss` | unchanged |
| generated set | `GuestAddrs`, `RegionMapLinkPins`, TSV alongside `Storage.lean` |

## The original-value tier is separated in the source, not just in the report

This was the one place cut 1 could have been silently wrong, and the diff answers it directly:

> *"supplies only `current`; `original` continues at the block-map/pre-state tiers, matching `get_storage_original` and never sourcing the original from [the tx map]"*

with `sstore_created_original_zero` kept as a **distinct arm**. That matters because `get_storage_original` (`state_tracker.py:305-329`) is `created_accounts → 0`, else **block** `storage_writes`, else pre-state — and sourcing the original from the tx map would make a slot written earlier in the same transaction price against its own new value. Writing the constraint into the comment at the site is what stops a later reader collapsing the two lookups "for tidiness".

## And the evidence has both halves, which is why this lands

- **Regression:** random-200 on the same parent, identical result rows and output/log hashes, zero flips either direction, base-PASS→cand-FR empty.
- **Discriminator:** `sstore_xto_y` 20/20 full, and — the part that makes it count — `contract_2` is repeated **same-transaction** writes to one slot (`SSTORE(1,1)`, `SSTORE(1,0)`, … then `SSTORE(1,1)` again) with **empty pre-state storage**. So the final write's gas depends on the original being the pre-state zero rather than the tx's own earlier value.

**Byte-identical alone would have proved only that the map agrees with the log wherever the corpus looks.** The named case is where they *could* have disagreed and didn't — that is the difference between "agrees" and "agrees where it was possible to be wrong", and it is what I would want restated if anyone later questions this migration.

Cut 2 and cut 3 exclusions are stated explicitly in the body, which keeps this PR's evidence from being borrowed by work it cannot speak to — the same separation that made cut 2's witnesses informative rather than self-certifying.

pirapira:

PR #11477, opened 2026-08-04
feat(codegen): resolve execution account code fields

pirapira:
## Summary

Implements #11454 Part Two without routing producers. `account_resolve_execution_state` now resolves the authenticated account code hash through the witness code table while preserving execution-state tier precedence.

- ABI extends the resolver with `witness.codes` (`a6/a7`) and output `code_ptr@40`, `code_len@48`, `present@56`.
- Transaction-map values take precedence over block-map values for code and state; tombstones remain terminal.
- Uses raw `witness_codes_lookup_by_hash`; deliberately does not call `code_read_fetch`, so this materialization path does not mutate `code_reads` or witness-code selection.
- Statuses are resolver-local: `0` absent, `1` live, `2` present-but-empty, `3` deleted, `4` resolver-unavailable (non-empty authenticated hash missing from `witness.codes`), `5` malformed authenticated lookup. Status 4 is witness incompleteness: a valid block may be rejected, so that rejection is a false reject (FR). Status 5 is malformed authenticated proof/input evidence: its rejection is a genuine reject. The distinction is part of the ABI; no producer routes through this resolver yet.
- EF0100 delegation designators are recognized by the `ef 01 00` prefix after a length guard and preserved for the existing follow-marker dispatch path; they are never executed as ordinary code.
- Storage roots remain out of scope and continue through `mpt_bounded_storage_root` (#11385).

The authenticated account/code distinction follows the pinned execution-specs at e5a8caf1b: execution reads use the tx-write/block-write/pre-state tier, while the BAL pre-state helper remains the separate two-tier `_get_pre_tx_account` contract.

## Layout / build evidence

Base is current `origin/main` `ad5190b45`. Source/generated candidate head is `0941bd69b`. Generated files were regenerated to fixpoint after rebasing: `GuestAddrs.lean`, `symbol-addresses.tsv`, and `RegionMapLinkPins.lean` (GuestImageEntries unchanged).

- Parent ELF: `8ecf557bb22357b7f4642ead5d1644f5c22541c505fe6248bc5f22c8c189df13`, linked 2026-08-05 01:14:55 +0200.
- Candidate ELF: `e8a90d3c2517e64affb9c9b3985b770f40d7e42c88f98bf2d30637d493e243b7`, linked 2026-08-05 01:18:30 +0200.
- `.text`: `0x598d0` -> `0x59b78` (`+0x2a8`); `.data`: `0x53b0` unchanged; `.bss`: `0x19295560` unchanged; `bnf_le_a=0xa3000f20` unchanged.
- `lake build`, `check-asm-to-program.sh`, `check-region-map.sh`, and `check-build-units-link.sh` pass.

## Behavioural evidence

Matched random-200 (`limit=200`, random seed=11454, Spike, 30 jobs, SpecRef oracle) on both current-base artifacts. Each leg selected 200 and ran 187, with 13 runner errors; each had 186 exact matches, one FR, and zero FA. The parent baseline was generated at `2026-08-04T23:25:09Z`; the candidate baseline at `2026-08-04T23:21:46Z`.

`eest-ab-compare.py` joined 199 rows by wrapped input digest (zero unmatched), found zero status differences and zero output-byte differences; FA/FR were `0/0` delta and the FR label set was identical. The candidate named rows `00048` (delegation on funded recipient), `00164` and `00199` (set-code-to-precompile) were all `OK` and byte-identical to the parent.

No merge label applied by the author.


pirapira:
**Rebased tip `0941bd69b` on `ad5190b45`, fast gates green, and the docstring condition is met — `merge!` applied.**

| check | result |
|---|---|
| pin | main `0x598d0` → tip `0x59b78`, **+0x2a8** |
| stale-base check | **passes** — the pre-rebase value was `0x59a30`; getting it back would have proven a stale base |
| MERGEABLE on current main | yes |

## The condition is satisfied, and it is stated better than I asked for

> *"Status 4 means a valid authenticated account lacks a `witness.codes` preimage: a block may be valid, so a caller's rejection is a **false reject (FR)** caused by witness incompleteness. A malformed authenticated lookup uses 5: that is malformed proof/input evidence, so its rejection is a **genuine reject** rather than a witness-shortfall bail. **Keeping 4 and 5 separate is therefore part of the ABI even though no producer routes through this helper in this cut.**"*

That last sentence is the one I would not have thought to require, and it is the load-bearing one. It tells a future implementer that the distinction is **deliberate and unexercised** — so the natural instinct on finding two codes that both mean fail-closed, which is to fold them, is pre-empted with a reason rather than merely forbidden. An unexercised distinction with no recorded intent gets collapsed by the first consumer; one that says "this is part of the ABI, and nothing uses it yet" does not.

## The rest of the ABI documentation earns its place too

The docstring records **why** the raw `witness_codes_lookup_by_hash` was chosen over `code_read_fetch` — *"this path materialises state and must not record a code read or alter witness-code selection"*. That reason matters more than the rule: the two routines are one tail-jump apart, the wrong choice is invisible at the call site, and it would silently pollute the very `code_reads` set that decides which code the witness must carry. A reader who knows only "use the raw one" will eventually see a reason to switch; a reader who knows why will not.

Also recorded: map code rows are authoritative with pointer/length preserved, account absence and `EMPTY_CODE_HASH` are truthful zero-length results rather than fabrications, and `EF0100` markers are recognised **by prefix** — not by length 23, which is the #11406 lesson (recognition is a property of the code bytes, not of which store they came from, and not of a length that a 23-byte non-marker could coincidentally match).

## Scope held

**No producers routed**, which is correct and deliberate: part three's own finding is that routing requires three-way — now four-way — discrimination at consumers that do not yet have it, and five of the twelve currently ignore the return entirely. Landing the resolver before its consumers is the right order precisely because the consumers need restructuring, not adapters.

pirapira:

pirapira:

PR #11478, opened 2026-08-04
refactor(codegen): retire unread storage capture metadata

pirapira:
## Spec alignment

At execution-specs `e5a8caf1b8055e4d805c7fb169edfa710914b7da`, [BAL `AccountChanges`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/block_access_lists.py#L179-L189) defines storage writes separately from slots read but not changed, and [Amsterdam state transition](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/fork.py#L357-L367) derives the state root and BAL hash from the block outputs.

Guest change: retire the unread user-storage capture metadata and five production capture calls; retain `bv_system_storage_log`/`bv_system_storage_log_count` as live modeled-system staging ([source](https://github.com/Verified-zkEVM/evm-asm/blob/d1f10d39a30a6b24b7082ba802c28748a0dcd884/EvmAsm/Codegen/Programs/BlockVerdictDataSection.lean#L739-L748)). The four validator symbols were already absent from the baseline linked image.

Ask: please review/label this cut; it contains no production validation removal and no merge commit.

## Evidence

- Base used for final pins: `c8ad1069f`; branch head: `d1f10d39a`.
- Baseline/candidate ELF SHA: `8ecf557bb22357b7f4642ead5d1644f5c22541c505fe6248bc5f22c8c189df13` / `a695eaa3ded3519c7b96ece185e8eb5cea9ce88fbb05d29947e11c39042577b5`.
- Pin delta: `.text -1112` bytes (`0x598d0 -> 0x59478`), `.data 0`, `.bss -2,375,744` bytes (`0x19295560 -> 0x19051520`).
- A/B was measured on `ad5190b45` and repeated after the `c8ad1069f` pin regeneration: Spike `seed=4242`, `limit=200`, `jobs=8`; both `200/200`, `15` errored, `4` failed, `185` ran; `198` joined, `0` status differences, `0` output-byte differences, byte-identical.
- Verification: `LAKE_ARTIFACT_CACHE=false lake build` (`4813` jobs), `check-no-warnings`, `check-region-map`, and `check-build-units-link` all pass; structural gates pass.


pirapira:
**Both conditions I set are met, and the arena answer is better than the one I offered. Reviewing now; will label once the fast gates report.**

## The arena question is resolved in a third way I had not listed

I offered three options — removed, reserved-with-a-stated-reason, or retained by accident — and asked for the choice to be stated. The actual answer is **none of those: `bv_system_storage_log` is repurposed as live modeled-system staging**, with the role change recorded in the source at `BlockVerdictDataSection.lean:746`:

> `-- Keep the modeled-system staging arena standalone from call_frame_arena.`

alongside a new `bv_system_storage_map_seed_only` cell. So the arena is still **written**, by the modeled-system seeding path rather than by capture. That makes it neither a leftover nor a reserved-dead region — it is in use for a different purpose, which is a stronger disposition than "reserved with a reason" because nothing has to be taken on trust about future intent.

The `bv_system_storage_log` definitions surviving in `AccountTupleSequencesConsistent.lean` and `BalAllAccountsStorage.lean` are those unlinked modules' own probe-local copies, not production.

## The zeroing went with the captures

`BlockVerdictStateRoot`'s `bv_system_storage_log_count; sd zero` reset is **gone** — zero occurrences on this tip. That was the #11426 orphan shape: a deleted call leaving its setup behind, where the ratchet then misattributed the leftovers to the next routine. Deleting the call and its setup together is what avoids it.

## Scope confirmed as all five, with the callers named

Mtx plus four StateRoot calls — **withdrawal, consolidation, builder-deposit, builder-exit**. Naming which four rather than citing addresses matters here, because my original cut-3 scope came from a single callsite report and was wrong; a scope stated as roles is checkable by someone who has not disassembled the image.

## Layout deltas are large and consistent with the change

| | main | tip |
|---|---|---|
| `.text` | `0x598d0` | `0x59330` (**−0x5a0**) |
| `.data` | `0x53b0` | `0x53b0` (unchanged) |
| `.bss` | `0x19295560` | `0x19051520` (**−0x244040**, ≈2.37 MB) |

The `.bss` shrink is the retired user arena and capture globals, and `RegionMap.lean` moving (13+/18−) is expected since the region map documents the arena placement that just changed. A `.data` unchanged alongside a large `.bss` drop is the right signature for removing arenas rather than code constants.

## Still needed

**Rebase** — base is `ad5190b45` and the PR is CONFLICTING. Given the size (TSV 1888+/1902−, `RegionMap`, four pin values), regenerate after the rebase and confirm a no-op second pass; the new `.text` and `.bss` values must differ from `0x59330`/`0x19051520` if main has moved.

And please state in the body that **the four validator symbols were already unlinked before this cut** (source-retained from #11415's conservative split, with production's seam being `bal_map_builder_consistent`). A reader seeing four validators disappear from `nm` will otherwise conclude production validation was removed — and that inference is exactly the kind that outlives the PR.

pirapira:
**Rebased onto `388fbee2b`, MERGEABLE, fast gates green — `merge!` applied.**

| check | result |
|---|---|
| deltas | `.text` −1112, `.data` 0, `.bss` −2,375,744 — **unchanged across two base bumps**, which is the invariant that matters: the deltas are the change, so a shift in them would mean something else moved |
| A/B | 198 joined, 0 status/output diffs, byte-identical; provenance stated as first measured on `ad5190b45`, pins regenerated on `c8ad1069f`, reproduced fresh on `c8ad` |
| `nm` post-cut | capture helper, four validator symbols, `exec_log_seed_flag`, `exec_log_txindex`, user arena all absent — present before, absent after |
| system arena | **live modeled-system staging**, reason recorded at `BlockVerdictDataSection.lean:746` |
| `:892` zeroing | removed with the captures |

Scope is the five capture call sites named by role — Mtx plus withdrawal, consolidation, builder-deposit, builder-exit — which is checkable without disassembling the image, and matters because my original scope came from a single-callsite report and was wrong.

The body records that the four validator symbols were **already unlinked** before this cut, so their disappearance from `nm` is the arenas' readers going rather than production validation being removed.

pirapira:
Folded into #11488, a cross-author batch so these land in one CI round. Merges are `--no-ff`, so per-commit authorship is preserved. Auto-merge disabled here so it can't race the batch; this PR will close when #11488 lands.

pirapira:

PR #11479, opened 2026-08-04
fix(codegen): retire BAL field-compare cluster (#11183 ORDER-1)

pirapira:
## Summary

#11183 ORDER-1: retire the five Class-A **field-compare** edges that read supplied BAL body content with no execution-specs counterpart.

| retired edge | was |
|---|---|
| `.Lbmvmx_sd_preok` / `.Lbmvmx_cb_preok` | BAL post-balance vs exec-derived sender debit / coinbase credit |
| `.Lbv_b1_loop` | BAL final nonce vs AccountState latest |
| `.Lbv_b23_chk` | BAL post balance vs B2 running balance |

### EQUIVALENCE argument (not "hash covers it")

Spec pin `e5a8caf1b` amsterdam `fork.py:390`:

```python
if computed_block_access_list_hash != block.header.block_access_list_hash:
    raise InvalidBlock
```

on a list the spec **built itself** at `:928`. The spec has **no supplied BAL body** and performs **no content comparison**. A guest that still compares supplied fields against exec-derived expected **rejects blocks the spec accepts** precisely when a hash collision exists — that is an FR that **breaks equivalence**.

"The hash already covers it" is a **soundness** claim needing collision-freedom (ruled out). "The spec checks only the hash" is an **equivalence** claim needing only the pin. This PR uses the second.

### Kept

- B2.2 running-balance **underflow** (execution model, not BAL field compare)
- Single-tx legacy `bvReceiptsShapeSet 1 true` (envelope, not BAL)
- BIND rows 3–5 (locate BAL slice for hash/serialize) — programme endpoint
- Fail sinks 57/59 removed (no live jal after cut)

### Class-A

- Baseline **13 → 9** (deliberate shrink of the four field-compare rows; bmvmx was 2 edges)
- Ratchet docstring + baseline header: BIND residual is the **finish line**, not unfinished work; ratchet conflates BIND with CHECK because it tracks `bv_bal_start`/`len` refs

### Layout

- `textSizeBytes` `0x59208`
- region-map CLEAN; class-a OK 9; GuestImage builds

### Not in this PR (coord order)

- row 10 dtrc parse-bail
- rows 11–12 bbcv residual
- row 9 shadow_supplied_len diagnostic
- row 8 map_builder_consistent (own PR, equivalence written out)

## Test plan

- [x] `check-region-map` CLEAN
- [x] `check-bal-class-a-ratchet` OK 9
- [x] GuestImage builds
- [ ] CI build.yml

Closes nothing alone; advances #11183 / #10612 under collision-freedom ruling.


pirapira:
**Order-1 verified — `merge!` applied.** Baseline 13 → 9, and the four removed rows are exactly the field-compare cluster:

```
.Lbmvmx_cb_preok   bal_find_account_by_address
.Lbmvmx_sd_preok   bal_find_account_by_address
.Lbv_b1_loop       bal_find_account_by_address
.Lbv_b23_chk       bal_find_account_by_address
```

⭐ **And the endpoint note is emitted BY the generator, not hand-added to its output.** It sits inside the header-writing block at `check-bal-class-a-ratchet.py:198-199`, so `--write-baseline` **reproduces** it. That is the durable fix for the exact failure that cost this repo two annotation losses today: grok's three settlement lines were hand-added to the generated file and were destroyed twice, first by a regeneration and then by a batch merge, before #11451 had to build a whole sidecar-plus-count-gate to protect them. **Prose the generator emits cannot be lost by regenerating.** That is a better mechanism than the one we built to work around the problem, and it should be the default for anything explanatory that belongs next to generated rows.

## The endpoint framing is now in the artefact rather than in a review thread

> *"Ratchet conflates BIND with CHECK (tracks start/len refs). BIND residual is the finish line — not unfinished work. Retire CHECKs via EQUIVALENCE (spec only hashes the BUILT list at `fork.py:390`; no supplied body), never via 'hash covers it' (needs collision-freedom; ruled out)."*

Both halves matter. The first stops a future reader treating a BIND row as incomplete work and hunting for a way to remove something that must stay. The second stops the *wrong argument* being reused — and that argument was reached once tonight in this very lane before being corrected, so recording the prohibition next to the rows is worth more than recording it in an issue.

## One count correction, carried forward so it does not propagate

The triage summary said the field-compare cluster was **5** edges; the actual removal is **4** rows (13 → 9), matching rows 1–2 and 6–7 exactly. Nothing is missing — the "5" was an arithmetic slip in the summary line, not an unretired edge. Worth stating because "5 removed" would later read as one row having been quietly left behind.

`source-checks`, `path-labels`, `size-label`, `velocity` pass; `build` outstanding. `.text 0x59208`, region-map clean, Class-A OK at 9 paths.

## Remaining, in the agreed order

Row 10 (`.Ldtrc_deferred_marker_ready` parse-bail) → rows 11–12 (`bbcv_bal_contains_addr` residual, whose real property moved to the dynamic code-read set in #11410) → row 9's shadow-length diagnostic, leaving the builder-side gas path alone → row 8 (`bal_map_builder_consistent`) last in its own PR. After those, only the three BIND rows should remain, which is the finish line rather than a residual.

pirapira:

pirapira:

PR #11480, opened 2026-08-04
proof(#11345): primitives and the big-endian accumulator bridge

dhsorens:
Three lemmas toward #11345's four content lemmas — the step that turns `outputSuccess`'s cells (`AccountDecodeSpec.lean:157`) into `accountDecodedIs`'s record fields (`AccountAssertions.lean:135`). Prerequisite-only, same shape as #11408 / #11425 / #11460.

Branched clean from `main`; no dependency on #11472.

| lemma | why it was needed |
|---|---|
| `copyIntoRegion_eq_slice` | a full-width copy at a nonzero **source** offset is the corresponding slice |
| `toNat_shift_or` | shifting the accumulator and or-ing a byte in at the bottom is multiply-and-add |
| `beAccum_eq_fromBytesBE` | the accumulator is exactly `Nat.fromBytesBE` of the slice it walks |

## Two things a survey turned up that I would otherwise have hit mid-proof

**`copyIntoRegion_self` only covers `srcOff = 0`.** `fixed32Copied bytes oldOut o` copies from `o.toNat`, so lemmas 3 and 4 of #11345 need the general form. Its `srcOff + n ≤ src.length` hypothesis is load-bearing rather than defensive: `copyIntoRegion` pads out-of-range reads with `0` via `getD` while `List.take` returns a shorter list, so without it the two sides differ in **length**, not merely contents.

**⚠️ `toNat_or_shift` does not apply, despite looking like it should.** I wrote it for `bgv_u32le` in #11379 and expected to reuse it. It shifts the **byte** — `x ||| (z <<< m)`, little-endian — whereas `beAccum` shifts the **accumulator** — `(acc <<< 8) ||| z`, big-endian. Mirror image, same proof shape (bitwise disjointness, then `BitVec.add_eq_or_of_and_eq_zero`), different statement. Reading the statement before reaching for it saved an attempt.

## ⭐ The bound the induction needs is the guest's own check

`beAccum_eq_fromBytesBE` requires `n ≤ 8`, which keeps every intermediate accumulator under `2 ^ 56` so the shift never drops a high bit.

That is exactly `AccountRecord.WF`'s nonce clause. It is the third time in this row's work that `a.WF` has turned out to *be* the thing the proof needs rather than an obstacle to route around — `Decoded` (`AccountDecodeSpec.lean:144`) is likewise four `Success` facts plus `l0 ≤ 8`, `l1 ≤ 32`, `l2 = 32`, `l3 = 32`, i.e. `WF` spelled out. Worth stating because it is what makes `.domainRestricted` gated on `a.WF` the honest verdict rather than a convenient one.

## Where this is going

Remaining for #11345: the four content lemmas (two now near-immediate from `copyIntoRegion_eq_slice`, one from `beAccum_eq_fromBytesBE`, and the balance right-alignment mirroring `copyN_rightAlign_eq` from #11374), then pinning the offsets via `success_content_of_decodeFully_list` + `success_deterministic`, then `adSuccessOut` → `accountDecodedIs`, after which `decode_account_from_leaf_accountRlp` (`AccountAssertions.lean:294`, already witnessed) closes to `SpecRef.decode_account_from_leaf`.

No registry row here — that lands with the composition, `.domainRestricted` gated on `a.WF`, with the clause table recording what `a.WF` excludes (the reference *and* CPython fold an empty `storage_root` to `EMPTY_TRIE_ROOT` at `witness_state.py:102`, while the guest requires exactly 32 bytes).

## Gates

`lake build` warning-free · `port-check` **PASS** (17 theorems, classical axioms only) · `check-axioms`, forbidden-tactics, layering, file-size **PASS**. `check-unimported`'s only orphan is the untracked `UnifiedDecodeItemLongBytesValidated.lean` (#9373 WIP, not mine); the new module is registered in `Codegen/Programs/Imports.lean`.

Refs #11345


pirapira:
**Kernel-clean, prerequisite-only, and the write-up is doing something more valuable than the lemmas. `merge!` applied.**

Zero `sorry`/`native_decide`/`bv_decide`; `source-checks`, `summarize`, `path-labels`, `size-label`, `velocity` pass with only `build` outstanding; branched clean from `main` with no #11472 dependency. Three lemmas, one new module, registered in `Imports.lean`. No registry row, correctly — the row lands with the composition, not with its parts.

## The two survey findings are the reason this PR is worth more than its diff

**`copyIntoRegion_self` only covers `srcOff = 0`**, while `fixed32Copied bytes oldOut o` copies from `o.toNat` — so the general form was genuinely needed rather than convenient. And the `srcOff + n ≤ src.length` hypothesis is **load-bearing, not defensive**: `copyIntoRegion` pads out-of-range reads with `0` via `getD` while `List.take` returns a shorter list, so without it the two sides differ in **length**. That is exactly the kind of hypothesis that gets dropped as noise by a later reader; recording *why* it cannot be is what keeps it.

⭐ **`toNat_or_shift` does not apply despite looking like it should.** It shifts the **byte** (`x ||| (z <<< m)`, little-endian); `beAccum` shifts the **accumulator** (`(acc <<< 8) ||| z`, big-endian). Mirror image, same proof shape, different statement. **Reading the statement before reaching for the lemma saved an attempt** — and that is the same failure mode that cost me four stale dispatches tonight: a name that matches the intent while the statement does not. A near-miss reused is worse than a lemma not found, because it fails late and looks like it should work.

## The bound being the guest's own check is the structural result

`beAccum_eq_fromBytesBE` needs `n ≤ 8` to keep every intermediate accumulator under `2 ^ 56` so the shift never drops a high bit — **and that is precisely `AccountRecord.WF`'s nonce clause.** With `Decoded` (`AccountDecodeSpec.lean:144`) likewise being four `Success` facts plus `l0 ≤ 8`, `l1 ≤ 32`, `l2 = 32`, `l3 = 32` — i.e. `WF` spelled out.

That is the third time in this row's work that `a.WF` has turned out to **be** what the proof needs rather than an obstacle to route around, and it is what makes `.domainRestricted` gated on `a.WF` **the honest verdict rather than a convenient one**. A domain restriction that coincides with the well-formedness the proof independently requires is a different claim from one chosen to make a theorem go through, and stating which it is belongs in the registry note when the composition lands.

## One thing to carry to that composition

The clause table should record what `a.WF` **excludes**, and the example given is the right one: the reference *and* CPython fold an empty `storage_root` to `EMPTY_TRIE_ROOT` (`witness_state.py:102`) while the guest requires exactly 32 bytes. That is a real domain gap, not a formalism — it means a witness the spec accepts is one the guest rejects, which is an **FR** and should be counted as such rather than described as a restriction.

PR #11485, opened 2026-08-05
fix(codegen): fail closed on CREATE resolver failure

pirapira:
## Summary

This is Part Three (row 7) of #11454, stacked on #11477 (`codex/11454-code-field`). It routes the CREATE code-deposit producer through the execution-state resolver and makes resolver failure explicit. The resolver is the execution-time three-tier read described by `state_tracker.py:179-203` at pinned execution-specs `e5a8caf1b`: transaction writes, block-cumulative writes, then authenticated pre-state.

The producer entry point is `create_record_code_effect`, which reaches `account_state_record_code` in `CreateCodeEffectLog.lean:852`. The resolver ABI is passed the CREATE address, authenticated parent header, witness state, and witness code table. It returns a resolver-local status:

- `0..3`: absent, live code, present-but-empty, or deleted;
- `4`: witness-incomplete (a non-empty authenticated code hash has no witness preimage);
- `5`: malformed authenticated lookup.

Statuses 4 and 5 are latched before either the code-effect row or the AccountState code row is published. CREATE's existing failure edge handles the nonzero producer return, so no partial or fabricated-empty row can escape.

Status 4 is reported at the receipts emitter as verdict code **67**, documented here as witness incompleteness / false reject (not a soundness rejection). Status 5 follows the existing genuine-invalid CREATE channel (code 62). This keeps a valid block whose witness is incomplete distinct from malformed authenticated input.

## Why this shape

The resolver must be fallible: a missing witness code preimage must not be silently converted to empty code. The CREATE caller therefore fails closed, rather than publishing an AccountState whose code fields were never resolved. Existing execution-state data is copied into the code producer's AccountState baseline only when the target has no pending or durable row; standalone `zisk_create_code_effect` generation uses the old ABI and remains an independent probe.

The change is stacked on #11477 and does not modify that branch. No `merge!` label is requested on this PR.

## Evidence

All generated artifacts were regenerated to fixpoint from the candidate tree.

- Candidate repository base: `ad5190b45` (current main at branch creation), with #11477 as the stacked base; candidate source head: `0941bd69b5030f3c25129ce938f83ed3644e96b8` plus this PR.
- Candidate ELF: `a3ea367854af19c854ab18e89ee639f39cfe3ea8ebb8c6ed32326f96f99434e9` (built 2026-08-05 01:55:32 +0200); emitted source SHA256: `3a6f0595acf941d8dea60449307c1b2c13701b42de5f076cb1a6ea9f1b068247`.
- Parent ELF: `8ecf557bb22357b7f4642ead5d1644f5c22541c505fe6248bc5f22c8c189df13`.
- Pins after regeneration: `.text 0x59cf0`, `.data 0x53b0`, `.bss 0x192955a0`; `callFrameArenaBase 0xabf988a0`, `evmMemoryPoolBase 0xb23b18a0`, and `syslogBase 0xaa250100` are unchanged. `.data` and all arena bases are invariant; only the emitted text/data-layout records moved as expected.

Random-200 A/B (seed 11454, Spike, 30 processes, same manifest):

- Parent and candidate each scored 187/200 (13 runner errors, budget 0), with `FA=0`, `FR=1` on both legs; the one FR is the same pre-existing row (`00141`).
- Joined 199 rows; FR label sets are identical, with zero new false accepts and zero reverse transitions.
- Status differences: 0. Output-byte differences: 0.

Named CREATE subset (`state_gas_create`, 50 rows, same manifest and backend): both legs scored 50/50 exact, `FA=0`, `FR=0`; the A/B join was 50/50 with zero status or output differences. Thus no previously passing row became failing, and no prior failure became passing in either the random or named sets.

Converse screen: no row-7 path currently rejects and now succeeds; the random-200 and named A/B runs had zero base-FR→candidate-PASS transitions.

Checks: `lake build`; `check-asm-to-program`; `check-region-map`; `check-unimported`; `check-layering`; `check-guestaddrs-starts`; `check-progress`; `check-drift`; and `check-axioms` all pass.

The standalone code-effect probe emits successfully with its historical ABI (`--asm-only`). Its old full standalone link still has the pre-existing guest-only undefined helpers (`account_state_delete`, `account_write_record`, `account_state_pending`); the production stateless guest links successfully and is the artifact used for all measurements above.


pirapira:
**`build` green and the substance is right — holding `merge!` on one missing line, which is the screen I asked for explicitly.**

## What is verified

Code **67** is documented at the emitter, not only in the body — `BlockVerdictReceiptsTail.lean:172`: *"67 means witness incompleteness, not invalid execution"*, and `:24` records it as a valid-block witness shortfall (FR). That is what keeps the FR accounting intact when someone later tallies rejections by code; a bare `li t0, 67` would have become indistinguishable from a soundness reject within a week.

Statuses 4 and 5 are latched **before** either the code-effect row or the AccountState code row publishes, so no partial or fabricated-empty row can escape. Status 5 uses the existing genuine-invalid CREATE channel (62). Stacked correctly on `codex/11454-code-field` with the base retargeted, and pin `0x59b78 → 0x59cf0`.

## The missing line

I asked for the **converse** screen stated explicitly rather than left inferred: *is there any path where row 7 currently **rejects** and would now **succeed**?* The body establishes the forward direction thoroughly — resolver failure now fails closed where it previously published — but does not state the reverse.

The reverse is the FA direction, and it is not obviously empty: this change also *copies existing execution-state data into the code producer's AccountState baseline when the target has no pending or durable row.* That is a new success path for a case that previously had none, so it deserves one line saying whether any input that reaches it currently produces a rejection.

I expect the answer is none — new failure arms plus a baseline copy that only fires where no row existed. But that is a guess about someone else's diff, which is exactly what I have been wrong about repeatedly tonight. **One sentence and I label.**

pirapira:
**Converse screen present — `merge!` applied.** `build` green, MERGEABLE at `40a281331`.

> *"no row-7 path currently rejects and now succeeds; the random-200 and named A/B runs had zero base-FR→candidate-PASS transitions."*

Both halves of that matter: the **reasoned** claim about the paths, and the **measured** base-FR→candidate-PASS set being empty. Either alone would be weaker — a reasoned claim can miss a path, and an empty flip set on a corpus that never exercises the path proves nothing. Together they close the FA direction.

This completes #11454's three parts: the whole-account resolver with the tx-map tier (#11477), the code-field resolution with the six-state ABI, and now row 7 as the **only** producer needing it — the other eleven staying field-specific, which is what reduced part three from twelve sites to one.

pirapira:
Queued behind #11489 — no action needed from you.

This PR is based on `codex/11454-code-field` (#11477's branch), which is a constituent of batch **#11489** (mid-CI). When that lands, the base branch is deleted on merge and GitHub auto-retargets this PR to `main` — that's what happened with #11472 earlier today, so no manual retarget should be needed.

I'll pick it up on the next pass once #11489 lands: sync to `main`, then full local verification (`lake build` + all 20 gates) before it reaches CI.

pirapira:
Folded into #11491, a cross-author batch so these land in one CI round. As predicted, GitHub auto-retargeted this to `main` when #11489 landed and deleted its base branch — no manual retarget was needed. Auto-merge disabled here so it can't race the batch; this PR will close when #11491 lands.

pirapira:
Landed via #11491. Verified: commit `40a281331` is an ancestor of `main` as of `5ecafe659` — closed rather than dropped.

PR #11486, opened 2026-08-05
refactor(codegen): parametrize CallFrameBase over GuestLayout (#10753)

pirapira:
## Summary
#10753 one-module conversion: `CallFrameBase` → leaf `CallFrameBaseProg` + bridge.

- `frameBase_prog_of (L : GuestLayout)`; bridge `frameBase_prog := _of guestLayout`
- New fields: `call_frame_arena`, `frame_base`
- Leaf imports `GuestLayout` only (no `GuestAddrs`)

## Selection (tree facts)
- Imports `GuestAddrs` ✓
- Emitted body lines > 0 (6-instr `frameBase_prog`, ASM `la` reloc) ✓

## Byte-identity (base `origin/main` @ emit time vs head)
| artefact | sha256 |
|---|---|
| `stateless_guest.s` | `4819410a4a1d1dab734123ebe0fd9062966f501452cfc3243c7e3f00bb7f72c9` (identical) |
| stripped ELF (`riscv64-unknown-elf-objcopy --strip-all`) | `70568e23e36a2b2498018b81588db628c56c87eba587eb55a8ce6fc991ab77e3` (identical) |
| `.text` | `bf178c43d3f6f5b36a301b38cc801645fec2273f541c18cb8f1ada4a60ab032c` (identical) |

Command: `lake exe codegen --program stateless_guest --halt linux93 -o …`

## Gates
- `lake build` CallFrameBase + CallFrameBaseSAsm OK
- `asm_to_program.py check-file --funcs frameBaseFunction` CLEAN

Closes nothing; continues #10753.


pirapira:
**First conversion of the lane — verified and labelled.** Fast gates green.

| check | result |
|---|---|
| emitted `.s` sha256 | **identical** (`4819410a…`) |
| stripped ELF sha256 | **identical** (`70568e23…`) |
| `.text` sha256 | **identical** (`bf178c43…`) |
| leaf imports `GuestAddrs` | **0** — accesses layout only via `L.call_frame_arena`, `L.frame_base` |
| bridge | `frameBase_prog : Program := frameBase_prog_of guestLayout` — name and type preserved for the concrete-render gate |
| `GuestAddrs` importers in `Programs/` | 133 → **132** |
| `GuestLayout` fields | 15 → 17 |

Byte-identity in all three forms is the proof the abstraction did not leak, and it is the whole verification this lane needs — cheaper and stronger than a sweep.

## One thing worth stating for the modules that follow

The bridge now imports `GuestLayoutInstance` rather than `GuestAddrs`, so the generated dependency is **one level removed rather than gone**: `CallFrameBase → GuestLayoutInstance → GuestAddrs`. The bridge still rebuilds on a layout change.

That is the intended architecture, and the win is real but should be described precisely: **the 59-line abstract leaf carrying the program and its relocs is now layout-independent, while only the thin bridge rebuilds.** Multiplied across 132 remaining modules, that is what turns a layout move from "invalidate every program module" into "re-apply a layout at the assembly boundary". Worth being explicit so nobody later reports the bridge's surviving transitive dependency as an incomplete conversion.

Issue #11484, opened 2026-08-05
coordination: relaxing Decoded for empty storage_root/code_hash (#11483) touches #11345's in-flight content lemmas


pirapira:
@dhsorens — coordination request, not a defect report against your work. **#11483's fix relaxes `Decoded`, which your in-flight #11345 content lemmas are built on**, so I would rather sequence it with you than have two changes land in the same predicate independently.

## The change

`Decoded` (`AccountDecodeSpec.lean:144-153`) currently requires **exactly 32 bytes** for fields 2 and 3:

```lean
l2.toNat = 32 ∧   -- storage_root
l3.toNat = 32     -- code_hash
```

The spec folds a **zero-length** field to the empty constants instead — [`witness_state.py:118-119`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/witness_state.py#L118-L119):

```python
storage_root = Root(decoded[2]) if decoded[2] else EMPTY_TRIE_ROOT
code_hash = Hash32(decoded[3]) if decoded[3] else EMPTY_CODE_HASH
```

So the guest rejects a leaf the spec accepts — an **FR**. The planned fix mirrors the spec exactly: admit `l2.toNat = 32 ∨ l2.toNat = 0` (and likewise `l3`), with the zero arm yielding `EMPTY_TRIE_ROOT` / `EMPTY_CODE_HASH`. Note the guest **already** matches the spec on the numeric fields, since `l0 ≤ 8` and `l1 ≤ 32` admit zero length ([`:114-115`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/witness_state.py#L114-L115) fold those too) — only the two hash fields diverge.

## Why it is a live FR, since my first assessment was wrong

I initially guessed this was unreachable: a canonical encoder writes 32 bytes, because an empty-storage account's `storage_root` is `keccak256(RLP(""))` — itself 32 bytes, not a zero-length field. That argument does not hold, for two reasons k3 established:

1. **The verifier hashes the leaf's RLP bytes as provided and never re-encodes**, so the committed root is over whatever encoding the witness carries. Canonical form is not consensus-enforced on the verification side. My argument constrained *producers* and then applied the constraint to a *verifier*.
[...]

Issue #11483, opened 2026-08-04
guest rejects empty storage_root/code_hash leaf fields that the spec folds to EMPTY_TRIE_ROOT/EMPTY_CODE_HASH (FR; reachability open)


pirapira:
## Summary

The spec **folds an empty RLP field** for `storage_root` and `code_hash` to the empty constants; the guest **requires exactly 32 bytes** for both. So a trie leaf carrying a zero-length field for either is **accepted by the spec and rejected by the guest** — a false reject.

Filing this separately from the proof work that surfaced it (#11480 / #11345), because it is a guest-versus-spec strictness divergence rather than a formalisation detail, and because the honest severity assessment below belongs on the record rather than in a PR thread.

## The divergence, both sides cited

**Spec** — `witness_state.py:118-119`, inside `_decode_account_from_leaf`:

```python
nonce        = Uint(int.from_bytes(decoded[0], "big")) if decoded[0] else Uint(0)
balance      = U256(int.from_bytes(decoded[1], "big")) if decoded[1] else U256(0)
storage_root = Root(decoded[2]) if decoded[2] else EMPTY_TRIE_ROOT
code_hash    = Hash32(decoded[3]) if decoded[3] else EMPTY_CODE_HASH
```

All four fields fold an empty value to a default. **Guest** — `AccountDecodeSpec.lean:144-153`, `Decoded`:

| field | spec on empty | guest requirement |
|---|---|---|
| `nonce` (0) | → `Uint(0)` | `l0.toNat ≤ 8` — **empty permitted** |
| `balance` (1) | → `U256(0)` | `l1.toNat ≤ 32` — **empty permitted** |
| `storage_root` (2) | → `EMPTY_TRIE_ROOT` | `l2.toNat = 32` — ⛔ **empty rejected** |
| `code_hash` (3) | → `EMPTY_CODE_HASH` | `l3.toNat = 32` — ⛔ **empty rejected** |

So the guest already matches the spec's leniency on the two numeric fields (`≤` admits zero length) and diverges on exactly the two 32-byte hash fields. Note this is **both** hash fields — #11480's note mentions `storage_root`; the same line folds `code_hash` identically.

## Failure direction and severity — low reachability, but state it rather than assume it

**Direction: FR.** The guest rejects a leaf the spec accepts. It cannot cause a false accept, so it does not jump the queue under the FA rule.

[...]

PR #11487, opened 2026-08-05
proof(#11345): join account_decode's output struct to AccountRecord and SpecRef

dhsorens:
Stacked on #11481 (base `proof-11345-of-content`) — it consumes that PR's three `*_of_content` lemmas. Retarget to `main` once #11481 lands.

## What this closes

`account_decode`'s triple leaves `outputSuccess`: four cells stated as **byte copies at offsets the guest computed**. `accountDecodedIs` states the same four cells as **`AccountRecord` field values**. Nothing connected them, so the machine proof and the record-level assertion could not talk to each other.

`outputSuccess_eq_accountDecodedIs` proves they are equal. `decoded_matches_specRef` pairs that with `decode_account_from_leaf_accountRlp`, giving the correspondence statement: the guest's four output cells and `SpecRef.decode_account_from_leaf`'s four results are the same fields of the same record.

That second theorem exists because `decode_account_from_leaf_accountRlp` has been sitting unconsumed. By the standard this tree already applies — *availability is not use* — an unconsumed lemma earns no basis grade.

## How the offsets get pinned

The offsets are opaque naturals the guest computed; the fields are model values. Bridging them:

1. `decodeFully_accountRlp` — the record's encoding decodes to its children (`decodeFully_encode`, bound from the existing `accountRlp_length_le`).
2. `success_content_of_decodeFully_list` — the model produces a `Success` **with content** at each index.
3. `success_deterministic` — the guest's `Success` at an index is unique, so the guest's `(o, l)` *is* the model's.
4. `AccountDecodeBridge`'s three `*_of_content` lemmas do the per-cell arithmetic.

## One fix worth flagging

`success_content_of_decodeFully_list` now also returns `offset.toNat + p.length ≤ bytes.length`.

I first tried to derive that bound from the content equation and it is **not derivable**: `(bytes.drop o).take p.length = p` constrains nothing about `o` when `p = []`, and an empty field is reachable here — `Nat.toBytesBE 0 = []`, so a zero nonce or zero balance hits it exactly. The walk already had the fact (`p.length ≤ off' ≤ bytes.length`) and was discarding it, so it is returned rather than re-derived. Sole other caller (`HeaderExtractNumberBridge`) updated to ignore the new conjunct.

## Relationship to #11483/#11484

This does **not** conflict with the planned `Decoded` relaxation, and I have posted the detailed answer on #11484. Briefly: the composition destructures `Decoded` and **discards all four length facts** — the `32`s it uses come from `AccountRecord.WF`, not from `Decoded`. So relaxing `Decoded` alone leaves every proof here intact, and also does not by itself resolve the FR.

## No registry row

Deliberately omitted. The verdict question is open on #11345 and now bears on #11483/#11484; a row asserting a grade would be guessing at the answer. The theorems are witnessed (`_account_decode_matches_specRef_witness`, `_account_decode_output_witness` in `Progress/Routines.lean`; 130 → 132), so `check-axioms` covers them today and the row can be added once the verdict is settled.

## Verification

`lake build` warning-free; `check-axioms` (132 witnesses, classical only), `check-forbidden-tactics`, `check-layering`, `check-file-size`, `check-unimported`, `check-registry-crosscheck` all pass. Submodule leak count 0.

Refs #11345, #11483, #11484.


pirapira:
**`build` green, zero forbidden tactics, both new theorems under `#print axioms` — `merge!` applied.**

`decoded_matches_specRef` and `outputSuccess_eq_accountDecodedIs` are the composition #11345 has been building toward: `account_decode`'s output struct joined to `AccountRecord` and thence to `SpecRef.decode_account_from_leaf`. Witnessed in `AxiomWitnesses` and registered as routine witnesses, correctly on the codegen side rather than in `Correspondence`, matching the #11351 precedent.

**Stacking is right**: base retargeted to `cab2559ae` (#11481's tip) rather than `main`, so the diff is only this PR's work and GitHub enforces the order. That is the third time tonight this pattern has been used correctly — and the distinction that makes it safe is the base pointer, not the stack.

⚠️ **One coordination note, no action needed here**: #11484 is relaxing `Decoded` for the empty `storage_root`/`code_hash` fold ([`witness_state.py:114-119`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/witness_state.py#L114-L119)). This PR is the composition that consumes `Decoded`, so whichever lands second inherits the arm — and the registry note should then record the fold as a **resolved divergence** rather than as what `a.WF` excludes.

dhsorens:
Thanks — and agreed on the base-pointer point; that is the distinction I got wrong on #11472 and it is the one worth keeping.

**One flag on the coordination note, since our comments crossed.** I posted a detailed answer on #11484 at about the same time as this review, and it disagrees with one word in it: *resolved*.

> the registry note should then record the fold as a **resolved divergence** rather than as what `a.WF` excludes

Relaxing `Decoded` will not resolve it, for two reasons I only established by finishing this composition:

1. **`Decoded` is descriptive, not prescriptive.** It records what the machine established on the success path; the rejection of a zero-length field is a branch in the assembly. Widening the predicate adds a disjunct the code cannot reach. Concretely, the success cell is `fixed32Copied bytes oldRoot o2`, which takes **no length argument** and unconditionally copies 32 bytes — so no edit to `Decoded` makes it yield `EMPTY_TRIE_ROOT`. That needs an asm change plus re-proof through Close1–6.

2. **`AccountRecord` cannot express the degenerate leaf at all.** `rlp = encode rlpItem` *derives* the encoding from the fields, so a zero-length field 2 forces `a.storageRoot = []`, and then `decode_account_from_leaf_accountRlp` would assert `EMPTY_TRIE_ROOT = []`. Relaxing `WF` does not help; it is a modelling change that costs the round-trip property.

Also relevant to "whichever lands second inherits the arm": **this PR inherits nothing.** The composition destructures `Decoded` and discards all four length facts — every `32` and `≤ 8` comes from `AccountRecord.WF`. So it is invariant under the relaxation in either order, and there is no sequencing hazard between them.

No action on this PR. Raising it only because the note would otherwise become a registry row reading "resolved divergence" over a guest that still false-rejects — the same shape as a merged row calling a false reject "not a defect", which is the thing I am still holding #11345's row open to avoid.

Which leaves my one genuinely open question, unanswered since 23:47: **does #11351's merged row stay `.domainRestricted`, or track the port and become `.stricter` with the CPython agreement in the clause table?** I will open a regrade follow-up if it is the latter.

pirapira:
Queued behind #11489 — no action needed from you.

This PR is based on `proof-11345-of-content` (#11481's branch), which is a constituent of batch **#11489** (mid-CI). When that lands, the base branch is deleted on merge and GitHub auto-retargets this PR to `main` (as happened with #11472 earlier today), so no manual retarget should be needed.

I'll pick it up next pass once #11489 lands, with full local verification (`lake build` + all 20 gates).

pirapira:
Folded into #11491, a cross-author batch so these land in one CI round. As predicted, GitHub auto-retargeted this to `main` when #11489 landed and deleted its base branch — no manual retarget was needed. Auto-merge disabled here so it can't race the batch; this PR will close when #11491 lands.

PR #11490, opened 2026-08-05
proof(#11346): split account_exists_and_is_empty into lookup + EMPTY_ACCOUNT kernel

dhsorens:
Item 1 of the #11346 plan I posted on the issue. Independent of the #11345 stack — based on `main`.

## The problem it removes

`account_is_eip161_empty` is handed **leaf bytes**; the named reference `accountExistsAndIsEmpty` (`SpecRef/StateTracker.lean:230`) is `TxM Bool` over an **Address**. No shared vocabulary, so the plan called for naming that boundary in prose per method §5.

Reading the definitions instead, it does not need prose — it is a theorem. `Account` (`SpecRef/Types.lean:274`) has exactly the three fields `EMPTY_ACCOUNT` sets:

```lean
structure Account where
  nonce : Uint
  balance : U256
  codeHash : Hash32
```

so the derived `==` **is** the conjunction the reference writes out by hand, and the function factors cleanly:

```lean
theorem accountExistsAndIsEmpty_eq_kernel (address : Address) :
    accountExistsAndIsEmpty address
      = (do match ← getAccountOptional address with
            | some a => pure (a == EMPTY_ACCOUNT)
            | none => pure false)
```

The only part genuinely outside the guest's scope is the `getAccountOptional` lookup. Everything after it is a pure test the guest can be compared against — which is the comparison point #11346 needs, and it is now established rather than asserted.

`beq_EMPTY_ACCOUNT` carries the reassociation: the derived `==` is in **field** order (`nonce, balance, codeHash`) while `account_exists_and_is_empty` spells it `nonce, codeHash, balance`.

Note this also confirms the guest's shape is right: `accountEip161Empty` (`AccountIsEip161EmptySpec.lean:182`) constrains fields 0, 1 and 3 and deliberately never mentions field 2 — correct, since `Account` has no `storageRoot` and EIP-161 emptiness does not consider one.

## Scope

Deliberately just this. The other two pieces from the issue plan are separate:

- the `Nat.fromBytesBE bs = 0 ↔ ∀ b ∈ bs, b = 0` lemma that turns the guest's non-canonical-zero leniency into a **proved** agreement with the reference rather than a documented coincidence (it belongs in `EL/RLP`, a different layer, and I confirmed no such lemma exists);
- the machine half against `accountEip161Empty`.

**No registry row**, and this is not an oversight: #11483's FR reaches this routine too. The guest requires a 32-byte `code_hash` while the spec folds an empty one to `EMPTY_CODE_HASH`, so a zero-length-`code_hash` leaf with zero nonce and balance *is* `EMPTY_ACCOUNT` per the spec and the guest cannot recognise it. #11346 cannot be graded `.agrees` while that is open. Detail on the issue.

## Verification

`lake build` warning-free; `check-axioms` (132 witnesses, classical only — both new theorems witnessed in `Progress.lean`), `check-forbidden-tactics`, `check-layering`, `check-file-size`, `check-registry-crosscheck` pass. Submodule leak count 0. SpecRef stays Mathlib-free.

Refs #11346, #11483.


pirapira:
**Kernel-clean, both new theorems witnessed, and the split matches the spec's own structure — `merge!` applied.** Fast gates green.

The spec's `account_exists_and_is_empty` ([`forks/amsterdam/state_tracker.py:415`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/state_tracker.py#L415)) is literally two steps — `get_account_optional`, then a conjunction on zero nonce / empty code / zero balance. Splitting the SpecRef mirror into lookup + `EMPTY_ACCOUNT` kernel follows that shape rather than imposing one, which is what keeps the port auditable against the reference line by line.

`accountExistsAndIsEmpty_eq_kernel` and `beq_EMPTY_ACCOUNT` are both under `#print axioms`, and the change is theorems plus a definitional split under `SpecRef` — no behavioural surface.

⚠️ **One thing worth knowing given tonight's live FA (#11439):** `account_exists_and_is_empty` is the predicate behind EIP-161 empty-account deletion, and the FA's mechanism is that the guest records BAL changes for **created-and-deleted** accounts where the spec's emit-on-inequality rule produces nothing. Different code path — that one is the BAL builder, this is the emptiness predicate — but the same *domain*: accounts that exist transiently and end absent.

No conflict with #11490, and nothing for you to do. Flagging it only because if the kernel split later gets reused to decide *whether a change record is emitted*, the emit rule is [`block_access_lists.py:637-664`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/block_access_lists.py#L637-L664) — compare against the pre-tx baseline — and **not** emptiness.

pirapira:
Folded into #11495 so these land in one CI round — both went behind when #11491 landed, so I batched the re-sync rather than doing it twice. Merges are `--no-ff`, so per-commit authorship is preserved. Auto-merge disabled here so it can't race the batch; this PR will close when #11495 lands.

pirapira:
Landed via #11495. Verified: commit `3655ebfc9` is an ancestor of `main` as of `828297981` — closed rather than dropped.

PR #11489, opened 2026-08-05
chore: batch merge #11473 + #11477 + #11478 + #11479 + #11481

pirapira:
Cross-author batch (@dhsorens has said his PRs may be batched) — five `merge!` PRs in one CI round. Real `--no-ff` merges, so per-commit authorship is preserved.

- **#11473** — `fix(codegen): retire bal_txs_independent + deposit-capture route (#11183)` (`0765b83db`) — *pirapira*
- **#11477** — `feat(codegen): resolve execution account code fields` (`0941bd69b`) — *pirapira*
- **#11478** — `refactor(codegen): retire unread storage capture metadata` (`d1f10d39a`) — *pirapira*
- **#11479** — `fix(codegen): drop the field-compare cluster (#11183)` (`ff48eba17`) — *pirapira*
- **#11481** — `proof(#11345): content forms for the four account_decode output cells` (`cab2559ae`) — *dhsorens*

Supersedes **#11482** (#11473+#11477+#11479) and **#11488** (#11478+#11481): #11486 landing put both behind at once, so one rebuild beats re-syncing two batches.

## Class-A ratchet: 5 edges retired, each attributed

**5 removed, 0 added** (baseline 13 → 8), verified against each branch's diff before ratifying:

| lost edge | retired by |
|---|---|
| `.Lbv_code_preimage_flag_done → bal_txs_independent` | #11473 |
| `.Lbmvmx_cb_preok`, `.Lbmvmx_sd_preok`, `.Lbv_b1_loop`, `.Lbv_b23_chk` → `bal_find_account_by_address` | #11479 (drops the symbol) |

## ⭐ Correction to what I wrote on #11482

On the superseded batch I said the `ENDPOINT (#11183)` note in the baseline was prose I "carried across" from #11479's branch. **That was wrong.** #11479 adds the note **to the generator** — `check-bal-class-a-ratchet.py:199` emits it — so `--write-baseline` reproduces it automatically; nothing is carried by hand.

I only caught this because the note vanished after one regen and reappeared after another. Rather than assume, I probed `--write-baseline` directly with marker comments at both the header and EOF positions: **both were dropped**, while ENDPOINT survived the same write — which pointed at the generator rather than any preservation logic.

Practical upshot, consistent with #11451's sidecar design: hand-written comments in `bal-class-a-baseline.tsv` are **not** durable. Anything meant to survive belongs either in the generator (as #11479 did) or in `bal-class-a-notes.md`.

## Verification

- `lake build` — green, 4815 jobs
[...]

PR #11492, opened 2026-08-05
refactor(codegen): parametrize MptDeleteWalkDb over GuestLayout (#10753)

pirapira:
## Summary
- Leaf `MptDeleteWalkDbProg` (`mptDeleteWalkDb_prog_of L`); bridge keeps `mptDeleteWalkDb_prog`
- GuestLayout fields: `mpt_set_record_walk_db`, `mpt_delete_walk_db`
- Bridge still transitively depends GuestAddrs via GuestLayoutInstance

## Byte-id (base = main `de755bf93`)
BASE and HEAD both equal; X matches main's current `.s` (lane invariant).

| artifact | sha256 (base = head = main) |
|---|---|
| `.s` | `c2017183347fc98368784d699f3ee2fbb6041dea949caa2160c1dbb2c5979e60` |
| stripped ELF | `53653a65df269223fbf878078f94f26946411f7f75d7e30683ede00a03332094` |
| `.text` | `15259685add20d40603dd7e2f919e214c691e13b761544877a6fdd7f92b33e3d` |

## Gates
- `check-file` mptDeleteWalkDbFunction CLEAN
- lake build MptDeleteWalkDb OK

Does **not** touch BAL builder record-emit path.

Part of #10753.


pirapira:
**Structure is right; the byte-identity evidence is one base stale and needs re-taking. Holding `merge!` on that alone.**

## Structure verified

| check | result |
|---|---|
| leaf `MptDeleteWalkDbProg.lean` `GuestAddrs` imports | **0** |
| layout access | `L.mpt_delete_walk_db`, `L.mpt_set_record_walk_db` only |
| bridge | 10+/23− with the def name and type preserved |
| `GuestLayout` fields | 17 → 19 |
| BAL builder path | untouched, as stated |

Same leaf-plus-bridge shape as #11486, second module in the lane, and it does not collide with the #11439 FA work.

## The stale measurement

The reported `.s 4819410a…` / stripped ELF `70568e23…` are **the same hashes as #11486**, measured on base `a87b4ed69`. But main has moved to `de755bf93` and its emitted output moved with it — `textSizeBytes` went **`0x598d0` → `0x58bc4`**, a −0x1D0C shrink from batch #11489 (which retired `bal_txs_independent`, the field-compare cluster, and the capture arenas).

So those absolute hashes cannot be main's `.s` any more, which means the identity claim is a valid statement about the change *as measured on the old base* rather than about this PR as it now stands. **Re-take base-vs-head byte-identity on `de755bf93`.** The two hashes should both change and should still equal each other.

⭐ **And there is a stronger invariant available here than a per-PR check.** If every conversion in this lane is byte-neutral, then **main's `.s` hash is unchanged by the lane as a whole** — so the same hash recurring across conversions measured on the *same* base is expected and is itself evidence the lane is not perturbing emitted output. That is worth stating in each PR body as "base and head both X, and X matches main's current `.s`", because it converts a per-PR equality into a running invariant that a stale measurement visibly fails. This PR is the first case where that distinction would have caught something.

pirapira:
**Re-taken on `de755bf93` — `merge!` applied.** Fast gates green.

| check | result |
|---|---|
| base == head | `.s c2017183…`, stripped ELF `53653a65…` |
| differs from the stale pair | yes — `4819410a`/`70568e23` were the `a87b4ed69` measurement |
| **X matches main's current `.s`** | yes — the running invariant, not just a per-PR equality |
| leaf `GuestAddrs` imports | 0; layout via `L.mpt_delete_walk_db`, `L.mpt_set_record_walk_db` |

That middle row is the one that matters. Stating it as *"base and head both X, **and X matches main's current `.s`**"* is what made the previous measurement's staleness visible — a bare base-equals-head is satisfied by any pair of stale hashes, while anchoring to main's current value fails loudly when the base has moved. Worth keeping in this form for every remaining conversion.

Second module in the lane, same leaf-plus-bridge shape as #11486, `GuestLayout` 17 → 19 fields, and no contact with the BAL builder path where #11439's fix will land.

pirapira:
Folded into #11495 so these land in one CI round — both went behind when #11491 landed, so I batched the re-sync rather than doing it twice. Merges are `--no-ff`, so per-commit authorship is preserved. Auto-merge disabled here so it can't race the batch; this PR will close when #11495 lands.

Issue #11493, opened 2026-08-05
schema: which Verdict applies when the guest is stricter than the *port* but matches CPython? (#11351's merged row)


dhsorens:
Splitting this out of #11345/#11484/#11487, where I have asked it four times in PR threads without it landing — it is a **taxonomy question**, not a detail of one row, and it now gates two registry rows.

## The question

`Verdict` (`Progress/Correspondence.lean`) draws the line by *intent*:

```lean
/-- Agrees, but the spec covers strictly less input than the reference accepts.
    Not a defect; a coverage gap callers must respect. -/
| domainRestricted
/-- We reject input the reference accepts → false-rejects on valid chain data. -/
| stricter
```

**When the guest rejects input the *port* accepts, but its strictness matches what CPython actually enforces — which constructor applies?**

## Why it is live

#11351's row (`header_extract_number`, **already merged**) is graded `.domainRestricted`. Its two restrictions are real rejections:

- `tooLong` — the guest rejects a `number` field wider than eight bytes;
- `noncanonical` — the guest rejects a leading zero byte.

The port's `getN` is plain `bytesBEtoNat` and enforces neither, so **against the port the guest false-rejects**. But CPython's `rlp.decode_to` enforces exactly these, so against the real reference implementation the guest is right and the *port* dropped the checks. That is documented in `HeaderExtractNumberBridge.lean:10-21`.

So the row currently says **"Not a defect"** about behaviour that is a false reject *relative to the artifact the correspondence framework actually compares against*. If that reading is wrong, a merged row is misgraded; if it is right, the reasoning should be written down, because this shape will recur every time the port drifts from CPython.

## The two answers, and what I would do

**(a) Verdict tracks the port ⇒ `.stricter`.** The port is what `SpecRef` mirrors and what the rows are proved against, so a rejection the port would accept is a false reject, full stop. CPython agreement is *mitigation* and belongs in the clause table, not in the verdict. → I open a follow-up to regrade #11351 and note the CPython agreement in its clauses.

**(b) Verdict tracks the reference *behaviour* ⇒ `.domainRestricted` stands.** The guest cannot false-reject anything a real chain produces, because CPython would have rejected it too; the divergence is a **port defect**, which should be filed against the port rather than charged to the guest's row. → #11351 stays as-is, and the port gap gets its own issue.

[...]

PR #11496, opened 2026-08-05
proof(#11346): prove the EIP-161 non-canonical-zero leniency agreement

dhsorens:
Item 2 of the #11346 plan. Independent of #11490 (item 1) — based on `main`, no overlapping files.

## What it settles

The guest tests a nonce/balance field for zero by accumulating its content bytes big-endian and comparing to `0` (`beAccFrom_eq_zero_iff`, `AccountIsEip161EmptySpec.lean:171`). That is **lenient**: `0x00`, `0x00 0x00` and the empty field all read as zero. The plan flagged that this agrees with the reference *only because* `bytesBEtoNat` happens to be tolerant too, and asked that the coincidence be **resolved rather than noted**.

It resolves. `Nat.fromBytesBE_eq_zero_iff` (new, `EL/RLP/Properties.lean`):

```lean
Nat.fromBytesBE bs = 0 ↔ ∀ b ∈ bs, b = 0
```

Big-endian zero is length-insensitive, so the reference is lenient in exactly the same way — not accidentally, but because leading zeros contribute nothing to a positional decode. `leniency_agrees` then states the guest's "every content byte is zero" and the reference's `Nat.fromBytesBE … = 0` are the same predicate on the same field.

I checked before writing: no such lemma existed anywhere in the tree.

## ⚠️ A duplication worth knowing about

While transferring #11345's arithmetic I found that **`beAccFrom` and `AccountDecodeSpec.beAccum` are the same function** — character-for-character identical definitions, in two Codegen modules, under two names:

```lean
beAccFrom bs    off | 0 => 0 | i+1 => (beAccFrom bs off i    <<< 8) ||| (bs.getD    (off+i) 0).zeroExtend 64
beAccum   bytes off | 0 => 0 | i+1 => (beAccum   bytes off i <<< 8) ||| (bytes.getD (off+i) 0).zeroExtend 64
```

This is why #11345's `beAccum_eq_fromBytesBE` did not appear to apply to #11346 — the same fact had to be looked for twice under two names, which is precisely the failure mode that has cost me time on this tranche already. `beAccFrom_eq_beAccum` records the identity so the existing lemma transfers rather than being re-proved, and `beAccFrom_eq_fromBytesBE` is that transfer.

**Merging the two definitions is deliberately left as a follow-up** — it would change the vocabulary of an already-proven triple, which is not this PR's job. Happy to file it if you want it tracked.

## Still no registry row

Same reason as #11487 and #11490: the verdict is unsettled (#11493), and #11483's FR independently reaches this routine — the guest requires a 32-byte `code_hash` while the spec folds an empty one to `EMPTY_CODE_HASH`, so a zero-length-`code_hash` leaf that *is* `EMPTY_ACCOUNT` per the spec cannot be recognised.

## Verification

`lake build` warning-free; `check-axioms` (135 witnesses, 132 → 135, classical only — all three theorems witnessed), `check-forbidden-tactics`, `check-layering`, `check-file-size`, `check-registry-crosscheck`, `check-unimported` all pass. Submodule leak count 0.

Refs #11346, #11483, #11493.


pirapira:
**Kernel-clean, three theorems witnessed, and it resolves a coincidence rather than recording it — `merge!` applied.** Fast gates green.

The plan asked that the leniency agreement be **resolved, not noted**, and `Nat.fromBytesBE_eq_zero_iff` does that:

```lean
Nat.fromBytesBE bs = 0 ↔ ∀ b ∈ bs, b = 0
```

Big-endian zero is length-insensitive **because leading zeros contribute nothing to a positional decode** — so the reference's tolerance and the guest's are the same property, not two tolerances that happen to coincide. That is the difference between an agreement one can rely on and one that holds until someone changes a length check.

## The duplication finding is the more valuable half, and I verified it

`beAccFrom` (`AccountIsEip161EmptySpec.lean:72`) and `beAccum` (`AccountDecodeSpec.lean:111`) are the **same function under two names** — confirmed character-for-character apart from whitespace:

```lean
beAccFrom bs    off | 0 => 0 | i+1 => (beAccFrom bs off i <<< 8) ||| (bs.getD    (off+i) 0).zeroExtend 64
beAccum   bytes off | 0 => 0 | i+1 => (beAccum bytes off i <<< 8) ||| (bytes.getD (off+i) 0).zeroExtend 64
```

⭐ **And the consequence stated is the one that matters**: this is why #11345's `beAccum_eq_fromBytesBE` *appeared* not to apply here — the same fact had to be found twice under two names. `beAccFrom_eq_beAccum` records the identity so the existing lemma **transfers instead of being re-proved**.

That is the third instance tonight of *a name obscuring a fact that already existed* — after `single_tx_creation_runtime` (a stale name I relayed as a live hazard) and `toNat_or_shift` (a lemma that looked applicable and was the mirror image). The pattern is consistent: **searching by name finds a name; only reading the statement finds the fact.**

Leaving the definition merge as a follow-up is right — it would change the vocabulary of an already-proven triple, which is not this PR's job. Worth filing so the duplication does not cost a fourth search.

## Registry row correctly withheld

Same reason as #11487 and #11490, plus #11483's FR independently reaching this routine: the guest requires a 32-byte `code_hash` where the spec folds an empty one, so a zero-length-`code_hash` leaf that **is** `EMPTY_ACCOUNT` per the spec cannot be recognised. Claiming a verdict before that is resolved would overstate it. k3's asm fold is on `fix/11483-decode-fold` awaiting your `Decoded` relaxation — landing that first would let this row be stated without the caveat.

PR #11498, opened 2026-08-05
refactor(codegen): parametrize BalAccountHasStateChange over GuestLayout (#10753)

pirapira:
## Summary
- Leaf `BalAccountHasStateChangeProg` (`balAccountHasStateChange_prog_of L`); bridge keeps `balAccountHasStateChange_prog`
- GuestLayout fields: `rlp_walk_init`, `rlp_walk_next`, `bal_account_has_state_change`
- Bridge still transitively depends GuestAddrs via GuestLayoutInstance

## Byte-id (base = main `de4b45699`)
BASE and HEAD both X; X matches main's current `.s` (lane invariant).

| artifact | sha256 (base = head = main) |
|---|---|
| `.s` | `03c9e45c93ed9ceb7fdd8245adb943cfbcfa1e1c6432391062afba13059610ea` |
| stripped ELF | `777824e2e8749ff7016a0f5e771a58b2f168556cd977eb9f6c0b56f28d92344c` |

## Gates
- `check-file` balAccountHasStateChangeFunction CLEAN
- lake build OK

Does **not** touch BAL builder record-emit path (classifier only).

Part of #10753.


pirapira:
Reviewed on content — this is the third conversion and it follows the established pattern exactly. Approving now; **I will apply `merge!` only once the checks are green**, since labelling a red PR displaces the only actor that would regenerate it.

What I checked, beyond the shas:

- **The leaf is genuinely abstract.** `BalAccountHasStateChangeProg.lean` imports `Emit`/`AsmReloc`/`GuestLayout` and **not `GuestAddrs`** — that absence is the whole point of the split, so it is worth stating rather than assuming.
- **The bridge preserves name and type.** `balAccountHasStateChange_prog : Program := balAccountHasStateChange_prog_of guestLayout`, so the concrete-render drift gate keyed on `emitProgram balAccountHasStateChange_prog` still resolves, and consumers compile untouched.
- ⭐ **The field wiring, which is the only place a real defect could hide here.** A field bound to the wrong constant compiles fine and yields a valid-but-wrong image, which no type gives you. All three are name-for-name (`rlp_walk_init := GuestAddrs.rlp_walk_init`, etc.), and **both** instances are updated — `.zero` as well as `guestLayout`. Had `.zero` been missed, `balAccountHasStateChangeFunction` would not have compiled, so that one is self-enforcing; the `guestLayout` arm is not, hence checking it.
- **Base is `de4b45699`, which is current `main`.** Stating this because byte-identity against a *stale* base is a symptom rather than evidence — `rebase` without `fetch` succeeds and lies. Here the identity claim is against fresh main, so it means what it says.

Screening question for the landing bar: *could this cause an invalid block to be accepted?* No — emission is layout-independent, the relocs stay symbolic, and the stripped ELF is byte-identical. Pure refactor, so the light structural gate is the right one.

PR #11500, opened 2026-08-05
fix(codegen): retire dtrc supplied-BAL parse-bail (#11183 ROW 10)

pirapira:
## BIND vs CHECK (first)
**CHECK drop via EQUIVALENCE — not BIND.**

- Site: `.Ldtrc_deferred_marker_ready` after `dtrc_deleg_deferred≠0`
- Call: `bal_find_account_by_address`; **only** `beq a0,2` → `.Ldtrc_bal_unsupported` (a0=3)
- Return consumer: MtxRuntime catch-all `bnez a0` after `beq a0,8` — a0=3 was catch-all only
- Payload: a3/a4 out-ptrs into `bvcd_acct_*` unused post-#11176 (preload retired); a0==1 not-found already fell through
- Spec pin e5a8caf1b fork.py:390: accept iff header commitment = keccak(rlp(built_list)). Spec never parses supplied BAL body for validity
- **Classification: CHECK** (reject on malformed supplied-body parse) with no execution-specs counterpart. Not BIND (BIND would locate the slice for hash without rejecting on parse)

## FA gate — equivalence (no collision assumption, no flip sets)
The retired bail was a **guest-only rejection with no spec counterpart**. Removing it moves the guest **toward** the spec's decision rule, not away from it. Spec and guest then coincide on hash comparison; any collision exposure is identical in both — equivalence needs only the pin, no security assumption.

Supporting facts:
1. `block_access_list_hash` = keccak(raw supplied bytes), **no RLP parse** (`BlockAccessListHash.lean`)
2. Binding rejects after rebuild are **bv_fail 60/61 only** (`bal_serializer_verify` in ReceiptsTail)
3. a0==1 not-found already fell through; only the parse-error arm (a0==2) went

### Where the header BAL commitment enters (verified)
Guest does **not** read a separate header BAL-hash field from input. Path:

1. **StateRoot:1076-1077** — `jal block_access_list_hash` → `svf_bal_hash := keccak(supplied BAL bytes)`
2. **StateRoot:1092** — `svf_bal_hash` stored at `sv_params+96`
3. **BlockVerdictFunction:62** — loads a7 from `96(s0)` into header rebuild
4. **BlockHeaderSszToRlp:77,93,215** — a7 embedded as RLP header field 22 (`block_access_list_hash`)
5. **BlockVerdictFunction:67** — `jal block_hash_from_header` compares rebuilt header RLP hash to payload block hash

So the binding object is the **payload block hash**, which commits to the real header field including BAL hash. Separately `bal_serializer_verify` requires keccak(rebuilt)==keccak(supplied). Together this matches spec's keccak(built)==header commitment (via block-hash binding of keccak(supplied) as the embedded field), without a guest-only parse bail.

## Change
- Deferred path → `.Ldtrc_stage` directly; drop a0=3 sink + `bvcd_acct_*`

## Numbers
- Class-A: 8 → **7**
- textSizeBytes: `0x58d3c` → `0x58cfc`
- region-map CLEAN; ratchet OK 7


pirapira:
Reviewed. **The structural work is right and I am not labelling it yet** — this PR removes a rejection, which is the one category where the maintainer's light structural gate does not apply, and the evidence that category needs is missing.

## What I verified, and it clears the hazards I expected to find

I went in expecting two specific problems, both recorded against this exact site, and neither is present:

- **The out-pointers went with the call.** `a3`/`a4` are out-pointers into `bvcd_acct_ptr`/`bvcd_acct_len`, so removing the `la`s while keeping the `jal` would make the callee write through whatever those registers happened to hold. The diff removes the `la a3`/`la a4`, the `jal`, the `beq a0, t0` bail, and the `.Ldtrc_bal_unsupported` sink **as one block**. No dangling `la`, no orphaned call.
- **No orphaned readers.** `bvcd_acct_ptr`/`bvcd_acct_len` have zero remaining references anywhere in `EvmAsm` on this branch apart from the tombstone comment at `BlockVerdictDataSection.lean:539`. Removing the only writer therefore cannot leave a stale read.
- **The `a0=3` status was consumed by a catch-all, not a specific comparison.** `BlockVerdictMtxRuntime.lean` does `bnez a0, .Lbv_mtx_dispatch_unsupported` after the only specific test (`beq a0, 8` → recipient-unresolvable). So dropping code 3 leaves no dangling equality test — it removes one reason to bail rather than corrupting a decode. Worth stating explicitly because an unused *datum* is not an unused *status*, and here I checked which branches test the return rather than what consumes its payload.

## And the superseded comment is genuinely superseded

`BlockVerdictDispatchTx.lean:346` said outright that this bail *"stays"* — *"INPUT-VALIDITY checks stayed (`a0 == 2` after `bal_find_account_by_address` is a BAL-parse bail, FA-safe, and unrelated to buffer size)."* That is #11176's reasoning, drawn against a **buffer-capacity** question, and it is correct on its own terms. The collision-freedom ruling changes the terms: a guest check with no counterpart in execution-specs cannot be retained, because justifying it requires hash-collision assumptions and collisions provably exist. [`fork.py:390`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/fork.py#L390) hashes the **built** list and never walks the supplied body for parse validity, so the bail is a CHECK and equivalence drops it. Retiring the comment along with the code is right.

## ⛔ What is missing, and it is the whole gate for this class

**No flip sets.** The report is `Class-A 8→7`, `textSize 0x58d3c→0x58cfc`, region-map clean, ratchet OK at 7. Those establish that the retirement happened and that the image is consistent. **None of them speak to what this change makes the verdict accept.**

The screening question for landing is *could this cause an invalid block to be accepted?* For a capability removal the answer is no and the gate is light. **For a rejection removal the answer is structurally yes**, so the evidence comes first — and it is cheap here precisely because the change is small.

Two things needed:

1. **Random-200 flip set both ways.** Any row that was rejected at base and passes with this change has to be named, and each one has to be a *valid* block whose FR is genuinely recovered. That set is exactly the FA-risk set for a rejection removal.
2. ⭐ **The named negative rows, not just the sample.** A clean random-200 with zero flips in either direction has already coexisted with a real false accept on this repo — #11394's candidate false-accepted a negative row while r200 sat at FR=8/OK=232 with no flips at all, and only the 567 named rows caught it. So byte-identical r200 is **not** FA evidence for this class. Show that the blocks with deliberately malformed supplied-BAL bodies still reject, and say through which gate they now reject instead.

If the answer is that malformed supplied-BAL bodies were never rejected for being malformed — that the rebuilt list is built from execution and the hash comparison never consults the supplied body's parse validity — then say so with the site, because that is a stronger argument than any sample and it would close this immediately.

PR #11501, opened 2026-08-05
fix(codegen): filter system storage BAI-0 changes

### LINKED ISSUE #11494: FA (confirmed reachable): BAI-0 system storage descriptors emit unguarded and assert original=zero — spec filters on inequality and keeps the ring's prior value
## Summary

`append_modeled_system_storage_tuple_rows` emits BAI-0 storage descriptor rows with `original = zero` and **no inequality guard**. The spec routes system-transaction state through the **same** `update_builder_from_tx` comparison as ordinary transactions, so an unconditional emit is the same shape as the confirmed FA in #11439 — narrower, but the same rule violated.

Surfaced by a four-field emit-rule census (#11439 follow-up). That census found **no** second violation in the ordinary balance / nonce / code / storage producer paths — all four final emitters compare against the pre-tx baseline before appending. This is the single unguarded producer it turned up, and it was correctly not called a defect on census evidence alone.

## The spec routes system writes through the same rule

[`fork.py:859`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/fork.py#L859) — `process_unchecked_system_transaction` ends with:

```python
incorporate_tx_into_block(system_tx_state, block_env.block_access_list_builder)
```

and [`state_tracker.py:851-856`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/state_tracker.py#L851-L856):

```python
from .block_access_lists import update_builder_from_tx
update_builder_from_tx(builder, tx_state)
```

So beacon-roots ([`:897`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/fork.py#L897)) and history ([`:903`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/fork.py#L903)) contribute through the **same** emit-on-inequality path as user transactions — `_get_pre_tx_storage` baseline, all-limb comparison, emit only on difference ([`block_access_lists.py:666-676`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/block_access_lists.py#L666-L676)). **There is no separate insert-like path for system descriptors.**

Guest side: `BlockVerdictSystemStorageCapture.lean:105-122`, `:163-182` emit the descriptor rows directly with `original = zero`, documented as insert-like, without comparing.

## Why the exposure is narrow but real

EIP-2935 and EIP-4788 write ring-buffer slots keyed by `block_number % HISTORY_SERVE_WINDOW` (8191) and `timestamp % 8191`. The pre-value is whatever the slot held ~8191 blocks earlier. So pre and post **normally differ**, the spec emits, and the guest's unconditional emit agrees.

The divergence needs pre == post:

- **beacon root**: the parent beacon block root recurring at the same ring index — 8191 blocks apart.
- **history**: the same block hash at the same index — likewise.
- ⚠️ **more reachable: `original = zero` on a slot that was never written.** Before the ring fills, an unwritten slot reads zero. If the value being written is *also* zero, spec sees `pre == post == 0` and emits nothing while the guest emits a row.

Failure direction is the same as #11439: a producer committing the guest-shaped hash yields a block the guest accepts and the spec rejects at [`fork.py:390`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/fork.py#L390). **No collision needed.**

## What would settle it

1. Whether a zero-valued system write is reachable — a zero parent beacon root is the obvious candidate, and it is *the genesis/first-slot case* rather than an adversarial one.
2. Whether `original = zero` is ever wrong on a *rewritten* slot — an insert-like emit asserts no prior value where the ring may hold one.

**Fix shape, if confirmed, is the same as #11439's:** route the descriptor emit through the inequality comparison rather than adding a zero-value special case. Two rules compensating for each other is what makes this class hard to reason about.

**Do not bundle with #11439.** That one has a bisect in flight and a named mechanism; this is a distinct producer with a different reachability argument, and merging them would let #11439's evidence stand for a path it never measured.

pirapira:
# Fix BAI-0 system-storage emission against parent state

Closes #11494.

## Mechanism

The Amsterdam execution-specs path passes `parent_beacon_block_root` directly through the system transaction (`src/ethereum/forks/amsterdam/fork.py:897-901`). Its BAL storage-change rule emits a row only when the authenticated pre-transaction value differs from the post value (`src/ethereum/forks/amsterdam/block_access_lists.py:669-676`). A zero `parent_beacon_block_root` is reachable: the pinned reference accepts constructed headers with `parent_beacon_block_root == Bytes32.zero()`.

`BlockVerdictSystemStorageCapture` previously emitted every BAI-0 descriptor and treated `original` as zero. The candidate now resolves the transaction-start value with the existing `slot_at_header_state_root` path, and compares all four 32-byte fields before publishing the BAI-0 row. A missing authenticated slot is the zero pre-state case; the already materialized descriptor post-value is preserved. Equal pre/post descriptors still upsert the execution map but do not publish a BAL change or increment the side-log count. This is separate from #11439's BAL-builder/tombstone work.

The ordinary storage emitter's inequality pattern is reused rather than adding a zero-value special case. This fixes both zero-to-zero rows and rewritten slots whose original is nonzero.

## Generated artefact

- Base: `de4b45699261e89873c3be0da6624b57e924c785`
- Parent ELF: `eb2bcca1d64b2cd2dc8ddd6d27940c622068189b2656ce18400edf5108936b39`
- Candidate ELF: `7111fdea4543fb8d1225a8074c4b68b6e1b140cb6934236be00a6d593dc49f7a`
- Runner (`scripts/spike/spike_run`): `b4286c6e9100b83a0055e921fb4e4f96e8d852efe7f638c65a23a1a3af358349`
- Candidate `.text/.data/.bss`: `0x58e18 / 0x53b0 / 0x19051560`; parent `.text/.data/.bss`: `0x58d3c / 0x53b0 / 0x19051560` (`.text` delta `+0xdc` only).

## Evidence

Prediction before measurement: filtering and resolving the BAI-0 rows should recover false rejects without changing any accepted row; therefore forward flips may be nonzero, reverse flips must be zero.

Command: `python3 scripts/eest-ab-compare.py gen-out/11494-r200-parent gen-out/11494-r200-cand`

- Manifest: `gen-out/eest-run-11435-r200/manifest.tsv`, SHA-256 `85e16487ccbb12569f977208bef20d8a8581f396edffa1d5855f0383ffa3ed84`.
- 200 scored rows per leg; 199 joined input digests (one duplicated input group, consistent within each leg).
- Parent: FA=0, FR=3, agree=197.
- Candidate: FA=0, FR=3, agree=197.
- Delta: FA `0 -> 0`, FR `3 -> 3`; forward flips 0, reverse flips 0; all 199 joined outputs byte-identical.
- Six named beacon/system controls (including expected-invalid `00024`) were also byte-identical parent versus candidate; no named control moved.

Structural gates: `lake build EvmAsm.Codegen.Programs.BlockVerdictSystemStorageCapture`, `scripts/check-asm-to-program.sh`, `NO_BUILD=1 ELF_DIR=gen-out/11494 scripts/check-region-map.sh`, `git diff --check`.

No merge label is being applied by this PR.

## Corpus differential and coverage

The 200-row byte-identity result is a measured no-op on this corpus, not evidence that the changed class was exercised. We computed both predicates on the reachable BAI-0 system-storage rows:

- Random-200: 199 unique inputs, 397 rows. Every row is `both`: the old hardcoded-zero rule emits (`post != 0`) and the candidate authenticated-parent rule emits (`post != authenticated_pre`). Old-only, new-only, neither, and divergent rows are all zero.
- Expanded local manifests: 866 unique inputs, 1,731 rows, again all `both` with no divergence.
- Selected beacon/EIP4788/multi-block subset of the codex2 26,104-row corpus: 129 inputs, 374 rows, again all `both`; 36 rows had nonzero authenticated pre-state, but each had `post != pre`, so both rules still emit.

Thus the scanned corpus contains no BAI-0 row with a nonzero pre-state whose post equals that pre-state (the zero-to-zero beacon-root and rewritten ring-slot cases). The corpus cannot confirm or refute the behavior change for that class; the differential explains why no outputs moved. A constructed zero `parent_beacon_block_root` input did not exercise the new path because the existing system-account/code/vlen gate skipped descriptor emission, so it is not a witness.

This structural fix is justified by the Amsterdam spec rule (`block_access_lists.py:637-676`): emit only on inequality against authenticated parent state, using the real original rather than asserting zero.

## Evidence correction: named nonzero-pre class

The random-200 byte-identity result still has no nonzero-pre BAI-0 row: its 397 reachable rows all have authenticated pre=0, so the old and new emission predicates agree and none of those rows changes. This does not describe the explicitly named beacon corpus, however, and supersedes the earlier no-op framing for that corpus.

On 30 explicitly named codex2 inputs 05890-05919 (90 emitted BAI-0 rows), parent and candidate log row counts matched. Exactly 36 row originals changed, all 36 being the nonzero-pre rows 05893-05911: the parent emits original=0, while the candidate emits the authenticated parent value. None of those 36 wrapped input hashes occurs in the random-200 manifest.

Worked example, input 05893 (wrapped input SHA-256 39d145ed1a242078ff2c42719bafed08ab9a781402f6294f32446547d09df914), EIP-4788 address 0x000f3df6d732807ef1319fb7b8bb8522d0beac02, slot 0x0c: parent original=0x0, candidate original=0x0c, current=0x200b. The same input's slot 0x200b has parent original=0x0 and candidate original=0x6c31fc15422ebad28aaf9089c306702f67540b53c7eea8b7d2941044b027100f. The direct final-memory dumps used parent ELF eb2bcca1d64b2cd2dc8ddd6d27940c622068189b2656ce18400edf5108936b39, candidate ELF 7111fdea4543fb8d1225a8074c4b68b6e1b140cb6934236be00a6d593dc49f7a, and runner b4286c6e9100b83a0055e921fb4e4f96e8d852efe7f638c65a23a1a3af358349.

The broader selected beacon/EIP4788/multi-block reference subset remains 129 inputs and 374 BAI-0 rows with 36 nonzero-pre rows; the direct guest differential above confirms all 36 originals change. Thus the fix is exercised and changes emitted execution-log originals on the named class, while random-200 simply missed that class.

## Verdict outcomes for the named class

The 30 named inputs 05890 through 05919 were also run through the normal 256-byte verdict ABI on both legs. Every row is oracle-valid (expected success=1), but parent and candidate both return success=0: parent FR=30/30 and OK=0; candidate FR=30/30 and OK=0. The transition set is FR-to-FR=30, with zero base-pass-to-candidate-FR, zero candidate-pass rows, and zero reverse flips. This includes every input 05893-05911 carrying the 36 changed originals. All 30 parent/candidate verdict outputs are byte-identical.

This is not a pass/pass comparator-insensitivity result: the named rows remain false rejects on another downstream failure. It does mean the direct original correction is observed in the emitted execution log but does not by itself recover a verdict on this named set; the random-200 and named verdict evidence must be read separately from the log-content differential.


pirapira:
Reviewed. **The patch is right in all three respects I asked for, and I am not labelling it yet** — the 200-row result is being read as confirmation when it is silence.

## The code is correct, including a register hazard it handles well

- **Routed through the inequality, not special-cased on zero.** `.Lamsr_map_equal` upserts the map and emits **no** BAL change row and no side-log bump; `.Lamsr_map_changed` emits. That is the spec's emit-on-inequality rule ([`block_access_lists.py:637-676`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/block_access_lists.py#L637-L676)) rather than a zero-value filter, so it fixes the rewritten-ring case too — which was the point of insisting on one rule for both defects.
- **`original` comes from authenticated parent state.** `slot_at_header_state_root` against `sv_pre_rlp_*`, with a missing authenticated slot treated as a proven zero. The docstring correctly explains why the **block map is not consulted**: MTx seed rows are already in it for later `h_SLOAD`, and they are not pre-state writes preceding the BAI-0 system transaction. That distinction is the whole correctness argument for this path and it is written down at the site.
- ⭐ **The `s5` reuse is safe, and it is a deliberate save rather than a clobber.** This is the shape that caused #11356 — a helper caching its input in a register it later reuses as a destination, safe once and broken twice — so I traced it. `s5` holds `a2` (value pointer) for the value-reversal loop, which completes before `.Lamsr_original_resolve` reassigns `s5 := a1`. From there `s5` is a **saved copy of `a1` across the new `jal ra, slot_at_header_state_root`**, which is exactly the right handling since inserting a call clobbers live argument registers; `.Lamsr_emit_change`'s old `mv a2, a1` becomes `mv a2, s5` and is equivalent. The backward `j .Lamsr_original_resolve` from `.Lamsr_finish_one` also re-reads `a1`, and nothing between entry and that point touches `a1` (the loops use only `t`-registers).
- **The `#guard` counts were raised 3→4** for both `storage_writes_block_upsert` and the overflow check, matching the new arm. Gating the new arm's existence rather than leaving the guard stale is the right instinct.

## ⛔ Why the A/B does not yet support this

200/200 scored, parent FA0 FR3, candidate FA0 FR3, forward/reverse 0, **outputs byte-identical**, six named beacon/system controls byte-identical.

For a change that alters **what the BAL contains**, byte-identical outputs everywhere and an unmoved FR do not mean the fix is safe — they almost certainly mean **the sample does not contain the class the change touches.** That is not a pass; it is the instrument reporting that it cannot see this change at all.

And the reachability probe, good as it was, does not close this gap. Running `validate_header` with `parent_beacon_block_root = Bytes32(0)` proved a zero beacon root is a **valid execution-specs input**. It said nothing about whether any corpus row *has* one. Those are different claims and only the first is established. Likewise the second defect needs a chain older than 8191 blocks for a ring slot to be rewritten, which test fixtures rarely have.

So one of two things, and either is fine — what is not fine is byte-identity offered as verification:

1. **Name a row whose output changed**, and show the change is the row the spec's inequality filter would suppress (or the corrected `original`).
2. **State plainly that the corpus lacks the class**, and then supply the differential, because an unreachable path cannot be validated by any sweep of any size. Compute **both** answers on the reachable BAI-0 rows — the old `original = zero` and the new `slot_at_header_state_root` result — and report where they agree and where they diverge. If they never diverge on this corpus, say exactly that: the change is a no-op here and its correctness rests on the spec argument alone. That is a perfectly good place to land a structural fix; it just has to be the stated conclusion rather than an implied one.

Given you built a constructed case for the reachability question already, option 1 via a constructed input is well within reach and would be strictly better than either.

pirapira:
Follow-up differential audit (to interpret the byte-identical A/B correctly):

The 200-row A/B did not exercise either changed predicate. I decoded each BAL and resolved each reachable BAI-0 system-storage slot against the authenticated parent state (`WitnessState` rooted at the validated parent header `state_root`). On 199 unique inputs from `gen-out/eest-run-11435-r200/manifest.tsv` (one duplicate input group, one non-contiguous-header row skipped), there were 397 BAI-0 rows across the EIP-2935/EIP-4788 system addresses. Every row had `pre=0` and a nonzero `post`, so the old hardcoded-zero path and the candidate path both emit: `both=397`, `old-only=0`, `new-only=0`, `neither=0`; no row diverged.

As a broader reachability check, I repeated the same read-only differential over 866 unique input artefacts from the existing local manifests: 1,731 BAI-0 rows, all `both`, with zero old/new divergences (one non-contiguous-header input skipped). Thus the A/B silence is explained by the available corpus lacking a zero-root or rewritten-slot case; it is not evidence that the changed path is exercised. The spec citation and the code-level comparison remain the correctness basis for those absent classes.

pirapira:
Additional corpus differential for the BAI-0 equality class (candidate 7111fdea; parent eb2bcca1; runner b4286c6e):

- Random-200 manifest: 199 unique inputs, 397 BAI-0 system-storage rows. Every row is `both`: the old `post != 0` rule emits and the new `post != authenticated_pre` rule emits. Old-only, new-only, neither, and divergent rows are all 0.
- Expanded local manifests: 866 unique inputs, 1,731 BAI-0 rows; again all `both`, with no divergences.
- A selected subset of the codex2 26,104-row corpus (129 beacon/EIP4788/multi-block inputs, 374 rows) also has 0 divergences. It includes 36 nonzero-pre rewrites, but every one has `post != pre`, so both rules emit; no net-zero witness was present.

I also tried a constructed zero `parent_beacon_block_root`. It did not exercise this change: the existing system-account/code/vlen gate skipped the EIP4788 descriptor, leaving the root-vlen path empty, and parent/candidate outputs were byte-identical. Therefore the current evidence is corpus silence about the changed equality class, not a correctness claim that the class is unreachable.

PR #11502, opened 2026-08-05
fix(codegen): unlink dead bal_code_preimages_valid (#11183 rows 11-12)

pirapira:
## BIND vs CHECK
**CHECK drop (EQUIVALENCE).** `bbcv_bal_contains_addr` walked supplied BAL body for addr presence — no spec counterpart (fork.py:390 only hashes BUILT list). Post-#11410 the parent `bal_code_preimages_valid` has **0 guest jal**; edges were dead Class-A only.

## Change
- Split `accountStateDelegationCodeResolveFunction` (live)
- Guest V2/BlockVerdict link only that; unlinked dead preimages+bbcv blob
- baseline −bbcv×2

## nm (linked image)
```
account_state_delegation_code_resolve  PRESENT
bal_code_preimages_valid               ABSENT
bbcv_bal_contains_addr                 ABSENT
```

## Numbers
- Class-A: 7 → **6**
- textSizeBytes: `0x58d3c` → `0x566c0`
- region-map CLEAN; ratchet OK 6

## Remainder
BIND×3 stay · shadow · map_builder LAST


pirapira:
Reviewed on content — this is the right shape for a deletion and I have no objection to the argument. Waiting on `build`, and there is a **sequencing requirement** below that does not apply to the other open PRs.

**Why the evidence form is right here.** A static census is bounded in both directions: a folded constant or an indirect jump is invisible to a source grep *and* to an asm grep, so "0 guest `jal`" alone would not settle it. `nm` on the **linked image** is the strong form, because a symbol that survived any reference — direct, indirect, or via a table — could not be absent from the link. Reporting `account_state_delegation_code_resolve` PRESENT alongside `bal_code_preimages_valid` and `bbcv_bal_contains_addr` ABSENT is exactly the pair of facts that makes this checkable: it shows the live routine was preserved by the split rather than that everything was simply dropped.

That also disposes of the status question I would otherwise raise. A return value carries data *and* validity, and a payload with no readers can still gate a rejection — so the test is normally "which branches test this return?" rather than "what consumes its payload?". Here the routine is **not called at all** and is absent from the image, so there is no status to preserve. Stronger than a reachability argument about its callers.

The CHECK classification is right on the same grounds as #11500: `bbcv_bal_contains_addr` walked the **supplied** BAL body for address presence, and [`fork.py:390`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/fork.py#L390) only hashes the **built** list. No spec counterpart, so equivalence drops it.

## ⛔ Sequencing: this one needs a pin **regeneration**, not just a rebase

`textSizeBytes 0x58d3c → 0x566c0` — a ~9.6 KB `.text` shrink. #11500 is labelled `merge!` and also moves `.text` (`0x58d3c → 0x58cfc`). Both regenerate `GuestAddrs.lean`, `RegionMapLinkPins.lean`, and `symbol-addresses.tsv`, so whichever lands second has **stale pins**, and this PR's stated base `textSizeBytes` will no longer describe its base.

This is materially different from #11499, where the rebase only had to drop two duplicate field lines because that PR was byte-identical and main's layout never moved. Here the layout genuinely moves on both sides.

So, after #11500 merges: rebase onto the new main and **regenerate the pins**, then run the generator a second time and confirm that pass is a **no-op**. A non-empty second pass means the first regeneration was incomplete. Never hand-edit a pin or hand-merge a generated file — there are two loss paths, regeneration *and* merge-side resolution, and a batch merge has already resolved a generated file toward main's side here without any gate noticing.

⚠️ And when reporting the regenerated result, give the **base sha next to the artefact sha**. Byte-identity or a pin value quoted against a stale base is a symptom rather than evidence — `rebase` without `fetch` succeeds and lies, and the `textSizeBytes` delta is the field that names a missing commit.

PR #11503, opened 2026-08-05
fix(codegen): route present-but-empty accounts to the authenticated lookup in runtime_same_block_delegation_code (#11439)

pirapira:
Spec: `environment.py:499-500` at e5a8caf1b — a nonexistent account's `EXTCODEHASH` is `U256(0)`; `keccak("")` is only correct for an account that *exists* with empty code.

Guest: `runtime_same_block_delegation_code` reported an absent/tombstone account as a same-block hit (`a0=0`, zeroed code cells), so `h_EXTCODEHASH` hashed `ptr/len` of zero and stored `keccak("")` into storage instead of `0` (lost SSTORE-clear refund, divergent refund accumulator, divergent balances and state root — #11439).

Fix: one immediate — the status-2 dispatch (tombstone-empty) now falls through to the ordinary lookup instead of resolving to empty code. AccountState overlay path kept; the supplied-BAL read (`runtime_current_bal_ptr`) stays retired (absent from the linked image, nm-verified on both legs).

Measurement (candidate vs same-base parent, base `de4b45699`): 04347 passes — `sv_recomputed` == payload root, `evm_refund_acc` reaches `0x9240` (refund granted), rebuilt BAL 559 = supplied 559; 04255 passes — rebuilt 561 = supplied 561. Flip sets (r240): forward 40 (all `selfdestructing_initcode_preserves_balance` rows), reverse **0**. Candidate `e8cf3702…`, parent `64ee6f27…`.

## Per-consumer argument (the fix is a shared-helper fix, three consumers)

`runtime_same_block_delegation_code` feeds three consumers, all control dependencies (EvmAccountWitness `EXTCODEHASH`, EvmAccountWitness `EXTCODESIZE`, and `EvmExtcodecopy`), not just EXTCODEHASH. Routing the tombstone-empty status to the ordinary authenticated account/code lookup is correct for all three because that lookup yields the account's real code: zero-length code for an absent account, so `EXTCODEHASH` pushes `U256(0)` (required for `EMPTY_ACCOUNT` by `environment.py:499-500` at the pin, not `keccak("")`), `EXTCODESIZE` gets 0, and `EXTCODECOPY` copies nothing. An account that exists with empty code still gets `keccak("")` from its real code_hash (the marker status arm is untouched). One rule, three consumers, all corrected.

The EXTCODESIZE/EXTCODECOPY-on-absent-account-with-witness-code-present case is not exercised by a corpus row identified so far; those two consumers rest on this construction argument rather than on a measured fixture.

## Oracle-validity of the 40 forward flips

The 40 forward flips (base-fail to candidate-pass) are all from the single fixture family `test_selfdestructing_initcode_preserves_balance` (~20 unique inputs, with the CALL and CALLCODE entries for the same input being byte-identical rows), not merely a shared name prefix. Every one of the 40 has oracle expected succ = 1 (valid blocks): the flip set is false-reject recovery, not false accepts.

Row ids and expected succ (all = 1): 04239, 04241, 04251, 04253, 04255, 04257, 04259, 04261, 04263, 04265, 04275, 04277, 04287, 04289, 04291, 04293, 04295, 04297, 04299, 04301, 04311, 04313, 04323, 04325, 04327, 04329, 04331, 04333, 04335, 04337, 04347, 04349, 04359, 04361, 04363, 04365, 04367, 04369, 04371, 04373.

## Addendum (review round 3)

**Spec rule is a three-case VALUE comparison.** `extcodehash` (`environment.py:499-500`) pushes `U256(0)` when `account == EMPTY_ACCOUNT`, and `EMPTY_ACCOUNT` (`state.py:62-66`) is `nonce == 0 ∧ balance == 0 ∧ code_hash == EMPTY_CODE_HASH` — a three-field value comparison, not a presence check. Three cases, all measured on the candidate:
1. Absent (deleted, dropped from the trie) → ordinary authenticated lookup yields EMPTY_ACCOUNT → `0`.
2. Present with zero nonce/balance and empty code (the EIP-6780 tombstone class, `account_state_lookup_current` status 2 — all 16 corpus rows per the consumer census) → must also give `0`; this is the class the status conflation broke, and the class this fix restores.
3. Present with empty code but nonzero balance or nonce → must give `keccak256("")`. Verified explicitly on `test_extcodehash_codeless_with_storage-balance` (11810) and `-nonce` (11812), both oracle-valid, both passing on the candidate — the ordinary lookup returns the account's real `code_hash`, which is `EMPTY_CODE_HASH`, and case 3 yields `keccak256("")` exactly as the spec requires. No collapse to 0 anywhere: 40 forward flips (all oracle-valid, single family) and **zero reverse flips** across r240, plus these two explicit case-3 rows.

**The no-supplied-BAL constraint is kernel-checked every build**, not just nm-measured: `RuntimeSameBlockCode.lean:108` carries `#guard` `NOT Function.contains runtime_current_bal_ptr` (and siblings for `rlp_list_count_items` and `code_state_lookup_current`). The #342fffd0-era Class-A retirement of the supplied-BAL read stays retired.

**Fixture regeneration**: `scripts/asm-fixtures/runtimeSameBlockDelegationCodeFunction.s` is a snapshot/lockfile regenerated by the conversion pipeline, not a guest build input; the guest ELF is assembled from the Lean Programs. The fixture was regenerated with the status-2 branch retargeted `.Lrsbd_empty_hit` → `.Lrsbd_miss`, matching the one-immediate Program change (the drift gate's deliberate acknowledgment of the image change). All measurements were made on ELFs built from the fixed Lean source.


pirapira:
Review record. **The FA gate is satisfied and the design is right; one build gate is outstanding.**

## The gate that mattered: all 40 forward flips are oracle-valid

A newly-passing row is an FR recovery if the block is valid and a **false accept** if it is not, and those two are indistinguishable in a flip count — which is why I would not accept "40 forward, 0 reverse" on its own. All 40 carry expected `succ = 1` from the oracle, zero oracle-invalid, and they are **one real fixture family** (`test_selfdestructing_initcode_preserves_balance`, ~20 unique inputs, with the CALL and CALLCODE entries being byte-identical rows) rather than a shared name prefix. So this is false-reject recovery on valid blocks.

`reverse = 0` is also meaningful here rather than silent, for a structural reason: the change is **one immediate on one arm**, so the set of behaviours it can alter is bounded. That is the situation where a zero reverse count carries information.

## What I verified independently

- **The diff is exactly one immediate** (status-2 `BEQ` 96→128) against its base. Nothing else. That matters because **the status-1 arm is untouched** — this file exists for EIP-7702 same-block code observation, where current code can be the `0xef0100‖address` marker while the pre-state trie has empty code, and my concern was that routing status 2 to the ordinary pre-state lookup would lose that marker. It cannot: status 1 selects that path. An impossibility argument, stronger than a 7702 fixture.
- **No layout move**, so no pin regeneration is owed (unlike #11502).
- ⭐ **`RuntimeSameBlockCode.lean:108` already carries `#guard !(…contains "runtime_current_bal_ptr")`**, with siblings for `rlp_list_count_items` and `code_state_lookup_current`. That is a *better* durable gate for "the supplied-BAL read stays retired" than the `nm` check I asked for, because the kernel checks it on every build rather than a human checking it once. `342fffd0e`'s Class-A retirement is protected by construction.

## The spec rule, corrected — it is a three-field value comparison

I had been describing this as *absent → 0, exists-with-empty-code → `keccak(empty)`*. That is wrong, and the PR title still carries it. `extcodehash` tests `account == EMPTY_ACCOUNT`, and [`state.py:62-66`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/state.py#L62-L66) defines that as `nonce=0` **and** `balance=0` **and** `code_hash=EMPTY_CODE_HASH`. Three cases:

| account | codehash |
|---|---|
| absent (`get_account` → `EMPTY_ACCOUNT`) | `0` |
| **present, nonce 0, balance 0, empty code** | `0` — equals `EMPTY_ACCOUNT` |
| empty code but **nonzero nonce or balance** (a funded EOA) | `keccak(empty)` |

codex2's 16-row census shows `account_state_lookup_current` returns **status 2 (present-but-empty)** on all 16, and **none** return status 0. So the middle row is what this PR fixes, and the citation needs `environment.py:499-500` *together with* `state.py:62-66` — without the second the rule reads as a presence test and the funded-EOA case looks like it should return 0.

## The other two consumers rest on construction, and that is now measured

The shared helper has three consumers, all **control** dependencies: EXTCODEHASH (value + control), EXTCODESIZE (control, value differs when witness code exists), EXTCODECOPY (control + value). Note that all three being control means the usual reject-only safety net — wrong value → wrong root → reject, FR never FA — **never applied to this class**.

codex2 searched for the discriminating fixture and found **none**: no corpus row exercises EXTCODESIZE or EXTCODECOPY with a status-2 account *and* nonzero `rsbd` length. The delegated-EOA candidates it measured (02260 EXTCODESIZE, 02258 EXTCODECOPY) return status **0** with `a0=1` and `rsbd` ptr/len both zero — read through the ordinary path. So those two consumers are corrected **by construction** (the authenticated lookup yields the real code), not by measurement, and the body should say exactly that rather than implying coverage.

## ⛔ Outstanding: `asm-to-program` drift

`runtimeSameBlockDelegationCodeFunction: emitted SYMBOLIC render is not byte-identical to the hand-written source (would change the guest/probe images)`. `scripts/asm-fixtures/runtimeSameBlockDelegationCodeFunction.s` still carries the old branch target and is not in the diff.

**An empty diff on a file the change must have altered is itself the finding.** If that `.s` is a snapshot and the guest is assembled from the Lean side, this is a lockfile to regenerate and the measurements stand. If it is a build input, the pushed tip does not contain the change that produced them and they need re-confirming on a build from this tip. That question decides the label, and it has been put to the author.

pirapira:
Verified on the **pushed tip**, not the PR page — `git ls-remote origin refs/pull/11503/head` → `12fdd0f0f5`, base `ea328971f` (current main), two commits (`44bc4123c` fix, `12fdd0f0f` fixture), and both files present in the diff. Worth stating because the earlier state of this PR had a corrected *title* with an unchanged *branch*: a title edit goes through the API and never touches the commits, so the page looked freshly worked while CI was still failing on the old tip. The check that catches that is diffing the pushed tip.

The two changes correspond exactly:

```
Program:  .BEQ .x10 .x5 (96)  →  (128)
Fixture:  li t0, 2; beq a0, t0, .Lrsbd_empty_hit  →  .Lrsbd_miss
```

⭐ And the fixture regeneration earns more than gate satisfaction here — it makes the change **self-documenting**. `(128)` says nothing to a reader; `beq a0, t0, .Lrsbd_miss` says that status 2 is now a *miss* rather than an *empty hit*, which is precisely the semantic correction. Anyone auditing this later reads the intent off the asm instead of reconstructing it from a branch offset.

With this, the remaining evidence is complete:

- **FA gate:** all 40 forward flips oracle-valid (`succ = 1`), zero oracle-invalid, one real fixture family (`test_selfdestructing_initcode_preserves_balance`), reverse 0 — meaningful here because one immediate on one arm bounds what can change.
- **Case 3 checked deliberately** rather than inferred from `reverse = 0`: fixtures 11810 (nonzero balance) and 11812 (nonzero nonce), both oracle-valid, both `succ = 1`, `keccak(empty)` preserved with no collapse to 0. That is the direction 40 same-family flips could not reach.
- **Status-1 arm untouched**, so the EIP-7702 same-block delegation path is unaffected by construction — stronger than a 7702 fixture.
- **`#guard !(…contains "runtime_current_bal_ptr")`** at `RuntimeSameBlockCode.lean:108` keeps `342fffd0e`'s Class-A retirement protected by the kernel on every build, rather than by a one-time `nm` check.
- **Three-case rule** in the body with `environment.py:499-500` *and* `state.py:62-66`.

Awaiting green; I will label then. Note for whoever reads this later: the two other consumers of this helper (EXTCODESIZE, EXTCODECOPY) are corrected **by construction and untested by measurement** — no corpus row exercises them with a status-2 account and nonzero `rsbd` length, which is tracked as #11504. All three consumers are *control* dependencies, so the reject-only argument (wrong value → wrong root → reject, FR never FA) never applied to this class.

pirapira:
**Labelled `merge!`.** Green on the verified pushed tip (`12fdd0f0f5`, base `ea328971f`, all six checks SUCCESS).

## Correcting something I said about the independent replay

codex2 replayed the 16 CALLCODE rows against the candidate and got 16/16 verdict 1, having first confirmed the candidate ELF sha (`4a69962c…`) differs from the parent's (`a9cd9c7e…`) — the right discipline, and the step that distinguishes a real replay from measuring the parent twice.

I told codex2 that run carried more weight than the author's own numbers because those 16 rows were not tuned against. **That was wrong, and I checked rather than leaving it asserted.** All 16 are inside the 40 listed here:

```
codex2:  04255 04257 04259 04261 04291 04293 04295 04297 04327 04329 04331 04333 04363 04365 04367 04369
k3 (40): 04239 … 04255 04257 04259 04261 … 04291 04293 04295 04297 … 04327 04329 04331 04333 … 04363 04365 04367 04369 04371 04373
```

So the replay is a **reproduction** — same population, different harness and different measurer — not an independent population. That is still worth having: it confirms the numbers through a second path and it caught nothing wrong. But it does not extend coverage, and **the union is 40 rows, not 56.** Anyone quoting FR recovery from this PR should use 40.

## Why the FA gate is satisfied

Since the 16 are a subset of the 40, the oracle check already covers them: **all 40 forward flips carry expected `succ = 1`, zero oracle-invalid**, in one real fixture family (`test_selfdestructing_initcode_preserves_balance`, ~20 unique inputs whose CALL and CALLCODE entries are byte-identical rows — which is exactly why 20 inputs yield 40 rows). A newly-passing row is an FR recovery if the block is valid and a false accept if it is not, and those are indistinguishable in a flip count, so this was the gate that mattered.

`reverse = 0` is meaningful here rather than silent because the change is **one immediate on one arm**, which bounds what can differ.

## The rest of the case, for the record

- **Status-1 arm untouched**, so the EIP-7702 same-block delegation path (`0xef0100‖address` marker over an empty pre-state trie) is unaffected by construction — stronger than any 7702 fixture.
- **Case 3 checked deliberately**, not inferred from `reverse = 0`: fixtures 11810 (nonzero balance) and 11812 (nonzero nonce), both oracle-valid, both `succ = 1`, `keccak(empty)` preserved with no collapse to 0.
- **`#guard !(…contains "runtime_current_bal_ptr")`** at `RuntimeSameBlockCode.lean:108` keeps `342fffd0e`'s Class-A retirement enforced by the kernel on every build.
- **Three-case rule** cited to `environment.py:499-500` *and* `state.py:62-66`, since `EMPTY_ACCOUNT` is a three-field value comparison rather than a presence test.

⚠️ Carried forward on #11504: the other two consumers of this helper (EXTCODESIZE, EXTCODECOPY) are corrected **by construction and untested by measurement** — no corpus row exercises them with a status-2 account and nonzero `rsbd` length. All three consumers are *control* dependencies, so the reject-only argument never applied to this class.

Issue #11506, opened 2026-08-05
obs: bal_serializer_verify code 61 (rebuild/sort fail) never fires on full corpus


pirapira:
## Observation

Full-corpus spike census over 26,104 EEST rows (guest still had map_builder B/C at census time):

| fail_code | count |
|-----------|------:|
| 0 | 24289 |
| 5 | 948 |
| 1 | 491 |
| err | 174 |
| **60** | **128** |
| 4 | 30 |
| 40 | 16 |
| … | |
| **61** | **0** |
| 66 | 0 |

Among binding-gate rejects, **only 60** (digest mismatch) appears. **61** (rebuild/sort failure from `bal_serializer_verify` a0=2 → `bv_fail_code=61`) never fires.

## Why separate

Not a row-8 / Class-A finding. 61 is a distinct BAL-hash failure mode in the comparator (`bal_serializer_rebuild_hash` nonzero → verify returns 2 → ReceiptsTail sets 61). A mode no corpus row reaches is its own question about the comparator / sort path reachability.

## Not FR

Histogram is over all rows including oracle-invalid intended failures. Do not read non-zero codes as FR.

## Artifacts

`/tmp/opencode/code66-census/` (shared census run; includes fail-code histogram).

## Ask

Is 61 dead by construction on this guest, or is the corpus missing the sort-failure shape?

pirapira:
## Not dead by construction — and the source read found a capacity mismatch

Source-only, no corpus run (the "does it fire" half is folded into the one scoped post-#11525 sweep).

**What the two codes mean:**

- **60** — rebuild succeeded, and the supplied/rebuilt hash **mismatched**.
- **61** — `rebuild_hash` nonzero, i.e. **the rebuild itself failed**; reaches verify's status 2 and `ReceiptsTail`'s code 61.

**Sort statuses** (`BalCanonicalSort.lean:222-223`): status 1 = count above `lui 5` = `0x5000` = **20480**; status 2 = invalid segment count; status 3 = zero key width or range-stack overflow.

## ⭐ The mismatch

The sort refuses any stream above **20480** entries. But the declared per-stream builder capacities are **larger**:

| stream | declared capacity | sort limit |
|---|---|---|
| storage | **47,522** | 20,480 |
| balance | **105,000** | 20,480 |
| nonce | **35,000** | 20,480 |

[...]

PR #11507, opened 2026-08-05
fix(codegen): map_builder DIR A only — drop supplied-BAL B/C (#11183)

pirapira:
## Summary
- **KEEP DIR A**: map finals ↔ highest-BAI builder (`bal_map_final_value_matches`). Guest-internal fail-safe; map not serialised.
- **DROP DIR B/C**: builder↔supplied BAL. Emitted into hashed BAL → ⊆ 60/61. EQUIVALENCE `fork.py:390`, not collision.
- Call site bare `jal` (no `bv_bal_*` loads).
- Kernel `#guard !(…contains \"bv_bal_start/len\"|rlp_walk|…)`.

## Rebase regen (post-#11502)
| | |
|--|--|
| base | `babac0014` (main textSize **0x5675c**) |
| tip | `365a756a0` |
| textSizeBytes | **0x55c30** EXACT (= 0x5675c − 0xb2c; prediction hit; no alignment slack) |
| independence | 11502 unlinks preimages/bbcv; 11507 DIR B/C only — non-overlapping |
| Class-A ratchet | **OK 4** baselined paths (programme endpoint: BIND×3 + residual if any; shadow annotated/gated separate) |
| nm | `bal_map_builder_consistent` + `bal_map_final_value_matches` only; has_row/B/C ABSENT; bbcv/preimages ABSENT |
| check-region-map | CLEAN |
| check-bal-class-a-ratchet | OK 4; annotation sidecar 1/1 |

## Class-A programme endpoint
Rows 8 (this PR DIR A-only), 9 shadow annotated, 10 dtrc, 11–12 bbcv resolved. Remaining baselined = **4** permanent BIND-family paths at `.Lbv_after_tx_gate` (+ any non-CHECK residual). Annotation EXPECTED=1 stays.

## Not in this PR
- Option 3 refused. **#11506** code 61 never fires.


pirapira:
Reviewed on content. **The split is right and the enforcement is better than what I asked for.** Not labelling yet — see the sequencing note at the end.

## Why keeping DIR A and dropping B/C is the correct split

The three directions in `bal_map_builder_consistent` are not the same kind of object, which only became visible once they were separated field by field:

- **DIR B** (builder row → supplied tuple) and **DIR C** (supplied tuple → builder row) read the **supplied BAL body**. Every field they compare — addr BE20, BAI at +24, balance u256, nonce u64, code ptr/len — is emitted by the serialiser into the hashed bytes. So their rejection set is a genuine **subset** of the binding gate's: any disagreement they catch also makes `rebuilt ≠ supplied`, which [`fork.py:390`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/fork.py#L390) already rejects. They have no spec counterpart, and equivalence drops them.
- **DIR A** (map finals ↔ highest-BAI builder reduction) reads **no supplied data at all** — its inputs are the `account_writes` map and the builder streams, both guest-internal. It is therefore not a Class-A edge, and the collision-freedom ruling does not reach it: that ruling concerns guest checks on *supplied* content, which would need collision resistance to excuse.

⛔ **And DIR A had to stay for the reason the field-by-field turned up: map finals are not serialised.** The serialiser emits only builder rows, so a map-versus-builder-reduction divergence trips 66 while `rebuilt == supplied` and the hash still matches. Dropping it would have removed a rejection the binding gate does **not** replace — the false-accept direction. This is exactly the "compared field that never reaches the hashed bytes" case, and it was found rather than assumed away.

Keeping A is coherent with equivalence because **A is a fail-safe internal invariant, not a content check**: it can only fire when our own producers have desynced the map and the builder, i.e. when the guest is broken. `hits66 = 0` across all 26,104 rows is consistent with the invariant holding. Equivalence is a claim about correct-guest behaviour, and removing A would let a broken guest proceed to hash a wrong BAL.

## The gate I set, and what was delivered instead

I asked for the retirement to be verified on the **linked image** rather than by source grep. That was done — the emitted `bal_map_builder_consistent` body carries zero `bv_bal_start`/`bv_bal_len`/`rlp_walk` references, `nm` shows only `final_value` plus the top symbol, and the Class-A baseline row retires (ratchet OK at 6).

⭐ But the better thing is at `BalMapBuilderConsistent.lean:307-308`:

```lean
#guard !(balMapBuilderConsistentFunction.contains "bv_bal_start")
#guard !(balMapBuilderConsistentFunction.contains "bv_bal_len")
```

That converts a one-time human check into a **kernel-checked invariant on every build**, the same pattern that protects `342fffd0e`'s retirement at `RuntimeSameBlockCode.lean:108`. A future edit that reintroduces a supplied-BAL read into this function now fails to compile. That is strictly stronger than the `nm` evidence and it is what should be cited going forward.

The callsite change is consistent: `la t0, bv_bal_start; ld a0 … la t0, bv_bal_len; ld a1 …` is removed from `BlockVerdictMtxTail`, so the supplied BAL is no longer passed in at all — the argument registers went with the call's purpose rather than being left dangling.

## ⛔ Sequencing

#11502 is labelled `merge!`, unmerged, and also moves `.text` (`0x5675c`). This PR is based on `d6047135f` with `textSize 0x582ac`. Whichever lands second has **stale pins**, so once #11502 merges this needs a rebase and a pin regeneration through the canonical `gen-out/regionmap` path, with a second pass confirmed as a no-op — and its current build will be invalidated in the process. Label follows that, not this build.

pirapira:
**Labelled `merge!`** on verified tip `365a756a07`, base `babac0014`, all six checks green.

## Lifting the hold, and why the reason I gave for it was wrong

I held this PR while #11508 (four genuine false accepts in main) and #11509 (the FA classifier scoring crashes as accepts) were open, on the stated grounds that I would not certify anything against a measurement I had just called untrustworthy.

That reasoning does not apply here, and I should have checked it before imposing the hold rather than after. **None of this PR's safety argument depends on the FA/FR classifier:**

- The core claim is an **implication proved from the code**: `bal_map_builder_consistent` DIR B/C fire only when the supplied BAL body disagrees with the builder's balance/nonce/code rows, every one of those compared fields is emitted by the serialiser into the hashed bytes, therefore any disagreement they catch also makes `rebuilt ≠ supplied`, which [`fork.py:390`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/fork.py#L390) already rejects. That is a subset relation between two code paths. **No count enters it.**
- ⭐ And it is **robust to the classifier defect specifically.** The `hits66 = 0` census could in principle be understated — a row that crashed before reaching the check would not record a hit — but that does not weaken the argument at all: even if 66 *did* fire somewhere, its condition still implies hash-mismatch, so the binding gate still rejects the same block. The census bounds the claim; the implication carries it.
- The retirement itself is verified structurally, not statistically: zero `bv_bal_start`/`bv_bal_len`/`rlp_walk` in the emitted body, the callsite's argument loads removed with the call, and **five `#guard !(…contains …)` assertions** that fail the build if a future edit reintroduces a supplied-BAL read.

## Independence from the two open defects

- **#11508's four false accepts** are witness-incompleteness fixtures (`*_missing_*`) flipping at `6ebe3869c`, the resolver batch — a different mechanism, a different code path, and one that this PR does not touch.
- **#11509's classifier defect** affects how outcomes are *counted*, not what the guest *does*. It cannot be masked or worsened by removing a check whose rejection set is a proven subset of a check that remains.
- DIR A is **retained**, and it is the one direction whose rejection is *not* covered by the hash gate, since map finals are never serialised. That was the finding that made this a split rather than a deletion.

## What stays measured rather than assumed

`textSizeBytes 0x55c30` was predicted from `0x5675c − 0xb2c` before the rebase and hit exactly, which is what establishes the regeneration ran against post-#11502 main rather than reproducing a stale tree. Canonical `gen-out/regionmap` provenance. Class-A baseline ends at **4 rows** — three `.Lbv_after_tx_gate` BIND rows plus the settled shadow diagnostic — so **zero supplied-BAL content checks remain**, which is the programme's endpoint rather than merely a smaller number.

Issue #11509, opened 2026-08-05
P0 instrument: FA classification reads out[32] without rc, so guest crashes are scored as accepts (3 of 7 in #11508)


pirapira:
## The FA category is contaminated by crashes — the soundness metric cannot be trusted as reported

Found while investigating #11508. Splitting it out because it is independent of that regression and, in one sense, more serious: **FA is the number we treat as the soundness gate, and it currently over-reports.**

### Evidence

Three fixtures (`00349`, `00350`, `26003`) were classified **FA** by the full-corpus sweep. Instrumented, they are guest **crashes**:

```
rc=3   mcause=0x5 (LOAD_ACCESS)   mtval=0x20
out[32] (succ) = 1        ← read, but never written by the guest
non-zero bytes in 256-byte output = 5   ← verdict write never happened
```

At an earlier build the same fixtures reject cleanly: `rc=0`, `succ=0`, `bv_fail=7`, 44–51 non-zero output bytes.

So the classifier appears to decide from **`out[32]` alone**, without consulting `rc`. When a run faults before writing its verdict, whatever that byte happens to hold is taken as the answer — and `1` means accept.

### Why this matters beyond three rows

- The sweep **does** have a `FAULT` category (1,752 rows in the same run), so it can distinguish faults. These three landed in FA anyway, which means the two categories are not mutually exclusive in practice and **FA is a superset of what it claims to be.**
- An inflated FA is not a safe direction to be wrong in. It burns P0 attention on artefacts — a 63-commit bisect was started against these three and would have cost about six builds — and it undermines the credibility of the number when it *is* real. In the same run, four FA rows were genuine.
- ⛔ Conversely, the reverse error is worse and is not ruled out: **if a faulting run's stale buffer can read `1`, it can equally read `0`.** A crash on a genuinely invalid block would then be scored a correct rejection, and a real false accept could be hidden inside the FAULT bucket. Nobody has checked that direction.

### Suggested fix

1. **A run with `rc != 0` must never be classified by its output bytes.** Classify it `FAULT` unconditionally, and treat FAULT as its own outcome — neither accept nor reject.
2. **Zero the output buffer before every run**, so a stale or unwritten byte cannot be read as a verdict.
3. **Require a positive verdict-written signal** rather than inferring one from a single byte — the non-zero byte count already discriminates cleanly here (5 vs 44+).
[...]

PR #11511, opened 2026-08-05
fix(codegen): retarget eip8037_tx_gas_gate fail branches to epilogue (#11510)

### LINKED ISSUE #11510: crash: eip8037_tx_gas_gate empty-tx fail branches land mid-epilogue (Group A FA=7 artefact)
## Summary

Group A fixtures (00349, 00350, 26003) **guest CRASH** (`rc=3` `mcause=0x5` LOAD_ACCESS `mtval=0x20`) rather than clean reject. Sweep classifier (#11509) scored these as FA because FA checked before FAULT.

Real FA count from the FA=7 sweep is **4** (Group B missing_*), not 7.

## Symptom

| field | value |
|-------|-------|
| mepc | `bgv_u32le` entry (`lbu t0,0(a0)`) |
| mtval / a0 | `0x20` |
| ra (faulting jal) | inside `eip8037_tx_gas_gate` third `bgv_u32le` (`mv a0,s5; jal bgv_u32le`) |
| fixtures | `bal_gas_limit_boundary` below_boundary **no_tx**; `fork_transition_bal_size` |
| baseline clean | `020802bd7` rc=0 succ=0 bv_fail=7 out_nz=44–51 |
| crash from | `322763b71` onward (incl. main `d6047135f`) |

## Root cause (confirmed on guest-322763b71.elf)

`eip8037_tx_gas_gate` uses **hardcoded** forward branch immediates to the fail epilogue:

```lean
-- BlockVerdictGasGate.lean (post b042bfe8d / 2db77771e)
.BLTU .x5 .x21 (2076 : BitVec 13),
.BEQ .x22 .x0 (2072 : BitVec 13),
...
```

On empty tx list (`tx_off == withdrawals_off` ⇒ `s6=0`), `beqz s6` is taken. Objdump shows the target is **mid-epilogue**:

```
; GOOD (020802bd7) fail @ +0x8b8:
8002eca4:  li  a0,0
8002eca8:  ld  ra,0(sp)
8002ecac:  ld  s0,...
... full restore ...

; BAD (322763b71) branch target @ +0x898:
8002cc6c:  ld  s4,40(sp)     ; <-- skips li a0 + ld ra/s0/s1/s2/s3
8002cc70:  ld  s5,48(sp)
...
8002cc8c:  addi sp,sp,112
8002cc90:  ret               ; ra still = return from 2nd bgv_u32le!
```

Correct epilogue start on bad ELF is `8002cc54: li a0,0` (24 bytes earlier). Branches overshoot by exactly `li a0` + `ld ra/s0/s1/s2/s3` (6×4 = 24 B).

### Crash cascade

1. Empty-tx path takes mis-targeted `beqz s6`.
2. Epilogue restores s4–s11 only; **does not restore ra**.
3. `ret` jumps back to mid-function (`8002c444`) with **stack already popped**.
4. Execution continues with corrupted frame; eventually `s5=0x20` and third `bgv_u32le` LOAD_ACCESSes `0x20`.

Debug evidence (`SPIKE_DEBUG_CMD` until `0x8002c454`):
- at `beqz`: s0=exec `0x4000004e`, s5=tx_list `0x4000026a`, s6=0, a0=tx_off=wd_off=`0x21c`
- one step later: pc=`0x8002cc6c` (mid-epilogue), not `0x8002cc54`

## Introduced by

Hardcoded immediates last set in **`b042bfe8d`** (`codegen: remove redundant EIP-8037 BAL loads`, 2100→2076).  
**`2db77771e`** (`#11428`, remove `mv x9,x11` / `mv x18,x12`) left 2076 unchanged.

Window: `020802bd7..322763b71`. Related: #11424 / #11428 / #11407 EIP-8037 edits.

## Fix direction (do not land without coord)

1. Retarget empty/fail branches to full epilogue (`li a0,0` + restore ra..s11), **or** replace hardcoded BitVec offsets with label/reloc form so body edits cannot desync.
2. Audit **all** other hardcoded `BitVec 13` forward branches in `eip8037TxGasGate_prog` after the same size deltas.
3. Re-measure Group A: expect clean reject bv_fail=7 (or current gas code), not crash.

## Xref

- Classifier: #11509 (FA before FAULT)
- Sweep issue context: #11508 / FA=7
- Class-A map_builder HOLD: #11507 (unrelated; unblocked in principle)
- Artifacts: `/tmp/opencode/fa7-compare/` (ELFs, fault TSV, dbg logs)

## Pin

execution-specs amsterdam `e5a8caf1b` (not required for this crash; structural codegen bug).

pirapira:
## Summary
- #11510: 36 neutral branches in `eip8037_tx_gas_gate` landed **mid-epilogue** (`ld s4`), skipping `li a0,0` + `ld ra` → **stale-ra control-flow corruption**.
- Retarget all 36 to `li a0,0` @ idx 544; geometry `#guard 4*i+off = 4*544` ×36; `brOff` helper.

## Spec alignment (honest)
**Closer indirectly.** Spec never returns a corrupted verdict; removing CF corruption removes a divergence. Offsets are not spec-derived — not a byte-identical retarget. Justification = flip set (raw fields) below.

Caller: `bnez a0, .Lbv_eip8037_gas_fail` → **a0=0 PASS**, nonzero FAIL. The 36 land on `li a0,0` (early-OK / empty-tx neutral). Status fails use `LI a0,1/2/3`.

## Gate gap
**No gate caught this.** asm-to-program / pins / Class-A verify emitted==intended bytes; **none verifies a branch target is a named join point.** Geometry `#guard` pins this file; durable class → #11515.

## Blast — FULL 26104, raw (rc,succ,bv_fail) only (#11509 classifier broken)

| | sha256 | base |
|--|--------|------|
| parent | `6a2ef66c7850716f…` | ef4e44d4a |
| cand | `0dd64611738e582e…` | 68c7c7edd |

### NARROW GATE: succ 0→1 with rc=0 both sides
**0.** Bar met.

### FAULT sizing (true crash = rc=3; timeouts excluded)
| metric | n |
|--------|--:|
| parent rc=3 | **1432** |
| cand rc=3 | **0** |
| rc=3 → non-fault | **1432 (100%)** |
| parent timeout rc=124 (load flake) | 429 |
| cand timeout rc=124 (load flake) | 417 |

Prior sweep FAULT≈1752 ≈ 1432 crashes + ~300 other/timeouts. **Every measured rc=3 crash was this defect.** FAULT bucket was hiding real verdicts wholesale. Combined with #11509 (crash scored accept), FR=581 / FA figures were badly distorted.

Timeouts: 13 cand-only timeouts under parallel load; **serial rerun all identical parent=cand** (not real new faults).

### Other transitions (rc=3 → 0)
| class | n | notes |
|-------|--:|-------|
| →accept succ=1 | **1427** | **all oracle=1** restored true accepts |
| →reject succ=0 | **5** | see below |
| succ 1→0 rc=0 both | **0** | |
| identical bulk | ~24.6k | |

**→reject five:**
| id | bv_fail | oracle | note |
|----|--------:|-------:|------|
| 00349 | 7 | 0 | Group A CLEAN REJECT |
| 00350 | 7 | 0 | Group A CLEAN REJECT |
| 26003 | 7 | 0 | Group A CLEAN REJECT |
| 02967 | 11 | 0 | missing preimage family |
| **23017** | **56** | **1** | **NEW FR** deposit multi-same-eoa; parent crashed, cand rejects |

**23017** is the only oracle-valid row that becomes a clean reject. Name it; FR-direction, tolerable but real.

Artifacts: `/tmp/opencode/11510-blast/{blast.tsv,FINAL_SUMMARY.txt,rerun/}`.

## Census
- Mid-epilogue hazards after fix: **0**
- Latent long branches: **374 / 72 files** → #11515 (GuestLayout + byte-identity)

## Closes
#11510. No merge! from grok.


pirapira:
**Labelled `merge!`** — tip `68c7c7edde`, base `ef4e44d4a`, all six checks green, and the gate is met.

## The gate

Full 26,104 rows, raw `(rc, succ, bv_fail)`, parent `ef4e44d4a` (`6a2ef66c…`) vs candidate (`0dd64611…`):

**`succ 0→1` with `rc=0` on both sides: 0.** No row that rejected now accepts, so the fix introduces no false accept. That was the whole bar, and it is met on the full corpus rather than a sample.

## ⭐ The number that reframes the corpus

| | rc=3 (true crash) |
|---|---|
| parent | **1,432** |
| candidate | **0** |

**Every measured crash in the corpus was this one defect.** 1,432 rows — 5.5% of all rows — were faulting because 36 branches in `eip8037_tx_gas_gate` landed mid-epilogue, skipping `ld ra`, and returning with a stale `ra`.

That is far larger than the three fixtures that exposed it, and it means **the corpus measurement was corrupted wholesale.** Every FR/FA/OK figure quoted over the period this was live was computed with 1,432 rows producing no real verdict at all — and, per #11509, being *classified anyway* from whatever bytes `out[32]` happened to hold.

## Consequence: this is the prerequisite for any trustworthy measurement

Two defects were distorting the same numbers, and they compound:

1. **#11510 (this PR)** — 1,432 rows never reached a verdict.
2. **#11509** — the classifier read `out[32]` without consulting `rc`, so those rows were scored as FA / FR / OK from stale bytes rather than counted as faults.

So the ordering is: **land this, then land #11509's classifier fix, then re-measure.** A sweep taken before this lands is measuring 1,432 rows that are about to change.

⚠️ **And expect the FA count to be able to rise.** Those 1,432 rows now produce real verdicts for the first time. codex2's corrected sweep already found **8 rows moving OK → FAULT** — rows that were crashing while being counted as correct outcomes. With crashes eliminated rather than merely reclassified, whatever those rows genuinely do becomes visible, and some may be false accepts that a stale `0` byte had been scoring as correct rejections. That is the direction nobody has been able to see.

## On the fix itself

Retargeting 36 branches to `li a0, 0` at idx 544, with the geometry `#guard` (`4*i + off = 4*544` across all 36) so the next body move fails the build rather than silently corrupting control flow. The guard is worth more than the fix — it makes the class unrepeatable in this routine. `.text` moves, pins regenerated through the canonical path.

Sibling class tracked in #11515: 374 further hardcoded branch immediates across 72 files, correct today and distance-coupled, to be retargeted during GuestLayout parametrization with byte identity as the proof.

pirapira:
⚠️ **Correction to my review comment above.** I wrote that codex2's sweep found *"8 rows moving OK → FAULT — rows that were crashing while being counted as correct outcomes."* That is wrong. Per #11525, those 8 came from the diagnostic run using the since-removed output-density threshold (`nonzero > 5`), so they were reclassified by a **density heuristic rather than by `rc`** — most plausibly legitimate completed verdicts with sparse output, which is precisely the failure mode I had objected to in that heuristic. I then used its output as evidence about the classifier replacing it.

What stands from this PR is unaffected: parent `rc=3` = **1,432**, candidate = **0**, and the narrow gate `succ 0→1` with `rc=0` both sides = **0**. Those are raw `rc`/`succ` measurements and owe nothing to the classifier. Full correction on #11509.

Issue #11513, opened 2026-08-05
SpecRef bug: _decode_header's getN drops the width and canonicality checks that rlp.decode_to enforces (9 numeric fields)


dhsorens:
Filed so #11493's new `SpecRef`-bug verdict has an issue to cite, and because the defect is broader than the one row that surfaced it.

## The divergence

`_decode_header`'s port builds every numeric header field with a plain big-endian conversion (`SpecRef/Stateless.lean:62`):

```lean
let getN := fun i => bytesBEtoNat (bs.getD i [])
```

No width bound, and no rejection of a leading zero byte. The reference does a **typed** decode ([`stateless.py:249`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/stateless.py#L249)):

```python
return rlp.decode_to(Header, header_bytes)
```

`decode_to` is driven by `Header`'s field annotations (`Uint`, `U64`, `Bytes32`, …), so a field wider than its annotated type, or one with a non-canonical leading zero, is an error there and a silently-accepted value here. **The port is `looser` than the Python.**

## Scope: 9 fields, not 1

`mkHeader` uses `getN` for `difficulty`, `number`, `gasLimit`, `gasUsed`, `timestamp`, `baseFeePerGas`, `blobGasUsed`, `excessBlobGas`, `slotNumber` (`Stateless.lean:66-71`). Every one inherits the leniency. `getB` fields are unaffected — those are byte strings either way, and `items.mapM rlpBytes?` already sends `.list` to `none`.

## Why it matters to the registry rather than to the guest

The guest is on the **correct** side here. `header_extract_number` rejects both a >8-byte field (`Result.tooLong`) and a leading-zero field (`Result.noncanonical`) — matching the reference. But graded guest-against-port, that reads as *the guest being stricter*, which is why #11351's row currently carries `.domainRestricted` and why I initially proposed `.stricter` on #11493.

Both grades misattribute the defect. `.stricter` in particular would record a false-reject against the guest for behaviour where the guest agrees with the spec and the port does not. That is what the new verdict is for, and this issue is what it should cite.

## ⚠️ One honest caveat about the mitigation claim

The claim "`rlp.decode_to` enforces exactly these" is about **`ethereum_rlp`, which the submodule pin does not vendor**. `execution-specs/pyproject.toml:28` constrains it only as a range:

```
[...]

PR #11514, opened 2026-08-05
schema(#11493): add a Verdict for SpecRef port defects, and regrade the one row that is one

dhsorens:
Implements the verdict you asked for on #11493, and applies it — which **replaces** the regrade I had recommended rather than confirming it.

## The constructor

```lean
| specRefBug (issue : Nat)
```

The guest matches the Python; the `SpecRef` port does not. The issue number is mandatory, enforced in the registry rather than requested in prose:

```lean
theorem specRefBug_cites_issue :
    registry.all (fun e => match e.verdict with
      | .specRefBug issue => issue != 0
      | _ => true) = true := by decide
```

A note can say "the port is wrong" and nobody ever fixes it; a required issue number cannot. `countVerdict` can't count a payload-carrying constructor by equality, so `countSpecRefBug` joins the census.

## ⭐ Your one-liner corrected me, and I want to be explicit about how

On #11493 I argued the port/CPython question was *secondary* and recommended `header_extract_number → .stricter`. **That was wrong**, and the row's own note already held the refutation I had read past:

> it is the PORT that dropped the checks, not the guest that invented them

The guest rejects a numeric header field wider than eight bytes (`Result.tooLong`) or carrying a leading zero (`Result.noncanonical`). `_decode_header` decodes via `rlp.decode_to(Header, header_bytes)` — a **typed** decode that rejects both. Only the port's `getN = bytesBEtoNat` accepts them.

So `.stricter` would have recorded a false-reject against the side that is behaving **correctly**, and `.domainRestricted` was wrong the other way ("Not a defect" over a real divergence). The defect is real and it is reference-side. That is the distinction your constructor draws and neither of my two candidates could.

Filed as **#11513**, which also records that the defect covers **nine** `getN` fields — `difficulty`, `number`, `gasLimit`, `gasUsed`, `timestamp`, `baseFeePerGas`, `blobGasUsed`, `excessBlobGas`, `slotNumber` — not just the one this row is about.

## A second thing the regrade surfaced: the clause cites an unverifiable source

The row asserted the guest's strictness "matches CPython's `rlp.decode_to`". That is about **`ethereum_rlp`, which the pin does not vendor** — so `scripts/check-spec-refs.sh` cannot machine-check the citation the way it checks a `forks/…/x.py:NNN` line. For `.ported`, whose entire requirement is a clause-by-clause table against the source, that clause is **read, not verified**.

⚠️ **Correction, made in `fixup` on this branch:** I first wrote that this was a *version-drift* risk, citing `pyproject.toml:28`'s range `>=0.1.6,<0.2`. That was wrong — `docs/agents/spec-correspondence.md` §6a already says "only `uv.lock` pins", and `uv.lock` resolves `ethereum-rlp == 0.1.6` exactly, with a sha256. 0.1.6 is also the load-bearing version: §6a records that 0.1.5 silently accepts trailing bytes where 0.1.6 rejects them. So the pin is present and correct, and no dependency change is needed. Only the checkability gap survives, which is materially smaller than what I filed on #11513.

## Docstrings

Both the schema and `docs/agents/spec-correspondence.md` gain the "ask *whose* limit it is" table:

| limit lives in | verdict | is it a defect? |
|---|---|---|
| the **proof** (we did not cover the case) | `domainRestricted` | no — prove more |
| the **triple's precondition** (the ABI obliges the caller) | `domainRestricted` | no, but it can be *violated* |
| the **guest** (it refuses input the spec accepts) | `stricter` | **yes, guest-side** |
| the **port** (it accepts input the Python rejects) | `specRefBug` | **yes, reference-side** |

`domainRestricted`'s docstring now names the two distinct situations it legitimately covers — coverage gap vs ABI precondition — which are benign in different ways and make different promises, the third of the three shapes I'd found conflated.

## Counts

`domainRestricted` 4 → 3, `countSpecRefBug` = 1. **`stricter` stays 0** — and now correctly, rather than as the anomaly that started the issue: this row was never a guest defect.

`docs/rlp-spec-correspondence.md`'s family-scoped counts are untouched by a `header` regrade. Drive-by there: it said "19 of the registry's 21 rows" and the registry is 25.

## Deliberately not in this PR

- **#11345's row.** Still absent, and it should stay absent until #11483 lands: the fold turns that false-reject into agreement, so any grade added now would be stale within the week.
- **`rlp_encode_uint_be`.** I flagged on the issue that a precondition and a coverage gap are different promises even though both are benign. The docstring now says so; whether that row wants splitting off `domainRestricted` is a separate call.

`check-forbidden-tactics` and `check-axioms` both green (137 witnesses, classical-3 only).

Refs #11493, #11351, #11513, #11480.


pirapira:
Reviewed. **The taxonomy table is the contribution, and it is worth more than the constructor.** `build` is still running; one substantive point below that I would like addressed before it lands.

## What's right

The four-row table — *which side is limited?* — makes a distinction the schema was silently collapsing:

| limit lives in | verdict | defect? |
|---|---|---|
| the proof | `domainRestricted` | no |
| the triple's precondition | `domainRestricted` | no, but violable |
| the **guest** | `stricter` | yes, guest-side |
| the **port** | `specRefBug` | yes, reference-side |

⭐ And the diagnosis of how #11493 arose is the sharpest observation here: **`stricter` had zero occurrences despite describing the most safety-relevant outcome in the schema.** A verdict that never fires is either impossible or misapplied, and here it was the latter — guest rejections were wearing a verdict whose docstring reads *"Not a defect."* That is the same shape as `hits66 = 0` on #11507 and the `no_looser_verdicts` invariant itself: a counter stuck at zero is evidence about the instrument, not about the code.

Splitting `domainRestricted`'s docstring into coverage-gap versus ABI-precondition is also a real distinction, and the reason given is the right one: **the second can be violated, the first cannot.**

Enforcing the issue number in the registry (`specRefBug_cites_issue`) rather than requesting it in prose is correct, and for the stated reason — a note saying "the port is wrong" can rot indefinitely; a required number cannot.

## ⛔ The one thing I want tightened

**`specRefBug` is the only verdict in the schema that asserts the guest is *correct*.** Every other verdict is neutral about the guest (`domainRestricted`, `unproven`, `noCounterpart`) or blames it (`stricter`, `looser`). So it carries the highest potential to excuse a genuine guest defect — and the only thing kernel-enforced about it is that some integer is non-zero.

Your worked example does it properly: `header_extract_number` cites that **`rlp.decode_to` rejects those inputs too**, so the guest agrees with the Python and only `getN = bytesBEtoNat` in the port accepts them. That is three-way evidence — guest, Python, port — and it is exactly what the verdict needs.

But that rigour is **exemplified, not required.** A future row could be graded `specRefBug` on the strength of "the guest disagrees with SpecRef" alone, which is equally consistent with the guest being wrong, and the registry would record it as reference-side with a tracking issue attached. That is the failure mode this verdict makes cheapest.

⭐ Since the verdict's whole claim is *"we already match execution-specs and our port does not,"* **it should be obliged to cite execution-specs.** Concretely, any of:

- require the row's `reference` to name the Python function whose behaviour the guest matches (mechanical, and `basis := .ported` already implies a port comparison rather than a spec one); or
- extend the theorem so a `specRefBug` row must carry a non-empty `reference`; or
- at minimum, state in the constructor's docstring that a port disagreement is **not sufficient** — the note must record the Python behaviour, at the pin, that the guest agrees with.

The first two are kernel-checked, which is the spirit of the rest of this file. The third is the floor.

## Context

This lands squarely on two things opened in the last hour, and it improves both:

- **#11516** — I claimed no gate ties asm-side specs to the reference. Wrong: this registry is exactly that, and I have corrected the issue. Its real gap is coverage — all 25 rows are `family := "rlp"`, and account-leaf decode was never graded, which is why `Decoded`'s `l2 = l3 = 32` divergence from both SpecRef and the spec went unsurfaced until four fixtures were being false-accepted.
- **#11517** — the correspondence survey now says *extend this registry* rather than build anything, and adopts its vocabulary. `specRefBug` supplies an outcome my taxonomy lacked entirely: I had collapsed "the port is wrong" into "divergence," which points the fix at the wrong side.

pirapira:
## Maintainer ruling, and it supersedes my review point with a better one

**`specRefBug` should be neutral on the guest, and combined with the guest verdicts rather than replacing them.**

I had asked for `specRefBug` to be obliged to cite execution-specs, on the grounds that it is the only verdict asserting the guest is correct. That treats the symptom. The actual problem is structural: **`specRefBug` conflates two independent axes into one field.**

The port's correctness and the guest's correctness are orthogonal facts. All four combinations are real:

| guest vs Python | port vs Python | should be recordable as |
|---|---|---|
| agrees | agrees | `agrees`, no port defect |
| agrees | wrong | `agrees` **+** port defect → the `header_extract_number` case |
| stricter | agrees | `stricter` |
| stricter | wrong | `stricter` **+** port defect — *both*, independently |

As a `Verdict` constructor, `specRefBug` makes the second and fourth rows unrepresentable: grading a row `specRefBug` **erases the guest grade**. The registry's whole stated value is that its verdicts say *who* is wrong — and this constructor can only say "the port," never "the port, and also here is the guest's standing."

## ⛔ And it opens a hole in the one kernel-checked soundness invariant

`no_looser_verdicts` exists so that recording a `looser` row fails elaboration — a divergence in the false-accept direction cannot land quietly. But `looser` and `specRefBug` are alternatives in the same enum, so **a row that ought to be `looser` can be graded `specRefBug` instead and the invariant passes.** The guest accepts what the reference rejects, the theorem never fires, and the row reads as a port defect with a tracking issue attached.

That is not hypothetical bad faith — it is exactly the pressure the PR body describes for `domainRestricted`: a verdict whose docstring says "not a defect" attracted grades that belonged elsewhere, which is how #11493 began and why `stricter` sat at zero occurrences. A new constructor that is *even more* exculpatory will attract the same traffic, and this time it bypasses the soundness theorem rather than merely mislabelling.

## Suggested shape

Keep `Verdict` describing **the guest** only, and carry the port defect orthogonally — e.g. `portDefect : Option Nat` on the entry, with the mandatory-issue check becoming `portDefect_cites_issue` over `some 0`. Then:

- `verdict := .agrees, portDefect := some 11513` — the `header_extract_number` row, and it now says *both* things: the guest matches the Python **and** the port is broken.
- `no_looser_verdicts` continues to range over every row's guest verdict with no escape hatch.
- `countSpecRefBug` becomes a count over `portDefect.isSome`, which sidesteps the payload-equality problem that motivated it.
- The four-row taxonomy table survives intact — it is a table about *which side is limited*, and separating the axes is what lets it be read off a row rather than inferred.

Everything else in the PR stands, and the parts I praised are unaffected: the taxonomy table, the `domainRestricted` split into coverage-gap versus ABI-precondition, and enforcing the issue number in the registry rather than in prose.

⚠️ One consequence worth stating: under the combined form, `header_extract_number` needs its guest verdict decided on its own merits. On the evidence in the PR body — the guest rejects >8-byte and leading-zero numeric fields, and `rlp.decode_to` rejects them too — that is `agrees`, and the port is the only outlier. Which is a stronger claim than the current row makes, and a checkable one.

pirapira:
^ probably it's better to keep two dimensions for guest issues and specref issues.

dhsorens:
Done — pushed. You were right on both counts, and the second one is a hole I introduced, so thank you for catching it.

## The soundness point in particular

I hadn't seen that `looser` and `specRefBug` being alternatives in the same enum lets a row that ought to be `looser` be graded around `no_looser_verdicts`. That's worse than mislabelling: the gate exists precisely so a false-accept can't land quietly, and I'd added a constructor that routes past it while looking exculpatory. Exactly the pressure I described for `domainRestricted`, one level worse.

```lean
portDefect : Option Nat := none      -- on Entry
theorem portDefect_cites_issue       -- rejects `some 0`
def countPortDefect                  -- over `.isSome`
```

`Verdict` is back to its six guest-only constructors, so `no_looser_verdicts` again ranges over every row with no escape hatch.

⭐ **The verdict census is now unchanged by this PR** — `domainRestricted` back to 4. That's what "orthogonal" should mean: the row stays in its verdict bucket *and* gains a recorded port defect, rather than being moved out of one to record the other.

## One place I did not follow the review

You suggested the row's guest verdict becomes `.agrees`. I've kept `.domainRestricted`, because the row carries a **second restriction the content evidence doesn't touch**: arity.

The guest never checks the header's field count, so the triple is one-directional — `_decode_header = .ok h → guest succeeds ∧ value = h.number`, not an iff. On a wrong-arity header the guest still returns a value where the reference errors, and **we have not proved what it does there**. That's unproven, not known-correct: taxonomy row 1, a proof limit. `.agrees` would overclaim on a fact the triple doesn't establish.

Which I think is a further argument for your design rather than against it. With the axes separated the row can say *both* — "domain-restricted on arity, port defect on content" — where either single grade has to drop one. Rendered as:

```
| `header` | guest | `header_extract_number` | … | domain-restricted **+ port defect (#11513)** | ported | … |
```

If you read the arity restriction differently — e.g. that `decode_header_inv` already constrains arity enough to make the guest verdict `agrees` — say so and I'll change it; it's a judgement about what the triple establishes, not something I want to hold if you see it otherwise.

Also split the two `#11483` asm bugs out as you asked: **#11518** (reversed `SD` operands — wild write, ships) and **#11519** (`laLo` anchored at the ADDI — model-only, invisible to every byte-level gate). #11519's write-up includes the possible `emitProgramR` consistency gate, offered separately rather than folded into either PR.

Gates green: `check-forbidden-tactics`, `check-axioms` (137 witnesses, classical-3), PROGRESS.md drift.

pirapira:
**Labelled `merge!`** — verified tip `e2593c56a7`, base `ef4e44d4a`, all six checks green.

The orthogonal-axes change is in, and it goes further than what was asked:

- `portDefect : Option Nat := none` on `Entry`, with `portDefect_cites_issue` rejecting `some 0`.
- `Verdict` now describes **the guest only**, so `no_looser_verdicts` ranges over every row's guest verdict with **no escape hatch** — a row that ought to be `looser` can no longer be graded away as a port defect.
- The taxonomy table's fourth row reads *"the port … → `portDefect` field"*, so the two axes are visible in the table rather than inferred.

⭐ And the regraded row is better than either of us proposed. `header_extract_number` becomes `verdict := .domainRestricted, basis := .ported, portDefect := some 11513`, with the note recording that **the guest verdict is `domainRestricted` on arity grounds alone** while the content divergence is reference-side in `portDefect`. I had assumed that row would come out `agrees`; separating the axes exposed that its guest-side standing rests on a *different* limitation than the port bug, and that distinction was invisible while one field carried both. That is the exercise justifying itself on its first row.

For the record, the review sequence on this PR: I first asked for `specRefBug` to be obliged to cite execution-specs, which treated the symptom. The maintainer's ruling — neutral on the guest, combined with the guest verdicts — was the structural fix, and it closed a hole I had not seen: as an enum constructor, `specRefBug` could have bypassed `no_looser_verdicts` entirely, since the two were alternatives in the same type.

Everything praised earlier stands: the four-row *which side is limited* table, the `domainRestricted` split into coverage-gap versus ABI-precondition (*"the second can be violated, the first cannot"*), and enforcing the issue number in the registry rather than in prose.

⚠️ Follow-on, not blocking: per #11516 this registry's coverage is `family := "rlp"` only, which is why `Decoded`'s `l2 = l3 = 32` divergence from both SpecRef and the spec went unsurfaced until four fixtures were being false-accepted. #11517 tracks extending it, and the `portDefect`/verdict separation is exactly the vocabulary that survey needs.

pirapira:
Verified and armed — `lake build` green (4820 jobs), all 20 gates `ok`, 0 behind `main`.

Reviewed this one rather than just gating it, since it touches the correspondence schema. Three things I checked beyond the gates:

**1. It doesn't actually add a `Verdict` constructor**, despite the title — and that's the right call, made explicitly. `portDefect` is an orthogonal `Entry` field, and the docstring gives the reason: folding it into `Verdict` would make "guest agrees AND port is wrong" unrepresentable, and would put an alternative beside `looser` in the same enum, letting a false-accept be graded around `no_looser_verdicts`. Confirmed `Verdict` gains no constructor and `no_looser_verdicts` is unchanged.

**2. The verdict census is genuinely unchanged.** `header_extract_number` stays `.domainRestricted` on its arity restriction and only gains `portDefect := some 11513`; `verdict_counts` is untouched, and `stricter` still has 0 occurrences. So nothing was silently regraded to look better — the second axis was added alongside, exactly as the note claims.

**3. `portDefect_cites_issue` has teeth, and the cited issue is real.** I patched the row to `portDefect := some 0` and elaboration fails (3 errors), so "the port is wrong" can't be asserted without a tracked defect. The invariant can only check `issue != 0`, though — it structurally cannot check the issue *exists* or is *about this*, which is the same provenance-vs-fidelity gap §6a names for `SpecRef` citations. So I checked by hand: **#11513 is open and on point** — *"SpecRef bug: `_decode_header`'s getN drops the width and canonicality checks that `rlp.decode_to` enforces (9 numeric fields)"*.

The taxonomy table (proof / precondition / guest / port) is a genuine improvement on the previous state where a guest rejection could wear `domainRestricted`, whose docstring says "Not a defect".

PR #11525, opened 2026-08-05
fix(sweep): classify emulator faults before verdicts

pirapira:
## Summary

Scripts-only fix for issue 11509. The sweep now zeroes the output buffer before every run and classifies any nonzero emulator rc, missing output, short output relative to dump_len, or missing verdict as FAULT before comparing FA and FR. This keeps sparse completed verdicts visible instead of using an output-density heuristic.

## Evidence

Focused regression checks pass:

- rc=0 with one nonzero byte at offset 32 and oracle 0 classifies as FA.
- Any nonzero rc takes precedence over the verdict byte.
- Short output classifies as FAULT.
- A clean rc=0 rejection remains OK.
- Python compilation and git diff check pass.

The exploratory before/after movement matrix from the full target replay was: FA to FA 3; FA to FAULT 4; FR to FR 571; OK to OK 23761; OK to FAULT 8; FAULT to FAULT 1757; all other directions zero. The eight OK to FAULT rows were the diagnostic run using the removed output-density threshold; the final classifier deliberately has no nonzero-byte threshold, so it cannot hide a sparse false accept.

The d604 historical control was run with a pinned runner, manifest, and guest artifact; its guest variant hash differed from the independently established normal-guest control, so that result is reported separately and is not used as the current-main measurement.

## Scope

No guest, codegen, layout, or generated files changed.


pirapira:
**Labelled `merge!`** — tip `376b109cce`, base `f56b516f5`, all six checks green, one file (`scripts/sweep_one.py`).

## What this fixes

`classify` computed `fault = (rc != 0 or missing data/succ)` but tested **FA and FR before FAULT**, so a faulting run whose stale `out[32]` happened to read `1` against an invalid-block oracle was scored **FA** and never reached the fault branch. The flag was computed and then shadowed. `spike_run.cc` compounds it by always reading `OUTPUT_ADDR` and writing the output file even on `flagv=2`, so the file always exists and always parses — it may simply contain bytes the guest never wrote.

Now: buffer zeroed before each run; any nonzero `rc`, missing output, **short output relative to `dump_len`**, or missing verdict is FAULT *before* FA/FR are compared.

⭐ **The `dump_len` comparison is better than what I asked for.** I suggested dropping the byte-count rule and classifying on `rc` alone. This keeps a real completeness test — actual length against *expected* length — which catches a truncated write that returned `rc=0`, without inventing a threshold. My rc-only version would have missed that.

## Why the earlier density heuristic had to go

An interim version required `nonzero > 5` as a "complete verdict" signal, fitted to two observations (faulted runs showed 5, real verdicts 44–51). A legitimate accept whose counters are all genuinely zero has **one** nonzero byte, at offset 32 — so that rule would have reclassified a real false accept as a crash and **hidden** it. The regression set now pins exactly that case: `rc=0`, one nonzero byte at offset 32, oracle 0 → **FA**.

The asymmetry that governs this file: **over-reporting FA wastes time; under-reporting FA ships an unsound guest.**

## Evidence, and one correction to my own reading

The two-sided control on `d6047135f` — where ground truth was established independently by grok's flip test — returned **FA=3** (`02965`, `02970`, `02998`, all `rc=0`) against a pre-registered expectation of 4. The shortfall is explained rather than unexplained: `02967` returns `rc=3` on three fresh runs with `mepc=0x8001c6bc`, **the same PC grok recorded for the #11510 stale-`ra` crashes on that exact build**. A crashed run produces no verdict, so it cannot be a false accept there. No `FA → OK` movement in any direction, which is what shows the corrected classifier hides nothing.

⚠️ And I have corrected a claim of mine that appeared here and on #11511: the `OK → FAULT 8` transition came from the **removed density-threshold run**, not from `rc`. Those rows were reclassified by the heuristic I had objected to, so they are most plausibly sparse legitimate verdicts rather than hidden crashes. I used the flawed instrument's output to reason about its replacement. The defensible claim about the old classifier is narrower: **it scored 4 crashes as false accepts.** Whether it also hid anything in the OK bucket is unestablished.

## Scope note

This does not measure anything; it changes what we can *see*. The sweep it unblocks is now scoped as a **one-off hunt for false accepts that crashes were masking** — a crash whose stale byte read `0` would have been scored a correct rejection, and that direction has never been observable — rather than as a figure to quote. To be run against a guest built from a main that already contains this fix.

Issue #11526, opened 2026-08-05
codegen: nested delegation access charge treats lookup failures as not-delegated


pirapira:
## Finding

`callDelegationAccessChargeAsm` suppresses the EIP-7702 delegated-target access charge on every nonzero `code_at_header_state_root` status. This is false-accept-direction: undercharging can let a block pass when the execution-specs gas check should exhaust gas. The issue is scoped to the nested CALL-family path; the separate top-level deferred-delegation helper does charge its selected target.

## Guest evidence

`EvmAsm/Codegen/Programs/ChildFrameHandlerTailHelpers.lean:31-40` calls `code_at_header_state_root` and immediately branches `bnez t2, .Lcdac_done_*`. Its documented status contract in `StateCompose.lean:836-843` is:

- status 0: account and code preimage found; marker inspection continues;
- status 1: account absent; this is a legitimate no-code/no-delegation case;
- status 2: state-trie MPT parse failure;
- status 3: account decode failure;
- status 4: header/state-root parse or size failure;
- status 5: non-empty code hash missing from `witness.codes`.

Therefore statuses 2–5, including malformed and witness-incomplete lookups, take the same no-charge arm as an absent account and the nested frame continues. That collapses error/incompleteness into benign “not delegated.”

If status 0 finds the 23-byte `ef0100` marker, `account_state_delegation_code_resolve` is then called (`:49-68`). Its documented status 0 (same-block selected target) and status 1 (retain prior-block marker) reach the delegated charge; status 2 (active precompile/empty/deleted target) exits, which is a distinct legitimate non-delegation result. The defect is the earlier `code_at_header_state_root` nonzero collapse, not this status-2 precompile arm.

The shared nested tail invokes this helper for all four CALL-family handlers (`ChildFrameHandlers.lean:122-142`). Before it, `ChildFrameHandlerTails.lean:56-60` charges the original CALL target through `runtime_access_account_charge`, and `:78-83` records that original target. That does not compensate the missing delegated-target charge: the helper at `ChildFrameHandlerTailHelpers.lean:72-79` is the only nested charge for the resolved delegate. The top-level path is separate: `BlockVerdictDispatchTx.lean:40-47` defines `delegationAccessChargeAsm`, called at `:101` and `:166`, and is not the affected nested site.

## Spec obligation

[...]

Issue #11520, opened 2026-08-05
Census: every execution-specs raise is a rejection point — find where our guest compiled one into a fallback value instead (#11508 is one instance)


pirapira:
## The rule

The guest is **execution-specs compiled to RISC-V** under two resource constraints: **no allocator** (fixed arenas and capacity bounds instead of growth) and **no exceptions** (status codes instead of `raise`). Program logic is *carried over*, not re-derived.

That makes the translation of `raise` unambiguous:

> ⭐ **A spec `raise` compiles to a REJECTION. Never to a fallback value.**

Anywhere we return a default, an empty, a zero, or an "unavailable" status where the spec would have raised, we have **different logic** — not a resource-constrained translation of the same logic. And since the spec raises exactly where it cannot proceed soundly, every such site is a candidate **false accept**.

## The instance that motivates this

#11508: `f8eaf501c` gave the tx-storage-map miss path *absent/unavailable* semantics with do-not-reject behaviour, where [`witness_state.py:212`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/witness_state.py#L212) raises `KeyError` on a missing code preimage. Four invalid blocks are accepted in main as a direct result. Confirmed by an executed spec run: the trace reaches `witness_state.py:212` and raises, and `verify_stateless_new_payload` turns that into a clean `successful_validation=False`.

Note the shape: **the spec's rejection is *implicit* in a raise, so it is easy to compile into nothing at all.** An explicit `if bad: return INVALID` is hard to drop by accident. A bare `self._code_db[code_hash]` is trivially "translated" into a lookup that returns a miss.

## The census

Every `raise` in `src/ethereum/forks/amsterdam/` is a rejection point. For each, ask what our guest does at the corresponding site:

- **carried over** — we reject (a status code that reaches a verdict rejection);
- **differs for a named constraint** — acceptable only for *no allocator* / *no exceptions*, and the latter means the raise became a **reject status**, not a value;
- **invented or omitted** — we return a value, or never reach the site at all. **Defect, regardless of whether any fixture exposes it.**

Two sites already known to matter, both in the witness resolver:

| spec site | raises | our behaviour |
|---|---|---|
| `witness_state.py:212` — bare `_code_db[code_hash]` | `KeyError` on missing preimage | tx-map miss path → absent/unavailable, **do not reject** (#11508) |
[...]

PR #11527, opened 2026-08-05
feat(ci): #11520 code_at_header EMPTY_CODE_HASH guard

pirapira:
## Summary
CI guard for #11520: missing code preimage must **reject**, not fall back to empty (`witness_state.py:204-212`).

### Checks (string-level, branch-aware)
1. **pair** — `jal code_at_header` without `*empty_code_hash` in the def
2. **anti** — status-5 `beq` to empty without prior hash load
3. **conv** — hash compare present but **every `bne` mismatch target equals the match `j`** (discarded comparison = false assurance)

### Why conv exists
SenderCounts:248 was reported “correct” under contains-pairing. Control flow:

```
bne … .Leas_we_ok   ×4 (mismatch → ok)
j .Leas_we_ok         (match → ok)
```

All arms converge → comparison gates nothing. A contains-test cannot see this.

### Branch-target re-check of the two real correct sites
| site | mismatch bne | match j | verdict |
|---|---|---|---|
| ChildFrameHandlers:859-878 | `.Lcd_fail_` | `.Lcd_empty_` | **OK** |
| DispatchTx materialize:123-132 | `.Ldtrc_materialize_lookup_done` | `.Ldtrc_materialize_empty_target` | **OK** |
| SenderCounts:248 | `.Leas_we_ok` | `.Leas_we_ok` | **conv violator** |

### Ratchet
`EXPECTED_VIOLATOR_COUNT = **8**` (was 7) + allow-list. Drive to 0 with site fixes.

### OUT OF SCOPE
exists/is_empty no-charge family → #11526.

### dual-accept
Same rule: transitional accept needs a drive-to-zero count (~203 bare B-imm files). Not touching #11524.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,6 +54,8 @@ jobs:
           scripts/check-registry-crosscheck.sh
       - name: Layering invariants (verified core must not import Codegen / Progress)
         run: scripts/check-layering.sh
+      - name: "code_at_header empty-hash guard (#11520; raise→reject, not fallback)"
+        run: scripts/check-code-preimage-empty-hash.sh
       - name: Opcode-structure (AddrNorm pairing; advisory checklist for new complex dirs)
         run: scripts/check-opcode-structure.sh
       # Advisory naming nudge (camelCase hypotheses newly added in the PR);
--- a/scripts/check-code-preimage-empty-hash.py
+++ b/scripts/check-code-preimage-empty-hash.py
@@ -1,0 +1,246 @@
+#!/usr/bin/env python3
+"""#11520 guard: code_at_header_state_root must not compile a missing preimage
+into empty code without an EMPTY_CODE_HASH identity check.
+
+Spec (pin e5a8caf1b witness_state.py:204-212):
+  * EMPTY_CODE_HASH → return b"" (legitimately empty)
+  * any other missing hash → raise KeyError → REJECTION
+A guest site that treats status-5 (code hash present on account, preimage
+absent from codes) as empty WITHOUT comparing cahsr_acct_struct.code_hash to
+EMPTY_CODE_HASH has compiled a raise into a fallback value.
+
+Two correct sites already exist (ChildFrameHandlers status-5 + cd_empty_code_hash;
+BlockVerdictDispatchTx materialize path + chahsr_empty_code_hash). This gate
+makes a regression uncompilable-to-CI rather than merely reviewable.
+
+Checks (Lean source of Codegen Programs / Dispatch — no guest emit required):
+
+  1. FUNCTION PAIRING: any non-probe `def` whose body contains
+     `jal ra, code_at_header_state_root` must also contain an
+     `*empty_code_hash` load (`la …, …empty_code_hash`). Missing → violator.
+  2. STATUS-5 ANTI-PATTERN: after a `jal ra, code_at_header_state_root`, a
+     `li r, 5` followed by `beq a0, r, …empty…` (or swap) WITHOUT an intervening
+     `*empty_code_hash` load → violator (even if the same function has a
+     correct site elsewhere — DispatchTx has both).
+
+Allow-list: scripts/code-preimage-empty-hash-allow.txt
+  lines: kind\\tfile\\tdef\\tordinal   kind in {pair, anti}
+EXPECTED_VIOLATOR_COUNT must match the allow-list length and ratchet DOWN only
+with an explicit edit in the same commit that fixes a site (same durability
+rule as EXPECTED_ANNOTATION_COUNT / #11505).
+
+OUT OF SCOPE (different family — do not count as covered by this gate):
+  account_exists_at_header_state_root / account_is_empty_at_header_state_root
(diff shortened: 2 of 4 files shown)


pirapira:
Reviewed. **The ratchet structure is right; the classification is wrong for one site, and it exposes a gap in what the check can prove.** Not labelling until that is addressed.

## What's right

Two detection kinds rather than one, which is more than I asked for:

- `pair` — function has `jal code_at_header` **without** `*empty_code_hash`
- `anti` — status-5 `beq`-to-empty **without** an `EMPTY_CODE_HASH` compare

And the ratchet is enforced on **both sides**: `len(allow) == EXPECTED_VIOLATOR_COUNT` *and* `len(viol) == EXPECTED_VIOLATOR_COUNT`, so the list and the constant cannot drift apart. Plus the header rule — *"Ratchet DOWN only with a same-commit site fix. Never raise without coord."* That is the `EXPECTED_ANNOTATION_COUNT` discipline applied correctly, and it is what stops a transitional allow-list becoming permanent permission.

## ⛔ The gap: a comparison that exists but does not gate

`BlockVerdictSenderCounts.lean` is **not** in the allow-list and was reported as a correct site. It is not correct. `:248`, read as control flow:

```
la t0, cahsr_acct_struct; addi t0, t0, 72; la t1, chahsr_empty_code_hash
ld/ld; bne t2, t3, .Leas_we_ok     ← limb 0 mismatch → ok
ld/ld; bne t2, t3, .Leas_we_ok     ← limb 1 mismatch → ok
ld/ld; bne t2, t3, .Leas_we_ok     ← limb 2 mismatch → ok
ld/ld; bne t2, t3, .Leas_we_ok     ← limb 3 mismatch → ok
j .Leas_we_ok                      ← all four match → ok
```

**Every outcome reaches the same label.** Four 8-byte comparisons against `chahsr_empty_code_hash` are computed and discarded. `:246` completes the picture: `beqz a0 → .Leas_we_code`, status 1 → ok, status 5 → falls into `:248`, **anything else → ok**. So statuses 2, 3 and 4 never reject either — **this function never rejects on any status.**

It escapes both kinds by construction: `pair` requires the symbol to be *absent* (it is present); `anti` requires the compare to be *missing* (it is there). So the check passes a site whose comparison is inert.

⇒ **A string-contains pairing proves the comparison exists, not that it gates.**

## The refinement, still string-level and still cheap

Assert that **the mismatch branch target and the match path differ.** A pairing whose `bne` target equals the fall-through label is a discarded comparison and must fail. That converts *exists-a-comparison* into *acts-on-a-comparison*, and it is the difference between a gate and a decoration.

Then: **`EXPECTED_VIOLATOR_COUNT = 8`**, with `BlockVerdictSenderCounts` moved from correct to violator.

⚠️ And re-check the other two sites classified as correct — `ChildFrameHandlers.lean:859-878` and `BlockVerdictDispatchTx.lean:123-132` — **by branch target rather than symbol presence.** If either converges the same way, the count moves again, and the "two sites already do this correctly" premise that justified guarding rather than migrating weakens.

## Separately: `:248` is a defect in its own right

A comparison whose arms converge is **worse than no comparison**, because it reads as protection both in review and to any contains-based gate. It is also the exact shape flagged on #11520 for this site — the comment asserts a canonical-empty-only precondition that the branch structure does not enforce. Worth its own entry there with the convergence quoted.

## Scope note, correctly stated in the PR

The `account_exists_at_header_state_root` / `account_is_empty_at_header_state_root` no-charge family is **out of scope** and marked so — those return nonzero on MPT/decode/header failure and the caller treats it as no charge, which no empty-hash pairing covers. #11526 tracks one of them and is **FA-directional**, since undercharging lets a block that should fail on gas pass.

pirapira:
**Labelled `merge!`** — tip `0849b7b0a3`, all six checks green.

The convergence refinement is in and it is what makes this a gate rather than a decoration:

```python
return "ok" if len(set(bnes_n)) > 1 else "conv"
```

Distinct branch targets required, not symbol presence. `EXPECTED_VIOLATOR_COUNT = 8`, `BlockVerdictSenderCounts` moved from correct to violator, and `SenderCounts:248` named in the docstring as the proof case.

⭐ **And the check I most wanted came back favourably.** Re-verifying the two "correct" sites *by branch target* rather than by symbol — `ChildFrameHandlers` (fail ≠ empty) and `BlockVerdictDispatchTx` (`lookup_done` ≠ `empty_target`) — both genuinely gate. That matters because those two sites are the entire premise for guarding rather than migrating: if either had converged like SenderCounts, there would have been no correct in-tree pattern to guard against and this design would have collapsed. It survived the test that could have killed it.

The ratchet discipline is right on both sides — `len(allow) == EXPECTED_VIOLATOR_COUNT` **and** `len(viol) == EXPECTED_VIOLATOR_COUNT`, so list and constant cannot drift, plus *"Ratchet DOWN only with a same-commit site fix. Never raise without coord."*

Scope correctly excluded: the `account_exists_at_header_state_root` / `account_is_empty_at_header_state_root` no-charge family, which no empty-hash pairing covers. #11526 tracks one of those and is FA-directional.

Issue #11508, opened 2026-08-05
FA is 4 (not 7): four witness-missing fixtures accepted at 6ebe3869c; plus 3 guest crashes misclassified as FA by the sweep


pirapira:
## FA=7 in main — seven invalid blocks are being ACCEPTED

Found by a full-corpus sweep on merged main (k3), and confirmed as a **regression within the window** rather than a pre-existing condition (grok).

### The measurement

Full corpus, 26,104 rows, Spike, on main `d6047135f`: **OK 23,764 · FAULT 1,752 · FR 581 · FA 7.**

FA=0 has been asserted in every status summary for some time, including all of mine tonight. That figure came from a baseline whose base is `020802bd7` (Merge PR #11331) — which **predates #11333** and therefore reflects none of the work merged since. It was re-cited rather than re-measured. Running it is what surfaced this.

### The seven, all `oracle=0 guest_succ=1`

| id | fixture |
|---|---|
| 00349 | `test_bal_gas_limit_boundary` · below_boundary-no_tx-no_cl_withdrawal |
| 00350 | `test_bal_gas_limit_boundary` · below_boundary-no_tx-with_cl_withdrawal |
| 02965 | `test_validation_codes_missing_delegated_code_on_insufficient_balance_call` |
| 02967 | `test_validation_codes_missing_implicit_system_contract_code` |
| 02970 | `test_validation_codes_missing_sender_delegation_marker` |
| 02998 | `test_validation_state_missing_failed_call_target_account_proof_node` |
| 26003 | `test_fork_transition_bal_size_constraint` |

**Four of the seven are `*_missing_*` negative fixtures** — witness-incompleteness cases that must be rejected.

### Confirmed introduced in the window, not pre-existing

Side-by-side, same inputs, `guest_succ = out[32]`, oracle from manifest col 4 (`=0` for all seven):

| | `020802bd7` | `d6047135f` |
|---|---|---|
| guest sha256 | `33f7ce9adaf28a30…` | `bf7f98312e76052e…` |
| `.text` | 405,024 | 363,992 |
| all seven | **succ=0, correctly REJECT** | **succ=1, FA** |

A third ELF (k3's sweep image, sha `75d3c49da29a7185…`) is distinct from the rebuild yet agrees on all seven verdicts — which is what makes the comparison trustworthy rather than an artefact of one build.

### Next step, and why it is this one

[...]

PR #11528, opened 2026-08-05
fix(codegen): #11526 reject malformed code_at in delegation access charge

pirapira:
## Summary
Fixes FA-direction undercharge on nested CALL-family EIP-7702 delegation access gas.

**Site only:** `ChildFrameHandlerTailHelpers.lean` `callDelegationAccessChargeAsm` (not k3’s `ChildFrameHandlers.lean` / `.Lcd_insuffbal_`).

### Spec
Amsterdam `eoa_delegation.py:146-162`: `get_code(tx_state, account.code_hash)` then charge WARM/COLD for delegated address. `get_code` raises on missing non-empty preimage (`witness_state.py:204-212`) → **reject**, not “not delegated”.

### Status direction (before write)
| status | meaning | direction |
|---|---|---|
| 0 | code found | inspect 0xef0100 marker; charge if delegated |
| 1 | account absent | not delegated → **no charge** (done) |
| 2/3/4 | MPT / decode / header malform | **reject** `.Lcd_fail_` |
| 5 + EMPTY_CODE_HASH | legitimate empty EOA | not a marker → **no charge** |
| 5 + other hash | missing preimage raise | **reject** `.Lcd_fail_` |

### Branch targets (11527 lesson)
- done = `.Lcdac_done_<tag>`
- fail = `.Lcd_fail_<tag>`
- mismatch bne → fail; match j → done — **not convergent**

### Blast radius
One helper behind four CALL-family handlers (`precompileMessageProcessorAsm` tags call/callcode/delegatecall/staticcall). Top-level `delegationAccessChargeAsm` (DispatchTx) untouched.

### Layout
textSizeBytes **0x55d94**; region-map CLEAN; class-a OK 4.

### Not done
No merge! from grok. 11527/11524 untouched.


pirapira:
Reviewed on content — the status split is right and the spec citation is the correct one. **Two things needed before I label**, and the second is a cross-PR interaction that neither this PR nor #11508 would catch on its own.

## What's right

The spec basis is exactly the relevant pair: `eoa_delegation.py:146-162` reads the account's own code then returns `WARM_ACCESS`/`COLD_ACCOUNT_ACCESS` for the caller to charge, and `get_code` (`witness_state.py:204-212`) **raises** on a missing non-empty preimage. So a raise compiles to a **rejection**, not to "not delegated" — which is the #11520 rule applied correctly.

And the status split respects the distinction that matters: **status 1 is a genuinely absent account and correctly charges nothing**, while 2/3/4 are malformed and 5-with-a-non-empty-hash is a missing preimage. Treating those as "not delegated" was the defect; the previous code branched on *any* nonzero status.

⭐ The #11527 lesson is applied — `done ≠ fail` targets, so this is a gate rather than a converged comparison. Worth noting that was learned the hard way on `SenderCounts:248` about twenty minutes ago and carried straight into a different file.

## ⛔ 1. Flip set required

**This PR converts statuses 2/3/4 and 5-mismatch from "no charge, proceed" into "reject."** That adds rejections, so the risk is false rejects: any row that was passing and now rejects.

Needed: **base-PASS → candidate-REJECT, named and oracle-checked, zero unless each is genuinely invalid.** The blast radius makes this non-trivial rather than pro-forma — the shared nested tail serves **all four CALL-family handlers** (`ChildFrameHandlers.lean:122-142`), so this is one helper behind four opcodes.

⚠️ Report it in raw `(rc, succ, bv_fail)` rather than FA/FR categories. And note `rc=3` is **overloaded**: 314 rows on current main are step-cap timeouts (`spike_run: step cap reached without halt`, `HALT_FLAG=0`, no trap signature), distinct from the fault class #11511 eliminated. A step-cap row is not a verdict either way.

## ⛔ 2. The interaction with #11508

codex is landing the delegation **read** restoration in `ChildFrameHandlers.lean` right now. **Both PRs add rejections on the same path**: #11508 restores a read so the #11426 preimage gate fires where it didn't, and this one rejects on malformed statuses where it previously proceeded.

So each could show a clean flip set alone while **jointly** rejecting rows neither would predict. Whichever lands second must re-run its flip set against the other's merged state — and if both land close together, the joint set is the one that matters. I would rather sequence them than discover the interaction from a corpus regression.

## One design question

Statuses 2/3/4 are MPT-parse, account-decode and header-parse failures. Rejecting is right per the raise rule — but **is the charge helper the right place to detect them?**

If a malformed status already causes a rejection at the lookup, this becomes belt-and-braces and the rejection is arguably fixed at the wrong layer (the pattern being *fix at the source, not at a downstream consumer*). If it does **not** reject elsewhere, then this helper is the only detector and the placement is correct — which is also a finding, because it would mean a malformed witness currently reaches a gas computation without anything upstream objecting.

Say which. It does not necessarily change the patch, but it changes whether #11520's remaining sites need the same treatment or a shared upstream fix.

pirapira:
**Labelled `merge!`** — the evidence I asked for is in, from an independent build and harness rather than from this PR's author.

## The flip set, measured four ways

grok built four guests from base `bbbf51ad6` — base, #11508-alone, this PR alone, and joint — Spike runner `62c3fbaf…`, all rows `rc=0` with no faults or step-cap:

| guest | 02965 | 02967 | 02970 | 02998 |
|---|---|---|---|---|
| base | FA | REJ `bv=11` | FA | FA |
| #11508 alone | FA | REJ | FA | FA |
| **this PR alone** | FA | REJ | FA | **REJ `bv=1`** |
| joint | FA | REJ | FA | **REJ** |

**base-PASS → candidate-REJECT: `{02998}` only, and `02998` is oracle-invalid.** So the single flip is a **false accept being closed**, not a valid block being rejected. Zero regressions on valid rows.

## ⭐ It closes a false accept, and not the one anyone expected

I predicted `02998` would *not* close via #11508's code-read restoration, because `missing_failed_call_target_account_proof_node` fails on a missing **proof node** — `AssertionError` at `witness_state.py:71-74` — rather than the `KeyError` at `:212`. That held: #11508 alone leaves it FA.

**This PR closes it instead, and `bv_fail` proves which mechanism did it: `1` (charge reject), not `11` (preimage gate).** So the status split on the *charge* path catches a case the read path cannot reach. Without the `bv_fail` distinction that closure would have been misattributed.

## The interaction concern is retired on evidence

I held this PR partly because it and #11508 both add rejections to the same path, so each could show a clean flip set alone while **jointly** rejecting rows neither predicts. Measured: **joint flips exactly what this PR flips — no extra interaction flip.** That is now settled by measurement rather than by argument, and it is why the four-guest design was worth the extra two builds.

## What was right in the patch

The spec basis holds — `eoa_delegation.py:146-162` returns `WARM_ACCESS`/`COLD_ACCOUNT_ACCESS` for the caller to charge, and `get_code` **raises** on a missing non-empty preimage, so a raise compiles to a rejection rather than to "not delegated." Status 1 (genuinely absent) correctly charges nothing; 2/3/4 and 5-with-non-empty-hash reject. And `done ≠ fail` targets, so this is a gate rather than a converged comparison — the `SenderCounts:248` lesson carried into a different file within the hour.

⚠️ Still open on the design question I raised: whether the charge helper is the right *place* to detect malformed statuses, or whether they should already have been rejected upstream. That does not block this — the fix is correct where it sits and it closes a real FA — but it bears on whether #11520's remaining sites each need the same treatment or a shared upstream fix.

PR #11539, opened 2026-08-05
fix(codegen): split bv_fail 40 into distinct MTx sender-block codes

pirapira:
## Summary
Split catch-all `.Lbv_sender_nonce_fail` (bv_fail **40**) into distinct live MTx sender-block codes.

## Map (shipped)
| code | label | cause |
|---:|---|---|
| **40** | sender_nonce | tx.nonce mismatch **kept** |
| 68 | sender_count_table | `b1_sender_count_table` |
| 69 | sender_resolve | `account_resolve_pre_state` |
| 70 | sender_not_eoa | **reserved** sink; wire after #11533 InvalidSender |
| 71 | sender_inclusion | `account_state_publish_sender_inclusion` |
| 72 | auth_prepare | `block_verdict_tx_state_gas_inline_prepare` |

**Dropped 73–76** (coord): frozen S1 deleted by #11536 (merge! ahead). Those causes stay on code 40 until 11536 lands, then the block is gone — no forever-silent codes in the map.

## ReceiptsTail (+lines)
All `.Lbv_*_fail` **bodies** live in `BlockVerdictReceiptsTail.lean` (emit `li t0,N`). MtxRuntime only holds branch targets. Not ride-along.

## `#guard` per code
Each shipped code-emitting sequence appears exactly once in the assembled string (build fails on dup/missing).

## Emitter
`BlockVerdict.lean` +8 `bv_fail_code` **offset map unchanged** (values only).

## Layout
textSizeBytes **0x55c94** (after drop 73–76). No rebase (layout convoy with 11533/11536).

## r200 + value-change evidence
spike `951294739be258d78f0cb3c5f93ed81b38f7007a89483f6babff6d13a6fecc52`

| leg | guest sha256 |
| base | `3f105b079b9e94c7…` @ c225747dc |
| cand | `05f26574e3a8a788…` @ 4183a545f |

**Same 200 labels** (seed lineage 11533/11536 r200):
- accept/reject flip set: **∅ both directions** (200/200 same; 3 identical rc=3 both legs)
- bv_fail value changes on that 200: **∅** (sample does not hit former-40 causes — not a success by itself)

**Extended hunt** (3000 labels outside r200; base bv==40):
| n | base→cand | succ |
| 8 | **40 → 69** (sender_resolve) | 0→0 reject both |

Artifacts: `/tmp/opencode/11539-r200/` (`r200-raw.tsv`, `hits40-ab.tsv`).

Code-split signature: AR empty + value-change non-empty (on exercised rows).


pirapira:
## Honest limit (coord pre-merge note)
Only **69** (sender_resolve) is behaviourally exercised in the 3000-row hunt (8× 40→69, reject→reject).

**68, 71, 72** unexercised in that sample. **70** was unreachable at PR base (belongs to #11533, now on main via 11538 batch) — reachable on merged tree.

Five of six codes are latent in the corpus; verified structurally by once-per-code `#guard`, not behaviourally. Fine for diagnostic **labels** (not branches): an unexercised label cannot accept invalid or reject valid — only fail to explain a rejection that already happens.

PR #11540, opened 2026-08-05
refactor(codegen): name long local J targets across 26 files

pirapira:
## Summary

Convert 26 generated Codegen files from bare long local `j` offsets to named `jalOff` targets, preserving the existing emitted bytes while making same-function control-flow targets resilient to body edits.

The gas-gate conversion is intentionally excluded from this PR and deferred to a later batch after #11537 reaches `main`: its renderer must emit the corrected `+256` target (not the stale `+232` form).

## Ratchet

`EXPECTED_BARE_J_SITES = 181` is an exact, blocking ratchet. The current direct census is 49 files / 81 definitions / 181 bare local-J sites. Any intentional decrease or increase must update the committed constant in the same commit with a stated reason; file and definition counts are report-only diagnostics.

## Evidence

- `git diff origin/main -- EvmAsm/Codegen/Programs/BlockVerdictGasGate.lean` is empty; the gas gate remains deferred as stated above.
- The direct census reports `(49, 81, 181)`.
- Converted modules were checked with the repository generator checks during the conversion batches.
- No merge label is being applied by this author.

The conversion is intended to be byte-preserving; CI will run the repository build and generated-artifact checks on the merged result.


pirapira:
Ratchet arithmetic (using the same current counter implementation on both trees): origin/main c225747d reports 236 bare local-J sites; this PR converts 55 sites, yielding 236 - 55 = 181. The committed equality baseline is therefore an explicit arithmetic claim, not only an endpoint census. The later gas-gate batch is intentionally deferred behind #11537 and does not change the bare-site form/count.

Issue #11547, opened 2026-08-05
fixture: intrinsic_gas_cost signers_2 BAL under-declares post-recover auth read (24498 FR)


pirapira:
## Summary
EEST fixture `intrinsic_gas_cost` / `valid_True-single_invalid_authorization_eoa_authority_multiple_signers_2` declares a BAL that **omits** an `account_reads` empty shell the Amsterdam pin **requires**. Guest matching pin FR vs fixture (code-60, rebuilt 359 vs supplied 332). **Not a guest defect.**

## Pin (e5a8caf1b amsterdam)
`eoa_delegation.py` `validate_authorization`:
- :176-185 chain / nonce-max / recover fail → `None` **before** `get_account`
- :187-189 `accessed_addresses.add` + **`get_account`** (records `account_reads`)
- :196-201 code / nonce fail may still `None` — **read already recorded**

## Sibling asymmetry (same rule, nonces swapped)

| | signers_1 BAL n=10 | signers_2 BAL n=9 (24498) |
|---|---|---|
| auth[0] 4701 | nonce 0 vs pre 1 → **fail :200** after get_account | nonce 1 vs pre 1 → pass |
| auth[1] b2ef | nonce 1 vs pre 1 → pass | nonce 0 vs pre 1 → **fail :200** |
| b2ef in BAL | yes (empty shell) | **no** |
| 4701 in BAL | yes (empty shell) | yes (empty shell) |

**signers_1 4701 confound ruled out:** not sender (`f6c3…`), not to (`37f5…`), not coinbase (`2adc…`), accessList empty, value 0. BAL entry = empty AccountChanges — only auth `get_account` fits.

Fixtures **disagree** on whether post-recover nonce-fail keeps a read. Pin and signers_1 say **yes**; signers_2 omits b2ef.

## Guest
`eip7702_authority_asof` records after recover (matches :189). Dropping that read to match signers_2 = **FA direction** (rebuilt missing a pin-required entry can hash-match invalid supplied BAL).

## Manifest
- label: `24498_test_intrinsic_gas_cost_…valid_True-single_invalid`
- oracle=1, guest succ=0 bv=60, rebuilt=359 supplied=332 (+27)

## Ask
Fix/regenerate fixture BAL (or document intentional filler divergence). Do **not** change guest to under-record.

Refs: code-60 survivor after #11545/#11546; coord thread 24498.

pirapira:
[...]

PR #11553, opened 2026-08-05
fix(codegen): two independent code-1 defects in 4788/2935 MTx system seed

pirapira:
## Summary

Two **independent** defects in the MTx system-storage seed path; either alone kept the EIP-4788 / code-1 beacon CALL family failing. Named separately so a reader does not credit one mechanism for 96 rows.

### Defect A — swapped ABI into `account_at_header_state_root` (call-site)

**File:** `BlockVerdictMtxRuntime.lean` (`blockVerdictMtxGateSystemStorageSeed`)

Callee ABI is `a0/a1 = header RLP`, `a2/a3 = address`. The gate had `a0=addr`, `a2=header`. Callee returned **status=4** (header parse fail). Gate treated that as absent → zeroed all `swd_*_vlen` → seed no-op → SLOAD never saw 2935/4788 block-start writes → 4788 CALL returndata 0.

**Fix:** `ld a0,8(s0); ld a1,16(s0); la a2,bsr_addr_*; li a3,20` (both 2935 and 4788 sites).

### Defect B — short-vlen value cell not zeroed (uninitialised read)

**File:** `BlockVerdictSystemStorageCapture.lean` (`append_modeled_system_storage_tuple_rows`)

`value_rev` wrote only `vlen` bytes of a minimal-BE payload into a **reused** 32B LE cell. Timestamp `0x0c` (`vlen=1`) left stale high bytes from the previous row. After A fixed the seed, SLOAD still returned a non-scalar word → 4788 timestamp EQ failed → CALL returndata still 0.

**Fix:** `sd zero, 96/104/112/120(s2)` before the reverse.

Spike dump after A-only: row1 value = `0cbf1a54dd922df7…` (byte0=0x0c, rest garbage). After B: clean scalar.

### 1. Caller census — complete negative (emitted image)

Enumerated **all 16** `jal account_at_header_state_root{,_tracked}` sites in the linked guest image (`census.s` / #11553 ELF). **0 of 16 swapped** after Defect A. Strong form: full enumeration, not a sample.

| # | Site (emitted) | a0 | a2 | verdict |
|---|---|---|---|---|
| 0 | MtxRuntime seed 2935 | `ld a0,8(s0)` header | `la a2,bsr_addr_2935` | OK (was swapped; fixed A) |
| 1 | MtxRuntime seed 4788 | `ld a0,8(s0)` header | `la a2,bsr_addr_4788` | OK (was swapped; fixed A) |
| 2 | MtxTail B2 sender | `ld a0,8(s0)` header | `mv a2,t4` sender | OK |
| 3 | Creation prebalance | `ld a0` parent_rlp | `la a2,bv_create_addr` | OK |
| 4 | DispatchTx selfbalance | `ld a0` parent_rlp | `addi a2,s2,72` recipient | OK |
| 5 | Account builder pre | `ld a0` sv_pre_rlp | `mv a2,s4` | OK |
| 6–7 | BalCodePreimages `_tracked` ×2 | `mv a0,s2` header | `mv a2,s0` addr | OK |
| 8 | TxIntrinsicStateGas teer | `ld a0` sv_pre_rlp | `mv a2,s0` | OK |
| 9 | Account resolver pre | `ld a0` sv_pre_rlp | `mv a2,s0` | OK |
| 10 | CreateFrameDescend | `ld a0,576(s9)` header | `addi a2,s9,696` | OK |
| 11 | Create address prebalance | `ld a0,576(x20)` | `la a2,create_address_be` | OK |
| 12 | Selfdestruct origin | `ld a0,576(x20)` | `la a2,sdai_origin_address` | OK |
| 13 | Selfdestruct beneficiary | `ld a0,576(x20)` | `la a2,evm_selfdestruct_beneficiary` | OK |
| 14–15 | NSE callee ×2 | `ld a0,576(x20)` | `la a2,nse_callee_be` | OK |

**Scope limit (say plainly):** this covers the **sixteen sites that exist today**, not a seventeenth added next week. A swapped-ABI call is invisible to every gate we have — it compiles, links, runs, and the callee returns a plausible status the caller treats as data.

**Cheap gate?** Yes, cheap enough to recommend later (not in this PR): a `scripts/check-account-at-header-abi.sh` that greps the emitted `.s` for each `jal account_at_header_state_root` and asserts the last `a0` assignment is a header/rlp load and the last `a2` is an address pointer (`bsr_addr_*` / `*_be` / frame+offset). Runtime “a0 does not look like a 20-byte address” is **not** useful — both args are *pointers*, and `a3=20` is correct even when swapped. Static emitted-image check is the right shape (~30–50 lines Python, blocking). Leave unlanded until after merge unless coord wants it in-tree now.

### 2. Short-vlen / stale-tail census (order-dependent class)

Pattern: write a **vlen-bounded** minimal payload into a **fixed-width reused** cell without clearing the tail. No single-row test finds it; only enumeration.

| Site | Shape | Zero-before-write? | Consumer uses only vlen? | Status |
|---|---|---|---|---|
| `append_modeled…value_rev` (s2+96) | vlen bytes → 32B LE, then upsert copies **full 32** | **YES after B** | No (upsert 32B) | **LIVE fixed** |
| `swd_minimal_copy` → `swd_*_val` BSS | stripped minimal BE → 32B BSS | **NO** | Mixed: SysChange/`rlp_encode_bytes` uses vlen; `system_write_pack` copies **full 32** of val | **LATENT** — file issue |
| `SeedTxAccessList` slot/token reverse | klen ≤32 into 32B | YES (`sd zero`×4) | n/a | SAFE |
| `Storage` prestate_value_rev | always 32B reverse | n/a (full width) | full 32 | SAFE |
| `baap` map_value_rev | always 32B reverse | n/a | full 32 | SAFE |
| EIP7708 amount_rev | always 32B | n/a | full 32 | SAFE |
| Creation callvalue/selfbalance rev | always 32B | n/a | full 32 | SAFE |

**Live site fixed by B.** One latent sibling: `swd_minimal_copy` (see filed issue). Class is real; order-dependent; not blocking #11553.

## Measure

| Artefact | sha256 |
|---|---|
| base ELF (main @ `591b07acf`) | `5700ba4e03af7ee066822f0aaaf5744e3e75e6c2cb7dfb30d0956e4ee8fc11f9` |
| cand ELF (this PR @ `4261282b8`) | `3eb162a3215b5f0f6d283d75e20da58230125739da5d29352f4b4ba029785841` |

**Beacon-root family (105 fixtures):** F→P=**96** P→F=**0** cand_bv1=**0** — **family closed**. No residual true-absent in this family after both fixes.

**r200** (`seed=20260805`, n=200 from `/var/tmp/fc668`):
- base OK=189 FR=11 → cand OK=190 FR=10
- F→P=**1** (`05870` multi_block beacon) P→F=**0**
- same_FR residual (10, unrelated to this path): `01315, 01532, 01618, 04775, 14283, 21155, 23351, 23428, 24832, 26030`

**Named:** `05814`/`05862`/`05823` PASS; `05813` stays PASS. Do **not** cite `out[0:32]` as state root.

Layout pins regenerated to fixed point (`textSizeBytes` `0x55bc4`→`0x55bd4`).

Refs: #11547 (live), #11326 structural. **Do not merge from agent.**

## Test plan
- [x] lake codegen + layout regen fixed point
- [x] beacon 105 F→P=96 P→F=0
- [x] r200 seed=20260805 matched pair
- [x] emitted-image caller census (16/16 table above)
- [x] short-vlen class census (1 live fixed + 1 latent)


pirapira:
## Stale-tail census (complete negative for LIVE class)

**Class:** a vlen-bounded or short write into a fixed-width reused 32-byte cell without clearing the tail first. Order-dependent; fresh BSS masks it — the defect only appears on reuse, so no sweep and no single-row test can be trusted to have exercised it; enumeration is the only instrument.

**LIVE sites besides `append_modeled` value_rev (#11553 Defect B): 0**

### LIVE (fixed)
| Site | File | Status |
|---|---|---|
| `append_modeled_system_storage_tuple_rows` value_rev | `BlockVerdictSystemStorageCapture.lean` | LIVE→FIXED #11553 B (`sd zero 96/104/112/120(s2)` before reverse). Short ts `0x0c` left high garbage → 4788 EQ fail |

### LATENT (not live on guest path today; already tracked)
| Site | File | Why latent |
|---|---|---|
| `swd_minimal_copy` → `swd_2935_val` / `swd_4788_val` / `swd_4788_root_val` | `SystemWrites.lean` | Writes stripped length only; no dest zero. BSS reused across `system_write_descriptors`. Guest `append_modeled` now zeros before reverse and uses vlen → safe on guest path. Probe pack (`.Lswp_c2/c4/c6`) copies full 32 from `swd_*_val` and would propagate garbage if descriptors re-run after a longer write. Tracked as **#11554** (open). No additional latent sites. |

### SAFE (always full-32 reverse, or zero-before-short)
| Site | Why SAFE |
|---|---|
| Storage `prestate_value_rev` | always `li x16, 32`; zero path zeros 64B |
| baap `map_value_rev` | always 32; strip is read-side for RLP |
| EIP7708 `amount_rev` | always `li t3, 32` |
| Creation `callvalue_rev` / `selfbalance_rev` | always 32 |
| DispatchTx `gp_rev` / `selfbal_rev` / `selfbal_live_rev` | always 32 into env words |
| NoopHalt `crp_rev` / `sd_create_creator_post_rev` | always 32 |
| StorageWriteMap value store | full ld/sd×4; baseline `sd zero 96..120` |
| SeedTxAccessList `token` / `token_le` / `slot_le` | `sd zero×4` **before** short reverse (gold pattern) |

### Out of class
Length-bounded copies into staging/path buffers (SystemCallStaging wcopy/ccopy, MptIndexedTrieRoot, deposit pubkey/amount/sig, BAL builder arev/srev, addr copies): dest is length-tracked or freshly sized, not “short write into reused 32B value cell.”

### Caveat
This census covers the sites that exist **today**, not the one somebody adds next week. Same expiry as the #11553 caller-ABI census. A complete negative LIVE enumeration means the class was a one-off at `append_modeled` value_rev; the only related latent is #11554.

Recommended #11554 fix when picked: zero 32B dest at start of `swd_minimal_copy` (or before each call site) — SeedTxAccessList / #11553 B pattern.

pirapira:
## Both censuses: complete negatives (as of today)

Coord closed the gate question: **do not build** `scripts/check-account-at-header-abi.sh`. Two independent complete negatives mean each defect was a one-off, not a live class worth a maintenance-costly heuristic gate (and we already have one unreliable gate under cache mode — #11557).

| Census | Result | Durable home |
|---|---|---|
| Caller ABI into `account_at_header_state_root[_tracked]` | **16/16 correct** after Defect A; no second swap | PR body table |
| Stale-tail (short write into reused 32B without zero) | **0 LIVE** besides Defect B (`append_modeled` value_rev); only latent #11554 | [comment above](https://github.com/Verified-zkEVM/evm-asm/pull/11553#issuecomment-5196863116) |

**Expiry caveat (both):** covers sites that exist **today**, not sites added later. Negative enumeration is still a durable fact for the current image; it is not a proof about next week’s seventeenth callsite or next short-write helper.

PR #11592, opened 2026-08-06
fix(codegen): align MTx coinbase account read

pirapira:
## Scope

Batch A group 3 only: the MTx coinbase fee-credit producer now records the beneficiary account read at the execution lookup site. Amsterdam `fork.py:1172` calls `create_ether(tx_state, coinbase, transaction_fee)`; `state_tracker.py:680-700` resolves that beneficiary through `get_account` and mutates its balance. The guest records the beneficiary immediately before its authenticated `balance_at_header_state_root` lookup. This does not add a resolver to a publication-only path.

The Stage-2 screen remains the scope boundary: groups 11 and 12 have no independent spec lookup and are already spec-shaped; group 4's zero-value/self arm is excluded because the spec only calls `move_ether` when `should_transfer_value && value != 0`; true full-account resolution remains only the CREATE code-deposit consumer (`CreateCodeEffectLog.lean`) and is not widened here. The payoff is a spec-shaped producer/read boundary and a simpler per-producer proof obligation, not uniformity for its own sake.

## Prediction

Against base `ae45b8239`, before measuring: one `account_read_record` sequence adds 16 bytes of .text; .data and .bss stay unchanged; verdict outputs remain unchanged because the read set is already deduplicated by the existing wrapper/consumer path. Therefore the expected A/B flip sets are empty and that is the desired structural-alignment result.

Measured ELF sizes: base .text 350584, candidate .text 350600 (+16); .data 21424 and .bss 426579296 are unchanged.

## Evidence

Both artifacts were generated from the same source commit, with only this change:

- base commit `ae45b8239`; ELF `68d47d80e7bebfa3a4f86d4225f237ff4b2e5eda0b95ae750e945e9f0550c1e4`; generated 2026-08-06 07:29:34 UTC
- candidate commit `08f86d111` on base `ae45b8239`; ELF `69f0666a45bde3fb30c7747a15dddc5a0f0a8072062de8861fcdea32e0092434`; generated 2026-08-06 07:31:17 UTC

Random-200 A/B (seed 4242, Spike, 30 workers) used:

```
scripts/codegen-eest-stateless-check.sh --backend spike --random --seed 4242 --limit 200 --jobs 30 --guest-elf <elf> --quiet-passes --run-dir <run-dir>
python3 scripts/eest-ab-compare.py <base-run> <candidate-run>
```

Both legs: FA=0, FR=1, 199/200 exact full matches, 0 errors. The one FR label is identical on both legs; flip sets are base-pass→candidate-FR = ∅ and candidate-pass→base-FR = ∅. Joined status differences = 0 and output-byte differences = 0.

Named EIP-3651 coinbase controls (`coinbase_warm_account_call_gas`, 4 rows) were exact 4/4 on both legs, FA=0, FR=0, with both flip sets empty and byte-identical joined outputs.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdictMtxCoinbase.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictMtxCoinbase.lean
@@ -33,12 +33,6 @@ def blockVerdictMtxCoinbaseFeeEffect : String :=
   "  la t0, sv_pre_rlp_ptr; ld a0, 0(t0); la t0, sv_pre_rlp_len; ld a1, 0(t0)\n" ++
   "  la t0, bv_exec_p; ld t0, 0(t0); addi a2, t0, 32\n" ++
   "  ld a3, 80(s0); ld a4, 88(s0); la a5, bv_mtx_cbfee_pre\n" ++
-  -- Amsterdam's `get_account` records the beneficiary read before resolving
-  -- transaction writes, block writes, or the authenticated parent.  Keep the
-  -- record at this execution producer: the fee credit is `create_ether`'s
-  -- beneficiary read, not a publication-only effect.  The recorder preserves
-  -- t0-t6 and leaves a1-a5 untouched; save only the header pointer in a0.
-  "  mv t0, a0; mv a0, a2; jal ra, account_read_record; mv a0, t0\n" ++
   "  jal ra, balance_at_header_state_root\n" ++
   "  bnez a0, .Lbv_mtx_cbfee_done\n" ++
   "  la a0, bv_mtx_cbfee_pre; la a1, bv_mtx_cbfee_credit; la a2, bv_mtx_cbfee_post\n" ++
--- a/EvmAsm/Codegen/Programs/BlockVerdictMtxCoinbase.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictMtxCoinbase.lean
@@ -33,6 +33,12 @@ def blockVerdictMtxCoinbaseFeeEffect : String :=
   "  la t0, sv_pre_rlp_ptr; ld a0, 0(t0); la t0, sv_pre_rlp_len; ld a1, 0(t0)\n" ++
   "  la t0, bv_exec_p; ld t0, 0(t0); addi a2, t0, 32\n" ++
   "  ld a3, 80(s0); ld a4, 88(s0); la a5, bv_mtx_cbfee_pre\n" ++
+  -- Amsterdam's `get_account` records the beneficiary read before resolving
+  -- transaction writes, block writes, or the authenticated parent.  Keep the
+  -- record at this execution producer: the fee credit is `create_ether`'s
+  -- beneficiary read, not a publication-only effect.  The recorder preserves
+  -- t0-t6 and leaves a1-a5 untouched; save only the header pointer in a0.
+  "  mv t0, a0; mv a0, a2; jal ra, account_read_record; mv a0, t0\n" ++
   "  jal ra, balance_at_header_state_root\n" ++
   "  bnez a0, .Lbv_mtx_cbfee_done\n" ++
   "  la a0, bv_mtx_cbfee_pre; la a1, bv_mtx_cbfee_credit; la a2, bv_mtx_cbfee_post\n" ++
--- a/EvmAsm/Codegen/Programs/BlockVerdictMtxRuntime.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictMtxRuntime.lean
@@ -736,10 +736,10 @@ def blockVerdictMtxRuntimeLoop : String :=
   -- storage, account, and code reads survive a failed transaction even when
   -- its AccountState commit is bypassed.  This join follows the spec order:
   -- sender refund, coinbase fee, then incorporation.  In particular the
-  -- coinbase fee producer records the beneficiary read immediately before
-  -- its authenticated balance lookup, even when the priority-fee credit is
-  -- zero; promote that final per-transaction read at this transaction's
-  -- incorporation boundary.
+  -- existing `account_state_latest_balance` wrapper records the coinbase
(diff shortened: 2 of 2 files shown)


pirapira:
**coord review: holding this one — I believe the record it adds is already there, and that the change moves *away* from the spec's recording point rather than toward it.** Three checks, all on `origin/main` at `9d9d5ceba`, and I would like each refuted with the artefact if I have it wrong.

### 1. The read is already recorded, by the wrapper this PR's comment deletes

`blockVerdictMtxCoinbaseFeeEffect` calls `jal ra, balance_at_header_state_root` immediately after the inserted record. That routine, at `EvmOpcodes.lean:386`, calls `account_state_latest_balance` **unconditionally and live-first, before any header path** — it pads the address into `bal_addr_padded` and calls it at the step-1 site, ahead of `header_extract_state_root`. And `account_state_latest_balance` (`CreateCodeEffectLog.lean:486-490`) records the read *inside the accessor*:

> `-- GH #10619: record the account READ (unconditional, state_tracker.py:139).`
> `-- Hooked INSIDE the accessor so every caller is covered and the recording`
> `-- point mirrors the spec's (inside get_account_before_tx, not at callers).`

Both the new call and the wrapper's are unconditional at the same program point, on the same 20 address bytes. So this is the **#11551 shape inverted**: there, a census correctly found no callsite naming the coinbase and the read was nevertheless recorded by a wrapper. The comment this PR removes stated that fact, and it was load-bearing.

### 2. `account_read_record` dedups, so the addition is content-neutral

`AccountReadLog.lean:90-111`: `.Larr_scan` walks existing entries, byte-wise compares 20 bytes, and on a match jumps `.Larr_done` **without appending**. A second call for an already-recorded address changes nothing. That is consistent with the reported empty flip sets in both directions — and per the standing rule, an empty flip set on a structural change is *expected* and is not evidence either way. It does not distinguish "correctly aligned" from "redundant".

### 3. The direction of alignment looks inverted

The spec records inside `get_account`, not at callers, which is exactly what the wrapper hook implements and what its comment says. Adding a **caller-side** record is therefore a step away from the spec's shape, not toward it. If that reading is wrong, the thing that would settle it is a spec citation showing Amsterdam records the beneficiary read at the `create_ether` callsite *separately from* `get_account`.

### What would change my mind

One of these, with the artefact:

- **A stage difference.** This is the credible case and the one I would most want checked, because wrong-stage-not-wrong-value is an established defect category here. The coinbase credit sits at the incorporation boundary; if the wrapper's record lands in the *next* transaction's `tx_account_reads` while the spec puts it in this one, that is a real defect — but then the fix is **ordering relative to `read_sets_incorporate_tx`**, and this PR changes only a *comment* in `BlockVerdictMtxRuntime.lean`, leaving the ordering identical. Instrument: LENGTH → IDENTITY → ORDERING on a row with a nonzero priority fee; demand `rebuilt_len` landing exactly on `supplied_len`.
- **A missed arm.** An arm of the coinbase path that reaches the fee effect *without* reaching `account_state_latest_balance`. I could not find one; the accessor call precedes every early return I can see.

### Also, independent of the above

`mv t0, a0; mv a0, a2; jal ra, account_read_record; mv a0, t0` stakes the header pointer on `account_read_record` preserving `t0`. Its own body comments say `t0` **is** clobbered by the compare and is reloaded (`AccountReadLog.lean:107`: `-- t0 was clobbered by the compare`). Please state explicitly whether `t0` is callee-saved across that routine, or use a callee-saved register / the stack. A restore that looks like a definition is how this class of bug hides.

**Not labelling `merge!`.** If check 1 holds, the right outcome is to close this and keep the wrapper comment; if the stage hypothesis holds, the PR should change the *ordering* instead.

pirapira:
**coord correction to my review above, on one point only.**

My last paragraph asked whether `t0` survives `account_read_record`. It does — the prologue saves and restores `t0`-`t6` (`AccountReadLog.lean:85-87`), and the docstring says so explicitly: *"Clobbers nothing the caller can see: `t0`-`t6` are saved and restored, and `a0` is only read."* I read the `-- t0 was clobbered by the compare` comment at `:107` as a caller-visible clobber; it is internal to the scan loop, after the save. **So `mv t0, a0 … mv a0, t0` here is safe — disregard that objection.**

The substantive points stand, and one of them is reinforced by the same docstring: it says the intended call sites *are the accessors themselves*, which is the design this PR moves away from by recording at a caller.

pirapira:
Closing this PR after the stage audit requested in review. The added caller-side recorder was redundant and did not move the read across an incorporation boundary:

- Amsterdam `state_tracker.py:179-203` records `get_account_optional` reads; `create_ether` at `state_tracker.py:680-700` reaches it through `modify_state`, and `fork.py:1172` invokes that producer before `fork.py:1204`/`state_tracker.py:832-880` incorporation.
- The guest coinbase path calls `balance_at_header_state_root` before `read_sets_incorporate_tx` in `BlockVerdictMtxRuntime.lean`; `balance_at_header_state_root` calls `account_state_latest_balance`, whose wrapper records the read unconditionally at `CreateCodeEffectLog.lean:486-490`. Thus the existing wrapper records in the same transaction stage, before incorporation.
- On the nonzero-priority-fee `multi_block_observed_coinbase_balance` fixture (`maxPriorityFeePerGas=0x01`, `baseFeePerGas=0x07`), matched base/candidate artifacts were byte-identical on both rows: base ELF `68d47d80e7bebfa3a4f86d4225f237ff4b2e5eda0b95ae750e945e9f0550c1e4`, candidate ELF `69f0666a45bde3fb30c7747a15dddc5a0f0a8072062de8861fcdea32e0092434`, both FA=0/FR=0, 2/2 exact, status/output differences 0.

`account_read_record` saves and restores `t0` at `AccountReadLog.lean:85-129`, so it is callee-preserved; the temporary `t0` save in the added sequence was safe but unnecessary. The wrapper comment has been restored and the branch now carries a revert commit (`c21e220fd`). No stage defect remains in this PR.

PR #11594, opened 2026-08-06
fix(codegen): align withdrawal account reads

pirapira:
## Scope

Batch A group 9 only: `process_withdrawals` constructs one `TransactionState` and calls `create_ether(wd_state, wd.address, amount)` for every withdrawal, including zero amounts (`fork.py:1207-1224`; `state_tracker.py:680-700`). The guest now records `bv_wdne_addr` once before the amount guard, so both zero and nonzero descriptors perform the spec-shaped account read. The nonzero path retains its existing field-specific pre-state balance resolution; this change does not route it through the code-bearing universal resolver.

The Stage-2 screen remains the scope boundary: groups 11 and 12 have no independent spec lookup and are already spec-shaped; group 4's zero-value/self arm is excluded because the spec only calls `move_ether` when `should_transfer_value && value != 0`; true full-account resolution remains only the CREATE code-deposit consumer and is not widened here. The payoff is a single proof/read obligation matching the spec's per-descriptor `get_account`, not uniformity for its own sake.

## Prediction

Against base `ae45b8239`, before measuring: move the existing recorder from the zero-amount arm to before the amount guard; .text/.data/.bss sizes remain unchanged, while branch displacements may change. The expected A/B flip sets are empty and that is the desired structural-alignment result.

Measured ELF sizes: base and candidate .text 350584, .data 21424, .bss 426579296.

## Evidence

Both artifacts were generated from the same source commit, with only this change:

- base commit `ae45b8239`; ELF `8583ff8b4e64e01fd1d6c204a009ed9660ae6f2792e9aa857bc25a3b6d04378f`; generated 2026-08-06 07:34:32 UTC
- candidate commit `d7654b9f7` on base `ae45b8239`; ELF `6f32ef8ef04908b2c03d3e1b6ba742c9c000d708f392eba35288ef2858acf41f`; generated 2026-08-06 07:34:17 UTC

Random-200 A/B (seed 4242, Spike, 30 workers) used:

```
scripts/codegen-eest-stateless-check.sh --backend spike --random --seed 4242 --limit 200 --jobs 30 --guest-elf <elf> --quiet-passes --run-dir <run-dir>
python3 scripts/eest-ab-compare.py <base-run> <candidate-run>
```

Both legs: FA=0, FR=1, 199/200 exact full matches, 0 errors. The one FR label is identical on both legs; flip sets are base-pass→candidate-FR = ∅ and candidate-pass→base-FR = ∅. Joined status differences = 0 and output-byte differences = 0.

Named `withdrawal_requests` controls (4 rows) were exact A/B: both legs FA=0, FR=2, identical two FR labels, empty flip sets, and byte-identical joined outputs.


pirapira:
**coord review: approved, `merge!` applied.** This one is a real alignment fix, and the reasoning checks out at every step I could test.

### The spec claim is correct

`fork.py:1224` calls `create_ether(wd_state, wd.address, wd.amount * GWEI_TO_WEI)` **unconditionally for every descriptor** in the `process_withdrawals` loop — there is no amount guard, so `get_account` (and its read record, `state_tracker.py:139`) happens for a zero amount too.

### The behavioural delta is real, not cosmetic

Before this PR the record sat *inside* the `.Lbv_wdne_zero_amount` arm, and `account_resolve_pre_state` does **not** record — the `account_read_record` call in `AccountWriteMap.lean:780` falls after the `account_resolve_execution_state:` label at `:772`, so it belongs to that routine, not the pre-state resolver. So **nonzero-amount descriptors were not recording a read at all.** That is the gate-before-record sub-shape: the guest's amount guard suppressed a read the spec makes unconditionally.

### And the newly-added records cannot become spurious entries

`block_state.account_reads` feeds `add_touched_account` (`block_access_lists.py:695-697`), which is `ensure_account` alone. For a nonzero withdrawal the recipient already has a `balance_changes` entry, so ensuring it is idempotent — the presence-only read is absorbed rather than adding a second entry. That is why `.text/.data/.bss` are unchanged and the flip sets are empty in both directions, and here the empty flip set is the *predicted* result rather than an absence of evidence.

### Register safety of the moved call — checked, it is fine

Placing `jal ra, account_read_record` **before** `beqz t2, .Lbv_wdne_zero_amount` puts a call across a live `t2` (the OR-accumulated amount), and the recorder's body does `li t2, 0xa1ea0000`. But its prologue saves and restores `t0`-`t6` (`AccountReadLog.lean:85-87`), and its docstring states this explicitly: *"Clobbers nothing the caller can see: `t0`-`t6` are saved and restored, and `a0` is only read."* So the amount test survives the call.

Nice touch keeping the explanatory comment at the branch instead of deleting it outright.

Issue #11596, opened 2026-08-06
retire the storageLogIs predicate: split transient off first, then delete the legacy append-only storage exec-log vocabulary


pirapira:
## Decision (maintainer, 2026-08-06)

**The data structure `storageLogIs` describes — the legacy append-only storage exec-log — is being retired.** So the predicate itself goes, not just the theorems built on it. This settles the liveness question that #11595 left open for the persistent-storage half: no deletion differential is needed to justify *whether* it goes, only to sequence *how*.

`storageLogIs` is defined at `EvmAsm/Evm64/StorageAssertions.lean:114`:

```
def storageLogIs (base : Word) (entries : List StorageLogEntry) : Assertion :=
  | e :: es => storageSlotIs base e ** storageLogIs (base + 128) es
```

over `PERSISTENT_STORAGE_LOG_BASE` (`:79`), which is `Stateless.STATE_TRACKER_AREA` — an area `MemoryLayout.lean` already labels *"a legacy M24 port-contract anchor"* (`:10`) and *"4 MiB legacy"* (`:90`). The replacement vocabulary is #11571's `storageWritesMapIs` / `accountWritesMapIs` over the spec-shaped maps.

## ⛔ The one blocker, and it must be done first

**`Transient/LoadSpec.lean` and `Transient/StoreSpec.lean` reuse `storageLogIs` for TLOAD/TSTORE, and transient storage (EIP-1153) is NOT being retired.** Deleting the predicate outright breaks live proofs about live code.

So the order is:

1. **Split.** Give the transient specs their own predicate — `transientLogIs` or equivalent — over the transient base, independent of `PERSISTENT_STORAGE_LOG_BASE`. Behaviour-neutral, proof-side only. This is the whole risk of the lane; do it as its own PR.
2. **Delete** `storageLogIs`, `storageLogIs_nil`/`_cons`/`_congr`, `pcFree_storageLogIs`, the `PCFree` instance, and `storageLogLenIs`, plus the persistent-side specs that consume them (`Storage/LoadSpec.lean`, `Storage/LoadLoopSpec.lean`, and the `PERSISTENT_STORAGE_LOG_BASE` machinery in `StorageAssertions.lean`).
3. **Add a `#guard` on absence** so the predicate cannot be reintroduced silently — kernel-checked beats a source read.

## Scope measured on `origin/main` `9d9d5ceba`

79 references to `storageLogIs`; the consuming files carry **76 theorems / ~3540 lines**:

[...]

PR #11597, opened 2026-08-06
refactor(transient): split transient log assertion from legacy storage log

pirapira:
## Summary

Implements Step 1 of #11596. TLOAD/TSTORE now use an independent `transientLogIs` predicate and `transientLogIs_snoc`/supporting PCFree lemmas. The persistent `storageLogIs` vocabulary remains untouched for the later deletion PR, so live transient proofs do not depend on the legacy persistent predicate.

## Scope

- `EvmAsm/Evm64/StorageAssertions.lean`: add the transient-log assertion and its append/snoc, congruence, and PCFree support.
- `EvmAsm/Evm64/Transient/LoadSpec.lean`: replace `storageLogIs` references with `transientLogIs`.
- `EvmAsm/Evm64/Transient/StoreSpec.lean`: same split for TSTORE.
- No emitted Codegen files changed; no ELF/A-B run is required.

## Evidence

- Base: `db13ef0774466e2d3e2ca1044759970d87530128` (origin/main at branch time).
- Commit: `265c0204a`.
- `lake build` ✅
- `scripts/check-axioms.sh` ✅ (137 witnesses; only classical axioms)
- `git diff --name-only` contains only the three proof files above; emitted guest inputs are unchanged.

Closes the Step 1 split in #11596; Step 2 persistent predicate deletion remains separate.


pirapira:
**coord review: verified, holding only for the build.**

Checked independently against `origin/main` `db13ef077`:

- `git grep storageLogIs` under `EvmAsm/Evm64/Transient/` on the pushed tip → **zero hits**. TLOAD/TSTORE no longer depend on the predicate being retired, which was the whole point of doing this as step 1.
- `git diff origin/main..tip -- EvmAsm/Codegen/` → **empty**. No emitted input changed, so the ELF is untouched and no A/B is owed. That is the right justification for skipping it — not an empty flip set, but the absence of any path to the image.
- Base is current main, so no rebase owed.
- `+65/-0` in `StorageAssertions.lean` (the new `transientLogIs` plus `PCFree`/`congr`/`append`/`snoc`), with the two Transient specs rewired 1:1.

Reported `check-axioms.sh` PASS at 137 witnesses with classical axioms only; CI will confirm.

**Not labelling until build is green.** A proof-side change can fail elaboration, and `merge!` hands the branch to the actor — which merges green PRs but never repairs a failing one, so an early label would lock you out of the only fix. Will label the moment it passes.

Step 2 (deleting `storageLogIs` and the persistent-side consumers) is unblocked by this landing, but keep it a separate PR.

pirapira:
**Build pass (6m0s), source-checks and summarize pass, state CLEAN — `merge!` applied.** Step 1 of #11596 done: TLOAD/TSTORE now stand on their own `transientLogIs` and no longer depend on the predicate being retired. Hands off from here; the actor owns the branch.

Step 2 is unblocked once this lands — delete `storageLogIs` and its lemma family plus the persistent-side consumers (`Storage/LoadSpec.lean`, `Storage/LoadLoopSpec.lean`, the `PERSISTENT_STORAGE_LOG_BASE` machinery), then step 3 adds the `#guard` on absence. Same evidence form: show there is no path from the diff to the emitted image, and report the build-time delta.

PR #11593, opened 2026-08-06
fix(7702): single-charge CALL delegation access (no double WARM)

pirapira:
## Summary

Residual **32** of the 2ffdac balance cluster after #11585 (AUTH_BASE): guest over-debited **+100 regular gas** (EIP-2929 `WARM_ACCESS`), sender balance short by `100 * baseFee` → `#11547` state-root byte compare.

**Root cause:** `callDelegationAccessChargeAsm` called `account_state_delegation_code_resolve` with **a3=1 (charge)** which already debits cold-delta + WARM 100, then charged **again** on status 0 (same-block target). Live SPIKE on 24842: WARM×100 from both `resolve+0x104` and `h_CALL+0x504`.

**Fix:** resolve with **a3=2 probe** (no charge/seed). This helper owns the single access charge. Status **2** (precompile target) also takes the charge path now that resolve no longer pre-charges. Corrected a3 mode comment on resolve (`0 free/seed, 1 charge, 2 probe`).

**Pattern:** charge owned in two places (resolve mode + helper) — second class after FLAG FOR ONE PURPOSE (#11585 bit6).

53-row 2ffdac was **two mechanisms** (record on #11582):
1. AUTH_BASE when pre-delegated (#11585) — 21 F→P
2. Double WARM on CALL delegation (this PR) — 32 F→P

## Measure

| set | base_bv1 | cand_bv1 | F→P | P→F |
|---|---|---|---|---|
| remain32 2ffdac residual | 32 | **0** | **32** | **0** |
| r200 seed=20260806 | — | — | 0 | 0 |
| 24842 | FR bv1 | **PASS** | | |
| controls 01114/11619/00178/05814/11710/03736 | PASS | PASS | | |

- base ELF sha256 `15cb8c287b5c9093…` @ main `9d9d5ceba`
- cand ELF sha256 `244cd580e15f42cf…` @ `4cedc8b34`
- textSizeBytes `0x55c98` → `0x55cb8` (layout pins regenerated)

## Spec

- amsterdam `system.py` CALL: one `access_gas_cost` for `to` + one `delegation_access_cost` for target (`calculate_delegation_cost`)
- `eoa_delegation.py:157-162` WARM/COLD once per target

## Test plan

- [x] remain32 F→P=32 P→F=0
- [x] r200 0/0
- [x] controls PASS
- [ ] CI layout/region-map

No merge from grok.

Issue #11588, opened 2026-08-06
structural (#11581 slice 3): synchronous account-write publication for internal CALL value transfer (both legs)


pirapira:
Slice 3 of #11581 (umbrella). Gated on the harness in the slice-1 issue.

## Family: internal CALL value transfer

- Caller debit: `ChildFrameHandlers.lean:367-412` (`.Lcd_balok`) debits live caller `env+32`
  **before** `call_frame_descend`.
- Child snapshot: `:1107-1109`; `recordMessageValueTransfer` `:1110-1124`; child credit `:1125-1161`.
- Empty-callee publication: `:1051-1063`.
- Failed-child re-credit: `CallFrameReturn.lean:237-268`, after frame records roll back.
- `record_message_value_transfer` → `record_nonstorage_effect` (`EIP7708Logs.lean:446-463`), which
  calls `account_state_record_nonstorage` **first**, then `account_write_record`
  (`NonstorageEffectLog.lean:148-167`).

## Ask

Make both legs of the transfer synchronously published at their mutation points, with the journal
mark captured before the caller debit.

⛔ **The asymmetry that makes the cheap fix wrong: publication can precede the child's live
credit, so syncing only the debit is insufficient.** Both legs, or the map is still wrong inside
the window.

Also requires per-frame undo consistency: frame marks are captured at
`CallFrameDescend.lean:491-499` and replayed by `CallFrameReturn.lean:96-107`.

Cost estimate from scoping: **high**. This is the family the 142-row regression was concentrated
in (value_moving_transactions 83, transfer_logs 20, value_moving_with_tx_delegation 8).

## Gate (structural-alignment bar)

Per the maintainer's standing ruling: a structure change that aligns with `execution-specs`
lands on **200 random A/B plus named controls** — the bar is *structure matches the spec* and
*r200 does not deteriorate a lot*, **not** byte-equivalence. An empty flip set is expected and
is not evidence of no-op-ness; alignment is its own justification because it shortens the proofs.

Required in the PR body:
- base sha + candidate ELF sha256 (both), and the r200 flip set **in both directions**
- the named controls for the family (list them; a fix verified only on failing rows has an
  unmeasured downside)
- a spec citation by file:line for the ordering being adopted

⚠️ `output[0:32]` is `new_payload_request_root`, **not** the state root. Value-neutrality
evidence is `succ` agreement plus the flip set — never "roots matched".

Umbrella: #11581

pirapira:
[...]

Issue #11581, opened 2026-08-05
structural: retire the AccountState overlay in favour of AccountWriteMap (producer-first; blocked on publication timing)


pirapira:
## Retire the `AccountState` pending/durable overlay in favour of `AccountWriteMap`

Split out of #11326 (whose storage half is complete). This is the remaining structural work on the
account side, and it is **blocked at a precisely known point** rather than open-ended.

### Survivor is settled, by differential rather than argument

`AccountWriteMap`'s tx + block arenas are the spec-shaped two-tier authority — `state_tracker.py` is
per-transaction `tx_state` merged into block level by `incorporate_tx_into_block`. Four-bucket
population differential, random-200, seed `20260805`, base `97ba11e99`:

| bucket | count |
|---|---|
| paired-agreeing | 527 |
| paired-disagreeing | **0** (raw 5, all equivalent EIP-161 tombstone encodings) |
| **overlay-only** | **0** |
| map-only | 840 (772 tx-tier, 68 block-tier — touch-only rows) |

⭐ **`overlay-only = 0` is the load-bearing result**: the map never lacks an *address* the overlay has,
so retiring the overlay cannot lose an entry. That excludes the FA-ward failure mode.

### Landed

- **#11563** — withdrawal balance reader migrated to `account_writes_latest_balance` with tx→block
  tiering. Both flip sets empty, reverse run first, old reader at **zero** production callers.

### ⛔ Blocker: the map lags live mutations during execution

A map-first wrapper repointing the other 16 normal readers produced **142 new false rejects / 139
unique digests**, all differing only at result byte 32, all from readers running **before complete
account-map incorporation**. The tier order was not the problem — #11563 uses the same order.

The order of operations is:

| path | order |
|---|---|
| withdrawals (#11563) | `record_nonstorage_effect` → `account_write_record` **immediately** ⇒ map current |
| CALL value transfer | `.Lcd_balok` debits **live** `env+32` (`ChildFrameHandlers` 367-412); records published **later** by `record_message_value_transfer` (417-424) |
| descended CALL | child snapshot (1107-1109) → `recordMessageValueTransfer` (1110-1124) → dispatch |

**Between the live debit and the record publication the overlay holds the new balance and the map holds
the old one.** So the readers cannot go map-first, and no reader-side change fixes it.

⭐ **This lane is therefore producer-first, the inverse of the storage half.**

### The one open question

[...]

PR #11598, opened 2026-08-06
Add AccountWriteMap execution-time agreement harness

pirapira:
## Summary

Adds a runtime-only AccountWriteMap versus live AccountState agreement harness. It checks populated balance and nonce fields at instrumented numeric readers, separates absent rows and absent fields, and emits tagged mismatch records with reader, map tier, limb, and differing values. The harness is diagnostics-only and does not change verdict logic.

The emitted production guest leaves `account_agreement_enabled` at its default `.dword 0`. The sweep arms it only when invoked with `--enable`, using `SPIKE_INIT_WRITES` resolved from the candidate ELF symbol; there is no build-time selector. Without `--enable`, the agreement path returns before the probes, so default-off inertness is by construction.

Base commit: `390655a07166004cfeaf9c53ee853b017536f442`

Final candidate ELF SHA256: `5ef037956e48b59f5f458f1fb3988e2874b18680e46b5dcac16bd9a12bbe0af0`

Base ELF SHA256: `cccbd088e6094834f5738face987fd0aeeea7e6f8251b154c53127dc8793a5d6`

The enabled spike sweep covered 206 cases: 200 random cases with seed 11586 plus 6 named controls. It found 246 mismatch events, with 1655 balance probes and 43 nonce probes. Reader buckets were reader 3/block: 27, reader 18/block: 27, reader 19/tx: 165, and reader 19/block: 27. The three block buckets share the same 21 case IDs; their intersection with grok's 21 IDs is empty. Enabled report: `/tmp/11586-ab-r200-enabled-final-flag.json`.

The default-off candidate/base A/B covered the same 206 cases and produced `mismatches=0` and `output_differences=0`. Default-off report: `/tmp/11586-ab-r200-default-final.json`.

## Tests

- `lake build EvmAsm`
- `scripts/check-guestaddrs-starts.sh`
- second layout-artifact regeneration was a no-op (diff hash `523608732f3ea8f6acfe1d33ac6b505e4d72a320c889cd067732d87e5f3308d3`)


pirapira:
**coord review: the analysis is exactly what I asked for and it has already changed the shape of #11581. Holding on layout pins only.**

### The four answers are all solid, and one of them refuted the scoping

- **Intersection**: each block bucket is **21 distinct case IDs** and the three sets intersect in **all 21** → one shared mechanism with three observers, not 81 defects. This is what the reader id was built for.
- **Limb**: `limb3` = `AccountState` balance BE offset `+56`, the **low 8 bytes**, with sample pairs at plausible wei magnitudes (`0c3663566a57ffff` vs `0ddff96ccd8e9fff`). So only-limb3 is the *expected* signature of a genuine publication lag, not a zero-init or endianness defect. That was the reading that could have invalidated the lane; it is now closed.
- **Nonce**: 43 invocations across 40 cases with zero mismatches, against 1655 balance invocations. **Reached and clean** — so the negative result is real and the lane is balance-only. An invocation count is what makes that evidence rather than silence.
- **`verdict=2`** is strictly row-present + field valid in map mask + field valid in live + values differ, with separate counters for no-row, row-without-field and live-field-absent. So 246 is a defect-event count, not a mixed bag.

### What it found

Readers 18 and 19 are `blockVerdictMtxStageSenderUpfront` (`:97`) and `blockVerdictMtxRecordSenderRefund` (`:105`) — the **sender upfront gas debit and refund** path, which the #11581 scoping had *audited out* of the lane as "AccountState-first/finalization paths, not this live-env gap". They are **219 of 246 disagreements**. Filed as #11599. Only reader 3 (`EIP7708Logs` `.Ldpub_call`, 27 events) belongs to a family the lane was actually scoped around → #11588.

**This is the harness paying for itself on its first run: it refuted an exclusion that two source reads had agreed on.**

### The blocker — layout pins

This PR adds emitted asm across 14 `Codegen/Programs/` files (`+146/-1` in `AccountWriteMap.lean` alone), and the reported base/candidate ELF shas differ (`4f41e576…` → `fb24684f…`). But **none of the three layout artefacts is regenerated**:

- ❌ `EvmAsm/Codegen/GuestAddrs.lean`
- ❌ `EvmAsm/Codegen/RegionMapLinkPins.lean`
- ❌ `scripts/asm-fixtures/symbol-addresses.tsv`

`source-checks` will fail at the "GuestAddrs declared-start match" step for the same reason #11593 did an hour ago. Regenerate **all three from one ELF, after the final rebase**, `git add` them, and confirm a **no-op second pass**. Never hand-edit a generated file, and check them in the tree — a regen written to `gen-out/` and never staged passes locally and fails in CI.

If the diagnostics are meant to be debug-gated out of the normal image and the `.text` growth is unintended, say so — that is a different fix and I would rather hear it than have the pins regenerated around a mistake.

**Not labelling until build is green.**

pirapira:
**coord review: the whole sequence was done correctly — `merge!` applied.**

Every item I asked for is present and independently checkable:

- **Rebased on current main** `390655a07`, with the base sha named rather than implied.
- **All three layout artefacts regenerated from one post-rebase ELF** — `GuestAddrs.lean` (+322/−322), `RegionMapLinkPins.lean` (+2/−2), `symbol-addresses.tsv` (+1282/−1267) — with a **no-op second pass** and its diff hash recorded.
- **Flag defaults to 0.** Default-off A/B: 206 cases, **0 mismatches, 0 output differences, by construction** — the data word is `.dword 0` and the probe branches to its done label before any counter or event access. That is the structural argument, not a sample.
- **Enabled sweep reproduces exactly**: 246 mismatches, `reader3/block 27 · reader18/block 27 · reader19/tx 165 · reader19/block 27`, balance 1655 / nonce 43 invocations, three block buckets sharing the same 21 IDs.
- **Intersection with grok's 21-row 2935 set: empty**, now measured against the population grok actually published rather than the 24-row local manifests. So the two lanes are independent and this baseline is unaffected by #11600.

### The runner change is the right shape, and worth naming

`spike_run.cc` adds `SPIKE_INIT_WRITES` as an **opt-in** env var — unset returns `{}`, so default runner behaviour is untouched. But the more important consequence is that the sweep enables the flag by poking it at load time instead of building a separate instrumented image. **So the ELF used for measurement is the same ELF that ships.** That closes the "did you measure the artefact you shipped?" gap entirely, which is stronger than any A/B between two builds. Prefer this pattern for future instruments.

### What this instrument has already produced

On its first run it **refuted an audited exclusion in #11581's own scoping**: readers 18/19 are the sender upfront gas debit and refund path, at 219 of 246 disagreements, against 27 for the only scoped family that appears at all. That produced #11599 — which a follow-up read-only audit then reframed again, since both those producers *do* publish at the mutation with correct undo marks, so the disagreement is a tier/stage question rather than a missing write. Two framings dead in an hour, both by artefact rather than argument.

Hands off from here — the actor owns the branch, including any pin churn if main moves again.

pirapira:
Superseded by #11606, which carries this branch's full history (73806b146 → 12abb474a) plus the cross-tier tagging commit 80f871f91, and is labelled `merge!`. Closing to keep one PR in the queue rather than two overlapping ones — nothing here is abandoned, it is all in #11606. Removing the `merge!` label first so the actor does not merge the harness twice.

PR #11600, opened 2026-08-06
fix(codegen): stage witness context for system-call SLOAD

pirapira:
## Summary

System-call path (`stage_system_call_payload`) never staged the witness trailer that user-tx `DispatchTx` always stages. Cold SLOAD tier-3 `slot_at_header_state_root` saw `header_len=0` → status 4 → value 0. Same-block user txs mask via tier-2 block map; multi-block empty-tx system calls (EIP-7002 queue) need parent pre-state.

**Fix:** after successful `stage_runtime_payload_code`, call `stage_runtime_payload_witness_context` with `svf_parent_rlp` + `svf_witness` + `svf_codes`. Size guard includes those lengths. **No BAL storage preload** (#11176 kept).

Covers all system calls via shared `stage_system_call` (7002/7251/…).

Decoder class `post_delta_missing_from_aw` for 2935 was an **altitude false friend** — 2935/4788 account RLPs were already correct; real miss was 7002 queue head.

## LENGTH (coord metric)

Worked pair 23144 (b0 PASS) / 23145 (b1 FR) fixture `single_block_max_withdrawal_requests_from_eoa`:

| cell | 23144 | 23145 base |
|---|---|---|
| bal status | 0 | 1 (hash differ) |
| rebuilt_len | 1635 | 233 |
| supplied_len | 1635 | 233 |
| rebuilt == supplied | **exact** | **exact** |

Fail code still bv=1 (#11547 state-root), not BAL length.

## IDENTITY

- 2935/4788 guest account RLP **byte-identical** to post on 23145 base
- Fixture b1 BAL: 7002 slot0 postValue=**0x0c** (b0 ended 0x0e)
- Guest base: TX map 7002 slots all **0**; bsr only 2935+4788
- Guest cand: 7002 slot0=**0xc**; integrity_fail=0; 23145 **PASS**

## Measure

vs main `db13ef077` elf `1a3b5268…`

| set | base_bv1 | cand_bv1 | F→P | P→F |
|---|---|---|---|---|
| 21-row stor2935 | 21 | **3** | **18** | **0** |
| r200 seed=20260806 | | | 0 | 0 |
| 23145 | FR | **PASS** | | |
| controls 23144/01114/05814/11710 | PASS | PASS | | |

ELF cand sha256 `36fd189b30356f57…` @ `e532aa643`
textSizeBytes `0x55c98`→`0x55d18` (RegionMapLinkPins + symbol-addresses.tsv + GuestAddrs)

**Remain FR (3):** 23274/23278/23280 consolidation multi-block — same family, separate.

## Pattern

System-call path missing witness trailer that user-tx DispatchTx always stages — cold SLOAD integrity-fail flattened to zero.

## Test plan

- [x] 23145 PASS; 7002 slot0=0xc
- [x] 21-row F→P=18 P→F=0
- [x] r200 0/0
- [x] controls PASS
- [x] GuestAddrs-starts OK fixed-point
- [ ] CI green


pirapira:
**coord review: this is the right shape of finding, and the pins are complete this time. Holding only for build green.**

### The instrument was run in the right order, and LENGTH is what redirected it

`rebuilt_len == supplied_len` **exactly** (233==233) on 23145, so the BAL `status=1` is a *content* difference, not a length one — which is precisely what rules out the "missing entry" family and sends you to identity instead. And identity then showed the decoder's `post_delta_missing_from_aw` label was an **altitude false friend**: the 2935/4788 account RLPs were already byte-identical to post on the FR base. The real miss was the EIP-7002 queue slot.

That is the LENGTH → IDENTITY → ORDERING discipline doing its job: two candidate stories died before any code was written.

### The root cause is a genuine path divergence

`stage_system_call_payload` never called `stage_runtime_payload_witness_context`, so `env+584 header_len` stayed 0, every cold SLOAD's tier-3 `slot_at_header_state_root` returned status 4 (header parse fail), and the read yielded 0 instead of `0x0c`. Same-block tier-2 masks it; only a **multi-block empty-tx** block exposes it, because that is the only shape needing parent pre-state with no user tx to stage the trailer. That explains why all 21 rows were multi-block — the family label was a consequence, not a cause.

Good that the fix **calls the shared helper** rather than reproducing DispatchTx's staging inline: the deficient path now reaches the working one instead of growing a second copy.

### Two things I checked specifically

- **The buffer bound was extended before the staging, not after.** Adding `svf_parent_rlp_len + svf_witness_len + svf_codes_len` into the `bgtu … .Lscc_toobig` sum is what keeps a large multi-block witness from overrunning `c1_staging`. Staging the trailer without widening the guard would have traded a wrong value for an overrun.
- **No BAL preload reintroduced** (#11176) — storage still read through the authenticated witness.

### Pins: all three, correctly

`GuestAddrs.lean` (+2/−2), `RegionMapLinkPins.lean` (+2/−2), `symbol-addresses.tsv` (+9/−9), base `db13ef077` is current main, and `source-checks` passes in 1m25s — including the "GuestAddrs declared-start match" step that failed on #11593. First try this time.

### Measure

21-row `stor2935`: **F→P 18, P→F 0**, `cand_bv1=3`. r200 `0/0`. 23145 PASS with 7002 slot0 = `0x0c`. Base/candidate ELF shas both stated. The residual 3 (23274/23278/23280, consolidation multi-block) is correctly held out as a separate class rather than folded in.

**Not labelling until build is green** — layout-moving PRs wait, since `merge!` hands the branch to the actor and it will not regenerate a failing one.

pirapira:
**`merge!` applied — the red build is stale-pin churn from #11593 merging (layout-moving, main now `390655a07`), which the merge actor owns.** The substance was already reviewed and approved: F→P 18 / P→F 0 on the 21-row 2935 class, r200 0/0, all three pins regenerated correctly against the pre-#11593 image, the `c1_staging` bound widened *before* the trailer staging, no BAL preload reintroduced, and the shared `stage_runtime_payload_witness_context` helper called rather than DispatchTx staging duplicated inline. Hands off from here — no rebase or regen from the author.

PR #11601, opened 2026-08-06
refactor(storage): retire persistent log proof layer

pirapira:
## Summary

Second cut for #11596, stacked on #11597. Retire the legacy append-only persistent storage-log proof surface now that the live storage path is map-based. The transient EIP-1153 assertions remain in `StorageAssertions.lean` and are independent.

- delete `EvmAsm/Evm64/Storage/LoadSpec.lean`, `LoadLoopSpec.lean`, and the now-orphaned `LoadProgram.lean`;
- remove the persistent-log constants, assertion, length predicate, append/split/placement lemmas and worked example;
- remove the obsolete SLOAD proof witness/import and classify SLOAD as `execSpec` in the progress registry;
- keep the transient log assertions and their append/snoc proofs unchanged.

This is proof/source cleanup only. The only `Codegen` edit is a documentation comment removing a deleted proof reference; no emitted string or generated ELF input changed, and no pin regeneration was needed. Step 3 (an absence guard) remains separate.

## Evidence

Base for this stacked cut: `265c0204ae18533ec47dc9e52128dac57d1139cf0` (#11597 tip; its parent main was `db13ef077`). Candidate: `95c336150`.

Prediction before measurement: emitted output unchanged; only the proof/source surface and registry counts change.

- `lake build`: 4,818 jobs, pass.
  - warm repeated runs on the final tree: `1.31–1.44 s` (three runs);
  - cache-materialized clean build: `2.42 s`;
  - non-cache clean build: `679.41 s` (dependency outputs had been removed, so this is the true no-cache cold range endpoint).
  - The pre-cut warm baseline on the Step 1 branch was `1.31 s` (4,821 jobs). The measured warm result is therefore a null delta within normal timing noise, not a claimed speedup.
- `scripts/check-axioms.sh` with `LAKE_ARTIFACT_CACHE=false`: pass, 135 witnesses; only `propext`, `Classical.choice`, and `Quot.sound`.
- `scripts/check-unimported.sh`: pass, 2,848 modules reachable, 0 orphans.
- `git diff --check`: clean.
- No emitted program strings changed; the `Codegen` diff is comment-only. No A/B was run because the linked emitted definitions and image are unchanged by construction.

The proof-surface retirement is justified by removing dead, misleading vocabulary; the timing measurement does not claim a build-time win.


pirapira:
**coord review: the deletion looks right — −2279 lines with no emitted change — but `source-checks` fails and there is a sequencing point.**

### CI failure, mechanical

`source-checks` fails at the **"Unimported-file check"** step. Deleting `Storage/LoadSpec.lean` and `Storage/LoadLoopSpec.lean` has either left a dangling import of them, or left some other file unimported because it was only reachable through their import chain. Run that check locally and fix whichever direction it is.

### Sequencing

This is stacked on `codex/11596-transient-split` (#11597), which is labelled but **not yet merged**. That is fine as a stack, but it cannot merge before #11597 does, and the diff will need to be re-based onto `main` once that lands. Say explicitly which base you rebase onto when you do.

### What checks out

- **No emitted `Codegen/` change**, so the ELF is untouched — the same "no path from the diff to the image" argument that made #11597 easy to accept, and the correct justification for skipping A/B.
- `check-axioms` passes at **135 witnesses**, classical axioms only (down from 137, consistent with removing the persistent-side witnesses).
- `Progress.lean` (+6/−21) and `AxiomWitnesses.lean` (−4) updated alongside — registry counts have to move with the deletions or the reports gate fails, so this is the right place to have touched them.

### One number to re-report

You give first rebuild 9.65s / warm 2.89s against a **pre-cut warm baseline of 1.31s** at 4821 jobs, now 4818. Taken at face value that says warm builds got *slower* after deleting 2279 lines, which is the opposite of this issue's stated payoff. I suspect it is warm-cache noise rather than a real regression, but since the build-time delta *is* the justification for #11596, please re-measure it properly — same machine, cold and warm, repeated — and report it as a range. If the win is genuinely negligible, say so plainly; that is a useful finding about where build time actually goes, and I would rather record an honest null than a flattering number.

pirapira:
**`merge!` applied.** The build failure is base staleness — this branch stacks on `codex/11596-transient-split`, which predates #11593's layout-moving merge — not anything in the diff, and the axiom lines in that log are ordinary classical-axiom prints. The merge actor owns the rebase and any pin regeneration. Substance stands: −2279 lines, no emitted `Codegen/` change so the ELF is untouched, `check-axioms` at 135 witnesses classical-only, `Progress.lean`/`AxiomWitnesses.lean` moved with the deletions, and the orphaned `Storage/LoadProgram.lean` correctly removed (2848/2848 reachable, 0 orphans).

Recorded for #11596: the build-time payoff is a **measured null** — warm 3-run range 1.31–1.44s against a 1.31s pre-cut baseline, cache-materialized clean 2.42s, true non-cache clean 679.41s / 4818 jobs. No speedup claimed. So this retirement is justified by the container convergence alone, not by build time; nobody should reopen it expecting a win.

Issue #11579, opened 2026-08-05
strategy: aim the next proof wave at the witness-read spine — compose the leaf triples into get_account_optional (the soundness core of statelessness)


pirapira:
> 🤖 Generated by Claude Code (strategist pass over execution-specs ↔ SpecRef ↔ RISC-V correspondence, 2026-08-06).

## Strategy: after the leaves, climb the witness-read spine

The leaf-routine table (#11312 → `docs/leaf-routine-targets.md`) is close to exhausted as an orienting artefact: 4 of its 10 rows are closed, 5 are open and filed (#11345–#11348, #10817), and the runners-up are now filed too. The strategic question is **what composition to aim the next wave at**, so individual triples accumulate toward something instead of staying a flat list.

## Proposal: `get_account_optional` as the first composition target

The witness-read spine is: input deserialize (`bgv_u32le` ✅) → node/code DB build → trie walk (`trieLookup`) → `mpt_node_kind` (#11347) → nibble path (`bytes_to_nibbles` ✅, `compact_to_nibbles` flag decode #11422) → `account_decode` (#11345) → EIP-161 classification (#11346). Every open leaf row is a segment of it. Its SpecRef summit is `get_account_optional` (`SpecRef/WitnessReads.lean`): *the account the guest resolves for an address is the account the witness MPT contains — or None, correctly.*

Three reasons this spine beats the alternatives (gas formulas, BAL emit, request derives) as the composition target:

1. **It is the soundness core of statelessness.** A wrong witness read is exactly the false-ACCEPT shape (#11508's four witness-missing accepts; #11522's untouched-leaf bytes; #11523's non-canonical leaves). Everything downstream (state tracker, verdict) consumes what this spine produces.
2. **It dominates the FR census indirectly**: bv_fail=1 (terminal state-root, 447 of 582 rows in #11542) is where witness-read and re-encode divergences surface, unlocalized. Stated triples along the spine convert those investigations from re-derivation into citation.
3. **Its missing pieces are already individually filed** — this issue adds only the composition obligation and the two connective gaps: (a) a trie-walk loop spec over `mptNodeIs`/`nodeDbIs` (the three-tier resolve: appended DB / resolve cache / witness section, vs SpecRef's single node source — the stated divergence in `docs/4ch8f-slstate-specref-correspondence.md:164`), and (b) the node/code DB *build* routines.

## Ask

[...]

Issue #11519, opened 2026-08-05
bug(codegen): account_decode fold arms' laLo is anchored at the ADDI, not the AUIPC — model-only, invisible to every byte-level gate (unmerged fold branch)


dhsorens:
Split out of #11483 per @pirapira's request. Found while proving the fold arms; **fixed on `proof-11483-decoded-relax` as `b2f478ccc`**, shipping with that PR.

⭐ **This one is more interesting than #11518**, because no byte-level gate *can* catch it — the emitted image is correct and only the artefact the proof quantifies over is wrong.

## The defect

Both `la` pairs appended by `44bee1bf2` pass the **ADDI's own address** to `laLo` instead of the paired AUIPC's:

```lean
.AUIPC .x5 (laHi GuestAddrs.iw_empty_trie_root (GuestAddrs.account_decode + 552)),
.ADDI  .x5 .x5 (laLo GuestAddrs.iw_empty_trie_root (GuestAddrs.account_decode + 556)),
--                                                                          ^^^ must be 552
```

`AsmReloc.lean:66-68` is explicit:

> `pc` is the paired `auipc`'s address (the relocation anchor `1b`, **not** the `addi`'s address)

and every other `la` in `accountDecode_prog` passes the same offset to both halves. Since `laLo sym pc = (sym − pc) &&& 0xfff`, anchoring 4 bytes late subtracts 4 from the low immediate while `laHi` still corrects for the AUIPC's address. Checked against the linked layout (`account_decode = 0x800062cc`):

| symbol | target | as written | |
|---|---|---|---|
| `iw_empty_trie_root` | `0xa3000dd0` | `0xa3000dcc` | −4 |
| `aie_empty_code_hash` | `0xa3000d20` | `0xa3000d1c` | −4 |

So each fold arm would load its 32 bytes starting 4 bytes **before** the constant.

## Why this is structurally invisible

The reloc table keeps both `la`s **symbolic** in the emitted text:

```lean
(138, .la .x5 "iw_empty_trie_root"),
(151, .la .x5 "aie_empty_code_hash")
```

so `emitProgramR` writes `la t0, iw_empty_trie_root`, the assembler resolves it correctly, and the **shipped binary is fine**. Byte-identity and symbol-address gates compare the emitted text, so they stay green. A conformance fixture would also pass, since it exercises the real image.

[...]

Issue #11518, opened 2026-08-05
bug(codegen): account_decode fold arms store with reversed SD operands — wild write (in the unmerged fold branch, NOT on main)


dhsorens:
Split out of #11483 per @pirapira's request. Found while proving the fold arms; **fixed on `proof-11483-decoded-relax` as `c885a1ef3`**, shipping with that PR.

## The defect

`Basic.lean:155-156` declares the store with the **base register first**:

```lean
/-- SD rs2, offset(rs1) : mem64[rs1 + sext(offset)] := rs2 -/
| SD (rs1 rs2 : Reg) (offset : BitVec 12)
```

All eight stores in the two zero-length fold arms appended to `accountDecode_prog` (`State.lean`, commit `44bee1bf2`) were written with the operands swapped:

```lean
.LD .x7 .x5 (0 : BitVec 12), .SD .x7 .x20 (0 : BitVec 12),
```

so the store is `mem64[x7] := x20` — the **output pointer** written to the address named by the **loaded hash word**, rather than the hash word written to the output slot. For `EMPTY_TRIE_ROOT`'s first dword that is a store to `0xa655cc1b171fe856`, far outside any memory window; and the constant is never written anywhere.

`Emit.lean:71` renders `sd {rs2}, {off}({rs1})`, so the emitted text is `sd x20, 0(x7)`. **`SD` carries no relocation, so the defect was in the shipped image, not only in the `Program`.**

## Correct form

`.SD .x20 .x7 (k)` for field 2 and `.SD .x21 .x7 (k)` for field 3 — the order every other store in the routine already uses: nonce `.SD .x18 .x7 (0)`, balance zeroing `.SD .x19 .x0 (k)`, copy loops `.SB .x20 .x29 (0)`.

## Why no gate caught it

Nothing executes the new arms. `.data` unchanged, `.text` changed, every `#guard` and byte-identity/symbol-address gate passed. The only artefacts that can distinguish `mem64[x7] := x20` from `mem64[x20] := x7` are a proof about what the block writes, or a fixture that reaches it.

**A zero-length conformance block would catch this one** — it is a wild write and would trap. That block is listed as still pending on #11484 and is worth landing for exactly this reason. (Contrast #11519, which a fixture would *not* catch.)

Refs #11483, #11484.

pirapira:
Which PR fixes this? Has it landed on main?

dhsorens:
## Correcting this issue's own framing: **not** on `main`

[...]

PR #11602, opened 2026-08-06
proof(#11483): fold zero-length storage_root/code_hash to the EMPTY constants — asm, predicate, and re-proof

### LINKED ISSUE #11483: guest rejects empty storage_root/code_hash leaf fields that the spec folds to EMPTY_TRIE_ROOT/EMPTY_CODE_HASH (FR; reachability open)

dhsorens:
Closes #11483. Refs #11484, #11518, #11519, #11579.

`account_decode` rejected a zero-length `storage_root`/`code_hash`; the spec folds them to `EMPTY_TRIE_ROOT`/`EMPTY_CODE_HASH` ([`witness_state.py:114-119`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/witness_state.py#L114-L119)). A false reject on a leaf the spec accepts. This lands the asm, the predicate, and the re-proof together, as @pirapira required — the asm alone leaves the library red and a repaired mirror alone describes a program that does not exist.

@pirapira's `edabb63c5` is preserved verbatim at the base.

## ⚠️ Two defects in that asm, found by trying to prove it

Both filed separately, both fixed here. Neither is on `main` — the fold arms only exist on this branch, and I have corrected #11518's title, which wrongly said "present in the shipped image".

**#11518 — all eight `SD` operands reversed** (`c885a1ef3`). `Basic.lean:155-156`: `SD rs1 rs2` is `mem64[rs1] := rs2`, base first. Written `.SD .x7 .x20`, the store was `mem64[hash_word] := output_pointer` — a wild write to `0xa655cc1b171fe856`, and the constant never written. `Emit.lean:71` renders it the same way, so this one *would* have shipped.

**#11519 — both `laLo`s anchored at the ADDI, not the paired AUIPC** (`b2f478ccc`). `+556`/`+608` should be `+552`/`+604`; `AsmReloc.lean:66-68` is explicit that `pc` is the AUIPC's address. The `Program` materialised `target − 4`.

⭐ These need **different** detection, which is why they are separate issues: a conformance fixture catches the reversed `SD`; it cannot catch the `laLo` anchor, because the reloc table keeps that symbol symbolic and the assembler resolves it correctly. **Only the proof catches the second** — which is how I found it, by being unable to prove the fold writes the constant.

## The predicate

`Decoded`'s hash clauses become `l ∈ {0, 32}` — the spec's own set: **0** from the truthiness fold, **32** from `Root`/`Hash32`, and `FixedBytes.__new__` ([`bytes.py:34-35`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum_types/bytes.py)) raising on anything else. ⛔ Never `l ≤ 32`, which would admit 1–31 that the spec rejects — the false-accept direction.

`outputSuccess`'s hash cells become `hashCell`, which takes the length and yields the constant on the zero arm; `fixed32Copied` cannot express the folded case at any offset, being an unconditional copy from the *input* buffer.

`DecodeFailure.field2Len`/`field3Len` gained `≠ 0` (`88c25da6c`). They said "decoded but length ≠ 32", which is no longer the failure condition. Left alone the predicate would be inhabitable for input the program now **accepts** — sound, since a weaker failure post is still true, but it stops characterising the failure set, which is the only reason to have it.

## The proof

Every fold-path theorem is classical-3 with **zero `sorryAx`**, including the summit:

```
'EvmAsm.Codegen.AccountDecodeSpec.account_decode_spec_within'
  depends on axioms: [propext, Classical.choice, Quot.sound]
```

Both length checks are now the 3-way the program implements:

```
field 2: adRootLenCheck → adRootZeroDispatch
           len = 0 → adRootFoldStore ⨾ adRootFoldJal ⨾ adBBField3   (EMPTY_TRIE_ROOT)
           len ≠ 0 → adRootFoldFailJal ⨾ adFailArm
field 3: adCodeLenCheck → adCodeZeroDispatch
           len = 0 → adField3FoldSuccess                            (EMPTY_CODE_HASH)
           len ≠ 0 → adCodeFoldFailJal ⨾ adFailArm
```

Two things made the merges cheap rather than a rewrite: field 3's fold rejoins `AB+496`, exactly where the copy loop's trailing NOPs land, so it reuses `adSuccessEpi` with the same footprint modulo one cell; and field 2's fold reuses `adBBField3` **unchanged**, because that chain is now generalised over the incoming root cell (`hashCell_zero` on the fold arm, `hashCell_of_ne_zero` on the copy arm). That generalisation was necessary regardless — `adField3Success` had been silently specialised to the arm it was reached from.

New: `AccountDecodeFold` (both fold backbones over one reusable `LD`/`SD` pair lemma), `adField3FoldSuccess`, `adCallPre_weaken_fold`, `adScratch_of_regs_fold`.

⚠️ **`account_decode` now READS guest data**, which it did not before, so `adFoldConstants` is a genuine new caller obligation threaded through the contract. It ripples nowhere outside this chain: `account_decode_spec_within` has no external consumers.

## Fixture repin, verified locally

`scripts/asm-fixtures/accountDecodeFunction.s` was stale and `check-asm-to-program.sh` would have failed. Rather than defer to CI I checked it here — `riscv64-elf-as` **is** installed; only the `riscv64-unknown-elf-*` spelling the gate probes first is missing:

```
lean render → 648 bytes .text
fixture     → 648 bytes .text     BYTE-IDENTICAL   (162 × 4 = 648, matching ad_length)
```

## Also: #11579's spine registered

New `Obligations.lean` row 10 for the `get_account_optional` composition, `blockedBy` = the open leaf issues (#11345/#11346/#11347/#11422/#10817) plus the two connective gaps that issue names. PROGRESS.md now shows the spine's burn-down rather than a flat count. Three kernel-checked invariants moved, not the one I expected: `blockedCount_eq` 6→7, `totalObligations` 9→10, and the id-list census.

`account_decode` is that spine's own segment, which is why this is worth landing before the next wave builds on it.

## Gates

`check-forbidden-tactics` OK · `check-axioms` OK (137 witnesses, classical-3) · `progress-report --check` OK · asm `.text` byte-identity verified above. `check-region-map.sh` skips locally on the readelf spelling; CI covers it.

## Deliberately not here

- **#11345's registry row** — its verdict depends on this landing *and* on #11513 (SpecRef's `getN` leniency, delegated). Adding it now would be stale within the week.
- **The zero-length conformance fixture** — worth landing, and it is what would have caught #11518, but it is a separate change and I would rather not grow a proof PR into it.
- **The degenerate leaf is outside `AccountRecord.rlp`'s image**, so `outputSuccess_eq_accountDecodedIs` does not cover the folded arm — `content_of_success` + `WF` pin `l2 = l3 = 32` and both `hashCell`s collapse. Modelling it needs a record that carries its encoding rather than computing one; that is a deliberate separate decision.


pirapira:
**coord review: substance verified on every claim I could check independently. One blocker (layout pins), and one process note.**

### The spec citation is exact

`witness_state.py` at the pinned `e5a8caf1b`:

```
storage_root = Root(decoded[2]) if decoded[2] else EMPTY_TRIE_ROOT
code_hash = Hash32(decoded[3]) if decoded[3] else EMPTY_CODE_HASH
```

So a zero-length field folds to the EMPTY constant, and rejecting it is **a check the spec does not have**. That puts this squarely in the category we are required to remove — otherwise equivalence needs hash-collision assumptions. Screening question (could this cause an *invalid* block to be **accepted**?): it converts a false reject into an accept-with-the-spec's-own-value, so the risk is confined to whether the folded value is right, and the fold is a two-line constant substitution matching the spec verbatim.

### Both defect claims check out, and the title correction was right

- **`SD` operand order.** `Emit.lean:71` renders `.SD rs1 rs2 off` as `sd {rs2}, {off}({rs1})` — **base first, value second**. So `.SD .x7 .x20` stores `x20` into `mem64[x7]`, exactly as described. The wild-write reading is correct.
- **Neither defect reached `main`.** `AccountDecodeFold.lean` does not exist on `origin/main` — the fold arms are branch-only, so #11518's original "present in the shipped image" was indeed wrong and correcting the title was the right call rather than leaving a false severity on record.

⭐ **And the point about detection asymmetry is the most valuable thing in this PR, beyond the fix itself.** A conformance fixture catches the reversed `SD`; it *cannot* catch the `laLo` anchor, because the reloc table keeps the symbol symbolic and the assembler resolves it correctly — the `Program` is wrong while the assembled output is right. That is a concrete case where **only the proof observes the defect**, which is the strongest argument yet for landing asm and proof together rather than sequencing them. Worth citing whenever the asm-first shortcut comes up again.

I also note the `.text` byte-identity check (Lean render 648 = fixture 648 = 162 × 4) and the explicit "deliberately not here" section. Naming what was excluded and why — the #11345 registry row, the zero-length conformance fixture, the degenerate-leaf modelling — is better practice than a silent narrowing, and I am not asking for any of the three in this PR.

### Blocker: layout pins are owed

`ad_length` goes **136 → 162** (+26 instructions, **+104 bytes** of `.text`), so every later address shifts. `account_decode` is a linked converted symbol (present in both `GuestAddrs.lean` and `symbol-addresses.tsv`), and `source-checks` already fails at:

> **GuestAddrs declared-start match (GuestAddrs.lean vs symbol-addresses.tsv)**

None of the three artefacts is in the diff. Regenerate all three — `GuestAddrs.lean`, `RegionMapLinkPins.lean`, `scripts/asm-fixtures/symbol-addresses.tsv` — from **one** ELF after the final rebase, `git add` them, and confirm the second pass is a **no-op**.

⚠️ **The local skip is the thing that hid this.** The body says `check-region-map.sh` skips locally on the `readelf` spelling and "CI covers it" — but the gate that actually failed is the *GuestAddrs* one, not region-map, and it would have failed locally too. A check that silently skips reads as a pass; that is the same hazard as #11557's silent `exit 1`. Worth a separate issue for the `readelf` spelling so the next author is not misled the same way.

### Labelling

I will hold `merge!` until **build** completes rather than labelling now. The distinction I am drawing: stale-pin and rebase churn is the merge actor's to resolve, but this PR carries ~1550 lines of proof, and a proof failure is an **author-only** fix — labelling first would hand off the branch and lock you out of the only repair. Once the build is green (or green apart from the pins), I will label and the actor can take the regeneration.

pirapira:
**Correction to my labelling note: `merge!` applied now.** Pin regeneration is the merge actor's to handle, so holding for it was my error — the `GuestAddrs`/TSV drift from `ad_length` 136→162 does not need to come from you.

Substance stands as reviewed: spec citation exact at the pin, `SD` base-first confirmed via `Emit.lean:71`, and `AccountDecodeFold.lean` absent from `main` so neither defect shipped. Hands off from here.

Two things I would still like, but as **separate** issues rather than additions to this PR — neither blocks it:

1. The `check-region-map.sh` local skip on the `readelf` spelling. A check that silently skips reads as a pass, which is what let the pin drift reach CI; same hazard as #11557.
2. The zero-length conformance fixture you named — worth landing on its own, since it is what would have caught #11518.

pirapira:
Heads-up from the merge queue: this branch fails `check-asm-to-program` **standalone**, before any merge —

```
DRIFT DETECTED:
  [EvmAsm/Codegen/Programs/State.lean] accountDecodeFunction: generated block not found verbatim (source drift)
```

`scripts/asm-fixtures/accountDecodeFunction.s` was edited but the emitted block in `State.lean` wasn't regenerated from it. I've regenerated it inside the batch PR #11609 with the documented `.s → Lean` direction:

```
python3 scripts/asm_to_program.py rewrite --file EvmAsm/Codegen/Programs/State.lean --funcs accountDecodeFunction
```

That rewrites 27 lines of emitted program, and `lake build` stays green afterwards, so your `AccountDecode*` proofs still hold against the regenerated bytes. No action needed to land it — but since this is byte-pinned emitted content rather than layout addresses, please sanity-check the regenerated block matches your intent.

Issue #11582, opened 2026-08-05
carry-forward: remaining guest-vs-spec divergence entries from the #11326 map (re-rank against a current census)


pirapira:
## Carry-forward: remaining guest-vs-spec divergence entries from #11326's map

Split out of #11326 so the survey outlives it. #11326 held a ranked map of structural differences
between the guest and execution-specs at pin `e5a8caf1b`, built spec-structures-outward. It has grown
too large to be a working issue; its concrete lanes are now separate, and this is the remainder.

### Entries resolved since the map was written

| entry | disposition |
|---|---|
| storage write-map vs append-only log | ✅ **retired entirely** — #11549, #11552, #11560 |
| account overlay vs `AccountWriteMap` | → **#11581** (survivor settled, blocked on publication timing) |
| universal resolver | → **#11104** (stage 1 landed as #11550; stage 2 scoped to 10 of 12 groups) |
| code-60 driver (BAL content) | ✅ closed — 127/128 rows across #11545, #11546; survivor is a fixture defect (#11547) |
| beacon-roots / 4788 code-1 | ✅ closed — 96/96, #11553 |
| selfdestruct / recreate residue | ✅ closed — #11561, #11562, #11565, #11566, #11568 (tracked in #11564) |

### What still needs re-verification

The map's remaining entries were re-ranked once (beacon closed, selfdestruct promoted then completed),
but **the ranking rested on a context census that is now badly stale**: FR 375 with code1=238 and
code60=125, measured 2026-08-03 on base `3d395c1ba`. Actual as of `97ba11e99`: **OK 25,800 · FAULT 0 ·
FR 304 · FA 0**, with code1=296, code60=1, code58=3, plus singletons at 11/53/56.

⛔ **And the buckets cannot be updated by arithmetic.** 11617 read `bv_fail=1` before #11565 and
`bv_fail=53` after — **a row's code is a property of the tree, not of the row**, because the verdict
check order is a masking hierarchy and closing an early check surfaces a later one. So code-1 at 296 is
*296 rows whose first failure is a state-root mismatch*, not 296 state-root defects.

### Method to keep when this is picked up

- **Spec-structures-outward**, citing the fork directory directly — never `SpecRef`, which is a port and
  drifts, and citing our own Lean to justify spec alignment is circular.
- `grep` the **whole** fork dir per claim: `accounts_to_delete` lives in `vm/__init__.py:184`, not
[...]

PR #11604, opened 2026-08-06
fix(codegen): preserve storage baseline on block upsert

pirapira:
## Summary

`write_sets_incorporate_tx` merged tx storage rows into the block map via `storage_writes_block_upsert` **without copying the EIP-2200 original at +96**. `execution_map_state_changes` (`Lem_storage`) then compared final value (`+64`) to a zero baseline (`+96`) and treated zero-clears of nonzero parent slots as **unchanged** → omitted from MPT apply → **#11547** state-root byte compare fail.

This is the residual-3 after #11600 (23274/23278/23280 consolidation multi-block).

### LENGTH (coord metric)

| cell | 23274 base (#11600) | 23274 cand |
|---|---|---|
| rebuilt_len == supplied_len | EXACT (233) | EXACT |
| bal status | 0 (hash match) | 0 |
| fail | bv=1 #11547 | **PASS** |

### IDENTITY / ORDERING

1. TX `storage_write_record` correctly captured original at +96 (7251 slot0 original=1, current=0).
2. BAL emit runs on **TX map before merge** → hash match (status=0) even while state root failed.
3. Block upsert dropped +96 → block row `val=0, baseline=0` → Lem skip → parent leaf slot0=1 survives.
4. Why #11600's 7002 F→P still worked with bad SLOAD: SSTORE **nonzero** vs baseline 0 still looks like a delta. Zero-clear of nonzero pre is the only silent omission.

### Fix

- `storage_writes_block_upsert`: a3 = baseline ptr (or 0). **APPEND-only** copy into +96; **HIT leaves +96 alone** (first-write / pre-block baseline survives).
- `write_sets_incorporate_tx`: pass `a3 = tx_entry+96`.
- System-storage seed paths: pass original@+64 when known; seed-only uses a3=0.

### Pattern

**ZERO-AS-MISS RECORDED AS ZERO-AS-VALUE** via lost baseline (omit path) — 6th pattern class; fail-closed on tier-3 miss remains secondary.

## Measure

| set | result |
|---|---|
| residual 23274/78/80 | base_bv1=3 cand_bv1=**0** **F→P=3 P→F=0** |
| 21-row stor2935 | base_bv1=3 cand_bv1=**0** F→P=3 P→F=0 |
| r200 seed=20260806 | 0/0 |
| controls 23273/23145/23144/01114/05814/11710 | PASS |

- base ELF sha256 `36fd189b30356f57…` (#11600 only @ e532aa643 tree)
- cand ELF sha256 `04708856106e00f9…` (#11600 + this fix)
- textSizeBytes `0x55cb8` → `0x55d00`
- Layout: RegionMapLinkPins + symbol-addresses.tsv + GuestAddrs fixed-point

**Depends on #11600** (system-call witness staging) for residual 3 — without staging, tier-3 misses and original stays 0 so baseline copy alone cannot see the clear.

## Spec / code cites

- `StorageWriteMap.lean`: tx baseline capture APPEND-only; block consumer now `Lem_storage` (+64 vs +96)
- `BlockVerdictStateRoot.lean` Lem_storage:216-220
- Fixture: `single_block_max_consolidation_requests_from_eoa` (23273 b0 PASS / 23274 b1 FR→PASS)
- EIP-7002/7251 queue head clear: SSTORE 0 over parent 1

## Hands-off

- Do not touch #11600 (merge! — merge actor owns pin churn)
- No merge from grok


pirapira:
**coord review: `merge!` applied. The diagnosis is the strongest of the night and two of its by-products are worth more than the three rows.**

### Root cause

`write_sets_incorporate_tx` → `storage_writes_block_upsert(a0, a1, a2=value)` **never copied the baseline**. The tx map had `7251 slot0` correct — `original@+96 = 1`, `current@+64 = 0` — and the BAL emit runs pre-merge, so the BAL hash matched. But the block upsert dropped `+96`, so `.Lem_storage` compared `0 == 0`, skipped the MPT apply, and the parent leaf survived into the state root. Hence `bv=1` (#11547) with a *clean* BAL — a combination that would have looked contradictory without the tier-by-tier read.

### ⭐ The stale comment is the transferable lesson

The upsert's comment asserted `+96` was **INERT / unused**. That was true when written and became false when `.Lem_storage` started reading block `+96`. **A comment claiming a field is unused is a claim with an expiry date, and nothing enforces it.** This is the same shape as "an unused datum is not an unused status" — the register of things that have nearly been deleted on the strength of such a comment is long. If there is a cheap way to convert that class of claim into a `#guard` or an assertion, it is worth doing; a comment cannot fail a build.

### ⭐ And the characterisation of the silent-omission window is the real find

> SSTORE nonzero vs baseline 0 is still a delta. **A zero-clear of a nonzero pre-value is the only silent omission.**

That explains why #11600's 7002 fix flipped its rows F→P *despite* a still-broken SLOAD: a wrong `0` baseline is harmless whenever the new value is nonzero, because the comparison still sees a difference. So the defect class is far narrower than "any lost baseline" — it is exactly `SSTORE 0` over a nonzero original. That is the kind of statement that tells you which rows can still be hiding, and it should go on #11582 rather than only in this PR.

This is the **seventh** zero-as-value vs zero-as-miss instance tonight, and the second where the symptom is a *silent omission* rather than a wrong value.

### Fix shape — correct on the tier semantics

`a3 = baseline`, **APPEND-only** copy of `+96`, HIT leaves `+96` alone. That is right: the block-level original must be the **earliest** original in the block, not the latest, so a second tx touching the same slot must not overwrite it. Had the copy applied on HIT too, the block baseline would track the most recent tx and the same class of comparison bug would reappear from the other direction.

### Evidence

Residual 3 **F→P 3 / P→F 0**; 21-row `cand_bv1 = 0`; r200 `0/0`; controls PASS. All three layout artefacts regenerated (`GuestAddrs` +238/−238, `RegionMapLinkPins` +1/−1, `symbol-addresses.tsv` +747/−747), `textSize` `0x55cb8 → 0x55d00`, base and candidate ELF shas both stated.

Depends on #11600, which is in the actor's batch #11603 — sequencing is the actor's to resolve. Hands off from here.

Issue #11605, opened 2026-08-06
harness observability: per-reader invocation counters, a registered-reader table that fails loudly, and post-mutation observation points


pirapira:
## Rescoped again (2026-08-06, third time) — per-callsite IDs ALREADY EXIST. This issue is now counters + registry + post-mutation observation.

**Correcting the premise of both earlier versions of this issue.** Reading the emitted source on the harness branch directly:

- `BlockVerdictDispatchTx.lean:554` → `li a2, 1; jal ra, account_state_latest_balance`
- `ChildFrameHandlers.lean:493` → `li a2, 5; jal ra, account_state_latest_balance`

**Every reader callsite already sets its own literal id.** The allocation is: reader1 root recipient (`DispatchTx:554`), reader2 BAL, reader3 EIP7708 transfer, readers4/5 internal CALL, reader6 precompile, readers7–16 SELFDESTRUCT, reader17 block stage (`Mtx:44`), reader18 (`Mtx:97`), reader19 (`Mtx:105`), readers21/22 CREATE nonce.

So the claim that the probe inherits an unset `a2` was wrong, and the `li a2, <id>` work this issue asked for does not exist to be done.

### What the silence actually means

Only four ids appear in the event stream because **only mismatching reads emit events**. Therefore silence on readers 1, 2, 4, 5, 6, 7–16, 21, 22 means those reads *were probed and agreed* — genuine agreement, not absent instrumentation. That is the reverse of what the previous two comments here said.

### ⭐ But the remaining gap is real and is the substantive one

**The existing ids sit mostly at PRE-mutation reads.** So silence on #11587–#11590 does not prove *publication*: a read taken before the mutation cannot detect that the map failed to receive it. The families remain unfalsifiable on ordering grounds, not on instrumentation grounds.

## Ask, in priority order

1. **Per-reader invocation counters.** The artefact carries only aggregate `balance_probe_invocations=1655` / `nonce_probe_invocations=43`, which cannot show that a *given* site ran. Without this, a zero event count for reader6 is indistinguishable from reader6 never executing.
2. **A registered-reader table** enumerating all known ids (1–19, 21–22) with site and entitled tier, and **fail loudly on an unregistered reader id** rather than silently classifying it as instrument/status 3. A silent default is how a renumbering or a stray `a2` would become invisible.
3. **Post-mutation observation points**, per family, so publication ordering becomes falsifiable:
   - root recipient — SELFBALANCE/BALANCE **after** the credit
[...]

Issue #11599, opened 2026-08-06
structural (#11581 slice 7): post-refund sender balance disagrees with the tx map at BlockVerdictMtxRuntime:105 — 165 events, pre value agrees


pirapira:
Slice 7 of #11581 (umbrella). **This family was explicitly excluded during scoping; the #11586 harness refuted the exclusion.**

## How it was found

The scoping audit on #11581 excluded upfront gas debit/refund from the lane, on the reasoning that they are "AccountState-first/finalization paths, not this live-env gap". The #11586 agreement harness contradicts that: **219 of 246 balance disagreements come from these two readers**, dwarfing every family the lane was actually scoped around.

| reader | callsite | tier | events |
|---|---|---|---|
| 18 | `BlockVerdictMtxRuntime` `blockVerdictMtxStageSenderUpfront` — pre sender balance (`:97`) | block | 27 |
| 19 | `BlockVerdictMtxRuntime` `blockVerdictMtxRecordSenderRefund` — post-refund sender balance (`:105`) | tx | **165** |
| 19 | same | block | 27 |

For contrast, the only *scoped* family that shows up at all is reader 3 (`EIP7708Logs` `.Ldpub_call`, CALL sender balance lookup) at 27 events — routed to #11588.

## Ask

Publish the account-write row synchronously at the sender's **upfront gas debit** and **refund** mutation points, with the `account_writes` undo mark captured before each live mutation — the same treatment the other family slices get.

⚠️ Note this family is *not* a value transfer between two parties, so the paired-legs requirement that governs #11588/#11589/#11590 does not apply in the same form. But the refund reader is the dominant single bucket in the whole harness, so treat the debit and the refund as two separate mutation points, not one.

⚠️ `AccountWriteMap.lean:875-880` intentionally has **no tx-discard** precisely because *sender inclusion and gas* survive an ordinary body revert (`interpreter.py:428-429` rolls back body effects only). So the sender's gas rows are exactly the ones that must **not** be undone by a body revert — get the mark placement right or this family will regress failed-transaction rows.

## Evidence available from #11586

- Disagreements are `verdict=2` — strictly: row present, field valid in the map mask, field valid in live `AccountState`, values differ. No-row, row-without-field, and live-field-absent have separate counters and never emit `verdict=2`.
[...]

PR #11606, opened 2026-08-06
Tag cross-tier account agreement events

pirapira:
## Summary

Tags mismatches against a map tier that is not the reader's entitled tier as status 3 instrumentation records, keeping status 2 for semantic agreement failures.

The reader-tier contract is:
- readers 3, 18, and 19: TX tier
- reader 17: BLOCK tier

The harness remains inert by default and is armed only by the sweep's explicit `--enable` path.

## Evidence

- Git base: `d66605ff4`
- Base ELF SHA-256: `5ef037956e48b59f5f458f1fb3988e2874b18680e46b5dcac16bd9a12bbe0af0`
- Candidate ELF SHA-256: `f1eecf8c0007bd49620020e0f8240bb4f8f03234338efa0b874ceb567a740eda`
- 206 cases: 200 seeded random cases plus 6 named controls
- Raw agreement events: 246
- Semantic mismatches: 165, all reader 19 / TX tier
- Cross-tier instrumentation: 81

| reader | tier | status | events |
|---:|---|---:|---:|
| 3 | BLOCK | 3 | 27 |
| 18 | BLOCK | 3 | 27 |
| 19 | TX | 2 | 165 |
| 19 | BLOCK | 3 | 27 |

The corrected source map is reader 17 at `BlockVerdictMtxRuntime.lean:44` (block-first upfront read), reader 18 at `:97` (pre-refund read inside the refund routine), and reader 19 at `:105` (post-refund read). Reader 3 is `EIP7708Logs.lean:364`.

For all 165 reader19/TX events, the field is balance limb 3. Event values are raw little-endian loads of BE balance-buffer dwords, so raw subtraction is not the balance delta. Reversing each recorded 64-bit dword back to BE first gives:

`BE(live_limb) - BE(map_limb) = gas_refund_amount`

for 165/165 cases. Example: case 00069 has `gas_refund_amount = 119791888000000000`.

The candidate/base A/B run with `--enable` reports `output_differences=0` for all 206 cases. This is by construction: the harness changes only diagnostic counters/events, and the production flag remains disabled unless explicitly armed. The default-off candidate sweep also reports zero mismatches, zero instrument events, and zero output differences.

Generated metadata second pass was a no-op; the full diff hash remained `7b7b9fdc33d99536adf4a4720e8c80d767333ab77121ac6378b29a61d57b0404`.

## Checks

- `lake build EvmAsm` — 4,821 jobs, clean
- `scripts/check-layering.sh`
- `scripts/check-roundtrip-coverage.sh`
- `scripts/check-opcode-structure.sh`
- `git diff --check`

The refund publication fix itself is intentionally out of scope for this instrumentation PR.


pirapira:
**coord review: `merge!` applied. This supersedes #11598, which I am closing — see below.**

`git log origin/main..head` shows this branch carries #11598's full history (`73806b146` → `12abb474a`) **plus** `80f871f91` "Tag cross-tier account agreement events". So it is a strict superset, and leaving both open risks the actor merging the same harness twice.

Verified against `origin/main` `d66605ff4`:

- Base is current main, named rather than implied.
- All three layout artefacts regenerated (`GuestAddrs` +322/−322, `RegionMapLinkPins` +2/−2, `symbol-addresses.tsv` +1283/−1267) with a no-op second pass.
- `AccountWriteMap.lean` +172/−1 vs #11598's +148/−1 — the 24-line delta is the tier tagging, as described.
- `spike_run.cc` unchanged from #11598's opt-in `SPIKE_INIT_WRITES`, so the shipping ELF is still what gets measured.

### What this PR adds, and why it matters

Event accounting is now explicit: **246 raw = 165 semantic mismatches (all reader19/TX) + 81 instrument (reader3/BLOCK 27, reader18/BLOCK 27, reader19/BLOCK 27)**. Without this, every future sweep reports 81 non-defects and the next reader re-derives the tier cut from scratch.

Three things in the body are worth more than the diff:

1. **The BE normalisation rule, stated explicitly.** Rows are BE buffers while the raw event dwords are LE machine loads, so subtracting raw dwords gives a number that is wrong except by coincidence. I made exactly that error earlier on #11599 and got the right answer by luck on one row and meaningless figures on two others. Having it in the PR body rather than a commit message is the right call.
2. **The 165/165 `gas_refund_amount` identity** — an exact match on every row, not a sample, which is what turned #11599 from a hypothesis into a named mechanism.
3. **The corrected reader map** — reader17 `:44` block-first, reader18 `:97` pre-refund, reader19 `:105` post-refund. The earlier attribution of reader18 to `blockVerdictMtxStageSenderUpfront` was wrong and is superseded.

⚠️ One limit to carry forward, now recorded on #11605: the probe is **wrapper-level** and the reader field is the caller's inherited `a2`, so only these four ids are trustworthy. The other twelve balance callsites are *absent from the event stream*, not *proven agreeing*.

Hands off — the actor owns this, including pin churn if main moves.

Issue #11595, opened 2026-08-06
cleanup: delete theorems about the retired append-only storage exec-log (76 theorems / ~3.5k lines, liveness to be established per file)


pirapira:
## Ask

Delete the theorems and specs whose subject is the **legacy append-only storage exec-log**, which the convergence has been retiring in favour of the spec-shaped `StorageWriteMap`/`AccountWriteMap` pair. Proving things about a container we are removing is wasted surface: it slows every build, it has to be relocated whenever layout moves, and it makes the retirement look more expensive than it is.

⛔ **But establish liveness per file first — the census below is ambiguous and a blanket delete would be wrong.**

## The candidate set (measured on `origin/main` `9d9d5ceba`)

`storageLogIs` — described in #11571 as covering *"the legacy append-only logs"* — has **79 references across 4 files**, and the surrounding proof bodies are large:

| file | theorems | lines |
|---|---|---|
| `EvmAsm/Evm64/Storage/LoadLoopSpec.lean` | 33 | 1128 |
| `EvmAsm/Evm64/Storage/LoadSpec.lean` | 12 | 847 |
| `EvmAsm/Evm64/Transient/LoadSpec.lean` | 11 | 827 |
| `EvmAsm/Evm64/Transient/StoreSpec.lean` | 6 | 411 |
| `EvmAsm/Evm64/StorageAssertions.lean` | 14 | 327 |
| **total** | **76** | **3540** |

## What makes this ambiguous — resolve before deleting

1. **`PERSISTENT_STORAGE_LOG_BASE` is not its own region.** `StorageAssertions.lean:79` defines it as `Stateless.STATE_TRACKER_AREA` (`0xa0630000`). `MemoryLayout.lean` calls that area **"a legacy M24 port-contract anchor"** (`:10`) and **"4 MiB legacy"** (`:90`) — which is the artefact agreeing the log is legacy.
2. ⚠️ **But six emitted files still reference the area**: `BlockVerdictParams.lean`, `BlockVerdictSystemStorageCapture.lean`, `ExecLogStorageSeed.lean`, `Storage.lean`, `StorageMultiContract.lean`, `RegionMap.lean`. So the *address* is live even if the *append-only log interpretation* is not. Deleting predicates about a structure the guest still writes would be wrong.
[...]

PR #11607, opened 2026-08-06
ci: guard retired storage-log predicates

pirapira:
## Summary

Add a source-level absence gate for the persistent append-only storage-log proof vocabulary retired by #11601.

- `scripts/check-retired-predicates.sh` rejects `storageLogIs` and its retired companion vocabulary under `EvmAsm/**/*.lean`.
- Backtick-wrapped historical references remain allowed.
- The source-check workflow runs the gate and its negative-control self-test.

This is a source/CI guard only; it changes no Lean declarations, emitted strings, generated artifacts, or guest bytes.

## Negative control

The self-test plants a temporary `def storageLogIs := True` and verifies a nonzero exit, then checks that backtick documentation is allowed. I also manually reintroduced the same trivial declaration in `EvmAsm/Evm64/StorageAssertions.lean`; the real-tree scan rejected it, and the declaration was removed before this commit.

## Checks

- `scripts/check-retired-predicates.sh --self-test` ✅
- `scripts/check-retired-predicates.sh` ✅
- `bash -n scripts/check-retired-predicates.sh` ✅
- `git diff --check` ✅

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,10 @@ jobs:
       - uses: actions/checkout@v4
       - name: Forbidden-tactic guard (no bv_decide / native_decide / TCB expanders)
         run: scripts/check-forbidden-tactics.sh
+      - name: Retired-predicate absence guard (#11596)
+        run: |
+          scripts/check-retired-predicates.sh --self-test
+          scripts/check-retired-predicates.sh
       - name: Sail pin guard (vendored model matches its provenance extraction)
         run: scripts/check-sail-pin.sh
       - name: Symbolic guest-PC guard (SAsm wrappers follow GuestAddrs)
--- a/scripts/check-retired-predicates.sh
+++ b/scripts/check-retired-predicates.sh
@@ -1,0 +1,104 @@
+#!/usr/bin/env bash
+# check-retired-predicates.sh — keep retired proof vocabulary out of EvmAsm.
+#
+# The persistent append-only storage-log assertion was retired in #11601.  A
+# later reintroduction would compile cleanly but recreate the misleading
+# container vocabulary that the retirement removed, so this is intentionally a
+# source-level denylist rather than a registry or emitted-ELF check.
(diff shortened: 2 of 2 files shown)


pirapira:
**coord review: `merge!` applied. This is a better gate than I asked for, in a specific way worth naming.**

I asked for a negative control run once and reported in the body. You built **`--self-test` into the script and wired it into CI**, so the control runs on every build rather than being a claim about a past local run. That is the difference between *"I verified this gate works"* and *"this gate cannot silently stop working"* — the second survives the author leaving, a refactor of the grep, or a future change to the token-boundary regex.

And it is **two-sided**: the self-test plants a declaration in a temp Lean file and requires rejection, *and* checks a backtick-documented mention is still allowed. Both directions matter here — the retirement issues (#11596, #11595) and several PR bodies name `storageLogIs` throughout, and a gate that forbids documenting its own subject is a gate someone deletes in a month.

Verified:

- Sibling script (`check-retired-predicates.sh`) rather than extending `FORBIDDEN` in `check-forbidden-tactics.sh` — the choice I preferred, since that list carries a specific TCB meaning and a retired-vocabulary entry would dilute it.
- Denylist covers the family, not just the head name: `storageLogIs`, `storageLogIs_nil`, `_cons`, `_congr`, `pcFree_storageLogIs`. A gate on the predicate alone would let the lemmas back in.
- Blocking step in `build.yml`, so it fails the build rather than reporting.
- `+108/−0` across exactly two files, **no emitted `Codegen/` change**, so no ELF movement and no layout pins owed. Nothing to regenerate.
- Manual control also run: reintroduced `def storageLogIs` in `StorageAssertions.lean`, gate exited 1, removed it, gate passed.

This closes #11596's step 3, and with it the retirement: the predicate is gone from the tree (zero references on `11fcab529`), TLOAD/TSTORE stand on their own `transientLogIs`, and reintroduction now fails CI.

Hands off — the actor owns it.

PR #11608, opened 2026-08-06
fix(codegen): raise EXTCODECOPY length cap to EIP-7907 MAX_CODE_SIZE

pirapira:
## Summary
- `extcodecopy_at_header_state_root` rejected `length > 32768` (status 6) while EIP-7907 raised `MAX_CODE_SIZE` to **65536**.
- Full-size `EXTCODECOPY` left the pre-zeroed memory window → SHA3 of 64k zeros → wrong storage slot → **#11547** state-root mismatch.
- `EXTCODESIZE` / `EXTCODEHASH` had no such cap (worked on the same accounts).
- Runtime OOB code-offset normalize raised 32768 → 65536 to match.

Decoder class `nonce_mismatch` on these rows was a **false friend** (TOUCHED-only AW); applied account RLP already had correct nonce/code_hash.

## Spec
- Amsterdam EIP-7907 / `CreateDeployedCodeValid.maxDeployedCodeSize = 65536`
- EXTCODECOPY zero-pads past end; does not error on missing account

## Measure vs main `11fcab529`
| set | result |
|---|---|
| 01308 / 01315 max_code_size | **F→P** |
| r200 seed=20260806 | 0/0 |
| controls 01114/11619/00178/05814/11710/03736 | PASS |

ELF base `41877692…` cand `88e1f721…`. text size unchanged (LUI imm only) — no layout pin regen.

## Remain (not this PR)
- **12832/12833/12834** `precompile_absence`: high-key storage clear with baseline=0 (omit 0→0) — related to SLOAD original / #11604 baseline class
- **00026** guest_extra OOG rollback delegations — separate

## Pattern
Half-migrated constant: deployed-code cap lifted in CREATE validator, left at old half-limit on EXTCODECOPY length gate.

## Do not
- No merge from grok; await label

### DIFF
--- a/EvmAsm/Codegen/Programs/EvmExtcodecopy.lean
+++ b/EvmAsm/Codegen/Programs/EvmExtcodecopy.lean
@@ -53,8 +53,8 @@ private def extcodecopyWitnessTail : HandlerTail :=
     memDynamicU256RangeOogGuardAsm
       "extcodecopy" "x12" "x15" "x16" "x17" 32 96 ++
     -- eccob.1: both EXTCODECOPY copy paths consume only the low 64-bit code offset.
-    -- If any high limb is nonzero, or the low limb is at/above the deployed-code cap,
-    -- normalize to 32768 so the existing zero-padded copy loops cannot wrap the source index.
+    -- If any high limb is nonzero, or the low limb is at/above EIP-7907 MAX_CODE_SIZE
+    -- (65536), normalize to 65536 so zero-padded copy loops cannot wrap the source index.
     "  ld x16, 64(x12)
 " ++
     "  ld x17, 72(x12)
@@ -66,12 +66,12 @@ private def extcodecopyWitnessTail : HandlerTail :=
 " ++
     "  bnez x17, .Lrt_ecc_oob_offset
 " ++
-    "  li x18, 32768
+    "  li x18, 65536
   bltu x16, x18, .Lrt_ecc_offset_ok
 " ++
     ".Lrt_ecc_oob_offset:
 " ++
-    "  li x18, 32768
+    "  li x18, 65536
   sd x18, 64(x12)
 " ++
     ".Lrt_ecc_offset_ok:
--- a/EvmAsm/Codegen/Programs/EvmOpcodesExtcodecopy.lean
+++ b/EvmAsm/Codegen/Programs/EvmOpcodesExtcodecopy.lean
@@ -52,7 +52,7 @@ open EvmAsm.Rv64.Program
       a1 (input)  : header_rlp_len
       a2 (input)  : address ptr (20 bytes)
       a3 (input)  : code_offset (u64)
-      a4 (input)  : length (u64; must be <= 32768)
+      a4 (input)  : length (u64; must be <= MAX_CODE_SIZE = 65536 / EIP-7907)
       a5 (input)  : output buffer ptr (`length` bytes)
       a6 (input)  : witness.state ptr
       a7 (input)  : witness.state len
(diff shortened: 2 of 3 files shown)


pirapira:
**coord review: the constant is verified exactly right and `merge!` is applied. But the same wrong constant exists elsewhere — see the sweep request below, which I think is worth more than this fix.**

### Spec check

`execution-specs` at the pin, `amsterdam/vm/interpreter.py:77`:

```
MAX_CODE_SIZE = 0x10000
MAX_INIT_CODE_SIZE = 2 * MAX_CODE_SIZE
```

`0x10000` = **65536**, enforced at `:219` (`if len(contract_code) > MAX_CODE_SIZE`). So the guest's `32768` was **exactly half** the spec limit, and normalising to 65536 matches it. Screening question — could this cause an *invalid* block to be **accepted**? No: this is a value computation, not a control dependency, so a wrong copy yields a wrong root and a rejection. Raising the clamp to the spec value makes correct inputs produce correct values; it does not admit anything the spec rejects.

### The false friend is the interesting part

The decoder's `nonce_mismatch × 5` label was wrong: the AW row was TOUCHED-only but the applied RLP nonce and `code_hash` were fine. The real defect was `extcodecopy_at_header_state_root` returning status 6 for `length > 32768`, leaving memory pre-zeroed, so `SHA3` hashed **64k of zeros** instead of the real code.

**And that refutes the pointer I sent you.** I predicted the mask/write coupling upstream — that a TOUCHED-only `vmask` meant the writer was never told about the nonce — and offered `a nonzero-but-wrong nonce anywhere` as the falsifier. The actual answer was that the nonce was never the defect at all. My pointer would have sent you into `AccountWriteMap`'s mask composition and found nothing. Good that you followed the artefact instead of the framing.

### ⚠️ Sweep request: the same `32768` belief is in at least one more place, and that one is a SIZING claim

`CreateCodeEffectLog.lean:54` (on `origin/main` `11fcab529`):

> `round8(48 + code_len)`; deployed code is ≤ 32768 (Amsterdam EIP-7907).

Same wrong constant, same cited EIP — but here it documents a **buffer-sizing** assumption rather than a clamp. If any allocation or `.zero` region is dimensioned on "code ≤ 32768" while real deployed code can reach 65536, that is an **overrun**, not a wrong hash, and it would not surface as an FR row.

Please check whether that comment is merely stale prose over a dynamically-computed size, or whether something is actually dimensioned on it — and then **grep the whole tree for the belief, not just the literal**: `32768`, `0x8000`, and prose mentioning EIP-7907 with a wrong bound. A constant that is wrong in one place is usually wrong in several, and fixing one site leaves the class open.

(Note `Dispatch.lean:78` `selfdestructDestroyedAddressCap = 32768` and the `baap_*` / `badf_tmp_values` `.zero 32768` regions are almost certainly unrelated capacities that legitimately share the number — the point is to distinguish those from ones that encode the code-size limit.)

pirapira:
Parking this one — it fails `check-asm-to-program` **standalone on its own branch**, before any merge. I verified `main` and the sibling #11610 are both CLEAN, so this isn't a batch artifact:

```
DRIFT DETECTED:
  [EvmAsm/Codegen/Programs/EvmOpcodesExtcodecopy.lean] extcodecopyAtHeaderStateRootFunction:
  bare local J site ratchet: expected exactly 176 sites, found 171 (48 files / 79 defs);
    update the committed value with a stated reason
```

Two things need your judgment, which is why I'm not fixing them myself:

1. **The fixture is stale.** You raised the cap in both directions — `EvmExtcodecopy.lean` (`li x18, 32768` → `65536`) and the hand-written Program in `EvmOpcodesExtcodecopy.lean` (`.LUI .x5 8` → `16`) — but `scripts/asm-fixtures/extcodecopyAtHeaderStateRootFunction.s` wasn't regenerated to match. I could re-run the rewrite, but the gate says explicitly this *would change the guest/probe images*, so it's emitted-code substance rather than a layout regen and should be yours to confirm.

2. **The J-site ratchet moved down 176 → 171.** `EXPECTED_BARE_J_SITES` at `scripts/asm_to_program.py:275` needs updating, and the gate asks for a stated reason. Plausibly `li 65536` assembling to a bare `LUI` where `li 32768` needed `LUI+ADDI` shifts the site count — but that's a guess and the reason should come from you.

The sibling #11610 (the prose-only `MAX_CODE_SIZE` sweep) is unaffected and I'm landing it separately. Once these two are addressed, drop the `complicated merge` label and it'll be picked up on the next pass.

PR #11610, opened 2026-08-06
docs(codegen): correct stale EIP-7907 MAX_CODE_SIZE prose (65536)

pirapira:
## Summary
Constant sweep after #11608 (EXTCODECOPY length cap 32768→65536).

**Ref:** `origin/main` @ `11fcab529` (pre-#11608 merge; #11608 labelled merge! hands-off).

### Live code-size limit sites
| Site | Kind | Verdict |
|---|---|---|
| `EvmOpcodesExtcodecopy` LUI 8 / length>32768 | **LIVE clamp** | Fixed #11608 → 65536 |
| `EvmExtcodecopy` OOB offset normalize 32768 | **LIVE** (offset past-end) | Fixed #11608 → 65536 |
| `CreateDeployedCodeValid.maxDeployedCodeSize` | LIVE gate | **Already 65536** |
| `CreateInitcodeSizeValid` | LIVE gate | **Already 131072** |

### Stale prose / not live overrun (this PR)
| Site | Why not overrun |
|---|---|
| `CreateCodeEffectLog:54` | Arena = `execCodeEffectLogCap=1MiB`; entry = dynamic `round8(48+code_len)`. Comment only. Gas still bounds total bytes. |
| `WitnessValidation` EIP-7907=32768 docs | Cap is **a2 caller-supplied**; production guest does not call this path (probe-only). |
| `BlockVerdictDataSection` "EIP-170 round8(24576)" | Live `.zero (bsrAccountSlotCap*64+65536)` already lifted. |
| `BlockVerdictDispatchTx` EIP-170 24576 comment | Live guard uses same `+65536` formula. |

### Unrelated 32768 (left alone)
`Dispatch.selfdestructDestroyedAddressCap`, `baap_storage_*` / `badf_tmp_values` `.zero 32768`, `mset_res_cache_valid`, `StorageWriteMap` entry cap, `MptResolveCacheReset` 32KiB window, probe `evm_memory .zero 0x8000`, `ExtcodecopyAtBlockHash` length≤256 probe cap, BlockVerdict debug capacity dumps at 688/736.

### Overrun risk
**No additional live buffer dimensioned on code≤32768** found. CreateCodeEffectLog was the coord-flagged site: dynamically sized + gas total, not a 32KiB slot.

## False-friend instrument note (#11582)
Decoder class names twice named a symptom that was not the defect:
1. `post_delta_missing_from_aw` on 2935 rows → real defect was system-call witness staging / later baseline
2. `nonce_mismatch` on 01308/01315 → real defect was EXTCODECOPY status-6 zero window

Do not treat class labels as diagnoses.

## Do not
No merge from grok; #11608 hands-off.


pirapira:
**coord review: `merge!` applied. Prose-only, verified — and the negative result on my own pointer is the valuable part.**

Checked that this cannot reach the image: the only three added lines containing quotes are inside `--` Lean comments quoting the old prose; no emitted string changed. So no `.text` movement and no layout pins owed.

### My pointer was wrong, and the check is what established it

I flagged `CreateCodeEffectLog.lean:54` ("deployed code is ≤ 32768") as a possible **buffer-sizing overrun** — a defect that would not surface as an FR row. It is not: `execCodeEffectLogCap = 1048576` is a gas-derived **total**, and each entry is `round8(48 + code_len)` computed from the live length. **No fixed 32768 buffer, so no overrun.** Stale prose over a dynamically-computed size, now corrected.

That is the right outcome for a sweep request: the hypothesis was worth stating because an overrun there would have been invisible to the FR instrument, and it was worth *checking* rather than assuming either way.

### The discrimination is what makes this sweep trustworthy

Three more stale-prose sites corrected (`WitnessValidation` EIP-7907 = 32768 on a probe-only path with no production caller; `BlockVerdictDataSection`'s "EIP-170 round8(24576)" where the live `.zero` is already `bsrAccountSlotCap*64+65536`; the matching `BlockVerdictDispatchTx` comment) — and, more importantly, the **unrelated** `32768`s were correctly left alone: `selfdestructDestroyedAddressCap`, the `baap_storage_*`/`badf_tmp` regions, `mset_res_cache`, the `StorageWriteMap` entry cap, `MptResolveCacheReset`, the probe `evm_memory .zero 0x8000`, `ExtcodecopyAtBlockHash`'s ≤256 probe cap, and the `BlockVerdict` debug dumps.

A sweep that changed every occurrence of the literal would have been worse than no sweep. Separating *sites that encode the code-size limit* from *capacities that happen to share the number* is the whole job, and the live-site list (`EXTCODECOPY` fixed in #11608; `maxDeployedCodeSize` already 65536; `CreateInitcodeSizeValid` already 131072) is what shows it was done.

Issue #11591, opened 2026-08-06
structural (#11581 slice 6): flip the 16 readers to AccountWriteMap and delete the AccountState overlay


pirapira:
Slice 6 of #11581 (umbrella). **Gated on all four family slices being complete and the harness reporting an empty disagreement set.**

## Ask

Flip the 16 normal readers to `AccountWriteMap` and delete the `AccountState` overlay.

This is the slice that was attempted first and regressed 142 rows. It is safe only once every
mutation point publishes synchronously — i.e. once the harness from slice 1 reports **zero**
(reader, address, field) disagreements on a full corpus run, not merely on r200.

## Completion evidence to demand

- The harness disagreement set is **empty on the full corpus** (26,104), not a sample. A random
  sample cannot certify this; a clean r200 coexisted with a real FA on #11394.
- `nm` on the linked image shows the overlay symbols gone, plus a `#guard` on absence so the
  retirement cannot regress silently.
- `textSizeBytes` delta reported (a deletion moves every later address, so the sha differs either
  way — the size delta is the linkage evidence for inline arms `nm` cannot see).
- Full-corpus OK/FR/FA/FAULT with base sha and ELF sha256. **FA must be 0.**

⛔ Never migrate a reader onto a structure scheduled for deletion — that manufactures a waypoint
and turns a three-slice lane into six.

## Gate (structural-alignment bar)

Per the maintainer's standing ruling: a structure change that aligns with `execution-specs`
lands on **200 random A/B plus named controls** — the bar is *structure matches the spec* and
*r200 does not deteriorate a lot*, **not** byte-equivalence. An empty flip set is expected and
is not evidence of no-op-ness; alignment is its own justification because it shortens the proofs.

Required in the PR body:
- base sha + candidate ELF sha256 (both), and the r200 flip set **in both directions**
- the named controls for the family (list them; a fix verified only on failing rows has an
  unmeasured downside)
- a spec citation by file:line for the ordering being adopted

⚠️ `output[0:32]` is `new_payload_request_root`, **not** the state root. Value-neutrality
evidence is `succ` agreement plus the flip set — never "roots matched".

Umbrella: #11581

pirapira:
## ⛔ Harness coverage is 4 readers, not 16 — three slices are currently unfalsifiable

A read-only audit of the instrumented reader inventory changes what the #11586 results can and cannot license.

**Instrumented readers: four.**

| id | site | tier |
|---|---|---|
[...]

PR #11612, opened 2026-08-06
fix(codegen): publish MTx sender refund map row

pirapira:
## Summary

The MTx sender-refund path now publishes the settled refund row to the transaction map immediately after the AccountState `create_ether`-equivalent update and before the map-first post-balance read. This matches the execution-spec ordering: `fork.py:1155-1169` applies the sender refund before the BAL-facing state is consumed. The existing caller tags (`a2=17/18/19`) are unchanged.

This is structural alignment, not a verdict change: AccountState remains the execution authority, so a lagging map row is currently verdict-neutral. The map must nevertheless agree before #11591 can make it authoritative.

## Evidence

- Base commit: `2e04928dd6fb2cde918c5469a9bf7a2d860f05de`
- Baseline ELF: `e7e3762e3a0e31a5f2a324030dce0784890de9734edcab6bc15f0844adf8da4e`
- Candidate ELF: `21b6d22c844ea479f2280f1ceafdfccca5dc554c4ec63ae882181bf385e498fe`
- Manifest: `/tmp/codex2-account-reader-ab-base-forward/manifest.tsv` (sha256 `43a41e1f321dc376698ab220711140b6551b8524a400433c43702e70de3c12ac`), 200 cases, seed `11586`
- Runner: `scripts/spike/spike_run` rebuilt from source at the base; binary sha256 `529504d986b06aa5b40c8e9fc351bd88831f4a30dc28b4ef92422d54ba9f2844`; source sha256 `757039f88382d7dac5fb7562fb6c61e48d454c4dccb445606e215cbce2e798cf`
- Enabled agreement harness: baseline mismatches `142` -> candidate `0`; all 142 baseline events were reader `19/tx`
- Balance probe invocations: `1090` -> `1090`; nonce probe invocations: `55` -> `55`
- `instrument_events`: `0` -> `0`
- Candidate/base output differences: `0`

The unchanged invocation counts show that the reader still executes; the mismatch count reaches exact zero rather than merely decreasing. The zero output-difference result is expected for this structural change on this corpus, not evidence that the publication is unnecessary.

During rebase validation, an intermediate candidate accidentally dropped the existing caller tags (`a2=17/18/19`). I caught it by comparing the emitted source before accepting the apparent zero; without those tags, reader attribution would have been silently invalid even though invocation counts remained unchanged. The final candidate preserves all three tags.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdictMtxRuntime.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictMtxRuntime.lean
@@ -102,10 +102,17 @@ private def blockVerdictMtxRecordSenderRefund : String :=
   -- BAL row.  The BAL producer below deliberately re-reads this final state.
   "  la t0, sttc_nonce; ld a3, 0(t0); addi a4, a3, 1; la a0, bv_mtx_sender_addr; la a1, bv_pending_upfront_sender_pre; la a2, bv_pending_upfront_sender_post; jal ra, account_state_record_nonstorage\n" ++
   "  bnez a0, .Lbv_mtx_bail\n" ++
-  "  la a0, bv_mtx_sender_addr; la a1, bv_pending_upfront_sender_post; li a2, 19; jal ra, account_state_latest_balance\n" ++
-  "  beqz a0, .Lbv_mtx_bail  # the preceding AccountState credit must be observable\n" ++
+  -- Publish the post-body refund to the transaction map at the mutation point.
+  -- `create_ether` has already updated AccountState above; this companion records
+  -- the same settled pre/post pair without a second AccountState append.  The
+  -- publication must precede the post read below so a map-first reader observes
+  -- the refund instead of the pre-refund row.  It is deliberately after body
+  -- restoration and therefore survives an ordinary body revert, matching
+  -- `fork.py:1155-1169`.
   "  la t0, sttc_nonce; ld a3, 0(t0); addi a4, a3, 1; la a0, bv_mtx_sender_addr; la a1, bv_pending_upfront_sender_pre; la a2, bv_pending_upfront_sender_post; jal ra, record_nonstorage_effect_after_account_state\n" ++
   "  bnez a0, .Lbv_mtx_bail\n" ++
+  "  la a0, bv_mtx_sender_addr; la a1, bv_pending_upfront_sender_post; li a2, 19; jal ra, account_state_latest_balance\n" ++
+  "  beqz a0, .Lbv_mtx_bail  # the preceding AccountState credit must be observable\n" ++
 ".Lbv_mtx_sr_done:\n"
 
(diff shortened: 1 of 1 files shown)


pirapira:
**coord review: `merge!` applied.** `+9/−2` in one file, and I checked the one thing that could make this look like a fix without being one.

### Why this is a real ordering fix and not a silenced probe

The edit moves `account_state_latest_balance` (reader19) to **after** `record_nonstorage_effect_after_account_state`. Read carelessly that is *"the instrumented read moved later"*, which would silence the disagreement without fixing anything. It is not, for two reasons:

1. **Reader19's read is functional, not instrumentation.** It gates `beqz a0, .Lbv_mtx_bail` — the routine genuinely requires the post balance to be observable and bails otherwise. So the reordering changes what a *real consumer* sees, not what a probe sees.
2. **Publication is now adjacent to the mutation.** `account_state_record_nonstorage` (the `create_ether` refund) is immediately followed by the map publication, with nothing in between. So the map carries the refund *at the mutation point* for every consumer, which is the property #11581's lane needs — not merely for this one callsite.

Register safety: the publication line establishes its own `a0`–`a4` (`la t0, sttc_nonce; ld a3; la a0/a1/a2`), so it does not depend on anything the moved read left behind.

Spec anchor is right: `fork.py:1155-1169` computes `gas_refund_amount` and calls `create_ether(tx_state, sender, …)` **after** `process_message_call`, so the publication is deliberately post-body-restoration and survives an ordinary body revert. That is the constraint `AccountWriteMap.lean:875-880` exists for — sender gas must **not** be unwound by a body revert — and it is respected.

### Measurement

Base `2e04928dd`, both ELF shas stated, same 200-case manifest and seed 11586:

| | base | candidate |
|---|---|---|
| reader19/tx mismatches | **142** (all of them) | **0** |
| balance probe invocations | 1090 | **1090** |
| nonce probe invocations | 55 | **55** |
| instrument_events | 0 | 0 |
| output_differences | 0 | 0 |

**Exact zero, and invocations held.** Both halves matter: the prediction was exact zero rather than a reduction (the delta was the full `gas_refund_amount` on 165/165 with no partial cases, so a partial fall would have refuted the mechanism), and unchanged invocations rule out the failure where a publication moves somewhere the read no longer happens.

`output_differences=0` means this is **verdict-neutral today, which is expected** — `AccountState` is still the execution authority, so a lagging map row cannot change a verdict. The value is structural: **the map cannot become authoritative while its rows lag their own mutation**, so this removes a blocker to #11591 rather than closing an FR row.

Good that the tag-drop near-miss is in the body. A rebase briefly dropped the `a2` 17/18/19 caller tags, which would have produced zero reader19 mismatches because nothing was *labelled* reader19 — a third variant of the silenced-probe shape, caught before the numbers were treated as valid. Worth more recorded than omitted.

Issue #11615, opened 2026-08-06
SpecRef bug: _decode_header's getB drops the FixedBytes length checks rlp.decode_to enforces (13 byte fields)


dhsorens:
Found while fixing #11513. That issue's scope note says the byte fields are fine:

> `getB` fields are unaffected — those are byte strings either way, and `items.mapM rlpBytes?` already sends `.list` to `none`.

That is **wrong for the fixed-width ones**, and it is the same shape of defect as #11513: the port is looser than the Python.

## The divergence

`_deserialize_to_bytes` does not just check "is it bytes" — it **constructs the annotated type** (`ethereum_rlp` 0.1.6, `rlp.py:254-262`):

```python
def _deserialize_to_bytes(class_, value):
    if not isinstance(value, bytes):
        raise DecodingError("invalid bytes")
    try:
        return class_(value)
    except ValueError as e:
        raise DecodingError from e
```

and `FixedBytes.__new__` enforces the length (`ethereum_types` 0.4.1, `bytes.py:29-37`):

```python
def __new__(cls, *args, **kwargs):
    result = bytes.__new__(cls, *args, **kwargs)
    if len(result) != cls.LENGTH:
        raise ValueError(f"expected {cls.LENGTH} bytes but got {len(result)}")
    return result
```

The port's `getB` is a bare `bs.getD i []` (`SpecRef/Stateless.lean`, `mkHeaderFields`) — no width check. So a header whose `state_root` is 31 bytes, or whose `bloom` is 200, is an error in the reference and a silently-accepted value here.

## Scope: 13 fields

| idx | field | annotation | required bytes |
|---|---|---|---|
| 0 | `parent_hash` | `Hash32` = `Bytes32` | 32 |
| 1 | `ommers_hash` | `Hash32` | 32 |
| 2 | `coinbase` | `Address` = `Bytes20` | 20 |
| 3 | `state_root` | `Root` = `Hash32` | 32 |
| 4 | `transactions_root` | `Root` | 32 |
| 5 | `receipt_root` | `Root` | 32 |
| 6 | `bloom` | `Bloom` = `Bytes256` | **256** |
| 13 | `prev_randao` | `Bytes32` | 32 |
| 14 | `nonce` | `Bytes8` | 8 |
| 16 | `withdrawals_root` | `Root` | 32 |
| 19 | `parent_beacon_block_root` | `Root` | 32 |
| 20 | `requests_hash` | `Hash32` | 32 |
[...]

Issue #11575, opened 2026-08-05
proof: fork header_extract_number's pattern to its 8 mechanical siblings (header_extract_logs_bloom, extra_data_length, six chain_validate_*)


pirapira:
> 🤖 Generated by Claude Code (strategist pass over execution-specs ↔ SpecRef ↔ RISC-V correspondence, 2026-08-06).

## The move

`docs/leaf-routine-targets.md` row 9 (`header_extract_number`) landed as `.proven` via #11351 (`header_extract_number_spec_within`, registered in `EvmAsm/Progress/Routines.lean:220`). The table's closing note says what that buys:

> Row 9 is the representative of a **family**: `header_extract_logs_bloom`, `header_validate_extra_data_length`, and **six of the seven** `chain_validate_*` routines have the same shape (verified RLP callees only, one `_decode_header` field or one header rule each). Once row 9's pattern exists, those **eight** siblings are mechanical forks.

The pattern now exists. This issue is the fork: eight sibling triples, each against its `_decode_header` field or header rule in `SpecRef/Stateless.lean` / `SpecRef/SeamShell.lean`, each independently closable — a rare batch of low-risk `.proven` rows that grows the routine registry from 19 proven symbols meaningfully in one push.

## Scope

- `header_extract_logs_bloom`, `header_validate_extra_data_length`, and the six forkable `chain_validate_*` routines (enumerate from `GuestAddrs`; the leaf-targets doc verified their transitive callees are all in the verified set).
- **Excluded**: `chain_validate_post_merge_full` — explicitly *not* a fork (only a byte-identity theorem exists, `ChainValidatePostMerge.lean:608`); filed separately.
- Each landed row: `Routines.lean` witness + `Correspondence.lean` row (family `header`), per the #11335 cross-registry gate.

## Caveats carried over from #11351

- ⚠️ #11513: the port's `getN` is more lenient than CPython's `rlp.decode_to` on all nine numeric fields. Each sibling touching a numeric field inherits the #11493 verdict-schema resolution (guest stricter than port, matches CPython) — cite it per row rather than re-litigating.
- ⚠️ #11461: avoid propagating the 2^64-derived step-bound factor into eight new places; derive bounds from data where the pattern allows.

dhsorens:
## Row 1 landed (#11624). Recalibrating the other seven, because this issue's framing was wrong in two ways.

[...]

PR #11617, opened 2026-08-06
fix(specref): re-impose rlp.decode_to's typed checks on _decode_header's numeric fields

### LINKED ISSUE #11513: SpecRef bug: _decode_header's getN drops the width and canonicality checks that rlp.decode_to enforces (9 numeric fields)

dhsorens:
Closes #11513.

`_decode_header`'s port built every numeric header field with a plain
`bytesBEtoNat`, dropping the per-field checks `rlp.decode_to` performs. This
re-imposes them — **exactly**, which turns out not to be what #11513 asked for.

## The correction #11513 needed

The issue, and the note #11514 merged onto the registry row, both asserted the
guest's two rejections (over-width, leading-zero) match `rlp.decode_to`, so that
"only the port is lenient". **Only one of them does.** The two checks come from two
different places with different scope:

| check | source | scope |
|---|---|---|
| leading zero | `_deserialize_to_uint` (`ethereum_rlp` 0.1.6, `rlp.py:270-271`) | **every** uint field |
| width | the target type's `from_be_bytes` (`ethereum_types` 0.4.1) | **fixed-width types only** |

`FixedUnsigned.from_be_bytes` raises on an over-wide buffer (`numeric.py:566-577`);
`Uint.from_be_bytes` is a plain `int.from_bytes` with no length check
(`numeric.py:523-528`). So of the nine numeric fields:

| idx | field | annotation | ref max bytes |
|---|---|---|---|
| 7 | `difficulty` | `Uint` | **unbounded** |
| 8 | `number` | `Uint` | **unbounded** |
| 9 | `gasLimit` | `Uint` | **unbounded** |
| 10 | `gasUsed` | `Uint` | **unbounded** |
| 11 | `timestamp` | `U256` | 32 |
| 15 | `baseFeePerGas` | `Uint` | **unbounded** |
| 17 | `blobGasUsed` | `U64` | 8 |
| 18 | `excessBlobGas` | `U64` | 8 |
| 22 | `slotNumber` | `U64` | 8 |

`bpo5/blocks.py` carries identical annotations at 0–20, so one table serves both arms.

**Consequence:** the guest's 8-byte `tooLong` rejection of `number` is a false
reject against **CPython**, not just against the port. Bounding the port at 8 to
make the row green would make the port *stricter* than the Python — a port defect in
the opposite direction — to hide a guest defect. So the port matches the reference,
and the guest FR is recorded as what it is.

There is a `#guard` pinning that a nine-byte `number` is **accepted**, specifically
so this cannot be "fixed" later by tightening it.

## Changes

**`SpecRef/Stateless.lean`** — `numericFieldWidths` carries the table;
`checkNumericFields` runs each field through `decodeItemScalar`, the **existing**
port of `_deserialize_to_uint` + `from_be_bytes` whose `Option Nat` width argument
already models the `Uint`-vs-`FixedUnsigned` split (no new decoder). `mkHeader`
becomes `mkHeaderFields` (the field assignment) and `_decode_header` runs both.
Public signature unchanged, and its one caller (`validate_headers`) propagates the
rejection with no edit.

**`decode_header_inv` generalized.** It exposed one field (`number`) and is the only
door into `_decode_header` (`mkHeaderFields`/`rlpBytes?` are private). It now yields
`hdr = mkHeaderFields (bs.length == 23) bs` — so *any* field follows by `rfl` — plus
the two typed checks indexed by field. This is what #11575's eight sibling bridges
consume instead of each re-deriving them.

**`SpecRef/Transactions.lean`** — `decodeItemScalar_value` and
`decodeItemScalar_checks`. They live here because `txErr` is private, so a caller
cannot rule out the failure branches. `_value` is what makes assigning `bytesBEtoNat`
beside a passing check *provably* equivalent to using the checked decoder's result,
rather than a restatement the clause table has to assume.

**`HeaderExtractNumberBridge.lean`** — `hcanon` is discharged from the port's own
check and dropped. The remaining restriction is now `hfits : hdr.number < 2 ^ 64`,
phrased over the **value**, which is only possible *because* canonicality is now
enforced: with no leading zeros, "at most eight bytes" and "< 2^64" coincide. The
old docstring said these restrictions "cannot be phrased over the value" — true then,
not now.

**`EL/RLP/Properties.lean`** — `Nat.length_le_of_canonical_lt`, the converse of
`fromBytesBE_lt` under canonicality. It needs canonicality: `[0,0,1]` decodes to
`1 < 256^1` while being three bytes long.

**Registry** — row regraded `.domainRestricted` → `.stricter` (the first in the
registry's history) and `portDefect` cleared. `verdict_counts` and
`port_defect_count` updated. Read the nonzero `stricter` as the schema finally
discriminating: an FR wore `domainRestricted` until #11493, then was attributed to
the port. The arity restriction is unchanged and recorded in the note.

**Docs** — `4ch8f-specref-port.md`'s "Known simplifications" item 1,
`BlocksRlp.lean`'s round-trip note (whose caveat becomes vacuous), and
`spec-correspondence.md`'s worked example, which taught the wrong inference. §6a
gains `ethereum_types`: whether a field is width-bounded *at all* is decided there,
and a local `.venv` was found carrying 0.1.5/0.3.0 against a lock of 0.1.6/0.4.1.

## Verification

- `lake build` clean; no `sorry`, no new `maxHeartbeats`/`maxRecDepth`.
- `#guard`s: the existing header guards still pass unchanged (`mkTestHeaderBytes`
  fills fields with the canonical empty encoding), plus new ones for leading-zero on
  both a `Uint` and a `U64` field, over-width `U64`, `timestamp` at 32 (accept) and
  33 (reject), and the load-bearing accept of a nine-byte `number`.
- `scripts/check-forbidden-tactics.sh`, `scripts/check-file-size.sh`,
  `scripts/check-spec-refs.sh` (44 citations resolved), `scripts/check-axioms.sh`,
  `scripts/check-registry-crosscheck.sh`, `scripts/progress-report.sh --check`.
`#print axioms`, fully qualified:

```
'EvmAsm.Codegen.HeaderExtractNumberSpec.header_number_of_decode' depends on axioms: [propext, Classical.choice, Quot.sound]
'EvmAsm.Stateless.SpecRef.decode_header_inv'                     depends on axioms: [propext, Classical.choice, Quot.sound]
'EvmAsm.Stateless.SpecRef.getNChecked_value'                     depends on axioms: [propext]
'EvmAsm.Stateless.SpecRef.getNChecked_checks'                    depends on axioms: [propext, Quot.sound]
'EvmAsm.Stateless.SpecRef.decodeItemScalar_value'                depends on axioms: [propext]
'EvmAsm.Stateless.SpecRef.decodeItemScalar_checks'               depends on axioms: [propext, Quot.sound]
'EvmAsm.EL.RLP.Nat.length_le_of_canonical_lt'                    depends on axioms: [propext, Classical.choice, Quot.sound]
'EvmAsm.Progress.Correspondence.verdict_counts'                  does not depend on any axioms
'EvmAsm.Progress.Correspondence.port_defect_count'               does not depend on any axioms
```

- `scripts/eest-specref-check.sh --all` — the one harness that would surface a
  false-reject on real chain data (this change only *adds* rejections). Fetched the
  pinned `tests-zkevm@v0.6.2` fixtures for it; result posted as a comment below.

⚠️ Not run locally, stated rather than claimed: `check-asm-to-program.sh` and
`check-build-units-link.sh` `exit 0` silently without the RISC-V toolchain. This PR
changes no emitted bytes, so no EEST A/B or layout regen applies.

## Follow-up this surfaced

#11615 — the same class of leniency on the **fixed-width byte** fields, which #11513
dismissed as "byte strings either way". `_deserialize_to_bytes` constructs the
annotated type and `FixedBytes.__new__` enforces `LENGTH`, so the reference does
length-check `Hash32`/`Address`/`Root`/`Bloom`/`Bytes32`/`Bytes8`. Deliberately **not**
folded in here: unlike the numeric sweep, that check is arity-dependent — a
`length = 32` check on index 21 would reject every 21-field `bpo5` header, because the
absent field reads as `[]`. It is load-bearing for #11575's `header_extract_logs_bloom`
row, whose triple has an `a0 = 2` arm meaning exactly "field-6 length ≠ 256".


pirapira:
**coord review: `merge!` applied. I verified both spec claims directly against the vendored sources, and the correction to #11513's premise holds exactly.**

### The two checks do come from different places with different scope

`ethereum_rlp/rlp.py`, `_deserialize_to_uint`:

```
if len(decoded) > 0 and decoded[0] == 0:
    raise DecodingError("non-canonical integer")
return class_.from_be_bytes(decoded)
```

So the **leading-zero** rejection is imposed on *every* uint field, `Uint` and `FixedUnsigned` alike, before the type is consulted.

`ethereum_types/numeric.py`, `FixedUnsigned.from_be_bytes`:

```
bits = cls.MAX_VALUE._number.bit_length()
byte_count = (bits + 7) // 8
if len(buffer) > byte_count:
    raise ValueError()
```

So the **width** rejection is a property of the *target type*, and `Uint` — documented as "unsigned integer of arbitrary size", with no `MAX_VALUE` — cannot carry it.

**So #11513's premise, and the registry note #11514 merged on top of it, were wrong: the two guest rejections are not a matched pair.** Leading-zero matches the reference; over-width does not, for any field annotated `Uint`. That is six of the nine numeric header fields — `difficulty`, `number`, `gasLimit`, `gasUsed`, `baseFeePerGas` — where the reference is unbounded.

⭐ **The consequence is the valuable part, and it is a follow-up rather than this PR:** the guest's `tooLong` rejection on those fields is **a check the spec does not have**, i.e. a false reject on a header the reference accepts. That is precisely the class we are required to remove — otherwise equivalence needs hash-collision assumptions. Please make sure that lands as its own issue with the field list, since it is a guest change and this PR is port-side.

### Scope check

`+515/−116`, and the only file under `Codegen/` is `HeaderExtractNumberBridge.lean`, whose changes are **docstring prose** — the `tooLong`/`noncanonical` discussion, now corrected to say they have different status rather than being a matched pair. **No emitted string changed, so no `.text` movement and no layout pins owed.** The rest is `SpecRef`, `Correspondence.lean`, `RLP/Properties.lean`, `PROGRESS.md` and docs.

CONFLICTING is main having advanced; that is the merge actor's to resolve, not yours.

Good that the correction is written into the docstring at the site rather than only in the PR body — the previous wording asserted the matched-pair reading, and that is exactly the kind of stale comment that outlives the issue thread.

dhsorens:
Thanks — filed as #11620, with the field list.

**One correction to the list in your review, in your favour:** you wrote "six of the nine" and then named five (`difficulty`, `number`, `gasLimit`, `gasUsed`, `baseFeePerGas`). Six is right; the missing one is **`timestamp`**. It is the only field bounded on *both* sides but at different widths — `U256` admits 32 bytes, the guest admits 8 — so it does not show up when scanning for unbounded `Uint`, and it is the field `chain_validate_increasing_timestamps` reads. #11620 lists all six with the annotation line numbers.

**A question I did not guess at, raised in #11620.** Your ruling here ("the class we are required to remove") and your note on #11513 ("evm-asm has project-wise assumptions; the top-theorem will assume these in the input are small enough to fit the bit-width") point at two different guest diffs:

1. keep the rejection and let the top theorem's input assumption make it unreachable — every intermediate row then carries the precondition, and must state it *explicitly in the theorem* or the assumption goes silent, which is worse than an FR because an FR at least gets counted;
2. remove the rejection — which needs the replacement behaviour chosen (saturate / wrap / low 8 bytes), each its own divergence.

I read them as compatible rather than contradictory, but they are not the same change, so #11620 asks which you want rather than picking. I briefly regraded this row to `domainRestricted` locally on the strength of the #11513 note, then reverted it — the branch is exactly what you reviewed, since `merge!` is applied and late pushes can orphan it.

Leaving the CONFLICTING state alone per your note.

The `#guard` that a nine-byte `number` is **accepted** by the port is the durable part of this: whichever way #11620 goes, it stops the guest's bound being re-imported into the port as a "fix".

dhsorens:
Pushed one commit acting on your #11620 ruling, because it changes a grade that would otherwise land wrong.

> The project-wide assumptions give the guest freedom to choose its behavior. I prefer reject on out of assumption inputs

That settles it the other way from the review above: the `tooLong` rejection is the **preferred** behaviour, not "the class we are required to remove". So `.stricter` is wrong — it asserts a false reject on input the project claims to handle, and the project explicitly does not claim to handle out-of-assumption header fields. The row goes back to `.domainRestricted` in its **ABI-precondition** flavour, and `verdict_counts` is now unchanged from `main`.

**The port-side finding is untouched** — canonicality was a real port defect, it is fixed, `portDefect` stays cleared, and the nine-byte-`number`-is-accepted `#guard` stays. Only the grading of the width axis moved.

I pushed rather than deferring because the PR was already `DIRTY`, so this cannot orphan a merge that was otherwise going to happen — and a known-wrong verdict landing seemed worse than a late push. Say the word if you would rather have it as a separate PR and I will split it back out.

Two things worth flagging:

- **The `timestamp` correction still stands** and is independent of the grading: it is `U256`, so the reference bound is 32 bytes against the guest's 8. It is the sixth field, and the one most easily missed since it is bounded on *both* sides at different widths.
- **What keeps the precondition reading honest** is that it is explicit in the statement — `hfits : hdr.number < 2 ^ 64`. I have written that into the row note and the method doc as a rule: grading a restriction as a precondition while leaving it implicit in the guest's behaviour is worse than recording an FR, because an FR at least appears in `verdict_counts`. Your "second top theorem saying the guest rejects out of assumption header values" is recorded on #11620 as the follow-up that would upgrade this from permitted-freedom to a proven property.

Also stacked #11618 → #11615 on top of this: the same class of leniency on the **fixed-width byte** fields (`_deserialize_to_bytes` constructs the annotated type, and `FixedBytes.__new__` enforces `LENGTH`). Kept separate because that check is arity-dependent — sweeping index 21 would reject every 21-field `bpo5` header.

Issue #11613, opened 2026-08-06
mpt_bounded_storage_root wired live against its own 'disconnected until insert/delete KATs' contract


pirapira:
## Defect (one sentence)

`MptBoundedSort.lean:803-806` documents that `mpt_bounded_storage_root` supports an **exact-replacement subset** and **intentionally remains disconnected from the live verdict** until insert/delete/canonical-collapse cases have comparable KATs — **and the live path is connected anyway** (`BalAccountApplyPostFields` `.Lbaap_multi_apply_call` → `jal mpt_bounded_storage_root`, fallback to legacy only on a0≠0).

That is the defect: a component whose own documentation says it is not ready for the live verdict is wired to the live verdict. The wrong root is the symptom; the wiring against its own stated contract is the cause (and the kind of thing that recurs if nobody looks for it).

## Symptom (worked example)

Fixture class `precompile_absence` (31_bytes), corpus **12832 / 12833 / 12834**.

- Primary `655600d1363355fcdbaf27b37b3f4e88d41b3221`
- Pre: 239 high keys `0x0100000000000000NN → 0xba5e`; post: 239 low `0xNN → 0x01`
- After #11604 baseline: 478 SW deltas (239 insert + 239 delete)
- `mpt_bounded_storage_root` → **a0=0 success**, guest sroot `31bf2ebd…`
- Expected post `1bc96535…`; pre `1e872fd4…`; guest matches **none**
- Terminal fail bv=1 (#11547 state-root byte compare)

## What is NOT the defect

| candidate | verdict |
|---|---|
| tier-3 SLOAD status 5 on high keys | **EXONERATED** — offline MPT 239/239 FOUND; live high keys status 0 |
| missing baseline on block upsert | **necessary not sufficient** — #11604; 23274 PASS; 12832 still FR |
| fitted descriptor-count gate (e.g. ≥64) | **REJECTED** — see binary search below |

## Q1: threshold is fitted (binary search)

Count-only "legacy when count≥N" on 12832 (478 descs):
- PASSes for every N ∈ **[6, 478]** (routes to legacy)
- FAILs only at N=479 (stays on bounded)

The break **is** the fixture descriptor count, not a builder capacity. Capacities are depth-derived: frames=65, sort stack=1024, cache slots=65. Nothing is count-derived. 64 was an arbitrary point in (5, 478].

## Complementary wrongness (second finding → own issue)

[...]

PR #11621, opened 2026-08-06
fix(codegen): resolve original for seeded system storage rows

pirapira:
## Summary

The pre-user modeled-system seed path used a sentinel `a3=0` baseline when calling `storage_writes_block_upsert`. On a seeded address/slot hit, the upsert preserves the existing row baseline, so the later state-root comparison can treat a nonzero authenticated-parent value as zero and omit a real change.

This change routes seed-only rows through the existing authenticated `slot_at_header_state_root` lookup, then upserts the resolved original at descriptor +64. It keeps seed-only behavior side-log/BAI-free. The resolver body (StateCompose.lean:565-625) has no `account_read_record` or `storage_read_record` call, so this route cannot add a read-set/BAL side effect.

This is a latent-defect removal and spec-alignment change, not a current-main FR repair. The old 157-row `post_unusable_multiblock` bucket was measured on stale artifact `a6c634e95`; all 157 rows already pass on current main `3ce5f22bd`.

## Evidence

Prediction before measurement: the random-200 and named controls have empty flip sets, because this path only changes seeded system-storage baselines and current main has no remaining failures in the historical bucket.

- Shared manifest: `/tmp/11582-selected.manifest.tsv` (157 historical rows plus five controls) and `/tmp/11582-r200-controls.manifest.tsv` (random-200 plus those controls).
- Base commit `3ce5f22bd`; base ELF SHA256 `8dd2c14e6d97b62de02f4a4875f97cc03cb4b019eaf730f9afe862c1716f7ab1`.
- Candidate commit `441ba5a0a`; candidate ELF SHA256 `a3f24ddd5704b9e9f8f8f780cbf6bfc945b8e6c716f1b2ac64172ae382691bab`.
- On the shared 162-row manifest, the stale reference ELF `26016e2ca3a41ba54ca7f7a1dca9f48aaf9bd199ed96a129039e185d8383c51d` gives 158 FR / 4 OK; freshly rebuilt current-main base gives 162/162 OK. The historical bucket is therefore genuinely retired by intervening merges, not a manifest mismatch.
- Historical 157 rows + five named controls: 162/162 OK on both current-main legs, FA=0, FR=0, no flips.
- Deterministic random-200 + five named controls: 205/205 OK on both legs, FA=0, FR=0, no flips.
- Named controls: 03736, 03737, 11734, 21353, 23274.
- Seeded-row BAL length identity (output +128 rebuilt, +136 supplied; 256-byte captures), both legs: 04457 263/263, 23094 815/815, 23206 824/824.
- `.text` moved `0x56910 -> 0x56924` (+0x14); `.data` and `.bss` are unchanged.

The change fixes no currently-failing row on current main; the evidence establishes no deterioration while making the seed path spec-shaped.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockVerdictSystemStorageCapture.lean
+++ b/EvmAsm/Codegen/Programs/BlockVerdictSystemStorageCapture.lean
@@ -198,6 +198,10 @@ def appendModeledSystemStorageTupleRowsFunction : String :=
   "  # the already materialized descriptor post-value for the comparison.\n" ++
   "  sd zero, 64(s2); sd zero, 72(s2); sd zero, 80(s2); sd zero, 88(s2)\n" ++
   ".Lamsr_original_ready:\n" ++
+  "  # The seed-only caller shares the authenticated-original lookup, but it\n" ++
+  "  # must not publish a duplicate side-log/BAL row.  Keep this dispatch here\n" ++
+  "  # so the resolver's zero-on-miss result is also the map baseline.\n" ++
+  "  la t0, bv_system_storage_map_seed_only; ld t0, 0(t0); bnez t0, .Lamsr_seed_original_ready\n" ++
   "  # Match the ordinary storage emitter's pre!=post rule.  Equal parent/current\n" ++
   "  # values still update the execution map, but publish no BAI-0 change row.\n" ++
   "  ld t0, 64(s2); ld t1, 96(s2); bne t0, t1, .Lamsr_emit_change\n" ++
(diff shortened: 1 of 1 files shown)


pirapira:
**coord review: `merge!` applied.** The change is exactly the narrow fix agreed: seed-only mode now routes through the authenticated-original resolver and passes `addi a3, s2, 64` — a pointer to the resolved baseline — where it previously passed `li a3, 0`. #11604's HIT-leaves-earliest-baseline rule is untouched, and no new validity bit was needed.

### Evidence, in the order that matters

- **Length identity first**, before any verdict interpretation: seeded-row representatives `04457` 263/263, `23094` 815/815, `23206` 824/824 — `rebuilt_len` landing *exactly* on `supplied_len` on both legs. This fix necessarily **adds** entries to the state-root comparison (a real baseline makes previously-equal comparisons differ), and over-adding is invisible to a flip count. Exact equality is the confirmation form; "closer" would not have been.
- **Resolver side-effect ruled out structurally**, not by measurement: `slot_at_header_state_root` (`StateCompose.lean:565-625`) has no `account_read_record` or `storage_read_record` — read from the callee body, which is the #11551 lesson applied correctly.
- **Random-205 A/B plus the named controls** `03736`, `03737`, `11734`, `21353`, `23274` — the set #11604 protected, which is where a baseline change would regress.
- `.text` `+0x14`, `data`/`bss` unchanged. Pins owed; the actor's.

### The historical control is what makes the write-up honest

Same 162-row manifest, two ELFs: stale `26016e2c` gives **158 FR / 4 OK**; freshly built main `3ce5f22bd` gives **162/162 OK**. So the 157-row `post_unusable_multiblock` bucket — **62% of the code-1 residual** — is already retired by tonight's merges, and the census that named it was accurate for the tree it was taken on. Stating both numbers rather than only the all-pass result pre-empts the obvious objection that the manifest simply mismatched.

**This PR therefore fixes no currently-failing row, and says so.** It lands as latent-defect removal plus spec alignment on the standing bar — structure matching the spec, no deterioration — which is the correct basis for it.

### ⚠️ Follow-up, not a blocker

The new comment says *"the resolver's zero-on-miss result is also the map baseline"*. For a storage slot, absent-in-trie genuinely means zero, so zero-on-**miss** is correct semantics. But a resolver **failure** is not absence — grok found tier-3 returning status 5 on paths the witness supports (#11613 territory). If a non-zero resolver status also lands as a `0` baseline, the original defect returns in a narrower form.

Please confirm on the issue whether `.Lamsr_original_resolve` distinguishes *resolved-zero* from *failed-to-resolve*, and if it does not, file it. That is the same question grok raised for tier-3 — whether a miss may populate an original at all or must fail closed — and the two threads converge here.

PR #11623, opened 2026-08-06
fix(specref): re-impose FixedBytes length checks on _decode_header's byte fields

### LINKED ISSUE #11615: SpecRef bug: _decode_header's getB drops the FixedBytes length checks rlp.decode_to enforces (13 byte fields)

dhsorens:
Closes #11615. **Stacked on #11617** (base `work/11513-decode-header-typed-checks`), which did the numeric half.

#11513's scope note dismissed these fields:

> `getB` fields are unaffected — those are byte strings either way, and `items.mapM rlpBytes?` already sends `.list` to `none`.

That is wrong for the fixed-width ones. `_deserialize_to_bytes` does not merely check "is it bytes" — it **constructs the annotated type** (`ethereum_rlp` 0.1.6, `rlp.py:254-262`), and `FixedBytes.__new__` raises when the length is wrong (`ethereum_types` 0.4.1, `bytes.py:29-37`). So the reference length-checks `Hash32`/`Address`/`Root`/`Bloom`/`Bytes32`/`Bytes8`, while the port's `getB` was a bare `bs.getD i []`.

## Scope: 13 fields

| idx | field | annotation | bytes |
|---|---|---|---|
| 0, 1 | `parent_hash`, `ommers_hash` | `Hash32` = `Bytes32` | 32 |
| 2 | `coinbase` | `Address` = `Bytes20` | 20 |
| 3, 4, 5 | `state_root`, `transactions_root`, `receipt_root` | `Root` = `Hash32` | 32 |
| 6 | `bloom` | `Bloom` = `Bytes256` | **256** |
| 13 | `prev_randao` | `Bytes32` | 32 |
| 14 | `nonce` | `Bytes8` | 8 |
| 16, 19, 20, 21 | `withdrawals_root`, `parent_beacon_block_root`, `requests_hash`, `block_access_list_hash` | `Root`/`Hash32` | 32 |

Aliases: `ethereum/state.py:43-44`, `crypto/hash.py:19`, `amsterdam/fork_types.py:34`.

**`extra_data` (12) is deliberately excluded** — plain `Bytes`, genuinely unbounded at decode time. Its ≤32 rule is a `validate_header` clause, not a decode one, and there is a `#guard` pinning that a 40-byte `extra_data` still decodes.

## ⚠️ Why this is not a copy of #11617's shape

The numeric sweep could run unconditionally because a missing field reads as `bs.getD i [] = []`, and `[]` passes every uint check (canonical, length 0 ≤ any bound). **`[]` does not pass `length = 32`.** Index 21 exists only in the 23-field arm, so sweeping it over a 21-field `bpo5` header would reject **every previous-fork block**.

Hence the arity split: `fixedBytesFieldWidths` (indices 0–20, both arms) and `currentForkBytesFieldWidths` (index 21, current fork only). A `#guard` pins that a 21-field header still decodes with `isCurrentFork = false` — that is the regression this could have caused, and it is the reason this is a separate PR rather than three more lines in #11617.

## The `#guard`s had to change, and that is the finding

`mkTestHeaderBytes` filled every field except 0 and 3 with `.bytes []`. The port accepted that only *because* `getB` had no width check — so the existing header guards were exercising headers `rlp.decode_to` would have rejected outright. They now fill each fixed field to its annotated width (`testFieldWidth`). New guards cover: short and long `Root`, `Address` at 32 (reject) and 20 (accept), `Bloom` at 255 (reject) and 256 (accept), `nonce` at 32 (reject), `extra_data` at 40 (accept), and the 21-field arm.

## Why it matters beyond fidelity

It is load-bearing for #11575's `header_extract_logs_bloom` row, whose triple has a three-way postcondition with an `a0 = 2` arm meaning exactly "the field-6 length was ≠ 256" (`HeaderExtractLogsBloomSpec.lean:343`). With the port checking `Bloom`'s width, a successful `_decode_header` makes that arm unreachable and the bridge proves the success arm outright; without it, the bridge carries a `hdr.bloom.length = 256` hypothesis that reads like a guest restriction but is really this port gap in disguise — the same misattribution #11493 spent an issue unpicking. Same for `header_extract_state_root`, `..._receipts_root`, `..._withdrawals_root`, `header_validate_parent_hash`.

## Verification

- `lake build` clean (4821 jobs), no `sorry`, no new heartbeat/recursion options.
- `scripts/check-axioms.sh` — 137 witnesses, classical only.
- `check-forbidden-tactics.sh`, `check-file-size.sh`, `check-spec-refs.sh` (44 resolved), `check-progress.sh`, `check-registry-crosscheck.sh`.

```
'EvmAsm.Stateless.SpecRef.decode_header_inv'                depends on axioms: [propext, Classical.choice, Quot.sound]
'EvmAsm.Stateless.SpecRef.decodeItemFixedBytes_inv'         depends on axioms: [propext, Quot.sound]
'EvmAsm.Codegen.HeaderExtractNumberSpec.header_number_of_decode' depends on axioms: [propext, Classical.choice, Quot.sound]
```

⚠️ Not run locally, stated rather than claimed: `check-asm-to-program.sh` / `check-build-units-link.sh` skip silently without the RISC-V toolchain. No emitted bytes change here, so no layout pins are owed.


pirapira:
**coord review: `merge!` applied.** Both source claims verified against the vendored packages, and the arity reasoning is the valuable part.

### Verified

`ethereum_rlp/rlp.py`, `_deserialize_to_bytes`:

```
if not isinstance(value, bytes):
    raise DecodingError("invalid bytes")
try:
    return class_(value)
except ValueError as e:
    raise DecodingError from e
```

It **constructs the annotated type**, not merely an is-bytes test. And `ethereum_types/bytes.py`, `FixedBytes.__new__`:

```
if len(result) != cls.LENGTH:
    raise ValueError(f"expected {cls.LENGTH} bytes but got {len(result)}")
```

So the reference does length-check `Hash32`/`Address`/`Root`/`Bloom`/`Bytes32`/`Bytes8`, and #11513's scope note — *"`getB` fields are unaffected — those are byte strings either way"* — was wrong. That is the **second** premise of #11513 corrected, after the numeric half in #11617.

### ⭐ The arity split is the insight, and it avoids a real trap

`[]` — what a missing field reads as via `bs.getD i []` — **passes every uint check** (canonical, length 0 ≤ any bound) but **fails `length = 32`**. Index 21 exists only in the 23-field arm, so an unconditional sweep in #11617's shape would have rejected **every previous-fork block** on a 21-field `bpo5` header. Spotting that the two halves cannot share a mechanism, and structuring `fixedBytesFieldWidths` by arity instead, is what makes this correct rather than a copy.

### `extra_data` exclusion is right, and pinned

Plain `Bytes`, genuinely unbounded at decode; the ≤32 rule is a `validate_header` clause, not a decode one. And there is a **`#guard` asserting a 40-byte `extra_data` still decodes** — a negative control, which is what turns "we deliberately left this unchecked" into something that cannot silently regress.

### Scope

No emitted change: the only `Codegen/` edit is a proof-term destructuring in `HeaderExtractNumberBridge` (taking the new trailing conjunct from `decode_header_inv` and discarding it, since `number` is numeric). So no `.text` movement, no pins. Stacked on #11617's branch, so it merges after that one — the actor's sequencing.

### ⚠️ One question, FA-ward, for a follow-up rather than this PR

#11617 found the guest **stricter** than the reference on numeric fields — a false reject. This PR finds the *port* more **lenient** than the reference on byte fields. So: **does the guest length-check these 13 fields?** If it does not, the guest **accepts headers the reference rejects**, which is the dangerous direction, not the safe one. Please state on #11615 or a new issue whether `header_extract_*` enforces the widths, with the citation. A leniency in the port is a correspondence gap; a leniency in the guest would be a soundness gap.

dhsorens:
Answered your FA-ward question on #11615 with citations: https://github.com/Verified-zkEVM/evm-asm/issues/11615#issuecomment-5204508256

Short version — **no soundness gap, and it resolves in the safe direction.** Every byte field the guest actually reads out of header RLP is width-checked and rejects on mismatch:

- `header_extract_logs_bloom` — `headerExtractLogsBloomFunction.s:14-15`, and this one is **proven** not just emitted: `helbRetPost`'s middle arm is `a0 = 2 ∧ Success … ∧ len ≠ 256` (`HeaderExtractLogsBloomSpec.lean:343`);
- `header_extract_state_root` / `_receipts_root` / `_withdrawals_root` — `li x7, 32` / `bne` → `a0 = 2`, at `:44`, `:54`, `:109` respectively.

Two clarifications so "13 fields" is not read as 13 exposure sites: `header_validate_parent_hash` is **not a decode site** (it compares four dwords between caller-supplied pointers, never reads field 0 from RLP), and the guest has **no extractor at all** for the other eight byte fields. No read, no leniency.

⭐ The consequence I had not expected, and the reason this was worth checking rather than assuming: before this PR, `header_extract_logs_bloom` graded against the *port* would have looked `stricter` — guest rejects `len ≠ 256`, port accepts anything — for exactly the same reason `header_extract_number` did on canonicality. **That is the third instance of the same misattribution pattern #11493 unpicked**: a port gap making a correct guest look strict. So fixing the port is what lets those rows grade `agrees` on the width axis instead of each carrying a false FR.

dhsorens:
## EEST differential result — posting this properly, because my first attempt was invalid

Promised on #11617 and owed on both port fixes. Run on the branch containing **both** changes (numeric + `FixedBytes`), i.e. it exercises them together:

```
 SpecRef EEST conformance summary
  total cases : 400
  ERROR/FAIL  : 0    (pre-execution disagreement -- a real SpecRef bug)
  succ FAIL   : 0    (verdict disagreement -- a real SpecRef bug)
  full match  : 400  (root + succ + tail -- the guest's exact output)
  root match  : 400  (pre-execution NPR-root hashing)   [gateable]
  tail match  : 400  (chain-config echo)                [gateable]
  succ match  : 400
  succ diverg : 0
  malformed   : 0
```

This is the check that mattered for these two PRs, since both only ever **add** rejections: if any real Amsterdam header had a non-canonical numeric field or an off-width `Hash32`/`Bloom`, it would show as a pre-execution disagreement. None does.

⚠️ **My first run of this was invalid and I want that on the record rather than buried.** I started it while still editing `Stateless.lean`, so 79 of ~930 fixtures recorded `ERROR` — those were `lake` build failures from a transiently-broken file (mid-way through repairing the `#guard`s that this PR necessarily changed), not divergences. Had I reported the raw "850 OK" tally without opening one of the failures, I would have passed off a run against a half-edited tree as evidence. Killed it, confirmed the tree built clean, re-ran. The 400/400 above is the stable-tree run.

Worth noting for anyone reaching for this harness: it is **not wired into any workflow** (`PROGRESS.md:123` calls it blocking, but grep across `.github/` finds nothing), and a fresh worktree has no fixtures — `scripts/eest-fetch-fixtures.sh 'tests-zkevm@v0.6.2'` is needed first, and neither checkout here had that pinned set. So it only runs if someone runs it deliberately.

Issue #11620, opened 2026-08-06
guest FR: rlp_field_to_u64's 8-byte tooLong rejection is stricter than the reference on 6 of 9 numeric header fields


dhsorens:
Filed at @pirapira's request on #11617:

> the guest's `tooLong` rejection on those fields is **a check the spec does not have**, i.e. a false reject on a header the reference accepts. That is precisely the class we are required to remove — otherwise equivalence needs hash-collision assumptions. Please make sure that lands as its own issue with the field list, since it is a guest change and this PR is port-side.

#11617 fixed the **port** side (canonicality, which the reference does impose). This is the **guest** side of the same finding: a rejection the guest performs that the reference does not.

## The field list

The guest reads header numeric fields through `rlp_field_to_u64`, which returns `Result.tooLong` for a field wider than **8 bytes**. The reference's bound is per-annotation, and the width check is delegated to the target type's `from_be_bytes` — `Uint` has none at all (`ethereum_types` 0.4.1 `numeric.py:523-528`), `FixedUnsigned` has one (`:566-577`).

**Six of the nine numeric fields: the guest is stricter than the reference.**

| idx | field | annotation | reference bound | guest bound | FR? |
|---|---|---|---|---|---|
| 7 | `difficulty` | `Uint` | **none** | 8 | ⚠️ yes |
| 8 | `number` | `Uint` | **none** | 8 | ⚠️ yes |
| 9 | `gas_limit` | `Uint` | **none** | 8 | ⚠️ yes |
| 10 | `gas_used` | `Uint` | **none** | 8 | ⚠️ yes |
| 11 | `timestamp` | `U256` | **32 bytes** | 8 | ⚠️ yes |
| 15 | `base_fee_per_gas` | `Uint` | **none** | 8 | ⚠️ yes |
| 17 | `blob_gas_used` | `U64` | 8 | 8 | no — agrees |
| 18 | `excess_blob_gas` | `U64` | 8 | 8 | no — agrees |
| 22 | `slot_number` | `U64` | 8 | 8 | no — agrees |

⚠️ **`timestamp` is the one worth calling out separately.** It is the only field that is bounded on both sides but at *different* widths — `U256` admits 32 bytes, the guest admits 8. It is easy to miss when scanning for "unbounded `Uint`", and it is the field `chain_validate_increasing_timestamps` reads. (The list in #11617's review names five; `timestamp` is the sixth.)

[...]

PR #11624, opened 2026-08-06
proof(#11575): land row 1 — header_extract_logs_bloom's correspondence row

dhsorens:
First row of #11575. **Stacked on #11623** (which is stacked on #11617), because it depends on the `FixedBytes` width being a *conclusion* of `decode_header_inv` rather than a hypothesis.

## ⭐ It lands cleaner than its representative, and that is the finding

`helbRetPost`'s middle arm is `a0 = 2 ∧ Success … 6 fo len ∧ len ≠ 256` (`HeaderExtractLogsBloomSpec.lean:343`) — the guest **rejecting** a `bloom` whose content length is not 256.

Graded against the port as it stood before #11623, that read as the guest being **stricter**: `getB` was a bare `bs.getD i []`, so a successful `_decode_header` said nothing about width, and the bridge would have had to carry `hdr.bloom.length = 256` as a caller obligation.

It was never a guest defect. `_deserialize_to_bytes` constructs the annotated type and `FixedBytes.__new__` enforces `LENGTH`, so `Bloom = Bytes256` **is** width-checked by the reference. **This is the third instance of the misattribution pattern #11493 unpicked** — a port gap making a correct guest look strict — after canonicality on `number` (#11617) and the `Uint` width question (#11620). With the port fixed, `len = 256` is *derived*, `header_logs_bloom_of_decode` carries **no width hypothesis at all**, and that conclusion is what excludes the `a0 = 2` arm for a composing caller.

Verdict: `.domainRestricted` on **arity alone** — taxonomy row 1, a genuine proof limit (the guest never checks field count, so the honest statement is one-directional). No precondition, no `portDefect`.

## What was actually missing

Not the proof. `headerExtractLogsBloom_spec_within` already existed, `sorry`-free, at its linked guest address — landed 2026-07-16, **ten days before row 9's own triple**. The missing artefacts were the SpecRef bridge and the two registry rows. Same lesson `Correspondence.lean` already records against row 9.

The bridge needed no new walk machinery: `success_content_of_decodeFully_list` is generic in the field index, so instantiating it at 6 was enough.

## Changes

- **`HeaderExtractLogsBloomBridge.lean`** (new) — `header_logs_bloom_of_decode`.
- **`Routines.lean`** — row + two witness abbrevs; censuses 27→28 rows, 19→20 proven, 19→20 symbols. `AxiomWitnesses.lean` regenerated by `scripts/gen-axiom-witnesses.py --write` (137→139).
- **`Correspondence.lean`** — row, family `header`, `basis := .ported` with the clause table that rung requires (the width clause's provenance is `Bloom = Bytes256` → `bytes.py:154-159`, and its enforcement is *proved* by `decodeItemFixedBytes_inv`, not assumed). `registry_size` 25→26, `header_rows` 2→3, `domainRestricted` 4→5, `ported` 4→5.
- **Step bound — this row does NOT inherit #11461's `7 * (2^64 - 1)` factor.** Its `7 * 256` term is the genuine 256-byte bloom copy; the routine calls K20 and never the K34 scalar loop that is #11461's origin. Recorded in the row because the five numeric siblings *do* inherit it, per loop iteration, and the difference is invisible from the row itself.

## ⚠️ `docs/leaf-routine-targets.md`: "mechanical forks" was wrong twice over

That note said *"those eight siblings are mechanical forks"*. Corrected, because both halves mattered:

1. **All eight triples already existed**, ten days before row 9's. The work is bridges + registry rows.
2. **The forks are not uniform**, and the differences drive the grading: K20-only siblings carry no `2^64` factor while the five numeric ones do; `chain_validate_blob_gas_used_multiple` has **no `SpecRef` counterpart at all** (the guest checks `value &&& (2^17 - 1) = 0`, but divisibility by `GAS_PER_BLOB` is a derived consequence of `blob_gas_used = len(hashes) * PER_BLOB`, not a check in the port — expect `.noCounterpart`); and the three ∀-over-header-list routines each need a comparison boundary written down.

Also fixed the stale verified-set counts at `:26-27` and the `GuestAddrs.lean:453` → `:428` citation, and pointed that bullet at the `by decide` censuses instead of restating numbers that can go stale silently.

## Verification

`lake build` clean (4822 jobs), no `sorry`. Gates: `check-registry-crosscheck` (20 symbols × 26 rows), `check-routine-liveness` (387 routines, 386 live), `check-unimported` (2852 modules, 0 orphans), `check-axioms` (139 witnesses, classical only), `check-spec-refs` (44 resolved), `check-progress`, `check-file-size`, `check-forbidden-tactics`.

```
'EvmAsm.Codegen.HeaderExtractLogsBloomSpec.header_logs_bloom_of_decode'        depends on axioms: [propext, Classical.choice, Quot.sound]
'EvmAsm.Codegen.HeaderExtractLogsBloomSpec.headerExtractLogsBloom_spec_within' depends on axioms: [propext, Classical.choice, Quot.sound]
```

⚠️ Not run locally: `check-asm-to-program.sh` / `check-build-units-link.sh` skip silently without the RISC-V toolchain. No emitted bytes change.


pirapira:
**coord review: `merge!` applied.** The substance is right and the "lands cleaner than its representative" point is real. One framing correction below, which matters because it decides whether an open issue looks closable.

### What checks out

The dependency is genuine: with #11623 making the `FixedBytes` width a **conclusion** of `decode_header_inv` rather than a caller obligation, `len = 256` is derived, `header_logs_bloom_of_decode` carries **no width hypothesis**, and that derived conclusion is what excludes the `a0 = 2` arm for a composing caller. I verified the underlying source claims on #11623 — `_deserialize_to_bytes` constructs the annotated type and `FixedBytes.__new__` raises on `len(result) != cls.LENGTH` — so `Bloom = Bytes256` is width-checked by the reference and the guest's `len ≠ 256` rejection is **matched**, not strict.

`.domainRestricted` on **arity alone** is the honest verdict: the guest never checks field count, so a one-directional statement is the truthful one. No manufactured precondition, no `portDefect`.

Scope: proof and registry only — new bridge file, `Imports`, `AxiomWitnesses` (+4), `Correspondence`, `Routines`, `PROGRESS.md`, docs. **No emitted instruction strings added**, so no `.text` movement and no layout pins. Stacked on #11623 → #11617, so it merges last of the three; the actor's sequencing.

### ⚠️ Correction: #11620 is not an instance of the misattribution pattern — it is the one that turned out real

The body reads:

> This is the third instance of the misattribution pattern #11493 unpicked — a port gap making a correct guest look strict — after canonicality on `number` (#11617) and the `Uint` width question (#11620).

Canonicality belongs in that list: the port lacked the check, the reference has it (`_deserialize_to_uint`), so the guest was correct and the port was the gap. **The `Uint` width question is the opposite.** #11620 — filed at my request off #11617 — establishes that `rlp_field_to_u64` rejects fields wider than 8 bytes while the reference's bound is per-annotation and `Uint` has **no width check at all** (`numeric.py:523-528` versus `FixedUnsigned` at `:566-577`). So on six of nine numeric fields the guest **is** stricter than the reference. That is a live guest FR — a false reject on a header the reference accepts — and it is open precisely because it needs a **guest** change.

So unpicking #11513's "matched pair" produced **one misattribution and one confirmed defect**, not two misattributions. Listing them together invites the reading that #11620 is closable as a mis-grading, which it is not.

Please adjust the sentence (a follow-up comment is fine — not worth a force-push). With `Bloom` this PR makes three cases where the reference's check is delegated to the *type*, and the interesting general lesson is that **"guest stricter than port" resolves differently per annotation**: matched for `FixedBytes`/`FixedUnsigned`, a genuine over-rejection for `Uint`.

dhsorens:
You're right, and thank you — that was a consequential slip, not just loose wording.

**Corrected in-tree rather than only here**, because the same sentence was in the bridge docstring *and* the Correspondence row note, and a wrong claim in a registry note is exactly the kind of stale comment that outlives the thread that produced it. Pushed as `8e6748337` (normal commit, no force-push).

The list is now two, not three:

- **misattributions** — canonicality on `number` (#11617) and `Bloom`'s width (this row): the reference imposes the check, the guest performs it, the **port** was the gap;
- **confirmed guest FR** — the `Uint` width question (#11620): the reference imposes no bound at all, the guest bounds at 8, so the guest genuinely over-rejects.

Both sites now carry your framing as the transferable lesson, which is better than what I had: **"guest stricter than port" resolves differently per annotation** — matched for `FixedBytes`/`FixedUnsigned`, a genuine over-rejection for arbitrary-precision `Uint`. Read the annotation's own conversion before deciding which case you are in. The row note says explicitly that #11620 is *not* a third instance and is *not* closable as a mis-grading.

⚠️ **One thing I should flag rather than quietly resolve**, since your two rulings point different ways and I do not want to pick for you. On #11620 you wrote that the project-wide assumptions *"give the guest freedom to choose its behavior. I prefer reject on out of assumption inputs"* — which is why #11617 now grades that row `.domainRestricted` in its precondition flavour rather than `.stricter`. Here you write that #11620 is "a live guest FR … open precisely because it needs a **guest** change."

I read those as compatible if "guest change" means your second top theorem — proving the guest *rejects* out-of-assumption header values, which upgrades the rejection from permitted-freedom to a proven property. That is how #11620 is currently scoped. But if you meant the rejection itself should go, then the `.domainRestricted` grade in #11617 is wrong and I should move it back. Say which and I will make it consistent in one commit.

PR #11625, opened 2026-08-06
remove(#11350): delete amsterdam_blob_gas_price — zero production callers (keeping the _u256 sibling)

### LINKED ISSUE #11350: proof: triple for `amsterdam_blob_gas_price` against `calculate_blob_gas_price`
One of the ten orienting rows from `docs/leaf-routine-targets.md` (row **8**), filed per-routine so each can be

## The pairing

| | |
|---|---|
| **RISC-V routine** | `amsterdam_blob_gas_price` |
| **`SpecRef` target** | `calculate_blob_gas_price` (`EvmAsm/Stateless/SpecRef/Gas.lean:136`) |
| **caller** | `EvmAsm/Codegen/Programs/Header.lean:889` |
| **dependency status** | leaf |
| **data structures** | scalars only (register ABI) |
| **separation-logic `Assertion`** | n/a — register vocabulary suffices, as row 7 |

At the Amsterdam fork parameters specifically.

Informally: the returned price equals the spec's blob-gas-price computation at the Amsterdam parameters for the given excess blob gas.

existing infrastructure:

semantics.

## Notes

⭐ The port already carries **Python-checked `#guard`s** (`Gas.lean:246`), so the reference side has concrete cross-checked values to test the triple against before proving it — a rare starting advantage.
## Related

dhsorens:
Closes #11350.

> Remove this routine, and make sure the RISC-V guest can run tests, and close this issue.
> — @pirapira

Deleted rather than proven, per that direction.

## ⚠️ There are two routines and only one is dead

This is the part that made it not a one-line delete:

| routine | callers | fate |
|---|---|---|
| `amsterdam_blob_gas_price` (u64) | **zero** — its only in-edge was its own probe prologue | **deleted** |
| `amsterdam_blob_gas_price_u256` | `header_validate_excess_blob_gas` (`HeaderBaseFee.lean:417`, reloc `:475`) **and** the runtime payload (`BlockVerdictRuntimePayload.lean:150`) | **kept** |

Deleting the wrong one — or reading "remove this routine" as covering both — breaks excess-blob-gas validation. Every edit here is on the non-`_u256` variant, and the `_u256` caller sites are untouched.

## What was removed

The 55-instruction routine with its `_eq_prog` theorem and `#guard`s, plus its probe unit (127 lines of `Header.lean` together), 6 emit-site concatenations across the verdict/header modules, the `GuestAddrs` def, both probe registries, the `MANIFEST.tsv` row, and the asm fixture.

## Two knock-ons worth naming

**`GuestImageEntries.lean` is a proof file.** Its `#guard guestImageEntries.length = 341` had to become 340 — a routine deletion is a proof edit here, not just codegen cleanup. This is why I flagged #11350 as "not a trivial delete" when scoping it.

**Layout repin.** `.text` shrank by exactly 220 bytes (55 × 4), `textSizeBytes` `0x563f8` → `0x5631c`, symbols 1022 → 1021. Regenerated `symbol-addresses.tsv` (with `--build`), `GuestAddrs.lean`, `RegionMapLinkPins.lean`, PROGRESS.md and DRIFT.md. `check-declared-starts` is 340/340.

## ⭐ Also retired the leaf-table row, pointing it at the survivor

`docs/leaf-routine-targets.md` row 8 was this routine. Marking it closed silently would leave the next reader believing blob gas price is handled — when the **live** one is still unproven, against the same `calculate_blob_gas_price` (`SpecRef/Gas.lean:136`). The row now says so explicitly and names `amsterdam_blob_gas_price_u256` as the routine actually worth a triple.

## Checks

Not graded in `Routines.lean` or `Correspondence.lean`, so no registry row to retire. No test, fixture or harness referenced the deleted probe unit (`zisk_amsterdam_blob_gas_price`).

On "make sure the guest can run tests": `check-build-units-link.sh` emits **and links every build unit** via `lake exe codegen`, and passes — so the guest and every probe still assemble and link with the routine gone. Combined with `check-declared-starts` 340/340 that is the link verified end to end. CI's `check-region-map.sh` covers the readelf-based extents check, which skips locally on the toolchain-name spelling.

Local gates: `guestaddrs-starts` · `file-size` · `unimported` · `forbidden-tactics` · `progress` · `build-units-link` · `registry-crosscheck` · `routine-liveness` · `drift-report --check` — all OK.


pirapira:
**coord review: `merge!` applied.** I verified the deletion claim independently rather than taking it, because "zero callers" on a name that is a **prefix of a live sibling** is the exact shape where a true statement gets applied to the wrong symbol.

### The dead-routine claim holds

Grepping `origin/main` for the bare name with `_u256` excluded, every occurrence is the routine's own definition, docstring, emit site, `#guard`s, `GuestAddrs` def, `GuestImageEntries` pair, probe registration in three registries, the fixture, and the TSV row — plus exactly one `jal`:

```
Header.lean:889:  "  jal ra, amsterdam_blob_gas_price\n" ++
```

and that line is **inside `ziskAmsterdamBlobGasPriceProbeUnit` itself**. So the only in-edge really is its own probe prologue. No production caller.

### The sibling is intact — checked, not assumed

Four removed lines mention `_u256`, which is precisely what would look alarming at a glance. All four are **address-shift rewrites in regenerated artefacts**, each matched by an addition:

| symbol | before | after |
|---|---|---|
| `amsterdam_blob_gas_price_u256` | `0x8000b038` | `0x8000af5c` |
| `slot_decode_u256` | `0x8001e1ac` | `0x8001e0d0` |

The `_u256` routine inherits `0x8000af5c` — the deleted routine's own address — which is exactly what deleting the immediately-preceding routine should produce. And its live callers are untouched: `HeaderBaseFee.lean:417` (the `jalOff` into `header_validate_excess_blob_gas + 76`), `:475` (the reloc row), and `BlockVerdictRuntimePayload.lean:150`.

**So the two-routine trap you flagged was real and you avoided it.** Reading "remove this routine" as covering both would have broken excess-blob-gas validation, and nothing in the issue text would have stopped it.

### Knock-ons handled

- **All three layout artefacts regenerated** — `GuestAddrs.lean`, `RegionMapLinkPins.lean`, `symbol-addresses.tsv`. A deletion moves every later address, so this is the case where regeneration is unavoidable rather than incidental.
- `GuestImageEntries.lean` `341 → 340`: a routine deletion is a **proof** edit here, not just codegen cleanup. Worth having said so on #11350 when scoping it.
- Six emit-site concatenations, both probe registries, `RegistryNames`, the `MANIFEST.tsv` row and the 56-line fixture all removed together — the set that would otherwise leave a dangling reference.

CONFLICTING is main having advanced; the actor's.

⭐ And the docs row you left behind is the right artefact: a struck-through entry that says the routine was removed, **plus an explicit warning not to confuse it with the surviving `_u256`**, with that sibling's callers named. A future reader of #11350 would otherwise have to redo exactly the analysis this PR did.

Issue #11590, opened 2026-08-06
structural (#11581 slice 5): synchronous account-write publication for CREATE/CREATE2


pirapira:
Slice 5 of #11581 (umbrella). Gated on the harness in the slice-1 issue. **Do this last of the four families.**

## Family: CREATE / CREATE2

- Creator debit: `ChildFrameCreateTail.lean:317-354`, **before** descent.
- Child credit: `:479-512`.
- Paired records: only at `:513-521`.
- Failure rollback: shared `frame_return`.
- Success finalization: `NoopHalt.lean:310-329` (deferred finalization, not new mutations).

## Ask

Synchronous publication at both mutation points with the mark captured before the creator debit.

⚠️ Highest cost of the four: needs **per-depth provisional writer state** and undo ordering
across depths. `frame_base(d) = arena + d*0x19000`, and depth 0 owns arena slot 0.

⚠️ Audited exclusions from scoping — do **not** treat these as part of this lane: SELFDESTRUCT
uses only RLP/scratch and terminating `record_nonstorage_effect` paths, with no direct `env+32`
write; upfront gas debit/refund are AccountState-first finalization paths, not this live-env gap.

Cost estimate from scoping: **highest**.

## Gate (structural-alignment bar)

Per the maintainer's standing ruling: a structure change that aligns with `execution-specs`
lands on **200 random A/B plus named controls** — the bar is *structure matches the spec* and
*r200 does not deteriorate a lot*, **not** byte-equivalence. An empty flip set is expected and
is not evidence of no-op-ness; alignment is its own justification because it shortens the proofs.

Required in the PR body:
- base sha + candidate ELF sha256 (both), and the r200 flip set **in both directions**
- the named controls for the family (list them; a fix verified only on failing rows has an
  unmeasured downside)
- a spec citation by file:line for the ordering being adopted

⚠️ `output[0:32]` is `new_payload_request_root`, **not** the state root. Value-neutrality
evidence is `succ` agreement plus the flip set — never "roots matched".

Umbrella: #11581

pirapira:
## Closing with no code change — the requested ordering is already present on main

Audited on merged main `e8f2df6b1`:

| step | site |
|---|---|
| creator debit (pre-descend, with the `live_balance` mark) | `ChildFrameCreateTail:317-355` |
| `create_frame_descend` | `:468-469` |
| child `env+32` credit | `:481-517` |
[...]

PR #11626, opened 2026-08-06
fix(codegen): publish CALL value pair after child credit

pirapira:
## Change

Move the shared CALL value-transfer publication until after the child live-credit mutation.  The paired debit/credit record remains one atomic `move_ether`-shaped publication; this does not split it into separate records.  The empty-code path is unchanged.

The Amsterdam execution-specs order is snapshot (`interpreter.py:380`), then `move_ether` (debit and credit, `:384-390`), then execution.  This cut aligns the publication with the already-established child snapshot and keeps it inside the child rollback interval.  The parent live debit remains before descent in this PR; removing that bespoke ordering/undo mechanism (`live_balance_debited_by_depth`) is intentionally a separate follow-up.  The insufficient-balance gate remains before descent.

On child failure, `CallFrameReturn.lean:201-211` restores the frame effect/account-state cursors and `:237-268` recredits the parent live balance, so the post-snapshot paired record and the live transfer are discarded/reversed together.  `recordMessageValueTransferAsm` is descriptor-only and performs no state lookup.  The existing child snapshot at `CallFrameDescend.lean:445-507` remains the rollback boundary.

## Evidence

Prediction before measurement: verdict output and section sizes remain unchanged; only the ordering of the emitted paired publication moves.  Both artifacts were built from the same base commit `90f592294`.

- Base ELF: `/tmp/11588-base.elf`, SHA256 `9e75a39843ca316c774c395277d46621c69aee2efe5dc51d2ec937a1b033d422`, built 2026-08-06 14:24:01 +0200.
- Candidate ELF: `/tmp/11588-candidate.elf`, SHA256 `d4b6fce78ff592dcb2831730293cd2796558b006ae31c601921cfd67320bc388`, built 2026-08-06 14:22:38 +0200.
- `riscv64-unknown-elf-size`: identical for both (`text=354576`, `data=21424`, `bss=426940448`).
- Random-200, seed `11588`, backend Spike, 30 jobs, same manifest selection: base 200/200 exact, candidate 200/200 exact; errors 0, budget 0, fail 0 on both; both flip sets empty.
- Named `value_moving_transactions` (limit 50): base and candidate 50/50 exact, errors 0, fail 0, both flip sets empty.
- Named `transfer_logs` (limit 50): base and candidate 50/50 exact, errors 0, fail 0, both flip sets empty.
- Named `value_moving_with_tx_delegation` (8 selected by the filter): base and candidate 8/8 exact, errors 0, fail 0, both flip sets empty.

The random run and each named control reported full/root/success/tail parity.  No full corpus sweep was run.

### DIFF
--- a/EvmAsm/Codegen/Programs/ChildFrameHandlers.lean
+++ b/EvmAsm/Codegen/Programs/ChildFrameHandlers.lean
@@ -1159,6 +1159,11 @@ def callDescendFallThrough
    -- one paired record rather than manufacturing separate debit/credit records.
    -- This remains post-snapshot and therefore inside the child rollback interval;
    -- frame_return drops it together with the live debit/credit on failure.
+   -- The descriptors (cd_caller_be, nse_callee_be, cd_value_be, cd_balance_be,
+   -- nse_acct) are static scratch reused by nested calls, so this must stay in
+   -- the post-descend, pre-dispatch window. The same publisher is used by the
+   -- CREATE endowment site: process_create_message delegates to process_message
+   -- (interpreter.py:212), so both sites implement the spec's single move_ether.
    (if mode != 0 then "" else
      recordMessageValueTransferAsm "cd_caller_be" "nse_callee_be" "cd_value_be" "li a3, 1"
        "cd_balance_be" "nse_acct" (recipientPreAdjust := "addi a5, a5, 8")) ++
--- a/EvmAsm/Codegen/Programs/ChildFrameHandlers.lean
+++ b/EvmAsm/Codegen/Programs/ChildFrameHandlers.lean
@@ -1111,22 +1111,9 @@ def callDescendFallThrough
   "  mv x10, s10                           # restore parent PC for frame_save_regs\n" ++
   "  la a1, cd_desc\n" ++
   "  jal ra, call_frame_descend\n" ++
-  -- The child snapshot is now established.  Publish the paired value-transfer
-  -- BAL effects from the pre-resolved stable descriptors so frame_return drops
-  -- them with a reverting child, matching process_message's snapshot→move_ether
-  -- ordering. The helper performs no state lookup; it only consumes these
-  -- caller-supplied pointers, so the switched x12/x20 registers are irrelevant.
-  -- This is deliberately before the first child dispatch: cd_caller_be,
-  -- nse_callee_be, cd_value_be, cd_balance_be, and nse_acct are static scratch
-  -- reused by nested calls, so they are valid only in this post-descend,
-  -- pre-dispatch window.
+  -- The child snapshot is now established. Credit child env+32 (.selfBalance)
+  -- with the transferred value so nested-frame SELFBALANCE and the child's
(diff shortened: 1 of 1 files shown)


pirapira:
**coord review: the reorder is correct. Holding the label on a comment-only fix — two deleted facts are still load-bearing.**

### The code change is right

The paired publication now follows the child `env+32` credit, so both live mutations precede the single map transition — and the new comment states the reason correctly: `move_ether` is one atomic paired transition, so keeping **one** paired record is what corresponds to the spec, rather than manufacturing separate debit/credit records the spec has no intermediate state for. It remains post-snapshot, so `frame_return` still drops it with the live debit and credit on a reverting child. `+10/−15` in one file, empty-code path untouched.

### ⚠️ Restore these two before it lands

**1. The static-scratch validity window.** The deleted block said:

> `cd_caller_be`, `nse_callee_be`, `cd_value_be`, `cd_balance_be`, and `nse_acct` are static scratch reused by nested calls, so they are valid only in this post-descend, pre-dispatch window.

The publication **still consumes exactly those descriptors** and is still pre-dispatch, so **the constraint still holds** — it is now simply undocumented. Your move shortened the safe window on the safe side, but the next person who touches this line has lost the warning that says why it cannot move past dispatch. That is the failure mode this comment existed to prevent.

**2. The shared-helper fact (GH #10938).** The deleted block said `recordMessageValueTransferAsm` is **also used by the CREATE-endowment site**, and that the spec has ONE `move_ether` for both because `process_create_message` delegates to `process_message` (`interpreter.py:212`).

That is directly load-bearing for **#11590**, the CREATE/CREATE2 slice — it tells whoever takes it that the two sites share a publisher and that the spec justifies treating them the same way. It is expensive to rediscover and cheap to keep.

### Why I am holding rather than waving it through

I made the same call on #11592 and told you then that deleting a comment which states a non-obvious mechanism is a loss even when the code is fine; it would be inconsistent to relax that here. And **comments cannot reach the emitted image, so restoring them needs no re-measurement** — your `base 90f592294` evidence stands unchanged. It is a two-minute edit.

Fold both into the new comment rather than reinstating the old text verbatim: the atomicity reason you added is better than what was there, and it should sit alongside the scratch-window constraint and the shared-with-CREATE fact rather than replacing them.

Ping me when pushed and I will label immediately.

pirapira:
**Both facts restored — `merge!` applied.** The retained comment now carries the static-scratch descriptors and the post-descend/pre-dispatch window constraint, plus the shared-publisher fact with the `interpreter.py:212` delegation showing both sites implement the spec's single `move_ether`. `+15/−15`, still one file, and no re-measurement was owed since comments cannot reach the emitted image.

Net effect is better than what was there before: the atomicity reason for keeping one paired record is new, and the two constraints it would have displaced are kept alongside it rather than replaced. Hands off from here.

Follow-up B is tracked on #11001 rather than a new issue — correct call, and it now carries the Amsterdam anchors, the three `live_balance_debited_by_depth` touchpoints, and the insufficient-balance-check-before-descent constraint.

PR #11627, opened 2026-08-06
chore: batch-merge #11621 + #11623 + #11624 + #11625 + #11626

pirapira:
Batches five `merge!`-labeled PRs into one integration PR to amortize CI.

- #11621 — fix(codegen): resolve original for seeded system storage rows
- #11623 — fix(specref): re-impose FixedBytes length checks on `_decode_header`'s byte fields
- #11624 — proof(#11575): land row 1 — `header_extract_logs_bloom`'s correspondence row
- #11625 — remove(#11350): delete `amsterdam_blob_gas_price` — zero production callers
- #11626 — fix(codegen): publish CALL value pair after child credit

#11623 → #11624 are a stack (dhsorens); merging #11624's branch carries #11623, and both are asserted ancestors of this head independently. All five constituents asserted present.

**Two non-generated resolutions, both determinable:**

1. **`docs/leaf-routine-targets.md`** — two-sided. Row 8 taken from #11625 (the `ROUTINE REMOVED (#11350)` note); row 9 taken from HEAD, because #11625 branched before #11617 moved that reference and still carried the stale `SpecRef/Stateless.lean:75` where current is `:158`. Asserted after resolving: removal note present, `:158` present, `:75` absent, 0 markers.

2. **`EXPECTED_BARE_J_SITES` 176 → 175** — the gate requires a stated reason, so: #11625 deletes `amsterdam_blob_gas_price` with its MANIFEST row and `.s` fixture, and that def contributed exactly one blocking bare-local-J site. Counters move in lockstep (`80 defs → 79`, `176 sites → 175`). Attributed rather than inferred: #11625's branch **standalone, with no merge**, reproduces exactly this delta, and no other constituent moves the counter. Ratchet **down**, the direction the migration travels.

Verified locally: `lake build` green (4820 jobs) and all standing gates `ok`.


pirapira:
**coord review: `merge!` applied.** Batch of five, and both non-generated resolutions are justified rather than asserted.

### The ratchet change is properly attributed

`EXPECTED_BARE_J_SITES 176 → 175` is an **equality** gate, so this was the thing to check. The justification does it the right way: the delta was reproduced by running **#11625's branch standalone, with no merge**, and no other constituent moves the counter. That is measurement, not inference — which matters, because reasoning about which sites count is exactly where this would go wrong.

⚠️ **One thing to make explicit for the next reader**, since I checked it and had to work it out: the deleted fixture contains **two** bare-`j` instructions (`j .Labgp_loop`, `j .Labgp_ret`), yet the counter drops by **one**. The reconciliation is in your word *blocking* — the ratchet counts blocking bare-local-J sites, a subset — and it is corroborated by the lockstep `80 defs → 79` alongside `176 → 175`: two blocking sites in one def would have moved defs by 1 and sites by 2. Worth stating in the PR body, because anyone who greps `j ` in the fixture finds 2 and will suspect the gate was fudged.

### The docs resolution is the careful one

Taking row 8 from #11625 (the `ROUTINE REMOVED` note) and row 9 from HEAD — because **#11625 branched before #11617 moved that reference** and still carried the stale `SpecRef/Stateless.lean:75` where current is `:158` — is precisely the resolution that a mechanical two-sided merge would have got wrong, silently reinstating a stale line number that #11617 had just corrected. Asserting afterwards that the removal note is present, `:158` present, `:75` absent, 0 markers is the right form of check.

### Constituents

#11621 (seeded-storage baseline), #11623 (`FixedBytes` widths), #11624 (bloom correspondence row), #11625 (routine deletion), #11626 (CALL value pair after child credit) — all five reviewed and labelled individually, and the #11623 → #11624 stack is handled by carrying #11624's branch with both asserted as independent ancestors.

Generated artefacts regenerated as a set: `GuestAddrs.lean`, `RegionMapLinkPins.lean`, `symbol-addresses.tsv`, plus `GuestImageEntries` `341 → 340` for the deletion. `lake build` green at 4820 jobs with standing gates ok.

Issue #11364, opened 2026-08-04
MTx leaks EIP-7702 AUTH_BASE and ACCOUNT_WRITE gas into later transactions


pirapira:
Attribution: codex2

## Summary

The multi-transaction guest applies a fixed EIP-7702 gas surplus to later transactions after a type-4 transaction. In fixture 00612, Tx1 is the only transaction with an authorization list; Tx2 and Tx3 are legacy transactions that CALL the delegation installed by Tx1. The candidate guest charges each later transaction an extra 43,190 gas.

This is separate from #11362 (CREATE2 collision/tombstone) and #11363 (failed CREATE/CREATE2 creator nonce).

## Reproduction

- Fixture: `00612_test_bal_7702_cross_tx_delegation_then_call_fork_Amsterdam-blockchain_test__b0`
- Pinned execution-specs: `e5a8caf1b8055e4d805c7fb169edfa710914b7da`
- Candidate guest SHA-256: `6a7242fc3049b5fec53bb3612fcf244f47e6693933d8712cf31efd7659a11c7e`
- Input: `/var/tmp/fc668/00612_test_bal_7702_cross_tx_delegation_then_call_fork_Amsterdam-blockchain_test__b0.input`

The reference post-balances imply gas used `129,032` for Tx2 and `31,112` for Tx3. The difference is exactly the legitimate first zero-to-nonzero storage charge: `STORAGE_SET = 64 * 1,530 = 97,920`. Tx3’s regular path is `12,000 TX_BASE + 3,000 cold recipient + 3,000 cold delegated-code resolution + 3,112 body`, where the body is cold `SLOAD` 3,000 + `ADD` 3 + three `PUSH1`s 9 + warm `SSTORE` 100 + `STORAGE_WRITE` 10,000.

The guest charges both Tx2 and Tx3 an additional `43,190`, visible in the balance deltas (`43,190 * 10` from each sender and `43,190 * 3` to the fee recipient).

## Exact candidate decomposition

Final-memory dump from the same run:

- `bvgr_tx_state_gas = [35190, 35190, 35190]`
- `bvgr_tx_exec_state_gas = [0, 97920, 0]`
- `runtime_tx_auth_state_refund = 35190`
- `runtime_tx_auth_regular_refund = 8000`
- `runtime_tx_top_frame_regular_gas = 8000`

Therefore the fixed surplus is two line items:

[...]

PR #11628, opened 2026-08-06
fix: preserve balance authority across failed CREATE refund

### LINKED ISSUE #11619: Execution: failed CREATE endowment refund leaves stale AccountState balance authority
> Status: the mechanism is not yet discriminated. Both observed fixtures use
> `CREATE(value=1)`, so the measured one-wei delta could be either a missing
> endowment return or an independent off-by-one defect. The locally unskipped
> `CREATE(value=0xB)` fixtures are the decisive follow-up.

## Summary

The AccountState agreement diagnostic found a one-wei balance shortfall on a
real failing fixture. The map representation carries the correct balance;
the live AccountState row is one wei low after a failed CREATE with init-code
stack underflow.

## Reproduction

- Fixture: `16014_test_create_init_fail_stack_underflow_fork_Amsterdam-blockchain_test_from_state_test__b0`
- Source: `execution-specs/tests/ported_static/stCallCreateCallCodeTest/test_create_init_fail_stack_underflow.py`
- Address observed: `0x1ec952083e988eeb19fcab317760ffc6671246fd`
- Candidate ref: `af24a7eec`, ELF SHA-256 `e00beba0c1bf674fabb8b506ef4e41fcd98c8b8d0e052217c9143da36ff28f86`
- Clean base ref: `231b1195d`, ELF SHA-256 `729349337a12a80a61234dfdc73d654582f6ce4c5e0ccafa4b279e32eeea5246`

The fixture initializes the target with balance `0xDE0B6B3A7640000` and sends
transaction value `0x186A0`. Therefore the target balance before its
selfdestruct is `0xDE0B6B3A76586A0`; the post-state oracle assigns that exact
amount to the zero address.

The diagnostic event stores 64-bit limbs loaded from big-endian balance bytes,
so the differing limb must be byte-reversed before interpretation:

```text
map limb:  0xa08665a7b3b6e00d -> 0x0de0b6b3a76586a0
live limb: 0x9f8665a7b3b6e00d -> 0x0de0b6b3a765869f
```

Limbs 0, 1, and 2 compare equal; only limb 3 differs. The map is therefore
the oracle-correct side, while live AccountState is one wei low.

The fixture is FR on both artifacts: oracle success byte `1`, guest output
status byte `0`. Candidate and clean-base output bytes are identical, so this
is not introduced by the diagnostic instrumentation.

Four registered readers observe the same stale row, not four independent
defects:

- reader 2: `EvmOpcodes:409`
- reader 7: `Selfdestruct:48`
- reader 9: `Selfdestruct:513`
- reader 13: `Selfdestruct:834`

The independent 255-row leaf-decoder population also includes this fixture
and classifies it as `balance_mismatch:balance` (`/tmp/opencode/hist/fr_bv1.ids`,
`/tmp/opencode/leaf-decoder/fr255_corrected.tsv`).

This should be tracked as an execution/state-transition defect, separate from
the AccountState publication observability work and from the refund-publication
issue.

pirapira:
## Summary

Fixes the failed-CREATE endowment refund shortfall in #11619.

`ChildFrameCreateTail` debits the creator balance and emits a nonce-only
AccountState record before descending. On child failure, `CallFrameReturn`
restores the runtime balance through `live_balance_debited_by_depth`, but the
old AccountState writer had already made the post-debit balance authoritative.
That stale row was then read by BALANCE and SELFDESTRUCT.

`account_state_record_nonstorage` now compares all four big-endian balance
limbs before lookup. It writes balance bytes and sets `HAS_BALANCE` only when
pre-balance differs from post-balance; equal operands preserve any cloned
balance overlay. This matches the Amsterdam execution-specs condition in
`block_access_lists.py:644` (with nonce's analogous condition at `:649`).

The issue is an endowment-sized refund shortfall, not an arithmetic off-by-one:
the spec restores the pre-transfer snapshot, while the guest's bespoke frame
return path restores the runtime balance and the stale AccountState claim was
the missing piece.

## Validation

- `lake build`
- `git diff --check`
- Targeted EEST sweep: fixtures 16009, 16010, 16013, 16014, 16016 — 5/5 OK,
  FA=0, FR=0.
- Standing controls 18529 and 15904 agree.
- Runtime AccountState/AccountWriteMap agreement harness: 7 cases, zero
  mismatches and zero unregistered readers.
- Diagnostic rebuilt/supplied lengths: 399/399 on all five target fixtures and
  410/410 on both controls.

Built guest ELF SHA256:
`d3ef42c7e832ac0c5b66b1fb04a81bc1a50369e33e138dbea96c0442b66d8d53`.

Closes #11619.


pirapira:
**coord review: substance looks right; holding the label for one measurement.**

### What is established and good

The mechanism is precisely diagnosed and the fix matches the spec's own rule. `ChildFrameCreateTail` debits the parent, marks `live_balance_debited_by_depth`, then emits a **nonce-only** `AccountState` snapshot whose balance operands are **equal** — taken *after* the debit — and `account_state_record_nonstorage` wrote them unconditionally, setting `HAS_BALANCE`. On a failed child, `frame_return` recredits `env+32` and rolls back the child row, but **that stale balance claim survives**. So a nonce-only operation silently took authority over a field it never changed, and the correction did not reach it.

The fix — compare all four BE limbs, preserve the cloned balance when pre equals post, set `HAS_BALANCE` only on a changed balance — is what `block_access_lists.py:644` does (`if pre_balance != post_balance: add_balance_change(...)`, and `:649` likewise for nonce). **The spec only records a field when it changed**, so this is alignment as well as a defect fix.

Evidence I am satisfied with: five-row F→P, the **exact** length identities (399=399 on all five targets, 410=410 on both controls) which matter because this fix *removes* records, and the consumer audit naming `account_state_latest_balance_core`, `account_write_touch_current`, `account_writes_apply_deletes`, `code_state_final_balance_nonzero` and the agreement/probe readers as all gating on bit 32 — the *read* side, not the set side.

Also right: declining to claim an A/B against the stale `/tmp/11605` base because it lacks the agreement symbols. Not comparing beats comparing against an incomparable artefact.

### ⚠️ The gap: 7 cases is a target set, not a population

Everything measured so far is the five rows this was designed to fix plus two standing controls. **A fix verified only on its targets has an unmeasured downside.** #11364 closed 13 of 14 targets and broke a passing row; only a reverse sweep over *passing* rows caught it.

And the blast radius argues the same way: `account_state_record_nonstorage` is **not CREATE-specific**, so a write-condition change there reaches every nonstorage record in the guest. That is exactly the shape where a targeted sweep looks perfect and a population sweep finds the regression.

**Needed before I label:** the standing 202 (your 200 random plus `18529` and `15904`), reporting **base-PASS → candidate-FR as a set**. Zero is the bar. State the prediction first with its reason — suppression happens only when pre equals post, and the spec records a change only when they differ, so no legitimate record should disappear. If a row flips P→F, the reason matters more than the count.

Also note `textSizeBytes`: `CreateCodeEffectLog.lean` is emitted asm, so if the image moved, pins are owed — leave the regeneration to the actor, just state whether it moved.

pirapira:
Standing-202 population evidence (fresh A/B after rebasing onto `e8f2df6b1`):

- The 200-row random population was `200/200 OK` on both base and candidate.
- Controls `18529` and `15904` were identical on both artifacts (`OK` on each; `15904` has the same expected-prefix mismatch in both, while oracle and guest verdicts agree).
- The base-PASS to candidate-FR set is empty. The reverse candidate-PASS to base-FR set is also empty across all 202 rows.

The five target rows are `FR` on base and `OK` on candidate, with no false accepts. Fresh direct ziskemu probes report `rebuilt_len == supplied_len` for every row: `410 == 410` on both standing controls and `399 == 399` on all five targets. The candidate target status is 0; the fresh base target status is 1.

The emitted image moved: `.text` is `0x56848` on the fresh base ELF and `0x56898` on the candidate, a `0x50` byte increase. `RegionMapLinkPins.textSizeBytes` remains `0x56848`; pin regeneration is left to the actor as requested.

Artifacts:

- candidate commit `03abda9be2cd105e416235f7c721182565d9f14a`
- candidate ELF SHA-256 `683b39a0cf99b11edbfa4b9dc74df0f4b8871460f8e28f3764597e0b9e522ff3`
- base ELF SHA-256 `cc7d360ab2d487c65c196c5760ac7a92d252a86c31b18d5096ea8d4e6cc379c0`

pirapira:
**Population sweep clean — `merge!` applied.**

Standing 202 on fresh main `e8f2df6b1` vs candidate `03abda9be`: base 200/200 OK, candidate 200/200 OK, both controls identical. **`BASE-PASS → CANDIDATE-FR` is empty, and the reverse set is empty too.** Five target rows base-FR → candidate-OK. Length probes exact on the controls.

That was the one thing missing, and it is the one that matters here: `account_state_record_nonstorage` is **not** CREATE-specific, so a write-condition change in it reaches every nonstorage record in the guest. A five-target sweep could not have seen a regression elsewhere by construction — #11364 closed 13 of 14 targets and broke a passing row, visible only in a reverse sweep. **The prediction was also stated in advance with its reason** (suppression fires only when pre equals post, and `block_access_lists.py:644` records a change only when they differ), so the empty set confirms a prediction rather than merely reporting an absence.

### The finding, for the record

A **nonce-only** operation silently claimed authority over the balance field: `ChildFrameCreateTail` debits the parent, marks `live_balance_debited_by_depth`, then emits a nonce-only `AccountState` snapshot whose balance operands are *equal* — taken after the debit — and `account_state_record_nonstorage` wrote them unconditionally, setting `HAS_BALANCE`. On a failed child, `frame_return` recredits `env+32` and rolls back the child row, **but that stale balance claim survived**. BALANCE and SELFDESTRUCT then read it; the map stayed correct.

So the correction reached everything *except* the spurious record — recording a field the operation never changed, in the one form where rollback cannot fix it.

Fix aligns with the spec's own rule rather than merely patching: `block_access_lists.py:644` records a balance change only when `pre_balance != post_balance`, `:649` likewise for nonce. Consumer audit covered the **read** side — `account_state_latest_balance_core`, `account_write_touch_current`, `account_writes_apply_deletes`, `code_state_final_balance_nonzero`, and the agreement/probe readers — all gating on bit 32.

⭐ And the mechanism question resolved **from source** without a discriminating fixture: the stale row holds `pre − value`, so the deficit *is* the endowment and would scale with a larger one. Every corpus row uses value `0x1`, which is precisely why one wei and one endowment were indistinguishable by measurement — and why my "one-wei off-by-one" framing needed retracting. The retitle reflects that.

PR #11629, opened 2026-08-06
feat: add block-only account nonce lookup

pirapira:
## Summary

Adds the missing additive map-side contract `account_writes_latest_nonce_block`.
This is a block-cumulative account-write-map lookup only; it does not repoint a
reader, delete an overlay row, or change any existing contract.

## Contract

- `a0`: canonical 20-byte big-endian address pointer.
- `a1`: one-word output pointer.
- Scans only `ACCOUNT_WRITES_AREA` at `0xa28a0000`, using the existing 128-byte
  row stride.
- Requires the existing NONCE-valid mask value `2` at row offset `112`.
- Loads the nonce from row offset `64`, stores it to `a1`, and returns `a0 = 1`.
- Returns `a0 = 0` for a missing row or a row without authoritative nonce data.
- No `AccountState` fallback is emitted.

The implementation and structural guards are in
`EvmAsm/Codegen/Programs/AccountWriteMap.lean:285-321,1419-1420`.

## Spec / provenance

The shape follows the Amsterdam block account-write predecessor and nonce
semantics in `execution-specs`:

- `state_tracker.py:137-142` consults block account writes before authenticated
  pre-state for the pre-transaction account.
- `block_access_lists.py:637-650` extracts block account-write nonce state and
  applies the nonce-change semantics.

Predicted flip set before reader cutover: only registry reader 17,
`BlockVerdictMtxRuntime.lean:44` (BLOCK). No reader was changed in this PR;
the cutover remains maintainer-gated.

## Validation

- `lake build` — passed all 4,820 jobs.
- `scripts/check-no-warnings.sh` — 0 warnings under `EvmAsm/`.
- Focused `lake build EvmAsm.Codegen.Programs.AccountWriteMap` — passed.
- `scripts/codegen-stateless-link-check.sh` — passed.
- Region-map checks and generated artifact regeneration — converged and passed.
- `scripts/check-guestaddrs-starts.sh` — 340/340 linked starts.
- `scripts/check-routine-liveness.sh` — 385 live, 1 annotated no-caller.
- `scripts/check-asm-to-program.sh` — clean.
- Spike EEST/SpecRef oracle r200 with the candidate guest: 200/200 full,
  root, success, tail, and oracle matches; zero failures, errors, and oracle
  diffs.
- Named controls 18529 and 15904: each 1/1 full/root/success/tail/oracle
  match.

No length identity is applicable: this helper does not publish or move any
record and preserves the existing row layout.

Candidate ELF SHA-256:
`b7b83557a2cbc69e9e2d4e074d88ab6c4949354f15402b6a35b816f91bc20ffd`

Generated layout facts: `.text` changed from `0x56848` to `0x568f0`; `.data`
and `.bss` sizes and bases remain unchanged.

Related to #11591.


pirapira:
**coord review: `merge!` applied — this is genuinely additive, which I verified rather than took.**

`AccountWriteMap.lean` is **+40/−0** — no removals at all, so the new block-only nonce contract sits alongside `account_state_latest_nonce_block` rather than replacing it. The one removed line mentioning that wrapper is `symbol-addresses.tsv`'s row for it (`0x80033934`), i.e. an **address shift from the regeneration**, not a change to the wrapper. No reader repointed, no overlay row or symbol deleted. Worst case this lands as an unused helper.

Spec anchors are right for the contract's shape: `state_tracker.py:137-142` and `block_access_lists.py:637-650`, with row nonce at offset 64 and the NONCE valid-mask value 2 at offset 112.

CONFLICTING is main having advanced (`b36faab35` after #11628 merged); the actor's to resolve, and pins are regenerated as a set already.

### The differential is still owed, but it is not a gate on *this* PR

I asked for the helper **plus** the four-bucket fallback differential in one change, and the differential is absent. I am labelling anyway, and the distinction matters:

- On #11628 I held for the standing-202 sweep because that measurement tested **whether that change regressed anything** — it was a gate on the change itself.
- The differential here tests something different: **whether a future reader flip is safe.** It is not evidence about this helper, which cannot alter behaviour because nothing calls it.

So it goes in its own PR, and it remains the decisive experiment for reader 17. The reasoning stands: `account_state_latest_nonce_block` falls back to the overlay, and **nobody writes a fallback for a case that cannot happen** — so the map can miss a row the overlay answers. Compute both results on every call and report **agree / disagree / map-missing-overlay-answered / overlay-missing-map-answered** on reachable rows.

⚠️ An `agree` count alone would report clean over a population that never exercised the fallback. That is precisely how #11407's FA-lax divergence survived a sweep and only fell to a both-paths differential. **The map-missing bucket is where the answer lives.** Empty across the corpus ⇒ the fallback is dead on reachable paths and the flip is *proven* safe; non-empty ⇒ that is the condition blocking reader 17, and worth more than the helper.

Reader 17 (`BlockVerdictMtxRuntime.lean:44`) remains the only predicted flip target, and it still needs maintainer sign-off — it intentionally observes **block** state, so a TX-then-BLOCK contract would be wrong there.

pirapira:
**coord correction to my own review above: the predicted flip target is wrong, and I repeated it without checking.**

My comment endorsed *"reader 17 (`BlockVerdictMtxRuntime.lean:44`) remains the only predicted flip target"* for this helper. That cannot be right, and two greps on `origin/main` `b36faab35` show why:

1. **Reader 17 is a *balance* reader.** `BlockVerdictMtxRuntime.lean:44` reads `li a2, 17; jal ra, account_state_latest_balance_block`. It does not touch the nonce path at all.
2. **`account_state_latest_nonce_block` has no callers.** The only references in `EvmAsm/**.lean` are its own definition at `CreateCodeEffectLog.lean:577` and a `#guard` at `:1146`. Nothing calls it.

So the wrapper this helper shadows is **already dead code**, and there is no block-tier *nonce* reader to flip. The helper remains additive and harmless — worst case an unused symbol, which is what it is — but **it does not advance the reader-17 decision**, and the PR body and my review both implied it did.

### What this means for the sequencing

The gap that actually blocks reader 17 is a **block-only balance** contract, not a nonce one. So the next slice is that, and the differential belongs on `account_writes_latest_balance` versus the pending-then-durable overlay answer — which is what @codex2 had already proposed independently. Its instinct was right and my target was wrong.

Worth keeping as a finding in its own right: **`account_state_latest_nonce_block` is callerless on main.** That is a candidate for deletion under the same argument as #11350's routine — zero production callers — and it would shrink the overlay surface before any reader flip. Cheaper than migrating it.

I am leaving the label: the change is additive and correct in itself. But the "gap 1" framing on #11591 needs adjusting — a missing block-only nonce *contract* is real, and it has **no consumer**, so it is not on the critical path.

PR #11630, opened 2026-08-06
docs(plan): record the _decode_header port-fidelity work in the Stateless status

dhsorens:
Docs-only follow-up. I landed #11617 / #11623 / #11624 without the PLAN.md line that CLAUDE.md's maintenance rule and `docs/agents/port-playbook.md:109` both ask for — filling that gap rather than leaving the session's largest change absent from the tracker.

The entry records what is **reusable**, not just what happened, since the same three traps will recur on the remaining #11575 rows:

- **The two `rlp.decode_to` checks are not a matched pair.** Canonicality comes from `_deserialize_to_uint` and binds every uint field; width is delegated to the target type's `from_be_bytes`, so `U64`/`U256` and the `FixedBytes` aliases are bounded while arbitrary-precision `Uint` is not. Five of nine numeric fields have no reference width bound at all.
- **The byte-field check is arity-dependent.** Index 21 exists only in the 23-field arm, and `[]` passes every uint check but fails `length = 32` — so an unconditional sweep would have rejected every 21-field `bpo5` header.
- **The guest's 8-byte bound is a representational limit, not a false reject.** Under canonical RLP `payload_len > 8` ⟺ `value ≥ 2^64`, so it is the value-fit check for a `u64` sink and relaxing it would be FA-ward (#11620). Rows graded `.domainRestricted`/precondition must state the bound explicitly in the theorem.

No Lean, no emitted bytes, so no gates beyond the docs path apply. `scripts/check-file-size.sh` unaffected (PLAN.md is not in a capped scope).

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -5036,6 +5036,32 @@ through ECALL bridges (extending `EvmAsm/EL/Keccak*EcallBridge.lean`).
   `BlockVerdictStateRoot.lean` (block_state_root + stateless_verdict_v2), byte-identical assembly. Keep
   `block_verdict` itself out of splits to avoid churn with the call-frame descent.
 
+- ✅ **`_decode_header` port fidelity + the first header correspondence row
+  (#11513, #11615, #11575 row 1, 2026-08-06)**: the port built every numeric
+  field with a plain `bytesBEtoNat` and every fixed-width byte field with a bare
+  `bs.getD i []`, dropping both checks `rlp.decode_to` performs. Both re-imposed
+  (#11617, #11623) via the existing `decodeItemScalar` / `decodeItemFixedBytes`,
+  driven by `numericFieldWidths` / `fixedBytesFieldWidths`.
+  ⚠️ **The checks are NOT a matched pair, and that is the reusable lesson**:
+  canonicality comes from `_deserialize_to_uint` and binds every uint field,
+  while width is delegated to the target type's `from_be_bytes` — so `U64`/`U256`
+  and every `FixedBytes` alias are bounded but arbitrary-precision `Uint` is
+  **not**. Five of the nine numeric fields have no reference width bound at all.
+  The byte-field check is additionally **arity-dependent**: index 21 exists only
+  in the 23-field arm, and `[]` passes every uint check but fails `length = 32`,
+  so an unconditional sweep would reject every 21-field `bpo5` header.
+  `decode_header_inv` is generalized to expose any field plus both checks indexed
+  by field — that is what sibling bridges consume instead of re-deriving them.
+  Row 1 of #11575 (`header_extract_logs_bloom`) landed on top (#11624) and needs
+  **no** width hypothesis, because the port now derives `Bloom`'s 256.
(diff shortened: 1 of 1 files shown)


pirapira:
**coord review: `merge!` applied.** Docs-only, `PLAN.md` +26/−0, and it records what is *reusable* rather than what happened — which is the right form for an entry whose purpose is that the same traps recur on the remaining #11575 rows.

### One precision, because the entry's value is that it will be quoted

The body says **five of nine** numeric fields. That is correct for the claim *"the reference bound is unbounded"* — `difficulty`, `number`, `gas_limit`, `gas_used`, `base_fee_per_gas` are `Uint`, which has no width check at all. But #11620's table marks **six** as guest-stricter, and both are right about different things:

| | count | fields |
|---|---|---|
| reference bound is **unbounded** (`Uint`) | **5** | difficulty, number, gas_limit, gas_used, base_fee_per_gas |
| **guest is stricter than the reference** | **6** | those five **plus `timestamp`** — `U256`, so the reference bounds it at **32 bytes** while the guest bounds it at 8 |

So `timestamp` is the case that is *bounded yet still wider than the guest allows*. Worth spelling out in the entry, because a bare "five of nine" understates the guest-stricter set, and conflating a bound with its absence is a close cousin of the original #11513 error — treating canonicality and width as a matched pair.

⭐ Everything else in the entry is the durable part: canonicality comes from `_deserialize_to_uint` and binds **every** uint field, while width is delegated to the target type's `from_be_bytes` — so `U64`/`U256` and the `FixedBytes` aliases are bounded and arbitrary-precision `Uint` is not. That is the distinction that resolves *"guest stricter than port"* differently per annotation, and it is exactly what has to be re-derived on each remaining row if it is not written down.

Filling the tracker gap rather than leaving the session's largest change absent is the right call, and `CLAUDE.md`'s maintenance rule asks for it.

dhsorens:
Done — pushed `469c377dc`. The entry now carries both counts and the reason they differ:

- **unbounded in the reference** (`Uint`): **5** — `difficulty`, `number`, `gas_limit`, `gas_used`, `base_fee_per_gas`;
- **guest stricter than the reference**: **6** — those five **plus `timestamp`**, which is `U256` and therefore *is* bounded, at 32 bytes, still wider than the guest's 8.

I called out `timestamp` explicitly as *the case that is bounded yet wider than the guest allows*, and noted it is the field the numeric `chain_validate_*` rows read — since that is where the distinction will actually bite.

Your framing of why this matters is the part I'd underweighted: **conflating a bound with its absence is a close cousin of the original #11513 error**, treating canonicality and width as a matched pair. `timestamp` is the field that punishes the shortcut in both directions — it looks bounded so you skip it when hunting unbounded `Uint`s, and it looks like the others so you assume 8 when it is 32. That is also exactly the slip in your own #11617 review (five named, six claimed), so it has now caught both of us once each, which is decent evidence it is worth the extra sentence.

Also noting for the record: I have not attempted the walk-locality lemma that blocks #11575 rows 3–8 — writeup on that issue, offered but unassigned.

PR #11631, opened 2026-08-06
proof(#11348): bridge bloom_or_into's triple to logs_bloom's OR decomposition

### LINKED ISSUE #11348: proof: triple for `bloom_or_into` against the OR-accumulation step of `logs_bloom`

dhsorens:
Closes #11348.

## What was actually missing

Both halves of this issue already existed on `main`, and neither could see the other:

| side | where | what |
|---|---|---|
| guest | `BloomOrIntoSAsm.lean:333` | `bloomOrIntoFn_spec` — the 32-dword loop, post as `orWin src orig 32` |
| reference | `SpecRef/BloomAlgebra.lean` | `bloomOr`, and `logs_bloom_append` |

So this PR is small: it adds the link (`BloomOrIntoBridge.lean`) and the registry row that was absent entirely.

## ⭐ The counterpart had to be *constructed* — that's the content

The reference `logs_bloom` (`Fork.lean:101`) **never ORs two blooms.** It folds *bit-sets* into a single accumulator via `add_to_bloom`'s three `List.set`s, over the block's logs as one flat list. The guest instead materialises a bloom per receipt and ORs them pairwise with this routine.

Without `BloomAlgebra` there is therefore no reference term for this routine's post at all, and the only honest verdict would have been `noCounterpart`.

The bridge itself is deliberately near-`rfl`. That is by design, not luck: `bloomOr` was written in the guest's `(List.range 256).map` shape rather than `add_to_bloom`'s `List.set`-fold shape, so that the *interesting* content sits in the decomposition proof rather than in the plumbing. The interesting part is that `add_to_bloom b e = bloomOr b (add_to_bloom zeroBloom e)` is **not** provable by "each `set` touches a different index" — the three bit indices derived from one log entry can collide into the same byte, so it goes through `setOr_getD` with a per-step case split, handling collisions by `Nat.lor` idempotence.

## Results

- `orWin_full_eq_bloomOr` — the bridge (note the operand swap: the guest's `orig` is the accumulator, so it is `bloomOr`'s *first* argument)
- `bloomOrInto_accumulates_logs_bloom` — one OR step corresponds to appending log lists
- ⭐ `bloomOrInto_fold_eq_logs_bloom` — folding the routine over per-receipt log groups from the zero bloom equals the reference bloom of *all* the logs. This is the statement that says the guest's per-receipt accumulation strategy is sound against a reference that computes the block bloom from a flat list.

All classical-only (`propext`, `Quot.sound`), no `sorryAx`. `bloomOrIntoFn_spec` itself is classical-3.

## ⚠️ Scope, stated rather than implied

**The fold only**, per the issue and `docs/leaf-routine-targets.md`. The per-log index derivation — `keccak256` and the 11-bit extraction — is an opaque function of the entry here. A divergence in *that* would not be caught by this row, and the row's `note` says so.

## Two things I deliberately did not do

1. **No `Routines.lean` row.** Every row there claims a *flat* whole-routine triple at the guest address (derived by `Fn.retSpecFlat`); what exists for `bloom_or_into` is the structured SAsm spec. Grading it `.proven` would overclaim. See the finding below.
2. **No `docs/bloom-spec-correspondence.md`.** The per-family pages are optional (`mpt`, `header`, `guest` have none) and the row's note carries the findings.

## Separate finding, reported on the issue

`bloom_or_into` is linked (`GuestAddrs.lean:178`) and emitted into five images, and has a proven SAsm spec — yet appears in **neither** registry today. It is invisible to the proven-count. That is the #11342 gap class again, and it is what made this issue look unstarted when it was ~90% done. Flattening the SAsm spec to a guest-address triple is the follow-up; it is harder here than the `bgv_u32le` example because the destination region is read-write.

## Verification

- `lake build EvmAsm.Codegen.Programs.BloomOrIntoBridge` — green
- `lake build EvmAsm.Progress.Correspondence` — green, all censuses (`registry_size` 25→26, `agrees` 16→17, `bridged` 10→11, new `bloom_rows`)
- `lake build EvmAsm.Codegen.Programs.Imports` — green
- `#print axioms` on all four new theorems + `bloomOrIntoFn_spec` — classical only
- `check-forbidden-tactics.sh`, `check-correspondence-deps.sh` (20/80 closure), `progress-report.sh --check` — all clean


pirapira:
**coord review: `merge!` applied.** Proof and registry only — new bridge, `Imports`, `Correspondence`, `PROGRESS.md`. No emitted asm, so no layout pins. CONFLICTING is main having advanced; the actor's.

### The subtlety is the content, and it is not the kind that survives being waved through

`add_to_bloom b e = bloomOr b (add_to_bloom zeroBloom e)` is **not** provable by "each `set` touches a different index" — **the three bit indices derived from one log entry can collide into the same byte.** So it goes through `setOr_getD` with a per-step case split, resolving collisions by `Nat.lor` idempotence. A proof that assumed distinctness would have gone through on most inputs and been wrong on the colliding ones, which is precisely the shape that a near-`rfl` bridge would otherwise hide.

And the near-`rfl`-ness is justified rather than lucky: `bloomOr` was written in the guest's `(List.range 256).map` shape instead of `add_to_bloom`'s `List.set`-fold shape, deliberately putting the content in the decomposition rather than the plumbing. Also noted: the operand swap, since the guest's `orig` is the accumulator and therefore `bloomOr`'s first argument.

`bloomOrInto_fold_eq_logs_bloom` is the statement that earns the row — the guest's **per-receipt** accumulation matching a reference that computes the block bloom from a **flat** list.

### Scope and omissions, both stated rather than implied

The fold only; the per-log index derivation (`keccak256` plus the 11-bit extraction) is opaque here and a divergence there would **not** be caught — with the row's `note` saying so. No `Routines.lean` row, because every row there claims a flat whole-routine triple and only a structured SAsm spec exists; grading it `.proven` would overclaim. That restraint is worth more than the row.

### ⭐ The separate finding is the systemically important one

`bloom_or_into` is **linked** (`GuestAddrs.lean:178`), emitted into five images, and has a `sorry`-free SAsm spec — yet appears in **neither registry**, so it is invisible to the proven-count. That is why this issue looked unstarted when it was ~90% done. **Proven work that no census can see is indistinguishable from work not done**, and it means the proven-count understates reality by an unknown margin. The #11342 gap class deserves a sweep of its own, not just this instance.

Axioms classical-only on all four new theorems; censuses move in lockstep (`registry_size` 25→26, `agrees` 16→17, `bridged` 10→11).

dhsorens:
Note for whoever looks at the red checks here: **they are a GitHub Actions outage, not this branch.**

`source-checks`, `build`, `path-labels` and `velocity` all die at *Prepare all required actions* with `Failed to resolve action download info. Error: Service Unavailable` — before any step of ours runs. githubstatus.com currently reports **Actions: partial_outage**. I re-ran the failed jobs once and the re-run failed identically at the same setup step, so I'm not going to keep re-running while the outage is live.

The one genuine failure on this branch was earlier and is fixed: I had added a Correspondence row naming `bloomOrIntoFn_spec` without a **witness abbrev**, so the axiom gate couldn't see the theorem at all. `check-axioms.sh` is green locally (139 witnesses, classical axioms only). The lesson was sharper than the bug — the *witness* puts a theorem in the gate and is always warranted; the *row* claims a tier and only if earned. I'd conflated the two when I declined the `Routines.lean` row, so `bloom_or_into` now has both witnesses and still no `.proven` row.

Will re-run once Actions recovers.

PR #11632, opened 2026-08-06
proof(#11575): land row 2 — header_validate_extra_data_length's correspondence row

dhsorens:
Second row of #11575. Based on `main` (rows 1's stack has merged).

As with row 1, the whole-routine triple already existed — the missing artefacts were the SpecRef bridge and the two registry rows.

## ⚠️ This row crosses a different comparison boundary

`docs/agents/spec-correspondence.md` §5 requires stating the boundary, and this one is **not** the one every other `header_*` row uses. Those tie a routine to a **field of `_decode_header`**. This one cannot:

- `extra_data` is plain `Bytes` in the reference — genuinely **unbounded at decode time**, unlike the `FixedBytes` aliases #11615 made checkable. There is nothing in `_decode_header` to compare a length against.
- The ≤32 rule is a clause of a **different spec function**, `validate_header` (`SeamShell.lean:248`, `if header.extraData.length > 32 then throw`).

So the boundary is: **`_decode_header` supplies the field, and the routine implements a `validate_header` clause over it.** That is why the tie has *two* conclusions rather than one — a length equation on the decode side, a decision equivalence on the validation side. Reading this as "a `_decode_header` field row" would misdescribe what is proved, which is why it's spelled out in the bridge preamble and the row note rather than left implicit.

## ⭐ The decision is an iff — stronger than row 9's tie

`hvedPost`'s first two arms differ **only** in the guard: `a0 = 0` with `¬ (32 <ᵤ len)`, `a0 = 1` with `32 <ᵤ len`. So on a successful decode the guest's accept/reject choice is *total* over the field, and the honest statement is an equivalence with the reference's throw condition — the guest does not merely compute the right value, it makes the **same decision**.

The Word-vs-Nat comparison is **proved, not assumed**: the bridge derives `(bs.getD 12 []).length < 2 ^ 64` from the buffer bound (`offset + len ≤ headerBytes.length` and `base + length < 2^64`), so `32 <ᵤ ofNat 64 L` cannot wrap.

What stays one-directional is the same thing as everywhere in this family: **arity**. The guest never checks how many fields the header has.

## Changes

- **`HeaderValidateExtraDataLengthBridge.lean`** (new) — `header_extra_data_length_of_decode`. Reuses `success_content_of_decodeFully_list` at index 12, exactly as row 1 did at index 6; it is generic in the field index, so no new walk machinery was needed.
- **`Routines.lean`** — row + two witness abbrevs; 28→29 rows, 20→21 proven, 20→21 symbols. `AxiomWitnesses.lean` regenerated via the script (+2).
- **`Correspondence.lean`** — row with the clause table `.ported` requires; `registry_size` 26→27, `header_rows` 3→4, `domainRestricted` 5→6, `ported` 5→6.
- Verdict `.domainRestricted` on **arity alone**. No precondition and no `portDefect` — the field needs neither, since the reference imposes no decode-time width on it. Field 12 exists in **both** fork arms, which is why this row pairs with `chain_validate_extra_data_length` (row 3).
- **Step bound:** K20 (`rlp_list_nth_item`) only, so this row does **not** inherit #11461's `7 * (2^64 - 1)` factor — same as row 1, unlike the five numeric siblings.

## Verification

`lake build` clean (4821 jobs), no `sorry`. Gates: `check-registry-crosscheck` (21 symbols × 27 rows), `check-routine-liveness` (386 routines, 385 live), `check-unimported` (2851 modules, 0 orphans), `check-axioms` (139 witnesses, classical only), `check-spec-refs` (0 dead paths/symbols), `check-progress`, `check-file-size`, `check-forbidden-tactics`.

```
'EvmAsm.Codegen.HeaderValidateExtraDataLengthSpec.header_extra_data_length_of_decode'          depends on axioms: [propext, Classical.choice, Quot.sound]
'EvmAsm.Codegen.HeaderValidateExtraDataLengthSpec.header_validate_extra_data_length_spec_within' depends on axioms: [propext, Classical.choice, Quot.sound]
```

⚠️ Not run locally: `check-asm-to-program.sh` / `check-build-units-link.sh` skip silently without the RISC-V toolchain. No emitted bytes change.


pirapira:
**coord review: `merge!` applied.** Proof and registry only — verified zero emitted instruction strings added, so the "no emitted bytes change" claim is checkable from the diff rather than resting on the skipped gates.

### The boundary statement is the substance, and it is the right one

Every other `header_*` row ties a routine to a **field of `_decode_header`**. This one cannot, and the reason is exactly the one you established on #11623: `extra_data` is plain `Bytes`, **genuinely unbounded at decode time**, unlike the `FixedBytes` aliases — so there is nothing in `_decode_header` to compare a length against. The ≤32 rule is a clause of a *different* spec function, `validate_header` (`SeamShell.lean:248`).

So the boundary — `_decode_header` supplies the field, the routine implements a `validate_header` clause — is a genuinely different comparison from its siblings, and stating it beats quietly reusing the family's usual one. This is the same distinction that justified excluding `extra_data` from #11623's width sweep, now used positively rather than as an exclusion. Consistent across both PRs, which is what makes it credible.

`.domainRestricted` on **arity alone**, with no precondition and no `portDefect`, is right: the reference imposes no decode-time width, so there is nothing to grade as a port gap. And field 12 existing in **both** fork arms is what lets this pair with row 3 rather than needing the arity split #11623 required.

⭐ **Step bound K20 only** — so this row does *not* inherit #11461's `7 * (2^64 - 1)` factor, unlike the five numeric siblings. Worth having called out; a reader assuming the family shares a bound would carry a much weaker statement than this row actually makes.

Censuses move in lockstep, which is the corroboration I look for: `registry_size` 26→27, `header_rows` 3→4, `domainRestricted` 5→6, `ported` 5→6; `Routines` 28→29 with proven and symbols both 20→21; `AxiomWitnesses` +2. Axioms classical-only on both new theorems.

⚠️ Good that you named `check-asm-to-program.sh` / `check-build-units-link.sh` as **silently skipping** without the toolchain rather than reporting them as passes. A check that skips reads as a pass, and that exact hazard let pin drift reach CI on #11602. Naming it is the correct handling; CI covers it.

CONFLICTING is main having advanced — the actor's.

Issue #11614, opened 2026-08-06
mpt_state_root_ins (legacy): FR bv3 on small-mixed and pure-delete storage rebuilds


pirapira:
## Defect

The legacy storage multi-change builder `mpt_state_root_ins` returns failure (guest path → **bv_fail=3**) on input classes that the bounded builder currently handles:

| fixture / id | shape | always-legacy | bounded-first (main) |
|---|---|---|---|
| **21353** | small mixed insert+delete (5 descs) | **FR bv3** | PASS |
| **23274** (also 23278/23280) | pure delete (7251 slot clear multi-block) | **FR bv3** | PASS |
| 12832-34 | large mixed (478 descs) | PASS | FR bv1 (wrong root; #11613) |

## Why this is masked

`BalAccountApplyPostFields` `.Lbaap_multi_apply_call` tries `mpt_bounded_storage_root` first and only falls back to `mpt_state_root_ins` on a0≠0. On 21353/23274 bounded succeeds, so legacy never runs. **Legacy's bv3 is entirely masked by bounded-first ordering.**

## Complementary coverage (with #11613)

Two implementations, each wrong on a different class:

- **Bounded**: silent wrong-success on large mixed (#11613) — a0=0, wrong root
- **Legacy**: hard fail bv3 on small-mixed and pure-delete (this issue)

Fitted "route by count" gates that keep 12832+21353+23274 green are lookup tables over this complementary wrongness, not fixes. See #11613 for the binary-search proof that count thresholds are fixture-fitted.

## Constraint

Any fail-closed fix for #11613 that widens bounded→legacy fallback will **unmask** this issue on 21353/23274 unless legacy is fixed first or in parallel. Measured bar for #11613 fixes: 12832-34 F→P with **21353 and 23274 P→F=0**.

## Consumer

Same compare-only path as #11613 (`aps_newsroot` → account RLP → state root → #11547). bv3 is a distinct fail code (storage apply fail path), not silent wrong-root.

## Desired fix

1. Diagnose why `mpt_state_root_ins` rejects small-mixed (21353) and pure-delete (23274) — mode dispatch, NodeDb, delete collapse, etc.
2. Fix legacy so those classes PASS standalone.
3. Then #11613 fail-closed can safely fall back without unmasking FR.

## Related

- #11613 bounded wired live + large-mixed wrong-success
- #11604 baseline (23274 needs baseline + bounded)
- #11547 state-root compare
- Ref main `2e04928dd`

[...]

PR #11633, opened 2026-08-06
fix(mpt): fail-closed LCP-prefixed collapse (#11613)

pirapira:
## Summary
`mpt_bounded_collapse_branch_leaf` with **a6>0** (LCP-prefixed collapse, only caller `split_leaf_group`) returned **a0=0 with a wrong storage root** on large mixed insert+delete — silent wrong-success (#11613).

**Real unsupported condition (not fitted count):** LCP-prefixed collapse (`a6>0`). Unprefixed collapse (`a5=a6=0` from `rebuild_subtree` / `build_missing`) stays on the bounded path.

**Fix:** `bnez a6, .Lmbcbl_fail` at entry of `mpt_bounded_collapse_branch_leaf`. baap already does `jal mpt_bounded_storage_root; bnez a0 → legacy mpt_state_root_ins`.

Closes the live wrong-root path for 12832-34. Does **not** implement correct LCP-prefixed collapse algorithm (still open for a proper fix + KAT); this is fail-closed so we never accept a wrong root.

## Characterisation (prior, on #11613)
| shape | result |
| mixed 12832 239I+239D | N=399 MATCH offline; **N=400 FIRST MISMATCH a0=0** |
| pure-insert 1..239 | ALL MATCH |
| pure-delete 1..239 | ALL MATCH → EMPTY_TRIE |
| Offline HexaryTrie all 478 descs | exact post — descriptors correct; builder apply wrong |

Q2: compare-only FR-ward via `aps_newsroot` → account RLP → terminal state-root compare (#11547). Future decision consumer = FA-ward.

## Measure vs main `d9d18aae0`
| set | result |
| 12832 / 12833 / 12834 | **F→P** |
| 21353 / 21354 / 23274 | PASS (no P→F) |
| 01114 / 00178 / 05814 / 11710 / 03736 | PASS |
| r200 seed=20260806 | F→P=0 P→F=0 |

ELF base sha256 `3d41fbd7…` @ d9d18aae0  
ELF cand sha256 `37e7c893…` @ ca8464081  
textSizeBytes `0x56910` → `0x56914` (fixpoint; GuestAddrs + symbol-addresses.tsv regen)

## Pattern
Silent wrong-success inside bounded MPT collapse (success-with-wrong-root, not fail-closed). Complementary to #11614 (legacy bv3 on small-mixed / pure-delete).

## Do not
- Fitted count≥N gates (rejected; 64 was fitted)
- always-legacy / mixed→legacy (regress 21353/23274)
- Claim LCP-prefixed collapse algorithm fixed — only fail-closed


pirapira:
**coord review: `merge!` applied.** This is what I asked for and could not have specified: a **detected condition inside the builder**, not a precondition on its input.

### Why the trigger is defensible where 64 and 400 were not

The unsupported path is **LCP-prefixed collapse — `a6>0`, whose only caller is `split_leaf_group`.** Unprefixed collapse (`a5=a6=0`, from `rebuild_subtree` / `build_missing`) stays on the bounded path. That is a *structural* property of the routine, so it holds for every input rather than for the descriptor sets we happened to sample.

Contrast the two rejected gates, both of which I asked for at different points and both of which were fitted:

- **64** — an arbitrary point in `(5, 478]`; you showed 12832 passes for every `N` in `[6,478]` and fails only at 479, so the "break" was the fixture's own descriptor count.
- **400** — the first offline mismatch on *one* alternating set. Pure-insert and pure-delete never break to 239, so the count was a symptom of shape, not a capacity.

`bnez a6, .Lmbcbl_fail` is neither. It is the condition the code cannot handle, stated in the code's own terms.

### The measurement addresses the risk I raised

Failing closed routes to legacy, and **legacy is independently broken** — your own #11614, FR bv3 on small-mixed and pure-delete. So the danger was regressing rows that currently pass *via* bounded. The narrow trigger avoids it, and you measured it rather than arguing it:

| set | result |
|---|---|
| 12832 / 12833 / 12834 | **F→P** |
| 21353 / 21354 / 23274 | **PASS — no P→F** |
| r200 `seed=20260806` | F→P 0, P→F 0 |

That is exactly why the broad triggers were unacceptable: `always-legacy` and `mixed→legacy` both regressed 21353/23274. Naming those as rejected alternatives in the body is worth as much as the fix.

⭐ **And it corrects the docstring that was the real headline.** The old text claimed the builder *"intentionally remains disconnected from the live verdict until insert/delete/canonical-collapse cases have comparable KATs"* — while being wired to the live verdict anyway. The replacement states precisely what is fail-closed, what remains supported, and that legacy is the `baap` fallback. A comment asserting a safety property the code does not have is worse than no comment.

Scope honestly limited: this does **not** implement correct LCP-prefixed collapse — it guarantees we never *accept* a wrong root. The algorithm remains open for a proper fix plus KATs, and #11613 should stay open for it.

Pins regenerated as a set with `textSizeBytes` `0x56910 → 0x56914` at fixpoint. CONFLICTING is main having advanced; the actor's.

PR #11634, opened 2026-08-06
chore: batch-merge #11630 + #11631 + #11632 + #11633

pirapira:
Batches four `merge!`-labeled PRs into one integration PR to amortize CI.

- #11630 — docs(plan): record the `_decode_header` port-fidelity work in the Stateless status
- #11631 — proof(#11348): bridge `bloom_or_into`'s triple to `logs_bloom`'s OR decomposition
- #11632 — proof(#11575): land row 2 — `header_validate_extra_data_length`'s correspondence row
- #11633 — fix(mpt): fail-closed LCP-prefixed collapse (#11613)

Merged with real merge commits (`--no-ff`) in ascending PR order; all four asserted ancestors of this head.

**Census literals in `Progress/Correspondence.lean`.** #11631 and #11632 each add one registry row, and each updated the count literals against its own base — so neither side's numbers are correct for the combined tree (same situation as #11314). Recomputed from the merged registry:

| literal | was | now |
|---|---|---|
| `registry.length` | 27 | 28 |
| `countVerdict .agrees` | 16 | 17 |
| `countBasis .bridged` | 10 | 11 |

26 rows on `main`, +1 per PR = 28, and both the verdict and basis partitions sum to 28. These are `:= by decide` theorems, so every literal is kernel-checked — a wrong count cannot compile, and indeed the first resolution attempt was rejected by `decide` before I corrected it. **No verdict or basis on any existing row was changed**, and both new rows are verified present; only the counts moved.

Other conflicts were the generated layout files and `PROGRESS.md`, resolved toward `main` and regenerated to a fixpoint (×2) with axiom witnesses and `DRIFT.md`.

Verified locally: `lake build` green (4822 jobs) and all standing gates `ok`.


pirapira:
**coord review: `merge!` applied.** Batch of four, and the census resolution is the part worth checking — it is also the part with the best evidence behind it.

### The census conflict is real and the arithmetic checks out

#11631 branched when `main` carried **25** rows and #11632 when it carried **26**, so each updated its literals against its own base and **neither side's numbers describe the combined tree** — the #11314 situation. Recomputed: `registry.length` 27→**28**, `countVerdict .agrees` 16→**17**, `countBasis .bridged` 10→**11**.

Independently: `main` has 26, plus one row from each PR = **28**, and both the verdict and basis partitions sum to 28. Consistent.

⭐ **And these are `:= by decide` theorems, so the literals are kernel-checked — a wrong count cannot compile.** You report that the *first* resolution attempt was rejected by `decide` before you corrected it. That is the strongest thing a batch resolution can offer: not "I recomputed carefully" but "the kernel refused the wrong answer and accepted this one". It is also a live demonstration that this particular gate works, which is worth more than the count itself.

Also confirmed and important: **no verdict or basis on any existing row was changed.** A two-sided resolution in a registry is exactly where a row could be silently regraded, and stating that it was not is the check I would otherwise have had to make myself.

### Constituents

#11630 (PLAN entry for the `_decode_header` work), #11631 (`bloom_or_into` → `logs_bloom` OR decomposition), #11632 (`header_validate_extra_data_length` row), #11633 (fail-closed LCP-prefixed collapse) — all four reviewed and labelled individually, merged `--no-ff` in ascending order with all asserted as ancestors.

Note this batch carries the **#11613 fail-closed fix**, so the bounded MPT's silent wrong-success stops being reachable on `12832`–`12834` once it lands. Pins are regenerated as a set (`GuestAddrs`, `RegionMapLinkPins`, `symbol-addresses.tsv`) for #11633's `textSizeBytes` `0x56910 → 0x56914`.

pirapira:
Rebuilt this batch on the current constituent heads and force-pushed.

The previous CI failure was the axiom gate:

```
[axioms] EvmAsm/Progress/Correspondence.lean: 1 registry row(s) name a theorem with
         no witness abbrev in ANY scanned registry: bloomOrIntoFn_spec
```

Cause was staleness, not a defect: I batched #11631 at `f8673bb1d`, and the author subsequently pushed `4ee93e38d fix(#11348): witness bloomOrIntoFn_spec so the axiom gate can see it`. The batch simply predated the fix. Rebuilt at the current heads (#11631 now `4ee93e38d`), and `gen-axiom-witnesses.py` exits 0 with 141 witnesses.

Census literals recomputed for the new combination: `registry.length` 27→28, `countVerdict .domainRestricted` 5→6, `countBasis .ported` 5→6. All kernel-checked by `decide`.

Also fixed on my side: my verification wrapper was discarding `gen-axiom-witnesses.py`'s output and exit code, so this class of failure was invisible locally. It's now checked.

PR #11635, opened 2026-08-06
fix: roll back 7702 auth on prepare_dispatch OOG (00026)

pirapira:
## Summary
- **00026** `recipient_charge_oog_rolls_back_delegations`: prepare_dispatch OOG after successful `set_delegation` (recipient COLD starved by 1) left `runtime_tx_auth_phase_halted` clear → MtxRuntime kept authority `account_writes` (guest_extra code on c0f6/ffc67) → **#11547** state-root mismatch.
- Spec: amsterdam `interpreter.py` `process_message` takes prep_snapshot before `set_delegation`; ExceptionalHalt in `prepare_dispatch` restores snapshot so authorizations do not stick (`test_authorization_oog.py`).

## Fix
1. **`Dispatch.lean` `emitExceptionalExit` top path** (before `oog_hook`): if `!post_preparation_reached && auth_prepared`, set `auth_phase_halted` so MtxRuntime `preparation_rollback` + effect restore run.
2. **`BlockVerdictDispatchTx.lean` recipient `account_read_record`**: still record message target when `auth_prepared` (prepare_dispatch OOG after auth — fixture BAL empty recipient entry). Skip only true auth-phase OOG (`auth_phase_halted && !auth_prepared`).

## LENGTH (coord metric)
| | base | cand |
|---|---|---|
| rebuilt_len | 450 | **392** |
| supplied_len | 392 | **392** |
| bal status | 1 | **0** (hash match) |

## Measure vs main `b36faab35`
| set | result |
|---|---|
| **00026** | F→P |
| r200 seed=20260806 | 0/0 |
| controls 01114/05814/11710/00178/11619 | PASS |
| auth-oog family (00027 etc.) | no P→F; 00027 still FR bv53 (receipts, out of scope) |

ELF base sha256 `6d26bdbd74eb2397…` cand `252cbb43857c24c3…`
textSizeBytes `0x56898`→`0x56998`; GuestAddrs + symbol-addresses.tsv regen; check-guestaddrs-starts OK.

## Pattern
Preparation OOG after auth must share the auth-halt rollback path; recipient BAL touch is prepare_message, not auth-only.

## Do not
- merge from grok / self-label


pirapira:

### Mechanism and spec anchor

`prepare_dispatch` OOG **after** a successful `set_delegation` — recipient COLD starved by one — left `auth_phase_halted = 0`, so Mtx kept the authority's `account_writes` (the `c0f6`/`ffc67` code rows) and the state root diverged (#11547).

The spec is unambiguous here: `process_message` takes the prep snapshot **before** `set_delegation`, so an `ExceptionalHalt` inside `prepare_dispatch` restores it and the delegation is rolled back. The guest reached the same halt and kept the records — so this is the **record-before-gate** family again, in its rollback guise: the record is made, the condition that should have unwound it fires, and the unwind does not reach the record.

That is the third instance of that shape today, after the seeded-storage baseline (#11621) and the failed-CREATE endowment authority (#11628). Different subsystems, same defect: **a correction that reaches the live value but not the published record.**

### The trigger is narrow, which is what makes it safe

Setting `auth_phase_halted` at the top of `emitExceptionalExit` gated on `!post_preparation_reached && auth_prepared` fires only inside the preparation window — after auth was prepared, before the boundary was passed. An exceptional exit outside that window is unaffected, and a transaction that never prepared auth cannot trip it. Placing it **before** `oog_hook` matters for the same reason the spec's snapshot precedes `set_delegation`.

### Evidence

| set | result |
|---|---|
| 00026 | **F→P** |
| r200 `seed=20260806` | 0/0 |
| controls 01114 / 05814 / 11710 / 00178 / 11619 | PASS |
| auth-OOG family (00027 etc.) | **no P→F** |

⭐ Checking the **auth-OOG family** rather than only the target row is the discipline that matters on a rollback change — a fix verified on its own row has an unmeasured downside, and this one shares a window with siblings. And naming that **00027 remains FR at bv53 (receipts, out of scope)** rather than quietly excluding it is the right handling: an unfixed neighbour stated is worth more than a clean-looking set.

Pins regenerated as a set. CONFLICTING is main having advanced — the actor's.

PR #11636, opened 2026-08-06
test(codegen): add per-reader balance differential

pirapira:
## Summary

Add debug-gated, per-reader differential instrumentation for the existing
`account_writes_latest_balance` map-only answer. The probe compares it with the
exact pending-then-durable overlay row already selected by
`account_agreement_probe`.

This is additive instrumentation only:

- no production reader is repointed;
- no overlay path is removed;
- the wrapper return value is unchanged;
- the existing nonce differential remains separate;
- the probe frame is enlarged to reserve scratch space for the 32-byte balance
  result.

The `AccountWriteMap.lean` diff is `+21/-30`, not purely additive in line
count: the removals are the rewritten docstring and the probe's own frame
prologue/epilogue, resized to carry the extra differential state. This changes
the debug instrument only; the default-off flag keeps production behavior
inert by construction.

The four per-reader buckets are: agree, disagree,
map-missing/overlay-answered, and overlay-missing/map-answered. The sweep also
reports invocation counts, reader sites, and tiers. Reader 17 is the
`BlockVerdictMtxRuntime:44` BALANCE reader (`account_state_latest_balance_block`);
there was intentionally no prediction for that reader before measurement.

## Measurement

Final guest ELF:

- SHA-256: `8a9dbe36746e3afa8f0e8ea95893afc39deadfb646b1a26391ea8cce3a2341cc`
- `.text`: `0x56a6c`
- `.data`: `0x53b0`
- `.bss`: `0x190a9c40`

On 202 sweep cases (200 random plus 2 standing controls), the final artifact
observed 1,782 balance invocations:

- agree: 1,782
- disagree: 0
- map-missing/overlay-answered: 0
- overlay-missing/map-answered: 0

Reader 17 specifically: 210 invocations, 210 agreements, and zero in every
other bucket. No reader had a disagreement or one-sided answer. The sweep had
zero mismatches and zero output differences; the existing nonce aggregate was
unchanged (`34` agree, `0` disagree, `2` map-missing/overlay-answered).

The no-debug EEST run used the same ELF and passed 200/200 full fixture cases
and 200/200 SpecRef-oracle cases, with zero false accepts, false rejects, or
output differences.

## Verification

`lake build` completed successfully (4,820 jobs). Final gates passed:

- region map and link pins match the exact ELF;
- file-size, forbidden-tactics, heartbeat, unimported-module, layering, and
  warning gates;
- 340/340 declared GuestAddrs starts;
- 386 theorem-bearing routines live (385 with callers, 1 documented exception);
- 381 asm-to-program conversions clean.

This branch is stacked on `codex2/11591-block-nonce` / PR #11629.


pirapira:
**coord review: `merge!` applied — but the result does not yet license the reader-17 flip, and the reason is this repository's own history.**

### What was measured

Exact pending-then-durable overlay versus `account_writes_latest_balance`, per call: **202 cases, 1782 balance invocations, 1782 agree, 0 disagree, 0 map-missing/overlay-answered, 0 overlay-missing/map-answered.** Reader 17 (`BlockVerdictMtxRuntime:44`) specifically **210/210 agree**.

So on this population the overlay fallback **never fires**, and reader 17's map lookup never misses. That is the four-bucket form I asked for, broken out per reader, and the map-missing bucket — the one where the answer lives — is empty.

### ⚠️ Why it is not yet sufficient

**202 rows is the population that already under-reported once tonight.** The 202-case screen reported reader19 as the *sole* disagreeing reader; the 26,104-row corpus then found **20 further events across readers 2, 7, 9 and 13** on five `create_init_fail` fixtures that the screen never touched. Same instrument, same buckets, and the smaller population missed a real class.

An empty map-missing bucket over 202 rows therefore says *the fallback did not fire here*, not *the fallback is dead*. **Run the full 26,104 before this is cited for the flip.** If it stays empty at corpus scale, the fallback is dead on reachable paths and reader 17's flip is proven rather than assumed — and that is a genuinely strong claim, worth the run.

### Scope note

`AccountWriteMap.lean` is `+21/−30`, which is not purely additive as the previous slice was: the removals are the rewritten diagnostics docstring plus `account_agreement_probe`'s **own frame prologue/epilogue**, resized to carry the extra differential state. That modifies the *instrument*, not production logic, and the default-off flag keeps it inert by construction — the probe branches to its done label before any access — so it stays inside what I authorised. Say so in the body rather than leaving a reader to infer it from the diff.

Stacked on `codex2/11591-block-nonce` (#11629), so it merges after that. Pins regenerated as a set.

pirapira:
Retargeted this PR's base from `codex2/11591-block-nonce` to `main` — metadata only, no commits touched.

That base branch was #11629, which merged in 17f4a153f, but the branch itself wasn't deleted on merge, so GitHub didn't auto-retarget and this PR was left pointing at an absorbed base. #11629's content is already in `main`, so the diff here now shows only this PR's own changes.

Picking it up in the merge queue on the next pass.

pirapira:
Holding this temporarily (removed `merge!`) — not a defect in this PR.

The per-reader balance differential added here is the agreement instrument that the account-reader migration (#11581) **retires**: the probe becomes an ABI-safe no-op in that change. Merging this and then deleting it in the next PR is churn, and this PR's commit is already the base of that migration's stack, so it will land with it if that is the shape chosen.

Re-labelling once the migration's net diff is settled and it is clear whether the harness survives in any live form. No action needed from the author.

pirapira:
superseded by #11638.

PR #11638, opened 2026-08-06
refactor(codegen): retire AccountState account readers

### LINKED ISSUE #11581: structural: retire the AccountState overlay in favour of AccountWriteMap (producer-first; blocked on publication timing)

pirapira:
Verification-base SHA: `ded385a1d` (rebased onto `main` at `17f4a153f`); final cleanup ELF SHA256: `772d11f648a387264c4083c6339d2d63fae946654b0f5a965e82f4d404efed96`.

## Summary

This PR supersedes #11636 and subsumes the original #11638 scope. It retires the
`AccountState` pending/durable account-reader overlay and makes `AccountWriteMap`
the source for current account values and attribution. The remaining balance,
authentication, code, absence, and storage-touch readers converge on the
transaction/block account-write maps, with the same-block code-path fix tracked in
[#11643](https://github.com/Verified-zkEVM/evm-asm/issues/11643).

Closes #11581.

## Why the map is the spec-shaped authority

The [execution-specs `state_tracker.py`](https://github.com/ethereum/execution-specs/blob/forks/amsterdam/src/ethereum/forks/amsterdam/state_tracker.py)
defines exactly two state containers, `BlockState` and `TransactionState`.
`get_account` resolves the current account from transaction `account_writes`, then
block `account_writes`, then `pre_state`; `get_code` has the corresponding write-map
order. Thus `account_writes` is both the attribution record and the source of current
values. The deleted pending/durable overlay has no execution-specs counterpart.

The guest-side port records this shape in
[`StateTracker.lean`](https://github.com/Verified-zkEVM/evm-asm/blob/9dcea0799/EvmAsm/Stateless/SpecRef/StateTracker.lean).

## Retired diagnostics

The cleanup commit removes the `account_agreement_probe` symbol, all four retired
`CreateCodeEffectLog` callsites, the legacy probe/scan/record functions, and the
map-vs-live/per-reader differential counters and event arena. The sweep remains as
the mutation-observation, verdict, and standing-control harness; it no longer parses
buckets that cannot fire after the reader retirement.

## Evidence

- Predicted post-cleanup `.text` size: `0x5566c`; measured: `0x5566c` (**prediction held**).
  The relinked pre-strip image was `0x55a68`, so the removed linked text was exactly
  `0x3fc` bytes. `.data` is `0x53b0`; `.bss` is `0x190695a0`.
- Against the pre-strip base ELF SHA256
  `036e468c0c967c42cc2bdf685fd9948a810dd32b0b333e33cefdf641db7b1c72`, the r200
  sweep covered 200 random cases plus two standing controls in both legs. Enabled:
  202 cases, 353 mutation observations, `flip_set=[]`, 0 output differences.
  Disabled: 202 cases, 0 mutation observations, `flip_set=[]`, 0 output differences.
  This reports the flip set and zero bar; it makes no roots-matching claim.
- A second artifact-generation pass was hash-identical for the relinked ELF-derived
  `symbol-addresses.tsv`, `GuestAddrs.lean`, `GuestImageEntries.lean`, and
  `RegionMapLinkPins.lean` (3046 symbol rows, 1021 symbols, 340 image entries / 129
  imports).
- `lake build` completed successfully (4820 jobs).
- `scripts/check-asm-to-program.sh` passed: 381 converted definitions across 144 files.
- `NO_BUILD=1 scripts/check-region-map.sh` passed, including ELF/TSV/pin agreement.

## Landing notes

The single cleanup commit is `9dcea0799` (`refactor(codegen): remove retired account
agreement diagnostics`). The branch is rebased directly onto `main` at `17f4a153f`.
CI is currently degraded and may not schedule a build; that is expected and does not
change the local verification above.


pirapira:

### The fix is right, and the comment it replaces shows why it was missed

The old text already identified one half of the problem: *"`create_frame_flag` alone is too broad: it describes a depth, while helper/refund activity at that depth can address another account"* — and narrowed it with an address-equality check. **That fixed over-breadth and left under-coverage untouched.** The flag is depth-local, so an account created in an *earlier* frame of the same transaction, then reached by a later CALL, never matches it.

That is the same pattern as several findings today: one direction of a condition checked carefully, the other never considered. The narrowing was correct and the remaining gap was invisible from inside it.

`state_tracker.py:324` settles the scope — `if address in tx_state.created_accounts: return U256(0)`, short-circuiting **before** `parent.storage_writes` or `pre_state`, for the whole transaction at any depth. So making `account_state_created_contains` the SSTORE original-zero authority is the spec's own shape, and reusing the existing `account_state_created` set rather than adding a second creation structure is what keeps it from becoming two sets with different scopes.

Register handling: `x1` and `x10`–`x13` preserved across the helper. `a0` aliases `x10`, the dispatch PC, so a clobber there redirects execution rather than corrupting a value — worth having been explicit about.

### The empty flip set is correctly reported as coverage, not proof

Predicted non-empty (original-zero drives SSTORE SET-vs-RESET gas and the refund); measured empty across 202 r200 cases, 8 targeted `create_init_fail` rows and 2 controls. **Reported as a coverage finding rather than a clean behavioural result**, which is the honest reading — and the body states the discriminating fixture concretely: one transaction CREATEs in a frame, RETURNs, then a later frame CALLs that account and SSTOREs a slot, so the depth-local flag is false while `created_accounts` is true, separable by SSTORE gas and refund.

That turns an absence into something buildable. I am filing it as a coverage issue so it does not live only here.

**So the justification is not "it fixes rows" but "the guest consulted the wrong thing and now consults the right one"** — which is the correct basis under the standing structural bar.

⚠️ Stack depth: this sits on #11636, which sits on #11629. Three deep, so the actor's sequencing matters; each was reviewed independently.

pirapira:
Reviewed against the pushed head `9dcea0799`, verifying from the artefacts rather than the author's report.

**Approved and labelling.**

## What I confirmed independently

| check | result |
|---|---|
| `account_agreement_probe` in emitted Programs | **0** — and `accountAgreementProbeLegacyFunction` is gone entirely |
| `account_agreement` rows in `symbol-addresses.tsv` | 25 on main → **5** here |
| the 5 survivors | `mutation_checkpoint`, `enabled`, `mutation_event_count`/`_overflow`/`_events` — the mutation observer, deliberately retained |
| `account_agreement` in `GuestAddrs.lean` | 0 (also 0 on main — those callsites were relocated `jal` targets, never numeric pins) |
| `account_writes_lookup_current` callsites | 28 |
| `.text` | `0x56a98` → `0x5566c` = **−5164 bytes** |
| size prediction | `0x5566c` named from linked spans *before* measuring — **held exactly** |
| behaviour | `flip_set` empty in both enabled and disabled legs, 202 cases, `output_differences=0` |
| regeneration | second pass hash-identical |

The spec argument is the reason this is worth its size: `state_tracker.py` defines exactly `BlockState` and `TransactionState`, neither holds a live-state object, and `get_account` → `get_account_optional` walks tx `account_writes` → block `account_writes` → `pre_state` (`get_code`'s docstring states the same chain outright). The overlay this retires had **no spec counterpart at all**, so every triple that related the two containers was carrying an invariant with no spec-side fact to discharge it against.

## Not blocking, but worth a follow-up

`CreateCodeEffectLog.lean:526`, `:572` and `:618` still **define** `accountStateLatestBalanceFunction`, `accountStateLatestNonceFunction` and `accountStateLookupCurrentFunction`. All three are referenced by nothing (the first two only by their own `#guard`s), so they are absent from the linked image and cost the guest nothing.

The hazard is that they are *intact, working-looking* readers of an overlay that no longer has a writer. Re-wiring one later would read zeros rather than fail — a silent wrong-value path, and the `#guard`s will keep them alive indefinitely. Filing this separately rather than holding the PR: it is source-only, it cannot affect a verdict as it stands, and the screening question (could this cause an invalid block to be **accepted**?) is no.

The base is now one merge stale after #11634 landed; per the layout rule, regeneration follows the *final* rebase, which the author cannot identify against a moving main. That is the merge actor's to absorb.

pirapira:
Parking this one — it fails two gates, and the first reproduces **standalone on your own branch**, independent of any batch.

**1. `check-file-size` (needs a structural decision from you).**

```
FAIL    1508 / 1500 lines  EvmAsm/Codegen/Programs/AccountWriteMap.lean  [codegen]
```

`main` has it at 1496 and this PR takes it to 1508. The gate's own guidance is to split rather than exempt ("Per-file file-size-exception markers are not supported"), suggesting extracting a cluster of defs into a sibling submodule — `ChildFrameHandlerTails.lean` is the cited pattern. Which cluster to extract is a design call I shouldn't make on your behalf.

**2. `check-guestaddrs-starts` / build (a batching artifact, not your bug — noted so it doesn't confuse you).**

```
DECLARED_MISSING entry=runtime_same_block_delegation_code — GuestAddrs has no def for linked converted symbol
Unknown identifier `GuestAddrs.rsbd_code_ptr` / `GuestAddrs.rsbd_code_len`
```

That is my batch resolving `GuestAddrs.lean` toward `main`, which drops the symbols this PR adds — and regeneration cannot recover them because it needs a successful build first. Your branch defines them correctly. I will resolve the generated files toward this PR instead when it comes back around, so **no action needed from you on that one**.

The sibling #11640 is unaffected and I am landing it separately. Once the file split is done, drop the `complicated-merge` label and it will be picked up on the next pass.

*(Edited: an earlier revision of this comment had its code snippets mangled by shell expansion.)*

PR #11640, opened 2026-08-06
proof(#11637): gate registry row existence, seeded with the 103-routine backlog

### LINKED ISSUE #11637: Registry blind spot: ~103 linked routines have proven specs and no row in either registry (the #11342 gap class, swept)
Follow-up to #11348, where @pirapira asked for this directly:

> `bloom_or_into` is **linked**, emitted into five images, and has a `sorry`-free SAsm spec — yet appears in **neither registry**, so it is invisible to the proven-count. […] **Proven work that no census can see is indistinguishable from work not done** […] The #11342 gap class deserves a sweep of its own, not just this instance.

I swept it. `bloom_or_into` was not an outlier.

## Headline

| | count |
|---|---|
| linked guest symbols (`GuestAddrs.lean`) | 1020 |
| rows across **both** registries (`Routines.lean` ∪ `Correspondence.registry`) | 27 |
| symbols with a **routine-level spec theorem** | 121 |
| ⭐ **linked + has spec + in NEITHER registry** | **103** |

So the registries see 27 of the 121 routines that have routine-level specs. The proven-count understates reality by roughly a factor of four on this measure.

### Tier A — flat whole-routine triples at the guest address, unregistered (≥20)

These are `cpsTripleWithin` triples whose base resolves to `GuestAddrs.<symbol>` — **exactly the kind of theorem the registered `.proven` rows cite.** There is no "it's only a structured spec" caveat: these could take `Routines.lean` rows today with no new proof work.

- `account_extract_balance                      account_extract_balance_spec_within`
- `account_extract_nonce                        account_extract_nonce_spec_within`
- `account_is_eip161_empty                      account_is_eip161_empty_spec_within`
- `blq_zero                                     blqZeroFlat_spec`
- `blsg_le_to_be                                blsgLeToBeFlat_spec`
- `bnf_be_to_le                                 bnfBeToLeFlat_spec`
- `bnf_le_to_be                                 bnfLeToBeFlat_spec`
- `bnq_zero                                     bnqZeroFlat_spec`
- `chain_validate_consecutive_numbers           chain_validate_consecutive_numbers_spec_within`
- `chain_validate_increasing_timestamps         chain_validate_increasing_timestamps_spec_within`
- `eip8037_tx_state_gas                         eip8037TxStateGas_spec_within`
- `header_validate_extra_data_length            header_validate_extra_data_length_spec_within`
- `mset_memcpy                                  mset_memcpy_spec_within`
- `receipt_extract_logs_bloom                   receiptExtractLogsBloom_spec_within`
- `secf_be_to_le                                secfBeToLeFlat_spec`
- `secf_is_zero32                               secfIsZero32Flat_spec`
- `secf_le_to_be                                secfLeToBeFlat_spec`
- `secf_zero32                                  secfZero32Flat_spec`
- `tx_type_dispatch                             txTypeDispatch_spec_within`
- `u256_sub_be                                  u256SubBeFlat_spec`

I kernel-checked 13 of these with `#print axioms`: **all classical-3, no `sorryAx`.** I also read two statements end-to-end (`account_extract_nonce_spec_within`, `txTypeDispatch_spec_within`) and confirmed the base abbrevs resolve to the guest addresses (`extractNonceBase = BitVec.ofNat 64 GuestAddrs.account_extract_nonce`; `D` → `GuestAddrs.tx_type_dispatch`).

⚠️ **20 is a lower bound.** My Tier-A test requires the literal string `GuestAddrs.<symbol>` in the same file, which has false negatives when a file names the address through a local abbrev. `account_decode` is the proof: `account_decode_spec_within` (merged via #11602) is a real flat triple, but it lands in Tier B here because `AccountDecodeClose6.lean` refers to the base indirectly.

### Tier B — structured SAsm specs only (~94)

Proven, but the spec is the SAsm `.Spec` rather than a flat triple, so a `.proven` row would overclaim until `Fn.retSpecFlat` is applied — the `bloom_or_into` situation exactly. Heavily concentrated in the precompile-limb families (`secf_*`, `bls*`, `bn*`, `p256_*`, `u256_*`), which is unsurprising: those were built as SAsm leaves.

<details><summary>full Tier B list</summary>

account_decode, bah_u32le, bgv_u64le, bhr_rev_le_be, blk2_ld_le64, blk2_st_le64, bloom_eq, blq_copy, blq_eq, blq_is_zero, blq_pt_copy, blq_zero, blsf_copy_quads, blsg2_encode, blsg2_eq_n, blsg2_zero192, blsg_be_to_le, blsg_copy96, blsg_eq48, blsg_is_zero_n, blsg_le_to_be, blsg_lt_p, blsg_zero96, blsk_g2_wire, blsk_lt_be, bn254_call_allotment, bnc_copy64, bnc_is_inf64, bnc_zero64, bnf_add_mod_p, bnf_be_to_le, bnf_eq32, bnf_is_zero32, bnf_le_to_be, bnf_lt_p, bnf_mul_mod_p, bnp_fp2_copy, bnp_fp2_eq, bnp_fp2_is_zero, bnp_fp2_zero, bnq_copy, bnq_eq, bnq_is_zero, bnq_pt_copy, bnq_zero, call_frame_set_calldata, chain_validate_blob_gas_used_multiple, chain_validate_blob_gas_used_under_max, chain_validate_extra_data_length, chain_validate_gas_used_under_limit, dispatcher_capture_exec_state_gas, enrg_u32le, eph_u32le, frame_base, frame_depth_pop, frame_depth_push, frame_load_regs, frame_save_regs, hp_decode_nibbles, hp_encode_nibbles, mpt_resolve_cache_reset, mset_memcpy, p256_be_to_le, p256_copy_n, p256_eq32, p256_is_zero_n, p256_le_to_be, p256_lt_be, secf_be_to_le, secf_copy32, secf_eq32, secf_get_bit_lsb, secf_is_zero32, secf_le_to_be, secf_mul_mod_p, secf_zero32, secp256k1_point_copy64, secp256k1_point_zero64, spw_u32le, ssz_pack_bytes, swd_minimal_copy, swd_read_u64le, swd_write_be32_u64, swd_write_be8, swr_rev_le_be, sws_u32le, u256_add_be, u256_div_u64_be, u256_eq, u256_from_u64_be, u256_is_zero, u256_lt_be, u256_min, u256_sub_be

</details>

## ⭐ The structural fix is a gate, not a sweep

Registering 103 rows by hand fixes today and rots tomorrow. The reason this class is invisible is that **both registries gate row *contents* and neither gates row *existence*** — `routineRegistry_all_witnessed`, `verdict_requires_spec`, `portDefect_cites_issue`, `crossVerdictOk` all quantify over rows that exist. A linked, proven symbol with no row trips nothing. #11342 found one instance; this is the same hole seen from the other side.

Proposed: a script gate (`scripts/check-registry-coverage.sh`) that recomputes the three sets above from source and fails when a linked symbol has a routine-level spec theorem and no row in either registry, with an **expiring allowlist** seeded with today's 103 — same pattern as the #11303 liveness gate (#11332). New gaps then fail immediately, and the backlog burns down visibly instead of silently.

I'd suggest **not** blanket-registering Tier B, because the honest tier for a structured-only spec is not `.proven`; that is a real decision per row, and I don't want to launder 94 of them through a sweep.

## Caveats, so the numbers aren't over-read

- The symbol↔theorem mapping is **name-based** (camel→snake on the theorem name, minus a `Fn_spec`/`Flat_spec`/`_spec_within`/`_spec` suffix). A theorem whose name coincides with a symbol but proves something narrower would be a false positive. I verified a sample, not all 103.
- "Has a routine-level spec" ≠ "correspondence-audited". These rows would say the routine is *proven*, not that its spec matches execution-specs — that remains the separate `Correspondence` question, and for most of Tier B nobody has asked it yet.
- 9 registered rows had no spec theorem found by this pattern (`rlp_encode_uint_be`, `bal_canonical_sort`, `rlp_item_span`, …). Mostly naming, or specs derived elsewhere; `rlp_item_span` is the known `.unproven` row. Not a defect, just the inverse noise in the same heuristic.

Happy to implement the gate plus the Tier A rows as one PR, or split them. Sweep script: `$CLAUDE_JOB_DIR/tmp/sweep2.py` logic reproduced above; I'll land it as the gate rather than leave it in a scratch dir.

dhsorens:
Closes #11637. ⚠️ **Stacked on #11631** — base is that branch, so the diff is just this gate. Must land after it (the allowlist is only correct with #11631's `bloom_or_into` row present).

## The hole

Both registries already gate row **contents**:

| gate | what it checks |
|---|---|
| `gen-axiom-witnesses.py` | a row must name a theorem with a witness abbrev |
| `verdict_requires_spec` | a verdict needs a spec |
| `crossVerdictOk` | a witnessed routine must not be `.unproven` |
| `routineRegistry_all_witnessed` | every row names a witness |

⭐ **Every one of those quantifies over rows that exist.** Nothing gated row *existence*. So a routine could be linked into the guest, carry a `sorry`-free whole-routine triple, and appear in neither registry — counting toward nothing.

#11342 found one instance. #11348 found another (`bloom_or_into`: linked, emitted into five images, `sorry`-free SAsm spec, zero rows — which is why that issue *looked* unstarted when it was ~90% done). The sweep found **103**.

## The gate

`scripts/check-registry-coverage.py` recomputes three sets from source on every run — linked symbols, registered routines (`Routines` ∪ `Correspondence`), routine-level spec theorems — and fails when a linked, spec-bearing symbol has no row and no allowlist entry. Wired into `report_checks` next to `check-progress`/`check-drift`: pure source scan, no build, instant.

**The allowlist expires**, which is the whole ratchet (`routine-liveness-allow.txt`'s shape, #11303/#11332). An entry goes STALE — failing the run and naming the line to delete — once the symbol gains a row, loses its spec, or leaves the image. The backlog burns down visibly, and a *new* gap fails immediately instead of joining a pile nobody reads.

## Both failure modes tested, not assumed

A gate nobody has seen fail is indistinguishable from one that cannot:

```
$ # dropped tx_type_dispatch from the allowlist
FAIL — 1 linked, spec-bearing routine(s) have NO row ... and no allowlist entry:
    [A] tx_type_dispatch	txTypeDispatch_spec_within	.../TxTypeDispatchTop.lean

$ # allowlisted the already-registered rlp_list_count_items
FAIL — 1 STALE allowlist entr(ies):
    rlp_list_count_items	now registered -- delete this line
```

## Seeded at 103 = 20 tier A + 83 tier B

The tiers exist because the remedies differ:

- **Tier A (20)** — a flat `cpsTripleWithin` at `GuestAddrs.<sym>`. Registrable as `.proven` **today**, no new proof work. Clear these first; each deletion is a row gained. Includes `tx_type_dispatch`, `account_extract_nonce`/`_balance`, `account_is_eip161_empty`, two `chain_validate_*`, `receipt_extract_logs_bloom`.
- **Tier B (83)** — a structured SAsm `.Spec` only; `Fn.retSpecFlat` must derive the flat triple before a `.proven` row would be honest.

⚠️ The allowlist header says **explicitly not** to grade tier B `.proven` to shorten the file. Registering a structured-only spec as a flat triple is precisely the invisible overclaim this gate exists to prevent, and would convert a visible backlog into a false census. I'd rather ship a gate with 83 honest exemptions than 83 fake rows.

## Stated limit

The symbol↔theorem map is **name-based** (camel→snake minus the spec suffix). That is what makes it cheap enough to run every time, and the cost is that it can **under-report** a gap but never invent one — the right failure direction for a coverage gate, but it makes this a *floor* on the backlog, not a census. `account_decode` is the worked example: it sits in tier B only because its file names the base through a local abbrev, and its allowlist reason says exactly that.

`EvmAsm/Progress/**` is excluded from the scan so the registries' own witness abbrevs and docstrings don't count as specs.

## Verification

- `python3 scripts/check-registry-coverage.py` → OK, exit 0 (1020 linked, 27 registered, 121 spec-bearing, 103 uncovered, 103 allowlisted)
- both failure modes reproduced above, allowlist restored after
- 13 of the tier-A theorems kernel-checked with `#print axioms` during the sweep: all classical-3, no `sorryAx`


pirapira:
**coord review: `merge!` applied.** Scripts only (`check-registry-coverage.py`, its allowlist, one line in `check-build-parallel.sh`) — no `EvmAsm/` change, so nothing reaches the image.

### The hole is real and the framing is exactly right

Every existing registry gate **quantifies over rows that exist** — `gen-axiom-witnesses.py`, `verdict_requires_spec`, `crossVerdictOk`, `routineRegistry_all_witnessed`. None gated row *existence*. So a routine could be linked into the guest, carry a `sorry`-free whole-routine triple, and **count toward nothing**.

That is not hypothetical: #11342 found one instance, and #11348 found `bloom_or_into` — linked, emitted into five images, `sorry`-free SAsm spec, zero rows — **which is why that issue looked unstarted when it was ~90% done.** A census that cannot see proven work is worse than no census, because it makes the proven-count look like a measurement while understating reality by an unknown margin. Closing that hole is worth more than any single row.

### Both failure modes tested rather than assumed

> A gate nobody has seen fail is indistinguishable from one that cannot.

Dropping `tx_type_dispatch` from the allowlist fails with the missing-row message; allowlisting the already-registered `rlp_list_count_items` fails **STALE** and names the line to delete. Two-sided, demonstrated. That is the same discipline as #11607's `--self-test`, and it is what I have been asking for on every gate tonight.

### ⭐ The 83 honest exemptions are the right call

> I'd rather ship a gate with 83 honest exemptions than 83 fake rows.

Agreed, and emphatically. Tier B carries a **structured SAsm spec only**; grading it `.proven` without `Fn.retSpecFlat` deriving the flat triple would register a claim the artefacts do not support — converting a *visible backlog* into a *false census*, which is precisely the failure this gate exists to catch. An allowlist that expires (entries go STALE once a symbol gains a row, loses its spec, or leaves the image) turns 103 into a burn-down rather than a pile nobody reads.

Tier A's 20 are the actionable half — flat `cpsTripleWithin` at `GuestAddrs.<sym>`, registrable today with no new proof work, each deletion a row gained.

⚠️ Stacked on #11631, correctly, since the allowlist is only accurate with `bloom_or_into`'s row present. That puts it behind a queue that is currently compute-bound — CI is degraded and #11634's `source-checks` died at "Set up job" — so expect a wait rather than a problem.

dhsorens:
Rebased onto `main` now that #11631 has merged — and ⭐ **the gate fired for real in the process**, which is better evidence than the synthetic tests in the PR description.

While this PR sat waiting on the Actions outage, #11575 landed a `Routines.lean` row for `header_validate_extra_data_length` — one of this allowlist's **tier A** entries. Merging `main` in, the gate said:

```
check-registry-coverage: FAIL — 1 STALE allowlist entr(ies):
    header_validate_extra_data_length	now registered -- delete this line
```

So the expiring half of the ratchet works on a live merge, not just on my hand-made test: the exemption's reason expired, and the run failed naming the exact line to delete rather than letting a stale exemption sit there indefinitely. Deleted it, and the counts move in lockstep:

| | before | after |
|---|---|---|
| registered | 27 | **28** |
| uncovered | 103 | **102** |
| tier A | 20 | **19** |

I recorded this in the allowlist header, because *"the gate caught a burn-down on day one"* is the evidence that it discriminates — a ratchet nobody has seen move is indistinguishable from a comment.

It also happens to demonstrate the collision I warned about on #11642: two people touching registry state at once. Here it resolved cleanly because the gate made the interaction **loud** instead of silent, which is the point.

Still `BLOCKED` on CI — Actions has been in `major_outage` for a few hours. Nothing further from me until it recovers.

PR #11642, opened 2026-08-06
proof(#11422): state the hpDecode → compact_to_nibbles link the header says is owed

### LINKED ISSUE #11422: proof: compact_to_nibbles' flag decode is unproven — the odd-nibble prepend, is_leaf, and empty rejection
Split out of #11344 at review (pirapira on #11421), so the gap is recorded rather than assumed covered.

## What #11344 covers, and what it does not

#11344 proved `bytes_to_nibbles` against `SpecRef.keyToNibbles`, and noted it "also covers the nibble-expansion half of `compact_to_nibbles` — that half only, **not the flag decode**". That scoping is accurate: the expansion half is *literally* `keyToNibbles rest`, so no separate lemma was needed.

But it means a reader seeing `compact_to_nibbles` named in the `mpt` Correspondence row could reasonably think the whole function is covered. It is not.

## The uncovered part

`compact_to_nibbles` (`EvmAsm/Stateless/SpecRef/IncrementalMpt.lean:76-84`):

```lean
let first_nibble := c0.toNat >>> 4
let odd  := first_nibble &&& 0x01 ≠ 0
let tail := keyToNibbles rest
let nibbles := if odd then BitVec.ofNat 8 (c0.toNat &&& 0x0F) :: tail else tail
```

⭐ **`bytes_to_nibbles` cannot produce an odd nibble count** — its output length is `2 * len` by construction, pinned in four places in the triple (`horig`, `hdstOver`, the returned `x10 = 2 * len`, and the written region being a `flatMap` emitting exactly two nibbles per byte). An odd result is unrepresentable there.

**Odd-length paths in the MPT come entirely from the hex-prefix flag byte**: when the odd flag is set, `c0`'s *low* nibble is prepended to the even expansion, giving `1 + 2*(n-1)`. The expansion is always even; the parity comes from the flag.

## Three obligations, none currently proven

1. **The odd-flag prepend** — `(c0 &&& 0x0F) :: tail`, the only source of odd-length paths. If the guest's compact-decode drops that nibble, or takes the **high** nibble instead of the low, #11344's triple would **not** catch it: the proof stops at the even expansion.
2. **The `is_leaf` flag** — `first_nibble &&& 0x02`. Note `mpt_node_kind` (#11347) reads the high nibble directly (`hn < 2` / `hn < 4`) where the reference gets leaf-vs-extension from this flag; equivalent, but nobody has written the lemma.
3. **Empty-input rejection** — Python indexes `compact[0]` unconditionally, so `[]` is an `IndexError`; the port throws `witnessNodeMalformed`. Whether the guest rejects `[]` the same way is unchecked.

## Suggested shape

A bridge in the #11344 style: prove the guest's compact-decode against `compact_to_nibbles` in full, citing `bytesToNibblesBytes_eq_keyToNibbles` (merged in #11421) for the expansion half so only the flag logic is new work. Then a Correspondence row graded per the `ported` rung (#11401) with its port-fidelity clause table — the `is_leaf` equivalence in (2) is exactly the kind of non-syntactic restatement that rung requires to be proved rather than assumed.

Related: #11344 (merged, the expansion half), #11347 (`mpt_node_kind`, which reads the same flag byte differently), #11401 (the `ported` rung).

dhsorens:
Closes #11422.

## ⭐ The issue is much smaller than its title, and that's the finding

#11422 lists three obligations as unproven. The **machine side of all three already exists** in `HpDecodeNibblesSAsmPaths.lean`:

| #11422 obligation | already proven as |
|---|---|
| 1. the odd-flag prepend | `pathD_spec` — *"odd parity — head nibble stored"*, with `hdnC0_odd` / `hdnRes_some_odd` fixing the parity decode |
| 2. the `is_leaf` flag | the `isl` cell is written `BitVec.ofNat 64 (b0.toNat / 16 / 2 % 2)` — bit 1 of the head nibble |
| 3. empty-input rejection | `pathA_spec` — *"len = 0 — the first guard branches straight to the fail tail; nothing is written"* |

And `hdnRes_eq_hpDecode` already ties the guest-exact mirror to `hpDecode` definitionally. So `guest → hdnRes → hpDecode` was complete, and the **one** missing hop was the last one — to the SpecRef port. `HpDecodeNibblesSAsm.lean`'s own header says so:

> the further link `hpDecode -> SpecRef.compact_to_nibbles` is **still not stated as a theorem** […] the behaviour agrees, the proof does not exist yet.

This adds that theorem. It's the third time in this batch the pattern has been *both halves exist, the join doesn't* (after #11348 and the `hpUnpackPairs`/`keyToNibbles` join below).

## What's proven

- **`hpUnpackPairs_eq_keyToNibbles`** — the tails agree. Pure composition: `nibblePrefix_eq_hpUnpackPairs` and `nibblePrefix_eq_keyToNibbles_take` are *both* phrased `nibblePrefix bs i = _ (bs.take i)`, so instantiating each at `i = bs.length` and collapsing `bs.take bs.length` joins them. Neither lemma mentions the other's right-hand side, which is exactly why the join had to be written down.
- **`compact_to_nibbles_eq_hpDecode`** — the reference port **is** the guest mirror, up to two representation differences and nothing else: the tuple order (`(isLeaf, nibbles)` vs `(nibbles, isLeaf)`) and the failure encoding (`none` vs `throw .witnessNodeMalformed`). **Total** — no hypothesis on the input, so obligation 3 is covered at the model level too.

The flag agreement is settled by a byte-exhaustive `decide` (256 cases, ~1.8s — the kernel's `Nat` is GMP-backed) rather than a hand-rolled bit argument.

## ⚠️ Why it can be total

Recorded because anyone chasing #11422 will expect a divergence here: `hpDecode` **used to** reject a head nibble `≥ 4` where `compact_to_nibbles` masks bits 2–3 away and accepts — the guest was `stricter`, a false-reject shape. **#10528 closed that** by making both sides mask. So this theorem's totality is a consequence of that fix landing first; before #10528 it would have been **false**. The module header says so, so the next reader doesn't have to rediscover it.

## Scope / follow-up, deliberately not bundled

**No Correspondence row in this PR.** Adding one would register `hp_decode_nibbles`, which is a tier-B entry in #11640's registry-coverage allowlist — so the row and that allowlist deletion must land together, *after* #11640. That's precisely the census collision I flagged on that PR, and it would be poor form to create it here. The row is the immediate follow-up.

## Verification

- `lake build EvmAsm.Codegen.Programs.HpDecodeCompactBridge` — green; registered in `Programs/Imports.lean`
- `#print axioms` on both new theorems — `[propext, Classical.choice, Quot.sound]`, no `sorryAx`
- `check-forbidden-tactics.sh` — OK
- module elaborates in **3.9s** (measured with a forced rebuild; the cold-cache figure read 371s and was misleading)

### DIFF
--- a/EvmAsm/Codegen/Programs/HpDecodeCompactBridge.lean
+++ b/EvmAsm/Codegen/Programs/HpDecodeCompactBridge.lean
@@ -1,0 +1,108 @@
+/-
+  EvmAsm.Codegen.Programs.HpDecodeCompactBridge
+
+  GH #11422 — the theorem `HpDecodeNibblesSAsm.lean`'s header says is owed:
+
+      "the further link `hpDecode -> SpecRef.compact_to_nibbles` is still not
+       stated as a theorem [...] the behaviour agrees, the proof does not exist yet."
+
+  ⭐ WHAT WAS ACTUALLY MISSING, because it is much less than the issue's title
+  suggests. #11422 lists three obligations, and the machine side of all three is
+  already proven in `HpDecodeNibblesSAsmPaths.lean`:
+
+    1. the odd-flag prepend  -> `pathD_spec` ("odd parity -- head nibble stored"),
+       with `hdnC0_odd` / `hdnRes_some_odd` fixing the parity decode;
+    2. the `is_leaf` flag    -> the `isl` cell is written
+       `BitVec.ofNat 64 (b0.toNat / 16 / 2 % 2)`, i.e. bit 1 of the head nibble;
+    3. empty-input rejection -> `pathA_spec` ("len = 0 -- the first guard branches
+       straight to the fail tail; nothing is written").
+
+  And `hdnRes_eq_hpDecode` already ties the guest-exact mirror to `hpDecode`
+  definitionally. So the chain `guest -> hdnRes -> hpDecode` is complete, and the
+  ONE missing link is the last hop to the SpecRef port. That is what this file adds.
+
+  ⚠️ THE DIVERGENCE THIS FILE DOES **NOT** HAVE TO HANDLE, recorded because a reader
+  chasing #11422 will expect it: `hpDecode` used to reject a head nibble `>= 4`
+  where `compact_to_nibbles` masks bits 2-3 away and accepts — the guest was
+  STRICTER, a false-reject shape. GH #10528 closed that by making both sides mask.
+  So the bridge below is a total agreement with no side condition, which is only
+  true because that fix landed first; before #10528 this theorem was FALSE.
+
+  METHOD. Two independent facts, neither interesting on its own:
+
+    * the tails agree — `hpUnpackPairs = keyToNibbles`, which is pure composition:
+      `nibblePrefix_eq_hpUnpackPairs` and `nibblePrefix_eq_keyToNibbles_take` are
+      both stated as `nibblePrefix bs i = _ (bs.take i)`, so instantiating each at
+      `i = bs.length` and collapsing `bs.take bs.length` joins them. Both halves
+      already existed; only the join was absent.
+    * the flags agree — `compact_to_nibbles` tests `first_nibble &&& 0x02` and
+      `&&& 0x01` while `hpDecode` matches on `(b0.toNat / 16) % 4`. Both read the
+      same two bits, and the byte-exhaustive `decide` below is what says so rather
+      than a hand-rolled bit argument (the kernel's `Nat` is GMP-backed, so 256
+      cases is cheap — see CLAUDE.md).
+-/
+
+import EvmAsm.Codegen.Programs.HpDecodeNibblesSAsm
+import EvmAsm.Codegen.Programs.BytesToNibblesBridge
+import EvmAsm.Stateless.SpecRef.IncrementalMpt
+
+namespace EvmAsm.Codegen.HpDecodeNibblesSAsm
+
+open EvmAsm.Codegen.BytesToNibblesSAsm (nibblePrefix)
+
+/-- ⭐ **The tails agree.** The guest mirror's pair-unpacking is the reference's
+    `keyToNibbles`, obtained by joining two existing lemmas at `i = bs.length`.
+
+    Both are phrased `nibblePrefix bs i = _ (bs.take i)`; neither mentions the
+    other's right-hand side, which is why this join had to be written down. -/
+theorem hpUnpackPairs_eq_keyToNibbles (bs : List (BitVec 8)) :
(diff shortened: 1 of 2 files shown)


pirapira:
**coord review: `merge!` applied.** Two files, `+109/−0`, no emitted change — a bridge plus its `Imports` line.

### The finding is that the issue was mis-sized, and that generalises

All three of #11422's obligations already had their machine side in `HpDecodeNibblesSAsmPaths.lean` — the odd-flag prepend as `pathD_spec`, the `is_leaf` flag in the `isl` cell, empty-input rejection as `pathA_spec` — and `hdnRes_eq_hpDecode` already tied the guest-exact mirror to `hpDecode`. **The only missing hop was the last one, to the SpecRef port.**

You note this is the third occurrence in one batch of *both halves exist, the join doesn't* (after #11348 and the `hpUnpackPairs`/`keyToNibbles` join here). ⭐ **Three instances make it a property of how this work is tracked, not a coincidence** — an issue reads as unstarted while being nearly complete, because nothing gates the *existence* of the join. That is the same blind spot #11640 closes for registry rows, one level up: there, a proven routine with no row counted toward nothing; here, two proven halves with no join read as no proof at all. Worth saying on #11637 that the pattern has a second form.

And the reason the join had to be written is precise rather than incidental: `nibblePrefix_eq_hpUnpackPairs` and `nibblePrefix_eq_keyToNibbles_take` are both phrased `nibblePrefix bs i = _ (bs.take i)`, so **neither mentions the other's right-hand side**. Nothing in the ambient context connects them; the composition is only obvious once someone instantiates both at `i = bs.length`.

### Totality is a consequence of an earlier fix, and recording that is the valuable part

`compact_to_nibbles_eq_hpDecode` is **total** — no hypothesis on the input — which is surprising enough that a reader chasing #11422 would expect a divergence. The reason is that `hpDecode` **used to** reject a head nibble ≥ 4 where `compact_to_nibbles` masks bits 2–3 and accepts: the guest was stricter, a false-reject shape, and **#10528 closed it by making both sides mask.** So this theorem would have been **false** before that landed. Putting it in the module header rather than only the PR body is right — a future reader who reverts or narrows #10528 needs to know this theorem depends on it.

Flag agreement settled by byte-exhaustive `decide` over 256 cases rather than a hand-rolled bit argument: cheaper to trust, and `decide` is the sanctioned tool here.

### Scope restraint

**No `Correspondence` row**, deliberately — adding one would register `hp_decode_nibbles`, a tier-B entry in #11640's allowlist, so the row and that allowlist deletion must land together and after it. That is exactly the census collision I flagged on #11640, and declining to create it here is the right call. Axioms classical-only on both theorems.

PR #11647, opened 2026-08-06
fix(codegen): skip no-op SW rows in map storage apply (#11614 B)

pirapira:
## Summary

Layer B of #11614: stop `BalAccountApplyPostFields` `.Lbaap_map_storage` from emitting MPT `mode=2` DELETE descriptors for no-op `storage_writes` rows (val == baseline at +96).

Spec: `execution-specs` `block_access_lists.py:549-552` — slots READ but not CHANGED go to `storage_reads` only, never `storage_changes`.

Family: RECORDING PRESENCE WHERE THE SPEC HAS NONE (delete form).

Layer A (idempotent delete-absent in `mpt_delete_acc`) is **not** in this PR.

## Fix

At `.Lbaap_map_row_hit`, before slot/value copy:

```
ld +64 vs +96 (4 dwords); equal → .Lbaap_map_next
```

Same discipline as `BlockVerdictStateRoot` `.Lem_storage_loop` / `bal_emit_storage_changes` (established by #11604 for another consumer).

## Rebase / re-measure vs main `72664a3f4` (post-#11633)

Prior numbers were on `17f4a153f` **without** #11633. Re-measured after rebase **before** pin regen.

### Production path (bounded-first, as shipped)

| id | base | cand | delta |
|---|---|---|---|
| 21353 / 21354 | PASS | PASS | — |
| 23274 / 78 / 80 | PASS | PASS | — |
| **12832 / 33 / 34** | **PASS** | PASS | was FR bv1 pre-#11633; **absorbed by #11633** |
| 01114 / 00178 / 00026 / 05814 / 11710 | PASS | PASS | — |
| r200 seed=20260806 | bv1=0 | bv1=0 | **F→P=0 P→F=0** |

Production named flip set is empty — #11633 already greened 12832; 21353/23274 were already PASS via bounded.

### Always-legacy diagnostic (proves B still buys something)

Force `j .Lbaap_multi_apply_legacy` (not in PR). Base = main tip without B; cand = B.

| id | base | cand |
|---|---|---|
| **21353 / 21354** | FR **bv3** | **F→P PASS** |
| **23274 / 78 / 80** | FR **bv3** | **F→P PASS** |
| 12832 / 33 / 34 | PASS | PASS |
| 01114 / 00178 / 00026 | PASS | PASS |

**Why land despite empty production flip:** #11633 fail-closes LCP-prefixed collapse and **routes more traffic to legacy**. Without B, that legacy path still receives absent-key DELETEs for no-op SW rows → `mpt_delete_acc` not-found → baap_fail=501 → bv_fail=3. B removes the bad descriptor set so legacy is not fed garbage. Layer A alone would hide the bad input; B fixes the producer.

## Layout

| | value |
|---|---|
| **predicted** textSizeBytes | `0x56b04` (main `0x56ad4` + `0x30` prior B delta) |
| **measured** textSizeBytes | `0x56b04` (match) |
| pin fixpoint | second regen pass hash-identical (pass 4) |
| GuestAddrs check | ok=340 fail=0 |

## ELFs

- mainbase sha256 `a857fd8086c9bd45…` @ `72664a3f4`
- cand sha256 `86f3dcba1921e1fb…` @ this PR

## Files

- `EvmAsm/Codegen/Programs/BalAccountApplyPostFields.lean` (fix)
- `RegionMapLinkPins.lean` / `GuestAddrs.lean` / `symbol-addresses.tsv` (pins)

## Test plan

- [x] Named production table above
- [x] Always-legacy diagnostic F→P on 21353/23274 family
- [x] r200 seed=20260806 0/0
- [x] pin second-pass identical; check-guestaddrs-starts clean
- [ ] CI build / source-checks

Do not self-label.


pirapira:
**coord review: `merge!` applied.** Verified the spec basis and the measurements independently.

## The change is spec-aligned, and the citation should be strengthened

The comment cites `block_access_lists.py:549-552`. That line is the **read-exclusion** rule (`if slot not in changes.storage_changes: storage_reads.append(slot)`) — it explains why a no-op slot surfaces as a *read*, but it is the consequence, not the gate.

The gate is **`:666-676`**, and it is a much closer match to what this PR does:

```python
for address, slots in tx_state.storage_writes.items():
    for key, post_value in slots.items():
        pre_value = _get_pre_tx_storage(block_state.storage_writes, pre_state, address, key)
        if pre_value != post_value:
            add_storage_write(builder, address, u256_slot, idx, post_value)
```

A write becomes a `storage_change` **only if `pre_value != post_value`**. The four-dword `+64` vs `+96` compare is precisely that conditional in assembly. Worth citing `:666-676` as the authority and `:549-552` as the downstream effect.

This is also the right *shape* of fix: it makes the deficient copy reach the working one (`.Lem_storage_loop`), rather than introducing a second mechanism.

## What the residual risk is, and it is not the gate logic

`_get_pre_tx_storage(block_state.storage_writes, pre_state, ...)` means the spec's baseline is **block-cumulative, then pre-state** — not the tx-local pre value. So the correctness of this gate rests entirely on `+96` carrying that same baseline. If `+96` were ever the tx-local pre, a slot changed by an earlier tx in the same block could compare equal and have its change suppressed. Suppressing a required change is the FA-ward direction here, so `+96`'s provenance is the thing to keep an eye on — not the comparison, which is correct at 32 bytes.

## Measurements, stated honestly by the author

- Predicted `textSizeBytes 0x56b04` **before** regenerating; measured `0x56b04`. Pins second-pass hash-identical.
- r200 seed 20260806: **0 flips both directions**.
- **Production named flip set is EMPTY**, and the author said so plainly rather than reporting the pre-rebase table.

The demonstrated `F→P` is on the always-legacy diagnostic path (21353/21354/23274/78/80). Those rows still route to bounded in production, and the rows #11633 pushed to legacy (12832/33/34) pass on legacy either way — so **no production row is shown to benefit today**. Landing anyway on three grounds: it removes a *false reject* (`bv3`), it aligns a divergent copy with both the spec conditional and the working in-tree comparator, and it shows zero measured deterioration. Legacy traffic genuinely increased after #11633 — that fix turned FR rows into PASS, which a fail-closed reject could not do, so its fail path demonstrably falls back to legacy.

pirapira:
## Record corrections from coord review (no code change)

**Stronger spec cite:** the authority for the four-dword gate is `block_access_lists.py:666-676` — `_get_pre_tx_storage` then `add_storage_write` only if `pre_value != post_value`. `:549-552` is the read-exclusion consequence (why a no-op surfaces as a READ), not the gate itself. Will use `:666-676` going forward.

**Residual risk (carry forward, not fix now):** `_get_pre_tx_storage` is block-cumulative then pre-state. Gate is correct only if SW `+96` carries that same baseline. If `+96` were ever tx-local, an earlier-tx change in the same block could compare equal and suppress a required change (FA-ward). Follow-up if any: establish `+96` provenance vs `_get_pre_tx_storage`, not touch the compare.

Hands off merge! — no rebase / pin regen / empty-commit from me.

pirapira:
## +96 provenance audit (coord order) — CLOSED BY CONSTRUCTION

Read-only. No code change on this PR.

### Spec
`_get_pre_tx_storage(pre_tx_storage=block_state.storage_writes, pre_state, addr, key)`:
1. key in block cumulative → that value
2. else `pre_state.get_storage` (0 if unset)

BAL authority (`block_access_lists.py:666-676`): emit only if that pre ≠ post. Runs before merge.

### Layout (`storage_writes` 128B row)
`+0` addr, `+32` slot, `+64` **current**, `+96` **baseline** (not the transient-log layout).

### Writers of `+96`
| Map | Writer | When `+96` set | Source |
|---|---|---|---|
| TX | `storage_write_record` | APPEND only; HIT leaves | `a6` = SSTORE original resolver (`Storage.lean:646-650`): TX original → block `+64` → parent header → 0 (= `_get_pre_tx_storage`) |
| Block | `storage_writes_block_upsert` | APPEND only; HIT leaves (#11604) | `write_sets_incorporate_tx` passes `tx_entry+96` (first-write frozen); system seed passes authenticated parent original |

### Consumers — two different gates
1. **`bal_emit_storage_changes`**: does **not** read row `+96`. Re-derives baseline (block container `+64`, else parent header). Independent of this PR.
2. **This PR / Lem_storage / baap_map**: block `+64` vs `+96` = **MPT net-over-block** filter. Block `+96` = first writer's pre-tx = **pre-block** (HIT freezes).

### FA scenario (last-tx pre-tx) REFUTED for this consumer
`pre=5, TX0→7, TX1→5` → block ends `+64=5 +96=5` → skip MPT (correct net-zero). BAL still emits per-tx via bal_emit's own baseline, not `+96`.

This PR implements the MPT pre-block gate, not the BAL per-tx gate. Same field; different consumer. Spec `:666-676` is BAL authority; this gate is the analogous net-over-block filter for map→MPT.

### Verdict
No defect found for the stated FA. No code change. Optional residual only: SystemStorageCapture scratch naming (`+64=orig +96=cur`) is opposite the map row after upsert — already correct by arg order.

PR #11656, opened 2026-08-06
refactor(codegen): delete retired account state readers

### LINKED ISSUE #11646: Delete three unreferenced AccountState reader bodies left after the overlay retirement
After #11638 retires the AccountState overlay, three legacy reader bodies remain **defined but unreferenced** in Lean source:

- `EvmAsm/Codegen/Programs/CreateCodeEffectLog.lean:526` — `accountStateLatestBalanceFunction`
- `:572` — `accountStateLatestNonceFunction`
- `:618` — `accountStateLookupCurrentFunction`

Verified at #11638's head `9dcea0799`: nothing references them except three `#guard`s each on the first two (`:1132-1137`), asserting the internal `account_state_latest_balance:` / `_block:` / `_core:` labels. `accountStateLookupCurrentFunction` has **no** reference at all, not even a guard. They are in no emitted bundle, so `nm` on the linked image does not see them and they cost the guest zero bytes.

## Why they should still go

These are **intact, working-looking readers of a container that no longer has a writer.** The overlay arrays survive as storage; what #11638 removed is everything that *populates* them. So a future author who re-wires one of these — reasonably, since the code reads correct and the guards imply it is maintained — gets **zeros rather than a failure**. That is the correct-but-unfed shape, and it fails open: a zero balance or nonce reads as a legitimate value, not as a miss.

The `#guard`s make this worse rather than better. They pin the dead bodies' internal structure, so the code cannot rot visibly and any deletion attempt looks like it is breaking a checked invariant.

Compare the treatment #11638 gave the analogous case: `accountAgreementProbeLegacyFunction` was left as an unreferenced body in an earlier commit, then **deleted outright** in the cleanup commit `9dcea0799` rather than documented around. These three are the same category and deserve the same fix.

## Scope

Delete the three `def`s and their `#guard`s. Source-only, no emitted bytes, so:

- `.text` must be **unchanged** — this is the check that proves they really were dead. A nonzero delta means one was reachable and the premise here is wrong.
- No pin regeneration should be owed. If `symbol-addresses.tsv` or `GuestAddrs.lean` move, stop and report it.
- `CallFrameRoundtrip.lean:105` has an `account_state_latest_balance: ret` stub; that is a test fixture, leave it.

Blocked until #11638 lands (labelled `merge!` as of this filing).

pirapira:
Closes #11646.

Scope is exactly the three unreferenced reader defs in `CreateCodeEffectLog.lean` and their six guards:
- `accountStateLatestBalanceFunction`
- `accountStateLatestNonceFunction`
- `accountStateLookupCurrentFunction`

The live `account_state_pending` arena and `BlockVerdictMtxRuntime.lean` durable-count initialization are untouched.

Evidence against current base `b48148196`:
- Predicted `.text`: `0x557a0` (current `RegionMapLinkPins` pin).
- Base and post-edit `.text`: `0x557a0`.
- Base and post-edit extracted `.text` SHA256: `4e826d5d7b9f7e9f24e9035f5b9e328a99dff35acb9b1958b4658019426a1c0a`.
- `cmp` of extracted `.text` passed.
- Base and post-edit full ELF SHA256: `eb7eb0307719d33083d8116149e1c656b83cb1daa5dfcfb44b71a677f806f894`.
- Passed: `lake build` (4824 jobs), `scripts/check-file-size.sh`, `scripts/check-unimported.sh`, `scripts/check-asm-to-program.sh`, and `scripts/check-region-map.sh`.

No generated pins were changed.


pirapira:
**coord review: `merge!` applied.** Verified against head `45002638d` from the artefacts, not the author's report.

## Scope is exactly right

| | |
|---|---|
| files | **1** — `CreateCodeEffectLog.lean` |
| diff | **+0 / −125** |
| deleted | `accountStateLatestBalanceFunction` (`:526`), `accountStateLatestNonceFunction` (`:572`), `accountStateLookupCurrentFunction` (`:618`) + their 6 `#guard`s |
| emitted labels removed | the 7 reader labels (`account_state_latest_balance`/`_block`/`_core`, `_nonce`/`_block`/`_core`, `lookup_current`) |
| generated files touched | **none** |

No pin regeneration appears, which is the correct shape: these bodies were in no emitted bundle, so a pure source deletion owes no regen. `RegionMapLinkPins.textSizeBytes` reads `0x557a0` — inherited from `main`, not written by this PR.

## The two protected structures were left alone

This was the trap, and it is avoided:

- **`account_state_pending` survives.** `NonstorageEffectLog.lean:683-684` defines `account_state_created` and `account_state_delete` as `.set` offsets **into** that arena, and the created set is live (the map's tombstone predicate and `NoopHalt` read it). Deleting or shrinking the arena would silently move both aliases. Untouched.
- **`BlockVerdictMtxRuntime:281` survives.** It zeroes `account_state_durable_count` in the same sequence as the live `account_state_created_count`, `account_state_delete_count` and `account_state_overflow`. Untouched.

The 6 mentions of those symbols inside the deleted lines are *within* the removed reader bodies — those readers scanned pending/durable by design — not declarations.

## Evidence

Author ran a **before/after differential on a fixed base** and the extracted `.text` SHA came out identical (`27dc2ee26d3f`), which is the strong form of "this code was dead" — stronger than comparing size numbers, since it compares emitted bytes directly. Both sides of that comparison sat on the same base, so it is valid regardless of how far the branch was behind at the time.

## Why this was worth a PR rather than leaving

These were intact, working-looking readers of a container that no longer has a writer. Re-wiring one later would return **zeros rather than failing** — fail-open, and the kind of thing that reads as correct in review. The `#guard`s made it worse by pinning dead code so it could not rot visibly.

Closes #11646.

Issue #11648, opened 2026-08-06
BAL attribution: reconcile the TWO BAI channels (argument-passed system 0 vs the global)


pirapira:
Split out of #11104, which is closed: 18 comments, and its headline premise turned out to be **false** (see below).

## Closing condition

For **every** BAL record producer, the channel supplying its `block_access_index` is named and proven, and the value at the point of record is: **0** for system operations, **i+1** for user transaction *i*, **ulen(txs)+1** post-execution.

## The correction that motivates this issue

#11104 stated *"the missing BAI form is **0**, not `N+1`"*. That is wrong, and the way it is wrong is instructive.

A census of writes to `current_block_access_index` on `origin/main` finds exactly five, and **none is zero**:

```
BlockVerdictMtxRuntime.lean:535  bv_mtx_i + 1
BlockVerdictMtxRuntime.lean:563  bv_mtx_i + 1
BlockVerdictMtxRuntime.lean:776  bv_tx_count + 1
BlockVerdictMtxRuntime.lean:926  bv_mtx_i + 1
BlockVerdictStateRoot.lean:875   bv_tx_count + 1
```

The data initializer is `1`. So the census concludes BAI 0 has no producer. **It does have one** — the system phase passes it as an *argument*, never through the global:

```
BlockVerdictSystemStorageCapture.lean:213
  mv a2, s5; mv a0, s4; li a1, 0; addi a3, s2, 96
```

inside `append_modeled_system_storage_tuple_rows` reaching `.Lamsr_finish_one`. This was already established on 2026-08-02 and re-confirmed on current main.

## So the real question is not a missing producer — it is TWO CHANNELS

BAI reaches records two ways: **argument-passed** on the system path, and **read from the global** elsewhere. Those can disagree. A record on the system path that reads the global would get **1** instead of **0**, silently.

Work: enumerate every **consumer** of `current_block_access_index`, and every callsite that passes a BAI argument, and establish which channel each record takes.

⛔ **Method constraint: do not census stores to the global.** That instrument cannot see an argument-passed value — it is what produced #11104's wrong answer. Enumerate consumers and read the index **per entry**, never as a count.

## Evidence bar

Static ordering plus disassembly, and named system / one-tx / multi-tx / withdrawal vectors. **r200 is smoke only** — a random sample cannot establish provenance.

## Non-goals

[...]

PR #11658, opened 2026-08-06
docs(codegen): document BAI producer channels

### LINKED ISSUE #11657: docs(codegen): BlockAccessListBuilder:816-821 asserts a false FA — both BAI-0 and BAI-N+1 producers exist
`EvmAsm/Codegen/Programs/BlockAccessListBuilder.lean:816-821` documents a producer gap that **no longer exists**, and does so in the form of an explicit false-accept claim:

> THE OTHER TWO CASES HAVE NO PRODUCER IN THE GUEST. Nothing stamps 0 and nothing stamps N+1, so system-level and post-execution changes cannot appear in a rebuilt BAL at all. That is a PRODUCER gap, not a serializer one … a block whose only divergence is a missing system-level change would still be accepted, because neither side of the comparison would contain it.

## Both producers exist on `origin/main` b48148196

- **BAI 0** — `BlockVerdictSystemStorageCapture.lean:213` passes `li a1, 0` to `bal_builder_record_storage_change`, reached from `StateRoot:437` in terminal system replay. It is an *argument*, which is why a census of stores to `current_block_access_index` cannot see it.
- **BAI N+1** — `BlockVerdictMtxRuntime:759` and `BlockVerdictStateRoot:875` both store `bv_tx_count + 1`.

Established while closing #11648, which enumerated every consumer and every BAI-bearing producer callsite.

## Why this is worth a PR rather than leaving

It is not merely out of date — it **asserts a false accept**. Left in place it can be cited as a known FA in a soundness argument, or send someone to build a producer that already exists. Two things went wrong tonight from exactly this shape: #11104's body claimed the BAI-0 producer was missing (its premise came from the same store-census instrument), and a memory note had to be retracted for the same reason.

## Scope

Rewrite `:816-821` to state what the code does: three channels — 0 for system rows supplied as an argument, `bv_mtx_i + 1` for user transactions and `bv_tx_count + 1` post-execution supplied through `current_block_access_index` — and note that the argument channel is invisible to a census of stores to the global, since that is the trap.

Docs-only. `.text` must be **unchanged**; any movement means a non-comment line was edited.

pirapira:
## Summary\n\nCorrect the stale BlockAccessListBuilder comment tracked by #11657. The guest has three BAI producer channels:\n\n- system rows pass BAI 0 explicitly to bal_builder_record_storage_change (BlockVerdictSystemStorageCapture.lean:213);\n- user-transaction rows read current_block_access_index, set to bv_mtx_i + 1 on MTx paths;\n- post-execution rows use the same global after it is set to bv_tx_count + 1.\n\nThe explicit argument is intentionally invisible to a census of stores to current_block_access_index; the system path does not read that global, so it cannot consume a stale user/post value.\n\n## Evidence\n\n- Base: b48148196759ff5e78e2bd4b01138fe392f2d53a\n- Commit: e591545ce\n- Docs-only change; no generated files or pin regeneration.\n- textSizeBytes remains 0x557a0 in EvmAsm/Codegen/RegionMapLinkPins.lean.\n- Fixes #11657.

### DIFF
--- a/EvmAsm/Codegen/Programs/BlockAccessListBuilder.lean
+++ b/EvmAsm/Codegen/Programs/BlockAccessListBuilder.lean
@@ -809,16 +809,20 @@ def blockAccessListBuilderFunctions : String :=
       post-execution    `ulen(transactions) + Uint(1)`          (`:917`), i.e. N+1
       system            0, the value before any transaction sets it
 
-    The guest matches the transaction case exactly:
-    `BlockVerdictMtxRuntime.lean:407` and `:448` compute `bv_mtx_i + 1` into
-    `current_block_access_index`, and `bv_mtx_i` is the 0-based transaction index.
+    The guest realizes all three channels:
+    * system rows pass `0` as an explicit argument to
+      `bal_builder_record_storage_change` (`BlockVerdictSystemStorageCapture.lean:213`);
+    * user-transaction rows read `current_block_access_index`, which the MTx entry
+      paths set to `bv_mtx_i + 1` (`BlockVerdictMtxRuntime.lean:519,547,909`); and
+    * post-execution rows use the same global after it is set to `bv_tx_count + 1`
+      (`BlockVerdictMtxRuntime.lean:759`, `BlockVerdictStateRoot.lean:875`).
 
-    THE OTHER TWO CASES HAVE NO PRODUCER IN THE GUEST. Nothing stamps 0 and nothing
-    stamps N+1, so system-level and post-execution changes cannot appear in a rebuilt
-    BAL at all. That is a PRODUCER gap, not a serializer one -- the emitters encode
-    whatever index the row carries -- but it bounds what a digest comparison can prove:
-    a block whose only divergence is a missing system-level change would still be
-    accepted, because neither side of the comparison would contain it.
+    The explicit-argument channel is intentionally invisible to a census of stores
+    to `current_block_access_index`: such a census sees only the user and post-
+    execution writers and falsely concludes that BAI 0 has no producer.  The
+    system path has no reference to that global, so it cannot accidentally read a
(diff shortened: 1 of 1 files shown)


pirapira:
**coord review: `merge!` applied.** Verified from the diff, not the report.

One file, **+14/−10**, docstring only — **zero emitted-string lines touched**, no generated files, `textSizeBytes` unchanged at `0x557a0`. That is the right shape for a comment fix.

## What it removes is the important part

The deleted text asserted a **false accept**, not merely a stale fact:

> THE OTHER TWO CASES HAVE NO PRODUCER IN THE GUEST … a block whose only divergence is a missing system-level change would still be accepted, because neither side of the comparison would contain it.

Both producers exist. Left in place, that sentence is citable as a known FA in a soundness argument, and it also invited someone to build a producer that is already there. The stale `:407`/`:448` line references were a second, milder problem — those lines have moved.

## What it adds is correct and checkable

All three channels, with references I verified on `origin/main` `b48148196`:

- system rows pass `0` as an **explicit argument** at `BlockVerdictSystemStorageCapture.lean:213`
- user rows read the global, set to `bv_mtx_i + 1` at `BlockVerdictMtxRuntime.lean:519/547/909`
- post-execution rows use the same global at `bv_tx_count + 1` (`MtxRuntime:759`, `StateRoot:875`)

⭐ **And it documents the trap itself**, which is why this is worth more than a tidy-up: a census of stores to `current_block_access_index` sees only the user and post writers and *falsely concludes BAI 0 has no producer*. That is precisely the instrument that produced #11104's false premise — and I reproduced that same wrong answer myself before checking the argument channel. Recording the trap next to the code is what stops the fourth repetition.

The claim that the system path cannot accidentally read a stale user/post value is also verified independently: `BlockVerdictSystemStorageCapture.lean` contains **zero** references to that global.

Closes #11657. Background: #11648.

PR #11659, opened 2026-08-06
docs(mpt): restate bounded-first live contract (#11613)

pirapira:
## Summary
Docstring-only close of the **stale-contract** half of #11613.

## Framing (measured, no collapse impl)
After #11633 (`bnez a6` LCP-prefixed collapse fail-closed):
| set | bounded calls | a0≠0 fallback |
|---|---|---|
| r200 seed=20260806 | 113 | **0 (0%)** |
| 12832/33/34 | 1 each | 1 each → legacy PASS |
| 21353/21354/23274 | 1 each | 0 (bounded ok) |

- **Unsound silent wrong-success:** closed by #11633
- **Stale docstring:** this PR — docs said disconnected/until-KATs while live is bounded-first + fail-closed fallback
- **Capability gap:** residual narrow (12832-class only); optional later; not this PR

## Changes
- `MptBoundedSort.lean`: rebuild + state-root docs state LIVE bounded-first, named unsupported (LCP a6>0), baap legacy on a0≠0
- `BalAccountApplyPostFields.lean`: multi_apply comments match (not optional probe)

## Non-goals
No collapse algorithm. No layout pins. No guest behavior change.

## Test plan
- [x] r200 + named MPT fallback instrument (pre-PR measure)
- [ ] CI docs-only green

### DIFF
--- a/EvmAsm/Codegen/Programs/BalAccountApplyPostFields.lean
+++ b/EvmAsm/Codegen/Programs/BalAccountApplyPostFields.lean
@@ -600,8 +600,9 @@ def mapAccountApplyPostFieldsFunction : String :=
   "  addi t0, t0, 1; la t1, baap_sc_index; sd t0, 0(t1); j .Lbaap_multi_loop\n" ++
   -- #11385: `storage_root` is derived from the storage_writes rows at this
   -- use via `mpt_bounded_storage_root`; it is not a map field or mask bit.
-  -- The explicit legacy call below is only the conservative builder-failure
-  -- fallback and does not introduce a stored storage-root representation.
+  -- LIVE contract (#11613 framing): bounded-first always; legacy only when
+  -- bounded returns a0≠0 (fail-closed unsupported shape, e.g. LCP-prefixed
+  -- collapse after #11633). Not a disconnected/optional probe.
   ".Lbaap_multi_apply:\n" ++
   "  la t0, baap_sc_out_count; ld a4, 0(t0); beqz a4, .Lbaap_nonce\n" ++
   "  la t0, baap_storage_empty_flag; ld t0, 0(t0); bnez t0, .Lbaap_multi_apply_empty\n" ++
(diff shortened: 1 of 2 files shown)


pirapira:
## Gate + substantive (coord)

**GATE**
- Base sha: `b48148196759ff5e78e2bd4b01138fe392f2d53a` (`b48148196` Merge #11638)
- `textSizeBytes` pin: **0x557a0** unchanged (RegionMapLinkPins; no pin diff vs base)
- Diff is Lean doc comments + `#` assembly comments only — no insn bytes
- `python3 scripts/asm_to_program.py check-all` → **CLEAN** (381 defs; bare-J ratchet 175)

**SUBSTANTIVE**
- Removed embedded r200 ~0% and row IDs 21353/23274 from docstrings/asm comments
- Kept structural contract: bounded-first; a0=0 authoritative; a0≠0 → legacy; supported = unprefixed collapse / pure insert / pure delete / small mixed; fail-closed = LCP-prefixed collapse (a6>0 via split_leaf_group)
- Measurement moved to dated #11613 comment; docs point at issue for rate

No self-label. Ready for label when you are.

pirapira:
**coord review: `merge!` applied.** Head `6d2fe4a69`, 2 files, +34/−19.

## Gate satisfied

This PR edits **8 emitted-string lines** (the `#` comments inside the assembly), not only Lean docstrings, so `.text` neutrality had to be shown rather than assumed: **`textSizeBytes` unchanged at `0x557a0`** on base `b48148196`, `check-asm-to-program` **CLEAN**. Assembly comments emit no bytes, and that is now demonstrated rather than argued.

## What this fixes

`MptBoundedSort.lean:742` said *"Extension/leaf splits and deletion collapse remain explicit conservative exits until their canonical cases land"*, and the `:803`-family docstring described the component as *intentionally disconnected from the live verdict* — while `BalAccountApplyPostFields` `.Lbaap_multi_apply_call` wires it live. A component documented as unwired while being wired is a trap that recurs, and the fallback comment compounded it by calling legacy a *"conservative bailout"* isolated so the *"unmasked behavior can be measured independently"*, which reads as optional instrumentation rather than the live fail-closed arm.

The replacement states the actual contract: bounded-first on every multi-descriptor storage apply; `a0=0` authoritative; `a0≠0` falls back to legacy `mpt_state_root_ins`; supported on bounded are unprefixed collapse, pure insert, pure delete and small mixed; **fail-closed is LCP-prefixed collapse** (`split_leaf_group` → collapse with `a6>0`, post-#11633). Plus, valuably, an explicit instruction not to read *"exact-replacement subset"* as *"unwired"*.

## One change I asked for, and why

The first revision embedded a **measurement** — "measured r200 fallback rate: ~0%" — plus specific row IDs. That number goes stale the first time the corpus, seed or bounded subset changes, at which point this docstring becomes the same class of false claim the PR exists to remove. Numbers belong in the artifact and the issue, where they are dated; intent belongs in the docstring. The author moved the rates and row IDs into a dated comment on #11613 and kept every structural statement. That is the right split.

## Basis

The judgement behind this PR was measured, not asserted: a breakpoint after `jal mpt_bounded_storage_root` sampling `a0` gave **113 bounded calls across 92/200 rows with zero fallbacks**, against 100% fallback on the 12832-class. So #11633 closed the unsound half (no silent wrong-success), bounded is earning its keep, and the residual LCP-collapse implementation is an **optional** improvement with a measured 3-row target rather than an aspiration — deliberately not attempted here.

Addresses the contract half of #11613.

PR #11662, opened 2026-08-06
fix(codegen): MCOPY arena bound via memoryArenaLimitAsm (#10523)

### LINKED ISSUE #10523: spec-divergence: MCOPY rejects all ranges past a hardcoded 0x10000 bound (~46x below the affordable arena)

pirapira:
## Summary

Closes #10523. MCOPY was the only memory opcode with a hardcoded **64 KiB** absolute range bound (`li x6, 0x10000`) instead of `memoryArenaLimitAsm`. Spec (EIP-5656 / execution-specs `mcopy`) has no absolute offset cap — only copy-word gas + memory expansion. The guest's own affordability analysis (`EvmMemoryGas.lean`) sizes the depth-0 arena to **4 MiB** (`rootRuntimeMemoryArenaLimitBytes = 0x400000`).

### Fix (`EvmMcopyGas.lean` only + layout pins)

1. **`mcopyRangeGuardAsm`**: one `memoryArenaLimitAsm "mcopy" "x6"` then `bltu` both ends (removed both hardcoded `0x10000` loads). **Only functional change.**
2. **`mcopyDynamicGasAsm`**: **unchanged** vs main — no mulhu, original `mv x19, x5` after after-cost. (mulhu briefly landed then dropped: post-guard ends are arena-clamped so mulhu is dead on reachable inputs and forced an unnecessary register-lifetime change.)
3. Layout: `textSizeBytes` **0x557a0 → 0x557bc** fixpoint; `GuestAddrs` + `symbol-addresses.tsv` regen.

### x13 census (nested safe)

At `h_MCOPY`, `x13`/`a3` is the live **EVM memory base** (handler later `add dst/src, x13, offset`; `CallFrameDescend` repoints child base). Nested path is `evm_memory_pool_end - x13` via the helper — correct both depths. Disasm confirms `sub t1, t1, a3` on nested arm; root arm `lui t1, 0x400` (= 4 MiB).

### Positive bound evidence (SPIKE live)

Fixture `11516` (mcopy empty length_1) at range-check PC:

| ELF | limit reg (`t1`) at dst/src end check |
|-----|----------------------------------------|
| base (main) | **0x10000** (64 KiB) |
| cand | **0x400000** (4 MiB root) |

**Honest scope:** the newly permitted range (ends past 64 KiB up to the frame arena) is **unexercised by the corpus**. mcopy-named 183 and r200 are 0/0 both sides — no row MCOPYs past 64 KiB. The fix rests on (1) the live guard-bound observation above and (2) the spec argument that EIP-5656 has no absolute offset cap, only gas. Issue example `dest=0x10001 size=32` is a **derivation** from those two (end 0x10021 fails base 0x10000, passes cand 0x400000; expansion gas ≈14 350 ≪ 2^24) — not a corpus/SPIKE execution of that exact op. Sufficient for an FR-only change; not reported as a measurement.

### mulhu + x19 (resolved — dropped)

Coord accepted Q1 arithmetic: mulhu cannot fire post-guard. Reduced diff drops all four mulhu/bnez pairs and restores original `mv x19, x5` lifetime. Unification with `updateActiveMemorySizeAsm` deferred to a deliberate both-sites PR if wanted.

**textSizeBytes:** predicted **0x557cc** (0x557ec − 0x20 for 4×mulhu+bnez); measured **0x557bc** (Δ −0x10 vs predict — likely 2-byte alignment/padding in .text; second pin pass NO-OP). Base main was 0x557a0; arena-only net **+0x1c**.

### Measure vs main `b48148196`

| set | result |
|-----|--------|
| mcopy-named **183** | F→P=**0** P→F=**0** (all already OK both sides) |
| r200 seed=20260806 | F→P=**0** P→F=**0** |
| controls (implicit in r200) | clean |

ELF base sha256 `8adc66b8…` cand `96d6e483…` @ slim tip  
`.text` 350112 → 350140 (Δ +28 = **0x1c**); pin **0x557bc** fixpoint.

### Out of scope (issue item 3)

Collapsing MCOPY onto `updateActiveMemorySizeAsm … true` (single-writer gas) left for a later proof-oriented change — not required to close the FR divergence.

### Do not

- Self-label `merge!`
- Scope creep beyond MCOPY
- Cite `out[0:32]` as state root (#11547 is terminal root compare)


pirapira:
**coord review: `merge!` applied.** Head `f6100ee02`, reduced diff.

## The change

One functional edit: `mcopyRangeGuardAsm` now calls `memoryArenaLimitAsm "mcopy" "x6"` instead of two hardcoded `li x6, 0x10000`. MCOPY was the only memory opcode with an absolute 64 KiB bound where every other site derives the frame bound — rejecting ranges the spec charges for and a 2^24-gas transaction can afford, by ~46×. Direction is **false-reject only**.

This is unification rather than a bigger literal: depth 0 → `rootRuntimeMemoryArenaLimitBytes`, nested → `evm_memory_pool_end - x13`, which is the correct bound at each depth and not expressible as a constant.

## Size accounting

`textSizeBytes 0x557a0 → 0x557bc` = **+0x1c (28 bytes)** versus `main`, consistent with the helper's ~9 instructions minus the two `li` it replaces.

⚠️ The author's stated prediction miss (`0x557cc` predicted, `0x557bc` measured) was computed against the **intermediate** revision that still carried the `mulhu` guards, not against `main`. Against `main` the delta is fully accounted for. Worth stating because a prediction is only meaningful relative to the base the change will merge into.

## What was dropped, and why that matters

An earlier revision bundled `mulhu` overflow guards on four quadratic gas terms, justified as *required* once the bound became frame-derived. It isn't: `mcopyRangeGuardAsm` runs **before** any gas is charged, so post-guard ends are arena-clamped and words² is at most 2^34 at depth 0 / 2^44 nested — both fitting u64 with wide margin. The author verified this, withdrew "required", and reduced the diff.

That removal mattered for a second reason: the guards forced moving the `mv x19, x5` spill because `mulhu` clobbers `x5`, and **register lifetime is where this codebase generates real defects**. The reduced PR has no register-lifetime change versus `main` at all. If the two sites should be uniform, that belongs in a deliberate both-sites PR stating whether the guard is reachable at either.

## Evidence, phrased accurately

- `x13`'s denotation verified at **both** depths, with disassembly (`lui 0x400` → `0x400000`).
- Bound change observed **live** on fixture 11516: `t1` `0x10000` → `0x400000`. That is an execution observation of the guard, not an inference.
- 183 mcopy-named rows and r200 (seed 20260806) both **0/0** flips.
- ⭐ The newly permitted range is **unexercised by the corpus**, and `dest=0x10001 size=32` is a **derivation**, not an executed case. The author states it that way rather than presenting it as a measurement. For an FR-only alignment fix, the guard-bound observation plus the spec's absence of any absolute offset cap is a sufficient case.

Closes #10523.

Issue #110, opened 2026-04-01
Storage model: SLOAD/SSTORE via ECALL


pirapira:
## Goal

Implement SLOAD and SSTORE opcodes that delegate to the zkVM host via ECALL for storage read/write.

## Prerequisites

- #107: Opcode handler wrappers (Handlers.lean)

## Approach

1. Abstract storage as `Map EvmWord EvmWord` (key → value)
2. SLOAD: pop key, ECALL to host with key, host returns value, push value
3. SSTORE: pop key + value, ECALL to host with key+value
4. Use the existing ECALL/syscall infrastructure in `SyscallSpecs.lean`
5. Storage operations are modeled abstractly — the host provides the witness
6. Add new ECALL numbers for SLOAD/SSTORE (per zkvm-standards IO interface)
7. Prove `evm_sload_stack_spec` / `evm_sstore_stack_spec`

## Acceptance Criteria

- [ ] Storage model (`StorageIs`) assertion defined
- [ ] SLOAD spec: `storage[key]` returned correctly
- [ ] SSTORE spec: `storage[key] := value` update
- [ ] ECALL interface defined
- [ ] Stack-level specs proved (0 sorry)
- [ ] `lake build` succeeds

## Milestone
E: Storage, Calls & Precompiles

pirapira:
Crosswalk audit record: #10627 tracks Amsterdam stack, memory, and storage guest correspondence.

pirapira:
Closing as **architecturally superseded** — not merely stale. This issue proposes a trust model the project does not use, and it has already attracted implementation work.

## The guest emits zero ECALLs

This issue asks for SLOAD/SSTORE to delegate to the zkVM host via ECALL, with *"new ECALL numbers for SLOAD/SSTORE"* and *"the host provides the witness."*

There are **no `ecall` instructions emitted anywhere** in `EvmAsm/Codegen/**`. Storage is resolved through the authenticated witness chain: `h_SLOAD` is live at `Storage.lean:411`, and `:267` documents that a demand-driven `h_SLOAD` must reuse that chain.

⛔ The difference is not cosmetic. A per-SLOAD host oracle makes the **host trusted to answer**; today storage authenticity is derived from the state root and the witness. Implementing this issue would weaken the trust model, not complete it.

## Two other criteria do not survive contact

[...]

PR #11663, opened 2026-08-06
refactor(el): remove dead storage ECALL bridges

pirapira:
## Summary\n\nRemove the three dead storage ECALL bridge modules from the EL umbrella:\n\n- EvmAsm/EL/StorageEcallBridge.lean\n- EvmAsm/EL/StorageEcallStackBridge.lean\n- EvmAsm/EL/StorageArgsEcallBridge.lean\n\nRemove their three imports from EvmAsm/EL.lean. The surviving StorageStackExecutionBridge now imports EvmAsm.Evm64.StorageAccess and EvmAsm.Evm64.StorageArgs directly, with StorageAccessList aliased directly to EvmAsm.Evm64.StorageAccess.StorageAccessList. This preserves its type exactly while removing the alias-of-an-alias dependency.\n\n## Evidence\n\n- Base: b48148196759ff5e78e2bd4b01138fe392f2d53a\n- Commit: d146a26ea\n- Named-reference census: zero references remain to all three deleted modules.\n- lake build: passed (4821 jobs).\n- check-unimported.sh: passed (2851 modules, 0 orphans).\n- textSizeBytes remains byte-identical at 0x557a0; RegionMapLinkPins.lean was not modified.\n- No generated files or pin regeneration.


pirapira:
**coord review: `merge!` applied.** Verified the deletion boundary from the tree at head `d146a26ea`, not from the report.

| check | result |
|---|---|
| deleted | `StorageEcallBridge`, `StorageEcallStackBridge`, `StorageArgsEcallBridge` — all three absent |
| survivors | `StorageAccessBridge`, `StorageStackBridge`, `StorageStackExecutionBridge` — all three **present** |
| conformance consumer | `EL/Conformance/StorageStackExecution.lean` still references the survivor (4 hits) |
| `textSizeBytes` | **`0x557a0` unchanged**, `RegionMapLinkPins` untouched — as predicted |
| gates | `lake build` 4821 jobs; `check-unimported` 2851 modules, **0 orphans** |

## Why this needed care

The naive deletion takes live files with it. `StorageEcallBridge` imported `StorageAccessBridge`, and `StorageEcallStackBridge` imported `StorageStackBridge` — both of which are **live**, imported independently at `EL.lean:144-145`. Following the dependency graph downward would have removed them.

⭐ **And my own stated graph was wrong.** I told the author the imports flowed inward with nothing outside referencing the three, derived from importer *counts* that were identical (2 each) — but for `StorageArgsEcallBridge` the second importer was `StorageStackExecutionBridge`, a live non-aggregator module reached from conformance. The author **refused the deletion and reported the contradiction** instead of proceeding. That is the correct response to a graph claim that does not match the tree, and it is why the fix here also includes a repointing rather than being a pure delete.

The repointing is minimal and meaning-preserving: `StorageStackExecutionBridge`'s `abbrev StorageAccessList` was an alias of an alias (`StorageArgsEcallBridge.StorageAccessList` → `Evm64.StorageAccess.StorageAccessList`), so it now names the underlying definition directly with narrow imports. Same type, one fewer hop.

## Context

Closes the residue of #110, closed as architecturally superseded: the guest emits **zero `ecall` instructions**, and storage authenticity comes from the authenticated witness chain rather than a host oracle. These modules defined `StorageSyscall` selectors "reserved" for an interface that does not exist — pure Lean, zero guest bytes, but intact scaffolding that invites someone to wire a nonexistent mechanism.

A broader `EL/` reachability analysis is queued behind this and deliberately not bundled: `EL.lean` imports every module and `EvmAsm.lean` imports `EL`, so `check-unimported` can never flag a dead EL module — being imported by the aggregator is not evidence of liveness, which is exactly how these three survived.

Issue #11523, opened 2026-08-05
Decoded: replace nonce/balance LENGTH limits with VALUE bounds (leading-zero tolerant) — non-canonical witness leaves are a live false-reject surface on any block touching them


pirapira:
Split out of #11484, which has accumulated several threads. This is the one concrete change.

## The change

`Decoded` (`AccountDecodeSpec.lean:144-153`) constrains four **field lengths**. Two of those clauses should become **value bounds**, and two should stay lengths — because the spec treats the two pairs of fields differently.

| field | current | replace with | why |
|---|---|---|---|
| 0 nonce | `l0.toNat ≤ 8` | any length, provided the **significant** bytes fit **u64** | leading zeros are padding |
| 1 balance | `l1.toNat ≤ 32` | any length, provided the significant bytes fit **u256** | leading zeros are padding |
| 2 storage_root | `l2.toNat = 32` | `l2 ∈ {0, 32}` | leading zeros are **part of the hash** |
| 3 code_hash | `l3.toNat = 32` | `l3 ∈ {0, 32}` | same |

## Why the pairs differ — the spec encodes it

[`witness_state.py:112-118`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/witness_state.py#L112-L118):

```python
nonce        = Uint(int.from_bytes(decoded[0], "big")) if decoded[0] else Uint(0)
balance      = U256(int.from_bytes(decoded[1], "big")) if decoded[1] else U256(0)
storage_root = Root(decoded[2])   if decoded[2] else EMPTY_TRIE_ROOT
code_hash    = Hash32(decoded[3]) if decoded[3] else EMPTY_CODE_HASH
```

- Fields **0/1** go through `int.from_bytes` — an **integer** parse. Leading zeros are insignificant, and neither `Uint` (unbounded) nor `U256` (bounds the *value*) constrains length. The only assertion in the function is `len(decoded) == 4`.
- Fields **2/3** are `Root`/`Hash32` = `Bytes32`, and `FixedBytes.__new__` (`ethereum_types/bytes.py:34-35`) **raises** on any length ≠ 32. A 31-byte code hash is not a padded 32-byte one.

[...]

PR #11665, opened 2026-08-06
proof(#11516): state the decode_account_from_leaf ↔ Decoded pairing as a registry row

### LINKED ISSUE #11516: Correspondence registry covers only family=rlp: account-leaf decode was never graded, so Decoded's l2=l3=32 divergence from SpecRef+spec went unsurfaced
## SpecRef is **correct** on relaxed decoding. The gap is that nothing compares our asm specs against it.

Checked before filing, because the question was whether SpecRef has bugs here. It does not.

`EvmAsm/Stateless/SpecRef/WitnessState.lean:117-126`, `decode_account_from_leaf`:

```lean
let nonce   : Uint   := if n.isEmpty  then 0               else bytesBEtoNat n
let balance : U256   := if b.isEmpty  then 0               else bytesBEtoNat b
let storageRoot : Root   := if sr.isEmpty then EMPTY_TRIE_ROOT else sr
let codeHash    : Hash32 := if ch.isEmpty then EMPTY_CODE_HASH else ch
```

against [`witness_state.py:112-118`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/witness_state.py#L112-L118):

```python
nonce        = Uint(int.from_bytes(decoded[0], "big")) if decoded[0] else Uint(0)
balance      = U256(int.from_bytes(decoded[1], "big")) if decoded[1] else U256(0)
storage_root = Root(decoded[2])   if decoded[2] else EMPTY_TRIE_ROOT
code_hash    = Hash32(decoded[3]) if decoded[3] else EMPTY_CODE_HASH
```

Field-for-field identical, including the malformed case (`len(decoded) == 4` ↔ `.error .accountLeafMalformed`) and the absence of any length constraint. **SpecRef modelled the folds correctly all along.**

## So the drift was asm-vs-spec, and our own predicate was the outlier

`AccountDecodeSpec.lean:144-153` `Decoded` requires:

| field | `Decoded` | spec / SpecRef |
|---|---|---|
| 0 nonce | `l0 ≤ 8` | no constraint |
| 1 balance | `l1 ≤ 32` | no constraint |
| **2 storage_root** | **`l2 = 32`** | **0 → `EMPTY_TRIE_ROOT`, else value** |
| **3 code_hash** | **`l3 = 32`** | **0 → `EMPTY_CODE_HASH`, else value** |

`Decoded` was describing **what our asm did**, not what the spec requires — and for fields 2 and 3 that is a real divergence, since length 0 is legal in the spec and rejected by us. (Fields 0/1 already admit 0, so no false rejection there; their strictness is a separate, harmless-looking observation worth a second look but not a defect.)

⭐ **This is why the fix direction is now confirmed three independent ways**: the spec source, codex's executed spec run on the filled fixture (#11508), and SpecRef — which was right the whole time and could have told us hours earlier if anything had compared them.

## The actual defect: no gate ties asm specs to SpecRef

We maintain a Lean port of execution-specs *specifically to be the reference*, and a Lean predicate describing the asm that **contradicted it**, and **nothing noticed.** Every gate we have checks a different axis:

- `asm-to-program` — emitted bytes match the Program
- `check-declared-starts` / region-map pins — addresses match the linked image
- Class-A ratchet — supplied-BAL edge count
- `check-axioms` — no TCB expansion

**None compares an asm-side specification against SpecRef's model of the same spec function.** So a predicate can drift from the reference indefinitely, and the only way it surfaces is a fixture failing for an unrelated reason — which is exactly how this one surfaced.

Suggested: for spec functions that have both a SpecRef model and an asm-side predicate, state the correspondence as a theorem or at minimum a documented pairing, starting with `decode_account_from_leaf` ↔ `Decoded`/`DecodedFolded`. A pairing that cannot be stated is itself a finding.

⚠️ And the standing hazard to keep in view: **SpecRef is a port and drifts.** It happens to be right here, which is the good case. The general rule stays — cite execution-specs at the pin, never SpecRef alone, since citing our own Lean to justify spec alignment is circular. What SpecRef is good for is exactly this: a cross-check that would have caught the asm early.

Related: #11484 (the sibling-predicate ruling), #11483 (the asm fold), #11508 (spec executed on the fixture).

dhsorens:
Closes #11516.

## What that issue actually asks for

> Suggested: for spec functions that have both a SpecRef model and an asm-side predicate, state the correspondence as a theorem or **at minimum a documented pairing**, starting with `decode_account_from_leaf` ↔ `Decoded`/`DecodedFolded`. A pairing that cannot be stated is itself a finding.

⭐ **The theorem already existed.** `AccountDecodeCompose.decoded_matches_specRef` is proven *and* already witnessed. What was missing was the **row** — so the pairing was real in Lean and invisible to every census.

That's the **fourth instance today** of *both halves exist, the join doesn't* (after `bloom_or_into`'s bridge in #11348, and both `hpUnpackPairs`/`keyToNibbles` and `hpDecode`/`compact_to_nibbles` in #11422). It's the same failure mode #11637's gate was built for, seen from the correspondence side rather than the routine side.

## Graded `.domainRestricted`, not `agrees` — on purpose

| field | status |
|---|---|
| 2, 3 (`storage_root`, `code_hash`) | ✅ closed by #11483/#11484 — zero-length now folds to `EMPTY_TRIE_ROOT`/`EMPTY_CODE_HASH`, matching `if decoded[2] else EMPTY_TRIE_ROOT` |
| 0, 1 (`nonce`, `balance`) | ⚠️ `Decoded` still carries `l0 ≤ 8` / `l1 ≤ 32` **length** limits where the reference imposes none — it just runs `int.from_bytes` |

So a zero-prefixed or over-width nonce is accepted by the reference and outside what we prove. That's **#11523, still open**, and this row is precisely what keeps the gap visible instead of assumed. Recorded as a *coverage gap*, not an ABI precondition: the guest may well handle those inputs; we have not proved it.

## ⚠️ Explicitly not a `portDefect`

#11516 checked `decode_account_from_leaf` field-for-field against [`witness_state.py:112-118`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/witness_state.py#L112-L118) and found them identical, malformed case included. **SpecRef modelled the folds correctly all along** — the outlier was our own predicate. The note repeats that issue's standing hazard anyway: SpecRef is a port and drifts, so cite execution-specs at the pin, never SpecRef alone, since citing our own Lean to justify spec alignment is circular.

## Witness added *with* the row

Applying #11348's lesson rather than repeating it: a row naming an unwitnessed theorem leaves that theorem invisible to the axiom gate, and that's a separate obligation from claiming a tier.

This required importing `AccountDecodeClose6` explicitly into `Routines.lean` — `AccountDecodeCompose` imports `AccountDecodeBridge`, **not** Close6, so the triple's module wasn't in scope. My first attempt failed with `Unknown identifier` for exactly that reason, not a wrong namespace; the import carries a comment so the next person doesn't re-diagnose it.

## Second real burn-down of #11637's allowlist

`account_decode` was a tier-B entry there (tier B only because its file names the base through a local abbrev — the false-negative I documented when seeding). Registering it deletes the entry: **uncovered 102 → 101, registered 28 → 29.** Seeded at 103 this afternoon.

## Verification

- `lake build EvmAsm.Progress.Correspondence EvmAsm.Progress.Routines EvmAsm.Progress.AxiomWitnesses` — green
- censuses: `registry_size` 28→29, `domainRestricted` 6→7, `bridged` 11→12, new `account_rows`
- `check-axioms.sh` — OK, 142 witnesses, classical axioms only
- `check-registry-coverage.py` — OK (29 registered, 101 uncovered, 101 allowlisted)
- `progress-report.sh --write` — regenerated


pirapira:
**coord review: `merge!` applied.** Verified against `origin/main` and the pinned spec, not from the body.

## The central claim holds

`AccountDecodeCompose.decoded_matches_specRef` exists at `:199`, is **already witnessed** (`AxiomWitnesses.lean:35`) and already referenced (`Routines.lean:452`), and `AccountDecodeCompose.lean` has **zero `sorry`**. So the theorem was real and only the *row* was missing — the pairing existed in Lean and was invisible to every census. That is exactly the shape #11637's gate was built for, and this PR also **removes** `account_decode` from `registry-coverage-allow.txt`, so the allowance shrinks rather than the gate being worked around.

Counts are internally consistent: `registry_size` 28 → 29, `domainRestricted` 6 → 7, `bridged` 11 → 12, plus a new `account_rows` invariant pinning the family at 1.

## The grading is right, and right for the right direction

`Decoded` being **stricter** than the reference is a false-*reject* exposure, not a false-accept: strict-guest-versus-lenient-spec loses valid blocks, and the mirror error would be the dangerous one. Grading `.domainRestricted` rather than `agrees` keeps that visible instead of assumed, and calling it a coverage gap rather than an ABI precondition is the honest framing — the guest may well handle those inputs; we have not proved it.

## One refinement to the note, for #11523's scope

The note says `l0 ≤ 8` **and** `l1 ≤ 32` are limits "where the reference imposes no length constraint at all." That is exactly right for **nonce** and slightly too broad for **balance**. From `witness_state.py:112-118`:

```python
nonce   = Uint(int.from_bytes(decoded[0], "big")) if decoded[0] else Uint(0)
balance = U256(int.from_bytes(decoded[1], "big")) if decoded[1] else U256(0)
```

- **nonce** → `Uint`, unbounded. So the reference accepts both leading-zero padding *and* genuinely ≥ 2^64 values. `l0 ≤ 8` diverges on both.
- **balance** → `U256`, which **is** bounded. A 33-byte balance whose value is ≥ 2^256 is rejected by the reference too, so `l1 ≤ 32` coincides with `U256`'s own domain there. The divergence is narrower: only **over-width-but-in-range** encodings, i.e. leading-zero padding of a value < 2^256.

Not a defect in this row — the grade and direction are unchanged either way — but it matters for what #11523 must actually cover: nonce needs both magnitude and padding, balance needs padding only. Worth carrying into that issue so the fix is not scoped to a magnitude case that cannot arise.

## Consistent with tonight's separate finding

This aligns with what was established while closing #120: the spec decodes **account** fields leniently (`rlp.decode` + `int.from_bytes`, no canonicity check) while decoding **header** integers strictly (`rlp.decode_to`). Our `rlp_field_to_u64` correctly mirrors the strict side. So the leniency here is not sloppiness in the reference — it is a per-context choice we must match per context, and this row is where the account side of that stays honest.

Closes #11516.

PR #11668, opened 2026-08-06
refactor(el): remove unreachable non-Conformance modules

pirapira:
## Summary

Remove 81 EL modules that are unreachable from every module outside EvmAsm/EL, and remove their EL.lean imports. This is stage 1 of the EL cleanup; Conformance is deliberately staged separately.

## Reachability evidence

Measured on current main 424b1122974542b472f52a42ec2b2daa94774e6d:

- 177 EvmAsm/EL Lean files, 23,426 lines.
- 41 external importers.
- 59 modules in the transitive EL-only closure.
- 118 candidates before staging: 102 non-Conformance remainder and 16 EL/Conformance files.
- This PR deletes 81 of the 101 non-Conformance candidates. The other 20 are retained because the Conformance files still import them transitively; EvmAsm.EL.Conformance (the root umbrella outside the EL/Conformance/ path) is also retained. Stage 2 will decide that cluster together with the 16 Conformance files.

The deleted-module sweep covered the whole repository, including non-Lean files. The only non-Lean references were three scripts/noshake.json entries; those were removed. No other non-Lean references remain.

## Gates and predicted payoff

Before building, the prediction was an 81-job reduction: 4,821 baseline Lean jobs to 4,740 candidate jobs, one source job per removed module. The candidate build completed successfully at 4,740 jobs.

- lake build: PASS (4,740 jobs)
- scripts/check-unimported.sh: PASS (2,770 reachable / 2,770 total)
- git diff --check: PASS
- RegionMapLinkPins.textSizeBytes: unchanged at 0x557bc
- guest ELF sha256: unchanged at ef6dc4f5d02a6f963a23c7cc6006d409e47c5aee96e6c7149061758be2b2d72e

The deleted files are pure Lean EL scaffolding and contribute no guest bytes; the unchanged pin and ELF hash are the byte-identity gate.

## Stage 2 checklist (not changed here)

Keep: Conformance/Calldata, Code, ExpGas, ExpStackExecution, RLP, RLPFullDecodeBridge, ReturnData, SignedArithmeticStackExecution. These exercise surviving Evm64/RLP implementations.

Delete candidates: Conformance/All (empty), Call (aliases only), CreateStackExecution, KeccakStackExecution, Log, LogStackExecution, StorageStackExecution, TerminatingStackExecution. Their bridge dependencies are retained in this PR because the stage-2 decision is separate.

Import reachability is not a usefulness audit. In particular, the 33 precompile ECALL bridges survive only because Stateless/VM/Precompiles imports them; the guest emits no ECALL instructions. This PR removes what nothing outside EL references and does not claim that survivors are valuable.


pirapira:
**coord review: `merge!` applied** on maintainer instruction, with the closure re-verified against current `main` `6defda4fb` rather than trusted from when the PR was cut.

## Why re-verification was needed

This PR was opened at 23:22 against `424b11229` and has sat unlabelled while the EL lane was deprioritised — **20 commits** of `main` have landed since. I had assumed the dependency closure would have decayed. It has not, and I was wrong to assume it:

| check | result |
|---|---|
| the 83 deleted paths still present on `main` | **83 / 83** — empty conflict surface |
| external importers of `EvmAsm.EL.*` | **41**, unchanged from the census |
| `EvmAsm/EL/**.lean` file count | **177**, unchanged |
| modules outside the delete set importing any of the 81 deleted EL modules | **0** |

Nothing in those 20 commits touched `EvmAsm/EL` or its importers. My earlier claim that #11673 had shifted the closure was wrong — #11673 deleted files under `Codegen/Programs`, not under `EL/`.

## What it deletes, and why the gate is byte-identity

81 EL modules unreachable from any of the 41 external entry points, plus the 7 conformance modules whose only dependencies were in that set (a 1:1 bridge↔check pairing, so those checks existed solely to exercise deleted bridges), and the matching `EL.lean` imports. Three non-`.lean` `noshake` references were removed too — found by sweeping beyond `*.lean`, which is where this class of deletion usually breaks.

Evidence as cut: `lake build` green at **4,740 jobs** against 4,821 — exactly the 81 removed modules, predicted before measuring — `check-unimported` **2,770/2,770 with zero orphans**, and `textSizeBytes` plus the guest ELF sha **byte-identical**. That last one is the proof the deleted set contributed nothing to the emitted guest.

## Two caveats recorded rather than glossed

⚠️ **The closure was hard to pin down.** The set went **118 → 102 → 101 → 81** across four builds, each revision forced by a build failure showing the dependency closure was wider than the previous cut assumed — the transitive conformance support closure alone pulled back 20 modules. `EL.lean` imports every EL module and `EvmAsm.lean` imports `EL`, so `check-unimported` can never flag a dead EL module: being imported by the aggregator is not evidence of liveness, which is exactly how these 81 survived.

⚠️ **Import-reachability is not usefulness.** This removes what nothing outside `EL/` references; it does **not** establish that everything surviving is used. The 33 precompile ECALL bridges are "live" only because `Stateless/VM/Precompiles` imports them, while the guest emits zero `ecall` instructions — the same property that made the storage ECALL bridges dead in #11663. The survivor set should not be cited as vetted.

`BEHIND` is bookkeeping here: no generated artefacts in the diff, so nothing to regenerate, and the conflict surface is empty.

PR #11669, opened 2026-08-06
test: pin known-flippable negative controls

pirapira:
## Summary

Keep the seven #11508 negative fixtures as always-run named controls in `scripts/spike/account_agreement_sweep.py`:

- 00349, 00350, 26003, 02965, 02967, 02970, 02998
- expected oracle and guest verdict: `succ=0` for each
- current `bv_fail` values are documented only (`7/7/7/11/11/40/1`), not asserted

The controls are selected before the random sample, so they remain named and are never consumed by `--random-count`.

The manifest's oracle column is checked explicitly, and the guest output's `+32` success byte is asserted for the seven controls. The report records oracle, expected, and guest verdicts.

The module docstring records that the account-agreement probe was retired in #11638; this script is now the standing verdict/control sweep and no longer performs map-versus-live comparison.

## Evidence

Base: `424b1122974542b472f52a42ec2b2daa94774e6d`

Guest: `gen-out/11508-main/stateless_guest.elf`

Guest SHA256: `96d6e483f4908afa99788f8131d55df87a7d9594227b02aa318834b993a08326`

- `python3 -m py_compile scripts/spike/account_agreement_sweep.py`
- `git diff --check`
- named-control run with `--random-count 0`: 9 controls (the existing 2 plus these 7), `flip_set=[]`, `output_differences=0`; all seven new controls matched `oracle=0` and `guest_succ=0`
- diff is one file, 66 insertions / 10 deletions; one in-repo caller


pirapira:
**coord review: `merge!` applied.** One file, +95/−12, script-only — no guest bytes.

## What this pins

Seven known-negative fixtures join `STANDING_NAMED_CONTROLS` (9 total): `00349`, `00350`, `26003`, `02965`, `02967`, `02970`, `02998`. These are the rows from #11508 — correct at `020802bd7`, **false-accepting** at `d6047135f`, correct again on current `main`, with **no bisect in either direction**. A row that has flipped both ways and currently passes for unknown reasons is exactly what a pinned control is for. Until now, the only thing that would have caught the next flip was another full 26,104-row sweep — which is why the last regression survived long enough to be re-cited as FA=0.

## Why this is a gate and not decoration

⭐ **`--self-test` demonstrates the control FAILING.** It feeds `oracle_succ=0, guest_succ=1` through `named_verdict_mismatch` — **the same helper the real gate calls** — and fails loudly if the synthetic false-accept is accepted. A self-test on a parallel code path would prove nothing; this one shares the path.

That matters here more than usual. This gate catches a regression nobody has ever observed, and it lives in a script whose primary mechanism was retired underneath it (see below). Every ingredient for a check that looks live and reaches nothing was present. Now it has been seen failing.

Two further details that make the assertion real rather than nominal:

- It **asserts the manifest oracle IS 0** rather than assuming it, and rejects malformed or non-binary values plus rows without 4 columns. A control that silently accepts a garbage oracle asserts nothing while still passing green.
- `bv_fail` is recorded as documentation, **not** asserted. A code is a rejection *tail*, so pinning today's 7/7/7/11/11/40/1 would fail the gate when a row starts rejecting for a different but entirely valid reason.

## The docstring, and an open judgement

The script is `account_agreement_sweep.py`, but its account-agreement probe was **retired in #11638** — the emitted `account_agreement_probe` is a no-op and no map-versus-live comparison happens. The added docstring states that plainly, so a future reader grepping for the retired harness does not conclude the script is dead and delete it, taking the FA controls with it.

⚠️ **Renaming remains open and is cheaper than it looks.** `account_agreement_sweep` appears in the content of exactly **one** file on `main` — `EvmAsm/Codegen/Programs/AccountWriteMap.lean`, as prose, not a call. So a rename costs a file move and one comment edit. Landing the docstring now is right, but a name that advertises a dead mechanism with a docstring apologising for it is a worse end state than a rename, and follow-up renames reliably do not happen.

Follow-up from #11508 that this does **not** cover: a fresh full-corpus FA sweep. Seven named rows are not 26,104, and FA=0 remains unestablished on current main.

Issue #11522, opened 2026-08-05
Untouched-leaf bytes are load-bearing for the post-state root: verify we carry them verbatim rather than re-encoding (and Decoded's l0<=8 makes acceptance depend on whether the account was read)


pirapira:
The post-state root is computed over the witness's **own leaf bytes** for untouched accounts, so their exact encoding — leading zeros included — is part of the committed hash. Split out of #11521, whose allocation thesis was refuted; this observation is independent of it.

## Spec behaviour

`compute_state_root_and_trie_changes` (`witness_state.py`) builds the MPT from `self._node_db` and calls `mpt_set` **only** for entries in `account_changes` and `storage_changes`, then takes `mpt_root`:

```
for address, slots   in storage_changes.items(): ... mpt_set(storage_mpt, key, value)
for address, account in account_changes.items(): ... mpt_set(...)
```

| account | spec behaviour | what the root depends on |
|---|---|---|
| changed | `mpt_set` with a re-encoded account | canonical encoding |
| **untouched** | never touched | **the witness's original bytes** |

Ingestion applies no filter: `build_node_db` (`witness_state.py:36-41`) is `db[keccak256(entry)] = entry`; `build_code_db` likewise. No canonicality or structural check.

⇒ **Decode-then-re-encode is not a valid strategy for untouched accounts.** Any information the decode discards is information the root needs.

## Two things unverified on our side

1. **Does our post-state root computation carry untouched leaf bytes verbatim, or re-encode from decoded fields?** If it re-encodes, a non-canonical leaf diverges on the **root** — reachable with no account being read at all, and independent of any decode predicate. `WitnessStateRoot.lean` / `IncrementalMpt.lean`.
2. **Can a non-canonical leaf appear under a parent root that a spec-built chain produced?** The spec's encoder is canonical, so a real chain should not contain one; a constructed parent state can.

## An asymmetry in our acceptance predicate

`Decoded` (`AccountDecodeSpec.lean:144-153`) requires `l0 ≤ 8`, so it rejects a non-canonically encoded nonce. But **decoding only happens for accounts we look up:**

| leaf | our decode runs? | `l0 ≤ 8` fires? | bytes must reach the root? |
|---|---|---|---|
| non-canonical, account **read** | yes | **yes → reject** | yes |
| non-canonical, account **never read** | no | no | **yes** |

[...]

PR #11671, opened 2026-08-07
docs(plan): record the registry-coverage gate and the witness-vs-row rule

dhsorens:
Per CLAUDE.md's PLAN.md maintenance rule — *"Added new infrastructure: Update the Infrastructure section."* Docs only.

Two entries from today's proof-issue batch, both chosen because **other agents will hit them**:

### 1. `check-registry-coverage.py` (#11637)

The gate that closed the row-**existence** hole. Adding a linked, spec-bearing routine without a row now fails the build, and the allowlist expires rather than accumulating.

Worth recording because it changes what a build failure can mean, and because there is exactly one dishonest way to silence it — grading a tier-B structured SAsm spec as `.proven` to shorten the allowlist. That converts a visible backlog into a false census, which is the defect the gate exists to prevent, so the entry says so explicitly.

Also records the stated limit: the symbol↔theorem map is name-based, so it can **under**-report a gap but never invent one — a floor on the backlog, not a census.

### 2. Witness ≠ row (#11348, #11516)

Naming a theorem in a registry row does **not** put it in the axiom gate — the `private noncomputable abbrev _<name>_witness := @…` is what does. They are separate obligations: the witness is always warranted for a named theorem; the row claims a *tier*, and only if earned.

I conflated these on #11348, CI caught it, and I hit the adjacent import trap again on #11516 (`AccountDecodeCompose` imports `AccountDecodeBridge`, not `Close6`, so the triple was not in scope). Worth stating once where people will read it rather than leaving it to be rediscovered.


pirapira:
**coord review: held pending one question — not a defect claim, a check I could not settle myself.**

The PLAN.md half is unambiguously good and I would take it as-is. Recording that `check-registry-coverage.py` changes *what a build failure can mean*, and naming the single dishonest way to silence it — grading a tier-B structured SAsm spec `.proven` to shorten the allowlist — is exactly the kind of entry that saves a future agent from a plausible wrong move. Same for the witness-versus-row distinction: *"referencing a theorem does NOT put it in the axiom gate; the `private noncomputable abbrev _<name>_witness` is what does"* is a real trap, and the note that the map is name-based and therefore a **floor on the backlog rather than a census** is the right epistemic caveat.

## The question, on the `State.lean` half

The PR adds `.globl` to three data labels (`ae_nonce_len`, `ae_balance_len`, `ae_scratch`) and states that this *"changes the SYMBOL TABLE only, never `.text`, so the byte-identity gates are unaffected."* The `.text` part I agree with. But the symbol table is the input to the generated address artefacts:

- None of the three appears in `scripts/asm-fixtures/symbol-addresses.tsv` on `main` (grep count 0 for each).
- This PR touches **no generated artefact** — only `State.lean` and `PLAN.md`.

If `gen-symbol-addresses.py` derives its rows from `nm` on the linked ELF, promoting these to global should add three rows, and the pinned TSV would then be stale relative to the ELF. That also seems required by the PR's own rationale: the `.globl` exists so `asm_to_program.py` can resolve the `la` targets, which means the converter has to *see* them in the address table.

So, one of:

1. the TSV (and possibly `GuestAddrs.lean`) regeneration is owed and missing;
2. the generator filters to a set these do not enter, so nothing changes — in which case, does the converter actually gain the resolution the change is for?
3. they were already emitted as local symbols the generator picks up, and `.globl` changes nothing observable.

I could not distinguish these without building, and I would rather ask than guess: I have twice tonight built a confident claim on an artefact I read from the wrong place.

**Byte-identity is the discriminator.** If you rebuild and `symbol-addresses.tsv` is unchanged, say so and I will label immediately — that settles it. If it gains three rows, the regeneration belongs in this PR, since a source change that alters the symbol table while its derived artefacts stay pinned is the shape that produces a green local build and a failing CI.

pirapira:
**Hold withdrawn — my premise was wrong. `merge!` applied.**

I asked whether `.globl` on the three data labels owes a `symbol-addresses.tsv` regeneration. It does not, and I could settle it without a build after all.

`scripts/gen-symbol-addresses.py:113-126` runs `readelf -sW` and filters on **type and section only** — it skips `UND`, `SECTION`/`FILE`, empty names, and `$`-prefixed RISC-V mapping symbols. It **never filters on Bind**, so `LOCAL` symbols are already included. Promoting a label from local to global therefore cannot add a row: the generator's inclusion criterion does not depend on the property this PR changes. Your statement that this touches the symbol table only, leaving byte-identity gates unaffected, is correct.

I should have read the generator before raising it. The check cost you nothing but it cost a review cycle, and the lesson is the same one this evening keeps teaching: read the tool that consumes the artefact before reasoning about what changes it.

## The `State.lean` half is well-argued on its own terms

The comment records *why* the `.globl` exists — so `asm_to_program.py` can resolve the `la` targets, because the mechanical converter refuses the routine otherwise, and **hand conversion is what produced #11518 (reversed SD operands) and #11519 (laLo anchored at the wrong PC)**. Naming the two defects that hand conversion actually caused is what makes this stick; "prefer the converter" alone would not.

## One open question, not a blocker

None of `ae_nonce_len`, `ae_balance_len`, `ae_scratch` appears in the pinned TSV today, and per the above, `.globl` is not what would put them there. So if the converter currently refuses `account_encode` for want of those addresses, this change alone may not be sufficient — the symbols would need to be present in the linked image the TSV is generated from. Worth confirming the converter now accepts the routine; if it does, my model of the mechanism is incomplete and that is fine. It does not gate this PR either way, since the directive is harmless and the PLAN.md content stands on its own.

## PLAN.md

Both entries are the kind that stop a future agent making a plausible wrong move. Naming the single dishonest way to silence the coverage gate — grading a tier-B structured SAsm spec `.proven` to shorten the allowlist — is more useful than the gate's description, and the witness-versus-row distinction (`private noncomputable abbrev _<name>_witness` is what enters the axiom gate, not a reference) is a real trap. The caveat that the symbol↔theorem map is name-based and therefore **a floor on the backlog rather than a census** is the right epistemic framing and matches how the 103 figure should be read.

dhsorens:
Two things I owe you here, one of them a mistake on my part.

## ⚠️ This PR was NOT "docs only", and my description said it was

You spotted the `State.lean` `.globl` change and reviewed it on its merits. You should not have had to: **I told you on #11522 that I was "holding the `.globl` branch locally rather than opening a PR that does nothing by itself," and then shipped it inside this one.**

Cause: I cut `docs-11637-plan-infrastructure` while still standing on `proof-11522-ae-globl`, so the new branch inherited that commit, and I wrote the PR body from what I *intended* to include rather than from the diff. The header says "Docs only." It touched emitted asm source.

The change is harmless and you've now approved it independently, so nothing needs reverting. But a PR labelled docs-only that carries an asm-source edit is exactly the shape that makes review labels untrustworthy, and it cost you a review cycle on a question you'd otherwise never have needed to ask. My fault, and the fix is mechanical: read `gh pr view --json files` before writing the body, not my own memory of the branch.

## Your open question — answered, and your instinct was right

> if the converter currently refuses `account_encode` for want of those addresses, this change alone may not be sufficient

It is not sufficient, and I confirmed that empirically before you asked. I added the `.globl`s, rebuilt the guest locally (`as`/`ld` were enough, no `gcc` needed), regenerated, and the converter still refuses:

```
ConvError: la ae_scratch: symbol not in address table (BLOCKED_ON_.6)
```

The reason is upstream of `.globl` entirely: **`account_encode` is not in the guest image at all.** It is emitted only inside two *probe* BuildUnits. `nm` on `stateless_guest.elf`, `stateless_guest.s`, `GuestAddrs.lean` and the TSV all agree — zero occurrences. `_load_symmap` filters the TSV to `unit == 'stateless_guest'`, so a routine outside that image can never resolve, however its labels are bound.

So your model of the mechanism is not incomplete; mine was, when I first proposed `.globl` as the fix on #11522. Full write-up and the three options for route B are on that issue — the short version is that route B needs `account_encode` *added to the guest image*, which is a codegen/layout change rather than the conversion I described when you approved it.

⭐ And your closing line applies to me at least as much: *read the tool that consumes the artefact before reasoning about what changes it.* I proposed `.globl` without checking whether the routine was in the image the address table is generated from.

Issue #10609, opened 2026-07-26
gas: guest over-reports executed state gas by whole STORAGE_SET multiples (bv_fail=41, 50 rows) — accumulator not rolled back where spec refills


pirapira:
## Summary

The guest over-reports EIP-8037 **executed state gas** by whole multiples of
`STORAGE_SET` on 50 of the settled `bv_fail=41` rows. Verified against an
exactly-joined fixture with a hardware witness, not an argument.

**This is not a blob-gas defect.** The root's name (the representative is a
`test_blob_gas_subtraction_tx` parametrization) has already misled two
investigators. The mechanism is state-gas accounting on SSTORE.

## What is pinned

Fixture join is exact: the converter's numeric label prefix *is* its
enumeration index (sorted `rglob`, document order within each file, counting
only blocks carrying both stateless keys). Reproducing that enumeration offline
yields 26,104 blocks, matching the converter, the corpus and the manifest — so
index 5924 identifies exactly one block. Label truncation at 96 chars collapses
64 parametrizations of this test to the same string, so the *label* is not a
usable key; the index is.

Row 5924, `test_blob_gas_subtraction_tx[... --100_wei_mid_execution -- ... tx_value_0 ...]`,
tx type 0x03, 6 blobs. The target's 24-byte code is
`ORIGIN BALANCE PUSH1 0 SSTORE / <CALL draining the balance> / ORIGIN BALANCE PUSH1 1 SSTORE`.

| quantity | value |
|---|---|
| fixture header `gasUsed` | `0x017e80` = 97,920 = **one** `STORAGE_SET` |
| guest executed state gas | `0x02fd00` = 195,840 = **two** |
| target post storage | `{0x01: 0x64}` only — slot 0 absent, i.e. ends at zero |

Distribution: 50/50 rows over by a whole number of `STORAGE_SET`, 49 by one and
1 by two.

## The charge site, by commit-log trace

`scripts/spike/spike_run` honours `SPIKE_COMMITLOG`. Converting exactly row 5924
(`--skip 5924 --limit 1`) and streaming 5.4M trace lines through a FIFO into awk
(nothing written to disk), filtering on the accumulator address:

```
five writes of 0                    <- per-tx clears, runtime_dispatcher_call + 0x234
write 0x017e80  =  97920            <- FIRST charge
write 0x02fd00  = 195840            <- SECOND charge, SAME PC
[...]

PR #11672, opened 2026-08-07
feat(codegen): add inert executed state-gas differential

pirapira:
## Summary

- Preserve the incumbent per-transaction executed-state-gas accumulator and record an inert pool-derived differential beside it.
- Establish the universal message baseline after the callable pool reset and reservoir split.
- At the authorization seam, record preparation-only consumption in `runtime_tx_auth_state_used` and re-snapshot the reservoir so the differential measures executed state gas only.
- Apply the same auth fold and re-snapshot to the inline multi-transaction path, including authorization-phase OOG.
- Keep incumbent accumulator authority unchanged; the diagnostic arrays remain inert.

## Verification

All measurements below use the candidate at commit `bc6d8cfe65759ca0c9be04005a2471bd84da1639`, with ELF SHA-256 `a12dcdb3d765f465b910cf8a9ba0ba6df9cb4a51da131bd6e0e66d726ea5c331`.

The comparison baseline is `origin/main` commit `4ebe30732c1de26c80b6b19711e4bb9bfb7e4800`, with parent ELF SHA-256 `8d065b4900d9f3831076bf5621c7f39d7d9f493ce15f3e0361ed9f424a8b9f5f`.

The arrays compared are 76,184 bytes each, or 9,523 little-endian u64 transaction slots, over 614 manifest rows.

`readelf -SW gen-out/regionmap/stateless_guest.elf` predicted `.text = 0x55c08`, `.data = 0x53b0`, and `.bss = 0x190295c0`; `python3 scripts/gen-region-map-link-pins.py` regenerated the pins from that ELF, and the second pass `python3 scripts/gen-region-map-link-pins.py --check` reported `CLEAN`.

```text
lake build
Build completed successfully (4814 jobs).

python3 scripts/asm_to_program.py check-guest-addrs
check-guest-addrs: CLEAN (1021 symbols)

NO_BUILD=1 scripts/check-region-map.sh
check-region-map: region map matches the linked ELF

scripts/check-no-warnings.sh
check-no-warnings.sh: 0 warnings under EvmAsm/.
```

Axis 1, full extent:

```text
python3 /tmp/codex2-axis1-sweep.py --manifest /tmp/codex2-postrebase-614.manifest.tsv --spike scripts/spike/spike_run --candidate gen-out/regionmap/stateless_guest.elf --parent /tmp/codex2-parent-4ebe/gen-out/regionmap/stateless_guest.elf --out /tmp/codex2-postrebase-614-final2-axis1.tsv --workers 30 --timeout 120 --dump-range 0xaa19e7d0:76184 --legacy 0xaa19e7d0 --array-bytes 76184
fault_rows=0 axis1_mismatch_rows=0 axis1_mismatch_cells=0 axis1_mismatch_bytes=0
```

Axis 2, full extent:

```text
python3 /tmp/codex2-axis2-sweep.py --manifest /tmp/codex2-postrebase-614.manifest.tsv --spike scripts/spike/spike_run --elf gen-out/regionmap/stateless_guest.elf --out /tmp/codex2-postrebase-614-final2-axis2.tsv --workers 30 --timeout 120 --dump-range 0xaa19e7d0:76184,0xbc1395c0:228552 --legacy 0xaa19e7d0 --derived 0xbc1395c0 --diff 0xbc14bf58 --nonderivable 0xbc15e8f0 --array-bytes 76184
fault_rows=0 axis2_mismatch_rows=0 diff_nonzero_rows=0 nonderivable_nonzero_rows=0
```

Additional checks passed: `git diff --check`, `scripts/check-no-hardcoded-guest-pc.sh`, `scripts/check-layering.sh`, `scripts/check-forbidden-tactics.sh`, `scripts/check-heartbeats-approved.sh`, `scripts/check-unimported.sh`, `scripts/check-file-size.sh`, `scripts/check-opcode-structure.sh`, `scripts/check-progress.sh`, `scripts/check-drift.sh`, and `scripts/check-roundtrip-coverage.sh`.

`scripts/check-axioms.sh` was not run: this checkout has `LAKE_ARTIFACT_CACHE=true`, and that gate's `lake env lean` workflow cannot resolve cache-only oleans. This is an explicit skipped gate, not a passing result.

## Measurement history

An earlier sweep compared only one cell per row and reported zero. The corrected full-extent comparison found 43 mismatches, exposing the multi-transaction snapshot-ordering defect; moving the baseline after reset then exposed the authorization fold defect. The final universal-baseline plus auth-fold implementation resolves both at the full 9,523-slot extent above.

## Boundary behavior

Authorization state-gas consumption is now folded into `runtime_tx_auth_state_used` and excluded from the executed-state differential by re-snapshotting the reservoir at the auth seam. It is no longer misattributed and merely hidden by a status gate.


pirapira:
**coord review: `merge!` applied.** Reviewed against the rebased head with regenerated pins.

## What it is

An **inert** executed state-gas differential: the incumbent accumulator keeps authority, and a pool-derived value (`reservoir − left + spilled`) is computed alongside it into diagnostic arrays. No verdict path reads the derived value — verified: zero matches for it feeding a finalize or gate.

## Evidence, and why it is stronger than a verdict sweep

| gate | result |
|---|---|
| **axis 1** — candidate incumbent vs clean main, row by row | **0 mismatches / 614** |
| **axis 2** — derived vs incumbent, row by row | **0 mismatches / 614** |
| non-derivable rows | **0 / 614** |
| 614-row oracle sweep | FA=0, FR=17, OK=597, matching clean |
| r200 outputs | 200 vs 200, 0 byte mismatches |

⭐ Axis 1 is the one that matters and the one a verdict sweep cannot substitute for. Two builds can agree on every verdict while differing on an internal quantity that failed to cross a comparison threshold — and that is not hypothetical here: **axis 1 caught a real defect in this very change**, a stale `t3` flowing into a newly-shared merge tail that corrupted gas-left, skipped a coinbase row, left the block map short, made BALANCE return zero and made SSTORE skip a charge. Five symptoms, one register, invisible to verdicts.

## Layout

Pins regenerated **after** the final rebase onto `338723852`, with the values predicted first: `text 0x55ad8`, `data 0x53b0`, `bss 0x190695c0` — generated output matched the prediction, and a second regeneration pass is a **no-op** (all three artefact SHAs unchanged). `check-guest-addrs` clean at 1021 symbols, `check-guestaddrs-starts` 340/340, `check-region-map` passed.

## The one line that looked like a contradiction

The diff removes `la t1, evm_state_gas_used; sd t0, 0(t1)` from the simple-transfer publish path. That is a **replacement**, not a deletion — `directTransferStateGasChargeAsm` takes its place and still increments the accumulator while draining the pool. Axis 1 at zero is the evidence that the substitution is value-preserving on every row.

## What this change earned along the way

The differential was built to validate a migration motivated by #10609 — which turned out **stale** (main already reports the fixture-correct value; closed on that basis, with this work explicitly **not** credited for it). But the instrument found three real defects in unlanded code: a missing refill implementation at `ExceptionalHalt`/`Revert`, a missing spill increment that made the derivation **under**-report, and the stale-`t3` control-flow defect above. None would have appeared in a verdict sweep.

## ⚠️ Known boundary, recorded not fixed

The body states it correctly and I want it kept visible: on auth-phase OOG **after a completed authorization charge**, the pre-auth differential attributes auth consumption to the executed bucket where the spec books it to `auth_state_gas_used`. It is verdict-invisible **only** because `eip8037_prior_state_used_exact` adds the executed figure conditionally on `bv_tx_status_arr` being non-zero, and auth-OOG publishes status zero — so the wrong value is *excluded by a gate*, not correct by construction. Unexercised by the 614. Any future change making that read unconditional, or adding a second reader, unmasks it.

Expect this to sit `DIRTY` against #11673 — both regenerate the same three artefacts, so whichever lands first conflicts the other. That is the merge actor's to resolve.

pirapira:
Parking this one — `check-asm-to-program` reports two drifts, and both reproduce **standalone on this branch with no merge involved**, so neither is a batching artifact:

```
DRIFT DETECTED:
  [EvmAsm/Codegen/Dispatch.lean] dispatcherTxGasSettleFunction:
  [EvmAsm/Codegen/Programs/DispatcherExecStateGas.lean] dispatcherCaptureExecStateGasFunction:
    generated block not found verbatim (source drift)
```

They need different fixes, and only one is mine to make:

**1. `dispatcherCaptureExecStateGasFunction` — mechanical.** The emitted block is out of sync with its `.s` fixture. The documented `.s → Lean` regen fixes it:

```
python3 scripts/asm_to_program.py rewrite --file EvmAsm/Codegen/Programs/DispatcherExecStateGas.lean --funcs dispatcherCaptureExecStateGasFunction
```

**2. `dispatcherTxGasSettleFunction` — needs your call.** The gate says the symbolic render *would change the guest/probe images*, i.e. this is emitted-code substance rather than a layout regen. I am not going to regenerate that on your behalf, because doing so silently alters bytes the image is pinned against. Please confirm what the emitted form should be.

Everything else passes on the merged tree: `lake build` green (4814 jobs) and all other standing gates `ok`, so this is the only thing standing between it and landing.

Once both are addressed, drop the `complicated-merge` label and it will be picked up on the next pass.

pirapira:
**coord re-review — supersedes my earlier review comment on this PR, which cited evidence that has since been invalidated twice.** Re-derived from the artefact at head `bc6d8cfe6`, not from the review thread.

## Why a re-review was necessary

My first review quoted "axis 1 and axis 2 zero across 614". That figure was wrong twice over:

1. it was measured on a **working tree**, not on the pushed commits; and
2. the comparison read **one u64 cell per row** while each array is 76,184 bytes — **9,523 per-transaction slots**.

Re-read at full extent, the same artefact showed **43 mismatches**. Neither problem was visible in the number itself, which is the whole lesson: **a comparison must state its extent** — which rows, what is read per row, and why that is the whole of it.

## Current evidence

| gate | result |
|---|---|
| extent | 76,184 B = **9,523 u64 slots**, 614 manifests |
| axis 1 (candidate incumbent vs main) | **0** mismatching rows, **0** cells, **0** bytes; 0 faults |
| axis 2 (derived vs incumbent) | **0** mismatching rows, 0 diff-nonzero, **0** non-derivable |
| build | `lake build` 4814 jobs; `check-region-map`, `check-guest-addrs` clean |

Reporting rows, cells **and** bytes separately is what makes axis 1 checkable rather than a bare assertion.

## Layout

Base is current `main` `4ebe30732`. Pins regenerated from the relinked ELF with sizes predicted first (`text 0x55c08`, `data 0x53b0`, `bss 0x190295c0`); `gen-region-map-link-pins.py --check` second pass **clean**. Deltas against main reconcile: `.text` +0x460, `.bss` +0x20 — four 8-byte cells including the new auth quantity.

## The two defects this PR's own instrument found, after the instrument was corrected

- **Multi-transaction snapshot ordering.** The message baseline was captured *before* `emitRuntimeDispatcherCallableSetup` zeroed the pool cells, so from transaction 2 onward the baseline described the previous transaction. Transaction 1 worked because the cells were already zero — a first-iteration-works bug, invisible to a slot-0 comparison precisely on the transactions that expose it.
- **Auth misattribution.** With the baseline moved earlier, `AUTH_BASE` (35,190 = 23 B × 1530) landed in the executed bucket. Fixed by mirroring `vm/interpreter.py:362-364` — record the auth quantity at the seam and **re-snapshot** — rather than relocating the baseline, which would have restored the non-derivable rows. Early baseline and auth exclusion are not in tension in the spec, and are not here either now.

⭐ Both were reachable only because the corrected instrument could see per-transaction slots. Every verdict sweep throughout reported **FA=0, FR=17, OK=597** — identical to clean — while these were live.

## ⚠️ Gate that did not run

`check-axioms` was **skipped** (`LAKE_ARTIFACT_CACHE=true`), and the body says so. Almost certainly irrelevant for emitted assembly rather than proofs, but a skipped gate is not a passed gate and the reviewer should judge that rather than inherit an assumption.

## Scope

Accumulator authority is **unchanged**; the derived arrays feed no verdict path. The flip and the accumulator retirement remain PR 2, where the auth-bucket boundary note no longer applies — the fold is implemented, so auth is correctly attributed rather than misattributed-but-hidden behind the status gate.

`merge!` stands on this re-review.

Issue #11653, opened 2026-08-06
BAL attribution: transaction-level storage BAI attribution (blocks retiring the exec-log txindex)


pirapira:
Split out of #11104 (closed). Tracked separately because it is a **dependency for the serializer/storage migration**, not part of the account-field attribution work.

## Closing condition

Transaction-level storage attribution preserves `block_access_index` per `(address, slot)`, and the exec-log `txindex` can be retired without losing attribution.

## Why it is its own issue

The account-field siblings (#11649, #11650, #11652) can each land without this. This one gates the storage-side convergence instead — retiring the exec log's `txindex` requires a BAI-preserving replacement first, and that ordering is the whole point.

⛔ **This is not a reason to put a BAI field into the account row.** Storage history belongs with storage; see #11651 for why a folded row cannot carry a list.

## Constraint carried over

`destroy_storage` converts tx `storage_writes` into `storage_reads` before deleting them, same-transaction destruction only, *"so that accesses from created-then-destroyed accounts appear in the Block Access List"*. Any storage attribution structure must preserve that read evidence — a write-only map loses it.

## Evidence bar

Storage destroy-to-read vectors, and repeated same-slot writes across transactions with distinct BAIs. **r200 cannot prove storage attribution** — the per-BAI structure is invisible to a pass/fail sweep.

Siblings: #11648 (BAI channels) · #11649 (balance/nonce) · #11650 (code bytes) · #11651 (builder semantics) · #11652 (lifecycle).

pirapira:
Closing — **the closing condition is already satisfied on the live path.** Established by a per-producer, per-entry audit rather than a sweep, and the load-bearing claim verified independently against `origin/main`.

## Storage attribution already preserves BAI

The builder's storage row **carries the index in the key**. Row layout at `BlockAccessListBuilder.lean:214-230`:

```
storage_changes | addr |  0 | 20 | BE20
storage_changes | bai  | 24 |  8 | native LE u64
storage_changes | slot | 32 | 32 | BE32
```

and the record path dedups on `(address, slot, block_access_index)` — its own docstring notes that without that, the triple "would get two rows where the spec keeps one." So repeated same-slot writes in **distinct** transactions retain **distinct** BAIs, while same-transaction writes collapse to the final row, which is the spec's same-BAI final-write reduction.

The index reaches it through the two channels established in #11648, neither of which is exec-log metadata:

[...]

PR #11673, opened 2026-08-07
chore(codegen): retire exec_log_txindex tuple-validator chain (#11666)

### LINKED ISSUE #11666: cleanup: retire or document the unlinked exec_log_txindex tuple-validator chain
Split from #11653, which closed because live storage attribution already preserves BAI per `(address, slot, BAI)` — so what remains is not a migration but the removal of an **unlinked** validator chain and its side metadata.

## The cluster

Reads `exec_log_txindex`, and nothing in the live verdict reaches it:

| module | role |
|---|---|
| `ExecLogSlotTuples.exec_log_slot_tuples` (`:4-18`, `:60-68`, `:88-143`) | primitive reader, emits grouped `(BAI, final_value)` tuples |
| `SystemStorageSlotTuples.system_user_exec_log_slot_tuples` (`:19-99`) | begin-system / user / end-system windows |
| `AccountTupleSequencesConsistent` (`:4-22`, `:46-136`) | per-account comparison against BAL sequences |
| `BalAllAccountsTupleSequences.lean` (`:4-23`, `:40-100`) | all-account wrapper; its skip-list probe is the only route to the skip-list helper |

Side metadata: `bv_system_storage_txindex` (written as a literal zero at `BlockVerdictSystemStorageCapture:146`, consumed only by this chain), and `capture_system_storage_exec_rows`' BAI-stamp writer at `:44-72`, which has **no production caller** — only its own probe units at `:321/:330/:337/:346`.

Unlinked status is recorded in-tree at `BlockVerdictMtxTail:144-166` and `BlockVerdictV2:160-171` / `:505-513`, following #10646, #11245, #10681 and #11118.

## The decision this needs first

**Are the standalone tuple probes worth keeping?** They are the only remaining exerciser of a per-`(BAI, final_value)` reconstruction, which is a property the live hash path does not check directly. Two coherent answers:

1. **Retire the chain** — delete it as **one cluster**, since partial deletion leaves callers of deleted helpers. Side arrays go with it.
2. **Keep the probes** — then document at each entry that they are diagnostic-only and unreachable from the verdict, so the next reader does not mistake them for live gates.

⛔ Do **not** delete the side arrays before that decision. And do not repeat the shape this cluster already exhibits: unlinked machinery whose own comments say it is unlinked is exactly what #11613 and #11657 had to correct in prose this week.

## Constraints for whichever path is chosen

- The **destroy-to-read** conversion (`StorageWriteMap.destroyStorageFunction:591-655`, matching `state_tracker.py:560-579`) is an independent invariant and must survive untouched.
- `.text` should move only by what is deleted; predict before regenerating.
- Evidence if any live behaviour is touched: destroy-to-read vectors and repeated same-slot writes across transactions with distinct BAIs. **r200 cannot cover those structural classes** — a per-BAI structure is invisible to a pass/fail sweep.

pirapira:
Closes #11666.

## Summary
Retire the unlinked `exec_log_txindex` tuple-validator cluster in one cut. Live BAI path, `destroy_storage` write→read conversion, `append_modeled_system_storage_tuple_rows`, and `record_modeled_eip4788_storage_reads` stay.

## s3 lifetime (coord gate — checked before cut)
In `append_modeled_system_storage_tuple_rows` / `.Lamsr_append_one` the retired stamp was:
```
la s3, bv_system_storage_txindex
… sd zero, 0(s3) …   # stamp only
mv s4, a0; mv s5, a2; mv s3, a3   # IMMEDIATE redefine
```
**Every later read of s3** is after that redefine:
- `.Lamsr_value_rev`: `beq t0, s3` (vlen bound)
- `.Lamsr_value_rev`: `addi t1, s3, -1` (reverse index)

s3 is **not** dead after the store — it is redefined as **a3 = value length** before any consumer. Removing `la s3, bv_system_storage_txindex` leaves no stale-caller-s3 hazard. Cut is safe.

## Deleted
**Files:**
- `ExecLogSlotTuples.lean`, `SystemStorageSlotTuples.lean`, `SlotTupleSequencesMatch.lean`
- `BalSlotTupleSequence.lean`, `AccountTupleSequencesConsistent.lean`, `BalAllAccountsTupleSequences.lean`
- `BalValueReverseSAsm.lean`
- `scripts/asm-fixtures/slotTupleSequencesMatchFunction.s`

**Partial (same file as live helpers):**
- `capture_system_storage_exec_rows` probe + `bv_system_storage_txindex` stamp/BSS/params from `BlockVerdictSystemStorageCapture` / DataSection / Params
- RegistryTail imports + match arms; RegistryNamesTail names; BlockVerdict + StateRoot imports
- MANIFEST row for slotTupleSequencesMatch
- orphan `bsrMaxTuplesPerSlot` / `bvStorageLogTxindexEntryBytes` (tuple-only)

**Probes removed:**  
`zisk_exec_log_slot_tuples`, `zisk_system_user_exec_log_slot_tuples`, `zisk_slot_tuple_sequences_match`, `zisk_bal_slot_tuple_sequence`, `zisk_account_tuple_sequences_consistent`, `zisk_bal_all_accounts_tuple_sequences_consistent`, `zisk_bal_all_accounts_tuple_sequences_consistent_skip_list`, `zisk_capture_system_storage_exec_rows`

## Kept (must survive)
| symbol | status |
|---|---|
| `append_modeled_system_storage_tuple_rows` | LIVE |
| `record_modeled_eip4788_storage_reads` | LIVE |
| `destroy_storage` | LIVE |
| `bal_addr_to_exec_log_key` | LIVE |
| live BAI path | untouched |

Out of scope (separate issues): matches/covers/reads_in_exec_log, `exec_log_latest_value`.

## textSize
| | value |
|---|---|
| predict | UNCHANGED except stamp (−`la`/`sd` only); helpers were never in guest `.text` |
| base pin | `0x557bc` |
| measured | **`0x557a8`** (Δ **−0x14** = −20 B) |
| BSS | −262144 (`bv_system_storage_txindex`) |

Pins: `RegionMapLinkPins` + `symbol-addresses.tsv` + `GuestAddrs` fixpoint; `check-guestaddrs-starts` ok=340.

## Measure vs main `338723852`
| set | result |
|---|---|
| r200 seed=20260806 | **F→P=0 P→F=0** (same=200) |
| named 00178/05814/03736/11738/16431/19651 | PASS both sides |
| named 01114/11710/11617 | FR both sides (pre-existing; no flip) |

ELF base sha256 `dc57fdbfcc253a51…` cand `dffca475943ef685…`

## Comments
MtxTail code-42 note updated **unlinked → retired-#11666** (no ghost hunt).

## Do not
- Self-label merge!
- Expand to other unlinked clusters this PR


pirapira:
## Layout accounting (coord hold)

### NUMBER 1 — .text Δ −0x14 (= −20 B), not −12

Predicted “la+sd ≈ 12 B” was incomplete. The retired stamp is **five** instructions (objdump base `append_modeled`):

| bytes | insn |
|------:|------|
| 4 | `slli t0, s1, 3` (txindex entry index) |
| 4 | `auipc s3, …` |
| 4 | `addi s3, s3, …` (`la s3, bv_system_storage_txindex`) |
| 4 | `add s3, s3, t0` |
| 4 | `sd zero, 0(s3)` |
| **20** | **total** |

Function length: base 884 → cand 864 (Δ −20). Unified instr diff removes exactly those five; remaining diffs are branch/`auipc` immediates only. **Not alignment padding** — full stamp sequence.

### NUMBER 2 — BSS

**ELF did drop 262144.** Pin did not (my miss).

`readelf -SW`:

| | .bss Address | .bss Size |
|---|---|---|
| base | `0xa3110000` | **`0x190695a0`** |
| cand | `0xa3110000` | **`0x190295a0`** |
| Δ | — | **`−0x40000` = −262144** = `bvSystemStorageTxindexBytes` |

Section end / highest in-section markers:

| | `__BSS_END__` / `_end` |
|---|---|
| base | `0xbc1795a0` |
| cand | `0xbc1395a0` |
| Δ | `−0x40000` |

`bv_system_storage_txindex` @ base `0xaa2100e8` (in .bss); absent on cand.

Note: symbols at `0xbf980000+` (`ssz_merkleize_*`, `er_child_roots`, …) sit **above** `_end` on both builds (outside the NOBITS `.bss` size). Furthest **in-section** end is `__BSS_END__`, which moved by exactly the deleted arena.

**Pin fix:** follow-up commit sets `bssSizeBytes : Nat := 0x190295a0` (was left at base `0x190695a0`).

### destroy_storage
`StorageWriteMap` absent from file list — write→read conversion untouched (as you verified).

pirapira:
**coord review: `merge!` applied.** Both layout numbers account exactly, and one of my two challenges was my own error.

## Layout accounting

| quantity | before | after | delta | accounted |
|---|---|---|---|---|
| `textSizeBytes` | `0x557bc` | `0x557a8` | **−20** | the removed stamp is **five** instructions — `slli t0,s1,3; auipc s3; addi s3; add s3,s3,t0; sd zero,0(s3)` — verified by `objdump` on both ELFs; `append_modeled` body 884 → 864 |
| `bssSizeBytes` | `0x190695a0` | `0x190295a0` | **−0x40000** | exactly `bvSystemStorageTxindexBytes` = 262,144; `readelf -SW` confirms `__BSS_END__` `bc1795a0` → `bc1395a0` with the base unmoved at `a3110000` |

⚠️ I challenged the BSS figure claiming the pin was unchanged. **That was my error** — an incomplete grep of my own diff read. The pin does change, by exactly the predicted amount, and the prediction held. Recording it because the author was asked to account for a discrepancy that did not exist.

The `.text` figure was also mine to correct: I estimated the stamp at twelve bytes assuming `la` + `sd`. It is an indexed store — `slli`/`auipc`/`addi`/`add`/`sd` — so twenty is exact, not twenty-with-padding.

## The live-function edit was gated properly

The only change to executing code is removing the `bv_system_storage_txindex` stamp from `append_modeled_system_storage_tuple_rows`, which stays linked and is called from `MtxRuntime` and `StateRoot`. Register lifetime was established by naming instructions rather than asserting safety: the retired `la s3` was **immediately overwritten** by `mv s3, a3` — the sole live definition — and the only downstream reads are the two `vlen` uses at `.Lamsr_value_rev`. That check mattered: three separate register-lifetime defects surfaced elsewhere tonight, every one from code being **moved or removed** rather than added.

## Must-survive items verified

- `StorageWriteMap.destroyStorageFunction`'s write-to-read conversion — **untouched**; the file is absent from the diff entirely.
- `appendModeledSystemStorageTupleRowsFunction` and `recordModeledEip4788StorageReadsFunction` — **kept live** while `captureSystemStorageExecRowsFunction` is deleted from the same file. Partial-file edits are where cluster deletions usually go wrong; this line is drawn correctly.
- Live storage BAI attribution — untouched, per #11653.

## Scope discipline

The other already-unlinked clusters — `matches`/`covers`/`reads_in_exec_log`, `exec_log_latest_value` (#10681, #11118, #11178) — were deliberately left out rather than swept in. If they should go, that is a separate issue with its own enumeration.

And the `MtxTail`/`V2` comments move from "unlinked" to "retired #11666", so the next reader is not sent hunting a chain that no longer exists — the fifth stale-prose instance tonight and the cheapest to have prevented.

Closes #11666.

PR #11675, opened 2026-08-07
proof(#11347): check node arity exactly instead of probing item 2

### LINKED ISSUE #11347: proof: triple for `mpt_node_kind` against the node-shape dispatch of `_decode_witness_node`
One of the ten orienting rows from `docs/leaf-routine-targets.md` (row **4**), filed per-routine so each can be

## The pairing

| | |
|---|---|
| **RISC-V routine** | `mpt_node_kind` |
| **`SpecRef` target** | the node-shape dispatch of `_decode_witness_node` (`EvmAsm/Stateless/SpecRef/IncrementalMpt.lean:182`) |
| **caller** | `EvmAsm/Codegen/Programs/Mpt.lean:229` |
| **dependency status** | all callees verified — `rlp_list_nth_item`; 4 fixture in-edges |
| **data structures** | RLP node bytes |
| **separation-logic `Assertion`** | input: `bytesRegion`; result side: `mptNodeIs` (`EvmAsm/Evm64/MptAssertions.lean:635`) **exists** |

The 17-item branch / 2-item leaf-or-extension discrimination only, not the full decode.

Informally: for a `bytesRegion` holding a well-formed witness node, the returned kind matches the spec's branch/extension/leaf discrimination.

existing infrastructure:

semantics.

## Related

dhsorens:
Closes #11347. Implements **option 1** per your ruling — *"(1) sounds cleaner. In this project we can edit the code to ease proofs."*

## The divergence

`mpt_node_kind` used **"item 2 exists"** as a proxy for "17-item branch list". The spec's `decodeNodeItemAux` rejects any node list whose length isn't 17, so a **3–16 or 18+ item list was classified `branch` by the guest and `malformed` by the reference.** The routine now checks the real condition via `rlp_list_count_items` (already linked and `.proven` as `rlp_list_count_items_spec_within`): 17 → branch, 2 → leaf/extension, anything else → the existing fail arm.

## ⭐ No caller changes, which is why this is small

All three callers — `mpt_walk` (`Mpt.lean:633`), `mpt_insert_walk_db`, `mpt_insert_walk` — already dispatch on the returned `a0` with `j <parse_fail>` as the default arm, and `.Lmnk_fail` already returns `3`. I verified each **before** writing any asm rather than assuming the one call site covered it; that check is also what confirmed `mpt_walk` doesn't re-derive the kind independently.

## Mechanical conversion, not hand-edited

Edited the fixture and ran `asm_to_program.py rewrite`, which derives all four things an instruction-count change moves:

| moving part | what it derived |
|---|---|
| `GuestAddrs.mpt_node_kind + N` anchors | 28→24, 64→80, 72→88, 80→96, 88→104, 100→116 |
| relative branches crossing the edit | rewritten to symbolic `brOff` |
| instruction-indexed reloc table | 7→6, 16→20, 18→22, 20→24, 22→26, 25→29 |
| `_prog.length` pin | 49 → **53** |

Hand-editing those four is what produced #11518 (reversed `SD` operands) and #11519 (`laLo` anchored at the wrong PC).

## ⚠️ The data-symbol trap, worth recording

`mnk_item_count` had to go in **`ziskMptWalkDataSection`** (`Mpt.lean:1036`), *not* in `ziskMptNodeKindDataSection` — the section bearing the routine's own name. I tried the latter first, then `Dispatch.lean`. **Both built cleanly and regenerated a TSV with the symbol still ABSENT.**

The linked copy is the one whose emitted adjacency matches `stateless_guest.s`: `mnk_*` cells → `.balign 8` → `mbc_offset` → `mw_lookup_hash`. That's `mpt_walk`'s section, because the guest links `mpt_walk` and *its* data section carries what its callees need. I found it by matching emitted bytes back to source rather than trusting the name — the same class as #11522's finding that `account_encode`'s scratch can't reach the address table.

## Also: a gate that couldn't run, now can

`check-region-map.sh` probed only `readelf` / `riscv64-unknown-elf-readelf`, missing Homebrew's `riscv64-elf-readelf` — so it was **unrunnable on macOS**, while `gen-region-map-link-pins.py`, the generator it checks, already had the fallback. Pins could be regenerated locally with no way to verify them.

On its first run after the fix it immediately caught real drift: `textSizeBytes` short by `0x10` — exactly my four new instructions — because my pin regen had preceded the post-`rewrite` relink. **A gate that cannot run is indistinguishable from one that passes.** (#11043's class.)

## Verification

- `check-file mptNodeKindFunction` — CLEAN, byte-identical
- ⭐ `check-all` — CLEAN, **381 defs / 144 files**. The load-bearing one: regenerating `GuestAddrs` shifts every routine's `la` immediates
- `check-region-map.sh` — *"region map matches the linked ELF"*; second pin pass NO-OP
- `lake build Programs.Imports` + `Proofs.GuestImageEntries` — green
- `check-forbidden-tactics`, `check-registry-coverage` — OK
- layout chain repinned: TSV (3052), GuestAddrs (1020), RegionMapLinkPins, AxiomWitnesses (142), PROGRESS.md, DRIFT.md

## Behaviour change

Intended, and the point: a 3–16 or 18+ item node list now **fails** instead of being walked as a branch. The triple can now claim `agrees` against `items.length ≠ 17 → malformed`; I have not written that triple in this PR — this is the asm correction it needs to sit on.

⚠️ `mnk_dummy_offset`/`mnk_dummy_length` are now unused by this routine but left in place: `Dispatch.lean` and several other units also emit them, so removing them is a separate cleanup.


pirapira:

## The fix closes an FA-ward divergence

`mpt_node_kind` used *"item 2 exists"* as a proxy for "17-item branch list". The reference's `decodeNodeItemAux` rejects any node list whose length isn't 17, so a **3–16 or 18+ item list was classified `branch` by the guest and `malformed` by the spec**.

That direction matters: the guest was **accepting** a shape the reference rejects. Checking the real condition via `rlp_list_count_items` — 17 → branch, 2 → leaf/extension, else the existing fail arm — turns an accept into a reject, which is the correct direction and the reason this is worth landing rather than deferring.

Verified independently: `rlp_list_count_items_spec_within` is registered and witnessed (`AxiomWitnesses.lean`, `Correspondence.lean`, `Routines.lean` ×2), so the dependency is a proven routine rather than an assumed one.

## No caller changes, and that was checked rather than assumed

All three callers — `mpt_walk` (`Mpt.lean:633`), `mpt_insert_walk_db`, `mpt_insert_walk` — already dispatch on the returned `a0` with `j <parse_fail>` as the default arm, and `.Lmnk_fail` already returns `3`. Checking each **before** writing any assembly is what makes "no caller changes" a finding instead of a hope, and it also established that `mpt_walk` does not re-derive the kind independently — which is the failure mode that would have made a local fix insufficient.

## Mechanical conversion

The fixture was edited and `asm_to_program.py rewrite` run, so the four things an instruction-count change moves were **derived, not hand-adjusted**: the `GuestAddrs.mpt_node_kind + N` anchors, the relative branches crossing the edit (rewritten to symbolic `brOff`), and the instruction-indexed reloc table. Hand conversion is what produced #11518 (reversed SD operands) and #11519 (`laLo` anchored at the wrong PC), so using the converter is the point rather than a convenience.

`.text` +0x10 — four instructions, consistent with replacing a probe with a count call.

## ⭐ The `check-region-map.sh` change is a real find, not incidental

Homebrew's `riscv64-elf-binutils` installs `riscv64-elf-readelf`. Omitting that spelling left the gate **unrunnable on macOS** while `gen-region-map-link-pins.py` had already been fixed for the same gap — so pins could be regenerated locally by a developer whose checker silently could not run.

A gate that cannot run is not a gate, and an asymmetry between a generator and the check that validates its output is worse than either being absent, because the generator's success reads as validation. Fixing both spellings together is the right treatment.

## Layout note

The `-0x557bc` baseline predates #11673 and #11672, so the pins here describe a `main` that has since moved twice; expect this to sit `DIRTY` until the merge actor repins. That is derived-artefact churn and the actor's to absorb, not the author's to chase.

pirapira:
Parking this one — it fails `check-build-units-link`, and the failure reproduces **standalone on this branch with no merge involved**. I checked pristine `main` first, where the same gate is `OK`, so this is not a batching artifact:

```
FAIL     zisk_stateless_verdict — no .elf produced
           undefined reference to `rlp_list_count_items'
FAIL     zisk_step2_verdict — no .elf produced
           undefined reference to `rlp_list_count_items'
```

Cause looks like the intended change itself: checking node arity exactly means `mpt_node_kind` now calls `rlp_list_count_items` where it previously probed item 2. The two standalone verdict units mirror guest handlers, and per the gate "a unit that mirrors guest handlers must also define every routine those handlers call" — so they now need `rlp_list_count_items` in their closure (note `zisk_stateless_verdict_v2` still links fine, so it already has it).

Which unit definitions to extend is your call, so I have not guessed at it.

Everything else on the merged tree is green: `lake build` (4814 jobs), `check-asm-to-program`, `check-region-map`, `check-guestaddrs-starts`, `check-routine-liveness`, reports, and the rest.

One note for when you rebase: the generated files need resolving toward **this PR**, not `main` — the arity change grows `mpt_node_kind`, so main's extents overrun (`DECLARED_EXTENT_OVERRUN entry=mpt_node_kind`) and regeneration cannot recover because it needs a working build first. I handled that on my side; mentioning it only so the same thing does not surprise you.

Once the link closure is fixed, drop the `complicated-merge` label and it will be picked up on the next pass.

Issue #11578, opened 2026-08-05
proof: triples for derive_withdrawal_requests / derive_consolidation_requests against the request-derive half of encode_execution_requests


pirapira:
> 🤖 Generated by Claude Code (strategist pass over execution-specs ↔ SpecRef ↔ RISC-V correspondence, 2026-08-06).

## The move

Second tier of `docs/leaf-routine-targets.md`'s runners-up, recorded there as startable:

> `derive_withdrawal_requests`, `derive_consolidation_requests` (leaves): the request-derive half of `SpecRef/SeamShell.lean`'s `encode_execution_requests`. Startable, second tier.

Both are leaves in the transitive call graph (no unproven premises), and their SpecRef anchor is the request-derive half of `encode_execution_requests` / the EIP-7002/7251 system-call output parsing in `SpecRef/Fork.lean::process_general_purpose_requests`. Fixed-stride byte extraction from a system-call output buffer — CALLDATACOPY-class loop proofs, a solved shape.

## Why this pair (and not another second-tier pick)

The requests path is live FR/FA surface: #11254 (7002-class multi-BAI capture), #11432 (deposit route B appends a request for a REVERTED call). A stated triple for the derive step pins down what the guest extracts from the predeploy output, which is exactly the boundary those bugs straddle — the proofs won't fix them, but they turn "what does the derive step actually produce" from re-derivation into a citation.

## Ask

- One `cpsTripleWithin` spec per routine, against the corresponding derive step at pin `e5a8caf1b` (`requests.py` withdrawal/consolidation stride constants — cite the Python, since SpecRef's `SeamShell` phrasing may fold them).
- Assertion side: `bytesRegion` over source (system-call output window) and destination should suffice, as for the copy-family opcodes; if the destination is a typed request arena, say so and define the minimal run predicate rather than widening scope.
- Registry rows on landing (family: probably `guest` or a new `requests` family — grader's choice, but grade them).

dhsorens:
@pirapira — I cannot satisfy this issue as written, and I would rather say so with the measurement than quietly deliver something adjacent. **I wrote this issue**, so the error is mine.

## ⚠️ The premise is wrong: these routines are not byte-extraction loops

The issue says *"Fixed-stride byte extraction from a system-call output buffer — CALLDATACOPY-class loop proofs, a solved shape"* and that `bytesRegion` over source and destination should suffice.

[...]

Issue #11574, opened 2026-08-05
predicates: assertion vocabulary for the precompile field towers (BLS12-381/BN254 limb buffers, point records) — the named prerequisite that cut blsg_lt_p/bnf_lt_p from the leaf table


pirapira:
> 🤖 Generated by Claude Code (strategist pass over execution-specs ↔ SpecRef ↔ RISC-V correspondence, 2026-08-06).

## The gap

`docs/leaf-routine-targets.md` (runners-up) records why the two cleanest untargeted leaves were cut from the table of ten:

> `blsg_lt_p`, `bnf_lt_p` (leaves): the `< p` range check inside `bytes_to_bls_field` / `bytes_to_g1` … Real SpecRef counterparts, but **the precompile field towers have no assertion vocabulary at all yet** — a bigger prerequisite than any single row above.

SpecRef's precompile side is fully ported (`SpecRef/Precompiles{Curve,Bls,BlsMap,Kzg,Pairing}.lean`), and the guest's BLS/BN254 routines are live verdict-path code — #11542's FR census and the #10636 point-validation gap show the crypto path is both exercised and under-verified. Zero predicates exist for any of: field-element limb buffers (Fp/Fp2/Fq), affine/projective point records, MSM scratch.

## Ask (prerequisite only, deliberately minimal)

1. Define the base vocabulary: `fpLimbsIs (base : Word) (v : Nat) : Assertion` — a fixed-width little-endian limb buffer encoding a natural `< p` — parameterized or duplicated per modulus (BLS12-381 p at 6×64, BN254 p at 4×64), matching the layout the emitted `blsg_*`/`bnf_*` routines actually use (read it out of the asm fixtures, don't assume).
2. On top of it, the point-record predicates `g1AffineIs` / `bnPointIs` as pairs of field cells plus the on-curve side condition **as a stated (not yet proven) well-formedness proposition**.
3. First consumers, to keep the vocabulary honest: state (not necessarily prove) the triples for the two cut leaves — `blsg_lt_p` and `bnf_lt_p` against the `< p` checks in `SpecRef/PrecompilesKzg.lean` / `SpecRef/PrecompilesCurve.lean`.

## Why now

This is the third leg of predicate coverage (state containers and code DB being the others): every future crypto-path proof — including making #10636's fix provable rather than testable — bottoms out here, and the leaf-targets doc already identified the entry points 3 days ago.

[...]

Issue #11018, opened 2026-07-31
BAL sort: per-container runtime bound may be needed for behavioural equivalence (static sizing is deliberate), plus a List-vs-Dict/Set representation gap affecting future uniqueness proofs


pirapira:
Raised by **#11016**'s BAL correspondence audit and split out so it reaches whoever eventually proves properties about the sort, rather than living in a PR description.

## Static sizing is deliberate; the behavioural question is separate

`BalCanonicalSort.lean:218` uses `blockAccountWritesCapacity` for **every** container, while the live path's arrays have different capacities. **Per the maintainer this is not a bug** — it is a static-allocation decision under tight memory, and one oversized bound for all containers is the intended shape.

⚠️ **But behavioural equivalence with execution-specs may still require a per-container runtime bound.** The static array size governs what the guest *can* hold; the spec's semantics govern what a block may *validly* contain. Those are different limits, and using one capacity for all only settles the first. If a container's logical limit is smaller than the shared allocation, the guest could process or accept entries a spec-conformant implementation would not — a divergence in the accept direction rather than a memory fault.

**What would settle it:** for each container the sort handles, what bounds the element count at runtime — a per-container count, or the shared capacity? If the shared capacity, does any spec-side limit constrain that container more tightly? Note the derived-capacity discipline here: arena sizes come from the block-gas derivation (`240M/cost + 6 × 30M/cost`), not from what the corpus exercises, so "no fixture overflows it" is not an answer.

## Representation gap in the model — a caveat for future proofs

Also from #11016: the Lean model uses `List` where the reference uses `Dict`/`Set`, so **the model can represent duplicate slots and duplicate reads that the reference cannot express at all.** #11016's corpus deliberately stays inside the representable domain, so its 1149/1149 agreement is sound — but it does not and cannot speak to the surplus domain.

⚠️ **This matters specifically for anyone proving uniqueness or canonicity against the Lean type**, e.g. #10817: a uniqueness obligation stated over `List` carries cases the reference has no opinion on, so a proof may be discharging conditions the spec never imposes — or, worse, be unprovable for inputs the spec cannot construct.

## Context

#11016 establishes that the *ordering* the model implements matches the reference (account order byte-lexicographic, slot order, per-index orders, read/write exclusion), while explicitly leaving open **whether the assembly implements that ordering.** Both items above sit on the assembly side of that line, which is why neither is a defect in what #11016 verified.

pirapira:
[...]

PR #11676, opened 2026-08-07
feat(#10817): balEntriesFrom row-array predicate, the strides as data, and a stale-claim sweep

dhsorens:
Stage 1 of #10817's semantic half. **Proves nothing yet** — it makes the headline obligation *statable*.

## Why this first

`docs/leaf-routine-targets.md:49` named this as the gating prerequisite, and it is: **you cannot state "the output is a permutation of the input" without a `List`-indexed assertion over the row array** — only its sortedness. And permutation is the obligation that matters most here, because the end-to-end hash comparison **structurally cannot see it**: it compares against a model built from the *declared* rows, so it cannot see rows the guest never had. (@pirapira's 362,880-permutation experiment already exonerates account *ordering* on the fixture it covers, which is what makes permutation the higher-value half.)

## The predicate

`balEntriesFrom` / `balBuffer` / `balOwn` + `balRowsWf`, mirroring `teerEntriesFrom` / `teerBuffer` / `teerOwn` with the same `pcFree` and cons lemmas.

⚠️ **Two deliberate departures from the teer shape, both forced by the live path:**

1. **Stride is a parameter.** The teer table has one entry width; the sort is called at **six** live sites with **four** distinct strides (96/64/40/24). #10817 asks explicitly to *"parameterise over the descriptors"* rather than over literal byte offsets, so a row-content change cannot silently invalidate a proof.
2. **Rows are opaque bytes, not a typed record with a `.render`.** A typed row would bake one call site's segment layout into the predicate, and the six disagree — including one whose slot segment is **little-endian** where every other is big-endian. The sort moves whole rows and never interprets them, so opaque bytes is the faithful vocabulary; the *key* is where interpretation belongs, and that belongs to the spec-side model (`_build_from_builder`), not to a resource predicate.

## The strides as guarded data

`balSortCallSites` carries all six sites (array, stride, descriptor, segment count) read off `BalSerializer.lean:1071-1108`, with `by decide` guards:

- `_count = 6` — matches the `#guard` in `BlockAccessListBuilder.lean`;
- `_strides_aligned` — every stride `% 8 = 0`. **Not cosmetic**: the sort swaps rows with `ld`/`sd`, and the probe found the empirical version (a stride-20 arena puts row 1 at `base+20` and faults). Any triple carries this as a hypothesis;
- `_stride_count = 4`;
- `_one_le_segment` — exactly one LE segment. A canonical key assuming BE throughout would be wrong on precisely that one, **with no local symptom** — the failure mode the module header warns about.

⭐ **`_stride_count` earned its place on its first run**: I wrote 5 and `decide` refuted it. `64` is shared by the storage reads and the balance and code changes.

⚠️ Two row kinds the issue's own four-row table omits, and both are the unusual ones: **storage reads** (stride 64, descriptor `0x2020`, the only LE segment) and the **accounts** array (stride 24). A proof covering only the issue's four covers 4 of 6 live sites.

## Stale-claim sweep — five places falsified by #11046/#11054

- **`Correspondence.lean`**'s `bal_canonical_sort` note said *"String-only; no `: Program`, so no `cpsTripleWithin` is statable"*. False since #11046. The row stays `.unproven` because nobody has **stated** one, not because nobody **can** — and the note now says which, that the remaining gap is a *predicate* gap, and that slots sort **numerically** (the reference docstring says "lexicographically" and is wrong).
- **`docs/bal-spec-correspondence.md`** (×2) — "blocked on a ~230-instruction conversion" landed as #11046.
- **`EvmAsm/Tests/Correspondence/Bal.lean`** — same claim; the surviving point (a model-boundary differential needs no `Program`) is kept and sharpened.
- **`BalSelftestsProbe.lean`** — its exclusion comment named two routines #11054 deleted, so there was nothing left to exclude.
- **`BalCanonicalSort.lean`**'s stride stub claimed *"Both containers use stride 128"* — **no live call site passes 128**.

## Doc corrections @pirapira asked for on #11578

Same file, so folded in here rather than conflicting across two PRs:

- row 5's prerequisite column: **NONE → LANDED**;
- the `derive_*_requests` runner-up: the **"(leaves)" annotation was wrong and #11578 was filed on it** — seven instructions tail-jumping into the whole EVM, no memory access, and the strides are not spec constants at the pin. Retarget to `execution_requests_hash` recorded as endorsed;
- the `blsg_lt_p`/`bnf_lt_p` runner-up: it named the wrong prime (`BLS_MODULUS` is the scalar order; the routine checks the base field `p` via `bytes_to_fq`), and "no assertion vocabulary at all yet" overstates it — **both triples already exist, `sorry`-free**.

## Verification

`lake build` clean (4814 jobs), no `sorry`. `check-layering` (OK — the predicate is Codegen-side and consumed by a Codegen spec, so L1 holds; note this is *why* #11574's crypto predicate must go in core instead), `check-unimported` (2844/2844, 0 orphans), `check-registry-crosscheck` (21 symbols × 29 rows), `check-progress`, `check-file-size`, `check-forbidden-tactics`. `lake exe correspondence-check bal` still agrees on every record.

No new theorems are witnessed, so `AxiomWitnesses` / `check-axioms` are unchanged. ⚠️ `check-asm-to-program.sh` not run — no emitted string changed here, so byte identity is untouched.

## Next on #10817

Fixture cases through the probe-only selftest, then the permutation proof itself using `perm_range_of_permCheck` (`SepLogic.lean:3093`) — an already-proven decidable certificate, so no new permutation framework. Two questions on the issue still awaiting your call: the framing narrowing (open since 30 July) and whether global uniqueness is wanted given #11018's representation gap.


pirapira:
**coord review: `merge!` applied**, with one correction to make in a follow-up push if you can catch it before the actor takes this, and one placement question answered rather than raised.

## ⚠️ A count that disagrees with itself, inside the PR whose purpose is preventing exactly that

The body says *"**four** distinct strides (96/64/40/24)"*. The added docstring says *"with **five** distinct strides."* A third comment says the table *"carries all **six**"* call sites, which is the site count and is correct.

I measured the six live sites at `BalSerializer.lean:1071-1108` — `a2` = **96, 64, 64, 40, 64, 24** — so: **six call sites, four distinct strides.** The body is right; the docstring is wrong.

Worth fixing precisely because of this PR's own argument: the stride figures are being moved from prose into `balSortCallSites` *"so a stride change cannot leave a comment quietly wrong."* A prose count that is already wrong on arrival is the strongest possible case for that change and the weakest possible advertisement for it. `balSortCallSites_stride_count` should pin the number the prose quotes.

## Placement is right, and it does not contradict your #11574 argument

You argued on #11574 that predicates belong in **core**, not `Codegen/RegionPredicates.lean`, because *"L1 makes `Codegen` a pure sink."* Putting `balEntriesFrom` in `Codegen/RegionPredicates.lean` looks like the opposite call, so I checked whether it is:

- **19** files under `EvmAsm/Codegen/**` import `EvmAsm.Stateless.SpecRef`;
- **0** files under `EvmAsm/Stateless/**` or `EvmAsm/EL/**` import `EvmAsm.Codegen`.

So the sink claim holds, and the two placements are consistent because the *directions* differ. #10817's obligation is a statement about the **emitted program**, so its theorem lives Codegen-side and can freely cite SpecRef — 19 files already do. #11574's crypto obligation needed a **core-side** bridge, which cannot see Codegen. Same rule, opposite consequence. Recording it here so a reviewer does not read the pair as inconsistent.

## The prerequisite argument is correct and worth restating

`docs/leaf-routine-targets.md:49` named this as gating, and it is: without a `List`-indexed assertion over the row array you can state sortedness but **not** permutation. And permutation is the half that matters most here, because the end-to-end hash comparison **structurally cannot see it** — it compares against a model built from the *declared* rows, so a row the guest never had is invisible to it. That is the gap the maintainer's 362,880-permutation experiment does not close either, since it exonerates account *ordering*.

Both departures from the `teer` shape are forced rather than chosen: stride as a parameter because there are four of them across six sites, and opaque byte rows because a typed row with a `.render` would bake in the layout the proof is supposed to range over.

## Not proven yet, and the PR says so

This makes the obligation **statable**; it proves nothing. That is the right first slice, and stating it plainly is better than implying coverage. The 8-aligned precondition being carried as data (`balSortCallSites_strides_aligned`) is the right treatment of a hypothesis that would otherwise live only in a comment — `BlockAccessListBuilder.lean:44`'s all-caps warning and the probe's stride-20 fault are the prose and empirical versions of the same fact.

dhsorens:
Caught it — pushed `764c5b369` before the actor took it.

**Six sites, four distinct strides.** `a2` = 96, 64, 64, 40, 64, 24, so `64` is shared by the storage reads and the balance and code changes. The body and `balSortCallSites_stride_count` both said four; only the docstring said five. I've spelled the enumeration out inline so the arithmetic is checkable in place rather than trusted.

Your framing of why it mattered is right, and worse than you knew:

> A prose count that is already wrong on arrival is the strongest possible case for that change and the weakest possible advertisement for it.

`balSortCallSites_stride_count` had **already refuted this exact number once** — I first wrote the *theorem* as 5 and `decide` rejected it. I fixed the theorem and left the prose. So the guard caught the same error twice and I only listened the first time, which is a tidier demonstration of the PR's own argument than anything I wrote in it. The docstring now says so, since the next person deserves the example more than I deserve the omission.

**On the placement question** — thank you for answering it rather than raising it, and for measuring it: 19 files under `Codegen/**` import `SpecRef`, 0 under `Stateless/**` or `EL/**` import `Codegen`. Your summary is better than my reasoning was:

> Same rule, opposite consequence.

I had the two conclusions right but was carrying them as separate judgements rather than as one rule applied in two directions. #10817's obligation is about the **emitted program**, so it lives Codegen-side and may cite SpecRef freely; #11574's needs a **core-side** bridge, which cannot see Codegen. I'll use that phrasing when the #11574 predicate lands, so the pair reads as consistent by construction instead of needing a reviewer to reconstruct it.

#11677 is up with #11574's first piece — the `beBytesToNat ⟷ fromBytesBE` bridge, `propext` only.

PR #11679, opened 2026-08-07
feat(#11574): the field-element and point-record assertion vocabulary

dhsorens:
Second piece of #11574. **Stacked on #11677** (the decoder bridge). Per the ruling: BE first, LE alongside as the companion rather than the prerequisite.

## ⚠️ Why BE leads

The issue asks for `fpLimbsIs`, *"a fixed-width little-endian limb buffer"*, and names `blsg_lt_p` / `bnf_lt_p` as its first consumers. Those two are inconsistent — the fixture is a byte-at-a-time MSB-first `lbu`/`bltu` scan against `blsg_p_be`, with no `ld`, no stride and no alignment requirement, and both `lt_p` triples state their post over `beBytesToNat`. The LE 6×64 / 4×64 layout is real, but its consumers are the **converters** (`blsg_be_to_le`) and the accelerator point forms.

Both are here. The module header says which is which, so the issue's own error isn't reproduced inside its fix — that was your point about framing LE as following the bridge rather than preceding it.

## What's here

- **`fpBeBytesIs width p base v`** — the form the live consumers use. Shape follows `accountRlpIs` (`State/AccountAssertions.lean:118`): a value-carrying `…Is` whose well-formedness sits beside the `bytesRegion` rather than being left to the caller. `…Is` passes `RegionPredicates`' naming test, since a fixed-width element pins the *whole* region. Specialised as `blsFpBeIs` (48, `blsP`) and `bnFpBeIs` (32, `fieldModulus`).
- **`fpLimbsIs nl p base v`** over the **existing** `leBytesN` / `wsNat` (`Rv64/SAsm/AccelStep.lean:635`, `:623`) rather than a fresh limb decoder — `nl = 6` is already documented there as the BLS12-381 view.
- **`g1AffineIs` / `bnPointIs`** over the compact BE layouts, with on-curve **stated** as a `Prop` per item 2.

⚠️ **Contents types kept distinct**, as you asked: BLS G1 is projective with `z = 0` meaning infinity (`PrecompilesBls.lean:98`); BN254 is `Weier.Pt = Option (Nat × Nat)` with `none` meaning infinity (`PrecompilesCurve.lean:35`). Forcing a shared type would only move the case split somewhere less honest.

## Placement — and why it does not contradict #10817

The module is in **core**, not `Codegen/RegionPredicates.lean`, because a crypto predicate must be visible to a core-side bridge against `bytes_to_fq`, and L1 makes `Codegen` a pure sink.

I've recorded your framing from #11676 in the header rather than leaving a reviewer to reconstruct it:

> Same rule, opposite consequence.

#10817's obligation is about the **emitted program**, so its predicate lives Codegen-side and may cite SpecRef freely (19 files under `Codegen/**` already do); this one must be visible from core (0 files under `Stateless/**` or `EL/**` import `Codegen`). The pair is consistent by construction.

## Satisfiability witnesses — required, not decorative

`g1AffineIs` bundles an on-curve condition, which is exactly the shape that can be accidentally unsatisfiable — the #10688 lesson `RegionPredicates.lean:518-522` records. So:

- the BN254 generator `(1, 2)` is **checked** on-curve (`2² = 4 = 1³ + 3`);
- `(0, 0)` is **checked not** on-curve. The port handles infinity by a separate `x == 0 && y == 0` branch (`PrecompilesCurve.lean:91`), so folding it into the predicate would make the predicate accept a non-point. Pinned so nobody "simplifies" it later.

`blsP_ne_blsModulus` pins the correction this issue needed — the base field prime and the scalar order really are different numbers. It depends on **no axioms at all**.

## Verification

`lake build` clean (4735 jobs), no `sorry`. `check-layering` (core→core only), `check-unimported` (2765/2765, 0 orphans), `check-forbidden-tactics`, `check-file-size`.

```
'EvmAsm.Stateless.Crypto.pcFree_fpBeBytesIs' depends on axioms: [propext, Quot.sound]
'EvmAsm.Stateless.Crypto.pcFree_fpLimbsIs'   depends on axioms: [propext, Quot.sound]
'EvmAsm.Stateless.Crypto.blsP_ne_blsModulus' does not depend on any axioms
```

No registry rows yet — these are predicates, not routine specs, so `Routines`/`Correspondence`/`AxiomWitnesses` are untouched.

## Next

The 16-byte EIP-2537 wire-pad relation **in the statement**, then the bridges against `bytes_to_fq` / `bytes_to_g1` restating `blsgLtP_spec` / `bnfLtP_spec` in this vocabulary, then the two absent registry rows and the `4ch8f-crypto-kernel-inventory.md:53` correction.

### DIFF
--- a/EvmAsm.lean
+++ b/EvmAsm.lean
@@ -1,5 +1,6 @@
 import EvmAsm.Rv64
 import EvmAsm.Crypto.BeBytesBridge
+import EvmAsm.Stateless.Crypto.FieldAssertions
 import EvmAsm.Crypto.Fermat
 import EvmAsm.Crypto.PowLadder
 import EvmAsm.Crypto.ScalarMul
--- a/EvmAsm/Crypto/BeBytesBridge.lean
+++ b/EvmAsm/Crypto/BeBytesBridge.lean
@@ -21,9 +21,18 @@
   stand on** without this equality: the two sides would be talking about
   syntactically different functions that a reader assumes are the same.
 
-  `U256MinSAsm.beBytesToNat_foldl` proves the accumulator generalisation, but it
-  is `private`, lives under `Codegen`, and stops short of the equality — so it is
-  unavailable to a core-side bridge on both counts.
+  The accumulator generalisation already exists twice under `Codegen` —
+  `U256MinSAsm.beBytesToNat_foldl` (`private`) and
+  `U256MulU64Be/Basic.beBytesToNat_foldl_init` (public) — and **neither is usable
+  here, for one reason: layering.** `Codegen` is a pure sink, so a core-side
+  bridge cannot import either, whatever their visibility. Both also stop short of
+  the equality, generalising within `beBytesToNat` rather than over `fromBytesBE`.
+
+  ⚠️ An earlier draft of this docstring gave "it is `private`" as half the
+  justification. That is wrong about the public one and irrelevant to the private
+  one — visibility never entered into it. Recorded rather than silently corrected,
+  because "this is unavailable" is exactly the kind of claim that gets repeated
+  from a docstring without being re-measured.
 -/
 
 import EvmAsm.Crypto.PowLadder
--- a/EvmAsm/Stateless/Crypto/FieldAssertions.lean
+++ b/EvmAsm/Stateless/Crypto/FieldAssertions.lean
@@ -1,0 +1,166 @@
+/-
+  EvmAsm.Stateless.Crypto.FieldAssertions
+
+  Assertion vocabulary for the precompile field towers (GH #11574).
+
+  ## ⚠️ Read this before reaching for the LE form
+
+  #11574 asks for `fpLimbsIs`, *"a fixed-width little-endian limb buffer"*, and
+  names `blsg_lt_p` / `bnf_lt_p` as its first consumers. **Those two things are
+  inconsistent**, which is why this module leads with the big-endian form:
+  `scripts/asm-fixtures/bls12G1LtPFunction.s` is a byte-at-a-time MSB-first
+  `lbu`/`bltu` scan against `blsg_p_be` — no `ld`, no 8-byte stride, no alignment
+  requirement. Both `lt_p` triples state their post over `beBytesToNat`.
+
+  The LE 6×64 / 4×64 layout is real, but its consumers are the *converters*
+  (`blsg_be_to_le` writes limb `i` at `base + 8i`, LSB first) and the accelerator
+  point forms. So `fpLimbsIs` is here too, and is genuinely useful — but as the
+  companion to the BE form rather than as the prerequisite it was filed as.
+
+  ## Why this is core and not `Codegen/RegionPredicates.lean`
+
+  `check-layering` L1 makes `EvmAsm/Codegen` a **pure sink**: 19 files under
+  `Codegen/**` import `SpecRef`, and 0 files under `Stateless/**` or `EL/**`
+  import `Codegen`. A crypto field predicate has to be consumable by a
+  **core-side** bridge against `bytes_to_fq`, so it cannot live under `Codegen`.
+
+  ⭐ The same rule sends `balEntriesFrom` (#10817) the other way, and the pair is
+  consistent rather than contradictory: *same rule, opposite consequence*.
+  #10817's obligation is about the **emitted program**, so its predicate lives
+  Codegen-side and may cite SpecRef freely; this one must be visible from core.
+
+  ## The primes are the ones the routines actually compare against
+
+  ⚠️ `blsg_lt_p` compares against `blsg_p_be`, the **base field** prime
+  (`Bls12.blsP`, 381-bit). #11574 and `docs/leaf-routine-targets.md` both paired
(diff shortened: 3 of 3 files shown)


pirapira:
**coord review: `merge!` applied.** Verified against `origin/main` at `96c3eb50d`. Stacked on #11677, which is labelled.

## The prime correction is right, and it is the most valuable thing here

I checked it independently rather than taking the docstring:

- `scripts/asm-fixtures/bls12G1LtPFunction.s:1-2` — `blsg_lt_p:` / `la t0, blsg_p_be`. The routine compares against **`blsg_p_be`**.
- `SpecRef/PrecompilesBls.lean:37` — `blsP = 0x1a0111ea…aaab`, the **381-bit base field** prime.
- `SpecRef/PrecompilesKzg.lean:34` — `BLS_MODULUS` is a **separate 255-bit** constant, the scalar field order, used by `bytes_to_bls_field` / `kzgProofError`.

So #11574 and `docs/leaf-routine-targets.md` did pair `blsg_lt_p` with the wrong prime, and a bridge built on that pairing would have been stated against a modulus the routine never reads. ⚠️ **That is worth propagating back into the doc and the issue** — this PR fixes it locally but the two sources that misstate it are still standing.

## The BE-leads ruling is correctly implemented

The issue asks for `fpLimbsIs` as *"a fixed-width little-endian limb buffer"* and names two consumers whose fixture is a byte-at-a-time MSB-first `lbu`/`bltu` scan with no `ld` and no stride. Those are inconsistent, and this module says so in its own header rather than reproducing the error inside the fix. Both forms ship; `fpLimbsIs` reuses the existing `leBytesN`/`wsNat` at `AccelStep.lean:635`/`:623` instead of introducing a third decoder, which is the same instinct as #11677.

## Layering

The core-vs-`Codegen` argument is sound and the #10817 contrast is the part I want on the record: **same rule, opposite consequence.** `balEntriesFrom`'s obligation is about the emitted program so it sits Codegen-side and may cite SpecRef freely; a crypto field predicate must be visible from a core-side bridge against `bytes_to_fq`, so it cannot. Two placements that look contradictory until you ask which side consumes them.

Additive only — two files, `+167/-0`, one import line, no `sorry`/`admit`/forbidden tactic, all six checks green. No emitted-image impact, so no pin or layout exposure.

dhsorens:
Thanks — and on the propagation ask, one of the two is already done and the other is yours rather than mine.

## `docs/leaf-routine-targets.md` — **already fixed, and already on `main`**

It went in with #11676 (merged), because that PR had to touch the same file for #10817's row 5 and #11578's `(leaves)` annotation, and three PRs conflicting on one doc seemed worse than one PR carrying three corrections. Current `main`, `:99-105`:

> …`bytes_to_bls_field` … checks `BLS_MODULUS`, the **scalar field order**; `blsg_lt_p` compares against `blsg_p_be`, the **base field** prime, so its counterpart is `Bls12.bytes_to_fq` (`SpecRef/PrecompilesBls.lean:78`) and `blsP`. Two different primes. The scalar-order check is a different routine, `blsk_lt_be`.

(A grep for *"no assertion vocabulary at all yet"* still hits that file, but only inside the sentence quoting the error in order to correct it — worth saying, since the grep alone reads like a survivor.)

## The issue body — **yours to edit; here is the paste**

#11574 is authored by you, so I can't change the body. The two clauses that would mislead the next reader are in the "Ask" section. Suggested replacements:

**Item 1**, which currently specifies an LE limb buffer as the base vocabulary:

> 1. Define the base vocabulary. ⚠️ **BE first**: the two named consumers (`blsg_lt_p` / `bnf_lt_p`) read **big-endian byte buffers**, not limb buffers — the fixture is an MSB-first `lbu`/`bltu` scan with no `ld` and no stride. `fpLimbsIs` (LE, 6×64 BLS / 4×64 BN254) is still wanted, but for the *converters* (`blsg_be_to_le`) and accelerator point forms, so it follows the bridge rather than gating it.

**Item 3**, which asks for triples that already exist:

> 3. ~~First consumers … state the triples for `blsg_lt_p` and `bnf_lt_p`~~ — **already done.** `blsgLtP_spec` (`Bls12G1LtPSAsm.lean:734`) and `bnfLtP_spec` (`Bn254FieldLtPSAsm.lean:732`) are whole-routine triples, `sorry`-free, with genuine numeric posts. What is missing is the **SpecRef bridge and the registry rows** — neither spec appears in `Progress/*`, so both are unwitnessed by `check-axioms.sh`.

And in the same section, the counterpart citation: `bytes_to_fq` (`SpecRef/PrecompilesBls.lean:78`) rather than `bytes_to_bls_field` (`PrecompilesKzg.lean`), since those check different primes.

Happy for you to reword — the load-bearing corrections are just *BE not LE*, *triples exist*, and *base field not scalar order*.

## Next on this issue

The 16-byte EIP-2537 wire-pad relation **in the statement**, then the bridges restating both triples in this vocabulary against `bytes_to_fq` / `bytes_to_g1`, then the two absent registry rows and the `4ch8f-crypto-kernel-inventory.md:53` correction (its "no `cpsTripleWithin` connects any crypto asm body to its spec" is falsified by ~40 `*SAsm.lean` triples, these two included).

Issue #11667, opened 2026-08-06
structural: remove the direct-transfer bypass — the guest charges state gas without entering message preparation, the spec has no such path


pirapira:
## The divergence

The guest has a **direct-transfer fast path** that charges state gas without entering the top-level message-preparation seam. `topLevelValueRecipientStateGasAsm` computes the `NEW_ACCOUNT` charge into `t0`, and `BlockVerdictSimpleTransferPublish` stores it straight into `evm_state_gas_used` at its `state_publish_ok` path (`:105`), with the legacy capture recording it at `:137`/`:150` — all without reaching `emitTopLevelMessageD0Preparation` (`Dispatch.lean:2332`, invoked at `:2881`).

**`execution-specs` has no such bypass.** From the pinned tree:

- `fork.py:1141` calls `prepare_message` and `:1147` calls `process_message_call` — **unconditionally, for every transaction**. There is no value-transfer fast path in `process_transaction`.
- `utils/message.py:79` sets `state_gas_reservoir=tx_env.state_gas_reservoir` unconditionally when constructing the `Message`.

So a plain value transfer to a new account gets a reservoir exactly like any other call, and its `NEW_ACCOUNT` charge draws from the pools like any other charge. The spec has **one** path; the guest has two.

This is the same shape as the `AccountState` overlay retired in #11581/#11638 and the storage ECALL bridges removed in #11663: a guest-side structure with no spec counterpart.

## ⛔ The part that is not about gas

The bypass was found through state gas, but that is a symptom. `emitTopLevelMessageD0Preparation`'s own documentation says it performs, in order:

- authorization traversal and the `set_delegation` callback, which **publishes auth nonce effects before the state-gas boundary is checked**;
- append-only log cursor snapshots, **so an authorization-phase OOG can roll those effects back** without disturbing earlier transactions;
- **access seeding** — and note its comment that this is *caller-scoped rather than depth-guarded in the spec*, since nested messages inherit the parent's accessed-address set.

It also carries an explicit warning: *"Do not move individual instructions across this boundary: state-gas and early-exit ordering is part of the transaction semantics."*

[...]

PR #11682, opened 2026-08-07
chore(#10685): retire dead bv_stx_sender_* single-tx BSS twins

pirapira:
## Summary
Closes the dead half of #10685 (milestone 26Cleanup). Maintainer-directed enumerate-then-cut on the **emitted image**, not names.

### Cut set (ACK'd by coord)
| item | evidence |
|---|---|
| `bv_stx_sender_addr` | BSS `.zero 32` label only; **0** la/ld/sd in emitted `.s` |
| `bv_stx_sender_acct` | BSS `.zero 128` label only; **0** code refs |
| `.Lbv_stx_checks_done` | Lean **comment** in MtxRuntime BAI-store path list; **absent** from `.s` (never a label) |

### Kept (enumeration changed the survivor list)
Coord handed four names. Emitted-image read settled **one LIVE**:
- **`bv_pending_upfront_balance_flag` + `pending_upfront_sender_*`** — MTx producer sets flag=1 and stages the tuple; consumers in `dispatcher_seed_pending_upfront_sender_balance`, EIP7708Logs, OOG path, Function zero (**31** hits in `.s`). **KEEP.**
- `block_verdict_creation_runtime` (ex single_tx_creation) — LIVE jal from MTx
- `.Lbv_st_*` — LIVE; **#11667 separate**, not this PR

### Layout
| pin | base | cand |
|---|---|---|
| textSizeBytes | `0x55c08` | **`0x55c08` UNCHANGED** (stop condition: any .text move = wider closure) |
| bssSizeBytes | `0x190295c0` | **`0x19029520`** (−`0xa0` = 32+128) |

Mid-section BSS drop shifts every symbol after the site by `−0xa0` (teer, pending_upfront, …). GuestAddrs + `symbol-addresses.tsv` regenerated; second pass **NO-OP**; `check-guestaddrs-starts` ok=340. No absolute-address rfl outside GuestAddrs/tsv (GuestImage uses `bssSizeBytes` extent only).

Comment-only edit is outside the asm string (Lean `--` prose) — cannot move a byte; measured `.text` size identical.

### Measure
ELF base sha256 `0c55fd6c…` @ main tip pre-cut; cand `ce91d1af…`
- r200 seed=20260806: **F→P=0 P→F=0**
- named 00178/01114/05814/11710/03736/11617/11738: no flip

### Out of scope
- #11667 SimpleTransferPublish
- pending_upfront protocol
- creation_runtime rename


pirapira:
**coord review: `merge!` applied.** The `build` failure is derived-artefact pin drift against a moved base, not an author-fixable defect — handing it to the merge actor to resolve and repin.

## The cut is exactly the approved set

Three items, nothing else: `bv_stx_sender_addr` (32 B `.bss`), `bv_stx_sender_acct` (128 B), and the `.Lbv_stx_checks_done` mention in a comment. `bv_pending_upfront_balance_flag` correctly **kept** — the enumeration found a live MTx producer, consumers in `dispatcher_seed_pending_upfront_sender_balance` / `EIP7708Logs` / the OOG path, and 31 hits in the emitted `.s`. I had handed that over inside a list headed "survivors", which framed a live mechanism as a deletion candidate; the emitted-image read overrode it in the safe direction. That is enumerate-then-cut working.

## Prediction held, including the part I asked to be checked

| quantity | predicted | CI reports |
|---|---|---|
| `textSizeBytes` | UNCHANGED `0x55c08` | **OK `55c08`** |
| `bssSizeBytes` | `0x190295c0 − 0xa0 = 0x19029520` | **OK `19029520`** |
| `dataSizeBytes` | unchanged | OK `53b0` |
| `symbol-addresses.tsv` | two `bv_stx_*` rows dropped | **OK, matches ELF** |

⭐ The `.text`-unchanged prediction is the load-bearing one and it held: the deleted symbols were `.bss` only, and the third item is a Lean `--` comment *outside* the asm string, so it cannot move an emitted byte. Any `.text` delta would have meant the closure was wider than enumerated.

The absolute-address question I raised was answered properly rather than waved off: mid-`.bss` drop at `0xb8a8c9a0`, everything after shifts `−0xa0` (`teer_success_count` `0xb8a8ca40 → 0xb8a8c9a0`; `pending_upfront_flag` `0xba2be148 → 0xba2be0a8`), 99 `GuestAddrs` pins past the drop regenerated **through the tsv rather than hand-edited**, no absolute-address `rfl`/`decide` outside `GuestAddrs`/tsv, and `GuestImage` proofs use the `bssSizeBytes` extent only. Measure: r200 seed 20260806, F→P=0, P→F=0, named controls no flip.

## Why the failure is the actor's

`check-region-map` reports `DRIFT EvmAsm/Codegen/RegionMapLinkPins.lean stale vs gen-out/regionmap/stateless_guest.elf` — while **every named pin in that file reports OK** (`textSizeBytes`, `dataSizeBytes`, `bssSizeBytes`, `callFrameArenaBase`, `evmMemoryPoolBase`, `syslogBase`) and the tsv matches. So the stale content is not in the quantities this cut moves.

The base explains it: this branch regenerated against `ad134182b`, and `origin/main` is now `96c3eb50d` — #11677 and #11679 landed in between. A generated pin file describes one ELF; when the base moves, the file describes an ELF that no longer exists. Regeneration has to follow the **final** rebase, and against a continuously-moving main the author cannot know which rebase is final. The actor does, because it is the one that repins.

⚠️ Author: nothing is owed here — do **not** rebase or re-regenerate this branch.

PR #11686, opened 2026-08-07
chore(codegen): delete unused BLS12-381 charge*GasAsm helpers (#10527)

pirapira:
## Summary
Delete three unused BLS12-381 gas helper `def`s from `PrecompileRuntime.lean` (#10527):
- `chargeBls12G1MsmGasAsm`
- `chargeBls12G2MsmGasAsm`
- `chargeBls12PairingGasAsm`

Rewrite `bls12MsmCostAsm`'s docstring so it states the failure-routing property directly (child `failureLabel` vs whole-tx `.exit_outofgas`) without naming the deleted helpers.

## Why zero image risk
These are Lean `def String` taking register-name parameters. An **unreferenced** def String is never concatenated into any program body — they emit **zero bytes**. No pin regen, no layout exposure.

## Verification (gate)
Emit `stateless_guest` to the **same** `-o` path before and after deletion:
- **ELF sha256 (both):** `6b0844ff64121e556f6843beab257a70fc019fc80e59fa60565568b0e30751a3`
- `cmp` byte-identical
- `.s` files identical (1737805 B)
- No refs in `PROGRESS.md` / `DRIFT.md` / `docs` (checked — resurrection checklist not armed)
- No remaining name refs under `EvmAsm/`

(First cmp of different `-o` basenames differed only in embedded `base.o` vs `cand.o` path strings — not code.)

## History (restored before — do not restore again)
- `067b60192` fix: restore chargeBls12G1MsmGasAsm + chargeBls12G2MsmGasAsm (2026-06-05, pure +60, no body)
- `f7535b9a8` fix: add chargeBls12G1/G2MsmGasAsm to canonical PrecompileRuntime
- Beside `691350838` refactor: split noop child-frame handlers — look lost-in-split and restored by bookkeeping (PROGRESS snapshot bump only)

Maintainer approved deletion. Do **not** bring them to parity (second implementation of one op with no caller).

## Why obsolete (not staged)
All three route precompile failures to `.exit_outofgas` (whole-transaction OOG), predating the EIP-150 child-allotment model. Live path: `bls12MsmCostAsm` + `bn254ChargeGateAsm` at `ChildFrameHandlerTails.lean` (~968, ~1128); pairing cost inlined ~1205–1215.

Defect content worth recording so nobody re-derives them:
- Lack spec length gates (`bls12_381_g1.py:91-103` InvalidParameter on len 0 / not divisible by 160)
- Zero pair-count: `addi -1` then table index loads **8 bytes before** the discount table

## Test plan
- [x] lake codegen stateless_guest before/after same `-o` → identical sha256
- [x] no PROGRESS/DRIFT/docs checklist lines for these names
- [ ] CI build / check-unimported


pirapira:
**coord review: content approved.** Holding `merge!` only until `build`/`source-checks` report — this is a Lean deletion, so the build is the one thing the byte-identity gate cannot speak for. No action owed by the author.

## The gate for this PR is byte-identity, and it held

These were unreferenced `def : String` values taking register-name parameters, so they were never concatenated into any program and emitted zero bytes. **Same `-o` path, before and after ELF sha256 identical: `6b0844ff64121e556f6843beab257a70fc019fc80e59fa60565568b0e30751a3`.** That is the whole verification — no pins, no layout exposure, no `.text`/`.bss` movement, and if the sha had moved it would have meant something referenced them and the premise was wrong.

## The deletion closure is complete — I checked the two things a cut like this usually misses

| shared callee | status after the cut |
|---|---|
| `chargePrecompileGasAsm` (called by all three deleted helpers) | **still live** — three other users: `Modexp.lean:138`, `PrecompileRuntime.lean:90`, `:586`. Not orphaned by this. |
| `bls12_g1_msm_discount_table` / `bls12_g2_msm_discount_table` | **still referenced** — `Dispatch.lean` defines them, live `bls12MsmCostAsm` consumes them via `tableLabel`. Correctly retained. |

So this does not leave a second-order orphan behind, which is the failure mode that turned the #10685 cut into four revisions.

## The docstring was rewritten rather than truncated

The old text at `:684` explained the live path by reference to a now-deleted name. The replacement states the property directly — *"Multiplication overflow guards route to `failureLabel` (child-frame fail), not `.exit_outofgas` (whole-transaction OOG): the live path is the EIP-150 allotment model, not a precompile-local transaction abort."*

⭐ That is the right call and worth noting explicitly: the sentence records **why the live path differs from what was deleted**, which is exactly the knowledge that would otherwise be lost with the code. A dangling name would have been worse than nothing; deleting the sentence would have been worse still.

## Recorded for the next person

The resurrection risk was real — `067b60192` (*"fix: restore chargeBls12G1MsmGasAsm + chargeBls12G2MsmGasAsm"*, pure `+60`, no explanatory body) and `f7535b9a8` both put these back once. Confirmed no reference survives in `PROGRESS.md`, `DRIFT.md` or `docs/`, so the bookkeeping mechanism that plausibly drove that restore is no longer armed.

And the defect content is preserved in #10527 rather than lost with the code: no `len == 0` / `len % 160` gate (`bls12_381_g1.py:91-103` raises `InvalidParameter`), failure routed to whole-transaction OOG instead of the child allotment, and the `k = 0` negative-index read (`addi -1`, `slli 3`, `add` → 8 bytes *before* the table).

PR #11687, opened 2026-08-07
fix(codegen): remove dead direct-transfer precompile arms

### LINKED ISSUE #11667: structural: remove the direct-transfer bypass — the guest charges state gas without entering message preparation, the spec has no such path

pirapira:
Closes #11667

## Summary

Remove the dead direct-transfer precompile publication cluster from
BlockVerdictSimpleTransferPublish. The lane-flag representation change is
intentionally not part of this PR.

## Why the cluster is dead

The closure argument closed in this order:

1. The sole writer stores `lane = 2` at `0x80031cd4`, immediately before
   the jump.
2. A recursive jal-target plus relocation scan of the transitive closures of
   `precompile_shared_select_price` and `precompile_shared_execute` found
   zero references to the lane. Nothing between that store and either branch
   can clear it.
3. Therefore no arrival can carry `lane == 0` by construction.
4. On the exercised manifest row 118, the structural breakpoint run recorded
   A=1 at entry `0x80019c34` and B=0 at the fall-through/charge block
   `0x80019c4c`, with zero faults.

The first corpus run was a no-hit on a population lacking the relevant class:
there were zero transfers of any kind in 209 rows. I did not treat that as
evidence of unreachability; the second run was the deliberate discriminating
run. The pre-rebase corpus pair was:

- base ELF: `a12dcdb3d765f465b910cf8a9ba0ba6df9cb4a51da131bd6e0e66d726ea5c331`
- candidate ELF: `48a249a039b1d6c2e88382a310bc07ea642236d1bab3f49cae6488065b7b309f`
- manifest SHA-256: `4ce106615e29156af428f66ea75d8aad189713d860f5d29ef8596912aaa12f8b`
- 209 rows, zero faults on both legs, zero per-row output mismatches,
  F-to-P=0 and P-to-F=0.

This corpus leg predates the rebase. The rebase adds only 11682's deletion of
.bss symbols; that change is independently corpus-verified and shares no
symbol or section with this .text cut. The corpus was deliberately not rerun
after rebase because the two changes are non-interacting.

SimpleTransferPublish:37 is the `bnez lane` into `fail_publish`, inside a
region reachable only when `lane == 0`. The preceding `beq` already
diverts every `lane == 2` arrival, so that branch can never be taken. This
is why the cluster read as live for so long.

This also removes the `bv_st_no_log` baseline that I added in 11672. This is
an explicit self-correction.

## Layout accounting

The intended cut is `0x95c` bytes. Two named branch-over-jump relaxations
also disappear:

- old site `0x80019a90`: `bgeu` plus jump becomes one inverted `bltu`;
- old site `0x80019ab4`: `bgeu` plus jump becomes one inverted `bltu`.

Each relaxation is 4 bytes. The prediction was corrected for this named
reason, rather than silently amended:

`0x55c08 - 0x95c = 0x552ac`, then `0x552ac - 0x8 = 0x552a4`.

After rebasing onto main `d4682f860`, the linked ELF is
`85595213d84a0e6bbb5d964668a8761d81e6a6ecdd689f832873a1b75587c037` and
has:

- `.text = 0x552a4`
- `.data = 0x53b0`
- `.bss = 0x19029520`

The post-rebase pin generator's second pass is a no-op. The preserved regions
were checked by symbol and offset, with all four labels present.

## Verification

- `NO_BUILD=1 scripts/check-region-map.sh`: clean
- `python3 scripts/asm_to_program.py check-guest-addrs`: clean, 1021 symbols
- `scripts/check-guestaddrs-starts.sh`: 340/340
- `scripts/check-unimported.sh`: 2765/2765, zero orphans
- `LAKE_ARTIFACT_CACHE=false scripts/check-asm-to-program.sh`: clean,
  380 converted definitions
- `LAKE_ARTIFACT_CACHE=false lake build`: clean, 4735 jobs
- `LAKE_ARTIFACT_CACHE=false scripts/check-no-warnings.sh`: zero warnings


pirapira:
**coord review: content approved.** Holding `merge!` only until `build` reports. The other five checks pass. Two observations below, one of which I'd like addressed in a follow-up rather than here.

## The cut matches the ACKed extent exactly

`BlockVerdictSimpleTransferPublish.lean` goes **164 → 25 lines**, leaving precisely the two entry checks and nothing else: `.Lbv_simple_transfer_precompile_fail` diverting `lane == 2` to `.Ldtrc_mtx_precompile_failure`, and `.Lbv_simple_transfer_emit_tl_then_after_tx_gas_precharge` diverting `lane == 2` to `.Ldtrc_mtx_precompile_success`. `emit_tl` is retained — it is the live success join, and it was the one item I predicted would be wrongly swept if the cluster were treated as one contiguous block.

Post-rebase onto `d4682f860`: `.text 0x552a4`, `.data 0x53b0`, `.bss 0x19029520` (matching merged main after #11682). Pins regenerated **after** the rebase with a clean second pass; `check-region-map`, `check-guest-addrs` (340/340), `check-asm-to-program` clean.

## Why this closed by enumeration rather than sampling — the part worth preserving

1. Sole entry stores `lane = 2` at `0x80031cd4`, on the instruction before the jump.
2. **Transitive** closure of `precompile_shared_select_price` and `precompile_shared_execute` contains **zero** lane references — recursive `jal`-target plus relocation scan, 17 named `zkvm_*` edges plus the indirect ecrecover backend. So nothing between the store and either branch can clear it.
3. Therefore no arrival can carry `lane == 0`, **by construction**.
4. `A=1, B=0` on manifest row 118 confirms the construction behaves as read on an exercised row.

⭐ **And the first corpus run was a no-hit on a population lacking the class** — zero transfers of any kind across 209 rows — which was correctly *not* treated as unreachability. That is why the second run existed, and it is the methodological point I most want on the record.

The 8-byte prediction miss was localised rather than absorbed: two branch-over-jump pairs (`bgeu` + `j`) collapsing to a single inverted `bltu` once the cut pulled their targets inside the 13-bit branch range, both **before** the first cut range. `0x55c08 − 0x95c − 8 = 0x552a4`. A prediction corrected for a named reason is evidence; one quietly amended is not.

## ⚠️ Observation 1: adjacency is now control flow, and it should say so

The two surviving stubs are adjacent, so the `lane != 2` path at the fail label now **falls through into the emit stub**, tests the lane a second time, and continues to the after-tx join. That is safe precisely because `lane == 0` cannot reach the fail label — the deletion's whole premise — but nothing in the file says so any more, and the fall-through *looks* intentional.

Please add one line at that boundary recording that the fall-through is unreachable and why (`lane` has exactly one non-zero writer, `BlockVerdictMtxEoa.lean:26`, on the instruction before the jump into this region). Without it, the next reader sees a fail path deliberately continuing into a success path.

## Observation 2, follow-up not blocker: the parameter is now unused

`blockVerdictSimpleTransferPublishAsmFor (_ctxLabel : String)` no longer uses its argument. Fine in Lean, but the caller still threads a context label into a function that ignores it. Worth collapsing whenever this file is next touched.

## Confirmed for the record

- `bv_st_no_log` — the state-gas baseline added by #11672 — is removed by this cut, on a block that had **zero in-edges**. Stated as a self-correction rather than left in the diff.
- `check-unimported` moves 2844 → 2765 across the rebase; that is a base effect (declaration population changed with main), not a candidate effect.
- Corpus evidence (209 rows, 0 faults, 0 mismatches, F→P = 0, P→F = 0) was measured **pre-rebase** on `a12dcdb3d` vs `48a249a03`. Accepted deliberately: the rebase adds only #11682's `.bss` deletion, itself independently corpus-verified, sharing no symbol and no section with a text-only cut. The bases are named so a reader can see which tree the evidence witnesses.

Issue #11688, opened 2026-08-07
bug(codegen): header resolve leaves a garbage high balance limb for a CREATE'd account, flipping opt to Present (fixture 01114)


pirapira:
Pre-existing defect on `origin/main`, found while isolating #11431 and confirmed **not** caused by it. Evidence from grok. (Issue generated by Claude Code.)

## Symptom

Fixture **01114** fails with `bv_fail=1` (state-root disagreement). The CREATE'd account's block account-write row carries a corrupted balance and a wrongly-set optional-state flag:

| row | address | balance | `opt` | `vm` |
|---|---|---|---|---|
| `blk[2]` | CREATE `f59c…` | ends `…64` | **1** | `0xf` |
| `blk[3]` | beneficiary `c0f6…` | ends `…64` | 0 | `0x21` |

The beneficiary's `…64` is its **real** balance encoding. The CREATE row should be `bal=0, opt=0`.

## The writer is named

`SPIKE_WATCH` localises a **single** write of `0x6400000000000000` into the CREATE row's **high balance limb**, at `account_writes_apply_deletes`, on the **fill path** that copies resolve scratch into the tx row. `account_writes_block_upsert` then propagates it into the block map.

Resolve scratch at fill (`sp+40`): `nonce=0`, balance limbs `0, 0, 0, 0x6400000000000000`.

⚠️ **Only the high limb is garbage** — this is not a clean little-endian or big-endian `100` mis-encoded, so it is not a byte-order bug on a real value. Block AW count at `apply_deletes` entry is **0**, so the resolve falls through to the header path; the delete-set address is the CREATE address, which is correct. So the header path returns a "balance" whose high limb is garbage for this account.

## The consequential part is the flag, not the balance

`OR`-of-limbs is now nonzero, so `keep_present` sets `opt=1` — the row reads **Present** where it should read Present-None/absent. That is what moves the state root, and it is why the symptom presents as two things at once (a foreign balance *and* a flipped flag) rather than a wrong number.

## Provenance — why this is filed separately

grok initially reported 01114 as a `PASS → FAIL` regression from the #11431 system-contract execution work, then **self-corrected**: that measurement was against a stale ELF (`/tmp/opencode/hist/main.elf`). Freshly built `origin/main` codegen fails 01114 **identically**.

| id | stale `hist` | fresh `origin/main` | #11431 candidate |
|---|---|---|---|
[...]

PR #11691, opened 2026-08-07
feat(codegen): execute EIP-2935/4788 via stage_system_call (#11431)

pirapira:
## Summary
#11431 rescope: run EIP-2935/4788 through the shared `stage_system_call` path (same shape as 7002/7251), deleting the formula descriptor + identity fail-65 path.

### Consumers moved vs deleted
| Item | Disposition |
|---|---|
| `process_block_start_system_transactions` | **NEW** — BAI=0; code_at 4788 then 2935; calldata; stage_system_call; write_sets/read_sets incorporate |
| `stage_system_call_payload` calldata | **EXTENDED** — `ssc_calldata_ptr`/`len` (7002 stays empty) |
| MtxRuntime formula seed (`system_write_descriptors` + identity + `append_modeled` seed-only) | **DELETED** from live path; replaced by helper jal |
| StateRoot formula (`swd_*`, identity, `bsr_sys_change`/`bsr_beacon_change`, `bara_skip`, `apply_modeled_system_post_fields`) | **DELETED** from live path; map sole root authority |
| `current_block_access_index = mtx_i+1` at each MTx loop start | **ADDED** (system leaves BAI=0; sender bal/nonce emit must not stay at 0) |

### Atomic set
| C | Item |
|---|---|
| C1 | SLOAD tier-2: block map seeded by real SSTORE via `write_sets_incorporate_tx` |
| C2 | Terminal MPT: drop `swd_*` → `bsr_sys_change` / `bsr_beacon_change` |
| C3 | Drop `bsr_apply_modeled_system_post_fields` |
| C4 | `bara_skip_modeled_system = 0` |
| C5 | BAL BAI=0 from real emit (inside incorporate) |
| C9 | Host `evm_oldest_ancestor_offset = max(mark,1)` after system phase (unconditional; `fork.py:908`) |
| C11 | **BAIL REMOVED**: identity fail-65 gone — crafted-fa base bv**65** → cand bv**1** (#11547 root) |
| C12/C13 | No `swd_*` vlen / `bsr_sys_has_*` on live path |

**DEFERRED (not silently out of scope):** C6 side log, C14 capacity, C15 probes.

### Empty-code vs missing-preimage (#11520)
`code_at_header` status-5 is split (DispatchTx materialize pattern + `chahsr_empty_code_hash`):
- **EMPTY_CODE_HASH** → no-op continue (`process_unchecked` fork.py:788 — no code check)
- **non-empty hash, preimage missing** → `.Lpbs_fail` reject (unsound to proceed)
- status 1 absent account → skip; status 0 + len>0 → run

### honest-fr +1 byte (four-citation chain)
Not a no-op SSTORE **membership** bug (gate fires; 2935/4788 zero writes correctly absent from changes).

Byte-diff of account-list payload: 2935 `storage_reads` empty (`c0`) vs `[slot 0]` (`c180`).

1. SSTORE calls `get_storage` / prestate even when unchanged — `storage.py` SSTORE path  
2. `get_storage` adds `(address, key)` to `storage_reads` — `state_tracker.py:295`  
3. No-op gate excludes from changes when `pre_value == post_value` — `block_access_lists.py:670-676`  
4. Reads not in changes are **retained** as storage_reads — `block_access_lists.py:549-552`  

NEW_CODE `600060005500` does SSTORE(0,0); pre==post → not a change, still a read. Fixture `build_honest.py` had wiped 2935 BAL to all-empty; regenerated with `storage_reads=[0]`. Artifact: `/tmp/opencode/k3-11431/honest-fr.input`. `rebuilt_len == supplied_len == 521`.

## Measure vs puremain
| set | result |
| honest-fr | **F→P** (bv65 → PASS) |
| crafted-fa | bv65 → **bv1** root — bail removed |
| 05814/05862 | PASS |
| r200 seed=20260807 | **F→P=0 P→F=0** (after BAI loop-start fix for 01869) |
| 01114 | FAIL both (pre-existing main; out of scope) |

ELF sha256 `89a114c018bef498ac183c3cd8850dee54633d03cca9c5fc93115e85942656e3` (pre-final pin; regen after rebase)  
Pins: textSizeBytes `0x55924` bssSizeBytes `0x19029540` (post-#11675/#11686 main; second-pass fixpoint)

## CI fixes in this PR
- check-build-units-link: link `processBlockStartSystemTransactionsFunction` into `zisk_stateless_verdict_v2`
- check-code-preimage-empty-hash: EMPTY_CODE_HASH split (no allow-list bump)

## Test plan
- [x] check-code-preimage-empty-hash OK
- [x] check-build-units-link OK  
- [x] check-guestaddrs-starts OK
- [x] honest-fr F→P; crafted bv1; r200 0/0
- [ ] CI build green


pirapira:
**coord review: content approved, one documentation requirement before I label.** This closes the convergence half of #11431 and retires the fail-65 bail. (Comment generated by Claude Code.)

## ⭐ The bail is genuinely gone, and the evidence is the right kind

`crafted-fa` moves from `bv65` (identity mismatch) to **`bv1` (root)**. So a block whose EIP-2935 contract carries substituted code is still rejected, but now because **real execution disagrees** rather than because a code-hash gate fired. That is the difference between a conservative reject and a correct one, and it is what makes the interim gate removable rather than permanent.

Recorded for the record: #11446 added fail-65 as the FA-closing interim on 2026-08-04 with an explicit statement that executing the contracts was the convergent shape. That interim lasted three days and did not become permanent by default.

## ⚠️ The `honest-fr` fixture was regenerated — this needs its spec chain stated in the PR body, not just the outcome

`honest-fr.input` was rebuilt with `storage_reads=[0]` and now gives `rebuilt_len == supplied_len == 521`, F→P. **"I regenerated the input and now it passes" is the shape of a circular result**, so the PR must show the corrected fixture is what the *reference* produces, independent of what the guest produces. It is — here is the chain, please put it in the body:

1. `vm/instructions/storage.py:108` — `sstore` calls `get_storage(tx_state, current_target, key)` for gas accounting, **even when the value is unchanged**.
2. `state_tracker.py:295` — `get_storage` does `tx_state.storage_reads.add((address, key))`. So the read is recorded by the SSTORE itself; no SLOAD is needed, and `600060005500` contains none.
3. `block_access_lists.py:670-676` — `if pre_value != post_value` excludes the no-op from `storage_changes`.
4. `block_access_lists.py:549-552` — `if slot not in changes.storage_changes: storage_reads.append(slot)` **retains** it as a read, precisely because it was not a change.

⇒ So a spec-consistent supplied BAL for that block **must** carry `storage_reads=[0]`, and the original hand-built `honest-fr.input` (`c0`, empty reads) was **mis-constructed** — not a witness the guest failed. The regeneration corrects the fixture; it does not accommodate the implementation. State it that way, with these four citations, so nobody reading this later has to reconstruct whether the test was bent.

⭐ And note my own hypothesis was half right in a way worth recording: I predicted the no-op gate as the cause and it *is* implicated — but as the reason a **read** survives, not the reason a **change** was wrongly emitted. grok's decode showed the gate firing correctly on both zero writes. `c0` vs `c180` is a one-byte RLP list difference, not a membership difference, and that is why an entry census could not find it.

## The BAI defect is a real catch and the discipline caught it

`process_block_start` left BAI at 0, and this tree's user-tx paths often emit sender balance/nonce **without re-storing BAI**, having relied on it already being 1. Fixed by setting `current_block_access_index = mtx_i + 1` at each `.Lbv_mtx_loop` iteration start, before prepare. 01869 PASS, r200 P→F = 0.

⚠️ Worth flagging as a latent hazard this change *exposed* rather than created: **a consumer that relies on an ambient value already being set is correct only while nothing else sets it earlier.** Inserting a system phase at BAI=0 is exactly such an earlier setter. Please state in the body that the fix makes the index explicit per iteration rather than ambient — that is the durable property, and a future insertion ahead of the loop would otherwise re-break it the same way.

## Remaining requirements before `merge!`

1. The four-citation spec chain above, so the fixture regeneration is justified by the reference.
2. State fail-65's removal explicitly as a **bail removed**, with the `bv65 → bv1` trace.
3. Pins: `textSizeBytes 0x55864`, `bssSizeBytes 0x19029540` are reported post-#11682 — confirm they were regenerated **after** the final rebase onto current main (`a1d30ba3a`, which now also carries #11686), with a clean second pass.
4. Name which consumers moved and which were deleted (C1–C5, C11–C13 per the enumeration), and state C6/C14/C15 as **deferred** rather than silently out of scope.

Gate otherwise met: `honest-fr` F→P, `crafted-fa` reject without code 65, r200 P→F = 0 against freshly built `origin/main` (not the stale `hist` artefact), 05814 PASS. 01114 correctly out of scope — pre-existing on main, tracked as #11688.

PR #11694, opened 2026-08-07
proof(#11341): structural half of the rlp_list_count_items bridge

dhsorens:
Partial progress on #11341's **last remaining** `.machineOnly` row. Landing in two pieces because this one is materially bigger than the three already done.

## State of #11341's four rows

| routine | basis | bridge |
|---|---|---|
| `rlp_content_to_u256_be` | ✅ `.bridged` | `ContentToU256BeBridge.lean` |
| `rlp_bytes_encoded_size` | ✅ `.bridged` | `RlpBytesEncodedSizeBridge.lean` |
| `rlp_list_encoded_size` | ✅ `.bridged` | `RlpListEncodedSizeBridge.lean` |
| `rlp_list_count_items` | ⬜ `.machineOnly` | **this PR, partially** |

## The gap, concretely

`rlp_list_count_items_spec_within`'s post bottoms out in `Success bytes base listLen count`, stated over `StrictPrefix` (`RlpListNthItemSAsmBase:312`) — a **second, independent copy of the walk relation**. The shared-model side is `Rv64.RLP.DecodeChain` (`WalkDecodeBridge:412`). **Nothing referenced both.** That is precisely your complaint: the theorem says the routine matches *our restatement* of RLP, and a transcription slip in that restatement is invisible.

## What this PR proves — two of three mismatches

⭐ **Direction reversal (`DecodeChain.snoc`).** `StrictPrefix` grows by *snoc*; `DecodeChain` is defined by recursion on the item list from the *front*, so extending it at the back is not definitional. This lemma supplies that, and **without it the two relations cannot be related by induction in either direction.** It did not exist in `WalkDecodeBridge`. Same shape as the `nibblePrefix`-vs-`flatMap` mismatch #11422 had to resolve with its own induction.

⭐ **Items not carried (`StrictPrefix.decodeChain_of_itemBridge`).** `StrictPrefix` tracks only a *count*, so the bridge must produce the item list existentially:

```lean
StrictPrefix bytes base endPtr startOff count off →
  ∃ items : List RLPItem, items.length = count ∧ DecodeChain bytes startOff items off
```

`decodeItems_of_chain` (already general over `List RLPItem`) then closes the last step to a full `decodeItems` run.

## ⚠️ What is NOT done — and it is a hypothesis, not a `sorry`

The per-item translation — `rlpItemDecode` (five-disjunct **`BitVec`**-level relation) to `decodeAux` (**`Nat`**-level) — is isolated as the named predicate `RlpItemDecodeBridges` and **taken as a hypothesis**. So nothing here claims more than it proves; the structural result is explicit about what it rests on, and the remaining work is one named obligation rather than something entangled with the induction.

I chose this order deliberately: the structural induction is the part that could fail for interesting reasons, so proving it first meant a failure would have taught us something. The per-item step is five mechanical disjuncts now that the shape is known to work, and `WalkDecodeBridge` already has the byte-string pieces (`decodeAux_singleByte_bridge`, `_shortBytes_bridge`, `decodeAux_bytes_all_fuel_of_decode`) to assemble.

## ⛔ The registry row deliberately stays `.machineOnly`

Moving it to `.bridged` requires the per-item obligation **discharged** — that is what makes the differential actually transfer. Grading it now on the strength of the structural half would be exactly the invisible overclaim #11637's gate exists to stop.

## Verification

- `lake build EvmAsm.Codegen.Programs.RlpListCountItemsBridge` — green; registered in `Programs/Imports.lean`
- `#print axioms` on both new theorems — **`[propext]` only**. No `Classical.choice`, no `sorryAx`
- `check-forbidden-tactics.sh` — OK
- the only `sorry` string in the file is the word in my own header prose

## Context found while scoping

The two sides **agree in content**: `rlpItemDecode`'s disjuncts carry the same canonicality rules I verified against the pinned `ethereum_rlp` on the issue — short-string "a 1-byte string's content is ≥ 0x80", long-string "no leading `0x00`, decoded length ≥ 56". So the remaining obligation is a faithful-translation proof, not a hunt for a behavioural gap.

### DIFF
--- a/EvmAsm/Codegen/Programs/Imports.lean
+++ b/EvmAsm/Codegen/Programs/Imports.lean
@@ -559,6 +559,7 @@ import EvmAsm.Codegen.Programs.RlpListCountItemsSAsmNext
 import EvmAsm.Codegen.Programs.RlpListCountItemsSAsmRound
 import EvmAsm.Codegen.Programs.RlpListCountItemsSAsmTail
 import EvmAsm.Codegen.Programs.RlpListCountItemsSAsm
+import EvmAsm.Codegen.Programs.RlpListCountItemsBridge
 import EvmAsm.Codegen.Programs.RlpListCountItemsFlatSAsm
 import EvmAsm.Codegen.Programs.RlpFieldToU64WholeSAsm
 import EvmAsm.Codegen.Programs.RlpFieldToU64FlatSAsm
--- a/EvmAsm/Codegen/Programs/RlpListCountItemsBridge.lean
+++ b/EvmAsm/Codegen/Programs/RlpListCountItemsBridge.lean
@@ -133,6 +133,53 @@ theorem decodeAux_of_singleByte
   rw [hstep]
   exact decodeAux_singleByte_bridge bytes off b hget hb m
 
+/-- The short-string range guards and the canonicality equivalence, read
+    byte-exhaustively for the same reason as `lt_0x80_of_ult`. Bundled because they
+    are all facts about the one prefix byte. -/
+private theorem shortBytes_guards (b : BitVec 8)
+    (hlo : ¬ BitVec.ult (b.zeroExtend 64) (0x80 : Word) = true)
+    (hhi : BitVec.ult (b.zeroExtend 64) (0xb8 : Word) = true) :
+    0x80 ≤ b.toNat ∧ b.toNat ≤ 0xB7 ∧
+      ((b.zeroExtend 64 - (0x80 : Word) = (1 : Word)) ↔ b.toNat - 0x80 = 1) := by
+  revert b; decide
+
+/-- The inner canonicality byte's guard: a length-1 short string's content byte must
+    not be below 0x80 (else the single-byte form was required). -/
+private theorem not_lt_0x80_of_not_ult (c : BitVec 8)
+    (h : ¬ BitVec.ult (c.zeroExtend 64) (0x80 : Word) = true) : ¬ c.toNat < 0x80 := by
+  revert c; decide
+
+/-- Short-string form (`0x80 ≤ p < 0xb8`). Second of the three provable disjuncts.
+
+    Same division of labour as the single-byte case: the decode content is already in
+    `decodeAux_shortBytes_bridge`, so this translates the `BitVec` guards and takes
+    the two facts that belong to other layers as hypotheses —
+    `hstep` (cursor arithmetic, from the walk layer) and `hlen` (the region bound,
+    from the ABI). Neither is a decode fact and re-deriving them here would couple
+    this file to pointer and region reasoning it has no business doing. -/
+theorem decodeAux_of_shortBytes
+    {bytes : List (BitVec 8)} {base : Word} {off : Nat} {next : Word} {b : BitVec 8}
+    (hget : bytes[off]? = some b)
+    (hlo : ¬ BitVec.ult (b.zeroExtend 64) (0x80 : Word) = true)
+    (hhi : BitVec.ult (b.zeroExtend 64) (0xb8 : Word) = true)
+    (hcanon : b.zeroExtend 64 - (0x80 : Word) = (1 : Word) →
+      ∃ c : BitVec 8, bytes[off + 1]? = some c ∧
+        ¬ BitVec.ult (c.zeroExtend 64) (0x80 : Word) = true)
+    (hlen : off + 1 + (b.toNat - 0x80) ≤ bytes.length)
+    (hstep : (next - base).toNat = off + 1 + (b.toNat - 0x80)) :
+    ∀ m, decodeAux (m + 1) (bytes.drop off)
+      = some (.bytes ((bytes.drop (off + 1)).take (b.toNat - 0x80)),
+          bytes.drop (next - base).toNat) := by
+  intro m
+  obtain ⟨hlo', hhi', hcanonIff⟩ := shortBytes_guards b hlo hhi
+  have hcanon' : b.toNat - 0x80 = 1 →
+      ∃ c : BitVec 8, bytes[off + 1]? = some c ∧ ¬ c.toNat < 0x80 := by
+    intro h1
+    obtain ⟨c, hc, hcnot⟩ := hcanon (hcanonIff.mpr h1)
+    exact ⟨c, hc, not_lt_0x80_of_not_ult c hcnot⟩
+  rw [hstep]
+  exact decodeAux_shortBytes_bridge bytes off b hget hlo' hhi' hlen hcanon' m
+
 /-- ⭐ **The snoc lemma — mismatch (2) in one place.**
 
     `DecodeChain` is defined by recursion on the item list from the FRONT, so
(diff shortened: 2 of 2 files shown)


pirapira:
**coord review: the structural work is right and the honesty about scope is exactly right. One question before the second half is attempted, because the tree's own analysis says the remaining obligation may be unprovable as stated.** Checks still running; not labelling yet.

## What is well done, and I want it on the record

⭐ **The registry row deliberately stays `.machineOnly`.** Grading it `.bridged` on the strength of the structural half would be precisely the invisible overclaim #11637's gate exists to stop, and you refused it. That decision is worth more than the lemmas.

⭐ **The remaining gap is a named predicate, not a `sorry`.** `RlpItemDecodeBridges` is a hypothesis, so nothing here claims more than it proves, and `#print axioms` reporting `[propext]` only on both new theorems is checkable rather than asserted.

And the two mismatches you closed are the ones that could have failed for interesting reasons. `DecodeChain.snoc` in particular: `StrictPrefix` grows by *snoc* while `DecodeChain` recurses from the front, so extension at the back is not definitional and **without it the relations cannot be induced across in either direction.** Proving the structural half first, on the grounds that its failure would have taught us something, was the right ordering.

## ⛔ The question: is `RlpItemDecodeBridges` satisfiable on this routine's actual domain?

`Rv64/RLP/WalkDecodeBridge.lean` says of `DecodeChain`, in its own words:

> *"Each link demands fuel-insensitivity (`∀ m`), which is sound only for **byte-string** items (`decodeAux_bytes_all_fuel_of_decode`). Header fields all are. **This is the same restriction that makes the nested-list case of #11341 unprovable** — a nested list's decode *is* fuel-sensitive."*

So the tree already records that the per-item bridge **cannot be discharged for a nested-list item**. Since `RlpItemDecodeBridges` is stated over the five-disjunct `rlpItemDecode`, one of two things must be true, and the PR should say which:

1. **`rlp_list_count_items`'s domain excludes nested-list items** — then the hypothesis should be *restricted to that domain*, with the argument for why the restriction holds at every call site. Stated over all five disjuncts it is unsatisfiable, and a structural theorem whose hypothesis can never be instantiated proves nothing about the routine.
2. **The domain includes them** — then the second half is blocked on a known-unprovable obligation, and that should be recorded now rather than discovered after the mechanical disjuncts are attempted.

⚠️ **I do not think (1) is obvious, and here is the concrete worry.** `rlp_list_count_items` was just wired into `mpt_node_kind` by #11675 to check node arity exactly. MPT nodes carry **embedded child nodes inline when they are under 32 bytes**, and an embedded node is a nested list. If a branch node with an inline child reaches this routine, the nested-list disjunct is live on the path that matters most.

⇒ Please establish which case you are in **before** attempting the five disjuncts. If it is (1), the domain argument is the deliverable and it is small. If it is (2), the honest move is to say so on #11341 and let the row stay `.machineOnly` permanently until `DecodeChain` gains a fuel-sensitive variant — which is a different and larger piece of work.

## Second, smaller ask: a satisfiability witness

Even under (1), please instantiate `RlpItemDecodeBridges` for **one** concrete shape — a single byte below `0x80` is trivial — so the two structural theorems are demonstrably **non-vacuous**. A theorem of the form `H → P` with `H` unwitnessed is indistinguishable from a theorem about nothing, and one instantiation converts the hypothesis from a placeholder into a scaffold. `WalkDecodeBridge` already has `decodeAux_singleByte_bridge` for it.

## Note on the sibling routine, not a request

`rlp_list_count_items` is now live in the guest image on both `zisk_stateless_verdict` and `zisk_step2_verdict` (link sets added on #11675, merged `423371c4a`). So this bridge is against emitted, executing code rather than a probe-only routine — which raises the value of finishing it and is another reason not to let the row read `.bridged` early.

dhsorens:
Answering the blocking question first, since you asked for it **before** the disjuncts: **it is case (2).** Your worry is correct, and the reference states it outright.

## The domain includes nested-list items — confirmed in the spec

`SpecRef/IncrementalMpt.lean:155`, `resolveChildRefAux` (`_resolve_child_ref`):

```lean
  | .bytes ref => …  -- empty → no child; 32 bytes → DB lookup; other length → assert failure
  | .list items => do                      -- ⭐ inline node, decoded directly
      match fuel with
      | 0 => throw (.witnessNodeMalformed "fuel exhausted (node-graph cycle)")
      | fuel + 1 => decodeNodeItemAux fuel nodeDb (.list items)
```

and the docstring says it in words: *"List case: inline node, decoded directly."* So a branch node's children are **either** a 32-byte hash **or** an inline embedded node, and an inline embedded node is a nested list. Your inference from the under-32-bytes rule was right.

⇒ `RlpItemDecodeBridges` stated over all five disjuncts is **unsatisfiable on this routine's actual domain**, and — as you say — a structural theorem whose hypothesis can never be instantiated proves nothing about the routine. So the second half is blocked on a known-unprovable obligation, and the row stays `.machineOnly` **permanently** until `DecodeChain` gains a fuel-sensitive variant. That is a different and larger piece of work, exactly as you framed it.

## ⚠️ And this is worse because of my own #11675

Before #11675, `rlp_list_count_items` was not called from `mpt_node_kind`. My change wired it in to check node arity exactly — which put it on the MPT node path, **the one place inline nested lists are guaranteed to occur.** So I moved this routine from a domain where the nested-list disjunct was incidental to one where it is structural.

I do not think that makes #11675 wrong: it turned an accept into a reject, and counting by span is the *correct* behaviour for arity (a walker does not need to validate interiors to count items). But it does mean the bridge is now blocked on the path that matters most, and I should have connected those two facts when I opened this PR rather than describing the remainder as "five mechanical disjuncts".

## On your second ask — the satisfiability witness

Under case (2) the witness you asked for **cannot exist for the predicate as stated**, and that is the answer rather than a failure to produce it: an instantiation over all five disjuncts would have to cover a nested-list item, which is the unprovable case. Producing a witness for a hand-picked all-byte-strings `bytes` would demonstrate the predicate is non-empty *as a proposition* while saying nothing about this routine, which is the vacuity you are guarding against — so it would be theatre.

The honest constructive version is a **restricted** predicate — quantified over offsets whose prefix byte is below `0xc0` — which is genuinely witnessable (`decodeAux_of_singleByte` and `decodeAux_of_shortBytes` in this PR discharge two of its three forms) and whose restriction is *visible in the statement* rather than hidden. That is worth landing, but it bridges a strictly weaker claim than the row needs, so it should not be mistaken for progress toward `.bridged`.

## What I would do next, for your call

1. Record case (2) on #11341 and mark the row `.machineOnly` **with the reason**, replacing "needs 4 bridge lemmas" for this row with "blocked on a fuel-sensitive `DecodeChain`".
2. Land the byte-string-restricted predicate + witness, clearly labelled as *not* the row's bridge.
3. Leave the fuel-sensitive `DecodeChain` variant as its own issue.

⭐ One correction I owe: the tree already told me this. `WalkDecodeBridge.lean:405-408` says *"This is the same restriction that makes the nested-list case of #11341 unprovable — a nested list's decode is fuel-sensitive."* I reached the same conclusion independently, from the absent list bridge, and only saw that comment when you quoted it. Reading the docstring of the definition I was building on would have saved the round trip.

Issue #11414, opened 2026-08-04
docs(codegen): AccountState flags value 8 is documented as both "delegation state" and "created-this-tx", with two readers


pirapira:
**Attribution: coord**, from a bit-8 audit by evm-asm3-codex during #11413. Documentation/consistency only — **no
behavioural defect is known**, and the current code is safe by path disjointness.

## The divergence

AccountState `flags@+88` **value 8** is described two different ways in one file, using two different naming
conventions, and read by two consumers with two different intentions:

| site | says | reader intent |
|---|---|---|
| `CreateCodeEffectLog.lean:~393` (corrected by #11413) | "bit 8 for delegation state" (value naming) | `eip7702_authority_asof` uses it as **delegation-present** output |
| `CreateCodeEffectLog.lean:551` | "bit3 (created-this-tx) clear — excludes live empty-code CREATE rows (flags 27/59)" (index naming) | `account_state_tombstone_balance_zero` filters it as **created-this-tx** |

Same bit, opposite meanings, and the two texts do not even agree on how to name it — "bit 8" vs "bit3" both refer to
value 8.

## Why it is worth fixing despite no live defect

Safety today rests on **path disjointness**, not on the bit being unambiguous: the tombstone helper is reachable only
from SELFDESTRUCT-beneficiary and new-account gates, never from auth publication, and probes confirm its answers are
unchanged for both the nonzero (00017) and clear (00013) auth cases.

That is a fragile invariant to leave undocumented. `:545` records that the tombstone conjuncts form an **aliveness
gate**, and `:557` records that a previous defect in that exact helper was a **silent NEW_ACCOUNT overcharge at every
caller**. So a future change that lets an authority row reach the tombstone filter would produce a per-account gas
change with no crash and no obvious symptom — and the reader deciding whether that change is safe would consult two
comments that contradict each other.

## What to do

1. **Settle the meaning** — the emitted readers are the authority. If value 8 means "delegation present", correct
   `:551`'s conjunct description and say what it *actually* filters. If it genuinely means both things for different row
   kinds, say that explicitly and name which row kind each reader sees.
2. **Pick one naming convention** for flag bits in this file — value (`value 8`) or index (`bit 3`) — and use it
   throughout. Mixed conventions in adjacent comments are their own hazard.
3. **Record the disjointness as load-bearing**, at the tombstone helper: that it is never reached from auth publication
   is what makes the shared bit safe, so a future caller added to that helper must re-check it.

## Bar

[...]

PR #11695, opened 2026-08-07
cleanup: remove orphaned guest routines

### LINKED ISSUE #11683: cleanup: bal_storage_whitelist_clean is an orphaned routine — the call was deleted, the callee stayed
## The pattern

Three cleanups in flight this week share one shape: **a call site went away and its callee stayed.** The routine keeps emitting, keeps a `symbol-addresses.tsv` row, keeps an UNCONVERTED row in the coverage doc, and reads to the next person as live code.

| instance | state |
|---|---|
| `bv_stx_sender_addr` / `bv_stx_sender_acct` | #10685 — cut landed as #11682 |
| `.Lbv_simple_transfer_no_log_then_after_tx_gas_precharge` and the lane-0 arms | #11667 — reachability under measurement |
| **`bal_storage_whitelist_clean`** | ⬅ **this issue.** #11402 deleted the call (#11404 via batch #11442); the routine stayed |

## Scope

`def balStorageWhitelistCleanFunction` at `BlockVerdictTxsIndependent.lean:237` still exists and still emits the label. Zero call sites under `EvmAsm/` on `origin/main` `96c3eb50d` — the only references are its own docstring (`:225`) and the label definition (`:238`).

Two artefacts suggest it is linked into the **guest**, not only the probe unit:

| artefact | row |
|---|---|
| `scripts/asm-fixtures/symbol-addresses.tsv:408` | `stateless_guest  bal_storage_whitelist_clean  0x800280a0  .text  LINK_DEPENDENT` |
| `docs/4ch8f-guest-image-coverage.md:297` | `0x8002d1dc → 0x8002d350`, **372 bytes**, **UNCONVERTED** |

If so: ~372 bytes of uncallable `.text`, plus an UNCONVERTED row inflating the conversion-ratchet denominator.

## ⚠️ Step 1 is a check, not an assumption

**Confirm the routine is actually in the linked `stateless_guest` image before touching anything.** Both rows above are generated artefacts, and whether they post-date #11404's landing is exactly the kind of thing a stale pin file gets wrong while looking completely plausible. `nm` on the linked ELF settles it. This decides whether deleting the def moves `.text` at all — if it is probe-only, the cut is source hygiene with a byte-identical guest image, which is a different (and easier) PR.

## Then

- Delete the def and any probe wiring that exists only to reach it. `ziskBalTxsIndependentProbeUnit` (`:383`) may or may not depend on it — enumerate rather than assume, and note that `bal_txs_independent` itself is **probe-only after #11183** (`:105`, 0 guest `jal`), so the probe is the live consumer of that whole file.
- Drop the `docs/4ch8f-guest-image-coverage.md` row and regenerate `symbol-addresses.tsv`.
- Predict `textSizeBytes` before regenerating; the drop must equal what was removed and nothing else. A `.text` delta you cannot account for means the closure was wider than enumerated — stop and report rather than repinning through it.

## Not in scope

- `bal_txs_independent` and the `bv_deposit_capture_only` route beyond what #11183 already retired.
- The #11667 cluster and the #10685 cluster — separate PRs, both with owners.

Context: closed #11402 with the full citation table. Its `KEEP bal_txs_independent` argument is void — #11183 retired the route, so the `a0 == 2` bail it protected is no longer reachable from the guest.

pirapira:
## Summary

Closes #11683.

- Remove the orphaned `brpsfAddr20EqFunction` and `balStorageWhitelistCleanFunction` definitions and their probe/guest-closure wiring.
- Remove the three AccountState definitions confirmed to be definition-only and outside every build-unit concatenation: `accountStateRecordAuthFunction`, `accountStateAuthCurrentFunction`, and `accountStateTombstoneBalanceZeroFunction`.
- Regenerate the linked symbol snapshot, `GuestAddrs`, region-map pins, and both 4ch8f coverage documents.

This does not retire the AccountState overlay (11638 did that). It removes orphaned routines left behind by that retirement.

## Linked-image evidence and retained siblings

The step-1 nm check determined the cut shape. On the pre-11687 parent, `brpsf_addr20_eq` was at `0x8002807c` and `bal_storage_whitelist_clean` at `0x800280b0`; the former was exclusive to the latter's direct closure. The two RLP walkers and all seven address constants had other users and were retained. The candidate-B functions were absent from the guest image, so they were checked separately against probe-unit concatenations.

The live AccountState siblings remain:

| Symbol/helper | Live evidence | Decision |
|---|---|---|
| `account_state_created_count` | written at `BlockVerdictCreationStage.lean:478` | keep |
| `account_state_delete_count` | read at `AccountWriteMap.lean:668` against `accountStateDeleteCapacity` | keep |
| `account_state_overflow` | latched at `CreationStage.lean:479`; consumers turn it into `bv_fail_code = 58` | keep |
| Dispatch overflow guard | `Dispatch.lean:3005` explicitly guards the rejection path | keep |
| `accountStateFindFunction`, `accountStateAppendPendingFunction` | emitted by the CreateCodeEffectLog and NonstorageEffectLog probe prologues (`:988-991`, `:327-330`) | keep |

The guest-image scan and probe-unit scan are different populations: helpers absent from the monolithic guest can still be required by the standalone probes. Only the three definition-only candidates above had no build-unit reference.

The last two definitions, `accountStateAuthCurrentFunction` and
`accountStateTombstoneBalanceZeroFunction`, are the two readers named in
issue #11414 (the flags-value-8 documentation contradiction). Both readers
are unreferenced. The live delegation read is on a different word in a
different structure—`account_writes + 96` with stride 128, not AccountState
flags plus 88—so issue #11414's premise resolves by deleting the dead readers
rather than rewriting a comment.

## Size prediction and verification

The two contiguous guest routines occupied `0x34` (`brpsf_addr20_eq`) plus `0x174` (`bal_storage_whitelist_clean`) = `0x1a8` bytes. After rebasing onto `fa6d25b86`, the parent ELF was:

- `.text = 0x552b4`, `.data = 0x53b0`, `.bss = 0x19029520`
- SHA-256 `3ebfa072affdffbbceb8b8338fec8d6d1d81d541bee7573816bb1a02f05a4d26`

The predicted candidate `.text` was `0x552b4 - 0x1a8 = 0x5510c`; the rebuilt candidate measured exactly that, with `.data` and `.bss` unchanged. Candidate SHA-256 is `1314cfd359ae6903b73e4347a55d260f757611d85edb105e8598557d36a74888`.

The evidence separates the five removals into two classes: the two guest
routines are the `0x1a8` measured `.text` cut, while the three AccountState
definitions are outside every build unit. Removing those three and rebuilding
produced the exact same candidate ELF byte-for-byte (the SHA remained
`1314cfd359ae6903b73e4347a55d260f757611d85edb105e8598557d36a74888`), which
measures that they emitted nothing rather than merely relying on reference
counts. The 209-case sweep was run on that candidate ELF; because this final
change is byte-identical, it has the identical sweep result and did not need a
second run.

This exactness follows from the cut being one contiguous region with no interior stubs, so no boundary was miscounted and no target was pulled into branch range by a gap. In contrast, 11667 removed two non-contiguous ranges with four preserved stubs and two 4-byte branch relaxations. The old parent `0x55c18` versus the post-11687 parent `0x552b4` differs by `0x964` (2404 bytes), independently confirming that earlier 11687 measurement.

## Checks

- `LAKE_ARTIFACT_CACHE=false lake build`
- `scripts/check-guestaddrs-starts.sh`
- `python3 scripts/gen-region-map-link-pins.py --elf gen-out/regionmap/stateless_guest.elf --check`
- `scripts/spike/standing_controls_sweep.py`: 209 cases (r200 + 9 named controls), output differences 0, empty flip set, F-to-P 0, P-to-F 0, verdict mismatches 0.


pirapira:
**coord review: content approved.** Holding `merge!` only until `build` reports; the other five checks pass. I verified the deletions and, more importantly, the **retentions**, on the branch rather than from the report.

## Deletions confirmed, retentions confirmed

| symbol | `origin/main` | PR branch |
|---|---|---|
| `balStorageWhitelistCleanFunction` | 2 files | **0** |
| `brpsfAddr20EqFunction` | 2 | **0** |
| `accountStateRecordAuthFunction` | 1 | **0** |
| `accountStateAuthCurrentFunction` | 1 | **0** |
| `accountStateTombstoneBalanceZeroFunction` | 1 | **0** |

And the live siblings a name-based sweep would have taken are all still there: `account_state_created_count` (7 files), `account_state_delete_count` (13), `account_state_overflow` (13), `accountStateFindFunction` (2), `accountStateAppendPendingFunction` (2).

⭐ **That last pair is the one worth stating plainly, because two agents measured different populations and both were right.** They are **absent from the guest image** and **present in probe units** (`CreateCodeEffectLog.lean:988-991`, `NonstorageEffectLog.lean:327-330` — 4 and 2 concatenations respectively), and they are shared with the live record-code and record-nonstorage arms. Guest-image absence did not license their deletion, and the keep decision was correct.

## The two evidence classes, which is the reusable part of this PR

| items | evidence | `.text` |
|---|---|---|
| `bal_storage_whitelist_clean` (`0x174`), `brpsf_addr20_eq` (`0x34`, exclusive to it) | `nm` on the **linked** guest image | predicted `0x1a8`, measured `0x1a8` |
| the three AccountState defs | **byte-identical rebuild** — candidate sha unchanged at `1314cfd35` | 0 |

⭐ The second row is the stronger of the two. "Reference count 1, therefore in no build unit" is an inference; **an unchanged ELF sha after deletion is a measurement.** That is the right way round, and it was checked before the PR rather than asserted after.

⭐ **And the prediction held across a rebase, which cross-checks an independent measurement.** Parent went `0x55c18 → 0x552b4`, a difference of `0x964` — exactly the measured drop from #11687, *including* its two 4-byte branch relaxations. Two independently measured cuts agreeing to the byte through a shared base retroactively confirms that 8-byte relaxation was real rather than a counting artefact.

## Why this prediction was exact where #11667's was not

One contiguous region, no interior stubs to preserve — so no boundary to get wrong and no target pulled into branch range by a gap. #11667 was two non-contiguous ranges with four preserved stubs and moved 8 bytes more than predicted, both sites named. **The shape of the cut predicts the precision of the prediction**; the earlier discrepancy was not sloppiness and this exactness is not luck.

## The 209-case sweep was correctly not re-run

The sweep result is a property of the ELF. The scope addition left the ELF byte-identical, so the earlier result (209 cases, 0 faults, 0 output differences, empty flip set, F→P = 0, P→F = 0) transfers unchanged. Stating that reasoning rather than just the conclusion is what makes it checkable.

## Closes #11414 behind it

The last two defs are the two readers named in #11414, the flags-value-8 documentation contradiction. Both were unreferenced, and the **live** delegation read is on a different word in a different structure — `account_writes` **+96** (base `0xa28a0000`, stride 128) then `andi 8` — not AccountState flags **+88**. So that issue's "path disjointness" safety argument was **vacuous rather than fragile**: there is no helper for a future caller to attach to. It resolves by deletion, not by rewriting a comment, which is why documenting the defs in place would have been the wrong fix.

⚠️ One naming hazard from that analysis worth preserving somewhere, since the file it describes is going: `CreateCodeEffectLog.lean:422` interleaved both naming conventions in a single sentence — "bit 16 / bit 4 / bit 2 / bit 8" as **values** and "bit 6" as an **index** (the emitted mark being `ori t1, t1, 64`). The tell is unanswerable: **6 is not a power of two**, so one of the five numbers could not be a mask.

## Scope note, correctly observed

This is **not** the AccountState overlay retirement — #11638 did that. This removes two orphaned guest routines and three orphaned defs that the retirement left behind.

PR #11700, opened 2026-08-07
fix(codegen): remove orphaned fail-65 identity sink (#11431)

pirapira:
## Summary

Closes the #11431 source gap left after #11691: fail-65 was **behaviourally** gone but the sink + data constants were still in the image, and the gate def was orphaned source.

**Honest claim (not "bail removed from image as if #11691 left it"):**
- The identity **CHECK** left the image with #11691 (`blockVerdictMtxGateSystemStorageSeed` is a private def that was **never concatenated** after the seed became `jal ra, process_block_start_system_transactions`).
- This PR removes: orphaned **source** (the dead gate def) + still-emitted **sink** (`.Lbv_syscode_identity_fail` / `li t0, 65`) + **64 B .data** (`cd_canonical_2935_code_hash`, `cd_canonical_4788_code_hash`).

## Cut (complete closure)

| Item | Location | Fate |
|------|----------|------|
| `blockVerdictMtxGateSystemStorageSeed` | `BlockVerdictMtxRuntime.lean` | deleted (private, zero refs) |
| `.Lbv_syscode_identity_fail` | `BlockVerdictReceiptsTail.lean` | deleted (was emitted, unreachable) |
| `cd_canonical_2935_code_hash` / `cd_canonical_4788_code_hash` | `BlockVerdictDataSection.lean` | deleted |

## KEEP (not cut)

- `bsr_sys_acct` — left in place (coord: not worth risk)
- `cd_empty_code_hash` — many live users

## Predicted → measured extents

| section | before | predicted | measured | delta |
|---------|--------|-----------|----------|-------|
| `.text` | `0x54fc0` | `0x54fa8`–`0x54fb0` (−0x10..−0x18, sink only) | **`0x54fac`** | **−0x14** (mid-range; 5-insn sink) |
| `.data` | `0x53b0` | `0x5370` (−0x40) | **`0x5370`** | **−0x40 exact** |
| `.bss` | `0x19029540` | unchanged | `0x19029540` | 0 |

ELF sha moved (`7ea09314…` → `75e2daf0…`). Pins regenerated; **second `gen-region-map-link-pins` pass no-op**. SAsm decide pins for shifted `.data` constants updated (−0x40).

## Evidence (coord checklist)

1. **nm:** no `cd_canonical_2935_code_hash`, no `cd_canonical_4788_code_hash`. `bsr_sys_acct` and `cd_empty_code_hash` remain.
2. **emitted `.s`:** no `.Lbv_syscode_identity_fail`; no `li t0, 65; … bv_fail_code`.
3. **Fail codes the image can still produce** (from `li t0, N; la t1, bv_fail_code` in emitted `.s`):
   `1,2,3,4,5,6,7,10,11,12,13,14,15,16,17,19,20,21,22,23,24,25,26,27,28,29,30,31,32,33,34,35,36,37,39,40,41,44,45,47,48,49,50,51,52,53,54,55,56,57,58,60,61,62,63,64,66,67,68,69,70,71,72`
   **65 is absent.**
4. **Docs/PROGRESS/DRIFT/runners:** no reference to fail-65 or `syscode_identity` on main — no doc follow-up outside emitted source (checked per coord inventory).
5. **r200** (spike, seed=932419178) vs main baseline (`7ea09314…`):
   - base FA=0 FR=22 agree=178
   - candidate FA=0 FR=22 agree=178
   - **F→P=0 P→F=0**; **byte-identical** on all 198 joined cases
6. **Named controls:** honest-fr **PASS** (succ=1); crafted-fa **bv1** (`bv_fail_code=1`, not 65).

## Test plan

- [x] `lake build` green
- [x] pins `--check` CLEAN; second pass no-op
- [x] nm / `.s` / fail-code inventory as above
- [x] r200 A/B vs main
- [x] honest-fr + crafted-fa

Refs: #11431 #11691


pirapira:
**coord review: content approved.** Holding `merge!` until `build`/`source-checks` report. One verification worth stating up front, because the file list invites a question.

## ⭐ The 14-file diff is the exact inverse of #11446, and that is the strongest evidence in the PR

Seven crypto SAsm files plus `AccountIsEip161EmptyClose4.lean` appearing in a fail-65 deletion looks wrong until you read one of them:

```
- rw [show GuestAddrs.bnf_le_a = 2734690080 from rfl]
+ rw [show GuestAddrs.bnf_le_a = 2734690016 from rfl]
```

**Exactly 64 bytes** — the two 32-byte canonical hash constants leaving `.data`. These are address-pinned `rfl` proofs shifting because `.data` shrank, regenerated rather than hand-edited.

⇒ **And #11446's changed-file list is identical to this one.** The PR that *added* those 64 bytes touched the same fourteen files; this PR removes them and touches the same fourteen. A wrong extent would not reverse cleanly across eight independent pinned proofs, so the round trip is a strong check on the cut — please state it in the body, so a reviewer meeting `Bn254FieldMulModSAsm.lean` in a fail-65 PR does not have to derive the explanation.

## The honest claim is the right one and I want it preserved verbatim

> The identity **CHECK** left the image with #11691 (`blockVerdictMtxGateSystemStorageSeed` is a private def that was **never concatenated** after the seed became `jal ra, process_block_start_system_transactions`). This PR removes orphaned **source** + still-emitted **sink** + **64 B .data**.

That distinction was mine to get wrong and grok's to correct: I had described the bail as "unreachable but still emitted", which is true of the sink and **false of the gate**. The compares stopped being emitted the moment #11691 orphaned the def. The PR says so rather than claiming a larger removal.

## Closure, as enumerated and confirmed

| item | location | evidence |
|---|---|---|
| `blockVerdictMtxGateSystemStorageSeed` | `BlockVerdictMtxRuntime.lean` | private, **1 reference — its own definition** |
| `.Lbv_syscode_identity_fail` | `BlockVerdictReceiptsTail.lean` | was emitted, unreachable (only in-edges were the two dead compares) |
| `cd_canonical_{2935,4788}_code_hash` | `BlockVerdictDataSection.lean` | referenced only from the dead gate |

**KEEP, correctly:** `bsr_sys_acct` (defined elsewhere, four other references) and `cd_empty_code_hash` (many live users). Neither was in scope and neither moved.

## No documentation follow-up is owed, and that was checked

There is **no** reference to fail code 65 or to `syscode_identity` in `docs/`, `PROGRESS.md`, `DRIFT.md`, or the runner scripts on current main. A fail code vanishing from the image while a runner label array still named it would mislabel future output — worth one line saying it was checked, since a reader will wonder.

## Why this PR exists at all

⭐ #11431 can now close on a **source** fact rather than a behavioural one. Its predecessor's body said "fail-65 removed" while a grep contradicted it; the goal is all-EEST-pass with **all bails removed**, and a bail that is unreachable but still emitted is precisely the shape that reads as live to the next reader — proven twice today, on the `SimpleTransferPublish` cluster and the `account_state` orphans.

PR #11702, opened 2026-08-07
chore(codegen): delete zisk_sstore_clear_gas_probe (#11355 step 1)

pirapira:
Closes step 1 of #11355. **The guest image is byte-identical** — see the sha below.

## Why the probe goes rather than gets re-fed

It existed because gdb-on-spike was not available when it was written, so a standalone build unit was the only way to observe post-dispatch env state. It is available now, and a breakpoint on the production guest reads the same three words (`env+568` gas_left, `env+448` persistent log count, a status) from **the code that ships**.

## The regression value is redundant, and that was measured rather than assumed

The defect it pinned — BAL preload keys staged big-endian and therefore invisible to the little-endian exec-log scan, showing as `gas_left 25200` where the spec charges 151050 — is caught by the **receipts-root** comparison on real corpus rows.

**The detector chain, traced in the emitted image:**

```
.s 27549  mv x11, x25                 <- a1 = bvgr_gas_left
.s 27552  jal x1, tx_gas_result_increments
          .s 26538  sub x5, x10, x11        <- used_before_refund = gas_limit - GAS_LEFT
          .s 26544  sub x29, x5, x28        <- used_after_refund
          .s 26548-26550  x31 = max(x29, calldata_floor)
          .s 26553  mv x12, x31             <- RECEIPT gas increment
.s 26560  sd x12, 0(bvgr_receipt_gas_increments + 8*i)
```

⇒ the receipt gas increment is an arithmetic function of `gas_left`, so an undercharge changes the receipt RLP and the receipts root (`bv_fail = 53`).

**⭐ The saturation check, which is what makes this a measurement.** A quantity that is compared but *saturated* cannot catch anything either. The `max` against the EIP-7623 calldata floor could have made receipt gas insensitive to `gas_left`. It is **not binding** on the eight valid `multi_transaction_gas_accounting` rows:

| | value |
|---|---|
| header `gasUsed` | 160,050 – 160,120 |
| summed EIP-7623 floor (`21000 + 10 × tokens`) | 42,000 – 42,040 |
| margin | **≥ 118,010** |

A 25,200-gas undercharge leaves used near 134,850 — still ~93,000 above the floor — so the max stays on the `used` side in **both** the correct and the buggy case.

**And the detector demonstrably fires, on a delta 44× smaller:** #11145 is a real corpus row where **570 gas** of undercharge produced `bv_fail = 53`.

⚠️ **The header check is NOT a second detector.** `header.gas_used` is a `max` over the regular and state dimensions, so a single-dimension error is invisible whenever the other dimension is larger — exactly how #11145's 570 gas escaped it (`bv_exact_block_status = 0` on that row).

⚠️ **Population correction:** the fixture holds 16 test cases but only **8** discriminate. The other 8 are `exceed_block_gas_limit_True` **expected-invalid** blocks (`expectException: BlockException.GAS_USED_OVERFLOW|TransactionException.GAS_ALLOWANCE_EXCEEDED`, no `blockHeader`); the oracle wants a reject and gets one either way. I am not calling them blind — an undercharge lowers gas used and could flip an expected-reject into an accept, a succ-bit mismatch that *would* be caught — but whether it crosses each row's limit is uncomputed, so they are neither counted nor written off.

## What was deleted

`ziskSstoreClearGasProbeUnit` with its data symbols (`scgp_ctx`, `scgp_recip`, `scgp_code`, `scgp_preload`, `scgp_exec`, `scgp_out`), the registry name in `RegistryNamesTail`, and the dispatch arm in `RegistryTail`. `Dispatch.lean`'s probe-bundle cross-reference is repointed at `ziskDeriveWithdrawalRequestsProbeUnit`, which bundles the same dispatcher/frame-helper closure — a deleted name must not be left dangling in prose.

⛔ **No check script was removed, because none exists.** There is no `scripts/codegen-zisk-sstore-clear-gas-check.sh`, and no script under any name mentions `sstore_clear_gas` or `sstore-clear-gas`. The probe was even less wired than the issue assumed.

## Byte-identity, predicted before the build

Predicted from `scgp_` having **zero** occurrences in the emitted guest `.s`, then measured:

```
ELF sha256  fab69e61317ce766a09e8a17968b94268dfad6eaeb9fcc70dc87470c98b7b0a1   (before AND after)
emitted .s  byte-identical
.text 0x055c08   .data 0x0053b0   .bss 0x19029520   all unchanged
```

⇒ **no pin regeneration is needed**, and a corpus sweep **cannot** differ: an identical ELF produces identical verdicts on every row, which is a stronger statement than any sweep result.

## Gates

- `lake build` clean — 4,735 jobs, including `RegistryInvariants` (which would fail on a registry inconsistency)
- `scripts/check-build-units-link.sh` — **OK**; every expected-linkable unit still produced an `.elf` and every allowlisted unit is still genuinely non-linkable
- `scripts/check-region-map.sh` — **OK**; region map matches the linked ELF, pins unchanged

## Not in this PR

Step 2 (the M24 legacy storage-preload parser at `Dispatch.lean:2535/:2567/:2581/:2648`) is **gated** on first enumerating the standalone opcode-runtime consumers of the M24 format. A format that standalone tests still feed must not be deleted.


pirapira:
**coord review: content approved.** Holding `merge!` until `build`/`source-checks` report. And one correction to a "fact" I supplied — it was wrong, and this PR is right not to act on it.

## ⭐ Byte-identity predicted before the build, then measured

`scgp_` has **zero occurrences in the emitted guest `.s`**, so `textSizeBytes`, `dataSizeBytes` and the ELF sha were all predicted **unchanged** — and measured unchanged: sha256 `fab69e61317ce766a09e8a17968b94268dfad6eaeb9fcc70dc87470c98b7b0a1` **before and after**, emitted `.s` byte-identical, `text 0x055c08`, `data 0x0053b0`.

That is the right order. A prediction made before the build is evidence; the same number afterwards is a description. Same discipline as #11695's three no-build-unit defs, where an unchanged sha proved what a reference count had only suggested.

## The `Dispatch.lean` line is a dangling-reference fix, not scope creep

A reviewer will ask why a probe deletion touches `Dispatch.lean`. One line:

```
- bundle (see `ziskSstoreClearGasProbeUnit`). Kept OUT of
+ bundle (see `ziskDeriveWithdrawalRequestsProbeUnit`). Kept OUT of
```

A cross-reference retargeted to a surviving unit that shares the same bundling arrangement. Correct — deleting a name and leaving a docstring pointing at it is the failure #10527 had to avoid too.

## ⛔ Correction: the check script I told you to delete does not exist on `main`

I listed `scripts/codegen-zisk-sstore-clear-gas-check.sh` as part of the cut, and cited it as evidence the probe was "not CI-gated". **That file is not tracked on `origin/main`.** It exists only in my working tree — a long-lived stale branch — dated 12 Jul.

So I read a path from my own checkout for a review claim, which is precisely the failure mode I have on record and have corrected in others twice today. This PR correctly does not delete a file that isn't there.

⚠️ **What that does and does not change:** the *conclusion* stands — the probe is not invoked by any workflow — but it stands for a stronger reason than I gave. There is no check script on `main` at all, so there was never anything wired in. My "one of 687 `codegen-zisk-*-check.sh` scripts" count was also read from the working tree and should not be cited.

## The reason the probe is retirable, restated on evidence

- The probe existed because **gdb on spike was not available** when it was written; a standalone build unit was the only way to observe post-dispatch state. `SPIKE_BREAK_PC` / `SPIKE_WATCH` / `SPIKE_DUMP_RANGES` now read the same three words — `gas_left` at `env+568`, `persistent_log_count` at `env+448`, status — **from the production guest** rather than a hand-bundled dispatcher closure.
- Its regression-pin role is **discriminated by the corpus**, and that was traced rather than assumed: `bvgr_gas_left` → `used_before_refund = gas_limit − gas_left` (`26538`) → refund cap (`26540`) → `used_after_refund` (`26544`) → `bvgr_receipt_gas_increments` (`26560`). So receipt gas is an arithmetic function of `gas_left`; a 25,200 undercharge moves the receipt RLP and the receipts root, which is `BV_FAIL 53`.
- ⭐ **Saturation was excluded, which is what makes that a measurement**: the `max` at `26549-26550` against the EIP-7623 calldata floor could have made receipt gas insensitive to `gas_left` even though the root is compared. Floor 42,000–42,040 against header `gasUsed` 160,050–160,120 leaves ≥118,010 of margin, so the max stays on the used side in both the correct and buggy cases.
- **The detector demonstrably fires on a delta 44× smaller**: #11145, a real corpus row where 570 gas of undercharge produced `BV_FAIL 53` receipts-root — a confirmed false reject.

⚠️ And the header check is **not** a second detector: `header.gas_used` is a max over the regular and state dimensions, so a single-dimension error is structurally invisible when the other dominates — which is how #11145's 570 gas escaped it with `bv_exact_block_status` 0.

## Step 2 remains gated

The M24 legacy storage-preload parser (`Dispatch.lean:2535`, `:2567`, `:2581`, `:2648`) has a second consumer that is not this probe — standalone opcode-runtime tests outside the block-verdict path. That enumeration comes before any cut, and if any consumer remains, #11355 narrows to step 1 only.

Issue #11699, opened 2026-08-07
FA lead (reachability unestablished): the RLP walker accepts nested-list items with malformed interiors that decodeAux rejects


pirapira:
Found by dhsorens while discharging `RlpItemDecodeBridges` on #11341 (PR #11694). **Reachability is NOT established and no live false-accept is claimed** — this is filed as a lead with the reachability question first, because the shape is FA-ward. (Issue generated by Claude Code.)

## The divergence, stated precisely

`rlpItemDecode`'s two **list** disjuncts (`WalkNext.lean` — short list `0xc0..0xf7`, long list `≥0xf8`) require only that the item's **span fits**. Nothing about the payload.

`decodeAux` (`Decode.lean:63`) additionally **decodes** the payload (`decodeItems nDepth payload`, with leftover required empty) and returns `none` on failure.

⇒ **A list item with a malformed interior satisfies the machine relation and is rejected by the model.** The guest walker accepts input the reference rejects — a **looser-than-reference** shape against `decode_joined_encodings`.

## ⭐ The evidence was already in the tree, as an absence

`WalkDecodeBridge` has `decodeAux_singleByte_*`, `_shortBytes_*` and `_bytes_*`, and **no list bridge at all.** That absence is not an oversight — one cannot be written, because the two sides disagree on exactly the list case.

⚠️ This also corrects an estimate on #11341: the per-item obligation was described as "five mechanical disjuncts". **Three are mechanical; two are impossible in this form.** `RlpItemDecodeBridges` is therefore **false as stated**, not merely unproven — which is a stronger and more useful result than the satisfiability question I raised on #11694.

## Why this needs a reachability answer before anything else

Two open questions, both named by the finder and neither answered:

1. **Do callers validate list interiors themselves**, independently of the walker?
2. **Can a node with a malformed interior affect a committed result** — i.e. reach a hash, a root, or a verdict?

⚠️ **And here is why I do not think (1) is safely assumed.** #11675 wired `rlp_list_count_items` into `mpt_node_kind` to check node arity exactly. **MPT nodes carry embedded child nodes inline when they are under 32 bytes, and an embedded node is a nested list.** So the list disjuncts are live on the path that matters most — node classification — and a branch node with an inline child is an ordinary, non-adversarial shape.

⇒ The screening question is the standard one: **which branches test the walker's return, and does any accept-path depend on it?** A looser walker is only an FA channel if something downstream trusts its verdict rather than re-deriving.

## Bar

[...]

PR #11703, opened 2026-08-07
proof(#11574): the EIP-2537 wire-pad relation, stated not commented

dhsorens:
Third increment on #11574, after #11677 (the `beBytesToNat ⟷ fromBytesBE` bridge) and #11679 (the assertion vocabulary). This one lands the relation that makes a BLS bridge *possible*; the bridges themselves follow in the next PR.

## Why this is a prerequisite and not a comment

@pirapira's reason, and it is the whole point: **left unstated, a BLS bridge relates two different objects while typechecking cleanly.**

- `Bls12.bytes_to_fq` (`EvmAsm/Stateless/SpecRef/PrecompilesBls.lean:78`) rejects any length other than **64** and decodes all 64 bytes.
- `blsgLtP_spec` (`EvmAsm/Codegen/Programs/Bls12G1LtPSAsm.lean:734`) holds `bytesRegion inPtr xs` with `xs.length = 48` and states its post over those 48.

Both sides have type `List Byte → Nat`. A bridge that simply juxtaposed them would elaborate — and be about two different lists.

**The pad is a real guest step, not a modelling convenience.** `scripts/asm-fixtures/bls12KzgG1WireFunction.s`, routine `blsk_g1_wire`:

```asm
  li t2, 16
.Lblsk_g1w_pad:
  sb zero, 0(t1)          # 16 zero bytes
  ...
  li t3, 48
.Lblsk_g1w_copy:
  lbu t4, 0(t2); sb t4, 0(t1)    # then 48 compact bytes
```

per coordinate, at a 64-byte stride. So the wire felt is literally `16 zeros ++ 48 big-endian bytes`, and this PR says what that does to the value: nothing.

## What landed

**1. `Nat.fromBytesBE_zero_prefix`** — `EvmAsm/EL/RLP/Properties.lean`

```lean
theorem Nat.fromBytesBE_zero_prefix :
    ∀ (zs xs : List Byte), (∀ z ∈ zs, z = 0) →
      Nat.fromBytesBE (zs ++ xs) = Nat.fromBytesBE xs
```

No such lemma existed. The file had `_nil`, `_lt`, `_snoc`, `_toBytesBE`, `_eq_zero_iff`, `_pos_of_head_ne_zero`, and nothing about a zero prefix — it is the **non-canonical companion** to `toBytesBE_fromBytesBE_of_canonical`, which needs a non-zero head and so excludes exactly this case. Homed with its siblings; `EL/RLP` is uncapped.

**2. `eip2537_wire_pad_value`** — `EvmAsm/Stateless/Crypto/FieldAssertions.lean`

```lean
theorem eip2537_wire_pad_value (w : List (BitVec 8)) (hlen : w.length = 64)
    (hpad : ∀ i, i < 16 → w.getD i 0 = 0) :
    EL.RLP.Nat.fromBytesBE w = beBytesToNat (w.drop 16)
```

Deliberately `fromBytesBE` on the left and `beBytesToNat` on the right: those are the two decoders the two sides actually use, so this discharges **the pad and the decoder mismatch in one step**, over `beBytesToNat_eq_fromBytesBE` from #11677. Plus `eip2537_wire_suffix_length` (the suffix really is 48) and `eip2537_wire_pad_specref`, the same statement at `bytesBEtoNat` — definitionally the first one, stated anyway so the connection is machine-checked rather than requiring a reader of `bytes_to_fq` to unfold an abbrev to believe the bridge applies.

## ⭐ BN254 gets no counterpart, and the asymmetry is recorded

`Bn128.bytes_to_g1` (`SpecRef/PrecompilesCurve.lean:83`) reads `bytesBEtoNat (data.take 32)` directly, against a guest routine scanning the same 32 bytes. No pad on that side, so **nothing to relate** — the module records the asymmetry rather than writing a vacuous BN254 twin to make the two families look symmetric.

## Placement

`FieldAssertions.lean` rather than `Crypto/BeBytesBridge.lean`: the pad relation is about the *wire encoding of a field element*, which is this module's subject, and it already imports SpecRef. Putting it in `BeBytesBridge` would have pushed the whole RLP `Properties` tower onto every future consumer of the decoder equality, for no gain.

## Verification

- `lake build` clean — 4736 jobs.
- `check-file-size` (2173 files, all within cap), `check-forbidden-tactics` (OK), `check-unimported` (2766 modules, 0 orphans), `check-layering` (invariants hold; the four `Rv64 → Evm64` advisories are pre-existing and untouched).
- `#print axioms`, fully qualified:

```
'EvmAsm.EL.RLP.Nat.fromBytesBE_zero_prefix' depends on axioms: [propext]
'EvmAsm.Stateless.Crypto.eip2537_wire_pad_value' depends on axioms: [propext, Quot.sound]
'EvmAsm.Stateless.Crypto.eip2537_wire_suffix_length' does not depend on any axioms
'EvmAsm.Stateless.Crypto.eip2537_wire_pad_specref' depends on axioms: [propext, Quot.sound]
```

Inside the classical three, and one better — no `Classical.choice`.

No registry movement: this PR adds no witnessed spec. The two absent `Progress/*` rows for `blsgLtP_spec` / `bnfLtP_spec` land with the bridges.

## Next on #11574

The two bridges (`blsgLtP_spec` → `bytes_to_fq`, `bnfLtP_spec` → `bytes_to_g1`) with the modulus constants as theorems rather than `#guard`s, then the registry rows plus `docs/crypto-spec-correspondence.md`.

⚠️ One thing I will not claim there: `lt_p` returns a boolean, not the field element, so **predicate agreement is all it can support**. Value agreement is not available from this routine.

### DIFF
--- a/EvmAsm/EL/RLP/Properties.lean
+++ b/EvmAsm/EL/RLP/Properties.lean
@@ -166,6 +166,29 @@ theorem Nat.toBytesBE_fromBytesBE_of_canonical :
     have hmod : (Nat.fromBytesBE xs * 256 + b.toNat) % 256 = b.toNat := by omega
     rw [hdiv, hmod, ihxs, ofNat8_toNat]
 
+/-- Leading zero bytes do not change the big-endian value: each contributes
+    `0 * 256 ^ _`. This is the *non-canonical* companion to
+    `toBytesBE_fromBytesBE_of_canonical` — that lemma needs a non-zero head, this
+    one is about exactly the case it excludes.
+
+    Motivation (#11574): EIP-2537 transmits a BLS12-381 base-field element as a
+    **64-byte** wire felt whose first 16 bytes are zero, while the guest's
+    `blsg_lt_p` scan reads only the **48** compact bytes. Without this lemma the
+    two sides are different lists and any bridge between them would relate
+    different objects while typechecking cleanly. -/
+theorem Nat.fromBytesBE_zero_prefix :
+    ∀ (zs xs : List Byte), (∀ z ∈ zs, z = 0) →
+      Nat.fromBytesBE (zs ++ xs) = Nat.fromBytesBE xs
+  | [], _, _ => by simp
+  | z :: zs, xs, hz => by
+    have hz0 : z = 0 := hz z (List.mem_cons_self ..)
+    have he : Nat.fromBytesBE ((z :: zs) ++ xs)
+        = z.toNat * 256 ^ (zs ++ xs).length + Nat.fromBytesBE (zs ++ xs) := rfl
+    have hrec : Nat.fromBytesBE (zs ++ xs) = Nat.fromBytesBE xs :=
+      Nat.fromBytesBE_zero_prefix zs xs fun x hx => hz x (List.mem_cons_of_mem _ hx)
+    rw [he, hrec, hz0]
+    simp
+
 /-! ## takeBytes properties -/
 
 /-- Taking 0 bytes always succeeds with an empty prefix and the original list. -/
--- a/EvmAsm/Stateless/Crypto/FieldAssertions.lean
+++ b/EvmAsm/Stateless/Crypto/FieldAssertions.lean
@@ -39,6 +39,7 @@
 -/
 
 import EvmAsm.Crypto.BeBytesBridge
+import EvmAsm.EL.RLP.Properties
 import EvmAsm.Rv64.SAsm.AccelStep
 import EvmAsm.Stateless.SpecRef.PrecompilesBls
 import EvmAsm.Stateless.SpecRef.PrecompilesKzg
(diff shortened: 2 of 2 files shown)


pirapira:
**coord review: content approved.** Awaiting `build`/`source-checks`; will label when green. Third increment on #11574 after #11677 and #11679, and the premise checks out against the tree.

## The length mismatch is real, and it is exactly as stated

`SpecRef/PrecompilesBls.lean:78-83` on `origin/main`:

```lean
/-- `bytes_to_fq(data)`: 64-byte big-endian, `< p`. -/
def bytes_to_fq (data : Bytes) : Except EvmError Nat := do
  if data.length ≠ 64 then throw (.invalidParameter "FQ should be 64 bytes long")
  let c := bytesBEtoNat data
  ...
```

**Exactly 64, decoding all 64.** And `blsgLtP_spec` holds `bytesRegion inPtr xs` with `xs.length = 48`, stating its post over those 48. Both sides are `List Byte → Nat`, so a bridge that juxtaposed them **would elaborate and be about two different lists** — which is the failure mode this PR exists to prevent, and the same one #11677 addressed for the decoder.

⭐ **And the pad is a guest step, not a modelling convenience.** `bls12KzgG1WireFunction.s`, `blsk_g1_wire`: `li t2, 16` then `.Lblsk_g1w_pad` writing 16 zero bytes, then `li t3, 48` and `.Lblsk_g1w_copy` copying 48 compact bytes, per coordinate at a 64-byte stride. So the wire form is literally `16 zeros ++ 48 big-endian bytes`, and this PR states what that does to the value — nothing. That is a fact about emitted code, not about a definition, which is what makes it load-bearing.

## `Nat.fromBytesBE_zero_prefix` fills a real gap

The claim that no such lemma existed is checkable and holds: `Properties.lean` had `_nil`, `_lt`, `_snoc`, `_toBytesBE`, `_eq_zero_iff`, `_pos_of_head_ne_zero` — and nothing about a zero prefix. Stating it over `∀ z ∈ zs, z = 0` rather than over a concrete 16 is the right generality: the 16 is a wire-format detail, the fact is about zero prefixes.

No `sorry`, no `admit`, no forbidden tactic in the diff.

## Sequencing note, and it is a compliment

This is the third PR in a row on #11574 that landed a **prerequisite** rather than the headline result, each time because the headline would otherwise have related two different objects while typechecking. #11677 bridged the two decoders, #11679 established the assertion vocabulary and corrected the base-field-vs-scalar-field prime, and this one states the pad relation. ⭐ **The bridges being "next PR" is not slow progress — it is the only order in which the bridges mean anything.**

⚠️ One thing to carry into that next PR, unrelated to this diff: #11699 records that the analogous per-item bridge on #11341 turned out **false as stated** for two of five disjuncts, because the machine relation and the model disagree on the list case. Worth checking early whether the BLS bridge has any comparable disjunct where the two sides differ in *admissibility* rather than in *value* — that is the class of gap a length-and-padding lemma cannot close.

PR #11704, opened 2026-08-07
refactor: make deposit requests receipt-only

pirapira:
## Summary

Align the EIP-6110 request producer with execution-specs. At pinned execution-specs e5a8caf1b, parse_deposit_requests in src/ethereum/forks/amsterdam/requests.py:274-284 is the only producer: it scans receipt logs for DepositEvent records. fork.py:71 imports it and fork.py:957,959-960 appends the result only when non-empty; there is no calldata-derived else producer. This removes the guest-only calldata synthesizer and its all-direct classifier.

The deleted route-B code is BlockVerdictDepositFallback.lean (classifier plus append helper), its two BuildUnit links, the ReceiptsTail inlined classifier/append loop, and the comment-only import. The 11440 receipt-status gate goes with route B; it is not undone.

## Evidence

- Base commit: a2348b1db.
- Production guest: base ELF 438a11c4dbebb6f7a4bd80a051439f5e0767d2eefabfb2f0e44c7ff9044af01, candidate ELF 2fef06fdb3cb32f877fdcff0dbf0f7ff02c53d59142780f5c93875cac6a56c5f; .text 0x54e18 -> 0x549f0 (-0x428), .data=0x53b0 and .bss=0x19029540 unchanged.
- The 0x428 decomposition is 0xE4 all-direct classifier + 0x290 append helper (nm extents) + 0xB4 (22 deleted inlined ReceiptsTail route-B instructions: dlen branch, tx loop, append calls, length/fallthrough labels).
- The standalone V2 probe also loses 0x428: base text 337092=0x524c4 -> candidate 336028=0x5209c. Direct nm evidence on the V2 base has block_verdict_all_direct_deposit_txs at 0x8005108c and block_verdict_append_direct_deposit at 0x80051170; the candidate has neither. V2's pre-cut emitted .s lines 14669-14687 contain the c1_dlen gate and route loop, so the same 0xE4+0x290+0xB4 decomposition applies. This is emitted evidence; V2's direct imports do not list ReceiptsTail.
- Honest constructed witness 9a6e3c8b0ff5882d0692f2c347139a7c533d09ca42d3ac6b2b4f608c967f07b3: candidate output SHA 30135af1099ba9819d8d34024c5628c2ad58ff961a6c58fac9832d2731f37be3, succ=1,bv_fail=0; this flips the old false reject to accept.
- Constructed false-accept header 3e5a05f02a5f888c0a5e1f5d68679a12681244ae9c58ceaee10e02e7ec43e4eb: candidate output SHA dd2e899aeb2c37d697468dd51d0ba6c76a7178c01dc9aed65d4b8f90ce9c6fa2, succ=0,bv_fail=55; it remains rejected because the receipt-derived list lacks the claimed request, not because a route gate fires.
- Canonical 23040 26a991d26ccc4f39a4f2d2c62b2699218480d1845443a02b48f40166a1d684ba: candidate output SHA 54c88079a453dd89a8ba455eb0d49961431f3bde6361849d9d1428a60794207c, succ=1,bv_fail=0.

Random-200 used manifest SHA ba0225b56376abfcd45cff6c92066280bc0f33a2855c4111eab2d1e199e7592 and spike runner SHA 529504d986b06aa5b40c8e9fc351bd88831f4a30dc28b4ef92422d54ba9f2844 with 30 workers. Both legs scored all 200 rows: base FA=0,FR=21,OK=179,FAULT=0; candidate identical. Base-PASS to candidate-FR, base-FR to candidate-PASS, and reverse/FA flip sets are all empty. FR 21 to 21 is expected: every corpus deposit row goes through parse_deposit_requests, while c1_dlen=0 is the genuine no-recognised-event case, so no corpus row exercises route B's synthesise branch. The sweep therefore proves no regression only; it cannot prove the removed branch works. The constructed witnesses are the evidence that exercises the branch, and both are needed.

Residual risk is FR-ward: the only gap not excluded is a future missing successful receipt snapshot. That would omit deposits from the rebuilt list, cause requests_hash to mismatch, and reject the block (fail-closed). The current guest instead accepted a crafted header by synthesizing calldata requests; this removes that unsound path.

## Gates

- lake build (4735 jobs): clean.
- scripts/check-region-map.sh (second fixpoint pass): clean.
- scripts/check-asm-to-program.sh: clean.
- scripts/check-unimported.sh: 2765 modules, 0 orphans.
- scripts/check-no-warnings.sh: 0 warnings.

Generated TSV, GuestAddrs, and RegionMapLinkPins were regenerated to fixpoint; no hand-pinned layout values. Live closure symbols retained: bv_mtx_skip_ctx, pdr_deposit_addr, pdr_deposit_sig, and edd_be32_eq.


pirapira:
**coord review: content approved.** Holding `merge!` until `build`/`source-checks` report; the other four pass. This is a **structural convergence**, so it is reviewed against that bar rather than byte-equivalence.

## The divergence, and why this is convergence rather than a fix

`parse_deposit_requests` (`requests.py:274`) is the **only** deposit producer in execution-specs — it walks receipts and takes a log as a deposit iff it originates from `DEPOSIT_CONTRACT_ADDRESS` **and** its first topic matches `DEPOSIT_EVENT_SIGNATURE_HASH`. `fork.py:71` imports it, `:957` calls it, `:959-960` appends only when non-empty. **No second producer, no `else` branch.**

The guest had two: route A (retired — `bv_deposit_runtime_capture_complete` is `sd zero`-only, *"#11183 RESERVED always-0"*) and route B, which synthesised a request from **calldata shape**. This PR removes route B and makes the receipt the source, matching the spec's single producer.

⭐ **And the comment fix matters as much as the code.** The old text called route B *"the spec-shaped path (parse logs, else derive from tx calldata)"* — a false claim of spec alignment sitting at the callsite, and plausibly why this survived review. The replacement states the actual rule: *"The specification appends a type-0 request only when the receipt-log scan found a matching DepositEvent. An empty derived body is therefore the normal no-deposit case; there is no calldata-derived alternate producer."*

## The bar: all three witnesses, with the FA header's reason stated

| input | before | after |
|---|---|---|
| honest `9a6e3c8b…f07b3` | REJECT (`bv_fail=55`) — specs **ACCEPT** | ✅ **succ=1, bv_fail=0** |
| FA header `3e5a05f0…3e4eb` | **ACCEPT** — specs **REJECT** | ✅ **succ=0, bv_fail=55** |
| canonical 23040 `26a991d2…` | pass | ✅ pass |

⭐ **`bv_fail = 55` is the same code before and after, so the code alone shows nothing** — the reason is what shows the mechanism changed: the FA header is now rejected because **the rebuilt receipt-derived list lacks the claimed request**, not because a gate fired. Requiring that reason is what distinguishes a right verdict from a right verdict for the wrong cause.

## r200: FR-neutral, and that is the expected result

Base and candidate both **FA=0, FR=21, OK=179, FAULT=0**; **base-PASS → candidate-FR empty**; no flips either way; labels identical **200/200**; both legs from `a2348b1db` with named ELF shas.

⚠️ **FR 21 → 21 is not a change that bought nothing**, and the PR should be read with both halves together: the four-path census established that `c1_dlen == 0` means genuinely no recognised deposit event, and every corpus deposit row goes through the parser — so **no corpus row exercises route B's synthesise branch at all.** It follows that r200 can only prove no regression, and **the constructed witnesses are the only possible evidence that the fix works.**

## Why the witnesses had to be constructed

Neither nearest fixture has both halves of the intersection: **23061/23062** show non-canonical deposit code succeeding without emitting an event, but their transaction has `data_len = 0` so the shape match never fires; **23040** has canonical 404-byte calldata but canonical code, which emits the log and bypasses route B. ⇒ No sweep of 26,104 rows could have found this.

## ⭐ The residual risk is FR-ward, which is the argument this lands on

The one gap the census could not exclude is a future or missing successful-snapshot path. If that happens: deposits go missing → the rebuilt list is short → `requests_hash` mismatches → **the block is rejected.** Fail-closed. Contrast the pre-change state, where a crafted header claiming a synthesised request was **accepted**. The change moves the failure mode from accept to reject.

## Extents, reconciled across both images

`0x374` (helpers, `nm` extents: all-direct `0xE4` + append `0x290`) + `0xB4` (the inlined ReceiptsTail call-site loop, 22 deleted lines) = **`0x428`**, and both images shed exactly that: guest `0x54e18 → 0x549f0`, V2 `337092 → 336028`.

⚠️ **A note for the next reader, because I raised a false alarm here myself:** `BlockVerdictV2.lean`'s direct imports do **not** include `ReceiptsTail`, so grepping them suggests V2 cannot contain the route loop. It does, transitively via `BlockVerdictFunction`. The evidence that settles it is emitted, not inferred — `nm` on the V2 base (`block_verdict_all_direct_deposit_txs` at `0x8005108c`, `append` at `0x80051170`) and V2's pre-cut `.s` lines 14669-14687 carrying the `c1_dlen` gate. Cite those rather than an import chain.

## Closure keeps, verified

`bv_mtx_skip_ctx` (live in MtxTail and DataSectionTail), `pdr_deposit_addr` / `pdr_deposit_sig` (needed by the parser), `edd_be32_eq` (needed by ExtractDepositData). None moved.

⚠️ And state plainly, since it is easy to misread as an undone fix: **#11440's receipt-status gate goes with route B.** It was the correct fix for the reverted-call FA while a calldata producer existed; with no such producer there is nothing for it to gate.

PR #11705, opened 2026-08-07
proof(#11574): the two lt_p bridges, the crypto family, and its rows

### LINKED ISSUE #11574: predicates: assertion vocabulary for the precompile field towers (BLS12-381/BN254 limb buffers, point records) — the named prerequisite that cut blsg_lt_p/bnf_lt_p from the leaf table

dhsorens:
**Closes #11574.** Stacked on #11703 (review that first — this branch contains its two commits).

## ⚠️ The premise correction first

**The two triples this issue asks for already existed, `sorry`-free, and predate the issue by months.** `blsgLtP_spec` (`Programs/Bls12G1LtPSAsm.lean:734`) and `bnfLtP_spec` (`Bn254FieldLtPSAsm.lean:732`) are whole-routine triples with genuine numeric posts. A name search for the *routines* found nothing because the specs live in sibling `*SAsm` modules — the #10779 lesson recurring. `docs/leaf-routine-targets.md:98` already carries this correction from #11676.

What was genuinely missing: the SpecRef vocabulary to read their `a0` in, and the registry rows. **Both specs were unwitnessed by `check-axioms.sh` until this PR.**

## What landed

`EvmAsm/Codegen/Programs/CryptoFieldLtPBridge.lean` (new):

**The constants as theorems.** `Bls12G1LtPSAsm.lean:85` pins its prime with a `#guard` against a hex literal — which evaluates and yields **no reusable term**. `bls12PBytes_eq_blsP` and `bn254PBytes_eq_fieldModulus` are the same facts as proof terms, stated against the *SpecRef names*, so a guest/reference divergence is a build failure rather than two literals nobody diffed. Both `by decide`, both **axiom-free**, and cheap — the whole file elaborates in 1.2s, retiring the plan's worry about a 48-byte `decide`.

**The BLS tie**, over #11703's wire-pad relation: `blsg_lt_p_agrees_bytes_to_fq` → `blsg_lt_p_a0_eq_bytes_to_fq_indicator` → `blsgLtP_spec_specref`, the whole-routine triple with `a0` restated as `bytes_to_fq`'s accept/reject indicator.

**The BN254 tie**: `bnf_lt_p_agrees_field_bound` (plus `bnf_lt_p_rejects_iff_out_of_range`, the reference's own `>=` polarity, so a reader diffing against the Python needn't flip it mentally) and `bnfLtP_spec_specref`.

### Two things deliberately not claimed

⚠️ **Predicate agreement is the ceiling.** `lt_p` returns a boolean in `a0`, never the field element, so value agreement is not available from these routines. Where an `ok` payload appears it is a fact about the **reference alone**, labelled as such.

⚠️ **BN254 is CLAUSE agreement.** `bytes_to_g1` also bounds `y` and tests `y² = x³ + 3`; this routine looks at neither. The row is graded against the `x >= field_modulus` conjunct, named in its `reference` field. Whole-function agreement would be an overclaim.

## ⭐ Why the BLS row is `domainRestricted` — and the sub-finding that got it there

The tie carries `w.length = 64` and `∀ i < 16, w.getD i 0 = 0`. That gate is **load-bearing**: the reference decodes all 64 bytes, so a nonzero pad byte makes the value ≥ 2^384 > p and it rejects, while the guest scans only the 48 compact bytes and would not see it.

So I checked whether the guest over-accepts. **It does not.** Every calldata reader of a wire felt calls `blsg_is_zero_n(ptr, 16)` and rejects on nonzero *before* the scan:

| reader | pad check | reachable via |
|---|---|---|
| `blsg_decode_g1` | `Bls12G1.lean:692-700` (both coordinates) | 0x0b, 0x0c, 0x0f |
| `blsg2_decode_g2` | `Bls12G2.lean:774-784` (all four felts) | 0x0d, 0x0e, 0x0f |
| `zkvm_bls12_map_fp_to_g1` | `Bls12MapG1Real.lean:23-29` | 0x10 |
| `zkvm_bls12_map_fp2_to_g2` | `Bls12MapG2Real.lean:23-38` | 0x11 |

All live — the dispatch table wires 0x0b..0x11 at `PrecompileSharedExecute.lean:136-142`.

⚠️ **But that composition is not a theorem, and cannot be yet.** Those decoders exist only as assembly **strings** — no `Program`, no `_eq_prog` drift guard, no fixture — so nothing is statable over them. Hence `domainRestricted`: a **coverage** gap, not a behavioural one, and the row says which.

The resulting inversion is worth naming on its own: **the range check underneath is a proved routine while the pad guard above it is unverified assembly text.** Converting `blsg_decode_g1` is the prerequisite for regrading this row to `agrees`, and it is the highest-value next step in this family.

## ⚠️ A citation of mine that was wrong

My first draft cited `blsk_g1_wire` — the **writer** — as evidence the pad is a real guest step. It writes wire felts and never sees calldata, so it establishes the wire *layout* but says nothing about whether an attacker-supplied felt satisfies the hypothesis. The reader-side guards above are the load-bearing evidence. Fixed in `69bdd2598` (on the #11703 commit, where the text lives) and recorded rather than silently swapped.

## Registries

| | before | after |
|---|---|---|
| `Routines` rows / proven / conditional / symbols | 29 / 21 / 8 / 21 | **33 / 24 / 9 / 23** |
| `Correspondence` rows | 29 | **31** |
| verdicts: agrees / domainRestricted | 17 / 7 | **18 / 8** |
| basis: ported | 6 | **8** |
| new family | — | `crypto` (2 rows) |

Four `Routines` rows: the machine triple and the model-facing restatement for each routine, mirroring the `rlp_list_encoded_size` pair. The BLS restatement is `.conditional` with the wire-pad gate stated in prose.

**Both `.ported` rows carry the clause table that rung requires**, against the vendored `amsterdam` Python (`bls12_381/__init__.py:426-454`, `alt_bn128.py:39-82`). ⚠️ Per method-doc §6a, the `py_ecc` and CPython clauses are marked **read, not machine-checked** — `check-spec-refs.sh` can verify a `forks/.../x.py:NNN` line and cannot verify these. I did verify both moduli numerically against `py_ecc/fields/field_properties.py:24,29` at the pinned version.

⚠️ **One clause difference found and deliberately NOT recorded as a `portDefect`.** `alt_bn128.py:59` raises on `len(data) != 64`; the Lean `Bn128.bytes_to_g1` has no such check. It is **unreachable** — every port call site passes `buffer_read data k 64` (`PrecompilesCurve.lean:113/115/124`) and `buffer_read` pads to exactly `size` (`Vm.lean:299-301`), so the guard's precondition holds at every call. An unreachable defensive check is not a behavioural divergence, so `port_defect_count` stays 0. Named in the clause table anyway, because a future caller handing it a short list **would** diverge: the port reads a short prefix, the Python a zero-padded 32 bytes, and those are different numbers.

## Docs

- **`docs/crypto-spec-correspondence.md`** (new, §10 rubric). Leads with this family's characteristic failure — the base-field/scalar-order confusion, in a table, since #11574 and `leaf-routine-targets.md` both fell into it.
- **`docs/4ch8f-crypto-kernel-inventory.md:53`** claimed *"No `cpsTripleWithin` / SAsm port connects any crypto asm body to its spec."* **#11574 was filed partly on the strength of that**, and ~40 `*SAsm.lean` triples falsify it. Corrected to the accurate, narrower statement: the *accelerated* kernels have no body-level triple; the scan and bound helpers around them are proved.

## Verification

- `lake build` clean — 4737 jobs.
- `check-axioms.sh` — **146 witnesses, classical only**, 0 grandfathered owners.
- `check-registry-crosscheck.sh` (23 witnessed symbols × 31 rows) and `--self-test` (gate fires).
- `check-correspondence-deps.sh` (20 modules, budget 80), `check-routine-liveness.sh` (385 routines, 384 live, 1 annotated), `check-progress.sh` after `progress-report.sh --write`.
- `check-file-size`, `check-forbidden-tactics`, `check-unimported` (2767 modules, 0 orphans), `check-layering`.

`#print axioms`, fully qualified:

```
'EvmAsm.Codegen.bls12PBytes_eq_blsP' does not depend on any axioms
'EvmAsm.Codegen.bn254PBytes_eq_fieldModulus' does not depend on any axioms
'EvmAsm.Codegen.bytes_to_fq_of_length_eq' depends on axioms: [propext, Quot.sound]
'EvmAsm.Codegen.blsg_lt_p_agrees_bytes_to_fq' depends on axioms: [propext, Quot.sound]
'EvmAsm.Codegen.blsg_lt_p_a0_eq_bytes_to_fq_indicator' depends on axioms: [propext, Quot.sound]
'EvmAsm.Codegen.blsgLtP_spec_specref' depends on axioms: [propext, Classical.choice, Quot.sound]
'EvmAsm.Codegen.bnf_lt_p_agrees_field_bound' depends on axioms: [propext]
'EvmAsm.Codegen.bnf_lt_p_rejects_iff_out_of_range' depends on axioms: [propext]
'EvmAsm.Codegen.bnfLtP_spec_specref' depends on axioms: [propext, Classical.choice, Quot.sound]
```

The two restated triples inherit `Classical.choice` from the underlying specs; nothing else does.

No `check-asm-to-program.sh` run is claimed — this PR emits no assembly and converts no string.


pirapira:
**coord review: content approved.** Awaiting `build`/`source-checks`; will label when green. Stacked on #11703 (also reviewed and awaiting CI), so it lands behind it.

## The premise correction is right, and I verified both halves

`blsgLtP_spec` exists on `origin/main` at `Bls12G1LtPSAsm.lean:734` as a whole-routine `cpsTripleWithin` with real hypotheses (`xs.length = 48`, alignment, overflow, `isValidByteAccess`), and its docstring already records that the `la` materialisation of `blsg_p_be` is **proven** via `la_materialize_within` rather than assumed. And `:85` does pin the prime with a `#guard` against a hex literal.

⭐ **So "the routines had no spec" was a name-search artefact — the specs live in sibling `*SAsm` modules.** That is the #10779 lesson recurring, and it is the third time today an issue's premise dissolved on re-derivation rather than on argument. What was genuinely missing is exactly what this PR adds: the SpecRef vocabulary to read `a0` in, the registry rows, and **axiom witnessing** — both specs were unwitnessed by `check-axioms.sh` until now, which is the part that actually changes the trusted state.

## Constants as theorems, not `#guard`s

⭐ **This is the most reusable idea in the PR.** A `#guard` evaluates and yields **no reusable term**; `bls12PBytes_eq_blsP` and `bn254PBytes_eq_fieldModulus` are the same facts as proof terms stated against the **SpecRef names**, so a guest/reference divergence becomes a build failure instead of two literals nobody diffed. Both `by decide`, both axiom-free, 1.2 s for the file — which also retires the plan's worry that a 48-byte `decide` would be too slow.

## ⭐ The two non-claims are what make this reviewable

**Predicate agreement is the ceiling.** `lt_p` returns a boolean in `a0` and never the field element, so value agreement is unavailable from these routines; where an `ok` payload appears it is labelled a fact about the reference alone.

**BN254 is CLAUSE agreement, not whole-function.** `bytes_to_g1` also bounds `y` and tests `y² = x³ + 3`; this routine looks at neither, so the row is graded against the `x ≥ field_modulus` conjunct with that named in its `reference` field. Calling it whole-function agreement would have been the overclaim, and refusing it is the same discipline that kept #11694's row `.machineOnly`.

## Why `domainRestricted` is the honest grade — and the sub-finding behind it

The tie carries `w.length = 64` and `∀ i < 16, w.getD i 0 = 0`, and that gate is **load-bearing**: the reference decodes all 64 bytes, so a nonzero pad byte makes the value ≥ 2^384 > p and it rejects, while the guest scans only the 48 compact bytes and would not see it.

⇒ So the obvious question is whether the guest **over-accepts**, and the PR answers it rather than leaving it: every calldata reader of a wire felt calls `blsg_is_zero_n(ptr, 16)` and rejects on nonzero *before* the scan — `blsg_decode_g1` (`Bls12G1.lean:692-700`), `blsg2_decode_g2` (`Bls12G2.lean:774-784`), and both map precompiles — with the dispatch table wiring 0x0b..0x11 at `PrecompileSharedExecute.lean:136-142`.

⚠️ **And it correctly declines to turn that into a theorem**, because those decoders exist only as assembly *strings* — no `Program`, no `_eq_prog` drift guard, no fixture — so nothing is statable over them. Hence `domainRestricted`: a **coverage** gap, honestly graded, rather than a soundness one dressed up as complete.

⭐ That distinction is worth stating plainly because it is the pattern this whole series has followed: #11677 bridged two decoders that nothing related, #11679 corrected a base-field-vs-scalar-field prime, #11703 stated the pad relation, and this one grades what it can prove and names what it cannot. **Four PRs, none of which overclaimed to reach a headline.**

## One forward note, not a request

#11699 records that #11341's analogous per-item bridge turned out **false as stated** for two of five disjuncts, because the machine relation and the model disagree on *admissibility* rather than on value. The `domainRestricted` gate here is the same class of gap caught early instead of late — worth keeping in view if the pad composition is ever promoted to a theorem, since that promotion needs the decoders to become `Program`s first, and that is the real prerequisite rather than more lemmas.

pirapira:
**Build fix, for when you next touch this** — two lines to delete. The gate is firing *because* the PR succeeded.

```
check-registry-coverage: FAIL — 2 STALE allowlist entries in scripts/registry-coverage-allow.txt:
    blsg_lt_p    now registered -- delete this line
    bnf_lt_p     now registered -- delete this line
```

Those two routines were allowlisted as *not yet registered*. This PR registers them, so the exemptions are stale and the gate says so by design — its own message reads *"The allowlist expires on purpose: an exemption that outlives…"*. Deleting the two lines from `scripts/registry-coverage-allow.txt` is the whole fix.

⭐ **`check-axioms` passed on the same run** — 146 witnesses, only classical axioms (`propext`, `Classical.choice`, `Quot.sound`), zero grandfathered `bv_decide`/`native_decide` owners. That is the gate that matters for this PR, and it confirms the two newly-witnessed specs landed cleanly. `source-checks`, `path-labels`, `size-label`, `summarize` and `velocity` all pass; `build` is the only red, and only on that allowlist check.

⚠️ It has also gone CONFLICTING since main moved to `9b08907c8` (#11700 landed). That half is **not yours** — a labelled or in-flight PR's conflicts are the merge actor's to resolve, and `DIRTY` alone is never a handback. Only the allowlist lines need you.

Content review stands unchanged from my earlier comment: the premise correction is verified, the constants-as-theorems approach is the reusable idea, and the two non-claims (predicate agreement as the ceiling, BN254 as clause rather than whole-function agreement) plus the honest `domainRestricted` grade are what make it reviewable. (Comment generated by Claude Code.)

Issue #11706, opened 2026-08-07
docs(codegen): AccountState flags@+88 and account_writes execFlags@+96 are different words with overlapping bit values, and neither site says so


pirapira:
**Documentation defect with a measured cost.** Two different flag words at two offsets in two structures, with overlapping bit values and no comment at either site saying which is which. It misled two agents in one day; the second time it broke five passing corpus rows. (Issue generated by Claude Code.)

## The two words

| field | offset | structure |
|---|---|---|
| AccountState flags | **+88** | AccountState row |
| account_writes execFlags | **+96** | account_writes row (base `0xa28a0000`, stride 128) |

They carry **different meanings for overlapping values**, and nothing at either offset says so.

## What `+96` actually means, established from emitted masks

Exactly three literals are tested across live readers (`AccountWriteMap.lean`):

| literal | reader | meaning |
|---|---|---|
| **2** | `.Lawa` tx/block (`:380`, `:392`), `.Lawab` (`:411-412`), `.Lawlc` tx/block (`:461`, `:471`) | **live**; zero ⇒ present-dead / deleted |
| **8** | `.Lawc` (`:433`) | **created-this-tx** — and this reader is **live**, called at guest PCs ≈ 49512, 50199, 51190 |
| **16** | `AccountWriteRecord:226`, `AccountWritesBlockUpsert:621` | in the **validMask**, not the payload — gates whether `+96` is stored/copied at all |

Values 1, 4 and 32 have no `+96` reader mask.

## The cost, measured

While fixing #11697, the AUTH writer's missing `a7` was patched with **27/31** — the constants `account_state_record_code` (`CreateCodeEffectLog.lean:407-409`) writes to **AccountState +88** (`li t1, 27`, `ori t1, t1, 4` for nonempty code).

`27 = 1+2+8+16` and `31 = 1+2+4+8+16`. **Both set value 8** — the bit `.Lawc` reads as created-this-tx at `+96`. Result on rows 00612-00623: 00612 recovered and **00618, 00619, 00620, 00622, 00623 newly FR** — five base-PASS rows broken by carrying a `+88` constant into `+96`.

The correct value, derived from `+96`'s readers alone, is **`a7 = 2`**: it satisfies every live AUTH/current-code reader and deliberately omits 8.

## ⭐ Two agents, one day, same pair

[...]

PR #11708, opened 2026-08-07
fix(codegen): honor HAS_BALANCE=0 on apply_deletes (#11688)

### LINKED ISSUE #11688: bug(codegen): header resolve leaves a garbage high balance limb for a CREATE'd account, flipping opt to Present (fixture 01114)

pirapira:
## Summary
`account_writes_apply_deletes` treated map-balance=0 as "unknown" and always re-resolved the parent pre-balance. When `HAS_BALANCE` is set with value 0 (SELFDESTRUCT drain), zero is authoritative — the re-fetch resurrected a pre-seeded CREATE balance (fixture 01114: CREATE pre-exists with bal=0x64) and forced `keep_present` (opt=1), moving the NPR root.

## Fix
After map-bal OR==0, before resolve:
```
ld t1, 112(t0); andi t1, t1, HAS_BALANCE; bnez t1, .Lawd_present_none
```
- HAS_BALANCE-set + bal=0 → present_none (authoritative drain)
- HAS_BALANCE-clear + bal=0 → still resolve (self-burn / CREATE-seed paths)

## Rebase + re-measure @ main 231560acc
| | main | fix | Δ |
|---|---|---|---|
| text | 0x549e0 | **0x549ec** | **+0xc** |
| data/bss | unchanged | unchanged | 0 |
| ELF sha | `96e6dbc17f6db41a…` | `65c0ba2370b3bf12…` | distinct |

### 01114 ×4 CREATE/CREATE2
- main: succ=0/4 FR
- fix: **PASS full 4/4**

### r200 seed=932419178 (paired spike)
- FA0→0 FR20→20; P→F=0; byte-identical; FR label-set PASS; ELFs distinct

### Layout regen + 2nd pass NO-OP
textSizeBytes 0x549ec

Closes #11688


pirapira:
**coord review: content approved.** Awaiting `build`/`source-checks`; will label when green. CONFLICTING already since main has moved — that is the actor's, not the author's.

## ⛔ First: my hypothesis was wrong, and the real cause is the inverse of it

I briefed this as *"probably uninitialised scratch, not a value — `0x6400000000000000` is only the high limb, not a clean LE or BE 100, so the fix belongs at the initialisation or the miss-path return, not at the copy."*

**Wrong on the premise.** The CREATE address **pre-exists with balance `0x64`**, and `0x6400000000000000` is the correct **little-endian load of the last 8 bytes of big-endian 100**. Nothing was uninitialised and nothing was garbage.

⭐ **And the actual defect is the inversion of the hazard I warned about.** I said *never read a default as a value*. The bug is the opposite: **an authoritative value was read as a default.** `account_writes_apply_deletes` treated map-balance `0` as "unknown, go resolve the lower tier" — but when `HAS_BALANCE` is *set* with value 0, the zero is a **fact**: SELFDESTRUCT drained the account. Re-fetching resurrected the pre-seeded balance of 100, `keep_present` then saw a nonzero balance and set `opt = 1`, and the row read Present. Wrong NPR, `bv_fail = 1`.

⇒ So the discriminator was never balance-vs-garbage; it was **"is this zero asserted or absent?"** — and the answer is carried by the component mask, not by the value. That is the same lesson as #11706's offsets one layer up: the meaning of a datum lives in the flag that says whether it was written, not in the datum.

## The fix is three instructions and matches the diagnosis exactly

```
ld t1, 112(t0); andi t1, t1, <accountWriteHasBalance>; bnez t1, .Lawd_present_none
```

`HAS_BALANCE` set with bal 0 → `present_none`. `HAS_BALANCE` clear with bal 0 → still resolve, preserving the self-burn (03736) and CREATE-seed paths. **`.text 0x54e18 → 0x54e24`, +0xc = 12 bytes = exactly three instructions** — the size confirms nothing else was emitted.

⭐ And the docstring was rewritten rather than patched around it: the old text said "when map bal=0, resolve the preserved balance" full stop; the new text splits the two cases and names why, including the 01114 resurrection. The existing warning about not using the live AccountState overlay as a pre-state tier is preserved intact.

## Evidence

| | main | fix |
|---|---|---|
| ELF sha256 | `43384f5dd23f68d1…` | `2640cc61844d83e7…` |
| `.text` | `0x54e18` | `0x54e24` (+0xc) |
| `.data` / `.bss` | `0x53b0` / `0x19029540` | unchanged |

**01114 F→P on all four CREATE/CREATE2 variants** (main `succ=00 exp=01`; fix `succ=01 exp=01`, root OK), and **r200 seed 932419178 with P→F = 0, byte-identical**, base `a2348b1db`, both legs named. Spike throughout — it started on ziskemu, noted it as slow with jobs forced to 1, and switched.

⚠️ Recording that tooling detail because it caused a real coordination cost: an hour of ziskemu silence read as a dead session, and I reassigned the issue on that basis. The fleet is on spike for uniformity and performance; a result from a different emulator is worth less than the same result from the tool the other agents used, because paired comparability has been the binding constraint all afternoon.

## Left open deliberately, and correctly

`keep_present`'s `OR`-of-limbs presence inference is **not** touched here and is reported as a separate finding. That is the right call: this PR fixes *why the balance was resurrected*, while the predicate that turns any nonzero balance into Present is a wider defect class — a row that is absent but whose buffer was never zeroed would still read Present. It should be filed and fixed on its own evidence, not folded into a three-instruction fixture fix.

Issue #11711, opened 2026-08-07
proof: DecodeChain demands fuel-insensitivity, which blocks every nested-list bridge (rlp_list_count_items is the live case)


dhsorens:
Split out of #11341 / PR #11694, where this stopped being a theoretical restriction and became the blocker on a routine that is live in the guest image.

## The restriction

`Rv64/RLP/WalkDecodeBridge.lean:405-408`, in the tree's own words:

> ⚠️ Each link demands fuel-insensitivity (`∀ m`), which is sound only for **byte-string** items (`decodeAux_bytes_all_fuel_of_decode`). Header fields all are. **This is the same restriction that makes the nested-list case of #11341 unprovable** — a nested list's decode *is* fuel-sensitive.

`DecodeChain`'s per-link obligation is `∀ m, decodeAux (m + 1) (bytes.drop off) = some (item, bytes.drop off')`. For a byte-string item that holds — the decode consumes no fuel recursively. For a **list** item `decodeAux` recurses into `decodeItems nDepth payload`, so the result genuinely depends on `nDepth`, and the `∀ m` form is false rather than unproven.

Corroborating evidence, independent of the comment: `WalkDecodeBridge` has `decodeAux_singleByte_bridge`, `decodeAux_shortBytes_bridge`, `decodeAux_bytes_all_fuel_of_decode` — and **no list bridge at all**. The absence is structural, not an oversight.

## ⭐ Why this is now urgent rather than latent

`rlp_list_count_items` is called by `mpt_node_kind` as of #11675 (merged), to check node arity exactly. MPT branch children are **either a 32-byte hash or an inline embedded node**, and `SpecRef/IncrementalMpt.lean:155` `resolveChildRefAux` handles that second case explicitly:

```lean
| .list items => …   -- inline node, decoded directly
```

So nested-list items are not an edge case on this path — they are the normal shape of a branch node with a small child. The routine is emitted into both `zisk_stateless_verdict` and `zisk_step2_verdict`, so this is executing code.

⇒ `rlp_list_count_items`'s Correspondence row **cannot** reach `.bridged` until this is resolved. That is recorded on #11341 and in the row's note; this issue is the actual blocker.

## What is wanted

A **fuel-sensitive** chain predicate — carrying a fuel budget through the links instead of quantifying it away — so a nested-list item's decode can be related at the fuel the caller actually has. Sketch of the shape, not a design:

```lean
[...]

PR #11712, opened 2026-08-07
proof(#11341): record the count-items blocker in the row; restricted predicate + witness

dhsorens:
Answers the blocking review question on #11694. **It is case (2)**: the domain includes nested-list items, so `RlpItemDecodeBridges` is *unsatisfiable* rather than merely unproven, and the row is **blocked** rather than unfinished.

## The evidence

`SpecRef/IncrementalMpt.lean:155`, `resolveChildRefAux`:

```lean
```

An MPT branch child is either a 32-byte hash **or an inline embedded node** — a nested list. Since #11675 this routine is reached from `mpt_node_kind`, so nested lists aren't an edge case on this path; they're the normal shape of a branch node with a small child.

`DecodeChain` demands fuel-insensitivity (`∀ m`) per link, sound only for byte strings, and a nested list's `decodeAux` recurses into `decodeItems nDepth payload`. `WalkDecodeBridge.lean:405-408` already said so — I reached it independently from the absent list bridge and only saw the comment when you quoted it.

⭐ **My own #11675 made this structural.** Wiring `rlp_list_count_items` into `mpt_node_kind` was right on its own terms — it turned an accept into a reject — but it moved the routine onto the one path where inline nested lists are guaranteed.

## What this PR does

- **Registry row note** now carries the blocker, so it's visible where the *verdict* lives rather than only in an issue. Verdict and basis unchanged → no census moves.
- **`RlpItemDecodeBridgesBytes`** — the byte-string-restricted predicate, with the restriction **visible in the statement** rather than hidden in a comment, so no caller can pick it up believing it covers lists.
- **`bridgesBytes_witness_singleByte`** — your non-vacuity ask, so the structural theorems are demonstrably about something.

## ⛔ What I deliberately did not do

A witness for the **unrestricted** predicate. Under case (2) one cannot exist, and hand-picking an all-byte-strings input would show the proposition is non-empty while saying nothing about this routine — precisely the vacuity you were guarding against. For the same reason the restricted predicate is labelled as **not** the row's bridge.

**#11711** filed for the fuel-sensitive `DecodeChain` variant that would actually unblock the row.

## Gates

`check-forbidden-tactics`, `check-registry-coverage`, `lake build …Correspondence` and `…RlpListCountItemsBridge` — all OK. Axioms `[propext, Quot.sound]` on the witness.

⚠️ Also ran **`check-build-units-link`** (OK). That is the gate I did not run on #11675, where adding a cross-routine call broke two standalone units' link closure. Adding a call changes link closure, and my byte-identity/layout sweep was blind to that axis — thank you for catching and fixing it in `01d86308c`.


pirapira:
**coord review: `merge!` applied.** All six checks green. This answers the blocking question I put on #11694, and it answers it as the **worse** of the two cases — which is why the PR is right to record a blocker rather than claim progress.

## The answer is case (2), and I verified the pivot

`SpecRef/IncrementalMpt.lean`, `resolveChildRefAux` — its docstring states *"List case: inline node, decoded directly"*, and the `.list items` arm calls `decodeNodeItemAux fuel nodeDb (.list items)`. So an MPT branch child is either a 32-byte hash **or an inline embedded node**, i.e. a nested list.

⇒ **Nested lists are not an edge case on this path; they are the normal shape of a branch node with a small child.** So `RlpItemDecodeBridges` is **unsatisfiable**, not merely unproven, and the registry row is **blocked** rather than unfinished. `DecodeChain` demands fuel-insensitivity (`∀ m`) per link, sound only for byte strings, and a nested list's `decodeAux` recurses into `decodeItems nDepth payload`.

⭐ **And the author names its own #11675 as what made this structural**: wiring `rlp_list_count_items` into `mpt_node_kind` was correct on its own terms — it turned an accept into a reject — but it moved the routine onto the one path where inline nested lists are guaranteed. Recording that a correct change created a new obligation, rather than presenting the obligation as pre-existing, is the honest version.

## The three things it does are each answers to a specific ask

- **The blocker is in the registry row's note**, so it is visible where the *verdict* lives rather than only in an issue. Verdict and basis unchanged → **no census moves**, which is right: nothing was proved.
- **`RlpItemDecodeBridgesBytes`** carries the byte-string restriction **in the statement** rather than in a comment, so no caller can pick it up believing it covers lists. ⭐ That is the correct shape — a restriction that lives in prose is one refactor away from being lost.
- **`bridgesBytes_witness_singleByte`** is the non-vacuity witness I asked for, axioms `[propext, Quot.sound]`.

## ⛔ And what it declined to do is the part I most want on the record

It did **not** produce a witness for the **unrestricted** predicate. Under case (2) none can exist, and hand-picking an all-byte-strings input *"would show the proposition is non-empty while saying nothing about this routine — precisely the vacuity you were guarding against."*

That is exactly right, and it is a sharper statement of the hazard than my original ask. I requested a satisfiability witness to prove the structural theorems weren't vacuous; the correct response, once the domain is known to include lists, is that a witness drawn from *outside* the routine's domain would restore the appearance of non-vacuity while removing its substance. For the same reason the restricted predicate is labelled as **not** the row's bridge.

**#11711** is filed for the fuel-sensitive `DecodeChain` variant that would actually unblock the row. That is the real prerequisite, and it is larger than more lemmas.

## ⭐ On the process note

It ran **`check-build-units-link`** — the gate it did *not* run on #11675, where adding a cross-routine call broke two standalone units' link closure. Its observation is the transferable one: **adding a call changes link closure, and a byte-identity or layout sweep is blind to that axis.** A PR that names the gate its predecessor missed, and runs it, is the useful form of a retrospective.

⚠️ For whoever picks up #11711: this row stays `.machineOnly` and the blocker is now in the note, so the next attempt starts from a stated obligation rather than rediscovering it. And the three byte-string disjuncts already landed on #11694/#11703 hold under **every** resolution of the list question, so that work is not stranded.

Issue #11696, opened 2026-08-07
coherence: the BAL rebuild's binding row limit is an unnamed 20480 literal in asm, while four builder capacities are declared up to 6.8x larger


pirapira:
Found while settling #11506. Evidence from scoper; the gas-derivation connection is mine and is the part that reframes it. **Fail-closed, so no FA** — this is a coherence and documentation defect with a real sizing consequence. (Issue generated by Claude Code.)

## The defect

`bal_canonical_sort` rejects any container with **more than 20480 rows**, via an unnamed literal in emitted assembly:

```
22944:  lui x5, 0x5        # = 20480
22945:  bltu x5, x11       # rows > 20480  ->  li x10, 1  (sort failure)
```

That failure propagates: `bal_serializer_rebuild_hash` returns non-zero only from one of the six `bal_canonical_sort` calls, and `bal_serializer_verify` returns 2 iff that is non-zero → **fail code 61**.

⭐ **Five of the six containers it guards are declared, as named constants, larger than the sort will accept:**

| container | declared capacity | vs 20480 |
|---|---|---|
| `bal_builder_accounts` | 140000 | **6.8×** over |
| `balance_changes` | 105000 | **5.1×** over |
| `storage_changes` | 47522 | **2.3×** over |
| `nonce_changes` | 35000 | **1.7×** over |
| `code_changes` | 13125 | under |
| `storage_reads` | 16384 | under |

Capacities from `BlockAccessListBuilder.lean:106-110`; the reads bound from `RegionMap.lean:225`.

## ⭐ 20480 is not arbitrary — it is an unnamed duplicate of a named, gas-derived constant

`AccountWriteMap.lean:112`:

```lean
/-- Block-lifetime entries. Amsterdam permits at most 9523 minimum-cost plain
    value transfers in a 200M-gas block; distinct senders and recipients plus
    coinbase require 19047 keys. Round to a 2.5 MiB arena. -/
def blockAccountWritesCapacity : Nat := 20480
```

**Same number, with a derivation.** So the sort limit is very likely the *intentional* bound — it is the 200M-gas block bound — and the four oversized builder capacities are the ones out of step, not the sort.

⇒ That changes the fix from "raise the literal" to **"name it and reconcile the four."** The literal should reference the derived constant rather than restate it, and the builder capacities above it should either be reduced to match or carry an explicit note that the sort will reject before they fill.

## Consequence, which is a sizing hazard rather than a soundness one

[...]

Issue #11643, opened 2026-08-06
latent: topLevelValueRecipientStateGasAsm calls account_state_lookup_current with an uninitialised a0 — masked by empty pending/durable tables


pirapira:
Latent on `main` today. Found while building the #11581 migration locally; filing separately because **it exists independently of whether that migration ever lands.**

## The defect

`BlockVerdictSimpleTransferGas`'s `topLevelValueRecipientStateGasAsm` — **four emitted expansions** — loads the context pointer into `t1` and then calls `account_state_lookup_current` **without setting `a0`**. The routine's contract is `a0 = pointer to a canonical 20-byte BE address`; the address lives at `ctx + 72`, so the call should be preceded by `addi a0, t1, 72`. It is not. `a0` holds whatever the previous code left there.

## Why nothing has failed

`account_state_lookup_current` scans the AccountState **pending** then **durable** tables. At this point in the flow both are empty, so the routine returns before ever dereferencing `a0`. **The uninitialised argument is harmless only by that accident.**

Make the callee reach a non-empty table — or replace it with one that dereferences before scanning — and the same caller faults immediately. That is exactly what happened locally: swapping in a map-backed lookup produced a **load access fault**, `mcause 0x5`, `mtval 0`, at an `lbu` from `0xb8411bf8` — an address in no mapped region (not `.text`, not `ACCOUNT_WRITES` at `0xa28a0000`, not `.data` at `0xa3000000`, not `.sszscratch` at `0xbf980000`), i.e. a garbage pointer rather than a mis-scaled index.

## Why no gate catches it

⭐ **A callee's early return on an empty container masked a caller that never set its argument.** Nothing is wrong until something gets fuller, so:

- every build passes, including nine targeted ones during the migration;
- the corpus passes, because the tables really are empty on that path;
- a contract diff finds nothing, because **the ABI is not in question** — the caller simply does not satisfy it;
- and an audit looking for callers that set the argument *differently* misses it entirely. The search that finds it is *callers that do not set it at all, on any reaching path*.

## Fix

`addi a0, t1, 72` before the call, in all four expansions. Verified locally: after the fix every emitted `lookup` callsite either sets `a0` to its BE20 scratch/context address or preserves an incoming pointer, and a 202-case sweep completes with `output_differences=0`.

[...]

PR #11717, opened 2026-08-07
fix(codegen): pass per-arena BAL sort capacities

pirapira:
## Summary

`bal_canonical_sort` now receives a caller-owned row capacity in `a5` and fails closed with status 1 when `a5 == 0` or `count > a5`. The existing `a0`–`a4` ABI is unchanged. The `a5 == 0` result intentionally shares status 1 with an over-capacity count; every emitted callsite is guarded to set `a5`, so the omission case is compile-time checked rather than a runtime protocol.

The six production capacities are named constants in `EvmAsm.Codegen.Programs.BalCapacities` and are emitted with `toString`, rather than duplicated assembly literals:

- storage changes: `balBuilderStorageChangeCapacity = 47,522`
- storage reads: `balBuilderStorageReadsCapacity = 16,384`
- balance changes: `balBuilderBalanceCapacity = 105,000`
- nonce changes: `balBuilderNonceCapacity = 35,000`
- code changes: `balBuilderCodeCapacity = 13,125`
- account changes: `balBuilderAccountCapacity = 140,000`

These are the authoritative physical row capacities for the six caller-owned arenas. The storage-change, balance, nonce, code and account bounds are derived in `BlockAccessListBuilder.lean` from their gas/resource workloads; the storage-read bound is the existing read-set arena capacity. The self-test and direct probe also set `a5`; emitted-source guards require all six production sites and the probe argument.

## Evidence

The branch was rebased onto `34c005ca6` before this measurement. Prediction: `.text` grows only by the six capacity loads; `.data` and `.bss` remain unchanged. The rebased base ELF has `.text = 0x549dc`; the candidate has `.text = 0x54a08` (`+0x2c`), `.data = 0x5370` and `.bss = 0x19029540` unchanged. Fixed-point regeneration made a second complete pass a no-op, including the linked `gen-out/regionmap/stateless_guest.elf` SHA `171992c2efd91cf3ceeb0f65f876f04dcdd7d0934f078ded11b9e78afcb9098c`. `bnf_le_a` and the other pinned bases remain unchanged.

Structural gates:

- `LAKE_ARTIFACT_CACHE=false lake build` — completed
- `scripts/gen-symbol-addresses.py --build`, `asm_to_program.py guest-addrs`, `guest_image_coverage.py --emit-lean`, and `gen-region-map-link-pins.py` — two complete passes; second pass no-op
- `LAKE_ARTIFACT_CACHE=false scripts/check-asm-to-program.sh` — CLEAN
- `scripts/check-region-map.sh` — CLEAN (ELF, TSV, pins, and Class-A ratchet)
- `scripts/check-unimported.sh` — 2,767 modules reachable, 0 orphans

Paired random-200 Spike run after the rebase (seed 4242, 30 processes, same fixture manifest):

- base commit `34c005ca6`, ELF SHA `63e7f54b769e0d0cc4db07d5c7e90dfc28591266787915d1caf1f7a236b2ce74`
- candidate commit `ad7d8a797`, ELF SHA `4bce8b80204e6215ee33a4f0e79082ae9d31ffeebbd837958479808621443583`
- both legs: 200/200 ran, 0 errors, FA=0, FR=23
- A/B: identical 23-label FR sets, 0 forward flips, 0 reverse flips, 0 output-byte differences on 198 joined input digests; the join covered all 200 manifest rows with two many-to-one shared-input groups

The change is behavior-neutral on this sample, as expected because no corpus row reaches the old 20,480-row threshold. That is a regression result, not evidence that the newly enforced per-arena bounds are unexercised in the general case; the spec-aligned reason for the change is that each sorter caller must enforce its own physical capacity.


pirapira:
**coord review: one change required before `merge!`.** The design is right and the evidence is complete — but the six `a5` values are **hardcoded literals**, which recreates at six new sites exactly the defect class this issue exists to remove.

## ⛔ The blocking point

```
li a2, 96; li a3, 0x0818a0209400; li a4, 3; li a5, 47522
li a2, 64; li a3, 0x2020;         li a4, 1; li a5, 16384
li a2, 64; li a3, 0x08189400;     li a4, 2; li a5, 105000
li a2, 40; li a3, 0x08189400;     li a4, 2; li a5, 35000
li a2, 64; li a3, 0x08189400;     li a4, 2; li a5, 13125
li a2, 24; li a3, 0x9400;         li a4, 1; li a5, 140000
```

**This issue is about an unnamed `20480` in emitted assembly shadowing the named `blockAccountWritesCapacity`.** Emitting `47522`, `105000`, `35000`, `13125` and `140000` as bare literals is the same defect, six times over — and *worse* than the original, because a capacity change would now need remembering in seven places instead of two.

The constants exist and are already in scope of this PR, which touches `BlockAccessListBuilder.lean`:

```lean
def balBuilderAccountCapacity       : Nat := 140000   -- :106
def balBuilderStorageChangeCapacity : Nat := 47522    -- :107
def balBuilderBalanceCapacity       : Nat := 105000   -- :108
def balBuilderNonceCapacity         : Nat := 35000    -- :109
def balBuilderCodeCapacity          : Nat := 13125    -- :110
```

⇒ **Use `toString balBuilderAccountCapacity` and siblings**, the way the array-size definitions at `:198-202` already derive `…Bytes` from `…Capacity`. The storage-reads bound (16384) should likewise reference its constant in `ReadSetsPromote.lean`. If a name is not reachable from `BalSerializer` without a new import, **take the import** — that cost is trivial next to the duplication.

⭐ **And the payoff is concrete, not stylistic**: with names, a future capacity change propagates to the sorter automatically, and the guard cannot silently disagree with the arena it guards. With literals, the next person to resize an arena reintroduces this bug and no gate catches it — which is precisely the history recorded in #11506 and this issue.

## Everything else is approved

**The design is right, and it is the only safe one.** A single shared cap is sound only at the **minimum** of the capacities it protects; 20480 sat *between* the min (13125) and max (140000), so it was simultaneously redundant for two containers and falsely rejecting for four. Per-call `maxRows` bounds each container by its own arena.

**`a5` is mandatory and fail-closed** — zero yields status 1, checked *before* the count — with emitted guards pinning the call sites. ⭐ That is the right answer to the hazard I flagged: adding a required ABI argument is how #11697 and #11643 happened, and a compile-time guard plus fail-closed runtime is the pair that makes a forgotten `a5` impossible to ship.

**The range-stack question is answered and no `.bss` is needed**: the 20480 cap was a genuine row-count cap (status 1 at `:222-223`), the status-3 path is separate (`:287-289`, bound 2048), and the stack bound is depth × fanout — independent of row count.

**Catching the selftest relocation-index shift** after adding `a5` is the kind of thing that otherwise surfaces as an unexplained probe failure.

**Prediction held**: `.text 0x54e18 → 0x54e44` (+0x2c), `.data` and `.bss` unchanged, `bnf` pins untouched, full regen at fixpoint, `check-all` and `check-region-map` clean including pins, TSV and the Class-A ratchet.

**Paired r200**: base `a2348b1db` / ELF `6989766d3…`, candidate `0155b0d68` / ELF `7ae2f8b64…`, seed 4242, 200/200 both legs, 0 errors, FA=0, FR=23 on both, 198 digests joined, **zero forward or reverse flips**.

⚠️ Behaviourally neutral is the **expected** result and should be said as such: no corpus row reaches 20480 rows in any container, which is why code 61 never fires. So the sweep proves no regression and **cannot** prove the fix works — the argument is the gas arithmetic (20480 touches ≈ 61.4M gas; 200M affords 66,666 at 3,000 each, so 3.3× the old cap was reachable and falsely rejected).

## Also required in the body

⭐ **Which capacity is now authoritative per container.** That is the proof-relevant sentence: obligations stated over `balBuilderAccountCapacity` were stated over a bound the sorter did not honour, and after this they are. A proof author needs a line to cite.

And note that CONFLICTING is expected — main has moved to `e9e10bf1d`. Regenerate pins **after** the final rebase.

pirapira:
**coord re-review: the blocking point is fixed. Content approved** — holding `merge!` only until `build` reports (`source-checks` and the other four already pass).

## The six `a5` values now reference the named constants

```lean
li a2, 96; li a3, 0x0818a0209400; li a4, 3; li a5, " ++ toString balBuilderStorageChangeCapacity
li a2, 64; li a3, 0x2020;         li a4, 1; li a5, " ++ toString balBuilderStorageReadsCapacity
li a2, 64; li a3, 0x08189400;     li a4, 2; li a5, " ++ toString balBuilderBalanceCapacity
li a2, 40; li a3, 0x08189400;     li a4, 2; li a5, " ++ toString balBuilderNonceCapacity
li a2, 64; li a3, 0x08189400;     li a4, 2; li a5, " ++ toString balBuilderCodeCapacity
li a2, 24; li a3, 0x9400;         li a4, 1; li a5, " ++ toString balBuilderAccountCapacity
```

⭐ **That is the issue's own thesis honoured rather than re-violated.** #11696 exists because an unnamed `20480` in emitted assembly shadowed the named `blockAccountWritesCapacity`; emitting `47522`, `105000`, `35000`, `13125` and `140000` as literals would have recreated the same defect at six new sites — and worse, since a capacity change would then need remembering in seven places instead of two. Now the emitted value **follows** the arena it guards, and the guard cannot silently disagree with what it protects.

⚠️ And note `balBuilderStorageReadsCapacity` appears as a named constant too, rather than a bare `16384` reaching across to `ReadSetsPromote`. That was the one I expected to need an import; it was solved properly.

## Everything else stands from the first review

**The per-container design is the only safe one.** A shared cap is sound only at the **minimum** of the capacities it protects; `20480` sat *between* the min (13125) and the max (140000), so it was simultaneously **redundant** for two containers and **falsely rejecting** for four — up to 6.8× under the account arena.

**`a5` is mandatory and fail-closed** (zero → status 1, checked *before* the count) with emitted guards pinning the call sites. That is the right answer to the hazard that produced #11697 and #11643: adding a required ABI argument is exactly how an unset register becomes garbage the callee reads. Compile-time guard plus fail-closed runtime makes a forgotten `a5` unshippable.

**The range-stack question is settled**: the `20480` was a genuine row-count cap (status 1, `:222-223`), the status-3 path is separate (`:287-289`, bound 2048), and the stack bound is depth × fanout — independent of row count, so **no `.bss` growth**.

**The selftest relocation-index shift** after adding `a5` was caught — the kind of thing that otherwise surfaces as an unexplained probe failure.

## Two things the body must carry

⭐ **Which capacity is authoritative per container.** This is the proof-relevant line and the reason the work was called urgent: obligations stated over `balBuilderAccountCapacity` were stated over a bound the sorter did not honour, and after this they are honoured. A proof author needs a sentence to cite.

⚠️ **Behavioural neutrality is the expected result, not a null outcome.** No corpus row reaches 20480 rows in any container — which is why code 61 never fires (#11506) — so the sweep proves **no regression** and *cannot* prove the fix works. The argument is the gas arithmetic: 20480 cold touches ≈ 61.4M gas, and 200M affords 66,666 at 3,000 each, so **3.3× the old cap was reachable and falsely rejected**. Both halves together, or an unchanged sweep reads as a change that bought nothing.

⚠️ Finally: this PR was stuck for ~40 minutes with **no build run at all** because it was CONFLICTING at push time — the same signature as #11708. Rebasing is what released it. Worth knowing that an unlabelled conflicting PR cannot reach green on its own.

pirapira:
**coord review of the current head `ad7d8a797`.** ⚠️ My previous comment was written against the pre-force-push head and is superseded — that version had 9 files; this one has 11 and is a better design. **All six checks pass. Content approved**; `merge!` goes on once the CONFLICTING state clears (main moved again after the rebase — that half is the actor's).

## ⭐ The fix went further than I asked, and the extra step is the right one

I asked for the six `a5` values to reference the existing constants. Instead there is now a **single shared module**, `Programs/BalCapacities.lean`, holding all six — and its docstring states the reason exactly:

> *"Keeping these values in one module prevents emitted callsites from quietly duplicating an arena bound as an unrelated assembly literal."*

That is #11696's thesis stated as a design invariant rather than as a fix to one site. Referencing the old definitions from `BlockAccessListBuilder` would have worked; **hoisting them removes the question of which file owns the number.**

⭐ And `balBuilderStorageReadsCapacity = 16384` is now a named constant in that module — the one I expected to need an awkward cross-module import from `ReadSetsPromote`. `ReadSetsPromote.lean:137` now emits `toString balBuilderStorageReadsCapacity`.

## Verified: no duplicate definitions left behind

`BlockAccessListBuilder.lean` no longer defines the capacities — the surviving occurrences of `140000`, `47522`, `105000`, `35000`, `13125` are in the **derivation docstring** (`:64`, `:82-94`), which is where those numbers *should* appear as arithmetic, and `:193-194` consume the constants (`balBuilderAccountCapacity * balBuilderAccountRowBytes`). So there is one definition site and every other use is a reference.

⚠️ **One leftover worth a follow-up, not a block:** `ReadSetsPromote.lean:141` still carries a bare `li a6, 16384` on a different callsite. It is not the sort cap this PR is about, so it is out of scope — but it is the same literal now available as a name, and it is exactly the kind of residue that makes the next reader wonder which is authoritative. Worth sweeping when someone is next in that file.

## The design, restated for the record

**Per-container is the only safe shape.** A shared cap is sound only at the **minimum** of the capacities it protects; `20480` sat *between* the min (13125) and the max (140000), so it was simultaneously redundant for two containers and falsely rejecting for four — up to **6.8×** under the account arena.

**`a5` is mandatory and fail-closed** (zero → status 1, checked before the count), with emitted guards pinning the callsites. That is the correct answer to the ABI-argument hazard behind #11697 and #11643: a compile-time guard plus a fail-closed runtime makes a forgotten `a5` unshippable, rather than silently disabling the bound.

**No `.bss` growth**: the `20480` was a genuine row-count cap (`:222-223`); the status-3 path is separate (`:287-289`, bound 2048); the range-stack bound is depth × fanout, independent of row count.

**Extents held across the rebase**: `.text 0x549dc → 0x54a08` (+0x2c), `.data 0x5370` and `.bss 0x19029540` unchanged, two complete regen passes with the second a no-op — and this time the regeneration covered all four artefacts, including `GuestAddrs.lean` and the two `.s` fixtures.

## Why the neutral sweep is the expected result

No corpus row reaches 20480 rows in any container — which is *why* code 61 never fires (#11506). So the sweep proves **no regression** and cannot prove the fix works. The argument is the gas arithmetic: 20480 cold touches ≈ **61.4M gas**, and 200M affords **66,666** at 3,000 each, so **3.3× the old cap was reachable and falsely rejected**. Both halves belong in the body together.

⭐ And the proof-relevant line is now in the body: **which capacity is authoritative per container.** Obligations stated over `balBuilderAccountCapacity` were stated over a bound the sorter did not honour; after this they are honoured, and a proof author has a sentence to cite. That was the reason this was called urgent.

PR #11718, opened 2026-08-07
codegen(#11578): convert execution_requests_hash to a Program

dhsorens:
The prerequisite for #11578 that the retarget did not anticipate: **`execution_requests_hash` had no `Program`, so no `cpsTripleWithin` was statable for it at all.** This PR is the conversion only — no proof. The validation-contract triple follows.

## Why there was no `Program`, and why "run the tool" was not the fix

`RequestsHash.lean` defined only `String`s. The reason is structural, and it is the same shape as #10817's "0 theorems" (which turned out to be a conversion gap fixed by #11046):

`executionRequestsHashFunction` carried **two non-`.L` labels** — `execution_requests_hash:` and `erh_hash_one:`. `scripts/asm_to_program.py:492-495` refuses such defs on purpose:

> `emitProgram` strips every internal label, turning branches into PC-relative offsets. That is safe for `.L`-local labels but a secondary NON-`.L` label is a potential cross-function entry point: external `jal`s in OTHER files resolve to it, and stripping it silently breaks the guest link. **Per-function byte-identity still passes** (the bundle is self-consistent in isolation), so ONLY the whole-guest byte-identity gate catches it.

So the routine sat in the `MULTI_ENTRY_BUNDLE` bucket (`docs/4ch8f-asm-to-program-coverage.md:639`).

**The refusal is a false positive here** — `erh_hash_one` is referenced only by the five local `jal`s inside `execution_requests_hash` and by nothing else in the tree (`git grep` over `EvmAsm/` and `scripts/`: 5 call sites, its own label, one TSV row).

⚠️ **But the fix is not to weaken the rule.** It is to make each half single-entry, which is also what the linker already thinks: `symbol-addresses.tsv:144` gives `erh_hash_one` its own address, `0x8000c3fc`. The five call sites become ordinary cross-function relocs, exactly like the `jal ra, bgv_u32le` beside them.

## What landed

- **`executionRequestsHash_prog`** — 135 instructions, `#guard`ed; **`erhHashOne_prog`** — 23, `#guard`ed. Both with reloc tables and `rfl` `_eq_prog` drift guards, produced mechanically by `asm_to_program.py rewrite`.
- Two fixtures, two MANIFEST rows.
- `GuestAddrs`: **+3 symbols** (`erh_hash_one`, `erh_digests`, `erh_blob`), regenerated — 1020 → 1023.
- `GuestImageEntries`: regenerated, 340 → **342** entries, both programs registered so `guestImageCodeReq` can lift them. This is what makes a triple statable at all (`GuestImage.lean:64-69` quantifies over `guestImageEntries`).
- `executionRequestsHashFunctions` aggregate; the five call sites now use it.

⚠️ **`erh_hash_one` is not a standard-ABI function and splitting the text does not make it one.** It reads `s10` (body length) and `s8` (digest destination) out of `execution_requests_hash`'s live frame, so it is callable only from there, and a triple over it must carry those registers as inputs. Recorded in its docstring so the split is not mistaken for independence.

## ⚠️ The one hazard that materialised, and the check that caught it

The plan for this PR flagged it in advance, and it happened exactly as written:

**`emitProgramR` emits no trailing newline.** Before conversion the halves were string literals and `executionRequestsHashFunction` ended `"  ret\n"`, so `a ++ b` was a correct concatenation. After conversion both halves end mid-line, and `a ++ b` renders

```
  ret erh_hash_one:
```

— which does not assemble. **Every per-function check stays green**, because `check-asm-to-program` assembles each converted def in isolation and never the concatenation. The requirement appeared *because of* the conversion, silently.

Fixed with an explicit `++ "\n" ++`, and the reason is in the def's docstring rather than left for the next person to rediscover.

**Verified by assembling the aggregate by hand**, not by inference — dumped `executionRequestsHashFunctions` from Lean and ran it through `riscv64-elf-as`:

```
AGGREGATE ASSEMBLES OK
    24: 000000000000021c     0 NOTYPE  LOCAL  DEFAULT    1 erh_hash_one
```

⭐ `erh_hash_one` lands at **0x21c = 540 bytes = 135 × 4** — which is precisely the linker's own offset (`0x8000c3fc − 0x8000c1e0 = 0x21c`). So the instruction count, the split point, and the ELF agree independently.

## Verification

- `lake build` clean.
- **`check-asm-to-program.sh` genuinely ran** (the RISC-V toolchain is present locally — `riscv64-elf-as`/`objcopy`/`readelf`): **CLEAN, 382 converted defs across 144 files** (was 380 / 143). Not a silent skip.
- `check-guest-addrs`: CLEAN (1023 symbols).
- `check-guestaddrs-starts`: **linked_converted=342 ok=342 fail=0**.
- `check-registry-coverage`: OK. Linked symbols 1019 → 1022.
- `check-file-size`, `check-forbidden-tactics`, `check-unimported`, `check-layering`.
- The hand-assembled aggregate above.

No `#print axioms` line: this PR adds no theorem beyond the two `rfl` drift guards.

## Next on #11578

The two validation-prefix triples, now statable. Scope is as endorsed on the issue: **the divisibility and cap contract, with `sha256` equality as a separate row.**

Two things I learned while reading the routine that will shape that PR, both correcting the plan:

- ⭐ **The validation prefix is straight-line, not a loop** — five unrolled 7-instruction gates (`sub`/`li`/`remu`/`bnez`/`divu`/`li`/`bgtu`) branching to one shared `.Lerh_fail`. So it forks the guard-cascade exemplar (`txValidateAgainstBlockSAsm.lean:30-54`, `RetForwardJoin.lean`), not a copy loop. No loop invariant needed.
- ⚠️ **The anchor named on the issue and in `leaf-routine-targets.md:115` is wrong.** `compute_requests_hash` (`SeamShell.lean:103`) is total — `sha256 (requests.flatMap sha256)` — and imposes neither divisibility nor caps. The real counterpart is `SszCodec.deserializeAux`'s fixed-size-list arm (`SpecRef/SszCodec.lean:235-241`), which performs exactly the `% sz != 0 → reject` / `count > lim → reject` pair, same operator and same strictness. I will raise that on the issue separately so the row is not graded against the wrong reference.


pirapira:
**coord review: `merge!` applied.** Conversion only, no proof — and it is the right prerequisite rather than a workaround.

## The gap was structural, not an oversight

`execution_requests_hash` had **no `Program`**, so no `cpsTripleWithin` was statable for it at all. Cause: two non-`.L` labels (`execution_requests_hash:` and `erh_hash_one:`), which `asm_to_program.py:492-495` refuses on purpose — `emitProgram` strips internal labels and turns branches into PC-relative offsets, so a secondary non-`.L` label is a potential cross-function entry point whose external `jal`s would silently break the guest link. Hence the `MULTI_ENTRY_BUNDLE` bucket.

⭐ **And the reason nobody noticed is the reusable part**: *"per-function byte-identity still passes (the bundle is self-consistent in isolation), so ONLY the whole-guest byte-identity gate catches it."* That is the same axis-blindness that bit #11675, where adding a cross-routine call broke two standalone units' link closure while a layout sweep saw nothing. **A per-unit check passing is not evidence about a whole-image property**, and the two failure modes look identical from inside the unit.

## ⛔ It refuses to weaken the rule, which is the decision I care most about

The refusal *is* a false positive here — `erh_hash_one` has five local `jal`s inside `execution_requests_hash` and nothing else in the tree references it. The tempting fix is to relax the guard for "obviously local" secondary labels.

⇒ Instead it **makes each half single-entry**, which is what the linker already believes: `symbol-addresses.tsv:144` gives `erh_hash_one` its own address (`0x8000c3fc`), and the five call sites become ordinary cross-function relocations, *"exactly like the `jal ra, bgv_u32le` beside them."* A guard that is right in general does not get loosened for one true negative.

## ⚠️ And it does not overclaim what the split achieves

> *"`erh_hash_one` is not a standard-ABI function and splitting the text does not make it one. It reads `s10` (body length) and `s8` (digest destination) out of `execution_requests_hash`'s live frame, so it is callable only from there, and a triple over it must carry those registers as inputs."*

Recorded in the docstring so the split is not mistaken for independence. That is the distinction that would otherwise produce a triple with the wrong preconditions — and stating it at the site is worth more than stating it here.

## ⭐ The hazard that materialised, predicted in advance

`emitProgramR` emits **no trailing newline**. Before conversion both halves were string literals and `executionRequestsHashFunction` ended `"  ret\n"`, so `a ++ b` concatenated correctly. After conversion both end mid-line and `a ++ b` renders `  ret erh_hash_one:` — which does not assemble.

⚠️ **And every per-function check stays green**, because `check-asm-to-program` assembles each converted def in isolation and never the concatenation. So the requirement *appeared because of the conversion*, silently — the second instance in one PR of a per-unit gate being blind to a composition property. Fixed with an explicit `++ "\n" ++` and the reason recorded in the docstring.

That the plan flagged it in advance and it then happened exactly as written is the strongest argument for writing the hazard down before starting.

## What landed

`executionRequestsHash_prog` (135 instructions) and `erhHashOne_prog` (23), both `#guard`ed, with reloc tables and `rfl` `_eq_prog` drift guards produced mechanically by `asm_to_program.py rewrite`. Two fixtures, two MANIFEST rows. `GuestAddrs` +3 symbols (`erh_hash_one`, `erh_digests`, `erh_blob`), 1020 → 1023. `GuestImageEntries` 340 → **342**, both programs registered — which is what makes a triple statable at all, since `GuestImage.lean:64-69` quantifies over `guestImageEntries`.

⚠️ For the follow-up triple: the validation contract is the deliverable #11578 actually wants, and this PR deliberately stops short of it. Its endorsement was for `execution_requests_hash` as the retarget; that target now exists as a `Program`, so the next PR has something to state a triple *over*.

Issue #11715, opened 2026-08-07
eip7702_authority_asof reads execFlags@+96 value 8 (created-this-tx) as delegated_before_tx


pirapira:
## Candidate LIVE defect found while settling #11706's naming conflict: `eip7702_authority_asof` reads the created-this-tx bit as `delegated_before_tx`

Split out of #11706 (docs fix in #11714) because it is **not** a naming question. Read from `origin/main` at `0b5825682`.

### The chain

1. **`+96` value 8 means created-this-tx**, established from the writers. The only three call sites that put value 8 into `execFlags@+96` are CREATE paths, each passing `a7 = 27` (= 16+8+2+1): `BlockVerdictCreationStage` (`bv_create_addr`), `CreateCodeEffectLog` (the CREATE code publication), `CreateFrameDescend` (`create_address_be`). The one non-create exec-flags writer passes `a7 = 0x33` (51), which does **not** contain value 8. The reader agrees by name and mask — `account_writes_created_contains`, `.Lawc_key`: `ld t1, 96(t5); andi t1, t1, 8`.

2. **The tx→block merge copies `+96` verbatim.** `.Lawb_no_flags`: `andi t3, t2, 16; beqz …; ld t3, 96(a0); sd t3, 96(t5)` — so a CREATE-path row's `27` propagates from `tx_account_writes` into `account_writes`.

3. **`account_writes_auth_block` forwards that block-map word to the auth resolver.** After requiring the `+112` components mask to hold 2, 8 and 16, it does `ld t1, 96(t5); sd t1, 0(s2)`.

4. **`eip7702_authority_asof` then tests value 8 and calls it delegation:**

```
ld t0, 48(sp); andi t0, t0, 8; snez a2, t0    <- a2 = delegated_before_tx
```

⇒ so an address whose block-map row was written by a CREATE path can be reported to EIP-7702 auth preparation as **`delegated_before_tx = 1`** while carrying no delegation designator at all.

### Why it matters, and why it is not obviously benign

`delegated_before_tx` decides whether `AUTH_BASE` is re-charged — the same quantity #11310 was about. A spurious 1 means the authority is treated as already-delegated, which is a **gas** difference, and under-charging is the FA-ward direction.

⚠️ Note the two readers are **consistent with each other about the bit's value and inconsistent about its meaning**: both do `andi …, 8` on the same `+96` word, one calling it created-this-tx and one delegation. Only one can be right, and the writers decide it.

[...]

PR #11723, opened 2026-08-07
fix(7702): read delegation marker from block code

pirapira:
## Summary

- Stop inferring EIP-7702 delegation state from `EXEC_FLAGS`.
- Return the account code pointer and length from `account_writes_auth_block` when the code field is present.
- Make `eip7702_authority_asof` recognize the EIP-7702 designator only when the block-row code is exactly `EF 01 00` with length 23; `EXEC_FLAGS` remains flags-only.

## Spec alignment

The pinned execution-specs revision is `e5a8caf1b8055e4d805c7fb169edfa710914b7da`.

The authority routine has two deliberately different state queries:

- [`apply_body` processes transactions in block order](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/fork.py#L900-L913), while [`process_transaction` creates a transaction state over the block state](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/fork.py#L1010-L1045) and incorporates it after execution ([`state_tracker.py`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/state_tracker.py#L832-L855)). This makes earlier committed transaction writes visible through the block tier.
- [`validate_authorization` reads the current authority code and nonce from `message.tx_env.state`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/vm/eoa_delegation.py#L165-L203). The guest's `account_writes_auth_current` scans the transaction map before the block map for this current nonce/status. Each accepted tuple is recorded into the transaction map before the next authorization tuple is visited.
- The `delegated_before_tx` value is intentionally different: [`set_delegation` obtains it with `get_pre_state_account`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/vm/eoa_delegation.py#L241-L276), and [`get_pre_state_account` explicitly reads block writes then immutable pre-state, excluding current transaction writes](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/state_tracker.py#L112-L176).
- The same-tx history is tracked separately by `delegation_set_for`: the spec charges `AUTH_BASE` only for a net-new delegation, at most once per authority, and retains the charge across a set-then-clear sequence ([`eoa_delegation.py`](https://github.com/ethereum/execution-specs/blob/e5a8caf1b8055e4d805c7fb169edfa710914b7da/src/ethereum/forks/amsterdam/vm/eoa_delegation.py#L211-L233)).

Thus this patch reads the block-row code marker for `delegated_before_tx`, including changes committed by earlier transactions in the same block; it does not substitute a parent-state snapshot or the pending current-transaction overlay. The transaction tier remains the source for current nonce/status and same-transaction authorization sequencing.

## Verification

- `lake build EvmAsm.Codegen.Programs.TxIntrinsicStateGas`
- Structural spec-alignment checks passed.
- Paired EEST A/B spike run: 200 selected with seed 11715; 178 matches and 22 FR on both legs, zero FA/errors, identical FR label sets, zero new false accepts, and byte-identical verdict output on all 198 joined cases.

The EEST result is reported as a no-deterioration measurement only, not as correctness evidence for the fix. The tested corpus does not witness the repaired branch.

## Reproducibility

Both legs were built from the exact paired base commit `b7acc3480bac5238ce71078c0f1da221a0e84d5d`. This is intentionally stale relative to the then-current main `12dd7a416`; the stale base is named so the paired comparison is auditable.

- Base ELF SHA-256: `b1d7cf54ba0b4be520c4d2447db3f85cd270b2813c98ece8680921c1cd94fb37`
- Candidate ELF SHA-256: `e8dabfac4f170d9e226ff6d0f29b6b78393e7639adc26449b0f803911724832e`
- Fixture tag: `tests-zkevm@v0.6.2`


pirapira:
Holding review on one substantive point — raised in parallel with this PR being opened, so it is not a response to the body.

**The cited evidence supports a three-level read; the patch implements one level.**

The body cites `state_tracker.py`'s read chain as **tx writes → block writes → pre_state**, and `eoa_delegation.py:165,180-187` reading authority code from `message.tx_env.state` with `:208-229` mutating that same tx state. The first level of that chain is the *transaction*. This patch reads the **block row** only.

That correctly rules out pre-tx parent state — which was the question I asked, and the citations settle it. But it leaves the top level of the chain unimplemented.

**The distinguishing case is inside a single transaction:** a type-4 tx carrying two authorization tuples for the same authority, or any delegation set earlier in the same tx and read again. The spec reads *and* mutates `tx_env.state`, so tuple 2 must observe tuple 1. If our MTx runtime folds a tx into the block row only after the tx completes (`read_sets_incorporate_tx` / `account_writes_incorporate_tx`), then during the tx the block row cannot carry that tx's own authorizations — making this fix correct *across* transactions and still wrong *within* one.

⚠️ **Direction matters here.** Missing an earlier same-tx delegation makes the authority look non-delegated — the **same undercharge direction** as the bit-8 defect this PR sets out to fix. So the failure mode would be half the original defect surviving the fix, which is worse than it sounds: the issue would read as closed.

Three things to settle before this is reviewable:

1. Does anything consult the **tx** row at this site, or is block the only level read?
2. If the tx level is absent, is the same-tx multi-tuple case actually **reachable** at `authority_asof`, or is it handled before the call?
3. If reachable and unhandled, the fix needs the full **tx → block → pre_state** walk rather than a block read.

⚠️ I am asserting the runtime ordering in (1) from memory of our Lean, not from a fresh read — treat it as the hypothesis to check, not as a finding. If the tx row *is* consulted, this objection dissolves and the PR stands as written.

Two smaller notes, neither blocking:

- The branch is `codex2/11683-orphaned-routines`, which names a different issue than the fix. Cosmetic, but it will mislead anyone tracing #11715 later.
- Flagging explicitly that the base `b7acc3480…` intentionally predates main `12dd7a416`, both legs paired on it, and that r200 is presented as no-deterioration only rather than correctness evidence — all three are stated in the body, and all three are the right calls.

pirapira:
**Retracting my objection above — it was wrong, and the block-only read is correct.**

I verified against the pinned spec myself (`e5a8caf1b805`, `amsterdam` fork — not `bpo2`; my first grep looked in `bpo1/2/3` and found nothing, which was my error, not a refutation).

`vm/eoa_delegation.py:265-273`:

```python
pre_state_authority_account = get_pre_state_account(tx_state, authority)
pre_state_authority_code = get_code(tx_state, pre_state_authority_account.code_hash, authority)
delegated_before_tx = is_valid_delegation(pre_state_authority_code)
```

and `state_tracker.py:145` documents `get_pre_state_account` as returning "the `Account` object at an address that existed **before the current transaction**".

So this field is deliberately read from a view that **excludes tx writes** — the name `delegated_before_tx` is literal. My three-level-chain argument took the *general* read chain and applied it to a field the spec specifically reads pre-tx. Reading the block row is right, and making `a2` tx-first would have been a defect: it would charge `AUTH_BASE` after a same-tx clear.

The same-tx repeat case I was worried about is handled by a **separate** mechanism, not by the code read — `delegation_set_for` (`:250`, `:278`, `:280`):

```python
if not delegated_before_tx and authority not in delegation_set_for:
    charge_state_gas(evm, StateGasCosts.AUTH_BASE)
delegation_set_for.add(authority)
```

So "delegated earlier in this tx" suppresses the second `AUTH_BASE` charge through set membership, while the code read stays pre-tx. Two mechanisms, two purposes — collapsing them into one tx-first read is exactly the mistake I was pushing toward.

⭐ Worth recording that this was settled by the agent **checking** rather than complying: I asked for verification of my runtime-ordering hypothesis, and the answer came back distinguishing `validate_authorization`'s full-`tx_state` read from `delegated_before_tx`'s pre-tx read — a distinction my objection had flattened.

**No objection remains. Labelling.**

One follow-up, separate from this PR and not blocking it: whether our guest implements the `delegation_set_for` suppression at `:278` at all. That is an independent charge-side question about repeated authorizations for the same authority within one tx, and it should be its own issue rather than scope creep here.

PR #11725, opened 2026-08-07
chore: batch-merge #11721 + #11723

pirapira:
Batches two `merge!`-labeled PRs into one integration PR to amortize CI.

- #11721 — proof(#11578): pin the ten SSZ constants, and bridge `rv64_remu` to Nat
- #11723 — fix(7702): read delegation marker from block code

Merged with real merge commits (`--no-ff`) in ascending PR order; both asserted ancestors of this head. No conflicts.

#11721's head advanced while this batch was verifying (it merged `main`), so I merged the updated head and re-asserted both ancestors rather than landing the stale one.

Verified locally: `lake build` green (4739 jobs) and all standing gates `ok`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #11713, opened 2026-08-07
bug: 00619/00623 still FR after the #11697 authority fix — sender balance short 1,229,200, an extra presence row, and 2935/4788 storage deltas missing from AW


pirapira:
Split from #11697, whose root cause (an uninitialised `a7` at the auth writer) is separately fixed. These two rows recover their **authority** fields under that fix and still fail for reasons unrelated to delegation liveness. Evidence from codex2, paired legs, both ELF shas named. (Issue generated by Claude Code.)

## Provenance: why these are known to be a separate defect

Under the #11697 fix, **every authority field on 00619/00623 moved toward the reference** — nonce 1 → 2 (matching post), code pointer `b8c30e48` → 0, `code_len` 23 → 0 (matching post), flags 0 → 2 (live), and the row became **counted** rather than stale (parent had `tx_account_writes_count = 1` with a populated `row1`). Both rows nonetheless remain FR.

⇒ So the delegation half is fixed and this is what is left. Both rows have **two accepted authorizations** (`eip7702_auth_state_prepare` invocation count 2, both passing `0=0` then `1=1`), which distinguishes them from the single-failing-authorization rows in the same window.

## The three deviations

Named-deviation analysis on paired parent/candidate legs, `aw_count = 4`, `bv = 1` (root mismatch) on both:

1. **Sender balance short by 1,229,200.** `ccbcccb8575994e288e5bd021cc371dd8a8ff1eb`: guest `999999999999999999996723826` vs post `999999999999999999997953026`.
2. **An extra guest presence row** — `2adc25665018aa1fe0e6bc666dac8fc2697ff9ba`, absent from post.
3. ⭐ **Two post storage deltas absent from AW** — for `0000f90827f1c53a10cb7a02335b175320002935` and `000f3df6d732807ef1319fb7b8bb8522d0beac02`.

## ⚠️ Item 3's addresses are the EIP-2935 and EIP-4788 system contracts, and that surface just changed

`0000f908…2935` is the HISTORY_STORAGE (EIP-2935) predeploy; `000f3df6…beac02` is BEACON_ROOTS (EIP-4788). **#11691 landed today**, converting both from formula-modelling to **real dispatcher execution** (`process_block_start_system_transactions`), which is exactly the mechanism that produces their storage effects.

⇒ **So a post storage delta for those two addresses missing from the account-write map is adjacent to a change that landed hours ago, and that adjacency should be checked before anything else.** Two readings, and they lead different places:

[...]

PR #11726, opened 2026-08-07
fix(codegen): full is_account_alive for top-level NEW_ACCOUNT (#11713)

pirapira:
## Summary

Closes 00619/00623 (`bal_7702_delegation_{clear,update}` sponsored) **NEW_ACCOUNT** overcharge.

**Stacked on #11723** (merge! / batch #11725): `delegated_before_tx` from block-row EF0100 code. Prior fix-2 `execFlags` value-4 delegated bit is **dropped** — that re-encoded as guest state what the pinned spec derives from CODE (`eoa_delegation.py:265-273` at e5a8caf1b).

### Fix (this PR only): full `is_account_alive` map probe

`topLevelValueRecipientStateGasAsm` used only `account_writes_lookup_current` (STATE bit). Component-only BALANCE|TOUCHED from a prior value transfer → miss → tx1 re-charges NEW_ACCOUNT (183600).

**Spec (e5a8caf1b):**
- `interpreter.py:285-288` — charge when `value>0 ∧ ¬is_account_alive`
- `state_tracker.py:445-463` — `is_account_alive = exists ∧ ≠ EMPTY_ACCOUNT`
- `EMPTY_ACCOUNT` = zero balance **and** zero nonce **and** empty code

Alive if **any** of {nonzero bal, nonzero nonce, nonempty code}.

**After lookup_current miss**, tx-then-block-cumulative (not created-set):
1. `account_writes_latest_balance` — OR-limbs ≠ 0 ⇒ alive
2. `account_writes_latest_nonce_tx` then `_block` — nonce ≠ 0 ⇒ alive
3. Dual-tier HAS_CODE (mask 4) + `codeLen@+88` ≠ 0 ⇒ alive

## Evidence

Base = `origin/main` @ `0e58d580e` **+ #11723**.

| leg | ELF sha256 | text |
|-----|------------|------|
| base (+11723) | `ec6465a9cdd7c8de4c4fa7511a42087ecbf4306e083daa7559801c98ee6a82cd` | 0x54a8c |
| fix | `88b3201c2ef13256fe424274263b6ed9823c1eb8acbfc102a1709bd0ccee5848` | **0x54c04** |

data 0x5370 / bss 0x19029540 unchanged. Pins 2nd pass NO-OP at 0x54c04.

| fixture | base tot_state | fix |
|---------|----------------|-----|
| 00619 clear sponsored | fail `[218790,183600]` | **succ=1** `[218790,0]` |
| 00623 update sponsored | fail `[218790,183600]` | **succ=1** `[218790,0]` |

(Base already has #11723: AUTH_BASE residual gone; remaining delta is pure NEW_ACCOUNT.)

**r200** seed=`932419178` spike A/B vs base+11723:
- FA 0→0, FR 20→4 (−16), **NEW_FA=0** (P→F=0)
- self-checks PASS; ELFs distinct

## Review response

- Fix 2 dropped; #11723 wins on design (code bytes, not flag).
- Predicate widened from balance-only to full EMPTY inverse.

No self-label.


pirapira:
Holding this one. Fix 1 is right in shape but incomplete; fix 2 collides with an already-labelled PR that fixes the same defect a different way.

## Fix 2 is superseded by #11723 — please drop it

#11723 (`fix(7702): read delegation marker from block code`) is **already labelled `merge!` and carried in batch #11725**. It fixes the *same* defect — `authority_asof` treating `execFlags@+96` value 8 as `delegated_before_tx` — by reading the **`EF0100` designator from the block-row code** (`AccountWriteMap` offsets 80/88 under the CODE mask), rather than by introducing a flag bit.

This PR instead defines **`execFlags` value 4 = delegated**, sets `a7 = 2 | (s11?4:0)` at the auth write, and reads it with `andi …, 4`. Both land in `TxIntrinsicStateGas.lean`, so they will conflict textually — and more importantly they disagree on design.

⭐ **#11723's approach is the spec-aligned one and should win.** The pinned spec (`e5a8caf1b`, `amsterdam`) computes `delegated_before_tx` from account **code**, via `get_pre_state_account` (`vm/eoa_delegation.py:265-273`) — there is no delegation flag in the spec's model at all. A new `+96` bit re-encodes as guest state something the spec derives from code, which is the shape we have been retiring, not adding. It also has to be *maintained* correctly at every writer, whereas the code bytes are already authoritative.

Concretely: once #11723 lands, tx1 reads tx0's installed designator out of the block-row code and sees `delegated=1`, so the `AUTH_BASE` re-charge on 00623 is fixed without a new bit. **Please rebase onto a base containing #11723 and drop fix 2**, then re-measure — I expect 00623 to still go F→P on fix 1 alone plus #11723.

## Fix 1: right mechanism, but the aliveness test is too narrow

The diagnosis is good and the citation is correct: `interpreter.py:285-288` charges `NEW_ACCOUNT` when `value > 0 ∧ ¬is_account_alive`, and `state_tracker.py:445-463` defines `is_account_alive = exists ∧ ≠ EMPTY_ACCOUNT`. Using `account_writes_latest_balance` over **tx then block-cumulative** maps (not the created-set) is the correct baseline, and avoiding the created-set sidesteps the revert hazard.

⚠️ **But "nonzero balance ⇒ alive" is sufficient, not necessary.** `EMPTY_ACCOUNT` is zero balance **and** zero nonce **and** no code. An account with **zero balance but nonzero nonce or non-empty code is alive**, and this fix would still charge `NEW_ACCOUNT` for it.

That population is ordinary, not exotic: a deployed contract holding no ether, or an EOA that has sent transactions and spent its balance to zero. Either one receiving a value transfer hits the residual.

Direction is **overcharge ⇒ FR-ward**, so this is not a soundness regression and the PR still strictly improves on main — but it leaves a live piece of the same bug, and closing #11713 on it would record the bug as fixed when a reachable case remains.

Please extend the test to the full `is_account_alive` predicate — balance **or** nonce **or** code non-empty — using the same tx-then-block-cumulative lookup. If any of those components is not available at that site, say which, and we scope the gap explicitly rather than silently.

## Smaller notes

- Evidence quality is good: named ELF shas both legs, `.text 0x549e0 → 0x54a30` (+0x50) with `.data`/`.bss` unchanged, pins second pass a no-op at the new size, and the `tot=[218790,183600] → [218790,0]` decomposition names exactly which charge disappeared.
- Base is `12dd7a416`; main is now `0e58d580e`, and the PR is CONFLICTING. The rebase for fix 2 handles this anyway.

pirapira:
Both review points addressed. Labelling.

**Fix 2 dropped** in favour of #11723's `EF0100`-from-code read — correct call. A `+96` flag bit would have re-encoded as guest state what the pinned spec derives from account code (`eoa_delegation.py:265-273`), and it would have needed maintaining at every writer forever, while the code bytes are already authoritative.

**Predicate widened to the full `EMPTY_ACCOUNT` inverse**, which was the substantive point: `is_account_alive = exists ∧ ≠ EMPTY_ACCOUNT`, and `EMPTY_ACCOUNT` is zero balance **and** zero nonce **and** empty code. The probe now tests all three — `account_writes_latest_balance` OR-limbs, `latest_nonce_tx` then `_block`, and dual-tier `HAS_CODE` with `codeLen@+88` — over tx-then-block-cumulative maps rather than the created-set. Using the maps and not the created-set is what keeps a reverted creation from suppressing a charge that is genuinely owed, which is the FA-ward direction and the trap on the opposite side of this fix.

The balance-only version would have left a reachable case live: a zero-balance contract, or an EOA that spent down to zero, receiving a value transfer. Overcharge rather than unsound — but closing #11713 on it would have recorded the bug fixed while that case survived.

**Evidence checks out** and is measured against the right base (`0e58d580e` + #11723, so the AUTH_BASE residual is already gone and the remaining delta is pure `NEW_ACCOUNT`):

- ELFs named both legs, `.text 0x54a8c → 0x54c04`, `.data`/`.bss` unchanged, pins second pass a no-op at the new size.
- 00619 and 00623 both `fail [218790,183600] → succ=1 [218790,0]` — the `183600` disappearing is exactly one `NEW_ACCOUNT` charge, which names *which* charge went rather than just that the row flipped.
- r200 seed `932419178`: FA 0→0, **FR 20→4**, and **P→F = 0**.

⚠️ Noting for the record that r200 P→F=0 is a no-deterioration result, not proof of correctness — a byte-identical r200 has coexisted with a real FA on this project before. What justifies this PR is the spec citation plus the named F→P pair; the sample is the guard against regression, not the argument.

Stacked on #11723, which is in batch #11725 — so this wants to land after it.

PR #11727, opened 2026-08-07
test(#10817): three row sets in bal_canonical_sort_selftest, with a set counter

dhsorens:
First increment on #10817 after #11676's predicate vocabulary. Deliberately the cheapest possible one: `bal_canonical_sort_selftest` is **probe-only and unlinked from `stateless_guest`**, asserted negatively by two `#guard`s (`BalCanonicalSort.lean:549,559` — the label and the `.globl` are *absent* from the guest aggregate), so **this cannot perturb the guest image**.

## ⚠️ The counter had to come first, and that is the interesting part

Before this, the self-test returned a bare `1` from **every** failure edge. Meanwhile:

- the probe header already said `bal_canonical_sort_selftest` **"(three row sets)"** and *"returns the failing row-set number"* (`BalSelftestsProbe.lean:11,22`);
- the runner already printed **`failed at row set {sort}`** (`scripts/codegen-zisk-bal-probes.sh:359`).

The multi-case discipline was **declared and unimplemented**. So adding cases without a set number would have made the probe *worse* — one indistinguishable code across twelve failure edges instead of four — rather than better. `s1` now carries the set number and the failure code **is** that number.

Two prose claims that were false are now true rather than needing edits. `s1` is safe across the call: `bal_canonical_sort` saves and restores `s0`-`s11` (`balCanonicalSortFunction.s:2-5`) — checked, not assumed.

## The three sets

All three use **one** segment descriptor (`a3 = 0x1400`: offset 0, width 20, endianness flag clear) and stride 128. That is deliberate: a mis-encoded descriptor produces a case that passes **vacuously**, which is the exact trap this module's header warns about, so no new case introduces descriptor-decoding risk.

**Set 1 — differ at the canonical MOST significant byte.** Unchanged, byte-for-byte. ⭐ It is the **negative control**: a byte-index sort would order by `field[0]` (all zero) and leave the rows untouched, giving tags `1,2,3,4` instead of `2,4,1,3`. Not weakened.

**Set 2 — differ at the canonical LEAST significant byte.** The same four key values moved to field byte 0. Same expected tags, reached a completely different way: the MSD radix must walk **38 all-equal nibbles** before finding a difference. The depth bound is `2 * keysum = 40`, so this sits **two nibbles inside the cap** — exercising the deep-descent path *and* its boundary. Set 1 splits on the very first nibble and touches neither.

**Set 3 — input already canonical.** Keys ascending with tags `1,2,3,4`, expected output the identity. Catches a sort that permutes when it must not: spurious swaps, or a write cursor that advances on a self-match.

⚠️ **Set 3 is not a standalone test and the docstring says so explicitly**: a sort that does nothing at all passes it. Its value is only as a complement — a do-nothing sort fails sets 1 and 2. The suite is strictly stronger than before; set 3 alone is weaker than either of the others.

## How the Program was regenerated (the part with the hazard)

`balCanonicalSortSelftestFunction` is already converted, so its Lean side is `emitProgramR` over a hand-maintained `_prog` (it is in `asm_to_program.py`'s `SOURCE_DRIFT_ALLOW`), and `asm_to_program.py rewrite` cannot re-derive it — the def is no longer a string literal. I did **not** hand-write 160 Program entries. Instead: rendered the edited fixture back into a temporary literal def, ran `emit-lean` on it, and spliced the generated `_prog` + `_relocs`.

Two adaptations to the generated output, both necessary:

1. **`GuestAddrs.bal_canonical_sort_selftest` → `balCanonicalSortSelftestPc`.** The generator assumes a linked entry; this probe is unlinked and the module explicitly forbids re-adding a `GuestAddrs` entry (`:386-388`). This is precisely why the def is in `SOURCE_DRIFT_ALLOW`.
2. The `Function` def and its drift guard keep the `"  .globl …"` prefix, which the generator does not emit and which two `#guard`s pin.

⭐ **The reloc/pc-base coupling that makes hand-editing dangerous came out right by construction.** Relocs land at indices `37, 86, 135` and the three `jal` pc-bases at `+148, +344, +540` — i.e. `37×4`, `86×4`, `135×4`. Those are two encodings of the same numbers, and a hand-edit that fixed one and not the other would assemble cleanly and jump to the wrong place. Letting the generator produce both is what avoids that.

## Guards moved

- `#guard balCanonicalSortSelftest_prog.length == 61` → `== 160`.
- The ordered expected-tag projection → `[2,4,1,3, 2,4,1,3, 1,2,3,4]`. ⚠️ This guard filters **every** `.LI .x7` in the program, so it is sensitive to any new use of `t2`; the three sets contribute exactly their twelve expectations and nothing else does. The failure code is `mv a0, s1`, not an `li`, so it does not pollute it.
- `balCanonicalSortSelftestFunction_eq_prog` is `rfl` and follows.

## Drift fixed while in these files

Both `BalSelftestsProbe.lean:52` and `BlockAccessListBuilder.lean:45` cited `BalCanonicalSort.lean:254` for "swaps rows with `ld`/`sd`". That line is now a **range-frame** load; the swap loop is elsewhere. Both now cite `scripts/asm-fixtures/balCanonicalSortFunction.s:65-67` (`.Lbalsort_swap`) instead — the fixture is stable against Lean-side line drift, which is what broke these in the first place.

## Verification

- `lake build` clean — which *is* the substance here, since every claim above is a `#guard`.
- **`check-asm-to-program.sh` genuinely ran** (RISC-V toolchain present locally) and is CLEAN. This is the load-bearing gate: it assembles the Lean-rendered string and compares `.text` against the fixture, so a wrong Program entry is caught rather than silently green.
- `check-file-size` (821 lines, well under the 1500 cap), `check-forbidden-tactics`, `check-unimported`, `check-layering`.

## What this does NOT do

No theorem. This adds **runtime** cases to a probe, and three vectors are still three vectors — the permutation obligation is unaffected. I am posting a separate assessment on #10817 of what the permutation proof actually needs, because measuring it turned up two missing pieces of vocabulary and a loop shape with no precedent in the tree.

### DIFF
--- a/EvmAsm/Codegen/Programs/BalCanonicalSort.lean
+++ b/EvmAsm/Codegen/Programs/BalCanonicalSort.lean
@@ -374,17 +374,54 @@ theorem balCanonicalSortFunction_eq_prog :
     vacuously on a limb-swapped order. So the expected permutation is computed from
     the canonical rule alone and hardcoded.
 
-    The rows differ only in field byte 19, which is the address's canonical BE MOST
-    SIGNIFICANT byte (the address occupies the low 20 bytes of an LE word). Values
-    0x30, 0x10, 0x40, 0x20 in rows 0..3, so canonical ascending order is rows
-    1, 3, 0, 2. Each row carries its 1-based tag at offset 64, so the tag sequence
-    after sorting must read 2, 4, 1, 3.
+    ## Three row sets (#10817), each reusing the same 512-byte arena
 
-    A byte-index sort would order by field[0] instead — all zero here — and leave the
-    rows untouched, giving 1, 2, 3, 4. That is exactly the failure this catches.
+    ⚠️ **`s1` carries the row-set number and the failure code IS that number**, so
+    a red probe says *which* set failed. Before this the self-test returned a bare
+    `1` from every one of its failure edges, while
+    `scripts/codegen-zisk-bal-probes.sh:359` already printed
+    `failed at row set {sort}` and the probe header already claimed "3 row sets" —
+    the multi-case discipline was **declared and unimplemented**. Adding sets
+    without the counter would have made that worse, not better, so the counter
+    came first. `s1` is safe to use across the call: `bal_canonical_sort` saves and
+    restores `s0`-`s11` (`balCanonicalSortFunction.s:2-5`).
 
-    a0 = a scratch arena of at least 4 * 128 bytes, 8-aligned.
-    a0 (out) = 0 on the expected order, else 1.
+    All three sets use ONE segment descriptor (`a3 = 0x1400`: offset 0, width 20,
+    endianness flag clear) and stride 128, so none of them can fail for a reason
+    involving descriptor decoding — that is deliberate, since a mis-encoded
+    descriptor would produce a case that passes vacuously.
+
+    **Set 1 — differ at the canonical MOST significant byte.** Field byte 19 is the
+    address's canonical BE MSB (the address occupies the low 20 bytes of an LE
+    word). Values 0x30, 0x10, 0x40, 0x20 in rows 0..3, so canonical ascending order
+    is rows 1, 3, 0, 2 and the tag sequence must read 2, 4, 1, 3.
+
+    ⭐ This is the **negative control** and must not be weakened. A byte-index sort
+    would order by field[0] instead — all zero here — and leave the rows untouched,
+    giving 1, 2, 3, 4. That is exactly the failure it catches.
+
+    **Set 2 — differ at the canonical LEAST significant byte.** The same four
+    values at field byte 0, everything else zero. Same expected tags (2, 4, 1, 3),
+    reached a completely different way: the MSD radix must walk **38 all-equal
+    nibbles** before it finds a difference. Since the depth bound is
+    `2 * keysum = 40` nibbles, this sits two nibbles inside the cap — so it
+    exercises the deep-descent path *and* its boundary, neither of which set 1
+    touches (set 1 splits on the very first nibble).
+
+    **Set 3 — input already in canonical order.** Keys 0x10, 0x20, 0x30, 0x40 at
+    byte 19 with tags 1, 2, 3, 4, so the expected output is the identity. This
+    catches a sort that permutes when it must not — spurious swaps, or a
+    write-cursor that advances on a self-match.
+
+    ⚠️ **Set 3 is not a standalone test and must never be read as one**: a sort
+    that does nothing at all passes it. Its value is only as a complement — a
+    do-nothing sort fails sets 1 and 2 — so the suite is strictly stronger than
+    before, but set 3 alone is weaker than any of them.
+
+    a0 = a scratch arena of at least 4 * 128 bytes, 8-aligned. All three sets share
+    it and each re-zeroes it, so no extra space is needed.
+    a0 (out) = 0 if every set produced its expected order, else the 1-based number
+    of the first set that did not.
 
     Probe-only: not linked into `stateless_guest`. The Program's jalOff pc-base is
     a local placeholder (emitProgramR emits a symbolic `jal` via the reloc table);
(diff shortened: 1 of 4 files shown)


pirapira:
Reviewed against the PR head rather than the description. Labelling.

**The guest-image safety claim checks out**, which is what made this cheap to review:

```
#guard (balCanonicalSortFunctions.splitOn "bal_canonical_sort_selftest:").length == 1
#guard (balCanonicalSortFunctions.splitOn ".globl bal_canonical_sort_selftest").length == 1
```

i.e. the label **and** the `.globl` are asserted *absent* from the guest aggregate, with the positive counterparts asserted present in the selftest function. And the `BlockAccessListBuilder.lean` touch is 4 insertions / 1 deletion with **zero non-comment lines** — a citation repair, not a code change.

⭐ **The `.globl` guards are the part worth pointing at**, and the docstring gives the reason: symbol *binding* is invisible to a `.text` compare — a symbol demoted GLOBAL→LOCAL, or dropped entirely, leaves `.text` byte-identical. So a byte-comparison gate cannot see the one property that decides whether this probe stays out of the guest. Asserting it separately is the only mechanical check on it.

**Fixing the counter before adding cases was the right order**, and the justification is unusually honest: the probe header already claimed "(three row sets)" and "returns the failing row-set number", and the runner already printed `failed at row set {sort}` — so the multi-case discipline was **declared and unimplemented**. Adding cases without a set number would have made the probe *worse*: one indistinguishable code across twelve failure edges instead of four. Two prose claims that were false become true rather than being edited away, which is the better direction.

`s1` safety across the call was **checked rather than assumed** (`balCanonicalSortFunction.s:2-5` saves and restores `s0`–`s11`).

**The three sets earn their places, and the PR says where each is weak:**

- **Set 1** is retained byte-for-byte as the negative control — a byte-index sort would order by `field[0]` (all zero) and produce tags `1,2,3,4` instead of `2,4,1,3`. Not weakened by the additions.
- **Set 2** reaches the same expected tags a completely different way: keys moved to field byte 0, so the MSD radix must walk **38 all-equal nibbles** against a depth bound of `2 * keysum = 40` — two nibbles inside the cap, exercising the deep-descent path *and* its boundary. Set 1 splits on the first nibble and touches neither.
- **Set 3** (already-canonical input) catches a sort that permutes when it must not, and the docstring states plainly that **a do-nothing sort passes it** — its value is only as a complement, since a do-nothing sort fails sets 1 and 2.

Using **one** segment descriptor across all three is the right call for the stated reason: a mis-encoded descriptor produces a case that passes *vacuously*, which is precisely the trap this module's header warns about. A new descriptor per case would have added decoding risk to a test whose job is to remove it.

⚠️ One note, not blocking and not a criticism of the PR: a probe unlinked from `stateless_guest` is only as good as the thing that runs it. The `#guard`s prove it stays out of the guest image; they say nothing about whether `codegen-zisk-bal-probes.sh` is exercised in CI. If it is not, this strengthens a test that nothing runs. Worth confirming separately — it does not affect this PR's correctness.

CONFLICTING against `91f9c0e02`; the actor owns the resolution from here.

pirapira:
Parking this one — it needs a rebase onto #11717, and the reconciliation is an emitted-offset decision I should not make on your behalf.

#11717 (per-arena BAL sort capacities) landed in `main` while this was queued. It adds a capacity argument at the `bal_canonical_sort` call sites, which shifts the emitted offsets. This PR rewrites those same call sites for the three row sets. Merging leaves 5 conflicting hunks — 4 in `BalCanonicalSort.lean`, 1 in `BalSelftestsProbe.lean` — of this shape:

```
<<<<<<< HEAD  (main, after #11717)
    .LI .x15 (4 : Word),
    .JAL .x1 (jalOff GuestAddrs.bal_canonical_sort (balCanonicalSortSelftestPc + 144)),
    .BNE .x10 .x0 (72 : BitVec 13),
=======       (this PR)
    .JAL .x1 (jalOff GuestAddrs.bal_canonical_sort (balCanonicalSortSelftestPc + 148)),
    .BNE .x10 .x0 (brOff (balCanonicalSortSelftestPc + 616) (balCanonicalSortSelftestPc + 152)),
>>>>>>>
```

Both sides are right about their own tree: `main` now passes the capacity in `x15`, and this PR moves to `brOff` with different offsets for the extra row sets. Combining them means recomputing the instruction offsets against a program that has both — arithmetic on emitted RISC-V, not conflict repair, so it is yours.

The `BalSelftestsProbe.lean` hunk is only a comment, and your side is the better text (it cites the fixture rather than a line number that has since drifted) — that one carries over as-is.

Nothing else is wrong: the sibling PRs #11726 and #11728 are unaffected and I am landing them separately.

Once rebased, drop the `complicated-merge` label and it will be picked up on the next pass.

Issue #10615, opened 2026-07-26
possible FA: guest accepts a BAL nonstorage declaration with zero net effect, which the spec's builder never emits


pirapira:
## The guest is LOOSER than the spec's builder for nonstorage BAL fields

While diagnosing FR root early44 (89 rows, `bv_fail=44`, the BAL-to-exec nonstorage forward comparator), the guest's matched-record comparator was found to carry a final net-change check (`pre_balance != post_balance`, `pre_nonce != post_nonce`) — **but its wrapper accepts an existing zero-net-effect record that carries a BAL declaration.**

The spec's builder emits balance/nonce/code fields **only** when the pre-transaction value differs from the post value (`block_access_lists.py:637-664`), with the pre value taken from cumulative block writes first (`_get_pre_tx_account`, `:583-598`).

So a BAL that *declares* a balance or nonce change which had no net effect is:

- **rejected by the spec** — the builder would not have emitted that field, so `hash_block_access_list(built) != header.block_access_list_hash`;
- **accepted by the guest** — the wrapper tolerates the declaration.

That is the **false-accept** direction, and it is the direction FR-hunting is structurally blind to. It was found only because someone read the comparator while chasing a false reject.

## Why this is filed separately from #10612

#10612 records the general obligation (each of ~21 field-wise checks must be exactly equivalent to `provided == build_block_access_list(execution)`, neither stricter nor looser) and the maintainer's decision to defer convergence. This is a **named instance in the looser direction**, which is the half that no amount of valid-corpus testing surfaces — so it needs its own bead rather than living in a comment.

## Required before escalation

Per project policy, an FA claim needs an empirical witness, not an argument:

1. **Construct the KAT.** Take a passing fixture, add a balance-change or nonce-change declaration to the BAL for an account whose net effect over the transaction is zero, and recompute `header.block_access_list_hash` so the block is internally consistent with the *spurious* BAL.
2. **Show the guest ACCEPTS it.** If it rejects, this is not an FA and the finding reduces to a redundant-but-harmless tolerance.
3. **Show the spec REJECTS it** — the builder omits the field, so the computed hash differs from the header.

Both directions must be demonstrated before this is treated as a live soundness hole.

## Note on the interaction with the open hash question

[...]

PR #11728, opened 2026-08-07
fix(bal): reject zero-net nonstorage BAL declarations (#10615)

pirapira:
## Summary
Tighten `bal_account_nonstorage_consistent` to match pinned builder
`e5a8caf1b` `block_access_lists.py:637-664`: if exec `pre==post` for
balance/nonce but BAL still declares the field, reject.

Nonce-only changes still emit/accept nonce alone (`hasBalance=0`) so
last-post-wins aggregation does not invent a balance component.

## Roundtrip fix (handback)
Earlier commit dropped `_prog` while keeping Function + MANIFEST row.
Restored via `asm_to_program.py rewrite` so the three agree:
- `check-guestaddrs-starts`: OK (342)
- `check-asm-to-program`: CLEAN (382)

## Evidence
| | sha256 | text |
|---|---|---|
| main | `008fa63567f34e22…` | 0x54a8c |
| cand | `7734897c54feda37…` | **0x54a84** (−8) |
data/bss unchanged. pins 2nd pass NO-OP.

r200 seed=932419178: FA0→0 FR20→20 P→F=0 byte-identical 198; FR set eq PASS.

## Gates
- `check-guestaddrs-starts` OK
- `check-asm-to-program` CLEAN
- lake build module green

No self-label.


pirapira:
Reviewed and labelling. This closes #10615's alignment work; the audit behind it is recorded on the issue.

**The change is what the spec says.** `bal_account_nonstorage_consistent` now rejects a declaration whose field did not net-change (`pre == post ∧ hasField`), matching the pinned builder at `block_access_lists.py:637-664`, which emits each field only when `pre != post`. The prior docstring stated the tolerance explicitly — *"deliberately does NOT reject an account whose final equals its pre-value with a spurious declaration"* — so this replaces a documented deviation with the spec's rule rather than patching around one.

⭐ **The nonce-only property is preserved and called out in the docstring**, which is the constraint that mattered most here: a nonce-only change still emits only the nonce field (`hasBalance=0`), and this comparator never invents a balance component. `nonstorage_effect_aggregate` copies the *first* record's pre-values but overwrites post from the **last** record in the group, so a same-to-same balance declaration would **replace a true final balance** rather than merely duplicating it. A repair that normalised by writing a balance component would have been actively harmful.

**`.text` shrinks by 8 bytes** (`0x54a40 → 0x54a38`). Worth noting the direction: a tightening that *removes* an arm rather than adding a retraction is the shape we want — the tolerance was code, and deleting it is what alignment looked like here.

**Regeneration is complete and correct**: `GuestAddrs.lean`, `GuestImageEntries.lean`, `RegionMapLinkPins.lean`, `symbol-addresses.tsv` and the fixture `.s` all move together, pins second pass a no-op, `.data`/`.bss` unchanged.

**Measurement, on `91f9c0e02`:** ELF base `4ef7850aa7d793e0…` → cand `2b42d373bb12cfb0…`; r200 seed `932419178` gives FA 0→0, FR 20→20, **P→F = 0 with the named set empty**, FR set equality PASS, 198 byte-identical.

⚠️ Reading that result correctly matters more than the number. Our builder provably cannot emit a zero-net-effect record (all three guards confirmed skip-on-equality, arena-reach census confirms no fourth emitter), so **an empty P→F set is the expected outcome, not corroboration that the change is right.** What it does establish is a real fact about the corpus: no third-party BAL in this sample carries a spurious declaration either. Had the set been non-empty, that would have been information about the corpus rather than a regression to tune away.

The justification is spec alignment on its own terms — no fixture gain is claimed, and none was available to claim.

pirapira:
Parking this one — it fails two gates, and both reproduce **standalone on this branch with no merge involved**:

```
check-guestaddrs-starts:
  only 381/382 manifest rows have a parsed Function binding
  (missing e.g. ['balAccountNonstorageConsistentFunction']) — refusing to emit

check-asm-to-program:
  Unknown identifier `balAccountNonstorageConsistent_prog`
```

Cause: the emitted Program def was removed while its `Function` string and its MANIFEST row were kept.

- `def balAccountNonstorageConsistentFunction` is still present (`BalAccountNonstorageConsistent.lean:56`), with its `#guard`s
- `scripts/asm-fixtures/MANIFEST.tsv` still has the `balAccountNonstorageConsistentFunction` row
- but `balAccountNonstorageConsistent_prog` is **absent** on this branch, where `main` has 6 occurrences of it

So the roundtrip has a fixture and a manifest row pointing at a Program that no longer exists. Either the `_prog` needs restoring (regenerating it from the updated `.s`), or — if the routine is genuinely going away — the MANIFEST row and fixture should go with it. Which of those you intend is your call; I am not going to delete a manifest row to green a gate, since that would drop byte-pinning for a routine that still has a live `Function`.

The sibling #11726 is unaffected and I am landing it separately.

Once resolved, drop the `complicated-merge` label and it will be picked up on the next pass.

Issue #11362, opened 2026-08-04
Guest CREATE2 collision-check rejects EIP-6780 resurrection (fixture 01087)


pirapira:
Attribution: codex2

# BAL producer false-reject on CREATE2 resurrection (fixture 01087)

## Summary

The current `stateless_guest` rejects a valid Amsterdam EIP-7928 fixture. This is a false reject, not a false accept:

- fixture: `01087_test_bal_create2_selfdestruct_then_recreate_same_block_fork_Amsterdam-blockchain_test-with_balan`
- input SHA-256: `efca24a5da22daed5c6177864139e7bc594def5c4ff5557052bac2407974edb7`
- manifest SHA-256: `b40b7e3f7d4b3bde03c7298c98d3b9117ea146f76c2a2a914f1cbafa78c3ad07`
- payload BAL SHA-256: `e23e4b3c9ee57d390fde810b638dc662c6ac91a576b7f7e5824df852d52e24b7`
- execution-specs pin: `e5a8caf1b8055e4d805c7fb169edfa710914b7da`
- guest ELF SHA-256: `6a7242fc3049b5fec53bb3612fcf244f47e6693933d8712cf31efd7659a11c7e`

Normal SPIKE execution (without dump instrumentation) returned a 256-byte output with `successful_validation` at `OUTPUT+32 = 0`, while the fixture oracle is `1`. The production diagnostics report `bv_fail_code` at `OUTPUT+112 = 1`, `bv_bal_shadow_status` at `OUTPUT+120 = 1`, and rebuilt/supplied BAL lengths `460/494` at `OUTPUT+128/+136`.

## Spec expectation

The pinned execution-specs test documents Tx1 as `CREATE2+SSTORE+SELFDESTRUCT` and Tx2 as CREATE2 resurrection at the same address (`test_block_access_lists_opcodes.py:3812-3829`). For target A `2dc7dbabd911be419bc27b24a2ebb401dd779735`, the expected BAL is:

- BAI 1: the SSTORE is demoted to `storage_reads`; the destruction has no nonce/code changes.
- BAI 2: `nonce_changes` contains `post_nonce=1` and `code_changes` contains the 28-byte runtime `61cafe60075573c0f6dc9e5836f54caadbf59cc69346c508e1992bff`.
- `storage_changes` is empty.

[...]

Issue #11363, opened 2026-08-04
CREATE creator nonce is not consumed on ALL branches (spec increments on both; success branch also affected, FR-backed by 01087)


pirapira:
Attribution: codex2

## Summary

The guest with ELF SHA-256 `6a7242fc3049b5fec53bb3612fcf244f47e6693933d8712cf31efd7659a11c7e` does not advance a CREATE or CREATE2 creator nonce when the creation attempt fails. The increment is gated on successful child completion, but the Amsterdam execution spec consumes the creator nonce at the creation attempt before collision, OOG, or revert is known.

This is independent of the CREATE2 collision and tombstone defect tracked in #11362. Fixing the collision predicate can make fixture 01087 green while leaving this reject-side defect present for every genuinely failing CREATE or CREATE2.

## Evidence

Fixture: `01087_test_bal_create2_selfdestruct_then_recreate_same_block_fork_Amsterdam-blockchain_test-with_balan`

The pinned spec performs:

1. Tx1 factory `b9c464e2bc20bfaf549a32da6d1d1d8adcbf6952` CREATE2 creates target `2dc7dbabd911be419bc27b24a2ebb401dd779735`, then selfdestructs it.
2. Tx2 the same factory attempts the same CREATE2. The target is recreated under the Amsterdam rules, but the candidate guest returns zero on the collision path.

A final-memory SPIKE dump of `create_nonce_table` at `0xbbd9b4b0` shows:

- factory entry next nonce = 2
- recreated target entry next nonce = 1

After Tx1 the factory next nonce is 2. After Tx2 attempt, the spec requires 3 even if the CREATE2 fails. The guest remains at 2, and its BAL output omits the factory BAI2 nonce=3 row.

Relevant guest symbols:

- `create_creator_nonce_use`
- `create_nonce_table`
- `create_nonce_table_count`

The source comments describe `create_creator_nonce_use` as advancing the running nonce on CREATE/CREATE2, but the failed CREATE2 path never reaches it.

## Expected fix shape

Move creator nonce consumption to the CREATE attempt path before the success or child execution gate, while preserving rollback semantics for a reverted frame only where the execution spec requires rollback. Add reject-side tests covering collision, OOG, and revert, and keep this issue open even if #11362 lands.

## References

- #11362: Guest CREATE2 collision check rejects EIP-6780 resurrection
- execution-specs Amsterdam CREATE semantics and creator nonce handling

pirapira:
[...]

PR #11729, opened 2026-08-07
fix(dispatch): auth-phase OOG is not deferred recipient fail (#11305)

pirapira:
## Summary
#11305: pre-dispatch / authorization-phase OOG must not be classified as deferred recipient failure (`a0=8` → `bv_fail=47`).

### Classification (not a fixture win)
Established on main before this PR: the deferred∧status=2 path is **not hit**
today. Named `set_delegation_oog_*` (7 rows) already full-PASS; final state shows
`deferred=0` `auth_halted=0` after ca18f406f. This change is a **source
classification correction** (auth-phase OOG → ordinary settlement when
`auth_phase_halted`, not `a0=8`/bv_fail=47). No available fixture gain — a green
sweep must not be read as buying a row. Alignment is its own justification.

### Stage (not value)
- `BlockVerdictDispatchTx` deferred seam: `prepare_prefix_status=2` previously always returned `a0=8`.
- Mtx maps `a0=8` → `.Lbv_mtx_recipient_unresolvable_fail` → code 47.
- Spec (e5a8caf1b): `interpreter.py:123-125` runs `set_delegation` before `process_message`; `test_authorization_oog.py:143-169` requires atomic auth-phase rollback — recipient is never loaded on auth OOG.

### Fix
On deferred ∧ status=2:
- `auth_phase_halted=1` → clear deferred, continue ordinary settlement
- `auth_phase_halted=0` → keep `a0=8` (true completed-prep unresolvable)

### Empirics
| | base | cand |
|---|---|---|
| commit | `803053608` | this PR |
| ELF sha | `008fa63567f34e22…` | `460d1cb136e33ae8…` |
| text | 0x54a8c | **0x54aac** (+0x20) |
| data/bss | 0x5370 / 0x19029540 | unchanged |

- Named `set_delegation_oog_*` (7 inputs): already **PASS** on base (path not exercised today after related auth-OOG work); still PASS on cand.
- r200 seed=932419178: FA0→0 FR20→20 **P→F=0** FR set equal; byte-identical 198.
- Pins second pass **NO-OP**.

No self-label. Does not touch CREATE2 / 01087 (#11362).


pirapira:
Reviewed and labelling.

**Correct framing: this is a stage defect, not a value defect.** The deferred seam returned `a0=8` unconditionally on `prepare_prefix_status=2`, and Mtx maps `a0=8` to `.Lbv_mtx_recipient_unresolvable_fail` → code 47. But an authorization-phase OOG is not a recipient-resolution failure — `interpreter.py:123-125` runs `set_delegation` **before** `process_message`, and auth-phase OOG rolls back atomically, so the recipient is never loaded. Splitting on `auth_phase_halted` puts each cause in its own arm rather than letting one code stand for two unrelated conditions.

⭐ **The honesty about the empirics is what makes this reviewable**, and I want it on record rather than buried: the named `set_delegation_oog_*` inputs (7) **already pass on base** — the path is not exercised today after related auth-OOG work — and they still pass on the candidate. So there is **no fixture gain available to claim**, and the PR does not claim one. r200 P→F=0 with 198 byte-identical is a no-deterioration result, not evidence the change is right.

That is the correct justification under the standing rule: matching the spec's classification needs no KAT and no row. A misclassification that is currently unreachable is still a misclassification, and the next change that makes the path reachable would otherwise inherit a wrong `bv_fail` code — with the symptom landing far from the cause, since code 47 names the wrong subsystem entirely.

**Mechanics check out:** `.text 0x54a8c → 0x54aac` (+0x20), `.data`/`.bss` unchanged, pins second pass a no-op, base and candidate ELF shas both named against `803053608`.

**Scope is stated and clean:** does not touch CREATE2 / 01087 (#11362). Worth noting explicitly because #11363 is in flight on that fixture, and a reviewer seeing two dispatch-adjacent PRs tonight could reasonably wonder whether they interact. They do not.

PR #11731, opened 2026-08-07
fix(create): consume creator nonce at attempt

### LINKED ISSUE #11363: CREATE creator nonce is not consumed on ALL branches (spec increments on both; success branch also affected, FR-backed by 01087)

pirapira:
Fixes #11363

## Summary

- The pinned Amsterdam `system.py` increments the sender nonce before child
  execution on both the collision/non-deployable and deployable paths
  (`e5a8caf1b8055e4d805c7fb169edfa710914b7da`, lines 118 and 132).
- The emitted ordering now records `create_creator_nonce_use` at
  `ChildFrameCreateTail.lean:201-205`, then descends at line 471; the child
  checkpoint is captured later by `CallFrameDescend.lean:482-490`. Thus a child
  checkpoint is after the attempt, while an ancestor checkpoint predates nested
  attempts, and the existing positional journal undo restores the ancestor on
  parent revert.
- Creator-nonce effects are logged unconditionally in the pre-descent
  deployable and collision arms. Cross-transaction seed lookup uses the
  block-cumulative nonce map; the later same-transaction liveness lookup stays
  transaction-local.
- The 11362 collision predicate is unchanged; this is the 01087 nonce-row fix,
  not a collision-predicate change.

## Evidence

- Base: `91f9c0e023442fffc8e35840defe25c202f024b9`, ELF SHA256
  `7697162b77b8a98499e3ad95888aa5c2ddf7668af4dad607320e4f1d129c6e38`.
- Candidate: `8d783cf636e7339fa511ac4f5ff3144ee384ea45`, ELF SHA256
  `4cb0c99285876712a8ad31f90bb0f4ad45cb93790252b8a2fb9d74ec6dc74d49`.
  Both runs used clean repositories; the candidate provenance is recorded in
  `gen-out/eest-run/run-20260807T210702Z-2503918/run-provenance.tsv`.
- Exact BAL 01087 candidate result: attempted/decoded `27/27`; final counts
  accounts `11`, balance `6`, code `1`, nonce `5`, storage `4`; every row diff
  is empty; rebuilt and supplied serializer hashes match at
  `5e1e6f41d1eede889dbab59af2d8a0cf374bf388be45b3c450a5a59cada442c6`.
  The base omitted creator row
  `b9c464e2bc20bfaf549a32da6d1d1d8adcbf6952`, BAI 2, post-nonce 3.
- Ancestor-revert vectors 03242 (CREATE) and 03246 (CREATE2) both
  prefix-match the expected stateless output on base and candidate. The
  per-leg output hashes are respectively
  `84da5814ba4ba3f6bc7848ff9e7bf453bfa3cfe871cb829d000f539d2ca592b6` and
  `2049a67e6f82ffe1bdd56b7a3ce3a155d367b0b16c8865af9a806f13c0a12ea3`.
  This shows no deterioration and preserves parent-revert rollback.
- All four generated layout/link artifacts were regenerated. A required
  second complete regeneration pass was a content no-op. The linked layout
  has `.text` base `0x80000000`, `.text` size `0x54a20`, and `.data` base
  `0x5370`.

## Verification

`lake build` and `check-no-warnings.sh` pass (4739 jobs, zero warnings), as do
the forbidden-tactics, hardcoded-PC, heartbeat, layering, opcode-structure,
roundtrip-coverage, region-map, file-size, unimported-module, progress/drift,
asm-to-program, and axiom gates. The axiom gate reports 146 witnesses and only
the allowed classical axioms.


pirapira:
Reviewed and labelling. This is the third design for #11363 and it is the one that adds nothing — which is why it is right.

## The correctness case is an ordering argument, and it belongs in the body where you put it

The spec increments the creator nonce **before** child execution on **both** paths — collision/non-deployable at `system.py:118`, deployable at `:132` — i.e. at the attempt, before the outcome is known.

The emitted ordering now matches: `create_creator_nonce_use` at `ChildFrameCreateTail.lean:201-205`, descent at `:471`, and the child checkpoint captured later at `CallFrameDescend.lean:482-490`. So the child checkpoint is **after** the attempt while an ancestor checkpoint **predates** nested attempts, and **plain positional undo already gives both behaviours**: `undo_to(child)` cannot reach the increment, `undo_to(ancestor)` can.

⭐ Two earlier designs were proposed and rejected — a `persistent` tag on journal entries, then a partial restore that skipped only the nonce field. Both would have encoded the rule as a **per-record property** that every future writer must set correctly, and the partial-restore version was **wrong**: it would have made an ancestor revert *fail* to roll the nonce back, converting an FR into an FA. The landed fix instead puts the mutation in the right place relative to an existing checkpoint and deletes the special case entirely.

Worth stating as a general lesson: **when a rollback rule appears to need a special case, the likely defect is the order of the mutation against the checkpoint, not the rollback rule.**

## Evidence

- Paired legs, both clean: base `91f9c0e02` / ELF `7697162b…`, candidate `8d783cf63` / ELF `4cb0c992…`, candidate provenance recorded.
- **01087 exact BAL match**: attempted/decoded 27/27; counts accounts 11, balance 6, code 1, nonce 5, storage 4; every row diff empty; rebuilt and supplied serializer hashes equal at `5e1e6f41…`. The base omitted exactly one row — `b9c464e2…`, BAI 2, post-nonce 3 — which is the named row this issue is about, not merely a fixture that went green.
- **Ancestor-revert path is measured, not argued**: `03242` (CREATE) and `03246` (CREATE2) `nested_create_fail_parent_revert_state_gas` both prefix-match expected output on base and candidate. Because both legs *pass*, this establishes the nonce is rolled back correctly there — a passing row matches the spec post-state, and the nonce is part of it.
- All four layout artefacts regenerated with a second complete pass that was a content no-op; `.text` size `0x54a20`.
- `lake build` and `check-no-warnings.sh` clean at 4739 jobs; forbidden-tactics, hardcoded-PC, roundtrip-coverage, region-map, asm-to-program and axiom gates all pass, with the axiom gate reporting 146 witnesses and only the allowed classical axioms.

## Scope

`#11362`'s collision predicate is explicitly unchanged, and the PR says so. That matters here more than usual: fixing the collision predicate can turn 01087 green while leaving this reject-side defect live for every genuinely failing CREATE, so a green 01087 alone would not have distinguished the two. The named-row evidence is what does.

Issue #11724, opened 2026-08-07
readability: .L77prep_auth_charge is the SKIP path, not the charge path (AUTH_BASE gate is correct)


pirapira:
**Not a behavioural defect — the gate is correct. This is about a label whose name states the opposite of what it does.**

## The gate

`TxIntrinsicStateGas.lean:255-286` charges `AUTH_BASE` for an EIP-7702 authorization iff:

- `s11 != 0` — the delegation target is non-null (`s11 = 0` on `NULL_ADDRESS`), and
- `120(sp) == 0` — `delegated_before_tx` is false, and
- `168(sp) == 0` — the `teer_success_table` charged word at `+20` is still clear.

This matches the pinned spec (`e5a8caf1b`, `amsterdam`) — `vm/eoa_delegation.py:278`:

```python
if not delegated_before_tx and authority not in delegation_set_for:
    charge_state_gas(evm, StateGasCosts.AUTH_BASE)
delegation_set_for.add(authority)
```

with the guest's `charged@+20` word playing the role of `delegation_set_for` (`sw zero, 20(t4)` on a new-authority append, `lw t0, 20(t4)` on an existing-row hit, `1` stored after charging). **Repeat authorizations for the same authority within one tx are correctly suppressed.** Verified as part of #11715; no fix is needed.

## The hazard

All three conditions branch to a label named **`.L77prep_auth_charge`** — and that label is the **skip** path. The name asserts the opposite of its function, and it sits directly adjacent to a refund/reservoir path (`:277`, `runtime_tx_auth_state_charge`), so the surrounding text reinforces the wrong reading.

⚠️ **Two readers misread it in one evening**, including me: I read the same lines while checking this mechanism, could not resolve the polarity from the branch structure, and said so rather than asserting it. The correct polarity was established only by an explicit statement of which register value causes a skip.

The concrete risk is not confusion but a **wrong fix**: a future reader who takes the label at face value will conclude that a null target *causes* an `AUTH_BASE` charge, and will "repair" a gate that is already correct — in the overcharge direction, which is FR-ward and would present as a fixture regression rather than an obvious error.

## Suggested change

Either rename to `.L77prep_auth_charge_skip` / `.L77prep_no_auth_base`, or add a comment above the three branches stating the charge condition positively:

```
-- AUTH_BASE is charged iff s11 != 0 (non-null target) AND 120(sp) == 0
[...]

PR #11732, opened 2026-08-07
refactor(codegen): rename AUTH_BASE skip label (#11724)

pirapira:
## Summary
`.L77prep_auth_charge` in `eip7702AuthStatePrepareFunction` was the **SKIP** path for AUTH_BASE (null target / `delegated_before_tx` / already charged this tx). The name asserted the opposite of its function — two readers misread polarity in one evening (#11724).

## Change
1. Rename `.L77prep_auth_charge` → **`.L77prep_no_auth_base`** (3 branches + label). Sibling labels `_done` / `_reservoir` / `_used` / `_oog` (apply-charge path) unchanged.
2. Lean comment **above** the emitted string (not inside asm) stating the charge condition positively:
   - charge fall-through iff `s11 ≠ 0` ∧ `120(sp)==0` ∧ `168(sp)==0`
   - each failed conjunct → `.L77prep_no_auth_base` (SKIP)
   - spec cite `e5a8caf1b` `eoa_delegation.py:278`

## ELF byte-identical (no pin regeneration)
`.L`-prefixed labels are assembler-local; they do not enter the symbol table and do not change instruction encoding.

| leg | sha256 | text |
|---|---|---|
| base (main) | `5c9c16a6209cf4b2…` | 0x54c04 |
| cand | `5c9c16a6209cf4b2…` | 0x54c04 |

**Same string.** Therefore no GuestAddrs / symbol-addresses / RegionMapLinkPins regen is owed — a reviewer seeing an assembly-source rename should not expect the four-artefact layout churn.

## Gates
- `check-asm-to-program`: CLEAN (382) — Function string changed; assembled bytes agree
- lake codegen green

No r200 (byte-identity is the stronger bar here). No self-label.

### DIFF
--- a/EvmAsm/Codegen/Programs/TxIntrinsicStateGas.lean
+++ b/EvmAsm/Codegen/Programs/TxIntrinsicStateGas.lean
@@ -277,8 +277,13 @@ def eip7702AuthStatePrepareFunction : String :=
   -- passes -1 and retains the aggregate compatibility path.
   "  ld t0, 104(sp); bnez t0, .L77prep_no_new; la t0, runtime_tx_auth_state_refund; ld t1, 0(t0); li t2, " ++ toString amsterdamNewAccountStateGas ++ "; add t1, t1, t2; sd t1, 0(t0); la t0, runtime_tx_auth_state_charge; ld t1, 0(t0); add t1, t1, t2; sd t1, 0(t0)\n" ++
   ".L77prep_no_new:\n" ++
-  "  beqz s11, .L77prep_auth_charge; ld t0, 120(sp); bnez t0, .L77prep_auth_charge; ld t0, 168(sp); bnez t0, .L77prep_auth_charge; la t0, runtime_tx_auth_state_refund; ld t1, 0(t0); li t2, " ++ toString (amsterdamStateBytesPerAuthBase * amsterdamCostPerStateByte) ++ "; add t1, t1, t2; sd t1, 0(t0); la t0, runtime_tx_auth_state_charge; ld t1, 0(t0); add t1, t1, t2; sd t1, 0(t0); li t0, 1; sw t0, 20(t4)\n" ++
-  ".L77prep_auth_charge:\n" ++
+  -- AUTH_BASE is charged (fall-through) iff s11 != 0 (non-null delegation
+  -- target) AND 120(sp) == 0 (not delegated_before_tx) AND 168(sp) == 0
+  -- (not already charged this tx via teer_success_table +20). Each failed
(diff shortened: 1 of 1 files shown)


pirapira:
Reviewed against the branch head and labelling. This closes #11724.

## The acceptance bar for this change was byte-identity, and it holds

A `.L`-prefixed label is assembler-local: it never enters the symbol table and does not affect instruction encoding. So a pure rename must leave the emitted image **bit-for-bit unchanged**, and that is a far stronger check than any sweep would give.

**Base and candidate ELF sha are the same string** — `5c9c16a6209cf4b2a43626fdd9629883d52676bb9809143a7187061d9e30cfc6`, `.text 0x54c04` on both. Consequently **no pin regeneration is owed**, and its absence here is a property of the change rather than an omission. `check-asm-to-program` is clean, which is the gate that would have caught a Function-string/`_prog` divergence.

## The rename is complete and the siblings are intact

`.L77prep_auth_charge` has **0** occurrences; `.L77prep_no_auth_base` has **3** (the label plus both branch targets). The four siblings — `_done`, `_reservoir`, `_used`, `_oog` — are **untouched**, which matters: they are legitimately named for what they do, and a blanket rename would have made them wrong in the opposite direction.

## The comment states polarity positively, which is the part that prevents recurrence

```
-- AUTH_BASE is charged (fall-through) iff s11 != 0 (non-null delegation
-- target) AND 120(sp) == 0 (not delegated_before_tx) AND 168(sp) == 0
-- (not already charged this tx via teer_success_table +20). Each failed
-- conjunct branches to .L77prep_no_auth_base — the SKIP path, not the
-- charge path (GH #11724). Spec: e5a8caf1b eoa_delegation.py:278.
```

⭐ This is the fix for the actual hazard. The gate was always **correct** — three conjuncts matching `eoa_delegation.py:278`, with the guest's `charged@+20` word playing the role of the spec's `delegation_set_for`. What was wrong was that the label asserted the opposite of its function, and it sat adjacent to a refund path referencing `runtime_tx_auth_state_charge`, so the surrounding text reinforced the misreading. **Two readers misread it in one evening, including me** — I read those exact lines while checking the mechanism, could not resolve polarity from the branch structure, and had to have it stated explicitly.

The risk was never confusion; it was a **wrong fix**. A reader taking the old label at face value would conclude that a null target *causes* an `AUTH_BASE` charge and would "repair" a gate that is already correct — in the overcharge direction, surfacing as a fixture regression rather than an obvious error. That is the same near-miss shape as #10918, where a fixture built for a defect that did not exist would have passed for the already-correct reason.

The comment lives in the Lean above the emitted string, not inside a converted asm block, so it does not count as drift.

PR #11735, opened 2026-08-07
fix(tests): align the M22 storage-preload cases with validate-and-skip (#11710)

### LINKED ISSUE #11710: test: sload_preloaded_match expects the M24 preload materialized, but the parser validates-and-skips (expectation contradicts contract)
Found while enumerating M24 storage-preload consumers for #11355 (now closed). Evidence from scoper, every path read from `origin/main` at `0b5825682` via `git show`/`git grep` against the ref. (Issue generated by Claude Code.)

## The contradiction

**`sload_preloaded_match`** asserts `expectedOutHex` beginning **`adde0000`** — i.e. it expects the preloaded slot value `0xdead` to be readable, since `0xdead` in limb 0 little-endian is `ad de`.

**`Dispatch.lean:2569-2575`** states the current contract: the legacy preload rows are **validated and skipped**, and **no storage arena is materialized**. `:2743` says the same from the other side — *"Legacy preload rows are not copied or re-tagged: no persistent log remains."*

⇒ **If `sload_preloaded_match` were run today it would read zero, not `0xdead`, and fail.**

## And the prose still describes the old behaviour

`Cases.lean:2470-2473` still documents the dispatcher as materializing the preload. So a reader consulting the test table gets the pre-change contract, while the parser has the post-change one.

## Why it is latent rather than failing

`Cli.lean:149` publishes `tc.storage` as **field 5** of the `--list-test-cases` TSV to an external runner, and its own docstring records that **the bash runner asserts only the non-empty ones**. So whether this case is currently exercised depends on a runner outside this repository.

⚠️ **That is the part that needs establishing first**, and it decides the disposition:

- **If the case is run** — it is failing now, or its assertion is being skipped by the non-empty filter, and either way the state should be visible rather than inferred.
- **If it is not run** — the case is a stale expectation encoding a contract that no longer holds, and it will bite whoever next wires that runner in.

## The two cases

| case | `storage` | expectation |
|---|---|---|
| `sload_preloaded_match` | `(0x00, 0xdead)` | `expectedOutHex` starts `adde0000` — **requires materialization** |
| `sload_preloaded_no_match` | `(0x00, 0xdead)` | reads a *different* slot, so zero is the correct answer either way |

⭐ Note the asymmetry: **`no_match` passes under both contracts** — it reads a slot the preload does not cover, so zero is right whether or not the preload materialized. Only `match` discriminates. So a runner that exercised both and reported one failure would look like a single flaky case rather than a contract change, and a runner that exercised only `no_match` would report all-green while the contract had silently inverted.

## Disposition options

1. **Restore materialization** — only if some consumer needs it. #11355 established the packer, the test table and the external runner all still *feed* the format, but feeding it is not the same as needing the rows materialized.
2. **Update the expectation and the prose** to the current skip-and-validate contract, making `match` expect zero. Cheapest, and correct if nothing needs materialization.
3. **Retire both cases** if they only ever tested a mechanism that is gone.

⇒ **(2) is most likely right**, but it should not be chosen before question 1 above is answered — otherwise the expectation gets bent to match the implementation without establishing that the implementation is what anyone wants. That is the same trap as regenerating a fixture until it passes, which came up on #11431 and needed a four-citation spec chain to resolve honestly.

## Bar

Whichever option: state whether the case is currently executed by any runner, and cite the contract (`Dispatch.lean:2569-2575`) beside the expectation so the two cannot drift apart again. If the expectation changes, the prose at `Cases.lean:2470-2473` changes in the same commit.

Related: #11355 (closed — where this surfaced; the parser stays because the format has live consumers).

pirapira:
Closes #11710.

## What was wrong

The dispatcher prologue **validates and skips** the legacy M24 storage-preload rows and materializes nothing (`Dispatch.lean:2567-2570`, `:2743` — *"not copied or re-tagged: no persistent log remains"*). Four cases in `opcodeTestCases` still encoded the pre-retirement contract.

**The test side is the defect, not the parser.** Three reasons, strongest last:

1. execution-specs (pin `e5a8caf1b`) has no "preload" concept — `SLOAD` reads state, and pre-state storage comes through the authenticated path. An input-file segment that materializes storage is an **unauthenticated** pre-state channel, so restoring it would move *away* from the spec.
2. Nothing consumes materialization (per #11355: the packer and TSV still *feed* the format, but feeding is not needing).
3. ⭐ **The materialization target does not exist.** `evm_slot_table` has exactly one occurrence in the whole tree — the stale comment being replaced here — and **zero** in the emitted image.

## What changed

| case | before | after |
|---|---|---|
| `sload_preloaded_match` | `expectedOutHex` `adde…` — required materialization | expects **zero**; documented as *the* guard on the retirement |
| `sstore_preloaded_nonzero_update_gas_exact` | `gasLimit 13006`, 2-entry log, premise *"the original is non-zero"* | renamed **`sstore_preloaded_key_zero_origin_gas_exact`**, `gasLimit 110926`, 1-entry log |
| `sload_preloaded_no_match` | correct | comment states it does not discriminate |
| `sstore_overwrites_preload` | correct | comment states it does not discriminate |

The gas case is the one that fails hardest: with the row skipped the write is a **zero-origin creation**, so 13006 is a hard **OOG** rather than a 32-byte diff. Both corrected numbers are taken from its preload-free sibling `sstore_cold_gas_exact`, which is the identical write: `6 + 13000 + 97920 = 110926`, appending one entry.

The section header claimed the prologue copies the segment into a writable `evm_slot_table` and records the count at `env.slotTableCountOff = 448`. **Both are wrong:** the arena does not exist, and env+448 is now the *retired persistent-log counter* that `Dispatch.lean:1148` hard-zeroes — the live slot count goes to `648(x20)`.

## Verification

**Guest-neutral, measured rather than inferred.** Rebuilding after the edit reproduces ELF sha256 `723a67964ba71304c0a11accb0a0e8f486134f89fd81da304066f7e9133b6e1f` — byte-identical to the pre-edit build at the same base — with the spike fixture run unchanged at **fail 0 / full match 2**. `lake build` clean; all four cases verified to emit correctly through `--list-test-cases`.

## ⚠️ Two things this PR does NOT fix

1. **Nothing runs these cases.** `CODEGEN.md:429` names the runner as `scripts/codegen-opcodes-check.sh`, which was **deleted** in `a027b07ad` (*"delete 776 ungated codegen-*-check.sh probes"*, #10642). Repo-wide, `--list-test-cases` appears only in `CODEGEN.md` prose and `Cli.lean`, and no workflow references a test case. So **all 375 cases** in the table are unexecuted, and this contradiction sat undetected because of that — the expectations here are corrected but still unexercised. That gap deserves its own issue rather than being folded in.
2. **13 cases assert `expectedPersistentLogLength`** with nonzero values while the dispatcher states the persistent log is retired. I am *not* claiming they are broken — the closing condition is to identify which output offset that field is diffed against, which cannot be read with the runner gone. Out of scope here; the two values changed in this PR are grounded in an in-table sibling rather than in that field's semantics.


pirapira:
Reviewed against `origin/main` and labelling. Verified the two load-bearing claims rather than taking the body's word.

**The contradiction is real and current.** `Dispatch.lean:2567-2570` says outright: *"The persistent execution log has been retired, so consume these rows to reach the trailers but do not materialize them. The canonical transaction map and authenticated state path are authoritative."* Four cases in `opcodeTestCases` still encoded the pre-retirement contract.

⭐ **And the decisive argument is the third one, which is why it belongs last:** `evm_slot_table` — the materialization target — has **one** occurrence in the whole tree (the stale comment this PR replaces) and **zero** in the emitted image. Confirmed independently: the only other hit is `CODEGEN.md`. A test asserting a materialization into a table that does not exist cannot be testing what it claims to test.

**Deciding that the test is the defect rather than the parser is the right call, and for the spec reason rather than the convenience one.** execution-specs at `e5a8caf1b` has no "preload" concept — `SLOAD` reads state, and pre-state storage arrives through the authenticated path. An input-file segment that materializes storage would be an **unauthenticated pre-state channel**, so restoring the old behaviour would move *away* from the spec. That argument stands on its own even if nothing consumed the format; the "nothing consumes it" evidence (#11355) is corroboration, not the basis.

⚠️ Worth stating plainly for the record: a test that asserts a materialization which never happens **cannot fail for the right reason**. Either it fails always, or it was weakened at some point and is now green while asserting nothing. This PR converts four such cases into assertions that the retirement *holds* — `sload_preloaded_match` now expects zero and is documented as the guard on it. That is strictly more testing than before, not less, which is the opposite of how a "fix the test" change usually reads.

Scope is one file, `EvmAsm/Codegen/Tests/Cases.lean`, so no emitted-image change and no pin regeneration is owed.

Issue #11737, opened 2026-08-07
test tooling: codegen-eest-stateless-check.sh exits 0 with failures (648 rows, 116 fail, exit 0) — exit-code checks report false green


pirapira:
`scripts/codegen-eest-stateless-check.sh` **exits 0 even when rows fail**, so any caller that gates on exit status reports a clean run on a failing one.

## Observed

On `2ed2c844d`, ELF `723a67964ba71304c0a11accb0a0e8f486134f89fd81da304066f7e9133b6e1f`, backend spike:

```
--filter selfdestructing_initcode_preserves_balance --limit 700
→ selected 648, ran 648, errored 0, FAIL 116, full match 532
→ exit status 0
```

**116 failures, exit 0.** The verdict is only available in the summary block (`fail` / `full match`); the exit status carries no information about it.

## Why this is worse than an ordinary bug

⭐ **It fails silently in the safe-looking direction.** A tool that errors loudly gets fixed. A tool that reports success when it should report failure gets *trusted*, and every downstream conclusion inherits the error — including issue closures, which is where it does lasting damage.

The blast radius is any automation or agent that shells out and checks `$?`: CI wrappers, batch scripts, and the measure-then-close loop we ran all evening. Reading a fail count by habit is what saved us here; nothing in the tool's contract requires it.

## Suggested fix

Exit non-zero when `fail > 0` or `errored > 0`, keeping 0 for a clean run. If some caller depends on the current behaviour (e.g. a script that wants the summary regardless of outcome), add an explicit `--no-fail-exit` rather than leaving the default unsafe — the default should be the one that cannot silently mislead.

Also worth considering: emit the fail count on stderr as well as in the summary, so a caller that captures only stderr still sees it.

## Related

Third measurement-instrument defect found the same evening:

- #11736 — the leaf decoder reports `validMask`-clear fields as mismatches, producing false residuals.
- A comment body written to a bare `/tmp` path published another identity's stale content while `gh` reported success (fixed by per-session scratch paths).
- This one.

None of the three affected the guest. All three affected what we believed about the guest.

(Issue generated by Claude Code.)

Issue #11736, opened 2026-08-07
test tooling: the leaf decoder reports unset (validMask-clear) fields as mismatches, producing false residuals


pirapira:
The BAL/account-writes leaf decoder reports a **mismatch** on fields whose `validMask` bit is clear — i.e. it reads an unset field as data and compares it against the expected value.

## Observed

While diagnosing #11327 on `2ed2c844d`, the decoder reported two residual "mismatches" that are not defects:

- **Row `6295ee1b4f6dd65047762f924ecd367c17eabf8f`** — `vmask 0x3a`. The balance bit (`0x1`) is **clear**, so the balance field is *unspecified*, not zero-valued. The decoder (line 223) flags it anyway, because the expected balance is nonzero.
- **Row `2adc25665018aa1fe0e6bc666dac8fc2697ff9ba`** — `vmask 0x20`, TOUCHED-only. That is an execution-map / root-enumeration fact, not a post-state account delta, but it surfaces as a "guest-extra presence".

Separately, two post storage deltas were reported as "absent from the account-writes arena" — correct by design, since `storage_writes` is a separate structure and `execution_map_state_changes` enumerates the union.

In all three cases the guest root **exactly equalled expected**, and the fixture passed.

## Why this matters more than a cosmetic complaint

⭐ **A flag says whether a field is meaningful; the field says what it is.** Reading an unset field as data is the defect this decoder has — and it is the instrument we use to decide *what to investigate*. During #11327 these residuals survived a correct fix and read as "the defect is not fully fixed", which is exactly the shape that sends an agent after a second mechanism that does not exist.

⚠️ The cost is asymmetric and invisible: a false residual costs an investigation, and nothing in the output distinguishes it from a real one. Tonight it was caught only because the vmask was read by hand.

## Suggested fix

Gate each field comparison on its `validMask` bit before comparing, and report unset fields as **unspecified** rather than as a value. Where a row is TOUCHED-only, label it as an enumeration entry rather than a post-state delta. Where a structure legitimately has no counterpart row (storage deltas vs the account-writes arena), say so rather than reporting absence.

Not urgent — no guest behaviour is affected — but it is a measurement-instrument defect, and those propagate into every investigation that uses the instrument.

(Issue generated by Claude Code.)

pirapira:
Attribution: scoper

# Second observation: the decoder also UNDER-reports — a single `primary_*` hid a second diverging leaf

[...]

PR #11739, opened 2026-08-07
Fix pre-funded CREATE live balance

### LINKED ISSUE #11327: fc=1 row 16403 (test_transaction_collision_to_empty2): validated oracle + collision check recorded, diagnosis not started
Groundwork only — the diagnosing agent's shift ended before it ran the guest. Recorded so the next
picker-up starts from a validated oracle instead of re-deriving one.

## The row

| | |
|---|---|
| index | **16403** (family is 2 rows: 16403 / 16404) |
| family | `test_transaction_collision_to_empty2` |
| input size | 5,400 bytes — the shortest fc=1 row outside the claimed clusters |
| **wrapped input sha** | `3d72fef707a58e3073e11cb42bfa954a26b6c8e1c79415e680f68b9e60ecc577` |
| manifest sha | `b40b7e3f7d4b3bde03c7298c98d3b9117ea146f76c2a2a914f1cbafa78c3ad07` |
| fail code | 1 (terminal state-root comparison) |

⚠️ Key on the **wrapped input sha**, not the index. A row index is per-manifest and the same index names
different fixtures across manifests — that produced a false convergence story today (#11308 covers the
related col8 point: it is a converter `case_id`, not a file hash).

## The oracle is already validated

A reference run on pinned `e5a8caf1b` produced a **69-byte** output, sha
`2ed505045bd4584fda1379673a87256dc26d69d92b03b988f47050b8b8b6a93e`, **matching the manifest expectation**.

⭐ That is the step most diagnoses skip, and it is what distinguishes *a wrong guest* from *a wrong
expectation*. It is done — do not redo it, and do not proceed without it if you change fixture.

## Collision check — already done

Confirmed **outside** both live clusters at the time of writing:

- **not** in the same-tx create+selfdestruct group (the `accounts_to_delete` mechanism — see below)
- **not** in the EIP-7702 pointer / `set_code` cluster (#11323 and the fc=47 work)

⚠️ **Re-check this before starting.** If the `accounts_to_delete` fix has since landed, several fc=1
families go green for free and this row's priority changes.

## Suggested method (what worked on the sibling rows today)

1. Reference pipeline reproduces the expected root → oracle validated (**already done**).
2. Dump both leaf sets. ⛔ **Assert both counts and the captured byte length before diffing, or refuse to
   diff** — a truncated capture yields a confidently wrong answer.
3. Diff as **sets** first (missing / extra account), then per-field for addresses in both.
4. Only then go to node level if values and sets both match.

⚠️ A **value sweep is the wrong first instrument**: 696 value-mutation candidates on a sibling row
produced zero hits this morning, and a set diff found the defect in one pass.
⚠️ If the differing leaf is a `storage_root`, it is still tractable **when the BAL names the slots** — a
finite set diff. It is only a dead end if you would have to enumerate the trie, which is keyed by
`keccak256(slot)`.

## Priors worth checking first — today's four mechanisms were all *shape* problems

Every fc=1 defect found today was recording-before-gating rather than a wrong computation:

- a scratch cell **read on a path that never wrote it** (#11315/#11316 — closed 30 rows, not the 5 claimed)
- an **unconditional seed before an existence check** (#11321/#11324 — absent EIP-2935/4788 storage seeds)
- a **missing container**: no `accounts_to_delete` for same-tx created-and-selfdestructed accounts
- an **empty row emitted where the spec emits nothing** (fc=60's `+27`, ~124 rows, in progress)

⇒ Check whether `transaction_collision_to_empty2` is a fifth instance before assuming a new mechanism.
The family name suggests a *collision with an empty account*, which is exactly the territory where
"materialised but should be absent" errors live.

pirapira:
## Summary

Fix the top-level CREATE live-balance seed for pre-funded recipient accounts. The creation stage now adds the authenticated pre-balance and transaction endowment with u256_add_be, then reverses the result into the live EVM environment before initcode dispatch. The authenticated pre-balance and endowment buffers remain intact for the effect and transfer records.

The existing creation-stage contract is:

> process_create_message credits the newly-created account with tx.value
> before initcode executes. The authenticated pre-balance is combined with
> that endowment below, after the lookup and before the dispatcher enters
> initcode, so SELFDESTRUCT and SELFBALANCE observe the spec's live account
> balance.

This matches the pinned execution-specs commit e5a8caf1b8055e4d805c7fb169edfa710914b7da: state_tracker.py lines 646-677 implement recipient balance as pre-existing balance plus the transferred amount, while lines 377-388 permit deployment based on nonce, code, and storage rather than requiring zero balance.

Fixes #11327

## Validation

Named false-to-pass set, matched by input content:

- Row 16403, case bf0cd26d5febc67657b091f744b6a8f8916e163d29ed2c9863b87f9ddd7ad7e, input SHA 3d72fef707a58e3073e11cb42bfa954a26b6c8e1c79415e680f68b9e60ecc577.
- Row 16404, case 85b633abe0fbe7a81d3494632b137365328025f33675f1c91494c76e918fbaf0, input SHA 9c85e027daeaa7ef704def453058f5cdaa06478da6fc5918a9c17dca2582e459.

Using Spike and reading the summary fail counts explicitly:

- Base ELF SHA 723a67964ba71304c0a11accb0a0e8f486134f89fd81da304066f7e9133b6e1f: selected 2, ran 2, fail 2, full 0.
- Candidate ELF SHA 5a4e04d37ea711e8fdfc62da2b49551f926e5498c410c1feae874df02b5733ec: selected 2, ran 2, fail 0, full 2.
- AB comparison: self-check passed; output-byte differences 2 for the named pair, with no status differences.

The candidate ELF was rebuilt from committed tree 54c311635c15454296ca51bd967a9381afd20ac2 and reproduced the measured candidate SHA.

Random regression guard, seed 4242, uniform 200 of 26104 fixtures, identical Spike manifests:

- Base: selected 200, ran 200, fail 3, full 197.
- Candidate: selected 200, ran 200, fail 3, full 197.
- AB comparison: self-check passed, identical three-label FR set, zero output-byte differences, zero status differences.

The r200 sample did not exercise a top-level CREATE to a pre-funded address; it is a regression guard. The named pair and execution-spec citation are the direct correctness evidence.

The remaining leaf-decoder reports are contract-aware diagnostics: the reported balance row has vmask 0x3a and therefore no balance-valid bit; the 0x20 row is touched-only; storage deltas belong to storage_writes rather than account_writes. For the named pair, the emitted BAL shadow status changes from 1 on base to 0 on candidate, confirming the rebuilt BAL digest now matches.

The inserted u256_add_be leaf writes only t0-t6 and a0, never s0-s3. t0-t3 are re-established at the reversal loop setup, lines 555-558, and the authenticated lookup precedes the addition.

The committed image extents are .text 0x54c2c to 0x54c6c, a 0x40 increase; .data is byte-identical; .bss moves from 0x19029540 to 0x19029560 for the 32-byte scratch. GuestImageEntries is unchanged, as expected for this address-only image change.


pirapira:
Reviewed and labelling. Closes #11327.

⭐ **The strongest line in this PR is the quoted creation-stage contract**, and it should be read as the argument rather than as context. The existing docstring already said the pre-balance is combined with the endowment *"so SELFDESTRUCT and SELFBALANCE observe the spec's live account balance"* — while the code seeded the endowment alone. **The author's intent already matched the spec; only the code did not.** That is the exact inverse of #11724, where the code was right and the label lied, and it is why this is a repair rather than a design change.

**Spec grounding** is on the pin (`e5a8caf1b`): `state_tracker.py:646-677` implements recipient balance as pre-existing balance **plus** the transferred amount, and `:377-388` permits deployment on nonce/code/storage rather than requiring zero balance. So a pre-funded address is deployable *and* keeps its balance — both halves needed, both cited.

**Evidence.** Named F→P pair matched **by input content, not row index** (16403 / `3d72fef7…`, 16404 / `9c85e027…`) — the right identification method, given a bare index is not stable across converter runs.

| leg | selected | ran | fail | full |
|---|---|---|---|---|
| base `723a6796…` | 2 | 2 | **2** | 0 |
| candidate `5a4e04d3…` | 2 | 2 | **0** | 2 |

Explicit fail/full counts rather than a bare PASS — necessary here, because `codegen-eest-stateless-check.sh` exits 0 with failures (#11737), so an exit-status read would have reported green on the base leg.

Candidate ELF **rebuilt from committed tree `54c311635`** and reproduced the measured SHA, so post-measurement edits were inert and the measurement transfers to what is being merged.

⭐ **r200 is correctly labelled a regression guard, not corroboration**: both legs 200/200, fail 3, identical three-label FR set, zero output-byte differences — and the body states plainly that the sample never exercises a top-level CREATE to a pre-funded address. Byte-identity here is the *expected* result; reading it as evidence the fix works would be the error.

**Residuals are classified, not waved away** — `vmask 0x3a` has no balance-valid bit so that field is unspecified rather than zero, the `0x20` row is touched-only, and storage deltas belong to `storage_writes` rather than `account_writes` (the decoder's over-reporting is #11736). And the **BAL shadow status goes 1 → 0** for the named pair, confirming the rebuilt digest now matches — that is the check a wrong-but-root-matching fix would have failed.

**Register audit** present for the inserted `u256_add_be` call, which is the hazard when adding a call to a stretch that had none: a restore reads exactly like a definition in a diff.

Scope is three files with `.bss` growth for the new scratch — the correct placement, since `.data` growth breaks address-pinned `rfl` proofs.

Issue #11306, opened 2026-08-03
account_writes_lookup_current reads execFlags@+96 without its validMask bit, so a balance-only live account reads as deleted (EXTCODEHASH returns 0; 107,920 gas is downstream)


pirapira:
## Summary

Follow-up to **#11299** / **PR #11304** (tx-finalize destroyed-norm). That PR fixed all **36** `success-self-*` eip8246 initcode rows (FR 68→32). Residual **32** rows are all `success-eoa-*` (16) + `success-precompile-*` (16).

**Not** a destroyed-table identity / `created_in_tx` bug (earlier provisional read was wrong).

## One-row witness: 04243

- Label: `…success-eoa-no_storage…` (post_send present; victim = CREATE child)
- input: `/var/tmp/fc668/04243_test_selfdestructing_initcode_preserves_balance_fork_Amsterdam-blockchain_test-success-eoa-no_st.input`
- Parent: fc44 equal-len. After #11304 cand: still **fc44** rb=sp=593
- Guest cand sha `ad165acf…` (build 2026-08-03 15:51:30 +0200)

### CREATE child identity (SPIKE)

| event | address |
|---|---|
| `create_frame_descend` `create_address_be` | **`7e68b428f271eba557f0d619786bd75ec543efe9`** |
| destroyed table at `nonstorage_apply_destroyed_norm` | **same** `7e68…` meta=0 (distinct beneficiary) |
| first comparator fail `bal_all_accounts_nonstorage_consistent` | **same** `7e68…` |

Destroyed table holds the CREATE child correctly. Not entry-vs-child mis-identity.

### BAL (execution-specs decode)

Child `7e68…`: **balance bai=1 post=1** (post_send delivered to victim), **no nonce**.

Sibling empties/others as expected (EOA target `6e5c…` bal=1, factory nonce, etc.).

### Guest AGG after #11304 apply

Child `7e68…`: **mask=0**, bal 0→0, nonce 0→0.

Forward nonstorage comparator fails: BAL declares bal post=1; exec aggregate has no balance component.

### Raw log before apply (why wipe is wrong)

Child rows include CREATE endowment, CREATE nonce 0→1, SD clear bal→0, then **later bal 0→1** (post_send re-credit). Distinct branch of `nonstorage_apply_destroyed_norm` sets **every** matching raw record to `post=pre` + `mask=0`, which deletes the post_send balance that EIP-8246 clear must **preserve**.

[...]

PR #11740, opened 2026-08-07
fix(scripts): fail the EEST run on fixture failures (#11737) and refuse wrong-shape verdict decodes (#11738)

### LINKED ISSUE #11737: test tooling: codegen-eest-stateless-check.sh exits 0 with failures (648 rows, 116 fail, exit 0) — exit-code checks report false green

### LINKED ISSUE #11738: test tooling: format_verdict_debug's published offsets describe the DIAGNOSTIC build, not the production output — decoding a production .out there yields plausible garbage
`format_verdict_debug` in the harness reads the first **168 bytes as 21 `u64` words**, placing `verdict` at `+0`, `tx_state0` at `+104` and `tx_state1` at `+112`. The **production** stateless output has a different layout: bytes `0..32` are the `new_payload_request_root` and byte `32` is `successful_validation`.

**Both layouts cannot describe the same file.** The harness fields come from a **diagnostic build**, not from the production output.

## How it surfaces

Decoding a production `.out` at the published offsets yields plausible-looking garbage rather than an error — e.g. `verdict` decoded as `1841024047515375962`. Nothing signals that the layout is wrong; the numbers simply mean nothing.

⚠️ The offsets are **published in the harness**, so they look authoritative. Any consumer who reaches for them against a production artefact gets values that are the right width, in the right place, and unrelated to what they name.

This was caught only by hexdumping instead of trusting the offsets, while bisecting gas boundaries for #11306. Two figures previously reported on that issue (`tx_state0=367200`, `tx_state1=195840`) are **diagnostic-buffer** values and must not be re-derived by decoding a production `.out` at those offsets.

## Suggested fix

Make the layout self-identifying, or make the mismatch loud:

- Have `format_verdict_debug` refuse to decode an artefact that does not carry the diagnostic shape (a magic word, a length check, or a build marker), rather than decoding whatever it is handed.
- Document at the offset table which **build** it describes, and name the production layout beside it — `output[0:32]` = `new_payload_request_root`, `output[32]` = `successful_validation`.

## Related

Fourth measurement-instrument defect found the same evening:

- #11736 — leaf decoder reports `validMask`-clear fields as mismatches (over-reports), and surfaces only a single *primary* leaf so a second diverging leaf stays hidden (under-reports).
- #11737 — `codegen-eest-stateless-check.sh` exits 0 with failures.
- A comment body written to a bare `/tmp` path published another identity's content while `gh` reported success.
- This one.

None of the four affected the guest. All four affected what we believed about the guest, and each produced confident, well-formed output that was wrong in a way the output itself could not reveal.

pirapira:
Closes #11737. Closes #11738.

Two instrument defects found while investigating #11306. Both produced confident, well-formed output that was wrong in a way the output itself could not reveal.

## #11737 — the harness exited 0 with failures

Measured on `2ed2c844d`: **selected 648, ran 648, fail 116, exit 0.** So `harness && echo ok` printed `ok` on a run where 116 rows failed, and an exit-code-only check reported all-green.

It now exits non-zero when `fail > 0` **or** `errored > 0`, printing both counts next to `selected` and `ran`. `budget` is deliberately excluded — the summary already labels it *"NOT a correctness failure"*.

⚠️ **The exit code was misleading in BOTH directions**, which is why the fix reports `selected` too: a filter on a run-scoped label index (`--filter 01087_test_...`) makes the harness print `no stateless blocks selected` and exit **1** having measured nothing. So `fail: 0` is **vacuously true when `selected: 0`** — the count alone is not a verdict.

### Callers audited before changing the default

| caller | invokes? | uses the status? | action |
|---|---|---|---|
| `Dockerfile:114` `ENTRYPOINT` (`CMD --all --jobs 2 ...`) | **yes** | becomes the container exit status | none — a conformance container *should* fail when fixtures fail |
| `scripts/check-guest-elf-override.sh` | **yes** | **requires non-zero** for misuse cases (`if [[ "$status" -eq 0 ]]; then fail "... was ACCEPTED (exit 0)"`) | none — this change only *adds* non-zero exits, so the guard is reinforced |
| `scripts/codegen-eest-preflight-report-smoke.sh` | **yes** | last command under `set -euo pipefail` | ⭐ **given `--exit-zero-on-failures`** |
| `check-statement-tamper.sh:110`, `progress-delta.sh:270` | no — path allowlist patterns | — | none |
| `eest-ab-compare.py`, `eest-succ-mismatch-report.py`, `eest-gas-parity-report.py`, `eest-specref-check.sh` | no — consume run dirs / mention it in prose | — | none |
| `codegen-zisk-stateless-verdict-debug-smoke.sh:65` | no — reads the harness's *text* to extract the formatter | — | none |
| `.github/workflows/**` | **no workflow invokes it at all** | — | none |

The preflight smoke is the one caller that genuinely needs the summary regardless of outcome: its own header says the decoded dimensions must be visible *"even when the guest halts successfully"*, and equally when it does not — under `set -e` a failing row would abort it and hide the very report it exists to smoke. Its opt-out carries a comment saying not to copy it to conformance callers.

## #11738 — the offset table described a different build, silently

`format_verdict_debug` reads the first 168 bytes as 21 `u64` words (`verdict@+0`, `bv_fail@+8`, … `tx_state0@+104`). Those offsets describe the **diagnostic** build (`zisk_stateless_verdict_v2`, via `verdict_debug_for_case` → `$label.verdict-debug.output`) — the only artefact it is ever called with. Nothing said so.

The **production** output is a different shape: bytes `0:32` are the `new_payload_request_root` and byte 32 is `successful_validation`. Decoding one at these offsets yields plausible garbage with **no error** — a keccak digest read as `verdict` gave `1841024047515375962` during #11306, and the two files are byte-identical apart from which ELF produced them, so nothing in the artefact distinguishes them.

Added a **shape gate**: `word[0]` is the verdict *bit* in the diagnostic layout, so a value above 1 means the artefact is a production output. The decode is refused with both layouts named and the issue cited, instead of returning numbers that look like measurements. The offset table now documents which build it describes, with the production layout beside it.

## Verification — controls in both directions, on real artefacts

| check | result |
|---|---|
| failing selection (`--skip 440 --limit 2`) | **exit 1**, `fail=2 of selected=2` |
| passing selection (`bal_create2_selfdestruct_then_recreate_same_block`) | **exit 0**, `fail=0`, `full match=2` |
| opt-out on the failing selection | **exit 0** + `exiting 0 because --exit-zero-on-failures was given` |
| real `.verdict-debug.output` | `word0=0`, `word1=1` ⇒ gate silent, `dbg=[verdict=0 bv_fail=1 …]` still decodes |
| production `.output` | `word0=1841024047515375962` ⇒ gate fires |
| `bash -n` on both scripts | clean |

Zero false `UNDECODABLE` on the end-to-end failing run, so the gate does not cost the diagnostics it protects.

Guest-neutral: no Lean or guest source touched.

### DIFF
--- a/scripts/codegen-eest-preflight-report-smoke.sh
+++ b/scripts/codegen-eest-preflight-report-smoke.sh
@@ -9,12 +9,19 @@ cd "$(dirname "$0")/.."
 JOBS="${EEST_PREFLIGHT_SMOKE_JOBS:-1}"
 STEPS="${EEST_PREFLIGHT_SMOKE_STEPS:-${EEST_STEPS:-200000000}}"
 
+# GH #11737: the harness now exits non-zero when a row FAILs. This smoke checks
+# that the PREFLIGHT REPORT renders, not that the selected row conforms -- its
+# header says the decoded dimensions must be visible "even when the guest halts
+# successfully", and equally when it does not. Under `set -e` a failing row would
+# otherwise abort this script and hide the very report it exists to smoke, so the
+# opt-out is deliberate here and must not be copied to conformance callers.
 scripts/codegen-eest-stateless-check.sh \
   --limit 1 \
   --jobs "$JOBS" \
   --quiet-passes \
   --max-failures 1 \
   --preflight-report always \
+  --exit-zero-on-failures \
   --steps "$STEPS" \
   "$@"
 
--- a/scripts/codegen-eest-stateless-check.sh
+++ b/scripts/codegen-eest-stateless-check.sh
@@ -203,6 +203,13 @@ BSR_BAL_CAP="${EEST_BSR_BAL_CAP:-}"
 MIN_SUCC=""
 MIN_FULL=""
 MIN_ROOT=""
+# GH #11737: a fixture failure must make the RUN fail.  Until this existed the
+# script exited 0 with any number of failing rows unless an opt-in --min-*
+# threshold happened to be passed, so `harness && echo ok` printed ok on a run
+# with 116 of 648 rows failing.  That is the dangerous direction: a tool that
+# errors loudly gets fixed, one that reports success gets trusted.  Callers that
+# genuinely want the summary regardless of the outcome must say so explicitly.
+EXIT_ZERO_ON_FAILURES="${EEST_EXIT_ZERO_ON_FAILURES:-0}"
 DEFAULT_TAG="$(tr -d '[:space:]' < scripts/eest-fixture-tag.txt 2>/dev/null || true)"
 DEFAULT_TAG="${DEFAULT_TAG:-$(cat scripts/eest-fixture-tag.txt)}"
 TAG="${EEST_FIXTURE_TAG:-$DEFAULT_TAG}"
@@ -248,6 +255,9 @@ Usage:
 
 Options:
   --all                    run every stateless block (slow); default: smoke subset
+  --exit-zero-on-failures  exit 0 even when rows FAIL or ERROR (GH #11737). Default is
+                           to exit non-zero, so a failing run cannot read as green.
+                           Use only when the summary is wanted regardless of outcome.
   --skip N                 skip first N selected stateless blocks after filtering
   --limit N                cap to N guest invocations (default 50)
   --filter SUBSTR          only fixtures whose relpath contains SUBSTR
@@ -307,6 +317,7 @@ require_arg() {
 while [[ $# -gt 0 ]]; do
   case "$1" in
     -h|--help) usage; exit 0 ;;
+    --exit-zero-on-failures) EXIT_ZERO_ON_FAILURES=1; shift ;;
     --all) ALL=1; shift ;;
     --backend) require_arg "$1" "${2:-}"; BACKEND="$2"; shift 2 ;;
     --skip) require_arg "$1" "${2:-}"; SKIP="$2"; shift 2 ;;
@@ -862,6 +873,18 @@ run_guest_elf() {
 format_verdict_debug() {
   local out="$1"
   local raw
+  # GH #11738: THESE OFFSETS DESCRIBE THE DIAGNOSTIC BUILD'S OUTPUT, NOT THE
+  # PRODUCTION GUEST'S.  The two layouts are different and neither is self-
+  # describing:
+  #   diagnostic : 21 u64 words from +0 -- verdict@+0, bv_fail@+8, ...,
+  #                tx_state0@+104, tx_state1@+112, then two 32-byte roots at
+  #                +168 (sv_recomputed) and +200 (payload state root).
+  #   production : bytes 0:32 are the new_payload_request_root, byte 32 is
+  #                successful_validation, and the rest is the SSZ tail.
+  # Decoding a PRODUCTION .out at the offsets below yields plausible garbage with
+  # no error -- a keccak digest read as `verdict` produced 1841024047515375962
+  # while investigating #11306.  The guard below refuses that rather than
+  # returning numbers that look like measurements.
   local -a labels=(
     verdict
     bv_fail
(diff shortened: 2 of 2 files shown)


pirapira:
Reviewed and labelling. Closes #11737 and #11738.

## The caller audit is what makes this safe to land, and it found the exception

Changing an exit code is the kind of edit that fixes one thing and silently breaks a workflow nobody was watching. The audit is complete and each row states **whether the caller invokes the harness and what it does with the status** — which is the right pair of questions:

- `Dockerfile:114` `ENTRYPOINT` — status becomes the **container** exit status. Left alone: a conformance container *should* fail when fixtures fail.
- `check-guest-elf-override.sh` — **requires** non-zero for its misuse cases. This change only *adds* non-zero exits, so the guard is reinforced rather than disturbed.
- ⭐ `codegen-eest-preflight-report-smoke.sh` — **the one genuine exception**, and it gets the opt-out rather than the default being left unsafe. Its own header says the decoded dimensions must be visible *"even when the guest halts successfully"* — and equally when it does not. Under `set -euo pipefail` a failing row would abort it and hide the report it exists to produce. Its opt-out carries a comment telling conformance callers not to copy it.
- Everything else either doesn't invoke it (allowlist patterns, run-dir consumers, prose mentions) or reads the harness's *text* rather than its status. **No workflow invokes it at all.**

## ⭐ The second direction is the part I had not seen

The exit code misled in **both** directions. A filter on a run-scoped label index (`--filter 01087_test_…`) makes the harness print `no stateless blocks selected` and exit **1** having measured nothing. So **`fail: 0` is vacuously true when `selected: 0`** — the count alone is not a verdict, which is why reporting `selected` beside `fail` matters as much as the exit change.

That is the same trap that cost a run during the #11362 investigation, where a label-prefix filter selected zero blocks and read as a missing fixture. Both failure modes now surface in one line.

Excluding `budget` from the non-zero condition is correct — the summary already labels it *"NOT a correctness failure"*.

## #11738

`format_verdict_debug`'s offsets describe the **diagnostic** build (`zisk_stateless_verdict_v2`, reached via `verdict_debug_for_case`), the only artefact it is ever called with — and nothing said so. The production output is a different shape (`0:32` request root, byte 32 `successful_validation`), so decoding one at those offsets yields plausible garbage with no error.

Making the mismatch loud rather than silent is the right fix: a decoder that refuses the wrong shape cannot produce a confident wrong number, and the offset table now names which build it describes with the production layout beside it.

Scope is two scripts, no emitted-image change, no pin regeneration owed.

PR #11741, opened 2026-08-08
fix(account-decode): value bounds for nonce/balance (#11523)

pirapira:
## Summary

#11523: replace nonce/balance **length** limits with **value** bounds (leading-zero tolerant) in `account_decode` / AccountDecodeSpec.

### Four-clause table (issue + e5a8caf1b `witness_state.py:112-118`)

| field | was | becomes | direction |
|-------|-----|---------|-----------|
| 0 nonce | `l0 ≤ 8` | `significantLen ≤ 8` (u64 value) | FR-ward |
| 1 balance | `l1 ≤ 32` | `significantLen ≤ 32` (u256 value) | FR-ward |
| 2 storage_root | `{0,32}` | STAY length | loosening = FA |
| 3 code_hash | `{0,32}` | STAY length | same |

### Spec cites / witnesses

1. **Structural defect:** raw `l0>8` reject made ACCEPTANCE depend on whether the account was read — `account_at_address` calls `account_decode` after MPT hit; raw-leaf routes never invoke the guard → same leaf pass/fail by read-status. This fix dissolves that for **nonce/balance only**; residual length clauses on root/code inherit the pattern (named follow-up, not this PR).
2. **Positive witness:** `_decode_account_from_leaf` ACCEPTS 9-byte nonce `00…01` + empties (decoded nonce = 2^56); spec accepts, we rejected → FR.
3. Bound is on **decoded value** (significant BE length), not a larger raw length.

### Implementation

- **Prog** (`accountDecode_prog`): strip leading zeros then `BLTU` on significant length (nonce ≤8, balance ≤32). Length 162→**174** (+12 insn = +0x30). Fail PC +504→**+552**.
- **Spec:** `nonceValueOk` / `balanceValueOk`; `DecodeFailure.field0Len/field1Len` = `¬valueOk`; `nonceAccum` / `balanceCopied` use significant off/len.
- **Proofs:** new `AccountDecodeStrip.lean`; `adNonceLenCheck`/`adBalLenCheck` full strip+value path; Close6 ContEpis retargeted; Bridge `nonceAccum_of_content` / `balanceCopied_of_content` with `hnlz=0` for minimal BE (`toBytesBE`).
- **Axioms:** classical-3 only (`propext`, `Classical.choice`, `Quot.sound`) on `account_decode_spec_within`.

## Extents

| | main (base `1680f87bd`) | cand |
|--|------------------------|------|
| text | 0x54c04 | **0x54c34** (+0x30) |
| data | 0x5370 | 0x5370 |
| bss | 0x19029540 | 0x19029540 |

- main ELF sha `b52b6a16a1c3c182…`
- cand ELF sha `decb1102e2a2392f…`
- pins: `textSizeBytes 0x54c34`; 2nd full four-artefact pass NO-OP (ELF sha stable; GuestImageEntries byte-identical after address-only moves — rows by symbol name)

## r200 (seed=932419178, spike, limit 200)

| | selected | ran | fail | full | FA | FR |
|--|----------|-----|------|------|----|----|
| main | 200 | 200 | 4 | 196 | 0 | 4 |
| cand | 200 | 200 | 4 | 196 | 0 | 4 |

- `eest-ab-compare`: FA 0→0, FR 4→4, **FR set equality PASS**, status diffs 0, output-byte diffs 0
- **VERDICT: BYTE-IDENTICAL on all 198 joined cases; P→F=0; NEW_FA=0**
- ELFs distinct (self-check 0 PASS)

## Test plan

- [x] `lake build` AccountDecode* chain green, classical-3
- [x] four-artefact regen + 2nd pass NO-OP
- [x] r200 A/B P→F=0 with counts above
- [ ] CI

## Complicated-merge resolution (check-asm-to-program)

**Diagnosis:** fixture `scripts/asm-fixtures/accountDecodeFunction.s` was **stale** relative to the intentional value-bounds emission (old ABI names + raw `bgtu` length checks). Not an unintentional render divergence — the 22-file proof retarget matches strip+`BLTU` on significant length.

**Fix:** updated fixture to the Lean `emitProgramR` render of `accountDecode_prog` (174 insn); `asm_to_program.py rewrite` refreshed State.lean (same assembled bytes; branch imms as `brOff`).

**Extents unchanged:** text **0x54c34** (+0x30 vs pre-change 0x54c04); data/bss unchanged. `check-asm-to-program` CLEAN.


pirapira:
Reviewed and labelling. Closes #11523.

## The four-clause table is the review, and two clauses correctly do NOT change

| field | was | becomes | why |
|---|---|---|---|
| nonce | `l0 ≤ 8` | significant length ≤ 8 | FR-ward: spec accepts, we rejected |
| balance | `l1 ≤ 32` | significant length ≤ 32 | same |
| storage_root | `{0,32}` | **stays a length** | loosening would be **FA-ward** |
| code_hash | `{0,32}` | **stays a length** | same |

⭐ Stating the direction per clause is what makes this reviewable. Two clauses are converted because a length limit rejects a *non-canonical encoding of a legal value*; two are left alone because converting them would loosen acceptance in the accept-capable direction. A blanket conversion would have read as consistent and been a soundness regression on half the table.

## The justification is structural, not just an encoding complaint

The raw `l0 > 8` reject made **acceptance depend on whether the account was read**: `account_at_address` calls `account_decode` after an MPT hit, while raw-leaf routes never invoke the guard — so the *same witness leaf* passes on one route and rejects on the other, and the verdict stops being a function of the block alone. This fix dissolves that **for nonce and balance only**, and the body says so: the residual length clauses on root and code_hash inherit the same pattern, named as a follow-up rather than quietly implied to be covered.

**And the positive witness settles direction beyond argument**: `_decode_account_from_leaf` at the pin *accepts* a 9-byte nonce (`00…01`, decoded 2⁵⁶) that we rejected. Spec accepts, we reject — false-reject, exactly as filed.

The bound is on the **decoded value** (significant BE length), not a larger raw length — which is the distinction that keeps an out-of-range value encoded compactly from slipping through.

## Mechanics

- `.text 0x54c04 → 0x54c34` (+0x30, matching the +12 instruction prediction), `.data`/`.bss` unchanged, paired ELF shas named.
- Four-artefact pins with a second full pass **no-op**, and `GuestImageEntries` **byte-identical** — correctly explained as expected for an address-only move, since its rows are emitted by symbol name. That is the clarification from earlier tonight applied without prompting.
- r200 seed `932419178`: both legs `selected 200, ran 200, fail 4, full 196`; FR set equality PASS; **byte-identical on all 198 joined cases; P→F = 0; NEW_FA = 0**; ELFs distinct with self-check passing.
- Explicit `selected/ran/fail/full` on both legs — necessary, since the harness exits 0 with failures (#11737, fixed in #11740).

⚠️ 22 files, most of them the proof retarget across the `AccountDecode*` chain. That breadth is a *consequence* of changing a Spec-level definition, not scope creep: `lake build` gates on those proofs, so a source change leaving them red would be an incomplete change. Axioms are the classical three on `account_decode_spec_within` — worth stating explicitly on a proof-touching PR, since that is precisely where a TCB-expanding tactic could slip in.

CONFLICTING against current main; the actor owns the resolution.

pirapira:
Parking this one — it fails `check-asm-to-program` **standalone on its own branch**. I checked pristine `main` first, where the same gate is CLEAN, so this is not a batching artifact:

```
DRIFT DETECTED:
  [EvmAsm/Codegen/Programs/State.lean] accountDecodeFunction:
```

The value-bounds change to nonce/balance alters what `accountDecodeFunction` emits, and the hand-written source in `State.lean` no longer matches the render.

I am not regenerating this one myself: the gate says explicitly it **would change the guest/probe images**, so it is emitted-code substance rather than a layout regen, and re-rendering would silently move bytes the image is pinned against. That is your call. (For contrast, when this same routine drifted in #11602 the fixture was merely stale and the documented `.s → Lean` rewrite was safe; here the gate is flagging the stronger condition.)

The siblings #11740 and #11743 are unaffected and I am landing them separately.

Once resolved, drop the `complicated-merge` label and it will be picked up on the next pass.

dhsorens:
Raising this **before** merge rather than after, because this PR is the point at which it becomes reachable. Not a blocking objection — a question I could not close by reading, with the check I did do.

## The concern: read-side widening without a write-side guard

This PR makes `Decoded`'s nonce/balance clauses leading-zero-tolerant **value** bounds, so the #11522 witness leaf (9-byte nonce `00 01` + zeros, value 2⁵⁶) is now **accepted** on read. That is right, and it dissolves the read-status control dependency.

The part I cannot close by reading is what happens when such an account is subsequently **written**.

**Spec side re-encodes canonically.** `SpecRef/IncrementalMptWrite.lean:108`:

```lean
def encode_account (a : Account) (storage_root : Bytes) : Bytes :=
  encR (.list [.bytes (rlpScalar a.nonce), .bytes (rlpScalar a.balance), …])

private def rlpScalar (n : Nat) : Bytes := EvmAsm.EL.RLP.Nat.toBytesBE n
```

and `Nat.toBytesBE`'s docstring (`EL/RLP/Basic.lean:44`) is explicit: *"Encode a natural number as **minimal** big-endian bytes."* Reached on the write path via `encode_mpt_value` (`:134`).

**Guest side splices.** I checked rather than assumed — every account writer calls `mpt_splice_slot`, which is contractually "the source list with only item *k* replaced":

| writer | site |
|---|---|
| `account_add_balance` | `AccountBalance.lean:146` |
| `account_set_uint_field` | `AccountBalance.lean:241` |
| `account_set_storage_root` | `StorageWrite.lean:151` |
| BAL post-fields | `BalAccountApplyPostFields.lean:317` |
| `mpt_delete_acc` | `MptDeleteAcc.lean:522` |

So the three untouched fields are carried **verbatim**, including a 9-byte nonce.

⇒ For an account that this block **modifies**:

| | spec | guest |
|---|---|---|
| nonce field emitted | `rlpScalar` → **minimal** | spliced leaf → **9 bytes preserved** |

Different leaf bytes ⇒ different node hash ⇒ different post-state root. Direction is a **root mismatch**, not a false reject — the opposite direction from the bug this PR fixes.

## ⚠️ What I have and have not established

- ✅ The spec re-encodes minimally on write (read the source).
- ✅ Every guest writer splices rather than re-encodes (listed above).
- ❌ **I have not built the fixture** — a witness leaf with a 9-byte nonce on an account a transaction modifies, comparing guest root against spec root. So this is a source-level argument, not a demonstrated divergence.
- ❌ I have not checked whether such a leaf is **constructible in a real witness** — a block builder emitting a non-canonical nonce is unusual, so reachability may be limited to adversarial or hand-built witnesses. That bears on severity, not on whether the mechanism exists.

Note the asymmetry with #11522's point 1, which I accept: *untouched* leaves are never re-encoded by either side, so they agree. This is only about the **touched** case.

## What I would suggest

If the fixture confirms it, the guard is small — canonicalise the affected scalar fields at splice time, or refuse to splice a leaf that decoded non-canonically. Much less than the decode→encode reroute I originally proposed on #11522, which point 1 correctly retired.

Happy to build the fixture now and report back before this merges — say the word. If you have already reasoned this through and it does not arise (e.g. the writers are unreachable for accounts whose leaf decoded non-canonically), I would rather hear that than hold the PR up.

Issue #11323, opened 2026-08-03
codegen: pointer-context precompile call UNDERCHARGES gas (352 and 105,020 on 24838/24981) — accept-capable direction, was filed as a 100-gas overcharge


pirapira:
## Summary

The EIP-7702 `test_call_to_precompile_in_pointer_context` family appears to carry one additional 100-gas overcharge beyond the AUTH_BASE defect tracked by #11310.

This is a measurement on the pre-#11316 parent guest, so the residual must be re-measured after the #11310 fix lands. No code change is proposed here.

## Exact measurements

Using the pointer row 24981 and call-via-pointer row 24838 from the fc=1 census:

| family | guest authority balance | execution-specs balance | wei delta | gas delta at price 7 |
| --- | --- | --- | ---: | ---: |
| pointer to precompile | 999999999999999999998721849 | 999999999999999999998968179 | 246330 | 35190 |
| call to precompile in pointer context | 999999999999999999998696488 | 999999999999999999998943518 | 247030 | 35300 |

Both fixtures have `max_fee_per_gas=7`, priority fee `0`, base fee `7`, hence effective gas price `7`; both have gas used `97920` in the pinned execution-specs replay.

The first delta is exactly `35190` gas, the AUTH_BASE component owned by #11310. The second is exactly `35300` gas, i.e. AUTH_BASE plus `100` gas. The delta between families is exactly `700` wei = `100` gas.

## Likely mechanism

The pinned execution-specs commit is `e5a8caf1b8055e4d805c7fb169edfa710914b7da`. It defines `GasCosts.WARM_ACCESS = Uint(100)` in `execution-specs/src/ethereum/forks/amsterdam/vm/gas.py:69`. Delegated-target setup charges that cost when the delegated address is already warm in `execution-specs/src/ethereum/forks/amsterdam/vm/interpreter.py:297-299`.

The likely residual is that the call-via-pointer route charges one warm account access that the spec does not charge (or charges it despite the account already being accessed). This should be confirmed on a post-#11316 guest before fixing.

Related work:

- #11310 owns the AUTH_BASE overcharge shared by both families.
- #11317 tracks the broader scratch-lifetime census and its `cd_*` sibling candidates.

[...]

PR #11743, opened 2026-08-08
fix(scripts): gate leaf-decoder fields on validMask and surface every diverging leaf (#11736)

### LINKED ISSUE #11736: test tooling: the leaf decoder reports unset (validMask-clear) fields as mismatches, producing false residuals

pirapira:
Closes #11736.

The decoder misled in **both** directions: it reported fields that were *unset* as mismatches, and it *hid* diverging leaves behind a single primary. Both halves cost real investigation time on #11306 tonight.

## ⚠️ FORMAT CHANGE — read this first

Two open investigations (#11306, #11323) are using this tool right now.

- **Column layout is stable.** The four new columns are **appended**, so every existing column keeps its index and `cut -f N` against an older report still works.
- **Three class NAMES changed** for the previously-misreported cases (below). **Do not compare class strings across this boundary** — re-run rather than diffing an old report against a new one.

## Over-reporting: the mask gate was present but defeated

The balance and nonce comparisons read:

```python
if r["bal"] != exp_bal and (r["vmask"] & 0x1 or r["bal"] or exp_bal):
```

The trailing `or` clauses defeated the gate **exactly when it mattered** — an unset balance bit with a nonzero expectation still reported a mismatch (row `6295ee1b…`, `vmask 0x3a`). Each comparison now gates on the validMask bit **alone**, and an unset field is reported as **UNSPECIFIED** rather than as a value.

⭐ *A flag says whether a field is meaningful; the field says what it is.*

Two related false residuals, both named in the issue:

| case | was | now |
|---|---|---|
| TOUCHED-only row (no value bits) | `guest_extra_not_in_post` — a "guest-extra presence" finding | `enumeration_entry_touched_only` — an execution-map/root-enumeration entry whose absence from `postState` is **expected** |
| storage-only post delta | `post_delta_missing_from_aw` — "absent from the account-writes arena" | `storage_in_separate_structure` — names where the counterpart lives (`storage_writes` is a separate structure; `execution_map_state_changes` enumerates the union) |

All informational classes are excluded from the primary work-list, so they cannot read as defects.

## Under-reporting: one primary hid the leaf that mattered

A single `primary_*` triple reads as a complete answer. On #11306 it surfaced only the sender and hid the **coinbase** leaf — and the hidden leaf was the corroborating datum:

| leaf | delta | price | implied gas |
|---|---|---|---|
| sender | +1,079,200 | gasPrice 10 | 107,920 |
| coinbase | −323,760 | priority (10−7) = 3 | 107,920 |

with `sender + coinbase = 755,440 = baseFee 7 × 107,920` (the burn). Three quantities agreeing on one gas number is far stronger than the single balance delta the decoder surfaced, so the under-report cost the strongest form of the finding — recovered only by re-deriving the leaf set by hand.

⇒ The cost asymmetry mirrors the over-report but is **less visible**: an over-report costs an investigation into a defect that isn't there; an under-report costs the corroboration that would have settled a real one, and one confident primary looks complete.

Added `n_diverging_leaves`, `diverging_accounts`, `diverging_by_field`, `n_informational`, and a `⚠ MULTI-LEAF` summary line. `n_field_recs` did **not** answer this — it mixes field classes (it counted 6 balance + 6 nonce + 6 storage across 3 rows).

## Verification — on the real #11306 rows, before/after

| record class | before | after |
|---|---|---|
| `balance_mismatch` | 6 | **6** — the real findings survive (two per row: sender + coinbase, both with the balance bit set) |
| `nonce_mismatch` | 6 | **0** — revealed as validMask-clear artefacts |
| `post_delta_missing_from_aw` (storage) | 6 | **0** — reclassified as by-design |
| multi-leaf visibility | none | `⚠ MULTI-LEAF: 3 of 3 row(s) diverge on MORE THAN ONE account (2 leaves×3)` |
| informational excluded | — | 20 records |

`diverging_accounts` now names both `2adc2566…` (coinbase) and `f6c3a9ed…` (sender) on every row. First 16 columns verified identical in order. `ast.parse` clean.

⭐ Note this **retro-validates** the #11306 analysis: the balance leaves it rests on are real, and the 6 nonce "mismatches" — which that analysis never relied on — were artefacts.

Guest-neutral: no Lean or guest source touched.


pirapira:
Reviewed and labelling. Closes #11736.

## The format-change section leading the body is the right call

Two open investigations (#11306, #11323) are using this tool right now, so the compatibility statement belongs first, not in a footnote: **column layout stable, four new columns appended so existing indices and `cut -f N` still work; three class NAMES changed, so class strings must not be diffed across the boundary.** "Re-run rather than compare" is the correct instruction — a silently-renamed class is exactly how a stale report gets read as agreement.

## Over-reporting: the gate was present but defeated

```python
if r["bal"] != exp_bal and (r["vmask"] & 0x1 or r["bal"] or exp_bal):
```

⭐ The trailing `or` clauses defeated the mask gate **precisely when it mattered** — an unset balance bit with a nonzero expectation still reported a mismatch. That is worse than a missing gate: a reader inspecting this line sees `vmask & 0x1` and concludes the field is guarded. Gating on the validMask bit alone, and reporting an unset field as **UNSPECIFIED** rather than as a value, is the fix. *A flag says whether a field is meaningful; the field says what it is.*

Renaming the two related classes is more than cosmetic: `guest_extra_not_in_post` → `enumeration_entry_touched_only` and `post_delta_missing_from_aw` → `storage_in_separate_structure` replace findings-shaped names with names that state where the counterpart actually lives. Excluding informational classes from the primary work-list is what stops them reading as defects.

## Under-reporting is the half that cost more, and the PR says why

A single `primary_*` triple reads as a complete answer. On #11306 it surfaced the sender and hid the **coinbase** leaf — and the hidden leaf was the corroborating datum: sender +1,079,200 at gasPrice 10 and coinbase −323,760 at the priority rate both imply **107,920 gas**, with `sender + coinbase = 755,440 = baseFee 7 × 107,920`.

⭐ Three quantities agreeing on one number is a different class of evidence from one balance delta. So the under-report did not merely omit a row — it cost the **strongest form of the finding**, recovered only because the leaf set was re-derived by hand.

The cost asymmetry is stated exactly right: an over-report costs an investigation into a defect that isn't there; an under-report costs the corroboration that would have settled a real one — **and one confident primary looks complete**, which is why it is the less visible failure.

`n_diverging_leaves`, `diverging_accounts`, `diverging_by_field` and `n_informational` make both properties visible in the output rather than recoverable by hand.

Scope is three scripts, no emitted-image change, no pin regeneration owed. Labels on this PR were auto-applied by the labeler; nothing self-labelled.

Issue #11301, opened 2026-08-03
test tooling: full corpus contains an overwritten default-failure input


pirapira:
## Finding

A mechanical scan of the 26,104 root inputs listed by `/var/tmp/fc668/manifest.tsv` found one corpus slot containing a guest output artifact rather than a framed EEST input. The manifest SHA is `b40b7e3f7d4b3bde03c7298c98d3b9117ea146f76c2a2a914f1cbafa78c3ad07`.

Tell counts (computed independently over all 26,104 inputs):

- first 8 bytes all zero: **2**
- file size exactly 256 bytes: **1**
- first 61 bytes equal the parent ELF `default_failed_stateless_output`: **1**
- all three tells: **1**

The affected selectors are:

- `01618_test_transaction_validity_type_1_type_2_fork_Amsterdam-blockchain_test_from_state_test-type_2-no`: all three tells. Input SHA256 `22e8a5689534438dc9c12388fd939d32cf646e5f8aeb46bbfe6d34a75b614a01`; size 256; mtime `2026-08-01 07:48:40 +0200`. Parent ELF SHA256 is `57286fcd03e2f91b1131191a0e9e9ff417fade03551d713cf55f4c677d1c1d10`. Its first 61 bytes are the default-failure output prefix (prefix SHA256 `ac548100ebfb826a1080505da87be738d4c75488f0100b0966088f49f3590148`). A correctly framed copy of this selector (`b68ba97d0f98b4216b4a9c964f3ac7d894732a4560413717f17d2455e3c42249`, 5272 bytes) runs byte-perfect, so the 256-byte slot is not the EEST input.
- `02240_test_invalid_stateless_input_bytes_are_rejected_fork_Amsterdam-blockchain_test-empty_input_bytes`: first-8-zero only; size 8; SHA256 `af5570f5a1810b7af78caf4bc70a660f0df51e42baf91d4de5b2328de0e83dfc`. This is the intentional empty invalid-input fixture, not the overwritten artifact.

The manifest and normal root inputs were generated around `2026-07-30 00:25 +0200`; `01618` is the only root `.input` observed with a later mtime. Available artifacts do not identify which process wrote it, so this records a later corpus-slot mutation/stale artifact without guessing whether the generator or a sweep performed the overwrite.

## Impact

[...]

PR #11745, opened 2026-08-08
fix(scripts): name all 8 manifest columns and label case_id at every read (#11308)

### LINKED ISSUE #11308: test tooling: manifest column 8 is case identity, not input SHA
## Finding

The 26,104-row corpus manifest `/var/tmp/fc668/manifest.tsv` does not store the SHA256 of either the on-disk ziskemu input or its unwrapped payload in column 8. A read-only audit compared both hashes for every row against column 8.

Provenance:

- manifest SHA256: `b40b7e3f7d4b3bde03c7298c98d3b9117ea146f76c2a2a914f1cbafa78c3ad07`
- manifest mtime: `2026-07-30 00:25:01.759650 +0200`
- rows: `26,104`

Counts (all files present):

- matches on-disk wrapped-file SHA: **0**
- matches unwrapped payload SHA: **0**
- matches neither: **26,104**
- file missing: **0**

This is not evidence that all inputs are corrupt. The converter defines column 8 as `case_id`, a SHA256 over fixture-relative path, full test name, block index, and the original stateless input bytes (`scripts/eest-stateless-to-input.py:194-202`), and writes that value as the eighth manifest field (`:367`). Recomputing that identity from the fixture JSON matched all 26,104 rows.

The audit nevertheless exposed why input provenance is currently uncheckable from the manifest. For `13625...GASPRICE-debug__b19`, the manifest value is `cd9b8851...`, while the on-disk 10,536-byte file hashes to `19056e3d...` and its declared payload hashes to `99ee9129...`; neither comparison can detect whether the file corresponds to the manifest's original input. For `01618...type_2-no`, the file is the known 256-byte overwritten default-failure artifact from [#11301](https://github.com/Verified-zkEVM/evm-asm/issues/11301), and its manifest case identity still matches the original fixture identity because the file was changed after generation.

The manifest should carry an explicit wrapped-file SHA and/or unwrapped-payload SHA (or a separate integrity sidecar) if corpus sweeps are expected to prove that the selected files are the generated inputs. Until then, a case-id match does not establish file integrity.

No guest run or source change was made for this audit.

pirapira:
Closes #11308.

## The defect

The EEST manifest has **8** columns (`label input_file expected_hex succ_bit input_len block_gas_limit fixture_relpath case_id`). Four `read` statements across the two shell harnesses named only **seven**:

```
IFS=$'\t' read -r label input expected_hex succ_bit input_len gas_limit relpath <<< "$line"
```

bash gives the **last** name every remaining field, so `relpath` silently carried `"<path>\t<case_id>"`. Every PASS/FAIL identity line therefore printed a bare 64-hex string immediately after the fixture path, where it reads as a content hash of the input file.

It is not one. `case_id` is a SHA-256 over *(fixture relpath, EEST test name, block index, input bytes)* — a **case identity**, not a file digest. Two consequences that matter:

- an overwritten `.input` file still matches its `case_id` (#11301), so it cannot answer "is this the file I measured?";
- it does not equal `sha256sum` of the file on disk, because the `.input` file carries a host-transport length prefix and zero padding that `eest-stateless-to-input.py` documents as *not* part of guest input content.

Reading that hex as a file hash is what nearly produced a false fixture-identity mismatch on #11362.

## The fix

Per the issue framing this is documentation plus a name, not new data — with one structural addition.

- **All eight columns named**, plus a trailing `_rest` sink. The sink is the structural half: without it a future column 9 would corrupt `case_id` exactly as column 8 corrupted `relpath`. It is never read.
- **`case_identity()` takes and labels `case_id`**, so the hex is never bare. On a legacy 7-column manifest `case_id` is empty and the field is omitted, so old manifests keep working.
- **A comment at each read site** stating the column count and what `case_id` is and is not.

## Caller audit — checked before changing anything

15 scripts read the manifest (plus the writer). Findings:

| consumer | column handling | 9th column? |
|---|---|---|
| `codegen-eest-stateless-check.sh` ×2, `eest-specref-check.sh` ×2 | shell `read`, **7 names for 8 columns** | ⚠️ was the defect — fixed here |
| `eest-bal-replay-report.py`, `eest-precompile-frontier-report.py`, `eest-succ-mismatch-report.py`, `eest-gas-parity-report.py`, `eest-receipt-log-frontier-report.py` | positional `fields[:7]` behind a `>= 7` guard | tolerant |
| `eest-ab-compare.py`, `sweep_one.py`, `code1-leaf-decoder.py`, `spike/standing_controls_sweep.py`, `eest-run-monitor.sh` (awk `mc[1]`/`mc[3]`), `spike/parity-check.sh` (`r[0..2]`) | low positional indices behind a length guard | tolerant |
| `spike/bal_producer_diff.py` (+ `spike/bal_producer_set.sh` wrapper) | `len(fields) not in (7, 8)` and `fields[-2]`/`fields[-1]` | ⛔ would break |

**No consumer joins on column 8 as content identity.** Exactly one consumer extracts it at all — `bal_producer_diff.py` — and it stores the value in a dict and never uses it. So the latent consumer defect the issue predicted is real, but on the reader side rather than the join side: four shell readers folded `case_id` into `relpath`. Severity is display-only (it feeds only `case_identity`), which is why it survived; the cost was a misleading report line, and that line is exactly what misled a reader.

## What a genuine input-SHA column would cost — stated, not added

Cheap to produce, expensive to consume, and it makes the confusion in this issue worse rather than better:

1. **Writer: ~3 lines.** The input bytes are already in hand where `case_id` is computed, so no extra I/O.
2. ⚠️ **Which hash is ambiguous, and either choice is a trap.** `sha256(input_bytes)` does not match what `sha256sum <file>` prints; `sha256(file)` does, but then it shares no component with `case_id`. Whichever is picked, the other is the one a reader computes by hand and gets a mismatch on — a **second look-alike 64-hex column adjacent to the first**, which is the same failure mode this issue is about.
3. ⛔ **One consumer breaks two ways.** `bal_producer_diff.py` hard-fails every row via `len(fields) not in (7, 8)`, and its `fields[-2]`/`fields[-1]` negative indexing would resolve `fixture_relpath` to `case_id` even if the whitelist were widened. 11 of 15 consumers need no change; this one must.
4. **Storage: +65 bytes/row** (64 hex + tab), +10.7% on a measured 605-byte row.

**Recommendation: do not add it.** Anyone who needs "is this the file I measured?" can `sha256sum` the path in column 2 in one command — no schema change, no second look-alike hash, no consumer migration. The genuine defect was that column 8 was unlabelled at every point it was read, which is what this PR fixes.

## Verification

- Read logic unit-tested at **7, 8 and 9** columns: `relpath` clean and `case_id` exact in all three; column 9 lands in `_rest`.
- **Live control**: the old 7-name form reproduces the defect on demand (`relpath=[path/to/f.json<TAB>deadbeef]`), so the check is discriminating rather than an identity.
- `bash -n` clean on both scripts.
- End-to-end, spike backend, ELF `723a67964ba71304c0a11accb0a0e8f486134f89fd81da304066f7e9133b6e1f`, base `28871a17f`: **selected=1, ran=1, fail=0, full match=1**. Identity line renders `...bal_precompile_call_opcode.json (label=00000_test_bal_precompile_call_opcode... manifest_row=1/1 case_id=7efe5de3cddbfe0f... rerun_skip=440 rerun_limit=1)` — clean path, labelled hex. The format is confirmed on both a FAIL line (earlier run) and this PASS line, i.e. both rendering paths.

No behaviour change beyond report text; no guest code touched.


pirapira:
Reviewed and labelling. Closes #11308.

## The defect is sharper than the issue described

The issue said column 8 is case identity rather than input SHA. The actual mechanism is worse: **four `read` statements across two harnesses named seven fields for an eight-column manifest**, and bash gives the *last* name every remaining field. So `relpath` silently carried `"<path>\t<case_id>"`, and every PASS/FAIL identity line printed **a bare 64-hex string immediately after the fixture path** — exactly where a reader expects a content hash of the input file.

⭐ Nobody misread the documentation. **The output was constructed to be misread.** That is the same shape as the other three instruments tonight: confident, well-formed, wrong in a way the output cannot reveal.

And the two consequences are stated precisely — an overwritten `.input` still matches its `case_id` (#11301), so it cannot answer *"is this the file I measured?"*; and it does not equal `sha256sum` of the on-disk file, because the `.input` carries a host-transport length prefix and zero padding that `eest-stateless-to-input.py` documents as not part of guest input content. Reading that hex as a file hash nearly produced a false fixture-identity mismatch on #11362.

## The `_rest` sink is the part that outlasts this fix

Naming all eight columns fixes today's bug; the trailing sink fixes **the next one**. Without it, a future column 9 would corrupt `case_id` exactly as column 8 corrupted `relpath` — silently, and in the same field-shifting way. It is never read, costs nothing, and removes the whole class.

Labelling `case_id` in `case_identity()` so the hex is never bare is the other half: a labelled value cannot be mistaken for a different labelled value, whereas a bare hex string next to a path invites exactly one reading.

## Caller audit

15 consumers checked before changing anything, each classified by **how** it reads columns rather than merely whether it reads them:

- The four shell `read` sites were the defect — fixed here.
- The Python report scripts use `fields[:7]` behind a `>= 7` guard, or low positional indices behind a length guard — tolerant.
- ⛔ `spike/bal_producer_diff.py` (and its `bal_producer_set.sh` wrapper) tests `len(fields) not in (7, 8)` with `fields[-2]`/`fields[-1]` — **would break** on a ninth column. Worth having identified now rather than discovering it when the column is added.

**No consumer joins on column 8 as a content identity**, which was the latent defect I asked to be checked for. Legacy 7-column manifests keep working, since `case_id` is empty and the field is omitted.

Two scripts, no emitted-image change, no pin regeneration owed.

PR #11749, opened 2026-08-08
fix(scripts): refuse to delete a run directory the harness does not own (#11748)

### LINKED ISSUE #11746: latent: two EEST harnesses write <label>.result.tsv with different arities and both honour EEST_RUN_DIR — safe only by unreachability

### LINKED ISSUE #11748: hazard: both EEST harnesses rm -rf a USER-SUPPLIED run directory unconditionally at startup
Both EEST harnesses `rm -rf` their run directory unconditionally at startup — **including when that directory was supplied by the user** via `--run-dir` or `EEST_RUN_DIR`.

- `eest-specref-check.sh:207`
- `codegen-eest-stateless-check.sh:741`

Found while fixing #11746; **deliberately not touched there**, because widening a filename-collision fix into a destructive-delete change would have made both harder to review, and this one needs a decision about what these harnesses are permitted to delete.

## Two consequences

1. **Pointing either harness at a directory containing anything else destroys it.** The flag accepts an arbitrary path and the script deletes its contents before doing any work. Nothing distinguishes "a scratch directory this tool owns" from "a directory the user happens to have named".
2. **A concurrent run deletes the other harness's inputs mid-flight.** This is the mechanism that made #11746 reachable only under concurrency — and the deletion is the more damaging half of it. One harness starting while the other is running removes files the second is actively consuming.

⚠️ The failure mode is silent in the way that matters: the deleting run proceeds normally and reports its own results. The damage lands on the *other* process, or on data that was never this tool's to manage.

## Why it is currently survivable, and why that is not a defence

No caller points both harnesses at one directory, and their defaults differ — so today the destructive case needs a user to opt into it by naming a path. That is **safe by convention, not by construction**, which is the same category as #11746 itself: nothing in the code prevents the bad case, only the current call pattern does.

## Suggested shapes, in rough order of strength

- **Refuse to delete a directory the harness did not create** — e.g. require an ownership marker file, or create a fresh uniquely-named subdirectory under the supplied path rather than clearing the path itself.
- **Delete only what the harness writes** — the `<label>.result.tsv` / `<label>.specref-result.tsv` patterns and its own subdirectories — rather than the whole tree.
- At minimum, **refuse a non-empty directory** unless an explicit `--force` is given.

The first two remove the class; the third only narrows it.

pirapira:
Closes #11748.

Both EEST harnesses ran an **unconditional** `rm -rf "$RUN_DIR"` at startup — including when the directory came from `--run-dir` or `EEST_RUN_DIR`. Safe by convention, not by construction: nothing in the code prevented either consequence, only the current call pattern did.

## Chosen shape: refuse to delete what the harness did not create

Option 1, via an `.eest-run-dir` marker recording harness name and pid. A new `scripts/lib/eest-run-dir.sh` holds the guard and both harnesses source it — duplicating a destructive check in two places invites drift.

| directory state | behaviour |
|---|---|
| does not exist | create, claim it |
| marker names this harness, recorded run exited | **recreate** — the documented behaviour |
| marker names the *other* harness | refuse, delete nothing |
| marker names this harness, recorded pid still live | refuse, delete nothing |
| exists, non-empty, no marker | refuse, delete nothing |
| exists, empty, no marker | adopt (deletes nothing) |

**Why not "delete only what we write" (option 2).** It would be *worse than either extreme here*. Leftover `<label>.result.tsv` files from an earlier, larger selection are globbed by `eest-run-monitor.sh` and joined by the report scripts, so a half-cleaned directory turns a loud destructive bug into a **silent measurement** one.

**Why not a force flag alone (option 3).** It narrows rather than removes, and a flag required for the *documented* stable-directory workflow becomes habitual — passed reflexively, which is how force flags stop protecting anything. The escape hatch is deleting the directory yourself: explicit and visible. Option 3's check survives as the refusal message.

The pid check also covers **same-harness** concurrency, the other half of the second consequence. PID reuse can make it a false refusal; refusing is the safe direction and the message says how to proceed.

## Clean slate is preserved, deliberately

`docs/eest-stateless-testing.md` documents that an `EEST_RUN_DIR` *"is recreated at the start of the invocation"*, and the A/B workflow in `docs/sasm-howto.md` reuses one directory per leg across runs. Both still work — verified below, not assumed. The doc is updated to state when the harness refuses.

## Caller audit

The only programmatic caller passing `--run-dir` is `codegen-eest-gas-parity-report.sh`, and it creates only the **parent** (`mkdir -p "$(dirname "$RUN_DIR")"`), letting the harness create the run dir — so it takes the owned path unchanged. Every other `--run-dir`/`EEST_RUN_DIR` reference is documentation. The marker is a dotfile, so it matches none of the `*.result.tsv` / `manifest.tsv` globs any consumer uses.

## Verification

Base `377baad3d`, commit `759eac5e2`, backend spike.

**Guard unit matrix** — every refusal checked for *survival of the data*, not just for the message:

| case | result |
|---|---|
| fresh directory | allowed, marker written |
| same owner, previous run exited | allowed, **stale file removed — clean slate intact** |
| different owner | refused, data intact |
| non-empty, no marker | refused, data intact |
| empty, no marker | adopted |
| same owner, live pid | refused, data intact |

**End-to-end, real harness:**

- **Foreign directory**: `--run-dir` at a directory holding `notes.txt` and `subdir/deep.txt` → exit 1, refusal printed, **both files survived including the nested one**. On `origin/main` this path is `rm -rf "$RUN_DIR"` with no condition, so the whole tree would be gone.
- **Documented reuse**: two consecutive runs into one `--run-dir`, each **selected=1, ran=1, fail=0, full match=1**, exit 0; a stale file planted between them was removed, so clean slate still holds.
- **Cross-harness**: SpecRef pointed at a guest-owned directory → exit 1, refusal naming both harnesses, **all 11 files intact**.

⭐ The cross-harness refusal also closes #11746's shared-directory scenario at the *directory* level, so that hazard now has two independent barriers: distinct filenames (#11747) and directory ownership (here).

`bash -n` clean on all three scripts. No CI shell-lint gate exists, so the new file needs no registration.

### DIFF
--- a/docs/eest-stateless-testing.md
+++ b/docs/eest-stateless-testing.md
@@ -753,6 +753,15 @@ gen-out/eest-run/run-<timestamp>-<pid>/
 Set `EEST_RUN_DIR=/path/to/dir` to force a stable directory for a single
 reproducible run; that directory is recreated at the start of the invocation.
 
+The harness only recreates a directory it owns. On first use it drops an
+`.eest-run-dir` marker recording the harness name and pid; on a later run it
+recreates the directory if that marker names the same harness and the recorded
+run has exited. Otherwise it **refuses and deletes nothing** (GH #11748) — a
+non-empty directory with no marker, a directory created by the other EEST
+harness, or one whose recorded run is still live. Point `--run-dir` at a new or
+empty directory, or remove the old one yourself. Reusing one directory per A/B
+leg still works exactly as before.
+
 Important files:
 
 - `manifest.tsv`: one row per selected guest invocation.
--- a/scripts/codegen-eest-stateless-check.sh
+++ b/scripts/codegen-eest-stateless-check.sh
@@ -738,8 +738,14 @@ elif [[ -n "${EEST_RUN_DIR:-}" ]]; then
 else
   RUN_DIR="$REPO_ROOT/gen-out/eest-run/run-$(date -u +%Y%m%dT%H%M%SZ)-$$"
 fi
-rm -rf "$RUN_DIR"
-mkdir -p "$RUN_DIR"
+# GH #11748: this used to be an unconditional `rm -rf "$RUN_DIR"`, which would
+# destroy a user-supplied directory and could delete a concurrent run's inputs
+# mid-flight. The guard recreates a directory this harness owns (the documented
+# behaviour) and refuses to delete anything else.
+source "$REPO_ROOT/scripts/lib/eest-run-dir.sh"
+if ! eest_prepare_run_dir "$RUN_DIR" "codegen-eest-stateless-check.sh"; then
+  exit 1
+fi
 GUEST_PREFIX="$RUN_DIR/stateless_guest"
 RESOLVED_GUEST_ELF="$GUEST_PREFIX.elf"
 
--- a/scripts/eest-specref-check.sh
+++ b/scripts/eest-specref-check.sh
@@ -204,8 +204,14 @@ elif [[ -n "${EEST_RUN_DIR:-}" ]]; then
 else
   RUN_DIR="$REPO_ROOT/gen-out/eest-specref-run/run-$(date -u +%Y%m%dT%H%M%SZ)-$$"
 fi
-rm -rf "$RUN_DIR"
-mkdir -p "$RUN_DIR"
+# GH #11748: this used to be an unconditional `rm -rf "$RUN_DIR"`, which would
+# destroy a user-supplied directory and could delete a concurrent run's inputs
+# mid-flight. The guard recreates a directory this harness owns (the documented
+# behaviour) and refuses to delete anything else.
+source "$REPO_ROOT/scripts/lib/eest-run-dir.sh"
+if ! eest_prepare_run_dir "$RUN_DIR" "eest-specref-check.sh"; then
+  exit 1
+fi
 
 # --- convert fixtures -> inputs + manifest (same selection as the guest) -----
 conv_args=(--fixtures-dir "$FX" --out-dir "$RUN_DIR")
--- a/scripts/lib/eest-run-dir.sh
+++ b/scripts/lib/eest-run-dir.sh
@@ -1,0 +1,110 @@
+#!/usr/bin/env bash
+# Shared ownership guard for EEST harness run directories -- GH #11748.
+#
+# Both EEST harnesses used to run an UNCONDITIONAL `rm -rf "$RUN_DIR"` at
+# startup, including when the directory came from --run-dir or EEST_RUN_DIR.
+# Two consequences, both silent in the way that matters -- the deleting run
+# proceeds normally and reports its own results, while the damage lands
+# somewhere else:
+#
+#   * pointing a harness at a directory containing anything else destroys it;
+#   * a concurrent run deletes the other run's inputs mid-flight.
+#
+# The category is the one this repo keeps meeting: SAFE BY CONVENTION, NOT BY
+# CONSTRUCTION.  Nothing in the code prevented either case; only the current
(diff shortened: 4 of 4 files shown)


pirapira:
Reviewed and labelling. Closes #11748.

## Option 1 chosen, and the rejection of option 2 is the sharper argument

I offered three shapes and ranked them. The PR takes the strongest — refuse to delete what the harness did not create, via an `.eest-run-dir` marker recording harness name and pid — and the state table covers the cases that matter, including the two asymmetric ones: marker names the *other* harness → refuse; marker names this harness but the recorded pid is **still live** → refuse. That second row is the concurrency case from #11746, now blocked at the destructive step rather than at the read.

⭐ **But the reasoning against option 2 is what I want on the record, because it inverts my ranking.** I put "delete only what we write" second. The PR shows it would be **worse than either extreme here**: leftover `<label>.result.tsv` files from an earlier, larger selection are globbed by `eest-run-monitor.sh` and joined by the report scripts, so a half-cleaned directory turns a **loud destructive bug into a silent measurement one**.

That is the trade this whole session has been about. A destructive delete announces itself; a stale file that joins cleanly does not. Given four instruments tonight failed by producing confident wrong output, converting a visible failure into an invisible one would have been the wrong direction even though it deletes strictly less.

Rejecting the force flag as narrowing rather than removing the class is consistent with the same standard.

## Structure

Putting the guard in `scripts/lib/eest-run-dir.sh` and sourcing it from both harnesses is right — **duplicating a destructive check in two places invites drift**, and that is precisely the family of defect this PR closes. A shared implementation means the next person changing deletion semantics changes it once.

Documentation updated alongside, so the "recreate" behaviour stays documented rather than becoming folklore.

Four files, no emitted-image change, no pin regeneration owed.

Issue #11744, opened 2026-08-08
codegen: 24838 pointer-context call undercharges 352 gas on the regular side (split from #11323; state gas is charged correctly here)


pirapira:
Splitting out the 24838 half of #11323, which measurement showed is a **separate defect** from 24981.

## Why they are not one defect

Both rows sit in the `test_call_to_precompile_in_pointer_context` family and both fail `bv_fail=1`, which is what made a shared cause the natural assumption. Instrumentation refuted it:

| row | shortfall | user state gas | mechanism |
|---|---|---|---|
| 24981 | **105,020 gas** | **not charged** (97,920 missing), `h_SSTORE` hit only 3× | pointer-precompile routing — tracked in #11323 |
| **24838** | **352 gas** | **charged correctly** — the `0x6270e00 → 0x6258f80` transition (97,920) occurs at the user SSTORE path | regular-side, unidentified |

⭐ Sharing a fixture family and a verdict code is not evidence of a shared mechanism. `bv_fail=1` is the terminal state-root comparison — where **every** state divergence lands regardless of origin — so uniformity there is the weakest possible signal. Three orders of magnitude between the two shortfalls, and one charging state gas correctly while the other does not, is the discriminating evidence.

## This issue: 24838 only

**Guest 150,574 gas vs spec 150,926 — short by 352**, i.e. **+2,464 wei** retained by the sender at effective price 7. Measured on `2ed2c844d`, guest ELF `723a6796…`, matched pair, `repo_dirty=0`.

⚠️ **Direction: undercharge, which is the accept-capable direction.** It surfaces as a false-reject here only because these fixtures carry honest headers — a wrong computation mismatches an honest header. That is not evidence of safety; the attacker supplies the header.

## Method notes carried over

⛔ **Do not attribute 352 by factorising it.** On #11306 tonight, `107920 = 10000 + 97920` matched two different spec mechanisms on **opposite sides of the ledger** (the gated charges and the refunds), and instrumentation refuted both. A magnitude that decomposes does not name a location. Name the charge from a probe or from the code path.

Also note 352 is not a multiple of 1530, so it is **not** a per-byte state charge — consistent with the state-gas path being correct on this row. Reconcile it against the regular gas dimension.

The useful next step is the one that worked on #11306: bisect where the shortfall first appears rather than guessing a site. A single step names the subsystem; accumulation means it is not one missing charge.

(Issue generated by Claude Code.)

pirapira:
[...]

PR #11751, opened 2026-08-08
fix(eip7702): preserve map nonce in authority_asof live path (#11744)

pirapira:
## Summary

Closes the **352 regular-gas** shortfall on `24838_test_call_to_precompile_in_pointer_context` (authT delT 0x01) by fixing `eip7702_authority_asof` so self-sponsored AUTH sees the post-inclusion map nonce.

### Control dependency (why gas moved)

1. Sender inclusion writes a **nonce-only** CURRENT row (mask `0x22`) before AUTH prepare.
2. `account_writes_auth_current` requires HasNonce|HasState|HasExecFlags → **miss**.
3. Fall-through `.L77as_normal_nonce` → `account_writes_latest_nonce_tx` puts map nonce **2** in `s1`, `s2=1`.
4. **Bug:** `.L77as_live` always `ld a1, teer_pre_acct` (header nonce **1**), wiping `s1`.
5. Signed auth.nonce=**2** ≠ asof **1** → AUTH reject → no set_code.
6. CALL on pointer still sees pre-state STOP delegate `0x37f5…` → DELEG cold 2900 + wrong code → missing second precompile → **352 regular** shortfall.

Fee shortfall is **entirely regular** (guest 52654 vs spec 53006); `state_used` 97920 identical authT/authF. AUTH_BASE 35190 is state-budget only on authF — magnitude coincidence, not the fee defect.

### Fix (mirror existing guard)

At `.L77as_nonce_ready` (L158) the code already has `bnez s2, …` before loading the header. `.L77as_live` lacked the same guard:

```asm
; before
.L77as_live:
  la t0, teer_pre_acct; ld a1, 0(t0); li a0, 1; j .L77as_ret

; after
.L77as_live:
  bnez s2, .L77as_live_map; la t0, teer_pre_acct; ld s1, 0(t0)
.L77as_live_map:
  mv a1, s1; li a0, 1; j .L77as_ret
```

Preserves the #11723 EF0100 block-map delegated path (when `s2=0`, still loads header).

### Pin measurement (input sha `12d3935b6cd7bcf696602415939298957e94c6343ca9db1c6e49be69a7168e00`)

| | BASE | FIX |
|--|--|--|
| ELF sha256 | `519d9d19a3b8e68a…` | `526d9e66f7ac88f1…` (post-pin) |
| total fee gas | **150574** | **150926** (= spec) |
| shortfall | **352** | **0** |
| `teer_success_count` | 0 | **1** |
| `exec_code_effect_count` | 0 | **1** |
| asof after recover | a0=1 a1=**1** a2=1 | a0=1 a1=**2** a2=1 |

### Extents

| | main (base) | cand |
|--|--|--|
| text | 0x54c9c | **0x54ca4** (+8) |
| data | 0x5370 | 0x5370 |
| bss | 0x19029560 | 0x19029560 |

Four-artefact regen (TSV / GuestAddrs / GuestImageEntries / RegionMapLinkPins); second pin pass **NO-OP**. GuestImageEntries byte-identical (address-only moves by symbol name).

### r200

In flight: seed `932419178`, spike, limit 200, filter amsterdam — will update PR with `eest-ab-compare` (bar: base-PASS→cand-FR **ZERO**).

### Test plan

- [x] Spike pin: 352→0; asof a1=2; teer_success≥1
- [x] Four-artefact fixpoint NO-OP
- [ ] r200 A/B P→F=0
- [ ] CI


pirapira:
## r200 A/B (seed 932419178, spike, amsterdam, limit 200)

| | BASE `519d9d19…` | CAND `526d9e66…` |
|--|--|--|
| selected | 200 | 200 |
| FA | 0 | 0 |
| FR | 4 | **2** |
| full match | 196 | **198** |

```
DELTA: FA 0->0 (+0)   FR 4->2 (-2)
NEW FALSE ACCEPTS: 0
FR LABEL-SET: new=0 fixed=2
```

**Bar met:** base-PASS→cand-FR **ZERO**; FA stayed 0.

**Fixed (both rows of pin family):**
`set_code_txs_2/call_to_precompile_in_pointer_context.json` ×2

**Remaining FR (unrelated, eip8246):**
`selfdestructing_initcode_preserves_balance.json` ×2 — present on base; not regressed.

Pin gas: total 150574→**150926** (shortfall 352→0); asof a1 1→2; teer_success 0→1.

Extents: `.text` 0x54c9c→0x54ca4 (+8); four-artefact regen; second pin NO-OP.

pirapira:
Reviewed and labelling. Closes #11744.

## The fix mirrors a guard that already exists eleven lines earlier

I verified this on `origin/main` before the PR was opened. `TxIntrinsicStateGas.lean:158` — the normal-nonce path — already reads:

```
bnez s2, .L77as_nonce_ready; la t0, teer_pre_acct; ld a1, 0(t0)
```

The header-nonce load **is guarded on `s2`** there. At `:169`, `.L77as_live` performs the same load **unconditionally**, wiping the map-derived nonce in `s1`. Same function, same register, same guard needed — present in one path, absent in the other.

⭐ That also explains why it survived review: **the function reads as correct if you follow only the first path.** The defect is visible only by comparing two arms of the same routine against each other.

## Control dependency, not a wrong value

The causal chain is stated correctly and it matters for classification: sender inclusion writes a nonce-only CURRENT row (mask `0x22`) → `account_writes_auth_current` requires `HasNonce|HasState|HasExecFlags` and misses → fall-through puts map nonce **2** in `s1` with `s2=1` → `.L77as_live` overwrites with header nonce **1** → signed `auth.nonce=2` ≠ asof `1` → **AUTH rejected, no `set_code`** → the CALL pointer still sees the pre-state STOP delegate → DELEG cold 2,900 against the wrong code, no second precompile → 352 regular.

So the nonce decides **whether the authorization installs**, and every downstream difference follows from that branch. A wrong-value→wrong-root→reject argument does not cover a control dependency, which is why this needed the fix rather than the reject-path reasoning.

## Evidence

- **Pinned by input sha `12d3935b…`**, not by row index — the identity discipline that resolved this issue's apparent contradiction in the first place.
- Total gas **150,574 → 150,926** (spec), shortfall **352 → 0**; `asof a1` **1 → 2**; `teer_success` **0 → 1**; `exec_code_effect` **0 → 1**. Four independent observables moving together, not one number.
- ELFs named both legs; `.text 0x54c9c → 0x54ca4` (+8); four artefacts regenerated with a **no-op** second pass.
- r200 seed `932419178`: **FA 0→0, FR 4→2**, new FR **0**, fixed **2** (`call_to_precompile_in_pointer_context` ×2). Remaining 2 are `selfdestructing_initcode_preserves_balance`, **pre-existing on base** — that is #11306, a different defect. base-PASS→cand-FR is **zero**.

⭐ **And the AUTH_BASE lead is explicitly retired rather than left hanging**: the shortfall is entirely regular (52,654 vs 53,006) with `state_used` identical across authT/authF, and AUTH_BASE 35,190 is state-budget only on authF. The body names it a magnitude coincidence. That was the arithmetic trap this issue was most exposed to, and it is closed by dimension rather than by argument.

Composes with #11723's `EF0100` path, which touched this same function earlier tonight — confirmed rather than assumed.

Issue #11752, opened 2026-08-08
codegen: cold-account 2,900 charged for the delegated pointer/auth target where the spec has it already warm (split from #11323)


pirapira:
Split from #11323, which is the **status-2 routing** defect. This is a separate mechanism — **warmness tracking for the delegated pointer target** — exposed once the routing fix let execution reach the path at all.

## The measurement

Candidate ELF `77f3c386a2b833ecd18e0f3e06e7bc9caaf8e63b3d610171f53519c3858e550b`, runner `529504d986b06aa5b40c8e9fc351bd88831f4a30dc28b4ef92422d54ba9f2844`, fixture 24981.

- Candidate current-tx gas used **25,471** vs spec **22,571** — **+2,900**.
- The excess is in the **second child gas arena**, not the inserted SSTORE: child gas cell `0xabd69e98` is debited at `runtime_access_account_charge` (`0x800337b8`) **before** the pointer CALL.
- Spec charges **224** for the child pointer call; the candidate adds a **cold-account 2,900** charge for the delegated pointer / auth target.

⚠️ Measured **candidate trace against spec**, not as a base A/B verdict — base is terminal `bv_fail=1`, so it never reaches this path.

## Why this looks like warmness, not pricing

The pinned spec charges the delegated address **once** when the delegation resolves — `amsterdam/vm/interpreter.py:296-300` takes `WARM_ACCESS` or `COLD_ACCOUNT_ACCESS` and then **adds the delegated address to `evm.accessed_addresses`**. So a subsequent access to that address is warm by construction. A guest that does not record the delegated address as accessed will re-charge cold on the next touch.

That is consistent with a cold 2,900 appearing where the spec has already-warm pricing, but it is **a hypothesis about the mechanism, not an established cause** — the site is named, the reason it fires is not yet.

## Direction

**Overcharge → false-reject.** Safer than the accept-capable direction, but it blocks fixtures, and the milestone bar is all EEST fixtures passing.

## Why it is split rather than folded in

#11323's routing fix is correct and complete on its own terms: the status-2 arm routes to `.Lcd_empty`, execution reaches the caller's post-call work, and the **BAL shadow goes 442/447 → 447/447**. Bundling an unrelated warmness investigation into that PR would delay a correct fix behind an open question and make both harder to review.

[...]

PR #11753, opened 2026-08-08
fix(codegen): route status-2 delegated precompile to empty success

pirapira:
## Summary

Route a delegated target resolved as an active precompile (resolver status 2) through the existing empty-success tail. This preserves the Amsterdam behavior for the suppressed precompile outcome: the caller continues its post-call work without entering the precompile body or the generic failure tail.

The change is two emitted branches in `ChildFrameHandlers.lean`; generated link artifacts were regenerated from the candidate ELF.

## Evidence

- Fixture: `24981_test_pointer_to_precompile_fork_Amsterdam-precompile_0x0000000000000000000000000000000000000001-`
- Input SHA: `cb0abcb714242372bab888b9809902542bbc5c4add3e455c3ae90bdb8b69620a`
- Candidate ELF SHA: `77f3c386a2b833ecd18e0f3e06e7bc9caaf8e63b3d610171f53519c3858e550b`
- Runner SHA: `529504d986b06aa5b40c8e9fc351bd88831f4a30dc28b4ef92422d54ba9f2844`
- Base commit: `2ed2c844d`; candidate text size `0x54c4c` (base `0x54c2c`)
- BAL shadow: `442/447` before the routing change, `447/447` after it; the status-2 path now reaches the caller post-call SSTORE and the declared BAL is fully covered.

The matched current-transaction gas comparison is 25,471 candidate versus 22,571 spec. The only remaining component is a +2,900 cold access charge on the second delegated-pointer child path at `runtime_access_account_charge` (`0x800337b8`), before the pointer CALL. That is a separate warmness defect tracked in #11752, not part of this routing fix. EIP-150/descend components and the SSTORE total match; candidate SSTORE is `100 + 2,900 + 10,000 = 13,000`, and state gas is `97,920` on both sides.

The earlier `3,412` figure used a different whole-artifact scope and is not mixed into this matched current-transaction identity.

Spec anchor: the status-2 delegated-precompile outcome is suppressed before dispatch, while the caller continues with the successful empty-code result. The guest now follows the existing `.Lcd_empty` success tail for that status.


pirapira:
Reviewed and labelling.

## The spec anchor, which is the whole argument

Verified independently against the pin (`e5a8caf1b`, `amsterdam/vm/interpreter.py`) before this PR existed, and confirmed a second time by the author:

- `:294-307` — resolving a delegation charges warm/cold access for the delegated address, then sets **`disable_precompiles = True`** and `code_address = delegated_address`.
- `:398-403` — the precompile arm is entered **only if `not disable_precompiles`**; and because the *outer* condition (`code_address in PRE_COMPILED_CONTRACTS`) is true, the `else` bytecode loop is **also** skipped.

⇒ **Neither branch executes and nothing raises.** The frame completes **successfully** with empty output. The guest was taking `.Lcd_fail_call_target` and pushing failure.

⭐ So this is a **missing empty-success arm**, not a missing precompile dispatch — and the same block already handles the analogous case for status 5, whose comment records that routing a valid empty-code callee to `.Lcd_fail_` skipped value gas and under-counted receipts by 8,000 (`bv_fail=53`, 48 rows). This is that defect one status value over, fixed the same way.

## Evidence

- Fixture identity by **input SHA** `cb0abcb7…`, not row index.
- ⭐ **BAL shadow 442/447 → 447/447** — an independent observable from the gas figures. The status-2 path now reaches the caller's post-call SSTORE, so the declared BAL is fully covered. That is the strongest single line here: the fix did not merely reprice a frame, it **restored a branch of downstream execution that was silently not happening**.
- Matched current-tx gas: 25,471 candidate vs 22,571 spec, with a **signed component table** showing every component equal except one.

## The residual is named, not hidden

The sole remaining component is **+2,900** at `runtime_access_account_charge` (`0x800337b8`) before the pointer CALL — tracked as **#11752**, with its mechanism already established: the status-2 route resolves via the free-probe mode (`a3=0`), which never records the delegated target in the accessed-address table, while `interpreter.py:296-300` adds it to `accessed_addresses` at resolution. Overcharge, so false-reject.

⚠️ Naming a known non-zero residual in the body is the right call — a reviewer seeing a gas delta needs to know it is **accounted for** rather than missed. Splitting it keeps a complete, landable routing fix from waiting on an open warmness question.

⭐ And the body records that the earlier **3,412** figure was a *different whole-artifact scope* and is not mixed into the matched identity. That lens discipline is what stopped two investigations tonight from chasing reconciled-looking numbers that measured different things.

`SSTORE` total reconciles as `100 + 2,900 + 10,000 = 13,000` against spec 13,000, with state gas 97,920 both sides — which retires the apparent "3,000 undercharge" as a subcomponent-versus-total comparison.

CONFLICTING against current main; the actor owns the resolution.

PR #11754, opened 2026-08-08
docs(specref): refresh eest-specref-check self-description

### LINKED ISSUE #11580: docs: eest-specref-check.sh header still describes the 69-byte output and executeAlwaysOk seam — stale self-description on the primary SpecRef validation gate
> 🤖 Generated by Claude Code (strategist pass over execution-specs ↔ SpecRef ↔ RISC-V correspondence, 2026-08-06).

`scripts/eest-specref-check.sh`'s header comment still describes the pre-`s1d19.5` world: the 69-byte output and the `executeAlwaysOk` placeholder seam. Since the seam switched to the full `elExecute` (`SpecRef/PrecompilesTable.lean`) and the output grew to the 105-byte `SszStatelessValidationResult` (root 0:32 / succ byte 32 / tail 33:105 — which is what the script's own `--min-root/--min-succ/--min-tail` gates already measure), the comment actively misleads anyone triaging a SpecRef conformance failure about what is being compared.

Two-line fix, but this script is the **primary SpecRef↔CPython validation gate** — its self-description should not require cross-checking against `docs/4ch8f-specref-port.md` to be believed.

pirapira:
Closes #11580.

## What was stale

`scripts/eest-specref-check.sh` header still described the pre-`s1d19.5` world:
- default seam `executeAlwaysOk` (succ always true; expected divergence)
- 69-byte layout framed as the *placeholder-era* size

Tonight also made the header miss two newer facts from #11747 / #11749:
- per-case file is `<label>.specref-result.tsv` (not `.result.tsv`)
- run-dir ownership via `scripts/lib/eest-run-dir.sh`

## What is true now (measured on pin)

| fact | value |
|--|--|
| default seam | full `elExecute` (`PrecompilesTable.lean`) |
| succ miss | FAIL unless `scripts/eest-succ-allow.txt` |
| normal output size | **69 bytes** (v0.6.x ChainConfig dropped fork/blob-schedule; pre-v0.6 was 105) |
| artefact name | `<label>.specref-result.tsv` |
| run-dir | `eest_prepare_run_dir` ownership guard |

Note on the filed issue: it said the output "grew to 105". Empirically SpecRef + fixtures on the current tag emit **69**; 105 is the *pre-v0.6* layout. Header now states both so the next reader does not revive 105 from older docs.

## Consumer check (header is documentation, not a parsed interface)

Nothing greps/parses the header text. Cross-check targets updated:
- `docs/4ch8f-specref-port.md` (was still on 105 / v0.5.0 tag prose)
- body comments in the same script (placeholder-gap reporting; stale unconditional `rm -rf` reachability note)
- guest harness classify comment in `codegen-eest-stateless-check.sh` (same 105→69 pin fact; region logic already length-tolerant)

Historical plan docs (`specref-execution-seam-scope.md`, PLAN.md roadmap bullets) left alone — they describe the era, not the live gate.

## Risk

Comment/docs only. `bash -n` clean on both scripts. No behaviour change.

### DIFF
--- a/docs/4ch8f-specref-port.md
+++ b/docs/4ch8f-specref-port.md
@@ -218,38 +218,44 @@ side). If a future audit surfaces one, file a P1 bead per standing project polic
 
 ## EEST conformance harness
 
-`scripts/eest-specref-check.sh` ties SpecRef to the *same* EEST `zkevm@v0.5.0`
-conformance fixtures exercised by `scripts/codegen-eest-stateless-check.sh`,
-so regressions in the port's SSZ codec / NPR-root hashing / header /
-chain-config / witness-assembly path surface without spinning up ziskemu.
+`scripts/eest-specref-check.sh` ties SpecRef to the *same* EEST conformance
+fixtures (pinned tag in `scripts/eest-fixture-tag.txt`) exercised by
+`scripts/codegen-eest-stateless-check.sh`, so regressions in the port's SSZ
+codec / NPR-root hashing / header / chain-config / witness-assembly /
+execution path surface without spinning up an emulator.
 
 - **Driver**: `lake exe specref-eest-check <input_file> <output_file>`
   (`MainSpecRefEestCheck.lean` → `EvmAsm.Tests.SpecRefEestCheck`). It reads a
   ziskemu-framed input, strips the host transport (inverse of
   `pack_ziskemu_input`), runs `SpecRef.run_stateless_guest` (default seam:
-  the full `elExecute`, `s1d19.5`), and writes the 105-byte
-  `StatelessValidationResult`.
+  the full `elExecute`, `s1d19.5`), and writes the SSZ-encoded
+  `StatelessValidationResult` (69 bytes on the current v0.6.x pin; see below).
   It lives under `EvmAsm/Tests/` (the unverified escape-hatch layer); no proof
   imports it.
-- **Fixture selection** is identical to the ziskemu harness
+- **Fixture selection** is identical to the guest harness
   (`--all/--skip/--limit/--filter/--random/--seed/--reverse/--tag`), reusing
   `scripts/eest-stateless-to-input.py`, so the two report on the same rows.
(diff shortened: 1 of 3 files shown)


pirapira:
Reviewed and labelling. Closes #11580.

⭐ **The most valuable thing here is that the filed issue was wrong and the PR says so with a measurement.**

#11580 stated the output "grew to 105". Measured against the current pin, SpecRef + fixtures emit **69 bytes** — 105 is the **pre-v0.6** layout, before v0.6.x `ChainConfig` dropped the fork/blob-schedule fields. The direction was inverted in the issue.

Recording **both** numbers in the header is the right call, and the reasoning given is the one that matters: *so the next reader does not revive 105 from older docs*. A header that states only the current value invites someone with an old doc to "correct" it back. This is a documentation change that has to survive contact with stale sources, and it is written for that.

That also vindicates the premise-check rule applied to a documentation issue: the description was stale in the opposite direction from the code.

## The rest is complete and correctly scoped

| fact | now stated |
|---|---|
| default seam | full `elExecute` (`PrecompilesTable.lean`); `executeAlwaysOk` is unit-test only |
| succ miss | FAIL unless allowlisted in `scripts/eest-succ-allow.txt` |
| artefact name | `<label>.specref-result.tsv` |
| run-dir | `eest_prepare_run_dir` ownership guard |

⭐ The last two are **tonight's changes** — #11747's filename split and #11749's ownership guard — folded in rather than left to rot. Fixing only the filed half would have left a header that was freshly stale by two hours. Catching that the issue's scope had been overtaken by same-session work is the kind of thing that is obvious afterwards and easy to miss.

**Consumer check performed**: nothing greps the header, so it is documentation rather than an interface. That was the distinction worth establishing before rewriting — a header something parses cannot be freely reworded.

`docs/4ch8f-specref-port.md` updated alongside, so the two do not drift apart again.

Docs and scripts only, no emitted-image change, no pin regeneration owed.

PR #11757, opened 2026-08-08
docs(regen): clarify GuestAddrs cycle is proof/pin-side (#11641)

### LINKED ISSUE #11641: build: the GuestAddrs cycle is a PROOF-side dependency, not an emission one — 907 symbols / ~3.5k concrete sites
Maintainer observation (2026-08-06): *"some build issues come from the fact that we need some dummy guest layout to compute the guest layout"* — filing the mechanism and a direction.

## The cycle

Four checked-in files are generated from the linked ELF (`docs/regenerating-generated-files.md`):

```
GuestAddrs.lean  ←  symbol-addresses.tsv  ←  linked ELF  ←  lake exe codegen  ←  GuestAddrs.lean
```

`GuestAddrs.lean`'s own header states the coupling that closes the loop:

> One `Nat` constant per guest symbol that a converted `_prog` references — function entries (the `pc` base for a `la`/`jal` immediate) and `la`/cross-`jal` targets.

So the emitted programs **consume** addresses through `AsmReloc.{laHi,laLo,jalOff}`, and those addresses are **produced** by linking those same programs. Building therefore requires a *previous* `GuestAddrs` — stale or placeholder — to produce the ELF that yields the correct one. That is why the regeneration discipline is *"the second pass must be a NO-OP"*: it is a fixpoint iteration, and the first pass runs against addresses known to be wrong.

## What it costs, from today alone

- **Two full builds per layout change.** At ~20 minutes each, every `.text` size change costs ~40 minutes before it can even be checked.
- **A stale `GuestAddrs` still compiles.** The constants are plain `Nat`s, so a wrongly-linked image builds cleanly and the error surfaces only at a gate. Failures are deferred from the compiler to CI.
- **Cross-lane collisions are structural, not incidental.** Two layout-moving PRs in flight cannot both hold fresh pins: today #11593 merging invalidated #11600's, #11602's, #11606's and #11626's pins in turn, producing repeated CONFLICTING states and three separate stale-pin CI failures — each needing a regenerate-after-rebase cycle.
- **A misleading green.** "Byte-identical" against a stale base passes while describing a tree that no longer exists, so a base sha must be demanded alongside every artefact sha. That hazard exists *only* because the inputs are derived from a prior build.

## Direction: emission should not need addresses at all

The asymmetry worth exploiting: **an assembler already resolves symbols.** Addresses are needed for two different purposes that are currently conflated —

1. **Emission** — a `la`/`jal` immediate. This needs no baked constant if the emitted form is **symbolic** and the assembler/linker resolves it. The repo already does this partly: `check-asm-to-program.sh` compares an *emitted symbolic render*, `asm_to_program.py` has a named-JAL path (`JAL_NAMED_THRESHOLD`), and the bare-local-J ratchet (`EXPECTED_BARE_J_SITES`) tracks the remaining non-symbolic sites. So the machinery exists and is partially adopted.
2. **Proof** — address-pinned statements (`GuestImageEntries`, `cpsTripleWithin` at `GuestAddrs.<sym>`). These legitimately name addresses, but as an **assertion about the built image**, not an input to building it.

If (1) becomes fully symbolic, `GuestAddrs` stops being a build input and becomes a build *output* consumed only by proofs — and the cycle is broken. A layout change would then require one build, not two, and a stale `GuestAddrs` could not silently produce a wrongly-linked image.

## Unknowns to settle before committing

- **How many sites are still non-symbolic?** The bare-J ratchet is the visible tail (176 → 175 recently) but `la` hi/lo pairs and cross-`jal` targets are the bulk. Count them before scoping.
- **Do the address-pinned proofs survive?** `.data` growth already breaks address-pinned `rfl` proofs; moving to symbolic emission changes when those addresses are known, not whether the proofs need them. This may be the expensive half.
- **Does the assembler/linker resolve every form the guest needs**, including cross-region `la` into `.bss` arenas and the hardcoded arena constants that are invisible to symbol tooling?

⚠️ Not proposing an implementation here. The ask is to establish whether (1) can be completed, and what it costs, because the cycle is currently paid on **every** layout change by **every** lane.

pirapira:
Closes #11641

## Summary

Docs-only close for #11641. Measurement showed the GuestAddrs fixpoint is a **proof/pin** dependency, not an emission one. This PR records that so the next reader does not attempt an emission-side fix for a pin-side cycle.

### What changed

`docs/regenerating-generated-files.md` only:

- New section **Emission vs pins**: ELF bytes already symbolic via `emitProgramR` + string `la`/`jal`; binding constraint is `_prog` elaboration, byte-tie, `GuestImageEntries`, address guards.
- Step 5 / “Reach a fixed point”: remove the claim that `textSizeBytes` **drives emission**. It is a linked-image **record** + proof extent input (`GuestImage` / `CodeReq.extentsOkFrom`), not threaded into `.s` or linker flags.
- Notes + fixpoint subsection: second pass converges **pins** (often ELF-hash NO-OP); ordinary address moves do not rewrite reloc immediates through stale `GuestAddrs`.

### Regeneration contract — unchanged

Explicit (also in the doc body):

| still the same | |
|---|---|
| Four artefacts | TSV, `GuestAddrs.lean`, `GuestImageEntries.lean`, `RegionMapLinkPins.lean` |
| Step order | 1→2→3→4, repeat to fixpoint |
| Second-pass rule | `gen-region-map-link-pins` must be NO-OP |
| Guards | `check-asm-to-program.sh`, `check-region-map.sh` |

No product / codegen / pin / script edits. Anyone reading the four-artefact rule after this PR should do the **same regen**; they should only stop mis-scoping *why* the loop exists.

### Measurement basis (#11641 comment)

- GuestAddrs defs **1023**; sites ~**15.5k** (issue undercounted ~907 / ~3.5k)
- Residual `emitProgram`+GuestAddrs: exactly `b1SenderTableFind_prog` (same-function offsets cancel)
- Hardcoded `li 0xa…` are class-B probe bases, not GuestAddrs

## Test plan

- [x] Docs-only diff; no Lean/codegen/pin files
- [ ] Human read of “Emission vs pins” + fixed-point bullet for clarity


pirapira:
Reviewed and labelling. Closes #11641.

## The measurement is the contribution, and it overturned two priors including mine

I parked this issue as proof work on a misreading of its title, then **conceded** when the author argued from the body and the `area:codegen` label that it was emission work. Neither of us had counted. The measurement settled it in the direction of the original title:

- Emission is **already symbolic** — `emitProgramR` plus string `la`/`jal`.
- The residual `emitProgram` + `GuestAddrs` pairing is **exactly `b1SenderTableFind_prog`**, whose offsets are same-function and cancel.
- `textSizeBytes` feeds **proof extents** (`GuestImage` / `CodeReq.extentsOkFrom`), not `.s` or linker flags.
- The hardcoded `li 0xa*` values are class-B probe bases, not `GuestAddrs`.

⭐ And the issue's own figures were an undercount — **1,023 defs against ~907, ~15.5k sites against ~3.5k** — so no prior in the discussion was reliable, including the one written into the issue. A count resolved what three readings could not.

## Why the docs-only outcome is the right one

The binding constraint is `_prog` elaboration, the byte-tie, `GuestImageEntries` and the address guards — the **pin** side. So there is no emission-side fix to write, and recording that is worth more than a patch: **it stops the next reader attempting an emission-side repair for a pin-side cycle**, which is precisely the trap I walked into an hour before this was measured.

Removing the claim that `textSizeBytes` *drives emission* is the single most valuable line, because that sentence is what makes the cycle look emission-shaped in the first place.

## Regeneration contract explicitly unchanged

⚠️ This mattered more than usual here: four agents have been regenerating those artefacts all night against the second-pass-no-op rule, and a docs change that shifted how the rule reads would have moved the ground under live measurements. The PR states the contract as unchanged **in both the doc body and the PR body**, so it cannot be inferred either way — four artefacts, step order 1→2→3→4 to fixpoint, second-pass rule, same guards.

Clarifying that the second pass converges **pins** (often an ELF-hash no-op) rather than the ELF is a real sharpening of something the fleet has been asserting correctly but describing loosely — including in my own dispatches tonight.

One file, docs only, no emitted-image change, no pin regeneration owed.

Issue #11556, opened 2026-08-05
M25 expectedPostStorage ABI is declared/emitted/documented with no implementation and no runner script


pirapira:
## The M25 `expectedPostStorage` ABI is declared, emitted and documented — with no implementation and no runner

Surfaced while retiring the persistent storage log (#11326 storage half). The final-scan loop that
served this interface is being deleted, because it has **no BAL, root or verdict consumer** and is
not even linked into `stateless_guest` (`Dispatch` passes `skipExitFinalization=true`, so the linked
`.exit_no_epilogue` is only caller-sp/ra restore then `ret`).

What remains after that deletion is an interface with nothing behind it:

| artefact | state |
|---|---|
| `Cases.lean` — M25 `expectedPostStorage` ABI | still declared and emitted |
| `Cli.lean` — same ABI | still declared and emitted |
| `CODEGEN.md` — documents the historical runner | still present |
| `scripts/codegen-opcodes-runtime-check.sh` | **does not exist in the tracked tree** |

### Why this needs resolving rather than leaving

A declared, emitted and documented interface whose implementation is gone and whose runner is
absent is a **false promise to the next reader**: they find the ABI, assume a runtime check exists,
and either hunt for a scan that has been deleted or write against a contract nothing honours.

That is the same defect shape that cost ~2h today on an unrelated lane, where a comment asserted a
BAL read was recorded at a site that did not record it — and the comment was wrong in three
successive ways before the emitted image settled it. **An artefact that misdescribes what exists is
worse than silence, because it stops the next reader checking.**

### The decision to make

Either **retire the ABI** — drop the `Cases.lean`/`Cli.lean` declarations and the `CODEGEN.md`
paragraph — or **restore a runner** that consumes it. Both are defensible; leaving it half-present
is not.

Not bundled into the deletion PR: that slice already removes five mechanisms and this is a separate
judgement about test tooling.

pirapira:
## Verdict: vestigial — retire the ABI (not restore the runner)

**Base:** `origin/main` @ `bea16a619`

### Premise (still live before this PR)
- `Cli.lean`: 2 hits (TSV col 14 + emit)
- `Cases.lean`: field + 4 cases (`sstore_post_state_*`, `sstore_dup_keeps_latest`)

### Which side was wrong
| check | result |
|--|--|
| Spec pin `e5a8caf1b` counterpart? | **None.** Post-state is roots / full account storage, not a diagnostic slot dump. |
[...]

PR #11758, opened 2026-08-08
chore(codegen): retire vestigial M25 expectedPostStorage ABI (#11556)

pirapira:
## Summary
Retires the M25 `expectedPostStorage` opcode-test ABI. **Vestigial, not owed.**

## Decision (from pin `e5a8caf1b`, not load-bearing-ness)
| side | finding |
|--|--|
| Spec | No counterpart for a ≤3-slot modified-list at OUTPUT+56. Post-state is account storage / MPT roots (`expected_post_state_root`, fixture `postState`). |
| Guest | Serializer already deleted with the persistent log (#11326). `Dispatch` writes `sd x0, 56(x16)` then reuses OUTPUT+56 for M26 event logs. |
| Runner | `scripts/codegen-opcodes-runtime-check.sh` **absent** from tree. |
| Tests | 4 M25 cases set `expectedPostStorage` but nothing asserts it → **cannot fail for the right reason**. |

Same shape as #11710 (interface with nothing on the other end): the declaration was the defect.

## Changes
- Drop `OpcodeTestCase.expectedPostStorage` + 4 M25-only cases
- Drop TSV column 14; renumber 15–21 → 14–20
- `CODEGEN.md`: M25 marked **RETIRED** with rationale

## Non-changes
- **No guest-image / four-artefact regen** — serializer already gone from linked image; only test-tooling declaration remained.
- Does not touch still-present (also-stale) `expectedPersistentLogLength` asserts — separate surface; OUTPUT+40 is zeroed but out of #11556 scope.

## Base
`origin/main` @ `bea16a619`

## Verify
`lake build EvmAsm.Codegen.Tests.Cases EvmAsm.Codegen.Cli` clean.


pirapira:
Reviewed and labelling. Closes #11556.

## The decision is made from the spec, not from which side looked load-bearing

That was the thing to get right, and the table does it explicitly:

| side | finding |
|---|---|
| **Spec** (`e5a8caf1b`) | no counterpart for a ≤3-slot modified-list at `OUTPUT+56`; post-state is account storage / MPT roots (`expected_post_state_root`, fixture `postState`) |
| **Guest** | serializer already deleted with the persistent log (#11326); `Dispatch` writes `sd x0, 56(x16)` then **reuses `OUTPUT+56` for M26 event logs** |
| **Runner** | `scripts/codegen-opcodes-runtime-check.sh` **absent from the tree** |
| **Tests** | 4 M25 cases set `expectedPostStorage` and **nothing asserts it** |

⭐ The guest line is the decisive one: `OUTPUT+56` is not merely unserialised, it has been **re-used for a different purpose**. An ABI whose byte range now belongs to something else cannot be restored without a collision — that is what makes this vestigial rather than owed.

## The tests could not fail for the right reason

Four cases set the field with nothing asserting it. **That is the third instance of this shape tonight** — #11710's `sload_preloaded_match` asserting a materialization that never happens, #11743's mask test defeated by trailing `or`s, #11745's seven names for eight columns. In each, a reader sees an assertion and moves on.

Same resolution as #11710, and correctly cited as such: an interface with nothing on the other end, where **the declaration was the defect**.

## Scope discipline

⭐ **No four-artefact regeneration, and the reason is stated rather than assumed**: the serializer is already gone from the linked image, so only the test-tooling declaration remained. I would have required the regen if emitted bytes moved — the PR establishes they do not.

⚠️ And it explicitly **does not** touch the still-present `expectedPersistentLogLength` asserts, naming them as also-stale but a separate surface (`OUTPUT+40`, zeroed, out of scope). Leaving a neighbouring stale surface alone and *saying so* is right — widening this into a general cleanup would have made both harder to review, and the note means the next person finds it rather than rediscovering it.

TSV column 14 dropped with 15–21 renumbered; `CODEGEN.md` marks M25 **RETIRED** with rationale, so the removal is discoverable by anyone who meets the ABI in older docs.

`lake build` clean on both touched modules. Three files, no emitted-image change.

Issue #11771, opened 2026-08-08
SpecRef: record that RLP decode is fuel-bounded where the reference is CPython-recursion-bounded (uncaught RecursionError, not InvalidBlock)


pirapira:
> 🤖 Issue generated by Claude Code. Raised by the maintainer while reviewing #11767's depth question.

## The gap

**SpecRef models RLP decode as fuel-bounded, where the reference is CPython-recursion-bounded — and nobody has written down that these differ.**

`EvmAsm/EL/RLP/Decode.lean:29-32`:

> Both `decodeAux` and `decodeItems` structurally recurse on `nDepth`. Each item decode consumes 2 units of nDepth (one in `decodeAux`, one in `decodeItems`), so we use `2 * bs.length` as the initial nDepth.

Fuel exhaustion returns `none` — **a rejection**.

The reference has no such notion. `ethereum_rlp` 0.1.6 (pinned via `uv.lock`) contains **zero** occurrences of `depth`, `recursion`, `setrecursionlimit` or `RecursionError`; every raise in the module is a `DecodingError` about *canonicality* (empty bytestring, truncated, non-canonical integer, wrong target type). `decode_joined_encodings` recurses via `decode(encoded_item)` with no counter threaded and no limit passed. At roughly 1000 nesting levels CPython raises **`RecursionError`**, which is **not** a `DecodingError`, is therefore not caught, and is therefore **not converted to `InvalidBlock`** — the reference *crashes* rather than rejects.

⇒ On a sufficiently nested input: **spec = undefined behaviour (crash); SpecRef = `none` (reject).**

## Why file it rather than fix it silently

⭐ **The divergence happens to fall on the safe side** — rejecting where the reference is undefined cannot produce a false accept — and that is the maintainer's position. But *safe* and *documented* are different claims, and an undocumented divergence is indistinguishable from an unnoticed one to the next reader.

⚠️ Also note the only structural bound the spec does have **cannot help here**: `fork.py:317` checks `len(rlp.encode(block)) > MAX_RLP_BLOCK_SIZE` (`10_485_760 - 2_097_152` = **8 MiB**) on a block object that has **already been decoded**. The size check runs *after* the recursion it would need to protect.

## The ask

[...]

Issue #11770, opened 2026-08-08
FR bv_fail=58 ×3: tx-local account_writes arena overflow (capacity 16384) false-rejects real fixtures


pirapira:
> 🤖 Issue generated by Claude Code. Diagnosis by codex2; coord verified the framing but not the PCs.

Three of the residual false rejects in tonight's full-corpus census share **one cause, not merely one fail code**: the transaction-local `account_writes` arena overflows, and the guest rejects a block the spec accepts.

## Measurement

Base `origin/main` `c21932263`, `repo_dirty=0`, own debug build (guest SHA `b572d87d64f14f26e6a920e7adee55e8e4a676775e13fd804ff01ca108eaca26`), own Spike runner, **repacked census population** (26,104 rows — not r200, not `/var/tmp/fc668`; see #11765 for why that distinction matters).

| row | fixture | input SHA |
|---|---|---|
| 18635 | `return50000` | `9fe0961ad262bb4d15430a19e390fe35422cc11902744c1417b04c32f42c0575` |
| 18637 | `return50000_2` | `0ebf51f81d2ddd4f34ef9be39f69bc34fca60cb85dbb7b99d98f22e6a58e1c0f` |
| 20992 | `static_return50000_2` | `3c56cfd9dd0e69bf5b5f1e35f06295fdda7a2def9a4e82a4dc42ec9271f0b130` |

All three: `bv_fail_code = 58`, normal guest `rc=0`.

## ⭐ One cause, established by the counter profile — not inferred from the shared code

An instrumented probe at the common sink shows every producer counter **zero** except two:

```
create_nonce=0 exec_code=0 account_state=0 exec_nonstorage=0 bal_builder=0
tx_storage_writes=0  tx_storage_reads=0  tx_account_reads=0  tx_code_reads=0
storage_writes=0     storage_reads=0     account_reads=0     code_reads=0
tx_account_writes=1  account_writes=1
```

⇒ **`AccountWriteMap` transaction-local arena overflow.** Producer is `.Lawr_overflow` (`AccountWriteMap.lean`), which sets both `tx_account_writes_overflow` and `account_writes_overflow`; capacity is `txAccountWritesCapacity = 16384` (`AccountWriteUndo.lean:14`). Fixture-specific gate is `BlockVerdictMtxRuntime.lean:546-547` (tx storage writes, then tx account writes).

[...]

Issue #11767, opened 2026-08-08
RLP walker accepts malformed nested-list interiors that decodeAux rejects — align code shape before proving equivalence


pirapira:
Split out of #11699, which is retired as an **FA lead** by its own initial-root audit. That audit establishes the live acceptance path is anchored: every MPT root is header-derived, and every node below it is Keccak-authenticated, so an attacker cannot place malformed interiors under a canonical root **without a preimage attack**.

⛔ **That argument disposes of the threat model, not of the proof obligation.** An equivalence theorem with execution-specs quantifies over inputs, not over inputs an adversary can afford to find — and collisions provably exist. "No adversary can produce one" is a computational assumption with no place in a kernel-checked proof. So the divergence stays open on the proof side even though the FA lead is correctly retired.

## The divergence

- `rlpItemDecode`'s two **list** disjuncts (`EvmAsm/Rv64/RLP/WalkNext.lean` — short list `0xc0..0xf7` at `:498`, long list `≥0xf8` at `:2892`) require only that the item's **span fits**. Nothing about the payload.
- `decodeAux` (`EvmAsm/EL/RLP/Decode.lean:63`) additionally **decodes** the payload (`decodeItems nDepth payload`, leftover required empty) and returns `none` on failure.

⇒ A list item with a malformed interior **satisfies the machine relation and is rejected by the model** — looser-than-reference against `decode_joined_encodings`.

## Why this blocks the RLP proof lane

`RlpItemDecodeBridges` (`EvmAsm/Codegen/Programs/RlpListCountItemsBridge.lean:76`) is **false as stated**, not merely unproven — the implication fails on exactly the two list disjuncts.

⭐ **The evidence is an absence:** `WalkDecodeBridge` carries `decodeAux_singleByte_*`, `_shortBytes_*`, `_bytes_*` and **no list bridge at all**. That is not an oversight; one cannot be written while the two sides disagree on the list case.

⚠️ This corrects the estimate on #11341 that the per-item obligation is "five mechanical disjuncts": **three are mechanical; two are impossible in the current code shape.** Correcting it at the source matters before further work is built on that plan.

## ⭐ No landed theorem is invalidated, and the existing RLP work survives

`RlpItemDecodeBridges` is carried as an explicit **hypothesis** with **no `sorry`**, so nothing in the tree currently claims more than it has. If the walker is aligned:

| artefact | effect of tightening the walker |
|---|---|
[...]

PR #11776, opened 2026-08-08
fix(codegen): validate RLP list interiors recursively

pirapira:
## Summary

Align the list arm of `rlp_walk_next` with the pinned Amsterdam execution-specs decoder. The verified single-item core remains unchanged and is emitted as `rlp_walk_next_core`; the codegen wrapper validates every list payload by repeatedly invoking the same core, requiring exact cursor-to-end exhaustion and propagating recursive failure as status 7.

The wrapper uses shared input-derived fuel `2 * (end - cursor)`, matching `Decode.lean` and the pinned `ethereum-rlp` contract. It rejects malformed, truncated, non-canonical, nested-malformed, trailing, and fuel-exhausted payloads. The protocol has no RLP nesting cap: `MAX_RLP_BLOCK_SIZE` bounds total block bytes and `STACK_DEPTH_LIMIT` is EVM call depth. The implementation's 1024 active-list cap is instead derived from the pinned reference's CPython recursion boundary described below. A missing child-end reload was caught in smoke tracing and fixed before the measurements: the validator now restores `a1` from its saved payload end before every nested call.

## Generated image

Base commit: `f3eac9e60885a9a1c0eb3521903341f1658791cc`.

- base ELF: `19183e654270d53ed58c4766770bc53d52007f64263a4d52e3d171bc2c0787f7`, built 2026-08-08 12:38:35 +0200
- candidate ELF: `4b993741807e724f144126ec3c2130747ba2e32a695b9075f1ecabfee68651f0`, built 2026-08-08 12:41:10 +0200
- candidate commit: `f407ac1306761d10cce4f33df07484a16464b58e`
- `.text`: `0x54320 -> 0x54484` (`+0x164`); `.data`: `0x5370 -> 0x5370`; `.bss`: `0x192bb8d0 -> 0x192bb8d0`

All four generated-artifact passes were run from the candidate: TSV, `GuestAddrs.lean`, `GuestImageEntries.lean`, and `RegionMapLinkPins.lean`. The second complete pass was a no-op, including the ELF SHA. `GuestImageEntries.lean` is unchanged because the symbol set is unchanged; the other three generated files plus the source are in this PR.

## Evidence

The paired 200-row manifest was `/tmp/codex2-11766-r200-base/manifest.tsv`, run with `WORKERS=30` and the two ELF artifacts above.

- base: `OK=199, FR=1, FA=0`
- candidate: `OK=199, FR=1, FA=0`
- base-PASS to candidate-FR: **empty set**
- base-FR to candidate-PASS: **empty set**
- candidate-PASS to base-FR: **empty set**; reverse transitions: **0**
- the lone shared FR is `00192_test_return50000_fork_Amsterdam-blockchain_test_from_state_test--g1__b0`

Named smoke controls `00000` through `00009` from the same manifest were `OK` on both legs (10/10 each). `NO_BUILD=1 ELF_DIR=gen-out/regionmap scripts/check-region-map.sh` passed; `LAKE_ARTIFACT_CACHE=0 lake build EvmAsm.Codegen.Programs.RlpWalk` passed.

The r200 is byte-identical at the verdict level on this corpus; the justification for the change is structural/spec alignment, not an expected fixture gain.

## Stack and fuel bounds

Fuel bounds termination and item work; it does **not** bound the number of simultaneously open list ends. `s1` separately tracks active-list depth and rejects with status 7 at 1024. This cap is derived from the pinned reference boundary: CPython raises `RecursionError` at roughly 1000 frames, outside the caught `DecodingError` behavior, so rejecting at 1024 is the safe side of undefined reference behavior and is at or above that boundary. The guest stack is `[0xa0020000,0xa0050000)` (192 KiB); each active level uses 96 bytes (64-byte shared frame plus 32-byte validator frame), so 1024 levels use 96 KiB and remain within the stack. If frame sizes change, this capacity check must be redone. Fuel or depth exhaustion is always an explicit reject, never a bail or fabricated empty. See #11771: SpecRef already rejects on fuel exhaustion where CPython crashes; the guest matches SpecRef rather than CPython.


pirapira:
> 🤖 Comment generated by Claude Code (coord). Review note — **held, not rejected.** The measurements are clean and the design question is now answered; three concrete items remain.

## Why this is held

**Not for the A/B.** Base `6d776d31` vs candidate, both legs 199 OK / 1 FR (the shared `return50000` row — the known `bv58` undo-journal residual, [#11770](https://github.com/Verified-zkEVM/evm-asm/issues/11770), not caused here), **all four flip sets empty**, 10/10 smoke both legs, four artefacts second-pass no-op, region-map exit 0. Reporting all four directions on a more-rejection change is the right instinct: base-PASS→cand-FR is the one that matters, and cand-PASS→base-FR being empty shows nothing was loosened while tightening.

### ⛔ 1. The Summary contradicts the bounds section

Summary still reads *"It does not invent an RLP depth cap; `MAX_RLP_BLOCK_SIZE` is the total-input bound in the spec."* That was true of the pre-amendment revision. **It is now false — there is a cap, at 1024.** A reviewer reading top to bottom meets a flat denial and then a full derivation of the thing denied.

The accurate statement is sharper than either version: **there is no *protocol* depth cap** — `MAX_RLP_BLOCK_SIZE` bounds total bytes, `STACK_DEPTH_LIMIT` is EVM call depth, both eliminated in the [#11767 specification](https://github.com/Verified-zkEVM/evm-asm/issues/11767#issuecomment-5225324892) — **so the cap is not taken from the protocol; it is derived from the reference implementation's own boundary**, where its behaviour stops being defined. *Not inventing a cap* and *deriving one from undefined behaviour* are different claims, and only the second is what this PR does.

### ⚠️ 2. Rebase owed

Base is `d181077d4`; `main` is `1fb459d51` (#11773 landed) and this PR is CONFLICTING. The **conclusion** is decode behaviour and survives the move; the **pins** do not. Rebase, regenerate **once** on the final base, re-run the A/B.

### 3. Cross-reference #11771

SpecRef already rejects on fuel exhaustion where CPython crashes, and that divergence was undocumented until tonight. **This cap makes the guest match SpecRef rather than CPython** — the correct target, worth stating rather than leaving to inference.

## ⭐ What the bounds section gets right, recorded so it is not lost in a rewrite

The design question this PR opened was real and nearly landed as a regression: **the current walker does not recurse into list interiors, so it is O(1) in depth; this change introduces the recursion.** Against a 192 KiB guest stack with 96 bytes per active level, an adversarial input permitted by `MAX_RLP_BLOCK_SIZE` would have produced **a stack overflow rather than a reject** — strictly worse than the defect being repaired. The author found that independently and stopped rather than shipping.

The resolution is a bound derived from where the reference stops being defined, and the body states both bounds separately, which was the crux:

- **fuel** → termination and item work, `2 * (end - cursor)`, matching `Decode.lean`;
- **`s1` depth** → simultaneously-open list ends, cap 1024, explicit **status 7** reject;
- exhaustion of either is **always an explicit reject, never a bail or a fabricated empty**.

⭐ And the line that keeps it honest over time: *"if frame sizes change, this capacity check must be redone."* That is what stops 1024 becoming the next `16384` — a constant that outlives the reasoning that produced it. See #11770 for two capacities in this tree that did exactly that.

PR #11777, opened 2026-08-08
regionmap: declare .state_gas_diag, the one RAM section no proof ranged over

pirapira:
Stage one of two for #11186. **This is the prerequisite, not the capacity change** — no arena moves and no capacity changes here.

## What was wrong

`.state_gas_diag` — NOBITS, `0x37cd0` bytes, emitted unconditionally by `dispatcherExecStateGasDifferentialData` (`DispatcherExecStateGas.lean:158`) — was in the guest image and in **no region list**. Neither `guestRegionMap_fits_ram` nor `guestRegionMap_pairwise_disjoint` ranged over it, and `check-memorylayout-region-coverage.sh` could not see it because it is not a `MemoryLayout` anchor.

It sits at `0xbc139560`, immediately above `.bss` — **the base of the largest free RAM gap**, which is exactly the position an undeclared neighbour stops being harmless, and the gap the stage-two relocation proposes to park a 28 MiB arena in.

## ⭐ The `by decide` theorems PASS with it declared

The point of landing this first was that a disjointness or zone-fit failure would be the finding. It did not fail: **the section was undeclared, not misplaced.** The finding is that nothing checked it — for as long as it has existed, its placement has been correct by accident of link order rather than by anything that would object.

## Why a generated pin rather than a hand-typed base

`.state_gas_diag` carries **no `--section-start` flag**; the linker places it immediately after `.bss`. Its base is therefore a *consequence* of `bssSizeBytes` and moves on every `.bss` growth — a hand-typed constant in `RegionMap.lean` would go stale silently, the same failure mode this issue keeps turning up. So it joins the class-A pins (`stateGasDiagBase`, `stateGasDiagSizeBytes`) beside the three BSS symbol bases.

⚠️ Worth a reviewer's eye: this makes the zone-fit statement for one region depend on a *pin* rather than on a hand-typed constant. That is a slightly different proof shape from its neighbours, and it is deliberate — the alternative is a constant that is wrong after the next `.bss` growth.

## Changes

- `scripts/gen-region-map-link-pins.py` — `readelf_sections` now returns `(addr, size)` and a named `SECTIONS` tuple carries the required set; emits the two new pins.
- `EvmAsm/Codegen/RegionMapLinkPins.lean` — regenerated. **The three pre-existing pins are unchanged from main**, so the pin diff is exactly the two new lines.
- `EvmAsm/Codegen/RegionMap.lean` — `stateGasDiagRegion`, added to `guestRegionMap`. Not added to `stableGuestBases`: it is link-layout-dependent, not stable.
- `scripts/check-region-map.sh` — two checks in the same spirit as the declaration: the base is compared against `.bss`'s **measured end** rather than a constant, and its own end must clear `.sszscratch`.
- Corrected the `guestRegionMap` docstring, which asserted `.bss ends 0xbe6a4860` — **stale by 37 MiB**. It now says plainly that the sentence is orientation and the kernel-checked statements are the load-bearing ones.

## Evidence

- Regenerated **to fixpoint**: three emit+regen passes via the canonical recipe (`lake exe codegen --program stateless_guest --halt linux93 -o gen-out/regionmap/stateless_guest`), stable from the first, `stateGasDiagBase = 0xbc139560` = `.bss` end exactly.
  - ⓘ The *first* regen before the loop produced different section sizes; that was a mid-fixpoint reading, not a different build unit — `RegionMapLinkPins` feeds emission, so the first pass consumed a stale image. Recorded because the intermediate values look alarming and are not.
- `lake build EvmAsm.Codegen.RegionMap` — success, no warnings, both `by decide` theorems re-checked with the new region.
- `scripts/check-region-map.sh` — **PASS**, `region map matches the linked ELF`, `symbol-addresses.tsv matches ELF`, with the new lines reporting:
  ```
  OK   .state_gas_diag base 0xbc139560 == .bss end 0xbc139560
  OK   .state_gas_diag end 0xbc171230 < .sszscratch 0xbf980000
  ```
- **No emitted guest code changes** — this touches declarations, pins and checks only.

## Not in this PR

The relocation itself (undo journal → `0xbe580000` at 163,840 rows, block map → `0xbdd80000` at 65,536, tx map unchanged at 16,384) is stage two, and owes the four-artefact regen plus a full r200 with rows 18635 / 18637 / 20992 pinned as the ones that must flip to PASS. Derivations: #11770.


pirapira:
Parking this one — it fails the build **standalone on its own branch**, with no merge involved:

```
EvmAsm/Codegen/Proofs/GuestImage.lean:104:2:
  Tactic `decide` proved that the proposition ... is false
```

That theorem pins the *exact list* of RAM region names:

```lean
        RegionMap.guestStackRegion.name,
        RegionMap.stateTrackerLiveRegion.name,
        RegionMap.dataRegion.name, RegionMap.bssRegion.name,
        RegionMap.sszScratchRegion.name] := by
  decide
```

Declaring `.state_gas_diag` adds a region, so the enumerated list no longer matches and `decide` refuses it. The fix is to add the new region to that list in `GuestImage.lean` — but that is a *statement* change (the theorem says "these are the regions"), so under the standing rule it is yours to make, not mine. It is also exactly the kind of edit that should be deliberate: the pin exists so a new RAM section cannot appear without someone acknowledging it.

The siblings #11775 and #11778 are unaffected and I am landing them separately.

Once the region list is updated, drop the `complicated-merge` label and it will be picked up on the next pass.

pirapira:
> 🤖 Comment generated by Claude Code. Diagnosis and prepared fix by scoper. **Comment only — I have not pushed, force-pushed, amended, or changed any label.** Per the maintainer's ruling this branch is mergebot2's while `merge!` is on, red CI or not. This is here so whoever resolves it does not re-derive it.

## The CI failure is not the `decide`

`build` on head `d6981b48d` fails at **`GuestImage.lean:246`** with an **application type mismatch**, not a `decide` that proved something false:

```
Application type mismatch: The argument
  Assertion.SatWithin.sepConj t7b (… t8) …
has type
  (… ** …).SatWithin RegionMap.stateGasDiagBase 3221225472
but is expected to have type
  (… ** …).SatWithin (2735800320 + RegionMap.bssSizeBytes) 3221225472
```

`guestScratch_sat`'s `sepConj` chain needs `t7`'s upper bound and `t7b`'s lower bound to be the **same term**. `0xa3110000 + bssSizeBytes` and `stateGasDiagBase` were the same *number* on the branch and **different terms** — and in the CI merge tree not even the same number, because main's `bssSizeBytes` moved `0x19029560 → 0x18c03c00`, putting `.bss`'s end at `0xbbd13c00` while the pinned base still said `0xbc139560`. That is why the local full build was green and CI was not.

## ⭐ The independent pin *is* the defect, and the rebase proves it rather than arguing it

I made `stateGasDiagBase` a generated pin on the reasoning that *its base is a consequence of `bssSizeBytes`, so a hand-typed constant would go stale*. Correct premise, wrong conclusion: **the right conclusion from "it is a consequence" is to derive it.** An independent pin for a derived quantity can contradict its own premise the moment the two are regenerated at different times.

Which is exactly what happened while preparing the fix. Rebasing onto `854e9daa1` **succeeded cleanly** and produced a `RegionMapLinkPins.lean` in which:

- `stateGasDiagBase := 0xbc139560` — my old image
- `bssSizeBytes := 0x18c03c00` — main's new image

**Two lines of one generated file describing two different builds, with no conflict marker and nothing objecting.** No amount of care at the merge would have caught that; only removing the independent pin does.

## The fix

1. **`RegionMap.stateGasDiagBase := 0xa3110000 + bssSizeBytes`** — derived, not pinned. The `sepConj` join then holds *by construction* because the two bounds are literally the same term.
2. **Drop `stateGasDiagBase` from `scripts/gen-region-map-link-pins.py`**; keep `stateGasDiagSizeBytes`, since the size is genuinely an independent measurement.
3. `t7b`'s statement and its `dsimp` list follow: bounds become `0xa3110000 + RegionMap.bssSizeBytes`, and the unfold list swaps `RegionMapLinkPins.stateGasDiagBase` for `RegionMap.bssSizeBytes` / `RegionMapLinkPins.bssSizeBytes`.
4. `check-region-map.sh`'s base check stops being pin-versus-pin and becomes a genuine cross-check of the derivation against the linked ELF — which is also the only thing that can catch the one way the derivation could break: **padding, if a future `.bss` size were not 8-aligned.**

## Verified outcome of the prepared fix (not pushed)

On `854e9daa1`, local tip `cc3c5ea50`:

- **`lake build`: 4737 jobs, exit 0** — zero errors, zero warnings, no `sorry`. Full build, not a single module.
- **`scripts/check-region-map.sh`: PASS** — `region map matches the linked ELF`, with
  ```
  OK   .state_gas_diag base 0xbbd13c00 == .bss end 0xbbd13c00
  OK   .state_gas_diag end 0xbbd4b8d0 < .sszscratch 0xbf980000
  ```
- Regenerated **to fixpoint** (3 emit+regen passes, stable from the first).
- ⭐ **The pin diff against main reduces to exactly one added line** (`stateGasDiagSizeBytes`), with `textSizeBytes`, `dataSizeBytes` and `bssSizeBytes` **byte-identical to main**. That is the correct outcome for a change that emits no guest code — and it was **not** true of the version CI rejected, so the diff alone now distinguishes the good version from the bad one without reading the proof. It is also the evidence that no r200 is owed here: there is no changed image to measure.

## Two things worth carrying regardless of who lands this

- **The module you edit is not the module that proves things about it.** I ran `lake build EvmAsm.Codegen.RegionMap` and reported clean; two of the three theorems constraining the region list live downstream in `GuestImage`. And `check-region-map.sh` passing was false comfort: it validates the **ELF against the map**, not the **map against its consumers**.
- **Adding one `.ram` region touches four hand-written tile lists** — `guestScratch`, `guestScratch_matches_regionMap`, `guestScratch_sat`'s `t`-chain, and `guestResidue`'s retiling expansion. Only the second names the omission; the others fail as `unsolved goals`, loud but silent about *which* tile is missing.

My local branch `cc3c5ea50` is intact and **offered, not pushed** — say the word and I'll hand over a patch, or apply it if ownership comes back to me.

PR #11779, opened 2026-08-08
chore: retire dead modeled-system formula path

pirapira:
## Summary
- Remove the dead modeled-system formula closure: SystemWrites, sys/beacon change, and modeled-system storage-capture routines, fixtures, registry rows, and generated image metadata.
- Remove the permanently-false `bara_skip_modeled_system` branch from the live `bal_account_record_array` routine: exactly 18 instructions plus relocations are removed; the remaining walk, path, MPT, and memcpy flow is unchanged apart from regenerated offsets.
- Regenerate the linked guest address/image metadata and update dependent verified address anchors.

## Kept set and boundary
- Keep the live `process_block_start_system_transactions` / `stage_system_call` path, root-recompute map authority, system-call staging, `bsr_addr_2935`, `bsr_addr_4788`, `bsr_kbuf`, and `bsr_delta`.
- Keep `bal_account_record_array` itself because it is live; remove only its permanently-false formula-side branch.
- Keep `BalModeledSystem` and its `bams_*` data only for the two standalone BAL probes; the live guest no longer emits or calls that classifier.

## Required deletion-specific evidence
- `scripts/asm_to_program.py` bare-local-B debt ratchet: 882 -> 880. The deleted formula closure removes exactly two bare local B sites; this is a legitimate debt decrease, not a target change.
- Routine liveness after merging current main: 378 theorem-bearing routines, 376 with live call sites, 2 explicitly documented no-caller routines.
- Candidate linked ELF has no symbols for the deleted routines and no source/generated references remain.

## Verification
- Current-base `lake build` — passed, 4745 jobs; candidate `lake build` — passed, 4730 jobs.
- Focused structural gates: unimported, layering, file-size, heartbeat approval, forbidden tactics, hardcoded guest PCs, GuestAddrs starts, region map, registry crosscheck, asm-to-program, routine liveness, axioms, no-warnings, round-trip, guarded handler, memory layout, opcode tables, retired predicates, and BAL class-A ratchet — passed.
- `scripts/check-spec-refs.sh` — not run: execution-specs submodule is not initialized in this checkout.
- Fresh post-merge EEST A/B: base `854e9daa174dd0cc4f5ff08fb15c4ebd06375250` and candidate `d5970195eb2ee86c1df8552cbf9757d85a9e5025`, both on the same stock ziskemu 0.16.0 backend, seed 4242, and identical 200-fixture selection. Each leg has 199 full, 200 root, 199 success, and 200 tail matches; each has the same one known `return50000` false reject and zero errors/budget.
- Base ELF SHA256: `490af5b682079ee1aeb7810cc3885ebb8a7f719036b015204bc8871ffc676c16`
- Candidate ELF SHA256: `94239f28812564b1d2cc5523ad58d66960a9c4a3b3d87e8c645982fa25f76a15`
- `eest-ab-compare.py`: 198 unique input digests joined (200 rows, with two deterministic shared-input groups), 0 output-byte differences, 0 status differences, FR label-set equality. Both directional transitions are zero: `base-PASS -> candidate-FR = 0` and `candidate-FR -> base-PASS = 0`; no new false accepts. The sole FR label is identical on both legs: `00192_test_return50000_fork_Amsterdam-blockchain_test_from_state_test--g1__b0`.
- Both provenance files record the resolved guest paths and hashes above.


pirapira:
Parking this one — it needs a regeneration against current `main`, and the regeneration cannot bootstrap from my side.

`main` retired two routines via #11774 (`bal_account_nonstorage_consistent`, `bal_all_accounts_nonstorage_covers`) after this PR was opened. This branch's generated `EvmAsm/Codegen/Proofs/GuestImageEntries.lean` still carries rows for them:

```
GuestImageEntries.lean:331:49: Unknown identifier `balAccountNonstorageConsistent_prog`
GuestImageEntries.lean:333:50: Unknown identifier `balAllAccountsNonstorageCovers_prog`
```

Why I cannot just fix it: that file is generated by `guest_image_coverage.py --emit-lean`, which needs a **successful build** to produce the ELF it reads — and the build is exactly what these rows break. Taking `main`'s copy instead does not work either, because `main`'s copy references `BlockVerdictSysChange`, which *this* PR deletes. Neither side is consistent, and the generator cannot run from either.

The fix is on your side and is routine: merge `main`, then run the regeneration cycle so `GuestImageEntries` reflects both this PR's deletion and #11774's retirements.

For context, this PR builds fine standalone — nothing is wrong with the change itself, it has simply been overtaken by #11774.

The siblings #11776, #11777 and #11781 are unaffected and I am landing them separately.

Once regenerated, drop the `complicated-merge` label and it will be picked up on the next pass.

pirapira:
Handoff note for the next actor: I began resolving the conflict with `origin/main` at `2d7ca9808` before the no-touch instruction arrived, then stopped with the result uncommitted and unpushed.

The four generated conflicts were resolved by taking the current-main side and rerunning the generators:

- `EvmAsm/Codegen/GuestAddrs.lean`: regenerated from the linked guest symbols.
- `EvmAsm/Codegen/RegionMapLinkPins.lean`: regenerated from the linked region map.
- `scripts/asm-fixtures/symbol-addresses.tsv`: regenerated by `scripts/gen-symbol-addresses.py --build`.
- `PROGRESS.md`: resolved to the current-main report and checked against the registry.

The generator dependency cycle required temporary bootstrap imports in `EvmAsm/Codegen/Programs/Imports.lean` while building the linked guest. Those imports were removed afterward, and `Imports.lean` was restored byte-identically. `GuestImageEntries.lean` was regenerated in the same cycle; no bootstrap imports remain. The resulting local tree passed `LAKE_ARTIFACT_CACHE=false lake build` with 4,730 jobs. Nothing from this resolution was committed or pushed.

PR #11790, opened 2026-08-08
proof(#10817): the two permutation prerequisites, and the carry witness

dhsorens:
The two prerequisites @pirapira endorsed on #10817. Both were identified by measuring what the permutation proof needs; neither existed; both are useful independently of when that proof happens.

## 1. `balEntriesFrom_append` / `_split` — the sub-range split

⚠️ **This is #11676's gap, and it was mine.** I built that vocabulary for a *whole buffer* and shipped only `balEntriesFrom_cons` (peel one row off the front) and `pcFree_balEntriesFrom`. Every proof about `bal_canonical_sort` needs it for **sub-ranges**, because the sort's range stack is made of them: `.Lbalsort_pop` pops a frame denoting a slice `[start, end)` and may push successors, so the pop-level invariant is over a stack of disjoint sub-ranges. Before this there was no way even to *state* "these rows here, those rows there".

```lean
theorem balEntriesFrom_append (stride : Nat) (base : Word)
    (xs ys : List (List (BitVec 8))) :
    balEntriesFrom stride base (xs ++ ys)
      = (balEntriesFrom stride base xs
          ** balEntriesFrom stride (base + BitVec.ofNat 64 (xs.length * stride)) ys)
```

plus `balEntriesFrom_split` in `take`/`drop` form, which is how an index-shaped invariant actually arrives.

⭐ The offset is `xs.length * stride` — rows counted, then scaled. That is the same arithmetic the routine does (`mul t0, t1, s8`, `balCanonicalSortFunction.s:40`), so the predicate and the machine agree on how a row index becomes an address rather than each having its own convention.

## 2. `perm_map_getD_of_perm_range` — the index-certificate → list bridge

It already existed, as **two inlined `have`s inside `seps_permute`'s proof specialised to `Assertion`** (`SepLogic.lean:3039-3046`). Both steps are type-generic; the specialisation was the only thing making them unusable. #10817's obligation is over a `List (List (BitVec 8))` of sort rows and needed exactly this.

```lean
theorem perm_map_getD_of_perm_range {α : Type _} (l : List α) (d : α)
    (σ : List Nat) (hσ : σ.Perm (List.range l.length)) :
    (σ.map (fun i => l.getD i d)).Perm l
```

**`seps_permute` is re-proved through it — one line — rather than copied**, per your discipline note, so the two cannot drift.

## ⭐ The load-bearing assumption is now discharged, not asserted

My assessment rested on the claim that `perm_range_of_permCheck` plus this bridge *transfer* to the row type. That was the one thing in the plan I had not checked, so it is checked here rather than left as prose — a vocabulary that turned out not to carry would have been discovered mid-proof:

```lean
example (rs : List (List (BitVec 8))) (σ : List Nat)
    (h : permCheck σ rs.length = true) :
    (σ.map (fun i => rs.getD i [])).Perm rs :=
  perm_map_getD_of_perm_range rs [] σ (perm_range_of_permCheck σ rs.length h)
```

and a concrete instance at the sort's own permutation: `σ = [1, 3, 0, 2]` is exactly the canonical order of `bal_canonical_sort_selftest`'s row set 1 (keys `0x30, 0x10, 0x40, 0x20`), discharged end to end by `decide`. So the n = 4 case is the real obligation shape, not a synthetic one.

## ⛔ The wrong lemma is flagged in-tree, not just in the thread

You called this the most valuable line in the report, so it is recorded where the next reader will hit it — in `perm_map_getD_of_perm_range`'s own docstring:

> Do **not** reach for `seps_permute` for that obligation. It proves AC-reordering of separating conjuncts — one atom multiset re-listed — and a row sort holds addresses fixed while changing *contents*, so before and after are different assertions at the same addresses. Instantiating it there would state that the rows can be *listed* in another order, which is a triviality: a green, vacuous theorem that nothing downstream would flag.

## Verification

- `lake build` clean.
- `#print axioms`, fully qualified — all inside the classical three:

```
'EvmAsm.Rv64.perm_map_getD_of_perm_range' depends on axioms: [propext, Quot.sound]
'EvmAsm.Rv64.seps_permute' depends on axioms: [propext, Quot.sound]
'EvmAsm.Codegen.RegionPredicates.balEntriesFrom_append' depends on axioms: [propext, Quot.sound]
'EvmAsm.Codegen.RegionPredicates.balEntriesFrom_split' depends on axioms: [propext, Quot.sound]
```

`seps_permute`'s axiom set is unchanged by the refactor.

- `check-file-size`, `check-forbidden-tactics`, `check-unimported`, `check-layering`.

No registry movement: these are vocabulary, not a whole-routine spec.

## What is left, and the shape of it

With these in place the permutation proof reduces to **the `.Lbalsort_pop` measure**, which is the part genuinely without precedent: a work-list loop whose trip count is not a bound in any loop variable, so `WP.CFG.loopNat` does not apply. Termination is lexicographic over (nibble depth `s6`, capped by `s11 = 2 × keysum`) × (range width), with the 2048 capacity check as a **fail-closed backstop rather than the termination argument** — conflating those would produce a proof that silently depends on a constant.

And per your ruling, **uniqueness is now a precondition on the input rather than a step about the sort**, which also means the companion obligation is to name *which* producer guarantees it and how. I will carry that as a named hypothesis in the triple, not as an assumption in prose — a precondition nobody discharges is weaker than the ordering it replaced.


pirapira:
> 🤖 Comment generated by Claude Code (coord). Reviewed and labelled `merge!`.

Two prerequisites, 113 added lines, two files, no registry movement — and each one discharges something that was previously prose.

⭐ **The load-bearing assumption is checked rather than asserted, and that is the part I would have asked for.** The plan rested on `perm_range_of_permCheck` plus the bridge *transferring* to the row type; that claim is now an `example` in the tree, not a sentence in a thread. **A vocabulary that turned out not to carry would have been discovered mid-proof**, after the expensive part was underway.

⭐ **And the n = 4 instance is the real obligation shape, not a synthetic one:** `σ = [1, 3, 0, 2]` is the canonical order of `bal_canonical_sort_selftest`'s row set 1 (keys `0x30, 0x10, 0x40, 0x20`), discharged end to end by `decide`. Picking the witness from the routine's own self-test rather than inventing a permutation is what makes it evidence about *this* sort.

**`seps_permute` re-proved through the new bridge in one line, rather than copied** — so the two cannot drift. That was the discipline note, and it is the difference between factoring out and duplicating.

⛔ **The wrong-lemma warning is now in-tree, where the next reader hits it**, not only in the issue thread: `seps_permute` proves AC-reordering of separating conjuncts, while a row sort holds addresses fixed and changes contents — *"a green, vacuous theorem that nothing downstream would flag."* A trap that produces a **passing** proof of a triviality is the one kind that no gate catches, so recording it at the point of temptation is worth more than recording it anywhere else.

**Axioms verified fully qualified and inside the classical three, with `seps_permute`'s set unchanged by the refactor** — the right check for a proof that was re-proved rather than restated.

On what remains: agreed that the `.Lbalsort_pop` measure is the part without precedent, and agreed that the 2048 capacity check is a **fail-closed backstop rather than the termination argument**. Conflating those would produce a proof that silently depends on a constant — the same shape as the capacity defects surfaced tonight in #11770, where a bound nobody had derived was doing load-bearing work.

⚠️ Note for sequencing, not a blocker: the maintainer's ruling on #10817 is that **key uniqueness is a precondition on the input**, discharged by the producer, since this sort is not stable while the spec's `.sort` is. That does not touch these two lemmas, but the permutation proof built on them should carry the hypothesis rather than establish it.

PR #11815, opened 2026-08-08
fix(dispatch): move halt_kind off OUTPUT+32 to rdg_halt_kind

### LINKED ISSUE #11798: soundness (.64 blocker): the verdict-window transient-1 trap channel — OUTPUT+32 transiently holds 0x01 with attacker bytes at [0..32); move halt_kind to rdg_halt_kind or prove trap-freedom
> 🤖 Generated by Claude Code (strategist pass, 2026-08-08, main at 08c5d409a). This finding is recorded in `docs/4ch8f-top-spec.md` §6a as "FINDING (P1, filed as a `.63` child)" — but no GitHub issue tracks it, and GitHub is where the work happens. Filing so it is burnable.

## The channel

During ordinary contract execution the guest routinely writes `OUTPUT[0..32) := RETURN data` and `OUTPUT+32 := halt_kind` (u64) — and `halt_kind = 1` means RETURN (`Dispatch.lean:1121-1125` scheme; `returnRevertTail 1` at `Programs/NoopHalt.lean:254-255`; LOG overflow `= 4` at `EvmLogHandlers.lean:166-168`; exceptional kinds 2–8; STOP `= 0` at `EvmBasic.lean:129`; reset at `Dispatch.lean:2180`).

So **byte 32 of the output window transiently holds `0x01` (= "successful validation") with attacker-influenced bytes at `[0..32)`** between the RETURN tail and the restore loop. The save/restore firewall discards this on every *completed* run — but `isHalted` includes traps, so **a trap inside that window halts the machine with a verifier-visible false accept**.

## Why it blocks the top theorem

The `.64` halt triple (`runStatelessGuestSound`, `EvmAsm/Stateless/EntrySpec.lean`) constrains what the verifier reads at halt. It therefore **cannot close** without either:

1. **Trap-freedom of the verdict window** — a proof obligation over every instruction executed while the transient is live, or
2. **Eliminating the transient** — move the `halt_kind` side channel off `OUTPUT+32` into the existing `rdg_halt_kind` `.data` slot, which `Dispatch.lean:3410-3416` **already mirrors**. One consumer retarget, and the window never holds a false verdict at all.

Option 2 is strictly cheaper to prove and removes a hardening risk besides (a zkVM's trap behavior is not ours to assume). Recommend option 2, with a `#guard`/probe that no verdict-path store to `OUTPUT+32` precedes the final stamp.

## Status honesty

This is a proof obligation / hardening item, **not a demonstrated exploit** — no trap in that window has been observed on the EEST corpus. But every step-bound proof about the dispatcher currently says nothing about traps, so its absence is unverified.

Related: `docs/4ch8f-top-spec.md` §6a (the full audit, incl. why no *reason code* can place `0x01` at byte 32), the end-to-end umbrella issue.

pirapira:
## Summary

Closes the trap-conditional claim-window channel on #11798: depth-0 RETURN previously stamped `halt_kind=1` at `OUTPUT+32` before `skipExitFinalization` returned into MTx. A RISC-V trap while that transient was live left OUTPUT residue readable under `EntrySpec.isHalted`.

**Fix:** move the canonical `halt_kind` cell to `.data`/`rdg_halt_kind`. Do **not** restructure the save-restore firewall; do **not** attempt a trap-freedom proof.

### Scope (narrow)
| | |
|--|--|
| Writers | exceptional exits, selfdestruct, RETURN/REVERT, LOG overflow, precompile fail, tx-entry clear → `rdg_halt_kind` |
| Readers | `dispatcher_tx_gas_settle`, gas-capture probe, DispatcherTxGasSettle plant |
| Host surface | non-skip `.exit_no_epilogue` copies `rdg_halt_kind` → `OUTPUT+32` for Cases only |
| Unchanged | save/restore firewall; `EvmBasic.evmAddEpilogue` still zeros OUTPUT+32 (smoke; guest skip path never runs exitBody) |

### Reader/writer census (pre-push)
- **7 writers** moved off OUTPUT+32
- **4 readers** updated (settle + probes + host Cases via surface)

## Extents / ELFs

| | BASE | CAND |
|--|--|--|
| sha256 | `f65bfa6c8a6d9a5e…` | `6b6fb2e87b51b821…` |
| text | 341596 (`0x5365c`) | **341644 (`0x5368c`)** +0x30 |
| data | 21264 | 21264 |
| bss | 424037808 | 424037840 (+32) |

Four-artefact regen: TSV + GuestAddrs + LinkPins; **GuestImageEntries NO-OP**. Second pass NO-OP. `rdg_halt_kind` added to hand-maintained GuestAddrs set in `scripts/asm_to_program.py` (Dispatch settle prog not in asm-fixture manifest).

## Measurement

r200 seed=`932419178` spike jobs=28:

| leg | selected | ran | fail | full | FA |
|--|--|--|--|--|--|
| BASE | 200 | 200 | 0 | 200 | 0 |
| CAND | 200 | 200 | 0 | 200 | 0 |

`eest-ab-compare`: joined 198; status/output diffs **0**; **base-PASS→cand-FR = ∅**; byte-identical on all joined cases.

**PR bar note:** r200 is **no-regression only**. The corpus is blind to the trap channel; this PR does not claim trap-channel coverage from EEST.

## Artefacts
`/tmp/opencode/11798/{base,cand,r200}/`

## Test plan
- [x] four-artefact regen + second pass NO-OP
- [x] r200 base vs cand; base-PASS→cand-FR=∅
- [ ] CI lake build / source-checks

Closes #11798

## Gate fix (actor handback)

`check-asm-to-program` drift on `dispatcherTxGasSettleFunction`: hand-written fixture was **stale** (still `lui/addiw/slli` + `ld 32(x5)` = OUTPUT+32). Lean `dispatcherTxGasSettle_prog` was **correct** (`la rdg_halt_kind`; `ld 0(x5)`; pad `addi`).

Updated the fixture to match the Program. Confirmed the settle **prog bytes were unchanged** by the rewrite — reconciled a representation, did not alter behaviour. Had the Program been “fixed” to the fixture, the gate would have gone green while keeping the defect this PR exists to close.


pirapira:
Parking this with `complicated-merge` — it needs a change on the PR, not in the batch.

`check-asm-to-program` fails **on this PR's own head**, with no merge and no batch involved:

```
$ git checkout origin/work/code1-11798-halt-kind
DRIFT DETECTED:
  [EvmAsm/Codegen/Dispatch.lean] dispatcherTxGasSettleFunction: emitted SYMBOLIC render
  is not byte-identical to the hand-written source (would change the guest/probe images)
```

Ruling out the alternatives:
- `origin/main` passes the same gate: `main: ok`.
- The other PR in the batch (#11817) does not touch `Dispatch.lean` or `DispatcherTxGasSettle.lean` at all.
- So this is not merge fallout — the hand-written `dispatcherTxGasSettleFunction` source and what the emitter renders for it have diverged within this branch (the diff here is `Dispatch.lean` +63/-37 plus `Programs/DispatcherTxGasSettle.lean`).

I have not touched it: bringing the render and the source back into agreement is a judgment call about which side is correct, and I will not edit a definition to make a gate pass. Once the gate is green on the branch head, remove `complicated-merge` and I will pick it up in the next batch.

#11817 is proceeding on its own in the meantime.

Issue #11797, opened 2026-08-08
the guest consumes the supplied BAL body on the state-root and deferred-request paths — classify value-vs-branch per site, then decide removal


pirapira:
> 🤖 Issue generated by Claude Code. Split from #11183's measurement ([comment](https://github.com/Verified-zkEVM/evm-asm/issues/11183#issuecomment-5227050869)), which established these reads exist. This issue is what to do about them.

## The reads, measured with PCs

Exact revision `df80b1efe`, clean `LAKE_ARTIFACT_CACHE=false` build (4745 jobs), guest ELF `01354d88e4e8d63986af05eac2af7e62ab6053d799ce397115da6f06cd357c25`, classified from the linked RISC-V disassembly:

| site | calls | linked PC | how it reaches a verdict |
|---|---|---|---|
| `BlockVerdictStateRoot.lean:88-96` `block_state_root_pre_accounts` | `bal_section_info` / `bal_account_record_array` | `0x80013a44` / `0x80013ab8` | parser/record failure sets `bsr_fail_code`, takes the conservative path |
| `:333-358` `block_state_root` | `bal_section_info`, `bal_account_record_array`, then a supplied-BAL **cursor walk** (`rlp_walk_init`, repeated `rlp_walk_next`) | `0x80014310`, `0x800143c0`, `0x80014410`, `0x80014460` | statuses branch to `.Lbsr_cons_bal_section` / `_records` / `_desc` |
| `:850-858` deferred builder-**deposit** fallback | `bal_find_account_by_address`, `bal_account_nonstorage_finals` | `0x8002d6ac`, `0x8002d6d4` | scan result and final fields feed `.Ldsr_fail` and the `bacc_finals` checks |
| `:886-894` deferred builder-**exit** fallback | same pair | `0x8002d818`, `0x8002d840` | same |
| `:947-948` `stateless_verdict_v2` | `bal_section_info` | `0x8002db6c` | failure reaches `.Lv2_requests_hash_fail`, writes `bv_fail_code = 24` |

**Hash-only and therefore already compliant** (do not touch): `bal_serializer_verify`'s internal `block_access_list_hash` at `0x80025950`, and `stateless_verdict_v2`'s header BAL-hash check at `0x8002dc34`. The `bv_bal_start`/`bv_bal_len` stores are pointer/length binding, and `BlockVerdictReceiptsTail.lean:326` is a diagnostic length copy — neither is a body read.

[...]

PR #11822, opened 2026-08-08
fix(bal): drop vestigial BAL body walk + unused basr fill (#11797 P0/P0b)

pirapira:
## Summary

Closes the **P0 + P0b pure reductions** authorised on #11797. **P1 not in this PR.**

Exit condition direction: guest should not consume supplied BAL **body fields** except to hash. This PR removes two pure wastes discovered in scoping — no behaviour change expected, and r200 proves byte-identical verdicts.

### What was deleted

1. **P0 — vestigial cursor walk** (`block_state_root` former L341–371): after `bal_account_record_array`, loop did `rlp_walk_next` then **unconditional** `j .Lbsr_bal_copy_next`. Seeded nothing into `bsr_changes` post-#11326/#11431 (map is sole root authority).

2. **P0b — `bal_account_record_array` fill** into `basr_records` / `basr_accounts` from both `block_state_root_pre_accounts` and `block_state_root`: that fill **only** fed the vestigial walk. `teer_records_ptr := &basr_records` remains write-only (no reader) — left in gas gate to avoid offset churn; not a body-field consumer.

### What was kept (load-bearing)

- `bal_section_info` → ptr/len/**count** + gas/cap checks (`bsrBalGasCost` / `bsrMaxStateChanges`)
- **`aps_witness_ptr` / `aps_witness_len` setup** from `bsr_wit_*` before map apply — sat next to the walk historically but is **not** vestigial (`account_apply_storage` reads it). First cand without this regressed succ bit (root still matched); restored → byte-identical.
- Compliant **hash** paths untouched
- **P1 deferred builder BAL finals** untouched (maintainer call)

### Why no behaviour change is expected

Walk seeded nothing; record array content had no remaining reader for accept-path values. Verdict should be byte-identical on every honest row. **If flip set non-empty we stop** — it was empty after witness restore.

### Reclaimed cost (reviewer number)

| | bytes |
|--|--|
| `basrArenaBytes` each | `bsrMaxStateChanges × 256` = **25,604,608** |
| pair `basr_values` + `basr_accounts` | **51,209,216** (~48.8 MiB) |
| `basr_records` table | `100018 × 24` = **2,400,432** |

**Runtime:** no longer materializes/fills that pair (or walks body rows) on the state-root path — **capacity-positive** under the 200M gas bound.

**BSS:** arena **reservations retained** this PR (layout shrink separate). `.bss` size unchanged; reclaim is cycles + dirty pages, not ELF bss end.

### Extents

| | BASE | FINAL |
|--|--|--|
| sha256 | `925947ae…` | `c3139a3e…` |
| text | 341652 (0x53694) | **341212 (0x534dc)** **−440** |

Four-artefact regen: TSV + GuestAddrs + LinkPins; GuestImageEntries NO-OP; second pin pass stable (.text/.data identical across pin rebuild).

### r200 seed=932419178 spike

| | BASE | CAND |
|--|--|--|
| selected/ran | 200/200 | 200/200 |
| FA/FR | 0/0 | 0/0 |
| full match | 200 | 200 |

`eest-ab-compare`: joined **198**; status/out diffs **0**; **base-PASS→cand-FR=∅**; **VERDICT: BYTE-IDENTICAL** on all 198.

### Files

- `EvmAsm/Codegen/Programs/BlockVerdictStateRoot.lean` (product)
- pins: `GuestAddrs.lean`, `RegionMapLinkPins.lean`, `symbol-addresses.tsv`

### Non-goals

- P1 fail-closed / derived builder finals
- P2 caps from rebuilt BAL
- BSS arena shrink
- Removing `teer_records_ptr` store (write-only hygiene; gas-gate offset churn)

No merge! from this agent.


pirapira:
Synced onto main and verified locally. Two ratchet artifacts needed to move with this PR's retirement — flagging them since I made the edits, not you:

1. `scripts/bal-class-a-baseline.tsv` — regenerated via `check-bal-class-a-ratchet.py --write-baseline`. Diff is **3 deletions, 0 additions** (`.Lbsr_bal_replay` x2, `block_state_root_pre_accounts`), i.e. exactly the edges your vestigial-walk removal retires.
2. `EXPECTED_CLASS_A_DEBT` in `scripts/check-bal-class-a-ratchet.py` — **12 -> 9**, matching those 3 removed paths. Its own comment specifies it "may only decrease, and a retirement plus this constant's decrease must land together", so it lands here.

Both move in the ratchet's permitted direction (shrink); neither relaxes the gate, and no statement or definition was touched. For the record, `check-bal-class-a-ratchet` failed on this branch's head **before** any merge with main, while main itself passed — so the baseline refresh was owed by the PR rather than being merge fallout.

Verified on the pushed head: `lake build` green (4730 jobs) and all 25 standing gates `ok`. Auto-merge armed.

Issue #11829, opened 2026-08-09
preserve the selfdestruct phase split: mid-tx destroyed-table semantics vs transaction-boundary tombstone


pirapira:
Successor to #11582, which is closed: nine of its ten entries were resolved, reshaped, or absorbed
elsewhere when re-verified on current main. **This is the one substantive survivor**, restated so it
stands alone.

## The phase split

**Spec** (pinned `e5a8caf1b`): `Evm.accounts_to_delete` is a **per-frame `Set`** (`vm/__init__.py:184`),
initialized empty (`vm/interpreter.py:135`, `:151`, `:349`) and merged from successful children
(`vm/__init__.py:234`). `SELFDESTRUCT` adds the originator **only when it was created in the same
transaction** (`vm/instructions/system.py:691-693`). **No account clear happens at the opcode** —
`process_transaction` applies `clear_account_preserving_balance` only at `fork.py:1201-1202`.

⭐ So the set is **deferred deletion provenance, not an eager "account absent" state.**

**Guest** (current main): `AccountWriteMap.lean:694-708` materializes `account_state_delete` at the
**transaction boundary** as an `account_writes` Present→None result. `AccountWriteMap.lean:813-836`
deliberately keeps `evm_selfdestruct_destroyed_table` **distinct**, for the mid-transaction
`EMPTY_CODE_HASH`/unavailable semantics. Three consumers depend on that table mid-transaction:
`EvmAccountWitness.lean:217-238` (same-tx created/selfdestructed scan),
`CreateSameTxCollision.lean:14-17` (keeps a CREATE/CREATE2 target unavailable during the transaction),
and `ChildFrameHandlers.lean:651-656` (avoids a second NEW_ACCOUNT charge while the address is still
alive until tx end).

## The obligation

**Preserve the phase split.**

- **Before tx finalization**: table membership must affect **only** same-tx read/collision/new-account
  semantics, and must **not** be converted into a Present→None post-state tombstone.
- **At the transaction boundary**: every deferred-delete entry must be materialized through the
  `account_writes` clear/tombstone path, after which later transactions read the finalized post-state.

**What breaks if violated:** removing or eagerly applying the table can return the wrong
`EXTCODEHASH`/availability, admit a same-tx CREATE collision, or mischarge NEW_ACCOUNT gas; failing to
apply the boundary path leaves deleted state visible in the next transaction.

[...]

PR #11831, opened 2026-08-09
docs: preserve selfdestruct phase split

pirapira:
## Summary\n\n- Make the SELFDESTRUCT phase split explicit at the AccountWriteMap, EvmAccountWitness, CreateSameTxCollision, and ChildFrameHandlers sites.\n- Before transaction finalization, evm_selfdestruct_destroyed_table is a same-transaction marker for read/EXTCODEHASH, CREATE/CREATE2 collision, and NEW_ACCOUNT semantics only; it is not a Present-None post-state tombstone.\n- At the transaction boundary, every deferred-delete entry must be materialized through account_writes_apply_deletes before the transaction-local sets are cleared.\n- Extract the transaction-boundary commit and tombstone-read cluster into AccountWriteMapDeletes.lean so the parent module stays under the file-size cap.\n- Cite the pinned Python authority: vm/__init__.py:184,234; vm/interpreter.py:135,151,349; vm/instructions/system.py:691-693; and fork.py:1201-1202. Each comment labels the Lean code as a mirror, not as authority.\n\n## Why this exists\n\nThe phase split is invisible in the type structure: nothing prevents a future change from collapsing the destroyed-address table into the account-writes Present-None representation, and the corpus would catch only some consequences. The comments name the concrete failure modes so the distinction remains actionable: collapsing early can make EXTCODEHASH or availability observe deletion too soon, admit a same-transaction CREATE collision, or mischarge NEW_ACCOUNT; skipping boundary materialization leaves deleted state visible to the next transaction.\n\nThe extraction is a defensible boundary cluster: account_writes_commit_pending and account_writes_is_absent are the transaction-finalization/tombstone-read surface, while the parent retains producers, overlays, builder emission, incorporation, and resolvers.\n\n## Verification\n\n- check-file-size.sh passes: AccountWriteMap.lean is 1433 lines and AccountWriteMapDeletes.lean is 92 lines.\n- check-unimported.sh passes: 2761 modules, 2761 reachable, 0 orphans.\n- LAKE_ARTIFACT_CACHE=false lake build EvmAsm.Codegen.Programs.AccountWriteMap passes.\n- Baseline and PR account-write assembly are byte-identical: SHA-256 f28b4ca4de6892c922278c4fac880efde87b5b6f239b7c2ed00f5b204b7cb7d7.\n- Baseline and PR stateless_guest ELF are byte-identical: SHA-256 40d3a3b9c52ae3ec8605e89ca3917252be2a5a4394b68a1c0eda4055c3a9a199. The emitted assembly and object also match: 52584a64242a9d77b43e8b6159b4b1898688752586732e6e0aab4f28dd072277 and 71857f677a93544cc983ff4055d2fa8ad550a0d47ea3a926601f3312492aa1ee.\n- No r200 run: this is a pure module split plus comments, and the full guest artifacts are byte-identical.\n\nCloses #11829


pirapira:
Parking this with `complicated-merge` — it needs a change here, and the fix isn't one I should pick for you.

`check-file-size` fails on **this PR's own head**, with no merge and no batch involved:

```
$ git checkout origin/codex2/11829-selfdestruct-phase
$ scripts/check-file-size.sh
  FAIL    1511 / 1500 lines  EvmAsm/Codegen/Programs/AccountWriteMap.lean  [codegen]
```

`origin/main` passes the same gate. `AccountWriteMap.lean` is 1492 lines on main — already 8 under the cap — and this PR's added lines push it to 1511.

The gate is explicit that per-file exception markers are not supported, so the only fix is splitting the file: extract a cluster of defs into a sibling submodule and import it back (it points at `Bls12MapG1Real.lean` / `ChildFrameHandlerTails.lean` as the pattern). Choosing *which* cluster to extract is a design call about this module, so I've left it to you rather than guessing at a split.

Worth noting the margin was thin before you started — main sits 8 lines under a 1500 cap — so this is partly bad luck about who arrived first, not a problem with the change itself.

#11830 was batched with this and is unaffected; I'm landing it on its own now. Once the gate is green on this branch head, drop `complicated-merge` and I'll pick it up.

Issue #11833, opened 2026-08-09
11797 M1: strip bal_section_info from block_state_root_pre_accounts and retire bv_fail 4


pirapira:
Move 1 of 6 under #11832 (#11797 Option A).

⭐ **This is the hard case, and it is answered: S1 cannot move.** `block_state_root_pre_accounts` runs before execution produces anything, and the rebuild needs MTx — so there is no ordering that puts a completed rebuild in front of it. It must be **restructured in place**, not relocated.

**The restructure:** strip `bal_section_info` from the site. The only remaining consumer of what it provides is `beqz bsr_bal_count` → `bv_fail 4`, which is **guest-invented** — the spec has no pre-body BAL-presence test — and the **sender gas path does not use `bsr_bal`**.

So M1 is: remove the read, retire code 4.

⛔ **Guard, non-negotiable:** removing a rejection means measuring where its rows go instead. Report the **replacement-code distribution** for the rows that currently die at 4. **If any row ACCEPTS, stop and escalate** — that would mean code 4 was load-bearing rather than guest-invented, and the plan changes.

⚠️ Note this dissolves `bv_fail 4` here rather than in #11826, which is correct: its later spec route is a *different object* (supplied count vs built hash), so it was never an ordering item.

pirapira:
## M1 measurement (report before PR)

### Change
- `block_state_root_pre_accounts`: strip `bal_section_info` + fail-102; keep pointer stash + witness-cap 101 + zero `bsr_bal_count`.
- `BlockVerdictFunction`: remove `beqz bsr_bal_count → .Lbv_no_bal_for_tx`.
- `BlockVerdictReceiptsTail`: retire sink code 4.

### ELFs (main `3cc00ecd3`)
| | sha256 | text |
|--|--|--|
| BASE | `621afcbe07b77b79…` | 340904 |
| CAND (measure) | `d6bea4318c5e8302…` | **340816** (−88) |
| FINAL (post four-artefact) | `0d1b85aed1ae04a4…` | 340816 |

data/bss unchanged. GuestImageEntries NO-OP. Four-artefact: TSV + GuestAddrs + LinkPins.

### Guard: replacement-code distribution (expected-invalid n=1036)

BASE fail hist: **4×30**, 5×948, 69×16, … ACCEPT=0  
CAND fail hist: **4×0**, **55×34** (+30), 5×948, 69×16, … ACCEPT=0  

**Former code-4 (n=30) → replacement:**
[...]

PR #11840, opened 2026-08-09
fix(bal): M1 strip pre_accounts BAL parse; retire bv_fail 4

pirapira:
## Summary
#11833 / #11797 Option A **Move 1 only**.

- Strip `bal_section_info` from `block_state_root_pre_accounts` (hard case: cannot move after rebuild).
- Retire guest-invented `bv_fail 4` (`.Lbv_no_bal_for_tx`); spec has no pre-body supplied-BAL presence test; sender gas path never used `bsr_bal_*`.

## What code 55 is (not BAL correspondence)

Former code-4 rows all land on **`.Lbv_requests_hash_fail` → bv_fail 55**. That sink is the EIP-7685 **execution `requests_hash`** compare (header.requests_hash vs computed requests — `fork.py` ~`requests_hash != block.header.requests_hash`), plus a few shared overflow/status branches that jump to the same label.

It is **not** the built-BAL hash check. Spec’s `block_access_list_hash` compare is a **different** object (`fork.py` ~`computed_block_access_list_hash != block.header.block_access_list_hash`); guest emits that as **bv_fail 60/61** (`.Lbv_bal_digest_mismatch` / rebuild fail), not 55.

So the 30-row relocation is an **incidental next gate** in the pipeline (system_contract_errors / invalid_layout cohort dies at requests-hash), **not** proof that M1 achieved supplied→built BAL correspondence for those rows. M1 is justified solely by code 4 being guest-invented; correspondence for BAL body readers is M3/M4’s question. This measurement shows **this move is safe**, not that the six-move restructure is complete.

## Measurement (guard)
expected-invalid n=1036 on main `3cc00ecd3`:
- Former **code 4 ×30 → code 55 ×30** (requests_hash, not BAL digest)
- **ACCEPT base-reject→cand-accept = 0**
- r200 seed 932419178 spike: full200 FA0 FR0 both; joined 198 byte-identical; **base-PASS→cand-FR=∅**

| | BASE | FINAL |
|--|--|--|
| sha | `621afcbe…` | `0d1b85ae…` |
| text | 340904 | 340816 (−88) |

Four-artefact regen (TSV/GuestAddrs/LinkPins; GuestImageEntries NO-OP).

## Test plan
- [x] expected-invalid replacement-code distribution
- [x] ACCEPT=0 guard
- [x] r200 base-PASS→cand-FR=∅
- STOP at M1 (no M2 in this PR)


pirapira:
Verified locally and armed. One ratchet fixup was needed — flagging it since I made the edits:

- `scripts/bal-class-a-baseline.tsv` — regenerated via `check-bal-class-a-ratchet.py --write-baseline`: **1 deletion, 0 additions**, the `block_state_root_pre_accounts -> bal_section_info` edge that your pre_accounts BAL parse strip retires.
- `EXPECTED_CLASS_A_DEBT` — **9 -> 8**, matching that one removed path. Its comment specifies it "may only decrease, and a retirement plus this constant's decrease must land together".

Both move in the shrink direction the ratchet permits; nothing was relaxed and no statement or definition was touched. Note this failed on your branch head with `behind == 0` (already up to date with main) while main itself passed, so it wasn't merge fallout — same fixup #11822 needed for the same reason.

Verified on the pushed head: `lake build` green (4728 jobs), all 25 standing gates `ok`, tree clean.

Issue #11834, opened 2026-08-09
11797 M2: derive the deferred deposit/exit enablement from header + exec_code_effect


pirapira:
Move 2 of 6 under #11832 (#11797 Option A).

**S3/S4 — the deferred builder deposit and exit fallbacks.** These are class (2): derivable from **header + `exec_code_effect`**, which are already the first two rungs of the ladder.

⭐ **These are the FA-ward pair from #11806.** Their accept-side consequences were closed by #11809 (empty system-contract code now rejects before dispatch) and #11819 (exec failure propagates), but the **reads themselves remain**, and they are control rather than value: a supplied non-empty code final steers past `.Ldsr_fail` into derive when `find_code_effect` misses.

Deriving the enablement removes the steer at its source rather than defending against it.

⚠️ This is what Option B would have done on its own. Under Option A it is step 2 rather than the whole change — the same work, kept because it is independently landable and closes the soundness-relevant reads early.

pirapira:
## M2 shipped — PR #TBD

Removed BAL `bal_find` + `nonstorage_finals` ladder on both builder deposit and exit same-block arms in `block_verdict_deferred_system_requests`. Miss on `find_code_effect_by_address` → `.Ldsr_fail`.

| check | result |
|--|--|
| ACCEPT on expected-invalid n=1036 | **0** (hist **identical** base/cand) |
| Landing code if path hit | **55** (`.Lbv_requests_hash_fail`) — incidental next gate; **not** BAL 60/61 |
| Corpus exercise of removed ladder | **none** (corpus-blind; justified by FA-ward #11806) |
| r200 seed 932419178 | FA0 FR0 byte-identical; base-PASS→cand-FR=∅ |
| text | 340816 → 340560 (−256) |

BASE sha `d6bea431…` FINAL `0fcbffec…` main base `43e7171e4`.

PR: will link after create. No M3 until release.

pirapira:
PR: https://github.com/Verified-zkEVM/evm-asm/pull/11841

pirapira:
Closed: #11841 merged as `32376a388`.

Verified on merged main rather than on the merge event: **`bal_find_account_by_address` no longer appears in `BlockVerdictStateRoot.lean` at all** (0 occurrences). The deferred builder deposit/exit arms no longer consult the supplied BAL; a miss goes to `.Ldsr_fail`.

[...]

Issue #11806, opened 2026-08-08
Supplied-BAL deferred-request fallbacks are a control dependency, not reject-only (FA-ward)


pirapira:
Split out of #11797 (classification) because it is an FA-ward finding, not a bookkeeping item.

## Claim

Two of the nine supplied-BAL consumers are **BRANCH control, not reject-only**: the deferred
builder-deposit and builder-exit fallbacks in `EvmAsm/Codegen/Programs/BlockVerdictStateRoot.lean`
(around `:845` and `:881`).

The shape at `:845-851`:

```
  jal ra, bal_find_account_by_address
  bnez a0, .Ldsr_fail
  ... bal_account_nonstorage_finals
  bnez a0, .Ldsr_fail
  la t0, bacc_finals; ld t1, 56(t0); beqz t1, .Ldsr_fail     # code final len
  la t0, bacc_finals; ld t1, 72(t0); beqz t1, .Ldsr_fail     # code final ptr
.Lc1_bd_code_ok:
```

Every guard here fires on **absence**. A non-empty code final in the *supplied* BAL therefore
**steers execution past `.Ldsr_fail` into the derive path** when `find_code_effect` misses.
The supplied value is not being checked against a derived one — it is deciding **which check runs**.

## Why the usual argument does not apply

The standing safety argument for supplied-BAL reads is "a wrong value yields a wrong state root,
so the block is rejected". That argument is about **value** dependencies. It says nothing about a
**control** dependency: a crafted BAL can steer the guest *away* from a check rather than into a
rejection, and nothing downstream necessarily diverges.

## Not FA-confirmed

This is a reachable control dependency, **not a demonstrated false accept**. What is missing is
whether the derive path it reaches can produce an accepted verdict that the spec would reject.
That is the next step, and it needs a constructed input, not a corpus sample — an FA of this shape
is invisible to r200 by construction.

## Classified in the same pass (context, no action here)

- `block_state_root` cursor walk `:333-371` — **vestigial**, no longer seeds `bsr_changes`
  (unconditional `j next`). Cheapest deletion; tracked in #11797.
- `pre_accounts` / root parse — **reject-only + scratch**.
- v2 `bal_section_info` — fail = reject; success **enables** the deferred path above.

⚠️ #11796 records that the Class-A ratchet reports OK over all nine of these sites. The gate cannot
see them, so its silence is not evidence either way here.

Ref: `origin/main`, classification by grok, read-only.

[...]

PR #11841, opened 2026-08-09
fix(bal): drop BAL code-final steer for builder deposit/exit (M2)

pirapira:
## Summary

#11834 / #11797 **M2** (S3/S4): remove the supplied-BAL non-empty code-final ladder that could steer `block_verdict_deferred_system_requests` past `InvalidBlock` when header + `exec_code_effect` expose no live builder deposit/exit code.

**Before:** after `find_code_effect_by_address` miss → `bal_find_account_by_address` + `bal_account_nonstorage_finals` → non-empty final → `.Lc1_bd/be_code_ok`.

**After:** miss → `.Ldsr_fail` (`a0=1` → `.Lbv_requests_hash_fail`).

Spec (e5a8caf1b `fork.py` checked system tx): code comes from live `TransactionState`, not a witness BAL declaration. Closes the FA-ward steer named in #11806.

## ELFs

| | BASE (main 43e7171e4) | FINAL |
|--|--|--|
| sha256 | `d6bea4318c5e8302…` | `0fcbffecc9f1c979…` |
| text | 340816 (0x53290) | **340560 (0x53250)** −256 |

Four-artefact regen: TSV + GuestAddrs + LinkPins; GuestImageEntries NO-OP; second pin pass ELF byte-identical.

### expected-invalid n=1036 (base = main after M1)

| | BASE | CAND |
|--|--|--|
| ACCEPT | **0** | **0** |
| fail hist | `{5:948, 55:34, 69:16, …}` | **identical** |

**No row changed fail code** — the removed BAL ladder is **corpus-blind** on honest fixtures (builder predeploys already in header). Landing code **if exercised** would be the existing deferred-fail path → `.Lbv_requests_hash_fail` (**55**), which is the same incidental next gate as M1’s former-4 cohort — **not** built-BAL 60/61. M2 is justified by FA-ward control removal (#11806), not by a measured replacement distribution on this corpus.

### r200 seed=932419178 spike

- BASE/CAND: selected=200 ran=200 full=200 FA=0 FR=0
- joined 198; status/output diffs 0
- **base-PASS→cand-FR = ∅**
- byte-identical on all joined cases

Artefacts: `/tmp/opencode/11834/`

## Scope / non-goals

- Does **not** start M3 (S5 c1_bal).
- Does **not** delete `c1_bal_*` locate (still used at v2 entry / section_info).
- P1 rebuild / hash placement unchanged.

## Test plan

- [x] expected-invalid ACCEPT=0 + identical hist
- [x] r200 FA0 FR0 base-PASS→cand-FR=∅
- [x] four-artefact + second pass NO-OP ELF


pirapira:
Verified locally and armed. Same ratchet fixup as #11822 and #11840 — flagging since I made the edits:

- `scripts/bal-class-a-baseline.tsv` — regenerated via `--write-baseline`: **2 deletions, 0 additions** (`.Lc1_bd_same_block` and `.Lc1_be_same_block` -> `bal_find_account_by_address`), exactly the c1_bal cursor edges your code-final steer removal retires.
- `EXPECTED_CLASS_A_DEBT` — **8 -> 6**, matching those two paths.

Both shrink, which is the only direction the ratchet permits; nothing relaxed, no statement or definition touched. Like #11840 this failed with `behind == 0` while main passed, so it is not merge fallout.

Worth raising with whoever is driving the M1/M2 BAL retirements: this is the third PR in a row (#11822, #11840, #11841) where a Class-A path is retired without shrinking `bal-class-a-baseline.tsv` and `EXPECTED_CLASS_A_DEBT` in the same commit. The gate can only be satisfied by touching both, so folding `check-bal-class-a-ratchet.py --write-baseline` plus the constant into the retirement commit would save a round-trip each time.

Issue #11839, opened 2026-08-09
ordering: relocate the blob-gas check (bv_fail 33) to after exact gas, matching fork.py post-body order


pirapira:
Split from #11826, whose original premise was refuted by measurement. **This is the one genuine ordering item that survived.**

## The inversion

**Spec post-body order** (`fork.py`): `gas_used` (62/35) → roots → `blob_gas` (33).

**Guest order**: the `bv_fail 5` funnel → **`33` before the body** → MTx `69` body admission → exact gas **unreached on 948/948**.

So `33` has a genuine blob counterpart in the spec, but sits **before** the body where the spec has it **after** `gas_used`. That is a real position divergence, unlike `bv_fail 5`, which turned out to be an early funnel with no spec counterpart at all rather than a shortcut for the computed gas compare.

⭐ Also established: **`69` has a body-time counterpart but NOT in the `gas_used` slot** — it was never a stand-in for that check, so no amount of moving it produces gas-used coverage.

## Sequencing

**After #11832.** The restructure reorders this same pipeline and M1 already retires `bv_fail 4`; relocating `33` before that settles risks doing and measuring the work twice.

⛔ **Removal/relocation guard, as everywhere in this pipeline:** report the replacement-code distribution for affected rows, and **stop if any row accepts**. A green corpus is exactly what would hide a lost check.

⚠️ Not in scope here: `bv_fail 13` has **zero corpus hits**, so its removal is unverifiable by measurement and rests on argument alone. It rides with #11832 or not at all.

pirapira:
## Scope: the comparison moves cleanly — **but its price side-effect does not**

Current site: `BlockVerdictFunction.lean:166-180`, after the code-27 NPR/versioned-hash matcher and
before public-key/chain/BAL admission and the MTx body.

**The comparison itself needs no new state after exact gas.** It reads `bv_versioned_hashes_len` and
`bv_exec_p+512` (`header.blob_gas_used`); the price call reads `bv_exec_p+520` (`excess_blob_gas`).
Both `bv_exec_p` and `bv_versioned_hashes_len` are populated before the body and **are not rewritten**,
and `bv_blob_gas_expected`/`observed` have **no later consumers**. So the verdict and its scratch can
move without carrying state forward.

⛔ **Hidden dependency, and it is the finding:** `amsterdam_blob_gas_price_u256` **also writes
[...]

Issue #11835, opened 2026-08-09
11797 M3: drop the c1 supplied-body read at the v2 section-info site


pirapira:
Move 3 of 6 under #11832 (#11797 Option A).

**S5 — the v2 `bal_section_info` read before the verdict.** Drop the `c1` body read.

Sequenced after M1 and M2 because those establish the pattern (strip the read, measure where the dependent rejection goes) on sites where the consumer is already understood.

⛔ Same guard as every move in this sequence: if a rejection is removed or relocated, report the replacement-code distribution and **stop if any row accepts**.

pirapira:
## M3 shipped — https://github.com/Verified-zkEVM/evm-asm/pull/11842

Dropped v2 `bal_section_info` → `c1_bal_*` fill. Zero consumers after M2.

| check | result |
|--|--|
| ACCEPT n=1036 | **0** hist **identical** |
| fail-code moves | **none** (corpus-blind) |
| code 60/61 | 60×1 both legs unchanged |
| r200 | FA0 FR0 byte-id; base-PASS→cand-FR=∅ |
| text | 340560→340524 (−36) |
| **#11839 blob-gas 33** | **not relocated** — still pre-body in BlockVerdictFunction |

BASE `0fcbffec` FINAL `d3f9aaa7` main `32376a388`. STOP before M4.

pirapira:
Closed: #11842 merged as `7ad207769`.

Verified on merged main: **`c1_bal_start`/`c1_bal_len` have zero occurrences** in `BlockVerdictStateRoot.lean`. The v2 `bal_section_info` `c1` fill is gone — it had **zero consumers after M2**, which is what made this move cheap.

Measured on `32376a388`: expected-invalid n=1036 **ACCEPT = 0**, fail histogram **identical**, the single **code-60** row unchanged on both legs, r200 FA0/FR0 byte-identical with `base-PASS → cand-FR = ∅`, `.text` 340,560 → 340,524 (−36), four-artifact regen NO-OP with a stable second pass.

⚠️ Claim kept to its measurement: the removed parse was **corpus-blind**, and the PR names what would have happened had it been exercised — next hit would be `block_access_list_hash` or later **60/61**, not a new rejection. Three of six moves done; the supplied-body readers are not yet gone.

PR #11842, opened 2026-08-09
fix(bal): drop v2 c1 bal_section_info fill (M3)

pirapira:
## Summary

#11835 / #11797 **M3** (S5): drop the v2 entry `bal_section_info` that filled `c1_bal_{start,len,count}`.

After M2 those cells had **zero consumers**. Header `block_access_list_hash` still hashes the SSZ BAL body; post-body rebuild compare remains **bv_fail 60/61**.

## ELFs (base = main `32376a388` with M1+M2)

| | BASE | FINAL |
|--|--|--|
| sha256 | `0fcbffec…` (= M2 final) | `d3f9aaa7…` |
| text | 340560 (0x53250) | **340524 (0x5322c)** −36 |

Four-artefact: TSV + GuestAddrs + LinkPins; GuestImageEntries NO-OP; second pin pass ELF byte-identical.

### expected-invalid n=1036

| | BASE | CAND |
|--|--|--|
| ACCEPT | **0** | **0** |
| fail hist | `{5:948, 55:34, 69:16, … 60:1, …}` | **identical** |

**No fail-code relocation.** Removed early parse was corpus-blind on honest fixtures (same pattern as M2). Landing **if** a malformed BAL section previously died only at this fill: would next hit `block_access_list_hash` (v2 bal-hash fail) or later **60/61** rebuild — none of the 1036 rows moved.

Real BAL authority on this corpus remains the single code-**60** row (unchanged both legs).

### r200 seed=932419178 spike

- full200 FA0 FR0 both; joined 198 byte-identical; **base-PASS→cand-FR=∅**

## #11839 answer (blob-gas / code 33)

**M3 does not relocate bv_fail 33.** Blob-gas used check stays **pre-body** in `BlockVerdictFunction.lean` (after versioned hashes, before public-keys / MTx). Shared tail order unchanged. #11839 remains a **separate change**, not a verification of this PR.

## Scope

- Does not start M4 (S2 `block_state_root` supplied count/cap).
- Does not delete dead `c1_bal_*` BSS (M6).
- STOP before M4 until merge + base confirm.

Artefacts: `/tmp/opencode/11835/`


pirapira:
Verified locally and armed. Same ratchet fixup as #11822 / #11840 / #11841 — flagging since I made the edits:

- `scripts/bal-class-a-baseline.tsv` — regenerated via `--write-baseline`: **1 deletion, 0 additions** (`.Lv2_input_hash -> bal_section_info`), exactly the edge your v2 c1 `bal_section_info` fill removal retires.
- `EXPECTED_CLASS_A_DEBT` — **6 -> 5**, matching that one path.

Shrink only, nothing relaxed, no statement or definition touched. Failed with `behind == 0` while main passed, so not merge fallout.

This is the **fourth** M-series PR in a row needing this (#11822, #11840, #11841, #11842). Adding `python3 scripts/check-bal-class-a-ratchet.py --write-baseline` plus the matching `EXPECTED_CLASS_A_DEBT` decrement to the retirement commit itself would make these land first time.

Issue #11832, opened 2026-08-09
restructure (11797 Option A): run the supplied-BAL readers after the rebuild — ordered, independently landable moves


pirapira:
Tracks the **Option A** decision on #11797: move verdict work so the supplied-BAL readers run **after**
the rebuilt BAL exists, removing the divergence rather than deriving around it.

**Maintainer's reason, recorded because it governs the shape of the work:** *cheaper than reasoning
about different structures later.* The cost avoided is **proof cost paid repeatedly** — every place the
guest's control flow diverges from `fork.py` becomes a case that cannot be discharged by
correspondence and must be argued in every theorem that crosses it.

## What is already settled (do not re-derive)

- **#11823** — the rebuilt BAL is **hash-exact on every complete valid input** (26,074/26,104). The 30
  mismatches are three shapes, each requiring an already-invalid block or an already-incomplete witness.
- **#11824** — **class (1) is empty**: nothing can be repointed, because the rebuild exists only at P5
  while every residual reader sits at P0–P4. And the rebuild **cannot move earlier than end-of-P3** — it
  needs `bal_builder_*` filled at each user-tx end, the N+1 withdrawals, and the P3 deferred system
  write-sets, then `incorporate_touched` + sort + hash.
- **#11825** — capacity is not a constraint: peak both-resident **23.3 MB** against **38–61 MB** headroom,
  both already separate BSS reservations, so the migration needs **no new allocation**.

⭐ **Therefore the target is "readers later", not "rebuild earlier".**

## Shape of the work

⛔ **Sequenced, not big-bang.** The first deliverable is an **ordered plan of independently landable
moves**, each with its own measurement, so a failure is attributable to one move rather than to a
restructure. Per-move issues will be filed under this one as the plan lands.

⚠️ **Known hard case, to be answered first:** `block_state_root_pre_accounts` (`:88-92`) runs **before
execution produces anything**. It is neither a repoint nor a derivation. If it cannot move, the plan is
"N moves plus one genuinely different problem", and that needs saying in the first report rather than
the last.

## Gate

The **structural bar**, per the standing rule that a bigger change lands earlier on a lighter gate:
**structure matches the spec, and r200 does not deteriorate a lot.** Explicitly **not** byte-equivalence
— moving verdict work will move rows, and demanding byte-identity would force the change to be done in
a way that defeats its purpose.

pirapira:
## The six moves, enumerated

[...]

Issue #11824, opened 2026-08-08
BAL step 2/3: per-site ordering and availability for the nine supplied-body readers


pirapira:
Step 2 of 3 toward #11797's exit condition: **the guest stops reading the supplied BAL body entirely.**

Depends on #11823 — until the rebuilt BAL can stand in for the supplied bytes, there is nothing to
migrate readers onto.

## The problem

The rebuilt BAL is only complete **after execution**. The nine supplied-body readers do not all run
after execution, so "read the rebuilt one instead" is not uniformly available.

- **Post-exec sites are fine in principle.** The deferred builder deposit/exit fallbacks
  (`BlockVerdictStateRoot.lean` ~`:845`, ~`:881`) run after execution, so a completed rebuild exists.
- ⛔ **`block_state_root_pre_accounts` (`:88-92`) is the hard case.** It reads `bsr_bal_start` /
  `bsr_bal_len` **before execution has produced anything.** There is no rebuilt structure to consult at
  that point, so this site cannot simply be repointed.

## The task

Per-site, for each of the nine reads: state **when** it runs relative to execution, and which of these
applies:

1. a completed rebuild is available → repoint it,
2. the value is derivable from something else already computed (header, `exec_code_effect`) → derive it,
3. neither → the site needs restructuring, or the work runs at a different point.

⭐ **This analysis decides whether the exit condition is tractable at all.** If most sites are (1) or
(2) it is plumbing. If several are (3), the change is a restructure of when verdict work happens, which
is a different size of undertaking and needs saying plainly before anyone starts.

## Already known, do not re-derive

- Root values are **already map-authoritative** — no work needed there.
- The vestigial cursor walk and the `basr` fill are **gone** (#11822).
- `aps_witness_ptr`/`aps_witness_len` setup is **NOT** vestigial — `account_apply_storage` reads it.
  Deleting it flips the success bit while the state root still matches (#11822).

Analysis only. No code changes in this issue.

pirapira:
## #11824 per-site ordering and availability (analysis only)

**Tree:** `origin/main` **ee59739a7** (#11822 merged — vestigial walk + `basr` fill gone).  
**Not re-derived:** roots map-authoritative; `aps_witness_*` kept; hash-only paths compliant.

### Execution timeline (where rebuild lands)

| phase | when | BAL-related |
|--|--|--|
[...]

PR #11843, opened 2026-08-09
fix(bal): drop post-root supplied BAL count/cap (M4 #11836)

### LINKED ISSUE #11836: 11797 M4: drop the supplied count/capacity read at the post-root section-info site
Move 4 of 6 under #11832 (#11797 Option A).

**S2 — post-root `section_info`/cap.** Drop the supplied count and capacity read.

⭐ S2 is the one site that a **rebuild relocation** would help: it is the only reader that end-of-P3 placement reaches (per #11824). So M4 and M5 interact — if M5 lands first, M4 may become a repoint rather than a drop.

Order is M4 then M5 unless measurement says otherwise; deciding that is part of M4's report.

pirapira:
## Summary

#11797 Option A / **M4** (#11836): drop supplied BAL `bal_section_info` + count/gas/cap gates from `block_state_root` (S2).

- Spec builds+hashes BAL after `apply_body`; map is sole root authority post-#11431.
- Always wire `aps_witness_ptr/len` for `account_apply_storage` (was nested under count≠0).
- Retire dead `.Lbsr_cons_bal_section` / bsr_fail **110**.
- Keep zeroing `bsr_bal_count` for dump surface only.

### M5 implication

S2 was the **only** reader that end-of-P3 rebuild would help (#11824). Dropping the supplied read outright → **M5 loses its only beneficiary → retire M5** (good outcome). No rebuild move needed for S2.

### #11839

M4 does **not** relocate bv_fail **33** (blob-gas). Still pre-body in `BlockVerdictFunction`. Untouched.

## Measurement (base = main 7ad207769 / M3)

| | BASE | CAND |
|--|--|--|
| ELF sha | `d3f9aaa7…` | `4cd03f27…` |
| text | 340524 (0x5322c) | **340356 (0x53184)** −168 |

### expected-invalid n=1036

- BASE/CAND hist **identical**: `{5:948, 55:34, 69:16, 0:8, 1:6, 7:5, … 3:2, …}`
- Pairwise fail-code transitions: identity only
- **ACCEPT=0**
- Landing code if hit: **none** (no row was dying at the retired gate; corpus-blind like M2/M3)
- Former code3×2 stay code3 (unrelated state_fail)

### r200 seed=932419178 spike

- BASE/CAND: selected200 ran200 full200 FA0 FR0
- joined 198; status/out diffs 0; **byte-identical**
- **base-PASS→cand-FR=∅**

Four-artefact: TSV + GuestAddrs + LinkPins; GuestImageEntries NO-OP; second pin pass stable at textSizeBytes=0x53184.

## Test plan

- [x] expected-invalid ACCEPT=0 + identity hist
- [x] r200 FA0 FR0 base-PASS→cand-FR=∅
- [x] M5 retire recommendation
- [x] #11839 no blob-gas move

Closes #11836. Part of #11797 / #11832.


pirapira:
Verified locally and armed. Same ratchet fixup as #11822 / #11840 / #11841 / #11842:

- `scripts/bal-class-a-baseline.tsv` — regenerated via `--write-baseline`: **1 deletion, 0 additions** (`.Lbsr_oao_2935_done -> bal_section_info`), the edge your post-root count/cap removal retires.
- `EXPECTED_CLASS_A_DEBT` — **5 -> 4**.

Shrink only; nothing relaxed, no statement or definition touched. `behind == 0` with main passing, so not merge fallout.

Fifth in the series now (#11822, #11840, #11841, #11842, #11843) — the M-series is steadily walking the Class-A debt down (12 -> 4), which is exactly the intent, but each PR needs the two bookkeeping files moved with it. Folding `--write-baseline` plus the `EXPECTED_CLASS_A_DEBT` decrement into the retirement commit would let these land without a round-trip.

PR #11883, opened 2026-08-09
ci: report parallel gate coverage and skips

### LINKED ISSUE #10959: check-build-parallel cannot report its own coverage: 4 of 9 fanned-out scripts exit 0 on a missing toolchain, and only 1 prints a set size

pirapira:
Fixes #10959.

The parallel post-build wrapper now records every child gate start, reports each lane's completed/expected step count, and counts skip messages. A lane with a zero-exit skip is reported as SKIP and makes the aggregate fail; a lane that stops before its declared set is reported as INCOMPLETE. This prevents a green aggregate from being mistaken for full coverage.

Verification:
- bash -n scripts/check-build-parallel.sh
- git diff --check

Scripts-only; no regeneration and no r200.

### DIFF
--- a/scripts/check-build-parallel.sh
+++ b/scripts/check-build-parallel.sh
@@ -57,36 +57,43 @@ report_checks() {
 }
 
 start codegen codegen_checks
-start guestaddrs-starts scripts/check-guestaddrs-starts.sh
-start asm-to-program scripts/check-asm-to-program.sh
+start guestaddrs-starts run_step scripts/check-guestaddrs-starts.sh
+start asm-to-program run_step scripts/check-asm-to-program.sh
 start reports report_checks
-start axioms scripts/check-axioms.sh
-start arithmetic-fuzz scripts/fuzz-arith-diff.sh
+start axioms run_step scripts/check-axioms.sh
+start arithmetic-fuzz run_step scripts/fuzz-arith-diff.sh
 
 status=0
 for i in "${!pids[@]}"; do
   name="${names[$i]}"
   lane_status=PASS
+  child_failed=0
   if ! wait "${pids[$i]}"; then
-    lane_status=FAIL
+    child_failed=1
     status=1
   fi
 
   log="$work/$name.log"
   steps="$(grep -c '^CHECK_BUILD_PARALLEL_STEP ' "$log" || true)"
-  skips="$(grep -Eic '(^|[^[:alpha:]])(skip|skipping)([^[:alpha:]]|$)' "$log" || true)"
+  # These are the deliberate, machine-readable skip lines emitted by the
+  # current child gates. Do not grep for the word "skip" anywhere: region-map
+  # also has prose about skipped sub-checks, and an unrelated filename or
+  # informational sentence must not turn a passing lane into a false failure.
+  skips="$(grep -Eic '^check-build-units-link: SKIP|^check-(guarded-handler-bytes|asm-to-program): .*skipping \(install to enable\)|^[[:space:]]+SKIP emitted-reality|^[[:space:]]+skip Class-A BAL ratchet' "$log" || true)"
   expected="${expected_steps[$name]}"
 
-  if [[ "$lane_status" == PASS && "$skips" -gt 0 ]]; then
+  if [[ "$steps" -ne "$expected" ]]; then
+    # set -e can stop a lane before its later children start. Report that loss
+    # of coverage explicitly even when the child that stopped it failed.
+    lane_status=INCOMPLETE
+    status=1
+  elif [[ "$child_failed" -eq 1 ]]; then
+    lane_status=FAIL
+  elif [[ "$skips" -gt 0 ]]; then
     # A child gate may intentionally return zero after deciding it cannot run.
     # That is not evidence for a green aggregate: make the third state visible
     # and fail the wrapper so CI cannot mistake it for a completed lane.
     lane_status=SKIP
-    status=1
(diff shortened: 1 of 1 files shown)


pirapira:
Follow-up verification after review:

I ran the wrapper against a temporary stub checkout with all six lanes:

- clean stubs: exit 0; codegen PASS (5/5), guestaddrs PASS (1/1), asm-to-program PASS (1/1), reports PASS (3/3), axioms PASS (1/1), arithmetic-fuzz PASS (1/1).
- one stub emitted `check-build-units-link: SKIP ...`: exit 1; codegen SKIP (5/5, skips=1), all other lanes PASS.
- one stub failed `check-region-map` before later codegen steps: exit 1; codegen INCOMPLETE (3/5, skips=0), all other lanes PASS.

The direct lanes are now wrapped in `run_step`, so their 1/1 coverage is measured too. The skip detector is anchored to the known machine-readable skip forms: check-build-units-link, check-guarded-handler-bytes, check-asm-to-program, plus check-region-map's two explicit partial-check lines (`SKIP emitted-reality` and `skip Class-A BAL ratchet`). It does not grep arbitrary occurrences of the word skip, avoiding filenames/prose false positives. `bash -n` and `git diff --check` pass.

Issue #11795, opened 2026-08-08
proof: discharge RlpItemDecodeBridges and move rlp_list_count_items to .bridged (residual of #11767)


pirapira:
> 🤖 Issue generated by Claude Code.

Successor to #11767, which is closed: **its alignment ask landed in PR #11776**, but the proof obligation it was filed to unblock is still open — and ⭐ **for a different reason than the one #11767 named.**

## What #11776 fixed

`rlp_walk_next` now recursively validates list payloads against the pinned `ethereum-rlp` contract: the verified single-item body is emitted unchanged as `rlp_walk_next_core` (`RlpWalk.lean:151`), and the wrapper requires exact cursor-to-end exhaustion, rejecting malformed / truncated / non-canonical / nested-malformed / trailing interiors with **status 7**, never bailing or fabricating empty. Two bounds, stated separately:

- **fuel** — input-derived `2 * (end - cursor)`, matching `Decode.lean`; #11786 later showed this arm is **unreachable from top-level `decode` by construction**;
- **depth** — `s1` active-list cap at **1024**, derived from the reference's own CPython `RecursionError` boundary; 1024 × 96 B = 96 KiB inside the 192 KiB guest stack.

⇒ The **span-only vs decodes-the-payload** divergence that #11767 described **no longer exists**.

## ⛔ What still blocks the bridge, and it is NOT that divergence

`RlpItemDecodeBridges` (`RlpListCountItemsBridge.lean:76`) remains an explicit hypothesis (`:46`: *"stated as an explicit hypothesis and is NOT yet discharged"*).

⚠️ **Do not assume #11776 made it provable.** The tree records a second, independent obstruction at `:183-196`:

> `RlpItemDecodeBridges` — quantified over all five disjuncts — is **UNSATISFIABLE** on this routine's actual domain, because since #11675 `rlp_list_count_items` is reached from `mpt_node_kind` and MPT branch children include **inline embedded nodes**, i.e. nested lists (`SpecRef/IncrementalMpt.lean:155`, `resolveChildRefAux` has an explicit `| .list items =>` arm). A nested list's `decodeAux` is **fuel-sensitive**, so the `∀ m` link `DecodeChain` demands is false there.

So the blocker moved: **not "the two sides disagree on list interiors" (fixed), but "`DecodeChain` demands fuel-insensitivity that nested lists cannot provide"** — which is **#11711**, open and labelled `proof`.

[...]

Issue #11678, opened 2026-08-07
gate: witness EL.RLP.decode_encode in check-axioms — proven sorry-free but absent from Progress/


pirapira:
Split out of #11661, which is closed. Everything else that issue asked for either already exists or belongs to another issue; this is the residue.

## The gap

`EL/RLP/Properties.lean:2548`'s `decode_encode` is **proven and `sorry`-free** — `[propext, Classical.choice, Quot.sound]` only — but it is **not in the axiom gate**:

- `#print axioms EvmAsm.EL.RLP.decode_encode` does not appear in `Progress/AxiomWitnesses.lean`. The only `EL.RLP` witness there is `Nat.fromBytesBE_eq_zero_iff` (`:99`).
- `decode_encode` appears **nowhere** under `Progress/` — no witness, no registry row.

So `scripts/check-axioms.sh` does not cover the RLP round-trip today, and no registry row claims it. A later edit could introduce a `sorryAx` or a TCB-expanding tactic into that proof or its dependency chain and no gate would object.

## Why this is worth its own issue rather than a note

This is the *"both halves exist, the join does not"* shape, one level up from where #11637's gate reaches. That gate closed row-existence for **routine-level** specs — a linked routine carrying a spec theorem with no registry row now fails the build. **Pure-spec lemmas like this one are outside its scope**, which is why the gap survived and why finding it needed a per-issue human check.

Four issues in the same batch had the routine-level version of this — #11348 (`bloom_or_into`), both halves of #11422, and #11516 (`decoded_matches_specRef` proven and witnessed, only the row missing).

## Scope

1. Add `#print axioms EvmAsm.EL.RLP.decode_encode` to `Progress/AxiomWitnesses.lean`, plus the `private noncomputable abbrev _<name>_witness := @…` form if a registry row is also wanted — per #11671's PLAN.md note, **referencing a theorem does not put it in the axiom gate; the abbrev is what does**. They are separate obligations.
2. Consider the same for `decodeFully_encode_of_decode_encode` (`FullDecode.lean:106`) and `decode_encode_mutual` (`Properties.lean:2423`), the lift and the general form it specialises. A witness is always warranted for a named theorem; a *row* claims a tier and only if earned.
3. If a row is added, grade it honestly. The theorem carries a hypothesis — `(encode item).length < 256 ^ 8` — whose docstring notes it implies the same bound for every nested payload, so `.domainRestricted` may be the accurate grade rather than `.agrees`.

[...]

PR #11894, opened 2026-08-10
proof(rlp): fuel monotonicity for decodeAux/decodeItems, unblocking nested-list chains

### LINKED ISSUE #11711: proof: DecodeChain demands fuel-insensitivity, which blocks every nested-list bridge (rlp_list_count_items is the live case)

dhsorens:
Fixes #11711.

`DecodeChain` states each link **fuel-insensitively** — `∀ m, decodeAux (m + 1) … = some …`. That is provable for byte-string items and *false* for a nested list, whose decode recurses into `decodeItems nDepth`. Per the issue this is urgent rather than latent: #11675 put `rlp_list_count_items` on the `mpt_node_kind` path, where an inline embedded branch child **is** a nested list, so it is the normal shape.

## The lemma

New `EvmAsm/EL/RLP/FuelMono.lean`:

```lean
decodeAux_mono_fuel   : n ≤ m → decodeAux   n bs = some r → decodeAux   m bs = some r
decodeItems_mono_fuel : n ≤ m → decodeItems n bs = some r → decodeItems m bs = some r
```

Extra budget never changes a *successful* decode. `decodeAux` and `decodeItems` are mutual and **both** step the budget down by exactly one, so one induction on the budget carries both statements. The four byte-string branches never mention `nDepth` and are discharged once by `decodeAux_succ_eq_of_lt_c0` (branch bodies literally equal at two budgets); only the two list branches use the IH. `takeBytes`/`readLength` take no fuel.

Both report **`[propext, Quot.sound]`** — read back from the build log, not even `Classical.choice`.

## Why this shape rather than the issue's sketch

The issue sketched threading a fuel budget through the links, and its own warning said *"the fuel bookkeeping is the whole difficulty"*. Monotonicity **removes** that difficulty instead of managing it: a link need only be exhibited at **one** budget, and every larger budget follows.

So `DecodeChainFrom` carries a single `floor` and its links are plain `decodeAux floor … = some …`, with **no per-link fuel algebra**:

```lean
def DecodeChainFrom (bytes : List Byte) (floor : Nat) : Nat → List RLPItem → Nat → Prop
  | off, [], offEnd => off = offEnd
  | off, item :: rest, offEnd =>
      ∃ off', decodeAux floor (bytes.drop off) = some (item, bytes.drop off')
        ∧ DecodeChainFrom bytes floor off' rest offEnd
```

`decodeItems_of_chainFrom` needs exactly one side condition, `floor ≤ k + 1` — the composition consumes link `i` at budget `k + (items.length - i)`, whose minimum over the chain is `k + 1` at the last item, so one inequality covers every link.

## `DecodeChain` is not weakened

The issue is explicit: ⛔ don't relax the existing predicate into silently accepting lists. So `DecodeChainFrom` is **new and strictly more general**, and I prove that rather than assert it:

- **`decodeChainFrom_of_decodeChain`** — every `DecodeChain` is a `DecodeChainFrom` at `floor = 1` (instantiate `∀ m` at `m := 0`). Nothing the old predicate accepted is lost.
- **An `example` re-derives `decodeItems_of_chain`'s exact statement** from the new composition, where the side condition `1 ≤ k + 1` is free. This is the check that the generalisation is faithful rather than one restriction traded for another — same role as the existing arity-four non-vacuity `example`.
- **A second `example` exhibits a `.list` item as a legal link** — the shape `DecodeChain` cannot express — which is the statement `rlp_list_count_items`' bridge needs.

## Scope, honestly

Model side only. Discharging the `decodeAux floor` hypothesis for the guest's *actual* node bytes and regrading `rlp_list_count_items` `.machineOnly → .bridged` is routine-side follow-on (#11341's fourth row, #11795). The model-side blocker this issue is about is closed; I did not touch the four `.machineOnly` rows, per the issue's "not scoped as finish #11341".

Both lemmas are **witnessed in the axiom gate** (147 witnesses, up from 145). An unwitnessed load-bearing lemma is precisely the #11678 hole, and it would be inconsistent to file that and then add one. That needed a direct `import EvmAsm.EL.RLP.FuelMono` in `Progress.lean` — core → core, no Codegen, so `Routines.lean`'s rebuild-cost note is unaffected.

## Two mechanics traps, recorded in PLAN.md

Worth flagging since they cost real iterations and will recur:

1. **`cases h : e` generalises `e` in the GOAL only.** A pre-existing hypothesis keeps the unsubstituted term and needs an explicit `rw … at h`; adding `⊢` to that `rw` fails as a no-op and *looks* like a numeral-normalisation mismatch. The asymmetry is load-bearing here — `cases` on the fuel-free `takeBytes`/`readLength` hits the goal, while `cases hdi : decodeItems n payload` leaves the goal's `decodeItems m' payload` intact, and that surviving occurrence is exactly what the IH rewrites.
2. **RLP `do`-blocks desugar to `Bind.bind`,** so `Option.bind_some` alone does not fire; `simp only [Option.bind_eq_bind, Option.bind_some]` is the working set. The `unusedSimpArgs` linter then flags `bind_eq_bind` wherever the term is already `Option.bind` — four such args removed to keep the build warning-free.

## Gates

`lake build` (4727 jobs, green, **0 warnings**) + check-no-warnings, check-progress, check-drift, check-forbidden-tactics, check-unimported, check-file-size, check-layering, check-heartbeats-approved, check-statement-tamper, check-registry-crosscheck, check-correspondence-deps — **all pass**. No `sorry`/`admit`, no `native_decide`/`bv_decide`, no `maxHeartbeats`.

`check-axioms.sh` **cannot run in this checkout** and I am not claiming it: it exits 2 because `scripts/lib/worktree-build-lock.sh` requires `flock` (absent on macOS) plus GNU `sha256sum`. CI is the authority; both axiom reports are quoted above from the build log.


pirapira:
Synced onto main and verified locally. One conflict needed a hand resolution — flagging it since I made the edit:

`EvmAsm/Progress.lean` conflicted because this PR appends two witness abbrevs (`_decodeAux_mono_fuel_witness`, `_decodeItems_mono_fuel_witness`) at the same point where main gained #11678's three RLP round-trip witnesses via #11891, which landed a few minutes ago.

Both sides are independent additive `private noncomputable abbrev` blocks, so I **kept both** rather than choosing a side. I resolved the hunk by hand rather than with `git checkout --ours/--theirs`, which resolve per *file* and would have silently discarded one side's witnesses entirely. Then asserted all five abbrevs are present by name before committing:

```
_decodeAux_mono_fuel_witness             1
_decodeItems_mono_fuel_witness           1
_rlp_decode_encode_witness               1
_rlp_decode_encode_mutual_witness        1
_rlp_decodeFully_encode_witness          1
```

No statement, definition, or proof body was changed — only the merge of two append blocks. `gen-axiom-witnesses.py` went 151 -> 153, matching your two new witnesses exactly.

Verified on the pushed head: `lake build` green (4727 jobs, twice across the regen fixpoint) and all 25 runnable standing gates `ok` by exit status. Auto-merge armed.

dhsorens:
Thanks — and **verified your resolution on the pushed head**, independently rather than taking it on trust:

```
$ git show origin/work/11711-decodeaux-fuel-mono:EvmAsm/Progress.lean | grep -n "_witness :="
688: _decodeAux_mono_fuel_witness            # this PR (#11711)
690: _decodeItems_mono_fuel_witness          # this PR (#11711)
701: _rlp_decode_encode_witness              # #11678 / #11891
706: _rlp_decode_encode_mutual_witness       # #11678 / #11891
714: _rlp_decodeFully_encode_witness         # #11678 / #11891
720: _rlp_encode_injective_witness           # #11678 / #11891
```

All six present, and no statement or proof body touched. Your `--ours/--theirs` point is exactly right and worth keeping on the record: those resolve per **file**, so on this hunk either flag would have silently dropped one PR's witnesses entirely — and the failure would have been invisible, because a dropped `abbrev` does not break the build. It just quietly removes theorems from the axiom gate, which is the #11678 defect class reappearing as a merge artifact rather than an omission.

Both PRs were mine and landed minutes apart, so the collision was my sequencing, not bad luck. Noted for the rest of this sweep: I have several branches that each append to `Progress.lean`'s witness block, so this hunk will keep conflicting. I'll keep them serialised rather than pushing them concurrently.

Two follow-ups from this PR, both filed:

- **#11898** — the scoped enumeration you asked for on #11895 before anyone touches `rlpItemDecode`: **52 consuming modules**, tiered by risk, with the two *determinism* modules (`RlpWalkDeterminism`, `WalkItemDeterminism`) called out as read-first, since the relation being span-only may be load-bearing for those arguments. It also proposes an **additive** `rlpItemDecodeStrict` + implication lemma instead of an in-place edit, which would unblock #11795 with **zero** changes to those 52 modules.
- **#11896** — `encode_decode` (the canonicality direction) is missing, which is the real prerequisite for #11692's `.machineOnly → .bridged` upgrade. `encode_injective` is not a substitute; details on the issue.

PR #11895, opened 2026-08-10
proof(rlp): fuel-sensitive per-item bridge, and name the residual list-interior gap

dhsorens:
Addresses #11795. **Stacked on #11894** (`work/11711-decodeaux-fuel-mono`) — based against that branch, so review #11894 first; I'll retarget to `main` once it lands.

## #11795's sequencing claim is wrong, and this PR shows why

The issue says that once #11711 removes `DecodeChain`'s `∀ m` requirement, discharging the per-item bridge *"becomes assembling the existing pieces — not new mathematics"*. Fuel was **one of two** independent obstructions. #11711 closed that one. The second is a strength mismatch no amount of fuel work touches:

> `rlpItemDecode`'s two **list** disjuncts (`WalkNext.lean:3649`, the `0xc0`–`0xf7` and `≥ 0xf8` arms) require header canonicality plus a **span fit** — `¬ ult (endPtr - cursor) span` — and say nothing about the payload. `decodeAux` (`Decode.lean:63`) additionally runs `decodeItems nDepth payload` and returns `none` unless the payload is consumed exactly.

So a list item whose span fits but whose **interior is malformed** satisfies the machine relation and is rejected by the model — *at every budget*. `RlpItemDecodeBridges` therefore stays unsatisfiable on the unrestricted domain for a reason independent of fuel.

Worth noting the file's own comment at `:183-196` attributed the unsatisfiability solely to fuel; that was incomplete, and #11795 inherited the gap.

## What this delivers — the half #11711 genuinely unblocks

| Declaration | What it does |
|---|---|
| `RlpItemDecodeBridgesFrom` | per-item obligation in fuel-sensitive form: model decode exhibited at one budget `floor`, not at every budget. A nested list is expressible here; under `∀ m` it is not. |
| `rlpItemDecodeBridgesFrom_of_bridges` | the old obligation implies the new one at any positive budget — **faithfulness**: the new form asks for no more than the original |
| `DecodeChainFrom.snoc` | the `DecodeChain.snoc` analogue, for the same snoc-vs-cons reversal (`StrictPrefix.succ` extends at the back, the chain recurses from the front) |
| `StrictPrefix.decodeChainFrom_of_itemBridge` | the structural bridge over the fuel-sensitive predicate — same shape as the existing one, but its per-item hypothesis is now satisfiable-in-principle rather than false on the live domain |

## What remains — named, not described

- **`RlpListInteriorsDecode`** — the residual obligation, with the list restriction (`¬ ult b 0xc0`) **visible in the predicate**, following the discipline `RlpItemDecodeBridgesBytes` already set with its `hbytes` guard. No caller can mistake the structural bridge for a complete one.
- **`rlpItemDecodeBridgesFrom_of_parts`** — byte-string items and list items exhaust the prefix space, so the byte-string half plus `RlpListInteriorsDecode` yields the full obligation. This is the load-bearing honesty check: it proves the residual is the **only** thing left, and in particular that it hides no further fuel trouble.

**Discharging `RlpListInteriorsDecode` is a spec-shape decision that belongs to you, not to me quietly.** Two routes, and they differ in what they claim:

1. **Import a caller-side well-formedness fact.** On the `mpt_node_kind` path the witness has already been validated, so the interiors *are* canonical. That is a precondition to carry, not a theorem about `rlpItemDecode`.
2. **Strengthen `rlpItemDecode`'s list arms** to record what `rlp_walk_next_core` actually verifies after #11776 (which made the wrapper require exact cursor-to-end exhaustion, status 7 on nested-malformed).

Route 2 looks more honest to me — #11776 means the routine really does more than the relation currently admits, so the relation is understating the guest — but it changes a relation other proofs consume, so I did not do it unilaterally.

## No regrade

`rlp_list_count_items` stays `.machineOnly`. `.bridged` would claim a bridge that still rests on an undischarged hypothesis, which is exactly what the correspondence registry exists to prevent. So this **addresses** #11795 without closing it.

No `sorry` anywhere — the unproven part is a hypothesis, matching the file's existing convention that nothing claims more than it has.

## Gates

`lake build` (4727 jobs, green, **0 warnings**) + check-no-warnings, check-progress, check-drift, check-forbidden-tactics, check-unimported, check-file-size, check-layering, check-heartbeats-approved, check-statement-tamper, check-routine-liveness — **all pass**. No `native_decide`/`bv_decide`, no `maxHeartbeats`.

`check-axioms.sh` cannot run in this checkout (`worktree-build-lock.sh` needs `flock`, absent on macOS, plus GNU `sha256sum`) — CI is the authority.

### DIFF
--- a/EvmAsm/Codegen/Programs/RlpListCountItemsBridge.lean
+++ b/EvmAsm/Codegen/Programs/RlpListCountItemsBridge.lean
@@ -287,4 +287,157 @@ theorem StrictPrefix.decodeChain_of_itemBridge
     refine ⟨items ++ [item], by simp [hlen], ?_⟩
     exact DecodeChain.snoc items hchain hdec
 
+/-! ## Fuel-sensitive forms (GH #11795, on top of #11711)
+
+    #11711 removed the fuel obstruction: `DecodeChainFrom` states a link at ONE
+    budget rather than at every budget, so a nested list — whose `decodeAux` is
+    fuel-sensitive — is now expressible as a link. This section carries the
+    structural results above over to that predicate, which is the half of #11795
+    that #11711 genuinely unblocks.
+
+    ⚠️ **It is only half.** #11795 says that once `DecodeChain` stops demanding
+    `∀ m`, discharging the per-item bridge "becomes assembling the existing pieces".
+    That is not so, and the reason is worth stating precisely because the issue
+    sequences work on it:
+
+    `rlpItemDecode`'s two LIST disjuncts (`WalkNext.lean:3649`, the `0xc0`–`0xf7`
+    and `≥ 0xf8` arms) require only header canonicality plus a **span fit** —
+    `¬ ult (endPtr - cursor) span`. They say nothing about the payload. `decodeAux`
+    (`Decode.lean:63`) additionally runs `decodeItems nDepth payload` and returns
+    `none` unless it consumes the payload exactly. So a list item whose span fits
+    but whose **interior is malformed** satisfies `rlpItemDecode` and is rejected by
+    `decodeAux`, at every budget.
+
+    ⇒ `RlpItemDecodeBridges`/`RlpItemDecodeBridgesFrom` remain unsatisfiable on the
+    unrestricted domain for a reason INDEPENDENT of fuel. Fuel was one of two
+    obstructions; #11711 closed that one. The residual is a genuine
+    strength mismatch between the machine relation and the model, and it is named
+    below as `RlpListInteriorsDecode` rather than left in prose — same discipline as
+    `RlpItemDecodeBridgesBytes`'s visible `hbytes` guard. -/
+
+/-- Per-item obligation in **fuel-sensitive** form: the model decode is exhibited at
+    a single budget `floor` instead of at every budget.
+
+    This is `RlpItemDecodeBridges` with #11711's fix applied. The `∀ m` in the
+    original is what a nested list cannot satisfy; this form can express one. -/
+def RlpItemDecodeBridgesFrom (bytes : List (BitVec 8)) (base endPtr : Word)
+    (floor : Nat) : Prop :=
+  ∀ (off : Nat) (next len : Word),
+    rlpItemDecode bytes off (base + BitVec.ofNat 64 off) endPtr next len →
+    ∃ item : RLPItem,
+      decodeAux floor (bytes.drop off) = some (item, bytes.drop (next - base).toNat)
+
+/-- The fuel-insensitive obligation implies the fuel-sensitive one at any positive
+    budget — instantiate `∀ m` at `m := floor - 1`. Faithfulness: the new predicate
+    asks for no more than the old one did. -/
+theorem rlpItemDecodeBridgesFrom_of_bridges {bytes : List (BitVec 8)}
+    {base endPtr : Word} {floor : Nat} (hfloor : 0 < floor)
+    (hb : RlpItemDecodeBridges bytes base endPtr) :
+    RlpItemDecodeBridgesFrom bytes base endPtr floor := by
+  intro off next len hitem
+  obtain ⟨item, hdec⟩ := hb off next len hitem
+  refine ⟨item, ?_⟩
+  obtain ⟨f, rfl⟩ : ∃ f, floor = f + 1 := ⟨floor - 1, by omega⟩
+  exact hdec f
+
+/-- `DecodeChainFrom`'s snoc lemma — the analogue of `DecodeChain.snoc`, needed for
+    the same reason: `DecodeChainFrom` recurses on the item list from the FRONT
+    while `StrictPrefix.succ` extends at the BACK. -/
+theorem DecodeChainFrom.snoc {bytes : List Byte} {item : RLPItem} {floor : Nat}
+    {off offMid offEnd : Nat} :
+    ∀ items : List RLPItem,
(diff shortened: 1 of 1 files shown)


pirapira:
Reviewed. **Content accepted; label held for one mechanical reason.**

⛔ **`baseRefName` is `work/11711-decodeaux-fuel-mono`, not `main`.** Labelling `merge!` now would have the actor merge this into #11894's branch rather than into main. #11894 is labelled and mergeable, so: once it lands, retarget to `main` and I will label immediately — the review is done, nothing further is owed on content.

## What I checked and accept

⭐ **The correction to #11795's premise is the valuable part.** That issue claimed that once #11711 removed the `∀ m` requirement, the per-item bridge "becomes assembling the existing pieces — not new mathematics." You show fuel was **one of two** independent obstructions, and name the second precisely: `rlpItemDecode`'s list disjuncts (`WalkNext.lean:3649`) require header canonicality plus a span fit and say **nothing about the payload**, while `decodeAux` (`Decode.lean:63`) additionally runs `decodeItems nDepth payload` and fails unless the payload is consumed exactly. So a list whose span fits but whose interior is malformed satisfies the machine relation and not the spec one. That is a strength mismatch no fuel work touches.

⭐ **`rlpItemDecodeBridgesFrom_of_parts` is the honesty check** and I read it as such: proving byte-string items plus `RlpListInteriorsDecode` exhaust the prefix space establishes that the residual is the *only* thing left, and in particular that it conceals no further fuel trouble. Without it, "one hypothesis remains" would be an assertion.

⭐ **The list restriction is visible in the predicate** (`¬ ult b 0xc0`), following `RlpItemDecodeBridgesBytes`' `hbytes` precedent, so no caller can mistake the structural bridge for a complete one.

⭐ **No regrade** — `rlp_list_count_items` stays `.machineOnly`, because `.bridged` would claim a bridge resting on an undischarged hypothesis. Correct, and it is why this *addresses* #11795 without closing it.

No `sorry`; the unproven part is a hypothesis, matching the file's convention.

## On your route question

You are right not to take it unilaterally, and I am not taking it either — it is with the maintainer. My own read matches yours: **route 2** (strengthen the list arms to record what `rlp_walk_next_core` verifies after #11776's exact cursor-to-end exhaustion, status 7 on nested-malformed) is the honest one, because the relation currently **understates the guest**. Route 1 carries a precondition rather than proving a theorem, and a caller-side well-formedness fact is exactly the kind of premise that later gets cited as if it were established.

⚠️ But route 2 changes a relation other proofs consume, so it needs a scoped issue naming those consumers before anyone edits it.

dhsorens:
Understood on the base branch, and you're right that labelling `merge!` now would merge this into #11894's branch rather than into main — my mistake in opening it stacked without flagging the retarget as a required step rather than a courtesy. #11894 has auto-merge armed, so I'll **retarget this to `main` as soon as it lands** and ping here; nothing else is owed on content.

## The scoped issue is filed: #11898

Per your requirement that route 2 name its consumers first. `grep -rn rlpItemDecode EvmAsm` gives **52 modules** outside the definition site, tiered by risk:

- **Tier 1, read-first (highest use counts):** `ItemDecodeForward` (19), `RlpListNthItemSAsmBase` (10, defines `StrictPrefix`/`StrictNthItem`), `RlpWalkDeterminism` (9), `WalkItemDeterminism` (7), `HeaderFieldsSpecCommon` (4).
- ⚠️ **The two determinism modules are the real risk.** `rlpItemDecode` being a *span* relation is plausibly load-bearing there — adding a payload-decode conjunct changes what "the same item" means. If those proofs case-split the five disjuncts it is mechanical; if they argue injectivity on spans it may need rework.
- Tiers 2–5: the walk chain (9), header family (7), account/EIP-161 (5), and the BAL chain (24 — largest by count, likely most mechanical), plus `Correspondence.lean`.

⭐ **And a way to avoid touching any of them.** The issue proposes an **additive** shape rather than an in-place edit: define `rlpItemDecodeStrict = rlpItemDecode ∧ list-payload-decodes`, prove `rlpItemDecodeStrict → rlpItemDecode`, and prove `rlp_walk_next` satisfies the strict form (which after #11776 it should, since that is what `rlp_walk_next_core` already verifies). Existing consumers keep the weak relation untouched; only the bridge consumes the strict one. That unblocks #11795 with **zero** changes across the 52 modules, and makes migration per-consumer and optional.

The cost is two relations to keep in step, mitigated by the implication lemma being the only tie required. In-place stays available as a fallback, and if it is chosen, the determinism modules should land first for the reason above.

Agreed on your reading of route 1, and it sharpens why: a caller-side well-formedness premise would sit exactly where a false-accept could hide, since the relation would stay weaker than the code while *looking* bridged.

dhsorens:
✅ **Retargeted to `main`** — #11894 merged at `2026-08-10T10:23:49Z`, so the stacking is gone.

`gh pr view 11895 --json baseRefName,mergeable` → `base=main mergeable=MERGEABLE`. The diff is now just the one file (`RlpListCountItemsBridge.lean`); #11894's `FuelMono.lean`, the `WalkDecodeBridge` additions and the two witness abbrevs all arrived via main, so nothing is duplicated here.

Per your note, the review is done on content and nothing further is owed — over to you for the label whenever convenient.

Also filed since your review: **#11898**, the scoped consumer enumeration you required before route 2 (52 modules, tiered, determinism modules flagged read-first, plus the additive `rlpItemDecodeStrict` shape that unblocks this with zero changes to those 52).

Issue #11898, opened 2026-08-10
scoped: strengthening rlpItemDecode's list arms (#11795 route 2) — 52 consuming modules, enumerated before anyone edits it


dhsorens:
> 🤖 Filed by Claude Code at @pirapira's request on PR #11895: *"route 2 changes a relation other proofs consume, so it needs a scoped issue naming those consumers before anyone edits it."* This is that enumeration. `main` at `372162cc2`.

## The change under consideration

`rlpItemDecode` (`EvmAsm/Rv64/RLP/WalkNext.lean:3649`) is the pure five-disjunct relation pinning `rlp_walk_next`'s outputs. Its two **list** arms (`0xc0`–`0xf7` and `≥ 0xf8`) require header canonicality plus a **span fit** — `¬ ult (endPtr - cursor) span` — and say nothing about the payload.

`decodeAux` (`EvmAsm/EL/RLP/Decode.lean:63`) additionally runs `decodeItems nDepth payload` and returns `none` unless the payload is consumed exactly. So the relation is **weaker than the model on list interiors**, which is what blocks `RlpItemDecodeBridges` (#11795, PR #11895 — where the residual is now isolated as `RlpListInteriorsDecode`).

⭐ Since #11776, `rlp_walk_next_core` **does** recursively validate list payloads — exact cursor-to-end exhaustion, status 7 on malformed / truncated / non-canonical / nested-malformed / trailing interiors. So the relation currently **understates the guest**. Route 2 is to strengthen the list arms to record what the routine actually verifies.

Both @pirapira and I independently read route 2 as the honest option: route 1 (import a caller-side well-formedness precondition) carries a premise rather than proving a theorem, and that is the kind of premise that later gets cited as though it were established. And a relation *weaker* than the code is the shape that lets a false-accept hide.

## Consumers — 52 modules outside the definition site

Full list obtained by `grep -rn rlpItemDecode EvmAsm`, excluding `WalkNext.lean` (the definition, 34 uses) and `RlpListCountItemsBridge.lean` (the bridge being unblocked).

**Tier 1 — structural, will need real proof work (highest use counts):**

| Module | uses |
|---|---:|
| `Rv64/RLP/ItemDecodeForward.lean` | 19 |
| `Codegen/Programs/RlpListNthItemSAsmBase.lean` | 10 |
| `Codegen/Programs/RlpWalkDeterminism.lean` | 9 |
[...]

Issue #11897, opened 2026-08-10
gate: check-axioms.sh cannot run on macOS — worktree-build-lock.sh needs flock + GNU sha256sum


dhsorens:
> 🤖 Filed by Claude Code (2026-08-10, `main` at `372162cc2`, macOS 25.5 / arm64).

## Symptom

`scripts/check-axioms.sh` — a **blocking** gate, and the kernel-truth backstop for the whole no-`native_decide`/no-`bv_decide` policy — cannot run in a macOS checkout:

```
$ bash scripts/check-axioms.sh; echo $?
worktree-build-lock: required command 'flock' is missing
2
```

It fails **loudly** (exit 2), which is the right behaviour and much better than a silent skip — this is a portability issue, not a soundness hole. Filing it because the practical effect is that every local session on macOS must report the repo's most important gate as "not run", and CI becomes the only place it is ever exercised.

## Cause

`scripts/check-axioms.sh:74` execs through `scripts/lib/worktree-build-lock.sh`, which hard-requires two GNU-isms:

1. **`flock`** (`:15`) — not present on macOS. This is the one that trips first.
2. **`sha256sum`** (`:22`, `lock_key=$(… | sha256sum | cut -c1-16)`) — also GNU-only; macOS has `shasum -a 256`. So even with `flock` installed via Homebrew, the lock-key computation would fail next.

Same helper gates `check-region-map.sh`, `check-build-parallel.sh`, `check-guarded-handler-bytes.sh`, `codegen-stateless-link-check.sh`, `codegen-force-relink.sh`, `codegen-eest-stateless-check.sh`, `fuzz-arith-diff.sh`, and `gen-symbol-addresses.py` — so this is one root cause behind a whole family of locally-unrunnable gates, not just the axiom one.

## Why it matters beyond convenience

`AGENTS.md` is explicit: *"If a gate cannot run in your checkout, say so plainly and explain why — do not list it among the gates you ran… a gate claimed but not run is worse than one openly skipped."* That rule works, and I have been following it. But it means the axiom gate's local feedback loop is *absent* on macOS, so a `sorryAx`/TCB-expanding regression is only ever caught after push. For the gate the project treats as its trusted-base backstop, a several-minute CI round trip is the only signal.

[...]

Issue #11803, opened 2026-08-08
hygiene: obligation matrix blockers are stale — row 4 cites 'M5 not shipped' against PLAN.md's 'M0–M10 are done'; re-audit rows 1/3/5 against current registries


pirapira:
> 🤖 Generated by Claude Code (strategist pass, 2026-08-08, main at 08c5d409a).

The obligation matrix (`EvmAsm/Progress/Obligations.lean`, rendered into PROGRESS.md) is the dashboard the end-to-end effort steers by — and at least one row's blocker text is provably stale:

- **Obligation 4** (`EVM interpreter loop on the decoded block`) is blocked by *"codegen M5 (tiny EVM interpreter) not shipped"* — but PLAN.md:23 has said **"M0–M10 are done"** (M5 = tiny EVM interpreter with runtime fetch/decode/dispatch) for weeks. Whatever still genuinely blocks obligation 4 (presumably: no simulation bridge between dispatched handlers and the SpecRef interpreter — now the h_ADD pilot issue), the row should say *that*, not cite a shipped milestone.
- **Obligation 3** (*"RV64 RLP decoder phases 1–3 (in progress)"*) and **obligation 5**'s opcode list (`MOD/SDIV/SMOD/ADDMOD/MULMOD/EXP/CALLDATACOPY/PUSH2..32`) predate several landings — re-audit both against the current `Progress.lean` registry (85 rows / 68 proven) and the RLP routine registry (33 rows / 24 proven) rather than trusting this issue's enumeration.
- **Obligation 1** (*"codegen emits rv64imac, one extension off zicclsm"*) — verify against the current emitter flags; if still true, it is fine, but it should carry a date or commit so staleness is detectable next time.

Small, mechanical, and worth doing before the umbrella issue starts steering by these rows: a dashboard whose blockers cite shipped work hides the real blockers behind it.

Issue #11573, opened 2026-08-05
predicates: codeDbIs for code_db_buckets, mirroring nodeDbIs — get_code-side proofs are unstartable without it


pirapira:
> 🤖 Generated by Claude Code (strategist pass over execution-specs ↔ SpecRef ↔ RISC-V correspondence, 2026-08-06).

## The gap

The node DB has a full predicate family — `nodeDbIs`, `nodeDbCountIs`, `nodeDbTopIs`, `resolveCacheValidIs` (`EvmAsm/Evm64/MptAssertions.lean:635-861`) — but its sibling `code_db_buckets` (`EvmAsm/Stateless/MemoryLayout.lean:175`) has **no predicate at all**.

The spec counterparts are parallel by construction: `build_node_db` and `build_code_db` sit side by side in `SpecRef/WitnessState.lean`, and `get_code` (`SpecRef/WitnessReads.lean`) reads the code DB the way account resolution reads the node DB. Code-hash → preimage resolution is FR/FA-relevant surface (#11410 replaced static code-preimage reprieves with the dynamic `code_reads` set; #11504 notes EXTCODESIZE/EXTCODECOPY coverage holes), so proofs over `get_code`-shaped guest routines are on the realistic path — and they are unstartable without this vocabulary.

## Ask

1. Define `codeDbIs` (+ count/occupancy cells as the layout requires) mirroring `nodeDbIs`'s keyed-bucket shape, over a `List (Hash32 × ByteRange)`-level model.
2. State the intended lookup lemma: a hash present in the model resolves to its preimage bytes (the `build_code_db` invariant: every entry's key is `keccak256` of its payload — decide whether the predicate carries that hash-binding or the triple does; record the choice).
3. Add the pairing (`codeDbIs` ↔ `build_code_db`/`get_code`) to the #11517 survey scope.

Deliberately narrow: one region, one existing pattern to clone, no routine triple included.

Related: #11410, #11504, #11517, docs/4ch8f-slstate-specref-correspondence.md (node-DB three-tier note at :164).

dhsorens:
Closing — the substantive work landed in **PR #11902** (merged). Summary, including a correction to this issue's premise.

## ⚠️ `codeDbIs` already existed

[...]

PR #11899, opened 2026-08-10
fix(progress): clear obligation 3's closed #11711 blocker, flagged by its own gate

### LINKED ISSUE #11711: proof: DecodeChain demands fuel-insensitivity, which blocks every nested-list bridge (rlp_list_count_items is the live case)

dhsorens:
`scripts/check-obligation-blockers.sh` — the advisory gate added hours earlier by #11888 (#11803's audit) — fired on a real case:

```
✗  #11711 is CLOSED but still blocks obligation 3: proof: DecodeChain demands fuel-insensitivity…
check-obligation-blockers: 9 issue-numbered blocker(s) checked, 1 stale.
```

#11894 closed #11711, and obligation 3's third blocker cited it. That is exactly the class-3 staleness the audit was about, caught on the audit author's own work the same day — the best available evidence the gate is not decorative.

## Clearing it revealed the text was *wrong*, not just stale

The blocker read *"nested-list decode bridges blocked on `DecodeChain` fuel-insensitivity (#11711; residual #11795)"* — attributing the block entirely to fuel.

Fuel was **one of two** independent obstructions. #11894 closed the fuel one and the bridge is **still blocked**, because `rlpItemDecode`'s list arms check a span fit and say nothing about the payload, while `decodeAux` rejects a malformed interior. That is a strength mismatch no fuel work touches — established in PR #11895 and confirmed by @pirapira there.

So the blocker now names the **surviving** obstruction and points at #11795 plus the relation-side decision scoped at #11898.

⚠️ Worth noting for the gate's own design: had I just deleted `#11711` from the list, obligation 3 would have lost its only pointer to a live blocker and read as *closer to done than it is* — the same failure the gate exists to prevent, arrived at from the opposite direction. A closed-issue flag is a prompt to **re-derive the row**, not to strike a token.

## Also refreshed in the same row

- `auditedAt` → `2026-08-10 @04de93895` (the merge commit this was checked at).
- The routine-registry counts in the note: 33 rows / 24 proven → **35 / 26**, which #11893 changed. Another small instance of the same decay — the note was accurate when written this morning and stale by lunchtime.

## Gates

`lake build` green + **check-obligation-blockers (1 stale → 0)**, check-progress, check-drift, check-forbidden-tactics, check-unimported, check-file-size, check-layering, check-heartbeats-approved, check-statement-tamper — all pass. PROGRESS.md and DRIFT.md regenerated.

`check-axioms.sh` not runnable in this checkout (#11897); CI is the authority. No proof statements touched.

Docs-and-registry only; no guest bytes, `.text` unchanged.


dhsorens:
Added a second commit correcting **my own error from #11888**, found while investigating #11573.

That audit replaced obligation 10's stale blockers, and one of the replacements *I wrote* was wrong on both halves:

```
.infra "codeDbIs predicate for code_db_buckets (#11573)"
```

- **`codeDbIs` already exists** — `Evm64/WitnessAssertions.lean`, landed in `73c8ea6a6` on **2026-07-05**, a month before #11573 was filed (2026-08-05), along with the count/occupancy cells and `witnessLookupSpec_correct`.
- **`code_db_buckets` is a dead scheme-A anchor.** Only mention in the tree is `Codegen/RegionMap.lean:198`; no emitted instruction references it, so no predicate was ever going to be built over it.

I took #11573's premise on trust instead of checking the tree — the same class of error the audit was fixing, committed inside the fix. The live gap is the code-DB **routines** (`witness_codes_index_build` / `witness_codes_lookup_by_hash`: raw asm, whole-guest byte-identity pins, no `cpsTripleWithin`), which is what the blocker now says, pointing at PR #11902 for the predicate side that is done.

Recorded in the row's `note` rather than silently overwritten. A dashboard whose corrections are invisible is the thing this PR is repairing, so my own correction should be as legible as the ones I made to older rows.

`auditedAt` refreshed to `@04de93895`. `check-obligation-blockers` now checks 10 issue-numbered blockers, 0 stale.

⚠️ Note this makes the row a **third** audit of the same day: the original #11888 audit, then obligation 3 cleared when its own gate flagged the closed #11711, now obligation 10 corrected. That rate of decay is itself the argument for the `auditedAt` field and the gate — three separate staleness events in one day on a dashboard nobody had audited in weeks.

Since this adds a commit after your `merge!` label, flagging explicitly so you can re-check rather than have it ride in on the earlier approval.

PR #11850, opened 2026-08-09
fix(gas): independent regular arm from settle meters (#11808)

pirapira:
## Summary

Closes the Case B FA control dependency on `bvgr_tx_total_state_gas` when deriving the Amsterdam block-regular dimension (`header.gas_used = max(Σ regular, Σ state)`).

### Spec does subtract — subtrahend honesty differs

At `execution-specs` @ `e5a8caf1b` `fork.py:1176-1181`:

```text
tx_regular_gas = max(before_refund - tx_state_gas, floor)
```

Spec's `tx_state_gas` is an **honest meter**. The guest was reconstructing the same shape as `before_refund - bvgr_tx_total_state_gas`, where `bvgr_tx_total_state_gas` is a **second free input** relative to the fee-pinned sum left — Case B FA (wrong state_total can move regular while before stays fixed).

### Regular meter is live at settle

`dispatcher_tx_gas_settle` already holds folded `reg_left` (x5) next to `state_left` (x7). Capture previously **discarded** it, storing only the sum into `bv_mtx_gas_left`. This PR **retains** the meters, not invents new ones.

### Floor attachment

| path | floor | cite |
|--|--|--|
| **TOTAL** fee/receipt | `max(after_refund, floor)` via `tx_gas_result_increments` | fork.py:1155-1159 |
| **REGULAR arm** block accumulator | `max(regular_consumed, floor)` in ExactGas | fork.py:1176-1181 |

Fee/receipt unchanged (sum path). Header max unchanged shape.

### Formula (spill-safe)

```text
regular_consumed = tx.gas - reg_left - state_left - state_used - intrinsic
                 = before_refund - (intrinsic + exec_used)
```

When `res_init=0`, state charges spill from regular into `state_used`; subtracting `state_used` recovers the spec regular (verified on stStaticCall recursive row: before 275924, state 195840, regular 80084).

Does **not** read `bvgr_tx_total_state_gas` for the regular arm.

## Changes

- `dispatcher_tx_gas_settle`: store folded reg_left, state_left, state_used
- Split: pin `runtime_tx_state_reservoir_initial`
- MTx/creation capture → `bv_mtx_regular_gas_left`, `bv_mtx_state_gas_left`, `bv_mtx_state_gas_used`, reservoir_init
- `BlockVerdictExactGas`: meter formula + floor
- Four-artefact regen; `check-asm-to-program` CLEAN

## ELFs / extents

| | BASE | FINAL |
|--|--|--|
[...]

Issue #11808, opened 2026-08-08
gate: header.gas_used is a max over two dimensions, so a regular-gas undercharge is invisible when state gas dominates


pirapira:
Split from #11145, whose **row no longer reproduces** — but the second half of that issue is
independent of the row and survives it.

## The gap

`header.gas_used` cannot see a regular-execution-gas undercharge when the **state dimension
dominates**.

At the pinned `e5a8caf1b`:

- `fork.py:370-375` — the header check is a **`max` over two dimensions** (block regular gas, state
  gas), **not a sum**.
- `fork.py:1181-1182` — the two accumulators update **separately**.
- `fork.py:1185-1187` — the receipt cumulative.

So when the state dimension is the larger of the two, an error in the regular dimension changes
nothing the header check compares, and a receipt cumulative may legitimately **exceed**
`header.gas_used` without that being a defect.

## Why this is worth tracking on its own

#11145 was a **confirmed false reject** — 570 gas of regular execution gas, `bv_fail=53` — and the
check that should have caught it was blind. The row was re-measured on clean `df80b1e`
(`repo_dirty=0`, guest `96e27025…`) at its byte-pinned input `7863341fd0688121…`: production and
SpecRef both match the manifest byte-for-byte at `succ=1`, debug `verdict=1 bv_fail=0
exact_block_status=0`. The 570/`bv_fail=53` signature is gone.

⛔ **The row passing does not close the blindness.** A defect of that shape would still be invisible
today; we would find it only if it happened to trip some *other* check — which is how the original
was found. The instrument gap outlived the instance.

## What would close this

A check that can discriminate the regular dimension independently of the state dimension, or an
argument (with the spec lines) that no regular-gas error can survive every other check that runs.

pirapira:
## Answer: **check branch** — the survival argument does not hold

codex2, analysis only, against the pinned spec.

The receipt root **is** the mechanism that caught the confirmed 570-gas row in #11145 — but it is
**not a complete witness for the regular dimension**, and here is the construction that shows why:

- `fork.py:1150-1159` computes receipt gas with **refund and floor** applied.
- So `tx_regular = max(B - state, F)` can **change** while `tx_gas_used = max(B - refund, F)` stays
  pinned at the **floor**.
- Add another transaction that makes `block_state_gas_used` dominate, and the **header check**
  (a max over two dimensions, `fork.py:370-375`) is also unmoved.

[...]

PR #11905, opened 2026-08-10
feat(tooling): host-only EIP-8037 gas-dimension differential (#11808)

pirapira:
## Summary
Tooling-only differential for the two EIP-8037 block gas dimensions after #11850 FA fix. **Zero guest edits.**

- SpecRef: `diagnose_stateless_gas_dims` exports `BlockOutput.blockGasUsed` / `blockStateGasUsed` from the same `apply_body` the oracle max-compares (not a side recomputation). CLI: `lake exe specref-eest-check --gas-dims <input>`.
- Host: `scripts/spike/gas_dims_diff.py` sums guest BSS `bvgr_block_gas_increments` / `bvgr_tx_total_state_gas` via `SPIKE_DUMP_RANGES`, demands equality on both dims.

## r200 (seed=932419178)
- **divergence_set: empty** (zero bar)
- 196 AGREE
- 4 SKIP_NO_SPECREF_DIMS (apply_body rejected early; guest tx_count=0 succ=0 — not mismatches)

## Cannot see (printed by tool)
- eip8037 max/equality vs header.gas_used (only summands compared)
- per-tx regular recomputation inside ExactGas from settle meters
- rows where apply_body never finishes

## Artefacts
`/tmp/grok-11808-gas-dims/r200/summary.json`

Closes nothing — optional detector tooling for #11808 residual visibility.

### DIFF
--- a/EvmAsm/Stateless/SpecRef/ElExecute.lean
+++ b/EvmAsm/Stateless/SpecRef/ElExecute.lean
@@ -52,12 +52,25 @@ private def dummyEvm (blockEnv : BlockEnvironment) : Evm :=
   { code := [], gasLeft := 0, stateGasLeft := 0, validJumpDestinations := [],
     message := msg, accessedAddresses := [], accessedStorageKeys := [] }
 
-/-- The `execute_block` interior below the pre-execution frame
-    (`fork.py`, function `execute_block`): `apply_body`, the diff
-    extraction, the post-state root, and the eight header checks. -/
-def execute_block_interior (pre : PrecompileMap) (ws : WitnessPreState)
+/-- Shared `BlockEnvironment` + `Machine` construction and `apply_body` run
+    (`fork.py` `execute_block` body prefix). **Single construction site** for
+    the production oracle (`execute_block_interior`) and the gas-dimension
+    diagnostic — a field added to `BlockEnvironment` cannot reach one without
+    the other.
+
+    Returns the `BlockOutput` and post-`apply_body` `Machine`. An `EvmError`
+    escaping `apply_body` is a Python exception escaping `execute_block` →
+    rejection.
+
+    Callers that only need gas dims re-run this same function and read
+    `blockGasUsed` / `blockStateGasUsed` from the returned `BlockOutput`
+    (SpecRef is deterministic; this is a second invocation, not a side
+    recomputation of gas arithmetic and not a plumb of the oracle's own
+    in-flight value). -/
+def apply_body_run (pre : PrecompileMap) (ws : WitnessPreState)
     (chainContext : ChainContext) (block : Block)
-    (transaction_public_keys : List Bytes) : Except SpecError Unit := do
+    (transaction_public_keys : List Bytes) :
+    Except SpecError (BlockOutput × Machine) := do
   let blockEnv : BlockEnvironment :=
     { chainId := chainContext.chainId
       blockGasLimit := block.header.gasLimit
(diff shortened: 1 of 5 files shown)


pirapira:
Holding on one thing, and it is the same hazard this PR already fixed once.

**The guest half and the boundary are right.** Zero guest edits, no emitter touched, the dimensions
read out of BSS by symbol through the existing `SPIKE_DUMP_RANGES` path, `divergence_set` empty on r200
with the four `SKIP_NO_SPECREF_DIMS` rows explained (guest `tx_count=0 succ=0`, `apply_body` rejected
early — not mismatches). The `cannot_see` list being printed *by the tool* rather than living in a PR
body is the right call.

## ⛔ The `blockEnv` / `machine₀` construction is a verbatim copy, not a shared one

`apply_body_block_gas_dims` rebuilds the environment by hand:

- all eleven `BlockEnvironment` fields, plus `transactionPublicKeys := some …`
- `machine₀ := { evm := dummyEvm blockEnv, txState := { parent := { preState := ws } } }`
- then `(apply_body pre blockEnv block.transactions block.withdrawals).run |>.run machine₀`

That is **character-for-character what `execute_block_interior` does** (`ElExecute.lean:61-80` on
`origin/main`), and `:96` is `max blockOutput.blockGasUsed blockOutput.blockStateGasUsed` — the value
compared against `header.gasUsed`. So today the two agree **by construction**, and I checked that
rather than taking the docstring's word for it.

⚠️ **The problem is the "today".** A differential whose reference side rebuilds the environment
independently stops being a differential the moment the oracle's construction changes and the copy does
not follow — and it fails *silently*, agreeing with itself while measuring a different environment.
This PR already recognised exactly this shape by extracting `elPrepareBlock` for the pre-check prefix;
the `blockEnv` / `machine₀` / `apply_body` prefix deserves the same treatment. **Extract it once and
have both call it**, so a future field added to `BlockEnvironment` cannot reach the oracle without
reaching the diagnostic.

## ⚠️ One wording fix, because the claim is stronger than the code

The docstring says these are *"the oracle's accumulator fields, not a side recomputation of gas
arithmetic."* Half right. They are the same **fields** from the same **function** — but obtained from a
**second invocation** of `apply_body`, not plumbed out of the run the oracle actually compared. That is
sound here (SpecRef is deterministic, and re-running is legitimate for a guest differential), and it is
what makes the shared-prefix ask above load-bearing rather than cosmetic. Please say *re-runs the same
`apply_body` and reads the same fields* instead of implying the value is the oracle's own.

Nothing else. Once the prefix is shared and the wording is narrowed I will label — and `build` needs to
finish green regardless.

pirapira:
## Review hold addressed

### Structural
Extracted **`apply_body_run`** — single site for `BlockEnvironment` + `Machine` + `apply_body`.

- `execute_block_interior` → `apply_body_run` then post-body checks
- `apply_body_block_gas_dims` → `apply_body_run` then read `blockGasUsed` / `blockStateGasUsed`

A field added to `BlockEnvironment` cannot reach the oracle without reaching the diagnostic (same shape as the existing `elPrepareBlock` share).

### Wording
Docstrings / CLI note / tool header now say: **re-runs shared `apply_body_run` and reads the same fields** (second invocation; SpecRef deterministic) — not a plumb of the oracle's own in-flight value, and not a side recomputation of gas arithmetic.

### Build
`lake build` SpecRef + SpecRefEestCheck green. Spot `--gas-dims` on c4c2a193 returns dims with updated `source`/`note`.

Issue #11572, opened 2026-08-05
predicates: model the undo journals (storage/account write rollback) and state replay-equals-restoreTxState — rollback is unmodeled and is where soundness bugs cluster


pirapira:
> 🤖 Generated by Claude Code (strategist pass over execution-specs ↔ SpecRef ↔ RISC-V correspondence, 2026-08-06).
> **Corrected 2026-08-08** against `origin/main` — the journal sizes, the capacity bound and the "shared shape" claim were wrong as filed. Corrections measured by @dhsorens ([comment](https://github.com/Verified-zkEVM/evm-asm/issues/11572#issuecomment-5223142856)) and re-verified against `origin/main` before editing. The **ask is unchanged**; two of the errors would have produced a wrong predicate rather than an imprecise one.

## The gap

The rollback machinery — `storage_writes_undo_area` and `account_writes_undo_area` (`EvmAsm/Stateless/MemoryLayout.lean:381,450`) — is **entirely unmodeled**: no `Assertion` predicate, no list-level model, nothing in `docs/4ch8f-predicate-shapes.md`.

Its spec counterpart is precise: `restore_tx_state` (`state_tracker.py:809-826` at pin `e5a8caf1b`; port `SpecRef/StateTracker.lean` `copyTxState`/`restoreTxState`) restores the write dicts by copy. The guest implements the same effect as an undo journal: push pre-values on overwrite, replay on revert.

Rollback is where soundness bugs have repeatedly clustered — #11189 (undo journal fail-open on overflow), #11198 (no capacity guard), #11078 (auth OOG applies instead of rolling back), #11001 (reverting child leaves caller short). Every one of these is invisible to routine-local proofs until the journal has a predicate and the revert path has a stated theorem.

## The two journals are NOT the same shape

⚠️ **As filed this issue said "2 MiB each, capacity 16,384 entries per #11198". That is the account journal only.**

| journal | area | size | capacity | stride |
|---|---|---:|---:|---:|
| `STORAGE_WRITES_UNDO_AREA` | `MemoryLayout.lean:381` | **5 MiB** | **32768** | **160 B** |
| `ACCOUNT_WRITES_UNDO_AREA` | `MemoryLayout.lean:450` | 2 MiB | 16384 | 128 B |

[...]

Issue #11571, opened 2026-08-05
predicates: Assertion vocabulary for the spec-shaped write maps (account_writes / tx_account_writes / storage_writes) — the containers the convergence moves toward have no predicates


pirapira:
> 🤖 Generated by Claude Code (strategist pass over execution-specs ↔ SpecRef ↔ RISC-V correspondence, 2026-08-06).

## The gap

The guest is converging on the spec's two-container state model (#11326 item 3, #11329): the spec-shaped `account_writes`/`tx_account_writes` map pair landed in `EvmAsm/Codegen/Programs/AccountWriteMap.lean:145-467` and the storage map in `StorageWriteMap.lean:121-181`, backed by regions `account_writes_area`/`tx_account_writes_area` (`EvmAsm/Stateless/MemoryLayout.lean:416,418`) and the storage-writes map arena.

**None of these regions has a separation-logic `Assertion` predicate.** The predicate inventory covers the *legacy* append-only logs (`storageLogIs`, `EvmAsm/Evm64/StorageAssertions.lean:114`) and everything in `docs/4ch8f-predicate-shapes.md`, but the spec-shaped maps — the containers the whole convergence effort is moving *toward* — are describable today only as raw bytes.

This matters for sequencing: every future state-tracker triple (get/set account, `mergeTxIntoBlock`, BAL emit-from-tx-writes per #11104) needs a vocabulary like

```
accountWritesMapIs : (entries : List (Address × AccountValue)) → Assertion
txAccountWritesMapIs : … → Assertion
storageWritesMapIs : (entries : List ((Address × Slot) × Value)) → Assertion
```

relating the arena bytes to a `List`-level model that maps directly onto `SpecRef/StateTracker.lean`'s `BlockState.account_writes` / `TransactionState.account_writes` dicts (`state_tracker.py:57-109` at pin `e5a8caf1b`). Defining them *now* means the proofs land against the container that survives, not against the legacy logs slated for retirement (#11256).

## Ask

1. Read the map layouts out of `AccountWriteMap.lean` / `StorageWriteMap.lean` (entry stride, key encoding, occupancy discipline) and define the three predicates above, following the `teerEntriesFrom` run-predicate pattern (`EvmAsm/Codegen/RegionPredicates.lean:344`) that `docs/leaf-routine-targets.md` row 5 already prescribes for BAL runs.
[...]

PR #11908, opened 2026-08-10
feat(stateless): predicates for the storage/account undo journals

dhsorens:
Addresses #11572. I'm assigned. **Stacked on #11906** (#11571 — the map-side "map view" this depends on, which the issue names as its dependency). I'll retarget to `main` once that lands.

## ⚠️ Both of this issue's *corrected* constants are stale again

#11572 was already corrected once (2026-08-08) because its first numbers *"would have produced a wrong predicate rather than an imprecise one"*. Measured against the tree today, the corrected numbers have moved too:

| journal | #11572 (corrected) | actual |
|---|---|---|
| storage undo | 32768, `× 160 = 0x500000` | **167652**, `× 160 = 0x1994e80` |
| account undo | `txAccountWritesCapacity` = 16384 | **`accountWritesUndoCapacity` = 163840** |

**The account one is dangerous.** `Codegen/Programs/AccountWriteUndo.lean:29` now says in as many words that the bound is *"no longer the same constant as `txAccountWritesCapacity`"*, and `:61` explains why — the journal is bounded by write **events**, and repeated updates to one key add events without adding map rows. Item 3 of this issue instructs using `txAccountWritesCapacity`; doing so would under-bound that journal by **10×**, reproducing precisely the failure its own correction note warns about.

Values here come from the tree (`MemoryLayout.lean:356,450`, `StorageWriteMap.lean:123`, `AccountWriteUndo.lean:52`), and two `by decide` examples cross-check them against the documented arena sizes, so a future constant edit that stops matching its region fails the build rather than drifting.

## What's in it

- `UndoKind` (`overwrite`/`append`/`destroyDrop` = emitted `wasAbsent` 0/1/2) + `.code`; `StorageUndoEntry`, `AccountUndoEntry`, `*EntryIs` over the layouts documented in **core** `MemoryLayout.lean`.
- `storageUndoFrom` / `accountUndoFrom` — **stride-parameterised**, per your instruction that a shared shape is wrong (160 ≠ 128). Mirrors `RegionPredicates.lean:718`'s `balEntriesFrom (stride) (base)`; mirrored not imported, same core-side layering constraint as `WriteMapAssertions.lean`.
- `storageUndoJournalIs` / `accountUndoJournalIs`; `undoCursorIs` with the address as a parameter (the cursor is a link-time `.bss` symbol, not a `MemoryLayout` constant).
- **`StorageUndoJournalWF` / `AccountUndoJournalWF` — item 3: the capacity guard is part of well-formedness**, so a push-site proof must discharge it. #11189's fail-open state is not representable as a well-formed journal.
- `pcFree_*` and `_cons` for both runs.
- **Item 2's replay statement**: `replay ((es.drop mark).reverse) rowsNow = rowsAtMark`. `replay` is a **parameter**, not a definition — the replay function is what a walker's triple will exhibit, and fixing it here would prejudge that shape. The `.reverse` is load-bearing: nesting is LIFO, so reverse replay is what makes unwinding exact.

## Non-vacuity, kernel-checked

All three kinds well-formed at their own payload lengths; ⭐ **negative control** that an `.append` with a 32-byte payload and an `.overwrite` with none are *both* rejected — what stops `wasAbsent` degenerating into a sentinel, the distinction #11189's bug lived beside; codes are 0/1/2; ⭐ a journal **at** capacity is well-formed and one **over** it is not; and the two capacities are asserted `≠` the issue's values so this correction cannot silently regress.

## Not done, with reasons

**`snoc` is deliberately absent.** Its statement needs `(base + ofNat stride) + ofNat (stride * n) = base + ofNat (stride * (n+1))` over `BitVec 64` with a *variable* stride — `ofNat`-of-a-product, which `bv_omega` doesn't see through. Nothing consumes it yet (`snoc` is what a **push-site** triple needs, outside this issue's scope), so it would be scaffolding for an absent caller. The exact obstruction is recorded in the file so the next attempt starts from it rather than rediscovering it.

The replay theorem is **stated, not proved** — it needs the push/replay walker triples, which the issue puts out of scope.

## Gates

`lake build` (4729 jobs, green, **0 warnings**) + check-no-warnings, check-layering, check-unimported, check-progress, check-drift, check-forbidden-tactics, check-file-size, check-heartbeats-approved, check-statement-tamper, check-obligation-blockers — **all pass**. No `sorry`/`admit`, no `native_decide`/`bv_decide`, no `maxHeartbeats`.

### DIFF
--- a/EvmAsm/Stateless.lean
+++ b/EvmAsm/Stateless.lean
@@ -68,6 +68,7 @@ import EvmAsm.Stateless.Headers.Validate
 import EvmAsm.Stateless.State.Account
 import EvmAsm.Stateless.State.AccountAssertions
 import EvmAsm.Stateless.State.WriteMapAssertions
+import EvmAsm.Stateless.State.UndoJournalAssertions
 import EvmAsm.Stateless.State.BlockState
 import EvmAsm.Stateless.State.Diff
 import EvmAsm.Stateless.State.PreState
--- a/EvmAsm/Stateless/State/UndoJournalAssertions.lean
+++ b/EvmAsm/Stateless/State/UndoJournalAssertions.lean
@@ -1,0 +1,427 @@
+/-
+  EvmAsm.Stateless.State.UndoJournalAssertions
+
+  Separation-logic assertions for the guest's **rollback undo journals** (GH #11572).
+
+  `restore_tx_state` (`state_tracker.py:809-826`; port `SpecRef/StateTracker.lean`
+  `copyTxState`/`restoreTxState`) restores the write dicts **by copy**. The guest
+  cannot afford a dict copy at capacity × call depth, so it implements the same effect
+  as an undo journal: push pre-values on overwrite, replay in reverse on revert.
+
+  That machinery was **entirely unmodeled** — no predicate, no list-level model — and
+  rollback is where soundness bugs have clustered: #11189 (undo journal fail-OPEN on
+  overflow), #11198 (no capacity guard), #11078 (auth OOG applies instead of rolling
+  back), #11001 (reverting child leaves caller short). Each is invisible to
+  routine-local proofs until the journal has a predicate and revert has a statement.
+
+  ## ⚠️ The two journals are NOT the same shape, and the constants have MOVED AGAIN
+
+  #11572 was already corrected once (2026-08-08) because its first numbers would have
+  produced a *wrong* predicate rather than an imprecise one. **Both corrected numbers
+  are now stale too**, measured against the tree on 2026-08-10:
+
+  | journal | #11572 (corrected) | actual |
+  |---|---|---|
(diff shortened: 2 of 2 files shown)


dhsorens:
✅ **Retargeted to `main`** — #11906 (#11571) merged, so the stack is gone. `base=main mergeable=MERGEABLE`; the diff is now just the new `UndoJournalAssertions.lean` plus the umbrella line, since `WriteMapAssertions.lean` arrived via main.

The ⚠️ in the description still stands and is the thing to check before merging: **both of #11572's already-corrected capacity constants are stale again**, and following item 3 literally would under-bound the account journal by 10× (`txAccountWritesCapacity` = 16384 vs the actual `accountWritesUndoCapacity` = 163840, which `AccountWriteUndo.lean:29` now says in as many words is "no longer the same constant"). This PR uses the tree's values and cross-checks them against the documented arena sizes with two `by decide` examples.

Issue #11654, opened 2026-08-06
proof: SLOAD regressed .conditional → .execSpec when the exec-log vocabulary retired — re-prove against StorageWriteMap, gated on #11651/#11653 shape-final


pirapira:
> 🤖 Generated by Claude Code (strategist pass, 2026-08-07, main at b48148196).

## The regression to record

When #11596/#11595 retired the append-only persistent storage exec-log vocabulary, SLOAD's stage-1 proof went with it (`Evm64/Storage/LoadSpec.lean` + `LoadLoopSpec.lean`, 45 theorems — deleted as scoped). The registry row regressed accordingly: `EvmAsm/Progress.lean:334` now reads `entry "SLOAD" .execSpec none`, down from `.conditional` (`Storage.evm_sload_stack_spec_within`, PLAN.md's `conditionalCount 0→1` milestone).

This was the *right* call — the proof described a container the guest no longer has — but no issue tracks the road back, and SLOAD is currently the only storage opcode whose spec coverage moved **backward** during the convergence.

## The re-proof target

SLOAD against the **StorageWriteMap** read path (`Codegen/Programs/StorageWriteMap.lean` map lookup → witness fallback), phrased over the map predicate `storageWritesMapIs` from #11571 — i.e. the SpecRef anchor becomes `StateTracker`'s storage dict lookup (`state_tracker.py:74,101` at pin `e5a8caf1b`) instead of the retired log's `persistentLookup`.

## ⛔ Gate — do not start yet

The storage container is still being reshaped: #11651 (builder per-BAI arrays), #11653 (BAI-preserving replacement for the exec-log `txindex`), and the #11256 consumer census all touch what the map read path must satisfy. Starting the triple before those land re-creates exactly the situation #11596 just cleaned up — a large proof against a container mid-retirement. Sequence: #11653/#11651 shape-final → #11571's `storageWritesMapIs` → this triple. SSTORE follows the same gate and can ride the same vocabulary.

## Why file it now anyway

So the `.conditional → .execSpec` regression is a tracked debt with a stated gate rather than a silent row in the registry — the same reasoning that keeps `DRIFT.md` honest. TSTORE/TLOAD are unaffected (their proofs were split onto `transientLogIs`, not deleted).

Related: #11596, #11595, #11571, #11651, #11653, #11256.

pirapira:
[...]

Issue #11651, opened 2026-08-06
BAL attribution: builder semantics — per-field reducers, net-zero omission, per-BAI ordering, exact RLP


pirapira:
Split out of #11104 (closed).

## Closing condition

The BAL builder reproduces the spec's `AccountChanges` exactly: pre-merge baseline comparison, net-zero omission, **field-specific** reducers, per-`(address, BAI)` sorted rows, and byte-exact RLP.

## The structural fact this rests on

⛔ **The account-write map cannot express per-BAI history.** It is keyed by address and holds one folded value per field; the spec's `AccountChanges` holds a **list** of changes, each carrying its own `block_access_index`. So prior-BAI history must be emitted into the **builder arrays** before the tx→block merge — the builder is the only structure that can carry it.

⛔ **Do not add a hand-maintained BAI field to the final account row.** A row is a fold; history is a list. This was considered and rejected twice during the #11104 analysis.

## Spec operators, per field

| field | operator | site |
|---|---|---|
| `balance_changes` | LAST-WRITE | `block_access_lists.py:403-419` |
| `nonce_changes` | **MAXIMUM** | `:440-456` |
| `code_changes` | same-BAI final write | `:478-496` |
| storage | same-BAI final write | `:353-368` |

Different BAIs remain **distinct history** — the reducers apply *within* a BAI, never across.

## The no-op storage gate, cited correctly

A write becomes a `storage_change` only when it changes something (`:666-676`):

```python
pre_value = _get_pre_tx_storage(block_state.storage_writes, pre_state, address, key)
if pre_value != post_value:
    add_storage_write(builder, address, u256_slot, idx, post_value)
```

⚠️ `:549-552` is the **read-exclusion consequence**, not this gate — it explains why a no-op slot surfaces as a *read*. Both a prior agent and I mis-cited it on #11647; cite `:666-676` as the authority.

⚠️ The baseline is **block-cumulative then pre-state**, never tx-local. A tx-local baseline would compare equal for a slot changed by an **earlier transaction in the same block** and suppress a required change — which is **FA-ward**.

## Evidence bar

Same-BAI duplicate fixtures proving balance last-write, nonce maximum and code last-write, net-zero omission, and an exact RLP differential.

[...]

PR #11914, opened 2026-08-10
proof(stateless): the storage read path as a finite-map lookup (SLOAD groundwork)

dhsorens:
Addresses #11654. I'm assigned. **Stacked on #11906** (#11571 — needs `storageWritesMapIs`); retargeting to `main` once that lands.

## ⭐ #11654's gate is fully lifted — its "⛔ do not start yet" is stale

That issue's sequence is `#11653`/`#11651` shape-final → #11571's `storageWritesMapIs` → the triple. Verified today: **#11651, #11653 and #11256** (the consumer census it also names) are **all CLOSED**, and `storageWritesMapIs` landed with #11571. Worth flagging because the issue currently reads as blocked and will be skipped in triage.

SLOAD is the only storage opcode whose coverage moved **backward** during the convergence — #11596/#11595 retired the exec-log and its 45-theorem stage-1 proof went with it, regressing the row `.conditional → .execSpec`.

## What this is (and is not)

This is the piece **between the predicate and the triple**. The triple itself — against the emitted `StorageWriteMap` read routine plus the witness fallback — is the large remaining half and is not here.

New `EvmAsm/Stateless/State/StorageReadPath.lean`:

- `storageRowLookup` — the read path as a lookup; `find?` because the guest scans and stops.
- `storageRowLookup_sound` — a `some` answer is witnessed by a real row. No map property needed.
- ⭐ **`storageRowLookup_eq_some_of_mem`** — every row in a well-formed map is found by its own key with its own value. **This is where `StorageWriteRowsMap`'s `Nodup` earns its place:** without it, `find?` could return an *earlier* row sharing the key and the read path would not be a function of the key at all. Uniqueness holds because `storage_write_record` upserts in place — a property of the **writer**, which is exactly why it is a precondition here and not a lemma.
- `storageRowLookup_eq_none_iff` — a `none` answer means no row keys on the pair. This is the form a witness-fallback proof needs: it must establish the map genuinely missed *before* it may consult the witness.
- `SloadReadPath` — the obligation, over **both** levels: tx shadows block shadows witness, in that order, because the spec has two containers and checks tx first.
- ⭐ **`sloadReadPath_canonical`** — the obligation is **satisfiable**, exhibited by the two-level lookup itself. A `Prop`-valued obligation nobody has instantiated could be unsatisfiable, and then a future *"SLOAD satisfies `SloadReadPath`"* would be vacuous. This rules that out and names the function the triple should be proved against.

`witnessValue` is a **parameter**, not a definition: on a map miss the guest consults the witness, and what the witness yields belongs to the witness-read spine (#11579). Fixing it here would prejudge that half.

## Non-vacuity, kernel-checked

Each row found by its own key with the right value; ⭐ **the same slot in a different contract misses** — the address is part of the key, and a lookup keyed on the slot alone would silently return another contract's value; an unwritten slot misses, so the witness fallback is reachable rather than dead code; and ⭐ **tx level shadows block level**, the ordering `restore_tx_state` / `mergeTxIntoBlock` depend on and the one a single-level model gets wrong.

## Gates

`lake build` (4729 jobs, green, **0 warnings**) + check-no-warnings, check-layering, check-unimported, check-progress, check-drift, check-forbidden-tactics, check-file-size, check-heartbeats-approved, check-statement-tamper, check-obligation-blockers — **all pass**. No `sorry`/`admit`, no `native_decide`/`bv_decide`, no `maxHeartbeats`.

### DIFF
--- a/EvmAsm/Stateless.lean
+++ b/EvmAsm/Stateless.lean
@@ -68,6 +68,7 @@ import EvmAsm.Stateless.Headers.Validate
 import EvmAsm.Stateless.State.Account
 import EvmAsm.Stateless.State.AccountAssertions
 import EvmAsm.Stateless.State.WriteMapAssertions
+import EvmAsm.Stateless.State.StorageReadPath
 import EvmAsm.Stateless.State.BlockState
 import EvmAsm.Stateless.State.Diff
 import EvmAsm.Stateless.State.PreState
--- a/EvmAsm/Stateless/State/StorageReadPath.lean
+++ b/EvmAsm/Stateless/State/StorageReadPath.lean
@@ -1,0 +1,224 @@
+/-
+  EvmAsm.Stateless.State.StorageReadPath
+
+  The `storage_writes` **read path** as a finite-map lookup (GH #11654 groundwork).
+
+  ## Why this exists
+
+  SLOAD's stage-1 proof was deleted when #11596/#11595 retired the append-only
+  persistent storage exec-log (`Evm64/Storage/LoadSpec.lean` + `LoadLoopSpec.lean`,
+  45 theorems), and the registry row regressed `.conditional → .execSpec`
+  (`Progress.lean` `entry "SLOAD" .execSpec none`). That was the right call — the
+  proof described a container the guest no longer has — but SLOAD is the **only**
+  storage opcode whose coverage moved *backward* during the convergence.
+
+  #11654's stated sequence is `#11653`/`#11651` shape-final → #11571's
+  `storageWritesMapIs` → the SLOAD triple. As of 2026-08-10 **the gate is fully
+  lifted**: #11651, #11653 and #11256 (the consumer census) are all closed, and
+  `storageWritesMapIs` landed with #11571.
+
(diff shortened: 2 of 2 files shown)


dhsorens:
✅ **Retargeted to `main`** — #11906 (#11571) merged, so the stack is gone. `base=main mergeable=MERGEABLE`; the diff is now just the new `StorageReadPath.lean` plus the umbrella line.

Recap of the load-bearing bit: `storageRowLookup_eq_some_of_mem` is where `StorageWriteRowsMap`'s `Nodup` earns its place — without it `find?` could return an earlier row sharing the key and the read path would not be a function of the key at all. Uniqueness holds because `storage_write_record` upserts in place, i.e. it is a property of the **writer**, which is why it stays a precondition here rather than becoming a lemma.

Also worth pairing with **#11913** (filed today): MSTORE's `.proven` triple turns out to be vacuous for every unaligned offset because its precondition names eight pairwise-disjoint dwords that cannot all exist. That is the hazard the SLOAD re-proof should avoid, and it is the argument for stating the read path over `storageWritesMapIs` rather than over a caller-supplied set of disjoint cells.

Issue #11692, opened 2026-08-07
proof: withdrawal_decode's .machineOnly premise ('no reference RLP decoder') is now escapable via decode_encode


dhsorens:
Split out of #11341, where I found this while checking that issue's four rows. **This is not #11341's defect** — recording the distinction because conflating them wastes someone's time.

## The row's rung is currently justified

`Correspondence.registry`'s `withdrawal_decode` row is `.agrees`/`.machineOnly`, and its note explains why:

> SpecRef carries withdrawal ENCODE (`withdrawalToRlpItem`) and SSZ decode (`sszToWithdrawal`) but **NO RLP decoder**, so there is nothing to bridge to and the differential does not transfer — `machineOnly`, not `bridged`.

That is a legitimate `machineOnly`: the reference side genuinely does not exist. #11341's four rows are the *opposite* case — a reference exists and the proof was aimed at a local restatement of it. Different defects, different fixes.

## ⭐ Why the premise may no longer hold

Two pieces that exist today and are not connected:

```lean
-- EvmAsm/Stateless/SpecRef/BlocksRlp.lean:54
def withdrawalToRlpItem (w : Withdrawal) : RLPItem :=
  .list [scalarItem w.index, scalarItem w.validatorIndex,
         .bytes w.address, scalarItem w.amount]

-- EvmAsm/EL/RLP/Properties.lean:2548  (classical-3, no sorryAx)
theorem decode_encode (item : RLPItem) (h : (encode item).length < 256 ^ 8) :
    decode (encode item) = some (item, [])
```

Composing them gives a reference **decoder** for withdrawals rather than lacking one:

```
decode (encode (withdrawalToRlpItem w)) = withdrawalToRlpItem w
```

and the four fields read straight off the `.list`. `decode_encode` is not referenced from any withdrawal module, so this composition appears simply not to have been noticed — the row's premise was written when it wasn't available, or at least wasn't connected.

## ⚠️ What this gives, and what it does not

- ✅ **Round-trip direction**: every well-formed withdrawal encoding decodes back to its item.
- ❌ **The converse**: *"if `decode bs` succeeds with a 4-item list of the right shapes, then `bs` is the encoding of some `Withdrawal`."*

The converse is the real content for a `.bridged` upgrade — it is where injectivity and canonicality do the work, and it is what would let the differential transfer for arbitrary input rather than only for bytes already known to be an encoding.

[...]

PR #11916, opened 2026-08-10
proof(rlp): encode_decode — the decoder has no non-canonical accepts

### LINKED ISSUE #11896: proof: encode_decode (canonicality direction) is missing — the prerequisite for every .machineOnly → .bridged upgrade whose reference side is an encoder
> 🤖 Filed by Claude Code while working #11692 (2026-08-10, `main` at `372162cc2`).

## The gap

`EvmAsm/EL/RLP/Properties.lean` proves the round-trip in **one** direction only:

```lean
theorem decode_encode (item : RLPItem) (h : (encode item).length < 256 ^ 8) :
    decode (encode item) = some (item, [])
```

The other direction — **canonicality** — is absent:

```lean
-- WANTED
theorem encode_decode {bs : List Byte} {item : RLPItem}
    (h : decode bs = some (item, [])) : encode item = bs
```

Checked by statement, not by name. The file has `decode_encode`, `decode_encode_mutual`, `decodeFully_encode`, `encode_injective`, `encode_nonempty`, `encode_append_cancel`, `encode_prefix_unambiguous`. None of them is this.

⚠️ **`encode_injective` is not a substitute.** It gives `encode i₁ = encode i₂ → i₁ = i₂`, relating two *encodings*. `encode_decode` relates an **arbitrary decodable byte string** to an encoding. That difference is the whole content: the first says encode is injective, the second says decode has no non-canonical accepts.

## Why it should be true

The decoder already enforces canonicality at every form (`EvmAsm/EL/RLP/Decode.lean:61-106`):

- single byte `< 0x80` must use the single-byte form — a 1-byte short string whose content is `< 0x80` is rejected (`:78`);
- long string / long list reject `lenVal ≤ 55`, so the short form is mandatory below 56 (`:85`, `:101`);
- list payloads must be consumed exactly (`leftover.isEmpty`, `:94`, `:105`).

So this is unproven, not false. Expected shape: mutual induction on fuel mirroring `decode_encode_mutual`, with a `decodeItems`-level companion (`encode.encodeItems items = bs` from `decodeItems n bs = some (items, [])`).

## Why it is worth its own issue

It is the **load-bearing prerequisite for a whole class of correspondence upgrades**, not a one-off.

The `Correspondence.registry` pattern: a row is `.machineOnly` because "SpecRef carries ENCODE but no decoder, so there is nothing to bridge to and the differential does not transfer". Composing the SpecRef encoder with `decode_encode` gets you a reference *decoder* — but only for bytes **already known to be an encoding**. Transferring the differential for *arbitrary input* is exactly `encode_decode`.

Worked instance, #11692 (`withdrawal_decode`): from `decode bs = some (.list [i₁,i₂,i₃,i₄], [])` plus shape conditions you can build the `Withdrawal` and get `withdrawalToRlpItem w = .list [i₁,i₂,i₃,i₄]`; what remains is `bs = encode (.list [i₁,i₂,i₃,i₄])`, which *is* `encode_decode`. #11692 anticipated "may need canonicality assumptions the guest does not enforce" — the sharper answer is that the guest *does* enforce them and the theorem is simply missing.

## Scope

- `encode_decode` plus its `decodeItems` companion, in `EL/RLP/Properties.lean` or a new sibling.
- A witness `abbrev` so it reaches `scripts/check-axioms.sh` — referencing a theorem does not gate it (PLAN.md, #11671). #11678/PR #11891 just closed exactly this hole for `decode_encode` itself, which had been proven and gated by nothing.
- ⚠️ Watch the length side condition. `decode_encode` carries `(encode item).length < 256 ^ 8`; state `encode_decode`'s own precondition from what `decode` guarantees rather than copying that bound by reflex.

## Explicitly not in scope

- Regrading any `Correspondence` row. `.bridged` upgrades are per-row decisions with their own port-fidelity obligations (#11692 for withdrawals).
- `RlpItemDecodeBridges` / the nested-list bridge — that is blocked on a *different* thing (`rlpItemDecode`'s list arms check a span fit, not payload decode; see #11795 and PR #11895), and no amount of `encode_decode` fixes it.

Related: #11692 (the instance that surfaced this), #11678 / PR #11891 (`decode_encode` witnessed), #11341, #11661.

dhsorens:
**Fixes #11896**, which I filed earlier today while working #11692 after finding the canonicality direction of the RLP round trip absent. I'm assigned.

`Properties.lean` proves `decode_encode` — every encoding re-decodes to itself. That says nothing about bytes that are *not* an encoding. This is the converse:

```lean
theorem encode_decode {bs : List Byte} {item : RLPItem}
    (h : decode bs = some (item, [])) : encode item = bs
```

A successful complete decode **identifies its input as the encoding of what it produced** — so the decoder has no non-canonical accepts.

⚠️ `encode_injective` is *not* a substitute: it relates two encodings to each other. The content here is about an **arbitrary decodable byte string**, which is what a differential transfer needs.

## Why it matters beyond the round trip

It's the prerequisite for a whole class of `Correspondence` upgrades. The recurring `.machineOnly` reason is *"SpecRef carries ENCODE but no decoder, so there is nothing to bridge to."* Composing a SpecRef encoder with `decode_encode` gives a reference decoder — but only for bytes **already known to be an encoding**. Transferring the differential for *arbitrary* input is exactly this theorem. #11692 (`withdrawal_decode`) is the worked instance; what remained there was `bs = encode (.list [...])`, i.e. this.

## Shape

`encode_decode_mutual`, fuel-parametric and mutual, mirroring `decode_encode_mutual` in the other direction — one induction on the shared budget carries both statements, since `decodeAux`/`decodeItems` both step it down by one.

⭐ **The `rest` generalisation is what makes it go through.** `decodeAux` returns a remainder, so the honest inductive statement is `encode item ++ rest = bs`; only the top-level wrapper specialises `rest := []`. The `decodeItems` case is then pure append-associativity over both IHs.

Corollaries: `encode_decodeAux`, `encode_decodeItems`, `encode_decode`, `encode_decodeFully`, `encode_decode_encode`, and ⭐ **`decode_injective`** — the dual of `encode_injective` and the form a transfer argument consumes: the decoder cannot map two distinct inputs to one item, so agreeing with the reference on decoded *items* pins agreement on *bytes*.

Note `encode_decode` needs **no length side condition**, unlike `decode_encode`'s `(encode item).length < 256 ^ 8`: that bound exists because encoding must fit the 8-byte length field, whereas here the input is *given* to decode successfully, so it's implied by the run rather than assumed.

## Where the canonicality actually lives — four checks, four negative controls

The theorem is true because of four decoder checks, and the proof is exactly the bookkeeping that each is enough. Since the interesting content is what the decoder **rejects**, each is a `by decide` control:

| check | control | why accepting would break it |
|---|---|---|
| single byte `< 0x80` must use the one-byte form | `decode [0x81, 1] = none` | `encode (.bytes [1]) = [1] ≠ [0x81, 1]` |
| long form must not encode length `≤ 55` | `decode [0xB8, 1, 0xFF] = none` | `encode` emits the short form |
| multi-byte length must not have a leading zero (`readLength`) | `decode ([0xB9, 0, 0x38] ++ replicate 56 7) = none` | length field wouldn't be minimal |
| list payload must be consumed exactly | `decode [0xC1, 0x83] = none` | payload overrun |

Drop any one and the theorem is **false**. Check 3 is what `Nat.toBytesBE_fromBytesBE_of_canonical` consumes. Also checked: empty list and empty string are distinct canonical encodings, and a nested list exercises the `decodeItems` half.

## Reuse

One new helper, `encodeBytes_long` — the `encodeBytes` counterpart of the existing `encode_list_long`; `Properties.lean` had `_nil`/`_single_small`/`_single_large`/`_pair`/`_triple` and the short generic form but no long one. Everything else is reused rather than re-proved: `takeBytes_eq_some_imp`, `readLength_eq_some_imp` (whose `0 < v → Nat.toBytesBE v = lenBytes` clause *is* the minimality fact), `encodeBytes_short_of_length_ne_one`, `encode_list_short`, `encode_list_long`.

Witnessed in the axiom gate (**156 witnesses**, up from 153): `encode_decode`, `encode_decode_mutual`, `decode_injective` — all `[propext, Classical.choice, Quot.sound]`, read back from the build log.

## Gates

`lake build` (4728 jobs, green, **0 warnings**) + check-no-warnings, check-progress, check-drift, check-forbidden-tactics, check-unimported, check-file-size, check-layering, check-heartbeats-approved, check-statement-tamper, check-obligation-blockers — **all pass**. **No `sorry`**, no `native_decide`/`bv_decide`, no `maxHeartbeats`.

### DIFF
--- a/EvmAsm/EL.lean
+++ b/EvmAsm/EL.lean
@@ -15,6 +15,7 @@ import EvmAsm.EL.RLP.Program
 import EvmAsm.EL.RLP.ProgramSpec
 import EvmAsm.EL.RLP.Properties
 import EvmAsm.EL.RLP.FuelMono
+import EvmAsm.EL.RLP.EncodeDecode
 import EvmAsm.EL.RLP.ReadLength
 import EvmAsm.EL.RLP.ReadLengthBridge
 import EvmAsm.EL.RLP.Scalar
--- a/EvmAsm/EL/RLP/EncodeDecode.lean
+++ b/EvmAsm/EL/RLP/EncodeDecode.lean
@@ -1,0 +1,425 @@
+/-
+  EvmAsm.EL.RLP.EncodeDecode
+
+  **Canonicality: `decode` only ever accepts an encoding** (GH #11896).
+
+      encode_decode : decode bs = some (item, []) → encode item = bs
+
+  ## Why this direction is the one that was missing
+
+  `Properties.lean` proves `decode_encode` — every encoding re-decodes to itself. That
+  says nothing about bytes that are *not* an encoding. This module proves the converse:
+  a successful decode identifies its input **as** the encoding of the decoded item, so
+  the decoder has no non-canonical accepts.
+
+  ⚠️ `encode_injective` is not a substitute: it relates two *encodings* to each other
+  (`encode i₁ = encode i₂ → i₁ = i₂`). The content here is about an arbitrary decodable
+  byte string.
+
+  ## What it unlocks
+
+  A whole class of `Correspondence` upgrades. The recurring `.machineOnly` reason is
+  "SpecRef carries ENCODE but no decoder, so there is nothing to bridge to". Composing
+  a SpecRef encoder with `decode_encode` yields a reference *decoder* — but only for
+  bytes already known to be an encoding. Transferring the differential for **arbitrary
+  input** is exactly this theorem. #11692 (`withdrawal_decode`) is the worked instance:
+  from a successful decode you can build the `Withdrawal` and get
+  `withdrawalToRlpItem w = .list [...]`; what remains is `bs = encode (.list [...])`,
+  i.e. this.
+
+  ## Where the canonicality actually lives
+
+  The theorem is true because `decodeAux`/`readLength` reject every non-minimal
+  encoding, and it is worth naming the four checks since the proof is exactly the
+  bookkeeping that each one is enough:
+
+  1. a single byte `< 0x80` **must** use the one-byte form — the short-string branch
+     returns `none` for `data = [b]` with `b < 0x80` (`Decode.lean:78`);
+  2. long forms **must not** encode a length `≤ 55` (`Decode.lean:85`, `:101`), so the
+     short form is mandatory below 56;
+  3. a multi-byte length field **must not** have a leading zero
+     (`readLength`, `Decode.lean:25`) — this is what makes the length field minimal,
+     and it is what `Nat.toBytesBE_fromBytesBE_of_canonical` needs;
+  4. a list payload must be consumed **exactly** (`leftover.isEmpty`).
+
+  Drop any one and the theorem is false. (2) is why `Nat.toBytesBE` of the recovered
+  length reproduces the original length bytes rather than a shorter field.
+
+  ## Shape
+
+  Mutual and fuel-parametric, mirroring `decode_encode_mutual` in the other direction:
+  one induction on the shared budget carries the `decodeAux` and `decodeItems`
+  statements together, since both step it down by one. The `rest` generalisation is
+  essential — `decodeAux` returns a remainder, so the honest statement is
+  `encode item ++ rest = bs`, and only the top-level wrapper specialises `rest := []`.
+-/
+
+import EvmAsm.EL.RLP.Properties
+
+namespace EvmAsm.EL.RLP
+
+/-- Long byte-string encoder shape (`payload > 55`), the `encodeBytes` counterpart of
+    the existing `encode_list_long`. Not present in `Properties.lean`, which has
+    `encodeBytes_nil` / `_single_small` / `_single_large` / `_pair` / `_triple` and the
+    short generic form but no long one. -/
+private theorem encodeBytes_long (data : List Byte) (h : 55 < data.length) :
+    encodeBytes data
+      = BitVec.ofNat 8 (0xB7 + (Nat.toBytesBE data.length).length)
+          :: (Nat.toBytesBE data.length ++ data) := by
+  match data with
+  | [] => simp at h
+  | [_] => exact absurd h (by simp)
+  | b :: c :: tl =>
+    simp only [encodeBytes]
+    rw [if_neg (Nat.not_le.mpr h)]
+    rfl
+
+/-- **Canonicality, mutual form.** A successful decode identifies its input as the
+    encoding of what it produced, followed by the returned remainder. -/
+theorem encode_decode_mutual : ∀ n : Nat,
+    (∀ (bs : List Byte) (item : RLPItem) (rest : List Byte),
+        decodeAux n bs = some (item, rest) → encode item ++ rest = bs)
+    ∧ (∀ (bs : List Byte) (items : List RLPItem) (rest : List Byte),
+        decodeItems n bs = some (items, rest) → encode.encodeItems items ++ rest = bs) := by
+  intro n
+  induction n with
+  | zero =>
+    refine ⟨?_, ?_⟩
+    · intro bs item rest hsome
+      rw [decodeAux_zero_fuel] at hsome
+      exact absurd hsome (by simp)
+    · intro bs items rest hsome
+      match bs with
+      | [] =>
+        simp only [decodeItems] at hsome
+        simp only [Option.some.injEq, Prod.mk.injEq] at hsome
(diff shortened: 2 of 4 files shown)


pirapira:
Reviewed. **Content accepted — labelling once `build` and `source-checks` finish** (they are still
running; `path-labels`, `velocity` and `size-label` are green).

I checked the load-bearing claims against `origin/main` rather than reading the PR body, since the
argument here is about what the decoder *rejects* and that is easy to assert and hard to see:

- **`decode_encode`'s length side condition is real and your asymmetry claim holds.**
  `Properties.lean:2289` `decode_encode_bytes` carries `hlen : data.length < 256 ^ 8`, and the docstring
  at `:2286` gives exactly the reason you cite — the 8-byte length field. So `encode_decode` needing no
  such hypothesis is a genuine difference in kind, not an oversight: the input is *given* to decode, so
  the bound is implied by the run.
- **`readLength_eq_some_imp` exists at `Properties.lean:238` and is reused, not re-proved**, and its
  `0 < v → Nat.toBytesBE v = lenBytes` clause is indeed the minimality fact check 3 needs.
- **The four negative controls are `by decide`** (`:411`, `:418`, `:428` plus the `0xC0`/`0x80`
  distinctness pair) — kernel-checked, no forbidden tactic anywhere in the diff, no `sorry`, no
  `maxHeartbeats`.
- **The three new axiom witnesses are actually wired**, not just claimed: `#print axioms` lines for
  `encode_decode`, `encode_decode_mutual` and `decode_injective` are added to
  `Progress/AxiomWitnesses.lean`. 153 → 156.

⭐ **The negative controls are what make this reviewable at all.** A canonicality theorem is a claim
about absent behaviour, and absent behaviour is exactly what a passing build cannot show — so
"drop any one of these four checks and the theorem is false", with a `decide` witness per check, is the
right way to present it. That table is the review.

⭐ **`decode_injective` is the corollary I expect to get used most**, and framing it as the dual of
`encode_injective` is right: agreeing with the reference on decoded *items* pins agreement on *bytes*,
which is the form a differential transfer actually consumes. Your note that `encode_injective` is **not**
a substitute — it relates two encodings, where the content here is an arbitrary decodable byte string —
is the distinction that makes #11692's residual discharge.

ⓘ On `check-axioms.sh` being unrunnable in your checkout: that is #11897, and **it is fixed** — the
portable-lock PR (#11909) landed, so `flock`/`sha256sum` are no longer hard requirements. Worth
re-running locally on your next branch; CI remains the authority either way.

Nothing to change. Holding only on CI.

dhsorens:
Thanks — and thanks especially for the #11897 heads-up, which was immediately actionable.

## `check-axioms.sh` now runs here, and I've stopped caveating it

Rebased my in-flight branch onto `main` (which now carries #11909 via the #11915 batch) and ran it for the first time on this machine:

```
==> acquired worktree build lock: …/evm-asm-build-8d2783d2d232e309.lock.d (mkdir fallback; flock missing)
check-axioms: OK — 153 witnesses; only classical axioms (propext, Classical.choice, Quot.sound),
  plus 0 grandfathered bv_decide/native_decide owner(s) in scripts/axiom-allow.txt.
exit=0
```

The `mkdir` fallback is what does it. So from here on the gate is in my run-list rather than my skipped-list — every PR I open after this reports a real local axiom result instead of deferring to CI. That closes the honest-but-annoying caveat I'd been repeating on every PR today, and it turned around in a few hours from filing, which I appreciate.

⚠️ One trap worth recording for whoever next debugs that script: `bash scripts/check-axioms.sh | tail -5; echo $?` reports **`tail`**'s status, not the script's. That is how I initially mis-read the flock failure as a *silent* pass rather than a loud one — the exit code was 2 all along, but the pipeline reported 0. `>/dev/null 2>&1; echo $?` is the only reliable form. I've corrected my own notes accordingly.

## On your review

Agreed on all four checked points, and on the framing: a canonicality theorem is a claim about **absent** behaviour, which a passing build cannot exhibit — so the four `by decide` controls are the review surface, not decoration. That is also why I stated "drop any one and the theorem is false" as a claim rather than a flourish: each control is the counterexample that would exist if the corresponding decoder check were removed.

`decode_injective` being the corollary that gets used most matches my expectation too — it is what #11692's residual actually needs, and I've noted there that its step 2 is now a composition rather than a blocker.

Nothing further from me on this PR.

Issue #10190, opened 2026-07-11
MSTORE topmost spec: migrate to evmMemoryIs interface + model guard band (like MLOAD #10180)


pirapira:
Sub-item of #53 (each EVM instruction should have a topmost spec over `IsEvmWord`/`evmMemoryIs` RISC-V memory, not raw per-machine-word cells). This is the **MSTORE analog** of the MLOAD region migration in #10180.

## Current state
`evm_mstore_stack_spec_within` (in `EvmAsm/Evm64/MStore/UnalignedFramedStackSpec.lean`) is still on the **old raw eight-cell form**: it takes explicit `loAddr*/hiAddr*/loVal*/hiVal*` witness words, a `start : Nat`, and four `mstoreLimbWindowOk` per-dword window predicates. It is classified `.proven` in `Progress.lean`, but via that raw-cell framing — not the `evmMemoryIs` region interface #53 asks for, and likely carrying the same aligned-only / over-constrained-precondition caveat MLOAD's old raw form had (where the unaligned public variant's overlapping-window precondition was effectively unsatisfiable).

By contrast, MLOAD (#10180) was migrated to a region-backed `evmMemoryIs` spec covering **every byte alignment** with the genuine `evmMemoryReadWord` post.

## Ask
Mirror the MLOAD migration for MSTORE:
1. Restate `evm_mstore_stack_spec_within` against `evmMemoryIs base capacity contents` (the EvmWord-level interface), covering all byte alignments, with the genuine post = the region after the 32-byte big-endian write at `offset` (`evmMemoryWriteWord`-style).
2. **Model the trailing guard band.** Per `CallFrameLayout.lean`, the EVM memory region (`frameMemBytes = 0x20000` at `frameMemOff = 0`) is immediately followed by the `frameStackGuardBytes = 512`-byte stack guard band (`frameStackGuardLoOff = frameMemOff + frameMemBytes`). MSTORE's dword-pair stores likely touch up to 8 bytes past `offset+32` at the region boundary, so the guard must be modeled to justify the boundary access and reduce the precondition to the natural `offset+32 ≤ capacity`.
3. Reach an **earned `.proven`** on the region interface (the guard discharges the tail unconditionally), not a relabel.

## MSTORE-specific care (differs from MLOAD)
MSTORE is a **write**, not a read. The pair mechanism must not *corrupt* the guard band or any byte outside `[offset, offset+32)`:
[...]

PR #11918, opened 2026-08-10
proof(mstore): mechanise the unaligned vacuity of evm_mstore_stack_spec_within

### LINKED ISSUE #11897: gate: check-axioms.sh cannot run on macOS — worktree-build-lock.sh needs flock + GNU sha256sum

### LINKED ISSUE #11913: registry: MSTORE is .proven but evm_mstore_stack_spec_within is VACUOUS for every unaligned offset (precondition unsatisfiable)
> 🤖 Filed by Claude Code while working #10190, 2026-08-10, `main` at `04de93895`. Surfaced by a subagent's analysis; **derivation below re-done and confirmed independently before filing.**

## Claim

`entry "MSTORE" .proven (some "evm_mstore_stack_spec_within")` — but that triple's precondition is **unsatisfiable whenever `offset % 8 ≠ 0`**, so the theorem is *vacuous* on every unaligned offset and has content only on the aligned eighth of its domain.

The rubric defines `.proven` as a complete triple "with **no input-domain precondition**". This is worse than a `.conditional` row: the restriction is not stated anywhere, and on the excluded region the statement is vacuously true rather than merely absent — which is the near-vacuity blind spot `coverRef` exists to catch.

## Derivation

`evm_mstore_stack_spec_within` (`Evm64/MStore/UnalignedFramedStackSpec.lean:1409`) separates **eight** memory cells in its precondition:

```lean
(loAddr0 ↦ₘ loVal0) ** (hiAddr0 ↦ₘ hiVal0) **
(loAddr1 ↦ₘ loVal1) ** (hiAddr1 ↦ₘ hiVal1) **
(loAddr2 ↦ₘ loVal2) ** (hiAddr2 ↦ₘ hiVal2) **
(loAddr3 ↦ₘ loVal3) ** (hiAddr3 ↦ₘ hiVal3)
```

and constrains them through four `mstoreLimbWindowOk … start …` hypotheses, one per limb, all sharing the same `start`. Each requires, for `i ∈ [0,8)`:

```lean
alignToDword (addrPtr + off_i) = MStore.mstoreDwordPairAddr loAddr hiAddr start i
byteOffset   (addrPtr + off_i) = (start + i) % 8
```

with (`ByteAlg.lean:88`)

```lean
def mstoreDwordPairAddr (loAddr hiAddr : Word) (start i : Nat) : Word :=
  if start + i < 8 then loAddr else hiAddr
```

The four limbs take byte-offset windows `0..7`, `8..15`, `16..23`, `24..31` (limb3…limb0 respectively, per the `h_window3 … h_window0` arguments). Offsets 8 apart share a dword residue, so all four limbs legitimately share one `start = byteOffset addrPtr`.

Let `D = alignToDword addrPtr`.

**Case `start = 0`.** Every `i < 8` satisfies `start + i < 8`, so all eight equations name `loAddr_k`, and **`hiAddr_k` is never constrained**. Choose the four `hiAddr` freely and distinct: satisfiable. ✅ This is where the theorem has content.

**Case `start = s`, `1 ≤ s ≤ 7`.** Bytes with `i < 8 - s` map to `loAddr_k`; the rest to `hiAddr_k`. Both are pinned:

| limb | first byte | `loAddr` | `hiAddr` |
|---|---|---|---|
| limb3 | `addrPtr + 0` | `D` | `D + 8` |
| limb2 | `addrPtr + 8` | `D + 8` | `D + 16` |
| limb1 | `addrPtr + 16` | `D + 16` | `D + 24` |
| limb0 | `addrPtr + 24` | `D + 24` | `D + 32` |

Hence **`hiAddr3 = loAddr2`**, **`hiAddr2 = loAddr1`**, **`hiAddr1 = loAddr0`**. Three pairs of the eight `**`-separated cells are the *same address*, and `memIs a v ** memIs a w` is unsatisfiable. ⛔ So the precondition is false and the triple says nothing.

Sanity: this is the arithmetic you'd expect — an unaligned 32-byte write touches **5** distinct dwords, not 8, so eight pairwise-disjoint cells cannot exist.

## Why this matters

- **The registry overstates coverage.** MSTORE reads as fully proven; it is proven on `offset % 8 = 0` only. The file name (`UnalignedFramedStackSpec`) actively suggests the opposite.
- **It is the caveat #10190 suspected**, now with a derivation rather than a hunch — which makes #10190 substantive rather than cosmetic. The `evmMemoryIs` migration is the fix, because an interface-level region assertion does not require the caller to name eight disjoint dwords.
- Nothing here says the **guest** is wrong. The emitted MSTORE is 32 individual `SB`s and never loads the destination, so unaligned writes are fine in the machine. The defect is in what the *proof* covers versus what the *row* claims.

## Suggested resolution

1. Either regrade MSTORE `.conditional` with `gate := "offset % 8 = 0 — the 8-cell precondition is unsatisfiable for unaligned offsets"`, **or** land #10190's `evmMemoryIs` re-statement and keep `.proven` honestly. The second is better and is already in progress (PR #11910 lands the bridge lemma `evmMemoryIs_quarter_pair_setBytes`).
2. Whichever is chosen, add a `coverRef` — a `…_precondition_reachable` lemma exhibiting a satisfying instance at `start = 0`. That is the anti-near-vacuity check the rubric asks for on restricted rows, and its absence is why this survived.
3. ⚠️ Worth a mechanized negative control: `example : ¬ (precondition …)` for a concrete `offset % 8 = 1`. I derived the collision from the statement and did **not** mechanize the unsatisfiability, so treat step 3 as the confirmation rather than this issue.

Related: #10190 (the migration that fixes it), PR #11910, #10180 (the MLOAD precedent).

dhsorens:
**Fixes #11913**, which I filed earlier today while working #10190. That issue derived the vacuity from the statement shape and explicitly deferred mechanising it ("⚠️ Worth a mechanized negative control… treat step 3 as the confirmation rather than this issue"). This is that confirmation, and both halves are machine-checked, so the argument has **no open link**.

MSTORE is registered `.proven` — which the rubric defines as a complete triple with **no input-domain precondition**. But `evm_mstore_stack_spec_within` separates **eight** memory cells while an unaligned 32-byte write touches only **five** distinct dwords, so the precondition is unsatisfiable and the theorem says nothing for any `offset % 8 ≠ 0`.

## Half 1 — the address collision, `decide`-checked at *every* unaligned residue

`mstoreDwordPairAddr … start i` routes byte `i` to `loAddr` when `start + i < 8`, else `hiAddr`; the four limb windows are 8 bytes apart. Hence `hiAddr3 = loAddr2`, `hiAddr2 = loAddr1`, `hiAddr1 = loAddr0`.

`alignToDword a = a &&& ~~~7#64` and `byteOffset a = (a &&& 7#64).toNat` are concrete `BitVec` ops, so this is checked at **every** `s ∈ 1..7` — the whole excluded region, not a sample — plus the limb2/limb1 and limb3/limb2 boundaries, plus a lemma confirming which cell each byte routes to at `start = 1`.

## Half 2 — the separation step, **proved**

```lean
theorem memIs_sepConj_same_addr_false (a v w : Word) :
    ¬ ∃ st, ((a ↦ₘ v) ** (a ↦ₘ w)) st
```

No state satisfies it, for any `v`/`w` **including `v = w`**: `sepConj` demands `∀ a, h1.mem a = none ∨ h2.mem a = none`, while both footprints are `singletonMem a`, which holds `some` at exactly `a`. Plus `memIs_sepConj_collide_false`, the form a reader of the precondition wants (the two cells already known equal).

## ⭐ The `s = 0` contrast — the part that keeps the finding from overclaiming

Also checked: at an **aligned** base, bytes 23 and 24 fall in *different* dwords, each limb sits inside one dword, and the four `hiAddr`s are left **unconstrained** by the window equations — so the eight cells *can* be chosen distinct.

That is why the triple has content on aligned offsets and none elsewhere, and why the row is **domain-restricted rather than simply wrong**. Without this case the report would be an overclaim in the opposite direction.

## Deliberately not done

No change to `evm_mstore_stack_spec_within`'s statement and **no registry regrade** — both are your call. The two routes:

1. regrade `.conditional` with the `offset % 8 = 0` gate named, plus a `coverRef` exhibiting a satisfying aligned instance; or
2. land #10190's `evmMemoryIs` re-statement and keep `.proven` honestly — an interface-level region assertion doesn't ask the caller to name eight disjoint dwords. **PR #11910 lands the bridge lemma for this, better, route.**

**No axiom witness added, and why:** `Progress.lean` is being edited by #11916 (labelled `merge!`), and you asked me to keep witness-block edits serialised after hand-resolving a conflict there. `memIs_sepConj_same_addr_false` is worth witnessing once #11916 lands — flagging rather than re-creating that collision.

## Gates

`lake build` (4729 jobs, green, **0 warnings**), check-no-warnings, **`check-axioms` — OK, 153 witnesses, classical axioms only, exit 0**, check-progress, check-drift, check-forbidden-tactics, check-unimported, check-file-size, check-layering, check-heartbeats-approved, check-statement-tamper, check-obligation-blockers — **all pass**.

⭐ First branch where `check-axioms.sh` actually **ran** here rather than being reported as skipped — your #11909 portable lock (mkdir + `shasum` fallbacks) fixed #11897 within hours of my filing it. Thanks; that retires a caveat I'd been repeating on every PR today.

No `sorry`/`admit`, no `native_decide`/`bv_decide` (plain `decide`, which the kernel's GMP-backed `Nat` handles fine on concrete 64-bit `BitVec`s), no `maxHeartbeats`. No guest bytes; `.text` unchanged.

### DIFF
--- a/EvmAsm/Evm64.lean
+++ b/EvmAsm/Evm64.lean
@@ -699,6 +699,7 @@ import EvmAsm.Evm64.MStore.Program
 import EvmAsm.Evm64.MStore.Spec
 import EvmAsm.Evm64.MStore.StackSpec
 import EvmAsm.Evm64.MStore.UnalignedFramedStackSpec
+import EvmAsm.Evm64.MStore.UnalignedVacuity
 import EvmAsm.Evm64.MStore.UnalignedStackSpec
 import EvmAsm.Evm64.MLoad.ByteAlg
 import EvmAsm.Evm64.MLoad.ByteWindow
--- a/EvmAsm/Evm64/MStore/UnalignedVacuity.lean
+++ b/EvmAsm/Evm64/MStore/UnalignedVacuity.lean
@@ -1,0 +1,184 @@
+/-
+  EvmAsm.Evm64.MStore.UnalignedVacuity
+
+  **`evm_mstore_stack_spec_within` is vacuous for every unaligned offset** (GH #11913).
+
+  MSTORE is registered `.proven`, which `Progress.lean`'s rubric defines as a complete
+  triple "with **no input-domain precondition**". But
+  `evm_mstore_stack_spec_within` (`MStore/UnalignedFramedStackSpec.lean:1409`)
+  separates **eight** memory cells:
+
+      (loAddr0 ↦ₘ loVal0) ** (hiAddr0 ↦ₘ hiVal0) ** (loAddr1 ↦ₘ loVal1) **
+      (hiAddr1 ↦ₘ hiVal1) ** (loAddr2 ↦ₘ loVal2) ** (hiAddr2 ↦ₘ hiVal2) **
+      (loAddr3 ↦ₘ loVal3) ** (hiAddr3 ↦ₘ hiVal3)
+
+  and constrains them via four `mstoreLimbWindowOk … start …` hypotheses sharing one
+  `start`. An unaligned 32-byte write touches **five** distinct dwords, not eight, so
+  eight pairwise-disjoint cells cannot exist — the precondition is unsatisfiable and the
+  theorem says nothing there. It has content only on `offset % 8 = 0`.
+
+  This module mechanises the arithmetic core of that argument, which #11913 flagged as
+  the outstanding confirmation step (the issue derived it from the statement shape but
+  did not machine-check it).
+
+  ## The collision
+
+  `mstoreDwordPairAddr loAddr hiAddr start i = if start + i < 8 then loAddr else hiAddr`
+  (`MStore/ByteAlg.lean:88`), and `mstoreLimbWindowOk` pins, for each byte `i` of a
+  limb's window, `alignToDword (addrPtr + off_i) = mstoreDwordPairAddr … start i`.
+
+  The four limbs take windows `0..7`, `8..15`, `16..23`, `24..31`. With `start = s ≠ 0`
+  each limb straddles two dwords, and the *high* dword of one limb is the *low* dword of
+  the next:
+
+  | limb | window | `loAddr` | `hiAddr` |
+  |---|---|---|---|
+  | limb3 | 0..7   | `D`      | `D + 8`  |
+  | limb2 | 8..15  | `D + 8`  | `D + 16` |
+  | limb1 | 16..23 | `D + 16` | `D + 24` |
+  | limb0 | 24..31 | `D + 24` | `D + 32` |
+
+  So `hiAddr3 = loAddr2`, `hiAddr2 = loAddr1`, `hiAddr1 = loAddr0` — three pairs of the
+  eight `**`-separated cells naming one address.
+
+  `alignToDword a = a &&& ~~~7#64` and `byteOffset a = (a &&& 7#64).toNat`
+  (`Rv64/Word.lean:209,212`) are concrete `BitVec` operations, so the collisions are
+  decidable at a concrete base. Below they are checked at every unaligned residue
+  `s ∈ 1..7`, which is the whole excluded region.
+
+  Both halves are here, so the argument has **no open link**:
+
+  * the **address collision** — `decide`-checked at every unaligned residue `s ∈ 1..7`,
+    with the contrasting `s = 0` case showing no collision (which is why the row is
+    domain-restricted rather than simply wrong);
+  * the **separation step** — `memIs_sepConj_same_addr_false`, *proved*: no state
+    satisfies `(a ↦ₘ v) ** (a ↦ₘ w)`, because `sepConj` demands
+    `∀ a, h1.mem a = none ∨ h2.mem a = none` while both footprints are
+    `singletonMem a`.
+
+  ⚠️ What is deliberately **not** done: rewriting
+  `evm_mstore_stack_spec_within`'s own statement, or regrading the registry row. Both
+  are the maintainer's call — see "What to do about it" below — and #11913 asked for the
+  confirmation, not the remedy.
+
+  ## What to do about it
+
+  Either regrade MSTORE `.conditional` with the `offset % 8 = 0` gate named, or land
+  #10190's `evmMemoryIs` re-statement and keep `.proven` honestly — an interface-level
+  region assertion does not ask the caller to name eight disjoint dwords. PR #11910
+  lands the bridge lemma for the second route, which is the better one.
+-/
+
+import EvmAsm.Evm64.MStore.ByteAlg
+import EvmAsm.Rv64.SepLogic
+
+namespace EvmAsm.Evm64.MStore
+
+open EvmAsm.Rv64
+
+/-! ## The dword collision at every unaligned residue
+
+    For a base whose `byteOffset` is `s ≠ 0`, byte `23` (the last byte of limb1's
+    window) and byte `24` (the first byte of limb0's window) lie in the **same** dword.
+    That is `hiAddr1 = loAddr0`.
+
+    Checked at a concrete base per residue rather than proved for a symbolic one: the
+    point is that the excluded region is nonempty and the collision is real, and
+    `alignToDword`/`byteOffset` are concrete `BitVec` operations, so `decide` settles it
+    without any TCB-expanding tactic. -/
+
+/-- `s = 1`: bytes 23 and 24 share a dword, so limb1's high cell IS limb0's low cell. -/
+example : byteOffset (0x40000001#64) = 1
+    ∧ alignToDword (0x40000001#64 + 23) = alignToDword (0x40000001#64 + 24) := by
+  refine ⟨?_, ?_⟩ <;> decide
+
+/-- `s = 2`. -/
+example : byteOffset (0x40000002#64) = 2
+    ∧ alignToDword (0x40000002#64 + 23) = alignToDword (0x40000002#64 + 24) := by
+  refine ⟨?_, ?_⟩ <;> decide
+
+/-- `s = 3`. -/
+example : byteOffset (0x40000003#64) = 3
+    ∧ alignToDword (0x40000003#64 + 23) = alignToDword (0x40000003#64 + 24) := by
+  refine ⟨?_, ?_⟩ <;> decide
+
+/-- `s = 4`. -/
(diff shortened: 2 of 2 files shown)


pirapira:
**Labelled.** Proof-only, no emitted bytes, no forbidden tactic anywhere in the diff (zero
`native_decide`/`bv_decide`/`sorry`; the controls are `example`s closed by `decide`, kernel-checked).

⭐ **This is the right way to report a vacuity, and the `s = 0` contrast is why.** A vacuity finding is a
claim that a theorem says *nothing*, which is unfalsifiable-sounding until you show where it *does* say
something. Checking that at an aligned base bytes 23 and 24 fall in different dwords and the four
`hiAddr`s are left unconstrained is what makes this **domain-restricted rather than simply wrong** — and
without it the report would have been an overclaim in the opposite direction. Both halves machine-checked,
so there is no open link in the argument.

Also right: `decide` at **every** residue `s ∈ 1..7` rather than a sample. That is the whole excluded
region, so the collision claim is exhaustive over the domain it concerns, not evidenced by instances.

## On the regrade — my recommendation, but it is the maintainer's call

**Route 1: regrade `.conditional` with the `offset % 8 = 0` gate named, plus a `coverRef` exhibiting a
satisfying aligned instance.** Reasons:

1. The rubric defines `.proven` as a complete triple with **no input-domain precondition**, and this one
   has an unsatisfiable precondition off the aligned residue. The grade as written asserts more than the
   theorem delivers, and **an invisible overclaim in a registry is worse than a missing row** — a census
   then reports coverage we do not have, and nobody re-derives it because the row looks settled.
2. `PROGRESS.md:269` **already hedges in prose** — *"aligned spec proven; unaligned `_public` variants in
   progress"*. So the restriction is half-disclosed today: the note is honest and the grade column is
   not. Route 1 makes them agree, which is a smaller change than it looks.
3. The `coverRef` is the part I care about most. A `.conditional` row without a satisfying instance is
   indistinguishable from a vacuous one — the same defect one level up. Exhibiting the aligned instance
   is what keeps the regrade from replacing an overclaim with an unfalsifiable hedge.

⚠️ What I would **not** do is weaken `evm_mstore_stack_spec_within`'s statement to make the grade true.
The theorem is fine; it is the label on it that is wrong.

ⓘ Related work in flight, so the pattern lands consistently: `chain_validate_gas_used_under_limit` is
being registered right now at an honest non-`.proven` grade for the same reason (complete machine triple,
`Fn.retSpecFlat` still owed per `registry-coverage-allow.txt:90`). Same principle, opposite direction —
one row understates, one overstates, and both get fixed by grading to the evidence.

PR #11922, opened 2026-08-10
proof(rlp): the additive strict item relation — route 2 without touching 52 modules

dhsorens:
Addresses #11898 — the scoped issue you asked for before anyone edits `rlpItemDecode`. Follows #11895 (#11795), which isolated the residual. I'm assigned on both.

## The problem, restated

`rlpItemDecode`'s two **list** arms require header canonicality plus a **span fit** and say nothing about the payload, while `decodeAux` additionally runs `decodeItems` and fails unless the payload is consumed exactly. So the machine relation is **weaker than the model** on list interiors — a strength mismatch no fuel work touches (#11711 closed the fuel half; this is the other one).

#11898 measured the cost of the obvious repair: **52 consuming modules**, with `RlpWalkDeterminism` and `WalkItemDeterminism` the real risk, since the relation being span-only may be load-bearing in a determinism argument.

## ⭐ Route 2, done additively — zero changes to those 52 modules

| declaration | what it does |
|---|---|
| `rlpItemDecodeStrict` | `rlpItemDecode` **conjoined** with the fact its list arms omit: on a list prefix, the payload decodes at budget `floor`, ending exactly at the item's span. First conjunct is the existing relation verbatim. |
| `rlpItemDecodeStrict_imp` | strict ⟹ weak, so **no existing consumer changes**. The 52 modules keep the weak relation; only the bridge consumes the strict one. |
| `rlpListInteriorsDecode_of_strict` | the strict relation discharges #11895's named residual |
| `rlpItemDecodeBridgesFrom_of_walkNextStrict` | composing with the already-proven byte-string half gives the full `RlpItemDecodeBridgesFrom` |

**That the discharge is near-immediate is the point, not a weakness.** It relocates the obligation from *"prove something false about the model"* to *"prove the routine establishes what it already checks"* — the difference between a blocked issue and a scheduled one. Before this, the gap read as a model-side impossibility.

## The residual is now one statement about the ROUTINE

`RlpWalkNextStrict`: `rlp_walk_next` accepts an item only when the strict relation holds.

After #11776 this should be **true** — `rlp_walk_next_core` recursively validates list payloads, requiring exact cursor-to-end exhaustion and rejecting malformed / truncated / non-canonical / nested-malformed / trailing interiors with status 7. The relation simply never recorded it, which is exactly your "the relation understates the guest" reading. Discharging it is the walker's triple, out of scope here.

## Deliberately not done

- **No edit to `rlpItemDecode`**, and no touch to any of the 52 consumers — that was the whole point of the additive shape, and it means the determinism proofs are untouched.
- **No registry regrade.** `rlp_list_count_items` stays `.machineOnly` until `RlpWalkNextStrict` is discharged; `.bridged` would claim a bridge resting on an undischarged hypothesis.
- No `sorry` — the open part is a hypothesis, matching the file's existing convention that nothing claims more than it has.

If you'd prefer the in-place edit after all, this is cheap to discard; the additive version is the one that needed no judgement call about the determinism proofs.

## Gates

`lake build` (4728 jobs, green, **0 warnings**), check-no-warnings, **`check-axioms` (exit 0 — running locally now, thanks to #11909)**, check-progress, check-drift, check-forbidden-tactics, check-unimported, check-file-size, check-layering, check-heartbeats-approved, check-statement-tamper, check-obligation-blockers — **all pass**.

### DIFF
--- a/EvmAsm/Codegen/Programs/RlpListCountItemsBridge.lean
+++ b/EvmAsm/Codegen/Programs/RlpListCountItemsBridge.lean
@@ -440,4 +440,103 @@ theorem rlpItemDecodeBridgesFrom_of_parts {bytes : List (BitVec 8)}
   · obtain ⟨inner, hdec⟩ := hlists off next len b hget hlt hitem
     exact ⟨.list inner, hdec⟩
 
+/-! ## The additive strict relation (GH #11898, route 2 without touching 52 modules)
+
+    #11795's residual is that `rlpItemDecode`'s two LIST arms require header canonicality
+    plus a **span fit** and say nothing about the payload, while `decodeAux` additionally
+    runs `decodeItems` and fails unless the payload is consumed exactly. So the machine
+    relation is **weaker than the model on list interiors**, and no fuel work fixes that
+    (#11711 closed the fuel half; this is the other one).
+
+    #11898 enumerated the cost of the obvious repair: `rlpItemDecode` has **52 consuming
+    modules**, and editing it in place risks the two determinism proofs
+    (`RlpWalkDeterminism`, `WalkItemDeterminism`) where the relation being span-only may
+    be load-bearing. @pirapira and I both read **route 2** — record what
+    `rlp_walk_next_core` actually verifies, since after #11776 the routine does more than
+    the relation admits — as the honest option, and a relation *weaker than the code* is
+    the shape that lets a false-accept hide.
+
+    ⭐ This is route 2 done **additively**, which is what makes it cheap: a new
+    `rlpItemDecodeStrict` that conjoins the missing payload fact, an implication back to
+    `rlpItemDecode` so **no existing consumer changes**, and the bridge discharge. The 52
+    modules keep the weak relation; only the bridge consumes the strict one.
+
+    ⚠️ What this does NOT do, and cannot: prove `rlp_walk_next` *satisfies* the strict
+    relation. That is a routine-side obligation needing the walker's triple, and it is
+    named below as `RlpWalkNextStrict` so the residual is one statement about the routine
+    rather than a false claim about the model. That relocation is the whole point — before
+    it, the gap looked like a model-side impossibility. -/
+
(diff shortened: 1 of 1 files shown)


pirapira:
**Labelled.** Route 2 additively is the right shape, and the reason is the sentence you put in yourself:
it relocates the obligation from *"prove something false about the model"* to *"prove the routine
establishes what it already checks"* — a blocked issue becoming a scheduled one. That is worth more than
the discharge being short.

⭐ **`rlpItemDecodeStrict_imp` is what makes this reviewable at 99 lines instead of 52 modules.** Strict ⟹
weak means the consuming modules keep the relation they were proved against, so the risk #11898 measured —
`RlpWalkDeterminism` / `WalkItemDeterminism` possibly *relying* on the relation being span-only — is not
taken at all rather than argued away. A strengthening that has to be defended across 52 consumers is a
different and much worse PR than a conjunction only the bridge consumes.

ⓘ **One thing I can now tell you about this exact file, from a measurement made twenty minutes ago:**
`Programs/RlpListCountItemsBridge.lean` was the sole codegen-touching change between `r20260810` and
current `main` (via #11895), and it was independently confirmed **byte-neutral** — the built guest ELF at
`04de93895` and at merge `6b03ab16c` is `08d1cd270a82360258e8d0fd1b54817e82ac880853c4ab13c2528388ae7a0ddb`,
identical to the tagged ELF, with the row passing at both. So additions to this module living under
`Codegen/Programs` do not reach the emitted image, and your +99 here inherits that precedent.

⚠️ That precedent is worth *checking* rather than assuming on any future change to this file, though, and
here is why I care: a proof module under `Codegen/Programs` that ever *did* move guest bytes would be a
defect in its own right, and it is exactly the kind of thing a `.text` sha comparison catches for free.

⭐ Naming the residual as one statement about the routine — `RlpWalkNextStrict`, that `rlp_walk_next`
accepts an item only when the strict relation holds — is the part that makes #11898 closable. An
obligation phrased about a routine has an owner and a method; one phrased about "the model being wrong"
has neither.

PR #11927, opened 2026-08-10
chore(asm): label 8 bare long-B epilogue heads (#11512)

### LINKED ISSUE #11512: codegen: 374 latent hardcoded long branches without named-target invariant
## Summary
Census after #11510 / PR #11511: **0** remaining mid-epilogue hazards (branches that skip `ld ra` into epilogue), but **374** hardcoded large-offset branches across **72** `EvmAsm/Codegen/Programs/*.lean` files still land on a `li a0,0` / first `ld ra` epilogue start without a named-target geometry guard.

These are not currently crashing (landing is correct *today*), but they are the same class as #11510: a body-length edit can silently retarget them mid-epilogue and corrupt control flow via stale `ra` without any gate failing.

## Criterion used in census
- Absolute B-type immediates with |off| large enough to cross routine structure
- Target is currently a valid fail/return epilogue entry
- No `brOff(namedTarget, pc)` / geometry `#guard` tying the immediate to a named index

## Proposed durable invariant
For any absolute branch beyond distance **N** (suggest N ≈ 64 bytes / 16 instr, or any branch that crosses a function-local label boundary):

```
require brOff(<named_target>, pc)
#guard 4*i + off = 4 * named_target_idx
```

So a hardcoded long branch cannot exist without naming what it lands on. Drift then fails the build instead of corrupting control flow.

## Non-goals tonight
Do **not** retarget the 374 in bulk without per-file geometry guards — that is the #11510 lesson. One instance fixed (#11510 gas gate); class needs the invariant.

## Related
- #11510 / PR #11511 (eip8037_tx_gas_gate mid-epilogue → stale-ra)
- No existing gate (asm-to-program, pins, Class-A ratchet) verifies branch targets are named join points

## Artifacts
Census from #11511 work: `/tmp/opencode/11510-fix/hazards.tsv` (post-fix mid-epilogue=0) and latent count 374/72 files.

pirapira:
## Summary

**PREP, not a fix.** Adds `.L`-local names at 8 previously unlabeled `li a0` epilogue heads so a later `brOff` retarget has named destinations. Does **not** address the #11512 retargeting hazard.

- **8 heads across 6 fixtures** (not 4): eip8037TxGasGate×2, headerExtract state/receipts/withdrawals×1 each, mptExtensionExtract×2, rlpFieldToU64×1.
- Diff is +8 label lines in `scripts/asm-fixtures/*.s` only.
- **No** `brOff` migration. **No** change to `EXPECTED_BARE_B_SITES` (still **872**).
- Does **not** close #11512 or any remainder of it.

## What this does and does not buy

| Claim | Status |
|---|---|
| Byte-neutral guest / Program emission | Yes — `emitProgram` strips internal `.L` labels; `assemble_cmp` text-identical on all 6; same mechanism as #11724 |
| Assembler / geometry gate checks these edges | **No** — branches remain bare `.+N` / bare BitVec-13 immediates |
| Retargeting hazard for body-length edits | **Unchanged** for all **872** bare long-B sites, **including the edges into these 8 heads** |
| `epi_li_a0_bare` 70→0 | Measures **naming only** (epilogue heads that lacked a label). **Not** protection. Exposure stays 872 |

Naming a destination gives a reader a handle and is a **prerequisite** for `brOff (entry+tgt) (entry+pc)`. It does not tie the branch immediate to that name.

## Labels added

| Fixture | Labels |
|---|---|
| `eip8037TxGasGateFunction` | `.Leip8037_tgg_st3`, `.Leip8037_tgg_ok` |
| `headerExtractStateRootFunction` | `.Lhesr_st1` |
| `headerExtractReceiptsRootFunction` | `.Lherr_st1` |
| `headerExtractWithdrawalsRootFunction` | `.Lhewr_st1` |
| `mptExtensionExtractFunction` | `.Lmee_st2`, `.Lmee_st1` |
| `rlpFieldToU64Function` | `.Lrfu_st1b` |

## Proof (prep only)

| Check | Result |
|---|---|
| `assemble_cmp` text-identical (6 fixtures) | PASS |
| `python3 scripts/asm_to_program.py check-all` | CLEAN; bare B ratchet **872** |
| Diff | +8 `.L` labels in fixtures only |

## Follow-on costed, not done (coord request)

Retarget the bare long-B edges whose destination PC is one of these 8 heads to `brOff(named)` in the Lean Programs (fixture labels already resolve; Lean still has bare BitVec-13).

| Quantity | Value |
|---|---:|
| Bare long-B sites in these 6 fixtures today | 79 |
| Of those, land on the 8 new heads | **70** (= the old `epi_li_a0_bare` 70) |
| Per-head absorb | tgg_ok 33, tgg_st3 2, hewr 18, herr 7, hesr 5, mee_st1 3, mee_st2 1, rfu_st1b 1 |
| Remaining bare long in these 6 (other targets) | 9 |
| Expected bytes | **Identical** — `brOff` reduces to the same BitVec-13 the literal already encodes |
| `EXPECTED_BARE_B_SITES` after head-only retarget | **872 → 802** |
| If full reconvert of all long bare in these 6 | 872 → 793 (extra −9) |

That follow-on is the first real invariant gain at zero byte cost and ratchets downward. Out of scope for this PR on purpose.

## Out of scope

- Any `brOff` / Program source edit
- Changing `EXPECTED_BARE_B_SITES`
- Closing #11512


pirapira:
Holding on **the PR body, not the diff.** The change itself is fine and I accept the byte-neutrality
argument — `emitProgramR`-converted routines, `emitProgram` strips internal `.L` labels, `assemble_cmp`
text-identical, `check-all` clean with the ratchet still at 872, same mechanism the #11724 rename
established. Eight labels, zero emitted bytes. No objection to any of that.

## ⛔ "Closes the actionable remainder of #11512" is not what this does

The body states plainly, two bullets later, that **branches remain bare `.+N` immediates — no `brOff`
migration**. So a body-length edit can still silently retarget every one of them, including these eight.
**That is the entire hazard #11512 was filed about.** Naming a destination gives a reader something to
look at; it gives the assembler nothing to check. The issue's proposed invariant was
`brOff(<named_target>, pc)` plus a guard tying the immediate to the named index, and bare `.+N` satisfies
neither.

⚠️ **And the metric makes it worse, not better:** `epi_li_a0_bare` **70 → 0** reads as a class
eliminated. What was eliminated is *"epilogue heads that lack a name"* — a documentation property. The
retargeting exposure is unchanged at every one of the 872 sites. A future reader who sees this row at
zero will reasonably conclude the epilogue-head class is handled.

**What I want changed — body only, no code:**

1. Retitle the intent as **prep**: labeling is a *prerequisite* for `brOff` (you cannot reference a
   target that has no name), not a fix. Drop "closes the actionable remainder".
2. State explicitly that **the retargeting hazard remains for all 872 sites, these 8 included**, and that
   `epi_li_a0_bare = 0` measures naming, not protection.
3. Fix the fixture count: the body says **8 unique heads in 4 converted fixtures**, then lists **six**
   fixtures, six changed files, and "text-identical on all 6". It is 8 heads across 6 fixtures.

## The follow-on worth costing (not in this PR)

Retarget those 8 to `brOff(named)`. Expected: **emitted bytes identical** — `brOff` computes the offset
the literal already encodes — while `EXPECTED_BARE_B_SITES` **drops below 872**. That is a real invariant
at zero byte cost, and it ratchets *downward*, which is strictly better than freezing the class where it
is. "Would move the ratchet" is listed as a reason it is out of scope; moving the ratchet **down** is the
goal, not a side effect.

⭐ To be clear about why I am not rejecting this: an eight-line byte-neutral prep step is exactly the
right size of first move. It is only the claim attached to it that would mislead.

PR #11928, opened 2026-08-10
fix(progress): register chain_validate_gas_used_under_limit as .proven

pirapira:
## Summary

Registers `chain_validate_gas_used_under_limit` as `.proven` in `EvmAsm/Progress/Routines.lean`, matching its twin `chain_validate_consecutive_numbers`/`chain_validate_increasing_timestamps`, and drains the tier-B allowlist entry that miscategorised it.

This is part of #11575 (register the 8 claimed "mechanical-fork" chain/header siblings). This PR covers **one** sibling only.

## Why `.proven`, not `.conditional`, and draining the allowlist

The top-level triple `chain_validate_gas_used_under_limit_spec_within` is a **direct `cpsTripleWithin`** over `fullCode = cvgulCode.union RlpFieldToU64SAsm.code`, structurally identical to the twin registered at line 270. There is no `Fn`/`Fn.retSpecFlat` SAsm structure in either the loop-close or loop files. The hypothesis set is ABI-only (`hspC` frame base, `hret` ret alignment, `hnWord` definitional, `hN` a 64-bit length bound, and the six `hAll*` per-header facts: align, len, salign, slack, over, `isValidByteAccess`). **No input-domain gate anywhere** — the three-way post is total over the header list. So the allowlist line that claimed "needs `Fn.retSpecFlat` before a `.proven` row is honest (#11637)" was a **miscategorisation**: grading this `.conditional` would have no gate to name, and leaving it tier B would put two structurally identical triples at different grades.

## Changes

- `EvmAsm/Progress/Routines.lean`: import `ChainValidateGasUsedUnderLimitLoopClose`, add the `.proven` row with a justification note, add the private witness abbrev, and bump the three count theorems:
  - `routineCount` **35 → 36**
  - `routineCountTier .proven` **26 → 27**
  - `routineSymbols.length` **25 → 26**
- `EvmAsm/Progress/AxiomWitnesses.lean`: regenerated via `scripts/gen-axiom-witnesses.py --write` — now **154 witnesses**, including the new `#print axioms` line.
- `scripts/registry-coverage-allow.txt`: drained the `chain_validate_gas_used_under_limit` entry (line 90).

## Gate status — please read

- `lake build` of `EvmAsm.Progress.Routines`: **clean** (2459 jobs).
- `check-registry-coverage.py`: **OK**.
- `check-progress.sh` / `check-drift.sh`: **exit 0**.
- `check-axioms.sh` **could NOT run in this environment** (artifact-cache mode — `lake env lean` cannot resolve cache-satisfied modules, issue #10537). I substituted a direct `lean_verify` of the triple, which passes with exactly the three classical axioms (`propext`, `Classical.choice`, `Quot.sound`) and 0 sorries. **CI is the authority and will run the real gate.**

## Untouched siblings (explicit non-claim)

Allowlist lines 87–89 still mark `chain_validate_blob_gas_used_multiple`, `chain_validate_blob_gas_used_under_max`, and `chain_validate_extra_data_length` as tier B claiming they "need `Fn.retSpecFlat`". If those triples take the same direct `cpsTripleWithin` form as this one they are **likely** miscategorised the same way — but I did **not** verify and did **not** touch them. That is an unverified suspicion, not a finding.


pirapira:
**Labelled.** I verified the load-bearing claim independently before accepting it, because this PR
overrides an instruction I had given its author:

- both `ChainValidateGasUsedUnderLimitLoopClose.lean` and `ChainValidateConsecutiveNumbersLoopClose.lean`
  contain **zero** occurrences of `retSpecFlat` or an SAsm `Fn` structure;
- the theorem at `:1283` is a direct `cpsTripleWithin` over `fullCode`, and the twin's is the same shape
  at `:877`;
- the hypothesis set is ABI-only as described — `hspC`, `hret`, `hnWord`, `hN : lengths.length < 2 ^ 64`,
  and the six `hAll*` facts — with **no input-domain gate anywhere**.

So `.proven` is honest, the twin's grading note at `Routines.lean:258-269` applies verbatim, and the
allowlist entry claiming `Fn.retSpecFlat` was a **miscategorisation**. Draining it is correct: an
allowlist row asserting a routine needs work it does not need is the same invisible wrongness the
registry exists to prevent, one level up.

⭐ **The reason I trust this PR is that its author refused both wrong answers and stopped.** `.proven` was
forbidden by my instruction; `.conditional` had no gate to name. Rather than pick the one that would pass
review, it reported the contradiction — and the contradiction was the finding, because **two structurally
identical triples graded differently is worse than either grade being wrong**: it makes the registry
unreadable and teaches the next reader that grades are arbitrary.

⭐ Both honesty caveats are carried where a reader meets them, which is why the evidence is usable:
`check-axioms.sh` **could not run** in that environment (#10537, artifact-cache mode) with `lean_verify`
substituted and **CI named as the authority** — so nobody infers the repo gate ran; and allowlist lines
87-89 are flagged as *likely* miscategorised the same way, explicitly **unverified and untouched**, as
*"an unverified suspicion, not a finding."*

ⓘ The count bumps are stated for checking rather than asserted — `routineCount` 35→36, proven 26→27,
`routineSymbols` 25→26, witnesses 154 — and `check-registry-coverage.py`, `check-progress.sh` and
`check-drift.sh` all pass. The underlying triple predates this PR entirely; nothing here proves anything
new, it makes proven work **visible**, which is the #11637 row-existence class.

PR #11900, opened 2026-08-10
feat(layout): relocate tranche-B arenas for #11186

pirapira:
## Summary
Closes the remaining #11186 tranche-B capacity raise (~19.71 MiB): block reads → 66666 and log arenas → protocol bounds, via layout relocation (not in-place growth).

### Geometry (production guest)
| item | value |
|---|---|
| `.data` | `0xa0b00000` |
| `.bss` | `0xa0b70000` size `0x1b38b8a0` end `0xbbefb8a0` |
| logs HEAD | event 44707×256, data 2095652, meta 44707×16 |
| scheme-A pack | SR/AR/CR@66666 + tx reads + SW/TSW through `0xa2e068c0` |
| main body | starts `0xa2e07000` (no legacy `0xa3110000` pin) |
| high pack | AW `0xbdb80000` / AU `0xbe380000` / TX_AW `0xbf780000` / SSZ keep |
| undo | derived `0xbbf5e000`, headroom `0x28d180` |
| OUTPUT / transient | KEEP ABI |

### Why not in-place
Log raise in old mid-bss floated undo into ACCOUNT_WRITES; high pack already SSZ-tight. Reclaim aspirational EVM_MEMORY hole into `.bss` HEAD + drop main-body base pin saves the envelope.

### Gates
- four-artefact regen fixpoint: NO-OP
- `check-asm-to-program`: CLEAN
- `check-region-map`: OK (bases, union, link-pins)
- `check-layout-residual-literals`: OK (new)
- r200 seed=932419178 spike: base/cand FA0 FR0 full200; joined198 **byte-identical**; base-PASS→cand-FR=∅

### ELFs
- base@610e4960d `7697c7ef…`
- cand `fada9ea4…`

Artefacts: `/tmp/grok-11186-impl/`

## Test plan
- [x] lake build codegen
- [x] four-artefact fixpoint
- [x] check-region-map / check-asm-to-program / residual
- [x] r200 A/B vs r20260810-2 base
- [ ] CI full suite

PR #11929, opened 2026-08-10
Reject transaction type 0xff at dispatch boundary

pirapira:
## Summary

Reject transaction prefix `0xff` at the transaction-type dispatch boundary.
The existing `0xc0..0xff` legacy classification is narrowed to the valid
legacy range `0xc0..0xfe`; `0xff` now takes the dispatch failure path while
the typed transaction cases remain unchanged. The new long branch uses
`brOff` to the existing fail tail, so the bare local-B ratchet remains exactly
872. The source-drift checker now accepts this legitimate site-level
named/bare relocation mix while still validating the `GuestAddrs` base and
computed offset.

## Validation

- Base: `5b2fc81140bbfb1c5d20aa67a7b5bbc96c123de7`
- Candidate: `f0400a9f3f60f8c70775d27c61a768495b864dbe`
- `repo_dirty=0` after the candidate commit; base and candidate guests were
  measured from those trees.
- Fixture tag: `tests-zkevm@v0.6.2`, random seed `4242`, Spike, 8 workers.
- Differential r200: both legs `200/200` `OK`; FA set `[]`, FR set `[]`,
  errored `0`, budget exhaustion `0`. There were zero status differences,
  zero output-match differences, and the explicit base-PASS -> candidate-FR
  set is `[]`. The 200 selected inputs were unique.
- The valid r200 population cannot exercise this path: no valid corpus row
  carries a transaction prefix `0xff`. Its clean result is therefore the
  expected no-regression/null result, not correctness evidence for the new
  guard.
- Production guest ELF SHA-256: base
  `fada9ea40962b0dbae36a1347bae7e36be5cc9d4a1d7df3efac53dffb72a399d`;
  candidate
  `9bff29bfce560043485b19accf4492d2cc97a36acb4061ebc8abcafcc810ceb4`.
- Structure delta against the current base is exactly the appended 12-byte
  guard: `.text`/`textSizeBytes` is `0x53348 -> 0x53354`; 674 symbols move
  by `+0xc`, 2335 common symbols are unchanged, and there are no other
  symbol deltas. `.data` (`0x5310`), `.bss` (`0x1b38b8a0`), and
  `stateGasDiagSizeBytes` (`0x61a78`) are unchanged. No symbols were added or
  removed. `GuestImageEntries.lean` is byte-identical.
- Gates: `lake build`; `check-no-warnings`; `check-asm-to-program` (bare
  local-B ratchet exactly `872`); `check-file-size`; `check-forbidden-tactics`;
  `check-layering`; `check-unimported`; `check-heartbeats-approved`;
  `check-no-hardcoded-guest-pc`; `check-drift`; `check-progress`; and
  `check-roundtrip-coverage` all pass. `check-axioms.sh` was not run because
  this checkout has `LAKE_ARTIFACT_CACHE=true`, under which the project guide
  says its `lake env lean` invocation cannot resolve cache-only oleans.

### Positive control

The unchanged exact row `0cd3bd9af8344ab2b9eadcab6bf89d47791fe83febf18eaa3addd734f560d47b`
(label `00027`, manifest row 28) passes on both current production guests:
`succ=1`.

For the discriminating control, I used that exact row, changed only the first
transaction byte from the checked-in artifact's `0x04` to `0xff`, and
recomputed the payload `block_hash` with execution-specs. The resulting
framed control input SHA-256 is
`984b82bb562b65db7ce5bf7e9e88666cd68373612aff41057b89b1909cddc8e1`.
Production `succ=0` on both legs. The diagnostic probes report:

- base `5b2fc8114`: `verdict=0`, `bv_fail=52`;
- candidate: `verdict=0`, `bv_fail=27`.

The current base and candidate diagnostic ELF SHAs are respectively
`37cf0af3f5dccfec99cf355d784fc76e2a83f1f921d56d5b3591ff16b6871a7d` and
`3f35116285d65777e2bfaf003ce53b53b313b002e2870d9591e052516ef949ab`.
Code `27` is the caller's outer versioned-hashes failure code, not a dedicated
0xff code. It is reached because
`ssz_tx_list_versioned_hashes_match` dispatches through `tx_type_dispatch`
before `tx_eip4844_decode` and before any RLP walker; the new guard therefore
changes the rejecting path at the type gate. Base code `52` is the old
legacy-classification path reaching the later public-key-sender failure.

The prior relay described this control as `f861800a8407270e ->
ff61800a8407270e`; that byte sequence is not present in the available exact
`0cd3...` artifact. The reproducible current control above is the same row
identity with the actual first byte changed `04 -> ff`, and it demonstrates
the same `52 -> 27` gate discrimination.

The r200 population is a no-regression sample; this positive control is the
correctness evidence that the new guard fires.

The generated pins are regenerated after rebasing onto current `main`
(`5b2fc8114`) and are committed with the source change. The stale repin
`a54a20009` is not part of this branch.


pirapira:
Parking with `complicated-merge` — two separate things, and the first is **not** your bug.

**1. Stale pins after main moved (needs a rebase + pin regeneration).** Your branch head builds perfectly on its own:

```
$ git checkout origin/codex2/11057-message-seam-rebased   # d2abd8c0d, unmerged
$ LAKE_ARTIFACT_CACHE=false lake build
  -> exit 0, 0 errors
```

But it is **32 commits behind main**, and merging main in produces 33 build errors of the repin family:

```
EvmAsm/Codegen/Programs/Bn254FieldMulModSAsm.lean:91:49: Type mismatch
EvmAsm/Codegen/Programs/Bls12G1LtPSAsm.lean:289:55: Tactic `decide` proved that the proposition ...
EvmAsm/Codegen/Programs/Secp256k1FieldCmpPSAsm.lean:354:55: ... is false
EvmAsm/Codegen/RegionMap.lean:733:64: ...
EvmAsm/Codegen/Programs/StorageWriteMap.lean:133:0: Expression ...
```

Your `d2abd8c0d chore(codegen): regenerate guest pins after tx dispatch guard` regenerated pins against the main you branched from. Since then **#11900 landed the tranche-B arena relocation**, which moves the layout again, so those pins are stale. I cannot repair this from the merge side: the repinned literals and the generated addresses have to move together, and `GuestAddrs` only regenerates *after* a successful build — so taking either side of the generated files deadlocks. It needs your regeneration re-run on top of current main.

**2. A ratchet growth that is yours, and is independent of the rebase.** On the unmerged branch head:

```
DRIFT DETECTED:
  bare local B site ratchet: expected exactly 872 sites, found 873 (97 files / 148 defs);
  update the committed value only with the corresponding source migration
```

That is +1 in the **growth** direction, which the gate only permits alongside a source migration. For contrast, #11512/#11927 labelled 8 such heads and landed at 872 cleanly. Please either avoid the new bare B site or land the corresponding migration with it — I will not raise the committed value to make the gate pass.

#11928 was batched with this and is unaffected; I am landing it on its own now. Clear `complicated-merge` once the branch is rebased with fresh pins and the B ratchet is back at 872, and I will pick it up.

pirapira:
Re-verified after `f0400a9f3 fix: name tx dispatch guard relocation` — both issues are resolved and auto-merge is armed.

**1. The pins.** This is the check that matters, because the branch built clean on its own last time too — the staleness only appears after merging main. Merged onto current main (35 behind -> 0, no conflicts):

```
$ LAKE_ARTIFACT_CACHE=false lake build
  -> exit 0, 0 errors     (previously: 33 repin-family errors)
```

So the rebase genuinely carried fresh pins past #11900's arena relocation.

**2. The B-site ratchet is back to 872**, and you fixed it the right way — naming the guard rather than raising the committed value:

```
check-all: CLEAN (371 converted defs across 138 files;
  blocking J ratchet 175 sites; blocking B ratchet 872 sites)
```

Full aggregate sweep on the merged tree:

```
==> codegen: PASS (steps=6/6, skips=0)
==> guestaddrs-starts: PASS (steps=1/1, skips=0)
==> asm-to-program: PASS (steps=1/1, skips=0)
==> reports: PASS (steps=4/4, skips=0)
==> axioms: PASS (steps=1/1, skips=0)
==> arithmetic-fuzz: PASS (steps=1/1, skips=0)
-> EXIT=0
```

`lake build` green (4733 jobs, twice across the regen fixpoint).

Worth recording for anyone hitting this next: any branch that regenerated guest pins **before** #11900 landed will build clean solo and then fail with `decide`-proved-false and type-mismatch errors in `Bn254Field*`/`Bls12*`/`Secp256k1*`/`RegionMap` once main is merged in. It cannot be repaired from the merge side — the repinned literals and generated addresses have to move together, and `GuestAddrs` only regenerates after a successful build. Rebase + regenerate is the fix, exactly as you did here.

Issue #11577, opened 2026-08-05
proof: whole-routine spec for rlp_item_span — the named .unproven blocker excluding two otherwise-startable routines from the leaf table


pirapira:
> 🤖 Generated by Claude Code (strategist pass over execution-specs ↔ SpecRef ↔ RISC-V correspondence, 2026-08-06).

## The move

`rlp_item_span` is one of only three routines carrying verdict `.unproven` in `EvmAsm/Progress/Correspondence.lean` (`:258`), and it is the *named blocker* in `docs/leaf-routine-targets.md`'s rule-firing paragraph:

> `eip8037_state_used_before_tx` and `runtime_same_block_delegation_code` *look* startable (every fixture-level callee verified) but transitively call `rlp_item_span`, which is `.unproven` — both excluded.

So one triple unblocks two otherwise-startable routines, and it lives in the family with the deepest existing toolkit: 19 rlp rows in the Correspondence registry, walk/size/nth-item specs already `.proven` in `Routines.lean`, and `rlpItemRegionFrom` for the typed phrasing.

## Ask

A whole-routine `cpsTripleWithin` spec for `rlp_item_span` against the span computation implied by SpecRef's RLP decode (`decode_to_item`-side prefix arithmetic — pick the exact SpecRef anchor from `docs/rlp-spec-correspondence.md`'s vocabulary, and grade the row with the same `Verdict`/`Basis` discipline as the other 19).

- If totality is awkward (malformed prefixes), `.conditional` with a stated input-domain gate is fine — `rlp_item_size` (`Routines.lean:158`) is the precedent, and the registry requires the gate be stated, not implied.
- Not in #10780's scope (that issue owns the **encode** side and the `rlp_item_size`/`rlp_encode_list_prefix` long forms); this is the decode-side span.

## Follow-on (do not fold in)

Once landed, `eip8037_state_used_before_tx` and `runtime_same_block_delegation_code` become startable rows — re-run the leaf-targets criterion and file them then, rather than pre-filing against an unproven premise.

pirapira:
## Leverage of `.conditional` (coord checkpoint answer)

**`.conditional` DOES lift the leaf-table exclusion.**

Evidence at current main:

[...]

PR #11936, opened 2026-08-10
proof(rlp): whole-routine rlp_item_span under short-list+WalkedSpanForm (#11577)

pirapira:
## Summary
Whole-routine `cpsTripleWithin` for `rlp_item_span` under a `.conditional` domain (short-list outer header + `WalkedSpanForm` on every walked prefix 0..i), with mandatory `coverRef`.

Also retargets the registry cross-check **self-test canary** after this PR lifts `rlp_item_span` off `.unproven` (that is a success of the proof, not a regression of the gate).

## Domain / leverage
- Gate: short-list outer (`payloadLen ≤ 55`) + `WalkedSpanForm 0..i` (every walked item is SpanForm). Fail arms unreachable under the gate.
- `coverRef`: `rlp_item_span_precondition_reachable` on `encode(.list [.bytes []]) = [0xc1, 0x80]`, i=0.
- **Leverage:** `.conditional` DOES lift leaf-routine-targets exclusion (verified set includes `.proven` and `.conditional`). Callers inherit the SpanForm domain. Issue comment: https://github.com/Verified-zkEVM/evm-asm/issues/11577#issuecomment-5240750389

## Files
- `RlpItemSpanSizeOffset.lean` — offset size callee (listBase%8=0, LBU at off)
- `RlpItemSpanMachine.lean` — frame/body geometry, pure helpers, coverRef
- `RlpItemSpanLoop.lean` — loop continue machinery (split under file-size cap)
- `RlpItemSpanBody.lean` — setup/header/exit/compose + `rlp_item_span_spec_within` via `abi_frame`
- Registry: Correspondence → `domainRestricted`/`machineOnly`; Routines +.conditional + witness; negative control moved to `rlp_encode_u64`

## Canary retarget (handback fix)
`scripts/check_registry_crosscheck.py` `CANARY`:
- **Was** `rlp_item_span` (this PR proved it → `--self-test` only fails; plain run still green).
- **Now** `bal_canonical_sort`.
- **Not** `rlp_encode_u64`: Routines.lean kernel negative control already points there; both controls on one row would break two gates when that row is proven.
- **Why bal_canonical_sort is the safer long-lived canary:** Correspondence note records a triple IS statable today and the row stays `.unproven` because nobody has stated one; headline obligation is **permutation** (a sort that silently drops rows is still sorted; e2e hash against declared rows cannot see it). Predicate gap nobody closes by accident.
- Open-PR check before commit: none of the open PRs touch `bal_canonical_sort` (titles + name-only diffs).

## Gates
- `check-registry-crosscheck.sh` plain OK
- `check-registry-crosscheck.sh --self-test` OK (fires on injected bal_canonical_sort)
- `check-file-size.sh` OK
- source-checks set (unimported, layering, forbidden tactics, roundtrip, heartbeats, hardcoded PC, opcode structure) OK
- classical axioms only; no guest edits; no sorry/native_decide/bv_decide

## Does not claim
- Full long-list / non-SpanForm coverage
- Closing every leaf that listed span as blocker without composing under the inherited domain


pirapira:
**Labelled.** Four things checked independently, because three of them are claims I had explicitly asked
to be justified rather than asserted.

**1. `.conditional` really does lift the leaf exclusion.** I doubted this and said so when I approved the
plan — the doc excludes `eip8037_state_used_before_tx` / `runtime_same_block_delegation_code` because
`rlp_item_span` is `.unproven`, and `.conditional` is not `.proven`. But `docs/leaf-routine-targets.md:26-28`
defines the **verified set** as *"the symbols with witnessed rows in `Routines.lean`"* — currently 28 rows,
**20 `.proven` and 8 `.conditional`**. So a `.conditional` row joins the verified set and the exclusion
lifts. The leverage claim I downgraded is restored, on the doc's own definition.

⚠️ With the caveat this PR states rather than hides: the callers become startable and **inherit the
SpanForm domain**. Startable-under-a-domain is not startable-unconditionally, and whoever picks up those
two routines needs to carry the gate.

**2. The negative control was re-pointed, not deleted.** The kernel-checked control at `Routines.lean`
asserted `crossVerdictOk ("rlp_item_span" :: routineSymbols) = false`; witnessing `rlp_item_span` would
have made it fire. The diff swaps it to **`rlp_encode_u64`** — another genuinely `.unproven`
Correspondence row — so the gate keeps a live counterexample instead of being quietly dropped. That is the
right move: the control's own docstring says a gate nobody has seen fail is indistinguishable from one
that cannot.

**3. The `coverRef` exhibits a concrete satisfying instance**, which is what stops a `.conditional` row
being vacuous one level up: `rlp_item_span_precondition_reachable` proves
`encode (.list [.bytes []]) = [0xc1, 0x80]` at index 0 satisfies `payloadLen ≤ 55`, `i < length`, and
`WalkedSpanForm`. A gate with no exhibited instance is the same defect as the MSTORE row (#11913/#11918),
one level up — that is now closed off.

**4. No forbidden tactic** anywhere in the diff: zero `sorry` / `native_decide` / `bv_decide`.

⭐ And the checkpoint discipline paid for itself here. The pre-work checkpoint found the `size` callee does
**not** drop in as-is — `SpanForm` short forms only, `ptr % 8 = 0` required while the loop cursor is
unaligned after item 0 — and eliminated two plausible shortcuts with reasons (`nth_item` has a different
ABI, content-for-strings vs full-span; `walk_next` models a sequential join, not a list-index span). That
is why this landed as an honest `.conditional` with a named gate instead of an over-claimed `.proven`.

pirapira:
Synced onto main and verified. One conflict needed judgement — flagging it since I made the edit.

`EvmAsm/Progress/Routines.lean` conflicted on the kernel-checked census theorems. **Neither side was correct for the merged tree**: your branch had 36/26/10 with 26 symbols, main had 43/34/9 with 33. So rather than picking a side, I parsed the merged registry rows and derived the counts:

```
total rows      : 44
tier .proven    : 34
tier .conditional: 10
distinct symbols: 34
```

which reconciles exactly as main's numbers plus this PR's one new `.conditional` row for `rlp_item_span` (43+1, 9+1, 33+1). Resolved to:

```lean
theorem routineCount_eq : routineCount = 44 := by decide
theorem routineProvenCount_eq      : routineCountTier .proven      = 34 := by decide
theorem routineConditionalCount_eq : routineCountTier .conditional = 10 := by decide
theorem routineSymbols_eq : routineSymbols.length = 34 := by decide
```

These are `by decide`, so the kernel is the check — a wrong count would fail to elaborate. The build passing is what confirms the numbers, not my arithmetic. `PROGRESS.md` took main's copy and was then regenerated.

No statement or definition was changed; only the literals a merge has to reconcile.

Verified on the pushed head with `scripts/check-build-parallel.sh` and `LAKE_ARTIFACT_CACHE=false`: `lake build` green (4738 jobs, twice across the regen fixpoint) and all six gate groups PASS (EXIT=0), tree clean. Auto-merge armed.

pirapira:
CI failed on the head I pushed — re-parking with `complicated-merge`. Auto-merge was disarmed first, so nothing lands on a re-run.

`check-file-size` fails:

```
FAIL    1654 / 1500 lines  EvmAsm/Codegen/Programs/RlpItemSpanBody.lean  [codegen]
```

`origin/main` passes the same gate, so this is the new file's size. The gate is explicit that per-file exception markers are not supported — the fix is splitting: extract a cluster of defs into a sibling submodule and import it back (it points at `Bls12MapG1Real.lean` / `ChildFrameHandlerTails.lean` as the pattern). Which cluster to extract is a design call about this module, so I have left it to you. #11831 hit the same gate recently and was fixed that way.

The census resolution I made is unaffected and still correct — `routineCount = 44`, `.proven = 34`, `.conditional = 10`, `routineSymbols.length = 34`, all kernel-checked by `decide` and confirmed by a clean build (4738 jobs). Only the file size blocks it.

**This is my fault for not catching it before arming, and it is the second time I have shipped a gap of this shape.** I had switched my local verification to `scripts/check-build-parallel.sh` after a hand-rolled gate list missed `check-memorylayout-region-coverage` — but that aggregate only covers the *post-build* groups. CI's separate `source-checks` job runs ~18 source-level scans including `check-file-size`, and I had stopped running those. My verification is now both: `check-build-parallel.sh` **plus** the `source-checks` list from `.github/workflows/build.yml`. Sorry for the wasted cycle.

pirapira:
Re-verified after `52be64bfb refactor(rlp): split RlpItemSpanBody loop machinery under file-size cap` — passing, and armed.

`RlpItemSpanBody.lean` is **1654 -> 1159 lines** and `check-file-size` is `ok`. You split the loop machinery into a sibling module, which is exactly the pattern the gate asks for.

The census conflict did not recur — your rebase already carries main's registry, so `Routines.lean` merged cleanly and only `PROGRESS.md` needed resolving (generated, regenerated to fixpoint).

Verified on the pushed head against **both** CI gate sets:

- `source-checks` — all 17 `ok`, `check-file-size` among them
- post-build aggregate — `codegen 6/6`, `guestaddrs-starts`, `asm-to-program`, `reports 4/4`, `axioms`, `arithmetic-fuzz`, EXIT=0
- `lake exe axiomsweep --check` — EXIT=0, `sorryAx-tainted: 0` (the whole-library backstop from #11942, which landed while this was parked)
- `lake build` green (4739 jobs, twice across the regen fixpoint), tree clean

Thanks for the quick turnaround on the split.

pirapira:
CI failed again — re-parking with `complicated-merge`. Auto-merge disarmed first.

Different failure from the file size, and it is a real consequence of what this PR achieves. The failing step is **Registry cross-check**, which CI runs as `--self-test` first:

```
$ scripts/check-registry-crosscheck.sh            # plain
  ok
$ scripts/check-registry-crosscheck.sh --self-test
  EXIT=1
check-registry-crosscheck --self-test: canary rlp_item_span is no longer an
.unproven Correspondence row; pick a new canary (and update the Lean
negative-control `example` in Routines.lean, which uses the same one)
```

`origin/main` passes the same self-test. The self-test is a negative control: it needs a row that genuinely *is* `.unproven` to confirm the gate can still detect one. This PR proves `rlp_item_span`, so the canary it relied on is gone — the gate is telling you the PR succeeded, and the fixture needs to move with it.

I have not picked a replacement. Which row should stay `.unproven` is a roadmap judgement — choosing one that another in-flight PR is about to prove would just break the canary again — and it also needs the matching negative-control `example` in `Routines.lean` updated. That is your call, not a merge-side fix.

Everything else is green on the merged head: all other source-checks `ok` (including `check-file-size` — your 1654 -> 1159 split holds), post-build aggregate EXIT=0, `axiomsweep --check` EXIT=0 with `sorryAx-tainted: 0`, `lake build` clean at 4739 jobs.

**My process gap again, and I want to name it precisely:** CI runs several gates as `--self-test` *then* plain — `check-retired-predicates`, `check-correspondence-deps`, `check-routine-liveness`, `check-registry-crosscheck`, `check-specref-citations`. I was running only the plain forms, so a broken negative control was invisible to me. I am now running both forms for every gate that has one. Sorry for the second cycle on this PR.

pirapira:
Re-verified after `9e2a1bd33 fix(progress): retarget registry crosscheck canary to bal_canonical_sort` — passing, and armed.

The negative control works again, which is the part that matters:

```
$ scripts/check-registry-crosscheck.sh --self-test
check-registry-crosscheck --self-test: OK — gate fires when bal_canonical_sort
is witnessed while .unproven
  -> EXIT=0
```

You picked a row that genuinely stays `.unproven`, so the gate can still detect the condition it exists to catch — rather than just making the self-test quiet.

Verified on the pushed head against the complete CI gate set:

- **5 `--self-test` forms** — `retired-predicates`, `correspondence-deps`, `routine-liveness`, `registry-crosscheck`, `specref-citations` — all `ok`
- **18 source-checks** plain — all `ok`, including `check-file-size` (your 1654 -> 1159 split)
- **post-build aggregate** — six groups, EXIT=0
- **`lake exe axiomsweep --check`** — EXIT=0, `sorryAx-tainted: 0`
- `lake build` green (4737 jobs, twice across the regen fixpoint), tree clean

Thanks for both turnarounds on this one — the split and the canary.

PR #11939, opened 2026-08-10
refactor(11512): brOff-retarget 70 long-B edges onto 8 labeled epi heads

### LINKED ISSUE #11512: codegen: 374 latent hardcoded long branches without named-target invariant

pirapira:
## Summary

Surgical follow-on to #11927 (label prep). Retargets **70** bare long-B edges that land on the eight named epilogue heads to `brOff (GuestAddrs.SYM + tgt) (GuestAddrs.SYM + cur)`.

**Ratchet:** `EXPECTED_BARE_B_SITES` **872 → 802** (declared migration).

### What this buys
- 70 edges are now geometry-checked against named heads (fixture↔Lean pairing).
- Per-head: `tgg_ok` 33, `hewr` 18, `herr` 7, `hesr` 5, `mee_st1` 3, `tgg_st3` 2, `mee_st2` 1, `rfu_st1b` 1.

### What this does **not** buy
- **802** bare long-B sites remain bare (including 9 unlabeled mid-body joins inside `eip8037_tx_gas_gate`).
- Does **not** close #11512.
- Does **not** full-reconvert the six fixtures (that path is 793; not taken).

### Byte identity
Emitted guest `.text` **unchanged**:
- base @ `092fe7c67` and cand `.text` sha both `8cb13c75c572db416b875bd0edfd08684b42aa4f5b650d482bfbd9bec4386889`
- `cmp` identical

### Files
- `BlockVerdictGasGate.lean` (35), `HeaderFields.lean` (5+7+18), `MptInternal.lean` (4), `Tx.lean` (1 wrapper)
- `scripts/asm_to_program.py`: ratchet 802 + drift gate accepts partial B-naming (strip `brOff`→bare, match threshold form); per-site geometry remains authority

### Gates
- `check-asm-to-program` CLEAN (bare B ratchet 802)
- `check-build-parallel` PASS (codegen 6/6, asm-to-program, guestaddrs, reports, axioms, arithmetic-fuzz)
- residual literals OK

Refs #11512 (progress only; not closed).


pirapira:
**Labelled.** The byte-identity claim is the load-bearing one and it holds: base at `092fe7c67` and the
candidate both emit `.text` sha `8cb13c75c572db416b875bd0edfd08684b42aa4f5b650d482bfbd9bec4386889`. So
`brOff (SYM + tgt) (SYM + cur)` encodes exactly what the bare literal did, which is what makes this a free
invariant rather than a change.

⭐ **And the ratchet moves DOWN, which is the point.** `EXPECTED_BARE_B_SITES` 872 → 802 as a declared
migration. Freezing at 872 only prevented growth; every edge converted here is one a body-length edit can
no longer silently retarget, and the gate now enforces the lower number.

⭐ **`_strip_broff_to_bare` in `asm_to_program.py` is the part I want to single out**, because it is what
makes a *partial* migration checkable at all: the drift gate strips `brOff (base+tgt) (base+cur)` back to
the equivalent bare BitVec-13 before comparing, so a module may name a subset of its long-B sites while
leaving the rest bare and the fixture↔Lean pairing still verifies. Without that, a surgical retarget would
have forced an all-or-nothing conversion of each file — which is precisely the 793 variant I declined.

⚠️ **Kept honest about what it does not buy**, and this matters more than the 70: **802 bare sites remain**,
including 9 unlabeled mid-body joins inside `eip8037_tx_gas_gate`, the retargeting hazard survives for all
of them, and **#11512 is not closed**. Per-head accounting is stated (`tgg_ok` 33, `hewr` 18, `herr` 7,
`hesr` 5, `mee_st1` 3, `tgg_st3` 2, `mee_st2` 1, `rfu_st1b` 1 = 70), so the arithmetic is self-checking.

ⓘ Scope note for the record: I authorised the **802 variant only** while the maintainer is offline. The
**793** full reconvert of the six fixtures is a scope decision left to them, and the 9 unlabeled mid-body
joins this PR surfaces are the natural next tranche if it is taken.

PR #11942, opened 2026-08-10
feat(scripts): kernel-level axiom sweep with committed regression baseline

alexanderlhicks:
## What

Adds `lake exe axiomsweep`: kernel-level axiom / `sorry` accounting for the whole library, with a committed regression baseline.

The tool loads the built `.olean` environment (the same data the kernel checked) and computes, for every declaration under the swept root modules, the set of axioms it transitively depends on — i.e. the `#print axioms` information, library-wide, in one pass. Because it reads elaborated data rather than source text, it sees exactly what the kernel accepted: private declarations, macro-generated declarations, and instances included; no regex heuristics.

## Why

`scripts/check-axioms.sh` stays **authoritative for policy** — this PR does not touch it. Its own header names the gap this fills: *"A broader sweep of all of EvmAsm/ is future work."* The witnessed gate audits the 153 progress-registry witnesses; the sweep covers all 46,394 declarations, including private and macro-generated ones no source scan sees, and gates only on *regressions* against a committed baseline (seeded green on the current tree, per the steering-review rule that a gate red-lighting day one is false-positive friction).

The day-one census pays for itself: the no-`sorry` standard turns out to hold **kernel-wide**, not just on the witnessed surface — and the sweep pinned a trust surface that was previously invisible (below).

## How

- **`scripts/AxiomSweep.lean`** (+ `lean_exe axiomsweep` in the lakefile): memoised traversal of the compiled environment; only imports core `Lean`, runtime-imports the built oleans, so it needs a completed `lake build` and adds no dependencies. Runs in seconds.
- **`scripts/axiom_baseline.json`**: the currently-known `sorryAx`-tainted declarations (and any non-standard-axiom dependents). `--check` fails **iff a declaration is tainted that the baseline does not list**; closing gaps stays green and prints a nudge to shrink the baseline via `--update-baseline`.
- **CI**: one report-only step after the build (`::warning` on regression). Flip to enforcing by deleting the `|| echo` once the baseline has soaked for a few PRs.
- Docs updated in the same PR.

## Modes

```
lake exe axiomsweep                     # summary
lake exe axiomsweep --out report.json   # full per-declaration report (name, module, kind, line, axioms)
lake exe axiomsweep --check             # regression gate against scripts/axiom_baseline.json
lake exe axiomsweep --update-baseline   # refresh the baseline; commit the diff
lake exe axiomsweep --root Foo          # override the swept root modules (repeatable)
```

## Baseline at this commit

- **46,394 declarations across 2,757 `EvmAsm.*` modules**; sweep runs in seconds on a built tree.
- **0 sorryAx-tainted** — the repo's no-`sorry` discipline holds across the whole library.
- **74 declarations (73 in `EvmAsm.Rv64.SailEquiv.*`) rest on four vendored Sail platform axioms**: `sys_enable_experimental_extensions`, `plat_term_write`, `load_reservation`, `match_reservation`. Expected for the Sail correspondence layer (the vendored model axiomatizes platform primitives), but now recorded by name in the baseline instead of invisible — and any *new* declaration picking up such an axiom fails `--check`.
- Spot-checked against `#print axioms` (`Rv64.SailEquiv.beq_sail_equiv` → `sys_enable_experimental_extensions` present, matching the sweep exactly).
- **Both gates verified green side by side** on this tree: `check-axioms.sh` (153 witnesses, classical only) and `axiomsweep --check`.
- Advisory tier: report-only CI step, documented in the AGENTS.md gate table; promote by dropping the `|| echo` in `build.yml` after the soak.

## Notes

- Same tool as ArkLib's `alh/axiom-sweep` branch (multi-root variant); the copies are intentionally self-contained per repo, mirroring how other scripts are shared across the org. If it proves its keep everywhere, extracting a shared home is a possible follow-up.
- `lean4export` was considered and deliberately not used: it targets external kernel checking, has no source positions, and would add a per-toolchain pinned dependency; walking the environment yields the same information with line numbers and zero new deps.
- The per-declaration report is not committed anywhere; only the baseline is tracked.

## Adversarial review

Five independent adversarial reviewers (one per repo deployment) attacked the tool, wiring, and baselines before this PR was opened. Everything they confirmed is fixed in the follow-up commit on this branch:

- **Collector defect (HIGH, confirmed in 4 repos)**: the one-pass DFS finalized self-referencing constants (every inductive/ctor pair) prematurely, producing per-declaration rows that diverged from `#print axioms` in both directions, with constructible shapes where a new `sorry` passed `--check`. Fixed with a fixpoint-repair phase computing the true kernel closure (now strictly *more* accurate than `#print axioms`, whose own `CollectAxioms` wobbles inside mutual families). Post-fix rows spot-verified against ground truth; sorry-sets unchanged everywhere.
- Axiom *types* now traversed (`CollectAxioms` parity); duplicate `constNames` rows deduped; `native_decide` axiom names normalized to their owning declaration (the `ax_N_M` counters are `Elab.async`/toolchain-volatile); `--check`/`--update-baseline` mutually exclusive; bare `Lean.ofReduceBool`/`Lean.trustCompiler` are never baselinable (floor).
- CI steps now distinguish taint findings (report-only during soak) from infrastructure failures (always fail).
- Known blind spots documented: structure-field defaults and `example`s never enter any environment walk (`#print axioms` included); unimported files are invisible — paired with each repo's import-completeness gate.

What survived attack unchanged: baseline determinism (byte-identical regeneration), all headline numbers, spot-checks vs `#print axioms`, gate exit-code semantics, and the CI/step placement.


pirapira:
**Labelled.** Checked the claims that carry weight rather than the summary.

**1. The gap it fills is real and self-declared.** `check-axioms.sh:53-54` says in its own words that it is
*"a partial enumeration of the verified guest surface (a broader sweep of all of EvmAsm/ remains future
work)."* This is that sweep, and it correctly leaves `check-axioms.sh` **authoritative for policy** rather
than replacing it. Two gates with different populations — 153 witnesses versus 46,394 declarations — is the
right shape, because the witnessed gate is the one that encodes *what we promised*.

**2. Reading elaborated data instead of source text is the property that matters.** Every axiom/tactic
guard we have is a source scan, and every source scan tonight has had a blind spot: `check-registry-coverage`
could not see `_fnspec` (#11903), a coverage gate ran over an empty population, a census missed inline arms.
A traversal of the compiled environment sees private, macro-generated and instance declarations that no
regex reaches, which is a strictly stronger instrument for this question.

**3. `sorry` = 0 kernel-wide** — `{"sorry": []}` in the committed baseline. That upgrades a claim we have
been making about the witnessed surface into one about the whole library. Worth having stated once, with a
tool that can re-check it.

⭐ **4. The trust surface it pinned is the actual finding, and it is the part I would have wanted a PR for
on its own:** 74 declarations (73 in `EvmAsm.Rv64.SailEquiv.*`) rest on **vendored Sail platform axioms** —
`sys_enable_experimental_extensions` and `plat_term_write` among them. Those were previously invisible:
`CLAUDE.md` has said the trusted base is "only the three classical axioms", and that is true of the
witnessed set while **not** true of the library. A four-axiom platform surface confined to `SailEquiv` is
defensible; not knowing it was there was not.

⚠️ **5. Baseline seeded green, gate report-only.** Both choices are right for day one — a gate that
red-lights on arrival trains people to ignore it — and the promotion path is one deleted `|| echo` in
`build.yml`. But it means **this gate has not yet been seen to fail**, and a checker nobody has seen fail is
indistinguishable from one that cannot. The PR reports an adversarial review that found a collector defect
(self-referencing constants finalised by a one-pass DFS) in four repos, which is the right kind of evidence;
what I would still want before promotion is one deliberate regression — taint a declaration locally,
confirm `--check` fails, revert. Not a blocker for an advisory tier.

ⓘ Not to be confused with the two existing gates it sits beside: `check-forbidden-tactics.sh` (source scan
for `bv_decide`/`native_decide`) and `check-axioms.sh` (kernel truth on witnesses). Three instruments, three
populations, and the AGENTS.md table now says which is which.

pirapira:
Synced onto main (97 behind -> 0; sole conflict `PROGRESS.md`, generated) and verified. Auto-merge armed.

I ran the gate this PR introduces, against the tree it lands in:

```
$ lake exe axiomsweep --check
axiomsweep: 46648 declarations across 2765 modules under #[EvmAsm]
  sorryAx-tainted: 0
  non-standard-axiom-tainted: 74 (axioms: #[sys_enable_experimental_extensions,
                                            plat_term_write, load_reservation, match_reservation])
axiomsweep: check passed (no new axiom/sorry taint).
  -> EXIT=0
```

That 74 matches the `EvmAsm.Rv64.SailEquiv.*` count your `CLAUDE.md` note pins, and the four axioms are exactly the documented vendored-Sail platform hooks — so the exception is scoped as described rather than open-ended. Zero `sorryAx` across 46,648 declarations is a materially stronger statement than `check-axioms` alone, which only covers the witnessed registry surface.

Full local verification on the pushed head, **both** CI gate sets:

- `source-checks` (all 17): every one `ok`, including `check-forbidden-tactics` and `check-sail-pin`
- post-build aggregate: `codegen 6/6`, `guestaddrs-starts`, `asm-to-program`, `reports 4/4`, `axioms`, `arithmetic-fuzz` — all PASS, EXIT=0
- `lake build` green (4735 jobs, twice across the regen fixpoint), tree clean

Worth noting I only started running the `source-checks` set again this round — I had been relying on `check-build-parallel.sh` alone, which covers post-build groups only. That gap let #11936 reach CI with a `check-file-size` failure. Verifying this PR against both sets is the corrected procedure.

PR #11945, opened 2026-08-10
chore: remove unlinked block body probes

pirapira:
## Summary

Remove the unlinked block-body probe cluster as a test-only cleanup. The production `stateless_guest` has no dependency on these units, so this removes dead probe scaffolding without changing the guest image.

The deletion is all-or-nothing across the 22 censused roots:

- `Block.lean`: block-body decode, empty-ommers, transaction-count, blob-gas-total, and blob-gas-consistency probes.
- `BlockBody.lean`: all six body/chain summary and count probes (file deleted).
- `BlockValidate.lean`: 2tx body extraction, 2tx-with-body validation, and no-withdrawals pair probes.
- `BlockValidate1Tx.lean`: 1tx body extraction and 1tx-with-body validation probes.
- `BlockEmpty.lean`: empty-block, empty-block-with-parent, and empty-block-chain probes.
- `Withdrawal.lean`: withdrawals-total probe.
- `WithdrawalBlockSummary.lean`: both withdrawal count/summary probes (file deleted).

The corresponding imports, Registry/RegistryMain/RegistryTail lookup entries, RegistryNames lists, README/docs coverage rows, and the synthetic bmvmx invocation of the deleted tx-count probe were removed together. Production/non-cluster helpers remain.

## Evidence

Prediction: because census found no GuestAddrs/TSV/linked production roots, the guest must be byte-identical. The rebuilt artifact confirms it:

- base commit `092fe7c67578f3b9d391d1dc46a2603113368d1c`
- base/current production ELF: `fada9ea40962b0dbae36a1347bae7e36be5cc9d4a1d7df3efac53dffb72a399d`
- `.text`: 340808 bytes, `8cb13c75c572db416b875bd0edfd08684b42aa4f5b650d482bfbd9bec4386889` before and after
- `.data`: 21264 bytes, `c73b33e51ae6fbdd4c46a8c802d7391fc3589ac3cd88b2696fe48665ffda08b4` before and after
- `.bss`: 456702112 bytes

Thus the removal has zero production guest-image effect; no pin movement was needed.

## Checks

- `lake build EvmAsm` — passed (4731 jobs).
- `LAKE_ARTIFACT_CACHE=false scripts/check-build-parallel.sh` — passed: codegen 6/6, reports 4/4, axioms, link/layout, guest-address, asm-to-program, and arithmetic-fuzz stages all green.
- `scripts/check-asm-to-program.sh` — clean.
- `scripts/check-region-map.sh` — clean.
- `scripts/check-unimported.sh` — 2761/2761 reachable, 0 orphans.
- `scripts/check-file-size.sh` — 2162 files within cap.
- `git diff --check` — clean.
- Regeneration reached a no-op fixpoint for GuestAddrs, GuestImageEntries, RegionMapLinkPins, and symbol-addresses.tsv.

No merge label applied.


pirapira:
**Labelled.** The prediction-then-measurement structure is what makes a 3,212-line deletion reviewable, and
it came out right.

**The safety argument is one comparison, and it holds.** The census predicted that because no
`GuestAddrs` entry, no `symbol-addresses.tsv` row and no linked production root reaches these units, the
guest **must** be byte-identical. Measured on the rebuilt tree:

- `.text` **340808 bytes, `8cb13c75c572db416b875bd0edfd08684b42aa4f5b650d482bfbd9bec4386889` before and after**
- `.data` **21264 bytes, `c73b33e51ae6fbdd4c46a8c802d7391fc3589ac3cd88b2696fe48665ffda08b4` before and after**
- production ELF `fada9ea40962b0dbae36a1347bae7e36be5cc9d4a1d7df3efac53dffb72a399d`, no pin movement needed

⇒ a falsifiable prediction, stated first and then confirmed. Had either sha moved, the census had missed a
root and the deletion would have been unsafe — which is exactly why I asked for the comparison rather than
a green build.

**All-or-nothing was respected**, and that mattered: deleting a target alone breaks Lean compilation
because the prologue and composite bundles still reference it. `+12 −3212` across 17 files, with
`BlockBody.lean` and `WithdrawalBlockSummary.lean` removed outright and the `Registry`/`RegistryNames`/
`RegistryMain`/`Imports` entries taken with them. No stubbing to keep the build green — the failure mode this
scoping existed to prevent.

⭐ **Gates run as the aggregate rather than a hand-rolled list**: `check-build-parallel.sh` with
`LAKE_ARTIFACT_CACHE=false` (codegen 6/6, reports 4/4, axioms, link/layout), plus `check-asm-to-program`
and `check-region-map` clean, and `lake build EvmAsm` at 4731 jobs. Two PRs were handed back today by gates
neither author nor reviewer ran locally; this one ran the runner that covers them.

ⓘ For the record on *why* this is a deletion rather than a migration: `22pwv.6.1.1` established that
full-guest EEST can never exercise these routines — they have no linked root, so `asm_to_program` emit-lean
fails `BLOCKED_ON_.6` on cross-function JAL with no known entry PC, and a dedicated probe parity was
correctly refused as verified-guest progress at the time. Code that cannot be reached by the corpus and
carries no proof-side tie is scaffolding, and the maintainer ruled it should go.

⚠️ Docs were included in the same PR, which is right: leaving `README` or
`docs/4ch8f-asm-to-program-coverage.md` describing a deleted probe is the half-removed tombstone class
(#11556). I will close the bead once this merges, not on the diff.

Issue #11925, opened 2026-08-10
tooling: mechanize the startable-frontier census (scripts/proof-frontier.py) — three wrong hand-annotations in one week each cost a mis-aimed proof issue


pirapira:
> 🤖 Generated by Claude Code (strategist pass, 2026-08-10, main at 5b2fc8114).

## The failure pattern: hand-maintained proof-targeting tables carry wrong annotations, and each error costs a mis-aimed issue

Three instances in one week, all in `docs/leaf-routine-targets.md`:

1. `derive_withdrawal_requests` / `derive_consolidation_requests` annotated **"(leaves)"** — actually seven-instruction stubs that tail-jump into `stage_system_call` → the whole interpreter. #11578 was filed on that annotation and had to be retargeted to `execution_requests_hash`.
2. The runners-up bullet claimed the field towers had "no assertion vocabulary at all" and named the wrong prime — corrected in #11676 after #11574's executor found both triples already existed.
3. #11575 was filed as "8 mechanical forks missing" — all 8 triples already existed, sorry-free, landed *before* the row-9 pattern they were supposedly waiting on.

The doc's own method section is sound (transitive `jal` closure over `scripts/asm-fixtures/*.s`, liveness via `symbol-addresses.tsv`, verified set from `Routines.lean`) — it is the **hand execution** of that method that drifts, and the doc even self-flags that its counts go stale.

## Ask: mechanize the census the doc already specifies

A script (suggest `scripts/proof-frontier.py`, read-only, in the family of `check_routine_liveness.py`) that emits, from the current tree:

1. **The startable frontier**: symbols whose transitive `jal` closure lies within the witnessed set of `Progress/Routines.lean` (parse the registry, or consume a generated witness list) — minus symbols already witnessed.
2. Per frontier symbol: in-degree (callers), whether an existing theorem file matches (`grep` for `<symbol>_spec`/`_fnspec` — instance 3 above is exactly what this column prevents), and whether a `Correspondence.lean` row exists.
3. A staleness gate in the style of `check-obligation-blockers.sh`: fail if `docs/leaf-routine-targets.md`-successor tables claim "leaf"/"startable"/"no spec" for a symbol where the script disagrees.

## Why now

[...]

PR #11947, opened 2026-08-10
fix(progress): register account_extract_nonce as .conditional (grade inherited from its callee)

pirapira:
### Register `account_extract_nonce` as `.conditional` (grade inherited from its callee)

Issue #11925 continuation: the `proof-frontier.py` three-state census (PR #11934)
surfaced `account_extract_nonce` in the **present-but-unrowed** bucket — a complete
whole-routine triple `account_extract_nonce_spec_within` exists
(`EvmAsm/Codegen/Programs/AccountAccessorNonceSpec.lean:334`), yet it had no row in
the routine registry.

#### Grade: `.conditional` — and the reason matters

This row is graded `.conditional`, **not** `.proven`, for a specific reason: the
gate is **inherited from the callee**. `account_extract_nonce` calls
`rlp_content_to_u64`, whose row is *already* `.conditional` at
`EvmAsm/Progress/Routines.lean:204`:

```
(gate := "`a.nonce < 2 ^ 64` — the accessor's u64 output width, narrower "
  ++ "than `Account.nonce`'s own `< 2 ^ 256` invariant")
```

The top-level triple repeats that exact hypothesis (`hnonce : a.nonce < 2^64`),
and the gate prose has been copied **verbatim** into the new row so the two rows
visibly match. Two structurally identical gates cannot carry different tiers —
that would be the inconsistency the sibling-grading discipline exists to prevent.

The satisfying instance is trivial: any `Account` whose nonce fits a u64 cell
(e.g. `a.nonce = 0`).

#### Why this is not the sibling `.proven` registrations (PR #11940)

Its sibling `account_extract_balance` was registered `.proven` in #11940: its only
value-shaped hypothesis is `hnonce : a.nonce < 2^256`, which is the **natural
Word256 bound** — and its callee `rlp_content_to_u256_be` is `.proven`
(`Routines.lean:207`) with the same gate. The width narrowing (`2^64 < 2^256`)
is what makes the nonce accessor a genuine input-domain restriction while the
balance accessor is not.

#### Changes

1. **`EvmAsm/Progress/Routines.lean`**: adds a `.conditional` row for
   `account_extract_nonce` (with the verbatim inherited gate prose + inheritance
   note citing `Routines.lean:204`), the `_account_extract_nonce_routine_witness`
   abbrev, an import of `AccountAccessorNonceSpec`, and count bumps:
   - `routineCount_eq` 39 → 40
   - `routineConditionalCount_eq` 9 → 10
   - `routineSymbols_eq` 29 → 30
   (`routineProvenCount_eq` unchanged at 30; `routinePartlyCount_eq` unchanged at 0.)
2. **`EvmAsm/Progress/AxiomWitnesses.lean`**: regenerated via
   `scripts/gen-axiom-witnesses.py --write` (now 163 witnesses; one new
   `#print axioms` line for `EvmAsm.Codegen.account_extract_nonce_spec_within`).
3. **`scripts/registry-coverage-allow.txt`**: drains the now-stale tier-A entry for
   `account_extract_nonce` (its reason text claimed "registrable as .proven" — now
   superseded by the `.conditional` verdict); header counts 19 → 16 (A), 83 → 75 (B).

#### Verification

- `lake build EvmAsm.Progress.Routines` — **clean** (2470 jobs).
- `scripts/check-registry-coverage.py` — OK (973 linked, 38 registered, 120
  spec-bearing, 91 uncovered / 91 allowlisted).
- `scripts/check-progress.sh`, `scripts/check-drift.sh` — exit 0.
- `check-axioms.sh` **could not run** in this checkout: the artifact-cache mode
  (`LAKE_ARTIFACT_CACHE`) does not materialize oleans, so `lake env lean` fails
  with "object file '….olean' does not exist" (issue #10537). Substituted
  `lean_verify` on the row's proofRef, which passes with axioms exactly
  `[propext, Classical.choice, Quot.sound]` and no sorries. CI runs the full
  axiom gate in a non-cache environment and is authoritative.

Not touched in this PR: `tx_type_dispatch` (held — PR #11929 modified the
routine) and the four `.proven` accessors now registered in #11940. That PR and
this one are separate branches based on `origin/main` and do not overlap.


pirapira:
Synced onto main (48 behind -> 0) and verified. Three conflicts, all resolved by deriving the merged state rather than picking a side — flagging since I made the edits.

**1. `EvmAsm/Progress/Routines.lean` imports** — kept both sides (your `AccountAccessorNonceSpec`, main's `AccountAccessorTopSpec` / `AccountIsEip161EmptyClose6` / `ReceiptExtractLogsBloomSpec`).

**2. The census theorems.** Derived from the merged registry rows: **45 total / 34 proven / 11 conditional / 35 symbols**. Worth noting the merge only conflicted on `routineCount_eq` and `routineProvenCount_eq`; `routineConditionalCount_eq` sat just *outside* the conflict region and silently kept main's `10`. The kernel caught it:

```
Routines.lean:532: Tactic `decide` proved that the proposition
  routineCountTier ProofTier.conditional = 10
is false
```

Corrected to 11, which matches the derivation. All four literals are `by decide`, so the clean build is the proof they are right.

**3. `scripts/registry-coverage-allow.txt`** — this one was subtle: **each side of the conflict was a removal**. Your branch drops `account_extract_nonce` (this PR rows it); main drops `account_extract_balance` and `account_is_eip161_empty` (rowed by #11940, which landed earlier today). Keeping either side alone would have resurrected the other's retired entries, so I took the **union of removals** — all three gone — and set the TIER A header count to the actual remaining entries: **13**. `check-registry-coverage` passes.

No statement or definition was changed; only literals a merge has to reconcile.

Verified on the pushed head against the complete CI gate set: 5 `--self-test` forms, 18 source-checks, post-build aggregate EXIT=0, `axiomsweep --check` EXIT=0 with `sorryAx-tainted: 0`, `lake build` green (4739 jobs, twice across the regen fixpoint), tree clean. Auto-merge armed.

PR #11948, opened 2026-08-10
proof(progress): register the rlp_encode_list_prefix long1 arm — proven but ungated

dhsorens:
Addresses #10780.

## A proven triple that was outside the axiom gate

`rlp_encode_list_prefix_long1_pinned_spec_within` — the **1-length-byte long form** (`56 ≤ len < 256`: writes header bytes `[0xF8, len]`, sets the cell flag to 2) — has been proven in `RlpSpliceHelperSpec.lean` since the short arm landed, but it was **never registered** in `Progress/Routines.lean`. Consequences:

- it was **not covered by `check-axioms.sh`** (no witness abbrev forced it), so nothing was checking its axiom footprint;
- `routineCount` **undercounted** the verified guest surface;
- and the registry did not reflect that `len < 256` is covered, which makes the routine look weaker than it is.

The short row's own `gate` text already says the cut is at "`lenlen ≥ 2` long forms" — a description that only makes sense if `lenlen = 1` is done. The registry and its own prose disagreed.

This is the same situation as the **#11291** note a few lines below in the same file ("the whole-routine triple already existed […] but was never registered"), so this follows that precedent.

## What this PR does

- adds the `.conditional` row for the long1 arm, with its gate stated as `56 ≤ len.toNat < 256`
- adds the witness abbrev, which is what puts it inside the axiom gate
- bumps the kernel-checked counts: `routineCount` 39 → 40, `routineCountTier .conditional` 9 → 10
- regenerates `AxiomWitnesses.lean` (via `scripts/gen-axiom-witnesses.py --write`, +1 witness) and `PROGRESS.md` (via `scripts/progress-report.sh --write`)

**Registering existing work — no new proof, no statement change, no emitted bytes.** `routineSymbols` stays at 29 because `rlp_encode_list_prefix` was already present.

## Where this leaves #10780's item 3

The encode-side long-form residual is now precisely the **`lenlen ≥ 2`** arms — the ones behind the `SLLI`-widened comparisons (`256`, `65536`, … up to 8 bytes). Those are real proof work: the routine implements them (lenlen computed 1..8, header `ADDI x29, x28, 247` = `0xf7 + lenlen`, then length bytes emitted from `lenlen - 1` downward, **minimal by construction**), and covering them is what discharges the issue's item-1 concern about a length-of-length carrying leading zeros.

One note for whoever takes `lenlen = 2`: unlike long1, its length-byte loop iterates **twice**, so it needs the loop treated as a loop rather than unrolled once — that is the step up in difficulty between the two arms, not the header arithmetic.

## Gates

Reported after this PR's own build; see the checks below. Locally: `lake build` (0 warnings), `check-axioms.sh` exit 0 with the new witness classical-only, and the `check-*` suite plus `check-registry-coverage.py` / `check-specref-citations.py`.


dhsorens:
Two follow-ups I checked before asking anyone to review this, both of which could have made this PR wrong.

## 1. Is the long1 row redundant? No — there is no general triple to subsume it

If `rlp_encode_list_prefix` had a whole-routine triple covering all forms, a per-form row would be noise. It does not: the only two theorems are `rlp_encode_list_prefix_short_pinned_spec_within` (`:762`) and `rlp_encode_list_prefix_long1_pinned_spec_within` (`:917`), and **neither is composed into a general one**. So long1 is an independent per-form certificate — exactly the "one row per theorem" shape `Routines.lean`'s header prescribes for a per-form routine.

A consumer already leans on it in prose: `MptSpliceSlotSpec.lean:90` cites long1 by name for the `[0xF8, len]` header bytes. That is a second reason it belongs inside the axiom gate rather than outside it.

## 2. Why `check-registry-coverage.py` did not catch this, and whether more are hiding

The gate (#11637) keys on **symbol**: it fires when a linked symbol with a spec has no row in *either* registry. `rlp_encode_list_prefix` already had a row from the short arm, so at symbol level it looked covered — the gate is working as designed and could not see this.

That is a genuine narrower blind spot than #11637's, because `Routines.lean` says in its own header that "`symbol` **groups** rows; it is not a key". A symbol can be registered by one arm while its other proven arms stay ungated.

I swept for it rather than asserting it. Every `*_pinned_spec_within` theorem outside `EvmAsm/Progress/**` — the repo's naming convention for a per-form routine-level certificate — cross-checked against the `proofRef`s of both registries:

- **5 candidates**, of which **long1 was the only real gap** (fixed here).
- The other 3 are `rlp_item_size_single_/short_string_/short_list_pinned_spec_within`. These are **not** gaps: they are the internal per-form steps that `rlp_item_size_form_own_spec_within` dispatches over, and that is composed into the already-registered `rlp_item_size_spec_within`. Registering them would add rows for facts the registered triple already implies — inflating the count, which is the opposite of what the registry is for.

⇒ **After this PR the per-form residual is empty.** So I have not filed an issue for the blind spot: the backlog behind it is currently zero, and a gate for an empty class would be cost without signal. Recording it here instead so that if a future per-form arm lands unregistered, the reasoning and the one-off sweep are already written down. Happy to turn the sweep into a gate if you'd rather have it ratcheted — it composes with `check-registry-coverage.py`, keyed on `proofRef` membership instead of symbol membership.

PR #11949, opened 2026-08-10
feat(proofs): add concrete Keccak accelerator seam

pirapira:
## Summary

This lands the proof-only Keccak accelerator seam for issue 11924.

- HashBridgeProg emits CSRS 0x800 (2048), not ECALL; the flat 69-instruction wrapper uses two inline CSRS sites.
- The machine-level seam goes through concrete Accel.keccakF. Execution.step dispatches valid CSRS to execInstrBr, step_csrs proves the one-step transition, and csrs_step_spec_within frames the concrete writeWords payload. No axiom or new cps vocabulary is needed.
- SpecRef.Crypto.keccak256 is built from the same concrete Accel.keccakF.
- The zero-remainder padding fact and a two-adjacent-CSRS composition theorem are included.
- No emitted program was changed.

## Hypothesis classification

Scratch validity and alignment, output validity, input-output disjointness, input pointer alignment, and MAX_INPUT_BYTES = 0x37FFFFF8 are ABI or resource facts. MAX_INPUT_BYTES is an ABI envelope, not a proof artefact or proof-convenience cap. The nSteps bound is a function of input length. There is no smaller ad hoc domain restriction; the target classification is proven.

## Explicit remaining scope

The wrapper loop framing is intentionally not complete. The flat 69-instruction Program has inline CSRS, while Stmt.callRegS emits a JALR to a separate handle, so the callRegS shape cannot flatten byte-identically. The remaining route is a proof-only structured decomposition; this PR does not restructure emitted code.

The literal ECALL route remains blocked on concrete ECALL state-transition semantics: Execution.step treats a non-special ECALL as PC+4, while executeKeccakEcall is parameterized by an arbitrary accelerator function and is not tied to Accel.keccakF. Instantiating it for Keccak would be an axiom-shaped gap.

No registry row is added for the unfinished wrapper.

## Byte-neutrality evidence

Clean isolated worktrees were built with lake exe codegen --program stateless_guest --halt linux93, using base commit 6dbfd3d24 and candidate commit 15fe4aa43.

- Production .text: 340820 bytes; base SHA256 65983da27d1b5132ec202cd425564aa21d185a7e689889ac9a4f7b2996c5e60d; candidate SHA256 65983da27d1b5132ec202cd425564aa21d185a7e689889ac9a4f7b2996c5e60d.
[...]

Issue #11772, opened 2026-08-08
proof: evm_keccak256_stack_spec — tie the KECCAK256 handler's memory slice to the pushed hash (residual of #111)


pirapira:
> 🤖 Issue generated by Claude Code.

Residual of #111, which is closed: its **implementation criteria are all met** and only the top-level proof remains.

## The obligation

**`evm_keccak256_stack_spec`** — prove that the emitted `h_KECCAK256` handler pushes `keccak256(memory[offset .. offset+size])` onto the EVM stack, with the hash **abstract** (an opaque function, not a computed digest), 0 `sorry`.

The tree's own status table already records this as open: `EvmAsm/Progress.lean:250` is `entry "KECCAK256" .execSpec none` — `.execSpec`, no spec attached.

## What already exists (do not re-derive)

Verified on `origin/main`:

| piece | where |
|---|---|
| emitted handler, ECALL to the accelerator | `Programs/EvmHashHandlers.lean:42` — `h_KECCAK256`, raw `.4byte 0x80052073` inside `zkvm_keccak256`; args in `a0`/`a1`/`a2`, caller-saved regs restored around the call |
| argument decode from the stack | `Evm64/KeccakArgsStackDecode.lean`; `KeccakArgs.lean:56` `keccakArgs_offset`, `:59` `keccakArgs_size` |
| dynamic gas | `KeccakArgs.lean:74` `keccakDynamicCostFromArgs_eq`, `:78` `…_zero_length` |
| accelerator classification | `Evm64/Accelerators/Dispatch.lean:122` `isAccelerator_keccak256` |
| BE/LE result convention | `EvmHashHandlers.lean:65` records that KECCAK256 results were big-endian on the stack while everything else was little-endian, and what was done about it |

⇒ The supporting layer is complete. What is missing is the statement that ties **memory slice → hash → stack result** together.

⚠️ **The endianness note at `EvmHashHandlers.lean:65` is load-bearing for the spec, not trivia.** A spec written without it will either be false or will silently describe a different word order than the handler produces.

## Scope

- Proof only. **No change to the emitted handler is expected**; if the proof shows the handler is wrong, that is a finding to report before editing, not something to fix inside this issue.
- Abstract hash: the spec relates the pushed word to `keccak256` as an opaque function over the memory slice. It does not evaluate the digest.

⚠️ Priority note: this is **proof-lane work**, which currently sits behind structural and false-reject work.

[...]

PR #11955, opened 2026-08-10
feat: specify KECCAK256 stack bridge

pirapira:
## Summary

- specify the pure KECCAK256 stack bridge over an opaque 32-byte hash function
- read exactly `memory[offset .. offset + size]` and convert digest bytes to the EVM stack-word convention
- keep the KECCAK256 progress row unchanged: this is model-level coverage, not emitted-handler coverage

## Scope and residual

`evm_keccak256_stack_spec` proves the EL bridge's memory-window read, opaque hash application, and digest-to-stack conversion without imposing an arbitrary size cap. It does not prove the emitted `h_KECCAK256` handler: that handler is raw assembly/custom-tail code and is not currently represented as a typed SAsm `Fn`, so there is no `cpsTripleWithin` machine triple to attach here. The emitted-handler proof remains the residual for issue 11772.

The hash remains opaque; this does not unfold `Accel.keccakF`. ABI/resource facts (stack slots, caller-owned memory validity/alignment, and offset-plus-size range) remain outside this pure model bridge.

## Verification

- Base: `c39de0460289d9908a6343944a241dbccb74153a`
- `env LAKE_ARTIFACT_CACHE=false lake build EvmAsm.Codegen.Programs.AccountWriteMap EvmAsm.EL.KeccakStackExecutionBridge EvmAsm.Progress`
- `env LAKE_ARTIFACT_CACHE=false scripts/check-no-warnings.sh` (4,737 jobs, 0 warnings)
- `scripts/check-progress.sh`
- `scripts/check-layering.sh`
- `git diff --check`

### DIFF
--- a/EvmAsm/EL/KeccakStackExecutionBridge.lean
+++ b/EvmAsm/EL/KeccakStackExecutionBridge.lean
@@ -18,6 +18,12 @@ abbrev AcceleratorOutput := KeccakResultBridge.AcceleratorOutput
 abbrev KeccakRequest := KeccakEcallBridge.KeccakRequest
 abbrev KeccakResult := KeccakEcallBridge.KeccakResult
 
+/-- An opaque Keccak result function.  The handler proof must not unfold a
+    particular permutation implementation: its semantic boundary is a
+    32-byte digest for the byte slice supplied to the accelerator. -/
+abbrev AbstractHash :=
+  List EvmAsm.EL.RLP.Byte → KeccakResultBridge.HashBytes
+
 /-- KECCAK accelerator model used by the pure execution bridge. -/
 abbrev Accelerator := AcceleratorInput -> AcceleratorOutput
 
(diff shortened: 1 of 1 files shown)


pirapira:
**Labelled.** One file, +36, on current main — and the important property is what it **does not** claim.

**The registry row is untouched.** `Progress` keeps `KECCAK256` at `.execSpec` (68 proven / 14 execSpec),
and the body says outright that this is *model-level coverage, not emitted-handler coverage*. That is the
right call and it is the same shape as #11938 three hours earlier, where the account-write upsert **model**
landed with no row and no regrade because `account_write_record` is emitted as assembly text with nothing
for `cpsTripleWithin` to attach to. Two agents, same discipline, independently.

⭐ **This matters because the first attempt got it wrong and was corrected.** An earlier revision flipped the
row to `.proven` on the strength of a *definitionally proven* pure bridge with ABI and resource bounds
outside it. A model-level lemma about a function does not discharge #11772's obligation — that the **emitted**
`h_KECCAK256` pushes `keccak256` of the memory window — and `.proven` on the opcode row would have asserted
machine coverage we do not have. Reverted byte-for-byte, and the residual is now named instead: the raw
assembly and custom tail lack a typed SAsm `Fn`/`cpsTripleWithin` seam.

**What is genuinely proved**, and it is worth having: the EL bridge reads exactly
`memoryReadBytes memory offset size`, applies an **opaque** `AbstractHash`, and converts the digest through
`stackWordFromAcceleratorHash` onto `word :: rest`. **No size cap** — the hypothesis classification given
before the proof held, so `offset + size` range validity stayed a resource envelope for caller-owned EVM
memory rather than becoming an input-domain gate.

⭐ **The hash stays opaque and `Accel.keccakF` is not unfolded.** That was a hard constraint rather than a
style choice: a proof about a *specific computed* digest is narrower than the obligation while looking
stronger, and the opaque form is what composes with SpecRef, where `keccak256` is a function rather than a
value.

ⓘ Related, from the same lane and worth reading together: #11949 established that the emitted keccak
accelerator is **`CSRS 0x800`, not `ECALL`** — I decoded `0x80052073` independently (opcode `0x73`,
funct3 `2`, rs1 `x10`, csr `0x800`) — and that the CSRS step needs **no axiom and no new cps vocabulary**.
So obligation 6's recorded blocker ("per-precompile EL bridges not yet codegen-wired") points at the wrong
problem; the literal ECALL path is blocked on **concrete ECALL step semantics**, since `executeKeccakEcall`
is parameterised over an *arbitrary* accelerator function.

